Issue #9889, opened 2013-05-29
[Quest] The Brothers Bronzebeard


TheMaaster:
when i take [The Brothers Bronzebeard] (http://www.wowhead.com/quest=12973) from [Brann Bronzebeard] (http://www.wowhead.com/npc=30107) (or [Brann Bronzebeard] (http://www.wowhead.com/npc=30382) ) nothing happens. 
Looking on youtube, i think it needs plenty scripting. (http://www.youtube.com/watch?v=SblDX-y8Wsc)
and when set to complete, [Velog Icebellow] (http://www.wowhead.com/npc=30401) still missing to turn it in to (phasing?)

grimgravy:
The mission're still bugged by the time it arrives in the area to complete the mission not complete

grimgravy:
The mission're still bugged by the time it arrives in the area to complete the mission not complete

grimgravy:
The mission're still bugged by the time it arrives in the area to complete the mission not complete

grimgravy:
The mission're still bugged by the time it arrives in the area to complete the mission not complete

grimgravy:
The mission're still bugged by the time it arrives in the area to complete the mission not complete

grimgravy:
The mission're still bugged by the time it arrives in the area to complete the mission not complete

grimgravy:
The mission're still bugged by the time it arrives in the area to complete the mission not complete

grimgravy:
The mission're still bugged by the time it arrives in the area to complete the mission not complete

grimgravy:
The mission're still buggy when it comes in the area to complete the mission not complete

Killyana:
Still waiting for a c++ dev to make a PR fixing the dismount issue.

Rushor:
even if we manually add:

machine->GetVehicleKit()->InstallAccessory(NPC_SOLDIER, 2, true, TEMPSUMMON_DEAD_DESPAWN, 0);

the accesoorynpc will cast 30383 on the vehicle and cicks the player out if the second soldier wants to jumb on the vehicle

Killyana:
Script updated, spell 56603 used by the player,  55089 by soldiers, and 43671 by Brann exactly as on retail, but still Brann is ejected when a soldier ride the plane. In this case I can't find the problem.

Killyana:
A solution by @whoami : http://pastebin.com/ESL4w3nJ
[...]

Issue #13298, opened 2014-10-10
DB/Quest When All Else Fails


Pitcrawler:
Currently Ricket is diplaying her gossip option regardless of this quest. So conditions are missing.
The script is missing, too.

Attention: This is not tested yet.

```sql
-- Fix quest When All Else Fails (12862,13060) 

UPDATE `creature_template` SET `unit_flags`=512, `VehicleId`=240, `InhabitType`=4 WHERE `entry`=30477;
UPDATE `creature_template` SET `minlevel`=80, `maxlevel`=80, `exp`=2, `unit_flags`=512, `VehicleId`=240, `InhabitType`=4 WHERE `entry`=30487;

UPDATE `creature_model_info` SET `bounding_radius`=0.305556, `combat_reach`=1 WHERE `modelid`=23647;

UPDATE `creature_template` SET `AIName`='SmartAI' WHERE `entry` IN (29428,30477,30487);
DELETE FROM `smart_scripts` WHERE `entryorguid` IN (29428,30477,30487) AND `source_type`=0;
DELETE FROM `smart_scripts` WHERE `entryorguid` IN (30477*100,30487*100) AND `source_type`=9;
INSERT INTO `smart_scripts` (`entryorguid`,`source_type`,`id`,`link`,`event_type`,`event_phase_mask`,`event_chance`,`event_flags`,`event_param1`,`event_param2`,`event_param3`,`event_param4`,`action_type`,`action_param1`,`action_param2`,`action_param3`,`action_param4`,`action_param5`,`action_param6`,`target_type`,`target_param1`,`target_param2`,`target_param3`,`target_x`,`target_y`,`target_z`,`target_o`,`comment`) VALUES
(29428,0,0,2,62,0,100,0,9917,0,0,0,11,58061,0,0,0,0,0,7,0,0,0,0,0,0,0,'Ricket - On gossip select - Spellcast Forcecast Summon Rocket (Alliance)'),
(29428,0,1,2,62,0,100,0,9917,1,0,0,11,58062,0,0,0,0,0,7,0,0,0,0,0,0,0,'Ricket - On gossip select - Spellcast Forcecast Summon Rocket (Horde)'),
(29428,0,2,0,61,0,100,0,0,0,0,0,72,0,0,0,0,0,0,7,0,0,0,0,0,0,0,'Ricket - On gossip select - Close gossip'),

[...]

Issue #8195, opened 2012-10-27
[Core/Vehicles] Not working vehicles


Pitcrawler:
There are some vehicles that cant be mounted. After some research I am quite convinced this is due to their seat flags. All of them that I checked were missing VEHICLE_SEAT_FLAG_CAN_ENTER_OR_EXIT. That were vehicleIds 240, 127 and 90 for example.

From a logical point of view it wouldnt make much sense to have a seat that cant be mounted...
In DBCEnums.h that flag has comment Lua_CanExitVehicle which makes much more sense to me. Maybe core handling must be changed and this flag should make the core decide wether to display the exit button when mounted or not.

Something more I noticed is that those vehicles are all flying and taking you on some kind of tour. So it'd make even more sense to be not able to exit voluntairily... So you wouldnt be able to fall to death when accidentally clicking the exit button.

Hash: ca276292dcce939d65649d434a08796269789901

dr-j:
https://github.com/TrinityCore/TrinityCore/issues/13298

StormBytePP:
Fixed by rev ac2845ee77d0f3e838267ad424fbcf4894040333

Issue #1290, opened 2011-05-16
[3.3.5][6.x] Core/Quest: Riding the Red Rocket


click:
The quest does not work

Alliace: http://old.wowhead.com/quest=12437
Horde: http://old.wowhead.com/quest=12432

Alliance Video: http://www.youtube.com/watch?v=lx4RDBHBt8U
Horde Video: http://www.youtube.com/watch?v=_yWB-ZLxh_Q

alein:
Any progress with this issue?

TrinityCore rev. 517ad446e973 2015-02-09 08:32:35 +0100 (3.3.5 branch) (Unix, Release)
Database: TDB_full_335.57_2014_10_19 + updates to the core above.

Regards.

Aokromes:
Generic bug #8195

sirikfoll:
Could reopen ?
Even when the vehicle's generic issue is resolved , it will be needed a script for the quest

StormBytePP:
Fixed by rev ac2845ee77d0f3e838267ad424fbcf4894040333

Killyana:
A little script still required for this quest.

sirikfoll:
Possible related spells:
http://wotlk.openwow.com/spell=49290
http://wotlk.openwow.com/spell=49257
http://wotlk.openwow.com/spell=49510
http://wotlk.openwow.com/spell=49372
http://wotlk.openwow.com/spell=49340

Asterc:
anyone has more information about this quest? as all the spells envolved or sniffs

Rushor:
```c++
/*######
## Quests: 12432, 12437 "Riding the Red Rocket"
## npc_rocket_propelled_warhead
######*/

enum QuestRidingTheRedRocket
{
    SPELL_RIDE_VEHICLE_HARDCODED = 46598,
    SPELL_PARACHUTE              = 44795,
    SPELL_KNOCKBACK_35              = 24199,
    SPELL_PERPETUAL_INSTABILITY  = 49163,
    NPC_ALLIANCE_LUMBERBOAT      = 27688,
    NPC_HORDE_LUMBERBOAT         = 27702,
    ITEM_ELEMENT_115             = 37664
};

class npc_rocket_propelled_warhead : public CreatureScript
{
public:
    npc_rocket_propelled_warhead() : CreatureScript("npc_rocket_propelled_warhead") { }

    CreatureAI* GetAI(Creature* creature) const
    {
[...]

PR #15376, opened 2015-08-28
Core/Vehicles: Renamed VEHICLE_SEAT_FLAG_UNK3 to VEHICLE_SEAT_FLAG_HAS_EXTRA_SEAT

StormBytePP:
This, enables a few vehicles (like the one for quest 12973), which has (unlike near all the others), 3 seats instead of 1 or 2.

EXPERIMENTAL, Should be tested along with info contained in issue #9889 

Help in testing / feedback needed

Based on work by @SeTM

(For 3.3.5 see PR #15377)


Shauren:
I don't get this, at all.
Vehicle seats are defined using seat fields in Vehicle.dbc, the one mentioned in #9889 has 4 seats in dbc - what does a flag on seat have to do with amount of usable seats?

sirikfoll:
right way of not, this also fixes the generic bug in #8195, and consequently #13298, #1290 and prob others.
I say it's worth a closer look '-'

StormBytePP:
@sirikfoll I am doing a much cleaner and proper fix, stay tunned please, I will post in issue #9889 as I'd need some testing

PR #15382, opened 2015-08-29
3.3.5 Fix BuildPointPath and check for MOVEMENTFLAG_DISABLE_GRAVITY

jameyboor:
BuildPointPath uses the 2 arguments incorrectly if _useStraigtLine is true, this means that the 2 3D vectors that are being constructed have the wrong coordinates.

ReadMovementInfo doesn't check  for MOVEMENTFLAG_DISABLE_GRAVITY which is set by lots of client cheating programs to allow flying.

### DIFF
--- a/src/server/game/Movement/PathGenerator.cpp
+++ b/src/server/game/Movement/PathGenerator.cpp
@@ -482,8 +482,9 @@ void PathGenerator::BuildPointPath(const float *startPoint, const float *endPoin
         memcpy(&pathPoints[VERTEX_SIZE * 0], startPoint, sizeof(float)* 3); // first point
 
         // path has to be split into polygons with dist SMOOTH_PATH_STEP_SIZE between them
-        G3D::Vector3 startVec = G3D::Vector3(startPoint[0], startPoint[1], startPoint[2]);
-        G3D::Vector3 endVec = G3D::Vector3(endPoint[0], endPoint[1], endPoint[2]);
+		//startPoint and endPoint are in format y,z,x while G3D::Vector3 expects x,y,z
+        G3D::Vector3 startVec = G3D::Vector3(startPoint[2], startPoint[0], startPoint[1]);
+        G3D::Vector3 endVec = G3D::Vector3(endPoint[2], endPoint[0], endPoint[1]);
         G3D::Vector3 diffVec = (endVec - startVec);
         G3D::Vector3 prevVec = startVec;
         float len = diffVec.length();
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -886,7 +886,7 @@ void WorldSession::ReadMovementInfo(WorldPacket &data, MovementInfo* mi)
         e.g. aerial combat.
     */
 
-    REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY) && GetSecurity() == SEC_PLAYER &&
+    REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY | MOVEMENTFLAG_DISABLE_GRAVITY) && GetSecurity() == SEC_PLAYER &&
         !GetPlayer()->m_mover->HasAuraType(SPELL_AURA_FLY) &&
         !GetPlayer()->m_mover->HasAuraType(SPELL_AURA_MOD_INCREASE_MOUNTED_FLIGHT_SPEED),
         MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY);
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -889,14 +889,11 @@ void WorldSession::ReadMovementInfo(WorldPacket &data, MovementInfo* mi)
     REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY | MOVEMENTFLAG_DISABLE_GRAVITY) && GetSecurity() == SEC_PLAYER &&
         !GetPlayer()->m_mover->HasAuraType(SPELL_AURA_FLY) &&
         !GetPlayer()->m_mover->HasAuraType(SPELL_AURA_MOD_INCREASE_MOUNTED_FLIGHT_SPEED),
-        MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY);
+        MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY | MOVEMENTFLAG_DISABLE_GRAVITY);
 
     //! Cannot fly and fall at the same time
     REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_CAN_FLY) && mi->HasMovementFlag(MOVEMENTFLAG_FALLING),
         MOVEMENTFLAG_FALLING);
-
-	//! Cannot have gravity disabled as player
-	REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_DISABLE_GRAVITY), MOVEMENTFLAG_DISABLE_GRAVITY);
 
     REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_SPLINE_ENABLED) &&
         (!GetPlayer()->movespline->Initialized() || GetPlayer()->movespline->Finalized()), MOVEMENTFLAG_SPLINE_ENABLED);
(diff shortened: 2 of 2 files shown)


src/server/game/Movement/PathGenerator.cpp
```diff
@@ -482,8 +482,9 @@ void PathGenerator::BuildPointPath(const float *startPoint, const float *endPoin
         memcpy(&pathPoints[VERTEX_SIZE * 0], startPoint, sizeof(float)* 3); // first point
 
         // path has to be split into polygons with dist SMOOTH_PATH_STEP_SIZE between them
-        G3D::Vector3 startVec = G3D::Vector3(startPoint[0], startPoint[1], startPoint[2]);
-        G3D::Vector3 endVec = G3D::Vector3(endPoint[0], endPoint[1], endPoint[2]);
+		//startPoint and endPoint are in format y,z,x while G3D::Vector3 expects x,y,z
```
Kittnz: tabs, rest looks good =)

joschiwald:
MOVEMENTFLAG_DISABLE_GRAVITY isn't a valid player movementflag, you can always remove it if player is not a gamemaster

src/server/game/Server/WorldSession.cpp
```diff
@@ -886,13 +886,13 @@ void WorldSession::ReadMovementInfo(WorldPacket &data, MovementInfo* mi)
...
-    REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY) && GetSecurity() == SEC_PLAYER &&
+    REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY | MOVEMENTFLAG_DISABLE_GRAVITY) && GetSecurity() == SEC_PLAYER &&
         !GetPlayer()->m_mover->HasAuraType(SPELL_AURA_FLY) &&
         !GetPlayer()->m_mover->HasAuraType(SPELL_AURA_MOD_INCREASE_MOUNTED_FLIGHT_SPEED),
-        MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY);
+        MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY | MOVEMENTFLAG_DISABLE_GRAVITY);
 
     //! Cannot fly and fall at the same time
-    REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_CAN_FLY | MOVEMENTFLAG_DISABLE_GRAVITY) && mi->HasMovementFlag(MOVEMENTFLAG_FALLING),
+    REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_CAN_FLY) && mi->HasMovementFlag(MOVEMENTFLAG_FALLING),
```
ccrs: sure about this?

jameyboor: Otherwise I wouldn't have PR'd it.

src/server/game/Movement/PathGenerator.cpp
joschiwald: can you explain this change?

jameyboor: BuildPointPath is called from BuildPolyPath which passes on 2 float pointers, namely : 
` float startPoint[VERTEX_SIZE] = {startPos.y, startPos.z, startPos.x};`
` float endPoint[VERTEX_SIZE] = {endPos.y, endPos.z, endPos.x};`

As you can see, they're not in the usual x, y, z format while they're being used that way when constructing 2 3D vectors in BuildPointPath, which is wrong.

Shauren: Are you sure that yzx is actually wrong here? :P

jameyboor: I am, you can check it for yourself, the 2 points are in y z and x and G3D Vector3 expects x y z in it's constructor :P

Shauren: Then you are wrong. While confusing, recast uses a different coordinate system than wow and using yzx here is correct. Please do not make changes basing only on parameter names again.

jameyboor: I know it uses yzx but that's not the case, the 2 vectors being created _DO NOT_ get used in any recast function, which is why it's wrong, it straight up copies them to _pathPoints which DOES expects x y z while the format is still yzx, which is invalid : 
`while (len > SMOOTH_PATH_STEP_SIZE)
        {
            len -= SMOOTH_PATH_STEP_SIZE;
            prevVec += diffVec;
            pathPoints[VERTEX_SIZE * pointCount + 0] = prevVec.x;
            pathPoints[VERTEX_SIZE * pointCount + 1] = prevVec.y;
            pathPoints[VERTEX_SIZE * pointCount + 2] = prevVec.z;
            ++pointCount;
        }`

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -482,8 +482,9 @@ void PathGenerator::BuildPointPath(const float *startPoint, const float *endPoin
...
+        G3D::Vector3 startVec = G3D::Vector3(startPoint[2], startPoint[0], startPoint[1]);
+        G3D::Vector3 endVec = G3D::Vector3(endPoint[2], endPoint[0], endPoint[1]);
```
boxa: good catch.

but need also change below to:
```
            pathPoints[VERTEX_SIZE * pointCount + 0] = prevVec.y; // Y
            pathPoints[VERTEX_SIZE * pointCount + 1] = prevVec.z; // Z
            pathPoints[VERTEX_SIZE * pointCount + 2] = prevVec.x; // X
```

src/server/game/Movement/PathGenerator.cpp
jackpoz: isn't https://github.com/jameyboor/TrinityCore/blob/3.3.5/src/server/game/Movement/PathGenerator.cpp#L550 already fixing the coordinate order in the G3D vector ?

jackpoz: any feedback to my comment above ?

jackpoz:
summing up this PR: 
- the xyz change is not going to be merged because of missing feedback to the objections
- is the movement flag part correct ? any how to reproduce steps ?

Kittnz:
@jameyboor what's the best way to test this PR?

jameyboor:
The xyz change was resolved with Shauren. The movement flag part wasn't though however, a legit client should never have that flag in the first place, it is added by some hack tools for fly hacking.

Kittnz:
I guess if it doesn't harm the normal client without hacking tools i don't see any harm.

jackpoz:
> The xyz change was resolved with Shauren.

https://github.com/TrinityCore/TrinityCore/pull/15382#discussion_r40465437 is still waiting for a reply

> it is added by some hack tools for fly hacking.

Without a link to the hack tools it's impossible to test the PR and verify that it actually doesn't allow anymore to flyhack

Shauren:
The xyz change was indeed resolved and should have been completely removed from this PR
Pasting relevant conversation from irc
```
[18:43] <Jameyboor> Shauren, you must've misunderstood me
[18:45] <%Shauren> Jameyboor: no, you misread the code (i read your reply)
[18:45] <%Shauren> it copies the points, thats true
[18:45] <%Shauren> but the destination is NOT '_pathPoints' but 'pathPoints'
[18:46] <@Aokromes> Clarkey if you only have those 2 files you have incomplete client
[18:46] <Jameyboor> I noticed just now
[18:46] <Jameyboor> I stand corrected
[18:46] <Jameyboor> Still, if you have a Vector3 with x y z params
[18:46] <Jameyboor> and you're not using it that way
[18:47] <Jameyboor> It's just vague
[18:47] <Jameyboor> Improvements can be made
[18:47] <%Shauren> true and not true at the same time
[18:47] * fallenoak (~textual@Rizon-79DA99C0.hsd1.mn.comcast.net) has joined #trinity
[18:47] <%Shauren> Vector3 is used here for the convenience methods (operator- and length())
[18:47] * Clarkey (~qwebirc@Rizon-B83AC057.as13285.net) Quit (Ping timeout: 240 seconds)
[18:47] <%Shauren> so it doesnt reimplement the code
[18:48] <Jameyboor> Yeah alright
[18:48] <Jameyboor> Still, that doesn't take away the needed MOVEMENTFLAG_DISABLE_GRAVITY check :P
```

src/server/game/Server/WorldSession.cpp
jackpoz: >send a modified packet to contain both disable_gravity and falling while having SPELL_AURA_MOD_INCREASE_MOUNTED_FLIGHT_SPEED (a flying mount).
you will freeze everyone nearby

this change must be reverted

Demonid:
@jackpoz

jackpoz:
Tested that tool, I can still flyhack with this pr.

Summing up again, half of this PR breaks pathfinding, a quarter freezes all the clients around and the remaining quarter doesn't fix anything (or the things supposed to be fixed are still unknown since there are no how to reproduce steps).

Reopen this PR once all changes will be tested ingame and once proper clear detailed How To Reproduce steps will be provided

PR #15392, opened 2015-08-31
Core/Spells: Fix tauren and night elf moonkin form

thomas33:
Fix tauren and night elf moonkin form


DDuarte:
Converted in a switch for clarity

Issue #14968, opened 2015-06-28
[Spell] Using a spell with a taunt effect on a player


chaodhib:
When using a spell with a taunt effect (example 6795, druid's Growl)  on a player:
a) It should say "Immune" above the head the of player.
b) The player shouldn't have the debuff "Taunted" applied to him.

![screenshot](http://i.imgur.com/Df3kRAe.jpg)

I tried to look at the code and modifying AuraEffect::HandleModTaunt() seems to be ineffective. The debuff is already applied when this method is reached.

An AuraScript could be used but, if i'm not mistaken, this imply to connect this script to every spell that has a Taunt effect. Isnt there any other solution?

edit: example on retail: https://www.youtube.com/watch?v=BYK6OkFwbv8 at 3:27. The dk death grip a rogue. a) immune above his head. b) no debuff with the Death Grip icon.

CORE revision number: https://github.com/TrinityCore/TrinityCore/commit/accab5451d3e62739e4aa376d25bc6623f0d3c52 (branch 3.3.5a, 26/06/15)
TDB 335.58 +updates up until 2015_06_26_02_world.

mik1893:
source for your issue? I mean, any video / proof from retail?

Goatform:
If I remember good, players are immune on taunt effects.

Keader:
Players take taunt seems wrong to me

Goatform:
Of course they will auto-target you because you casted spell on player and he doesn't have a target, try to reproduce that when player have a target and then use taunt, if player switch targets then it's high issue, but mostly like it's just visual as I said.

chaodhib:
a) is mostly cosmestic yes.
However b) is important imo. A useless debuff like that is bad.

emptyus:
Researching this issue can maybe reveal some other core problems. Like that one that screws up phases on LK encounter using taunt. I doubt its purely cosmetic issues or visual...

Goatform:
That isn't same issue. PvE and PvP system are different as you know.

chaodhib:
@Goatform this isnt about the player changing his target. It's about immunity to the spell effect and to the debuff.

Goatform:
Yeah I understood you very well.

Mack36:
The same thing is with warrior's Mocking Blow (spellId 694). It shouldn't deal damage to a player (due to description).

Goatform:
Mocking Blow - " If the target is tauntable, also deals weapon damage." 

[...]

PR #15354, opened 2015-08-25
[WIP] Spells/Death Grip fix. The pull effect shouldn't work on some bosses

chaodhib:
Fixes https://github.com/TrinityCore/TrinityCore/issues/14870
- An extra flag CREATURE_FLAG_EXTRA_NO_PULL_BY_GRIP has been created to handle the immunity to the pull component of the spell Death Grip
- a spell script has been written to the third spell called by the Death Grip chain in order to send to the client the message "immune" if the creature has the flag mentioned above.
- the 3 spell scripts are grouped together in the spell_dk for convenience.

Patchwerk (16028) is taken as the dummy:
wowscrnshot_082615_012611
This screenshot is taken after the fix is applied. It show that indeed the taunt is applied but not the pull.

For some reason, the dk didnt got the same threat as the highest in list and only got aggro for the 3s duration of the taunt. Unsure if this is linked to the PR.

To do:
- [X] implement the extra flag
- [X] Make use of it to implement the immunity
- [x] Message "immune" should be shown to the caster if the mob is immune to the pull effect.
- [ ] Investigate the thing about the bugged threat. See if it's linked to this PR.
- [ ] Update the flag of each boss. And then, update the known exceptions. (SQL queries)

PR #15402, opened 2015-09-01
Core/Spells: Make players immune to taunt

chaodhib:
Fixes https://github.com/TrinityCore/TrinityCore/issues/14968.

It makes players immune to both the spell effect ATTACK ME and the taunt aura. Also, the "immune" message correctly shows up when a taunt ability (warrior, druid, ..) is used against a player. An exception to this being Death Grip (see https://github.com/TrinityCore/TrinityCore/pull/15354#issuecomment-138095254). That should be the concern of another PR.

edit: update to the PR current situation.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -2510,6 +2510,17 @@ void Player::RemoveFromWorld()
     }
 }
 
+bool Player::IsImmunedToSpellEffect(SpellInfo const* spellInfo, uint32 index) const
+{
+    // players are immune to taunt (the aura and the spell effect).
+    if (spellInfo->Effects[index].IsAura(SPELL_AURA_MOD_TAUNT))
+        return true;
+    if (spellInfo->Effects[index].IsEffect(SPELL_EFFECT_ATTACK_ME))
+        return true;
+    
+    return Unit::IsImmunedToSpellEffect(spellInfo, index);
+}
+
 void Player::RegenerateAll()
 {
     //if (m_regenTimer <= 500)
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -1118,6 +1118,8 @@ class Player : public Unit, public GridObject<Player>
 
         static bool BuildEnumData(PreparedQueryResult result, WorldPacket* data);
 
+        bool IsImmunedToSpellEffect(SpellInfo const* spellInfo, uint32 index) const override; // override Unit::IsImmunedToSpellEffect
+
         void SetInWater(bool apply);
 
         bool IsInWater() const override { return m_isInWater; }


Trisjdc:
Players also need to be immune to SPELL_EFFECT_ATTACK_ME. And I would suggest using a IsImmuneToSpellEffect override for Player, because shitty code can override ApplySpellImmune during a player object's lifetime.

chaodhib:
The spell effect handler of SPELL_EFFECT_ATTACK_ME check himself if the target can have an agro list and does nothing if that's the case.

I've used ApplySpellImmune() precisely because I would like to keep the generic nature of IsImmuneToSpellEffect(), if that's possible. A call to ApplySpellImmune(..., false) during the object's lifetime worries me too. Unsure what to do about that.

Trisjdc:
Neither of the effects does jack to players, but if you want it to properly display immune then the player has to be immune to both. IsImmuneToSpellEffect is already overriden for creatures.. and it'll remain 'generic', you'll still call the Unit version.

chaodhib:
To be blizz-like, the core will need to support spell effect hit results (see https://github.com/TrinityCore/TrinityCore/issues/14968#issuecomment-118374137). Meaning if a spell with 2 spell effects is casted on some unit and the unit is immune to one of the 2 spell effects, one "immune" should show up. The target doesn't have to be immune to the whole spell or to every spell effects for the "immune" message to show up. See the retail video proof in the first post.

And this functionality should be added on another PR since it has nothing to do specific with the current issue (and it will required quite a bit of changes I think).

I agree on everything else.

chaodhib:
PS: I dont think players should be immune SPELL_EFFECT_ATTACK_ME. It should just not affect them, which is the current behavior. Otherwise, once the "spell effect hit results" thing is implemented, we'll get 2 immunes messages when taunting a player, which is in contradiction with the proof see above.

chaodhib:
https://github.com/TrinityCore/TrinityCore/issues/14968#issuecomment-137302475

chaodhib:
There you go. Players are now immune to both the aura and the spell effect ATTACK_ME. The result is that now, the "immune" message shows up when a taunt ability (warrior, druid, ..) is used against a player. 

The only problem being Death Grip (always, that bloody ability). The "immunity" doesnt show up because the target isnt immune to the first spell effect (SPELL_EFFECT_DUMMY) and if IsImmuneToSpell() is tweaked to ignore SPELL_EFFECT_DUMMY, the spell script is executed (and the spell handling the pull, 49575, is not launched).

In any case, this should be fixed in another PR specific to that (bloody) ability. This PR is ready imo.

MrSmite:
Wouldn't it be better to add the check in *Unit::IsImmunedToSpellEffect()*?

chaodhib:
Because it's cleaner that way imo. Specific parts are in a specific function.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -2510,6 +2510,18 @@ void Player::RemoveFromWorld()
...
 }
 
+bool Player::IsImmunedToSpellEffect(SpellInfo const* spellInfo, uint32 index) const
+{
+    // players are immune to taunt (the aura and the spell effect).
+    if (uint32 aura = spellInfo->Effects[index].ApplyAuraName)
+        if (aura == SPELL_AURA_MOD_TAUNT)
+            return true;
+    if (spellInfo->Effects[index].Effect == SPELL_EFFECT_ATTACK_ME)
+        return true;
```
joschiwald: you could use spellinfo helpers `SpellEffectInfo::IsEffect` and `SpellEffectInfo::IsAura`

src/server/game/Entities/Player/Player.h
```diff
@@ -1118,6 +1118,8 @@ class Player : public Unit, public GridObject<Player>
 
         static bool BuildEnumData(PreparedQueryResult result, WorldPacket* data);
 
+        bool IsImmunedToSpellEffect(SpellInfo const* spellInfo, uint32 index) const override; // override Unit::IsImmunedToSpellEffect
```
joschiwald: imho, useless comment since override is a c++ keyword

src/server/game/Entities/Player/Player.cpp
chaodhib:
```C++
if (spellInfo->Effects[index].IsAura())
    if (spellInfo->Effects[index].ApplyAuraName == SPELL_AURA_MOD_TAUNT)
        return true;
if (spellInfo->Effects[index].IsEffect())
    if (spellInfo->Effects[index].Effect == SPELL_EFFECT_ATTACK_ME)
        return true;
```
Like this you mean?

src/server/game/Entities/Player/Player.h
chaodhib: it's with Creature::IsImmunedToSpellEffect(). I figured why not.

src/server/game/Entities/Player/Player.cpp
joschiwald: those helpers accept the name/effect as parameter

chaodhib: oh ok

src/server/game/Entities/Player/Player.h
chaodhib: and the Unit class has parent classes too. The comment explicitly states the overwritten method, which isn't a bad thing to have I suppose.

MrSmite:
> Because it's cleaner that way imo. Specific parts are in a specific function.

Perhaps but since Player inherits from Unit and the override isn't providing any unique functionality or require a different parameter list, it seems unnecessary, IMO.

chaodhib:
Overriding a method isn't about providing new functionalities. It's about changing the implementation of a method in order to make it more specific to the child class. Which is done here.

Problem with putting the code in Unit::IsImmuneToSpellEffect() is that those 2 checks will be done every time IsImmuneToSpellEffect() is called on a sub class of Unit, including every Creatures. Here we are using the polymorphism capability of the language to automatically avoid that.

MrSmite:
> Overriding a method isn't about providing new functionalities. It's about changing the implementation of a method in order to make it more specific to the child class. Which is done here.

Well, changing implementation is providing new functionalities

> Problem with putting the code in Unit::IsImmuneToSpellEffect() is that those 2 checks will be done every time IsImmuneToSpellEffect() is called on a sub class of Unit, including every Creatures. Here we are using the polymorphism capability of the language to automatically avoid that.

Not necessarily, the checks only need to be performed if __GetTypeId() == TYPEID_PLAYER__. Trinity does checks like this all over the code.

chaodhib:
You are right. It's the same in term of efficiency really. 

However, putting code specific to players in the Player class is easier to maintain and therefore good practice.

MrSmite:
> However, putting code specific to players in the Player class is easier to maintain and therefore good practice.

Yes, that I will agree with.

Anyway, I'm just curious if there are any cases where NPCs can taunt a player? I haven't been on retail since WotLK but I thought there were cases where NPCs could cause a player to target them.

chaodhib:
> I'm just curious if there are any cases where NPCs can taunt a player?
>  I thought there were cases where NPCs could cause a player to target them.

Hm, unsure. As far as I can tell, no. Maybe the shaman's Earth elemental?

Also, it's worth noting that being forced to change his target doesn't equal to being taunted. Being taunted means you are forced to attack (by auto attacks or spells) a target for a predetermined duration.

MrSmite:
> Also, it's worth noting that being forced to change his target doesn't equal to being taunted. Being taunted means you are forced to attack (by auto attacks or spells) a target for a predetermined duration.

Right, I was just curious if there are any spells like that which apply a taunt effect along with some other effect.

ccrs:
I agree @chaodhib, its cleaner to override an inherited method to add specific functionalities.

chaodhib:
@MrSmite Each spell has one or multiple spell effects. In general, they work independently of each other. Meaning if the target is immune to one of the spell effects (for example the taunt), the other spell effects will still applied (for example here the pull).

This is a simplification of how Death Grip actually works but this is to show you the big picture.

chaodhib:
Bump...

jackpoz:
@joschiwald any final feedback about this pr ?

joschiwald:
looks good to me

jackpoz:
thank you for the fix and the long tests

chaodhib:
My pleasure. I just wished the PR process was faster..

Issue #15406, opened 2015-09-02
[Request] Allow gms inside of an instance to log back in without getting teleported out even with .gm off


jackpoz:
Like a real issue, "Title says it all". Send a message to the player notifying him about the behavior, something like "Naios just saved you from getting kicked out of the instance..."
Commit: https://github.com/TrinityCore/TrinityCore/commit/24a624b38c5d436341a4b87eef6fbe66afc4bfb6

jackpoz:
@Naios : there was a time when you had this issue, fixed now

PR #15408, opened 2015-09-02
Core/Player: Always allow accounts with RBAC_PERM_COMMAND_GM to enter…

jackpoz:
Core/Player: Always allow accounts with RBAC_PERM_COMMAND_GM to enter instances on login

Always allow accounts with RBAC_PERM_COMMAND_GM to enter instances on login even if player had .gm off .
Send an ingame message to notify the user about the avoided kick.

The text should be moved to SQL.
Tested with a GM account and a normal user account in Black Temple, the GM received the ingame message without getting kicked and the user got teleported out.

Closes https://github.com/TrinityCore/TrinityCore/issues/15406

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -2561,6 +2561,11 @@ void Player::SetGameMaster(bool on)
     UpdateObjectVisibility();
 }
 
+bool Player::CanBeGameMaster() const
+{
+    return m_session && m_session->HasPermission(rbac::RBAC_PERM_COMMAND_GM);
+}
+
 void Player::SetGMVisible(bool on)
 {
     if (on)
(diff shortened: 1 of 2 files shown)


Shauren:
Not sure if the message is needed.

jackpoz:
It's a way to avoid tickets like "Bug: Players doesn't get kicked out of instance on login if without a party"

Naios:
:+1: :shipit:

src/server/game/Entities/Player/Player.cpp
```diff
@@ -18986,17 +18991,28 @@ bool Player::CheckInstanceLoginValid()
     {
         // cannot be in raid instance without a group
         if (!GetGroup())
-            return false;
+            return IsInstanceLoginGameMasterException();
```
joschiwald: if player has gm command you will bypass all other checks, so i don't think this is not the right way to go

i guess you just need to adjust the gm level check which allows the player to enter in gm mode

jackpoz: > if player has gm command you will bypass all other checks

this is the whole point of the PR as described in the PR description, I don't get the point of your comment

jackpoz: so @joschiwald ? what's the point of your comment ?

PR #15409, opened 2015-09-03
AHBot "rebuild" command update

MrSmite:
Update AHBot "rebuild" command to allow specifying a single house to expire auctions from.

example (neutral only): .ahbot rebuild all n
example (alliance only): .ahbot rebuild all a

### DIFF
--- a/sql/updates/world/2015_09_02_05_world_ahbot_expire_by_house.sql
+++ b/sql/updates/world/2015_09_02_05_world_ahbot_expire_by_house.sql
@@ -1,0 +1,2 @@
+DELETE FROM `command` WHERE `name` = 'ahbot rebuild';
+INSERT INTO `command` (`name`, `permission`, `help`) values('ahbot rebuild','791','Syntax: .ahbot rebuild [all] [[a]lliance [h]orde [n]eutral]\r\n\r\nExpire all actual auction of ahbot except bided by player. Binded auctions included to expire if \"all\" option used. If one of \"a\" \"h\" or \"n\" option is used, only the auctions of that house will expire otherwise all houses will expire. Ahbot re-fill auctions base at current settings then.');
--- a/src/server/scripts/Commands/cs_ahbot.cpp
+++ b/src/server/scripts/Commands/cs_ahbot.cpp
@@ -146,12 +146,35 @@ public:
     static bool HandleAHBotRebuildCommand(ChatHandler* /*handler*/, const char* args)
     {
         char* arg = strtok((char*)args, " ");
+        char* ahMapIdStr = strtok(NULL, " ");
+        uint32 ahMapID = MAX_AUCTION_HOUSE_TYPE;
 
         bool all = false;
         if (arg && strcmp(arg, "all") == 0)
             all = true;
 
-        sAuctionBot->Rebuild(all);
+        if (ahMapIdStr)
+        {
+            switch (ahMapIdStr[0])
+            {
+                case 'n':
+                case 'N':
+                    ahMapID = AUCTION_HOUSE_NEUTRAL;
+                    break;
+                case 'a':
+                case 'A':
+                    ahMapID = AUCTION_HOUSE_ALLIANCE;
+                    break;
+                case 'h':
+                case 'H':
+                    ahMapID = AUCTION_HOUSE_HORDE;
+                    break;
+                default:
+                    break;
+            }
+        }
+        
+        sAuctionBot->Rebuild(all, ahMapID);
         return true;
     }
 
--- a/src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
+++ b/src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
@@ -398,10 +398,13 @@ void AuctionHouseBot::PrepareStatusInfos(AuctionHouseBotStatusInfo& statusInfo)
     }
 }
 
-void AuctionHouseBot::Rebuild(bool all)
+void AuctionHouseBot::Rebuild(bool all, uint32 houseId)
 {
     for (uint32 i = 0; i < MAX_AUCTION_HOUSE_TYPE; ++i)
     {
+        if ((houseId != MAX_AUCTION_HOUSE_TYPE) && (i != houseId))
+            continue; // if specific house ID sent, increment until we get a match
+
         AuctionHouseObject* auctionHouse = sAuctionMgr->GetAuctionsMap(AuctionHouseType(i));
         for (AuctionHouseObject::AuctionEntryMap::const_iterator itr = auctionHouse->GetAuctionsBegin(); itr != auctionHouse->GetAuctionsEnd(); ++itr)
             if (!itr->second->owner)                        // ahbot auction
--- a/src/server/game/AuctionHouseBot/AuctionHouseBot.h
+++ b/src/server/game/AuctionHouseBot/AuctionHouseBot.h
@@ -283,7 +283,7 @@ public:
     void SetItemsAmount(uint32(&vals)[MAX_AUCTION_QUALITY]);
     void SetItemsAmountForQuality(AuctionQuality quality, uint32 val);
     void ReloadAllConfig();
-    void Rebuild(bool all);
+    void Rebuild(bool all, uint32 houseId);
 
     void PrepareStatusInfos(AuctionHouseBotStatusInfo& statusInfo);
 private:


MrSmite:
Hmm, so Travis Build doesn't like *_stricmp*?

src/server/scripts/Commands/cs_ahbot.cpp
```diff
@@ -146,12 +146,28 @@ class ahbot_commandscript : public CommandScript
...
+        uint32 ahMapID = MAX_AUCTION_HOUSE_TYPE;
 
         bool all = false;
         if (arg && strcmp(arg, "all") == 0)
             all = true;
 
-        sAuctionBot->Rebuild(all);
+        if (ahMapIdStr)
+        {
+            // NOTE: Should use _stricmp() for case insenstive, stricmp() is deprecated
```
Shauren: You can just remove this comment, pretty much ALL posix functions with _ prefix are msvc specific, we ignore that warning for a reason.

Also - linux case: https://github.com/TrinityCore/TrinityCore/blob/6.x/src/common/Common.h#L79

MrSmite: Done. Thanks for the explanation.

jackpoz: aren't all these if/else just a switch on ahMapIdStr[0] ? No need for any stricmp

MrSmite: Yes, good point.

sql/updates/world/2015_09_02_05_world_ahbot_expire_by_house.sql
```diff
@@ -0,0 +1,2 @@
+DELETE FROM `command` WHERE `name` = 'ahbot rebuild';
+INSERT INTO `command` (`name`, `permission`, `help`) values('ahbot rebuild','791','Syntax: .ahbot rebuild [all] [[a]lliance [h]orde [n]eutral]\r\n\r\nExpire all actual auction of ahbot except bided by player. Binded auctions included to expire if \"all\" option used. If one of \"a\" \"h\" or \"n\" option is used, only the auctions of that house will expire otherwise all houses will expire. Ahbot re-fill auctions base at current settings then.');
```
tkrokli: missing New Line at the end of the file.

MrSmite: Fixed, thanks.

What's the purpose of that anyway?

tkrokli: It is mostly about the way the file is stored, what kind of (invisible) character the file ends with.

MrSmite: Right, I just always assumed that Notepad++ would add an EOF marker to the end of the file which would be sufficient. I often forget to add a new line.

tkrokli: The way I remember to add a new line is by thinking that I don't want the code to be highlighted while reading the code, making a new line to "rest" the highlighted line after the code. :)

tkrokli:
Try squashing your commits to see if Travis will show a new and updated status of your PR.

MrSmite:
> Try squashing your commits to see if Travis will show a new and updated status of your PR.

I first need to fix a bunch of typos in the switch() statement. For starters I used double quotes instead of single...

tkrokli:
Good point. I didn't catch that when looking at the build errors in Travis.

MrSmite:
Question:

Since I'm performing a switch() on a char*, which is preferred by Trinity:

*switch(*ahMapIdStr)* __or__ *switch(ahMapIdStr[0])*

they both compile properly...

jackpoz:
ahMapIdStr[0] means "first element of the array", *ahMapIdStr means "dereference the pointer" .
They both give the same result but ahMapIdStr[0] is more clear so please use that one.

MrSmite:
> They both give the same result but ahMapIdStr[0] is more clear so please use that one.

Done

sql/updates/world/2015_09_02_05_world_ahbot_expire_by_house.sql
Shauren: Its for people who use 'cat' util to merge updates

Rochet2:
> Hmm, so Travis Build doesn't like _stricmp?

When I used that in a small test program, antivirus claimed the program was a troijan.

Maybe should make it so that the command returns false if something is provided in the command, but the provided input is not "all" or any of the a,h,n special characters. Just to make it so that invalid input doesnt make stuff happen and that ppl not knowing of change wont get stumped by different things happening from before without notice.

jackpoz:
it looks kinda complete to me, someone just has to test it ingame (and maybe debug it too while at it)

MrSmite:
@Rochet2 

> Just to make it so that invalid input doesnt make stuff happen and that ppl not knowing

I get what you're saying but invalid input is simply ignored. For example, *bool all = false* remains that way if the first argument isn't "all" which is in line with the help definition. Same with the *house* option. If none is supplied then it effects all houses which is also in keeping with the help definition.

So essentially *.ahbot rebuild damn you* is the same as *.ahbot rebuild* and *.ahbot rebuild all sucker* is the same as *.ahbot rebuild all* (which is actually how it worked before I started messing with it).

sql/updates/world/2015_09_02_05_world_ahbot_expire_by_house.sql
MrSmite: > Its for people who use 'cat' util to merge updates

This is one reason why I should leave my Windows world. It's easy to forget that other people do things differently.

jackpoz:
does it allow ".ahbot rebuild neutral" without "all" ? or it's ok not to support that case ?

tkrokli:
Just in case this PR gets merged: you need to set up your git config with your github user e-mail address so your commits can link to your username + avatar properly (current commit has no such link).

MrSmite:
> does it allow ".ahbot rebuild neutral" without "all" ? or it's ok not to support that case ? Just to know the behavior

Yes, when I tested it, omitting the "all" was supported. Someone may want to verify that though.

Also, I've been thinking more about what Rochet said and I realized one possible point of confusion would be if someone did: __*.ahbot rebuild all now*__ it would see the "n" in "now" as a valid parameter because the switch() only considers the whole word.

In this case it would only rebuild the neutral house and the user might have expected every house to be rebuild.

MrSmite:
__UPDATE:__ I'm going to test the command again by omitting the *all* parameter, just waiting for the bot to fill my auctionhouse.

MrSmite:
Ok so omitting the *all* comand just ignores the house parameter and wipes any auction with no bid. Essentially, *.ahbot rebuild h* is the same as *.ahbot rebuild*

So I need to parse the parameter list differently...

Shauren:
Just a reminder: in no command we do some magic like optional middle parameter - you don't have to think about this case.

Rochet2:
I would still like `.ahbot rebuild h` to error (since h != all and something was provided in place of all) instead of having some complex check or simply ignoring the argument.
That would  be easy to do. Just make it return false if `strcmp(arg, "all") == 0` is false.

MrSmite:
> I would still like .ahbot rebuild h to error (since h != all and something was provided in place of all)

Except that the original commands are:

> .ahbot rebuild
> .ahbot rebuild all

Then I added [a,h,n]:

> .ahbot rebuild all [a,h,n]
> .ahbot rebuild [a,h,n]

So to error with no "all" command would be incorrect.

MrSmite:
New code tested as follows:

*rebuild g j* -> correctly failed
*rebuild u* -> correctly failed
*rebuild all p* -> correctly failed
*rebuild h* -> correctly passed
*rebuild all* -> correctly passed

I need to wait for the bot to refill the auctionhouse now...

sql/updates/world/2015_09_02_05_world_ahbot_expire_by_house.sql
tkrokli: I actually missed out on some minor flaws in your SQL when I commented on the missing Line Feed. I also did some spell checking and corrected the grammar of some words. I was uncertain about the exact meaning in the following:
`Expire all actual auction of ahbot except bided by player.` -- bid, bid, bid - just like 'cast', right?
`Binded auctions included to expire if \"all\" option used.`. -- Bind, bound, bound -or- bid, bid, bid?

I have tried correcting them to the following, though I may have missed on the intended meaning:
`Expire all actual auctions of ahbot, except bids by player.`
`Auction bids included to expire if \"all\" option used.` - though I am uncertain if that is what you mean.

Sample with additional grammar corrections:
```sql
DELETE FROM `command` WHERE `name` = 'ahbot rebuild';
INSERT INTO `command` (`name`, `permission`, `help`) VALUES
('ahbot rebuild',791,'Syntax: .ahbot rebuild [all] [[a]lliance [h]orde [n]eutral]\r\n\r\nExpire all actual auctions of ahbot, except bids by player. Auction bids included to expire if \"all\" option used. If one of \"a\" \"h\" or \"n\" option is used, only the auctions of that house will expire. Otherwise, all houses will expire. Ahbot will then refill auctions based on current settings.');
```
Check to see if the intended meanings remain. Remember to add a line break if you copy-paste the above sample. ;)

MrSmite:
Question:

I noticed when using *account rebuild all* from the console, I get spammed with:

*AccountMgr::HasPermission: Wrong accountId 0*

I know AHBot uses no account but I'm not sure where that error is generated. The two functions I'm touching don't use the account for anything.

MrSmite:
> I also did some spell checking and corrected the grammar of some words

Just noticed this comment... those errors exist in the current version. I can fix them if you want.

tkrokli:
I wish I had more time to look over the entire source and supply grammatical fixes. I would need a list of priority files, though, to save time looking for "engrish" text and write the PRs for them.

MrSmite:
Preview of the cleaner version of the SQL:

```SQL
DELETE FROM `command` WHERE `name` = 'ahbot rebuild';
INSERT INTO `command` (`name`, `permission`, `help`) VALUES ('ahbot rebuild','791','Syntax: .ahbot rebuild [all] [[a]lliance [h]orde [n]eutral]\r\n\r\nRebuild command with no options expires all AHBot auctions in all houses that have no bid.\r\n\r\nTo include auctions with bids use the \"all\" option.\r\nTo expire auctions from a specific house use one of the [a,h,n] options.\r\nAHBot will re-fill auctions per CONF settings.');
```

Thoughts?

MrSmite:
The more I work on this, the more I don't really like the option of no parameters. Most other commands will give you the help text if you supply no parameters but here the bot will expire all auctions without bids.

I think the first parameter should be mandatory: *all* or *nobid* and leave the A,H,N parameter as optional (eg: all houses if omitted). At least this way a *.ahbot rebuild* can then print the help message instead of surprising the user by wiping the auctionhouse.

Thoughts?

tkrokli:
> Preview of the cleaner version of the SQL:
***
> Thoughts?

1. SQL codestyle:

   a)  Always place the insert values on a new line. Insert a line break after "VALUES".
   b) Never use ' ' for number insert, only for text insert. (,791, instead of ,'791',)
2. Text: consider using more commas and periods to shorten the sentences for easier reading. 

Examples:
`To include auctions with bids, use the \"all\" option.`
`To expire auctions from a specific house, use one of the [a,h,n] options.`

Also consider quoting the commands directly when you refer to them in the text:
The '.ahbot rebuild' command **with no options** will expire all AHBot auctions with no bids, in all houses.

(Yes, I have done a lot of nitpicking here, since you asked for my thoughts on the subject.) :)

Rochet2:
@MrSmite
Hmm, you are right that erroring is not right. However `.ahbot rebuild [a,h,n]` is not valid.

The valid commands are
- .ahbot rebuild  
.ahbot rebuild x (same effect as previous)
- .ahbot rebuild all
- .ahbot rebuild all [a,h,n]
- .ahbot rebuild x [a,h,n]

where x is something else than `all`

I agree using some keyword instead of "nothing" or "something but the keyword" is clearer.
Its also relatively simple to code :)
I feel like writing something like `.ahbot rebuild asd a` would be silly (and the only way ..) to achieve the last command.

sql/updates/world/2015_09_02_05_world_ahbot_expire_by_house.sql
Aokromes: And because possible missing newlines is why i start my commits with -- to avoid broken "set" etc :)

MrSmite:
FYI: I haven't forgotten about this, it's been a busy week (and a half?). I hope to have something new for you guys within the next few days as things settle down here.

MrSmite:
Update:

Got sidetracked with actually playing the game for a change ;) and my recent AHBot fix. Haven't forgotten about this though.

MrSmite:
Sorry about the delay here, I've had some IRL things keeping me busy.

Also, I need to update my core since I think there was a fix to combine faction houses recently?

Shauren:
No. A, H, N split is still there (what was "combined" on 3.3.5 was only AH mail senders)

MrSmite:
> No. A, H, N split is still there (what was "combined" on 3.3.5 was only AH mail senders)

Ah, ok. I guess I misread the reply to that thread.

Carbenium:
@MrSmite: is it done from your point of view?

jackpoz:

jackpoz:
any update ?

jackpoz:
Please ask a TrinityCore member to reopen this Pull Request once you provide the requested feedback. There is no need to open a new Pull Request with the same changes.

PR #15422, opened 2015-09-03
ipv6 support for bnetserver and worldserver

TonyHoyle:
This is ipv6 support.  It works on my system and should be OK on non-ipv6 systems - I've written it such that it avoids touching ipv6 code if it's not configured to do so (to avoid nasties like partially configured routing and such).

To enable ipv6 on the worldserver, just set address6 in auth.realmlist (and of course enable it on the client).

There's also support for ipv6 the bnetserver although the current blizzzard client always connects over ipv4 for this even when a good ipv6 connection is available.. I anticipate they'll add it soon anyway and it's not many extra lines.

To enable it in the bnetserver change the BindIP to '::' or '::1' depending on whether you're using a global or loopback address.

Todo: 
- Test on machines with no ipv6/no v6 routing.  I've tried to code defensively but there's no substitute for a proper test
- Test on Windows... unfortunately I have no easy way to compile on that platform.

### DIFF
--- a/sql/base/auth_database.sql
+++ b/sql/base/auth_database.sql
@@ -583,6 +583,7 @@ CREATE TABLE `realmlist` (
   `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `name` varchar(32) NOT NULL DEFAULT '',
   `address` varchar(255) NOT NULL DEFAULT '127.0.0.1',
+  `address6` varchar(255) NOT NULL DEFAULT '',
   `localAddress` varchar(255) NOT NULL DEFAULT '127.0.0.1',
   `localSubnetMask` varchar(255) NOT NULL DEFAULT '255.255.255.0',
   `port` smallint(5) unsigned NOT NULL DEFAULT '8085',
@@ -605,7 +606,7 @@ CREATE TABLE `realmlist` (
 
 LOCK TABLES `realmlist` WRITE;
 /*!40000 ALTER TABLE `realmlist` DISABLE KEYS */;
-INSERT INTO `realmlist` VALUES (1,'Trinity','127.0.0.1','127.0.0.1','255.255.255.0',8085,0,2,1,0,0,20444,2,1);
+INSERT INTO `realmlist` VALUES (1,'Trinity','127.0.0.1','','127.0.0.1','255.255.255.0',8085,0,2,1,0,0,20444,2,1);
 /*!40000 ALTER TABLE `realmlist` ENABLE KEYS */;
 UNLOCK TABLES;
 
--- a/sql/updates/auth/2015_09_02_00_auth.sql
+++ b/sql/updates/auth/2015_09_02_00_auth.sql
@@ -1,0 +1,15 @@
+SET @s = (SELECT IF(
+    (SELECT COUNT(*)
+        FROM INFORMATION_SCHEMA.COLUMNS
+        WHERE table_name = 'realmlist'
+        AND table_schema = DATABASE()
+        AND column_name = 'address6'
+    ) > 0,
+    "SELECT 1",
+    "ALTER TABLE realmlist ADD address6 VARCHAR(255) NOT NULL DEFAULT ''"
+));
+
+PREPARE stmt FROM @s;
+EXECUTE stmt;
+DEALLOCATE PREPARE stmt;
+
--- a/sql/updates/auth/2015_09_02_00_auth.sql
+++ b/sql/updates/auth/2015_09_02_00_auth.sql
@@ -1,15 +1,1 @@
-SET @s = (SELECT IF(
-    (SELECT COUNT(*)
-        FROM INFORMATION_SCHEMA.COLUMNS
-        WHERE table_name = 'realmlist'
-        AND table_schema = DATABASE()
-        AND column_name = 'address6'
-    ) > 0,
-    "SELECT 1",
-    "ALTER TABLE realmlist ADD address6 VARCHAR(255) NOT NULL DEFAULT ''"
-));
-
-PREPARE stmt FROM @s;
-EXECUTE stmt;
-DEALLOCATE PREPARE stmt;
-
+ALTER TABLE `realmlist` ADD `address6` VARCHAR(255) NOT NULL DEFAULT ''
--- a/sql/updates/auth/2015_09_02_00_auth.sql
+++ b/sql/updates/auth/2015_09_02_00_auth.sql
@@ -1,1 +1,15 @@
-ALTER TABLE `realmlist` ADD `address6` VARCHAR(255) NOT NULL DEFAULT ''
+SET @s = (SELECT IF(
+    (SELECT COUNT(*)
+        FROM INFORMATION_SCHEMA.COLUMNS
+        WHERE table_name = 'realmlist'
+        AND table_schema = DATABASE()
+        AND column_name = 'address6'
+    ) > 0,
+    "SELECT 1",
+    "ALTER TABLE realmlist ADD address6 VARCHAR(255) NOT NULL DEFAULT ''"
+));
+
+PREPARE stmt FROM @s;
+EXECUTE stmt;
+DEALLOCATE PREPARE stmt;
+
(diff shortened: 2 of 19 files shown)


sql/updates/auth/2015_09_02_00_auth.sql
```diff
@@ -0,0 +1,15 @@
...
+        AND column_name = 'address6'
+    ) > 0,
+    "SELECT 1",
+    "ALTER TABLE realmlist ADD address6 VARCHAR(255) NOT NULL DEFAULT ''"
+));
+
+PREPARE stmt FROM @s;
+EXECUTE stmt;
+DEALLOCATE PREPARE stmt;
+
```
DDuarte: Any reason not to execute the query directly? Also missing change in auth_database.sql file

src/server/bnetserver/Realms/RealmList.cpp
```diff
@@ -119,41 +120,56 @@ void RealmList::UpdateRealms(boost::system::error_code const& error)
                 }
 
                 ip::address externalAddress = (*endPoint).endpoint().address();
+              
+		ip::address externalAddress6;
```
DDuarte: tabs to spaces

DDuarte:
How useful is ipv6 if the client can't support it?

sql/updates/auth/2015_09_02_00_auth.sql
TonyHoyle: Just in case the script gets run on a database with the column already there - mysql lacks an 'if column exists' syntax.

DDuarte: These sqls are not required to be re-appliable.

TonyHoyle:
The client supports it for connection to worldservers  and has done for some time.  I was somewhat surprised it wouldn't do it for the battlenet server but as that's a trivial part of the patch (just the BindIp support) and they're likely to add it soon anyway it doesn't hurt to be ahead.

DDuarte:
Oh, I was misinformed, thanks

sql/updates/auth/2015_09_02_00_auth.sql
TonyHoyle: OK I'll just simplify it.

src/server/shared/Realm/Realm.cpp
```diff
@@ -42,8 +42,8 @@ ip::tcp::endpoint Realm::GetAddressForClient(ip::address const& clientAddr) cons
         {
             realmIp = LocalAddress;
         }
-        else
-            realmIp = ExternalAddress;
+        else
```
tkrokli: Is the extra space needed at the end of these 2 lines? Or could they stay as they were?

sql/base/auth_database.sql
```diff
@@ -583,6 +583,7 @@ CREATE TABLE `realmlist` (
   `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `name` varchar(32) NOT NULL DEFAULT '',
   `address` varchar(255) NOT NULL DEFAULT '127.0.0.1',
+  `address6` varchar(255) NOT NULL DEFAULT '',
```
DDuarte: Any particular reason this field can't be shared? `address` for both IPV4 and IPV6 addresses

TonyHoyle: The realm server returns both addresses to the client as separate lists, and the client makes the choice (there's a flag in the client under System/Network plus it probably does some local connectivity checks as well) - therefore the realmlist needs to know both.

It could I guess be done as a comma separated list but you have to then ensure there's only one of each and parse the strings rather than just chucking them at boost with the right flag set.

src/server/shared/Realm/Realm.cpp
TonyHoyle: That's just vi adding random spaces as it does sometimes.. I'll remove them

TonyHoyle:
Hopefully fixed the email in git this time..

sql/base/auth_database.sql
DDuarte: Ah, it's clear now. Two fields are ok then, never been a fan of lists in the DB

tkrokli:
Yes, github is now able to show your user avatar as well as link your username from your commits. :+1:

src/server/bnetserver/Realms/WorldListener.cpp
```diff
@@ -65,7 +68,7 @@ void WorldListener::HandleOpen()
 {
     try
     {
-        _worldSocket->bind(std::string("tcp://*:") + std::to_string(_worldListenPort));
+        _worldSocket->bind(std::string("tcp://"+_bindIp+":") + std::to_string(_worldListenPort));
```
Shauren: Can you explain this change? ZMQ sockets are used for ipc, they have nothing to do with clients

sql/base/auth_database.sql
pete318: Presumably you want it to bind to both IPv4 and IPv6 interfaces. So best to have the ability to set both addresses.

src/server/bnetserver/Realms/WorldListener.cpp
TonyHoyle: This simply makes sure that bindIp is honoured.  The old code simply bound to all addresses which is not good on a multihomed server.  Binding always to loopback is also a solution provided the IPC is always local.

It's possible that particular change belongs in its own pull request as it's not directly related to ipv6 - I just noticed it when I was going through the various bits of networking code.

Shauren: Indeed it belongs to its own PR.
Also it doesnt have to be always local - bnetserver can be somewhere else.

TonyHoyle:
Tested on a VM with no ipv6 stack, and everything worked normally.

Currently working on the opposite case of an ipv6 only worldserver.

Aokromes:

tkrokli:
Not sure if you are aware of this, but you need to rebase your own 6.x branch to solve the conflicts.
Your branch is currently 13 commits behind TrinityCore:6.x . This is mostly due to the new 6.2.2 updates.

TonyHoyle:
It's done on my machine, alas I haven't rebased before and lost a bunch of stuff so I'll have to get back to a working state first.

TonyHoyle:
Tested with an ipv6 only worldserver.. works OK.  Also did some cleaning up of the patch (trailing spaces mostly).

sql/updates/auth/2015_09_02_00_auth.sql
TonyHoyle: I put the check back - the bnetserver does try to reapply these files if it thinks they've changed, and it was stopping me starting the server at all...

DDuarte: Don't. That re-apply on update is a feature. It does not re-apply under normal conditions (once this is added to the repo).

src/server/bnetserver/Server/Session.cpp
```diff
@@ -1141,7 +1144,11 @@ Battlenet::WoWRealm::ListUpdate* Battlenet::Session::BuildListUpdate(Realm const
         version << buildInfo->MajorVersion << '.' << buildInfo->MinorVersion << '.' << buildInfo->BugfixVersion << '.' << buildInfo->Build;
 
         listUpdate->Version = version.str();
-        listUpdate->Address = realm->GetAddressForClient(GetRemoteIpAddress());
+        // Prioritise ipv4, for compatibility
```
Shauren: This address is always v4 - client reads it as such.

TonyHoyle: In the case where this happens (ipv6 only worldserver), there isn't an ipv4 address available.  The client seems to handle it.

Shauren: Because this block is not sent by default.

Aokromes:
+ rebase plz.

Aokromes:
@TonyHoyle 

If TonyHoyle don't answers, anyone can takeover this work?

jackpoz:
This feature will be implemented sometime in the future by Shauren, taking care of merge conflicts with current sources. Thank you for the contribution.

PR #15434, opened 2015-09-05
Core/Taxi: Updated TaxiMaskSize to 6.2.2.20444.

Luzifix:
Hello i update the TaxiMaskSize, to the new Max from db2 file


Shauren:
We don't load 6.2.2 DBC/DB2 files yet. This is currently being worked on.

PR #15439, opened 2015-09-05
Split worldSocket bind fix 

TonyHoyle:
Split from PR #15422

The existing code creates a ZMQ socket bound to all addresses, which is somewhat unexpected on a multihomed server (and an issue if you decide to run multiple servers bound to IPs).

This simply passes the BindIp parameter to RealmList so it obeys the user configuration.

### DIFF
--- a/src/server/bnetserver/Realms/WorldListener.cpp
+++ b/src/server/bnetserver/Realms/WorldListener.cpp
@@ -32,8 +32,9 @@ WorldListener::HandlerTable::HandlerTable()
 #undef DEFINE_HANDLER
 }
 
-WorldListener::WorldListener(uint16 worldListenPort) : _worldListenPort(worldListenPort)
+WorldListener::WorldListener(const std::string & bindIp, uint16 worldListenPort) : _worldListenPort(worldListenPort)
 {
+    _bindIp = bindIp;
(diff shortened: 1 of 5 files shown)


src/server/bnetserver/Realms/WorldListener.cpp
```diff
@@ -65,7 +66,7 @@ void WorldListener::HandleOpen()
 {
     try
     {
-        _worldSocket->bind(std::string("tcp://*:") + std::to_string(_worldListenPort));
+        _worldSocket->bind(std::string("tcp://"+_bindIp+":") + std::to_string(_worldListenPort));
```
DDuarte: This changes the default from * to 0.0.0.0, is there any impact in that or those two are equivalent?

TonyHoyle:
They appear to be equivalent according to my tests (they should be - 0.0.0.0 is the standard way of saying 'any').

I checked anyway just in case - in zmq in tcp_address.cpp (line 234) it has the comment ' "*" resolves to INADDR_ANY or in6addr_any' which is the same as 0.0.0.0, so they're the same.

Gacko:
0.0.0.0 is the standard way of saying 'any IPv4 address'. Keep IPv6 in mind.

Shauren:
No longer relevant - zmq will be removed.

PR #15483, opened 2015-09-10
DB/World schema backport: updated quest_template table

FrancescoBorzi:
All changes are made according to 6.x


StormBytePP:
The full database modification is missing, rendering new installs unable to start as of different fields

DDuarte:
world_database.sql should ONLY be updated in TDB releases, there's nothing missing in this commit

PR #15485, opened 2015-09-10
Core/Taxi: Implemented taxi routes

StormBytePP:
This PR implements taxi in 6.x branch.
Help is needed for testing/resolve the last problems

Fixed current problems:
- [x] When you start a flight, you always see in client the message "There's no vendor nearby" despite flight starts
- [x] Missing request early landing
- [x] Flight end does not end where it should
- [x] Update the starting known nodes, as at least for human, both StormWind and Ironforge are added in retail (investigate all races in this regard)
- [x] Remove _special_ nodes (like quest nodes) from all nodes graph (partially done)
- [x] Try to solve issue #15490 and port it to 6.x
- [x] Investigate in IDA, how client "draws" the path that will be chosen, to adjust algorithm (currently I made it using logic + several client path drawing tests (current algorithm uses client testing + logic, but no reverse)
- [x] Resolve early teleport when you go to a different map, like the route from Ironforge to Shattered Sun Staging Area in which you are teleported when you're still inside IronForge, instead of teleporting later when actually map has to change ( maybe faulty code: https://github.com/TrinityCore/TrinityCore/pull/15485/files#diff-649b0ea06820eb06283d1ccc8c83efdbR548 )

**THIS PR IS OBSOLETE, WORK IS BEING DONE IN TC BRANCH https://github.com/TrinityCore/TrinityCore/tree/6.x_implement_taxi**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,4 +31,4 @@ script:
   - cat sql/updates/hotfixes/*.sql | mysql -utrinity -ptrinity hotfixes
   - mysql -uroot < sql/create/drop_mysql.sql
   - cd bin
-  - make -j 10
+  - make -j 8
--- a/src/server/game/CMakeLists.txt
+++ b/src/server/game/CMakeLists.txt
@@ -139,6 +139,7 @@ include_directories(
   ${CMAKE_CURRENT_SOURCE_DIR}/Entities/Object/Updates
   ${CMAKE_CURRENT_SOURCE_DIR}/Entities/Pet
   ${CMAKE_CURRENT_SOURCE_DIR}/Entities/Player
+  ${CMAKE_CURRENT_SOURCE_DIR}/Entities/Taxi
   ${CMAKE_CURRENT_SOURCE_DIR}/Entities/Totem
   ${CMAKE_CURRENT_SOURCE_DIR}/Entities/Transport
   ${CMAKE_CURRENT_SOURCE_DIR}/Entities/Unit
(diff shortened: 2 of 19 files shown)


StormBytePP:
I am starting to think that clang gets killed and travis fails to build, maybe because too many nested templates in boost graph and property maps that make travis VM run out of memory?

.travis.yml
```diff
@@ -31,4 +31,4 @@ script:
   - cat sql/updates/hotfixes/*.sql | mysql -utrinity -ptrinity hotfixes
   - mysql -uroot < sql/create/drop_mysql.sql
   - cd bin
-  - make -j 10
+  - make -j 8
```
StormBytePP: This change is to prevent travis running out of memory, and clang being killed as of boost graph templates, just in case someone wonders why I change travis build parallel number

Naios: > Travis CI VMs run on 1.5 virtual cores. This is not exactly a concurrency, which allows to parallelize a lot, but it can give a nice speedup depending on your use case.
Which means 8-10 is too high anyway.

Keldo:
https://github.com/Keldo/TrinityCore/commit/2c5653addd214097e8dabe413d2c33a498db64cc

Multiple build errors.
https://github.com/Keldo/TrinityCore/issues/1

joschiwald:
@Keldo it's your fault

Keldo:
I pulled the source from 04de58c and it successfully compiled along with the hunter fixes. Uploading to my Linux Server now to compile and test

Keldo:
I'm using what Boost was available from Debian which is libboost-dev (1.49.0.1), if I need a different version then please let me know. If you don't want to populate this thread with a conversation I'm available at scott@twedev.com

PR #15493, opened 2015-09-11
Core/Scripts: make "The Crone" in Karazhan attackable to players

### LINKED ISSUE #15426: [3.3.5a]The Crone doesn't become attackable during the Opera Event
Revision: 2bcd9c543f1f

After killing Roar, Dorothee, Tinhead, Strawman and Tito, The Crone correctly spawns, but does not become attackable. You drop combat, and if you teleport out of the instance and come back you can start up another event. The Crone will still be present during the new encounter.

dubitabam:
fixes #15426 

second approach :)

### DIFF
--- a/src/server/scripts/EasternKingdoms/Karazhan/bosses_opera.cpp
+++ b/src/server/scripts/EasternKingdoms/Karazhan/bosses_opera.cpp
@@ -105,6 +105,9 @@ void SummonCroneIfReady(InstanceScript* instance, Creature* creature)
     {
         if (Creature* pCrone = creature->SummonCreature(CREATURE_CRONE, -10891.96f, -1755.95f, creature->GetPositionZ(), 4.64f, TEMPSUMMON_TIMED_OR_DEAD_DESPAWN, HOUR*2*IN_MILLISECONDS))
         {
+            if (pCrone->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_PC))
+                pCrone->RemoveFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_PC);
+
             if (creature->GetVictim())
                 pCrone->AI()->AttackStart(creature->GetVictim());
         }


MitchesD:
this isn't correct way, these flags are removed when EnterCombat, so I am not sure if that issue is really valid

Killyana:
The problem is EnterCombat is never triggered due to UNIT_FLAG_IMMUNE_TO_PC. BTW it's the only boss in bosses_opera.cpp  that use EnterCombat to remove this flags, the others are using MoveInLineOfSight to remove UNIT_FLAG_NON_ATTACKABLE

tkrokli:
@daevidpreis : Please update this page to use the correct NPC name in the title and your commit:
"The Crone" - http://www.wowhead.com/npc=18168/the-crone

Since you need to edit the title anyway, I suggest the following wording based on TC standards:
**Core/Scripts: remove UNIT_FLAG_IMMUNE_TO_PC from "The Crone" (Karazhan)**

tkrokli:
@daevidpreis : I checked with the devs in #trinity and verified that @Expecto's  suggestion is correct.

daevidpreis:
@tkrokli ok changed that and the one in line 704 `EnterCombat` too

MitchesD:
As I said before, it's wrong, removing flag in constructor is nonsense, if so, then this flag will be removed for all time, then its useless to remove it in EnterCombat, I think proper way is remove it in MoveLineInSight, or when event starts, but I am sure, that in constructor it's WRONG.

daevidpreis:
@MitchesD advise taken. you think this is the correct way?

MitchesD:
@daevidpreis no, that's not correct, the result is the same, removing flag on spawn. I think correct way is removing on attack start or on move, but for sure not on spawn.

daevidpreis:
i think attack start will never happen as long as the flag is set. i tried it in MoveInLineOfSight, but then players can drop combat and regenerate before engaging her.

MitchesD:
this is still wrong, if you are still interested in, find a better way.

Rushor:
sadly @daevidpreis no longer has time for anything related to tc :/

Issue #4688, opened 2012-01-07
[SPELL] Gem Perfection


jokerstar:
Spell - Gem Perfection ID 55534 http://www.wowhead.com/spell=55534
 When you finish quest you will gain Gem perfection. http://www.wowhead.com/quest=12918
 Problem is when you producing jewels you have chance to make a perfection gem, but it doesn´t work, you still produce normal jewels

Aokromes:
15cafba

PR #15495, opened 2015-09-11
Core/Spell: Implement "Gem Perfection" (and framework for similar spells in the future)

Treeston:
Implements "Gem Perfection".
Adds new DB table skill_perfect_item_template that can be used for similar spells in the future.

(Also removes some magic numbers in the profession specialization script because I stumbled across them while looking for the files I needed.)

Closes #4675 and #4688.

### DIFF
--- a/src/server/game/Skills/SkillExtraItems.cpp
+++ b/src/server/game/Skills/SkillExtraItems.cpp
@@ -20,10 +20,97 @@
 #include "DatabaseEnv.h"
 #include "Log.h"
 #include "Player.h"
+#include "ObjectMgr.h"
 #include <map>
 
 // some type definitions
 // no use putting them in the header file, they're only used in this .cpp
+
+// struct to store information about perfection procs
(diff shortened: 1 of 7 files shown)


src/server/game/Skills/SkillExtraItems.cpp
```diff
@@ -20,11 +20,97 @@
...
+void LoadSkillPerfectItemTable()
+{
+    uint32 oldMSTime = getMSTime();
+
+    SkillPerfectItemStore.clear(); // reload capability
+
+    //                                                  0               1                      2                  3
+    QueryResult result = WorldDatabase.Query("SELECT spellId, requiredSpecialization, perfectCreateChance, perfectItemType FROM skill_perfect_item_template");
+
+    if (!result) {
```
tkrokli: TC codestyle suggests that you move the opening bracket { to a new line and align it with its closing bracket } .

PR #15501, opened 2015-09-13
Creature/SmartAI: Fix creature aggression behavior

### LINKED ISSUE #7197: Rogue: Distract
Distract currently erroneously holds creatures in place if they are engaged while distracted.

From stealth, distract a creature.
While it is distracted, engage it in combat
Run away from it and observe:  The creature will simply face you, but not move until the effects of distract would have normally worn off.

Expected result:  The creature should not be affected by distract in any way once combat has begun.

### LINKED ISSUE #15482: [3.3.5] Creatures putting players in combat despite already being engaged
Branch 3.3.5, revision 6cdfe9dc095b+, TDB 335.59

Walking within aggro range of an aggressive creature that is currently engaged in combat with another creature will make the player enter combat with the first creature. Expected behavior is to have the creature ignore the player if it already has entries on its threat table (unless the player engages it).

Repro examples are the soldier vs undead battles in borean tundra, as well as the sporelok vs bog giant battles in zangarmarsh.

Treeston:
Clean up SmartAI::MoveInLineOfSight. Remove tons of duplicate logic and forward to CreatureAI::MoveInLineOfSight instead of using its own (incorrect) implementation.

Fixes and closes #15482.

Justifications for removed checks:

```
HasReactState(REACT_PASSIVE)? Handled in Creature::CanCreatureAttack
CanAIAttack? Is called by Creature::CanCreatureAttack (to check the above)
CanStartAttack? Is checked by CreatureAI::MoveInLineOfSight
IsHostileTo? Checked by Unit::_IsValidAttackTarget (called from Creature::CanCreatureAttack)
IsWithinDistInMap? Done by Creature::CanStartAttack
IsWithinLOSInMap? Done by Creature::CanStartAttack
```

Also fixes non-SAI creatures not moving to engage while affected by Rogue distract. This fixes and closes #7197.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -453,45 +453,15 @@ void SmartAI::MoveInLineOfSight(Unit* who)
 
     GetScript()->OnMoveInLineOfSight(who);
 
-    if (me->HasReactState(REACT_PASSIVE) || AssistPlayerInCombat(who))
+    if (AssistPlayerInCombat(who))
         return;
 
-    if (!CanAIAttack(who))
-        return;
-
-    if (!me->CanStartAttack(who, false))
-        return;
-
-    if (me->IsHostileTo(who))
-    {
-        float fAttackRadius = me->GetAttackDistance(who);
-        if (me->IsWithinDistInMap(who, fAttackRadius) && me->IsWithinLOSInMap(who))
-        {
-            if (!me->GetVictim())
-            {
-                // Clear distracted state on combat
-                if (me->HasUnitState(UNIT_STATE_DISTRACTED))
-                {
-                   me->ClearUnitState(UNIT_STATE_DISTRACTED);
-                   me->GetMotionMaster()->Clear();
-                }
-
-                AttackStart(who);
-            }
-            else/* if (me->GetMap()->IsDungeon())*/
-            {
-                who->SetInCombatWith(me);
-                me->AddThreat(who, 0.0f);
-            }
-        }
-    }
+    CreatureAI::MoveInLineOfSight(who);
 }
 
 bool SmartAI::CanAIAttack(const Unit* /*who*/) const
 {
-    if (me->GetReactState() == REACT_PASSIVE)
-        return false;
-    return true;
+    return !(me->HasReactState(REACT_PASSIVE));
 }
 
 bool SmartAI::AssistPlayerInCombat(Unit* who)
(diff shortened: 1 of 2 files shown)


tkrokli:
If this change works as intended, it looks like a much needed cleanup. :+1:

pete318:
Yeah I noticed that behaviour when I was working on the stealth changes. But I forgot to look into it further.

One thing you removed though is the clearing of distracted state. Which might post a problem. Can you try for example with a rogue with your PR distracting a creature then moving into combat range? See if they engage you AND chase?

One of the things I corrected was that if you had a creature distracted (either with normal distract, or with the creature alert I added) it would aggro you (correct) but not clear distracted state, so wouldn't chase you until it ended.

src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -453,45 +453,15 @@ void SmartAI::MoveInLineOfSight(Unit* who)
...
     GetScript()->OnMoveInLineOfSight(who);
 
-    if (me->HasReactState(REACT_PASSIVE) || AssistPlayerInCombat(who))
+    if (AssistPlayerInCombat(who))
         return;
 
-    if (!CanAIAttack(who))
-        return;
-
-    if (!me->CanStartAttack(who, false))
```
MitchesD: I think this should not be removed, it's not part of CreatureAI::MoveInLineOfSight(who); and then it's useless to modify CanAIAttack method when its not called there

src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -453,45 +453,15 @@ void SmartAI::MoveInLineOfSight(Unit* who)
 
     GetScript()->OnMoveInLineOfSight(who);
 
-    if (me->HasReactState(REACT_PASSIVE) || AssistPlayerInCombat(who))
+    if (AssistPlayerInCombat(who))
         return;
 
-    if (!CanAIAttack(who))
```
MitchesD: I think this should not be removed, it's not part of CreatureAI::MoveInLineOfSight(who); and then it's useless to modify CanAIAttack method when its not called there

pete318:
When I looked at this before. I was actually thinking this should be resolved at the GridNotifiers stage. E.g. when the creature recognizes your presence and usually acts to engage. It should ask itself "Am I already busy smashing someone elses head in?" if so, not put the new player in combat.

However, this is a bit different in an instance situation I think. So, it's a bit complex.

Treeston:
Yeah, and as it turns out, in current TC core creatures actually exhibit exactly that behavior if they aren't SmartAI - because that check is only done in SmartAI code.

Fixed now.

pete318:
That's odd. One of the creatures I was testing with was the boars in HFP. Sure they're not SmartAI.. At least I wouldn't expect so. They just melee really. Hmm, maybe they have an enrage.

Probably though, it's right it should have gone there in the first place.

Treeston:
Rebased branch and squashed commits.

Issue #15177, opened 2015-07-28
[3.3.5] Hover Mode not working (55$ bounty)


Keader:
In the air phase of Mimiron, he does not remain floating. After a few seconds of battle it ceases to hover and keep attacking on the floor.

Mimiron (Airphase)
![wowscrnshot_070815_125924](https://cloud.githubusercontent.com/assets/7120913/8575259/6f0d8f58-2571-11e5-8a73-db00d86baf36.jpg)

> bug caused by incomplete hover movement implementation
> 
> current hover movement implementation works for static npcs only, they start moving and hover position goes to ground
> 
> on movement target destination is relocated to ground, hover offset is missing

rev b63082971fac8e239f08db0d16113c46c7c1e3ee

joschiwald:
>bug caused by incomplete hover movement implementation

>current hover movement implementation works for static npcs only, they start moving and hover position goes to ground

Keader:
@LuqJensen Have you made any progress on the hover mode? (https://github.com/TrinityCore/TrinityCore/pull/12268#issue-35626540)

LuqJensen:
no, im not working on this anymore. this bug is beyond me and i have no intention to spend time on trying to fix it

Keader:
Nobody is working on this?

Aokromes:
I added 15$ bounty here from trinitycore rewards, also i enabled something called "Tweet from @Bountysource (+$20)?" so maybe it's increased to 35$.

sirikfoll:
For the brave soul that takes interest in this issue, this is my failed attempt https://gist.github.com/sirikfoll/0ef4c294d6973fbb29328fe414b361e2
includes a little mod to easily test the hover in Mimiron(forcing phase 3 right before aggro)

Keader:
@Nawuko any news on this :question:

funjoker:
For 6.x too?

Keader:
i think yes

w5860363:
Flags_extra 128 submitted to the
The problem is that I don't remember the date.

Keader:
@w5860363 what?

w5860363:
@ke

tkrokli:
@Keader : perhaps his native language needs less words to say things than English does. Take a look at
```cpp
if ((c->HasReactState(REACT_AGGRESSIVE) || c->IsTrigger()) && !c->HasUnitState(UNIT_STATE_SIGHTLESS))
```
[...]

Issue #15161, opened 2015-07-25
[3.3.5] VMap data is sometimes not correct


jameyboor:
Rev : :link: https://github.com/TrinityCore/TrinityCore/commit/7d9843c380cf00f38ad0c5bc75fd387e50ee11da 

~~First thing: The Z position of collision enabled GO's isn't taken into account when updating the movement flags of creatures in `Creature::UpdateMovementFlags()`. It's "allowed" Z position is taken with an overloaded function of `Map::GetHeight()` that doesn't even use the dynamictree.
`float ground = GetMap()->GetHeight(GetPositionX(), GetPositionY(), GetPositionZMinusOffset());` uses `float Map::GetHeight(float x, float y, float z, bool checkVMap /*= true*/, float maxSearchDist /*= DEFAULT_HEIGHT_SEARCH*/) const` which doesn't take the dynamictree into account, it should be using `float Map::GetHeight(uint32 phasemask, float x, float y, float z, bool vmap/*=true*/, float maxSearchDist/*=DEFAULT_HEIGHT_SEARCH*/) const` so I've changed that.~~

Next thing, and my real issue: Some positions on certain GameObject's that are spawned in the air don't have correct FloorZ coords and they just use the FloorZ of the ground, which is GroundZ , this causes creatures that are aggro'd on GameObject's to fall through the GameObject because the core thinks the creature is flying. This doesn't happen to players because (I think) the client does have the correct height data of these GameObject's and the player's position is checked client side too, which allows them to stay on the GameObject.

An example of this problem is with the GameObject `Nexus Raid Platform` entry : `193070`
Me standing on the center of the platform using .gps : http://puu.sh/jcltV/a4e28abfed.jpg , this shows FloorZ at ~40 which is totally correct.
Me standing on a random spot somewhere else using .gps : http://puu.sh/jclrR/07d43fd663.jpg , this shows FloorZ at 0.5 which is not correct at all, it's most likely the FloorZ of the Ground beneath me.

Another example is with a creature spawned, the .gps information shown on the right is from the only creature on that platform, so is the `UpdateMovementFlags` call because I added a conditional breakpoint : http://puu.sh/jcniX/508acaad94.jpg

If anyone has some more information regarding this I'd love to hear it.

jameyboor:
Note, this has nothing to do with phasemasks, this has to do with GO's in general.

tkrokli:
[...]

Issue #15032, opened 2015-07-08
[3.3.5][Ulduar] XT002 Entering the ground


Keader:
XT002 - 33293

During the fight, depending on where you move this happens:

XT002
![wowscrnshot_070815_125151](https://cloud.githubusercontent.com/assets/7120913/8575252/62c39404-2571-11e5-9416-893b5ed9ab8e.jpg)

rev b574c99549bb8cf1d95bd4bc0c319548a678799a

Rushor:
you tried disabling the movement maps?
http://rgtc.wikia.com/wiki/Disables_tc2 - well pretty old link but meh

joschiwald:
mimiron bug caused by incomplete hover movement implementation

xt bug caused by the stairs under the floor

Killyana:
For XT it's the mmaps but the vmaps, with vmap.enableHeight = 0 the issue doesn't happen.

P-Kito:
Movementmaps have nothing to do with this.

Keader:
@Rushor this is a good choice? No other boss would have problems?

Rushor:
i was just my first idea, but as the others said, it should be done in another way

try to use use:
33554432 = Hover mode (bytes1) in creature_template_addon for mimiron

joschiwald:
Won't work, current hover movement implementation works for static npcs only, they start moving and hover position goes to ground

ccrs:
@joschiwald i think its related to how movementflags are updated

joschiwald:
on movement target destination is relocated to ground, that's all, hover offset is missing

Keader:
No one has any idea of correction?

Frytiks:
Disable mmaps for boss

Keader:
How? And that would not cause others boss give problem?

tkrokli:
@Rushor : the TC wiki has a better page for the `disables` table: http://collab.kpsn.org/display/tc/disables

@Keader : try the following Query (not sure if it is 100% correct, not tested) :
```sql
DELETE FROM `disables` WHERE `entry` IN (33293,33350);
INSERT INTO `disables` (`sourceType`,`entry`,`flags`,`params_0`,`params_1`,`comment`) VALUES
(7,33293,0,0,0,'Disable mmaps - XT002'),
(7,33350,0,0,0,'Disable mmaps - Mimiron');

Rushor:
the documetation was down at the time i wrote this

[...]

PR #15502, opened 2015-09-13
Creature/BossAI/Sapphiron: Fixed minor bugs/quirks

Treeston:
- Sapphiron will now properly re-engage if the previous attempt ended while he was in flight.
- Sapphiron will now properly be hovering while in the air phase instead of standing idly on the ground.

### DIFF
--- a/src/server/scripts/Northrend/Naxxramas/boss_sapphiron.cpp
+++ b/src/server/scripts/Northrend/Naxxramas/boss_sapphiron.cpp
@@ -43,6 +43,7 @@ enum Spells
     SPELL_BERSERK           = 26662,
     SPELL_DIES              = 29357,
     SPELL_CHILL             = 28547,
+    SPELL_CHECK_RESISTS     = 60539,
 };
 
 enum Phases
@@ -67,7 +68,8 @@ enum Events
     EVENT_EXPLOSION,
     EVENT_LAND,
     EVENT_GROUND,
-    EVENT_BIRTH
+    EVENT_BIRTH,
+    EVENT_CHECK_RESISTS
 };
 
 enum Misc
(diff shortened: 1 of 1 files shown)


src/server/scripts/Northrend/Naxxramas/boss_sapphiron.cpp
```diff
@@ -123,7 +123,12 @@ class boss_sapphiron : public CreatureScript
                 _Reset();
 
                 if (_phase == PHASE_FLIGHT)
+                {
                     ClearIceBlock();
+                    me->SetReactState(REACT_AGGRESSIVE);
+                }
+
+                me->SetFullHealth();
```
chaodhib: Is this necessary?  Doesn't _Reset() take care of that?

Treeston: It doesn't reset react state. Try resetting Sapphiron (vanish etc) while he's in flight phase - proximity aggro will be disabled.

Shauren: using SetFullHealth is not correct, regen after evade needs to be fixed (in the core, not scripts)

ccrs: script should be updated to use UpdateVictim()

Treeston: UpdateVictim() is already used. It's just below `events.Update(diff)` (line 238 in latest).

src/server/scripts/Northrend/Naxxramas/boss_sapphiron.cpp
```diff
@@ -303,6 +306,7 @@ class boss_sapphiron : public CreatureScript
                             case EVENT_LIFTOFF:
                                 Talk(EMOTE_AIR_PHASE);
                                 me->SetDisableGravity(true);
+                                me->SetHover(true);
```
Shauren: I cannot confirm this without sniffs but from experience I can tell you that true hover is very rarely used - so this is likely to be incorrect

Treeston: What's a better way to tell the client to display the hover idle animation?

tkrokli: Browse through [3.3.5/Ulduar] Miniron's Hover Mode (#15177),  [3.3.5/Ulduar] XT002 Entering the ground (#15032) and  [3.3.5] VMap data is sometimes not correct (#15161) for ideas.

PR #15503, opened 2015-09-13
Creature/BossAI/Patchwerk: Fix Hateful Strike behavior

Treeston:
- Hateful Strike can only target the highest 1 (10m)/ 2 (25m) threat targets that aren't the current target. Out of these targets, Hateful Strike hits the highest current HP target.
- Hateful Strike now properly adds threat to all targets eligible for it, plus the current target if it is in melee range. Actual threat amount is not from sniffs - please update this if you have conclusive retail data.

### DIFF
--- a/src/server/scripts/Northrend/Naxxramas/boss_patchwerk.cpp
+++ b/src/server/scripts/Northrend/Naxxramas/boss_patchwerk.cpp
@@ -49,6 +49,11 @@ enum Misc
     ACHIEV_MAKE_QUICK_WERK_OF_HIM_STARTING_EVENT  = 10286
 };
 
+enum HatefulThreatAmounts
+{
+    HATEFUL_THREAT_AMT  = 1000,
+};
+
 class boss_patchwerk : public CreatureScript
 {
 public:
@@ -92,8 +97,8 @@ public:
             _EnterCombat();
             Enraged = false;
             Talk(SAY_AGGRO);
-            events.ScheduleEvent(EVENT_HATEFUL, 1000);
-            events.ScheduleEvent(EVENT_BERSERK, 360000);
+            events.ScheduleEvent(EVENT_HATEFUL, 1 * IN_MILLISECONDS);
+            events.ScheduleEvent(EVENT_BERSERK, 6 * MINUTE * IN_MILLISECONDS);
 
             instance->DoStartTimedAchievement(ACHIEVEMENT_TIMED_TYPE_EVENT, ACHIEV_MAKE_QUICK_WERK_OF_HIM_STARTING_EVENT);
         }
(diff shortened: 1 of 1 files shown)


src/server/scripts/Northrend/Naxxramas/boss_patchwerk.cpp
```diff
@@ -49,6 +49,14 @@ enum Misc
     ACHIEV_MAKE_QUICK_WERK_OF_HIM_STARTING_EVENT  = 10286
 };
 
+// These numbers are taken from anecdotal statements found on the interwebs.
+// Mainly based off of statements about tanks' TPS "going up by X amount on patchwerk" and then reverse calculating how much every hateful must give to reach this.
+// If you have actual numbers taken from retail, please update these!
+enum HatefulThreatAmounts {
+    HATEFUL_THREAT_10M = 2000,
+    HATEFUL_THREAT_25M = 2000,
```
ccrs: same values?

Treeston: Intentional - separated in case they end up being different when/if someone updates with "real" data.
The compiler optimizes it out anyway if the values are the same.

src/server/scripts/Northrend/Naxxramas/boss_patchwerk.cpp
```diff
@@ -111,37 +119,68 @@ class boss_patchwerk : public CreatureScript
...
                     {
-                        //Cast Hateful strike on the player with the highest
-                        //amount of HP within melee distance
-                        uint32 MostHP = 0;
-                        Unit* pMostHPTarget = NULL;
+                        // Hateful Strike targets the highest non-MT threat in melee range on 10man
+                        // and the higher HP target out of the two highest non-MT threats in melee range on 25man
+                        float MostThreat = 0.0f;
+                        Unit* pSecondThreatTarget = NULL;
+                        Unit* pThirdThreatTarget = NULL;
```
ccrs: u should rename those to TC stardard thirdThreatTarget or _thirdThreatTarget

ccrs:
````note
A rapid and powerful melee strike on a single target/two targets for 20,000-27,000/~80,000 physical damage. Cast every second. Hateful Strike also adds threat to the three players who have the highest threat. The target of Hateful Strike is determined by three requirements:

- Proximity: Targets must be within melee range of Patchwerk.
- Threat: Targets must be less than first in threat, but at least second/third highest threat. This requirement does not apply if only one target is in melee range.
- Current Health: The target must have the highest health of all available targets.
````

Is "Current Health" requirement still checked?

**Comment by Treeston** [comment]
154-161:
```
Unit* pHatefulTarget = NULL;
if (!pThirdThreatTarget)
    pHatefulTarget = pSecondThreatTarget;
else if (pSecondThreatTarget)
    pHatefulTarget = (pSecondThreatTarget->GetHealth() < pThirdThreatTarget->GetHealth()) ? pThirdThreatTarget : pSecondThreatTarget;

if (!pHatefulTarget)
    pHatefulTarget = me->GetVictim();
```

chaodhib:
It doesn't seemed mentioned yet so I'm throwing this wowhead comment on the boss: http://www.wowhead.com/npc=16028/patchwerk#comments:id=552113

chaodhib:
Are you sure the current target isn't eligible for hateful strike? Also, are you sure the current target shouldn't have a bonus threat too?

Some math should be done on this but let's image this: Tank A has aggro and generates X TPS. Tank B is second in threat and generates Y TPS. With your commit, tank B will actually have (Y + 1k/CD of HS) TPS. If this amount is higher than X, the tank and the off tank will swap aggro during the whole fight.

PR #15508, opened 2015-09-14
AuctionHouse: Deposit - Fixed config value of 0 being ignored.

lucas-wb:
Please talk slowly I'm a little dull

-- Realized I used tabs instead of spaces as per the coding standards, added another commit. Enlighten me if there is a better way to do so

### DIFF
--- a/src/server/game/AuctionHouse/AuctionHouseMgr.cpp
+++ b/src/server/game/AuctionHouse/AuctionHouseMgr.cpp
@@ -63,6 +63,9 @@ AuctionHouseObject* AuctionHouseMgr::GetAuctionsMap(uint32 factionTemplateId)
 
 uint32 AuctionHouseMgr::GetAuctionDeposit(AuctionHouseEntry const* entry, uint32 time, Item* pItem, uint32 count)
 {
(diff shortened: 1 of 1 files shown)


DDuarte:
`GetAuctionDeposit` returns `AH_MINIMUM_DEPOSIT` (100) if the deposit is lower than that. Why shouldn't that be the case for 0?

Tomatodev:
That's the problem, in the world config you put in a multiplier for the deposit cost. If you put 0 into the config, indicating you want there to be no deposit, it multiplies the deposit by 0, but then immediately sets it to AH_MINIMUM_DEPOSIT since its < (100).

jackpoz:
This PR looks ok, could you squash the commits so it can be merged ?

StormBytePP:
@Tomatodev why close without reason?

Tomatodev:
Had problems squashing, deleted the wrong source repo. Was going to resubmit with a new source repo

StormBytePP:
but you can undelete it, re-clone, and try squash again :)

jackpoz:
@Tomatodev there's no hurry, take your time to learn how to squash. you need to learn it only once, after than it's gonna be easier :)

PR #15525, opened 2015-09-16
Scripts/SmartAI: Edit a bunch of scripts to not spam "is calling for help" / "is trying to run away" emotes.

Treeston:
Just a quickie. Title says it all.


tkrokli:
SQL-only fixes are not supposed to be put in a PR.
Create a regular issue for this and put the SQL in a box at the end of your post.

PR #14199, opened 2015-02-21
Combat with distant creatures

r00ty-tc:
This PR should resolve issue:

Combat bug due to HostileReferences out of range (https://github.com/TrinityCore/TrinityCore/issues/12100)

and any others that are generally called "the combat bug"

Here's what happens:

When you engage combat with a creature, you're set in combat with it and it with you. However, if for some reason you leave the creature behind and it remains in another cell (with no players in), the Update() for that creature won't be run any more. This it will never leave combat. This was easy to reproduce by entering combat with NPCs that will aid you, let them pick up the slack and move on.

The solution is to add the creature to the active objects list while still in combat with a player or pet. When leaving combat, if this active mode was set it's cleared (but only if we set it).

For me it resolves most times it happens. But I think it needs some testing by others, specifically those with more players than 1.

PR #15538, opened 2015-09-17
Combat with distant creatures (Mk II)

r00ty-tc:
New PR from closed https://github.com/TrinityCore/TrinityCore/pull/14199

Totally rewrote this to instead check if player is in combat, and then iterate the player hostile unit list. If a unit is a creature and is more than the player's visibility distance away, the cell for that creature will be visited.

This avoids the pitfalls of setting a creature as active.

Any comments on the viability of this solution are welcomed.

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -678,6 +678,26 @@ void Map::Update(const uint32 t_diff)
         player->Update(t_diff);
 
         VisitNearbyCellsOf(player, grid_object_update, world_object_update);
+
+        // Handle updates for creatures in combat with player and are more than 60 yards away
+        if (player->IsInCombat())
+        {
+            std::vector<Creature*> updateList;
+            HostileReference* ref = player->getHostileRefManager().getFirst();
+
+            while (ref)
+            {
+                if (Unit* unit = ref->GetSource()->GetOwner())
+                    if (unit->ToCreature() && unit->GetMapId() == player->GetMapId() && !unit->IsWithinDistInMap(player, GetVisibilityRange(), false))
+                        updateList.push_back(unit->ToCreature());
+
+                ref = ref->next();
+            }
+
+            // Process deferred update list for player
+            for (Creature* c : updateList)
+                VisitNearbyCellsOf(c, grid_object_update, world_object_update);
+        }
     }
 
     // non-player active objects, increasing iterator in the loop in case of object removal


DDuarte:
Can it be guaranteed that this eventually exists the while loop?

pete318:
You mean ref might be empty? Then the while loop will be skipped entirely surely? The while (ref) has an implied if (ref) attached I'd say.

DDuarte:
I mean `ref->next()` never returning null

pete318:
That's the condition for the end of the list.

So, if the initial ref is null, then it never enters the while loop. If it has something in, it enters at least once. In the event that there was only one creature on the hostile references it would reach ref->next get null, and exit the loop there. If there were more it would go around until ref->next returned null.

So, I think it's safe.

src/server/game/Maps/Map.cpp
```diff
@@ -678,6 +678,20 @@ void Map::Update(const uint32 t_diff)
...
+
+        // Handle updates for creatures in combat with player and are more than 60 yards away
+        if (player->IsInCombat())
+        {
+            HostileReference* ref = player->getHostileRefManager().getFirst();
+
+            while (ref)
+            {
+                if (Unit* unit = ref->GetSource()->GetOwner())
+                    if (unit->GetTypeId() == TYPEID_UNIT && !unit->IsWithinDist2d(player->PositionXYStream().Pos, 60.0f))
```
pete318: I am wondering whether there is a point to checking the distance. Since, if the cell has already been visited (by the player iterator above for example) then it won't be visited again anyway.

Maybe someone knows how expensive the distance check vs the checks in the Visit route are and can advise?

jackpoz: > someone knows how expensive the distance check vs the checks in the Visit route are

only a profile session knows the answer to this question

jackpoz: btw why do you need
```cpp
player->PositionXYStream().Pos
```
to get the position ? what's wrong with "player" itself ?

pete318: Good point. I didn't see IsWithinDistInMap. I also think I need to make a check for being on the same map. Generally you're removed from combat when shifting map.. But still.

streetrat:
isn't VisitNearbyCellsOf "visiting" all objects in the cells nearby the target unit?
if yes and player has multiple (lets say 100) creatures on his/her threathlist in the same grid/cell far away then it will visit that same cell 100 times but 1 would be enough.

i would better make *all* creatures active while they are in any kind of combat so they can evade and run home whenever they need, the only drawbacks i can think of are some mini-world-events where creatures are attacking eachother, but actually those fights should be mostly only be "visual" fights, without actual combat.

pete318:
Yes, and no. The cell visit code is quite clever in this respect. It only visits a cell once per Update(). Inside the visit code it marks a cell once visited. If it's marked it won't visit again. That list is reset at the start of the Update() code.

```C++
            if (isCellMarked(cell_id))
                continue;

            markCell(cell_id);
```

The original version of this set the creatures active with setActive. A few problems with that.

1: It causes crashes, pretty much impossible to replicate since it needs a lot of activity going on. This was reported with the previous PR and as such I've moved to this method.
2: It also makes the creatures visible ANYWHERE in the grid. Now imagine a very busy grid with many players all enaged in combat.. The amount of data sent to every client, and the fact you'd be seeing tons of combat going on far away.

No, setActive isn't the solution here.

First of all if a player is in combat with 100 creatures it would be a VERY rare situation that they were all more than 60 yards away. But, if they were all more than 60 yards away, they'd have to all be in different cells to make 100 cell visit calls.

I did consider just calling creature Update() however, the main reason for this combat bug to happen is when the creature is engaged with another NPC. In which case, just updating the creature would create the odd situation that a low level non elite would slog away for ages at a likely elite NPC and eventually win, leaving the player in combat for a very long time and not producing a realistic result.

Hence I settled on this solution.

pete318:
Honestly not sure what can be done about this one.

The exact same method to iterate is used everywhere. While I understand the concept of what may be happening (list is updated in another thread while this is being processed). I don't understand why this doesn't happen anywhere else the exact same method is used.

Now, there's a couple of things I could try here. From what I can see this same iterator is used EVERYWHERE.. However, not in Map.cpp. So it could be something to do with the multiple map threads.

I could move the code to Player Update(). However, in my mind even if Player functions exist in one thread, the maps are still busily being updated around it. So, not convinced it will solve the problem. But I feel I'd be making a hasty stab in the dark based on only a partial knowledge of how this aspect works.

Some input from those with an understanding of the threading used in TC would be helpful here.

jackpoz:
It looks much more simple to me, without even involving multiple threads: VisitNearbyCellsOf(unit, grid_object_update, world_object_update) is modifying player->getHostileRefManager() linked list.

pete318:
It's (as you can see by the commit, made literally a minute or so before your comment) the exact same conclusion I came to.

One question of course is.. Do I go this route (process ALL player combat creatures at once) or, have the list stored and iterated per player?

But either should prevent the crash I think.

pete318:
I was looking at list vs vector. And vector won in this scenario. I'll check on set. I think set it (mainly) designed for use with a key + value. But, let me see.

One possible solution I looked at was keeping it in the player iterator, and pre-allocating the vector to the size of the threat list.. Sure vector might end up with a few less in, but the cost should be less.

pete318:
I am tending to agree set might be better and keep it where it is. So a unique iteration per Update() for creatures in combat with a player.

jackpoz:
Moving the container inside the player loop might keep the iterator to smaller sizes, it might also be easier to debug having a player update followed by his hostile creatures' updates. In this case any container will be ok since all elements are unique (set allows to store unique elements, unordered_set removes the order).
It's mostly a codestyle matter so there's no correct answer.

pete318:
I think, this is going to be one of those questions of scale (the only time this kind of thing matters at).

The main question is, whether the cost of inserting into a set to ensure Visit is only run on unique creatures is higher or lower than the cost of the check in VisitNearby.... for whether the cell has already been visited, coupled with whether multiple players are likely to be in combat AND far away from the same creature.

My gut feeling is, the most likely scenario is on a busy server there will be at any time a number of creatures in combat with a player.. A much smaller number more than visibility distance away (another change I forgot to document there) and much much smaller the number that are in combat with more than one player, both of which are out of visibility range.

So, maybe it's best to stay with Vector, keep it in the player iterator (so the events happen in a more logical order as you say) and see how it goes from there.

src/server/game/Maps/Map.cpp
```diff
@@ -678,6 +678,31 @@ void Map::Update(const uint32 t_diff)
...
+                ref = ref->next();
+            }
+
+            // Process deferred update list for player
+            if (!updateList.empty())
+            {
+                for (uint32 i = 0; i<updateList.size(); i++)
+                    VisitNearbyCellsOf(updateList[i], grid_object_update, world_object_update);
+
+                updateList.clear();
```
Shauren: such a minor thing but ill say it anyway - no point in clearing this container, it goes out of scope after this block anyway

pete318: Yep. I wasn't sure if there were memory implications of letting it go out of scope with data in. But I read since (when comparing the various types discussed) that they have a proper destructor anyway that will take care of it.

StormBytePP: It's storing pointers, and STL containers, *never* deletes the pointers when they are being destructed, so it's safe.

pete318:
Removed un-needed clear, and also the check on unit type (converting to creature implies that test).

Squashed to a single commit.

src/server/game/Maps/Map.cpp
```diff
@@ -678,6 +678,26 @@ void Map::Update(const uint32 t_diff)
...
+        // Handle updates for creatures in combat with player and are more than 60 yards away
+        if (player->IsInCombat())
+        {
+            std::vector<Creature*> updateList;
+            HostileReference* ref = player->getHostileRefManager().getFirst();
+
+            while (ref)
+            {
+                if (Unit* unit = ref->GetSource()->GetOwner())
+                    if (unit->ToCreature() && unit->GetMapId() == player->GetMapId() && !unit->IsWithinDistInMap(player, GetVisibilityRange(), false))
```
Runico: I've seen creature's are in the HostileReference of player and they are not in the same map of player

Issue #2947, opened 2011-09-08
[Spell] Slime Spray


Anonimo17:
Rotface ability Slime Spray (id: 69508) isn't hitting the original trigger's cone, but he hits a cone in front of him, given by the tank usually.

I don't know if it's the same issue as Ick's spray toxic waste.

Rev: e67a33eaaa5e1c1f46ba268e02254a087b0f2ec3

Golrag:
Try
```diff
diff --git a/src/server/game/Entities/Unit/Unit.cpp b/src/server/game/Entities/Unit/Unit.cpp
index 4597f18..63c722f 100644
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12765,7 +12765,15 @@ Unit* Creature::SelectVictim()
 
     if (target && _IsTargetAcceptable(target) && CanCreatureAttack(target))
     {
-        SetInFront(target);
+        if (WorldObject* channelObject = sObjectAccessor->GetWorldObject(*this, GetChannelObjectGuid()))
+        {
+            if (IsTrackingChannelObject())
+                if (!isInFront(channelObject))
+                    SetInFront(channelObject);
+        }
+        else
+            SetInFront(target);
+
         return target;
     }
 
@@ -17986,3 +17994,12 @@ void Unit::Whisper(uint32 textId, Player* target, bool isBossWhisper /*= false*/
     ChatHandler::BuildChatPacket(data, isBossWhisper ? CHAT_MSG_RAID_BOSS_WHISPER : CHAT_MSG_MONSTER_WHISPER, LANG_UNIVERSAL, this, target, bct->GetText(locale, getGender()), 0, "", locale);
     target->SendDirectMessage(&data);
 }
+
+bool Unit::IsTrackingChannelObject() const
+{
+    if (Spell* spell = GetCurrentSpell(CURRENT_CHANNELED_SPELL))
+        if (SpellInfo const* spellInfo = spell->GetSpellInfo())
+            return spellInfo->HasAttribute(SPELL_ATTR1_CHANNEL_TRACK_TARGET);
+
+    return false;
+}
diff --git a/src/server/game/Entities/Unit/Unit.h b/src/server/game/Entities/Unit/Unit.h
index 39b1c34..6cfcbe3 100644
--- a/src/server/game/Entities/Unit/Unit.h
+++ b/src/server/game/Entities/Unit/Unit.h
[...]

Issue #10803, opened 2013-09-13
Boss Cast Directions - Anub'arak and Ingvar


drpepper2:
Core: 458f5b9502da
DB: 335.52

So this is weird, not really sure how to categorize this:

Two bosses I can think of in particular do this: 

Ingvar the Plunderer with his Smash and Dark Smash, and Anub'arak in Azjol Nerub.

How it's supposed to work: He is supposed to cast smash/dark smash/pound, and if the tank stays in front of him, takes a bunch of damage, but if the tank moves (usual path was to run through them), then they don't take this damage.

What happens: The boss stays facing the correct direction while casting, but then immediately turns and hits the tank, no matter what, making this damage unavoidable.

NPCs
Ingvar - http://www.wowhead.com/npc=23954
Anub'arak - http://www.wowhead.com/npc=29120

Abilities
Smash - http://www.wowhead.com/spell=59706
Dark Smash - http://www.wowhead.com/spell=42723
Pound - http://www.wowhead.com/spell=59433

Treeston:
Still happening as of 0cdf3b7a73d8, see referenced issue.

Exodius:
@Treeston close this as per f481ae1048aac32ba3a40491f0304d0317234295?

PR #15546, opened 2015-09-19
Core/Spell: Fixed serverside creature orientation while casting

Golrag:
Closes https://github.com/TrinityCore/TrinityCore/issues/2947
Any feedback is welcome ! :)

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12764,10 +12764,7 @@ Unit* Creature::SelectVictim()
         return NULL;
 
     if (target && _IsTargetAcceptable(target) && CanCreatureAttack(target))
-    {
-        SetInFront(target);
         return target;
-    }
 
     // Case where mob is being kited.
     // Mob may not be in range to attack or may have dropped target. In any case,
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -1370,11 +1370,11 @@ bool Creature::CanStartAttack(Unit const* who, bool force) const
         return false;
 
     // This set of checks is should be done only for creatures
-    if ((HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_NPC) && who->GetTypeId() != TYPEID_PLAYER)                                   // flag is valid only for non player characters 
+    if ((HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_NPC) && who->GetTypeId() != TYPEID_PLAYER)                                   // flag is valid only for non player characters
         || (HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_PC) && who->GetTypeId() == TYPEID_PLAYER)                                 // immune to PC and target is a player, return false
         || (who->GetOwner() && who->GetOwner()->GetTypeId() == TYPEID_PLAYER && HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_PC))) // player pets are immune to pc as well
         return false;
-    
+
     // Do not attack non-combat pets
     if (who->GetTypeId() == TYPEID_UNIT && who->GetCreatureType() == CREATURE_TYPE_NON_COMBAT_PET)
         return false;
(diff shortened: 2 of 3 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12764,10 +12764,7 @@ Unit* Creature::SelectVictim()
         return NULL;
 
     if (target && _IsTargetAcceptable(target) && CanCreatureAttack(target))
-    {
-        SetInFront(target);
         return target;
```
joschiwald: does it work without trouble? is serverside orientation still updated correctly?

Golrag: Yes, it is handled here: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L180
note: some spells should allow NPC's to walk/attack while channeling (like bone flurry) but this doesn't work (yet)

joschiwald: movementgenerator update is blocked while casting (and channeling), is serverside orientation updated correctly while casting?

Golrag: Good question, will test this asap.. Shouldn't SPELL_ATTR0_CAST_TRACK_TARGET take care of this ?

Golrag: Just tested it with a simple .cast back 11366, and it is not (spell was cast but serverside the npc didn't change his orientation until the cast was done).
But, if a player is tanking an NPC and NPC cast spell on another player, without this change it would have had his serverside orientation looking at the tank.

joschiwald: > But, if a player is tanking an NPC and NPC cast spell on another player, without this change it would have had his serverside orientation looking at the tank.

and exactly this is the issue of #10803

Golrag: Maybe set caster in front of target in spell:update like I did in this PR but with other spells ? I just don't know which spells need to track their focustarget :/
Will take a look at this

src/server/game/Entities/Creature/Creature.h
```diff
@@ -659,7 +659,7 @@ class Creature : public Unit, public GridObject<Creature>, public MapObject
 
         // Handling caster facing during spellcast
         void SetTarget(ObjectGuid guid) override;
-        void FocusTarget(Spell const* focusSpell, WorldObject const* target);
+        void FocusTarget(Spell const* focusSpell, ObjectGuid targetGuid);
```
joschiwald: pls revert this change, or explain why this is needed

Golrag: To allow focus on "nothing" with ObjectGuid::Empty
I can change it back and use ObjectGuid::Empty if target does not exist if you prefer that :)

Warpten:
```c++
To allow focus on "nothing" with ObjectGuid::Empty
```

This doesnt make sense

Golrag: Should I just remove the creatures target then ?

Golrag: Some creatures stop targetting while casting example: http://www.wowhead.com/npc=72762 when casting http://www.wowhead.com/spell=147310/gust-of-wind

joschiwald: the wotlk one is (should be) handled by SPELL_ATTR5_DONT_TURN_DURING_CAST

> Should I just remove the creatures target then ?

don't call FocusTarget if caster == target like it is used in other places

Golrag: If caster == target FocusTarget will be called but with no target now. So _focusSpell will have a value. Will fix that asap, still one question: Should I call FocusTarget when the spell does not have a target ? If I do, when the spell is finished ReleaseFocus will be called and creature will start attack the tank again, If I don't.. ehm he won't I guess, so I will need to look at that :p

joschiwald: > Should I call FocusTarget when the spell does not have a target ?

i don't see any need to focus nothing

Golrag: Updated, instead of calling focusTarget with no target, I removed the creatures target when it has no target or when target is self

Golrag:
PR by @Treeston https://github.com/TrinityCore/TrinityCore/pull/15641

PR #15547, opened 2015-09-19
Core/SmartScripts: added new action SMART_ACTION_SET_POWER_TYPE

r4d1sh:


mthsena:
where it will be applied in the blizzlike?

tkrokli:
According to the branch name Tonghost is using in his repo, this is meant for 3.3.5.
The missing description should be filled in, though. I suppose a TC dev will comment on that later.

Carbenium:
Since every situation where the action could be used is very likely a hack, I'm going to close this PR.
The problems still have to be resolved in the core.
Anyway, thanks for the PR.

PR #15551, opened 2015-09-20
Scripts/Naxxramas: Patchwerk - quick bugfix (incorrect SpellID)

Treeston:
Fixed an incorrect spellid for Naxx10 Hateful Strike. This is from sniffs.
Previously used spellid is actually the one used by Supremus in BT (which had the same mechanic).

### DIFF
--- a/sql/updates/world/2015_09_20_00_world.sql
+++ b/sql/updates/world/2015_09_20_00_world.sql
@@ -1,0 +1,3 @@
+-- remove existing entries
+DELETE FROM `spelldifficulty_dbc` WHERE id IN (28308,41926);
+INSERT INTO `spelldifficulty_dbc` (`id`,`spellid0`,`spellid1`) VALUES (28308,28308,59192);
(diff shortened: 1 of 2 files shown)


ariel-:
Is this confimed by sniffs or something?

http://www.wowhead.com/spell=28308/hateful-strike#comments
> This is from the Old 40 man Naxx

Also the Spell ID is low, suggesting old patch.

Related: http://www.wowhead.com/npc=16028/patchwerk#article
According to this, 41926 and 59192 are the correct spells.

Aokromes:
http://pastebin.com/WTH9zFyW

Killyana:
41926 is used by Supremus at BT
``` sql
ServerToClient: SMSG_SPELL_START (0x08FC) Length: 47 ConnIdx: 2 Time: 10/15/2012 19:09:43.341 Number: 254932
Caster GUID: Full: 0xF1305972000176DB Type: Creature Entry: 22898 Low: 95963
Caster Unit GUID: Full: 0xF1305972000176DB Type: Creature Entry: 22898 Low: 95963
Cast Count: 0
Spell ID: 41926 (41926)
Cast Flags: PendingCast, HasTrajectory (3)
Time: 0
Time2: 0
Target Flags: Unit (2)

joschiwald:
old naxx spells are upgraded to new naxx 10 version

Treeston:
Sniffs say this is the spell used in naxx10, so it is the spell used in naxx10.

Issue #15593, opened 2015-09-24
[Crash] ObjectAccessor::AddCorpse


Pirricli:
I just had this crashlog.

http://pastebin.com/FsE45RXv

a3dbb0dd9caa1af61ff4e313ef1b75a98c17e30b
3.3.5

Aokromes:
No rows, useless crashlog.

Pirricli:
ummm... it's on debug mod like others... I don't know the reasons...

DDuarte:
This might be related to the guid PR by rooty.
The crashlog is not *that* useless since there's only two asserts in `AddCorpse`, still it would be better if the log was more complete.

pete318:
Just for completeness. Here's how it should have looked.

http://pastebin.com/qyw9giDy

shenhuyong:
confirmed

Shauren:
Fixed in cd27fd38b93884797c094a82dc7beb0498e2f124 (TODO: port to 3.3.5)

PR #15604, opened 2015-09-26
(Core/Maps) Remove corpses from ObjectAccessor collection, when loading maps.

r00ty-tc:
If a map is unloaded (instance maps mainly) with corpses inside, when the players return to the instance, if the server wasn't shutdown in between would load the corpses from DB, and attempt to insert again into the map in ObjectAccessor causing an assert when it already existed.

The ensures that before a game map is loaded, any corpses loaded from DB are first removed from the ObjectAccessor map.

### DIFF
--- a/src/server/game/Globals/ObjectAccessor.cpp
+++ b/src/server/game/Globals/ObjectAccessor.cpp
@@ -269,6 +269,19 @@ void ObjectAccessor::RemoveCorpse(Corpse* corpse)
     }
 }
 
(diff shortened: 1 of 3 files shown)


Pirricli:
Maybe it closes #15593

pete318:
Looks like the same crash, from the weird limited crashlog. There may be a different solution to this, but in the meantime this should stop the crash and can easily be reverted for whatever final solution is devised.

jackpoz:
is there a reason to remove the corpse on map load instead of map unload ?

Shauren:
This will be solved in a different way.

PR #15619, opened 2015-09-27
(3.3.5/Core/Auctions) - Changes to Auction House (use of guids)

r00ty-tc:
Changes 
- Stop storing guid for auctioneer on auction records.
- Store instead house ID
- No separate ID for various houses. Only Horde, Alliance and Neutral.
- Mails from AH are now either Blackwater, Horde or Alliance. None from Stormwind, etc.

NOTE: The SQL script will expire all current auctions and return item to player (or
award to high bidder) in order to prepare database for the changes.

This resolves the issue with disappearing auctions since map guid changes.


src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -61,6 +61,21 @@ AuctionHouseObject* AuctionHouseMgr::GetAuctionsMap(uint32 factionTemplateId)
         return &mNeutralAuctions;
 }
 
+AuctionHouseObject* AuctionHouseMgr::GetAuctionsMapByHouseId(uint8 auctionHouseId)
+{
+    if (sWorld->getBoolConfig(CONFIG_ALLOW_TWO_SIDE_INTERACTION_AUCTION))
+        return &mNeutralAuctions;
+
+    if (!auctionHouseId)
```
DDuarte: switch? :)

src/server/game/AuctionHouse/AuctionHouseMgr.h
```diff
@@ -76,11 +76,8 @@ struct AuctionEntry
     uint32 bidder;
     uint32 deposit;                                         //deposit can be calculated only when creating auction
     AuctionHouseEntry const* auctionHouseEntry;             // in AuctionHouse.dbc
-    uint32 factionTemplateId;
 
     // helpers
-    uint32 GetHouseId() const { return auctionHouseEntry->houseId; }
```
DDuarte: Wouldn't it be easier to keep this method and change it to `return houseId;`? As a away to avoid breaking the "API"

pete318:
Test cases run were as follows:
 - Items limited to faction
 - Neutral AH functions correctly.
 - Correct deposit, AH cut taken.
 - Outbid alert correctly sent, only to correct player.
 - Auction sold alert correctly sent, only to correct player.
 - Cancelled auction, correct mail sent.
 - Auction won, correct mail sent.
 - Auction outbid, correct mail sent.
 - Auction sold, correct mail sent.
 - AllowTwoSide.Interaction.Auction - correctly handle existing auctions and new auctions. All above work correctly. NOTE - all auctions use goblin higher deposit/cut. This appears to be the functionality before this change though.
 - AuctionHouseBot. Tested buyer, seller functionality. Noted message in log "AccountMgr::HasPermission: Wrong accountId 0" appears to be due to bot having not having player ID and triggering error with RBAC. Not a problem with this PR. Otherwise functions fine.

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
pete318: Not sure it'd be benficial. The options are CONFIG_ALLOW_TWO_SIDE_INTERACTION_AUCTION, we can't switch on that.

!auctionHouseId, well we can assume that means 0 and we could switch on that.
2 and 6
something else.

So we could have switch and check for 0, 2, 6 and default. But, would it really be worthwhile?

src/server/game/AuctionHouse/AuctionHouseMgr.h
pete318: The way I saw it, GetHouseId() existed before because it needed to actually lookup the value from the entry. Now it's actually already existing as exposed from the same level.

I changed all references to it. So, anyone now doing this knows to use houseId.

So, on this basis I'm disagreeing. But, I don't mind re-instating it. I don't think it makes a lot of difference.

src/server/game/AuctionHouse/AuctionHouseMgr.h
```diff
@@ -64,7 +64,7 @@ enum MailAuctionAnswers
 struct AuctionEntry
 {
     uint32 Id;
-    uint32 auctioneer;                                      // creature low guid
+    uint8 houseId;                                      // creature low guid
```
jackpoz: >// creature low guid

the comment looks outdated

pete318: Yeah, I saw that. But somehow managed to forget to remove it.

PR #15646, opened 2015-10-01
Core/Player: implement ResetCoolDownAfterDuel configurable feature

FrancescoBorzi:
Reset the cooldown of both players after duel. Disabled by default.
- In order to prevent abuse, it does the reset only if the player didn't have any cooldown before starting the duel
- It considers the _arena cooldowns_ only (all cooldowns that last less than 10 minutes)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -7610,6 +7610,12 @@ void Player::DuelComplete(DuelCompleteType type)
     duel->opponent->SetGuidValue(PLAYER_DUEL_ARBITER, ObjectGuid::Empty);
     duel->opponent->SetUInt32Value(PLAYER_DUEL_TEAM, 0);
 
+    if (sWorld->getIntConfig(CONFIG_RESET_COOLDOWN_AFTER_DUEL) == 1)
+    {
+        GetSpellHistory()->ResetAllCooldowns();
+        duel->opponent->GetSpellHistory()->ResetAllCooldowns();
```
joschiwald: profession cooldowns? ^^

ShinDarth:
Updated, tested and working.

digz6666:
:+1:

src/server/game/Entities/Player/Player.h
```diff
@@ -1946,6 +1947,10 @@ class Player : public Unit, public GridObject<Player>
         void SetHonorPoints(uint32 value);
         void SetArenaPoints(uint32 value);
 
+        bool m_hasCoolDownBeforeDuel;
```
Takenbacon: This needs to be initialized in player constructor.

MitchesD:
is that blizzlike or custom?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7610,6 +7610,21 @@ void Player::DuelComplete(DuelCompleteType type)
     duel->opponent->SetGuidValue(PLAYER_DUEL_ARBITER, ObjectGuid::Empty);
     duel->opponent->SetUInt32Value(PLAYER_DUEL_TEAM, 0);
 
+    if (sWorld->getBoolConfig(CONFIG_RESET_COOLDOWN_AFTER_DUEL) &&
+        type != DUEL_INTERRUPTED)
+    {
+        std::string hadCoolDownBeforeDuelText = "You had cooldowns before starting the duel, so your cooldowns haven't been reset.";
```
Shauren: Text needs to be trinity_string'ed for localization

MitchesD:
I wonder if we can move these custom changes somewhere else, to don't make Player class even more bigger, it could make that class a bit smaller if we move custom things to somewhere else and just call them here

joschiwald:
@MitchesD PlayerScripts

ShinDarth:
@Shauren  @Takenbacon done

@MitchesD it's just few lines inside Player.cpp

src/server/game/Entities/Player/Player.h
ShinDarth: done

L30m4nc3r:
@ShinDarth make 2 version.
1) After Duel.
2) Before Duel.
-- You have cooldowns, /duel
-- Duel Timer 3, 2, 1, DUEL_START_COOLDOWN_RESET

For me work this code:
    void OnDuelStart(Player* player, Player* plTarget)// override
    {
        for (uint32 phase = 2; phase <= ULONG_MAX / 2; phase *= 2)
        {
        RemoveArenaSpellCooldowns(true);
        return;
        }
    }
But i have micro-lag after 1 second.

ShinDarth:
@L30m4nc3r that would nullify my check to prevent abuse of this.

Rochet2:
> But i have micro-lag after 1 second.

@L30m4nc3r The for-loop is causing the lag and looks to be unnecessary, unless that is not the whole code.

Keader:
If I use Divine Shield before duel  and use stun(Hammer of Justice) during the duel, the right would not be the hammer of justice have reset cooldown?
The current form, with any cooldown before the duel, no spell has reset cooldown.

(sorry bad english)

ShinDarth:
@Keader yes,  it is that way in order to prevent any abuse, as I wrote in the PR description

PR #15671, opened 2015-10-06
[WIP] Dynamic script hotswapping

Naios:
## [WIP] Hotswap - Dynamic script hotswapping

> "Trinity should consider something like a C++ Engine..."

:white_check_mark: Mission success ...

---

6.x: [Build Status](https://travis-ci.org/Naios/TrinityCore) 3.3.5: [Build Status](https://travis-ci.org/Naios/TrinityCore)

**[The 3.3.5 version is available now!](https://github.com/Naios/TrinityCore/commits/hotswap_335)**

---

output_zw3zhg

---

**Recompiles on script changes:**
savetocompile

---

**Dynamic loaded modules based on source changes:**
preview2

---

**Reloadable command scripts based on file notifier (watches the linked `.dll` for changes):**
reload_commands

---

**Demo video on Youtube:**

[Demo video](https://www.youtube.com/watch?v=y2sRbcCWMH4)

---
### Roadmap
- [x] Readd "missing script counter"
- [x] Auto-Configure compilation mode (Debug/Release)
- [x] Clear outdated script.dll's from the script directory
- [x] Route compiler output to TC logger.
- [x] Detect binary incompatibility (**Partial:** warns about mismatching revision hashes)
- [x] Fix script counter
- [x] Fix install prefix correction in build tree
- [x] Fix reloading of ScriptReloadMgr config
- [x] Auto rerun CMake when sources are added/removed
- [x] Check scheduled scripts
- [x] Distinguish between default scripts and "script" scripts to replace default scripts when core scripts are loaded.
- [x] Make `static std::unique_ptr<std::vector<ChatCommand>> commandTableCache` exchange atomic
- [x] Check script load order (strange warnings) **-> not reproduceable on 3.3.5**
- [x] Readd f809e5e13eac66add49d36
- [x] Fix crash on shutdown
- [x] Handle `Unit::i_disabledAI` (charms)
- [x] Reinject all AI's before calling Reset
- [x] Remove the test stuff
### Summary

This PR is splitted into multiple commits which cover:
- **[:white_check_mark: merged]** **Transformation of the sub libraries to shared libraries** to expose singletons to shared libraries.
  - It's still possible and recommended to link the server through full static linking (~20MB size .exe) except for developing purposes.
  - Shared linking is enabled when `DYNAMIC_LINKING=1` in CMake is set or if a script is dynamic linked.
  - The main disadvantage is that `TRINITY_*_API` macros needs to be added to all public interfaces.
  - The travis build tests the full shared linkage to check for missing `TRINITY_*_API` macros.
  - Singleton `instance()` methods were moved into the .cpp files to prevent duplicated instantiations of singletons on unix.
- **[:clock4: outstanding]** **Splitting the scripts into independent modules**.
  - Makes it possible to select the linkage (`disabled`, `static`, `dynamic`) for each module.
    - Scripts marked as `disabled` are excluded from the build.
    - Scripts marked as `static` are statically linked into the worldserver (instead of the game library which caused circular dependencies)
    - Scripts marked as `dynamic` are linked into seperate shared libraries (".dll's", ".so's") - one for each script subdirectory.
  - **[:white_check_mark: merged]** The scriptloaders were splitted and are smaller now.
  - **[:clock4: outstanding]**: The CMake `SCRIPTS` variable is responsible for the default linkage now.
    - `SCRIPTS="none"`: Disable all scripts
    - `SCRIPTS="static"`: Link all scripts statically (old `SCRIPTS=1` option)
    - `SCRIPTS="dynamic"`: Link all scripts dynamically
    - `SCRIPTS="minimal-static"`: Link the command and spell scripts statically (old `SCRIPTS=0` option)
    - `SCRIPTS="minimal-dynamic"`: Link the command and spell scripts dynamically
- **[:clock4: outstanding]** **Adding the HotSwap main system which reloads script libraries on changes**.
  - The ScriptReloadMgr is responsible for unloading/reloading shared libraries, just copy your .dll's or .so's to the scripts directory or delete it, **the worldserver will load/unload it immediatly**.
  - The reloader is also capable to **invoke a build on filechanges**, just set the `HotSwap.EnableReCompiler` worldserver.conf option.
  - Scripts can strong reference it's module to prevent it from unloading (useful for spells), which means **multiple versions of a script can be loaded at the same time** -> **Support for lazy unloading**.
  - The ScriptMgr was rewritten entirely to support database bound and unbound storage layout for better performance.
### Tutorial

``` sh
# Add my fork to your repository
git remote add naios https://github.com/Naios/TrinityCore.git
git fetch naios
git checkout hotswap
# or "git checkout hotswap_335" for the 3.3.5 version

# Keep your local copy up to date
git fetch naios
git reset --hard naios/hotswap
# or "git reset --hard naios/hotswap_335" for the 3.3.5 version

# Use dynamic linking
# Windows:
# Set the `SCRIPTS` option in the cmake gui to "dynamic" or "minimal-dynamic"

# Unix:
mkdir build
cd build
cmake .. -DUSE_COREPCH=1 -DUSE_SCRIPTPCH=1 -DSCRIPTS="dynamic" -DSERVERS=1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install_dir
#                                           ^^^^^^^^^^^^^^^^^^^^

make -j4
make install
```

It's important to set the `CMAKE_INSTALL_PREFIX` to the location where you want to install the worldserver (even on windows!).
Use the **INSTALL** target in Visual Studio!

To build script projects as modules you need to set it's linking option to `dynamic`.
It's recommended to disable some script modules for testing (because the solution will build faster).
Test sources are located in the latest commit which provides a small testing environment at the gmisland.

Don't forget to check out the `HotSwap.EnableReCompiler` worldserver.conf option to reload a script immediatly on a change.

**Reviews, suggestions, criticism, testing and corrections are wished.**


Gigi1237:
Do you need any help testing this?
I have both a linux and a windows box on which I can test it on. Possibly even a Mac if need be, although I'd rather not have to test on that machine.

Naios:
It would be nice if you could help testing the builds and if there are obviously major bugs which occur (it needs to be tested on all 4 major platforms/compiler combinations, see [the checklist](#roadmap)).
Basically it's enough to set the cmake variable `DYNAMIC_LINKING=1` through `-DDYNAMIC_LINKING=1` for example.
The following log message will show up when `DYNAMIC_LINKING` was set.

screenshot_2

Dynamic linked scripts aren't loaded into the server as of now which means if you mark a script module as `dynamic` it will compile it in a seperate shared library but it won't load the scripts.

Gigi1237:
GCC 4.8 Linux Build Fails:
```
Linking CXX shared library libcommon.so
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: ../../dep/jemalloc/libjemalloc.a(jemalloc.c.o): relocation R_X86_64_32 against `arenas_lock' can not be used when making a shared object; recompile with -fPIC
../../dep/jemalloc/libjemalloc.a: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
src/common/CMakeFiles/common.dir/build.make:933: recipe for target 'src/common/libcommon.so' failed
make[2]: *** [src/common/libcommon.so] Error 1
CMakeFiles/Makefile2:585: recipe for target 'src/common/CMakeFiles/common.dir/all' failed
make[1]: *** [src/common/CMakeFiles/common.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 13%] Built target generate_ScriptPCH.cpp
Makefile:117: recipe for target 'all' failed
make: *** [all] Error 2
```

Naios:
Could you post your `CMAKE_CXX_FLAGS``? (Maybe through adding a CMake message or by taking a look into the `CMakeCache.txt` file).

Gigi1237:
Got none apparently: 
image

Naios:
```cmake
diff --git a/CMakeLists.txt b/CMakeLists.txt
index bb56151..aa98b8d 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,3 +85,5 @@ add_subdirectory(dep)
 
 # add core sources
 add_subdirectory(src)
+
+message("cxx flags: ${CMAKE_CXX_FLAGS}")
```

Gigi1237:
Whoops, sorry. Thought they'd show up in cmake GUI...
    `cxx flags: -std=c++11 -fPIC -fvisibility=hidden`

Naios:
@Gigi1237 i fixed the issue, could you try it again with the latest revision?

Gigi1237:
@Naios Still fails. Different problem though
```
Linking CXX executable bnetserver
c++: error: unrecognized command line option ‘--no-undefined’
src/server/bnetserver/CMakeFiles/bnetserver.dir/build.make:568: recipe for target 'src/server/bnetserver/bnetserver' failed
make[2]: *** [src/server/bnetserver/bnetserver] Error 1
CMakeFiles/Makefile2:1214: recipe for target 'src/server/bnetserver/CMakeFiles/bnetserver.dir/all' failed
```
[Full Build Log](http://paste2.org/O770UhOm)

Naios:
@Gigi1237 removed the flag, gcc has some strange issued with it.

Gigi1237:
@Naios GCC build still fails:
```
Linking CXX executable worldserver
../scripts/libscripts.a(cs_gm.cpp.o): In function `gm_commandscript::HandleGMListIngameCommand(ChatHandler*, char const*)':
cs_gm.cpp:(.text._ZN16gm_commandscript25HandleGMListIngameCommandEP11ChatHandlerPKc[_ZN16gm_commandscript25HandleGMListIngameCommandEP11ChatHandlerPKc]+0x17): undefined reference to `HashMapHolder<Player>::_lock'
cs_gm.cpp:(.text._ZN16gm_commandscript25HandleGMListIngameCommandEP11ChatHandlerPKc[_ZN16gm_commandscript25HandleGMListIngameCommandEP11ChatHandlerPKc]+0x187): undefined reference to `HashMapHolder<Player>::_lock'
cs_gm.cpp:(.text._ZN16gm_commandscript25HandleGMListIngameCommandEP11ChatHandlerPKc[_ZN16gm_commandscript25HandleGMListIngameCommandEP11ChatHandlerPKc]+0x2db): undefined reference to `HashMapHolder<Player>::_lock'
../scripts/libscripts.a(cs_reset.cpp.o): In function `reset_commandscript::HandleResetAllCommand(ChatHandler*, char const*)':
cs_reset.cpp:(.text._ZN19reset_commandscript21HandleResetAllCommandEP11ChatHandlerPKc[_ZN19reset_commandscript21HandleResetAllCommandEP11ChatHandlerPKc]+0xdf): undefined reference to `HashMapHolder<Player>::_lock'
collect2: error: ld returned 1 exit status
src/server/worldserver/CMakeFiles/worldserver.dir/build.make:227: recipe for target 'src/server/worldserver/worldserver' failed
make[2]: *** [src/server/worldserver/worldserver] Error 1
CMakeFiles/Makefile2:1495: recipe for target 'src/server/worldserver/CMakeFiles/worldserver.dir/all' failed
make[1]: *** [src/server/worldserver/CMakeFiles/worldserver.dir/all] Error 2
Makefile:117: recipe for target 'all' failed
```
[Full Build Log](http://paste2.org/DaXE6NNa)

Naios:
@Gigi1237 i spent some time to work on the GCC build but it seems like GCC has some strange issues.
Could you first test the Clang/MSVC build out please?

Gigi1237:
Build tested on:
- [x] MSVC 14: Success
- [ ] MSVC 12
- [ ] Clang 3.4+

Naios:
@Gigi1237 All builds were fixed (MSVC 12 & GCC). Also the startup error was fixed, thanks for the crashlog.The windows core startup works without any issues, could you test it again please?

Naios:
**/Push**

I'm still looking for testers, who can find possible bugs in this commit.

The basic module reloading part, is working now.
Seems like just the AI and corescript support is missing.

preview2

u0yiji:
@Naios  I build success  but how to load scripts_spells.dll?

tkrokli:
@u0yiji : Remember: Always build the 6.x source with
- [x] TOOLS

to get those extra files.

Naios:
Loading of the script modules isn't working as of now, since this part is very complicated... i'm working on it.

Naios:
/Push Dynamic script reloading should work for all script types now.
Some final cleanup and corrections are needed but all in all this PR seems almost completed.

I urgently need tests on this pr and code reviews.

Naios:
Fixed 2 Issues:

- The recompiler project build type (Release/ Debug) now depends on the .dll/.so build type, thanks @Rochet2 for figuring this out.
- Fixed a minor issue regarding to checking before deleting a nullptr, thx @jackpoz

Rochet2:
Just a bunch of midnight comments and thoughts. Havent actually tested any scripts yet.
And just got to beginning of `b/src/server/game/Scripting/ScriptMgr.cpp` in https://github.com/TrinityCore/TrinityCore/compare/6.x...Naios:hotswap.diff with looking through code. Need to continue some other day. Posting this now since gtg.

Is `(void)code` there to avoid compiler warnings about unused variable? Or is it something else? Isnt the code variable clearly used here though.. hmm.
https://github.com/Naios/TrinityCore/blob/b5b555e131611a7158a2d8c35eec6ffd01cb7a8d/src/server/game/Scripting/ScriptReloadMgr.cpp#L374

The "locked" case not handled? Or assumed not to be locked / doesnt matter.
https://github.com/Naios/TrinityCore/blob/91d8966c34ee0481101ae81707ac34d0847dbd4d/src/server/game/Entities/Creature/Creature.cpp#L744-L748
https://github.com/Naios/TrinityCore/blob/91d8966c34ee0481101ae81707ac34d0847dbd4d/src/server/game/Scripting/ScriptMgr.cpp#L341

From what I read it seems as if config reload is not supported (yet) for the config settings.
The enabling and disabling of the dll reload itself might not be that useful to reload, but at least disabling/enabling the auto compile on code save might be a nice thing to have if you want to develop for some without the auto compile. (talking about .reload config with changing the hotswap config settings)

After compiling the dll files are in the bin folder. I think this is pretty nasty, though if the core is installed with INSTALL elsewhere then its fine, but is there anything that could be done about this?
I think one way would be (for me) to just use the config option to redirect loading dlls from the server root instead of the scripts and ignore the install step - or is install forced by the auto compile? If so, then it would mess that up.
Also I noticed the install step doesnt move the debug files from the bin folder root to the scripts folder. Debugging seemed to work fine when I tested so I can only assume the debug files are loaded from server folder root (working directory) and not from where the dll actually is at. .. which is weird I guess. :thought_balloon: 

Is there any documentation going to be written for this (planned)? Comments, doxygen, wiki? Or is current state going to be the final documentation?
I mean documenting the code, not the how to use and add scripts.

Sure is a lot of code to read through and a lot of it is cmake or complex/scattered/unfamiliar so that it is not easy to see (for me) what is going on at points.
Probably takes a while to read, understand and test. A common user isnt going to know how to use this even if they would know how to compile and add code to the old script system.
( even though this is pretty simple - but thats what they always say about everything in CS :smiley: )

Some other scripting related: :thought_balloon: 
The scripting api (hooks) could always be extended. There is no good way to store data per player without locks or a core mod to player class. Script classes cant be merged (multiple inheritance for example) - at least in the old system due to crash on cleanup.

Cant wait to resolve all the merge conflicts with stuff :D

jackpoz:
what happened to https://github.com/TrinityCore/TrinityCore/blob/6.x/src/common/GitRevision.cpp#L75 ?

jackpoz:
"static GuildFinderMgr* instance()" ended up in copyright header
```
diff --git a/src/server/game/Battlefield/BattlefieldMgr.h b/src/server/game/Battlefield/BattlefieldMgr.h
index d370add..034f6bb 100644
--- a/src/server/game/Battlefield/BattlefieldMgr.h
+++ b/src/server/game/Battlefield/BattlefieldMgr.h
@@ -11,7 +11,7 @@
  * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
  * more details.
  *
+ * You should have received a copy of the GNU General Public License alongstatic GuildFinderMgr* instance()
  */
 ```

**Comment by jackpoz** [comment]
no need to NULL-check here
```
+    if (i_AI)
+    {
+        delete i_AI;
+        i_AI = nullptr;
+    }
```

is it intended to ignore the return value of AIM_Destroy() here ? is there a case of leaking memory (even without dynamic script enabled) ?
```
@@ -748,12 +766,12 @@ bool Creature::AIM_Initialize(CreatureAI* ai)
         return false;
     }
 
-    UnitAI* oldAI = i_AI;
+    AIM_Destroy();
```

**Comment by jackpoz** [comment]
what happened to "UnusedScriptNames" ?

**Comment by jackpoz** [comment]
is this a dynamic library limitation or does it affect static library too ?
```
+// Don't change this functions name!
+// Name must match:
+// void Add##${NameOfDirectory}##Scripts()
```

**Comment by jackpoz** [comment]
I see you reverted https://github.com/TrinityCore/TrinityCore/commit/9d6c8add0e46a4700bed1cfcfdda03775ddcc7c6 , please check PCH and NO-PCH builds (I see the matrix in the summary above doesn't specify if they are pch/no-pch/both)

**Comment by Naios** [comment]
Thanks for your feedback :-)

About the `m_ai_locked`... the `AIDestroy` methods will never fail in this context since the AI is locked only at a `Map::Update` call.

I'm still unsure whether the AI lock is needed or even safe to apply... i tracked the change back to https://github.com/mangos/mangos-svn/commit/99fb2dd5ef29698473068c0033819c6be0941ac9 but however i personally would remove it since it will fail only when an AI tries to reinitialize itself which is never the case.

-> Ask @Shauren

**Comment by Naios** [comment]
@Rochet2
Thanks again for your review.

> Is (void)code there to avoid compiler warnings about unused variable? Or is it something else? 
Isnt the code variable clearly used here though.. hmm.
https://github.com/Naios/TrinityCore/blob/b5b555e131611a7158a2d8c35eec6ffd01cb7a8d/src/server/game/Scripting/ScriptReloadMgr.cpp#L374

Yep it's to avoid compiler warnings. Boost filesystems `remove` behaves strange, and throws exceptions when there is no code variable given although it returns `false` on failure.
Don't know what is going on there....
I added a comment to clarify the usage of the voidcast.

> The "locked" case not handled? Or assumed not to be locked / doesnt matter.
https://github.com/Naios/TrinityCore/blob/91d8966c34ee0481101ae81707ac34d0847dbd4d/src/server/game/Entities/Creature/Creature.cpp#L744-L748
https://github.com/Naios/TrinityCore/blob/91d8966c34ee0481101ae81707ac34d0847dbd4d/src/server/game/Scripting/ScriptMgr.cpp#L341

Explained above, it's assumed not to be locked.

> From what I read it seems as if config reload is not supported (yet) for the config settings.
The enabling and disabling of the dll reload itself might not be that useful to reload, but at least disabling/enabling the auto compile on code save might be a nice thing to have if you want to develop for some without the auto compile. (talking about .reload config with changing the hotswap config settings)

Good point which annoyed me too when testing the recompiler, I will fix that soon.

> After compiling the dll files are in the bin folder. ...

/Fixed, good catch there

> Also I noticed the install step doesnt move the debug files from the bin folder root to the scripts folder. Debugging seemed to work fine when I tested so I can only assume the debug files are loaded from server folder root (working directory) and not from where the dll actually is at. .. which is weird I guess. :thought_balloon:

I never build hotswap as debug build... probably needs some testing but i could move the debug symbols with the shared libraries when there is a need to do so.

> Is there any documentation going to be written for this (planned)? Comments, doxygen, wiki? Or is current state going to be the final documentation?
I mean documenting the code, not the how to use and add scripts.

Actually no... i already noticed that my implementation lacks of comments (like TC overall oO),
which makes it even harder to understand the complex metaprogramming in the `ScriptMgr`.
Maybe i could add more comments...

> Script classes cant be merged (multiple inheritance for example) - at least in the old system due to crash on cleanup.

Could you explain more about this? I've never heard about that before.

> Some other scripting related: :thought_balloon: 
The scripting api (hooks) could always be extended. There is no good way to store data per player without locks or a core mod to player class.

It was discussed recently in following issue: https://github.com/TrinityCore/TrinityCore/pull/15732.
I proposed a custom storage in the player class which is threadsafe and could be used to store custom data:

- - -

> Maybe we could add a "custom" store in player which is adaptable but not global.
We could use it to replace several other less used optional core or custom stores,
which would also increase the compression factor for unused entries:

```c++
enum class PlayerExtraUINT32Store
{
    STORE_PLAYER_HEALTH_BEFORE_DUEL,
    STORE_PLAYER_MANA_BEFORE_DUEL
};

Optional<uint32> Player::GetCustomStorageEntry(PlayerExtraUINT32Store)
{
    ...
}

bool Player::ClearCustomStorageEntry(PlayerExtraUINT32Store)
{
    ...
}
```

Naios:
@jackpoz 
Thanks for your feedback

> what happened to https://github.com/TrinityCore/TrinityCore/blob/6.x/src/common/GitRevision.cpp#L75 ?

The compiler flags in the `revision_config.h` were removed in https://github.com/Naios/TrinityCore/commit/9f1f6cf173673c3326cbd7f1cbef008a119c8e75
The reason is that the adapted C & CXX flags are not stored in the cache and therefore not available when generating the revision_config as part of the build process - also clarified in the commit message, which means the C & CXX flag entry depends on the generator -> bugged & unused -> removed.

> "static GuildFinderMgr* instance()" ended up in copyright header

How this could happen... ^^ /fixed

> no need to NULL-check here

Was fixed in my development branch already

> what happened to "UnusedScriptNames" ?

The unused scripts are handled in a different way now: https://github.com/Naios/TrinityCore/commit/9f2bf1abbb6c3d656facbd4be3767034a687b503#diff-3d759b20335cedcebd74c619a5057557R840 where no global variable is needed anymore.

> is this a dynamic library limitation or does it affect static library too ?

This is only needed for the main function of the scriptloader sources `cs_script_loader.cpp`, `spell_script_loader.cpp`... 
In dynamic builds as well as in static builds since the generated internal scriptloader https://github.com/Naios/TrinityCore/blob/hotswap_wip/src/server/scripts/ScriptLoader.cpp.in.cmake needs to know these functions.

> I see you reverted 9d6c8ad , please check PCH and NO-PCH builds (I see the matrix in the summary above doesn't specify if they are pch/no-pch/both)

Yep you are right, we need to test this, added it to the checklist

Yehonal:
Hi 

It's a very nice job!

I was implementing such a similar "modular engine" some time ago but it appeared to be too extreme according with rest of community: https://community.trinitycore.org/topic/10536-trinitycore-as-a-true-mmo-framework 

However in this all-in-one repo you can find some useful cmake code for sources inclusion automatism etc: https://github.com/AzerothShard/wowengine-full

Feel free to contact me if you need an hand with this proj.

Naios:
Thanks

Your framework seems tobe very interesting, i thought about something like this already but as you said it's far too intrusive for users and the source code (it's very difficult to modularize scripts already).

Some time ago I worked with OSGI (the microservice Java framework where Eclipse is built on) and it was very impressive.
It goes a step further and allows coupling through interfaces only meanwhile the implementation is completely hidden to the user.
In C++ it's possible too but it would need some heavy buildsystem and compiler magic to work effectively (LLVM Clang/JIT).

Rochet2:
@Naios 

> > Script classes cant be merged (multiple inheritance for example) - at least in the old system due to crash on cleanup.
>
> Could you explain more about this? I've never heard about that before.

For example this would crash on old system because of multiple script class inheritance:
```c++
struct myscript : CreatureScript, WorldScript
{
    myscript() : CreatureScript("asd"), WorldScript("asd") {}

    // any hooks here
}

void AddSC_myscript()
{
    new myscript();
}
```

The reason it crashes (on shutdown) is because the scriptmgr unloads per script type.
https://github.com/TrinityCore/TrinityCore/blob/8524faae94601621b16002aadb96f6d931d29f64/src/server/game/Scripting/ScriptMgr.cpp#L274-L307
So once the worldscript part calls delete, what happens when creaturescript calls delete? - undefined behavior since the pointer is being deleted twice.

> It was discussed recently in following issue: #15732.
> I proposed a custom storage in the player class which is threadsafe and could be used to store custom data:

Yeah, something like that. In my mind it would be really nice to have a feature that works without an edit to the system though, so its just API.
My ideas were something along the lines of:

```c++
struct Data // this is abstract and can be implemented by user thus containing anything he wants
{
};

// in player class
std::unordered_map<uint32, std::shared_ptr<Data> > player_data;
static std::atomic<uint32> player_dataid = 0;

// my code
static const uint32 mydataid = Player::player_dataid++; // guaranteed to be unique
...
    player->player_data[mydataid] = new MyData(); // store anything made up by me here
...
```

My functions looked horrible so I just used vars directly here.
In terms of stuff being clean this looks horrible. But in terms of dynamically bumping in stuff it would work fine and would cause no leaks or have thread issues when using from map update for example.
boost::any would probably work here too instead of the pointers, though I dont like that myself  .. probably because of unfamiliarity.

jackpoz:
> player->player_data[mydataid] = new MyData(); // store anything made up by me here

just as unrelated note, this is not thread-safe and can actually invalidate all iterators http://www.cplusplus.com/reference/unordered_map/unordered_map/operator[]/ . Better use a fixed-sized array storing Data* with an enum as index that users can customize, with a nice
```
Data* GetCustomData(DataEnum index) { return player_data[static_cast<size_t>(index)]; }
```

it's unclear if the user will have to deal with race conditions on the actual Data based on each best case (atomics, R/W locks, mutexes) . There could be different Data base classes with each different locking mechanism.

Rochet2:
>> player->player_data[mydataid] = new MyData(); // store anything made up by me here
>
> just as unrelated note, this is not thread-safe and can actually invalidate all iterators

That would only be unsafe if someone stored an iterator over time and that would be unsafe in all non fixed size storages since the element could be destroyed. This storage was not supposed to be used from multiple threads at the same time, but using it from map update thread is safe which is the point of it. (global storage is not safe to access from map threads)
The API should not allow get any iterators to the player_data container. It is just the internal storage type. The API would simply offer to place, access and erase data from the storage.

I guess this is a bit off topic, but .. if there should be no race conditions so why would the data need any locking mechanism (for data or container holding the data)? You can only operate on a player safely in general at all only from the map he is on or the world thread, no? So any other action is already not safe so why would such need to be implemented now?
Bracing for stored pointers to the data is same as bracing for users storing player and creature pointers.

Shauren:
Custom data storage is not in the scope of this PR

Rochet2:
MSVC 2015 non pch is not building:
[log.txt](https://github.com/TrinityCore/TrinityCore/files/49715/log.txt)

jackpoz:
what's the impact of all these changes to existing custom scripts/patches ? could you simulate what kind of conflicts a user would have to figure, for example https://community.trinitycore.org/topic/10-patch-telenpc2-pvpranks-passive-anticheat-for-trinitycore-2015-11-03-ea9c112b3085-35a-only/?do=findComment&comment=15 ?

Naios:
The impact i very minimal, there are only two incompatibilities:
- To register a script (`new AnticheatScripts();`) we need to call `ScriptMgr::SetScriptContext(std::string const& context)` and after we registered all scripts we need to call `ScriptMgr::SwapScriptContext()` other usage is guarded by ASSERT's for improved error tracing.
That's why all core internal scripts are initialized in `ScriptMgr::Initialize` now: https://github.com/Naios/TrinityCore/blob/hotswap_wip/src/server/game/Scripting/ScriptMgr.cpp#L818
- Classes/functions which are accessible from scripts need a TRINITY_*_API macros now.

Naios:
**/Push** The non PCH build was fixed

Rochet2:
Tried building the source from naios/hotswap and got this log:
https://gist.github.com/Rochet2/c1241372d7cd50d4e74f
2ca79bab270aa4ac7b5d6459b595df8169b92fce

Naios:
Strange, the travis build should fail when there are unexported symbols... I will have a look at it.

Naios:
**/Fixed and rebased**, thanks for reporting.
Seems like MSVC doesn't export methods of templated classes although the class was instantiated.

Naios:
Changes:
* Old script modules are uninstalled when they are exported from the static library already.
* The script reloader config is reloadable now.

Naios:
**/Push** I did a *Happy new Year* rebase and updated this PR to to latest HEAD.
All in all this PR is finished but it is impossible to improve it further due to missing user testing.

I put this PR on hold as long as there is no interest in testing and reviewing this PR.

Asterc:
Any chance to make it available to the 3.3.5 branch? It would be more likely to get testers

Naios:
I planned to port this to 3.3.5 as soon as this PR was taken into the 6.x branch because it's a huge amout of work to port it over... (especially the TRINITY_*_API commits).
However i would appreciate it if someone starts to port the commits to 3.3.5 since i currently don't have much time.

LuigiElleBalotta:
News on this?

Naios:
I will port this over to 3.3.5 for testing as soon as my examen phase is finished.
All in all the system is completed but it needs to be tested carefully (hoping this will be done on the 3.3.5 branch since there are no 6.x specific bugs).

lorac2k14:
Naios with the database handling change does this mean you have decided to put this idea to the pastures? as the idea is great In my humble opinion
for many reasons it does open TC up to be a multi platform server
not to mention the ability to use TC for more than just wow if one wanted to
Not to mention it pulls a lot of weight out of the exes and puts it on the dlls

Naios:
No I'm still working on getting this merged into the codebase.
Currently I'm merging this PR partially in small steps into 3.3.5 and 6.x see #16795 and the recent changes to singletons.

The next step will be to enable shared compilation for the libraries through merging the `Core/*: Converted the * library to a shared library` commits, which will show very early if there are any issues with the compilation model. -> the scripting system will stay the same but the server will be linkable to shared libraries.

lorac2k14:
Thank you Naios for your quick reply I think your idea is brilliant and I'm reviewing so I can actually understand how you did this so maybe I can provide a hand porting over to current 6x / 3.3.5
Ah yes see that now sorry I saw the 6698586a97c46398897832889036dffd28e26971 on top commit. I forgot it merges it last for compatibility. I also saw the incubation in title which threw me.
 I had a blond moment.
I truly love what you have done here. I will review and start trying to port to 3.3.5a and see how it goes if successful would you like me to open a pr on your git?

Naios:
I highly appreciate your offer but some things changed with the recent commits to the 3.3.5 branch and  I plan to solve some issues differently which makes it very difficulty to port the current PR over to 3.3.5.

I will let you know when the issues are solved and I could need a helping hand :-)

lorac2k14:
no problem :) just give me a yell if you need me

Naios:
@lorac2k14 I could need help porting the PR #16810 over to 3.3.5 . If you are still interested it would be amazing if you could PR it to my fork.

lorac2k14:
Sure thing Naios  :)
I just completed porting 
hash: ea8f2d90d10d357973261ea21ce279aaa7d3fcee 
I need to call it a night its almost 5am here 
Ill get back on it as soon as i get up
You only want me to fire a PR to you once its compiled right or you want it at current stage?

Naios:
Thanks for your effort.
You don't need to create a PR for it, I'll just take a look at your fork.

lorac2k14:
Tabs replaced sorry for newb mistake I should have checked editor
you are very welcome

Naios:
The 3.3.5 version of the hotswap system is **[available for testing now!](https://github.com/Naios/TrinityCore/commits/hotswap_335)**
The description above was updated to reflect the latest changes, also instructions were added how to test the 3.3.5 port.

Naios:
Just a quick note about the current state of development:
In the last 2 weeks I implemented some missing features like (auto cmake rerun when files are added or removed) and fixed various issues:

- The script reloader is fully asynchronous, atomic and nonblocking now -> doesn't affect the difftime anymore (through blocking while waiting for external processes)
- Fixed multiple crashes caused by scripts in instances when the script library is removed (it's possible to remove the library while being infight now -> thanks to lazy unloading and the worldserver which inject sdefault AI's then).
- Major code improvements, and minor improvements

All in all this PR seems to be very stable for me which means **I finally plan to merge it on monday into both branches**..

Let me know when you find any bugs meanwhile.

Dekadencee:
@Naios Very nice system to fast develop of scripts, are you planning to port this to 4.3.4?

DDuarte:
@Dekadencee: 4.3.4 is not a branch of TrinityCore

Dekadencee:
@DDuarte Yes, I know that its not a oficial branch of TrinityCore but still develop on https://github.com/TrinityCoreLegacy/TrinityCore

DDuarte:
That isn't TC. That's someone's private playground.

Naios:
@Dekadencee I'm glad to hear that you like the system :-)
As far as I know @Aokromes tried to port it to one of the unofficial 4.3.4 branches, but I don't know the progress of the port.

Issue #15672, opened 2015-10-07
[3.3.5] Innkeeper Boorand Plainswind (3934) Has Two Browse Goods Gossip Options


Espionage724:
Build: https://github.com/TrinityCore/TrinityCore/commit/cdb16c5181b0ad02e5dd3a4afa44151fb119b62e

DB GUID: 14139

The Innkeeper at The Crossroads has two "Let me browse your goods." options that both seem to do the same thing.

![wowscrnshot_100615_231935](https://cloud.githubusercontent.com/assets/6378956/10336583/bb06a244-6cc7-11e5-80e2-7fc0361c820f.jpg)

joschiwald:
from massive sniff dump:
```sql
INSERT INTO `gossip_menu_option` (`menu_id`, `id`, `option_icon`, `option_text`, `box_coded`, `box_money`, `box_text`) VALUES
(344, 2, 5, 'Make this inn your home.', 0, 0, ''), -- 6746
(344, 3, 1, 'Let me browse your goods.', 0, 0, ''), -- 6746
```

tkrokli:
TrinityCore rev. 30eaf23b6af4 2015-10-08 02:37:55 +0200 (3.3.5 branch) (Win64, Release)
Issue confirmed. "Let me browse your goods." appears twice in the NPC's gossip menu.

@joschiwald : although I'm not sure where you are going with that SQL, seeing that the number of columns to be inserted does not match the number of actual columns in the query (and hence would not work as an actual SQL insert) and that 6746 is a match for Innkeeper Pala in Thunder Bluff, I found that Innkeeper Boorand Plainswind <Innkeeper> (NPC 3934) is the only one with 2 matching entries in `gossip_menu_option` for `option_text` = 'Let me browse your goods.' :
```sql
SELECT * FROM `gossip_menu_option` WHERE `option_text` = 'Let me browse your goods.';
```
```
menu_id      id  option_icon  option_text                OptionBroadcastTextID  option_id  npc_option_npcflag  action_menu_id  action_poi_id  box_coded  box_money  box_text  BoxBroadcastTextID  
-------  ------  -----------  -------------------------  ---------------------  ---------  ------------------  --------------  -------------  ---------  ---------  --------  --------------------
    342       3            1  Let me browse your goods.                   8097          3                 128               0              0          0          0                               0
    344       1            1  Let me browse your goods.                   8097          3                 128               0              0          0          0                               0
[...]

PR #15673, opened 2015-10-07
Fix for #15672

S0ulSeller:


DDuarte:
Thanks for the PR. However do not submit PRs with SQL only fixes. Write down this fix in the issue #15672 and I'll tag it as Patch-Fix.

DanielMo1603:
Sorry, it was my first Pr :)

joschiwald:
the gossip option is present in sniffs, the other options (0-2) are maybe wrong

Issue #15643, opened 2015-10-01
[335] Auctionhouse does not calculate correct deposit for multiple stacks


MrSmite:
__Core:__ 747ea7ca2c2b757dc9c5a672746d801a8365ee98
__DB:__ TDB_full_335.59_2015_07_14 + updates

__Bug:__

The deposit required to create an auction of multiple stacks (same item) is not calculated and displayed correctly in the auctionhouse GUI, resulting in the possibility of a "not enough money" error.

__Details:__

* Give yourself 18x 4496 (Small Brown Pouch)
* Give yourself 5 silver, 22 copper (hint: this is what the auctionhouse will calculate)
* Open the auctionhouse to create a new auction
 * Put the bag in the item slot of the auction GUI
 * Stack size: 1
 * Number of stacks: 14
 * Price: Per item
 * Starting bid: 2 gold, 25 silver, 0 copper
 * Buyout: 2 gold, 25 silver, 0 copper
 * Duration: 24 hours
* Notice the deposit shown is 5 silver, 22 copper
* Click the "Create auction" button
* Result
 * 4 bags will be placed into auction
 * An error will appear: "You do not have enough money"
 * Your character will still have 1 silver, 22 copper

tkrokli:
TrinityCore rev. 1dcf186dfd24 2015-09-30 22:31:00 +0200 (3.3.5 branch) (Win64, Release) / TDB 335.59
Confirmed. In addition to showing the wrong price (independent of the bid price or buyout price), the AH also shows that it wants 6c extra (36/6) for each fourth item (4) you place in the Create Auction GUI.

Average deposit is supposedly is 36 copper / item, but AH won't accept any lower deposit than 100c (1s). The big deal, though, is that the actual deposit price is 1 silver/item, a total of 1400c (14s) in your case.
```
 bags:  Deposit shown:   Actual deposit:
 -----  --------------  ----------------
  1              100           100
  2              100           200
  3              108           300
  4              150           400
  5              186           500
  6              222           600
  7              258           700
  8              300           800
  9              336           900
 10              372          1000
 11              408          1100
 12              450          1200
 13              486          1300
 14              522          1400
 15              558          1500
 16              600          1600
 17              636          1700
 18              672          1800
 19              708          1900
 20              750          2000
```
Is the GUI calculating the AH cut price instead of the Deposit cost?

pete318:
No, looks much simpler than that. It looks like it's applying the minimum deposit per item.

[...]

PR #15674, opened 2015-10-07
[Core/Auctionhouse] Improved Auction house deposit handling.

r00ty-tc:
Opening this as a PR to get some comments. Since, while I think the code works and is safe. I think it adds some extra processing (deferring deposit handling, while auctions are ongoing) for what is quite a minor problem.

So I'd like opinions as to whether this is a worthwhile endeavour.

Related to https://github.com/TrinityCore/TrinityCore/issues/15643

Commit text:

Improved Auction house deposit handling.

Including the weird quirks.

Changes:
- Deposit calculated according to same rule as client. (detailed below). Should always match client deposit now
- Deposit for multiple auctions now only collected once. Deferred calculation of deposit until all auctions listed.

Deposit calculation is as follows:

Base deposit calculation = MSV x 15/75% (depending on AH). However this is not rounded. Case to int is used (so always round down)
The remainder is held in a float.
The base is then multiplied by number of items, and the time multiplier (x2 for 24 hour, x4 for 48 hour)
The nearest (no of items or lower) no of items when multiplied by the remainder that creates a whole number is then multiplied by the time multiplier (x1/x2/x4) and then added to deposit.

Example:

Item sell price 1s25. Deposit 18.75c (15% of 1s25). So base deposit = 18, remainder 0.75. Time 24h (x2) Item count 1 = 36c. Remainder = 0.75 (n) Min = 1s. Deposit 1s
Item sell price 1s25. Deposit 18.75c (15% of 1s25). So base deposit = 18, remainder 0.75. Time 24h (x2) Item count 2 = 72c. Remainder = 1.50 (n)  Min = 1s. Deposit 1s
Item sell price 1s25. Deposit 18.75c (15% of 1s25). So base deposit = 18, remainder 0.75. Time 24h (x2) Item count 3 = 108c. Remainder = 2.25 (n) Min = 1s. Deposit 1s08c
Item sell price 1s25. Deposit 18.75c (15% of 1s25). So base deposit = 18, remainder 0.75. Time 24h (x2) Item count 4 = 144c. Remainder = 3.00 (n) Min = 1s. Deposit 1s50c (144c + (3c \* 2))

etc. Yes, it's weird. But it's how the client calculates deposits, and so the server does too.

I don't like this personally, but Blizzard must have some similar kludge
or handle auctions entirely differently. Since, the packet structure would
appear to support not only multiple stacks of items, but multiple
different items too.

However, packet structure might support it. The client still sends only a
stack of items at a time, including if items that don't stack. With
nothing clearly identifying that more packets follow, it's hard to track.

### DIFF
--- a/src/server/game/AuctionHouse/AuctionHouseMgr.cpp
+++ b/src/server/game/AuctionHouse/AuctionHouseMgr.cpp
@@ -83,12 +83,23 @@ uint32 AuctionHouseMgr::GetAuctionDeposit(AuctionHouseEntry const* entry, uint32
 
     float multiplier = CalculatePct(float(entry->depositPercent), 3);
     uint32 timeHr = (((time / 60) / 60) / 12);
-    uint32 deposit = uint32(((multiplier * MSV * count / 3) * timeHr * 3) * sWorld->getRate(RATE_AUCTION_DEPOSIT));
+    uint32 deposit = uint32(MSV * multiplier * sWorld->getRate(RATE_AUCTION_DEPOSIT));
+    float remainderbase = float(MSV * multiplier * sWorld->getRate(RATE_AUCTION_DEPOSIT)) - deposit;
+
+    deposit *= timeHr * count;
+
+    int i = count;
+    while (i > 0 && (remainderbase * i) != uint32(remainderbase * i))
+        i--;
+
+    if (i)
+        deposit += remainderbase * i * timeHr;
 
     TC_LOG_DEBUG("auctionHouse", "MSV:        %u", MSV);
     TC_LOG_DEBUG("auctionHouse", "Items:      %u", count);
     TC_LOG_DEBUG("auctionHouse", "Multiplier: %f", multiplier);
     TC_LOG_DEBUG("auctionHouse", "Deposit:    %u", deposit);
+    TC_LOG_DEBUG("auctionHouse", "Deposit rm: %f", remainderbase * count);
 
     if (deposit < AH_MINIMUM_DEPOSIT * sWorld->getRate(RATE_AUCTION_DEPOSIT))
         return AH_MINIMUM_DEPOSIT * sWorld->getRate(RATE_AUCTION_DEPOSIT);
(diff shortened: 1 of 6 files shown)


src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -389,6 +400,118 @@ bool AuctionHouseMgr::RemoveAItem(uint32 id, bool deleteItem)
...
+        pendingAuctionMap[player->GetGUID()] = thisAH;
+        pendingAuctionCountMap[player->GetGUID()] = 0;
+    }
+    thisAH->push_back(aEntry);
+}
+
+uint32 AuctionHouseMgr::PendingAuctionCount(Player* player)
+{
+    if (pendingAuctionMap.find(player->GetGUID()) != pendingAuctionMap.end())
+        return pendingAuctionMap.find(player->GetGUID())->second->size();
```
Naios: You could use

```
auto const itr = pendingAuctionMap.find(player->GetGUID());
if (itr != pendingAuctionMap.end())
        return itr->second->size();
```
to reduce the map lookups.

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -389,6 +400,119 @@ bool AuctionHouseMgr::RemoveAItem(uint32 id, bool deleteItem)
...
+
+    CharacterDatabase.CommitTransaction(trans);
+    pendingAuctionMap.erase(player->GetGUID());
+    delete thisAH;
+    player->ModifyMoney(-int32(totaldeposit));
+}
+
+void AuctionHouseMgr::UpdatePendingAuctions()
+{
+    if (pendingAuctionMap.size() == 0)
```
DDuarte: this check is not needed (redundant) since it the for won't do any loop if the map is empty

Shauren: also do empty checks with .empty()

pete318:
There's been little comment. Either to the overall relevancy, or to the content. So, if there's no reason stated here not to. I intend to merge tomorrow.

jackpoz:
@pete318 what happened to this pr ?

pete318:
I've been somewhat busy, work, university and upcoming holiday travel too.

Later on Saturday I might have a bit of time. If so I'll look to merging this.

PR #15677, opened 2015-10-08
Core/Player: implemented Heirloom Collection

MitchesD:
- implemented all things related to that collection
- you can create heirloom by click on item in collection
- also you can upgrade heirlooms by using specific kits
- old heirlooms are converted to new ones

I am creating it as PR, because I may overlooked something, I had to change it many times, so, if someone can check it, I will be pleased


MitchesD:
no respond? I know about that travis error, will fix that when it will be ready to merge, from my side it's ready in all sides, so any comments here? otherwise I will merge it

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -5960,3 +5981,14 @@ void Spell::EffectUncageBattlePet(SpellEffIndex /*effIndex*/)
     plr->DestroyItem(m_CastItem->GetBagSlot(), m_CastItem->GetSlot(), true);
     m_CastItem = nullptr;
 }
+
+void Spell::EffectUpgradeHeirloom(SpellEffIndex /*effIndex*/)
+{
+    if (effectHandleMode != SPELL_EFFECT_HANDLE_HIT)
+        return;
+
+    Player* player = m_caster->ToPlayer();
```
joschiwald: missing nullptr check

tkrokli:
Sorry for not having tested this or trying to help with this. I'm just not very efficient with my time at testing other things than compiling and starting up the servers.

src/server/game/Entities/Player/CollectionMgr.cpp
```diff
@@ -76,3 +76,209 @@ void CollectionMgr::ToySetFavorite(uint32 itemId, bool favorite)
...
+}
+
+void CollectionMgr::CheckHeirloomUpgrades(Item* item)
+{
+    Player* player = _owner->GetPlayer();
+    if (!player)
+        return;
+    
+    // Convert old heirloom to new ones
+    if (HeirloomEntry const* heirloom = sDB2Manager.GetHeirloomByOldItem(item->GetEntry()))
```
Shauren: This conversion should be done on database (should be a simple matter of updating entry).
This method is called in _LoadInventory, BEFORE bagslot and slot are set on the item - the returned position for DestroyItem is always bag 255 (main inventory, including equipped slots) slot 0 which is the currently equipped helm!

MitchesD:
https://github.com/TrinityCore/TrinityCore/commit/9fe29de260b704603f40a8912c66dccf76fd0db8

Issue #14700, opened 2015-05-15
[3.3.5] Core/Scripts: Death's Challenge & The Endless Hunger


tkrokli:
TrinityCore rev. 5bd7c29 2015-05-13 22:29:49 +0200 (3.3.5 branch) (Win64, Release)

Quest: Death's Challenge http://www.wowhead.com/quest=12733/deaths-challenge (ID: 12733)
```
Olrun the Battlecaller has ordered you to win 5 duels against other death knights.
You may choose to battle death knight initiates or players.
```
- Death Knights defeated in a duel (5)

Actually 2 issues here, 1 practical (game mechanics) and 1 cosmetic :
- According to the quest description, it should be possible to choose between challenging the NPCs named "Death Knight Initiate" and challenging other players. I have tested with 3 simultaneous accounts, two normal (gmlevel 0) accounts and one gmlevel 3 account, even tried using a Night Elf vs Blood Elf with PvP on, but it appears impossible to challenge each other the same way as the NPCs. The player character names still appear in blue color, not green.

![wowscrnshot_051515_074200](https://cloud.githubusercontent.com/assets/4903600/7647763/8ccf8fe4-fada-11e4-9bbf-71353630e187.jpg)

- When my player character challenges a *Death Knight Initiate*, there are 8 (9) lines of text to choose from in the database, but 2 of them use the **$n** with the intention of showing the target name in the /Say text scripted in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp#L497 :

![wowscrnshot_051515_060657](https://cloud.githubusercontent.com/assets/4903600/7647753/76922c6e-fada-11e4-97cc-afbfb994a1aa.jpg)

```cpp
sCreatureTextMgr->SendChat(creature, SAY_DUEL, NULL, CHAT_MSG_ADDON, LANG_ADDON, TEXT_RANGE_NORMAL, 0, TEAM_OTHER, false, player);
```
The text is pulled from the database table `creature_text` and looks like this:
```sql
(28406,0,0,'No potions!',12,0,100,0,0,0,29265,0,'dk initiate SAY_DUEL_A'),
(28406,0,1,'Remember this day, $n, for it is the day that you will be thoroughly owned.',12,0,100,0,0,0,29270,0,'dk initiate SAY_DUEL_B'),
[...]

PR #15678, opened 2015-10-08
Core/Scripts: The Endless Hunger, NPC text

ghost:
Solves an issue where the untargeted variable $r appears in the creature say line
for quest ID 12848, The Endless Hunger. The NPC script `npc_unworthy_initiate`
is supposed to target the player and refer to the player character's race.

Matching the unworthy initiate IDs in the script to the text entries in
`creature_text` 29519, 29520, 29565, 29566 and 29567, the affected entries are:
- "You are hopelessly outmatched, $r." (`BroadcastTextId` 30212)
- "Sate your hunger on cold steel, $r!" (`BroadcastTextId` 30214)

Updates issue #14700

PR #15680, opened 2015-10-09
Core/Scripts: The Endless Hunger & Death's Challenge, creature text

ghost:
This PR solves the following say text issues:
In quest ID 12733, Death's Challenge, the creature script npc_death_knight_initiate is supposed to target the player and refer to the player character's name in these 2 lines from `creature_text`.`entry` 28406:
- "You don't stand a chance, $n." (`BroadcastTextId` 29267)
- "Remember this day, $n, for it is the day that you will be thoroughly owned." (29270)

In quest ID 12848, The Endless Hunger, the creature script npc_unworthy_initiate is supposed to target the player and refer to the player character's race in the following 2 lines from `creature_text`.`entry` 29519, 29520, 29565, 29566 and 29567:
- "You are hopelessly outmatched, $r."  (`BroadcastTextId` 30212)
- "Sate your hunger on cold steel, $r!" (`BroadcastTextId` 30214)

The untargeted variables $r and $n appear in the say lines because the existing script assigns the text to the PC instead of the NPCs. It fails to produce any $race or $name information in /say or chat.

This Pull Request closes issue #14700 and replaces PR  #15678.

### DIFF
--- a/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
+++ b/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
@@ -93,11 +93,6 @@ class npc_unworthy_initiate : public CreatureScript
 public:
     npc_unworthy_initiate() : CreatureScript("npc_unworthy_initiate") { }
 
-    CreatureAI* GetAI(Creature* creature) const override
-    {
-        return new npc_unworthy_initiateAI(creature);
-    }
-
     struct npc_unworthy_initiateAI : public ScriptedAI
     {
         npc_unworthy_initiateAI(Creature* creature) : ScriptedAI(creature)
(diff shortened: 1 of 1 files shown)


MitchesD:
have you tested it?

tkrokli:
Yes, indeed I have tested it, in at least 2 stages. I would not want to create the PR without testing the result first, especially because I want the in-game result to be as blizzlike as possible.

I can verify that the NPC/creature 'say' text now works as intended, both as I remember it from playing the death knight class back in 2010 and also from quite a few YouTube videos I looked up in case I would need to document what it used to be like on the Blizzard servers. The NPC creatures now say the text instead of the player character, as well as solving the issue of `$n` and `$r` appearing in the text,
instead of naming the actual target or referencing the target's race.

PR #15687, opened 2015-10-10
[Need Help - WIP] Core/Battleground allow to join both RBG and BG config

FrancescoBorzi:
## PR Description:

By default the player is allowed to queue for 2 specific BGs at the same time (example: Warsong Gulch and Arathi Basin at the same time) and can jump from one to another, or to queue for RBG only.

The player is **not** allowed to queue for both RBG and a specific BG (example: RBG and Warsong Gulch at the same time).

This PR aims to introduce a configurable setting (disabled by default) to allow the player to queue for 2 BGs even if one of those 2 is RBG.
## PR Current behavior:

You are in a specific BG and queued for RBG.  When RBG starts and you try to jump to it:
- it **works** fine if the RBG type is **different** than the type of the BG which you were in.
- it does **not** work if the RBG type is the **same** of the BG which you were in. 
- the vice versa has the same results (trying to start with RBG first and then jumping to specific BG works only if RBG type != specific BG type).

Example: if you joined specific Arathi Basin and queued for RBG, when RBG starts and you try to jump to it, it will work **only** if the RBG will not be Arathi Basin too. (And vice versa).

Any suggestions?


Aokromes:
What's needed to complete this?

ShinDarth:
@Aokromes I've explained everything in the PR description. It doesn't work if the RBG type is the same of the BG which you were in (or vice versa). It works in all other cases.

I guess that is an easy fix, but I don't know where to look...

jackpoz:
What's the difference of queuing for random plus 2 specific ? doesn't random include all bg types already ?

ShinDarth:
By default you can't queue both for random and for a specific

jackpoz:
> doesn't random include all bg types already ?

Doesn't queueing for random queues you already to all possible BGs ? So if I queue for random, it's like if I queue for WSG, AB, EOTS, etc

ShinDarth:
no, if you queue for random it is NOT like you queue for WSG, AB, etc...

suppose some players queue for specific WSG and some others for RBG, then even if the RBG will be WSG, those players will be in separated instances. A nice idea would be to merge those queues, but that goes behind the scope of this PR.

jackpoz:
lol, the current queue system makes no sense then. is this some weird blizzlike behavior or just plain broken BG queue system ?

Aokromes:
Blizzlike.

jackpoz:
is it blizzlike on 6.x too ?

ShinDarth:
anyway, this PR is "almost" ready and it's a shame it can't be merged. I'm pretty sure that the only problem left it is a really easy to solve, I didn't solve it because I had no idea where that specific piece of logic is, and now I don't have nor the time nor a ready development environment to do it.

jackpoz:
From a player point of view this PR makes no sense, in fact I would have expected specific BG queue to let me join BGs with players queued as random BG. It would make more sense to implement mixed queues (and maybe add an optional config to disable them for unknown reasons I can't think of).

Aokromes:
Some people like to play smalls bgs while waiting for random bgs. Also, if this pr can be increased  to allow LFG queue while on random bg queue can be amazing.

Shauren:
@jackpoz I think you are misunderstanding something here - players that queued random and players that queued specific are not in separate selection pools. Players can be queued for up to 2 battlegrounds at the same time ("random" is just another selection on battleground list) with the restriction that if "random" was selected, it must be the only selection (no specific) and this PR aims to lift that restriction. 
The problem here is that TC does not allow teleporting from one instance of map X to another instance of map X.

jackpoz:
@Shauren so https://github.com/TrinityCore/TrinityCore/pull/15687#issuecomment-226725721 is not correct ?
>suppose some players queue for specific WSG and some others for RBG, then even if the RBG will be WSG, those players will be in separated instances

Shauren:
eh i dunno actually, but what i said about teleports is still true and blocks this PR

ShinDarth:
> The problem here is that TC does not allow teleporting from one instance of map X to another instance of map X.

well, we are now a step further since we know where the problem is

Aokromes:

Shauren:
Seems like the interest was lost on top of the author not really knowing how to solve the teleportation issue

PR #15688, opened 2015-10-10
Line break in bnetserver.conf.dist for consistency

Espionage724:
- Same as https://github.com/TrinityCore/TrinityCore/pull/15628


tkrokli:
If you wonder why the build fails, it has got nothing to do with this PR. It is caused by one of TrinityCore's own updates to the 6.x auth database, forgetting to update the `updates` table at the same time as pushing the incomplete commit.

PR #15692, opened 2015-10-10
Scripts/Commands: enhanced .instance get/setbossstate commands

FrancescoBorzi:
This allows the GM to see the name of the Encounter States without the need of checking the source code.


Treeston:
While you're at it, have you considered checking if you can somehow get it to read encounter names as well?

ShinDarth:
Yes @Treeston I would like to work on that too in future

tkrokli:
Your SQL filename /sql/updates/world/2015_10_10_03_world.sql is taken by commit 1f02e39c3ebd36f07ac85f508fe1cfd66dbc0028 already.

Nyeriah:
This is cool. The command initially had this but I scrapped it off when merging because I thought it to be redundant back then, seeing that to set states you should know what the values represent in the first hand.

ShinDarth:
@tkrokli Fixed, thanks.
@Nyeriah merge?

PR #15697, opened 2015-10-11
Core/Player rewritten duel CD reset system

FrancescoBorzi:
This rewrites the duel CD reset system introduced in https://github.com/TrinityCore/TrinityCore/pull/15646
- Fixed issue caused by some spells such as Stealth and Prowl, which are often cast before duel starts and broke the reset
- Now all CD are saved before the duel, reset while the duel is running, and then restored when the duel is finished.
- Tested and working

Thanks @pete318 for help.


src/server/game/Spells/SpellHistory.h
```diff
@@ -117,7 +117,8 @@ class SpellHistory
     void BuildCooldownPacket(WorldPacket& data, uint8 flags, uint32 spellId, uint32 cooldown) const;
 
     CooldownStorageType::size_type GetCooldownsSizeForPacket() const { return _spellCooldowns.size(); }
-    uint16 GetArenaCooldownsSize();
+    void SaveCoolDownStateBeforeDuel();
+    void RestoreCoolDownStateAfterDuel(Player* player);
```
joschiwald: useless parameter

src/server/game/Spells/SpellHistory.cpp
```diff
@@ -637,21 +637,38 @@ void SpellHistory::BuildCooldownPacket(WorldPacket& data, uint8 flags, PacketCoo
...
-    uint16 count = 0;
+    _spellCooldownsBeforeDuel = _spellCooldowns;
+}
 
-    for (auto itr = _spellCooldowns.begin(); itr != _spellCooldowns.end();)
-    {
-        SpellInfo const* spellInfo = sSpellMgr->EnsureSpellInfo(itr->first);
+void SpellHistory::RestoreCoolDownStateAfterDuel(Player* player)
+{
+    _spellCooldowns = _spellCooldownsBeforeDuel;
```
joschiwald: what happen if you start a profession cooldown during duell?

Pirricli:
Is the life and mana restored too?

src/server/game/Spells/SpellHistory.h
ShinDarth: Player* player is used inside RestoreCoolDownStateAfterDuel method

ShinDarth:
@Pirricli no it isn't, I can consider the idea but not in this PR

src/server/game/Spells/SpellHistory.h
joschiwald: use owner of the spellhistory instead

src/server/game/Spells/SpellHistory.cpp
ShinDarth: just tested and the CD remains, even if you relog

ShinDarth:
@joschiwald 

- Tested profession CD and it's not abusable (profession CD remains even if you relog, tested both beofore duel and while duelling)

-  Remove useless player parameter

Nyeriah:
This is very unrelated but cooldown != cool down

http://www.thefreedictionary.com/cool+down
http://www.thefreedictionary.com/cooldown

Just seen you naming things "CoolDown" so was wondering

tkrokli:
Hehe, nice to see that I'm not the only one with a penchant for correct grammar and spelling. I suspect that it is because of the very popular initialized form "CD", even though it is not a proper acronym.

ShinDarth:
@Nyeriah  @tkrokli thanks I didn't know it XD

PR #15731, opened 2015-10-18
Convert the Duel Cooldown Reset feature to scripts

FrancescoBorzi:


jackpoz:
no changes in https://github.com/ShinDarth/TrinityCore/blob/toscri/src/server/game/Scripting/ScriptLoader.h ?

ShinDarth:
@jackpoz which changes should I do? I declared void ```AddSC_duel_reset();``` inside the .cpp file.

src/server/scripts/Custom/duel.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+class DuelResetScript : public PlayerScript
+{
+    public:
+        DuelResetScript() : PlayerScript("DuelResetScript") { }
+
+        // Called when a duel starts (after 3s countdown)
+        void OnDuelStart(Player* player1, Player* player2)
+        {
+            if (sWorld->getBoolConfig(CONFIG_RESET_DUEL_COOLDOWNS))
```
Naios: try to move this into the script validation hook if there is any

src/server/scripts/World/duel_reset.cpp
```diff
@@ -0,0 +1,56 @@
...
+                player1->GetSpellHistory()->SaveCooldownStateBeforeDuel();
+                player2->GetSpellHistory()->SaveCooldownStateBeforeDuel();
+
+                player1->RemoveArenaSpellCooldowns(true);
+                player2->RemoveArenaSpellCooldowns(true);
+            }
+        }
+
+        // Called when a duel ends
+        void OnDuelEnd(Player* winner, Player* loser, DuelCompleteType type)
```
tkrokli: At the moment, it looks like Travis is only complaining about `type` being unused.

Runico:
@ShinDarth an exploit exist in this script

How to reproduce
1. Use Item [Dimensional Ripper - Area 52] ( SpellId = 36890 ) has 4 Hours cooldown
2. Do a normal duel
3. Item [Dimensional Ripper - Area 52] doesn't have cooldown anymore

tkrokli:
Working links, in case you want to look up the spells/items:
http://www.wowhead.com/item=30542/dimensional-ripper-area-52 (item crafted by engineering)
http://www.wowhead.com/spell=36890/area52-transporter (spell used by the item itself)
http://www.wowhead.com/spell=36954/dimensional-ripper-area-52 (crafting spell to create the item)

ShinDarth:
@Runico 

thanks for the report, but it is just a visual bug. You can't actually use that item again. The client thinks that there is no cooldown, but server side the cooldown is still here, so there is no exploit possible. If you try to relog, the client will see again the real (correct) cooldown.

The visual bug is caused in this line:
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellHistory.cpp#L713

the condition causing the visual bug is ```cooldownDuration > 10 * MINUTE * IN_MILLISECONDS```

logically, we shouldn't use that condition. But that check was needed that in order to avoid some strange bugs like visual cooldowns of 1 hour on some spells (like priest shield or warrior stances), which were very annoying on player side.

It is possible that [this commit](https://github.com/TrinityCore/TrinityCore/commit/7dcddd90be6e49281ca04239de5839e757ef3f7a) resolved that bug, so we may remove the condition ```cooldownDuration > 10 * MINUTE * IN_MILLISECONDS``` and fix the visual bug that you reported. But I'm not sure of it.

You may test by removing that condition, it will surely remove the visual bug that you reported, but it may introduce those other old bugs and be aware that they are not very easy to reproduce (I and @GigaDev90 spent a lot of time in order to debug them).

Runico:
@Shindarth, I've test this many times and yes it's works and it's not just a visual bug

ShinDarth:
@Runico you are right, there is an exploit on newer revisions. I was testing on an older one where there is no exploit: https://github.com/TrinityCore/TrinityCore/commit/302623a1516e487ad64f2028aec287b99372e33a

ShinDarth:
@Runico  please try with https://github.com/TrinityCore/TrinityCore/pull/15947

Runico:
@ShinDarth Now it's just visual bug, better than old one xD

Thanks

PR #15728, opened 2015-10-18
Core/Player: implement Duel Reset HP and Mana config

GigaDev90:
This PR introduces a new configurable setting (disabled by default) which reset the mana and health of the player before and after Duel.
Thanks to shindarth for help me during the work.

PR #15732, opened 2015-10-18
Scripts/Duel: implement duel reset of health and mana and improve duel reset

FrancescoBorzi:
This PR is about the Duel Reset script and it is composed by several parts:
## Implementation of Health & Mana reset

Originally written by @GigaDev90 (https://github.com/TrinityCore/TrinityCore/pull/15728), rewritten using scripts.

This PR introduces a configurable feature to reset health and mana before and after duel.
- Before duel it saves health and mana(if any), and set them to the max value
- After duel it restore the old health and mana(if any)
## Improvement of the whole Duel Reset system

This PR also introduces a global improvement for the Duel Reset system: the reset should not be performed if duel isn't started (DUEL_INTERRUPTED) or fled (DUEL_FLED).
## Fix restoration of the Druid mana

The initial part didn't handle the restoration of the Druid mana, with a commit of Giga we handle that too. (by @GigaDev90)
## Fix onHold spells

Fixed bug when a player accept duel with a spell on onHold true (like sthealt of dudu when is active). Now in any condition of onHold never happens visual bug or wrong store of cooldown. (by @GigaDev90)


src/server/game/Entities/Player/Player.h
```diff
@@ -1946,6 +1946,12 @@ class Player : public Unit, public GridObject<Player>
         uint32 GetMaxPersonalArenaRatingRequirement(uint32 minarenaslot) const;
         void SetHonorPoints(uint32 value);
         void SetArenaPoints(uint32 value);
+        
+        // duel health and mana reset methods
+        void SaveHealthBeforeDuel() { healthBeforeDuel = GetHealth(); }
+        void SaveManaBeforeDuel() { manaBeforeDuel = GetPower(POWER_MANA); }
+        void RestoreHealthAfterDuel() { SetHealth(healthBeforeDuel); }
+        void RestoreManaAfterDuel() { SetPower(POWER_MANA, manaBeforeDuel); }
```
joschiwald: you could save this stuff in script too

Shauren: no, you cant. PlayerScript is not the same as an AI script - there is only one global instance of it

joschiwald: ... by using a std::unordered_map with guid as key

ShinDarth: I think the current one is the most elegant way....

Shauren: @joschiwald which would be again global - and not threadsafe

ShinDarth:
I would like to avoid commit squashing, since there are commits doing different things and/or with different authors

src/server/game/Entities/Player/Player.h
Naios: Maybe we could add a "custom" store in player which is adaptable but not global.
We could use it to replace several other less used optional core or custom stores,
which would also increase the compression factor for unused entries:

```c++
enum class PlayerExtraUINT32Store
{
    STORE_PLAYER_HEALTH_BEFORE_DUEL,
    STORE_PLAYER_MANA_BEFORE_DUEL
};

Optional<uint32> Player::GetCustomStorageEntry(PlayerExtraUINT32Store::STORE_PLAYER_MANA_BEFORE_DUEL)
{
    ...
}

bool Player::ClearCustomStorageEntry(PlayerExtraUINT32Store::STORE_PLAYER_MANA_BEFORE_DUEL)
{
    ...
}
```

Shauren: Optional still takes memory, even if not initialized

Naios: No... the value would be stored in a hashmap... The optional is only the return type

Shauren: Then again, it doesnt have to be an uint32 (hence boost::any suggestion)

Naios: Right

ShinDarth:
This PR fixed all bugs of the duel reset system and added new features to it.

I've been testing it for a while, and it seems to work fine. There is no reason for this PR to be still open, please merge it

tkrokli:
Nice job, ShinDarth! :+1:

PR #15746, opened 2015-10-20
Core/Spells: Reduce Duration for EffectEnchantHeldItem-Spells

### LINKED ISSUE #15739: [3.3.5][6.x][Core/Spell] Venomhide Poison - Enchant with wrong duration
rev: https://github.com/TrinityCore/TrinityCore/commit/c6c3e72b55846ccfb565fd09580509800bd25c2a

Facts:
http://www.wowhead.com/spell=14792/venomhide-poison
You get this echantspell if you fight against http://www.wowhead.com/npc=6508/venomhide-ravasaur

Problem:
The duration of the enchant:
Enchant Held Item (OBSOLETE): Venomhide Poison (1003)
holds 9 HOURS

How it should be:
Enchant Held Item (OBSOLETE): Venomhide Poison (1003)
The duration should be 5 MINUTES.
The procchance and the procflags are all correct. it's only an issue with the duration of the enchant

Source:
Venomhide Poison is a temporary weapon enchantment that lasts for 5 minutes.
http://wowwiki.wikia.com/wiki/Venomhide_Poison

What i tried to fix the issue:
https://github.com/TrinityCore/TrinityCore/blob/6.x/src/server/game/Spells/SpellEffects.cpp#L2692
I added the spellinfo:
```c++
else if (m_spellInfo->Id == 14792)
        duration = 300;                                     // 5 mins
```
but this didn't changed anything. so i'm pretty lost here :D 

sideproblem: many people abuse this enchant to deal further damage in raids

Lumber-Jack-2:
closes #15739
- maybe there is another option to get the real duration of the enchant from spells with Effect: SPELL_EFFECT_ENCHANT_HELD_ITEM
- but since there are only 3 spells which use this effect, we can change the multiplier here 
- ID - 53160 Dancing Rune Weapon Visual and ID - 19108 MHTest02 are the other spells, which aren't affected from any change in the duration
- the fix itself works fine > http://i.imgur.com/lEAOEKR.jpg

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -4373,11 +4373,14 @@ void Spell::EffectEnchantHeldItem(SpellEffIndex effIndex)
     if (m_spellInfo->Effects[effIndex].MiscValue)
     {
         uint32 enchant_id = m_spellInfo->Effects[effIndex].MiscValue;
-        int32 duration = m_spellInfo->GetDuration();          //Try duration index first ..
+        int32 duration = m_spellInfo->GetDuration();  // Try duration index first ..
         if (!duration)
-            duration = damage;//+1;            //Base points after ..
+            duration = damage;//+1;                   // Base points after ..
         if (!duration)
-            duration = 10;                                  //10 seconds for enchants which don't have listed duration
+            duration = 10 * IN_MILLISECONDS;          // 10 seconds for enchants which don't have listed duration
+
+        if (m_spellInfo->Id == 14792) // Venomhide Poison
+            duration = 5 * MINUTE * IN_MILLISECONDS;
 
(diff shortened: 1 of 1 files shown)


Rushor:
updated the stuff from https://github.com/TrinityCore/TrinityCore/issues/15739#issuecomment-149549923 
feel free to merge this if noone is against the hardcoded check for spellid 14792

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -4379,6 +4379,9 @@ void Spell::EffectEnchantHeldItem(SpellEffIndex effIndex)
         if (!duration)
             duration = 10;                                  //10 seconds for enchants which don't have listed duration
```
Shauren: should multiply this value by IN_MILLISECONDS (since SetEnchantment expects value in ms and it was removed from line 4397 as it didnt make sense to mutliply a value already in ms (from spell duration) again)

PR #15755, opened 2015-10-23
Fixed some rare stealth/prowl visual bugs after duel cd reset

FrancescoBorzi:
Sometimes stealth and prowl spells cause a huge cooldown which should not be sent to the client to avoid a rare visual bug.

All cooldowns we give back to the user should be between 0 and 10 minutes, this should prevent any kind of visual bugs.

Tested and working.


Runico:
this bug still happens for druids

tkrokli:
@Runico : if you can document it (and there is no open issues for it), you could open a new issue.

ShinDarth:
@MitchesD  we need https://github.com/TrinityCore/TrinityCore/pull/15732 which fixes every bug the reset duel script has.

PR #15763, opened 2015-10-25
[3.3.5] Fix conflicts between duration and cast time of channeled spells

FrancescoBorzi:
Method **Unit::ModSpellCastTime** .

Facts:
- It is used both for modify the cast time of all spells and the duration time of channelled spells
- It returns and does nothing when a duration of a channelled spell should not be affected by haste
- Some rare channelled spells has both cast time and channelling duration time (example: Seduction)

Problems:
- When we try to modify the cast time of a channelled spell using that method, we can't do nothing if the duration shouldn't be affected by haste (but this should be unreleated with the cast time)
- There are useless repeated checks

Solution:
- Split this method into **Unit::ModSpellCastTime** and **Unit::ModSpellDurationTime**. The first should be used to modify the cast time **only**, and the latter should be used to modify the duration time **only**.
- No need to check if spell duration should be affected by haste before calling **Unit::ModSpellDurationTime**, because the method already does it

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12987,9 +12987,6 @@ void Unit::ModSpellCastTime(SpellInfo const* spellInfo, int32 & castTime, Spell*
     if (!spellInfo || castTime < 0)
         return;
 
-    if (spellInfo->IsChanneled() && !spellInfo->HasAttribute(SPELL_ATTR5_HASTE_AFFECT_DURATION))
-        return;
-
     // called from caster
     if (Player* modOwner = GetSpellModOwner())
         modOwner->ApplySpellMod(spellInfo->Id, SPELLMOD_CASTING_TIME, castTime, spell);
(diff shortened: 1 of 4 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12992,15 +13010,15 @@ void Unit::ModSpellCastTime(SpellInfo const* spellInfo, int32 & castTime, Spell*
 
     // called from caster
     if (Player* modOwner = GetSpellModOwner())
-        modOwner->ApplySpellMod(spellInfo->Id, SPELLMOD_CASTING_TIME, castTime, spell);
+        modOwner->ApplySpellMod(spellInfo->Id, SPELLMOD_DURATION, time, spell);
```
ShinDarth: visual github bug, the line is:
```modOwner->ApplySpellMod(spellInfo->Id, SPELLMOD_DURATION, time, spell);```

**Comment by iarkoff** [comment]
This pull makes the talent part of #13669 fixable, am i wrong ?

File: src/server/game/Entities/Unit/Unit.cpp [hunk 2]
```diff
@@ -12982,9 +12982,27 @@ int32 Unit::ModSpellDuration(SpellInfo const* spellProto, Unit const* target, in
     return std::max(duration, 0);
 }
 
-void Unit::ModSpellCastTime(SpellInfo const* spellInfo, int32 & castTime, Spell* spell)
+void Unit::ModSpellCastTime(SpellInfo const* spellInfo, int32 & time, Spell* spell)
```
joschiwald: pls rename back to `castTime`

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12982,9 +12982,27 @@ int32 Unit::ModSpellDuration(SpellInfo const* spellProto, Unit const* target, in
...
+    if (!(spellInfo->HasAttribute(SPELL_ATTR0_ABILITY) || spellInfo->HasAttribute(SPELL_ATTR0_TRADESPELL) || spellInfo->HasAttribute(SPELL_ATTR3_NO_DONE_BONUS)) &&
+        ((GetTypeId() == TYPEID_PLAYER && spellInfo->SpellFamilyName) || GetTypeId() == TYPEID_UNIT))
+        time = int32(float(time) * GetFloatValue(UNIT_MOD_CAST_SPEED));
+    else if (spellInfo->HasAttribute(SPELL_ATTR0_REQ_AMMO) && !spellInfo->HasAttribute(SPELL_ATTR2_AUTOREPEAT_FLAG))
+        time = int32(float(time) * m_modAttackSpeedPct[RANGED_ATTACK]);
+    else if (spellInfo->SpellVisual[0] == 3881 && HasAura(67556)) // cooking with Chef Hat.
+        time = 500;
+}
+
+void Unit::ModSpellDurationTime(SpellInfo const* spellInfo, int32 & time, Spell* spell)
```
joschiwald: and here `duration`

ShinDarth:
@joschiwald done

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -13003,6 +13000,27 @@ void Unit::ModSpellCastTime(SpellInfo const* spellInfo, int32 & castTime, Spell*
...
+{
+    if (!spellInfo || duration < 0)
+        return;
+
+    if (spellInfo->IsChanneled() && !spellInfo->HasAttribute(SPELL_ATTR5_HASTE_AFFECT_DURATION))
+        return;
+
+    // called from caster
+    if (Player* modOwner = GetSpellModOwner())
+        modOwner->ApplySpellMod(spellInfo->Id, SPELLMOD_DURATION, duration, spell);
```
ShinDarth: visual github bug, the line is:
modOwner->ApplySpellMod(spellInfo->Id, SPELLMOD_DURATION, time, spell);

ShinDarth:
Any feedback? waiting for almost a month ^^

ccrs:
I tested this for like a week, and still no errors found.
Havent tested anything specific though; if you have any "special" spell to test I can give some feedback.

ShinDarth:
@ccrs thanks for your feedback.

Some time ago a change like this was enough to fix the Improved Succubus warlock talent, but now it is not enough (there is something else broken with that), but I think this can be a good base to start.

ccrs:
both

PR #15764, opened 2015-10-25
Core/Conditions: More beautiful code

r4d1sh:


Shauren:
what if i prefer it the way i originally wrote it?

PR #15771, opened 2015-10-28
Core/Spells: Fixed a mistake in SpellInfo::CalcPowerCost

Zedron:
Fixed a mistake which caused some SPELLMOD_COST modifiers to give (lots) of power instead of taking power in Spell::TakePower

### DIFF
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -2729,7 +2729,8 @@ std::vector<SpellInfo::CostData> SpellInfo::CalcPowerCost(Unit const* caster, Sp
     else
         collector(sDB2Manager.GetSpellPowers(Id, caster->GetMap()->GetDifficultyID()));
 
-    std::remove_if(costs.begin(), costs.end(), [](CostData const& cost) { return cost.Amount <= 0; });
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/SpellInfo.cpp
```diff
@@ -2729,7 +2729,8 @@ std::vector<SpellInfo::CostData> SpellInfo::CalcPowerCost(Unit const* caster, Sp
     else
         collector(sDB2Manager.GetSpellPowers(Id, caster->GetMap()->GetDifficultyID()));
 
-    std::remove_if(costs.begin(), costs.end(), [](CostData const& cost) { return cost.Amount <= 0; });
+    // POWER_RUNES is handled by SpellRuneCost.db2, and cost.Amount is always 0 (see Spell::TakeRunePower)
+    costs.erase(std::remove_if(costs.begin(), costs.end(), [](CostData const& cost) { return cost.Power != POWER_RUNES && cost.Amount <= 0; }), costs.end());
```
jackpoz: is there a difference between
```
std::remove_if(
```
and
```
costs.erase(std::remove_if(
```
?

Shauren: apparently there is. std::remove_if is stupid and very confusing...
turns out std::remove_if only MOVES the elements in the sequence (while std::list<T>::remove_if REMOVES them) and you need to call .erase by yourself

PR #15778, opened 2015-10-29
Updates

Ayik0:


sql/updates/world/2015_10_30_Stealth_Fix_world.sql
```diff
@@ -0,0 +1,3 @@
+REPLACE INTO `spell_linked_spell` (`spell_trigger`, `spell_effect`, `type`, `comment`) VALUES ('1784', '158185', '2', 'Stealth mod');
```
tkrokli: First of all, TC never uses `REPLACE INTO`. Use the combination queries `DELETE FROM` and `INSERT INTO`.
Second, never use ' ' around numbers.
Third, with `INSERT INTO` you can place the values on their separate lines, only needing 1 `INSERT INTO`.
```sql
DELETE FROM `spell_linked_spell` WHERE `spell_trigger` IN (1784,115191) AND `spell_effect` IN (158185,158188);
INSERT INTO `spell_linked_spell` (`spell_trigger`, `spell_effect`, `type`, `comment`) VALUES
(1784,158185,2,'Stealth mod'),
(1784,158188,2,'Stealth shapeshift'),
(115191,158185,2, 'Stealth mod');

Shauren:
duplicate #15779

PR #15782, opened 2015-10-30
Core/Threads: Replace Boost TLS with C++11 one

jackpoz:
Replace boost::thread_specific_ptr<T> thread-local storage with C++11 thread_local to remove libboost_thread dependency from common project.

Tested on VS2015 authserver and worldserver with multiple map threads, a debugging session showed how tls object were being created once for each thread (map threads, boost asio thread, startup thread). VS2013 Update 4 and GCC compatibility should be checked before merging.


src/common/Utilities/Util.cpp
```diff
@@ -31,19 +30,14 @@
   #include <arpa/inet.h>
 #endif
 
-static boost::thread_specific_ptr<SFMTRand> sfmtRand;
+static thread_local SFMTRand* sfmtRand = nullptr;
```
Naios: `unique_ptr`? The block is deallocated on application exit anyway but it could prevent false positives from valgrind.

Naios: And:

```c++
class SFMTRandGetter
{
public:
    SFMTRandGetter() { }

    SFMTRand* operator-> ()
    {
        if (!_impl)
            _impl = Trinity::make_unique<SFMTRand>();

        return _impl.get();
    }

private:
    std::unique_ptr<SFMTRand> _impl;
};

static thread_local SFMTRandGetter sfmtRand;

// sfmtRand->
```

hides `GetRng()` completely.

jackpoz: can you pr (or push) that ? As long as there's no boost anything is fine to me

Naios:
VS2013 Update 4 doesn't support the `thread_local` keyword.

Naios:
Jackpoz see [Naios/TrinityCore@thread_local](https://github.com/Naios/TrinityCore/commits/thread_local).

Shauren:
@Naios no good, even if you #define thread_local to __declspec(thread) on vs 2013 you are limited to POD types (this means a simple pointer is as far as you can go)

DDuarte:
It might be a bit *hardcore* but that SFMT stuff could be replaced with `<random>`

Shauren:
Thats not the point! (also it would achieve nothing, you would still need thread local generator)

jackpoz:
this PR will have to wait for VS2015 (and GCC/clang version) made mandatory

Naios:
Maybe we could collect all changes on another tc `c++14` branch so we dont loose it.

jackpoz:
you could open an issue C++14 and reference this pr

Naios:
What about using thread_local conditionally on msvc 2013 as `__declspec(thread)` as shauren suggested?

Shauren:
its pointless because thread_local also crashes the thing that @jackpoz is doing

Naios:
@Shauren Why?

jackpoz:
Now it would be interesting to check if this is supported by the current compiler versions

Issue #15190, opened 2015-07-29
[6.x] Core/Movement: creature_formations - FollowerGUID follows LeaderGUID very bad while flying


Rushor:
hash: https://github.com/TrinityCore/TrinityCore/commit/3e083c5eac91c82d2b8bbc2387f169430aafab0f
Problem: Creatures with Inhabitype 4 and creature_formations are following the Leader very bad until the follwerguid stucks in the air

How to reproduce:
1) Apply this https://gist.github.com/Rushor/121732b9cc767b8a39c2 for the wyvern in Orgrimmar
2) .go c 310863
3) look at creatureguid 310863 and 310862

Result:
You will see that the follower310862 will start following the leader 310863 - but soon the follower 310862 is going to fly to the ground with a very slow speed. The followerGUID seems to be unable to fly. The LeaderGUID meanwhile keeps flying around until it is out of sight and becomes inactive. This causes the FolloerGUID to become inactive and it starts stucking in the ground and does not move any longer.

Mainproblem which has to be improved:
FollowerGUID should always be able to fly

Rushor:
i will make a vid to demonstrate all - without this fix orgrimmar cannot be completed xD

Rushor:
kay done - https://youtu.be/fX9Q4e02oiI - here you can see how the follower goes down to the earth (yes i now it's 4.x client, but i'm unable to make a vid of 6.x and in 6.x it looks the same)

Rushor:
```c++
diff --git a/src/server/game/Entities/Creature/CreatureGroups.cpp b/src/server/game/Entities/Creature/CreatureGroups.cpp
index 2ccf447..c55ee9f 100644
--- a/src/server/game/Entities/Creature/CreatureGroups.cpp
+++ b/src/server/game/Entities/Creature/CreatureGroups.cpp
@@ -244,13 +244,14 @@ void CreatureGroup::LeaderMoveTo(float x, float y, float z)
         Trinity::NormalizeMapCoord(dx);
         Trinity::NormalizeMapCoord(dy);
 
-        member->UpdateGroundPositionZ(dx, dy, dz);
+        //member->UpdateGroundPositionZ(dx, dy, dz);
 
         if (member->IsWithinDist(m_leader, dist + MAX_DESYNC))
             member->SetUnitMovementFlags(m_leader->GetUnitMovementFlags());
         else
             member->SetWalk(false);
 
+        member->UpdateMovementFlags();
[...]

PR #15791, opened 2015-10-31
Core/CreatureGroups: Don't set positionZ to ground for flying creatures.

Golrag:
Should fix some pathing with the windriders in orgrimmar :)


Killyana:
It will fix this: https://github.com/TrinityCore/TrinityCore/issues/15190 ?

Golrag:
Yes, just wondering if I should check if creature is flying or if creature can fly and dZ > groundZ :p

PR #15801, opened 2015-11-01
[BlackrockMountain] Remove Sripts Boss Anubshiah

ghost:
The script can be removed from the use of SAI.

[commit](https://github.com/TrinityCore/TrinityCore/commit/2819437da01281afc87c8078aec56a4725b2f27f)


MitchesD:
when removing a file you have to remove it from cmake and scipt loader as wel

PaulHellCore:
My error.

Issue #6541, opened 2012-05-17
[Instance] ICC Sindragosa - Ice Tomb Summon


deny70:
well first off this is my first one but never mind lol.So its known sindragosa is realy screwed with the thombs......they have line of sight isues, DMG from bombs are took from all raid members :3 thanks for your time :D

tkrokli:
> @Rushor can u send me the code please? I cannot open the websites http://pastebin.com/9ywfCnUK,sorry.

I suppose you are referring to this page & code: http://pastebin.com/9ywfCnUK
(your link was broken by making ".sorry" part of your own link) Anyway, the code looks like this:
```diff
 -1335,6 +1335,13 @@
     if (!IsInMap(obj))
         return false;
 
+    // hack for ice tomb's gameobject
+    if (obj->GetTypeId() == TYPEID_UNIT)
+        if (obj->GetEntry() == 36980 /* Ice Tomb */)
+            return true;
+    if (GetTypeId() == TYPEID_UNIT)
+        if (GetEntry() == 36980 /* Ice Tomb */)
+            return true;
     float ox, oy, oz;
     obj->GetPosition(ox, oy, oz);
     return IsWithinLOS(ox, oy, oz);
```
Note: that code is tagged as a hack fix.

D4R4:
@tkrokli the hackfix worked 100% for quite some but recently with it only direct spells hit the Tomb and aoe spells like warrior's Whirwind or for example priest's mind seer deal no damage and still cant get passed the LoS check somehow, need to fix the hack fix :) working on that atm

tkrokli:
Good to hear that, @D4R4 :dragon_face: it will be nice to bring some good news to my fellow players later on. :dragon:

D4R4:
@tkrokli @Dorsch I managed to do what Dorsch said, and I must admit it is looking pretty good,

so baisically the GO doesn't spawn when the IceTomb npc is created but instead only for a friction of second it is spawned BEFORE the FrostBomb hits the gound and instantly Deleted AFTER it has done it's damage,

``_owner->CastSpell((Unit*)NULL, SPELL_FROST_BOMB, false, NULL, NULL, _sindragosaGUID);``

seems to be working with no problem, gonna post the code later today

tkrokli:
Good job! Nice to see that your investigation has paid off! :+1:

D4R4:
https://gist.github.com/D4R4/2f0fe5736eb174d60de6

[...]

PR #15807, opened 2015-11-03
Core/Object: Use object edges to calculate LoS

Golrag:
Also fixed https://github.com/TrinityCore/TrinityCore/issues/6541 :)

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1146,9 +1146,13 @@ bool WorldObject::IsWithinLOSInMap(const WorldObject* obj) const
     if (!IsInMap(obj))
         return false;
 
-    float ox, oy, oz;
-    obj->GetPosition(ox, oy, oz);
-    return IsWithinLOS(ox, oy, oz);
+    float x, y, z;
+    if (obj->GetTypeId() == TYPEID_PLAYER)
+        obj->GetPosition(x, y, z);
+    else
+        obj->GetHitSpherePointFor(GetPosition(), x, y, z);
+
+    return IsWithinLOS(x, y, z);
 }
 
 float WorldObject::GetDistance(const WorldObject* obj) const
@@ -1232,9 +1236,34 @@ bool WorldObject::IsWithinLOS(float ox, float oy, float oz) const
     VMAP::IVMapManager* vMapManager = VMAP::VMapFactory::createOrGetVMapManager();
     return vMapManager->isInLineOfSight(GetMapId(), x, y, z+2.0f, ox, oy, oz+2.0f);*/
     if (IsInWorld())
-        return GetMap()->isInLineOfSight(GetPositionX(), GetPositionY(), GetPositionZ()+2.f, ox, oy, oz+2.f, GetPhaseMask());
+    {
+        float x, y, z;
+        if (GetTypeId() == TYPEID_PLAYER)
+            GetPosition(x, y, z);
+        else
+            GetHitSpherePointFor({ ox, oy, oz }, x, y, z);
+
+        return GetMap()->isInLineOfSight(x, y, z + 2.0f, ox, oy, oz + 2.0f, GetPhaseMask());
+    }
 
     return true;
+}
+
+Position WorldObject::GetHitSpherePointFor(Position const& dest) const
+{
+    G3D::Vector3 vThis(GetPositionX(), GetPositionY(), GetPositionZ());
+    G3D::Vector3 vObj(dest.GetPositionX(), dest.GetPositionY(), dest.GetPositionZ());
+    G3D::Vector3 contactPoint = vThis + (vObj - vThis).directionOrZero() * GetObjectSize();
+
+    return Position(contactPoint.x, contactPoint.y, contactPoint.z, GetAngle(contactPoint.x, contactPoint.y));
+}
+
+void WorldObject::GetHitSpherePointFor(Position const& dest, float& x, float& y, float& z) const
+{
+    Position pos = GetHitSpherePointFor(dest);
+    x = pos.GetPositionX();
+    y = pos.GetPositionY();
+    z = pos.GetPositionZ();
 }
 
 bool WorldObject::GetDistanceOrder(WorldObject const* obj1, WorldObject const* obj2, bool is3D /* = true */) const
--- a/src/server/game/Entities/Object/Object.h
+++ b/src/server/game/Entities/Object/Object.h
@@ -487,6 +487,8 @@ class WorldObject : public Object, public WorldLocation
         bool IsWithinDistInMap(WorldObject const* obj, float dist2compare, bool is3D = true) const;
         bool IsWithinLOS(float x, float y, float z) const;
         bool IsWithinLOSInMap(WorldObject const* obj) const;
+        Position GetHitSpherePointFor(Position const& dest) const;
+        void GetHitSpherePointFor(Position const& dest, float& x, float& y, float& z) const;
         bool GetDistanceOrder(WorldObject const* obj1, WorldObject const* obj2, bool is3D = true) const;
         bool IsInRange(WorldObject const* obj, float minRange, float maxRange, bool is3D = true) const;
         bool IsInRange2d(float x, float y, float minRange, float maxRange) const;
--- a/sql/updates/world/yyyy_mm_dd_xx_world.sql
+++ b/sql/updates/world/yyyy_mm_dd_xx_world.sql
@@ -1,0 +1,1 @@
+UPDATE `creature_model_info` SET `BoundingRadius`=7.5, `CombatReach`=6 WHERE `DisplayID`=30890;


Aokromes:
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Entities/Object/Object.cpp:1150:33: fatal error: 
      taking the address of a temporary object of type 'Position'
      [-Waddress-of-temporary]
    float angle = obj->GetAngle(&GetPosition());
                                ^~~~~~~~~~~~~~

WoWShards:
All checks have passed
1 successful check

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1147,7 +1147,12 @@ bool WorldObject::IsWithinLOSInMap(const WorldObject* obj) const
         return false;
 
     float ox, oy, oz;
-    obj->GetPosition(ox, oy, oz);
+    float angle = obj->GetAngle(GetPositionX(), GetPositionY());
+    float range = std::min(GetObjectSize(), GetExactDist(obj->GetPositionX(), obj->GetPositionY(), obj->GetPositionZ()));
```
pete318: If I'm understanding what you're doing right, IsWithinLosInMap with object is checking the position of the object and getting the edge based on the angle of the source. Then passing that into the co-ordinate version of the same function which will do the same for the source object. Like this:

image

In that case, should the above not be obj->GetObjectSize() ?

It's possible I'm misunderstanding things a bit. But if the picture is right, maybe the above should be changed.

Overall this is a good thing. I've seen the LoS issues for a while and it was on my list of things to "look at" but admittedly quite some way down :)

Golrag: Nice catch ! Made some changes on local will push it ASAP :)

However there is a bug now with those changes: http://puu.sh/l7QM8/b75b680eeb.jpg

Don't know how to fix this and I don't know if it is a bug :/

pete318: Well the purple line is what I think it would have done using the GetObjectSize rather than obj->GetObjectSize. e.g. it would use the size of the from object as bounds for the to object.. Which would be a much bigger problem for example if the from object was much bigger than the "to" object.. :)

As for your problem.. Maybe put a TC_LOG_INFO with the results of the checks for both calls and see if anything stands out as odd?

Golrag: It taking the bottom position of the objects and checks if those are in los (which, in the case of the screenshot, are not.. :/)

pete318: Hmm. Z position isn't already middle of object? Easily changed at least.

Golrag: Tried
```cpp
oz = GetPositionZ() + range;
```
Will try to set it to the middle of the object :)

Golrag: .debug LoS says I am in LoS with it, wondering why https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/Spell.cpp#L4843
it checks if target is in los with caster

Golrag: @pete318 changed it a bit. You will always be in LoS now if GetDistance() returns 0.0f

Carbenium:
@pete318 could you take a look whether this PR is fine for you? If yes, feel free to merge i guess

pete318:
Yeah I'll see if I can't set this up and run some tests later.

Shauren:
As a player you definitely should not be able to run your face into a thin wall and cast through it (if wall width < player->GetObjectSize())

Golrag:
@Shauren, mh will take a look a that.. Also wanted to remove the + 2.f's, should I replace them with GetObjectSize() ? Cause if I do, Sindragosa moves strangely. But not sure if it's blizzlike or not

Golrag:
I'm stuck.. I've got an image of possible scenarios, first two based on the ice tomb, three others based on @Shauren's example: http://puu.sh/loNnB/e0a6d3bc00.png (note: top is retail returns, bot is what my implementation returns).

What I've tried now, when the edge position is not in los with center position, relocate the edge position to center. Like when you want to place a totem but destPosition in not in LoS so it spawns at your position. But that breaks example 1 & 2.
And I have no idea how to get wall width :/

Shauren:
GetObjectSize is width, not z (height) (+2 or rather, CollisionHeight?) - not the part you should be looking at

Treeston:
Use center for players, use edge for creatures. Sounds simple?

Golrag:
Was thinking about that but I don't know if that's blizzlike behavior :/

Will do it anyway to reduce possible exploits.

Golrag:
Updated to https://github.com/TrinityCore/TrinityCore/commit/e6a7818a9a12b085c8ea07cfb8b3c6d1d700a112

Using center for players now

DevRival:
It still does not help, but can do damage AoE, but can not be attacked in the early mile.

Golrag:
@DevRival I don't have LoS issues with Sindragosa's Ice Block with this PR. If you have more information on how to reproduce it, please share it :)

DevRival:
It seems to me you forgot something.
``` SQL 
UPDATE `creature_model_info` SET `BoundingRadius`=7.5, `CombatReach`=6 WHERE `DisplayID`=30890;

**Comment by Treeston** [comment]
This has lots of LoS issues  in Azjol'Nerub, especially in the first sloped corridor (try pulling the first boss there) and the Hadronox area.

**Comment by Golrag** [comment]
@Treeston thanks for the report :) will look at it next week if I have some time !

**Comment by Treeston** [comment]
I assume it's because you are using 2d nearpoint. You should be using 3d nearpoint.

**Comment by Golrag** [comment]
@Treeston I agree, kinda looks like that. LoS should be calculated from midsection to midsection. Will make it so that it will only use the edges for creatures (not for players, and not for gameobjects). So I can use https://github.com/TrinityCore/TrinityCore/blob/4f78efd4633f79285f176b61367a067b2cd90e2b/src/server/game/DataStores/DBCStructure.h#L311

**Comment by Treeston** [comment]
I disagree. LoS should be calculated from closest point on the hit sphere. You are already doing this, but you are only using 2D hit circles rather than 3d hit spheres.

You need to offset the position in 3D using the 3D angle between the positions, not in 2D using the 2D angle.

**Comment by Golrag** [comment]
Thought that it was a cylindir cause of http://wow.gamepedia.com/index.php?title=Line_of_sight
Will see what I can do to make it a sphere, Still only for creatures :p

**Comment by Golrag** [comment]
@Treeston any idea on how to get that position with a sphere ? Tried something with G3D::Vector3::moveTowards but I don't think the result is what I really want :/

**Comment by Treeston** [comment]
```cpp
    G3D::Vector3 vThis(GetPositionX(), GetPositionY(), GetPositionZ());
    G3D::Vector3 vObj(obj->GetPositionX(), obj->GetPositionY(), obj->GetPositionZ());
    G3D::Vector3 contactPoint = (vObj - vThis).directionOrZero() * GetObjectSize();
```

If you use `G3D::Vector3` for it. There may be more efficient ways in the core's own logic, I dunno.

Golrag:
Tried it in the Test zone
```
contactPoint {x=1.74494815 y=-5.69913864 z=0.689188242 }
vObj {x=3.30446625 y=-14.1103020 z=-142.708649 }
vThis {x=-1.76020002 y=2.43129992 z=-144.709000 }
```
Z position seems wrong :p

Treeston:
Err.

```cpp
    G3D::Vector3 vThis(GetPositionX(), GetPositionY(), GetPositionZ());
    G3D::Vector3 vObj(obj->GetPositionX(), obj->GetPositionY(), obj->GetPositionZ());
    G3D::Vector3 contactPoint = vThis + (vObj - vThis).directionOrZero() * GetObjectSize();
```

Of course.

Treeston:
Or maybe switch the roles of `obj` and `this` looking at current code.

Golrag:
@Treeston thanks ! Still looking into that CollisionHeight instead of +2.0f :)

Golrag:
Updated it ! Moved calculation of the object's edge to it's own function. :)

Treeston:
The method shouldn't be in `Creature`, but in `Object`. Also, provide an override that returns a `Position` and only takes one argument - and rename it to something more descriptive.

Golrag:
Updated, better name now ? My english is limited so if you have any suggestion feel free :D

Treeston:
Something like `GetHitSpherePointFor`?

Golrag:
Renamed ! :)

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2893,3 +2893,12 @@ void Creature::ClearTextRepeatGroup(uint8 textGroup)
     if (groupItr != m_textRepeat.end())
         groupItr->second.clear();
 }
+
+Position Creature::GetHitSpherePointFor(Position const dest) const
```
Treeston: Why not `Position const& dest`?

src/server/game/Entities/Object/Object.h
```diff
@@ -487,6 +487,7 @@ class WorldObject : public Object, public WorldLocation
         bool IsWithinDistInMap(WorldObject const* obj, float dist2compare, bool is3D = true) const;
         bool IsWithinLOS(float x, float y, float z) const;
         bool IsWithinLOSInMap(WorldObject const* obj) const;
+        virtual Position GetHitSpherePointFor(Position const dest) const;
```
Treeston: Why do you use virtual polymorphisms here instead of checking for player in whatever function calls it?

IMO it would be cleaner to have the "proper" method here and then just choose not to use it for players in the calling method.

Golrag: It looked cleaner to me like that, for players I intend(ed) to use ```GetPositionZ() + Player::GetCollisionHeight```. Instead of + 2.0f
I could do it in one function, but won't it be too messy if I do so ? :/

Treeston: I mean have just `GetHitSpherePointFor`, without `virtual`, in `Object` (with the functionality from `Creature`).

Then, wherever you call `GetHitSpherePointFor`, check if it's a player, and if yes, use the `+2.0f` position instead of calling `GetHitSpherePointFor`.

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1232,11 +1236,28 @@ bool WorldObject::IsWithinLOS(float ox, float oy, float oz) const
...
     if (IsInWorld())
-        return GetMap()->isInLineOfSight(GetPositionX(), GetPositionY(), GetPositionZ()+2.f, ox, oy, oz+2.f, GetPhaseMask());
+    {
+        Position pos;
+        if (GetTypeId() == TYPEID_PLAYER)
+            pos = { GetPositionX(), GetPositionY(), GetPositionZ() + 2.0f };
+        else
+            pos = GetHitSpherePointFor({ ox, oy, oz });
+
+        return GetMap()->isInLineOfSight(pos.GetPositionX(), pos.GetPositionY(), pos.GetPositionZ(), ox, oy, oz, GetPhaseMask());
```
Treeston: Shouldn't you use modified position for non-player creature casters too?

src/server/game/Entities/Object/Object.h
```diff
@@ -487,6 +487,7 @@ class WorldObject : public Object, public WorldLocation
         bool IsWithinDistInMap(WorldObject const* obj, float dist2compare, bool is3D = true) const;
         bool IsWithinLOS(float x, float y, float z) const;
         bool IsWithinLOSInMap(WorldObject const* obj) const;
+        Position GetHitSpherePointFor(Position const& dest) const;
```
Treeston: Might want to consider providing an overload like
`void GetHitSpherePointFor(Position const& dest, float& x, float& y, float& z) const;`
as well for use cases (like your own) that want to work directly with coords instead of having to construct temporary position objects.

src/server/game/Entities/Object/Object.cpp
Golrag: Ehm, I don't get it :x
Which position are you referring to ? ox, oy, oz or 'pos' ? I assume modified position is position get from GetHitSpherePointFor

Treeston: ox/oy/oz

Golrag: These should be the modified position, if it's called from ```bool WorldObject::IsWithinLOSInMap(const WorldObject* obj) const``` if it's used directly then I assume that that position is fixed and not related to any worldobject :p

Treeston: Alright, nevermind then.

src/server/game/Entities/Object/Object.h
Golrag: Done !

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1232,11 +1236,36 @@ bool WorldObject::IsWithinLOS(float ox, float oy, float oz) const
...
     return true;
 }
 
+Position WorldObject::GetHitSpherePointFor(Position const& dest) const
+{
+    G3D::Vector3 vThis(GetPositionX(), GetPositionY(), GetPositionZ());
+    G3D::Vector3 vObj(dest.GetPositionX(), dest.GetPositionY(), dest.GetPositionZ());
+    G3D::Vector3 contactPoint = vThis + (vObj - vThis).directionOrZero() * GetObjectSize();
+
+    return Position(contactPoint.x, contactPoint.y, contactPoint.z);
```
ccrs: orientation?

Treeston: Isn't needed in this scenario, but I suppose specifying `GetAngle` or something wouldn't hurt, agree.

Golrag: Updated :)

Keader:
LoS dont work anymore o.O
You can use spell in targets out of LoS after last commit

Golrag:
Any examples ? I walked around orgrimmar quickly and everything seemed fine :/

Keader:
In my core, in arenas like Circle of Blood spell can be used in targets behind walls (pillars)

Greymane:
I can confirm what @Keader said, @Golrag after that patch LoS does not work anymore, you can try on any arena.

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1146,9 +1146,13 @@ bool WorldObject::IsWithinLOSInMap(const WorldObject* obj) const
     if (!IsInMap(obj))
         return false;
 
-    float ox, oy, oz;
-    obj->GetPosition(ox, oy, oz);
-    return IsWithinLOS(ox, oy, oz);
+    float x, y, z;
+    if (obj->GetTypeId() == TYPEID_PLAYER)
+        GetPosition(x, y, z);
```
Greymane: Shouldnt it be `obj->GetPosition(x,y,z);` instead?

Golrag: Nice catch @Greymane , haven't got the time to test it a lot right now. Will take a loot at this PR next week. Will commit this change ASAP tho :)

Greymane: Nice :D

sirikfoll:
@Golrag The LoS issue mentioned above was fixed, but there's a new one, go to `.go xyz -553 2200 51 631` and use death knight's spell Death and Decay, the spell will hit npcs that are on the upper floor, this seems to also occur to similar spells.

sirikfoll:
I'm using this PR for a while, and everything seems to be working just fine ^^

Aokromes:

Treeston:
@Golrag anything you feel is missing?

Golrag:
@Treeston Not really. Thanks for all the help :)

Treeston:
Alright, I will review + merge later today then.

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1232,11 +1236,36 @@ bool WorldObject::IsWithinLOS(float ox, float oy, float oz) const
...
     if (IsInWorld())
-        return GetMap()->isInLineOfSight(GetPositionX(), GetPositionY(), GetPositionZ()+2.f, ox, oy, oz+2.f, GetPhaseMask());
+    {
+        float x, y, z;
+        if (GetTypeId() == TYPEID_PLAYER)
+            GetPosition(x, y, z);
+        else
+            GetHitSpherePointFor({ ox, oy, oz }, x, y, z);
+
+        return GetMap()->isInLineOfSight(x, y, z + 2.0f, ox, oy, oz + 2.0f, GetPhaseMask());
```
Treeston: Hmm. Unsure about this. I think we want to offset `+2.0f` _first_, and _then_ find contact point.

Golrag: Isn't the result the same ? Adding 2.f to both of the sphere's z position won't change the x,y position or am I mistaken ?

Treeston: Right, good point. We add it for both positions so the angle doesn't change.

SeTM:
It's wrong commit in all? 
For example:
Sindra is in LoS: http://i.prntscr.com/bbe2fbe3598745eca1388517b4080d79.jpeg
Sindra is NOT IN LoS: http://i.prntscr.com/3f4b67d160064fa59fd066dbb67aeec4.jpeg

It's OK?

SeTM:
Also, the coordinates of the model can be calculated at any wall, so check LoS fail, although we are directly in the model.

Treeston:
Pretty sure Sindragosa still has some weird hacks in the script itself that might be messing with it.

Wall thing is a good point though, @Golrag what happens if spheres overlap?

SeTM:
With Sindragosa I checked only by its model, on a dummy, using the command "debug los", so that no scripts here are not affected.

This is not normal when we see (LoS) the boss of the tomb, and do not see BEFORE tomb.

Golrag:
Hm... @SeTM in your first screenshot, Should sindragosa be in LoS or not ? For the second one, I think the best way to handle this is to return true on LoS when you are close enough (in objectsize range).

Can't commit/PR a fix tho, not enough free time to test things :(

SeTM:
First screen is ok, Sindragosa should be in LoS.

PR #14570, opened 2015-04-18
Core/Battleground: Removed _CheckSafePositions, this is handled by areatriggers now.

Golrag:
Removed _CheckSafePositions, this is handled by areatriggers now.

Any feedback is welcome ! :)

About the areatriggers for starting areas, should I move them to Battleground::HandleAreatrigger or handle them in the specific battlegroundclass ? (Like BattlegroundTP::HandleAreatrigger)

Edit: Updated to https://github.com/TrinityCore/TrinityCore/commit/007ab5b801fd173d35dc6ba7e07b4015e08a4c43

Added missing alterac valley areatriggers
Removed m_StartMaxDist

PR #15841, opened 2015-11-09
Core/Battleground: Removed _CheckSafePositions, this is handled by areatriggers now

Golrag:
Made a mistake in https://github.com/TrinityCore/TrinityCore/pull/14570 when rebase, had to do it again :p

### DIFF
--- a/sql/updates/world/yyyy_mm_dd_xx.sql
+++ b/sql/updates/world/yyyy_mm_dd_xx.sql
@@ -1,0 +1,4 @@
+ALTER TABLE `battleground_template`
+	DROP COLUMN `StartMaxDist`;
+
+UPDATE `battleground_template` SET `MaxLvl`=100;
--- a/src/server/game/Battlegrounds/Battleground.cpp
+++ b/src/server/game/Battlegrounds/Battleground.cpp
@@ -151,7 +151,6 @@ Battleground::Battleground()
 
     m_MapId             = 0;
     m_Map               = NULL;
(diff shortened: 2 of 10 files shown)


Shauren:
https://github.com/TrinityCore/TrinityCore/pull/14570#issuecomment-120704315
^ Still wanting this

soulfrost:
@Shauren if i had understand - u care about modfied wow.exe where cmsg_area_trigger can be disabled. what about making some protect for this in other way ( not old one, cos it's hard for cpu ),
client sending needed areaTrigger ids all time on player entering bgs, arenas  : make boolean variable for player and set it to true if server get cmsg_area_trigger from player on bg entering in first 15 seconds f.e., if no? kick him from server.

Warpten:
no. i can forge packet.

Kinzcool:
By the way, maybe 2 years ago, I wanted to try out loading a modified AreaTrigger.dbc file into the client to specifically exploit this flaw. I didn't do it in the end, but the point I wanted to bring is that I did glitch and do "early starts", and I wasn't teleported back to the spawn place.

jackpoz:
This PR is not going to be merged till the old _CheckSafePositions() checks are added back due to possible exploits

Golrag:
Updated https://github.com/Golrag/TrinityCore/commit/941574c4015d7822ce68579f4e5c1be58360c9c4

Aokromes:
Cannot be reopen.

Issue #15833, opened 2015-11-08
[Crash] related to extractQuotedArg


mthsena:
Rev. 7d5e7102cdd407846fa340892df39ea49938ee69

How to reproduce:
Use .guild rename and do not put the second argument
```Example: .guild rename "guildname"```

it happens to all the commands than need two arguments

http://pastebin.com/G0mwzDgm

jackpoz:
do you have any custom change applied ?

mthsena:
No, clean build, you cannot reproduce ?

PR #15847, opened 2015-11-10
Core/Scripts: Fix crash related to extractQuotedArg

Rochet2:
strtok may return null which crashes in the `ChatHandler::extractQuotedArg`.
See https://github.com/TrinityCore/TrinityCore/blob/c92faf627712d26a7abdce2db62ffb8b938a4522/src/server/scripts/Commands/cs_guild.cpp#L212

### DIFF
--- a/src/server/game/Chat/Chat.cpp
+++ b/src/server/game/Chat/Chat.cpp
@@ -1131,7 +1131,7 @@ void ChatHandler::extractOptFirstArg(char* args, char** arg1, char** arg2)
 
 char* ChatHandler::extractQuotedArg(char* args)
 {
-    if (!*args)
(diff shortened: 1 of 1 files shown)


Shauren:
This is definitely a valid fix but before we merge it there is something that needs to be decided
whether it should be checked inside like this or it should be callers responsibility

tkrokli:
@Rochet2 : since your original PR title was "Fix TrinityCore#15833", you may want to include the following text on a single line at the bottom of your commit text and PR description:
`Closes #15833`

(the intention is to trigger TrinityCoreBot to close issue 15833 when this PR gets merged)

Aokromes:
abd4e2e

PR #15881, opened 2015-11-20
Fix build // Update to gcc 4.9 and g++ 4.9 and clang to 3.6

imbavirus:
The warning causing the build issue is a bug from gcc 4.8 and has been fixed in gcc 4.9, without the clang update I got another build error, but the build passed by updating all 3 of these.


Vincent-Michael:
no, we support still clang 3.5 / gcc 4.8

Carbenium:
clang 3.6 / gcc 4.9 is tested on an internal jenkins instance

PR #15883, opened 2015-11-22
Wyrmrest taxi gossip

mysterious-static:
Fixes gossip options for Wyrmrest Temple taxi flights between levels.


Aokromes:
Better convert data to read from db.

velinath:
Ah, unfortunately I've no idea about where to start from that direction. Is there an example of this that I can take a look at to work from? Would be happy to take a stab at it but just unsure where to start looking as I am pretty unfamiliar with c++

Aokromes:
https://github.com/TrinityCore/TrinityCore/commit/6c0ff5717f14e90dfd92cf963c0805adae3c668b

velinath:
Thank you very much. I'll take a look at that and see what I can do. Could be nothing, could be I'm back here in a week, either way I'm sure I'll have learned something!

PR #15889, opened 2015-11-23
Core/MMaps: Fixed mmap loading after recent format change

Zedron:
bug introduced in 1d2aafd


jackpoz:
which recent format change ? A PR description would be nice...

Zedron:
edited...

Zedron:
hmm my bad it was actually 1d2aafd. that makes more sense. edited.

PR #15891, opened 2015-11-24
DB/Misc: Fix wrong npcflag's

F03SD:
This small fixes wrong npc flags on start location's.
Before fix:
wowscrnshot_112415_100434


F03SD:
- "sql/updates/characters/2015_11_23_00_characters.sql" is missing in the 'characters'.'updates' table.
Fatal error:
  The Database Updater is broken for the 'characters' database,
  because the applied update is missing in the 'characters'.'updates' table.

Aokromes:
Thx but we don't use PRs for sql fixes, open a ticket for it.

PR #15904, opened 2015-11-27
Core/Player: cleanup function CheckInstanceLoginValid

ForesterDev:
CleanUP and optimize checks added in https://github.com/TrinityCore/TrinityCore/commit/5f917e2286809d7aafa8871be38042ae7966c26b

### DIFF
--- a/sql/updates/world/2015_11_27_00_world_355.sql
+++ b/sql/updates/world/2015_11_27_00_world_355.sql
@@ -1,0 +1,3 @@
+DELETE FROM `trinity_string` WHERE `entry` = 11010;
+INSERT INTO `trinity_string` (`entry`,`content_default`) VALUES
+(11010, 'You didn\'t get kicked out of the instance even if Player::CheckInstanceLoginValid() returned false and without .gm on flag');
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -2563,7 +2563,7 @@ void Player::SetGameMaster(bool on)
 
 bool Player::CanBeGameMaster() const
 {
-    return m_session && m_session->HasPermission(rbac::RBAC_PERM_COMMAND_GM);
+    return GetSession()->HasPermission(rbac::RBAC_PERM_COMMAND_GM);
 }
 
 void Player::SetGMVisible(bool on)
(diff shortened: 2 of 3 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -18672,31 +18672,31 @@ bool Player::Satisfy(AccessRequirement const* ar, uint32 target_map, bool report
...
 
     // do checks for satisfy accessreqs, instance full, encounter in progress (raid), perm bind group != perm bind player
-    return sMapMgr->CanPlayerEnter(map->GetId(), this, true) || IsInstanceLoginGameMasterException();
+    return sMapMgr->CanPlayerEnter(map->GetId(), this, true);
 }
 
 bool Player::IsInstanceLoginGameMasterException() const
 {
-    if (CanBeGameMaster())
+    if (IsGameMaster() || CanBeGameMaster())
```
Shauren: With CanBeGameMaster you don't need to check IsGameMaster - CanBeGameMaster() checks if you have the rbac permission to use .gm on, while IsGameMaster() returns true if you used .gm on

src/server/game/Entities/Player/Player.cpp
```diff
@@ -18672,31 +18672,31 @@ bool Player::Satisfy(AccessRequirement const* ar, uint32 target_map, bool report
 
 bool Player::CheckInstanceLoginValid(Map* map)
 {
-    if (!map->IsDungeon() || IsGameMaster())
+    if (!map->IsDungeon() || IsInstanceLoginGameMasterException())
```
jackpoz: Isn't this going to send the ingame message even if the player is actually in a raid/group ?

Please provide a list of the tests you did ingame.

ForesterDev: Yes, player will receive message even if he is in raid group. But we can send this warning-message only if player is not in group, if needed, just add (!GetGroup()) check in IsInstanceLoginGameMasterException

src/server/game/Entities/Player/Player.cpp
```diff
@@ -2563,7 +2563,7 @@ void Player::SetGameMaster(bool on)
 
 bool Player::CanBeGameMaster() const
 {
-    return m_session && m_session->HasPermission(rbac::RBAC_PERM_COMMAND_GM);
+    return GetSession()->HasPermission(rbac::RBAC_PERM_COMMAND_GM);
```
ccrs: is this safe?

Shauren: Yes, player session will never be null.

src/server/game/Entities/Player/Player.cpp
jackpoz: That will duplicate the group checks, the idea behind IsInstanceLoginGameMasterException() is to replace any previous "return false;" in Player::CheckInstanceLoginValid() .
I'd like to merge this PR without the changes in Player::CheckInstanceLoginValid(), the rest of the PR is ok

sql/updates/world/2015_11_27_00_world_355.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM `trinity_string` WHERE `entry` = 11010;
```
jackpoz: this file should be called 335 btw, not 355

PR #15912, opened 2015-11-28
Core/Util: Add a urandms(min,max) helper...

Treeston:
…to make urand(min_IN_MILLISECONDS, max_IN_MILLISECONDS) shorter to write for random time intervals (boss scripts etc).

### DIFF
--- a/src/common/Utilities/Util.h
+++ b/src/common/Utilities/Util.h
@@ -82,6 +82,9 @@ int32 irand(int32 min, int32 max);
 /* Return a random number in the range min..max (inclusive). */
 uint32 urand(uint32 min, uint32 max);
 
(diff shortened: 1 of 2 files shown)


Vincent-Michael:
why no "ASSERT(max >= min)"?

src/common/Utilities/Util.h
```diff
@@ -21,6 +21,7 @@
 
 #include "Define.h"
 #include "Errors.h"
+#include "Common.h"
```
jackpoz: Util.cpp already include Common.h , please move the function implementation in the cpp

Treeston: Will inline still be handled properly if definition is in the cpp? Don't want to introduce extra call overhead, and the likelyhood of something using Util.h without using Common.h is practically nonexistant.

jackpoz: please let's not start with micro-optimization and compiler optimization talks...

PR #15947, opened 2015-12-06
Core/Spell: fixed possible exploit in Duel Reset script

FrancescoBorzi:
https://github.com/TrinityCore/TrinityCore/pull/15731#issuecomment-162257714

### DIFF
--- a/src/server/game/Spells/SpellHistory.cpp
+++ b/src/server/game/Spells/SpellHistory.cpp
@@ -695,9 +695,7 @@ void SpellHistory::RestoreCooldownStateAfterDuel()
         // check for spell with onHold active before and during the duel
         for (auto itr = _spellCooldownsBeforeDuel.begin(); itr != _spellCooldownsBeforeDuel.end(); ++itr)
         {
-            if (!itr->second.OnHold &&
-                _spellCooldowns.find(itr->first) != _spellCooldowns.end() &&
-                !_spellCooldowns[itr->first].OnHold)
+            if (!itr->second.OnHold && !_spellCooldowns[itr->first].OnHold)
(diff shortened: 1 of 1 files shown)


jackpoz:
what about https://github.com/TrinityCore/TrinityCore/pull/15731#issuecomment-162269065 ?

ShinDarth:
@jackpoz read the whole discussion,

this PR is just a revert of a small part of that commit

jackpoz:
the whole discussion ends with
> @ShinDarth Now it's just visual bug, better than old one xD
> 
> Thanks

which means there is still a bug, so what's about it ?

ShinDarth:
@jackpoz  yes after this PR there will be a visual bug (resolving the exploit).

About the visual bug I said [that](https://github.com/TrinityCore/TrinityCore/pull/15731#issuecomment-162258667), explaining that I'm not sure about how to resolve that visual bug without introducing worse annoying bugs.

Maybe we can meanwhile merge this PR and resolve the exploit at least?

Asterc:
That's a tricky thing, another exploit.
If you use an ability that provides a buff, like Starfall, then start the duel, Starfall will already be active and you'll be able to use it again, since the cooldown will be reset.
2 starfalls in a row '-'

Wouldn't be better to let things simpler and only reset the cooldowns after the duel?

jackpoz:
@Asterc can you also copy your message to a new issue ? so it doesn't get lost

ShinDarth:
No it wouldn't be better to only reset the cooldowns after the duel, because it would introduce a very worse exploit: have a cooldown? Just do a duel to reset it.

So **we do need** to restore the cooldowns the player had just before the duel.

Also, that is not really a problematic exploit. The opponent player can see clearly if the player just used some spells which have a meaning cooldown.

PR #15960, opened 2015-12-08
Misc: Removed unneeded info about character escaping from .conf.dist …

panaut0lordv:
…and improved example MotD

### DIFF
--- a/src/server/worldserver/worldserver.conf.dist
+++ b/src/server/worldserver/worldserver.conf.dist
@@ -1040,9 +1040,8 @@ BeepAtStart = 1
 
 #
 #    Motd
-#        Description: Message of the Day, displayed at login.
-#                     Use '@' for a newline and be sure to escape special characters.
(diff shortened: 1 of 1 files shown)


jackpoz:
what's the deal with character escaping no longer needed ? the PR doesn't specify anything about it

tkrokli:
The weird thing about using apostrophes in the MOTD message in Worldserver.conf is that for some people the backslash will be visible in-game, while for others it will work just as intended. My 3.3.5 compiles seem to work with the documented method, but I have not compared with 6.x yet.

panaut0lordv:
In case your config needs commenting, please specify OS, architecture and boost lib version.

jackpoz:
tried on Win 10 X64 and escaping the character \' displayed it like \' ingame so escaping is unneeded in this configuration

jackpoz:
same behavior on Ubuntu 15.10 (335 branch at least)

PR #15996, opened 2015-12-14
Core/Player: Update combo points for 6.x

myran2:
Since 6.x, combo points are treated like any other power.
I tried to reflect this in the code by removing a lot of combo point specific stuff.
I was pretty conservative with my changes so there might be some other stuff that should be removed.


src/server/game/Entities/Player/Player.h
```diff
@@ -1660,13 +1660,9 @@ class Player : public Unit, public GridObject<Player>
         void SetTarget(ObjectGuid const& /*guid*/) override { } /// Used for serverside target changes, does not apply to players
         void SetSelection(ObjectGuid const& guid) { SetGuidValue(UNIT_FIELD_TARGET, guid); }
 
-        uint8 GetComboPoints() const { return m_comboPoints; }
```
DDuarte: Keep this function but make it call `GetPower(POWER_COMBO_POINTS)` to avoid having to replace all the `GetComboPoints()` calls with GetPower() (also breaking the "API")

sarnex: I wouldn't have made that mistake.

DDuarte: @sarnex: I'm not quite sure what you are referring too but this is not a contest.

sarnex: I won

myran2:
Any ideas why the build is failing? Looking at the compile log all I can see is
'clang: fatal error: unable to execute command: Killed' which isn't all that helpful.

Shauren:
Merged

Issue #15089, opened 2015-07-14
[3.3.5][Core/Spell] SPELL_ATTR1_NO_THREAT + SPELL_ATTR3_NO_INITIAL_AGGRO


Rushor:
related to https://github.com/TrinityCore/TrinityCore/commit/dc1a015d0b48f8684aa4174b9b84eba910467a48

There must be a new Checkmethode to check either for SPELL_ATTR1_NO_THREAT or SPELL_ATTR3_NO_INITIAL_AGGRO. If this Checkmethode is implemented, we can remove the 'db-hack'

If any of these Attributes is given in spelldbc, the spell should not generate any threat

Possible fix by @joschiwald: https://gist.github.com/joschiwald/2f5564d680fc74a55ff8

This fix needs some testing

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- Could not find branch name (3.3.5? 6.x?)

Please read http://bit.ly/tc-issuetracker-and-you . Thanks.

Killyana:
Some time when you use an item to control a npc or when you ride a vehicle which use a spell using control unit effect, the npc or the vehicle will attacks you once the control expired or when you dismount, I think it's related.

robinsch:
SPELL_ATTR1_NO_THREAT != SPELL_ATTR3_NO_INITIAL_AGGRO imo.

joschiwald:
pls post the spellid too

Killyana:
Example of spells used to ride a vehicle:
http://wotlk.openwow.com/spell=49078
http://wotlk.openwow.com/spell=48290
using: AttributesEx1: 0x00000400 (SPELL_ATTR1_NO_THREAT)

LuqJensen:
there are lots of issues with threat/aggro. a lot of dots/hots trigger talents which starts combat, when it shouldnt

Keader:
Any news about this?

joschiwald:
b4b43d03b3fbff38ebcf84be27388202fc6cb08f

robinsch:
I don't think this is correct, because of spells stated in https://github.com/TrinityCore/TrinityCore/issues/15089#issuecomment-121410868

joschiwald:
taunts have a special spelleffect that generates threat

PR #16020, opened 2015-12-16
CreatureAI: Adjust AI behavior when a charm ends

Treeston:
CreatureAI: Adjust AI behavior when a charm ends:
- OnCharmed will be invoked on the creature's base (non-charmed) AI, allowing the AI to react to the creature no longer being charmed.

### DIFF
--- a/src/server/game/AI/CoreAI/PassiveAI.cpp
+++ b/src/server/game/AI/CoreAI/PassiveAI.cpp
@@ -58,6 +58,12 @@ void PossessedAI::KilledUnit(Unit* victim)
         victim->RemoveFlag(UNIT_DYNAMIC_FLAGS, UNIT_DYNFLAG_LOOTABLE);
 }
 
+void PossessedAI::OnCharmed(bool /*apply*/)
+{
+    me->NeedChangeAI = true;
+    me->IsAIEnabled = false;
+}
+
 void CritterAI::DamageTaken(Unit* /*done_by*/, uint32&)
 {
     if (!me->HasUnitState(UNIT_STATE_FLEEING))
--- a/src/server/game/AI/CoreAI/PetAI.cpp
+++ b/src/server/game/AI/CoreAI/PetAI.cpp
@@ -588,6 +588,12 @@ void PetAI::ReceiveEmote(Player* player, uint32 emote)
         }
 }
 
+void PetAI::OnCharmed(bool /*apply*/)
+{
+    me->NeedChangeAI = true;
+    me->IsAIEnabled = false;
+}
+
 void PetAI::ClearCharmInfoFlags()
 {
     // Quick access to set all flags to FALSE
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -29,11 +29,13 @@
 #include "Language.h"
 
 //Disable CreatureAI when charmed
-void CreatureAI::OnCharmed(bool /*apply*/)
+void CreatureAI::OnCharmed(bool apply)
 {
-    //me->IsAIEnabled = !apply;*/
-    me->NeedChangeAI = true;
-    me->IsAIEnabled = false;
+    if (apply)
+    {
+        me->NeedChangeAI = true;
+        me->IsAIEnabled = false;
+    }
 }
 
 AISpellInfoType* UnitAI::AISpellInfo;
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -653,8 +653,8 @@ void SmartAI::OnCharmed(bool apply)
 {
     GetScript()->ProcessEventsFor(SMART_EVENT_CHARMED, NULL, 0, 0, apply);
 
-    if (!apply && !me->IsInEvadeMode() && me->GetCharmerGUID())
-        if (Unit* charmer = ObjectAccessor::GetUnit(*me, me->GetCharmerGUID()))
+    if (!apply && !me->IsInEvadeMode())
+        if (Unit* charmer = me->GetCharmer())
             AttackStart(charmer);
 }
 
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13712,6 +13712,9 @@ void Unit::UpdateCharmAI()
                     delete i_AI;
                     i_AI = i_disabledAI;
                     i_disabledAI = nullptr;
+
+            if (GetTypeId() == TYPEID_UNIT)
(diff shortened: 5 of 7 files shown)


Killyana:
It's not related to SPELL_ATTR1_NO_THREAT and SPELL_ATTR3_NO_INITIAL_AGGRO ? https://github.com/TrinityCore/TrinityCore/issues/15089

Treeston:
Pretty much any issue with smartai mobs behaving weirdly while MCed relates to this. I didn't look.

src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -650,9 +650,7 @@ void SmartAI::OnCharmed(bool apply)
 {
     GetScript()->ProcessEventsFor(SMART_EVENT_CHARMED, NULL, 0, 0, apply);
 
-    if (!apply && !me->IsInEvadeMode() && me->GetCharmerGUID())
-        if (Unit* charmer = ObjectAccessor::GetUnit(*me, me->GetCharmerGUID()))
-            AttackStart(charmer);
+    CreatureAI::OnCharmed(apply);
```
ccrs: I debated this with @Shauren some time ago, and this is not ok.

He claimed that AI change should not be done (it will prevent smartscripts to trigger)

I dont know the right solution, but this is what i came up with:

* Enable AI change only on CHARM_TYPE_CHARM
This is like a middle term solution, but needs also to check for 
- If creature is vehicle
- if creature is possessed
If charm is one of those cases just store the ReactState to fix the aggro behaviour
Else use the ````CreatureAI::OnCharmed(apply)```` to change into the PetAI

* Enable the AI change just as u did
Prevents any smartscript to trigger, and there are some creatures that use them while charmed
(texts for example)

* Implement PetAI functionallities inside the SmartAI
That would be the perfect solution but a huuuuuuuuuge load of work:
- Autocast
- Follow
- Disable events with cast action (or just disable cast without a special flag)
- ...

src/server/game/AI/CoreAI/PetAI.cpp
```diff
@@ -597,6 +597,12 @@ void PetAI::ReceiveEmote(Player* player, uint32 emote)
         }
 }
 
+void PetAI::OnCharmed(bool /*apply*/)
+{
```
ccrs: not needed, It will call ````CreatureAI::OnCharmed(apply)````

src/server/game/AI/CoreAI/PassiveAI.cpp
```diff
@@ -58,6 +58,12 @@ void PossessedAI::KilledUnit(Unit* victim)
         victim->RemoveFlag(UNIT_DYNAMIC_FLAGS, UNIT_DYNFLAG_LOOTABLE);
 }
 
+void PossessedAI::OnCharmed(bool /*apply*/)
+{
```
ccrs: not needed, It will call ````CreatureAI::OnCharmed(apply)````

src/server/game/AI/CreatureAI.cpp
```diff
@@ -28,11 +28,13 @@
 #include "CreatureTextMgr.h"
 
 //Disable CreatureAI when charmed
-void CreatureAI::OnCharmed(bool /*apply*/)
+void CreatureAI::OnCharmed(bool apply)
 {
-    //me->IsAIEnabled = !apply;*/
-    me->NeedChangeAI = true;
-    me->IsAIEnabled = false;
+    if (apply)
```
ccrs: If u add this, u will need all the other overrides, so better to have this just in one place :)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -13690,6 +13690,9 @@ void Unit::UpdateCharmAI()
             delete i_AI;
             i_AI = i_disabledAI;
             i_disabledAI = NULL;
+
+            if (GetTypeId() == TYPEID_UNIT)
```
ccrs: inside RemoveCharmBy this call is made, so this will cause errors

Treeston: That is the point. RemoveCharmBy's call is handled by the charmed AI, while this call will be handled by the creature's _original_ AI.

src/server/game/AI/SmartScripts/SmartAI.cpp
Treeston: With all due respect, any smartscript mob is currently a huge pile of steaming bugs when mind controlled. They actively fight player control, cast spells without player input, and move without player input.

If you weigh this against the questionable benefit of having smart scripts still trigger while controlled (how many creatures with elaborate smartscripts can you think of that can be targeted by such abilities), there should be no question as to which is infinitely preferable.

PS: Your fix does not work. Mind Control is a possess, and is by far the most broken type of charm for SmartAI.

src/server/game/AI/CreatureAI.cpp
Treeston: You completely missed the point of the PR. The point is that the original creatureAI will now also get an OnCharmed invocation with apply=false, thus meaning the base CreatureAI has to check whether this flag is true or false before initiating the AI change...

src/server/game/AI/SmartScripts/SmartAI.cpp
ccrs: I totally agree with u, SAI Charm mechanics are incomplete.

My "solution" works. There are several important points:
* If its one of the cases in which no PetAI is needed, just store the ReactState on charm and restore on remove. On a clean TC it works, I've tested it.
* SmartScripts on charms without PetAI continue and events like casts or texts do trigger. Example: http://www.wowhead.com/item=38302/ruby-beacon-of-the-dragon-queen
* SmartScripts on charms which replace SAI for PetAI dont trigger. Like I said this is the incorrect point of this solution, to avoid handling PetAI inside SAI.

The perfect solution is still the last one I wrote down, implement PetAI inside SAI.
Like I said, huge effort for a small group of cases; but still the right thing to do.


To handle events that should not be triggered while charm, for example, we could implement a flag like ````SMART_EVENT_FLAG_TRIGGER_ONCHARM = 512````

src/server/game/Entities/Unit/Unit.cpp
ccrs: Oh, now I get it. Right when the AI change is restored, u call the oncharm hook. Smart.

Treeston:
This PR introduces some issues in conjunction with controlled pets for drakuru quest that use SmartAI. Labeling WIP.

ccrs:
There are many charmed creatures that have SAI to trigger credit, texts, casts,... Just like I told you xd

Treeston:
Removed SmartAI changes from the scope of this PR.

Issue #16040, opened 2015-12-17
[3.3.5] Penance triggering a cooldown if facing the wrong way


Treeston:
Branch 3.3.5, commit 0bb2ba7b35c7cd136b4371f6d284eba0bd34b6e5

Repro:
1. Target a hostile
2. Cast Penance (53007) at the target
3. Penance fails as you are facing the wrong way, but still triggers a cooldown

Expected:
Penance does not trigger a cooldown in step 3

P-Kito:
```
            SpellCastResult CheckCast()
            {
                Player* caster = GetCaster()->ToPlayer();
                if (Unit* target = GetExplTargetUnit())
                {
                    if (!caster->IsFriendlyTo(target))
                    {
                        if (!caster->IsValidAttackTarget(target))
                            return SPELL_FAILED_BAD_TARGETS;
                        if (!caster->isInFront(target))
                            return SPELL_FAILED_UNIT_NOT_INFRONT;
                    }
                }
                return SPELL_CAST_OK;
            }
```

Treeston:
Interesting thing: Penance Rank 1 (47540) and Rank 2 (53005) make you turn to face the target if you cast at a hostile.
Other ranks (53006 or 53007) do not.

PR #16066, opened 2015-12-19
Spells/Penance: Fix erroneous cooldown on misplaced hostile target

Treeston:
Spells/Penance: If hostile target is not in front, now fails in the initial cast rather than the triggered cast, thus not erroneously causing a cooldown.

Thanks @P-Kito


Treeston:
Closes #16040

PR #16082, opened 2015-12-22
DB: Fix a DB error.

ghost:
Error correction, if wrong to cancel.


MitchesD:
I don't see anything wrong, it's normal actionlist

Killyana:
What's the DBerror?

DevRival:
Why are we then use entryorguid = 2229300?

Aokromes:
we don't edit existing sql files unless they can't be applied, also, we don't use pull requests for sql only fixes.

MitchesD:
because it's actionlist? it means entry * 100

DevRival:
Oh God, I did not see( excuse the stupidity(

PR #16089, opened 2015-12-23
[3.3.5] AI/CreatureAI: Complete re-write of evade boundary system

Treeston:
Alright, I finally found time to tackle something that I've been meaning to do for a while - fixing the heavily broken/practically nonexistant evade boundary system for bosses by replacing it with something that fulfills the following:
- Easy to adapt for new instances being added. No fiddling in the boss script in 99% of cases.
- Not needlessly restricted to bosses only. Could be adapted for any creature.
- Robust, yet simple. Can fit (almost) any shape of boss room.
- Easily extendable. Adding support for additional kinds of boundaries is simple.

Allow me to provide an outline of important files:
1. AreaBoundary: The meat of the new system. The virtual AreaBoundary class defines a basic interface that allows checking whether a Position (note that Unit and friends are all subclasses of Position) is "inside" or "outside" of its object. Its implementing subclasses then add support for basic geometric shapes that I feel should cover most boss rooms. Any AreaBoundary can also be inverted (a boolean flag) which makes it match elements that are _not_ inside the object.
2. CreatureAI: Implements the higher-level logic of checking in bounds. A Position is considered to be "in bounds" if it matches all boundaries on the CreatureAI in question - inverted boundaries can be used as exclusions (allowing one to form more complex shapes by combining basic shapes). Specific AI implementations can override this if necessary.
3. InstanceScript: Implement a simple loader method for loading boundary information for the instance's bosses. Allows one to maintain a simple initializer list of boundaries in the instance script.
4. Creature.cpp: Periodically invoke the boundary check (every 2.5 seconds while in combat).
5. cs_debug.cpp: Implement a `.debug boundary` command that flood fills the targeted creature's boundary and marks the edge. This makes for quick and painless testing and rapid iterating on boss boundaries.


src/server/game/AI/CreatureAI.cpp
```diff
@@ -271,6 +271,112 @@ bool CreatureAI::_EnterEvadeMode()
...
+        me->Talk("VisualizeBoundary warning: Current position is out of bounds", CHAT_MSG_MONSTER_YELL, LANG_UNIVERSAL, 50000.0f, nullptr);
+        return;
+    }
+    bool hadBoundsWarning = false;
+    Q.push({ 0,0 });
+    while (!Q.empty())
+    {
+        coordinate front = Q.front();
+        bool hasOutOfBoundsNeighbor = false;
+        for (coordinate off : std::list<coordinate>({ {1, 0}, { 0,1 }, { -1,0 }, { 0,-1 } }))
```
DDuarte: Replace this list by an array and create it outside of the loop.

Shauren:
why the doubles? position is a float everywhere (and in the packets)

src/server/game/AI/CreatureAI.cpp
```diff
@@ -271,6 +271,112 @@ bool CreatureAI::_EnterEvadeMode()
...
+void CreatureAI::VisualizeBoundary(uint32 duration, Unit* owner, bool fill) const
+{
+    typedef std::pair<int32, int32> coordinate;
+    
+    if (!owner)
+        return;
+
+    if (!_boundary || _boundary->empty())
+    {
+        me->Talk("VisualizeBoundary error: Creature movement is unbounded", CHAT_MSG_MONSTER_YELL, LANG_UNIVERSAL, 50000.0f, nullptr);
```
Naios: Move this to TC_LOG maybe? Error messages shouldn't be leaked to users.

src/server/game/AI/CreatureAI.cpp
```diff
@@ -271,6 +271,112 @@ bool CreatureAI::_EnterEvadeMode()
...
+        return;
+    }
+
+    std::queue<coordinate> Q;
+    std::unordered_set<coordinate> alreadyChecked;
+    std::unordered_set<coordinate> outOfBounds;
+
+    float meX = me->GetPositionX(), meY = me->GetPositionY(), meZ = me->GetPositionZ(), spawnZ = meZ + BOUNDARY_VISUALIZE_SPAWN_HEIGHT;
```
Naios: Prefer `Position` or `g3d::Vec3D` over x, y, z

src/server/game/AI/CreatureAI.cpp
```diff
@@ -271,6 +271,112 @@ bool CreatureAI::_EnterEvadeMode()
...
+    
+    if (!CheckBoundary(me))
+    {
+        me->Talk("VisualizeBoundary warning: Current position is out of bounds", CHAT_MSG_MONSTER_YELL, LANG_UNIVERSAL, 50000.0f, nullptr);
```
Naios: Same as

> Move this to TC_LOG maybe? Error messages shouldn't be leaked to users.

Treeston:
Numerical precision mid-operation. If your coordinates are magnitude 10⁴, and you then calculate a scalar product, then you start to lose precision on the lower digits (which you're then missing when you subtract the values).

src/server/game/AI/CreatureAI.cpp
Treeston: @Shauren Numerical precision mid-operation. If your coordinates are magnitude 10⁴, and you then calculate a scalar product, then you start to lose precision on the lower digits (which you're then missing when you subtract the values).

Treeston: @naio

Treeston: @Naios it's from a .debug command, but agree that it should use a different method. Something that is displayed in client would be preferred though. Any ideas?

src/server/game/AI/CreatureAI.cpp
```diff
@@ -271,6 +271,112 @@ bool CreatureAI::_EnterEvadeMode()
...
+                if (!hadBoundsWarning)
+                {
+                    hadBoundsWarning = true;
+                    me->Talk("VisualizeBoundary warning: Reached fail-safe flood boundary - check is your boundary is unbounded!", CHAT_MSG_MONSTER_YELL, LANG_UNIVERSAL, 50000.0f, nullptr);
+                }
+                continue;
+            }
+            if (alreadyChecked.find(next) == alreadyChecked.end()) // never check a coordinate twice
+            {
+                Position nextPos(meX + next.first*BOUNDARY_VISUALIZE_STEP_SIZE, meY + next.second*BOUNDARY_VISUALIZE_STEP_SIZE, meZ);
```
Naios: As mentioned above if you replace the hardcoded `float x,y, z` through a Vector3 you could do scalar multiplication such as
```c++
Vector3D vec1 = ...
Vector3D vec2 = ...
auto vec 3 = vec1 + vec2 * BOUNDARY_VISUALIZE_STEP_SIZE
```

Treeston:
@Shauren Numerical precision mid-operation. If your coordinates are magnitude 10⁴, and you then calculate a scalar product, then you start to lose precision on the lower digits (which you're then missing when you subtract the values).

src/server/game/Maps/AreaBoundary.h
```diff
@@ -0,0 +1,147 @@
...
+};
+
+class TriangleBoundary : public AreaBoundary
+{
+    public:
+        TriangleBoundary(Position const& pointA, Position const& pointB, Position const& pointC, bool isInverted = false);
+        TriangleBoundary(DoublePosition const& pointA, DoublePosition const& pointB, DoublePosition const& pointC, bool isInverted = false);
+
+    protected:
+        bool IsWithinBoundaryArea(const Position* pos) const override;
```
Naios: Nice usage of virtual inheritance here

src/server/game/Maps/AreaBoundary.h
```diff
@@ -0,0 +1,147 @@
...
+        ZRangeBoundary(float minZ, float maxZ, bool isInverted = false);
+
+    protected:
+        bool IsWithinBoundaryArea(const Position* pos) const override;
+
+    private:
+        const float _minZ, _maxZ;
+};
+
+#endif //TRINITY_AREA_BOUNDARY_H
```
Naios: newline

src/server/game/AI/CreatureAI.cpp
Naios: A direct message to the chat handler of the user who invoked the debug function

src/server/game/AI/CreatureAI.cpp
Naios: Or you could use `std::initializer_list`  instead of `std::list` [which the compiler will optimize out:
](https://goo.gl/DqxFGt)
```c++
for (int i : std::initializer_list<int>{1, 2, 3, 4})
{
    // ...
}
```

src/server/game/AI/CreatureAI.cpp
Treeston: So I'd need to pass the chat handler through to the function, but doable, I guess.

src/server/game/AI/CreatureAI.cpp
Treeston: Would need to turn the Vector3D back to a Position though for boundary check. Probably not worth it overhead wise?

src/server/game/AI/CreatureAI.cpp
```diff
@@ -271,6 +271,112 @@ bool CreatureAI::_EnterEvadeMode()
...
+        return true;
+
+    switch (GetBoundaryEvadeType())
+    {
+        case EVADE_TYPE_RETURN:
+            EnterEvadeMode();
+            break;
+        case EVADE_TYPE_RESPAWN:
+            me->DespawnOrUnsummon();
+            me->SetRespawnTime(15);
```
Shauren: Not sure I agree with this. On retail the despawn happens whenever creature enters evade mode, not only if running out of room. (should probably be inside EnterEvadeMode for these bosses, maybe as BossAI/ScriptedAI member _despawnAtEvade?)

src/server/game/AI/CreatureAI.cpp
```diff
@@ -271,6 +271,112 @@ bool CreatureAI::_EnterEvadeMode()
...
+}
+
+bool CreatureAI::CheckBoundary(Position* who) const
+{
+    if (!who)
+        who = me;
+
+    if (_boundary)
+        for (CreatureBoundary::const_iterator it = _boundary->begin(); it != _boundary->end(); ++it)
+            if ((*it) && !(*it)->IsWithinBoundary(who))
```
jackpoz: can *it be NULL at this point ?

src/server/game/AI/CreatureAI.cpp
Treeston: According to @Aokromes, Ymiron (in Utgarde Pinnacle) runs back if it resets normally, but despawns if it leaves the room. Not sure on this one.

NorthStrider: it's depending on the boss. Needs special scripting on the ai codes.

Treeston: Pass an argument to EnterEvadeMode maybe indicating what the cause of evading is?

NorthStrider: sth like   virtual void CreatureAI::EnterEvadeMode(bool roomLeft = false) ?

Treeston:
```c++
enum EvadeReason {
    EVADE_CAUSE_NO_THREAT,
    EVADE_CAUSE_OUT_OF_BOUNDS,
    EVADE_CAUSE_OTHER
}
```

`virtual void CreatureAI::EnterEvadeMode(EvadeReason why);`

Keader: On SOO (Siege of Orgrimmar), all boss give "despawn" regardless of which way the evade occurred. On Retail (WoD), all boss on Icecrown Citadel give despawn for all evade conditions too.

ccrs:
In retail there was another evade mechanic:
While in instance, if creature cant reach victim (no available path) it gets evade flag until player is reachable again (without actually returning home, just standing still).

Its kinda unrelated to this PR, but I thought its worth saying.

Demonid:
@ccrs Useful for flyhack users while Mmaps are enabled.

Runico:
Yes, deathbringer saurfang easily can get killed by one hunter, it took 30 mins depended on hunter gear

ccrs:
LOL thats something Id like to see hahaha

Treeston:
Alright, so here is a list of boundary information that is missing before this can be merged.

* Ulduar - Auriaya: Does she reset if you drag her out of the center ring of the antechamber?
* Onyxia's Lair: Does she reset if you drag her out of her lair?
* Obsidian Sanctum - Sartharion: Does he reset if dragged? Do the drakes if they engage with him? (Anecdotally: No on both accounts, so only a sanity check boundary around the entire zone is needed)
* Obsidian Sanctum - Minibosses: Do the drakes reset while fought on their own if they are dragged away from their spawn?
* Naxxramas - Patchwerk: Does he ever reset?
(Anecdotally: No)
* Naxxramas - Grobbulus: Does he ever reset?
* Ruby Sanctum - Halion: Does he reset if pulled away from the arena?

Once I have information on all of the above, Wrath-level raids will be done, and I'll consider this PR ready for merging into the core.

ccrs:
I can confirm these ones:
* Onyxia: evades if out of combat area
* Sartharion: evades if out of combat area
* OS pre bosses: they dont evade
* Halion: evades if it chases someone who is left out the fight area

What im not sure about is the behaviour of the boss when it finishes the evade movent: engage combat right away if someone is in aggro range (which is likely) or has something simillar to a small delay.
Right now in retail it disappears for a small period of time to let the players run to a safe area; cant confirm what was the 3.3.5 behaviour.

src/server/game/Instances/InstanceScript.h
```diff
@@ -221,7 +209,7 @@ class InstanceScript : public ZoneScript
         virtual bool SetBossState(uint32 id, EncounterState state);
         EncounterState GetBossState(uint32 id) const { return id < bosses.size() ? bosses[id].state : TO_BE_DECIDED; }
         static std::string GetBossStateName(uint8 state);
-        BossBoundaryMap const* GetBossBoundary(uint32 id) const { return id < bosses.size() ? &bosses[id].boundary : NULL; }
+        CreatureBoundary const* GetBossBoundary(uint32 id) const { return id < bosses.size() ? &bosses[id].boundary : NULL; }
```
ccrs: If no bundary is set what about adding a "standard" square size boundary whose size depends on the boss size?

src/server/game/Maps/AreaBoundary.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "Unit.h"
+#include "TemporarySummon.h"
+
+// ---== RECTANGLE ==---
+RectangleBoundary::RectangleBoundary(float southX, float northX, float eastY, float westY, bool isInverted) :
+    AreaBoundary(BoundaryType::BOUNDARY_RECTANGLE, isInverted), _minX(southX), _maxX(northX), _minY(eastY), _maxY(westY) { }
+bool RectangleBoundary::IsWithinBoundaryArea(const Position* pos) const
+{
+    if (!pos)
+        return true;
```
ccrs: why do you return true here?

ccrs:
Im now wondering, would it be necessary for a particular boss to have multiple boundary types?
(depending on the different boss phases for example)

Treeston:
@ccrs the boss ai can either ignore boundary-related evades outright during that phase if it's an RP phase, or can call SetBoundary as necessary to adjust the boss boundary during transition.

src/server/game/Instances/InstanceScript.h
Treeston: There is far too much variance between different boss rooms for this to work.

PR #16120, opened 2015-12-29
AHBot item class modifiers

mysterious-static:
Adds per-class modifiers to worldserver.conf.

### DIFF
--- a/src/server/worldserver/worldserver.conf.dist
+++ b/src/server/worldserver/worldserver.conf.dist
@@ -3019,6 +3019,29 @@ AuctionHouseBot.Items.Orange.Price.Ratio = 100
 AuctionHouseBot.Items.Yellow.Price.Ratio = 100
 
 #
+#    AuctionHouseBot.Class.CLASS.Price.Ratio
+#       Description: Percentage by which the price of items sold of each class is incremented / decreased (for all houses)
+#       Default:   100 - (No change)
+
+AuctionHouseBot.Class.Consumable.Price.Ratio = 100 
+AuctionHouseBot.Class.Container.Price.Ratio = 100 
+AuctionHouseBot.Class.Weapon.Price.Ratio = 100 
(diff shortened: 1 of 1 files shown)


DDuarte:
`config.SetPriceRatioPerClass(ITEM_CLASS_CONSUMABLE, sAuctionBotConfig->GetConfig(CONFIG_AHBOT_CLASS_CONSUMABLE_PRICE_RATIO) / 100);`

This is dividing an uint32 by an int, which is an int. If the value in the config is 150, the result will be 1, having no effect on the outcome.
In the previous code, we had the class ratio being multiplied by the quality ration and only then an integer division by 100, which produces different results if you do the division first.

My suggestion is to remove the code changes from the PR and only add the missing config values.

P.S When you say "divided by zero" you mean by 100, right?
P.P.S I expect ratios to be floats between around 1 and percentages around 100, it's a bit weird (for me) to mix both concepts. (This isn't an issue with this PR, only a deeper issue in the existing code.)

velinath:
@DDuarte Thanks for the feedback, helps me understand what is going on. If you think it would be worthwhile I could update those config values to floats in AuctionHouseBot.h and update the worldserver.conf to reflect that, that doesn't look difficult.

Re: PS: Yes. I am terrible. Edited.

NathanHandley:
I don't think float will help very much, and suggest going with DDuarte's suggestion of keeping things the same except the missing values in the config.

However, looking at SetPricesOfItem, there's a lot of work that could be done there.  Nearly every line there has implicit rounding which will have a compounded deviation.  It's exacerbated at the very low end of the ranges.  For example, if classRatio is 3 and qualityRatio is 3, you'll have a priceRatio of zero after that line resolves.  So, if you were going to take it a step further, remove the implicit rounding from that method.

velinath:
All right, I'll start looking into that and for now just make this PR for the worldserver.conf change.

velinath:
Rebased.

PR #16101, opened 2015-12-26
Updated AuctionHouseSeller to have item class stack control

NathanHandley:
For each item class (consumable, glyph), you can now edit the worldserver.conf to configure the ratio (percent) of random compared to single stack AuctionHouseSeller auction postings.  For example: you can configure it so that all Glyphs are non-stacked, gems are sometimes stacked, reagents are always randomly stacked, etc.  I did this to give a more blizzlike behavior.

**Example 1**
AuctionHouseBot.Class.RandomStackRatio.Consumable = 0
- New consumable auctions posted by AuctionHouseSeller will always be single stack (1).

**Example 2**
AuctionHouseBot.Class.RandomStackRatio.Consumable = 100
- New consumable auctions posted by AuctionHouseSeller will always be of a random stack size (1...n).

**Example 3**
AuctionHouseBot.Class.RandomStackRatio.Consumable = 20
- New consumable auctions posted by AuctionHouseSeller have a 20% chance of being a random stack size (1...n), and an 80% chance of being single stack size (1).

_Note:_ This is my first pull request.  I emulated the coding conventions in the files I changed, including terms (ratio).  Please point out anywhere that I egregiously violated the coding guidelines.

PR #16123, opened 2015-12-29
Improve AHBot seller math

mysterious-static:
As discussed in #16120, the AHBot has rounding issues when selling items. This PR should mitigate those issues by using floating point math in SetPricesOfItem.

Many thanks go out to @NathanHandley for code review before pushing this PR. Being a server guy who dabbles in PHP, I don't know a lot about C++. He really went above and beyond to help out.

One of the outstanding questions that I have is convention related. I'm using `basePriceFloat` as a variable name right now. Would `tempBasePrice` be more appropriate or is there a project convention for this?

### DIFF
--- a/src/server/game/AuctionHouseBot/AuctionHouseBotSeller.cpp
+++ b/src/server/game/AuctionHouseBot/AuctionHouseBotSeller.cpp
@@ -565,13 +565,13 @@ void AuctionBotSeller::LoadSellerValues(SellerConfiguration& config)
             break;
     }
 
-    config.SetPriceRatioPerQuality(AUCTION_QUALITY_GRAY, PriceRatio * sAuctionBotConfig->GetConfig(CONFIG_AHBOT_ITEM_GRAY_PRICE_RATIO) / 100);
-    config.SetPriceRatioPerQuality(AUCTION_QUALITY_WHITE, PriceRatio * sAuctionBotConfig->GetConfig(CONFIG_AHBOT_ITEM_WHITE_PRICE_RATIO) / 100);
-    config.SetPriceRatioPerQuality(AUCTION_QUALITY_GREEN, PriceRatio * sAuctionBotConfig->GetConfig(CONFIG_AHBOT_ITEM_GREEN_PRICE_RATIO) / 100);
-    config.SetPriceRatioPerQuality(AUCTION_QUALITY_BLUE, PriceRatio * sAuctionBotConfig->GetConfig(CONFIG_AHBOT_ITEM_BLUE_PRICE_RATIO) / 100);
-    config.SetPriceRatioPerQuality(AUCTION_QUALITY_PURPLE, PriceRatio * sAuctionBotConfig->GetConfig(CONFIG_AHBOT_ITEM_PURPLE_PRICE_RATIO) / 100);
-    config.SetPriceRatioPerQuality(AUCTION_QUALITY_ORANGE, PriceRatio * sAuctionBotConfig->GetConfig(CONFIG_AHBOT_ITEM_ORANGE_PRICE_RATIO) / 100);
-    config.SetPriceRatioPerQuality(AUCTION_QUALITY_YELLOW, PriceRatio * sAuctionBotConfig->GetConfig(CONFIG_AHBOT_ITEM_YELLOW_PRICE_RATIO) / 100);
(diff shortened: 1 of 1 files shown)


src/server/game/AuctionHouseBot/AuctionHouseBotSeller.cpp
```diff
@@ -708,15 +708,15 @@ void AuctionBotSeller::SetPricesOfItem(ItemTemplate const* itemProto, SellerConf
 
     if (sAuctionBotConfig->GetConfig(CONFIG_AHBOT_BUYPRICE_SELLER))
         buyPrice = sellPrice;
+	
+	float basePriceFloat = (buyPrice * stackCount * priceRatio) / (itemProto->GetClass() == 6 ? 200.0f : static_cast<float>(itemProto->GetBuyCount())) / 100.0f;
+    float range = basePriceFloat * 0.04f;
+	
+    buyp = urand(static_cast<uint32>(basePriceFloat - range + 0.5f), static_cast<uint32>(basePriceFloat + range + 0.5f)) + 1;
```
DDuarte: perhaps frand can be used here instead?

NathanHandley: Replacing with frand and then converting to int is a good alternate.  The net effect will be similar.

src/server/game/AuctionHouseBot/AuctionHouseBotSeller.cpp
```diff
@@ -708,15 +708,15 @@ void AuctionBotSeller::SetPricesOfItem(ItemTemplate const* itemProto, SellerConf
 
     if (sAuctionBotConfig->GetConfig(CONFIG_AHBOT_BUYPRICE_SELLER))
         buyPrice = sellPrice;
+	
+	float basePriceFloat = (buyPrice * stackCount * priceRatio) / (itemProto->GetClass() == 6 ? 200.0f : static_cast<float>(itemProto->GetBuyCount())) / 100.0f;
```
NathanHandley: Replace the tab with spaces, please.

src/server/game/AuctionHouseBot/AuctionHouseBotSeller.cpp
velinath: Like this? Finding documentation for frand was hell so I'm kind of guessing. :\

```cpp
buyp = static_cast<uint32>(frand(basePriceFloat - range, basePriceFloat + range) + 1.5f);
```

src/server/game/AuctionHouseBot/AuctionHouseBotSeller.cpp
velinath: All set on this one. No idea how that happened.

src/server/game/AuctionHouseBot/AuctionHouseBotSeller.cpp
NathanHandley: Yes, exactly like that.

Looking at the +1 that was originally there (and now 1.5f) looks like it was there to avoid assigning the buy price as 0, which would break things.  I suggest just making it 0.5f, and having a conditional that follows as "if (buyp == 0) then 1 else buyp".  It would make this part of code very clear and explicit on intent.  Any have any alternate opinions?

velinath: I'll make that change here.

Looking back, L717-719 may benefit from:
```cpp
basePriceFloat = buyp * 0.5f;
range = buyp * 0.4f;
bidp = static_cast<uint32>(frand(basePriceFloat - range, basePriceFloat + range) + 1.5f);
```

instead? Or is it not worth it since we've already lost precision due to `buyp` being uint32?

DDuarte:
Word of advice: make sure that these changes are actually fixing something, tested in-game. Changing ints to floats for no obvious reasons is not worth it and may introduce bugs.

NathanHandley:
I think I can speak to the 'why' behind the change.  When I was testing the work for #16101, I noticed that the current logic was 'rounding away' and being fairly inaccurate on seller price-- specifically at the low end.  This is why I said something on vellnath's other pull request.

For example, if you have the following values set for an item:
Class Ratio = 15
Price Ratio = 25
Quality Ratio = 20

[As all int]
(Class Ratio * Price Ratio) / 100 = (15 * 25) / 100 = 3.75 => 3
(3 * Quality Ratio) / 100 = (3 * 20) / 100 = 0.6 = 0

Which, through the function, was setting the value of items to 1 in the end.

PR #16156, opened 2016-01-03
Game/Maps: Instance handling bug-fix follow-up

### LINKED ISSUE #16150: [3.3.5] Entering instance resets you to homebind in specific circumstances
3.3.5, b6ed3d3fc2b8447059ff0aa0e5b6de36947fd006

Bug with new instance handling. Repro:
1. Zone into instance while not partied
2. Get invited by someone else while in the instance
3. Your party leader zones into the same instance as you
4. You leave and re-enter. This should send you to the party's instance. Instead, it sends you to your homebind.

Treeston:
Game/Maps: Instance handling follow-up:
- Fix a bug where a player could maintain a conflicting non-perm solo bind if they were in the instance when invited to group. Closes #16150.
- If a group is created while the leader is in an instance that nobody is bound to, the group will take over the instance and bind to it. This stops the homebind timer when reforming group after disconnects and the like.

### DIFF
--- a/src/server/game/Groups/Group.h
+++ b/src/server/game/Groups/Group.h
@@ -194,6 +194,7 @@ class Group
         bool   AddMember(Player* player);
         bool   RemoveMember(ObjectGuid guid, const RemoveMethod &method = GROUP_REMOVEMETHOD_DEFAULT, ObjectGuid kicker = ObjectGuid::Empty, const char* reason = NULL);
         void   ChangeLeader(ObjectGuid guid);
+ static void   ConvertLeaderInstancesToGroup(Player* player, Group* group, bool switchLeader);
         void   SetLootMethod(LootMethod method);
         void   SetLooterGuid(ObjectGuid guid);
         void   SetMasterLooterGuid(ObjectGuid guid);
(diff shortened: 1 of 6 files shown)


src/server/game/Maps/MapInstanced.cpp
```diff
@@ -163,7 +163,11 @@ Map* MapInstanced::CreateInstanceForPlayer(const uint32 mapId, Player* player, u
             {
                 groupBind = group->GetBoundInstance(this);
                 if (groupBind)
+                {
+                    // solo saves should be reset when entering a group's instance
+                    player->UnbindInstance(GetId(), player->GetDifficulty(IsRaid()));
```
Shauren: Where is MapInstanced::CreateInstanceForPlayer called from? UnbindInstance is not threadsafe

src/server/game/Groups/Group.h
```diff
@@ -194,6 +194,7 @@ class Group
         bool   AddMember(Player* player);
         bool   RemoveMember(ObjectGuid guid, const RemoveMethod &method = GROUP_REMOVEMETHOD_DEFAULT, ObjectGuid kicker = ObjectGuid::Empty, const char* reason = NULL);
         void   ChangeLeader(ObjectGuid guid);
+ static void   ConvertLeaderInstancesToGroup(Player* player, Group* group, bool switchLeader);
```
Shauren: u wot

ccrs: tld u m8

PR #16171, opened 2016-01-03
Core/Scripts/DB: move Gryphoneer Windbellow to SAI

mysterious-static:
This should correctly migrate Gryphoneer Windbellow, NPC 20235, to SAI / database. Feedback on conditions SQL would be appreciated; I want to make sure that I am using ElseGroup correctly. Will remove excess comments after code review, but wanted to make sure everything was clear for this. When testing, the behavior was identical to the current npc_taxi script.

If someone has a live account with a conveniently correct-level character (I do not), it would be nice to see whether the Shatter Point -> Honor Point flight point is indeed only valid while a player has the quest, or whether it continues to be available after the quest "Go to the Front" (ID: 10382) has been turned in. The behavior in the existing C++ script and in this PR is that it becomes unavailable after the quest is turned in.

Will squash when it's deemed ready to merge!

### DIFF
--- a/sql/updates/world/2016_01_03_05_world.sql
+++ b/sql/updates/world/2016_01_03_05_world.sql
@@ -1,0 +1,25 @@
+-- Add sniffed gossip texts to menu
+DELETE FROM `gossip_menu_option` WHERE `menu_id` IN (8095, 8096, 8100);
+INSERT INTO `gossip_menu_option` (`menu_id`, `id`, `option_icon`, `option_text`, `OptionBroadcastTextID`, `option_id`, `npc_option_npcflag`, `action_menu_id`, `action_poi_id`, `box_coded`, `box_money`, `box_text`, `BoxBroadcastTextID`) VALUES
+(8095, 0, 2, 'Show me where I can fly.',12271,1,1,0,0,0,0,'',0),
+(8096, 0, 0, 'Send me to Honor Point!',18200,1,1,0,0,0,0,'',0),
+(8096, 1, 0, 'Send me to the Abyssal Shelf!',17936,1,1,0,0,0,0,'',0),
+(8100, 0, 0, 'Send me to Shatter Point!',17937,1,1,0,0,0,0,'',0);
+
+-- Migrate NPC to use SmartAI
+UPDATE `creature_template` SET `AIName` = 'SmartAI', `ScriptName` = '', `gossip_menu_id` = 8096 WHERE `entry` = 20235;
+
+-- Condition: Gossip menu ID 0 needs either quest ID 10163 or quest ID 10346 to be incomplete.
+-- Condition: Gossip menu ID 1 needs quest ID 10382 to be complete
+DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId`= 15 AND `SourceGroup` = 8096 AND `SourceEntry`IN (0,1);
+INSERT INTO `conditions` (`SourceTypeOrReferenceId`,`SourceGroup`,`SourceEntry`,`SourceId`,`ElseGroup`,`ConditionTypeOrReference`,`ConditionTarget`,`ConditionValue1`,`ConditionValue2`,`ConditionValue3`,`NegativeCondition`,`ErrorTextId`,`ScriptName`,`Comment`) VALUES
+(15,8096,0,0,0,9,0,10163,0,0,0,0,'','Gryphoneer Windbellow - Show gossip option 0 if player has taken quest ID 10163'),
+(15,8096,0,0,1,9,0,10346,0,0,0,0,'','Gryphoneer Windbellow - Show gossip option 0 if player has taken quest ID 10346'),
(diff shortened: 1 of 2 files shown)


sql/updates/world/2016_01_XX_XX_world.sql
```diff
@@ -0,0 +1,26 @@
...
+DELETE FROM `gossip_menu_option` where menu_id = 8095;
+INSERT INTO `gossip_menu_option` (`menu_id`, `id`, `option_icon`, `option_text`, `OptionBroadcastTextID`, `option_id`, `npc_option_npcflag`, `action_menu_id`, `action_poi_id`, `box_coded`, `box_money`, `box_text`, `BoxBroadcastTextID`) VALUES
+(8095,0,0, 'Send me to the Abyssal Shelf!', 16615,1,1,0,0,0,0,NULL,0), -- There are multiple entries for this text. Review would be appreciated by anyone who can see which text is correct.
+(8095,1,0,'Send me to Honor Point!',17936,1,1,0,0,0,0,NULL,0);
+
+-- Migrate NPC to use SmartAI
+UPDATE `creature_template` SET `AIName` = 'SmartAI', `ScriptName` = '' WHERE `entry` = 20235;
+
+-- Condition: Gossip menu ID 0 needs either quest ID 10163 or quest ID 10346 to be incomplete.
+-- Condition: Gossip menu ID 1 needs quest ID 10382 to be taken and not rewarded. I wasn't sure how to read the CPP condition.
```
joschiwald: change condition to is quest completed

sql/updates/world/2016_01_XX_XX_world.sql
```diff
@@ -0,0 +1,26 @@
...
+-- I think I am using ElseGroup correctly but please advise if I am not.
```
joschiwald: change ElseGroup 78 -> 0, 79 -> 1, 80 -> 0

sql/updates/world/2016_01_XX_XX_world.sql
```diff
@@ -0,0 +1,26 @@
+-- THERE ARE A LOT OF EXCESS COMMENTS IN THIS FILE. They will be removed after code review.
+-- Add sniffed gossip texts to menu
```
joschiwald: i found this in my sniffs
```sql
INSERT INTO `gossip_menu_option` (`menu_id`, `id`, `option_icon`, `option_text`, `box_coded`, `box_money`, `box_text`) VALUES
(8095, 0, 2, 'Show me where I can fly.', 0, 0, ''), -- 20234
(8096, 0, 0, 'Send me to Honor Point!', 0, 0, ''), -- 20235
(8096, 1, 0, 'Send me to the Abyssal Shelf!', 0, 0, ''), -- 20235
```

joschiwald:
take a look at my comments, they are not outdate

velinath:
@joschiwald thanks for feedback and especially for sniffs - very helpful to make sure I have things correct! I will make changes shortly.

sql/updates/world/2016_01_XX_XX_world.sql
velinath: as a general reference, then, I can just use ElseGroup starting enumeration at 0 for each, that's not a global thing? That will make things much easier.

sql/updates/world/2016_01_XX_XX_world.sql
velinath: replacing both negative conditions? Looks like the line notes did something weird so it's not pointing me to the right place, sorry

joschiwald: yes

sql/updates/world/2016_01_XX_XX_world.sql
joschiwald: right

sql/updates/world/2016_01_XX_XX_world.sql
velinath: sorry, unclear comment: there are 2 "Send me to the Abyssal Shelf!" strings in `broadcast_text` and I'm not sure how to tell which is the correct one to use in this case. The other one, I presume, is used for the Horde NPC Wing Commander Brack (19401).

joschiwald: that's possible but nothing found in my sniff, assign the higher broadcasttext id to the gossip of 20235

velinath:
Thanks @joschiwald! I went ahead and implemented those suggestions, cleaned up a couple things I missed, and squashed.

joschiwald:
according to my sniffs, gossips are still wrong
i suggest to add all 4 of them because they are not in db, adjust your script to use gossip 8096 (8085 is used by another npc), and let the others unimplemented

velinath:
got it - sorry, misunderstood. Rebased again.

tkrokli:
Well done. Nice work. :+1:

Tauriella:
Excellent :+1:

PR #16182, opened 2016-01-04
Core/Scripts: remove unused scripts

mysterious-static:
This PR removes scripts for taxi NPC's that are already on SmartAI in the database and no longer have references to `npc_taxi` script. It also cleans up a comment reference to a script that was previously moved to SmartAI in Moonglade.


tkrokli:
Nice cleanup. :+1: Thank you for keeping up the pace. :smile:

Issue #7315, opened 2012-08-07
[Talent Bug] Tidal Waves


Nividan:
Greethings,

I found bug in "Tidal Waves" Talent.

What is supposed to do:
After you use the skill Chain Heal or Riptide you get 2 stuck of buff named Tidal Waves.
When you use  Lesser Healing Wave or Healing Wave its get 1 stuck down.
When you use  Lesser Healing Wave or Healing Wave again its remove the buff completely.

What happened:
After you use the skill Chain Heal or Riptide you get 2 stuck of buff named Tidal Waves.
When you use Healing Wave its get 1 stuck down.
When you use Lesser Healing Wave or Healing Wave again its remove the buff completely.
But when you use Lesser Healing Wave its get 1 stuck down.
When you use Lesser Healing Wave again you still see the buff and it still give you the bonus.

Tidal Waves - http://old.wowhead.com/spell=51564/tidal-waves
Lesser Healing Wave - http://www.wowwiki.com/Lesser_Healing_Wave
Healing Wave - http://old.wowhead.com/spell=331/healing-wave

Hash: f523192   Tdb: 335.11.47+

Keader:
Still bugged on rev 5c97594a1691b78ab881437985dc5eeebb31a979
@Aokromes can you reopen this?

ariel-:
Apparently the problem is that charges are being dropped twice by the core, one for each call to Unit::IsSpellCrit(), this function is called by Spell::DoAllEffectOnLaunchTarget and Spell::DoAllEffectOnTarget.

Does this mean spell crit is calculated twice? it uses roll_chance_f so maybe first time its crit, then its not, or the other way round.

Issue #15529, opened 2015-09-17
[3.3.5] Warrior - Spell Recklessness aura stuck


Keader:
http://wotlk.openwow.com/spell=1719

Spell Description:
Your next 3 special ability attacks have an additional 100% to critically hit but all damage taken is increased by 20%. Lasts 12 sec.

Problem:
If you use Recklessness with spells of one target it will work normally.
However use Recklessness with spells of multiple targets, the last charge of the buff does work (is not consumed and not give the effects)

Reproduction:
Use Recklessness buff and use the spell Cleave (id = 47520) in a location with multiple targets. The first 2 charge of Recklessness will be consumed normally, however the latter will not function.

Little Vid: https://youtu.be/Z2V-vAKb5UE

Using comments in this piece of code: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L7695 is called 2 times, when it is used spells of multiple target like a cleave

Trinity Hash: 9a43dd35cdd40228db5b8519b53ca5b6438c3c18

Maybe related to: #7315

Keader:
@robinsch I do not know if you are in time, but it is one of the few who is interested / have expertise in spells correction could take a look then ?!

robinsch:
The entire proc system never really worked good with charges - but I currently don't have the time to fix it.

For ´hack-fixing the charges drop you could just add a check for 
if (GetCharges() > 1) *handled = true;
but this will not fix the crit chance.

Keader:
Related too: https://github.com/TrinityCore/TrinityCore/issues/932#issuecomment-4033950

Keader:
@ariel-  this maybe related #16186 ?
Can you take a look?

ariel-:
@Keader Recklessness drops one charge per target, I don't know if it should be like this or should drop a charge per SPELL

Keader:
But if by target should not consume  three charges in the first Cleave?
When there are 3 Targets when using cleave are still consumed only 2 charge

ariel-:
Exactly,  spellMod charges (as in the part that gives next 3 spells 100% crit) will now be consumed once per target, however the aura gets stuck at 2 charges now because when you hit 3 targets, the modspells are consumed to 0, and then the aura stacks are never updated, giving the aura "stuck" at 2 stacks.

[...]

PR #16186, opened 2016-01-04
Core/Spells: Fix redundant mod charge drop and spell crit calc

ariel-:
Note that critical is already calculated in Spell::DoAllEffectOnLaunchTarget()

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2359,7 +2359,7 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
     // Do healing and triggers
     if (m_healing > 0)
     {
-        bool crit = caster->IsSpellCrit(unitTarget, m_spellInfo, m_spellSchoolMask);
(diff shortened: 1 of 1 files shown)


Keader:
This fix #7315 for me

ariel-:
Note that this will fix only duplicated modSpell consuming, there is still issue of Aura stacks not synchronizing properly with the charges.

Detailed explanation in https://github.com/TrinityCore/TrinityCore/issues/15529#issuecomment-168849319

jackpoz:
could a solution to the other issue require to revert this PR ? maybe it's better to find a solution for all the cases ?

ariel-:
Not necessarily, IsSpellCrit is more than an observer, it calculates crit chance for a spell using active crit spellmods, and in doing so removes the charge from the spellmod.

This PR only replaces the (extra, IMO) call to IsSpellCrit, using member TargetInfo::crit, this member is already initialized with proper value (crit or not), as stated, in Spell::DoAllEffectOnLaunchTarget()

PR #16204, opened 2016-01-07
Scripts/Shadowmoon Valley: Dragonmaw Peon AI Update

ghost:
Fixes issue #16210.

### DIFF
--- a/src/server/scripts/Outland/zone_shadowmoon_valley.cpp
+++ b/src/server/scripts/Outland/zone_shadowmoon_valley.cpp
@@ -514,8 +514,6 @@ public:
             Initialize();
         }
 
-        void EnterCombat(Unit* /*who*/) override { }
-
         void SpellHit(Unit* caster, const SpellInfo* spell) override
         {
             if (!caster)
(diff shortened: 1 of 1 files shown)


Saben65:
Seems to be some additional issues with the scripting. The NPCs will stay aggro'd to you indefinitely without evading due to distance. If you do drop aggro (I could only do so by enabling GM mode) they will stay wherever they dropped aggro and do their mining emote instead of running back to their spawn point.

Not sure how to fix this.

MitchesD:
you should check for victim

Saben65:
I assume you are talking about this?

```C++
            if (!UpdateVictim())
                return;
```

That fixed the issue where they wouldn't reset.

src/server/scripts/Outland/zone_shadowmoon_valley.cpp
```diff
@@ -514,6 +514,8 @@ class npc_dragonmaw_peon : public CreatureScript
             Initialize();
         }
 
+        void EnterCombat(Unit* /*who*/) override { }
```
jackpoz: is this changed needed for a particular reason ?

Saben65: Nothing in particular. I just added it based on it being there in similar scripts. I will remove it.

jackpoz:
I aggroed 2, moved a bit away, died with ".die" and 1 of the peons moved to his position while the other did the mining animation for a few seconds and then moved to his position. Not sure if there's a way to make them do the mining position only after going back to their position

PR #16239, opened 2016-01-12
[3.3.5] Core/Scripts: move npc_shadowfang_prisoner gossip option to DB

ghost:
Getting rid of one more incorrect and hardcoded text
which now will be read correct and from the DB instead.

Wrong text: "Thanks, I'll follow you to the door." (core)
Correct text: "Please unlock the courtyard door." (DB)

Thanks to @Aokromes, @r00ty, @velinath, @Nayd, @Treeston, @Zedron
and lots of other regular TrinityCore developers and members for the
inspiration and motivation, as well as helpfulness along the way.

### DIFF
--- a/sql/updates/world/2016_01_12_12_world335.sql
+++ b/sql/updates/world/2016_01_12_12_world335.sql
@@ -1,0 +1,5 @@
+-- Insert missing gossip_menu_option 21213 for Sorcerer Ashcrombe and 21214 for Deathstalker Adamant:
+DELETE FROM `gossip_menu_option` WHERE `menu_id` IN (21213,21214) AND `OptionBroadcastTextID`= 2802;
+INSERT INTO `gossip_menu_option` (`menu_id`,`id`,`option_icon`,`option_text`,`OptionBroadcastTextID`,`option_id`,`npc_option_npcflag`,`action_menu_id`,`action_poi_id`,`box_coded`,`box_money`,`box_text`,`BoxBroadcastTextID`) VALUES
+(21213,0,0,'Please unlock the courtyard door.',2802,1,1,0,0,0,0,'',0),
+(21214,0,0,'Please unlock the courtyard door.',2802,1,1,0,0,0,0,'',0);
--- a/src/server/scripts/EasternKingdoms/ShadowfangKeep/shadowfang_keep.cpp
+++ b/src/server/scripts/EasternKingdoms/ShadowfangKeep/shadowfang_keep.cpp
@@ -63,8 +63,6 @@ enum Creatures
     NPC_ASH                 = 3850
 };
 
-#define GOSSIP_ITEM_DOOR        "Thanks, I'll follow you to the door."
-
 class npc_shadowfang_prisoner : public CreatureScript
 {
 public:
@@ -93,7 +91,7 @@ public:
         InstanceScript* instance = creature->GetInstanceScript();
 
         if (instance && instance->GetData(TYPE_FREE_NPC) != DONE && instance->GetData(TYPE_RETHILGORE) == DONE)
-            player->ADD_GOSSIP_ITEM(GOSSIP_ICON_CHAT, GOSSIP_ITEM_DOOR, GOSSIP_SENDER_MAIN, GOSSIP_ACTION_INFO_DEF + 1);
+            player->ADD_GOSSIP_ITEM_DB(Player::GetDefaultGossipMenuForSource(creature), 0, GOSSIP_SENDER_MAIN, GOSSIP_ACTION_INFO_DEF + 1);
 
         player->SEND_GOSSIP_MENU(player->GetGossipTextId(creature), creature->GetGUID());
 


tkrokli:
Please note: I will need some assistance in troubleshooting why the gossip menu option will not trigger the core script like it used to do. It may be that I have missed just a minor setting, a small, but important trigger of some sort to link the gossip menu option to the core script.

src/server/scripts/EasternKingdoms/ShadowfangKeep/shadowfang_keep.cpp
```diff
@@ -93,7 +91,7 @@ class npc_shadowfang_prisoner : public CreatureScript
         InstanceScript* instance = creature->GetInstanceScript();
 
         if (instance && instance->GetData(TYPE_FREE_NPC) != DONE && instance->GetData(TYPE_RETHILGORE) == DONE)
-            player->ADD_GOSSIP_ITEM(GOSSIP_ICON_CHAT, GOSSIP_ITEM_DOOR, GOSSIP_SENDER_MAIN, GOSSIP_ACTION_INFO_DEF + 1);
+            player->ADD_GOSSIP_ITEM_DB(Player::GetDefaultGossipMenuForSource(creature), 0, GOSSIP_SENDER_MAIN, GOSSIP_OPTION_GOSSIP);
```
ForesterDev: player->ADD_GOSSIP_ITEM_DB(Player::GetDefaultGossipMenuForSource(creature), 0, GOSSIP_SENDER_MAIN, GOSSIP_ACTION_INFO_DEF + 1);

tkrokli: OK, that is absolutely worth while testing. Since it does not work as it is, I will test.
Have you got previous experience on using that argument with this function? :)

ForesterDev: Yes, this is how it should be. Look at the definition:

#define ADD_GOSSIP_ITEM_DB(h, i, c, d)   PlayerTalkClass->GetGossipMenu().AddMenuItem(h, i, c, d)

// h - Menu ID from DB
// i - Menu item ID from DB
// c - Sender(this is to identify the current Menu with this item)
// d - Action (identifys this Menu Item)

tkrokli: Very nice, my bad for not reading the details on that function. :-p Just compiled a new test now.
Going to update this PR if it works as expected. :smirk_cat:

Treeston:
💙

tkrokli:
I'm so glad I found this "new" tool to get rid of those "1-liner" hardcoded texts without having to go to the job of re-scripting the events in SAI, just to get rid of a few lines of hardcoded text. I'm grinning almost all the time now, it will help velinath and me a lot with our work on that long list. :sweat_smile:
(Like I mentioned in #trinity, I know that some of these events might do better when moved to SAI, but for now, the important part of me is to get on with the work at hand.) :+1:

sql/updates/world/2016_01_12_12_world335.sql
```diff
@@ -0,0 +1,5 @@
+-- Insert missing gossip_menu_option 21213 for Sorcerer Ashcrombe and 21214 for Deathstalker Adamant:
+DELETE FROM `gossip_menu_option` WHERE `menu_id` IN (21213,21214) AND `OptionBroadcastTextID`= 2802;
+INSERT INTO `gossip_menu_option` (`menu_id`,`id`,`option_icon`,`option_text`,`OptionBroadcastTextID`,`option_id`,`npc_option_npcflag`,`action_menu_id`,`action_poi_id`,`box_coded`,`box_money`,`box_text`,`BoxBroadcastTextID`) VALUES
+(21213,0,0,'Please unlock the courtyard door.',2802,1,1,0,0,0,0,'',0),
+(21214,0,0,'Please unlock the courtyard door.',2802,1,1,0,0,0,0,'',0);
```
Killyana: Why there's 2 gossip_menu_option with the same text? (if there's a gossip for each faction so conditions are required)
This text : "Thanks, I'll follow you to the door." was moved to the DB ?

tkrokli: "Thanks, I'll follow you to the door." is a made-up text and not used by Blizzard as far as I could find. The broadcast_text ID 2802 is correct for these 2 NPCs. Please also note that there are 2 different `gossip_menu_option`.`menu_id`s, 1 for each NPC (21213 for alliance, 21214 for horde). You would probably say that it doesn't matter, but the 2x `gossip_menu`.`entry` values also need to link to 2 different `text_id`s, a.k.a. `npc_text`.`ID` 798 and 799. Those already existed in TDB 335.60:
```sql
SELECT * FROM `gossip_menu` WHERE `entry` IN (21213,21214);
```
```
 entry  text_id
------  ---------
 21213      798
 21214      799
```
The way I see it, I have followed the rules to the letter. If you can find sniffs contradicting this, I would think that the `broadcast_text` and `gossip_menu`  needs to be revised for these 2 NPCs.

PR #16243, opened 2016-01-12
[3.3.5] Core/Script: move Myranda the Hag's gossip option to DB

ghost:
Removing hardcoded text from the core script and into the DB:
"I am ready for the illusion, Myranda."

This quest is 3.3.5 _only_. Removed in patch 4.0.3a.

Addendum: enum ILLUSION_GOSSIP = 4773 , instead of "magic number".


tkrokli:
Tested and fully working. But phew, what a long chain of quests I had to do to get there. :-p

PR #16256, opened 2016-01-13
Core/Entities: Some corrections related to Movement Flags

elecyb:
- Redefined MOVEMENTFLAG_PITCH_UP and MOVEMENTFLAG_PITCH_DOWN as TURNING flags, these are sent by the client when aiming in cannons, for example, and can be sent in conjunction with MOVEMENTFLAG_ROOT
- Fixed visual bug of ghosts players "falling" under water due to incorrect remove of the MOVEMENTFLAG_WATERWALKING.


src/server/game/Server/WorldSession.cpp
```diff
@@ -845,28 +845,8 @@ void WorldSession::ReadMovementInfo(WorldPacket &data, MovementInfo* mi)
...
-            MOVEMENTFLAG_MASK_MOVING);
-    }
-    else
-    {
-        // Only remove here for non vehicles
-        REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_ROOT),
-            MOVEMENTFLAG_ROOT);
-    }
+    REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_ROOT) && mi->HasMovementFlag(MOVEMENTFLAG_MASK_MOVING),
+        MOVEMENTFLAG_ROOT);
```
Shauren: I asked this on the PR that last changed this function and I will ask it here
What is the point of removing the ROOT flag instead of *_MOVING? ROOT is "stronger" than movement and should be preveiting it

elecyb: Is the same, clients sending ROOT + MOVING flags is only possible with 3rd party programs, if the "cheater" manage to set both flags then he is able to unroot itself, the whole point of this function is to prevent client freezing for legal players.

Shauren: then remove _MOVING serverside, not ROOT.

Kittnz: Any update here @elecyb ? Seems like an easy fix for this to get merged.

Aokromes:
@elecyb .

If elecyb don't answers, anyone can takeover this work?

Aokromes:
anyone can takeover this work?

tkrokli:
If needed (last resort), I could create a PR to enable updating it with suggestions from the other TC developers, although I would need suggestions from the developers to implement new changes.

jackpoz:
As usual, PRs should focus on ONE single issue instead of 2 (pitch and waterwalking).
The waterwalking part looks fine and could be merged, the pitch one is still waiting for Shauren's comments to be implemented.

jackpoz:
Water walking part of this PR has been merged

PR #16257, opened 2016-01-13
DB/Creature: CREATURE_FLAG_EXTRA_NO_PLAYER_DAMAGE_REQ

Treeston:
DB/Creature: Add a new extra_flag CREATURE_FLAG_EXTRA_NO_PLAYER_DAMAGE_REQ (0x00200000) that awards players credit for killing the creature even if less than 50% of damage was dealt by players.

This can be used to fix quests such as "Heated Battle" or "If Valgarde Falls..." that are currently very hard to complete at the appropriate level due to "friendly" NPCs killing off the required targets too quickly for players to get 50% of damage in.


tkrokli:
Nice efforts all around, you've been busy. :+1:

LordUsagi:
Excellent work. Would this require a revert of 946fdbc06b46ee3a357ebfff1128d9d9b38d83d0? Then potentially add flags to the NPCs that cause issues instead?

Aokromes:
No, this is for non-helper mobs.

Treeston> Aokromes: yeah that was for guardians etc
Treeston> this is for stuff where random npcs shouldn't fuck credit
Treeston> for example the valgarde thing where there's guards fighting vrykul
Treeston> and you should help the guards

jackpoz:
are there cases where killer credit shouldn't be awarded if a friendly NPC dealt too much damage ?

Treeston:
@jackpoz Yes, any mob that isn't specifically flagged for it (this is blizzlike). This is to prevent players from dragging mobs to a city and having guards help them get the kill.

Treeston:
@LordUsagi like @Aokromes said, this is specifically for friendly creatures that are not linked to the player, for example the ones in Valgarde Keep I mention in the SQL.

Friendly creatures that are summoned by the player (this includes guardians summoned by trinket effects and the like) should be treated as "player damage" and thus always decrease playerdamagereq.

LordUsagi:
@Treeston Thanks, it just looked like two very similar checks and I was thinking perhaps those creatures that were getting killed by "friendly summons" also maybe had the flag.

PR #16263, opened 2016-01-14
DB Fix SQL script delete to avoid GUID collision [3.3.5]

NathanHandley:
In performing a local update, I noticed that the /world/2015_12_20_00_world.sql wasn't performing a wide enough delete before insert.  If this is ran twice, it'll fail on the last insert of the first set due to a unique collision.


Kittnz:
This is already been fixed in 8dd072722ab523b611fa65e8bd6b84653dd7cacb

PR #16264, opened 2016-01-14
Looting/Group: Do not display looter in tooltip when FFA loot is enabled

Treeston:
Just a quickie. Do not display the "Loot: <name>" tag on dead creatures while group has FFA loot on.


OScriptTC:
Works just fine.

PR #16265, opened 2016-01-14
Clean-up DYNFLAG_LOOTABLE update behavior for group member quest items

Treeston:
Creature/Loot: Do not display "lootable" sparkles for player even if others in the group still have quest items to be looted from creature.

Repro for issue this fixes:
1. Have two characters (A and B) grouped, set loot to FFA
2. Have character B on a quest that requires quest items from mobs that are only visible while on quest
3. Kill a mob that drops character B's quest item
4. Loot all visible items with character A. The quest item will remain, as A cannot see it
5. The mob will still sparkle for A, even though there is no loot remaining that A could get.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -17517,6 +17517,8 @@ bool Player::isAllowedToLoot(const Creature* creature)
     const Loot* loot = &creature->loot;
     if (loot->isLooted()) // nothing to loot or everything looted.
         return false;
+    if (!loot->hasItemForAll() && !loot->hasItemFor(this)) // no loot in creature for this player
+        return false;
 
     if (loot->loot_type == LOOT_SKINNING)
         return creature->GetSkinner() == GetGUID();
--- a/src/server/game/Handlers/LootHandler.cpp
+++ b/src/server/game/Handlers/LootHandler.cpp
@@ -370,13 +370,10 @@ void WorldSession::DoLootRelease(ObjectGuid lguid)
                 loot->roundRobinPlayer.Clear();
 
                 if (Group* group = player->GetGroup())
-                {
                     group->SendLooter(creature, NULL);
-
-                    // force update of dynamic flags, otherwise other group's players still not able to loot.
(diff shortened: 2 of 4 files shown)


OScriptTC:
Mob doesn't spark, but you can still hover over it and see a brown bag. When clicked on it opens an empty loot table.

Treeston:
Yeah, I know. Trying to figure out what's causing that.

LordUsagi:
That's because 
    creature->RemoveFlag(UNIT_DYNAMIC_FLAGS, UNIT_DYNFLAG_LOOTABLE);
isn't called after dyn flags are removed. It doesn't do anything. Even if it did, it'd remove the lootable flag for all players I believe, not just the person who couldn't see the quest items.

Treeston:
The reason why the PR adjusts `isAllowedToLoot` behavior is this check in [Unit.cpp](https://raw.githubusercontent.com/TrinityCore/TrinityCore/3.3.5/src/server/game/Entities/Unit/Unit.cpp) line 17803:
```c++
if (!target->isAllowedToLoot(creature))
    dynamicFlags &= ~UNIT_DYNFLAG_LOOTABLE;
```
Which dynamically removes DYNFLAG_LOOTABLE for any player (and only those players) that are not allowed to loot the creature. However, this does not appear to remove the right-click functionality. I am still trying to find out what flag controls whether a corpse is right-clickable.

LordUsagi:
I don't believe UNIT_DYNFLAG_LOOTABLE is being removed properly in this case

Treeston:
That's because they aren't. Dynamic flags are flags that are different for each player - the creature should be _LOOTABLE for one player, while not being _LOOTABLE for another.

Thus, _LOOTABLE should not be removed from the creature itself until it's no longer lootable for _any_ player.

Naios:
Any news about this?

Treeston:
I still have no idea about the issue mentioned above. Unless we find someone that understands how lootable flags on the client work, this won't be finished.

jackpoz:
I don't think there will be further tests for this PR so please merge it if it's working to you

PR #16267, opened 2016-01-14
Core/Spells: Fixed Beacon of Light not healing pet(s)

### LINKED ISSUE #15609: [3.3.5a] Beacon of Light is not healing pet
Beacon of Light is not healing pet.

Revision: 747a3f1d38abdb9b711877c168243f0d398b615f

Goatform:
Closes #15609

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -6382,8 +6382,10 @@ bool Unit::HandleDummyAuraProc(Unit* victim, uint32 damage, AuraEffect* triggere
                 }
                 else
                 {    // Check Party/Raid Group
-                    if (Group* group = IsPet() ? GetOwner()->ToPlayer()->GetGroup() : ToPlayer()->GetGroup())
+                    if (Group* group = ToPlayer()->GetGroup())
+                    {
                         for (GroupReference* itr = group->GetFirstMember(); itr != NULL; itr = itr->next())
+                        {
                             if (Player* member = itr->GetSource())
                             {
                                 // check if it was heal by paladin which cast this beacon of light
@@ -6404,6 +6406,8 @@ bool Unit::HandleDummyAuraProc(Unit* victim, uint32 damage, AuraEffect* triggere
                                     break;
                                 }
                             }
+                        }
+                    }
                 }
 
                 if (triggered_spell_id && beaconTarget)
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -6382,27 +6382,28 @@ bool Unit::HandleDummyAuraProc(Unit* victim, uint32 damage, AuraEffect* triggere
                 }
                 else
                 {    // Check Party/Raid Group
-                    if (Group* group = ToPlayer()->GetGroup())
```
jackpoz: did the old code trigger a NULL dereference exception at this point when IsPet() returned true ?

jackpoz: Actually, how can IsPet() return true here inside the else of https://github.com/Goatform/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L6375 check ? Was this change tested ingame AND debugged with both cases IsPet() true/false ?

Goatform: I tested it and I was using this fix for a long time and none found any issue related to it.

jackpoz: I wrote above what's wrong with this code, there is dead code in it https://en.wikipedia.org/wiki/Dead_code

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -6382,27 +6382,28 @@ bool Unit::HandleDummyAuraProc(Unit* victim, uint32 damage, AuraEffect* triggere
...
-                    {
+                    if (Group* group = IsPet() ? GetOwner()->ToPlayer()->GetGroup() : ToPlayer()->GetGroup())
                         for (GroupReference* itr = group->GetFirstMember(); itr != NULL; itr = itr->next())
-                        {
```
Treeston: Code style says brackets should be used whenever there is more than one line in the code block.

src/server/game/Entities/Unit/Unit.cpp
Treeston: @Goatform The line will never be reached for pets, there is a check for typeid being `TYPEID_PLAYER` further up.

Goatform:
@jackpoz & @Treeston check it now.

jackpoz:
It looks better now

joschiwald:
disagree, whole beacon of light code is ugly

cemak:
check guardian too ?

PR #15008, opened 2015-07-04
Scripts/Ulduar: implement Thorim Encounter

joschiwald:
TODO:
- [x] Loot (nh and hm is swapped?)
- [x] Sif's blizzard 
- [x] Known issue: Thorim targets (and turns! while on balcony, this didn't happen before)
- [x] and everything that need to be done before merge

i will merge and squash the commits myself when it is ready to be merged
(pls ignore stuff that should not be there)

Issue #15151, opened 2015-07-23
IsWithinDistInMap vs SPELL_EFFECT_APPLY_AREA_AURA_*


killradio:
Rev:  https://github.com/TrinityCore/TrinityCore/commit/d20ed095b143a9ab6238242b192dfdda9f51d0cf (fresh compile)
DB: TDB 335.59

Problem: Seems like maximum distance calculation for these spells is broken. From bool WorldObject::_IsWithinDist:
```cpp
    float sizefactor = GetObjectSize() + obj->GetObjectSize();
    float maxdist = dist2compare + sizefactor;
```
Logically its seems right, *owner_brounding_radius + spell_radius + target_bounding_radius*. But in game, seems like owner's radius not needed. Especially you can see this wrong calculations for spells with area animation, such as Anti-Magic Zone (Id: 52894), used by Koltira Deathweaver in DK's quests chain - it has 5yd radius and purple "bubble" which indicates actually this zone, but unit/target can be affected by this aura far away from zone like 8yd.

Video from off: https://youtu.be/dSax8Z-6WT4?t=1m20s

pete318:
Well, maybe try changing it to see if it exactly lines up. Since you've isolated a potential solution already.

Of course that's not the end of the story. _IsWithinDist and parent calls are used in many places. The next task is to see if all of those are wrong. If so, the change is fine. If not, another solution (optional parameter to the _IsWithinDist call chain perhaps) is needed.

joschiwald:
The whole target selection is wrong

killradio:
@pete318 Already changed and range looks ok, tested multiple spells with .dist command.
https://gist.github.com/killradio/43e71952bf2ab0e388d3

Aokromes:
@killradio (or anyone else) can you create a PR for this?

tkrokli:
It is unlikely that killradio will make a PR, seeing that he does not have any of the current TC branches forked to his repository. I can make a PR based on his suggestion from a year ago, hoping that it will be possible to fit his code in the current source, but I will need constructive feedback on changes.

Shauren:
No, that patch is not good

killradio:
@tkrokli check this pr https://github.com/TrinityCore/TrinityCore/pull/16290

tkrokli:
@killradio : Well, thanks for the pointer, but I'm not yet fluent enough in coding to discover what I can use there. Do you mind showing me more directly what your patch needs to work in current source?
[...]

Issue #3923, opened 2011-11-11
PBAOEs having less radius than what tooltip says


PKX:
Core revision: 583f6575e1976e177f6c0234bce1914cc91280ae
Db: TDB 335.11.43 + every update in this core revision

Afaik this bug is present in every revision.

> Steps to reproduce the bug (having a gm account):
> 1. Find an attackable npc which stays in place and doesn't move (I used Booty Bay's guards);
> 2. using command .distance, position yourself 9.9y away from that npc (use 3D distance, NOT exact 3D);
> 3. cast an AOE of your choice which has 10y radius. For GTAOEs you can use ".cast #spellid" to cast it precisely at your coordinates.

What should happen: the npc is hit by the aoe.
What appens instead: the npc is not hit by the aoe (is out of range).

> Steps to reproduce the bug (not having a gm account):
> 1. Take a rogue and a friend warrior (or whatever class han a 10y radius AOE), bring them to Gurubashi Arena (if they are of the same faction);
> 2. use the range check from rogue's untalented Blind spell to position yourself between 9-10y yards away from the warrior (note: DO NOT use talented Sap to do this, it won't work);
> 3. make the warrior cast Demoralizing Shout: you won't be hit.

By using .distance command and making some tries you'll discover that the AOE has an actual radius of 8.5y instead of 10y.
After some more tests you can find that **every AOE has 1.5y radius less than the one written in tooltip**.

Now, I know that every Unit has a fixed value (called COMBAT_REACH or MELEE_REACH if i remember right) which is used in calculating distance between 2 Unit and doing range checks for spells using a similar formula:

```
FinalRange = ExactDistanceBetweenUnits(Caster, Target) + Caster.MELEE_REACH + Target.MELEE_REACH
```

MELEE_REACH for Players is exactly 1.5y so my question is: Is this a bug or is normal because AOEs should not consider caster's MELEE_REACH? (if you think about it, an AOE is centered in a particular spot determined by coordinates so there is no reason why a point of size 0 should have a MELEE_REACH).

Update: this is bugged only for PBAoE, other AoEs are ok this way. The bug is that they don't count caster's size in rangechecks.
##

chaodhib:
If I'm not mistaken, this issue has been fixed either long time ago or by #16290.

Jildor:
Hi!
[...]

PR #16290, opened 2016-01-16
Core/Spells: fix wrong distance calculations in AoE spells [Needs testing]

chaodhib:
This PR attempts to fix an issue encountered in several places including here: https://github.com/TrinityCore/TrinityCore/pull/15008 regarding AoE spells. The AoE calculations are broken. The size of the aura's dynamic object and the size of the target are taken into account which shouldn't happen.

Changes:
- the calculation doesn't take into account the size of the dynamic object (the object at the center of the aoe) anymore
- the calculation doesn't take into account the size of the player anymore
- the distance calculation is done now in 2D.

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2230,9 +2230,9 @@ void WorldObject::GetContactPoint(const WorldObject* obj, float &x, float &y, fl
     GetNearPoint(obj, x, y, z, obj->GetObjectSize(), distance2d, GetAngle(obj));
 }
 
-float WorldObject::GetObjectSize() const
+float WorldObject::GetObjectSize() const // this method name is missleading. the returned value is only dependant on the combat reach. a unit's object size is something different.
 {
-    return (m_valuesCount > UNIT_FIELD_COMBATREACH) ? m_floatValues[UNIT_FIELD_COMBATREACH] : DEFAULT_WORLD_OBJECT_SIZE;
+    return (m_valuesCount > UNIT_FIELD_COMBATREACH) ? m_floatValues[UNIT_FIELD_COMBATREACH] : 0.0f;
 }
 
 void WorldObject::MovePosition(Position &pos, float dist, float angle)
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1113,9 +1113,11 @@ float WorldObject::GetDistanceZ(const WorldObject* obj) const
     return (dist > 0 ? dist : 0);
 }
 
-bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D) const
+bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D, bool incOwnRadius, bool incTargetRadius) const
 {
-    float sizefactor = GetObjectSize() + obj->GetObjectSize();
+    float sizefactor = 0;
+    sizefactor += incOwnRadius ? GetObjectSize() : 0;
+    sizefactor += incTargetRadius ? obj->GetObjectSize() : 0;
     float maxdist = dist2compare + sizefactor;
 
     if (GetTransport() && obj->GetTransport() && obj->GetTransport()->GetGUID().GetCounter() == GetTransport()->GetGUID().GetCounter())
@@ -1226,9 +1228,9 @@ bool WorldObject::IsWithinDist(WorldObject const* obj, float dist2compare, bool 
     return obj && _IsWithinDist(obj, dist2compare, is3D);
 }
 
-bool WorldObject::IsWithinDistInMap(WorldObject const* obj, float dist2compare, bool is3D /*= true*/) const
+bool WorldObject::IsWithinDistInMap(WorldObject const* obj, float dist2compare, bool is3D /*= true*/, bool incOwnRadius /*= true*/, bool incTargetRadius /*= true*/) const
 {
-    return obj && IsInMap(obj) && InSamePhase(obj) && _IsWithinDist(obj, dist2compare, is3D);
+    return obj && IsInMap(obj) && InSamePhase(obj) && _IsWithinDist(obj, dist2compare, is3D, incOwnRadius, incTargetRadius);
 }
 
 bool WorldObject::IsWithinLOS(float ox, float oy, float oz, VMAP::ModelIgnoreFlags ignoreFlags) const
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1120,19 +1120,19 @@ bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool
     sizefactor += incTargetRadius ? obj->GetCombatReach() : 0.0f;
     float maxdist = dist2compare + sizefactor;
 
-    const Position* a = this;
-    const Position* b = obj;
+    Position const* thisOrTransport = this;
+    Position const* objOrObjTransport = obj;
 
     if (GetTransport() && obj->GetTransport() && obj->GetTransport()->GetGUID().GetCounter() == GetTransport()->GetGUID().GetCounter())
     {
-        a = &m_movementInfo.transport.pos;
-        b = &obj->m_movementInfo.transport.pos;
+        thisOrTransport = &m_movementInfo.transport.pos;
+        objOrObjTransport = &obj->m_movementInfo.transport.pos;
     }
 
     if (is3D)
-        return a->IsInDist(b, maxdist);
+        return thisOrTransport->IsInDist(objOrObjTransport, maxdist);
     else
-        return a->IsInDist2d(b, maxdist);
+        return thisOrTransport->IsInDist2d(objOrObjTransport, maxdist);
 }
 
 bool WorldObject::IsWithinLOSInMap(const WorldObject* obj, VMAP::ModelIgnoreFlags ignoreFlags) const
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1116,33 +1116,23 @@ float WorldObject::GetDistanceZ(const WorldObject* obj) const
 bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D, bool incOwnRadius, bool incTargetRadius) const
 {
     float sizefactor = 0;
-    sizefactor += incOwnRadius ? GetObjectSize() : 0;
-    sizefactor += incTargetRadius ? obj->GetObjectSize() : 0;
+    sizefactor += incOwnRadius ? GetObjectSize() : 0.0f;
+    sizefactor += incTargetRadius ? obj->GetObjectSize() : 0.0f;
     float maxdist = dist2compare + sizefactor;
+
+    const Position* a = this;
+    const Position* b = obj;
 
     if (GetTransport() && obj->GetTransport() && obj->GetTransport()->GetGUID().GetCounter() == GetTransport()->GetGUID().GetCounter())
     {
-        float dtx = m_movementInfo.transport.pos.m_positionX - obj->m_movementInfo.transport.pos.m_positionX;
-        float dty = m_movementInfo.transport.pos.m_positionY - obj->m_movementInfo.transport.pos.m_positionY;
-        float disttsq = dtx * dtx + dty * dty;
-        if (is3D)
-        {
-            float dtz = m_movementInfo.transport.pos.m_positionZ - obj->m_movementInfo.transport.pos.m_positionZ;
-            disttsq += dtz * dtz;
-        }
-        return disttsq < (maxdist * maxdist);
+        a = &m_movementInfo.transport.pos;
+        b = &obj->m_movementInfo.transport.pos;
     }
 
-    float dx = GetPositionX() - obj->GetPositionX();
-    float dy = GetPositionY() - obj->GetPositionY();
-    float distsq = dx*dx + dy*dy;
     if (is3D)
-    {
-        float dz = GetPositionZ() - obj->GetPositionZ();
-        distsq += dz*dz;
-    }
-
-    return distsq < maxdist * maxdist;
+        return a->IsInDist(b, maxdist);
+    else
+        return a->IsInDist2d(b, maxdist);
 }
 
 bool WorldObject::IsWithinLOSInMap(const WorldObject* obj, VMAP::ModelIgnoreFlags ignoreFlags) const
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1108,7 +1108,7 @@ InstanceScript* WorldObject::GetInstanceScript()
 float WorldObject::GetDistanceZ(const WorldObject* obj) const
 {
     float dz = std::fabs(GetPositionZ() - obj->GetPositionZ());
-    float sizefactor = GetObjectSize() + obj->GetObjectSize();
+    float sizefactor = GetCombatReach() + obj->GetCombatReach();
     float dist = dz - sizefactor;
     return (dist > 0 ? dist : 0);
 }
@@ -1116,8 +1116,8 @@ float WorldObject::GetDistanceZ(const WorldObject* obj) const
 bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D, bool incOwnRadius, bool incTargetRadius) const
 {
     float sizefactor = 0;
-    sizefactor += incOwnRadius ? GetObjectSize() : 0.0f;
-    sizefactor += incTargetRadius ? obj->GetObjectSize() : 0.0f;
+    sizefactor += incOwnRadius ? GetCombatReach() : 0.0f;
+    sizefactor += incTargetRadius ? obj->GetCombatReach() : 0.0f;
     float maxdist = dist2compare + sizefactor;
 
     const Position* a = this;
@@ -1151,31 +1151,31 @@ bool WorldObject::IsWithinLOSInMap(const WorldObject* obj, VMAP::ModelIgnoreFlag
 
 float WorldObject::GetDistance(const WorldObject* obj) const
 {
-    float d = GetExactDist(obj) - GetObjectSize() - obj->GetObjectSize();
+    float d = GetExactDist(obj) - GetCombatReach() - obj->GetCombatReach();
     return d > 0.0f ? d : 0.0f;
 }
 
 float WorldObject::GetDistance(const Position &pos) const
 {
-    float d = GetExactDist(&pos) - GetObjectSize();
+    float d = GetExactDist(&pos) - GetCombatReach();
     return d > 0.0f ? d : 0.0f;
 }
 
 float WorldObject::GetDistance(float x, float y, float z) const
 {
-    float d = GetExactDist(x, y, z) - GetObjectSize();
+    float d = GetExactDist(x, y, z) - GetCombatReach();
     return d > 0.0f ? d : 0.0f;
 }
 
 float WorldObject::GetDistance2d(const WorldObject* obj) const
 {
-    float d = GetExactDist2d(obj) - GetObjectSize() - obj->GetObjectSize();
+    float d = GetExactDist2d(obj) - GetCombatReach() - obj->GetCombatReach();
     return d > 0.0f ? d : 0.0f;
 }
 
 float WorldObject::GetDistance2d(float x, float y) const
 {
-    float d = GetExactDist2d(x, y) - GetObjectSize();
+    float d = GetExactDist2d(x, y) - GetCombatReach();
     return d > 0.0f ? d : 0.0f;
 }
 
@@ -1195,22 +1195,22 @@ bool WorldObject::IsInMap(const WorldObject* obj) const
 
 bool WorldObject::IsWithinDist3d(float x, float y, float z, float dist) const
 {
-    return IsInDist(x, y, z, dist + GetObjectSize());
+    return IsInDist(x, y, z, dist + GetCombatReach());
 }
 
 bool WorldObject::IsWithinDist3d(const Position* pos, float dist) const
 {
-    return IsInDist(pos, dist + GetObjectSize());
+    return IsInDist(pos, dist + GetCombatReach());
 }
 
 bool WorldObject::IsWithinDist2d(float x, float y, float dist) const
 {
-    return IsInDist2d(x, y, dist + GetObjectSize());
+    return IsInDist2d(x, y, dist + GetCombatReach());
 }
 
 bool WorldObject::IsWithinDist2d(const Position* pos, float dist) const
 {
-    return IsInDist2d(pos, dist + GetObjectSize());
+    return IsInDist2d(pos, dist + GetCombatReach());
 }
 
 bool WorldObject::IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D /*= true*/) const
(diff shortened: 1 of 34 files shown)


src/server/game/Entities/Object/Object.cpp
```diff
@@ -1141,6 +1143,11 @@ bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool
     return distsq < maxdist * maxdist;
 }
 
+bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D) const
+{
+    return _IsWithinDist(obj, dist2compare, is3D, true, true);
```
jackpoz: isn't this the same as https://github.com/TrinityCore/TrinityCore/pull/16290/files#diff-29afb15ead2a8204cac07d94f074f260R1114 with the last 2 arguments are optional ?

jackpoz:
does changing the dist calc to 2D mean a very high Z diff will still make the AoE spell affect the target ? is that how it's supposed to work ?

src/server/game/Entities/Object/Object.cpp
chaodhib: WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D) const is overriden in GameObject. I didn't want to rewrite that method because I'm not sure what it does exactly/ how it does it.

So I created a separate method in order to not affect the rest of the code. Maybe not the best approach..

Treeston:
You tracked down that bug? Nice :+1:

chaodhib:
@jackpoz I'm beginning to have doubts. I'm not sure

jackpoz:
I have no idea how your code behaves with high Z diff and how was the blizzlike behavior but it's worth to investigate

chaodhib:
it's not the main concern of this PR. I could drop it. But yeah, That would be also interesting to know.

Aokromes:
As for frost nova it can hit flying player few around 6-8 yards over floor.

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1111,9 +1111,11 @@ float WorldObject::GetDistanceZ(const WorldObject* obj) const
     return (dist > 0 ? dist : 0);
 }
 
-bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D) const
+bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D, bool incOwnRadius, bool incTargetRadius) const
 {
-    float sizefactor = GetObjectSize() + obj->GetObjectSize();
+    float sizefactor = 0;
+    sizefactor += incOwnRadius ? GetObjectSize() : 0;
+    sizefactor += incTargetRadius ? obj->GetObjectSize() : 0; // sie
```
jackpoz: little typo here "// sie"

Treeston:
@Aokromes can you check if it's a sphere or just a xy ring plus z range?

E.g. is the height at which you can hit less or or the same at the edge vs near the center?

Aokromes:
Can't anymore, ptr is down, and it will very likely will be down for months.

Treeston:
Could someone with a retail account check this then, please? Only thing left before this can get merged.

If we can't, then @chaodhib just assume it's a 2d check plus z diff check (not sphere, just check if abs(targetz-ownz) <= aoe radius in addition existing 2d check) and we merge that.

chaodhib:
I rather not rush on this PR. I know it's affecting a lot of encounters but I would like to have more data before we implement a new implementation.

Treeston:
The difference between xyz sphere and xy sphere + z range is miniscule in 99% of cases.

jackpoz:
is the 2D/3D the only unclear thing ? it could be restored to be 3D so the rest can be merged

Treeston:
Just make it a cylinder imo, that's what it "feels like" on retail (super inaccurate, I know, honestly it makes no practical difference).

chaodhib:
I agree with the cylinder part, except it wont work great in a non flat area.

Currently, we have this: (which is not precise but ok)
gtaoe cylindar and sphere non flat2b
Using a cylinder, we'll have this: (which is bad)
gtaoe cylindar and sphere non flat2a
A better solution would be this:
gtaoe cylindar and sphere non flat2c
which is a "cylinder above the ground". So we'll need to use MMAPs for this.

Or we could use a double cylinder (one above and one under the horizontal plane). Would be less precise but would work too. The simpler calculations would be good for the speed. However, we need to ask ourselves if this approximation is good enough. (the answer to that question depends on the maximum slope a player or NPC can go upward. if that value is low enough, we'll never encounter the situation in the picture below and therefore we'll be fine)
gtaoe cylindar and sphere non flat2d

My point is, the 2D/3D subject is harder than it looks. I prefer to drop it from this PR and focus on the main issue.

Regarding the main issue: Quick reminder; AoEs can be divided into 2 categories: GTAoE (ground targeted AoE, example: Blizzard, Shadowfury,...) and PBAoE (proximity based AoE. example: Frost Nova, the hunter trap I suppose, etc...). Can we say that:
- in case of a GTAoE, nor the target's radius nor the caster's radius are to be taken into account in the distance calculation?
- in case of a PBAoE, only the caster's radius are to be taken into account?

In both case, do we have proof?

PS: there seems to be an open issue linked to this PR: https://github.com/TrinityCore/TrinityCore/issues/3923

jackpoz:
The cylinder in 2nd screenshot should have same height as the sphere radius, that way the diff in volume between sphere and cylinder is very little

Treeston:
Target hitbox has to be taken into account btw. Otherwise, y'know, dragon bosses.

Came to mind during Sapphiron today :D

Keader:
Testing this PR in Halion...
Some spells still dont hit he in right distance.
Divine storm (8 yards - spellid:  53385) and Whirlwind (8 yards - spellid:  50622).

Using .dist on Boss:
The distance is: (3D) 6.415426 (2D) 6.415344 (Exact 3D) 25.915426 (Exact 2D) 25.915344 yards. 
_Works, boss get damage._
But...
The distance is: (3D) 6.761860 (2D) 6.761780 (Exact 3D) 26.261860 (Exact 2D) 26.261780 yards. 
_Dont work, boss dont get any damage._

Keader:
Any news about https://github.com/TrinityCore/TrinityCore/pull/16290#issuecomment-196499053 ?

sirikfoll:
@chaodhib , must be due to the 3d->2d thing, that this issue happens:
go to `.go xyz -553 2200 51 631` and use death knight's spell Death and Decay, the spell will hit npcs that are on the upper floor, this seems to occur for any spells with `SPELL_EFFECT_PERSISTENT_AREA_AURA`.

chaodhib:
I'm working on this but I've nothing concrete yet. I'm doing sniff analysis to determine the formulas used in retail to calculate the ranges. I got tons of data but nothing makes much sense to me so far. I'll post the data soon.

Aokromes:
Maybe the way to go is ingame inspection while 2 characters does things.

chaodhib:
@Aokromes What do you mean? You mean doing tests on retail with 2 players while sniffing? If so, I agree.

I managed to get the right formulas for the auto attack attack, melee spells but for aoes, I can't use the same trick since the client has no aoe range validation. 

Instead what I did was wrote a software that analyze sniffs and look for times when an unit is hit by a spell at a distance higher than the range of the spell. I managed to have a working version and got some data but I've found any way to interpret these data. I'll dump as much data as I can when I'll find time, probably not before the end of this month.

On the other hand, taking the approach of doing tests on a live retail server would be probably faster and better..

Aokromes:
I mean with 3, 2 doing test and a 3rd looking the tests.

chaodhib:
When doing the test, we need the exact positions (of the players and the spell origin), which can only be obtained by parsed sniffs. Therefore, only 2 players with at least one of them sniffing is enough I think

tkrokli:
@chaodhib : good job working on this. Do you have any opinions to share in issue #15151 ? :-)

chaodhib:
AoE spells are more complicated than I thought. I think that the formula that determines if a unit is in range of an AoE actually depends on the ImplicitTarget flags of the aoe spell.

My current understanding is that:
- for GTAoEs, the formula is `range = spell radius + caster combat reach + target combat reach`
- for PBAoE, the formula is `range = spell radius`
- for AoEs created by dynamic objects, the formula is : `range = spell radius`. 

To make things smoother, this PR will only fix dynObj based AoEs.

Treeston:
This would prevent mages from hitting, say, dragon bosses with Arcane Explosion (it's PBAoE, which wouldn't account for hitbox). I agree for GTAoE and dynobj AoEs, but not for PBAoE.

chaodhib:
@Treeston My bad. I mistakenly reversed the 2. I fixed it now.

Treeston:
Hm, that means that DnD would only hit a dragon boss if you cast it on its center, not its nose? Seems weird, but I can't remember how it worked.

Should be easily to verify on retail though.

chaodhib:
DnD uses a dynObj

Treeston:
Right. What's an example of GTAoE? Did you test target hitbox for dynobj AoE separately?

chaodhib:
Actually strike that, even if DnD uses a dynObj, it's the same formula as GTAoEs, so yeah, if i'm right, you would need to cast the spell pretty close to the center of the unit.

The reason I think that dynObj AoEs and GT AoEs share the same formula is spells like Flamestrike. This spell has 2 spell effect. One which uses TARGET_DEST_DYNOBJ_ENEMY (dynObj formula) for the dot component. And the other which uses TARGET_UNIT_DEST_AREA_ENEMY (GTAoE formula) for the direct damage component. That's not a full proof but it would make sense that the 2 spell effects have the same effective radius.

Treeston:
Yeah, but I'm wondering if they might disregard caster (dynobj) but take into account target (potentially large dragon boss).

Like I said, should be easy to test on retail - just walk up to Sapphiron or something and drop a DnD so it only overlaps the edge of his hitbox.

chaodhib:
@Treeston Yes. It's possible that it takes into account the target combat reach as well, though my current understand is that it does not.

Would be cool if someone with a retail account could test that (and record it if possible).

Treeston:
@Aokromes ?

Aokromes:
https://onedrive.live.com/redir.aspx?cid=83725776ab2e4745&resid=83725776AB2E4745!1874&parId=83725776AB2E4745!143&authkey=!AHJWo0hk4QSBwLw&ithint=video%2cmp4

Treeston:
So it accounts for target, but doesn't account for caster. Makes sense.

@chaodhib can we get this finished with that info?

chaodhib:
So, if I understan correctly, these are the formulas:

- for PBAoEs, the formula is `range = spell radius + caster combat reach + target combat reach`
- for GTAoEs, the formula is `range = spell radius + target combat reach`
- for AoEs created by dynamic objects, the formula is : `range = spell radius + target combat reach`

I made the change for dynObj based AoEs. I'll work on the rest later.

Treeston:
In more simple terms:
* Is it PBAoE? Then the AoE applies from edge of our hitbox.
* Is it non-PB AoE effect? Then the AoE applies as a circle from the destination point.

In both cases, any target whose hitbox overlaps the AoE will be hit.

Treeston:
@chaodhib is this ready for merge yet?

chaodhib:
@Treeston No, I need time to review this and I need to analyze more sniffs of woltk to confirm what was discussed here. Sadly, I won't have much available time in the following weeks.

Keader:
@chaodhib This PR needs update :)

chaodhib:
Since the combat radius of players is so small, I don't think the impact of this issue is big. Also, it's very hard to prove the correctness of the new formulas, using sniff data. And believe me, I tried.

Treeston:
Can we just make the logic assume what I stated in [this comment](https://github.com/TrinityCore/TrinityCore/pull/16290#issuecomment-237676040) (which is the most accurate guess we have), @chaodhib?

Then review & merge.

chaodhib:
Ok. Will do that.

Treeston:
Actually, doesn't that just mean that we should give dynobjects a hitbox radius of zero?

Aokromes:
Conflicting files
src/server/game/Entities/Object/Object.h
src/server/game/Grids/Notifiers/GridNotifiers.h
src/server/game/Spells/Auras/SpellAuras.cpp

chaodhib:
Alright, I'm more confident about the formulas that we came up with now (though still not 100% sure). I'm gonna start to work on the implementation.

Treeston:
> Actually, doesn't that just mean that we should give dynobjects a hitbox radius of zero?
Either way, awesome :+1:

chaodhib:
I think dynObj do not have a size on their own. Their size is the spell radius. Another way to see this is to say that their hitbox radius is zero, yeah. Either way, it's the same thing.

chaodhib:
Done. Ready for extensive testing and review.

Treeston:
:heart:

src/server/game/Spells/Auras/SpellAuras.cpp
```diff
@@ -2552,7 +2552,7 @@ void DynObjAura::FillTargetMap(std::unordered_map<Unit*, uint8>& targets, Unit*
         }
         else
         {
-            Trinity::AnyAoETargetUnitInObjectRangeCheck u_check(GetDynobjOwner(), dynObjOwnerCaster, radius, false, true);
+            Trinity::AnyAoETargetUnitInObjectRangeCheck u_check(GetDynobjOwner(), dynObjOwnerCaster, radius, NULL, false, true);
```
Keader: we not use nullptr instead NULL?

Treeston: :+1: `nullptr` please

chaodhib: I'm not sure. I thought it was the opposite? In existing code, it can see both.

Aokromes: NULLs are moving into nullptr

chaodhib: ok

chaodhib:
To synthesize, here are the changes:

**BEFORE:**
- for PBAoEs, `effective range = spell radius + target combat reach`
- for GTAoEs, `effective range = spell radius + target combat reach`
- for AoEs created by UnitAura, `effective range = spell radius + target combat reach` + caster combat reach
- for AoEs created by DynObjAura, `effective range = spell radius + target combat reach` + DEFAULT_WORLD_OBJECT_SIZE

**AFTER:**
- for PBAoEs, `effective range = spell radius + caster combat reach + target combat reach`
- for GTAoEs, `effective range = spell radius + target combat reach`
- for AoEs created by Auras, `effective range = spell radius + target combat reach`

Aokromes:
@chaodhib  irc :P

Treeston:
Alright, as per @Aokromes test results on IRC, these are the hit formulas:

```range = radius + target hitbox radius
if PBAoE:
    range += caster hitbox radius
target is affected iff ((distance2d(origin, target) <= range) AND (abs(origin.z-target.z) <= range))```

**Comment by chaodhib** [comment]
Could I get access to the data?

**Comment by chaodhib** [comment]
Done. Awaiting for test & review.

**Comment by Aokromes** [comment]
@chaodhib this can be merged? if yes and no one tells something to stop the merge i will merge it tomorrow.

**Comment by chaodhib** [comment]
@Aokromes Apparently, there is no negative feedback. So yeah, in my opinion, it's ready to be merged.

**Comment by ccrs** [comment]
correct codestyle first

**Comment by chaodhib** [comment]
@ccrs could you be more specific please?

**Comment by tkrokli** [comment]
I read through all of the changes in the files, but could not find any obvious codestyle issues, aside from minor column indent variations in some of the `#define`s in a couple of the `.h` files (but that's not important). I also noticed that `IsWithinDoubleVerticalCylindar` is a tiny typo, but that doesn't matter either because all 5 of them have the same name. So yes, need to have any codestyle issues pointed out.

File: src/server/game/Entities/Object/Position.h [hunk 4]
```diff
@@ -229,6 +229,32 @@ struct TC_GAME_API Position
...
+    /*
+    side view:
+     _____     
+    |     |     ^
+    |     |     | height
+    |_____|     v
+       ^
+       center
+    */
+    bool IsWithinVerticalCylindar(Position const* center, float radius, float height) const;
```
Treeston: This is unused and should be removed.

src/server/game/Entities/Object/Position.h
```diff
@@ -229,6 +229,32 @@ struct TC_GAME_API Position
...
+    |     |     ^
+    |     |     | height
+    |__v__|     v 
+    |  ^  |     ^                   
+    |     |     | height         
+    |_____|     v
+                     
+    The center is at v
+                     ^
+    */
```
Treeston: Just `dist2d < radius && abs(dz) < height` summarizes it well enough

src/server/game/Entities/Object/Position.h
```diff
@@ -229,6 +229,32 @@ struct TC_GAME_API Position
     }
 
     bool IsWithinBox(const Position& center, float xradius, float yradius, float zradius) const;
```
Treeston: Is this still used anywhere? If not, remove.

src/server/game/Entities/Object/Position.h
```diff
@@ -229,6 +229,32 @@ struct TC_GAME_API Position
...
+    |     |     | height
+    |__v__|     v 
+    |  ^  |     ^                   
+    |     |     | height         
+    |_____|     v
+                     
+    The center is at v
+                     ^
+    */
+    bool IsWithinDoubleVerticalCylindar(Position const* center, float radius, float height) const;
```
Treeston: Refactor `Cylindar` -> `Cylinder`

Treeston (approved):
Final CS adjustments, then merge.

src/server/game/Spells/Spell.cpp
```diff
@@ -1220,6 +1220,10 @@ void Spell::SelectImplicitAreaTargets(SpellEffIndex effIndex, SpellImplicitTarge
 
     radius *= m_spellValue->RadiusMod;
 
+    // if this is a proximity based aoe (Frost Nova, Psychic Scream, ...), include the caster's own combat reach
+    if(targetType.IsProximityBasedAoe())
```
ccrs: here

src/server/game/Spells/Spell.cpp
```diff
@@ -7761,8 +7765,20 @@ WorldObjectSpellAreaTargetCheck::WorldObjectSpellAreaTargetCheck(float range, Po
...
+    if (target->ToGameObject())
+    {
+        // isInRange including the dimension of the GO
+        bool isInRange = target->ToGameObject()->IsInRange(_position->GetPositionX(), _position->GetPositionY(), _position->GetPositionZ(), _range);
+        if (!isInRange)
+            return false;
+    }
+    else
+    {
+        bool isInsideCylinder = target->IsWithinDist2d(_position, _range) && std::abs(target->GetPositionZ() - _position->GetPositionZ()) <= _range;
```
ccrs: maybe a comment is better than using a local variable to explain what the check does, I dont know if there is a convention about this

src/server/game/Grids/Notifiers/GridNotifiers.h
```diff
@@ -1041,14 +1067,23 @@ namespace Trinity
                 if (_spellInfo && _spellInfo->HasAttribute(SPELL_ATTR3_ONLY_TARGET_PLAYERS) && u->GetTypeId() != TYPEID_PLAYER)
                     return false;
 
-                return i_funit->_IsValidAttackTarget(u, _spellInfo, i_obj->GetTypeId() == TYPEID_DYNAMICOBJECT ? i_obj : nullptr) && i_obj->IsWithinDistInMap(u, i_range);
+                if (!i_funit->_IsValidAttackTarget(u, _spellInfo, i_obj->GetTypeId() == TYPEID_DYNAMICOBJECT ? i_obj : nullptr))
+                    return false;
+
+                float searchRadius = i_range;
+                if (i_incOwnRadius) searchRadius += i_obj->GetCombatReach();
```
ccrs: here

src/server/game/Grids/Notifiers/GridNotifiers.h
```diff
@@ -965,7 +979,17 @@ namespace Trinity
...
 
-                return !_refUnit->IsHostileTo(u) && u->IsAlive() && _source->IsWithinDistInMap(u, _range);
+                if (_refUnit->IsHostileTo(u))
+                    return false;
+
+                if (!u->IsAlive())
+                    return false;
+
+                float searchRadius = _range;
+                if (i_incOwnRadius) searchRadius += _source->GetCombatReach();
```
ccrs: here

src/server/game/Grids/Notifiers/GridNotifiers.h
```diff
@@ -927,27 +927,41 @@ namespace Trinity
...
             bool operator()(Unit* u) const
             {
-                if (u->IsAlive() && i_obj->IsWithinDistInMap(u, i_range) && i_funit->IsFriendlyTo(u) && (!i_playerOnly || u->GetTypeId() == TYPEID_PLAYER))
-                    return true;
-                else
+                if (!u->IsAlive())
+                    return false;
+
+                float searchRadius = i_range;
+                if (i_incOwnRadius) searchRadius += i_obj->GetCombatReach();
```
ccrs: here

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1108,39 +1108,31 @@ InstanceScript* WorldObject::GetInstanceScript()
...
-bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D) const
+bool WorldObject::_IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D, bool incOwnRadius, bool incTargetRadius) const
 {
-    float sizefactor = GetObjectSize() + obj->GetObjectSize();
+    float sizefactor = 0;
+    sizefactor += incOwnRadius ? GetCombatReach() : 0.0f;
+    sizefactor += incTargetRadius ? obj->GetCombatReach() : 0.0f;
     float maxdist = dist2compare + sizefactor;
 
+    const Position* a = this;
```
ccrs: better use other name for the local variables (a & b), everything would be more readable

ariel-: Convention is lowerCamelCase for local variable, const after type not before.

src/server/game/Entities/Object/Position.h
chaodhib: it's used somewhere

src/server/game/Entities/Object/Position.h
chaodhib: ok

src/server/game/Entities/Object/Position.h
chaodhib: ok

src/server/game/Entities/Object/Position.h
chaodhib: ok

src/server/game/Spells/Spell.cpp
chaodhib: what is wrong here?

src/server/game/Spells/Spell.cpp
chaodhib: i disagree

src/server/game/Grids/Notifiers/GridNotifiers.h
chaodhib: you mean the statement on a different line then the if, right?

src/server/game/Entities/Object/Object.cpp
chaodhib: ok

src/server/game/Spells/Spell.cpp
Treeston: `if (`

src/server/game/Grids/Notifiers/GridNotifiers.h
Treeston: yes

src/server/game/Entities/Object/Object.cpp
chaodhib: any idea of names I could use for a and b?

chaodhib:
I think that's all

src/server/game/Spells/Spell.cpp
chaodhib: ty

src/server/game/Spells/Spell.cpp
ccrs: Given the past dramas I understand the... answer I guess 😄
But no worries, it was more of a headsup to open a debate about it since its one of the few times I've seen this type of codestyle in the core and actually anywhere else (using codestyle to show meaning instead of comments)

DDuarte: Avoid comments but use them when you *need* them.

https://blog.codinghorror.com/coding-without-comments/

> While comments are neither inherently good or bad, they are frequently used as a crutch. You should always write your code as if comments didn't exist. This forces you to write your code in the simplest, plainest, most self-documenting way you can humanly come up with.

PR #16314, opened 2016-01-17
Player/Quest: Fix GO group credit

Treeston:
Player/Quest: Credit group for gameobject activation if they are in proximity.

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -1353,7 +1353,15 @@ void GameObject::Use(Unit* user)
                         break;
                 }
 
-                player->KillCreditGO(info->entry, GetGUID());
+                if (Group* group = player->GetGroup())
+                {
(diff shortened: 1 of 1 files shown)


jackpoz:
are there GOs that shouldn't award kill credits to party members ? Otherwise this looks ready to be merged

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -1353,7 +1353,19 @@ void GameObject::Use(Unit* user)
                         break;
                 }
 
-                player->KillCreditGO(info->entry, GetGUID());
+                if (Group* group = player->GetGroup())
+                {
+                    GroupReference const* ref = group->GetFirstMember();
+                    while (ref)
```
DDuarte: Change to a for loop as in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Totem/Totem.cpp#L136-L141

PR #16327, opened 2016-01-19
Core/Scripts: Change door ID and name in Blackwing Lair

Exodius:
as per 7d454c66f70c19c8ff402fd126392e27700398ba


src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/blackwing_lair.h
```diff
@@ -59,7 +59,7 @@ enum CreatureIds
 enum GameObjectIds
 {
     GO_BLACK_DRAGON_EGG         = 177807,
-    GO_BOSSGATE01               = 175946,
+    GO_BOSSGATE01               = 176965,
```
joschiwald: pls update the name too

Exodius: What is the correct name? Don't have the DB installed and, to be honest, I don't have enough drive space to re-install everything...

MitchesD: @Exodius http://i.imgur.com/kiT81g7.png

tkrokli: OK, so maybe a name like `GO_PORTCULLIS` would be useful here.

Exodius:
@joschiwald Edited and squashed. Is it OK like this?

Exodius:
Or should it be a bit flexible and be named GO_RAZORGORE_PORTCULLIS?

tkrokli:
I think the name `GO_PORTCULLIS` is more flexible and that `GO_RAZORGORE_PORTCULLIS` is more specific.
But that is just my opinion. :)

PR #16334, opened 2016-01-20
Core/Build: Fix ZMQ library for Visual Studio 

Emtec:
temporary solution, when ZMQ publish prebuild with VC14 support this solution should be removed


DDuarte:
ZMQ compiled for VS 2013 is compatible with VS 2015.
You can set ZMQ_LIBRARY to `libzmq-v120-mt-4_0_4.lib` even if you compile TC for VS 2015, therefore including ZMQ in the repo is not necessary.

Thanks for the PR but we prefer to have some sources outside of the repo, the same way that MySQL sources were removed some time ago.

Emtec:
Thanks. I don't know why i've problem with zmq for VS 2013. Maybe I made missclick and I set .dll as ZMQ_LIBRARY. So thanks ;)

PR #16335, opened 2016-01-20
[3.3.5] Scripts/Commands: Small QoL adjustments

Treeston:
- .debug hostil now shows spawn ID (DBGUID) in addition to current GUID (so you can .go creature to it)
- .npc temp now takes an additional argument before the creature entry that determines whether the spawned creature instantly despawns upon death. Default is instant despawn (current behavior).
- Add .npc evade command.
- Add .pet level command.
- .server shutdown and .server restart fail if time is less than a config parameter (default = 30 seconds) and another player is connected. Add new `force` param to override this behavior.


Aokromes:
NPCs spawned from .npc add temp now despawn 2 minutes after death instead of instantly upon death to allow for looting

IMHO this must be a parameter.

how about don't allow shutdown / restart times under 30s if people is online  without a parameter (server restart/shutdown 10 force)

Treeston:
> NPCs spawned from .npc add temp now despawn 2 minutes after death instead of instantly upon death to allow for looting
>IMHO this must be a parameter.

It's expected behavior imo. Maybe a param to disable it?

> how about don't allow shutdown / restart times under 30s if people is online without a parameter (server restart/shutdown 10 force)

Sounds reasonable.

Shauren:
30 * IN_MILLISECONDS is not 2 minutes

jackpoz:
I wonder if GMs do events spawning bosses and then manually give unrelated loot and would not like this change in .npc add spawn

LordUsagi:
A param to disable loot would be a fantastic idea.

jackpoz:
cs_debug.cpp change can be merged, the cs_npc.cpp one still need to keep current behavior by default

Treeston:
Cleaned everything up by now.

Waiting on tests and review.

PR #16344, opened 2016-01-22
[3.3.5] Core/Scripts/DB: move core script 'npc_braug_dimspirit' to SAI

ghost:
Additional change: The answers have been changed from containing only names
to include the text `is my answer.` based on the lines in `broadcast_text`.
- created DB creature_text entries for Braug Dimspirit
- created DB gossip_menu_option entries for the NPC
- implemented NPC say lines in reply to wrong answers
- replaced core text with DB SAI script, added DB creature updates
- zone_stonetalon_mountains.cpp has now got only 1 quest script left

This PR is tested several times on a horde character (horde-only quest).
It now works as it was originally intended, with NPC response lines.

### DIFF
--- a/src/server/scripts/Kalimdor/zone_stonetalon_mountains.cpp
+++ b/src/server/scripts/Kalimdor/zone_stonetalon_mountains.cpp
@@ -19,12 +19,11 @@
 /* ScriptData
 SDName: Stonetalon_Mountains
 SD%Complete: 95
-SDComment: Quest support: 6627, 6523
+SDComment: Quest support: 6523
 SDCategory: Stonetalon Mountains
 EndScriptData */
 
 /* ContentData
-npc_braug_dimspirit
 npc_kaya_flathoof
 EndContentData */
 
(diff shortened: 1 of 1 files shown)


tkrokli:
I would like to make a posssible improvement on this quest by adding NPC say text (`creature_text`) to NPC 4489, Braug Dimspirit, by adding the following texts to `creature_text` and have him say something in reply to the selected answer, just like Parqual Fintallas in UC does in his part of the chain:
```sql
SELECT * FROM `broadcast_text` WHERE `ID` IN (8382,8386,8387,8388,8393);
```
```
    ID  Language  MaleText                                               VerifiedBuild
------  --------  -----------------------------------------------------  ---------------
  8382         0  Ah, Ysera the Dreamer.                                         18019
  8386         0  Neltharion, hmm? As good an answer as any, yes?                18019
  8387         0  Alexstrazsa, the Life-Binder. A magnificent creature.          18019
  8388         0  Malygos, is it?                                                18019
  8393         0  That is the wrong answer, $n. Be off with you.                 18019
```
At least, I think Braug should say `That is the wrong answer, $n. Be off with you.` when the wrong answer is clicked and he fires up the spell to teleport the player to Camp Taurajo.

I could not find a proper DB text that matches the gossip option "Nozdormu is my answer.", but then again, broadcast_text is from build 18019 (5.4.7), so maybe it would be just as possible to replace "Neltharion" with "Nozdormu" if those creature_text lines could be implemented. What do you others think?

tkrokli:
This PR is no longer in [WIP] status. Comments are welcome, if there is anything to comment on.

Killyana:
Add the broadcast on creature_text, the rest is fine.

tkrokli:
OK, will do. I will have to leave 1 `broadcast_text`.`ID` out of the list because I changed the name to make it fit with my usage of the creature say replies. (Yes, I know. I cheated a little bit.) Another solution would be, if all 5 of them need to have the ID, is to change the name from `Neltharion` to `Nozdormu` in `broadcast_text`.`ID` 8386, because the SAI reads the `broadcast_text`.`MaleText` directly when available. Tested it and I can't use `broadcast_text`.`ID` 8386 directly: `Neltharion, hmm? As good an answer as any, yes?` because it won't fit with my setup where the say text only applies when the answer is wrong. ...unless someone knows what the missing line for Nozdormu should be. Anyway, this is a 3.3.5a-only quest, so it should not be transferred to 6.x.

tkrokli:
OK, if nothing else needs to be changed, the PR is ready to use.

Treeston:
Gossip options shouldn't show while he's actively talking. You can simply spam all the options before he finishes kicking you out for the incorrect one to complete quest.

tkrokli:
OK, have to find a condition for it to disable the gossip or follow your suggestion in #trinity:
`easy way: set a data when gossip selected, reset it after punishment spell is cast`

tkrokli:
Exploits avoided by removing `npcflags` from the NPC while he talks, then restore again after spellcast.

Treeston:
Exploit still works. Flags need to be stored at end of the [2 second cast](http://www.wowhead.com/spell=6766/test-of-lore), not at start.

tkrokli:
Thanks, I will make the required change shortly. Just need to test another PR first.

Treeston:
The linked action happens at the same time as the action it is linked to, so the flags are restored at the same time as casting the spell begins.

I can select an incorrect answer, then open the gossip menu as the NPC starts channeling his "teleport to barrens" spell, then select the correct answer and receive credit for it before getting kicked out.

tkrokli:
Thanks again for precise feedback. Will use a different action or sourcetype for resetting the `npcflags`.

tkrokli:
I have checked and double-checked that the latest change
`"Braug Dimspirit - on target spellhit 'Test of Lore' - set npcflags Gossip & Questgiver"`
works as intended. No chance to interact before teleport is complete.

Treeston:
What happens if the player somehow is not available to be targeted anymore (the cast fails)? Would this cause the mob to be permanently stuck without flags until restart? :trollface:

(IMO, just use a linked action with SMART_ACTION_CREATE_TIMED_EVENT to reset the flags 3 seconds after the cast starts.)

PR #16345, opened 2016-01-22
[3.3.5][6.x] Core/Creatures: Reset unit flags/dynamic flags on Creature::SetDeathState(JUST_RESPAWNED)

Trisjdc:
Creatures should pretty much completely reset on respawn, and currently flags are only set on creation.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -1604,13 +1604,29 @@ void Creature::setDeathState(DeathState s)
 
         UpdateMovementFlags();
 
-        CreatureTemplate const* cinfo = GetCreatureTemplate();
-        SetUInt32Value(UNIT_NPC_FLAGS, cinfo->npcflag);
         ClearUnitState(uint32(UNIT_STATE_ALL_STATE & ~UNIT_STATE_IGNORE_PATHFINDING));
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1607,10 +1607,29 @@ void Creature::setDeathState(DeathState s)
         CreatureTemplate const* cinfo = GetCreatureTemplate();
         SetUInt32Value(UNIT_NPC_FLAGS, cinfo->npcflag);
         ClearUnitState(uint32(UNIT_STATE_ALL_STATE & ~UNIT_STATE_IGNORE_PATHFINDING));
-        SetMeleeDamageSchool(SpellSchools(cinfo->dmgschool));
+
+        if (!IsPet())
+        {
+            CreatureData const* creatureData = GetCreatureData();
+            if (CreatureTemplate const* cinfo = GetCreatureTemplate())
```
joschiwald: variable `cinfo` already defined

Trisjdc: Indeed. Fixed!

joschiwald: cinfo is never null

Trisjdc:
Anything wrong?

Issue #13787, opened 2014-12-27
[3.3.5][6.x] Respawntimer should start ticking straight after killing the unit + creature should be respawned as completely new unit with new guid $20 


ghost:
General Core issue. Revision while reporting this issue: 5d4f26d82028823f4672e370deaf0f6069fb225f

While I was playing on retail, something got my attention while I was killing some units in some starting zones. When I kill a unit and do not loot the corpse, the npc remains in world but also a new unit gets spawned after a while which is supposed to be the respawned unit.

So instead of removing the corpse from the world and re-use the guid, blizz creates completely new units.

The core currently respawns the unit by re-using the corpse instead of spawning a new one and destroy the old one after looting /  decay time expired.

So the progress should be:
1. Unit dies
2. Respawntimer to create a new npc at the spawn location starts
3. Create a completely new npc after timer expires.

So the old corpse is being completely ignored on retail. It just despawns and is completely removed from the world. This also explains why the respawned units in sniffs have different guids

This would also fix the annoying exploit that some players keep the corpses unlooted to prevent them respawning so the area stays empty after a while.
##

pete318:
The linked PR by me is an interim solution to the issue of insufficient creatures being around in busy areas. It's an interim solution that just adds a minimum number of spawns and then chooses a current corpse to revive using specific rules to minimize player annoyance.

If you look further down and indeed into the forum thread on the subject you'll see it moves onto this exact issue and some possible solutions are suggested.

I'm currently working on a design and ultimately a change to the pooling system that will address this. But, it will mean all creatures would need to be in pools to be covered by the new functionality.

I'm not convinced the cartoon could (safely) be done. e.g. you renumber the guid of the unit itself. But, I've not looked into that. I'd be just generating new guids. As far as I know SAI works on the template and not the creature guid itself. So, changing either way wouldn't hurt that (I think at least).

MrSmite:
This could be done relatively easy by randomly generating a GUID when the creature is spawned. The main issue is the possibility of duplicate GUIDs which could be overcome by generating them with a unix timestamp based on YYYYMMDDHHNNSSss (where NN = minutes, SS= seconds, ss=milliseconds).

I think the main problem is the amount of core rewrite that would be necessary to accomodate dynamic GUIDs would be daunting.

Aokromes:
3.x client can't handle a such high guid like 20.150.101.101.010

MrSmite:
[...]

PR #16346, opened 2016-01-22
Core/Creatures: Remove resetting of UNIT_FLAG_IMMUNE_TO_PC on combat clearing

Trisjdc:
This is most likely leftover code from those days where UNIT_FLAG_IMMUNE_TO_PC still allowed the creature to engage players if the player got close enough. Not the case anymore, and setting this flag can be confusing in scripts.

This fixes cases such as Violet Hold bosses becoming unattackable after a reset. I found out that Lich King's script relied on the shitty flag behavior, but there might be other scripts. Would like help to find those out!

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -11398,10 +11398,8 @@ void Unit::SetInCombatState(bool PvP, Unit* enemy)
         if (enemy)
         {
             if (IsAIEnabled)
-            {
                 creature->AI()->EnterCombat(enemy);
-                RemoveFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_PC); // unit has engaged in combat, remove immunity so players can fight back
-            }
+
             if (creature->GetFormation())
                 creature->GetFormation()->MemberAttackStart(creature, enemy);
         }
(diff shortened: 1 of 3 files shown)


Shauren:
The Lich King does not "rely" on this "shitty" behavior - this is blizzlike that he fully resets on every attempt and requires the intro sequence (not blizzlike part here is that instead of despawning he runs to start position)
It is probably is blizzlike for most npcs or switched per-npc

Trisjdc:
Yes @Shauren, he should fully reset on every attempt. If it's a respawn, I actually made another commit that would give the same behavior: https://github.com/Trisjdc/TrinityCore/commit/e532026626a915c7862157a297145379645b385d

P-Kito:
Shouldn't DespawnAtEvade be doing this?

Trisjdc:
What's 'DespawnAtEvade'?

Shauren:
@Trisjdc https://github.com/TrinityCore/TrinityCore/blob/6.x/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp#L587

Trisjdc:
Any respawn should have this behavior and fully reset unit flags. Makes all the more sense considering that with https://github.com/TrinityCore/TrinityCore/issues/13787, the respawned unit would be a completely new one.

Treeston:
This breaks Malygos. Script needs to be modified to properly remove IMMUNE_TO_PC on engage.

Aokromes:
@Trisjdc read https://github.com/TrinityCore/TrinityCore/pull/16346#issuecomment-183771667 
If Trisjdc don't answers, anyone can takeover this work?

Trisjdc:
If any other scripts break, it's just a matter of manually removing UNIT_FLAG_IMMUNE_TO_PC before the combat call.

w1sht0l1v3:
1st boss in toc raid resets instead of agroing

ccrs:
BRS - blackhand, BT, ToCrusader and who knows how many quests...

PR #16348, opened 2016-01-22
Core/InstanceScript: Allow more than one condition for boss doors to open

Trisjdc:
Useful in the case of Zul'Aman/Black Temple. Essentially allows for AND door conditions.

### DIFF
--- a/src/server/game/Instances/InstanceScript.cpp
+++ b/src/server/game/Instances/InstanceScript.cpp
@@ -181,20 +181,21 @@ void InstanceScript::UpdateDoorState(GameObject* door)
     if (range.first == range.second)
         return;
 
+    // more than one condition can be needed for the door to open
     bool open = true;
-    for (; range.first != range.second && open; ++range.first)
+    for (; range.first != range.second; ++range.first)
     {
         DoorInfo const& info = range.first->second;
         switch (info.type)
         {
             case DOOR_TYPE_ROOM:
-                open = (info.bossInfo->state != IN_PROGRESS);
+                open = open && (info.bossInfo->state != IN_PROGRESS);
                 break;
             case DOOR_TYPE_PASSAGE:
-                open = (info.bossInfo->state == DONE);
+                open = open && (info.bossInfo->state == DONE);
                 break;
             case DOOR_TYPE_SPAWN_HOLE:
-                open = (info.bossInfo->state == IN_PROGRESS);
+                open = open && (info.bossInfo->state == IN_PROGRESS);
                 break;
             default:
                 break;


joschiwald:
does it change anything? there is a `&& open` check in for loop, so once the condition fails the door will never open

Treeston:
@joschiwald at a glance, if two conditions are met, then the door will close (as opposed to only obeying the latter condition)

Trisjdc:
@joschiwald, it would essentially make this usable:

```
DoorData const doorData[] =
{
	{ GO_TEMPLE_DOOR,           DATA_SHADE_OF_AKAMA,    	DOOR_TYPE_PASSAGE },
	{ GO_TEMPLE_DOOR,           DATA_TERON_GOREFIEND,    	DOOR_TYPE_PASSAGE },
	{ GO_TEMPLE_DOOR,           DATA_GURTOGG_BLOODBOIL,    	DOOR_TYPE_PASSAGE },
    { GO_TEMPLE_DOOR,           DATA_RELIQUARY_OF_SOULS,    DOOR_TYPE_PASSAGE },
    { 0,                        0,                          DOOR_TYPE_ROOM } // END
};
```

src/server/game/Instances/InstanceScript.cpp
```diff
@@ -181,6 +181,7 @@ void InstanceScript::UpdateDoorState(GameObject* door)
     if (range.first == range.second)
         return;
 
+    // more than one condition can be needed for the door to open
     bool open = true;
     for (; range.first != range.second && open; ++range.first)
```
joschiwald: this line, once open is false you wil exit the for loop

Treeston: correct, this would need to be removed

joschiwald: it won't change anything

and iirc door handling is working as intended

Treeston: if the check were to be removed then the PR would work as advertised - multiple conditions for the same door would turn into an AND instead of only abiding the last condition.

I see no issue with making that change.

Treeston: I take that back and agree with @joschiwald, current code already does this.

Treeston:
Have you actually tested on 3.3.5 base if the behavior is "bugged" like you expect?

Trisjdc:
@Treeston: For this commit, nope. I just thought about the use case, looked at the code, and went "Ah, that wouldn't work". I checked the result after the change, which is working as I want. I didn't think of checking before.

Nonetheless code is probably a bit clearer this way. If not, close the pull.

Treeston:
@Trisjdc on current 3.3.5, open gets set to false the first time a condition isn't matched, and the loop then exits (the loop condition checks for `open`).

PR #16378, opened 2016-01-24
Core/Spells: Fixed a possible crash in spell_pal_light_s_beacon

Zedron:
Fixes a possible crash in spell_pal_light_s_beacon aurascript


joschiwald:
did you got such crash? actiontarget should not be null in this situation

Zedron:
in Unit::Kill, victim->ProcDamageAndSpell(NULL

jackpoz:
> No description provided.

Please provide a description

joschiwald:
this aura should not proc in Unit::Kill

please provide a crash log

Shauren:
@joschiwald it doesn't matter - crashing depending on database data (proc flag overrides are possible after all) is not a good thing.

PR #16395, opened 2016-01-26
Core/SAI: Change SmartScript::GetTargets to return an ObjectList instead of an ObjectList*

DDuarte:
Applied the same treatment to GetWorldObjectsInDist.

SmartAI's "ObjectList" is now a std::vector instead of a std::list because no where it was being used as an actual list.

Honestly no idea why the original idea insisted so much on using pointers. No more missing null checks or deletes.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScriptMgr.h
+++ b/src/server/game/AI/SmartScripts/SmartScriptMgr.h
@@ -1380,39 +1380,38 @@ struct SmartScriptHolder
 
 typedef std::unordered_map<uint32, WayPoint*> WPPath;
 
-typedef std::list<WorldObject*> ObjectList;
+typedef std::vector<WorldObject*> ObjectList;
 
 class ObjectGuidList
 {
-    ObjectList* m_objectList;
-    GuidList* m_guidList;
+    ObjectList m_objectList;
+    GuidList m_guidList;
     WorldObject* m_baseObject;
 
 public:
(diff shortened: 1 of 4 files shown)


src/server/game/AI/SmartScripts/SmartScriptMgr.h
```diff
@@ -1380,39 +1380,38 @@ struct SmartScriptHolder
...
+    ObjectList m_objectList;
+    GuidList m_guidList;
     WorldObject* m_baseObject;
 
 public:
-    ObjectGuidList(ObjectList* objectList, WorldObject* baseObject)
+    ObjectGuidList(ObjectList objectList, WorldObject* baseObject)
     {
-        ASSERT(objectList != NULL);
+        ASSERT(!objectList.empty());
```
jackpoz: what's wrong passing an empty list ? the old code didn't assert in this case, this whole assert should just be removed in the new code

ariel-:
Maybe the whole pointer thing was for avoiding list copies. Perhaps they should be `shared_ptr`s instead? (to 'forget' about proper GC)

jackpoz:
the whole idea of this PR is to remove pointers and use a simple list copied on assignment. when one wants to avoid list copies, the first question to ask is "why?" and there's no clear answer to that question in this case

DDuarte:
On hold.

PR #16404, opened 2016-01-26
Core/Scripts: SMSG_PLAY_ONE_SHOT_ANIM_KIT can now be used in waypoint_scripts

Lopfest:
DB devs will be able to create waypoints (waypoint_data) with waypoint_scripts where AnimKits are used.


Shauren:
How about no?
Where is the packet class for it?
Where is animKitId validation against dbc?

Lopfest:
How about now?
Im not using any AI, just pure waypoints. Thats why I needed waypoint_script

Shauren:
Still missing that validation on startup when loading the script table

Lopfest:
Ok added. I hope everything is correct now :)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -13429,6 +13429,17 @@ void Unit::SendDurabilityLoss(Player* receiver, uint32 percent)
     receiver->GetSession()->SendPacket(packet.Write());
 }
 
+void Unit::PlayOneShotAnimKitId(uint16 animKitId)
+{
+    if (animKitId && !sAnimKitStore.LookupEntry(animKitId))
```
jackpoz: is animKitId = 0 a valid id to be sent ?

DDuarte:
Looks good to me :shipit:

Lopfest:
in AnimKit.dbc there is no entry 0

DDuarte:
I would simply remove the check for a valid entry entirely (like the code before this PR allows). Let the client handle that. (But by all means, keep the startup data validation.)

Wait for someone else opinion on this before doing any change.

jackpoz:
Mine was a legit question, I have no idea either if 0 is meant to be sent sometimes or if the old check was not needed.

Shauren:
@DDuarte I disagree, the check is better there - since anyone can use this method with hardcoded id in c++ scripts

DDuarte:
So perhaps PlayOneShotAnimKitId can have a debug error log on invalid entries

DDuarte:
The last commit isn't related to this PR, is it?

This is one of the reasons why PRs should be done from specific branch, not master (6.x in this case).

Lopfest:
Yes ignore that (already deleted) commit.

I believe that debug error is not needed because the entries with non-existing AnimKit ID are not loaded during "void ObjectMgr::LoadScripts(ScriptsType type)"

Ascathor:
So, anything else that has to be done here ? This looks pretty much clean entirely.

Ulduar:
Play Cinematic too not bad to have. In SAI and SCRIPT_COMMAND.
For quests ofc.
I mean the following.
``` CPP
case SMART_ACTION_PLAYCINEMATIC:
        {
            ObjectList* targets = GetTargets(e, unit);
            if (!targets)
                break;

            for (ObjectList::const_iterator itr = targets->begin(); itr != targets->end(); ++itr)
            {
                if (!IsPlayer(*itr))
                    continue;

                (*itr)->ToPlayer()->SendCinematicStart(e.action.cinematic.entry);
            }

            delete targets;
            break;
        }
```
``` CPP
case SCRIPT_COMMAND_PLAYCINEMATIC:
                // Source must be Player.
                if (Player* player = _GetScriptPlayer(source, true, step.script))
                    player->SendCinematicStart(step.script->PlayCinematic.CinematicID);
                break;
```

PR #16428, opened 2016-01-30
Collision/MMap: Fix a memleak reported by valgrind

Carbenium:
Tiles which are owned by the mesh should have DT_TILE_FREE_DATA to ensure the deallocation of the memory on removal from the mesh.

Log:
==23516== 6,181,976 bytes in 6 blocks are definitely lost in loss record 15 of 15
==23516==    at 0x4C298A0: operator new[](unsigned long) (vg_replace_malloc.c:389)
==23516==    by 0x2052660: dtCustomAlloc(int, dtAllocHint) (Memory.h:11)
==23516==    by 0x29D7FDA: dtAlloc(int, dtAllocHint) (DetourAlloc.cpp:43)
==23516==    by 0x230AA70: MMAP::MMapManager::loadMap(std::string const&, unsigned int, int, int) (MMapManager.cpp:166)
==23516==    by 0x1DD585B: Map::LoadMMap(int, int) (Map.cpp:125)
==23516==    by 0x1DD7266: Map::LoadMapAndVMap(int, int) (Map.cpp:200)
==23516==    by 0x1DD93C6: Map::EnsureGridCreated_i(CoordPair<64u> const&) (Map.cpp:453)
==23516==    by 0x1DD9129: Map::EnsureGridCreated(CoordPair<64u> const&) (Map.cpp:429)
==23516==    by 0x1DED100: Map::GetGrid(float, float) (Map.cpp:2200)
==23516==    by 0x1DE0F4C: Map::GetAreaFlag(float, float, float, bool*) const (Map.cpp:2349)
==23516==    by 0x198521B: Map::GetZoneId(float, float, float) const (Map.h:353)
==23516==    by 0x197CFF0: WorldObject::GetZoneId() const (Object.cpp:1277)

### DIFF
--- a/src/common/Collision/Management/MMapManager.cpp
+++ b/src/common/Collision/Management/MMapManager.cpp
@@ -179,7 +179,7 @@ namespace MMAP
         dtTileRef tileRef = 0;
 
         // memory allocated for data is now managed by detour, and will be deallocated when the tile is removed
-        if (dtStatusSucceed(mmap->navMesh->addTile(data, fileHeader.size, 0/*DT_TILE_FREE_DATA*/, 0, &tileRef)))
+        if (dtStatusSucceed(mmap->navMesh->addTile(data, fileHeader.size, DT_TILE_FREE_DATA, 0, &tileRef)))
         {
             mmap->loadedTileRefs.insert(std::pair<uint32, dtTileRef>(packedGridPos, tileRef));
             ++loadedTiles;


Carbenium:
Open for review.
/cc @streetrat 

Change which removed the flag: https://github.com/TrinityCore/TrinityCore/commit/a68f19af5638f67778ab57b9ee9883c370362f2b#diff-d637f5263f89a4f65c9bd9c5dff32969R156

jackpoz:
What about phased tiles ? Are they deleted too each time they are removed from the mesh ? Did you valgrind the changed code too ?

src/common/Collision/Management/MMapManager.cpp
```diff
@@ -540,7 +540,7 @@ namespace MMAP
             loadedPhasedTiles[swap].insert(packedXY);
 
             // add new swapped tile
-            if (dtStatusSucceed(navMesh->addTile(ptile->data, ptile->fileHeader.size, 0, 0, &loadedTileRefs[packedXY])))
+            if (dtStatusSucceed(navMesh->addTile(ptile->data, ptile->fileHeader.size, DT_TILE_FREE_DATA, 0, &loadedTileRefs[packedXY])))
```
Carbenium: @jackpoz: phased tiles are added with the flag, so memory is freed on removal

Carbenium:
I did valgrind the changes and the leaks were gone.

src/common/Collision/Management/MMapManager.cpp
jackpoz: so the phased tile is created in memory every time it's swapped and there's no reference anywhere storing the ?
because to me it looks like https://github.com/Carbenium/TrinityCore/blob/mmap-memleak/src/common/Collision/Management/MMapManager.cpp#L467 is going to delete the phased tile

PR #16430, opened 2016-01-30
Core/Scripts/DB: hardcoded text in 'npc_tirion_fordring' moved to DB

ghost:
- created new gossip_menu for NPC 1855 Tirion Fordring
- replaced hardcoded text with enum links to DB entries
- using `ADD_GOSSIP_ITEM_DB` instead of `ADD_GOSSIP_ITEM`
- replaced quest ID with enum

Tested with both alliance and horde character, works as intended (like before).

I did want to move this script to SAI, but I could not find a condition
that checks for `UNIT_STAND_STATE_SIT` (only `UNIT_STATE` (21) )


Tonghost:
:+1:

tkrokli:
Closed by #16680

PR #16449, opened 2016-02-01
[6.x] DB/Filenames: Rename incorrectly named sql files

imbavirus:
The one file was named with 2015 instead of 2016 and there was also two files marked as _04 for the same day so I renamed them so they are correctly named and in the correct order


jackpoz:
> No description provided.

Please always provide a description

jackpoz:
@Rushor why is this pr marked as Branch-3.3.5a ? Aren't those files about cherrypicks from 335 ?

Rushor:
i don't even know what the issue is about due to missing description. add it pls @imbavirus 
and well no issue on 3.x with this stuff? kay removing

Issue #16445, opened 2016-02-01
[3.3.5][6.x][Druid] Core/Spell: Feral Charge (Cat)


Treeston:
Branch 3.3.5, rev. 8a0bbc386982

Feral Charge (Cat) should:
* Not cause target to aggro the druid if the druid is stealthed
* Cause the druid to land facing the target('s back)

http://www.wowhead.com/spell=132302/wild-charge

tkrokli:
What is the current observed bug? The opposite of your listed effects? :)

Treeston:
* Using Feral Charge (Cat) while stealthed makes the targeted mob enter combat with the druid
* After using Feral Charge (Cat), the druid lands facing whatever direction the arc of the charge had

Keader:
On Spellmgr.cpp add atribute

```cpp
spellInfo->AttributesEx3 |= SPELL_ATTR3_NO_INITIAL_AGGRO;
```
does not solve the problem of combat?

joschiwald:
https://gist.github.com/joschiwald/2f5564d680fc74a55ff8

there is also a similar issue

Treeston:
Also (not sure if this a bug), but you can use Feral Charge (Cat) while rooted.

Eliminationzx:
https://github.com/Eliminationzx/TrinityCore/commit/106898cdb453a2d1be9977642ddab5a6f1a01093

Rushor:
thx for sharing :)

Keader:
close this please :)

Treeston:
Ref. ff80f24 / cfed2d7 fixes the orientation issue.

Initial aggro issue still exists.

Keader:
add in spellmgr.cpp flag >spellInfo->AttributesEx3 |= SPELL_ATTR3_NO_INITIAL_AGGRO; is hack?

Treeston:
Yes, unless there is no better way.

joschiwald:
b4b43d03b3fbff38ebcf84be27388202fc6cb08f

PR #16450, opened 2016-02-01
[3.3.5] Core/Movements: Fix jump dest orientation

Eliminationzx:
Closes https://github.com/TrinityCore/TrinityCore/issues/16445.
Also fixed cast spells "[Cat charge](http://wotlk.openwow.com/spell=49376)", "[Ghoul leap](http://wotlk.openwow.com/spell=47482)" while rooted.

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -366,10 +366,10 @@ void MotionMaster::MoveJumpTo(float angle, float speedXY, float speedZ)
     float moveTimeHalf = speedZ / Movement::gravity;
     float dist = 2 * moveTimeHalf * speedXY;
     _owner->GetClosePoint(x, y, z, _owner->GetObjectSize(), dist, angle);
-    MoveJump(x, y, z, speedXY, speedZ);
+    MoveJump(x, y, z, _owner->GetOrientation(), speedXY, speedZ);
 }
 
-void MotionMaster::MoveJump(float x, float y, float z, float speedXY, float speedZ, uint32 id)
+void MotionMaster::MoveJump(float x, float y, float z, float o, float speedXY, float speedZ, uint32 id, bool hasOrientation)
 {
     TC_LOG_DEBUG("misc", "Unit (GUID: %u) jump to point (X: %f Y: %f Z: %f)", _owner->GetGUID().GetCounter(), x, y, z);
     if (speedXY <= 0.1f)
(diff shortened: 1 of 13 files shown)


src/server/game/Entities/Object/Position.h
```diff
@@ -50,8 +50,6 @@ struct Position
     float m_positionX;
     float m_positionY;
     float m_positionZ;
-    // Better to limit access to _orientation field, to guarantee the value is normalized
-private:
```
Shauren: no.

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -382,6 +382,7 @@ void MotionMaster::MoveJump(float x, float y, float z, float speedXY, float spee
     init.MoveTo(x, y, z, false);
     init.SetParabolic(max_height, 0);
     init.SetVelocity(speedXY);
+    init.SetFacing(o);
```
Treeston: Not all jump movesplines have a facing attached. These cases should be handled.

How about passing `nanf` or similar value as orientation in those cases, and not setting facing for these?

lashia:
err

Eliminationzx:
@lashia, `MotionMaster::MoveJump` takes 5 arguments.
`me->GetMotionMaster()->MoveJump(2134.79f, -263.03f, 419.84f, me->GetOrientation(), 30.0f, 20.0f);`

Shauren:
@Eliminationzx ignore him
@lashia do not report that unmerged PRs don't compile with this...

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -382,6 +382,8 @@ void MotionMaster::MoveJump(float x, float y, float z, float speedXY, float spee
     init.MoveTo(x, y, z, false);
     init.SetParabolic(max_height, 0);
     init.SetVelocity(speedXY);
+    if (HasDst)
```
Treeston: What is your reasoning behind using a bool flag instead of [`std::nanf`](http://www.cplusplus.com/reference/cmath/nanf/) or similar non-numeric float constants?

Shauren: nans are evil, would rather see optional used here BUT given that orientation logically belongs after Z (wouldnt make sense to make it last arg) i think your solution of using bool with default value as last argument is perfect.

Eliminationzx:
Updated.

PR #16132, opened 2015-12-30
Scripts/Deadmines: Fix #6624 Mr. Smite's event text, and phase switch…

henhouse:
…ing.
- Fixes Mr. Smite announcing to instance once doors have been destroyed.
- Add missing dialogue to encounter phases.
- Now properly runs in front of his chest, kneels down and equips items, stands, and reengages like he's supposed to.
- Fix weapon model for first phase when he switched to double axes.

Closes #6624

PR #16453, opened 2016-02-02
[WIP][3.3.5] Core/Scripts: Deadmines instance / Mr. Smite

ghost:
This is a Work in Progress for what needs to be done to make
the yells from Mr. Smite work, as well as a kneeling emote
that did not work before (unsupported emote method).

I will need lots of constructive feedback to make it a working script.

Target points for this PR:
- remove 2 hardcoded lines from the instance script
- add all 3 missing say(yell) lines to creature text
- use entries from creature_text via proper enum
- make Mr. Smite use a proper kneeling emote
- apply proper code style to enum sections

With all 4 yells for Mr. Smite available in creature_text,
this PR is open for suggestions on how to implement
the 2 missing yells in extra events in the scripts.

### DIFF
--- a/src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
+++ b/src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
@@ -31,19 +31,22 @@ EndScriptData */
 
 enum Sounds
 {
-    SOUND_CANNONFIRE                                     = 1400,
-    SOUND_DESTROYDOOR                                    = 3079,
-    SOUND_MR_SMITE_ALARM1                                = 5775,
-    SOUND_MR_SMITE_ALARM2                                = 5777
+    SOUND_CANNONFIRE         = 1400,
+    SOUND_DESTROYDOOR        = 3079,
+    SOUND_MR_SMITE_ALARM1    = 5775,
+    SOUND_MR_SMITE_ALARM2    = 5777
 };
 
-#define SAY_MR_SMITE_ALARM1 "You there, check out that noise!"
-#define SAY_MR_SMITE_ALARM2 "We're under attack! A vast, ye swabs! Repel the invaders!"
-
-enum Misc
+enum MrSmiteYells
 {
-    DATA_CANNON_BLAST_TIMER                                = 3000,
-    DATA_PIRATES_DELAY_TIMER                               = 1000
+    SAY_CHECK                = 0,
+    SAY_AGGRO                = 1
+};
+
+enum Timers
+{
+    DATA_PIRATES_DELAY_TIMER = 1000,
+    DATA_CANNON_BLAST_TIMER  = 3000
 };
 
 class instance_deadmines : public InstanceMapScript
@@ -74,9 +77,22 @@ class instance_deadmines : public InstanceMapScript
             ObjectGuid DefiasCompanionGUID;
 
             uint32 State;
+            ObjectGuid MrSmiteGUID;
             uint32 CannonBlast_Timer;
             uint32 PiratesDelay_Timer;
             ObjectGuid uiSmiteChestGUID;
+
+            void OnCreatureCreate(Creature* creature) override
+            {
+                switch (creature->GetEntry())
+                {
+                    case NPC_MR_SMITE:
+                        MrSmiteGUID = creature->GetGUID();
+                        break;
+                    default:
+                        break;
+                }
+            }
 
             virtual void Update(uint32 diff) override
             {
@@ -91,9 +107,9 @@ class instance_deadmines : public InstanceMapScript
                 {
                     case CANNON_GUNPOWDER_USED:
                         CannonBlast_Timer = DATA_CANNON_BLAST_TIMER;
-                        // it's a hack - Mr. Smite should do that but his too far away
-                        //pIronCladDoor->SetName("Mr. Smite");
-                        //pIronCladDoor->MonsterYell(SAY_MR_SMITE_ALARM1, LANG_UNIVERSAL, NULL);
+
+                        if (Creature* mrsmite = instance->GetCreature(MrSmiteGUID))
+                            mrsmite->AI()->Talk(SAY_CHECK);
                         pIronCladDoor->PlayDirectSound(SOUND_MR_SMITE_ALARM1);
                         State = CANNON_BLAST_INITIATED;
                         break;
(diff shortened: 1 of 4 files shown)


src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
```diff
@@ -87,13 +95,17 @@ class instance_deadmines : public InstanceMapScript
                 if (!pIronCladDoor)
                     return;
 
+                // Loads Mr. Smite's grid. We need this as he must be active in order for his texts to work.
+                instance->LoadGrid(MrSmite.GetPositionX(), MrSmite.GetPositionY());
```
Nyeriah: This should be called on the same event his text is. The placement of this here will cause it to be executed on every update (unnecessary overload)

tkrokli: Thanks. I will move it to the current 2 positions where existing events for yells are.

src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
```diff
@@ -74,10 +77,16 @@ class instance_deadmines : public InstanceMapScript
...
 
             uint32 State;
+            uint32 MrSmiteGUID;
             uint32 CannonBlast_Timer;
             uint32 PiratesDelay_Timer;
             ObjectGuid uiSmiteChestGUID;
 
+            void OnCreatureCreate(Creature* creature) override
+            {
+                MrSmiteGUID = creature->GetGUID();
```
ccrs: entry check

src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
```diff
@@ -184,11 +203,16 @@ class instance_deadmines : public InstanceMapScript
             {
                 switch (go->GetEntry())
                 {
-                    case GO_FACTORY_DOOR:   FactoryDoorGUID = go->GetGUID(); break;
-                    case GO_IRONCLAD_DOOR:  IronCladDoorGUID = go->GetGUID();  break;
-                    case GO_DEFIAS_CANNON:  DefiasCannonGUID = go->GetGUID();  break;
-                    case GO_DOOR_LEVER:     DoorLeverGUID = go->GetGUID();     break;
-                    case GO_MR_SMITE_CHEST: uiSmiteChestGUID = go->GetGUID();  break;
+                    case GO_FACTORY_DOOR:   FactoryDoorGUID = go->GetGUID();
```
ccrs: new line also for the assignations

src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
tkrokli: do you mean check if he is present in the instance?

src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
tkrokli: Ah, right. Will fix.

src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
Riztaz: switch (creature->GetEntry()) case urNpcEntry: ...

Right now your hook looks like this: Assign guid from any creature created in instance to MrSmiteGUID

src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
```diff
@@ -74,10 +77,16 @@ class instance_deadmines : public InstanceMapScript
             ObjectGuid DefiasCompanionGUID;
 
             uint32 State;
+            uint32 MrSmiteGUID;
```
Riztaz: use ObjectGuid here

src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
```diff
@@ -91,9 +100,12 @@ class instance_deadmines : public InstanceMapScript
                 {
                     case CANNON_GUNPOWDER_USED:
                         CannonBlast_Timer = DATA_CANNON_BLAST_TIMER;
-                        // it's a hack - Mr. Smite should do that but his too far away
-                        //pIronCladDoor->SetName("Mr. Smite");
-                        //pIronCladDoor->MonsterYell(SAY_MR_SMITE_ALARM1, LANG_UNIVERSAL, NULL);
+
+                        // Loads Mr. Smite's grid. We need this as he must be active in order for his texts to work.
+                        instance->LoadGrid(MrSmite.GetPositionX(), MrSmite.GetPositionY());
```
Riztaz: incorrect

if (Creature* smite = instance->GetCreature(MrSmiteGUID))
{
    Position pos = smite->GetPosition();
    instance->LoadGrid(pos.GetPositionX(), pos.GetPositionY());
}
But i think loading grid here is not required, try changing textRange in `creature_text`
Ranges:
0 - Normal (Default)
1 - Area
2 - Zone
3 - Map
4 - World

LordUsagi:
This could be joined with @henhouse's work

tkrokli:
@LordUsagi : have you got a link to an example? I found it difficult to understand where it is listed among his repositories. Would be interesting to look at it if you can link what you refer to.

henhouse:
I made a pull request about 2 months ago.

https://github.com/TrinityCore/TrinityCore/pull/16132

henhouse:
Our code is incredibly similar, I just have a few more minor things changed like his axes, the door triggering the first emote, and exact positioning in front of his chest, etc.

tkrokli:
Thanks for the feedback, would be interesting to see if you have got suggestions for improvements. :smile:

tkrokli:
Just finished checking the PR core & DB settings. "You there! Check out that noise!" (SAY_CHECK) is said instantly upon firing the cannon at the ironclad door. 3 seconds after, the second yell is used:
"You landlubbers are tougher than I thought! I'll have to improvise!" (SAY_AGGRO)
This yell is also triggered upon aggroing Mr. Smite again at the ramp up to the ship.

Remaining "issue": When is Mr Smite supposed to kneel? Have to check it, but don't know exactly when.

LordUsagi:
Doesn't he kneel when he changes weapons from the crate?

henhouse:
See this video: https://www.youtube.com/watch?v=PY7frfcqvvw

He does not yell the first two every again -- that is only upon the door being destroyed. I have reason to believe he may not yell the second one unless the player walks into the next area (where the boat is), but it was trivial so I simply set a ~5 second timer on it.

In between phases he stomps, exists attack mode & attack animations, runs over to his chest, kneels and equips different weapons each time. He then runs back and reenters combat with players. I alternate between the two different say phrases each phase change.

In regards to your quests why the creature_text was in the header files, it's just old code. Someone hacked that stuff in instead of creating working solutions years ago, which is why I updated them.

tkrokli:
Sorry, I see now that I have copy-pasted the wrong line as the repeated line Mr Smite used while I was testing. The correct line should be: "We're under attack! Avast, ye swabs! Repel the invaders!"

Treeston:
@tkrokli and @henhouse which of the PRs do you feel should be the one to be continued?

tkrokli:
I recommend continue using @henhouse's PR since this one lacks the extra events as well as I have had trouble finding out how to make timed actions in SAI (distracted, not able to focus long enough).

PR #16457, opened 2016-02-02
[3.3.5] Add condition for unit sitting/standing

Treeston:
DB/Conditions: Add new `CONDITION_STAND_STATE` (41). value1 selects the type of state check to apply, value2 specifies the particular state we are looking for.

Valuing:
- `0,x` - has exactly state `x`
- `1,0` - has any standing state
- `1,1` - has any sitting state

### DIFF
--- a/src/server/game/Conditions/ConditionMgr.cpp
+++ b/src/server/game/Conditions/ConditionMgr.cpp
@@ -100,7 +100,8 @@ ConditionMgr::ConditionTypeInfo const ConditionMgr::StaticConditionTypeData[COND
     { "Health Value",         true, true,  false },
     { "Health Pct",           true, true, false  },
     { "Realm Achievement",    true, false, false },
-    { "In Water",            false, false, false }
+    { "In Water",            false, false, false },
+    { "Sit/stand state",      true,  true, false }
 };
 
 // Checks if object meets the condition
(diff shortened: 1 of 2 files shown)


DDuarte:
Any particular reason to re-use a condition type instead of creating a new one?
The way it was done is not backwards compatible.

Tonghost:
#16430
Now possible to move `npc_tirion_fordring` script to SAI

Treeston:
@DDuarte Avoid clutter, I guess. I suppose we could make a new condition for it as well, but it feels excessive.

Also, included sql converts any old conditions to new format.

DDuarte:
I'd understand that reasoning if the CondValue1 was an actual id used by the client (e.g something that would not require specific code (a switch) per value). It has the disadvantage of CondValue2 changing meaning based on CondValue1.

Let's wait for someone else's opinion :)

jackpoz:
can one be drunk and sitting at same time ?

tkrokli:
I thought that there is room in the existing conditions list to create a new condition, but I don't know.

Treeston:
Re-done. No longer re-uses existing condition, but adds a new condition.

tkrokli:
I have tested this PR now with the following conditions + SAI + gossip menus:
https://gist.github.com/tkrokli/717fbc26b49ce7c51845#file-redemption_tirions_tale-sql
Sadly, any variant of the condition values will not allow the gossip to be shown. I have tested with 1,1 (any sitting state), 0,1 (exact sitting state), 1,0 (any standing state) and 0,0 (exact standing state), but no change. I have even tested if there are any undocumented settings to `ConditionTarget` instead of 0, but it did not make any difference. I have tested this PR alone and also merged with the regular 3.3.5 branch, no difference. I am unable to make the following condition apply:
```sql
DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId`= 15 AND `SourceGroup`= 3502;
INSERT INTO `conditions` (`SourceTypeOrReferenceId`,`SourceGroup`,`SourceEntry`,`SourceId`,`ElseGroup`,`ConditionTypeOrReference`,`ConditionTarget`,`ConditionValue1`,`ConditionValue2`,`ConditionValue3`,`NegativeCondition`,`ErrorType`,`ErrorTextId`,`ScriptName`,`Comment`) VALUES
(15, 3502,0, 0,1, 41,0, 1, 1,0,0,0,0, '', 'Show gossip menu option 3502 if player is sitting');
```

Treeston:
Fixed & merged.

Issue #15682, opened 2015-10-09
[3.3.5] Trash not despawning on soft reset after boss is killed


Treeston:
3.3.5, Rev. 50cf74f0dbe1

After a boss is killed, any trash associated with that boss should despawn after a soft reset of the instance (nobody enters the instance for 30 minutes, raid is disbanded/reformed, or server is restarted). This does not happen.

Krudor:
You are very much correct!
The systems managing raids and dungeons are starting to get outdated and would need a rework somewhere down the line, I might take a look at this after the reverse engineers do their magic with fixing most of the opcodes now that the expansion is coming to a halt.

I should probably make a thread about this somewhere but there should be a new way of handling creature spawns, which this would be related to.

w5860363:
Do you mean with instance, the BOSS of the front of the monster should be gone?
I did not have this problem, in addition to Odul, as if it is because the header file only defines 16 BOSS

Treeston:
1. Find an instance with an optional trash pack (ex.: Naxxramas, Spiderling packs before Anub'rekhan)
2. Kill the boss those trash packs are linked to (ex.: Anub'rekhan)
3. Exit the instance and disband/reform the raid (or: Restart the world server) to force a soft instance reset

Expected behavior:
4. After re-entering the instance, the optional trash pack should be despawned.

Instead:
4. The optional trash pack is still present.

MrSmite:
For reference: http://wowwiki.wikia.com/wiki/Instance_reset

> A soft reset is when all party members leave the instance, and then come back in. This puts people out of combat, making mobs run back to their original places.

> The term "soft reset" can also mean having all party/raid members leave the instance for at least 30 minutes. This causes any trash mobs that have been killed to respawn. The most common reason to do this (deliberately) is to reset a boss fight that bugged out part way through the encounter.

@Treeston 

You're correct about leavng the instance for 30 minutes but I'm not sure a server restart should automatically repop the instance, it just reads from `creature_respawn`. I suppose this could be a CONF option to allow server restarts to immediately repop instance mobs.

Keader:
Mimiron (Ulduar) on Air Phase dont despawn your adds too xD

mvelasco93:
linked_respawn where the leader is the boss?

Jinnaix:
Cant speak for master, but works now for 3.3.5 on rev fb227be9eb53581863a45002a3878566b3c1e22c
[...]

PR #14884, opened 2015-06-13
Core/Scripts Boss Ahune

sirikfoll:
I know that spawns should only appear when queuing with LFG, but that can be changed when this system is implemented, all npc's that i added to database, needs to be there to work properly.

feedback will be appreciated ^^

Issue #1516

PR #16458, opened 2016-02-02
[3.3.5] Fix instance trash spawn behavior

Treeston:
Implement new CONDITION_SOURCE_TYPE_CREATURE, which allows the condition system to prevent spawning of creatures.

The immediate use is that this allows us to have blizzlike trash respawn behavior in instances, with trash no longer respawning after the associated boss is dead, and also despawning on soft reset if the boss is dead (issue #15682).

If a condition for checking instance group's LFD id is implemented, this system could also be used to spawn holiday bosses only when queued for them using LFD (see discussion in PR #14884).

As an example, the included sql update file adds conditions to all appropriate creatures in Naxxramas, causing them to have correct respawn behavior.


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -222,7 +222,11 @@ void Creature::AddToWorld()
 
         GetMap()->GetObjectsStore().Insert<Creature>(GetGUID(), this);
         if (m_spawnId)
+        {
             GetMap()->GetCreatureBySpawnIdStore().insert(std::make_pair(m_spawnId, this));
+            if (!sConditionMgr->IsObjectMeetingNotGroupedConditions(CONDITION_SOURCE_TYPE_CREATURE, m_spawnId, this))
+                AddObjectToRemoveList();
```
Shauren: not approving this hack

Aokromes:

Treeston:
Like I said a few times already - this won't be merged.
It'll be part of dynamic spawns once that gets merged, and @r00ty-tc and myself will find a less hacky way of implementing it.

r00ty-tc:
Pretty sure the version already in dynspawn branch stops the removal after spawning. They're simply not spawned at all if they don't meet the condition.

Treeston:
Still hacky with creating the dummy worldobject.

PR #16467, opened 2016-02-03
[3.3.5] Creature/AI: Fix incorrect evade logic

### LINKED ISSUE #16461: [3.3.5] Core/Creature: Creatures not evading properly if affected by damage over time effects
3.3.5 d5868a7db31a13ac147a66a25cb4c3a6b8a93f8a

Repro case:
* Take any character that can use damage over time effects (I used a cat druid)
* Engage a creature (I used Tuskarr in Nagrand, no idea if it works for others)
* Place a damage over time effect on the creature, then die

Expected behavior:
* The creature immediately realizes its threat list is empty and enters evade mode, removing the debuff and returning to spawn

Observed behavior:
* The creature fails to enter evade mode until a few seconds _after_ the final damage over time effect has expired. It will display an empty threat list (.debug threat), but will still be in combat (UNIT_STATE_COMBAT in .npc info) and will not regenerate health
* If all players leave the creature's grid (release spirit) while it is in this state, any debuffs present on the creature will become "stuck", never expiring and instead remaining perpetually with a 0 second duration (unless something causes the debuff to be removed)

Treeston:
Creature/AI: Move a LastDamagedTime check to CanCreatureAttack (from SelectVictim) to prevent overworld creatures from not evading while a damage-over-time spell is active.

Fixes and closes #16461.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2142,6 +2142,11 @@ bool Creature::CanCreatureAttack(Unit const* victim, bool /*force*/) const
     if (GetMap()->IsDungeon())
         return true;
 
+    // if the mob is actively being damaged, do not reset due to distance unless it's a world boss
+    if (!isWorldBoss())
+        if (time(NULL) - GetLastDamagedTime() <= MAX_AGGRO_RESET_TIME)
+            return true;
+
     //Use AttackDistance in distance check if threat radius is lower. This prevents creature bounce in and out of combat every update tick.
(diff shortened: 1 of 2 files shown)


joschiwald:
:+1:

ccrs:
If there are no problems with other uses of CanCreatureAttack, I totally agree and look forward to this merge :)

joschiwald:
@ccrs the LastDamagedTime stuff should just skip the evade-if-too-far-from-homeposition stuff

ccrs:
I mean other uses of the function like in some random script for example

PR #16469, opened 2016-02-03
[3.3.5] Implement AuctionHouse features: GetAll scan and search throttling

r00ty-tc:
Implements two standard features of the Auction House.
- GetAll scan, retrieves all auctions and sends them in a single packet.
  There's a limitation on how often a player can do this
- Search throttling. For normal searches, the server can send a time
  in milliseconds to the client, the client will wait that long between
  searches. Delay set in config

I don't think this needs to be a long discussion. The PR shouldn't break anything. In theory the large packets could pose a problem. I tested up to 22k items and it took 30secs over internet (not LAN). Other players were able to function fine during this time.

The variable delay is nothing that could cause a problem either. So, mainly it's in case I missed something, or coding mistakes (I've not properly reviewed this yet) and so on.


src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -648,105 +648,128 @@ void AuctionHouseObject::BuildListOwnerItems(WorldPacket& data, Player* player,
...
     std::wstring const& wsearchedname, uint32 listfrom, uint8 levelmin, uint8 levelmax, uint8 usable,
     uint32 inventoryType, uint32 itemClass, uint32 itemSubClass, uint32 quality,
-    uint32& count, uint32& totalcount)
+    uint32& count, uint32& totalcount, bool getall)
 {
     int loc_idx = player->GetSession()->GetSessionDbLocaleIndex();
     int locdbc_idx = player->GetSession()->GetSessionDbcLocale();
 
     time_t curTime = sWorld->GetGameTime();
+    time_t throttleTime = GetAllThrottleMap.find(player->GetGUID()) != GetAllThrottleMap.end() ? GetAllThrottleMap[player->GetGUID()] : curTime;
```
Treeston: You are doing two lookups for the same key in the map. This is inefficient. Use the iterator you get from .find instead.

src/server/game/AuctionHouse/AuctionHouseMgr.h
```diff
@@ -110,6 +110,7 @@ class AuctionHouseObject
     }
 
     typedef std::map<uint32, AuctionEntry*> AuctionEntryMap;
+    typedef std::unordered_map<ObjectGuid, time_t> PlayerGetAllThrottleMap;
```
Rochet2: I understand the ease of this, but this storage is never cleared of records that are unused/useless/expired.

Carbenium:
:+1:

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -648,105 +648,128 @@ void AuctionHouseObject::BuildListOwnerItems(WorldPacket& data, Player* player,
...
+            if (!item)
+                continue;
 
-        if (itemClass != 0xffffffff && proto->Class != itemClass)
-            continue;
+            ++count;
+            ++totalcount;
+            Aentry->BuildAuctionInfo(data);
+        }
+        GetAllThrottleMap[player->GetGUID()] = curTime + sWorld->getIntConfig(CONFIG_AUCTION_GETALL_DELAY);
```
DDuarte: Perhaps add a return after this statement so the 100 lines after this one do not need to be moved to the `else` (it makes code reviewing harder)

src/server/game/AuctionHouse/AuctionHouseMgr.h
pete318: Agreed, I considered adding something to the AH update, it's not often run and so shouldn't be a problem.

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
pete318: Yep, will take a look later.

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
pete318: And.. Also agreed. Will change later.

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
jackpoz: what if this little time_t gets moved to Player class so there's no global container, no lookup, no need to clear any expired thing.

pete318: I did consider that. But, wasn't sure how well adding to player would be, for a minor little used AH feature.

All the information to support this has been around for a while, I suspect it wasn't added because most people didn't really know about it. I doubt it would get a lot of use. Hence I thought a separate map might be a better solution.

But, maybe a straw poll to decide which way is better and I'll then make the change if required.

pete318:
Incidentally, the getAll scan seems to be very different in 6.x. So this can't be directly moved.

Shauren:
30 seconds for 22k items for one player? was this time spent in building the packet or after it had been queued into the socket?

pete318:
That wasn't an internal benchmark. It might have included processing time by Auctioneer too.

I'll benchmark each phase a bit later.

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
jackpoz: The logout thing makes sense, please add these notes in the code too so the container won't be moved to Player class in the future

Shauren:
Don't forget about 2^23 bytes packet size limit in 3.3.5

pete318:
That could be the limiting factor. Since the 45k item limit was on retail. It might be possible to maybe send multiple large packets. But, it might also be that 45k is the limit and for anything more you need to use old school page by page scan.

Sadistic server admins can of course force a 10sec wait between pages now too.

Shauren:
Btw that limit is 2^19 in 6.x

pete318:
Yeah, in 6.x it looks like GetAll uses a different packet type and sends one packet per auction. But not 100% sure on that.

WPP didn't show anything at all. But I could see them in the .PKT. Probably it had to be changed in part because of the smaller max packet.

pete318:
Commited 3aaeb574050668e5a240078f6e40337c3975d110

PR #16484, opened 2016-02-06
Grids/Notifiers: Notify triggers of units moving within their LoS

Treeston:
What the title says. Make MoveInLineOfSight be invoked for triggers again (restores really old behavior). Fixes stuff like Yogg not triggering, no flying guards in cities etc.


ccrs:
why was this removed on frist place?

PR #16491, opened 2016-02-07
[3.3.5] Add target boundary check to BossAI CanCreatureAttack call

Treeston:
AI/BossAI: Make creature never chase targets that are out of bounds
- Add missing const attribute on CheckBoundary arg.
- Prevent AI from attacking a target that is out of bounds. This prevents exploids such as killing Gluth/Deathbringer from out of bounds.


Shauren:
What about scripts currently overriding that method? Wouldn't you want to also include it there *if* they are using both boundary and CanAIAttack?

Issue #4524, opened 2011-12-29
[Core/Spells] Magic Absorption Calculation


maxxx:
http://www.wowhead.com/spell=29444#comments
Increases all resistances by 1 per level 
now it gives "level-1" resistance (79 on level 80). Yes it' small difference, but may be in other place importance will be higher.
wrong calculation here (not sure at 100%):
SpellEffectInfo::CalcValue
...
level -= int32(_spellInfo->SpellLevel);

f283a7b116d7bdcf1b35361df85b2d07927107f2

Rushor:
https://github.com/TrinityCore/TrinityCore/pull/16494

mik1893:
mmh this should be closed... fixed by #16494

Rushor:
Reference:

https://github.com/TrinityCore/TrinityCore/commit/343c2fbf526a279964ca7d1927a2e245821607c3
https://github.com/TrinityCore/TrinityCore/commit/07bb3037dec38b6e5bb8672e3dd03683d4702db1

PR #16494, opened 2016-02-07
[3.3.5]Core/Spell: Magic Absorption Calculation

Lumber-Jack-2:
- closes https://github.com/TrinityCore/TrinityCore/issues/4524
- by @maxxx
- passive spells like Magic Absorption should not be affected by this calulation

what will be fixed with this?
before: mage 80 lvl 79 resistance...
with fix: mage 80 lvl 80 resistance


jackpoz:
is there a list of affected spells ?

Rushor:
@Goatform: do you tried it already? any related "new" issues? (and no i cannot recheck this atm, i need to get fresh spell_dbc first)

jackpoz:
> is there a list of affected spells ?

jackpoz:
Merging this PR since it seems none will ever provide the list of affected spells anyway. Revert is if it breaks anything.

PR #16525, opened 2016-02-09
[6.x] Scripts/Draenor: Add skeleton for upcoming Draenor scripts

P-Kito:
Just did some basic work for upcoming Draenor scripts


joschiwald:
does not match trinity standards

many wrong stuff

PR #16528, opened 2016-02-09
Core/Scripts port _DespawnAtEvade functionality to 3.3.5 branch

sirikfoll:
backport @Shauren commit fefee58de7d8c72ba6b34d7f84ee2d626dad59ce to 3.3.5 branch
Also Implement the _DespawnAtEvade functionality to boss Lich King.

I hope to promote discussion on how it should be used.

I used the _DespawnAtEvade found in 6.x, with the addiction of a SetBossState to FAIL and a parameter to allow customize the respawn time, keeping 29s as default.

As for  the Lich kings's script:
- Moved the code from JustReachedHome() and Reset() to InitializeAI() and JustRespawned().
  Thanks @Treeston
- Change npc Highlord Tiirion from permanent spawn to a summon.  I think this is the way it should be, since he despawn along with the Lich King.
- Misc needed changes for turning Tirion into a summon

### DIFF
--- a/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
+++ b/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
@@ -547,6 +547,29 @@ void BossAI::UpdateAI(uint32 diff)
     DoMeleeAttackIfReady();
 }
 
+void BossAI::_DespawnAtEvade(uint32 delayToRespawn)
+{
+    if (delayToRespawn < 2)
+    {
+        TC_LOG_ERROR("scripts", "_DespawnAtEvade called with delay of %u seconds, defaulting to 2.", delayToRespawn);
+        delayToRespawn = 2;
+    }
+
+    uint32 corpseDelay = me->GetCorpseDelay();
+    uint32 respawnDelay = me->GetRespawnDelay();
+
+    me->SetCorpseDelay(1);
+    me->SetRespawnDelay(delayToRespawn - 1);
+
+    me->DespawnOrUnsummon();
+
+    me->SetCorpseDelay(corpseDelay);
+    me->SetRespawnDelay(respawnDelay);
+
+    if(instance)
+        instance->SetBossState(_bossId, FAIL);
+}
+
 // WorldBossAI - for non-instanced bosses
 
 WorldBossAI::WorldBossAI(Creature* creature) :
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
@@ -285,6 +285,7 @@ enum Phases
 #define PHASE_TWO_THREE  (events.IsInPhase(PHASE_TWO) ? PHASE_TWO : PHASE_THREE)
 
 Position const CenterPosition     = {503.6282f, -2124.655f, 840.8569f, 0.0f};
+Position const TirionSpawn        = {505.2118f, -2124.353f, 840.9403f, 3.141593f};
 Position const TirionIntro        = {489.2970f, -2124.840f, 840.8569f, 0.0f};
 Position const TirionCharge       = {482.9019f, -2124.479f, 840.8570f, 0.0f};
 Position const LichKingIntro[3]   =
@@ -514,13 +515,33 @@ class boss_the_lich_king : public CreatureScript
                 _vileSpiritExplosions = 0;
             }
 
-            void Reset() override
+            void InitializeAI() override
+            {
+                SetupEncounter();
+            }
+
+            void JustRespawned() override
+            {
+                SetupEncounter();
+            }
+
+            void SetupEncounter()
             {
                 _Reset();
                 me->SetReactState(REACT_PASSIVE);
                 events.SetPhase(PHASE_INTRO);
                 Initialize();
                 SetEquipmentSlots(true);
+
+                // Reset The Frozen Throne gameobjects
+                FrozenThroneResetWorker reset;
+                Trinity::GameObjectWorker<FrozenThroneResetWorker> worker(me, reset);
+                me->VisitNearbyGridObject(333.0f, worker);
+
+                // Reset any light override
+                me->GetMap()->SetZoneOverrideLight(AREA_THE_FROZEN_THRONE, 0, 5000);
+
+                me->SummonCreature(NPC_HIGHLORD_TIRION_FORDRING_LK, TirionSpawn, TEMPSUMMON_MANUAL_DESPAWN);
             }
 
             void JustDied(Unit* /*killer*/) override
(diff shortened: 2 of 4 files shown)


src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
```diff
@@ -547,6 +547,23 @@ void BossAI::UpdateAI(uint32 diff)
     DoMeleeAttackIfReady();
 }
 
+void BossAI::_DespawnAtEvade()
+{
+    uint32 corpseDelay = me->GetCorpseDelay();
+    uint32 respawnDelay = me->GetRespawnDelay();
+
+    me->SetCorpseDelay(1);
+    me->SetRespawnDelay(29);
```
Treeston: Why this instead of `me->SetRespawnTime(seconds)` after `me->DespawnOrUnsummon()`?

src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
```diff
@@ -547,6 +547,23 @@ void BossAI::UpdateAI(uint32 diff)
     DoMeleeAttackIfReady();
 }
 
+void BossAI::_DespawnAtEvade()
+{
+    uint32 corpseDelay = me->GetCorpseDelay();
+    uint32 respawnDelay = me->GetRespawnDelay();
+
+    me->SetCorpseDelay(1);
```
Treeston: Is this necessary? `me->DespawnOrUnsummon()` does not leave a corpse.

src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
```diff
@@ -514,6 +515,8 @@ class boss_the_lich_king : public CreatureScript
                 _vileSpiritExplosions = 0;
             }
 
+            void JustRespawned() override { }
```
Treeston: Won't this prevent `Reset()` from being called on respawn? Is this intended?

DDuarte:
The original commit is https://github.com/TrinityCore/TrinityCore/commit/fefee58de7d8c72ba6b34d7f84ee2d626dad59ce
(the PR description should have included this info)

src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
Shauren: yes, this counts in m_respawnTime calculation in Creature::setDeathState

Treeston: Wouldn't be necessary if you use `me->SetRespawnTime(seconds)` after `me->DespawnOrUnsummon()` though, which IMO is the proper way (instead of using a hack and temporarily changing respawn delay).

sirikfoll:
Ty @DDuarte i haven't found the commit before ^^

src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
Treeston: As per IRC discussion, this is to prevent a core crash from leaving the DB with a next-week respawn time (`->SetRespawnTime` does not save to DB).

src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
sirikfoll: That's the point, Reset is called at https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Creature/Creature.cpp#L1718. If I do not override JustRespawned(), it will be called twice after a wipe

Nawuko:
Is there a creature flag or something for the despawn? Does this every boss in an raid or are there raid boss exceptions? I could visit every old raid and check with my hunter on retail

Treeston:
On retail, most of them are, but that is not 3.3.5 blizzlike.

Despawn would need to be handled in the `EnterEvadeMode` override of the respective BossAI.

Nawuko:
the old boss (bc, wotlk) evade mechanic now behave differently than back in 3.3.5? Is 30 seconds a fixed time for all bosses?

Treeston:
Good point. @sirikfoll can we get an optional arg that specifies time to respawn, default 30?

sirikfoll:
you mean change it to something like `void _DespawnAtEvade(uint32 delay = 30);` ?
if you want i can change it, but i think 30 seconds is the time used in retail

Treeston:
Yes, but more optional arguments never hurt anyone.

sirikfoll:
Added the argument.
What do you guys think about the way it is been used in lich king?

src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
Treeston: Move logic from Reset() to JustRespawned(), maybe?

src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
```diff
@@ -547,6 +547,23 @@ void BossAI::UpdateAI(uint32 diff)
     DoMeleeAttackIfReady();
 }
 
+void BossAI::_DespawnAtEvade(uint32 delayToRespawn)
+{
+    uint32 corpseDelay = me->GetCorpseDelay();
+    uint32 respawnDelay = me->GetRespawnDelay();
+
+    me->SetCorpseDelay(1);
+    me->SetRespawnDelay(delayToRespawn);
```
Shauren: minor, but this should be delayToRespawn -1 (the 1 is used on corpse delay one line above), also probably should add some check that noone passes a value lower than 2

sirikfoll: `ASSERT(delayToRespawn >= 2);` or `me->SetRespawnDelay(delayToRespawn < 2 ? 1 : delayToRespawn - 1);` ?

Treeston: I prefer the `ASSERT`. Immediately tells the person writing the script "this doesn't work".

jackpoz: ASSERT is used for illogic cases that are never supposed to happen, passing the number 2 here  doesn't apply to that case. If you really want to notify the developer use TC_LOG_ERROR and std:max(delayToRespawn, 2) - 1

src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
```diff
@@ -547,6 +547,23 @@ void BossAI::UpdateAI(uint32 diff)
     DoMeleeAttackIfReady();
 }
 
+void BossAI::_DespawnAtEvade(uint32 delayToRespawn)
+{
+    uint32 corpseDelay = me->GetCorpseDelay();
+    uint32 respawnDelay = me->GetRespawnDelay();
+
+    me->SetCorpseDelay(1);
+    me->SetRespawnDelay(std::max(delayToRespawn, (uint32)2) - 1);
```
Treeston: Ew, C-style cast. `2u` does what you want, though a simple
```c++
if (delayToRespawn < 2)
{
    TC_LOG_ERROR("scripts", _DespawnAtEvade called with delay of %d seconds, defaulting to 2.", delayToRespawn);
    delayToRespawn = 2;
}
```
is cleaner - not as compact, of course, but we aren't playing code golf.

jackpoz: it looks very clear and it does all the jobs it's supposed to do :) please use %u for uint32 tho

Issue #16495, opened 2016-02-07
[3.3.5][Priest] Core/Spell: Devouring Plague - Damage Negative 


Keader:
if you have insufficient spell penetration versus a class with elemental resistance, the devouring plague starts giving negative damage when hp of target is low.

Steps for reproduce:

1. Put Mage and Priest em Duel
2. Add some spell penetration in Priest ( .add aura 64130 )
3. Add some Resistence on Mage (On test i used .add aura  48169 [Shadow Resist] and Mage Armor)
4. Let the mage HP around 100.
5 . See damage "heal" mage :octopus: 

Vid1: https://youtu.be/usZTefH8eX4

Vid2: https://www.youtube.com/watch?v=2je0i-4JCb4&feature=youtu.be

Mage Vision 
![wowscrnshot_020716_121741](https://cloud.githubusercontent.com/assets/7120913/12873280/ac0f28f2-cd97-11e5-885a-7272bc7febfc.jpg)

Trinity Hash: cb8f64cb63c9262717c7a019a581bd284579583c

Code related: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/Auras/SpellAuras.cpp#L1298-L1313

Rushor:
related https://github.com/TrinityCore/TrinityCore/commit/af62cf5939948ee0fc6cc28a150f43076d095491 ?

Keader:
No, the server that I play, verison is 3-month ago and the problem occurs also

Eliminationzx:
Try it
```diff
src/server/game/Spells/Auras/SpellAuras.cpp | 5 +++--
 1 file changed, 3 insertions(+), 2 deletions(-)

diff --git a/src/server/game/Spells/Auras/SpellAuras.cpp b/src/server/game/Spells/Auras/SpellAuras.cpp
index 1ca5df6..d6a7a2d 100644
--- a/src/server/game/Spells/Auras/SpellAuras.cpp
+++ b/src/server/game/Spells/Auras/SpellAuras.cpp
@@ -1301,9 +1301,10 @@ void Aura::HandleAuraSpecificMods(AuraApplication const* aurApp, Unit* caster, b
                     // Improved Devouring Plague
                     if (AuraEffect const* aurEff = caster->GetDummyAuraEffect(SPELLFAMILY_PRIEST, 3790, 1))
                     {
[...]

PR #16534, opened 2016-02-10
[3.3.5] Core/Spells: Fix wrong damage calculation for SPELL_AURA_PERIODIC_LEECH

Eliminationzx:
Closes https://github.com/TrinityCore/TrinityCore/issues/16495
Spells with aura attribute `SPELL_AURA_PERIODIC_LEECH` has negative damage calculation.
(For example spell [Devouring Plague](http://wotlk.openwow.com/spell=2944))

Damage should be calculated by `GetMaxHealth()` instead of `GetHealth()` in that case.

> ...GetHealth function will calculate currently amount of HP on target e.g. if target has 500/1000 hp it will calculate amount of 500 and if it's GetMaxHealth function, it will calculate maximal hp of target (1000 HP in this example)...

Thanks to @Goatform for detailed description.

### DIFF
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -5998,8 +5998,8 @@ void AuraEffect::HandlePeriodicHealthLeechAuraTick(Unit* target, Unit* caster) c
 
     caster->CalcAbsorbResist(target, GetSpellInfo()->GetSchoolMask(), DOT, damage, &absorb, &resist, m_spellInfo);
 
-    if (target->GetHealth() < damage)
-        damage = uint32(target->GetHealth());
+    if (target->GetMaxHealth() < damage)
+        damage = target->GetMaxHealth();
 
     TC_LOG_DEBUG("spells.periodic", "PeriodicTick: %s health leech of %s for %u dmg inflicted by %u abs is %u",
         GetCasterGUID().ToString().c_str(), target->GetGUID().ToString().c_str(), damage, GetId(), absorb);


src/server/game/Spells/Auras/SpellAuras.cpp
```diff
@@ -1301,8 +1301,7 @@ void Aura::HandleAuraSpecificMods(AuraApplication const* aurApp, Unit* caster, b
                     // Improved Devouring Plague
                     if (AuraEffect const* aurEff = caster->GetDummyAuraEffect(SPELLFAMILY_PRIEST, 3790, 1))
                     {
-                        uint32 damage = caster->SpellDamageBonusDone(target, GetSpellInfo(), GetEffect(0)->GetAmount(), DOT);
-                        damage *= caster->SpellDamagePctDone(target, GetSpellInfo(), SPELL_DIRECT_DAMAGE);
+                        uint32 damage = caster->SpellDamageBonusDone(target, GetSpellInfo(), GetEffect(0)->GetAmount(), DOT) * caster->SpellDamagePctDone(target, GetSpellInfo(), SPELL_DIRECT_DAMAGE);
```
Shauren: There is no change here - only code formatting changes should not be submitted in PRs that fix other things

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -5998,8 +5998,8 @@ void AuraEffect::HandlePeriodicHealthLeechAuraTick(Unit* target, Unit* caster) c
 
     caster->CalcAbsorbResist(target, GetSpellInfo()->GetSchoolMask(), DOT, damage, &absorb, &resist, m_spellInfo);
 
-    if (target->GetHealth() < damage)
-        damage = uint32(target->GetHealth());
+    if (target->GetMaxHealth() < damage)
+        damage = target->GetMaxHealth();
```
Shauren: What is the result of this change? Calculating heal from all damage including overkill?

jackpoz:
> Closes #16495

This does not really look like a proper PR description, you should describe in detail what behavior your code changes, why it does that and how it does that

src/server/game/Spells/Auras/SpellAuraEffects.cpp
Goatform: Well, GetHealth function will calculate currently amount of HP on target e.g. if target has 500/1000 hp it will calculate amount of 500 and if it's GetMaxHealth function, it will calculate maximal hp of target (1000 HP in this example).

src/server/game/Spells/Auras/SpellAuras.cpp
Goatform: Yes I agree with @Shauren. This part should be reverted.

src/server/game/Spells/Auras/SpellAuraEffects.cpp
Shauren: Yes, but why is this needed here?

Eliminationzx:
Thank you for the feedback.

src/server/game/Spells/Auras/SpellAuraEffects.cpp
Goatform: I think that @Eliminationzx will need to answer on that question. From my aspect of looking as dev, this isn't very smart way because this could affect other things as well, at least in my opinion.

Goatform:
@jackpoz, I think that this could be marked for Branch-6.x.  as well.

Goatform:
@jackpoz, I think that this could be marked for Branch-6.x.  as well.

Keader:
This PR Works for me :+1:

Goatform:
@Keader, I already explained that.

Shauren:
Rejecting due to insufficient description on how this fixes the issue or what the actual issue is.
Even if it has any visible effect it only masks the real issue (if one exist, I could not reproduce it given steps in #16495)

Issue #14446, opened 2015-03-25
[3.3.5] Seal of of Righteousness - Damagecalculation


Expecto:
1 - Create Paladin
2 - Use Seal of Righteousness
3 - In Seal of Righteousness indicate correct damage 21 per tick in reality damage 20 per tick...
4 - Look Screenshot: 

Patch in Kernel: Russian localization for locales_creature

http://www.imageup.ru/img293/2084250/wowscrnshot_032515_203634.jpg

Rev Hash: 9a2e9ecccc5ed4d300025f3b6d1cffc6af11265d
TDB Last Rev for 3.3.5a

MrSmite:
Shouldn't it be effected by resistances? Even low level mobs have a chance to partially resist magic.

Nyeriah:
Since when is there resistance to holy damage?

MrSmite:
There is no specific holy resistance but I've seen my Paladin's seals and judgements resisted. I never thought much of it since it doesn't happen often.

Nyeriah:
Yeah, and damage can't really be resisted ("mitigated") unless the creature has specifically resistance to the damage school. I would also assume it was something among these lines if it was a full resist but in this case it's not.

When did you get these judgment resists though? There was an ancient bug that caused druid's marks to add holy resistance to players. And also some oddity with some old bosses such a Maexna having holy resistance set, but I thought they were long gone

Nyeriah:
Anyway

```
Shanalla's Seal of Righteousness hits Defias Thug for 1 Holy.
```

At level 1 when the tooltip says it should tick for 2

MrSmite:
I don't believe NPCs are supposed to have resistances to specific schools, they simply roll a chance to resist on cast / each tick. The "no resistance for Holy" is for players so they won't resist a heal. If you do a google search for resistance and read the various wiki entries, they're inconsistent as to wether NPCs actually have a resistance stat.

The Judgement resist was a while ago but I just figured it was supposed to be a "miss" and it was being picked up wrong by the client. There are a few things the client misinterprets from the core when it comes to Lua functions.

Anyway, in this case it sounds like the core is possibly not adding spell power or some other stat in *DealDamage()*

ariel-:
I think the issue is a low-level one, but simple to correct, you see, when the spell basepoints are calculated, the core uses a naive float to int cast.

In SpellEffectInfo::CalcValue

```c++
    // base amount modification based on spell lvl vs caster lvl
[...]

PR #16535, opened 2016-02-10
[3.3.5] Core/Spells: Fix incorrect Seal of Righteousness calculation.

Eliminationzx:
Source http://wowwiki.wikia.com/wiki/Spell_power_coefficient
Closes https://github.com/TrinityCore/TrinityCore/issues/14446

### DIFF
--- a/src/server/scripts/Spells/spell_paladin.cpp
+++ b/src/server/scripts/Spells/spell_paladin.cpp
@@ -1358,7 +1358,7 @@ class spell_pal_seal_of_righteousness : public SpellScriptLoader
                 float ap = GetTarget()->GetTotalAttackPowerValue(BASE_ATTACK);
                 int32 holy = GetTarget()->SpellBaseDamageBonusDone(SPELL_SCHOOL_MASK_HOLY);
                 holy += eventInfo.GetProcTarget()->SpellBaseDamageBonusTaken(SPELL_SCHOOL_MASK_HOLY);
-                int32 bp = int32((ap * 0.022f + 0.044f * holy) * GetTarget()->GetAttackTime(BASE_ATTACK) / 1000);
+                int32 bp = int32((ap * 0.0225f + 0.0355f * holy) * GetTarget()->GetAttackTime(BASE_ATTACK) / 1000);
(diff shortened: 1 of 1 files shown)


Treeston:
[WoWwiki history agrees](http://wowwiki.wikia.com/wiki/Spell_power_coefficient?oldid=2193140).

c9240709-b07d-4439-9212-94f05f7a3212:
:+1:

ariel-:
No, 35.5% of SP and 22.5% of AP is the coefficient for the JUDGEMENT part of this spell, not for the "extra holy attack" part. This PR mixes coefficients.

The "by 1" off issue (#14446) is caused by int32/uint32 casting of floats, wich truncates the decimal part, std::lroundf should be used instead, as explained in https://github.com/TrinityCore/TrinityCore/issues/14446#issuecomment-91328897 (but maybe only in the calculation part of this spell, not in the SpellEffectInfo, which maybe will break something else)

c9240709-b07d-4439-9212-94f05f7a3212:
@ariel- You might be wrong:

Here is the spell info http://wotlk.openwow.com/spell=21084.
_Fills the Paladin with holy spirit for 30 min, granting each melee attack **[MWS*(0.022*AP+0.044*SPH)]** additional Holy damage. Only one Seal can be active on the Paladin at any one time. Unleashing this Seal's energy will cause **[1+0.2*AP+0.32*SPH]** Holy damage to an enemy._

The first part is before patch 3.3.0 (mistake on openwow) as it might be seen here on patch 3.2.0: http://www.wowhead.com/forums&topic=113856/sor-vs-socom-some-math. It was lately changed to the spell coefficient stated on WoWWiki ([Seal of Righteousness] (judgement) 35.5% 22.5% Patch 3.3), but that _(judgement)_ might be a mistake from WoWWiki as this spell was upgraded during years as now deals also damage on a area of 8 yards.

We should use _lroundf_ (as stated in your sugestion https://github.com/TrinityCore/TrinityCore/issues/14446#issuecomment-91328897) also on this formula (**which this PR updated it to patch 3.3.0**) but also on the tooltip (if it can be modified).

PR #16554, opened 2016-02-11
[3.3.5] Beast Mastery hunter talent Animal Handler

### LINKED ISSUE #15361: [3.3.5][Hunter] Animal Handler
http://wotlk.openwow.com/spell=34454

Not increase the power of pet attack 10%

4c420344fc4c8508b2142a5031d9bf8b16c0e7fd
3.3.5 branch

ghost:
The goal of this PR is to restore the correct Animal Handler talent AP values for a BM hunter.

**Animal Handler, Rank 1**
- increases your pet's attack power by 5% and
- increases the duration of your Master's Call effect by 3 sec.

**Animal Handler, Rank 2**
- increases your pet's attack power by 10% and
- increases the duration of your Master's Call effect by 6 sec.

Viewed in talent calculator: http://wotlk.openwow.com/talent#cVbhzTGRb

by **Rushor**, closes  #15361

### DIFF
--- a/src/server/game/Entities/Unit/StatSystem.cpp
+++ b/src/server/game/Entities/Unit/StatSystem.cpp
@@ -1327,6 +1327,11 @@ void Guardian::UpdateAttackPowerAndDamage(bool ranged)
             }
 
             bonusAP = owner->GetTotalAttackPowerValue(RANGED_ATTACK) * 0.22f * mod;
+            if (AuraEffect* aurEff = owner->GetAuraEffectOfRankedSpell(34453, EFFECT_1, owner->GetGUID())) // Animal Handler
+            {
+                AddPct(bonusAP, aurEff->GetAmount());
+                AddPct(val, aurEff->GetAmount());
+            }
             SetBonusDamage(int32(owner->GetTotalAttackPowerValue(RANGED_ATTACK) * 0.1287f * mod));
         }
         else if (IsPetGhoul() || IsRisenAlly()) //ghouls benefit from deathknight's attack power (may be summon pet or not)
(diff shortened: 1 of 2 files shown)


Ascathor:
Wouldn't a spellscript make more sense ? Just asking to be sure.

Aokromes:

tkrokli:
Label detail: how is [Comp-C++Script] related to `src/server/game/Entities/Unit/StatSystem.cpp` ?

Aokromes:
If there is no reason to merge this, i will merge it on 24h.

src/server/game/Entities/Unit/StatSystem.cpp
```diff
@@ -1327,6 +1327,11 @@ void Guardian::UpdateAttackPowerAndDamage(bool ranged)
             }
 
             bonusAP = owner->GetTotalAttackPowerValue(RANGED_ATTACK) * 0.22f * mod;
+            if (AuraEffect* aurEff = owner->GetAuraEffectOfRankedSpell(34453, EFFECT_1, owner->GetGUID())) // Animal Handler
+            {
+                AddPct(bonusAP, aurEff->GetAmount());
+                AddPct(val, aurEff->GetAmount());
```
joschiwald: why is the bonus applied twice?

tkrokli: Good question, need feedback from @Rushor on this one.

ariel-: val is base AP
bonusAP is, well, bonus AP

Total AP = (val + bonusAP), so it's not applying twice, it's distributing multiplication

tkrokli: Thanks for the explanation. Looking at that code again, I wonder if the sequence of those 2 lines should be swapped, but maybe the order is indifferent like when performing a regular addition?

PR #16585, opened 2016-02-12
Core/Spells: Disallow druid flight form while shapeshifted

### LINKED ISSUE #16354: [3.3.5] Core/Mechanic: Druid flight form not abiding mount restrictions re: shapeshifts
3.3.5 0316dff24530da83ef3466500ee8bd4e1af548ba

Druids are capable of shifting into flight form while shapeshifted into (non-druid) forms that do not allow mounting. Example: [Drakuru chain ghoul form](http://www.wowhead.com/item=38699/ensorcelled-choker). The result is swimming zombies.

Treeston:
Core/Spells: Disallow druid flight form while in shapeshift forms that prevent mounting. No more swimming zombies. Closes #16354.


Shauren:
Doesn't break using flight form while shapeshifted into another druid form?

Treeston:
No. The client cancels the other druid form before casting flight form.

(Other druid forms don't allow flight form unless `autoUnshift` is set on the client, even without this PR.)

PR #16615, opened 2016-02-14
Fix build failing since https://github.com/TrinityCore/TrinityCore/commit/52758c1a0b8e37e198d744803460e34600d115b5

Rochet2:
Fix build failing since https://github.com/TrinityCore/TrinityCore/commit/52758c1a0b8e37e198d744803460e34600d115b5

```
/TrinityCore/src/server/game/Maps/Map.cpp:2448:22: fatal error: 
      variable 'areaId' is used uninitialized whenever 'if' condition is false
      [-Wsometimes-uninitialized]
        if (GridMap* gmap = const_cast<Map*>(this)->GetGrid(x, y))
```

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -2439,7 +2439,7 @@ uint32 Map::GetAreaId(float x, float y, float z, bool *isOutdoors) const
             atEntry = sAreaTableStore.LookupEntry(wmoEntry->AreaTableID);
     }
 
-    uint32 areaId;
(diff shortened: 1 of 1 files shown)


tkrokli:
:+1: Good point, I was wondering why non-related files caused my latest PR to fail. :shipit:

Keader:
Same? ff45a422c5e525c4d9465e91200803ffa95700d2

tkrokli:
Yes, but ff45a42 is on 3.3.5 and we need the same warning fix here on 6.x too.

PR #16633, opened 2016-02-16
Core/Packets: Adjust 'TalentsInspecting' option

Treeston:
Core/Packets: Adjust 'TalentsInspecting' option so it no longer doesn't actually do what it says (but would instead disable talent inspecting completely).


tkrokli:
Looks nice, thanks for the contribution! :+1:

DDuarte:
I think 0 and 1 should be swapped to maintain backwards compatibility. Someone that updates the core but doesn't update the config (quite common) will suddenly have inspection enabled on all characters.

Treeston:
Which is the same behavior that setting `1` had before this change.
The only value that changes behavior is `0` (nobody can inspect -> only own faction can inspect).

DDuarte:
In that case why isn't the default value the same?

Treeston:
Because the new default (`0`) is blizzlike behavior.

Warpten:
Easier to manage if you set 0,1,2 as config values then do -1

Treeston:
Yes, but not doable while maintaining both sensible valuing and backwards compatibility.

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1203,7 +1203,8 @@ void WorldSession::HandleInspectOpcode(WorldPacket& recvData)
     WorldPacket data(SMSG_INSPECT_TALENT, guid_size+4+talent_points);
     data << player->GetPackGUID();
 
-    if (sWorld->getBoolConfig(CONFIG_TALENTS_INSPECTING) || _player->IsGameMaster())
+    // config value can be 0, 1 or INT_MAX - +1 guarantees that we end up with 0-3 (2-3 valid)
+    if (GetPlayer()->CanBeGameMaster() || sWorld->getIntConfig(CONFIG_TALENTS_INSPECTING) + (GetPlayer()->GetTeamId() == player->GetTeamId())+1 > 1)
```
jackpoz: do we really need arithmetic operations on boolean type ?

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1203,7 +1203,8 @@ void WorldSession::HandleInspectOpcode(WorldPacket& recvData)
     WorldPacket data(SMSG_INSPECT_TALENT, guid_size+4+talent_points);
     data << player->GetPackGUID();
 
-    if (sWorld->getBoolConfig(CONFIG_TALENTS_INSPECTING) || _player->IsGameMaster())
+    // config value can be 0, 1 or INT_MAX - +1 guarantees that we end up with 0-3 (2-3 valid)
```
jackpoz: is INT_MAX actually a valid config value ?

Treeston: `getIntConfig` returns `uint32`, `-1` -> `INT_MAX`, unsigned overflow is well-defined.

Shauren: -1 is not INT_MAX. (INT_MAX is 0x7FFFFFFF, or 2147483647; INT_MIN is 0x80000000, or -2147483648)

Treeston: Whatever the max value of uint32 is then (2*INT_MAX +-1 or something). Doesn't really make a difference in functionality, I can change the comment.

Shauren: then change the comment?

Issue #15379, opened 2015-08-28
[3.3.5] MindcontrolAI Controlled players do not use spells


Keader:
In boss like Lady Deathwhisper, Blood-Queen Lana´thel and all other boss who control players during the fight, should use the spells of players while they are controlled.
On TC controlled players give only melee damage and do not use spells.

Rev 5c97594a1691b78ab881437985dc5eeebb31a979

If there was another issue about it, excuse me. I could not find

P-Kito:
Mindcontrol AI - Yep.

Rushor:
or better CharmAI :D 

some basic stuff
https://github.com/Lillecarl/HGCore/commit/a421c5ba72e9bd57356bbd65564954024c811a72
https://github.com/Lillecarl/HGCore/blob/118e2d9f1c4b684b891de6e5674e419909334c5b/src/game/PlayerAI.cpp#L307
https://github.com/Lillecarl/HGCore/blame/118e2d9f1c4b684b891de6e5674e419909334c5b/src/game/PlayerAI.h
https://www.google.de/#q=void+MageAI::UpdateAI%28uint32+diff%29

P-Kito:
There is a CharmAI currently in the code for creatures. Someone just needs to create some functions for the player case :D

Rushor:
go for it @P-Kito :>

ccrs:
not that easy, player has no AI when its mind controlled.

P-Kito:
Because there is that in the code: (I saw it in SimpleCharmAI I think.. not sure though)
if (IsPlayer())
   return;

but if you would initialize an AI instead of a return, that would probably work.

ccrs:
that code is not used, its just a placeholder. you need a spellscript or a core mechanic to set the AI to the player when the charm aura is applied.

P-Kito:
Well I guess spellscript is not good, because its not generic. 

Not sure whether you should use cooldowns when mindcontrolled outside of an instance/raid.

Treeston:
Closed by 25c5570.

Runico:
Player under mind control aura character size should increase ?

Keader:
Idk if is general
In lady deathwhisper and kelthuzad should.

Treeston:
That's an attribute of the specific MC aura. #16634 fixes it for KT.

Runico:
@Treeston, I hear mind controlled players should start casting after a few seconds not instant, is that true ?

PR #16634, opened 2016-02-17
[3.3.5] Naxxramas: Kel'thuzad rework

Treeston:
Full rewrite of the last boss missing in this instance before we can call it quits.

Current WIP status:
- [x] **Phase 1** - fully working
- [x] **Phase 2** - fully working
- [x] **Phase 3** - fully working


Rushor:
@Treeston : see https://github.com/TrinityCore/TrinityCore/issues/15379 for:
> Figure out a way to make mind controlled players act smarter without resorting to hacks like the old script did (custom charm AI maybe?)

Keader:
>/home/travis/build/TrinityCore/TrinityCore/src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp:200:58: fatal error: 
      unused parameter 'damage' [-Wunused-parameter]
            void DamageTaken(Unit* /*attacker*/, uint32& damage) override

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
```diff
@@ -268,355 +144,218 @@ class boss_kelthuzad : public CreatureScript
...
-
-            instance->SetData(DATA_ABOMINATION_KILLED, 0);
-
-            if (GameObject* trigger = ObjectAccessor::GetGameObject(*me, instance->GetGuidData(DATA_KELTHUZAD_TRIGGER)))
+            void InitializeAI() override { Reset(); }
+            void JustRespawned() override
             {
-                trigger->ResetDoorOrButton();
-                trigger->SetPhaseMask(1, true);
+                instance->SetBossState(BOSS_KELTHUZAD, NOT_STARTED);
```
sirikfoll: you don't need InitializeAI()/JustRespawned() calling for Reset(), it will be called without them, and the boss state is already set in _Reset();

The need for a  void JustRespawned() override { } would be only to avoid Reset() being called twice(which wouldn't matter in this case)

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
```diff
@@ -268,355 +144,218 @@ class boss_kelthuzad : public CreatureScript
...
+                            if (Is25ManRaid())
+                                events.ScheduleEvent(EVENT_CHAINS, urandms(60, 80), 0, PHASE_TWO); // @todo sniff timer
                             break;
-                        case EVENT_NOVA:
-                            DoCastAOE(SPELL_FROST_BOLT_AOE);
-                            events.Repeat(15000, 30000);
+                        
+                        case EVENT_FROSTBOLT_VOLLEY:
+                            DoCastAOE(SPELL_FROSTBOLT_VOLLEY);
+                            if (!urand(0, 3)) // sometimes he skips one volley - is this truly random? i couldn't discern any pattern
```
sirikfoll: I assume this may be caused by the spell being locked by category(some used here uses category 1152), i never understood how this is handled by blizz

Treeston: Yes, but this is a solo priest sniff, which isn't a class that can interrupt...

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
sirikfoll: did you check my comment above?

joschiwald:
SPELL_DARK_BLAST = 28457,
is handled wrong, take a look at spell 28458

Treeston:
Fixed, I'll push the updated version once I have a first draft of charm AI finished.

Treeston:
Still need to figure out which of the triggered cast flags I'm using makes the spells instant cast. I don't want the spells to be instant cast.

sirikfoll:
isn't there a simple way to make the SimpleCharmedPlayerAI more generic? Like turning kel'thuzad into a parameter and get the guardians by looking for nearby allies.
Looks really good, btw :+1:

Treeston:
Yeah, I've been thinking about it, but current AI is too mean for generic. Having it be mean is fine when getting MCed by a boss is a failure condition, but having it do this for generic mobs is a bit questionable.

sql/updates/world/9999_99_99_99_KELTHUZAD.sql
```diff
@@ -0,0 +1,180 @@
...
+UPDATE `gameobject` SET `position_x`=3732.656, `position_y`=-5026.173, `position_z`=152.7197, `rotation0`=0, `rotation1`=0, `rotation2`=-0.7743921, `rotation3`=0.632706 WHERE `guid`=150155; -- portal 01
+UPDATE `gameobject` SET `position_x`=3784.165, `position_y`=-5062.077, `position_z`=152.5704, `rotation0`=0, `rotation1`=0, `rotation2`=-0.957571, `rotation3`=0.2881973 WHERE `guid`=150156; -- portal 02
+UPDATE `gameobject` SET `position_x`=3760.238, `position_y`=-5175.256, `position_z`=152.5706, `rotation0`=0, `rotation1`=0, `rotation2`=0.8698883, `rotation3`=0.4932488 WHERE `guid`=150157; -- portal 03
+UPDATE `gameobject` SET `position_x`=3698.601, `position_y`=-5187.073, `position_z`=152.7199, `rotation0`=0, `rotation1`=0, `rotation2`=0.6149149, `rotation3`=0.7885935 WHERE `guid`=150158; -- portal 04
+
+-- ================================== --
+-- Chains spell script (scale factor) --
+-- ================================== --
+DELETE FROM `spell_script_names` WHERE `spell_id`=28410;
```
Keader: TC always use script name for for deletes.

Treeston:
Finished.

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
```diff
@@ -45,578 +37,523 @@ enum Texts
...
             {
-                if (GameObject* portal = ObjectAccessor::GetGameObject(*me, instance->GetGuidData(DATA_KELTHUZAD_PORTAL01 + i)))
-                    if (!((portal->getLootState() == GO_READY) || (portal->getLootState() == GO_NOT_READY)))
-                        portal->ResetDoorOrButton();
+                for (ObjectGuid summon : summons)
+                    if (Creature* cSummon = ObjectAccessor::GetCreature(*me, summon))
+                        if (cSummon->IsAlive() && cSummon->GetEntry() == NPC_GUARDIAN)
+                        {
+                            cSummon->AI()->DoAction(ACTION_KELTHUZAD_DIED);
+                            summons.Despawn(cSummon); // prevent them from being despawned by _JustDied
```
ccrs: crash here, erase element so itr is invalidated

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
```diff
@@ -45,578 +37,523 @@ enum Texts
...
+
+                        case EVENT_TRANSITION_SUMMON:
+                        {
+                            uint8 selected = urand(_guardianCount, nGuardianSpawns - 1);
+                            if (selected != _guardianCount)
+                                std::swap(_guardianGroups[selected], _guardianGroups[_guardianCount]);
+
+                            std::list<TempSummon*> summons;
+                            // server-side spell 28454 is used on retail - no point replicating this in spell_dbc
+                            me->SummonCreatureGroup(_guardianGroups[_guardianCount++], &summons);
```
ccrs: same name as summons on BossAI

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
```diff
@@ -45,578 +37,523 @@ enum Texts
...
+            void DoAction(int32 action) override
+            {
+                switch (action)
+                {
+                    case ACTION_BEGIN_ENCOUNTER:
+                        if (instance->GetBossState(BOSS_KELTHUZAD) != NOT_STARTED)
+                            return;
+                        instance->SetBossState(BOSS_KELTHUZAD, IN_PROGRESS);
+                        events.SetPhase(PHASE_ONE);
+                        DoZoneInCombat();
```
ccrs: you have to erase UNIT_FLAG_IMMUNE_TO_PC, the encounter doesnt start

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
```diff
@@ -45,578 +37,523 @@ enum Texts
...
+                        DoCastAOE(SPELL_VISUAL_CHANNEL);
+                        Talk(SAY_SUMMON_MINIONS);
+
+                        for (uint8 group = SUMMON_GROUP_MINION_FIRST; group < SUMMON_GROUP_MINION_FIRST + nMinionGroups; ++group)
+                        {
+                            std::list<TempSummon*> summons;
```
ccrs: same name as summons on BossAI

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
Treeston: No. `SummonList::Despawn` is just confusingly named, all it really does is delete the element from map.

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
```diff
@@ -45,578 +37,523 @@ enum Texts
...
+{
+    double angle = rand_norm() * M_2_PI;
+    double relDistance = rand_norm() + rand_norm();
+    if (relDistance > 1)
+        relDistance = 1 - relDistance;
+    return Position(center.GetPositionX() + std::sin(angle)*relDistance*radius, center.GetPositionY() + std::cos(angle)*relDistance*radius, center.GetPositionZ());
+}
 
-class boss_kelthuzad : public CreatureScript
+struct KelThuzadCharmedPlayerAI : public SimpleCharmedPlayerAI
```
ccrs: struct inherits a class? why? xD

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
Shauren: indeed, use `summons.DespawnIf(predicate)`

ccrs: he doesnt want to make them despawn, I guess he will have to change it into ````c++ itr = summons.erase(asd); ````

src/server/scripts/Northrend/Naxxramas/boss_kelthuzad.cpp
Treeston: doesn't really make a difference anyway vOv

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -2235,7 +2235,7 @@ void GameObject::BuildValuesUpdate(uint8 updateType, ByteBuffer* data, Player* t
 
             if (index == GAMEOBJECT_DYNAMIC)
             {
-                uint16 dynFlags = 0;
+                uint16 dynFlags = m_uint32Values[GAMEOBJECT_DYNAMIC];
```
ccrs: so I guess u have to be a collab to make this type of changes so the rest wont go all retard about mixed changes #TClogic

tkrokli: Nice tweet ;-)

Shauren: Explain this change please, it does not look like something that should be in a script PR.

Treeston: For some reason, the core completely ignores any `GAMEOBJECT_DYNAMIC` flags set on GOs and _only_ ever sends the ones specifically ORed in inside that method - this is because the dynflags to be sent value is zero initialized instead of using the flag value.

I can move it to a separate PR if you think it necessary.

Shauren: And that ignoring was intended to prevent scripts from messing with the values like you do

Treeston: Then what is the intended way of setting `LO_ACTIVATE`?

Treeston:
Requesting final code review for this. Merge tomorrow if no objections.

PR #16638, opened 2016-02-18
Core/Creature: Fix a random equipment selection for summoned creatures.

### LINKED ISSUE #16587: [3.3.5]Core: Creature equipement
After this commit https://github.com/Tonghost/TrinityCore/commit/cc3ae640a48417aae31eac8828842ab50099069b all summoned npcs doesn't load equipment, example Jaraxus.
The correct behavior must:
- If the npc is a spawn, the equipment is set or not according to equipment_id on creature table (this works correctly)
- If the npc is summoned (not a spawn) equipment must always be set, And if the npc must spawn without equipment it must be handled by the script.
rev. ebf6c7188369

r4d1sh:
Closes #16587

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -369,13 +369,19 @@ bool Creature::InitEntry(uint32 entry, CreatureData const* data /*= nullptr*/)
     SetByteValue(UNIT_FIELD_BYTES_0, 2, minfo->gender);
 
     // Load creature equipment
-    if (data && data->equipmentId != 0)
+    if (data)
     {
-        m_originalEquipmentId = data->equipmentId;
-        LoadEquipment(data->equipmentId);
+        if (data->equipmentId != 0)
+        {
+            // If creature exist in `creature` table we must use `creature.equipment_id` (can be -1, it means random equipment)
+            m_originalEquipmentId = data->equipmentId;
+            LoadEquipment(data->equipmentId);
+        }
+        else
+            LoadEquipment(0, true); // Line in `creature` table exist, but `creature.equipment_id` == 0. Creature must spawn without equipment.
     }
     else
-        LoadEquipment(0, true);
+        LoadEquipment(); // Creature is summoned. Use random equipment (if available)
 
     SetName(normalInfo->Name);                              // at normal entry always
 
@@ -1399,7 +1405,7 @@ void Creature::SetCanDualWield(bool value)
     UpdateDamagePhysical(OFF_ATTACK);
 }
 
-void Creature::LoadEquipment(int8 id, bool force /*= true*/)
+void Creature::LoadEquipment(int8 id /*= -1*/, bool force /*= true*/)
 {
     if (id == 0)
     {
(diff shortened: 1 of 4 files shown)


Tonghost:
Maybe better load random equipment for summoned creatures. I will be glad to your opinions.

Killyana:
Yes, better to load a random equipment if there's more than one.

Tonghost:
Ready to test and merge.

Killyana:
For koltira, this must fix the error:
``` sql
DELETE FROM `creature_equip_template` WHERE `CreatureID` IN (28912);
INSERT INTO `creature_equip_template` (`CreatureID`, `ID`, `ItemID1`, `ItemID2`, `ItemID3`, `VerifiedBuild`) VALUES
(28912, 1, 38633, 0, 0, 0);

**Comment by Tonghost** [comment]
Good. Your sql can be commited regardless of this PR.

**Comment by ariel-** [comment]
Tested and it works :+1:

**Comment by Tonghost** [comment]
:shipit:

File: src/server/game/Entities/Creature/Creature.cpp [hunk 1]
```diff
@@ -369,13 +369,19 @@ bool Creature::InitEntry(uint32 entry, CreatureData const* data /*= nullptr*/)
...
-        m_originalEquipmentId = data->equipmentId;
-        LoadEquipment(data->equipmentId);
+        if (data->equipmentId != 0)
+        {
+            // If creature exist in `creature` table we must use `creature.equipment_id` (can be -1, it means random equipment)
+            m_originalEquipmentId = data->equipmentId;
+            LoadEquipment(data->equipmentId);
+        }
+        else
+            LoadEquipment(0, true); // Line in `creature` table exist, but `creature.equipment_id` == 0. Creature must spawn without equipment.
```
Shauren: This is definitely not correct - the intent is that creature.equipment_id does not have to use a value different than 0 for a "default selection"
Want one spawn have a different equipment than template? use creature.equipment_id
Want one spawn without equipment? define no default in creature_template and set creature.equipment_id to all spawns except then one you want to exclude (rare use case)

Summing up, this change breaks all existing `creature` spawns that rely on the behavior I described (which was what was intended when this got implemented)

Tonghost:
@Shauren You say that `creature.equipment_id = 0` means "default selection", but @Killyana said that creature must spawn without equipment in this case (#16433). So I'm confused.

> Want one spawn without equipment? define no default in creature_template and set creature.equipment_id to all spawns except then one you want to exclude (rare use case)

Some creatures may be not exist in `creature`

DDuarte:
This is the commit that changed how equipments work: 028c72a9f2c606e8af391d284931c60247ad0160 (3 years ago, by me). At the time I didn't think about summoned creatures but do **not** change the behaviour of the values already in the DB.

ariel-:
Maybe revert cc3ae640a48417aae31eac8828842ab50099069b and set that a value of -1 in `creature.equipment_id` means "no equipment at all", and 0 means "default value, just use random equip", altough that makes it exactly the opposite of the core (0 = noequip, -1 = randomequip)

Other solution to this problem could be manually setting to -1 rows in the spawn DB, were random equip is intended.

cronayu:
as @ariel- said, reverting the deletion of those 2 main lines fixed the issue of creatures spawning without equipment

Trisjdc:
I agree with @ariel-. Add a -1 value for no equipment, maybe change the constants on the source code for consistency.

The code on the pull doesn't work for a bunch of existing creatures otherwise.

sirikfoll:
@Tonghost Won't you be working with this issue anymore?

tkrokli:
Would the content in this PR make any difference for issue #16759 ?

PR #16641, opened 2016-02-18
Core/Spells: check aura dummy cooldown instead of trigger cooldown

### LINKED ISSUE #10840: [3.3.5] Core/Spells: Sacred Shield on Pets
Spell: http://wotlk.openwow.com/spell=53601
Effect: http://wotlk.openwow.com/spell=58597

All works well with this spell except when you cast it in pets, if you cast http://wotlk.openwow.com/spell=53601 in pets, the effect http://wotlk.openwow.com/spell=58597 doesnt have internal CD, maybe it has easy fix in spell_paladin.cpp in

```
void HandleEffectRemove(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
        {
            if(Unit* pCaster = GetCaster())
                if(pCaster->ToPlayer())
                    pCaster->ToPlayer()->AddSpellCooldown(PALADIN_SPELL_SACRED_SHIELD_EFFECT, 0, time(NULL) + 6);
        }
``` 
If you add  pCaster->ToCreature()->_AddCreatureSpellCooldown it doesnt work.
It its very abusable if paladin with warlock partner cast sacred shield on pet, the pet is immortal because everytime it has the shield.

ariel-:
**Changes proposed**:
- Changes the check of cooldown on aura dummy triggered spells to a check of the triggering aura itself (except for a few where we would end up adding a cd on a player spell)

**Target branch(es)**: 335

**Issues addressed**: Closes #10840 

**Tests performed**: Builds, tested IG

**Known issues and TODO list**: Internal item CDs, remove TYPEID_PLAYER checks

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -22482,19 +22482,37 @@ void Player::ApplyEquipCooldown(Item* pItem)
             continue;
 
         // wrong triggering type (note: ITEM_SPELLTRIGGER_ON_NO_DELAY_USE not have cooldown)
-        if (spellData.SpellTrigger != ITEM_SPELLTRIGGER_ON_USE)
+        if (spellData.SpellTrigger > ITEM_SPELLTRIGGER_CHANCE_ON_HIT)
             continue;
+
+        uint32 equipCooldown = 30; // default equip cooldown is 30 seconds for 'On Use' spells
+
+        switch (spellData.SpellTrigger)
+        {
+            case ITEM_SPELLTRIGGER_ON_EQUIP:
+            case ITEM_SPELLTRIGGER_CHANCE_ON_HIT:
+                if (SpellProcEventEntry const* procEvent = sSpellMgr->GetSpellProcEvent(spellData.SpellId))
+                    equipCooldown = procEvent->cooldown;
+                break;
+            default:
+                break;
+        }
 
         // Don't replace longer cooldowns by equip cooldown if we have any.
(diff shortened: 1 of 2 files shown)


Shauren:
Does not fix #10840 - there still is a TYPEID_PLAYER check
Also breaks all trinkets which proc from a dummy aura if you unequip and equip them again while the proc is on cooldown.

ariel-:
i have tested it casting SS on a pet and it does not proc back to back

Shauren:
*updated comment

ariel-:
Well, it can be merged with this PR, what do you think about starting internal trinket CD on item equip?
(as it is now, only items with a "on use" spell trigger the 30 sec cd), starting internal cd on equip should be blizzlike.

ariel-:
It is a compromise solution as I understand it, previously the general rule was "check triggered spell cd", and there were some exceptions that "check triggering aura cd", now the rule is "check triggering aura cds" with the exception being "check triggered spell cd", like SS case, or Imp. Leader of the pack

Shauren:
You misunderstood my comment. Casting the proc aura itself will fail due to cooldown (and it will never be retried) leading to non-functioning trinket.

ariel-:
Equip Spells are casted triggered, with the changes in 3f886025a2559c4973ac260fc633ff74c2096932, this shouldn't happen.

joschiwald:
btw. QAston added with his new proc system the methods Aura::IsProcOnCooldown and Aura::AddProcCooldown to handle this problem

ariel-:
Interesting... so the plan originally was to make Auras responsible of procs instead of Unit like now?

joschiwald:
yeah, but there is a problem, how to handle spells like "Leader of the Pack"?

Shauren:
Closing, I'm planning to clean all of proc in the near future, finishing the new proc system

PR #16650, opened 2016-02-21
Core/Misc: DB size optimization, don't store default values to database

ariel-:
**Changes proposed**: 
- Move usually unused columns to separate tables.
- Don't store default values in database.
- Split up and unblob several characters and item_instance fields, most of them were always default values.
- This will save data for cases where the values are distinct from default, producing a noticeable decrease in size of the characters DB

**Target branch(es)**: 335

**Tests performed**: (Does it build? Tested in-game?) Build correct with MSVC2015upd1, tried to break it as much as I could (deleting and restoring character, then perm delete, etc.)

**Tests by Aokromes**:

```
Original Database ~50k characters:
1556M Feb 19 07:02 characters.sql

Optimized Database:
1241M Feb 19 17:10 characters_optimized.sql
```

**Tests by Treeston**:

```
Original Database test 2 97 characters:
before 3.2mb

Optimized Database test 2:
615.4kb
```

**Tests by xjose93**:

```
Original DB:
4680.646 MB
Optimized DB:
4065.738 MB

Character count: 128533
```

**Known issues and TODO list**: None as I know.


DDuarte:
Judging by Aokromes' reports the optimizations in the DB size are very clear (the size needed in a beefy database is reduced a lot) however it's not clear if the optimizations have any impact the query performance.

sql/updates/characters/2016_02_19_00_characters.sql
```diff
@@ -0,0 +1,274 @@
+--
```
DDuarte: Missing updates on `sql/base/characters_database.sql`.

ariel-:
The performance shouldn't matter too much, as most of queries are executed async, maybe for the client is some extra delay when logging in, but shouldn't be much of a difference core-wise.

Aokromes:
Also, shauren said: Shauren> probably a bit slower queries but these are joins on indexs so its not THAT bad

jackpoz:
please profile your changes and compare them with the previous behavior.

ariel-:
@jackpoz: How to do that? I'm using windows 64 bit and MSVC 2015 upd 1

jackpoz:
check MySQL documentation and see if they provide any tool to profile query performance

xjose93:
From 4680,646mb to 4065,738mb.

AriDEV:
´´´
From: 1448.960mb
To: 1159.168mb
´´´

ariel-:
Profiling data (thanks @AriDEV)

```
Before optimization: 54.013 ms
After optimization: 60.84 ms
```

Of course the results of only one run with each core can't be taken into account, but here it is for reference.

On each core (restored DB, applied update manually and did the following steps);
 - Login account
 - Login with character 1
 - Logout character
 - Delete character 2 (worldserver set to "unlink" delete)
 - Logout account
 - In server console I do .character deleted restore 2
 - Login account again
 - Delete character 3
 - Logout account, now in server console:
 - .character deleted list
 - .character deleted delete 3
 - .character erase 2
 - .saveall

src/server/game/Entities/Player/Player.h
```diff
@@ -2097,8 +2092,9 @@ class Player : public Unit, public GridObject<Player>
         uint8 m_forced_speed_changes[MAX_MOVE_TYPE];
 
         bool HasAtLoginFlag(AtLoginFlags f) const { return (m_atLoginFlags & f) != 0; }
-        void SetAtLoginFlag(AtLoginFlags f) { m_atLoginFlags |= f; }
```
Vincent-Michael: revert this rename pls

Aokromes:
@ariel- can you update this? :)

jackpoz:
this PR has come down to the question: do we consider database space reduction at the cost of query performance (and database complexity with multiple joins instead of denormalized database) an optimization ?

jackpoz:
it is db complexity if I have to write a SELECT with 10 JOINs instead of a simple SELECT with 0 JOINs to get the same data

Aokromes:
What i mean is, the pr is "DB size optimization, don't store default values to database" not "DB complexity reduction" if you have a magic way to reduce db size without increase complexity it's welcomed. Anyway it also reduces complexity of some tables. (killing blobs)

jackpoz:
as written at https://github.com/TrinityCore/TrinityCore/pull/16650#issuecomment-226763214 , it's not a matter of finding a way to reduce db size without increasing complexity, it's a matter to decide if we want to decrease the db size at all at cost of the 2 drawbacks described

Aokromes:
as i wrote on https://github.com/TrinityCore/TrinityCore/pull/16650#issuecomment-226764480 it's not the 1st time we merge such things, quest_template_addon, split character fields, etc, if those where merged, i don't see why this cannot be merged.

jackpoz:
where did you read it cannot be merged ?

Aokromes:
@ariel- irc :P

malcrom:
Default assumes all have same value and then we can't get stuff to work because its the same for all creatures. Like right now having some of an entry fly while others walk.

PR #16651, opened 2016-02-21
Core/Unit: Fix typo

ariel-:
**Changes proposed**: immuneMask needs to be shifted by (mechanic - 1) instead of (mechanic)
- Caused creature to be immune to SPELL_AURA_USE_NORMAL_MOVEMENT_SPEED only when creature had either MECHANIC_STUN or MECHANIC_DISCOVERY in its immunemask, instead of the intended.

**Target branch(es)**: 335/6x

**Tests performed**: (Does it build? Tested in-game?) Aye

**Known issues and TODO list**: None as I know.


Goatform:
I think that it will be nice if you report that issue on tracker and then link that issue.

PR #16654, opened 2016-02-21
Core/Spells Create fallback for item-cast spells while character is still logging in

### LINKED ISSUE #16604: [6.x] Core: Equiped Heirloom items crash worldserver on login
Logging into a character that has a Heirloom equipped crashes the world server.

Crashlog: [52758c1a0b8e_worldserver.exe_[13-2_21-16-13].txt](https://github.com/TrinityCore/TrinityCore/files/129456/52758c1a0b8e_worldserver.exe_.13-2_21-16-13.txt)

In the function void WorldSession::HandlePlayerLogin(LoginQueryHolder* holder) in CharacterHandler.cpp

LoadDB will load all items, including the Heirloom, these Heirlooms have the following spells (depending on the item):
http://www.wowhead.com/spell=57353/heirloom-experience-bonus-10
http://www.wowhead.com/spell=71354/heirloom-experience-bonus-5

Which the system tries to cast in void Spell::prepare(SpellCastTargets const* targets, AuraEffect const* triggeredByAura)

However the GetOwner() function attempts to retrieve the Player out of a HashMapHolder<Player>, which isn't filled until way later in HandlePlayerLogin(LoginQueryHolder* holder) (line 1023 in CrashHandler.cpp) which causes a chain of null returns etc.

Reproduction:
1. Make a new character
2. .additem in any heirloom (see collection overview)
3. equip heirloom
4. leave game
5. join game
6. crash

TrinityCore rev. 52758c1a0b8e 2016-02-12 20:40:43

KarrokDC:
**Changes proposed**:
- Create a fallback for spellcasting items to cover character-loading.
- Create an early-exit if both primary and secondary method of items casting spells fail

**Target branch(es)**: 6x

**Issues addressed**: Fixes #16604

**Tests performed**: (Does it build? Tested in-game?)

After building, I used a low level character and straight upon logging in killed a creature of the same type and level. A couple of times while logging in in full heirloom gear and several times without to ensure the XP bonus is active.

no known issues due to presumibly small testset (only used full heirloom agi leather + jewelry/trinkets)

closes #16604

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2861,7 +2861,22 @@ void Spell::prepare(SpellCastTargets const* targets, AuraEffect const* triggered
     {
         m_castItemGUID = m_CastItem->GetGUID();
         m_castItemEntry = m_CastItem->GetEntry();
-        m_castItemLevel = int32(m_CastItem->GetItemLevel(m_CastItem->GetOwner()));
+
+        Player* owner = dynamic_cast<Player*>(targets->GetUnitTarget());
+        if (m_CastItem->GetOwner() != nullptr)
+        {
+            m_castItemLevel = int32(m_CastItem->GetItemLevel(m_CastItem->GetOwner()));
+        }
+        else if (owner)
+        {
+            m_castItemLevel = int32(m_CastItem->GetItemLevel(owner));
+        }
+        else
+        {
+            SendCastResult(SPELL_FAILED_EQUIPPED_ITEM);
+            finish(false);
+            return;
+        }
     }
 
     InitExplicitTargets(*targets);


src/server/game/Spells/Spell.cpp
```diff
@@ -2861,7 +2861,22 @@ void Spell::prepare(SpellCastTargets const* targets, AuraEffect const* triggered
     {
         m_castItemGUID = m_CastItem->GetGUID();
         m_castItemEntry = m_CastItem->GetEntry();
-        m_castItemLevel = int32(m_CastItem->GetItemLevel(m_CastItem->GetOwner()));
+
+        Player* owner = dynamic_cast<Player*>(targets->GetUnitTarget());
+        if (m_CastItem->GetOwner() != nullptr)
```
Keader: I think it could be ```if(m_CastItem->GetOwner())```

KarrokDC: Possibly indeed, I'll leave this one to veteran TC devs to make the call and apply the suggested change.

Shauren:
Merged - 7afb542a493813b7e53d53b2c72102f688562aba

PR #16666, opened 2016-02-22
[3.3.5] Fix bosses losing interest in the person running at them

Treeston:
Core/Threat: Prevent bosses (and other stuff using DoZoneInCombat) from switching off of the person pulling if they are pulled without generating threat.

Why? Because anything that used DoZoneInCombat in the EnterCombat hook (like the default BossAI implementation) would add players to the threat list in an unpredictable manner (rather than adding the person pulling first), thus causing the boss to go after a random target rather than the one pulling.

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -134,7 +134,10 @@ void CreatureAI::MoveInLineOfSight(Unit* who)
         return;
 
     if (me->HasReactState(REACT_AGGRESSIVE) && me->CanStartAttack(who, false))
+    {
+        me->AddThreat(who, 0.0f); // ensure our initial target is the first thing added to threat list so we don't randomly switch off if DoZoneInCombat is called during the EnterCombat hook
         AttackStart(who);
+    }
     //else if (who->GetVictim() && me->IsFriendlyTo(who)
     //    && me->IsWithinDistInMap(who, sWorld->getIntConfig(CONFIG_CREATURE_FAMILY_ASSISTANCE_RADIUS))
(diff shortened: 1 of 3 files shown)


src/server/game/AI/CreatureAI.cpp
```diff
@@ -134,7 +134,10 @@ void CreatureAI::MoveInLineOfSight(Unit* who)
         return;
 
     if (me->HasReactState(REACT_AGGRESSIVE) && me->CanStartAttack(who, false))
+    {
+        me->AddThreat(who, 0.0f); // ensure our initial target is the first thing added to threat list so we don't randomly switch off if DoZoneInCombat is called during the EnterCombat hook
```
Shauren: wouldnt just adding 1 threat instead of 0 do the trick? (hack, i know)

Treeston: Not sure what the benefit would be?

Shauren: not changing ThreatManager?

Treeston: Is there really any relevant overhead to an emptiness check on a container?

Shauren: none. was rather meant as smaller change more than anything else

ccrs: I think it should be added in Attack or in AttackStart (attack may return false)

Shauren: valid point about returning false - but it cannot be in AttackStart because that can be overridden by scripts (and you really dont want to care about such detail in every script yourself if you do that)

PR #15796, opened 2015-11-01
Blink v4

kvipka:
- -[x] blink in abyss
- -[x] blink by path ( rechecking each step )
- -[x] where player can walk - blink will available there
- -[x] implemented check on dynamic objects (locked doors and etc)
- -[x] implemented total path of blink.

video : http://www.youtube.com/watch?v=wpymzrkwNTg

PR #16688, opened 2016-02-26
Blink v4 (cont'd)

Treeston:
(Copied from previous PR #15796)
- -[x] blink in abyss
- -[x] blink by path ( rechecking each step )
- -[x] where player can walk - blink will available there
- -[x] implemented check on dynamic objects (locked doors and etc)
- -[x] implemented total path of blink.

video : http://www.youtube.com/watch?v=wpymzrkwNTg


kvipka:
stop it

DDuarte:
@kvipka: what's wrong?

kvipka:
this pr will not applied in TC, just wait a Naios version and close this pr

w5860363:
@DDuarte This is now in the test https://github.com/TrinityCore/TrinityCore/pull/15796

Treeston:
Bug for whoever picks this back up:
Blinking underwater can lead to the player falling through the world and dying.

kvipka:
Video, place, info?

Treeston:
`.go xyz 2682.910400 662.913391 -28.269423 571 1.356954`

Run straight ahead "into" the slope and cast blink while doing so.

jackpoz:
@kvipka any news about the issue reported above ?

Shauren:
Closing.
Still has issues, noone really know what is going on there, attempts to refactor have failed, multiple expensive vmap calls

msoky:
if is that better and less buged than blink which is actually used, then maybe someone can push new blink system and fix minor bugs later..

Treeston:
It's not a case of "minor bugs", it's a case of "potentially significant performance impact".

That being said, if you would like to use it on your server, there is nothing stopping you from merging the PR into your local deployment branch and using it.

kvipka:
upd: http://pastebin.com/z0FSezvi

tkrokli:
@kvipka : of course. You tried to push to https://github.com/TrinityCore/TrinityCore, not your own repository. Try pushing to your own repository, which should be something different than "origin". You need to fork the TC repository to your user account here on github to push any updates you want to add.

kvipka:
tkrokli, exactly

But branch with blink in trinity repo. Not PR.

tkrokli:
Yes, but only the TrinityCore **members** (members in the developer team with TC access) can push to the https://github.com/TrinityCore/TrinityCore.git location. All us other users need to update a personal fork where the PR originated (if you are the author of the PR in question) to update a current PR.

jackpoz:
You can pull TC repo, swtch to blink branch, apply your changes, push to your own github fork and link here your branch, no need to push to TC repo

kvipka:
okay=(

jackpoz:
https://github.com/TrinityCore/TrinityCore/compare/3.3.5...kvipka:blink shows the diff between 335 and your branch

Keader:
Ok, i'm using a old version about it.
Using code present in this PR, blink is working?

Treeston:
The PR itself works just fine. Most of the concerns are with code style and/or readability.

@kvipka feel free to create a new PR about this if you've cleaned up the code significantly. I only created this as a back-up since you seemed to be deleting your commits on the old branch.

sirikfoll:
@kvipka  Not really important, just to let you know, in Dalaran Arena, if you use blink on certain places, on the top platform,you will be teleported to the lower plataform, video: https://youtu.be/y4iWICR6mqU

jackpoz:
The mix of missing comments, variables reused, magic numbers and strange code style (like declaring a index variable outside of for() loop and not using it anywhere else) makes the code less readable than how it could be and kinda unclear.
It would be nice to have a proper detailed explaination of the issues, the solution and how it was implemented. From what I understand the idea is to check the vmap Z of the whole blink path, little step by little step ? That doesn't sound enough to fix all issues, is it expected that there will still be cases where this patch might fail ?

Treeston:
Does vmap/mmap not have the capability of "move along the ground distance `d` in 2d direction `(x,y)`"?
I'm not sure about the implementation, but I assume that mmaps is some sort of polygonal approximation of the floor, which should make this kind of along-the-floor movement of a `Position` possible (and not too expensive either).

Maybe @Shauren might know, it's something that came up in discussion with @r00ty-tc before about contact point logic.

r00ty-tc:
There isn't a "move in x direction following terrain" option. But, mmap should provide something similar if the end co-ordinate has a valid, or close to valid Z. One of the things I was working on was using mmap to get a better Z fix for movement. But, it created other problems.

With vmaps you can use rays (pointing toward a texture, not from underground) to find ground level. But, you need to be sure you're starting from above (but not the next floor above).

I was trying to work out a solid "work all the time" method to do this with current vmap (and vmap/mmap in unison) tools. I couldn't find a solution which wasn't A: A stopgap measure which only fixed some cases. B: Fix many cases, create new broken cases. C: Took significantly more rays to achieve thus hurting performance.

To me, it's a serious problem. But finding a programatically cheap solution has alluded me, so far.

kvipka:
@sirikfoll , good info -> checking

Treeston:
> There isn't a "move in x direction following terrain" option.

Not right now, no - my question is whether this would be possible (and if not, why it isn't). Following a polygonal surface in a direction is not exactly rocket science.

r00ty-tc:
The main problem I found wasn't following the mmap polys. There are some places where the mmap Z  isn't definitive in which valid vmap z it relates to. The resolution is low such that z varies by up to 6.0f in places I found and potentially even more. Now the question isn't whether the problem can be solved. There could be a whole decision tree to resolve all possible outcomes. The issue arises, as to the processing cost. When it got too high, I stopped.

jackpoz:
@Treeston MMap Z approximation is totally unreliable, so ppl often break the path into small steps and fix the Z using vmap data (much more accurate).

I'm not even sure which are the current issues and which part of the code fixes them:
- blink in abyss : what does this mean ? destination Z is below the floor and the player falls below the ground ? or destination X/Y is behind a wall/rock and player falls below the ground?
- blink by path ( rechecking each step ) : is this a straight path ? what kind of check is done in each "step" (whatever "check" and "step" mean in this context)
- where player can walk - blink will available there : how is this implemented ? on which data does this "player can walk" condition relies on ?
- implemented check on dynamic objects (locked doors and etc) : is this for both blinking on top of dyn objects and through dyn objects ? does it support phasemask too ?
- implemented total path of blink : I'm not sure what "total path" means here and how is different from current behavior (current is only a "partial path" ?)

kvipka:
@sirikfoll  https://www.youtube.com/watch?v=xYiQ5Qweodw&feature=youtu.be

Keader:
looks nice now.

sirikfoll:
@kvipka  You weren't able to reproduce or have you fixed it?

kvipka:
i did nothing. idk what blink you are using. Try to use my vmaps and etc. https://drive.google.com/file/d/0B95oF05riRW0VEkxOHo3SlhyU0E/view?usp=sharing

sirikfoll:
@kvipka  I'm using clean trinityCore + blink pr, 
It only happen in that exact case that i showed in the video, orientation matters too `.go xyz 1291.749634 798.676392 7.114820 617` facing the crystal and `.go xyz 1331.003174 763.371399 3.1590707 617` facing the right edge of the boxes

kvipka:
check your code. I think you using clear blink version (TC version).

sirikfoll:
@kvipka  This one: https://github.com/TrinityCore/TrinityCore/compare/3.3.5...kvipka:blink

kvipka:
i can't reproduce

wrathix:
why not push this and fix whatever small issues it may have in the future?  Its certainly better than the current blink

tripleslash:
Why don't you just use the recast raytracing method for this. I believe the method is called moveAlongSurface. You only need to fix the height of the final point that you get from moveAlongSurface after because the poly mesh has very low quality heights.

Treeston:
@r00ty-tc same question I asked for movement tracing. You've looked into it?

Shauren:
Just no.

PR #16710, opened 2016-03-02
Spells/SpellMgr: Add a LoadSpellInfoCorrection case for Remove Weapon…

Heaven31415:
**Changes proposed**:

Spell named **Remove Weapon** (_108055_) will be used at you if you create a Pandaren Mage character. This spell job is to delete your main and off-hand weapon, however without this correction, off-hand weapon isn't deleted. This bug is caused by wrong data in Spell Effect 1. Instead of target TARGET_UNIT_CASTER there is a NO_TARGET set.

One may argue whether this kind of correction is really needed and I think the answer for this question is: yes.

Mage is not the only one character which is using main and off-hand weapon. There is also a rogue and shaman. Rogue is using spell with ID: _108058_ and shaman spell with ID: _108056_. This spells have the same Spell Effects structure as _108055_ do, however for Spell Effect 1 they use TARGET_UNIT_CASTER instead of NO_TARGET which is not the case for _108055_.

After making 108055 identical to mentioned spells in case of target, everything is working well.

**Target branch(es)**: 6x

**Issues addressed**: I haven't been able to find issues related to this problem.

**Tests performed**: I have built it myself and tested, everything was working well.

I don't know any other issues related to this matter and there is nothing left to do.

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -2999,6 +2999,9 @@ void SpellMgr::LoadSpellInfoCorrections()
 
         switch (spellInfo->Id)
         {
+            case 108055: // Remove Weapon (should have TARGET_UNIT_CASTER instead of NO_TARGET)
+                const_cast<SpellEffectInfo*>(spellInfo->GetEffect(EFFECT_1))->TargetA = SpellImplicitTargetInfo(TARGET_UNIT_CASTER);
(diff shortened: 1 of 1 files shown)


Shauren:
Thanks for your contribution but your changes are only targeting a single spell - cf2fead644fe76b3d1b2814ebf1877c8d7ee516c fixed this in a generic way for all spells.

Issue #16549, opened 2016-02-11
[3.3.5][6.x] Core/Creature: Missing Z Position update in Combat


sirikfoll:
Example:
The boss gets out of melee reach, but for some unknown reason, he do not move closer to the player
Does not seem to be an "out of LoS" problem, because if used something like `me->IsWithinLOS(me-GetVictim())` , the return value will be true.
Guess the Z position value is not being handled well somewhere

Video: https://www.youtube.com/watch?v=i1e_UqyJTAs

vmaps and mmaps are enabled.

hash b4b43d03b3fb
branch 3.3.5

P-Kito:
Oh, probably since https://github.com/TrinityCore/TrinityCore/commit/89d8d50b2e1185a14896bd8b8b6679da578d4ed4

Keader:
@P-Kito nop, it happened before that.

Rushor:
i think this can be reproduced with all creatures. (@Keader, yep also happens on a 2 years old rev.)

Shauren:
It looks like creature is in range of player just fine (2d distance) but attack range is checked in 3d so it fails like that

P-Kito:
Anyone who is interested in checking the math behind this, here you can find it: 

bool Unit::IsWithinMeleeRange(const Unit* obj, float dist)
bool Unit::IsWithinCombatRange(const Unit* obj, float dist2compare)

Demonid:
Didn't someone say that if the boss cant reach the player, it must go into evade mode ? Even if that not the issue over here, it still a nice feature that should be added.

P-Kito:
That doesn't fix the issue, the boss will then randomly evade if we don't fix this problem first.

chaodhib:
The problem lies in `TargetedMovementGeneratorMedium<T, D>::_setTargetLocation`. Except for the fact that this method needs a rewritte because it's a mess, it uses `i_target->GetContactPoint(owner, x, y, z);` which works in 2D: 
```
GetNearPoint2D(x, y, distance2d+searcher_size, absAngle);
z = GetPositionZ();
...
```

I'll see what I can do.

chaodhib:
Is it even possible to walk up these stairs when the Lich King is engaged in retail? Cause handling correctly this type of situation (mob with a big combat reach + high slope) will complicate the code, without adding much benefit for most of the time.

chaodhib:
[...]

PR #16736, opened 2016-03-06
Solve many of the creature falling through floor issues.

r00ty-tc:
Remove check for map height above current Z in getHeight:

This check has been present forever as far as I can see, but to me the premise seems flawed. Generally getHeight is used to find the correct ground height and more often than not the problems of creatures falling through the terrain are because a relative location to a known good location is used, and often the Z is below the ground the core wants to move the object to. This check limits the ability to choose options above the current ground level to vmap lookups.

This is a PR for a simple change because, it's called EVERYWHERE and has the potential to break a lot of things if it's not a perfect solution. Hence I'd like some testing.

Target Branch: 335, potentially 6x

Tests Performed:

Pet pathing in Sepulcher is resolved, creature pathing around the temple close to Gundrak, and many other places.

It doesn't resolve even close to all the problems obtaining the Z position. But it seems to improve, and from my testing so far not make matters worse anywhere.

Potentially this could cause problems. But in my testing so far I've not found any examples.

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -2328,9 +2328,9 @@ float Map::GetHeight(float x, float y, float z, bool checkVMap /*= true*/, float
         {
             // we have mapheight and vmapheight and must select more appropriate
 
-            // we are already under the surface or vmap height above map heigt
+            // vmap height above map height
             // or if the distance of the vmap height is less the land height distance
-            if (z < mapHeight || vmapHeight > mapHeight || std::fabs(mapHeight - z) > std::fabs(vmapHeight - z))
+            if (vmapHeight > mapHeight || std::fabs(mapHeight - z) > std::fabs(vmapHeight - z))
                 return vmapHeight;
             else
                 return mapHeight;                           // better use .map surface height


Keader:
This fix https://github.com/TrinityCore/TrinityCore/issues/16549 ?

r00ty-tc:
No, that seems to be a discrepancy between distance checks. For moving into range it seems to use 2d distance, for checking melee range it seems to be using 3d distance. Therefore it moved into 2d range, but cannot reach in 3d range.

Just a guess. But, that's how it looks. Nothing to do with this change.

Rushor:
one good exampe to test this can be found here http://i.imgur.com/3BlfEYl.jpg

without the fix, all creatures drop below the ground at this position

tkrokli:
2 spots where NPCs consistently fall through the map (from what I remember):
- path between Sputtervalve and the bank building in Ratchet (Ratchet Bruiser pathing)
- Rellian Greenspyre in Darnassus, grass corner on the opposite side of the road (pathing Protectors)

r00ty-tc:
Just to clarify. It's great to hear what it may fix (and it's good for people to list problems spots). But I know it doesn't fix everything (I spent quite some time trying to produce something that wouldn't eat CPU, but resolves all the times it goes wrong, and gave up). I can list some places that will still cause problems.

Mainly what I'd like to know is where this might cause a problem that didn't previously exist. There's no point fixing one issue and introducing another.

jackpoz:
Please include some positions that got fixed by this PR with the expected Z so we can write some unit tests too in the future.

Riztazz:
IIRC in duskwood, on west side of the map(near zombies) pets and other creatures used to fall to cave below, stand behind fence and send pet on one of the zombies

http://i.imgur.com/04Avb54.jpg this spot

Raydor:
https://github.com/TrinityCore/TrinityCore/issues/15032#issuecomment-124469517 This issue (my post, can't test the OP issue) is fixed by your code, tested a lot and doesn't seem to generate other issues in that place.

r00ty-tc:
OK so, the positions I tested:

**Pet pathing:**
 - The Supulcher (Map 0, [522.150, 1640.514, 125.194]). Pet pathing will often fall through the the ceiling of the crypt below. Usually when the pet's "position" beside the player is on the raised ground. It places the pet slightly under the ground and puts it onto the next "floor" it can find. Note that this doesn't fix pet pathing from failing if you take the pet into the mountain wall.

**Creature pathing:**
 - XT-002 (Map 603, [886.960, 2.866, 409.714]). XT's path will usually make him fall underground.
 - Altar of Quetz'lun (Map 571, [5815.150, -4420.841, 389.454]) - Creatures pathing over the mounds of snow will fall through them to the sunseen floor below.

tkrokli:
I can confirm that by now, using rev. b82ffb88f7e6 2016-04-26 22:25:16 +0100 (3.3.5 branch),  my 2 examples in https://github.com/TrinityCore/TrinityCore/pull/16736#issuecomment-192947256 appear to be fixed, which suggests that many NPC Z axis map issues are now fixed.

PR #16739, opened 2016-03-07
SmartScript: Added SetBossState support

ariel-:
**Changes proposed**:
- Adds Boss State capability to smart scripts, making it possible to fully script simple bosses using DB
- Still needs SetBossNumber set on instance .cpp, done this for DIre Maul, which has every boss with SmartAI

**Target branch(es)**: 335/6x

**Issues addressed**: Will help fixing DM tribute run (commit referenced)

**Tests performed**: (Does it build, tested in-game, etc)  Build correct

**Known issues and TODO list**: None as I know

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -929,9 +929,21 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                 break;
             }
 
-            instance->SetData(e.action.setInstanceData.field, e.action.setInstanceData.data);
-            TC_LOG_DEBUG("scripts.ai", "SmartScript::ProcessAction: SMART_ACTION_SET_INST_DATA: Field: %u, data: %u",
-                e.action.setInstanceData.field, e.action.setInstanceData.data);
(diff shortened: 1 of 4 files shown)


Shauren:
fkin hell... bosses arent supposed to be scripted in sai

NorthStrider:
@Shauren  the newer complex one yes, but it looks to me that this can be used for old classic bosses since there seems to be some sort of client notification if a boss dies according to that boss state UI element.

soulfrost:
@NorthStrider
```
some sort of client notification if a boss dies in WoD according to that boss state UI element.
```
made by SMSG_BOSS_KILL_CREDIT, and encounterID from by taken from DungeonEncounter.db2 - where row should be finded by creature_display_id

Runico:
Please add an options for heroic attempts, like ToC

ariel-:
Heroic attempts counter is handled by world states, this are the kind of instances that should be handled by a cpp script.

joschiwald:
Dislike

ariel-:
There are still some instances which don't have AIs in .cpp, but in SmartScript, like DIre Maul case, what should be done with those then? I think moving the (simple) scripts to cpp is overkill.

Aokromes:
I don't see any reason to don't merge this for custom projects.

PR #16763, opened 2016-03-11
Added locales id for "ptBR" and "itIT" on worldserver.conf.dist

### LINKED ISSUE #16762: [6.x] Add suporte to ptBR DBC Locale
Please, add suport to ptBR locale on DBC... =)

mateusdemboski:
Added locales id for "ptBR" and "itIT" on worldserver.conf.dist

**Target branch**: 6x

**Issues addressed**: Fixes #16762

### DIFF
--- a/src/server/worldserver/worldserver.conf.dist
+++ b/src/server/worldserver/worldserver.conf.dist
@@ -589,6 +589,9 @@ StrictPetNames = 0
 #                     6   - (Spanish)
 #                     7   - (Spanish Mexico)
 #                     8   - (Russian)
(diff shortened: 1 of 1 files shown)


MitchesD:
eh, how it can fix the issue? it's only a comment in config

r00ty-tc:
@MitchesD I thought that too. But it looks like there is at least the ENUMs in the source and they match the config. So, might just have been an oversight that they weren't added to config.

Shauren:
Merging for the sake of clarifying config, ackowledged that this fixes nothing.

PR #16772, opened 2016-03-12
[3.3.5/Comp/Core] Activate creatures and objects during opening cinematics and other flyby

### LINKED ISSUE #4755: Creature/Object View Distance During Loading
So this issue has been persistant among all of the WoW private server distros since the beginning. Basically you know how in the .conf file you can specify view distance for objects, creatures, players ect... Well when a character is first created and the game takes you through the intro video around the terrain of the game you do not see any animals, objects, players, ect... Until you get within the view distance of your character, that you specified in the .conf file... Well this is not blizz-like, all animals, objects, and players should be noticeable during the intro video then it should return back to the conf file settings.

The way to circumvent this is to have the view distance for objects, players, creatures overridden from what is in the configuration file during the loading screen. You would change the view distance for all options to the radius of the starting zones.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65473-creature-object-view-distance-during-loading?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

r00ty-tc:
I started this, thinking it unlikely I'd work it out. Hence the branch name folly_cinematic. It will essentially show objects/npc's during the cinematic fly by, for character creation and elsewhere in the game.

Anyway, it all seems to work, but I wonder if the solution is a little hacky. Hence I've chosen to PR this first. Comments and suggestions would be good as well as some testing.

Target branch: 335 (for 6.x the camera structure needs changing)

Fixes #4755 

Tests performed: (Does it build, tested in-game, etc)

Builds fine, Travis OK. Ran with helgrind and valgrind (valgrind spotted a problem and it's resolved now). Tested with each race. They all work.

Known issues and TODO list:
- I did see at some point client crashes caused. I think these were due to invalid co-ordinates sent. I've since added checks to the "look ahead" time on the cinematics and it's not happened since. It will still cause a problem if a cinematic is started in the wrong map I believe. Which shouldn't happen in normal operation.
- There was a server crash, Resolved when valgrind spotted a problem.


r00ty-tc:
Just updated it to the changes committed today (which added conflicts).

src/tools/map_extractor/System.cpp
```diff
@@ -1025,6 +1026,48 @@ void ExtractDBCFiles(int locale, bool basicLocale)
...
+        std::string filename = path;
+        filename += (iter->c_str() + strlen("Cameras\\"));
+
+        if (boost::filesystem::exists(filename))
+            continue;
+
+        if (ExtractFile(iter->c_str(), filename))
+            ++count;
+    }
+    printf("Extracted %u camera files\n\n", count);
```
Goatform: `\n\n`

I know that it's just visual but I think that you could use it just one time.

`\n`

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1062,11 +1062,19 @@ void WorldSession::HandleSetActionButtonOpcode(WorldPacket& recvData)
...
+    // If player has sight bound to visual waypoint NPC we should remove it
+    if (Player* thisPlayer = GetPlayer())
+        thisPlayer->EndCinematic();
+
     TC_LOG_DEBUG("network", "WORLD: Received CMSG_COMPLETE_CINEMATIC");
 }
 
 void WorldSession::HandleNextCinematicCamera(WorldPacket& /*recvData*/)
 {
+    if (Player* thisPlayer = GetPlayer())
```
Goatform: Codestyle:

```
    if (Player* thisPlayer = GetPlayer())
        thisPlayer->BeginCinematic();
```

joschiwald: GetPlayer() is always not null at this point

src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -866,6 +866,8 @@ void WorldSession::HandlePlayerLogin(LoginQueryHolder* holder)
 
     pCurrChar->SendInitialPacketsBeforeAddToMap();
 
+    bool sentCinematic = false;
```
joschiwald: useless

r00ty-tc: That's actually a hangover from my "test" code that actually implemented it at this point just to see if my idea would work. Totally right that and the setting to true later is pointless.

src/server/game/Entities/Player/Player.h
```diff
@@ -2250,6 +2250,17 @@ class Player : public Unit, public GridObject<Player>
         std::string GetMapAreaAndZoneString() const;
         std::string GetCoordsMapAreaAndZoneString() const;
 
+        // Cinematic camera data and remote sight functions
+        uint32 GetActiveCinematicCamera() const { return m_activeCinematicCameraId; }
+        void SetActiveCinematicCamera(uint32 cinematicCameraId = 0) { m_activeCinematicCameraId = cinematicCameraId; }
+        bool onCinematic() const { return (m_cinematicCamera != nullptr); }
```
Takenbacon: Shouldn't be lower camel cased imo, doesn't follow the naming convention for the _majority_ of TC's functions (or rather, your other functions in this same PR as well).

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1062,11 +1062,17 @@ void WorldSession::HandleSetActionButtonOpcode(WorldPacket& recvData)
 
 void WorldSession::HandleCompleteCinematic(WorldPacket& /*recvData*/)
 {
+    // If player has sight bound to visual waypoint NPC we should remove it
+    if (Player* thisPlayer = GetPlayer())
```
joschiwald: same here, GetPlayer() never be null

while you are on these lines, pls remove the useless network log messages

src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -873,10 +873,14 @@ void WorldSession::HandlePlayerLogin(LoginQueryHolder* holder)
...
+            uint32 cinematicSequence = 0;
             if (cEntry->CinematicSequence)
-                pCurrChar->SendCinematicStart(cEntry->CinematicSequence);
+                cinematicSequence = cEntry->CinematicSequence;
             else if (ChrRacesEntry const* rEntry = sChrRacesStore.LookupEntry(pCurrChar->getRace()))
-                pCurrChar->SendCinematicStart(rEntry->CinematicSequence);
+                cinematicSequence = rEntry->CinematicSequence;
+
+            if (cinematicSequence)
+                pCurrChar->SendCinematicStart(cinematicSequence);
```
joschiwald: you could revert all changes here, nothing changed

src/server/game/Grids/Notifiers/GridNotifiers.cpp
```diff
@@ -120,10 +120,12 @@ void VisibleChangesNotifier::Visit(CreatureMapType &m)
 void VisibleChangesNotifier::Visit(DynamicObjectMapType &m)
 {
     for (DynamicObjectMapType::iterator iter = m.begin(); iter != m.end(); ++iter)
+    {
```
Ascathor: Do we even need brackets here ? Seems redundant imo.

r00ty-tc: Nope. Another spot I had either logging, or added something previously which was removed later.

src/server/game/Entities/Player/Player.h
r00ty-tc: Actually changing it to IsOnCinematic()

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -110,6 +110,11 @@ class debug_commandscript : public CommandScript
         // #cinematicid - ID decimal number from CinemaicSequences.dbc (1st column)
         if (!*args)
         {
+            for (uint32 i = 0; i < sCinematicCameraStore.GetNumRows(); i++)
```
Ascathor: Same here, no Brackets necessary. No front friend!

src/server/game/Maps/Map.cpp
```diff
@@ -706,6 +706,15 @@ void Map::Update(const uint32 t_diff)
 
         VisitNearbyCellsOf(player, grid_object_update, world_object_update);
 
+        // If player is using far sight, visit that object too
+        if (WorldObject* viewPoint = player->GetViewpoint())
+        {
```
Ascathor: Brackets *g* :P

src/server/scripts/Commands/cs_debug.cpp
r00ty-tc: The whole section shouldn't remain. It was for testing :)

Ascathor: Leave it, a debug function for this makes sense. :)

r00ty-tc:
So, code style and this kind of thing seem to have settled down. Can people run some actual tests and see if they can re-create the client or server crashes any more.. Spot any problems with the actual working etc?

src/server/shared/DataStores/DBCStore.h
```diff
@@ -26,6 +26,140 @@
...
+ // Structures for M4 file. Source: https://wowdev.wiki
+template<typename T>
+struct M2SplineKey
+{
+    T p0;
+    T p1;
+    T p2;
+};
+
+struct M2Vector3
```
Shauren: G3D::Vector3

src/server/shared/DataStores/DBCStore.h
```diff
@@ -26,6 +26,140 @@
...
+{
+    float x;
+    float y;
+    float z;
+};
+
+struct M2Vector4
```
Shauren: use G3D::Vector4

src/server/game/DataStores/DBCStores.cpp
```diff
@@ -717,10 +724,222 @@ void LoadDBCStores(const std::string& dataPath)
...
+        angle += 2 * float(M_PI);
+
+    work.x = DBCPosition->x + (distance * sin(angle));
+    work.y = DBCPosition->y + (distance * cos(angle));
+    work.z = DBCPosition->z + z;
+    return work;
+}
+
+// Number of cameras not used. Multiple cameras never used in 3.3.5
+void readCamera(M2Camera const* cam, uint32 /*numCams*/, M2Header const* header, CinematicCameraEntry const* dbcentry)
```
Shauren: You should just do all of this preprocessing in map extractor and only write out stuff you will use in worldserver (and use G3D types for vectors/aabox instead of defining your own)

r00ty-tc: I agree with Vector/AABox (and done). I was mostly lazy to find out what to include for those.

However, moving this to map-extractor seems like a lot more work for nothing.

I'd have to create some container to hold the data.
I'd have to move the processing there.
I'd have to do it after DBC is extracted (and include code to read the DBC files)

Why though? It's run once on server startup and take milliseconds at worst. The files once extracted are already in format.

Just seems like making a lot of work for a most minuscule gain.

Shauren: Because only Vector3/Vector4 guarantee that their memory layout matches the client (AABox does not and may include padding)

Code reading dbc files is already present in map extractor for reading map/liquids (access using row/column indexes)
Also something I see missing is some kind of validation/sanity checks accessing M2 struct members (the ones that cast to pointers!)

r00ty-tc: Yep the validation is a good point. I put the comment acknowledging that the file size boundaries aren't checked and that corrupt files would crash the process. It didn't make it to here.

I'm still largely of the opinion that fixing and validating in place makes more sense. Processing to an intermediate file and then reading that one instead just feels like a real step in the wrong direction.

src/tools/map_extractor/System.cpp
```diff
@@ -1025,6 +1026,48 @@ void ExtractDBCFiles(int locale, bool basicLocale)
     printf("Extracted %u DBC files\n\n", count);
 }
 
+void ExtractCameraFiles(int locale, bool basicLocale)
```
Shauren: This function should access the camera dbc and extract only models pointed by that dbc - reason for this is porting to 6.x, CASC does not have a listfile and you cannot iterate a directory to find cameras

r00ty-tc: It needs changing for 6.x anyway. But it's a good point.

r00ty-tc:
Anything outstanding here codestyle/with the code in general? Has anyone actually done some testing?

r00ty-tc:
Planning on merging this later, unless someone posts here to say why I shouldn't.

jaredjones:
@r00ty-tc Since you seem to know the mapping system so well, do you think you'd be able to fix the issue of NPCs in instances have poor view distances unless they are set to Active(), which is a bad practice? For instance, if you enter Icecrown Citadel, you won't be able to see Lord Marrowgar from across the hall, however on retail you can see him.

r00ty-tc:
Well, at the moment view distance is fixed based on player and location (except for active as you say). In this case there's a view distance for open world (non instanced maps) which I think is 90 yards. Then inside instances which is 150-160 or somewhere there. Finally in battlegrounds the distance is 533.33r (max grid size).

There's two ways to solve this (and maybe both should be an option).

- Creatures can have custom view distance. I think before even considering this I'd need to take a look at the function that decides the view distance. As to whether that would have unexpected problems.
- Custom view distance by zone ID.

A question I'd ask is, was it the case you could see Marrogar from that distance on 3.3.5 retail? Then it's a case of whether bosses in general should have a longer range..

jaredjones:
It's not just bosses, it's also the creatures that you can see down the hall.

Killyana:
Distance view can be modified on worldserverconf
```
# VISIBILITY AND DISTANCES
Visibility.Distance.Continents = 90
Visibility.Distance.Instances = 170
Visibility.Distance.BGArenas = 533

msoky:
version for 7.x is here https://github.com/r00ty-tc/TrinityCore/commit/3b8da259e8fd2ea6f7f2df725ec904722058b800

joschiwald:
no, that is still 6.x

jackpoz:
when in doubt not @ me

PR #16774, opened 2016-03-13
[Feature] Implement passive anticheat system

ariel-:
**Changes proposed**: Original patch taken from here: (created by @maanuel)
https://community.trinitycore.org/topic/10-patch-telenpc2-pvpranks-passive-anticheat-for-trinitycore-2016-02-24-4e4b2b9-35a-only/
- This adds the core capabilities to passively detect player hacking/cheating using 3rd party tools, it just scans for some flags/speed mismatch and emits a warning to every GM (or rather any account with rbac::RBAC_PERM_RECEIVE_GLOBAL_GM_TEXTMESSAGE)
- The original version was massively updated, it was optimized and updated to reflect core changes along the way.

**Target branch(es)**: 335/6x

**Issues addressed**: None, feature request.

**Tests performed**: (Does it build, tested in-game, etc) It builds and correctly detects many kinds of 3rd party tools, issuing a warning.

- [x] Move anticheat.sql to correct folder and update base sql
- [x] Move singleton out of header
- [ ] Re evaluate logic of existing checks (specially non-efficient ones)


sql/updates/world/2016_mm_dd_nn_world.sql
```diff
@@ -0,0 +1,28 @@
...
+(@ENTRY+4,  'Information about player %s'),
+(@ENTRY+5,  'Average: %f || Total Reports: %u'),
+(@ENTRY+6,  'Speed Reports: %u || Fly Reports: %u || Jump Reports: %u'),
+(@ENTRY+7,  'Walk on Water Reports: %u || Teleport to Plane Reports: %u'),
+(@ENTRY+8,  'Climb Reports: %u'),
+(@ENTRY+9,  '|cFFFFFC00[AC] |cffffffff|Hplayer:%s|h[%s]|h|r|cFF00FFFF Possible cheater!|r'),
+(@ENTRY+10, '============================='),
+(@ENTRY+11, 'Players with the lowest averages:'),
+(@ENTRY+12, 'Player: %s Average: %f Total Reports: %u'),
+(@ENTRY+13, 'Players with the more reports:'),
```
myran2: Players with the most reports?

ariel-: Heh, that's what you get for brainless copy pasting hardcoded texts xD

Nelidon:
You should rebase your branch and squash commits.

ariel-:
When it's deemed for merge into the core, that is.

Until then I prefer to leave the commit history intact so changes may be seen step by step (also different authors on commits)

sql/base/characters_database.sql
```diff
@@ -1,8 +1,8 @@
--- MySQL dump 10.13  Distrib 5.6.26, for Win64 (x86_64)
```
Shauren: I see many unrelated changes in the full sql dump - charsets/unsigned on some fields

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,482 @@
...
+{
+    if (!sWorld->getBoolConfig(CONFIG_ANTICHEAT_ENABLE))
+        return;
+
+    // Disable anticheat for Game Masters
+    if (player->CanBeGameMaster())
+        return;
+
+    ObjectGuid::LowType lowGuid = player->GetGUID().GetCounter();
+    AnticheatData& plData = m_playerMap[lowGuid];
```
Rochet2: I dont think this is safe. `StartHackDetection` can be called from multiple threads at the same time and access to `m_playerMap` is not guarded by anything from multiple accesses and modifications at the same time. There may be other thread safety related issues also.

Shauren: Looks like it is safe to me entry in m_playerMap is created on login (threadsafe context) and destroyed on logout (threadsafe context)
Modifying a single map value is not unsafe for m_playerMap itself but can be unsafe for the value - however this is not a problem either because only one thread can process movement packets for one player.

src/server/game/Anticheat/AnticheatMgr.h
```diff
@@ -0,0 +1,153 @@
...
+    SPEED_HACK_DETECTION            = 0x01,
+    FLY_HACK_DETECTION              = 0x02,
+    WALK_WATER_HACK_DETECTION       = 0x04,
+    JUMP_HACK_DETECTION             = 0x08,
+    TELEPORT_PLANE_HACK_DETECTION   = 0x10,
+    CLIMB_HACK_DETECTION            = 0x20
+};
+
+// GUIDLow is the key.
+typedef std::unordered_map<ObjectGuid::LowType, AnticheatData> AnticheatPlayersDataMap;
```
Shauren: Please do not use lowguid as key, it should only be used in database (I know there are MANY uses for it still in the code but don't add new ones)

Keader:
The Hitchhiker's Hack  is detected?

src/server/game/Anticheat/AnticheatMgr.cpp
Rochet2: It would be better probably to change the `[]` operator to `.at`. This way we can detect in case unintended behavior occurs. And it looks safe as you say, however the code also uses `m_playerMap.clear();` in one of the commands, which in my mind makes it unsafe since all data is deleted even though ppl are online and get detected.

sql/updates/characters/2016_03_13_00_characters.sql
```diff
@@ -0,0 +1,29 @@
...
+  `guid` int(10) unsigned NOT NULL DEFAULT '0',
+  `creation_time` int(10) unsigned NOT NULL DEFAULT '0',
+  `average` float NOT NULL DEFAULT '0',
+  `total_reports` bigint(20) unsigned NOT NULL DEFAULT '0',
+  `speed_reports` bigint(20) unsigned NOT NULL DEFAULT '0',
+  `fly_reports` bigint(20) unsigned NOT NULL DEFAULT '0',
+  `jump_reports` bigint(20) unsigned NOT NULL DEFAULT '0',
+  `waterwalk_reports` bigint(20) unsigned NOT NULL DEFAULT '0',
+  `teleportplane_reports` bigint(20) unsigned NOT NULL DEFAULT '0',
+  `climb_reports` bigint(20) unsigned NOT NULL DEFAULT '0',
```
Rochet2: Seems like the bigint is unnecessary as data is written in uint32, which is unsigned int. And data is never loaded, or I cant see that.

jackpoz:
I would like this patch to be tested with valgrind, helgrind and a static analyzer of your choice

src/server/scripts/Commands/cs_anticheat.cpp
```diff
@@ -0,0 +1,264 @@
...
+        handler->PSendSysMessage(LANG_COMMAND_ACPLAYER_CLIMB, climb_reports);
+        return true;
+    }
+
+    static bool HandleAntiCheatHandleCommand(ChatHandler* handler, char const* args)
+    {
+        if (!*args)
+            return false;
+
+        std::string strCommand = const_cast<char*>(args);
```
Rochet2: doesnt look like a necessary cast. The value should be copied by the string constructor.

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -345,6 +346,9 @@ void WorldSession::HandleMovementOpcodes(WorldPacket& recvData)
         plrMover->SetInWater(!plrMover->IsInWater() || plrMover->GetBaseMap()->IsUnderWater(movementInfo.pos.GetPositionX(), movementInfo.pos.GetPositionY(), movementInfo.pos.GetPositionZ()));
     }
 
+    if (plrMover)
+        sAnticheatMgr->StartHackDetection(plrMover, movementInfo, Opcodes(opcode));
```
Rochet2: The Opcodes cast looks unnecessary as opcode is already `uint16` and `StartHackDetection` takes an `uint16`

src/server/game/Anticheat/AnticheatMgr.cpp
Shauren: Oh, the existence of command changes this completely - indeed it becomes unsafe

src/server/game/Anticheat/AnticheatData.h
```diff
@@ -0,0 +1,89 @@
...
+
+    inline void SetLastReportTimer(uint32 time) { lastRepTime = time; }
+    inline uint32 GetLastReportTimer() const { return lastRepTime; }
+
+    inline void SetDailyReportState(bool b) { hasDailyReport = b; }
+    inline bool GetDailyReportState() const { return hasDailyReport; }
+
+private:
+    uint16 lastOpcode;
+    MovementInfo lastMovementInfo;
```
joschiwald: this is the same like `player->GetMovementInfo()`

sql/updates/characters/2016_03_13_00_characters.sql
ariel-: Data is queried when using .anticheat global command, however I'll have to rethink if this is necessary at all.

ariel-:
Well, I'm open to suggestions, because I don't know about multi threaded programming. To solve the unsafe issue, then I don't have to clear the map when the command is issued right? Can it be fixed by 0'ing out the associated data instead of clearing the whole map?

And, lastMovementInfo is saved to compare with new movement, see https://github.com/TrinityCore/TrinityCore/pull/16774/files#diff-cd67866d024cb7e5288fca94a3cd3358R172

joschiwald:
`player->GetMovementInfo()` contains the same data as lastMovementInfo, so you don't need to save it twice

Shauren:
Instead of trying to make access to m_playerMap from multiple threads safe (and slow!) you should just drop the delete command entirely

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -16,6 +16,7 @@
  * with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "AnticheatMgr.h"
```
Shauren: there is no change in this file - include not needed

src/server/scripts/Commands/cs_anticheat.cpp
```diff
@@ -0,0 +1,210 @@
...
+        {
+            handler->SendSysMessage(LANG_COMMAND_ACDISABLED);
+            handler->SetSentErrorMessage(true);
+            return false;
+        }
+
+        static WorldLocation const jailPos(1, 16226.5f, 16403.6f, -64.5f, 3.2f);
+
+        Player* pTarget = nullptr;
+        char* command = strtok(const_cast<char*>(args), " ");
```
Ascathor: const_cast is overkill here, simply replace it with a strtok(*args, " ");
That's how commands were handled until now.

Rochet2: Not quite sure what you mean. Const cast isn't an overhead (probably optimized away as it is only typesafety). Also before it was still casted with a C style cast: `strtok((char*)args, " ");`
https://github.com/TrinityCore/TrinityCore/blob/478cc756eb05591f36857be3ffd1895684529710/src/server/scripts/Commands/cs_character.cpp#L956

Ascathor: Ah, yeah. I forgot the actual casting that still needs to happen. But that's the case, we do not need an actual const_cast type of convertion, a simple c-style cast is how we handled it until now. It solves the issues.

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,447 @@
...
+        TC_LOG_TRACE("entities.player.character", "AnticheatMgr:: Speed-Hack detected on player %s (%s)", player->GetName().c_str(), player->GetGUID().ToString().c_str());
+        return true;
+    }
+
+    return false;
+}
+
+bool FlyHackCheck::OnCheck(Player* player, MovementInfo const& /* movementInfo */, uint16 /*opcode = 0*/) const
+{
+    if (AnticheatData* playerData = GetPlayerData(player->GetGUID()))
```
Ascathor: Travis complains here. Unused variable

src/server/scripts/Commands/cs_anticheat.cpp
ariel-: How it was in the past != how it should be done.

I need to remove const-ness because strtok requires a char*.

Ascathor: Are those commands different to the way other commands are handled ?

ariel-: Not really, as Rochet2 pointed out in that link, const-ness is casted away as it is here (except the cast there is not explicit, rather a C-style cast, this is why I prefer const_cast, because it's explicit of my intention)

src/server/game/Anticheat/AnticheatMgr.h
```diff
@@ -0,0 +1,151 @@
...
+{
+    friend class AnticheatCheck;
+
+    AnticheatMgr();
+    ~AnticheatMgr();
+
+    public:
+        static AnticheatMgr* instance()
+        {
+           static AnticheatMgr instance;
```
jackpoz: please keep an eye on TC development and apply the latest standard, for example https://github.com/TrinityCore/TrinityCore/commit/b7f936891aeb86eeb4eb67caa78e672858aa1e20

src/server/game/Anticheat/AnticheatMgr.h
```diff
@@ -0,0 +1,151 @@
...
+
+        // use sparingly, triggers a global anticheat save to DB, requires to lock player map and does some synchronous queries
+        bool AnticheatGlobalCommand(ChatHandler* handler);
+
+        void ResetDailyReportStates();
+
+        void LoadSettings();
+
+    private:
+        AnticheatData* GetPlayerData(ObjectGuid const& guid)
```
jackpoz: is it safe to return a pointer to a inner container element ?

Shauren: Sure it is - ObjectMgr does that all the time and here its a private method so it shouldn't be a problem (unless its used in an evil way and exposed to code outside of this class)

src/server/game/Anticheat/AnticheatMgr.h
```diff
@@ -0,0 +1,151 @@
...
+class AnticheatCheck
+{
+    public:
+        virtual ~AnticheatCheck() { }
+        virtual bool OnCheck(Player* player, MovementInfo const& movementInfo, uint16 opcode = 0) const = 0;
+
+        virtual void HackReport(Player* player) const = 0;
+
+    protected:
+        static AnticheatData* GetPlayerData(ObjectGuid const& guid) { return sAnticheatMgr->GetPlayerData(guid); }
```
jackpoz: @Shauren sAnticheatMgr->GetPlayerData(guid) exposed and used in places like this one ?

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,453 @@
...
+bool SpeedHackCheck::OnCheck(Player* player, MovementInfo const& movementInfo, uint16 /*opcode = 0*/) const
+{
+    AnticheatData* playerData = GetPlayerData(player->GetGUID());
+    if (!playerData)
+        return false;
+
+    // We also must check the map because the movementFlag can be modified by the client.
+    // If we just check the flag, they could always add that flag and always skip the speed hacking detection.
+    // 369 == DEEPRUN TRAM
+    if (player->m_movementInfo.HasMovementFlag(MOVEMENTFLAG_ONTRANSPORT) && player->GetMapId() == 369)
```
Nawuko: Why exclude the deeprun tram here? If the gameobjects (tram) are correctly marked as Transports the flag will be applied automaticly. I don't like the idea of hardcoding exceptions like these.

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,453 @@
...
+
+    if (movementInfo.HasMovementFlag(MOVEMENTFLAG_FALLING))
+        return false;
+
+    //DEAD_FALLING was deprecated
+    //if (player->getDeathState() == DEAD_FALLING)
+    //    return;
+    float x, y, z;
+    player->GetPosition(x, y, z);
+    float ground_Z = player->GetMap()->GetHeight(x, y, z);
```
Nawuko: This will be triggered on every movement package (even if the player only changes facing) which can be many many GetHeight queries if the player for example only spins arround. Maybe some kind of throttling is a good idea. Something like this is quite expensive if someone has 2k+ Players and every movement package is checked.

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,453 @@
...
+        WorldPacket data;
+        std::string str = Trinity::StringFormat(sObjectMgr->GetTrinityStringForDBCLocale(LANG_AC_CHEAT_REPORT), player->GetName().c_str(), player->GetName().c_str());
+        ChatHandler::BuildChatPacket(data, CHAT_MSG_SYSTEM, LANG_UNIVERSAL, NULL, NULL, str);
+        sWorld->SendGlobalGMMessage(&data);
+    }
+}
+
+bool SpeedHackCheck::OnCheck(Player* player, MovementInfo const& movementInfo, uint16 /*opcode = 0*/) const
+{
+    AnticheatData* playerData = GetPlayerData(player->GetGUID());
```
Nawuko: I dont think this is needed anymore (was used for the last saved movement info which is now directly used from the Player class)

src/server/game/Anticheat/AnticheatMgr.cpp
ariel-: Well, the actual "hack checking" code is the only thing remnant from the original; any suggestions here, besides throttle? Maybe caching the value elsewhere (is it good idea?)

src/server/game/Anticheat/AnticheatMgr.h
ariel-: It's only an accessor, used in checks context, checks are only called on movementInfo handlers, which in turn only get called when player is logged in, data is already existing at this point.

src/server/game/Anticheat/AnticheatMgr.cpp
Shauren: The problem with this is that only deeprun tram is NOT a Transport* serverside, it is the same gameobject type as elevators what means it does not maintain a passenger list

Nawuko: @Shauren It's marked as GAMEOBJECT_TYPE_TRANSPORT in the database so the client automaticly sends MOVEMENTFLAG_ONTRANSPORT. Atleast on 3.3.5. I just checked the sended movement package from the client on the latest rev.

Shauren: Yes but only MO_TRANSPORT are Transport* - this kind of check/exclusion is useless if you cant verify if the player really is on the tram or not (the cheater would just need to send  ONTRANSPORT flag with any random transport gameobject guid present on map and he would be able to speedhack all he wants)

src/server/game/Anticheat/AnticheatMgr.h
Rochet2: Assumptions. Assumptions lead to crashes. (if not now, then later)
Im sure jackpoz would feel a lot safer if it returned a copy or the pointer was wrapped in `std::unique_ptr`, right?

src/server/game/Anticheat/AnticheatMgr.cpp
Nawuko: Yes you are totally right. I tested 5 different cheats and one actually does exactly this.

Nawuko:
I just tested some random cheats for 3.3.5

**TeleportPlaneCheck** does in 4/5 cases not work. (Player has falling flag).

**SpeedhackCheck** dosent work in 3/5 cases (Player has OnTransportFlag or the cheat skips/obscure packages, send many many small packages or delays packages).

**WalkOnWaterCheck** fails in 5/5 cases (all cheats dosent send the waterwalk movementflag but are still able to walk on water)

src/server/game/Anticheat/AnticheatMgr.h
ariel-: Wait, what assumption? Return value is always checked against nullptr, I see no reason this could crash.

jackpoz: > Return value is always checked against nullptr, I see no reason this could crash.

null dereference is not really the only C++ kind of crash, especially when returning a pointer to a inner container element that can be deleted at anytime simply by calling container.clear() .

@Rochet2: I don't like any kind of std::cheat_ptr , it's still worth to note in PR review which are the points of attention, also to evaluate the quality and safety of code by reading the author replies.

Shauren: How about you remove this method entirely and just pass AnticheatData* as parameter to OnCheck?

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,453 @@
...
+    stmt->setUInt32(++index, playerData.GetCreationTime());
+    trans->Append(stmt);
+}
+
+bool AnticheatMgr::AnticheatGlobalCommand(ChatHandler* handler)
+{
+    SQLTransaction trans = CharacterDatabase.BeginTransaction();
+    // Mutex protected part
+    {
+        boost::shared_lock<boost::shared_mutex> lock(*HashMapHolder<Player>::GetLock());
```
jackpoz: @Shauren : is the lock needed here ?

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,453 @@
...
+
+        HashMapHolder<Player>::MapType const& m = ObjectAccessor::GetPlayers();
+        for (auto const& pair : m)
+            SavePlayerData(pair.second->GetGUID(), trans);
+    }
+    CharacterDatabase.CommitTransaction(trans);
+
+    QueryResult resultDB = CharacterDatabase.Query("SELECT guid,average,total_reports FROM players_reports_status WHERE total_reports > 0 ORDER BY average ASC LIMIT 5");
```
jackpoz: are the CommitTransaction() and Query() both sync ? otherwise you have a database race condition

src/server/game/Anticheat/AnticheatMgr.h
```diff
@@ -0,0 +1,151 @@
...
+    private:
+        AnticheatData* GetPlayerData(ObjectGuid const& guid)
+        {
+            auto itr = m_playerMap.find(guid);
+            if (itr != m_playerMap.end())
+                return &itr->second;
+            return nullptr;
+        }
+
+        AnticheatPlayersDataMap m_playerMap;
```
jackpoz: I always wonder if we should keep maps using Player guid as key or just store the data in Player class itself

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,453 @@
...
+    AnticheatData& plData = m_playerMap[player->GetGUID()];
+
+    if (result)
+        plData.SetDailyReportState(true);
+}
+
+void AnticheatMgr::HandlePlayerLogout(ObjectGuid const& guid)
+{
+    // We must also delete it at logout to prevent have data of offline players in the db when we query the database (IE: The GM Command)
+    PreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_DEL_AC_PLAYER_REPORTS);
```
jackpoz: is there a way to avoid executing an additional query on logout for every player ?

src/server/worldserver/worldserver.conf.dist
```diff
@@ -1321,6 +1322,55 @@ Warden.BanDuration = 86400
...
 
 ###################################################################################################
+# ANTICHEAT SETTINGS
+#
+#     Anticheat.Enable
+#        Description: Enables or disables the Anticheat System functionality
+#        Default:     1 - (Enabled)
+#                     0 - (Disabled)
+
+Anticheat.Enable = 1
```
jackpoz: Warden is disabled by default so it makes sense to keep this disabled by default too.

src/server/scripts/Commands/cs_anticheat.cpp
```diff
@@ -0,0 +1,212 @@
...
+            handler->SendSysMessage(LANG_PLAYER_NOT_FOUND);
+            handler->SetSentErrorMessage(true);
+            return false;
+        }
+
+        ChatHandler(pTarget->GetSession()).SendSysMessage(LANG_COMMAND_ACWARN);
+        return true;
+    }
+
+    static bool HandleAnticheatJailCommand(ChatHandler* handler, char const* args)
```
jackpoz: this command doesn't check any AntiCheat related data, it's just a generic .jail command, yet it requires anticheat to be enabled. The "!sWorld->getBoolConfig(CONFIG_ANTICHEAT_ENABLE)" check should be removed and the command should be moved to a more generic commands group, maybe the same of the freeze command (if it exists)

src/server/game/Anticheat/AnticheatMgr.cpp
Shauren: If running from World::Update thread? No.

src/server/game/Anticheat/AnticheatMgr.cpp
Shauren: CommitTransaction is async, Query is sync

src/server/game/Anticheat/AnticheatMgr.h
ariel-: clear() was removed entirely, and passing the pointer as a parameter sounds good.

src/server/game/Anticheat/AnticheatMgr.cpp
ariel-: Yes, however it'll lead to not-updated-enough data, maybe split this command in two: anticheat save and anticheat global?

src/server/game/Anticheat/AnticheatMgr.cpp
ariel-: I can leave the data in DB, and then do an async delete on login, anyways all data is wiped on server startup, so it can work.

src/server/game/Anticheat/AnticheatMgr.cpp
jackpoz: All these commands are evil enough to allow sync DirectCommitTransaction(...) which will wait for the transaction to be over before executing the Query(...)

src/server/game/Anticheat/AnticheatMgr.cpp
jackpoz: is this query executed even if the whole AntiCheat stuff is disabled ?

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,453 @@
...
+    for (AnticheatCheck* check : m_checks)
+        if (check->OnCheck(player, movementInfo, opcode))
+            check->HackReport(player);
+
+    plData.SetLastOpcode(opcode);
+}
+
+void AnticheatMgr::HandlePlayerLogin(Player* player, PreparedQueryResult result)
+{
+    AnticheatData& plData = m_playerMap[player->GetGUID()];
```
jackpoz: the whole idea of this line in case result is null is to create the AnticheatData element in the container ?

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,453 @@
...
+    if (!playerData)
+        return false;
+
+    // We also must check the map because the movementFlag can be modified by the client.
+    // If we just check the flag, they could always add that flag and always skip the speed hacking detection.
+    // 369 == DEEPRUN TRAM
+    if (player->m_movementInfo.HasMovementFlag(MOVEMENTFLAG_ONTRANSPORT) && player->GetMapId() == 369)
+        return false;
+
+    uint32 distance2D = (uint32)movementInfo.pos.GetExactDist2d(&player->m_movementInfo.pos);
```
Runico: while player have flyable aura ( like fly mounts ) player could easily bypass this, you should check 3D distance while player have ability to fly

How to reproduce
1. mount up ( fly )
2. change your speed 20 ( by cheats tools )
3. fly as 3D ( easiest way move by mouse or press both w and q )

src/server/game/Anticheat/AnticheatMgr.cpp
ariel-: Yes, indeed.

ryoting:
this can check player`s flyhack?
ago,this anticheat check gm`s fly only,can`t check player`s fly

Demonid:
@ryoting This is true, at least the last time i checked AC2, it only checks Flyhack for accouts with GM Level > 0

Dont know if this is the case for this PR, but if someone could give it a try :+1:

Aokromes:
This is not based on ac2.

Demonid:
@Aokromes Wouldn't hurt to check just in case.

Then again, what about adding some kind of movement restriction, kinda like this Anticheat does:

https://bitbucket.org/lordpsyan/lordpsyan-patches/src/bd5c2b08de445bafbb4cc70cba56bc391e5f5848/Individual/Anti-Cheats/2016_02_29-Non-Passive-AC2.patch?at=master&fileviewer=file-view-default

Just a suggestion, nothing more.

mthsena:
> @ryoting This is true, at least the last time i checked AC2, it only detects Flyhack for accounts with GM Level > 0
> 
> Dont know if this is the case for this PR, but if someone could give it a try :+1:

@Demonid, @ryoting  Related to WorldSession.cpp. On my version, I removed these lines and it worked fine for accounts 0.

```
REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY) && GetSecurity() == SEC_PLAYER &&
!GetPlayer()->m_mover->HasAuraType(SPELL_AURA_FLY) &&
!GetPlayer()->m_mover-HasAuraType(SPELL_AURA_MOD_INCREASE_MOUNTED_FLIGHT_SPEED),
MOVEMENTFLAG_FLYING | MOVEMENTFLAG_CAN_FLY);
```

Aokromes:
@Demonid this is intended to be PASSIVE anticheat.

Nawuko:
I dont want to disrespect your work. But maybe we should implement something like this: https://github.com/Nawuko/TrinityCore/commit/88d662f7b867f804629cbe1dbfb8fd2988de81d0
And not an old, only partly working code. Just an thought. Like i said i tested this anti cheat with the most popular 5 cheats out there and the result is not very great.

Aokromes:
@nawuko that code is under consideration.

tripleslash:
@Nawuko This is not supposed to be merged whatsoever. It is my personal testing playground. I plan to rewrite the entire movement system from scratch in a blizzlike way during the next semester holidays.

ryoting:
any news of this?

dandi2001:
Is this heading anywhere? D:

src/server/game/Anticheat/AnticheatMgr.cpp
maanuel: Hi everyone.

It has been long time since I wrote the original AC code. I can't really recall the idea behind that check.

I think that I could be trying not to divide by zero (in a trigonometric function?) in the GetHeight function, I don't know if that could really happen.
Another reason could be that I detected something in the  WoWEmuHacker's (for 3.3.5) behaviour.

Independently of the reasons above explained, I think that the check is not correct. GetPositionZ is a float number, so it must be checked against an epsilon.

About the performance concerns, at that moment I did not understand the TC threading model. Now I understand that anticheat checks are made in the same thread that is processing packets, so it is not a nice idea to perform heavy computations. 

I think that the options available to mitigate this drawback, if the anticheat checks are made in the same thread, are the followings:

Randomize this check, I mean it could happen with certain probability. If you are detected positive, your probability increases otherwise decreases.
Inspecting the GetHeight() function I noticed that there are two metrics used: maps data and vmaps data. We could consider just using maps data, in case it were faster than vmap computations (I don't really know about the peformance costs of each).

Finally, we could consider doing all this checks in a separated thread, in case that map or vmap height were thread safe. @Shauren Could you enlighten us about this? 

If I am not wrong vmap height is thread-safe unless you set isHeightCalcEnabled as false, but it is just set at startup.

Palabola:
Instead of try to create the perfect movement calculation on server side and looking after mismatch, my advise focus to catch Hack Engines poor designed packets and logics.

We need calculation only for speed and Z movement.

**Example:**
Some Hack Engine only spam 2 types of Movement packet(Fall & Stop Fall) with always the same flags.

tripleslash:
Then I'll go ahead and implement one that circumvents your suggestion. For an open source emulator a proper anticheat is necessary and not something like this.

Ghost-R:
@tripleslash any news ?

Aokromes:
@ariel- you think this can be merged? it can be enhanced  on the future.

src/server/game/Anticheat/AnticheatMgr.cpp
```diff
@@ -0,0 +1,443 @@
...
+
+    // in this case we don't care if they are "legal" flags, they are handled in another parts of the Anticheat Manager.
+    if (player->IsInWater() ||
+        player->IsFlying() ||
+        player->IsFalling())
+        return false;
+
+    Position playerPos = player->GetPosition();
+
+    float deltaZ = fabsf(playerPos.GetPositionZ() - movementInfo.pos.GetPositionZ());
```
jackpoz: please avoid using C functions and use the C++ std::* counterpart instead

src/server/game/Anticheat/AnticheatMgr.cpp
jackpoz: Checking only half of height data (only maps and not vmaps) makes no sense since you have partial informations.
Since this is just an optional feature, you could explain in the configs about the performance drawbacks that these check will do and then it's up to the server owner to choose if it's worth to enable them or not.

src/server/game/Anticheat/AnticheatMgr.cpp
jackpoz: ping on my questions above

src/server/game/Anticheat/AnticheatMgr.cpp
jackpoz: ping on the comment above about the lock not being needed (if running from World::Update thread: is it ? )

src/server/game/World/World.cpp
```diff
@@ -1289,6 +1290,13 @@ void World::LoadConfigSettings(bool reload)
     m_bool_configs[CONFIG_PDUMP_NO_OVERWRITE] = sConfigMgr->GetBoolDefault("PlayerDump.DisallowOverwrite", true);
     m_bool_configs[CONFIG_UI_QUESTLEVELS_IN_DIALOGS] = sConfigMgr->GetBoolDefault("UI.ShowQuestLevelsInDialogs", false);
 
+    // Passive anticheat
+    m_bool_configs[CONFIG_ANTICHEAT_ENABLE] = sConfigMgr->GetBoolDefault("Anticheat.Enable", true);
```
jackpoz: why is it enabled by default ?

Demonid:
@Aokromes  Needs update after https://github.com/TrinityCore/TrinityCore/commit/7026c7ed7a39ae258eb0ba01f6946c8bc1594ce0

tripleslash:
I'd be sad to see this thing merged into the official TC master...
I'd suggest abandoning this branch and starting to fix the fundamental issues (like movement ACK handling) first. As soon as that's done one can implement the missing handlers for the movement packets and stop handling every different opcode in one handler function and start doing proper timestamp synchronization.

Nawuko:
@tripleslash :+1: 

Also TrinityCore gets more and more cluttered (AHBot, Custom Duel stuff, Custom PvP Token stuff) now a half-assed AntiCheat that accomplish nothing in the most cases.

Aokromes:
@Nawuko trinitycore was made with the mind of add custom features since the start.

jackpoz:
> cluttered

?

Demonid:
Could anyone check if this PR is still working?

I tried after the last change but it doesn't detect anything, maybe the detection system broke after all the code changes?

Thanks!

Keader:
i cant see problems with custom stuffs. Using "disabled" by default, why not be present?

Demonid:
Could anyone check if this PR is still working?

I tried testing it after the last commit 2fb0418 but it doesn't detect anything, maybe the detection system broke after all the code changes?

Thanks!

Mrgrasser:
Break passive ? not work now :(

Mrgrasser:
@Demonid  this AntiCheat have some problem for back jump , Improved Charge warrior , spell Engineering , Some time speed rouge , Aspect of the Pack hunter , Blazing Speed mage 
If the configuration is correct is with no trouble AntiCheat.

> Then again, what about adding some kind of movement restriction, kinda like this Anticheat does:
>
> https://bitbucket.org/lordpsyan/lordpsyan-patches/src/bd5c2b08de445bafbb4cc70cba56bc391e5f5848/Individual/Anti-Cheats/2016_02_29-Non-Passive-AC2.patch?at=master&fileviewer=file-view-default
>
>Just a suggestion, nothing more.

chaodhib:
I don't think this should get merged neither to be honest. I think we should keep the good stuff (which is the anti cheat logic, basically every OnCheck method in AnticheatMgr.cpp) and throw the rest out. 

The anti cheat logic will already use additional CPU resources. I don't think adding a system that tracks cheating players and does database queries on top of that is a good idea. Also, keeping small the anti movement cheat code will make it easier to maintain and improve. Imo, we should just add a hook in the MovementHandler: It will call a method like `bool verifyClientPosition()` that takes as input the latest verified state of the player and the latest received. If it returns false, the handler should ignore the last packet, update the client to a valid position and potentially do one (or more) of the following options (configurable by config file): log the event, kick the player and or ban the player. This change won't take much more than 300 LOC.

I think the focus should be put on the validating incoming packets (example: if the unit is already falling, ignore any new MSG_MOVE_JUMP, etc...), not all the extra stuff around it.

PR #16786, opened 2016-03-14
Core/AI Remove Reset() call from JustRespawned() hook

### LINKED ISSUE #16672: [3.3.5][6.x] Core/Creatures: Respawned creatures call for Reset() twice
**Description**: Reset() hook runs twice at creature Respawn, 

On Creature.cpp
AI()->Reset() is called at https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Creature/Creature.cpp#L1718

in the next update tick, AI()->JustRespawned(); is called at https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Creature/Creature.cpp#L496.

Note JustRespawned definition : 
```cpp
virtual void JustRespawned() { Reset(); }
```
found at https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/CreatureAI.h#L140 

Result, reset called twice.
I suggest remove the Reset() from Justrespawned(), seems not to be needed

**Current behaviour**: if the creature has a hook for Reset(), actions on it will run twice at creature's respawn

**Expected behaviour**: Reset should run one time at respawn

1. Script a creture, put something on Reset() hook
2. despawn and respawn the creature
3. Reset will run twice.

**Branch(es)**: 335/6x

**TC hash/commit**: dab661c8dad6

**TDB version**: 335.60

sirikfoll:
**Changes proposed**:
- With the current implementation Reset() hook is called twice at creature respawn
- So, I propose to remove it from JustRespawned()
- Why this shouldn't break anything:
- Reset is called at creature spawn in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/CoreAI/UnitAI.h#L130
- Reset is called at creature respawn in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Creature/Creature.cpp#L1718

**Target branch(es)**: 335/6x

**Issues addressed**: Fixes #16672

**Tests performed**: Built and tested

- Didn't found any so far

### DIFF
--- a/src/server/game/AI/CreatureAI.h
+++ b/src/server/game/AI/CreatureAI.h
@@ -137,8 +137,8 @@ class CreatureAI : public UnitAI
         virtual void AttackedBy(Unit* /*attacker*/) { }
         virtual bool IsEscorted() const { return false; }
 
-        // Called when creature is spawned or respawned (for reseting variables)
-        virtual void JustRespawned() { Reset(); }
+        // Called when creature is spawned or respawned
+        virtual void JustRespawned() { }
 
         // Called at waypoint reached or point movement finished
         virtual void MovementInform(uint32 /*type*/, uint32 /*id*/) { }


ccrs:
check all the JustRespawned overrides, if there are no conflicts, this should be ok.

jackpoz:
> Tests performed: Built and tested

what is "tested" supposed to mean in this context ? did you test all the UnitAI scripts in current sources ?

sirikfoll:
like log in and test if it solved the main problem and debug a spawn and respawn sittuation to check if the call is happening yet.
I just can't think in any way this can break anything.
The only way Reset will not be called is if we override InitializeAI() and only for the spawn sittuation, and this behavior exists without this PR.
Reset will still be called at respawn by https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Creature/Creature.cpp#L1718 regardless if I override InitilizeAI() or JustRespawned() or anything else, as it already happens without this PR too.
The only change is that it will be callled once, instead of twice.

PR #14865, opened 2015-06-10
Scripts/Naxxramas: Rework of the Gluth encounter

chaodhib:
Fixes of the follow bugs:

1) Gluth should eat the zombies close to him (https://github.com/TrinityCore/TrinityCore/issues/14806). The animation used now is not 100% blizzlike but it's an improvement nonetheless.
2) The zombies should put the "Infected Wound" debuff on their auto-attack targets.
3) Zombies should have a normal threat table before decimate and should loose all kind of aggro behavior and run toward Gluth after decimate.
4) Gluth should be not be affected by the decimate damage.
5) Newly poped zombies should have all players in their aggro list as soon as they spawn. So far, they were not moving and were only aggroed by proximity.
6) and several minor fixes (emotes, useless data in db, ...).

PR #16806, opened 2016-03-19
Adding a helper/wrapper method in MotionMaster to move toward a unit …

chaodhib:
Adding a helper/wrapper method in MotionMaster as discussed with @r00ty-tc . It adds a missing functionality: move closer to a unit until we reach a certain distance and then stop. I need it for Gluth and I suppose it could be useful elsewhere in the scripts.

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -302,6 +302,27 @@ void MotionMaster::MovePoint(uint32 id, float x, float y, float z, bool generate
     }
 }
 
+void MotionMaster::MoveCloserAndStop(uint32 id, Unit* target, float distance)
+{
+    float distanceToTravel = _owner->GetExactDist2d(target) - distance;
+    if (distanceToTravel > 0.0f)
+    {
+        float angle = _owner->GetAngle(target);
+        float destx = _owner->GetPositionX() + distanceToTravel * std::cos(angle);
+        float desty = _owner->GetPositionY() + distanceToTravel * std::sin(angle);
+        MovePoint(id, destx, desty, target->GetPositionZ());
+    }
+    else 
+    {
+        // we are already close enough. We just need to turn toward the target without changing position.
+        Movement::MoveSplineInit init(_owner);
+        init.MoveTo(_owner->GetPositionX(), _owner->GetPositionY(), _owner->GetPositionZMinusOffset());
+        init.SetFacing(target);
+        init.Launch();
+        Mutate(new EffectMovementGenerator(id), MOTION_SLOT_ACTIVE);
+    }
+}
+
 void MotionMaster::MoveLand(uint32 id, Position const& pos)
 {
     float x, y, z;
--- a/src/server/game/Movement/MotionMaster.h
+++ b/src/server/game/Movement/MotionMaster.h
@@ -173,6 +173,12 @@ class MotionMaster //: private std::stack<MovementGenerator *>
             { MovePoint(id, pos.m_positionX, pos.m_positionY, pos.m_positionZ, generatePath); }
         void MovePoint(uint32 id, float x, float y, float z, bool generatePath = true);
 
+        /*  Makes the unit move toward the target until it is at a certain distance from it. The unit then stops. 
+            Only works in 2D.
+            This method doesn't account for any movement done by the target. in other words, it only works if the target is stationary.
+        */
+        void MoveCloserAndStop(uint32 id, Unit* target, float distance);
+
         // These two movement types should only be used with creatures having landing/takeoff animations
         void MoveLand(uint32 id, Position const& pos);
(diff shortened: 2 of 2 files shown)


src/server/game/Movement/MotionMaster.cpp
```diff
@@ -302,6 +302,23 @@ void MotionMaster::MovePoint(uint32 id, float x, float y, float z, bool generate
...
+    {
+        float angle = _owner->GetAngle(target);
+        float destx = _owner->GetPositionX() + distanceToTravel * std::cos(angle);
+        float desty = _owner->GetPositionY() + distanceToTravel * std::sin(angle);
+        MovePoint(id, destx, desty, target->GetPositionZ());
+    }
+    else 
+    {
+        // we are already close enough. We just need to turn toward the target without changing position.
+        _owner->SetFacingToObject(target);
```
chaodhib: I'm thinking about removing this part. Looks potentially buggy.

Naios: What happens when the target moves meanwhile (in or against the direction the target is moving)?

Pesthuf:
Doesn't ChaseMovementGenerator already have a method to chase at a specific angle and distance?

It was quite broken if I remember right, because it makes the creature chase at a fixed angle and makes the creature run away if it's less than the distance away from the target, but I've never seen this behavior on Blizzard.

Maybe this function should be fixed instead?

chaodhib:
@Naios 
I don't have time to test it but this new line should prevent any issue.

src/server/game/Movement/MotionMaster.cpp
chaodhib: I misunderstood your question. Good point. This method actually assumes that the target is not moving. I should mention this in the comments.

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -302,6 +302,24 @@ void MotionMaster::MovePoint(uint32 id, float x, float y, float z, bool generate
     }
 }
 
+void MotionMaster::MoveCloserAndStop(uint32 id, Unit* target, float distance)
+{
+    float distanceToTravel = _owner->GetExactDist2d(target) - distance;
+    if (distanceToTravel > 0.0)
```
jackpoz: 0.0 is double but distanceToTravel is float

Naios:
My opinion is that it's much better to implement this as a movement generator instead of a oneshot method because you can't cover moving targets with the current way.

Implementing this as a movement generator would bring the benefit that you can rely on that when the `OnPointReached` hook is called that the npc is really `distance` meters away from the target and not close to the point the target was 20s ago, which means we could save the work in `UpdateAI` to check periodically for the distance to the target.

chaodhib:
@Pesthuf @Naios 
Good points.

First of all, the way I see it, `MoveFollow()` and `MoveCloserAndStop()` do not serve the same purpose, even though they are some similarities. With `MoveFollow`, the unit should follow the target until the end of times (unless disturbed or if another method of `MotionMaster `is called). However, with `MoveCloserAndStop`, the unit should try reach the target and once it's done, "detach" from it.

On a related note, I noticed too that `MoveFollow` behaves weirdly sometimes and should probably be re-implemented: When the moving unit gets too close to the target, the moving unit backs off, which is unblizzlike. This should be reported into a bug issue, if it's not already done.

Second point: 
Naios, I agree with you that the restriction on this implementation (that it only works if the target is not moving) sucks. To be honest however, I don't want to extend the scope of this PR. I have too many open PRs and I want to focus on my unresolved ones. I need `MoveCloserAndStop` for https://github.com/TrinityCore/TrinityCore/pull/14865 and I do not need for it to handle a moving target. This method can always be re-implemented to alleviate the restriction later on by someone who needs it and it will stay compatible with cases like mine.

Shauren:
What you intend there (a stationary target) is fine - but write it in the comment on that method that it doesn't account for any movement done by target after calling it so others trying to use it will know

chaodhib:
@Shauren yep ok

chaodhib:
Seems finished to me. Only needs more testing.

Naios:
Are you sure you don't want to spend some more time in this PR and try to solve this through the MotionMaster (maybe through fixing the existing methods or adding a new one)?
The result would be much better and much more reliable.

I don't like the Idea of adding a new half working way to the codebase which just decorates the misbehaviour of the correct method.

chaodhib:
@Naios In another PR maybe. Like I said, I want to stay focused and I opened this PR because I need the code for another one of my PRs.

chaodhib:
Nevermind, I need an implementation that also works with a moving targets for another part of the same PR (the zombie movement after they are hit by decimate). I'll see what I can do...

chaodhib:
I changed my mind (again, sorry). Could this get merged? I need it for Gluth. I'll open a follow up PR later, to extend the possibilities of use of the method (if it's in my skill set).

PR #16810, opened 2016-03-20
Introducing the shared linking compilation model

Naios:
# Introducing the shared linking compilation

**This commits will make it possible to link all executable targets dynamically -> HotSwap #15671 precommits (3/4)**

I guess this is are the most intrusive changes which are required for the hotswap system.
Basically the commits enable the cmake `WITH_DYNAMIC_LINKING` option which enables shared compilation for TC's major libraries.

The `TRINITY_*_API` macros are required for the windows platform since MSVC doesn't auto export and import symbols.
The recent introduced cmake 3.4 auto export feature doesn't help here since we still need to import  all symbols.
I know the macros are ugly but we are forced to add it if we want automatic c++ script hot reloading to work.

**Target branch(es)**: 335/6x

**Issues addressed**: Fixes #

MSVC and Clang build should work

- [ ] We need to check weired behaviour with the shared linking (but this can also be done in the main branches - since it doesn't affect the static linking).


MitchesD:
isn't there any "cleaner" way instead of adding everywhere TRINITY_*_API etc. ? I read your description but still :D

jackpoz:
A macro before every class declaration is an annoying clutter that we'll have to deal with every time we define a new class. I'd prefer not to see anything like that in TC sources.

Naios:
Adding a new public class which is used by other libraries happens not that often,
the disadvantage compared to the advantage of reloadable scripts is very minimal.

Just to make it clear, it's not every class which needs a macro, a macro is only needed when you access the implementation of a class/function from another library which isn't fully contained in the header.

This also makes it possible to annotate TC's public API which will change rarely.

For example:

```
class MyClass
{
  void doSth() { /* ... */ }
};
```
Needs no macro.
* * *

```
class MyPrivateClass
{
  void doSth();
};
```
Needs no macro when it's used inside (privately) the library only.
* * *

```
class MyPublicClass
{
  void doSth();
};
```
Needs a macro when it's used outside the library.
* * *

src/common/Define.h
```diff
@@ -95,6 +95,51 @@
...
+#    define TRINITY_API_IMPORT
+#  else
+#    error compiler not supported!
+#  endif
+#else
+#  define TRINITY_API_EXPORT
+#  define TRINITY_API_IMPORT
+#endif
+
+#ifdef TRINITY_API_EXPORT_IPC
```
Shauren: IPC no longer exists

Naios: Yep thanks

Naios:
Merged in cd777f97a182c582221bf5194e4075d92c532679

jackpoz:
doesn't this mess up with the ongoing cherry-picks and TDB release ?

Naios:
No, I've checked that. Shauren cherry-picked every important core commit which means it was the perfect time to merge it.

Nelidon:
When compiling the core with dynamic linkings option it's impossible for Visual Studio (2015) to find out pdb files for dll even if we force the software to load it.
Only the worldserver's pdb is loaded.

PR #16819, opened 2016-03-22
Core/Pdump: Fix pdump write after 1925778e44

Yehonal:
...however we should consider to use field names instead of index to avoid future issues after table altered


tkrokli:
Looks good, although I wish you had put more description in the commit text & first post.

PR #16827, opened 2016-03-23
Tools/ConnectionPatcher: patch 'versions' by default to prevent auto-…

DJScias:
…updating of executables

**Changes proposed**:
- On executable drag and drop it will find the current build number and apply the correct versions patch.
- This prevents patched executables from auto-updating and causing incorrect version errors.
- Also removed "-e" functionality as this change renders it obsolete.

**Target branch(es)**: 6x

**Issues addressed**: Requested by Aokromes to stop people from having issues with their executables auto-updating and stopping to work and thus making many posts on the forums about it ;).

Cutting the ties to the actual versions file causes no issues regardless if people are on the latest or not latest version, it only makes things easier.

**Tests performed**: (Does it build, tested in-game, etc) It builds.

**Known issues and TODO list**: This should be final, unless someone finds issues with it that may have slipped through.

### DIFF
--- a/src/tools/connection_patcher/Program.cpp
+++ b/src/tools/connection_patcher/Program.cpp
@@ -96,7 +96,7 @@ namespace Connection_Patcher
         }
 
         template<typename PATCH, typename PATTERN>
-        void do_patches(Patcher* patcher, boost::filesystem::path output, bool patchVersionPath, uint32_t buildNumber)
+        void do_patches(Patcher* patcher, boost::filesystem::path output, uint32_t buildNumber)
         {
             std::cout << "patching Portal\n";
             // '.logon.battle.net' -> '' to allow for set portal 'host'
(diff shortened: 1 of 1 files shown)


Shauren:
Doesn't fully prevent accidental patching, battle.net app can still do it if running in background

DJScias:
@Shauren Of course there's always that, but this kicks out the option of it happening by people opening their executable when it's closed.

Also, some people also create a second WoW folder for TC needs (as I have) which battle.net does not patch and with this fix neither can the executables give version errors anymore either.

Kind regards,
DJScias

DDuarte:
Shouldn't the behavior be inverted? Specify a command flag to *not* prevent updates.

DJScias:
@DDuarte This is something you should ask @Aokromes for, as this behaviour is how he wanted it.

I'll try to break down what making it default behaviour does in regards to which build you're on:
- If you're on the latest build: This will patch the executable like normal and sever the versions file connection. This makes sure that if Blizzard patches their own versions file your executable won't stop working with 'incorrect version' errors. It also stops the client from starting an auto-update when that happens.
- If you're not on the latest build: This will patch the executable like normal and sever the versions file connection. However in this case if the files that you want to connect to are, let's say 6.1.2 as an example, it will allow you without giving you countless of errors as well. Eventually it also stops the client from starting an auto-update when that happens.

But I'd still say you should talk to Aokromes for a full explanation on why he wanted this behaviour.

Kind regards,
DJScias

DDuarte:
Dropping -e and making it default is fine as it is. All I am asking it to replace -e by something else that does not prevent patching (behavior before this PR).

DJScias:
Well @DDuarte 

As @Shauren explained. Opening up the battle.net app and allowing it to patch the normal way will still work and basically overwrite these changes again.. So would the -e functionality honestly still be needed?

Kind regards,
DJScias

DDuarte:
Probably not needed, fair enough.

Carbenium:
We only have to keep https://github.com/trinity6/trinity6.github.io updated
(this should only be a reminder, no con)

Nelidon:
In fact it's not enough because Blizzard can remove data files from his server and you won't be able to run the game again.
To avoid this you have to create a download server which contains all game data stored by Blizzard.

Warpten:
This has nothing to do with remote casc

lorac2k14:
I have a newb question 
So this patcher calls on https://github.com/trinity6/trinity6.github.io for the patching info now? vs blizzards patch info?
or is it fully disabled now?
Either way awesome work :)

PR #16834, opened 2016-03-24
[3.3.5] Entities/Unit: Finally fix no-path evasion

Treeston:
Entities/Unit: Finally fix no-path evasion
- When a creature cannot find a path to its victim, it begins evading all attacks and regenerating health.
- If this persists for 5 seconds, it evades back to spawn position with new EvadeReason value EVADE_REASON_NO_PATH.
- Also some SmartAI cleanup (why oh why does it have so much duplicated code) and getting rid of #defines in favor of type-checked compile-time constants.

PS: Sorry for fixing your favorite bug, you know who you are.

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -268,6 +268,7 @@ bool CreatureAI::_EnterEvadeMode(EvadeReason /*why*/)
     me->SetLootRecipient(NULL);
     me->ResetPlayerDamageReq();
     me->SetLastDamagedTime(0);
+    me->SetCannotReachTarget(false);
 
     if (me->IsInEvadeMode())
         return false;
@@ -275,11 +276,11 @@ bool CreatureAI::_EnterEvadeMode(EvadeReason /*why*/)
     return true;
 }
 
-#define BOUNDARY_VISUALIZE_CREATURE 15425
-#define BOUNDARY_VISUALIZE_CREATURE_SCALE 0.25f
-#define BOUNDARY_VISUALIZE_STEP_SIZE 1
-#define BOUNDARY_VISUALIZE_FAILSAFE_LIMIT 750
-#define BOUNDARY_VISUALIZE_SPAWN_HEIGHT 5
+const uint32 BOUNDARY_VISUALIZE_CREATURE = 15425;
+const float BOUNDARY_VISUALIZE_CREATURE_SCALE = 0.25f;
+const int8 BOUNDARY_VISUALIZE_STEP_SIZE = 1;
+const int32 BOUNDARY_VISUALIZE_FAILSAFE_LIMIT = 750;
+const float BOUNDARY_VISUALIZE_SPAWN_HEIGHT = 5.0f;
 int32 CreatureAI::VisualizeBoundary(uint32 duration, Unit* owner, bool fill) const
 {
     typedef std::pair<int32, int32> coordinate;
@@ -295,13 +296,13 @@ int32 CreatureAI::VisualizeBoundary(uint32 duration, Unit* owner, bool fill) con
     std::unordered_set<coordinate> outOfBounds;
 
     Position startPosition = owner->GetPosition();
-    if (!CheckBoundary(&startPosition)) // fall back to creature position
-    {
+    if (!CheckBoundary(&startPosition))
+    { // fall back to creature position
         startPosition = me->GetPosition();
         if (!CheckBoundary(&startPosition))
-        {
+        { // fall back to creature home position
             startPosition = me->GetHomePosition();
-            if (!CheckBoundary(&startPosition)) // fall back to creature home position
+            if (!CheckBoundary(&startPosition))
                 return LANG_CREATURE_NO_INTERIOR_POINT_FOUND;
         }
     }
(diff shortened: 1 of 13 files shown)


Ascathor:
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Globals/ObjectMgr.cpp:490:27: fatal error: 
      use of undeclared identifier 'CREATURE_MAX_SPELLS'
    for (uint8 i = 0; i < CREATURE_MAX_SPELLS; ++i)
                          ^

src/server/game/AI/CreatureAI.cpp
```diff
@@ -275,11 +276,11 @@ bool CreatureAI::_EnterEvadeMode(EvadeReason /*why*/)
...
-#define BOUNDARY_VISUALIZE_CREATURE 15425
-#define BOUNDARY_VISUALIZE_CREATURE_SCALE 0.25f
-#define BOUNDARY_VISUALIZE_STEP_SIZE 1
-#define BOUNDARY_VISUALIZE_FAILSAFE_LIMIT 750
-#define BOUNDARY_VISUALIZE_SPAWN_HEIGHT 5
+const uint32 BOUNDARY_VISUALIZE_CREATURE = 15425;
+const float BOUNDARY_VISUALIZE_CREATURE_SCALE = 0.25f;
+const int8 BOUNDARY_VISUALIZE_STEP_SIZE = 1;
+const int32 BOUNDARY_VISUALIZE_FAILSAFE_LIMIT = 750;
+const float BOUNDARY_VISUALIZE_SPAWN_HEIGHT = 5.0f;
```
soulfrost: static ?

Treeston: Irrelevant. I can add it if code style demands it. Dunno if we have a guideline for global constants.

ccrs:
- The evade shouldnt happen on instance: on retail npcs got stuck in evade mode till it reaches the target
- Should bosses use the invalid path evade or should they just ignore unreachable targets?
- 5 seconds seems like too short wait time

Nawuko:
> The evade shouldnt happen on instance: on retail npcs got stuck in evade mode till it reaches the target

This is not completely true. Maybe in instances its a more combined thing between time and distance.

> Should bosses use the invalid path evade or should they just ignore unreachable targets?

Bosses normally reset if the current victim is not reachable. To protect against exploits.

src/server/game/Entities/Creature/Creature.h
```diff
@@ -631,6 +633,9 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         virtual uint8 GetPetAutoSpellSize() const { return MAX_SPELL_CHARM; }
         virtual uint32 GetPetAutoSpellOnPos(uint8 pos) const;
 
+        void SetCannotReachTarget(bool cannotReach) { if (cannotReach == m_cannotReachTarget) return; m_cannotReachTarget = cannotReach; m_cannotReachTimer = 0; }
```
jackpoz: could you please move each statement on its own line to make it debuggable ?

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -37,7 +37,10 @@ void TargetedMovementGeneratorMedium<T, D>::_setTargetLocation(T* owner, bool up
         return;
 
     if (owner->GetTypeId() == TYPEID_UNIT && !i_target->isInAccessiblePlaceFor(owner->ToCreature()))
+    {
+        owner->ToCreature()->SetCannotReachTarget(true);
```
jackpoz: nice how this new feature is provided without any additional mmap check

Treeston: TargetedMovementGenerator was already aware of whether a path could be found, it just didn't do much with it.

Treeston:
I suspect that 5-man instances _do_ have this reset behavior, while raids do _not_ have this reset behavior.

sirikfoll:
Probably because of #16549, your script won't work on mostly situations that involves Z position variation.
Quick example, aggro a creature and fly above it, creature can't reach you, but it won't evade.

Treeston:
That's a core bug that doesn't really impact this PR. Once it gets fixed, that situation will work as well.

Keader:
Using Clean TC + This Pr, the faction fight in Trial of the crusader dont work.
After moving around (because fear), some boss give evade and despawn...
i dont kill any boss, and only 3 left... http://i.imgur.com/w7g5JUO.jpg
can you take a look?

jackpoz:
isn't any mmap-related functionality broken in ToC ?

Goatform:
I had a feeling to mention this commit 826717de since I think that it might be useful.

Keader:

tkrokli:
[[TrinityCore/TrinityCore/tree/3.3.5]](https://github.com/TrinityCore/TrinityCore/tree/3.3.5) : This branch is 2637 commits ahead, 5964 commits behind 6.x
[[Treeston/TrinityCore/tree/3.3.5-evademore]](https://github.com/Treeston/TrinityCore/tree/3.3.5-evademore) is 2588 commits ahead, 5964 commits behind TrinityCore:6.x

Looks like `3.3.5-evademore` needs 49 updates from the main source to become aligned to TC 3.3.5.

jackpoz:
there's no technical reason for a PR to "become aligned"

tkrokli:
No, but it is usually the easiest way to resolve a branch & PR conflict.

Keader:
TOC problem is solved ?

Treeston:
PR doesn't cause evades in any raid maps.

src/server/game/Entities/Creature/Creature.h
jameyboor: When only setting a member of primitive type you should avoid checking `if (cannotReach == m_cannotReachTarget) return;` only causes overhead.

jackpoz:
Wonderful addition to mmaps

SUN-Kelno:
I'm not 100% sure about this since this is just a quick comment reading your PR:
Won't m_cannotReachTimer keep incrementing if the creature is being feared or confused? (= ChaseMovementGenerator stop being updated because a generator of MOTION_SLOT_CONTROLLED  is running on top of it).
If so, the creature will evade/incorrectly regenerate health if the control was applied while the current target was unreachable.

Treeston:
Control can't be applied while the target is unreachable because the spell would be evaded.

kelno:
Oh indeed :)

Issue #16818, opened 2016-03-21
[3.3.5] Core/Spell: Creature spellcast facing. The creature is respawned and not updated


chaodhib:
I found a bug introduced in this PR: https://github.com/TrinityCore/TrinityCore/pull/15641#issuecomment-199421377

When a unit casts a spell, `SendUpdateToPlayer(player)` is called here: https://github.com/TrinityCore/TrinityCore/pull/15641/files#diff-f3f953a63e2a008d22de69cb68f735c4R2726

From what I understood, it updates players of the new `UNIT_FIELD_TARGET` of the caster mob. The problem is that the wrong packet is sent. `SendUpdateToPlayer()` calls `BuildCreateUpdateBlockForPlayer()` to construct the update packet.

The problem lies in the type of update sent. A `UpdateType: CreateObject1` is sent instead of an `UpdateType: Values`. In other words, the packet sent will re-spawn the caster mob from scratch instead of updating it. Visually, it looks like the mob is jumping, pretty weird stuff.

I asked @Treeston about it in IRC and apparently, it's related to the grid system. @r00ty-tc could you look into this when you find time please? Thanks.

bug observed in: 0cc1af721e05c89d82429aaf04c2b8241b768252

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you . Thanks.

Treeston:
What I meant is that the mentioned snippet that handles target updating should be re-written using a grid worker instead of the hacky loop it is now.

Don't have the time to do it myself, sadly.

Treeston:
```
<chaodhib> treeston: checkout my gluth branch
<chaodhib> pull gluth, wait that he eat a zombie
<chaodhib> when he does, when he casts the killing spell (DoCast(zombieToBeEaten, SPELL_ZOMBIE_CHOW_SEARCH_SINGLE);)
<chaodhib> that's when the bug occurus
<chaodhib> ServerToClient: SMSG_COMPRESSED_UPDATE_OBJECT (0x01F6) Length: 208 ConnIdx: 0 EP: 127.0.0.1:60947 Time: 03/21/2016 23:37:12.577 Number: 20034
<chaodhib> Count: 1
<chaodhib> [0] UpdateType: CreateObject1
<chaodhib> [0] GUID: Full: 0xF130003E3C000309 Type: Creature Entry: 15932 Low: 777
[...]

PR #16839, opened 2016-03-25
Adjustments to spell facing

chaodhib:
Cherry pick of @r00ty-tc's fix to the issues described here: https://github.com/TrinityCore/TrinityCore/issues/16818

Here is an excerpt of IRC chat:

> [00:26] @r00ty Treeston / chaodhib. https://github.com/r00ty-tc/TrinityCore/commit/c0693a6ee52d31840ab2ee878c4881884b857638 can either of you check that resolves all problems, but still works for spell facing etc (and doesn't cause explosions elsewhere). My only testing is, it compiles and should work in theory :P - I'm away for a week so your other option is to wait that long.
> [00:26] @r00ty Feel free to cherry pick/commit if it's all good, otherwise again it'll be next week ;)

I have tested it (not extensively however) and it looks good.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2786,17 +2786,17 @@ bool Creature::FocusTarget(Spell const* focusSpell, WorldObject const* target)
             )
         )
         {
-            const MapRefManager& mapPlayers = GetMap()->GetPlayers();
-            for (MapRefManager::const_iterator it = mapPlayers.begin(); it != mapPlayers.end(); ++it)
-                if (Player* player = (*it).GetSource())
+            std::list<Player*> playersNearby;
+            GetPlayerListInGrid(playersNearby, GetVisibilityRange());
+            for (Player* player : playersNearby)
+            {
+                // only update players that are known to the client (have already been created)
+                if (player->HaveAtClient(this))
                 {
-                    // only update players that can both see us, and are actually in combat with us (this is a performance tradeoff)
-                    if (player->CanSeeOrDetect(this, false, true) && IsInCombatWith(player))
-                    {
-                        SendUpdateToPlayer(player);
-                        shouldDelay = true;
-                    }
+                    SendUpdateToPlayer(player);
+                    shouldDelay = true;
                 }
+            }
             if (shouldDelay)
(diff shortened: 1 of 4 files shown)


Nawuko:
I think with your changes the IsInCombatWith(player) check can be removed. It is not retail behavior.

chaodhib:
@Nawuko Yeah, I agree. Since this is not my code, I prefer to have confirmation before I do the change.

r00ty-tc:
I agree, but that code was already there. Might be worth double checking with treeston to be certain but yes it doesn't make sense to skip players not in combat. 

Also I think the for loop could be turned into a c++11 style one (I forget the feature name) but that's just code style.

chaodhib:
@r00ty-tc In java, it's called a for each loop, not sure for c++.

The change is made. @Treeston any comment?

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2786,17 +2786,18 @@ bool Creature::FocusTarget(Spell const* focusSpell, WorldObject const* target)
             )
         )
         {
-            const MapRefManager& mapPlayers = GetMap()->GetPlayers();
-            for (MapRefManager::const_iterator it = mapPlayers.begin(); it != mapPlayers.end(); ++it)
-                if (Player* player = (*it).GetSource())
+            std::list<Player*> targets;
+            this->GetPlayerListInGrid(targets, GetVisibilityRange());
```
Treeston: Don't use `this->`

Treeston:
Looks good. :+1:

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2786,17 +2786,17 @@ bool Creature::FocusTarget(Spell const* focusSpell, WorldObject const* target)
             )
         )
         {
-            const MapRefManager& mapPlayers = GetMap()->GetPlayers();
-            for (MapRefManager::const_iterator it = mapPlayers.begin(); it != mapPlayers.end(); ++it)
-                if (Player* player = (*it).GetSource())
+            std::list<Player*> playersNearby;
+            GetPlayerListInGrid(playersNearby, GetVisibilityRange());
+            for(Player* player : playersNearby)
```
MitchesD: codestyle for _space_ (

r00ty-tc:
Yep seems fine (at least from what I can see on the phone screen) 

The stray this was there because I used it to find the function name. Then never removed it.

Issue #16808, opened 2016-03-19
Core/Movement: Ack messages are handled incorrectly


tripleslash:
Currently if any player around you mounts up YOUR own client will send an Ack message with the foreign players GUID as mover. The reason is that SMSG_FORCE_(SPEED)_CHANGE and other messages are sent to the current set instead of the local mover only. Foreign players should receive SMSG_(SPEED)_CHANGE and not SMSG_FORCE_(SPEED)_CHANGE.

Fix:
SMSG_FORCE_... opcodes should be sent to the person that currently controls this unit i.e. the  `Unit::m_movedPlayer`.

Example:

Stand in a crowded place and watch players around you mount up and dismount, then watch your clients outgoing packet traffic.

Packet log: http://underflow.in/i/yUvI3.png
None of these packets actually belong to me.

Fixing this would result in lots of reduced trash traffic in crowded areas.

I encountered this when I started to implement a proper movement anticheat into TC and have fixed it in a commit, you can cherrypick the relevant parts from here:

https://github.com/tripleslash/TrinityCore/commit/88d662f7b867f804629cbe1dbfb8fd2988de81d0

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you . Thanks.

sirikfoll:
really interesting :+1:

kvipka:
it's strange, but fear will call disconnect for target ><

tripleslash:
What? Do you mean in my commit? Theres lots of other broken stuff there. I never said it was supposed to be merged, just cherrypicking the relevant change will do for now until this mess is rewritten from scratch.

kvipka:
yep, i understand, just note for you =)

chaodhib:
fixed in https://github.com/TrinityCore/TrinityCore/issues/16843

Not part of this issue but it's included in @tripleslash's commit: I'm not sure if the system of resending FORCE_SPEED_CHANGE messages until an ACK message has been sent back is implemented. If this isn't the case, an issue should be opened about that.

Also, I don't understand why the need for ACK messages. Isn't WoW using TCP? And therefore, we have the guarantee that every packet sent by the server is received by the client (the TCP layer takes care of that). So why the need for the client to send `ACKNOWLEDGE` messages? Is it related to cheat protection?

Aokromes:
So, we can close this?

[...]

PR #16843, opened 2016-03-26
Core/Movement: Clean up and improvements on Unit::SetSpeed

chaodhib:
Clean up of Unit::SetSpeed. The biggest changes are:
- the opcodes used to send the change of the speed of a unit were reworked. until now, MSG_MOVE_SET_\* were sent for every units which isn't like retail behavior (more info in the comments of the code). 
- removed the method argument "forced" from Unit::SetSpeed since it is nor used nor useful.
- removed the method argument "forced" from Unit::UpdateSpeed since it is nor used nor useful.

Minor changes of the PR:
- renamed Unit::SetSpeed to Unit::SetSpeedRate (to stay coherent with the methods GetSpeed and GetSpeedRate)
- added a utility method Unit::SetSpeed (which takes into parameter the flat new value)

---

About the testing done:

Test when unit::SetSpeed() is called on a player. Test done with 2 players on the map:

> ServerToClient: SMSG_FORCE_RUN_SPEED_CHANGE (0x00E2) Length: 11 ConnIdx: 0 EP: 127.0.0.1:49302 Time: 03/26/2016 12:06:14.844 Number: 3500
> Guid: Full: 0x00000004 Type: Player Low: 4 Name: Druiderino
> MoveEvent: 0
> Unk Byte: 1
> New Speed: 7
> 
> ServerToClient: MSG_MOVE_SET_RUN_SPEED (0x00CD) Length: 52 ConnIdx: 0 EP: 127.0.0.1:49298 Time: 03/26/2016 12:06:14.844 Number: 3501
> GUID: Full: 0x00000004 Type: Player Low: 4 Name: Druiderino
> Movement Flags: Forward, StrafeLeft, Falling (4101)
> Extra Movement Flags: None (0)
> Time: 64753
> Position: X: 3672.161 Y: -1265.754 Z: 243.5119
> Orientation: 2.537083
> Fall Time: 0
> Fall Velocity: -7.955547
> Fall Sin Angle: -0.9836842
> Fall Cos Angle: -0.179904
> Fall Speed: 11.9
> Speed: 7
> 
> ClientToServer: CMSG_FORCE_RUN_SPEED_CHANGE_ACK (0x00E3) Length: 56 ConnIdx: 0 EP: 127.0.0.1:49302 Time: 03/26/2016 12:06:14.847 Number: 3516
> Guid: Full: 0x00000004 Type: Player Low: 4 Name: Druiderino
> Movement Counter: 0
> Movement Flags: Forward, StrafeLeft, Falling (4101)
> Extra Movement Flags: None (0)
> Time: 493594943
> Position: X: 3670.382 Y: -1266.079 Z: 244.4983
> Orientation: 2.588134
> Fall Time: 152
> Fall Velocity: -7.955547
> Fall Sin Angle: -0.9836842
> Fall Cos Angle: -0.179904
> Fall Speed: 11.9
> New Speed: 7

---

Test when unit::SetSpeed() is called on a NPC. Test done with 2 players on the map:

> ServerToClient: SMSG_MOVE_SPLINE_SET_RUN_SPEED (0x02FE) Length: 11 ConnIdx: 0 EP: 127.0.0.1:49348 Time: 03/26/2016 12:10:06.259 Number: 9460
> GUID: Full: 0xF130003E3C000309 Type: Creature Entry: 15932 Low: 777
> Amount: 36
> 
> ServerToClient: SMSG_MOVE_SPLINE_SET_RUN_SPEED (0x02FE) Length: 11 ConnIdx: 0 EP: 127.0.0.1:49350 Time: 03/26/2016 12:10:06.259 Number: 9461
> GUID: Full: 0xF130003E3C000309 Type: Creature Entry: 15932 Low: 777
> Amount: 36

In both cases, the right packets are sent to the right destinations. No more no less.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -782,7 +782,7 @@ void Creature::DoFleeToGetAssistance()
         cell.Visit(p, grid_creature_searcher, *GetMap(), *this, radius);
 
         SetNoSearchAssistance(true);
-        UpdateSpeed(MOVE_RUN, false);
+        UpdateSpeed(MOVE_RUN);
 
         if (!creature)
             //SetFeared(true, EnsureVictim()->GetGUID(), 0, sWorld->getIntConfig(CONFIG_CREATURE_FAMILY_FLEE_DELAY));
@@ -1620,7 +1620,7 @@ void Creature::setDeathState(DeathState s)
         if (HasSearchedAssistance())
         {
             SetNoSearchAssistance(false);
-            UpdateSpeed(MOVE_RUN, false);
+            UpdateSpeed(MOVE_RUN);
         }
 
         //Dismiss group if is leader
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -381,10 +381,10 @@ bool Creature::InitEntry(uint32 entry, CreatureData const* data /*= nullptr*/)
 
     SetFloatValue(UNIT_MOD_CAST_SPEED, 1.0f);
 
-    SetSpeed(MOVE_WALK,   cinfo->speed_walk);
-    SetSpeed(MOVE_RUN,    cinfo->speed_run);
-    SetSpeed(MOVE_SWIM,   1.0f); // using 1.0 rate
-    SetSpeed(MOVE_FLIGHT, 1.0f); // using 1.0 rate
+    SetSpeedRate(MOVE_WALK,   cinfo->speed_walk);
+    SetSpeedRate(MOVE_RUN,    cinfo->speed_run);
+    SetSpeedRate(MOVE_SWIM,   1.0f); // using 1.0 rate
+    SetSpeedRate(MOVE_FLIGHT, 1.0f); // using 1.0 rate
 
     // Will set UNIT_FIELD_BOUNDINGRADIUS and UNIT_FIELD_COMBATREACH
     SetObjectScale(cinfo->scale);
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12493,9 +12493,11 @@ void Unit::SetSpeed(UnitMoveType mtype, float rate, bool forced)
                 pet->SetSpeed(mtype, m_speed_rate[mtype], forced);
     }
 
-    if (GetTypeId() == TYPEID_PLAYER && ToPlayer()->m_mover->GetTypeId() == TYPEID_PLAYER)
+    if (m_movedPlayer) 
     {
-        // Send notification to self. only sent to one client (the client of the player concerned by the change).
+        // unit controlled by a player.
+
+        // Send notification to self. this packet is only sent to one client (the client of the player concerned by the change).
         WorldPacket self;
         self.Initialize(moveTypeToOpcode[mtype][1], mtype != MOVE_RUN ? 8 + 4 + 4 : 8 + 4 + 1 + 4);
         self << GetPackGUID();
@@ -12503,7 +12505,7 @@ void Unit::SetSpeed(UnitMoveType mtype, float rate, bool forced)
         if (mtype == MOVE_RUN)
             self << uint8(1);                               // unknown byte added in 2.1.0
         self << float(GetSpeed(mtype));
-        ToPlayer()->GetSession()->SendPacket(&self); // this is a hack. technically, players can control some NPcs (like 19405) and it will result in a SMSG_FORCE_RUN_SPEED_CHANGE).
+        m_movedPlayer->GetSession()->SendPacket(&self);
 
         // Send notification to other players. sent to every clients (if in range) except one: the client of the player concerned by the change.
         WorldPacket data;
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12468,100 +12468,59 @@ void Unit::SetSpeed(UnitMoveType mtype, float rate, bool forced)
 
     propagateSpeedChange();
 
-    WorldPacket data;
-    if (!forced)
+    // Spline packets are for units controlled by AI. "Force speed change" (wrongly named opcodes) and "move set speed" packets are for units controlled by a player.
+    static Opcodes const moveTypeToOpcode[MAX_MOVE_TYPE][3] =
     {
-        switch (mtype)
-        {
-            case MOVE_WALK:
-                data.Initialize(MSG_MOVE_SET_WALK_SPEED, 8+4+2+4+4+4+4+4+4+4);
-                break;
-            case MOVE_RUN:
-                data.Initialize(MSG_MOVE_SET_RUN_SPEED, 8+4+2+4+4+4+4+4+4+4);
-                break;
-            case MOVE_RUN_BACK:
-                data.Initialize(MSG_MOVE_SET_RUN_BACK_SPEED, 8+4+2+4+4+4+4+4+4+4);
-                break;
-            case MOVE_SWIM:
-                data.Initialize(MSG_MOVE_SET_SWIM_SPEED, 8+4+2+4+4+4+4+4+4+4);
-                break;
-            case MOVE_SWIM_BACK:
-                data.Initialize(MSG_MOVE_SET_SWIM_BACK_SPEED, 8+4+2+4+4+4+4+4+4+4);
-                break;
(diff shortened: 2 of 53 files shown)


Kittnz:
Related to https://github.com/TrinityCore/TrinityCore/issues/16808 ?

chaodhib:
@Kittnz I didn't had in mind any issue to fix when I wrote this PR but actually, yes ^^

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12468,100 +12468,59 @@ void Unit::SetSpeed(UnitMoveType mtype, float rate, bool forced)
...
+    {
+        // Send notification to self. only sent to one client (the client of the player concerned by the change).
+        WorldPacket self;
+        self.Initialize(moveTypeToOpcode[mtype][1], mtype != MOVE_RUN ? 8 + 4 + 4 : 8 + 4 + 1 + 4);
+        self << GetPackGUID();
+        self << (uint32)0;                                  // Movement event encounter. Unimplemented at the moment! NUM_PMOVE_EVTS = 0x39Z. 
+        if (mtype == MOVE_RUN)
+            self << uint8(1);                               // unknown byte added in 2.1.0
+        self << float(GetSpeed(mtype));
+        ToPlayer()->GetSession()->SendPacket(&self); // this is a hack. technically, players can control some NPcs (like 19405) and it will result in a SMSG_FORCE_RUN_SPEED_CHANGE).
```
tripleslash: `m_movedPlayer->GetSession()->SendPacket(...)`

chaodhib: Indeed. `m_movedPlayer `is a better choice than `ToPlayer()` here. And indeed, it should be converted to a GUID.

chaodhib: And also, i don't understand why we store both `m_mover` and `m_movedPlayer`. With the former, we can get the latter.

tripleslash:
@chaodhib Well the whole movement handling is crying for a rewrite imo. It was written by people who had no clue about how it actually works.

chaodhib:
@tripleslash Yeah I completely agree. We have a better understanding of the packets now. It helps a lot.

chaodhib:
I think I'm finished.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12468,100 +12468,59 @@ void Unit::SetSpeed(UnitMoveType mtype, float rate, bool forced)
...
+        if (!IsInCombat())
+            if (Pet* pet = ToPlayer()->GetPet())
+                pet->SetSpeed(mtype, m_speed_rate[mtype], forced);
+    }
+
+    if (GetTypeId() == TYPEID_PLAYER && ToPlayer()->m_mover->GetTypeId() == TYPEID_PLAYER)
+    {
+        // Send notification to self. only sent to one client (the client of the player concerned by the change).
+        WorldPacket self;
+        self.Initialize(moveTypeToOpcode[mtype][1], mtype != MOVE_RUN ? 8 + 4 + 4 : 8 + 4 + 1 + 4);
```
KarrokDC: Might just be me, but I always try to avoid "magic numbers", as in the 8 + 4+ 4 .. etc. Could consider pulling these to a properly named local const variable, especially if this sequence is used more in the code.

chaodhib: these numbers are directly related to the lines of code below. they are the number of bytes that new each data added to the packet (using `data << ***`) takes.

Not particularly nice to read or write but it's useful and I haven't found any better to way.

src/server/scripts/Commands/cs_modify.cpp
```diff
@@ -525,7 +525,7 @@ class modify_commandscript : public CommandScript
         if (handler->needReportToTarget(target))
             ChatHandler(target->GetSession()).PSendSysMessage(LANG_YOURS_SPEED_CHANGED, handler->GetNameLink().c_str(), Speed);
 
-        target->SetSpeed(MOVE_RUN, Speed, true);
+        target->SetSpeedRate(MOVE_RUN, Speed);
```
KarrokDC: local variable "Speed" should probably be "speed" all lowercase

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -381,10 +381,10 @@ bool Creature::InitEntry(uint32 entry, CreatureData const* data /*= nullptr*/)
 
     SetFloatValue(UNIT_MOD_CAST_SPEED, 1.0f);
 
-    SetSpeed(MOVE_WALK,   cinfo->speed_walk);
-    SetSpeed(MOVE_RUN,    cinfo->speed_run);
-    SetSpeed(MOVE_SWIM,   1.0f); // using 1.0 rate
-    SetSpeed(MOVE_FLIGHT, 1.0f); // using 1.0 rate
+    SetSpeedRate(MOVE_WALK,   cinfo->speed_walk);
+    SetSpeedRate(MOVE_RUN,    cinfo->speed_run);
+    SetSpeedRate(MOVE_SWIM,   1.0f); // using 1.0 rate
```
KarrokDC: is 1.0 some default rate? if so, can pull that out to a const value somewhere to be more descriptive

src/server/scripts/Commands/cs_modify.cpp
chaodhib: this is not my code. I just renamed the name of the method and removed one parameter. The main objective of this PR is the changes of implementation of that method.

src/server/game/Entities/Creature/Creature.cpp
chaodhib: cf above

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12455,7 +12455,12 @@ float Unit::GetSpeed(UnitMoveType mtype) const
     return m_speed_rate[mtype]*(IsControlledByPlayer() ? playerBaseMoveSpeed[mtype] : baseMoveSpeed[mtype]);
 }
 
-void Unit::SetSpeed(UnitMoveType mtype, float rate, bool forced)
+void Unit::SetSpeed(UnitMoveType mtype, float newValue)
```
ccrs: If u have modified all the uses of SetSpeed, why do u keep it?

chaodhib:
@ccrs I understand the confusion. What I did is:
- `Unit::SetSpeed` (which takes the new speed rate as parameter) was renamed to `Unit::SetSpeedRate`
- then I introduced a new method called SetSpeed, which takes the new speed but as a flat value this time.

Why these changes you may ask?
Because now, the Unit class has those 4 clean methods:
```
float GetSpeed(UnitMoveType mtype) const
float GetSpeedRate(UnitMoveType mtype) const
void SetSpeed(UnitMoveType mtype, float newValue
void SetSpeedRate(UnitMoveType mtype, float rate)
```

ccrs:
and what are they used for? each of them.

chaodhib:
Let's imagine we have a creature with a base run speed of 11.

- A call to `float GetSpeed(MOVE_RUN) const` on that creature will return 11
- float GetSpeedRate(MOVE_RUN) const will return 11/7

If you want to set the speed of the creature to 15, simply call: `creature->SetSpeed(MOVE_RUN, 15);`

On the other hand, if you want to increase the speed of the creature by 20%, you'll simply call `creature->SetSpeedRate(MOVE_RUN, 1.2);` .... or not. I just realized this last thing is not true. `creature->SetSpeedRate(MOVE_RUN, 1.2*11/7);` is the correct call.

I'm wandering where [these values](https://github.com/TrinityCore/TrinityCore/blob/6.x/src/server/game/Entities/Unit/Unit.cpp#L75) come from.

Shauren:
@chaodhib They were picked as most commonly appearing values in sniffs (creatures with no modifiers)

chaodhib:
I see. Also, it looks like they are the base speeds for players. 

I'm beginning to think that we should stop using them when dealing with creatures. Dividing the speed of a creature by the average speed of all creatures doesn't represent anything useful. Might as well manipulate and store the value of the speed itself. It has no added value and it's error prone.

On the other hand, the ratio `current speed / base speed of that creature` means more and it could be interesting to have for people who work on scripts or spells.

chaodhib:
Since it appaears to cause a lot of confusion, should I remove the renaming (Unit::SetSpeed to Unit::SetSpeedRate) in order to focus on the most important which is: 

- Unit::SetSpeed method has been completely rewritten (sending now the correct packets in each situation and reducing the traffic) 
- and one (unused) method parameter was removed from the same method.

?

Issue #15040, opened 2015-07-09
[SPELL/DK] Pestilence


P-Kito:
I am not sure if it's a problem on current TC. But maybe someone can approve this.

Should:
Take a point between the target and dk and spread diseases to all targets within 10y (15y with Glyph).

Current:
Only spreads diseases to targets infront of the DK. And when having the Glyph to increase range to 15y, Pestilence won't trigger when DK is standing too close to the target.

Can someone confirm this on TC?

Information to this spell:
Patch 3.1.0 (14-Apr-2009): Does no damage and has only one rank. This ability will now spread disease centered on a position between the death knight and its target, rather than the target’s center position (for very large creatures, this used to cause no disease spread at all).
http://wowwiki.wikia.com/Pestilence
Uses 3 Spell_effect.
One for Single target, one for 10y AOE Enemy and 1 with 5yard target_front.

Keader:
TC Hash?

P-Kito:
As I said, someone has to approve this, since I don't have a runable TC on this computer.

Would be cool if you log in and try it with a DK and put ur hash in here :+1:

DDuarte:
We got enough *real* issues, creating potential issues doesn't seem like a good idea.

If someone does confirm this I'll re-open the issue.

fat4l:
tested on:TrinityCore rev. e77744b48eac 2015-09-30 19:16:09 +0200 (master branch) (Unix, Release)

Confirm this: "Only spreads diseases to targets infront of the DK. "
Pestilence doesn't spread if enemies are behind the DK(it doesn't spread as AOE).

Can't confirm this: "And when having the Glyph to increase range to 15y, Pestilence won't trigger when DK is standing too close to the target."

Aokromes:
There is no such rev on trinitycore.

tkrokli:
Nor is there any 'master' branch on TrinityCore since November 2014. Use TC source for support.

fat4l:
ah, what about this ? 
https://github.com/TrinityCore/TrinityCore/commit/d476e0eff24af5849bbbc21128aa461a24d4fe9b
that's the last commit that we merged at.

Aokromes:
To old commit/who knows how much custom code to trust you report.

PR #16849, opened 2016-03-26
[3.3.5] Game/SpellMgr: DBC correction for Pestilence

P-Kito:
**Changes proposed**:
- Pestilence currently only works for targets infront of the DK, remove cone effect which makes no sense

**Target branch(es)**: 335

**Issues addressed**: https://github.com/TrinityCore/TrinityCore/issues/15040

**Tests performed**: Ingame tested

**Known issues and TODO list**:     --------

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -3337,6 +3337,10 @@ void SpellMgr::LoadSpellInfoCorrections()
                 //! HACK: This spell break quest complete for alliance and on retail not used °_O
                 spellInfo->Effects[EFFECT_0].Effect = 0;
                 break;
+            case 50842: // Pestilence
+                // original has cone effect which does not seem to make sense
+                spellInfo->Effects[EFFECT_1].TargetA = SpellImplicitTargetInfo(TARGET_UNIT_TARGET_ENEMY);
+                break;
             // VIOLET HOLD SPELLS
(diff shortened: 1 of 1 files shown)


P-Kito:
https://github.com/TrinityCore/TrinityCore/issues/15040
Some discussion there

NorthStrider:
Isn't effect_1 a dummy effect and according to that not the actual trigger? effect_2 is a 10 yards script effect which should handle the spreading. So the spellscript seems to be more inaccurate.

Shauren:
What really happens here:
1. EFFECT_0 targeting - the unit we selected to spread diseases from
2. EFFECT_1 targeting - moves DEST 5 yards in front of caster
3. EFFECT_2 targeting - selects all enemies in 10 yard radius centered on DEST moved in EFFECT_1

Not sure if this is a bug at all.

P-Kito:
According to the discussion diseases will only spread in that cone effect. But you are right @Shauren , it should be like that exactly.

ccrs:
not cone, but 10 yards around 5 yards in front of caster

P-Kito:
Suggestion: Rewrite the spellscript according to Shaurens idea.

Greymane:
Just use a OnDestinationTargetSelect hook into EFFECT_1 and move the Dest back to EFFECT_0 target's position, that will fix the problem.

Issue #16836, opened 2016-03-25
[3.3.5][Hunter] Core/Spell: talent "Sniper Training" triggering while dead


Treeston:
3.3.5 f37682b

The hunter talent [Sniper Training](http://wotlk.openwow.com/spell=53304) applies a buff to the player if they stand still for 6 seconds.

The current core attempts to apply this buff even if the player is dead, leading to the player receiving a "You are dead" error message every 5 seconds while standing still as a ghost.

Kodekc:
Same with talent http://wotlk.openwow.com/spell=48108!
With the explosion of fire bombs, proc will attempt to activate a dead aim

P-Kito:
Is there any proc in WoW that requires to proc when being dead?

mik1893:
i think we should just return... cant think of something that procs when you are dead

Treeston:
I can pretty much guarantee that there is some niche spell somewhere that procs while dead, yes.

mik1893:
of course, it's bl1z....

Treeston:
That is not how core development works.

Shauren:
As I said in the closed PR that attempted to fix it - Sniper Training is NOT A PROC
It's a periodic trigger aura.

Nawuko:
My take on this: https://gist.github.com/Nawuko/c9f489418b6b89d1fc1ecb09a29f12d7
No gurantees and line count dosent match (You have to apply manually)

Cheers.

Treeston:
Tested, @Nawuko? Looks like a clean solution to me (curious that `TRIGGERED_DONT_REPORT_CAST_ERROR` isn't already checked there).

Shauren:
Why not here https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/Spell.cpp#L3791?

Treeston:
Agree with @Shauren. We should remove all the various `TRIGGERED_DONT_REPORT_CAST_ERROR` checks from CheckCast and have it checked in `SendCastResult` instead.

Nawuko:
Yea, i try always to change little as i can when it comes to something like the spell system. @Shaures sugestion makes much more sense. So go go open a PR tree <3

Treeston:
Will do later today when I get home.

Shauren:
[...]

PR #16844, opened 2016-03-26
[3.3.5/6.x] Game/Spell: Do not proc when player is dead

P-Kito:
**Changes proposed**:
- Procs will no longer proc when the "proccer" is dead

**Target branch(es)**: 335/6x

**Issues addressed**: Fixes https://github.com/TrinityCore/TrinityCore/issues/16836

**Tests performed**: Not tested, needs review, is this the right position for the IsAlive check?

PR #16857, opened 2016-03-27
[3.3.5/6.x] Game/Spell: Do not proc when player is dead

P-Kito:
**Changes proposed**:
- Procs will no longer proc when the "proccer" is dead

**Target branch(es)**: 335/6x

**Issues addressed**: Fixes https://github.com/TrinityCore/TrinityCore/issues/16836

**Tests performed**: Not tested ingame

is this the right position for this check?

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -14075,6 +14075,9 @@ void Unit::ProcDamageAndSpellFor(bool isVictim, Unit* target, uint32 procFlag, u
     // Player is loaded now - do not allow passive spell casts to proc
     if (GetTypeId() == TYPEID_PLAYER && ToPlayer()->GetSession()->PlayerLoading())
         return;
+    // Player is dead - do not proc on dead player
+    if (!IsAlive())
+        return;
     // For melee/ranged based attack need update skills and set some Aura states if victim present
     if (procFlag & MELEE_BASED_TRIGGER_MASK && target)
     {


tkrokli:
This PR replaces PR #16844.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -14075,6 +14075,9 @@ void Unit::ProcDamageAndSpellFor(bool isVictim, Unit* target, uint32 procFlag, u
     // Player is loaded now - do not allow passive spell casts to proc
     if (GetTypeId() == TYPEID_PLAYER && ToPlayer()->GetSession()->PlayerLoading())
         return;
+    // Player is dead - do not proc on dead player
```
jackpoz: are you sure "this" is a Player at this point ?

tkrokli: So if "this" is not necessarily a player, we need to check for `GetTypeId() == TYPEID_PLAYER` in addition to `!IsAlive()` --- is that the preferred way to do it?

Sdyessdev: If it's only for players then you should make sure that "this" is a player and not something else. Also, hi Jackpoz.

Shauren:
Sniper Training is not a proc aura, it's a periodic trigger https://github.com/P-Kito/TrinityCore/blob/proc/src/server/scripts/Spells/spell_hunter.cpp#L764
Seeing that this does not fix the issue and can potentially break procs expecting dead units I'm closing this PR.

Nawuko:
Quick update: Something like that dosent change any spell behavior but supresses the error message: https://gist.github.com/Nawuko/0df637d4ccd7e5855735b9e2f9a7b40a (triggered spells should show a error message if the victim is dead)

PR #16876, opened 2016-03-29
Core/Movement: Fix a bug related to unit movements

chaodhib:
Fix a bug related to unit movements:
Let's say creature A is attacking unit B. If creature A then starts to attack (or follow) unit C and if unit C is very close to Creature A, Creature A will move away from unit C for a few meters before auto-attacking (or following). This isn't natural looking and it isn't blizz-like behavior.

It is mentioned at least here (and probably at other places): https://github.com/TrinityCore/TrinityCore/pull/16806#issuecomment-198741472 & https://github.com/TrinityCore/TrinityCore/pull/16806#issuecomment-199519967

I have tested the change IG and it looks good.

### DIFF
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -48,6 +48,9 @@ void TargetedMovementGeneratorMedium<T, D>::_setTargetLocation(T* owner, bool up
     {
         if (!i_offset)
         {
+            if (i_target->IsWithinDistInMap(owner, CONTACT_DISTANCE))
+                return;
+
             // to nearest contact position
             i_target->GetContactPoint(owner, x, y, z);
         }


chaodhib:
wow, sorry about that. slightly tired

Takenbacon:
This in fact blizzlike for at least WoD (not sure about WoTLK, but can't imagine it would be any different), if you are attacking a mob on retail and step forward into his radius, after a short delay the mob will always back up and continue attacking you from a specific distance away.

chaodhib:
Do you have a proof? a video or something

Takenbacon:
@chaodhib I'll jump on retail and make a video super quick, give me ~20 minutes.

chaodhib:
Sure. Personally, i have this as a potential counter proof: https://youtu.be/3BC9KpeN2UM?t=4m48s

Takenbacon:
@chaodhib https://www.youtube.com/watch?v=Oqux5vc9Qdo Disgusting ass quality because apparently OBS is misconfigured some how, but whatever, you get the point.

Nawuko:
I think @chaodhib is right. I cant remeber creatures adjusting back in 3.3.5. If i recall correctly this was added in MoP or WoD (i lean more towards WoD).

Kittnz:
Eitherway, adding it as an option is also a good idea and can be cherry picked to 6.x

chaodhib:
Besides the fact that this functionality was probably implemented after Wotlk, technically what this PR fixes and what @Takenbacon is talking about are not the same thing. 

This PR fixes a re-positioning after the creature changes target and that target is too close to it. This re-positioning only happens after a target change. You can't test this 1vs1. you need another NPC or a player besides you to test this.

In the example of Gluth I shown above, first he attacks the player, then he changes target and goes on the zombie. Finally, he returns on the player. When he does, he is very close to the player and yet we can see that Gluth is only rotating to face the player, not re-positioning himself.

Kittnz:
Wouldn't it be good to do this with the other Movement Generators too? Like what @Takenbacon means? Would be nice to have this as an option.

chaodhib:
@Kittnz You are mistaken. Again: this PR does not implement a new functionality. It's a bug-fix. The functionality evoked by @Takenbacon was not implemented in Woltk. Even though they share some similarities on the player point of view, this PR does not try to implement that functionality. In the technical side, it has nothing to do with it.

The code involve just adds a missing test. The same test is correctly done [line 79](https://github.com/chaodhib/TrinityCore/blob/move_chase_follow_repositioning_fix/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L79) in case the boolean [line 49](https://github.com/chaodhib/TrinityCore/blob/move_chase_follow_repositioning_fix/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L49) is evaluated to `false`. It was missing in case that `if` was evaluated to `true`.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -48,6 +48,9 @@ void TargetedMovementGeneratorMedium<T, D>::_setTargetLocation(T* owner, bool up
     {
         if (!i_offset)
         {
+            if (i_target->IsWithinDistInMap(owner, 1.0f))
```
Warpten: This will fail for large models - it makes more sense to check against bounding box.

chaodhib: What do you mean by "bounding box"?

By the way, [_IsWithinDist](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Object/Object.cpp#L1124) takes into account the objectSize of both the moving unit and the target.

chaodhib:
Images to demonstrate the bug is probably better. 

This is shown with 1 creature and 2 players. First, one of the players takes aggro of the creature. Then the second player gets very close to the creature (at a distance inferior to `Creature ->ObjectSize+Player->ObjectSize + 0.5` to be precise) and takes aggro back.

Before the fix: the creature backs away a way meters before turning towards the player
http://i.giphy.com/l3V0aSEiRuW7O4ysg.gif

After the fix: After the change of aggro, the creature just turns toward its new target.
http://i.giphy.com/26AHy9Km5oikIr7kA.gif

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -48,6 +48,9 @@ void TargetedMovementGeneratorMedium<T, D>::_setTargetLocation(T* owner, bool up
     {
         if (!i_offset)
         {
+            if (i_target->IsWithinDistInMap(owner, CONTACT_DISTANCE));
```
tkrokli: Remove the semicolon at the end of this line.

chaodhib: oups. thank you.

PR #16884, opened 2016-03-31
moved GetSkill function to more appropriate ItemTemplate class

Yehonal:
**Changes proposed**:

GetSkill is an Item function that just do checks on ItemTemplate, so it can be easily moved under ItemTemplate class and be used when you just have a prototype object

**Target branch(es)**: 335

**Tests performed**: Compilation works and supports legacy code

### DIFF
--- a/src/server/game/Entities/Item/ItemPrototype.h
+++ b/src/server/game/Entities/Item/ItemPrototype.h
@@ -737,6 +737,41 @@ struct ItemTemplate
         return std::max<float>(0.f, itemLevel);
     }
 
+    uint32 GetSkill() const
+    {
+        const static uint32 item_weapon_skills[MAX_ITEM_SUBCLASS_WEAPON] =
+        {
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Item/ItemPrototype.h
```diff
@@ -737,6 +737,41 @@ struct ItemTemplate
         return std::max<float>(0.f, itemLevel);
     }
 
+    uint32 GetSkill() const
```
jackpoz: unrelated to this PR and related to the existing code too: what is the point of placing all the logic in the header ?

Yehonal: Indeed, i've placed it there to avoid different ItemTemplate "code-standard"

jackpoz:
This PR looks fine. The merger should move the logic to .cpp and cleanup the .h while at it.

It will probably have to wait for next TDB release before being merged tho.

Naios:
Note that variables only placed in header files could lead to multiple instances of this variable when linking dynamically.

PR #16890, opened 2016-04-01
Core/Spells: Fix PvP trinkets animations

efectokoleos:
**Changes proposed**:
- Add spellscript for animations of pvp trinkets 

**Target branch(es)**: 6x

**Issues addressed**: No issue created yet

**Tests performed**:  build and tested ingame

### DIFF
--- a/sql/updates/world/2016_04_01_04.world.sql
+++ b/sql/updates/world/2016_04_01_04.world.sql
@@ -1,0 +1,3 @@
+DELETE FROM `spell_script_names` WHERE `spell_id`= 42292;
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
+(42292, 'spell_gen_pvp_trinket');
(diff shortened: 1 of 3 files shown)


sql/updates/world/2016_04_01_04.world.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM `spell_script_names` WHERE `spell_id`= 42292;
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
+(42292, 'spell_gen_pvp_trinket');
```
Keader: missing new line.
And tc use spell name on Delete querry like:
```sql
DELETE FROM `spell_script_names` WHERE `ScriptName`= 'spell_gen_pvp_trinket';
```

P-Kito:
nice work! 

But please rename the SQL to:

9999_99_99_00_world.sql

efectokoleos:
Ok thanks for the advises i think that is correct now

src/server/scripts/Spells/spell_generic.cpp
```diff
@@ -2710,6 +2710,48 @@ class spell_gen_profession_research : public SpellScriptLoader
         }
 };
 
+enum TrinketSpells
+{
+    SPELL_PVP_TRINKET_ALLIANCE  = 97403,
+    SPELL_PVP_TRINKET_HORDE     = 97404
+};
+
```
P-Kito: // spellid - spellname

this helps alot when searching specific spells later. But it's more or less a visual suggestion.

PR #16916, opened 2016-04-05
minor Coverity issue fix proposal

KarrokDC:
**Changes proposed**:
- Test on unchanged FirstLocale
- Exit with error if first locale is -1 as indicator of undefined locales

**Target branch(es)**: 6x

**Issues addressed**: Fixes Coverity CID 1341054

[x] Performed local build

none

CID 1341054

In the (very) unlikely event the LOCALE enum is gone missing, the FirstLocale local variable will remain -1, further code would then use a -1 to index an array in the OpenCascStorage(FirstLocale) line.

Two fixes come to mind:
a. check for FirstLocale == -1 and exit with an error message
b. set the default to enUS ( initialize FirstLocale = 0; and declare it as uint instead of int.

I chose the first as in the good flow it won't change anything and I was unsure of the impact if I put enUS as default if no locales were defined at all.

### DIFF
--- a/src/tools/vmap4_extractor/vmapexport.cpp
+++ b/src/tools/vmap4_extractor/vmapexport.cpp
@@ -488,8 +488,8 @@ int main(int argc, char ** argv)
 #endif
                     ))
             success = (errno == EEXIST);
-    int defaultFirstLocale = -1;
-    int FirstLocale = defaultFirstLocale;
+
+    int FirstLocale = -1;
     for (int i = 0; i < TOTAL_LOCALES; ++i)
     {
         if (i == LOCALE_none)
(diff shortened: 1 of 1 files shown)


jackpoz:
are you sure this affects 335 branch too ?

src/tools/vmap4_extractor/vmapexport.cpp
```diff
@@ -488,8 +488,8 @@ int main(int argc, char ** argv)
 #endif
                     ))
             success = (errno == EEXIST);
-
-    int FirstLocale = -1;
+    int defaultFirstLocale = -1;
```
jackpoz: please declare this as const so it doesn't get modified by accident in the next lines

KarrokDC:
Unsure, removed 3.3.5 as target branch.

src/tools/vmap4_extractor/vmapexport.cpp
KarrokDC: From discussion @ irc, @Shauren suggested to not even declare that default and instead just use -1 straight up here and in the test.

jackpoz:
To the merger: please classify the Coverity issue and mark it as Fix Submitted once merged

PR #16940, opened 2016-04-11
Adding new GM commands to control the combat reach & bounding radius

chaodhib:
Add .modify combatreach XXX and .modify boundingradius IG as GM commands

It needs some clean up before it gets merged.

I'm posting in part so that for those who are curious can test what I described here: https://community.trinitycore.org/topic/12515-unambiguously-defining-auto-attack-and-spell-range/#comment-78179

### DIFF
--- a/src/server/scripts/Commands/cs_modify.cpp
+++ b/src/server/scripts/Commands/cs_modify.cpp
@@ -51,6 +51,8 @@ public:
         {
             { "arenapoints",  rbac::RBAC_PERM_COMMAND_MODIFY_ARENAPOINTS,  false, &HandleModifyArenaCommand,         "" },
             { "bit",          rbac::RBAC_PERM_COMMAND_MODIFY_BIT,          false, &HandleModifyBitCommand,           "" },
+            { "boundingradius",  rbac::RBAC_PERM_COMMAND_MODIFY_SCALE,        false, &HandleModifyBoundingRadiusCommand,         "" },
+            { "combatreach",  rbac::RBAC_PERM_COMMAND_MODIFY_SCALE,        false, &HandleModifyCombatReachCommand,         "" },
             { "drunk",        rbac::RBAC_PERM_COMMAND_MODIFY_DRUNK,        false, &HandleModifyDrunkCommand,         "" },
             { "energy",       rbac::RBAC_PERM_COMMAND_MODIFY_ENERGY,       false, &HandleModifyEnergyCommand,        "" },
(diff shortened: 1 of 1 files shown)


src/server/scripts/Commands/cs_modify.cpp
```diff
@@ -1400,6 +1402,100 @@ class modify_commandscript : public CommandScript
...
+        //    if (handler->HasLowerSecurity(player, ObjectGuid::Empty))
+        //        return false;
+
+        //    handler->PSendSysMessage(LANG_YOU_CHANGE_SIZE, handler->GetNameLink(player).c_str(), Scale);
+        //    if (handler->needReportToTarget(player))
+        //        ChatHandler(player->GetSession()).PSendSysMessage(LANG_YOURS_SIZE_CHANGED, handler->GetNameLink().c_str(), Scale);
+        //}
+
+        
+        std::string output(". modifying Combat Reach to: ");
```
MitchesD: we don't use raw strings, use trinity_strings

src/server/scripts/Commands/cs_modify.cpp
```diff
@@ -1400,6 +1402,100 @@ class modify_commandscript : public CommandScript
...
+        //    if (handler->HasLowerSecurity(player, ObjectGuid::Empty))
+        //        return false;
+
+        //    handler->PSendSysMessage(LANG_YOU_CHANGE_SIZE, handler->GetNameLink(player).c_str(), Scale);
+        //    if (handler->needReportToTarget(player))
+        //        ChatHandler(player->GetSession()).PSendSysMessage(LANG_YOURS_SIZE_CHANGED, handler->GetNameLink().c_str(), Scale);
+        //}
+
+
+        std::string output(". modifying Bounding Radius to: ");
```
MitchesD: the same here

Shauren:
Yay, more commands to edit updatefields directly. Also not agreeing that it is a "useful" command to be placed as modify subcommand, it fits more under debug commands.
Also the same can be achieved by using
```.debug setvalue index_of_combatreach_field combatreach_value 0```
You can get value of combatreach field index from visual studio tooltips

chaodhib:
> Also the same can be achieved by using
> ``.debug` setvalue index_of_combatreach_field combatreach_value 0``

Erf, I was looking for that command. I guess you are right. Since it already exists in that form and since it's mostly to be a debug tool, I'm dropping this PR.

PR #16955, opened 2016-04-13
Fix flight form & fall damage

### LINKED ISSUE #16627: [3.3.5] Core/Spell: Druid flight form & fall damage
3.3.5 fef06781f510b

Repro:
1. Dismount/unshift from flight form at high altitude and start falling
2. Shortly before hitting the ground, switch to flight form
3. Wait a second or two, then cancel flight form

Expected result:
- As the player only fell a short distance, no/trivial fall damage should be taken

Actual result:
- Player dies of fall damage when hitting the ground as if they had fallen all the way from the original altitude

vtronko:
Fixes #16627 issue.
This handles all variations of that uninteded behavior, including druid flight form and .gm fly commands. Tested in-game, it works flawlessly.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -26090,6 +26090,9 @@ bool Player::SetCanFly(bool apply, bool packetOnly /*= false*/)
     if (!packetOnly && !Unit::SetCanFly(apply))
         return false;
 
+    if (!apply)
+        SetFallInformation(0, GetPositionZ());
+
     WorldPacket data(apply ? SMSG_MOVE_SET_CAN_FLY : SMSG_MOVE_UNSET_CAN_FLY, 12);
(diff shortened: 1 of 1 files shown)


Shauren:
I'm liking this idea but I would like you to change two things
* Codestyle - don't wrap the if in one line and dont wrap single statements in brackets
* Use SetFallInformation(0, GetPositionZ()); instead of assigning m_lastFallZ directly (you aren't resetting fall time variable properly)

innermous:
@Shauren i was thinking about SetFallInformation(0, GetPositionZ()), but we *already* set fallTime to 0 in Unit::SetCanFly, so we will touch this variable two times.
But if you think it's better idea - here is updated diff.

Shauren:
Unit::SetCanFly sets that value in movementInfo.fallTime field, SetFallInformation modifies m_lastFallTime (member of Player class)

Issue #16959, opened 2016-04-14
[Quest] Fields of Grief: Captured Scarlet Zealot poisoning


adeutscher:
**Description**: 

Captured Scarlet Zealot is not dying as expected after turning in "Fields of Grief".

**Current behaviour**:

"Captured Scarlet Zealot" transforms into a ghoul after turning in "Fields of Grief" (quest 407). He runs around, and says a few lines.

The Zealot does not transform back into a human, and will stay in place until a GM intervenes or the area gets otherwise de-loaded.

**Expected behaviour**:

In addition to the current behavior, the Zealot dies after he is done running around, allowing other players who turn in the quest to watch the same event once the Zealot respawns.

**Steps to reproduce the problem**:

1. Complete Fields of Grief (quest 407).

**Branch(es)**: 335

**TC hash/commit**: b756962

**TDB version**: 60

**Operating system**: Linux

adeutscher:
To confirm from the PR that I mistakenly filed, I'd like to replace the Zealot's final action with a SMART_ACTION_KILL_UNIT event, since this is the only place that Death Touch seems to be cast in `smart_scripts`:

```sql
UPDATE `smart_scripts` SET `action_type`=51, `action_param1`=0 WHERE `entryorguid`=193100 AND `id`=7;
```

tkrokli:
:-) Nice, thanks for posting this issue and solution. I noticed some time ago that the zealot turned into a ghoul did not die as expected, but I never took the time to check what could be causing it. I always play alone and usually don't have multiple characters online at the same time doing the same quests.

Rushor:
hm any reason why spell "Death Touch" is not working in script9 ?

tkrokli:
I have not re-tested this yet, not sure how to debug that script. I will empty my DB completely and test again. Could there be some kind of condition or parameter blocking the execution of that command?

adeutscher:
The cause of the issue seems to be that Death Touch is marked as a cheat spell. If a player tries to cast it without GM mode being on, they'll get an error that they're not 

The check is done in `Spell.cpp` at line 4712:
```cpp
    if (m_spellInfo->HasAttribute(SPELL_ATTR7_IS_CHEAT_SPELL) && !m_caster->HasFlag(UNIT_FIELD_FLAGS_2, UNIT_FLAG2_ALLOW_CHEAT_SPELLS))
    {
        m_customError = SPELL_CUSTOM_ERROR_GM_ONLY;
[...]

PR #16960, opened 2016-04-14
DB/Quest: Fields of Grief: Captured Scarlet Zealot should die after b…

adeutscher:
**Changes proposed**:

The issue that I'm submitting a fix for affects the "Fields of Grief" quest. When the "Captured Scarlet Zealot" (entry 1931, guid 44548) is fed a pumpkin for "Fields of Grief" (Quest no. 407), he is supposted to become a ghoul, wander around, and then die. He becomes a ghoul as intended, but does not die. The Zealot will not respawn on account of his own scripting, requiring a  GM intervention.

Steps to replicate:
- Complete Fields of Grief (Quest 407)

**Target branch(es)**: 335 

**Issues addressed**:  #16959 

The problem seems to be linked to the extra protections around Death Touch (spell 5) that prevent non-GMs from using it. To confirm this problem, instead try setting the spell to be cast to something else. For a laugh, I replaced the spell cast with the shrinking "Transporter Malfunction" for my proof-of-concept and had the spell successfully cast: 

``` sql
UPDATE `smart_scripts` SET `action_param1`=36893 WHERE `entryorguid`=193100 AND `id`=7;
```

This is the one and only entry in `smart_scripts` that tries to cast Death Touch, so for a fix I'd like to adjust the script to instead perform a SMART_ACTION_KILL_UNIT action instead of writing in a zany edge cast to the cheat spell check.

For an actual fix:

``` sql
UPDATE `smart_scripts` SET `action_type`=51, `action_param1`=0 WHERE `entryorguid`=193100 AND `id`=7;
```

### DIFF
--- a/sql/updates/world/2016_04_14_00_world.sql
+++ b/sql/updates/world/2016_04_14_00_world.sql
@@ -1,0 +1,2 @@
+--
+UPDATE `smart_scripts` SET `action_type`=51, `action_param1`=0 WHERE `entryorguid`=193100 AND `id`=7;


Aokromes:
Thx but we don't use PRs for sql only fixes.

adeutscher:
Doh, sorry about the hiccup. For future SQL nudges, should I just bring them up in the issue thread?

Aokromes:
Yes.

PR #16978, opened 2016-04-15
Core/Instance: If a boss kills itself, still save the group to the instance

### LINKED ISSUE #16971: Boss dead by self inflicted damage don't trigger instance bind
**Description**:  If a boss dies from self inflicted damage, he won't trigger instance binding to the players

**Current behaviour**: Kill Baron Geddon, at 2%health he will self kill, no instance binding is triggered

**Expected behaviour**: Killing him should bind players to the instance.

1. .go crea id 12056
2.  Kill Baron Geddon
3. Note that you won't be saved to instance

**Branch(es)**: 335

**TC hash/commit**: e38e1c0c549c

**TDB version**: 335.61

**Operating system**: Windows

**Linking model**: dynamic

myran2:
**Changes proposed**:
- Before this change, if an instance boss kills itself, the instance group will not be saved to the instance.
- This change fixes this issue by adding a check to Unit::Kill

**Target branch(es)**: 335/6x

**Issues addressed**: Fixes #16971

Builds, Tested by sirikfoll

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13772,12 +13772,19 @@ void Unit::Kill(Unit* victim, bool durabilityLoss)
             Player* creditedPlayer = GetCharmerOrOwnerPlayerOrPlayerItself();
             /// @todo do instance binding anyway if the charmer/owner is offline
 
-            if (instanceMap->IsDungeon() && creditedPlayer)
+            if (instanceMap->IsDungeon() && (creditedPlayer || this == victim))
             {
                 if (instanceMap->IsRaidOrHeroicDungeon())
                 {
                     if (creature->GetCreatureTemplate()->flags_extra & CREATURE_FLAG_EXTRA_INSTANCE_BIND)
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -13772,20 +13772,28 @@ void Unit::Kill(Unit* victim, bool durabilityLoss)
...
+            if (instanceMap->IsDungeon() && (creditedPlayer || this == victim))
             {
                 if (instanceMap->IsRaidOrHeroicDungeon())
                 {
                     if (creature->GetCreatureTemplate()->flags_extra & CREATURE_FLAG_EXTRA_INSTANCE_BIND)
-                        ((InstanceMap*)instanceMap)->PermBindAllPlayers(creditedPlayer);
+                    {
+                        // if the boss killed itself we still need to bind players to the instance
+                        if (!creditedPlayer)
+                            creditedPlayer = ((InstanceMap*)instanceMap)->GetPlayers().getFirst()->GetSource();
```
Shauren: What if the instance becomes empty? Creatures can still update in grid if they are set to active

PR #17024, opened 2016-04-24
Core/Misc: Move STOP_TIME_FOR_PLAYER value to config.

r4d1sh:


src/server/worldserver/worldserver.conf.dist
```diff
@@ -1658,6 +1658,14 @@ ListenRange.TextEmote = 40
 ListenRange.Yell = 300
 
 #
+#    Creature.MovingStopTimeForPlayer
+#        Description: Time (in milliseconds) during which creature will not moving after
+#                     interaction with player.
```
DDuarte: "will not move" instead of "will not moving"

Issue #17019, opened 2016-04-23
[Instance] Dark Iron Land Mines


adeutscher:
**Description**:

Dark Iron Land Mines (`creature_template` entry 8035) in Gnomeregan are acting strangely.

**Current behaviour**:

* When Dark Iron Land Mines a 5 seconds away from arming, they play the Avatar of Hakkar's aggro sound ("Die, mortals!"). While I think that this hilarious given how much those things can hurt in groups, it isn't exactly Blizz-like.
* Dark Iron Landmines

**Expected behaviour**:

* Land mines should play a more appropriate sound effect. Off the top of my head, they play the same sound effect as "Taunt".
* Land mines should not move.

**Steps to reproduce the problem**:

1. Go to the tunnel in Gnomeregan before the Final Chamber
2. To observe the misplaced sound, stay in the area long enough for the Dark Iron Agents to refresh their mines.
3. To observe the moving land mines, aggro a Dark Iron Agent or find one that likes to move around a lot while passive.

**Branch(es)**: 335/6x  

**TC hash/commit**:  b91bed0

**TDB version**:  335.61

**Operating system**:  Linux

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you . Thanks.

adeutscher:
To address the sound problem (setting the same sound effect as Taunt for the moment):
```mysql
UPDATE `creature_text` SET `sound`=3216 WHERE `entry` = 8035 AND `groupid` = 1 AND `id` = 0;
```

I'll investigate the movement issue more in my next session, but I have made some observations:
* The land mines were missing UNIT_FLAG_DISABLE_MOVE. Adding this flag successfully made some manually added mines on GM Island stay still when they originally did not, but the "natural" Gnomeregan ones still follow their summoner. Aside from the database problem, I think that there is also a hiccup with an NPC's follow or pet behaviour ignoring this flag.
* `.npc info` reports that the mines still have their flag, so it isn't getting outright removed (unless `.npc info` wouldn't catch onto per-spawn changes?)

To add `UNIT_FLAG_DISABLE_MOVE` (partial fix, but no observable effect in the wild):
```mysql
UPDATE creature_template set `unit_flags`= `unit_flags` | 4 WHERE `Entry`=8035;
```

adeutscher:
Updated original post to include a commit hash.

[...]

Issue #15651, opened 2015-10-03
[3.3.5] "Earthgrab Totem" chases target


Treeston:
Branch 3.3.5, rev 774ea0b4bd1e0d469cfce8013f2e557f030c7c97, TDB 335.59

Creature-summoned [Earthgrab Totem](http://www.wowhead.com/npc=6066/earthgrab-totem) will chase its target, when it should be immobile (as it's a totem, duh).

P-Kito:
uhm... really? I cant believe :D

Killyana:
If the npc enter combat when it's rooted it will enter in evade mode: https://github.com/TrinityCore/TrinityCore/issues/15304

tkrokli:
Thanks for the link, Killyana. I knew I had seen this problem mentioned before, but could not find it by searching only for "Earthgrab Totem", because it is reported as Mosh'ogg Witch Doctor & Ward of Laze.

Killyana:
@Treeston can you provide the npc that summon this totem?

Treeston:
One type of thistleshrubs in Tanaris (I believe it is [Gnarled Thistleshrub](http://www.wowhead.com/npc=5490/gnarled-thistleshrub)), and I also vaguely remember seeing it in Stranglethorn somewhere.

dr-j:
This arisen because of   http://www.wowhead.com/npc=670/skullsplitter-witch-doctor   a7fcd11a0/sql/old/6.x/world/01_2015_03_21/2015_07_03_05_world_2015_05_26_06.sql#L94 which fixes it for above case and since there is only one entry for the npc 'Earthgrab Totem' probably fixes all.

~~~sql
DELETE FROM `smart_scripts` WHERE  `entryorguid`=6066;
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES 
[...]

Issue #15304, opened 2015-08-18
[3.3.5] (Ward of Laze) Mosh'ogg Witch Doctor totems not stationary and missing aura


Funkeln:
rev https://github.com/TrinityCore/TrinityCore/commit/30fc4f9f63b7a0ce01e5aad4667902991b70e298

http://www.wowhead.com/npc=2667 is not stationary and chases players. Should be stationary as it's a totem. Also, it does not cast http://www.wowhead.com/spell=3826/ward-of-laze-effect (Ward of Laze effect) on nearby players

They're summoned by http://www.wowhead.com/npc=1144/moshogg-witch-doctor

tkrokli:
````sql
UPDATE `creature_template` SET `speed_walk`= 0, `speed_run`= 0, `unit_flags`= 4, `spell1`= 3826 WHERE `entry`= 2667;

Aokromes:
speed 0 is hack.

tkrokli:
OK, I'll remove the speed settings and let `unit_flags`= 4 `UNIT_FLAG_DISABLE_MOVE` be the fix.

Killyana:
Better to do it with SAI: just summoned - desable movement  and IC - cast 3825 every 3 secs
Also spell 1, 2, 3 ect.. are used for spell bars like vehicles and pets.

tkrokli:
WIP :
```sql
UPDATE `creature_template` SET `unit_flags`= 4, `AIName`= 'SmartAI' WHERE `entry`= 2667;

-- Ward of Laze SAI
-- (need to continue this tomorrow)

ShinDarth:
@tkrokli  You're welcome! And thank you for the fix!

tkrokli:
@Killyana : does my SAI look OK now, or does it need more adjustments?
@Funkeln : can you test this script for me, I don't have my computer with me today.

Killyana:
The npc has the extraflag 2 so it will never attack
The timer is in ms so 3 must 3000
The unitflag 4 doesn't work on summoned units that follow the summoner
If you use the action SMART_ACTION_SET_ROOT instead of set_unitflag and remove the extra flag, the npc will enter in evade on aggro, as it can't reach the target.
I got a similar issue for a quest and no idea how to fix it.

tkrokli:
New attempt based on the feedback, using `SMART_ACTION_SET_ROOT` and no flags.
```sql
[...]

PR #17040, opened 2016-04-26
[#17019] Targeted movement is now ignored by units with UNIT_FLAG_DIS…

adeutscher:
**Changes proposed**:
- Units with UNIT_FLAG_DISABLE_MOVE should not be subject to targeted movement via TargetedMovementGenerator.

**Target branch(es)**: 335, 6.x (?, see note)

**Issues addressed**: Partial progress towards #17019, though database adjustments are still required (see issue).

This should also make it easier to solve similar cases where a creature summons another creature that is intended to stay in one place. Some similar "Summoned X moves but shouldn't" issues that I could find:
- #15304
- #15651

**Tests performed**:

To add UNIT_FLAG_DISABLE_MOVE to an entry without interfering with any other flags (replace FOO):

``` sql
UPDATE creature_template SET `unit_flags`=`unit_flags` | 4 WHERE entry=@FOO;
```
- With the database adjustments brought up in #17019, teleport to Gnomeregan and fight a Dark Iron Agent (entry 6212) inside of the instance. The Dark Iron Land Mines should **NOT** run to meet you.
- Add a `unit_flag` value of 4 to Earthgrab Totem (NPC entry 6066, brought up in #15651), and fight a Thistleshrub Rootshaper in Tanaris. The Earthgrab Totem should **NOT** chase you.

- If 6.x has a similar problem, this should also be applied to 6.x. However, I do not have a 6.x server or client to test with.


Killyana:
Here some sniffs, it looks like the UNIT_FLAG_DISABLE_MOVE is not used
https://gist.github.com/Killyana/1344e0838be054427f78c84413dc621d

Shauren:
```[3] Movement Flags: Root (2048)```
As expected.

**Comment by Killyana** [comment]
So this Movement flag must be implemented?

**Comment by adeutscher** [comment]
While this PR is open, I'd like to confirm that people like this solution the best as opposed to any alternative and resolve any doubts that came from giving this another look.

My understanding of the problem is it's better to cut problem cases off when they're getting their location, since movement generators seem to have more double-checks in their `DoInitialize` or `_setTargetLocation` (e.g. "is the mover alive?") instead of in `MotionMaster::UpdateMotion`, which just checks for states like `UNIT_STATE_ROOT` or `UNIT_STATE_STUNNED`. Is this the correct approach?

If this change is accepted, I'd also like to apply the same to the other movement generators, since a unit with a "Disable Move" flag should never move at all. Should that similar of a change be made in separate PRs, since this change was originally made for a specific problem (Dark Iron Land Mines), or should the PR be renamed and expanded upon since it is essentially the same pair of lines each time?

**Comment by Aokromes** [comment]
@adeutscher 

**Comment by adeutscher** [comment]
Sorry about that. I'll look into the conflicts this weekend.

**Comment by adeutscher** [comment]
Doh, or right now!

**Comment by Shauren** [comment]
This flag is not meant to be used for generic "disable movement", its related to movement behaving like player or like creature (for example, applying it allows the player to climb steeper hills/wallclimb)

**Comment by adeutscher** [comment]
Could I confirm the location where this is used for steepness/wallclimbing? If this is the case, then the wiki also and good number of other places might also need to be corrected and/or clarified.

The comment field for `UNIT_FLAG_DISABLE_MOVE` in the [creature_template](https://trinitycore.atlassian.net/wiki/display/tc/creature_template) page outright says "Prevents the creature from moving at all", dating back to v125 of the page in January of 2016. I mostly based my decision to use this flag based on this comment.

The flag is also used in a few different places in 3.3.5 that suggest to me that it's already being used for disabling movement (code lines not directly involving `UNIT_FLAG_DISABLE_MOVE` are for context).

In the Player and Unit classes, it mostly seems to be used when dealing with flight paths or possession (the possessor gets the flag set for them).

For Kel'thuzad (initial aggro?):
```cpp
Talk(SAY_SUMMON_MINIONS);
me->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_NON_ATTACKABLE | UNIT_FLAG
_DISABLE_MOVE | UNIT_FLAG_NOT_SELECTABLE);
```

For Ionar in Halls of Lightning when dispersing:
```cpp
me->AttackStop();
me->SetVisible(false);
me->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_NON_ATTACKAB
LE|UNIT_FLAG_NOT_SELECTABLE|UNIT_FLAG_DISABLE_MOVE);
```
Ionar resetting:
```cpp
lSparkList.DespawnAll();
Initialize();
me->RemoveFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_NON_ATTACKABL
E|UNIT_FLAG_NOT_SELECTABLE|UNIT_FLAG_DISABLE_MOVE);
```

Hodir's Flash Freeze blocks:
```cpp
struct npc_flash_freezeAI : public ScriptedAI
    {
        npc_flash_freezeAI(Creature* creature) : ScriptedAI(creature)
        {
            Initialize();
            instance = me->GetInstanceScript();
           me->SetDisplayId(me->GetCreatureTemplate()->Modelid2);
           me->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_DISABLE_MOVE | UNIT_FL
AG_STUNNED | UNIT_FLAG_PACIFIED);
    }
```

It is also already in practice in places like `MotionMaster::MoveChase` and `MotionMaster::MoveFollow` in a way that suggests "if the flag is here, then don't bother continuing further into the method":
```cpp
void MotionMaster::MoveChase(Unit* target, float dist, float angle)
{
    if (!target || target == _owner || _owner->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_DISABLE_MOV
E))
    return;
```

In addition, the flag is also set in the database for a number of other `creature_template` entries that sound like they should not be moving, separate from my meddling with land mines:

* Argent Cannon
* Eye Stalk
* Combat Dummy
* Legion Flak Cannon
* Wintergrasp Tower Cannon
* Riven Widow Cocoon
* Rocket Strike
* VX-001 (Mimiron's phase 2)
* Toasty Fire (Hodir?)
* Snowpacked Icicle Target

Shauren:
All of them are wrong.
As for the location - reversing the client (UNIT_FIELD_FLAGS is sent to clients)

r00ty-tc:
I think the real question to be asking is, what do we call it? What is an apt name for it? Does it have a different effect when applied to a creature?

This piqued my interest though.

>(for example, applying it allows the player to climb steeper hills/wallclimb)

I wonder if this is part of the fear falling through floor/running through mountains problem that I was always stuck with when trying to fix the general Z position problem? For sure, much of it was inappropriate path generation, and the client not being too happy about that generated path. Other oddities were where on the client you seem to be "jumping" through the air. But the generated path definitely had you on ground level.

In fact, I think that server side it was never falling through the ground. But, if fear finished at a point the client "thought" it was in a mountain or underground, then the client takes over sending location and thus you're now falling.

No idea if it's the case or not. But, something worth looking at, at some point. Unless that flag is already set of course.

Shauren:
@r00ty-tc https://github.com/TrinityCore/TrinityCore/commit/b36da771857554c5bd652bf8ccc1386d4be607b0#diff-21815ce03789521b9bbf6217e3b97d64R618

r00ty-tc:
Yep, makes sense.

adeutscher:
Dang. To confirm, if `UNIT_FLAG_DISABLE_MOVE` is entirely off the table, does that make setting a rooted state via `smart_scripts` the only accepted way of disabling all movement for non-bosses?

Shauren:
Anything that ends up calling Unit::SetRooted is the way to go

adeutscher:
Will do. Going back to `UNIT_FLAG_DISABLE_MOVE`, perhaps the new name could be something like `UNIT_FLAG_IGNORE_SLOPES` or `UNIT_FLAG_DIRECT_PATHING`?

Shauren:
It's already renamed

tkrokli:
@adeutscher : https://github.com/TrinityCore/TrinityCore/pull/17040#issuecomment-222137077 (https://github.com/TrinityCore/TrinityCore/commit/b36da771857554c5bd652bf8ccc1386d4be607b0#diff-21815ce03789521b9bbf6217e3b97d64L618)

adeutscher:
Sorry about that, I should've checked the commit history before putting forward name ponderings.

If it's alright with everyone, I'd like to take a stab at tackling all of the spots where the former `DISABLE_MOVE` flag was used incorrectly, and updating them to root the subject instead. I'd be submitting the work in a new PR, of course.

PR #17041, opened 2016-04-26
Adjustment to Player::SetViewpoint to allow for Mind Vision "hops".

adeutscher:
**Changes proposed**:

Import of changes applied to resolve OregonCore/OregonCore#1212 that are outstanding in TrinityCore.
- When Mind Vision "hops" (casting on a new target while currently casting on another target) to another seer-able-unit, unit visibility should be updated so that the .
- Adjusted a direct assignment of `m_seer` to use `SetSeer`

**Target branch(es)**: 335

1. Cast Mind Vision (Rank 2, spell ID 10909) on a party member over 100-ish yards away on the same continent. The main goal is to mind vision something that is out of your normal sight range.
2. Refresh Mind Vision on the same target, or target another creature in range.

After doing these two steps, you should still see when the target of mind vision moves, and NPCs will continue to move. Without this change, players will stand or run in place, as will NPCs once they have finished the last waypoint that they told the caster's client about.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -23968,7 +23968,7 @@ void Player::SetViewpoint(WorldObject* target, bool apply)
         // farsight dynobj or puppet may be very far away
         UpdateVisibilityOf(target);
 
-        if (target->isType(TYPEMASK_UNIT | TYPEMASK_PLAYER) && target != GetVehicleBase())
+        if (target->isType(TYPEMASK_UNIT) && target != GetVehicleBase())
         {
             static_cast<Unit*>(target)->AddPlayerToVision(this);
             SetSeer(target);
@@ -23984,7 +23984,7 @@ void Player::SetViewpoint(WorldObject* target, bool apply)
             return;
         }
 
-        if (target->isType(TYPEMASK_UNIT | TYPEMASK_PLAYER) && target != GetVehicleBase())
+        if (target->isType(TYPEMASK_UNIT) && target != GetVehicleBase())
             static_cast<Unit*>(target)->RemovePlayerFromVision(this);
 
         //must immediately set seer back otherwise may crash
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -23968,7 +23968,7 @@ void Player::SetViewpoint(WorldObject* target, bool apply)
         // farsight dynobj or puppet may be very far away
         UpdateVisibilityOf(target);
 
-        if (target->isType(TYPEMASK_SEER) && target != GetVehicleBase())
+        if (target->isType(TYPEMASK_UNIT | TYPEMASK_PLAYER) && target != GetVehicleBase())
         {
             static_cast<Unit*>(target)->AddPlayerToVision(this);
             SetSeer(target);
@@ -23984,7 +23984,7 @@ void Player::SetViewpoint(WorldObject* target, bool apply)
             return;
         }
 
-        if (target->isType(TYPEMASK_UNIT) && target != GetVehicleBase())
+        if (target->isType(TYPEMASK_UNIT | TYPEMASK_PLAYER) && target != GetVehicleBase())
             static_cast<Unit*>(target)->RemovePlayerFromVision(this);
 
         //must immediately set seer back otherwise may crash
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -23969,10 +23969,8 @@ void Player::SetViewpoint(WorldObject* target, bool apply)
         UpdateVisibilityOf(target);
 
         if (target->isType(TYPEMASK_UNIT) && target != GetVehicleBase())
-        {
             static_cast<Unit*>(target)->AddPlayerToVision(this);
-            SetSeer(target);
-        }
+        SetSeer(target);
     }
     else
     {
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -23968,8 +23968,11 @@ void Player::SetViewpoint(WorldObject* target, bool apply)
         // farsight dynobj or puppet may be very far away
         UpdateVisibilityOf(target);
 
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -23968,8 +23968,11 @@ void Player::SetViewpoint(WorldObject* target, bool apply)
         // farsight dynobj or puppet may be very far away
         UpdateVisibilityOf(target);
 
-        if (target->isType(TYPEMASK_UNIT) && target != GetVehicleBase())
+        if (target->isType(TYPEMASK_SEER) && target != GetVehicleBase())
```
Shauren: This change is not valid. TYPEMASK_SEER includes TYPEMASK_DYNAMICOBJECT but the code uses a static_cast<Unit*>. DynamicObject is not Unit and will crash.

Goatform: Hell no. I agree with @Shauren.

adeutscher: Other than my change and its definition, `TYPEMASK_SEER` is only used at one point in `Player::GetViewpoint`. Is there any situation that would set the player's view to a dynamic object such that `Player::GetViewpoint` would return a dynamic object? If setting a viewpoint to a dynamic object would be a server-crasher. If it is not the case that `Player::GetViewpoint` would ever return a dynamic object, I'd like to remove `TYPEMASK_DYNAMICOBJECT` from `TYPEMASK_SEER`.

adeutscher: If adjusting `TYPEMASK_SEER` is not an option, then I will change the line being discussed to say "unit or player".

Shauren: Shaman farsight is a case for DynamicObject

adeutscher: Will do, I've pushed a new commit so that `SetViewpoint` will only call `SetSeer` and `AddPlayerToVision` for objects that specifically have `TYPEMASK_UNIT` or `TYPEMASK_PLAYER`. I also updated the "removed seer" case so that `RemovePlayerFromVision` is 1-to-1 with the number of times that `AddPlayerToVision` is called.

Goatform:
Squash commits please.

Shauren:
@Goatform not needed anymore, github has a new feature that lets us squash it ourselves during merge

Goatform:
Oh, I didn't noticed that actually. Anyway, thanks for info!

Naios:
Seems legit for me, anyone against merging this PR? (@Shauren)

tkrokli:
It would be nice to have the label [Feedback-FixOutdatedMissingWIP] replaced with the label
[Feedback-PatchFix] before merging to see that it also was evaluated properly before merging.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -23981,11 +23984,11 @@ void Player::SetViewpoint(WorldObject* target, bool apply)
             return;
         }
 
-        if (target->isType(TYPEMASK_UNIT) && target != GetVehicleBase())
+        if (target->isType(TYPEMASK_UNIT | TYPEMASK_PLAYER) && target != GetVehicleBase())
```
Shauren: Players also have TYPEMASK_UNIT flag set so this change isn't needed

src/server/game/Entities/Player/Player.cpp
```diff
@@ -23968,8 +23968,11 @@ void Player::SetViewpoint(WorldObject* target, bool apply)
         // farsight dynobj or puppet may be very far away
         UpdateVisibilityOf(target);
 
-        if (target->isType(TYPEMASK_UNIT) && target != GetVehicleBase())
+        if (target->isType(TYPEMASK_UNIT | TYPEMASK_PLAYER) && target != GetVehicleBase())
```
Shauren: Players also have TYPEMASK_UNIT flag set so this change isn't needed

adeutscher:
Well, I feel incredibly foolish considering the TYPEMASK debate earlier. Making a new commit to revert this line.

Shauren:
One more thing that escaped my attention before
Why is SetSeer(target) called only if target is unit and not vehicle in apply case but always called in unapply case?

adeutscher:
In the unapply case, the target is `this` instead of `target`, so I assumed that it was because the unit getting their seer reset was trusted to be a reliable player target that didn't need any double-checking.

The `SetSeer(this)` line change was mostly housekeeping since every other place seemed to not directly assign to `m_seer` instead of using `SetSeer`. `SetSeer` is a glorified assignment to `m_seer`, but in the off-chance `SetSeer` needs to be expanded upon for some reason in the future there won't be a loose end that's still working manually.

Shauren:
Since you know exactly how to reproduce it and what behavior to look for, can you tell what happens when a shaman chains Far Sight like that?

adeutscher:
When I cast Far Sight over a long range, I seem to be getting a freeze-frame of the current view. NPC's and players will stay in place, whether on the first cast or later cast (though like with Mind Vision, NPCs will complete their current waypoint). Their position will be updated when I re-cast, though I think this is only because the client forgets about what it knew about the target area when vision hops back to the caster for a re-cast. This happens both with and without my change (expected since the typemask is only for Units, not Dynamic Objects).

I'm sorry to say that I've never tested this through Far Sight before, as I only learned of this flavour of issue through OregonCore and a Mind Vision issued, so I'd filed Far Sight away as working. OregonCore's Far Sight looks to be broken in the same way.

Results of a few mock builds:

* AddPlayerToVision crashes TC, as feared above
* Placing `SetSeer` outside of the TYPEMASK_UNIT check will not crash anything (for Mind Vision or Far Sight), but Far Sight troubles remain.

Other than crashing the server, I was not able to make Far Sight work differently with my initial tinkering. For this PR, do you want me to move SetSeer outside of the `if`, with Far Sight's problems to be filed as a separate issue?

Shauren:
I would like it moved outside the if, to avoid any future confusion for other developers reading the code as to why applying doesn't always set the seer

Aokromes:

Aokromes:
If there is no reason to merge this, i will merge it on 24h.

PR #17050, opened 2016-04-27
[6.x] Core/Spell: Jade Lightning (117952 & 117959)

Infamous-devel:
**Changes proposed**:
- Jade Lightning - (www.wowhead.com/spell=117952)
- Jade Lightning - (www.wowhead.com/spell=117959)

**Target branch(es)**: 6x

### DIFF
--- a/src/server/scripts/Spells/spell_monk.cpp
+++ b/src/server/scripts/Spells/spell_monk.cpp
@@ -59,9 +59,8 @@ class spell_monk_crackling_jade_lightning : public SpellScriptLoader
         {
             if (GetCaster())
             {
-                GetCaster()->CastSpell(GetCaster(), SPELL_MONK_CRACKLING_JADE_LIGHTNING_AURA, true);
-                GetCaster()->CastSpell(GetCaster(), SPELL_MONK_CRACKLING_JADE_LIGHTNING_AURA, true, NULL, aurEff);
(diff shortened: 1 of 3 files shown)


src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -0,0 +1,143 @@
...
+                return false;
+            return true;
+        }
+
+        void OnApply(AuraEffect const* aurEff, AuraEffectHandleModes /*mode*/)
+        {
+            if (GetCaster())
+            {
+                GetCaster()->CastSpell(GetCaster(), SPELL_MONK_CRACKLING_JADE_LIGHTNING_AURA, true);
+                GetCaster()->CastSpell(GetCaster(), SPELL_MONK_CRACKLING_JADE_LIGHTNING_AURA, true, NULL, aurEff);
```
Shauren: Casting the same spell twice?

src/server/scripts/Spells/spell_script_loader.cpp
```diff
@@ -1,52 +1,54 @@
-/*
```
Shauren: ooks like you pushed the file with CR+LF line endings instead of expected LF so it shows up as changing the entire file - this needs to be fixed

Infamous-devel: Not sure how to revert that and only change the lines I edited,

PR #17078, opened 2016-05-04
Core/Guilds: Fix Guild::HandleSetMemberNote logic according to method…

r4d1sh:
… declaration.

This change has no effect on server behavior.


Shauren:
It would have been easier to just rename the argument in declaration... https://github.com/Tonghost/TrinityCore/blob/3.3.5/src/server/game/Guilds/Guild.h#L675

Tonghost:
@Shauren I know. I just prefer the current name of the argument.

Aokromes:

Tonghost:
:shipit:

PR #17113, opened 2016-05-11
Core/Auras: Implement SPELL_AURA_MOD_SPELL_COOLDOWN_BY_HASTE

efectokoleos:
**Changes proposed**:
- Implement aura SPELL_AURA_MOD_SPELL_COOLDOWN_BY_HASTE

**Target branch(es)**: 6x

**Tests performed**: Tested and working

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12460,6 +12460,7 @@ Unit* Unit::SelectNearbyTarget(Unit* exclude, float dist) const
 void Unit::ApplyAttackTimePercentMod(WeaponAttackType att, float val, bool apply)
 {
     float remainingTimePct = (float)m_attackTimer[att] / (GetAttackTime(att) * m_modAttackSpeedPct[att]);
+
     if (val > 0)
     {
         ApplyPercentModFloatVar(m_modAttackSpeedPct[att], val, !apply);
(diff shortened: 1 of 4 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12468,7 +12469,18 @@ void Unit::ApplyAttackTimePercentMod(WeaponAttackType att, float val, bool apply
         if (GetTypeId() == TYPEID_PLAYER)
         {
             if (att == BASE_ATTACK)
+            {
                 ApplyPercentModFloatValue(UNIT_FIELD_MOD_HASTE, val, !apply);
+
+                const Unit::AuraEffectList list = GetAuraEffectsByType(SPELL_AURA_MOD_SPELL_COOLDOWN_BY_HASTE);
+
+                for (Unit::AuraEffectList::const_iterator itr = list.begin(); itr != list.end(); ++itr)
```
Naios: You can simplify the two lines by using:
```c++
for (auto& effect : GetAuraEffectsByType(SPELL_AURA_MOD_SPELL_COOLDOWN_BY_HASTE))
```

Shauren: We don't use auto for non-iterator types, also AuraEffectList is a std::list<AuraEffect*>, there is no need to have a reference to pointer there

Naios:
Looks good to me, anyone could test it?

Aokromes:

Naios:
No, I'll merge this later today.

Shauren:
I just looked at how this aura is handled by the client and checked sniffs - your implementation is wrong.
No spellmod is used on retail for this (in fact you actually break it *clientside* by making it apply haste TWICE)

PR #17133, opened 2016-05-14
Core/Unit: Pet shouldn't be notified for DOT damage on the owner 

### LINKED ISSUE #16894: [335] Core/Pet: DoT Damage Causes Pet Aggro Per Tick (Patch included)
**Description**:

Each tick of a DoT on the Hunter causes the pet to aggro

**Current behaviour**:

When a mob casts a DoT on the Hunter and you then run away to exit combat, each DoT tick will put the pet in combat and if it is on defensive it will run to attack the original caster.

**Expected behaviour**:

The pet should only aggro onto the caster on the inital cast. Subsequent damage from DoT ticks should be ignored by the pet and not put it back into combat.

1. Get a Hunter
2. Get a pet, put it on defensive
3. Find something to cast a DoT on the Hunter
4. Put the pet on passive and run away to exit combat (without killing the mob)
5. Put the pet back on defensive after combat drops
6. Observe the pet as it tries to chase down the caster each time the DoT damges the Hunter (hint: click 'follow' when it begins to chase).

**Branch(es)**: 335

**TC hash/commit**: cebdf03349ed0c50a40cf52ddae887cf666d1eb5 and https://github.com/TrinityCore/TrinityCore/commit/974703671ed5f08ec0c1a5b3943fc68508f2f4d6

**TDB version**: TDB_full_world_335.60_2015_11_07.sql

**Operating system**: Win XP / 7

mik1893:
**Changes proposed**:
- The pet should only aggro onto the caster on the inital cast. Subsequent damage from DoT ticks should be ignored by the pet and not put it back into combat

**Target branch(es)**: 335

**Issues addressed**: Fixes #16894

**Tests performed**: tested in game

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -599,11 +599,14 @@ uint32 Unit::DealDamage(Unit* victim, uint32 damage, CleanDamage const* cleanDam
 
     if (victim->GetTypeId() == TYPEID_PLAYER && this != victim)
     {
-        // Signal to pets that their owner was attacked
-        Pet* pet = victim->ToPlayer()->GetPet();
+        // Signal to pets that their owner was attacked - except when DOT.
+        if (damagetype != DOT)
+        {
+            Pet* pet = victim->ToPlayer()->GetPet();
 
-        if (pet && pet->IsAlive())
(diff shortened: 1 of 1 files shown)


MrSmite:
Thank you.

PR #17144, opened 2016-05-17
Issue mind control flag

### LINKED ISSUE #17143: [3.3.5][Priest] Core/Spell: Mind Control not flagging PvP
**Description**: Mind Control does not flag the caster for PvP when used against a PvP-flagged target. This makes for some very fun shenanigans in city sieges, but definitely isn't intended.

**Current behaviour**:

A player can cast Mind Control without being flagged for PvP. This only seems to apply to successful casts, as misses still flag the user for PvP.

**Expected behaviour**: 

A player will be flagged for PvP if they cast Mind Control on a flagged target.

1. Make a character, and teach them Mind Control (Id: 605).
2. Cast Mind Control on a flagged target while ***not*** PvP flagged.

**Branch(es)**: 335

**TC hash/commit**:  b91bed0

**TDB version**:  335.61

**Operating system**:  Linux

I currently have a crude fix from copying the `UpdatePvP` call in `DoAllEffectOnTarget` to `AuraEffect::HandleModPossess`, but I am still looking into how to do this earlier in the chain. Not entirely happy with the current fix since PvP flagging isn't really dealt with anywhere else in `SpellAuraEffects.cpp`, so there must be a better way.

```cpp
// if target is flagged for pvp also flag caster if a player
if (target->IsPvP() && caster->GetTypeId() == TYPEID_PLAYER)
    caster->ToPlayer()->UpdatePvP(true);
```

adeutscher:
**Changes proposed**:
- Spells such as Mind Control should flag the caster for PvP when cast on a PvP-flagged target.

**Target branch(es)**: 335

**Issues addressed**: Fixes #17143 

**Tests performed**: 
1. Made a character, and taught them Mind Control (Id: 605).
2. Cast Mind Control on a flagged target while **_not**_ PvP flagged. Tested against flagged players in capital city (e.g. Alliance target in Orgrimmar).


adeutscher:
Oh dear. Accidentally tried to suggest merging 3.3.5 into 6.x.

PR #17145, opened 2016-05-17
[#17143] Make certain spells (e.g. Mind Control) properly flag the ca…

### LINKED ISSUE #17143: [3.3.5][Priest] Core/Spell: Mind Control not flagging PvP

adeutscher:
Re-submission of #17144  against the proper target branch. Whoops.

**Changes proposed**:
- Spells such as Mind Control should flag the caster for PvP when cast on a PvP-flagged target.

**Target branch(es)**: 335

**Issues addressed**: Fixes #17143 

**Tests performed**: 
1. Made a character, and taught them Mind Control (Id: 605).
2. Cast Mind Control on a flagged target while **_not**_ PvP flagged. Tested against flagged players in capital city (e.g. Alliance target in Orgrimmar).

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2294,8 +2294,13 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
         }
     }
 
+    bool enablePvP = false; // need to check PvP state before spell effects, but act on it afterwards
+
     if (spellHitTarget)
     {
+        // if target is flagged for pvp also flag caster if a player
+        if (unit->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
+            enablePvP = true; // Decide on PvP flagging now, but act on it later.
         SpellMissInfo missInfo2 = DoSpellHitOnUnit(spellHitTarget, mask, target->scaleAura);
         if (missInfo2 != SPELL_MISS_NONE)
         {
@@ -2448,8 +2453,7 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
         // Needs to be called after dealing damage/healing to not remove breaking on damage auras
         DoTriggersOnSpellHit(spellHitTarget, mask);
 
-        // if target is flagged for pvp also flag caster if a player
-        if (unit->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
+        if (enablePvP)
             m_caster->ToPlayer()->UpdatePvP(true);
 
         CallScriptAfterHitHandlers();
@@ -2509,10 +2513,6 @@ SpellMissInfo Spell::DoSpellHitOnUnit(Unit* unit, uint32 effectMask, bool scaleA
 
     if (m_caster != unit)
     {
-        // if target is flagged for pvp also flag caster if a player
-        if (unit->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
-            m_caster->ToPlayer()->UpdatePvP(true);
-
         // Recheck  UNIT_FLAG_NON_ATTACKABLE for delayed spells
         if (m_spellInfo->Speed > 0.0f && unit->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_NON_ATTACKABLE) && unit->GetCharmerOrOwnerGUID() != m_caster->GetGUID())
             return SPELL_MISS_EVADE;
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2448,7 +2448,7 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
         // Needs to be called after dealing damage/healing to not remove breaking on damage auras
         DoTriggersOnSpellHit(spellHitTarget, mask);
 
-        // if target is fallged for pvp also flag caster if a player
+        // if target is flagged for pvp also flag caster if a player
         if (unit->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
             m_caster->ToPlayer()->UpdatePvP(true);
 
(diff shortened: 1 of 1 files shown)


tkrokli:
Nice update, just remember to add the line "Fixes #17143" to your commit message when PR is ready
(makes github auto close issue 17143 when this PR gets merged or pushed as a separate commit).

adeutscher:
Adjusted the commit message. First time squashing commits, so I think I went through a bit more kerfuffle than I needed to. Everything seems in order now, though.

Aokromes:
You don't need to squash anymore, now github allows to squash on push.

src/server/game/Spells/Spell.cpp
```diff
@@ -2448,7 +2448,7 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
         // Needs to be called after dealing damage/healing to not remove breaking on damage auras
         DoTriggersOnSpellHit(spellHitTarget, mask);
 
-        // if target is fallged for pvp also flag caster if a player
+        // if target is flagged for pvp also flag caster if a player
         if (unit->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
             m_caster->ToPlayer()->UpdatePvP(true);
```
jackpoz: any idea about why this call is not enough ?

adeutscher: This is still throwing me for a loop. I've taken another look at it this morning. Some of my observations:

In `DoAllEffectOnTarget`, `spellHitTarget` must not be NULL in order to get at the only call to `DoSpellHitOnUnit`, where my fix is. There are many calls to `HandleEffects`, but only one call to `DoSpellHitOnUnit` in the entire project.

As far as I can see, the only way that `spellHitTarget` will be reset to NULL is if the return value is **NOT** `SPELL_MISS_NONE` (~Line 2299 of Spell.cpp). But if we've gotten to the point of calling `HandleEffects` to actually make the wheels spin on the spell, then the only thing that `DoSpellHitOnUnit` can return is `SPELL_MISS_NONE` (~Line 2653 of Spell.cpp).

jackpoz: I suggest to use Step Into/Step Over debugger features, http://www.codeproject.com/KB/cs/MasteringInDebugging.aspx has some nice tips on How To Replace Printf Debugging With Proper Debugging Tools :)

adeutscher: I'm on Linux using Geany/make at the moment, so the first step for Visual Studio for me would be wine shenanigans or getting a Windows installation.

I found the core reason for the problem. I was so focused on Spell Hit that I forgot the original bug: Mind Control not flagging the caster for PvP.

`spellHitTarget` is working exactly as intended. The problem is that the following important things are happening:

1. Player casts Mind Control, eventually leading to `DoAllEffectOnTarget`.
2. `DoSpellHitOnUnit is called, and my PvP flag update copied from the bottom of `DoAllEffectOnTarget` takes effect.
3. The various effects of Mind Control are applied. This is where the problem is. After the target is charmed, the victim becomes an **unflagged** minion.
4. Exit back out to `DoAllEffectOnTarget`.
5. Continue to the PvP check block. If the spell successfully hit (note that misses and immunes still work as intended), then the target is not flagged and the method continues without flipping the caster over.

So this looks like it's much more specific of a problem than I first thought.

To resolve this, would there be any disadvantage have the original PvP check set a boolean flag in `DoAllEffectOnTarget` immediately before the call to `DoSpellHitOnUnit`, then act on it where it currently is?

```cpp
if (unit->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
    enablePvP = true;
...
if (enablePvP)
   m_caster->ToPlayer()->UpdatePvP(true);
```

Shauren:
> To resolve this, would there be any disadvantage have the original PvP check set a boolean flag in DoAllEffectOnTarget immediately before the call to DoSpellHitOnUnit, then act on it where it currently is?
Sounds good to me

tkrokli:
Sorry for bothering you with this cosmetic detail again. I just thought you would like to know what I meant.

PR title: **Core/Spell: properly flag caster for PvP if cast on a PvP-flagged target**

Commit message text:
```
Core/Spell: properly flag caster for PvP if cast on a PvP-flagged target

Make certain spells (e.g. Mind Control) properly flag the caster for PvP
if cast on a PvP-flagged target.

Fixes issue #17143
```
It is mostly a suggestion based on how the preferred TC format seems to be in most PRs. :)

src/server/game/Spells/Spell.cpp
jackpoz: it's always nice to see an issue debugged, understood and fixed without affecting any other functionality. it takes some more time but it's worth it.
thank you for the fine work and for the fix

PR #17163, opened 2016-05-21
Core/Spell: do not activate PVP flag on selfcast

mik1893:
**Changes proposed**:
- Do not activate PVP flags when a spell is a selfcast

**Target branch(es)**: 335/6x

**Issues addressed**: Fixes #9766

**Tests performed**: Builds and tested in game

- none (had this for a while running)

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2495,8 +2495,8 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
         // Needs to be called after dealing damage/healing to not remove breaking on damage auras
         DoTriggersOnSpellHit(spellHitTarget, mask);
 
-        // if target is fallged for pvp also flag caster if a player
-        if (unit->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
+        // if it's not selfcast and target is flagged for pvp also flag caster if a player
+        if (unit != m_caster && unit->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
             m_caster->ToPlayer()->UpdatePvP(true);
 
         CallScriptAfterHitHandlers();


src/server/game/Spells/Spell.cpp
```diff
@@ -2495,8 +2495,8 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
         // Needs to be called after dealing damage/healing to not remove breaking on damage auras
         DoTriggersOnSpellHit(spellHitTarget, mask);
 
-        // if target is fallged for pvp also flag caster if a player
-        if (unit->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
+        // if it's not selfcast and target is flagged for pvp also flag caster if a player
+        if (spellHitTarget != m_caster && spellHitTarget->IsPvP() && m_caster->GetTypeId() == TYPEID_PLAYER)
```
Shauren: This is wrong. `spellHitTarget` points back to m_caster if spell was reflected (but it should still flag for PvP), using `unit` was correct here.

Shauren:
I wonder how did you test it ingame if the condition can never be satisfied anyway?
It only flags the caster for pvp if target passes `IsPvP()` check - which would mean the caster is already flagged if caster = target

mik1893:
I am taking pieces of code from my old report - sometimes I copy paste the wrong commit as I don't really remember what I did 4+ years ago and I don't have the time to check. I will find the right commit and squash it

mik1893:
ehm, didnt see that the code just changed on a rev of the 18/05... will change this again ?¬

MrSmite:
This isn't the proper way to handle Shadowmeld, it is not a hostile spell and should not flag the player for combat under any circumstance.

http://www.wowhead.com/spell=58984/shadowmeld

>     Stops auto-attack
    Generates no threat

Since there is no threat generated, any nearby players or NPCs that are flagged for PvP should not cause the player to flag.

Shauren:
This will not fix it.
```
ID - 58984 Shadowmeld (Racial)
Effect 1: Id 3 (SPELL_EFFECT_DUMMY)
BasePoints = 1
Targets (18, 31) (TARGET_DEST_CASTER, TARGET_UNIT_DEST_AREA_ALLY)
Radius (Id 12) 100,00
```
This effect hits nearby friendly units which satisfies the condition unit != m_caster

@MrSmite positive spells also can flag for pvp: healing and buffs

MrSmite:
> positive spells also can flag for pvp: healing and buffs

Yes but those generate threat. This spell explicitly states it generates no threat.

> http://wowwiki.wikia.com/wiki/Threat
> Threat caused by beneficial effects is divided amongst all NPCs that are aware of the character. [...] Healing threat is global, and is normally .5x of the amount healed. Healing effects cause no threat if the target is already at full health.

Shauren:
But resurrecting does not generate threat and does flag for pvp - threat and pvp flagging are unrelated

zaphod77:
Resurrection doesn't flag you for pvp, it just doesn't remove it if you still have it. You can unflag if you wait 5 minutes while dead.

Threat DOES matter.

Shadowmeld has a dummy aura that targets every npc within 100 yards.

But because it's a no threat spell it doesn't flag you.

Making no threat spells not flag you for PVP is 100% correct.

PR #17167, opened 2016-05-22
Core/Instances: Add functionality for changing instance entrance

Krudor:
**Changes proposed**: Add the ability for instances to set new locations that players will be teleported to when entering the instance

**Target branch(es)**: 6x

**Tests performed**: Tested in-game

**Known issues and TODO list**: No known functional issues

Instances have since MoP (Possibly even Cataclysm) been able to set new zone-in locations, depending on dungeon progress.

A theoretic example of this could be after a certain boss has been
defeated, a new entrance location is set to bring players entering the
instance closer to the relevant content.
A real example would be Siege of Orgrimmar changing entrance locations on multiple occasions as the raid progresses.

Added SetEntranceLocation for setting entrance locations that will be
saved to DB the next time the instance is saved, and
SetTemporaryEntranceLocation for when you don't want the value to be
saved to the database.

### DIFF
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -620,13 +620,13 @@ void WorldSession::HandleAreaTriggerOpcode(WorldPackets::Misc::AreaTrigger& pack
 
     if (!teleported)
     {
-        const WorldSafeLocsEntry* entranceLocation = nullptr;
+        WorldSafeLocsEntry const* entranceLocation = nullptr;
         InstanceSave* instanceSave = player->GetInstanceSave(at->target_mapId);
         if (instanceSave)
         {
             // Check if we can contact the instancescript of the instance for an updated entrance location
             if (Map* map = sMapMgr->FindMap(at->target_mapId, player->GetInstanceSave(at->target_mapId)->GetInstanceId()))
-                if (InstanceScript* instanceScript = ((InstanceMap*)map)->GetInstanceScript())
+                if (InstanceScript* instanceScript = dynamic_cast<InstanceMap*>(map)->GetInstanceScript())
                     entranceLocation = sWorldSafeLocsStore.LookupEntry(instanceScript->GetEntranceLocation());
 
             // Finally check with the instancesave for an entrance location if we did not get a valid one from the instancescript
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -619,7 +619,26 @@ void WorldSession::HandleAreaTriggerOpcode(WorldPackets::Misc::AreaTrigger& pack
     }
 
     if (!teleported)
-        player->TeleportTo(at->target_mapId, at->target_X, at->target_Y, at->target_Z, at->target_Orientation, TELE_TO_NOT_LEAVE_TRANSPORT);
+    {
+        const WorldSafeLocsEntry* entranceLocation = nullptr;
+        InstanceSave* instanceSave = player->GetInstanceSave(at->target_mapId);
+        if (instanceSave)
+        {
+            // Check if we can contact the instancescript of the instance for an updated entrance location
+            if (Map* map = sMapMgr->FindMap(at->target_mapId, player->GetInstanceSave(at->target_mapId)->GetInstanceId()))
+                if (InstanceScript* instanceScript = ((InstanceMap*)map)->GetInstanceScript())
+                    entranceLocation = sWorldSafeLocsStore.LookupEntry(instanceScript->GetEntranceLocation());
+
+            // Finally check with the instancesave for an entrance location if we did not get a valid one from the instancescript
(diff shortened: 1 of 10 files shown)


Shauren:
Do these really come from WorldSafeLocs? That would be amazing

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -619,7 +619,26 @@ void WorldSession::HandleAreaTriggerOpcode(WorldPackets::Misc::AreaTrigger& pack
     }
 
     if (!teleported)
-        player->TeleportTo(at->target_mapId, at->target_X, at->target_Y, at->target_Z, at->target_Orientation, TELE_TO_NOT_LEAVE_TRANSPORT);
+    {
+        const WorldSafeLocsEntry* entranceLocation = nullptr;
```
Treeston: `WorldSafeLocsEntry const*`, code style

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -619,7 +619,26 @@ void WorldSession::HandleAreaTriggerOpcode(WorldPackets::Misc::AreaTrigger& pack
...
+        InstanceSave* instanceSave = player->GetInstanceSave(at->target_mapId);
+        if (instanceSave)
+        {
+            // Check if we can contact the instancescript of the instance for an updated entrance location
+            if (Map* map = sMapMgr->FindMap(at->target_mapId, player->GetInstanceSave(at->target_mapId)->GetInstanceId()))
+                if (InstanceScript* instanceScript = ((InstanceMap*)map)->GetInstanceScript())
```
Treeston: Code style, don't do C-style casts.

src/server/game/Instances/InstanceSaveMgr.cpp
```diff
@@ -96,6 +96,9 @@ InstanceSave* InstanceSaveManager::AddInstanceSave(uint32 mapId, uint32 instance
         return NULL;
     }
 
+    if (!sWorldSafeLocsStore.LookupEntry(entranceId))
+        TC_LOG_WARN("misc", "InstanceSaveManager::AddInstanceSave: invalid entranceId = %d defined for instance save with mapid = %d, instanceid = %d!", entranceId, mapId, instanceId);
```
Treeston: Reset to some sensible default value?

src/server/game/Instances/InstanceScript.h
```diff
@@ -246,6 +246,15 @@ class TC_GAME_API InstanceScript : public ZoneScript
...
         uint32 GetCompletedEncounterMask() const { return completedEncounters; }
 
+        // Sets the entrance location (WorldSafeLoc) id
+        void SetEntranceLocation(uint32 worldSafeLocationId);
+
+        // Sets a temporary entrance that does not get saved to db
+        void SetTemporaryEntranceLocation(uint32 worldSafeLocationId) { temporaryEntranceId = worldSafeLocationId; }
+
+        // Get's the current entrance id
+        uint32 GetEntranceLocation() const { return temporaryEntranceId > 0 ? temporaryEntranceId : entranceId; }
```
Treeston: `> 0` check is unnecessary.

src/server/game/Instances/InstanceSaveMgr.cpp
```diff
@@ -96,6 +96,9 @@ InstanceSave* InstanceSaveManager::AddInstanceSave(uint32 mapId, uint32 instance
         return NULL;
     }
 
+    if (!sWorldSafeLocsStore.LookupEntry(entranceId))
```
Shauren: Will print error if 0 is used (no entrance override)

src/server/game/Instances/InstanceScript.h
Krudor: True! :+1:

src/server/game/Instances/InstanceSaveMgr.cpp
Krudor: Indeed.

Krudor:
I've looked at WorldSafeLocs and they seem to cover all the dynamic instance locations I can think of.
I'm going to suggest another change to lfg_dungeon_template in DB to change  pos_x/y/z/o to WorldSafeLoc's as well.

src/server/game/Instances/InstanceSaveMgr.cpp
Krudor: Oh right, thanks!

Aokromes:

Krudor:
I can see a reason for changing this down the road.
There are instances that change entrance location right as the instance starts depending on for example difficulty (I'm looking at you, Challenge Mode Auchindoun).

This commit will not help achieve that purpose but it will suffice for most of the cases where the entrance changes depending on dungeon progress or other reasons.
Hopefully we can make table areatrigger_teleport even more dynamic in the future, or simply provide another alternative method for overriding teleport locations depending on various conditions.

PR #17200, opened 2016-05-28
Core/Quest: remove duplicated call to SendQuestComplete().

### LINKED ISSUE #15613: [3.3.5[ Double Quest complete Message
Core TrinityCore rev. 3f24853e29c2
TDB TDB 335.59 + all updates

Many quest give the text twice (quest complete)
All Escortquests have this Bug and many normal Quests

wowscrnshot_092715_122046

SnapperRy:
**Changes proposed**:
- Remove wrong call to SendQuestComplete() that caused a wrong "Objective complete" to pop up in the client.

**Target branch(es)**: 335.

**Issues addressed**: Fixes #15613.

**Tests performed**: Tested with quests that call AreaExploredOrEventHappens(), like 2520 and 938.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -14839,12 +14839,8 @@ void Player::CompleteQuest(uint32 quest_id)
             SetQuestSlotState(log_slot, QUEST_STATE_COMPLETE);
 
         if (Quest const* qInfo = sObjectMgr->GetQuestTemplate(quest_id))
-        {
             if (qInfo->HasFlag(QUEST_FLAGS_TRACKING))
                 RewardQuest(qInfo, 0, this, false);
-            else
-                SendQuestComplete(quest_id);
-        }
     }
 
     if (sWorld->getBoolConfig(CONFIG_QUEST_ENABLE_QUEST_TRACKER)) // check if Quest Tracker is enabled


src/server/game/Entities/Player/Player.cpp
```diff
@@ -15973,12 +15973,7 @@ void Player::AreaExploredOrEventHappens(uint32 questId)
             QuestStatusData& q_status = m_QuestStatus[questId];
 
             if (!q_status.Explored)
-            {
                 q_status.Explored = true;
-                m_QuestStatusSave[questId] = QUEST_DEFAULT_SAVE_TYPE;
```
joschiwald: this line is still required

SnapperRy: Because it won't save the completed objective to DB without it, right?

Sorry, didn't think about that.

joschiwald: yes

SnapperRy: Done, thanks.

Shauren:
This is actually wrong. On retail "Objective complete" is shown only for special quests that require custom completion like AreaExploredOrEventHappens. Regular kill X creatures or gather Y items should not show it.

You should remove it from CompleteQuest, not here

SnapperRy:
Makes sense, didn't know it was like that for other kind of quests. Done, thanks.

tkrokli:
Yay, thanks for looking into that issue and finding the correct source of the problem! I was trying to think of how to find it myself, but I get too easily distracted with everything else. Thank you for the fix!

SnapperRy:
To add another two cents (wasn't sure if making another comment would re-open the request, thanks tkrokli for confirming to me it doesn't :P): tested and working.

Also found proof that it's acting as on retail now: https://youtu.be/ohTCJ1SDI1w?t=182 (video was made before patch 3.3.0, since that quest still required 7 Young Nightsabers). I'm sure there are plenty of other (and with better quality) videos out there, but that should be enough.

Thanks again ;) wouldn't have gotten this one right without Shauren's input.

PR #17206, opened 2016-05-29
Core/Pet: Fix pet at rest distance

### LINKED ISSUE #10865: [Core / Pet] [Patch] Pets at rest are too far away
__Rev:__ All

__Bug:__

When pets are idle, they are too far away from their owner

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1412994-core-pet-patch-pets-at-rest-are-too-far-away?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Aokromes:
**Changes proposed**:
- Fix pet at rest distance

**Target branch(es)**: 335

**Issues addressed**: Closes #10865

By MrSmite, closes #10865

### DIFF
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -172,7 +172,7 @@ bool TargetedMovementGeneratorMedium<T, D>::DoUpdate(T* owner, uint32 time_diff)
         //More distance let have better performance, less distance let have more sensitive reaction at target move.
         float allowed_dist = 0.0f;
 
-        if ((owner->GetTypeId() == TYPEID_UNIT && owner->IsPet()) && (owner->GetCharmerOrOwnerGUID() == i_target->GetGUID()))
+        if (owner->IsPet() && (owner->GetCharmerOrOwnerGUID() == i_target->GetGUID()))
             allowed_dist = 1.0f; // pet following owner
         else
             allowed_dist = owner->GetCombatReach() + sWorld->getRate(RATE_TARGET_POS_RECALCULATION_RANGE);
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -73,8 +73,8 @@ void TargetedMovementGeneratorMedium<T, D>::_setTargetLocation(T* owner, bool up
             //   doing a "dance" while fighting
             if (owner->IsPet() && i_target->GetTypeId() == TYPEID_PLAYER)
             {
-                dist = i_target->GetCombatReach();
-                size = i_target->GetCombatReach() - i_target->GetObjectSize();
+                dist = 1.0f; //i_target->GetCombatReach();
+                size = 1.0f; //i_target->GetCombatReach() - i_target->GetObjectSize();
             }
             else
             {
@@ -168,8 +168,15 @@ bool TargetedMovementGeneratorMedium<T, D>::DoUpdate(T* owner, uint32 time_diff)
     if (i_recheckDistance.Passed())
     {
         i_recheckDistance.Reset(100);
+
         //More distance let have better performance, less distance let have more sensitive reaction at target move.
-        float allowed_dist = owner->GetCombatReach() + sWorld->getRate(RATE_TARGET_POS_RECALCULATION_RANGE);
+        float allowed_dist = 0.0f;
+
+        if ((owner->GetTypeId() == TYPEID_UNIT && owner->IsPet()) && (owner->GetCharmerOrOwnerGUID() == i_target->GetGUID()))
+            allowed_dist = 1.0f; // pet following owner
+        else
+            allowed_dist = owner->GetCombatReach() + sWorld->getRate(RATE_TARGET_POS_RECALCULATION_RANGE);
+
         G3D::Vector3 dest = owner->movespline->FinalDestination();
         if (owner->movespline->onTransport)
             if (TransportBase* transport = owner->GetDirectTransport())


src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -73,8 +73,8 @@ void TargetedMovementGeneratorMedium<T, D>::_setTargetLocation(T* owner, bool up
             //   doing a "dance" while fighting
             if (owner->IsPet() && i_target->GetTypeId() == TYPEID_PLAYER)
             {
-                dist = i_target->GetCombatReach();
-                size = i_target->GetCombatReach() - i_target->GetObjectSize();
+                dist = 1.0f; //i_target->GetCombatReach();
```
jackpoz: what are the original values ?

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -168,8 +168,15 @@ bool TargetedMovementGeneratorMedium<T, D>::DoUpdate(T* owner, uint32 time_diff)
     if (i_recheckDistance.Passed())
     {
         i_recheckDistance.Reset(100);
+
         //More distance let have better performance, less distance let have more sensitive reaction at target move.
-        float allowed_dist = owner->GetCombatReach() + sWorld->getRate(RATE_TARGET_POS_RECALCULATION_RANGE);
+        float allowed_dist = 0.0f;
+
+        if ((owner->GetTypeId() == TYPEID_UNIT && owner->IsPet()) && (owner->GetCharmerOrOwnerGUID() == i_target->GetGUID()))
```
jackpoz: are you sure == TYPEID_UNIT check is required in addition to IsPet() ?

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
tkrokli: I thought the original values could be variable since the checks for `GetCombatReach()` and `GetObjectSize()` were used. I don't know if the results could be a fixed value. Maybe @MrSmite knows more about it, since he wrote it.

jackpoz: please don't add comments that don't give any additional info then

SnapperRy: A moth from the Draenei starting area has the same CombatReach and ObjectSize of 1.5f.

Dist would be 1.5f, and size would be 0.

Other pets might have other values.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
GitVerge: he isn't doing any cast after this check, so you are probably right.

MrSmite: > are you sure == TYPEID_UNIT check is required in addition to IsPet() ?

Probably not anymore. The original IsPet() was this:

```C++
bool IsPet() const { return m_unitTypeMask & UNIT_MASK_PET; }
```

and I used the TYPEID_UNIT check just to make sure that owner was in fact a unit. The new IsPet() is quite different though:

```C++
bool IsPet() const { return GetHigh() == HighGuid::Pet; }
```

HighGuid is something that was added after my original commit.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
MrSmite: > CombatReach and ObjectSize values are 1.5f for a Draenei

These values are for the pet, not the player. All pets should have the same reach.

MrSmite: > please don't add comments that don't give any additional info then

My original commit https://github.com/TrinityCore/TrinityCore/commit/d1e6b603394357f3772e81779ff2741c6fdb6e8f had these comments:

```C++
dist = 1.0f; //i_target->GetCombatReach();
size = 1.0f; //i_target->GetCombatReach() - i_target->GetObjectSize();
```
I was trying to normalize the values because all pets should have the same reach / size.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
jackpoz: 335 still checks "m_unitTypeMask & UNIT_MASK_PET"

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
jackpoz: "please don't add comments that don't give any additional info then" was about a comment that the original author removed, not to your code. now it all look weird o.O

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
Shauren: so does 6.x, you must have used grep or viewed it in some other form completely out of context - GetHigh is a method inside ObjectGuid, Unit::IsPet is `m_unitTypeMask & UNIT_MASK_PET`

MrSmite: > you must have used grep or viewed it in some other form completely out of context

I quoted from here: https://github.com/TrinityCore/TrinityCore/blob/9bbca372b6395ebf97d898f018459cc88f3c3a38/src/server/game/Entities/Object/ObjectGuid.h

jackpoz: GetTypeId() == TYPEID_UNIT checks if the object is a Creature, not a Unit.
If there's no chance for "GetTypeId() != TYPEID_UNIT && IsPet()" to be true then the GetTypeId() check can be removed.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
SnapperRy: > These values are for the pet, not the player. All pets should have the same reach.

But in that case, i_target is the player, so the player's combat reach/object size are used.

It makes sense that pets stay further away from, say, a Tauren compared to a Blood Elf.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
MrSmite: > If there's no chance for "GetTypeId() != TYPEID_UNIT && IsPet()" to be true

Yeah, that's something I wasn't sure about. Is there ever a possibility a pet can't be a unit (eg: an object)?

SnapperRy:
Seems fine, on retail pet distance from owner is the same regardless of a race's model size, and the values on the pr seem correct.

[Distance without PR](http://imgur.com/0rAOxQo) - [Distance with PR](http://imgur.com/YCTLSyY) - [Retail](https://imgur.com/uAHRegC) (thanks Aok!).

allowed_dist is correct too, pets must be more responsive to player movement.

PR #17213, opened 2016-05-29
Core/Spells: check database conditions for targets of chain spells

SnapperRy:
**Changes proposed**:

This additional check (partly taken from Spell::CheckCast()) allows database conditions to filter targets of chained spells. For example, the "Searing Wrath" spell used by red drakes in Oculus. The spell should only hit the other dragons, but if you go near the construct and any other NPC, the chain spell will hit them too (in this case, conditions to prevent this already exist in the database, but aren't actually checked against).

**Target branch(es)**: 335

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -1935,6 +1935,17 @@ void Spell::SearchChainTargets(std::list<WorldObject*>& targets, uint32 chainTar
             break;
         target = *foundItr;
         tempTargets.erase(foundItr);
+
+        // check spell cast conditions from database
+        {
+            ConditionSourceInfo condInfo = ConditionSourceInfo(m_caster, target);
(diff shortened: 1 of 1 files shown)


joschiwald:
chain targets are filterable with CONDITION_SOURCE_TYPE_SPELL_IMPLICIT_TARGET (13) since a recent commit

SnapperRy:
I didn't know that, but this actually affects mainly CONDITION_SOURCE_TYPE_SPELL (17), as that's the one used to say "only this this and that target can be hit by this spell, nobody else", unless I'm missing something in the picture.

joschiwald:
CONDITION_SOURCE_TYPE_SPELL (17) means you can cast the spell or not on your explicit target

chain targets count as implicit target, because you don't explicit select the target

SnapperRy:
Oh I see, it makes sense now. Thanks. I'll try that for Oculus. This can be scrapped, then.

PR #17232, opened 2016-05-30
[3.3.5] Core/Player: Added support cast spell for some class spells on login

SnapperRy:
**Changes proposed**:
- Bring c8eb69d from the WoD branch to 335 for appropriate cases (Battle Stance and Blood Presence cast at first login).

**Target branch(es)**: 335


sql/updates/world/3.3.5/9999_99_99_00_world.sql
```diff
@@ -0,0 +1,7 @@
+DROP TABLE IF EXISTS `playercreateinfo_cast_spell`;
+CREATE TABLE IF NOT EXISTS `playercreateinfo_cast_spell` (
```
Aokromes: put the 2 files inside one plz. we don't use multiple files anymore on a single commit.

SnapperRy: Done!

PR #15022, opened 2015-07-07
SmartAI: Possibility to despawn object with SMART_ACTION_FORCE_DESPAWN

robinsch:
As requested in #15020.

PR #17259, opened 2016-06-01
Core/SmartAI: remove unneeded check for SMART_ACTION_FORCE_DESPAWN.

SnapperRy:
**Changes proposed**: remove unneeded check related to gameobjects targeted with SMART_ACTION_FORCE_DESPAWN. There's no reason to have it, as the despawn functionality doesn't require SmartAI.

**Target branch(es)**: 335

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -1037,10 +1037,7 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                         target->DespawnOrUnsummon(e.action.forceDespawn.delay);
                 }
                 else if (GameObject* goTarget = (*itr)->ToGameObject())
-                {
-                    if (IsSmartGO(goTarget))
-                        goTarget->SetRespawnTime(e.action.forceDespawn.delay + 1);
-                }
+                    goTarget->SetRespawnTime(e.action.forceDespawn.delay + 1);
             }
 
             delete targets;


jackpoz:
SmartAI checks are usually meant to ensure that multiple AIs are not mixed, i.e. a SmartAI despawning a GO spawned by a C++ script.

SnapperRy:
I'll post what I'm working on:

```sql
INSERT INTO `smart_scripts` VALUES
('137300','9','4','0','0','0','100','0','1000','1000','0','0','41','10','0','0','0','0','0','14','1037','269','0','0','0','0','0','Jarven Thunderbrew - On Script - Despawn Closest Gameobject \'Guarded Thunder Ale Barrel\'');
```

This fails due to that check, because the gameobject "Guarded Thunder Ale Barrel" isn't set to use SmartAI. It didn't work even when I assigned a SmartGameObjectAI to it, but not sure about the reason for that one.

jackpoz:
> skipped to prevent crash.

well this part of the log raises a few eyebrows. the check was added back in https://github.com/TrinityCore/TrinityCore/commit/cebf8590e39129fdd8526eaa4302d143340a1cff but it doesn't give any detail about how it can crash

SnapperRy:
Yep, I think it's copypasted after the creature despawn code (that one actually requires SmartAI and has calls to ENSURE_AI, so it makes sense it'd crash).

I'm experiencing no crashes with this patch, although I'm testing with only one game object so it's not bulletproof. I think there are no risks, though (you're just despawning a gameobject using GameObject::SetRespawnTime()).

jackpoz:
it seems this code has been discussed already in the original PR https://github.com/TrinityCore/TrinityCore/pull/15022#discussion_r34023763 but it was merged even if this case wasn't handled or explained

SnapperRy:
Oh, thanks for finding it.

To answer your question there (in case it's still valid, it's been a year), forceDespawn.delay specifies after how many seconds the target will respawn.

It's not really a despawn, but a respawn.

tkrokli:
Travis is fixed via TC `.travis.yml` so you can refresh your PR on your repo to reactivate Travis here.

PR #17270, opened 2016-06-02
Update worldserver.conf.dist

Espionage724:
**Changes proposed**:
- Minor typo

**Target branch(es)**: 335/6x


tkrokli:
... or 'alignment fix' would probably be a better wording to use. But it's OK. :-)

PR #17271, opened 2016-06-03
Core/Movement: interrupt a creature's cast when fleeing/roaming

### LINKED ISSUE #12823: [Core/Movement] MovementType 1 & 2 creatures casting non-movement spells
db : up-to-date
core : 14cb6e4235ebb84ad41d6ceb4eaf56c0b55d9238

When a creature has movementType 1 (random move), when it goes into combat, and cast a spell (which needs cast time, and the spell can not cast when move), you will find the creature can cast when move (until it reach the random move point).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3658755-fix-bug-problem-about-cast-spell-when-creature-s-movementtype-is-1-or-2?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SnapperRy:
**Changes proposed**: creatures using MovementType 1 or 2, as well as those using SMART_ACTION_FLEE_FOR_ASSIST) can and will often cast in movement spells that shouldn't be castable while moving. This fixes that!™

**Target branch(es)**: 335.

**Issues addressed**: Fixes #12823.

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -534,6 +534,7 @@ void MotionMaster::MoveSeekAssistance(float x, float y, float z)
         TC_LOG_DEBUG("misc", "Creature (Entry: %u GUID: %u) seek assistance (X: %f Y: %f Z: %f)",
             _owner->GetEntry(), _owner->GetGUID().GetCounter(), x, y, z);
         _owner->AttackStop();
+        _owner->CastStop();
         _owner->ToCreature()->SetReactState(REACT_PASSIVE);
         Mutate(new AssistanceMovementGenerator(x, y, z), MOTION_SLOT_ACTIVE);
     }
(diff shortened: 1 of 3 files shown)


GitVerge:
Since the problem is creatures casting while moving not vice versa maybe you should probably treat the cast instead of "interrupting" it by stopping the movement.

Did you try interruptnonmeleespells?

SnapperRy:
It goes both ways. They start casting while moving, they move after having already started casting.

GitVerge:
interrupting the spell should work for both ways

SnapperRy:
Thinking about it, creatures shouldn't (I think?) interrupt a cast to execute a (random) movement, they should stop, complete the cast and resume moving.

Correct me if I'm wrong :P

GitVerge:
Without the fleeing part looks better, but you don't need the first couple of checks because the creature will stop moving when deathstate is changed to JUST_DIED.

SnapperRy:
It's an unrelated change (not really worthy of a separate pull request), since it makes no sense to update it if the creature isn't alive :P

This isn't working for some creatures, though (Defias Pillagers), so back to the drawing board.

SnapperRy:
There, this now works as one would expect and actually closes that issue.

mik1893:
i'm pushing this up now for some testing

PR #17301, opened 2016-06-06
Core/Commands: Implement .neargraveyard (find nearest graveyard from dbc)

xjose93:
**Changes proposed**:
- Possibility to find nearest graveyard from dbc if doesn't exist on db.

**Target branch(es)**: 335 (can be cherrypicked to 6xx too)


Keader:
You make a new PR to "merge all commits in one"?
you can use squash it, look: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

src/server/scripts/Commands/cs_misc.cpp
```diff
@@ -2621,6 +2622,47 @@ class misc_commandscript : public CommandScript
...
+    {
+        Player* player = handler->GetSession()->GetPlayer();
+
+        float x = player->GetPositionX();
+        float y = player->GetPositionY();
+        float z = player->GetPositionZ();
+
+        const WorldSafeLocsEntry* nearestLoc = nullptr;
+        bool isNear = false;
+        float distNearest = 20000.0f;
```
jackpoz: you could use std::numeric_limits<float>::max() from <limits> and simplify the code in the for loop

src/server/scripts/Commands/cs_misc.cpp
```diff
@@ -2621,6 +2622,47 @@ class misc_commandscript : public CommandScript
...
+                {
+                    isNear = true;
+                    distNearest = dist;
+                    nearestLoc = loc;
+                }
+            }
+        }
+
+        if (nearestLoc)
+            handler->PSendSysMessage(LANG_NEARGRAVEYARD, nearestLoc->ID, nearestLoc->x, nearestLoc->y, nearestLoc->z);
```
jackpoz: is there a case when a graveyard could not be found so that we need to tell the user with another message ? or that shouldn't matter (arenas/instances maybe ?)

Treeston:
Would you mind putting the command somewhere that isn't the top-level command namespace? `.debug neargraveyard` maybe? Just `.debug graveyard` would work too.

There's a ton of clutter in the top-level namespace as is, no need to make it worse.

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -1485,6 +1487,63 @@ class debug_commandscript : public CommandScript
...
+            bool isNear = false;
+            float distNearest = std::numeric_limits<float>::max();
+            
+            for (uint32 i = 0; i < sWorldSafeLocsStore.GetNumRows(); ++i)
+            {
+                const WorldSafeLocsEntry* loc = sWorldSafeLocsStore.LookupEntry(i);
+                if (loc && loc->map_id == player->GetMapId())
+                {
+                    float dist = (loc->x - x) * (loc->x - x) + (loc->y - y) * (loc->y - y) + (loc->z - z) * (loc->z - z);
+                    if (isNear)
```
jackpoz: missed the "simplify the code in the for loop" part. there's no need to use isNear at all since any dist will always be < std::numeric_limits<float>::max(), so the whole "else" block can be removed

xjose93:
Changed to .debug neargraveyard, with possibility of .debug neargraveyard linked which shows the nearest graveyard linked on db.

Aokromes:
soo, any opposition to merge this?

Issue #17298, opened 2016-06-06
Core/ThreatManager: doAddThreat doesnt send correct ThreatRefStatusChangeEvent to container.


ccrs:
**Description**:

``doAddThreat`` function doesnt send the correct event to container: on threat change, the container doesnt reorder the list as expected.
If there aren't any other events that make the list dirty (and then make a reorder on next update), it will continue forever unordered.
Therefore, functions like SelectTarget may return incorrect targets if a position is used.
(the target value (or order) managed by SelectTarget its the same as threatlist)

``doAddThreat`` is used in ``void Spell::EffectTaunt(SpellEffIndex /*effIndex*/)`` so any taunt spell wont make the threatlist reorder after threat consolidation.

**Expected behaviour**:
The list should be reordered (set as dirty) if the top threat value is not the first element on it.

**Steps to reproduce the problem**:

1. Go to Lich King with two characters.
2. Make it change to phase 2.
3. Make the character without LK aggro use a taunt spell on it.
4. See how defile is always used on the current "tank".

related code to the example:
````c++
if (Unit* target = SelectTarget(SELECT_TARGET_RANDOM, 1, 0.0f, true, -SPELL_HARVEST_SOUL_VALKYR))
{
    Talk(EMOTE_DEFILE_WARNING);
    DoCast(target, SPELL_DEFILE);
}
````

**Branch(es)**: 3.3.5 / 6.x

**TC hash/commit**:  
7ff0f3c65fd0c068560351eac92e96a8527c96a7

jackpoz:
strange how there's a particular function for taunt too https://github.com/TrinityCore/TrinityCore/blob/caa1b82b88edf1d425aa147c8f5851df0a5d14d2/src/server/game/Combat/ThreatManager.cpp#L485
which events do make the list dirty and trigger a reorder ?

ccrs:
> which events do make the list dirty and trigger a reorder ?

````c++
enum UNIT_EVENT_TYPE
{
    // Player/Pet changed on/offline status
    UEV_THREAT_REF_ONLINE_STATUS        = 1<<0,

    // Threat for Player/Pet changed
    UEV_THREAT_REF_THREAT_CHANGE        = 1<<1,

    // Player/Pet will be removed from list (dead) [for internal use]
    UEV_THREAT_REF_REMOVE_FROM_LIST     = 1<<2,

    // Player/Pet entered/left  water or some other place where it is/was not accessible for the creature
[...]

PR #17309, opened 2016-06-07
Core/UnitAI: SelectTarget correction

ccrs:
Set CurrentVictim as first element on targetList if necessary, which is then filled with the threatlist elements that satisfy the predicate.

ref #17298

**Target branch(es)**: 335/6x

### DIFF
--- a/src/server/game/AI/CoreAI/UnitAI.h
+++ b/src/server/game/AI/CoreAI/UnitAI.h
@@ -151,11 +151,11 @@ class TC_GAME_API UnitAI
                 return nullptr;
 
             std::list<Unit*> targetList;
-            HostileReference* currentVictim = me->getThreatManager().getCurrentVictim();
+            Unit* currentVictim = me->getThreatManager().getCurrentVictim()->getTarget();
 
             // Current victim always goes first
             if (predicate(currentVictim))
-                targetList.push_back(currentVictim->getTarget());
+                targetList.push_back(currentVictim);
 
             for (ThreatContainer::StorageType::const_iterator itr = threatlist.begin(); itr != threatlist.end(); ++itr)
                 if ((*itr)->getTarget() != currentVictim && predicate((*itr)->getTarget()))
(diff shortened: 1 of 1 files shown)


src/server/game/AI/CoreAI/UnitAI.h
```diff
@@ -148,15 +148,21 @@ class TC_GAME_API UnitAI
...
             ThreatContainer::StorageType const& threatlist = me->getThreatManager().getThreatList();
             if (position >= threatlist.size())
-                return NULL;
+                return nullptr;
 
             std::list<Unit*> targetList;
+            HostileReference* currentVictim = me->getThreatManager().getCurrentVictim();
+
+            // Current victim always goes first
+            if (predicate(currentVictim))
```
Treeston: Do we need to null check `currentVictim` here?

Treeston:
Never thought of it, but yeah, first element of targetList does not have to be the tank (even though most scripts assume that it is).

:+1:

src/server/game/AI/CoreAI/UnitAI.h
ccrs: as this check is done before
````c++
if (position >= threatlist.size())
````
the list will always have something in it, so there will always be a currentVictim afaik

Keader:
> /home/travis/build/TrinityCore/TrinityCore/src/server/game/AI/CoreAI/UnitAI.h:161:41: fatal error: 
>       comparison of distinct pointer types ('Unit *' and 'HostileReference *')
>                 if ((*itr)->getTarget() != currentVictim && predicate((*...
>                     ~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~

src/server/game/AI/CoreAI/UnitAI.h
Shauren: you want to use `.getCurrentVictim()->getTarget()` to deal with units instead of HostileReference

Shauren:
this is good

src/server/game/AI/CoreAI/UnitAI.h
ccrs: derp

Aokromes:
Soo, any opposition to merge this?

PR #17313, opened 2016-06-07
Scripts/Ulduar: Fix Biting Cold

### LINKED ISSUE #16999: [3.3.5] Hodir- NPC's are being affected by biting cold
NPC's during Hodir encounter are getting high amounts of stacks of biting cold and dying super quickly.
http://www.wowhead.com/npc=32845/hodir#article (NPC links)

NPC's suffer from biting cold http://www.wowhead.com/spell=62039/biting-cold, thus dying and making the fight much more difficult and impossible for Hardmode

NPC's should not be affected by biting cold

Go to Ulduar and start Hodir encounter
Break NPC's out of Flash Freeze

3.3.5

Commit hash:    4afea55b2c6035a986162a8d800a57fb1ae4cf98

TDB 335.60

Win7

adeutscher:
**Changes proposed**:
- When cycling through targets, Biting Cold should ignore anything that is not a player.

**Target branch(es)**: 335

**Issues addressed**: Fixes #16999 

**Tests performed**:

Engaged Hodir, and the NPCs did not gain Biting Cold stacks.

### DIFF
--- a/src/server/scripts/Northrend/Ulduar/Ulduar/boss_hodir.cpp
+++ b/src/server/scripts/Northrend/Ulduar/Ulduar/boss_hodir.cpp
@@ -967,7 +967,7 @@ class spell_biting_cold : public SpellScriptLoader
 
                 for (TargetList::iterator itr = listOfTargets.begin(); itr != listOfTargets.end(); ++itr)
                 {
-                    if (itr->first != target->GetGUID())
+                    if (itr->first != target->GetGUID() || target->GetTypeId() != TYPEID_PLAYER)
                         continue;
 
                     if (itr->second >= 4)


joschiwald:
better use DoCheckAreaTarget but it is still hack, because area auras need generic fix, they do not follow generic target selection

adeutscher:
Doh, will do. Closing this PR then, since `SPELL_ATTR3_ONLY_TARGET_PLAYERS` would be a much better general fix.

mik1893:
@adeutscher this is fixed by #17321

adeutscher:
Excellent, thanks for coming up with a cleaner solution!

PR #17321, opened 2016-06-09
Core/SpellAuras: Add target check to area auras

### LINKED ISSUE #16999: [3.3.5] Hodir- NPC's are being affected by biting cold

mik1893:
**Changes proposed**:
- Add target spell validity check to area auras

**Target branch(es)**: 335/6x

**Issues addressed**: Fixes #16999

**Tests performed**: Build - to be tested extensively in game

-  [x] Extensively test in game (try different bosses with area auras)
-  [x] Clean up code and add additional conditions (if needed)

### DIFF
--- a/src/server/game/Spells/Auras/SpellAuras.cpp
+++ b/src/server/game/Spells/Auras/SpellAuras.cpp
@@ -2471,6 +2471,9 @@ void UnitAura::FillTargetMap(std::map<Unit*, uint8> & targets, Unit* caster)
 
         for (UnitList::iterator itr = targetList.begin(); itr!= targetList.end();++itr)
         {
+            if (GetSpellInfo()->CheckTarget(caster, *itr, true) != SPELL_CAST_OK)
+                continue;
+
             std::map<Unit*, uint8>::iterator existing = targets.find(*itr);
             if (existing != targets.end())
                 existing->second |= 1<<effIndex;
@@ -2524,6 +2527,9 @@ void DynObjAura::FillTargetMap(std::map<Unit*, uint8> & targets, Unit* /*caster*
 
         for (UnitList::iterator itr = targetList.begin(); itr!= targetList.end();++itr)
         {
+            if (GetSpellInfo()->CheckTarget(dynObjOwnerCaster, *itr, true) != SPELL_CAST_OK)
+                continue;
+
             std::map<Unit*, uint8>::iterator existing = targets.find(*itr);
             if (existing != targets.end())
                 existing->second |= 1<<effIndex;
--- a/src/server/game/Spells/Auras/SpellAuras.cpp
+++ b/src/server/game/Spells/Auras/SpellAuras.cpp
@@ -614,6 +614,49 @@ void Aura::UpdateTargetMap(Unit* caster, bool apply)
     }
 }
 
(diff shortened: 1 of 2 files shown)


mik1893:
@Keader - this is a kind of generic solution, i'm not entirely sure I'm addressing it properly but if you can try this we can see how it behaves

mik1893:
although this works, i'm not entirely sure that the implementation is correct.
I think the problem is in the Spell::CheckCast function, that executes all the condition and target checks only when m_targets returns an ObjectTarget or a UnitTarget - not going through the list of all the targets...

joschiwald:
@mik1893 what do you mean?

mik1893:
i  mean that the way i have done it, it does some check for targets, but its not all that it should do - there are other things (e.g. conditions) that are not checker.
Also it's quite ugly to have a check on a spellcast result on an aura application (my personal 2 cents)
might need to write a method that properly executes some target checks for auras

joschiwald:
checkcast is a bit different to that you need here, checkcast do checks on caster and on explicit target, here you need implicit target checks
one big problem here is, how are targets searched

mik1893:
i will doing some basics to fix the errors related to anubarak and hodir, then will likely add smth better on the way

mik1893:
i have left only the checks that I THINK might be useful.

Keader:
Now looks like more clean ^^

P-Kito:
I dont know why, but I have a feeling that this might break something or make spellscripts useless that had some filtering, which we need to take out then.

mik1893:
well we tested it quite extensively and it is looking fine - 
the spellscripts doing target filtering will filter after this list is done - so in the worst case they are useless

P-Kito:
Okay then I have no doubts. :P

QAston:
LGTM, though there's always risk of breaking some scripts.

mik1893:
I'll merge soon need to do a small change

Treeston:
This breaks the "Ghost" debuff from being applied to dead characters releasing.

ccrs:
Where is the conditions check?

Aokromes:
@mik1893 

mik1893 commented on Jul 12
I'll merge soon need to do a small change

Shauren:
Do not merge - https://github.com/TrinityCore/TrinityCore/pull/17321#issuecomment-233002317

Aokromes:
I known i was waking mik to fix that :P

Aokromes:
@mik1893 stop changing diapers for a while and give love to your another baby :P

Aokromes:
Conflicting files
src/server/game/Spells/Auras/SpellAuras.cpp

Shauren:
Since it breaks generic stuff in favor of some special situations - closing.

PR #17325, opened 2016-06-10
Core/Misc: Remove duplicate code

r4d1sh:
Refactor


Rushor:
make shure to use spaces (4) instead of tabs

CraftedRO:
is this PR still valid @r4d1sh ? I see it was never commited and the code seems the same.

r4d1sh:
> is this PR still valid @r4d1sh ? I see it was never commited and the code seems the same.

Closed as you can see. This is not such an important change, many years ago I was just learning to write code and make pull requests.

PR #17369, opened 2016-06-14
Core/Player: implemented account-wide mounts

MitchesD:
**Changes proposed**:
Implemented account wide mounts

**Target branch(es)**: 6x

**Issues addressed**: Updates https://github.com/TrinityCore/TrinityCore/issues/13825

**Tests performed**: Tested ingame

- [x] After learning a new mount from item or spell, you can't set it as favorite till you re-log, can't figure out why, maybe there is some dynamic field or some flag
- [x] Move SQL to proper folders and update base auth and updates
- [x] Make more tests for faction mounts behavior


src/server/game/Entities/Player/CollectionMgr.cpp
```diff
@@ -259,3 +260,137 @@ bool CollectionMgr::CanApplyHeirloomXpBonus(uint32 itemId, uint32 level)
...
+        stmt->setUInt32(0, _owner->GetBattlenetAccountId());
+        stmt->setUInt32(1, mount.first);
+        stmt->setBool(2, mount.second);
+        trans->Append(stmt);
+    }
+}
+
+bool CollectionMgr::UpdateAccountMounts(uint32 spellId, bool isFavourite)
+{
+    return _mounts.insert(MountContainer::value_type(spellId, isFavourite)).second;
```
P-Kito: _mounts.insert(MountContainer::value_type(spellId, isFavourite).second);

Maybe because of this, mounts are not being updated upon learning?
Not sure tho..

src/server/game/Entities/Player/CollectionMgr.h
```diff
@@ -78,12 +80,25 @@ class TC_GAME_API CollectionMgr
...
+    MountContainer const& GetAccountMounts() const { return _mounts; }
 
 private:
     WorldSession* _owner;
 
     ToyBoxContainer _toys;
     HeirloomContainer _heirlooms;
+    MountContainer _mounts;
+
+    static MountDefinitionMap _mountDefinitions;
```
Naios: Avoid using non constant static variables in header files since this may cause issues with dynamic linking.

Since you are using `_mountDefinitions` only in your compilation unit you could move it directly to `CollectionMgr.cpp` without referencing it in your `CollectionMgr` class or you make it non-static because you are always working directly on the `CollectionMgr` anyway.

MitchesD: I did that stuff some time ago, before you linking changes, so didn't look on it from this angle, but thanks, will fix it

src/server/game/Entities/Player/CollectionMgr.cpp
```diff
@@ -259,3 +260,137 @@ bool CollectionMgr::CanApplyHeirloomXpBonus(uint32 itemId, uint32 level)
...
+        return;
+
+    itr->second = favorite;
+
+    // Temporary container, just need to store only selected mount
+    MountContainer tempMounts;
+    tempMounts.insert(*itr);
+    
+    WorldPackets::Misc::AccountMountUpdate mountUpdate;
+    mountUpdate.IsFullUpdate = false;
```
GitVerge: did you try with full update to see if the favorite state changes?

MitchesD: @GitVerge problem is that it never goes here, problem is that CMSG_MOUNT_SET_FAVORITE isn't even sent when you click on Set Favorite

GitVerge: oh, I see. Will let you know if something pops up

GitVerge: maybe you need to send AccountMountUpdate inside addmount, because the only thing that I see fixes the issue at relog is that update when added to map

MitchesD: @GitVerge thanks, I think I might found a bit logic fail inside, let me check ;)

Pixel-Pirate:
It lacks a function to remove mount from table battlenet_account_mounts whenplayer unlearn his mount ;)

Aokromes:
that's supposed to be blizzlike?

Shauren:
@Pixel-Pirate no, it's not missing, you cannot unlearn mounts on retail either (also trying to implement it adds a problem unsolvable with current codebase, synchronization of data between realms)

Anddrw:
> unsolvable with current codebase, synchronization of data between realms

Heil Mangos!

PiskelDEV:
what about this, I just can confirm that it works properly - good job :+1:

soulfrost:
image
working fine for me too

MitchesD:
been busy last days, from me it's completed, will try to merge it when I have some time

wowharley (commented):
does anyone have the current opcode doe this

CMSG_MOUNT_SET_FAVORITE

Kittnz:
@MitchesD had enough time? ;) :dancer:

nyhmii:
Can we use this for 3.3.5a TC??

Issue #657, opened 2011-05-16
[3.3.5][Priest] - Shadowform


click:
Spell: [http://old.wowhead.com/spell=15473]

Error: Cast any spell, and put shadowform before end global CD, not apply shadowform, and no longer lets you get Shadowform.

Core: https://github.com/TrinityCore/TrinityCore/commit/790086ea2f06a659d3995811bb42212771b5b3ce

Raydor:
Still bugged https://github.com/TrinityCore/TrinityCore/commit/48fa7fa992c02d7c17c39c08cb78c6457d034f49

cguliver:
when was  posted first time was  on [ May 17, 2011 ] so i think is required 10 more  years to fix this LOL :"

sirikfoll:
Reopen?
Still bugged  0148832

Aokromes:
I am using https://github.com/TrinityCore/TrinityCore/issues/657#issuecomment-8433390 for years.

mik1893:
note also that the same should be applied to tricks of the trade and misdirection
```
    if (strict && !(_triggeredCastFlags & TRIGGERED_IGNORE_GCD) && HasGlobalCooldown()) 
    {
        switch (m_spellInfo->Id)
        {
            case 15473: // Shadowform
            case 34477: // Misdirection
            case 57934: // Tricks of the Trade
                return SPELL_FAILED_FIZZLE; // !Hack - Returning "Not ready" bugs Shadowform client side
            default:
                return SPELL_FAILED_NOT_READY;
        }
    }

Keader:
Dont work for MD and Tricks of the Trade (https://github.com/TrinityCore/TrinityCore/issues/657#issuecomment-226107058)
Already using this fix, and still bug sometimes.
Shadowform works well :)

mik1893:
mmmh to me MD and tricks of the trade work...
not entirely sure I did anything different from this to make it work

PR #17376, opened 2016-06-15
Core/Spells: enable Shadowform, Misdirection, Tricks of the Trade to be casted while in GCD

### LINKED ISSUE #12913: Core/Spell: Misdirection, Tricks of the Trade, Shadowform stuck
Hi all

I've opened this issue as I can´t find any other regarding this ability: http://www.wowhead.com/spell=34477

The problem here seems to be with its cooldown, or the way it's handled.

Sometimes, randmly (i cant find a pattern or an explanation) it gets stuck in the first "phase".

This first "phase" is when u click and the button is grey colored. When ur meant to attack a creature to start the Misdirection.

The spell gets stuck there and players cant cast it again (this persist even after dead)

hash: a8ff0c990b68eda2bf3c22bbf4d72331bc8842b9

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3991067-misdirection?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mik1893:
**Changes proposed**:
- Fix issues related to the DISABLED_WHILE_ACTIVE

**Target branch(es)**: 335/6x

**Issues addressed**: Fixes #12913 and #657

**Tests performed**: build - to be re-tested in game

- [ ]  Test again the 3 spells
- [ ]  is SPELL_FAILED_FIZZLE the right one to use?

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -4739,6 +4739,8 @@ SpellCastResult Spell::CheckCast(bool strict)
         {
             if (m_triggeredByAuraSpell)
                 return SPELL_FAILED_DONT_REPORT;
+            else if (m_spellInfo->HasAttribute(SPELL_ATTR0_DISABLED_WHILE_ACTIVE))  // Shadowform, Tricks of the Trade, Misdirection
+                return SPELL_FAILED_FIZZLE;
             else
                 return SPELL_FAILED_NOT_READY;
         }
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/Spell.cpp
```diff
@@ -4739,6 +4739,8 @@ SpellCastResult Spell::CheckCast(bool strict)
         {
             if (m_triggeredByAuraSpell)
                 return SPELL_FAILED_DONT_REPORT;
+            else if (m_spellInfo->HasAttribute(SPELL_ATTR0_DISABLED_WHILE_ACTIVE))  // Shadowform, Tricks of the Trade, Misdirection
```
Shauren: Failing SpellHistory::IsReady check should continue to return SPELL_FAILED_NOT_READY - it means the cast failed because it was already active or we deactivated it and triggered normal cooldown. Sending any other value than SPELL_FAILED_NOT_READY will make the client clear cooldown on it.

mik1893: tested it now - it will anyway work fine without this part 
the only part I m missing now is why do tricks of the trade and misdirection not work with the same piece of code - I figure I might have dome some other mods unrelated to this in the past, I'm figuring it out now

mik1893:
So, for misdirection and tricks of the trade, if target is not in range, the spells will fail and stuck 
- need more investigation.

codie72:
added this to my server, its much better thanks

PR #17384, opened 2016-06-16
[6.x] Core/Spell: Provoke (115546) 

Infamous-devel:
**Changes proposed**:
- Provoke - (www.wowhead.com/spell=115546)

**Target branch(es)**:  6x


sql/updates/world/6.x/9999_99_99_99.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM `spell_script_names` WHERE `ScriptName` IN ('spell_monk_provoke');
+INSERT INTO `spell_script_names` (`spell_id`,`ScriptName`) VALUES
+(115546,'spell_monk_provoke');
```
Aokromes: missing newline.

src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -125,8 +125,53 @@ class spell_monk_crackling_jade_lightning_knockback_proc_aura : public SpellScri
     }
 };
 
+// Provoke - 115546
```
P-Kito: Some codestyle is not correct in this spellscript. Have a look at the other spellscripts, concerning TABS

src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -125,8 +125,53 @@ class spell_monk_crackling_jade_lightning_knockback_proc_aura : public SpellScri
...
+                return SPELL_FAILED_BAD_TARGETS;
+            else if (!target->IsWithinLOSInMap(GetCaster()))
+                return SPELL_FAILED_LINE_OF_SIGHT;
+            return SPELL_CAST_OK;
+        }
+
+        void HandleDummy(SpellEffIndex /*effIndex*/)
+        {
+            if (Unit* caster = GetCaster())
+                if (caster->getClass() == CLASS_MONK && caster->GetTypeId() == TYPEID_PLAYER)
```
P-Kito: Checking the class is never done and is therefore not needed.

(Just curious: Dark Simulacrium will steal "Provoke" - DK can't cast it then, because he is not a monk?^^)

src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -30,10 +30,10 @@ enum MonkSpells
 {
     SPELL_MONK_CRACKLING_JADE_LIGHTNING_CHANNEL = 117952,
     SPELL_MONK_CRACKLING_JADE_LIGHTNING_CHI_PROC = 123333,
-    SPELL_MONK_STANCE_OF_THE_SPIRITED_CRANE = 154436,
-
     SPELL_MONK_CRACKLING_JADE_LIGHTNING_KNOCKBACK = 117962,
     SPELL_MONK_CRACKLING_JADE_LIGHTNING_KNOCKBACK_CD = 117953,
+    SPELL_MONK_PROVOKE = 118635,
```
P-Kito: http://www.wowhead.com/spell=116189/provoke
http://www.wowhead.com/spell=118635/provoke

your spell is the second, which is a area taunt... this is wrong and I think it should be casted  by the statue if I remember correctly.

src/server/scripts/Spells/spell_monk.cpp
P-Kito: Oh and:
It's a dummy effect because if you are brewmaster and you have a statue up, the statue will cast the AOE version of Provoke.

> This ability can be targeted on your Statue of the Black Ox, causing the same effect on all enemies within  10 yards of the statue

P-Kito:
Please fix the following issues:

[ ] Codestyle, tabs, unneeded if-cases
[ ] Provoke will cast http://www.wowhead.com/spell=116189/provoke if NOT targeting the statue of the Ox
[ ] Provoke will cast http://www.wowhead.com/spell=118635/provoke if targeting the statue of the Ox, which will make the statue cast this spell (118635)

src/server/scripts/Spells/spell_monk.cpp
GitVerge: someone once wrote that caster can't be null in spellscripts and instead of that monk/player check (as p-kito said), use spellinfo validation.

P-Kito:
Well good job so far. 

I think something is still missing when targeting the statue of the ox.
Someone has to test if THE STATUE will cast the spell or THE MONK with statue as target.

Is the statue attackable by the monk? If not ->GetTarget() will not return the statue I think.
In Deathpact script in spell_dk.cpp (3.3.5 branch) they selected the target by accessing Unit_field_target (or however its called). Maybe this will help too.

if you add brackets ( } { ) make sure to have 1 space infront and after them, this makes codestyle better and gives a lil better overview.

Shauren:
@P-Kito lol no. You can cast this spell with a focus/mouseover target, UNIT_FIELD_TARGET will not return correct data

P-Kito:
Is the statue attackable by the monk?

P-Kito:
Hmmm, okay. Not sure how to get the statue then :-1:

P-Kito:
Okay, I tested it ingame and you can use the spell provoke the statue, you now have to get the Target and let it cast the AOE Provoke with monk as caster.

```c++
if (Unit* hitUnit = GetHitUnit())
{
   if (hitUnit->GetEntry() == BLACK_OX_STATUE)
       hitUnit->CastSpell(........, GetCaster()->GetGUID());
   else
       GetCaster()->CastSpell(hitUnit, SPELL_MONK_PROVOKE);
}
```

P-Kito:
Check the target, if its the statue, let statue cast the aoe with monk as caster.... rest is up to you...

Issue #16133, opened 2015-12-31
[3.3.5] DB/Quest: The Hunt is On (11794)


Treeston:
Branch 3.3.5, revision 79c9af11b9a90b051ecc26b0a60e879e3d9c4134

[Quest: The Hunt is On](http://www.wowhead.com/quest=11794/the-hunt-is-on#screenshots:id=262507)

Quest functionality works fine, however quest targets do not display a dark glow around them while the player has the [Righteous Vision](http://www.wowhead.com/spell=46078/righteous-vision) buff. In addition, quest target [Tom Hegger](http://www.wowhead.com/npc=25827/tom-hegger) does not have the proper target set for his aggro talk (it displays $c).

Saben65:
Looks like the spell needs core support. The NPCs were missing the aura, and it still doesn't show after adding it.

```SQL
DELETE FROM `creature_template_addon` WHERE `entry` IN (25248,25827,25828);
INSERT INTO `creature_template_addon` (`entry`, `mount`, `bytes1`, `bytes2`, `auras`) VALUES
(25248, 0, 0x0, 0x1, '46077'), -- 25248 - 46077
(25827, 0, 0x0, 0x1, '46077'), -- 25827 - 46077
(25828, 0, 0x0, 0x1, '46077'); -- 25828 - 46077

DELETE FROM `creature_addon` WHERE `guid` IN (117733,111073,117769);
INSERT INTO `creature_addon` (`guid`, `mount`, `bytes1`, `bytes2`, `auras`) VALUES
(117733, 0, 0x0, 0x1, '46077'), -- 25827 - 46077 - 46077
(111073, 0, 0x0, 0x1, '46077'), -- 25248 - 46077 - 46077
(117769, 0, 0x0, 0x1, '46077'); -- 25828 - 46077 - 46077
```

tkrokli:
About Tom Hegger not having a proper target, he probably needs to target his speech/yell like this (https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/zone_borean_tundra.cpp#L2316-L2320) :
```diff
                            case NPC_TOM_HEGGER:
-                               Talk(SAY_HIDDEN_CULTIST_3);
+                               Talk(SAY_HIDDEN_CULTIST_3, player);
                                uiEventTimer = 5000;
                                uiEventPhase = 2;
                                break;
```
That is at least the method that worked for `npc_mootoo_the_younger`
[...]

PR #17395, opened 2016-06-17
Core/Scripts: Tom Hegger must target player in Say text

ghost:
**Changes proposed:**

Make NPC 25827, Tom Hegger, say his aggro text while targeting player to use $c correctly.

[zone_borean_tundra.cpp](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/zone_borean_tundra.cpp)
Script: [npc_hidden_cultist](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/zone_borean_tundra.cpp#L2200)
Quest: The Hunt is On, 11794 (http://www.wowhead.com/quest=11794/the-hunt-is-on)
NPC:  [Tom Hegger, 25827](http://www.wowhead.com/npc=25827/tom-hegger)

Before this change, Tom Hegger does not target player when saying his
aggro text "You don't know who you're messing with, $c!   Death beckons!"
and hence the text $c will be shown instead of the player class.

**Target branch(es):** 3.3.5 (I do not know if this issue exists in 6.x)

**Issues addressed**: Updates #16133 [[3.3.5] DB/Quest: The Hunt is On (11794)](https://github.com/TrinityCore/TrinityCore/issues/16133)

Builds without issues, works as intended/expected in-game. :+1:

No known issues with this script modification, it compiles and works in-game.
The added line is identical to another line used in the same script (reused code).

### DIFF
--- a/src/server/scripts/Northrend/zone_borean_tundra.cpp
+++ b/src/server/scripts/Northrend/zone_borean_tundra.cpp
@@ -2314,7 +2314,8 @@ public:
                                 uiEventPhase = 2;
                                 break;
                             case NPC_TOM_HEGGER:
-                                Talk(SAY_HIDDEN_CULTIST_3);
+                                if (Player* player = ObjectAccessor::GetPlayer(*me, uiPlayerGUID))
+                                    Talk(SAY_HIDDEN_CULTIST_3, player);
                                 uiEventTimer = 5000;
                                 uiEventPhase = 2;
                                 break;


Naios:
If this doesn't happen aleady, maybe we can log an error when the formatter fails to deduce the playername.

Shauren:
@Naios formatting happens clientside.

Naios:
We could also check against that.

DDuarte:
String search for $C, $G and so on and if no target error log?

Naios:
Yep, better precheck all strings on startup.

PR #17434, opened 2016-06-21
Core/Scripts: Fix array out of bounds in boss_mother_shahraz.cpp.

MisterZok:
**Changes proposed**:
- Fix an error in Prismatic Shield auras and Fatal Attraction teleport points selection (was introduced in https://github.com/TrinityCore/TrinityCore/commit/fb63c8e915986650f787de62859e1f818d1d96c7)

**Target branch(es)**: 335/6x

**Tests performed**: Tested in-game


DDuarte:
Valid, urand is inclusive in both ends.

PR #17447, opened 2016-06-23
Core/MMaps: Fix crash introduced in commit 32e8b65

### LINKED ISSUE #16910: [6.x] Crash regarding MMaps and Terrain Swap
**Description**: This is a crash issue that happens when MMaps are enabled and you try to teleport or enter a location that uses Terrain Swap

**Current behaviour**: The server crashes with the following call stack https://gist.github.com/DJScias/38fa629f607dc6970be96a6bc09850fc (I know it says revision unknown, but git is being a tool. It uses the TC Hash/commit found below)

**Expected behaviour**: The server shouldn't crash and use the proper MMaps in the Terrain Swap.

1. Enable MMaps
2. Enable Terrain Swap on a location like Lion's Landing, Dominion Point , this is done through the DB table terrain_swap_defaults and terrain_phase_info. (If you don't know the TerrainSwapMap id's, it's 1061 for Dominion Point and 1062 for Lion's Landing)
3. Go to said location that uses Terrain Swaps (like Lion's Landing, Dominion Point), 
4. Get instantly crashed with the above crashlog

**Branch(es)**: 6x

**TC hash/commit**: https://github.com/TrinityCore/TrinityCore/commit/363731aac827f3eab051064ba466cef49f6cc111 I'm still using 6.2.3, reluctant to go 6.2.4 yet. But this would not be fixed even if I was on 6.2.4

**TDB version**: latest TDB 6.03

**Operating system**: Windows 10 Pro, 64-bit

et65:
**Changes proposed**:
- Fix crash introduced in commit 32e8b6506fc064fdba134d692dfe99298e43f055 related to mmaps and terrain swap.
- Also fix remaining memory leaks scenarios in MMapManager.
- Now map base tiles are loaded at map loading, like phased tiles, and unloaded at map unloading (also like phased tiles). Non phased maps are not loaded into base tiles attribute but when unloading map, retrieved by "dtNavMesh::removeTile" data parameter and free by "dtFree" function. No more usage of `DT_TILE_FREE_DATA` which allowed recast navigation to auto free datas at tile remove that is not what we want.

**Target branch(es)**: 6x

**Issues addressed**: Closes #16910

**Tests performed**: Move in game through many locations. Also tested scenario described in the issue.

### DIFF
--- a/src/common/Collision/Management/MMapManager.cpp
+++ b/src/common/Collision/Management/MMapManager.cpp
@@ -282,7 +282,7 @@ namespace MMAP
             if (dataItr != phasedTileItr->second.end())
             {
                 TC_LOG_DEBUG("phase", "MMAP:UnloadPhaseTile: Unloaded phased %04u%02i%02i.mmtile for root phase map %u", phaseMapId, x, y, phasedMapData->first);
-                delete dataItr->second->data;
+                dtFree(dataItr->second->data);
                 delete dataItr->second;
                 phasedTileItr->second.erase(dataItr);
             }
@@ -589,7 +589,7 @@ namespace MMAP
         auto itr = _baseTiles.find(packedGridPos);
         if (itr != _baseTiles.end())
         {
-            delete itr->second->data;
+            dtFree(itr->second->data);
             delete itr->second;
             _baseTiles.erase(itr);
         }
--- a/src/common/Collision/Management/MMapManager.cpp
+++ b/src/common/Collision/Management/MMapManager.cpp
@@ -179,7 +179,7 @@ namespace MMAP
         dtTileRef tileRef = 0;
 
         // memory allocated for data is now managed by detour, and will be deallocated when the tile is removed
-        if (dtStatusSucceed(mmap->navMesh->addTile(data, fileHeader.size, DT_TILE_FREE_DATA, 0, &tileRef)))
+        if (dtStatusSucceed(mmap->navMesh->addTile(data, fileHeader.size, 0, 0, &tileRef)))
         {
             mmap->loadedTileRefs.insert(std::pair<uint32, dtTileRef>(packedGridPos, tileRef));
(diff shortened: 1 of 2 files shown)


src/common/Collision/Management/MMapManager.cpp
```diff
@@ -518,32 +523,21 @@ namespace MMAP
 
         // header xy is based on the swap map's tile set, wich doesn't have all the same tiles as root map, so copy the xy from the orignal header
         header->x = oldTile->header->x;
-        header->y = oldTile->header->y;
+        header->y = oldTile->header->y;;
```
Aokromes: double ;;

et65: Bad copy paste, thanks x)

jackpoz:
did you test this using a memory leak detector like valgrind ?

et65:
@jackpoz : No, I'm a little busy these days.

Can anyone do the test?

et65:
@jackpoz : I've just run Dr Memory, and no more memory leaks concerning mmaps. Just a warning fixed in a new commit.

jackpoz:
I doubt users will run Dr Memory/valgrind so it's more of debugging it line but line and checking with valgrind all the possible cases to ensure no new leaks/crashes are added

et65:
@jackpoz : It's what I've done and no more memory leaks ;)

DJScias:
@et65 Hello there,

Sorry for the late responses, github didn't beep me.
I'll apply these changes to my local and see what happens.

Kind regards,
DJScias

DJScias:
@et65 Hello there,

Can confirm the crash is gone, there's still a bug related to pathwalking of the NPCs in regards of terrain swaps though but that wasn't part of the issue.

Kind regards,
DJScias

et65:
@DJScias : Thanks for your feedback.

PR #17455, opened 2016-06-25
Core/Event: prevent seasonal quests from resetting at server restart

### LINKED ISSUE #16015: [3.3.5] Core/Quests: Seasonal Quests Reset
The reset of seasonal quests isn't working properly, tested in Winter's Veil.

The player can accept the quest, but can't complete or reward it.

Even if you complete and reward forced, with GM commands, the next quest won't be offered.

SS:
wowscrnshot_121615_155645

Branch 3.3.5
Hash cc685ac334335d9f6a25a7c1c739a905f919328f

### LINKED ISSUE #16413: [3.3.5][6.x] Core/Event: On core restart, Lunar Fesitval Elders give out quests/coins again
TrinityCore rev. 3f886025a255 2016-01-22 22:14:28 +0100 (3.3.5 branch) (Unix, Release)

I was testing a smart script patch for a different ticket when I noticed that Lunar Festival Elders don't think I've turned in the quest for their individual coins after I restart the server.

Noticed on:
Elder Bladeleaf (http://www.wowhead.com/npc=15595/elder-bladeleaf)
Elder Stormbrow (http://www.wowhead.com/npc=15565/elder-stormbrow)

I have no reason to believe these to be unique, but I can test others further if we need it.

I'm guessing this has to do with it being a "roughly" yearly event.

SnapperRy:
**Changes proposed**:
- Only reset seasonal quests at event startup if event manager is active (after server startup). Prevents seasonal quests from resetting at server restart.

This pr uses worldstates to check if an event has already been started.
- When an event is started, its worldstate is set to the current unix timestamp.
- When an event ends (automatically stopped by the event manager or via commands), the worldstate is set to 0.
- At every event manager update tick (which happens whenever an event is set to be started or stopped), every inactive event has its worldstate set to 0. This allows for cleanup of events that end while the server is offline,

**Target branch(es)**: 335 (but should be cherrypicked to 6.x too).

**Issues addressed**:
Closes #16015.
Closes #16413.

### DIFF
--- a/src/server/game/Events/GameEventMgr.cpp
+++ b/src/server/game/Events/GameEventMgr.cpp
@@ -139,6 +139,9 @@ bool GameEventMgr::StartEvent(uint16 event_id, bool overwrite)
             if (data.end <= data.start)
                 data.end = data.start + data.length;
         }
+
+        // When event is started, set its worldstate to current time
+        sWorld->setWorldState(event_id, time(NULL));
         return false;
     }
     else
@@ -173,6 +176,9 @@ void GameEventMgr::StopEvent(uint16 event_id, bool overwrite)
 
     RemoveActiveEvent(event_id);
     UnApplyEvent(event_id);
+
+    // When event is stopped, clean up its worldstate
+    sWorld->setWorldState(event_id, 0);
 
     if (overwrite && !serverwide_evt)
     {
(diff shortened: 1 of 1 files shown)


Shauren:
mmm, this is no good either. what if the server was offline during the moment event should start?

SnapperRy:
I'm not sure how that case can be covered, as the server cannot know if the event is starting for the first time or not.

Maybe something can be done with the existing game_event_save characters table, will try this evening.

ariel-:
I think maybe just add an extra column on game_event with status, if server starts and flag is set, it means the event was running when server stopped, when event ends flag is reset.

SnapperRy:
@ariel-: that would use the world database to save temporary data, it's a very big no I guess :P

Using the characters database for it is needed, either as worldstate or as a new table. But since this falls into the category of what worldstates exist for, I think it makes sense.

PR #17464, opened 2016-06-26
Event/Midsummer: improve pole ribbon functionality

SnapperRy:
**Changes proposed**:
- Bunny creature should hover, which puts it in the intended position.
- Use correct visual and internal spells to handle cosmetics and timers.

**Target branch(es)**: 335

- [x] The [bunny npc](http://www.wowhead.com/npc=17066/ribbon-pole-debug-target) should alternate between casting [this spell](http://www.wowhead.com/spell=46842/flame-ring) and [this other one](http://www.wowhead.com/spell=46836/flame-patch) roughly every five seconds, as long as atleast one player is using the ribbon pole. Haven't been able to find a not-abominable way to implement it yet.
- [x] [This other spell](http://www.wowhead.com/spell=45407/reveler-applause-cheer) is triggered at set intervals, but functionality is (probably?) not yet implemented. (this is unrelated to this PR though, should be fixed separately)

### DIFF
--- a/src/server/scripts/World/go_scripts.cpp
+++ b/src/server/scripts/World/go_scripts.cpp
@@ -1197,6 +1197,28 @@ public:
     }
 };
 
+enum MidsummerPoleRibbon
+{
+    SPELL_POLE_DANCE      = 29726,
+    SPELL_BLUE_FIRE_RING  = 46842,
+    NPC_POLE_RIBBON_BUNNY = 17066
+};
+
+class go_midsummer_ribbon_pole : public GameObjectScript
+{
+public:
+    go_midsummer_ribbon_pole() : GameObjectScript("go_midsummer_ribbon_pole") { }
+
+    bool OnGossipHello(Player* player, GameObject* go) override
+    {
+        if (Creature* creature = go->FindNearestCreature(NPC_POLE_RIBBON_BUNNY, 10.0f))
+        {
+            creature->CastSpell(go, SPELL_BLUE_FIRE_RING, true);
+            player->CastSpell(creature, SPELL_POLE_DANCE);
+        }
+        return true;
+    }
+};
 
 enum ToyTrainSpells
 {
@@ -1274,5 +1296,6 @@ void AddSC_go_scripts()
     new go_veil_skith_cage();
     new go_frostblade_shrine();
     new go_midsummer_bonfire();
+    new go_midsummer_ribbon_pole();
     new go_toy_train_set();
 }
--- a/src/server/scripts/World/go_scripts.cpp
+++ b/src/server/scripts/World/go_scripts.cpp
@@ -1213,8 +1213,8 @@ public:
     {
         if (Creature* creature = go->FindNearestCreature(NPC_POLE_RIBBON_BUNNY, 10.0f))
         {
-            creature->CastSpell(go, SPELL_BLUE_FIRE_RING, true);
-            player->CastSpell(creature, SPELL_POLE_DANCE);
+            creature->GetAI()->SetData(0, 0);
+            player->CastSpell(creature, SPELL_POLE_DANCE, true);
         }
         return true;
     }
--- a/src/server/scripts/World/go_scripts.cpp
+++ b/src/server/scripts/World/go_scripts.cpp
@@ -1201,7 +1201,8 @@ enum MidsummerPoleRibbon
 {
     SPELL_POLE_DANCE      = 29726,
     SPELL_BLUE_FIRE_RING  = 46842,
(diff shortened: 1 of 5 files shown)


horn:
There are two ways to do it:
1) AI script - NPC / go can handle the timers itself and cast proper spells 
2) as I can see, both spells have 5 secs duration - you can add scripts to them and cast another spell in OnRemove hook on expiration only

SnapperRy:
@horn: thanks for the suggestions! Couldn't get OnRemove() hook to work for those spells, so I'm going for a core script.

src/server/scripts/World/npcs_special.cpp
```diff
@@ -580,6 +580,107 @@ class npc_torch_tossing_target_bunny_controller : public CreatureScript
...
+};
+
+class npc_midsummer_bunny_pole : public CreatureScript
+{
+public:
+    npc_midsummer_bunny_pole() : CreatureScript("npc_midsummer_bunny_pole") { }
+
+    struct npc_midsummer_bunny_poleAI : public ScriptedAI
+    {
+        npc_midsummer_bunny_poleAI(Creature* creature) : ScriptedAI(creature) { Reset(); }
```
jackpoz: please avoid calling virtual functions inside the constructor

src/server/scripts/World/go_scripts.cpp
```diff
@@ -1213,8 +1213,8 @@ class go_midsummer_ribbon_pole : public GameObjectScript
     {
         if (Creature* creature = go->FindNearestCreature(NPC_POLE_RIBBON_BUNNY, 10.0f))
         {
-            creature->CastSpell(go, SPELL_BLUE_FIRE_RING, true);
-            player->CastSpell(creature, SPELL_POLE_DANCE);
+            creature->GetAI()->SetData(0, 0);
```
Tuzka: Why not use DoAction instead of SetData ? Confusing arguments imho.

SnapperRy: Changed, thanks :) didn't know about DoAction.

src/server/scripts/World/npcs_special.cpp
SnapperRy: Right, I thought I saw reset functions being called there, but checking again I found none. Changed, thanks.

src/server/scripts/World/npcs_special.cpp
```diff
@@ -580,6 +580,107 @@ class npc_torch_tossing_target_bunny_controller : public CreatureScript
...
+                if (checkNearbyPlayers())
+                {
+                    Reset();
+                    return;
+                }
+
+                if (GameObject* go = me->FindNearestGameObject(GO_RIBBON_POLE, 10.0f))
+                    me->CastSpell(go, SPELL_RED_FIRE_RING, true);
+
+                events.ScheduleEvent(EVENT_CAST_BLUE_FIRE_RING, 5000);
```
Keader: events.ScheduleEvent(EVENT_CAST_BLUE_FIRE_RING, Seconds(5));

src/server/scripts/World/npcs_special.cpp
```diff
@@ -580,6 +580,107 @@ class npc_torch_tossing_target_bunny_controller : public CreatureScript
...
+                if (checkNearbyPlayers())
+                {
+                    Reset();
+                    return;
+                }
+
+                if (GameObject* go = me->FindNearestGameObject(GO_RIBBON_POLE, 10.0f))
+                    me->CastSpell(go, SPELL_BLUE_FIRE_RING, true);
+
+                events.ScheduleEvent(EVENT_CAST_RED_FIRE_RING, 5000);
```
Keader: events.ScheduleEvent(EVENT_CAST_RED_FIRE_RING, Seconds(5));

src/server/scripts/World/npcs_special.cpp
SnapperRy: Good, I thought it was only for TaskManager :P thanks.

src/server/scripts/World/npcs_special.cpp
jackpoz: Usually there's a Initialize() function that sets the fields to default values called from the ctor and from Reset(), so you don't have to duplicate code.
Please make sure all class fields are initialized (i.e. "running")

SnapperRy: Changed again, thanks! Will keep that in mind.

Aokromes:
Sooo this can be merged?

jackpoz:
it should, was it tested again after all recent changes ?

SnapperRy:
Yep, tested and working.

PR #17469, opened 2016-06-27
Core/Loading: output error when specified emote id for some quest-related tables is invalid

SnapperRy:
**Changes proposed**: title says it all.

**Target branch(es)**: 335

**Tests performed**:  Compiles and works.

### DIFF
--- a/src/server/game/Quests/QuestDef.cpp
+++ b/src/server/game/Quests/QuestDef.cpp
@@ -133,9 +133,9 @@ void Quest::LoadQuestDetails(Field* fields)
 {
     for (int i = 0; i < QUEST_EMOTE_COUNT; ++i)
     {
-        if (fields[1+i].GetUInt16() >= EMOTE_MAX)
+        if (!sEmotesStore.LookupEntry(fields[1+i].GetUInt16()))
         {
-            TC_LOG_ERROR("sql.sql", "Table `quest_details` has non-existing Emote%u %u set for quest %u. Skipped.", 1+i, fields[1+i].GetUInt16(), fields[0].GetUInt32());
+            TC_LOG_ERROR("sql.sql", "Table `quest_details` has non-existing Emote%i (%u) set for quest %u. Skipped.", 1+i, fields[1+i].GetUInt16(), fields[0].GetUInt32());
             continue;
         }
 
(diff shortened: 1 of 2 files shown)


src/server/game/Quests/QuestDef.cpp
```diff
@@ -142,13 +150,25 @@ void Quest::LoadQuestRequestItems(Field* fields)
...
     RequestItemsText = fields[3].GetString();
 }
 
 void Quest::LoadQuestOfferReward(Field* fields)
 {
     for (int i = 0; i < QUEST_EMOTE_COUNT; ++i)
+    {
+        if (fields[1+i].GetUInt16() >= EMOTE_MAX)
+        {
+            TC_LOG_ERROR("sql.sql", "Table `quest_offer_reward` has non-existing Emote%u %u set for quest %u. Skipped.", 1+i, fields[1+i].GetUInt16(), fields[0].GetUInt32());
```
Carbenium: Same as above

src/server/game/Quests/QuestDef.cpp
```diff
@@ -132,7 +132,15 @@ Quest::Quest(Field* questRecord)
 void Quest::LoadQuestDetails(Field* fields)
 {
     for (int i = 0; i < QUEST_EMOTE_COUNT; ++i)
+    {
+        if (fields[1+i].GetUInt16() >= EMOTE_MAX)
```
Shauren: Don't define EMOTE_MAX and don't check it like this
Verify the emote exists in Emotes.dbc instead (we load it, lookup sEmotesStore)

src/server/game/Quests/QuestDef.cpp
```diff
@@ -132,7 +132,15 @@ Quest::Quest(Field* questRecord)
...
+        {
+            TC_LOG_ERROR("sql.sql", "Table `quest_details` has non-existing Emote%u %u set for quest %u. Skipped.", 1+i, fields[1+i].GetUInt16(), fields[0].GetUInt32());
```
jackpoz: why is %i suggested for uint16() ?

Carbenium: Because the variable ```i``` is an int

jackpoz: lul TC code is not even consistent about `%i` and `%d`

PR #17488, opened 2016-07-02
Core/Spell: allow use of ground mounts while swimming

SnapperRy:
**Changes proposed**: title says it all. Ground (but not flying) mounts should be usable while swimming. Client checks are quite accurate in this case, and allow ground mounts when in water. Also, source [here](http://wowwiki.wikia.com/wiki/Swim?oldid=2300441): `Unlike ground mounts, flying mounts cannot be mounted while in water.`

Proof [here](https://youtu.be/FlGxBOs2YTw?t=22)! (thanks @Aokromes!)

**Target branch(es)**: 335

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -5498,7 +5498,7 @@ SpellCastResult Spell::CheckCast(bool strict)
             }
             case SPELL_AURA_MOUNTED:
             {
-                if (m_caster->IsInWater())
+                if (m_caster->IsInWater() && m_spellInfo->HasAura(SPELL_AURA_MOD_INCREASE_MOUNTED_FLIGHT_SPEED))
                     return SPELL_FAILED_ONLY_ABOVEWATER;
 
                 // Ignore map check if spell have AreaId. AreaId already checked and this prevent special mount spells


src/server/game/Spells/Spell.cpp
```diff
@@ -5498,7 +5498,7 @@ SpellCastResult Spell::CheckCast(bool strict)
             }
             case SPELL_AURA_MOUNTED:
             {
-                if (m_caster->IsInWater())
+                if (m_caster->IsInWater() && m_spellInfo->HasAura(SPELL_AURA_MOD_INCREASE_MOUNTED_FLIGHT_SPEED))
```
Takenbacon: I would suggest just moving this statement if it's correct to the SPELL_AURA_MOD_INCREASE_MOUNTED_FLIGHT_SPEED case ~30 lines below.

On an unrelated note: I could swear I remember it only letting me mount while I was at the top of the water line on retail during WoTLK, however I cannot find any proof of this. Older patch perhaps or am I just not remembering things correctly?

tkrokli: @Takenbacon : I remember running around on a DK character, exploring while riding on [[Path of Frost](http://www.wowhead.com/spell=3714/path-of-frost)] on the sea. When I for some reason found myself submerged in the water, I used to jump up on top of the Path of Frost again to mount. Not sure if that changed in one of the latest patches to WoTLK, but I remember it like you do.

Goatform: I would agree with @Takenbacon. I think that this patch is only valid for 6.x.x branch and not for 3.3.5 patch but I may be wrong again.

SnapperRy: Proof [here](https://youtu.be/FlGxBOs2YTw?t=22) :P

tkrokli: OK, good point. Pity that it wasn't better documented in the  3.1.0 patch notes.

PR #17516, opened 2016-07-06
Core/Mmaps: Charge ignoring z position of target

### LINKED ISSUE #15011: [3.3.5][6.x][MMAPS] Charge ignoring z position of target
It allow caster to charge his target who's e.g. 5 yards above caster without any object (any path) to reach that target. On retail if your target is above/under you for 5 yards or more and there is no availabe path for you it will sent you "No path" error message. On TC it will let you to charge your target ignoring z position.

Here are few examples:

This is how charge works on TC:
8jcdsrv

- So ask you can see on that picture, it let you to charge target who's ... probably flying or something like that without checking z position. And on picture you can see that there are two lines which means that sometimes it will let you to "fly and charge target" and sometimes you will charge under him in "same path of your target".

This is how charge should to work (retail):
sqmrdyo

- It shouldn't let you to charge in at all and it will sent you "No path" error.

This is situation when you are able to charge your target who's above you 5y or more.
qx37pnp

- So only if you have object (availabe path) for charge, it will let you to do that.

TESTED ON:
3.3.5 : dfc7176
6.x : d418ae5

Aokromes:
**Target branch(es)**: 335/6x

**Issues addressed**: Closes #15011

By cemak

### DIFF
--- a/src/server/game/Movement/PathGenerator.cpp
+++ b/src/server/game/Movement/PathGenerator.cpp
@@ -943,7 +943,7 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
     }
 }
 
-bool PathGenerator::InValidPathZ(Unit* target)
+bool PathGenerator::IsInvalidDestinationZ(Unit* target)
 {
     G3D::Vector3 endPos = GetActualEndPosition();
 
--- a/src/server/game/Movement/PathGenerator.cpp
+++ b/src/server/game/Movement/PathGenerator.cpp
@@ -942,3 +942,10 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
         nextVec = currVec; // we're going backwards
     }
 }
+
+bool PathGenerator::InValidPathZ(Unit* target)
+{
+    G3D::Vector3 endPos = GetActualEndPosition();
+
+    return (target->GetPositionZ() - endPos.z) > 5.f;
+}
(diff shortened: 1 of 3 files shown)


src/server/game/Movement/PathGenerator.cpp
```diff
@@ -942,3 +942,10 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
         nextVec = currVec; // we're going backwards
     }
 }
+
+bool PathGenerator::InValidPathZ(Unit* target)
```
Shauren: Rename this - "InValid" is not correct here, I suggest "IsInvalidDestinationZ"

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -942,3 +942,10 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
         nextVec = currVec; // we're going backwards
     }
 }
+
+bool PathGenerator::IsInvalidDestinationZ(Unit* target)
+{
+    G3D::Vector3 endPos = GetActualEndPosition();
+
+    return (target->GetPositionZ() - endPos.z) > 5.f;
```
Shauren: replace both lines with `(target->GetPositionZ() - GetActualEndPosition().z) > 5.0f;` and add proper const to function (don't forget adjusting declaration in PathGenerator.h)
`bool PathGenerator::IsInvalidDestinationZ(Unit const* target) const`

for final effect
```
bool PathGenerator::IsInvalidDestinationZ(Unit const* target) const
{
    (target->GetPositionZ() - GetActualEndPosition().z) > 5.0f;
}
```

PR #17520, opened 2016-07-06
Core/Unit: remove duplicate resist roll in Unit::MeleeSpellHitResult

ariel-:
**Changes proposed**: Remove duplicate mechanic resist roll in melee spell hit function
- Resist was being calculated twice for melee spells, GetMechanicResistChance function body was doing the same than the misc code removed, but also there was a GetMechanicResistChance call!
  result: two spell resist calc, and twice the chance for a spell to be (unfairly?) resisted.

**Target branch(es)**: 335/6x

**Tests performed**: Builds correctly


Shauren:
Nice find.

PR #17525, opened 2016-07-06
Core/Instances: implemented handling for SMSG_BOSS_KILL_CREDIT

Ovahlord:
**Changes proposed**:
- Adds packet structure for the opcode and implements a instancescript function to call it via internal script or directly within a boss script that can access the instancescript.

**Target branch(es)**: 6x

**Tests performed**: Tested ingame, works as intended

The opcode connection type has not been changed to instance because the opcode can also be sent to players within the open world for example by killing Supreme Lord Kazzak.
The packet structure itself was not in the client function name list that I got via gist so I used the most common names of that list

The following screenshot has been taken by using my IronDocks branch at https://github.com/Ovahlord/TrinityCore/tree/IronDocks

wowscrnshot_070516_121231


src/server/game/Server/Packets/InstancePackets.h
```diff
@@ -236,6 +236,15 @@ namespace WorldPackets
             int32 InCombatResCount = 0;
             uint32 CombatResChargeRecovery = 0;
         };
+
+        class InstanceBossKillCredit final : public ServerPacket
+        {
+        public:
+            InstanceBossKillCredit() : ServerPacket(SMSG_BOSS_KILL_CREDIT, 0) { }
```
MitchesD: just a small thing, use size 4 instead of 0 as default packet size

Shauren:
CONNECTION_INSTANCE does not strictly mean inside instance maps but map/world connection

Shauren:
Why not automate sending it by placing the call in this method https://github.com/TrinityCore/TrinityCore/blob/6.x/src/server/game/Instances/InstanceScript.cpp#L680? You have access to DungeonEncounter dbc entry there

src/server/game/Server/Packets/InstancePackets.h
```diff
@@ -236,6 +236,15 @@ namespace WorldPackets
             int32 InCombatResCount = 0;
             uint32 CombatResChargeRecovery = 0;
         };
+
+        class InstanceBossKillCredit final : public ServerPacket
```
Shauren: also please rename this class to BossKillCredit as in SMSG_BOSS_KILL_CREDIT (there is no INSTANCE word here)

Ovahlord:
@Shauren I do not automate it because there are many boss encounters that does not use that handling like the Black Temple and some other bosses.

Shauren:
Maybe its only used for wod+?

Ovahlord:
It is used for everything that has been reworked with MoP and also some Cataclysm raid bosses like Al'Akir use that mechanic. Looks to me like Blizz did not finish their job in upgrading all codes because it is all random.

Ovahlord:
I would still keep that instance function at place and maybe add another one to send it to single players for world bosses and to send it for non-tracked encounters like in Blackwing Descent.

Shauren:
While it is useful to have a method allowing to send it for world bosses, its current implementation is incorrect because only players that will receive credit for the kill should get this notification, not every nearby player. Seeing as we currently don't have a way to support that condition it would be best to not add an incorrect implementation - please revert Unit::SendBossKillCredit

PR #17528, opened 2016-07-06
Core/Fishing: implement retail-like fishing skill-up functionality

SnapperRy:
**Changes proposed**:

On Retail, ever since the beginning of time (probably?), the chance of gaining a skill point in Fishing isn't really random. Instead, you get a skill point after a certain amount of successful attempts, which is based on your current skill level.

The amount of successful fishing attempts required to skill up is not linear or esponential. For more information, read [here](https://web.archive.org/web/20100305183419/http://www.elsanglin.com/role_of_skill.html) and [here](http://wowwiki.wikia.com/wiki/Fishing?oldid=2356392) (the attempts-per-skill-points tables are the most useful data there, although the two sources conflict. I think El's Extreme Anglin' is more trustworthy, due to being entirely dedicated to Fishing in WoW).

This PR uses guessed formulas to get as close as possible (well, as close as I can) to the retail's skill-up curve. [This is the result](http://pastebin.com/XgrCabUi).

**Target branch(es)**: 335


PiskelDEV:
someone know how 6.x works ?

PR #17529, opened 2016-07-06
[6.x] Fix camera and player collision not updating when object scale changes

### LINKED ISSUE #9285: [6.x] Camera height and collision box
TrinityCore rev. 2013-02-22 23:10:30 +0100 (2672d134436b)
DB rev TDB_full_434.04_2013_02_14 with updates

Hello, there is a problem with the camera and the player's character collision box when the character grow or shrink or when the character is mounted. I'm not sure if it is a known issue. This doesn't happening on blizzard's server.
It seems that the camera and the collision box aren't updated until a logout or a map change.

screen1

As you can see above the camera it at the head level.
But when the character grow with spells like "Elite" "Experienced" or "Unholy Growth" for example, the camera's height stay the same.
The same applies to the collision box, the character can pass through objects.

screen2
screen3

But after a logout or a map change the camera is updated at the head position with the collision box.

screen4
screen5

A same thing happen with mounts, after a dismount the camera goes to the ground.

screen6
screen7

And after a logout or a map change the camera come back to the head.

screen8

I remember a long time ago that the camera and the collision box were updated immediately after the character modification.

Voilà, it's not an important issue but I hope this will be useful.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/112491-4-3-4-camera-height-and-collision-box?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Krudor:
**Changes proposed**:
- Send collision height update when object scale changes

**Target branch(es)**: 335/6x

**Issues addressed**: Closes #9285

**Tests performed**: Build and in-game tested, would appriciate if more people could test and verify.

Does this apply to 3.3.5 branch as well maybe?


Shauren:
This method is large to deserve a move to Player.cpp

Krudor:
Agreed, shall I make the changes?

Shauren:
yep

src/server/game/Entities/Player/Player.cpp
```diff
@@ -1761,6 +1761,15 @@ void Player::RemoveFromWorld()
     }
 }
 
+void Player::SetObjectScale(float scale)
+{
+    Unit::SetObjectScale(scale);
+    SetFloatValue(UNIT_FIELD_BOUNDINGRADIUS, scale * DEFAULT_WORLD_OBJECT_SIZE);
+    SetFloatValue(UNIT_FIELD_COMBATREACH, scale * DEFAULT_COMBAT_REACH);
+    if (IsInWorld())
+        SendMovementSetCollisionHeight(scale * GetCollisionHeight(false));
```
Shauren: now then, scale can be adjusted while mounted too so just pass IsMounted() instead of always false

Krudor: Ooo right!

tkrokli:
Nice, thanks for the contribution! :+1:

PR #17543, opened 2016-07-07
Core/Creature: implement INHABIT_ROOT to apply UNIT_STATE_ROOT to creature

### LINKED ISSUE #17507: [3.3.5][Quest] "It's All Fun and Games" stopped working
**Description**: The eye pursues the player, it needs to stand still.

**Current behaviour**: the eyes pursuing the player when he was attacked

**Expected behaviour**: the eyes should stand still

1. quest 12887 and 12892
2. http://www.wowhead.com/quest=12892/its-all-fun-and-games

**Branch(es)**:
3.3.5

**TC hash/commit**:  
TrinityCore rev [1ad7321](https://github.com/TrinityCore/TrinityCore/commit/1ad73212dca0cf8a829d15ffdbcc4cd611e64d4e)
**TDB version**:  
335.61
**Operating system**:  
Debian

previously, the quest worked
not working disable_move
please help

SnapperRy:
**Changes proposed**:

Old UNIT_FLAG_DISABLE_MOVE used to handle this. However, after searching some sniffs, it turns out this flag (currently UNIT_FLAG_REMOVE_CLIENT_CONTROL) is not present on some creatures that should be rooted. As such, flags_extra is a good idea I guess (better than using SmartAI/Core script to add UNIT_STATE_ROOT, atleast).

**Target branch(es)**: 335

**Issues addressed**: Closes #17507.


Shauren:
UNIT_FLAG_DISABLE_MOVE? Never did. It never meant "disable move" and only had any visible ingame effect before wotlk (how long before wotlk, that i dont know)

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -481,6 +481,9 @@ bool Creature::UpdateEntry(uint32 entry, CreatureData const* data /*= nullptr*/)
         ApplySpellImmune(0, IMMUNITY_EFFECT, SPELL_EFFECT_ATTACK_ME, true);
     }
 
+    if (cInfo->flags_extra & CREATURE_FLAG_EXTRA_ROOT)
+        AddUnitState(UNIT_STATE_ROOT);
```
Shauren: AddUnitState is a wrong method to call here, it should be SetControlled(true, UNIT_STATE_ROOT);

Shauren:
Also not sure if flags_extra is the right way to implement it, maybe new flag to InhabitType?

SnapperRy:
It was used to skip movement in the chase/follow movegens [here](https://github.com/TrinityCore/TrinityCore/commit/b36da77#diff-7f0072ea7e8aa2a3e1d6580a7e14dfdeL244) and a few lines below. Not its intended behavior, but people added that flag to root creatures (checked sniffs for some of them, most that have it in db don't have it on sniffs).

PR #17544, opened 2016-07-08
Core/Items: bags will now get equipped automaticly

Ovahlord:
**Changes proposed**:
- Bags that has been looted, purchased and are not getting soulbound on equip and quest rewards will now get equipped automaticly as it is a default mechanic since patch 4.x

**Target branch(es)**: 6x

**Tests performed**: Tested ingame, works as intended.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -11662,7 +11662,7 @@ Item* Player::StoreNewItem(ItemPosCountVec const& pos, uint32 itemId, bool updat
 
         // Since 4.x bags are getting equipped automaticly when they have no bind on equip boundary
         // This includes looted bags, bags from quest rewards and purchased bags from vendors
-        if (item->IsBag() && item->GetTemplate()->GetBonding() != BIND_WHEN_EQUIPED)
+        if (item->IsBag() && item->GetTemplate()->GetBonding() != BIND_ON_EQUIP)
             item = StoreItem(pos, item, update, true);
         else
             item = StoreItem(pos, item, update, false);
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -11660,7 +11660,12 @@ Item* Player::StoreNewItem(ItemPosCountVec const& pos, uint32 itemId, bool updat
         for (int32 bonusListID : bonusListIDs)
             item->AddBonuses(bonusListID);
 
-        item = StoreItem(pos, item, update);
+        // Since 4.x bags are getting equipped automaticly when they have no bind on equip boundary
+        // This includes looted bags, bags from quest rewards and purchased bags from vendors
+        if (item->IsBag() && item->GetTemplate()->GetBonding() != BIND_WHEN_EQUIPED)
+            item = StoreItem(pos, item, update, true);
+        else
+            item = StoreItem(pos, item, update, false);
 
         if (allowedLooters.size() > 1 && item->GetTemplate()->GetMaxStackSize() == 1 && item->IsSoulBound())
         {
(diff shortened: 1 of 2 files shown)


tkrokli:
You may want to verify your git e-mail address for your commits to be identified as yours.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -11076,12 +11081,24 @@ Item* Player::StoreNewItem(ItemPosCountVec const& pos, uint32 itemId, bool updat
...
     Item* lastItem = pItem;
+
+    if (lootedBag && pItem->GetTemplate()->GetClass() == ITEM_CLASS_CONTAINER && pItem->GetTemplate()->GetSubClass() == ITEM_SUBCLASS_CONTAINER)
+        for (uint32 i = INVENTORY_SLOT_BAG_START; i < INVENTORY_SLOT_BAG_END; i++)
+        {
+            if (!m_items[i])
+            {
+                EquipNewItem(i, pItem->GetEntry(), true);
+                return lastItem;
+                continue;
```
jackpoz: please either return or continue, you can't do both

Shauren:
I disagree with this solution, it would be better to place it in
`InventoryResult Player::CanStoreItem(uint8 bag, uint8 slot, ItemPosCountVec& dest, uint32 entry, uint32 count, Item* pItem = nullptr, bool swap = false, uint32* no_space_count = nullptr) const;`
changing first argument in the last CanStoreItem_InInventorySlots call to INVENTORY_SLOT_BAG_START (if the new item doesnt bind on equip and is a regular bag)

Ovahlord:
Lemme check this out later. I just followed the core routine to that point and thought that this place was the best one. I gonna try your way later on and see what is gonna happen

Shauren:
Thats understandable since `CanStoreItem` doesn't only check if the item can be stored but also gives you a possible store slot

Ovahlord:
Is that call getting triggered only on looting / getting items? Because it is not suposed to auto-equip the bags if they have been taken out of the bank for example.

Aokromes:
Any new on this?

Aokromes:
Conflicting files
src/server/game/Entities/Player/Player.h

Aokromes:
@Ovahlord  can you fix conflict?

src/server/game/Entities/Player/Player.cpp
Aokromes: @Ovahlord

Aokromes:
@Ovahlord can you fix conflict?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -11660,7 +11660,12 @@ Item* Player::StoreNewItem(ItemPosCountVec const& pos, uint32 itemId, bool updat
         for (int32 bonusListID : bonusListIDs)
             item->AddBonuses(bonusListID);
 
-        item = StoreItem(pos, item, update);
+        // Since 4.x bags are getting equipped automaticly when they have no bind on equip boundary
+        // This includes looted bags, bags from quest rewards and purchased bags from vendors
+        if (item->IsBag() && item->GetTemplate()->GetBonding() != BIND_ON_EQUIP)
```
Mihapro: wouldn't it be better to move IsBag() and GetBonding() checks inside Player::StoreItem() since it is called in both cases anyway?

Mihapro: You can use ```pItem->HasFlag(ITEM_FIELD_FLAGS, ITEM_FIELD_FLAG_NEW_ITEM);``` in Player::StoreItem(), no need to have an additional argument lootedBag.

DoctorKraft:
@Ovahlord / @funjoker any news about this pr?

Ovahlord:
No. I tried to port it to a newer revision a few weeks ago but it failed. I personally think this whole slot handling for items need a rewrite with better namings cuz I never felt that lost in that code.

src/server/game/Entities/Player/Player.cpp
Ovahlord: not supported on 335 or 4.x so bad idea.

Issue #15927, opened 2015-12-01
[3.3.5][6.x] DB/Gameobject: Netherwing Egg - Linking Spawn to Trap Netherwing Egg Dust (Visual)


Rushor:
rev: https://github.com/TrinityCore/TrinityCore/commit/3c4f4356298d974c507fbd4c3fa54c1d4075e091

object=185915
trap - object=185600

Issue:

object=185915 (the egg itself) should be linked to object=185600 (the dust visual). Each time The egg 185915 spawn, the dust 185600 should spawn in the same location like the egg.

IMPORTANT: We cannot simply add a dbspawn here because the eggspawn itself should be pooled.

What do you tried to fix the issue:
I already tried to add dbspawn and inserts in https://www.trinitycore.info/display/tc/linked_respawn ,  but this does not solved the issue.

Basically we need a script which executes two things:
1) If the egg 185915 spawn from the pool, the egg-go should spawn the dustgo 185600
2) If the egg-go 185915 was looted and changes it lootstate, the dustgo 185600 should despawn.

Problem A) Moreover each time if an egg 185915 from the pool respawns, the dustgo 185600 must be summoned again.
Problem B) Actually egg-go 185915 contains datas on the trap - object=185600 in the gameobject_template. But somehow the trap will not be spawned.
(Summoning @Gacko for this, because he maybe knows more about it)

Note:
I have no fucking idea how to solve this with SAI. I already worked out a hackshit coresolution but i could not solve Problem A. I will post the script here later.

Labels: Core + DB, not shure where it must/could be fixed

Shauren:
You can use a pool of pools
instead of pooling eggs you pool pools of egg+dust pairs (each egg+dust pair would be in its own separate pool managed by the global one)

Killyana:
Same issue as https://github.com/TrinityCore/TrinityCore/issues/12139

Kittnz:
I guess it's also the same for these mana crystal crates that have a glow on them and if the gob is looted the glow goes away too

Rushor:
Found even more datas in a githubproject which i don't want to link here because i don't know if they like it

Text from the issue:



Issue: Some gameobjects should be scripted to spawn and despawn together with other gameobjects. For example gameobjects with type trap, that should spawn and despawn with the gameobject that triggers the trap. Or gameobjects that are supposed to give visual effect (glow) around another gameobject.

[...]

PR #17553, opened 2016-07-09
Core/Gameobject: handle despawning/respawning of nearby linked traps on gameobject despawn/respawn

### LINKED ISSUE #12139: [3.3.5][GameObject]The Trap Object 181584 do not despawn after cast
core: latest
db: up-to-date

The gameobject 181584 (it's a trap, type 6, cast spell 29420) do not despawn after cast,
even I have changed it data4 = 1 in database.

Do I miss something? Or Is it a bug?
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2340545-3-3-5-gameobject-the-trap-object-181584-do-not-despawn-after-cast?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SnapperRy:
**Changes proposed**:

While we don't know whether or not linked traps are permanently spawned, or if the gameobject itself spawns them, what we know is that they should respawn when the parent gameobject is respawned, and despawned when the parent is despawned.

**Target branch(es)**: 335

**Issues addressed**: Closes #12139, Updates #15927

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -429,9 +429,8 @@ void GameObject::Update(uint32 diff)
                     m_usetimes = 0;
 
                     // If nearby linked trap exists, respawn it
-                    if (uint32 linkedEntry = GetGOInfo()->GetLinkedGameObjectEntry())
-                        if (GameObject* linkedTrap = FindNearestGameObject(linkedEntry, 10.0f))
-                            linkedTrap->SetLootState(GO_READY);
+                    if (GameObject* linkedTrap = GetLinkedTrap())
+                        linkedTrap->SetLootState(GO_READY);
 
                     switch (GetGoType())
                     {
@@ -619,9 +618,8 @@ void GameObject::Update(uint32 diff)
         case GO_JUST_DEACTIVATED:
         {
             // If nearby linked trap exists, despawn it
-            if (uint32 linkedEntry = GetGOInfo()->GetLinkedGameObjectEntry())
-                if (GameObject* linkedTrap = FindNearestGameObject(linkedEntry, 10.0f))
-                    linkedTrap->SetLootState(GO_JUST_DEACTIVATED);
+            if (GameObject* linkedTrap = GetLinkedTrap())
+                linkedTrap->SetLootState(GO_JUST_DEACTIVATED);
 
             //if Gameobject should cast spell, then this, but some GOs (type = 10) should be destroyed
             if (GetGoType() == GAMEOBJECT_TYPE_GOOBER)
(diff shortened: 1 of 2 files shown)


jackpoz:
what happens if there are 2 traps spawned closely ?

robinsch:
They should be spawned on top of each other so nearest is always the correct one even if there are multiple within 10.0f.

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -419,6 +419,11 @@ void GameObject::Update(uint32 diff)
                     m_SkillupList.clear();
                     m_usetimes = 0;
 
+                    // If nearby linked trap exists, respawn it
+                    if (uint32 linkedEntry = GetGOInfo()->GetLinkedGameObjectEntry())
+                        if (GameObject* linkedTrap = FindNearestGameObject(linkedEntry, 10.0f))
```
ariel-: Suggestion: instead of using a generic "shotgun" grid searcher, why don't save spawned linked GO spawn GUID in the GameObject, then just do a ObjectAccessor:GetGameObject(*this, linkedGoGUID) to despawn only the linked one.

SnapperRy:
@ariel-: calling for feedback :P

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -2227,6 +2235,27 @@ bool GameObject::IsLootAllowedFor(Player const* player) const
...
+GameObject* GameObject::GetLinkedTrap()
+{
+    uint32 linkedEntry = GetGOInfo()->GetLinkedGameObjectEntry();
+
+    if (!linkedEntry)
+        return nullptr;
+
+    GameObject* linkedTrap = nullptr;
+
+    if (m_linkedTrap.Empty())
```
ariel-: m_linkedTrap should be set when spawned... see Spell::EffectSummonObjectWild for example

Using a grid searcher can get two nearby gameobjects controlling the same linked go spawn for example

SnapperRy: The issue with that approach is that we have no control over which gameobject is spawned first, if the trap or the "master".

Am I missing something?

Shauren: We do. Master is spawned first, https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellEffects.cpp#L3533 - this code block summons linked trap (and you can link them together here)

SnapperRy: Nice, thanks! Didn't know about that.

Aokromes:
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Entities/GameObject/GameObject.cpp:2247:9: fatal error: 
      type 'const ObjectGuid' does not provide a call operator
    if (m_linkedTrap.Empty())
        ^~~~~~~~~~~~~~~~~~

PR #17556, opened 2016-07-09
Core/Quests: Remove quest items that were looted if they are unique, …

### LINKED ISSUE #15860: [3.3.5] Core/Item: Questitem not deleted after canceling the quest
How to test (example):
.qu add 256 (WANTED: Chok'sul / Alliance quest)
.npc add 1210  (Chok'sul)
Kill him
Pick up the head (questitem) from loot.

Cancel the quest. A window appears and is asking if you want to cancel the quest and to delete the questitem (The head). Say yes.

The questitem should be deleted but is not.

runningnak3d:
**Changes proposed**:
- Properly remove bound quest items when abandoning a quest
- Fix quest item removal on quest fail (wrong define used)

**Target branch(es)**: 335/6x

**Issues addressed**: Closes #15860

- It builds
- .quest add 6
- .add 182
- abandon quest, and Garrick's head is now deleted as it should be

- none

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -15098,11 +15098,11 @@ void Player::FailQuest(uint32 questId)
             SendQuestFailed(questId);
 
         // Destroy quest items on quest failure.
-        for (uint8 i = 0; i < QUEST_SOURCE_ITEM_IDS_COUNT; ++i)
+        for (uint8 i = 0; i < QUEST_ITEM_OBJECTIVES_COUNT; ++i)
             if (quest->RequiredItemId[i] > 0 && quest->RequiredItemCount[i] > 0)
                 // Destroy items received on starting the quest.
                 DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true, true);
-        for (uint8 i = 0; i < QUEST_OBJECTIVES_COUNT; ++i)
+        for (uint8 i = 0; i < QUEST_SOURCE_ITEM_IDS_COUNT; ++i)
             if (quest->ItemDrop[i] > 0 && quest->ItemDropQuantity[i] > 0)
                 // Destroy items received during the quest.
                 DestroyItemCount(quest->ItemDrop[i], quest->ItemDropQuantity[i], true, true);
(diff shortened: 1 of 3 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -15098,17 +15098,34 @@ void Player::FailQuest(uint32 questId)
             SendQuestFailed(questId);
 
         // Destroy quest items on quest failure.
-        for (uint8 i = 0; i < QUEST_OBJECTIVES_COUNT; ++i)
+        for (uint8 i = 0; i < QUEST_SOURCE_ITEM_IDS_COUNT; ++i)
```
Shauren: this was not a mistake
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Quests/QuestDef.h#L283

src/server/game/Entities/Player/Player.cpp
```diff
@@ -15098,17 +15098,34 @@ void Player::FailQuest(uint32 questId)
...
             if (quest->RequiredItemId[i] > 0 && quest->RequiredItemCount[i] > 0)
                 // Destroy items received on starting the quest.
                 DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true, true);
-        for (uint8 i = 0; i < QUEST_SOURCE_ITEM_IDS_COUNT; ++i)
+        for (uint8 i = 0; i < QUEST_OBJECTIVES_COUNT; ++i)
```
Shauren: same as above
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Quests/QuestDef.h#L285

src/server/game/Entities/Player/Player.cpp
runningnak3d: That was my bad. I had an old branch, and forgot that I had already found that I was wrong. However, QUEST_ITEM_OBJECTIVES_COUNT is correct: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Quests/QuestDef.h#L283

runningnak3d:
@Shauren Would it be better if I changed FailQuest to FailOrAbandonQuest with a bool for abandon?

Shauren:
@runningnak3d not sure about that - in 6.x there are new quest flags to control this behavior (and not delete items when abandoning or completing the quest) and I am not sure if they should also work with failing the quest

runningnak3d:
Gotcha -- I will remove 6x from the PR comment.

Shauren:
No need, for 6.x it would be just a matter of extra flags check inside this new method you are adding

PR #17562, opened 2016-07-10
Core/AI: Add better support for tracking DynObjs

Krudor:
**Changes proposed**
- Add functionality for getting ALL dyn/gameobjects summoned by a spell through a unit.
- Add hooks for when dyn/gameobjects are registered to a creature, much like summoned creatures through CreatureAI.

**Target branch(es)**: 6x

**Tests performed**: Builds, haven't tested in-game yet.

I've been scripting some more modern raid encounters, and found a need to get more than just the last summoned dynamicobject linked to a creature, as well as having events for when a dynamicobject is registered.

Went along and added same functionality for gameobjects as well.

This was added to increase scripting capability in the future, this is also a setup for one of my next commits (if time allows).

Would like some feedback on what you think about this change.


src/server/game/Entities/Unit/Unit.h
```diff
@@ -2030,10 +2030,12 @@ class TC_GAME_API Unit : public WorldObject
         void _RegisterDynObject(DynamicObject* dynObj);
         void _UnregisterDynObject(DynamicObject* dynObj);
         DynamicObject* GetDynObject(uint32 spellId);
+        std::list<DynamicObject*> GetDynObjects(uint32 spellId);
```
Shauren: is there any reason for using std::list over std::vector? if not then use vector

Krudor: Chose list because the container used to store the objects, are lists.
What is the advantage of returning a different container in this case?

```
typedef std::list<DynamicObject*> DynObjectList;
DynObjectList m_dynObj;

typedef std::list<GameObject*> GameObjectList;
GameObjectList m_gameObj;
```

Shauren: vector is pretty much faster in every operation - list only makes it easier to erase elements from the middle since it doesnt invalidate any iterators)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -4671,24 +4671,30 @@ void Unit::ApplyStatPercentBuffMod(Stats stat, float val, bool apply)
 void Unit::_RegisterDynObject(DynamicObject* dynObj)
 {
     m_dynObj.push_back(dynObj);
+    if (GetTypeId() == TYPEID_UNIT && ToCreature()->IsAIEnabled)
```
Shauren: IsAIEnabled is a member of Unit class, no need for ToCreature() here

Krudor: Bad habit of looking at other code and figuring as THAT code was commited, it's probably better than what I originaly wrote :P

Krudor:
Running in-game tests with these changes now, In case you want to merge this, I suggest you hold off until I finish testing.

Krudor:
Done doing in-game tests, runs fine.

PR #17565, opened 2016-07-10
Update Object.h

lively158:
**Changes proposed**:
## 
## 
- 

**Target branch(es)**: 335

**Issues addressed**: N/A

**Tests performed**: (Does it build, tested in-game, etc) Tested , works like retail

Changed combat reach to be more blizzlike, instead of hitting pvp players from an insane distance.

### DIFF
--- a/src/server/game/Entities/Object/Object.h
+++ b/src/server/game/Entities/Object/Object.h
@@ -32,7 +32,7 @@
 
 #define CONTACT_DISTANCE            0.5f
 #define INTERACTION_DISTANCE        5.0f
-#define ATTACK_DISTANCE             5.0f
+#define ATTACK_DISTANCE             1.0f
 #define INSPECT_DISTANCE            28.0f
 #define TRADE_DISTANCE              11.11f
 #define MAX_VISIBILITY_DISTANCE     SIZE_OF_GRIDS           // max distance for visible objects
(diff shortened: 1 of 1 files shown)


DDuarte:
Any source for those values?

lively158:
Trial and error whilest comparing to retail servers I'm afriad, no access to sniff devices :(

DDuarte:
I'm afraid we need something better than trial & error in this case. Those values have been there since ever and are quite a fundamental part of the combat system. I think sniffs wouldn't help that much anyway.

jackpoz:
what happened to the PR description ?

tkrokli:
Yes, the title should be more specific, something like
**[3.3.5] Core/Object: normalize Combat Reach distance**

... and then write details about the change in the expected paragraphs in the first PR post, like "Update the combat reach distance values in `Object.h` to realistic distances." or even more specific explanation in the top lines.

Shauren:
Sorry but this is wrong. This is the sniffed value for player's combat reach
```
[0] UNIT_FIELD_COMBATREACH: 1069547520/1.5
```
As you see, it matches current value of DEFAULT_COMBAT_REACH. Next is range calculation itself - autoattacks have the same range as melee abilities that I very recently researched the formula for range where the current value of NOMINAL_MELEE_RANGE constant (5) is also correct and blizzlike.

Summing up, current range IS blizzlike.

PR #17566, opened 2016-07-10
[6.x] Core/Spells: Implement SPELL_AURA_INTERFERE_TARGETTING

Krudor:
**Changes proposed**: Implement SPELL_AURA_INTERFERE_TARGETTING

**Target branch(es)**: 6x

**Tests performed**: Builds, tested in-game.

**Description**: Implements functionality to abilities like smoke bomb / hunter camouflage / other target obscuring effects.
A player affected by this aura effect cannot cast spells on units without the aura(from the same caster) and vice versa. Melee and AOE spells are unaffected by this interference.

EDIT: _Added description by request_


Shauren:
Shouldnt this also affect area spells?

Krudor:
No, AOE effects are unaffected by this

Temppus:
But what about Camouflage ? http://www.wowhead.com/spell=51753/camouflage#comments
Some melee abilities mentioned in comments should be able to cast, like :

Hemorrhage
Multilate.

Krudor:
Very good point, we're not checking if the spell casted is a ranged/not melee spell!

Goatform:
@Krudor, is this something related to rogue's spell Smoke Bomb?

Krudor:
Yes

src/server/game/Spells/Spell.cpp
```diff
@@ -4973,6 +4973,23 @@ SpellCastResult Spell::CheckCast(bool strict)
         if (castResult != SPELL_CAST_OK)
             return castResult;
 
+        // If it's not a melee spell, check if vision is obscured by SPELL_AURA_INTERFERE_TARGETTING
+        if (m_spellInfo->DmgClass != SPELL_DAMAGE_CLASS_MELEE)
```
Temppus: Be careful, this condition should belong only to Camouflage (try check SPELL_AURA_MOD_CAMOUFLAGE), not for Smoke Bomb.
Anyway I am no sure if SPELL_DAMAGE_CLASS_MELEE is good check, according to commnets some melee spells do not work, some does.

Krudor: I added this because it makes sense for smoke bomb to also be unaffected by melee attacks.
Unless you can provide proof of your claim, I can't make any more changes.

Temppus: Camouflage has it in description:

> You and your pet blend into your surroundings, becoming untargetable by ranged attacks. 

Smoke bomb does not. That is why I am not sure :)

PiskelDEV:
I suggest to add some information about that that SPELL_AURA, GJ

Krudor:
Description added.
I'm also guessing this currently won't work with hunter camouflage, as the hunter's vision will be obscured by his own camouflage, I'll push a fix for this soon.

Krudor:
Should now behave properly, if anyone wants to give it a go.
Tested with different combinations of friend/friend targeting, friend/foe, foe/friend and self-casting on both smoke bomb and camouflage.

Aokromes:
@Shauren @joschiwald this can be merged?

michelR1900:
Can you test this conditions
1- use divine shield and try to attack a rogue in his own smoke bomb from long distance 
2- make 2x rogue and use smoke bomb at same time and try to attack with long distance attack spells

tkrokli:
@michelR1900 : While I agree that the PR author should test any claims made in the PR description, it would be preferable if you test https://github.com/Krudor/TrinityCore/tree/spell_aura_interfere_targetting and post the results you get, then discuss the results here, especially if it does not work as intended.

Goatform:
It doesn't work correctly for me.

1. I am still able to melee attack my target(s) inside Smoke Bomb and they are able to do the same thing.
2. Target outside of Smoke Bomb is still able to cast spells on his target inside Smoke Bomb.

Krudor:
@Goatform You need to a bit more specific than that.
On the first point, the code I presented is supposed to allow you to use melee abilities on targets in/out of smokebomb.
Your second point is not well written and there are multiple scenarios to consider when casting ranged spells into the smoke bomb.
Alliance caster, casting into Horde smoke bomb, on Horde.
Alliance caster, casting into Alliance smoke bomb on Horde.
Alliance caster, casting into Horde smoke bomb, on Alliance.
Alliance caster, casting into Alliance smoke bomb on Alliance.

I would suggest everyone to post video proof about some things not working as intended, just so there are no missunderstands to what's actually happening. I reworked the logic for this several times until I came up with this, rather simple and what seemed to be fully functioning behavior.

Krudor:
Any reason for not merging?

tkrokli:
Good question. Now is a good time for other users to post updated test results from using this content.

P-Kito:
You can use your pvp trinket for smokebomb. This will remove the effect and your healer outside can heal you once again.

Krudor:
That has nothing to do with this PR

P-Kito:
Just saying, since you were discussing about smokebomb :p

PR #17577, opened 2016-07-13
[3.3.5/6.x] Game/Entities: Fix math problem "Disk Point Picking" in GetRandomPoint

P-Kito:
**Changes proposed**:

The calculations made for GetRandomPoint are wrong.
To generate random points over the unit disk, it is incorrect to use two uniformly distributed variables Image and Image and then take 
Image
Image

This wrong calculation leads to more points to center as you can see on the LEFT image.
Image

More information:
http://mathworld.wolfram.com/DiskPointPicking.html
http://stackoverflow.com/questions/5837572/generate-a-random-point-within-a-circle-uniformly

**Target branch(es)**: 335/6x

**Tests performed**: Tested Ingame, Tested Build

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1403,7 +1403,8 @@ void WorldObject::GetRandomPoint(const Position &pos, float distance, float &ran
 
     // angle to face `obj` to `this`
     float angle = (float)rand_norm()*static_cast<float>(2*M_PI);
-    float new_dist = (float)rand_norm()*static_cast<float>(distance);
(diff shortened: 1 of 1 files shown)


DDuarte:
Would it cost anything to do the intermediate calcs with doubles (avoiding all those casts) and only in the end converting to float (**if needed**)?

P-Kito:
Is this in master yet?

tkrokli:
@P-Kito : not that I can see. Check https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Object/Object.cpp#L1831 and notice that it is unchanged compared to this PR.

PR #17590, opened 2016-07-15
[6.x] Core/Spells: Removed a faulty aura handler

Krudor:
**Changes proposed**:
- Remove faulty generic removal of auras triggered by spells

**Target branch(es)**: 6x

**Tests performed**: Build and in-game tested

**Description**:
Currently, auras triggered by SPELL_AURA_PERIODIC_TRIGGER_SPELL are improperly being removed upon the parent aura's removal.
Auras that do obey this behavior will now properly have it handled by Unit::RemoveAurasWithInterruptFlags when called upon, see Unit.h AURA_INTERRUPT_FLAG_...

If anyone knows if this applies to 3.3.5 branch as well, leave a comment and I'll fix the target branches.


Shauren:
what interrupt flag?

Krudor:
None in particular...?
"AURA_INTERRUPT_FLAG_..." = any aura interrupt flags.
Unit::RemoveAurasWithInterruptFlags removes the need for this incredibly generic and absolutely missfunctioning method.

Krudor:
RemoveAurasWithInterruptFlags properly interrupts auras when needed, since the spells have the interrupt flags, and this function just goes in and removes all triggered auras, with no exceptions.

Shauren:
I really don't see how this will work as expected for proc trinkets applying a buff which in turn periodically triggers another buff.

Krudor:
The exception should not govern the general behavior.
You make easily make exceptions from this with scripts / db entries.

The fact is that this function is completely 100% wrong

Shauren:
I gave examples of why its needed, expecting some proof on why this function is wrong now

Krudor:
Okay fair enough, so where would be best to fix these cases, can it be done in db?

Krudor:
These are spells that periodically applies an aura, and then expires, expecting the aura to still be applied.
http://www.wowhead.com/spell=100051/submerge
http://www.wowhead.com/spell=122547/draw-power

Krudor:
I'm pretty sure Blizzard does not design spells like the one you linked anymore.
The ones I've seen with similiar behavior in later expansions now have durations.

I  think it's best to just deal with the spells like the one that you just linked and handle them case by case with a spellscript / db entry (not sure if db can deal with this).

Shauren:
In this case I think you should provide such script for the trinkets that are known to be broken by this

Krudor:
I don't mind.

Krudor:
Shall I update this PR with the spell proc fixes or make a new one, refering to this PR?

Krudor:
Added some aura proc fixes to this PR, let me know if you want separate them.

QAston:
The aura scripts could be replaced by a single generic script, so that it can be easily added to other spells using db, and would save a LOT of code duplication.

Just check in validate() if the spell has a correct aura with existing triggered spell, and in register() register the handler only to that effect. No need to check if the spell you're scripting exists in db as the script loader does that for you, you only need to check existence of spells you're referencing.

Krudor:
I disagree with using a single script to handle multiple spells, just because some of them happen to use the same behavior as other scripts.

Aokromes:

Aokromes:
Conflicting files
src/server/scripts/Spells/spell_item.cpp

Shauren:
Files changed: 575? That can't be right

tkrokli:
Happened in commit d4dd6e2 (the latest one) from what I can see.
Reset the branch back to previous commit and then rebase or merge origin before pushing the update.

Krudor:
Yeah I tried updating from base branch, and got a load of merge errors that I can't fix, not sure what happened.

Krudor:
Not showing up in history, but I've reverted the bad merges, let's see if I can do it right this time

Krudor:
Alright, seems to have solved it.

tkrokli:
Looks a lot better, it should be good to test for other users now. 👍

PR #17594, opened 2016-07-15
Core/Spells GameObjects needs to be spawned in order to be acceptable as a SpellFocus

sirikfoll:
**Changes proposed**:
- If the GameObject is not spawned, then return SPELL_FAILED_REQUIRES_SPELL_FOCUS
- Remove outdated code due to the change above

**Target branch(es)**: 335/6x

**Tests performed**: Built and tested with quest wotlk.openwow.com/quest=9452

### DIFF
--- a/src/server/game/Grids/Notifiers/GridNotifiers.h
+++ b/src/server/game/Grids/Notifiers/GridNotifiers.h
@@ -620,6 +620,9 @@ namespace Trinity
                 if (go->GetGOInfo()->spellFocus.focusId != i_focusId)
                     return false;
 
+                if (!go->isSpawned())
+                    return false;
+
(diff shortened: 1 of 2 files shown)


src/server/game/Spells/Spell.cpp
```diff
@@ -4997,7 +4997,7 @@ SpellCastResult Spell::CheckCast(bool strict)
     if (m_spellInfo->RequiresSpellFocus)
     {
         focusObject = SearchSpellFocus();
-        if (!focusObject)
+        if (!focusObject || !focusObject->isSpawned())
```
joschiwald: what about moving this check to GameObjectFocusCheck?

sirikfoll: Done :)

PR #17601, opened 2016-07-16
Core/Mail: implement a way to specify NPC sender for mails when they should be different from the quest ender

SnapperRy:
**Changes proposed**:

Right now, all mails sent by NPCs when turning in a quest show as being sent by the questender NPC. However, some mails are sent by a different NPC.

Adding ~~a column in quest_template_addon~~ a new table to specify a different NPC's entry to show as sender seems the more sane way to handle the few mails that behave like that.

**Target branch(es)**: 335

**Tests performed**: testing in progress...

### DIFF
--- a/sql/updates/world/3.3.5/9999_99_99_99_world.sql
+++ b/sql/updates/world/3.3.5/9999_99_99_99_world.sql
@@ -1,0 +1,16 @@
+--
+ALTER TABLE `quest_template_addon` ADD COLUMN `RewardMailSenderEntry` INT(5) UNSIGNED DEFAULT 0 NOT NULL AFTER `ProvidedItemCount`;
+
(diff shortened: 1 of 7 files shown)


DDuarte:
There will be very few npcs that need this. Wouldn't it be softer to create a table just for this instead of adding a new field to all the existing rows?

SnapperRy:
I don't know, there are other quest_template_addon fields that only have values for very few entries too.

I guess it depends on whether or not #16650 is accepted. Separate slower tables with less default values against fast and more compact tables with too much useless data.

Going to move it to a separate table either way :P thanks.

Shauren:
@SnapperRy performance impact is much less important in world database where all its content is just selected once at startup

PR #17604, opened 2016-07-16
Core/Auras: Increase number of auras sent to client except for group …

### LINKED ISSUE #17051: [3.3.5][6.x] Core/AuraApplication: Limitation to 64 aura slots.
**Description**:
There is a limit to 64 maximum visible auras for each unit.
At least for 3.3.5, the limit should be at 255.

**Branch(es)**: 335/6x  

**TC hash/commit**:  b82ffb88f7e6342f6abcb09b789ad5c46efe158c

**TDB version**:  TDB 335.61

**Linking model**: static

Gerhood:
**Changes proposed**:

For everything except SMSG_PARTY_MEMBER_STATS_FULL and SMSG_PARTY_MEMBER_STATS we can safely send more than 64 auras. This should cover > 99% of cases where players cant see all auras.

**Target branch(es)**: 335, a similar fix could work on 6.x

**Issues addressed**: Closes #17051

**Tests performed**: Tested and working on higher pop server for several weeks now.

### DIFF
--- a/src/server/game/Handlers/GroupHandler.cpp
+++ b/src/server/game/Handlers/GroupHandler.cpp
@@ -779,7 +779,7 @@ void WorldSession::BuildPartyMemberStatsChangedPacket(Player* player, WorldPacke
     {
         uint64 auramask = player->GetAuraUpdateMaskForRaid();
         *data << uint64(auramask);
-        for (uint32 i = 0; i < MAX_AURAS; ++i)
+        for (uint32 i = 0; i < MAX_AURAS_GROUP_UPDATE; ++i)
         {
             if (auramask & (uint64(1) << i))
             {
@@ -861,7 +861,7 @@ void WorldSession::BuildPartyMemberStatsChangedPacket(Player* player, WorldPacke
         {
             uint64 auramask = pet->GetAuraUpdateMaskForRaid();
             *data << uint64(auramask);
-            for (uint32 i = 0; i < MAX_AURAS; ++i)
+            for (uint32 i = 0; i < MAX_AURAS_GROUP_UPDATE; ++i)
             {
                 if (auramask & (uint64(1) << i))
                 {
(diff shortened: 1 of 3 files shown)


src/server/game/Spells/Auras/SpellAuras.cpp
```diff
@@ -78,7 +78,7 @@ _flags(AFLAG_NONE), _effectsToApply(effMask), _needClientUpdate(false)
             TC_LOG_DEBUG("spells", "Aura: %u Effect: %d put to unit visible auras slot: %u", GetBase()->GetId(), GetEffectMask(), slot);
         }
         else
-            TC_LOG_DEBUG("spells", "Aura: %u Effect: %d could not find empty unit visible slot", GetBase()->GetId(), GetEffectMask());
```
Aokromes: imho we don't need this log change.

Gerhood: Well, if there are more than 255 visible auras on a NPC / player, something's fishy...

src/server/game/Handlers/GroupHandler.cpp
```diff
@@ -779,7 +779,7 @@ void WorldSession::BuildPartyMemberStatsChangedPacket(Player* player, WorldPacke
     {
         uint64 auramask = player->GetAuraUpdateMaskForRaid();
         *data << uint64(auramask);
-        for (uint32 i = 0; i < MAX_AURAS; ++i)
+        for (uint32 i = 0; i < MAX_AURAS_GROUP_UPDATE; ++i)
```
GitVerge: uint8

GitVerge: uint8

Shauren: doesnt matter

RAlexis: May I ask why? Not trying to be a smartass, I honestly don't know why it doesn't matter, thought it would save space. Thanks!

GitVerge: because of scope this variable will be destroyed when we end the loop, that's _probably_ the reason.

jackpoz:
how many auras did you send in your tests ?

Gerhood:
157

PR #17616, opened 2016-07-17
Core/AI Prevent bosses respawn when BossState is set to DONE

sirikfoll:
**Changes proposed**:
- If bossState is set to DONE, then block respawn.
- Currently, the only thing that prevents their respawn, is if they are linked in `linked_respawn`, which is not very effective anymore, since we need to remove that in order to use `_DespawnAtEvade`, for example. Also, just set `spawntimesecs` to a high value isn't enough because now instance cooldowns can be extended.

**Target branch(es)**: 335/6x

**Tests performed**: Built and tested


Shauren:
seems legit

PR #17620, opened 2016-07-18
Game/Entities: Player- Allow non-empty bags to be swapped with larger bag.

Furzmund:
**Changes proposed**:
Allow an equipped bag with items to be swapped with a bag of the same size or larger.

**Target branch(es)**: 335

It builds and runs. When tested the bags swap and the items are put into the new bag. Gives appropriate error message if the bag size is too small or of the wrong type.


Shauren:
This isn't blizzlike, you can only swap EQUIPPED nonempty bags (together with their contents)

Furzmund:
Good catch. I'll look into it some more. Thanks

PR #17630, opened 2016-07-19
Core\UnitAI: correct logic for SpellTargetSelector. CID 1357405

ForesterDev:
**Changes proposed**:
- correct logic for SpellTargetSelector. CID 1357405

**Target branch(es)**: 335/6x

**Issues addressed**: CID 1357405

**NOTE** If your Pull Request is SQL only create a ticket instead

**SUGESTION** If your Pull Request contains SQL give it one imposible date, for example 9999_99_99_99_database.sql on that way it will be free on merge.

### DIFF
--- a/src/server/game/AI/CoreAI/UnitAI.cpp
+++ b/src/server/game/AI/CoreAI/UnitAI.cpp
@@ -267,7 +267,7 @@ SpellTargetSelector::SpellTargetSelector(Unit* caster, uint32 spellId) :
 
 bool SpellTargetSelector::operator()(Unit const* target) const
 {
-    if (target && _spellInfo->CheckTarget(_caster, target) != SPELL_CAST_OK)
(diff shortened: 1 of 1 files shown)


Shauren:
`SpellTargetSelector::operator` must always be called with a target, instead of changing it like this, the whole Spell::CheckRange copypasta should be adjusted to remove all `if (target)` conditions.

src/server/game/AI/CoreAI/UnitAI.cpp
```diff
@@ -267,9 +267,6 @@ SpellTargetSelector::SpellTargetSelector(Unit* caster, uint32 spellId) :
 
 bool SpellTargetSelector::operator()(Unit const* target) const
 {
-    if (!target)
```
Shauren: but this check should stay

PR #17631, opened 2016-07-19
The Kalecgos evade condition.

ghost:
**Changes proposed**:
- Kalecgos evade condition.
## \- Proof http://us.battle.net/wow/en/blog/8896363#raids_dungeons

**Target branch(es)**: 335/6x
6.x only.
Not.
Not testing.

**NOTE** If your Pull Request is SQL only create a ticket instead

**SUGESTION** If your Pull Request contains SQL give it one imposible date, for example 9999_99_99_99_database.sql on that way it will be free on merge.

The Kalecgos encounter in The Sunwell can now be completed by a single player

### DIFF
--- a/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kalecgos.cpp
+++ b/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kalecgos.cpp
@@ -173,7 +173,7 @@ public:
         {
             if (me->HasAura(SPELL_BANISH))
                 return;
-                
+
             bJustReset = true;
             me->SetVisible(false);
             me->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_NON_ATTACKABLE + UNIT_FLAG_NOT_SELECTABLE);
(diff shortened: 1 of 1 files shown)


Ovahlord:
I would put this right under the if(!UpdateVicitim()) check in UpdateAI. this one here will cause the encounter to break things if the evade is called by an whipe while the aura was applied.

Ulduar:
Maybe you right, but Kalecgos can cast spell SPECTRAL_BLAST on player, even if has aura banish.

Krudor:
This will prevent Kalecgos from ever entering evade mode if he's pushed into his spectral form... Why would you want this?

Ovahlord:
it is more that everything after that check like
```
             bJustReset = true;
              me->SetVisible(false);
              me->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_NON_ATTACKABLE + UNIT_FLAG_NOT_SELECTABLE);
```

and so on is getting fucked up if he evades due to an wipe and has that aura.

Ulduar:
Banish aura not removed in script. Event cannot be completed even without evade condition. But after wipe cleanup working.

src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kalecgos.cpp
```diff
@@ -171,6 +171,9 @@ class boss_kalecgos : public CreatureScript
 
         void EnterEvadeMode(EvadeReason why) override
         {
+            if (me->HasAura(SPELL_BANISH))
+                return;
+
```
Aokromes: remove whitespaces :)

Aokromes (changes requested):

Aokromes (approved):

PR #17650, opened 2016-07-22
Game/Entities: Player- Allow an empty bag of equal or larger size to be swapped with an equipped bag containing items.

Furzmund:
**Changes proposed**:
Allow an empty bag of equal or larger size to be swapped with an equipped bag containing items.

**Target branch(es)**: 335

It builds and was tested in game. It now allows an empty bag to be swapped with an equipped bag containing items as long as there is room for all the items in the new bag. Equipped bags with items cannot be moved into the inventory.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -10233,7 +10233,7 @@ InventoryResult Player::CanStoreItem_InSpecificSlot(uint8 bag, uint8 slot, ItemP
 
     uint32 need_space;
 
-    if (pSrcItem && pSrcItem->IsNotEmptyBag() && !IsBagPos(uint16(bag) << 8 | slot))
+	if (pSrcItem && pSrcItem->IsNotEmptyBag() && !IsBagPos(uint16(bag) << 8 | slot) && !swap)
         return EQUIP_ERR_CAN_ONLY_DO_WITH_EMPTY_BAGS;
 
     // empty specific slot - check item fit to slot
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -10233,7 +10233,7 @@ InventoryResult Player::CanStoreItem_InSpecificSlot(uint8 bag, uint8 slot, ItemP
 
     uint32 need_space;
 
-    if (pSrcItem && pSrcItem->IsNotEmptyBag() && !IsBagPos(uint16(bag) << 8 | slot))
+	if (pSrcItem && pSrcItem->IsNotEmptyBag() && !IsBagPos(uint16(bag) << 8 | slot) && !swap)
```
Aokromes: tabs

jackpoz:
Out of curiosity, is this a 335 blizzlike feature ?
Bags handling is the most used vector for cheaters, could you ensure this change doesn't allow any new exploit ?

ariel-:
https://github.com/TrinityCore/TrinityCore/pull/17620

> This isn't blizzlike, you can only swap EQUIPPED nonempty bags (together with their contents)

Shauren:
I don't know about 3.3.5 but in 6.x you can do this kind of swap without first clearing the contents of currently equipped bag
bagswap

In any case, this pull request is wrong and doesn't handle this properly opening cheat possibilities. You are not handling moving the contents to new bag updating all item references so this pull request will remain closed.

Furzmund:
That's not true, at least in 335. You can swap bags that are empty from your inventory to the equipped bag with items. You can also swap them and empty equipped bag with another equipped bag that has items. But if you require it to work for 6xx too, which this doesn't, then I understand.

ariel-:
I stand corrected. You're misundersanding, this is closed not because it's not blizzlike, but because when swapping containers, item backreferences must be updated (ie using a loop)

PR #17657, opened 2016-07-24
[Legion] Update MAX_LEVEL & DEFAULT_MAX_LEVEL

funjoker:
**Changes proposed**:
- Update MAX_LEVEL to 110
- Update DEFAULT_MAX_LEVEL to 110

**Target branch(es)**: 7x Only necessary for Legion

**Tests performed**: 
- Enables the player_levelstats for 101-110


funjoker:
Not Sure if both need to be 123 =/

funjoker:
I think now it fits ... i hope ^^

Shauren:
Well this is interesting, both should be 110 btw (why do we even keep 2 separate defines for this when NONE of them is used for gt code)

funjoker:
yeah tell me why :D
so better 110 ? ^^
Description says we need to check what the max level in gt is so i did :P

PR #17660, opened 2016-07-25
Core/Players: Fix logic regarding HandleCharEnum and DHs

DJScias:
**Changes proposed**:
- If CONFIG_CHARACTER_CREATING_MIN_LEVEL_FOR_DEMON_HUNTER was set to 0 (disable in the config), it would still not allow you to create a DH in case there were no prior characters.
- If RBAC_PERM_SKIP_CHECK_CHARACTER_CREATION_DEMON_HUNTER was true (AKA allow DH creation always) it would not allow DH creation regardless of existing characters or not considering canAlwaysCreateDemonHunter was not checked for validity on charEnum.IsDemonHunterCreationAllowed.

**Target branch(es)**: 6x (technically 7x)

**Issues addressed**: Fixed the logic, making canAlwaysCreateDemonHunter actually taken in account and work -always- as it should.

**Tests performed**: (Does it build, tested in-game, etc) It's been tested with an empty character list both with the minimum level being 0 and the SKIP check. Works in all cases after commit.

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -266,6 +266,8 @@ void WorldSession::HandleCharEnum(PreparedQueryResult result)
 {
     uint8 demonHunterCount = 0; // We use this counter to allow multiple demon hunter creations when allowed in config
     bool canAlwaysCreateDemonHunter = HasPermission(rbac::RBAC_PERM_SKIP_CHECK_CHARACTER_CREATION_DEMON_HUNTER);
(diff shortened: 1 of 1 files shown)


DJScias:
It's kind of illogical that the build is failing since https://github.com/DJScias/TrinityCore/commit/ff812a37f26cb59e5b3489f0defc5ba9b708c386 actually fixed the 'fail' on that branch.

I've no clue why it's happening.

Kind regards,
DJScias

Ovahlord:
change
`charEnum.IsDemonHunterCreationAllowed = (!charEnum.HasDemonHunterOnRealm && charEnum.HasLevel70OnRealm) || canAlwaysCreateDemonHunter;`

to

`charEnum.IsDemonHunterCreationAllowed = (!charEnum.HasDemonHunterOnRealm && charEnum.HasLevel70OnRealm  || canAlwaysCreateDemonHunter);`

PR #17672, opened 2016-07-27
Core/Quest: Finally delete ALL quest bound items on quest reward, fai…

runningnak3d:
**Changes proposed**: Currently only the required amount of quest bound items are deleted on reward, fail, or abandon of a quest. However, there are quests like: http://www.wowhead.com/quest=582/headhunting that allow you to collect more than the required amount due to drop rates.
- Fix Player::RewardQuest to delete ALL quest bound items on quest turn in
- Fix Player::AbandonQuest to delete ALL quest bound items -- period
- Fix Player::FailQuest to delete ALL quest bound items -- period

**Target branch(es)**: 335/6x

- Builds
- Tested quest turn in with and without quest bound items.
- Tested abandoning quests with greater than the required amount of quest bound items
- Tested failing quests with greater than the required amount of quest bound items
- Tested quest turn in with items that are not quest bound to make sure only the required amount were deleted
- Tested abandoning quests with items that are not quest bound to make sure nothing is deleted since they are just normal drops
- Tested failing quests with items that are not quest bound to make sure nothing is deleted since they are just normal drops
- Think I pretty well covered it...

- None

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -14887,9 +14887,9 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver, 
         {
             if (quest->RequiredItemCount[i] > 0 && (itemTemplate->Bonding == BIND_QUEST_ITEM || itemTemplate->Bonding == BIND_QUEST_ITEM1))
                 DestroyItemCount(quest->RequiredItemId[i], 9999, true, true);
+            else 
+                DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true, true);
         }
-        else 
-            DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true, true);
     }
     for (uint8 i = 0; i < QUEST_SOURCE_ITEM_IDS_COUNT; ++i)
     {
@@ -14897,9 +14897,9 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver, 
         {
             if (quest->ItemDropQuantity[i] > 0 && (itemTemplate->Bonding == BIND_QUEST_ITEM || itemTemplate->Bonding == BIND_QUEST_ITEM1))
                 DestroyItemCount(quest->ItemDrop[i], 9999, true, true);
+            else
+                DestroyItemCount(quest->ItemDrop[i], quest->ItemDropQuantity[i], true, true);
         }
-        else
-            DestroyItemCount(quest->ItemDrop[i], quest->ItemDropQuantity[i], true, true);
     }
 
     RemoveTimedQuest(quest_id);
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -14888,8 +14888,8 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver, 
             if (quest->RequiredItemCount[i] > 0 && (itemTemplate->Bonding == BIND_QUEST_ITEM || itemTemplate->Bonding == BIND_QUEST_ITEM1))
                 DestroyItemCount(quest->RequiredItemId[i], 9999, true, true);
         }
-        else if (quest->RequiredItemId[i])
-                DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true, true);
+        else 
+            DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true, true);
     }
     for (uint8 i = 0; i < QUEST_SOURCE_ITEM_IDS_COUNT; ++i)
     {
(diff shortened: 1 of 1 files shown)


Expery:
:+1: Nice

src/server/game/Entities/Player/Player.cpp
```diff
@@ -14882,16 +14882,24 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver,
...
-        if (quest->RequiredItemId[i])
-            DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true);
-
+    {
+        if (ItemTemplate const* itemTemplate = sObjectMgr->GetItemTemplate(quest->RequiredItemId[i]))
+        {
+            if (quest->RequiredItemCount[i] > 0 && (itemTemplate->Bonding == BIND_QUEST_ITEM || itemTemplate->Bonding == BIND_QUEST_ITEM1))
+                DestroyItemCount(quest->RequiredItemId[i], 9999, true, true);
+        }
+        else if (quest->RequiredItemId[i])
```
Shauren: This else doesn't make sense, item is not allowed to exist if GetItemTemplate returns nullptr (so there will be nothing to delete anyway)

runningnak3d: Durrr ... thanks. Will fix :)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -14882,16 +14882,24 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver,
...
-        if (quest->RequiredItemId[i])
-            DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true);
-
+    {
+        if (ItemTemplate const* itemTemplate = sObjectMgr->GetItemTemplate(quest->RequiredItemId[i]))
+        {
+            if (quest->RequiredItemCount[i] > 0 && (itemTemplate->Bonding == BIND_QUEST_ITEM || itemTemplate->Bonding == BIND_QUEST_ITEM1))
+                DestroyItemCount(quest->RequiredItemId[i], 9999, true, true);
+        }
+        else
```
Shauren: what i said earlier still holds... there is no "else" here. if template doesnt exist then DestroyItemCount on next line will do nothing - because the player cannot ever have an item without a template

src/server/game/Entities/Player/Player.cpp
```diff
@@ -14882,16 +14882,24 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver,
     uint32 quest_id = quest->GetQuestId();
 
     for (uint8 i = 0; i < QUEST_ITEM_OBJECTIVES_COUNT; ++i)
-        if (quest->RequiredItemId[i])
-            DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true);
-
+    {
+        if (ItemTemplate const* itemTemplate = sObjectMgr->GetItemTemplate(quest->RequiredItemId[i]))
+        {
+            if (quest->RequiredItemCount[i] > 0 && (itemTemplate->Bonding == BIND_QUEST_ITEM || itemTemplate->Bonding == BIND_QUEST_ITEM1))
```
Shauren: aaaaaaaand you just broke turning in quests requiring items that arent quest bound - you need only to override the count if item bonding is quest, not skip the call

runningnak3d: Eff me. I removed the wrong one locally, fixed it, and forgot to commit. Anyway, I will fix it again...

As for the second parameter, there are two DestroyItemCount methods:
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Player/Player.cpp#L12346
and
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Player/Player.cpp#L12346

The second one only checks inventory (as you can see). I am not sure why both are needed, but that was outside the scope of this PR.

src/server/game/Entities/Player/Player.cpp
runningnak3d: The else is to destroy only the required amount in the case of it being something like runecloth/ If it is quest bound, destroy all, if not, only destroy the quest requirements.

src/server/game/Entities/Player/Player.cpp
Shauren: In this case its the same DestroyItemCount, what you are changing is the value passed for unequip_check (it has a default value false) - and after looking inside DestroyItemCount I can say this change is wrong

runningnak3d: @Shauren This only destroys the items. Player::CanCompeteQuest is what checks the requirements. Just tested, this breaks nothing. The if was redundant though.

src/server/game/Entities/Player/Player.cpp
Shauren: But this else is linked to "if (ItemTemplate exists)" not to bonding check!

runningnak3d: I am fucking blind! Also, I just checked. I know why my second test passed "with this code". No make install after the test compile, so I was still using a core with my old code and the else if which checked for quest->RequiredItemId[i] *sigh*

PR #17675, opened 2016-07-27
Core/Loot: Partially fix group loot. Loot rules should not apply to party loot.

runningnak3d:
**Changes proposed**: Currently if you are in a group, and you all accept a quest, only one member can loot party loot -- no matter the loot mode. An example would be quest 3904 (Milly's Harvest).
The reason I say that this partially fixes the problem is that you have to revert: https://github.com/TrinityCore/TrinityCore/commit/a49544cc187d3a156b4615907d6b82057364fcc4 for this to work correctly. I have not had a chance to look through that commit to find out why, but depending on loot mode, creature->ForceValuesUpdateAtIndex(UNIT_DYNAMIC_FLAGS); is getting sent to players that haven't looted yet. No sparkle, no loot. 

If you don't want to back out that patch, the "workaround" is to have the first person in the party that loots leave their loot window open until everyone has looted. 
- Not apply loot rules to party loot

**Target branch(es)**: 335/6x

- Builds
- Tested with 3 players .. all were able to receive party loot while in a group.

- None for this PR. There will be an additional PR when I fully figure out why dynamic flag updates are getting sent to players that haven't looted.

### DIFF
--- a/src/server/game/Loot/LootMgr.cpp
+++ b/src/server/game/Loot/LootMgr.cpp
@@ -354,7 +354,7 @@ LootItem::LootItem(LootStoreItem const& li)
 
     ItemTemplate const* proto = sObjectMgr->GetItemTemplate(itemid);
     freeforall  = proto && (proto->Flags & ITEM_PROTO_FLAG_MULTI_DROP);
-    follow_loot_rules = proto && (proto->FlagsCu & ITEM_FLAGS_CU_FOLLOW_LOOT_RULES !proto->Flags & ITEM_PROTO_FLAG_MULTI_DROP);
+    follow_loot_rules = proto && (proto->FlagsCu & ITEM_FLAGS_CU_FOLLOW_LOOT_RULES && !proto->Flags & ITEM_PROTO_FLAG_MULTI_DROP);
 
     needs_quest = li.needs_quest;
 
--- a/src/server/game/Loot/LootMgr.cpp
+++ b/src/server/game/Loot/LootMgr.cpp
@@ -354,7 +354,7 @@ LootItem::LootItem(LootStoreItem const& li)
 
     ItemTemplate const* proto = sObjectMgr->GetItemTemplate(itemid);
     freeforall  = proto && (proto->Flags & ITEM_PROTO_FLAG_MULTI_DROP);
-    follow_loot_rules = proto && (proto->FlagsCu & ITEM_FLAGS_CU_FOLLOW_LOOT_RULES);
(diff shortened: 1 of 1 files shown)


Killyana:
I'm not able to reproduce this issue, example quest=7489 I was able to loot the quest item on 2 characters.
Also for the quest  3904, gobs despawn once looted I don't see the problem in this case.

runningnak3d:
@Killyana Were both characters in a group? If so what group loot mode did you have set? I have tested on a clean HEAD with FFA and round robin (still not sure about master looter -- I mean it is broke, but I am not sure how party loot should be handled). 

For me the steps to reproduce this are:
- Create two accounts, and login with both
- Make a group
- .quest add 3904 on both
- go out in the grape field and try to loot.

The first character that loots gets the grapes, and you are right, the GO doesn't immediately despawn, however, when the second member of the group tries to loot the GO -- the loot list is empty. Finally, after the second member of the group attempts to loot, the GO despawns as it should.

How it SHOULD work, is that both members of the party should be able to loot the grapes.

By eliminating party loot from group loot rules, this works as expected.

Killyana:
Yes the 2 characters are in party and the loot rules=party loot
If the quest item is looted from a npc, both characters could get the item, so there's no problem here.
If the quest item is looted from a gob, only one character could get the item (the gob will instantly despawn once looted), the other one must wait for the respawn or get the it from another gob (It always was like this, and it's the correct behavior)
Btw I tested the PR and it changes nothing the loot behavior is the same.

runningnak3d:
I will test again when I get home. Not the first time that I was on crack :)

If you get a chance, would you mind testing with a quest that requires a quest item drop (not just a normal drop that happens to be a quest objective) from an NPC, and use either FFA or RR loot modes?

Killyana:
I tested all loot modes to get the item for quest=7489, and in all cases the 2 players were able to get their own item without any restriction.

runningnak3d:
So, I built a fresh server from HEAD, and I grabbed a fresh client -- no loot problems. Then I connected with my old client to the new server, and there was no loot problems on FFA, then I switched to RR, and loot problems. So, yup, dunno how a client could be messed up and run fine but have that kind of problem. I have been beating my head against the wall on this for over a month now. As for the "fix" in this PR -- you are right -- does nothing. The only reason that I *thought* it did was the fact that the problem was intermittent. Any way -- PR closed.
@Killyana thank you very much for testing. I would have continued to beat my head against a wall.

PR #17676, opened 2016-07-28
Core/Conditions: Add CONDITION_SOURCE_TYPE_QUEST_AVAILABLE to condition manager.

ghost:
**Changes proposed**: Add the ability for quest availability to be determined in condition table if player has completed the required quest. This is my first pull request and is a small partial fix to minimize the complaints.

**Target branch(es)**: 335/6x

**Issues addressed**: Inability of core to properly make a quest available only after two previous quest are completed. 

**Tests performed**: It builds.

- [ ] Need Gossip system to use this condition to determine if quest is made available.

### DIFF
--- a/src/server/game/Conditions/ConditionMgr.cpp
+++ b/src/server/game/Conditions/ConditionMgr.cpp
@@ -58,6 +58,7 @@ char const* const ConditionMgr::StaticSourceTypeData[CONDITION_SOURCE_TYPE_MAX] 
     "Spell Proc",
     "Terrain Swap",
     "Phase"
+    "Quest Available"
 };
 
 ConditionMgr::ConditionTypeInfo const ConditionMgr::StaticConditionTypeData[CONDITION_MAX] =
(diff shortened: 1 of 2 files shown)


src/server/game/Conditions/ConditionMgr.cpp
```diff
@@ -58,6 +58,7 @@ char const* const ConditionMgr::StaticSourceTypeData[CONDITION_SOURCE_TYPE_MAX]
     "Spell Proc",
     "Terrain Swap",
     "Phase"
```
Nyeriah: Missing comma?

DDuarte:
> Inability of core to properly make a quest available only after two previous quest are completed.

See https://trinitycore.atlassian.net/wiki/display/tc/quest_template **Chain of quests with multiple start quests.**. If that doesn't work anymore then it should be fixed instead of adding yet another way to do the same thing.

malcrom:
I complained about this over a year ago while spawning Felwood, nothing was done. Anyway I am looking at CONDITION_SOURCE_TYPE_QUEST_SHOW_MARK to see it it can prevent quest from being added to gossip rather than checking gossip to hide the quest mark.

DDuarte:
Can you paste code showing that the method described in the quest_template wiki does not work?

malcrom:
Can't test with current 6.X core. Tkrokli tried last night.

DDuarte:
It's the same "system" on 3.3.5

PR #17686, opened 2016-07-29
Core/Ticket System: Ticket comment visual output bug

### LINKED ISSUE #16091: [3.3.5] Core/Ticket System: Ticket comment visual output bug
When a GM adds a comment to a ticket which is no assigned, it says that the ticket is assigned to the GM who is commenting even if the ticket is not actually assigned to anyone.

rev: https://github.com/TrinityCore/TrinityCore/commit/cc685ac334335d9f6a25a7c1c739a905f919328f

lucas-wb:
**Changes proposed**:
- Do not send the Gm's name to the ticket handler when commenting on a ticket

**Target branch(es)**: 335 (6.x?)

**Issues addressed**: Closes #16091

Builds, tested commenting on a ticket, assigning tickets, and unassigning tickets

### DIFF
--- a/src/server/scripts/Commands/cs_ticket.cpp
+++ b/src/server/scripts/Commands/cs_ticket.cpp
@@ -199,7 +199,8 @@ public:
         ticket->SaveToDB(trans);
         sTicketMgr->UpdateLastChange();
 
-        std::string msg = handler->PGetParseString(LANG_COMMAND_TICKETLISTADDCOMMENT, player ? player->GetName().c_str() : "Console", comment);
+        std::string msg = ticket->FormatMessageString(*handler, NULL, (ticket ->GetAssignedToName().empty() ? nullptr : ticket->GetAssignedToName().c_str()), NULL, NULL, NULL);
+        msg += handler->PGetParseString(LANG_COMMAND_TICKETLISTADDCOMMENT, player ? player->GetName().c_str() : "Console", comment);
         handler->SendGlobalGMSysMessage(msg.c_str());
 
         return true;
--- a/src/server/scripts/Commands/cs_ticket.cpp
+++ b/src/server/scripts/Commands/cs_ticket.cpp
@@ -199,7 +199,6 @@ public:
         ticket->SaveToDB(trans);
         sTicketMgr->UpdateLastChange();
 
-        std::string msg = ticket->FormatMessageString(*handler, NULL, NULL, NULL, NULL, NULL);
         msg += handler->PGetParseString(LANG_COMMAND_TICKETLISTADDCOMMENT, player ? player->GetName().c_str() : "Console", comment);
         handler->SendGlobalGMSysMessage(msg.c_str());
 
--- a/src/server/scripts/Commands/cs_ticket.cpp
+++ b/src/server/scripts/Commands/cs_ticket.cpp
@@ -199,7 +199,7 @@ public:
         ticket->SaveToDB(trans);
         sTicketMgr->UpdateLastChange();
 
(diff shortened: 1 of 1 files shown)


src/server/scripts/Commands/cs_ticket.cpp
```diff
@@ -199,7 +199,7 @@ class ticket_commandscript : public CommandScript
         ticket->SaveToDB(trans);
         sTicketMgr->UpdateLastChange();
 
-        std::string msg = ticket->FormatMessageString(*handler, NULL, ticket->GetAssignedToName().c_str(), NULL, NULL, NULL);
+        std::string msg = ticket->FormatMessageString(*handler, NULL, NULL, NULL, NULL, NULL);
```
Nyeriah: What's the point of his line after this is removed? It's not passing any data into the function at all, it's all NULLs

Tomatodev: Silly me. Thanks for catching that

src/server/scripts/Commands/cs_ticket.cpp
```diff
@@ -199,7 +199,6 @@ class ticket_commandscript : public CommandScript
         ticket->SaveToDB(trans);
         sTicketMgr->UpdateLastChange();
 
-        std::string msg = ticket->FormatMessageString(*handler, NULL, ticket->GetAssignedToName().c_str(), NULL, NULL, NULL);
         msg += handler->PGetParseString(LANG_COMMAND_TICKETLISTADDCOMMENT, player ? player->GetName().c_str() : "Console", comment);
```
Naios: This will cause an error now...

src/server/scripts/Commands/cs_ticket.cpp
```diff
@@ -199,7 +199,6 @@ class ticket_commandscript : public CommandScript
         ticket->SaveToDB(trans);
         sTicketMgr->UpdateLastChange();
 
-        std::string msg = ticket->FormatMessageString(*handler, NULL, NULL, NULL, NULL, NULL);
         msg += handler->PGetParseString(LANG_COMMAND_TICKETLISTADDCOMMENT, player ? player->GetName().c_str() : "Console", comment);
```
P-Kito:
```c++
std::string msg = handler->PGetParseString(LANG_COMMAND_TICKETLISTADDCOMMENT, player ? player->GetName().c_str() : "Console", comment);
```

Naios:
Could you explain the benefit of this change please?

Tomatodev:
The line currently when you comment on an open ticket, you become assigned to it. removing that line will no longer assign you to the ticket when commenting on it.

Naios:
`FormatMessageString` and `GetAssignedToName` both never change the assignment of the commented ticket.

Also the current codebase looks like not to change the assignment of the ticket through commenting.
Did you test your change?

Tomatodev:
The 3rd argument of `FormatMessageString` is `szAssignedToName` which does this:
`ss << handler.PGetParseString(LANG_COMMAND_TICKETLISTASSIGNEDTO, szAssignedToName);`

The original bug report implies it doesn't actually assign it, but it is only visual. I did test it with this change, and it no longer visually assigns it

Naios:
Ah I see, so it's a visual bug only, I guess it says something like `Assigned to` without any name because the empty string is translated to a c-string and the handler thinks it must parse the string.

Maybe you could replace `ticket->GetAssignedToName().c_str()` by `ticket->GetAssignedToName().empty() ? nullptr : ticket->GetAssignedToName().c_str()` which isn't the best style but stillo better then displaying nothing.

PR #17689, opened 2016-07-30
[6.x/7.x-Prepatch] Scripts/Spells: Fix Shaman talent Earthen Rage

P-Kito:
**Changes proposed**:
- Fix Earthen Rage (http://www.wowhead.com/spell=170374)
- Shoots stones at last damaged target, as spell description says

**Target branch(es)**: 6.x/7.x-Prepatch

**Tests performed**: Working ingame

### DIFF
--- a/src/server/scripts/Spells/spell_shaman.cpp
+++ b/src/server/scripts/Spells/spell_shaman.cpp
@@ -341,15 +341,8 @@ class spell_sha_earthen_rage_proc_aura : public SpellScriptLoader
                 }
             }
 
-            void CalcPeriodic(AuraEffect const* /*aurEff*/, bool& /*isPeriodic*/, int32& amplitude)
-            {
-                // On retail there have been 1-6 procs randomly distributed over the duration
-                amplitude = urand(1000, 4000);
-            }
-
             void Register() override
             {
-                DoEffectCalcPeriodic += AuraEffectCalcPeriodicFn(spell_sha_earthen_rage_proc_aura_AuraScript::CalcPeriodic, EFFECT_0, SPELL_AURA_PERIODIC_DUMMY);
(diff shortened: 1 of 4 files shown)


P-Kito:
I broke it by removing some hacks - I might need a little help here now @Shauren 
Bascially it's not finding procTarget..?

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -266,6 +269,76 @@ class spell_sha_earth_shield : public SpellScriptLoader
...
+                        return false;
+                    if (!sSpellMgr->GetSpellInfo(SPELL_SHAMAN_EARTHEN_RAGE_DAMAGE))
+                        return false;
+                    return true;
+                }
+
+                void HandleEffectProc(AuraEffect const* /*aurEff*/, ProcEventInfo& eventInfo)
+                {
+                    PreventDefaultAction();
+                    _procTarget = eventInfo.GetProcTarget();
```
Ovahlord: simply use  if (Unit* target = eventInfo.GetProcTarget())   and after  if (Unit* caster = eventInfo.GetActor()) so you can drop that Unit* _procTarget;  pointer.

Shauren:
This will never work like that - what you have here are 2 separate instances of spell_sha_earthen_rage_AuraScript, belonging to 2 separate auras. Putting all this code together inside one class only adds confusion.

P-Kito:
Talent working ingame, tested build.

joschiwald:
what's the advantage by using an new container in aura class instead of saving data in scripts (we can access them with Aura::GetScriptByName)?

P-Kito:
ok so obviously Spellwork didnt show the amplitude, I can remove that hack then :)

Aokromes:
@Shauren @joschiwald this can be merged?

Shauren:
3fedeb1c9d6b2b218f46774991bd8e841477ad1a

PR #17691, opened 2016-07-30
Core/Scripts: Remove useless SD2 error log

ghost:
**Changes proposed:**

Another "removal" of hardcoded text in the outdated #define format.
This time it is not so much about removing the text as using the
preferred format and placing of the error log string in TC_LOG_ERROR.
- remove "#define ERROR_INST_DATA" (old format)
- move output string to the TC_LOG_ERROR command

**Issues addressed:** Updates https://github.com/velinath/TrinityCore/issues/2

Builds without issues, this is just deleting 1 label & moving a text string.

None. This is only about moving a text string, no changes to the script code functionality.

### DIFF
--- a/src/server/scripts/Outland/BlackTemple/illidari_council.cpp
+++ b/src/server/scripts/Outland/BlackTemple/illidari_council.cpp
@@ -163,7 +163,7 @@ public:
                 Council[1] = instance->GetGuidData(DATA_VERAS_DARKSHADOW);
                 Council[2] = instance->GetGuidData(DATA_LADY_MALANDE);
                 Council[3] = instance->GetGuidData(DATA_HIGH_NETHERMANCER_ZEREVOR);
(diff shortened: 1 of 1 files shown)


joschiwald:
useless log message

tkrokli:
Sure, but do you have a suggestion for what to do instead? All I did was to remove the `#define` part and move the actual text string back to the actual log message. I am not introducing it, I am merely getting rid of the `#define` part.

I am open to suggestions on how to deal with that log message, but I don't have any ideas myself. Please, feel free to suggest any changes that would improve on what you see as useless about the logging part.

sirikfoll:
Remove the define, the "else" and the log ^^

tkrokli:
Thanks, that sounds like a good solution to me. **joschiwald**, do you concur with sirikfoll's suggestion?

tkrokli:
If this commit is acceptable, I will rename the PR title to "Remove useless SD2 error log".

Issue #17539, opened 2016-07-07
[3.3.5] Melee Range spells have less range than autoattacks


Raydor:
**Description**: After @Shauren changes to range of melee spells, melee spells can't be casted at the same distance at maximum range of autoattack

**Current behaviour**: Can't cast melee range spells at maximum autoattack range

**Expected behaviour**: You should be able to cast melee range spells at the same distance as you can autoattack

**Steps to reproduce the problem**:

1. Have 2 people and duel between them
2. Walk pixel by pixel until you can autoattack the enemy, then stop
3. You can autoattack the enemy but you can't cast spells like Mortal Strike (melee range spells)

**Branch(es)**: Tested in 3.3.5, don't know if this also happens in 6.x

**TC hash/commit**:  https://github.com/TrinityCore/TrinityCore/commit/d0c377f86fec3d8ae3740fed303edfd6ec49a7ee

Proof (you can see how I can autoattack the enemy but when I try to cast Mortal Strike it returns Out of Range):
![](http://puu.sh/pTqVt/32aa97fd3d.jpg)

SnapperRy:
Wonder if it's the autoattack range that's too high.

Raydor:
Yeah probably the bug is on autoattack range side, I also thought that

Shauren:
autoattack needs -1, see https://github.com/TrinityCore/TrinityCore/commit/e6a52d4aae9e8facc46c3b05e0560d540a928d73#commitcomment-18104812

Raydor:
Tried just removing "dist" from the float
```c++
float maxdist = sizefactor;
```
for a fast testing and then is just the opposite, I can cast melee spells farther than I can autoattack. Probably you know that but just in case.

Shauren:
Ok, then it will need a bit more work than I expected

killradio:
@Raydor can you plz check after recent changes? Now I can cast even if staying a bit far for autoattack. Checked distance for autoattack with off - with TC you should stay closer.

Shauren:
@killradio see https://github.com/TrinityCore/TrinityCore/commit/14b93c04ee07e9fe6e2767460e7fc7cc2b3e6808#commitcomment-18191617 - this part is missing to set min range 5

chaodhib:
@killradio could you re-test with https://github.com/TrinityCore/TrinityCore/issues/17693 plz?

Raydor:
[...]

PR #17693, opened 2016-07-30
Yet another attempt to fix the melee range checks + clean up

chaodhib:
**Changes proposed**:
- Duplication of logic is bad. So I tried to regroup the "melee range" logic into one function and let the other parts of the game reuse that function.
- I simplyfied Spell::CheckRange() and broke it into 2 functions.
- Fix what is mentioned [here
  ](https://github.com/TrinityCore/TrinityCore/commit/14b93c04ee07e9fe6e2767460e7fc7cc2b3e6808#commitcomment-18191617) and [here](https://github.com/TrinityCore/TrinityCore/commit/e6a52d4aae9e8facc46c3b05e0560d540a928d73#commitcomment-18104786)
- Fix a potential issue regarding non melee spell ranges on non unit targets:
  In order to write [this commit](https://github.com/TrinityCore/TrinityCore/commit/e6a52d4aae9e8facc46c3b05e0560d540a928d73), @Shauren used [this decompiled code ](http://paste2.org/2FUCZ4Ix)of wow client 6.2.3 (cf https://community.trinitycore.org/topic/12515-unambiguously-defining-auto-attack-and-spell-range). In it, we can see how the client manages the range validation. And long story short, Shauren didn't translate completely the line 86 from the decompiled code. This is how I wrote this fix.

**Target branch(es)**: 335

**Issues addressed**: https://github.com/TrinityCore/TrinityCore/issues/17539 (which is strange because the ticket is closed but the issue is still unsolved)

**Tests performed**: It builds. Did not test IG to be honest.

**Left to do**: Clean up SpellTargetSelector::operator() in UnitAI.cpp which also duplicate logic :(

This PR is easier to read commit by commit I think.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -492,12 +492,7 @@ bool Unit::IsWithinMeleeRange(Unit const* obj) const
 
 float Unit::GetMeleeRange(Unit const* target) const
 {
-    float range = this->GetCombatReach() + 4.0f / 3.0f;
-    if (target)
-        range += target->GetCombatReach();
-    else
-        range += this->GetCombatReach();
-
+    float range = this->GetCombatReach() + target->GetCombatReach() + 4.0f / 3.0f;
     return std::max(range, NOMINAL_MELEE_RANGE);
 }
 
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -485,9 +485,20 @@ bool Unit::IsWithinMeleeRange(Unit const* obj) const
     float dz = GetPositionZMinusOffset() - obj->GetPositionZMinusOffset();
     float distsq = dx*dx + dy*dy + dz*dz;
 
-    float maxdist = GetCombatReach() + obj->GetCombatReach() + 4.0f / 3.0f;
+    float maxdist = GetMeleeRange(obj);
 
     return distsq <= maxdist * maxdist;
+}
+
+float Unit::GetMeleeRange(Unit const* target) const
+{
+    float range = this->GetCombatReach() + 4.0f / 3.0f;
+    if (target)
+        range += target->GetCombatReach();
+    else
+        range += this->GetCombatReach();
+
+    return std::max(range, NOMINAL_MELEE_RANGE);
 }
 
 void Unit::GetRandomContactPoint(const Unit* obj, float &x, float &y, float &z, float distance2dMin, float distance2dMax) const
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -492,7 +492,7 @@ bool Unit::IsWithinMeleeRange(Unit const* obj) const
 
 float Unit::GetMeleeRange(Unit const* target) const
 {
-    float range = this->GetCombatReach() + target->GetCombatReach() + 4.0f / 3.0f;
+    float range = GetCombatReach() + target->GetCombatReach() + 4.0f / 3.0f;
     return std::max(range, NOMINAL_MELEE_RANGE);
 }
 
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -5835,7 +5835,7 @@ SpellCastResult Spell::CheckRange(bool strict)
     float minRange = 0.0f;
     float maxRange = 0.0f;
     
-    GetMinMaxRange(strict, &minRange, &maxRange);
+    GetMinMaxRange(strict, minRange, maxRange);
 
     // get square values for sqr distance checks
     minRange *= minRange;
@@ -5866,13 +5866,13 @@ SpellCastResult Spell::CheckRange(bool strict)
     return SPELL_CAST_OK;
 }
 
-void Spell::GetMinMaxRange(bool strict, float* minRange, float* maxRange)
+void Spell::GetMinMaxRange(bool strict, float& minRange, float& maxRange)
 {
     Unit* target = m_targets.GetUnitTarget();
     float rangeMod = 0.0f;
     if (strict && IsNextMeleeSwingSpell()) 
     {
-        *maxRange = 100.0f;
+        maxRange = 100.0f;
         return;
     }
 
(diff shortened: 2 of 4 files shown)


chaodhib:
I've tested it IG and it works. Also fixed non melee spell range on non unit targets (using decompiled code that @Shauren used to make the original commit https://github.com/TrinityCore/TrinityCore/commit/e6a52d4aae9e8facc46c3b05e0560d540a928d73.

killradio:
@chaodhib Hey. Now it looks like ok, but can't compare with off-server.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -485,11 +485,22 @@ bool Unit::IsWithinMeleeRange(Unit const* obj) const
...
 
     return distsq <= maxdist * maxdist;
 }
 
+float Unit::GetMeleeRange(Unit const* target) const
+{
+    float range = this->GetCombatReach() + 4.0f / 3.0f;
+    if (target)
+        range += target->GetCombatReach();
+    else
```
killradio:
```cpp
bool Unit::IsWithinMeleeRange(Unit const* obj) const
{
    if (!obj || !IsInMap(obj) || !InSamePhase(obj))
        return false;
...
    float maxdist = GetMeleeRange(obj);
,,,
```
?

chaodhib: I'm not sure I understand what you mean

chaodhib: The cases where `Unit::GetMeleeRange(NULL)` is called does not come from `Unit::IsWithinMeleeRange `but from `Spell::GetMinMaxRange `which also make use of the function. But yeah, that part is a bit strange to explain, so I suppose it should be refactored.

chaodhib: done

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -485,11 +485,17 @@ bool Unit::IsWithinMeleeRange(Unit const* obj) const
...
 
-    float maxdist = GetCombatReach() + obj->GetCombatReach() + 4.0f / 3.0f;
+    float maxdist = GetMeleeRange(obj);
 
     return distsq <= maxdist * maxdist;
 }
 
+float Unit::GetMeleeRange(Unit const* target) const
+{
+    float range = this->GetCombatReach() + target->GetCombatReach() + 4.0f / 3.0f;
```
ccrs: fear the silly uses in scripts, I wouldnt remove the nullptr check xd

Shauren:
From what I see this is just refactoring, not fixes here

killradio:
@Shauren yep, with mentioned in https://github.com/TrinityCore/TrinityCore/commit/14b93c04ee07e9fe6e2767460e7fc7cc2b3e6808 by @joschiwald missing changes.

chaodhib:
@Shauren 
- one fix regarding https://github.com/TrinityCore/TrinityCore/commit/14b93c04ee07e9fe6e2767460e7fc7cc2b3e6808#commitcomment-18191617 (by avoid duplication of logic)

- one small fix https://github.com/TrinityCore/TrinityCore/pull/17693/files?diff=unified#diff-952c711803610aa78110e4f92ff2516aR5892 (based on the same decompiled wow client code that you used)
Also yeah, a bit of refactoring here and there.

src/server/game/Entities/Unit/Unit.cpp
killradio: Imo target should be checked before calling GetMeleeRange, cuz using caster as target if !target -- may be not what it should be in some cases.

chaodhib: yeah I agree with @killradio. Even if I check if target is null, I don't see how I could handle that case anyway. This function shouldn't be called with NULL. And it shouldn't be called in AI scripts anyway. Only by the core.

jackpoz: please don't use "this->" syntax

chaodhib: why not?

jackpoz: Because it's not needed and causes confusion when mixed with code that doesn't use it (as in the rest of TC sources).

Aokromes:

chaodhib:
done

src/server/game/Spells/Spell.cpp
```diff
@@ -5831,46 +5831,69 @@ SpellCastResult Spell::CheckRange(bool strict)
...
+        if (m_caster->GetExactDistSq(m_targets.GetDstPos()) > maxRange)
+            return !(_triggeredCastFlags & TRIGGERED_DONT_REPORT_CAST_ERROR) ? SPELL_FAILED_OUT_OF_RANGE : SPELL_FAILED_DONT_REPORT;
+        if (minRange > 0.0f && m_caster->GetExactDistSq(m_targets.GetDstPos()) < minRange)
+            return !(_triggeredCastFlags & TRIGGERED_DONT_REPORT_CAST_ERROR) ? SPELL_FAILED_OUT_OF_RANGE : SPELL_FAILED_DONT_REPORT;
+    }
+
+    return SPELL_CAST_OK;
+}
+
+void Spell::GetMinMaxRange(bool strict, float& minRange, float& maxRange)
```
donato: Feel free to disagree with me, as I'm new to TrinityCore.

I think this would work better as 
```c++
std::tuple<float, float> Spell::GetMinMaxRange(bool strict) 
```
Generally I'm not a fan of using "return parameters"

src/server/game/Spells/Spell.cpp
```diff
@@ -5831,46 +5831,69 @@ SpellCastResult Spell::CheckRange(bool strict)
     if (!strict && m_casttime == 0)
```
donato: why not include 
```c++ 
&& target != m_caster
```
here, and just exit early?

src/server/game/Spells/Spell.cpp
donato: PS, what this would do to the invoking code is
```c++
     float minRange, maxRange;
     std::tie(minRange, maxRange) = GetMinMaxRange(strict);
```

chaodhib: Indeed. It's probably better. I will change that.

src/server/game/Spells/Spell.cpp
chaodhib: I'm not quite sure I follow

donato: I edited the code example included in the previous comment. I was trying to say, that if the target is the caster, then (my understanding is) it will always be within range. Thus

```c++
if (target == m_caster)
    return SPELL_CAST_OK;
```

Removing the need for the check on line 5844

killradio: @donato you forgot about line 5857 and case when target is null.

Shauren: @killradio but m_caster can never be NULL

killradio: @Shauren its true, but target can be NULL. When target == m_caster and !target -- two different cases? Peace of code where ```if (m_targets.HasDst() && !m_targets.HasTraj())``` is not needed if target == m_caster?

chaodhib: I rather not extend the scope of this PR. Feel free to open a new PR if you would like. Then things can be discussed and tested more thoroughly there.

Shauren: leave it alone then

src/server/game/Spells/Spell.cpp
chaodhib: Done. Thank you for your help.

chaodhib:
I updated the description.

sirikfoll:
@chaodhib Don't you know how to fix #16549? Just asking because seems to be related to the code you are working on

src/server/game/Spells/Spell.cpp
```diff
@@ -5831,46 +5831,70 @@ SpellCastResult Spell::CheckRange(bool strict)
...
+
+std::pair<float, float> Spell::GetMinMaxRange(bool strict)
+{
+    float rangeMod = 0.0f;
     float minRange = 0.0f;
     float maxRange = 0.0f;
-    float rangeMod = 0.0f;
-    if (strict && IsNextMeleeSwingSpell())
+
+    if (strict && IsNextMeleeSwingSpell())
```
sirikfoll: Whitespace here

tkrokli: I think it improves readability, but it would be nice to have consistency in the spacing.
I would remove the blank line before `return SPELL_CAST_OK;` instead (Line 5862).
Maybe make it a rule to have a blank line before each `if ()` check or no extra LF at all?

chaodhib:
@sirikfoll I'll have a look into it

PR #17704, opened 2016-07-31
[6.x] Core/Spell: Restore SPELL_AURA_MOD_DAMAGE_PERCENT_TAKEN behavior

Regigicas:
**Changes proposed:**

Restore SPELL_AURA_MOD_DAMAGE_PERCENT_TAKEN behavior on spells cases. Currently is working on melee damage

**Target branch(es):** 6.x

**Issues addressed:** None i think

**Tests performed:** (Does it build, tested in-game, etc):
.unaura all
.aura 60988
/cast Dispersion

Tested on a shadow priest lvl 110
Currenty the damage taken is around 610 per tick, if you activate Dispersion you still take 610 per tick.
With this fix 610 is reduced to 244 which is correct according to the 60% damage reduction.

This commit only restore these lines https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L9977-L9979 as they are missing in 6.x branch

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8033,6 +8033,10 @@ uint32 Unit::SpellDamageBonusTaken(Unit* caster, SpellInfo const* spellProto, ui
                 TakenTotalCasterMod += (float((*i)->GetAmount()));
         }
 
+        // from positive and negative SPELL_AURA_MOD_DAMAGE_PERCENT_TAKEN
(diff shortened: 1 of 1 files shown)


Shauren:
Got a link to the commit that removed it?

Regigicas:
https://github.com/TrinityCore/TrinityCore/commit/83c2e79d52a57ec670b439ed6f822dd18e0ac55e#diff-4522eeffcbc961602e86620011a2d579L8693

Issue #11418, opened 2014-01-08
[Script/SQL/Fix] Fumping and The Big Bone Worm


Raszagal:
TrinityCore rev. b1431bdd47a2+ 2013-12-29 23:31:07 -0330 (master branch) (Win32, Release) (worldserver-daemon)
(The + is because of this script.)
Using World DB: TDB 335.52

I take NO credit for this! Only credit I take is testing to see if it works.

ALL Credit goes to @Discover- it is his script!

SQL...
```sql
-- Addition to spell_script_names for Fumping and The Big Bone Worm
DELETE FROM spell_script_names WHERE spell_id IN (39238,39246);
INSERT INTO spell_script_names VALUES (39238,'spell_fumping_bone_sifters'),(39246,'spell_fumping_gigantic_bone_worm ');
```

Script... Add to: scripts > Source Files > spell_quest.cpp
```cpp
enum Fumping
{
    NPC_MATURE_BONE_SIFTER  = 22482,
    NPC_SAND_GNOME          = 22483,
    NPC_HAISHULUD           = 22038,
};
 
// Quests 'Fumping' and 'The Big Bone Worm'
class spell_fumping_bone_sifters : public SpellScriptLoader
{
    public:
        spell_fumping_bone_sifters() : SpellScriptLoader("spell_fumping_bone_sifters") { }
 
        class spell_fumping_bone_sifters_SpellScript : public SpellScript
        {
            PrepareSpellScript(spell_fumping_bone_sifters_SpellScript)
 
            void HandleDummy(SpellEffIndex /*effIndex*/)
            {
                if (Player* player = GetCaster()->ToPlayer())
                    urand(0, 1) == 0 ? player->SummonCreature(NPC_MATURE_BONE_SIFTER, player->GetPositionX()+rand()%10, player->GetPositionY()+rand()%10, player->GetPositionZ()+5, 0.0f, TEMPSUMMON_DEAD_DESPAWN, 0) : player->SummonCreature(NPC_SAND_GNOME, player->GetPositionX()+rand()%10, player->GetPositionY()+rand()%10, player->GetPositionZ()+5, 0.0f, TEMPSUMMON_DEAD_DESPAWN, 0);
            }
 
            void Register()
            {
                OnEffectHit += SpellEffectFn(spell_fumping_bone_sifters_SpellScript::HandleDummy, EFFECT_ALL, SPELL_EFFECT_ANY);
            }
        };
 
[...]

PR #17706, opened 2016-08-01
Core/Conditions Exclude creatures not spawned from CONDITION_NEAR_CREATURE search

sirikfoll:
**Changes proposed:**
-  Exclude creatures not spawned from CONDITION_NEAR_CREATURE

**Issues addressed:** Updates #11418

**Tests performed:** Built and tested with quest The Big Bone Worm(10930)

### DIFF
--- a/src/server/game/Conditions/ConditionMgr.cpp
+++ b/src/server/game/Conditions/ConditionMgr.cpp
@@ -284,11 +284,7 @@ bool Condition::Meets(ConditionSourceInfo& sourceInfo) const
         }
         case CONDITION_NEAR_CREATURE:
         {
-            Creature* crea = GetClosestCreatureWithEntry(object, ConditionValue1, (float)ConditionValue2, bool(!ConditionValue3));
-            if (!crea || (ConditionValue3 && crea->getDeathState() == DEAD))
-                condMeets = false;
-            else
-                condMeets = true;
+            condMeets = GetClosestSpawnedCreatureWithEntry(object, ConditionValue1, (float)ConditionValue2, bool(!ConditionValue3)) ? true : false;
             break;
         }
         case CONDITION_NEAR_GAMEOBJECT:
--- a/src/server/game/Conditions/ConditionMgr.cpp
+++ b/src/server/game/Conditions/ConditionMgr.cpp
@@ -284,7 +284,11 @@ bool Condition::Meets(ConditionSourceInfo& sourceInfo) const
         }
         case CONDITION_NEAR_CREATURE:
         {
-            condMeets = GetClosestCreatureWithEntry(object, ConditionValue1, (float)ConditionValue2, bool(!ConditionValue3)) ? true : false;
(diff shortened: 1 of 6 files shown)


tkrokli:
:+1: Nice, I like the idea. Not competent to comment on the code itself, though.

src/server/game/Conditions/ConditionMgr.cpp
```diff
@@ -284,7 +284,11 @@ bool Condition::Meets(ConditionSourceInfo& sourceInfo) const
         }
         case CONDITION_NEAR_CREATURE:
         {
-            condMeets = GetClosestCreatureWithEntry(object, ConditionValue1, (float)ConditionValue2, bool(!ConditionValue3)) ? true : false;
+            Creature* crea = GetClosestCreatureWithEntry(object, ConditionValue1, (float)ConditionValue2, bool(!ConditionValue3));
+            if (!crea || (ConditionValue3 && crea->getDeathState() == DEAD))
```
jackpoz: doesn't GetClosestCreatureWithEntry() already check for death state ?

Shauren: 3rd argument controls this - but here you must remember what deathstate values mean
CORPSE = is dead but still visible/spawned
DEAD = despawned, waiting for respawn

If i'm reading this right, this check is for the case when GetClosestCreatureWithEntry is allowed to return dead creatures via ConditionValue3

jackpoz: it would make more sense to change GetClosestCreatureWithEntry() to ignore despawned creatures then (or add a new function that does that)

Treeston: @Shauren any particular reason why we named them that to begin with?

Treeston: Also this isn't going to work if there's, say, a despawned corpse at 15yd and a living creature at 20yd.

sirikfoll: @Treeston I'm not sure i'm following, if ConditionValue3 is true, GetClosestCreatureWithEntry() will already have excluded all creatures with ALIVE deathState

@jackpoz I don't think change GetClosestCreatureWithEntry() to ignore despawned creatures is wise, it could be used like, for example, to find and respawn a dead creature. Also, I don't know how to create a grid searcher to do this :/

Treeston: @sirikfoll This comment chain is from before you changed it, when you simply took GetClosestCreatureWithEntry and then checked death state for that.

jackpoz: > Also, I don't know how to create a grid searcher to do this :/

there's no hurry

sirikfoll:
@jackpoz  This is the way?

jackpoz:
looks good at first sight

Treeston:
@shauren might know for sure

Shauren:
After inspecting nearly all FindNearestSpawnedCreature uses, it was enough to just throw this check into  NearestCreatureEntryWithLiveStateInObjectRangeCheck

PR #17732, opened 2016-08-04
Core/Command: Check account password max length

ForesterDev:
**Changes proposed:**
-  account password is limited by client to 16 characters, like the account name. We should check this, when create new account with command .account create

**Target branch(es):** 3.3.5

**Tests performed:** (Does it build, tested in-game, etc) tested


tkrokli:
The filename `2016_08_04_01_world.sql` already exists on TrinityCore. Change `_01_` to a higher number than existing filenames in https://github.com/TrinityCore/TrinityCore/tree/3.3.5/sql/updates/world/3.3.5 so the conflict can be resolved. :-)

sql/updates/world/3.3.5/2016_08_04_05_world.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM `trinity_string` WHERE `entry` = 1031;
+INSERT INTO `trinity_string` (`entry`, `content_default`) VALUES
+('1031','An account password can NOT be longer than 16 characters (client limit). The account was NOT created.');
```
Aokromes: no ' around numbers.

PR #17737, opened 2016-08-05
Scripts/SmartScripts: Fix charmed behavior for SmartAI

Treeston:
# hype

Fix SmartAI to no longer issue any movement/combat orders while the creature is controlled by a player.


Shauren:
Please open a new PR with no renames, no code reformatting, no replacing NULL with nullptr and no removing commented out code so we can actually see what you changed.

PR #17738, opened 2016-08-05
Scripts/SmartScripts: Fix charmed behavior for SmartAI

Treeston:
# hype

Fix SmartAI to no longer control movement or cast spell while the creature is controlled by a player.
Also add new `SMART_EVENT_FLAG_WHILE_CHARMED` event flag that allows overriding this behavior.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -28,6 +28,7 @@
 
 SmartAI::SmartAI(Creature* c) : CreatureAI(c)
 {
+    mIsCharmed = false;
     // copy script to local (protection for table reload)
 
     mWayPoints = nullptr;
@@ -68,6 +69,15 @@ SmartAI::SmartAI(Creature* c) : CreatureAI(c)
     mInvincibilityHpLevel = 0;
 
     mJustReset = false;
+}
+
+bool SmartAI::IsAIControlled() const
+{
+    if (me->IsControlledByPlayer())
+        return false;
+    if (mIsCharmed)
+        return false;
+    return true;
 }
 
 void SmartAI::UpdateDespawn(const uint32 diff)
@@ -155,7 +165,7 @@ void SmartAI::PausePath(uint32 delay, bool forced)
         return;
     if (HasEscortState(SMART_ESCORT_PAUSED))
     {
-        TC_LOG_ERROR("misc", "SmartAI::StartPath: Creature entry %u wanted to pause waypoint movement while already paused, ignoring.", me->GetEntry());
+        TC_LOG_ERROR("misc", "SmartAI::PausePath: Creature entry %u wanted to pause waypoint movement while already paused, ignoring.", me->GetEntry());
         return;
     }
     mForcedPaused = forced;
@@ -199,7 +209,10 @@ void SmartAI::EndPath(bool fail)
     mLastWP = nullptr;
 
     if (mCanRepeatPath)
-        StartPath(mRun, GetScript()->GetPathId(), true);
+    {
+        if (IsAIControlled())
+            StartPath(mRun, GetScript()->GetPathId(), true);
+    }
     else
         GetScript()->SetPathId(0);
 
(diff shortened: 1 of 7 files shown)


src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -545,8 +576,12 @@ void SmartAI::JustReachedHome()
 
 void SmartAI::EnterCombat(Unit* enemy)
 {
-    me->InterruptNonMeleeSpells(false); // must be before ProcessEvents
     GetScript()->ProcessEventsFor(SMART_EVENT_AGGRO, enemy);
+
+    if (!IsAIControlled())
+        return;
+
+    me->InterruptNonMeleeSpells(false); // must be before ProcessEvents
```
Shauren: well this is obviously not good - see the comment

ccrs:
"TODO: Some sort of system for smart actions to indicate whether they want to be executed during charm or not."

We talked about that months ago :smile:
Add a event flag similar to the ones that already exist to set which ones should trigger while charmed or not. Thats the easy part... the fun part comes when converting the tons of scripts that now are triggered when charmed (and usually they just check auras :D).

ccrs:
Eitherway I'm sure this already breaks a considerable amount of scripts xd

Treeston:
@ccrs the difference is that this time, it's a matter of "everything still _works_ while charmed, now we need to selectively figure out what _shouldn't_ work". Before it was "nothing works while charmed".

ccrs:
Casts are not the only thing that is made while charmed (texts, explored credit, ...)
IMO the param should be on the event (like the param that currrently exists to not repeat the event) instead of the cast.

Treeston:
Yes, but texts etc should all still fire while charmed, no? The only thing we want to prevent are things that "fight" for control with the controller (casts and movement). Movement is prevented (allowing it is pointless, since the controller can always override it), and casting has the flag.

ccrs:
but then again, right now the only way to mark events for "charm only" is with link on aura for example or using conditions. How simple would it be just marking a flag? xd

Treeston:
That's kind of the intended use case for conditions though.

ccrs:
then why using the smart cast flag of combat move? its the same logic

Plus some charms cannot be "detected" with conditions.

Treeston:
What kind of charms? `IsCharmed` should still be true?

PS: If there's no condition for charmed, then we need one of those instead.

Treeston:
Implementing "_only_ while charmed" using SAI flags is kind of silly anyway. If anything, we'd have a flag for "can be done while charmed", otherwise it'd be "only while not charmed".

Treeston:
Re conditions: 9785546

ccrs:
I dont see how making an event flag is "dumber" than have to create one separate condition for every single event that has to trigger only on charm :P

Treeston:
It's gonna cost us some storage space because event flags now exceed `uint8`, but there you go @ccrs

ccrs:
its the only way to not rewrite every script with charm mechanic :)

Treeston:
It's still gonna break some scripts, but those can be fixed manually. Can't make this omelet without breaking some eggs, I'm afraid.

Killyana:
This also fix the issue with vehicles using sai?

Treeston:
Yes.

Treeston:
Merged in 96f9038 e9eacae

lost-illusion:
@Treeston Hi, can i ask you about this commit. For example we have spell 32802, and this spell summon npc 18734. In SummonPropertiesEntry type seted to SUMMON_TYPE_GUARDIAN, and after summon npc "controlled by player" and don't cast OOC Healing player (In update you not set flag for spell casts) . Can you explain me please why ? I mean it guard, player can't directly controll her. Or we should manualy set flag if needed ? I'm sorry if i understand this wrong.

Treeston:
Flag should be manually set for any guardians that should use it, @lost-illusion. Open an issue about your specific case, it's a very quick fix.

Treeston:
0f73fc8

Treeston:
See 0f7efc8

lost-illusion:
@Treeston  Thanks. Maybe we should found all guardians and check their ? I mean non controlled by player. Minions, guardians.

Treeston:
If you find any guardians missing the flag, open an issue for it.

lost-illusion:
Ok.

PR #17748, opened 2016-08-06
Core/Scripts: move #define to enum

ghost:
**Changes proposed:**

Move #define variable and value to enum in script generic_creature,
core script file mob_generic_creature.cpp

(Part of the ongoing series of removing hardcoded #define variables in core scripts.)

**Target branch(es):** 3.3.5 (not sure if this file is in use in 6.x/7.x)

**Issues addressed:** Updates https://github.com/velinath/TrinityCore/issues/2

Builds without issues, only a text change by moving a #define to enum.

None that I know of.

### DIFF
--- a/src/server/scripts/World/mob_generic_creature.cpp
+++ b/src/server/scripts/World/mob_generic_creature.cpp
@@ -27,7 +27,10 @@ EndScriptData */
 #include "ScriptedCreature.h"
 #include "PassiveAI.h"
 
-#define GENERIC_CREATURE_COOLDOWN   5000
+enum GenericCreature
+{
+    GENERIC_CREATURE_COOLDOWN = 5000
+};
 
 class generic_creature : public CreatureScript
 {


offl:
Guess all three scripts from this file are obsolete because SAI or c_t_a can replace their functions.

tkrokli:
Quite possible. I will wait for a TC team/collaborator developer to confirm if the file should be removed.

src/server/scripts/World/mob_generic_creature.cpp
```diff
@@ -27,7 +27,10 @@ EndScriptData */
 #include "ScriptedCreature.h"
 #include "PassiveAI.h"
 
-#define GENERIC_CREATURE_COOLDOWN   5000
+enum GenericCreature
+{
+    GENERIC_CREATURE_COOLDOWN = 5000
```
Expecto: 5*IN_MILLISECONDS maybe?

Naios: Use std::chrono pls (`Seconds(5)`)

tkrokli: OK, so I will drop both the enum and #define completely and just insert `std::chrono pls (`Seconds(5)`)` at each occurrence in the code instead. Very well. Thanks for the tip. :+1:

Naios: Better define a static const variable for that

Shauren: or just dont bother and leave it in enum like it was, less chance to screw up something

Shauren:
pfft, this script isnt even used, closing for that reason

src/server/scripts/World/mob_generic_creature.cpp
tkrokli: and I would also need to
```cpp
#include <chrono>
```
right? BTW, there are only 3 occurrences where it is used in this script.
It would be just as clean to insert `std::chrono::seconds(5)` 3 times.

tkrokli:
OK, so what about deleting the file?

src/server/scripts/World/mob_generic_creature.cpp
Naios: TC has a typedef header for that: `#include "Duration.h"`

Shauren:
One script from there is registered but no idea if ever used (trigger_periodic)

PR #17753, opened 2016-08-07
Core/Spells: Implemented the function to set the next crit/noncrit of a spell

Seyden:
**Changes proposed:**
-  Set the next spell crit/noncrit for a specific target
-  Set Spellvalues over SpellScript
-  Possibility to set Crit over SpellValue

**Target branch(es):** 6.x

Tested with Demon hunter spell Chaos Strike.
Chaos Strike should have a shared crit roll for main and off hand and it worked.
If Main hand crits, off hand will too.. if main hand doesn't crit, the off hand won't too

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8156,6 +8156,13 @@ float Unit::GetUnitSpellCriticalChance(Unit* victim, SpellInfo const* spellProto
     // not critting spell
     if ((spellProto->HasAttribute(SPELL_ATTR2_CANT_CRIT)))
         return 0.0f;
+	
+    if (victim && HasNextSpellCritData(victim, spellProto->Id))
+    {
+        float nextSpellCritChance = IsNextSpellACrit(victim, spellProto->Id) ? 100.0f : 0.0f;
+        const_cast<Unit*>(this)->RemoveNextSpellCritData(victim, spellProto->Id);
+        return nextSpellCritChance;
+    }
 
     float crit_chance = 0.0f;
     switch (spellProto->DmgClass)
@@ -16178,4 +16185,35 @@ void Unit::SendCombatLogMessage(WorldPackets::CombatLog::CombatLogServerPacket* 
 bool Unit::VisibleAuraSlotCompare::operator()(AuraApplication* left, AuraApplication* right) const
 {
     return left->GetSlot() < right->GetSlot();
+}
+
+bool Unit::IsNextSpellACrit(Unit* target, uint32 spellId) const
+{
+    auto itr = m_spellCritMap.find(std::pair<Unit*, uint32>(target, spellId));
+
+    return itr != m_spellCritMap.end() && itr->second == true;
+}
+
+void Unit::SetNextSpellCrit(Unit* target, uint32 spellId, bool crit)
+{
+    auto itr = m_spellCritMap.find(std::pair<Unit*, uint32>(target, spellId));
+
+    if (itr == m_spellCritMap.end())
+        m_spellCritMap.emplace(std::pair<Unit*, uint32>(target, spellId), crit);
+}
+bool Unit::HasNextSpellCritData(Unit* target, uint32 spellId) const
+{
+    auto itr = m_spellCritMap.find(std::pair<Unit*, uint32>(target, spellId));
+
+    return itr != m_spellCritMap.end();
+}
+
+void Unit::RemoveNextSpellCritData(Unit* target, uint32 spellId)
+{
+    auto itr = m_spellCritMap.find(std::pair<Unit*, uint32>(target, spellId));
+
+    if (itr != m_spellCritMap.end() && !m_spellCritMap.empty()) {
+
+        m_spellCritMap.erase(std::pair<Unit*, uint32>(target, spellId));
+    }
 }
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -2707,6 +2707,12 @@ void Unit::_UpdateSpells(uint32 time)
         }
     }
 
+    for (auto nexctSpellCrit : m_spellCritMap)
+    {
+        if (!nexctSpellCrit.first.first)
+            m_spellCritMap.erase(nexctSpellCrit.first);
+    }
+
     _spellHistory->Update();
 }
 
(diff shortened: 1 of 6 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -16179,3 +16186,34 @@ bool Unit::VisibleAuraSlotCompare::operator()(AuraApplication* left, AuraApplica
...
+    auto itr = m_spellCritMap.find(std::pair<Unit*, uint32>(target, spellId));
+
+    return itr != m_spellCritMap.end();
+}
+
+void Unit::RemoveNextSpellCritData(Unit* target, uint32 spellId)
+{
+    auto itr = m_spellCritMap.find(std::pair<Unit*, uint32>(target, spellId));
+
+    if (itr != m_spellCritMap.end() && !m_spellCritMap.empty()) {
```
P-Kito: codestyle

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -8156,6 +8156,13 @@ float Unit::GetUnitSpellCriticalChance(Unit* victim, SpellInfo const* spellProto
     // not critting spell
     if ((spellProto->HasAttribute(SPELL_ATTR2_CANT_CRIT)))
         return 0.0f;
+	
+    if (victim && HasNextSpellCritData(victim, spellProto->Id))
```
P-Kito: does victim need to be checked here?

PortgasDAce: yes look a bit further, there is a victim check too

Pixel-Pirate:
nice job

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -8156,6 +8156,13 @@ float Unit::GetUnitSpellCriticalChance(Unit* victim, SpellInfo const* spellProto
...
+    {
+        float nextSpellCritChance = IsNextSpellACrit(victim, spellProto->Id) ? 100.0f : 0.0f;
+        const_cast<Unit*>(this)->RemoveNextSpellCritData(victim, spellProto->Id);
```
jackpoz: does the function still need to be const ?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -16179,3 +16186,33 @@ bool Unit::VisibleAuraSlotCompare::operator()(AuraApplication* left, AuraApplica
...
+}
+
+bool Unit::HasNextSpellCritData(Unit* target, uint32 spellId) const
+{
+    auto itr = m_spellCritMap.find(std::pair<Unit*, uint32>(target, spellId));
+
+    return itr != m_spellCritMap.end();
+}
+
+void Unit::RemoveNextSpellCritData(Unit* target, uint32 spellId)
```
jackpoz: what happens if the target creature is removed from world before the spell crit is used ? will m_spellCritMap still hold a reference to it ?

Naios:
I don't like the approach of adding a whole hash map to the unit class because it is used for creatures and players and entails heap memory allocation on every unit/player object creation -> fragmentation.

Isn't it possible to add the option to the `Spell` class and launch it through `CastCustomSpell` or something?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -16179,3 +16186,33 @@ bool Unit::VisibleAuraSlotCompare::operator()(AuraApplication* left, AuraApplica
 {
     return left->GetSlot() < right->GetSlot();
 }
+
+bool Unit::IsNextSpellACrit(Unit* target, uint32 spellId) const
+{
+    auto itr = m_spellCritMap.find(std::pair<Unit*, uint32>(target, spellId));
```
Naios:
```c++
auto itr = m_spellCritMap.find(std::make_pair(target, spellId));
```

P-Kito:
We have added that function for Auras. Once the PR is merged, we can give various parameters of any datatype from one aura to another. 
This was not yet implemented for spell casts (yet).

In this case both spells are triggered in one "head" spell for a reason.
Main spell
 -> effect 0: triggers mainhand damage
 -> effect 1: triggers offhand damage
When effect 0 crits, effect 1 should crit too. Simple as that.

src/server/game/Entities/Unit/Unit.cpp
PortgasDAce: ty

src/server/game/Entities/Unit/Unit.cpp
PortgasDAce: thx, fixed

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -2707,6 +2707,12 @@ void Unit::_UpdateSpells(uint32 time)
         }
     }
 
+    for (auto nexctSpellCrit : m_spellCritMap)
+    {
+        if (!nexctSpellCrit.first.first)
```
jackpoz: could you explain more in detail what this check does and how it does it please ?

PortgasDAce: the first "first" gets me the pair.. and with the other first i'll get the Unit.. so i check if the Unit in the map is null or not.. if the Unit is null, i remove them..

Do you want to say i should comment this line?

jackpoz: did you debug this code ? I'm quite sure a debug session will show that the NULL check has nothing to do with checking if a Unit* is still valid

PortgasDAce: not yet, will do if i'm home.. wanted to test it. Pushed it first in the repo.. but thanks for your hint, i will try later

PortgasDAce: ahhh should i use IsInWorld?

jackpoz: please don't make us review untested code. test it, debug it then push it.

PortgasDAce: Sry will test it first.

src/server/game/Spells/Spell.h
```diff
@@ -274,13 +274,21 @@ class TC_GAME_API SpellCastTargets
         std::string m_strTarget;
 };
 
+enum SpellValueCritFlag : uint8
```
Temppus: Shouldn't this be regular enum values instead of "flag" enum ?

tkrokli: It is a valid way of making sure that the value returned by `CritFlag` via `SpellValueCritFlag` is `uint8` instead of regular `int`. Check how they are used in `Spell.cpp` above and notice how `CritFlag` and `SpellValueCritFlag` don't need to be defined as `uint8` there.

P-Kito: I think you got that wrong tkrokli. He means not to store flags eg:  1, 2, 4, 8, 16.

Temppus: Yep P-Kito got me right. It seems that these should be normal values, not flags. My point is to rename enum and values naming. Word "flag" should not be there, cause it is not bitmask.

funjoker:
Wtf

Shauren:
i said it will not be merged

funjoker:
due to which reasons ?
Maybe feedback will make people rework etc.

PR #17759, opened 2016-08-09
Core/Scripts: remove 2 unused scripts

ghost:
**Changes proposed:**

These scripts are no longer in use, may as well remove them.

**Issues addressed:** None in particular, just general cleanup.

**Tests performed:** (Does it build, tested in-game, etc) Builds without issues. Script removal only.

None that I know about.


src/server/scripts/World/mob_generic_creature.cpp
```diff
@@ -236,7 +79,6 @@ class trigger_death : public CreatureScript
 
 void AddSC_generic_creature()
 {
-    //new generic_creature;
     new trigger_periodic();
     //new trigger_death;
```
Shauren: as you see here, trigger_death is also commented out (and unused)

tkrokli: Thanks for confirming. :+1: Will fix.

src/server/scripts/World/world_script_loader.cpp
```diff
@@ -41,7 +40,6 @@ void AddWorldScripts()
 {
     AddSC_areatrigger_scripts();
     AddSC_emerald_dragons();
-    AddSC_generic_creature();
```
Shauren: now you went overboard with changes...

tkrokli: Oh. right. That is only needed when removing the file. I will undo that.

PR #17763, opened 2016-08-10
Core/Quest: Delete only required amount of items unless item is quest

runningnak3d:
**Changes proposed:**
-  Only delete all quest bound items on turn in if they can not be used on another quest

**Issues addressed:** Closes # 17758

- Builds

- [ ]  Need to check if there are any quests that would require all quest bound, multi-drop items to NOT be deleted
- [ ] Need to check if there are any quests the would require all quest bound, but NOT multi-drop items to be deleted.

``` SQL
SELECT `entry`, `name` from `item_template` WHERE `bonding`=4 AND Flags & 0x800; // Use this list to check the first issue
SELECT `entry`, `name` from `item_template` WHERE `bonding`=4 AND Flags &~ 0x800; // Use this query to check the second issue
```


runningnak3d:
Closing -- better solution incoming. Thanks @Shauren

PR #17764, opened 2016-08-10
Core/Quest: Only delete bound quest items on turn in if the following

### LINKED ISSUE #17758: [3.3.5] Quest items related to multiple quests and repeatable quests being all destroyed on quest turn in
**Description**:
There are quest items that can relate to multiple quests or are used by repeatable quests and due to the latest changes they're all destroyed now whenever the first quest they relate to is turned in.

They're not supposed to be destroyed... just the instance of them required to complete the quest being turned in.

1.  Add quests 4723, 4728, 4730, 4733 ("Beached Sea Creature")
2.  Try to turn in any of the aforementioned quests
3.  Notice every Sea Creature Bones in your inventory are deleted leaving you incapable of turning in any of the remaining quests, unless you drop them and pick up again.

The same happens to the stranglethorn fishing extravaganza quest, where on the first fish turn in you get all of the 250 fishes you may have destroyed...

video_1470745853

335

**TC hash/commit**:  
https://github.com/TrinityCore/TrinityCore/commit/ac35f93c2d6d5436054be474da06c9483e311c36

**TDB version**:  
TDB 335.61

[//]: # (For fixes containing c++ create a Pull Request)

runningnak3d:
**Changes proposed:**
-  Delete ALL bound quest items on turn in only if they meet the following conditions, otherwise delete only the required amount for the quest:
1. Not repeatable 
2. The item isn't a requirement of another quest in the players quest log  

**Issues addressed:** Closes #17758 #17806 

- Builds
- Tested with quests 4730 and 4733
  - .quest add 4730
  - .quest add 4733
  - Turn in 4730, and the second bones are still in your bag
- Tested with quest 582 after modifying the quest requirement to be item 12242 (seemed fitting)
- Test 1:
  - .quest add 582
  - .add 12242 21
  - Turn in quest and all 21 bones were deleted
- Test 2 (this is why I changed the requirement to item 12242 -- no other quest needs the heads):
  - .quest add 582
  - .quest add 4730
  - .quest add 4733
  - .add 12242 26
  - Turn in quest and only the required amount were deleted, leaving bones to be turned in for the other quests.

- [x]  Check if it works
- [x] Test additional quests to make sure that nothing was broken. 
- I removed the fringe case WIP because 1 - I don't think there are any. 2 - if there are, someone will have to open a ticket if they find them


jackpoz:
> Have no ability to test this in game since I am at work :)

please find the time to test it

runningnak3d:
I have tested every quest that was listed in the original issues that brought on the original PR, and I have tested all the quests listed as issues since then. I need other people to test to make sure there are no fringe cases.

jackpoz:
do the last changes require a new session of tests ? since it's the last version that is going to be merged

runningnak3d:
The last version is the one that was tested -- and passed :)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -16440,12 +16441,16 @@ bool Player::HasQuestForItem(uint32 itemid) const
...
-        if (q_status.Status == QUEST_STATUS_INCOMPLETE)
+        if (q_status.Status == QUEST_STATUS_INCOMPLETE || q_status.Status == QUEST_STATUS_COMPLETE)
         {
             Quest const* qinfo = sObjectMgr->GetQuestTemplate(questid);
             if (!qinfo)
                 continue;
 
+            if (questid == excludeQuestId)
+                excludeQuestId = 0;
+                return true;
```
Shauren: passed my ass.

runningnak3d: Yes, it was tested with my cock in your ass. What the fuck is wrong with you dude?

DDuarte: Brackets are needed. It's currently always returning true. Indentation alone is not enough.

runningnak3d: Yea, I get that. It just so happens that always returning true just happened to make it look like the patch was working as intended. As I explained to Shauren in IRC. A simple, "Hey, some braces here...." would have been enough. If I say I test something, then I test it.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -16440,12 +16441,18 @@ bool Player::HasQuestForItem(uint32 itemid) const
 
         QuestStatusData const& q_status = qs_itr->second;
 
-        if (q_status.Status == QUEST_STATUS_INCOMPLETE)
+        if (q_status.Status == QUEST_STATUS_INCOMPLETE || q_status.Status == QUEST_STATUS_COMPLETE)
```
Shauren: Including QUEST_STATUS_COMPLETE here allows quest items to drop even if player completes the objective, see the use of Player::HasQuestForItem in LootItem::AllowedForPlayer

runningnak3d: Now that was a constructive comment. That is very easily fixed by adding a bool because I need to check for both, but yes, I see how that breaks other uses.

src/server/game/Entities/Player/Player.cpp
jackpoz: Please debug your code line by line and add breakpoints to ensure all cases are tested.
There's no need to say bad words to who is reviewing your code for mistakes you made.

Aokromes:
What's missing here?

tkrokli:
@runningnak3d : you need to rebase your repo on current 3.3.5 source to solve the conflicts. There has been a lot of changes in the core files since you opened this PR, so you need to incorporate the changes into your PR branch files.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -16445,7 +16445,7 @@ bool Player::HasQuestForItem(uint32 itemid) const
 
         QuestStatusData const& q_status = qs_itr->second;
 
-        if (q_status.Status == QUEST_STATUS_INCOMPLETE)
+        if ((q_status.Status == QUEST_STATUS_INCOMPLETE) || (turnIn && (q_status.Status == QUEST_STATUS_INCOMPLETE || q_status.Status == QUEST_STATUS_COMPLETE)))
```
Shauren: You don't need to check `q_status.Status == QUEST_STATUS_INCOMPLETE` twice

runningnak3d: Thank you

src/server/game/Entities/Player/Player.cpp
```diff
@@ -16460,7 +16460,7 @@ bool Player::HasQuestForItem(uint32 itemid) const
             // This part for ReqItem drop
             for (uint8 j = 0; j < QUEST_ITEM_OBJECTIVES_COUNT; ++j)
             {
-                if (itemid == qinfo->RequiredItemId[j] && q_status.ItemCount[j] < qinfo->RequiredItemCount[j])
+                if ((itemid == qinfo->RequiredItemId[j] && q_status.ItemCount[j] < qinfo->RequiredItemCount[j]) || (turnIn && itemid == qinfo->RequiredItemId[j] && q_status.ItemCount[j] <= qinfo->RequiredItemCount[j]))
```
Shauren: The same applies here, duplicated `itemid == qinfo->RequiredItemId[j]` is not needed
`if (itemid == qinfo->RequiredItemId[j] && (q_status.ItemCount[j] < qinfo->RequiredItemCount[j] || (turnIn && q_status.ItemCount[j] <= qinfo->RequiredItemCount[j])))`
I would also want to know why are you using `<=` comparison during quest turn in, what does it change?

runningnak3d: Yea, I saw this after I already committed the removal of the other redundant check, but I had already logged out for the day. As soon as I can get back on my dev box I was going to push it. But again, thanks.

The <= is needed because this check would fail if (for example) you had 2 quests that needed item 1234 and you had 4 of that item. The first quest needs 2, and the second quest needs 2. If you just check for < it would fail because you have the correct amount for the second quest. So on turnin, all of item 1234 would be deleted and you are up the creek trying to turning the second quest. There is never a situation where you would be turning in a quest and need to check for just < and of course we don't care about > :)

Aokromes:
@runningnak3d this is completed?

runningnak3d:
I listed the tests performed, and the issues it addresses :)

I tried to find any fringe cases where you would want to delete all of an item even though you have other quests in your log that required that item, and I couldn't find any. If I recall, @Shauren even pointed out that these are the things the client checks for ...

PR #17774, opened 2016-08-12
Core/Custom: Race masquerading system

Treeston:
We haven't had one of these "fun custom features" added to the core for a while, have we?

Well, I think it's about time we changed that. To that end, I've managed to finalize a pet project that I'd been idly playing with for a while now, with thanks to some client internals help from @Shauren.

Introducing: the race masquerading system. It's a purely cosmetic (optional, default off) system that allows you to play any class on any faction while _looking_ like any race you want. Here's how this works.

As far as the core is concerned, the character is the same as it's always been - a valid class/race combination; say, a Human Rogue. However, using the `.character masquerade` command, you ask the core to let you look like a Tauren Rogue instead.
Thanks to the masquerading system, the core complies and changes your native display ID to that used by a Tauren. In addition, it also tricks other players' clients into showing your character's race as "Tauren" (this is actually necessary due to how the client displays character skins), as well as (thanks to some update timing magic) makes your own client _display_ a properly-skinned Tauren Rogue while still treating it as a Human Rogue, meaning you have access all the languages and spellbook tabs you would normally have.

This means you can now play a Tauren Rogue, with all the high-velocity beef stealth this implies, while still being treated as a Human Rogue gameplay wise - the transformation is purely cosmetic. You get the Human racials, are member of the Alliance, have access to all the respective reputation factions and so forth.

:client magic:


Sar777:
:panda_face:

Keader:

tkrokli:
Treeston's branch compared to TC 3.3.5:
- `3.3.5-race-masquerade`: `This branch is 3411 commits ahead, 6551 commits behind TrinityCore:6.x.`
- `TrinityCore:3.3.5`: `This branch is 3411 commits ahead, 6551 commits behind 6.x.`

Must be something in one of the last 6 commits to `TrinityCore:3.3.5` which is needed to fix the conflict.

Treeston:
Yeah, the version I pushed to head has a different comment. Fixing.

tkrokli:
Sorry for bothering you with this detail, but this PR adds a blank line in your `Player.cpp L 1660` :
https://github.com/TrinityCore/TrinityCore/pull/17774/files#diff-5589ee47b3a7ad220cff105a8fe7eac3R1660

Shauren:
Custom features are fine to be added when their impact isn't too large
This however does not fit into that category with the huge hacks used to cheat the wow client (`WorldObject::BuildUpdate`, `Unit::BuildValuesUpdate` with the latter also being a point of past performance issues)
Given the reasons above it cannot be accepted, even if it can be disabled with a config (because the performance penalty in BuildValuesUpdate is paid anyway, even if disabled). Using preprocessor directives to enable it at compile time would also not be acceptable (because that in turn leads to unmaintainable/unreadable code)

PR #17782, opened 2016-08-14
Core/GameEvent: weekend xp rate

Aokromes:
**Changes proposed:**
-  Allows you to increase XP rate during weekends.


src/server/game/World/World.cpp
```diff
@@ -460,6 +460,8 @@ void World::LoadConfigSettings(bool reload)
     rate_values[RATE_XP_BG_KILL]  = sConfigMgr->GetFloatDefault("Rate.XP.BattlegroundKill", 1.0f);
     rate_values[RATE_XP_QUEST]    = sConfigMgr->GetFloatDefault("Rate.XP.Quest", 1.0f);
     rate_values[RATE_XP_EXPLORE]  = sConfigMgr->GetFloatDefault("Rate.XP.Explore", 1.0f);
+    rate_values[RATE_XP_WEEKEND]  = ConfigMgr::GetFloatDefault("Rate.XP.Weekend", 1.0f);
```
lorac2k14: Shouldn't this be `rate_values[RATE_XP_WEEKEND]  = sConfigMgr->GetFloatDefault("Rate.XP.Weekend", 1.0f);` ?

src/server/game/World/World.cpp
```diff
@@ -460,6 +460,8 @@ void World::LoadConfigSettings(bool reload)
...
+    m_int_configs[CONFIG_RATE_XP_WEEKEND_EVID] = ConfigMgr::GetIntDefault("Rate.XP.Weekend.EVID", 65);
```
lorac2k14: `m_int_configs[CONFIG_RATE_XP_WEEKEND_EVID] = sConfigMgr->GetIntDefault("Rate.XP.Weekend.EVID", 65);`
or am I incorrect

src/server/game/World/World.cpp
tkrokli: Very likely. Copy & Paste errors, all 11 of them. :keycap_ten: + :one:

Aokromes: Maybe xd it's 4 years old code no one updated xd

lorac2k14: I'm just trying to help ya out ;)
I noticed inconsistencies and just wanted to point them out and maybe this will help debug build :)

src/server/game/World/World.cpp
tkrokli: The value there is an int, not a float, so it looks like it doesn't have to be `sConfigMgr`.
Look in the travis-ci compile log and you will only find that it picks on L#463, not L#464.

Odyssey: I feel like the event Id shouldn't be a config setting since if it gets merged everyone using a clean database will have the same official default id set in the sql file above.

lorac2k14:
I for one would love to see something like this implemented and it is a step twords "even though this is for 3.3.5a"  implementing the weekly events for WoD and above

src/server/game/World/World.cpp
```diff
@@ -460,8 +460,8 @@ void World::LoadConfigSettings(bool reload)
     rate_values[RATE_XP_BG_KILL]  = sConfigMgr->GetFloatDefault("Rate.XP.BattlegroundKill", 1.0f);
     rate_values[RATE_XP_QUEST]    = sConfigMgr->GetFloatDefault("Rate.XP.Quest", 1.0f);
     rate_values[RATE_XP_EXPLORE]  = sConfigMgr->GetFloatDefault("Rate.XP.Explore", 1.0f);
-    rate_values[RATE_XP_WEEKEND]  = sConfigMgr::GetFloatDefault("Rate.XP.Weekend", 1.0f);
-    m_int_configs[CONFIG_RATE_XP_WEEKEND_EVID] = sConfigMgr::GetIntDefault("Rate.XP.Weekend.EVID", 65);
+    rate_values[RATE_XP_WEEKEND]  = sConfigMgr->GetFloatDefault("Rate.XP.Weekend", 1.0f);
+    m_int_configs[CONFIG_RATE_XP_WEEKEND_EVID] = sConfigMgr->GetIntDefault("Rate.XP.Weekend.EVID", 65);
```
Nawuko: Default should be 66 not 65

Aokromes: right, the patch is such old that when it was made the id 65 was free :)

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2029,10 +2029,14 @@ Rate.Drop.Item.ReferencedAmount = 1
 #        Default:     1 - (Rate.XP.Kill, affects only kills outside of Battlegrounds)
 #                     1 - (Rate.XP.Quest)
 #                     1 - (Rate.XP.Explore)
-
-Rate.XP.Kill    = 1
-Rate.XP.Quest   = 1
-Rate.XP.Explore = 1
+#                     1 - (Rate.XP.Weekend)
+#                    66 - (Rate.XP.Weekend.EVID)
```
Treeston: Some explanation on what this is in the .conf comment would be nice.

Aokromes: Thinking on switch to http://www.boost.org/doc/libs/1_39_0/doc/html/date_time/gregorian.html to avoid the need of use event.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -6478,6 +6478,7 @@ void Player::CheckAreaExploreAndOutdoor()
                     XP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*sWorld->getRate(RATE_XP_EXPLORE));
                 }
 
+                XP = uint32(XP * (IsEventActive(sWorld->getIntConfig(CONFIG_RATE_XP_WEEKEND_EVID)) ? sWorld->getRate(RATE_XP_WEEKEND) : 1.0f));
```
donato: This code is duplicated 5 times, it may be worthy of a helper function!
```c++
uint32 getModifiedXP(uint32 startingXP) {
    return startingXP * (IsEventActive(sWorld->getIntConfig(CONFIG_RATE_XP_WEEKEND_EVID)) ? sWorld->getRate(RATE_XP_WEEKEND) : 1.0f));
}
```

Treeston: That's a terrible name for a method.

donato: I'm open to constructive feedback, if you have a better idea :)

Ascathor: uint32 getFinalizedXP(uint32 initial XP)

Treeston: Alright, but make it more easy to later expand as well. Get rid of the compact syntax, and expand it into an if branch.

```
uint32 getFinalizedXP(uint32 initialXP)
{
    if (whatever)
        initialXP *= whatever;
    return initialXP;
```

Ascathor: Seems good that way. I like it (y)

src/server/game/Miscellaneous/Formulas.h
```diff
@@ -228,6 +229,16 @@ namespace Trinity
             sScriptMgr->OnGroupRateCalculation(rate, count, isRaid);
             return rate;
         }
+
+        inline float xp_on_weekend_rate(uint32 initial_rate = 1.0f)
+        {
+            if(IsEventActive(sWorld->getIntConfig(CONFIG_RATE_XP_WEEKEND_EVID)))
```
Aokromes: TODO: convert it to boost::gregorian or c++ equivalent

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2029,10 +2029,14 @@ Rate.Drop.Item.ReferencedAmount = 1
...
-Rate.XP.Quest   = 1
-Rate.XP.Explore = 1
+#                     1 - (Rate.XP.Weekend)
+#                    66 - (Rate.XP.Weekend.EVID EventID (from game_event.eventEntry))
+
+Rate.XP.Kill         = 1
+Rate.XP.Quest        = 1
+Rate.XP.Explore      = 1
+Rate.XP.Weekend      = 1
+Rate.XP.Weekend.EVID = 66
```
Aokromes: TODO: convert it to boost::gregorian or c++ equivalent

src/server/game/Miscellaneous/Formulas.h
greenbagels: Just noticed I gave the uint32 initial_rate a float value, should be a float rate...

r00ty-tc:
Just a bit of an idea here. Could the days config not be bitmapped to allow any combination of days, rather than just a string of days together starting Fri/Sat?

Rochet2:
> Would it be easy to implement?

You can do `return (1 << now.tm_wday) & weekendXPDays != 0;`
This would mean if config is 0 then nothing happens,
if config is 1 sunday is double, if config is 2 monday is double, if config is 3 sunday and monday is double and so on, where each day is a power of 2.

This would be an edit to IsWeekendActive and would probably require renaming the function etc.

r00ty-tc:
Pretty much that, and of course annotate the worldserver.conf.dist to reflect the change.

Aokromes:
@Rochet2 if you want / can takeover this pr it's welcomed :P btw acording @ariel- some quest bonus thing is needed.

r00ty-tc:
Another thing I mentioned in IRC was that it might be nicer to make a generic "Bonus XP" class, and have that go off to only weekend XP now (but leave the way open for other optional bonus XP features to be plugged in). So, the actual hook in the XP gain places is done once, and bonus XP checks config and applies any bonus XP due, due to config.

I can help, but time is limited as always with me :(

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2843,6 +2843,23 @@ AllowTrackBothResources = 0
 
 PlayerStart.AllReputation = 0
 
+#
+#    Weekend.XP.Days
```
joschiwald: i would suggest to use a game_event to enable holiday xp rates

Kittnz: Several id's?

joschiwald: why?

Kittnz: Oh nevermind, i didn't read it correctly.

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2843,6 +2843,23 @@ AllowTrackBothResources = 0
...
+#        Description: Enables XP increase during last days of the week.
+#
+#        Default:     0 - (Disabled)
+#                     2 - (Weekend - Saturdays and Sundays)
+#                     7 - (Always active)
```
Krudor: Wouldn't it make more sense to just have 0 and 1 as values, for if weekend bonus experience should be applied or not?
7 for "Always active" doesn't make much sense, you'd just use the other config for xp rates.
0, 2 and 7 doesn't make much sense tbh.

There's also another option of making this a bitmask, allowing any week day to be chosen as the bonus xp "weekend".
But then joschiwald's game_event suggestion starts to make more sense.

Either way I'd remove the "Always" active options and stick with 0 and 1 as values for bonus weekend is "not active"/"active"

jackpoz:
would anyone like to complete this PR with the suggestions above ?

illfated:
Looks like Rotchet2 is going to test it in his own repo first. ^^

src/server/game/World/World.cpp
```diff
@@ -452,6 +452,10 @@ void World::LoadConfigSettings(bool reload)
     rate_values[RATE_XP_BG_KILL]  = sConfigMgr->GetFloatDefault("Rate.XP.BattlegroundKill", 1.0f);
     rate_values[RATE_XP_QUEST]    = sConfigMgr->GetFloatDefault("Rate.XP.Quest", 1.0f);
     rate_values[RATE_XP_EXPLORE]  = sConfigMgr->GetFloatDefault("Rate.XP.Explore", 1.0f);
+
+    m_int_configs[CONFIG_XP_BOOST_DAYMASK] = sConfigMgr->GetIntDefault("XP.Boost.Daymask", 0);
+    rate_values[RATE_XP_BOOST] = sConfigMgr->GetFloatDefault("XP.Boost.Rate", 1.0f);
```
Rochet2: Seems I forgot to change the default rate from 1 to 2 here as it was changed in config.
```suggestion
    rate_values[RATE_XP_BOOST] = sConfigMgr->GetFloatDefault("XP.Boost.Rate", 2.0f);
```

illfated: Nice suggest feature.

Faq: Shouldn't that magic number taken directly from config? So now if needed rate 3.0, it is needed to edit in 2 places?

illfated: I think you have misread the code (although I could be wrong).
`1.0f` / `2.0f` is the default value, while `XP.Boost.Rate` reads the config value.
I presume the default is included in case the config value is missing, out of scope or invalid.

Rochet2:
The above fixes the bonus applying to quests, implements mask for the days and renames weekend to boost. (after all is merged to TClegacy repo)
Was there some other things needed?

src/server/scripts/World/boosted_xp.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "ScriptMgr.h"
+#include "World.h"
+
+#include <boost/date_time/posix_time/posix_time.hpp>
+
+namespace
+{
+    bool IsXPBoostActive()
+    {
+        auto now = boost::posix_time::to_tm(boost::posix_time::from_time_t(GameTime::GetGameTime()));
```
jackpoz: I wonder if a return early case like "if (weekdayMaskBoosted == 0) return false;" will avoid checking every time for gametime and weekday.

src/server/scripts/World/world_script_loader.cpp
```diff
@@ -56,4 +57,5 @@ void AddWorldScripts()
     if (sWorld->getBoolConfig(CONFIG_IP_BASED_ACTION_LOGGING))
         AddSC_action_ip_logger(); // location: scripts\World\action_ip_logger.cpp
     AddSC_duel_reset();
+    AddSC_xp_boost();
```
jackpoz: maybe add the script only if sWorld->getIntConfig(CONFIG_XP_BOOST_DAYMASK) != 0 to avoid calling it every time even if disabled ?

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2843,6 +2843,30 @@ AllowTrackBothResources = 0
 
 PlayerStart.AllReputation = 0
 
+#
+#    XP.Boost.Daymask
+#        Description: Enables experience boost during the defined days. This field is a bitmask.
+#
+#        Default:     0 - (Disabled)
+#                     1 - (Sunday)
```
jackpoz: does this depend on the OS weekday definition ? (for example Egypt has a different weekend)

jackpoz (changes requested):
Please check the comments in the review (I marked them as "Request changes" but it's more of an open discussion)

src/server/scripts/World/world_script_loader.cpp
Rochet2: Hmm, that would disable the ability to use `.reload config` to enable and disable the script though.

src/server/worldserver/worldserver.conf.dist
Rochet2: The code uses `now.tm_wday` and tm_wday is defined as `days since Sunday – [0, 6]` as seen here https://en.cppreference.com/w/c/chrono/tm
I tested locally to enable monday only yesterday (which is the first day of the week here and in my system) and the system works as documented in the config (monday = 2).
From what I see it is not related to OS or locale etc. according to my tests or the documentation for `tm` struct.

src/server/scripts/World/world_script_loader.cpp
jackpoz: well it seems like 2 lines above that happens already with if (sWorld->getBoolConfig(CONFIG_IP_BASED_ACTION_LOGGING))

src/server/worldserver/worldserver.conf.dist
jackpoz: Thanks for the info 👍

src/server/scripts/World/boosted_xp.cpp
Rochet2: Hmm, not sure both of these changes are needed. Just either one.
During runtime the optimization in IsXPBoostActive would be just an overhead as the script is already disabled in scriptloader if the script is off. Though I guess adding a single IF statement probably isnt a huge bottleneck..

jackpoz: yeah, the check in AddWorldScripts() is enough

src/server/scripts/World/world_script_loader.cpp
ariel-: Could a WorldScript (afaik that was the name) be used to hook reload config event and enable/disable script on demand?

Shauren: No, there is currently no way to disable a single script

jackpoz (approved):

jackpoz:
thanks all for the help for this PR :)

PR #17784, opened 2016-08-15
Core/DungeonFinder: disable map only for LFG

Aokromes:
**Changes proposed:**
-  disable map only from looking for dungeon will allow to disable one map for dungeon finder and allow to enter people walking

### DIFF
--- a/src/server/game/Conditions/DisableMgr.cpp
+++ b/src/server/game/Conditions/DisableMgr.cpp
@@ -118,6 +118,7 @@ void LoadDisables()
             case DISABLE_TYPE_QUEST:
                 break;
             case DISABLE_TYPE_MAP:
+            case DISABLE_TYPE_LFG_MAP:
             {
                 MapEntry const* mapEntry = sMapStore.LookupEntry(entry);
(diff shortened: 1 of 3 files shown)


Aokromes:
ariel sugested to change this to a flag.

Takenbacon:
You need to increment MAX_DISABLE_TYPES

RedSonja:
I am not sure if this is the correct place to put this or not, but the Dungeon Finder does not work at all for me. Every player is locked out of every dungeon even if they have never been in any of the ones in the Dungeon Finder. I have looked but this was the only thing I could find related to it. For me, it has not worked for years. Just wondering if it has anything to do with this disable map you are talking about, and if that was put into the core on Nov 25, 2016, then shouldn't it be working now?

illfated:
You should report the problem in a new issue ticket (if no open issue ticket exists). Necro-posting in old and completed pull requests will never be practical for reporting the problem in a way that it can be followed up on.

RedSonja:
Thank you made a new ticket.

PR #17787, opened 2016-08-15
Core/Movement: Improve unit chase and follow in 3D

chaodhib:
**Changes proposed:**
-  Improved the code of MoveChase and MoveFollow (and more specificly the MoveGen used by these methods) in order to better handle high slopes and chases in 3D.

**Target branch(es):** 3.3.5

**Issues addressed:** close https://github.com/TrinityCore/TrinityCore/issues/16549

**Tests performed:** Build and tested IG

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2236,11 +2236,25 @@ void WorldObject::GetNearPoint2D(float &x, float &y, float distance2d, float abs
     Trinity::NormalizeMapCoord(y);
 }
 
-void WorldObject::GetNearPoint(WorldObject const* /*searcher*/, float &x, float &y, float &z, float searcher_size, float distance2d, float absAngle) const
+void WorldObject::GetNearPoint(WorldObject const* searcher, float &x, float &y, float &z, float searcher_size, float distance2d, float absAngle, bool is3D) const
 {
-    GetNearPoint2D(x, y, distance2d+searcher_size, absAngle);
-    z = GetPositionZ();
-    // Should "searcher" be used instead of "this" when updating z coordinate ?
+    float totalDistance = distance2d + searcher_size;
+    if (is3D)
+    {
+        G3D::Vector3 vectSearcher(searcher->GetPositionX(), searcher->GetPositionY(), searcher->GetPositionZ());
+        G3D::Vector3 vectMe(GetPositionX(), GetPositionY(), GetPositionZ());
+        G3D::Vector3 contactPoint = (vectSearcher - vectMe).fastUnit() * totalDistance + vectMe;
+        x = contactPoint.x;
+        y = contactPoint.y; 
+        z = contactPoint.z;
+    }
+    else
+    {
+        GetNearPoint2D(x, y, totalDistance, absAngle);
+        // Should "searcher" be used instead of "this" when updating z coordinate ?
+        z = GetPositionZ();
+    }
+    z += 1.0f; // go up a bit before the Z normalization to avoid getting under the landscape. 
     UpdateAllowedPositionZ(x, y, z);
 
     // if detection disabled, return first point
@@ -2259,7 +2273,7 @@ void WorldObject::GetNearPoint(WorldObject const* /*searcher*/, float &x, float 
     // loop in a circle to look for a point in LoS using small steps
     for (float angle = float(M_PI) / 8; angle < float(M_PI) * 2; angle += float(M_PI) / 8)
     {
-        GetNearPoint2D(x, y, distance2d + searcher_size, absAngle + angle);
+        GetNearPoint2D(x, y, totalDistance, absAngle + angle);
         z = GetPositionZ();
         UpdateAllowedPositionZ(x, y, z);
         if (IsWithinLOS(x, y, z))
@@ -2302,7 +2316,7 @@ Position WorldObject::GetRandomNearPosition(float radius)
 void WorldObject::GetContactPoint(const WorldObject* obj, float &x, float &y, float &z, float distance2d /*= CONTACT_DISTANCE*/) const
 {
     // angle to face `obj` to `this` using distance includes size of `obj`
-    GetNearPoint(obj, x, y, z, obj->GetObjectSize(), distance2d, GetAngle(obj));
+    GetNearPoint(obj, x, y, z, obj->GetObjectSize(), distance2d, GetAngle(obj), true);
 }
 
 float WorldObject::GetObjectSize() const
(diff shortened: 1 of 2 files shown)


chaodhib:
The solution works but there are 2 issues:
* First, I'm not sure about the efficiency of this solution
* Second, there is an issue regarding the recheck distance mechanism in TargetedMovementGeneratorMedium<T, D>::DoUpdate. It lacks precision when fighting mobs with high combat reach. Let me explain:

When an unit chases another unit, first a "contact point" is determined. Then a path to arrive to this contact point is calculated. The contact point will be the point where the chaser will stand once it has finished its movement. On that contact point, the chaser will be in melee range of the target. 

On another hand, if the target itself moves, the contact point should be recalculated, right? Well, that's not always true. As an optimization, the contact point is recalculated only if the distance between the old position of the target and its new position is higher than a certain threshold. Well, long story short this optimization does not work well when the chaser has a "big" (higher than 3) combat reach, like here with the Lich King.

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2234,9 +2234,21 @@ void WorldObject::_GetNearPoint(float &x, float &y, float &z, float distance, fl
     UpdateAllowedPositionZ(x, y, z);
 }
 
-void WorldObject::GetNearPoint(float &x, float &y, float &z, float distance, float absAngle) const
+void WorldObject::GetNearPoint(float &x, float &y, float &z, float distance, float absAngle, bool distanceIn3D /*= false*/) const
 {
     _GetNearPoint(x, y, z, distance, absAngle);
+    const float ACCEPTABLE_ERROR = 0.5f;
```
Nawuko: Instead of doing something like this you could calculate a proper `distance` calculation that considers height difference from `this` object and the object that `requests` the GetNearPoint.

chaodhib:
@Nawuko you mean something like this?

chaodhib:
It's tested and it's working as intended. I think is solution is also pretty efficient. The change 2d -> 3d wont be noticeable in term of performance if I'm not mistaken.

Nawuko:
Have a look at here: https://github.com/TrinityCore/TrinityCore/compare/3.3.5...Nawuko:feature/meele_range_fix_tc?expand=1

Thats from my personal repo all changes that are fixing this issue on a clean tc.

Some notes: 
 - GetNearPoint3D is maybe a wrong name to describe the function correct.
 - GetNearPoint does nothing else than pointing to GetNearPoint3D
 - Removed 1 useless no longer needed worldserver.conf entry.

sirikfoll:
@Nawuko I'm saying like it it the same problem, forgive me if it's out of the scope, but an example:

.go xyz 6638.332031 1066.633057 297.816132 571 and aggro a nearby Undying Minion.
On clean TC the mob will climb the object, but not all the way to the top, he won't reach the player, and will not evade(because I guess he thinks that it can reach the player, same as LK case),

with your code, if the player gets right on the edge, the npc will climb everything until it reaches the player.
If you put .gm fly on and fly right above a creature he still no evade, and sometimes theystart acting weird, like if it's in a loop trying to find a path.

Just a side note, in retail those npcs don't even think in climb that object ^^

Nawuko:
@sirikfoll Yep, its out of scope. If you try this on a clean TC the npc tries to do the same (But you dont see it cause he dosent scan for different angles

Maybe there should be 1 more check to avoid constantly calling the GetNearPoint function if the target is in a unreachable location, I've already done that but i did it different as Treeston [inside the PathGen with a custom flag] so i can not port it easily back over to TC).

Treeston started work on this and from a TargetetMovementGenerator point of view its implemented if I recall correctly

Oh yea and the code was only there to help @chaodhib to maybe get another idea/point of view. Its not a PR and should not be used with a clean TC :)

chaodhib:
There are (at least) 3 different bugs at play here:

- First, the bug described in the issue https://github.com/TrinityCore/TrinityCore/issues/16549. Probably the easiest of the 3 to solve. 

Basically, the search for a "contact point" (the point where the NPC will be standing once it has finished its movement and can auto attack) is done in 2D. This way of searching is not working well when we are fighting in a high slope (like the Lich King's stairs). 

In this PR, the commit https://github.com/TrinityCore/TrinityCore/pull/17787/commits/64f1b3c9c41fa3414a0f1653d07cd4197b8d9b4d and the ones before are enough to fix this. A fully 3D search has been implemented and further testing needs to be made but it doesn't look much slower than the 2D method. And this solution is more efficient than @Nawuko's solution, in situations like the LK on the stairs, since the right point is found at the first try.

- Right at the end of the search for the contact point, the potential contact point has its Z coordinate normalized. What I mean by that is the Z is changed (moved down) until the ground is hit. Doing so ensure that if the moving unit is nor flying nor swimming, then the contact point should be on the ground. The problem is that no check is done to ensure that this contact point "post-Z-normalization" is at the required range (cf https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Object/Object.cpp#L2244 if the point is in LoS, no further check is done and the function call finishes)

- Once a good contact point has been determined, it is the PathGenerator's job to find a path from the current position of the unit to the contact point. And its also its job to NOT generate any path in case the unit should not be able to reach that point because of physics and the map geometry. But because it is generating a valid path, the npc does not go into evade mode in situations where it should.

Also, side note, one more bug in the movement system:
- From sniff analysis, I believe pets follow their masters at a 3 constant yard distance.

Anyway, I'll see how I can combine @Nawuko's code and mine to fix the issues 1 and 2. Issues 3 and 4 should be the subject of another PR.

Nawuko:
Yep summs it up pretty well. @chaodhib the method in https://github.com/TrinityCore/TrinityCore/pull/17787/commits/64f1b3c9c41fa3414a0f1653d07cd4197b8d9b4d is faster cause of 1 less mapheight call. But it dosent work for custom absAngle also there could be a problem with cliffs or when the searcher is undefined. Since GetNearPoint is used in many different cases i opted to remove a very similar code.

Also yes, Point 3 is currently a PathGenerator problem, since the PathGen generates paths that are "as close as possible" and did not abort on "uncompleted" paths. I added a custom flag that gets set if a absolut closes point is reached but the path is not completed. In this case the mob should evade (Retail behavior is similar, the mob "tries" first to get as close as possible and after that it waits shortly and then evades if the path still cant completed).

src/server/game/Entities/Object/Object.h
```diff
@@ -436,7 +436,7 @@ class TC_GAME_API WorldObject : public Object, public WorldLocation
         virtual void RemoveFromWorld() override;
 
         void GetNearPoint2D(float &x, float &y, float distance, float absAngle) const;
-        void GetNearPoint(WorldObject const* searcher, float &x, float &y, float &z, float searcher_size, float distance2d, float absAngle) const;
+        void GetNearPoint(WorldObject const* searcher, float &x, float &y, float &z, float searcher_size, float distance2d, float absAngle, bool is3D = false) const;
```
donato: I'm not sure this is the right API.

Given we have GetNearPoint2D, the clearest thing to me, would be to have a GetNearPoint3D, and avoid the whole bool-flag thing.

In general, [rules](http://www.planetgeek.ch/wp-content/uploads/2014/11/Clean-Code-V2.4.pdf) of thumb should be 
 * No flag arguments in functions 
 * Method with too many arguments

chaodhib:
superseded by https://github.com/TrinityCore/TrinityCore/pull/17850

PR #17791, opened 2016-08-16
Core/Gameobject: Fix greeting (Questgiver-)Gameobjects

Gerhood:
**Changes proposed:**

_Is:_ If a questgiver-GO has no text assigned and a player is not on or eligible for a quest with that GO, a default message will be issued (Greetings, $N)

_Should be:_ In such cases the player should not be able to interact with that GO

**Target branch(es):** 3.3.5

**Tests performed:** Thoroughly tested, did not recheck sniffs

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -2240,8 +2240,9 @@ void GameObject::BuildValuesUpdate(uint8 updateType, ByteBuffer* data, Player* t
                 switch (GetGoType())
                 {
                     case GAMEOBJECT_TYPE_QUESTGIVER:
-                        if (ActivateToQuest(target))
-                            dynFlags |= GO_DYNFLAG_LO_ACTIVATE;
+                        // Interaction should only be possible if GO has gossip or is active for quest
+                        if (!ActivateToQuest(target) && !GetGOInfo()->questgiver.gossipID)
+                            dynFlags |= GO_DYNFLAG_LO_NO_INTERACT; // This flag will disable interaction, not setting GO_DYNFLAG_LO_ACTIVATE will not!
                         break;
                     case GAMEOBJECT_TYPE_CHEST:
                     case GAMEOBJECT_TYPE_GOOBER:


tkrokli:
Good idea, I have been somewhat annoyed by those meaningless empty gossip windows. :unamused:

Nelidon:
Thank you.

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -2240,8 +2240,9 @@ void GameObject::BuildValuesUpdate(uint8 updateType, ByteBuffer* data, Player* t
                 switch (GetGoType())
                 {
                     case GAMEOBJECT_TYPE_QUESTGIVER:
-                        if (ActivateToQuest(target))
-                            dynFlags |= GO_DYNFLAG_LO_ACTIVATE;
+                        // Interaction should only be possible if GO has gossip or is active for quest
+                        if (!ActivateToQuest(target) && !GetGOInfo()->questgiver.gossipID)
+                            dynFlags |= GO_DYNFLAG_LO_NO_INTERACT; // This flag will disable interaction, not setting GO_DYNFLAG_LO_ACTIVATE will not!
```
joschiwald: that breaks questgivers with GO_FLAG_INTERACT_COND

Gerhood: I can't find any Questgiver-GOs with that flag in DB

joschiwald: Query: select * from gameobject_template where flags & 4 and type = 2

170 row(s) affected

Nawuko:
Why not something simple like this? 

If has a quest or has a gossip add flag `GO_DYNFLAG_LO_ACTIVATE`
if none of the above add flag `GO_DYNFLAG_LO_NO_INTERACT `

I dont if this is the right behavior. Needs to be in-depth tested.

https://gist.github.com/Nawuko/796ef95dad6109569eab865c6daf6c70

src/server/game/Entities/GameObject/GameObject.cpp
Gerhood: I'm updating my DB

Gerhood:
Fixing the GO_FLAG_INTERACT_COND thing (yet I don't even know how it "breaks" that) should not be a problem. I just didn't know questgiver-GOs with that flag were a thing

Gerhood:
Now I see that the desired effect can also be achieved by simply adding missing GO_FLAG_INTERACT_COND flags:
`UPDATE gameobject_template SET flags = flags | 4 WHERE type = 2 AND Data3 = 0`

PR #17797, opened 2016-08-17
Core/ObjectAccessor: optimize FindPlayerByName

ariel-:
**Changes proposed:**
-  Adds a template specialization for ObjectAccessor::Add/RemoveObject(Player*)
- Saves the (normalized) names in a map, so instead of iterating every player we can do a simple lookup on this new map.

**Target branch(es):** 3.3.5

**Tests performed:** (Does it build, tested in-game, etc) Builds ok

### DIFF
--- a/src/server/game/Globals/ObjectAccessor.cpp
+++ b/src/server/game/Globals/ObjectAccessor.cpp
@@ -81,17 +81,27 @@ HashMapHolder<Player>::MapType const& ObjectAccessor::GetPlayers()
 template class TC_GAME_API HashMapHolder<Player>;
 template class TC_GAME_API HashMapHolder<Transport>;
 
-void PlayerNameMapHolder::Insert(Player* p)
+namespace PlayerNameMapHolder
+{
+typedef std::unordered_map<std::string, Player*> MapType;
+
+MapType& GetContainer()
+{
+    static MapType _objectMap;
+    return _objectMap;
+}
+
+void Insert(Player* p)
 {
     GetContainer()[p->GetName()] = p;
 }
 
-void PlayerNameMapHolder::Remove(Player* p)
+void Remove(Player* p)
 {
     GetContainer().erase(p->GetName());
 }
 
-Player* PlayerNameMapHolder::Find(std::string const& name)
+Player* Find(std::string const& name)
 {
     std::string charName(name);
     if (!normalizePlayerName(charName))
@@ -100,12 +110,7 @@ Player* PlayerNameMapHolder::Find(std::string const& name)
     typename MapType::iterator itr = GetContainer().find(charName);
     return (itr != GetContainer().end()) ? itr->second : nullptr;
 }
-
-PlayerNameMapHolder::MapType& PlayerNameMapHolder::GetContainer()
-{
-    static MapType _objectMap;
-    return _objectMap;
-}
+} // namespace PlayerNameMapHolder
 
 WorldObject* ObjectAccessor::GetWorldObject(WorldObject const& p, ObjectGuid const& guid)
 {
--- a/src/server/game/Globals/ObjectAccessor.cpp
+++ b/src/server/game/Globals/ObjectAccessor.cpp
@@ -101,7 +101,7 @@ Player* PlayerNameMapHolder::Find(std::string const& name)
     return (itr != GetContainer().end()) ? itr->second : nullptr;
 }
 
-auto PlayerNameMapHolder::GetContainer() -> MapType&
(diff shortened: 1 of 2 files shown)


src/server/game/Globals/ObjectAccessor.h
```diff
@@ -65,6 +65,26 @@ class TC_GAME_API HashMapHolder
...
 };
 
+class TC_GAME_API PlayerNameMapHolder
+{
+    //Non instanceable only static
+    PlayerNameMapHolder() { }
+
+public:
+
+    typedef std::map<std::string, Player*> MapType;
```
ariel-: Should this be unordered?

Shauren: Yes

src/server/game/Globals/ObjectAccessor.cpp
```diff
@@ -81,6 +81,41 @@ HashMapHolder<Player>::MapType const& ObjectAccessor::GetPlayers()
...
+}
+
+void PlayerNameMapHolder::Remove(Player* p)
+{
+    boost::unique_lock<boost::shared_mutex> lock(*GetLock());
+
+    GetContainer().erase(p->GetName());
+}
+
+// note that input string shall be normalized by the caller
```
Shauren: I don't like this requirement - you can easily normalize the name inside (pass by value), before acquiring the lock

Shauren:
Is the mutex even needed here? Players can only be added/removed to HashMapHolder during login and logout in threadsafe context of World::Update (single threaded)

ariel-:
Well since HashMapHolder uses a mutex, I thought it was needed here too

src/server/game/Globals/ObjectAccessor.cpp
```diff
@@ -81,6 +81,46 @@ HashMapHolder<Player>::MapType const& ObjectAccessor::GetPlayers()
...
+
+    {
+        boost::shared_lock<boost::shared_mutex> lock(*GetLock());
+
+        typename MapType::iterator itr = GetContainer().find(charName);
+        return (itr != GetContainer().end()) ? itr->second : nullptr;
+    }
+}
+
+auto PlayerNameMapHolder::GetContainer() -> MapType&
```
Shauren: /meh, why use trailing return type? just write MapType& PlayerName...

Shauren:
HashMapHolder has a mutex because it used to be a global storage for ALL object on all maps, not only Player and Transport - Player doesn't need it, not 100% sure on Transport (probably doesn't need it either)

src/server/game/Globals/ObjectAccessor.cpp
ariel-: It's a c++11 custom, to do that I'd have to scope the name MapType.

With auto keyword this is not needed as we already scoped inside PlayerNameMapHolder (return type is writen after :: )

Shauren: Yeah, no. Don't auto.

Shauren:
Now then, PlayerNameMapHolder is not used directly by other code, why not turn it into anonymous namespace living in ObjectAccessor.cpp only?

src/server/game/Globals/ObjectAccessor.cpp
```diff
@@ -81,6 +81,37 @@ HashMapHolder<Player>::MapType const& ObjectAccessor::GetPlayers()
...
+    GetContainer().erase(p->GetName());
+}
+
+Player* Find(std::string const& name)
+{
+    std::string charName(name);
+    if (!normalizePlayerName(charName))
+        return nullptr;
+
+    typename MapType::iterator itr = GetContainer().find(charName);
```
Shauren: good place to use auto (iterator)

src/server/game/Globals/ObjectAccessor.cpp
```diff
@@ -81,6 +81,37 @@ HashMapHolder<Player>::MapType const& ObjectAccessor::GetPlayers()
 template class TC_GAME_API HashMapHolder<Player>;
 template class TC_GAME_API HashMapHolder<Transport>;
 
+namespace PlayerNameMapHolder
+{
+typedef std::unordered_map<std::string, Player*> MapType;
+
+MapType& GetContainer()
+{
+    static MapType _objectMap;
```
Shauren: you can move _objectMap directly into the namespace and remove GetContainer() to avoid magic static initialization (it must be threadsafe so compiler ensure that for you)

ariel-:
:shipit:

Shauren:
no shipping before travis completes its job

PR #17798, opened 2016-08-17
Tools/mmaps_generator: Add percentageDone to console

eilo:
**Changes proposed:**
-  We all know that mmaps extractor takes a lot of time, but never know when it will finish, at least never had an idea unless you did it before, this will help new people not to rush and be patient while console shows general percentage done of the task.
-  Add percentageDone method and two holders m_totalTiles, m_totalTilesBuilt for the purpose.
-  People will do something else while the machine is extracting mmaps instead posting help posts / download mmaps posts.

**Tests performed:** Built under Windows10, VS2015, Corei5, All Built for x64

captura de pantalla 2016-08-17 04 24 01
captura de pantalla 2016-08-17 04 24 08

### DIFF
--- a/src/tools/mmaps_generator/MapBuilder.h
+++ b/src/tools/mmaps_generator/MapBuilder.h
@@ -138,7 +138,7 @@ namespace MMAP
             float m_maxWalkableAngle;
             bool m_bigBaseUnit;
             uint32 m_totalTiles;
-            uint32 m_totalTilesDone;
+            std::atomic<uint32> m_totalTilesDone;
 
             // build performance - not really used for now
             rcContext* m_rcContext;
--- a/src/tools/mmaps_generator/MapBuilder.h
+++ b/src/tools/mmaps_generator/MapBuilder.h
@@ -138,7 +138,7 @@ namespace MMAP
             float m_maxWalkableAngle;
             bool m_bigBaseUnit;
             uint32 m_totalTiles;
-            std::atomic<uint32> m_totalTilesDone;
+            std::atomic<uint32> m_totalTilesBuilt;
 
             // build performance - not really used for now
             rcContext* m_rcContext;
--- a/src/tools/mmaps_generator/MapBuilder.h
+++ b/src/tools/mmaps_generator/MapBuilder.h
@@ -123,6 +123,7 @@ namespace MMAP
             bool shouldSkipMap(uint32 mapID);
             bool isTransportMap(uint32 mapID);
             bool shouldSkipTile(uint32 mapID, uint32 tileX, uint32 tileY);
+            uint32 percentageDone(uint32 totalTiles, uint32 totalTilesDone);
 
             TerrainBuilder* m_terrainBuilder;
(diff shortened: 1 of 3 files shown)


Aokromes:
Have you tested performance before and after change?

eilo:
@Aokromes I'm on it, last one without change in my Surface Pro3 i5 8gbRam (screenshots machine) took about 4-5hours, with the change took similar time.

Im doing it now second try, i will let you know results in a couple of hours :)

Aokromes:
Thx, i ask because when trinitycore removed progress bars startup time improved a lot.

Shauren:
@Aokromes  We removed progress bars because flushing and constantly rewriting **the same line** is slow, adding a simple number into new lines is fine.

However, this PR has issues on its own
* Do not use NULL for integers. Use 0.
* m_totalTilesDone++; is not threadsafe, that code can run from multiple threads at the same time
* It is possible that m_totalTilesDone at the end will not be the same as m_totalTiles (the first one estimates the number of tiles based on number of vmap files)

eilo:
Thanks @Shauren 

- [ ] - Do not use NULL for integers. Use 0. (**done**, also in method percentageDone return 0)
- [ ] - m_totalTilesDone++; is not threadsafe, that code can run from multiple threads at the same time (ill be working on it tomorrow night, same time of today)
- [ ] - It is possible that m_totalTilesDone at the end will not be the same as m_totalTiles (the first one estimates the number of tiles based on number of vmap files) (ill check with threadsafe issue)

src/tools/mmaps_generator/MapBuilder.h
```diff
@@ -136,6 +137,8 @@ namespace MMAP
 
             float m_maxWalkableAngle;
             bool m_bigBaseUnit;
+            uint32 m_totalTiles;
+            uint32 m_totalTilesDone;
```
DDuarte: I guess this could be a `std::atomic_int`

Shauren: Always prefer template std::atomic<T> over std::atomic_T typedef. Not sure about current state in most recent compilers but these used to not be the same

eilo:
@Aokromes @Shauren @DDuarte Please check commit https://github.com/TrinityCore/TrinityCore/pull/17798/commits/25321bcedcc9ebddaf69983ccc17eb9af7d09edb

I think i done the last point on Shauren's punchlist, please let me know if meet all requirements.

DDuarte:
If the PR is ready, remove all the debug code (partially revert last commit).

DDuarte:
LGTM

tkrokli:
@eilo : You no longer need to squash your commits, it is done in the merge process now.

eilo:
@tkrokli thanks for the advise, anyway i will rebase and make one single clean commit to show in commit log after TC staff gives green light to proceed

jackpoz:
The % is broken as described in https://github.com/TrinityCore/TrinityCore/issues/18832

PR #17800, opened 2016-08-17
Pet/DK: Make Risen Ghoul and Bloodworms no longer suicidal

Treeston:
Pretty much as the title says. No more rambo pets.

Current behavior: Ghouls and Bloodworms will aggress any nearby creature, including creatures that aren't currently in combat with the owner. They will even sometimes decide to do so right after being summoned.

New behavior: Ghouls and Bloodworms will never attack a target that isn't already in combat with the DK the guardians are owned by.

### DIFF
--- a/sql/updates/world/3.3.5/9999_99_99_99_BLOODWORMS.sql
+++ b/sql/updates/world/3.3.5/9999_99_99_99_BLOODWORMS.sql
@@ -1,0 +1,2 @@
+-- add AI to bloodworms to make them not stupid and suicidal
+UPDATE `creature_template` SET `ScriptName`='npc_pet_dk_guardian', `AIName`='' WHERE `entry` IN (26125,28017);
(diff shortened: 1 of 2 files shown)


Shauren:
Title says absolutely nothing. 
What is the current behavior? 
What is the new behavior?

Treeston:
Current behavior: Ghouls and Bloodworms will aggress any nearby creature, including creatures that aren't currently in combat with the owner. They will even sometimes decide to do so right after being summoned.

New behavior: Ghouls and Bloodworms will never attack a target that isn't already in combat with the DK the guardians are owned by.

tkrokli:
I can confirm the bloodworm behaviour on current 3.3.5 source (recent rev. tested: 7f5756804ede). They charge any hostile NPC, even if player does not currently aggro any NPC at all. My DK pet does not seem to be quite as aggressive, it does not seem go out of aggro range to attack, like the bloodworms do.

Treeston:
This is about untalented ghoul.

tkrokli:
OK, I can confirm that behaviour of the untalented ghoul too, mostly while doing the DK starter area. Because of that behaviour, I avoided using that one until I had spent enough talents for perma-ghoul.

PR #17813, opened 2016-08-18
Core/Scripts: move npc_saat to DB/SAI

ghost:
**Changes proposed:**

Remove the core script `npc_saat` and use DB data instead.
The core script only contains gossip handling + conditions:
- remove script from file the_black_morass.cpp
- thereby removing 1 line of hardcoded text
- remove redundant SD2 comments in header
- add SAI, gossip and conditions to DB

**Target branch(es):** 3.3.5 and 6.x

**Issues addressed:** No existing open issue related to Sa'at and the quest gossip in this dungeon.
- Updates https://github.com/velinath/TrinityCore/issues/2 , [**Gossip migration from core scripts to SAI/DB ToDo List**](https://github.com/velinath/TrinityCore/issues/2)

- Builds without any issue, core script removal.
- Works as expected in-game, tested with a level 80 character.

- [ ] The conditions for when to show the gossip options seem a bit unclear to me, because the player receives a Chrono-Beacon (item 24289) when accepting quest 10297 The Opening of the Dark Portal and should in theory not be presented with the option to ask for it before the quest is taken, but I am uncertain what the retail behaviour is for this gossip. Maybe @Killyana or @dr-j know more about it?

### DIFF
--- a/sql/updates/world/3.3.5/2016_10_32_17813_world.sql
+++ b/sql/updates/world/3.3.5/2016_10_32_17813_world.sql
@@ -1,0 +1,34 @@
+SET @Saat := 20201;
+SET @Menu := 8088;
+
+UPDATE `creature_template` SET `AIName` = 'SmartAI', `ScriptName`= '' WHERE `entry` = @Saat;
+
+DELETE FROM `smart_scripts` WHERE (source_type = 0 AND entryorguid = @Saat);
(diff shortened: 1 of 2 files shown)


Pixel-Pirate:
good job

Aokromes:
What's missing here?

tkrokli:
Nothing that I know of, when you compare the change to the original script.
Maybe there could be some improvement done to it, but I'm uncertain what that would be.

Killyana:
Conditions are not correct, also there's a missing text.
I will try to fix it.

PR #17814, opened 2016-08-19
Core/Scripts: move forge_of_souls gossip to DB

ghost:
**Changes proposed:**

Forge of Souls: Lady Jaina Proudmoore & Lady Sylvanas Windrunner
- remove 2 lines of hardcoded text from forge_of_souls.cpp
- remove the unused enum `GOSSIP_SPEECHINTRO`
- remove the `OnGossipHello()` hook from both NPC scripts
- replace the `OnGossipSelect()` hook with `sGossipSelect`
- remove the `DoAction` hook because of sGossipSelect
- move the actions into the `OnGossipSelect()` hook
- add Sylvanas' gossip option to the gossip_menu_option table

(gossip option text for Jaina is already in the DB table gossip_menu_option)

**Issues addressed:** Updates https://github.com/velinath/TrinityCore/issues/2 , [**Gossip migration from core scripts to SAI/DB ToDo List**](https://github.com/velinath/TrinityCore/issues/2)

**Tests performed:** (Does it build, tested in-game, etc) Builds without issues. Text modifications only.
Gossip tested in-game, still works as expected.

- [x] Remove `OnGossipHello()` 
- [x] Replace `OnGossipSelect()` with `sGossipSelect`


joschiwald:
the whole ongossiphello stuff should be moved to db

tkrokli:
OK, I thought it would be difficult to change that because of the instance script, but if that won't interfere, I will make those changes. Thanks for the suggestion. :+1:

ccrs:
both scripts can be done with sai afaik

tkrokli:
Very possible, let me push my latest change and then we can check which row of actions to use.

PR #17832, opened 2016-08-20
Core/Scripts: move #define to enum in boss_kalecgos.cpp

ghost:
**Changes proposed:**

Remove hardcoded text from the core script file boss_kalecgos.cpp
- remove hardcoded #define text and use spellcast errors instead
- ( none of the 2 old texts are used anywhere else, likely made up )
- move #define int values to enum

**Issues addressed:** Updates Aokromes' list of hardcoded text to be removed.

**Tests performed:** (Does it build, tested in-game, etc) Builds OK, not tested in-game yet.

Request for comment, feedback & suggestions for what to use instead of the guessed text messages.
The only thing I have done is to insert suggestions for using already defined error messages.

- [x] Need help & feedback on which method to use for correct error handling

### DIFF
--- a/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kalecgos.cpp
+++ b/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kalecgos.cpp
@@ -79,22 +79,21 @@ enum SWPActions
     DO_BANISH                                   =  2
 };
 
-#define GO_FAILED   "You are unable to use this currently."
+enum Misc
+{
+    FLY_X                         = 1679,
+    FLY_Y                         = 900,
+    FLY_Z                         = 82,
+    CENTER_X                      = 1705,
+    CENTER_Y                      = 930,
+    RADIUS                        = 30,
+    MAX_PLAYERS_IN_SPECTRAL_REALM = 0 // over this, teleport object won't work, 0 disables check
+};
 
 #define EMOTE_UNABLE_TO_FIND    "is unable to find Kalecgos"
 
-#define FLY_X           1679
-#define FLY_Y           900
-#define FLY_Z           82
-
-#define CENTER_X        1705
-#define CENTER_Y        930
-#define RADIUS          30
-
 #define DRAGON_REALM_Z  53.079f
 #define DEMON_REALM_Z   -74.558f
-
-#define MAX_PLAYERS_IN_SPECTRAL_REALM 0 //over this, teleport object won't work, 0 disables check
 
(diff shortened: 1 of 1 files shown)


src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kalecgos.cpp
```diff
@@ -760,9 +758,8 @@ class boss_sathrovarr : public CreatureScript
                     }
                     else
                     {
-                        me->TextEmote(EMOTE_UNABLE_TO_FIND);
                         EnterEvadeMode();
-                        return;
+                        return SPELL_FAILED_TARGET_NOT_IN_INSTANCE;
```
Sarratas: You cant return anything here, UpdateAI is a void function. Thats why your pull request doesnt compile. Leave it as it was before.

src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kalecgos.cpp
```diff
@@ -562,8 +561,7 @@ class kalecgos_teleporter : public GameObjectScript
 
         if (player->HasAura(AURA_SPECTRAL_EXHAUSTION) || SpectralPlayers >= MAX_PLAYERS_IN_SPECTRAL_REALM)
         {
-            player->GetSession()->SendNotification(GO_FAILED);
-            return true;
+            return SPELL_FAILED_CANT_DO_THAT_RIGHT_NOW;
```
Sarratas: OnGossipHello returns boolean value, true or false. It doesnt make sense to return anything else. Dont change it, it was just ok.

tkrokli: Got a suggestion for where to find the correct error message?
I need the correct error message, otherwise this PR has no meaning.

Shauren: There is no text on retail.

tkrokli: Thanks. I removed the message altogether. Let's see if the result is acceptable.

Aokromes:
What's missing here?

tkrokli:
Nothing that I know of now. I got feedback from Shauren that the error messages were incorrect, so I removed those. The rest is moved into enum (except for the float values). Goal achieved, anyway.

PR #17843, opened 2016-08-22
[6.x] Core/ObjectMgr: Load correct Player Create XP Data

### LINKED ISSUE #17822: wrong experience
1 level should be 400, error 12000
starting with level 2, a strip of smooth experience level 1 = 400
wowscrnshot_081916_204139
wowscrnshot_081916_204123
7.0.3
https://github.com/TrinityCore/TrinityCore/commit/650ab693a8366847d7db8bd8443b3b4aaf99ff85

funjoker:
**Target branch(es):** 6.x

**Issues addressed:** Closes #17822 

- Check other GT relevant stuff (Mitigation is working correct)

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -3559,7 +3559,7 @@ void ObjectMgr::LoadPlayerInfo()
     {
         uint32 oldMSTime = getMSTime();
 
-        _playerXPperLevel.resize(sXpGameTable.GetTableRowCount() + 1, 0);
+        _playerXPperLevel.resize(sXpGameTable.GetTableRowCount() - 1, 0);
 
         //                                               0      1
         QueryResult result = WorldDatabase.Query("SELECT Level, Experience FROM player_xp_for_level");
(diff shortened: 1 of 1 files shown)


funjoker:
Not sure if it was a unqiue failure or if it's a GT loading problem.

If it's a GT loading problem this should not be merged.

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -3559,14 +3559,14 @@ void ObjectMgr::LoadPlayerInfo()
     {
         uint32 oldMSTime = getMSTime();
 
-        _playerXPperLevel.resize(sXpGameTable.GetTableRowCount() + 1, 0);
+        _playerXPperLevel.resize(sXpGameTable.GetTableRowCount() - 1, 0);
```
Shauren: this change is wrong, you will access `_playerXPperLevel[sXpGameTable.GetTableRowCount()]`

funjoker: Tested it. All Level xp were loaded.
The Table has 124 rows.

Row 0 is Header stuff.
So 123 rows with data.

Getting the row count and removing the Header will still leave 123 rows. Starting loading from Level 1 Not 0 (which doesnt exist) will cause full loading. 

But i can undo this Part

Shauren: sure, if the vector has allocated space for more rows. but you are still going out of bounds

PR #17847, opened 2016-08-23
Core/Chat: removed useless code

FrancescoBorzi:
**Changes proposed:**
-  removed useless code

it builds and executes, I've tested running commands like `.server info`,  `.serv inf` and that "match" variable is never set to true (both commands work)


Shauren:
Try `.ser i`

Shauren:
This code is supposed to handle a situation where one command full name is a valid abbreviation for another command
Example:
`
.additemset
.additem
`
Without this part, if `.additemset` appeared before `.additem` in the command table, it would be impossible to execute `.additem` command

ShinDarth:
@Shauren thanks for the explanation

it would be better to have some comments in those lines of code to make it clear

PR #17849, opened 2016-08-23
Core/Chat: add some safety checks to prevent crashes + add comment

FrancescoBorzi:
**Changes proposed:**
-  added some safety checks, preventing crashes
-  added a comment

**Tests performed:** it builds and runs


Shauren:
These checks are useless, Name can never be null. See https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Chat/Chat.h#L45

Issue #14120, opened 2015-02-12
Pet Pathing & Pet LoS


ccrs:
I couldn't find any other issue regarding this.

I suppose this is a well known issue/exploit but I'll explain it briefly:
- Any Pet which can be commanded to attack (ie: any warlock pet, hunter pet, frost mage pet,... etc) will go through walls and solid gobjects, mainly doors, to attack its objective.

This offers exploit opportunities to players such as starting combat with locked bosses and avoid doing the whole instance (ie Ymiron at Utgarde Pinnacle).

hash: c5dac48436c8a00d40640af9aecaaa71d549135a
Using mmaps.
##

Runico:
confirm

jackpoz:
Walls or doors ? Because doors are ingame gameobjects spawned at runtime, walls are already present in client files and handled by MMAPs

Aokromes:
doors.

jackpoz:
then it's a know old issue, mmaps don't take into account any dynamically spawned gameobject

ccrs:
well I just wanted to have it on an open issue ^^

jackpoz:
it has been a known issue for 2 years at least and it's going to be like that for 2 more years too

ccrs:
nice attitude ^^

jackpoz:
It's not an attitude, it's the simple consequence of the limitations of recastnavigation. We already worked on this issue months ago, people before us tried too but no steps forward have been made (and I'm actually the only dev who fixed an issue in recast and got his PR accepted).
You could study how recastnavigation works and see with your own eyes what l'm talking about.

jackpoz:
If anyone would like to mess with recast, here's an idea https://github.com/memononen/recastnavigation/issues/56 

>we'll have to get the bounding box of the go, change the mesh layout to create a new polygon with the size of the bounding box, when the go is spawned at runtime we set the flags as "don't cross this" and when the go is despawned(or the door is opened) we reset the flags

jackpoz:
> it has been a known issue for 2 years at least and it's going to be like that for 2 more years too

1 year passed, 1 left to go

Aokromes:
Not high anymore, with the new boundaries system from Treeston you can't pull ingvar anymore.

jackpoz:
> it has been a known issue for 2 years at least and it's going to be like that for 2 more years too

2 years passed! ding!

chaodhib:
[...]

PR #17850, opened 2016-08-23
[WIP] Core/Movement: Targeted MoveGen Improvements

### LINKED ISSUE #9206: [MMaps] Exploit (easy kills)
Basically with MMaps, on any server I've seen with it including TrinityCores latest implementation of the system, there's an exploit where you can fly above the creatures range. If the creature does not have flying, he will stand idly at your feet while you cast spells on him.

Pretty big exploit if you don't have decent anti-cheat as it can be used in dungeons and raids to kill bosses solo.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/112567-mmaps-exploit-easy-kills?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #16549: [3.3.5][6.x] Core/Creature: Missing Z Position update in Combat

### LINKED ISSUE #16647: [3.3.5] Core/Spell: Pets and LoS, spell cast fail
**Description**: Pets dont use spells on target outside of line of sight. Pet should cache spell casted and use again when pet is close to target.
**Current behaviour**: After using spells at targets outside the pet's line of sight, the spell fails.
**Expected behaviour**: The correct is that the pet chase the target until he can use the spell.
Little Vid : https://www.youtube.com/watch?v=IIxiusAehl0 (Video provided by @Goatform )
https://youtu.be/fgqmpbg8Cds (On Retail)

1. Create one Warlock Pet (succubus ou any other)
2. Put warlock on circle of blood (.go xyz 2825 5911 3 530 )
3. Put another character out of LoS (.go xyz 2831 5920 3 530)
4. Try use any spell of warlock pet on another character

**Branch(es)**: 335/6x

**TC hash/commit**: b872565fb6841efccc0fd0374c380c3ad4104278

**TDB version**: 335.60

**Operating system**: Win10

chaodhib:
# Work in progress PR
## **Planned changes:**

Major changes:
- Changing how the way the destination point of each NPC path is determined. Up until now, each time a creature had to move, first a destination point was computed. And only then a path leading to that point was generated. Now it's the other way around. The destination point is actually deduced from the path itself. We first generate a path to the target's location and then remove from the path a certain length: the distance at which the npc should stand.

In the only commit yet, this is already implemented but only for MoveChase. Also, ReducePathLenghtByDist needs to be improved in order to make sure that after that method call, the last point of the path is always in LoS of the target & at the right distance.

Minor changes:
-  From sniff analysis, I found that the standard distance at which NPCs stand after chasing another unit is 'Chaser->CombatReach + Target->CombatReach' instead of 'Chaser->CombatReach + Target->CombatReach + 0.5'.
- From sniff analysis, I believe pets follow their masters at a 3 constant yard distance.
-  Planning to remove the third parameter of MoveChase (the angle).

**Target branch(es):** 3.3.5

**Issues addressed:** Seems to at least 
fix #16549
fix #9206 
fix #16647

**Tests performed:** WIP. Build and tested IG for what I have so far.

### DIFF
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -132,7 +132,9 @@ void TargetedMovementGenerator<T, D>::SetTargetLocation(T* owner, bool updateDes
             if (GetTarget()->IsWithinDistInMap(owner, CONTACT_DISTANCE))
                 return;
 
-            GetTarget()->GetContactPoint(owner, x, y, z);
+            x = GetTarget()->GetPositionX();
+            y = GetTarget()->GetPositionY();
+            z = GetTarget()->GetPositionZ();
         }
         else
         {
@@ -167,6 +169,16 @@ void TargetedMovementGenerator<T, D>::SetTargetLocation(T* owner, bool updateDes
     bool forceDest = (owner->GetTypeId() == TYPEID_UNIT && owner->ToCreature()->IsPet() && owner->HasUnitState(UNIT_STATE_FOLLOW));
 
     bool result = _path->CalculatePath(x, y, z, forceDest);
+
+    float distFromTarget;
+    if(!_offset)
+        distFromTarget = owner->GetCombatReach() + GetTarget()->GetCombatReach() + CONTACT_DISTANCE;
+    else if(owner->IsPet() && GetTarget()->GetTypeId() == TYPEID_PLAYER) // special case for offset != 0 and pets. cf above
+        distFromTarget = 1.0f + GetTarget()->GetCombatReach() + _offset;
+    else
+        distFromTarget = owner->GetCombatReach() + GetTarget()->GetCombatReach() + _offset;
+
+    _path->ReducePathLenghtByDist(distFromTarget, GetTarget()->ToUnit());
     if (!result || (_path->GetPathType() & PATHFIND_NOPATH))
     {
         // can't reach target
--- a/src/server/game/Movement/PathGenerator.cpp
+++ b/src/server/game/Movement/PathGenerator.cpp
@@ -904,7 +904,7 @@ float PathGenerator::Dist3DSqr(G3D::Vector3 const& p1, G3D::Vector3 const& p2) c
     return (p1 - p2).squaredLength();
 }
 
-void PathGenerator::ReducePathLenghtByDist(float dist)
+void PathGenerator::ReducePathLenghtByDist(const float dist, Unit* target)
 {
     if (GetPathType() == PATHFIND_BLANK)
     {
@@ -915,35 +915,121 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
     if (_pathPoints.size() < 2) // path building failure
         return;
 
-    uint32 i = _pathPoints.size();
-    G3D::Vector3 nextVec = _pathPoints[--i];
-    while (i > 0)
+    /*
+    evolution of idx:
+     x---->x---->x---->X
+     0     1     2     3
+                idx   idx+1
+          idx  idx+1
+    idx  idx+1
+    ---------------------------
+    diffVec when idx==1 in example above:
+            ---->
+    */
+    int idx = _pathPoints.size() - 2;
+    float totalLengthSoFar = 0.0f;
+    G3D::Vector3 diffVec;
+    for ( ; idx >= 0; idx--)
     {
-        G3D::Vector3 currVec = _pathPoints[--i];
-        G3D::Vector3 diffVec = (nextVec - currVec);
-        float len = diffVec.length();
-        if (len > dist)
-        {
-            float step = dist / len;
-            // same as nextVec
-            _pathPoints[i + 1] -= diffVec * step;
-            _sourceUnit->UpdateAllowedPositionZ(_pathPoints[i + 1].x, _pathPoints[i + 1].y, _pathPoints[i + 1].z);
-            _pathPoints.resize(i + 2);
+        diffVec = _pathPoints[idx + 1] - _pathPoints[idx];
+        totalLengthSoFar += diffVec.length();
+        if (totalLengthSoFar >= dist)
             break;
+    }
+
+    if (totalLengthSoFar > dist)
+    {
+        float delta = totalLengthSoFar - dist;
+        G3D::Vector3 newFinalPoint = _pathPoints[idx] + diffVec.unit() * delta;
+        _sourceUnit->UpdateAllowedPositionZ(newFinalPoint.x, newFinalPoint.y, newFinalPoint.z);
+
+        /*
+        3 potential problems could have occured up until this point:
+        1) the "actual end position" (closest position to the target found by the PathGen) could be not exactly on the target.
+        Which means the distance from the target to the last point of the path BEFORE THE CUT could be > 0
+        and add into that the cutted distance and this can result in the target being further away than dist to the 
+        path's end position AFTER CUT.
+        2) AFTER THE CUT, the Z of the new last point is normalized 
+        (to make sure our point is on the ground in case non flying/swimming unit).
+        This means the distance from the target to the path's last point AFTER CUT can be higher than dist.
+        3) The path's last point AFTER CUT can be out of LoS from the target.
+
+        We need to make sure that in ANY case: 
+        a) the last point after cut is at MAX at "dist" yards from the target + marginal error factor (0.5f)
+        b) the last point after cut is in LoS of the target
+
+        In order to do that, we iterate on the points of the path starting by the point on idx+1 and we cut the path at the first
+        point that is both in LoS and in range.
+        Worst case scenario, we use the last position, which should be close (if not equal) to the target position.
+        */
+        if (IsValidFinalPoint(newFinalPoint, target, dist))
+        {
+            _pathPoints[idx + 1] = newFinalPoint;
+            _pathPoints.resize(idx + 2);
         }
-        else if (i == 0) // at second point
+        else
+        {
+            CutAtFirstValidPoint(idx + 1, target, dist);
+        }
+    }
+    else if (totalLengthSoFar == dist && idx > 0)
+    {
+        if (IsValidFinalPoint(_pathPoints[idx], target, dist))
+        {
+            _pathPoints.resize(idx + 1);
+        }
+        else
+        {
+            CutAtFirstValidPoint(idx + 1, target, dist);
+        }
+    }
+    else
+    {
+        // if initial position is "valid", no need to move
+        if (IsValidFinalPoint(_pathPoints[0], target, dist))
         {
             _pathPoints[1] = _pathPoints[0];
             _pathPoints.resize(2);
-            break;
         }
-
-        dist -= len;
-        nextVec = currVec; // we're going backwards
+        else
+        {
+            CutAtFirstValidPoint(1, target, dist);
+        }
     }
 }
 
 bool PathGenerator::IsInvalidDestinationZ(Unit const* target) const
 {
     return (target->GetPositionZ() - GetActualEndPosition().z) > 5.0f;
+}
+
+bool PathGenerator::IsValidFinalPoint(const G3D::Vector3 finalPoint, Unit* target, float dist) const
+{
+    return target->GetExactDist(finalPoint.x, finalPoint.y, finalPoint.z) <= dist + 0.5f
+        && target->IsWithinLOS(finalPoint.x, finalPoint.y, finalPoint.z);
+}
+
+uint32 PathGenerator::SearchForFirstValidPoint(int startAtIndex, Unit* target, float dist) const
+{
+    uint32 index = startAtIndex;
+    for ( ; index < _pathPoints.size() ; index++)
+    {
+        if (IsValidFinalPoint(_pathPoints[index], target, dist))
+            break;
+    }
+    return index;
+}
+
+void PathGenerator::CutAtFirstValidPoint(int startAtIndex, Unit* target, float dist)
+{
+    // directly test the last point in order to avoid iterating over all points in case of unreachable target
+    if (!IsValidFinalPoint(_pathPoints[_pathPoints.size() - 1], target, dist))
+    {
+        // even the last position of the generated path, which should be close if not on top of the target, was not good enough. we mark the path as unusable.
+        _type = PATHFIND_NOPATH;
+        return;
+    }
+
+    uint32 indexToCutAt = SearchForFirstValidPoint(startAtIndex, target, dist); // should never return "_pathPoints.size" because of the check done above
+    _pathPoints.resize(indexToCutAt + 1);
 }
--- a/src/server/game/Movement/PathGenerator.h
+++ b/src/server/game/Movement/PathGenerator.h
@@ -74,7 +74,12 @@ class TC_GAME_API PathGenerator
 
         PathType GetPathType() const { return _type; }
 
-        void ReducePathLenghtByDist(float dist); // path must be already built
+        /*
+        Reduce the total length of the path by removing excess waypoints and cuting at the right distance
+        while making sure that the last point is in LoS with the target. If by the first cut, the new last point
+        isn't in LoS with the target, the cutted length will be smaller than "dist".
+        */
+        void ReducePathLenghtByDist(float dist, Unit* target); // path must be already built
 
     private:
 
@@ -133,6 +138,10 @@ class TC_GAME_API PathGenerator
         dtStatus FindSmoothPath(float const* startPos, float const* endPos,
                               dtPolyRef const* polyPath, uint32 polyPathSize,
                               float* smoothPath, int* smoothPathSize, uint32 smoothPathMaxSize);
+
+        bool IsValidFinalPoint(const G3D::Vector3 finalPoint, Unit* target, float dist) const;
+        uint32 SearchForFirstValidPoint(int startAtIndex, Unit* target, float dist) const;
+        void CutAtFirstValidPoint(int startAtIndex, Unit* target, float dist);
 };
 
 #endif
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -5318,7 +5318,7 @@ SpellCastResult Spell::CheckCast(bool strict, uint32* param1 /*= nullptr*/, uint
                     else if (m_preGeneratedPath->IsInvalidDestinationZ(target)) // Check position z, if in a straight line
                             return SPELL_FAILED_NOPATH;
 
-                    m_preGeneratedPath->ReducePathLenghtByDist(objSize); // move back
+                    m_preGeneratedPath->ReducePathLenghtByDist(objSize, target); // move back
                 }
                 break;
             }


Treeston:
This logic is _really_ interesting. @r00ty-tc I know you'll appreciate.

Aokromes:
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Movement/PathGenerator.cpp:933:17: fatal error: 
      comparison of unsigned expression >= 0 is always true
      [-Wtautological-compare]
    for ( ; idx >= 0; idx--)
            ~~~ ^  ~

Treeston:
Naughty infinite loop.

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -915,31 +915,45 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
...
+          idx  idx+1
+    idx  idx+1
+    ---------------------------
+    diffVec when idx==1 in example above:
+            ---->
+    */
+    uint32 idx = _pathPoints.size() - 2;
+    float totalLengthSoFar = 0.0f;
+    G3D::Vector3 diffVec;
+    uint32 newPathSize;
```
tkrokli:
```
C:\projects\trinitycore\src\server\game\Movement\PathGenerator.cpp(932): error C2220: warning treated as error
 - no 'object' file generated [C:\projects\trinitycore\build\src\server\game\game.vcxproj]
C:\projects\trinitycore\src\server\game\Movement\PathGenerator.cpp(932): warning C4101: 'newPathSize':
unreferenced local variable [C:\projects\trinitycore\build\src\server\game\game.vcxproj]
```

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -915,31 +915,45 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
...
+    for ( ; idx >= 0; idx--)
```
tkrokli:
```
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Movement/PathGenerator.cpp:933:17: fatal error: 
      comparison of unsigned expression >= 0 is always true
      [-Wtautological-compare]
    for ( ; idx >= 0; idx--)
            ~~~ ^  ~
```

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -926,10 +926,9 @@ void PathGenerator::ReducePathLenghtByDist(const float dist, Unit* target)
     diffVec when idx==1 in example above:
             ---->
     */
-    uint32 idx = _pathPoints.size() - 2;
+    int idx = _pathPoints.size() - 2;
```
Treeston: Ew. Stick with `uint32` and just change your loop.

```
while (idx)
{
    --idx;
    /* do stuff */
```

chaodhib:
First version ready for testing. Only MoveChase without offset has been implemented. The rest is broken for now.

There is probably a lot more work to do but for now it looks stable and it should fix A LOT of movement related issues. At least the first 3 issues exposed [here](https://github.com/TrinityCore/TrinityCore/pull/17787#issuecomment-240851396). Maybe more.

src/server/game/Movement/PathGenerator.cpp
chaodhib: idx can be equal to -1 at some point. unsigned int would not be enough.

chaodhib:
@Treeston @r00ty-tc @jackpoz could you let me know what do you think about this PR when you find time plz?

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -106,6 +107,7 @@ void TargetedMovementGeneratorMedium<T, D>::_setTargetLocation(T* owner, bool up
         && owner->HasUnitState(UNIT_STATE_FOLLOW));
 
     bool result = i_path->CalculatePath(x, y, z, forceDest);
+    i_path->ReducePathLenghtByDist(owner->GetObjectSize() + i_target->GetObjectSize() + CONTACT_DISTANCE, i_target->ToUnit()); // this distance is correct only for MoveChase with no offset
```
jackpoz: there is little point reducing the path length before checking the result in the lines below (also it's length, not lenght)

jackpoz:
could you describe more in detail how pathfinding calls are handled in this PR, in particular the "Now it's the other way around" part ? Does this change very little how and when pathfinding is used or does it change some things we should check (i.e. more/less queries, more/less detailed queries, etc).

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -904,7 +904,7 @@ float PathGenerator::Dist3DSqr(G3D::Vector3 const& p1, G3D::Vector3 const& p2) c
     return (p1 - p2).squaredLength();
 }
 
-void PathGenerator::ReducePathLenghtByDist(float dist)
+void PathGenerator::ReducePathLenghtByDist(const float dist, Unit* target)
```
jackpoz: It would be nice if target wouldn't be mandatory (right now the code crashes with NULL dereference exception of the caller passes NULL since there's no validation at all), PathGenerator is not supposed to always have a Unit as target.

Treeston:
Current logic:
1. Take a straight line from self to target
2. Shorten this line by `offset`
3. Take the resulting point, and generate a path to it

New logic:
1. Generate a path from self to target
2. Shorten this path by `offset`

jackpoz:
> Should fix quite a lot of issues related to NPC movements

could you provide test cases with previous and new behavior ?

Treeston:
I haven't tested, but this should fix any issue where NPCs would path to just below you, then be unable to attack you due to range.

Ex.: Central storm peaks, stand on the rocks and pull a Magnataur.

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -915,35 +915,121 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
...
+    return index;
+}
+
+void PathGenerator::CutAtFirstValidPoint(int startAtIndex, Unit* target, float dist)
+{
+    // directly test the last point in order to avoid iterating over all points in case of unreachable target
+    if (!IsValidFinalPoint(_pathPoints[_pathPoints.size() - 1], target, dist))
+    {
+        // even the last position of the generated path, which should be close if not on top of the target, was not good enough. we mark the path as unusable.
+        _type = PATHFIND_NOPATH;
```
jackpoz: this code doesn't handle PATHFIND_SHORT case (and it's still to be defined after which distance it should be set to PATHFIND_SHORT)

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -915,35 +915,121 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
...
 }
 
 bool PathGenerator::IsInvalidDestinationZ(Unit const* target) const
 {
     return (target->GetPositionZ() - GetActualEndPosition().z) > 5.0f;
 }
+
+bool PathGenerator::IsValidFinalPoint(const G3D::Vector3 finalPoint, Unit* target, float dist) const
+{
+    return target->GetExactDist(finalPoint.x, finalPoint.y, finalPoint.z) <= dist + 0.5f
```
jackpoz: do you need the exact dist or is the squared distance comparison good too ?

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -915,35 +915,121 @@ void PathGenerator::ReducePathLenghtByDist(float dist)
...
+        Which means the distance from the target to the last point of the path BEFORE THE CUT could be > 0
+        and add into that the cutted distance and this can result in the target being further away than dist to the 
+        path's end position AFTER CUT.
+        2) AFTER THE CUT, the Z of the new last point is normalized 
+        (to make sure our point is on the ground in case non flying/swimming unit).
+        This means the distance from the target to the path's last point AFTER CUT can be higher than dist.
+        3) The path's last point AFTER CUT can be out of LoS from the target.
+
+        We need to make sure that in ANY case: 
+        a) the last point after cut is at MAX at "dist" yards from the target + marginal error factor (0.5f)
```
jackpoz: actually you have not guarantee that a valid path could ever be at MAX at dist, i.e. PATHFIND_INCOMPLETE case which by definition doesn't reach the destination

jackpoz:
> 1. Take a straight line from self to target

isn't the current 1st step "find a point close to target" https://github.com/TrinityCore/TrinityCore/pull/17850/files#diff-268a0e4eced8ee7bb5139f6c2b47034cL61 ?
what's the difference between "find a point close to target then walk to the point" to "walk to target then find a point close to target" ? Just the angle between where the unit is walking from, the target and when the unit will stand when it stops ?

Treeston:
Example: Target is standing on a hill, with a path leading up the side.

**Current behavior:**
now

**New behavior:**
then

jackpoz:
2nd SS looks better, I think a 3rd SS with both "try different angles and try to shorten the path" will get the best of both.
Keep in mind in the worst case the end point of the path is not in LoS with the target and reducing the path might not always give a point in LoS (in corridor/wall with enemy being the end of the wall but close to the angle)

chaodhib:
There is a fundamental issue with trying different angles is that it's asymmetrical by nature. By which direction should we start searching? By the right or by the left? For each choice, some situations will be handled awkwardly.

The map Blade Edge Arena is a great example of this problem in the current version of the core:
[current core](https://we.tl/R5fQGBxrhL)
[with my changes](https://we.tl/DNJnfCEsJv)

here is a representation (I wanted to do join in the Paint party):
npc pathing example

src/server/game/Movement/PathGenerator.cpp
chaodhib: this method is called for now only at 2 points and for each of those cases, if i'm not mistaken, unit will never be null.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
chaodhib: ReducePathLenghtByDist also valide the last point of the path and if no right point is found in range and in los of the target, the path is set to flagged as unusable. This is why. (regarding the spelling, i didn't create the method. i just rewrote it, but yeah i will fix that)

src/server/game/Movement/PathGenerator.cpp
chaodhib: this method should not be called when dealing with such path. my mistake if it's the case right now.

src/server/game/Movement/PathGenerator.cpp
chaodhib: to add the "acceptable error" of 0.5f, i need the exact

jackpoz:
what if the target is very close to the edge and the cut path will be from the left side of the edge and so not in LoS ? I think a mix of the 2 solutions is required to find a good place in as many cases as possible

chaodhib:
Getting closer and closer to the target until a point that satisfy both the los and the distance is found. That works too and has the advantage of being consistent and predictable. I'm really not fan of that 2D circle solution because we haven't even touched the issue of 3D yet (cf https://github.com/TrinityCore/TrinityCore/issues/16549)

chaodhib:
Getting closer and closer to the target until a point that satisfy both the los and the distance is found, that works too and it has the advantage of being consistent and predictable. I really dont think we should keep the 2D circle solution because of the issues related to it and we haven't even touched the 3D aspect yet (cf https://github.com/TrinityCore/TrinityCore/issues/16549).

jackpoz:
how do you handle cases where the target is not in the walkable mesh ? how do you choose the direction to get closer to the target in that case ? i.e. between a wall and a statue that is not marked as walkable by recast.

Treeston:
If the target isn't walkable then pathing will fail.

jackpoz:
> If the target isn't walkable then pathing will fail.

Well the current code handles that case, if you think the new one doesn't then it has to support it.

Treeston:
It shouldn't, though. If the current code does, that's a bug.

If you are standing in a position that a mob can't reach, then it _should_ no-path evade.

Treeston:
For example: Sunken Temple (pre-cata), the green drake bosses had a knockback. If you tanked them against the wall in some spots and got knocked onto a ledge a few feet in the air, they'd soft evade on you, even though they'd still easily be able to hit you if they walked up to you.

jackpoz:
The mob doesn't have to reach the exact x,y,z position of the target, it's enough to get in combat reach. In fact the circle code was added to handle a case of player behind a statue/pillar that is walkable by players but not by npcs, all the mob has to do is to move a bit to be in LoS and attack the player that way.

Another unrelated case: how does this PR handle targets behind dynamically spawned gameobjects (ignored by recast) ?

src/server/game/Movement/PathGenerator.cpp
jackpoz: are you sure you can't use the squared distance just because of 0.5f ?

Treeston: Square the RHS, probably cheaper.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
jackpoz: there's no point to set the path as unusable (PATHFIND_NOPATH) if the path is already unusable, so if i_path->GetPathType() & PATHFIND_NOPATH is already true before this call.

src/server/game/Movement/PathGenerator.cpp
jackpoz: if it will never be null then add an ASSERT( )

src/server/game/Movement/PathGenerator.cpp
jackpoz: @Treeston: sometimes it's better to give time to the users to let them find the solution and learn than giving them the solution.

Aokromes:
Sorry, but https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-243159611 is incorrect, i have multiple points on retail where mobs takes short path.

Aokromes:
You was allowed to hide on the small hole on 2nd screenshot and mob was unable to reach you.
http://i.imgur.com/6KCqbKM.jpg
http://i.imgur.com/YmfrVUS.jpg

src/server/game/Movement/PathGenerator.cpp
chaodhib: either way, thank you

Raydor:
This fixes https://github.com/TrinityCore/TrinityCore/issues/16647#issuecomment-186828967, confirmed.

akrom23:
Any news on this? What is needed to merge?

Treeston:
@akrom23 most things, e.g. it actually working for movechase with offset and whatnot.

chaodhib:
I'm away from home atm. I will get back on this in a few weeks.

akrom23:
@chaodhib: Any news?

chaodhib:
I could work again on this but I need a clearer overview of the weaknesses of this new path finding method so I can find a solution for them or, if i can't, drop this PR.

In my opinion, this method that I'm proposing, without being perfect, is stronger than what is currently implemented in TC. However, I'm really confused wherever or not this will get merged if I finish it.

jackpoz:
@chaodhib what's your plan about handling https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-243519503 ?

chaodhib:
@jackpoz Could I get as much detail as possible regarding this particular case? Where is it? How to reproduce? What is the behavior in retail? What will be the behavior with this PR?

I noticed that in retail server, NPCs could cross very high slopes but they are unable to get on top of a cube standing on the floor, even if the height of the cube is 0.5 yards.

jackpoz:
it's the same described from https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-243252027 and below, @Aokromes can tell more details about it if needed (which is the same I asked back then)

Eliminationzx:
Any news?
@chaodhib , can you check why pets falling under textures when target on transport?

Aokromes:
that's mmaps thing.

Aokromes:
Conflicting files
src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp

chaodhib:
@Aokromes regarding https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-243519503, if I'm understanding you correctly, in retail, a player could go in that corner and mobs would stop chasing me, is that correct? If so, that was true for which mob? The boss in the next room? Or one of his humanoid adds?

Aokromes:
No, that bug exists (or used to exist) on tc, idk on retail.

chaodhib:
I'm not following you then. What is the bug and what is the expected behavior? If you don't know wherever or not this behavior exists in retail, how can you prove that it's a bug?

Aokromes:
well, having ability to hide there of mobs and be unkillable looks a big bug for me :P (also i wrote it there because jackpoz asked) xd

chaodhib:
Oh ok. But this bug exists even in the main branch (3.3.5), right? It's not specific to this PR, is it?

Aokromes:
Yes, at least used to exist on the far past xd

jackpoz:
> But this bug exists even in the main branch (3.3.5), right?

No it doesn't, https://github.com/TrinityCore/TrinityCore/commit/8927a042536d4671e94ad901e9209efcf78ffa2c fixed the issue, it was explained in https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-243168276 too

JunkyBulgaria:
http://prntscr.com/h3hzza
npcs with bigger hitbox gets stuck

example: sholazar, x: 5539, y:4949, z: -136

chaodhib:
Resuming work on this.

So, I tested @Aokromes's [example ](https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-243519503) (the small hole at the entrance of Blackwing Lair): NPCs are able to reach the player both in 3.3.5 and in my branch.

I will look at the remaining issues and will implement the remaining of the functionalities needed for merge this.

Also, the PR's description has been edited to reflect that this PR should fix at least 3 different issues.

chaodhib:
Let's be clear about one thing:

This PR will fix a lot of issues related to pathing but it is far from perfect. It merely replaces the way we use Detour to do path finding. And as such, we are still bound to the limitations of Detour to our use case. This PR gets rid of the usage of the "test multiple angles around the target until a suitable point is found" method (let's call it the "Angle method" from now on). Instead, a more predictable and consistent method is used. This method will be better in 99% of the cases. However, it will notably fail to take the shortest path in some situation (the one described by @jackpoz in the first paragraph [HERE](https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-243398938)). 

To that, there are 2 things to note:

- The Angle method can also sometimes fail to find the shortest path (since a) it is asymmetrical. it will start by looking at one side before the other. b) it starts by the angle between the target and the mover. thus sometimes not picking the right point).
- A proper solution to this situation is not the Angle method. It is to modify Detour itself in order to adapt it to our use case. Which is significantly harder to do of course. And this is what Blizzard does. By that, I mean that they use their own path finding system and are not limited by a pathfinding library such as Detour. My point being, we can't achieve Blizz-like path-finding unless we fork and modify Detour to adapt it to our particular use case.

In conclusion, this PR uses a method that is better in 99% of the cases than the current one. And the only way to do better is by modifying the path finding library that we use, which is significantly more work.

Treeston:
Can you briefly describe the "new method" in relatively simple terms?

chaodhib:
In simple terms, in order to attack its target, a NPC needs to move closer to its target until the distance between itself and its target is below a known value (the combat reach, see note 1 below). So, what we do now is:

1. Find a walkable path from the point A (current position of the mover) to point B (position of the target), using Detour. It's worth repeating that here, we are only thinking in terms of 2 points and a path between those two points. No notion of unit with a certain combat reach or anything.
step 1


If we were to stop here and simply ask the NPC to follow this path, then at the end of the path, the NPC will be exactly on its target position. This is obviously overkill because the NPC doesn't need to get as close. Each NPC has a certain a certain range. Getting close **enough** to the target is what we need.

2. In this step, we simply "cut" from the path an extra distance: the combat reach. We can cut this extra distance because the NPC does not need to travel further in order to attack its target from there.

step 2

After that, we are done. The location pointed by the orange arrow will be our final position after moving. And the orange arrow will be our path. The final position satisfy our problem: a) we moved the least amount of distance. b) we are now close enough to our target in order to attack it.


Note:
1. I made a simplification regarding what i called the combat reach. Its value actually depends on the characteristics of both the mover and the target. The formula is in the code somewhere (and uses a field actually called "combat reach"). But this detail is irrelevant to the topic.
2. The example above explains the basic and most simple situation. The more general solutions takes also care of a few issues with this method (line of sight, ..).

Treeston:
OK - that's exactly what I wanted. :+1: from me.

jackpoz:
as suggested at https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-243252027 , I think a mix of both New and Angle solutions should be implemented, in particular using your New method first, check LoS with target and then apply Angle to find a point in LoS (the way angles are chosen could be changed to jump from a side to the other instead of checking them like around the clock). I'm not sure if a 2nd path should be calculated to reach the new point, maybe just from the last point of New.

I'd suggest to merge this PR with the above suggestion first, then we can check how to modify Detour (what would you like to modify in Detour btw ?)

chaodhib:
Could you describe as clearly as possibly a situation where using a combination of the New method and the Angle method would be more suitable than just the New method?

> the way angles are chosen could be changed to jump from a side to the other instead of checking them like around the clock

Indeed. It would be an improvement compared to the current way of doing. Nonetheless, it's still asymmetrical. And therefore, some situations will be handled awkwardly, depending on if the NPC approaches by the left or by the right (like in the blade's edge arena, when you go up on the platform. i will provide more details later).

> what would you like to modify in Detour btw ?

Detour allows to find a walkable path from point A to point B. Ideally, the concept of LoS and combat reach should be integrated into Detour as well. It would solve the [situation mentioned above](https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-243398938).

jackpoz:
> Could you describe as clearly as possible a situation where using a combination of the New method and the Angle method would be more suitable than just the New method?

Imagine an enemy hiding behind a crate: the New method finds a path (that would walk throught the crate), reduces the length by the combat reach and sets the end point in front of the crate with the enemy not in LoS.
At this point, checking for other points in a circle around the enemy allows to find a point in LoS.

image

Rochet2:
Would it be more efficient and good enough to walk along the originally calculated path rather than calculate the points to different spots around the target?

For example calculate path to target. Cut it by the combat reach. Check LOS at that point. Reduce the cut by X. Check LOS at that point. Repeat until our cut is 0.
This way we calculate just one route and check LOS cut_portion/X times at maximum.

Treeston:
@jackpoz detour would route you around the crate, no?

chaodhib:
@Rochet2 Yes. I didn't mentioned it in the example above because i wanted to keep it simple. But yes, my algorithm works exactly the way you described.

@jackpoz I agree with @Treeston. Detour will find a path that will go around the obstacle. Unless the crate is a DynObj. In which case, the problem is the same if you increase the size of the crate. The Angle approach would not help. A true solution to this situation would be to support Dynamic Objects in the pathfinding. Which is an entirely different topic.

jackpoz:
If you think Detour will find a path around a GameObject then you should go back studying how TC uses Recast and Detour.
How does the Angle approach not solve the crate issue, no matter what the size is ? Any of the points I highlighted in the screenshot above will be valid no matter how big the crate is (since obviously the crate can't grow on the upper side where there's a NPC standing there, NPC can't be inside a crate).

chaodhib:
> If you think Detour will find a path around a GameObject then you should go back studying how TC uses Recast and Detour.

I said exactly the opposite. In your previous post, you didn't specified if the obstacle was a GO or not (sorry, in my previous post, by dynobj i meant GO). So there are 2 cases:
- If the obstacle is not a GO, Recast will find a path around. --> there is no issue. 
- If the obstacle _is_ a GO, it's a whole different problem. As you know, GO are not taken into account by Detour. And the solution to that is not to add a circle approach. But to fix [this](https://github.com/TrinityCore/TrinityCore/issues/14120). The Angle approach is too unreliable in order to be a good solution, even if it is used in combination with the New approach.

ccrs:
And dont forget about slopes.

Path calculated -> path reduced -> movegen checks real distance -> recalculates path (loop)

jackpoz:
> you didn't specify if the obstacle was a GO

> Imagine an enemy hiding behind a crate

really ? why do you think I chose a crate ? it was obviously a GO.

> along with it comes a lot of issues because of its unreliability (point under the map, asymmetrical, ...)

which point under map ? which asymmetrical ? where are all these "lot of issues" ?

Why is it so hard for you to just implement a part of algo that fixes some corner cases that your algo totally ignores ? are we discussing about philosophy here or are we trying to improve TC ? Just add it at the end so we can merge this PR and move on.

jackpoz:
> @Rochet2 Yes. I didn't mentioned it in the example above because i wanted to keep it simple. But yes, my algorithm works exactly the way you described.

@chaodhib does that means the cutting will notice that the target is not in LoS in my crate example and as result make the creature move to the exact location as the target ? that would mean the creature will be able to attack it and will cause pretty much no issue. If this is the case then this is how your New algo handles the crate case and it looks good to me. Please confirm if that's the case, then we can merge it.

chaodhib:
Yes, in worst case scenario, the creature will move to the exact location of the target. It will cut less and less until it finds a position where the creature is both close enough and in LoS of the target.

Treeston:
Out of curiosity, does this do anything to address #20173 too?

ccrs:
it should

chaodhib:
No it doesn't. Last time I checked, that issue was relatively simple to solve (I could be mistaken though). I will take a look again when I can.

src/server/game/Movement/PathGenerator.cpp
jackpoz: please add an ASSERT(target); here

src/server/game/Movement/PathGenerator.cpp
jackpoz: please compare the squared distances

jackpoz (changes requested):
Couple of minor things and we can merge this one

chaodhib:
Besides your 2 points @jackpoz, there are remaining fix to implement before this PR can be merged: support of MoveChase with offset. And support of MoveFollow.

I'll get into it as soon as i can.

Aokromes:
Conflicting files
src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp

jackpoz:
@chaodhib , do you have any plan to continue working on this ?

chaodhib:
@jackpoz At the moment, I'm focused on other PRs. I still believe the idea behind this one is sound though. Maybe I'll get back to it in the future. Until then, if someone wants to pick up where I left off and work on this, I'm more than happy to give the reins. Help is welcome.

illfated:
Add the label **[help wanted]** , maybe?

chaodhib:
Sure

jackpoz:

PR #17852, opened 2016-08-23
Core/SmartAI: add param2 to SMART_ACTION_UPDATE_TEMPLATE to prevent change of level between entry updates

### LINKED ISSUE #16371: Quest - Inoculation - randomly leveling owlkin?
TrinityCore rev. 3f886025a255 2016-01-22 22:14:28 +0100 (3.3.5 branch) (Unix, Release)

Quest:  http://www.wowhead.com/quest=37444/inoculation

When you use the Inoculating Crystal (Item #22962), sometimes the level 3 Nestlewood Owlkin level to level 4.  This happened on 4 of 5 level 3 Nestlewood Owlkin I tested it on.  It didn't happen on a level 4 (to level 5).  They get the little light swirl and everything.  Not a bad bug, but made me look twice several times.

SnapperRy:
**Changes proposed:** most (if not all? Can't think of any exception) creatures should keep their current level when changing their entry.

With this change, param2 can be set to 1 to force a level reroll (based on template levels) and 0 (default for every current use of the action) allows to keep the original level.

Core scripts can use the third parameter in UpdateEntry() (set to false) to achieve the same. I'll open another PR for that once I know for sure which creatures should change their level (if any).

**Target branch(es):** 3.3.5

**Issues addressed:** Closes #16371

**Tests performed:** tested and working

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -1013,7 +1013,7 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
 
             for (ObjectList::const_iterator itr = targets->begin(); itr != targets->end(); ++itr)
                 if (IsCreature(*itr))
-                    (*itr)->ToCreature()->UpdateEntry(e.action.updateTemplate.creature);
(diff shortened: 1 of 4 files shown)


tkrokli:
:+1: Nice, thanks for looking into this. Will be nice to see less "level up" animations for scripted creatures.

PR #17854, opened 2016-08-23
Core/Achievements: fix statistics related to epic/legendary items looted and received

SnapperRy:
**Changes proposed:** I'll be the first to say that this is not really a good way to handle it. I'm looking for feedback and suggestions on a way to improve it.

So, statistics that keep track of looted and obtained epic and legendary items use the same achievement criteria. It means that both will increase at the same pace even when not supposed to.

Example: loot one legendary item, notice that the epic items obtained statistic increases too.

**Target branch(es):** 3.3.5

- [ ] Throw the hack out the window

### DIFF
--- a/src/server/game/Achievements/AchievementMgr.cpp
+++ b/src/server/game/Achievements/AchievementMgr.cpp
@@ -1920,12 +1920,30 @@ bool AchievementMgr::RequirementsSatisfied(AchievementCriteriaEntry const* achie
                 return false;
             break;
         case ACHIEVEMENT_CRITERIA_TYPE_LOOT_EPIC_ITEM:
+        {
+            // Only Epic items should count
+            if (!miscValue1)
+                return false;
+            ItemTemplate const* proto = sObjectMgr->GetItemTemplate(miscValue1);
+            if (!proto || proto->Quality != ITEM_QUALITY_EPIC)
+                return false;
+            break;
+        }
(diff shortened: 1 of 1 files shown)


ariel-:
One word: AchievementCriteriaScript

SnapperRy:
@ariel-: tried that, but couldn't find a way to get a hold of the looted item to check its quality.

Can it be done with the current system and I just don't know how, or would it require some heavy lifting?

ariel-:
There, just use ACHIEVEMENT_CRITERIA_DATA_TYPE_S_EQUIPED_ITEM (19) in DB `achievement_criteria_data`

param1 is min item level (can be left as 0)
param2 is item quality

However, I agree AchievementCriteriaScript should have an accessor or a parameter returning a pointer to AchievementEntry const (just like ConditionScript does)

SnapperRy:
Do you think creating a new achievement_data_type would be better in this case? Checks make sure item entry cannot be 0, and it makes sense in that case.

SnapperRy:
ACHIEVEMENT_CRITERIA_DATA_TYPE_S_EQUIPED_ITEM isn't enough sadly, it has the same issue I'm trying to solve. It checks if the quality is "x or higher", which is what makes legendary items count for the  epic looted item criteria in the first place.

I'm leaning towards a new criteria data type that checks for the exact same quality.

Issue #15780, opened 2015-10-30
[3.3.5] Instance Save Weirdness


ariel-:
Hi! I detected this weird entry on instance saves, the data "U U" is for ulduar raid, however mapid=189 corresponds to Scarlet Monastery

```
mysql> SELECT * FROM characters.instance WHERE id = 29;
+----+-----+------------+------------+---------------------+------------------------------------------------------------+
| id | map | resettime  | difficulty | completedEncounters | data                                                       |
+----+-----+------------+------------+---------------------+------------------------------------------------------------+
| 29 | 189 | 1446176358 |          0 |                 127 | U U 3 3 3 3 3 3 3 0 0 3 5 0 0 5 3 3 3 2 61 0 0 0 0 0 579 0 |
+----+-----+------------+------------+---------------------+------------------------------------------------------------+
1 row in set (0.00 sec)

```

641f34c78a41e3ebe1428c431dbbd2808a543cf9

Nyeriah:
I can confirm this... can happen to any raid. First time it happened to me it was in Molten Core, the instance ID simply disappeared and then whenever I killed a boss it saved me for Maraudon instead.

Shauren:
ok. recycling instace ids is a bad thing.
TODO: remove that.

ariel-:
Got this at worldserver

SQL(p): INSERT INTO instance (id, map, resettime, difficulty, completedEncounters, data) VALUES (64, 249, 0, 0, 0, 'O L 5 ')
 [ERROR]: [1062] Duplicate entry '64' for key 'PRIMARY'

Aokromes:
Yeah i get that, and 128, and 196 from time to time.

Voidx:
yep same problem here.

Voidx:
Some change on this?

Aokromes:
15$ bounty added here.

ariel-:
> Yeah i get that, and 128, and 196 from time to time.

Interesting point, all multiples of 32

Related code: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Maps/MapManager.cpp#L319

xinef1:
@ariel- well, dont use vector<bool> if you do not know how it really works or if you have math problems :D try this patch (was written in notepad but should work), it removes runtime recycling (instances are recycled at every restart so no benefits). Should fix some problems with overriding ids.
PS. whole instance saving should be rewritten

<details>

```diff
[...]

PR #17856, opened 2016-08-24
Core/Instance: Fix GetInstanceSave sometimes checking the wrong map

roc13x:
**Changes proposed:**
-  Always check the correct map ID in GetInstanceSave. If the player was in a group, it would check for a save on the players current map instead of the requested mapID

**Tests performed:** Builds, works correctly ingame


jackpoz:
thank you for the fix. do you know if this solves any known issue too ?

Nyeriah:
Any chance this helps with https://github.com/TrinityCore/TrinityCore/issues/15780 ?

roc13x:
I don't know of any reported issues on the tracker, but it affected the "change dungeon entrance" feature on 6.x (you could no longer exit an instance through the portal)

@Nyeriah It wouldn't help with that, but I am currently testing a fix for that bug and will PR if it works

PR #17861, opened 2016-08-25
Core/Instances: Stop recycling instance IDs

### LINKED ISSUE #15780: [3.3.5] Instance Save Weirdness

roc13x:
**Changes proposed:**
-  Remove the system that re-uses instance ID numbers, instead just increment. This is a lot safer and avoids problems with instances overwriting eachother.

**Issues addressed:** Closes #15780

**Tests performed:** Builds, appears to work fine from my tests

- I did not find any problems in my checks, but I advise others to test thoroughly before merging


Treeston:
Permanent lockouts that persist through reboot are going to push your initial ID higher and higher, eventually causing the server to be unable to launch any instances whatsoever.

Nyeriah:
> Permanent lockouts that persist through reboot are going to push your initial ID higher and higher, eventually causing the server to be unable to launch any instances whatsoever.

Shauren had suggested the removal of the recycle system at https://github.com/TrinityCore/TrinityCore/issues/15780#issuecomment-152447380

roc13x:
It will take you quite a while to launch 4.3 billion instances though :P

Nawuko:
So this would keep up to a maximum of 0xFFFFFFFF ids in the database even when the instance id is no longer accessible by any player? Sounds like a "cheap" solutions for me.

I dont prefer the current method. But at least this should add some kind of instance id validation to remove invalid/expired ids from the database.

Or a better thread safe implementation of the current system.

Summoning master @Shauren for word of advice :)

Treeston:
As far as I can tell, this still deletes old IDs.

Nawuko:
@Treeston you are right, i was thinking the FreeInstanceId was in charge of removing the database entry as well.

But i noticed a removal of line https://github.com/TrinityCore/TrinityCore/pull/17861/files#diff-ba31b87912fb8436628b3452e1089c22L373 I dont know if this is called elsewhere.

Shauren:
As this without a doubt will overflow instance ids at some point, you need to offer an alternative approach. One way would be to reset the sequence during weekly instance reset and renumber ids of anything that doesnt reset at that point (the weird 3 day AQ20, also extended saves) to start from 1

@Nawuko thats for realm first achievements (its implemented wrong in 3.3.5, 6.x has the correct and blizzlike way for that)

Treeston:
Lots of permanent lockouts will persist through weekly lockout reset due to extension system.

roc13x:
@Shauren Would re-numbering instances at startup be acceptable? I imagine trying to re-number them while the server is running and instances are in progress would be a nightmare

Shauren:
Could be acceptable if you applied extra conditions, for example:
* Don't renumber if MAX(instanceId) - MIN(instanceId) < COUNT(*) * 2 // every other id is used
* Renumber if MAX(instanceId) - MIN(instanceId) > MIN(instanceId)  // large unused gap in front

src/server/game/Instances/InstanceSaveMgr.cpp
```diff
@@ -286,8 +286,32 @@ void InstanceSaveManager::LoadInstances()
...
 
+    // Reduce instance ID numbers when they climb too high, to prevent overflow
+    QueryResult result = CharacterDatabase.Query("SELECT MIN(id),MAX(id) FROM instance");
+    if (result)
+    {
+        uint32 minId = (*result)[0].GetUInt32();
+        uint32 maxId = (*result)[1].GetUInt32();
+
+        // If the gap at the front is larger than the ID range, shift all IDs down to start at 1
+        if (maxId > minId && minId > maxId-minId)
```
jackpoz: did you mean "maxId != minId" instead of "maxId > minId" ? the > instead of != looks quite confusing

src/server/game/Instances/InstanceSaveMgr.cpp
```diff
@@ -286,8 +286,32 @@ void InstanceSaveManager::LoadInstances()
...
+        {
+            uint32 gap = minId - 1;
+            CharacterDatabase.DirectPExecute("UPDATE instance SET id = id-%u ORDER BY id ASC", gap);
```
jackpoz: all these statements should be in a single transaction, you don't want to update a table but fail another table

src/server/game/Maps/MapManager.cpp
```diff
@@ -306,69 +306,19 @@ uint32 MapManager::GetNumPlayersInInstances()
...
-    {
-        if ((i < _instanceIds.size() && !_instanceIds[i]) || i >= _instanceIds.size())
-        {
-            _nextInstanceId = i;
-            break;
-        }
-    }
-
-    if (newInstanceId == _nextInstanceId)
+    if (_instanceId >= 0xFFFFFFFE)
```
jackpoz: _instanceId with value 0xFFFFFFFE is not valid after this point ?

Aokromes:
why no re-number only if it reachs 4.000 millons? I seriously doub any server will reach that number on years.

Treeston:
Agree with @Aokromes, don't re-number unless you need to.

roc13x:
Yeah looking back at it I agree, that last commit isn't a good way of going about it, reverting that bit.

Re-numbering seems unnecessary, it only becomes a problem after many years, and if you ever manage to reach the limit, just clear the few active saves and you're back to 1

Treeston:
Re-numbering is fine - just don't do it unless max > a certain arbitrary threshold.

roc13x:
Yea I can restore the function, just want to make those changes such as putting it in a transaction. What'd be a reasonable threshold? 4bn? 2bn?

Treeston:
4bn should be more than enough. Maybe add a warning log when handing out instance ID 4,100,000,000 too warning that reboot may be required soon?

src/server/game/Instances/InstanceSaveMgr.cpp
```diff
@@ -286,8 +286,38 @@ void InstanceSaveManager::LoadInstances()
...
+
+        // If IDs have reached 4 billion, re-number
+        if (maxId >= uint32(0xEE6B2800))
+        {
+            TC_LOG_INFO("server.loading", "Instance IDs exceed 4000000000, re-numbering...");
+
+            uint32 gap = minId - 1;
+            SQLTransaction trans = CharacterDatabase.BeginTransaction();
+
+            trans->PAppend("UPDATE instance SET id = id-%u ORDER BY id ASC", gap);
```
Shauren: this will fail if someone has a very low id and keeps extending the save forever.

By renumbering I meant also compacting the ids, for example
[4, 23, 546, 2132] would become [1, 2, 3, 4], not, [1, 20, 543, 2129]

ariel-:
Perhaps an easier solution would be keep on recycling but fix the logic fail that causes it to reuse current ids

Aokromes:
what's the status of this PR?

src/server/game/Instances/InstanceSaveMgr.cpp
Aokromes: @roc13x

Aokromes (changes requested):

Treeston:
Create a temporary table of pairs `(oldid,newid)` where oldid is just instanceid from existing table sorted asc, and newid is the index of the row in that query.

Then just use that table to consistently update all other tables to the new IDs.

Aokromes:
Conflicting files
src/server/game/Instances/InstanceSaveMgr.cpp
src/server/game/Maps/MapManager.cpp

PR #17865, opened 2016-08-27
Enhance quest status hook

Rochet2:
**Changes proposed:**
- One thing to note is that there already exists hooks for quest accept and quest reward(end of quest). However those hooks are for CreatureScript and GameObjectScript and this hook is for PlayerScript
- This PR enhances quest status hook by making it catch more, if not all quest status changes. For example quest add, fail, remove
- Before this hook only tracked when the quest status was changed through quest status method, which is not the only thing that changes quest statuses - as you can see by looking at `Player::GetQuestStatus` function contents.
- Unfortunately this changes the API to not pass the quest status through hook parameters, but if needed you can get it with `player->GetQuestStatus(questid)`, fortunately TC doesnt use the hook.

**Issues addressed:** https://community.trinitycore.org/topic/12843-c-quest-status-problem/

**Tests performed:**
- Does build
- Tested in game that the start and end as well as status changes like quest objective completion of a couple of quests were catched (first two/three draenei and human start quests)
- There seemed to be no duplicate catches
- Only tested on 3.3.5 branch

- [ ] Need more testing

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -14832,6 +14832,8 @@ void Player::AddQuest(Quest const* quest, Object* questGiver)
         // add to Quest Tracker
         CharacterDatabase.Execute(stmt);
     }
+
+    sScriptMgr->OnQuestStatusChange(this, quest_id);
 }
 
 void Player::CompleteQuest(uint32 quest_id)
@@ -15074,6 +15076,8 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver, 
 
     //lets remove flag for delayed teleports
     SetCanDelayTeleport(false);
+
+    sScriptMgr->OnQuestStatusChange(this, quest_id);
 }
 
 void Player::FailQuest(uint32 questId)
@@ -15734,7 +15738,7 @@ void Player::SetQuestStatus(uint32 questId, QuestStatus status, bool update /*= 
     if (update)
         SendQuestUpdate();
 
-    sScriptMgr->OnQuestStatusChange(this, questId, status);
+    sScriptMgr->OnQuestStatusChange(this, questId);
 }
 
 void Player::RemoveActiveQuest(uint32 questId, bool update /*= true*/)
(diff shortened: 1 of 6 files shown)


jackpoz:
Should Player::RemoveActiveQuest() trigger the hook instead of every caller of Player::RemoveActiveQuest() having to do it ? Is there any drawback or issue doing so ?

Rochet2:
> Should Player::RemoveActiveQuest() trigger the hook instead of every caller of Player::RemoveActiveQuest() having to do it ? Is there any drawback or issue doing so ?

I was somewhat afraid of having cases where the hook triggers two times unintentionally in a way.
For example the code in cs_quest.cpp calls that as well as RemoveRewardedQuest. The ultimate change is to remove the quest, so the hook kind of should only trigger once with the final quest status. But if it is possible to trigger it twice it is in a way wrong to have some kind of middle state instead of just triggering the hook when quest status is that the quest is removed.

I however did not inspect if there are such cases possible where the hook could trigger twice like that and I am not quite sure if it is desired to trigger twice if there is a case like that.



> Also what does player->GetQuestStatus(questid) return in current code after RemoveActiveQuest() is called ?

It returns QUEST_STATUS_NONE or QUEST_STATUS_REWARDED depending on if the quest was rewarded.
After RemoveRewardedQuest it can return any quest status except QUEST_STATUS_REWARDED.
When both are called it can return only QUEST_STATUS_NONE.
.. As far as I can see.

So I guess cases where the hook calling matters is when a quest can be rewarded and can be in some other state as well, like completed, failed, incomplete. Maybe that applies for repeatable quests.

jackpoz:
it makes sense to trigger the hook only once for every status changed, so if the hook call can't be moved to RemoveActiveQuest() we'll keep it like this.
QUEST_STATUS_NONE when a quest has just been abandoned is more of a documentation thing, now we know what it returns (since there's no "QUEST_STATUS_JUST_ABANDONED" )

jackpoz:
@Rochet2 so this is ready to be merged ?

Rochet2:
@jackpoz Yes, I believe so.

PR #17885, opened 2016-09-01
[3.3.5] AHBot Characters

iridinite:
**Changes proposed:**
- This PR implements auction house bot characters. The idea is that you create a dummy account and add some characters to it (they only need to be there for the name) and whenever the ahbot creates or bids on an auction, it will pick one of those characters as the owner/bidder. The reason for this is that I think it adds a tiny bit of life to the AH, instead of having your auctions being bought up by an empty string.
- I added a setting in the worldconfig that specifies the account ID that belongs to the AHbot. On server startup, it will load all associated character IDs. Note: If the account ID is zero, old behaviour is preserved.
- The bot characters will never receive auction-related mail (see my changes to AuctionHouseMgr.cpp) so they will not clutter up the item / mail tables just because they're 'real' characters.

**Target branch(es):** 3.3.5

**Tests performed:** Built and tested (release, x64) locally for several days, works fine. I have _not_ tested with ahbot disabled or account ID set to zero, though the code should work with this also.

### DIFF
--- a/src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
+++ b/src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
@@ -19,6 +19,7 @@
 #include "Item.h"
 #include "World.h"
 #include "Config.h"
+#include "AccountMgr.h"
 #include "AuctionHouseMgr.h"
 #include "AuctionHouseBot.h"
 #include "AuctionHouseBotBuyer.h"
@@ -55,6 +56,36 @@ bool AuctionBotConfig::Initialize()
 
     _itemsPerCycleBoost = GetConfig(CONFIG_AHBOT_ITEMS_PER_CYCLE_BOOST);
     _itemsPerCycleNormal = GetConfig(CONFIG_AHBOT_ITEMS_PER_CYCLE_NORMAL);
+
+    if (GetConfig(CONFIG_AHBOT_ACCOUNT_ID))
+    {
+        // check character count
+        uint32 charcount = AccountMgr::GetCharactersCount(GetConfig(CONFIG_AHBOT_ACCOUNT_ID));
+        if (charcount)
+        {
+            // find account guids associated with ahbot account
+            PreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_SEL_CHARS_BY_ACCOUNT_ID);
+            stmt->setUInt32(0, GetConfig(CONFIG_AHBOT_ACCOUNT_ID));
+            PreparedQueryResult result = CharacterDatabase.Query(stmt);
+            
+            for (uint32 charno = 0; charno < result->GetRowCount(); charno++)
+            {
+                _AHBotCharacters.push_back((*result)[0].GetUInt32());
+                result->NextRow();
+            }
+
+            TC_LOG_DEBUG("ahbot", "AuctionHouseBot found %i characters", charcount);
+        }
+        else
+        {
+            _AHBotCharacters.push_back(uint32(0));
+            TC_LOG_WARN("ahbot", "AuctionHouseBot Account ID has no associated characters.");
+        }
+    }
+    else
+    {
+        _AHBotCharacters.push_back(uint32(0));
+    }
 
     return true;
 }
@@ -111,6 +142,8 @@ void AuctionBotConfig::SetConfig(AuctionBotConfigFloatValues index, char const* 
 //Get AuctionHousebot configuration file
 void AuctionBotConfig::GetConfigFromFile()
 {
+	SetConfig(CONFIG_AHBOT_ACCOUNT_ID, "AuctionHouseBot.Account", 0);
+
     SetConfigMax(CONFIG_AHBOT_ALLIANCE_ITEM_AMOUNT_RATIO, "AuctionHouseBot.Alliance.Items.Amount.Ratio", 100, 10000);
     SetConfigMax(CONFIG_AHBOT_HORDE_ITEM_AMOUNT_RATIO, "AuctionHouseBot.Horde.Items.Amount.Ratio", 100, 10000);
     SetConfigMax(CONFIG_AHBOT_NEUTRAL_ITEM_AMOUNT_RATIO, "AuctionHouseBot.Neutral.Items.Amount.Ratio", 100, 10000);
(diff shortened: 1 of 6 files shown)


Aokromes:
The version on core was pushed to avoid the need of this.

DDuarte:
I'm okay with this as long as it is optional

iridinite:
@Aokromes Sorry, what do you mean?

@DDuarte Yes, it is optional. If account 0 is specified in the config (which it is by default) then old behaviour is preserved.

src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
```diff
@@ -111,6 +142,8 @@ void AuctionBotConfig::SetConfig(AuctionBotConfigFloatValues index, char const*
 //Get AuctionHousebot configuration file
 void AuctionBotConfig::GetConfigFromFile()
 {
+	SetConfig(CONFIG_AHBOT_ACCOUNT_ID, "AuctionHouseBot.Account", 0);
```
tkrokli: Replace `<Tab>` with 4 spaces. GitHub displays `<Tab>` as 8 spaces and TC will not accept Tabs in the code. :)

iridinite: @tkrokli  Corrected, thanks!

src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
```diff
@@ -272,6 +305,31 @@ char const* AuctionBotConfig::GetHouseTypeName(AuctionHouseType houseType)
...
+// to have another character than the auction owner place bids
+uint32 AuctionBotConfig::GetRandCharExclude(uint32 exclude) const
+{
+    // avoid freezing if only one ahbot char (which defeats the purpose but oh well)
+    if (_AHBotCharacters.size() == 1) return GetRandChar();
+
+    uint32 result;
+    do {
+        result = GetRandChar();
+    } while (result == exclude);
```
tkrokli: Recommended TC code style:
```cpp
    do
    {
        result = GetRandChar();
    }
    while (result == exclude);
```

Naios:
Isn't it better to make virtual players account independent?

We could create a table "virtual_characters" that reserve names so they can be used vor various virtual tasks like player bots ot AH bot...

iridinite:
@Naios That'd be cool, but I think that's beyond the scope of this PR. :P

Naios:
No it isn't. Before adding a dirty workarround that is hard revertable, we could add it right instead.

I'm sure that the effort to implement this idea into the PR isn't that huge.

iridinite:
Hmm. I'm interested but not quite convinced of the necessity. I'm also new to this codebase, so apologies in advance if I get stuff wrong.
How would you structure such a `virtual_characters` table? I mean not just in SQL, but also, such a bot characters table would require special handling for them in several places. What would make a virtual character - especially player bots - so different that it would require a seperate table? They'd still need to have an ID (which cannot overlap with any in the `characters` table), a name, possibly inventory and guild entries, etc. And afaik - correct me if I'm wrong - the client still wants a character ID on its auctions, for which it can then look up the associated name.
I agree that creating a dummy account for them is a bit weird, but I also think that implementing virtual characters would require rewriting and/or adding to some of the core account/character logic.
Would it perhaps be an idea to add a column to `auth.account` that describes whether the account represents a bot or not?

tkrokli:
Maybe @jackpoz would be interested, since it touches on the subject of player bots like his botfarm?

src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
ariel-: Actually it's

```c++
do
{
    // loop body here
} while (/*loop condition*/);
```

jackpoz:
the whole point of botfarm is to require 0 changes in tc code so any discussion about tc does doesn't relate to botfarm

src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
```diff
@@ -272,6 +305,34 @@ char const* AuctionBotConfig::GetHouseTypeName(AuctionHouseType houseType)
...
+{
+    return _AHBotCharacters[urand(0, _AHBotCharacters.size() - 1)];
+}
+
+// Picks a random AHBot character, but excludes a specific one. This is used
+// to have another character than the auction owner place bids
+uint32 AuctionBotConfig::GetRandCharExclude(uint32 exclude) const
+{
+    // avoid freezing if only one ahbot char (which defeats the purpose but oh well)
+    if (_AHBotCharacters.size() == 1) return _AHBotCharacters[0];
```
jackpoz: please don't put if and return in same line

src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
tkrokli: Well, that would work too. As long as the first `{` is moved underneath the `do` command, it's OK.
Both forms have been used in TC source code previously, so I suppose you can use both.

src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
```diff
@@ -272,6 +305,35 @@ char const* AuctionBotConfig::GetHouseTypeName(AuctionHouseType houseType)
...
+        result = GetRandChar();
+    }
+    while (result == exclude);
+    
+    return result;
+}
+
+bool AuctionBotConfig::GetIsBotChar(uint32 characterID) const
+{
+    return !characterID || find(_AHBotCharacters.begin(), _AHBotCharacters.end(), characterID) != _AHBotCharacters.end();
```
jackpoz: please use std:: prefix for std:: functions

jackpoz:
@Naios the way this PR handles optional AH characters is very easy to setup for the user and satisfies all current needs. 
"dirty workarround that is hard revertable" and "the effort to implement this idea into the PR isn't that huge" can't go together since you are already asking to revert the workaround and implement it in another way. On top of that we are never gonna merge PlayerBots, that requires too many changes to  https://github.com/TrinityCore/TrinityCore/tree/3.3.5/src/server/game .

src/server/game/AuctionHouseBot/AuctionHouseBotBuyer.cpp
```diff
@@ -391,11 +391,11 @@ void AuctionBotBuyer::BuyEntry(AuctionEntry* auction, AuctionHouseObject* auctio
     SQLTransaction trans = CharacterDatabase.BeginTransaction();
 
     // Send mail to previous bidder if any
-    if (auction->bidder)
+    if (auction->bidder && !sAuctionBotConfig->GetIsBotChar(auction->bidder))
         sAuctionMgr->SendAuctionOutbiddedMail(auction, auction->buyout, NULL, trans);
 
     // Set bot as bidder and set new bid amount
-    auction->bidder = 0;
+    auction->bidder = sAuctionBotConfig->GetRandCharExclude(auction->owner);
```
jackpoz: what is the *Exclude() part needed for here ?

iridinite: So the bidding character will not also be the auction owner. Cosmetic, really, nothing more.

src/server/game/AuctionHouseBot/AuctionHouseBot.h
```diff
@@ -224,13 +225,17 @@ class TC_GAME_API AuctionBotConfig
 
     uint32 GetItemPerCycleBoost() const { return _itemsPerCycleBoost; }
     uint32 GetItemPerCycleNormal() const { return _itemsPerCycleNormal; }
+    uint32 GetRandChar() const;
+    uint32 GetRandCharExclude(uint32 exclude) const;
+    bool GetIsBotChar(uint32 characterID) const;
```
jackpoz: IsBotChar() sounds a bit better than GetIsBotChar()

iridinite: Wasn't sure about that one, figured I'd stick with the Get* prefix the other functions have. Will change momentarily.

PR #17890, opened 2016-09-03
Fixed issue #17884: make broadcast_text_locale work.

### LINKED ISSUE #17884: hotfixes.broadcast_text_locale doesn't work
**Description**:
I added a entry in hotfixes.broadcast_text_locale but server still refer to the text in table hotfixes.broadcast_text for gossip_menu_option.

server doesn't get the text from table hotfixes.broadcast_text_locale for gossip_menu_option.

get the localized text from hotfixes.broadcast_text_locale for gossip_menu

1. chat with any city guard.
2. 
3. 

6.x

**TC hash/commit**:  
 59fe9087447dc58c3f41fb6f647bc2b50a95b9b5

**TDB version**:  
6.04

**Operating system**:  
Windows.

[//]: # (For fixes containing c++ create a Pull Request)

Druson:
**Changes proposed:**
## 

**Target branch(es): 6.x

**Issues addressed:** Closes #17884

**Tests performed: yes, tested in game.

- [none ] 
- [ ]

### DIFF
--- a/src/server/game/DataStores/DB2Stores.cpp
+++ b/src/server/game/DataStores/DB2Stores.cpp
@@ -926,7 +926,7 @@ static void LoadBroadcastTextLocales()
         count++;
     } while(result->NextRow());
 
-    TC_LOG_INFO("server.loading", ">> Loaded broadcast_text locale strings in %u ms", count, GetMSTimeDiffToNow(oldMSTime));
+    TC_LOG_INFO("server.loading", ">> Loaded %u broadcast_text locale strings in %u ms", count, GetMSTimeDiffToNow(oldMSTime));
 }
 
 void DB2Manager::LoadHotfixData()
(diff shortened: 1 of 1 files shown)


Shauren:
This is just a bad workaround, locale data should be loaded through `DB2DatabaseLoader::LoadStrings`

Druson:
@Shauren  I don't think this is a bad workaround. I have checked that ther is no locale data  loaded.  You may take a look at the table layout of hotfixes.broadcast_text, there is no locale column in the table now, the locale columns have been moved to table hotfixes.broadcast_text_locale.

Shauren:
You are missing the point, all DB2 stores should load their locale data with the code already in place
Loaded from here https://github.com/TrinityCore/TrinityCore/blob/6.x/src/server/game/DataStores/DB2Stores.cpp#L263

Druson:
@Shauren Checked the code in DB2Stores.cpp,  seems the bug is in https://github.com/TrinityCore/TrinityCore/blob/6.x/src/server/game/DataStores/DB2Stores.cpp#L256

I changed "defaultLocale == i" to "DEFAULT_LOCALE == i", and it works now.

Issue #16859, opened 2016-03-28
Undocumented config options


DDuarte:
**Description**:

The following config options are missing from master worldserver.conf:

``` ini
GuarderSight = 50.0

Guild.SaveInterval = 15
```

**Branch(es):** master 

**TC hash/commit**: 070bedae0c78a870ef69f6796ac956896c8e7fa2/ 580773dc1a52cf6c8177647baf15afccd557be60

**Operating system**: Win10

tkrokli:
TrinityCore rev. 4afea55b2c60 2016-03-28 05:47:41 +0200 (3.3.5 branch) (Win64, Release)
```
Missing name DisconnectToleranceInterval in config file worldserver.conf, add "DisconnectToleranceInterval = 0" to this file
Missing name MonsterSight in config file worldserver.conf, add "MonsterSight = 50.000000" to this file
Missing name GuarderSight in config file worldserver.conf, add "GuarderSight = 50.000000" to this file
Will clear `logs` table of entries older than 1209600 seconds every 10 minutes.
Missing name Battleground.PrematureFinishTimer in config file worldserver.conf, add "Battleground.PrematureFinishTimer = 300000" to this file
Missing name Battleground.PremadeGroupWaitForMatch in config file worldserver.conf, add "Battleground.PremadeGroupWaitForMatch = 1800000" to this file
```
Worldserver says `Battleground.PrematureFinishTimer` + `Battleground.PremadeGroupWaitForMatch` are missing, although both those 2 values and their documentation is present in both my `Worldserver.conf.dist` and `Worldserver.conf` files:
```
#    Battleground.PrematureFinishTimer
#        Description: Time (in milliseconds) before battleground will end prematurely if there are
#                     not enough players on one team. (Values defined in battleground template)
#        Default:     300000 - (Enabled, 5 minutes)
#                     0      - (Disabled, Not recommended)

BattleGround.PrematureFinishTimer = 300000

#
#    BattleGround.PremadeGroupWaitForMatch
#        Description: Time (in milliseconds) a pre-made group has to wait for matching group of the
#                     other faction.
#        Default:     1800000 - (Enabled, 30 minutes)
#                     0       - (Disabled, Not recommended)

BattleGround.PremadeGroupWaitForMatch = 1800000
```
[...]

PR #17906, opened 2016-09-05
[6.x] Update worldserver.conf.dist

alexandre433:
**Target branch(es):** 6.x

**Tests performed:** It builds and it was tested (Does it build, tested in-game, etc)

Change of some errors which give on the server.log for not beeing there.

I don't really know some descriptions thus them beeing "None"

### DIFF
--- a/src/server/worldserver/worldserver.conf.dist
+++ b/src/server/worldserver/worldserver.conf.dist
@@ -1084,6 +1084,22 @@ Guild.BankEventLogRecordsCount = 25
 Guild.NewsLogRecordsCount = 250
 
 #
+#   Guild.SaveInterval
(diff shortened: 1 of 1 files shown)


tkrokli:
Regarding `MonsterSight` and `GuarderSight`: those options will not be added to Worldserver.conf without a proper description of what they are / what they do. See also [**Undocumented config options**](16859) #16859 for previous discussion on this topic.

alexandre433:
So should I edit the description from what I understand what the names of those options do?

tkrokli:
If you have got correct information about what the options do, it would definitely be an improvement. :)

Shauren:
Adding these values with no description serves no purpose

PR #17920, opened 2016-09-08
Core/Spell: do not send duplicate updates for auras

SnapperRy:
**Changes proposed:**

When the server sends SMSG_AURA_UPDATE_ALL, it also sends a SMSG_AURA_UPDATE for every affected aura. This results in duplicate updates (one from UPDATE_ALL and one from UPDATE).

Sniffs show that the server only sends SMSG_AURA_UPDATE_ALL.

**Target branch(es):** 3.3.5


Shauren:
This is incorrect, Player::SendAurasForTarget happens when some unit becomes visible and you need to see all its auras but other nearby players already see the unit and need to receive normal SMSG_AURA_UPDATE. Your change would break that.

PR #17921, opened 2016-09-09
Core/Gossip: improve gossips related to Dual Talent Specialization

SnapperRy:
**Changes proposed:**
- Correctly hide and show info gossip based on dual spec requirements.
- Show correct gossip after learning dual spec.

Currently, the dual spec info gossip is always shown on the trainer regardless of your level and whether or not you already have it.
Moreover, there is a missing text that should be sent after learning dual spec (as seen [here](https://youtu.be/BdHswNkXa8U?t=30)).

**Target branch(es):** 3.3.5

- [ ] I'm not entirely sure if using a new gossip option for it is the right way™, but to me it seems much cleaner than having two conditions for every gossip menu (amounting to ~400 new conditions), as well as having to implement a new condition type to check for dual spec.

### DIFF
--- a/sql/updates/world/3.3.5/9999_99_99_99_world.sql
+++ b/sql/updates/world/3.3.5/9999_99_99_99_world.sql
@@ -1,0 +1,3 @@
+--
+UPDATE `gossip_menu_option` SET `option_id`=20 WHERE `OptionBroadcastTextID`=33762;
+UPDATE `gossip_menu_option` SET `action_menu_id`=10373 WHERE `option_id`=18;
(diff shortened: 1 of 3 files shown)


tkrokli:
Looks like a clean and neat solution to me, especially since it seems to fit the K.I.S.S. principle. :-)

tkrokli:
Could we have some constructive feedback and/or criticism from other developers regarding this change?
Moreover, it would also be nice to get feedback regarding not displaying the option(s) on low levels.

PR #17939, opened 2016-09-13
Script/Quest: Trail of Fire

SnapperRy:
**Changes proposed:**
-  Script the quest [Trail of Fire](http://www.wowhead.com/quest=11241/trail-of-fire), replacing the previous script that was very far from blizzlike.

It's not perfect (most things are taken from sniffs, but I wasn't able to figure out what is used to equip the torch at the beginning. The creature_equip_template entry is probably a hack, but it's the only way I can think of. Still, works as one would expect it to).

**Target branch(es):** 3.3.5

### DIFF
--- a/src/server/scripts/Northrend/zone_howling_fjord.cpp
+++ b/src/server/scripts/Northrend/zone_howling_fjord.cpp
@@ -40,13 +40,38 @@ EndContentData */
 ######*/
 enum Entries
 {
-    NPC_APOTHECARY_HANES         = 23784,
-    FACTION_ESCORTEE_A           = 774,
-    FACTION_ESCORTEE_H           = 775,
-    NPC_HANES_FIRE_TRIGGER       = 23968,
-    QUEST_TRAIL_OF_FIRE          = 11241,
-    SPELL_COSMETIC_LOW_POLY_FIRE = 56274,
-    SPELL_HEALING_POTION         = 17534
+    NPC_APOTHECARY_HANES = 23784,
+    FACTION_ESCORTEE_H   = 775,
+    QUEST_TRAIL_OF_FIRE  = 11241,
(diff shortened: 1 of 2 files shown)


src/server/scripts/Northrend/zone_howling_fjord.cpp
```diff
@@ -97,19 +126,89 @@ class npc_apothecary_hanes : public CreatureScript
             if (Player* player = GetPlayerForEscort())
                 player->FailQuest(QUEST_TRAIL_OF_FIRE);
         }
-
-        void UpdateEscortAI(uint32 diff) override
-        {
+
```
tkrokli: redundant whitespace (L 129,8 + 131,12)

src/server/scripts/Northrend/zone_howling_fjord.cpp
```diff
@@ -97,19 +126,89 @@ class npc_apothecary_hanes : public CreatureScript
...
+                        me->HandleEmoteCommand(EMOTE_ONESHOT_KNEEL);
+                        break;
+                    case EVENT_TALK_2:
+                        if (Player* player = ObjectAccessor::GetPlayer(*me, _player))
+                            Talk(TALK_2, player);
+                        me->LoadEquipment(EQUIP_TORCH);
+                        me->SetSheath(SHEATH_STATE_MELEE);
+                        break;
+                    case EVENT_BURN_CRATES:
+                        me->CastSpell(me, SPELL_BURN, true);
```
Keader: me == npc_apothecary_hanes?
if yes, change for: DoCastSelf(SPELL_BURN, true);

SnapperRy:
I could have sworn I had already pushed the changes... :P

PR #17941, opened 2016-09-14
[3.3.5-proc] Unit::HandleAuraProc

ccrs:
@ariel- 

you are not alone bby


sql/updates/world/3.3.5/2016_99_99_99_world .sql
```diff
@@ -0,0 +1,12 @@
+DELETE FROM `spell_proc` WHERE `SpellId` IN (70672, 72455, 72832, 72833, 71564);
```
Shauren: dat space in filename

src/server/scripts/Northrend/IcecrownCitadel/boss_blood_prince_council.cpp
```diff
@@ -1498,6 +1498,27 @@ class spell_taldaram_ball_of_inferno_flame : public SpellScriptLoader
...
         }
+
+        class spell_taldaram_ball_of_inferno_flame_AuraScript : public AuraScript
+        {
+            PrepareAuraScript(spell_taldaram_ball_of_inferno_flame_AuraScript);
+
+            void HandleStackDrop(AuraEffect const* /*aurEff*/, ProcEventInfo& /*eventInfo*/)
+            {
+                PreventDefaultAction();
+                GetTarget()->RemoveAuraFromStack(GetId());
```
ariel-: ModStackAmount(-1);

ccrs: no need to, that function already does that and in case there is no stack left it removes the aura... so yeah no need to change it

src/server/scripts/Spells/spell_item.cpp
```diff
@@ -1757,6 +1757,49 @@ class spell_item_six_demon_bag : public SpellScriptLoader
...
+                    return false;
+                return true;
+            }
+
+            void HandleProc(AuraEffect const* aurEff, ProcEventInfo& /*eventInfo*/)
+            {
+                PreventDefaultAction();
+
+                Unit* target = GetTarget();
+                int32 bp0 = CalculatePct(target->GetMaxHealth(), GetSpellInfo()->Effects[EFFECT_0].CalcValue());
```
ariel-: Just use aurEff->GetAmount()

src/server/scripts/Spells/spell_item.cpp
```diff
@@ -333,6 +333,74 @@ class spell_item_blessing_of_ancient_kings : public SpellScriptLoader
...
+            bool Validate(SpellInfo const* /*spellInfo*/) override
+            {
+                if (!sSpellMgr->GetSpellInfo(SPELL_DEADLY_PRECISION))
+                    return false;
+                return true;
+            }
+
+            void HandleDummy(SpellEffIndex /*effIndex*/)
+            {
+                SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(SPELL_DEADLY_PRECISION);
```
ariel-: As the spellInfo exists, because of Validate hook, use AssertSpellInfo so Coverity doesn't complain

ariel-:
So, turns out the combustion issues were caused by the check you mentioned.

Mage casting times were another bug, and you included them in the sql even when I had event entries already :P

You were missing SpellFamilyFlags on several entries, which caused auras to incorrectly proc off other spells.

That being said, thanks for your help!

ccrs:
no problem at all, my pleasure

but next time dont delete my branch XD

PR #17860, opened 2016-08-25
[Still pretty decently WIP] Culling of Stratholme rewrite

Treeston:
Just started this, creating PR so those interested can track progress.

Full rewrite, same scale as Azjol-Nerub (if not bigger due to RP events). Feedback on new code is welcomed.

PR #17946, opened 2016-09-15
[3.3.5] SplineChain movement generation

Treeston:
So, there's a pretty significant difference between waypoint movement on TC and waypoint movement on retail. That much doesn't come as a surprise.

The first approximation we came up with was the waypoint motion scheme, where creatures simply move from one point to the next, with interpolation between them provided by mmaps. This was alright enough for general gameplay, but didn't really provide a smooth feeling for pre-composed sequences or RP events.

The second approximation that I've started recently using in the Stratholme PR (#17860) is what I call "spline shortcutting". Allow me to explain using MS paint:
Image
The layout at the top is what the retail servers send for pre-scripted RP sequences, with each color representing a different movespline (new SMSG_ON_MONSTER_MOVE packet), all of them with smoothing flag set which makes the client seamlessly transition between them, providing fluent movement.
The red spline at the bottom is what smooth pathing in the Stratholme PR currently sends, produced by "shortcutting" the splines, inserting artificial waypoints where a new spline begins and creating one single long movespline that is then dumped to the client.
This works decently well, but there is still some minor stuttering and moonwalking with this at the transition points.

The question remains: Can we do better at replicating retail movement?

In this PR, I'm trying to answer that to the affirmative. The new SplineChainMovementGenerator takes a set of movespline waypoints along with timings for when to override them with the subsequent spline, all of which can easily be parsed from packet sniffs. This is obviously not dynamic generation, but for a static sequence of events, it really doesn't need to be.

Also included is a debug instance that lets you view the new movement generation in action - build the PR branch (will require cmake re-run), then `.go creature id 27891` and perform any emote targeting him to start movement.

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -19,6 +19,7 @@
 #include "MotionMaster.h"
 #include "CreatureAISelector.h"
 #include "Creature.h"
+#include "ScriptSystem.h"
 
 #include "ConfusedMovementGenerator.h"
 #include "FleeingMovementGenerator.h"
@@ -28,6 +29,7 @@
 #include "TargetedMovementGenerator.h"
 #include "WaypointMovementGenerator.h"
 #include "RandomMovementGenerator.h"
+#include "SplineChainMovementGenerator.h"
 #include "MoveSpline.h"
 #include "MoveSplineInit.h"
 
@@ -466,6 +468,33 @@ void MotionMaster::MoveSmoothPath(uint32 pointId, Movement::PointsArray const& p
     Mutate(new EffectMovementGenerator(pointId), MOTION_SLOT_ACTIVE);
     //Position pos(pathPoints[pathSize - 1].x, pathPoints[pathSize - 1].y, pathPoints[pathSize - 1].z);
     //MovePoint(EVENT_CHARGE_PREPATH, pos, false);
+}
+
+void MotionMaster::MoveAlongSplineChain(uint32 pointId, uint32 dbChainId, bool walk)
+{
+    Creature* owner = _owner->ToCreature();
+    if (!owner)
+    {
+        TC_LOG_ERROR("misc", "MotionMaster::MoveAlongSplineChain: non-creature %s tried to walk along DB spline chain. Ignoring.", _owner->GetGUID().ToString().c_str());
+        return;
+    }
+    SplineChain const* chain = sScriptSystemMgr->GetSplineChain(owner, dbChainId);
+    if (!chain)
+    {
+        TC_LOG_ERROR("misc", "MotionMaster::MoveAlongSplineChain: creature with entry %u tried to walk along non-existing spline chain with DB id %u.", owner->GetEntry(), dbChainId);
+        return;
+    }
+    MoveAlongSplineChain(pointId, *chain, walk);
+}
+
+void MotionMaster::MoveAlongSplineChain(uint32 pointId, SplineChain const& chain, bool walk)
+{
+    Mutate(new SplineChainMovementGenerator(pointId, chain, walk), MOTION_SLOT_ACTIVE);
+}
+
+void MotionMaster::ResumeSplineChain(SplineChainResumeInfo const& info)
+{
+    Mutate(new SplineChainMovementGenerator(info), MOTION_SLOT_ACTIVE);
 }
 
 void MotionMaster::MoveFall(uint32 id /*=0*/)
--- a/src/server/game/Movement/MotionMaster.h
+++ b/src/server/game/Movement/MotionMaster.h
@@ -24,6 +24,7 @@
 #include "SharedDefines.h"
 #include "Object.h"
 #include "MoveSplineInitArgs.h"
+#include "SplineChain.h"
 
 class MovementGenerator;
 class Unit;
@@ -53,7 +54,9 @@ enum MovementGeneratorType
     FOLLOW_MOTION_TYPE    = 14,
     ROTATE_MOTION_TYPE    = 15,
     EFFECT_MOTION_TYPE    = 16,
-    NULL_MOTION_TYPE      = 17
+    NULL_MOTION_TYPE      = 17,
+    SPLINE_CHAIN_MOTION_TYPE = 18,                          // SplineChainMovementGenerator.h
+    MAX_MOTION_TYPE                                         // limit
 };
 
 enum MovementSlot
@@ -196,6 +199,10 @@ class TC_GAME_API MotionMaster //: private std::stack<MovementGenerator *>
         void MoveCirclePath(float x, float y, float z, float radius, bool clockwise, uint8 stepCount);
         void MoveSmoothPath(uint32 pointId, G3D::Vector3 const* pathPoints, size_t pathSize, bool walk);
         void MoveSmoothPath(uint32 pointId, Movement::PointsArray const& points, bool walk);
+        // Walk along spline chain stored in DB (script_spline_chain_meta and script_spline_chain_waypoints)
+        void MoveAlongSplineChain(uint32 pointId, uint32 dbChainId, bool walk);
+        void MoveAlongSplineChain(uint32 pointId, SplineChain const& chain, bool walk);
+        void ResumeSplineChain(SplineChainResumeInfo const& info);
         void MoveFall(uint32 id = 0);
 
         void MoveSeekAssistance(float x, float y, float z);
--- a/src/server/game/Movement/MovementGenerators/SplineChainMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/SplineChainMovementGenerator.cpp
@@ -1,0 +1,140 @@
+/*
+* Copyright (C) 2008-2016 TrinityCore <http://www.trinitycore.org/>
+*
+* This program is free software; you can redistribute it and/or modify it
+* under the terms of the GNU General Public License as published by the
+* Free Software Foundation; either version 2 of the License, or (at your
+* option) any later version.
+*
+* This program is distributed in the hope that it will be useful, but WITHOUT
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+* FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+* more details.
+*
+* You should have received a copy of the GNU General Public License along
+* with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "SplineChainMovementGenerator.h"
+#include "Creature.h"
+#include "CreatureAI.h"
+#include "MoveSplineInit.h"
+#include "MoveSpline.h"
+#include "Log.h"
+
+uint32 SplineChainMovementGenerator::SendPathSpline(Unit* me, Movement::PointsArray const& wp) const
+{
+    uint32 numWp = wp.size();
+    ASSERT(numWp > 1 && "Every path must have source & destination");
+    Movement::MoveSplineInit init(me);
+    if (numWp > 2)
+        init.MovebyPath(wp);
+    else
+        init.MoveTo(wp[1], false, true);
+    init.SetWalk(_walk);
+    return init.Launch();
+}
+
+void SplineChainMovementGenerator::SendSplineFor(Unit* me, uint32 index, uint32& toNext)
+{
+    ASSERT(index < _chainSize);
+    TC_LOG_DEBUG("movement.splinechain", "%s: Sending spline for %u.", me->GetGUID().ToString().c_str(), index);
+
+    SplineChainLink const& thisLink = _chain[index];
+    uint32 actualDuration = SendPathSpline(me, thisLink.Points);
+    if (actualDuration != thisLink.ExpectedDuration)
+    {
+        TC_LOG_DEBUG("movement.splinechain", "%s: Sent spline for %u, duration is %u ms. Expected was %u ms (delta %d ms). Adjusting.", me->GetGUID().ToString().c_str(), index, actualDuration, thisLink.ExpectedDuration, int32(actualDuration) - int32(thisLink.ExpectedDuration));
+        toNext = uint32(double(actualDuration)/double(thisLink.ExpectedDuration) * toNext);
+    }
+    else
+    {
+        TC_LOG_DEBUG("movement.splinechain", "%s: Sent spline for %u, duration is %u ms.", me->GetGUID().ToString().c_str(), index, actualDuration);
+    }
+}
+
+void SplineChainMovementGenerator::Initialize(Unit* me)
+{
+    if (_chainSize)
+    {
+        if (_nextFirstWP) // this is a resumed movegen that has to start with a partial spline
+        {
+            if (finished)
+                return;
+            SplineChainLink const& thisLink = _chain[_nextIndex];
+            if (_nextFirstWP >= thisLink.Points.size())
+            {
+                TC_LOG_ERROR("movement.splinechain", "%s: Attempted to resume spline chain from invalid resume state (%u, %u).", me->GetGUID().ToString().c_str(), _nextIndex, _nextFirstWP);
+                _nextFirstWP = thisLink.Points.size()-1;
+            }
+            Movement::PointsArray partial(thisLink.Points.begin() + (_nextFirstWP-1), thisLink.Points.end());
+            SendPathSpline(me, partial);
+            TC_LOG_DEBUG("movement.splinechain", "%s: Resumed spline chain generator from resume state.", me->GetGUID().ToString().c_str());
+            ++_nextIndex;
+            if (!_msToNext)
+                _msToNext = 1;
+            _nextFirstWP = 0;
+        }
+        else
+        {
+            _msToNext = std::max(_chain[_nextIndex].TimeToNext, 1u);
+            SendSplineFor(me, _nextIndex, _msToNext);
+            ++_nextIndex;
+            if (_nextIndex >= _chainSize)
+                _msToNext = 0;
+        }
+    }
+    else
+    {
+        TC_LOG_ERROR("movement", "SplineChainMovementGenerator::Initialize - empty spline chain passed for %s.", me->GetGUID().ToString().c_str());
+    }
+}
+
+void SplineChainMovementGenerator::Finalize(Unit* me)
+{
+    if (!finished)
+        return;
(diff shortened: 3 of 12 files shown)


src/server/game/Movement/Spline/MovementPacketBuilder.cpp
```diff
@@ -98,7 +98,7 @@ namespace Movement
 
     void WriteLinearPath(const Spline<int32>& spline, ByteBuffer& data)
     {
-        uint32 last_idx = spline.getPointCount() - 3;
+        uint32 last_idx = spline.getPointCount() - 2;
```
Treeston: This one is something that I stumbled across when reviewing existing spline logic, and something that I couldn't understand. I'd love some input from @Shauren and other people that were around when splines were first implemented.

The way I see it, `.getPointCount()` and `.getPoint(i)` are simply wrappers for `.size()` and `[i]` on the underlying points array. The first point is the current location, and the last point is the spline endpoint. `last_idx` should be set to the index of the endpoint relative to the second point (see definition of `real_path` just below this comment).

Obviously, the last element of the array will have index `.size() - 1` in relation to the first element (`&spline.getPoint(0)`). In relation to the second element (`&spline.getPoint(1)`) it will have index `.size() - 2`.

However, the current core sends the point with relative index `.size() - 3` as the spline endpoint instead, and I have no explanation for this. This leads to the final point of any path being silently discarded by the core, never reaching the client. This change compiles and runs fine without any issue.

I really have no idea what the thought process behind the current logic is. Is there anything I'm not getting here?

Warpten: The way that spline ends is used by the client to determine if the path is linear or not. Etc.

Shauren: The old code is correct, client magic, do not touch. These ending points are used serverside for catmullrom interpolation only and should not be sent

Treeston: @Shauren Are any of those auto-generated points actually sent to the client?
Say I have a movespline like this in a sniff:
```
Position: X: 1607.088 Y: 802.5474 Z: 122.7526
(MovementMonsterSpline) (MovementSpline) Flags: 1048576 (Unknown5)
(MovementMonsterSpline) (MovementSpline) [0] Points: X: 1613.629 Y: 795.9061 Z: 121.8324
(MovementMonsterSpline) (MovementSpline) [0] WayPoints: X: 1609.108 Y: 799.9767 Z: 122.7925
(MovementMonsterSpline) (MovementSpline) [1] WayPoints: X: 1609.608 Y: 799.2267 Z: 122.7925
(MovementMonsterSpline) (MovementSpline) [2] WayPoints: X: 1612.608 Y: 796.9767 Z: 122.2925
```
which of these points should be passed to `MovebyPath` to replicate this packet in TC? Right now I pass the WayPoints in order, followed by the endpoint (the single Points entry). Is this correct?

Shauren: You just pass all points you see in the packet (and final destination)
First and last points in spline are internal generated points (yes, catmullrom init is used for linear splines too!), that makes it -2 points to send. On top of these points one is sent separately (destination), making it -3

Treeston:
Right, so, here's what this test case illustrates: the core's duration for a movespline currently does not match the client's. In the current test case, this is what the retail servers send for the first spline in the set:

```
Position: X: 1613.629 Y: 795.9061 Z: 121.8324
(MovementMonsterSpline) Id: 1652
(MovementMonsterSpline) Destination: X: 0 Y: 0 Z: 0
(MovementMonsterSpline) CrzTeleport: False
(MovementMonsterSpline) StopDistanceTolerance: 0
(MovementMonsterSpline) (MovementSpline) Flags: 1048576 (Unknown5)
(MovementMonsterSpline) (MovementSpline) AnimTier: 0
(MovementMonsterSpline) (MovementSpline) TierTransStartTime: 0
(MovementMonsterSpline) (MovementSpline) Elapsed: 0
(MovementMonsterSpline) (MovementSpline) MoveTime: 4320
(MovementMonsterSpline) (MovementSpline) JumpGravity: 0
(MovementMonsterSpline) (MovementSpline) SpecialTime: 0
(MovementMonsterSpline) (MovementSpline) Mode: 0
(MovementMonsterSpline) (MovementSpline) VehicleExitVoluntary: 0
(MovementMonsterSpline) (MovementSpline) TransportGUID: Full: 0x0
(MovementMonsterSpline) (MovementSpline) VehicleSeat: -1
(MovementMonsterSpline) (MovementSpline) Face: 0
(MovementMonsterSpline) (MovementSpline) PointsCount: 1
(MovementMonsterSpline) (MovementSpline) PackedDeltasCount: 2
(MovementMonsterSpline) (MovementSpline) HasSplineFilter: False
(MovementMonsterSpline) (MovementSpline) HasSpellEffectExtraData: False
(MovementMonsterSpline) (MovementSpline) [0] Points: X: 1623.677 Y: 799.6086 Z: 120.5017
(MovementMonsterSpline) (MovementSpline) [0] WayPoints: X: 1619.153 Y: 798.0073 Z: 121.4171
(MovementMonsterSpline) (MovementSpline) [1] WayPoints: X: 1622.903 Y: 799.2573 Z: 120.9171
```
Note the duration, which is `4320ms`.

If we start in the same location on TC and create a spline using the three points in the sniff, we instead end up with a spline duration of `4351ms`, 31ms longer than expected. This causes a desync which adds up on subsequent movespline creations, causing stuttering and moonwalking behavior as the client tries to cope.

Any input is appreciated at this point.

Treeston:
Alright, DB implementation done. This now works!

Check it out at `.go creature id 27891`, fully sourced from DB!

Treeston:
If you want to parse data from a sniff, [this parser](https://gist.github.com/Treeston/6d189e72d1085d25f7d7a642602f295d) parses spline chains from 6.x/7.x era sniffs (WPP output format sensitive).

Treeston:
At this point, this is merge ready. Expeditious code review is appreciated.

src/server/game/Movement/MovementGenerators/SplineChainMovementGenerator.cpp
```diff
@@ -0,0 +1,140 @@
...
+void SplineChainMovementGenerator::SendSplineFor(Unit* me, uint32 index, uint32& toNext)
+{
+    ASSERT(index < _chainSize);
+    TC_LOG_DEBUG("movement.splinechain", "%s: Sending spline for %u.", me->GetGUID().ToString().c_str(), index);
+
+    SplineChainLink const& thisLink = _chain[index];
+    uint32 actualDuration = SendPathSpline(me, thisLink.waypoints);
+    if (actualDuration != thisLink.expectedDuration)
+    {
+        TC_LOG_DEBUG("movement.splinechain", "%s: Sent spline for %u, duration is %u ms. Expected was %u ms (delta %d ms). Adjusting.", me->GetGUID().ToString().c_str(), actualDuration, thisLink.expectedDuration, int32(actualDuration) - int32(thisLink.expectedDuration));
```
tkrokli:
```
src/server/game/Movement/MovementGenerators/SplineChainMovementGenerator.cpp:47:117: fatal error: 
      more '%' conversions than data arguments [-Wformat]
  ...for %u, duration is %u ms. Expected was %u ms (delta %d ms). Adjusting."...
                                                          ~^
src/common/Logging/Log.h:199:56: note: 
      expanded from macro 'TC_LOG_DEBUG'
    TC_LOG_MESSAGE_BODY(filterType__, LOG_LEVEL_DEBUG, __VA_ARGS__)
                                                       ^
src/common/Logging/Log.h:179:32: note: 
      expanded from macro 'TC_LOG_MESSAGE_BODY'
                    check_args(__VA_ARGS__);                            \
                               ^
```

tkrokli:
Not quite sure where in the file culling_of_stratholme.cpp the error from AppVeyor belongs:
```
culling_of_stratholme.obj : error LNK2019: unresolved external symbol
 "public: struct SplineChainResumeInfo __cdecl SplineChainMovementGenerator::GetResumeInfo(class Unit const *)const "
 (?GetResumeInfo@SplineChainMovementGenerator@@QEBA?AUSplineChainResumeInfo@@PEBVUnit@@@Z)

 referenced in function
 "public: virtual void __cdecl debugscript_splinechains_malcolm::ai::UpdateAI(unsigned int)"
 (?UpdateAI@ai@debugscript_splinechains_malcolm@@UEAAXI@Z)

[C:\projects\trinitycore\build\src\server\scripts\scripts_kalimdor.vcxproj]

C:\projects\trinitycore\build\bin\RelWithDebInfo\scripts\scripts_kalimdor.dll :
 fatal error LNK1120: 1 unresolved externals 

[C:\projects\trinitycore\build\src\server\scripts\scripts_kalimdor.vcxproj]
```

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -468,6 +470,29 @@ void MotionMaster::MoveSmoothPath(uint32 pointId, Movement::PointsArray const& p
...
+}
+
+void MotionMaster::MoveAlongSplineChain(uint32 pointId, SplineChain const& chain, bool walk)
+{
+    Mutate(new SplineChainMovementGenerator(pointId, chain, walk), MOTION_SLOT_CONTROLLED);
+}
+
+void MotionMaster::ResumeSplineChain(SplineChainResumeInfo const& info)
+{
+    Mutate(new SplineChainMovementGenerator(info), MOTION_SLOT_CONTROLLED);
```
ccrs: so you take the info from the current movegen to create the new one in the future, based on how many points are left, k
but wouldn't it be better to simply add a hook creature->motionmaster->movegen to stop/resume movement on demand?

src/server/game/Movement/MovementGenerators/SplineChainMovementGenerator.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include "Creature.h"
+#include "CreatureAI.h"
+#include "MoveSplineInit.h"
+#include "MoveSpline.h"
+#include "Log.h"
+
+uint32 SplineChainMovementGenerator::SendPathSpline(Unit* me, Movement::PointsArray const& wp) const
+{
+    uint32 numWp = wp.size();
+    ASSERT(numWp > 1 && "Every path must have source & destination");
```
Treeston: If you send a wrong spline, your dev core crashes and tells you where you messed up. Hard to miss that.

src/server/game/Movement/MotionMaster.h
```diff
@@ -53,7 +54,8 @@ enum MovementGeneratorType
     FOLLOW_MOTION_TYPE    = 14,
     ROTATE_MOTION_TYPE    = 15,
     EFFECT_MOTION_TYPE    = 16,
-    NULL_MOTION_TYPE      = 17
+    NULL_MOTION_TYPE      = 17,
+    SPLINE_CHAIN_MOTION_TYPE = 18                           // SplineChainMovementGenerator.h
```
Treeston: :legacycode:

src/server/game/Movement/MotionMaster.cpp
Treeston: Movegen might get overridden and lose the info. This is more powerful.

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -468,6 +470,29 @@ void MotionMaster::MoveSmoothPath(uint32 pointId, Movement::PointsArray const& p
...
+    {
+        TC_LOG_ERROR("misc", "MotionMaster::MoveAlongSplineChain: creature with entry %u tried to walk along non-existing spline chain with DB id %u.", owner->GetEntry(), dbChainId);
+        return;
+    }
+    MoveAlongSplineChain(pointId, chain, walk);
+}
+
+void MotionMaster::MoveAlongSplineChain(uint32 pointId, SplineChain const& chain, bool walk)
+{
+    Mutate(new SplineChainMovementGenerator(pointId, chain, walk), MOTION_SLOT_CONTROLLED);
```
Treeston: explain the difference to me? CONTROLLED is stuff like polymorph I guess, that should override?

src/server/game/Movement/MotionMaster.h
```diff
@@ -196,6 +198,10 @@ class TC_GAME_API MotionMaster //: private std::stack<MovementGenerator *>
         void MoveCirclePath(float x, float y, float z, float radius, bool clockwise, uint8 stepCount);
         void MoveSmoothPath(uint32 pointId, G3D::Vector3 const* pathPoints, size_t pathSize, bool walk);
         void MoveSmoothPath(uint32 pointId, Movement::PointsArray const& points, bool walk);
+        // Walk along spline chain stored in DB (script_spline_chain_meta and script_spline_chain_waypoints)
+        void MoveAlongSplineChain(uint32 pointId, uint32 dbChainId, bool walk);
+        void MoveAlongSplineChain(uint32 pointId, SplineChain const& chain, bool walk);
```
Treeston: On one hand, yes - on the other hand, we can't (easily) parse that from sniffs, while current spline data is fully sniff parsable.

Also, I haven't seen any case where creatures switch between walk and run within a single spline chain. Not sure how client would even handle it.

src/server/game/Movement/MotionMaster.cpp
Treeston: guess it should really be IDLE

src/server/game/Movement/MotionMaster.h
Treeston: fixed in latest

src/server/game/Movement/MotionMaster.cpp
Treeston: changed to IDLE

src/server/game/Movement/MotionMaster.h
Treeston: Or is there some way to get walk/run from WPP output?

src/server/game/Movement/MotionMaster.cpp
ccrs: I'd say safer, but feels like a overkill to me xd

src/server/game/Movement/MovementGenerators/SplineChainMovementGenerator.cpp
ccrs: the point was, not to have a crash because of a silly dev xd

Treeston: They'd instantly notice it in their local tests. Remember, spline chains are static data, they (generally) don't change on every iteration.

src/server/game/Movement/MotionMaster.cpp
Treeston: waypoint is IDLE

src/server/game/Movement/MotionMaster.cpp
Treeston: intended use is "enter combat, pause splinechain, store info, on exit combat, leash then resume once home reached".

Shauren:
What is the main difference between this and regular waypoint tables and why can't that be used?
This looks like a FIFTH way to handle waypoints
* `waypoints`
* `waypoint_data`
* `script_waypoint`
* `smart_scripts`
* this

src/server/game/Movement/MotionMaster.cpp
Treeston: IDLE does weird things though where it doesn't properly yield on motionmaster clear.

Bumping back up to ACTIVE.

Treeston:
Regular waypoint tables work as a sequence of movepoints, using mmaps pathfinding as interpolation between them. This is great for relatively dynamic content, as well as easy to implement for a creature - not a lot of sniff data is needed to get an accurate waypoint path using only corner vertices.

However, this ends up with all the issues that come with mmaps pathfinding - mobs clipping through the floor, jumping through the air, cutting corners and similar - that happen because mmaps pathfinding doesn't match what the client expects.

This new movement system is _only_ workable for static situations (where a creature moves in a fully pre-determined manner, e.g. RP sequences) - but in those situations, it provides a retail-like movespline output, causing the scenes to play fluently, without any movement weirdness whatsoever.

Treeston:
PS: I'd argue this could also be used as an improvement over the `script_waypoint` system for escort quests if enough sniff data is available to construct a full spline chain.

Kittnz:
Isn't there already enough data to construct the path?

Treeston:
@Kittnz ?

Treeston:
This has two things that current waypoint systems don't have:

* It stores full spline waypoint sets, not just the endpoint. This means no mmaps pathing is done.
* It doesn't wait for the creature to reach the endpoint of the spline before "chaining" the next spline, causing the client to smoothly animate the transition (as opposed to move-stop-move-stop-move-stop behavior like with current WP).

Kittnz:
Maybe i should have asked what exactly i need to construct the path.

Treeston:
Sniff data. Spline chains come straight from sniffs.

It's pretty much the concept "to get retail-like pathing, you just need to use retail spline waypoints".

Shauren:
Why not improve `script_waypoint` then?

Treeston:
It's a completely different concept. `script_waypoint` is mmaps+sequential movepoints, spline chaining is full spline data.

Existing script_waypoint data isn't compatible with this system.

ccrs:
so... instead of sending mmap created paths correctly, create a new movement with not calculated paths?

Treeston:
Since current mmaps doesn't match what client expects? yes.

ccrs:
dont get me wrong, I get that the solution is nice to have instant results
and on top of that pathfinding is just a tc depedency.
ironically the problem is not pathfinding, but mmaps creation xd

Treeston:
Yes, but obviously Blizzard uses a different (better? arguable) pathing solution, very likely either homebrew or proprietary. Our mmaps output will _never_ match it 100%.

For RP sequences, having a 100% match is very much a visual necessity.

ccrs:
yeah, seems necessary
the focus is not here then, but to erase those systems which are redundant
but as always, thats another matter xd

Treeston:
Yeah, I definitely agree that some of the old systems are redundant with each other and should be slimmed down.

However, this PR constitutes an entirely different approach to path movement that is distinct from the existing (movepoint-based) systems, not related to that.

Shauren (changes requested):

BAndysc:
This was really nice idea and works flawlessly!

PR #17948, opened 2016-09-15
Script/Quest: improvements for Let Them Eat Crow.

### LINKED ISSUE #16737: [3.3.5][6.x] DB/Quest: Let Them Eat Crow
**Description**: 
The quest itself works and is detachable but the Plaguehound is buggy and could also make perfomance problems.

In the Quest it comes that kills you crows to get their meat.
This meat (5stk) you feed the dog.
Now comes the problem: One can the dog after 2sec again summoning and the old remains. So you can theoretically spawn 100p before it disappears.
As far as I have in mind pursued this dog a and when you summon a new course, should the old disappearance

**Branch(es)**: 335

**TC hash/commit**: ec62a02f5763b7ed39a9441f5989e109c72e4c2a

**TDB version**: 2016_03_04_03_world.sql

**Operating system**: Linux

SnapperRy:
**Changes proposed:**
- Improve the quest [Let Them Eat Crow](http://www.wowhead.com/quest=11227/let-them-eat-crow) by summoning the Hungry Plaguehound as pet, with the Sic'em spell useable to aggro Fjord Crows.

I can't find any pattern in SummonProperties.dbc to give this poor npc a pet bar, hence why I change the summon type in SpellEffects.cpp.
This can also be used for other creatures that should be summoned as pets yet don't have the pet category/type already set.

**Target branch(es):** 3.3.5

**Issues addressed:** Closes #16737

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -2185,7 +2185,7 @@ void Spell::EffectSummonType(SpellEffIndex effIndex)
     if (!entry)
         return;
 
-    SummonPropertiesEntry const* properties = sSummonPropertiesStore.LookupEntry(m_spellInfo->Effects[effIndex].MiscValueB);
+    SummonPropertiesEntry* properties = const_cast<SummonPropertiesEntry*>(sSummonPropertiesStore.LookupEntry(m_spellInfo->Effects[effIndex].MiscValueB));
     if (!properties)
     {
         TC_LOG_ERROR("spells", "EffectSummonType: Unhandled summon type %u.", m_spellInfo->Effects[effIndex].MiscValueB);
@@ -2230,6 +2230,15 @@ void Spell::EffectSummonType(SpellEffIndex effIndex)
         default:
             numSummons = 1;
             break;
+    }
+
+    // Some creatures must be summoned as pets, yet have a different category or type set in SummonProperties.dbc.
+    // Until a generic way is found to handle them, replace the existing values.
+    switch (properties->Id)
+    {
+    case 628: // Hungry Plaguehound
+        properties->Category = SUMMON_CATEGORY_PET;
+        break;
     }
 
     switch (properties->Category)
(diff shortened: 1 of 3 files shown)


src/server/game/Spells/SpellEffects.cpp
```diff
@@ -2232,6 +2232,15 @@ void Spell::EffectSummonType(SpellEffIndex effIndex)
             break;
     }
 
+    // Some creatures must be summoned as pets, yet have a different category or type set in SummonProperties.dbc.
+    // Until a generic way is found to handle them, replace the existing values.
+    switch (properties->Id)
+    {
+    case 628: // Hungry Plaguehound
+        properties->Category = SUMMON_CATEGORY_PET;
```
jackpoz: is this threadsafe ?

SnapperRy: Absolutely no idea, sorry. I see no reason it would cause any issue, though, as this happens before the creature itself is created.

jackpoz: Better get an idea about the TrinityCore threading model (WordUpdateLoop, Map::Update, MapInstance::Update, WorldSession::Update) to ensure no race conditions are added. In this case it seems you are modifying a singleton field every time Spell::EffectSummonType() is called for a particular summon, technically triggering a race condition if it is possible for multiple threads to call this function at same time (leaving aside for a moment the technicality of 2 threads writing the exact same value).

sirikfoll: imo, this fix belongs to SpellMgr.cpp, see https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellMgr.cpp#L3821

SnapperRy: You're right, I really missed that. Thanks!

Treeston:
Wouldn't your current conditions setup cause two players to potentially feed each others' Plaguehounds?

SnapperRy:
@Treeston: it would. It's the same for the [Venomhide Hatchling](http://www.wowhead.com/item=46362) and other quests like that handled via conditions, as we can't (yet) target the player's pet like that.

Does the same happen or retail, or does it only work for your own Plaguehound (if anyone wants to check :P)?

PR #17952, opened 2016-09-15
Core/Transport: always update position of non-static passengers

SnapperRy:
**Changes proposed:**

Currently, the position of players inside transports is updated quite accurately when the transport is moving. However, this is not the case when the transport is not moving. Somehow, when the transport's waypoint movement is stopped, the player's position desyncs from the client.

The best way to check this is the quest [Mutiny on the Mercy](http://www.wowhead.com/quest=11527/mutiny-on-the-mercy). On the ship's lower levels, there are hostile mobs. When the player is standing still or the ship is moving, they can attack and be attacked by the player just fine. But if the ship is standing still and the player moves, the mobs will reset and evade, unable to reach the player.

This fix greatly improves the behavior in such cases. I'm not sure if it's the proper way™ to fix it, but it's a start if anyone knows enough about the transport system to look into it.

**Target branch(es):** 3.3.5

### DIFF
--- a/src/server/game/Entities/Transport/Transport.cpp
+++ b/src/server/game/Entities/Transport/Transport.cpp
@@ -201,6 +201,8 @@ void Transport::Update(uint32 diff)
     if (_positionChangeTimer.Passed())
     {
         _positionChangeTimer.Reset(positionUpdateDelay);
+        UpdatePassengerPositions(_passengers);
+
         if (IsMoving())
         {
             float t = CalculateSegmentPos(float(timer) * 0.001f);


tkrokli:
Interesting solution. Will it have any positive effect on the NPCs on the airships in Icecrown? :-)

SnapperRy:
I don't think it will affect them, as the Icecrown ships are always moving. I'm going to check anyway :P

tkrokli:
OK, good point. I presume it won't be affecting that case, then. Thank you for checking it, though.

SnapperRy:
http://www.wowhead.com/npc=32302/knight-captain-drosche this npc seems to get less stuck than before, but I'm not sure if it's due to this change or if it was already like that. I remember it being hard to talk with.

I'm not familiar with any other ship bugs, seems fine to me though :P

PR #17959, opened 2016-09-16
Created Old SQL Merger For ALL Old TDB's

alexandre433:
**Changes proposed:**
-  Adding this whould help loots of people who have older TDB's installed and it's easier to update their db's to have more fixes on their own compilations.
- Created a file which was modified, which creates all .sql's of the older TDB's.

**Target branch(es):** 3.3.5

**Tests performed:** Tested and works!


tkrokli:
You need to put your suggested patch in the correct branch of your own repository. This won't work.

PR #17960, opened 2016-09-16
Created Old SQL Merger For ALL Old TDB's

alexandre433:
**Changes proposed:**
-  Adding this whould help loots of people who have older TDB's installed and it's easier to update their db's to have more fixes on their own compilations.
- Created a file which was modified, which creates all .sql's of the older TDB's.

**Target branch(es):** 3.3.5

**Tests performed:** Tested and works!


DDuarte:
This isn't needed since the worldserver can automatically update the DBs.

alexandre433:
This is for OLD TDB's not new ones.

The old one's can't update themselves.

Treeston:
This isn't cross platform, and updating DBs that old is more effort than just getting latest TDB. Fixes from that long ago aren't going to work in the latest core anyway.

Treeston:
(Feel free to put it on the forums though.)

PR #17986, opened 2016-09-23
Core/World: properly re-initialize daily quest reset time when resetting daily quests.

SnapperRy:
**Changes proposed:**

Right now, the server adds one day to the daily quest reset timer after it has been processed.

This assumes that the timer is always set to the proper daily reset time (6:00 AM). However, this is not always the case.
If the server is shut down before the reset time and restarted after the reset time, the timer is set to the time the last daily quest has been completed, which can be at any hour of the day.

Examples: if a daily quest is completed at 5:00 AM, and the server is shut down to be restarted at 7:00 AM, the timer would be equal to 5:00 AM. Adding one day to that means the next reset will not be at 6:00 AM as it's supposed to be.

**Target branch(es):** 3.3.5

- [x] I'm not a hundred percent sure that it's "safe" for it to execute a SELECT query in `InitDailyQuestResetTime()` right after the DELETE in `ResetDailyQuests()`. Should it skip the SELECT if the server is not loading? Looking for advice here.

### DIFF
--- a/src/server/game/World/World.cpp
+++ b/src/server/game/World/World.cpp
@@ -2103,7 +2103,7 @@ void World::Update(uint32 diff)
     if (m_gameTime > m_NextDailyQuestReset)
     {
         ResetDailyQuests();
-        m_NextDailyQuestReset += DAY;
+        InitDailyQuestResetTime();
     }
 
     /// Handle weekly quests reset time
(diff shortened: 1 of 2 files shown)


ariel-:
If both queries are synch (thread blocking), you'll have no problem executing a select after a delete (data will be consistent)

SnapperRy:
Uhm. DELETE query uses Execute() which is async. It should probably stay that way. Going to make it skip the SELECT when server is loading.

PR #17989, opened 2016-09-23
Core/Battlegrounds: Fix lag caused by arena points distribution

ibakula:
Distribution of arena points caused worldserver lag spikes.
Arena teams list contained 4900 items. (ie the table contains 4927 arena teams)
The prepared statements contained large amount of queries and blocked the thread.
This seemed to be working out well when there are plenty inactive teams.

### DIFF
--- a/src/server/game/Battlegrounds/ArenaTeam.cpp
+++ b/src/server/game/Battlegrounds/ArenaTeam.cpp
@@ -899,6 +899,10 @@ void ArenaTeam::SaveToDB()
 
     for (MemberList::const_iterator itr = Members.begin(); itr != Members.end(); ++itr)
     {
+        // Save the effort and go
+        if (itr->WeekGames == 0)
+            continue;
+
         stmt = CharacterDatabase.GetPreparedStatement(CHAR_UPD_ARENA_TEAM_MEMBER);
         stmt->setUInt16(0, itr->PersonalRating);
         stmt->setUInt16(1, itr->WeekGames);
@@ -919,8 +923,12 @@ void ArenaTeam::SaveToDB()
     CharacterDatabase.CommitTransaction(trans);
 }
 
-void ArenaTeam::FinishWeek()
+bool ArenaTeam::FinishWeek()
 {
+    // No need to go further than this
+    if (Stats.WeekGames == 0)
+        return false;
+
     // Reset team stats
     Stats.WeekGames = 0;
     Stats.WeekWins = 0;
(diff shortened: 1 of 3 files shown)


src/server/game/Battlegrounds/ArenaTeam.cpp
```diff
@@ -899,6 +899,10 @@ void ArenaTeam::SaveToDB()
 
     for (MemberList::const_iterator itr = Members.begin(); itr != Members.end(); ++itr)
     {
+        // Save the effort and go
+        if (itr->WeekGames == 0)
```
jackpoz: is this enough to ensure no other variables changed ?
which ArenaTeam values are still needed to be saved even if nothing changed ?

Aokromes: @ibakula

ibakula (commented):
@Aokromes @jackpoz 
It is enough.
Other variable values are intact, I have responded before but my comments show as "pending".

src/server/game/Battlegrounds/ArenaTeamMgr.cpp
```diff
@@ -186,8 +186,9 @@ void ArenaTeamMgr::DistributeArenaPoints()
     {
         if (ArenaTeam* at = titr->second)
         {
-            at->FinishWeek();
-            at->SaveToDB();
+            if (at->FinishWeek())
+                at->SaveToDB();
+            
             at->NotifyStatsChanged();
```
jackpoz: should this be moved inside the "if (at->FinishWeek())" check too ?

ibakula: I think its fine as the function will only let them know that the flush has been completed, even if their score is intact.

ibakula: What I meant to say is, it could be done as following:
`if (!at->FinishWeek())
     continue;
at->SaveToDB();
at->NotifyStatsChanged();`

Because we don't really have to notify the members of their stats as there was no alterations made from their side as they didn't play this week but I rather chose to inform them as that's not the underlying problem that freezes the thread.

My commit should be accepted, imo.

jackpoz:
Merging this PR hoping the author tested thoroughly

Kittnz:
lol @jackpoz (Quote: Is this even tested before it was merged?) xD

jackpoz:
Well from my comment it's very clear I merged it without testing it so none has to wonder if it was or wasn't tested :) 
it is safe to assume it will cause issues, unless the author did a very good job and made sure it worked already :) if that's not the case I'll add the author to my personal watchlist of dangerous (to the code) authors

PR #18001, opened 2016-09-25
Core/Entity : Conversation

### LINKED ISSUE #17992: Core/Entity : new dialogs in WoW Legion
**Description**:

Hello, I am trying to recreate the initial mission of the hunters of demons, someone has some idea of since it could do http://oi68.tinypic.com/x5umtc.jpg
What in warcraft the IIIrd was a dialog, now my question is, for this type of dialogs it is a necessary algun script in lua or in c ++

**Current behaviour**: not work

--removed--

1.It is necessary opcodes, or is just a matter of scrips

**Tags 7.0.3/22594**

Edited by DDuarte

Traesh:
**Changes proposed:**
-  Implement Conversation Entity, used to display ingame dialog
-  Command .debug conversation # convId
-  Command .Reload conversation_template
- Add 5 tables for templates, actors & lines

**Target branch(es):** 6.x

**Issues addressed:** Closes #17992

Built
Tested ingame with conversation id 705    (.debug conv 705)
Tested ingame with conversation id 1025  (.debug conv 1025)

- None ATM


src/server/game/Entities/Conversation/Conversation.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    if (!conversationTemplate)
+        return false;
+
+    SetUInt32Value(CONVERSATION_FIELD_LAST_LINE_DURATION, conversationTemplate->LastLineDuration);
+    _duration = (conversationTemplate->Lines.size() - 1) * 10000 + conversationTemplate->LastLineDuration; // Estimated duration, need further investigation
+
+    for (ConversationActorTemplate actor : conversationTemplate->Actors)
+    {
+        AddDynamicValue(CONVERSATION_DYNAMIC_FIELD_ACTORS, actor.Id);
```
Shauren: If each actor requires several entries in the updatefield then use "structured dynamic values" - look at how item gems and artifact powers are implemented

src/server/game/Entities/Object/Object.cpp
```diff
@@ -815,7 +816,7 @@ void Object::BuildDynamicValuesUpdate(uint8 updateType, ByteBuffer* data, Player
             data->resize(data->size() + arrayBlockCount * sizeof(UpdateMask::BlockType));
             for (std::size_t v = 0; v < values.size(); ++v)
             {
-                if (updateType == UPDATETYPE_VALUES ? _dynamicChangesArrayMask[index][v] : values[v])
+                if (updateType != UPDATETYPE_VALUES || _dynamicChangesArrayMask[index][v])
```
Shauren: I don't like this change - messing with code here affects every object type, generally only non zero values are sent on create

akrom23:
How about sceneobjects? :)

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9634,3 +9634,115 @@ void ObjectMgr::LoadSceneTemplates()
 
     TC_LOG_INFO("server.loading", ">> Loaded %u scene templates in %u ms.", count, GetMSTimeDiffToNow(oldMSTime));
 }
+
+void ObjectMgr::LoadConversationTemplates()
```
Shauren: All queries here are used only in this method and only during startup, there is no need to make them prepared statements (reloading is seen as development purpose, should not affect deciding whether to use a prepared statement or not)

src/server/game/Globals/ObjectMgr.h
```diff
@@ -1414,7 +1420,6 @@ class TC_GAME_API ObjectMgr
 
             return nullptr;
         }
-
```
Shauren: this empty line should be here

sql/updates/world/6.x/9999_99_99_00_world.sql
```diff
@@ -0,0 +1,72 @@
+CREATE TABLE IF NOT EXISTS `conversation_template` (
```
Shauren: Write these as DROP TABLE IF EXISTS + CREATE TABLE

sql/updates/world/6.x/9999_99_99_00_world.sql
```diff
@@ -0,0 +1,72 @@
...
+CREATE TABLE IF NOT EXISTS `conversation_lines` (
+  `ConversationId` INT (10) UNSIGNED NOT NULL,
+  `ConversationLineId` INT (10) NOT NULL,
+  `Idx` INT (10) UNSIGNED NOT NULL DEFAULT 0,
+  `VerifiedBuild` INT (10) NOT NULL DEFAULT 0,
+  PRIMARY KEY (`ConversationId`, `ConversationLineId`)
+) ENGINE=MyISAM DEFAULT CHARSET=utf8;
+
+
+DELETE FROM conversation_actor_template WHERE Id IN (49825, 50124);
```
Shauren: since the purpose of this sql is to create tables, you can omit deletes

sql/updates/world/6.x/9999_99_99_00_world.sql
```diff
@@ -0,0 +1,72 @@
...
+INSERT INTO `conversation_lines`(`ConversationId`,`ConversationLineId`,`Idx`,`VerifiedBuild`) VALUES
+(705, 1683, 0, 22624),
+(705, 1684, 1, 22624),
+(1025, 1445, 0, 22624),
+(1025, 1446, 1, 22624);
+
+DELETE FROM conversation_template WHERE Id IN (705, 1025);
+INSERT INTO `conversation_template`(`Id`,`LastLineDuration`,`VerifiedBuild`) VALUES
+(705, 6544, 22624),
+(1025, 3458, 22624);
```
Shauren: You are adding new commands, this sql should also insert data to `command` table with help text

Shauren (changes requested):

src/server/game/Entities/Object/Object.cpp
Traesh: I understand, but in my sniffs there is a lot of DYNAMIC_FIELD send even if value is 0 at creation.

Conversation does not work if 0 value is not sent, so this is mandatory, but of course there is perhaps a better way to do it

src/server/game/Entities/Conversation/Conversation.cpp
```diff
@@ -0,0 +1,131 @@
...
+    _dynamicValuesCount = CONVERSATION_DYNAMIC_END;
+}
+
+Conversation::~Conversation()
+{
+}
+
+void Conversation::AddToWorld()
+{
+    ///- Register the AreaTrigger for guid lookup and for caster
```
Shauren: that comment copypaste

Traesh: Don't know what you're talking about, i'm innocent

Traesh:
Now at #18010 due to a fail on previous repo

PR #18021, opened 2016-09-28
Core/Spell: set quest to rewarded instead of complete when targeted by SPELL_EFFECT_QUEST_COMPLETE.

SnapperRy:
**Changes proposed:**

Title says it all.

Example of an issue with current behavior:
- Take the quest [Arm the Wards!](http://www.wowhead.com/quest=11523/arm-the-wards), and add four [Mana Remnants](http://www.wowhead.com/item=34338/mana-remnants) to yourself.
- Use the Mana Remnants near the crystal in Isle of Quel'Danas.
- Relog.
- Notice that in your quest log [this ugly quest](http://www.wowhead.com/quest=11496/the-sanctum-wards) is now staring defiantly back at you, even though you never accepted it (and it's a deprecated quest, to boot).

**Target branch(es):** 3.3.5

**Issues addressed:** Fixes most (all?) cases where players find strange quests in their quest logs.

**Tests performed:** tested 'n' working.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -15055,10 +15055,7 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver, 
 
     RemoveActiveQuest(quest_id, false);
     if (quest->CanIncreaseRewardedQuestCounters())
-    {
-        m_RewardedQuests.insert(quest_id);
-        m_RewardedQuestsSave[quest_id] = QUEST_DEFAULT_SAVE_TYPE;
-    }
+        SetRewardedQuest(quest_id);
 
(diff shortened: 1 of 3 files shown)


Traesh:
Wouldn't it be better to simply ignore the spell if quest isn't in quest log ?

Initial comment say "for internal use", maybe this should just have been dropped, is there a case were this code is usefull ?

Shauren:
@Traesh no. These quests need to be set to rewarded. To give you an example, http://www.wowhead.com/spell=217435/illusion-nightmare - you "learn" enchant illusions with effects like that, then the client only checks all existing illusions and locks the ones you don't meet the PlayerCondition for (the condition is "complete quest"). Known illusion list is not saved/sent anywhere explicitly.

Traesh:
Indeed, my bad, thanks for the infos

Issue #16937, opened 2016-04-11
[6.x][Death Knight] Core/Spells: Defile talent


Ulduar:
**Description**:
http://www.wowhead.com/spell=152280
Defile.
**Current behaviour**: Tell us what happens
Not working.
**Expected behaviour**: Tell us what should happen instead

**Steps to reproduce the problem**:

1. learn spell 152280
2. .cast spell 152280
3. Not Working

**Branch(es)**: 6x

**TC hash/commit**:
https://github.com/TrinityCore/TrinityCore/commit/348eac096dce4e939fe31cd2d7fa698bf1fe602d
**TDB version**:
6.x
**Operating system**:

[//]: # (This template is for problem reports, for other type of reports edit it accordingly)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)

Spell Work Data:
``` CPP
ID - 152280 Осквернение
=================================================
Description: Оскверняет выбранное рыцарем смерти место. Раз в $t3 сек., если в зоне поражения "Осквернения" есть противники, заклинание наносит им $156000s1 ед. урона от темной магии льда, а зона его действия и наносимый им урон увеличиваются на ${$m2/100}.1%. $?c1[

Противники, находящиеся в зоне действия "Осквернения", наносят вам на $s4% меньше урона.][]

Заменяет "Смерть и разложение".

Category = 0, SpellIconID = 9751, activeIconID = 0, SpellVisual = (36630,0)
Family SPELLFAMILY_DEATHKNIGHT, flag [0] 0x00000020 [1] 0x00000000 [2] 0x00000000 [3] 0x00000000

SpellSchoolMask = 48 (SPELL_SCHOOL_MASK_FROST, SPELL_SCHOOL_MASK_SHADOW)
DamageClass = 1 (SPELL_DAMAGE_CLASS_MAGIC)
PreventionType = 5 (5)
=================================================
AttributesEx1: 0x10000000 (SPELL_ATTR1_DONT_DISPLAY_IN_AURA_BAR)
AttributesEx5: 0x04000000 (SPELL_ATTR5_UNK26)
AttributesEx8: 0x00002000 (SPELL_ATTR8_UNK13)
AttributesEx13: 0x00100000 (SPELL_ATTR13_UNK20)
=================================================
Targets Mask = 0x00000040 (TARGET_FLAG_DEST_LOCATION)
Spell Level = 0, base 0, max 0

Category = 0
[...]

Issue #16744, opened 2016-03-07
6.2.3 Warlord of Draenor. Need support SAI and C++ hook for created areatrigger.


ghost:
**Description**:
https://github.com/TrinityCore/TrinityCore/blob/6.x/src/server/game/Spells/SpellEffects.cpp
https://github.com/TrinityCore/TrinityCore/blob/5576f06f20a79fc49dca39fb33802239b52e052f/src/server/game/Spells/SpellEffects.cpp#L5816
**Current behaviour**: Tell us what happens
C++ and SAI don't have support hook for add scripts to created(likely summoned) areatrigger with miscvalueA.
**Expected behaviour**: Tell us what should happen instead
I don't know, but https://github.com/TrinityCore/TrinityCore/search?q=SMART_EVENT_AREATRIGGER_ONTRIGGER&type=Code&utf8=%E2%9C%93
allow script only for static areatrigger from dbc. For created = not.
**Steps to reproduce the problem**:
1. Add any SAI script.
2. 
3. 

**Branch(es)**: 335/6x
Only Warlord of Draenor 6.x.x
**TC hash/commit**:
https://github.com/TrinityCore/TrinityCore/commit/ce5def332fa12ea4b2f3223fbaa4e41750da9cfb
**TDB version**:
Latest.
**Operating system**:
Win 64.

Example number one:

``` SQL
ID - 151810 Замораживающая бомба
=================================================
Description: $@spelldesc152267

Category = 1152, SpellIconID = 8632, activeIconID = 0, SpellVisual = (35412,0)
Family SPELLFAMILY_GENERIC, flag [0] 0x00000000 [1] 0x00000000 [2] 0x00000000 [3] 0x00000000

SpellSchoolMask = 16 (SPELL_SCHOOL_MASK_FROST)
DamageClass = 1 (SPELL_DAMAGE_CLASS_MAGIC)
PreventionType = 3 (3)
AttributesEx9: 0x00000010 (SPELL_ATTR9_SPECIAL_DELAY_CALCULATION)
=================================================
Spell Level = 0, base 0, max 0

Category = 1152
DispelType = 0 (DISPEL_NONE)
Mechanic = 0 (MECHANIC_NONE)
SpellRange: (Id 37) "Extended Range":
    MinRangeNegative = 0, MinRangePositive = 0
    MaxRangeNegative = 50, MaxRangePositive = 50

Speed 1,50
Cast time (Id 4) = 1,00
[...]

PR #18035, opened 2016-10-03
Areatrigger template

Traesh:
**Changes proposed:**
-  Basic AreaTrigger Entities handlers.
-  Currently handle non-moving, non-scaling AreaTrigger created by spell effect 179
-  handle type SPHERE, BOX, CYLINDER, POLYGON
-  Added C++ script hooks

**Target branch(es):** 6.x

**Issues addressed:**
- Related to #16744
- Related to #16937

Builded
Tested ingame with spell 187651 (AreaTrigger 4424), type Cylinder
Tested ingame with spell 163559 (AreaTrigger 2472), type Polygon

- This is the basic system, ToDo won't be fixed in this PR :
- [ ] ~~SAI Events~~
- [x] Moving areatriggers (Spline)
- [ ] Scaling areatriggers
- [ ] Find UNK Flags
- [ ] Implement unhandled flags

(Re-Opening PR due to a fail on previous repo, all old reviews have been fixed)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -1602,6 +1602,9 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
             // remove all dyn objects
             RemoveAllDynObjects();
 
+            // remove all areatriggers entities
+            RemoveAllAreaTriggers();
```
matukaa: Shouldn't you resummon all area triggers that are created by Apply Aura: AreaTrigger? Spells like [Windwalking](http://www.wowhead.com/spell=157411/windwalking) are applied only once, therefore the trigger is only created once, but they should be present at any time

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -100,14 +148,256 @@ void AreaTrigger::Update(uint32 p_time)
         Remove(); // expired
 
     WorldObject::Update(p_time);
+    _timeSinceCreated += p_time;
+
+    sScriptMgr->OnAreaTriggerEntityUpdate(this, p_time);
+
+    switch (GetTemplate()->GetType())
+    {
+        case AREATRIGGER_TYPE_SPHERE:
```
tkrokli:
```
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp:155:13: fatal error: 
      enumeration value 'AREATRIGGER_TYPE_NONE' not handled in switch [-Wswitch]
    switch (GetTemplate()->GetType())
            ^
```

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -662,6 +666,64 @@ class ScriptRegistrySwapHooks<InstanceMapScript, Base>
...
+    void BeforeUnload() final override
+    {
+        ASSERT(!swapped);
+    }
+
+private:
+    bool swapped;
+};
+
+/// This hook is responsible for swapping AreaTriggerEntityScript's
```
RAlexis: Typo

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -662,6 +666,64 @@ class ScriptRegistrySwapHooks<InstanceMapScript, Base>
     bool swapped;
 };
 
+/// This hook is responsible for swapping SceneScript's
```
Krudor: Does this belong to this PR?

Krudor:
Gonna see if I can make time to test this, in any case it's an awesome PR I've personally been looking forward to!

src/server/game/Scripting/ScriptMgr.cpp
Shauren: Its a missed change from previous PR (already merged)

Krudor:
Tested with a 292 vertice polygon along with a script and works like a charm, what a joy to see this working, great job!

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
tkrokli: Maybe you could add just the 2 lines
```cpp
        case AREATRIGGER_TYPE_NONE:
            break;
```
on top of the existing switches?

Traesh: missing default added :)

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -77,18 +88,55 @@ bool AreaTrigger::CreateAreaTrigger(ObjectGuid::LowType guidlow, uint32 triggerE
...
+
+    SetUInt32Value(AREATRIGGER_DURATION, spell->GetDuration());
+
     SetObjectScale(1);
 
     SetGuidValue(AREATRIGGER_CASTER, caster->GetGUID());
     SetUInt32Value(AREATRIGGER_SPELLID, spell->Id);
     SetUInt32Value(AREATRIGGER_SPELLVISUALID, spellVisual);
     SetUInt32Value(AREATRIGGER_DURATION, spell->GetDuration());
+    SetUInt32Value(AREATRIGGER_TIME_TO_TARGET_SCALE, spell->GetDuration()); // Todo
```
Shauren: This is the time it takes for the client to play "change scale" animation, you probably should not set it to total duration

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -100,14 +148,258 @@ void AreaTrigger::Update(uint32 p_time)
...
+    HandleUnitEnterExit(targetList);
+}
+
+void AreaTrigger::SearchUnitInPolygon()
+{
+    std::list<Unit*> targetList;
+
+    Trinity::AnyUnitInObjectRangeCheck check(this, GetTemplate()->MaxSearchRadius);
+    Trinity::UnitListSearcher<Trinity::AnyUnitInObjectRangeCheck> searcher(this, targetList, check);
+    this->VisitNearbyObject(GetTemplate()->MaxSearchRadius, searcher);
```
Shauren: unneeded this->

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -834,22 +835,46 @@ class TC_GAME_API GroupScript : public ScriptObject
 
 class TC_GAME_API SceneScript : public ScriptObject
 {
+protected:
```
joschiwald: indent like it was before and you will get a better git diff

src/server/game/Entities/AreaTrigger/AreaTrigger.h
```diff
@@ -36,12 +38,36 @@ class TC_GAME_API AreaTrigger : public WorldObject, public GridObject<AreaTrigge
...
+        ObjectGuid _casterGuid;
+
         int32 _duration;
         uint32 _spellXSpellVisualId;
+        uint32 _timeSinceCreated;
+
+        std::vector<AreaTriggerPolygonVertice> _polygonVertices;
+
+        AreaTriggerTemplate const* _areaTriggerTemplate;
+        std::set<ObjectGuid> _insideUnits;
```
joschiwald: there is already a GuidSet?

sql/updates/world/6.x/9999_99_99_17997_world.sql
```diff
@@ -0,0 +1,41 @@
...
+  `Idx` int(10) unsigned NOT NULL,
+  `VerticeX` float NOT NULL DEFAULT '0',
+  `VerticeY` float NOT NULL DEFAULT '0',
+  `VerticeTargetX` float NOT NULL DEFAULT '0',
+  `VerticeTargetY` float NOT NULL DEFAULT '0',
+  `VerifiedBuild` int(10) unsigned NOT NULL DEFAULT '0',
+  PRIMARY KEY (`areatriggerId`,`Idx`)
+) ENGINE=MyISAM DEFAULT CHARSET=utf8;
+
+INSERT INTO `areatrigger_template`(`id`,`flags`,`moveCurveId`,`scaleCurveId`,`morphCurveId`,`facingCurveId`,`data0`,`data1`,`data2`,`data3`,`data4`,`data5`,`scriptName`) VALUES
```
joschiwald: correct the column names

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
Krudor: It appears to be default behavior though

Shauren: It has no impact on current implementation (because you send final scale before the areatrigger is visible). You would need to modify scale after it is added to world to notice this

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -834,22 +835,46 @@ class TC_GAME_API GroupScript : public ScriptObject
...
+
+    // Called when a scene is canceled
+    virtual void OnSceneCancel(Player* /*player*/, uint32 /*sceneInstanceID*/, SceneTemplate const* /*sceneTemplate*/) { }
+
+    // Called when a scene is completed
+    virtual void OnSceneComplete(Player* /*player*/, uint32 /*sceneInstanceID*/, SceneTemplate const* /*sceneTemplate*/) { }
+
+};
+
+class TC_GAME_API AreaTriggerEntityScript : public ScriptObject
```
joschiwald: what about using the existing AreaTriggerScript?

Traesh: Bad idea imho, there is no reason to add an "OnUpdate" function in AreaTriggerScript by example. This is two totally different things, even if they have close names

Krudor: I think he ment inherit AreaTriggerScript

src/server/database/Database/Implementation/WorldDatabase.cpp
```diff
@@ -91,4 +91,6 @@ void WorldDatabaseConnection::DoPrepareStatements()
     PrepareStatement(WORLD_UPD_CREATURE_ZONE_AREA_DATA, "UPDATE creature SET zoneId = ?, areaId = ? WHERE guid = ?", CONNECTION_ASYNC);
     PrepareStatement(WORLD_UPD_GAMEOBJECT_ZONE_AREA_DATA, "UPDATE gameobject SET zoneId = ?, areaId = ? WHERE guid = ?", CONNECTION_ASYNC);
     PrepareStatement(WORLD_SEL_GUILD_REWARDS_REQ_ACHIEVEMENTS, "SELECT AchievementRequired FROM guild_rewards_req_achievements WHERE ItemID = ?", CONNECTION_SYNCH);
+    PrepareStatement(WORLD_SEL_AREATRIGGER_TEMPLATES, "SELECT Id, Flags, MoveCurveId, ScaleCurveId, MorphCurveId, FacingCurveId, Data0, Data1, Data2, Data3, Data4, Data5, ScriptName FROM `areatrigger_template`", CONNECTION_SYNCH);
```
Shauren: Please don't use prepared statements for loading templates on startup

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
Shauren: Why not parse this field with WPP then and store in db?

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -100,14 +148,258 @@ void AreaTrigger::Update(uint32 p_time)
...
+
+        return GetDistance(unit) > radius || unit->GetPositionZ() < minZ || unit->GetPositionZ() > maxZ;
+    });
+
+    HandleUnitEnterExit(targetList);
+}
+
+void AreaTrigger::HandleUnitEnterExit(std::list<Unit*> newTargetList)
+{
+    for (Unit* unit : newTargetList)
```
Shauren: I believe this would be a better way to handle this:
```
GuidUnorderedSet exitUnits = _insideUnits;
_insideUnits.clear();
for (Unit* unit : newTargetList)
{
    if (exitUnits.erase(unit) == 0) // erase(key_type) returns number of elements erased
    {
        // handle entering, chat handler and script hook
    }
    _insideUnits.insert(unit->GetGUID());
}

for (ObjectGuid const& exitUnitGuid : exitUnits)
{
    // handle exit, no conditions
}
```
And you don't need to check IsAlive() anywhere, AnyUnitInObjectRangeCheck does that so `newTargetList` will never contain dead units

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef TRINITYCORE_AREATRIGGER_TEMPLATE_H
+#define TRINITYCORE_AREATRIGGER_TEMPLATE_H
+
+#include "Object.h"
+#include "ObjectAccessor.h"
```
Shauren: Don't include ObjectAccessor here, you don't need it (do it in .cpp)

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+                if (pointDist > MaxSearchRadius)
+                    MaxSearchRadius = pointDist;
+            }
+
+            break;
+        }
+        case AREATRIGGER_TYPE_CYLINDER:
+        {
+            MaxSearchRadius = std::sqrt((CylinderDatas.Radius * CylinderDatas.Radius) + (CylinderDatas.Height * CylinderDatas.Height));
```
Shauren: VisitNearbyObject expects 2D distance for its first argument, also don't turn this into a sphere if it's a cylinder

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.cpp
```diff
@@ -0,0 +1,80 @@
...
+    switch (GetType())
+    {
+        case AREATRIGGER_TYPE_SPHERE:
+        {
+            MaxSearchRadius = SphereDatas.Radius;
+            break;
+        }
+        case AREATRIGGER_TYPE_BOX:
+        {
+            MaxSearchRadius = std::max(std::max(BoxDatas.Extents[0], BoxDatas.Extents[1]), BoxDatas.Extents[2]);
```
Shauren: Don't take Extents[2] into account here, if the height is large you will make search visit unneccessarily large number of grids.
On top of that the equation should be `std::sqrt(BoxDatas.Extents[0] * BoxDatas.Extents[0] / 4 + BoxDatas.Extents[1] * BoxDatas.Extents[1] / 4)`

src/server/game/Entities/Object/Object.cpp
```diff
@@ -509,111 +509,165 @@ void Object::BuildMovementUpdate(ByteBuffer* data, uint32 flags) const
...
+        {
+            *data << float(0); // X
+            *data << float(0); // Y
+            *data << float(0); // Z
+        }
 
-    //    if (HasScaleCurveID)
-    //        *data << uint32(ScaleCurveID);
+        if (hasScaleCurveID)
+            *data << areaTriggerTemplate->ScaleCurveId;
```
Shauren: always explicitly specify the type when appending to packets (ObjectGuid is the only exception)
`*data << uint32(areaTriggerTemplate->ScaleCurveId);`

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9628,3 +9629,73 @@ void ObjectMgr::LoadSceneTemplates()
...
+        areaTriggerTemplate.ScaleCurveId = fields[3].GetInt32();
+        areaTriggerTemplate.MorphCurveId = fields[4].GetInt32();
+        areaTriggerTemplate.FacingCurveId = fields[5].GetInt32();
+
+        for (uint8 i = 0; i < MAX_AREATRIGGER_ENTITY_DATA; ++i)
+            areaTriggerTemplate.DefaultDatas.Data[i] = fields[6 + i].GetFloat();
+
+        areaTriggerTemplate.ScriptId = sObjectMgr->GetScriptId(fields[12].GetCString());
+
+        PreparedStatement* verticesStmt = WorldDatabase.GetPreparedStatement(WORLD_SEL_AREATRIGGER_POLYGON_VERTICES);
```
Shauren: Select vertices once at the start and cache them in map<areatriggerId, vector<vertice>>, then just std::move that vector when looping over result from templates

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -100,14 +148,258 @@ void AreaTrigger::Update(uint32 p_time)
...
         RemoveFromWorld();
         AddObjectToRemoveList();
     }
 }
+
+void AreaTrigger::SearchUnitInSphere()
+{
+    std::list<Unit*> targetList;
+
+    Trinity::AnyUnitInObjectRangeCheck check(this, GetTemplate()->MaxSearchRadius);
```
Shauren: You will need to extend AnyUnitInObjectRangeCheck constructor to add 3rd argument check3D and pass false to make its inner check IsWithinDistInMap also check only 2D distance here (but it must default to true to keep its current behavior for all other uses in TC)

Shauren (changes requested):
Almost good to go.

src/server/game/Entities/AreaTrigger/AreaTrigger.h
```diff
@@ -36,12 +38,36 @@ class TC_GAME_API AreaTrigger : public WorldObject, public GridObject<AreaTrigge
...
+        ObjectGuid _casterGuid;
+
         int32 _duration;
         uint32 _spellXSpellVisualId;
+        uint32 _timeSinceCreated;
+
+        std::vector<AreaTriggerPolygonVertice> _polygonVertices;
+
+        AreaTriggerTemplate const* _areaTriggerTemplate;
+        GuidSet _insideUnits;
```
Shauren: And use GuidUnorderedSet here

src/server/game/Entities/Object/Object.cpp
Traesh: May i ask why ? Wouldn't it imply to modify code in 2 location if type change ?

streetrat:
is that normal that you can run thru for example a hunter trap between two update ticks without triggering it if you are fast enough (mounted) ?

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
Traesh: I updated the box/cylinder/polygon, sphere should check in 3D

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
Traesh: I felt like this was the worst part of my PR, Thanks very much :)

Krudor:
I maaay have noticed the same behavior as streetrat described, any thoughts about it Traesh?
It's so responsive on retail, I thought they used a client opcode for sending info whenever a player entered a dynamic areatrigger, much like the static ones, but they don't.

Traesh:
Weird, AreaTrigger update is called at each update there is no delay. Is there a delay between two player coordinates calculation ?

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9634,9 +9634,33 @@ void ObjectMgr::LoadAreaTriggerTemplates()
 {
     uint32 oldMSTime = getMSTime();
     _areaTriggerTemplateStore.clear();
+    std::map< uint32, std::vector<AreaTriggerPolygonVertice> > verticeByAreatrigger;
```
Shauren: dat space padding in template argumets, such stronk c++03

streetrat:
player move hearthbeat is sent from client about once per second if you run straight

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -383,3 +385,70 @@ bool AreaTrigger::CheckIsInPolygon2D(Position* pos) const
...
+
+void AreaTrigger::AddAuras(Unit* unit)
+{
+    for (AreaTriggerAuras aura : GetTemplate()->Auras)
+    {
+        if (UnitFitToAuraRequirement(unit, aura.TargetType))
+        {
+            if (Unit* caster = ObjectAccessor::GetUnit(*this, _casterGuid))
+            {
+                caster->AddAura(aura.AuraId, unit);
```
joschiwald: some spells are castable while dead, does the spells have limited range that could case too far away issues?

that gameobject cast is bad, don't look on it

AddAura will do nothing for spells with non apply aura spelleffects

Shauren:
@streetrat every half a second to be exact

Ulduar:
I know it do not matter at the moment, but nevertheless:
On Future:
``` CPP
struct
        {
            uint32 spellId;
            uint32 type;
        } removeDynamicObjectOrAreatrigger;
        
case SMART_ACTION_REMOVE_DYNAMIC_OBJECT_OR_AREATRIGGER:
        {
            if (!me)
                break;
                
            if (e.action.removeDynamicObjectOrAreatrigger.type == 0)
            {
                me->RemoveDynObject(e.action.removeDynamicObjectOrAreatrigger.spellId);
            }
            else if (e.action.removeDynamicObjectOrAreatrigger.type == 1)
            {
                me->RemoveAllDynObjects();
            }
            else if (e.action.removeDynamicObjectOrAreatrigger.type == 2)
            {
                me->RemoveAreaTrigger(e.action.removeDynamicObjectOrAreatrigger.spellId);
            }
            else if (e.action.removeDynamicObjectOrAreatrigger.type == 3)
            {
                me->RemoveAllAreaTriggers();
            }
            else if (e.action.removeDynamicObjectOrAreatrigger.type == 4)
            {
                me->RemoveAllDynObjects();
                me->RemoveAllAreaTriggers();
            }
            else
            {
                TC_LOG_ERROR("sql.sql", "SmartScript: Invalid type for SMART_ACTION_SET_SPELL_VISUAL_STATE, skipping");
                break;
            }
            break;
        }
```

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -77,18 +89,56 @@ bool AreaTrigger::CreateAreaTrigger(ObjectGuid::LowType guidlow, uint32 triggerE
...
     SetGuidValue(AREATRIGGER_CASTER, caster->GetGUID());
     SetUInt32Value(AREATRIGGER_SPELLID, spell->Id);
     SetUInt32Value(AREATRIGGER_SPELLVISUALID, spellVisual);
     SetUInt32Value(AREATRIGGER_DURATION, spell->GetDuration());
+    SetUInt32Value(AREATRIGGER_TIME_TO_TARGET_SCALE, GetTemplate()->TimeToTargetScale != 0 ? GetTemplate()->TimeToTargetScale : spell->GetDuration());
+
+    switch (GetTemplate()->GetType())
+    {
+        case AREATRIGGER_TYPE_SPHERE:
+            SetUInt32Value(AREATRIGGER_BOUNDS_RADIUS_2D, GetTemplate()->SphereDatas.Radius);
```
joschiwald: SetFloatValue

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -100,14 +150,305 @@ void AreaTrigger::Update(uint32 p_time)
...
+{
+    for (AreaTriggerAuras aura : GetTemplate()->Auras)
+    {
+        unit->RemoveAurasDueToSpell(aura.AuraId, _casterGuid);
+    }
+}
+
+bool AreaTrigger::UnitFitToAuraRequirement(Unit* unit, AreaTriggerAuraTypes targetType) const
+{
+    switch (targetType)
```
Traesh: Missing default, again :/

Traesh:
If player coordinates are updated each 500 ms i don't think i'll be able to be make areatriggers more reactive than that.

streetrat:
well, g3d library can calculate from 2 different xyz coords of it hits an object or not :)
store the the player position on every update, but before storing check if the last pos + current pos's vector hit the trigger, "simple" :D

GitVerge:
@streetrat What if the player uses a spell like mage's blink? Maybe that vectors addition will make him 'hit' the areatrigger

streetrat:
you can see it if its a near teleport or movement packet, and maybe IsBeingTeleported is true for these too in core
but i think this whole stuff isnt needed, the smallest areatriggers should be hunter traps, and those have radius = 2 yards in the spelleffect, so thats 4 yards, if player runs 4 yards in 1 server tick that is possibly not a normal player :)

streetrat:
why is spell_areatrigger needed? if you use the miscId as triggerId it simpy works

Shauren:
Not blizzlike, miscId != entry in packets

streetrat:
hm, true
and do you maybe know what is extra_curva_value (dont know the exact name now, im at work) updatefield for? i tested with mage-blizzard, and there was no iceshard anim until i set extra curve thing to the sniffed value

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9628,3 +9629,140 @@ void ObjectMgr::LoadSceneTemplates()
 
     TC_LOG_INFO("server.loading", ">> Loaded %u scene templates in %u ms.", count, GetMSTimeDiffToNow(oldMSTime));
 }
+
+void ObjectMgr::LoadAreaTriggerTemplates()
+{
+    uint32 oldMSTime = getMSTime();
+    _areaTriggerTemplateStore.clear();
```
joschiwald: you can not clear this storage on reload since areatriggers have a pointer to container element

src/server/game/Entities/Object/Object.cpp
```diff
@@ -509,111 +509,165 @@ void Object::BuildMovementUpdate(ByteBuffer* data, uint32 flags) const
...
+        data->WriteBit(hasUnk3);
+        data->WriteBit(hasUnk4);
+        data->WriteBit(hasAreaTriggerSphere);
+        data->WriteBit(hasAreaTriggerBox);
+        data->WriteBit(hasAreaTriggerPolygon);
+        data->WriteBit(hasAreaTriggerCylinder);
+        data->WriteBit(hasAreaTriggerSpline);
+        data->WriteBit(hasAreaTriggerUnkType);
+
+        if (areaTrigger->GetTemplate()->HasFlag(AREATRIGGER_FLAG_UNK3))
```
joschiwald: if (hasUnk3)

joschiwald:
i found areatriggers with different curveids, maybe there is a relation to spellmisc

at

joschiwald:
https://gist.github.com/joschiwald/55073970187a50decb45368cb54237a6

looks so, CurveIds depend on SpellMisc

Traesh:
Indeed, could you please check if there is other values that change depending of misc ?

For example : TimeToTargetScale

streetrat:
im playing around with the movement stuff for this
does anyone have the struct for SMSG_AREA_TRIGGER_RE_SHAPE ?
seems like blizz send reshape instead of repath when changing the speed of a moving AT (like frozen orb) idea is from sniffs ofc

Traesh:
https://github.com/TrinityCore/WowPacketParser/commit/0667d875f7c57e73324b2d31259319c664dabcf4

streetrat:
bah, thanks, my wpp is a week old..

Traesh:
Just implanted the spline system for Areatriggers, will need tests

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -141,6 +168,16 @@ void AreaTrigger::Update(uint32 p_time)
     WorldObject::Update(p_time);
     _timeSinceCreated += p_time;
 
+    if (GetTemplate()->HasFlag(AREATRIGGER_FLAG_HAS_ATTACHED))
+    {
+        if (Unit* caster = GetCaster())
+        {
+            Relocate(caster);
```
Shauren: Relocate does not trigger visibility updates, you need to implement something like Map::CreatureRelocation

src/server/game/Entities/Object/Object.cpp
```diff
@@ -572,8 +572,8 @@ void Object::BuildMovementUpdate(ByteBuffer* data, uint32 flags) const
         {
             std::vector<Position> splines = areaTrigger->GetSplines();
 
-            *data << float(areaTrigger->GetTotalDuration());    // TimeToTarget
-            *data << float(areaTrigger->GetTimeSinceCreated()); // ElapsedTimeForMovement
+            *data << float(areaTriggerMiscTemplate->TimeToTarget);
+            *data << float(0); // ElapsedTimeForMovement, always 0 because sent at creation
```
Shauren: This is a wrong assumption, a player entering visibility range of the areatrigger after it has been created will see a nonzero value

tkrokli:
https://travis-ci.org/TrinityCore/TrinityCore/builds/171328746 :
```
$ cat sql/updates/world/master/*.sql | mysql -utrinity -ptrinity world
ERROR 1064 (42000) at line 345: You have an error in your SQL syntax;
 check the manual that corresponds to your MySQL server version
 for the right syntax to use near '`VerifiedBuild` int(10) unsigned NOT NULL DEFAULT '0',
  PRIMARY KEY (`MoveCurve' at line 7
The command "cat sql/updates/world/master/*.sql | mysql -utrinity -ptrinity world" exited with 1.
```

Ulduar:
@Traesh thank your for start working on sat system (spellareatrigger)

tkrokli:
```
$ cat sql/updates/world/master/*.sql | mysql -utrinity -ptrinity world
ERROR 1054 (42S22) at line 382: Unknown column 'MoveCurveId' in 'field list'
The command "cat sql/updates/world/master/*.sql | mysql -utrinity -ptrinity world" exited with 1.
```

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -96,14 +169,374 @@ void AreaTrigger::Update(uint32 p_time)
         Remove(); // expired
 
     WorldObject::Update(p_time);
+    _timeSinceCreated += p_time;
+
+    if (GetTemplate()->HasFlag(AREATRIGGER_FLAG_HAS_ATTACHED))
+    {
+        if (Unit* caster = GetCaster())
```
joschiwald: is it always the caster? it depends on aura target

sql/updates/world/master/9999_99_99_17997_world.sql
```diff
@@ -0,0 +1,76 @@
...
+  `AuraId` int(10) unsigned NOT NULL,
+  `TargetType` int(10) unsigned NOT NULL DEFAULT '0',
+  `CastType` int(10) unsigned NOT NULL DEFAULT '0',
+  `VerifiedBuild` int(10) unsigned NOT NULL DEFAULT '0'
+) ENGINE=MyISAM DEFAULT CHARSET=utf8;
+
+DROP TABLE IF EXISTS `spell_areatrigger`;
+CREATE TABLE `spell_areatrigger` (
+  `SpellMiscId` int(10) unsigned NOT NULL,
+  `AreatriggerId` int(10) unsigned NOT NULL,
```
joschiwald: for consistency: `AreaTriggerId`

Seyden:
Would it be possible, if you add areatrigger transport? So SPELL_AURA_AREATRIGGER always attach to the caster and moves with him. Would be nice, and thanks for the nice pr

joschiwald:
since template is splitted into areatrigger_template and spell_areatrigger, not all flags depend on areatrigger_template
i think we should set "flags" on runtime and give the areatrigger_template a type field, or just save some special flags on db

```sql
INSERT INTO `areatrigger_template` (`Id`, `Flags`, `Data0`, `Data1`, `Data2`, `Data3`, `Data4`, `Data5`, `BoundRadius`, `VerifiedBuild`) VALUES
(10003, 1064964, 1.5, 1.5, 0, 0, 0, 0, 0, 22522),
(10003, 16388, 1.5, 1.5, 0, 0, 0, 0, 1.5, 22522),
(10003, 278532, 1.5, 1.5, 0, 0, 0, 0, 1.5, 22522);

INSERT INTO `spell_areatrigger` (`SpellId`, `Id`, `ScaleCurveId`, `MorphCurveId`, `FacingCurveId`, `MoveCurveId`, `DecalPropertiesId`, `TimeToTarget`, `TimeToTargetScale`, `TimeToTargetExtraScale`) VALUES
(0, 10003, 0, 0, 0, 0, 2112736, 0, 3221248287, 2558549339),
(0, 10003, 0, 0, 0, 0, 0, 0, 0, 0),
(0, 10003, 0, 0, 0, 0, 2112736, 0, 2147506463, 2558549371),
(196744, 10003, 0, 0, 0, 0, 0, 0, 0, 0),
(196745, 10003, 0, 0, 0, 0, 0, 0, 0, 0),
(0, 10003, 0, 0, 0, 0, 0, 0, 3221254391, 0),
(196747, 10003, 0, 0, 0, 0, 0, 0, 0, 0),
(0, 10003, 0, 0, 0, 0, 565248000, 0, 0, 0),
(0, 10003, 0, 0, 0, 0, 1077936128, 0, 227, 4),
(0, 10003, 0, 0, 0, 0, 0, 0, 1, 1077936128),
(0, 10003, 0, 0, 0, 0, 2112232, 0, 2147505522, 3053469017),
(0, 10003, 0, 0, 0, 0, 2112720, 0, 1073763735, 2558548105),
(196744, 10003, 0, 0, 0, 0, 0, 0, 5000, 0),
(0, 10003, 0, 0, 0, 0, 2112720, 0, 3221247383, 2558551247),
(0, 10003, 0, 0, 0, 0, 2112720, 0, 1073763735, 2558552097),
(196745, 10003, 0, 0, 0, 0, 0, 0, 0, 30565),
(0, 10003, 0, 0, 0, 0, 2503933952, 0, 0, 0),
(0, 10003, 0, 0, 0, 0, 2112720, 0, 3221247383, 2558548105),
(0, 10003, 0, 0, 0, 0, 905969664, 0, 0, 0),
(0, 10003, 0, 0, 0, 0, 0, 0, 1073763735, 0),
(0, 10003, 0, 0, 0, 0, 2112720, 0, 2147505559, 2558548058),
(0, 10003, 0, 0, 0, 0, 2112720, 0, 3221247383, 2558547872),
(0, 10003, 0, 0, 0, 0, 2112720, 0, 21911, 2558551235),
(0, 10003, 0, 0, 0, 0, 2109588, 0, 1073745178, 1824538911),
(0, 10003, 0, 0, 0, 0, 1077936128, 0, 227, 17),
(0, 10003, 0, 0, 0, 0, 2109564, 0, 29129, 2558552522),
(0, 10003, 0, 0, 0, 0, 2109564, 0, 2147512777, 2558552733),
(196746, 10003, 0, 0, 0, 0, 0, 0, 0, 0),
(196744, 10003, 0, 0, 0, 0, 0, 0, 0, 3),
(0, 10003, 0, 0, 0, 0, 2109564, 0, 1073770953, 2558547465),
(0, 10003, 0, 0, 0, 0, 2109564, 0, 2147512777, 2558550560),
(0, 10003, 0, 0, 0, 0, 960687107, 0, 927201924, 3527786931),
(0, 10003, 0, 0, 0, 0, 2112240, 0, 17791, 2558551115),
(0, 10003, 0, 0, 0, 0, 2103004, 0, 1073763321, 2558550146),
(0, 10003, 0, 0, 0, 0, 4210688, 0, 1627389952, 0),
(0, 10003, 0, 0, 0, 0, 7, 0, 4278194175, 4268817407),
(0, 10003, 0, 0, 0, 0, 2112720, 0, 1073750178, 2558550583),
(0, 10003, 0, 0, 0, 0, 2112720, 0, 2147492002, 2558548220),
(0, 10003, 0, 0, 0, 0, 8968, 0, 3564822543, 422578045),
(0, 10003, 0, 0, 0, 0, 3172532224, 0, 0, 0),
(0, 10003, 0, 0, 0, 0, 17583, 0, 0, 1191182336);
```

akrom23:
Also,
``SMSG_MOVE_APPLY_MOVEMENT_FORCE, SMSG_MOVE_REMOVE_MOVEMENT_FORCE, SMSG_MOVE_UPDATE_APPLY_MOVEMENT_FORCE, SMSG_MOVE_UPDATE_REMOVE_MOVEMENT_FORCE, CMSG_MOVE_APPLY_MOVEMENT_FORCE_ACK, CMSG_MOVE_REMOVE_MOVEMENT_FORCE_ACK`` are connected with areatrigger movement - i don't know how exactly.

Traesh:
"SMSG_MOVE_APPLY_MOVEMENT_FORCE" is not directly related to areatrigger_template system, the only link is that sometime it is send when the player enter an areatrigger :)

This can already be done using C++ hooks of areatriggers

src/server/game/Entities/AreaTrigger/AreaTrigger.h
```diff
@@ -36,12 +39,53 @@ class TC_GAME_API AreaTrigger : public WorldObject, public GridObject<AreaTrigge
...
+
         int32 _duration;
+        int32 _totalDuration;
         uint32 _spellXSpellVisualId;
+        uint32 _timeSinceCreated;
+
+        std::vector<AreaTriggerPolygonVertice> _polygonVertices;
+        std::vector<Position> _splines;
+
+        float _totalDistance;
```
joschiwald: not initialized in constructor

src/server/game/Maps/Map.cpp
```diff
@@ -1091,6 +1092,38 @@ void Map::DynamicObjectRelocation(DynamicObject* dynObj, float x, float y, float
     ASSERT(integrity_check == old_cell);
 }
 
+void Map::AreaTriggerRelocation(AreaTrigger* at, float x, float y, float z, float orientation)
+{
+    Cell integrity_check(at->GetPositionX(), at->GetPositionY());
+    Cell old_cell = at->GetCurrentCell();
```
joschiwald: old_cell contains 0 and will trigger assertion
Map::AddToGrid block with areatrigger specialisation is missing?

src/server/game/Entities/AreaTrigger/AreaTrigger.h
Traesh: Done

tkrokli:
```
ERROR 1064 (42000) at line 318048: You have an error in your SQL syntax;
check the manual that corresponds to your MySQL server version for the right syntax to use near
'`,`class`,`button`,`action`,`type`) VALUES
(7, 3, 0, 193455, 0),
(7, 3, 9, 20589' at line 1
```

sql/updates/world/master/9999_99_99_17997_world.sql
```diff
@@ -0,0 +1,77 @@
...
+DROP TABLE IF EXISTS `spell_areatrigger`;
+CREATE TABLE `spell_areatrigger` (
+  `SpellMiscId` int(10) unsigned NOT NULL,
+  `AreaTriggerId` int(10) unsigned NOT NULL,
+  `MoveCurveId` int(10) unsigned NOT NULL DEFAULT '0',
+  `ScaleCurveId` int(10) unsigned NOT NULL DEFAULT '0',
+  `MorphCurveId` int(10) unsigned NOT NULL DEFAULT '0',
+  `FacingCurveId` int(10) unsigned NOT NULL DEFAULT '0',
+  `TimeToTarget` int(10) unsigned NOT NULL DEFAULT '0',
+  `TimeToTargetScale` int(10) unsigned NOT NULL DEFAULT '0',
```
joschiwald: missing verifiedbuild

sql/updates/world/master/9999_99_99_17997_world.sql
```diff
@@ -0,0 +1,77 @@
...
+  `SplineX` float NOT NULL DEFAULT '0',
+  `SplineY` float NOT NULL DEFAULT '0',
+  `SplineZ` float NOT NULL DEFAULT '0',
+  `VerifiedBuild` int(10) unsigned NOT NULL DEFAULT '0',
+  PRIMARY KEY (`MoveCurveId`,`Idx`)
+) ENGINE=MyISAM DEFAULT CHARSET=utf8;
+
+DROP TABLE IF EXISTS `areatrigger_template_auras`;
+CREATE TABLE `areatrigger_template_auras` (
+  `AreaTriggerId` int(10) unsigned NOT NULL,
```
joschiwald: i guess this should also depend on spellmisc now?

matukaa:
it's just an observation, but maybe you should add a check for SpellEffIndex :) [this](http://www.wowhead.com/spell=123986/chi-burst) spell for example creates two AreaTriggers that need two different scripts, therefore two different templates.

Shauren:
@matukaa not needed, MiscValue from SpellEffect is used to determine template, not spellid

src/server/game/Entities/Object/Object.cpp
```diff
@@ -509,111 +509,179 @@ void Object::BuildMovementUpdate(ByteBuffer* data, uint32 flags) const
...
+        data->WriteBit(hasAreaTriggerCylinder);
+        data->WriteBit(hasAreaTriggerSpline);
+        data->WriteBit(hasAreaTriggerUnkType);
+
+        if (hasUnk3)
+            data->WriteBit(0);
+
+        if (hasAreaTriggerSpline)
+        {
+            std::vector<Position> splines = areaTrigger->GetSplines();
```
Shauren: Is sending rotated spline points intended here?

Traesh: Yes, client expect absolute position

Shauren: And polygon vertices are not sent rotated? Thats some strange design by blizz then.
Anyway you need to undo the rotation when parsing in wpp

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -96,14 +163,377 @@ void AreaTrigger::Update(uint32 p_time)
...
+    float currentDistanceFromLastPoint = currentDistance - _distanceToPoints[lastPositionIndex];
+    float percentFromLastPoint = currentDistanceFromLastPoint / (_distanceToPoints[lastPositionIndex + 1] - _distanceToPoints[lastPositionIndex]);
+
+    Position lastPosition = _splines[lastPositionIndex];
+    Position nextPosition = _splines[lastPositionIndex + 1];
+    float moveAngle = lastPosition.GetAngle(&_splines[lastPositionIndex + 1]);
+
+    m_positionX = lastPosition.GetPositionX() + (currentDistanceFromLastPoint * cos(moveAngle));
+    m_positionY = lastPosition.GetPositionY() + (currentDistanceFromLastPoint * sin(moveAngle));
+    m_positionZ = lastPosition.GetPositionZ() + (percentFromLastPoint * (nextPosition.GetPositionZ() - lastPosition.GetPositionZ()));
```
Traesh: @Shauren : Should i use "AreaTriggerRelocation" here too ?

sql/updates/world/master/9999_99_99_17997_world.sql
Traesh: Don't think so, i feel like spellmisc mostly handle AT movements & scaling, while AreaTriggerId handle AT shape & casted spell.

But i may be wrong, if we found any AreaTrigger where spell casted depend of spellmisc, i'll change it

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
Shauren: Yes

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
```diff
@@ -0,0 +1,174 @@
...
+    AREATRIGGER_FLAG_UNK5                       = 0x00400
+};
+
+enum AreaTriggerTypes
+{
+    AREATRIGGER_TYPE_SPHERE     = 0,
+    AREATRIGGER_TYPE_BOX        = 1,
+    AREATRIGGER_TYPE_POLYGON    = 2,
+    AREATRIGGER_TYPE_CYLINDER   = 3,
+    AREATRIGGER_TYPE_MAX        = 4
```
Traesh: Is 2 unused ? Or is this the "hasAreaTriggerUnkType" value ?

joschiwald: all areatriggers with the hasAreaTriggerUnkType bit are spheres

sql/updates/world/master/9999_99_99_17997_world.sql
```diff
@@ -0,0 +1,77 @@
...
+  `VerticeY` float NOT NULL DEFAULT '0',
+  `VerticeTargetX` float NOT NULL DEFAULT '0',
+  `VerticeTargetY` float NOT NULL DEFAULT '0',
+  `VerifiedBuild` int(10) unsigned NOT NULL DEFAULT '0',
+  PRIMARY KEY (`areatriggerId`,`Idx`)
+) ENGINE=MyISAM DEFAULT CHARSET=utf8;
+
+DROP TABLE IF EXISTS `areatrigger_template_splines`;
+CREATE TABLE `areatrigger_template_splines` (
+  `MoveCurveId` int(10) unsigned NOT NULL,
```
joschiwald: i found areatriggers with splines and without movecurve

Traesh: Damn :( Could you send me your examples ?

src/server/game/Entities/Object/Object.cpp
```diff
@@ -509,111 +509,179 @@ void Object::BuildMovementUpdate(ByteBuffer* data, uint32 flags) const
...
+            *data << float(areaTrigger->GetTimeSinceCreated()); // ElapsedTimeForMovement
 
-    //    if (HasScaleCurveID)
-    //        *data << uint32(ScaleCurveID);
+            data->FlushBits();
+            data->WriteBits(splinePoints.size(), 16);
 
-    //    if (HasMorphCurveID)
-    //        *data << uint32(MorphCurveID);
+            for (G3D::Vector3 spline : splinePoints)
```
joschiwald: const&

src/server/game/Entities/AreaTrigger/AreaTrigger.h
```diff
@@ -36,12 +39,51 @@ class TC_GAME_API AreaTrigger : public WorldObject, public GridObject<AreaTrigge
...
         int32 _duration;
+        int32 _totalDuration;
         uint32 _spellXSpellVisualId;
+        uint32 _timeSinceCreated;
+
+        std::vector<AreaTriggerPolygonVertice> _polygonVertices;
+        ::Movement::Spline<int32> _spline;
+
+        bool _reachedDestination;
+        int lastSplineIndex;
```
Shauren: dont use `int` type anywhere

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9628,3 +9629,176 @@ void ObjectMgr::LoadSceneTemplates()
...
+            continue;
+        }
+
+        areaTriggerTemplate.Type  = static_cast<AreaTriggerTypes>(type);
+        areaTriggerTemplate.Flags = fields[2].GetUInt32();
+
+        for (uint8 i = 0; i < MAX_AREATRIGGER_ENTITY_DATA; ++i)
+            areaTriggerTemplate.DefaultDatas.Data[i] = fields[3 + i].GetFloat();
+
+        areaTriggerTemplate.ScriptId            = sObjectMgr->GetScriptId(fields[9].GetCString());
```
joschiwald: GetScriptId accepts std::string

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -96,14 +165,359 @@ void AreaTrigger::Update(uint32 p_time)
...
+
+    std::list<Unit*> targetList;
+
+    Trinity::AnyUnitInObjectRangeCheck check(this, GetTemplate()->MaxSearchRadius, false);
+    Trinity::UnitListSearcher<Trinity::AnyUnitInObjectRangeCheck> searcher(this, targetList, check);
+    VisitNearbyObject(GetTemplate()->MaxSearchRadius, searcher);
+
+    float halfExtentsX = extentsX / 2.0;
+    float halfExtentsY = extentsY / 2.0;
+    float halfExtentsZ = extentsZ / 2.0;
```
joschiwald: missing float suffix (2.0f)

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -57,34 +67,93 @@ void AreaTrigger::RemoveFromWorld()
...
+
     SetObjectScale(1);
 
     SetGuidValue(AREATRIGGER_CASTER, caster->GetGUID());
     SetUInt32Value(AREATRIGGER_SPELLID, spell->Id);
     SetUInt32Value(AREATRIGGER_SPELL_X_SPELL_VISUAL_ID, spellXSpellVisualId);
     SetUInt32Value(AREATRIGGER_DURATION, spell->GetDuration());
+    SetUInt32Value(AREATRIGGER_TIME_TO_TARGET, GetMiscTemplate()->TimeToTarget != 0 ? GetMiscTemplate()->TimeToTarget : spell->GetDuration());
+    SetUInt32Value(AREATRIGGER_TIME_TO_TARGET_SCALE, GetMiscTemplate()->TimeToTargetScale != 0 ? GetMiscTemplate()->TimeToTargetScale : spell->GetDuration());
+    SetFloatValue(AREATRIGGER_BOUNDS_RADIUS_2D, GetTemplate()->MaxSearchRadius);
```
joschiwald: new in 7.1: AREATRIGGER_CREATING_EFFECT_GUID is the cast guid of the spell with SPELL_EFFECT_CREATE_AREATRIGGER

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -341,6 +331,32 @@ void AreaTrigger::HandleUnitEnterExit(std::list<Unit*> newTargetList)
     }
 }
 
+void AreaTrigger::UpdatePolygonOrientation()
+{
+    float newOrientation = GetOrientation();
+
+    // No need to recalculate, orientation didn't changed
```
Traesh: Damn, saw it after commit, hoped i would be able to fix it before anyone notice x)

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -6724,3 +6724,25 @@ void AuraEffect::HandlePlayScene(AuraApplication const* aurApp, uint8 mode, bool
...
     }
 }
+
+void AuraEffect::HandleCreateAreaTrigger(AuraApplication const* aurApp, uint8 mode, bool apply) const
+{
+    if (!(mode & AURA_EFFECT_HANDLE_REAL))
+        return;
+
+    Unit* target = aurApp->GetTarget();
+    if (!target)
```
joschiwald: target never null

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -5613,17 +5613,16 @@ void Spell::EffectCreateAreaTrigger(SpellEffIndex /*effIndex*/)
     if (effectHandleMode != SPELL_EFFECT_HANDLE_HIT)
         return;
 
+    Unit* target = unitTarget != nullptr ? unitTarget : GetCaster();
```
joschiwald: unitTarget always null here

you must have a valid destTarget here, otherwise something went wrong

Shauren: @joschiwald not really, only unitTarget/gameObjTarget/itemTarget are exclusive with each other, destTarget can be set together with them

joschiwald: i was talking about SPELL_EFFECT_HANDLE_HIT phase (the right one for this effect)

Aokromes:
Conflicting files
src/server/game/Entities/AreaTrigger/AreaTrigger.cpp

Aokromes:
@Traesh

matukaa:
Traesh is back! Yaay :D

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -96,13 +157,386 @@ void AreaTrigger::Update(uint32 p_time)
...
 void AreaTrigger::Remove()
 {
     if (IsInWorld())
     {
+        if (Unit* caster = GetCaster())
+            caster->_UnregisterAreaTrigger(this);
+
+        for (ObjectGuid insideUnitGuid : _insideUnits)
+            if (Unit* leavingUnit = ObjectAccessor::GetUnit(*this, insideUnitGuid))
+                sScriptMgr->OnAreaTriggerEntityUnitExit(this, leavingUnit);
```
joschiwald: that will not remove auras correctly, i think it is better to just do `HandleUnitEnterExit({});` here

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -96,13 +157,386 @@ void AreaTrigger::Update(uint32 p_time)
...
+    for (Unit* unit : newTargetList)
+    {
+        if (exitUnits.erase(unit->GetGUID()) == 0) // erase(key_type) returns number of elements erased
+        {
+            if (Player* player = unit->ToPlayer())
+                if (player->isDebugAreaTriggers)
+                    ChatHandler(player->GetSession()).PSendSysMessage("You entered areatrigger %u", GetTemplate()->Id);
+
+            AddAuras(unit);
+            sScriptMgr->OnAreaTriggerEntityUnitEnter(this, unit);
```
joschiwald: you can't use GetInsideUnits() in this hook, i think it is better so add the unit to a enterUnits container and handle hooks and auras after all units are reinserted in _insideUnits

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -6647,3 +6647,23 @@ void AuraEffect::HandlePlayScene(AuraApplication const* aurApp, uint8 mode, bool
...
+{
+    if (!(mode & AURA_EFFECT_HANDLE_REAL))
+        return;
+
+    Unit* target = aurApp->GetTarget();
+
+    if (apply)
+    {
+        AreaTrigger* areaTrigger = new AreaTrigger();
+        if (!areaTrigger->CreateAreaTrigger(GetMiscValue(), GetCaster(), target, GetSpellInfo(), target->GetPosition()))
```
joschiwald: what about spellvisual?

Seyden:
Maybe i did something wrong, but i tested areatrigger with splines and the position i gave my areatrigger got sended.. but he did nothing.. he spawned, stood on his spawn position and then despawned...

Correct spline position got sended when i debugged the send method

matukaa:
@Seyden same here. it did move server-side though. possibly we are doing something wrong?

src/server/game/Entities/Object/Object.cpp
```diff
@@ -568,42 +564,35 @@ void Object::BuildMovementUpdate(ByteBuffer* data, uint32 flags) const
...
-        {
-            *data << float(0); // X
-            *data << float(0); // Y
-            *data << float(0); // Z
-        }
+            *data << G3D::Vector3::zero();
 
         if (hasScaleCurveID)
-            *data << int32(areaTriggerMiscTemplate->ScaleCurveId);
+            *data << uint32(areaTriggerMiscTemplate->ScaleCurveId);
```
Shauren: @joschiwald curve ids are signed in packet (not that it really matters, but i would like to preserve original structure signs)

joschiwald: https://gist.github.com/Vincent-Michael/6e8a1e9d4cc085f1b4e7b88deb969ffc#file-gistfile1-c-L82-L109

Shauren: nvm

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -57,34 +67,65 @@ void AreaTrigger::RemoveFromWorld()
...
-    SetObjectScale(1);
+
+    SetObjectScale(1.0f);
 
     SetGuidValue(AREATRIGGER_CASTER, caster->GetGUID());
+    SetGuidValue(AREATRIGGER_CREATING_EFFECT_GUID, castId);
+
     SetUInt32Value(AREATRIGGER_SPELLID, spell->Id);
     SetUInt32Value(AREATRIGGER_SPELL_X_SPELL_VISUAL_ID, spellXSpellVisualId);
     SetUInt32Value(AREATRIGGER_DURATION, spell->GetDuration());
```
matukaa: Should be Int32Value, since duration can be -1 ([proof](http://www.wowhead.com/spell=157411/windwalking))
Same for TimeToTarget, and TimeToTargetScale, since you are defining them after Duration.

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -96,13 +137,415 @@ void AreaTrigger::Update(uint32 p_time)
         Remove(); // expired
 
     WorldObject::Update(p_time);
+    _timeSinceCreated += p_time;
+
+    if (GetTemplate()->HasFlag(AREATRIGGER_FLAG_HAS_ATTACHED))
+    {
+        if (Unit* target = GetTarget())
+            GetMap()->AreaTriggerRelocation(this, target->GetPositionX(), target->GetPositionY(), target->GetPositionZ(), target->GetOrientation());
```
matukaa: What if target changes map? it can also happen ([same spell again](http://www.wowhead.com/spell=157411/windwalking))

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -96,13 +137,415 @@ void AreaTrigger::Update(uint32 p_time)
         Remove(); // expired
 
     WorldObject::Update(p_time);
+    _timeSinceCreated += p_time;
```
matukaa: Should check for negative duration, and skip checks for it @line 134 (it will instantly remove areatriggers with duration = -1)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -4899,6 +4899,59 @@ void Unit::RemoveAllGameObjects()
...
+{
+    if (m_areaTrigger.empty())
+        return;
+    for (AreaTriggerList::iterator i = m_areaTrigger.begin(); i != m_areaTrigger.end();)
+    {
+        AreaTrigger* areaTrigger = *i;
+        if (areaTrigger->GetSpellId() == spellId)
+        {
+            areaTrigger->Remove();
+            i = m_areaTrigger.erase(i);
```
matukaa: will crash the server immediately (invoked from SpellAuraEffects::HandleCreateAreaTrigger, due to the fact that areaTrigger->Remove() calls _UnregisterAreaTrigger, which removes the trigger from m_areaTrigger, thus modifying the vector, erase(i) will fail and crash the server. I suggest using
`i = m_areaTrigger.begin()`

matukaa (changes requested):

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
joschiwald: duration handling is incomplete

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
joschiwald: GetTarget() returns null on map change

what happens on retail on map change?

src/server/game/Entities/Unit/Unit.cpp
joschiwald: yeah i wrote that in one of the other hundred comments :D

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
matukaa: Well, I will go and sniff, would that help? But the triggers is moved with the caster. Also, it is not always created client-side, which is really weird. I tried sniffin Windwalking, but there was absolutely no smsg_update_object related to it. I got an aura update (157411 -> aura areatrigger), and immediately after it, i get the aura update for the movespeed buff spell, nothing between them, after them related to areatrigger ><

joschiwald: no, it is the aura target, do you have any example where it must be the caster?

matukaa: Sorry, i am clumsy, it is moved with the aura target ofcourse :) should we reapply such auras after the target is added to the map?

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
joschiwald: fixed

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -187,7 +197,17 @@ void AreaTrigger::SetDuration(int32 newDuration)
...
+    // negative duration (permanent areatrigger) sent as 0
+    SetUInt32Value(AREATRIGGER_DURATION, std::max(newDuration, 0));
+}
+
+void AreaTrigger::_UpdateDuration(int32 newDuration)
+{
+    _duration = newDuration;
+
+    // don't broadcast update
+    UpdateUInt32Value(AREATRIGGER_DURATION, _duration);
```
Shauren: but UpdateUInt32Value sends the value (always, it skips the check if new value is different to old one)

joschiwald: and it does not do `AddToObjectUpdateIfNeeded()`

Shauren: Its strange indeed but if another field changes then the value will be sent

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -5613,7 +5613,7 @@ void Spell::EffectCreateAreaTrigger(SpellEffIndex /*effIndex*/)
         return;
 
     AreaTrigger* areaTrigger = new AreaTrigger();
-    if (!areaTrigger->CreateAreaTrigger(effectInfo->MiscValue, GetCaster(), nullptr, GetSpellInfo(), destTarget->GetPosition(), m_SpellVisual, m_castId))
+    if (!areaTrigger->CreateAreaTrigger(effectInfo->MiscValue, GetCaster(), nullptr, GetSpellInfo(), destTarget->GetPosition(), GetSpellInfo()->GetDuration(), m_SpellVisual, m_castId))
```
matukaa: Incorrect behavior due to SPELLMOD_DURATION.
Consecration's base duration (DurationEntry->Duration, which is returned by GetSpellInfo()->GetDuration()) is 9000 ms, but learning 3 ranks of http://www.wowhead.com/spell=209218/consecration-in-flame will extend it to 12000 ms. It will cause the periodic dummy to deal damage 12 times, but the visual of consecration fades after 9 seconds

matukaa: I wonder why is this outdated by default

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -5612,17 +5612,11 @@ void Spell::EffectCreateAreaTrigger(SpellEffIndex /*effIndex*/)
...
-        pos = destTarget->GetPosition();
-
-    // trigger entry/miscvalue relation is currently unknown, for now use MiscValue as trigger entry
-    uint32 triggerEntry = effectInfo->MiscValue;
+        return;
 
-    AreaTrigger * areaTrigger = new AreaTrigger;
-    if (!areaTrigger->CreateAreaTrigger(GetCaster()->GetMap()->GenerateLowGuid<HighGuid::AreaTrigger>(), triggerEntry, GetCaster(), GetSpellInfo(), pos, m_SpellVisual))
+    AreaTrigger* areaTrigger = new AreaTrigger();
+    if (!areaTrigger->CreateAreaTrigger(effectInfo->MiscValue, GetCaster(), nullptr, GetSpellInfo(), destTarget->GetPosition(), GetSpellInfo()->GetDuration(), m_SpellVisual, m_castId))
```
matukaa: Incorrect behavior due to SPELLMOD_DURATION.
Consecration's base duration (DurationEntry->Duration, which is returned by GetSpellInfo()->GetDuration()) is 9000 ms, but learning 3 ranks of http://www.wowhead.com/spell=209218/consecration-in-flame will extend it to 12000 ms. It will cause the periodic dummy to deal damage 12 times, but the visual of consecration fades after 9 seconds

matukaa (changes requested):

src/server/game/Spells/SpellEffects.cpp
joschiwald: the hole spell duration stuff is bad

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
joschiwald: i want to send the updated duration in create object packets

src/server/game/Spells/SpellEffects.cpp
matukaa: hm :/ do you have any better idea, or at least any idea how is it handled on retail. I doubt that they set TimeToTarget explicitly for all areatriggers... that is not even affected by these spellmods. it has to be related to spell duration. what is it exactly that you don't like in the current handling method?

joschiwald: we are calculating spell duration everywhere in the code, somewhere incomplete, somewhere incorrect and somewhere in strange ways

all i know about TimeToTarget is that it is the current duration and TimeToTargetScale is max duration

matukaa: Yeah, you are right about duration calculations :/ although for Areatriggers, I believe it would be enough applying SPELLMOD_DURATION to get the final duration. So far I have never encountered an areatrigger that had a duration different from the spell's duration (only spellmod_duration affects it, haste doesn't, etc).

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
matukaa: Any ideas about this one? Shauren did a [commit](https://github.com/TrinityCore/TrinityCore/commit/bc758991140e961ceac46b17b766b0d54b8129fb) for fixing spellmods after map change, I thought we could do something similar. Ofcourse, not inside that function, since it's for packets, but before/after calling it? Would the tsrget still be null?

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -2431,7 +2431,8 @@ void ScriptMgr::OnAreaTriggerEntityUpdate(AreaTrigger* areaTrigger, uint32 diff)
     GET_SCRIPT(AreaTriggerEntityScript, areaTrigger->GetScriptId(), tmpscript);
     tmpscript->OnUpdate(areaTrigger, diff);
 }
-
+// place include here to signify its debug intention only!!!
+#include "AreaTriggerTemplate.h"
```
joschiwald: can you explain it?

i'm going to change
```
player->SummonCreature(WORLD_TRIGGER, areaTrigger->GetPosition(), TEMPSUMMON_TIMED_DESPAWN, areaTrigger->GetMiscTemplate()->TimeToTarget);
```
to get the TimeToTarget directly from areatrigger, because it can change during runtime
can i remove the include then?

PR #18057, opened 2016-10-08
Core/Spells: Effect leap back fix

Krudor:
**Changes proposed:**
-  Change SPELL_EFFECT_LEAP_BACK so that the victim of the spell effect leaps, and not the caster

**Issues addressed:** Found no issues related to fix

**Tests performed:** Builds, tested with spells that seemed to be related to the issue and they properly worked after the suggested code change

**Known issues and TODO list:** No known issues or todo's


tkrokli:
Nice fix, thank you for this contribution. 👍

PR #18077, opened 2016-10-12
Clean up locale functionality to proper add the case of enGB.

lbalan791:
**Changes proposed:**
-  Clean up locale functionality to proper add the case of enGB.
-  Eliminate an ugly locale check for enGB that would block specific clients from logging in.
-  Add error state to the GetLocaleByName so it does not return enUS in case of error ( it was a hack before ).

**Target branch(es):** 6.x

**Tests performed:** (Does it build, tested in-game, etc) Yes. A test for enGB locale client should be done.

- [ ] Test a client connection with enGB locale.

### DIFF
--- a/src/common/Common.cpp
+++ b/src/common/Common.cpp
@@ -18,8 +18,9 @@
 
 #include "Common.h"
 
-char const* localeNames[TOTAL_LOCALES] =
+char const* localeNames[TOTAL_LOCALES+1] =
 {
+  "enGB",
   "enUS",
   "koKR",
   "frFR",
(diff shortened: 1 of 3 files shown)


Shauren:
Having to distinct enum names for the same value (0) does nothing. In C++ after compilation enums are just integers.
enGB locale does not exist anymore since 6.0, it is handled purely clientside so no client should ever send "enGB" string in that packet (I use the locale that was once known as enGB but client sends enUS in there)

lbalan791:
Thank you for the quick feedback. 

Agree. The LOCALE_enGB was there for readability. Didn't know enGB is not used. 
The main issue was the default enUS set and the check required after: 
if (GetLocaleByName(logonRequest->locale()) == LOCALE_enUS && logonRequest->locale() != "enUS")

Added a new commit which addresses only this part.

PR #18093, opened 2016-10-16
Core/Totem: Remove aura due to spell glyph of stoneclaw shield

hcvcastro:
**Changes proposed:**
When the stoneclaw totem is destroyed it also should destroy glyph of stoneclaw shield
- When the stoneclaw totem is destroyed it destroy glyph of stoneclaw shield

**Issues addressed:**

### DIFF
--- a/src/server/game/Entities/Totem/Totem.cpp
+++ b/src/server/game/Entities/Totem/Totem.cpp
@@ -123,6 +123,10 @@ void Totem::UnSummon(uint32 msTime)
     if (GetEntry() == SENTRY_TOTEM_ENTRY)
         GetOwner()->RemoveAurasDueToSpell(SENTRY_TOTEM_SPELLID);
 
+    // Remove Glyph of Stoneclaw shield
+    if (GetEntry() == STONECLAW_TOTEM_ENTRY)
+        GetOwner()->RemoveAurasDueToSpell(55277);
+
     //remove aura all party members too
     if (Player* owner = GetOwner()->ToPlayer())
     {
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Totem/Totem.cpp
```diff
@@ -123,6 +123,10 @@ void Totem::UnSummon(uint32 msTime)
     if (GetEntry() == SENTRY_TOTEM_ENTRY)
         GetOwner()->RemoveAurasDueToSpell(SENTRY_TOTEM_SPELLID);
 
+    // Remove Glyph of Stoneclaw shield
+    if (GetEntry() == STONECLAW_TOTEM_ENTRY)
+        GetOwner()->RemoveAurasDueToSpell(55277);
```
joschiwald: magic number

Keader:
Fixing the magic number, looks good to me :)

Leastra:
The shield of the glyph of stoneclaw totem should **not** be destroyed when the stoneclaw totem is destroyed.

http://www.wowhead.com/forums&topic=90563/stoneclaw-totem-glyph : 
"You don't even need to keep the totem after it's been placed, you can switch to another Earth totem safely. The shield will stick to you, it's not linked to the totem."

http://www.dual-boxing.com/archive/index.php/t-25179.html
"It absorbs around 4k dmg after you drop the totem. You can immediately replace the stoneclaw totem with others (earthbind, tremor). And not lose the buff. "

http://www.arenajunkies.com/topic/137848-shaman-stoneclaw-totem/
"Keep in mind that you get the shield buff as soon as you put it down and don't need the totem down to maintain it. As such you can put it down and then redrop other valuable totems such as tremor/ earthbind for your team. "

http://www.wowhead.com/forums&topic=179126/my-new-favourite-totem-stoneclaw
"you don’t actually need to keep the Stoneclaw Totem up to continue to receive the shield"

PR #18128, opened 2016-10-23
Core/Movement: Using player position extrapolation to smooth range checks

chaodhib:
**Changes proposed:**
When a player moves, the client only send the current position of the player at the beginning and at the end of the movement and every 0.5s in between. This is not enough if we require high precision of the player position, during range checking for example. 

However, to counter that limitation, we can extrapolate the player position using his last known position + his speed and direction. This is what I have implemented.

**Target branch(es):** 3.3.5. Probably useful for the other branches.

**Issues addressed:** For now, the solution greatly mitigate the problem described here: https://github.com/TrinityCore/TrinityCore/issues/17539#issuecomment-242581530. At term, I'm hoping it completely fixes it. Also, the extrapolation of the player movement could probably be useful to other parts of the game, like spell hit checking which will improve the player experience.

**Tests performed:** tested IG.

- So far, I've implemented the 2D side. 3D support isn't done yet.
- The solution can have a 1 to 5% inaccuracy because of lag and request processing time. I'm hoping to fix that by working on the time synchronization done between the client and the server (cf [this relevant read](http://www.mine-control.com/zack/timesync/timesync.html)).

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -508,6 +508,100 @@ void Unit::GetRandomContactPoint(const Unit* obj, float &x, float &y, float &z, 
         , GetAngle(obj) + (attacker_number ? (static_cast<float>(M_PI/2) - static_cast<float>(M_PI) * (float)rand_norm()) * float(attacker_number) / combat_reach * 0.3f : 0));
 }
 
+/*
+    The client only send the current position of moving players at the beginning and at the end of their movements and every 0.5s in between.
+    This is not enough if we require high precision. However, we can extrapolate players position using their last known position + their speed and direction.
+    Keep in mind that what this function evaluates is a projected position: It does not take into account any limitation imposed by the map!!
+*/
+Position Unit::GetCurrentPosition() const
+{
+    if (!m_movementInfo.HasMovementFlag(MOVEMENTFLAG_MASK_MOVING))
+        return GetPosition();
+
+    Position pos;
+    if (m_movementInfo.pitch != 0.0) // swimming or flying. 3D
+    {
+        // todo
+    }
+    else // 2D
+    {
+        float orientation = GetCurrentMovementOrientation();
+        float time = (getMSTime() > m_movementInfo.time) ? getMSTime() - m_movementInfo.time : 0.0f; // in ms
+        float distancetTraveled = GetCurrentSpeed() * time/1000.0f;
+        pos.m_positionX = GetPositionX() + distancetTraveled * std::cos(orientation);
+        pos.m_positionY = GetPositionY() + distancetTraveled * std::sin(orientation);
+        pos.m_positionZ = GetPositionZ();
+    }
+
+    return pos;
+}
+
+float Unit::GetCurrentMovementOrientation() const
+{
+    float orientation = m_movementInfo.pos.GetOrientation();
+    if (!m_movementInfo.HasMovementFlag(MOVEMENTFLAG_MASK_MOVING))
+        return orientation;
+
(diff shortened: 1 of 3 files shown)


chaodhib:
@Raydor Could you test this and see if it helps?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -508,6 +508,100 @@ void Unit::GetRandomContactPoint(const Unit* obj, float &x, float &y, float &z,
...
+        else
+            return GetSpeed(UnitMoveType::MOVE_FLIGHT);
+    }
+    else
+    {
+        if (m_movementInfo.HasMovementFlag(MOVEMENTFLAG_BACKWARD))
+            return GetSpeed(UnitMoveType::MOVE_RUN_BACK);
+        else
+            return GetSpeed(UnitMoveType::MOVE_RUN);
+    }
```
joschiwald: checkout Movement::SelectSpeedType

Raydor:
I can't implement it on my server, im not there anymore and im not in any project because I don't and won't have free time until a lot of months, sorry :/

chaodhib:
No worries

chaodhib:
@Shauren 
Indeed. Also, the extrapolation should also be able to handle slopes and uneven landscape. Otherwise, its usefulness will be quite limited. I hope this wont be too resource intensive.

Shauren:
If you include terrain it very much will be resource intensive

chaodhib:
After analysing a fair share of sniffs, I found no evidence that the server actually runs the physics simulation (updating the player positions on each game tick, including collisions with terrain, based on the player's keyboard/mouse inputs). It's commonly done in FPS games (Valve games, Halo, Unreal Tournament, ..). In WoW however, it will be too CPU intensive (which is fine since the precision on player positions is not as important in MMORPGs as to FPS games).

In conclusions, I'm 99% sure my changes are wrong. I'm dropping this.

Shauren:
Player position is actually extrapolated on retail, you can see this in pet movement when running into closed gates, like the starting gate in Arathi Basin bg

Warpten:
its also extrapolated for charge - witnessed it more times than i can count on warrs charging rogues

chaodhib:
@Shauren I'm not sure I see what you mean. Could you please elaborate?

@Warpten Goods point. I'll look more into it . Thanks.

Shauren:
@chaodhib Pets move a few yards forward (beyond the gate) while you are stuck on the closed gate

ryancheung:
@chaodhib @Shauren  I'm interested in the 3D extrapolation support. 

Current movement updating mechanism can't satisfy some situations. For example, weird mob chasing in Mage aoe farming:
- Mob can melee mage even the distance between mob and player is large than 5yards, because we have 0.5s position update delay.
- Mobs may be scattered at most of the time on chasing instead of all mobs staying at one point, or staying in a line, also because of the player position update delay.

In other case:
- Spell distance check on casting may be inaccurate
- Melee spell cant use on target even target is in melee rage
- AOE spell should hit targets in radius but sometimes target would get hit even in the radius

So I think only we do player position extrapolation in server can fix these issues. I'd like to implement the 3D extrapolation, but currently I don't have a available way to do it. The mainly question is how to do the 3D extrapolation when terrain(slopes, uneven landscapes, walls) or physic gravity is accounted.

Could you guys give me any insight of these?

Raydor:
@chaodhib Sorry for the late response, got really busy last months. Just tested your patch, like you said, it completely fixes the issue I reported. I don't mind having things not blizzlike in my source so I have a question, can this patch as the state as it is generate other bugs? Do I have to worry about other things?

PR #18136, opened 2016-10-25
Core/Unit: Correct speed calculus when affected by SPELL_AURA_MOD_MINIMUM_SPEED

### LINKED ISSUE #4169: [Spell] Wings of the Damned
Hello,

There seems to be a slight bug, with handling of spell auras 305 (SPELL_AURA_MOD_MINIMUM_SPEED)

Right now, the code states:

``` c++
if (float minSpeedMod = (float)GetMaxPositiveAuraModifier(SPELL_AURA_MOD_MINIMUM_SPEED))
{
        float min_speed = minSpeedMod / 100.0f;
        if (speed < min_speed)
        speed = min_speed;
}
```

The point is, the value of the "Minimum speed" is taken as an absolute value, instead of as value relative to the unit's base speed rate.

ie. If we have two units,
unit A with base speed rate 1.0,
unit B with base speed rate 0.6, 

And we apply a -70% slow to both of them,

unit A will be slowed only by 50% ( down to 0.5 ), which is correct, but 
unit B will be slowed down to 0.5, instead of 50% of 0.6 ( 50% of base speed rate ).

How to quickly test in practice:
Place unit A and unit B, apply a slow debuff on both of them ( ie. Cone of Cold ), observe how they both run at same speed.

Note: I'm not 100% about correct expected behaviour, couldn't verify for or against it.

Thank you.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65665-spell-wings-of-the-damned?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sirikfoll:
**Changes proposed:**
New aproach is from the patch in https://github.com/TrinityCore/TrinityCore/issues/4169#issuecomment-2989390

**Issues addressed:** Closes #4169

**Tests performed:** Built and tested during the Lich King fight, with the Valkyr Shadowguard

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9268,7 +9268,7 @@ void Unit::UpdateSpeed(UnitMoveType mtype)
 
     if (float minSpeedMod = (float)GetMaxPositiveAuraModifier(SPELL_AURA_MOD_MINIMUM_SPEED))
     {
-        float min_speed = minSpeedMod / 100.0f;
+        float min_speed = CalculatePct(GetTypeId() == TYPEID_UNIT ? ToCreature()->GetCreatureTemplate()->speed_run : 1.0f, minSpeedMod);
         if (speed < min_speed)
             speed = min_speed;
     }
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9269,7 +9269,7 @@ void Unit::UpdateSpeed(UnitMoveType mtype)
     if (float minSpeedMod = (float)GetMaxPositiveAuraModifier(SPELL_AURA_MOD_MINIMUM_SPEED))
     {
         float baseMinSpeed = 1.0f;
-        if (!(GetOwnerGUID().IsPlayer() || IsHunterPet()) && GetTypeId() == TYPEID_UNIT)
+        if (!GetOwnerGUID().IsPlayer() && !IsHunterPet() && GetTypeId() == TYPEID_UNIT)
             baseMinSpeed = ToCreature()->GetCreatureTemplate()->speed_run;
 
         float min_speed = CalculatePct(baseMinSpeed, minSpeedMod);
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9268,7 +9268,11 @@ void Unit::UpdateSpeed(UnitMoveType mtype)
 
     if (float minSpeedMod = (float)GetMaxPositiveAuraModifier(SPELL_AURA_MOD_MINIMUM_SPEED))
     {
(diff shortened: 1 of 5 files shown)


ccrs:
check the interaction with the shaman spirit wolf spell and its related talent

sirikfoll:
@ccrs  it definitely broke that :/

I know that set the Valkyr speed directly in the database may be considered a hack, but that, along with the second edit in unit.cpp, would solve the problem without break anything.
Any comments?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9166,7 +9166,7 @@ void Unit::UpdateSpeed(UnitMoveType mtype)
 
     if (float minSpeedMod = (float)GetMaxPositiveAuraModifier(SPELL_AURA_MOD_MINIMUM_SPEED))
     {
-        float min_speed = minSpeedMod / 100.0f;
+        float min_speed = (minSpeedMod / 100.0f) * GetSpeedRate(MOVE_RUN);
```
P-Kito: I suggest:
```cpp
CalculatePct(GetSpeedRate(MOVE_RUN), minSpeedMod);
```

Aokromes:
@sirikfoll this can be merged?

Shauren:
This is definitely incorrect in its current form. It breaks in case when movement speed slow aura is applied before SPELL_AURA_MOD_MINIMUM_SPEED. You would be treating the slowed speed as "normal"

P-Kito:
Maybe introduce a new function that subtracts the negative speed mods and positive speed mods of a speedvalue. Then there would only be the DB value with scripted changes left. Just a random idea...

Keader:
Any feedback here guys?

Noryad:
Related to Valkyr Shadowguard speed is ok.

Shauren:
Player pets need to ignore template base speed - same thing as with resistances applies here

Noryad:
like 949b38b?

sirikfoll:
Is it correct now?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9268,7 +9268,11 @@ void Unit::UpdateSpeed(UnitMoveType mtype)
 
     if (float minSpeedMod = (float)GetMaxPositiveAuraModifier(SPELL_AURA_MOD_MINIMUM_SPEED))
     {
-        float min_speed = minSpeedMod / 100.0f;
+        float baseMinSpeed = 1.0f;
+        if (!(GetOwnerGUID().IsPlayer() || IsHunterPet()) && GetTypeId() == TYPEID_UNIT)
```
Shauren: please, dont try putting the negation outside, it makes the code hard to decipher

Shauren:
Is good now.

PR #18151, opened 2016-10-27
Core/Conditions: merge CONDITION_SOURCE_TYPE_QUEST_SHOW_MARK with CONDITION_SOURCE_TYPE_QUEST_ACCEPT.

SnapperRy:
**Changes proposed:**

Title says it all.

CONDITION_SOURCE_TYPE_QUEST_SHOW_MARK is basically useless, since it makes no sense for a quest to be available without its mark being visible (only low level quests have this behavior, which is handled elsewhere).

Right now it's just an exact copy of CONDITION_SOURCE_TYPE_QUEST_ACCEPT, making it rendundant and confusing.

Some quests' availability was wrongly fixed using CONDITION_SOURCE_TYPE_QUEST_SHOW_MARK alone. This has been corrected in 95f2aed.

**Target branch(es):** 3.3.5

**Tests performed:** tested 'n' working.


Shauren:
good! i always wondered why the fuck was that even added like that

PR #18172, opened 2016-10-30
Core/GameObject: spawn linked traps at gameobject creation instead of using DB spawns.

### LINKED ISSUE #11328: Dreaming Glory BUFF
https://github.com/TrinityCore/TrinityCore/commit/e5f7beecf5fabefa13fbe2645cc7d7ca707e46e8

when herb Dreaming Glory obj:181271  u must buffed spell:28694
but bug
http://www.wowhead.com/object=181271#comments
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1412715-dreaming-glory-buff?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SnapperRy:
**Changes proposed:**

Title says it all. Recent changes only affect gameobjects spawned via spell effects, but there are also a lot of spawned gameobjects that should also spawn a linked trap and respawn/despawn it.

GUID list comes from:

``` sql
SELECT `guid` FROM `gameobject`
 WHERE `id` IN (SELECT `data7`  FROM `gameobject_template` WHERE `type`=3)
    OR `id` IN (SELECT `data3`  FROM `gameobject_template` WHERE `type`=1)
    OR `id` IN (SELECT `data2`  FROM `gameobject_template` WHERE `type`=8)
    OR `id` IN (SELECT `data12` FROM `gameobject_template` WHERE `type`=10)
 ORDER BY `guid`;
```

A bit ugly, but apparently the gameobject table doesn't like delete queries by id.

Closes #11328.

**Target branch(es):** 3.3.5


Killyana:
Good Idea, it's better than spawning hundreds of gobs https://github.com/TrinityCore/TrinityCore/issues/11328

SnapperRy:
Thanks for the issue link :P indeed this fixes all currently non-spawned traps as well.

runningnak3d:
Dunno if it is outside the scope of this PR, but spell focus GOs should be handled the same way.

Either way, awesome idea dude!

SnapperRy:
@runningnak3d this already handles traps linked to spellfocus GOs :P do you have any example of wrong behavior with that type?

runningnak3d:
durrr. Type=8 right there in your query. You rock dude :)

tkrokli:
Edited your SQL to make the contents more obvious. Does this PR affect issue #12139 in any positive way?

SnapperRy:
@tkrokli yep, it does fix that one too, albeit indirectly (main fix for that issue was 5d9d25d, but that only works for objects spawned by spells). Nice find :) was hard to find due to misleading title.

PR #18189, opened 2016-11-02
Core/Movement: Use time synchronisation to better interpret client timestamps

chaodhib:
**Changes proposed:**
- Implement [Simple Network Time Protocol](https://tools.ietf.org/html/rfc4330) (SNTP) using the `SMSG_TIME_SYNC_REQUEST `/  `CMSG_TIME_SYNC_RESPONSE` exchange. No need to read however the RFC. The mechanism is quite simple and it is explained in the code. This packet exchange initiated by the server provides to the server the following information: the difference between its own clock and the clock of the client. With that clock delta, the server can precisely interpret timestamps sent by the client.
- On top of SNTP, I implemented the approach described [HERE](https://web.archive.org/web/20180430214420/http://www.mine-control.com/zack/timesync/timesync.html). That approach is necessary to use on top of SNTP because plain SNTP is designed to work on UDP, not TCP. The fact that TCP can resend dropped packets without any warning requires extra care.
- Using the server newly found information, which is the clock delta, we can now interpret the only timestamps sent by the client to the server: The movement timestamps. Now, when movement events occur on the client, we can precisely know their time on server clock (such as the player has started to move forward, for example). And send these movement timestamps on server clock to the observing nearby players. This change should improve the visual perseption of the other players on screen.

**Target branch(es):** 3.3.5

**Issues addressed:** should improve the visual perseption of the other players on screen. Should help with https://github.com/TrinityCore/TrinityCore/issues/19271.

**Tests performed:** Build and tested IG with artificial latency and packet drop on my machine. Larger scale testing should be performed

Nothing that I'm aware of.

### DIFF
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -1239,7 +1239,7 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
     // we are going to make 2 assumptions:
     // 1) we assume that the request processing time equals 0.
     // 2) we assume that the packet took as much time to travel from server to client than it took to travel from client to server.
-    uint32 roundTripDuration = getMSTimeDiff(serverTimeAtSent, getMSTime());
+    uint32 roundTripDuration = getMSTimeDiff(serverTimeAtSent, GameTime::GetGameTimeMS());
     uint32 lagDelay = roundTripDuration / 2;
 
     /*
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -1229,7 +1229,7 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
     uint32 counter, clientTimestamp;
     recvData >> counter >> clientTimestamp;
 
-    if (counter != _player->m_timeSyncCounter - 1)
+    if (counter != m_timeSyncCounter - 1)
     {
         TC_LOG_DEBUG("network", "Wrong time sync counter from player %s (cheater?)", _player->GetName().c_str());
         // todo: send a new one?
@@ -1237,8 +1237,11 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
     }
 
     // time it took for the request to travel to the client, for the client to process it and reply and for response to travel back to the server.
-    uint32 roundTripDuration = getMSTimeDiff(_player->m_timeSyncServer, getMSTime());
-    uint32 lagDelay = roundTripDuration / 2; // we assume that the request processing time equals 0.
+    // we are going to make 2 assumptions:
+    // 1) we assume that the request processing time equals 0.
+    // 2) we assume that the packet took as much time to travel from server to client than it took to travel from client to server.
+    uint32 roundTripDuration = getMSTimeDiff(m_timeSyncServer, getMSTime());
+    uint32 lagDelay = roundTripDuration / 2;
 
     /*
     clockDelta = serverTime - clientTime
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -1226,20 +1226,31 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
 {
     TC_LOG_DEBUG("network", "CMSG_TIME_SYNC_RESP");
 
-    uint32 counter, clientTicks;
-    recvData >> counter >> clientTicks;
+    uint32 counter, clientTimestamp;
+    recvData >> counter >> clientTimestamp;
 
     if (counter != _player->m_timeSyncCounter - 1)
+    {
         TC_LOG_DEBUG("network", "Wrong time sync counter from player %s (cheater?)", _player->GetName().c_str());
+        // todo: send a new one?
+        return;
+    }
 
-    TC_LOG_DEBUG("network", "Time sync received: counter %u, client ticks %u, time since last sync %u", counter, clientTicks, clientTicks - _player->m_timeSyncClient);
+    // time it took for the request to travel to the client, for the client to process it and reply and for response to travel back to the server.
+    uint32 roundTripDuration = getMSTimeDiff(_player->m_timeSyncServer, getMSTime());
+    uint32 lagDelay = roundTripDuration / 2; // we assume that the request processing time equals 0.
 
-    uint32 ourTicks = clientTicks + (GameTime::GetGameTimeMS() - _player->m_timeSyncServer);
+    /*
+    clockDelta = serverTime - clientTime
+    where
+    serverTime: time that was displayed on the clock of the SERVER at the moment when the client processed the SMSG_TIME_SYNC_REQUEST packet.
+    clientTime:  time that was displayed on the clock of the CLIENT at the moment when the client processed the SMSG_TIME_SYNC_REQUEST packet.
 
-    // diff should be small
-    TC_LOG_DEBUG("network", "Our ticks: %u, diff %u, latency %u", ourTicks, ourTicks - clientTicks, GetLatency());
-
-    _player->m_timeSyncClient = clientTicks;
+    Once clockDelta has been computed, we can compute the time of an event on server clock when we know the time of that same event on the client clock,
+    using the following relation:
+    serverTime = clockDelta + clientTime
+    */
+    _player->m_timeSyncClockDelta = (int64) (_player->m_timeSyncServer + lagDelay) - (int64) clientTimestamp;
 }
 
 void WorldSession::HandleResetInstancesOpcode(WorldPacket& /*recvData*/)
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -1229,18 +1229,17 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
     uint32 counter, clientTimestamp;
     recvData >> counter >> clientTimestamp;
 
-    if (counter != m_timeSyncCounter - 1)
-    {
-        TC_LOG_DEBUG("network", "Wrong time sync counter from player %s (cheater?)", _player->GetName().c_str());
-        // todo: send a new one?
+    if (pendingTimeSyncRequests.count(counter) == 0)
         return;
-    }
+
+    uint32 serverTimeAtSent = pendingTimeSyncRequests.at(counter);
+    pendingTimeSyncRequests.erase(counter);
 
     // time it took for the request to travel to the client, for the client to process it and reply and for response to travel back to the server.
     // we are going to make 2 assumptions:
     // 1) we assume that the request processing time equals 0.
     // 2) we assume that the packet took as much time to travel from server to client than it took to travel from client to server.
-    uint32 roundTripDuration = getMSTimeDiff(m_timeSyncServer, getMSTime());
+    uint32 roundTripDuration = getMSTimeDiff(serverTimeAtSent, getMSTime());
     uint32 lagDelay = roundTripDuration / 2;
 
     /*
@@ -1253,10 +1252,11 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
     using the following relation:
     serverTime = clockDelta + clientTime
     */
-    int64 clockDelta = (int64)(m_timeSyncServer + lagDelay) - (int64)clientTimestamp;
+    int64 clockDelta = (int64)(serverTimeAtSent + lagDelay) - (int64)clientTimestamp;
     timeSyncClockDeltaQueue.push_back(std::pair<int64, uint32>(clockDelta, roundTripDuration));
     ComputeNewClockDelta();
 
+    TC_LOG_ERROR("custom", "CMSG_TIME_SYNC_RESP. counter %u clientTimestamp %u serverTimeAtSent %u", counter, clientTimestamp, serverTimeAtSent);
     TC_LOG_ERROR("custom", "CMSG_TIME_SYNC_RESP. roundTripDuration %u", roundTripDuration);
     TC_LOG_ERROR("custom", "CMSG_TIME_SYNC_RESP. new delta: %i for player  %s", timeSyncClockDelta, _player->GetName().c_str());
 }
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -1239,7 +1239,7 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
     // we are going to make 2 assumptions:
     // 1) we assume that the request processing time equals 0.
     // 2) we assume that the packet took as much time to travel from server to client than it took to travel from client to server.
-    uint32 roundTripDuration = getMSTimeDiff(serverTimeAtSent, GameTime::GetGameTimeMS());
+    uint32 roundTripDuration = getMSTimeDiff(serverTimeAtSent, recvData.GetReceivedTime());
     uint32 lagDelay = roundTripDuration / 2;
 
     /*
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -1255,10 +1255,6 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
     int64 clockDelta = (int64)(serverTimeAtSent + lagDelay) - (int64)clientTimestamp;
     timeSyncClockDeltaQueue.push_back(std::pair<int64, uint32>(clockDelta, roundTripDuration));
     ComputeNewClockDelta();
-
-    TC_LOG_ERROR("custom", "CMSG_TIME_SYNC_RESP. counter %u clientTimestamp %u serverTimeAtSent %u", counter, clientTimestamp, serverTimeAtSent);
-    TC_LOG_ERROR("custom", "CMSG_TIME_SYNC_RESP. roundTripDuration %u", roundTripDuration);
-    TC_LOG_ERROR("custom", "CMSG_TIME_SYNC_RESP. new delta: %i for player  %s", timeSyncClockDelta, _player->GetName().c_str());
 }
 
 void WorldSession::HandleResetInstancesOpcode(WorldPacket& /*recvData*/)
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -1250,7 +1250,12 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
     using the following relation:
     serverTime = clockDelta + clientTime
     */
-    _player->m_timeSyncClockDelta = (int64) (_player->m_timeSyncServer + lagDelay) - (int64) clientTimestamp;
+    int64 clockDelta = (int64)(m_timeSyncServer + lagDelay) - (int64)clientTimestamp;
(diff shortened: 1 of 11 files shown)


Aokromes:
TrinityCore/TrinityCore/src/server/game/Handlers/MovementHandler.cpp:353:162: fatal error: 
      format specifies type 'int' but the argument has type 'int64' (aka 'long')
      [-Wformat]
  ...Clock delta: %d", movementInfo.time, plrMover->m_timeSyncClockDelta);
                  ~~                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                  %ld

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -345,14 +345,15 @@ void WorldSession::HandleMovementOpcodes(WorldPacket& recvData)
...
     }
 
-    uint32 mstime = getMSTime();
-    /*----------------------*/
-    if (m_clientTimeDelay == 0)
-        m_clientTimeDelay = mstime - movementInfo.time;
-
     /* process position-change */
     WorldPacket data(opcode, recvData.size());
-    movementInfo.time = movementInfo.time + m_clientTimeDelay + MOVEMENT_PACKET_TIME_DELAY;
```
sucklifex: Please also remove this #define MOVEMENT_PACKET_TIME_DELAY 0 at line 35

jackpoz:
what kind of tests have been performed ?

jackpoz:
isn't that you want to revert the whole https://github.com/chaodhib/TrinityCore/commit/2cca3982d4dd87be452ab13332dd809e02d2a24b too ? You removed every use of m_clientTimeDelay but kept the class member field

Palabola:
@Jackpoz I'm not sure, but since 5.X this delaytime can cause really stange desyncronization. Mainly because the current movement system use much less packet number then before so 1-2 wrong timed packet can cause seeable problems.

jackpoz:
@Palabola this PR is for 335 tho

chaodhib:
@jackpoz  `isn't that you want to revert the whole chaodhib@2cca398 too ? You removed every use of m_clientTimeDelay but kept the class member field`
Didn't saw the commit but yeah pretty much. I removed the field now.

Regarding the testing done: I havent done much so far. i just verified that the values (round trip delay and the clock delta) computed made sense. Do you have tests in mind that I could do?

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1448,16 +1448,27 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
...
-    TC_LOG_DEBUG("network", "Our ticks: %u, diff %u, latency %u", ourTicks, ourTicks - clientTicks, GetLatency());
+    // time it took for the request to travel to the client, for the client to process it and reply and for response to travel back to the server.
+    uint32 roundTripDuration = getMSTimeDiff(_player->m_timeSyncServer, getMSTime());
+    uint32 lagDelay = roundTripDuration / 2; // we assume that the request processing time equals 0.
 
-    _player->m_timeSyncClient = clientTicks;
+    /*
+    clockDelta = serverTime - clientTime
+    where
+    serverTime: time that was displayed on the cloak of the SERVER at the moment when the client processed the SMSG_TIME_SYNC_REQUEST packet.
```
tkrokli: cloak = a cape or drape, a sleeveless outdoor overgarment that hangs loosely from the shoulders.
clock = an instrument to indicate, keep, and co-ordinate time.

chaodhib: oups, thank you.

tkrokli: (same typo on the line below, obviously. :)

Nawuko:
@chaodhib The world server processes never ticks faster than 50ms. So thats the smallest possible delay you can get if the client response packet gets processed on the next tick.

chaodhib:
@Nawuko Yep, already gotten help from IRC. Thanks though.

jackpoz:
When I ported that old commit I did few tests with more than 1 client to verify client smooth movements. I think I used Valgrind to have high update time diff to simulate lag.

Palabola:
I preffer to use this: https://www.softperfect.com/products/connectionemulator/  for simulating high latency.

chaodhib:
@jackpoz I tried running with multiple clients and I see no change of behavior. This PR basically brings 2 things:
- To the server, it gives a more accurate time-stamp of the moment the player started his change of movement. As far as I can tell however, other than my changes proposed in https://github.com/TrinityCore/TrinityCore/pull/18128, `movementInfo.time` is unused by the server at the moment.
- To the client, it also give the same time-stamp but as I as I can tell, **the time-stamp contained in MSG_MOVE packets at destination to the client are unused by the client**

I'm gonna elaborate. Please challenge me and tell me if I'm wrong in something. My current understanding is as follow: The server sends (at least) 2 types of messages to the client that contains a time-stamp (I'm talking about 3.3.5 branch): SMSG_SPELL_GO and MSG_MOVE packets. And in both cases, the time-stamp is completely ignored by the client. This is specially true since as far as I can tell, the client has no way of interpreting that time-stamp because the client and the server run on a different clock and as opposed to the server, the client has no way of computing the clock delta between its clock and the server's clock.

The server regularly sends to each client a SMSG_TIME_SYNC_REQUEST packet. The client answer to it with a CMSG_TIME_SYNC_RESPONSE packet which contains the client current time-stamp at the moment he received the request. This exchange of packets has only one goal: it allows the server to compute the clock delta. With that information, each time the client sends to the server a packet containing a time-stamp (so far I know only of MSG_MOVE messages), the server can accurately deduce when the event described by the message occurred (aka when the player pressed a movement key on his computer).

tripleslash:
@chaodhib It allows the server to fix the timestamp in outgoing movement packets in order to enable proper movement extrapolation for other players (no more lagging when they mouse turn and move etc).

Shauren:
The biggest problem with this PR is that it assumes that request processing time is 0, which definitely is not the case for packets processed for players already in world. It can range anywhere between 0 to max(WORLD_SLEEP_CONST, last_update_diff)

chaodhib:
@tripleslash Yeah. The premise of my post above "And in both cases, the time-stamp is completely ignored by the client" is wrong. I proved it by sending to the client different MSG_MOVE packets with different time-stamp and there was a noticeable difference. If the timestamp was the server time, the character moved as soon as the client received the packet. If the timestamp was server time + 1000ms, the character moved after a delay (that I assume was 1s).

@Shauren I assume that the request processing time is 0 for only one packet: `SMSG_TIME_SYNC_REQUEST`. As soon as the client receives that packet, it just replies with a `CMSG_TIME_SYNC_RESPONSE` that contains the current time on the client clock. The time it takes to process that request is probably negligible most of the time.

Shauren:
@chaodhib but the packet needs to be queued from worldsocket to worldsession and it must be processed, that is the time i was talking about - the time it takes to go from WorldSocket::ReadDataHandler to WorldSession::HandleTimeSyncResp (its not a simple call chain)

jackpoz:
> If the timestamp was the server time, the character moved as soon as the client received the packet. If the timestamp was server time + 1000ms, the character moved after a delay (that I assume was 1s).

this sounds like the same research done maaaany years ago that led to the same commit I ported to TC
- https://github.com/TrinityCore/TrinityCore/commit/2cca3982d4dd87be452ab13332dd809e02d2a24b
- https://sourceforge.net/p/ascentemubackup/code/1637/

you can check the work done in the svn history of that repo if it helps.

chaodhib:
@Shauren True. I'm not sure what to do with that (major) issue. It looks like `CMSG_PING` packets are handled in a synchronous way. Maybe the same could be done for `CMSG_TIME_SYNC_RESPONSE` packets... I really don't have much knowledge about that part..

@jackpoz 
1. Your ported commit does not take into account fluctuating latency. Mine does. Better yet, if I implement the method described [here](http://www.mine-control.com/zack/timesync/timesync.html), it is even possible to compensate for the error provoked by the TCP layer resending a dropped packet.
2. The field `_player->m_timeSyncClient` is, until now, completely unused. In other words, the method `WorldSession::HandleTimeSyncResp` has no effect and the TIME_SYNC packets exchange amounts to nothing. 
3. I'm not yet sure if this change is blizz-like. This is still a work in progress but anyway: Each time a movement packet is sent from the client to the server, it is then broadcasted to the other clients. With this PR, the broadcasted packet's timestamp will be the time the movement packet was sent by the client **on server clock**, as opposed to the time the movement packet was processed on server clock.

Seyden:
if this is really fixing #19271 then merge itttt please :D

Aokromes:
Conflicting files
src/server/game/Handlers/MiscHandler.cpp
src/server/game/Handlers/MovementHandler.cpp
src/server/game/Server/WorldSession.cpp

chaodhib:
@Seyden It does not. And it's far from being ready to be merged.

SeTM:
what about CMSG_MOVE_TIME_SKIPPED?

chaodhib:
CMSG_MOVE_TIME_SKIPPED is sent by the client when it freezed for a certain period of time. Thus not running its internal clock during that time. The packet indicated the amount of time "lost". It's easy to see in sniffs since we know for sure that heartbeat packets are sent exactly 500ms after the last movement packet if the player is still moving. If the client freezed inbeween, the difference in the timestamp will be different then 500ms BUT only if you do not take into account the TIME_SKIPPED packet.

The implementation of the 2 TIME_SKIPPED packet is not very difficult. I will try to get into it asap (or someone else can and i will be happy to code review and merge) but it does not help with this PR.

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -345,14 +343,15 @@ void WorldSession::HandleMovementOpcodes(WorldPacket& recvData)
...
+    int64 movementTime = (int64) movementInfo.time + plrMover->m_timeSyncClockDelta; // time of the event on the server clock.
```
Shauren: And you probably should be using packet sender's clock delta, not the currently controlled player's (if mind controlled) since the timestamp comes from controllers client (not moved player)

dRaiNe:
I believe it's related to this: https://github.com/TrinityCore/TrinityCore/issues/10158

ArcEMU distributed/calculated the timing properly, back in WoW version 3.3.5 (or so). Maybe we could analyze its code, provided with the timing system didn't change over the years. I don't have WoW client at the moment.

akrom23:
@chaodhib 
What do you think about that: https://github.com/lh-server/core/blob/development/src/shared/WorldPacket.h#L65
Tested on local and works good. This timer is registered when the worldsocket received packet. so we can use it instead of ``GameTime::GetMsTime()`` to properly calculate, how long took the client to respond to SMSG packet - this resolves the main problem described by @Shauren .
But we can use this timer in every movement packet. so this timer in the server time, when the packet is received by client - we can assume that the clientstamp sent in packet is processed right here (in case of no client lag)?.

chaodhib:
@akrom23 
Thanks for the inspiration. I resumed working on this following your comment. In the end, I took the approach that we currently use to handle `CMSG_PING` packets: handle TIME_SYNC packets directly in `WorldSocket::ReadDataHandler()`.

With that, I added quite a lot of changes. Here is the new list of things provided by this PR:
- Implement SNTP using the `SMSG_TIME_SYNC_REQUEST `/  `CMSG_TIME_SYNC_RESPONSE` exchange. This packet exchange initiated by the server provides to the server the following information: the difference between its own clock and the clock of the client. With that clock delta, the server can precisely interpret timestamps sent by the client.
- On top of SNTP, I implemented the approach described [HERE](https://web.archive.org/web/20180430214420/http://www.mine-control.com/zack/timesync/timesync.html). That approach is necessary to use on top of SNTP because plain SNTP is designed to work on UDP, not TCP. The fact that TCP can resend dropped packets without any warning requires extra care.
- Using the server newly found information, which is the clock delta, we can now interpret the only timestamps sent by the client to the server: The movement timestamps. Now, when movement events occur on the client, we can precisely know their time on server clock (such as the player has started to move forward, for example). And send these movement timestamps on server clock to the observing nearby players. This change should improve the visual perseption of the other players on screen.

src/server/game/Server/WorldSocket.cpp
```diff
@@ -360,6 +360,29 @@ WorldSocket::ReadDataHandlerResult WorldSocket::ReadDataHandler()
             if (_worldSession)
                 _worldSession->ResetTimeOutTime(true);
             break;
+
+        case CMSG_TIME_SYNC_RESP:
```
chaodhib: Feedback on this touch part would be nice

jackpoz:
Does that SNTP stuff allow to know the timezone of the client (and save the calendar events with the right server time) ?

chaodhib:
@jackpoz I don't think so, no. I think the TIME_SYNC packet exchange is only useful to network player movement more smoothly. It only provide a clock delta, which allows the server to interpret in server clock timestamps that were sent in client clock. That's it. There is no notion of timezone.

chaodhib:
Let's merge this in two weeks if there is no opposition/review comments

jackpoz:
Check no-pch build please

src/server/game/Server/WorldSocket.cpp
Shauren: It's not good actually. You have a race on `WorldSession::timeSyncClockDelta` in two threads - the network thread that this socket is bound to and any map update thread where movement packets are processed

I assume you need the timestamp of when this packet was received to be as accurate as possible (without the delay from packet queue) so my suggestion would be to somehow pass the timestamp to regular packet queue, possibly by extending WorldPacket class. Marking some opcodes as "high priority" would get the timestamp with `getMSTime()` while regular ones would use lower precision `GameTime::GetGameTimeMS()`

Shauren (changes requested):

Shauren:
Also codestyle (naming WorldSession members), either use our default style or just m_ that stuff for consistency with existing code - don't make a third naming variant

src/server/game/Server/WorldSocket.cpp
chaodhib: > It's not good actually. You have a race on WorldSession::timeSyncClockDelta in two threads - the network thread that this socket is bound to and any map update thread where movement packets are processed

Indeed. Wouldn't using `std::atomic<int64> timeSyncClockDelta` instead of `int64 timeSyncClockDelta` fix the issue though? This is the way `CMSG_PING` and `CMSG_KEEP_ALIVE` are handled.

Though, adding a received timestamp in WorldPacket (or a subclass), as suggested by you and @akrom23, would work too and would be cleaner I suppose.

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1226,20 +1226,35 @@ void WorldSession::HandleTimeSyncResp(WorldPacket& recvData)
...
-    uint32 counter, clientTicks;
-    recvData >> counter >> clientTicks;
+    uint32 counter, clientTimestamp;
+    recvData >> counter >> clientTimestamp;
 
-    if (counter != _player->m_timeSyncCounter - 1)
-        TC_LOG_DEBUG("network", "Wrong time sync counter from player %s (cheater?)", _player->GetName().c_str());
-
-    TC_LOG_DEBUG("network", "Time sync received: counter %u, client ticks %u, time since last sync %u", counter, clientTicks, clientTicks - _player->m_timeSyncClient);
+    if (pendingTimeSyncRequests.count(counter) == 0)
```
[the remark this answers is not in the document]
chaodhib: This wouldn't really have any noticeable impact on the performance (this portion of the code is not in a hot path and all 3 operations are in log complexity). However, the current implementation is more readable, in my opinion. Better to keep it like this I think.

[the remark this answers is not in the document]
Shauren: I disagree with the readability part, anyone familiar with c++ should also be familiar with iterators and be able to immediately determine what is going on here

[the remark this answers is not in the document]
chaodhib: - We both agree that peformance wise, they are both the same.
- And we can probably both agree that using the public API of map is _at least_ as maintainable as using its iterator.

So both solutions are similar. Therefore, there is no reason left to spend any further time on this.

src/server/game/Server/WorldSocket.cpp
chaodhib: Done. The time when `CMSG_TIME_SYNC_RESP` packets are received are now recorded (before they are put into the packet queue). And then, they are handled like any other `PROCESS_THREADSAFE` packet.

chaodhib:
nopch build and codestyle fixed. Anything else?

src/server/game/Server/WorldSession.cpp
```diff
@@ -125,17 +125,22 @@ WorldSession::WorldSession(uint32 id, std::string&& name, std::shared_ptr<WorldS
...
-    m_clientTimeDelay(0),
     m_TutorialsChanged(TUTORIALS_FLAG_NONE),
     recruiterId(recruiter),
     isRecruiter(isARecruiter),
     _RBACData(nullptr),
     expireTime(60000), // 1 min after socket loss, session is deleted
     forceExit(false),
-    m_currentBankerGUID()
+    m_currentBankerGUID(),
+    _timeSyncClockDeltaQueue(boost::circular_buffer<std::pair<int64, uint32>>(6)),
```
Shauren: Why are you initializing by creating a temporary circular_buffer object instead of directly passing "6" to _timeSyncClockDeltaQueue constructor?

src/server/game/Maps/Map.cpp
```diff
@@ -753,8 +753,11 @@ void Map::Update(uint32 t_diff)
         {
             //player->Update(t_diff);
             WorldSession* session = player->GetSession();
+
             MapSessionFilter updater(session);
             session->Update(t_diff, updater);
+
+            session->UpdateForPlayersOnMap(t_diff);
```
Shauren: Isn't the point of updater/filter argument on WorldSession::Update to do exactly this? Execute code based on update context

chaodhib: i could move the code inside WorldSession::UpdateForPlayersOnMap to WorldSession::Update but I would have to use a dynamic cast on the `updater` reference:
````C++
try
{
    dynamic_cast<MapSessionFilter&>(updater);
    // Send time sync packet every 10s.
    if (_timeSyncTimer > 0)
    {
        if (diff >= _timeSyncTimer)
            SendTimeSync();
        else
            _timeSyncTimer -= diff;
    }
}
catch (std::bad_cast exp)
{
}
````

Wouldn't that have a noticable performance impact?

src/server/game/Server/WorldSession.cpp
chaodhib: fixed

src/server/game/Maps/Map.cpp
Shauren: You don't need to rely on dynamic_cast, you can place the code under `updater.ProcessLogout()` - thats the other function of the filter that determines context

chaodhib: Done. In that case, the virtual function `ProcessLogout` should probably be renamed into something more meaningful since i'm using it out of the log-out context.

Shauren: Yes, I already renamed it long time ago in master branch https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Server/WorldSession.h#L848

chaodhib: Done

chaodhib:
Merging this in 2 days.

jackpoz:
Could you wait until after the 15th ? We are planning on releasing April TDB for 335, after that it's the perfect time to merge long standing pull requests :)

chaodhib:
@jackpoz Sure, of course :)

jackpoz:
TDB has been released :)

chaodhib:
Merged. Thank you!

PR #18219, opened 2016-11-08
[master] Core/Misc: Define unknown quest type

Krudor:

**Changes proposed:**

-  Define previously unknown quest type
QUEST_TYPE_UNK -> QUEST_TYPE_TASK

**Target branch(es):** master


joschiwald:
@Krudor try to create a fresh new master branch and use this branch as base for your pr branches to stop looking your prs crazy

Krudor:
@joschiwald Thanks, I suck at git^^

Palabola:
It's a good feature, tricky part how to manage Areas where Bonus Objectives are active.

Shauren:
iirc the proper name for these types of quests is "Task"

Krudor:
Then commit and close this, let's just give it a name since we knows what it does :)
@Palabola I think all we need is a table mapping Task quest ids to areas, and add logic in the core for adding the quest when entering a specific area, and then hiding the quest when the player leaves (Haven't looked to much in sniffs how its done).
Not sure if it needs to check for conditions, like phases & other quest statuses to trigger.
In WoD they seem to just be there on the map, available whenever you enter the area.
Perhaps add a phasegroup id to the table, along with the area id as well.

Palabola:
@Krudor I guess using winding number algorithm based on Task POIs can reproduce the best funcionality.

Krudor:
There's no need to use the winding number algorithm.
You get the quests when you enter a specific area, simply tune into the area change event.

Shauren:
Nope, the area isnt a trigger

Krudor:
The "Task" quests I've look at in Draenor trigger when you enter a specific area. I've tried entering from different sides of the area where the Task quest is, and the 0.01 yard you enter the area, it triggers the Task quest.
How is this not triggered by area id?

Krudor:
https://www.youtube.com/watch?v=T2rUsx5mBSE

Palabola:
Areas are acceptiable way, but there is several Task which is active in bigger range than an Area.

Shauren:
Thats fine then - was just mentioning that even in WoD there are some area based triggers which are not areatriggers (clientside or serverside) or just areas - garrison entrance comes to mind here (huge custom shape)

PR #18220, opened 2016-11-08
Core/Movement: added safety checks

FrancescoBorzi:
**Changes proposed:**

-  added some safety checks in order to prevent crashes

- [x] 3.3.5
- [x] master

**Tests performed:** builds

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -58,12 +58,15 @@ void MotionMaster::InitDefault()
     if (_owner->GetTypeId() == TYPEID_UNIT)
     {
         MovementGenerator* movement = FactorySelector::selectMovementGenerator(_owner->ToCreature());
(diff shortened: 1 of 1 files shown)


src/server/game/Movement/MotionMaster.cpp
```diff
@@ -58,9 +58,15 @@ void MotionMaster::InitDefault()
...
+
+        if (movement)
+        {
+            Mutate(movement, MOTION_SLOT_IDLE);
+            return;
+        }
     }
-    else
+
+    if (&si_idleMovement)
```
Shauren: hahahahahaha. no.

ShinDarth: ok... can you elaborate? ^^

Shauren: You are taking the address of a variable, it will never be null.

Shauren:
Does not solve the crash it is attempting to fix http://paste.ubuntu.com/23448340/

PR #18260, opened 2016-11-16
Core/Scripts: Fixed Vaelastrasz bomb mechanic

HelloKitty:
**Changes proposed:**

-  Fix SPELL_CLEAVE targeted spell to be http://www.wowhead.com/spell=19983/cleave
-  Fix SPELL_BURNINGADRENALINE no cast fault.
-  Fix SPELL_BURNINGADRENALINE targeting of non-pet mana users in the raid. Cleaned up logic.

- [x] 3.3.5

Fixed broken Vael encounter mechanic

Tested it in-game. Bombs are applied only to mana users and tanks on the respective cooldown.

- [x] I'm not sure the bomb is dealing damage on death. I believe the comment said that the victim/target should cast it on themselves to avoid that issue, or something, but the way it was writtern prevented SPELL_BURNINGADRENALINE from ever being applied.

- [x] Currently the SPELL_BURNINGADRENALINE can be reapplied to a target that already has the debuff. This is not intended behavior.

**Commit Message**

Vaelstraz was suppose to bomb raid members that are mana users every 15
seconds. He was not doing that with the current implementation.

This implementation allows him to bomb the raid and to select the proper
targets. This is also done in a cleaner way.


HelloKitty:
Added an AuraScript to handle the AoE the debuff is suppose to cause at the end of its duration. Contrary to what the tooltip says it doesn't happen only on death. It happens whenever it falls off. This was discovered here: http://qaliaresponse.stage.lithium.com/t5/WoW-Mayhem/Surviving-Burning-Adrenaline-For-tanks/td-p/48609

HelloKitty:
This is working now. I've tested it and it is functional.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_vaelastrasz.cpp
```diff
@@ -188,24 +191,19 @@ class boss_vaelastrasz : public CreatureScript
                         break;
                     case EVENT_BURNINGADRENALINE_CASTER:
                         {
-                            Unit* target = NULL;
-
-                            uint8 i = 0;
-                            while (i < 3)   // max 3 tries to get a random target with power_mana
+                            //selects a random target that isn't the current victim and is a mana user (selects mana users) but not pets
+                            //it also ignores targets who have the aura. We don't want to place the debuff on the same target twice.
+                            if (Unit *target = SelectTarget(SELECT_TARGET_RANDOM, 1, [&](Unit* u) { return !u->IsPet() && u && u->getPowerType() == POWER_MANA && !u->HasAura(SPELL_BURNINGADRENALINE); }))
```
jackpoz: this code will trigger a static analysis issue about null checking "u" after it has been dereferenced

jackpoz (changes requested):

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_vaelastrasz.cpp
HelloKitty: I can add that one second.

sql/updates/world/3.3.5/2016_11_16_00_world.sql
```diff
@@ -0,0 +1,2 @@
+-- Adds a reference to the spell_vael_burning_adrenaline script for Burning Adrenaline (Vael Fight)
+INSERT INTO `world`.`spell_script_names` (`spell_id`, `ScriptName`) VALUES ('18173', 'spell_vael_burning_adrenaline');
```
HelloKitty: Ahhh, ok. Sorry, I'm new to TC contributions so I guess I should have looked more at the format of the SQL people were submitting.

HelloKitty: Commit 8b8fe3aa429394ccf93f94b577fd0ca6d810ab0d fixes it.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_vaelastrasz.cpp
HelloKitty: Added null check in Commit 192521b932f110341ac56ec5120affef55c6b1f7 but I do wonder if a null Unit will ever be provided to the PREDICATE. I've no experience in C++ but should I just be nullchecking any points TC spits out?

Also, GetTarget() says it won't ever be null. Does not null checking that cause static analysis issues?

sql/updates/world/3.3.5/2016_11_16_00_world.sql
```diff
@@ -1,2 +1,2 @@
 -- Adds a reference to the spell_vael_burning_adrenaline script for Burning Adrenaline (Vael Fight)
-INSERT INTO `world`.`spell_script_names` (`spell_id`, `ScriptName`) VALUES ('18173', 'spell_vael_burning_adrenaline');
\ No newline at end of file
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES ('18173', 'spell_vael_burning_adrenaline');
```
Keader: missing a new line

HelloKitty: Commit 94fb968d2be166f6e79304ba90eea2ff1743f698 fixes it.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_vaelastrasz.cpp
```diff
@@ -244,7 +242,39 @@ class boss_vaelastrasz : public CreatureScript
...
+    class spell_vael_burning_adrenaline_AuraScript : public AuraScript
+    {
+        PrepareAuraScript(spell_vael_burning_adrenaline_AuraScript);
+
+        void OnAuraRemoveHandler(AuraEffect const* aurEff, AuraEffectHandleModes /*mode*/)
+        {
+            //The tooltip says the on death the AoE occurs. According to information: http://qaliaresponse.stage.lithium.com/t5/WoW-Mayhem/Surviving-Burning-Adrenaline-For-tanks/td-p/48609
+            //Burning Adrenaline can be survived therefore Blizzard's implementation was an AoE bomb that went off if you were still alive and dealt
+            //damage to the target. You don't have to die for it to go off. It can go off whether you live or die.
+            GetTarget()->CastSpell(GetTarget(), SPELL_BURNINGADRENALINE_EXPLOSION, true, nullptr, aurEff);
```
HelloKitty: Only what? That should be correct for the behavior that is expected.

HelloKitty: I'm too new to TC to know exactly if that could be that. Give me some time to research that.

sql/updates/world/3.3.5/2016_11_16_00_world.sql
```diff
@@ -0,0 +1,2 @@
+-- Adds a reference to the spell_vael_burning_adrenaline script for Burning Adrenaline (Vael Fight)
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES ('18173', 'spell_vael_burning_adrenaline');
```
Keader: missing a delete line before insert (this querry should run multiple times)

something like:

``` sql
DELETE FROM `spell_script_names` WHERE `ScriptName`='spell_vael_burning_adrenaline';
```

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_vaelastrasz.cpp
HelloKitty: What exactly differs from the behavior of those two CastSpell overloads?

Keader: You can test in game, if ```aurEff``` change something in this case ^^

sql/updates/world/3.3.5/2016_11_16_00_world.sql
HelloKitty: This is for the purpose of making sure no DB updates cause conflicts yes?

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_vaelastrasz.cpp
HelloKitty: I will do that now, but it would be great if the behavior was documented somewhere. Do you have a reference to what the differences between all these overaloads are?

sql/updates/world/3.3.5/2016_11_16_00_world.sql
Keader: yes

sql/updates/world/3.3.5/2016_11_16_00_world.sql
```diff
@@ -1,2 +1,3 @@
 -- Adds a reference to the spell_vael_burning_adrenaline script for Burning Adrenaline (Vael Fight)
-INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES ('18173', 'spell_vael_burning_adrenaline');
+DELETE FROM `spell_script_names` WHERE `ScriptName`='spell_vael_burning_adrenaline';
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES ('18173', 'spell_vael_burning_adrenaline');
```
HelloKitty: Commit f30050c6d6ac8f4c89eb97c46c4afb7ca1ceda19 should fix it.

Keader:
You can use irc to get some help too: http://rizon.net/chat , #Trinity channel :)

HelloKitty:
Ahhh yes, I'll probably be stopping by in the future. I have a lot to learn. Started on Tuesday and it has been a wild ride so far lol.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_vaelastrasz.cpp
HelloKitty: Thank you =). I traced through what actually happens with the triggeredAura and it seems it won't have an effect in this particular case. So, the simpler overload that you suggested should work.

HelloKitty:
Travis failed for some reason. Had to close and reopen to kick it back off. There was no reason it should fail building.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_vaelastrasz.cpp
jackpoz: That's a good question and the answer can usually be found in the implementation of the function itself. You could check if SelectTarget( ) ever passes a NULL Unit* .
Imo it makes little sense to pass NULL there (but that's out of scope of your PR).

We periodically run static analysis and come back to original authors notifying them about actual issues (or modify the code just to shut the warning, even if it's not a real issue, with either a NULL check or an ASSERT).

HelloKitty:
When I get back to my computer I'll take a closer look at why Travis has been failing. I could have broken something and not recompiled to verify.

jackpoz (approved):

HelloKitty:
@Aokromes ahhh, let me re-cmake the solution with warnings enabled. That occured when I switched the SpellCast overload I was using. Can I get your input and opinion on which overload would be proper in this case?

@Keader said that the the triggered aura overload was uneeded. After looking through the source I saw that he was probably right in this case. Is it best practice to provide the aura that triggered this cast though? I'm very much interested in if there is official written documentation about the differences and behavior of these overloads. I'd prefer not to dig through and read the source to understand the API. Is there written documentation on this stuff?

jackpoz:
to fix "unused parameter" issues you can simply comment out the parameter in your function, i.e.:
```
void OnAuraRemoveHandler(AuraEffect const* /*aurEff*/, AuraEffectHan...
```

HelloKitty:
@jackpoz I see, it's a very odd thing coming from C#. Also, no matter what analysis or warning ruleset I enable the warning never comes up for me in VS2015. Very odd.

Rochet2:
@HelloKitty You can get it by using /Wall or /W4
http://i.imgur.com/XDLxzVh.png
/W3 seems to be default no matter if with_warnings is selected in cmake

HelloKitty:
@Rochet2 Ahh thank you, I see that as a cumbersome setup though. I spent abit of time and found that adding 4100 warning as an error settings off the unreferenced parameter warning even with warnings, or lower than w4 level, disabled. I think I'll take that approach.

I apologize again to everyone here. I've essentially never written any C++ outside of simple labs for homework. I'm mainly a C#/.Net developer so there is a lot of arcane stuff here. I'm sorry that such a small change resulted in 30 comments. Next PR I'll make sure that these same issues don't come up again.

Also, as far as I'm concerned I believe this is complete. Thank you everyone for pitching in.

Keader:
dont worry about comments numbers.
We are here to help you improve your code.

PR #18290, opened 2016-11-21
[3.3.5/master] Core/Creature: Mount Inhabit Type

Krudor:
**Changes proposed:**

-  Add another Inhabit Type flag for creature mounts

**Description:**
Eliminates the issues to WHERE a creature belongs once it mounts / dismounts by giving the creature a separate Inhabit Type flag for its mount.
Most noticeable change is that ground creatures on flying mounts will now properly fall if dismounted (No more ground units swimming in the air when they shouldn't!)
This is a suggestion how you can fix it, maybe the more experienced developers have a better way of doing it, but it made the most sense to me and was made possible with very little code changes.
This also won't magically fix every creature with mount InhabitType issues, they will need creature_template and creature_template_addon updates per case.

**Tests performed:** Builds and tested in-game

- ~~Ground creatures on flying mounts dismounting will stutter when falling to the ground, related?~~
- No other issues observed

[YouTube - Before example](https://www.youtube.com/watch?v=4ltSzOBB72k)
[YouTube - After example](https://www.youtube.com/watch?v=ws6Wvmi-YSE)


sql/updates/world/master/2016_11_xx_00_world.sql
```diff
@@ -1,2 +1,2 @@
-ALTER TABLE `creature_addon` ADD COLUMN `mountInhabitType` TINYINT(3) UNSIGNED NOT NULL DEFAULT '3' AFTER `mount`;
-ALTER TABLE `creature_template_addon` ADD COLUMN `mountInhabitType` TINYINT(3) UNSIGNED NOT NULL DEFAULT '3' AFTER `mount`;
\ No newline at end of file
+ALTER TABLE `creature_addon` ADD COLUMN `mountInhabitType` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER `mount`;
+ALTER TABLE `creature_template_addon` ADD COLUMN `mountInhabitType` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER `mount`;
```
Keader: missing new line

src/server/game/Entities/Creature/Creature.h
```diff
@@ -787,6 +788,8 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         CreatureData const* GetCreatureData() const { return m_creatureData; }
         CreatureAddon const* GetCreatureAddon() const;
 
+        uint8 GetInhabitType() const { return IsMounted() ? GetCreatureAddon()->mountInhabitType : GetCreatureTemplate()->InhabitType; }
```
Shauren: Crash: GetCreatureAddon() can return nullptr

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9182,6 +9182,8 @@ void Unit::Mount(uint32 mount, uint32 VehicleId, uint32 creatureEntry)
     }
 
     RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_MOUNT);
+    if (GetTypeId() == TYPEID_UNIT)
```
Shauren:
```
if (Creature* creature = ToCreature())
    creature->UpdateMovementFlags();
```

ToCreature is designed to return nullptr on invalid cast

Shauren (changes requested):

src/server/game/Entities/Unit/Unit.cpp
Krudor: Yeah, there's alot of checks like this though, and
```
if (creature->IsAIEnabled)
    creature->AI()->...
```
instead of
```
if (CreatureAI* AI = creature->AI())
    AI->...
```
that makes me unsure of what approach to use :P

Shauren: Don't use AI as an example, IsAIEnabled is a separate boolean to disable AI (yes, its a valid state) - also its never null for creatures

src/server/game/Entities/Creature/Creature.h
```diff
@@ -722,9 +723,9 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         bool IsCivilian() const { return (GetCreatureTemplate()->flags_extra & CREATURE_FLAG_EXTRA_CIVILIAN) != 0; }
         bool IsTrigger() const { return (GetCreatureTemplate()->flags_extra & CREATURE_FLAG_EXTRA_TRIGGER) != 0; }
         bool IsGuard() const { return (GetCreatureTemplate()->flags_extra & CREATURE_FLAG_EXTRA_GUARD) != 0; }
-        bool CanWalk() const { return (GetCreatureTemplate()->InhabitType & INHABIT_GROUND) != 0; }
-        bool CanSwim() const { return (GetCreatureTemplate()->InhabitType & INHABIT_WATER) != 0 || IsPet(); }
-        bool CanFly()  const override { return (GetCreatureTemplate()->InhabitType & INHABIT_AIR) != 0; }
+        bool CanWalk() const { return (GetInhabitType() & INHABIT_GROUND) != 0; }
+        bool CanSwim() const { return (GetInhabitType() & INHABIT_WATER) != 0; }
```
Shauren: What about the removed IsPet() check?

Krudor: Good question, that was unintentional :/

Krudor:
```
Creature::UpdateMovementFlags()
{
    ...
    SetSwim((GetInhabitType() & INHABIT_WATER) && IsInWater());
}
```
Shouldn't this be a call to CanSwim()? It's missing a IsPet() check as well.
Looks like there's alot of GetInhabitType checks that could be replaced with CanWalk/Swim/Fly.

Aokromes:
we really need 2 new columns for few creatures? it looks to killer for me more when it's a custom field. we need proper inhabitattype handling

Krudor:
I don't see a reason why this field shouldn't exist.
Not choosing to implement it because the tables become bloated is not a good enough reason in my opinion.
How would you otherwise handle this behavior?

Aokromes:
^^
we need proper inhabitattype handling

Kittnz:
We should be able to change the inhabit type ingame. For example: https://github.com/TrinityCore/TrinityCore/issues/12581 . These NPCs should change inhabittype when they are on the ground (after the waypoint). Or we should have a way to force change the inhabit type. Currently this is not posible because the data from database overwrites it. Even if you try to change it.

Krudor:
Bad example perhaps, this PR is a perfect example that actually fixes the problem you described Kittnz, but I see your point.

Kittnz:
Well yes, it does. But we need another way then this.

ccrs:
but we need both eitherway, without this all ground units that fly with mount would need a specific script to set the flying moveflag, and thats quite an amount of cases.

Shauren (approved):

malcrom:
The creatures don't change inhabittype. They drop the hoverflag from bytes_1.

Aokromes:
2411093

PR #18292, opened 2016-11-21
Scripts/Spells: Fixed an issue between Fire Nova and LoS

### LINKED ISSUE #18268: [Shaman] Fire Nova and LoS

**Description:**
Shaman's spell Fire Nova (all ranks) trigger Fire Nova from Fire Totem(s)

**Current behaviour:** (Tell us what happens.)
Currently, when player is in LoS with his totem and cast Fire Nova his spell will go on cooldown but there is no effect from totem.

**Expected behaviour:** (Tell us what should happen instead.)
When casted, Fire Nova should ignore LoS between player and totem and trigger its effect.

1. Create shaman
2. Spawn any fire totem and go in LoS with it
3. Cast Fire Nova

**Branch(es):** 3.3.5 / master (Tell us which branch(es) this issue affects.)
3.3.5 and master

**TC rev. hash/commit:**  503259e6fe0a1e6328090a5a2e88832b6a017516

**TDB version:**  335.62

**Operating system:**  Windows 7 x64

Goatform:

**Changes proposed:**

-  If player is out of LoS then totem will trigger spell by itself

- [x] 3.3.5

**Issues addressed:** Closes #18268

**Tests performed:** Builds and tested in-game

**Known issues and TODO list:** Nothing so far.

### DIFF
--- a/src/server/scripts/Spells/spell_shaman.cpp
+++ b/src/server/scripts/Spells/spell_shaman.cpp
@@ -609,6 +609,11 @@ class spell_sha_fire_nova : public SpellScriptLoader
                     return SPELL_FAILED_CUSTOM_ERROR;
                 }
 
+                float distance = 30.0f;
+                Creature* totem = GetCaster()->GetMap()->GetCreature(GetCaster()->m_SummonSlot[1]);
+                if (GetCaster()->GetExactDist2d(totem->GetPositionX(), totem->GetPositionY()) > distance)
+                    return SPELL_FAILED_OUT_OF_RANGE;
+
                 return SPELL_CAST_OK;
             }
 
--- a/src/server/scripts/Spells/spell_shaman.cpp
+++ b/src/server/scripts/Spells/spell_shaman.cpp
@@ -619,7 +619,7 @@ class spell_sha_fire_nova : public SpellScriptLoader
                 {
                     uint8 rank = GetSpellInfo()->GetRank();
                     if (totem->IsTotem())
-                        totem->CastSpell(totem, sSpellMgr->GetSpellWithRank(SPELL_SHAMAN_FIRE_NOVA_TRIGGERED_R1, rank), true);
+                        totem->CastSpell(totem, sSpellMgr->GetSpellWithRank(SPELL_SHAMAN_FIRE_NOVA_TRIGGERED_R1, rank), true, nullptr, nullptr, caster->GetGUID());
                 }
             }
 
--- a/src/server/scripts/Spells/spell_shaman.cpp
+++ b/src/server/scripts/Spells/spell_shaman.cpp
@@ -619,7 +619,7 @@ class spell_sha_fire_nova : public SpellScriptLoader
                 {
                     uint8 rank = GetSpellInfo()->GetRank();
                     if (totem->IsTotem())
-                        caster->CastSpell(totem, sSpellMgr->GetSpellWithRank(SPELL_SHAMAN_FIRE_NOVA_TRIGGERED_R1, rank), true);
+                        totem->CastSpell(totem, sSpellMgr->GetSpellWithRank(SPELL_SHAMAN_FIRE_NOVA_TRIGGERED_R1, rank), true);
                 }
             }
 


Shauren:
What if i told you this might not be blizzlike for the totem to cast it?

Goatform:
I don't agree with you either at all but that is my opinion. Blizzlike or not, there is no need to make something more complicated instead of solving it in easier and most likely better way.

ccrs:
add the player as original caster and u get the same packet :D

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -619,7 +619,7 @@ class spell_sha_fire_nova : public SpellScriptLoader
                 {
                     uint8 rank = GetSpellInfo()->GetRank();
                     if (totem->IsTotem())
-                        caster->CastSpell(totem, sSpellMgr->GetSpellWithRank(SPELL_SHAMAN_FIRE_NOVA_TRIGGERED_R1, rank), true);
+                        totem->CastSpell(totem, sSpellMgr->GetSpellWithRank(SPELL_SHAMAN_FIRE_NOVA_TRIGGERED_R1, rank), true);
```
matukaa: as @ccrs suggested, use
`totem->CastSpell(totem, sSpellMgr->GetSpellWithRank(SPELL_SHAMAN_FIRE_NOVA_TRIGGERED_R1, rank), true, nullptr, nullptr, caster->GetGUID());`

Goatform: Okay, I will test that a bit later to make sure that it works correctly.

Goatform:
Tested ingame:
eb61c939a0c64cbd8577045b9545feb6

Shauren:
And another question, can you show us some proof that LoS should be ignored? Maybe the entire cast should fail completely with an error message instead of just going on cooldown with no effect?

Mrgrasser:
> And another question, can you show us some proof that LoS should be ignored?

For example in the map battle arena Dalaran_Sewers , when water loss in the middle maybe

Keader:
> For example in the map battle arena Dalaran_Sewers , when water loss in the middle maybe

it's a example not a proof...
any video or something that support your example?

offl:
```cpp
if (totem->IsTotem())
```

That's a funny check.

Rushor:
moreover with this fix, the range check of (30? yards or sth. like that) will no longer be given. You will be able to cast the spell 100 yards away from the totem (so a rangecheck must be added then).

Goatform:
@Rushor, test it now mate :+1:

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -609,6 +609,11 @@ class spell_sha_fire_nova : public SpellScriptLoader
                     return SPELL_FAILED_CUSTOM_ERROR;
                 }
 
+                float distance = 30.0f;
+                Creature* totem = GetCaster()->GetMap()->GetCreature(GetCaster()->m_SummonSlot[1]);
+                if (GetCaster()->GetExactDist2d(totem->GetPositionX(), totem->GetPositionY()) > distance)
```
Keader: should be >= no?
whats happens when is exactly 30.0 yards? :)

Rushor:
yep that works fine

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -609,6 +609,11 @@ class spell_sha_fire_nova : public SpellScriptLoader
                     return SPELL_FAILED_CUSTOM_ERROR;
                 }
 
+                float distance = 30.0f;
+                Creature* totem = GetCaster()->GetMap()->GetCreature(GetCaster()->m_SummonSlot[1]);
```
ccrs: ugh, really ugly

get the caster -> get the map -> look for a creature -> get the summon for the magic number summon slot 1

go for: get the summon slot and null check it if necessary and cast it to a creature if necessary too

ariel-: You can use ObjectAccessor

matukaa: Use object accessor, as @ariel- suggested, but you should always opt for non-hardcoded values. Like check if the original spell has rangeindex set, evaluate that, use that value. Same summonslot. Get the spelleffect that summons the totem, miscvalueb gives the summonproperties entry, look it up in the store, use the "slot" field to get totem from m_SummonSlot.
These are all optional, since 3.3.5 is not going to change anytime soon, but it's a good habit to always use dbc values, so if blizzard changes them, you won't have to bother with changing these scripts

jackpoz:
Please note there are some comments that need to be addressed before this PR can be merged.
You can update this one and ask here to reopen it when ready. 
Every contributor is still able to comment on it even if closed.

Issue #18214, opened 2016-11-08
[3.3.5a][master] cherry-pick BoundingIntervalHierarchy from CMaNGOS


funjoker:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:** typo in Vmap BoundingIntervalHierarchy generation
Due to this typo, some objects could be left out during BIH generation and thus never checked for intersection. This breaks height and LoS calculations in some areas.

https://github.com/cmangos/mangos-classic/pull/215

File name & path in TC: https://github.com/TrinityCore/TrinityCore/blob/master/src/common/Collision/BoundingIntervalHierarchy.cpp

**Current behaviour:** Height and LoS calculation breaks in some areas.

**Expected behaviour:** LoS and Height calculation should not break, NPCs should behave normally.

**Steps to reproduce the problem:**
The occurrence of this bug can be seen in "Md_Mushroomcave.wmo". It can be found in several locations, for example (-6140, -2960, 400), map 0. In some parts of the cave mobs will either fall through the ground or fly up into the sky when chasing the player. LoS is not obstructed there as well.

**Branch(es):** 3.3.5 **and** master

Killyana:
Fixes: https://github.com/TrinityCore/TrinityCore/issues/9753

jackpoz:
As usual refactor changes should be left out from commits that affect code functionality so only https://github.com/cmangos/mangos-classic/pull/215/commits/f68416c9a13da4d116ce38ed1d6b3173f4021976 should be tested and merged

Aokromes:
@jackpoz this needs vmaps re-extraction? if yes then maybe we need to increase vmaps version no?

Kittnz:
Any testing been done with this commit?

Issue #9753, opened 2013-05-04
Mobs air-stepping into cave ceilings


ghost:
My character is only in Westfall, but so far all of the caves he's been in so far, the mobs while fleeing will start to step up into the air untill they have vanished into the roof of the cave. I have tried with MMaps on and off and still get the same results. I have tried re-compiling the map extractors in Tools and re extract the maps then deleted all folders in my bin/release folder and copied the new folders into the bin/release folder and rerun mmaps_generator, and still not fixed. Any ideas out there on how to fix this?

TrinityCore rev. b1e8c422593f 2013-05-01 21:56:39 +0200 (master branch) (Win64, Release)
TDB 335.51 `2013_05_04_01_world_creature_template.sql`
##

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1413005-mobs-air-stepping-into-cave-ceilings?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).

Killyana:
This issue can be observed here: Map: 571 (Northrend) Zone: 210 (Icecrown) Area: 4595 (Hall of Horrors)  X: 8002.013184 Y: 3312.416748 Z: 632.395935 Orientation: 4.051054
Spawn any npc and aggro him, he will go to the roof.

Killyana:
This issue seems happening only if the nps is in the water.

Killyana:
https://github.com/TrinityCore/TrinityCore/commit/0df0cb30ca9de2015ec774e50bb69337382576d3

PR #18297, opened 2016-11-22
Fix typo in VMap BIH generation

Aokromes:
Description: typo in Vmap BoundingIntervalHierarchy generation
Due to this typo, some objects could be left out during BIH generation and thus never checked for intersection. This breaks height and LoS calculations in some areas.

cmangos/mangos-classic#215

File name & path in TC: https://github.com/TrinityCore/TrinityCore/blob/master/src/common/Collision/BoundingIntervalHierarchy.cpp

Current behaviour: Height and LoS calculation breaks in some areas.

Expected behaviour: LoS and Height calculation should not break, NPCs should behave normally.

The occurrence of this bug can be seen in "Md_Mushroomcave.wmo". It can be found in several locations, for example (-6140, -2960, 400), map 0. In some parts of the cave mobs will either fall through the ground or fly up into the sky when chasing the player. LoS is not obstructed there as well.

Branch(es): 3.3.5 and master

### DIFF
--- a/src/common/Collision/BoundingIntervalHierarchy.cpp
+++ b/src/common/Collision/BoundingIntervalHierarchy.cpp
@@ -152,13 +152,13 @@ void BIH::subdivide(int left, int right, std::vector<uint32> &tempTree, buildDat
         else if (left > right)
         {
             // all right
+            right = rightOrig;
             if (prevAxis == axis && G3D::fuzzyEq(prevSplit, split)) {
                 // we are stuck here - create a leaf
                 stats.updateLeaf(depth, rightOrig - left + 1);
                 createNode(tempTree, nodeIndex, left, rightOrig);
                 return;
             }
-            right = rightOrig;
             if (clipR >= split) {
                 // keep looping on right half
                 gridBox.lo[axis] = split;
--- a/src/common/Collision/BoundingIntervalHierarchy.cpp
+++ b/src/common/Collision/BoundingIntervalHierarchy.cpp
@@ -155,8 +155,8 @@ void BIH::subdivide(int left, int right, std::vector<uint32> &tempTree, buildDat
             right = rightOrig;
             if (prevAxis == axis && G3D::fuzzyEq(prevSplit, split)) {
                 // we are stuck here - create a leaf
-                stats.updateLeaf(depth, rightOrig - left + 1);
-                createNode(tempTree, nodeIndex, left, rightOrig);
+                stats.updateLeaf(depth, right - left + 1);
+                createNode(tempTree, nodeIndex, left, right);
                 return;
             }
             if (clipR >= split) {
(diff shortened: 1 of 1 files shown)


Aokromes:
No feedback on this? it needs to update vmaps? if no i will merge this on monday.

lorac2k14:
I have been testing this and vmap output since last commit and so far it looks as though this has corrected a few places ... I can not say it corrects all because Its a big world and there is a lot of places to check but the above mentioned places are fixed. 
Imho I think it is safe 
Other note.. it is nice to see npcs not flying after me or falling through the ground :-) nice work

zorix:
In my option it should be merged.

It's a typo from https://github.com/cmangos/mangos-classic/commit/7fe0ee46eacbe80a9a83a86ef2da8bc68a7a4c6c#diff-ed0561f398a3b842031d0a47932bbd85R146

Reason why this typo happened:
First check Sunflow repository on which vmaps are based.
https://github.com/fpsunflower/sunflow/blob/master/src/org/sunflow/core/accel/BoundingIntervalHierarchy.java#L276

As you can see here that `if (clipR >= split)` and `if (prevAxis == axis && prevSplit == split)` are swapped. Same thing happened for `if (clipL <= split)` and `if (prevAxis == axis && prevSplit == split)` (one if above)

Why it was swapped?
Because when you make it the same like in Sunflow vmaps will stuck in endless loop when generating (tested by me).

This typo occurred when those ifs was being swapped, the first `if` was taken with `right = rightOrig;` what is wrong because `right` should be updated before doing anything (if you check logic in Sunflow code).

Conclusion for me it's good change and if you want make it as it should be done when it was porting, the change should looks like this
```
@@ -152,13 +152,13 @@ void BIH::subdivide(int left, int right, std::vector<uint32> &tempTree, buildDat
         else if (left > right)
         {
             // all right
+            right = rightOrig;
             if (prevAxis == axis && G3D::fuzzyEq(prevSplit, split)) {
                 // we are stuck here - create a leaf
                 stats.updateLeaf(depth, right - left + 1);
                 createNode(tempTree, nodeIndex, left, right);
                 return;
             }
-            right = rightOrig;
             if (clipR >= split) {
                 // keep looping on right half
                 gridBox.lo[axis] = split;
```

Aokromes:
@zorix  so like this?

zorix:
Yes.

jackpoz:
what about https://github.com/cmangos/mangos-classic/pull/215#issuecomment-263087820 comment ?

zorix:
They didn't said that it's new bug (or I misunderstood) after this change and probably it's another bug in VMaps.

Rushor:
related to https://github.com/TrinityCore/TrinityCore/issues/9753 ?

zorix:
Related to VMaps but probably not to this change.

funjoker:
#18214 ?

Chrissy2520:
Hey, well I am having issuses aswell as u had. I did everything correctly, but when it comes 2 launching the worldserver.exe it close down immediately and this pop up:
WORLD: MMap data directory is: ./mmaps
VMap support included. LineOfSight: 1, getHeight: 1, indoorCheck: 1
VMap data directory is: ./vmaps
VMap file './vmaps/000.vmtree' does not exist
Please place VMAP files (*.vmtree and *.vmtile) in the vmap directory (./vmaps/), or correct the DataDir setting in your worldserver.conf file.
Unable to load critical files - server shutting down !!!

I couldnt find any 2017 threads about this, so I consider making a new one

PR #18298, opened 2016-11-22
Creature/Movement: Enable swimming creatures to chase targets in 3d space

### LINKED ISSUE #11060: Creatures don't chase/follow pure Z axis moves while swimming
For ba1a4cd0ce51a0a6389317199c66fd85deffac02

Steps to reproduce: attack a creature in a body of deep water, and make sure you're right in the same place as them. Then swim directly down or up. When you move out of their combat range, they'll stop being able to attack, and will not follow you. (This also happens with hunter pets; or at least the ones for which swimming isn't broken.)

Expected behavior: They follow you to continue the attack.

The problem seems to be in TargetedMovementGenerator.cpp, in DoUpdate. Creatures which can't fly won't register that the target (i.e., the player attacking them) has moved away if the movement is entirely up/down, because non-flying creatures use IsWithinDist2d to check that.

I don't know if this should just be switched to always use IsWithinDist3d, or if a check should be added to see if the creature which is moving is in the water.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1412792-creatures-don-t-chase-follow-pure-z-axis-moves-while-swimming?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Krudor:

**Changes proposed:**

-  Enable swimming creatures to chase targets in 3d space

**Issues addressed:** Closes #11060

**Tests performed:** Builds, works great in game

**Description:** Swimming creatures are currently unable to detect if targets are in range or not if they change their Z position, as they only calculate distance in 2d space.
This PR enables swimming creatures to calculate distance to target in 3d space, just like flying creatures does.

### DIFF
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -166,7 +166,7 @@ bool TargetedMovementGeneratorMedium<T, D>::DoUpdate(T* owner, uint32 time_diff)
                 transport->CalculatePassengerPosition(dest.x, dest.y, dest.z);
 
         // First check distance
-        if (owner->GetTypeId() == TYPEID_UNIT && owner->ToCreature()->CanFly())
+        if (owner->GetTypeId() == TYPEID_UNIT && (owner->ToCreature()->CanFly() || owner->ToCreature()->CanSwim()))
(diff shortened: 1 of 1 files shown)


Subv:
What happens if the creature can swim but is not in the water?

PR #18320, opened 2016-11-26
Add support for QUEST_OBJECTIVE_CRITERIA_TREE

rstebler:

**Changes proposed:**

-  Add support for QUEST_OBJECTIVE_CRITERIA_TREE

- [ ] 3.3.5
- [x] master

- [x] Does it build?
- [x] Tested in-game? Tested with quest 38312

### DIFF
--- a/src/server/game/Achievements/AchievementMgr.cpp
+++ b/src/server/game/Achievements/AchievementMgr.cpp
@@ -65,7 +65,7 @@ bool AchievementMgr::CanUpdateCriteriaTree(Criteria const* criteria, CriteriaTre
 {
     AchievementEntry const* achievement = tree->Achievement;
     if (!achievement)
-        return false;
+        return true;
 
     if (HasAchieved(achievement->ID))
     {
@@ -96,7 +96,7 @@ bool AchievementMgr::CanCompleteCriteriaTree(CriteriaTree const* tree)
 {
     AchievementEntry const* achievement = tree->Achievement;
     if (!achievement)
-        return false;
+        return true;
 
     // counter can never complete
     if (achievement->Flags & ACHIEVEMENT_FLAG_COUNTER)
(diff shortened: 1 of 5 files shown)


src/server/game/Achievements/AchievementMgr.cpp
```diff
@@ -65,7 +65,7 @@ bool AchievementMgr::CanUpdateCriteriaTree(Criteria const* criteria, CriteriaTre
 {
     AchievementEntry const* achievement = tree->Achievement;
     if (!achievement)
-        return false;
+        return true;
```
Krudor: What?
If you can't find an achievement linked to the criteriatree, return true no matter what? This is really really bad.

rstebler: You are right. Sorry I'm new to TC and still learning how everything works :)
What about the CriteriaHandler check in the new commit?

Krudor: If it's not an achievement then it's not supposed to update it, this is the achievement manager, it wasn't designed to handle quest's criteria trees.
If there's another system that starts using criterias then I guess we need a NEW manager for it.

Krudor: Take a look at Player::UpdateCriteria, it sends updates to the achievement manager, scenario, and guild system, which are all separate criteria handlers.

rstebler: Yea a new manager for that would be the best thing. If I have time I will have a look at it but I think it's best to leave this work to you and other more experienced people than me :)
Thanks for your time.

Aokromes:
If no one have opposition i will merge this tomorrow.

src/server/game/Achievements/AchievementMgr.cpp
jackpoz: This point still needs to be addressed. Post here to reopen the PR once done

rstebler: This correction needs an additional check? 6688ad03d35bb264e41f5fbfcc5cee377c84d913

jackpoz: @Krudor can you check if the changes are ok ?

Aokromes:
btw @rstebler i sugest you to jump irc irc.rizon.net #trinity

PR #18328, opened 2016-11-27
[3.3.5] Core/AuctionHouse: Auction Bidder Table

iridinite:
**Changes proposed:**
- Save a list of all characters who have placed a bid on an auction, to the characters database. When client requests list of auctions (s)he is involved in, use this table, instead of only returning auctions where that character is the current highest bidder.
- This prevents auctions that you placed a bid on earlier from disappearing from the Bids tab if you are outbid and you relog or the server is restarted.

**Target branch(es):**
- [x] 3.3.5
(it might also apply to master, I have not tested this)

**Todo:**
- Could someone please help me understand under what circumstances the client [requests certain auction IDs](https://github.com/iridinite/TrinityCore/blob/4ac9ae29447316d66b7a19671606305ba63a969f/src/server/game/Handlers/AuctionHouseHandler.cpp#L652) in `WorldSession::HandleAuctionListBidderItems` and if it matters if we just ignore that entirely? I mean, using this bid table system, it's completely irrelevant if the client happens to remember a handful of auction IDs.

**Tests performed:**
Built and tested in game.


Shauren:
is this how it works on retail?

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -892,6 +896,21 @@ bool AuctionEntry::LoadFromDB(Field* fields)
         TC_LOG_ERROR("misc", "Auction %u has not a existing item : %u", Id, itemGUIDLow);
         return false;
     }
+
+    // get list of bidders
+    PreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_SEL_AUCTION_BIDDERS);
+    stmt->setUInt32(0, Id);
+
+    if (PreparedQueryResult result = CharacterDatabase.Query(stmt))
```
Shauren: this is going to be terribly slow if you launch a separate query for every single auction during startup

iridinite:
Confirmed with a friend who has access to retail.

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
iridinite: Would it be a better idea to load all of the table at once and keep it around in memory until all auctions are loaded?

Shauren: All auctions are loaded on startup with one query in `AuctionHouseMgr::LoadAuctions`, just run another query (for the entire table) there and parse it before the loop calling `AuctionEntry::LoadFromDB`

src/server/game/AuctionHouse/AuctionHouseMgr.h
```diff
@@ -85,6 +85,7 @@ struct TC_GAME_API AuctionEntry
     ObjectGuid::LowType bidder;
     uint32 deposit;                                         //deposit can be calculated only when creating auction
     uint32 etime;
+    std::unordered_set<uint32> bidderlist;
```
Shauren: if this is meant to be a guid, use ObjectGuid::LowType, never the underlying type (master branch portability issue, its a different type there)

iridinite:
It seems like the Travis build breaks on the SQL patch. I can't quite make sense of the error, what's going on there?

Aokromes:
It's not your issue, the guy who merges the patch must rename the file and add it to updates.

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -349,21 +349,44 @@ void AuctionHouseMgr::LoadAuctions()
...
+    std::unordered_map<uint32, std::unordered_set<ObjectGuid::LowType>> bidderlists;
+    PreparedStatement* stmt2 = CharacterDatabase.GetPreparedStatement(CHAR_SEL_AUCTION_BIDDERS);
+    
+    uint32 countBidders = 0;
+    if (PreparedQueryResult resultBidders = CharacterDatabase.Query(stmt2))
+    {
+        do
+        {
+            Field* fields = resultBidders->Fetch();
+            auto it = bidderlists.find(fields[0].GetUInt32());
```
Shauren: not needed, operator[] creates entries in map if not present.
just bidderlists[fields[0].GetUInt32()].insert(fields[1].GetUInt32());

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -373,13 +396,17 @@ void AuctionHouseMgr::LoadAuctions()
             continue;
         }
 
+        auto it = bidderlists.find(aItem->Id);
+        if (it != bidderlists.end())
+            aItem->bidderlist = std::unordered_set<ObjectGuid::LowType>(it->second);
```
Shauren: for extra performance&fun, std::move it->second

Shauren (changes requested):

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -349,21 +349,46 @@ void AuctionHouseMgr::LoadAuctions()
...
+    
+    uint32 countBidders = 0;
+    if (PreparedQueryResult resultBidders = CharacterDatabase.Query(stmt2))
+    {
+        do
+        {
+            Field* fields = resultBidders->Fetch();
+            uint32 auctionId = fields[0].GetUInt32();
 
+            auto it = bidderlists.find(auctionId);
```
Shauren: not needed, operator[] creates entries in map if not present.
just bidderlists[fields[0].GetUInt32()].insert(fields[1].GetUInt32());

iridinite:
@Shauren Done, thanks for your comments!

iridinite:
Any comments?

Faq:
@joschiwald could you please check?

src/server/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -845,6 +865,10 @@ void AuctionEntry::DeleteFromDB(SQLTransaction& trans) const
     PreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_DEL_AUCTION);
     stmt->setUInt32(0, Id);
     trans->Append(stmt);
+
+    PreparedStatement* stmt2 = CharacterDatabase.GetPreparedStatement(CHAR_DEL_AUCTION_BIDDERS);
```
joschiwald: you can reuse `stmt`

tkrokli:
The branch [`iridinite:3.3.5-bidders`](https://github.com/iridinite/TrinityCore/tree/3.3.5-bidders) needs to be updated/rebased or at least `.travis.yml` needs to be updated to latest version (https://github.com/TrinityCore/TrinityCore/commit/bf187a195ee8bd80cb1c7e7a03eacca20125223e) in order for travis-ci to work on this PR.

src/server/game/Handlers/AuctionHouseHandler.cpp
```diff
@@ -497,6 +498,11 @@ void WorldSession::HandleAuctionPlaceBid(WorldPacket& recvData)
         stmt->setUInt32(2, auction->Id);
         trans->Append(stmt);
 
+        PreparedStatement* stmt2 = CharacterDatabase.GetPreparedStatement(CHAR_INS_AUCTION_BIDDERS);
+        stmt2->setUInt32(0, auction->Id);
+        stmt2->setUInt32(1, auction->bidder);
+        trans->Append(stmt2);
```
joschiwald: you only need to fire this query if bidder was not present in bidderlist before

sql/base/characters_database.sql
```diff
@@ -214,6 +214,17 @@ LOCK TABLES `auctionhouse` WRITE;
 UNLOCK TABLES;
 
 --
+-- Table structure for table `auctionbidders`
+--
+
+DROP TABLE IF EXISTS `auctionbidders`;
+CREATE TABLE `auctionbidders` (
+	`id` INT(10) UNSIGNED NOT NULL DEFAULT '0',
+	`bidder` INT(10) UNSIGNED NOT NULL DEFAULT '0',
```
joschiwald: we don't like tabs and better name will be `BidderGuid`

iridinite:
Well I can't say I was expecting this PR to be dug up again. Thanks for the comments, I'll reinstall the tools and see what I can do.

Is there an easy way to rebase this PR on HEAD of 3.3.5 without screwing up the version history? Am a bit of a git noob

tkrokli:
To update your branch `3.3.5-bidders` with the 3.3.5 TC branch, make sure you choose 'rebase' to base your changes on top of the merge.

I only use Git Extensions, so I use the GUI approach. There is of course a git command for doing this, but I would prefer if someone doing it regularly could advice you on what it looks like in your case, if you use git on the command line.

tkrokli:
BTW, you may want to include an update in the `sql/base/characters_database.sql` file if you want to see a full green result from travis-ci.

Example, insert the following at the end of **line 2571** of your `sql/base/characters_database.sql` file:
```sql
,('9999_99_99_18328_characters.sql','790DB3C1D6E0E50349E951A573CC8D2DC271B5F0','RELEASED','2017-11-12 16:56:23',0)
```
before the ending semicolon (` ; `).

That content will have to be changed again when merging due to different file name and date/time info.

r00ty-tc:
The way I do it is, something like this:

Do this once for the local repository. If you clone a new one you must do it again.
```
git remote add TrinityCore https://github.com/TrinityCore/TrinityCore
```

Then every time you want to update

```
git fetch TrinityCore
git rebase TrinityCore/3.3.5
```

Of course, if it cannot auto merge your changes on top of the current latest version it will list conflicts, which you need to fix. Once you're sure all are fixed (after fixing each file you need to add with git add) you can then enter (only if there were conflicts)

```
git rebase --continue
```

If you did a good job of fixing conflicts it will complete and you'll have a repo in the state that your commits appear after the latest commits from the TrinityCore 3.3.5 branch.

Now you should ensure that it compiles and works as before. If so, then you must force push to your own github repo with git push -f. The reason to force it is because you're rewriting the history of your repo by rebasing.

Now, the way to do it without rewriting history is git merge. But, for the purposes of updating a PR I think it's better to rebase.

This will cause a spam of text in our IRC channel for each commit.. :)

iridinite:
@r00ty-tc "This will cause a spam of text in our IRC channel for each commit.. :)"
I assume you don't mind that, then, because I was just about to start rebasing :P

r00ty-tc:
Well I personally think the bot should mute/summarize these multiple commits that happen in a short period. But, that's not your problem :)

r00ty-tc:
Actually looks like I was wrong. Just a single line sent. I think it's because I was working on a branch on TrinityCore itself and that logs each commit. So, all good.

iridinite:
Neat-o. I'll look into the requested changes!

iridinite:
@tkrokli What hash algorithm is used for the characters.updates table? Doesn't seem like throwing SHA1 at the file gets me the same value as in your comment.

Aokromes:
md5

r00ty-tc:
It is sha1sum, but if you're on windows it will give the wrong result. Try using dos2unix (if you have git bash it will be there I think) or similar to convert to unix text file and then do sha1sum and the value should then be correct.

tkrokli:
Sorry, I generated a SHA1 hash because that was what came up when I checked the DB table info for the `updates` table. If you are able to generate a hash locally from the auto import of the file, use yours.

tkrokli:
To look up your local hash, run this query in `mysql` or your SQL DB editor:
```sql
SELECT * FROM `characters`.`updates` WHERE `name`= '9999_99_99_18328_characters.sql';
```
If you use a different name for your `characters` table, remove/replace that part in my  suggested query.

iridinite:
Come on, gents, don't leave me hanging again. If you think this is a shitty patch, just tell me, then at least I know which way is up.

Shauren:
No, this is actually pretty good

Shauren (approved):

tkrokli:
Looks good this time around, travis-ci has accepted the updated `sql/base/characters_database.sql` file (and all the other SQL parts).

Issue #15952, opened 2015-12-07
[3.3.5] Spells movement problem - Scatter Shot visual bug


Re3os:
Rev https://github.com/TrinityCore/TrinityCore/commit/bffaa92633937d220ab19f565faa5906995530a2
Branch 3.3.5
Patch: Clean TC.

**Example:**
If you use Scatter Shot in moviment targets, he "teleport" in desorientation moviment.

Old Commit has fixed that, dont work anymore (https://github.com/TrinityCore/TrinityCore/pull/7975)

Videos of servers based at Trinity

https://www.youtube.com/watch?v=28aRmHO-WpQ
https://www.youtube.com/watch?v=pSDQ5yZGmx4

How should work : https://www.youtube.com/watch?time_continue=119&v=PVhwiHmV1BI
(Even as a private server based on TC, is there working properly, i dont have time for search vid on retail)

Tc Rev 8a0bbc386982612077e6f83f1816b02be757e371

Old Issues: #2689 #2013 #466 #1011

old fix for this was https://github.com/TrinityCore/TrinityCore/commit/3d644d8d119b9009f10ea637c52a01dad32708c9

but this no longer works

Close issue still problem
https://github.com/TrinityCore/TrinityCore/issues/9475
https://github.com/TrinityCore/TrinityCore/issues/11530
https://github.com/TrinityCore/TrinityCore/issues/11530
https://github.com/TrinityCore/TrinityCore/issues/11597

digz6666:
Confirmed.

Goatform:
Confirmed.

DevRival:
Maybe hack, but work for me. 
[Fix](https://gist.github.com/DevRival/52696a4a3c6b49442629)

Re3os:
@DevRival it's not work.
Maybe most change?

Rushor:
the must be some option to stop the client-side movement. there is a similar problem if you cast scatter on someone. https://github.com/TrinityCore/TrinityCore/issues/466 

example:
hunter A und mage B start duel
mage B start running in any direction
hunter A casts scatter on mage B
if the disorient effect hits mage B, he directly stops movement (from the point of viwe of the mage)
if the disorient effect hits the mage, he runs 1-2 seconds in the direction he originally wants to reach and then he teleports back to the position where the disorient effect was applied (from the point of viwe of the hunter A)

Rushor:
[...]

PR #18336, opened 2016-11-28
Core/Spells: Fix Scatter Shot issue on moving players

chaodhib:
From https://github.com/TrinityCore/TrinityCore/issues/15952#issuecomment-263414400:

> First, I noticed that the target was not stopped right there, as soon as the spell lands, if the target is a player. By sniffing the communication between the client and TC while testing (something that more developers should do, it really helps debugging), i realized that the server wasn't sending the packet to make the player stop. By logging some key methods, I managed the trace the problem. First, @robinsch's change is needed because `unit->IsStopped()` returns true on moving players for some reason. Second, the call to `unit->StopMoving()` does nothing if the condition movespline->Finalized() is verified, which is usually the case for players.

**Changes proposed:**
-  Bypassing `unit->StopMoving()` in `ConfusedMovementGenerator<T>::DoInitialize` for now because it does nothing if the condition `movespline->Finalized()` is verified, which is usually the case for players.

**Target branch(es):**

- [X] 3.3.5

**Issues addressed:** Closes #  https://github.com/TrinityCore/TrinityCore/issues/15952

**Tests performed:** Tested IG. Works against moving players.

This is a quick and dirty solution. A solution involving a cleaner method needs to be found before this can get merged.

### DIFF
--- a/src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
@@ -33,10 +33,14 @@ void ConfusedMovementGenerator<T>::DoInitialize(T* unit)
     if (!unit->IsAlive())
         return;
 
+    // send to clients the order to immobilize the unit and make it face a random direction.
     Movement::MoveSplineInit init(unit);
     init.MoveTo(i_x, i_y, i_z, false, true);
+    init.SetFacing(frand(0.0f, 2*(float)M_PI));
+    init.SetWalk(true);
     init.Launch();
 
+    unit->ClearUnitState(UNIT_STATE_MOVING);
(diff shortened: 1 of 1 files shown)


Persephone:
Tested ingame, and it does fix it. It does however, also mess with Fear pathing. On flat ground, fear pathing seems improved a lot. But it does have some odd movements in for example Warsong Gulch. I think your on the right track tho :)

robinsch:
@Persephone ConfusedMovementGenerator has nothing to do with fear movement.

chaodhib:
@Persephone Thanks. Like @robinsch said though, I didn't make any change to fear or the MovementGenerator its using.

Keader:
@chaodhib please rebase your PR to fix travis ^^

chaodhib:
fix rewritten here: https://github.com/TrinityCore/TrinityCore/issues/18889

PR #18367, opened 2016-11-30
Minor MotionMaster refactor

chaodhib:
**Changes proposed:**
a bit of clean up to MotionMaster::UpdateMotion()

- [x] 3.3.5

**Issues addressed:** Does not change any behavior. This is just code clean up.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -467,6 +467,7 @@ void Unit::Update(uint32 p_time)
 
     UpdateSplineMovement(p_time);
     i_motionMaster->UpdateMotion(p_time);
+    UpdateUnderwaterState(GetMap(), GetPositionX(), GetPositionY(), GetPositionZ());
 }
 
 bool Unit::haveOffhandWeapon() const
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -88,10 +88,10 @@ void MotionMaster::UpdateMotion(uint32 diff)
     ASSERT(!empty());
 
     _cleanFlag |= MMCF_UPDATE;
-    bool IsMoveGenUpdateSuccess = top()->Update(_owner, diff);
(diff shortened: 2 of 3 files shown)


ccrs:
Merging in a couple mins, if you dont want to include anything else.

chaodhib:
Nop, that's all I have for now. Thanks

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -467,6 +467,7 @@ void Unit::Update(uint32 p_time)
 
     UpdateSplineMovement(p_time);
     i_motionMaster->UpdateMotion(p_time);
+    UpdateUnderwaterState(GetMap(), GetPositionX(), GetPositionY(), GetPositionZ());
```
Shauren: expensive...

chaodhib: This method was called in UpdateMotion() (the line above). So it was called on each Unit::Update() call anyway. And putting this call here is better than in MotionMaster::UpdateMotion() imo.

Takenbacon: Seems like it's pretty pointless considering it's already called in Unit::UpdatePosition

chaodhib: I removed it from Unit::UpdatePosition and moved it in Unit::Update. This method better belong here than in the MotionMaster.

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -88,37 +88,14 @@ void MotionMaster::UpdateMotion(uint32 diff)
     ASSERT(!empty());
 
     _cleanFlag |= MMCF_UPDATE;
-    if (!top()->Update(_owner, diff))
-    {
-        _cleanFlag &= ~MMCF_UPDATE;
+    bool IsMoveGenUpdateSuccess = top()->Update(_owner, diff);
```
Shauren: you call it a refactor but dont follow codestyle (local variable naming)...

chaodhib: fixed

PR #18463, opened 2016-12-12
Core/Chat: make Player commands behave like retail + configurable

Estaira:
**Changes proposed:**
Allows to disable or enable player commands just like in retail.

- [x] 3.3.5
- [x] master

**Issues addressed:**
Fixes issue with players able to type in commands making it not blizzlike standards. 

**Tests performed:** Builds and has been tested in-game. 

**Known issues and TODO list:** None.

### DIFF
--- a/src/server/game/Chat/Chat.cpp
+++ b/src/server/game/Chat/Chat.cpp
@@ -397,6 +397,7 @@ bool ChatHandler::ParseCommands(char const* text)
     ASSERT(*text);
 
     std::string fullcmd = text;
+    
     if (m_session && AccountMgr::IsPlayerAccount(m_session->GetSecurity()) && !sWorld->getBoolConfig(CONFIG_ALLOW_PLAYER_COMMANDS))
        return false;
     /// chat case (.command or !command format)
--- a/src/server/game/Chat/Chat.cpp
+++ b/src/server/game/Chat/Chat.cpp
@@ -397,6 +397,9 @@ bool ChatHandler::ParseCommands(char const* text)
     ASSERT(*text);
 
     std::string fullcmd = text;
+	
+	if (m_session && AccountMgr::IsPlayerAccount(m_session->GetSecurity()) && !sWorld->getBoolConfig(CONFIG_ALLOW_PLAYER_COMMANDS))
+       return false;
 
     /// chat case (.command or !command format)
     if (m_session)
--- a/src/server/game/Chat/Chat.cpp
+++ b/src/server/game/Chat/Chat.cpp
@@ -397,10 +397,8 @@ bool ChatHandler::ParseCommands(char const* text)
     ASSERT(*text);
 
     std::string fullcmd = text;
-	
-	if (m_session && AccountMgr::IsPlayerAccount(m_session->GetSecurity()) && !sWorld->getBoolConfig(CONFIG_ALLOW_PLAYER_COMMANDS))
+    if (m_session && AccountMgr::IsPlayerAccount(m_session->GetSecurity()) && !sWorld->getBoolConfig(CONFIG_ALLOW_PLAYER_COMMANDS))
        return false;
-
     /// chat case (.command or !command format)
     if (m_session)
     {
--- a/src/server/game/World/World.cpp
+++ b/src/server/game/World/World.cpp
@@ -662,6 +662,7 @@ void World::LoadConfigSettings(bool reload)
     m_int_configs[CONFIG_TRADE_LEVEL_REQ] = sConfigMgr->GetIntDefault("LevelReq.Trade", 1);
     m_int_configs[CONFIG_AUCTION_LEVEL_REQ] = sConfigMgr->GetIntDefault("LevelReq.Auction", 1);
     m_int_configs[CONFIG_MAIL_LEVEL_REQ] = sConfigMgr->GetIntDefault("LevelReq.Mail", 1);
+	m_bool_configs[CONFIG_ALLOW_PLAYER_COMMANDS] = ConfigMgr::GetBoolDefault("AllowPlayerCommands", 1);
     m_bool_configs[CONFIG_PRESERVE_CUSTOM_CHANNELS] = sConfigMgr->GetBoolDefault("PreserveCustomChannels", false);
     m_int_configs[CONFIG_PRESERVE_CUSTOM_CHANNEL_DURATION] = sConfigMgr->GetIntDefault("PreserveCustomChannelDuration", 14);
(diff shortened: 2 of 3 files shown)


joschiwald:
you can control it over rbac

Kizarr:
This is a more efficient and blizzlike standard way of doing this. It would be easier to disable from worldserver configs and to implement this. Say a player try's to type in command it will not show up as saying no such sub command or of the sorts instead just like retail it shows up .gm fly on in text instead.

tkrokli:
Remember to add the config option in `worldserver.conf.dist` too, with description.

joschiwald:
you have a RBAC_PERM_COMMANDS_NOTIFY_COMMAND_NOT_FOUND_ERROR

src/server/game/World/World.cpp
```diff
@@ -662,6 +662,7 @@ void World::LoadConfigSettings(bool reload)
     m_int_configs[CONFIG_TRADE_LEVEL_REQ] = sConfigMgr->GetIntDefault("LevelReq.Trade", 1);
     m_int_configs[CONFIG_AUCTION_LEVEL_REQ] = sConfigMgr->GetIntDefault("LevelReq.Auction", 1);
     m_int_configs[CONFIG_MAIL_LEVEL_REQ] = sConfigMgr->GetIntDefault("LevelReq.Mail", 1);
+	m_bool_configs[CONFIG_ALLOW_PLAYER_COMMANDS] = ConfigMgr::GetBoolDefault("AllowPlayerCommands", 1);
```
tkrokli:
```
/home/travis/build/TrinityCore/TrinityCore/src/server/game/World/World.cpp:665:60: fatal error: 
      call to non-static member function without an object argument
  ...= ConfigMgr::GetBoolDefault("AllowPlayerCommands", 1);
       ~~~~~~~~~~~^~~~~~~~~~~~~~
```

Kizarr:
My bad forgot to add the comment.

tkrokli:
Please also replace tabs with spaces in Chat.cpp + World.cpp and match the indenting above & below.

Kizarr:
Okay

src/server/worldserver/worldserver.conf.dist
```diff
@@ -1804,6 +1804,14 @@ ChatLevelReq.Whisper = 1
 ChatLevelReq.Say = 1
 
 #
+#    AllowPlayerCommands
+#        Description: Allow players to use commands.
+#        Default:     1 - (Enabled)
```
DDuarte: Default doesn't match value below

src/server/game/World/World.cpp
```diff
@@ -662,6 +662,7 @@ void World::LoadConfigSettings(bool reload)
     m_int_configs[CONFIG_TRADE_LEVEL_REQ] = sConfigMgr->GetIntDefault("LevelReq.Trade", 1);
     m_int_configs[CONFIG_AUCTION_LEVEL_REQ] = sConfigMgr->GetIntDefault("LevelReq.Auction", 1);
     m_int_configs[CONFIG_MAIL_LEVEL_REQ] = sConfigMgr->GetIntDefault("LevelReq.Mail", 1);
+    m_bool_configs[CONFIG_ALLOW_PLAYER_COMMANDS] = ConfigMgr::GetBoolDefault("AllowPlayerCommands", 1);
```
DDuarte: Use true instead of 1

src/server/game/Chat/Chat.cpp
```diff
@@ -397,7 +397,8 @@ bool ChatHandler::ParseCommands(char const* text)
     ASSERT(*text);
 
     std::string fullcmd = text;
-
+    if (m_session && AccountMgr::IsPlayerAccount(m_session->GetSecurity()) && !sWorld->getBoolConfig(CONFIG_ALLOW_PLAYER_COMMANDS))
+       return false;
```
Kizarr: Okay.

Warpten:
Someone explain to me how being able to enter commands on retail is a thing - and blizzlike, on top of that.

tkrokli:
@Kizarr : is it currently impossible to use _any_ chat commands for normal (non-GM) players?

Kizarr:
You can use /commands but not any ! or .commands.

Kizarr:
@Warpten I believe you're a bit confused and far off from what my intention is from this commit. What I am referring to disabling player command function. Say for instance you type in command .gm on. It will show for a player no such command. Instead in retail say if you type command .  or ! you type .gm on it will say out loud .gm on appear as there no command system for . ! Just like in retail it this should be added because this a blizzlike standard and has been implemented in retail.

nawuko:
@Kizarr You are sure that RBAC_PERM_COMMANDS_NOTIFY_COMMAND_NOT_FOUND_ERROR isn't working? ExecuteCommandInTable should return false when the command is not available for a player. If this is the case the logic in ExecuteCommandInTable should be fixed instead of adding a new code.

Kizarr:
Yes this isn't working RBAC_PERM_COMMANDS_NOTIFY_COMMAND_NOT_FOUND_ERROR. This was code was never finished in the core. Although it was partially implemented. CONFIG_ALLOW_PLAYER_COMMANDS, line 105 https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/World/World.h

joschiwald:
i'm sure this code (pr) was dropped because of rbac implementation

Kizarr:
If it was dropped then this is not working correctly RBAC_PERM_COMMANDS_NOTIFY_COMMAND_NOT_FOUND_ERROR

It should not even notify players about commands etc. read above.

joschiwald:
it requires that players not have this permission

Kizarr:
Let me get you picture's for example to better explain.

Kizarr:
Yes.

Pixel-Pirate:
What is the utility of this script? I do not see

Kizarr:
With Allow Player Commands Disabled:
http://prntscr.com/dj5y1q

Without player Commands disabled using non-blizzlike standards when players shouldn't be able use commands.
http://prntscr.com/dj5z8s

joschiwald:
works for me with my own rbac configuration

Kizarr:
Go on to retail and try to type in command .commands .gm on !commands or something you'll see that players can't use command at all.

joschiwald:
you have to adjust rbac permissions according to your needs

Shauren:
RBAC_PERM_COMMANDS_NOTIFY_COMMAND_NOT_FOUND_ERROR actually does not work for subcommands (its a bug that needs fixing, not a config workaround like this)

PR #18513, opened 2016-12-19
Core/Chat: Added option to allow 5-man parties to use raid warnings.

### LINKED ISSUE #17889: [3.3.5] Raid Warning in parties
3.3.5 f5cbc17

While in a party, you can't use `/rw`. Expected behavior is being able to use `/rw` (any member can do so, no need to be leader).

adeutscher:
***Changes proposed:***

These changes would allow for the use of Raid Warnings when in a 5-man party. All party members have permission to create raid warnings.

This feature is controlled by a setting in `worldserver.conf`, and is off by default in order to be Blizzlike for 3.3.5.

***Target branch(es):*** 3.3.5

***Issues addressed:*** Closes #17889

***Tests performed:***

Built and tested in game under 3.3.5.

* Tested to confirm that non-leaders/assistants in raid groups are still unable to use raid warnings.
* 5-man party members are now able to use raid warnings with the new setting enabled.

***Details***

A TC server will **currently** return without broadcasting if one of the following is True:

* Player is not in a group.
* Player is not in a raid group.
* Player is not the leader or an assistant of the group.
* Player is in a BG group.

My current plan would return without broadcasting if one of the following is True:

* Player is not in a group.
* Player is in a raid group where they are not the leader or an assistant.
* Player is in a BG group.
* The new raid warnings setting is not enabled.


src/server/game/World/World.cpp
```diff
@@ -1349,6 +1349,9 @@ void World::LoadConfigSettings(bool reload)
     // prevent character rename on character customization
     m_bool_configs[CONFIG_PREVENT_RENAME_CUSTOMIZATION] = sConfigMgr->GetBoolDefault("PreventRenameCharacterOnCustomization", false);
 
+    // Allow 5-man parties to use raid warnings
+    m_int_configs[CONFIG_CHAT_PARTY_RAID_WARNINGS] = sConfigMgr->GetIntDefault("PartyRaidWarnings", 0);
```
DDuarte: Perhaps this could be in m_bool_configs instead of int instead.

adeutscher: That makes sense to me. Pushing upwards after a quick test build is done.

src/server/game/Handlers/ChatHandler.cpp
```diff
@@ -412,7 +412,7 @@ void WorldSession::HandleMessagechatOpcode(WorldPacket& recvData)
         case CHAT_MSG_RAID_WARNING:
         {
             Group* group = GetPlayer()->GetGroup();
-            if (!group || !group->isRaidGroup() || !(group->IsLeader(GetPlayer()->GetGUID()) || group->IsAssistant(GetPlayer()->GetGUID())) || group->isBGGroup())
+            if (!group || (group->isRaidGroup() && !(group->IsLeader(GetPlayer()->GetGUID()) || group->IsAssistant(GetPlayer()->GetGUID()))) || group->isBGGroup() || !sWorld->getBoolConfig(CONFIG_CHAT_PARTY_RAID_WARNINGS))
```
ccrs: Correct me if I'm wrong, but this will make anyone in the raid group be able to send raid warnings if the option is enabled

adeutscher: If the user is in a raid group and not a leader/assistant, then the IF statement will short-circuit when `(group->isRaidGroup() && !(group->IsLeader(GetPlayer()->GetGUID()) || group->IsAssistant(GetPlayer()->GetGUID())))` evaluates to True and never attempt the config check at the very end.

The configuration setting will only be checked if they are in a group that is not a raid group or a BG group.

adeutscher: @ccrs Looked over my work again after 8a07fe7, and you were entirely correct. Sorry about the hiccup.

PR #18523, opened 2016-12-19
Handle Quests.IgnoreAutoAccept in the post-processing phase of quest …

### LINKED ISSUE #17403: Worldserver.conf - Quests.IgnoreAutoAccept - breaks starting quests
**Description**:
Starting quests in starting zones (level 1-6) are not allowing characters to accept

Upon using the setting "Quests.IgnoreAutoAccept = 1" in worldserver.conf then low level quests in starting zones are not allowed to be manually accepted

Quests should act as any normal quests do of higher level, allow you to click the accept button and roll out!

In my ignorant attempts to recompile last night I reused my worldserver.conf file, figured it would save me time instead of having to input new DB info and such.  After using a fresh one today and figuring out that the following setting was causing the quests to not work:

```
#    Quests.IgnoreAutoAccept
#        Description: Ignore auto accept flag. Clients will have to manually accept all quests.
#        Default:     0 - (Disabled, DB values determine if quest is marked auto accept or not.)
#                     1 - (Enabled, clients will not be told to automatically accept any quest.)

Quests.IgnoreAutoAccept = 0
```

If I set Quests.IgnoreAutoAccept = 1  then no starting quests will work at all.  

So instead of the bug being that the quests dont work it is that the "Quests.IgnoreAutoAccept = 1" breaks being able to accept starting quests.

**Branch(es)**: 3.3.5

**TC hash/commit**:  
TrinityCore rev. 90ff3ce 2016-06-16 17:53:14 +0200 (3.3.5 branch) (Win32, Release)

**TDB version**:  
335.61

**Operating system**:  
Host: Server 2008R2 
Client: Windows 10 & Windows 7 both tested.

adeutscher:
**Changes proposed:**

- During post-processing in `ObjectMgr::LoadQuests` when quests are loaded from the database, remove the `QUEST_FLAGS_AUTO_ACCEPT` flag if we are dealing with an auto-accept quest and `Quests.IgnoreAutoAccept` is set to True (1) in `worldserver.conf`
-  Change `Quest::IsAutoAccept` to only check the `Quest` object's flag, as the post-processing addition should make the configuration check here redundant.

**Target branch(es):** 3.3.5

**Issues addressed:** Closes #17403

Built and tested in game.

- When `Quests.IgnoreAutoAccept` is set to ***0***, the quest is accepted the moment that the dialog comes up, which is consistent with how things were before this PR. I don't think that this is entirely Blizzlike, but that's a hiccup for another issue.
- When `Quests.IgnoreAutoAccept` is set to ***1***, the quest now appears in the quest log after the "Accept" button is pressed. Previously, the quest dialog would disappear but no quest would be accepted.

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -4711,6 +4711,10 @@ void ObjectMgr::LoadQuests()
             qinfo->SetSpecialFlag(QUEST_SPECIAL_FLAGS_TIMED);
         if (qinfo->RequiredPlayerKills)
             qinfo->SetSpecialFlag(QUEST_SPECIAL_FLAGS_PLAYER_KILL);
+
+        // If forcing manual quest acceptance through config, then remove auto-accept flag
+        if (qinfo->Flags & QUEST_FLAGS_AUTO_ACCEPT && sWorld->getBoolConfig(CONFIG_QUEST_IGNORE_AUTO_ACCEPT))
+            qinfo->Flags &= ~QUEST_FLAGS_AUTO_ACCEPT;
     }
 
     // check QUEST_SPECIAL_FLAGS_EXPLORATION_OR_EVENT for spell with SPELL_EFFECT_QUEST_COMPLETE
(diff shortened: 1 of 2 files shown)


adeutscher:
Closing this pull request, hopefully made obsolete by #18525.

@Shauren pointed out that this PR's method could introduce client caching issues if the option was changed and the server configuration were to be reloaded.

PR #18528, opened 2016-12-19
Core/Vmaps: Stop M2s from occluding for spellcast LoS

HelloKitty:
**Changes proposed:**

Thanks to Robinsch for his help.

-  Change LoS/Raycasting API to include optional flag that acts as layer mask.
-  Create flags that can mask out objects that reside on layers (WMO, M2 and etc)  
-  Fix the spell casting LoS to not account for M2 objects in the world based on  Subv's [2012 commit](https://github.com/TrinityCore/TrinityCore/commit/c53d722aa29ff3a39d881de7d6ef86b79deffe48) that was eventually reverted due to a pathfinding bug (a bug that is fixed in this version).

**Target branch(es):** 3.3.5

- [x] 3.3.5

This commit addresses the fact that M2 objects should not occlude targets when calculating LoS. However, M2s should be included for pathfinding. The tight coupling of the raycasting/intersection system and LoS and Pathfinding makes it difficult to produce a clean solution. Because there was no encapsulation for the concept of a Ray or RayTestInfo, that you may find in other physics systems, it was not possible to extend without an API change. This appears to essentially preform the role of the 2012 commit without breaking the pathfinding like it did before.

Closes #3495
Closes #5312

Builds. Tested pathfinding and LoS. LoS ignores M2s. Pets and pathfinding still respect M2 placement. Runs on a public 3.3.5 server too. No crashes so far from this change. Behavior seems to be correct.

None that I know of.

P.S. I left the [2012 commit](https://github.com/TrinityCore/TrinityCore/commit/c53d722aa29ff3a39d881de7d6ef86b79deffe48) as is even though I think it should be cleaned up. It was merged so I figured it was good enough.

### DIFF
--- a/src/common/Collision/Management/VMapManager2.cpp
+++ b/src/common/Collision/Management/VMapManager2.cpp
@@ -28,6 +28,7 @@
 #include "Log.h"
 #include "VMapDefinitions.h"
 #include "Errors.h"
+#include "ObjectIgnoreDefinitions.h"
 
 using G3D::Vector3;
 
@@ -162,7 +163,8 @@ namespace VMAP
         }
     }
 
-    bool VMapManager2::isInLineOfSight(unsigned int mapId, float x1, float y1, float z1, float x2, float y2, float z2)
+    //SPELL LOS CALLSTACK
+    bool VMapManager2::isInLineOfSight(unsigned int mapId, float x1, float y1, float z1, float x2, float y2, float z2, ObjectIgnoreFlags ignoreFlags)
     {
         if (!isLineOfSightCalcEnabled() || IsVMAPDisabledForPtr(mapId, VMAP_DISABLE_LOS))
             return true;
@@ -174,7 +176,7 @@ namespace VMAP
             Vector3 pos2 = convertPositionToInternalRep(x2, y2, z2);
             if (pos1 != pos2)
             {
-                return instanceTree->second->isInLineOfSight(pos1, pos2);
+                return instanceTree->second->isInLineOfSight(pos1, pos2, ignoreFlags);
             }
         }
 
@@ -279,7 +281,7 @@ namespace VMAP
         return false;
     }
 
-    WorldModel* VMapManager2::acquireModelInstance(const std::string& basepath, const std::string& filename)
+    WorldModel* VMapManager2::acquireModelInstance(const std::string& basepath, const std::string& filename, uint32 flags/* Only used when creating the model */)
     {
         //! Critical section, thread safe access to iLoadedModelFiles
         std::lock_guard<std::mutex> lock(LoadedModelFilesLock);
@@ -295,6 +297,9 @@ namespace VMAP
                 return NULL;
             }
             VMAP_DEBUG_LOG("maps", "VMapManager2: loading file '%s%s'", basepath.c_str(), filename.c_str());
+
+            worldmodel->Flags = flags;
+
             model = iLoadedModelFiles.insert(std::pair<std::string, ManagedModel>(filename, ManagedModel())).first;
             model->second.setModel(worldmodel);
         }
--- a/src/common/Collision/Management/VMapManager2.cpp
+++ b/src/common/Collision/Management/VMapManager2.cpp
@@ -28,7 +28,6 @@
 #include "Log.h"
 #include "VMapDefinitions.h"
 #include "Errors.h"
-#include "ObjectIgnoreDefinitions.h"
 
 using G3D::Vector3;
 
--- a/src/common/Collision/Management/VMapManager2.cpp
+++ b/src/common/Collision/Management/VMapManager2.cpp
@@ -163,7 +163,6 @@ namespace VMAP
         }
     }
 
-    //SPELL LOS CALLSTACK
     bool VMapManager2::isInLineOfSight(unsigned int mapId, float x1, float y1, float z1, float x2, float y2, float z2, ObjectIgnoreFlags ignoreFlags)
     {
         if (!isLineOfSightCalcEnabled() || IsVMAPDisabledForPtr(mapId, VMAP_DISABLE_LOS))
--- a/src/common/Collision/Management/VMapManager2.h
+++ b/src/common/Collision/Management/VMapManager2.h
@@ -24,7 +24,6 @@
 #include <vector>
 #include "Define.h"
 #include "IVMapManager.h"
-#include "ObjectIgnoreDefinitions.h"
 
 //===========================================================
 
--- a/src/common/Collision/Management/VMapManager2.h
+++ b/src/common/Collision/Management/VMapManager2.h
@@ -24,6 +24,7 @@
 #include <vector>
 #include "Define.h"
 #include "IVMapManager.h"
+#include "ObjectIgnoreDefinitions.h"
 
 //===========================================================
 
@@ -107,7 +108,7 @@ namespace VMAP
             void unloadMap(unsigned int mapId, int x, int y) override;
             void unloadMap(unsigned int mapId) override;
 
-            bool isInLineOfSight(unsigned int mapId, float x1, float y1, float z1, float x2, float y2, float z2) override ;
+            bool isInLineOfSight(unsigned int mapId, float x1, float y1, float z1, float x2, float y2, float z2, ObjectIgnoreFlags ignoreFlags) override ;
             /**
             fill the hit pos and return true, if an object was hit
             */
@@ -119,7 +120,7 @@ namespace VMAP
             bool getAreaInfo(unsigned int pMapId, float x, float y, float& z, uint32& flags, int32& adtId, int32& rootId, int32& groupId) const override;
             bool GetLiquidLevel(uint32 pMapId, float x, float y, float z, uint8 reqLiquidType, float& level, float& floor, uint32& type) const override;
 
-            WorldModel* acquireModelInstance(const std::string& basepath, const std::string& filename);
+            WorldModel* acquireModelInstance(const std::string& basepath, const std::string& filename, uint32 flags = 0);
             void releaseModelInstance(const std::string& filename);
 
             // what's the use of this? o.O
--- a/src/common/Collision/Maps/MapTree.cpp
+++ b/src/common/Collision/Maps/MapTree.cpp
@@ -23,6 +23,7 @@
 #include "Log.h"
 #include "Errors.h"
 #include "Metric.h"
+#include "ObjectIgnoreDefinitions.h"
 
 #include <string>
 #include <sstream>
@@ -33,14 +34,14 @@ using G3D::Vector3;
 
 namespace VMAP
 {
-
+    //SPELL LOS CALLSTACK
     class MapRayCallback
     {
         public:
-            MapRayCallback(ModelInstance* val): prims(val), hit(false) { }
+            MapRayCallback(ModelInstance* val, ObjectIgnoreFlags ignoreFlags): prims(val), hit(false), flags(ignoreFlags) { }
             bool operator()(const G3D::Ray& ray, uint32 entry, float& distance, bool pStopAtFirstHit=true)
             {
-                bool result = prims[entry].intersectRay(ray, distance, pStopAtFirstHit);
+                bool result = prims[entry].intersectRay(ray, distance, pStopAtFirstHit, flags);
                 if (result)
                     hit = true;
                 return result;
@@ -49,6 +50,7 @@ namespace VMAP
     protected:
         ModelInstance* prims;
         bool hit;
+        ObjectIgnoreFlags flags;
     };
 
     class AreaInfoCallback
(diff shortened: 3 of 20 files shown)


Ovahlord:
does this include things like invisible walls? I mean iron docks had some of them which were preventing skulloc from hitting the other lanes cuz the LoS was blocking it.

HelloKitty:
@Ovahlord I'm not sure, I've only tested this on 3.3.5. Basically it's Subv's 2012 commit that excludes objects that are derived from M2 files. Not sure if invisible walls would have that flag on it.

You could extend the flags though if you wanted to. It is left open ended. Not beautiful but at least slightly extendable.

src/common/Collision/Models/ObjectIgnoreDefinitions.h
```diff
@@ -0,0 +1,29 @@
...
+#define _OBJECTIGNOREDEFINITIONS_H_
+
+enum ObjectIgnoreFlags
+{
+	IGNORE_NONE = 0x00,
+	IGNORE_M2 = 0x01
+};
+
+
+#endif
```
Aokromes: no new line.

HelloKitty: Ok, I untabified and added a newline there. Sorry about that.

HelloKitty:
Oh just as a warning I added some checks for SPELL_EFFECT_CHARGE or whatever it was called to prevent charging through M2s. Also did the same thing for an unknown spell attribute that all Death Grip spells seems to share.

There may be other abilities that require both a path and LoS. This leads me to believe that these spells should be not checking LoS but something to the effect of isPathTo(...) && IsWithinLoS(...) but someone else should make those sorts of design decisions. I could be very wrong but I'd think Blizzard probably autogenerated these checks based on the attributes on spells.

Anyway, other spells may need "full" LoS checks too. Like Feral Charge - Cat. I don't know.

src/common/Collision/Management/VMapManager2.cpp
```diff
@@ -279,7 +280,7 @@ namespace VMAP
         return false;
     }
 
-    WorldModel* VMapManager2::acquireModelInstance(const std::string& basepath, const std::string& filename)
+    WorldModel* VMapManager2::acquireModelInstance(const std::string& basepath, const std::string& filename, uint32 flags/* Only used when creating the model */)
```
robinsch: I would rather keep this as it is now and make a setter for flags. Just personal preferences because it feels wrong imo

HelloKitty: Yea, I agree. I think in general that method shouldn't be both used for loading and creating.

However, since this was how the accepted and merged commit you linked me the other day [had it](https://github.com/TrinityCore/TrinityCore/commit/c53d722aa29ff3a39d881de7d6ef86b79deffe48) I left it as is.

It has a default defined at least. I made a note that I didn't fully agree with how the originally commit was implemented, I believe it exposed a public field too, but I didn't want to make changes to it since it had actually been accepted in the past that way.

DDuarte:
Up to discussion: wouldn't it be better if the enum was added to an existing header?

HelloKitty:
@DDuarte Yes definitely! I've nearly no C++ experience so maybe I was missing something but it seemed like there was no viable place to put it. Anywhere I saw an opportunity to put it either wasn't referenced by what needed it or referencing it would cause a circle reference.

If you can find a good header to include this in I very much wanted that. Maybe I missed something but I just couldn't find one such header to do that to unless I added it to maybe something like Common.h or something.

src/server/game/Entities/Object/Object.h
```diff
@@ -24,6 +24,7 @@
 #include "UpdateMask.h"
 #include "GridReference.h"
 #include "ObjectDefines.h"
+#include "ObjectIgnoreDefinitions.h"
```
Eliminationzx: superfluous inclusion

src/server/game/Maps/Map.cpp
```diff
@@ -31,6 +31,7 @@
 #include "MapInstanced.h"
 #include "ObjectAccessor.h"
 #include "ObjectMgr.h"
+#include "ObjectIgnoreDefinitions.h"
```
Eliminationzx: superfluous inclusion

src/server/game/Maps/Map.h
```diff
@@ -31,6 +31,7 @@
 #include "DynamicTree.h"
 #include "GameObjectModel.h"
 #include "ObjectGuid.h"
+#include "ObjectIgnoreDefinitions.h"
```
Eliminationzx: superfluous inclusion

src/server/game/Movement/MovementGenerators/FleeingMovementGenerator.cpp
```diff
@@ -55,7 +55,7 @@ void FleeingMovementGenerator<T>::_setTargetLocation(T* owner)
                                                                                 mypos.m_positionX,
                                                                                 mypos.m_positionY,
                                                                                 mypos.m_positionZ + 2.0f,
-                                                                                x, y, z + 2.0f);
+                                                                                x, y, z + 2.0f, ObjectIgnoreFlags::IGNORE_NONE);
```
Eliminationzx: I guess ObjectIgnoreFlags::IGNORE_NONE looks better without namespace -> IGNORE_NONE

src/server/game/Entities/Object/Object.h
```diff
@@ -489,8 +490,8 @@ class TC_GAME_API WorldObject : public Object, public WorldLocation
         // use only if you will sure about placing both object at same map
         bool IsWithinDist(WorldObject const* obj, float dist2compare, bool is3D = true) const;
         bool IsWithinDistInMap(WorldObject const* obj, float dist2compare, bool is3D = true) const;
-        bool IsWithinLOS(float x, float y, float z) const;
-        bool IsWithinLOSInMap(WorldObject const* obj) const;
+        bool IsWithinLOS(float x, float y, float z, ObjectIgnoreFlags ignoreFlags = ObjectIgnoreFlags::IGNORE_NONE) const;
```
Eliminationzx: ObjectIgnoreFlags::IGNORE_NONE  -> IGNORE_NONE

src/server/game/Spells/Spell.cpp
```diff
@@ -1915,7 +1916,7 @@ void Spell::SearchChainTargets(std::list<WorldObject*>& targets, uint32 chainTar
                 if (Unit* unit = (*itr)->ToUnit())
                 {
                     uint32 deficit = unit->GetMaxHealth() - unit->GetHealth();
-                    if ((deficit > maxHPDeficit || foundItr == tempTargets.end()) && target->IsWithinDist(unit, jumpRadius) && target->IsWithinLOSInMap(unit))
+                    if ((deficit > maxHPDeficit || foundItr == tempTargets.end()) && target->IsWithinDist(unit, jumpRadius) && target->IsWithinLOSInMap(unit, ObjectIgnoreFlags::IGNORE_M2))
```
Eliminationzx: ObjectIgnoreFlags::IGNORE_M2 -> IGNORE_M2

src/server/game/Spells/Spell.cpp
```diff
@@ -54,6 +54,7 @@
 #include "Battlefield.h"
 #include "BattlefieldMgr.h"
 #include "TradeData.h"
+#include "ObjectIgnoreDefinitions.h"
```
Eliminationzx: superfluous inclusion

src/common/Collision/Management/VMapManager2.cpp
```diff
@@ -28,6 +28,7 @@
 #include "Log.h"
 #include "VMapDefinitions.h"
 #include "Errors.h"
+#include "ObjectIgnoreDefinitions.h"
```
Eliminationzx: superfluous inclusion

src/common/Collision/Management/VMapManager2.h
```diff
@@ -24,6 +24,7 @@
 #include <vector>
 #include "Define.h"
 #include "IVMapManager.h"
+#include "ObjectIgnoreDefinitions.h"
```
Eliminationzx: superfluous inclusion

src/common/Collision/Maps/MapTree.cpp
```diff
@@ -23,6 +23,7 @@
 #include "Log.h"
 #include "Errors.h"
 #include "Metric.h"
+#include "ObjectIgnoreDefinitions.h"
```
Eliminationzx: superfluous inclusion

src/common/Collision/Maps/MapTree.h
```diff
@@ -22,6 +22,7 @@
 #include "Define.h"
 #include "BoundingIntervalHierarchy.h"
 #include <unordered_map>
+#include "ObjectIgnoreDefinitions.h"
```
Eliminationzx: superfluous inclusion

src/common/Collision/Models/WorldModel.h
```diff
@@ -27,6 +27,14 @@
 
 #include "Define.h"
 
+//Redef from export library
```
Eliminationzx: I guess better do `#include "ModelInstance.h"` in `WorldModel.cpp`

jackpoz:
I think binary comparison works now with *.mmtile files, that way you know for sure that mmaps are generated correctly.

HelloKitty:
@jackpoz I never actually touched mmap generation I think. As close as I got was adding [this](https://github.com/TrinityCore/TrinityCore/commit/c53d722aa29ff3a39d881de7d6ef86b79deffe48) to models. It only happens at runtime when checking for intersections.

This commit was accepted in 2012 and it doesn't touch extraction in any way that I can see.

src/server/game/Entities/Object/Object.h
HelloKitty: Hmmmm, yea I suppose that would work. Are there guidelines on Enum usage in Trinitycore? My personal preference, since I come from C#, is to include Type information. I think C++11 was moving that direction too with the idea of scoped enums? Not really sure though, my C++ is very limited.

If that's how Trinitycore does enums I'll change it though.

ariel-: Enum scoping is required in case of enum class (see ObjectGuid.h for an example)

Palabola:
@HelloKitty With this tool you can visualize the problematic places not only the M2 objects cause some glitch in Vmap system. 

https://github.com/tizbac/NavMeshViewer

HelloKitty:
There seems to be debate whether charge should include M2s in LoS check. At least on CMangos. I believe that to be correct. Charge didn't path around things till like WoD or Legion so full LoS should be required to charge since it follows a linear path (excluding height).

If anyone has found some information showing otherwise in 3.3.5 then I'd happily concede. I tried to look for such information and found none though.

r00ty-tc:
Regarding path find for charge. Not absolute proof this was on retail, but if so:

https://www.youtube.com/watch?v=fRcEa1sIHEw would imply that on that version at least there was only a distance check, and then is a path available, not length of path. Allowing stupidly long paths like that,

HelloKitty:
@r00ty-tc That must be a glitch. I never saw anything like that on retail when I played. The video also says it's a glitch in the video title. So it's unlikely that anyone should use that as a reference for how charge works in my opinion

Definitely pretty ridiculous to see a warrior charge all the way up an AV tower.

r00ty-tc:
A glitch, in that there probably aren't many cases where you can do it, perhaps. But it proves without doubt that warrior charge path finds, or at least did in 2.x. Now, whether charge ALSO did a LOS check, I don't know. That's another story. I just knew I remembered warriors at least changing trajectory mid-charge and wanted to find proof it could happen.

As for charge including M2s in the check. I suspect it has to. But, assuming retail WSG has the same tree stump now as it did then, someone can try it there :P

HelloKitty:
@r00ty-tc It still does pathfind on TC though even with this fix. It just doesn't use any path that, if followed, wouldn't be in LoS and it includes the WMO, ADT and M2 model data layers. That behavior in the video is the behavior you get if you don't add a full LoS check and don't do a length length check. Both those things are done on TC. Length check was already done before this change too. I have a gif of the behavior on TC if you don't [here](https://i.gyazo.com/3a764aaf0eafb1f0b8ab0bec4e700927.gif) I never tested it much further on other things since it's not the behavior that was on retail.

If you wanted to get the behavior in the video you need to change or remove

```
float range = m_spellInfo->GetMaxRange(true, m_caster, this) * 1.5f + objSize; // can't be overly strict
```

and

```
 if (!target->IsWithinLOSInMap(m_caster)) //Do full LoS/Path check. Don't exclude m2
                        return SPELL_FAILED_LINE_OF_SIGHT;
```

But that would be silly because it's not the behavior that was on retail. What you link is just a glitch or bug.

r00ty-tc:
Yeah it's fine, and I agree that all those checks should be made. I just read it like you were suggesting we remove the pathfind aspect, and that didn't seem quite right to me.

HelloKitty:
@r00ty-tc Oh no, I'm only saying you can't rely on a valid path exclusively. You must also check that the target has LoS or alternatively check that the path if generated is straight. Checking LoS was easier to do with the current API so I went with that.

diegobfl:
I wonder if a solution was found to fix attacks between Objects, Hills, Mountains Walls? if it has been corrected please send me the Commit link

tkrokli:
Ask in [IRC #trinity @ irc.rizon.net](https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130200/IRC) for questions like that.
This is not a forum (even though @HelloKitty & @r00ty-tc made it look like one above).

PR #18562, opened 2016-12-23
Core/Vehicle: Remove immunity to Battering Ram for Ulduar vehicles

### LINKED ISSUE #15185: [3.3.5][6.x][Ulduar] Flame Leviathan -Spell Battering Ram not affecting vehicles.
Spell: Battering Ram (62376)

http://www.wowhead.com/spell=62376/battering-ram

Description: Rams the target, dealing damage and knocking it forward.

Effect#3 : 
Apply Aura: Mod % Damage Taken (Arcane, Fire, Frost, Holy, Nature, Physical, Shadow)
Value: 50%
Radius: 20 yards

Problem: The debuff is applied only if a player is hit by the spell. The correct is that cars also suffer the debuff.

Prov: Observe the right of the screen, you notice the icon of the debuff. (test on WoD)

wow-64 2015-07-20 22-07-04-75 avi_snapshot_00 49_ 2015 07 29_09 57 07

TC Rev df92d2c8d259a272bf5abfc41293f88781d03457

Lumber-Jack-2:
**Target branch(es):** 3.3.5

- [X] 3.3.5

**Issues addressed:** Closes #15185

**Tests performed:** No compiling errors, no crashes

### DIFF
--- a/src/server/game/Entities/Vehicle/Vehicle.cpp
+++ b/src/server/game/Entities/Vehicle/Vehicle.cpp
@@ -211,6 +211,11 @@ void Vehicle::ApplyAllImmunities()
             // why we need to apply this? we can simple add immunities to slow mechanic in DB
             _me->ApplySpellImmune(0, IMMUNITY_STATE, SPELL_AURA_MOD_DECREASE_SPEED, true);
             break;
+        case 335: // Salvaged Chopper
+        case 336: // Salvaged Siege Engine
+        case 338: // Salvaged Demolisher
+            _me->ApplySpellImmune(0, IMMUNITY_STATE, SPELL_AURA_MOD_DAMAGE_PERCENT_TAKEN, false); // Battering Ram
+            break;
         default:
             break;
     }


src/server/game/Entities/Vehicle/Vehicle.cpp
```diff
@@ -211,6 +211,11 @@ void Vehicle::ApplyAllImmunities()
             // why we need to apply this? we can simple add immunities to slow mechanic in DB
             _me->ApplySpellImmune(0, IMMUNITY_STATE, SPELL_AURA_MOD_DECREASE_SPEED, true);
             break;
+        case 335: // Salvaged Chopper
+        case 336: // Salvaged Siege Engine
+        case 338: // Salvaged Demolisher
+            _me->ApplySpellImmune(0, IMMUNITY_STATE, SPELL_AURA_MOD_DAMAGE_PERCENT_TAKEN, false); // Battering Ram
```
Shauren: actually I would like to know why such immunity is applied to all vehicles in the first place

Rushor: maybe Machiavelli could tell us the reason since he added it :)

but i think it was added to prevent using http://wotlk.openwow.com/spell=20233 on vehicles. Since we cannot heal vehicles i think it's fair to disable damagereduces in general too

Issue #14194, opened 2015-02-20
[3.3.5] crash MoveSpline.cpp


ikir83:
Hi all i'm running the server since November and sometimes i got this crash. Now i'm on version cb3fcea26e9d 2015-02-05 and still happens:

Any idea?

/Users/ikirwowserver/Downloads/TrinityCore/src/server/game/Movement/Spline/MoveSpline.cpp:252 in _updateState ASSERTION FAILED:
  minimal_diff >= 0
##

Aokromes:
plz supply a full crashlog.

ikir83:
Process:         worldserver [473]
Path:            /Users/*/worldserver
Identifier:      worldserver
Version:         0
Code Type:       X86-64 (Native)
Parent Process:  sudo [468]
Responsible:     Terminal [277]
User ID:         0

Date/Time:       2015-02-20 14:20:52.440 +0100
OS Version:      Mac OS X 10.9.5 (13F34)
Report Version:  11
Anonymous UUID:  8002C839-8A18-AD14-9690-447CC6FE85CA


Crashed Thread:  6

Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Codes: KERN_INVALID_ADDRESS at 0x0000000000000000

VM Regions Near 0:
--> 
    __TEXT                 000000010616b000-00000001073b0000 [ 18.3M] r-x/rwx SM=COW  /Users/*

Thread 0:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff8856a716 __psynch_cvwait + 10
1   libsystem_pthread.dylib       	0x00007fff92a65c3b _pthread_cond_wait + 727
2   libc++.1.dylib                	0x00007fff8efcbd43 std::__1::condition_variable::wait(std::__1::unique_lock<std::__1::mutex>&) + 47
3   worldserver                   	0x00000001066d14f5 MapUpdater::wait() + 69
4   worldserver                   	0x00000001066cf205 MapManager::Update(unsigned int) + 133
5   worldserver                   	0x000000010681c656 World::Update(unsigned int) + 1974
6   worldserver                   	0x00000001061745dd WorldUpdateLoop() + 349
7   worldserver                   	0x0000000106171021 main + 8497
8   libdyld.dylib                 	0x00007fff898445fd start + 1

Thread 1:
0   libsystem_kernel.dylib        	0x00007fff8856b64a kevent + 10
[...]

PR #18589, opened 2016-12-25
Possible fix for #14194 crash

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Issues addressed:** Possible fix for #14194 

**Tests performed:** Compiles properly, cannot repeat the crash so no tests

This change makes the velocity check consistent with MotionMaster::MoveJump and MotionMaster::MoveKnockbackFrom manual checks.

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1578,6 +1578,14 @@ bool WorldObject::CanSeeOrDetect(WorldObject const* obj, bool ignoreStealth, boo
                             corpseVisibility = true;
                 }
             }
+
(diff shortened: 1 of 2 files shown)


DDuarte:
The two commits seem unrelated, did you intend to make two PRs?

sirikfoll:
The minimum velocity was changed in 10 July of this year, and the crash report goes back to the previous year.
Also, there's proof(#17555) that the velocity can go bellow 0.1, maybe the other checks should be changed then ^^

xinef1:
@DDuarte Yes :< I accidentally clicked push after commit...
@sirikfoll If so, other check should be changed in this case, this request can be marked as invalid

PR #18592, opened 2016-12-26
Fixed problem where item dependent auras are not properly updated

### LINKED ISSUE #18506: [3.3.5] Spell/Warrior: Poleaxe Specialization not work properly with equipment methods

**Description:**
Poleaxe Specialization (Rank 5): Increases your chance to get a critical strike and the critical damage caused with Axes and Polearms by 5%.
If the warrior equips an axe/polearm Poleaxe Specialization works ok (crit chance increases by 5%). The problem is when the warrior (with axe/polearm equipped) changes the weapon by another (not axe/polearm) for example a mace. If the method used is "drag and drop" Poleaxe Specialization works ok (crit chance decreases 5%) but if the method used is "right click" or "macros" the crit chance does not decrease.

**Current behaviour:** 
The "right click" and "macros" equipment methods do not work with the Poleaxe specialization.

**Expected behaviour:** 
The "right click" and "macros" equipment methods should work ok with the Poleaxe specialization.

1. Look the initial crit chance and equip an axe.

2. Swap weapon using "drag and drop" method. Poleaxe Specialization works ok (crit chance decreases 5%).

3. Equip the axe again.

4. Swap weapon using "macros" or "right click". Poleaxe Specialization not work (crit chance not decrease).

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  2dbda3d

**TDB version:**  TDB_full_world_335.62_2016_10_17 + updates

**Operating system:**  Windows

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Issues addressed:** Closes #18506

**Tests performed:** Compiles and fixes the problem

### DIFF
--- a/src/server/game/Handlers/ItemHandler.cpp
+++ b/src/server/game/Handlers/ItemHandler.cpp
@@ -253,6 +253,11 @@ void WorldSession::HandleAutoEquipItemOpcode(WorldPacket& recvData)
             _player->EquipItem(eSrc, pDstItem, true);
 
         _player->AutoUnequipOffhandIfNeed();
+
+        // if inventory item was moved, check if we can remove dependent auras, because they were not removed in Player::RemoveItem (update was set to false)
+        // do this after swaps are done, we pass nullptr because both weapons could be swapped and none of them should be ignored
+        if ((srcbag == INVENTORY_SLOT_BAG_0 && srcslot < INVENTORY_SLOT_BAG_END) || (dstbag == INVENTORY_SLOT_BAG_0 && dstslot < INVENTORY_SLOT_BAG_END))
+            _player->ApplyItemDependentAuras((Item*)nullptr, false);
     }
 }
 
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -13063,9 +13063,6 @@ void Player::SwapItem(uint16 src, uint16 dst)
     RemoveItem(dstbag, dstslot, false);
     RemoveItem(srcbag, srcslot, false);
 
-    if (srcbag == INVENTORY_SLOT_BAG_0 && srcslot < INVENTORY_SLOT_BAG_END)
-        ApplyItemDependentAuras(pSrcItem, false);
-
     // add to dest
     if (IsInventoryPos(dst))
         StoreItem(sDest, pSrcItem, true);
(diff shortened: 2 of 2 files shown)


src/server/game/Handlers/ItemHandler.cpp
```diff
@@ -253,6 +253,11 @@ void WorldSession::HandleAutoEquipItemOpcode(WorldPacket& recvData)
             _player->EquipItem(eSrc, pDstItem, true);
 
         _player->AutoUnequipOffhandIfNeed();
+
+        // if inventory item was moved, check if we can remove dependent auras, because they were not removed in Player::RemoveItem (update was set to false)
+        // do this after swaps are done, we pass nullptr because both weapons could be swapped and none of them should be ignored
+        if ((srcbag == INVENTORY_SLOT_BAG_0 && srcslot < INVENTORY_SLOT_BAG_END) || (dstbag == INVENTORY_SLOT_BAG_0 && dstslot < INVENTORY_SLOT_BAG_END))
+            _player->ApplyItemDependentAuras((Item*)nullptr, false);
```
Naios: Might be too late but use `static_cast` instead of raw c-casts please.

Keader: @Naios should be used in SpellScripts too? has alot of ```me->CastSpell((Unit*) nullptr, SPELL_ID);```

Naios: Yep

ariel-: Even for nullptr casting?

Naios: Technically it doesn't make any difference when casting a nullpointer, however one should never use a c-cast from C++ because there are 4 alternatives which express your intention better then c-casts:

See http://stackoverflow.com/questions/103512/in-c-why-use-static-castintx-instead-of-intx

> The first problem is that it's almost impossible to tell which one will occur in a C-style cast without looking at large and disperse pieces of code and knowing all the rules.

xinef1: There can be only one intention of nullptr cast, but I agree that we should use c++ casts.

Shauren: There is one valid case for casting a nullptr - function overload resolution, like in the case of CastSpell

Issue #16355, opened 2016-01-23
Desync between client and server health


Treeston:
3.3.5 0316dff24530da83ef3466500ee8bd4e1af548ba

If a DamageTaken AI hook sets damage taken to zero, the client is never notified of the fact that the creature still has full health, leading to a visual desync. The client's displayed health drops, while the server health remains full (as displayed by .npc info).

Example: Phase 1 Gothik the Harvester.

Aokromes:
this issue is still valid?

CraftedRO:
b7471244adb08f112214634b27d7daf2da19aa0f If I understand correctly it seems still valid if you .go c id 16060 then damage it half hp with gm then .npc info and see base max current  health values are same (836700) even if in focus target the boss is half hp bar ~418.350.

PR #18593, opened 2016-12-26
[3.3.5/master] Core/Mechanic: Server client health desync

Krudor:

**Changes proposed:**

-  Eliminate one (common) scenario where spell damage was modified by an underlying function, but old unmodified damage value was sent to the client causing a client/server desync in health values.
- Semi unrelated change, prevent players from being able to kill themselves with self-inflicted damage with god mode cheat enabled.

**Tests performed:** Builds and appears to work as expected in-game


Shauren:
Does this mean that UnitAI DamageDealt and DamageTaken modifications are reflected in values sent in client packets? If yes then this change is incorrect (see training dummies with 1 hp - but that is not the only case)

Krudor:
This change is indeed incorrect after looking into it further.
I guess there's some missunderstanding regarding the rules of managing health.

If I understand it correct, you should NOT be allowed to modify creature damage taken server-side like DamageTaken and  DamageDealt is doing right now since it will lead to desync.

Related issue #16355

Shauren:
No. Because doing it like that for creatures staying at any fixed health (and being unkillable) is blizzlike - and combat log keeps reporting unmodified damage as well (back to training dummies example, you need original unmodifed damage sent to measure your dps as thats the entire point of these dummies)

Krudor:
Yes, but how do you support a creature staying at any fixed health, while still preserving original melee/spell damage when the damage info client package is the one telling the client whether or not a creature should take a visible chunk out of it's health?
Doesn't the damage deal packet tell the client both how much damage an ability did + it uses the same packet to remove health from the creature?
This is from what I've understood, please correct me if I'm wrong.

Krudor:
Is there a single creature on retail that has a certain health locked, except the obvious case of 1 hp.

Shauren:
http://www.wowhead.com/npc=97219/solsten this guy locks it at 2hp

Krudor:
Indeed, then I'm out of ideas for fixing this desync issue :P
A flag for setting/unsetting a units ability to die would be nice though.
Lot's of

```
if (me->GetHealth() <= damage)
    damage = me->GetHealth() - 1;
```

code to specifically make a creature unkillable
Blizzard doesn't appear to be using damage taken hook too much when scheduling health % events, it seems to be more of a repeated time event checking for hp threshholds.
On TrinityCore bosses kind of instantly SNAP into their new phases once they hit a certain health through the DamageTaken hook, compared to live where it seems to be checked periodically.

Issue #18549, opened 2016-12-21
[3.3.5 / Ruby Sanctum] Baltharus Warborn cast interrupt by UpdateSplineMovement


Keader:
**Description:** Cast of spell 76221 (SPELL_BARRIER_CHANNEL) is interrupted by movement

**Expected behaviour:** Cast of SPELL_BARRIER_CHANNEL should be interrupted only when Baltharus enter combat

**Steps to reproduce the problem:**

1. go to Baltharus (.go crea id 39751)
2. observe channel cast

Callstack from Spell::cancel() -> https://gist.github.com/Keader/943ee8c122814a0370c09b2e291e7d91

Looks like similar problem fixed in 90a58117017a93fa33864fe1892e7304e2f3bbf4

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  c809c220e29f7094121cc8489f5588d8bfee414d

xinef1:
Try this, movement shouldn't be allowed if spell is interrupted by movement
```c++
 src/server/game/Spells/SpellInfo.cpp | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/server/game/Spells/SpellInfo.cpp b/src/server/game/Spells/SpellInfo.cpp
index 8394ca7..f1c49d5 100644
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -1211,7 +1211,7 @@ bool SpellInfo::IsChanneled() const
 
 bool SpellInfo::IsMoveAllowedChannel() const
 {
-    return IsChanneled() && HasAttribute(SPELL_ATTR5_CAN_CHANNEL_WHEN_MOVING);
+    return IsChanneled() && (HasAttribute(SPELL_ATTR5_CAN_CHANNEL_WHEN_MOVING) || (ChannelInterruptFlags & (AURA_INTERRUPT_FLAG_MOVE | AURA_INTERRUPT_FLAG_TURNING)));
 }
 
 bool SpellInfo::NeedsComboPoints() const
```

Keader:
@xinef1 Tested and does not work for Baltharus

xinef1:
Well, spell 76221 has 0x1000101F channelinterruptflags (both AURA_INTERRUPT_FLAG_MOVE and AURA_INTERRUPT_FLAG_TURNING). This means that some movegen allows to move without checking for IsMovementPreventedByCasting
[...]

PR #18595, opened 2016-12-26
Prevent remove of auras interrupted by turning if the turn could be caused by float calculation error

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiled and tested with #18549

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13485,7 +13485,7 @@ bool Unit::UpdatePosition(float x, float y, float z, float orientation, bool tel
     }
 
     // Check if angular distance changed
-    bool const turn = M_PI - fabs(fabs(GetOrientation() - orientation) - M_PI) > 0.001f;
+    bool const turn = G3D::fuzzyGt(M_PI - fabs(fabs(GetOrientation() - orientation) - M_PI), 0.0f);
 
     // G3D::fuzzyEq won't help here, in some cases magnitudes differ by a little more than G3D::eps, but should be considered equal
     bool const relocated = (teleport || 
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13484,7 +13484,8 @@ bool Unit::UpdatePosition(float x, float y, float z, float orientation, bool tel
         return false;
     }
 
-    bool const turn = (GetOrientation() != orientation);
+    // Check if angular distance changed
+    bool const turn = M_PI - fabs(fabs(GetOrientation() - orientation) - M_PI) > 0.001f;
 
     // G3D::fuzzyEq won't help here, in some cases magnitudes differ by a little more than G3D::eps, but should be considered equal
     bool const relocated = (teleport || 


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -13484,7 +13484,8 @@ bool Unit::UpdatePosition(float x, float y, float z, float orientation, bool tel
         return false;
     }
 
-    bool const turn = (GetOrientation() != orientation);
+    // Check if angular distance changed
+    bool const turn = M_PI - fabs(fabs(GetOrientation() - orientation) - M_PI) > 0.001f;
```
Shauren: Thats what G3D::fuzzy* functions are for

Shauren (changes requested):

src/server/game/Entities/Unit/Unit.cpp
xinef1: Checking orientation cannot be made with simple subtraction, as we have to compare angular distance. This formula will leave result in range of 0..PI (no two points can have greater distance than PI). I tried to search for similar formula in G3D but failed to do so. Advise required.

Shauren: Position::NormalizeOrientation(GetOrientation() - orientation) // result is now in [0, 2PI) range
then turn could be expressed as `turn = G3D::fuzzyGt(Position::NormalizeOrientation(GetOrientation() - orientation), 0.0f);`

xinef1: If current orientation (GetOrientation) is 2pi-0.000001 and orientation variable equals to 0.000001 your formula will give 2pi-0.000002 and this will of course be greater than eps, but their real distance is 0.000002 (if you take the shorter path)

At lowest you would have to check 
G3D::fuzzyGt(Position::NormalizeOrientation(GetOrientation() - orientation), 0.0f) && G3D::fuzzyGt(Position::NormalizeOrientation(orientation - GetOrientation()), 0.0f);

Keader:
> [214] GUID: Full: 0xF130009B4700BB4A Type: Creature Entry: 39751 (Baltharus the Warborn) Low: 47946
> [214] Object Type: Unit (3)
> [214] Update Flags: Living, StationaryObject (96)
> [214] Movement Flags: None (0)
> [214] Extra Movement Flags: None (0)
> [214] Time: 2333322435
> [214] Position: X: 3153.061 Y: 389.4861 Z: 86.25955
> [214] Orientation: 5.637414

Current orientation is right.
And without change orientation this dont fix #18549

ref (https://github.com/TrinityCore/TrinityCore/issues/18549#issuecomment-269154707)

Idk if is acceptable change orientation like that

src/server/game/Entities/Unit/Unit.cpp
xinef1: I used the same check that was added below to keep it the same, should i correct lower checks also?

Shauren: meh, no

xinef1: I think below comment may also apply to this case "// G3D::fuzzyEq won't help here, in some cases magnitudes differ by a little more than G3D::eps, but should be considered equal"

Shauren: The code below was changed to deal with cancelling hover spells - 45f8135ac45799e555449bebb6d2a78869a439b6

xinef1: Changed as requested

Shauren (approved):

PR #18596, opened 2016-12-26
Fixed bug where spell could fail to hit a target properly even if all conditions were met

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compilation

Basicly a copy of the comment. I've commented out !m_spellInfo->StartRecoveryTime, it forces instant spells with cooldown to be processed in spell::update. As a result a spell that passed CheckCast and should be processed instantly may suffer from this delayed process. The easiest bug to observe is LoS check in AddUnitTarget, even if spell passed the CheckCast LoS check the situation can change in spell::update
 because target could be relocated in the meantime, making the spell fly to the air (no targets can be registered, so no effects processed, nothing in combat log)

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -3069,7 +3069,11 @@ void Spell::prepare(SpellCastTargets const* targets, AuraEffect const* triggered
             TriggerGlobalCooldown();
 
         //item: first cast may destroy item and second cast causes crash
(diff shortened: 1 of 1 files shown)


Shauren:
StartRecoveryTime is GCD but why is that even there in the first place

xinef1:
True, wrote the comment from memory without checking. This is probably leftover of some old code and was never changed because it was there forever, I commented this out in 2012 or 2013, as far as I've checked should not bring any problems to TC.

PR #18605, opened 2016-12-26
Fixed exploit where killing tamed pet could yield emblems

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiled and tested

Fixed exploit where killing tamed pet could yield emblems (eg King Dred killed on early 3.x patch). This may be valid for master branch.


Nyeriah:
But you can't tame King Dred...

xinef1:
@Nyeriah no true, you can do it. Anyway if an npc yields emblem and it can be tamed, emblems will be rewarded even if pet is killed while being tamed.

PR #18611, opened 2016-12-26
Fixed a typo in MovementHandler.cpp

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles

Fixed a typo according to the comment.


Kittnz:
Did you test this?

xinef1:
Had this changed since 2014 and no complaints. The logic is simple, if environmental damage didn't kill player (because of immunity aura or something else), we should kill him by force.

chaodhib:
Thanks for the PR.

I did not test the changes but it does make sense.

Palabola:
This change work, tested at Eye of the Storm and some other Zones.

PR #18613, opened 2016-12-26
Fixed empty slots handling in equipment manager

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles

Fixes wrong behavior and some bugs with equipment manager

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -1489,7 +1489,7 @@ void WorldSession::HandleEquipmentSetSave(WorldPacket& recvData)
         recvData >> itemGuid.ReadAsPacked();
 
         // if client sends 0, it means empty slot
-        if (itemGuid.GetRawValue() == 0)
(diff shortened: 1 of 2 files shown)


src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -1488,6 +1488,13 @@ void WorldSession::HandleEquipmentSetSave(WorldPacket& recvData)
         ObjectGuid itemGuid;
         recvData >> itemGuid.ReadAsPacked();
 
+        // if client sends 0, it means empty slot
+        if (itemGuid.GetRawValue() == 0)
```
Shauren: please use .IsEmpty() instead of accessing raw value for this

Shauren (changes requested):

src/server/game/Handlers/CharacterHandler.cpp
xinef1: Done

Shauren (approved):

Issue #14274, opened 2015-03-01
[DEV] Performance Improvements


jackpoz:
This is the list of known performance issues and possible regions of optimization with the goal of allowing a smooth experience in a high-populated server:
- [ ] ACE to boost asio transition seems to have decreased the performances, the 2 libraries should be compared once a proper baseline is found (e.g. 10 groups of 50 players around the world sending small packets)
- [ ] PathGenerator currently calculates the whole path from start to end instead of just retrieving the list of polygons in the path and calculating a portion of the points of that path. The target will most likely move so a lot of calculations will be wasted. The idea is to calculate only the portion that the creature will be able to walk at every Update()
- [ ] Log::GetLoggerByType() called for every TC_LOG_TRACE/DEBUG/INFO/WARN seems to be a waster of resources, reaching 3% CPU in a profiling sessions. https://github.com/jackpoz/TrinityCore/commit/65c1add47c03379498fd0dc8ba05b46fc8414a4c tackles this issue by removing at build time all these calls and leaving only TC_LOG_ERROR/FATAL logs. @DDuarte suggested to allow to specify in cmake which log types should be enabled/disabled in a more generic way

Intel:
For Log::GetLoggerByType, what about about finding highest log level during logger initialization and then ignore any higher levels in Log::ShouldLog before logger lookup happens? These cmake options and ifdefs make code really ugly and if you want to enable logs you would have to recompile core.

jackpoz:
How is code "really ugly" ? it's just a macro definition hidden somewhere in a header that doesn't require any change in all the TC_LOG_ calls.

The highest log level idea should indeed provide the same performance gains without requiring any cmake option, I'll change that and profile it.

n4ndo:
ACE better than BOOST, the performance/delay difference is very notorious.

joschiwald:
as note about loggers: https://github.com/TrinityCore/TrinityCore/commit/7228bd3664d9fa6a808e338a2fb2f519a6370672

jackpoz:
@n4ndo could you provide some data to support your comment ?

Lordron:
https://www.daniweb.com/software-development/cpp/threads/387446/performance-of-boost-asio-and-ace

jackpoz:
@Lordron any part of that discussion that you'd like to highlight ?

F03SD:
ace is better in fine-tuning and flexibility , i guess.

Lordron:
[...]

PR #18633, opened 2016-12-27
Greatly optimize achievement system by splitting large types with miscvalue

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles, basic ingame tests for a few types

This patch splits largest achievement categories by miscvalues. As of now, every spell cast or spellhit makes thousands of unnecessary loops even if this spell doesn't update any achievement. This patch should reduce load of achievement system by 90% or more. More tests required.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -6495,7 +6495,7 @@ void Player::CheckAreaExploreAndOutdoor()
     {
         SetUInt32Value(PLAYER_EXPLORED_ZONES_1 + offset, (uint32)(currFields | val));
 
-        UpdateAchievementCriteria(ACHIEVEMENT_CRITERIA_TYPE_EXPLORE_AREA);
+        UpdateAchievementCriteria(ACHIEVEMENT_CRITERIA_TYPE_EXPLORE_AREA, GetAreaId());
 
         if (areaEntry->area_level > 0)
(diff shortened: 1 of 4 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -17675,6 +17675,7 @@ bool Player::LoadFromDB(ObjectGuid guid, SQLQueryHolder *holder)
 
     _LoadDeclinedNames(holder->GetPreparedResult(PLAYER_LOGIN_QUERY_LOAD_DECLINED_NAMES));
 
+    // Should be removed
```
Shauren: should not. as bad as it seems to be, its actually blizzlike

Aokromes:
TrinityCore/TrinityCore/src/server/game/Achievements/AchievementMgr.cpp:745:13: fatal error: 
      70 enumeration values not handled in switch:
      'ACHIEVEMENT_CRITERIA_TYPE_REACH_LEVEL',
      'ACHIEVEMENT_CRITERIA_TYPE_COMPLETE_QUEST_COUNT',
      'ACHIEVEMENT_CRITERIA_TYPE_COMPLETE_DAILY_QUEST_DAILY'... [-Wswitch]
    switch (type)
            ^

xinef1:
Added defaults

PR #18636, opened 2016-12-27
Remade who list processing

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles and basic ingame tests

Requests are now processed in maps. Player entries are now copied every 5 seconds to dedicated storage (avoids usage of hashmapholder mutex)

**Improvements:** Move processing to separate thread


src/server/game/Storages/WhoListStorage.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return &instance;
+}
+
+void WhoListStorageMgr::Update()
+{
+    // clear current list
+    _whoListStorage.clear();
+    _whoListStorage.reserve(sWorld->GetPlayerCount()+1);
+
+    boost::shared_lock<boost::shared_mutex> lock(*HashMapHolder<Player>::GetLock());
```
Shauren: I've been thinking a bit about this, is this lock even needed for player storage? entries are only added/removed from the main thread (and you are looping this in main thread as well)

src/server/game/Storages/WhoListStorage.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+        std::string gname = sGuildMgr->GetGuildNameById(itr->second->GetGuildId());
+        std::wstring wgname;
+        if (!Utf8toWStr(gname, wgname))
+            continue;
+        wstrToLower(wgname);
+
+        std::string aname;
+        if (AreaTableEntry const* areaEntry = sAreaTableStore.LookupEntry(itr->second->GetZoneId()))
+            aname = areaEntry->area_name[sWorld->GetDefaultDbcLocale()];
```
Shauren: you must not store the area name, this will break lookups in other locales

src/server/game/Storages/WhoListStorage.h
```diff
@@ -0,0 +1,84 @@
...
+    ObjectGuid::LowType GetGuidLow() const { return _guidLow; }
+    uint32 GetTeam() const { return _team; }
+    AccountTypes GetSecurity() const { return _security; }
+    uint8 GetLevel() const { return _level; }
+    uint8 GetClass() const { return _class; }
+    uint8 GetRace() const { return _race; }
+    uint32 GetZoneId() const { return _zoneid; }
+    uint8 GetGender() const { return _gender; }
+    bool IsVisible() const { return _visible; }
+    std::wstring const& Getwpname() const { return _wpname; }
```
Shauren: these names are horrible...

src/server/game/Storages/WhoListStorage.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+        _whoListStorage.push_back( WhoListPlayerInfo(itr->second->GetGUID().GetCounter(), itr->second->GetTeam(), itr->second->GetSession()->GetSecurity(), itr->second->getLevel(),
```
Shauren: and since we are actually using c++11 features you could just .emplace_back(arg list) it

Shauren (changes requested):

src/server/game/Storages/WhoListStorage.cpp
xinef1: Is caching names for all locales ok?

src/server/game/Storages/WhoListStorage.cpp
xinef1: True, it's unneeded because this is done in main loop and even if other thread could run parallel to this it should not modify the player storage (except for reading)

src/server/game/Storages/WhoListStorage.cpp
Rochet2: Do you really need to cache the name at all? Just cache the ID.
Its already just 2 array lookups which are constant and lockless. (not even a hashmap lookup). Caching probably gives little to no benefit. Or then I just dont see something here.

xinef1: Right, already done that

Shauren:
Whats up with the lowguid love? Always prefer full guid unless you are saving it to database (also a bit wishful thinking: if cross-realm lfg/bg ever happens then you need full guid as it will contain realm identifier part)

xinef1:
A habit of mine

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -225,110 +226,96 @@ void WorldSession::HandleWhoOpcode(WorldPacket& recvData)
...
+    data << uint32(matchCount);                           // placeholder, count of players matching criteria
+    data << uint32(displayCount);                         // placeholder, count of players displayed
 
-    HashMapHolder<Player>::MapType const& m = ObjectAccessor::GetPlayers();
-    for (HashMapHolder<Player>::MapType::const_iterator itr = m.begin(); itr != m.end(); ++itr)
+    WhoListInfoVector const& whoList = sWhoListStorageMgr->GetWhoList();
+    for (WhoListInfoVector::const_iterator itr = whoList.begin(); itr != whoList.end(); ++itr)
     {
-        Player* target = itr->second;
+        WhoListPlayerInfo target = *itr;
```
ariel-: const& this :P

xinef1: Done

PR #18637, opened 2016-12-27
Store cached static data queries, instead of building them in every query opcode

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles and simple ingame tests

As stated in title.


Palabola:
Looks promising, but is it affect the load more than ~1% of decrease?

zorix:
You are creating `ByteBuffer` always with reserved only 1 byte maybe change this to for example to 1000 (to reduce reallocation) and add this.

```diff
diff --git a/src/server/shared/Packets/ByteBuffer.h b/src/server/shared/Packets/ByteBuffer.h
index d24a91e..5d75e75 100644
--- a/src/server/shared/Packets/ByteBuffer.h
+++ b/src/server/shared/Packets/ByteBuffer.h
@@ -422,6 +422,11 @@ class TC_SHARED_API ByteBuffer
                 _storage.reserve(ressize);
         }
 
+        void shrink_to_fit()
+        {
+            _storage.shrink_to_fit();
+        }
+
         void append(const char *src, size_t cnt)
         {
             return append((const uint8 *)src, cnt);
```

When packet was generated (and should never be changed) you can call `shrink_to_fit` to reduce memory usage. http://en.cppreference.com/w/cpp/container/vector/shrink_to_fit

xinef1:
True, you can make pr

Shauren:
@zorix vector::push_back allocates capacity()*{1.5 or 2, depends on implementaion} more storage, there is no need to worry about too many allocations that are done once during startup

zorix:
But when you will have packet that require 129 bytes of memory and starting with 1 byte it will reallocate 1->2->4->8->16->32->64->128->256 (in case *2) and 256-129= 127 bytes it's not too much but still you can get better start-up reducing this reallocations and packets for quests are bigger because of strings.

xinef1:
Well in terms of loading time it wont change much. Only memory saving aspect is left.

Palabola:
Based on internal DoS protection:
                                                                          // Mysql use / World Server Load
case CMSG_QUERY_QUEST_INFO:                     //   0               1.5
case CMSG_QUERY_GAME_OBJECT:                    //   0               1.5
case CMSG_QUERY_CREATURE:                       //   0               1.5

case CMSG_QUEST_POI_QUERY:                      //   0              25         very high upload bandwidth usage

ariel-:
@zorix ByteBuffer's (which is base class for WorldPacket) default constructor reserves 1000 bytes on it's internal vector storage, so no 1 byte vector

Anyways i'm not a big fan of shrink_to_fit, it's non-binding function according to standards, and may not resize the vector (what can be done to force it is creating a temp vector copying the original one and then swapping them)

Shauren:
@ariel- only if you dont pass any size to the constructor - this PR does (and it passes 1)

zorix:
@ariel- Default capacity is 0x1000 -> 4096 https://github.com/TrinityCore/TrinityCore/blob/9666f7c8b1b563d64b57212f65dde8a255205267/src/server/shared/Packets/ByteBuffer.h#L64 so it's big when you need just 100

@Shauren In case of passing 1 to `Initialize` @ariel- is right because first of all `WorldPacket` is create with 4096 bytes buffer and then calling `Initialize` change this buffer BUT https://github.com/xinef1/TrinityCore/blob/9666f7c8b1b563d64b57212f65dde8a255205267/src/server/game/Server/WorldPacket.h#L69-L74 calls clear which doesn't change capacity (http://en.cppreference.com/w/cpp/container/vector/clear) and next calling reserve (http://en.cppreference.com/w/cpp/container/vector/reserve) which change only capacity in case when new capacity is greater so after everything capacity will be 4096 bytes.

Conclusion each packet will have at least 4096 bytes when some of them will probably never reach 200.

@ariel- About `shrink_to_fit` yes it may do nothing so capacity will remain same but in some cases maybe reduce memory usage.

Only what I want to say is that is something is generated at start will all the time use allocated memory should be reduce if possible because when you will count number of items, questes, etc. it's not small number when you have to multiply it by 4KiB.

Shauren:
@zorix no, you are wrong here. WorldPacket default constructor calls ByteBuffer reserving constructor with 0 size
https://github.com/xinef1/TrinityCore/blob/9666f7c8b1b563d64b57212f65dde8a255205267/src/server/game/Server/WorldPacket.h#L30

zorix:
Ye, you are right sorry.

So first version is correct and only need take a look at case when extending capacity wasting some memory (described above [256-129=127](https://github.com/TrinityCore/TrinityCore/pull/18637#issuecomment-269395181)) if you don't care about this ignore it.

Only what can be changed here is add casting for example in case like [this](https://github.com/TrinityCore/TrinityCore/pull/18637/files#diff-9e1cd0b614ef7e65b47b689ee52665f5R255) like all other packets doing it even when variable has proper type;

Tonghost:
Any news?

jackpoz:
how much memory does this use ?

xinef1:
From normal programming point of view - a lot (300MB?)
From normal server perspective, not so much.

A bit can be saved if we split this by locales related components, but again we want to improve performance so a bit of memory can be sacrificed

Ill add some more optimization prs not requiring additional memory, when I find more time to port them.

jackpoz:
300 MB sounds alright for a server, not sure tho for small ones with limited resources. a memory allocation profiling is always nice to have before merging these changes

xinef1:
Ill add config option, how about that?

DDuarte:
How does this behave with changing DB data and then .reload'ing it?

xinef1:
Ill fix reloading tomorrow, never really used those commands so i forgot about them

jackpoz:
very nice

joschiwald:
since we have packet classes in 3.3.5 i think it would be better to use them to make cherry pick to master branch easier

src/server/game/Entities/Item/ItemTemplate.cpp
```diff
@@ -122,3 +126,152 @@ uint32 ItemTemplate::GetSkill() const
...
+    queryTemp << Delay;
+    queryTemp << AmmoType;
+    queryTemp << RangedModRange;
+
+    for (int s = 0; s < MAX_ITEM_PROTO_SPELLS; ++s)
+    {
+        // send DBC data for cooldowns in same way as it used in Spell::SendSpellCooldown
+        // use `item_template` or if not set then only use spell cooldowns
+        SpellInfo const* spell = sSpellMgr->GetSpellInfo(Spells[s].SpellId);
+        if (spell)
```
Shauren: this shit is wrong and was wrong before you changed it too, only db data should ever be sent

xinef1:
@joschiwald I'll be out for the next week. It can be merged in current form and Ill adapt it sometime later if it's not a problem

PR #18638, opened 2016-12-27
Move some packets processing to map context

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles

Tested for a long time, should not create any new crashes (as far as I checked in current TC code)


Shauren:
I'm not really a fan of current PROCESS_THREADSAFE handling - processing it only in map update, it creates unneccessary early processing loop exits when you keep getting THREADSAFE, THREADUNSAFE, THREADSAFE, THREADUNSAFE sequences

xinef1:
Still, it's better to process as much as possible in maps than in single thread. If rest of the core is optimized well, you can get so low diff times that it doesn't matter at all.

Takenbacon:
With the current way opcodes are processed in TrinityCore at the moment, I would have to disagree with this commit. As a better than current, but not ideal solution, the lightweight threadsafe opcodes should be marked as PROCESS_INPLACE instead, it will have the benefit of _sometimes_ being executed in parallel, while not interrupting the packet processing loop.

jackpoz:
any helgrind log with the affected packets ?

xinef1:
Sadly, no. I'm not running a server anymore.
@Shauren CMSG_QUESTGIVER_ACCEPT_QUEST is currently thread unsafe, as far as i know party quest accept should not be sent to players on other map. Can you state correct behavior?

HelloKitty:
I see a lot of conversation about how to handle packets that have the potential to be done concurrently. Blizzard defines various locking policies that general act as supersets of the lower policies.

**They have two types of handling strategies**:

1. Simple single-threaded message queue pooling.

2. Concurrent message handling (with lock policies).

**Five main types of locking policies**:

1. Default lock* 
2. Handler lock
3. Protocol lock
4. Global service lock
5. Custom**

\* The default is a policy makes the promise that for a given client their JAM messages will never be handled concurrently or out of order.

\*\*They also have custom policies that can "inherit" the policy in the main types but also extend the idea with modern read/writing locking for better throughput. For example, for a given protocol lock you may only be read locking for some messages in a given protocol and only fully write lock on some operation codes within the protocol.

jackpoz:
Let's see what happens merging this PR

HelloKitty:
@Shauren It may have taken a long long time for me to finally appreciate your frustration with the behavior you described in the first post. The design truly is unfortunate. The design also impacts the ability to limit handling the amount of processing of non-PROCESS_THREADSAFE to spend less time per frame in WorldSession::Update in World::Update. Because it seems if you do, you'll potentially stall handling of PROCESS_THREADSAFE packets due to the early exists described.

What can be done about this though? It seems like a fundamental design change would need to occur for packet handling for players. Not easy, and out of the scope of my skills in C++. Has any public work been done on this?

What you posted too @Takenbacon is accurate. But doesn't marking so many handlers as PROCESS_INPLACE dramatically increase the time spend in World::Update's WorldSession:Update? The *sometimes* multithreaded processing of these packets seems, although I don't have data, to be less likely to happen. In your experience, does this not bottleneck things at high load? Maps in TrinityCore can't start updating until UpdateSessions exists for each session. So many PROCESS_INPLACE packets sent by players could stop Map updates from starting!

Overall, it's disappointing that there isn't anything I can do about the design and that even you guys are frustrated and dislike it too. What can we do?

HelloKitty:
After looking into it, it seems the only viable path if to resolve/manage race conditions with PROCESS_THREADUNSAFE on a case-by-case basis and bring them all to map processing on a case-by-case basis. It's the only way to avoid that early exit behavior from both entries of session updating imo.

Way easier said than done though.

jackpoz:
You could use grafana to get some additional stats

PR #18684, opened 2016-12-30
Few small optimizations here and there

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles, simple ingame tests

Review and tests required.

Changes list:
- Added CharacterGuidByNameContainer which contains name -> guid unordered map (updated along CharacterInfo)
- Extended CharacterInfo structure with GuildId
- Extended CharacterInfo structure with ArenaTeamId[3], for all possible teams (2v2, 3v3, 5v5)
- Removed CHAR_SEL_GUID_BY_NAME and CHAR_SEL_CHAR_GUID_BY_NAME synchronous queries, name -> guid can be now retrieved in World::GetCharacterGuidByName
- Removed CHAR_SEL_GUID_RACE_ACC_BY_NAME synchronous query, guid can be retrieved by name and rest of the data can be retrieved by guid
- Removed CHAR_SEL_CHAR_LEVEL synchronous query, level can be retrieved by guid
- Changed CHAR_SEL_CHARACTER_ACTIONS_SPEC to asynchronous query, action bars are now loaded asynchronously using _specSwitchActionsCallback
- Removed CHAR_SEL_CHARACTER_NAME_CLASS synchronous query, guid can be retrieved by name and rest of the data can be retrieved by guid
- Removed CHAR_SEL_PLAYER_ARENA_TEAMS and CHAR_SEL_ARENA_TEAM_ID_BY_PLAYER_GUID synchronous queries, arena teams are now stored in CharacterInfo
- Replaced synchronous db calls with CharacterInfo lookups
- Removed ObjectMgr::GetPlayerGUIDByName, as it used db query
- Replaced some unnecessary UpdateObjectVisibility() calls because they were either duplicated (called few lines above in other function) or it is enough to call DestroyForNearbyPlayers because object is being removed or should be invisible and DestroyForNearbyPlayers is faster
- Corrected typo in Player::DestroyForPlayer, only items in slots 0 to EQUIPMENT_SLOT_END are sent to other players
- Renamed Player::GetGuildIdFromDB to Player::GetGuildIdFromCharacterInfo and changed the function to use CharacterInfo structure
- Renamed Player::GetArenaTeamIdFromDB to Player::GetArenaTeamIdFromCharacterInfo and changed the function to use CharacterInfo structure
- Renamed Player::GetLevelFromDB to Player::GetLevelFromCharacterInfo and changed the function to use CharacterInfo structure
- Removed GameEventMgr::_questToEventLinks and associated functions, eventId is now stored in Quest class under _eventIdForQuest variable
- Changed some functions checking quest status to use other functions for quest status check instead of duplicating code
- Removed _addFriendCallback callback, because we can get the guid from appropriate storage, no need to make db query
- Removed _addIgnoreCallback callback, because we can get the guid from appropriate storage, no need to make db query
- Added _openWrappedItemCallback to unwrap wrapped items asynchronously
- Removed synchronous select in tutorials to check if we have any entries in db, if any entry exists in db, m_TutorialsChanged variable will have TUTORIALS_FLAG_LOADED_FROM_DB flag added and it is no longer necessary to query database
- Pushed missing change


Keader:
Better description would be nice :)

joschiwald:
even better, 1 pr for each fix

xinef1:
@joschiwald maybe i could split this into 3 pr but i decided that this is unnecessary. It's really hard to create separate pr when the changes interlace. Ill edit pr with better description.

src/server/game/Handlers/SpellHandler.cpp
```diff
@@ -231,34 +238,53 @@ void WorldSession::HandleOpenItemOpcode(WorldPacket& recvPacket)
 
         stmt->setUInt32(0, item->GetGUID().GetCounter());
 
+        _openWrappedItemCallback.SetFirstParam(item->GetPos());
+        _openWrappedItemCallback.SetSecondParam(item->GetGUID());
+        _openWrappedItemCallback.SetFutureResult(CharacterDatabase.AsyncQuery(stmt));
+
         PreparedQueryResult result = CharacterDatabase.Query(stmt);
```
Sarratas: .

src/server/game/Handlers/NPCHandler.cpp
```diff
@@ -391,12 +391,12 @@ void WorldSession::SendSpiritResurrect()
         if (corpseGrave != ghostGrave)
             _player->TeleportTo(corpseGrave->map_id, corpseGrave->x, corpseGrave->y, corpseGrave->z, _player->GetOrientation());
         // or update at original position
-        else
-            _player->UpdateObjectVisibility();
+        //else
+        //    _player->UpdateObjectVisibility(); called in ResurrectPlayer
```
ccrs: If you are sure 100% that this is not needed, just erase it.
If not, trace the code and debug if necessary till you are 100% sure that this is needed or not.

src/server/game/Maps/Map.cpp
```diff
@@ -879,10 +879,15 @@ void Map::RemovePlayerFromMap(Player* player, bool remove)
 {
     sScriptMgr->OnPlayerLeaveMap(this, player);
 
+    player->getHostileRefManager().deleteReferences(); // multithreading crashfix
```
ccrs: Kinda weird that only this is cleaned.
What about the references on the attacker, and the attack state and all the other related things?

xinef1: Called in ResurrectPlayer so absolutely not needed. Will apply all suggestions if someone decides it's worth merging.

src/server/game/Handlers/NPCHandler.cpp
xinef1: It is called in ResurrectPlayer so absolutely not needed, ill apply all suggestions if someone decides it's worth merging.

src/server/game/Maps/Map.cpp
xinef1: References are two sided, if unit clears it's hostile reference, hostile unit removes it's threat reference. The rest is for the ai to handle.

src/server/game/World/World.cpp
```diff
@@ -3293,6 +3289,14 @@ CharacterInfo const* World::GetCharacterInfo(ObjectGuid const& guid) const
     return nullptr;
 }
 
+ObjectGuid World::GetCharacterGuidByName(std::string const& name) const
```
ccrs: Why did you choose World as the holder of this container?
IIRC everything is stored in ObjectMgr/ObjectAccessor

xinef1: Because rest of CharacterInfo is stored in world, if this should be moved to ObjectMgr then CharacterInfo should be as well.

ccrs:
Also, seems like there are a couple of conflicts.

The problem with the PR is that there are changes not related between them.
I agree its a nightmare, and I even had some argues about my PRs, but at least those had one topic in common.

src/server/game/World/World.cpp
ccrs: Since this is not using CharacterInfo, you can freely move it to its correct place, right?
Leave CharacterInfo for another PR.

xinef1: ObjectAccessor is used for accessing ingame units (players and other entities). ObjectMgr is used to manage / store world database. The purpose of this container is to store name -> guid for all players (even those offline) and I assume this the initial reason why CharacterInfo was placed in World (which also bad place for this). The best solution would be to create a separate manager to handle the cached db data.

Shauren: KILL OBJECTMGR WITH FIRE.

xinef1: I was using grenades mostly, fire leaves ashes :D

frozen4:
awesome

src/server/game/Server/WorldSession.cpp
```diff
@@ -1178,6 +1162,24 @@ void WorldSession::ProcessQueryCallbacks()
         HandleStableSwapPetCallback(result, param);
         _stableSwapCallback.FreeResult();
     }
+
+    //- HandleOpenItemOpcode
+    if (_openWrappedItemCallback.IsReady())
+    {
+        uint16 pos = _openWrappedItemCallback.GetFirstParam();
```
Shauren: note to myself: need to write nicer api for async query callbacks

src/server/game/Maps/Map.cpp
```diff
@@ -879,10 +879,15 @@ void Map::RemovePlayerFromMap(Player* player, bool remove)
...
+
+    bool inWorld = player->IsInWorld();
     player->RemoveFromWorld();
     SendRemoveTransports(player);
 
-    player->UpdateObjectVisibility(true);
+    if (!inWorld) // if was in world, RemoveFromWorld() called DestroyForNearbyPlayers()
+        player->DestroyForNearbyPlayers(); // previous player->UpdateObjectVisibility()
```
Shauren: doesnt this leave garbage data in visible client guids set?

Keader:

Shauren:
@Keader not really a huge problem, github now has a "resolve conflicts" button

PR #18723, opened 2017-01-02
Core/Player: Add config option for IsGroupVisibleFor()

Staleness89:
**Changes proposed:**

-  Add a new configurable option for IsGroupVisibleFor() in worldserver.conf for stealthed players to remain invisible, even while in the same group. Because why not?

- [X] 3.3.5
- [?] master

None

Builds and works, looks fine.

None

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -2626,6 +2626,7 @@ bool Player::IsGroupVisibleFor(Player const* p) const
         default: return IsInSameGroupWith(p);
         case 1:  return IsInSameRaidWith(p);
         case 2:  return GetTeam() == p->GetTeam();
+        case 3:  return nullptr;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -2626,6 +2626,7 @@ bool Player::IsGroupVisibleFor(Player const* p) const
         default: return IsInSameGroupWith(p);
         case 1:  return IsInSameRaidWith(p);
         case 2:  return GetTeam() == p->GetTeam();
+        case 3:  return nullptr;
```
DDuarte: Shouldn't it be false instead?

Staleness89: this also worked, but that makes more sense. I'll change that and test it again

DDuarte: It doesn't even compile (with a properly configured compiler): https://travis-ci.org/TrinityCore/TrinityCore/builds/188351642#L1047 :)

Staleness89: like I said it compiled and worked fine for me, tested it before I opened this PR, but now it's changed and it still works.

Warpten:
because that was blizzlike

Shauren:
A config? Blizzlike? Only the default value (which is unchanged)

PR #18733, opened 2017-01-04
Core/Creature: Implement groupAI 3 and 4 (formation without following)

Lumber-Jack-2:

**Changes proposed:**

-  Update https://trinitycore.atlassian.net/wiki/display/tc/creature_formations#creature_formations-groupAI
-  Currently all members of the group follow the leader if groupAI 0,1 or 2 is set in creature_formations
-  Add groupAI 3 and 4.
- groupAI 3: Group members will assist all group members and don't follow each other
- groupAI 4: Group member will assist only their leader if attacked and don't follow each other

- [X] 3.3.5
- [X] master

yes and yes

Example:

```sql
DELETE FROM `creature_formations` WHERE `leaderGUID`=48704;
INSERT INTO `creature_formations` (`leaderGUID`, `memberGUID`, `dist`, `angle`, `groupAI`, `point_1`, `point_2`) VALUES 
(48704, 48704, 0, 0, 3, 0, 0),
(48704, 48703, 0, 0, 3, 0, 0);
```

It would be useful for these two creatures in Ragefire Chasm. Creature 48704 and 48703 should attack together if one of the was attacked, but 48703 should not follow 48704, because 48704 has rndmmovement and 48703 should simply stay on it's spawn position the whole time.

### DIFF
--- a/src/server/game/Entities/Creature/CreatureGroups.cpp
+++ b/src/server/game/Entities/Creature/CreatureGroups.cpp
@@ -176,7 +176,7 @@ void CreatureGroup::MemberAttackStart(Creature* member, Unit* target)
     if (!groupAI)
         return;
 
-    if ((groupAI == FLAG_AGGRO_ON_AGGRO || groupAI == FLAG_AGGRO_ON_AGGRO_NOFOLLOW) && member != m_leader)
+    if ((groupAI == GROUP_AI_ASSISTLEADER_FOLLOW || groupAI == GROUP_AI_ASSISTLEADER_NOFOLLOW) && member != m_leader)
         return;
 
     for (CreatureGroupMemberType::iterator itr = m_members.begin(); itr != m_members.end(); ++itr)
@@ -233,7 +233,7 @@ void CreatureGroup::LeaderMoveTo(float x, float y, float z)
             continue;
 
         uint8 groupAI = sFormationMgr->CreatureGroupMap[member->GetSpawnId()]->groupAI;
-        if (groupAI == FLAG_TO_AGGRO_ON_AGGRO_NOFOLLOW || groupAI == FLAG_AGGRO_ON_AGGRO_NOFOLLOW)
+        if (groupAI == GROUP_AI_ASSISTGROUP_NOFOLLOW || groupAI == GROUP_AI_ASSISTLEADER_NOFOLLOW)
             return;
 
         if (itr->second->point_1)
--- a/src/server/game/Entities/Creature/CreatureGroups.cpp
+++ b/src/server/game/Entities/Creature/CreatureGroups.cpp
@@ -176,7 +176,7 @@ void CreatureGroup::MemberAttackStart(Creature* member, Unit* target)
     if (!groupAI)
         return;
 
-    if ((groupAI == 1 || groupAI == 4) && member != m_leader)
+    if ((groupAI == GROUP_AI_ASSISTLEADER_FOLLOW || groupAI == GROUP_AI_ASSISTLEADER_NOFOLLOW) && member != m_leader)
         return;
 
     for (CreatureGroupMemberType::iterator itr = m_members.begin(); itr != m_members.end(); ++itr)
@@ -233,7 +233,7 @@ void CreatureGroup::LeaderMoveTo(float x, float y, float z)
             continue;
 
         uint8 groupAI = sFormationMgr->CreatureGroupMap[member->GetSpawnId()]->groupAI;
-        if (groupAI == 3 || groupAI == 4)
+        if (groupAI == GROUP_AI_ASSISTGROUP_NOFOLLOW || groupAI == GROUP_AI_ASSISTLEADER_NOFOLLOW)
             return;
 
         if (itr->second->point_1)
--- a/src/server/game/Entities/Creature/CreatureGroups.cpp
+++ b/src/server/game/Entities/Creature/CreatureGroups.cpp
@@ -176,7 +176,7 @@ void CreatureGroup::MemberAttackStart(Creature* member, Unit* target)
     if (!groupAI)
         return;
 
-    if (groupAI == 1 && member != m_leader)
+    if ((groupAI == 1 || groupAI == 4) && member != m_leader)
         return;
 
     for (CreatureGroupMemberType::iterator itr = m_members.begin(); itr != m_members.end(); ++itr)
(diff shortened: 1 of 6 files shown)


xinef1:
I would change the AI 0/1/2/3/4 to flags which maps specific behavior (follow leader, assist leader etc)

Shauren:
eww magic numbers

Rushor:
magic number was already there before :( 

how should be call these flags:

0 = GROUP_AI_..what?..
1 = GROUP_AI_..what?..

r00ty-tc:
My suggestion is something like:

GROUP_AI_NOASSIST_FOLLOW (0)
GROUP_AI_ASSISTLEADER_FOLLOW (1)
GROUP_AI_ASSISTGROUP_FOLLOW (2)
GROUP_AI_ASSISTGROUP_NOFOLLOW (3)
GROUP_AI_ASSISTLEADER_NOFOLLOW (4)

Mainly, just needs to be descriptive though.

Shauren:
flags pls

r00ty-tc:
@Shauren I agree flags would be nicer, but wouldn't that require changing all the current logic in SmartAI etc? It changes a search replace into changing the if/etc everywhere to match the flags instead?

Rushor:
magicnumber killed ty. mangos has way to many flags http://docs.getmangos.com/en/latest/database/world/creature-linking.html. i don't know how to implemement them all. For now, those 2 additional ai-types are enough (to my mind)

r00ty-tc:
Looking at those flags.. I am sure some of them are surely NEVER used.

- the slave respawn when the master dies
- the slave despawns on master respawn

Anyone seen those used in retail?

Rushor:
> the slave respawn when the master dies

i saw this behaviour in one of the deadmines commits https://github.com/cmangos/wotlk-db/commit/432c11dd0449f7ef50e1e9e69cb616bd5df5f218#diff-4a8aac6a535d0b13c74f517f42a2b9dcR96

make ask X-Savior there

Shauren:
Oh I wasn't asking you to implement them all, just look at them and only support the ones we already support with groupAI column (keeping cmangos values)

Rushor:
well then FLAG_AGGRO_ON_AGGRO and FLAG_TO_AGGRO_ON_AGGRO seems to be the same like groupai 1 and 2 that is currently used on tc. However creature_linking is a combiantion of creature_formation and linked_respawn without followangle, we cannot simply copy those flags. (And btw, there are not that much creature_linking entries that need to be ported to tc anymore. most of them are all handled by groupai 2 already)

Shauren:
But you can copy the ones referring to aggro, no problem here

xinef1:
Flag is a flag :D use unique bits for each flag
0x1, 0x2, 0x4, 0x8, 0x10 etc and use bitwise comparison

Shauren:
Now you changed values and need SQL to convert existing db rows
Also operator== is not how you check for flags

Rushor:
but if we change it to a flagcheck now like (groupAI & FLAG_TO_AGGRO_ON_AGGRO), FLAG_TO_AGGRO_ON_AGGRO (which is 0x00000002) will also contain FLAG_AGGRO_ON_AGGRO(which is 0x00000001) in the check no? the current checks will not work anymore? cannot we stay simply at the groupai sytem with the definition of GROUP_AI_ASSISTLEADER_FOLLOW etc. ?

Rushor:
sry i cannot change more in this pr

Rushor:
alright thx. But then again, it doesn't even make sence to copy those flags because it's totally different implemented (i understood creature_linking wrong in the beginning)

mangos:
1 = FLAG_AGGRO_ON_AGGRO = member attacks if leader attacks
2 = FLAG_TO_AGGRO_ON_AGGRO = leader attacks if member attacks

but then again our definition of groupai is simply different
1 = member attacks if leader attacks
2 = all member help each other

so if we want to copy those flags we frist have to implement FLAG_AGGRO_ON_AGGRO and FLAG_TO_AGGRO_ON_AGGRO like it is in mangos and all current groupais with number 2 have to be changed to 3 (bitwise it contains FLAG_AGGRO_ON_AGGRO and FLAG_TO_AGGRO_ON_AGGRO).

But then again i don't see the reason why we should do that

Rushor:
Alright i updated it to the flagsystem now.

flag 0 = If any creature from group is attacked, members won't assist and won't follow
flag 1 = The member aggroes if the leader aggroes
flag 2 = The leader aggroes if the member aggroes
flag 4 = The member will follow the leader

This means that:
- current flag 2 will be flag 7 in the future.
- current flag 1 will be flag 5 in the future.
- current flag 0 will be flag 4 in the future.

This also enables the feature, that creatures will assist each other, but don't follow each other (flag 1+2)

Rushor:
some more detailed info how it works now:

flags: function
(0)       = Noone assists noone and member don't follow the leader
(4)       = Noone assists noone and member follow the leader
(1)       = The member aggroes if the leader aggroes
(2)       = The leader aggroes if the member aggroes
(1|2)    = Everyone assists everyone and member don't follow the leader
(1|2|4) = Everyone assists everyone and member follow the leader

src/server/game/Entities/Creature/CreatureGroups.cpp
```diff
@@ -196,7 +193,7 @@ void CreatureGroup::MemberAttackStart(Creature* member, Unit* target)
         if (other->GetVictim())
             continue;
 
-        if (other->IsValidAttackTarget(target))
+        if ((other != m_leader && groupAI & FLAG_AGGRO_ON_AGGRO) || (other == m_leader && groupAI & FLAG_TO_AGGRO_ON_AGGRO) && other->IsValidAttackTarget(target))
```
ForesterDev: if (((other != m_leader && groupAI & FLAG_AGGRO_ON_AGGRO) || (other == m_leader && groupAI & FLAG_TO_AGGRO_ON_AGGRO)) && other->IsValidAttackTarget(target))

I think this will be correct

Rushor: will not work and everyone attacks everyone even with other flags

Aokromes:
TrinityCore/TrinityCore/src/server/game/Entities/Creature/CreatureGroups.cpp:196:125: fatal error: 
      '&&' within '||' [-Wlogical-op-parentheses]
  ...|| (other == m_leader && groupAI & FLAG_TO_AGGRO_ON_AGGRO) && other->IsValidAttackTarget(target))
     ~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Entities/Creature/CreatureGroups.cpp:196:125: note: 
      place parentheses around the '&&' expression to silence this warning
  ...(other == m_leader && groupAI & FLAG_TO_AGGRO_ON_AGGRO) && other->IsValidAttackTarget(target))
                                                             ^
     (                                                                                            )
1 error generated.

Killyana:
Someone can check this PR and merge it?

Aokromes:
If no oposition i will merge this tomorrow.

Shauren:
No, you will not. I told you many times to stop posting that.

Aokromes:
and this is why you don't have a single db guy working on master.

RAlexis:
Did you just... merge it!?

Treeston:
Yes, that's why it says "@Treeston merged commit 0770848 into `TrinityCore:3.3.5`" 😃

Issue #18615, opened 2016-12-26
[3.3.5] NPC Steward of Time 


Tauriella:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:**
You can't fly with Steward of Time 
Comes only the message "is too far away"

![hdz](https://cloud.githubusercontent.com/assets/10637389/21483109/821d20a0-cb7d-11e6-8ff9-4f3b76cdeb06.jpg)

**Steps to reproduce the problem:**

1. .go 8165 -4773 35 1
2. Talk with Steward of Time
 

**Branch(es):** 3.3.5 / master (Tell us which branch(es) this issue affects.)

**TC rev. hash/commit:**  
TrinityCore rev. bf7e95bd5703 2016-12-26 14:25:32 +0100 (335 branch)

**TDB version:** 
TDB_full_world_335.62_2016_10_17.sql + updates 

**Operating system:**  
Win7 64Bit

Kittnz:
tried without gm on?

Tauriella:
GM is off, and i have testet with normal Player Char
Same result " i to far away "

Staleness89:
confirm, I tried giving the SAI spellcasts the triggered CastFlag but I guess it doesn't ignore range requirements. Weird though because it's an SMART_ACTION_INVOKER_CAST, shouldn't have a range problem.

Staleness89:
I looked in the Taxi dbcs, found the starting position of this taxi, moved the Steward closer to this position, and that fixes it. Could it be that the Steward is spawned in the wrong spot?

Aokromes:
it was working for years, and i don't remember any commit moving it, also another mob on dragonbright have the same issue.

Aokromes:
#15398

ariel-:
For this particular issue, I found in sniff creature casting first 45331 (serverside spell). After this player self casts 34891, still taxis activated through spells check against the dbc node position.

There is no way to associate the caster with the npc triggering the event, as it's the same player. Should it be better to actually drop the check entirely if activating taxi through a spell?

Tauriella:
@ariel- 
Maybe they look in the cmangos code as the have solved.
There it works.

Issue #15398, opened 2015-09-01
[3.3.5a][6.x] Gossip taxi not working from middle floor of Wyrmrest Temple


Treeston:
Branch 3.3.5a, revision 0cdf3b7a73d8, TDB 335.59

Both Taxi gossip options from Lord Afrasastrasz in the middle floor of Wyrmrest Temple print "You are too far away from the taxi stand" instead of transporting you to the ground/top of the temple, respectively.

Rexidia:
It seems to work for me, you just have to stand behind him. Try this to see if it works for you.

Killyana:
Yes if you stand behind the npc you can take the fly
The question is: the position of "taxi stand" is somewhere on the DB or taken directly from DBC?

Trisjdc:
It's distance from the first WP point, iirc. I don't think there should be even such a short distance check to begin with. It should be either be removed or massively increased. I recall a Coldarra quest with a taxi spell that occasionally breaks because of the same problem.

tkrokli:
Oh. You mean like in Quest 11969 "Springing the Trap", issue #5137 ?

tkrokli:
Confirmed that this issue still exists on commit cc685ac when making PR #15966 .

Staleness89:
Assuming this code in Player.cpp is related?
```C++
    // check node starting pos data set case if provided
    if (node->x != 0.0f || node->y != 0.0f || node->z != 0.0f)
    {
        if (node->map_id != GetMapId() || !IsInDist(node->x, node->y, node->z, 2 * INTERACTION_DISTANCE))
        {
            GetSession()->SendActivateTaxiReply(ERR_TAXITOOFARAWAY);
            return false;
        }
    }
```

PR #18774, opened 2017-01-09
Core/Player: increase ActivateTaxiPathTo() distance check

Staleness89:
**Changes proposed:**
Remove the distance check from ActiveTaxiPathTo()
*Should* Fix the failure to activate taxi paths for at least one quest (#18615)
Not sure if this is better than increasing the distance. Feedback please?

- [x] 3.3.5
- [x] master

#15398
#18615

None. Feedback please?

None

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -21113,7 +21113,7 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc 
         return false;
     }
 
-	// check node starting pos data set case if provided
+    // check node starting pos data set case if provided
     if (node->x != 0.0f || node->y != 0.0f || node->z != 0.0f)
     {
         if (node->map_id != GetMapId() || !IsInDist(node->x, node->y, node->z, 6 * INTERACTION_DISTANCE))
@@ -21122,7 +21122,6 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc 
             return false;
         }
     }
-
     // node must have pos if taxi master case (npc != NULL)
     else if (npc)
     {
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -21116,11 +21116,8 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc 
     // check node starting pos data set case if provided
     if (node->x != 0.0f || node->y != 0.0f || node->z != 0.0f)
     {
-        if (node->map_id != GetMapId() || !IsInDist(node->x, node->y, node->z, 2 * INTERACTION_DISTANCE))
-        {
-            GetSession()->SendActivateTaxiReply(ERR_TAXITOOFARAWAY);
-            return false;
-        }
+        GetSession()->SendActivateTaxiReply(ERR_TAXITOOFARAWAY);
+        return false;
     }
     // node must have pos if taxi master case (npc != NULL)
     else if (npc)
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -21113,14 +21113,8 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc 
         return false;
     }
 
-    // check node starting pos data set case if provided
-    if (node->x != 0.0f || node->y != 0.0f || node->z != 0.0f)
-    {
-        GetSession()->SendActivateTaxiReply(ERR_TAXITOOFARAWAY);
-        return false;
-    }
     // node must have pos if taxi master case (npc != NULL)
-    else if (npc)
+    if (npc)
     {
         GetSession()->SendActivateTaxiReply(ERR_TAXIUNSPECIFIEDSERVERERROR);
         return false;
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -21113,8 +21113,18 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc 
         return false;
     }
 
+	// check node starting pos data set case if provided
+    if (node->x != 0.0f || node->y != 0.0f || node->z != 0.0f)
+    {
+        if (node->map_id != GetMapId() || !IsInDist(node->x, node->y, node->z, 6 * INTERACTION_DISTANCE))
+        {
+            GetSession()->SendActivateTaxiReply(ERR_TAXITOOFARAWAY);
+            return false;
+        }
+    }
+
     // node must have pos if taxi master case (npc != NULL)
-    if (npc)
+    else if (npc)
     {
         GetSession()->SendActivateTaxiReply(ERR_TAXIUNSPECIFIEDSERVERERROR);
         return false;


ForesterDev:
I think, this changes are completely incorrect. Before this changes, error returns when some checks are fail, now error returns always, because you remove this checks.

Staleness89:
@ForesterDev Yes I was very tired at that moment, I can fix that.

jackpoz:
> Tests performed: (Does it build, tested in-game, etc.)
> None. Feedback please?

What do you mean with "none" ? No tests at all ?

Staleness89:
No tests at all, obviously as I did it wrong the first time. It's a pretty straight forward change, the question is does it break anything else related, and I can't think of anything. I can also change it to increase the distance rather than remove the check completely.

Aokromes:
for sure the check is to stop hacks.

Rushor:
4 * INTERACTION_DISTANE would not be enough here?

Staleness89:
@Aokromes Yes I thought the check might be an anti-hack. I can test at least one spot affected by this and find out if 4 * INTERACTION_DISTANCE is enough for it.

Shauren:
And now it will always fail with flight masters (npc != nullptr)

Staleness89:
4 * INTERACTION_DISTANCE is definitely not enough. It would have to be over 100 to fix the Steward of Time taxi. Maybe someone else could find a different way to measure it, or fix this another way. Closing for now, if no one else does anything I'll try again later after testing.

Aokromes:
Meh i checked on 4.3.4 db and it got diff coordinates, then compared with mangos, we have same coords, checked sniff, bad luck Steward of Time is on proper coords.

Staleness89:
Tested with 6 * INTERACTION_DISTANCE, fixes #18615.

Treeston:
feels like hack

src/server/game/Entities/Player/Player.cpp
```diff
@@ -21113,15 +21113,16 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc
...
     if (node->x != 0.0f || node->y != 0.0f || node->z != 0.0f)
     {
-        if (node->map_id != GetMapId() || !IsInDist(node->x, node->y, node->z, 2 * INTERACTION_DISTANCE))
+        if (node->map_id != GetMapId() || !IsInDist(node->x, node->y, node->z, 6 * INTERACTION_DISTANCE))
         {
             GetSession()->SendActivateTaxiReply(ERR_TAXITOOFARAWAY);
             return false;
         }
     }
+
```
jackpoz: it is good habit to review all changes pushed, trying to minimize the codestyle changes

jackpoz:
please include the original commit that set the value to 2 in your PR

src/server/game/Entities/Player/Player.cpp
```diff
@@ -21113,15 +21113,16 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc
         return false;
     }
 
-    // check node starting pos data set case if provided
+	// check node starting pos data set case if provided
```
tkrokli: Please remove this change, it is unnecessary. TC source does not use TABs.
GitHub will use 1 TAB = 8 spaces in preview. Correct indent is 4 spaces.

Treeston:
Why do we even check distance from taxi node here instead of distance from npc (if present)?

Staleness89:
@jackpoz - 2 was the original value
@Treeston how would you find the npc's position?

Treeston:
npc is passed to ActivateTaxiPathTo

Staleness89:
I tried this:
```C++
		if (node->map_id != GetMapId() || !IsInDist(npc->GetPositionX(), npc->GetPositionY(), npc->GetPositionZ(), 2 * INTERACTION_DISTANCE))
```
which causes a server crash. Any idea on how to fix that?

Treeston:
npc can be null

Staleness89:
@Treeston I need a better hint.

tkrokli:
if you are doing this within
```cpp
bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc /*= nullptr*/, uint32 spellid /*= 0*/)
```
, you check that `npc` is not null, either `if (npc)` or `if (!npc)` \ `return false;` and go from there.

Treeston:
`npc == null` should still be handled (same check as current).

Please don't give advice if you have no idea what you are talking about.

Eliminationzx:
I guess it should be like this
```
// check node starting pos data set case if provided
    if (node->x != 0.0f || node->y != 0.0f || node->z != 0.0f)
    {
        if (node->map_id != GetMapId() || npc && !IsInDist(npc->GetPositionX(), npc->GetPositionY(), npc->GetPositionZ(), 2 * INTERACTION_DISTANCE))
        {
            GetSession()->SendActivateTaxiReply(ERR_TAXITOOFARAWAY);
            return false;
        }
    }
    // node must have pos if taxi master case (npc != NULL)
    else
    {
        if (npc)
        {
            GetSession()->SendActivateTaxiReply(ERR_TAXIUNSPECIFIEDSERVERERROR);
            return false;
        }
    }
```

Staleness89:
@Eliminationzx That fixed the crash, but now comes with the out of range error again (Still testing on Steward of Time) I believe it's because the npc doesn't actually start the taxi, the SAI involved makes the player self-cast the taxi start spell, so there's really no NPC involved in that calculation.

Eliminationzx:
@Staleness89, post updated.

Staleness89:
@Eliminationzx still doesn't seem to work.

ariel-:
> it was working for years, and i don't remember any commit moving it, also another mob on dragonbright have the same issue.

Would be better to actually debug the root cause for the problem instead of touching code that isn't related to the problem being there in first place.

ariel-:
I retract. Maximum interaction distance tolerance is what caused the issue in the first place:

Commit 875844845f77f85b20cd8f908adb4fde61ab69ae changed maximum distance from (2 * INTERACTION_DISTANCE)³ ie 1000 yds to (2 * INTERACTION_DISTANCE) ie 10 yds

Treeston:
8758448 was an intentional change. Old limit was practically non-existant, and open to hacks.

Staleness89:
so.. where should it be fixed if not here?

ariel-:
I know, but it's too restrictive

Treeston:
It isn't. The issue here (as discussed in #15398) is that the core compares against the wrong position.

ariel-:
I miscalculated, squared distances compared against squared constant, so max distance was sqrt(1000) instead 1000 ie 31.62 yds. Updated my previous comment

Still a typo, though.

Treeston:
No. Proper distance check would be `sqrt(2^2+2^2+2^2)`, or `3.46`*`INTERACTION_DISTANCE`, if you _really_ cared.

Realistically you'll never be off in more than two directions so it'd be more of `2.82`*`INTERACTION_DISTANCE`.

ariel-:
INTERACTION_DISTANCE is 5 yds, those are multiplications, (2*5)*(2*5)*(2*5) = 10*10*10 = 1000. Compared squared distance against a squared distance ie sqrt(1000).

I don't know how much leeway allows the client, if any.

Eliminationzx:
Can we do something like this?
```diff
diff --git a/src/server/game/Entities/Player/Player.cpp b/src/server/game/Entities/Player/Player.cpp
index 65e7ec8..4d3206d 100644
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -21116,7 +21116,9 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc
     // check node starting pos data set case if provided
     if (node->x != 0.0f || node->y != 0.0f || node->z != 0.0f)
     {
-        if (node->map_id != GetMapId() || !IsInDist(node->x, node->y, node->z, 2 * INTERACTION_DISTANCE))
+        bool isInValidDistance = GetExactDistSq(node->x, node->y, node->z) < pow(2 * INTERACTION_DISTANCE, 3);
+
+        if (node->map_id != GetMapId() || !isInValidDistance)
```

Staleness89:
> There is no way to associate the caster with the npc triggering the event, as it's the same player. Should it be better to actually drop the check entirely if activating taxi through a spell?

PR #18815, opened 2017-01-13
Core/Object: fixed traps visibility for enemies. Fixes #18251

### LINKED ISSUE #18251: [3.3.5] Spell/Hunter: Traps are always visible to enemies

**Description:**
The hunter traps are always visible and should not, this makes it very easy to avoid the traps. Enemies can only see traps up to 10 yards away, in this case it is very likely that the enemy will fall into the trap. Only rogues with Detect Traps can see traps easily.

**Current behaviour:** 
Hunter's traps are always visible.

**Expected behaviour:** 
Enemies can only see traps 10 yards or less away.

1. With a hunter use any trap.
2. Any enemy can see the trap from afar.

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  674ac56

**TDB version:**  TDB_full_world_335.62_2016_10_17 + updates

**Operating system:**  Windows

ForesterDev:
**Changes proposed:**
-  fixed typo that made traps are always visible for enemies. 

**Target branch(es):** 3.3.5

**Issues addressed:** Closes #18251  

**Tests performed:** (Does it build, tested in-game, etc.) Tested, works

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1715,6 +1715,7 @@ bool WorldObject::CanDetectStealthOf(WorldObject const* obj, bool checkAlert) co
     if (!HasInArc(float(M_PI), obj))
         return false;
 
+    GameObject const* go = obj->ToGameObject();
     for (uint32 i = 0; i < TOTAL_STEALTH_TYPES; ++i)
     {
         if (!(obj->m_stealth.GetFlags() & (1 << i)))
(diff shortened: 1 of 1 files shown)


jackpoz:
any info about which commit added the bug in the first place ?

ForesterDev:
@jackpoz , this changes was made in https://github.com/ForesterDev/TrinityCore/commit/7c36e3a298ce243631848b94364c185a863d1526

jackpoz:
thanks

Shauren:
Then this kind of reverts that part of the commit, what if you just kept the code as it was (with local var) just change its value?

ForesterDev:
@Shauren, for why we should leave this local variable, if it uses only once in this function?

Shauren:
Its used in a loop

ForesterDev:
@Shauren thanks for explanation

PR #18848, opened 2017-01-15
Core/Spell: SPELL_FAILED_TRY_AGAIN should be send after cast is finished

robinsch:
**Changes proposed:**

-  Moved SPELL_FAILED_TRY_AGAIN to effects

- [x] 3.3.5
- [ ] master

-


Shauren:
No. Do not abort spells in effects, that is bad design (for example an earlier effect could already have been executed)

PR #18857, opened 2017-01-15
Changed error message when loading outdated Vmap file

### LINKED ISSUE #18431: mmaps not finding tree file

**Description:**
mmaps_generator doesn't run correctly against vmaps directory made on master

**Current behaviour:** (Tell us what happens.)
When run it states that vmaps directory is not present or is empty

**Expected behaviour:** (Tell us what should happen instead.)
vmaps directory is present and was populated by vmap4_extractor/assembler of same commit hash
tool should generate the mmaps

1. follow instructions to extract map data for linux, with tools built from current master
2. get to step involving mmaps_generator
3. cry

Confirmed resolution by separating the error print statement out. The fault believed that the directory was present but empty. The current vmaps tools generate a .dtree file, and the mmaps_generator tool is searching specifically for a *.vmtree file. Changing the filter to *.dtree allows it to run correctly.

**Branch(es):** 3.3.5 / master (Tell us which branch(es) this issue affects.)
master

**TC rev. hash/commit:**  

**TDB version:**  

**Operating system:**  
ubuntu 16.04

devtekve:
Added a more descriptive message in case the Vmap file fails to load due
being outdated or if it is infact on in the folder.

**Changes proposed:**

-  When a Vmap file cant be loaded because of incompatibility, it should say it, instead of saying that the file does not exist, otherwise you may think it is a filesystem /permission problems.

- [x] 3.3.5

**Issues addressed:** closes #18431

Builds, no in-game modifications done. Works as expected.

***Function names should be corrected:***
- [Game->Maps->Map.cpp, line 105] Function should be renamed from ExistVMap to CheckVMap or something else that implies that it not only checks if file exist but also if it can be loaded! 
- [Game->Maps->Map.cpp, line 121] This method should be renamed to 'canLoadMap'. it is not clear when it couldn't load because didn't exist or couldn't load because of file mismatch!

### DIFF
--- a/src/common/Collision/Management/IVMapManager.h
+++ b/src/common/Collision/Management/IVMapManager.h
@@ -1,20 +1,20 @@
 /*
-* Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
-* Copyright (C) 2005-2010 MaNGOS <http://getmangos.com/>
-*
-* This program is free software; you can redistribute it and/or modify it
-* under the terms of the GNU General Public License as published by the
-* Free Software Foundation; either version 2 of the License, or (at your
-* option) any later version.
-*
-* This program is distributed in the hope that it will be useful, but WITHOUT
-* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
-* FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
-* more details.
-*
-* You should have received a copy of the GNU General Public License along
-* with this program. If not, see <http://www.gnu.org/licenses/>.
-*/
+ * Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
+ * Copyright (C) 2005-2010 MaNGOS <http://getmangos.com/>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
 
 #ifndef _IVMAPMANAGER_H
 #define _IVMAPMANAGER_H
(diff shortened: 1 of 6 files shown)


Keader:
Change Tabs to 4 Spaces.

devtekve:
@Keader Done, would you main giving it a second check? I hope I did it right, as you can see, I am new with this stuffs.

src/server/game/Maps/Map.cpp
```diff
@@ -102,23 +102,32 @@ bool Map::ExistMap(uint32 mapid, int gx, int gy)
...
+                return false;
+            }
+            
+            
+            bool couldLoad = vmgr->existsMap((sWorld->GetDataPath() + "vmaps").c_str(), mapid, gx, gy); //TODO: This method should be renamed to 'canLoadMap' because, it is not clear when it couldn't load because didn't exist or couldn't load because of file mismatch!
+            if (!couldLoad)
+            {
+                std::string name = vmgr->getDirFileName(mapid, gx, gy);
+                TC_LOG_ERROR("maps", "VMap file '%s' couldn't be loaded", (sWorld->GetDataPath() + "vmaps/" + name).c_str());
+                TC_LOG_ERROR("maps", "This may be due a mismatch version, lack of permissions or a corrupted file. Try re-extracting the maps again.");
```
Aokromes: imho add version check separate to lack of permissions or corrupted files.

Shauren:
Perhaps instead of creating multiple methods that check different things it would be better to return error code and select proper message based on that instead of current bool return

src/server/game/Maps/Map.cpp
tkrokli: If the error message is kept as is in the final stage of this PR, please consider rewriting line 2 to something like this:
```
[...] "This might be due to a version mismatch, [...]
[...] "This could be caused by a mismatched version, [...]
```

src/common/Collision/Management/VMapManager2.cpp
```diff
@@ -326,6 +326,20 @@ namespace VMAP
         return StaticMapTree::CanLoadMap(std::string(basePath), mapId, x, y);
     }
 
+	bool VMapManager2::isPathAccessibleForMap(const char* _basePath, unsigned int mapID)
+	{
+        std::string basePath = std::string(_basePath);
```
Keader: and missing identation inside of this function

src/server/game/Maps/Map.cpp
```diff
@@ -102,23 +102,32 @@ bool Map::ExistMap(uint32 mapid, int gx, int gy)
...
+            bool exists = vmgr->isPathAccessibleForMap((sWorld->GetDataPath() + "vmaps").c_str(), mapid);
             if (!exists)
             {
                 std::string name = vmgr->getDirFileName(mapid, gx, gy);
-                TC_LOG_ERROR("maps", "VMap file '%s' does not exist", (sWorld->GetDataPath()+"vmaps/"+name).c_str());
-                TC_LOG_ERROR("maps", "Please place VMAP-files (*.vmtree and *.vmtile) in the vmap-directory (%s), or correct the DataDir setting in your worldserver.conf file.", (sWorld->GetDataPath()+"vmaps/").c_str());
+                TC_LOG_ERROR("maps", "VMap file '%s' does not exist", (sWorld->GetDataPath() + "vmaps/" + name).c_str());
+                TC_LOG_ERROR("maps", "Please place VMAP-files (*.vmtree and *.vmtile) in the vmap-directory (%s), or correct the DataDir setting in your worldserver.conf file.", (sWorld->GetDataPath() + "vmaps/").c_str());
+                return false;
+            }
```
joschiwald: bad design: open close open file
better: handle the log message inside existsMap or let it return a load result instead of bool
similar loading logic in vmap and mmap would be great

src/server/game/Maps/Map.cpp
devtekve: @Aokromes Is the magic number implemented on VMAP files already?

Aokromes: https://github.com/TrinityCore/TrinityCore/commit/ec9a8345a49938ddd9d42e9fdf7053ac25477b4d

devtekve:
Ok guys, I did again this code, using instead of just true or false, I am using different levels of error, so we can now if it is a Version mismatch, a file not found or something else. ( @Aokromes ) 

Also deleted my methods and used the ones that were exisiting already, there as no need to make new methods anyways, thanks for your tip @Shauren 

Should I do a totally new pull request? 

( @Keader )

src/server/game/Maps/Map.cpp
```diff
@@ -102,24 +102,36 @@ bool Map::ExistMap(uint32 mapid, int gx, int gy)
...
+			std::string name = vmgr->getDirFileName(mapid, gx, gy);
+			switch (result)
+			{
+			case 1:
+				TC_LOG_ERROR("maps", "VMap file '%s' does not exist", (sWorld->GetDataPath() + "vmaps/" + name).c_str());
+				TC_LOG_ERROR("maps", "Please place VMAP-files (*.vmtree and *.vmtile) in the vmap-directory (%s), or correct the DataDir setting in your worldserver.conf file.", (sWorld->GetDataPath() + "vmaps/").c_str());
+				return false;
+
+			case 2:
+				TC_LOG_ERROR("maps", "VMap file '%s' couldn't be loaded", (sWorld->GetDataPath() + "vmaps/" + name).c_str());
```
Aokromes: i suggest you to copy the version error from maps.

devtekve:
@Aokromes I really don't find a proper way of showing File Version and Expected Version because all of that check is done in StaticMapTree::CanLoadMap, it would require either moving the whole file check to Map.cpp or instead of passing an int, passing an array with result code and error message, which I do not recomment, is there any suggestion to do it?

src/common/Collision/Management/IVMapManager.h
```diff
@@ -1,20 +1,20 @@
...
- *
- * This program is distributed in the hope that it will be useful, but WITHOUT
- * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
- * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
- * more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program. If not, see <http://www.gnu.org/licenses/>.
- */
+* Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
```
Aokromes: don't change this.

src/common/Collision/Management/VMapManager2.cpp
```diff
@@ -1,20 +1,20 @@
...
- *
- * This program is distributed in the hope that it will be useful, but WITHOUT
- * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
- * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
- * more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program. If not, see <http://www.gnu.org/licenses/>.
- */
+* Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
```
Aokromes: don't change this.

src/common/Collision/Management/VMapManager2.h
```diff
@@ -1,20 +1,20 @@
...
- *
- * This program is distributed in the hope that it will be useful, but WITHOUT
- * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
- * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
- * more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program. If not, see <http://www.gnu.org/licenses/>.
- */
+* Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
```
Aokromes: don't change this.

src/common/Collision/Maps/MapTree.cpp
```diff
@@ -1,20 +1,20 @@
...
- *
- * This program is distributed in the hope that it will be useful, but WITHOUT
- * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
- * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
- * more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program. If not, see <http://www.gnu.org/licenses/>.
- */
+* Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
```
Aokromes: don't change this.

src/common/Collision/Maps/MapTree.h
```diff
@@ -1,20 +1,20 @@
...
- *
- * This program is distributed in the hope that it will be useful, but WITHOUT
- * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
- * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
- * more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program. If not, see <http://www.gnu.org/licenses/>.
- */
+* Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
```
Aokromes: don't change this.

devtekve:
@Aokromes  Sorry for the modification in the copyright section, was a mistake when formatting tabs to spaces!

Vincent-Michael:
maybe should convert in enum
and comment style is ugly

tkrokli:
About comments in the code; maybe use `// comment here` instead (2x slash + 1 space) for all inserted lines?

devtekve:
@Vincent-Michael Sorry for my comment style :D Already removed them for bein redundant since I took you `enum` advise, thank you!.

@tkrokli On the comment I left, I followed your advise, thank you for reviewing my code!.

src/common/Collision/Management/IVMapManager.h
```diff
@@ -55,7 +55,8 @@ namespace VMAP
 
             virtual int loadMap(const char* pBasePath, unsigned int pMapId, int x, int y) = 0;
 
-            virtual bool existsMap(const char* pBasePath, unsigned int pMapId, int x, int y) = 0;
+            ///Changed from Bool to Int, in order to be able to use the same method with a little bit more of error descriptor.
```
DDuarte: Remove this comment, it doesn't really add anything.

Change the int to a new enum describing the possible return states:


Something like this would do:

```c++
enum class MapLoadError
{
     None,
     FileNotFound,
     VersionMismatch,
     Other
}

Issue #18739, opened 2017-01-04
[NPC] Ursal the Mauler - Deathscript


Rushor:
source: https://github.com/cmangos/classic-db/issues/24

http://www.wowhead.com/npc=2852/enslaved-druid-of-the-talon#comments:id=152934

By adashiel (274,298 – 13·81·927) on 2007/09/14 (Patch 2.1.3)
The screenshot shows one awake, but you'll nearly always find them sleeping. However, when Ursal the Mauler dies, they awaken, transform into Freed Druid of the Talon and fly away.

http://www.wowhead.com/npc=2853/freed-druid-of-the-talon

https://youtu.be/ISmLr3mnO1I?t=206
found a text for the creatures (https://youtu.be/8Ce3sznVtzU?t=246) it's from cata, but it should be in classic available too.

The nightmare of Ursal is over! We are free!

That script is missing.

**Branch(es):** 3.3.5 / master (Tell us which branch(es) this issue affects.)
**3.3.5** 

**TC rev. hash/commit:**  https://github.com/TrinityCore/TrinityCore/commit/bddf15d6965d7deafed3f0da3efda6b8e41de9e0

**Operating system:**  win 10

Rushor:
fix:

```sql
-- Ursal the Mauler SAI
SET @ENTRY := 2039;
UPDATE `creature_template` SET `AIName`="SmartAI" WHERE `entry`=@ENTRY;
DELETE FROM `smart_scripts` WHERE `entryorguid`=@ENTRY AND `source_type`=0;
INSERT INTO `smart_scripts` (`entryorguid`,`source_type`,`id`,`link`,`event_type`,`event_phase_mask`,`event_chance`,`event_flags`,`event_param1`,`event_param2`,`event_param3`,`event_param4`,`action_type`,`action_param1`,`action_param2`,`action_param3`,`action_param4`,`action_param5`,`action_param6`,`target_type`,`target_param1`,`target_param2`,`target_param3`,`target_x`,`target_y`,`target_z`,`target_o`,`comment`) VALUES
(@ENTRY,0,0,0,0,0,80,0,2000,2000,4000,4000,11,15793,0,0,0,0,0,1,0,0,0,0,0,0,0,"Ursal the Mauler - In Combat - Cast 'Maul'"),
[...]

Issue #17568, opened 2016-07-10
Core: Respawn position after movements


Killyana:
For many npcs, if they despawn after a waypoint or a follow action they will respawn where the despawn happen instead of the home position.

**Steps to reproduce the problem**:
1. Execute this sql: https://github.com/TrinityCore/TrinityCore/issues/16915#issuecomment-208766974
2. .go c id 8023
3. .quest add 2994  and   .add 9472
4. Open the cage and wait until the end of the event
5. wait until the npc respawn 2min
6. go -119.83095, -3499.7751, 119.578827
7. The npc will be there where it despawns instead of the home position.

This issue happen for many other npcs.

EDIT: This issue happen if the npc despawn while z_position > z_ground

**Branch(es)**: 3.3.5 / 6.x

**TC hash/commit**:  rev. 8edc1caf3263

tkrokli:
Related to [Core/SAI] `SMART_ACTION_SET_HOME_POS` #11593 ?

jackpoz:
this sounds like a change added by a commit in last weeks that changed the behavior

Killyana:
First time I noticed this was the 17th April.

jackpoz:
Was there a PR that changed waypoint functionality setting Home at every waypoint to avoid npc going back to start after evading ?

dr-j:
Yes this does not seem to effect every creature with wp but confirming for Brunhilldar Prisoners and all 3 OOX homing robots.

Killyana:
I reverted https://github.com/TrinityCore/TrinityCore/pull/11609 but the issue is not fixed.

jackpoz:
what about https://github.com/zengwf/TrinityCore/blob/6fac5dd859491b40a4638f970fb296227157161a/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp#L117 ? it seems home position is set to last waypoint when path ends

Shauren:
Not related, see https://github.com/TrinityCore/TrinityCore/blob/6.x/src/server/game/Entities/Creature/Creature.cpp#L2372

Killyana:
It's not related waypoints only, as npc 8023 despawn after a follow action.

Shauren:
Do you have any way to reproduce it shorter than watching a 2 minute event?

Kittnz:
Isn't it related to https://github.com/TrinityCore/TrinityCore/issues/17363 ?

Killyana:
[...]

Issue #16915, opened 2016-04-05
[3.3.5] DB/Quest: Saving Sharpbeak (quest 2994)


tea91:
**Description**:
http://www.wowhead.com/quest=2994/saving-sharpbeak
Saving Sharpbeak, in The Hinterlands, is half working. The quest can be completed perfectly normally, however, upon releasing him from the cage, we should auto mount him and Sharpbeak should fly to Aerie Peak and iirc some dialog should happen, and we can normally turn the quest in.
 Working quest, but not perfectly :P
**Current behaviour**: Tell us what happens
There is no SAI
**Expected behaviour**: Tell us what should happen instead
1. The event has to look this. 
2.  http://www.youtube.com/watch?v=QnytlvxuQmA
   http://www.youtube.com/watch?v=5yGyB1nfWyQ
3. Sharpbeat's Mother and Father have to come to rescue him.

**Steps to reproduce the problem**:
1. The event has to look this. 
2.  http://www.youtube.com/watch?v=QnytlvxuQmA
   http://www.youtube.com/watch?v=5yGyB1nfWyQ
3. Sharpbeat's Mother and Father have to come to rescue him.

**Branch(es)**: 335/6x
335
**TC hash/commit**:
d59e1b2a2966dacc400c9214d4824561912c90db
**TDB version**:
lastest update from 3.3.5. 60
**Operating system**:
win10 64bit

Rushor:
could not find any sniffs for this but a self made version is:

```sql
-- Saving Sharpbeak
-- Sharpbeak's Cage SAI
SET @ENTRY := 144070;
UPDATE `gameobject_template` SET `AIName`="SmartGameObjectAI" WHERE `entry`=@ENTRY;
DELETE FROM `smart_scripts` WHERE `entryorguid`=@ENTRY AND `source_type`=1;
INSERT INTO `smart_scripts` (`entryorguid`,`source_type`,`id`,`link`,`event_type`,`event_phase_mask`,`event_chance`,`event_flags`,`event_param1`,`event_param2`,`event_param3`,`event_param4`,`action_type`,`action_param1`,`action_param2`,`action_param3`,`action_param4`,`action_param5`,`action_param6`,`target_type`,`target_param1`,`target_param2`,`target_param3`,`target_x`,`target_y`,`target_z`,`target_o`,`comment`) VALUES
[...]

PR #18864, opened 2017-01-16
Game/Entities: Teleport Units with AI to db-spawnlocation on respawn

Lumber-Jack-2:
**Changes proposed:**

-  Teleport Units with AI to db-spawn location on respawn

- [X] 3.3.5

- https://github.com/TrinityCore/TrinityCore/issues/17568
- https://github.com/TrinityCore/TrinityCore/issues/18739
- https://github.com/TrinityCore/TrinityCore/issues/16915

yes and yes

- I know this code is wrong and it's a hack. anyway this PR is a good way to make core devs think about the real reason of the issue. I don't know what causes the issue and i don't know how to fix it propperly, but it breaks around 150 scripts ingame.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -2057,6 +2057,20 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
             delete targets;
             break;
         }
+		case SMART_ACTION_GO_SET_GO_STATE:
+        {
+            ObjectList* targets = GetTargets(e, unit);
+
+            if (!targets)
+                break;
+
+            for (ObjectList::const_iterator itr = targets->begin(); itr != targets->end(); ++itr)
+                if (IsGameObject(*itr))
+                    (*itr)->ToGameObject()->SetGoState((GOState)e.action.goState.state);
+
+            delete targets;
+            break;
+        }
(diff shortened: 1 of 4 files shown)


Ovahlord:
I dunno if this is a proper solution for that issue since it appears that Blizz is creating entirely new creatures instead of recycling the old ones.

Rushor:
yea but that's a complete different issue

streetrat:
i think setposition or what is it called would be enough before adding the unit to the map, so it won't trigger any teleport packets

Rushor:
no setposition in respawn is also a hack, we need to find the reason in core why the creatures respawns on despawnposition

Rushor:
and btw @streetrat would you like to look at https://github.com/TrinityCore/TrinityCore/issues/18865 and discuss a bit about it (i think it's not that huge amount of changes that need to be done in core)

Rushor:
and yea tried it with

```c++
        float x, y, z, o;
        GetRespawnPosition(x, y, z, &o);
        SetPosition(x, y, z, o);
```

result: same issue as without the addition

jackpoz:
> I don't know what causes the issue and i don't know how to fix it propperly

take your time to debug the respawn and see why it respawns in the wrong place. you can post your findings on the issue so others can contribute to the discussion too.

Rushor:
@jackpoz 

@BAndysc already found the reason. pls give a skilled comment to it https://github.com/TrinityCore/TrinityCore/issues/17568#issuecomment-272994471 ty

PR #18771, opened 2017-01-08
Core/Movement: Improve player movement

chaodhib:
**Changes proposed:**

Goal: Significantly improve player movement smoothness and reduce jerkiness
- [X] Add clean support of speed changes to all units
- [X] Add clean support of collision height changes to units controlled by players
- [X] Add clean support for knock backs to units controlled by players
- [X] Add clean support of the toggle-able movement flags (root, hover, gravity disabled, water walk, feather fall) to all units.
- [X] Add clean support of special movement flags (can fly, can transition between fly and swim) to units controlled by players. 
- [X] Add clean support of teleport to all units.

The biggest change is the remodeling of the way movement changes (change of speed, movement flag, etc..) are applied on units controlled by players. Let's take the example of a change of speed:
**Before**: when `SetSpeed(MOVE_RUN, 10)` was called on any unit, the change was active as soon as the method returned.
**Now**: when `SetSpeed(MOVE_RUN, 10)` is called on an unit moved by a player, first the server sends a packet to the player's client to inform it of the changes. The client then replies with a confirmation. Only when the confirmation is received by the server that the change is actually applied, and then propagated to the client of the other players. The reason why this change was needed can be found [here](https://github.com/TrinityCore/TrinityCore/issues/18582).

More information on this can be found in the header of this file: https://github.com/chaodhib/TrinityCore/blob/c792f35f0e6492d933c3aabc2cd20ebc42c38aad/src/server/game/Movement/MovementPacketSender.h

These changes take into account player taking possession of another player, of a creature or of a vehicle.
These changes also include an implementation of the movement counter and a queue to keep track of pending player changes (this change is also needed for the implementation of solid anti speed hack detection mechanism).

**Videos:**
Before: https://youtu.be/KnsQ_Zqz0kQ
After: https://youtu.be/MwpWViAxhbs

**Target branch(es):** 3.3.5

**Issues addressed:** Closes at least https://github.com/TrinityCore/TrinityCore/issues/18582

**Tests performed:** Tested IG.

**Known issues and TODO list:** (add/remove lines as needed)

[...]

Issue #18306, opened 2016-11-24
[3.3.5]Spells not interacting properly with target movement


Aknunx:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:** When you cast a spell against a player in movement, the player who's affected by that spell will see the normal "outcome" of whatever spell has been casted against him, but the player who casted the spell will see some strange and mixed results depending on the spell.
Here we can see it with a clone against a player who's trying to LoS the clone, you'll see how, in one of the vids, after the clone, the target "tp" to his real position and in the other he does it mid-clone, making things really awkward for the caster.

![](https://puu.sh/sqWPs/e5af58d765.mp4)

![](https://puu.sh/sqWNu/a7cb0e2cf5.mp4)

**Current behaviour:** Spells do not interact properly with the target movement, for example, the clone, that after landed, teleports the player to his real position.

**Expected behaviour:** A clone should never tp a player neither mid-clone or at the end of it, and if the clone tps the player behind out of LoS of the caster, means that when the clone landed the target was already out of LoS, in which case the clone should've never landed in first instance.

**Steps to reproduce the problem:**

1.  Create a druid with clone
2.  Try to clone an enemy player who's moving
3.  See how after or mid-clone the player will TP to his real position.

**Branch(es):** 3.3.5 / master (Tell us which branch(es) this issue affects.)

**TC rev. hash/commit:**   https://github.com/TrinityCore/TrinityCore/commit/ab4ec5bd01122d503af38fc8464ff7528d297604

**TDB version:**  TDB 62

**Operating system:**  Ubuntu 14.04

[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing C++, create a Pull Request.)

Persephone:
Confirm. Similar effects happends with fearing, sometimes fearing looks "okay", other times it appears as if the targeted player just teleports from place to place.

chaodhib:
Cyclone applies a stun on the target, which roots the target (by looking at sniff, we can see a MSG_MOVE_ROOT packet sent). Roots (and a lot of other movement flags) are not handled properly at the moment. It should be fixed by https://github.com/TrinityCore/TrinityCore/pull/18771 when it's ready.

[...]

PR #18889, opened 2017-01-19
Core/Movement: Fix Scatter Shot issue on moving players - cleaner approach 

chaodhib:
Cleaner approach to https://github.com/TrinityCore/TrinityCore/pull/18336

The first commit is sufficient as a fix on its own (and as you can see, it does not actually touch the Confused MoveGen code). In the second commit I try to make Confused MoveGen more retail like but it's minor changes.

The general idea of the fix is that the methods Unit::IsStopped() and Unit::StopMoving() should now work properly on player-moved units, which was not the case until now.

The PR is ready to be reviewed and potentially merged. However, the changes here have much broader implications than in the first PR, so extensive testing is needed.

**Target branch(es):**

- [X] 3.3.5

**Issues addressed:** Closes #  https://github.com/TrinityCore/TrinityCore/issues/15952

**Tests performed:** Tested IG. Works against moving players.

**Known issues and TODO list:** Nothing that I'm aware of.

### DIFF
--- a/src/server/game/AI/PlayerAI/PlayerAI.cpp
+++ b/src/server/game/AI/PlayerAI/PlayerAI.cpp
@@ -1300,7 +1300,7 @@ void SimpleCharmedPlayerAI::UpdateAI(const uint32 diff)
             _forceFacing = true;
         }
 
-        if (me->IsStopped() && !me->HasUnitState(UNIT_STATE_CANNOT_TURN))
+        if (!me->HasUnitState(UNIT_STATE_MOVING) && !me->HasUnitState(UNIT_STATE_CANNOT_TURN))
         {
             float targetAngle = me->GetAngle(target);
             if (_forceFacing || fabs(me->GetOrientation() - targetAngle) > 0.4f)
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -504,6 +504,10 @@ void Unit::UpdateSplineMovement(uint32 t_diff)
 
 void Unit::UpdateSplinePosition()
 {
+    // this method could be called on players. players usually do not have a movespline initialized. the following condition is therefore needed.
+    if (!movespline->Initialized())
+        return;
+
     static uint32 const positionUpdateDelay = 400;
 
     m_movesplineTimer.Reset(positionUpdateDelay);
@@ -11056,13 +11060,15 @@ void Unit::StopMoving()
     ClearUnitState(UNIT_STATE_MOVING);
 
     // not need send any packets if not in world or not moving
-    if (!IsInWorld() || movespline->Finalized())
+    if (!IsInWorld() || !isMoving())
         return;
 
     // Update position now since Stop does not start a new movement that can be updated later
-    UpdateSplinePosition();
+    if(!movespline->Finalized())
+        UpdateSplinePosition();
     Movement::MoveSplineInit init(this);
     init.Stop();
+    RemoveUnitMovementFlag(MOVEMENTFLAG_MASK_MOVING);
 }
 
 void Unit::SendMovementFlagUpdate(bool self /* = false */)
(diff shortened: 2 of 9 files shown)


ForesterDev:
Maybe we should use same methods to fix "fear teleport" report? https://github.com/TrinityCore/TrinityCore/issues/18306#issuecomment-263516101

chaodhib:
@ForesterDev I replied in the thead. I don't think it will help with Fear. The movement generator that fear uses probably needs to be rewritten.

Aokromes:
So, close https://github.com/TrinityCore/TrinityCore/pull/18336 ?

shenhuyong:
I tested it, I found a bug, the waypoints of creatures are not working.

chaodhib:
could you go more into the details? What is not working exactly? How to reproduce? Could you provide screenshots and/or a video?

shenhuyong:
Go into Karazhan, you can find any mob or npc with waypoints will not move and only stay at the start point.

chaodhib:
Indeed. Thanks for the feedback. I will see how to fix that.

chaodhib:
Should work correctly now.

Eliminationzx:
Looks ok to me.

chaodhib:
Bump. This PR is also, as far as I can tell, ready to be merged.

Aokromes:
jump into irc :)

Treeston:
@chaodhib rebase & fix conflicts please then I can merge

tripleslash:
@chaodhib Doesn't the issue mentioned in the issue post from you [here](https://github.com/TrinityCore/TrinityCore/issues/15952#issuecomment-263153057) still apply? Specifically this one:

When the player gets affected by the spell, he will be at a certain position. That position is what the players will see on their screens when the spell land. However because the server does not use extrapolation, it thinks that the player is still at the latest received position. Therefore, when the Confused MoveGen is gonna kick in, it's gonna start the spline at the wrong place.

chaodhib:
@tripleslash 
No, I was wrong in my initial investigation:

The teleportation had 2 sources of "lag" and they were adding up. The first source was what i explained in the comment you mentioned. The second source however is much much bigger than the first. To the point where, the first source is actually a non issue. The glitch disappaears entirely once the second source is removed.

So what was the second source of lag? cf description of https://github.com/TrinityCore/TrinityCore/pull/18336#issue-192134046

I will resume working on this PR shortly after https://github.com/TrinityCore/TrinityCore/pull/18771 is merged.

Killyana:
If you cast Scatter Shot on a player with a flying mount it will lands directly and execute Confused movements on ground.

Aokromes:
Conflicting files
src/server/game/Entities/Unit/Unit.cpp
src/server/game/Entities/Unit/Unit.h
src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
src/server/game/Movement/MovementGenerators/PointMovementGenerator.cpp
src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp

robinsch:
This is still a valid issue. Fix above doesn't work.

chaodhib:
I can't reproduce the issue anymore, from 3.3.5 branch. Is it still valid?

Killyana:
Still able to reproduce it, use 2 characters, put one on auto run with a mount and use Scatter Shot on it, he will play the confuse movement then instead stop moving he will kind of slide to a position and continue running.

robinsch:
All movement gens have this issued. You can spam w and s and then fear the unit. It will desync the movement

chaodhib:
@Killyana 
The only aim of this PR was to solve the small jump the target had at the start/end of the Scatter Shot effect (which made the follow up with Freezing Arrow very hard to execute). Since that jump is already fixed on the 3.3.5 branch, I don't see the point of this PR anymore. The animation at the end of the effect is out of scope. I don't think my PR ever fixed that problem sadly.

@robinsch I'm confused. Which issue are we talking about?

PR #18904, opened 2017-01-21
Core/Entities: hover movement fix

### LINKED ISSUE #15177: [3.3.5] Hover Mode not working (55$ bounty)

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Issues addressed:** Closes #15177 closes #18923

**Tests performed:** Compiles, basic ingame tests

The idea behind this patch is to keep the hovered unit on ground height (server side) and just hover it on client. This approach may be wrong, this needs some more consultations. Hovering above 2-3yds should be always combined with disabled gravity.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -905,15 +905,6 @@ bool Creature::Create(ObjectGuid::LowType guidlow, Map* map, uint32 phaseMask, u
             break;
     }
 
-    //! Need to be called after LoadCreaturesAddon - MOVEMENTFLAG_HOVER is set there
-    if (HasUnitMovementFlag(MOVEMENTFLAG_HOVER))
-    {
-        z += GetFloatValue(UNIT_FIELD_HOVERHEIGHT);
-
-        //! Relocate again with updated Z coord
-        Relocate(x, y, z, ang);
-    }
-
     uint32 displayID = GetNativeDisplayId();
     CreatureModelInfo const* minfo = sObjectMgr->GetCreatureModelRandomGender(&displayID);
     if (minfo && !IsTotem())                               // Cancel load if no model defined or if totem
@@ -1106,7 +1097,7 @@ void Creature::SaveToDB(uint32 mapid, uint8 spawnMask, uint32 phaseMask)
     {
         data.posX = GetPositionX();
         data.posY = GetPositionY();
-        data.posZ = GetPositionZMinusOffset();
+        data.posZ = GetPositionZ();
         data.orientation = GetOrientation();
     }
     else
@@ -2268,7 +2259,7 @@ bool Creature::LoadCreaturesAddon()
         //! Check using InhabitType as movement flags are assigned dynamically
         //! basing on whether the creature is in air or not
         //! Set MovementFlag_Hover. Otherwise do nothing.
-        if (GetByteValue(UNIT_FIELD_BYTES_1, UNIT_BYTES_1_OFFSET_ANIM_TIER) & UNIT_BYTE1_FLAG_HOVER && !(GetCreatureTemplate()->InhabitType & INHABIT_AIR))
+        if (GetByteValue(UNIT_FIELD_BYTES_1, UNIT_BYTES_1_OFFSET_ANIM_TIER) & UNIT_BYTE1_FLAG_HOVER/* && !(GetCreatureTemplate()->InhabitType & INHABIT_AIR)*/)
             AddUnitMovementFlag(MOVEMENTFLAG_HOVER);
     }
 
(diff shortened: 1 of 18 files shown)


Keader:
i cant test it again, but you fixed issues with melee range?
E.G. Melees can reach Mimiron in air phase?

xinef1:
No, this was not changed. The easiest way would be to check for hoverheight in IsWithinMeleeRange function (currently the hover is subtracted in this function so my patch do not change this functionality).

xinef1:
Other approach is presented here https://github.com/xinef1/TrinityCore/commit/2b7ad67465430a683cd94a5bd63ee06d7f552fb5

Unit is hovered both at client and at server. Possible problems may arise with splines launched from different parts of the core.

ccrs:
For me this is the correct solution, hover animation seems like just a client side effect (only a visual)

But I disagree with the Disable gravity part, its not needed and probably not blizzlike.

If you are seeing visual "glitches" without adding disable gravity to the creature is because the server is sending ground Z values to the client on UpdatePosition (splines and teleports).

To solve that you will need to correct the value sent on the splines and the values sent on the creature actual position (UpdateSplinePosition -> UpdatePosition). In both cases those values must have the hoverheight added to its z.

Also to accomplish that you will also need to check methods in Object such as UpdateAllowedPositionZ or NormalizeZforCollision

ccrs:
BTW have you checked Hover on unit death? talking about floating corpses.

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2268,7 +2259,7 @@ bool Creature::LoadCreaturesAddon()
         //! Check using InhabitType as movement flags are assigned dynamically
         //! basing on whether the creature is in air or not
         //! Set MovementFlag_Hover. Otherwise do nothing.
-        if (GetByteValue(UNIT_FIELD_BYTES_1, UNIT_BYTES_1_OFFSET_ANIM_TIER) & UNIT_BYTE1_FLAG_HOVER && !(GetCreatureTemplate()->InhabitType & INHABIT_AIR))
+        if (GetByteValue(UNIT_FIELD_BYTES_1, UNIT_BYTES_1_OFFSET_ANIM_TIER) & UNIT_BYTE1_FLAG_HOVER/* && !(GetCreatureTemplate()->InhabitType & INHABIT_AIR)*/)
             AddUnitMovementFlag(MOVEMENTFLAG_HOVER);
```
ccrs: This should be sethover manual AddUnitMovementFlag are not cool

src/server/game/Entities/Object/Object.cpp
```diff
@@ -361,10 +361,7 @@ void Object::BuildMovementUpdate(ByteBuffer* data, uint16 flags) const
 
             *data << object->GetPositionX();
             *data << object->GetPositionY();
-            if (isType(TYPEMASK_UNIT))
-                *data << unit->GetPositionZMinusOffset();
-            else
-                *data << object->GetPositionZ();
+            *data << object->GetPositionZ() + (unit ? unit->GetHoverHeight() : 0.0f);
```
ccrs: uh, better check if hover is active or not

ccrs: Same with the other uses of GetHoverHeight

src/server/game/Entities/Creature/Creature.cpp
ariel-: That was the original code, though

ccrs: yeah I know, but its still incorrect xD

xinef1:
@ccrs thanks for the feedback, i don't really remember why disablegravity flag affecting code was added but there had to be a reason for it (maybe pussywizard will know, probably unit sometimes felt visually) :D Data sent in packets is properly appended in both value updates and spline packets, there should be no issues there. I've would like to get shauren idea about this approach but i won't dare to call him xD 

I've linked other approach which relocates the unit on server also, create/update packets sends direct z position not affected (because unit is really on increased z position) and subtracts the hover height for path calculations. The paths are then offseted with hover in spline system. Possible problems may arise because i've only validated main movement generators and there are various places in core, from where spline can be launched but in my opinion they should pose no problems. Additionally distance calculations for targeted movement were fixed to account for the hover increase.

src/server/game/Entities/Creature/Creature.cpp
xinef1: SetHover with !IsHovering, because this can be doubly called (home movement generator for example)

ccrs:
The other approach is incorrect IMO, having to adjust all the calculations (paths, range, etc) for just a visual element feels just bad.

Treeston:
@xinef1 please resolve conflicts so I can review this. Also, imo refactor `GetHoverHeight` -> `GetCurrentHoverHeight` to make the "only if hovering" explicit.

PR #18923, opened 2017-01-22
Core/Movement: Rework of the Hover movement flag handling

chaodhib:
**Changes proposed:**
The server does not need to keep track of an additional Z component when handling hovering units. The movement flag only puts one restriction on the Z coordinate of the unit: a force is applied on the unit until it reaches z=ground height + hover height. Knowing this, we need to make sure that a hovering unit is always at the right distance from the ground. When this is taking care of, then no additional care is needed.

The Unit::GetPositionZMinusOffset() method has been deleted and every occurrence has been replaced by GetPositionZ() since it is not needed.

Also cf https://github.com/TrinityCore/TrinityCore/issues/15177#issuecomment-274336307

**Target branch(es):** 3.3.5

**Issues addressed:** Closes https://github.com/TrinityCore/TrinityCore/issues/15177 closes #18904

**Tests performed:** Tested IG against Mimiron P3 and the Valkyr twins.

- [x] The Chase MoveGen has been adapted to take into account that the mover is hovering. I provided a crude implementation for proof of concept value. However, it is pretty lacking: 
- the Valkyr twins will stand too far away to AA.
- The ACU (Mimiron P3) will get unnecessarily close to the player its chasing. EDIT: wont fix. this problem is not part of the PR. I don't know how to flag the ACU (Mimiron P3) as a "ranged" attacker.
This point needs to be solved before the PR can get merged.


joschiwald:
#18904

chaodhib:
@xinef1 and me have a different approach to the issue. cf discussion about it here: https://github.com/TrinityCore/TrinityCore/issues/15177#issuecomment-274336307

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2782,22 +2777,22 @@ Unit* Creature::SelectNearestHostileUnitInAggroRange(bool useLOS) const
     return target;
 }
 
-void Creature::UpdateMovementFlags()
+void Creature::UpdateMovementFlags() // i'm wandering if this method really needs to be called periodically. shouldn't this be done only once, at creation of the creature ? And for transitions (land -> swim, air -> land, etc..), they are usually scriptted therefore the change of movement flag should be done by script and not automatically applied.
```
Shauren: To answer your question - what if the floor under creature changes state? GameObjects are part of height calculations

src/server/game/Entities/Player/Player.cpp
```diff
@@ -1937,7 +1937,10 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
         {
             Position oldPos = GetPosition();
             if (HasUnitMovementFlag(MOVEMENTFLAG_HOVER))
-                z += GetFloatValue(UNIT_FIELD_HOVERHEIGHT);
+            {
+                float ground = GetMap()->GetHeight(GetPhaseMask(), x, y, z);
```
Shauren: this is completely wrong, teleport destination can be anywhere, including high in the air

src/server/game/Entities/Creature/Creature.cpp
malcrom: Creature::UpdateMovementFlags() issue is in here too. hard coded Inhabittype in template is wrong for most flying creatures.

ariel-: It does some expensive vmap queries each update, maybe it could be throttled a little

chaodhib:
The PR is ready to be tested and review. All should work as intented.

src/server/game/Entities/Player/Player.cpp
chaodhib: Yes, a unit can be TPed anywhere, including in the air. However, for a unit without Hover & DisableGravity flag, as soon as such unit is put at a new position in the air, the client will apply a force (gravity) on that unit until the unit reaches the ground.

It works more or less the same way for hovering units: As long as a unit is not at z=ground_height + hover_height, the client will apply a force on that unit to make the unit move to that stable position.

Ok, that's great, but where am I going with this? Well, there is something missing in the core. If the client applies a force on units controlled by the server, the server needs to do the same. Otherwise, the unit will move in the client but not in the server. And a desynchronization of the unit's position as seen by the server and by the client will occur.

In order to circonvent this problem (and until such time that gravity on falling creature  is implemented), this PR's approach is to avoid putting creatures in "unstable" positions. And as long as creatures are in "stable" positions (= on the ground for units subjected to gravity, = ground_height + hover_height for hovering units), all is well.

I'm not saying this approach is perfect. I will see what I can do next.

ccrs:
Sooo essentially your approach is to set the creatures server and clientside to the hover position and use disable gravity flag to prevent the creature to be located to the ground (those annoying blinks from air to ground).

Said it in the other PR, for me this is not the correct way.

chaodhib:
1. DisableGravity and Hover are 2 different movement flags (CanFly and Flying are 2 others but that's another discussion).
2. these 2 flags only changes the way of the physical engine treats affected units: 
* Gravity is disabled for DisableGrabity. 
* For hover, gravity is applied toward z=floor level + hover height. So if the player is above that z, it will fall. if it's precisely at that level, it will not move and if it's under, it will ascend.
3. these 2 flags only change the forces applied on the affected unit. It does NOT have a direct impact of their positions. If a hovering unit is at (0, 0, 5.0), its position will be (0, 0, 5.0) and not (0, 0, 0). Technically, we could model the (0,0,0) in the server and deal with side effects but that approach is just masochistic in my opinion. It will force the server to deal with a bunch of issues (including melee range validation which the other PR needs to do and which my PR does not).

On [this branch ](https://github.com/chaodhib/TrinityCore/tree/hover_experiment) I added those commands:
- .debug hover on
- .debug hover off
- .debug sethoverheight #value
- .debug disablegravity on
- .debug disablegravity off

[Here is a video ](https://youtu.be/l2tLTgJA0hM) of me playing around with these commands to manipulate the 2 movement flags. I suggest that everyone go and play around with the flags to get a better feeling on how they make the client react. It's pretty interesting.

PS: the blinks that you see are the result of de-synchronization between the unit's position as seen by the server and by the client. Having the same position on client and the server at all times guarantees that the blinks never happen.

I will give more information about my approach later. In the meantime, I would appreciate if I could get feedback from IG testing, independently of the code itself.

ccrs:
If blink happens can be for several reasons
1 - the spline sent has one or more points with an incorrect Z value
2 - the server is correctiong the creature Z value to the ground incorrectly
3 - the client corrects visually the position of the model to hover height because the movementflag is set

Setting both server and client positions to hoverheight simply fucks up everything, and also remind that:
- Range calculations should be like if the unit is not hovering
- Position relocations should be done like if the unit is not hovering
- ....
There is a nice list of things that forces you to revert the logic to satisfy their logic

ariel-:
What did sniffs say about it? Coords sent are ground or hover level? Moveflags etc

ccrs:
So far I've only seen ground positions for hovering units, correct me if I'm wrong

chaodhib:
@ccrs "Range calculations should be like if the unit is not hovering" That is incorrect. In the client, range calculations are based on the real position of the unit (in the air):
- In P3 of Mimiron fight, the ACU isn't reachable by melees, simply because the real position of the unit is in the air and its combat reach is lower than its hover height. 
- Against the Vaklyr twins on the other hand, even though they are hovering in the air, their hover height is smaller than their combat reach, therefore they are attackable by melee players on the ground.

The fact that range calculations are done using the real height of the unit and not the height on the ground is a pretty good indicator that the server uses the position in the air internally.

"Having both position values synchronized and not affecting the other core mechanics is something really really difficult to accomplish, and the result will be exactly the same."
I'm sorry but you are completely wrong with this sentence. Simply because, "Having both position values synchronized " is exactly what the core is already doing for every non hovering creatures. The core has the position of each unit, which can change on each game tick if the unit is moving. And it's the core's job to make sure that the position of each units on the player's screen is in alignment with its own. All the more reason to not treat Hover separately.

ccrs:
"In the client, range calculations are based on the real position of the unit (in the air)"

So? Client gets those values from the positions sent by the server. We still dont know what the values are serverside, and thats the key matter here. How we handle serverside creature position.

1. Both positions are the same
2. Server keeps position without hover while it sends the client the correct values to have the correct visual.

And thats pretty much it, decide which approach is the best.

Keader:
> Bradley@Keader-Laptop MINGW64 ~/Documents/Trinity (3.3.5)
> $ git apply hover2.diff
> hover2.diff:332: trailing whitespace.
>                 float A = i_target->GetObjectSize() + owner->GetObjectSize() + CONTACT_DISTANCE; // default 2D distance search used by GetContactPoint
> hover2.diff:360: trailing whitespace.
> 
> hover2.diff:372: trailing whitespace.
>                 float A = i_target->GetObjectSize() + size + i_offset; // the 2D distance search used by GetClosePoint(x, y, z, size, i_offset, i_angle)
> hover2.diff:393: trailing whitespace.
>         float maxAllowedDistance = 4.0f / 3.0f - CONTACT_DISTANCE;
> warning: 4 lines add whitespace errors.

Getted by https://github.com/TrinityCore/TrinityCore/pull/18923.diff

Keader:
And testing with Mimiron i get it when arial phase starts

> c:\users\bradley\documents\trinity\src\server\game\maps\Map.h:622 in Map::getNGrid ASSERTION FAILED:
>   x < MAX_NUMBER_OF_GRIDS && y < MAX_NUMBER_OF_GRIDS

and server crash

chaodhib:
@Keader Thanks for the feedback.

Regarding Illidan, Hover is not used. The packet SMSG_MOVE_SPLINE_DISABLE_GRAVITY is sent and not SMSG_SPLINE_MOVE_SET_HOVER. And Illidan's hover height is 1.0 which is the default value. On creatures that do use Hover, this value is usually higher then 1. So it's outside of the scope of this PR. I could look into it next though. Does it work without the code of my PR?

Regarding Mimiron, indeed it's crashing after the latest changes I did. I'll look into it.

chaodhib:
Mimiron P3 issue fixed

chaodhib:
@ccrs 

> So? Client gets those values from the positions sent by the server. We still dont know what the values are serverside, and thats the key matter here.

True point.


> Again, I think its much better to just take option 2 and ""add the values to the packets sent to the client"" (making it simple) like the other PR does.
The alternative solution is simpler short term to implement but it will backfire and will be harder to manage in the long term when we will need to deal with edge cases. It's an illusion to think that we are saving time and minimizing complexity with that approach imo.

> the problem is that this modification will affect literally every calculation in which 3D position is used
It's more complicated than that actually. It's not all black and white. The alt solution also has to do adjustments with 3D positions. Its the other solution that has to include changes to the range validation in order to work. Not mine. I did not modify a single LOC in range validation. It's also the other solution that is modifying MoveSpline and MoveSplineInt, not mine. However, I did had to include more code to deal with the path finding.

xinef1:
You MUST modify the data that is sent by spline. It does not matter where you do it (alter final z in targeted movement generator, etc.). Just remember that anything returned from movemaps will have z position on ground and you have to alter this. Some random MovePoint in icecrown script called with ground position. There is no point in changing literally every call from everything with every possible data to account for hover at the moment of call. It is much better to assume that as for now, everything was called with ground position and the path passed to spline is on ground height, therefore it is sufficient to modify height just at this one point.

chaodhib:
I tested both of your commits. Frankly it's not working great compared to what I have. Either you did not test your stuff or I messed up something in my TC.

Some outside opinions would be nice.

Palabola:
@chaodhib  Agree, we should touch Movement system and do not keep old outdated parts anymore.

xinef1:
Both of my patches work as intended - tested in game and see no problem, but this is not even needed and slides away from the main issues. It is enough to analyze the code to see results and possible problems.

> Please. Most of the movement codebase needs some serious rewrite. Don't tell me that because some > code has been there for a few years that it should be kept that way. Client side shift is a lazy and 
> messy approach that will burden future TC contributors with the same or even a higher amount of 
> complexity than dealing with hover in a more regular way.

And what this has to do with what I've written? The hover does not work at all in TC. I referred to my server which I ran for a few years and the hover change posed no problems at all. I do not see any complexity addition for future programmers, you do not touch code you don't understand and if you do understand it - what is the problem? Your example with GetPositionZ() is completely missed, you don't care what is returned by the function and assume it is correct (and it is no matter the approach, from the perspective of the server). No matter the approach you don't care about internals and no matter the approach you will never have to worry about hover if you script a creature or do something else. Ill give you better example. You have an hovering unit, let's call MovePoint in SAI to move to another object. At this point your patch already exploded because it only changes final position generated in TargetedMovementGenerator. Turn on MoveMaps and your patch explodes even in TargetedMovementGenerator, because whole generated path will be on ground height, maybe except the last point. I've wrote this in few of my comments regarding this topic and if you still don't see the issue, I give up. Some higherups should join the discussion.

chaodhib:
Using your changes, I recorded a try on Mimiron P3 and the Valkyr twins [HERE](https://youtu.be/wmZPKlPHZtw). For me, that's not fine. The only thing I have that is not up to date are my MMAPs (v5 instead of v6) but I doubt that it changes anything.

Anyway, I'm wasting time testing your stuff and arguing with you instead of working on my own stuff. I'm not going to further discuss with you, because a) why enter into technical details about the code if your changes are not even working? b) I don't agree with your approach so let's agree to disagree on that one.

xinef1:
You did not even modify twin valkyrs code (its path for hover and not twin valkyrs), to turn hover on ;) you do not even read code and analyze what is going on, you just guess and see what appears on client, it pointless to discuss any further.

chaodhib:
"you do not even read code and analyze what is going on, you just guess and see what appears on client"

You have saying that stuff since we started this conversation. Personal attacks followed by personal attacks. Meanwhile a) my changes work smoothly and yours don't. b) you don't seem to understand that because I'm using a different approach than yours, the changes that you think are necessary are actually not needed.

Because of your attitude, I won't explain to you how and why my approach works.

Now piss off.

src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/trial_of_the_crusader.h
```diff
@@ -62,8 +62,8 @@ const Position ToCCommonLoc[]=
     {563.833008f, 195.244995f, 394.585561f, 0}, //  5 - Center
     {573.5f, 180.5f, 395.14f, 0},               //  6 Move 0 Right
     {553.5f, 180.5f, 395.14f, 0},               //  7 Move 0 Left
-    {573.0f, 170.0f, 395.14f, 0},               //  8 Move 1 Right
```
sucklifex: I think these data was took from sniff

Kittnz:
You both have to calm down. We are all here to try to improve the code, if someone disagrees with something you have written try to understand it in a meaningfull discussion.

Ontopic, i don't understand why we would have to change the code of valkyr twins for the hover to work?

Killyana:
For the valkyr twins we need just to add bytes1= 33554432 = Hover mode on creature template addon, Hover must be handled via the DB and not the cpp.

Viste:
don't understand why you change bosses cpp code it's hacky

chaodhib:
oh, that's strange. i thought my "review" comments were already posted. looks like not. TIL

src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/trial_of_the_crusader.h
chaodhib: Nop: https://gist.github.com/chaodhib/02508362678cf1e28505d885b74db5e4#file-twin-vaklyr-hover-analysis-txt

No waypoint seems to be blizz like for this boss it seems at the moment.

chaodhib:
The changes that I did on Mimiron are completely retail like. The way the third phase is orchestrated is described in the second paragraph [here ](https://github.com/TrinityCore/TrinityCore/issues/15177#issuecomment-274336307). Basically, the ACU starts with the DisableGravity movement flag set. Then DisableGravity needs to be removed at some point. And at the same time Hover needs to be set. 

And actually with the changes that I have done, I actually didn't go far enough: he should first spawn inside the ceiling and then come down to where he is currently spawned at. This can be seen [here](https://youtu.be/2RrgpASuIsc?t=5m15s) (and in sniff). I will let someone else fix this detail since it has nothing to do with Hover.

The changes on the twins on the other hand.. yeah, it's not really mergable. Way-points for this boss are unretail-like and should be rewritten. As it can be seen [here](https://gist.github.com/chaodhib/02508362678cf1e28505d885b74db5e4#file-twin-vaklyr-hover-analysis-txt).

@Killyana I gotta admit I'm not really knowable about the field "bytes1" not its particular value 33554432  even though I have seen it appear in sniffs. What does it do exactly?

chaodhib:
Partially replying to my own question: UNIT_FIELD_BYTES_1 appears to be named "animTier" based on some decompiled code.

Treeston:
@chaodhib I'm curious about the "force" (<-> gravity) you mentioned earlier, and how this is not implemented. If I understood you correctly, this can be reduced to:

- Each creature has a "stable" Z position.
  - For flying units, this is the current position.
  - For nonflying units, this is ground Z
  - For hovering units, this is ground Z + hover height
- If the creature is not at its Z position, its client position Z is constantly adjusted to reach this stable Z position ("falling" upwards for hover)

Am I correct? If yes, this should be a "fairly" simple adjustment I could make in `Unit::Update`. Do you have a reproducible case where this incorrect falling behavior can currently be observed on a plain core?

Aokromes:
This is still needed?

chaodhib:
I guess not.

PR #18950, opened 2017-01-24
Core/Spells: Define and implement some new spell attributes

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles

Any feedback is welcomed

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -17804,7 +17804,7 @@ void Player::_LoadAuras(PreparedQueryResult result, uint32 timediff)
             }
 
             // negative effects should continue counting down after logout
-            if (remaintime != -1 && !spellInfo->IsPositive())
+            if (remaintime != -1 && ((!spellInfo->IsPositive() && spellInfo->Id != 15007) || spellInfo->HasAttribute(SPELL_ATTR4_FADES_WHILE_LOGGED_OUT))) // Resurrection sickness should not fade while logged out
             {
                 if (remaintime/IN_MILLISECONDS <= int32(timediff))
                     continue;
--- a/src/server/game/Miscellaneous/SharedDefines.h
+++ b/src/server/game/Miscellaneous/SharedDefines.h
@@ -467,7 +467,7 @@ enum SpellAttr4
 {
     SPELL_ATTR4_IGNORE_RESISTANCES               = 0x00000001, //  0 spells with this attribute will completely ignore the target's resistance (these spells can't be resisted)
     SPELL_ATTR4_PROC_ONLY_ON_CASTER              = 0x00000002, //  1 proc only on effects with TARGET_UNIT_CASTER?
-    SPELL_ATTR4_UNK2                             = 0x00000004, //  2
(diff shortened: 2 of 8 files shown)


tkrokli:
Cosmetic detail: if `SPELL_ATTR3_NO_PUSHBACK` gets approved, consider improving its comment (25) text ;-)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -17847,7 +17847,7 @@ void Player::_LoadAuras(PreparedQueryResult result, uint32 timediff)
             }
 
             // negative effects should continue counting down after logout
-            if (remaintime != -1 && !spellInfo->IsPositive())
+            if (remaintime != -1 && ((!spellInfo->IsPositive() && spellInfo->Id != 15007) || spellInfo->HasAttribute(SPELL_ATTR4_FADES_WHILE_LOGGED_OUT))) // Resurrection sickness should not fade while logged out
```
Shauren: Uh, I think it should fade when logged out but let me test it on retail first

xinef1: I'm 99% sure it didn't fade on wotlk, but ill adjust the code if necessary :)

src/server/game/Miscellaneous/SharedDefines.h
```diff
@@ -519,9 +519,9 @@ enum SpellAttr6
     SPELL_ATTR6_CAN_TARGET_UNTARGETABLE          = 0x01000000, // 24
     SPELL_ATTR6_NOT_RESET_SWING_IF_INSTANT       = 0x02000000, // 25 Exorcism, Flash of Light
     SPELL_ATTR6_UNK26                            = 0x04000000, // 26 related to player castable positive buff
-    SPELL_ATTR6_UNK27                            = 0x08000000, // 27
+    SPELL_ATTR6_LIMIT_PCT_HEALING_MODS           = 0x08000000, // 27 some custom rules - complicated
```
ariel-: Defined but not implemented

src/server/game/Entities/Player/Player.cpp
Aokromes: At least on legion Resurrection sickness fades while logged out.

Aokromes: https://us.battle.net/forums/en/wow/topic/1670088555
https://us.battle.net/forums/en/wow/topic/2228173818
https://us.battle.net/forums/en/wow/topic/2369682005

so, @xinef1 is right

ariel-: Better to apply spellinfo correction removing the attribute?

r00ty-tc: The issue isn't spellinfo from what I can see. It's a negative spell and the check for not positive traps it. The spell doesn't have the fades flag set and this is correct.

The question we need to ask is whether all the spells that have the flag are the correct list and if so, then we can drop the positive check. If not, then we keep in the exception.

Shauren (approved):

PR #18952, opened 2017-01-24
Properly stop attacking after player leaves ffa area (like Gurubashi Arena)

### LINKED ISSUE #13815: [3.3.5] Bug autoattack doesn't cancel when get out of PvP Areas like Gurubashi Arena
If you are autoattacking an enemy in, for example, Gurubashi Arena (Battle Ring) and then the enemy and you get out of it, you will continue autoattacking the enemy until he dies/you stop autoattacking/you swap or cancel the target.

This is a very old bug but it isn't fixed yet, players can abuse this very easily.

Using lastest rev https://github.com/TrinityCore/TrinityCore/commit/cc00321409707e50da0826b758bcd1242d7fcec7
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7363153-3-3-5-bug-autoattack-doesn-t-cancel-when-get-out-of-pvp-areas-like-gurubashi-arena?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Issues addressed:** Closes #13815

**Tests performed:** Compiles, no ingame tests

Review and feedback is welcomed.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -5777,6 +5777,24 @@ bool Unit::AttackStop()
     return true;
 }
 
+void Unit::ValidateAttackersAndOwnTarget()
+{
+    // iterate attackers
+    AttackerSet toRemove;
+    AttackerSet const& attackers = getAttackers();
+    for (AttackerSet::const_iterator itr = attackers.begin(); itr != attackers.end(); ++itr)
+        if (!(*itr)->IsValidAttackTarget(this))
+            toRemove.insert(*itr);
+
+    for (AttackerSet::const_iterator itr = toRemove.begin(); itr != toRemove.end(); ++itr)
+        (*itr)->AttackStop();
+
+    // remove our own victim
+    if (Unit* victim = GetVictim())
+        if (!IsValidAttackTarget(victim))
+            AttackStop();
+}
+
 void Unit::CombatStop(bool includingCast)
 {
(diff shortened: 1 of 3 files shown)


ariel-:
I think faction changes should be somehow broadcasted, making the nearby players stop attacking.

ie, when attacking some hostile creature and you .debug setvalue 55 35, you keep auto attacking it

xinef1:
Cannot check the code now, but I recall this is the only place that can change FFA flag. If you feel similar code should be put somewhere else you can wrap it in a function and apply to faction change also, no need to broadcaast antything as it clears our current target and targets currently attacking us :)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -5777,6 +5777,24 @@ bool Unit::AttackStop()
     return true;
 }
 
+void Unit::ValidateAttackersAndOwnTarget()
+{
+    // iterate attackers
+    AttackerSet toRemove;
```
ariel-: Change this one to a vector please

xinef1: done

PR #18977, opened 2017-01-26
Core/Auth: Use QueryCallbackProcessor insted boost::optional.

r4d1sh:
Fixes build for boost 1.55.0

**Changes proposed:**

-  Failure to use boost::optional because it not support move semantic on 1.55.0
- Add IsEmpty() method to QueryCallbackProcessor

### DIFF
--- a/src/server/authserver/Server/AuthSession.cpp
+++ b/src/server/authserver/Server/AuthSession.cpp
@@ -20,6 +20,7 @@
 #include "Log.h"
 #include "AuthCodes.h"
 #include "Database/DatabaseEnv.h"
+#include "QueryCallback.h"
 #include "SHA1.h"
 #include "TOTP.h"
 #include "openssl/crypto.h"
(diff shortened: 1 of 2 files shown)


jackpoz:
is boost 1.55 needed for a particular reason ?

Tonghost:
Debian still ships 1.55

[More info](https://github.com/TrinityCore/TrinityCore/commit/4c27203c8f36dd2a5df0a4ae69fbdc4c9140b29d#diff-1bfb148491dedaab3a6311bec32f7e7eR172)

Naios:
There are other standalone implementations of `optional` which we could use.

Shauren:
Lets not.

src/server/authserver/Server/AuthSession.h
```diff
@@ -24,6 +24,7 @@
 #include "Socket.h"
 #include "BigNumber.h"
 #include "QueryCallback.h"
```
Shauren: remove QueryCallback.h

Shauren (changes requested):

src/server/authserver/Server/AuthSession.cpp
```diff
@@ -288,7 +287,7 @@ bool AuthSession::HandleLogonChallenge()
     std::string login((const char*)challenge->I, challenge->I_len);
     TC_LOG_DEBUG("server.authserver", "[AuthChallenge] '%s'", login.c_str());
 
-    if (_queryCallback)
+    if (!_queryProcessor.IsEmpty())
```
Shauren: These checks are not needed anymore since _status guards login sequences and you can only attempt it once per connection

Issue #18241, opened 2016-11-12
gameobject commands not working


Rochet2:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:**
The gameobject commands dont seem to be working properly together.
The gameobject target gives links that contain the gameobject DB guid.
Because of the new guid changes .gps command however searches with the ingame guid (So a gameobject that has DB guid X will have ingame guid Y which has a dynamically generated guid counter).

I guess the root causes here are that there are two guids and both are needed by the commands.
Also some commands need the entry for which there is a completely another link type.
It would be probably nice to have something like `|color|Hgameobject:entry:dbguid:ingameguid|h[name]|h|r` so one link can be used for everything and it would work/not work depending on if the info needed is available in the link.

Also .gob add command does not work for some reason. The gameobjects dont save up in DB.

**Current behaviour:**
Gameobject link from .gob tar gives DB guids. However .gps command uses core generated guids to search. The error message says player not found and should say gameobject not found.
The .gob add command does not save gameobjects to DB.

**Expected behaviour:**
Gameobject link from .gob tar should probably contain the ingame guid as well. The gps command needs to know the ingame guid to search properly (uniquely identify the gameobject) It also needs the DB guid or gameobject entry to make the guid to search with.
The error message should vary depending on what arguments were provided. For example when giving a gameobject link, it should say that no gameobject was found.
The spawned gameobjects should save to db.

**Steps to reproduce the problem:**

1. Try spawning a gameobject with .gob add 35
2. See that it is there
3. Try target it with .gob target and see that it fails to target the gameobject
4. See the gameobject table in DB and see that there are no spawns there.
5. Restart the server and go to where you spawned the object and see that it isnt there.
6. Conclusion: No gameobjects are saved and they cannot be targeted. Targeting fails because it searches the objects from DB where they dont exist.
1. Go to an existing gameobject.
2. Use .gob target on it and it should succeed because it is in DB
3. Use .gps <link of the gameobject>
4. See that it fails and the error message incorrectly says "Cannot find player"
5. Conclusion: the gps command cannot be used on the object links (even though it is intended to)

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  ed2cd6b34f849f9cc9adb86318457155e47f7314

**TDB version:**  62

**Operating system:**  windows

[...]

Issue #16733, opened 2016-03-05
[3.3.5] Core/Command: .npc delete guid and wp show off not working properly


Magnifikator:
d55d32fa0f7a

Both commands don't work properly because of the same fault. 
.wp show off doesn't delete the visual waypoint marker (because it can't find the guid)
.npc del guid also doesn't work because of the same reason.

Looking to the code:

cs_npc.cpp line 492:
unit = handler->GetSession()->GetPlayer()->GetMap()->GetCreature(ObjectGuid(HighGuid::Unit, cr_data->id, lowguid));

unit is always 0 (null)

cs_wp.cpp line 1025:
Creature\* creature = handler->GetSession()->GetPlayer()->GetMap()->GetCreature(ObjectGuid(HighGuid::Unit, VISUAL_WAYPOINT, guid));

creature is always 0 (null)

Unfortunately I couldn't find exactly where the problem is. ObjectGuid() seems to be correct. It contains the correct guid. Maybe it's GetCreature().

Magnifikator:
Ok, so far as I know it has something to do with using object lowguid instead of object counter at ObjectGuid()

Shauren:
Ok I know what is wrong in npc delete, still looking at wp show

Magnifikator:
cs_wp.cpp line 1024
ObjectGuid::LowType guid = fields[0].GetUInt32();

That's also a database guid and not a counter. Same issue same reason.

I just don't know if you should give another parameter or change the ObjectGuid() method.

Magnifikator:
@Shauren 

Need more information ?

Rochet2:
> Need more information ?

he needs a cloning machine.

Magnifikator:
What does that mean ... a cloning machine ?   :-)

tkrokli:
@Magnifikator : I don't know if you have understood what Rochet2 meant yet, but I will explain:
"Shauren has got so much to do, he must be cloned to become available enough for our questions."

Magnifikator:
No, I haven't, good to know, thanks :-)

Aokromes:
117d1c3

PR #18863, opened 2017-01-16
Scripts/Commands: Fix trinity_string usage and enhance commands

Rochet2:
[//]: # (***************************)
[//]: # (** FILL IN THIS TEMPLATE **)
[//]: # (***************************)

**Changes proposed:**

~~Waiting for https://github.com/TrinityCore/TrinityCore/pull/18978 to merge first!~~

Tries to fix trinity_string related format differences and more:
- Change many %d to %u
- Change many %i to %u
- Change many %u and %llu to %s and change uint64 values to std::to_string(val).c_str()
- Change int64 to string before formatting and change all format characters for them to %s
- Change uint64 to string before formatting and change all format characters for them to %s
- Change many atoi to atoul when trying to get an uint32 value from user
- Change many atoi to atoul when trying to get an uint32 value from user
- Add .c_str() call to convert std::string to const char* before formatting
- Remove format characters that were not used in code (in DB there was %s but nothing in c++)
- Remove arguments that had no format characters (in c++ there was something formatted, but in DB there was no format character like %s)
- Change types in code to match the surrounding code better (for example return type was uint32 and returned variable was int)
- Remove unnecessary `(char*)` casts for `const char*` variables
- `.modify faction` now works without arguments as described in `.help`
- LANG_COMMAND_DELOBJREFERCREATURE (274) no longer says that a nonexistant creature references it, because owner can be more than a creature. Instead it says that a nonexistant owner references it. Printed guid should tell what type the owner is.
- Change __some__ usages of snprintf to use Trinity::StringFormat instead
- Fix typos (IE: Wypoint scipt -> Waypoint script)
- Convert __some__ `container.size()` to string before formatting
- Change __some__ std::strtoull to atoull

**Target branch(es):** 3.3.5/master

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes #  (insert issue tracker number)
Closes #19127
Closes #20023

**Tests performed:** (Does it build, tested in-game, etc.)
Does build.

**Known issues and TODO list:** (add/remove lines as needed)

- [x] Fix the code (the IF condition here looks funny) and the message (the guids are in wrong order) and the types in here: https://github.com/TrinityCore/TrinityCore/pull/18978#discussion_r100573183 (usage of LANG_COMMAND_DELOBJREFERCREATURE in cs_gobject.cpp)
- [x] Fix the PR after https://github.com/TrinityCore/TrinityCore/pull/18978 is merged.

PR #18278, opened 2016-11-19
Scripts/Commands: Fix guid usage and enhance commands

Rochet2:
[//]: # (***************************)
[//]: # (** FILL IN THIS TEMPLATE **)
[//]: # (***************************)

**Changes proposed:**

- Fix targeting gameobjects and creatures for .dist and .gps. The old code used DB guid to search from core generated guid lists.
- Fix some error messages telling no player found even if creatures and gameobjects and players were searched
- Change atoi to atoul when receiving lowguids as string or other uint32 values
- Use ObjectGuid::LowType instead of uint32 for lowguid type
- Allow .gob info to take in gameobject guid link as well as entry link in addition to gameobject entry and update the documentation of the command
- Change .gob delete to post the spawnid (dbguid) instead of the ingame generated guid (which is basically never seen or used by the player in any other commands that I see)
- Allow spawntimesecs to be negative when spawning a gameobject
- Disable searching by ingame spawn id in commands. (it makes little sense for a user to provide a number which is then attempted to be used as DB guid and non DB guid when both could have a match)
- Update .gob turn documentation to tell about possibility to provide orientation in command
- Correct the types in .gob add command result trinity_string
- Correct the creature search for .npc del <lowguid>
- Change .event <id> to .event info <id>
- Add .go offset command
- Remove the .go command for empty string (no subcommand given) because it just used .go xyz
- Extend .gob info to show size, faction, flags and model dimensions
- Remove some unnecessary casts
- Document .wp show better (first and last were not documented)
- Insert spawnid to .wp show info error message
- Fix spawntime printing for .gobject target
- Fix guid targetting for .npc set movetype <guid> <type>
- Fix query by spawnid for .wp show info
- Fix deleting of existing waypoints when twice doing the command .wp show on
- Fix deleting of existing waypoints when doing the command .wp show off
- Change wpguid column in DB and in core to uint32, which is what the spawnid type is in core and db.

**Target branch(es):** 3.3.5/master

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes #  (insert issue tracker number)
Closes https://github.com/TrinityCore/TrinityCore/issues/18241
Closes https://github.com/TrinityCore/TrinityCore/issues/16733

**Tests performed:** (Does it build, tested in-game, etc.)

- [x] Does it build?
- [x] Tested ingame? Most important parts tested. For more see TODO

**Known issues and TODO list:** (add/remove lines as needed)
The PR was made for 3.3.5, but probably also applies to master branch as well.
[...]

PR #18978, opened 2017-01-26
Scripts/Commands: Fix guid usage and enhance commands

Rochet2:

This is a cherry pick of https://github.com/TrinityCore/TrinityCore/pull/18278 and is needed to finish up https://github.com/TrinityCore/TrinityCore/issues/18863

**Changes proposed:**

- Fix targeting gameobjects and creatures for .dist and .gps. The old code used DB guid to search from core generated guid lists.
- Fix some error messages telling no player found even if creatures and gameobjects and players were searched
- Change atoi to atoul when receiving lowguids as string or other uint32 values
- Use ObjectGuid::LowType instead of uint32 for lowguid type
- Allow .gob info to take in gameobject guid link as well as entry link in addition to gameobject entry and update the documentation of the command
- Change .gob delete to post the spawnid (dbguid) instead of the ingame generated guid (which is basically never seen or used by the player in any other commands that I see)
- Allow spawntimesecs to be negative when spawning a gameobject
- Disable searching by ingame spawn id in commands. (it makes little sense for a user to provide a number which is then attempted to be used as DB guid and non DB guid when both could have a match)
- Update .gob turn documentation to tell about possibility to provide orientation in command
- Correct the types in .gob add command result trinity_string
- Correct the creature search for .npc del <lowguid>
- Change .event <id> to .event info <id>
- Add .go offset command
- Remove the .go command for empty string (no subcommand given) because it just used .go xyz
- Extend .gob info to show size, faction, flags and model dimensions
- Remove some unnecessary casts
- Document .wp show better (first and last were not documented)
- Insert spawnid to .wp show info error message
- Fix spawntime printing for .gobject target
- Fix guid targetting for .npc set movetype <guid> <type>
- Fix query by spawnid for .wp show info
- Fix deleting of existing waypoints when twice doing the command .wp show on
- Fix deleting of existing waypoints when doing the command .wp show off
- Change wpguid column in DB and in core to uint32, which is what the spawnid type is in core and db.

- [x] master

Closes https://github.com/TrinityCore/TrinityCore/issues/18241
Closes https://github.com/TrinityCore/TrinityCore/issues/16733

- [x] Does it build?


src/server/game/Accounts/RBAC.h
```diff
@@ -755,7 +755,8 @@ enum RBACPermissions
     RBAC_PERM_COMMAND_SCENE_CANCEL                           = 848,
     RBAC_PERM_COMMAND_LIST_SCENES                            = 849,
     RBAC_PERM_COMMAND_RELOAD_SCENE_TEMPLATE                  = 850,
-    RBAC_PERM_COMMAND_GO_OFFSET                              = 851,
+    RBAC_PERM_COMMAND_RELOAD_AREATRIGGER_TEMPLATE            = 851,
+    RBAC_PERM_COMMAND_GO_OFFSET                              = 852,
```
Rochet2: hmm, need to fix the SQLs for this too before merge

Rochet2:
Note me if I need to do something about this before merging.

Merging this will allow us to fix the master/3.3.5 rbac permission desync (before it gets any worse) and continue to work on fixing commands for both versions.

Aokromes:
there is no desync anymore.

Rochet2:
I see, it was fixed by ariel- on https://github.com/TrinityCore/TrinityCore/commit/9e65fead38c486ec8a971a09a95ffae912943a83

src/server/scripts/Commands/cs_gobject.cpp
```diff
@@ -373,7 +363,7 @@ class gobject_commandscript : public CommandScript
             Unit* owner = ObjectAccessor::GetUnit(*handler->GetSession()->GetPlayer(), ownerGuid);
             if (!owner || !ownerGuid.IsPlayer())
             {
-                handler->PSendSysMessage(LANG_COMMAND_DELOBJREFERCREATURE, ownerGuid.ToString().c_str(), object->GetGUID().ToString().c_str());
+                handler->PSendSysMessage(LANG_COMMAND_DELOBJREFERCREATURE, ownerGuid.GetCounter(), guidLow);
```
Shauren: bug here (both in existing code and proposed)
db entry 274
>Game Object (GUID: %u) is referenced by nonexistent creature %u in GO list, can't be deleted.

counter/lowguid is uint64, %u is not sufficient
always prefer printing guid using ToString(), not its low part

Shauren (changes requested):

src/server/scripts/Commands/cs_gobject.cpp
Rochet2: This is fixed by the other PR that tackles the types in general and fixes most if not all such differences in commands. https://github.com/TrinityCore/TrinityCore/pull/18863

However it uses tostring on the lowguid instead of the full guid. I guess in this case the correct way is as you say to use the tostring because its not a gameobject link.

But in general most of the code for commands gives out the lowguids instead of full guids since the commands use lowguid to refer to objects.
Hmm.

tkrokli:
Any other changes needed in this PR before merging?

Rochet2:
Couldnt find other things to do myself atm.

Shauren (approved):

PR #18999, opened 2017-01-28
Core/Database: Prepared statement parameter preallocation

ariel-:

**Changes proposed:**

-  Rename CheckValidIndex -> AssertValidIndex
-  Cached prepared size on the worker pool as it's shared among all connections (original implementation cached this info on every connection, so there were multiple instances, needlessly wasting memory)
 - Removed cached query data map, only was used for logging and sql query is now saved on the MySQLPreparedStatement proper
- Cached size is now passed on to constructor, added assertions to check out of range parameter indices

Based on https://github.com/Naios/TrinityCore/commit/7d13a6c6a3e95bf97ad41278176d7a54e86c9921

**Target branch(es):** 3.3.5

**Issues addressed:** Ref #14274

**Tests performed:** Builds and works as intended

### DIFF
--- a/src/server/database/Database/MySQLConnection.h
+++ b/src/server/database/Database/MySQLConnection.h
@@ -35,17 +35,6 @@ enum ConnectionFlags
     CONNECTION_BOTH = CONNECTION_ASYNC | CONNECTION_SYNCH
 };
 
-struct PreparedStatementInfo
-{
-    PreparedStatementInfo(std::string const& queryString, ConnectionFlags connFlags) :
-        Query(queryString), Flags(connFlags) { }
-
-    std::string const Query;
-    ConnectionFlags const Flags; // sync/async
-};
-
-typedef std::unordered_map<uint32 /*index*/, PreparedStatementInfo> PreparedStatementContainer;
-
 struct TC_DATABASE_API MySQLConnectionInfo
 {
     explicit MySQLConnectionInfo(std::string const& infoString)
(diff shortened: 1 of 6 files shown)


src/server/database/Database/MySQLConnection.h
```diff
@@ -35,6 +35,17 @@ enum ConnectionFlags
     CONNECTION_BOTH = CONNECTION_ASYNC | CONNECTION_SYNCH
 };
 
+struct PreparedStatementInfo
+{
+    PreparedStatementInfo(std::string const& queryString, ConnectionFlags connFlags) :
+        Query(queryString), Flags(connFlags) { }
+
+    std::string const Query;
+    ConnectionFlags const Flags; // sync/async
```
ariel-: On a second thought, stored connection flags for the Prepared statement are not used at all.

Query string could be moved to MySQLPreparedStatement class, and remove this container entirely, so there is only one container, and it's a vector so it's fast

src/server/database/Database/PreparedStatement.cpp
```diff
@@ -19,202 +19,172 @@
...
+    ASSERT(index < statement_data.size());
     statement_data[index].type = TYPE_NULL;
 }
 
-MySQLPreparedStatement::MySQLPreparedStatement(MYSQL_STMT* stmt) :
-m_stmt(NULL),
-m_Mstmt(stmt),
-m_bind(NULL)
+MySQLPreparedStatement::MySQLPreparedStatement(MYSQL_STMT* stmt, std::string const& queryString) :
+m_stmt(nullptr), m_Mstmt(stmt), m_bind(nullptr), m_queryString(queryString)
```
Naios: `std::string const& queryString)` -> `std::string queryString)`
and
`m_queryString(queryString)` -> `m_queryString(std::move(queryString))`

Naios (approved):
Seems good, nice that you finally finished my changes :-)

PR #19003, opened 2017-01-29
Fixed durability problems for wrapped item

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles


src/server/game/Entities/Item/Item.cpp
```diff
@@ -457,7 +457,9 @@ bool Item::LoadFromDB(ObjectGuid::LowType guid, ObjectGuid owner_guid, Field* fi
     SetUInt32Value(ITEM_FIELD_DURABILITY, durability);
     // update max durability (and durability) if need
     SetUInt32Value(ITEM_FIELD_MAXDURABILITY, proto->MaxDurability);
-    if (durability > proto->MaxDurability)
+
+    // do not overwrite durability for wrapped items
```
Shauren: why?

xinef1: because proto durability value of gift (gifts overwrites entries for original items) is 0

PR #19004, opened 2017-01-29
Build Fix

Jildor:

**Compile Error:**
https://gist.github.com/Jildor/92311667b50df04b0de2d3418521e75b

Caused by:
https://github.com/TrinityCore/TrinityCore/commit/684a5fd3f1895703a52cff7e7f762883c74c5aba#diff-93a9b3c0592a8842b0dbb5020df4b45aL26

https://github.com/TrinityCore/TrinityCore/commit/62b62ddd63942090b23f8881f2c6ad93d0a251dc#diff-212788e7639954c00075a15a422da430L29

**Target branch(es):**

- [x] 3.3.5
- [ ] master

Debian

### DIFF
--- a/src/server/authserver/Server/AuthSession.h
+++ b/src/server/authserver/Server/AuthSession.h
@@ -23,6 +23,7 @@
 #include "ByteBuffer.h"
 #include "Socket.h"
 #include "BigNumber.h"
+#include "QueryResult.h"
 #include "QueryCallbackProcessor.h"
 #include <memory>
 #include <boost/asio/ip/tcp.hpp>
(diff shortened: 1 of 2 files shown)


src/server/authserver/Server/AuthSession.h
```diff
@@ -23,6 +23,7 @@
 #include "ByteBuffer.h"
 #include "Socket.h"
 #include "BigNumber.h"
+#include "QueryCallback.h"
```
Shauren: Including QueryCallback is not the proper way to do this - it is not used anywhere in this file. I have a strong feeling that the error is something else but including QueryCallback.h indirectly includes the required header.

Shauren (changes requested):

src/server/authserver/Server/AuthSession.h
Jildor: Done ;)

PR #19011, opened 2017-01-30
Core/Misc: do not kick GameMasters from instances

### LINKED ISSUE #18228: [3.3.5] GMs kicked from raid when server restart

**Description:** When you restart server with GM on, when server back, sometimes you are kicked from instance.

**Current behaviour:** You are kiked from instances when server restart.

**Expected behaviour:** You stay inside of instance.

1.  Put a character with some instance.
2. Restart server.
3.  If you stay in instance, repeat steps 1 and 2.

This commit should fix it: https://github.com/TrinityCore/TrinityCore/commit/9573ec52f0ea2fbda0fdcac1bd9ed2617d359338. But sometimes fail and show message in console:

> Player::LoadFromDB: Player 'Keader' (GUID Full: 0x0000000000000001 Type: Player Low: 1) Map: 631, X: -59.841766, Y: 2210.268555, Z: 27.902609, O: 3.149650. Areatrigger not found.

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  70a867531d6c6dba13dcd56a44db855b39aa6471

**TDB version:**  

**Operating system:**  

sirikfoll:
**Changes proposed:**

-  Even when GameMaster mode is deactivated, do not remove the GM from the instance (happens after a server shutdown/restart)
-  Always allow GMs to enter raid instances without GM mode active and/or raid group

- [x] 3.3.5

**Issues addressed:** Closes #18228

**Tests performed:** Built and tested with a GM character

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -21589,7 +21589,7 @@ uint32 Player::GetMaxPersonalArenaRatingRequirement(uint32 minarenaslot) const
 void Player::UpdateHomebindTime(uint32 time)
 {
     // GMs never get homebind timer online
-    if (m_InstanceValid || IsGameMaster())
+    if (m_InstanceValid || CanBeGameMaster())
     {
         if (m_HomebindTimer)                                 // instance valid, but timer not reset
         {
--- a/src/server/game/Maps/MapInstanced.cpp
+++ b/src/server/game/Maps/MapInstanced.cpp
@@ -153,7 +153,11 @@ Map* MapInstanced::CreateInstanceForPlayer(const uint32 mapId, Player* player, u
             if (loginInstanceId) // if the player has a saved instance id on login, we either use this instance or relocate him out (return null)
             {
                 map = FindInstanceMap(loginInstanceId);
-                return (map && map->GetId() == GetId()) ? map : nullptr; // is this check necessary? or does MapInstanced only find instances of itself?
+                if (map && map->GetId() == GetId()) // is this check necessary? or does MapInstanced only find instances of itself?
+                    return map;
+
+                if (!player->CanBeGameMaster())
+                    return nullptr;
             }
 
             InstanceGroupBind* groupBind = nullptr;
--- a/src/server/game/Maps/MapManager.cpp
+++ b/src/server/game/Maps/MapManager.cpp
@@ -146,7 +146,7 @@ Map::EnterState MapManager::PlayerCannotEnter(uint32 mapid, Player* player, bool
         return Map::CANNOT_ENTER_DIFFICULTY_UNAVAILABLE;
 
     //Bypass checks for GMs
-    if (player->IsGameMaster())
+    if (player->CanBeGameMaster())
         return Map::CAN_ENTER;
 
     char const* mapName = entry->name[player->GetSession()->GetSessionDbcLocale()];


FireEmerald:
Shouldn't a GM handled like a normal player if he doesn't has GM mode on?

Keader:
Old fix about it: 9573ec52f0ea2fbda0fdcac1bd9ed2617d359338

@FireEmerald it's not always true. Has some RBAC checks to skip some stuffs if you are/can be GM.

sirikfoll:
That's up for discussion, I'm just leaving the PR here to see what happens ^^

jackpoz:
I'd prefer to keep a GM char with .gm off to behave as a normal player, that's the whole point of .gm on/off

sirikfoll:
What about keeping just this part? "Even when GameMaster mode is deactivated do not remove the GM from the instance(happens after a server shutdown/restart)"

Because of 5f917e2, GMs with .gm off are kept in the instance at login(if there is a instance lock active), it would just improve it a bit, keeping them in instance even when there is no instance lock.

Aokromes:
I will try to see what happens on ptr when a restart is send and i stay inside instance :)

Keader:
Is annoying reenter in instance everytime that you restart server when you are scripting something.

Krudor:
If I understand the problem correct, we're getting kicked out of the instance because we're not GM's.
So the underlying problem is that GM state is not saved and resets upon relogging.
How about saving the GM state and retrieve it upon login? This and other possibly related problems should go away.

Kittnz:
Same with all other gm things, such as invisibility

jackpoz:
> If I understand the problem correct, we're getting kicked out of the instance because we're not GM's

nope, that's handled by https://github.com/TrinityCore/TrinityCore/commit/5f917e2286809d7aafa8871be38042ae7966c26b . You are getting kicked out because of you don't have a valid InstanceSave on login, see https://github.com/TrinityCore/TrinityCore/pull/19011/files#diff-f84464597f685f99be0d1e1f7d09f3e4L156

Aokromes:
well, i had my char stuck inside one ptr instance for weeks, and even server restarted multiple times it continued to be there. so imho we can merge this.

Krudor:
I don't like this hack fix, should be solved another way.

jackpoz:
> well, i had my char stuck inside one ptr instance for weeks, and even server restarted multiple times it continued to be there. so imho we can merge this.

are you a GameMaster in PTR ?

Aokromes:
na, but if players aren't kicked gms they must do the same.

Krudor:
I think you're being kicked to the graveyard, which is a behavior they haven't changed in a long time.
They just changed some graveyard locations to be the entrance location for instances and raids.

Krudor:
If it can't get the graveyard, you're kicked to hearthstone location.
Appears to happen when you die inside an instance, log out and then log back in after the instance has soft reset.

jackpoz:
> na, but if players aren't kicked gms they must do the same.

but we kick both players and GMs and this PR is making an exceptions for GMs. This PR doesn't implement the PTR behavior you described.

Treeston:
As discussed above - GMs with `.gm off` should be treated the same as regular players for most intents and purposes, this included.

If you fix the underlying behavior (again, as discussed above), please open a separate PR.

PR #19023, opened 2017-02-01
Core/Player: UpdateCraftSkill

DoctorKraft:

**Changes proposed:**

- Adding new system where profession skill give more than 1 skill point
For example : http://www.wowhead.com/spell=167950/research-warbinders-ink (Rewards 3 skill points)
-  Rewrite for better understanding

- [x] master

**Issues addressed:** Closes #  Unk

**Tests performed:** Build & Tested in-game

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -5245,10 +5245,10 @@ bool Player::UpdateCraftSkill(uint32 spellid)
             uint32 craft_skill_gain = _spell_idx->second->NumSkillUps * sWorld->getIntConfig(CONFIG_SKILL_GAIN_CRAFTING);
 
             return UpdateSkillPro(_spell_idx->second->SkillLine, SkillGainChance(SkillValue,
-                 _spell_idx->second->TrivialSkillLineRankHigh,
-                 (_spell_idx->second->TrivialSkillLineRankHigh + _spell_idx->second->TrivialSkillLineRankLow)/2,
-                 _spell_idx->second->TrivialSkillLineRankLow),
-                 craft_skill_gain);
+                _spell_idx->second->TrivialSkillLineRankHigh,
+                (_spell_idx->second->TrivialSkillLineRankHigh + _spell_idx->second->TrivialSkillLineRankLow)/2,
+                _spell_idx->second->TrivialSkillLineRankLow),
+                craft_skill_gain);
         }
     }
     return false;
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -5242,13 +5242,13 @@ bool Player::UpdateCraftSkill(uint32 spellid)
                     LearnSpell(discoveredSpell, false);
             }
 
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -5243,12 +5243,24 @@ bool Player::UpdateCraftSkill(uint32 spellid)
...
             uint32 craft_skill_gain = sWorld->getIntConfig(CONFIG_SKILL_GAIN_CRAFTING);
+            int skill_gain_chance = SkillGainChance(SkillValue, _spell_idx->second->TrivialSkillLineRankHigh, (_spell_idx->second->TrivialSkillLineRankHigh + _spell_idx->second->TrivialSkillLineRankLow) / 2, _spell_idx->second->TrivialSkillLineRankLow);
 
-            return UpdateSkillPro(_spell_idx->second->SkillLine, SkillGainChance(SkillValue,
-                _spell_idx->second->TrivialSkillLineRankHigh,
-                (_spell_idx->second->TrivialSkillLineRankHigh + _spell_idx->second->TrivialSkillLineRankLow)/2,
-                _spell_idx->second->TrivialSkillLineRankLow),
-                craft_skill_gain);
+            if (_spell_idx->second && _spell_idx->second->NumSkillUps > craft_skill_gain)
+                craft_skill_gain = _spell_idx->second->NumSkillUps;
```
joschiwald: could it be also `_spell_idx->second->NumSkillUps * sWorld->getIntConfig(CONFIG_SKILL_GAIN_CRAFTING)`?

DoctorKraft: Sure ! CONFIG_SKILL_CHANCE_ORANGE better no ? Because all skill with NumSkillUps > 1 is orange

src/server/game/Entities/Player/Player.cpp
```diff
@@ -5243,12 +5243,24 @@ bool Player::UpdateCraftSkill(uint32 spellid)
...
+
+            auto skillId = _spell_idx->second->SkillLine;
+            if (UpdateSkillPro(skillId, skill_gain_chance, craft_skill_gain))
+            {
+                if (SkillLineEntry const* skillEntry = sSkillLineStore.LookupEntry(skillId))
+                {
+                    if (uint32 parentSkillId = skillEntry->ParentSkillLineID)
+                        return UpdateSkillPro(parentSkillId, skill_gain_chance, craft_skill_gain);
```
Shauren: This is for child skills like MoP cooking so this is no good. You need to give guaranteed skill value equal to highest child skilline, not grant a separate roll (also you need to make sure.
I suggest only doing the multiple skillups in this PR as the title says.

Shauren (changes requested):

src/server/game/Entities/Player/Player.cpp
DoctorKraft: Ok thanks, i have missed this. It's a very old code ! Fixed :)

DoctorKraft:
I have updated, say me if you prefer with old codestyle

Shauren (approved):

PR #19030, opened 2017-02-02
Moved CharacterInfo out of world to separate class

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles

Moved CharacterInfo out of world to separate class
Moved code related to CharacterInfo, placed elsewhere in the core to the new class

### DIFF
--- a/src/server/game/Cache/CharacterCache.h
+++ b/src/server/game/Cache/CharacterCache.h
@@ -66,8 +66,8 @@ class TC_GAME_API CharacterCache
         uint32 GetCharacterAccountIdByGuid(ObjectGuid guid) const;
         uint32 GetCharacterAccountIdByName(std::string const& name) const;
         uint8 GetCharacterLevelByGuid(ObjectGuid guid) const;
-        ObjectGuid::LowType CharacterCache::GetCharacterGuildIdByGuid(ObjectGuid guid) const;
-        uint32 CharacterCache::GetCharacterArenaTeamIdByGuid(ObjectGuid guid, uint8 type) const;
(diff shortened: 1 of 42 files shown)


src/server/game/Cache/CharacterCache.h
```diff
@@ -0,0 +1,75 @@
...
+    uint32 AccountId;
+    uint8 Class;
+    uint8 Race;
+    uint8 Sex;
+    uint8 Level;
+    ObjectGuid::LowType GuildId;
+    uint32 ArenaTeamId[3];
+};
+
+typedef std::unordered_map<std::string, ObjectGuid> CharacterGuidByNameContainer;
```
Shauren: How about changing this map to <string, CharacterCacheEntry*> pointing inside the other container to avoid double lookups in *ByName methods? (and add Guid to CharacterCacheEntry)

Shauren: also, you don't need these typedefs at all (especially in this header! you don't provide any api returning iterator directly so there is no need for them being here)

src/server/game/Cache/CharacterCache.h
```diff
@@ -0,0 +1,75 @@
...
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef __CHARACTERCACHE_H
+#define __CHARACTERCACHE_H
+
+#include "Common.h"
```
Shauren: If you remove the typedefs from this file you can reduce the include set to just
"Define.h" (difference between Define and Common is huge right now)
"ObjectGuid.h"
<string>

PR #19036, opened 2017-02-04
Core/Spells: corrected spell fail result calculations and added some new

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles

Added SPELL_FAILED_NOTHING_TO_STEAL error
Corrected Dispel check in sanctuary area
Added root check for SPELL_EFFECT_JUMP and SPELL_EFFECT_JUMP_DEST


tkrokli:
Would it be practical to add a new function `IsPvPAttackable` the same way as `IsInSanctuary`
```cpp
bool IsPvPAttackable() const { return HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_PVP_ATTACKABLE); }
```
... or is it a waste of time?

xinef1:
Sure, IsPvPAttackable + SetPvPAttackable(bool state) but this should be done in other PR imo

PR #19037, opened 2017-02-04
If roll winner is unable to store loot, block the won item for him

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles


Keader:
It parcial fix #7672?

PR #19039, opened 2017-02-04
Corrected execution time calculation for dst targets

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -811,7 +811,8 @@ void Spell::SelectSpellTargets()
         }
         else if (m_spellInfo->Speed > 0.0f)
         {
-            float dist = m_caster->GetDistance(*m_targets.GetDstPos());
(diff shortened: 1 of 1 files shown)


joschiwald:
there are more places where this calculation is done

xinef1:
This is the only place where that kind of calculation is done

joschiwald:
Spell::AddUnitTarget
Spell::AddGOTarget

are also affected by caster size

xinef1:
That is something else, this is delayTime calculation to destination point

joschiwald:
why is the delay to destination not affected by caster size, while delay to units or gameobjects is?

xinef1:
Because this matches clients animation (check Malleable Goo cast by Profesor Putricade)

Treeston:
This is flat out incorrect for certain spells (Halion's Meteor Strike (74637) among others).

There is more to this than "hitbox size is not taken into account for ground targeting" (if only because that'd be really weird).

PR #19048, opened 2017-02-05
Core/Spells: Remade trajectory target selection

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compile, some ingame tests

Remade trajectory target selection, created separate target category (TARGET_SELECT_CATEGORY_TRAJ). Remodeled all collision with trajectory calculations, at first all not in line objects are not even included in the targetlist thanks to WorldObjectSpellTrajTargetCheck. Fixed existing bestDist range problem, many times trajectory spells have too short range (which should be in such case taken from triggered spell), limited this just in case we have trigger with 50000yds. Simplified collision calculations, we only have to check if trajectory is on appropriate height to collide with unit.

Added possibility to apply conditions to trajectory spells
Allow to apply conditions to chain targets (they are treated as area)
Properly recalculate delay time if CMSG_UPDATE_PROJECTILE_POSITION is received


src/server/game/Spells/Spell.cpp
```diff
@@ -2915,8 +2861,8 @@ void Spell::prepare(SpellCastTargets const* targets, AuraEffect const* triggered
         m_triggeredByAuraSpell  = triggeredByAura->GetSpellInfo();
 
     // create and add update event for this spell
-    SpellEvent* Event = new SpellEvent(this);
-    m_caster->m_Events.AddEvent(Event, m_caster->m_Events.CalculateTime(1));
+    _spellEvent = new SpellEvent(this);
```
joschiwald: that leads to dangling pointer after the event is executed and deleted by event processor

xinef1: When SpellEvent is deleted, it deletes the Spell itself

joschiwald: good

src/server/game/Spells/Spell.cpp
```diff
@@ -1516,12 +1531,14 @@ void Spell::SelectImplicitTrajTargets(SpellEffIndex effIndex)
...
+    Position srcPos = *m_targets.GetSrcPos();
+    srcPos.SetOrientation(m_caster->GetOrientation());
+    float srcToDestDelta = m_targets.GetDstPos()->m_positionZ - srcPos.m_positionZ;
 
     std::list<WorldObject*> targets;
-    Trinity::WorldObjectSpellTrajTargetCheck check(dist2d, m_targets.GetSrcPos(), m_caster, m_spellInfo);
+    Trinity::WorldObjectSpellTrajTargetCheck check(dist2d, &srcPos, m_caster, m_spellInfo, targetType.GetCheckType(), m_spellInfo->Effects[effIndex].ImplicitTargetConditions);
     Trinity::WorldObjectListSearcher<Trinity::WorldObjectSpellTrajTargetCheck> searcher(m_caster, targets, check, GRID_MAP_TYPE_MASK_ALL);
-    SearchTargets<Trinity::WorldObjectListSearcher<Trinity::WorldObjectSpellTrajTargetCheck> > (searcher, GRID_MAP_TYPE_MASK_ALL, m_caster, m_targets.GetSrcPos(), dist2d);
+    SearchTargets<Trinity::WorldObjectListSearcher<Trinity::WorldObjectSpellTrajTargetCheck> > (searcher, GRID_MAP_TYPE_MASK_ALL, m_caster, &srcPos, dist2d);
```
ariel-: SearchTargets still expects 3D distance to be passed

1.

xinef1: Nope :D https://github.com/TrinityCore/TrinityCore/pull/19048/commits/a24859cba59ed1093ae085e85d41f0024e157107#diff-952c711803610aa78110e4f92ff2516aR7719

ariel-: yup: https://github.com/xinef1/TrinityCore/blob/a24859cba59ed1093ae085e85d41f0024e157107/src/server/game/Spells/Spell.cpp#L1748

ariel-: Brain fart

ariel-:
I purposedly left out the IoC fix, so the PR didn't have an unrelated change :P, I'm leaving that for the battleground rewrites.

PR #19054, opened 2017-02-05
Core/Spells: Corrected aura SPELL_AURA_MOD_CASTING_SPEED_NOT_STACK handling for auras with very high values

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles

### DIFF
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -3911,6 +3911,13 @@ void AuraEffect::HandleModCastingSpeed(AuraApplication const* aurApp, uint8 mode
 
     Unit* target = aurApp->GetTarget();
 
+    // Do not apply such auras in normal way
+    if (GetAmount() >= 1000)
+    {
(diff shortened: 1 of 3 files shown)


src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -3868,6 +3868,13 @@ void AuraEffect::HandleModCastingSpeed(AuraApplication const* aurApp, uint8 mode
 
     Unit* target = aurApp->GetTarget();
 
+    // Do not apply such auras in normal way
+    if (GetAmount() >= 1000)
+    {
+        target->SetInstantCast(apply);
```
ariel-: What happens if 2 such auras are applied, and then one is removed?

xinef1: It's not possible in 3.3.5 (very few of such auras). If theoretically this could happen, we would need to loop speed_not_stack auras to see if there is other such aura and prevent unsetting

PR #19055, opened 2017-02-05
Core/Object: Fix LoS issue

Golrag:
**Changes proposed:**
- Limit the ObjectSize radius calculation so that we do not check for the point behind the caster when caster is in ObjectSize of target. (see https://github.com/TrinityCore/TrinityCore/pull/15807#issuecomment-268077802)

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes https://github.com/TrinityCore/TrinityCore/pull/15807#issuecomment-268077802

**Tests performed:** 
- Builds
- Tested in game

Any feedback is welcome :)


Keader:
If has no oppositions i will merge it in 24h

PR #19056, opened 2017-02-05
[3.3.5] Dynamic Creature/Go spawning

### LINKED ISSUE #2858: InhabitType 4 (Flying)
Creatures with only InhabitType=4 seems to respawn under the map (or fall) after you have killed them.

so you can only kill/target them once a serverstart.

it's annoying when this creature is unique and you require this for some quest - so you have to wait for a server down.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/255439-inhabittype-4-flying?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #13787: [3.3.5][6.x] Respawntimer should start ticking straight after killing the unit + creature should be respawned as completely new unit with new guid $20

### LINKED ISSUE #15222: NPC with minLevel <> maxLevel levelup animation on respawn
Rev: https://github.com/TrinityCore/TrinityCore/commit/747ea7ca2c2b757dc9c5a672746d801a8365ee98 (fresh compile)
DB: TDB 335.59

Problem: NPC with minLevel <> maxLevel set in db should not has levelup animation during respawn. To reproduce the problem just find any NPC with minLevel <> maxLevel, kill him and wait. If old level will not match new level at respawn, you will see levelup animation for this creature.

This issue mb related for both 6.x and 3.3.5 branches.

r00ty-tc:

I've finally promoted this to PR. So I can actually have some proper testing feedback and somewhere to log any issues people have/changes requested (since I forget some of the things requested previously)

**Changes proposed:**

- Replace existing spawn/respawn system
- Add spawn group functionality
- Add dynamic respawn functionality
- Add persistent grid functionality.
- Controlled restart upon GUID exhaustion

- [x] 3.3.5
- [ ] master

**Issues addressed:** 
There are several (some have been closed) that are indirectly solved by this. Some examples I could find.
Closes #2858 
Closes #8661 (with application of grid persistance in WG code)
Closes #13787 
Closes #15222 (it's already closed, but still)

Compiles fine.
Tested listed features work as intended (although, it's possible some merged changes could have potentially adversely affected, I've not re-tested them after each merge!)
Tested levelling to level 30.

For this, currently no known issues. But, for such a core change I expect this area to be updated with a list of issues to fix.

Provided some more detailed documentation on the changes, and how to use some of the new features to aid in testing.
[Dynamic Spawning.pdf](https://github.com/TrinityCore/TrinityCore/files/753246/Dynamic.Spawning.pdf)


Treeston:
Did you adjust for b3d44d6c366c7e4f12a6ce30ee5cf6b9bff68a06

src/server/game/AI/ScriptedAI/ScriptedEscortAI.cpp
```diff
@@ -297,8 +297,10 @@ void npc_escortAI::UpdateAI(uint32 diff)
                 uint32 groupFlags = 0;
 
                 if (const CreatureData* cdata = me->GetCreatureData())
```
ariel-: CreatureData const*

src/server/game/AI/ScriptedAI/ScriptedEscortAI.cpp
```diff
@@ -297,8 +297,10 @@ void npc_escortAI::UpdateAI(uint32 diff)
...
+                {
                     if (const CreatureGroupTemplateData* groupData = cdata->groupdata)
```
ariel-: CreatureGroupTemplateData const*

src/server/game/Maps/Map.cpp
```diff
@@ -3158,27 +3149,20 @@ void Map::RespawnCreatureList(const RespawnVector& RespawnData, bool force)
             }
             else
             {
-                if (ri->spawnId > 0xFFFFFF)
+                const ObjectGuid::LowType spawnId = ri->spawnId;
```
ariel-: ObjectGuid::LowType const

src/server/game/Maps/Map.cpp
```diff
@@ -3158,27 +3149,20 @@ void Map::RespawnCreatureList(const RespawnVector& RespawnData, bool force)
...
+                if (const CreatureData* cdata = sObjectMgr->GetCreatureData(spawnId))
```
ariel-: CreatureData const*

src/server/game/AI/ScriptedAI/ScriptedEscortAI.cpp
```diff
@@ -289,11 +294,21 @@ void npc_escortAI::UpdateAI(uint32 diff)
             {
                 TC_LOG_DEBUG("scripts", "EscortAI failed because player/group was to far away or not found");
 
-                if (m_bCanInstantRespawn)
+                uint32 groupFlags = 0;
+
+                if (const CreatureData* cdata = me->GetCreatureData())
```
ariel-: CreatureData const*

horn:
With this PR I guess Map::SummonCreatureGroup() and WorldObject::SummonCreatureGroup() (both in Object.cpp) can be replaced by ObjectMgr::SpawnCreatureGroup().

r00ty-tc:
It's slightly different I think. SummonCreatureGroup is a group summoned by a specific creature/go (as in they become the owner).

This spawns them with no owner. So I think both mechanisms still have their place.

Killyana:
I don't know if it the right place to talk about this, but respawntimesecs on creature table is modified by something else that make the respawn time really huge, example:
.go c id 32239
.die
.npc inf
SpawnTime: Full:2m Remaining:31m57s
There's 30 min extra for the respawn because the npc has boss level
an other example of elite npcs:
Entry: 32175.
SpawnTime: Full:5m Remaining:7m28s

r00ty-tc:
@Killyana yes that's right. I think there needs to be some kind of spawn time update in the DB overall. But of course be careful to only apply it once.

Well, at least it's my opinion that respawn time should never have been influenced by a modifier in game to increase the time based on creature rank (by increasing corse despawn time). The respawn time int he DB should be correct regardless.

ariel-:
It comes from worldconfig:

```
#
#    Corpse.Decay.NORMAL
#    Corpse.Decay.RARE
#    Corpse.Decay.ELITE
#    Corpse.Decay.RAREELITE
#    Corpse.Decay.WORLDBOSS
#        Description: Time (in seconds) until creature corpse will decay if not looted or skinned.
#        Default:     60   - (1 Minute, Corpse.Decay.NORMAL)
#                     300  - (5 Minutes, Corpse.Decay.RARE)
#                     300  - (5 Minutes, Corpse.Decay.ELITE)
#                     300  - (5 Minutes, Corpse.Decay.RAREELITE)
#                     3600 - (1 Hour, Corpse.Decay.WORLDBOSS)

Corpse.Decay.NORMAL=60
Corpse.Decay.RARE=300
Corpse.Decay.ELITE=300
Corpse.Decay.RAREELITE=300
Corpse.Decay.WORLDBOSS=3600

#
#    Rate.Corpse.Decay.Looted
#        Description: Multiplier for Corpse.Decay.* to configure how long creature corpses stay
#                     after they have been looted.
#         Default:    0.5

Rate.Corpse.Decay.Looted=0.5
```

r00ty-tc:
Well, let's just be clear here. If I change it to add that time on, it means the respawn time for that kind of creature can NEVER be less than that. Removing one of the reasons this concept arrived in the first place. In fact removing one of the issues I linked that it is meant to close.

sql/updates/auth/3.3.5/2017_02_04_99_auth.sql
```diff
@@ -0,0 +1,15 @@
+DELETE FROM `rbac_permissions` WHERE `id` BETWEEN 846 AND 850;
+INSERT INTO `rbac_permissions` (`id`, `name`) VALUES
+(846, 'Command: npc spawngroup'),
+(847, 'Command: npc despawngroup'),
+(848, 'Command: gobject spawngroup'),
+(849, 'Command: gobject despawngroup'),
+(850, 'Command: list respawns');
```
joschiwald: ensure that these permission ids don't conflict with master branch

r00ty-tc: RBAC_PERM_COMMAND_GO_OFFSET already broke that sync. If I add the master ones in, one will be missing. I can't renumber offset as part of this PR, that would be silly. So, what is the proper procedure when this happens?

Treeston:
@Killyana that would be fixed after merge on a case by case basis. Current respawn times in DB are too inconsistent for a blanket DB adjustment, imo.

src/server/worldserver/worldserver.conf.dist
```diff
@@ -1685,6 +1686,177 @@ Creature.MovingStopTimeForPlayer = 180000
...
+#                     This can be based upon either number of players in the given radius, or
+#                     number of creatures alive with the same entry, vs dead within the given radius.
+#                     The latter mode requires a SQL script run with the same parameters, if the 
+#                     radius is changed.
+#                     Only creatures in a group which has the flag set for dynamic respawn will be
+#                     considered.
+#        Default:     0 - No dynamic respawn function
+#                     1 - Use number of players in radius
+
+Respawn.DynamicMode = 1
```
tkrokli: Use default value here or tag this value as default in the description.

horn:
@r00ty-tc It seems like duplicated stuff to me, you are right that there are some differences, but maybe you could adjust this PR a bit and rewrite SummonCreatureGroup() to use your system instead and add summons to their summoner's list if possible. No idea how complicated it would be though.

Treeston:
Doesn't SpawnCreatureGroup trigger natural respawns after that?

SummonCreatureGroup uses tempsummons.

r00ty-tc:
Pretty much yeah. I don't think we should go about trying to replace more than one system like that. I provided this, because there's a bit of a gap. I don't think I want to deliberately try to replace something that works fine, for it's own purposes.

Keader:
It should fix: #19071 too

Viste:
Can some one help to ajust WG code to this pr ? What need to do )

ariel-:
Sure thing, WG code can be revamped after this one is merged.

LuigiElleBalotta:
Why still not merged?

Aokromes:
Because it needs serious testing?

LuigiElleBalotta:
Can this be tested on master too?

r00ty-tc:
Nope, it will need to be ported over there first.

Treeston:
I've been running it for the last few weeks with no issue. Hopefully I can square away an evening and do actual reviewing soon.

Viste:
Bosses in ICC respawn in 24h Lady Deahthwispher have instant respawn(0d0m0s) ((

Aokromes:
update your core, that bug was fixed yesterday.

Treeston:
No, @Aokromes, not in dynspawn respawn logic. It's a conflict with 94a6dd7b3c2f336cde5d283d0e70980eb9370827 (@r00ty-tc will fix once he gets around to rebasing)

Viste:
@Aokromes its happen with this PR in clean core all right ) and i don't post issue just report in pr... ((

Viste:
to many bosses have 0d0m0s spawntime ((( almost all

Treeston:
> It's a conflict with 94a6dd7 (@r00ty-tc will fix once he gets around to rebasing)

r00ty-tc:
That was rebased. But, perhaps it doesn't play well with this.. I'll have to see.

r00ty-tc:
image

Push incoming.

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -384,7 +384,9 @@ void Creature::RemoveCorpse(bool setSpawnTime, bool destroyForNearbyPlayers)
         if (setSpawnTime)
         {
             uint32 respawnDelay = m_respawnDelay;
-            m_respawnTime = time(NULL) + respawnDelay;
+            if (setSpawnTime)
```
steven-foster: redundant?

r00ty-tc:
Yep. Tired coding is almost as bad as dry-coding.

Viste:
wowscrnshot_022117_155047
Now ))

Viste:
On last rev with pr elevators are broken and ships on gunship battle to you just stay in place and object move without you
wowscrnshot_022717_012125

r00ty-tc:
Hmm, I can't even test it now. But looking at recent commits to 3.3.5 I can't see anything that would do that.. I'll check it out when I'm home Wed/Thu.

Viste:
@r00ty-tc sorry its my db issue

r00ty-tc:
Ah, OK. That's good to hear.

Killyana:
Better to push the PR to have more feed back and before having more conflicts to resolve.

r00ty-tc:
@Killyana I've done about as much testing as I can do on my own (of course there's the risk that without regression testing, commits I've worked around could be affecting anything I already tested).

At this point, the only issue is whether other people can also fix problems that arise, since my availability is variable at best.

src/server/game/Maps/Map.h
```diff
@@ -253,6 +278,9 @@ struct ZoneDynamicInfo
 typedef std::map<uint32/*leaderDBGUID*/, CreatureGroup*>        CreatureGroupHolderType;
 
 typedef std::unordered_map<uint32 /*zoneId*/, ZoneDynamicInfo> ZoneDynamicInfoMap;
+typedef std::unordered_multimap<uint32, RespawnInfo*> respawnInfoMultiMap;
```
Treeston: Refactor type names to uppercase.

Treeston (approved):
Waiting for final squash and rebase by @r00ty-tc. Then merge. It's happening.

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -3120,3 +3216,20 @@ void Creature::ClearTextRepeatGroup(uint8 textGroup)
     if (groupItr != m_textRepeat.end())
         groupItr->second.clear();
 }
+
+bool Creature::IsEscortNPC(bool isEscorting)
+{
+    if (!GetAI())
+        return false;
+
+    if (npc_escortAI* escortAI = dynamic_cast<npc_escortAI*> (AI()))
```
Shauren: this need to be virtual method in CreatureAI*, do not create a dependency on ai type here

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9174,6 +9559,24 @@ CreatureTemplate const* ObjectMgr::GetCreatureTemplate(uint32 entry) const
     return nullptr;
 }
 
+uint32 ObjectMgr::GetCreaturesInGroup(uint32 groupid, std::vector<uint32>& creatureList)
```
Shauren: arg type mismatch, declared `std::vector<ObjectGuid::LowType>` in the header

src/server/game/Grids/Notifiers/GridNotifiers.h
```diff
@@ -566,6 +566,22 @@ namespace Trinity
     };
 
     template<class Check>
+    struct PlayerListSearcherByPosition
```
Shauren: this is an exact copypaste of `PlayerListSearcher`, remove

src/server/game/Maps/Map.cpp
```diff
@@ -2769,6 +2888,587 @@ void Map::SendObjectUpdates()
...
+            CharacterDatabase.CommitTransaction(trans);
+
+        if (respawnObject)
+            RespawnGameObjectList(rv, true);
+        else
+            DeleteGameObjectRespawnInfo(spawnId, gridId, cellAreaZoneId, false);
+    }
+}
+
+void Map::addRespawnInfo(respawnInfoMultiMap& gridList, respawnInfoMultiMap& cellAreaZoneList, respawnInfoMap& spawnList, RespawnInfo& Info, bool replace)
```
Shauren: codestyle, rename (method, arg types, arg names)

src/server/game/Maps/Map.cpp
```diff
@@ -2769,6 +2888,587 @@ void Map::SendObjectUpdates()
...
+
+    if (replace)
+    {
+        // In case of replace, delete this spawn ID everywhere, then self call with no replace
+        deleteRespawnInfo(gridList, cellAreaZoneList, spawnList, Info.spawnId, 0, 0, false);
+        addRespawnInfo(gridList, cellAreaZoneList, spawnList, Info, false);
+    }
+}
+
+bool Map::getRespawnInfo(respawnInfoMultiMap const& gridList, respawnInfoMultiMap const& cellAreaZoneList, respawnInfoMap const& spawnList, RespawnVector& RespawnData, ObjectGuid::LowType spawnId, uint32 gridId, uint32 cellAreaZoneId, bool onlyDue)
```
Shauren: codestyle, rename (method, arg types, arg names)

src/server/game/Maps/Map.cpp
```diff
@@ -2769,6 +2888,587 @@ void Map::SendObjectUpdates()
...
+            }
+            ++iter;
+        }
+        return true;
+    }
+
+    return false;
+}
+
+void Map::deleteRespawnInfo(respawnInfoMultiMap& gridList, respawnInfoMultiMap& cellAreaZoneList, respawnInfoMap& spawnList, ObjectGuid::LowType spawnId, uint32 gridId, uint32 cellAreaZoneId, bool onlyDue)
```
Shauren: codestyle, rename (method, arg types, arg names)

src/server/game/Maps/Map.cpp
```diff
@@ -2769,6 +2888,587 @@ void Map::SendObjectUpdates()
...
+            break;
+        case RESPAWNSCOPE_ZONE:
+            RespawnCellAreaZoneGameObject(zoneId);
+            break;
+        default:
+            ASSERT("INVALID RESPAWN SCOPE");
+    }
+}
+
+void Map::transformRespawnList(RespawnVector& RespawnData, uint32 numPlayers, float adjustFactor, uint32 minAdjustSpawn, uint32 mode)
```
Shauren: codestyle, rename (method, arg names)

src/server/game/Maps/Map.cpp
```diff
@@ -3552,64 +4280,102 @@ void Map::UpdateIteratorBack(Player* player)
         m_mapRefIter = m_mapRefIter->nocheck_prev();
 }
 
-void Map::SaveCreatureRespawnTime(ObjectGuid::LowType dbGuid, time_t respawnTime)
+void Map::SaveCreatureRespawnTime(ObjectGuid::LowType spawnId, uint32 entry, time_t respawnTime, uint32 cellAreaZoneId, uint32 gridId, bool WriteDB, bool replace, SQLTransaction respawntrans)
```
Shauren: codestyle, rename (arg name)

src/server/game/Maps/Map.cpp
```diff
@@ -3552,64 +4280,102 @@ void Map::UpdateIteratorBack(Player* player)
...
-    stmt->setUInt32(2, GetInstanceId());
-    CharacterDatabase.Execute(stmt);
+    if (respawntrans)
+        respawntrans->Append(stmt);
+    else
+        CharacterDatabase.Execute(stmt);
 }
 
-void Map::SaveGORespawnTime(ObjectGuid::LowType dbGuid, time_t respawnTime)
+void Map::SaveGORespawnTime(ObjectGuid::LowType spawnId, uint32 entry, time_t respawnTime, uint32 cellAreaZoneId, uint32 gridId, bool WriteDB, bool replace, SQLTransaction respawntrans)
```
Shauren: codestyle, rename (arg name)

src/server/game/Maps/Map.h
```diff
@@ -633,6 +705,7 @@ class TC_GAME_API Map : public GridRefManager<NGridType>
         void UpdateActiveCells(const float &x, const float &y, const uint32 t_diff);
 
         void SendObjectUpdates();
+//        uint32 m_respawnDiff;
```
Shauren: dont add commented out code

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -950,12 +970,21 @@ void Creature::Motion_Initialize()
...
 }
 
-bool Creature::Create(ObjectGuid::LowType guidlow, Map* map, uint32 phaseMask, uint32 entry, float x, float y, float z, float ang, CreatureData const* data /*= nullptr*/, uint32 vehId /*= 0*/)
+bool Creature::Create(ObjectGuid::LowType guidlow, Map* map, uint32 phaseMask, uint32 entry, float x, float y, float z, float ang, CreatureData const* data /*= nullptr*/, uint32 vehId /*= 0*/, bool dynamic)
 {
     ASSERT(map);
     SetMap(map);
     SetPhaseMask(phaseMask, false);
 
+    if (!sWorld->isGuidAlert() && guidlow > sWorld->getIntConfig(CONFIG_RESPAWN_GUIDALERTLEVEL))
```
Shauren: wouldn't ObjectGuidGenerator be a better place for this?

Shauren (changes requested):

src/server/game/World/World.cpp
```diff
@@ -194,6 +199,59 @@ const char* World::GetMotd() const
     return m_motd.c_str();
```
tkrokli: To resolve the conflict in this file, remove your line numbers 190-201 ( `motd` changes in TC).

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1483,9 +1525,12 @@ bool Creature::LoadCreatureFromDB(ObjectGuid::LowType spawnId, Map* map, bool ad
 
     m_respawnTime = GetMap()->GetCreatureRespawnTime(m_spawnId);
 
-    // Is the creature script objecting to us spawning? If yes, delay by one second (then re-check in ::Update)
-    if (!m_respawnTime && !sScriptMgr->CanSpawn(spawnId, GetEntry(), GetCreatureTemplate(), GetCreatureData(), map))
-        m_respawnTime = time(NULL)+1;
+    if (m_respawnCompatibilityMode)
+    {
+        // Is the creature script objecting to us spawning? If yes, delay by one second (then re-check in ::Update)
+        if (!m_respawnCompatibilityMode && !m_respawnTime && !sScriptMgr->CanSpawn(spawnId, GetEntry(), GetCreatureTemplate(), GetCreatureData(), map))
```
ariel-: m_respawnCompatibilityMode is always true here, checked one line above

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -226,11 +226,16 @@ void GameObject::RemoveFromWorld()
     }
 }
 
-bool GameObject::Create(ObjectGuid::LowType guidlow, uint32 name_id, Map* map, uint32 phaseMask, Position const& pos, G3D::Quat const& rotation, uint32 animprogress, GOState go_state, uint32 artKit /*= 0*/)
+bool GameObject::Create(ObjectGuid::LowType guidlow, uint32 name_id, Map* map, uint32 phaseMask, Position const& pos, G3D::Quat const& rotation, uint32 animprogress, GOState go_state, uint32 artKit /*= 0*/, bool dynamic, ObjectGuid::LowType spawnid)
 {
     ASSERT(map);
     SetMap(map);
 
+    if (!sWorld->isGuidAlert() && guidlow > sWorld->getIntConfig(CONFIG_RESPAWN_GUIDALERTLEVEL))
```
ariel-: If the warning code is moved to generators, this code can be removed to avoid coupling.

src/server/game/Entities/Creature/Creature.cpp
r00ty-tc: possible I moved it to the wrong place on one of the many rebase operations. It should be outside that check.

I'll get to looking at this stuff maybe, this weekend.

src/server/game/Grids/Notifiers/GridNotifiers.h
r00ty-tc: No, it's not. 

        PlayerListSearcherByPosition(uint32 phaseMask, std::list<Player*> &objects, Check & check)
            : i_phaseMask(phaseMask), i_objects(objects), i_check(check) { }

vs

        PlayerLastSearcher(WorldObject const* searcher, Player*& result, Check& check) : i_phaseMask(searcher->GetPhaseMask()), i_object(result), i_check(check)
        {
        }

Primarily the difference is all other searchers need an object and derive phasemask (and usually position) from the object. This has phasemask passed as argument.

Unless, I'm misunderstanding the objection.

r00ty-tc:
All requested changes applied, aside from the one I objected to (see reasons above)

src/server/worldserver/worldserver.conf.dist
```diff
@@ -1696,6 +1697,177 @@ Creature.MovingStopTimeForPlayer = 180000
...
+#        Default:     1800 - (30 minutes)
+
+Respawn.WarningFrequency = 1800
+
+#
+#    Respawn.GuidAlertLevel
+#        Description: The point at which the highest guid for creatures or gameobjects in any map must reach
+#                     before the alert logic is enabled. A restart will then be triggered for 30 mins from that
+#                     point. The maximum guid per map is 16,777,216. So, it must be less than this value.
+#        Default:     16000000 - 12 million
```
Raydor: 16 million

src/server/worldserver/worldserver.conf.dist
```diff
@@ -1696,6 +1697,177 @@ Creature.MovingStopTimeForPlayer = 180000
...
+#    Respawn.DynamicMode
+#        Description: Select which mode (if any) should be used to adjust respawn of creatures.
+#                     This can be based upon either number of players in the given radius, or
+#                     number of creatures alive with the same entry, vs dead within the given radius.
+#                     The latter mode requires a SQL script run with the same parameters, if the 
+#                     radius is changed.
+#                     Only creatures in a group which has the flag set for dynamic respawn will be
+#                     considered.
+#        Default:     1 - Use number of players in radius
+                      0 - No dynamic respawn function
```
Raydor: Missing the #

Raydor:
So what is needed here in order to merge it? @r00ty-tc said he applied all the changes above.

Keader:
Last time I heard about it, Rooty would travel and tell us to not merge before he came back

Raydor:
Wow that's really good news, can't wait for this to get merged finally :+1:

src/server/game/AI/SmartScripts/SmartScriptMgr.h
```diff
@@ -584,8 +584,12 @@ enum SMART_ACTION
     SMART_ACTION_REMOVE_ALL_GAMEOBJECTS             = 126,
     SMART_ACTION_STOP_MOTION                        = 127,    // stopMoving, movementExpired
     SMART_ACTION_PLAY_ANIMKIT                       = 128,    // don't use on 3.3.5a
+    SMART_ACTION_SPAWN_CREATUREGROUP                = 129,    // Group ID, min secs, max secs
+    SMART_ACTION_DESPAWN_CREATUREGROUP              = 130,    // Group ID, min secs, max secs
+    SMART_ACTION_SPAWN_GAMEOBJECTGROUP              = 131,    // Group ID, min secs, max secs
+    SMART_ACTION_DESPAWN_GAMEOBJECTGROUP            = 132,    // Group ID, min secs, max secs
```
joschiwald: conflicts with master branch

src/server/game/Miscellaneous/Language.h
```diff
@@ -1135,7 +1135,24 @@ enum TrinityStrings
     LANG_COMMAND_MUTEHISTORY_EMPTY      = 5060,
     LANG_COMMAND_MUTEHISTORY_OUTPUT     = 5061,
 
-    // Room for more Trinity strings      5062-9999
+    // Strings added for dynamic spawns
+    LANG_NPCINFO_GROUP_ID               = 5062,
```
joschiwald: also conflicts with master branch

r00ty-tc: Should have resolved both now.

Demonid:
Push the Button!

👍

r00ty-tc:
Something was pointed out yesterday evening that could cause performance problems (potentially). So, before merging I need to think of a way to either measure/resolve, or just work around the bottleneck.

It's a core part of the process. So, not easy to workaround.

ariel-:
> Something was pointed out yesterday evening that could cause performance problems (potentially). So, before merging I need to think of a way to either measure/resolve, or just work around the bottleneck.

What if that call is cached and queried only if necessary? SunwellCore had some optimization like that

r00ty-tc:
Maybe, it might help at least. I just need a block of clear time to be able to work on it.

tkrokli:
Conflict in 3.3.5 after commit https://github.com/TrinityCore/TrinityCore/commit/9e65fead38c486ec8a971a09a95ffae912943a83#diff-ff3b59d5f6fc8303cc01120ec5c190fc :
```
<<<<<<< dynamic_spawning
    RBAC_PERM_COMMAND_NPC_SPAWNGROUP                         = 853,
    RBAC_PERM_COMMAND_NPC_DESPAWNGROUP                       = 854,
    RBAC_PERM_COMMAND_GOBJECT_SPAWNGROUP                     = 855,
    RBAC_PERM_COMMAND_GOBJECT_DESPAWNGROUP                   = 856,
    RBAC_PERM_COMMAND_LIST_RESPAWNS                          = 857,
=======
    RBAC_PERM_COMMAND_RELOAD_CONVERSATION_TEMPLATE           = 853, // not on 3.3.5a
    RBAC_PERM_COMMAND_DEBUG_CONVERSATION                     = 854, // not on 3.3.5a
>>>>>>> 3.3.5
```

r00ty-tc:
Yeah @tkrokli tbh it's getting quite boring spending more time fixing conflicts than actually working on the branch.

tallang:
Just give it some stupid high permission id until you're merge ready.

Treeston:
Just give it some stupid high permission id until you're merge ready.

Raydor:
@r00ty-tc So the only thing needed by this PR is fixing the perfomance issues? Are those that bad to not merge this and fix it in the (I hope close) future when you have a block of free time?

ccrs:
sadly no, the spotted issue is too big to just hope for the best

r00ty-tc:
There have been some good suggestions here, as to how to resolve it (and indeed probably slightly improve overall performance too).

When I get a chance, I will look into it. Again though, I've been busy. However, it's possible anyone else can also resolve this either by commit to 3.3.5 or to this branch.

Treeston:
@r00ty-tc can you quickly elaborate on what the actual issue is?

ccrs:
this https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Maps/Map.cpp#L2443
is used often in the new system to, for example, set respawn timers or list the number of players in the area

r00ty-tc:
Yeah.

Basically https://github.com/TrinityCore/TrinityCore/pull/19056/files#diff-9eea3ad0b189ac095427d3c6c6ff68e7R709

For any object that is active (players, setactive creatures) it will on every update tick get the area and zone id. The reason being that it needs to check for respawns, and the scope may be Zone or Area. 

Now this can be stopped by disabling zone/area scope. However this really is useful for GO spawns because of the fact that currently you need to visit the location of the previous object to activate the respawn, which may be elsewhere and this won't actually appear until you visit the new location. This way if a player is in the zone, the pooled respawn will happen anyway.

Now solutions, I see two.

1: Cache the zone/area info in an object. This is a core change separate to this IMO. Probably the best overall solution since this is something that probably impacts core anyway.

2: Don't check for player presence to handle respawns. Actually I don't see this as a bad thing. Currently the respawns expire but only become actioned when an active object enters an area needing a respawn. In theory, if there's only a few players around then there won't be many respawns. If there's many respawns it's likely that players are active anyway. So, this might be a false economy in the first place. In the event a grid isn't loaded for a queued respawn, who cares? It will be respawned when the grid is loaded next time anyway. So, discard.

r00ty-tc:
Pretty much

Raydor:
I like where that conversation is going... :smile:

Treeston:
I looked into the caching outlined above, but this would require un-consting of GetArea/ZoneId methods, which would in turn un-const a whole bunch of other things. 
Not sure I like that - I'll look into less invasive approaches, but can't promise anything.

Shauren:
@Treeston `mutable` keyword, look it up (using it for caches is perfectly fine)
https://ideone.com/eTjHKN

But imo you should NOT calculate and store it inside GetArea/Zone methods - creature can move and you don't know when to invalidate the cache (a better way would be to implement a new vmap ray callback that retrieves COMPLETE area/liquid/height info with a single call - we already do liquid call in Unit::Update/or one of the methods called by it)

Treeston:
@Shauren my idea was to store old x/y/z values and update at some diff, but I agree that it feels like a hack.

I was also thinking about doing something along the lines of Unit::Update (since nonunits never move), but I'd need to figure out how vmaps work first.

Treeston:
See #19840 which spawned from this discussion.

Treeston:
@r00ty-tc #19840 is now merged, go go go!

Treeston:
Reserved RBAC permissions 856-860 on all relevant branches so we can stop having to rebase every few weeks. Also did the last rebase to update RBACs on this branch and resolve conflicts.

@r00ty-tc can you take a look at my merge of `Creature::ForcedDespawn` when you have a moment? In addition, on the same method, are you sure that your force logic for non-compat does what you want? It seems as if you use `m_respawnDelay` instead of the force timer.

Raydor:
I found a way to crash the server with this system:

Steps:
1 - Kill a NPC (with .damage for example)
2 - Respawn the npc (with .resp for example)
3 - You will see that there is a new creature alive and you will see the dead body of the other npc in the ground
4 - Remove (with .npc del) the dead NPC
5 - Kill the alive NPC (with .damage for example)
6 - Use .respawn without targetting the NPC

I don't know if this is intended, but when the server restarts the NPC is not there so is like its trying to respawn an NPC that doesnt exist and so that it crashes.

Treeston:
Interesting. `.npc del` does real-time modification to spawn data - I suspect that queued respawns/existing instances of the creature aren't being removed.

Raydor:
So this means it can only be exploited by GMs? there is no other situation where a npc despawn itself or something like that right? Just curious

src/server/game/Maps/Map.cpp
```diff
@@ -679,10 +704,95 @@ void Map::VisitNearbyCellsOf(WorldObject* obj, TypeContainerVisitor<Trinity::Obj
             cell.SetNoCreate();
             Visit(cell, gridVisitor);
             Visit(cell, worldVisitor);
+
+            uint32 zoneId, areaId;
+            obj->GetZoneAndAreaId(zoneId, areaId);
+            // Handle possible respawns
+            RespawnCellAreaZone(cell.GetCellCoord().GetId(), zoneId, areaId);
```
Shauren: I'm sure this part doesn't work properly - each cell is visited exactly once (isCellMarked/markCell) but there can be more than 1 area/zone in a given cell so some areas/zones might never respawn depending on visiting order

src/server/game/Maps/Map.cpp
```diff
@@ -516,6 +526,21 @@ bool Map::EnsureGridLoaded(const Cell &cell)
     return false;
 }
 
+void Map::GridMarkNoUnload(uint32 x, uint32 y)
+{
+    _gridNoUnload.insert(GridCoord(x, y).GetId());
```
Shauren: Each grid already has a way to prevent unloading it - https://github.com/TrinityCore/TrinityCore/blob/7874bee7bfb70e0e039f91173cff212e9572de09/src/server/game/Grids/NGrid.h#L57

src/server/game/Maps/Map.cpp
```diff
@@ -679,10 +704,95 @@ void Map::VisitNearbyCellsOf(WorldObject* obj, TypeContainerVisitor<Trinity::Obj
...
+
+    uint32 oldZoneCount = 0;
+    uint32 newZoneCount = 0;
+
+    // Get old zone if exist, without creating element if not
+    auto oldItr = _zonePlayerCountMap.find(oldZone);
+    if (oldItr != _zonePlayerCountMap.end())
+        oldZoneCount = oldItr->second;
+
+    // Sanity check, we're leaving a zone (that isn't the default) and there's no players there (should be at least one)
```
Shauren: sounds like a job for ASSERT instead

Treeston:
Putting this here since too since you aren't responding on IRC, @r00ty-tc:
```
<Treeston> so my view of the visit logic as you have it right now:
<Treeston> 1. if update mode is cell, it visits and updates all cells in grid activation range of any active object, with each cell being visited exactly once (so far, so good)
<Treeston> 2. if update mode is area or zone, it visits any area/zone that has at least one active object in them, but not any area/zone that may be in activation range of the active object
<Treeston> 3. for update mode area/zone, each area/zone can also be visited multiple times if it has multiple active objects
<Treeston> 1. is correct behavior, 2. and 3. are problematic
<Treeston> here's my take on it
<Treeston> 2. we have no cheap way (to my knowledge) of finding out which areas/zones are in a given cell/grid - thus, i'd recommend just making modes continously increasing (e.g., if update mode is area or zone, still respawn cell too - this should give the behavior you want)
<Treeston> 3. i (personally) don't see this being a significant issue - RespawnCellAreaZoneX methods check time since last check almost immediately, so the call will return after (effectively) one map lookup
<Treeston> only way we could make 3. faster is by using a faster container (bitset?) but idk if it's really worth it
```

Treeston:
PS also as a consequence I'd prefer if you separated cell respawns from area/zone respawns. Also, do we really need _area_ respawns? IMO cell or zone is enough choice and should greatly clean up the logic (but up to you).

Treeston:
Alright, I've taken a few hours to work on this.

After coordinating with @r00ty-tc yesterday, I've made the following adjustments:
* World config based scoping is dead. It made the logic a mess, introduced a bunch of overhead and had the potential to cause significant edge case issues.
* All relevant logic has been changed to act as if zone scoping was enabled. A zone can be processed multiple times (once per active object in it), but any calls after the first will immediately return after a single map lookup.

In addition, I've taken the liberty of making the following adjustments:
* A set of minor refactors to better reflect the nature of `Map` methods (`RespawnXType` -> `ProcessTypeRespawns`)
* Adjust `GetRespawnData` (only used by debug commands) to improve efficiency. Also removed a bunch of unused arguments.
* `getMSTime` -> `GameTime` for cheap performance improvements
* Removed the auth sql files from this PR. All necessary adjustments to auth were already pushed to 3.3.5 base in 55d8c2b and 06c5d13. This should fix appveyor builds.
* Renamed world sql file to make finding it easier.
* Changed the behavior of `list respawns` to generally be more useful.
* Some minor changes to format strings to make them more readable.
* Adjust documentation in various places to eliminate references to arbitrary "regions" or "scopes", replace with "zone" to reflect new behavior.

With this, the PR is once again in a state I'd consider fully functional and merge ready. However, I have another point I'd like to raise:
* Currently, there is strong code duplication due to all methods and containers being split between `Creature` respawns and `GameObject` respawns. I personally don't feel this separation at all levels of the call stack is needed, and could just as well be served by a `RespawnObjectType` member in `RespawnData`.
* @r00ty-tc, is there any strong reasoning behind maintaining this separation? If not, I'd like to merge all of this logic - possibly maintaining separation at lower levels of the call stack, e.g. the actual respawn logic, as appropriate. I feel this would greatly reduce the amount of duplicated code and likely also provide a minor performance speed-up (at negligible memory cost).

src/server/game/Maps/Map.cpp
```diff
@@ -3142,14 +3095,14 @@ bool Map::GetRespawnInfo(RespawnInfoMultiMap const& gridList, RespawnInfoMultiMa
...
 }
 
-void Map::DeleteRespawnInfo(RespawnInfoMultiMap& gridList, RespawnInfoMultiMap& cellAreaZoneList, RespawnInfoMap& spawnList, ObjectGuid::LowType spawnId, uint32 gridId, uint32 cellAreaZoneId, bool onlyDue)
+void Map::DeleteRespawnInfo(RespawnInfoMultiMap& gridList, RespawnInfoMultiMap& zoneList, RespawnInfoMap& spawnList, ObjectGuid::LowType spawnId, uint32 gridId, uint32 zoneId, bool onlyDue)
 {
     // For delete, a bit more important to ensure only one present (or none to erase all)
-    if ((spawnId && (gridId || cellAreaZoneId)) || (gridId && (spawnId || cellAreaZoneId)) || (cellAreaZoneId && (spawnId || gridId)))
-        ASSERT("too many active parameters passed to Map::deleteRespawnInfo");
+    if ((spawnId && (gridId || zoneId)) || (gridId && (spawnId || zoneId)) || (zoneId && (spawnId || gridId)))
+        ASSERT(false && "too many active parameters passed to Map::deleteRespawnInfo");
```
Shauren: Please, ASSERT signature is ASSERT(condition) or ASSERT(condition, msg, ...args), you can use the second version here

Shauren:
I'm pretty sure that creature respawn timers are not zone wide, extensive farming of ogres in one camp does not make respawning faster in entire zone but only affects the ogres in that camp

The same goes for gameobject quest objectives - but at the same time there is zone wide gathering node pooling

r00ty-tc:
@Shauren it was down to scope originally, but to simplify we moved to zone only spawn scope. This is a just a stepping stone though on the dynamic rate subject. 

The correct (and verified by me on current retail) way, is down to pool group using a min/max spawn setting, where respawns are forced to instant when a pool count drops below a certain level (I can provide a location to prove it if required). 

But, since we don't (yet) have creature pool groups. I decided to go with an overall system that isn't perfect, but will make things a bit better. Right now there's nothing. So I think it's an improvement of sorts, until there's a proper pooling system in place.

src/server/game/Maps/Map.cpp
Treeston: :+1:

Treeston:
@Shauren I agree that this is not blizzlike - however, as @r00ty-tc outlined, we need additional info in the DB for blizzlike spawn behavior to work, and we don't have that info. Consider this a custom feature to alleviate the pain for large servers in the meantime (it's off by default).

Treeston (approved):
Alright, I think I'm done on my end. @Shauren, @r00ty-tc, the ball's back in your court now. All tests I ran (general respawn behavior, dynamic scaling, manual spawn groups etc.) perform as expected.

Kittnz:
Don't compare the pooling from legion to 335 standards tho.. Instant respawning didn't happen back then.

ccrs:
Instant respawn? where is that defined? as far as I can see its a dynamic respawn which can just be close to 0 if the conditions meet for a scenario

r00ty-tc:
@Kittnz 

I think it did, if you ever were on a realm where many people decided to roll level 1's (new realm, or just some online project) you'd see that there was never a shortage of the quest creatures.

The example I've found is quite good for visually seeing it happen (it's a single pool with a minimum spawn in a small area). But generally it happened, and you'd have seen it when in a really busy area killing creatures you'd see one respawn in front of your eyes.

Now, I know you saw it more in cata+, but I actually think this is because they reduced the number of spawnpoints when they remade the world cata+. I think in 3.3.5 and below there were many spawnpoints, of which only say 50-60% were used at any one time. So even when this fast respawn was invoked, you generally didn't get one respawn right on top of one you just killed. But, I don't think anyone can prove/disprove that now.

jackpoz:
As with all big PRs, some checks with valgrind memcheck, helgrind and pvs-studio or other tools of your choice might be nice.

Treeston:
I can definitely say that it wasn't yet implemented by 3.0, @r00ty-tc - the utter starvation of leveling through Northrend with the "main pack" of players on launch was real.

Every raiding guild tried to get their people ahead of the "main pack" so they could level in mostly empty zones, or just spammed instanced dungeons instead.

Treeston:
PS: Not quite ready yet, though of one more efficiency speedup we can make.

bumbummen99:
@ccrs that's exactly what it is. I can remember that back in 2009 i did not have to wait for a corpse to be looted and disappeared until the respawn-timer started (What is current TC 3.3.5 behaivour).

Treeston:
We're talking about completely different things, @bumbummen99. You're talking about creatures being able to respawn while a corpse still persists - that's both blizzlike and implemented in this PR regardless of world config.

What everyone else is talking about is dynamic respawn time scaling, e.g. respawns happening faster because there's more players doing stuff (so nobody runs out of mobs to kill). That's an optional (custom) feature on this branch, enabled in the worldserver config.

Kittnz:
As far i can remember at launch of woltk there were empty spots everywhere. There were no instant respawns for players, but the begin zone npcs all had shorter respawn timers. So you still had to wait for mobs to respawn. I could be wrong ofc.

Treeston:
Alright, _now_ it's ready.

src/server/worldserver/worldserver.conf.dist
```diff
@@ -1714,6 +1715,123 @@ Creature.MovingStopTimeForPlayer = 180000
...
+#        Description: The point at which the highest guid for creatures or gameobjects in any map must reach
+#                     before the warning logic is enabled. A restart will then be queued at the next quiet time
+#                     The maximum guid per map is 16,777,216. So, it must be less than this value.
+#        Default:     12000000 - 12 million
+
+Respawn.GuidWarnLevel = 12000000
+
+#
+#    Respawn.WarningMessage
+#        Description: This mesage will be periodically shown (Frequency specified by Respawn.WarningFrequency) to
```
vanasis: `Description: This **mesage** will `
a small typo here

morpheus4hire:
@Treeston Received world server crash when looting Remedy of Arugal book for quest Arugals Folly. Replicated multiple times crashes with map.cpp assertion failed couldn't get it all written down before it closed the world server.

Treeston:
Will investigate, thanks for testing!

Treeston:
That took me way longer than it had any right to, @morpheus4hire. GO didn't want to become lootable for some reason.

Anyway, once it did, bug was obvious, I missed a `break;` and was continuing into a fallback case. Woops.

PS: Run worldserver from console instead of by double-click to preserve the output.

morpheus4hire:
@Treeston 
My apologies I'm still learning all of this. Love the work all of you do and I hope to be able to help more in the future. To everyone who works on this project I hope you all realize there are people who appreciate all the work you all do. Right now though until i get more experienced testing is about all i can do to try and help and I will take your advice on running from console thanks,

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1467,31 +1481,41 @@ bool Creature::LoadCreatureFromDB(ObjectGuid::LowType spawnId, Map* map, bool ad
...
+    m_respawnCompatibilityMode = !data->spawnGroupData || (data->spawnGroupData->flags & SPAWNGROUP_FLAG_COMPATIBILITY_MODE);
     m_creatureData = data;
     m_respawnradius = data->spawndist;
     m_respawnDelay = data->spawntimesecs;
-    if (!Create(map->GenerateLowGuid<HighGuid::Unit>(), map, data->phaseMask, data->id, data->posX, data->posY, data->posZ, data->orientation, data))
+
+    // Is the creature script objecting to us spawning? If yes, delay by one second (then re-check in ::Update)
+    if (!m_respawnCompatibilityMode && !m_respawnTime && !sScriptMgr->CanSpawn(spawnId, data->id, data, map))
+    {
+        SaveRespawnTime(urand(4,7));
```
Shauren: comment does not match code

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -2211,15 +2203,123 @@ void ObjectMgr::LoadGameobjects()
...
+        SpawnGroupTemplateData& data = _spawnGroupDataStore[0];
+        data.groupId = 0;
+        data.name = "Default Group";
+        data.mapId = 0;
+        data.flags = SPAWNGROUP_FLAG_SYSTEM;
+        data.isActive = true;
+    }
+
+    if (result)
+        TC_LOG_INFO("server.loading", ">> Loaded %zu spawn group templates in %u ms", _spawnGroupDataStore.size(), GetMSTimeDiffToNow(oldMSTime));
```
Shauren: Use the portable format specifier from Define.h - SZFMTD

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -2211,15 +2203,123 @@ void ObjectMgr::LoadGameobjects()
...
+            {
+                TC_LOG_ERROR("server.loading", "Spawn group %u has map ID %u, but spawn (%u,%u) has map id %u - spawn NOT added to group!", groupId, groupTemplate.mapId, uint32(spawnType), spawnId, data->GetMapId());
+                continue;
+            }
+            const_cast<SpawnData*>(data)->spawnGroupData = &groupTemplate;
+            _spawnGroupMapStore.emplace(groupId, data);
+        }
+    } while (result->NextRow());
+
+    TC_LOG_INFO("server.loading", ">> Loaded %zu spawn group members in %u ms", _spawnGroupMapStore.size(), GetMSTimeDiffToNow(oldMSTime));
```
Shauren: Use the portable format specifier from Define.h - SZFMTD

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -2211,15 +2203,123 @@ void ObjectMgr::LoadGameobjects()
...
+}
+
+
```
Shauren: double \n

src/server/game/Globals/ObjectMgr.cpp
Treeston: Is `%zu` not portable? I was under the impression that it always matched `size_t`.

src/server/game/Globals/ObjectMgr.h
```diff
@@ -1254,6 +1267,16 @@ class TC_GAME_API ObjectMgr
...
         }
 
+        SpawnData const* GetSpawnData(SpawnObjectType type, ObjectGuid::LowType guid)
+        {
+            if (type == SPAWN_TYPE_CREATURE)
+                return GetCreatureData(guid);
+            else if (type == SPAWN_TYPE_GAMEOBJECT)
+                return GetGameObjectData(guid);
+            else
+                ASSERT(false, "Invalid spawn object type %u", uint32(type));
```
Shauren: Errors.h is not included in this file but brought in by some other header and I don't like that this code abuses that (move to .cpp?)

src/server/game/Globals/ObjectMgr.h
```diff
@@ -1202,8 +1209,14 @@ class TC_GAME_API ObjectMgr
...
-        uint32 GenerateCreatureSpawnId();
-        uint32 GenerateGameObjectSpawnId();
+        ObjectGuid::LowType GenerateCreatureSpawnId();
+        ObjectGuid::LowType GenerateGameObjectSpawnId();
+
+        bool SpawnGroupSpawn(uint32 groupId, Map* map, bool ignoreRespawn = false, bool force = false, std::vector<WorldObject*>* spawnedObjects = nullptr);
+        bool SpawnGroupDespawn(uint32 groupId, Map* map, bool deleteRespawnTimes = false);
+        void SetSpawnGroupActive(uint32 groupId, bool state) { auto it = _spawnGroupDataStore.find(groupId); if (it != _spawnGroupDataStore.end()) it->second.isActive = state; }
+        bool IsSpawnGroupActive(uint32 groupId) const { auto it = _spawnGroupDataStore.find(groupId); return (it != _spawnGroupDataStore.end()) && it->second.isActive; }
+        Trinity::Containers::IteratorPair<SpawnGroupLinkContainer::const_iterator> GetSpawnDataForGroup(uint32 groupId) const { return _spawnGroupMapStore.equal_range(groupId); }
```
Shauren: Worth porting fdd9227b232db9aae9bc4b2c60ca4de2cdaa0b54

src/server/game/Maps/Map.cpp
Shauren: no reply?

Treeston: will change

src/server/game/Globals/ObjectMgr.h
Treeston: at that point I'd rather return nullptr, I want the definition to be transparent to the compiler

src/server/game/Globals/ObjectMgr.cpp
joschiwald: msvc uses `%Iu`
https://msdn.microsoft.com/en-us/library/tcxf1dw6.aspx

src/server/game/Globals/ObjectMgr.h
Treeston: (or explicitly pull `Errors.h` in (how big is that one?))

src/server/game/Globals/ObjectMgr.cpp
Treeston: they support `%zu` though, as do all the unix compilers

src/server/game/Globals/ObjectMgr.h
Treeston: :+1:

src/server/game/Maps/Map.h
```diff
@@ -271,7 +273,37 @@ struct ZoneDynamicInfo
 
 typedef std::map<uint32/*leaderDBGUID*/, CreatureGroup*>        CreatureGroupHolderType;
 
+struct RespawnInfo; // advance declaration
```
Shauren: its called forward declaration, not advance

src/server/game/Maps/Map.cpp
```diff
@@ -2867,6 +2951,325 @@ void Map::SendObjectUpdates()
...
+void Map::AddRespawnInfo(RespawnInfo& info, bool replace)
+{
+    if (!info.spawnId)
+        return;
+
+    RespawnInfoMap& bySpawnIdMap = GetRespawnMapForType(info.type);
+    
+    RespawnInfo*& existing = bySpawnIdMap[info.spawnId]; // get a reference so we only do the map lookup once
+
+    if (existing) // spawnid already has a respawn scheduled
```
Shauren:
```
auto itr = map.find(key);
if (itr == map.end())
    itr = map.emplace(std::piecewise_construct, std::forward_as_tuple(key), std::forward_as_tuple(value_ctor_args)).first;
```
thats the standard insert-into-map-if-not-exists-and-do-something-with-value pattern to avoid double lookups - that way you dont have to turn your value into a pointer

Treeston: I'm not turning it into a pointer - the map stores pointers anyway. However, will change.

src/server/game/Maps/SpawnData.h
```diff
@@ -0,0 +1,76 @@
...
+struct SpawnGroupTemplateData
+{
+    uint32 groupId;
+    std::string name;
+    uint32 mapId;
+    SpawnGroupFlags flags;
+    bool isActive;
+};
+
+struct SpawnData : public WorldLocation
```
Shauren: I don't like this abuse of Position inheritance, make this a member of the struct

Treeston: It allows us to pass it to any util that takes `Position` though, how is it abuse?

src/server/game/World/World.h
```diff
@@ -859,7 +860,21 @@ class TC_GAME_API World
         AutobroadcastsWeightMap m_AutobroadcastsWeights;
 
         void ProcessQueryCallbacks();
+
+        void SendGuidWarning();
+        void DoGuidWarningRestart();
+        void DoGuidAlertRestart();
         QueryCallbackProcessor _queryProcessor;
+
+        std::string respawnWarningMsg;
```
Shauren: confusing naming (all of them), they look like local variables when used

Treeston: not mine, but will fix

src/server/scripts/Commands/cs_gobject.cpp
```diff
@@ -38,6 +38,11 @@ EndScriptData */
 #include "RBAC.h"
 #include "WorldSession.h"
 
+struct npc_commandscript // partial declaration so we can map the spawn group command to both .npc and .gobject
```
Shauren: C++ does not support partial class/struct definitions, you are violating one definition rule. http://en.cppreference.com/w/cpp/language/definition

This would only be ok if npc_commandscript was a namespace

src/server/scripts/Commands/cs_npc.cpp
```diff
@@ -1546,6 +1542,86 @@ class npc_commandscript : public CommandScript
...
+            std::string thisArg = arg;
+            std::transform(thisArg.begin(), thisArg.end(), thisArg.begin(), ::tolower);
+            if (thisArg == "removerespawntime")
+                deleteRespawnTimes = true;
+            else if (thisArg.empty() || !(std::count_if(thisArg.begin(), thisArg.end(), ::isdigit) == (int)thisArg.size()))
+                return false;
+            else
+                groupId = atoi(thisArg.c_str());
+
+            arg = strtok(NULL, " ");
```
Shauren: use nullptr in new code

Shauren (changes requested):

src/server/game/Maps/Map.cpp
Shauren: That is what I meant, it is a pointer to support exactly this scenario (but does not have to be)

src/server/scripts/Commands/cs_gobject.cpp
Treeston: they're in different compilation units

src/server/game/Maps/Map.cpp
Treeston: It's a pointer because it should be consistent with the data in respawn heap. If I modify something on one container, I don't want to do a lookup on the other container to make sure both sets of data are consistent. That'd just be weird.

src/server/scripts/Commands/cs_npc.cpp
Treeston: :+1:

src/server/game/Maps/SpawnData.h
Shauren: But this is not "respawn location", this is "SpawnData" - and spawn data _has_ a position, not is one.

Treeston:
Rebased to get d24ce17 (cp of fdd9227)

src/server/game/Maps/Map.cpp
Shauren: From what I understood its the map that "owns" the RespawnInfo structure, heap can keep pointing to value stored inside the map

Treeston: Not saying it's an impossible change, I just don't see much benefit - anything that takes values from the map would still need to get a pointer (since it's possible for a respawninfo for a spawnid to not exist)

Treeston:
Static analysis now reports no issues in any of the code touched by the PR.

Treeston:
I actually went through the effort of compiling this branch in full debug configuration and running it in Valgrind, despite the 10 minute world load time that incurs. I tested the following scenarios:

* Regular gameplay. I cleared all of Dun Morogh (level 1-10) on a new character, completing quests and killing NPCs as normal. No irregularities were apparent, and no issues were detected by dynamic anaylsis.
* I added three static creature spawns using `.npc add`, then deleted one of them via `.npc delete`. Dynamic analysis reported no issue.
* I added the two remaining spawns to a spawn group using `MANUAL` flag, then restarted the core. I then manually spawned the group (using `.npc spawngroup`), then despawned it again. I then spawned it one final time and used `.npc delete` to delete both creatures. This caused an issue with a leftover database entry, as well as a dangling map entry, that will be addressed in the following commit.

Treeston:
And I think that's the last of the edge cases I can think of. `.npc delete` is such a stupid, stupid command. As is `.npc add`.

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -2298,6 +2298,21 @@ void ObjectMgr::LoadSpawnGroups()
...
 }
 
+void ObjectMgr::OnDeleteSpawnData(SpawnData const* data)
+{
+    if (!data->spawnGroupData->groupId)
+        return;
+    auto pair = _spawnGroupMapStore.equal_range(data->spawnGroupData->groupId);
+    for (auto it = pair.first; it != pair.second; ++it)
+    {
+        if (it->second != data)
```
Shauren: This code looks to me like Trinity::Containers::MultimapErasePair

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -2211,15 +2193,148 @@ void ObjectMgr::LoadGameobjects()
...
+}
+
+void ObjectMgr::OnDeleteSpawnData(SpawnData const* data)
+{
+    auto templateIt = _spawnGroupDataStore.find(data->spawnGroupData->groupId);
+    ASSERT(templateIt != _spawnGroupDataStore.end(), "Creature data for (%u,%u) is being deleted and has invalid spawn group index %u!", uint32(data->type), data->spawnId, data->spawnGroupData->groupId);
+    if (templateIt->second.flags & SPAWNGROUP_FLAG_SYSTEM) // system groups don't store their members in the map
+        return;
+
+    auto pair = _spawnGroupMapStore.equal_range(data->spawnGroupData->groupId);
```
Shauren: This code looks to me like Trinity::Containers::MultimapErasePair

Treeston:
Latest revision reports no issues in static analysis.

Treeston:
After spawngroup-delete-add-despawngroup-delete tests, no leaks related to this code are reported by valgrind.

Treeston (approved):
Static and dynamic analysis are clean, and all edge cases should be handled.

src/server/game/Globals/ObjectMgr.cpp
Treeston: mine terminates after the first found element (since I know the entries are unique)

morpheus4hire:
@Treeston 
When starting world server you get a bunch of these 
Loading Spawn Group Data...
Spawn with ID (1,21) is listed as a member of spawn group 4, but is already a member of spawn group 2. Skipping.
Spawn with ID (1,1801) is listed as a member of spawn group 4, but is already a member of spawn group 2. Skipping.
I only listed a few this didn't happen unil your latest commits.

Treeston:
Interesting - I'll take a look.

Treeston:
@morpheus4hire Issue should be resolved by the latest revision.

morpheus4hire:
@Treeston 
Yes it is thanks 👍

Shauren (approved):

Raydor:
WE READY BOYS, PRESS THE BUTTON!

bumbummen99:
What is with the bounty in #13787? @Treeston

Treeston:
I don't feel it'd be fair for me to claim that - @r00ty-tc can do it.

PR #19059, opened 2017-02-05
Core/Misc: Various crash fixes

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles

Fixed various crashes, also related to multithreading. Replaced not needed calls to FindPlayer with GetPlayer.

### DIFF
--- a/src/server/game/DungeonFinding/LFGMgr.h
+++ b/src/server/game/DungeonFinding/LFGMgr.h
@@ -304,7 +304,7 @@ class TC_GAME_API LFGMgr
 
         // World.cpp
         /// Finish the dungeon for the given group. All check are performed using internal lfg data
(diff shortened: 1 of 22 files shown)


src/server/game/DungeonFinding/LFGMgr.h
```diff
@@ -303,7 +304,7 @@ class TC_GAME_API LFGMgr
 
         // World.cpp
         /// Finish the dungeon for the given group. All check are performed using internal lfg data
-        void FinishDungeon(ObjectGuid gguid, uint32 dungeonId);
+        void FinishDungeon(ObjectGuid gguid, uint32 dungeonId, const Map* currMap);
```
Shauren: Map const*

PR #19070, opened 2017-02-07
Make MMapFactory and VMapFactory singleton creation thread safe indep…

chaodhib:
**Changes proposed:**

- Make MMapFactory and VMapFactory singleton creation thread safe independently of creation context. Does not solve any issue. Just cleaning up some old code. 

- [X] 3.3.5
- [X] master

**Issues addressed:** Does not solve any issue.

**Tests performed:** Build. Server starts correctly.


src/common/Collision/Management/MMapFactory.cpp
```diff
@@ -21,24 +21,10 @@
...
-    {
-        if (g_MMapManager)
-        {
-            delete g_MMapManager;
-            g_MMapManager = NULL;
-        }
+        static MMapManager g_MMapManager;
+        return &g_MMapManager;
     }
 }
```
Keader: new line xd

chaodhib: oups. fixed

jackpoz:
> Tests performed: Build. Server starts correctly.

could you please test tools too ?

joschiwald:
why not removing these factorys and declare the singletons how we do it everywhere?

chaodhib:
@joschiwald Factory pattern is used.  It was useful in the past (for VMAP). It doesn't hurt to keep it just in case it could be useful again in the future.

@jackpoz Done. Starts correctly too.

joschiwald:
I don't think it is useful, it is just a file with one singleton and it is
inconsistent to our other singletons

Shauren:
I don't see any benefit in this change - it actually has negative consequences (exceptionally rare). You see, current approach assumes that you can have different IVMapManager interface implementations (the last time we gained new one was with vmaps overhaul few years ago). If you are going to remove it then also remove the interface, drop 2 from VMapManager2 class name etc...
All that for no gain and only cluttering history

chaodhib:
> I don't see any benefit in this change
Are you talking about the use of the factory pattern or this PR?

This PR's gain is just to clean up the code as we go. In the long run, this approach helps to keep the code concise & clean wherever possible.

I will remove the factories, the interface and rename VMapManager2 to VMapManager tonight

Shauren:
And then you step right into the other point I made - cluttering history with a change that brings absolutely nothing

chaodhib:
I'm not going to argue the benefits of [code refactoring](https://en.wikipedia.org/wiki/Code_refactoring) with you _again_, Shauren. Because yes, I remember arguing with you about the same thing in the past. I already spend enough time harassing people about code quality at my day job . I'm not going to do it when I code for fun.

I've done the changes. Builds and runs correctly (server & tools). The fact is: It's 200 less LOC and 5 unneeded files removed.

PS: Since you know C++ better than me, what would had been more constructive is access whether or not the "negative consequences" that you mentioned (= change of the order of the destructor calls which may or may not cause issues) are real.

Shauren:
You should damn well know what my opinion on refactoring only PRs is. Just NO.

jackpoz:
none thinks about my CLI branch :(

btw https://travis-ci.org/TrinityCore/TrinityCore/builds/202077771#L867

chaodhib:
@jackpoz your CLI branch?

src/common/Collision/Management/VMapManager.cpp
```diff
@@ -33,14 +33,21 @@ using G3D::Vector3;
...
         thread_safe_environment = true;
     }
 
-    VMapManager2::~VMapManager2(void)
+    // using c++ 11 static initializer, we have the guarantee of creating only one MapManager instance (aka singleton)
+    VMapManager* VMapManager::createOrGetVMapManager()
+    {
+        static VMapManager gVMapManager;
+        return &gVMapManager;
+    }
```
joschiwald:
```c++
VMapManager& VMapManager::Instance()
{
    static VMapManager instance;
    return instance;
}
```
for more consistence

chaodhib: createOrGetXMapManager() returns a pointer. not a reference. To change that would require to modify the code of every calls to this method. I would like to avoid that.

chaodhib:
I'm dropping this. This is minor changes and since there is opposition, I don't want to spend more time on this.

PR #18910, opened 2017-01-21
Ensure that all actions are compared to fixed point in time (ie. world update start)

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Issues addressed:** Closes #10530, can potentially fix #10218 if arming time / 2 fix is correct

**Tests performed:** Compiles, simple ingame tests

Ensure that all actions are compared to fixed point in time (eg. world update start) and are not dependent on processing moment. Increased GameObjects cooldown resolution to milliseconds, fixes arming time of traps to be exactly one second and not something from range (1000, 1999)

PR #19101, opened 2017-02-10
Core/GameObject: regenerate GO loot if respawn time passed since last loot generation time and not all items were taken

### LINKED ISSUE #8952: Looting issue with gameobject type 3 for quest items (cause by players)
I would like to report an issue which I have encountered with gameobjects (type 3) which contain items which are required for quests,

an example object is http://www.wowhead.com/object=290 which contains http://www.wowhead.com/item=841 which is required for http://www.wowhead.com/quest=64/the-forgotten-heirloom

But have previously encountered this issue with other gameobjects, the issue is if a player opens the gameobject but does not loot the quest item then another player trys to loot the object it will appear empty, if the player who originally opened it goes back and loots the item then it will work again but this wont always happen and object will end up tagged to that player and noone else will be able to loot quest item until server restart unless the player to which it is tagged loots item.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/104231-looting-issue-with-gameobject-type-3-for-quest-items-cause-by-players?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Issues addressed:** Closes #8952

**Tests performed:** Compile and simple ingame test

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.h
+++ b/src/server/game/Entities/GameObject/GameObject.h
@@ -794,6 +794,8 @@ class TC_GAME_API GameObject : public WorldObject, public GridObject<GameObject>
         void AddLootMode(uint16 lootMode) { m_LootMode |= lootMode; }
         void RemoveLootMode(uint16 lootMode) { m_LootMode &= ~lootMode; }
         void ResetLootMode() { m_LootMode = LOOT_MODE_DEFAULT; }
+        void SetLootGenerationTime() { m_lootGenerationTime = time(NULL); }
+        uint32 GetLootGenerationTime() const { return m_lootGenerationTime; }
 
         void AddToSkillupList(ObjectGuid::LowType PlayerGuidLow) { m_SkillupList.push_back(PlayerGuidLow); }
         bool IsInSkillupList(ObjectGuid::LowType PlayerGuidLow) const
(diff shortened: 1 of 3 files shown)


src/server/game/Entities/GameObject/GameObject.h
```diff
@@ -789,6 +789,8 @@ class TC_GAME_API GameObject : public WorldObject, public GridObject<GameObject>
         void AddLootMode(uint16 lootMode) { m_LootMode |= lootMode; }
         void RemoveLootMode(uint16 lootMode) { m_LootMode &= ~lootMode; }
         void ResetLootMode() { m_LootMode = LOOT_MODE_DEFAULT; }
+        void SetLootGenerationTime() { m_lootGenerationTime = time(NULL); }
```
DDuarte: Should this receive the same treatment as #18910 or it doesn't matter in this case?

xinef1: Should, but in this case its rather irrelevant (duration that is mostly measured in minutes)

Tonghost:
:shipit:

Keader:
nice

PR #19105, opened 2017-02-11
Core/Spells: Corrected points per level calculation for some spells

ghost:

- [x] 3.3.5

**Tests performed:** Compiles


tkrokli:
Spell Level: L-E-V-E-L [(Order of the Stick joke ^^)](http://www.giantitp.com/comics/images/oots0012.gif)

PR #19121, opened 2017-02-12
Core/Misc: Added some missing SetFallInformation calls and some more

ghost:
**Target branch(es):** 3.3.5

- [x] 3.3.5

**Tests performed:** Compiles, simple ingame tests

Added some missing SetFallInformation calls
Fixed HP drop bug on zeppelins
Fixed possible packet relocation problems on huge transports

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -1929,7 +1929,7 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
 
         // this will be used instead of the current location in SaveToDB
         m_teleport_dest = WorldLocation(mapid, x, y, z, orientation);
-        SetFallInformation(0, z);
+        SetFallInformation(0, GetPositionZ());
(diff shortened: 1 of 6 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -1929,7 +1929,7 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
 
         // this will be used instead of the current location in SaveToDB
         m_teleport_dest = WorldLocation(mapid, x, y, z, orientation);
-        SetFallInformation(0, z);
+        SetFallInformation(0, GetPositionZ());
```
Shauren: wont using the position from before teleportation cause fall damage?

xinef1: We are not yet teleported, don't see any reason why we should relocate fallZ to teleportZ. This is done when teleport ack packet is received (movement cannot be processed while being teleported)

PR #19148, opened 2017-02-17
Core/Instances: Loading boss state as FAIL now defaults to NOT_STARTED

Krudor:

**Changes proposed:**

-  Set boss state to NOT_STARTED if it loads as FAIL

**Tests performed:** Builds, tested in-game


Kittnz:
Any reason for this change?

Krudor:
It's mainly causing some unwanted behavior in a script I'm working on because it's, as the status IN_PROGRESS, not expected to stay in that state upon loading instance saves.

It is in most cases not a problem because most bosses set their status manually upon Reset to NOT_STARTED, but some bosses don't.

FAIL is much like IN_PROGRESS, not a state that an encounter state should be loaded into.

Another question is why these states are even saved to DB in the first place.

PR #19076, opened 2017-02-08
Core/Scripts: implement boss Apothecary Hummel ("Love is in the Air" event boss)

sirikfoll:
…nt boss)

**Changes proposed:**

-  Implement the fight and add the related objects

**Target branch(es):** 3.3.5/master
- [x] 3.3.5

**Issues addressed:** Closes #5114

**Tests performed:** Built

**Known issues and TODO list:** (add/remove lines as needed)

- [x] Gossip menu option don't appear, find out the correct text too (help needed)
- [x] Clean-up
- [x] Research Crazed Apothecary spawns (dropped, it was only existing in 3.3.5 and there's no sniffs of it to find out correct spawn locations)
- [x] Spells Perfume and Cologne Spill doesn't do damage (help needed)
The spells are 68798 and 68614
```
Effect 0: Id 27 (SPELL_EFFECT_PERSISTENT_AREA_AURA)
BasePoints = 250
Targets (18, 8) (TARGET_DEST_CASTER, TARGET_UNIT_DEST_AREA_ENTRY)
Aura Id 226 (SPELL_AURA_PERIODIC_DUMMY), value = 250, misc = 0 (0), miscB = 0, periodic = 1000
Radius (Id 15) 3,00
```
They have to do damage to players in range, the damage spells are 68927 and 68934
```
Effect 0: Id 2 (SPELL_EFFECT_SCHOOL_DAMAGE)
BasePoints = 2375 to 2626
Targets (25, 0) (TARGET_UNIT_TARGET_ANY, NO_TARGET)

```

PR #19168, opened 2017-02-19
Core/Entities Allow DynamicObjects with UNIT_FLAG_IMMUNE_TO_PC to select targets

sirikfoll:
Fix suggested by xinef1

**Changes proposed:**

- GameObjects will select targets regardless of having UNIT_FLAG_IMMUNE_TO_PC

- [x] 3.3.5

**Issues addressed:** Fix one issue with #19076

**Tests performed:** Built and tested with #19076


xinef1:
Maybe it would be more readable if we wrap this condition in a separate bool?

Kittnz:
Can you update this @sirikfoll ?

sirikfoll:
How?

Kittnz:
What xinef1 wrote, more readable i guess

sirikfoll:
@xinef1  could you take this over please?

PR #19171, opened 2017-02-19
Areatrigger rework script system + fix priest angelic feather

Traesh:
**Changes proposed:**

- Rework AreaTrigger's script system to add "AreaTriggerAI" class
- Add default values for AREATRIGGER_EXTRA_SCALE_CURVE, work for 90% areatriggers and fix visuals
- Add http://www.wowhead.com/spell=121536/angelic-feather spell/areatrigger script

**Target branch(es):**

- [ ] 3.3.5
- [x] master

Built, tested in-game

- [ ] Todo in future : Add AREATRIGGER_EXTRA_SCALE_CURVE to DB ?


Takenbacon:
As far as I know, wasn't angelic feather changed in WoD to prioritize the caster? (Probably needs to be researched).

Traesh:
Indeed, thanks for the infos ! It explain the dummy :)

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -1327,6 +1330,75 @@ class spell_pri_power_word_shield : public SpellScriptLoader
...
+        };
+
+        SpellScript* GetSpellScript() const override
+        {
+            return new spell_pri_angelic_feather_trigger_SpellScript();
+        }
+};
+
+// Angelic Feather areatrigger - created by SPELL_PRIEST_ANGELIC_FEATHER_AREATRIGGER
+class areatrigger_pri_angelic_feather : public AreaTriggerEntityScript
```
Shauren: ugh, this is really shit (not your code) - AreaTriggerEntityScript being a global unbound script, can you see every single script being executed for every single areatrigger that we spawn? this cant be solved like this

Traesh: Well it's not totally unbound, it's linked with an areatrigger by it's areatrigger_template's scriptname.

But i do see your point, i'm creating an AreaTriggerAI and i'll probably transfom this PR when finished :)

Shauren: yeah well, with current approach you can't store any extra state in the script
but you shouldn't give it *AI* name

Traesh: Damn, saw your comment too late. I created an "AreaTriggerAI" class, why shouldn't i give it the "AI" name ?

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -1327,6 +1331,99 @@ class spell_pri_power_word_shield : public SpellScriptLoader
...
+            {
+                if (caster->IsFriendlyTo(unit))
+                {
+                    // If target already has aura, increase duration to max 130% of initial duration
+                    if (Aura* aura = unit->GetAura(SPELL_PRIEST_ANGELIC_FEATHER_AURA))
+                    {
+                        int32 newDuration = aura->GetDuration() + aura->GetMaxDuration();
+                        int32 maxPossibleNewDuration = CalculatePct(aura->GetMaxDuration(), 130);
+
+                        aura->SetDuration(newDuration > maxPossibleNewDuration ? maxPossibleNewDuration : newDuration);
```
matukaa: hm, shouldn't this be implemented elsewhere? sometime back in WoD, they added this feature, that upon reapply auras can be increased in duration to a maximum of 130% of the original duration. it's not incorrect, but it should be something more general, we can't possibly do this feature for all spells from inside a script

matukaa: nvm, my mistake, the original blue post sounds like this:
"Recasting periodic damage over time and healing over time effects that are already on the target now extends those effects to up to 130% of the normal duration of the effect."
keep the code, my previous comment does not apply here

src/server/game/AI/CoreAI/AreaTriggerAI.cpp
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
+ * Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
+ *
```
Aokromes: new file no mangos (C)

src/server/game/AI/CoreAI/AreaTriggerAI.h
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
+ * Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
```
Aokromes: new file no mangos (C)

src/server/scripts/Spells/spell_priest.cpp
Traesh: Indeed, Gamepedia made me think it was a specific feature.

I delete this code, we'll have to handle it somewhere else in a more general way as you said.

src/server/scripts/Spells/spell_priest.cpp
Shauren: because its not really AI - how would you even imagine AI for something such fully reactive like area triggers?

src/server/game/AI/CoreAI/AreaTriggerAI.h
```diff
@@ -0,0 +1,58 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef TRINITY_AREATRIGGERAI_H
+#define TRINITY_AREATRIGGERAI_H
+
+#include "Define.h"
+#include "Object.h"
+#include "AreaTrigger.h"
```
Shauren: nope

src/server/game/AI/CoreAI/AreaTriggerAI.h
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef TRINITY_AREATRIGGERAI_H
+#define TRINITY_AREATRIGGERAI_H
+
+#include "Define.h"
+#include "Object.h"
```
Shauren: nope

src/server/game/Entities/AreaTrigger/AreaTrigger.h
```diff
@@ -37,6 +40,12 @@ class TC_GAME_API AreaTrigger : public WorldObject, public GridObject<AreaTrigge
         void AddToWorld() override;
         void RemoveFromWorld() override;
 
+        bool AIM_Initialize(AreaTriggerAI* ai = nullptr);
```
Shauren: do you really want to support changing the script?

src/server/game/Entities/AreaTrigger/AreaTrigger.h
```diff
@@ -21,11 +21,14 @@
 #include "Object.h"
 #include "Position.h"
 #include "Spline.h"
+#include "AreaTriggerAI.h"
```
Shauren: dont include

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -839,34 +840,14 @@ class TC_GAME_API GroupScript : public ScriptObject
...
-        virtual void OnUnitEnter(AreaTrigger* /*areaTrigger*/, Unit* /*unit*/) { }
+    AreaTriggerEntityScript(const char* name);
 
-        // Called when an unit exit the AreaTrigger, or when the AreaTrigger is removed
-        virtual void OnUnitExit(AreaTrigger* /*areaTrigger*/, Unit* /*unit*/) { }
+public:
 
-        // Called when the AreaTrigger is removed
-        virtual void OnRemove(AreaTrigger* /*areaTrigger*/) { }
+    // Called when a GameObjectAI object is needed for the gameobject.
```
Shauren: copypaste comment is strong here

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -839,34 +840,14 @@ class TC_GAME_API GroupScript : public ScriptObject
...
+    virtual AreaTriggerAI* GetAI(AreaTrigger* /*at*/) const { return NULL; }
```
Shauren: nullptr

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -1327,6 +1331,99 @@ class spell_pri_power_word_shield : public SpellScriptLoader
...
+        {
+            if (Unit* caster = at->GetCaster())
+            {
+                std::vector<AreaTrigger*> areaTriggers = caster->GetAreaTriggers(SPELL_PRIEST_ANGELIC_FEATHER_AREATRIGGER);
+
+                if (areaTriggers.size() >= 3)
+                    areaTriggers.front()->SetDuration(0);
+
+                // Caster is prioritary
+                if (caster->IsWithinDist(at, at->GetTemplate()->SphereDatas.Radius))
```
Shauren: is the areatrigger even created when targeting the caster on retail?

src/server/scripts/Spells/spell_priest.cpp
Traesh: Well it make sense but i saw it more like a convention.
For Gameobject we use "GameObjectAI", still gobs are also fully reactive

src/server/scripts/Spells/spell_priest.cpp
matukaa: no, it is not, just sniffed :)
also, could use "MaxSearchRadius" as well, more generic (although it's just codestyle, functionality is the same)

Shauren: then the "if caster" should just be handled in the dummy, casting the buff directly

src/server/scripts/Spells/spell_priest.cpp
Traesh: Indeed this is a special case, but i think you're still right, it should be handled in a more general way :) I removed the code and i'll check this later, it's a minor detail

src/server/game/AI/CoreAI/AreaTriggerAI.h
```diff
@@ -18,9 +18,8 @@
 #ifndef TRINITY_AREATRIGGERAI_H
 #define TRINITY_AREATRIGGERAI_H
 
-#include "Define.h"
```
Shauren: define is still needed (for TC_GAME_API and int types)

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -164,17 +192,11 @@ void AreaTrigger::Update(uint32 p_time)
...
-    if (GetDuration() != -1)
+    if (IsAIEnabled)
     {
-        if (GetDuration() > int32(p_time))
-            _UpdateDuration(_duration - p_time);
-        else
-        {
-            Remove(); // expired
-            return;
-        }
```
joschiwald: why did you moved this block? i think we should update position and worldobject before despawn to have actual values available in remove hooks

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -839,34 +840,14 @@ class TC_GAME_API GroupScript : public ScriptObject
...
-        virtual void OnCreate(AreaTrigger* /*areaTrigger*/) { }
-
-        // Called on each AreaTrigger update
-        virtual void OnUpdate(AreaTrigger* /*areaTrigger*/, uint32 /*diff*/) { }
-
-        // Called when the AreaTrigger reach splineIndex
-        virtual void OnSplineIndexReached(AreaTrigger* /*areaTrigger*/, int /*splineIndex*/) { }
-
-        // Called when the AreaTrigger reach its destination
-        virtual void OnDestinationReached(AreaTrigger* /*areaTrigger*/) { }
```
joschiwald: imo we should keep these hooks, maybe someone has use for it

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
```diff
@@ -164,6 +164,8 @@ class AreaTriggerMiscTemplate
         TimeToTarget        = 0;
         TimeToTargetScale   = 0;
 
+        ExtraScaleCurve     = { 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 1.0000001f, 1.0f };
```
joschiwald: ExtraScaleCurve6 is wrong, should be uint(1)

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
matukaa: I agree with @joschiwald, OnRemove and OnUnitExit (in case of Remove()) will be behind with an update cycle

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
Traesh: Nope, ExtraScaleCurve6 is a float (example : [5] AREATRIGGER_EXTRA_SCALE_CURVE + 6: 1073741831/2.000002) with default value 1

src/server/game/Scripting/ScriptMgr.h
Traesh: Well it's pure duplicate and it's not available since a long time, switching to new system only require two copy/paste, i don't think we should keep it

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
joschiwald: AREATRIGGER_EXTRA_SCALE_CURVE + 5: 1065353217/1
AREATRIGGER_EXTRA_SCALE_CURVE + 6: 1/1.401298E-45

that are the defaults

Traesh: Well the value still can be a float, so it's more logical to define it as float no ?

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
Traesh: I of course had a very good reason, i just can't remember it, so let's do like it never happened

src/server/game/Entities/AreaTrigger/AreaTrigger.h
Traesh: Not especially, is this a bad thing ?

Shauren: complicates the code a bit - by not exposing such functionality (no SetAI) you dont also need to care about _AI_locked (maybe you don't even need IsAIEnabled, just check for null)
also note for naming
AIM (M is movement, creature calls Motion_Initialize there)

actually you dont need 80% of that creature copypasta, method arg is unneeded, AIM_Destroy is unneeded

Traesh: Noted, thanks for the explainations

src/server/game/Entities/AreaTrigger/AreaTrigger.h
Traesh: It's the only  '#include "AreaTriggerAI.h"' left, remove it won't break non-PCH build ?

Naios:
You also need to add the functionality to the `ScriptMgr` that it unloads and reloads the scripts correctly from areatriggers. This is required because you made the AreaTrigger scripts bound to a specific entity.

Traesh:
@Naios I take creature/gob as example, it seem good to me (it work ingame), could you please check and warn me if i missed something ?

Naios:
@Traesh Seems good

streetrat:
merge it?

joschiwald:
@streetrat if you know more about extra scale curves ;)

Traesh:
@joschiwald The point of this PR is to set default value to debug visuals in most case, not to fix all extra scale curves :)

joschiwald:
default values are still wrong, see https://github.com/TrinityCore/TrinityCore/pull/19171#discussion_r103997287

Traesh:
merge it? :p

Aokromes:
ERROR 1064 (42000) at line 797773: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'UPDATE `smart_scripts` SET `event_flags`=`event_flags`|0x200 WHERE  `entryorguid' at line 2

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
```diff
@@ -74,6 +75,30 @@ struct AreaTriggerAction
...
 
+struct AreaTriggerScaleInfo
+{
+    AreaTriggerScaleInfo()
+    {
+        memset(OverrideScale, 0, sizeof(OverrideScale));
+        memset(ExtraScale, 0, sizeof(ExtraScale));
+
+        ExtraScale[5].AsFloat = 1.0000001f;
+        ExtraScale[5].AsInt32 = 1;
```
Shauren: you are setting the same memory location twice - 1.0000001f is never used

Traesh: Yes, was a typo sorry, fixed in https://github.com/TrinityCore/TrinityCore/pull/19171/commits/08bc8b58fc126329fd564a4f78748c11b63adc4f

src/server/game/Entities/AreaTrigger/AreaTrigger.h
Shauren: you will need to add it in .cpp files that need it (instead of all that include AreaTrigger.h but don't need the AI header)

Seyden:
Merge it?

PR #19183, opened 2017-02-22
Core/Battlegrounds: disable health regen for battleground vehicles

### LINKED ISSUE #950: [Core/Entities] Regenerating HP of vehicles part 2
CORE revision number : e4c0a8033680e
Database: TDB 335.11.38
Patches: None

Continuation from this ticket http://www.trinitycore.org/t/trinitycore/ticket/317 it was not fixed. New things i have learned:
- It seems only in raids in battlegrounds there is not applied this condition to not regenerate HP from vehicles, because in one by one testing it seems to be working sometimes vehicle is destroyed after reload vehicle with less HP.

What steps will reproduce the problem?
1. Get on vehicle for example in sota
2. Damge this vehicle to have less HP
3. Get off and on vehicle again it has full HP even when in DB is disabled 

It seems it is NOT DB problem because when I set RegenHealth=0 it is still regenerating after reload the vehicle.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/66411-core-entities-regenerating-hp-of-vehicles-part-2?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sirikfoll:
**Changes proposed:**

-  Disable health regeneration for battlegrounds vehicles

- [x] 3.3.5

**Issues addressed:** Partially fixes #950

**Tests performed:** Built and tested in **Strand of the Ancients** with the Demolishers

### DIFF
--- a/src/server/game/Battlegrounds/Battleground.cpp
+++ b/src/server/game/Battlegrounds/Battleground.cpp
@@ -1526,6 +1526,9 @@ Creature* Battleground::AddCreature(uint32 entry, uint32 type, float x, float y,
     if (respawntime)
         creature->SetRespawnDelay(respawntime);
 
+    if (creature->IsVehicle())
+        creature->setRegeneratingHealth(false);
+
     return creature;
 }
 


Noryad:
Incomplete, Battlefield demolisher can not regenerate hp, but if the vehicle dies later appears dead.

Re3os:
@Noryad try this 
```sql 
diff --git a/src/server/game/Battlegrounds/Battleground.cpp b/src/server/game/Battlegrounds/Battleground.cpp
index 628d7ae..e105bbb 100644
--- a/src/server/game/Battlegrounds/Battleground.cpp
+++ b/src/server/game/Battlegrounds/Battleground.cpp
@@ -1693,6 +1693,12 @@ Creature* Battleground::AddCreature(uint32 entry, uint32 type, float x, float y,
     if (respawntime)
         creature->SetRespawnDelay(respawntime);
 
+    if (creature->IsVehicle())
+    {
+        if (creature->IsAlive())
+            creature->setRegeneratingHealth(false);
+    }
+    
     return creature;
 }
 
```

Shauren:
@Noryad It is natural for dead creature to appear dead...

Jildor:
@Shauren, the Battlegrounds respawn the vehicles in X period of time, and should appear alive, when vehicles respawn appear dead

(the same if put regenHealth=0 in creature_template)

sirikfoll:
@jildor @noryad This was fixed yesterday, please test again

Keader:
Test after https://github.com/TrinityCore/TrinityCore/commit/b3b4054eb44fdb9b0efe47d1a7ec3021b87a5fed

Noryad:
After b3b4054 it works ok :smile:

Jildor:
Mmm not is more easy change it in DB?

And should be only bg vehicles or all?

Aokromes:
db only changed failed to stop vehicles hp regeneration.

Jildor:
@Aokromes I think is the same xD

Aokromes:
from comment of ticket: "
It seems it is NOT DB problem because when I set RegenHealth=0 it is still regenerating after reload the vehicle."

Jildor:
You are free to try it xD
The comment you are dating is old, I try it change ir in DB and the result is the same (obviously, with this fix will only work if spawn the BG)
So that's why I ask if it's not better than do it by DB

But....It's ok xD

Demonid:
Why isn't this merged?

Noryad already stated that after b3b4054eb44fdb9b0efe47d1a7ec3021b87a5fed this PR works fine, changing it in DB would not be correct cuz TrinityCore, so everything seems fine for a merge.

ariel-:
This fix isn't correct. We keep on patching Battleground::AddCreature. But the real issue is, supposedly, not honoring RegenHealth flag in DB.

ariel-:
And that issue is non existant. So we can safely revert this patch and update affected creature templates in DB

PR #19214, opened 2017-02-27
Contrib: add debian start/stop scripts

MarbasFR:
Adding start/stop script for debian => /etc/init.d/trinity

**Changes proposed:**

-  Adding a script to start/stop/show bnetserver and worldserver

- [ ] 3.3.5
- [x] master

**Issues addressed:** 

**Tests performed:** 
- [x] tested on debian 8

- [x] screen must be installed in order to work => apt-get install screen


DDuarte:
>  screen must be installed in order to work => apt-get install screen

I think a better solution would be supporting daemon mode in TC (like it already happens on Windows)

MarbasFR:
@DDuarte sure, but for the moment i was needed something to start my new compiled server so i share my script :)

DDuarte:
Alright, one more thing for my wishlist :)

Btw, I think it would make sense to have two init.d files, one for worldserver and one for bnet/auth. There's no requirement to run both at the same time or in the same machine.

ariel-:
Doesn't Debian have an alternatives system? You can use it to call a symlink instead of screen, so users may keep using their favorite multiplexer (like tmux)

tkrokli:
It might not seem important for Linux files in the `contrib` folder, but make a habit of ending your files with a `NewLine` (Enter) character. Will be easier to remember doing it when you edit source & DB files later.

MarbasFR:
@ariel- In fact, i don't really know Debian :$ I've done these scripts because i was needed a solution to start Trinity on my test server. I didn't know tmux before you said me lol

Aokromes:
From one friend debian dev:

```
Adding a sysvinit init script is not a good idea as it's a deprecated init system that all major distributions, including Debian and Ubuntu, have migrated away from. These init script should be instead a systemd service file, which is a lot easier to write and maintain, and can additionally be tweaked to add security features to the started daemon, like restriction of syscalls, capabilities and other permissions. See https://lwn.net/Articles/709755/ for some examples.

PR #19247, opened 2017-03-05
Core/Scripts: Add support for spawn specific C++ scripts

Krudor:
**Changes proposed:**

-  Add script name column to creature and gameobject table to enable spawn specific C++ scripts

**Tests performed:** Builds, in-game tested

The system will now check if there's a valid script connected to a specific creature/gameobject spawn and default to the template script if none was found.


Krudor:
Here's a suggestion for an added script support for any creatures that need special C++ Scripted AI.
This is one proposal for changes that needs to be made to add more spawn specific control to C++ scripts.

Currently we're not allowed to use db guids for targetting specific creature/gameobject spawns.
This change would add support for giving a specific spawn its own C++ script.
It would however not fix the fact that we sometimes need to target specific creature spawns for spells or other events.

I've encountered many cases where there are generic triggers spawned with identical entry ids, that are targets for different spells.
It's currently a bad idea to use db guid for targetting these spawns as the db guids can change at any time, and scripts could be trying to target dead or invalid db guids.
So the proposal is for a new table that maps db guids to a static entry that will never change once assigned, unless removed from the script using it.
That means if you change spawn guids in the database, there's no need to update the scripts using these these guids, only the new table referencing them.
This new id would pretty much just be a secondary, C++ script safe db guid, to combat the ever changing db guids.

In a best case scenario, db guids shouldn't change once assigned, and we should be able to use the db guid, but we do have the need for a guid rebase once in a while since TrinityCore is always changing, and this is why I think this table is needed.

I will make a separate PR for this suggestion but thought I'd mention it here, as this PR is only a partial potential fix to the bigger issue of targetting specfic spawns.

If you're still with me so far, thanks for reading my [pile of text](https://www.youtube.com/watch?v=63UlBsdElsY)!

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2500,6 +2500,9 @@ std::string Creature::GetScriptName() const
 
 uint32 Creature::GetScriptId() const
 {
+    if (CreatureData const* creatureData = GetCreatureData())
```
Warpten: that is pretty much hiding the template scripts, afaik, since CreatureData is never null for db spawns. you should check `scriptID != 0` rather than a null pointer. same for gameobjects

Krudor: It's never null for db spawns, but not all are db spawns. Battlegrounds create creatures without creature data.
And yes it is hiding intentionally hiding template script id :)

Warpten: My point is, if there is no script defined in `creature|gameobject`, it won't fallback to the one in `(creature|gameobject)_template`.

Krudor: Check ObjectMgr changes.
If we can't find a valid script from the creature table, we check creature_template.

Warpten: Oh. Duh.

Aokromes:
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Spells/SpellEffects.cpp:5222:43: fatal error: 
      unused parameter 'effIndex' [-Wunused-parameter]
void Spell::EffectTitanGrip(SpellEffIndex effIndex)
                                          ^
1 error generated.
make[2]: *** [src/server/game/CMakeFiles/game.dir/Spells/SpellEffects.cpp.o] Error 1


/home/travis/build/TrinityCore/TrinityCore/src/server/game/Loot/LootMgr.cpp:405:119: fatal error: 
      comparison of integers of different signs: 'int32' (aka 'int') and
      'uint32' (aka 'unsigned int') [-Wsign-compare]
  ...&& int32(player->GetItemCount(itemid, true)) >= pProto->GetMaxCount())
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~

Keader:
👍

src/server/game/Entities/Creature/Creature.h
```diff
@@ -543,6 +543,7 @@ struct CreatureData
     uint32 dynamicflags;
     uint32 phaseid;
     uint32 phaseGroup;
+    uint32 ScriptID;
```
ariel-: Uninitialized variable. Causes erratic behaviour: #19405

PR #19264, opened 2017-03-09
Core/Misc : Creature Scaling System

### LINKED ISSUE #17898: [7.x] Creature Scaling System (100$ bounty)
**Description**:
In Legion, the creatures in Dungeons and leveling zones scale to your level, but only visible and occuring to you.

http://www.polygon.com/2015/11/6/9684580/world-of-warcraft-legion-zones-level-scaling

Example:
Player1 Level 101
Player2 Level 110
Moblevel seen by Player1 = 101
Moblevel seen by Player2 = 110
(this is the Same unique mob at same time)

**Current behaviour**: Nothing

**Expected behaviour**: 
- Damage done must be adapted
- Damage received must be adapted
- Health must be adapted
- Experience must be adapted
- everything i forgot to point here

How is level and damage being influced by group with different character levels?
Does it scale with itemlevel? Artifact Scenarios mobs scaling with itemlevel?

**Branch(es)**: master

**TDB version**:  703.00

**Bountysource**
https://www.bountysource.com/issues/37569917-7-x-creature-scaling-system
[Bountysource](https://www.bountysource.com/issues/37569917-7-x-creature-scaling-system?utm_source=37569917&utm_medium=shield&utm_campaign=ISSUE_BADGE)

Traesh:
**Changes proposed:**

-  Implementing legion creature scaling system.
- Add new table "creature_template_scaling" with scaling delta
- If the creature has it's entry in "creature_template_scaling", the levels will scale between template.minlevel and template.maxlevel

- [ ] 3.3.5
- [x] master

**Issues addressed:** Closes #17898

**Tests performed:**
Compiled, tested ingame with NPC 23836

`INSERT INTO creature_template_scaling VALUES (23836, 0);`


sql/updates/world/master/9999_99_99_99_world.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `creature_template` ADD COLUMN `levelScalingDelta` SMALLINT(5) DEFAULT 0 NOT NULL AFTER `maxlevel`, CHANGE `flags_extra` `flags_extra` BIGINT(20) UNSIGNED DEFAULT 0 NOT NULL;
```
joschiwald: what about a new table `creature_template_scaling` with the 3 level fields

Traesh: It was another valable idea that i discussed with funjoker, but i felt it would be a duplicate of min & max levels

Eredar:
Tested it ingame, overall it's fine, but I found few issues:
1. Incoming damage lowered by target level, but numbers in combatlog not lowered. So for example as level 100 I'm getting 50k portrait and combatlog hits from rare legion npc but it hits for about 5% of 300k healthpool. Can post a screenshot if needed.
2. Npc levels should not adjust instantly when target reaches new level. Instead it should be done only when player have above 50% of current level exp.

Traesh:
@Eredar Thanks for testing !

Do you have any sniff of point 2 ? Level switching is handled client side, i'd like to see if the feature you're talking about is handled by PLAYER_FIELD_SCALING_PLAYER_LEVEL_DELTA

Shauren (dismissed):

Seyden:
Instead of an extra flag, i would do a new table.. so if there is a row for the entry, you know that this creature does have scaling data.

Traesh:
@Seyden I hesitated too so i'm ready to make a new table if most people feel better with it. The thing making me add an extra flag was that this new table would only have "entry - levelScalingDelta" value, i felt that a table for 1 value was a little overkill

src/server/game/Miscellaneous/Formulas.h
```diff
@@ -176,7 +176,10 @@ namespace Trinity
             {
                 float xpMod = 1.0f;
 
-                gain = BaseGain(player->getLevel(), u->getLevel());
+                if (creature && creature->HasScalableLevels())
```
Seyden: Instead of doing an if for all things, i would change all getLevels to GetlevelForTarget and then let the virtual override handle the rest.. put the scaling if in the GetLevelForTarget in the creature override and let it handle there.

Seyden:
@Traesh nope, well.. i thought first too about using min and maxlevel for it.. but thats too confusing, other creatures use it as random level between this and.. i think its cleaner with minlevel, maxlevel and delta in a new table

Traesh:
@Seyden Then what do you put in template min/maxlevel ?

- Same values ? Then duplicate, should better using them
- only max level ? That doesn't make sense to me
- both at 0 or both at 1 ? That's not true, it's not real min/max levels of creature and would be confusing if dev don't bother to look at scaling table

Seyden:
I use entry, minlevel, maxlevel and deltalevel for the table

if i remember correctly, there was a reason why i didn't used the minlevel, maxlevel in creature_template for it..

Traesh:
@Seyden Yes i understand that you are using the values from the new table, my question is "then which value do you set in creature_template minl/max levels ?"

In your version, those values are not used anymore. What do you recommend so the dev seeing creature_template values know "ah yeah, modifying those values won't do anything" ?

That's some real questions, as i said  i also hesitated about creating a new table or no because of data integrity, i'm willing to change it if i'm wrong :)

Seyden:
I would use the sniff maxlevel value for the creatures that doesn't have any scaling data and any minlevel data

Shauren:
@Traesh because UNIT_FIELD_LEVEL does not have to always be equal to maxlevel (as long as it is in the range is ok)

Seyden:
@Shauren if a creature does have a scalable level, the UNIT_FIELD_LEVEL is always maxlevel

ALWAYS! Sniff confirmed

Traesh:
Yup, confirmed with sniffs of all legion zones while pexing

Eredar:
@Traesh About 2. I will try to find it in sniffs, but it's simply my observation. First I thought that it was some time delay, but after getting same behaviour for 9 levels and even after relog I'm 100% sure that while you have below 50% of level - npcs have your previous one.

Traesh:
@Eredar Thanks for the infos, i'm doing some tests ingame right now.

1 is fixed, and 3 isn't related to scaling as you said

Eredar:
PLAYER_FIELD_SCALING_PLAYER_LEVEL_DELTA: 4294967295/NaN
This string occured few times during two levelups sniff.

Traesh:
@Eredar 2 fixed :)

Traesh:
@funjoker I really don't understand what's difficult about understanding that duplicate values in a DB is just wrong ^^

They. are. the. same. data.

In one case, the core should handle them a "rand between them", and for scaling, it should be "the creature should be able to scale between thoses values", yet the data remain the same. As i said on IRC, it's just like saying "we should have 30 gameobject_template tables because data0 are handled differently depending of the gameobject type"

Creature_template of pretty much every legion template have level 1, including non-scaling creature, what's your point ?

funjoker:
Well isnt this a hint that actually the Scaling stuff is non Creature_template data ?

Traesh:
It's just the hint that no level has been defined in creature_template ^^

funjoker:
Well :D.. Shauren Check and merge :D

Eredar:
Any news?

Traesh:
Waiting for @Shauren to check and merge ^^

dikkaa:
@Shauren Any news for the merge ?

Exodius:
March 28 Hotfixes, Undocumented Changes in 7.2 Including Mobs Scaling with Item Level
http://bit.ly/2nzTEKP

Shauren:
I declare that part as out of scope for this PR

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -136,20 +136,23 @@ void WorldPackets::Spells::SandboxScalingData::GenerateDataForPlayerToNpc(Player
...
     TargetMaxScalingLevel   = creatureTemplate->maxlevel;
-    TargetScalingLevelDelta = (int8)creatureTemplate->levelScalingDelta.value();
+    TargetScalingLevelDelta = (int8)creatureTemplate->levelScalingDelta.get();
 }
 
 void WorldPackets::Spells::SandboxScalingData::GenerateDataForNpcToNpc(Creature* attacker, Creature* target)
 {
     CreatureTemplate const* creatureTemplate = target->GetCreatureTemplate();
 
+    if (!creatureTemplate->levelScalingDelta.is_initialized)
```
Shauren: is_initialized is a method

soulfrost:
SandboxScaling.txt game table should not be used here?

src/server/game/Server/Packets/SpellPackets.h
```diff
@@ -252,6 +246,15 @@ namespace WorldPackets
 
         class SandboxScalingData
         {
+            enum SandboxScalingDataType
+            {
+                TYPE_PLAYER_TO_PLAYER_HEALING    = 1, // NYI
+                TYPE_CREATURE_TO_PLAYER_DAMAGE   = 2,
+                TYPE_PLAYER_TO_CREATURE_DAMAGE   = 3,
+                TYPE_CREATURE_TO_CREATURE_DAMAGE = 4,
+                TYPE_PLAYER_TO_PLAYER_DAMAGE     = 32 // NYI
```
soulfrost:
```
TYPE_PLAYER_TO_PLAYER_DAMAGE     
```
dont seems rly exist on 720

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -92,6 +92,51 @@ WorldPacket const* WorldPackets::Spells::SendUnlearnSpells::Write()
...
+
+    Type                    = 3;
+    PlayerLevelDelta        = attacker->GetUInt32Value(PLAYER_FIELD_SCALING_PLAYER_LEVEL_DELTA);
+    PlayerItemLevel         = attacker->GetAverageItemLevel();
+    TargetLevel             = target->getLevel();
+    Expansion               = creatureTemplate->RequiredExpansion;
+    Class                   = creatureTemplate->unit_class;
+    TargetMinScalingLevel   = creatureTemplate->minlevel;
+    TargetMaxScalingLevel   = creatureTemplate->maxlevel;
+    TargetScalingLevelDelta = (int8)creatureTemplate->levelScalingDelta.value();
```
soulfrost: TargetScalingLevelDelta not exist on sniffs with that Type

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -92,6 +92,51 @@ WorldPacket const* WorldPackets::Spells::SendUnlearnSpells::Write()
...
+
+    Type                    = 2;
+    PlayerLevelDelta        = target->GetUInt32Value(PLAYER_FIELD_SCALING_PLAYER_LEVEL_DELTA);
+    PlayerItemLevel         = target->GetAverageItemLevel();
+    TargetLevel             = target->getLevel();
+    Expansion               = creatureTemplate->RequiredExpansion;
+    Class                   = creatureTemplate->unit_class;
+    TargetMinScalingLevel   = creatureTemplate->minlevel;
+    TargetMaxScalingLevel   = creatureTemplate->maxlevel;
+    TargetScalingLevelDelta = (int8)creatureTemplate->levelScalingDelta.value();
```
soulfrost: TargetScalingLevelDelta not exist on sniffs with that Type

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -92,6 +92,51 @@ WorldPacket const* WorldPackets::Spells::SendUnlearnSpells::Write()
...
+}
+
+void WorldPackets::Spells::SandboxScalingData::GenerateDataForPlayerToNpc(Player* attacker, Creature* target)
+{
+    CreatureTemplate const* creatureTemplate = target->GetCreatureTemplate();
+
+    Type                    = 3;
+    PlayerLevelDelta        = attacker->GetUInt32Value(PLAYER_FIELD_SCALING_PLAYER_LEVEL_DELTA);
```
soulfrost: PlayerLevelDelta not exist on sniffs with that Type

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -92,6 +92,51 @@ WorldPacket const* WorldPackets::Spells::SendUnlearnSpells::Write()
...
+
+    return true;
+}
+
+void WorldPackets::Spells::SandboxScalingData::GenerateDataForNpcToPlayer(Creature* attacker, Player* target)
+{
+    CreatureTemplate const* creatureTemplate = attacker->GetCreatureTemplate();
+
+    Type                    = 2;
+    PlayerLevelDelta        = target->GetUInt32Value(PLAYER_FIELD_SCALING_PLAYER_LEVEL_DELTA);
```
soulfrost: PlayerLevelDelta not exist on sniffs with that Type

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -92,6 +92,51 @@ WorldPacket const* WorldPackets::Spells::SendUnlearnSpells::Write()
...
+    Expansion               = creatureTemplate->RequiredExpansion;
+    Class                   = creatureTemplate->unit_class;
+    TargetMinScalingLevel   = creatureTemplate->minlevel;
+    TargetMaxScalingLevel   = creatureTemplate->maxlevel;
+    TargetScalingLevelDelta = (int8)creatureTemplate->levelScalingDelta.value();
+}
+
+void WorldPackets::Spells::SandboxScalingData::GenerateDataForPlayerToNpc(Player* attacker, Creature* target)
+{
+    CreatureTemplate const* creatureTemplate = target->GetCreatureTemplate();
```
soulfrost: check for null?

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -92,6 +92,51 @@ WorldPacket const* WorldPackets::Spells::SendUnlearnSpells::Write()
...
+        GenerateDataForNpcToPlayer(attacker->ToCreature(), target->ToPlayer());
+    else if (attacker->IsPlayer() && target->IsCreature())
+        GenerateDataForPlayerToNpc(attacker->ToPlayer(), target->ToCreature());
+
+    return true;
+}
+
+void WorldPackets::Spells::SandboxScalingData::GenerateDataForNpcToPlayer(Creature* attacker, Player* target)
+{
+    CreatureTemplate const* creatureTemplate = attacker->GetCreatureTemplate();
```
soulfrost: check for null?

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -137,6 +139,19 @@ void WorldPackets::Spells::SandboxScalingData::GenerateDataForPlayerToNpc(Player
...
+{
+    CreatureTemplate const* creatureTemplate = target->GetCreatureTemplate();
+
+    Type                    = TYPE_CREATURE_TO_CREATURE_DAMAGE;
+    TargetLevel             = target->getLevel();
+    Expansion               = creatureTemplate->RequiredExpansion;
+    Class                   = creatureTemplate->unit_class;
+    TargetMinScalingLevel   = creatureTemplate->minlevel;
+    TargetMaxScalingLevel   = creatureTemplate->maxlevel;
+    TargetScalingLevelDelta = (int8)creatureTemplate->levelScalingDelta.value();
```
soulfrost: TargetScalingLevelDelta not exist on sniffs with that Type

src/server/game/Server/Packets/SpellPackets.cpp
Traesh: A creature must have a template, not need to check for null

src/server/game/Server/Packets/SpellPackets.cpp
Traesh: A creature must have a template, not need to check for null

src/server/game/Server/Packets/SpellPackets.h
Traesh: https://pastebin.com/gv024wcP

Don't know in which conditions it is used

soulfrost: khm... are u had corrected retarded struct in WPP ?

soulfrost:
```
                DONT_SCALE                      = 0,
                PLAYER_TO_PLAYER                = 1,
                CREATURE_TO_PLAYER_DAMAGE       = 2,
                PLAYER_TO_CREATURE_DAMAGE       = 3,
                CREATURE_TO_CREATURE_DAMAGE     = 4,
```

looks like that
```c++

  if ( ( Type - 3 ) > 1 )
    return 1.0;
```
part from client handler

Traesh: Yup, bad WPP struct didn't pay enough attention, my bad

Traesh:
I will recheck later. And are you sure about that ? Check SandboxScaling.gt column names, it is clearly "Player-to-player Healing"

Shauren:
By healing blizz also means absorbs such as power word: shield

soulfrost:
ya, seems like that, and this type of scaling seems exactly clear
```
  if ( *LODWORD(v1) == 1 )
  {
    a1 = GetLevel(28, v5 - 1, 0, 1.0);
    v47 = GetLevel(28, LODWORD(v53) - 1, 0, 1.0);
    return a1 / v47;
  }
```

src/server/game/Server/Packets/SpellPackets.cpp
Traesh: not exist = it's not sent (condition in write() function) or value is default (= 0) ?

Shauren: Always sent. There is no condition in client packet reader

src/server/game/Server/Packets/SpellPackets.h
```diff
@@ -265,10 +265,8 @@ namespace WorldPackets
             uint8 TargetMaxScalingLevel = 1;
             int8 TargetScalingLevelDelta = 1;
```
soulfrost:
```
            uint8 Class = 1;
            uint8 TargetMinScalingLevel = 1;
            uint8 TargetMaxScalingLevel = 1;
            int8 TargetScalingLevelDelta = 1;
```
by default 0 at some types

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -92,6 +92,79 @@ WorldPacket const* WorldPackets::Spells::SendUnlearnSpells::Write()
...
+    if (   (!attacker->ToCreature() || !attacker->ToCreature()->HasScalableLevels())
+        && (!target->ToCreature()   || !target->ToCreature()->HasScalableLevels()))
+        return false;
+
+    if (attacker->IsCreature() && target->IsPlayer())
+        GenerateDataForUnits(attacker->ToCreature(), target->ToPlayer());
+    else if (attacker->IsPlayer() && target->IsCreature())
+        GenerateDataForUnits(attacker->ToPlayer(), target->ToCreature());
+    else
+        GenerateDataForUnits(attacker->ToCreature(), target->ToCreature());
```
soulfrost: think that here's should be smthng like this:
```c
     if (attacker->IsCreature() && target->IsPlayer())
         GenerateDataForUnits(attacker->ToCreature(), target->ToPlayer());
     else if (attacker->IsPlayer() && target->IsCreature())
         GenerateDataForUnits(attacker->ToPlayer(), target->ToCreature());
     else if (attacker->ToCreature() && target->ToCreature())
         GenerateDataForUnits(attacker->ToCreature(), target->ToCreature());
else if (attacker->IsPlayer(), target->IsPlayer())
         GenerateDataForUnits(attacker->ToPlayer(), target->ToPlayer());
else
         return false;

&&

template<>
 bool WorldPackets::Spells::SandboxScalingData::GenerateDataForUnits<Player, Player>(Player* attacker, Player* target)
 {
     Type                    = TYPE_PLAYER_TO_PLAYER;
     TargetLevel             = target->getLevel();
     return true;
 }
```

Shauren: No, because scaling player healing is not always enabled (and damage never is)

soulfrost: ya, but need other condition here

Traesh: Yes, it also need to be called from the corresponding SendXXXXXLog, that's why i commented "NYI" near the TYPE_PLAYER_TO_PLAYER. It will need some more research, i'll do it in a future PR, this one will only handle creatures scaling

Seyden: Player to player is only in dungeons enabled

src/server/game/Server/Packets/SpellPackets.cpp
```diff
@@ -93,7 +93,7 @@ WorldPacket const* WorldPackets::Spells::SendUnlearnSpells::Write()
 }
 
 template<class T, class U>
-bool WorldPackets::Spells::SandboxScalingData::GenerateDataForUnits(T* attacker, U* target)
+bool WorldPackets::Spells::SandboxScalingData::GenerateDataForUnits(T*, U*)
```
Seyden: why not T* /*attacker*/, U* /*target*/ to clarify, what is what

src/server/game/Server/Packets/SpellPackets.cpp
Traesh: What about grouped played in scaling level zones ? The healer can't correctly heal his teammate if he have a higher level ?

Seyden: I only knew it from dungeons.. but we should test it.. if its enabled in fight while staying in a level zone

Eredar:
Tested again. Player damage not scaled up. 200k hits at level 100 make 10% dmg to 300k hp target.

funjoker:
still would not use min max level column for scaling levels :/ because the min max is intended to make the npc spawn with rnd level (static real level).

Aokromes:
superseded by https://github.com/TrinityCore/TrinityCore/pull/20026

PR #19289, opened 2017-03-13
Core/AuctionHouse: use DisableMgr to disable specific items

ForesterDev:
…nHouse or AuctionHouseBot

- [x] 3.3.5


src/server/game/Conditions/DisableMgr.cpp
```diff
@@ -381,6 +392,7 @@ bool IsDisabledFor(DisableType type, uint32 entry, Unit const* unit, uint8 flags
         case DISABLE_TYPE_OUTDOORPVP:
         case DISABLE_TYPE_ACHIEVEMENT_CRITERIA:
         case DISABLE_TYPE_MMAP:
+        case DISABLE_TYPE_AUCTIONHOUSE_ITEM:
```
Aokromes: while you are on it, how about also add DISABLE_TYPE_ITEM ? :)

ForesterDev: Where else can DISABLE_TYPE_ITEM be used?

Aokromes: vendors, drop, quest rewards. hmmm maybe better
DISABLE_TYPE_DROP_ITEM 
DISABLE_TYPE_VENDOR_ITEM 
DISABLE_TYPE_QUEST_ITEM

ForesterDev: Then, I think, will be better to implement DISABLE_TYPE_ITEM with different flags to avoid multiply Types. Give me a time and i'll implement this

src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
```diff
@@ -146,7 +146,6 @@ void AuctionBotConfig::GetConfigFromFile()
     SetConfigMax(CONFIG_AHBOT_NEUTRAL_ITEM_AMOUNT_RATIO, "AuctionHouseBot.Neutral.Items.Amount.Ratio", 100, 10000);
 
     SetAHBotIncludes(sConfigMgr->GetStringDefault("AuctionHouseBot.forceIncludeItems", ""));
-    SetAHBotExcludes(sConfigMgr->GetStringDefault("AuctionHouseBot.forceExcludeItems", ""));
```
ariel-: I would keep this for compatibility

DDuarte: I'm kinda split, backwards compatibility is nice but two ways of doing the same thing is not.

ariel-: The problem here is that if anyone edited his/her config, it will be silently discarded

Aokromes: maybe we can add a warn of unused setting (but instead to startup to the end of it)

MrSmite: Is it possible to have the DisablesMgr read that line, parse it and enter it into the `disables` table and then remove it from the CONF?

You could then print a message in the console: *AuctionHouseBot.forceExcludeItems automatically ported to `disables` table and removed from config*

Shauren: no, std::string does not have operator bool

ariel-: And you lose info if world db gets reset (ie new TDB)

MrSmite: I only used std::string as an example but what about

```C++
std::string disabledItems = sConfigMgr->GetStringDefault("AuctionHouseBot.forceExcludeItems", "")
if (boost::lexical_cast<bool>(disabledItems))
{
    // do stuff
}

As far as losing data on a db update, you would lose that anyway because those entries are entirely user defined so the user would need to maintain a separate SQL file for their AHBot stuff, along with the other AHBot items in the CONF file.

Comment (Aokromes) [inline src/server/game/AuctionHouseBot/AuctionHouseBot.cpp]: wiki suggest to save your changes.

Comment (ariel-) [inline src/server/game/AuctionHouseBot/AuctionHouseBot.cpp]: > Anyway, as far as losing data on a db update, you would lose that anyway because those entries are entirely user defined so the user would need to maintain a separate SQL file for their AHBot stuff, along with the other AHBot options in the CONF file.

Which wasn't lost before with tdb updates since those entries are in the conf file.

What about adding a deprecation notice to config and core?

Comment (MrSmite) [inline src/server/game/AuctionHouseBot/AuctionHouseBot.cpp]: > Which wasn't lost before with tdb updates since those entries are in the conf file.

Right which is why I was suggesting the `disables_custom` table. Users could enter their IDs here just as they used to do in the CONF and not risk losing all their data and without the necessity of maintaining a SQL file somewhere.

It's pretty trivial for the prepared statement to be updated to read from two tables.

**Comment by Aokromes** [comment]
Conflicting files
src/server/game/AuctionHouse/AuctionHouseMgr.h
src/server/game/AuctionHouseBot/AuctionHouseBot.h

File: src/server/game/AuctionHouse/AuctionHouseMgr.h [hunk 3]
```diff
@@ -22,6 +22,7 @@
 #include "Common.h"
 #include "DatabaseEnv.h"
 #include "DBCStructure.h"
+#include "DisableMgr.h"
```
jackpoz: could you move this include in the .cpp ?

src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
jackpoz: I vote to keep the config

Aokromes:
Conflicting files
src/server/game/AuctionHouse/AuctionHouseMgr.h
src/server/game/AuctionHouseBot/AuctionHouseBot.h

jackpoz:
Config deprecation steps:
1.  Keep both options but log a warning when values are found in the config with instructions about how to migrate to the new table system
2. After some months, remove support for the config and log an error
3. After some months, remove the error

Please implement the step 1 of the list above, then it can be merged

src/server/game/AuctionHouseBot/AuctionHouseBot.cpp
```diff
@@ -147,7 +147,12 @@ void AuctionBotConfig::GetConfigFromFile()
...
+    std::string ahBotExcludes = sConfigMgr->GetStringDefault("AuctionHouseBot.forceExcludeItems", "");
+    if (!ahBotExcludes.empty())
+    {
+        TC_LOG_WARN("server.loading", "AuctionHouseBot.forceExcludeItems will be deleted soon. Please, use DisableMgr instead.");
```
Shauren: Minor note about warning log - end user doesn't neccessarily know what `DisableMgr` is but knows `AuctionHouseBot.forceExcludeItems` and knows `disables` db table (mention that instead)

src/server/game/Handlers/AuctionHouseHandler.cpp
```diff
@@ -208,6 +209,12 @@ void WorldSession::HandleAuctionSellItem(WorldPacket& recvData)
         if (itemEntry == 0)
             itemEntry = item->GetTemplate()->ItemId;
 
+        if (DisableMgr::IsDisabledFor(DISABLE_TYPE_ITEM, itemEntry, nullptr, ITEM_DISABLE_AUCTIONHOUSE))
```
Shauren: With this you completely change the old meaning of `AuctionHouseBot.forceExcludeItems` configuration, this check will not prevent regular player from ever listing that specific item on AH manually, you can't have this tied to AH bot configuration

Shauren (changes requested):
I don't know what to think of this PR, why is the current config bad?

jackpoz:
It kinda makes sense to put all disabled thingies in a single place, it probably breaks down though to personal preference (all AHBot settings together or all disabled items together ?).
Depending on how many items you want to disable, the entry in .conf can get quite long and poorly maintenable, while in SQL you can easily query the item id.

ForesterDev:
Summon @jackpoz

jackpoz:
looks good, did you test ingame latest changes too ?

jackpoz:
Thanks for the PR :)

jackpoz:
This is actually completely broken because when disables table is loaded, before items_template, making the check about "does the item exist ?" always return false. See https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/World/World.cpp#L1698

jackpoz:
This PR has been reverted

PR #19306, opened 2017-03-16
Core/SocialMgr: fix case when contacts appears in friend list many times

### LINKED ISSUE #19266: Core: Duplicated entry on friend list

**Description:** On the friend list some contacts appears many times.

**Current behaviour:** http://i.imgur.com/1TJ2zQ8.jpg
http://i.imgur.com/oV7igXl.jpg

**Expected behaviour:** The friend must appear only once

Unknown
Nothing weird or unusual on the DB

**Branch(es):** 3.3.5

**TC rev. hash/commit:** rev. 8089f1816c9b

ForesterDev:

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #19266  (insert issue tracker number)

**Tests performed:** (Does it build, tested in-game, etc.) tested

** How to reproduce bug: add contacts to ignore list.

### DIFF
--- a/src/server/game/Entities/Player/SocialMgr.cpp
+++ b/src/server/game/Entities/Player/SocialMgr.cpp
@@ -125,12 +125,13 @@ void PlayerSocial::SendSocialList(Player* player, uint32 flags)
 {
     ASSERT(player);
 
-    uint32 size = GetNumberOfSocialsWithFlag(SocialFlag(flags));
-
     uint32 count = 0;
-    WorldPacket data(SMSG_CONTACT_LIST, (4 + 4 + size * 25)); // just can guess size
+
+    WorldPacket data(SMSG_CONTACT_LIST, (4 + 4 + _playerSocialMap.size() * 25)); // just can guess size
     data << uint32(flags);                                    // 0x1 = Friendlist update. 0x2 = Ignorelist update. 0x4 = Mutelist update.
-    data << uint32(size);                                     // friends count
+
+    size_t wpos = data.wpos();
+    data << uint32(0);                                        // placeholder for contacts count
 
     for (PlayerSocialMap::value_type& v : _playerSocialMap)
     {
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Player/SocialMgr.cpp
```diff
@@ -125,7 +125,7 @@ void PlayerSocial::SendSocialList(Player* player, uint32 flags)
 {
     ASSERT(player);
 
-    uint32 size = _playerSocialMap.size();
+    uint32 size = GetNumberOfSocialsWithFlag(SocialFlag(flags));
```
jackpoz: is this the same as the count variable below ?

ForesterDev: Yes, but we need this variable to initialize SMSG_CONTACT_LIST packet size.

ForesterDev:
@jackpoz i think this solution will be vore cleaner

jackpoz:
It looks better imo. Could you repeat the tests just to be sure ? Then I think it can be merged.

ForesterDev:
@jackpoz works fine. 

I also have a proposal: rename FriendInfo struct to ContactInfo, because not only friends, but also ignored players use this strcture

jackpoz:
Better keep refactoring changes for another PR and focus only on the bug in this one

src/server/game/Entities/Player/SocialMgr.cpp
```diff
@@ -125,40 +125,43 @@ void PlayerSocial::SendSocialList(Player* player, uint32 flags)
...
     {
-        if (!(v.second.Flags & flags))
+        FriendInfo friendInfo = v.second;
+        if (!(friendInfo.Flags & flags))
             continue;
 
-        ++count;
-        sSocialMgr->GetFriendInfo(player, v.first, v.second);
+        ObjectGuid friendGUID = v.first;
+        sSocialMgr->GetFriendInfo(player, friendGUID, friendInfo);
```
ariel-: You now broke functionality that kept map info updated, because you get the friendInfo in a copy.

PR #19314, opened 2017-03-17
Core/Spell Implement talent warrior level 110 anger management.

HannibalRoG:

**Changes proposed:**

-  Implement talent level 110 Anger Management.

- [x] master

Compiled and tested

### DIFF
--- a/sql/updates/world/master/RoG_2017_03_16_00_world.sql
+++ b/sql/updates/world/master/RoG_2017_03_16_00_world.sql
@@ -1,0 +1,5 @@
++-- Class: Warrior
++-- Spell: Anger Management
++
++DELETE FROM `spell_script_names` WHERE `ScriptName` = 'spell_warr_anger_management';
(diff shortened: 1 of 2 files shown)


Shauren:
This is not good at all, you need to track rage spent for all spells so casting a spell with 5 rage cost followed by another 5 rage would reduce the cooldown by 1 second
`SpellHistory::ModifyCooldown` does not need HasCooldown check before it
`CalcPowerCost` consumes spellmods with charges

DoctorKraft:
A playerscript with function OnModifyPower is a better method ?

HannibalRoG:
Let's deal with playerscript, we had doubts about that but his comments made it clear that we can use it, thank you.

DoctorKraft:
i think, it's strange this proc on spell... but with your spellscript system i don't think we can do otherwise

wonder2011:
I search for the playerscript event OnModifyPower or somethig like that but i don't find anything. I miss something here?

wonder2011:
@Shauren if @HannibalRoG hold the spent rage and in case of 5 + 5 or similars, reduce the CD's, the PR then would be fine?

PR #19331, opened 2017-03-19
Implement quest_request_items_locale

AlcDenat:
**Changes proposed:** Add a new table quest_request_items_locale and code

**Target branch(es):** master

**Issues addressed:** Missing locale table

**Tests performed:** no errors, code works ingame

**Known issues and TODO list:** none

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -4824,6 +4824,36 @@ void ObjectMgr::LoadQuestOfferRewardLocale()
     TC_LOG_INFO("server.loading", ">> Loaded " SZFMTD " Quest Offer Reward locale strings in %u ms", _questOfferRewardLocaleStore.size(), GetMSTimeDiffToNow(oldMSTime));
 }
 
+void ObjectMgr::LoadQuestRequestItemsLocale()
+{
+    uint32 oldMSTime = getMSTime();
+
+    _questRequestItemsLocaleStore.clear(); // need for reload case
+                                          //         0     1          2
+    QueryResult result = WorldDatabase.Query("SELECT Id, locale, CompletionText FROM quest_request_items_locale");
+    if (!result)
+        return;
+
+    do
+    {
+        Field* fields = result->Fetch();
+
+        uint32 id = fields[0].GetUInt32();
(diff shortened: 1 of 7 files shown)


src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -4824,6 +4824,36 @@ void ObjectMgr::LoadQuestOfferRewardLocale()
...
+        std::string localeName = fields[1].GetString();
+
+        std::string CompletionText = fields[2].GetString();
+
+        QuestRequestItemsLocale& data = _questRequestItemsLocaleStore[id];
+        LocaleConstant locale = GetLocaleByName(localeName);
+        if (locale == LOCALE_enUS)
+            continue;
+
+        AddLocaleString(CompletionText, locale, data.RequestItemsText);
```
Shauren: copypaste is strong here, `CompletionText`

Shauren (changes requested):

src/server/game/Globals/ObjectMgr.cpp
AlcDenat: While 98% of this stuff is copy & paste, CompletionText is the field name in the original table and thus in the _locale table I created. I'd say it's ok to just let it be - It's never reused anyway.

joschiwald: `CompletionText` should be fine, rename `RequestItemsText` -> `CompletionText`

Shauren: just rename it to RequestItemsText to remove inconsistencies

Shauren: The reason I recommend keeping the `RequestItems` name is because that is the field name used in packet

AlcDenat: Changed it to RequestItems, should be fine now.

joschiwald:
@Shauren i don't get your point, in packet it is named `CompletionText`

or something changed since 6.x?
```c#
public class ClientQuestGiverRequestItems
{
    public string QuestTitle                { get; set; }
    public QuestGiverRequestItems QuestData { get; set; }
    public string CompletionText            { get; set; }
}
```

Shauren:
fine then

PR #19365, opened 2017-03-25
Core/ScriptReloadMgr: Fixed scripts hot swapping on OS X

svranesevic:
**Changes proposed:**
-  Added appropriate dynamic linking include, and shared library extension for OS X
-  Canonicalized world server executable path in ScriptReloadMgr::GetDirectoryOfExecutable()

- [x] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**
- [x] Builds
- [x] Tested in-game

### DIFF
--- a/src/server/game/Scripting/ScriptReloadMgr.cpp
+++ b/src/server/game/Scripting/ScriptReloadMgr.cpp
@@ -67,6 +67,9 @@ namespace fs = boost::filesystem;
 #ifdef _WIN32
     #include <windows.h>
     #define HOTSWAP_PLATFORM_REQUIRES_CACHING
+#elif __APPLE__
+    #include <dlfcn.h>
+    #define HOTSWAP_PLATFORM_REQUIRES_CACHING
 #else // Posix
     #include <dlfcn.h>
     // #define HOTSWAP_PLATFORM_REQUIRES_CACHING
@@ -87,11 +90,13 @@ static char const* GetSharedLibraryPrefix()
 #endif
 }
 
-// Returns "dll" on Windows and "so" on posix.
+// Returns "dll" on Windows, "dylib" on OS X, and "so" on posix.
 static char const* GetSharedLibraryExtension()
 {
 #ifdef _WIN32
     return "dll";
+#elif __APPLE__
+    return "dylib";
 #else // Posix
     return "so";
 #endif
@@ -112,7 +117,7 @@ static fs::path GetDirectoryOfExecutable()
     if (path.is_absolute())
         return path.parent_path();
     else
-        return fs::absolute(path).parent_path();
+        return fs::canonical(fs::absolute(path)).parent_path();
 }
 
 class SharedLibraryUnloader
--- a/cmake/macros/ConfigureScripts.cmake
+++ b/cmake/macros/ConfigureScripts.cmake
@@ -85,10 +85,12 @@ function(IsDynamicLinkingRequired variable)
   set(${variable} ${IS_REQUIRED} PARENT_SCOPE)
 endfunction()
 
-# Stores the native variable name 
+# Stores the native variable name
 function(GetNativeSharedLibraryName module variable)
   if(WIN32)
     set(${variable} "${module}.dll" PARENT_SCOPE)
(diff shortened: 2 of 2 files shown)


Naios:
Nice that finally someone fixed hotswap for OSX :-)

src/server/game/Scripting/ScriptReloadMgr.cpp
```diff
@@ -67,6 +67,9 @@ namespace fs = boost::filesystem;
 #ifdef _WIN32
     #include <windows.h>
     #define HOTSWAP_PLATFORM_REQUIRES_CACHING
+#elif __APPLE__
+    #include <dlfcn.h>
+    #define HOTSWAP_PLATFORM_REQUIRES_CACHING
```
Naios: Is caching of binaries actually required on OSX?
The BSD kernel should behave similar to Linux not locking the loaded shared libraries on usage.

src/server/game/Scripting/ScriptReloadMgr.cpp
```diff
@@ -112,7 +117,7 @@ static fs::path GetDirectoryOfExecutable()
     if (path.is_absolute())
         return path.parent_path();
     else
-        return fs::absolute(path).parent_path();
+        return fs::canonical(fs::absolute(path)).parent_path();
```
Naios: 👍

src/server/game/Scripting/ScriptReloadMgr.cpp
```diff
@@ -87,11 +90,13 @@ static char const* GetSharedLibraryPrefix()
...
 }
 
-// Returns "dll" on Windows and "so" on posix.
+// Returns "dll" on Windows, "dylib" on OS X, and "so" on posix.
 static char const* GetSharedLibraryExtension()
 {
 #ifdef _WIN32
     return "dll";
+#elif __APPLE__
+    return "dylib";
```
Naios: You are also required to add an entry in the CMake function [here](https://github.com/TrinityCore/TrinityCore/blob/86b98686a95e23247ecb774fb23ecd5b8d94b97b/cmake/macros/ConfigureScripts.cmake#L89) so the shared library gets removed properly on uninstall.

Naios (changes requested):
Well done :-)
The commit is only missing the support for uninstalling `.dylib` as mentioned in the review.

src/server/game/Scripting/ScriptReloadMgr.cpp
svranesevic: Roger.

src/server/game/Scripting/ScriptReloadMgr.cpp
svranesevic: It is, I will check again tho.

svranesevic: It is needed indeed,
Without caching compilation and reload procedures are triggered, however after reload old scripts are still used by the core.

For the sake of my curiosity, is there anything else that could be causing this behaviour ?

Naios (approved):

src/server/game/Scripting/ScriptReloadMgr.cpp
Naios: How did you see that the scripts are still usable by the core?
Maybe you could explain your impressions in detail.

Caching is required only if you can't delete shared libraries out of the `scripts` directory through your file explorer (when caching is disabled).

svranesevic: I was playing around with '.gm' command from GM script, and making changes in notification system (eg. GM mode on/off) so I could have visual feedback on functionality of loaded script.
And without caching script used by the core is not reloaded at all, even though whole compile/reload procedure successfully completes after script file change.

Also , OS X definitely does not lock loaded shared libraries.

Naios: Ah ok, don't know why that is the case, but then just let the system cache the binaries for the OSX platform.

Shauren (approved):

Shauren:
Looks good enough to me

Issue #14790, opened 2015-05-28
[3.3.5][6.x][Exploit] learn extra spells


Rochet2:
The trainer system allows training of few extra spells with hacks.
For example the npc_trainer table holds data for Death Knight Strangulate spell: http://wotlk.openwow.com/spell=47476#see-also-ability
And the ranks for it like and the others http://wotlk.openwow.com/spell=49913
However only the first spell linked is normally learnable. The other strangulate versions do not show on the training list.
It seems that the core sends all the strangulate spells to the client, but the client only shows the correct spell.
The learning does not do any extra checks either so the spells that do not show are learnable.
- 3.3.5, all?
- https://github.com/TrinityCore/TrinityCore/commit/1161295f4f933661ea9738e0c57049a1e38ec8f2
- TDB 335.58
- a lot of custom stuff, but probably none for this part anyways
##

dr-j:
well class trainers have conditions on the gossip option to access the trainer so only the class for which trainer teaches spells can access the trainer for most this works but believe there may be some bugs have still heard about some trainers teaching spells which should not.

Ps about DK spells it should not be possible for a non-dk to reach archeus to access dk trainers there as the flight npcs to fly up there cant be accessed by non dk characters.

Rochet2:
@dr-j Is that somehow related to this issue? :/
This is about DK learning extra DK spells.
What you are talking about is related to this: https://github.com/TrinityCore/TrinityCore/issues/14586

Rochet2:
It seems that this is able to filter all the bad spells:
Only used on class spells.
BadSpells:(1785,1786,1787,6783,8649,8650,9913,11197,11198,26866,48669,49913,49914,49915,49916,51426,51427,51428,51429,53720)
```c++
    static bool IsValidHelper(SpellInfo const* info)
    {
        SkillLineAbilityMapBounds bounds = sSpellMgr->GetSkillLineAbilityMapBounds(info->Id);
        SkillLineAbilityMap::const_iterator _spell_idx = bounds.first;
        if (_spell_idx == bounds.second)
            return false; // This is the only return false that was reached
/*
        for (; _spell_idx != bounds.second; ++_spell_idx)
        {
            SkillLineEntry const* pSkill = sSkillLineStore.LookupEntry(_spell_idx->second->skillId);
            if (!pSkill)
                return false;
            switch (pSkill->categoryId)
            {
                case SKILL_CATEGORY_WEAPON:
[...]

PR #19394, opened 2017-03-31
Fix exploit for learning spells that the player should not get

Rochet2:

**Changes proposed:**

- First off I would like to note that I don't know what GetSkillLineAbilityMapBounds contains exactly and if there is a better / more correct way to do this. The base issue is that there are class spells which should not be learned by any class and the suggested code rules them out. The suggested code cannot be used on all spells most likely, only the ones a class trainer can teach. 
- Add a check to race&class check. The check will now return false if the spell is a class specific spell (spell family matches class) and it is not in skill line ability store.
- Fix an error message that claimed a spell was nonexistent even if the spell exists, but it is a talent

The spells this new check filters out do not appear on client training list even if sent to the client. They show up in general tab if learned. They should not be learnable. They can be learned through .learn all my class command before this PR and through hacks by altering the spell being learned at a trainer.

List of bad spells from linked issue:
1785, 1786, 1787, 6783, 8649, 8650, 9913, 11197, 11198, 26866, 48669, 49913, 49914, 49915, 49916, 51426, 51427, 51428, 51429, 53720

- [x] 3.3.5
- [x] master - possibly, same spells are not in DB

Closes https://github.com/TrinityCore/TrinityCore/issues/14790

- Tested only on 3.3.5
- Compiles
- Set up debugging to log all spells that the modification returns false for, logged in with each class, leveled to 80, performed .learn all my class, added some money to the player, went to a class trainer and learned all spells he had to offer. In the end only the same list of spells was being blocked as mentioned in the issue linked above.
- I also tried to check if some (but not all) profession trainers had spells blocked by this unintentionally, but I did not notice any on First Aid, Leatherworking, Cooking

- [ ] This modification does not alert that there are bad spells in trainer table, just silently ignores them and does not let player learn them due to red status.
- [ ] This modification does not remove the bad spells from DB, I'm unsure if the DB data in npc_trainer is blizzlike or not. It could be, even if these spells are not learnable (they are ignored by client).

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -23361,11 +23361,9 @@ bool Player::IsSpellFitByClassAndRace(uint32 spell_id) const
     {
         if (SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(spell_id))
         {
-            if (ChrClassesEntry const* classEntry = sChrClassesStore.LookupEntry(getClass()))
-            {
-                if (spellInfo->SpellFamilyName == classEntry->spellfamily)
-                    return false;
-            }
+            ChrClassesEntry const* classEntry = sChrClassesStore.LookupEntry(getClass());
+            if (spellInfo->SpellFamilyName == classEntry->spellfamily)
+                return false;
         }
         return true;
     }
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -23358,7 +23358,17 @@ bool Player::IsSpellFitByClassAndRace(uint32 spell_id) const
 
     SkillLineAbilityMapBounds bounds = sSpellMgr->GetSkillLineAbilityMapBounds(spell_id);
     if (bounds.first == bounds.second)
+    {
+        if (const SpellInfo* spellInfo = sSpellMgr->GetSpellInfo(spell_id))
+        {
+            if (ChrClassesEntry const* classEntry = sChrClassesStore.LookupEntry(getClass()))
+            {
+                if (spellInfo->SpellFamilyName == classEntry->spellfamily)
+                    return false;
+            }
+        }
         return true;
+    }
 
     for (SkillLineAbilityMap::const_iterator _spell_idx = bounds.first; _spell_idx != bounds.second; ++_spell_idx)
     {
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -23359,7 +23359,7 @@ bool Player::IsSpellFitByClassAndRace(uint32 spell_id) const
     SkillLineAbilityMapBounds bounds = sSpellMgr->GetSkillLineAbilityMapBounds(spell_id);
     if (bounds.first == bounds.second)
     {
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -23358,7 +23358,17 @@ bool Player::IsSpellFitByClassAndRace(uint32 spell_id) const
 
     SkillLineAbilityMapBounds bounds = sSpellMgr->GetSkillLineAbilityMapBounds(spell_id);
     if (bounds.first == bounds.second)
+    {
+        if (const SpellInfo* spellInfo = sSpellMgr->GetSpellInfo(spell_id))
```
ariel-: SpellInfo const*

src/server/game/Entities/Player/Player.cpp
```diff
@@ -23358,7 +23358,17 @@ bool Player::IsSpellFitByClassAndRace(uint32 spell_id) const
...
+        {
+            if (ChrClassesEntry const* classEntry = sChrClassesStore.LookupEntry(getClass()))
```
ariel-: getClassMask is called earlier, and class gets checked on player Load. Is it necessary to check ChrClassesEntry against null?

Shauren:
GetSkillLineAbilityMapBounds -> parsed SkillLineAbility.dbc

src/server/game/Entities/Player/Player.cpp
```diff
@@ -23358,7 +23358,17 @@ bool Player::IsSpellFitByClassAndRace(uint32 spell_id) const
 
     SkillLineAbilityMapBounds bounds = sSpellMgr->GetSkillLineAbilityMapBounds(spell_id);
     if (bounds.first == bounds.second)
+    {
+        if (SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(spell_id))
+        {
+            if (ChrClassesEntry const* classEntry = sChrClassesStore.LookupEntry(getClass()))
```
ariel-: Can never be null

ariel-:
> This modification does not remove the bad spells from DB, I'm unsure if the DB data in npc_trainer is blizzlike or not. It could be, even if these spells are not learnable (they are ignored by client).

It's not, spells should be removed from DB entirely, just checked sniffs.

For paladin for example:
```
ServerToClient: SMSG_TRAINER_LIST (0x01B1) Length: 6784 ConnIdx: 0 Time: 05/15/2010 01:14:13.000 Number: 286834
GUID: Full: 0xF1300003A007B3F4 Type: Creature Entry: 928 Low: 504820
Type: Class (0)
Count: 177
```

In TC we send 178 spells, one more. That 'extra' spell is 53720, which shouldn't be there in the first place.

Rochet2:
Then a query should be made to delete them

```sql
DELETE FROM `npc_vendor` WHERE SpellID IN (1785, 1786, 1787, 6783, 8649, 8650, 9913, 11197, 11198, 26866, 48669, 49913, 49914, 49915, 49916, 51426, 51427, 51428, 51429, 53720);
```

ariel-:
Pushing the sql part, I vote for logging on startup

Rochet2:
Im closing this because the logging is difficult to do (needs to be done only on class spells) and it may not be correct as it is based on a guess on how to filter the spells.
The issue were also already solved in the commits made so far.

PR #19418, opened 2017-04-07
Core/Creature: Fixed ForcedDespawn

ghost:
Existing code prevents forceRespawnTimer from being used to determine respawn time if creature is already dead before DespawnOrUnsummon is used.


Shauren:
The `IsAlive()` check is there for a reason - you can't just expect everything will work if you call setDeathState(JUST_DIED) on a dead creature.
Just removing it is a wrong approach.

malcrom:
I can't see what reason.

malcrom:
I can change to:

    else
        if (IsAlive())
            setDeathState(JUST_DIED);

But the first setDeathState(JUST_DIED) must be there to update the respawn timer.

malcrom:
This is working. I need this pushed.

Issue #17265, opened 2016-06-02
[3.3.5][6.x] Sindragosa - Icy Grip


ghost:
**Description**: Icy grip should move player to center of Sindragosa

**Current behaviour**: Icy grip move player to border of Sindragosa model

**Branch(es)**: 3.3.5 

**TC hash/commit**:  325143960074a562b1d419bfe23cf9cbf1b75e11

**TDB version**:  61

**Operating system**:  Debian

Rushor:
possible fix in https://github.com/TrinityCore/TrinityCore/issues/1087

moreover there must be some exceptions for the jump.
1.) it should not move the tank
2.) it should not move player which are in the ice block in phase 2 switch
3.) it should not move people which are not in los (f.ex. behind the iceblocks)

Runico:
@Rushor, that fix doesnt work anymore cause of new changes

P-Kito:
Here comes the hackfix!
```        
case 70122: // Sindragosa - Icygrip jump
                spellInfo->Effects[EFFECT_0].TargetA = SpellImplicitTargetInfo(TARGET_DEST_TARGET_ANY);
                spellInfo->Effects[EFFECT_0].Effect = SPELL_EFFECT_JUMP_DEST;
                break;
```

Keader:
Sometimes icy grip is not casted too ._.

Keader:
@chaodhib any idea here? :)

chaodhib:
The current implementation of SPELL_EFFECT_JUMP is to move the caster 'x' yards away from the target, where x = target's combat reach + caster's combat reach + 0.5y.

This can easily be changed to move the caster exactly on top of the target (like this spell would require). 

However, this will break spells that uses SPELL_EFFECT_JUMP to move the caster but not exactly at the same place as where the target currently is standing at. The question is: does it happen? Is there any spells that shouldn't move the caster to where the target is standing?

If so, the answer may be in the additional data of the spell effect (https://trinitycore.atlassian.net/wiki/display/tc/Spell+Effects+Reference#SpellEffectsReference-SPELL_EFFECT_JUMP=41) which may be unimplemented or misunderstood.

chaodhib:
The original issue has been fixed.

PR #19438, opened 2017-04-10
Core/Spells: change to the way the spell effect SPELL_EFFECT_JUMP work

chaodhib:
**Changes proposed:**
- The spell effect 41 (SPELL_EFFECT_JUMP) moves the caster at the exact target's position instead of moving it to the closest "contact point" with the target.

**Target branch(es):** 3.3.5

**Issues addressed:** Closes https://github.com/TrinityCore/TrinityCore/issues/17265#issuecomment-292783050

**Tests performed:** Tested IG. Appears to work well.

**Left to do:**
What is left is verify through sniffs analysis that this change is indeed retail like. I observed 3 spells with that spell effects (56729, 70122 and 52631). The 3 of them seems to follow the new logic.

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -1024,12 +1024,9 @@ void Spell::EffectJump(SpellEffIndex effIndex)
     if (!unitTarget)
         return;
 
-    float x, y, z;
-    unitTarget->GetContactPoint(m_caster, x, y, z, CONTACT_DISTANCE);
-
     float speedXY, speedZ;
-    CalculateJumpSpeeds(effIndex, m_caster->GetExactDist2d(x, y), speedXY, speedZ);
-    m_caster->GetMotionMaster()->MoveJump(x, y, z, 0.0f, speedXY, speedZ, EVENT_JUMP, false);
+    CalculateJumpSpeeds(effIndex, m_caster->GetExactDist2d(unitTarget), speedXY, speedZ);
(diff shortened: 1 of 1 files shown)


ccrs:
IMO the point here is to research why is it necessary to get the contact point instead of the exact position.

chaodhib:
What I tried to say is that for all the spells using SPELL_EFFECT_JUMP that I observed in sniffs, it is NOT the contact point that is used, but the position of the target itself. I know that my data sample is small but unless we can find an exception to this rule, then the current implementation is simply wrong.

ccrs:
ummm I'm thinking about death grip from DKs, that might be the only reason behind the contact thing

chaodhib:
Death Grip (49575) uses the spell effect 42 (SPELL_EFFECT_JUMP_DEST). So it's out of scope.

chaodhib:
I will merge this PR in one week if no opposition. Meanwhile, I will analyze more spells with the same spell effect.

Treeston:
What are the spells that use this effect?

chaodhib:
From SpellWork:
image

ariel-:
Update your SpellWork and connect it to DB to see the spells from spell_dbc table too! :P

updated spellwork

chaodhib:
@ariel- Neat. Thanks!

Keader:
Looks good to me

chaodhib:
Now that the sniff repo is up, i also got the chance to analyse the sniffs of an another spell. Exactly the same behavior as the first 3. I'm pretty confident about the changes so I'm merging now.

PR #19497, opened 2017-04-23
Scripts/Commands: Modify some Graveyard references to WorldSafeLoc

Krudor:

**Changes proposed:**

-  Change ".go graveyard" to ".go worldsafelocation"
-  Modify some related trinity strings

**Target branch(es):** master

**Tests performed:** Builds, appears to work as intended in game

Is probably valid for 3.3.5 if someone wants to port it.


runningnak3d:
What is the point of this?

Aokromes:
that's why i was thinking xd

Krudor:
World Safe Location's are not only graveyard teleport locations, they are general purpose locations.
"go graveyard" is a not a good command name for what it actually does.

runningnak3d:
I understand that, but this command has been around for so long that changing it for those *very few* fringe cases when you aren't teleporting to a graveyard doesn't make sense. 

Just throwing my 2 cents worth out there :)

Shauren:
I agree that its not worth renaming a command that has been around for years - what you could do is just provide a name alias (add to command table with same handler/permission but different name)

Krudor:
Yeah that sounds like a better idea

PR #19504, opened 2017-04-23
[3.3.5] Core/Spells: don't skip resisted damage calculations for binary spells cast by creatures

Xyventh:
**Changes proposed:**

Binary spells cast by NPCs should allow for the damage to be partially resisted via normal resistance calculations, as opposed to those cast by players.

Right now, it can only be fully resisted or not resisted at all.

One example for this behavior is [here](https://youtu.be/V-qgBmWcEjg?t=46): the Druid is partially resisting Frost Nova's damage.

Another example is [here](https://youtu.be/Qtk3JXUvjWQ?t=122): the Warrior resists most of Prophet Skeram's Earth Shock (a binary spell) damage due to high Nature Resistance.

**Target branch(es):** 3.3.5

**Tests performed:** builds and works just fine.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -1678,8 +1678,8 @@ uint32 Unit::CalcSpellResistedDamage(Unit* victim, uint32 damage, SpellSchoolMas
         if (spellInfo->HasAttribute(SPELL_ATTR4_IGNORE_RESISTANCES))
             return 0;
 
-        // Binary spells can't have damage part resisted
-        if (spellInfo->HasAttribute(SPELL_ATTR0_CU_BINARY_SPELL))
+        // Damage dealt by binary spells cast by players cannot be partially resisted
+        if (GetTypeId() == TYPEID_PLAYER && spellInfo->HasAttribute(SPELL_ATTR0_CU_BINARY_SPELL))
             return 0;
     }
 


ariel-:
What? I thought the whole point of binary spells was to either resist or hit, not to have partial resistances

Xyventh:
@ariel-: I think it's only a PvP mechanic to prevent people from stacking resistances and making casters useless.

In PvE it's not that game-breaking. Stacking resistance is sometimes supported by Blizzard's design for some bosses, so it would make sense to reduce a spell's damage, be it binary or not :)

tkrokli:
Sounds like you have looked into an area of the source which could (in theory) be related to the item (bag) [[Foror's Crate of Endless Resist Gear Storage]](http://wotlk.openwow.com/item=23162) :-D

Xyventh:
@tkrokli: that's not far from the truth :D

(was mad I couldn't solo Prophet Skeram when people did that on retail... took a shovel and started digging for info :P)

ariel-:
Let's ping @xinef1

xinef1:
I can't say if such a generic rule exist (no binary resistance for spells cast by npc) but I can't guarantee that binary setting rules are 100% blizzlike - they were created according to various descriptions of this mechanic and as we all know, they may skip some things. Maybe spells with family generic or damage class none should be always non-binary? Or if those two spells hava a special attribute? Can't check that now by myself so if @ariel- could I would be greatful.

Xyventh:
@xinef1: the only common attribute that's not shared with other spells cast by players is SPELL_ATTR0_LEVEL_DAMAGE_CALCULATION.

Other than that, they're both classified as SPELLFAMILY_GENERIC.

Here's the two spells for reference: [Earth Shock](https://pastebin.com/kHsZw4jk) and [Frost Nova](https://pastebin.com/p6XqeGtX).

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -1678,8 +1678,8 @@ uint32 Unit::CalcSpellResistedDamage(Unit* victim, uint32 damage, SpellSchoolMas
         if (spellInfo->HasAttribute(SPELL_ATTR4_IGNORE_RESISTANCES))
             return 0;
 
-        // Binary spells can't have damage part resisted
-        if (spellInfo->HasAttribute(SPELL_ATTR0_CU_BINARY_SPELL))
+        // Damage dealt by binary spells cast by players cannot be partially resisted
+        if (GetTypeId() == TYPEID_PLAYER && spellInfo->HasAttribute(SPELL_ATTR0_CU_BINARY_SPELL))
```
ccrs: shouldnt this also check for victim's typeid?

Xyventh: I don't think so, only caster is relevant here - target can be of any type.

ccrs: so player vs player also ignores binary? thats weird

Xyventh: This checks considers binary spells only when they're cast by player, it ignores binary spells when caster is creature but not when it's player, so PvP behavior isn't altered :P

xinef1: What about pets?

joshwhedon: Should use IsControlledByPlayer() in order to handle pets correctly.
Now my understanding is that binary resistances are only handled when the caster is a player, therefore we also need to add a check in MagicSpellHitResult().

joshwhedon:
Xyventh: this spell, Cataclysmic Bolt, is it binary ? I don't think it is, since it has a direct damage component.

ariel-:
And it stuns

src/server/game/Entities/Unit/Unit.cpp
ariel-: > Now my understanding is that binary resistances are only handled when the caster is a player, therefore we also need to add a check in MagicSpellHitResult().

That's wrong, what would the purpose of stacking resistance on magic heavy fights in that case?

joshwhedon: For partial resists on non binary spells ? The problem with binary resists in PVE is that they can have bad effect on the fight. I remember for example on the jormungars at the trial of the crusader, once in a while the tank was resisting the bite mechanic, and therefore, other players could not go to him to get the other debuff removed.
If not, then should IGNORE_HIT_RESULT attribute have the same effect as IGNORE_RESISTANCE for binary spells (if it is not already the case, maybe it is).

jackpoz:

Issue #19490, opened 2017-04-22
[7.2] Addons communicating with other clients


shocky4:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

**Description:**
I've noticed a few addons aren't communicating with other clients/players like it should. Best example is Total RP 3, it doesn't seem to be communicating with other players clients.
Here's how the tooltip should look when you hover over somebody that has the addon, and the bar when you click the player:
https://i.imgur.com/NTGoQun.jpg

**Current behaviour:**
Currently, when you hover over a player that has the addon, nothing happens, it's just the standard tooltip, and no bar from the addon when you click the player.

**Expected behaviour:**
It should show the tooltip, as well as the players version of the addon and all their information and customization that they put in their side of the addon.

**Steps to reproduce the problem:**
2 players have the Total RP 3 addon, if you hover over the player or click them, you don't see what you should with the addon. It affects other addons as well.

**Branch(es):**
Master

**TC rev. hash/commit:**
Both the last 7.1.5 commit: https://github.com/TrinityCore/TrinityCore/commit/886e9e839b304b7567dd655e4fa3afb2251c96ca
And latest 7.2 commit: https://github.com/TrinityCore/TrinityCore/commit/a316ba576b4c5316da2960340c4e05274e19b258

**TDB version:**  TDB 720.00

**Operating system:** Tried on both Windows 10 and Debian 8/Linux

[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (For fixes containing C++, create a Pull Request.)

MrSmite:
This might be a bug with the addon itself:

> https://mods.curse.com/addons/wow/total-rp-3
> *Posted 5 hours ago*
> Huh...the scan for roleplay doesn't detect anyone anymore..in fact it hasn't for a few updates...what's happened?

shocky4:
Hmm I don't think it's the addon itself since it works fine on retail for me when I tried on 2 separate accounts like the image, but doesn't seem to work on the latest Trinity. But I didn't know about the info you linked, that's informative

MrSmite:
> But I didn't know about the info you linked, that's informative

Yeah, sorry I don't have a working 7.x client / server yet so even though I understand LUA, I can't do any real debugging to see if it is the addon itself.

Pixel-Pirate:
TrinityCore don't provide support for a problem with specific addon

gmillerd:
[...]

PR #19534, opened 2017-05-01
Core/Channel: Added prefix to function to allow addon communication through the channel addon opcode

Syloxus:
**Changes proposed:**
-  Added 'prefix' to the 'Say' command in Channel to allow addon communication, referencing issue: https://github.com/TrinityCore/TrinityCore/issues/19490

- [x] master

**Issues addressed:**
https://github.com/TrinityCore/TrinityCore/issues/19490

- [x] Builds, also tested it and addons now receive a message back from the channel opcode.

- The addons that use the MSP (Mary Sue Protocol) still don't communicate to other players clients correctly, it doesn't send anything back to the clients, only requests.

### DIFF
--- a/src/server/game/Chat/Channels/Channel.cpp
+++ b/src/server/game/Chat/Channels/Channel.cpp
@@ -695,7 +695,7 @@ void Channel::Announce(Player const* player)
     UpdateChannelInDB();
 }
 
-void Channel::Say(ObjectGuid const& guid, std::string const& what, uint32 lang) const
+void Channel::Say(ObjectGuid const& guid, std::string const& what, uint32 lang, std::string const& prefix) const
 {
     if (what.empty())
         return;
@@ -727,10 +727,10 @@ void Channel::Say(ObjectGuid const& guid, std::string const& what, uint32 lang) 
 
         WorldPackets::Chat::Chat* packet = new WorldPackets::Chat::Chat();
         if (Player* player = ObjectAccessor::FindConnectedPlayer(guid))
-            packet->Initialize(CHAT_MSG_CHANNEL, Language(lang), player, player, what, 0, GetName(localeIdx));
+            packet->Initialize(CHAT_MSG_CHANNEL, Language(lang), player, player, what, 0, GetName(localeIdx), DEFAULT_LOCALE, prefix);
         else
         {
-            packet->Initialize(CHAT_MSG_CHANNEL, Language(lang), nullptr, nullptr, what, 0, GetName(localeIdx));
+            packet->Initialize(CHAT_MSG_CHANNEL, Language(lang), nullptr, nullptr, what, 0, GetName(localeIdx), DEFAULT_LOCALE, prefix);
             packet->SenderGUID = guid;
             packet->TargetGUID = guid;
         }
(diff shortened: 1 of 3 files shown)


src/server/game/Handlers/ChatHandler.cpp
```diff
@@ -384,7 +384,7 @@ void WorldSession::HandleChatMessage(ChatMsg type, uint32 lang, std::string msg,
             if (Channel* chn = ChannelMgr::GetChannelForPlayerByNamePart(target, sender))
             {
                 sScriptMgr->OnPlayerChat(sender, type, lang, msg, chn);
-                chn->Say(sender->GetGUID(), msg.c_str(), lang);
+                chn->Say(sender->GetGUID(), msg.c_str(), lang, nullptr);
```
Shauren: This will cause a segfault. You cannot construct a std::string using nullptr

Syloxus: Oh, I didn't realize. What would be best to put there?

Shauren: Empty string, ""

DDuarte: Couldn't prefix in Say have the default value of ""?

Shauren: It can

Syloxus: So in Channel.h, for the function do: `void Say(ObjectGuid const& guid, std::string const& what, uint32 lang, std::string prefix = "") const;`, in Channel.cpp, make the function `void Channel::Say(ObjectGuid const& guid, std::string const& what, uint32 lang, std::string prefix /*= ""*/) const`, then just "" instead of nullptr?

Keader: yes

src/server/game/Handlers/ChatHandler.cpp
```diff
@@ -384,7 +384,7 @@ void WorldSession::HandleChatMessage(ChatMsg type, uint32 lang, std::string msg,
             if (Channel* chn = ChannelMgr::GetChannelForPlayerByNamePart(target, sender))
             {
                 sScriptMgr->OnPlayerChat(sender, type, lang, msg, chn);
-                chn->Say(sender->GetGUID(), msg.c_str(), lang, nullptr);
+                chn->Say(sender->GetGUID(), msg.c_str(), lang, "");
```
Keader: "" is new default value, so it can be omitted
chn->Say(sender->GetGUID(), msg.c_str(), lang); should work

src/server/game/Chat/Channels/Channel.cpp
```diff
@@ -704,6 +704,11 @@ void Channel::Say(ObjectGuid const& guid, std::string const& what, uint32 lang,
     if (sWorld->getBoolConfig(CONFIG_ALLOW_TWO_SIDE_INTERACTION_CHANNEL))
         lang = LANG_UNIVERSAL;
 
+    Player* player = ObjectAccessor::FindConnectedPlayer(guid);
+
+    if (!player->GetSession()->IsAddonRegistered(prefix))
```
DDuarte: I don't think that this condition should be here. The message to be sent might not be related to addons at all.

DDuarte: It probably can be moved to `HandleChatAddonMessage`

Syloxus: Hmm, you're probably right, although I added this condition and it's the only way the addon would get a response back from the broadcast channel when debugged.

PR #19560, opened 2017-05-06
Autodetect msvc versions

KonTy:
**Changes proposed:**

-  
-  
-  

- [X] 3.3.5
- [X] master

### DIFF
--- a/dep/boost/CMakeLists.txt
+++ b/dep/boost/CMakeLists.txt
@@ -6,7 +6,7 @@
 #
 # This program is distributed in the hope that it will be useful, but
 # WITHOUT ANY WARRANTY, to the extent permitted by law; without even the
-# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. "lib${PLATFORM}-msvc*"
 
 if(WIN32)
  set(BOOST_DEBUG ON)
@@ -15,8 +15,17 @@ if(WIN32)
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 19.0)
       set(BOOST_LIBRARYDIR ${BOOST_ROOT}/lib${PLATFORM}-msvc-12.0)
     else()
-      set(BOOST_LIBRARYDIR ${BOOST_ROOT}/lib${PLATFORM}-msvc-14.0)
+	  # new boost 64.* upped the version to 14.1 from 14.0
+	  # adding logic to detect version automatically
+	  file(GLOB MSVC_LIB_DIRECTORIES ${BOOST_ROOT}/lib${PLATFORM}-msvc*)
+	  if(DEFINED MSVC_LIB_DIRECTORIES)
+		list(GET MSVC_LIB_DIRECTORIES 0 BOOST_LIBRARYDIR)
+	  else()
+		set(BOOST_LIBRARYDIR ${BOOST_ROOT}/lib${PLATFORM}-msvc-14.0)
+	  endif()
+	  
     endif()
+	message("BOOST_LIBRARYDIR is set to ${BOOST_LIBRARYDIR}")
   else()
     message(FATAL_ERROR "No BOOST_ROOT environment variable could be found! Please make sure it is set and the points to your Boost installation.")
   endif()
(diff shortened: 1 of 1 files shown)


dep/boost/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 #
 # This program is distributed in the hope that it will be useful, but
 # WITHOUT ANY WARRANTY, to the extent permitted by law; without even the
-# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. "lib${PLATFORM}-msvc*"
```
Aokromes: i suppose this is unintended change

Rochet2:
What happens if there are multiple boost installed like this?
I _think_ it would choose completely wrong boost, like 8.0
explorer_2017-03-24_21-13-25

tkrokli:
@Rochet2 : isn't Boost selected by checking the Windows Environment Variable `BOOST_ROOT` ?
That value will usually be set manually by the user to match the desired CMake & VS setup.

dep/boost/CMakeLists.txt
KonTy: Yes this was unintended thanks for catching

KonTy:
@Rochet2 what is the "right" boost version in your mind? I will chose the first boost it sees, how do you think we should fix it? We could have a wild card include 14 like this: lib${PLATFORM}-msvc.14* 
but this will miss next versions of boost 15,16 etc.

Aokromes:
what Rochet2 mentions remind me when we tried to do the same for mysql, the result, a lot of people where unable to compile because auto selected mysql connector instead mysql.

KonTy:
@Aokromes in this case we should fix the docs, because here it says it is recommended to use boost 64
https://trinitycore.atlassian.net/wiki/display/tc/Windows+Requirements 
"1.63.0 is the minimum version required, but version 1.64.0 is recommended (fully supports Visual Studio 2017)"

but the script doesn't support 64 it will not compile with 64 installed because 64 uses "lib32-msvc-14.1" not lib32-msvc-14.0.

There has to be a better way...

Aokromes:
or fix the code :P

ariel-:
3049d19b53db7e7f2042ee837771f7a71ad31373 ?

Rochet2:
> @Rochet2 : isn't Boost selected by checking the Windows Environment Variable BOOST_ROOT ?
> That value will usually be set manually by the user to match the desired CMake & VS setup.

That only sets the root, those folders are inside the root. The boost root will only show where boost is installed, not which boost version to use.

> @Rochet2 what is the "right" boost version in your mind? I will chose the first boost it sees, how do you think we should fix it? We could have a wild card include 14 like this: lib${PLATFORM}-msvc.14*

The one right for the compiler. IE 14.1 if 8.0 and 14.0 and 14.1 are installed.

> 3049d19 ?

@ariel- I _guess_ that fixed the problem for 3.3.5 but isnt applied to master yet?
I wonder how it works with multiple libraries and compilers selected like that.

jackpoz:
where's the PR description ?

KonTy:
@jackpoz it's in the title there isn't much more to add to that.
@ariel- 
> 3049d19 I personally do not like that solution because it is hardcoding the versions.
perhaps for 2017 we can pick a range? ex: <= 14.0 

here is what I did and tested it extensively

`list(LENGTH MSVC_LIB_DIRECTORIES MSVC_LIB_DIRECTORIES_LENGTH)
MATH(EXPR MSVC_LAST_ELEMENT "${MSVC_LIB_DIRECTORIES_LENGTH}-1")
list(SORT MSVC_LIB_DIRECTORIES)
list(GET MSVC_LIB_DIRECTORIES ${MSVC_LAST_ELEMENT} BOOST_LIBRARYDIR)`

DDuarte:
What's the behavior of the official FindBoost.cmake on this topic?

ariel-:
> What's the behavior of the official FindBoost.cmake on this topic?

The one I ported in 3049d19b53db7e7f2042ee837771f7a71ad31373, it's TC that hardcodes and "auto selects" best version here: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/dep/boost/CMakeLists.txt#L11-L33

joschiwald:
> What's the behavior of the official FindBoost.cmake on this topic?
yeah we should update our FindBoost.cmake, many stuff changed since last 2 years
for me, it is working well

Aokromes:
Conflicting files
dep/boost/CMakeLists.txt

Aokromes:
why close?

PR #19784, opened 2017-05-22
Core/Scripts : Proof of concept for creature DB ScriptParam

Traesh:
**Changes proposed:**

-  Add a new field in "creature" and "creature_template" tables
-  This field can be set in "creature_template" and overwritten in "creature"
-  The value can be used in C++ script to customize script depending of creature entry/guid

**Target branch(es):**

- [x] master

None ATM, wait for the concept review

None


Keader:
Is not the same thing ? https://github.com/TrinityCore/TrinityCore/commit/5a36a2eb17aeec937b4375f341239fdb9dce5fda

Traesh:
No, https://github.com/TrinityCore/TrinityCore/commit/5a36a2eb17aeec937b4375f341239fdb9dce5fda was about allowing ScriptName by guid, this one is about allowing passing param to C++ script by DB (entry or guid)

ariel-:
I'm not sure, can't the same result be achieved using SmartAI (for eg Set invincibilty pct)

Traesh:
Yes it can for some actions, but not all atm.

The "npc_creature_damage_limit " is an example saw yesterday for a Maclrom request "limit maximum health loss by a creature if hit by another creature", the point is that health pct depend of the creature entry or guid.

Firsts reaction mentioned the fact that adding new fields in creature_template was a bad idea, Shauren suggested creating new table(s) "creature_template_script_param" and/or "creature_script_param"

It could probably be a SmartAI (new event 78 "SMART_EVENT_HEALTH_AFTER_DAMAGED" with "uint32 healthPct" as param1 and "bool fromCreatureOnly" as param2, new action 129 "PREVENT_DAMAGE" ?) but i feel like it would add a lot of lines in SmartAI for nothing, and loose a lot of readability. Moreover, adding a new C++ script is pretty easy and Malcrom's request could be addressed easily, adding a new SmartAI Event/Action take more time

Examples :

old
`
INSERT INTO `smart_scripts` (entryorguid, source_type, id, event_type, event_chance, event_param1, event_param2, action_type, target_type) VALUES
(xxxxx, 0, 0, 78, 100, 80, 1, 129, 1);
`

new
`UPDATE creature_template SET ScriptName = "npc_creature_damage_limit", ScriptParam0 = 80 WHERE entry = xxxxx;`

jackpoz:
> Shauren suggested creating new table(s) "creature_script_param_template" and/or "creature_script_param"

I'm in favour of creating a new table with multiple columns for each parameter type (some uint64 columns, some float columns, some varchar columns) to allow this system to scale up to future requirements.

Traesh:
@jackpoz Maybe "float" and "string" would be enough, is this usefull to have an int and a float field ?

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -431,8 +431,8 @@ void ObjectMgr::LoadCreatureTemplates()
                                              "spell1, spell2, spell3, spell4, spell5, spell6, spell7, spell8, VehicleId, mingold, maxgold, AIName, MovementType, "
     //                                        65           66           67              68                   69            70                 71             72              73
                                              "InhabitType, HoverHeight, HealthModifier, HealthModifierExtra, ManaModifier, ManaModifierExtra, ArmorModifier, DamageModifier, ExperienceModifier, "
-    //                                        74            75          76           77                    78           79
-                                             "RacialLeader, movementId, RegenHealth, mechanic_immune_mask, flags_extra, ScriptName FROM creature_template");
+    //                                        74            75          76           77                    78           79               80
+                                             "RacialLeader, movementId, RegenHealth, mechanic_immune_mask, flags_extra, ScriptName, ScriptParam0 FROM creature_template");
```
tkrokli: in your next PR update, please reduce indent for '80' by 5 spaces:
```cpp
    //                                        74            75          76           77                    78           79          80
                                             "RacialLeader, movementId, RegenHealth, mechanic_immune_mask, flags_extra, ScriptName, ScriptParam0 FROM creature_template");
```

Traesh: Next update will delete this modification to create new table(s)

jackpoz:
> and lose a lot of readability

By the way readability was dropped long time ago when SAI was added. As you said this case (and probably most of them) can be handled in SAI. It would keep everything in the same table instead of having to look around for all the possible ways to script a creature.

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -1857,8 +1924,8 @@ void ObjectMgr::LoadCreatures()
     QueryResult result = WorldDatabase.Query("SELECT creature.guid, id, map, modelid, equipment_id, position_x, position_y, position_z, orientation, spawntimesecs, spawndist, "
     //   11               12         13       14            15         16          17          18                19                   20                    21
         "currentwaypoint, curhealth, curmana, MovementType, spawnMask, eventEntry, pool_entry, creature.npcflag, creature.unit_flags, creature.unit_flags2, creature.unit_flags3, "
-    //   22                     23                24                   25
-        "creature.dynamicflags, creature.phaseid, creature.phasegroup, creature.ScriptName "
+    //   22                     23                24                   25                   26
+        "creature.dynamicflags, creature.phaseid, creature.phasegroup, creature.ScriptName, creature.ScriptParam0 "
```
PalWow: creature.ScriptParam0  ？

Traesh: Missing remove from "switch to new table" commit, thx

Ascathor:
I still don't see the exact reason for a new table over adding features to SmartAI. That's actually exactly what SmartAI was supposed to be used for, avoiding c++ scripts over a maintainable, robust functionality with multiple uses. If we actually need to look at readability, we should overhaul the way a specific guid or entry is referenced. Seems more useful imo.

malcrom:
Are you mad. I tried to get Sai overhauled for years to no avail. Anyway we need a way to do sparing. It's in every zone and needed to spawn properly.

tkrokli:
@Traesh : Sorry, but you need to rebase your PR on current TC source to include Shauren's [**Core/Game: Include cleanup part 5**](https://github.com/TrinityCore/TrinityCore/commit/b453e124231a90321fe79fbf3a62acdcfa54a691) commit https://github.com/TrinityCore/TrinityCore/commit/b453e124231a90321fe79fbf3a62acdcfa54a691 from 2 days ago, 2017-06-04 01:00 CEST. All the `enum CreatureDifficultyFlags` (1-7) has moved to https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Creature/CreatureData.h in the TC source, so the TC file `Creature.h` is less than half of its original length. Sorry about that, but all I did in my latest commit to this PR was to move a remaining TC source line to above your still remaining `enum CreatureDifficultyFlags` sections.

PR #19788, opened 2017-05-23
Scripts/Commands: modify command .dismount to target any selected player

ForesterDev:
**Changes proposed:**

-  Improve .dismount command  to target any selected player, not only current player

3.3.5a


src/server/scripts/Commands/cs_misc.cpp
```diff
@@ -645,7 +645,13 @@ class misc_commandscript : public CommandScript
 
     static bool HandleDismountCommand(ChatHandler* handler, char const* /*args*/)
     {
-        Player* player = handler->GetSession()->GetPlayer();
+        Player* player = handler->getSelectedPlayerOrSelf();
+        if (!player)
```
DJScias: Pretty sure that `getSelectedPlayerOrSelf()` will never be null, because player becomes yourself if you're not targeting any actual player.

This sorta makes this check unnecessary.

ForesterDev: getSelectedPlayerOrSelf can return NULL

ForesterDev: This check is unnecessary now, but it will prevent server to crash if someone allow to use this command from console

DJScias: Logic dictates that nobody should run player commands through the console, but you're not wrong I guess.

Good stuff. 🎉

myran2: Doesn't the "false" on line 64 of this file mean that the dismount command can't be used from the console?

Raydor: Yep

ariel-:
This could potentially introduce privilege escalation, as the .dismount command is available to players by default.

Maybe you should add another RBAC permission that allows using dismount on other players

myran2:
Aren't gm commands disabled by default for player accounts?

ariel-:
> Aren't gm commands disabled by default for player accounts?

Yes, but .dismount isn't a GM command, like .help, .commands, .account lock...

ForesterDev:
WoW Client supports /dismount command, that have same functional. Why do we need to have this command for players? Maybe we should change security lvl for this command for gms?

Aokromes:
It looks logical.

Ascathor:
I agree to this last change. /dismount is sufficient for players. 

Pull it, looks good imo.

Aokromes:
Conflicting files
sql/base/auth_database.sql

src/server/scripts/Commands/cs_misc.cpp
jackpoz: if the NULL check is not needed then please remove it

jackpoz: please remove the null check

jackpoz: please remove the null check

jackpoz (changes requested):

tkrokli:
Will you be revisiting this PR anytime soon, @ForesterDev ?

jackpoz (approved):

jackpoz:
I removed the null check, anyone up to update the auth_database.sql file so it can be merged ?

tkrokli:
Still working on producing the updates info...

PR #19813, opened 2017-05-27
Add functionality to allow low level quests/kills/discoveries to grant experience

NathanHandley:
**Changes proposed:**

-  Add WorldServer configuration 'MinQuestScaledXPRatio'
This determines the minimum returned experience for completing a quest with a 0...100 ratio, based on a player level to quest level index.  For example, completing a level 20 quest with MinQuestScaledXPRatio set to 75 will make it so the quests will return no less than 75% of the experience it would have granted a level 20 player.  If set to 100, quest EXP will always reward full experience.  If set to 0 (default) there is no minimum enforced.
-  Add WorldServer configuration 'MinCreatureScaledXPRatio'
Similar to above, this determines the minimum returned experience for killing a monster based with a 0...100 ratio indexed on monster level.  For example, killing a level 15 experience-granting target with MinCreatureScaledXPRatio set to 50 will ensure a return of at least 50% of the experience a level 15 player would receive.  If set to 100, full experience will be returned regardless of player level.  If set to 0 (default) there is no minimum enforced.  
As a special note, this does properly take into account spell effects and triggers which use the "... grants experience or honor" mechanic, such as Drain Soul and Curse of Doom.  This was also noted in the worldserver.conf comment area for the config value to avoid confusion.
-  Add WorldServer configuration 'MinDiscoveredScaledXPRatio'
Similar to above, this determines the minimum returned experience for discovering a subpoint on a map based with a 0...100 ratio indexed on player level.  For example, discovering a level 20 map location with MinDiscoveredScaledXPRatio set to 25 will ensure no less than 25% of the experience it would have granted a level 20 player.  And so on.  If set to 0 (default) there is no minimum enforced.  

In all of the above changes, normal scaling in the engine is unchanged.  These settings only influence the minimum or 'low bar' of experience returned from the actions, should the calculated scaled experience drop below the configured ratio.

- [X] 3.3.5

**Issues addressed:**
New functionality.

**Tests performed:**
Builds and works properly.  See below attached for the test scripts executed:
testresults


src/server/game/Entities/Player/Player.cpp
```diff
@@ -6596,24 +6596,27 @@ void Player::CheckAreaExploreAndOutdoor()
...
+                    scaledXP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*exploration_percent/100*sWorld->getRate(RATE_XP_EXPLORE));
                 }
                 else
                 {
-                    XP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*sWorld->getRate(RATE_XP_EXPLORE));
+                    scaledXP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*sWorld->getRate(RATE_XP_EXPLORE));
                 }
 
+                uint32 minScaledXP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*sWorld->getRate(RATE_XP_EXPLORE)) * sWorld->getIntConfig(CONFIG_MIN_DISCOVERED_SCALED_XP_RATIO) / 100;
+                uint32 XP = uint32(minScaledXP >= scaledXP ? minScaledXP : scaledXP);
```
ariel-: Use std::max

src/server/game/Quests/QuestDef.cpp
```diff
@@ -211,17 +211,31 @@ uint32 Quest::XPValue(Player* player) const
...
+            minScaledXP = 10 * ((minScaledXP + 5) / 10);
+        else if (minScaledXP <= 1000)
+            minScaledXP = 25 * ((minScaledXP + 12) / 25);
+        else
+            minScaledXP = 50 * ((minScaledXP + 25) / 50);
+        minScaledXP = minScaledXP * sWorld->getIntConfig(CONFIG_MIN_QUEST_SCALED_XP_RATIO) / 100;
 
-        return xp;
+        // Use greater between non-level scaled and scaled
+        return uint32(minScaledXP >= scaledXP ? minScaledXP : scaledXP);
```
ariel-: Here too

src/server/game/Miscellaneous/Formulas.h
```diff
@@ -134,26 +133,33 @@ namespace Trinity
...
                 else
-                    baseGain = 0;
+                    baseGainScaled = 0;
             }
 
+            // Calculate unscaled based gain min.  Use mob level instead of player level to avoid overscaling on gain
+            uint32 baseGainUnscaledMin = (mob_level * 5 + nBaseExp) * sWorld->getIntConfig(CONFIG_MIN_CREATURE_SCALED_XP_RATIO) / 100;
+
+            // Use largest
+            uint32 baseGain = baseGainUnscaledMin >= baseGainScaled ? baseGainUnscaledMin : baseGainScaled;
```
ariel-: Here too

NathanHandley:
Thanks for the review feedback, ariel.  Done.

src/server/game/Miscellaneous/Formulas.h
```diff
@@ -134,26 +133,33 @@ namespace Trinity
...
-                    baseGain = (pl_level * 5 + nBaseExp) * (ZD + mob_level - pl_level) / ZD;
+                    baseGainScaled = (pl_level * 5 + nBaseExp) * (ZD + mob_level - pl_level) / ZD;
                 }
                 else
-                    baseGain = 0;
+                    baseGainScaled = 0;
             }
 
+            // Calculate unscaled based gain min.  Use mob level instead of player level to avoid overscaling on gain
+            uint32 baseGainUnscaledMin = (mob_level * 5 + nBaseExp) * sWorld->getIntConfig(CONFIG_MIN_CREATURE_SCALED_XP_RATIO) / 100;
```
Treeston: Move this and the `std::max` call into an if branch checking your world config to make it explicit that this is custom option code.

Also, remove the rename above (leave `baseGain`, replace its value in the if branch if appropriate).

src/server/game/Entities/Player/Player.cpp
```diff
@@ -6596,24 +6596,27 @@ void Player::CheckAreaExploreAndOutdoor()
...
-                    XP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*exploration_percent/100*sWorld->getRate(RATE_XP_EXPLORE));
+                    scaledXP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*exploration_percent/100*sWorld->getRate(RATE_XP_EXPLORE));
                 }
                 else
                 {
-                    XP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*sWorld->getRate(RATE_XP_EXPLORE));
+                    scaledXP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*sWorld->getRate(RATE_XP_EXPLORE));
                 }
 
+                uint32 minScaledXP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*sWorld->getRate(RATE_XP_EXPLORE)) * sWorld->getIntConfig(CONFIG_MIN_DISCOVERED_SCALED_XP_RATIO) / 100;
```
Treeston: Move this into an if branch on your config value. Revert the rename above (leave it as `XP`, then replace `XP` in your if branch if appropriate).

src/server/game/Quests/QuestDef.cpp
```diff
@@ -211,17 +211,31 @@ uint32 Quest::XPValue(Player* player) const
...
+        else if (scaledXP <= 500)
+            scaledXP = 10 * ((scaledXP + 5) / 10);
+        else if (scaledXP <= 1000)
+            scaledXP = 25 * ((scaledXP + 12) / 25);
         else
-            xp = 50 * ((xp + 25) / 50);
+            scaledXP = 50 * ((scaledXP + 25) / 50);
+
+        // Calculate non-level scaled xp
+        uint32 minScaledXP = xpentry->Exp[_rewardXPDifficulty];
```
Treeston: This is some serious code duplication right there. Get rid of it and do your replacement before all the if branchings (and, as above, put the replacement in an explicit if branch on your config entry).

Treeston (commented):
This is custom functionality. As such, it needs to be clearly separate from vanilla logic, evident from the source code structure at a glance.

Also, don't refactor variable names for no particular reason, it just makes your diff confusing to review.

NathanHandley:
Thanks.  I'll get working on the rewrite.

NathanHandley:
Completed.  Please re-review.  I also re-executed the test scripts on the original post to ensure accuracy after the refactor.  It's good.

src/server/game/Quests/QuestDef.h
```diff
@@ -388,6 +388,9 @@ class TC_GAME_API Quest
         uint32 _startItemCount        = 0;
         uint32 _rewardMailSenderEntry = 0;
         uint32 _specialFlags          = 0; // custom flags, not sniffed/WDB
+
+        // Helpers
+        uint32 RoundXPValue(uint32 xp) const;
```
ariel-: This helper function doesn't need access to any of the Quest object members, it can be made static, or even live only in the .cpp as non-member function

NathanHandley: Which would you prefer?  I'm thinking static having to choose between the two, since it's relevant only to the scope of quest.

Also, is this a general pattern adherence used in this code base?  That is, not to have private methods for internal class methods?  Or is it due to the nature of the purpose of the Quest class (abstraction).  If so, I'll keep that in mind from now on when I need class subroutines.

Treeston: If it has no use for object properties, then there's no reason for it to be a method on the object.

NathanHandley: Understood.  I'll put as a non-class bound method in the cpp then, and rename appropriately.

Treeston: Leave it as `static`. Random in-file methods are ugly.

NathanHandley: Okay

NathanHandley:
Done.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -6614,6 +6614,12 @@ void Player::CheckAreaExploreAndOutdoor()
                     XP = uint32(sObjectMgr->GetBaseXP(areaEntry->area_level)*sWorld->getRate(RATE_XP_EXPLORE));
                 }
 
+                if (sWorld->getIntConfig(CONFIG_MIN_DISCOVERED_SCALED_XP_RATIO) != 0)
```
Treeston: `if (sWorld->getIntConfig(CONFIG_MIN_DISCOVERED_SCALED_XP_RATIO))`

src/server/game/Entities/Player/Player.cpp
```diff
@@ -23656,7 +23662,7 @@ bool Player::isHonorOrXPTarget(Unit* victim) const
     uint8 k_grey  = Trinity::XP::GetGrayLevel(getLevel());
 
     // Victim level less gray level
-    if (v_level <= k_grey)
+    if (v_level <= k_grey && sWorld->getIntConfig(CONFIG_MIN_CREATURE_SCALED_XP_RATIO) == 0)
```
Treeston: `if (v_level <= k_grey && !sWorld->getIntConfig(CONFIG_MIN_CREATURE_SCALED_XP_RATIO))`

src/server/game/Miscellaneous/Formulas.h
```diff
@@ -154,6 +154,13 @@ namespace Trinity
                     baseGain = 0;
             }
 
+            if (sWorld->getIntConfig(CONFIG_MIN_CREATURE_SCALED_XP_RATIO) != 0)
```
Treeston: `if (sWorld->getIntConfig(CONFIG_MIN_CREATURE_SCALED_XP_RATIO))`

src/server/game/Quests/QuestDef.cpp
```diff
@@ -211,15 +211,12 @@ uint32 Quest::XPValue(Player* player) const
...
-        if (xp <= 100)
-            xp = 5 * ((xp + 2) / 5);
-        else if (xp <= 500)
-            xp = 10 * ((xp + 5) / 10);
-        else if (xp <= 1000)
-            xp = 25 * ((xp + 12) / 25);
-        else
-            xp = 50 * ((xp + 25) / 50);
+        uint32 xp = RoundXPValue(diffFactor * xpentry->Exp[_rewardXPDifficulty] / 10);
+        if (sWorld->getIntConfig(CONFIG_MIN_QUEST_SCALED_XP_RATIO) != 0)
```
Treeston: `if (sWorld->getIntConfig(CONFIG_MIN_QUEST_SCALED_XP_RATIO))`

NathanHandley:
Made the changes requested, Triston.  Thanks for the quick feedback on these change sets!

I try to avoid implicit integer->boolean compiler conversions for conditionals from a non local scope variable due to potential future data type changes and/or future overload behavior patterns.  However given that the method name suggest the data type is fixed and how close-to-stock this code base works with the primitive types, the risk of this change having ill impact later is far too low to really protest over.  

I'll also add this to my notes of code conduct for this project tree to make it quicker next time.  If I may suggest, please add this to the code guidelines for trinitycore since it's not addressed there, nor is it present in the referenced google C++ style guide.

Treeston:
Generally, we recommend readable source code. That means that, if possible, variables should be boolean-checkable in a sensible way (true = "enabled"/"yes"/"present"/"empty", false = "disabled"/"no"/"not present"/"empty" etc.)

In this case, that means that boolean true <-> config key is enabled, and boolean check conveys the developer intent properly.

NathanHandley:
Without being strongly familiar with the code base there's no clarity of the boolean comparitor is an existence check ("true == a config has been set") vs a value result ("true == the value of the config or default is non-zero").  I bring this up because I've worked in a code base where the bool comparison overload was used as an existence check. (i.e, set in a config), regardless of default/result -- similar to null checking.  Specifically the code on player.cs line 23659 needs to fire if there is a non zero value in that value regardless if a player set anything in the worldserver.conf.

Treeston:
There's no such thing as an unset config value in the code base. Everything has a default value.

NathanHandley:
Yep, totally with you.  This is ignorance on my behalf because I haven't worked in this specific code based very much and without looking at parent context, that's not apparent.  That's partially why I'd like it added to the code standards/guidelines since those are designed (I think) for non-member contributors.

Treeston:
Oh, you sweet summer child. Still so wide-eyed and hopeful 🙄.

Issue #19485, opened 2017-04-21
[3.3.5] Guardians put target in combat before reaching target


Keader:
**Description:** If you send a Guardian to attack your target, it will put the target in combat even if your Guardian is out of range.
E.g.  Frost Mage, Summon Water Elemental (31687)

When you press attack button (CTRL+1), the target will enter combat before the Water Elemental reaches target.

**Expected behaviour:** Target should enter combat when the Guardian reaches target (like when using pets).

PS: It works for pets, but does not work for "temp Pets/Guardians"

**Steps to reproduce the problem:**

1. Cast spell 31687 (Summon Water Elemental)
2. Send water elemental to attack some player out of range
3. Player enters combat state before water elemental reaches him.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** f42ff8c6ba0609c2f774440dd30981356c1ffd63

ariel-:
https://github.com/TrinityCore/TrinityCore/commit/69fd6245dc41dca658cadb4e99c1ee6593fe000b

ccrs:
actually no

ariel-:
> Closes #19768, #19485 and #10921.

Based on commit msg xd

Treeston:
Was my mistake, we found out it breaks in some niche situations still. #19816 will address once we sort out bugs with aggro behavior on that one.

jackpoz:
what's the status of this PR and its relation to the other referenced PR ?

ccrs:
this is not a PR xd

tkrokli:
(been there myself, made that mistake not too long ago :-P )

Killyana:
Fixed by https://github.com/TrinityCore/TrinityCore/commit/a891eb364e941a5146fba4af16e1ecf65cdd9239

Issue #4943, opened 2012-01-24
[General] Evade/Movement Issues


Amit86:
Hey
I didnt really know how to classify these problems im having

It started with the issue in ToC where some mobs got stuck in evade - this was fixed by applying this #4877 by @Vincent-Michael

The problem persist with other, more general mobs like world mobs etc
Players attack a mob, he keeps evading and when they try to run away he keeps following them to the end of time (or till server crashes/they die)

I have no idea how to reproduce this but im sure many of you have seen this problem or heard of it at least 1 time

Please confirm.

Hash: 8776158
##

ibakula:
One way to reproduce this "evade bug" is to go into the Zul'Aman and fight "Nalorakk", let him kill you and then release the spirit, enter the instance and fight him again.. which you'll be unable to because he'll evade all of your attacks or you'll possibly see "out of range" bug.

While I was debugging I stumbled upon the following.

Callstack:
Creature::CanStartAttack(who, false) Line: 1387
CreatureAI::MoveInLineOfSight(who) Line: 135
boss_nalorakk::boss_nalorakkAI::MoveInLineOfSight(Unit* Who) Line: 218

Autos:
CanFly() returned false
GetDistanceZ(who) returned 7.xxxf

I am not sure if I was of any help but this is what I've managed to debug on my own.
Under "CreatureAI::MoveInLineofSight", Line on which you can find "me->CanStartAttack" then returns false and exits the MoveInLineOfSight function.

What would have fixed this issue is:
HomeMovementGenerator<Creature>::Finalize
If the HomeMovementGenerator<Creature>::arrived was evaluated to true, because the NPC would clear its Evade state.
Or:
boss_nalorakk::boss_nalorakkAI::Reset() Line 166
If you added "me->ClearUnitState(UNIT_STATE_EVADE);" after the line above

Both of these ways may be the incorrect way of fixing this.

Also, when scripting a creature movement and then using a Clear function, for some reason the boss becomes un-attackable.. "out of range"

Example of this issue can be found in "Zul'Jin"

After the clear function happens, the creature's unit state flags are equal to: 
UNIT_STATE_CASTING | UNIT_STATE_MELEE_ATTACKING

Zul'Jin in the eagle phase should not be moving.

I may not be good at programming or debugging at all but I'd be interested to see what's the proper way to resolve this.

[...]

PR #19814, opened 2017-05-27
Fix evade on aggro switch on neutral (yellow) mobs

### LINKED ISSUE #19768: [3.3.5] Faction bug
Description:
https://github.com/TrinityCore/TrinityCore/commit/15a207fcac15d336b848f9759f75386e12b6a35f
After this commit i found strange bug. For example - boss Midnight in Karazhan.
Npc 16151 have faction 7 (this faction don't have any friendly/hostile factions in dbc files), when you start attacking midnight by first player, and after re-aggro boss to second player (by abilities), this check
(https://github.com/TrinityCore/TrinityCore/commit/15a207fcac15d336b848f9759f75386e12b6a35f#diff-f3f953a63e2a008d22de69cb68f735c4R2292) is failing and boss entering evade mode. I think all npc with this faction can have this problem, or, with simular factions.

If creature in AGRESSIVE react state, then traget can't switch target to new, and continue attacking first target using "// search nearby enemy before enter evade mode" case.

Can you please review this situation ? if need i can record video. Can be reprodusing on last rev (3.3.5 branch). In additional you can check this using 20774 npc.

Branch(es):
3.3.5

TC rev. hash/commit:
(last rev)
TrinityCore rev. edd2c1c03995 2017-05-17 07:16:46 +0300 (master branch) (Unix, Release, Static) (worldserver-daemon)

Treeston:
Partial revert of 15a207f, which was causing issues (ref #4943 and #19768). Instead, fixes the underlying issue from 15a207f one level further down - `Unit::getAttackerForHelper()` shouldn't return units that we aren't in combat with (victim can be such a unit for players/player pets, which can startattack from a distance without entering combat).

Fixes the following issues:
- Player pets would aggro neutral mobs as soon as they start autocasting (Imp's Firebolt) if they're in react range (due to victim != null, autocast counts as autoattack and sets victim)
- Neutral mobs would randomly evade when aggro switched between targets.

Closes #19768 and #19485.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -5768,13 +5768,22 @@ void Unit::_removeAttacker(Unit* pAttacker)
 
 Unit* Unit::getAttackerForHelper() const                 // If someone wants to help, who to give them
 {
-    if (GetVictim() != NULL)
-        return GetVictim();
+    if (Unit* victim = GetVictim())
+        if (!IsControlledByPlayer() || IsInCombatWith(victim) || victim->IsInCombatWith(this))
+            return victim;
 
     if (!m_attackers.empty())
         return *(m_attackers.begin());
 
-    return NULL;
+    if (Player* owner = GetCharmerOrOwnerPlayerOrPlayerItself())
+    {
+        HostileRefManager& refs = owner->getHostileRefManager();
+        for (Reference<Unit, ThreatManager> const& ref : refs)
+            if (Unit* hostile = ref.GetSource()->GetOwner())
+                return hostile;
+    }
+
+    return nullptr;
 }
 
 bool Unit::Attack(Unit* victim, bool meleeAttack)
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2291,13 +2291,8 @@ bool Creature::_IsTargetAcceptable(Unit const* target) const
     Unit const* targetVictim = target->getAttackerForHelper();
 
     // if I'm already fighting target, or I'm hostile towards the target, the target is acceptable
-    if (GetVictim() == target || IsHostileTo(target))
+    if (IsInCombatWith(target) || IsHostileTo(target))
         return true;
-
-    // a player is targeting me, but I'm not hostile towards it, and not currently attacking it, the target is not acceptable
-    // (players may set their victim from a distance, and doesn't mean we should attack)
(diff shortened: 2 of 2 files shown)


ccrs:
``Unit::getAttackerForHelper() shouldn't return units that we aren't in combat with```
why? guardians without charm info use it to enter combat (for example assist on owner damaged)

Treeston:
Owner damaged means we've been attacked, so we're in combat.

The only situation where a Unit can be not in combat with its victim is when it's player controlled, where "victim" = "thing the player right-clicked to start auto attack". Here, combat should not trigger until the actual attack happens.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -5768,13 +5768,14 @@ void Unit::_removeAttacker(Unit* pAttacker)
 
 Unit* Unit::getAttackerForHelper() const                 // If someone wants to help, who to give them
 {
-    if (GetVictim() != NULL)
-        return GetVictim();
+    if (Unit* victim = GetVictim())
+        if (!IsControlledByPlayer() || IsInCombatWith(victim))
+            return GetVictim();
```
ariel-: return victim;

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -5768,13 +5768,22 @@ void Unit::_removeAttacker(Unit* pAttacker)
 
 Unit* Unit::getAttackerForHelper() const                 // If someone wants to help, who to give them
 {
-    if (GetVictim() != NULL)
-        return GetVictim();
+    if (Unit* victim = GetVictim())
+        if (!IsControlledByPlayer() || IsInCombatWith(victim) || victim->IsInCombatWith(this))
```
ariel-: Isn't IsInCombatWith reciprocal?

ccrs:
This brings up a mistake on some hooks like
````
OwnerAttackedBy OwnerAttacked
````
they are only triggered for Pet, not for Minions (Guardian etc)

src/server/game/Entities/Unit/Unit.cpp
Treeston: No, it checks threat list.

Treeston:
That's beyond the scope of this PR, tbf. (but yes)

Regrad:
Last revision.
Video of this bug:
https://youtu.be/gtvi9TgfLBs

Treeston:
@Regrad is that with this PR applied?

Regrad:
With the last patch the problem was completely solved. Many thanks.

PR #19824, opened 2017-05-29
Pet/Guardian AI hook re-organizing

Treeston:
Pet/Guardian AI hook re-organizing:
* Adjust OwnerAttacked/OwnerAttackedBy hooks on CreatureAI to fire for all owned units, not just player pets. This should allow guardians to more reliably recognize valid targets.
* Kill off the AttackedBy hook. While it was defined in CreatureAI.h as virtual, it was only ever invoked for player pets in specific situations. This makes it classic developer bait.
  * Adjust PetAI to use DamageTaken instead of AttackedBy.
  * Adjust behavior of AttackStart on PetAI to compensate.

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -136,17 +136,17 @@ void CreatureAI::MoveInLineOfSight(Unit* who)
         AttackStart(who);
 }
 
-void CreatureAI::OwnerAttackedBy(Unit* attacker)
+void CreatureAI::_OnOwnerCombatInteraction(Unit* target)
 {
-    if (!attacker || !me->IsAlive())
+    if (!target || !me->IsAlive())
         return;
 
-    if (!me->HasReactState(REACT_PASSIVE) && me->CanStartAttack(attacker, true))
+    if (!me->HasReactState(REACT_PASSIVE) && me->CanStartAttack(target, true))
     {
         if (me->IsInCombat())
-            me->AddThreat(attacker, 0.0f);
+            me->AddThreat(target, 0.0f);
         else
-            AttackStart(attacker);
+            AttackStart(target);
     }
 }
 
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -132,15 +132,22 @@ void CreatureAI::MoveInLineOfSight(Unit* who)
     if (me->GetVictim())
         return;
 
-    if (me->GetCreatureType() == CREATURE_TYPE_NON_COMBAT_PET) // non-combat pets should just stand there and look good;)
-        return;
-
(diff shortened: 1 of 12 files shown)


Shauren:
Shouldn't AttackedBy fire on events that cause aggro but not deal damage like taunt/debuff/dot without instant damage component?

Treeston:
In current core, EnterCombat is used if it's initial aggro, otherwise spell related hooks.
If we want AttackedBy to actually be a functional hook, that'd be a different PR - in its current state, it does nothing.

Shauren:
Well, all of the scenarios I described are valid even as non-initial-aggro

Treeston:
Yes, but currently `AttackedBy` won't fire in these scenarios.

src/server/game/AI/CreatureAI.cpp
```diff
@@ -132,15 +132,22 @@ void CreatureAI::MoveInLineOfSight(Unit* who)
...
-    //    && me->CanStartAttack(who->GetVictim(), true)) /// @todo if we use true, it will not attack it when it arrives
-    //    me->GetMotionMaster()->MoveChase(who->GetVictim());
+}
+
+void CreatureAI::OwnerAttackedBy(Unit* attacker)
+{
+    if (!attacker || !me->IsAlive())
+        return;
+
+    if (!me->HasReactState(REACT_PASSIVE) && me->CanStartAttack(attacker, false))
```
ccrs:
````
me->CanStartAttack(attacker, false)
````
prevents combat with neutral units that the owner is attacking
````
me->CanStartAttack(attacker, true)
````
testing with force = true

Comment (ccrs) [inline src/server/game/AI/CreatureAI.cpp]: actually makes sense for it to be true, faction check can be sort of ignored and distance check is useless

Comment (ccrs) [inline src/server/game/AI/CreatureAI.cpp]: "sort of ignored"

File: src/server/game/AI/CreatureAI.h [hunk 2]
```diff
@@ -152,15 +150,14 @@ class TC_GAME_API CreatureAI : public UnitAI
         virtual void ReceiveEmote(Player* /*player*/, uint32 /*emoteId*/) { }
 
         // Called when owner takes damage
-        virtual void OwnerAttackedBy(Unit* /*attacker*/) { }
+        virtual void OwnerAttackedBy(Unit* attacker);
 
         // Called when owner attacks something
-        virtual void OwnerAttacked(Unit* /*target*/) { }
+        virtual void OwnerAttacked(Unit* target) { CreatureAI::OwnerAttackedBy(target); }
```
Shauren: You talked about traps for scripters but here you add another one - it would be better to introduce a separate (nonvirtual) method that is called by both OwnerAttacked and OwnerAttackedBy. Someone less experienced might think that if he overrides OwnerAttackedBy in his own script but not OwnerAttacked that it will call his overriden OwnerAttackedBy (but it will be the base CreatureAI one)

Shauren (changes requested):

PR #19829, opened 2017-05-30
Core/Scene: fix possible PlayScene crash

Traesh:
**Changes proposed:**

-  Fix possible crash if PlayScene called from aura with non-handled scenes

**Target branch(es):**

- [x] master

none


codeman8214:
Works like a charm

P-Kito:
I don't understand how that crashed when the template was checked in the returning function anyways Oô

PR #19830, opened 2017-06-01
Core/Config: Add configurable scaling to in-game (server) time

Staleness89:
**Changes proposed:**

-  Adds a float config option to scale up (or down) the in game time. Default value in the config is set at 1, and is multiplied by the default game speed.

- [x] 3.3.5

n/a

Builds, tested in game with various values (100, 1000) and it works fine.

none


Staleness89:
Changing maps seemed to reset the server time back to what it would be (at least client side)

Issue #19761, opened 2017-05-15
[3.3.5] On entering cave in durotar you can chat in tanaris


dikkedeur:
**Description:**

There is a cave in Durotar where you when you are at a certain spot can talk in tanaris.

**Current behaviour:**

Spot in durotar makes you able to talk in tanaris

**Expected behaviour:**

Since we are in durotar we should only be able to talk in durotar, not tanaris

**Steps to reproduce the problem:**

1. go to dustwind cave in durotar

![schermafbeelding 2017-05-15 om 14 58 18](https://cloud.githubusercontent.com/assets/11842306/26058636/009c51fc-397f-11e7-91d1-db88d1630ccd.png)

2. go into the cave.. and at a certain point you see that you are in general tanaris

![schermafbeelding 2017-05-15 om 14 20 05](https://cloud.githubusercontent.com/assets/11842306/26058662/1f68d4b6-397f-11e7-8dbd-1e37abdbddc4.png)

![schermafbeelding 2017-05-15 om 14 20 14](https://cloud.githubusercontent.com/assets/11842306/26058681/2a08e82a-397f-11e7-9e9a-3e1da6b0d08c.png)

3. you are now able to speak with people in tanaris 👍 

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. e036d2232dae+ 2017-05-10 02:41:50 +0200 (3.3.5 branch) (Unix, Debug, Static)

**TDB version:**  TDB 335.63

**Operating system:** Ubuntu server 16.04

tkrokli:
TrinityCore rev. 2ff76e828f7c 2017-05-14 21:40:11 +0200 (3.3.5 branch) (Win64, Release, Static)
TDB 335.63 + updates up to and including `2017_05_14_01_world.sql`

Confirmed.

![wowscrnshot_051517_233002](https://cloud.githubusercontent.com/assets/4903600/26084179/7895555e-39da-11e7-81dc-88d45d51b4eb.jpg)

r00ty-tc:
There's several areas where this happens.

One in Winterspring where you enter AV general chat (it thinks you're in Winterax Hold) is for sure one I remember.

 I did investigate a bit once. But didn't spent a serious amount of time on it.

offl:
Can't reproduce in aba0704

illfated:
Pretty simple to re-create still. I just walked right into "discovering" Lost Rigger Cove and entering Tanaris chat:

[...]

PR #19840, opened 2017-06-02
[3.3.5] Get zone/area IDs from vmap data in the liquid update

Treeston:
Right now, every call to `GetZoneId`/`GetAreaId` (which seem innocuous enough) actually invokes a full vmaps lookup to find the zone/area id of the object. This is kind of bad.

However, as discussed in #19056, we already do liquid state updates on every server tick for everything, and that's a vmap lookup. And that vmap lookup gives us the zone/area data, we just don't use it.

And even that is kind of stupid, because why are we doing liquid updates every _tick_ if it liquid data is static? Instead, we now do liquid update whenever the WorldObject's position changes.

Then we save it on the WorldObject, and use that info to turn `GetZoneId` and `GetAreaId` into simple member accesses.

### DIFF
--- a/src/common/Collision/Management/IVMapManager.h
+++ b/src/common/Collision/Management/IVMapManager.h
@@ -22,6 +22,7 @@
 #include <string>
 #include "Define.h"
 #include "ModelIgnoreFlags.h"
+#include "Common.h"
 
 //===========================================================
 
@@ -49,6 +50,27 @@ namespace VMAP
     #define VMAP_INVALID_HEIGHT       -100000.0f            // for check
     #define VMAP_INVALID_HEIGHT_VALUE -200000.0f            // real assigned value in unknown height case
 
+    struct AreaAndLiquidData
+    {
+        struct AreaInfo
+        {
+            AreaInfo(int32 _adtId, int32 _rootId, int32 _groupId, uint32 _flags) : adtId(_adtId), rootId(_rootId), groupId(_groupId), mogpFlags(_flags) { }
+            int32 const adtId;
+            int32 const rootId;
+            int32 const groupId;
+            uint32 const mogpFlags;
+        };
+        struct LiquidInfo
+        {
+            LiquidInfo(uint32 _type, float _level) : type(_type), level(_level) { }
+            uint32 const type;
+            float const level;
+        };
+        
+        float floorZ = VMAP_INVALID_HEIGHT;
+        Optional<AreaInfo> areaInfo;
+        Optional<LiquidInfo> liquidInfo;
+    };
     //===========================================================
     class TC_COMMON_API IVMapManager
     {
@@ -102,6 +124,8 @@ namespace VMAP
             */
             virtual bool getAreaInfo(unsigned int pMapId, float x, float y, float &z, uint32 &flags, int32 &adtId, int32 &rootId, int32 &groupId) const=0;
             virtual bool GetLiquidLevel(uint32 pMapId, float x, float y, float z, uint8 ReqLiquidType, float &level, float &floor, uint32 &type) const=0;
+            // get both area + liquid data in a single vmap lookup
+            virtual void getAreaAndLiquidData(unsigned int mapId, float x, float y, float z, uint8 reqLiquidType, AreaAndLiquidData& data) const=0;
     };
 
 }
--- a/src/common/Collision/Management/VMapManager2.cpp
+++ b/src/common/Collision/Management/VMapManager2.cpp
@@ -279,6 +279,37 @@ namespace VMAP
         return false;
     }
 
+    void VMapManager2::getAreaAndLiquidData(unsigned int mapId, float x, float y, float z, uint8 reqLiquidType, AreaAndLiquidData& data) const
+    {
+        if (IsVMAPDisabledForPtr(mapId, VMAP_DISABLE_LIQUIDSTATUS))
+        {
+            data.floorZ = z;
+            int32 adtId, rootId, groupId;
+            uint32 flags;
+            if (getAreaInfo(mapId, x, y, data.floorZ, flags, adtId, rootId, groupId))
+                data.areaInfo.emplace(adtId, rootId, groupId, flags);
+            return;
+        }
+        InstanceTreeMap::const_iterator instanceTree = GetMapTree(mapId);
+        if (instanceTree != iInstanceMapTrees.end())
+        {
+            LocationInfo info;
+            Vector3 pos = convertPositionToInternalRep(x, y, z);
+            if (instanceTree->second->GetLocationInfo(pos, info))
+            {
+                data.floorZ = info.ground_Z;
+                uint32 liquidType = info.hitModel->GetLiquidType();
+                float liquidLevel;
+                if (!reqLiquidType || (GetLiquidFlagsPtr(liquidType) & reqLiquidType))
+                    if (info.hitInstance->GetLiquidLevel(pos, info, liquidLevel))
+                        data.liquidInfo.emplace(liquidType, liquidLevel);
+                
+                if (!IsVMAPDisabledForPtr(mapId, VMAP_DISABLE_AREAFLAG))
+                    data.areaInfo.emplace(info.hitInstance->adtId, info.rootId, info.hitModel->GetWmoID(), info.hitModel->GetMogpFlags());
+            }
+        }
+    }
+
     WorldModel* VMapManager2::acquireModelInstance(const std::string& basepath, const std::string& filename, uint32 flags/* Only used when creating the model */)
     {
         //! Critical section, thread safe access to iLoadedModelFiles
--- a/src/common/Collision/Management/VMapManager2.cpp
+++ b/src/common/Collision/Management/VMapManager2.cpp
@@ -287,7 +287,7 @@ namespace VMAP
             int32 adtId, rootId, groupId;
             uint32 flags;
             if (getAreaInfo(mapId, x, y, data.floorZ, flags, adtId, rootId, groupId))
-                data.areaInfo.emplace(adtId, rootId, groupId, flags);
+                data.areaInfo = boost::in_place(adtId, rootId, groupId, flags);
             return;
         }
         InstanceTreeMap::const_iterator instanceTree = GetMapTree(mapId);
@@ -302,10 +302,10 @@ namespace VMAP
                 float liquidLevel;
                 if (!reqLiquidType || (GetLiquidFlagsPtr(liquidType) & reqLiquidType))
                     if (info.hitInstance->GetLiquidLevel(pos, info, liquidLevel))
-                        data.liquidInfo.emplace(liquidType, liquidLevel);
+                        data.liquidInfo = boost::in_place(liquidType, liquidLevel);
                 
                 if (!IsVMAPDisabledForPtr(mapId, VMAP_DISABLE_AREAFLAG))
-                    data.areaInfo.emplace(info.hitInstance->adtId, info.rootId, info.hitModel->GetWmoID(), info.hitModel->GetMogpFlags());
+                    data.areaInfo = boost::in_place(info.hitInstance->adtId, info.rootId, info.hitModel->GetWmoID(), info.hitModel->GetMogpFlags());
             }
         }
     }
--- a/src/common/Collision/Management/VMapManager2.h
+++ b/src/common/Collision/Management/VMapManager2.h
@@ -118,6 +118,7 @@ namespace VMAP
 
             bool getAreaInfo(unsigned int pMapId, float x, float y, float& z, uint32& flags, int32& adtId, int32& rootId, int32& groupId) const override;
             bool GetLiquidLevel(uint32 pMapId, float x, float y, float z, uint8 reqLiquidType, float& level, float& floor, uint32& type) const override;
+            void getAreaAndLiquidData(unsigned int mapId, float x, float y, float z, uint8 reqLiquidType, AreaAndLiquidData& data) const override;
 
             WorldModel* acquireModelInstance(const std::string& basepath, const std::string& filename, uint32 flags = 0);
             void releaseModelInstance(const std::string& filename);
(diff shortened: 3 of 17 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -3357,14 +3357,33 @@ bool Unit::IsUnderWater() const
...
+{
+    PositionFullVMapData data;
+    GetMap()->getFullVMapDataForPosition(GetPositionX(), GetPositionY(), GetPositionZ(), data);
+    
+    m_zoneId = m_areaId = data.areaInfo.areaId;
+    if (AreaTableEntry const* area = sAreaTableStore.LookupEntry(m_areaId))
+        if (area->zone)
+            m_zoneId = area->zone;
+
+    /*if (GetTypeId() == TYPEID_PLAYER)
```
Treeston: Heads-up, enable this block if you're reproing to get me some useful info (also, give me a `.gps`).

ccrs: how about a timer to be able to know how much it takes?

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1074,6 +1074,12 @@ void WorldObject::_Create(ObjectGuid::LowType guidlow, HighGuid guidhigh, uint32
     m_phaseMask = phaseMask;
 }
 
+void WorldObject::AddToWorld()
+{
+    Object::AddToWorld();
+    GetMap()->GetZoneAndAreaId(m_zoneId, m_areaId, GetPositionX(), GetPositionY(), GetPositionZ());
```
Shauren: Why not `GetBaseMap()`?

src/server/game/Entities/Object/Object.h
```diff
@@ -627,6 +628,10 @@ class TC_GAME_API WorldObject : public Object, public WorldLocation
         // transports
         Transport* m_transport;
 
+        // zone/area ID updated in Unit::Update (only Units can move)
```
Shauren: False. GameObjects and DynamicObjects can move when on transport.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -3357,14 +3357,33 @@ bool Unit::IsUnderWater() const
     return GetBaseMap()->IsUnderWater(GetPositionX(), GetPositionY(), GetPositionZ());
 }
 
-void Unit::UpdateUnderwaterState(Map* m, float x, float y, float z)
+void Unit::UpdatePositionData()
+{
+    PositionFullVMapData data;
```
Shauren: FullTerrainStatus (naming - matters here because you are returning GridMap data if on flat/nowmo terrain)

Shauren (changes requested):

nawuko:
It's somewhat unrealated to this PR, but since you are currently working on it this could an intressting issue to solve: https://github.com/TrinityCore/TrinityCore/issues/19761. I did not look into this issue at all, so i'm not sure if zone/area id is even the root cause here.

Treeston:
That'd be an issue with base gridmap/vmap data, @nawuko. Not something this PR will (or aims to) address.

src/server/game/Entities/Object/Object.h
Treeston: Hm, should we do the call in their `::Update` too? Or are transports always their own area?

nawuko:
okay, just wanted to be sure its not a problem in the lookup logic. Thanks for looking into that 👍

Treeston:
I honestly can't say whether it is, @nawuko - though it seems unlikely.

src/common/Collision/Management/VMapManager2.cpp
```diff
@@ -298,19 +295,19 @@ namespace VMAP
...
-                floorZ = info.ground_Z;
-                liquidType = info.hitModel->GetLiquidType();
-                if (!reqLiquidType || (GetLiquidFlagsPtr(liquidType) & reqLiquidType))
-                    if (info.hitInstance->GetLiquidLevel(pos, info, liquidLevel))
-                        hasLiquid = true;
+                data.floorZ = info.ground_Z;
+                data.liquidInfo.type = info.hitModel->GetLiquidType();
+                if (!reqLiquidType || (GetLiquidFlagsPtr(data.liquidInfo.type) & reqLiquidType))
+                    if (info.hitInstance->GetLiquidLevel(pos, info, data.liquidInfo.level))
+                        data.hasLiquid = true;
```
Shauren: Looks like a good candidate to use Optional

Treeston: Would make it inconsistent with other vmap methods though.

Treeston:
@Shauren it is my understanding that any movement of a worldobject while it's added to a map should go through one of the methods calling `->Relocate` in `Map` - is this correct?

Raydor:
So are these changes good to go? Imo this should get more attention from shauren if you can, please. Dynamic spawn is almost done and its been in proccess for so long, dont delay it more by not getting enough attention to this pr please :(

src/common/Collision/Management/VMapManager2.cpp
Shauren: use it.

src/common/Collision/Management/VMapManager2.h
```diff
@@ -118,6 +118,7 @@ namespace VMAP
 
             bool getAreaInfo(unsigned int pMapId, float x, float y, float& z, uint32& flags, int32& adtId, int32& rootId, int32& groupId) const override;
             bool GetLiquidLevel(uint32 pMapId, float x, float y, float z, uint8 reqLiquidType, float& level, float& floor, uint32& type) const override;
+            virtual void getAreaAndLiquidData(unsigned int mapId, float x, float y, float z, uint8 reqLiquidType, AreaAndLiquidData& data) const override;
```
Shauren: virtual keyword not needed

src/common/Collision/Management/IVMapManager.h
```diff
@@ -49,6 +49,23 @@ namespace VMAP
     #define VMAP_INVALID_HEIGHT       -100000.0f            // for check
     #define VMAP_INVALID_HEIGHT_VALUE -200000.0f            // real assigned value in unknown height case
 
+    struct TC_COMMON_API AreaAndLiquidData
```
Shauren: *_API macro is only needed when your struct/class has methods implemented in .cpp file

src/server/game/Entities/Object/Object.h
```diff
@@ -627,6 +629,10 @@ class TC_GAME_API WorldObject : public Object, public WorldLocation
         // transports
         Transport* m_transport;
 
+        virtual void _ProcessPositionDataChanged(PositionFullTerrainStatus const& data);
```
Shauren: don't prefix method names with _

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -3357,14 +3356,27 @@ bool Unit::IsUnderWater() const
     return GetBaseMap()->IsUnderWater(GetPositionX(), GetPositionY(), GetPositionZ());
 }
 
-void Unit::UpdateUnderwaterState(Map* m, float x, float y, float z)
+void Unit::_ProcessPositionDataChanged(PositionFullTerrainStatus const& data)
+{
+    WorldObject::_ProcessPositionDataChanged(data);
+    
+    /*if (GetTypeId() == TYPEID_PLAYER)
```
Shauren: remove debug commented out code

src/server/game/Maps/Map.cpp
```diff
@@ -2596,6 +2600,104 @@ ZLiquidStatus Map::getLiquidStatus(float x, float y, float z, uint8 ReqLiquidTyp
...
+        data.liquidInfo.depth_level = vmapData.floorZ;
+        data.liquidInfo.entry = liquidType;
+        data.liquidInfo.type_flags = 1 << liquidFlagType;
+
+        float delta = vmapData.liquidInfo.level - z;
+        if (delta > 2.0f)
+            data.liquidStatus = LIQUID_MAP_UNDER_WATER;
+        else if (delta > 0.0f)
+            data.liquidStatus = LIQUID_MAP_IN_WATER;
+        else if (delta > 0.1f)
```
Shauren: `Map::getLiquidStatus` checks against -0.1f here

src/server/game/Maps/Map.cpp
```diff
@@ -2596,6 +2600,104 @@ ZLiquidStatus Map::getLiquidStatus(float x, float y, float z, uint8 ReqLiquidTyp
...
+            data.liquidStatus = LIQUID_MAP_WATER_WALK;
+        else
+            data.liquidStatus = LIQUID_MAP_ABOVE_WATER;
+    }
+    // look up liquid data from grid map
+    if (gmap)
```
Shauren: broken. you are overwriting vmap result. `Map::getLiquidStatus` returns early if vmap result is found

src/server/game/Maps/Map.h
```diff
@@ -155,6 +155,21 @@ struct LiquidData
     float  depth_level;
 };
 
+struct PositionFullTerrainStatus
+{
+    bool hasFullAreaInfo;
```
Shauren: pls, Optional (no has*** members)

src/server/game/Maps/Map.cpp
```diff
@@ -2596,6 +2600,104 @@ ZLiquidStatus Map::getLiquidStatus(float x, float y, float z, uint8 ReqLiquidTyp
...
+        data.areaInfo.mogpFlags = vmapData.areaInfo.flags;
+    }
+
+
+    GridMap* gmap = const_cast<Map*>(this)->GetGrid(x, y);
+    float mapHeight = gmap->getHeight(x, y);
+
+    // area lookup
+    AreaTableEntry const* areaEntry = nullptr;
+    if (vmapData.hasAreaInfo && !(z + 2.0f > mapHeight && mapHeight > vmapData.floorZ))
```
Shauren: dont bring out the negation before (), especially when you have conditions you can easily change (> to <=)

src/server/game/Maps/Map.cpp
```diff
@@ -2596,6 +2600,104 @@ ZLiquidStatus Map::getLiquidStatus(float x, float y, float z, uint8 ReqLiquidTyp
...
+    VMAP::AreaAndLiquidData vmapData;
+    vmgr->getAreaAndLiquidData(GetId(), x, y, z, reqLiquidType, vmapData);
+    if ((data.hasFullAreaInfo = vmapData.hasAreaInfo))
+    {
+        data.areaInfo.adtId = vmapData.areaInfo.adtId;
+        data.areaInfo.rootId = vmapData.areaInfo.rootId;
+        data.areaInfo.groupId = vmapData.areaInfo.groupId;
+        data.areaInfo.mogpFlags = vmapData.areaInfo.flags;
+    }
+
```
Shauren: 2 newlines

src/server/game/Maps/Map.cpp
```diff
@@ -2596,6 +2600,104 @@ ZLiquidStatus Map::getLiquidStatus(float x, float y, float z, uint8 ReqLiquidTyp
     return result;
 }
 
+void Map::getFullTerrainStatusForPosition(float x, float y, float z, PositionFullTerrainStatus& data, uint8 reqLiquidType) const
```
Shauren: method name? make it start with uppercase (you can leave the one in vmapmanager as-is)

src/server/game/Entities/Unit/Unit.h
```diff
@@ -2309,6 +2308,9 @@ class TC_GAME_API Unit : public WorldObject
 
         void DisableSpline();
 
+        void _ProcessPositionDataChanged(PositionFullTerrainStatus const& data) override;
```
Shauren: don't prefix method names with _

Shauren (changes requested):

Treeston:
Made most of the refactors you requested. Use of `Optional` needs some significant refactors.

src/server/game/Entities/Unit/Unit.h
```diff
@@ -2309,6 +2308,9 @@ class TC_GAME_API Unit : public WorldObject
 
         void DisableSpline();
 
+        void ProcessPositionDataChanged(PositionFullTerrainStatus const& data) override;
```
tkrokli:
```
src/server/game/Entities/Unit/Unit.h:2311:80: fatal error: 
      only virtual member functions can be marked 'override'
  ...const& data) override;
                  ^~~~~~~~
```

Treeston:
Travis build fail is due to outdated boost (it has 1.55, we require [1.59](https://trinitycore.atlassian.net/wiki/display/tc/Windows+Requirements)).

Dropping in place construction (`.emplace`, the Boost 1.56 feature we need) is a pretty significant performance impact, which I'm not overly willing to accept at this point in a core vmap call.

ariel-:
Ref boost::optional move semantics, not the first time this happens:
https://github.com/TrinityCore/TrinityCore/commit/4c27203c8f36dd2a5df0a4ae69fbdc4c9140b29d#diff-1bfb148491dedaab3a6311bec32f7e7eR172.

Is the performance hit that bad? IIRC POD types were just copied instead moved, you can still use boost::in_place to perform an in-place initialization

Shauren:
We use boost::in_place everywhere in packets on master branch

src/server/game/Maps/Map.h
```diff
@@ -155,6 +159,16 @@ struct LiquidData
     float  depth_level;
 };
 
+struct PositionFullTerrainStatus
+{
+    typedef VMAP::AreaAndLiquidData::AreaInfo AreaInfo;
```
Shauren: I know this is convenient but don't do this. Don't make all users of Map class also include VMapManager

Shauren (changes requested):

Shauren:
https://github.com/TrinityCore/TrinityCore/blob/c5d3dd90bea3889ef5fcd33c9ef0d59d7c544f8a/src/server/game/Handlers/AuthHandler.cpp#L36

This is the zero-overhead pattern that is being used in master branch for building packets - assigning boost::in_place() no-arg version is the same as calling .emplace() with no args
after that you just assign each field

Shauren (approved):

PR #19858, opened 2017-06-06
Scripts/Spells: Hunter spell Exhilaration

### LINKED ISSUE #19605: Core/Script The spell Exhilaration has an issue.
**Description:**

The spell Exhilaration [Spell ID: 109304] of the Hunter class, has an issue.

**Current behaviour:**

The spell Exhilaration is only healing the Hunter and not the pet aswell .

**Expected behaviour:**

Once you use the spell Exihilaration it should heal the Hunter for 30% health and the Hunter's pet for 100% maximum health.

1. Create a Hunter character.
2. Reach level 48 in order to learn the spell ''Exihilaration'' (rank 2) [Spell ID: 109304].
3. Use ''Exihilaration'' [Spell ID: 109304] while you and your pet are both below 90% health or lower.

**Branch(es):**

Master

**TC rev. hash/commit:** 

Description in wowhead http://www.wowhead.com/spell=109304/exhilaration

MartGon:
**Changes proposed:**

-  Simple fix for Exhilaration spell http://www.wowhead.com/spell=109304/exhilaration
with Exhilaration (Rank 2) http://www.wowhead.com/spell=231546/exhilaration

- [ ] 3.3.5
- [x] master

**Issues addressed:** 
Closes #19605

**Tests performed:** 
Tested in-game on the following scenarios:
- Without Rank 2 the spell only heals the hunter.
- Tested with Rank 2 spell heals both hunter and pet.
- The spell does not glitch if pet is dead or despawned.

None as far as I know.

### DIFF
--- a/src/server/scripts/Spells/spell_hunter.cpp
+++ b/src/server/scripts/Spells/spell_hunter.cpp
@@ -286,8 +286,7 @@ class spell_hun_exhilaration : public SpellScriptLoader
             void HandleOnHit()
             {
                 if (GetCaster()->HasAura(SPELL_HUNTER_EXHILARATION_R2))
-                    if (Unit* pet = GetCaster()->GetGuardianPet())
-                        GetCaster()->CastSpell(pet, SPELL_HUNTER_EXHILARATION_PET, true);
+                    GetCaster()->CastSpell((Unit*)0, SPELL_HUNTER_EXHILARATION_PET, true);
             }
 
             void Register() override
--- a/src/server/scripts/Spells/spell_hunter.cpp
+++ b/src/server/scripts/Spells/spell_hunter.cpp
@@ -37,6 +37,7 @@ enum HunterSpells
     SPELL_HUNTER_CHIMERA_SHOT_HEAL                  = 53353,
     SPELL_HUNTER_EXHILARATION                       = 109304,
     SPELL_HUNTER_EXHILARATION_PET                   = 128594,
+    SPELL_HUNTER_EXHILARATION_R2                    = 231546,
     SPELL_HUNTER_FIRE                               = 82926,
     SPELL_HUNTER_GENERIC_ENERGIZE_FOCUS             = 91954,
     SPELL_HUNTER_IMPROVED_MEND_PET                  = 24406,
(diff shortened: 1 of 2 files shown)


src/server/scripts/Spells/spell_hunter.cpp
```diff
@@ -264,6 +265,43 @@ class spell_hun_disengage : public SpellScriptLoader
...
+                ({
+                    SPELL_HUNTER_EXHILARATION_R2
+                });
+            }
+
+            void HandleOnHit()
+            {
+                if (GetCaster()->HasAura(SPELL_HUNTER_EXHILARATION_R2))
+                    if (Unit* pet = GetCaster()->GetGuardianPet())
+                        GetCaster()->CastSpell(pet, SPELL_HUNTER_EXHILARATION_PET, true);
```
joschiwald: this spell targets the pet implicitly

MartGon: Mmm I don't get what you mean. Do you mean that it should be casted by the pet?

Keader: even if casted by player, will hit pet

MartGon: So there's no need on checking if the pet exists?

MartGon:
I made the checks I did before and it works like the same. So I suppose this solution is better.

src/server/scripts/Spells/spell_hunter.cpp
```diff
@@ -264,6 +265,42 @@ class spell_hun_disengage : public SpellScriptLoader
...
+                return ValidateSpellInfo
+                ({
+                    SPELL_HUNTER_EXHILARATION_R2
+                });
+            }
+
+            void HandleOnHit()
+            {
+                if (GetCaster()->HasAura(SPELL_HUNTER_EXHILARATION_R2))
+                    GetCaster()->CastSpell((Unit*)0, SPELL_HUNTER_EXHILARATION_PET, true);
```
Keader: GetCaster()->CastSpell((Unit*)nullptr, SPELL_HUNTER_EXHILARATION_PET, true);

MartGon:
I used (Unit*)0 because it was the default value on CastCustomSpell, never used null pointer on target before.

MartGon:
With the last commit, now the code it is as Keader said, in case my last comment was confusing

Keader:
@MartGon has others examples in core: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_muru.cpp#L125 xD

MartGon:
@Keader Thanks for the example! Now I undestand what you said, it is null target because the spell already targets the pet by itself by the target field = TARGET_UNIT_PET. But what I was trying to mean is that I think the PR is ready to merge, if you like of course xD

Keader:
can be merged @joschiwald ?

Shauren (approved):

Issue #15798, opened 2015-11-01
[3.3.5] Strand of the Ancients LoS on gates


Golrag:
https://github.com/TrinityCore/TrinityCore/issues/6228 gameobject issue has been fixed but not the issue itself.

Screenshot in game: http://puu.sh/l5yKk/4b394bd373.jpg

> But it's not the GameObject, that is bugged but vmaps (I think).
> Little screenshot: http://puu.sh/kNsFY/e75f206abe.jpg

rev: https://github.com/TrinityCore/TrinityCore/commit/570e6baa090793bb5be33a4dbc417fb7b11d0b51

streetrat:
isnt that the mmaps debug output on the 3d screen?

Golrag:
Yes it is, I thought it might be useful information :)

Aokromes:
I am sure this is duplicate, but can't find the other ticket.

Golrag:
@Aokromes https://github.com/TrinityCore/TrinityCore/issues/6228 ? :p wasn't reopened so I made a new one ^^

Aokromes:
Na, there was another open iirc.

Killyana:
This one: https://github.com/TrinityCore/TrinityCore/issues/8397

Aokromes:
Eh right, the issue was on comments, unrelated to original ticket, so better leave this open.

Aokromes:
Btw Golrag jump into irc :)

Rushor:
hm what happens if we delete the whole object? still los issues?

pete318:
> hm what happens if we delete the whole object? still los issues?

Tried it, still LoS failure. Definitely not the GO itself.

Rushor:
yeah i tried this yesterday too, maybe even a blizzlike issue?

Rushor:
no not ablizzlike feature.

Aokromes:
This bug is still valid?

Keader:
yes

Warpten:
Can you try

```difff
diff --git a/src/tools/vmap4_extractor/wmo.cpp b/src/tools/vmap4_extractor/wmo.cpp
index 6318796355..201030b829 100644
--- a/src/tools/vmap4_extractor/wmo.cpp
+++ b/src/tools/vmap4_extractor/wmo.cpp
@@ -348,15 +348,17 @@ int WMOGroup::ConvertToVMAPGroupWmo(FILE *output, WMORoot *rootWMO, bool precise

[...]

PR #19863, opened 2017-06-08
Core/Maps: Correct some GetHeight functions

Golrag:
**Changes proposed:**

-  Return ground_z instead of liquid level when above water, only if ground_z > liquid level

**Target branch(es):** Both

- 3.3.5
- master

**Issues addressed:** Not a 100% fix for #15798 follow up

Tested in game

- I made this in the Test map really quick: https://puu.sh/wekrI/70683bbd24.png
As you can see wrong floorZ is present on the red marker while the blue one is correct
- A real example: https://puu.sh/wekvj/71fd07ac29.png Should be +- 394 for floorZ

- Also, I am not sure about the changes in object.cpp

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2558,7 +2558,7 @@ float WorldObject::GetFloorZ() const
 {
     if (!IsInWorld())
         return m_staticFloorZ;
-    return std::max<float>(m_staticFloorZ, GetMap()->GetGameObjectFloor(GetPhaseMask(), GetPositionX(), GetPositionY(), GetPositionZ()));
+    return std::max<float>(m_staticFloorZ, GetMap()->GetHeight(GetPositionX(), GetPositionY(), GetPositionZ()));
 }
 
 template TC_GAME_API void WorldObject::GetGameObjectListWithEntryInGrid(std::list<GameObject*>&, uint32, float) const;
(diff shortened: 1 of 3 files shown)


src/server/game/Maps/Map.cpp
```diff
@@ -2331,7 +2331,15 @@ float Map::GetWaterOrGroundLevel(uint32 phasemask, float x, float y, float z, fl
         LiquidData liquid_status;
 
         ZLiquidStatus res = GetLiquidStatus(x, y, ground_z, MAP_ALL_LIQUIDS, &liquid_status);
-        return res ? liquid_status.level : ground_z;
+		switch (res)
```
jackpoz: tabs, please edit your IDE to replace tabs with 4 spaces

Golrag: Oops my bad

jackpoz:
@Warpten please take a look at this when you can

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2558,7 +2558,7 @@ float WorldObject::GetFloorZ() const
 {
     if (!IsInWorld())
         return m_staticFloorZ;
-    return std::max<float>(m_staticFloorZ, GetMap()->GetGameObjectFloor(GetPhaseMask(), GetPositionX(), GetPositionY(), GetPositionZ()));
+    return std::max<float>(m_staticFloorZ, GetMap()->GetHeight(GetPositionX(), GetPositionY(), GetPositionZ()));
```
Shauren: Why? m_staticFloorZ should be exactly equal to Map::GetHeight (the call without checking _dynamicTree)

Golrag: Ooh, yea I understand now, will revert it

PR #19870, opened 2017-06-10
Modify group invite behavior while creating group to be blizzlike

### LINKED ISSUE #17258: [3.3.5] Group creation behavior inconsistent with retail
3.3.5 e3aa192

On current TC, a group is created when the first player accepts the group invite. This means you cannot invite additional people before the first member accepts the group, and you also cannot "cancel" a group invite by leaving the group early (using `/run LeaveParty()`).

On retail, a group is created when the player first sends a group invite. You can begin inviting additional members immediately, and also leave this not-yet-visible group using `/run LeaveParty()` or similar.

Treeston:
Modify group invite behavior while creating group to be blizzlike:
- Leader can invite multiple people before the first invite is accepted
- Leader can cancel group formation by sending CMSG_GROUP_DISBAND (using /run LeaveParty() or similar)

Fixes #17258.

### DIFF
--- a/src/server/game/Handlers/GroupHandler.cpp
+++ b/src/server/game/Handlers/GroupHandler.cpp
@@ -74,68 +74,71 @@ void WorldSession::HandleGroupInviteOpcode(WorldPacket& recvData)
         return;
     }
 
-    Player* player = ObjectAccessor::FindPlayerByName(membername);
+    Player* invitingPlayer = GetPlayer();
+    Player* invitedPlayer = ObjectAccessor::FindPlayerByName(membername);
 
     // no player
-    if (!player)
+    if (!invitedPlayer)
     {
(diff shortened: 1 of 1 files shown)


aje:
1. log in 3 players A, B and C
2. player A invites players B and C to join his group (the two invitations are launched before B or C answer)
3. player B (respectively C) decline the invitation
4. player C (respectively B) accepts the invitation
5. **Current behaviour:** Nobody is in a group. **Expected behaviour:** Player A should be in group with player C (respectively B)

**Branche:**
TrinityCore rev. 4b6351e6a5a2+ 2017-06-15 04:19:28 +0200 (3.3.5 branch) (Win64, RelWithDebInfo, Static) (worldserver-daemon)

PR #19893, opened 2017-06-15
[3.3.5] Implement Pet combo point system

Treeston:
* Implement Pet combo point system
* Implement a hack to add combo points and make Wolverine Bite useful (fixes #752).
* Fix an issue where players could use client hacks to use combo points on a different target from the one they're applied to.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -2106,6 +2106,12 @@ void Unit::AttackerStateUpdate(Unit* victim, WeaponAttackType attType, bool extr
         DamageInfo dmgInfo(damageInfo);
         ProcSkillsAndAuras(damageInfo.target, damageInfo.procAttacker, damageInfo.procVictim, PROC_SPELL_TYPE_NONE, PROC_SPELL_PHASE_NONE, dmgInfo.GetHitMask(), nullptr, &dmgInfo, nullptr);
 
+        if (dmgInfo.GetHitMask() & PROC_HIT_CRITICAL) // @todo is there a better place to put this hack? there doesn't seem to be any aura applied on talent
+            if (Pet* pet = ToPet())
+                if (pet->HasSpell(53508)) // Wolverine Bite
+                    pet->AddComboPoints(damageInfo.target, 1);
+
+
         if (GetTypeId() == TYPEID_PLAYER)
             TC_LOG_DEBUG("entities.unit", "AttackerStateUpdate: (Player) %u attacked %u (TypeId: %u) for %u dmg, absorbed %u, blocked %u, resisted %u.",
                 GetGUID().GetCounter(), victim->GetGUID().GetCounter(), victim->GetTypeId(), damageInfo.damage, damageInfo.absorb, damageInfo.blocked_amount, damageInfo.resist);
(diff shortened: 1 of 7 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -2106,6 +2106,12 @@ void Unit::AttackerStateUpdate(Unit* victim, WeaponAttackType attType, bool extr
         DamageInfo dmgInfo(damageInfo);
         ProcSkillsAndAuras(damageInfo.target, damageInfo.procAttacker, damageInfo.procVictim, PROC_SPELL_TYPE_NONE, PROC_SPELL_PHASE_NONE, dmgInfo.GetHitMask(), nullptr, &dmgInfo, nullptr);
 
+        if (dmgInfo.GetHitMask() & PROC_HIT_CRITICAL) // @todo is there a better place to put this hack? there doesn't seem to be any aura applied on talent
+            if (Pet* pet = ToPet())
+                if (pet->HasSpell(53508)) // Wolverine Bite
+                    pet->AddComboPoints(damageInfo.target, 1);
```
Treeston: This segment is obviously a hack. However, I didn't find any auras applied by the talent - in fact, I didn't even find any global pet auras that proc on white crit - despite my best efforts.

@Shauren any ideas on how to best handle this? Leaving it here feels wrong.

ccrs: script hook and playerscript?

ariel-: Unit::ProcSkillsAndReactives seems a better place for this

ariel-:
What about making combo point related methods virtual? so only one call instead of branching

Treeston:
They're only implemented by `Pet` and `Player`. Not sure what's cheaper here - @Shauren, insights?

src/server/game/Entities/Unit/Unit.cpp
ariel-: Also, check for special attribute in Wolverine Bite (IIRC had one)

Treeston: It has `SPELL_ATTR4_TRIGGER_ACTIVATE` and `SPELL_ATTR1_UNK27`, no idea if either of those is interesting.

ariel-: SPELL_ATTR4_TRIGGER_ACTIVATE check that one :)

Treeston: `TRIGGER_ACTIVATE` is a client side flag that interacts with other flags (in this case `REQ_COMBO_POINTS1`) to make the icon gray out if not available. No server handling for these spells.

NB: Most other spells with this attribute have proc info (f.ex. Art of War).

Shauren:
In this case? Whatever produces cleaner code

src/server/game/Entities/Unit/Unit.cpp
Shauren: Lets see, a spell using a single combo point to activate, does not scale from combo points...
Yep, thats a reactive, just like warrior Overpower - handle them together in Unit::ProcSkillsAndReactives

ariel-: I recall researching for the WB issue, I think SMSG_PET_UPDATE_COMBO_POINTS was sent even when the pet didn't have the spell, so a mere AddComboPoints on every crit should do in that case

Treeston:
OK, I'll extend this to cover a full CP system rewrite then.

Treeston:
PR superseded by #19914.

PR #19895, opened 2017-06-15
Detect and punish client-side fall delaying

Treeston:
Cross-check client-submitted fall time with server-measured time. When excessive difference is detected, punish player.


P-Kito:
Forgot the issue, thanks for catching up the issue

PR #19904, opened 2017-06-17
Core/Quest: fix order of RewardChoiceItems and reputation reward.

Weyrath:
**Changes proposed:**

The current quest reward order is a bit off: reputation is rewarded before the SMSG_QUESTGIVER_QUEST_COMPLETE is sent, and RewardChoiceItems are given before RewardItems.

As can be seen on [this](https://youtu.be/-CqiRz4i-lo?t=468) video from retail (just as an example - every retail video from WotLK shows the same behaviour), reputation is given after the other rewards. Moreover, when the author of the video [opens his bags](https://youtu.be/-CqiRz4i-lo?t=472), it can be seen that the three healing potions (which are the non-choice reward) was given before the three elixirs (which are the choice reward).
The client also lists the rewards in that order (non-choice -> choice) in the chat window.

**Target branch(es):**

- 3.3.5

**Tests performed:**

- Builds and works.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -15001,19 +15001,6 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver, 
 
     RemoveTimedQuest(quest_id);
 
-    if (quest->GetRewChoiceItemsCount() > 0)
-    {
-        if (uint32 itemId = quest->RewardChoiceItemId[reward])
-        {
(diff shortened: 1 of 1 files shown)


Aokromes:
Conflicting files

Weyrath:
Fixed, thanks.

Weyrath:
Would someone mind reviewing this? Hasn't been outdated since the last commit, and it's a small enough change.

Thanks :+1:

PR #19906, opened 2017-06-18
Proper client timeout detection

Treeston:
Properly detect client timeout when logged into a character after a configurable time (default 10s) has passed without the client sending any packets.

Fixes issues with crashed clients leaving characters in the world for a very long time (default 15 minutes), as well as edge case exploits involving intentionally pausing client execution for some amount of time.


jackpoz:
10s might be quite low, like in cases of Windows Updates randomly starting and using the whole bandwidth. Maybe 1 minute is high enough anyway ?

Treeston:
You'd need to not send _any_ packets for 10s to kill the connection - that's not really something that's gonna happen just because some other program is using bandwidth.

jackpoz:
It just has to happen once to trigger the disconnect, where now any client can have any kind of network lag spike without causing the disconnect (wifi signal degradation, mobile data connection with low performance, network saturation).
How often is the client sending a packet for sure anyway ? is there any operation on the computer that might trigger the disconnect (map loading on slow computer + other background unrelated processes, etc) ?

Treeston:
Yes, but a client _shouldn't_ be able to stay connected for prolonged periods of time without data - the WoW client/server model simply isn't designed to handle it gracefully.

Keeping a player connected in these scenarios will make them visually teleport for others (impossible to fight against in PvP), as well as allowing them to clip through locked doors (as they aren't being sent in time), ignore NPC aggro (by moving from one side of the aggro range to the other before the next movement packet is sent) and many many more. It's simply not sustainable while maintaining a consistent gameplay environment (which is why the retail servers would act like this PR does).

Shauren:
And 10s is too short (thats both CMSG_PING and CMSG_KEEPALIVE are on 30s timers)

Treeston:
40s then?

Shauren:
1 minute suggested earlier by jackpoz seems like a good idea to me

src/server/game/Server/WorldSocket.cpp
```diff
@@ -358,6 +358,8 @@ WorldSocket::ReadDataHandlerResult WorldSocket::ReadDataHandler()
         }
         case CMSG_KEEP_ALIVE:
             LogOpcodeText(opcode, sessionGuard);
+            if (_worldSession)
```
Shauren: This is very bad - you are accessing _worldSession while sessionGuard is not locked

Shauren (changes requested):

src/server/game/Server/WorldSocket.cpp
Treeston: The only modification being made is to a `std::atomic` though. Still an issue?

Shauren: its the session pointer in WorldSocket that needs to be protected, not any member of WorldSession

jackpoz: if a race condition can happen here on _worldSession then its value could become NULL after the NULL-check, causing a NULL dereference crash.
if that's the case then it's just a matter of calling sessionGuard.lock() and then sessionGuard.unlock() (or whatever is the syntax of this lock) in the CMSG_KEEP_ALIVE case ?

Treeston: Makes sense, thanks @Shauren - I'll adjust this to lock/unlock.

Shauren: @jackpoz the type of sessionGuard is unique_lock (constructed with deferred lock flag), there is no need to unlock it

Treeston:
@Shauren any other feedback or can this be merged?

Shauren (approved):

Issue #16542, opened 2016-02-10
[3.3.5][6.x] Core/Vehicle: Vehicle accessorys threat


Killyana:
Threat generated by accessory must be transferred to the vehicles, but actually all accessory generate 0 threat, example Salvaged Siege Engine http://www.wowhead.com/npc=33060 use Salvaged Siege Turret http://www.wowhead.com/npc=33067 and the vehicle is supposed to be used as tank but as the turret threat is ignored it cannot be
Rev: d67e87733b2

sirikfoll:
Kinda related to this issue, threat system needs to be reworked for vehicles, probably not valid for all vehicles, but a lot of them aren't supposed to even have a threatList, cannons are one example.
Antipersonnel Cannon(27894) in Strand of the Ancients often bug players in combat, problem which wouldn't exist if the cannon don't have a ThreatList.

Rushor:
as already said in https://github.com/TrinityCore/TrinityCore/issues/15591

> hmm just found this https://github.com/TrinityCore/TrinityCore/blob/d5fb18d3c3a557acaf60a39be4ad5444969dd663/src/server/game/Entities/Unit/Unit.cpp#L10657

>we can add a flagextra for alle creatures which should not generate any thread

Issue #15591, opened 2015-09-24
Core/Vehicles: combat bug with mounted NPCs


Killyana:
Aggro any npc mounted on a vehicle, then run away, the player will remain in combat until reconnection or he dies. (chance to reproduce 100%)
Example of Npc: .go crea 96359
Ps: Once you're stacked in combat switch in gm mode than gm off to clean the combat mode, than aggro any other npc and run away, you will be in combat mode again.
TrinityCore rev. 2eafbe6bf089

jackpoz:
so the npc never evades ?

Killyana:
The vehicle evade but not the accessory.

Rushor:
same bug exists, if a player mounts on the vehicle. any other player which attacks the vehicle will remain in combat.

Killyana:
For this issue may adding an evade action for the accessories one the vehicle evade can fix the problem.

Rushor:
hmm just found this https://github.com/TrinityCore/TrinityCore/blob/d5fb18d3c3a557acaf60a39be4ad5444969dd663/src/server/game/Entities/Unit/Unit.cpp#L10657

we can add a flagextra for alle creatures which should not generate any thread

sirikfoll:
Maybe is related to this, sometimes MKII(Mimiron's vehicle), after a wipe, remains with threat in a player who died, release and is outside the instance. The instance gets stuck in combat

Aokromes:
Very likely it's unrelated, raid stuck on combat  with players after wipe appears on a lot of bosses without vehicles.

Kittnz:
Anyone tried to debug these stuck combat in a raid?

sirikfoll:
These mimiron's case, i know how to reproduce. With two players,  leave player1 inside the instance, like at the entrance. A second player start the fight and, as soon as the boss mount the vehicle, second player kill himself. The boss will remain with aggro with the player1, player1 can now leave the instance and will still have the boss aggro.

velinath:
Is this related to Argent Tournament NPC jousting opponent evade issues? I don't want to open a new issue if I'm just looking at another specific case of the general problem.

Killyana:
@Treeston this issue is not fixed by Combat/threat system rewrite.

Rushor:
anyone else encounter this issue:
[...]

Issue #13103, opened 2014-09-11
[335] PvP Flag - Manual toggle does not work properly (PvE Realm)


MrSmite:
**Core:** 0392bf9bdfaed726a1ef473bb477b338765f4806
**DB:** N/A
### Bug:

On a PVE realm (realm type 6) when a player types _/PVP_ to turn on their flag, it is not properly tracked across logout or zone changes from hearthing. The core doesn't handle it properly when the player is spawned into the world.

Finally, if you type /PVP to turn _off_ your flag, when the timer runs out it does not remove the player's PvP status.
### Details:

**_Example 1_**
- Log in and go to a non-sanctuary zone (such as Goldshire)
- Type /PVP to turn on your flag
- Log out
- Log in

At this point the player will NOT have the PVP flag on their portrait **and** _IsPvP()_ returns FALSE however typing _/PVP_ thinking it will turn the flag on displays the message:

> You will be unflagged for PvP in five minutes...

**_Example 2_**
- Type /PVP to turn on your flag
- Use your Hearthstone

At this point the player will NOT have the PVP flag on their portrait and  typing _/PVP_ thinking it will turn the flag on displays the message:

> You will be unflagged for PvP in five minutes...

**_Example 3_**
- Type /PVP to turn on your flag
- Type /PVP to turn off your flag

Wait until the timer runs out (don't hearth or change zones). Notice you are still flagged for PvP until you hearth or change zones.
### Conclusion

The core doesn't seem to be properly tracking the PvP status. In some cases, _IsPvP()_ it thinks it's disabled but in terms of _/PVP_ it thinks it's enabled. Other cases it simply doesn't remove PVP status when it should.
### Notes:
- I honestly can't remember if it should be disabled at logout or not. I don't think it should unless logged out longer than 5 minutes but I can't be sure.
- I don't know about other realm types but I don't think this is realm specific, it seems more like a generic PvP issue.
- I haven't tested on normal zone changes (eg: loading screens into / out of the Deeprun Tram)
##

Rushor:
so probably we can try to find out first, why typing /pvp after relog returns the error-message. there must be something which saves the pvp-status after login and logout :/

Rushor:
i think the problem is somewhere here in the toggleaction:
https://github.com/Rushor/TrinityCore/blob/3.3.5/src/server/game/Handlers/MiscHandler.cpp#L467

Frytiks:
Confirmed.

[...]

PR #19930, opened 2017-06-20
[3.3.5] Combat/threat system rewrite

Treeston:
_(When reviewing this PR, start reading with [CombatManager.h](https://github.com/Treeston/TrinityCore/blob/3.3.5-combatrewrite/src/server/game/Combat/CombatManager.h) and [ThreatManager.h](https://github.com/Treeston/TrinityCore/blob/3.3.5-combatrewrite/src/server/game/Combat/ThreatManager.h). Everything else will be easier to understand after that.)_

Our current combat system is kind of a mess. Combat is not necessarily mutual, `IsInCombatWith` only works for units with a threat list, and so on and so forth. I'm ripping it all out and replacing it.

**Improvements:**
* PvE combat is now always mutual
* PvP combat is now also tracked, and almost always mutual. That means we can easily determine a list of players we're fighting.
* `IsInCombatWith` is now reliable
* Blizzlike engage behavior with projectile spells (combat on launch, threat on impact)
* Threat system is no longer a steaming pile of hacks
* Scripts can `AddThreat` without misdirection fucking them up (lol)
* Vehicle aggro now properly goes to the "main" vehicle body (tagging #16542)
* And various edge case behaviors that I didn't even notice

**Todo:**
* Bug hunting


src/server/game/Combat/CombatManager.h
```diff
@@ -0,0 +1,86 @@
...
+        void EndAllCombat();
+
+        void Update(uint32 tdiff);
+
+        CombatManager(CombatManager const&) = delete;
+        CombatManager& operator=(CombatManager const&) = delete;
+
+    protected:
+        void putReference(ObjectGuid const& guid, CombatReference* ref);
+        // NB: does not actually deallocate the ref - that's done in ref::EndCombat (this should never be called from anywhere else)
```
Shauren: NB? What is that

sirikfoll: Probably stands for "Nota bene", like "pay attention"..

Treeston: what @sirikfoll said

src/server/game/Combat/NewThreatManager.h
```diff
@@ -0,0 +1,144 @@
...
+    private:
+        Unit* const _owner;
+        bool _ownerEngaged;
+        ThreatReference const* _currentVictimRef;
+        uint32 _updateClientTimer;
+
+        ThreatReference const* reselectVictim();
+
+        // send opcodes (all for my own threat list)
+        void sendClearAllThreatToClients() const;
```
ariel-: Codestyle, all member functions must be UpperCamelCase

src/server/game/Combat/NewThreatManager.cpp
```diff
@@ -0,0 +1,394 @@
...
+    if (it == _myThreatListEntries.end())
+        return;
+    ThreatReference* ref = it->second;
+    _myThreatListEntries.erase(it);
+
+    if (_currentVictimRef == ref)
+        _currentVictimRef = nullptr;
+    
+    // start at back since that's what ClearThreat does (this makes ClearThreat take linear instead of quadratic time)
+    for (size_t i = _sortedThreatList.size(); (i--) > 0;)
```
ariel-: If it's sorted can't you use binary search? then swap with end and pop_back.

Treeston: It's not sorted until it needs to be (which isn't now). Swap and pop is reasonable though.

Treeston:
Small aside: My heap logic in `ThreatManager::reselectTarget` is wrong in this version, and might terminate early in edge cases. Too tired to rewrite it now, will do tomorrow.

src/server/game/Combat/CombatManager.cpp
```diff
@@ -0,0 +1,177 @@
...
+    while (!_pvpRefs.empty())
+        _pvpRefs.begin()->second->EndCombat();
+}
+
+void CombatManager::Update(uint32 tdiff)
+{
+    auto it = _pvpRefs.begin(), end = _pvpRefs.end();
+    while (it != end)
+    {
+        PvPCombatReference* const ref = reinterpret_cast<PvPCombatReference*>(it->second);
```
Shauren: why the cast? your map already stores this type

src/server/game/Combat/CombatManager.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+void CombatManager::putReference(ObjectGuid const& guid, CombatReference* ref)
+{
+    if (_owner->IsControlledByPlayer())
+        std::cout << _owner->GetName() << " put " << guid.GetRawValue() << " (" << ref->_isPvP << ")" << std::endl;
+    if (ref->_isPvP)
+    {
+        auto& inMap = _pvpRefs[guid];
+        ASSERT(!inMap && "Duplicate combat state detected - memory leak!");
+        inMap = reinterpret_cast<PvPCombatReference*>(ref);
```
Shauren: use static_cast, not reinterpret_cast

src/server/game/Combat/CombatManager.cpp
Treeston: Leftover from a previous implementation.

src/server/game/AI/ScriptedAI/ScriptedCreature.h
```diff
@@ -210,11 +210,16 @@ struct TC_GAME_API ScriptedAI : public CreatureAI
     //Plays a sound to all nearby players
     void DoPlaySoundToSet(WorldObject* source, uint32 soundId);
 
-    //Drops all threat to 0%. Does not remove players from the threat list
-    void DoResetThreat();
-
-    float DoGetThreat(Unit* unit);
-    void DoModifyThreatPercent(Unit* unit, int32 pct);
+    // Add specified amount of threat directly to victim (ignores redirection effects) - also puts victim in combat and engages them if necessary
+    void AddThreat(Unit* victim, float amount, Unit* who = nullptr) { (who ? who : me)->AddThreat(victim, amount, nullptr, true, true); }
```
Shauren: Move all of them to source file. This header is not supposed to require Unit/Creature definitions (read the comment on top for Creature.h include)

src/server/game/Combat/ThreatManager.cpp
```diff
@@ -1,593 +1,581 @@
...
- *
- * This program is distributed in the hope that it will be useful, but WITHOUT
- * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
- * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
- * more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program. If not, see <http://www.gnu.org/licenses/>.
- */
+* Copyright (C) 2017-2017 TrinityCore <http://www.trinitycore.org/>
```
Aokromes: aling with /* :)

src/server/game/Combat/ThreatManager.h
```diff
@@ -1,6 +1,5 @@
 /*
- * Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
- * Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
+ * Copyright (C) 2017-2017 TrinityCore <http://www.trinitycore.org/>
```
Aokromes: 2008- :)

Killyana:
As you're working on this, can you take a look to this issue https://github.com/TrinityCore/TrinityCore/issues/16542  ?

Treeston:
Does this apply to _all_ auras with `RIDE_VEHICLE`, @Killyana?

src/server/game/Combat/ThreatManager.h
Treeston: Isn't it the year of the file being created?

src/server/game/AI/ScriptedAI/ScriptedCreature.h
Treeston: I'm not sure what the benefit would be. Are we intending to remove the `Creature.h` include?

If not for that, this is the perfect candidate for inlining (so we save call overhead since compiler is just gonna inline it).

Killyana:
Yes, you can test with https://woehead.way-of-elendil.fr/?npc=33060
Also: https://github.com/TrinityCore/TrinityCore/issues/15591

Treeston:
So, while riding a vehicle, all threat caused should go the vehicle? We can do that fairly easily using the redirect system.

Killyana:
And Accessories must evade at the same time as the vehicle (for vehicles not controlled by players)

src/server/game/Combat/CombatManager.cpp
```diff
@@ -1,19 +1,19 @@
...
-*
-* This program is distributed in the hope that it will be useful, but WITHOUT
-* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
-* FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
-* more details.
-*
-* You should have received a copy of the GNU General Public License along
-* with this program. If not, see <http://www.gnu.org/licenses/>.
-*/
+ * Copyright (C) 2017-2017 TrinityCore <http://www.trinitycore.org/>
```
Raydor: One more 2017-2017 remaining :P

src/server/game/AI/ScriptedAI/ScriptedCreature.h
jackpoz: lul it's 2017 and people still take sourcecode-related decisions because of inlining

src/server/scripts/EasternKingdoms/Karazhan/boss_prince_malchezaar.cpp
```diff
@@ -455,9 +454,7 @@ class boss_malchezaar : public CreatureScript
                             if (target)
                             {
                                 axe->AI()->AttackStart(target);
-                                //axe->getThreatManager().tauntApply(target); //Taunt Apply and fade out does not work properly
-                                                                // So we'll use a hack to add a lot of threat to our target
-                                axe->AddThreat(target, 10000000.0f);
+                                AddThreat(target, 10000000.0f, axe);
```
ariel-: Since the original comment was removed, does the value come from sniffs?

Treeston: None of the changes in these very old hack scripts are sniff sourced (or even tested, for that matter).

src/server/game/AI/ScriptedAI/ScriptedCreature.h
Treeston: lul it's 2017 and compilers still can't generally inline definitions that aren't available in the current compilation unit

If I don't have to incur jump overhead every time a simple helper is called, I see no reason to do it.

jackpoz: by now we should not worry at all about "jump overhead" (just like we don't worry about "if" overhead or virtual/override overhead for the matter).

Shauren: If you really care about inlining then build with LTO

Treeston: Alright, I cede that - however, I have to admit I still don't understand the thought process behind not wanting to "have to" include `Creature.h` in `ScriptedCreature.h`.
What use case do you see where a script would include the latter without also wanting to include the former? Have you ever seen a script that doesn't use any of `Creature`'s defined methods directly?

If the answer to that is "no" (and I strongly believe it to be), I don't see any harm in leaving these methods as inline definitions - though the comment on the include should be amended in that case, of course.

Shauren: The general idea is to make each header only include things it needs for itself - all kind of convenience helpers don't belong in this category and should be moved to .cpp file

Treeston:
there you go @Killyana

Treeston:
And with this, it does everything that the old one did (except better). Review + testing, please!

Treeston:
NB: Auth database update queries (RBAC renames) currently not included to avoid conflicts.

ccrs:
some sort of guide to know the new behaviour would be necessary here

is AI AttackStart still the main method to start combat with the target?
whats the purpose of engage logic? differencies with isincombat(with)?
whats the behaviour when addthreat is used in combat and out of combat?
are encounters that use a trigger as combat controller (examples valithria or halion) tested?
is trigger combat handled correctly? if a trigger that owns a damaging aura enters combat and then evades the aura will be lost due to enterevademode (example slime triggers on rotface)
do charms lose combat status correctly when all attackers evade/die?

src/server/game/Combat/ThreatManager.h
```diff
@@ -15,284 +14,154 @@
...
-        bool iAccessible;
+    friend class ThreatManager;
+    friend struct CompareThreatLessThan;
 };
 
-//==============================================================
-class ThreatManager;
-
-class TC_GAME_API ThreatContainer
+struct CompareThreatLessThan : std::binary_function<ThreatReference*, ThreatReference*, bool>
```
Shauren: std::unary/binary_function structs are deprecated since C++11 and will be removed in C++17

src/server/game/Combat/ThreatManager.h
```diff
@@ -15,284 +14,154 @@
...
+        void UnregisterRedirectThreat(uint32 spellId);
+        void UnregisterRedirectThreat(uint32 spellId, ObjectGuid const& victim);
+        void ForwardThreatForAssistingMe(Unit* assistant, float baseAmount, SpellInfo const* spell = nullptr, bool ignoreModifiers = false);
     private:
-        void _addThreat(Unit* victim, float threat);
+        Unit* const _owner;
+        bool _ownerEngaged;
+
+        // send opcodes (all for my own threat list)
+        void sendClearAllThreatToClients() const;
```
Shauren: func naming

src/server/game/Combat/ThreatManager.h
```diff
@@ -15,284 +14,154 @@
...
+        float GetThreat(Unit const* who, bool includeOffline = false) const;
+        std::vector<ThreatReference*> const& GetThreatList() const { return _sortedThreatList; } // NB: Not necessarily structured in any way unless you call HeapifyThreatList()/SortHeapifiedThreatList() first!
+
+        bool IsThreateningAnyone(bool includeOffline = false) const;
+        std::unordered_map<ObjectGuid, ThreatReference*> const& GetThreatenedByMeList() const { return _threatenedByMe; }
+
+        void UpdateOnlineStates(bool meThreateningOthers = true, bool othersThreateningMe = true);
+        ///== AFFECT MY THREAT LIST ==
+        void HeapifyThreatList();
+        void SortHeapifiedThreatList();
```
Shauren: these 2 functions seem like a good place to use them incorrectly - have you looked at boost.heap (d_ary_heap, backed by vector specifically)?

It provides two interfaces - regular .begin/.end (raw begin/end on heapified vector) and .ordered_begin/.ordered_end (iterates in sorted order)
Creating .ordered_end() isn't free like a vector .end() so you could expose this to scripts using our Trinity::Containers::IteratorPair

Shauren (changes requested):
This is not a complete review btw

src/server/game/Combat/ThreatManager.h
Treeston: what's the suggested replacement?

Treeston:
@ccrs mhm, I'll do something a bit down the line.

> is AI AttackStart still the main method to start combat with the target?

`EngageWithTarget` should be used. (though yes, `AttackStart` should still work, as always).

> whats the purpose of engage logic? differencies with isincombat(with)?

The mob can be in combat before it engages the target. For example, if I shoot a shadowbolt at a kobold, the kobold enters combat the moment my cast time finishes, but doesn't engage until my shadowbolt smacks him in the face (threat list not empty).

> whats the behaviour when addthreat is used in combat and out of combat?

Identical. Out of combat it also starts combat (threat implies combat).

> are encounters that use a trigger as combat controller (examples valithria or halion) tested?

Don't see any way they could've broken, but I can check.

> is trigger combat handled correctly? if a trigger that owns a damaging aura enters combat and then evades the aura will be lost due to enterevademode (example slime triggers on rotface)

I can test it, though I see no reason why it should be broken. Got an example?

> do charms lose combat status correctly when all attackers evade/die?

Should involve a faction change, which _should_ call that method that stops combat with a faction? I haven't touched that logic, so I assume it still works. Can test when I get around to it.

* > logic behind threat list heapify? when is it used? why is the list not sorted all/almost all times? (previous logic sorted when the list was set to dirty)

Sorting is kinda expensive, heapification is cheaper. 99% of the time we don't need the sorted list (and in the cases where we do, we end up just doing a full heap sort anyway through repeated `pop` calls).
```

File: src/server/game/Combat/ThreatManager.h (hunk 13 above)
Comment (Treeston) [inline src/server/game/Combat/ThreatManager.h]: Interesting, `boost::heap` provides mutability interfaces (which were lacking in `std::priority_queue`). Will look into it, could be a very nice improvement (fibonacci heaps give us constant complexity on increasing element priority, which is by far the most common - threat being added).

File: src/server/game/Combat/ThreatManager.h (hunk 12 above)
Comment (Treeston) [inline src/server/game/Combat/ThreatManager.h]: elaborate?

Comment (Shauren) [inline src/server/game/Combat/ThreatManager.h]: should not start with lowercase

Comment (Treeston) [inline src/server/game/Combat/ThreatManager.h]: Do we have any way to distinguish between private/public member functions in CS?
In this PR, lowercase = private, uppercase = public, though I can obviously refactor that if required (just hearing conflicting versions of "CS" here).

Comment (Shauren) [inline src/server/game/Combat/ThreatManager.h]: No, all functions follow the same convention

**Comment by ccrs** [comment]
````quote
Sorting is kinda expensive, heapification is cheaper. 99% of the time we don't need the sorted list (and in the cases where we do, we end up just doing a full heap sort anyway through repeated pop calls).
````
now talking without knowing the exact behaviour, what about threat packets? are they sent as a list of values to the clients on threat change on update? if they are constantly sent it means the list is constantly sent ordered or client addons that show aggro do order those values?

File: src/server/game/Combat/ThreatManager.h (hunk 11 above)
Comment (ariel-) [inline src/server/game/Combat/ThreatManager.h]: Just remove the inheritance

**Comment by BioGram** [comment]
This exploit may related to this system:

While instance/boss is in combat nightelf's can leave combat by using shadowmeld spell and relogin with another wow then its combat will gone

**Comment by jackpoz** [comment]
@Treeston I was wondering if it made sense to merge the rename commits (that don't change any functionality) before all the other refactoring to ease the PR review. Just a thought :)

**Comment by Treeston** [comment]
@ccrs sent in no particular order - addons handle sorting themselves.
@jackpoz could be an option, yes

**Comment by Treeston** [comment]
@jackpoz I thought about this some more - what changes would you be thinking about when you say "rename commits" - can't think of any changes that "just" rename stuff.

**Comment by Shauren** [comment]
getThreatManager() -> GetThreatManager()
IsInCombat() -> IsEngaged() (i know this is not a rename but you can temporarily add IsEngaged function which will do nothing except return IsInCombat until this is merged)

**Comment by Treeston** [comment]
Hm, I suppose. I'll look into it once I finish heap rewrite.

**Comment by Treeston** [comment]
Refactors/compat layer is now in #19966.

**Comment by Treeston** [comment]
Rebased on latest core. Now that #19966 is merged, work will resume here.

**Comment by jackpoz** [comment]
nice, only 44 files modified

File: src/server/game/Combat/CombatManager.cpp [hunk 14]
```diff
@@ -17,6 +17,33 @@
...
+    // ...the two units need to be in the world
+    if (!a->IsInWorld() || !b->IsInWorld())
+        return false;
+    // ...the two units need to both be alive
+    if (!a->IsAlive() || !b->IsAlive())
+        return false;
+    // ...the two units need to be on the same map
+    if (a->GetMap() != b->GetMap())
+        return false;
+    // ...the two units need to be in the same phase
```
ccrs: algalon, halion, etc?

Treeston: At least Halion uses a controller dummy (that's in both phases) to keep players in combat. I'll need to check Algalon.

ccrs: Sartharion too (portal made by first miniboss.. Vesperon?), valithria adds attacking someone that enters the portal?

dragons do like phases uh?

Treeston: Apparently, yes. Vali adds should continue to attack you while you're not in their phase? That sounds like a bug to me (would be completely unique behavior).

ccrs: I was thinking about they should keep you referenced but cannot actually attack like the actual canaiattack

Treeston: That'd be a change from current behavior, though (and would open a potential huge can of worms with phase transitions and stuck in combat bugs).

Treeston:
Rebased to incorporate 2739a5c and 62d5113 from base.

Treeston:
Tested Algalon - everything seems to be working as expected:
* If everybody phases out, Algalon will reset.
* If any given raid member phases out, their threat is offlined, but maintained. When they re-enter the main phase, the threat still exists.
* Since the threat is only offlined, combat state between Algalon and players is maintained while the players are phased.

src/server/game/Combat/CombatManager.cpp
Treeston: @ccrs also keep in mind that this check is only called when the combat initially starts - it's mostly a safeguard against scripts doing dumb things. Once combat is started, it'll persist even if the player leaves the phase.

Killyana:
I noticed that if a player get CM like during Lady Deathwhisper encounter, all the threat generated by this player will be reset, I don't know if this is the correct behavior.

Treeston:
@Killyana Definitely intended within the context of this PR - friendlies cannot be in combat with each other.

src/server/game/Combat/CombatManager.cpp
ccrs: > If any given raid member phases out, their threat is offlined, but maintained. When they re-enter the main phase, the threat still exists.
> Since the threat is only offlined, combat state between Algalon and players is maintained while the players are phased.

Thats exactly what I meant 😄

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12559,6 +12557,8 @@ void Unit::RemoveCharmedBy(Unit* charmer)
         }
         player->SetClientControl(this, true);
     }
+    
+    EngageWithTarget(charmer);
```
Shauren: What if charming the player made the creature enter evade mode?

Shauren: But that was my point - on retail if a creature/boss charms a player who is the only entry on the threat list, it evades.

Keader: ^ true, i saw it alot in Lady Deathwhisper

Treeston: It _should_ still do that. The charm isn't removed until _after_ the evade happens - at this point the mob is in evade mode, and `EngageWithTarget` will fizzle.

I'll make a note to test that, though.

morpheus4hire:
Treeston I added this to test on base 3.3.5 repo and when player attacks anything it is flagging player as pvp. After 5 minutes of no activity the flag drops but as soon as you attack something again it flags player as if they are in pvp. Went back to regular repo and issue is not present only happens with combat/threat rewrite. The pvp issue seems similar to issue #13103. Was wondering if anyone else is having this issue.

src/server/game/Combat/ThreatManager.cpp
```diff
@@ -291,6 +292,13 @@ void ThreatManager::UpdateOnlineStates(bool meThreateningOthers, bool othersThre
             pair.second->UpdateOnlineState();
 }
 
+static void SaveCreatureHomePositionIfNeed(Creature* c)
+{
+    MovementGeneratorType const movetype = c->GetMotionMaster()->GetCurrentMovementGeneratorType();
+    if (movetype == WAYPOINT_MOTION_TYPE || movetype == POINT_MOTION_TYPE || (c->IsAIEnabled && c->AI()->IsEscorted()))
```
ccrs: Point?

src/server/game/Combat/CombatManager.cpp
```diff
@@ -48,7 +66,12 @@
         if (!a->IsHostileTo(b) && !b->IsHostileTo(a))
             return false;
     }
-    Player *playerA = a->GetCharmerOrOwnerPlayerOrPlayerItself(), *playerB = b->GetCharmerOrOwnerPlayerOrPlayerItself();
+    Creature const* creatureA = a->ToCreature();
+    Creature const* creatureB = b->ToCreature();
+    if ((creatureA && creatureA->IsTrigger()) || (creatureB && creatureB->IsTrigger()))
```
ccrs: some trigger flagged creatures can enter combat, like the controllers in VDW, BPC or Halion

Treeston: realized that during my regular thinking about shit time (aka gym).

src/server/game/Combat/ThreatManager.cpp
Treeston: Same as [current core](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L8730-L8734).

Treeston:
@morpheus4hire I'll investigate.

Treeston:
woopsie, that was a typo - there you go, @morpheus4hire

src/server/game/Entities/Creature/Creature.h
```diff
@@ -101,6 +101,14 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         bool HasReactState(ReactStates state) const { return (m_reactState == state); }
         void InitializeReactState();
 
+        // These shadow the definitions on Unit and override the default value for keepCombat
```
Shauren: missing override keyword on all of them.

Shauren (changes requested):

src/server/game/Entities/Creature/Creature.h
Treeston: they're currently not virtual, though i should probably change that (they're not called often enough that overhead matters)

Shauren: They MUST be virtual.

Or you get this.
```
Unit* unit = target;
unit->SetImmuneToNPC(true, true); // calls Unit::SetImmuneToNPC
unit->ToCreature()->SetImmuneToNPC(true, true); // and now this calls Creature::SetImmuneToNPC
```

Treeston: yup, i know that's what happens, but they do the same thing (only thing being changed is default value for second arg)

Shauren: imo in this case it would be less confusing to have one arg version virtual and the two arg non-virtual and without default arg value in function declaration

Treeston: mhm

i'm also wondering if we ever need the method on base unit. is there ever a situation where non-creatures are set as immune_to_x?

src/server/game/Entities/Creature/Creature.h
```diff
@@ -102,12 +102,15 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
...
-        void SetImmuneToNPC(bool apply) { Unit::SetImmuneToNPC(apply, HasReactState(REACT_PASSIVE)); }
-        void SetImmuneToNPC(bool apply, bool keepCombat) { Unit::SetImmuneToNPC(apply, keepCombat); }
+        using Unit::IsImmuneToAll;
+        using Unit::SetImmuneToAll;
+        void SetImmuneToAll(bool apply) override { Unit::SetImmuneToAll(apply, HasReactState(REACT_PASSIVE)); }
+        using Unit::IsImmuneToPC;
+        using Unit::SetImmuneToPC;
+        void SetImmuneToPC(bool apply) override { Unit::SetImmuneToPC(apply, HasReactState(REACT_PASSIVE)); }
+        using Unit::IsImmuneToNPC;
+        using Unit::SetImmuneToNPC;
```
ccrs: why do you need this?

Treeston: Because that's how C++ scoping works. The override for the one-arg version of `SetImmuneToX` would shadow the base class definition of the two-arg version.

(The `IsImmuneToX` decl is here for clarity)

src/server/game/Entities/Unit/Unit.h
```diff
@@ -1213,12 +1213,15 @@ class TC_GAME_API Unit : public WorldObject
         CombatManager const& GetCombatManager() const { return m_combatManager; }
         void AttackedTarget(Unit* target, bool canInitialAggro);
         
-        void SetImmuneToAll(bool apply, bool keepCombat = false);
         bool IsImmuneToAll() const { return IsImmuneToPC() && IsImmuneToNPC(); }
-        void SetImmuneToPC(bool apply, bool keepCombat = false);
+        void SetImmuneToAll(bool apply, bool keepCombat);
+        virtual void SetImmuneToAll(bool apply) { SetImmuneToPC(apply, false); }
```
Raydor: Is it correct to have SetImmuneToPC in SetImmuneToAll or is it a typo?

Treeston: Yeah, that's a typo. Good catch.

ccrs:
dont worry, that script never worked

Treeston:
Well, it works now. Yay for accidental progress.

Seyden:
Wait.. if all your threat is redirected from the vehicle to the driver.. who does the creature attack? the driver or the vehicle? because while fighting against Leviathan, the damage should redirected to the vehicle..

Treeston:
It's redirect _from_ the driver _to_ the vehicle, not the other way around. Creatures attack the vehicle.

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -2239,14 +2241,21 @@ void AuraEffect::HandleFeignDeath(AuraApplication const* aurApp, uint8 mode, boo
                 }
             }
         }
-        target->CombatStop();
+
+        if (target->GetMap()->IsDungeon()) // feign death does not remove combat in dungeons
```
Shauren: Does this mean that if the hunter is the last target alive during boss encounter and he uses feign death then boss doesn't evade with this change?

src/server/scripts/Spells/spell_rogue.cpp
```diff
@@ -843,6 +844,7 @@ class spell_rog_tricks_of_the_trade : public SpellScriptLoader
     public:
         spell_rog_tricks_of_the_trade() : SpellScriptLoader("spell_rog_tricks_of_the_trade") { }
 
+        class spell_rog_tricks_of_the_trade_SpellScript; // advance declaration so we can friend it from aurascript
```
Shauren: or, you could make a public setter

Shauren (changes requested):

src/server/scripts/Spells/spell_rogue.cpp
Treeston: feels ugly though

src/server/game/Spells/Auras/SpellAuraEffects.cpp
Treeston: No, because feign death adds `UNIT_STATE_DEAD`, which is checked in `ThreatReference::selectOnlineState` and sets the threat ref to offline. It only maintains the actual combat state, e.g. you don't get potion CDs back and such.

src/server/scripts/Spells/spell_rogue.cpp
Shauren: no, it doesnt

Treeston: meh, alright them

src/server/scripts/Spells/spell_rogue.cpp
```diff
@@ -884,8 +883,8 @@ class spell_rog_tricks_of_the_trade : public SpellScriptLoader
                 OnEffectProc += AuraEffectProcFn(spell_rog_tricks_of_the_trade_AuraScript::HandleProc, EFFECT_1, SPELL_AURA_DUMMY);
             }
 
+        public:
```
Shauren: i said a public setter...

void SetRedirectTarget(ObjectGuid const& target) { _redirectTarget = target; }

Shauren (changes requested):

HannibalRoG:
This NPC http://www.wowhead.com/npc=29307/drakkari-colossus with this PR at 50% he will split into an elemental but only goes to start and regenerate his live and attack again.

Treeston:
@HannibalRoG fixed by latest commits - thanks for testing!

Treeston:
Alright, finally got around to tracking down `UNIT_FLAG_PET_IN_COMBAT` in 3.3.5 sniffs. Its behavior is actually quite strange:
* For player pets: Whether the pet is currently going after a target
* For all other units: Whether a minion of that unit is in combat
So it's doing double duty (though, since player pets can never have minions of their own, there is no conflict).

IlPicasso:
is there any eta for completing the merge? Or more bug hunting is expected?

Treeston:
We're currently giving merge priority to #19056 - more than one huge core PR at a time doesn't end well.

Kizarr:
@Treeston #19056 has been completed now we can work on this :)

Treeston:
I'm aware that #19056 is completed, however #18771 is next.

Eliminationzx:
There is two strange behavior:
- Combat stopped sometimes when you attack the target in melee.
cause: https://github.com/Treeston/TrinityCore/blob/c3db3a1df2c43cde176259e1dd93efae0e6ebce6/src/server/game/Entities/Unit/Unit.cpp#L8611
- After use mage spell [invisibility ](https://db.rising-gods.de/?spell=66) in pvp combat stopped for caster and target, should be stopped only for caster.
probably something wrong here: 
https://github.com/Treeston/TrinityCore/blob/c3db3a1df2c43cde176259e1dd93efae0e6ebce6/src/server/game/Combat/CombatManager.cpp#L64

Treeston:
> Combat stopped sometimes when you attack the target in melee.

Your "cause" makes no sense, and "sometimes" is terrible repro. Be more specific, please.

> After use mage spell invisibility in PvP combat stopped for caster and target, should be stopped only for caster.

That's impossible. Combat is _always_ reciprocal - the only exception is if your water elemental is still in combat with the target.

ccrs:
maybe he is talking about the 5s delay in pvp combat, mage shouldnt have delay while the target yes; which I dont think its correct.

Eliminationzx:
Sorry, my bad. first case which I described is invalid,.
But second is still valid - [proof](http://wowwiki.wikia.com/wiki/Invisibility).

@Treeston , can you take a look at rogue spell [vanish ](https://db.rising-gods.de/?spell=26889) ?
This spell also should remove combat as invisibility.

Evidence:
http://wowwiki.wikia.com/wiki/Vanish.
https://www.youtube.com/watch?v=zuKwiM5sUiY 6:08-6:12

I tried to fix combat removing for rogue spell [vanish ](https://db.rising-gods.de/?spell=26889) and I got this behavior:
> After use rogue spell vanish in PvP combat stopped for caster and target, should be stopped only for caster.

```diff
diff --git a/src/server/game/Spells/SpellEffects.cpp b/src/server/game/Spells/SpellEffects.cpp
index b200d29759..ed3f12b586 100644
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -764,6 +764,7 @@ void Spell::EffectTriggerSpell(SpellEffIndex effIndex)
             // Vanish (not exist)
             case 18461:
             {
+                unitTarget->CombatStop();
```

Shauren:
@Eliminationzx That is wrong, Vanish (and Invisibility) are kind of special and behave differently depending on who you are in combat with.
Using them on dungeon and raid bosses should make you "offline on threat list", not fully remove combat (so you don't get additional potion during boss fights) but otherwise they fully remove you from combat (even if someone else is in combat with the creature you were fighting)

Eliminationzx:
Combat can't be mutual in PvP, you can clearly see it in PvP movie from retail.

> https://www.youtube.com/watch?v=zuKwiM5sUiY

src/server/game/Combat/ThreatManager.cpp
lorac2k14: 2017-2017?

Treeston:
This is on hold pending #18771

qyh214:
Thanks!~Can you fix some conflicts in recent core and I can test

Treeston:
This now has merge priority over #18771. Resuming work.

Treeston:
Latest changes Vanish/FD behavior to match retail tests, please review @Eliminationzx.

chaodhib:
Sorry for indirectly slowing down the merge of this PR. I thought I could get https://github.com/TrinityCore/TrinityCore/pull/18771 merged sooner. But in the end, better not wait for it indeed.

Anyway, nice work.

src/server/scripts/World/npcs_special.cpp
```diff
@@ -1817,8 +1817,10 @@ class npc_training_dummy : public CreatureScript
                             // If unit has not dealt damage to training dummy for 5 seconds, remove him from combat
                             if (itr->second < now - 5)
                             {
-                                if (Unit* unit = ObjectAccessor::GetUnit(*me, itr->first))
-                                    unit->getHostileRefManager().deleteReference(me);
+                                auto& refs = me->GetCombatManager().GetPvECombatRefs();
+                                auto it = refs.find(itr->first);
+                                if (it != refs.end())
+                                    (*it).second->EndCombat();
```
Shauren: why are you doing this, `it->`

Treeston: yeah, sure

Treeston:
OK, so these last few commits ensure that our combat managers are in a valid state by the time we call out to the AI hooks.

Before, one side of the relation would get its AI hook called when the `CombatReference` was still dangling (not inserted into the second side). Calling `EndCombat` on such a reference would've crashed the core. That's generally not something we want to happen.

As a result, a bunch of refactoring happened. Combat state is now sequenced safely - both refs are inserted, *then* both combat states are updated, *then* both AI hooks are invoked if needed. This ensures that the managers are in a valid state any time outside code is invoked.

Oh, and 6113b9d was pulled in by the last rebase, which is a refactor commit pushed straight to 3.3.5 base.

src/server/game/Combat/CombatManager.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (!a->IsInWorld() || !b->IsInWorld())
+        return false;
+    // ...the two units need to both be alive
+    if (!a->IsAlive() || !b->IsAlive())
+        return false;
+    // ...the two units need to be on the same map
+    if (a->GetMap() != b->GetMap())
+        return false;
+    // ...the two units need to be in the same phase
+    if (!(a->GetPhaseMask() & b->GetPhaseMask()))
```
Shauren: use `WorldObject::InSamePhase` (phase ids on master)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -14068,10 +13662,13 @@ void Unit::StopAttackFaction(uint32 faction_id)
             ++itr;
     }
 
-    getHostileRefManager().deleteReferencesForFaction(faction_id);
 
-    for (ControlList::const_iterator itr = m_Controlled.begin(); itr != m_Controlled.end(); ++itr)
-            (*itr)->StopAttackFaction(faction_id);
+    for (auto const& pair : m_combatManager.GetPvECombatRefs())
+        if (pair.second->GetOther(this)->GetFactionTemplateEntry()->faction == faction_id)
+            pair.second->EndCombat();
```
Shauren: iterator invalidation here

src/server/scripts/EasternKingdoms/MagistersTerrace/boss_priestess_delrissa.cpp
```diff
@@ -855,19 +855,12 @@ class boss_yazzai : public CreatureScript
             if (Blink_Timer <= diff)
             {
                 bool InMeleeRange = false;
-                ThreatContainer::StorageType const& t_list = me->GetThreatManager().getThreatList();
-                for (ThreatContainer::StorageType::const_iterator itr = t_list.begin(); itr!= t_list.end(); ++itr)
-                {
-                    if (Unit* target = ObjectAccessor::GetUnit(*me, (*itr)->getUnitGuid()))
+                for (auto const& pair : me->GetCombatManager().GetPvECombatRefs())
+                    if (pair.second->GetOther(me)->IsWithinMeleeRange(me))
```
Shauren: brackets pls

src/server/scripts/EasternKingdoms/MagistersTerrace/boss_priestess_delrissa.cpp
```diff
@@ -949,19 +942,12 @@ class boss_warlord_salaris : public CreatureScript
             if (Intercept_Stun_Timer <= diff)
             {
                 bool InMeleeRange = false;
-                ThreatContainer::StorageType const& t_list = me->GetThreatManager().getThreatList();
-                for (ThreatContainer::StorageType::const_iterator itr = t_list.begin(); itr!= t_list.end(); ++itr)
-                {
-                    if (Unit* target = ObjectAccessor::GetUnit(*me, (*itr)->getUnitGuid()))
+                for (auto const& pair : me->GetCombatManager().GetPvECombatRefs())
+                    if (pair.second->GetOther(me)->IsWithinMeleeRange(me))
```
Shauren: {}

src/server/game/Combat/CombatManager.cpp
```diff
@@ -0,0 +1,353 @@
...
+    if (!a->IsInWorld() || !b->IsInWorld())
+        return false;
+    // ...the two units need to both be alive
+    if (!a->IsAlive() || !b->IsAlive())
+        return false;
+    // ...the two units need to be on the same map
+    if (a->GetMap() != b->GetMap())
+        return false;
+    // ...the two units need to be in the same phase
+    if (!WorldObject::InSamePhase(a, b))
```
Shauren: its not static.
a->InSamePhase(b)
also you forgot to remove direct phasemask check

Shauren (changes requested):

src/server/game/Combat/CombatManager.cpp
Treeston: removal done. there's a static version now (to make symmetry explicit)

funjoker:
dies it come to master too ?

tkrokli:
Eventually, it will be cherry-picked or adapted to `master`, but it depends on when Shauren or joschiwald has got time to check what is needed and how much of it needs to be adapted.

Shauren:
@funjoker it crashes so not yet.

funjoker:
ok

tje3d:
@Shauren no more crashes :)

Tatara902:
@Treeston Hello I just reverted the core before this PR and many issues were fixed:
#21881 
#21503
#21564
#21537
#21571 
#21501

Maybe you have time to check it, thanks!

PR #19943, opened 2017-06-23
[PLAYERBOTS]fixed the table name error

miszero:
Core/plugins: fixed the table name error
-fixed the table name "characters",this may  cause the worldserver to crash when join the LFG


Aokromes:
learn to target pull requests.

PR #19955, opened 2017-06-24
Core/SpellAuraEffects: Implement SPELL_AURA_DETECT_AMORE

sirikfoll:
**Changes proposed:**

Alright, I need advice here about the coding.
Got the idea from commit: https://github.com/mangos/mangos/commit/b58b29c9581327129daaf43b9c368810db850771

This was only confirmed with 3.3.5 spells:
-  SpellId=46630 have MiscValue=1 and needs to send value 1 in order to properly detect spell 46901;
-  SpellId=44069 have MiscValue=2 and needs to send value 2 in order to properly detect spell 44067;
-  SpellId=46078 have MiscValue=3 and needs to send value 4 in order to properly detect spell 46077;

So, is there a way to make the code more generic?

- [x] 3.3.5

**Tests performed:** Built and tested with the spell mentioned above

- [x] Improve code

### DIFF
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -229,7 +229,7 @@ pAuraEffectHandler AuraEffectHandler[TOTAL_AURAS]=
     &AuraEffect::HandleAuraModRangedAttackPowerPercent,           //167 SPELL_AURA_MOD_RANGED_ATTACK_POWER_PCT
     &AuraEffect::HandleNoImmediateEffect,                         //168 SPELL_AURA_MOD_DAMAGE_DONE_VERSUS            implemented in Unit::SpellDamageBonus, Unit::MeleeDamageBonus
     &AuraEffect::HandleNoImmediateEffect,                         //169 SPELL_AURA_MOD_CRIT_PERCENT_VERSUS           implemented in Unit::DealDamageBySchool, Unit::DoAttackDamage, Unit::SpellCriticalBonus
-    &AuraEffect::HandleNULL,                                      //170 SPELL_AURA_DETECT_AMORE       various spells that change visual of units for aura target (clientside?)
+    &AuraEffect::HandleDetectAmore,                               //170 SPELL_AURA_DETECT_AMORE       used to detect various spells that change visual of units for aura target
     &AuraEffect::HandleAuraModIncreaseSpeed,                      //171 SPELL_AURA_MOD_SPEED_NOT_STACK
     &AuraEffect::HandleAuraModIncreaseMountedSpeed,               //172 SPELL_AURA_MOD_MOUNTED_SPEED_NOT_STACK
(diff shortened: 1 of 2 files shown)


src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -1531,6 +1531,22 @@ void AuraEffect::HandleModStealthLevel(AuraApplication const* aurApp, uint8 mode
     target->UpdateObjectVisibility();
 }
 
+void AuraEffect::HandleDetectAmore(AuraApplication const* aurApp, uint8 mode, bool apply) const
+{
+    if (!(mode & AURA_EFFECT_HANDLE_REAL))
```
sirikfoll: Which mode is the correct one to use here?

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -1531,6 +1531,22 @@ void AuraEffect::HandleModStealthLevel(AuraApplication const* aurApp, uint8 mode
...
+
+    Unit* target = aurApp->GetTarget();
+
+    if (target->GetTypeId() != TYPEID_PLAYER)
+        return;
+
+    if (GetMiscValue() == 1)
+        target->ApplyModByteFlag(PLAYER_FIELD_BYTES2, PLAYER_FIELD_BYTES_2_OFFSET_AURA_VISION, GetMiscValue(), apply);
+    else
+        target->ApplyModByteFlag(PLAYER_FIELD_BYTES2, PLAYER_FIELD_BYTES_2_OFFSET_AURA_VISION, GetMiscValue() + 1, apply);
```
Shauren: this cant be right

sirikfoll: Doesn't look good, but was the only way i could get it to work with the spells mentioned in the description

src/server/game/Spells/Auras/SpellAuraEffects.cpp
ariel-: I think AURA_EFFECT_HANDLE_SEND_FOR_CLIENT_MASK would suffice

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -1531,6 +1531,19 @@ void AuraEffect::HandleModStealthLevel(AuraApplication const* aurApp, uint8 mode
...
+{
+    if (!(mode & AURA_EFFECT_HANDLE_REAL))
+        return;
+
+    Unit* target = aurApp->GetTarget();
+
+    if (target->GetTypeId() != TYPEID_PLAYER)
+        return;
+
+    target->ApplyModByteFlag(PLAYER_FIELD_BYTES2, PLAYER_FIELD_BYTES_2_OFFSET_AURA_VISION, 1 << (GetMiscValue() - 1), apply);
```
ariel-: Handler should make sure there aren't any other auraeffects with same miscvalue applied before removing the flag

sirikfoll:
Any further feedback on this? Would like to merge it so I can proceed and fix a few follow-up bugs 😄

Shauren (approved):

PR #19974, opened 2017-07-01
Linear time RandomResize

Treeston:
The current `RandomResize` implementation uses `O(n)` calls to `std::advance` and `O(n)` calls to `std::erase`. This means that for `std::vector` (`std::erase` in linear time) and `std::list` (`std::advance` in linear time) it takes quadratic time in the worst case.

This PR proposes an improved implementation of `RandomResize` which cuts asymptotic performance to `O(n)` for all containers.

Runtime breakdown (resizing from size `n` to size `k`):
* Exactly `k` calls to `std::move` (each call takes `ϴ(1)`)
* `<= n` calls to `urand` (each call takes `ϴ(1)`)
* `<= n+k` increments of `container::iterator` (each call takes `ϴ(1)`)
* A single `container.erase` call that erases `(n-k)` elements (takes `O(n-k) = O(n)`)

-> Overall runtime is `O(n)`.

### DIFF
--- a/src/common/Utilities/Containers.h
+++ b/src/common/Utilities/Containers.h
@@ -55,7 +55,6 @@ namespace Trinity
 
         // resizes <container> to have at most <requestedSize> elements
         // if it has more than <requestedSize> elements, the elements to keep are selected randomly
-        // note: after calling this method, the elements remaining in <container> will not necessarily have the same order as they did before the call
         template<class C>
         void RandomResize(C& container, std::size_t requestedSize)
         {
--- a/src/common/Utilities/Containers.h
+++ b/src/common/Utilities/Containers.h
@@ -53,17 +53,30 @@ namespace Trinity
             return size;
         }
 
+        // resizes <container> to have at most <requestedSize> elements
+        // if it has more than <requestedSize> elements, the elements to keep are selected randomly
+        // note: after calling this method, the elements remaining in <container> will not necessarily have the same order as they did before the call
         template<class C>
         void RandomResize(C& container, std::size_t requestedSize)
         {
-            uint32 currentSize = uint32(Size(container));
-            while (currentSize > requestedSize)
+            static_assert(std::is_base_of<std::forward_iterator_tag, typename std::iterator_traits<decltype(std::begin(container))>::iterator_category>::value, "Invalid container passed to Trinity::Containers::RandomResize");
+            if (Size(container) <= requestedSize)
+                return;
+            auto keepIt = std::begin(container), curIt = std::begin(container);
+            uint32 elementsToKeep = requestedSize, elementsToProcess = Size(container);
+            while (elementsToProcess)
             {
-                auto itr = std::begin(container);
-                std::advance(itr, urand(0, currentSize - 1));
-                container.erase(itr);
-                --currentSize;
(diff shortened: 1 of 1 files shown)


src/common/Utilities/Containers.h
```diff
@@ -66,6 +66,25 @@ namespace Trinity
             }
         }
 
+        template<class T>
+        void RandomResize(std::list<T>& container, std::size_t requestedSize)
+        {
+            if (container.size() <= requestedSize)
+                return;
+            uint32 elementsToKeep = requestedSize;
```
jackpoz: any reason to use uint32 instead of size_t ?

Shauren:
How about shuffle, advance(itr, num_to_keep), erase(itr, end)

Treeston:
Current implementation keeps order of remaining elements, so shuffle is out.

src/common/Utilities/Containers.h
Treeston: we have no randomness function that handles `size_t`

Treeston:
Taking inspiration from @Shauren's comment - rewritten to work in linear time for all containers for which copy constructing an iterator takes `ϴ(1)` (this includes all STL containers).

Ordering is not preserved, but none of the current uses of the method rely on this.

src/common/Utilities/Containers.h
```diff
@@ -53,17 +53,30 @@ namespace Trinity
...
 
+        // resizes <container> to have at most <requestedSize> elements
+        // if it has more than <requestedSize> elements, the elements to keep are selected randomly
+        // note: after calling this method, the elements remaining in <container> will not necessarily have the same order as they did before the call
         template<class C>
         void RandomResize(C& container, std::size_t requestedSize)
         {
-            uint32 currentSize = uint32(Size(container));
-            while (currentSize > requestedSize)
+            static_assert(std::is_base_of<std::forward_iterator_tag, typename std::iterator_traits<decltype(std::begin(container))>::iterator_category>::value, "Invalid container passed to Trinity::Containers::RandomResize");
```
Shauren: C::iterator is a thing btw, don't nee this ugly decltype (this function obviously cannot be used on arrays)

Shauren:
Also I am not sure how good the randomness of it is

src/common/Utilities/Containers.h
Warpten:
```c++
template <typename T, typename = void> struct is_iterable : std::false_type {};
template <typename T> struct is_iterable<T, std::void_t<decltype(std::declval<T>().begin()), decltype(std::declval<T>().end())>> : std::true_type {};
```

jackpoz:
I wrote a little something that give a n sized container that must be randomly resized to m, first gets m random indexes and then removes each element from the container:
https://ideone.com/eEpsvp

In case the random index was already marked to be deleted I go to next element and restart at start() if end() is reached. This part might not be random but it's a way to avoid infinite random() calls.

The container is iterated once (other than the call to get the size of the container)

Treeston:
@Shauren perfect uniform distribution on the assumption that `urand` is.

Treeston:
Alright, requested change made, randomness proof is here. Anything else we need, @Shauren?

PR #19975, opened 2017-07-01
Change offset behavior of UnitAI::SelectTarget(List)

Treeston:
Current behavior:
* Ignore the first `offset` units in `targetType` order that match `predicate`.

Proposed behavior:
* Ignore the first `offset` units in `targetType` order, regardless of whether they match `predicate`.

**Example:**
`Unit* target = SelectTarget(SELECT_TARGET_TOPAGGRO, 1, 10.0f, true);`
**Expected (New):** Selects the highest threat non-tank in 10yd.
**Current:** Selects the highest threat non-tank within 10yd, unless the tank is not within 10yd, then selects the second highest threat non-tank within 10yd.

Also includes some major code deduplication (changing `SelectTarget` to call `SelectTargetList` to handle filtering and sorting).

### DIFF
--- a/src/server/game/AI/CoreAI/UnitAI.h
+++ b/src/server/game/AI/CoreAI/UnitAI.h
@@ -171,54 +171,7 @@ class TC_GAME_API UnitAI
                 return nullptr;
 
             std::list<Unit*> targetList;
-            if (targetType == SELECT_TARGET_MAXDISTANCE || targetType == SELECT_TARGET_MINDISTANCE)
-            {
-                for (ThreatReference const* ref : mgr.GetUnsortedThreatList())
-                {
-                    if (ref->IsOffline())
-                        continue;
-
-                    targetList.push_back(ref->GetVictim());
-                }
-            }
-            else
-            {
-                Unit* currentVictim = mgr.GetCurrentVictim();
-                if (currentVictim)
-                    targetList.push_back(currentVictim);
-
-                for (ThreatReference const* ref : mgr.GetSortedThreatList())
-                {
-                    if (ref->IsOffline())
-                        continue;
-
-                    Unit* thisTarget = ref->GetVictim();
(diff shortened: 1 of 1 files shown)


Keader (approved):
Looks good to me

Keader:
Maybe can help someone that will review this pr:
<details>
```
<Treeston> now: selects the 2nd highest aggro person that matches predicate
<Treeston> after: selects the highest aggro person that matches predicate (except the 1st in aggro)
<Keader> both do same stuff to me o.o
<Treeston> no
<Treeston> like
<Treeston> the difference is if the tank doesn't match
<Treeston> for example
<Treeston> you have 3 dudes in melee
<Treeston> warrior is tank
<Treeston> kitty is dps
<Treeston> rogue is also dps
<Treeston> but rogue is a good rogue and keeps using tricks so he's below kitty in aggro
<Treeston> now
<Treeston> you call SelectTarget(SELECT_TARGET_TOPAGGRO, 1, 10.0f);
<Treeston> and it will always pick the kitty (both now and with the change)
<Treeston> but now
<Treeston> the warrior wants to kite the boss
<Treeston> so he runs ahead of him by 15y
<Treeston> and is no longer in 10.0f range
<Treeston> now the current core would pick the rogue
<Treeston> and the new version would pick the kitty
<Treeston> makes sense?
<Keader> hmm
<Keader> based in your example, looks correct
<Keader> (new behavior)
<Treeston> the difference is
<Treeston> for offset = 1
<Treeston> current: select 2nd highest aggro person from (people in 10yd)
<Treeston> new: select highest aggro person from (people in 10yd that aren't the highest aggro person)
<Treeston> so new always skips top aggro no matter if he's in 10yd
<Treeston> current only skips top aggro if he's in 10yd otherwise skips the next highest thing that's in 10yd
```
</details>

Shauren (approved):

Issue #18832, opened 2017-01-14
[3.3.5] MMaps Generator shows wrong %


jackpoz:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:**
The mmaps generator shows 64% and then shows the message "Finished. MMAPS were built". This makes the % totally useless.
Issue added in https://github.com/TrinityCore/TrinityCore/pull/17798

**Current behaviour:** (Tell us what happens.)
The mmaps generator shows 64% and then shows the message "Finished. MMAPS were built".

**Expected behaviour:** (Tell us what should happen instead.)
The % should show in the range [0,100] % .

**Steps to reproduce the problem:**

1. Delete any previous generated mmaps folder
2. .\mmaps_generator --threads 4
3. See this log showing
>64% [Map 598] Building tile [32,32]
[Map 576] [32,31]:  Building movemap tiles...
[Map 598] [32,32]:  Building movemap tiles...
[Map 598] [32,32]:  Building navmesh tile...
[Map 598] [32,32]:  Adding tile to navmesh...
[Map 598] [32,32]:  Writing to file...
[Map 598] Complete!
[Map 576] [32,31]:  Building navmesh tile...
[Map 576] [32,31]:  Adding tile to navmesh...
[Map 576] [32,31]:  Writing to file...
64% [Map 576] Building tile [33,30]
[Map 576] [33,30]:  Building movemap tiles...
[Map 548] [33,32]:  Building navmesh tile...
[Map 548] [33,32]:  Adding tile to navmesh...
[Map 548] [33,32]:  Writing to file...
64% [Map 548] Building tile [34,30]
[Map 548] [34,30]:  Building movemap tiles...
[Map 576] [33,30]:  Building navmesh tile...
[Map 576] [33,30]:  Adding tile to navmesh...
[Map 576] [33,30]:  Writing to file...
64% [Map 576] Building tile [33,31]
[Map 576] [33,31]:  Building movemap tiles...
64% [Map 548] Building tile [34,31]
[Map 548] [34,31]:  Building movemap tiles...
[Map 576] Complete!
[Map 548] [34,31]:  Building navmesh tile...
[Map 548] [34,31]:  Adding tile to navmesh...
[Map 548] [34,31]:  Writing to file...
64% [Map 548] Building tile [34,32]
[...]

PR #20013, opened 2017-07-09
Get all tiles in beginning and calulate proper total tile count

Rochet2:
**Changes proposed:**

- Add the mapid restriction from commandline parameters to the MapBuilder::shouldSkipMap function to allow calculating proper tile count for both single map and all maps processing
- Remove an old print that was commented out
- Move all tile discovering to the part that discovers tiles
- Now that all tiles are discovered at the beginning and we have access to shouldSkipMap in it's entirety we now calculate the actual tile count to be processed per map. The tile count only reflects the amount of tiles to build for a single or all maps however and is not and should not be used for single file or tile processing configurations. See https://github.com/TrinityCore/TrinityCore/blob/07593f6c3488666423093c2c32514a0a378da942/src/tools/mmaps_generator/PathGenerator.cpp#L288-L295
- Move `++m_totalTilesProcessed;` to be after the map tile processing in MapBuilder::buildMap so we dont reach 100% prematurely (and since there is no print we actually never see 100% printed). I felt that having 100% print before we are there and upping the count of a tile being processed before it is processed to be illogical

Example log output
[mmaps_output.txt](https://github.com/TrinityCore/TrinityCore/files/1133867/mmaps_output.txt)

- [x] 3.3.5
- [x] master

- Closes https://github.com/TrinityCore/TrinityCore/issues/18832

- checked that percentage only grows and doesnt shift back and forth between two values
- checked that percentage does reach 99% on both a single map and all maps run
- Built mmaps with old and new extractor and took md5 hashes of the mmap file contents to verify that the output was not changed. Also compared the folder sizes which matched
- Verified moving tile discovering to single threaded part is not spending noticeable amount of time

### DIFF
--- a/src/tools/mmaps_generator/MapBuilder.cpp
+++ b/src/tools/mmaps_generator/MapBuilder.cpp
@@ -54,7 +54,7 @@ namespace MMAP
 {
     MapBuilder::MapBuilder(float maxWalkableAngle, bool skipLiquid,
         bool skipContinents, bool skipJunkMaps, bool skipBattlegrounds,
-        bool debugOutput, bool bigBaseUnit, char const* offMeshFilePath) :
+        bool debugOutput, bool bigBaseUnit, int mapid, char const* offMeshFilePath) :
         m_terrainBuilder     (nullptr),
         m_debugOutput        (debugOutput),
         m_offMeshFilePath    (offMeshFilePath),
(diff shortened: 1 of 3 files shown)


Shauren:
100% where?

Rochet2:
> 100% where?

The percentage is only printed when a tile is starting to be processed.
So when we reach 100% tiles processed there will be no print of "100%".
100% is when the output says "Finished".

jackpoz:
md5 comparison sounds enough for me (making sure the file count is the same too).
LGTM

Issue #19865, opened 2017-06-08
Destructible GameObjects as a floor


Golrag:
**Description:**

Some destructible gameobjects that are used as a floor return the wrong floorZ when using .gps command. This leads to creatures falling through them. This is not always the case on the gameobject as on some places it returns the right value.

**Current behaviour:**
See above.

**Expected behaviour:**

.gps command should return the right z coordinate

**Steps to reproduce the problem:**

1.  .go xyz 483.304047 -2168.442627 840.856812 631
2. .gps
3. Notice the floorZ value.
4. Aggro Lich King and let him chase you

**Branch(es):**

- 3.3.5
- master ?

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/107d6260668f825a2a72e7694b43e7940b3c88aa

**Operating system:** Win10

Treeston:
Also happens on 1660bb7d27d6f42b49012a6b57e3c2b2eab20fd3 - not caused by f6c849729b27b77228704b595de3adaf24da2c10.

Console reports
```
Could not load MMAP name:Icecrown Citadel, id:631, x:31, y:36 (mmap rep.: x:31, y:36)
VMap file './vmaps/631.vmtree' does not exist
```

tkrokli:
@Treeston : I found `631.vmtree` in my latest extraction of vmaps, maybe you haven't extracted new vmaps after commit https://github.com/TrinityCore/TrinityCore/commit/c35793941b3aae9b54dce92389aa528d40050754 (**Core/VMaps: Fix LoS in Strand of the Ancients** (#19847))?

Treeston:
> That file does exist.

tkrokli:
Yes, I misread. Sorry.

Treeston:
```
<Shauren> its caused by this
<Shauren> https://github.com/TrinityCore/TrinityCore/commit/c35793941b3aae9b54dce92389aa528d40050754
<Shauren> i think i need to remind you that both ToC floor and LK area are gameobjects
<Shauren> that means no pathfinding
<Shauren> except apparently there is that unrendered WMO spawn there (just like the sota gates) which provided pathfinding (and also blocked it in sota)
```
Tagging and assigning to @Golrag

Golrag:
I juste made the PR, I am afraid I don't know much about extracting vmaps & wmo's etc and maybe @Warpten would be a better choice for fixing this :/

Treeston:
[...]

Issue #19976, opened 2017-07-02
[Trial of the Crusader] Jormungar stuck after submerge


GreatRedem:
**Description:** Jormungars boss gets stuck in floor after submerge

this happens after this commit: c35793941b3aae9b54dce92389aa528d40050754

**Expected behaviour:** They shouldn't get stuck in under floor
**Branch(es):** 3.3.5
**TC rev. hash/commit:**  e35092c6a164e929f9684b350093b2beefee5d86
**TDB version:**  63
**Operating system:** Debian 9

Demonid:
Confirmed

Keader:
@Warpten

Demonid:
Btw, you can see a lot of random issues on LK too, like Adds getting inside the floor and being unable to be taunted.

HannibalRoG:
I can confirm that with LK encounter with adds.

Warpten:
Uhhhh.

Golrag:
try
```diff
diff --git a/src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/boss_northrend_beasts.cpp b/src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/boss_northrend_beasts.cpp
index 17ce945890..6ffdcfb145 100644
--- a/src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/boss_northrend_beasts.cpp
+++ b/src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/boss_northrend_beasts.cpp
@@ -651,7 +651,10 @@ struct boss_jormungarAI : public BossAI
         events.SetPhase(PHASE_SUBMERGED);
         events.ScheduleEvent(EVENT_EMERGE, 5*IN_MILLISECONDS, 0, PHASE_SUBMERGED);
         me->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_NON_ATTACKABLE | UNIT_FLAG_NOT_SELECTABLE);
-        me->GetMotionMaster()->MovePoint(0, ToCCommonLoc[1].GetPositionX()+ frand(-40.0f, 40.0f), ToCCommonLoc[1].GetPositionY() + frand(-40.0f, 40.0f), ToCCommonLoc[1].GetPositionZ());
+        float x = ToCCommonLoc[1].GetPositionX() + frand(-40.0f, 40.0f);
[...]

Issue #19185, opened 2017-02-22
Core/Vmaps: vmaps for WMO


Killyana:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:** Players can cast throw WMO

**Current behavior:** Players can cast throw WMO, looks like vmaps are not generated for those map objects.

**Expected behavior:** Players mustn't be able to cast throw WMO

**Steps to reproduce the problem:**

1. Find any WMO
2.  .npc add 321
3.  cast any spell on the npc when you're hidden behind the WMO

Example: .go xyz 304.09, 119.107, 46.741245 0  then cast a spell on the spider when you're behind the tree.

**Branch(es):** 3.3.5 

**TC rev. hash/commit:**  rev. d535b98b057f

[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing C++, create a Pull Request.)

Aokromes:
oh, blizzard is very tricky on this.

Killyana:
You can reproduce this in Warsong Gulch with a tree trunk.

Aokromes:
i mean, blizzard does strange things with obstacles, some trees blocks los others no.

Shauren:
but trees generally are not wmos, they are m2 models that do not block los

Killyana:
It's not only trees but also some building and rocks.

[[http://i.imgur.com/tPHHcX7.jpg|alt=octocat]]

joshwhedon:
https://github.com/TrinityCore/TrinityCore/issues/5430
For reference, this is this issue, and it is back.

robinsch:
I'm pretty sure you could cast through almost everything in WSG.

akrom23:
https://github.com/TrinityCore/TrinityCore/pull/18528 - more info.

HelloKitty:
This was the behavior on retail. It was accepted as the behavior in a 2012 commit [here](https://github.com/TrinityCore/TrinityCore/commit/c53d722aa29ff3a39d881de7d6ef86b79deffe48) which was only reverted because it caused issues with pet pathing.

[...]

PR #20024, opened 2017-07-14
Core/VMaps: Fix no collision triangles

Golrag:
**Changes proposed:**

-  Fix collision check when generating vmaps

- 3.3.5
- master

**Issues addressed:** Closes https://github.com/TrinityCore/TrinityCore/issues/19865

- Builds
- Tested in game

**Notes:**
- Reextracting vmaps

### DIFF
--- a/src/common/Collision/VMapDefinitions.h
+++ b/src/common/Collision/VMapDefinitions.h
@@ -25,8 +25,8 @@
 
 namespace VMAP
 {
-    const char VMAP_MAGIC[] = "VMAP_4.3";
-    const char RAW_VMAP_MAGIC[] = "VMAP043";                // used in extracted vmap files with raw data
+    const char VMAP_MAGIC[] = "VMAP_4.4";
+    const char RAW_VMAP_MAGIC[] = "VMAP044";                // used in extracted vmap files with raw data
     const char GAMEOBJECT_MODELS[] = "GameObjectModels.dtree";
 
     // defined in TileAssembler.cpp currently...
--- a/src/tools/vmap4_extractor/vmapexport.cpp
+++ b/src/tools/vmap4_extractor/vmapexport.cpp
@@ -65,7 +65,7 @@ bool preciseVectorData = false;
 
 //static const char * szWorkDirMaps = ".\\Maps";
 char const* szWorkDirWmo = "./Buildings";
-char const* szRawVMAPMagic = "VMAP043";
+char const* szRawVMAPMagic = "VMAP044";
 
 // Local testing functions
 
--- a/src/tools/vmap4_extractor/wmo.cpp
+++ b/src/tools/vmap4_extractor/wmo.cpp
@@ -349,10 +349,8 @@ int WMOGroup::ConvertToVMAPGroupWmo(FILE *output, WMORoot *rootWMO, bool precise
         {
             // Skip no collision triangles
             bool isRenderFace = (MOPY[2 * i] & WMO_MATERIAL_RENDER) && !(MOPY[2 * i] & WMO_MATERIAL_DETAIL);
-            bool isDetail = (MOPY[2 * i] & WMO_MATERIAL_DETAIL) != 0;
-            bool isCollision = (MOPY[2 * i] & WMO_MATERIAL_COLLISION) != 0;
-
-            if (!isRenderFace && !isDetail && !isCollision)
+            bool isCollision = MOPY[2 * i] & WMO_MATERIAL_COLLISION || isRenderFace;
+            if (!isCollision)
                 continue;
 
             // Use this triangle


ccrs:
wrong

just remove https://github.com/TrinityCore/TrinityCore/commit/c35793941b3aae9b54dce92389aa528d40050754#diff-97edc20817c42622adaefabfd511126fR514

leave the rest

Golrag:
Reverted my changes, applied what you suggested but then SotA bugged again
http://i.imgur.com/n6FZbUW.png

Warpten:
> you cant have both at the same time, sota bug is caused by paths not being calculated under a dynamic environment but a static one. That line ignores dynamic wmos so the core can calculate those dynamic paths without restrictions.
> 
> so yeah, cant have it till the core functionality is developed

this is true iff the wmo giving mmap support on LK is destructible, and im not sure it is. i dont see why blizzard would make it destructible anyway because it doesnt provide clientside collision asn i very much pointed out by copy-pasting clientside checks into the extractor.

Golrag:
Not sure if this should be intended:
http://i.imgur.com/lp9pi1i.jpg
Also, the rock does not provide LoS (and it's not a gameobject)

Golrag:
Ugh, looks like this broke sindragosa's Ice Tomb LoS

Warpten:
I know for a fact the trees in elwynn forest aren't supposed to give LoS. Last I checked they did on TC

Killyana:
m2 doesn't provide LoS https://github.com/TrinityCore/TrinityCore/issues/19185

Golrag:
Trees in Elwynn forest still do with this PR.
I also noticed that without this PR, the if-test was never true (on wotlk)
List of affected wmos (7K+ all of them ?): https://gist.github.com/Golrag/06d58e3d2432a5e53acd67e28aa99401

Aokromes:
@Golrag jump into irc :)

Warpten:
Here is the trick, there are M2s part of wmos, and I'm not sure the extracter is smart enough to pull them out of the WMO. Not that it affects elwynn forest trees, but that is something worth thinking about

akrom23:
So, it's a valid fix or not? This bug causes many problems with destructible gos.

Golrag:
Don't think M2s are affected by this, with or without this PR trees in elwynn forest are extacted, the spell system just ignores M2 when LoS checking.
I just implemented the previous check back with updated enum values which matches the comment ```// Skip no collision triangles``` and https://wowdev.wiki/WMO/v17#MOPY_chunk
```cpp
bool isRenderFace() { return F_RENDER && !F_DETAIL; }
bool isCollidable() { return F_COLLISION || isRenderFace(); }
```

Aokromes:
@akrom23 what problems it causes?

Golrag:
I think he is talking about #19865

akrom23:
Yes.

Aokromes:
and how is this pr related to that ticket if that ticket is without this pr?

Kizarr:
Any news on this?

Keader:
sould fix #19976 too

Golrag:
No, as some point combinations (x,y) with ```ToCCommonLoc[1].GetPositionZ()``` are going slightly under the gameobject (it's not a flat gameobject) which results in looking at the floor under the one we actually want. I proposed a fix where we just add 5.f but it doesn't feel like a proper fix,

Warpten:
Do I still need to look into this? Kind of forgot about it and moved away from wow for a couple months

ccrs:
https://github.com/TrinityCore/TrinityCore/pull/20024#issuecomment-315467582

The problem is pathgenerator not the extractor, so either revert the logic so the data takes into account dynamic situations or fix pathgenerator.

Warpten:
Again, no. Client map files (.adts) have models for the doors spawned it, it just so happens that everything gets flagged as no-render. This is your issue here. It's not caused by dynamic los. It's caused by duplicate data where the static tree has references to the wmos that the dynamic tree also has

Warpten:
I will finally take the time to check this all tomorrow.

Golrag:
This location https://puu.sh/y3yEZ/2850816ad5.png
should not provide height data (no path, and npc's fall through it on retail).
Yet with my PR here it does :(
Will test soon if it ignore height data without this PR (might take some time)

Warpten:
its returning ground 60 feet below, its fine

Golrag:
Okay, tried the position I used in my example here on retail: https://github.com/TrinityCore/TrinityCore/issues/19865#issuecomment-315352458
http://i.imgur.com/xvGqDnk.jpg

Got on the spike with a demon hunter, tried to meta on it while I was standing on the spike, no path so I am assuming this is working as intended.

Just trying to find things that might be broken due to this PR. Not easy to do tho. I have to check if it was present without this PR and if there is a change, I have to check retail. I think that's the safest way to check if the PR didn't change things in an unintended way, but time consuming & nearly impossible alone. So I started looking for interesting spots on retail and checked if those were working as intended with/without this PR :p (only really interesting point was the door in black rock (spire?) where you can walk on it and I found that spot in darnassus).

I wish I could be a better help with this issue tho, learning all that stuff and find a way to get (& visualise ?) all the data from wow but I'm afraid I don't have the knowledge & tools (yet).

@Warpten is there something holding you back from accepting this PR ? Would like to know what it is if you can share it. If it's just some testing I can try to make some free time for this and maybe put up a nice list of differences between: current rev, this PR and retail.

Faq:
@Warpten any comment on this?

Warpten:
Checking out now

Warpten (approved):

Warpten:
Regarding ADT floor: appears to be a possible bug in my code, as every single cave entrance is affected

PR #20074, opened 2017-07-23
[3.3.5] Addon channel GM command protocol

Treeston:
Little-known fact for most of you: I write addons.
Pet peeve: Integrating with TC GM commands sucks major balls.

Thus, fixing that.

This PR implements a well-formed protocol for addons to communicate with the server, outlined below:

* All communication happens over the addon channel (`LANG_ADDON` in terms of the core, `CHAT_MSG_ADDON`/`SendAddonMessage` for the client). The prefix used for all messages is `TrinityCore` (in client terms - in core terms, every message starts with `TrinityCore\t`).
* In each message, the first character is the opcode. The following four characters are a unique identifier for the invocation in question, and will be echoed back by the server in every message related to that invocation. Following is the message body, if any.
* The following opcodes are supported:
  * Client to server:
    * `p` - Ping request. The core will always respond by ACKing with the passed identifier. No body.
    * `i` or `h` - Command invocation. The message body is the command text without prefix.
      `i` requests machine-readable output, `h` requests human-readable.
  * Server to client:
    * `a` - ACK. The first message sent in response to any invocation (before any output). No body.
    * `m` - Message. Sent once per line of output the server generates. Body = output line.
    * `o` - OK. Indicates that the command finished processing with no errors. No body.
    * `f` - Failed. Indicates that command processing is done, but there was an error. No body.

Expected overhead is minimal, and this integrates seamlessly with existing command scripts (no changes necessary).

PS: There's also a client-side addon library that exposes this protocol in a developer-friendly way over at [Treeston/LibTrinityCore-1.0](https://github.com/Treeston/LibTrinityCore-1.0).

### DIFF
--- a/src/server/game/Chat/Chat.cpp
+++ b/src/server/game/Chat/Chat.cpp
@@ -16,7 +16,6 @@
  * with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include <boost/algorithm/string/replace.hpp>
 #include "Chat.h"
 #include "AccountMgr.h"
 #include "CellImpl.h"
@@ -34,6 +33,7 @@
 #include "Realm.h"
 #include "ScriptMgr.h"
 #include "World.h"
+#include <boost/algorithm/string/replace.hpp>
 
 ChatCommand::ChatCommand(char const* name, uint32 permission, bool allowConsole, pHandler handler, std::string help, std::vector<ChatCommand> childCommands /*= std::vector<ChatCommand>()*/)
     : Name(ASSERT_NOTNULL(name)), Permission(permission), AllowConsole(allowConsole), Handler(handler), Help(std::move(help)), ChildCommands(std::move(childCommands))
--- a/src/server/game/Chat/Chat.cpp
+++ b/src/server/game/Chat/Chat.cpp
@@ -16,6 +16,7 @@
  * with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <boost/algorithm/string/replace.hpp>
 #include "Chat.h"
 #include "AccountMgr.h"
 #include "CellImpl.h"
@@ -347,6 +348,7 @@ bool ChatHandler::ExecuteCommandInTable(std::vector<ChatCommand> const& table, c
                 SendSysMessage(table[i].Help.c_str());
             else
                 SendSysMessage(LANG_CMD_SYNTAX);
+            SetSentErrorMessage(true);
         }
 
         return true;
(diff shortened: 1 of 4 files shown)


Rochet2:
Before the pain has been mostly (for me at least) that there is no way to determine when a command output starts and ends properly.

So my understanding is that with this we can execute GM commands using addon messages and we can be sure when the gm command output has ended. And we get the status message so we know if the command had an error or if it succeeded (as long as the command is coded correctly).

Example of exchange:
```
c->s: i1234gm on
s->c: a1234
s->c: m1234Syntax: .gm on|off
s->c: f1234
```

I see the opcode `h` there as well. Is that WIP?

Treeston:
Not listed yet - requests human-readable output. Look at [Treeston/LibTrinityCore-1.0](https://github.com/Treeston/LibTrinityCore-1.0), it implements the client half for easy accessibility.

Ascathor:
Would love to have an example file at some point to see how this could be implemented in code. Still have a while to go with lua. Just a general idea. Good thing though, looks clean

src/server/game/Chat/Chat.cpp
```diff
@@ -16,6 +16,7 @@
  * with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <boost/algorithm/string/replace.hpp>
```
Shauren: include order should be
* Main header for the source (chat.h)
* all other TC includes
* 3rd party lib includes
* stl includes
* c includes

src/server/scripts/Commands/cs_server.cpp
```diff
@@ -132,7 +132,10 @@ class server_commandscript : public CommandScript
     // Display the 'Message of the day' for the realm
     static bool HandleServerMotdCommand(ChatHandler* handler, char const* /*args*/)
     {
-        handler->PSendSysMessage(LANG_MOTD_CURRENT, Motd::GetMotd());
+        if (handler->IsHumanReadable())
```
Shauren: I don't really like having to check this in every command

Treeston: Mhm, but I'm not entirely sure thing what a better idea would be.

Treeston:
@Ascathor 
> Look at Treeston/LibTrinityCore-1.0, it implements the client half for easy accessibility.

src/server/scripts/Commands/cs_server.cpp
Shauren: How about this
```
template<typename... Args>
void PSendSysMessage(uint32 entry, Args&&... args)
{
    if (IsHumanReadable()) // virtual
        SendSysMessage(PGetParseString(entry, std::forward<Args>(args)...).c_str());
    else
        SendSysMessage(JoinUnformattedValues(std::forward<Args>(args)...)); // this one would just join all values with some separator to make parsing in addon easier and send it
}
```

Treeston: Eh, main issue I have with that one is addons having to rely on the output order (since we wouldn't send any sort of identifying tag, e.g. `guid: whatever` or `npcflag: stuff`)

Shauren: But that would be exactly the same as your sample (well, bad one because motd has only one arg) just handled one level lower and transparent to command handler

Shauren: Alternative would be to switch all command output to fmtlib named arguments (see http://fmtlib.net/latest/index.html), then you would have access to arg names

Treeston: yea, my example only has that _because_ there's only one arg

named arguments sound like a great idea, but would bloat the PR workload. can we implement that in a follow-up?

Shauren: I can agree to that if you either scrap the current example or go with my initial suggestion (and dont implement the IsHumanReadable flag, always return true) to avoid having to change the same code twice

Treeston:
Removed example + adjusted include order. Merge ready.

src/server/game/Chat/Chat.h
```diff
@@ -111,6 +112,7 @@ class TC_GAME_API ChatHandler
         virtual LocaleConstant GetSessionDbcLocale() const;
         virtual int GetSessionDbLocaleIndex() const;
 
+        bool IsHumanReadable() const { return humanReadable; }
```
Shauren: Right, this was supposed to be virtual and returning true and `bool humanReadable` moved to AddonCliHandler

src/server/game/Chat/Chat.h
```diff
@@ -181,4 +185,22 @@ class TC_GAME_API CliHandler : public ChatHandler
         Print* m_print;
 };
 
+class TC_GAME_API AddonCliHandler : public ChatHandler
```
Shauren: maybe change `Cli` to `Command` (not a command line interface)

Shauren (changes requested):

Shauren:
_Almost_ merge ready.

Treeston:
..._Fully_ merge ready? :P

Treeston:
..._Fully_ merge ready? 😛

Rochet2:
It should be noted that the addon has been moved to https://github.com/TrinityCore/LibTrinityCore-1.0

Issue #20080, opened 2017-07-26
Learn spells at wrong class trainer


Rochet2:
**Description:**

This is just a minor exploit. In commit https://github.com/TrinityCore/TrinityCore/commit/0cd5179726f91cc163910dd0806d622fe86d9a8c
I may have just missed it, but I cannot see a check for if the npc has the trainerid available. This opens a possibility to a hack where a player can learn his spells at any trainer.

I did not test this hack in practice.

Solutions:
- One solution could be saving the trainer guid with the trainerid. that way we can check that the NPC and the trainerid are correct.
- We could have a list of trainerids that the npc has
- One list we already have that has the trainerid possibly is the gossip options list. So one possibility is to check that.

**Current behaviour:**

You can learn spells with hacks at other class trainers after visiting your own class trainer.

**Expected behaviour:**

When sending a spell list to player the trainerid and guid should be saved.
Then later we can check if the trainerid and the trainer guid are correct in the packet sent by client and thus we know that the NPC next to the player is a valid one that has given the player the trainer id originally.

**Steps to reproduce the problem:**

- For example lets say I visit my warrior trainer that has guid W in stormwind. My trainerid is set to X.
- Then I walk into some wilderness and in there happens to be a rogue trainer with guid Y.
- I walk near to him and send a packet that is handled by `WorldSession::HandleTrainerBuySpellOpcode`.
- The packet holds the trainerid X and the guid Y and the spellid Z of a valid spell I can learn.
- The handler will check that the npc with guid Y is found next to me and has trainer flag
- It also checks that the trainerid X exists and that the spellid Z is from the trainerid X and that it is learnable by me.
- The spell is teached to be then without checking that npc Y has the trainerid X or that the npc is the one that was present when setting the trainerid X (so check that Y == W).

**Branch(es):**

master

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/0cd5179726f91cc163910dd0806d622fe86d9a8c

**TDB version:**

**Operating system:**

Shauren:
You mean this? https://github.com/TrinityCore/TrinityCore/commit/0cd5179726f91cc163910dd0806d622fe86d9a8c#diff-dfce1c1a49fc16b89987c23c06b5dd2eR135

Rochet2:
@Shauren no.

Shauren:
[...]

PR #20085, opened 2017-07-27
Implement a general way to store selection data for gossip and fix trainer exploit

Rochet2:
**Changes proposed:**

- Make a single place to store sender guid and related information for gossip actions - now sending a new gossip menu erases trainerid for example
- Fix trainer exploit by checking that trainer guid is same when sending trainer menu and when learning spells
- Use the sender guid storage to store mailbox guid to check later interaction with the mailbox
- Implement CloseInteraction packet that will call reset for the interaction data.

- [x] master

Closes https://github.com/TrinityCore/TrinityCore/issues/20080

Compiles.
Gossip at stormwind city guard works.

- [x] After implement https://github.com/TrinityCore/TrinityCore/pull/20085#discussion_r130220802

### DIFF
--- a/src/server/game/Entities/Creature/GossipDef.h
+++ b/src/server/game/Entities/Creature/GossipDef.h
@@ -241,7 +241,7 @@ class InteractionData
 
         void Reset()
         {
-            SourceGuid = ObjectGuid::Empty;
+            SourceGuid.Clear();
             TrainerId = 0;
         }
 
--- a/src/server/game/Entities/Creature/GossipDef.h
+++ b/src/server/game/Entities/Creature/GossipDef.h
@@ -234,18 +234,18 @@ class TC_GAME_API QuestMenu
         QuestMenuItemList _questMenuItems;
 };
 
-class SelectionData
+class InteractionData
 {
     public:
-        SelectionData() { Reset(); }
+        InteractionData() { Reset(); }
 
         void Reset()
         {
-            SenderGuid = ObjectGuid::Empty;
+            SourceGuid = ObjectGuid::Empty;
             TrainerId = 0;
         }
 
-        ObjectGuid SenderGuid;
+        ObjectGuid SourceGuid;
         uint32 TrainerId;
 };
 
@@ -257,7 +257,7 @@ class TC_GAME_API PlayerMenu
 
         GossipMenu& GetGossipMenu() { return _gossipMenu; }
         QuestMenu& GetQuestMenu() { return _questMenu; }
-        SelectionData& GetSelectionData() { return _selectionData; }
+        InteractionData& GetInteractionData() { return _interactionData; }
 
         bool Empty() const { return _gossipMenu.Empty() && _questMenu.Empty(); }
 
@@ -289,6 +289,6 @@ class TC_GAME_API PlayerMenu
         GossipMenu _gossipMenu;
         QuestMenu  _questMenu;
         WorldSession* _session;
-        SelectionData _selectionData;
+        InteractionData _interactionData;
 };
 #endif
--- a/src/server/game/Entities/Creature/GossipDef.h
+++ b/src/server/game/Entities/Creature/GossipDef.h
@@ -237,7 +237,7 @@ class TC_GAME_API QuestMenu
 class SelectionData
 {
     public:
-        SelectionData() { Reset(); };
+        SelectionData() { Reset(); }
 
         void Reset()
         {
--- a/src/server/game/Entities/Creature/GossipDef.h
+++ b/src/server/game/Entities/Creature/GossipDef.h
@@ -241,12 +241,12 @@ class SelectionData
 
         void Reset()
         {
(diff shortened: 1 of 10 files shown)


src/server/game/Entities/Creature/GossipDef.h
```diff
@@ -237,6 +234,21 @@ class TC_GAME_API QuestMenu
         QuestMenuItemList _questMenuItems;
 };
 
+class TC_GAME_API SelectionData
```
Shauren: TC_GAME_API is not needed when the class is fully contained in header

src/server/game/Entities/Creature/GossipDef.h
```diff
@@ -237,6 +234,21 @@ class TC_GAME_API QuestMenu
         QuestMenuItemList _questMenuItems;
 };
 
+class SelectionData
+{
+    public:
+        SelectionData() { Reset(); };
```
joschiwald: useless extra ; at the end

src/server/game/Entities/Creature/GossipDef.h
```diff
@@ -237,6 +234,21 @@ class TC_GAME_API QuestMenu
...
+
+        void Reset()
+        {
+            senderGuid = ObjectGuid::Empty;
+            trainerId = 0;
+        }
+
+        ObjectGuid senderGuid;
+        uint32 trainerId;
```
joschiwald: UpperCamelCase for public members? ;)

src/server/game/Entities/Creature/GossipDef.h
```diff
@@ -237,6 +234,21 @@ class TC_GAME_API QuestMenu
         QuestMenuItemList _questMenuItems;
 };
 
+class SelectionData
+{
+    public:
+        SelectionData() { Reset(); }
+
+        void Reset()
```
joschiwald: should CMSG_CLOSE_INTERACTION do a reset?

src/server/game/Entities/Creature/GossipDef.h
```diff
@@ -237,6 +234,21 @@ class TC_GAME_API QuestMenu
...
+        {
+            SenderGuid = ObjectGuid::Empty;
+            TrainerId = 0;
+        }
+
+        ObjectGuid SenderGuid;
```
joschiwald: maybe InteractionData and TargetGuid are better words? what do you think?

Rochet2: Interaction sure, but TargetGuid I am not sure about.
I would rather use Sender or Source or something like that myself. I also think in may places source is used in core already.

src/server/game/Entities/Creature/GossipDef.h
Rochet2: Not sure what sends that or where/how it is used.
But if that is sent on gossip ending or closing windows like trainer window then it should reset most likely to prevent buying stuff etc after window has been closed - though I wouldnt consider that a big plus as you could prevent sending the packet through hacks anyways and then do the actions even after closing the window.

Seems that this system could also be applied to mails as Im looking at the comments in https://github.com/TrinityCore/TrinityCore/blob/b453e124231a90321fe79fbf3a62acdcfa54a691/src/server/game/Handlers/MailHandler.cpp#L385

Rochet2: Not sure how to implement the packet handler though.
Enough to have this?
```cpp
void WorldSession::HandleCloseInteraction(WorldPacket& /*recvData*/)
{
    _player->GetInteractionData().Reset();
}
```

joschiwald: you need to add a new packet class, the packet has one guid iirc, but maybe we can declare implementing that packet of of score of this pr

src/server/game/Entities/Creature/GossipDef.h
```diff
@@ -234,18 +234,18 @@ class TC_GAME_API QuestMenu
...
+class InteractionData
 {
     public:
-        SelectionData() { Reset(); }
+        InteractionData() { Reset(); }
 
         void Reset()
         {
-            SenderGuid = ObjectGuid::Empty;
+            SourceGuid = ObjectGuid::Empty;
```
joschiwald: `SourceGuid.Clear()`

src/server/game/Entities/Creature/GossipDef.h
DoctorKraft: Base of packet If this can help you [here](https://gist.github.com/DoctorKraft/78c3002b896ccf9f153aede043503617)

Rochet2:
Are guids unique throughout different maps?
If creature/gameobject/item guids are not comparable between maps, then we need to also call reset when player changes maps.

Shauren:
Map id is part of the guid, it is guaranteed that you will not find any unit with the same guid on different map

Shauren (approved):

PR #20103, opened 2017-08-02
[WIP][3.3.5] Implement spawn group management in InstanceScript

Treeston:
What it says on the tin - begin dynspawn follow-up work by gradually phasing out `linked_respawn` in favor of spawn groups for instance spawn control.

Medium term, once all instances have appropriate DB data, this will allow us to remove the `instanceId` column from the respawn tables altogether, relying exclusively on this system.

New DB table `instance_spawn_groups`:
* `instanceMapId` - duh
* `bossStateId` - duh
* `bossStates` - bitmask, `1 << state` for `EncounterState` values
* `flags` - bitmask, currently allows `0x1` (`FLAG_ACTIVATE_SPAWN`) and `0x2` (`FLAG_BLOCK_SPAWN`)

A spawn group is activated if _any_ of its `FLAG_ACTIVATE_SPAWN` conditions are met, unless _any_ of its `FLAG_BLOCK_SPAWN` conditions are met - this allows for some neat complex management if necessary.

PS: Instance spawn data is currently being crowdsourced - check [this thread](https://community.trinitycore.org/topic/13650-the-great-instance-spawn-crowdsource-initiative-of-2017).


src/server/game/Instances/InstanceScript.cpp
```diff
@@ -347,6 +381,13 @@ bool InstanceScript::_SkipCheckRequiredBosses(Player const* player /*= nullptr*/
     return player && player->GetSession()->HasPermission(rbac::RBAC_PERM_SKIP_CHECK_INSTANCE_REQUIRED_BOSSES);
 }
 
+void InstanceScript::Create()
+{
+    for (int8 i = 0; i < bosses.size(); ++i)
```
tkrokli:
```
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Instances/InstanceScript.cpp:386:24:
 fatal error:
 comparison of integers of different signs: 'int8' (aka 'signed char') and 'size_type' (aka 'unsigned long') [-Wsign-compare]
    for (int8 i = 0; i < bosses.size(); ++i)
                     ~ ^ ~~~~~~~~~~~~~
```

src/server/game/Instances/InstanceScript.h
```diff
@@ -21,6 +21,7 @@
 
 #include "ZoneScript.h"
 #include "Common.h"
+#include "ObjectMgr.h"
```
Shauren: objection.
not this header.

Treeston: Hrm, just use the actual type (`std::unordered_multimap<` etc) instead of the typedef?

src/server/game/Instances/InstanceScript.cpp
Treeston: Yes, I am capable of reading travis output. Thank you.

src/server/game/Instances/InstanceScript.h
Treeston: Or no, that won't work either (type itself is defined in `ObjectMgr.h`). I'll think of something.

src/server/game/Instances/InstanceScript.h
```diff
@@ -135,6 +135,22 @@ struct MinionInfo
...
+        FLAG_BLOCK_SPAWN = 0x02,
+
+        FLAG_ALL = (FLAG_ACTIVATE_SPAWN | FLAG_BLOCK_SPAWN)
+    };
+    uint8 bossStateId;
+    uint8 bossStates;
+    uint32 spawnGroupId;
+    uint8 flags;
+};
+typedef std::unordered_multimap<uint16, InstanceSpawnGroupInfo> InstanceSpawnGroupMap;
```
Shauren: I still don't like this, none of these headers should include each other (in any direction)
Proposed fix: change the container to `std::unordered_map<uint16, std::vector<InstanceSpawnGroupInfo>>`
and hold a pointer to that vector in InstanceScript

Treeston: Yeah, but where do we define `InstanceSpawnGroupInfo`? Class definitions need to be available at point of substitution, iirc?

Shauren: It doesn't matter. vector permits incomplete types

Treeston: Ah, great.

Krudor:
Hell yes

Keader:
@Treeston , @Ovahlord we need update the wiki with this table.

illfated:
Oh right. I see. At least we need a new table name entry and the preliminary field names,
so others can fill in the necessary explanations based on the following SQL :
```sql
CREATE TABLE `instance_spawn_groups` (
  `instanceMapId` smallint(5) unsigned not null,
  `bossStateId` tinyint unsigned not null,
  `bossStates` tinyint unsigned not null,
  `spawnGroupId` int unsigned not null,
  `flags` tinyint unsigned not null,
  PRIMARY KEY (`instanceMapId`,`bossStateId`,`spawnGroupId`,`bossStates`)
```

PR #20110, opened 2017-08-04
[3.3.5] Core/Pet: Fix pet action bar loosing autocast settings (#16211)

### LINKED ISSUE #16211: [3.3.5] Core/Pet: Hunter's Pet actionbars doesn't save data
After relog/dismount the hunter's pet action bars data will reset

How to reproduce:
1. Tame a pet ( for example - entry: 19350 )
2. in your spell book pet's spell will appear so you could change spell to automatic cast or self cast ( by right click on them ) make pet's spell self cast
3. Mount up then dismount, the spell goes as same as first time it was

Revision: 94e4bddb903d581293b5a7f50f9b32d1262d2c20

necropola:
reverts one line of 6c21ddd9b98d372cb802c9abf034e2aa3ecc70b9
Target branch:  3.3.5
Issues addressed: Closes #16211
Tests performed: build succeeded and tested/verified in game (me and Eronox)

### DIFF
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -1520,7 +1520,7 @@ bool Pet::removeSpell(uint32 spell_id, bool learn_prev, bool clear_ab)
     }
 
     // if remove last rank or non-ranked then update action bar at server and client if need
-    if (m_charmInfo->RemoveSpellFromActionBar(spell_id) && !learn_prev && clear_ab)
+    if (clear_ab && !learn_prev && m_charmInfo->RemoveSpellFromActionBar(spell_id))
     {
         if (!m_loading)
             GetOwner()->PetSpellInitialize(); // need update action bar for last removed rank


src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -1520,7 +1520,7 @@ bool Pet::removeSpell(uint32 spell_id, bool learn_prev, bool clear_ab)
     }
 
     // if remove last rank or non-ranked then update action bar at server and client if need
-    if (m_charmInfo->RemoveSpellFromActionBar(spell_id) && !learn_prev && clear_ab)
```
necropola: I changed it back to what is was. intentionally. order matters in short-circuit boolean evaluation and executing  RemoveSpellFromActionBar() in any case broke the pet action bar autoccast settings.

Treeston: > order matters in short-circuit boolean evaluation

That's the point. You're changing the order of the `RemoveSpellFromActionBar` invocation w.r.t. two different bools (`clear_ab` and `learn_prev`) - which of the two is the intended change (or is it both).

Treeston: Upon looking at diff history - both is intended.

necropola: I want the function call to only happen when both ``clear_ab == true`` and  ``!learn_pre == true`` (``learn_prev == false``) or rather I want the order to be restored to what it was before commit 6c21ddd9b98d372cb802c9abf034e2aa3ecc70b9. My patch simply reverts one line of this commit and thereby fixes an issue obviously caused by it, i.e it's 100% reproducable/fixed without/with my patch.
I would be happier myself, if I could give a better explanation, but I just started a few days ago to familiarize myself with the _TrinityCore_ source, actually with the intention to fix this (and maybe some other) issue(s).   .

Treeston:
@necropola btw to avoid cluttering the history of every PR you make by basing them all off the same branch,  I'd recommend creating a separate branch based on 3.3.5 for every PR you make (f.ex., `3.3.5-petautocast`, `3.3.5-thisawesomething` etc) and then PRing from those.

necropola:
will do. still need to learn a lot. but I'm getting there.

Treeston:
No worries, you'll get there.

PS: Consider getting on IRC.

necropola:
My second patch/commit is now in a separate 3.3.5-branch of my fork.
https://github.com/necropola/TrinityCore/tree/3.3.5-PetSpellMapSyncDB
Do I have to discard the current pull request and make a new one based on the new branch?
.. or is there a way to adjust the branch/commit in my current pull request?

Treeston:
No, sadly. Just keep the PR as-is and start using separate branches going forward.

PR #20112, opened 2017-08-04
[3.3.5] Core/Pet: Fix PetSpellMap <-> DB sync

### LINKED ISSUE #16211: [3.3.5] Core/Pet: Hunter's Pet actionbars doesn't save data

necropola:
``PetSpellMap`` entries with ``state == PETSPELL_REMOVED`` must not be erased before DB sync. 
* Target branch: 3.3.5
* Issues addressed: Finally closes #16211
* Tests performed:
  * build success
  * tested/verified in game
  * DB ``characters.pet_spells`` checked
  * previous spell ranks are now correctly removed from DB

The previous patch was only a partial fix for the issue and didn't actually address the root cause, which is ``PetSpellMap`` and ``characters.pet_spell``are getting out of sync. The current fix only helps with (ranked non-talent) spells that are toggled on the pet action bar, because those are also written to ``characters.character_pet``.
Ranked non-talent spells that are not on the pet action bar, but put on auto via the pet spell book only are not correctly remembered without this second fix.

### DIFF
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -1307,7 +1307,6 @@ bool Pet::addSpell(uint32 spellId, ActiveStates active /*= ACT_DECIDE*/, PetSpel
     {
         if (itr->second.state == PETSPELL_REMOVED)
         {
-            m_spells.erase(itr);
             state = PETSPELL_CHANGED;
         }
         else if (state == PETSPELL_UNCHANGED && itr->second.state != PETSPELL_UNCHANGED)
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -1306,23 +1306,22 @@ bool Pet::addSpell(uint32 spellId, ActiveStates active /*= ACT_DECIDE*/, PetSpel
     if (itr != m_spells.end())
     {
         if (itr->second.state == PETSPELL_REMOVED)
-        {
             state = PETSPELL_CHANGED;
-        }
-        else if (state == PETSPELL_UNCHANGED && itr->second.state != PETSPELL_UNCHANGED)
-        {
-            // can be in case spell loading but learned at some previous spell loading
-            itr->second.state = PETSPELL_UNCHANGED;
-
-            if (active == ACT_ENABLED)
-                ToggleAutocast(spellInfo, true);
-            else if (active == ACT_DISABLED)
-                ToggleAutocast(spellInfo, false);
-
-            return false;
-        }
         else
+        {
+            if (state == PETSPELL_UNCHANGED && itr->second.state != PETSPELL_UNCHANGED)
+            {
+                // can be in case spell loading but learned at some previous spell loading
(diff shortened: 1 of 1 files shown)


Treeston:
I'm not intricately familiar with the relevant code paths. Could you outline the incorrect logic vs your correct fix.?

necropola:
The idea - as far as I understood after analyzing the code - is that ``PetSpellMap`` is the server-side memory representation of the pet's spell book which is made persistent by saving to / loading from certain tables in the ``character`` database (``pet_spell`` in particular). For performance reasons the database isn't updated on every change of the map but as a bulk operation on certain events , e. g. mounting/unmounting. 

``PetSpellMap`` basically stores the currently available spells along with an``active`` value (autocast on/off) and a ``state`` value (map entry needs to be written to / deleted from database). A map entry must not be erased from the map until it is deleted from the database. Instead it needs to be flagged for deletion by setting the state value to ``PETSPELL_REMOVED``. Which means only the DB syncing code should be actually erasing entries from``PetSpellMap``.

There are currently **two** occurrences of ``m_spells.erase()`` outside of the DB sync code. One doesn't seem to cause problems (unchanged), but the other (removed) can actually lead to the _pathological_ case of erasing an entry from the ``PetSpellMap`` before it is also deleted from the DB. In fact it happens when``addSpell`` cleans up ``PetSpellMap`` from lower ranked spells. The result is that ``PetSpellMap`` in memory and its representation on disk (DB) get out of sync. Multiple ranks of the same spell are still in ``pet_spell`` and get loaded from the DB on unmount / pet call, passed into ``addSpell`` and finally lead to autocast flags not being properly restored.

I originally just noticed that the pet bar autocast issue only happens when there are entries for different ranks of the same spell in the ``pet_spell`` table (DB) and I was able to fix it (per pet) by manually deleting the lower ranks from the DB. Then I noticed how much effort the code puts into removing lower rank spells from ``PetSpellMap`` and that it fails to correctly update the DB, too. And finally I spotted the reason,

But why did my previous patch seem to have fixed autocast settings?
Well, it only fixed it for spells that are on the pet bar, not for those that are only toggled via spell book. The _ugly_ fact that autocast settings are stored in two places (``pet_spell`` for all spells, and ``character_pet`` for spells on the action bar) was able to obscure the root cause for a long time ...

Treeston:
:+1: great analysis and good catch

necropola:
Seems you integrated a slightly different fix. Guess that's alright. It reminds me to keep my work in a separate branch (3.3.5-my_work) to avoid merge conflicts when pulling "3.3.5" from upstream. :-)

Treeston:
You can simply base your branches on `upstream/3.3.5` then commit just the PR branch itself to your fork. No reason to keep your fork's main `3.3.5` branch up-to-date all the time.

PR #20128, opened 2017-08-09
[3.3.5] Core/Movement: Pets go for the back unless being the victim

### LINKED ISSUE #19925: [3.3.5] Pets should attack from behind the target $150
**Description:**

Pets in PvE should try to position themselves behind targets to attack them and avoid being parried.
https://us.battle.net/forums/en/wow/topic/1965837204

**Current behaviour:**  Pets always attack from the front.

**Expected behaviour:**  The movement generator for pets should try to position them behind NPC targets when attacking

1. Summon a hunter or warlock pet
2. Send it to attack a creature

**Branch(es):**  3.3.5

**TC rev. hash/commit:**  https://github.com/TrinityCore/TrinityCore/commit/9f765a162092e8b65cc9bc924e0c456f719b3678

necropola:
This restores the pet's original behavior to **not always** attack from the front and get cleaved down.
* Target branch: 3.3.5
* Issues addressed: Closes #19925 
* Tests performed:
  * build success
  * tested/verified in game

I was neither able to pinpoint the commit that broke it nor do I think I found the best place/way to fix it, but it's very effective and should make warlocks and hunters happy again.

Demo showing the new (restored) pet behavior: https://youtu.be/9p4CYmOOWaE

### DIFF
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -140,6 +140,7 @@ void TargetedMovementGenerator<T, D>::SetTargetLocation(T* owner, bool updateDes
             float size = owner->GetCombatReach();
 
             if (owner->IsPet())
+            {
                 if (GetTarget()->GetTypeId() == TYPEID_PLAYER)
                 {
                     distance = 1.0f;
@@ -149,6 +150,7 @@ void TargetedMovementGenerator<T, D>::SetTargetLocation(T* owner, bool updateDes
                     _angle = float(M_PI); // then pet goes for the back
                 else
                     _angle = 0.f; // else pet goes for the front
+            }
 
             if (GetTarget()->IsWithinDistInMap(owner, distance))
                 return;
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -139,11 +139,16 @@ void TargetedMovementGenerator<T, D>::SetTargetLocation(T* owner, bool updateDes
             float distance = _offset + 1.0f;
             float size = owner->GetCombatReach();
 
-            if (owner->IsPet() && GetTarget()->GetTypeId() == TYPEID_PLAYER)
-            {
-                distance = 1.0f;
-                size = 1.0f;
-            }
+            if (owner->IsPet())
+                if (GetTarget()->GetTypeId() == TYPEID_PLAYER)
+                {
+                    distance = 1.0f;
+                    size = 1.0f;
+                }
+                else if (GetTarget()->GetVictim() != owner) // if pet is not the target's victim
+                    _angle = float(M_PI); // then pet goes for the back
+                else
+                    _angle = 0.f; // else pet goes for the front
 
             if (GetTarget()->IsWithinDistInMap(owner, distance))
                 return;


Shauren (dismissed):
Seems legit?

necropola:
I mean it's clearly not the right/best place to adjust the`` _angle`` value. But seeing that the code already adjusts ``offset`` and ``size`` for the *pet runs to owner* case, I thought "what the heck, why not fix it here!?"

ccrs (approved):
If you've tested it... 👍

necropola:
I'm running around with a warlock and a hunter for some time now and it's clearly an improvement. I haven't seen any *pet dance* regression (pet tries to reach the target's back while target tries to attack from the front) and the pet does no longer actively go for the front when it's not the target.

It's still not 100% what it used to be. In older versions of TC-3.3.5 the pet would just approach the target in a straight line and attack from there. Like it still does when using _Charge_ (which uses a different ``MovementGenerator``).

necropola:
Anything I can do to increase the likelihood that this gets merged into TC-3.3.5?

tkrokli:
Not really, it will be approved and merged when enough developers have had the chance to review it.

Shauren:
Alright, looks like I found a possible problem with it
By manipulating `_angle` you prevent this code from triggering when reaching target https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L92

necropola:
I understand that ``_angle`` is used to calculate a point at ``offset`` distance from the target relative to where the target is currently facing, i. e. if ``_angle == 0``  this point is in front of the target and if ``_angle == M_PI`` it's in its back.
Currently the ``TargetedMovementGenerator`` seems to be used with ``_angle == M_PI/2`` for the case that the pet returns to it's owner, and with ``_angle == 0`` to approach a hostile target which has the undesired effect of going for the front, even when the target is after someone else (the tank).
My patch simply catches the latter case and prevents the pet from going suicidal.

ccrs:
Thing is the pet must face the creature when reaches the target (thats what the section shauren linked does). The pet will face its target visually (clientside) because it is targeting it (unit::SetTarget) and client makes the unit always face its target.

But serverside the orientation will remaing incorrect, maybe being autocorrected for a second if it casts, but returning to old (incorrect) after the cast. This can break many things like autoattacks or spells that require facing.

necropola:
Could we change it so that the pet (server-side) faces a hostile target even when in its back?
In older versions of TC the pet would simply approach in a straight line like it still does when using it's _Charge_ ability  and then attack from the point where it reached the target (ignoring the target's orientation by obviously using a different MovementGenerator, maybe Point?).
At some point along the line (when?) it has probably been changed to use the ``TargetedMovementGenerator`` and thereby introduced pet suicide mode ...

Jildor:
any news?

tkrokli:
If I have understood the conversation correctly, we are still waiting for a solution to the issue Shauren pointed out in https://github.com/TrinityCore/TrinityCore/pull/20128#issuecomment-323595682 .

Keader:
Any news on this @necropola ?

chaodhib:
Please don't solve this through the MoveGen. Like you said, it's not the best place to put that code. And the reason pet logic is already in the MoveGen is because of the laziness of past contributors. Let's not do the same.

PS: I'm in the process of cleaning up that MoveGen.

chaodhib (changes requested):
Please don't put that logic in the MoveGen. Like you already pointed it out, it's not a suitable place for pet logic (or any non movement logic really).

The reason there already is pet logic there is because of the laziness from past contributors. Let's not keep doing the same thing.

PS: I'm in the process of refactoring the MoveGen to clean it up from this kind of stuff. 
PS2: I'm also in the process of improving the MoveGen. The overhead of complexity created by these unnecessary coupling between the different parts of the core are slowing down my progress significantly.
PS3: Nice work nonetheless. I know contributing to the project for the first time is hard, because of the size and the complexity. Keep it up!

tkrokli:
I do hope @necropola is still alive and around to receive the feedback.
Would be sad to see useful & practical feedback go to waste.

Shauren:
Replaced by 2a84562dc85516f432bb1e5de9add23c28b26ce4

PR #20129, opened 2017-08-09
Core/Script: Fix Heigan runaway eruption event

Foereaper:
By scheduling the eruption event without cancelling the previously registered eruption events, Heigan's eruption (or dance) would overlap and cause a runaway scenario, where eruptions would not follow a proper timer.

By using Reschedule and not locking the Eruption event to the fighting phase, the dance now works as intended.

**Changes proposed:**

-  Replace multiple event schedules with re-scheduling and not locking to specific phase.

- [x] 3.3.5
- [x] master

- None I could see

- Builds, tested solo and with a raid


Keader:
@Treeston this changes are correct? if yes, please merge :)

PR #20134, opened 2017-08-10
Core/BG: Flag Wintergrasp grids as always active preventing grid unloading

Foereaper:
This fixes the issues regarding Wintergrasp gameobjects despawning and making it impossible to complete the BG.

**Changes proposed:**

-  Flag all Wintergrasp related grids as active on battlefield ~~startup~~ setup.
- Possibly add a config param for whether to keep the grids active or not?

**Target branch(es):** 

- [x] 3.3.5
- [x] master

**Issues addressed:**

#8661

**Tests performed:**

- Builds, tested over the span of two weeks from first iteration with larger playerbase.

- [ ] Possibly add config option if necessary

### DIFF
--- a/src/server/game/Battlefield/Zones/BattlefieldWG.cpp
+++ b/src/server/game/Battlefield/Zones/BattlefieldWG.cpp
@@ -64,11 +64,25 @@ Position const WintergraspStalkerPos   = { 4948.985f, 2937.789f, 550.5172f,  1.8
 Position const WintergraspRelicPos     = { 5440.379f, 2840.493f, 430.2816f, -1.832595f };
 QuaternionData const WintergraspRelicRot    = { 0.f, 0.f, -0.7933531f, 0.6087617f };
 
+uint8 const WG_MAX_GRID             = 8;
 uint8 const WG_MAX_OBJ              = 32;
 uint8 const WG_MAX_TURRET           = 15;
 uint8 const WG_MAX_TELEPORTER       = 12;
 uint8 const WG_MAX_WORKSHOP         = 6;
 uint8 const WG_MAX_TOWER            = 7;
+
+// Grids to mark as active and locked
+GridCoord const WGGridCoord[WG_MAX_GRID] =
+{
+    { 40, 35 },
+    { 40, 36 },
+    { 40, 37 },
+    { 40, 38 },
+    { 41, 36 },
+    { 41, 37 },
+    { 42, 36 },
+    { 42, 37 }
+};
 
 // *****************************************************
 // ************ Destructible (Wall, Tower..) ***********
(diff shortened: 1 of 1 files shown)


Warpten:
Unlock them when the battlefield enfs

Foereaper:
This would re-introduce the issue with the objects despawning, as this is the exact issue we are facing now. 

If players stay in the zone during the entire "cooldown" period, then everything would work as it should with the current implementation as well, however players do not stick around when no battle is ongoing, causing the objects to unload and it breaks the BG.

The only other "solution" would be to constantly respawn the objects when the grids load, apply damage states and the likes. This was also discussed in the issue tagged above.

Shoxxo:
But this Patch will work as it is in the description or not?

Foereaper:
Yes, it will flag all the grids as active on startup causing the objects to stay loaded and prevent them from despawning and breaking the bg.

Shoxxo:
And the grids will unload after that? I see no unload function atm?

Foereaper:
No, the grids are supposed to stay locked to prevent them from unloading during the time between battles, because the unloading breaks the bg.

Shoxxo:
Okay, i will try a test later, need first modify some files on my server to stay tuned with eluna and waloria tests.

ccrs:
rewrite it, no more hacks

r00ty-tc:
@ccrs the thing is, any proper rewrite will still need this, or some way to store the state on grid unload. 

I think, this should be enabled if "Wintergrasp.Enable = 1" is set in the config. E.g. if wintergrasp is enabled then enable the persistent grids.

Yeah, the whole thing needs rewriting. But, this change would be part of a rewrite I think anyway.

Foereaper:
The only other way to deal with this would be:

- store damage states of every single game game object at the end of every single battle. This would have to be saved to the database in case of a core crash or restart. 
- on grid load, check every single object in the grid to make sure all the objects are spawned, and if they are not, respawn the object and reapply their damage states. 

I'm sure there's more that would have to be taken into consideration as well, however I'm not sure what would be less of a hack to be honest. 

Like r00ty said, ensuring the grids are loaded is most likely going to be part of a larger rewrite regardless.

ccrs:
Ill commit a rwrite when im done with movement without locking any grid

Foereaper:
Alright :)

Adizbek:
There is branch. __3.3.5-battleground__. It is safe to merge it with __3.3.5__ ?

ccrs:
no

Shoxxo:
Expect us WHY no and don't close without any reason. If you say NO, WHy no, please more details. Here for learning, not for NO

ccrs:
ROFL just shut up #20144

Shoxxo:
That is not friendly. You should leaen respect other users and comments internetprollo!!!

tkrokli:
His "no" answer was not disrespectful, it was a simple answer to the question
" 3.3.5-battleground. It is safe to merge it with 3.3.5 ? "

PR #20146, opened 2017-08-14
Scripts/ICC: Lord Marrowgar - Coldflame Resistance

Demonid:
**Changes proposed:**

-  Fix Resistance for Coldflame

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes https://github.com/TrinityCore/TrinityCore/issues/13240

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -3898,6 +3898,12 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->Effects[EFFECT_0].RadiusEntry = sSpellRadiusStore.LookupEntry(EFFECT_RADIUS_100_YARDS); // 100yd
     });
 
+    // Coldflame (Lord Marrowgar)
+    ApplySpellFix({ 69146, 70823, 70824, 70825 }, [](SpellInfo* spellInfo)
+    {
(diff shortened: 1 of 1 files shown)


Keader:
Sources? i'm not sure if issue is valid.

offl:
> The second ability is an ability called Cold Flame. Lord Marrowgar will shoot out the Cold Flame against a target and the Cold Flame will go in a straight line until gone. If you stand in the fire you take 6000 frost damage every second for three seconds. The Cold Flame will start at Lord Marrowgar and go towards it’s target, so melee need to be careful with the flame as it may start right on them even if it’s going for another target. **It’s adviced to use Frost Resistance Aura or Frost Resistance Totem**. 

[Source](http://www.ownedcore.com/forums/world-of-warcraft/world-of-warcraft-guides/273775-icecrown-citadel-instance-tactics-explanation-etc-questions-allowed.html). But i'm not sure if it is a good source.

Demonid:
In this guide from 2009 from Tankspot: the video says that is a good idea to run frost resistance:

https://www.youtube.com/watch?v=RSbBt9NHsXE&feature=youtu.be&t=1m54s

In this other video from 2009, you can see the tank getting hit multiple times from Coldflame on MSBT, it clearly shows part of the damage being resisted.

https://www.youtube.com/watch?v=YUlvH8n_Zwg

Shauren:
Yep, definitely getting resited there on these videos

Issue #7266, opened 2012-08-03
Val'kyr Shadowguard


siley:
http://www.wowhead.com/npc=36609

DB: TDB 48
core: https://github.com/TrinityCore/TrinityCore/commit/8e35762b35a07ca3ad096b6430962104d0b97dd4

after stun, root, freeze, slow etc Val'kyr drop player and despawn instead of continue in to move to border of platform
##

sirikfoll:
This can be pushed?
```sql
-- Valkyr Shadowguard Imunity
UPDATE `creature_template` SET `mechanic_immune_mask` = 
1|          -- charm
2|          -- disorient
4|          -- disarm
8|          -- distract
16|         -- fear
32|         -- grip
64|         -- root
256|        -- silence
512|        -- sleep
4096|       -- freeze
8192|       -- knockout
65536|      -- polymorph
131072|     -- banish
524288|     -- shackle
1048576|    -- mount
4194304|    -- turn
8388608|    -- horror
33554432|   -- interrupt
536870912   -- sapped
WHERE `entry` IN (36609,39120,39121,39122);

Jenkyins:
confirmed on  ae1a5c6c2b82918850cb7b851e2f7dff33a40c57

lashia:
can anyone share hack fix for this

tkrokli:
No. Such things belong in the forum, in the section for custom code.

Aokromes:
If anything here is valid, plz open a pr for this.

Aokromes:
This bug/fix is still valid after fe68b2e99851a888c5182d7bebddfea67459afd3 ?

Keader:
@Kittnz , @Killyana  what about https://github.com/TrinityCore/TrinityCore/issues/7266#issuecomment-174006201 ?

Rushor:
according to http://www.wowhead.com/npc=36609/valkyr-shadowguard#comments:id=987567, those posted immunities are fine :)

Noryad:
Confirmed on d939018.

Noryad:
This bugs:
1. Sometimes Valkyr Shadowguard "blink" when get stun.
2. Valkyr Shadowguard dropping the targets.

Both caused by a movement generator issue.

sirikfoll:
About the bug with the spell Charge, it's a core issue in
[...]

PR #20147, opened 2017-08-14
Core/Scripts: More fixes in boss Lich King

sirikfoll:
**Changes proposed:**
- Fix Valkyr charge spell, they will no longer ignore Z position and become unreachable.
- Correct height of Spirit Bomb and added the 3 seconds delay on his explosion.
- Set the Trigger inside frostmourne room as active, to avoid problems with reseting the room, e.g: Spirit Bombs not despawning and Wicked Spirits stuck in evade.
- Use correct InhabitType for Wicked and Vile Spirits, this avoid a situation where they could spawn falling.

- [x] 3.3.5
- [x] master

**Issues addressed:** Updates #7266

**Tests performed:** Built and tested during boss fight, with 2 players

### DIFF
--- a/sql/updates/world/3.3.5/2017_08_17_00_world.sql
+++ b/sql/updates/world/3.3.5/2017_08_17_00_world.sql
@@ -1,1 +1,2 @@
+-- Vile Spirit + (1),(2),(3) / Wicked Spirit + (1),(2),(3)
 UPDATE `creature_template` SET `InhabitType`=4 WHERE `entry` IN (39190,39287,39288,39289,37799,39284,39285,39286);
--- a/sql/updates/world/3.3.5/9999_99_99_99_world.sql
+++ b/sql/updates/world/3.3.5/9999_99_99_99_world.sql
@@ -1,0 +1,1 @@
+UPDATE `creature_template` SET `InhabitType`=4 WHERE `entry` IN (39190,39287,39288,39289,37799,39284,39285,39286);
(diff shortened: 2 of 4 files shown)


sirikfoll:
If someone remembers any other problem related with this fight, please let me know.

Killyana:
If you stunn a Val'kyr it will teleport some yards forward when the stunn effect fades.

sirikfoll:
@Killyana  yeah, same as #7266, that one I can't fix. if I remember right, is something with TargetedMovementGenerator and chaodhib was working on it.

Keader:
@ccrs valkyr teleport was not fixed?

Keader (approved):
Looks good to me 👍

tkrokli:
Maybe throw in a comment line above or behind the SQL command? Example:
```sql
-- Vile Spirit + (1),(2),(3) / Wicked Spirit + (1),(2),(3)
UPDATE `creature_template` SET `InhabitType`=4 WHERE `entry` IN (39190,39287,39288,39289,37799,39284,39285,39286);
```

Shauren:
Need to check sniff for spirit movement flags

Shauren:
Right so - the spirits have these flags on spawn according to old 3.3.0 sniff: None (0) but SMSG_MONSTER_MOVE has Spline Flags: Flying
And in 4.3 sniff they have Movement Flags: DisableGravity (512)

Which is InhabitType = 4

PR #20152, opened 2017-08-17
Core/Misc: Several gold handling improvements

### LINKED ISSUE #15936: [3.3.5] Core/Guild Vault: Deposit resets to 0 after it exceeds (2^64 − 1) copper
Revision: https://github.com/TrinityCore/TrinityCore/commit/c703e01471a3598f68051abb17e598e316b744ef

If your Guild Vault has (2^64 − 1) copper deposited (just set it to (2^64 − 1) in Database - Characters - Guild - BankMoney) and you go and deposit 1 copper, it will reset back to 0.

No Guild Vaults will have such tons of gold in reality, but maybe (just in case of a fun server) it should tell you that you are at the gold limit (limited by game client).

### LINKED ISSUE #15937: [3.3.5] Core/Guild Vault: Deposit is not displayed correctly if higher than (2^51 - 1) copper
Revision: https://github.com/TrinityCore/TrinityCore/commit/c703e01471a3598f68051abb17e598e316b744ef

If your Guild Vault has a deposit bigger than (2^51 -1) copper, you have to deposit or withdraw any amount of gold / silver / copper in order to see the REAL deposit.

wowscrnshot_120315_141916
wowscrnshot_120315_141923
wowscrnshot_120315_141927

### LINKED ISSUE #19195: [master] Auction House

**Description** Auction House can not be used to place auctions

**Current behaviour:** You can't create an auction in any Auction House.

**Expected behaviour:**
You should be able to create auction in any auction house and other players should be able to view and buy them.

1. Speak with any auctioneer NPC
2. Press the 'Auctions' tab
3. Drag any sellable item to the Auction item slot
4. Press the 'Create Auction' button
5. Notice that the auction is not created.

**Branch(es):** master

**TC rev. hash/commit:** 7081201c9160+

**TDB version:**  703.00 + updates

**Operating system:**  Windows 7 64-bits

roc13x:
**Changes proposed:**
- Update gold limit to 10 million
- Updated auction house gold handling to 64-bit
- Fixed some vendor checks for when BuyCount > 1
- Tweaked some checks for available gold space
- Updated guild bank gold handling to 64-bit and fix withdraw limits
- Enforce gold limit on guild bank
- Check correct rank right when withdrawing from bank
- Other small changes to update money to 64-bit

**Target branch(es):**
- [x] master

**Issues addressed:**
Closes #19195
Closes #15936

- Compiles
- Tested all updated code in-game


src/server/game/Entities/Player/Player.cpp
```diff
@@ -22425,13 +22425,20 @@ bool Player::BuyItemFromVendorSlot(ObjectGuid vendorguid, uint32 vendorslot, uin
...
-        if (creature->GetVendorItemCurrentCount(crItem) < pProto->GetBuyCount() * count)
+        if (creature->GetVendorItemCurrentCount(crItem) < count)
         {
             SendBuyError(BUY_ERR_ITEM_ALREADY_SOLD, creature, item, 0);
             return false;
         }
     }
 
+    // Can only buy full stacks
+    if (count % pProto->GetBuyCount())
```
Shauren: This is false.
Take http://www.wowhead.com/item=3371/crystal-vial for example. It is sold in stacks of 20 but you can buy only 1 on retail
Only things bought with special currencies cannot be divided

Shauren:
Does not fix #15937 (unfixable, lua limitation)

Are all fields in packets for things you changed 64bit ints? There still are some packets sending gold as 32bit (strange but true)

Shauren (changes requested):

src/server/game/Entities/Player/Player.cpp
roc13x: I know you definitely used to be able to do that on previous versions. I tried on the 725 client and shift-clicking would not let me buy 1. Did it work for you?

roc13x:
Sorry, was a bit hasty tagging that issue, removed ^^
Yea I made sure not to change fields that were still being sent as 32bit. Rank money per day is still 32bit

src/server/game/Entities/Player/Player.cpp
Shauren: It does not work with default UI but you can do this with addons - I use https://mods.curse.com/addons/wow/gnomishvendorshrinker (heavy warning - might not work on 7.2.5 without modifications, source link https://github.com/TekNoLogic/GnomishVendorShrinker/commits/master)

PR #20201, opened 2017-08-26
Core/Entities: add SellResult "You cannot sell items to this merchant" for Ethereal Soul-Trader

toto6:
**Changes proposed:**

-  Add SellResult "You cannot sell items to this merchant"
-  Use it for NPC Ethereal Soul-Trader 27914

- 3.3.5

### DIFF
--- a/src/server/game/Entities/Item/ItemDefines.h
+++ b/src/server/game/Entities/Item/ItemDefines.h
@@ -128,12 +128,13 @@ enum BuyResult
 
 enum SellResult
 {
-    SELL_ERR_CANT_FIND_ITEM                      = 1,
-    SELL_ERR_CANT_SELL_ITEM                      = 2,       // merchant doesn't like that item
-    SELL_ERR_CANT_FIND_VENDOR                    = 3,       // merchant doesn't like you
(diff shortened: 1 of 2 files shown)


nawuko:
There is a good list of npc that you cannot sell to in WoW (altho its only 1 in 3.3.5 to my knowledge) Adding a check like this is bad practice. Maybe a custom flag or something would be better. @Shauren could give advise here.

Faq:
Anyone has idea how this should be done? Scripted or with flag? @Shauren

Shauren:
I think a flag on the vendor would work

jackpoz:
anyone up to implement the flag ?

ForesterDev:
I implement the feature this way

if (CreatureTemplate const* creatureInfo = creature->GetCreatureTemplate())
    {
        if (creatureInfo->type == CREATURE_TYPE_NON_COMBAT_PET)
        {
            _player->SendSellError(SELL_ERR_CANT_SELL_TO_THIS_MERCHANT, creature, itemguid, 0);
            return;
        }
    }

Shauren:
That would break argent gruntling pet and would allow selling items to http://www.wowhead.com/npc=111737/celadine-the-fatekeeper

ariel-:
Superseded by #21642

PR #20202, opened 2017-08-26
Core/Unit: make neutral guards attack players involved in PvP

### LINKED ISSUE #17635: Core/unit: Guards no longer attaking players
**Description**:
Guards like http://www.wowhead.com/npc=3502/ratchet-bruiser are supposed to attacks any player that hit any other player, but with the recent changes, they just turn aggressive, but they doesn't attack and they cannot be attacked (wrong target as message)

1.  connect 2 characters horde and alliance
2.  .go c id 19488
3.  hit the player in front of you and notice the react of the guard

**Branch(es)**: 3.3.5 / 6.x 

**TC hash/commit**:  rev. cdf8f3c2e1f0

toto6:
**Changes proposed:**
-  Fix neutral guards not attacking players involved in PvP

- [x] 3.3.5

closes #17635

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8923,6 +8923,9 @@ bool Unit::_IsValidAttackTarget(Unit const* target, SpellInfo const* bySpell, Wo
         Player const* player = playerAffectingAttacker ? playerAffectingAttacker : playerAffectingTarget;
         Unit const* creature = playerAffectingAttacker ? target : this;
 
+        if (creature->IsContestedGuard() && player->HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_CONTESTED_PVP))
+            return true;
+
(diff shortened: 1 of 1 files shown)


Faq:
@Shauren is this correct way to do it?

jackpoz:
could you please explain more in detail how your change fixes the issue ? there is very little in the PR description or in the code about what's happening there.

Killyana:
I just tested it, it fixes the issue.

aje:
before this PR, there was this issue https://github.com/TrinityCore/TrinityCore/issues/17635 : after having been involved in PvP, neutral guards won't attack you, and you won't be able to attack neutral guards.
I think the issue was because the function IsValidAttackTarget returned False because the player had reputation >= to neutral with the faction of the guards.
So I added a check to see if the NPC is a guard and if the player has been involved in PvP before the faction check. If this is the case, then IsValidAttackTarget returns true

tkrokli:
Good description. Belongs in the PR description (original post) as well. :-)

PR #20235, opened 2017-09-01
Enable CMSG_CLOSE_INTERACTION

Rochet2:
Coded in https://github.com/TrinityCore/TrinityCore/pull/20085

**Changes proposed:**

- Enable already coded opcode
- https://github.com/TrinityCore/TrinityCore/commit/e5b9c207f8889010c88f66ea39160659137402d3#commitcomment-24025415

- [ ] 3.3.5
- [x] master

- [ ] I have tried with STATUS_LOGGEDIN. Only one problem when you speak with a NPC (open trainer for example) and after, speak with another NPC (in same range) without close your actual window (trainer), the second npc do not react (close only your actual windows but don't open the new). You need to go out of range interact for this to work

### DIFF
--- a/src/server/game/Server/Protocol/Opcodes.cpp
+++ b/src/server/game/Server/Protocol/Opcodes.cpp
@@ -323,7 +323,7 @@ void OpcodeTable::Initialize()
     DEFINE_HANDLER(CMSG_CLEAR_RAID_MARKER,                                  STATUS_LOGGEDIN,  PROCESS_THREADUNSAFE, &WorldSession::HandleClearRaidMarker);
     DEFINE_HANDLER(CMSG_CLEAR_TRADE_ITEM,                                   STATUS_LOGGEDIN,  PROCESS_THREADUNSAFE, &WorldSession::HandleClearTradeItemOpcode);
     DEFINE_HANDLER(CMSG_CLIENT_PORT_GRAVEYARD,                              STATUS_LOGGEDIN,  PROCESS_THREADUNSAFE, &WorldSession::HandlePortGraveyard);
-    DEFINE_HANDLER(CMSG_CLOSE_INTERACTION,                                  STATUS_UNHANDLED, PROCESS_THREADSAFE,   &WorldSession::HandleCloseInteraction);
+    DEFINE_HANDLER(CMSG_CLOSE_INTERACTION,                                  STATUS_LOGGEDIN, PROCESS_THREADSAFE,   &WorldSession::HandleCloseInteraction);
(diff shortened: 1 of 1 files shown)


matukaa:
Please maintain whitespace :)

src/server/game/Server/Protocol/Opcodes.cpp
```diff
@@ -323,7 +323,7 @@ void OpcodeTable::Initialize()
     DEFINE_HANDLER(CMSG_CLEAR_RAID_MARKER,                                  STATUS_LOGGEDIN,  PROCESS_THREADUNSAFE, &WorldSession::HandleClearRaidMarker);
     DEFINE_HANDLER(CMSG_CLEAR_TRADE_ITEM,                                   STATUS_LOGGEDIN,  PROCESS_THREADUNSAFE, &WorldSession::HandleClearTradeItemOpcode);
     DEFINE_HANDLER(CMSG_CLIENT_PORT_GRAVEYARD,                              STATUS_LOGGEDIN,  PROCESS_THREADUNSAFE, &WorldSession::HandlePortGraveyard);
-    DEFINE_HANDLER(CMSG_CLOSE_INTERACTION,                                  STATUS_UNHANDLED, PROCESS_THREADSAFE,   &WorldSession::HandleCloseInteraction);
+    DEFINE_HANDLER(CMSG_CLOSE_INTERACTION,                                  STATUS_LOGGEDIN, PROCESS_THREADSAFE,   &WorldSession::HandleCloseInteraction);
```
joschiwald: change `PROCESS_THREADSAFE` to `PROCESS_THREADUNSAFE` too and it can fix your not responding npcs, but anyway in that packet we should only reset the saved guid if it matches the sent guid

DoctorKraft: `PROCESS_THREADUNSAFE` don't fix this. Remember: we need a trainer with gossip for try it (not with creature_default_trainer)

Rochet2: Can you post an entry of such a creature if one exists? to test with.

DoctorKraft: https://gist.github.com/DoctorKraft/a9b6562c838e2f23a0ae34c59232ca8f

joschiwald: trainerlist, gossiphello, gossipselect are processes threadunsafe, this opcode must also processed threadunsafe to ensure the right order of received packets like the client sends them, closeinteraction before gossiphello and others

PR #20245, opened 2017-09-05
Fix quest "Rescue the survivors" in Draenei starting zone.

### LINKED ISSUE #20240: [Master] Quest Rescue the survivors in Draenei start zone is broken.
**Description:**

Quest 9283 Rescue the survivor has script in azuremyst_isle.cpp and link in DB for Draenei survivor id 16483.

http://www.wowhead.com/quest=9283/rescue-the-survivors

**Current behaviour:**

Quest dont work even with all data in core and DB

**Expected behaviour:**

Complete the quest when you use gift of the Naaru on Draenei Survivor.

1- Create a Draenei
2- Add quest 9283 with level 2
3- Try to complete the quest

**Branch(es):**

Master

**TC rev. hash/commit:** 

52c7f12ed35f321dc97d8caea2b112287f5ac05b

HannibalRoG:
**Changes proposed:**

-  Restrict quest "Rescue the survivor" to spell Gift of the Naaru.

- [ ] 3.3.5
- [x] master

closes #20240

Compiled, tested in game.

### DIFF
--- a/src/server/scripts/Kalimdor/zone_azuremyst_isle.cpp
+++ b/src/server/scripts/Kalimdor/zone_azuremyst_isle.cpp
@@ -51,7 +51,6 @@ enum draeneiSurvivor
     SAY_THANK_FOR_HEAL     = 0,
     SAY_ASK_FOR_HELP       = 1,
     SPELL_IRRIDATION       = 35046,
-    SPELL_GIFT_OF_THE_NAARU = 59544,
     SPELL_STUNNED          = 28630,
     EVENT_CAN_ASK_FOR_HELP = 1,
     EVENT_THANK_PLAYER     = 2,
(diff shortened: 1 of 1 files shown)


src/server/scripts/Kalimdor/zone_azuremyst_isle.cpp
```diff
@@ -51,6 +51,7 @@ enum draeneiSurvivor
     SAY_THANK_FOR_HEAL     = 0,
     SAY_ASK_FOR_HELP       = 1,
     SPELL_IRRIDATION       = 35046,
+    SPELL_GIFT_OF_THE_NAARU = 59544,
```
joschiwald: each class has its own "Gift of the Naaru" spell

HannibalRoG: Well, i dont know why but it work with every class.

HannibalRoG:
I found what happend, the spell i use is SPELLFAMILY_GENERIC and affect all draenei racial.
Now i think the best way to fix this is change 
`spell->SpellFamilyFlags[2]` & 0x080000000 && !_tappedBySpell
BY
`spell->SpellFamilyFlags[2]` & 0x80000000 && !_tappedBySpell

Shauren:
Your change does nothing, thats the same number...
Also, thats not why this quest is broken (and its only broken for priests and monks)
It does not work for these two classes is because their version of Gift of the Naaru has no value in `SpellFamilyFlags`

HannibalRoG:
THANKS @Shauren

Issue #20287, opened 2017-09-09
[3.3.5] Dynamic Spawn vs DespawnAtEvade


Keader:
**Description:** Dynamic Spawn show some conflict with DespawnAtEvade method. When creature respawn with Dynamic Spawn system, a new script is loaded (is a new creature !).
But some events triggered by areatriggers should occour just 1 time. Other bosses should have same issue related with events that should happen 1 time only.
Normally was used class constructor to handle with it, but since creature load a new script it's not possible anymore (constructor is "called again")

**Current behaviour:** DespawnAtEvade kill boss, boss respawns with a new script, areatrigger events is called again.

**Expected behaviour:** Areatrigger events called 1 time only.

**Steps to reproduce the problem:**

1.Go to Lady Deathwhisper, Teron Gorefiend or any boss with areatrigger + DespawnAtEvade
2.  Pass to areatriggers (normally in main door).
3. Start fight/Evade, pass to areatrigger again

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 4d654f548b29bf61552d9f90da8254ecea91b63e

1 suggested solution for it (ccrs solution) is store this "events" in instance_script.

(issue created to remember of this problem xd)

Shauren:
Change script to store state in instance script

Keader:
971ed856a49dc56ea33d9a0779b1386eea5cd3ca

PR #20288, opened 2017-09-09
Core/Scripts: Implemented OnlyOnceAreaTriggerScript

Treeston:
Tagging #20287, #20306

To transition:
* Inherit from `OnlyOnceAreaTriggerScript` instead of `AreaTriggerScript`
* Rename `OnTrigger` to `_OnTrigger`


Krudor:
I'm all for improving the scripting experience, but personally don't think manually storing/getting state from instancescript is that much of a hassle for the cases where you need to trigger once.

Treeston:
Neither is adding a new implementing subclass (and it saves us from duplicating the same code 500 times).

Krudor:
Yep, fully aware.
I like the idea, just don't like the implementation.

Ovahlord:
This implementation is wrong anyways if it is not stored in the instance data. Cuz this would lead to an reset of all such areatriggers when the server crashes / gets restarted.
You should instead extend the exsiting area trigger class to add a "trigger once" variable and add a optional instance data overload to handle things automated.

Treeston:
Greeting areatriggers are reset on instance reload on retail, too. This is intentional.

Ovahlord:
What about areatriggers that are handling events that should not repeat, once they are done?

ccrs:
I agree with Treeston, the less code duplication the better.
Though the option to save the state to DB should be available.

Treeston:
@Ovahlord those already use boss states (see ICC events)

ccrs:
And using boss states is completely wrong, Svala event for example. I'll commit a refactor/rewrite/we soon (TM). Dont use boss states to store event states.

Keader:
Ok, PR is ready for feedbacks :)

Keader:
@Treeston can be merged? :)

PR #20325, opened 2017-09-15
Core/Items: Implement ItemLevelSelector.db2

roc13x:
**Changes proposed:**
- Implement ItemLevelSelector.db2
- Use the ItemLevelSelectorID field in bonus tree nodes
- Remove the functionality of ITEM_BONUS_ITEM_LEVEL_OVERRIDE. Client ignores this since 7.2

**Target branch(es):**
- [x] master

**Tests performed:**
- Built
- Tested with instance loot


src/server/game/DataStores/DB2Stores.cpp
```diff
@@ -1510,8 +1525,27 @@ std::set<uint32> DB2Manager::GetItemBonusTree(uint32 itemId, uint32 itemBonusTre
...
+            }
+            else if (bonusTreeNode->ItemLevelSelectorID)
+            {
+                ItemLevelSelectorEntry const* selector = sItemLevelSelectorStore.LookupEntry(bonusTreeNode->ItemLevelSelectorID);
+                if (!selector)
+                    continue;
+
+                int16 levelChange = int16(selector->ItemLevel) - proto->ItemLevel;
+                
+                auto itr = _selectorToBonus.find(levelChange);
```
Shauren: This is wrong. You are supposed to use ItemBonusListLevelDelta here (no need for custom filled container based on hardcoded ids)

src/server/game/DataStores/DB2Stores.h
```diff
@@ -118,6 +118,7 @@ TC_GAME_API extern DB2Storage<ItemDisenchantLootEntry>              sItemDisench
 TC_GAME_API extern DB2Storage<ItemEffectEntry>                      sItemEffectStore;
 TC_GAME_API extern DB2Storage<ItemEntry>                            sItemStore;
 TC_GAME_API extern DB2Storage<ItemExtendedCostEntry>                sItemExtendedCostStore;
+TC_GAME_API extern DB2Storage<ItemLevelSelectorEntry>               sItemLevelSelectorStore;
```
Shauren: Not used outside of DB2Stores.cpp, don't declare.

src/server/game/DataStores/DBCEnums.h
```diff
@@ -713,7 +713,7 @@ enum ItemBonusType
     ITEM_BONUS_SCALING_STAT_DISTRIBUTION    = 11,
     ITEM_BONUS_DISENCHANT_LOOT_ID           = 12,
     ITEM_BONUS_SCALING_STAT_DISTRIBUTION_2  = 13,
-    ITEM_BONUS_ITEM_LEVEL_OVERRIDE          = 14,
+    ITEM_BONUS_ITEM_LEVEL_OVERRIDE          = 14,                 // Functionality removed in 7.2.0
```
Shauren: The comment is also incorrect - functionality isn't entirely removed, it still makes the client display '+' sign next to item level when viewing tooltip (in quest rewards or dungeon journal loot)

Shauren (changes requested):

src/server/game/DataStores/DB2Stores.cpp
roc13x: Ah damn, no idea how I missed that file. I wasn't happy with _selectorToBonus at all, so thanks for the info!

src/server/game/DataStores/DBCEnums.h
roc13x: I see, so it's only a client-side visual flag now. Should it be renamed or just change the comment?

Shauren: Rename it

roc13x:
Requested changes added

Shauren (approved):
pretty good

PR #20418, opened 2017-09-25
[3.3.5] Scripts/Quest: A Rogue's Deal improvements

Wyreth:
**Changes proposed:**

- Make the event for the quest [A Rogue's Deal](http://www.wowhead.com/quest=590) more blizzlike, with correct timers, creature text and emotes.

Used [this video](https://youtu.be/5CjYDkAUPy0?t=173) as reference.

**Target branch(es):**

- [x] 3.3.5


tkrokli:
I noticed a detail on the page http://www.wowhead.com/spell=7737/food regarding the spell you have chosen for the NPC when drinking:
`Can only be used during nighttime` -- do we know if Deathknell is in a perpetual "night time" for this to work 24/7 ?
What about other drink spells from the list of spells from 3.3.5 or earlier:
http://www.wowhead.com/uncategorized-spells/name:drink/live-only:on?filter=21;4;30305
- http://www.wowhead.com/spell=46583/cosmetic-drink
- http://www.wowhead.com/spell=42871/cosmetic-drink-tankard

Would any of those spells be more correct?

Wyreth:
@tkrokli: that one is the only drink spell lasting 15 seconds that I could find. It fits perfectly with the event timing shown in the video.

I think the "nighttime only" is a mistake - if you check the items that use the spell, there's stuff like [Raw Brilliant Smallfish](http://www.wowhead.com/item=6291) - that certainly can be used at any time of the day :)

tkrokli:
Hehe, very well. The only hunch I have got left is that maybe one of the other spells could have lasted the correct time if normal player eat/drink animation would last as long as the spell duration, which never seems to be the case in TC. I think the player animation some time long ago lasted as long as the actual spell duration on Blizzard servers. (I wonder if it has ever worked at all or if it is a core issue that will never be solved in the TC source, but that is a different matter entirely.)

Wyreth:
Got to check that one again. I recall in TC the animation (sparkles) lasts correctly, but the character remains sitting in the ground until you move or stand manually.

I don't recall how it was handled on retail. So many small details I'd like to have taken the time to write down back then :P

src/server/scripts/EasternKingdoms/zone_tirisfal_glades.cpp
```diff
@@ -54,82 +60,59 @@ class npc_calvin_montague : public CreatureScript
...
+                        _events.ScheduleEvent(EVENT_DRINK, Seconds(5));
                         break;
-                    case 2:
-                        if (Player* player = ObjectAccessor::GetPlayer(*me, m_uiPlayerGUID))
+                    case EVENT_DRINK:
+                        if (Player* player = ObjectAccessor::GetPlayer(*me, _playerGUID))
                             player->AreaExploredOrEventHappens(QUEST_590);
-
+                        _playerGUID.Clear();
                         DoCast(me, SPELL_DRINK, true);
```
Keader: DoCastSelf(SPELL_DRINK, true);

src/server/scripts/EasternKingdoms/zone_tirisfal_glades.cpp
```diff
@@ -54,82 +60,59 @@ class npc_calvin_montague : public CreatureScript
...
-                        EnterEvadeMode();
+                    case EVENT_STAND:
+                        me->SetStandState(UNIT_STAND_STATE_STAND);
                         break;
                 }
-
-                return;
             }
 
             if (!UpdateVictim())
```
Keader: Should be on top, no? (before events update)

Keader (commented):
Please check it

src/server/scripts/EasternKingdoms/zone_tirisfal_glades.cpp
Wyreth: Events should be executed when out of combat. Some other scripts have this before DoMeleeAttackIfReady(), so I left it there.

src/server/scripts/EasternKingdoms/zone_tirisfal_glades.cpp
Wyreth: Removing the triggered parameter altogether, not needed here. Thanks :)

src/server/scripts/EasternKingdoms/zone_tirisfal_glades.cpp
Keader: You can control it using "events.IsInPhase" like: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Outland/BlackTemple/boss_illidan.cpp#L779

Wyreth: I'm not sure I understand, sorry :P

This isn't a boss, this is a simple quest script for a NPC. There's no phase, no complex events or anything else. It should just do something when out of combat, and do melee attacks in combat. Stop.

All other similar scripts use the same structure - why is this one any different?

Keader: Normally ccrs complain alot because of it.
But i agree with you btw....
It is tested, can be merged already?

Wyreth: Yep, tested and working. I only have the [video](https://www.youtube.com/watch?v=5CjYDkAUPy0&feature=youtu.be&t=173) for reference, and with those fixes the quest works exactly as seen there.

I'd say it can be merged :)

Keader:
Ty

PR #20420, opened 2017-09-25
Core/SAI: remove double calls to SMART_EVENT_RESPAWN for creatures and gameobjects

Wyreth:
**Changes proposed:**

-  SMART_EVENT_RESPAWN is already called in the respective InitializeAI() methods for both creatures and gameobjects. The other existing calls are wrongly executed on reset. Since there's already SMART_EVENT_RESET, this double script call is confusing and harmful for scripts that must execute only at respawn.

**Target branch(es):**

- [x] 3.3.5

**Issues addressed:** I think I saw an issue related to this somewhere... but can't find it now.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -552,6 +552,7 @@ void SmartAI::JustAppeared()
         me->RestoreFaction();
     mJustReset = true;
     JustReachedHome();
+    GetScript()->ProcessEventsFor(SMART_EVENT_RESPAWN);
     mFollowGuid.Clear(); // do not reset follower on Reset(), we need it after combat evade
     mFollowDist = 0;
     mFollowAngle = 0;
@@ -694,10 +695,7 @@ void SmartAI::InitializeAI()
     GetScript()->OnInitialize(me);
 
     if (!me->isDead())
-    {
         GetScript()->OnReset();
-        GetScript()->ProcessEventsFor(SMART_EVENT_RESPAWN);
-    }
 }
 
 void SmartAI::OnCharmed(bool apply)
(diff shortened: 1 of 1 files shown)


Wyreth:
Any update for this one?

Shauren:
I am not sure removing it from JustAppeared is the right thing to do. Legacy respawn mode is still supported after all

Wyreth:
Tested with creatures set to the Legacy spawn group, it still works as expected (a new creature object is created, so it goes through initialization again).

Is there any other test I can try?

Shauren:
Wait what? legacy respawn is the case when old corpse is reused

Wyreth:
I thought that "reuse old corpse" behaviour was completely removed and "legacy" just meant "don't calculate respawn time dynamically". Can you tell me how to enable it for a creature? Tried playing with the spawn_group tables but no luck.

tkrokli:
Maybe @r00ty-tc or @Treeston can reply/comment on that.

Treeston:
Just add them to the legacy ("compatibility mode") spawn group. Dynamic respawn time is unrelated (requires the creature to be in a group with dynamic flag).

See [SpawnGroupFlags](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Maps/SpawnData.h#L39-L49).

Wyreth:
Got it to work, thanks! Removed the calls in InitializeAI() instead, so it behaves correctly for both spawn types.

Treeston:
Just check for compatibility mode in the `Reset` call?

Wyreth:
`Reset` is called at spawn for both modes, so there's no need for further checks I think? I might be wrong, but this way works for everything so far.

Treeston:
Actually isn't there a `JustSpawned` hook...? I think we addedone

Wyreth:
There's `JustAppeared`. I meant to say that one, not Reset. Sorry.

Treeston:
Alright, `JustAppeared` is the correct place to put it.

Wyreth:
Nice :) with this change SMART_EVENT_RESPAWN is called on `SmartAI::JustAppeared` for creatures and `SmartGameObjectAI::Reset` for gameobjects. The other two calls are harmful in that they cause double execution of scripts.

Hopefully this covers every case (that I could test, at least).

ccrs (approved):

Treeston (approved):

tkrokli:
Are you satisfied with the current content, that you don't need to make any further changes?

Aokromes:
Plz, don't add "[3.3.5]" or [master] to prs, it needs to rename before merge, we can see targeted branch on pr target:
Wyreth  wants to merge 2 commits into TrinityCore:3.3.5 from Wyreth:sai_action_respawn

Wyreth:
Alright, I'll avoid that from now on :D

Shauren (approved):

PR #20462, opened 2017-09-28
Scripts/Northrend: fix event for quest A Suitable Test Subject

Wyreth:
**Changes proposed:**

When using the quest item for [this quest](http://www.wowhead.com/quest=11719), when the spell aura vanishes, the nearby questgiver should turn toward the player and whisper a line.

This is handled via spell event for [this spell](http://www.wowhead.com/spell=45997).

[Video here](https://www.youtube.com/watch?v=oH_CGGqolns).

**Target branch(es):**

- [x] 3.3.5
- [x] master

### DIFF
--- a/sql/updates/world/3.3.5/2017_09_28_99_world_bloodspore_ruination.sql
+++ b/sql/updates/world/3.3.5/2017_09_28_99_world_bloodspore_ruination.sql
@@ -1,5 +1,5 @@
 --
-DELETE FROM `spell_script_names` WHERE `spell_id`=45997;
+DELETE FROM `spell_script_names` WHERE `ScriptName`="spell_q11719_bloodspore_ruination_45997";
 INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
 (45997, "spell_q11719_bloodspore_ruination_45997");
 UPDATE `creature_template` SET `ScriptName`="npc_bloodmage_laurith" WHERE `entry`=25381;
--- a/sql/updates/world/3.3.5/2017_09_28_99_world_bloodspore_ruination.sql
+++ b/sql/updates/world/3.3.5/2017_09_28_99_world_bloodspore_ruination.sql
@@ -1,0 +1,8 @@
+--
+DELETE FROM `spell_script_names` WHERE `spell_id`=45997;
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
+(45997, "spell_q11719_bloodspore_ruination_45997");
+UPDATE `creature_template` SET `ScriptName`="npc_bloodmage_laurith" WHERE `entry`=25381;
+DELETE FROM `creature_text` WHERE `CreatureID`=25381;
+INSERT INTO `creature_text` (`CreatureID`, `GroupID`, `ID`, `Text`, `Type`, `Language`, `Probability`, `Emote`, `Duration`, `Sound`, `BroadcastTextId`, `TextRange`, `comment`) VALUES
+(25381, 0, 0, "How positively awful! You were totally incapacitated? Weak? Hot flashes?", 15, 0, 100, 21, 0, 0, 24992, 0, "Bloodmage Laurith");
--- a/src/server/scripts/Northrend/zone_borean_tundra.cpp
+++ b/src/server/scripts/Northrend/zone_borean_tundra.cpp
@@ -2384,6 +2384,103 @@ public:
     }
 };
 
+enum BloodsporeRuination
+{
+    NPC_BLOODMAGE_LAURITH = 25381,
+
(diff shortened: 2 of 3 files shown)


ccrs:
think about multiple people triggering the event at the same time or while its running

src/server/scripts/Northrend/zone_borean_tundra.cpp
```diff
@@ -2384,6 +2384,103 @@ class spell_windsoul_totem_aura : public SpellScriptLoader
...
+                            Talk(0, player);
+                        _events.ScheduleEvent(EVENT_RESET_ORIENTATION, Seconds(5));
+                        break;
+                    case EVENT_RESET_ORIENTATION:
+                        me->SetFacingTo(me->GetHomePosition().GetOrientation());
+                        break;
+                }
+            }
+
+            if (!UpdateVictim())
```
ccrs: move this up
you dont want texts running while the creature is in combat, right?

Wyreth: The text is a whisper, I don't think combat makes any difference with that. Can't check on retail, though.

Wyreth:
Should probably just whisper everyone involved in this case. I know the questgiver npcflag is never removed for this NPC, so that's what I guess would happen on retail.

Faq:
so ready for merge? :)

tkrokli:
If you have tested it and it works for you and others (especially with more than one player),
then yes, certainly. Feedback from testing is always welcome.

Wyreth:
Works for me. It's not outdated, so that label is wrong :P

Aokromes:
@Keader this can be merged?

src/server/scripts/Northrend/zone_borean_tundra.cpp
```diff
@@ -2384,6 +2384,108 @@ class spell_windsoul_totem_aura : public SpellScriptLoader
...
+
+            _events.Update(diff);
+
+            if (uint32 eventId = _events.ExecuteEvent())
+            {
+                switch (eventId)
+                {
+                    case EVENT_TALK:
+                        if (Player* player = ObjectAccessor::GetPlayer(*me, _playerGUID))
+                            Talk(0, player);
```
Keader: magic number

Wyreth: Done. Not sure what name it should have - copied that from other scripts.

src/server/scripts/Northrend/zone_borean_tundra.cpp
```diff
@@ -2387,6 +2387,8 @@ class spell_windsoul_totem_aura : public SpellScriptLoader
 enum BloodsporeRuination
 {
     NPC_BLOODMAGE_LAURITH = 25381,
+    
+    TALK_0 = 0,
```
Keader: uneeded space '-'

sql/updates/world/3.3.5/2017_09_28_99_world_bloodspore_ruination.sql
```diff
@@ -0,0 +1,8 @@
+--
+DELETE FROM `spell_script_names` WHERE `spell_id`=45997;
```
Keader: use delete by ScriptName instead spell_id

Keader:
after author do this changes, can be merged.

src/server/scripts/Northrend/zone_borean_tundra.cpp
Wyreth: But but but space is good ',--,' :D

src/server/scripts/Northrend/zone_borean_tundra.cpp
```diff
@@ -2384,6 +2384,108 @@ class spell_windsoul_totem_aura : public SpellScriptLoader
     }
 };
 
+enum BloodsporeRuination
+{
+    NPC_BLOODMAGE_LAURITH = 25381,
+    TALK_0 = 0,
```
tkrokli: Because the word `Talk` is used already in the script as the action, use something beginning with `SAY_` to signify that the referenced text is said by the NPC. Use something descriptive like `SAY_BLOODMAGE_LAURITH` or use something directly related to the event or naming the event (`SAY_BLOODSPORE_RUINATION_TEST`), especially if the NPC in question has got many Say lines for different occasions (if the text said by the NPC is aggro, player kill or own death, it is usually named `SAY_AGGRO` `SAY_SLAY` or `SAY_DEATH` which is common in boss scripts).

Cosmetic, but worth noting: the old standard for creature_text comments when the text was used in a C++ script was to include the enum name as either part of the `creature_text`.`comment` field in DB or using the enum name as the only comment content.

PR #20465, opened 2017-09-29
Core/Quest: don't send "Received item" chat message for quest rewards

Wyreth:
**Changes proposed:**

As can be seen [here](https://www.youtube.com/watch?v=rb2LhTzv-eY&feature=youtu.be&t=412) (and any other video showing a quest reward), quest rewards shouldn't trigger the green-colored "You receive item x" in chat.

The quest packets already take care of that.

**Target branch(es):**

- [x] 3.3.5

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -14000,7 +14000,7 @@ void Player::SendItemDurations()
     }
 }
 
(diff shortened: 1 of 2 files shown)


tkrokli:
👍 Thank you for suggesting this fix. I have noticed that double "You receive item x" chat message in-game and been mildly annoyed about it, but never taken the time to find out why it happens.

PR #20494, opened 2017-09-30
Core/DataStores: implement operator [ ] on DB2Storage

Sovak:
**Changes proposed:**

-  It just looks prettier imo, faster to use & shorter

- [x] 3.3.5
- [x] master

- Builds & works


Ascathor:
That's actually pretty sweet. Considering lookupentry already checks for null, this seems like a safe method. I'd say pull it.

jackpoz:
looks good to me, we'll just have to remember it's not an array when using [ ]. let's keep all existing LookupEntry() calls the way they are tho

Shauren (approved):
Got "ok" from another dev, merging

PR #20508, opened 2017-10-02
[3.3.5] Scripts/ScarletMonastery: Mograine and Whitemane boss fight improvements

### LINKED ISSUE #15793: Scarlet Monastery/Cathedral: High Inquisitor Whitemane's door is open
TrinityCore rev. 60da7c655273 2015-10-31 16:04:14 +0100 (HEAD branch)

The door must be closed

kloster

Wyreth:
**Changes proposed:**

- Fix call for help when boss is aggroed (this fails when called in EnterCombat, so moved it to UpdateAI).
- Correctly handle the door to High Inquisitor Whitemane so that it opens only when supposed to.
- Fix wrong creature text for High Inquisitor Whitemane (was rehauled after WotLK, so the broadcast_text entry is not valid for WotLK).
- Whitemane should attack the player that killed Mograine first.

Closes #15793.

**Target branch(es):**

- [x] 3.3.5

### DIFF
--- a/src/server/scripts/EasternKingdoms/ScarletMonastery/boss_mograine_and_whitemane.cpp
+++ b/src/server/scripts/EasternKingdoms/ScarletMonastery/boss_mograine_and_whitemane.cpp
@@ -96,6 +96,7 @@ public:
         bool _bHasDied;
         bool _bHeal;
         bool _bFakeDeath;
+        bool _canCallForHelp;
 
         void Reset() override
         {
@@ -120,8 +121,7 @@ public:
         {
             Talk(SAY_MO_AGGRO);
             DoCast(me, SPELL_RETRIBUTIONAURA);
-
-            me->CallForHelp(VISIBLE_RANGE);
+            _canCallForHelp = true;
         }
 
         void KilledUnit(Unit* /*victim*/) override
@@ -129,7 +129,7 @@ public:
             Talk(SAY_MO_KILL);
         }
 
-        void DamageTaken(Unit* /*doneBy*/, uint32 &damage) override
+        void DamageTaken(Unit* attacker, uint32 &damage) override
         {
             if (damage < me->GetHealth() || _bHasDied || _bFakeDeath)
                 return;
@@ -140,6 +140,7 @@ public:
                 instance->SetBossState(DATA_MOGRAINE_AND_WHITE_EVENT, IN_PROGRESS);
 
                 Whitemane->GetMotionMaster()->MovePoint(1, 1163.113370f, 1398.856812f, 32.527786f);
+                Whitemane->GetAI()->AttackStart(attacker);
 
                 me->GetMotionMaster()->MovementExpired();
                 me->GetMotionMaster()->MoveIdle();
@@ -197,6 +198,12 @@ public:
 
                     _bHeal = true;
                 }
+            }
+
+            if (_canCallForHelp)
+            {
+                me->CallForHelp(VISIBLE_RANGE);
+                _canCallForHelp = false;
             }
 
             //This if-check to make sure mograine does not attack while fake death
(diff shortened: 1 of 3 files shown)


Ovahlord:
Why did you set the broadcast text id = 0?

Wyreth:
The broadcast text was changed when the instance was rehauled after WotLK. Instead of making another broadcast text, Blizzard reused that one. So it's not valid for WotLK.

offl:
Closes #15793?

Wyreth:
Right, thanks!

src/server/scripts/EasternKingdoms/ScarletMonastery/instance_scarlet_monastery.cpp
```diff
@@ -117,6 +113,10 @@ class instance_scarlet_monastery : public InstanceMapScript
                             HandleGameObject(PumpkinShrineGUID, false);
                         }
                         break;
+                    case DATA_MOGRAINE_AND_WHITE_EVENT:
+                        if (state == DONE || state == IN_PROGRESS)
+                            HandleGameObject(HighInquisitorsDoorGUID, true);
```
joschiwald: why you are not using generic doors handling?

Wyreth: Generic door handling is what made it stay open all the time. I'm not sure how it should be set up for a door that should only open mid-fight.

tkrokli:
Preferably preface SQL files with a comment line ( `--  ` ) , empty or with text.

src/server/scripts/EasternKingdoms/ScarletMonastery/boss_mograine_and_whitemane.cpp
```diff
@@ -295,6 +303,26 @@ class boss_high_inquisitor_whitemane : public CreatureScript
                 damage = me->GetHealth() - 1;
         }
 
+        void SetGUID(ObjectGuid attacker, int32 action) override
```
tkrokli:
```
src/server/scripts/EasternKingdoms/ScarletMonastery/boss_mograine_and_whitemane.cpp:306:49:
 fatal error: unused parameter 'action' [-Wunused-parameter]
        void SetGUID(ObjectGuid attacker, int32 action) override
                                                ^
```
Comment out the unused parameter ( `int32 /*action*/` ) -- unless you plan to modify the code to use it.

Wyreth: Didn't actually mean to push this, it got pushed automatically when I pushed another branch :D gonna take this out.

src/server/scripts/EasternKingdoms/ScarletMonastery/instance_scarlet_monastery.cpp
Shauren: By using DOOR_TYPE_SPAWN_HOLE instead of DOOR_TYPE_ROOM

Shauren: If it should open mid-fight and stay open after encounter then it needs manual handling

Wyreth:
Going to scrap this for now - planning a rewrite for the whole place since the Corrupted Ashbringer event is also missing. Eventually™

src/server/scripts/EasternKingdoms/ScarletMonastery/instance_scarlet_monastery.cpp
Tauriella: @Wyreth 
For the Door handling eventual take a look in the cMangos code.
In Cmangos, the doors work correctly.

illfated:
Hoping to hear back from you again,, @Wyreth (especially regarding any plans to work on this). 😃

PR #20588, opened 2017-10-10
Core/DBLayer: db check client major and minor version

freekode:
**Changes proposed:**

-  checks only major and minor version (5.1) , avoid subversion (5.1.27)

- [x] 3.3.5
- [x] master

tested on 3.3.5 server

### DIFF
--- a/src/server/database/Database/DatabaseWorkerPool.cpp
+++ b/src/server/database/Database/DatabaseWorkerPool.cpp
@@ -55,10 +55,14 @@ DatabaseWorkerPool<T>::DatabaseWorkerPool()
     : _queue(new ProducerConsumerQueue<SQLOperation*>()),
       _async_threads(0), _synch_threads(0)
 {
+    unsigned long usedMysqlMajorMinorVersion = mysql_get_client_version() / 100;
+    unsigned long compileMysqlMajorMinorVersion = MYSQL_VERSION_ID / 100;
+
     WPFatal(mysql_thread_safe(), "Used MySQL library isn't thread-safe.");
     WPFatal(mysql_get_client_version() >= MIN_MYSQL_CLIENT_VERSION, "TrinityCore does not support MySQL versions below 5.1");
-    WPFatal(mysql_get_client_version() == MYSQL_VERSION_ID, "Used MySQL library version (%s) does not match the version used to compile TrinityCore (%s). Search on forum for TCE00011.",
-        mysql_get_client_info(), MYSQL_SERVER_VERSION);
+    WPFatal(usedMysqlMajorMinorVersion == compileMysqlMajorMinorVersion,
+        "Used MySQL library version (%s) does not match the version used to compile TrinityCore (%s). Your version should be >= %s. Search on forum for TCE00011.",
+        mysql_get_client_info(), MYSQL_SERVER_VERSION, compileMysqlMajorMinorVersion * 100);
 }
 
 template <class T>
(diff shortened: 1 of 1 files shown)


src/server/database/Database/DatabaseWorkerPool.cpp
```diff
@@ -55,10 +55,14 @@ DatabaseWorkerPool<T>::DatabaseWorkerPool()
...
+    unsigned long usedMysqlMajorMinorVersion = mysql_get_client_version() / 100;
+    unsigned long compileMysqlMajorMinorVersion = MYSQL_VERSION_ID / 100;
+
     WPFatal(mysql_thread_safe(), "Used MySQL library isn't thread-safe.");
     WPFatal(mysql_get_client_version() >= MIN_MYSQL_CLIENT_VERSION, "TrinityCore does not support MySQL versions below 5.1");
-    WPFatal(mysql_get_client_version() == MYSQL_VERSION_ID, "Used MySQL library version (%s) does not match the version used to compile TrinityCore (%s). Search on forum for TCE00011.",
-        mysql_get_client_info(), MYSQL_SERVER_VERSION);
+    WPFatal(usedMysqlMajorMinorVersion == compileMysqlMajorMinorVersion,
+        "Used MySQL library version (%s) does not match the version used to compile TrinityCore (%s). Your version should be >= %s. Search on forum for TCE00011.",
+        mysql_get_client_info(), MYSQL_SERVER_VERSION, compileMysqlMajorMinorVersion * 100);
```
tkrokli:
```
src/server/database/Database/DatabaseWorkerPool.cpp:65:56:
 fatal error: format specifies type 'char *' but the argument has type 'unsigned long' [-Wformat]
        mysql_get_client_info(), MYSQL_SERVER_VERSION, compileMysqlMajorMinorVersion * 100);
                                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/common/Debugging/Errors.h:50:110: note: expanded from macro 'WPFatal'
#define WPFatal(cond, ...) ASSERT_BEGIN do { if (!(cond)) Trinity::Fatal(__FILE__, __LINE__, __FUNCTION__, ##__VA_ARGS__); } while(0) ASSERT_END
                                                                                                             ^~~~~~~~~~~
src/server/database/Database/DatabaseWorkerPool.cpp:450:32: note: in instantiation of member function 'DatabaseWorkerPool<LoginDatabaseConnection>::DatabaseWorkerPool' requested here
template class TC_DATABASE_API DatabaseWorkerPool<LoginDatabaseConnection>;
                               ^
1 error generated.
```

Aokromes:
Like i told you on irc the check is intended.
https://github.com/TrinityCore/TrinityCore/commit/0a27f8bce264b8f6d8c55efa8ebc97f6e9cb7d11

jackpoz:
what's the problem about checking subversion too ?

tkrokli:
I presume some may see it as being redundant / not needed. Then again, are there any cases where a MySQL sub version diff could break something or cause data corruption if the compile is done with a sub version different from the dependency file in the core folder?

Aokromes:
I see it as "fuck repacks users".

freekode:
I want to use trinitycore deb packages. Then to start worldserver I don't need exact version of mysql with which worldserver was compiled.

jackpoz:
I would prefer to have the exact same version to avoid any possible (even if remote) issue. You can install locally the mysqlclient version of that deb package

PR #20610, opened 2017-10-13
Core/AI: implemented AI hook for interrupting spells

Ovahlord:
**Changes proposed:**

-  implement an AI hook that gets called when a creature's spell cast gets interrupted. Needed for Cataclysm+ boss implementations. Maybe 3.3.5 as well.

- [x] 3.3.5
- [x] master

- Tested ingame with some scripts, works as intended.

### DIFF
--- a/src/server/game/AI/CreatureAI.h
+++ b/src/server/game/AI/CreatureAI.h
@@ -157,6 +157,9 @@ class TC_GAME_API CreatureAI : public UnitAI
 
         void OnCharmed(bool apply) override;
 
+        // Called when a spell cast gets interrupted
+        virtual void OnSpellCastInterrupt(SpellInfo const* /*spell*/) { }
+
         // Called at reaching home after evade
         virtual void JustReachedHome() { }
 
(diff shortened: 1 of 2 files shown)


r00ty-tc:
It's a good one for sure. Is there consensus that an SAI event for this would also be useful?

Keader:
Looks good to me.
Idk if is usefull to 3.3.5 :)

Ovahlord:
I don't know any regular npc that uses this mechanic. Only bosses so far, so from my side, I dont see a reason (yet) to implement such an SAI function. Feel free to add it afterwards

Shauren:
Looks good but I have a question, are you aware of any boss mechanic that would require more info about spell being interrupted (such as remaining channel time or something like that)

Ovahlord:
Hmm, good question, but as far as I know I don't know any boss that uses such a mechanic. But yes, this could be expanded if needed. Do you know a boss, that uses such a information?

Shauren:
Can't remember, thats why I asked

Ovahlord:
Me neither. So far I've only seen spells, that needs that for interrupt events.

Shauren (approved):

PR #20611, opened 2017-10-13
Scripts/Commands: Add failure state for quest remove

mysterious-static:
**Changes proposed:**

-  Add more verbose messaging to the `.quest remove` command. This could be useful for debugging; there's another issue that I _think_ this command might have but I can't tell without better output.

**Target branch(es):**
3.3.5. It's probably relevant for master too, but I branched off of 3.3.5 so I see no problem starting here. :)

**Tests performed:**
It builds, which is a good start since I don't really know much about C++. The logic checks out to me, but I haven't been able to test it yet due to cost constraints on my end.

**Questions:**

- Should there be a separate, unique message for this failure state rather than reusing `LANG_COMMAND_QUEST_NOTFOUND`?
- If so, what should it say?
- If not, should the conditions be moved up to line 124, which already outputs `LANG_COMMAND_QUEST_NOTFOUND`, instead of here?

### DIFF
--- a/src/server/scripts/Commands/cs_quest.cpp
+++ b/src/server/scripts/Commands/cs_quest.cpp
@@ -128,27 +128,26 @@ public:
             return false;
         }
 
-        // remove all quest entries for 'entry' from quest log
-        for (uint8 slot = 0; slot < MAX_QUEST_LOG_SIZE; ++slot)
-        {
-            uint32 logQuest = player->GetQuestSlotQuestId(slot);
-            if (logQuest == entry)
-            {
-                player->SetQuestSlot(slot, 0);
-
-                // we ignore unequippable quest items in this case, its' still be equipped
-                player->TakeQuestSourceItem(logQuest, false);
-
-                if (quest->HasFlag(QUEST_FLAGS_FLAGS_PVP))
-                {
-                    player->pvpInfo.IsHostile = player->pvpInfo.IsInHostileArea || player->HasPvPForcingQuest();
-                    player->UpdatePvPState();
-                }
-            }
-        }
-        
         if (player->GetQuestStatus(entry) != QUEST_STATUS_NONE)
         {
+            // remove all quest entries for 'entry' from quest log
+            for (uint8 slot = 0; slot < MAX_QUEST_LOG_SIZE; ++slot)
+            {
+                uint32 logQuest = player->GetQuestSlotQuestId(slot);
+                if (logQuest == entry)
+                {
+                    player->SetQuestSlot(slot, 0);
+
+                    // we ignore unequippable quest items in this case, its' still be equipped
+                    player->TakeQuestSourceItem(logQuest, false);
+
+                    if (quest->HasFlag(QUEST_FLAGS_FLAGS_PVP))
+                    {
+                        player->pvpInfo.IsHostile = player->pvpInfo.IsInHostileArea || player->HasPvPForcingQuest();
+                        player->UpdatePvPState();
+                    }
+                }
+            }
             player->RemoveActiveQuest(entry, false);
             player->RemoveRewardedQuest(entry);
 
--- a/src/server/scripts/Commands/cs_quest.cpp
+++ b/src/server/scripts/Commands/cs_quest.cpp
@@ -147,7 +147,7 @@ public:
             }
         }
         
-        if(player->GetQuestStatus(entry) != QUEST_STATUS_NONE /* || player->IsActiveQuest(entry) ? */ )
+        if (player->GetQuestStatus(entry) != QUEST_STATUS_NONE /* || player->IsActiveQuest(entry) ? */ )
         {
             player->RemoveActiveQuest(entry, false);
             player->RemoveRewardedQuest(entry);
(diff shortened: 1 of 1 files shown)


src/server/scripts/Commands/cs_quest.cpp
```diff
@@ -146,14 +146,21 @@ class quest_commandscript : public CommandScript
...
-        player->RemoveRewardedQuest(entry);
+            sScriptMgr->OnQuestStatusChange(player, entry);
 
-        sScriptMgr->OnQuestStatusChange(player, entry);
-
-        handler->SendSysMessage(LANG_COMMAND_QUEST_REMOVED);
-        return true;
+            handler->SendSysMessage(LANG_COMMAND_QUEST_REMOVED);
+            return true;
+        } else {
```
vanasis: Move `else {` to new line

src/server/scripts/Commands/cs_quest.cpp
```diff
@@ -146,14 +146,21 @@ class quest_commandscript : public CommandScript
                 }
             }
         }
+        
+        if(player->GetQuestStatus(entry) != QUEST_STATUS_NONE /* || player->IsActiveQuest(entry) ? */ ) {
```
vanasis: change it to 
```
if (player->GetQuestStatus(entry) != QUEST_STATUS_NONE /* || player->IsActiveQuest(entry) ? */ )
{
    ....
```

vanasis (commented):
CodeStyle

src/server/scripts/Commands/cs_quest.cpp
```diff
@@ -147,15 +147,17 @@ class quest_commandscript : public CommandScript
             }
         }
         
-        if(player->GetQuestStatus(entry) != QUEST_STATUS_NONE /* || player->IsActiveQuest(entry) ? */ ) {
+        if(player->GetQuestStatus(entry) != QUEST_STATUS_NONE /* || player->IsActiveQuest(entry) ? */ )
```
vanasis: some small things, use  `if (...)` instead of `if(...)` ( one space )

src/server/scripts/Commands/cs_quest.cpp
```diff
@@ -147,15 +147,17 @@ class quest_commandscript : public CommandScript
...
             player->RemoveActiveQuest(entry, false);
             player->RemoveRewardedQuest(entry);
 
             sScriptMgr->OnQuestStatusChange(player, entry);
 
             handler->SendSysMessage(LANG_COMMAND_QUEST_REMOVED);
             return true;
-        } else {
+        }
+        else {
```
vanasis: also change
`else  { `
to
```
else
{
```

vanasis (commented):
Small changes

velinath:
I think at this point most of what I'm waiting on is dev feedback for this PR per the questions in the original PR text. As far as I could tell `player->IsActiveQuest(entry)` was unnecessary so I'll be removing that from the PR

src/server/scripts/Commands/cs_quest.cpp
```diff
@@ -146,14 +146,24 @@ class quest_commandscript : public CommandScript
...
+            handler->SendSysMessage(LANG_COMMAND_QUEST_REMOVED);
+            return true;
+        }
+        else
+        {
+            handler->SendSysMessage(LANG_COMMAND_QUEST_NOTFOUND);
+            handler->SetSentErrorMessage(true);
+            return false;
+        }
+
```
tkrokli: cosmetic/codestyle: this blank line can be removed.

velinath:
Just wanted to check in on this as I can't finish without some feedback from project maintainers.

Outstanding questions before this is ready to go:

- Should there be a separate, unique message for this failure state rather than reusing LANG_COMMAND_QUEST_NOTFOUND?
- If so, what should it say?
- If not, should the conditions be moved up to line 124, which already outputs LANG_COMMAND_QUEST_NOTFOUND, instead of here?

jackpoz:
By looking at https://github.com/velinath/TrinityCore/blob/05c763844f32f1093f3bfa883c2121923d23fc71/src/server/scripts/Commands/cs_quest.cpp#L191 it seems that ".quest complete #entry" command already sends the same LANG_COMMAND_QUEST_NOTFOUND error if the player doesn't have the quest.

For this PR it's ok to keep the same behavior, if you want you can add a new message later on in another PR.

I wonder tho if the for loop at https://github.com/velinath/TrinityCore/blob/05c763844f32f1093f3bfa883c2121923d23fc71/src/server/scripts/Commands/cs_quest.cpp#L132 could be moved inside the "if" check in the case the player has the quest.

velinath:
the if stmt starting at L150? I don't really see a reason why it couldn't since it wouldn't need to do anything if the quest didn't exist, you're right

src/server/scripts/Commands/cs_quest.cpp
```diff
@@ -128,32 +128,40 @@ class quest_commandscript : public CommandScript
...
             {
-                player->SetQuestSlot(slot, 0);
+                uint32 logQuest = player->GetQuestSlotQuestId(slot);
+                if (logQuest == entry)
+                {
+                    player->SetQuestSlot(slot, 0);
 
-                // we ignore unequippable quest items in this case, its' still be equipped
-                player->TakeQuestSourceItem(logQuest, false);
+                    // we ignore unequippable quest items in this case, its' still be equipped
```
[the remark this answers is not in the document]
velinath: I'm not entirely sure what it does or whether it's worth keeping the comment at all, would prefer someone else speak to what to do with this one.

PR #20615, opened 2017-10-15
Core/Creature: aggro radius calculation reworked

Ovahlord:
**Changes proposed:**

-  reworked aggro range calculation to get rid of incorrect aggro ranges

- [x] 3.3.5
- [x] master

- tested ingame, works so far as I can compare it to retail


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1646,40 +1646,44 @@ bool Creature::CheckNoGrayAggroConfig(uint32 playerLevel, uint32 creatureLevel)
 
 float Creature::GetAttackDistance(Unit const* player) const
 {
+    // WoW Wiki: the minimum radius seems to be 5 yards, while the maximum range is 45 yards
+    float maxRadius = (45.0f * sWorld->getRate(RATE_CREATURE_AGGRO));
+    float minRadius = (5.0f * sWorld->getRate(RATE_CREATURE_AGGRO));
     float aggroRate = sWorld->getRate(RATE_CREATURE_AGGRO);
-    if (aggroRate == 0)
-        return 0.0f;
+    uint8 expansionMaxLevel = GetMaxLevelForExpansion(GetCreatureTemplate()->expansion);
```
tkrokli:
```
src/server/game/Entities/Creature/Creature.cpp:1653:78:
 fatal error: no member named 'expansion' in 'CreatureTemplate'
    uint8 expansionMaxLevel = GetMaxLevelForExpansion(GetCreatureTemplate()->expansion);
                                                      ~~~~~~~~~~~~~~~~~~~~~  ^
```

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1646,40 +1646,44 @@ bool Creature::CheckNoGrayAggroConfig(uint32 playerLevel, uint32 creatureLevel)
 
 float Creature::GetAttackDistance(Unit const* player) const
 {
+    // WoW Wiki: the minimum radius seems to be 5 yards, while the maximum range is 45 yards
+    float maxRadius = (45.0f * sWorld->getRate(RATE_CREATURE_AGGRO));
+    float minRadius = (5.0f * sWorld->getRate(RATE_CREATURE_AGGRO));
     float aggroRate = sWorld->getRate(RATE_CREATURE_AGGRO);
-    if (aggroRate == 0)
-        return 0.0f;
+    uint8 expansionMaxLevel = GetMaxLevelForExpansion(GetCreatureTemplate()->Expansion);
```
tkrokli:
```
src/server/game/Entities/Creature/Creature.cpp:1653:78:
 fatal error: no member named 'Expansion' in 'CreatureTemplate'
    uint8 expansionMaxLevel = GetMaxLevelForExpansion(GetCreatureTemplate()->Expansion);
                                                      ~~~~~~~~~~~~~~~~~~~~~  ^
1 error generated.
```

Ovahlord: wtf...

tkrokli: Yes, I know. Sometimes the cause of an error is not as obvious as it would seem.

Ovahlord: Ye, I mean what happened to that field? where has it gone or how to I access it. Cant open VS right now to look for it

Shauren: There are 2 expansion fields `HealthScalingExpansion` (determines max health, our) and `RequiredExpansion` (client restricts interaction based on this, WDB field)

Ovahlord: which one would you recommend to use to identify the creature's expansion?

tkrokli: `RequiredExpansion` should be the correct one in relation to when a creature was added to the game, but I could be wrong if this is about health vs level.

Keader:
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Entities/Creature/Creature.cpp:1653:31: fatal error: use of undeclared identifier 'GetMaxLevelForExpansion'
    uint8 expansionMaxLevel = GetMaxLevelForExpansion(GetCreatureTemplate()->RequiredExpansion);
                              ^
1 error generated.

tkrokli:
OK, so the difference now is that you need to declare `GetMaxLevelForExpansion` .

Ovahlord:
is some1 arround here to look for the proper check to get the expansion for the creature? I currently have no open master branch arround here so I might need some1 for that

tkrokli:
You mean something like this? (I hope this can be of some help to you)
https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Handlers/ReferAFriendHandler.cpp#L70
```cpp
    else if (target->getLevel() >= GetMaxLevelForExpansion(target->GetSession()->GetExpansion()))
```
I think you also need to declare it first, like in this line
https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Handlers/ReferAFriendHandler.cpp#L25
```cpp
inline uint32 GetMaxLevelForExpansion(uint32 expansion)
```
This is only based on what I found in a simple TC site search:
https://github.com/TrinityCore/TrinityCore/search?q=GetMaxLevelForExpansion&type=Code&utf8=✓

Teppic1:
I believe that the aggro distance is based on the colour of the mob (orange, red, etc) rather than a simple 1 level per yard calculation. The formula is already implemented for determining the colours.

Faq:
So only thing this PR missing to merge is compile error?

tkrokli:
Not sure, but the error is important enough. Do you have a suggestion for solving the error? :)

Faq:
For example this compiles
```
    uint8 GetMaxLevelForExpansion(uint8 expansionMaxLevel);
    uint8 expansionMaxLevel = (GetCreatureTemplate()->RequiredExpansion);
```

Aokromes:
It looks it needs some CP from 3.3.5a branch.

tkrokli:
OK, but first let us get rid of the declaration error so we can discover if there are any other compile errors in the changes.

Shauren:
@Faq but it obviously does not work since you are not calling the function!
@tkrokli please dont integrate changes just anyone posts in commends

tkrokli:
Very well. Should I revert the offending commit?

Shauren:
at this point no

funjoker:
Scaling Creatures are ignored.. Needs fix.

PR #20626, opened 2017-10-16
[3.3.5] DB/Scripts: Rivern Frostwind

mysterious-static:
**Changes proposed:**

-  Migrates another vendor.

**Target branch(es):**
- [X] 3.3.5

**Issues addressed:**
Migrates another NPC out of C++ scripts

Builds. Works ingame.

- [ ] Need verify gossip_menu_option again, I searched broadcast_text for terms that might work and couldn't find anything, and I don't know anyone exalted with this faction to test on live.


src/server/scripts/Kalimdor/zone_winterspring.cpp
```diff
@@ -40,48 +39,6 @@ EndContentData */
 #include "TemporarySummon.h"
```
Aokromes: check if gossip includes is unneed after this changes :)

velinath: I thought it was because go_elune_fire uses GossipHello but I now see that that's not in ScriptedGossip.h, you're right. Thanks for the heads up.

PR #19409, opened 2017-04-05
Core/Entities: Warlock pet base damage table fix

PolarCookie:
remake of https://github.com/TrinityCore/TrinityCore/pull/19360 on a new branch
tldr: suggested fix to the warlock pet base damage, placing base damage into the pet_levelstats table.
Fills world.pet_levelstats with base damage values and stats for the 5 common warlock pets for all levels.

references:
[1] simcraft
https://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/simulationcraft/simc-335-1-win32.zip
[2] old wowwiki
http://wowwiki.wikia.com/wiki/Felguard_(warlock_minion)?oldid=2409171
http://wowwiki.wikia.com/wiki/Felhunter?oldid=2714647
http://wowwiki.wikia.com/wiki/Succubus_(warlock_minion)?oldid=2408937
http://wowwiki.wikia.com/wiki/Voidwalker_(warlock_minion)?oldid=2353207
http://wowwiki.wikia.com/wiki/Imp_(warlock_minion)?oldid=2410392
[3] voidwalker tanking
https://www.youtube.com/watch?v=SWqecvALz6Q
https://www.youtube.com/watch?v=WO-E2craHms
[4] my work files
https://www.mediafire.com/folder/rbt5u2yu1w232a9,7mt04maidc7u0uu,gv97znicm4y5sb5,rv7jsd0d32gc7sc,990y8fgqhk5vgcc,je7y4dfqg574cug,quft16relsflk6f,ur88x7lih6d56x1/shared

base_dmg_nonmono
This is a plot of all the wowwiki datapoints, with stock trinity and sunwells felguard for comparison. There are many datapoints that are missing, if anyone have any sources to fill in let me know asap, but there's reason to suspect Blizzard reduced the damage at lvl 60 and 70 for balancing reason back in the day. Other than that it seems to be more or less smooth curves.

Dalaran-wow and ToD (rip) solved this issue by linear scaling; dalaranwow/dalaran-wow#3237.
This works for lvl 80, but all pets become massively overpowered at lower levels.

base_dmg
This solution will scale the pets as illustrated on this image.
I interpolated the tables from wowwiki and kept it strictly monotonic, ignoring damage values lower than previous levels. This can be discussed, but the alternative would be the zigzagged progression as shown on the graph above. I suspect this monotonic will be closer to the actual blizzlike values as the wowwiki table is missing far to many data points in the 60-80 range.
[...]

PR #20632, opened 2017-10-17
[WIP] [3.3.5] Warlock Infernal, Research help needed

PolarCookie:
Disclaimer: this is a continuation on https://github.com/TrinityCore/TrinityCore/pull/19409 which, as of 17.10.2017, is not merged, but I wanted to make this a separate PR to keep the research conversation about Infernal stats focused.

SQL FILE WILL BE UPDATED IF ANY NEW DATA APPEARS

**Changes proposed:**
The Warlock Infernal is missing it's Immolation passive and does next to no damage.
I found this recently and started looking at the numbers.
https://youtu.be/aqvgRj9qJvM?t=3m36s

*Damage*
The average melee hit on the dummy is 1799.
The armor on an expert dummy is 3791 (speculative, from the TDB, need confirmation)
Adjusting for armor the average melee damage is 2247
If [Immolation](https://wow.gamepedia.com/index.php?title=Immolation&oldid=2060744) is hitting for 543, the warlock would have 3353sp.
Infernal AP from warlock would be 1911 and if we assume the same primary stats as the other demons (speculative, need confirmation), the avg base damage is now 1887

*Health*
I put together some ICC gear with proper enchants/gems to get close to 3353sp, stamina was 1343 giving 1007 stam to the Infernal. To reach 43686hp, assuming infernal gains 11hp/stam (spec. need conf.), base health was set to 32609hp.

*Armor*
[This](https://wow.gamepedia.com/index.php?title=Warlock_minion&oldid=2478495) mentions Infernal having more armor than Voidwalker. As I don't have any hard stats on this I just gave Infernal Voidwalkers armor numbers.

*50-79*
Stats for level 50-79 were scaled to follow the same curve as Voidwalker.
Lvl 50 Infernal starts out at 12873 base hp and 621 avg. base damage, which might be a bit OP...

*workfiles*
https://www.mediafire.com/folder/rbt5u2yu1w232a9,990y8fgqhk5vgcc,je7y4dfqg574cug,rv7jsd0d32gc7sc,ur88x7lih6d56x1,gv97znicm4y5sb5,quft16relsflk6f,jnv9bhejn5rs8q8/shared

**Target branch(es):**
- [x] 3.3.5

**Tests performed:**
ICC gear close to 3353SP, for demo with demonic knowledge and proper gems/enchants, to get the stamina right.
wowscrnshot_101717_152646

Same gear, but with a new trinket and just runed cardinal rubies to hit 3353 to test damage.
wowscrnshot_101717_152932

- [ ] More data points for any relevant stat.


sql/updates/world/3.3.5/9999_99_99_99_world_2_335.sql
```diff
@@ -0,0 +1,94 @@
+-- Infernal SAI
+SET @ENTRY := 89;
+UPDATE `creature_template` SET `AIName`="SmartAI" WHERE `entry`=@ENTRY;
+DELETE FROM `smart_scripts` WHERE `entryorguid`=@ENTRY AND `source_type`=0;
+INSERT INTO `smart_scripts` (`entryorguid`,`source_type`,`id`,`link`,`event_type`,`event_phase_mask`,`event_chance`,`event_flags`,`event_param1`,`event_param2`,`event_param3`,`event_param4`,`action_type`,`action_param1`,`action_param2`,`action_param3`,`action_param4`,`action_param5`,`action_param6`,`target_type`,`target_param1`,`target_param2`,`target_param3`,`target_x`,`target_y`,`target_z`,`target_o`,`comment`) VALUES
+(@ENTRY,0,0,0,54,0,100,0,0,0,0,0,11,19483,0,0,0,0,0,1,0,0,0,0,0,0,0,"Infernal - On Just Summon - Cast 'Immolation'"),
+(@ENTRY,0,1,0,54,0,100,0,0,0,0,0,11,32233,0,0,0,0,0,1,0,0,0,0,0,0,0,"Infernal - On Just Summon - Cast 'Avoidance'");
+
```
Killyana: Move all this SAI to creature_template_addon

Keader:
Demonic Knowledge is handled in Pet::CastPetAuras(true)

PolarCookie:
I'd love to have a video of someone testing out their new Infernal at lvl 50. Logs too would be nice.

tkrokli:
Looks like the SQL file import fails because the SQL content is in 2 separate files and the SQL import does not seem to import the files in the order you were hoping for.

Consider merging the 2 SQL files into 1 so you can be absolutely sure that the column `min_dmg` gets created before any `INSERT INTO` or `UPDATE` queries are executed.

PolarCookie:
@tkrokli, should be fine now.

tkrokli:
Good job. Glad to see that it worked.
Keep up the good work. 👍

src/server/game/Globals/ObjectMgr.h
```diff
@@ -657,7 +657,7 @@ typedef std::pair<ExclusiveQuestGroups::const_iterator, ExclusiveQuestGroups::co
 
 struct PetLevelInfo
 {
-    PetLevelInfo() : health(0), mana(0), armor(0)
+    PetLevelInfo() : health(0), mana(0), armor(0), min_dmg(0), max_dmg(0)
```
ccrs: minDamage or minimumDamage

PolarCookie: updated it in the parent branch #19409

tkrokli: Most likely `minDamage` because 'minimumDamage' did not yield any results in a search through my local 3.3.5 files (https://github.com/TrinityCore/TrinityCore/commit/b507acf1c04ff6e9d0a075f9663e9549d5b238b4) or in the master branch (https://github.com/TrinityCore/TrinityCore/search?q=minimumDamage&utf8=✓) while `minDamage` gives search results both locally and in master (https://github.com/TrinityCore/TrinityCore/search?q=minDamage&utf8=✓).

PolarCookie: thanks anyway!

velinath:
I linked this already in #5136 but maybe you will also find it helpful for this? 

[SimulationCraft warlock script from 3.3.5a](https://gist.github.com/velinath/223531327f41e6d5088e969632d32b7b)

PolarCookie:
@velinath, sorry for the delayed respones.
I did take a look at simcraft before hand, but I get a bit uneasy by it, I highly believe simcraft is incorrect on this one.
1) https://gist.github.com/velinath/223531327f41e6d5088e969632d32b7b#file-sc_warlock-cpp-L3592
Infernal doesn't get Demonic Knowledge, but the video I used as a base clearly shows the aura being applied.
2) https://gist.github.com/velinath/223531327f41e6d5088e969632d32b7b#file-sc_warlock-cpp-L454
Infernal doesn't get AP from the warlock. This one I'd like to get a second source on before I believe it.
3) https://gist.github.com/velinath/223531327f41e6d5088e969632d32b7b#file-sc_warlock-cpp-L917
Infernal straight up gets 100% of warlock SP, and then
https://gist.github.com/velinath/223531327f41e6d5088e969632d32b7b#file-sc_warlock-cpp-L900
does 40+20% of that SP in Immolation damage. It should have been 40+100% of Infernal SP and infernal getting 15% of warlocks SP, as every other demon I would assume.
4) https://gist.github.com/velinath/223531327f41e6d5088e969632d32b7b#file-sc_warlock-cpp-L927
Infernal melee damage is (412.5+(331×2−20)×2÷14)×3.2 = 1613.4857 (base damage + ap from strength times the damage modifer). But just from the video I used this should be impossible because the applied average damage there was 1799 after armor reduction. 

Maybe I have been reading the simcraft file all wrong, but I'm cautious.

sphericalsilver:
Yeah indeed, Affliction is the only one that has any business using the Infernal or Doomguard. As for resummoning the pet thereafter, the way you did it on retail was you'd just summon the Infernal in the last minute of the fight. 

I'm really stuck on this issue of Infernal damage formulas myself, and it doesn't help that it's so difficult to find logs that feature Warlocks using Infernals. I figure if you had enough logs you can use Immolation Aura's dmg as an indicator of the Warlock's SP, and after that, you take that into consideration the buffs and debuffs present on the target to analyze the Infernal's melee dmg, and gradually basically reverse-engineer the raw damage formula. 

Simcraft is most surely wrong about the Infernal not inheriting the Warlock's SP as AP, if that's indeed what it says.

Another matter I'm curious about is the Crit chance. Specifically the Agi -> Crit conversion, and whether it benefits from Demonic Tactics or Improved Demonic Tactics.

Keader:
Still work on it @PolarCookie?

illfated:
Maybe add the label [help wanted] to make it show up in the TC repository subtitle
(currently " 2 issues need help " in the repo list  - https://github.com/TrinityCore )

PolarCookie:
@Keader I forgot this existed, I've been so busy playing on Felmyst, we got 3/6 SSC (Leo down) yesterday, but now...
I'll need some time.

PolarCookie:
Originally I was waiting for #19409 to be merged before continuing on this one, as it relies on it.
I'm was very happy to see the merge happen at the time.

PolarCookie:
Having a lot of trouble getting updated with the current TC correctly, starting over on a new PR soon.

PR #20652, opened 2017-10-18
[WIP] [3.3.5] Pet hit rating inheritance

PolarCookie:
**Changes proposed:**
Warlock (and hunter) pets do not inherit hit from their owners.
This used to be done through scaling auras, but these seem to have been deprecated in favour of the StatSystem (right?)
Only, StatSystem includes everything for pets but combat ratings, so here's a fix for hit.

This fix feels kinda "tacked on", any thoughts on this @ariel-? you wrote most of the StatSystem, right?

**Target branch(es):**
- [x] 3.3.5

**Issues addressed:** Closes probably many that I don't remember...

**Tests performed:**
Tested for Felhunter, was using Instructor Razuvious on 25 man as my dummy.

Here is the current situation, hit rating was 368 (14.03%)

This PR at 145 hit rating (5.53%), with 3/3 suppression.

This PR at 368 (14.03%), with 3/3 suppression.

### DIFF
--- a/src/server/game/Entities/Creature/TemporarySummon.h
+++ b/src/server/game/Entities/Creature/TemporarySummon.h
@@ -85,6 +85,11 @@ class TC_GAME_API Guardian : public Minion
         void UpdateAttackPowerAndDamage(bool ranged = false) override;
         void UpdateDamagePhysical(WeaponAttackType attType) override;
 
+        void UpdateMeleeHitChances();
+        void UpdateSpellHitChances();
+        void UpdateExpertise(WeaponAttackType attack);
+
+
         int32 GetBonusDamage() const { return m_bonusSpellDamage; }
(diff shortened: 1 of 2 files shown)


Shauren:
No, doing it with auras is the preferred way, it was supposed to be done with auras but never properly implemented

PolarCookie:
well, until they work, this will do for now.

robinsch:
You simply prevent mistakes by implementing the auras, for example Ghoul should not get owners hit scaling (they don't have an aura for it).

PolarCookie:
What is the problem with the scaling auras then? why do they not work/what needs to be fixed?

Keader:
They dont have a "base point", you need set amount by scripts :)
hard part is: know how much

robinsch:
@PolarCookie  Base points / talent scaling etc.

Issue #9469, opened 2013-03-20
Quest: Forced Hand / The Mist Isn't Listening


RedSonja:
Trinity Revision: 4e1a4b3ce4d5+ Database Version: TDB 335.51
http://www.wowhead.com/quest=12576

Forced Hand the required npcs are all unattackable (invalid target), they can however be killed by aoe or pet damage.
##

Rushor:
try this:

```c++
diff --git a/src/server/game/Reputation/ReputationMgr.cpp b/src/server/game/Reputation/ReputationMgr.cpp
index 69e677b..e373fb0 100644
--- a/src/server/game/Reputation/ReputationMgr.cpp
+++ b/src/server/game/Reputation/ReputationMgr.cpp
@@ -459,6 +459,10 @@ void ReputationMgr::SetAtWar(RepListID repListID, bool on)
 
 void ReputationMgr::SetAtWar(FactionState* faction, bool atWar) const
 {
+    if (atWar)
+        if (faction->ID == 1104 || faction->ID == 1105)
+            faction->Flags |= FACTION_FLAG_AT_WAR;
+
     // not allow declare war to own faction
     if (atWar && (faction->Flags & FACTION_FLAG_PEACE_FORCED))
         return;
```

Killyana:
@Rushor your fix works, if it's not considered as hack it can be pushed.

Rushor:
oh well, it creates a new bug - if you reached a friendly rep with the npcs, you can still attack them :(

tkrokli:
Interesting. You need to be at war with the NPCs if you have got either the initial quest or the daily, but not to the faction you take the quest from. Does this need to be defined in the quest or in the zone script?
My first guess would be the script zone_sholazar_basin.cpp, since the bug seems to manifest when you leave the zone and then enter it again.

dr-j:
could script only set at war if player has lower then friendly reputation with either of these factions/

Killyana:
Yes I was testing the conditions: it works
```
diff --git a/src/server/game/Reputation/ReputationMgr.cpp b/src/server/game/Reputation/ReputationMgr.cpp
index 69e677b..e373fb0 100644
--- a/src/server/game/Reputation/ReputationMgr.cpp
+++ b/src/server/game/Reputation/ReputationMgr.cpp
@@ -459,6 +459,10 @@ void ReputationMgr::SetAtWar(RepListID repListID, bool on)

[...]

PR #20661, opened 2017-10-20
Core/Reputation: Aldor & Scryer guards should attack and be attackable when hostile

toto6:
(These guards that cast a spell "exile" on you teleporting you out of Shattrath)

**Changes proposed:**
Allow to set at war for factions flagged as "rival" (only Aldor & Scryer in 335 I think)

- [x] 3.3.5

### DIFF
--- a/src/server/game/Reputation/ReputationMgr.cpp
+++ b/src/server/game/Reputation/ReputationMgr.cpp
@@ -463,8 +463,8 @@ void ReputationMgr::SetAtWar(RepListID repListID, bool on)
 
 void ReputationMgr::SetAtWar(FactionState* faction, bool atWar) const
 {
-    // not allow declare war to own faction
-    if (atWar && (faction->Flags & FACTION_FLAG_PEACE_FORCED))
+    // Do not allow to declare war to our own faction. But allow for rival factions (eg Aldor vs Scryer).
+    if (atWar && (faction->Flags & FACTION_FLAG_PEACE_FORCED) && !(faction->Flags & FACTION_FLAG_RIVAL))
         return;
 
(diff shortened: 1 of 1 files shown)


Aokromes:
Maybe this fixes also https://github.com/TrinityCore/TrinityCore/issues/9469 ?

aje:
I'm not sure; I work only on TBC so I'm not familiar with this

Aokromes:
This is still needed after https://github.com/TrinityCore/TrinityCore/commit/299323ce5d2e478f2eb7fcd6572c5fbf87b98c6e ?

aje:
yes, I just tested

aje:
oups I closed by error

tkrokli:
Looks like you deleted your PR branch. Unless you are able to undelete it, you need to open a new PR on a new branch.

aje:
cant you add this code :D its only 2 lines easy

tkrokli:
We need to get some feedback from one of the main TC developer members to validate that it can be added to the source. I do see that there is only 1 line of code + one comment changed, but we still need one of the major developers to validate the added change.

tkrokli:
It would also be useful to hear from other users if the code has been tested to solve all the issues related to the change.

tkrokli:
BTW, the TC member who pushes the commit will have to know that the change works as intended and to either have tested it well enough to be able to validate it, or rely on other testers to confirm that it works exactly as intended.

tkrokli:
@Killyana : Thank you for the quick follow-up. I was unable to check this and verify on short notice. There is also the fact that you have invested a lot more time in it than me. 👍

Issue #20310, opened 2017-09-11
Core/Movement: waypoint system issues [45$]


minkokomaung:
**Description:**
When you click Inconspicuous Mine Car, you'll be inside Mine Car and Scarlet Miner should pull that mine car toward Light's Point.
Currently it doesn't happen.

http://www.wowhead.com/quest=12701/massacre-at-lights-point

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 62f7debffdb0

**TDB version:**  TDB 3.3.5.63

**Operating system:** Windows 10 x64

dikkedeur:
i can confirm, the mine cart seems not to be working

Foereaper:
Seems to be happening after the following commit:

https://github.com/TrinityCore/TrinityCore/commit/7fff83d67526efff63867d41b9e036a19a9287b3

Foereaper:
Copied from issue 20341:

Several features seem to have stopped working as intended after the above mentioned commit and subsequent updates. I will list things I discover below as I go along:

**Quests:**

- Massacre at Light's Point(12701), the minecart no longer gets pulled down to the boats.
- Territorial Trespass (13051), the dragon no longer lands upon quest progression.

**Instances:**

- Azjol'Nerub, Hadronox keeps resetting as soon as one of the debuffs runs out. Impossible to complete without GM intervention.
- The Culling of Stratholme, Arthas keeps getting stuck on adds and no longer continues to follow his waypoints, impossible to complete without GM intervention.

I will keep this updated with more examples as I find them, but these are the ones I have noticed so far.

Infinito47:
I think a priority tag would suit well; there is a lot broken after 7fff83d

ConradBunton:
Violet hold works fine, for me.

Infinito47:
@ConradBunton On what commit are you?

tkrokli:
Yes, please always include TC commit rev. hash when confirming or disputing issues. A statement about what works or not on TC is almost worthless without the actual local TC rev. hash for comparison.

ConradBunton:
I'm on 20bc7d7

Foereaper:
After debugging Massacre at Light's Point(12701) (see chapter1.cpp, npc_scarlet_miner) I can see that, while Start() is called (EscortAI), the waypoints are never actually started, and WaypointReached() never reached at all, preventing any of the subsequent events from firing.

[...]

PR #20684, opened 2017-10-22
Core/SAI: allow to pause/resume waypoint for non-escorting NPCs

Wyreth:
**Changes proposed:**

Some SAI actions related to waypoints (SMART_ACTION_WP_PAUSE and SMART_ACTION_WP_STOP) wrongly check for SMART_ESCORT_ESCORTING. But NPCs that aren't involved in escort quests should be able to benefit from those actions too.

This change fixes many non-trivial waypoint SAI scripts that use SAI's waypoint scripting functions.

[Example of such a creature](http://www.wowhead.com/npc=18146).

**Target branch(es):**

- [x] 3.3.5
- [x] master (unsure)

**Issues addressed:** updates #20310 (only the ones related to SAI). 

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -128,9 +128,6 @@ bool SmartAI::LoadPath(uint32 entry)
 
 void SmartAI::PausePath(uint32 delay, bool forced)
 {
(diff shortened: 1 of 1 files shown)


Shauren (approved):

Shauren:
This one looks good to me

PR #20686, opened 2017-10-22
Core/AI: fix EscortAI functionality for summoned creatures

Wyreth:
**Changes proposed:**

The underlying issue is the way JustAppeared() is called.

The current flow for summoned creatures (and any other creature under the new dynamic spawns system) is the following:

1. Constructor
2. Script initialization (including Reset())
3. First AI tick calls JustAppeared() (which in turn calls Reset() again)

This causes scripts to reset, do their initialization and then reset again.

This only really affects summoned creatures, since they use dynamic spawn system by default. All other ~~EventAI~~ EscortAI cases can be disabled.

I'd have commented on #20310 to seek opinions, but it's locked. So, opinions here!

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:** updates (and probably closes, more testing needed) #20310.

### DIFF
--- a/src/server/game/AI/ScriptedAI/ScriptedEscortAI.cpp
+++ b/src/server/game/AI/ScriptedAI/ScriptedEscortAI.cpp
@@ -123,8 +123,6 @@ void EscortAI::JustDied(Unit* /*killer*/)
 
 void EscortAI::JustAppeared()
 {
-    _escortState = STATE_ESCORT_NONE;
-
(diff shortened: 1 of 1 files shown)


ccrs:
> All other EventAI...

Dont disturb the dead

Wyreth:
Yeah going to wash my hands and turn to Wicca for some cleansing rituals :D

PR #20694, opened 2017-10-23
Core/SAI: Reimplement EVENT_FRIENDLY_HEALTH_PCT

### LINKED ISSUE #20536: Core/SAI: SMART_EVENT_FRIENDLY_HEALTH_PCT is broken
**Description:**

**Current behaviour:**

**Branch(es):**

both

**TC rev. hash/commit:** 

**Steps to Reproduce**

Alternatively:

3. Nothing will ever happen

Teppic1:
**Changes proposed:**

EVENT_FRIENDLY_HEALTH_PCT is fundamentally broken, as any ACTION that uses a target will not work with it. It attempts to implement an action into an event, which doesn't work.

This change will reimplement it to be consistent with other events (like EVENT_FRIENDLY_HEALTH) and will allow the use of TARGET_ACTION_INVOKER, which is also consistent with other events.

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes #20536

**Tests performed:**

Builds on current 3.3.5 branch and tested with an NPC, works as expected.

A hard coded range is necessary as we only have four fields for an event - two are for the per cent range and two for the repeat timer.

All current SAI using this event is broken anway, but to fix it, all it needs is the target type changed to TARGET_ACTION_INVOKER, as the reimplementation is otherwise compatible with existing fields.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScriptMgr.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScriptMgr.cpp
@@ -868,21 +868,6 @@ bool SmartAIMgr::IsEventValid(SmartScriptHolder& e)
                     TC_LOG_ERROR("sql.sql", "SmartAIMgr: Entry %d SourceType %u Event %u Action %u has pct value above 100, skipped.", e.entryOrGuid, e.GetScriptType(), e.event_id, e.GetActionType());
                     return false;
                 }
-
-                switch (e.GetTargetType())
-                {
-                    case SMART_TARGET_CREATURE_RANGE:
-                    case SMART_TARGET_CREATURE_GUID:
-                    case SMART_TARGET_CREATURE_DISTANCE:
-                    case SMART_TARGET_CLOSEST_CREATURE:
-                    case SMART_TARGET_CLOSEST_PLAYER:
-                    case SMART_TARGET_PLAYER_RANGE:
-                    case SMART_TARGET_PLAYER_DISTANCE:
-                        break;
-                    default:
-                        TC_LOG_ERROR("sql.sql", "SmartAIMgr: Entry %d SourceType %u Event %u Action %u uses invalid target_type %u, skipped.", e.entryOrGuid, e.GetScriptType(), e.event_id, e.GetActionType(), e.GetTargetType());
-                        return false;
-                }
                 break;
             case SMART_EVENT_DISTANCE_CREATURE:
                 if (e.event.distance.guid == 0 && e.event.distance.entry == 0)
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -3155,40 +3155,20 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
             if (!me || !me->IsEngaged())
                 return;
 
-            ObjectVector targets;
-            switch (e.GetTargetType())
+            // Get any friendly creature missing health within 30 yard radius
+            Unit* target = DoSelectLowestHpFriendly(30.0f, 1);
+            if (!target || !target->IsInCombat())
             {
-                case SMART_TARGET_CREATURE_RANGE:
-                case SMART_TARGET_CREATURE_GUID:
-                case SMART_TARGET_CREATURE_DISTANCE:
-                case SMART_TARGET_CLOSEST_CREATURE:
-                case SMART_TARGET_CLOSEST_PLAYER:
-                case SMART_TARGET_PLAYER_RANGE:
-                case SMART_TARGET_PLAYER_DISTANCE:
-                    GetTargets(targets, e);
-                    break;
-                default:
-                    return;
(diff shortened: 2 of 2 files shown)


jackpoz:
I don't like the way the range is hardcoded.

Are both min and max required by the event ? One solution could be to have only either max or min, just like how SMART_EVENT_FRIENDLY_HEALTH works. Another solution would be to add a 5th parameter, changing how SAI works.

Killyana:
We cannot take the range directly from the spell that will be casted?

Teppic1:
With only four parameters available, two are needed for the repeat timer (no question there), which leaves two for the health range or for the distance range to check. The health range is obviously more important for this event.

One option would be to use a single value for max health and one for max distance but it wouldn't be very consistent with other SAI, where min/max is normal.

The event only checks for nearby friendly creatures at a given health, there's no connection to any spell (the action could be literally anything, not just a cast).

jackpoz:
Well if we want to keep EVENT_FRIENDLY_HEALTH_PCT consistent with EVENT_FRIENDLY_HEALTH, then the min/max health is not needed. these are the parameters for the EVENT_FRIENDLY_HEALTH one:
```
struct
{
  uint32 hpDeficit;
  uint32 radius;
  uint32 repeatMin;
  uint32 repeatMax;
} friendlyHealth;
```

if the min/max is really needed (and maybe it could be needed for the EVENT_FRIENDLY_HEALTH event too), then we could really just add a 5th parameters. 

do you have a case where you need both min and max ?

Teppic1:
SMART_EVENT_HEALT_PCT and SMART_EVENT_TARGET_HEALTH_PCT however do use min/max HP. The current (broken) implementation for FRIENDLY_HEALTH_PCT uses the same fields as these two so an advantage of keeping it the same was that existing scripts would "auto fix", as well as consistency. The disadvantage was fixing a distance for checking for friendly targets (this is totally unrelated to any action ranges, like spell casts). Adding extra fields to the table for one event seems a bit messy.

jackpoz:
one thing in this PR is not working: DoSelectLowestHpFriendly(..) returns a single Unit with the lowest hp but then the health % is compared to minHpPct. This means if there is an event with minHpPct set to 50% and 2 allies, one with 20% and one with 60%, the event will not be triggered.

Same goes if there are 2 allies and the lowest hp one is not in combat due to the IsInCombat() check.

jackpoz:
btw about the range topic, what's wrong with the old code about checking the target type and retrieving the targets in that way ?

Teppic1:
Because the only way to pass anything from the event code is to use TARGET_ACTION_INVOKER. The target is always determined by the action code, you can't use it for the event code as it'll just get ignored when the action uses the target itself.

jackpoz:
https://github.com/TrinityCore/TrinityCore/commit/1c60af632888433b29a27bee76e82e96632d4096 there you go, now you have 5 parameters

jackpoz:
any update on this ?

Teppic1:
Busy with loot work at the moment, will come back to it when I can focus.

jackpoz:
@Teppic1 got time to finish this PR ?

src/server/game/AI/SmartScripts/SmartScriptMgr.cpp
```diff
@@ -868,21 +868,6 @@ bool SmartAIMgr::IsEventValid(SmartScriptHolder& e)
                     TC_LOG_ERROR("sql.sql", "SmartAIMgr: Entry %d SourceType %u Event %u Action %u has pct value above 100, skipped.", e.entryOrGuid, e.GetScriptType(), e.event_id, e.GetActionType());
                     return false;
                 }
-
-                switch (e.GetTargetType())
```
jackpoz: could you describe a bit more in detail why this was removed ?

Teppic1: It's been a very long time since I did this so will have to look over it properly again. Off the top of my head though the event is just a broken hack, and it's trying to force specific targets (like range), when none of these actually work since they'll just get a new list of targets and everything originally checked for health % is lost.

jackpoz:
Continuing at https://github.com/TrinityCore/TrinityCore/pull/23377

jackpoz:
Another broken thing: you copied the parameters of EVENT_FRIENDLY_HEALTH_PCT  with min/max but then left the implementation of EVENT_FRIENDLY_HEALTH that uses only 1 parameter validation as HP deficit. This means that DoSelectLowestHpFriendly() will always find the lowest hp unit around, even if we set max % to 10

Rushor:
now we have eventparam5 and could use it here

Teppic1:
This ended up being rejected. The original problem still exists though in the current code (which is totally illogical), so any SAI that uses the event is broken.

jackpoz:
This PR was closed because no work was done for months,  even if we added the required support for additional parameters. 
It was continued at https://github.com/TrinityCore/TrinityCore/pull/23377 where a shitton of other issues were found. 
If you care about the issue,  please check that PR, fix the remaining issues and open a new one.

PR #20710, opened 2017-10-24
Core/SAI: keep a creature active while it's executing an ActionList

Wyreth:
**Changes proposed:**

When a creature is executing an ActionList, it would make sense for it to stay active and actually finish the ActionList even if there are no nearby players. This prevents complex SAI scripts from getting stuck in the middle of execution.

Every script should behave like this (think a NPC saying some lines of text after quest completion while the player hearthstones away, and anything that sets/removes npcflags and gets stuck).

**Target branch(es):**

- [x] 3.3.5
- [x] master


Teppic1:
This should probably also be done for long waypoints scripts too (e.g. Stitches walking across the whole of Duskwood).

Wyreth:
In those specific long events maybe it's better to just add/remove active state at start/end of script, this is a more generic case for every small script out there :P

r00ty-tc:
The thing is, setActive doesn't only keep a creature active. It has visibility effects too (it becomes grid wide visible).

Regarding the long waypoint creatures. I've had a generic core solution for this on the cards for a while. But it's some way down the priority list.

For this specific issue, the core solution wouldn't fix it automatically, it would create a way to "just" keep a creature active without changing visibility state.

Shauren:
Automatic objection to anything using setActive.

Wyreth:
Would it be enough to split setActive into two methods to handle active status and visibility separately?

r00ty-tc:
Well, my original intention was to create multiple options that could be specified on creature template/addon/creature/addon (whichever was appropriate) that would allow something like.

Activation:
- Normal (visual range)
- Grid (larger range)
- Zone (for large roaming creatures)
- IF possible, a list of grids that are part of the path

Visibility:
- Normal (default visibility range)
- Custom (specified range, not sure there's a real world use for it though)
- Grid wide (current maximum)
- Zone wide (Fel reaver et al), for gameobjects, think WG walls too.

Of course these modes would still need to be carefully applied still just like setActive, but would at least reduce the amount of unneeded activation, and allow more.

But, that's much more than you're doing right now. I know. It's also some way down the list of things I'd like to do and, as such not likely to happen soon.

Wyreth:
All that sounds very useful, yeah :P I'll see what I can do to help with that.

Naios:
/push @r00ty-tc I would also prefer the solution mentioned by you

jackpoz:
doesn't current TC already have split setActive and SetFarVisible ?

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -3354,15 +3354,22 @@ void SmartScript::UpdateTimer(SmartScriptHolder& e, uint32 const diff)
...
                         {
                             i->enableTimed = true;
+                            isActionListOver = false;
                             break;
                         }
                     }
+
+                    // Set object to non-active if there's nothing else to process
+                    if (isActionListOver)
+                        GetBaseObject()->setActive(false);
```
jackpoz: I wonder about manual setActive() calls, this will set it to false even if unintended. It should restore the previous value (but that complicates things a bit)

Wyrserth: @jackpoz is this acceptable? https://pastebin.com/CAvrscB8

Shauren:
This is still a no.

Wyrserth:
Keeping it alive until a better alternative comes up, for now this is the only way to avoid db scripts being paused when the last player leaves the area :P

Issue #8988, opened 2013-01-18
[Quest / NPC] Razormaw from quest 'Razormaw' does not attack [$15]


MrSmite:
__Rev:__ 98f502f32f092fe1bcb9e19c32c189d2ea552d19

__Bug:__

[Razormaw](http://www.wowhead.com/npc=17592) from quest [Razormaw](http://www.wowhead.com/quest=9689) does not fight, he just stands there while you attack.

__Notes:__

I'm not sure if it is because he becomes attackable before he lands or not. I was spamming Hunter's Mark which kept saying "invalid target". Once it actually cast I sent my pet in. We killed Razormaw without him attacking.

When I went to loot his body, I noticed he was still a few yards off the ground.

Killyana:
The inactive grid cell is a general issue, increasing Visibility.Distance.Continents  to 170 on server config fix the problem in this case.

For the script:

``` sql
UPDATE creature_template SET InhabitType=4, unit_flags=64 WHERE Entry=17592;
DELETE FROM `smart_scripts` WHERE `entryorguid`=17592 AND `source_type`=0 AND `id`IN(5, 6, 7);
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(17592, 0, 5, 6, 40, 0, 100, 0, 10, 17592, 0, 0, 101, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 'Razormaw - On Waypoint 10 Reached - Set Home Position'),
(17592, 0, 6, 0, 61, 0, 100, 0, 0, 0, 0, 0, 8, 2, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 'Razormaw - On Waypoint 10 Reached - Set Agressive'),
[...]

PR #20725, opened 2017-10-26
Core/Entities: remove grid-wide visibility from setActive and implement another method for it

Wyreth:
**Changes proposed:**

An attempt to make setActive less troublesome by moving the grid-wide visibility (which is never used by anything on retail as far as I know, but I might be wrong on that) to another bool.

**Target branch(es):**

- [x] 3.3.5
- [x] master

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -998,7 +998,7 @@ void MovementInfo::OutDebug()
 }
 
 WorldObject::WorldObject(bool isWorldObject) : WorldLocation(), LastUsedScriptID(0),
-m_name(""), m_isActive(false), m_isWorldObject(isWorldObject), m_zoneScript(nullptr),
+m_name(""), m_isActive(false), m_isVisibleFromFarAway(false), m_isWorldObject(isWorldObject), m_zoneScript(nullptr),
 m_transport(nullptr), m_zoneId(0), m_areaId(0), m_staticFloorZ(VMAP_INVALID_HEIGHT), m_currMap(nullptr), m_InstanceId(0),
 m_phaseMask(PHASEMASK_NORMAL), m_notifyflags(0)
 {
@@ -1056,6 +1056,14 @@ void WorldObject::setActive(bool on)
         else if (GetTypeId() == TYPEID_DYNAMICOBJECT)
             map->RemoveFromActive((DynamicObject*)this);
     }
+}
+
+void WorldObject::setVisibleFromFarAway(bool on)
+{
+    if (GetTypeId() == TYPEID_PLAYER)
+        return;
+
+    m_isVisibleFromFarAway = on;
 }
 
 void WorldObject::CleanupsBeforeDelete(bool /*finalCleanup*/)
@@ -1500,7 +1508,7 @@ float WorldObject::GetGridActivationRange() const
 
 float WorldObject::GetVisibilityRange() const
 {
-    if (isActiveObject() && !ToPlayer())
+    if (isVisibleFromFarAway() && !ToPlayer())
         return MAX_VISIBILITY_DISTANCE;
     else
         return GetMap()->GetVisibilityRange();
@@ -1512,7 +1520,7 @@ float WorldObject::GetSightRange(WorldObject const* target) const
     {
         if (ToPlayer())
         {
-            if (target && target->isActiveObject() && !target->ToPlayer())
+            if (target && target->isVisibleFromFarAway() && !target->ToPlayer())
                 return MAX_VISIBILITY_DISTANCE;
             else if (ToPlayer()->GetCinematicMgr()->IsOnCinematic())
                 return DEFAULT_VISIBILITY_INSTANCE;
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -998,7 +998,7 @@ void MovementInfo::OutDebug()
 }
 
 WorldObject::WorldObject(bool isWorldObject) : WorldLocation(), LastUsedScriptID(0),
-m_name(""), m_isActive(false), m_isVisibleFromFarAway(false), m_isWorldObject(isWorldObject), m_zoneScript(nullptr),
+m_name(""), m_isActive(false), m_isFarVisible(false), m_isWorldObject(isWorldObject), m_zoneScript(nullptr),
 m_transport(nullptr), m_zoneId(0), m_areaId(0), m_staticFloorZ(VMAP_INVALID_HEIGHT), m_currMap(nullptr), m_InstanceId(0),
 m_phaseMask(PHASEMASK_NORMAL), m_notifyflags(0)
 {
@@ -1058,12 +1058,12 @@ void WorldObject::setActive(bool on)
     }
 }
 
-void WorldObject::setVisibleFromFarAway(bool on)
+void WorldObject::SetFarVisible(bool on)
 {
     if (GetTypeId() == TYPEID_PLAYER)
         return;
 
-    m_isVisibleFromFarAway = on;
+    m_IsFarVisible = on;
 }
 
 void WorldObject::CleanupsBeforeDelete(bool /*finalCleanup*/)
@@ -1508,7 +1508,7 @@ float WorldObject::GetGridActivationRange() const
 
 float WorldObject::GetVisibilityRange() const
 {
-    if (isVisibleFromFarAway() && !ToPlayer())
+    if (IsFarVisible() && !ToPlayer())
         return MAX_VISIBILITY_DISTANCE;
     else
         return GetMap()->GetVisibilityRange();
@@ -1520,7 +1520,7 @@ float WorldObject::GetSightRange(WorldObject const* target) const
     {
         if (ToPlayer())
         {
-            if (target && target->isVisibleFromFarAway() && !target->ToPlayer())
+            if (target && target->IsFarVisible() && !target->ToPlayer())
                 return MAX_VISIBILITY_DISTANCE;
             else if (ToPlayer()->GetCinematicMgr()->IsOnCinematic())
                 return DEFAULT_VISIBILITY_INSTANCE;
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1063,7 +1063,7 @@ void WorldObject::SetFarVisible(bool on)
     if (GetTypeId() == TYPEID_PLAYER)
         return;
 
-    m_IsFarVisible = on;
+    m_isFarVisible = on;
 }
 
 void WorldObject::CleanupsBeforeDelete(bool /*finalCleanup*/)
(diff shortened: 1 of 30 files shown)


src/server/game/Entities/Object/Object.h
```diff
@@ -414,6 +414,8 @@ class TC_GAME_API WorldObject : public Object, public WorldLocation
 
         bool isActiveObject() const { return m_isActive; }
         void setActive(bool isActiveObject);
+        bool isVisibleFromFarAway() const { return m_isVisibleFromFarAway; }
+        void setVisibleFromFarAway(bool isVisibleFromFarAway);
```
ccrs:
```
SetVisibleFar
SetFarVisible
```
keep it simple

Wyreth: Done, thanks!

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1058,6 +1058,14 @@ void WorldObject::setActive(bool on)
     }
 }
 
+void WorldObject::SetFarVisible(bool on)
+{
+    if (GetTypeId() == TYPEID_PLAYER)
+        return;
+
+    m_IsFarVisible = on;
```
tkrokli:
```cpp
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Entities/Object/Object.cpp:1066:5:

 fatal error: use of undeclared identifier 'm_IsFarVisible'; did you mean 'm_isFarVisible'?
    m_IsFarVisible = on;
    ^~~~~~~~~~~~~~
    m_isFarVisible

/home/travis/build/TrinityCore/TrinityCore/src/server/game/Entities/Object/Object.h:449:14:

 note: 'm_isFarVisible' declared here
        bool m_isFarVisible;
             ^
```

Wyreth: Ouch, fixed. Thanks!

tkrokli:
BTW, the AppVeyor fail is caused by something in the latest changes in `src/server/game/Battlegrounds/Zones/BattlegroundAB.h` and nearby affected .cpp files. Something about `conversion from 'double' to 'const float' requires a narrowing conversion`.

offl:
c2f152d903cbbf16271189b86cfebd7eacce09f5?

Shauren:
Not enough. There might be setActive uses in scripts that used it exactly for visibility distance. I know of at least one such use - https://github.com/TrinityCore/TrinityCore/blob/618cd8cd49fd3abd35d0ddd3fb7f379b0a7941b6/src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp#L990

r00ty-tc:
Can I suggest that rather than adding extra functions, just adapt the existing setActive:

```c++
 void setActive(bool isActiveObject, bool extendVisibility = true);
```

That way, existing uses will still provide the extra visibility, but now it's also possible to only use it for the activity aspect.

Could also add a parameter to setActive on SAI to use the new mode too.

Shauren:
No, that is not a good idea, we want it to default to false (check and fix existing uses)

r00ty-tc:
Then, the two functions are fine and it's a case of find all instances of setActive, make a judgement call.

This is a lesson to @Wyreth. Nothing is as simple as it seems. 😆

Shauren:
*cough* `m_respawnCompatibilityMode`

Wyreth:
Nothing is ever as easy as we'd like it to be 😄 

I'd like to keep the two separate methods if possible - makes it easier to extend this to custom visibility values eventually™

About instances, isn't everything supposed to have max visibility range in instanced content? I vaguely recall being able to /target final bosses from the entrance in retail, although I wouldn't bet on it.

r00ty-tc:
@Wyreth the issue is, that it's fine to make this change. But, if we're visiting this, we should find ALL places setActive is used, and attempt to derive whether it was set for activation purposes, visibility purposes, or both. Then change those locations too.

This probably should include checking SAI, since if you make this change, suddenly all places where SAI sets something active, will no longer also set the visibility too. In most cases I expect that will be fine, but there might be a few exceptions.

Wyreth:
@r00ty-tc yeah, you're right.

Hopefully in almost every case the increased visibility is a bug and not intended - setActive is mostly used to keep the grid active so that the NPC can do its own thing without being interrupted. There's only a few exceptions I can think of (Fel Reaver, Putridus the Ancient and Wintergrasp walls, neither of which is currently affected, and the elevator @Shauren mentioned).

I've checked both SAI and core scripts - added SetFarVisible calls where I think it can be helpful to avoid issues. No SAI script needs it as far as I can tell - in all cases it's used to avoid interruptions in the script rather than for visibility.

Aokromes:
I remember one quest on azuremyst isle or the other draenei isles, where you have to summon a drake or something like tha and that uses SAI IIRC.

r00ty-tc:
Ah, I think it might be the SAI I did for the elite dragon group quest.. Razor something?

Wyreth:
Yep I know [that dragon](http://www.wowhead.com/quest=9689), but it shouldn't be visible from further than normal as far as I can tell :P unless the dragon not being visible affects waypoint movement, at which I'd just go sulk in a corner or something.

r00ty-tc:
https://github.com/TrinityCore/TrinityCore/issues/8988#issuecomment-81966255

That one there.

r00ty-tc:
Yeah, could be right that actually visibility shouldn't be changed there.

Wyreth:
Turns out he should actually be visible from any distance when spawned, checked other videos where it's clearly visible.

Good find, back to the drawing board :P

src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
```diff
@@ -1002,9 +1002,15 @@ class instance_icecrown_citadel : public InstanceMapScript
                         // set the platform as active object to dramatically increase visibility range
                         // note: "active" gameobjects do not block grid unloading
                         if (GameObject* precipice = instance->GetGameObject(ArthasPrecipiceGUID))
+                        {
                             precipice->setActive(state == IN_PROGRESS);
```
Shauren: active should be removed here - being the author of this script i guarantee it

Wyreth: Only for precipice, or for the platform too?

Wyreth:
Just a small note about Razormaw - it doesn't seem to work for me, even after reverting the changes in this PR. It will start working correctly if I fly to the dragon's spawn position. But otherwise he just spawns into an unactive grid and doesn't receive updates.

Wyreth:
Is increasing grid activation distance a bad idea? Right now it's equal to the view distance, which means any event targeting a nearby grid (like Razormaw) will not happen unless the player physically visits the nearby grids.

Playing around with WorldObject::GetGridActivationRange() returning higher values turned out to allow improving certain edge-case events (Razormaw, #20728 to mention some) as well as anything else that should happen even if outside the current nearby player's view range,, while only adding a few MBs to overall ram usage (more grids get loaded).

r00ty-tc:
The main reason to not activate creatures outside view range is to save CPU (not calling Update() on those creatures). It's the realization of "If a tree falls and there's no-one around, does it make a sound" in wow terms, it doesn't even fall..

Now, I think this is Blizz like. Back in the end of Wrath and beginning of cata, my old realm REALLY dropped in population, and I mean REALLY low. Quite often, you'd see these same artifacts. Flying on an FTP for example it'd be fairly normal to see shaman creatures and similar casting their expired buffs back onto themselves as you passed. So, I don't think we're doing things wrong here.

I do thing there are situations when creatures need activating out of range. Always active should be VERY rate (setActive), but we should really have some larger scopes available for specific use. General use, visual range is fine and is needed to keep CPU usage down.

Wyreth:
I see, makes sense :P I was on a fairly populated realm back then, so I don't think I ever noticed that happening.

I guess it could be solved by allowing specific creatures to override grid activation range (instead of using the default map range), that is assuming creatures themselves can activate grids (I'm honestly not sure).

I think I'll step down from related issues for now and try to get a better feeling on how the grid system works - I feel like I'm just making mostly wrong assumptions right now :P

r00ty-tc:
The main wrong assumption is that a grid is activated. Generally grids are only used as a way of reducing the number of items in any search, and as a block of objects to load/unload.

Activation happens at a much smaller level. Cell or visible range.

malcrom:
Certain creatures need to remain active as long as the area is active.

Faq:
@malcrom some example would be nice.

Aokromes:
i remember some creatures with huge paths and waiting where they start / end the path the creature reachs the end.

r00ty-tc:
Oh it definitely works (and worked on retail as far back as BC). We should handle it, as right now people might well wait forever

src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
Shauren: both

Wyreth: Done, thanks!

r00ty-tc:
@Wyreth is there anything else you want to do here, have you checked all uses and applied the extended visibility whrere applicable?

Killyana:
We need 'SetFarVisible' on the DB: creature_template, creature, gameobjet_template and gameobject

Aokromes:
maybe custom flag or whatever.

r00ty-tc:
@Killyana I don't think that's for this change.

That's part of the visibility/activity scope overhaul that I'd like to do or someone else can (just say so else there will be duplication). There will be some way (either addon or on template/creature instance) to define these parameters (and it can be done in code).

This is just a change for the setActive for the interim for current scripting use.

dikkedeur:
i came here since i was looking for a sollution for the quest razormaw, after using the bones at the pyre you here the sound ( kick sound? ) anyway.. the dragon is not spawned or at least i dont see it.
what would be the best way to fix this, since i saw aokromes mentioning something about visibility?
is that the visibility in the config?

Kittnz:
FYI
https://github.com/LightsHope/server/commit/c58d027807b639cc6affc1c9163a1b936bcfa4bc
https://github.com/LightsHope/server/commit/a7f8c58c6077f1a57ad808428d06f113d30caa84

Mihapro:
👍https://imgur.com/a/X6HGP
Tested in Halls of Origination where this is also needed and it works great. When far away (without going within "active" range), creatures normally move on paths, door open - Anraphet won't come down, but that's what far-visibility is for, not loading AI, right?  

💭 Also it is yet to be discovered if there's a flag for it - all of bosses and visual bunnies I find have either unit_flag2 0x00400000 or 0x00200000 set.

src/server/game/Entities/Object/Object.h
```diff
@@ -414,6 +414,8 @@ class TC_GAME_API WorldObject : public Object, public WorldLocation
 
         bool isActiveObject() const { return m_isActive; }
         void setActive(bool isActiveObject);
+        bool IsFarVisible() const { return m_isFarVisible; }
+        void SetFarVisible(bool m_isFarVisible);
```
Mihapro: m_isFarVisible => visible ❓ 💭

jackpoz:
thanks for the PR :)

tkrokli:
I hope @Wyreth will be online again this year to notice the merge.

PR #20736, opened 2017-10-27
Scripts/Deadmines: New script for Edwin VanCleef

Teppic1:
**Changes proposed:**

Move VanCleef from SAI to C++ to correct various evade/boss/guard behaviour. Also fix issues with the existing scripting.

- Blackguards are now summoned properly as part of the encounter
- Blackguards and VanCleef will despawn/respawn on evade
- Fixed scripted actions/speech on health
- Removed Blackguards from world DB
- Corrected VanCleef's damage

**Target branch(es):** 3.3.5

- [x] 3.3.5

Built and tested. 

None as yet.


jackpoz:
> Move VanCleef from SAI to C++

lol half of the PRs are about moving a script from a system to another xD

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,176 @@
...
+                Talk(SAY_KILL);
+        }
+
+        void EnterEvadeMode(EvadeReason /*why*/) override
+        {
+            summons.DespawnAll();
+            _DespawnAtEvade();
+        }
+
+        void Reset() override
```
jackpoz: should this reset the _phase too ?

Aokromes:
well, bosses are supposed to be c++ quest etc if possible SAI :P

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,176 @@
...
+    boss_vancleef() : CreatureScript("boss_vancleef") { }
+
+    struct boss_vancleefAI: public BossAI
+    {
+        boss_vancleefAI(Creature* creature) : BossAI(creature, 0)
+        {
+            _phase = PHASE_NORMAL;
+        }
+
+        void Initialize()
```
jackpoz: as a standard we use Initialize() only to set the class fields to a default value, called in the ctor and in Reset(). Maybe this code fits better in Reset().

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,176 @@
...
+            {
+                DoMeleeAttackIfReady();
+            });
+        }
+
+        void _SummonBlackguards()
+        {
+            for (auto BlackguardPosition : BlackguardPositions)
+                DoSummon(NPC_BLACKGUARD, BlackguardPosition, 60000,
+                        TEMPSUMMON_CORPSE_TIMED_DESPAWN);
```
jackpoz: please keep all DoSummon() parameters on 1 single line, it's a short line anyway

jackpoz:
The PR overall looks good, there are some minor things here and there. I dunno if we should handle the "from 100% hp to 20% hp" in a particular way.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
Keader: yeap, it should be in reset

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,176 @@
...
+        {
+            // don't repeat
+            if (_phase == Phase)
+                return;
+
+            _phase = Phase;
+
+            switch (_phase)
+            {
+            case PHASE_66_PCT:
```
Keader: identation

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+            scheduler.Update(diff, [this]
+            {
+                DoMeleeAttackIfReady();
+            });
+        }
+
+        void _SummonBlackguards()
+        {
+            for (auto BlackguardPosition : BlackguardPositions)
```
Keader: why auto?

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,176 @@
...
+                else if (_phase == PHASE_50_PCT && HealthBelowPct(33))
+                    _SwitchPhase(PHASE_33_PCT);
+                else if (_phase == PHASE_33_PCT && HealthBelowPct(25))
+                    _SwitchPhase(PHASE_25_PCT);
+
+                task.Repeat(Seconds(1));
+            });
+        }
+
+    protected:
```
Keader: private

Keader:
What you say about use new script model?
Like in: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Outland/BlackTemple/boss_illidan.cpp#L1360-L1410

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        void _SummonBlackguards()
+        {
+            for (auto BlackguardPosition : BlackguardPositions)
+                DoSummon(NPC_BLACKGUARD, BlackguardPosition, 60000,
+                        TEMPSUMMON_CORPSE_TIMED_DESPAWN);
+        }
+
+        void _SwitchPhase(uint8 Phase)
```
ccrs: "_" points that this is private, make it private then

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,176 @@
...
+            _DespawnAtEvade();
+        }
+
+        void Reset() override
+        {
+            BossAI::Reset();
+            Initialize();
+        }
+
+        void UpdateAI(uint32 diff) override
```
ccrs: no need to override this, its the class default behaviour IIRC

offl:
Perhaps everything here can be done via SAI >_< Maybe Killyana can confirm that
I also can't undersand the manipulating with health, why he should say or cast only between 25-50%?

Aokromes:
^^ https://github.com/TrinityCore/TrinityCore/pull/20736#issuecomment-340165004

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
Teppic1: After evade he'll respawn and so it gets reset anyway.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
Teppic1: Ok, I put it there mainly because the compiler complained if I didn't :)

Teppic1:
Also if it's left to SAI when he evades he will just run back to his spawnpoint, which isn't how bosses are supposed to behave. I originally tried with SAI using a spawn group for his initial guards but had multiple issues with them not despawning properly, him not aggroing them, etc.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
Teppic1: The variable's only used for the loop, so why not? :)

Keader: Naah, put position there
we only use auto in some iterators case

ForesterDev:
I think, is better to use CreatureGroups for guards instead of hardcoding summon

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,170 @@
...
+                Talk(SAY_KILL);
+        }
+
+        void EnterEvadeMode(EvadeReason /*why*/) override
+        {
+            summons.DespawnAll();
+            _DespawnAtEvade();
+        }
+
+        void UpdateAI(uint32 diff)
```
ccrs: uh? just remove the whole thing

ForesterDev: Why tou override this function? It's the same implementation as https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp#L542

ForesterDev:
Also, you should use new script model according https://github.com/TrinityCore/TrinityCore/commit/fb87ac8e8da3d215d4988b8c2fa25d7a98cf9ee0#diff-5d3914b5fea050ec9beba58cf348400d

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
Teppic1: I was following what I saw in existing scripts, but it's removed now.

Teppic1:
Tested without the ``UpdateAI()`` override and he doesn't do any events any more. With it back in it works again.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,176 @@
...
+            default:
+                break;
+            }
+        }
+
+        void ScheduleTasks() override
+        {
+            scheduler.Schedule(Seconds(1), [this](TaskContext task)
+            {
+                if (_phase == PHASE_NORMAL && HealthBelowPct(66))
```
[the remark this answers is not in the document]
Teppic1: At the moment everything "skipped" will be queued up and executed, I'm not sure if that's desirable or not.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
Teppic1: Changed to all your suggestions, plus used the new model.

Keader:
> Tested without the UpdateAI() override and he doesn't do any events any more. With it back in it works again.

It's because you using TaskScheduler to handle with events.
You have 2 choices here: Or keep your UpdateAI() or make a new PR adding scheduler.Update(diff) here: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp#L547

Teppic1:
@Keader It's possible just to add the line ``scheduler.Update(diff);`` to ``BossAI::UpdateAI()`` and it not break tons of stuff? I tried it and removed the ``UpdateAI()`` from my code and it was fine as far as that was concerned.

Keader:
will not, you just making generic UpdateAI update all events based in TaskScheduler
It should be default behavior, idk is not present in current revs...
Is better add this line in a new PR

Keader:
Hmm
Thinking better, just restaure your UpdateAI.
We always use EventMap in events, update TaskScheduler too is unecessary.

nawuko:
I would rather not recommend to use licensed files (only remove the header). Not only can this be troublesome from the legal site for you, it also is a huge problem for OSS repositories like TrinityCore to get a DMCA takedown.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,160 @@
...
+Position const BlackguardPositions[] =
+{
+        { -78.2791f, -824.784f, 40.0007f, 2.93215f },
+        { -77.8071f, -815.097f, 40.0188f, 3.26377f }
+};
+
+struct boss_vancleef : public BossAI
+{
+    public:
+        boss_vancleef(Creature* creature) : BossAI(creature, 0) { }
```
Keader: magic number
should have a DATA_BOSS_VANCLEEF or something like that in deadmines.h
Btw deadmines.h should be called Deadmines.h

Teppic1:
@nawuko I spent hours looking through multiple files (including one you wrote) to learn TC's system, it's not simply your file with a header change.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
Teppic1: Yeah I wondered about that but since there's only one boss with BossAI I didn't know if it would be necessary.

Keader: yes, is

Teppic1: That's fixed now. I added Mr Smite since he's currently the only other boss scripted, but doesn't use BossAI (that probably needs to be rewritten too)

Teppic1:
@tkrokli  - this was already done, the change below was local only

> Hmm
> Thinking better, just restaure your UpdateAI.
> We always use EventMap in events, update TaskScheduler too is unecessary.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,160 @@
...
+            if (!UpdateVictim())
+                return;
+
+            scheduler.Update(diff, [this]
+            {
+                DoMeleeAttackIfReady();
+            });
+        }
+
+        void ScheduleTasks() override
```
Keader: What Nawuko suggested is true.
This phase events should be handled in DamageTaken hook instead a 1 sec event

Teppic1: I can rewrite that, but it is wasting my time if it'll never get accepted due to false accusations.

tkrokli:
>  @tkrokli - this was already done, the change below was local only

Well, my reason for changing the label was more based on the fact that it looked like there are still some details to sort out. But if it will make you feel better, I can change the label back while still leaving the decision on when to merge up to the main developer members.

Teppic1:
I've addressed all the comments, except the last one which I'm waiting for a response on.

Keader:
@nawuko has some prof of "stolen work" ?

tkrokli:
Would be a good idea to be very clear about the details for such accusations.
What has been said so far seems cryptic and/or obscure to me.

Teppic1:
I can provide a copy of the file he is claiming is just a whitespace change (it's licensed under the GPL2 after all). There are many similarities since this is a pretty simple script, not a raid boss, and so it's naturally overriding the same functions etc.

The situation is that we were working together on a TC derived emulator, and so he distributed his code to me to run on the server I was paying for. If I wanted to upload something of his I'd not choose some simple Deadmines boss that I'm trying to learn TC scripting by doing myself.

nawuko:
> I can provide a copy of the file he is claiming is just a whitespace change (it's licensed under the GPL2 after all).

Troublesome from a legal standpoint, and also wrong.

> There are many similarities since this is a pretty simple script

And so you end up with same function names, same timing, same percentages (which was made up by me and is not 100% true from retail). Even the same file structure (where function are withing in class). Even small errors are from me are included.

> If I wanted to upload something of his I'd not choose some simple Deadmines boss that I'm trying to learn TC scripting by doing myself.

So lets upload DBC files directly from the WoW Client, we are all here to learn something right?

 **No** - this is not how this works, you have to respect copyright even if its "for learning". The reason why you do it dosen't justify the action.

Also, like i said i have no problem to get inspired, but i have a problem with using my copyrighted work (and offering to distribute it). Also you just **admitted** in a Private Reddit message that you **took parts from my work**. 

So to close this, since the script has come a long way and is now far from the original, you can continue your work. But i highly discourage you from taking parts from my work again **without permission.** 

Also to add to the conversation, i would remove the ::EnterEvadeMode code section (the base class handles the despawn of summons.). Since the behavior from _DespawnAtEvade can A) cause some respawn issues atm. B) Is not blizzlike at all for classic/tbc dungeon bosses. If you don't use _DespawnAtEvade their is no reason to summon the first 2 guards (in the ::Reset overwrite) via script. Since this part is also not blizzlike, you where able to kill the guards during an encounter -> wipe and not fight them again. With this implementation you have to fight them every time the boss resets.

Teppic1:
Here is a copy of the file this ridiculous drama is about:

https://pastebin.com/mMs91NE8

And here is the file in this PR:

https://github.com/Teppic1/TrinityCore/blob/815aa0e4610646270291c5ad0bc7d12f3d82c3b8/src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp

Some things are identical, since they are identical in 1000 other TC scripts too. Some things are not: the guard despawn in this file doesn't work, it does in mine; I use the Blizzard spell 5200 to summon guards in combat, this file uses DoSummon(). There are lots of other changes as can be clearly seen.

The part I used from the above file was the phase section, which seemed perfectly suitable and mostly trivial, since it's just if/else. It's natural to see small sections of code in any file you view and reuse them, which is part of the reason the GPL exists.

I would appreciate any further drama be directed to me instead of made public in this extremely immature and offensive way.

Wyreth:
All this talk about copyright is quite discouraging, since we're all mostly here to learn and help improve on something we all love (I guess?).

I'm not sure if a bunch of variables and API calls (which is what TC scripting is, after all) fall into the copyright laws. But that's beyond the point.

nawuko:
> Some things are identical, since they are identical in 1000 other TC scripts too

First of all, you are admitting that you took parts. Second of all even the variable names are the same for the function _SummonGuards (which has also the same name, and is on the same place in the class)

> The part I used from the above file was the phase section

Clearly not the case, explained above and in the above posting.

> I would appreciate any further drama be directed to me instead of made public in this extremely immature and offensive way.

Unable to read? Also you are the guy that started with insults in the private reddit message, not me.

@Wyreth Yes and no, GPL is really the wrong license for TC (and they know it, but its impossible to change that since you would need an agreement from everyone that has contributed [if the code from them is still in there]). Since a server does not distribute any form of binary many requirements from GPL are not valid.

Teppic1:
@Wyreth In a script this simple it's little more than a set of overrides that are used identically in every other current TC script that needs to perform the same actions. It made sense to choose a boss like this to work on rather than something that does actually involve non-trivial logic.

Kittnz:
Will be reopened tomorrow.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,142 @@
...
+                    Talk(SAY_THREE);
+                    break;
+                default:
+                    break;
+            }
+        }
+
+        void DamageTaken(Unit* /*attacker*/, uint32& /*damage*/) override
+        {
+            if (_phase == PHASE_ONE && HealthBelowPct(66))
```
ccrs: HealthBelowPctDamaged is better
Also I'd change the order of the checks, example scenario: a level 80 uses a spell that brings the boss to 5% health, which phase should be triggered? One or Last?

ccrs: And also, DamageTaken is NOT a standard in TC for phase switching
I personally prefer to check it on update since DamageTaken can be called multiple times on the same update tick, from different damage incomes, giving the chance to multiple same triggered actions

jackpoz: @ccrs I wondered the same about hp % at https://github.com/TrinityCore/TrinityCore/pull/20736#issuecomment-340165344

Keader: you can control multiple calls using a variable, i dont see any problem about it.
But i agree with you, is better change the order.

Teppic1:
I thought it might be better just to get rid of the phase stuff, do a flag check every time for the guard summons (since this is max priority) and then fall through to the speech checks, I guess in reverse order from what it currently is.

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,128 @@
...
+struct boss_vancleef : public BossAI
+{
+    public:
+        boss_vancleef(Creature* creature) : BossAI(creature, DATA_VANCLEEF) { }
+
+        void Reset() override
+        {
+            BossAI::Reset();
+
+            _guardsCalled = false;
```
jackpoz: please initialize all class field members in the constructor too to avoid static analysis issues

jackpoz (changes requested):
The script looks all good except a couple of class fields that need to be initialized in the constructor.
Other than that, is the PR ready to be merged ?

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
```diff
@@ -0,0 +1,128 @@
...
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "ScriptMgr.h"
+#include "ScriptedCreature.h"
```
Aokromes: move the last include between the 2 script* like on https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/IcecrownCitadel/boss_lord_marrowgar.cpp

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
tkrokli: Looks like this was done in the following commit (https://github.com/TrinityCore/TrinityCore/pull/20736/commits/7664213442fd7e97259d5c23c78cae38b1065c1d).
Does the PR look good enough for your approval now? :-)

Teppic1: Yeah I forgot to comment here that I'd addressed that (all the other comments made have been addressed too).

jackpoz (approved):

jackpoz:
Thanks for the PR

src/server/scripts/EasternKingdoms/Deadmines/boss_vancleef.cpp
ariel-: A little late but IIRC blizz uses update to check hp

PR #20746, opened 2017-10-30
Scripts/Fireland: Fix static analysis issues CID 1379866 and CID 1379865

DDuarte:
Hacktoberfest 2/4


offl:
Since Oktoberfest is over, maybe better to call it Hacklloween

PR #20748, opened 2017-10-30
Scripts: Fix static analysis issues CID 1381648, 1379869, 1377302 and…

DDuarte:
Hacktoberfest 4/4 :spam:

### DIFF
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_sindragosa.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_sindragosa.cpp
@@ -1549,7 +1549,7 @@ class spell_frostwarden_handler_focus_fire : public SpellScriptLoader
                 PreventDefaultAction();
                 if (Unit* caster = GetCaster())
                 {
-                    if (SpellEffectInfo const* effect = GetSpellInfo()->GetEffect(target->GetMap()->GetDifficultyID(), EFFECT_1))
+                    if (SpellEffectInfo const* effect = GetSpellInfo()->GetEffect(caster->GetMap()->GetDifficultyID(), EFFECT_1))
                         caster->AddThreat(GetTarget(), -float(effect->CalcValue()));
                     caster->GetAI()->SetData(DATA_WHELP_MARKER, 0);
                 }
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_sindragosa.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_sindragosa.cpp
@@ -1549,7 +1549,8 @@ class spell_frostwarden_handler_focus_fire : public SpellScriptLoader
                 PreventDefaultAction();
                 if (Unit* caster = GetCaster())
                 {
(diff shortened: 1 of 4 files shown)


src/server/scripts/Northrend/IcecrownCitadel/boss_sindragosa.cpp
```diff
@@ -1549,7 +1549,8 @@ class spell_frostwarden_handler_focus_fire : public SpellScriptLoader
                 PreventDefaultAction();
                 if (Unit* caster = GetCaster())
                 {
-                    caster->AddThreat(GetTarget(), -float(GetSpellInfo()->GetEffect(caster->GetMap()->GetDifficultyID(), EFFECT_1)->CalcValue()));
+                    if (SpellEffectInfo const* effect = GetSpellInfo()->GetEffect(target->GetMap()->GetDifficultyID(), EFFECT_1))
```
Keader: `if (SpellEffectInfo const* effect = GetSpellInfo()->GetEffect(caster->GetMap()->GetDifficultyID(), EFFECT_1))`

Riztazz:
@DDuarte fk that was smart, should have done that as well - static analysis that is :P

src/server/scripts/Northrend/IcecrownCitadel/boss_sindragosa.cpp
tkrokli: Yes, travis-ci points out `use of undeclared identifier 'target'` which indicates that it should be `caster` like before.

tkrokli:
Hmm. In hindsight, this change may not be the intended fix here, based on the static analysis?
```diff
-                    if (SpellEffectInfo const* effect = GetSpellInfo()->GetEffect(target->GetMap()->GetDifficultyID(), EFFECT_1))
+                    if (SpellEffectInfo const* effect = GetSpellInfo()->GetEffect(caster->GetMap()->GetDifficultyID(), EFFECT_1))
```
(sorry if this was an unwanted/wrong correction, but if `target` should be used here, it needs to be declared.)

DDuarte:
Thanks, good catch

tkrokli:
Thank Keader, he pointed it out first. Then again, it was pretty obvious in the travis-ci output. ;-)

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -265,8 +265,13 @@ class spell_sha_earth_shield : public SpellScriptLoader
 
             bool CheckProc(ProcEventInfo& eventInfo)
             {
-                if (!eventInfo.GetDamageInfo() || !HasEffect(EFFECT_1) || eventInfo.GetDamageInfo()->GetDamage() < GetTarget()->CountPctFromMaxHealth(GetEffect(EFFECT_1)->GetAmount()))
+                if (!eventInfo.GetDamageInfo() || !HasEffect(EFFECT_1))
                     return false;
+
+                if (AuraEffect* effect = GetEffect(EFFECT_1))
```
Shauren: False positive - HasEffect takes care of it (and more since it not only checks that it exists on aura but also that it is applied on current target). Don't fix.

Shauren (dismissed):

src/server/scripts/Spells/spell_shaman.cpp
Aokromes: @DDuarte

PR #20757, opened 2017-10-31
Core/Entities: add DB-support for conversation actorGuids

funjoker:
**Changes proposed:**

-  Added the option to add permanent spawns
-  Use only for "creature_text" like conversations.

**Target branch(es):** 3.3.5/master

- [ ] 3.3.5
- [x] master

**Issues addressed:** 

**Tests performed:** (Does it build, tested in-game, etc.)
Tested Malfurion Ysera conversation in Val'sharah, with sniffed conversation data.

wowscrnshot_103117_041118
wowscrnshot_103117_041125
wowscrnshot_103117_041132
wowscrnshot_103117_041143

**Known issues and TODO list:** (add/remove lines as needed)

- [ ] Whatever TC wants
- [ ]

PR #20758, opened 2017-10-31
Core/Scripting: implement Conversation scripts

funjoker:
**Changes proposed:**

-  Option to make ConversationScripts
-  Added ScriptName column to conversation_template

- [x] master

Same tests as in #20757 


src/server/game/Scripting/ScriptMgr.h
```diff
@@ -879,6 +880,17 @@ class TC_GAME_API SceneScript : public ScriptObject
...
 
+class TC_GAME_API ConversationScript : public ScriptObject
+{
+protected:
+// Manages registration, loading, and execution of scripts.
+    ConversationScript(const char* name);
+
+public:
+
+    virtual void OnConversationCreate(Conversation* /*conversation*/, Unit* /*creator*/) { }
```
Traesh: Nice PR, but you should not rely on user calling "SetValidState", instead i think you should make "OnConversationCreate" return a boolean (default true) and set Conversation state using this boolean, so the user just have to return true or false from his script

src/server/game/Entities/Conversation/Conversation.cpp
```diff
@@ -138,6 +140,16 @@ bool Conversation::Create(ObjectGuid::LowType lowGuid, uint32 conversationEntry,
...
         SetDynamicStructuredValue(CONVERSATION_DYNAMIC_FIELD_LINES, linesIndex++, line);
 
+    sScriptMgr->OnConversationCreate(this, creator);
+
+    // return if there is no actor added (actors start at idx = 0), missing actors cause client crash
+    if (!GetDynamicStructuredValue<ConversationDynamicFieldActor>(CONVERSATION_DYNAMIC_FIELD_ACTORS, 0))
+        return false;
+
+    // option to prevent AddToMap by external script, e.g. ConversationScript or InstanceScript
+    if (!_isValid)
```
joschiwald: like i already said to you, don't use such boolean that the user have to set
the crash does not happen cause of missing actor on index 0, it happens because of the line does not have an actor
you need to verifiy that your conversation have actors for all lines, if that fails, add a log message

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -113,6 +115,10 @@ template<>
 struct is_script_database_bound<SceneScript>
     : std::true_type { };
 
+template<>
+struct is_script_database_bound<ConversationScript>
+: std::true_type { };
+
```
joschiwald: did you check that hotswap support too?

src/server/game/Entities/Conversation/Conversation.cpp
funjoker: Since the data is taken by sniff and it is unkown how many actors are there until the developer adds actors, i donno how to Check it if you dont know it.

joschiwald: you know it, actors depend on lines

funjoker: Ah yeah ok. Gonna add a Check for it.

src/server/game/Scripting/ScriptMgr.cpp
funjoker: Never worked with hotswap will read it

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -778,6 +784,35 @@ class ScriptRegistrySwapHooks<SpellScriptLoader, Base>
     bool swapped;
 };
 
+/// This hook is responsible for swapping ConversationScript's
```
funjoker: Never worked with hotswap.
If it's incorrect feel free to adjust :)

src/server/game/Entities/Conversation/Conversation.cpp
```diff
@@ -135,9 +137,25 @@ bool Conversation::Create(ObjectGuid::LowType lowGuid, uint32 conversationEntry,
...
+    std::list<uint16> _totalActorIdx;
     for (ConversationLineTemplate const* line : conversationTemplate->Lines)
+    {
         SetDynamicStructuredValue(CONVERSATION_DYNAMIC_FIELD_LINES, linesIndex++, line);
 
+        _totalActorIdx.push_back(line->ActorIdx);
+    }
+
+    _totalActorIdx.sort();
+    _totalActorIdx.unique();
```
joschiwald: use std::set to prevent duplicates and they are sorted by default

src/server/game/Entities/Conversation/Conversation.cpp
```diff
@@ -137,18 +137,24 @@ bool Conversation::Create(ObjectGuid::LowType lowGuid, uint32 conversationEntry,
     }
 
     uint16 linesIndex = 0;
+    std::list<uint16> _totalActorIdx;
```
Shauren: local var naming

Shauren (changes requested):

src/server/game/Entities/Conversation/Conversation.cpp
funjoker: si senor

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -879,6 +880,24 @@ class TC_GAME_API SceneScript : public ScriptObject
...
+    public:
+    
+        // Called when Conversation is created but not added to Map yet
+        virtual bool OnConversationCreate(Conversation* /*conversation*/, Unit* /*creator*/) { return true; }
+
+        // Called when an Conversation::AddActor is used
+        virtual void OnAddActor(Conversation* /*conversation*/, ObjectGuid /*actorGuid*/, uint16 /*actorIdx*/) { }
+
+        // Called when a Participant is added to conversation
+        virtual void OnAddParticipant(Conversation* /*conversation*/, ObjectGuid /*participantGuid*/) { }
```
joschiwald: can use provide an example where OnAddActor and OnAddParticipant is useful for?

funjoker: No idea yet. But thought it might be useful for someone

src/server/game/Globals/ConversationDataStore.cpp
```diff
@@ -137,6 +138,7 @@ void ConversationDataStore::LoadConversationTemplates()
             conversationTemplate.Id                 = fields[0].GetUInt32();
             conversationTemplate.FirstLineId        = fields[1].GetUInt32();
             conversationTemplate.LastLineEndTime    = fields[2].GetUInt32();
+            conversationTemplate.ScriptId           = sObjectMgr->GetScriptId(fields[3].GetCString());
```
joschiwald: GetString()

src/server/scripts/World/conversation_scripts.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
+ * Copyright (C) 2008-2017 TrinityCore <https://www.trinitycore.org/>
```
Shauren: lul, im pretty sure its http in all headers

joschiwald: should be changed on next copyright update

joschiwald:
merging when travis succeeds

PR #20768, opened 2017-10-31
Scripts/ScarletEnclave: Stable Master Kitrik's spell Repo (52265)

Wyreth:
**Changes proposed:**

Script spell [Repo](http://www.wowhead.com/spell=52265) used by NPC [Stable Master Kitrik](http://www.wowhead.com/npc=28683) so that it actually dismounts and attacks the player.

**Target branch(es):**

- [x] 3.3.5
- [x] master

### DIFF
--- a/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
+++ b/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
@@ -1234,7 +1234,45 @@ class npc_scarlet_miner : public CreatureScript
         }
 };
 
-// npc 28912 quest 17217 boss 29001 mob 29007 go 191092
+enum StableMasterRepo
+{
+    SPELL_REPO = 52265
+};
+
+class spell_stable_master_repo : public SpellScriptLoader
+{
+public:
+    spell_stable_master_repo() : SpellScriptLoader("spell_stable_master_repo") { }
+
+    class spell_stable_master_repo_AuraScript : public AuraScript
+    {
+        PrepareAuraScript(spell_stable_master_repo_AuraScript);
+
+        bool Validate(SpellInfo const* /*spellInfo*/) override
+        {
+            return ValidateSpellInfo({ SPELL_REPO });
+        }
+
+        void HandleDummy(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+        {
+            Creature* target = GetTarget()->ToCreature();
+            if (Unit* owner = target->GetCharmerOrOwner())
+                GetCaster()->GetAI()->AttackStart(owner);
+
+            target->DespawnOrUnsummon(Seconds(2));
+        }
+
+        void Register() override
+        {
+            AfterEffectApply += AuraEffectApplyFn(spell_stable_master_repo_AuraScript::HandleDummy, EFFECT_0, SPELL_AURA_DUMMY, AURA_EFFECT_HANDLE_REAL);
+        }
+    };
+
+    AuraScript* GetAuraScript() const override
+    {
+        return new spell_stable_master_repo_AuraScript();
+    }
+};
 
 void AddSC_the_scarlet_enclave_c1()
 {
@@ -1250,4 +1288,5 @@ void AddSC_the_scarlet_enclave_c1()
     new npc_scarlet_ghoul();
     new npc_scarlet_miner();
     new npc_scarlet_miner_cart();
+    new spell_stable_master_repo();
 }
--- a/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
+++ b/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
@@ -1255,6 +1255,9 @@ public:
 
         void HandleDummy(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
         {
+            if (target->GetTypeId() != TYPEID_UNIT)
+                return;
+
             Creature* target = GetTarget()->ToCreature();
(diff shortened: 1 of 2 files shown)


tkrokli:
F.Y.I. -- about the PR title change, the head part should normally indicate where in the source this change is made.
The rest is only my personal interpretation of how it could be written to make most sense in a short form.

src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
```diff
@@ -1234,7 +1234,45 @@ class npc_scarlet_miner : public CreatureScript
...
+        PrepareAuraScript(spell_stable_master_repo_AuraScript);
+
+        bool Validate(SpellInfo const* /*spellInfo*/) override
+        {
+            return ValidateSpellInfo({ SPELL_REPO });
+        }
+
+        void HandleDummy(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+        {
+            Creature* target = GetTarget()->ToCreature();
```
Riztazz: If in any way this spell will be casted on player, worldserver will crash due to this part :P

Riztazz: If in any way this spell will be casted on player, worldserver will crash due to this part :P

ccrs:
```
Attributes: 0x21000000 (SPELL_ATTR0_CASTABLE_WHILE_MOUNTED, SPELL_ATTR0_UNAFFECTED_BY_INVULNERABILITY)
AttributesEx2: 0x00000004 (SPELL_ATTR2_CAN_TARGET_NOT_IN_LOS)
AttributesEx3: 0x10000000 (SPELL_ATTR3_UNK28)
AttributesEx5: 0x00000008 (SPELL_ATTR5_USABLE_WHILE_STUNNED)
AttributesEx6: 0x00001000 (SPELL_ATTR6_CASTABLE_WHILE_ON_VEHICLE)
````
agreed, this spell needs TYPEID validation at least

Comment (Wyreth) [reply src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp]: Sorry, forgot to mention - conditions are applied in #20767.

Still better to avoid any chance of crash :P yep.

File: src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp [hunk 2]
```diff
@@ -1234,7 +1234,48 @@ class npc_scarlet_miner : public CreatureScript
...
+public:
+    spell_stable_master_repo() : SpellScriptLoader("spell_stable_master_repo") { }
+
+    class spell_stable_master_repo_AuraScript : public AuraScript
+    {
+        PrepareAuraScript(spell_stable_master_repo_AuraScript);
+
+        bool Validate(SpellInfo const* /*spellInfo*/) override
+        {
+            return ValidateSpellInfo({ SPELL_REPO });
```
Mihapro: You don't need this. I believe it's only for spells that you cast in functions below.

src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
Riztazz: Ah makes more sense now :P

src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
ccrs: the id which the spellscript is binded into is validated on loading

Wyreth: Nice, thanks! Copied that from another script, thought it was needed.

Mihapro:
Who should be targeted by this spell? Mount/creature? Player?

Wyreth:
@Mihapro the spell only targets the mount.

Mihapro:
ok, then you check if it's a creature instead of unit:
```C++
Creature* creature = GetTarget()->ToCreature();
if (!creature)
    return;

//and then use creature instead of GetTarget() to make it easier to read for others :P
```
Not sure if spell needs conditions or something, how does it find the mount?
SpellInfo says: Targets (25, 0) (TARGET_UNIT_TARGET_ANY, NO_TARGET)

Wyreth:
@Mihapro you're right, that's a better way to handle it.

Conditions are already implemented in 1cd554e (should have put them here after all, but what is done is done :P).

src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
```diff
@@ -1234,7 +1234,38 @@ class npc_scarlet_miner : public CreatureScript
...
+        PrepareAuraScript(spell_stable_master_repo_AuraScript);
+
+        void HandleDummy(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+        {
+            Creature* creature = GetTarget()->ToCreature();
+            if (!creature)
+                return;
+
+            if (Unit* owner = creature->GetCharmerOrOwner())
+                GetCaster()->GetAI()->AttackStart(owner);
```
Shauren: attack charmer? seems suspicious - pretty sure this will not work at this point since its friendly during the charm

offl: On which source the script is based?
Maybe this should be done here?
https://github.com/TrinityCore/TrinityCore/pull/20897/files#diff-baeb010bab3defc8b23a63d10965f7e8R829

Sorikoff:
Any news on this?

PR #20777, opened 2017-11-01
Core/SAI: don't call OnReset right after a creature using SAI is summoned, and fix WaypointReached usage outside of escorts

### LINKED ISSUE #20771: [3.3.5] Territorial Trespass Quest 13051
**Description:**
Veranus flies to the nest but does not land.
Thorim does not appear.
You can not finish the quest.

**Expected behaviour:**
http://www.wowhead.com/quest=13051/territorial-trespass
https://www.youtube.com/watch?v=Ucdzb4oG5-0

**Branch(es):**

 3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. aa4bf288e77f 2017-11-01 03:44:59 +0100 (335a branch)

**TDB version:**  TDB_full_world_335.63_2017_04_18 + updates

**Operating system:**Win7

Wyreth:
**Changes proposed:**

The waypoint change is related to #20684.

About the OnReset change: when JustAppeared() is called by a summoned creature, it also resets invoker and on-going scripts that have been called via SMART_EVENT_JUST_SUMMONED and similar.

This change makes it so spawned creatures don't reset at their fist update tick (they're new creatures, they've already reset - no need to do so again).

Also cleaned up JustAppeared and JustReachedHome a bit, and removed a now useless bool.

**Target branch(es):**

- [x] 3.3.5
- [x] master

**Issues addressed:** closes #20771, updates #20310.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -31,8 +31,8 @@
 
 SmartAI::SmartAI(Creature* creature) : CreatureAI(creature), mIsCharmed(false), mFollowCreditType(0), mFollowArrivedTimer(0), mFollowCredit(0), mFollowArrivedEntry(0), mFollowDist(0.f), mFollowAngle(0.f),
     _escortState(SMART_ESCORT_NONE), _escortNPCFlags(0), _escortInvokerCheckTimer(1000), _currentWaypointNode(0), _waypointReached(false), _waypointPauseTimer(0), _waypointPauseForced(false), _repeatWaypointPath(false),
-    _OOCReached(false), _waypointPathEnded(false), mRun(true), mEvadeDisabled(false), mCanAutoAttack(true), mCanCombatMove(true), mInvincibilityHpLevel(0), mDespawnTime(0), mDespawnState(0), mJustReset(false),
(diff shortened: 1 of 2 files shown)


src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -541,8 +541,8 @@ void SmartAI::JustAppeared()
     me->SetVisible(true);
     if (me->GetFaction() != me->GetCreatureTemplate()->faction)
         me->RestoreFaction();
-    mJustReset = true;
-    JustReachedHome();
+    if (!me->IsSummon())
+        GetScript()->OnReset();
```
joschiwald: i think this one need a generic fix and not only sai summons

tkrokli:
Cosmetic (I know), but still: Google "github commit title length"
(https://www.google.com/search?&q=github+commit+title+length)
Consider adding the explanatory description in the comment message body, not in the title line.
(it is comparable to writing an email message entirely in the Subject: field)

lost-illusion:
Hi @Wyreth, in additional we needn't resets all variables in SmartAI::JustAppeared, because we everytime have new creature and we can change some variables in smart events, like respawn and others.
Will be good if you can review this case.

PR #20745, opened 2017-10-29
Scripts/EasternKingdoms: implement Children of Goldshire event

Kittnz:
http://wowwiki.wikia.com/wiki/Children_of_Goldshire

**Changes proposed:**

-  Implement Childeren of Goldshire Event

**Target branch(es):** 
335 for now

**Issues addressed:** Closes #  (insert issue tracker number)
#19090

**Known issues and TODO list:** (add/remove lines as needed)

- [x] Formation of the childeren does not work with MovePath function for unknown reason. @Shauren any chance you could look into this?
- [x] Add correct waypoints for 3.3.5
- [x] Set correct timings for waypoints
- [x] Remove SetActive, here for testing

PR #20779, opened 2017-11-02
Core/Movement: Add MovePoint with facing

Kittnz:
Thanks to Riztazz

**Changes proposed:**

-  Core/MotionMaster: Add MovePoint with facing

**Target branch(es):** 3.3.5

- [x] 3.3.5

Works as is.

Notes: This is needed for Children of Goldshire event #20745


src/server/game/Movement/MotionMaster.h
```diff
@@ -129,6 +129,7 @@ class TC_GAME_API MotionMaster
             MovePoint(id, pos.m_positionX, pos.m_positionY, pos.m_positionZ, generatePath);
         }
         void MovePoint(uint32 id, float x, float y, float z, bool generatePath = true);
+        void MovePoint(uint32 id, Position const& pos, bool generatePath, bool _setFinalFacing);
```
Keader: setFinalFacing

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -339,6 +339,21 @@ void MotionMaster::MovePoint(uint32 id, float x, float y, float z, bool generate
     }
 }
 
+void MotionMaster::MovePoint(uint32 id, Position const& pos, bool generatePath = true, bool _setFinalFacing = false)
```
Keader: setFinalFacing

tkrokli:
About the title edit: I simply picked the directory name 1 level up
(which also is the common denominator for all files in this PR).

Keader (approved):
Looks good to me

ccrs:
doesnt feel right.. everything should be just handled with Position and update all the x,y,z uses..
stall this for the weekend, maybe I can perform the changes

Riztazz:
Afaik shauren didnt want to use NAN to determine if there is orientation set, because by default orient is set to some value, 0.0f i think? Even tho 0.0f is still valid orient ;p

Keader:
Maybe has some useful discussion about it [here](https://github.com/TrinityCore/TrinityCore/pull/18893)

Shauren:
#include "Optional.h"

Use that instead of float+bool arg pair (dont shove it into Position)

jackpoz:
@Kittnz can you use the suggestion above about using Optional ? Then this PR could be merged

Kittnz:
Ill look into it.

Naios (approved):

Naios (changes requested):
As said before, the optional part is missing

src/server/game/Movement/MovementGenerators/PointMovementGenerator.cpp
```diff
@@ -50,6 +50,10 @@ void PointMovementGenerator<T>::DoInitialize(T* owner)
     init.MoveTo(_x, _y, _z , _generatePath);
     if (_speed > 0.0f)
         init.SetVelocity(_speed);
+
+    if (_orient)
+        init.SetFacing(*_orient);
```
jackpoz: are these 2 lines correct ?

jackpoz:
@Naios could you check if it's ok now ?

Shauren (changes requested):
No, Optional was meant to propagate in api back to MotionMaster::MovePoint, with the overload that has Position argument not being added

src/server/game/Movement/MovementGenerators/PointMovementGenerator.cpp
Naios: Yes, this should be correct

Naios (approved):
Optional used correctly instead of 0 float,
Please address Shaurens comment before merging please

jackpoz:
@Shauren better now ? (I actually deleted the overload before reading your comment since it didn't make sense to add a new one to me either)

Shauren:
`Optional<float> = nullptr` only works because its we defined `BOOST_OPTIONAL_USE_OLD_DEFINITION_OF_NONE` to work around msvc linker errors, you should default construct it with `Optional<float> = {}`

Shauren (approved):

PR #20785, opened 2017-11-04
Core/Gameobject: Add WorldEffectID to Gameobject

funjoker:
Example:
Purple Frame around treasures

wowscrnshot_110417_042412

wowscrnshot_110417_043315

Left without WorldEffectId | Right with WorldEffectId

**Changes proposed:**

-  Adding WorldEffectID

- [ ] 3.3.5
- [x] master

- [x] No Idea


funjoker:
Unsure if there is the need for gameobject_addon
Or if
gameobject_template_addon is enough.

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -1228,6 +1228,7 @@ void ObjectMgr::LoadGameObjectAddons()
         gameObjectAddon.ParentRotation = QuaternionData(fields[1].GetFloat(), fields[2].GetFloat(), fields[3].GetFloat(), fields[4].GetFloat());
         gameObjectAddon.invisibilityType = InvisibilityType(fields[5].GetUInt8());
         gameObjectAddon.InvisibilityValue = fields[6].GetUInt32();
+        gameObjectAddon.worldEffectID = fields[7].GetUInt32();
```
joschiwald: is there a way to validate this id?

funjoker: Maybe WorldEffect.db2
Need to investigate it.

PR #20788, opened 2017-11-04
Core/Movement: Rewrite confused movement

Mihc3:
Based on `15595_2012-08-22_22-02-32_looking_for_dungeon_delete_item_set.pkt`.

I gathered waypoints and made a nice graph/map of a few confusion effects:
https://www.desmos.com/calculator/cipr26agwb

**Changes proposed:**
- Unit will now make 5 moves and not 2 (during confusion that lasts 4 seconds).
    - New waypoint calculated right after 800ms, no wait time.
- Distance from original point seems to be more or less rounded value.
- Less unnecessary IF statements and indentation.

- 3.3.5 (probably?)
- master ✔️

* Build: ✔️
* Tests:
    * Confusion (41397) cast on enemy creatures nearby ✔️
    * Confusion (170349) cast on self ✔️

* None.

### DIFF
--- a/src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
@@ -56,44 +56,38 @@ bool ConfusedMovementGenerator<T>::DoUpdate(T* unit, uint32 diff)
     if (unit->HasUnitState(UNIT_STATE_ROOT | UNIT_STATE_STUNNED | UNIT_STATE_DISTRACTED))
         return true;
 
-    if (i_nextMoveTime.Passed())
-    {
-        // currently moving, update location
-        unit->AddUnitState(UNIT_STATE_CONFUSED_MOVE);
+    // Check if we must generate movement.
+    i_nextMoveTime.Update(diff);
+    if (!i_nextMoveTime.Passed())
+        return true;
+    
+    // Get original position (where we started)
+    Position pos;
+    pos.Relocate(i_x, i_y, i_z);
 
-        if (unit->movespline->Finalized())
-            i_nextMoveTime.Reset(urand(800, 1500));
-    }
-    else
-    {
-        // waiting for next move
-        i_nextMoveTime.Update(diff);
-        if (i_nextMoveTime.Passed())
-        {
-            // start moving
-            unit->AddUnitState(UNIT_STATE_CONFUSED_MOVE);
+    // Generate vector
+    float dist = float(urand(1, 4));
+    float angle = 2 * float(M_PI) * float(rand_norm());
+    
+    // Check how far we can move (collision check).
+    unit->MovePositionToFirstCollision(pos, dist, angle);
+    
+    // Generate path
+    PathGenerator path(unit);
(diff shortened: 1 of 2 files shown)


src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
```diff
@@ -56,44 +56,38 @@ bool ConfusedMovementGenerator<T>::DoUpdate(T* unit, uint32 diff)
     if (unit->HasUnitState(UNIT_STATE_ROOT | UNIT_STATE_STUNNED | UNIT_STATE_DISTRACTED))
         return true;
 
-    if (i_nextMoveTime.Passed())
-    {
-        // currently moving, update location
-        unit->AddUnitState(UNIT_STATE_CONFUSED_MOVE);
```
joschiwald: what happend with this unit state?

Mihapro: Isn't adding it on init and reset enough? Or does something remove it?

Mihapro: It doesn't seem anything in core would clear all unit states or specifically confused move - only confused movement generator itself removes it

src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
```diff
@@ -31,73 +31,76 @@ void ConfusedMovementGenerator<T>::DoInitialize(T* unit)
...
+
+    // Check how far we can move (collision check).
+    unit->MovePositionToFirstCollision(pos, dist, angle);
+
+    // Generate path
+    PathGenerator path(unit);
+    path.SetPathLengthLimit(10.0f);
+    bool result = path.CalculatePath(pos.m_positionX, pos.m_positionY, pos.m_positionZ);
+    if (!result || (path.GetPathType() & PATHFIND_NOPATH))
+        return true; // Return without setting timer, generate another path on next update.
```
tkrokli:
```cpp
src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp:88:9:
 fatal error: void function '_move' should not return a value [-Wreturn-type]

        return true; // Return without setting timer, generate another path on next update.
        ^      ~~~~
```
(`_move` in Line 70 - https://github.com/TrinityCore/TrinityCore/pull/20788/files#diff-6924fe8c6e57209e1beffb3de0491242R70)

jackpoz:
this PR has a lot of changes about "Less unnecessary IF statements and indentation." that makes it harder to review. could you just include the bare minimum changes required ?

Mihapro:
I only changed from ```if/else``` to ```if(not)->return``` so it doesn't need to check timer twice in one update (timer.passed() thing), I didn't touch most of the code 😄

ccrs:
this doesnt apply to 335, its already implemented and most likely improved

jackpoz:
So master just needs some cherrypicks ?

MaxtorCoder:
> 
> 
> So master just needs some cherrypicks ?

Yes, definitely :P

jackpoz:
Closing this PR then

PR #20793, opened 2017-11-06
Scripts/EasternKingdoms: Ragged John

mysterious-static:
**Changes proposed:**

- Migrates Ragged John to SAI
- Removes zone_burning_steppes.cpp

**Target branch(es):** 3.3.5

**Tests performed:**
Builds. Verified SQL script was syntactically correct.

- [x] The included SAI is a direct port from C++. I still need to verify that the event that I noted in SQL comments will actually fire or whether that also needs to be scripted. Will be doing this tonight.


src/server/scripts/EasternKingdoms/eastern_kingdoms_script_loader.cpp
```diff
@@ -175,7 +175,7 @@ void AddSC_instance_zulgurub();
 //void AddSC_alterac_mountains();
 void AddSC_arathi_highlands();
 void AddSC_blasted_lands();
-void AddSC_burning_steppes();
+void AddSC_duskwood();
```
Aokromes: duplicate row.

velinath: How in the world did I manage that?

sql/updates/world/3.3.5/9999_99_99_99_world.sql
```diff
@@ -0,0 +1,20 @@
...
+DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId` = 22 AND `SourceGroup` = 1 AND `SourceEntry` = 9563 AND `SourceId` = 0;
+INSERT INTO `conditions` (`SourceTypeOrReferenceId`, `SourceGroup`, `SourceEntry`, `SourceId`, `ElseGroup`, `ConditionTypeOrReference`, `ConditionTarget`, `ConditionValue1`, `ConditionValue2`, `ConditionValue3`, `NegativeCondition`, `ErrorType`, `ErrorTextId`, `ScriptName`, `Comment`) VALUES
+(22, 1, 9563, 0, 0, 1, 0, 16468, 1, 0, 0, 0, 0, '', 'SAI event 0 for Ragged John will execute if target has aura Mother\'s Milk.'),
+(22, 1, 9563, 0, 0, 32, 0, 16, 0, 0, 0, 0, 0, '', 'SAI event 0 for Ragged John will execute if target is player.');
+
+-- Missing events?
+
+-- per wowhead - player dialog w/NPC "Milk me, John." "With pleasure... just let me savor this moment for a minute longer." "Do it... do it now."
+-- verifiable w/ broadcast_text id 5833, 5834, 5835
+-- no video available that I can find, unfortunately, so I have no idea how this relates at all to the actual event action above, or whether that may be handled by the spell (I'll check on this before PR is opened.)
```
offl:
```sql
SELECT * FROM `broadcast_text` WHERE `ID` BETWEEN 5833 AND 5835;
```
```
    ID  Language  MaleText                                                             FemaleText           
------  --------  -------------------------------------------------------------------  ---------------------
  5833         0  Milk me, John.                                                       Milk me, John.       
  5834         0  With pleasure... just let me savor this moment for a minute longer.                       
  5835         0  Do it... Do it now.                                                  Do it... Do it now.  
```

Quite possible he should cast the spell only after player select the gossip.

velinath: Could be gossip, in which case this whole thing becomes a lot easier. I figured based on that Wowhead comment it was scripted /say text - just based on the word "dialogue" - but I was hoping someone would have a video somewhere - without other sources, I usually default to doing a direct conversion from the C++ script.

I'm fine with moving it to gossip if devs are also comfortable with that - maybe we can get a check in from them on that :)

offl: One text already present in npc_text, so..
```sql
SELECT * FROM `npc_text` WHERE `ID`=3496;
```
```text
    ID  text0_0                                                              
------  ---------------------------------------------------------------------
  3496  With pleasure... just let me savor this moment for a minute longer.  
```

velinath: good catch, that's enough for me to be going on with! I'll make necessary changes sometime today or tomorrow.

velinath:
actually this is fine now, i was running into a phantom problem because i had leftover conditions for smart scripts to fire

this is ready to go, tested ingame

Issue #20591, opened 2017-10-10
Quest: A Little Slime Goes a Long Way


STARRHELD:
**Description:**

**Current behaviour:**

The provided quest items "Empty Tainted Ooze Jar" and "Empty Cursed Ooze Jar" can be used on alive Oozes as many times you want. You can simply finish this quest on a single Ooze without even killing it.

**Expected behaviour:**

To use the provided quest items you are supposed to kill the Ooze first. The Ooze must be dead so you can use the quest item. After you use it, the Ooze is supposed to disappear so can't use it on the Ooze again!

**Steps to reproduce the problem:**

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 1af6df4

CHANGEME Copy the first line of the `worldserver`, `authserver` or `bnetserver` startup.
For example: TrinityCore rev. 0000000000 2000-01-09 11:31:41 +0100 (my branch) (Win64, RelWithDebInfo, Static) (bnetserver)

**TDB version:**  3.3.5.62

**Operating system:** Linux

STARRHELD:
> We need conditions to ensure that the correct spell is used on the specific creature.
```SQL
DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId`=17 AND `SourceEntry`=15698;
INSERT INTO `conditions` (`SourceTypeOrReferenceId`, `SourceGroup`, `SourceEntry`, `SourceId`, `ElseGroup`, `ConditionTypeOrReference`, `ConditionTarget`, `ConditionValue1`, `ConditionValue2`, `ConditionValue3`, `NegativeCondition`, `ErrorType`, `ErrorTextId`, `ScriptName`, `Comment`) VALUES 
(17, 0, 15698, 0, 0, 31, 1, 3, 7086, 0, 0, 0, 0, '', 'Filling Empty Jar can be used on Cursed Ooze'),
(17, 0, 15698, 0, 0, 36, 1, 0, 0, 0, 1, 0, 0, '', 'Target must be dead');

DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId`=17 AND `SourceEntry`=15699;
[...]

PR #20795, opened 2017-11-06
Core/AI: implement AI function DelayedCast to delay a spell cast

gerripeach:

- [X] 3.3.5
- [?] master - has to be ported & maybe fix compile due to slightly different structure

does build
tested with several creatues

### DIFF
--- a/src/server/game/AI/CoreAI/UnitAI.cpp
+++ b/src/server/game/AI/CoreAI/UnitAI.cpp
@@ -193,6 +193,33 @@ void UnitAI::DoCastAOE(uint32 spellId, bool triggered)
     me->CastSpell(nullptr, spellId, triggered);
 }
 
+BasicEvent* UnitAI::DoCastDelayed(Unit* victim, uint32 spellId, uint32 delay, bool triggered, bool aliveCheck)
+{
+    struct DelayedCast : public BasicEvent
+    {
+        DelayedCast(Unit *caster, Unit *target, uint32 spellId, bool triggered, bool aliveCheck) : BasicEvent(),
+            _caster(caster), _target(target), _spellId(spellId), _triggered(triggered), _aliveCheck(aliveCheck) { }
+
+        bool Execute(uint64, uint32)
+        {
+            if ((_aliveCheck && _caster->IsAlive()) || !_aliveCheck)
+                _caster->CastSpell(_target, _spellId, _triggered);
+            return true;
+        }
+
+        Unit *_caster;
+        Unit *_target;
+        uint32 _spellId;
+        bool _triggered;
+        bool _aliveCheck;
+    };
+
+    BasicEvent* castEvent = new DelayedCast(me, victim, spellId, triggered, aliveCheck);
+    me->m_Events.AddEvent(castEvent, me->m_Events.CalculateTime(delay));
+    return castEvent;
+}
+
+
 uint32 UnitAI::GetDialogStatus(Player* /*player*/)
 {
     return DIALOG_STATUS_SCRIPTED_NO_STATUS;
--- a/src/server/game/AI/CoreAI/UnitAI.h
+++ b/src/server/game/AI/CoreAI/UnitAI.h
@@ -39,6 +39,7 @@ class Player;
 class Quest;
 class SpellInfo;
 class Unit;
+class BasicEvent;
 struct AISpellInfoType;
 enum DamageEffectType : uint8;
 enum SpellEffIndex : uint8;
@@ -298,6 +299,7 @@ class TC_GAME_API UnitAI
         void DoCastSelf(uint32 spellId, bool triggered = false) { DoCast(me, spellId, triggered); }
         void DoCastVictim(uint32 spellId, bool triggered = false);
         void DoCastAOE(uint32 spellId, bool triggered = false);
+        BasicEvent* DoCastDelayed(Unit* victim, uint32 spellId, uint32 delay, bool triggered = false, bool aliveCheck = true);
 
         float DoGetSpellMaxRange(uint32 spellId, bool positive = false);
 


src/server/game/AI/CoreAI/UnitAI.cpp
```diff
@@ -193,6 +193,35 @@ void UnitAI::DoCastAOE(uint32 spellId, bool triggered)
...
+    {
+        DelayedCast(Unit *caster, Unit *target, uint32 spellId, bool triggered, bool aliveCheck) : BasicEvent(),
+            _caster(caster), _target(target), _spellId(spellId), _triggered(triggered), _aliveCheck(aliveCheck) { }
+
+        bool Execute(uint64, uint32)
+        {
+            if (_aliveCheck && _caster->IsAlive())
+                _caster->CastSpell(_target, _spellId, _triggered);
+            else
+                _caster->CastSpell(_target, _spellId, _triggered);
```
joschiwald: nice copy paste from somewhere

gerripeach: the idea from DoCastDelayed is not from me. Tell me the real author and I change the commit message

Shauren: Thats not what he is talking about - look closely at the code. If branch is exactly the same as else branch

gerripeach: I don't see the problem. My trinity core branch is 3(?) years behind. So there might be some issues with different code

src/server/game/AI/CoreAI/UnitAI.cpp
```diff
@@ -193,6 +193,35 @@ void UnitAI::DoCastAOE(uint32 spellId, bool triggered)
...
+            return true;
+        }
+
+        Unit *_caster;
+        Unit *_target;
```
joschiwald: this leads to crashes
btw. since we have a taskscheduler i don't see any use for this function

src/server/game/AI/CoreAI/UnitAI.cpp
Seyden: Just look closely dude.. your condition doesn't make any sense

Shauren: You don't understand, I'm not talking about git branches but this https://en.wikipedia.org/wiki/Branch_(computer_science)

Shauren (changes requested):
fix the if

Aokromes:
https://github.com/TrinityCore/TrinityCore/pull/20795/files#diff-ab7e4a0d883137967b73ba1fd4e3c241R205 he changed the if days ago.

ccrs:
it still makes no sense

Keader:
any example where it can be used?

gerripeach:
If you have any dead creature and this creatures has to cast after a specific time a spell nothing that uses the update can be used

src/server/game/AI/CoreAI/UnitAI.cpp
gerripeach: ty: shauren

Krudor:
I've encountered a similiar issue with this before, where Firelands bosses cast the legendary aura that allows players to drain their essence, a few seconds after their death.

I would either fix the underlying issue with after-death events (UpdateAI not triggering right?), or work around it with a BasicEvent like you're doing here, just don't make it generic.

This is more like a generic hackfix to circumvent the real issue.

Seyden:
Bro.. this if clause is always shitty.. don't try to bypass your problem with an else if..

try it with only ONE if condition.. you can use multiple condition in 1 if clause

Treeston:
> for exaple have a look at chogall. this boss summons adds. when these adds die they will casted 4 sec after death 2 spells. *I could either call a method on chogall and let him do the update* or use my method

Yeah, so what's the point of this?

PS: Are you even sure the creatures are fully dead in your example? Sounds like something Blizzard would use feign death for, dead creatures practically never cast spells.

Krudor:
> PS: Are you even sure the creatures are fully dead in your example? Sounds like something Blizzard would use feign death for, dead creatures practically never cast spells.

I'm going to assume you did not read my reply, and this I'll repeat it again down below.

I know for a fact Firelands bosses cast spells after they have died, to give you an example.

We need SOMETHING to support this, but as I mentioned, THIS implementation feels hacky and more like a workaround.

Treeston:
Calling `UpdateAI` for dead creatures is a lot of extra jumps for very few special cases, I'm not a big fan.

Is there some issue with using `Event`s?

src/server/game/AI/CoreAI/UnitAI.cpp
Krudor: Because TaskScheduler wouldn't work if its not being updated from UpdateAI?

Treeston: `m_Events` still processes while dead. Why is this not sufficient?

Krudor: Sure this will suffice, but doesn't it feel hacky?
This would only allow us to cast spells with a certain delay, that aren't affected by death for example.

How about when we need to do more than just cast a delayed spell upon death.
I don't have a suggestion right now, but I don't see why we shouldn't be able to schedule after-death events properly, even though in 99% of the cases it's not relevant.

Treeston: You seem to be misunderstanding - I'm not agreeing with the approach of the PR.

I'm trying to figure out why (or whether) you think the current state of affairs - using custom `Event`s for the very few edge cases where this is relevant and no other master AI is around to handle it cleanly - is insufficient.

Krudor: I think using custom Events for special cases like this is fine.

Killyana:
In some cases like https://github.com/TrinityCore/TrinityCore/issues/20591 we need to despawn the dead npc on spell hit, r00ty-tc already did an initial work https://github.com/r00ty-tc/TrinityCore/commit/74f640f4c348415686c8ff7ef8c6992e3b789994

src/server/game/AI/CoreAI/UnitAI.cpp
jackpoz: who won here ? is it still useful to have this event ?

Treeston: Doesn't seem like there's any convincing argument for it. Closing.

PR #20828, opened 2017-11-07
Core/AI: only call JustAppeared for entities that are using the old respawn system

Wyreth:
**Changes proposed:**

This hopefully fixes everything that's wrong with ~~the world~~ AI not behaving like it should.

JustAppeared was meant only for re-used entities (old respawn system). So allow it to be called only by those entities.

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:** updates #20310.


Treeston:
> JustAppeared was meant only for re-used entities (old respawn system).

Huh, what? Who says that?

PR #19288, opened 2017-03-13
Scripts/Commands: Modify fury & pain

Hadifaren:
**Changes proposed:**

- Modify some files to edit Fury & Pain
- class Demon Hunter

**Target branch(es):** 3.3.5/master
- [X] master

**Tests performed:** (Does it build, tested in-game, etc.)
- [X] build successful
- [X] tested in game

**Known issues and TODO list:** (add/remove lines as needed)

PR #20833, opened 2017-11-09
Scripts/Commands: command .modify power (all powers)

Hadifaren:
`.modify power` `#powercount`

Closes #19288

- [x] master

Tested And Works


tkrokli:
```
- "sql/updates/auth/2017_11_16_20833_auth.sql" is missing in the 'auth'.'updates' table.
error:
  The Database Updater is broken for the 'auth' database,
  because the applied update is missing in the 'auth'.'updates' table.
How to fix:
  Insert the missing names of the already applied sql updates
  to the 'updates' table of the 'auth' base dump ('sql/base/auth_database.sql').
```

Hadifaren:
i dont know how to solve the error you writed ^_^
+ cannot understand how to fix

Hadifaren:
:| why its closed ! @tkrokli

src/server/scripts/Commands/cs_modify.cpp
```diff
@@ -997,6 +998,66 @@ class modify_commandscript : public CommandScript
...
     }
+    	// Edit Player Power 
+	static bool HandleModifyPowerCommand(ChatHandler* handler, const char* args)
+	{
+		int32 power, maxpower;
+		Player* target = handler->getSelectedPlayerOrSelf();
+		int8 const powerMultiplier = 10;
+		if (CheckModifyResources(handler, args, target, power, maxpower, powerMultiplier))
+		{
+			switch (target->getPowerType())
```
tkrokli:
```
src/server/scripts/Commands/cs_modify.cpp:1009:12:
 fatal error:
 13 enumeration values not handled in switch: 'POWER_COMBO_POINTS', 'POWER_RUNES', 'POWER_SOUL_SHARDS'... [-Wswitch]
                        switch (target->getPowerType())
                                ^
1 error generated.
```
https://github.com/TrinityCore/TrinityCore/blob/2d75d8071bb6d85987bdc5d88ed2271822fcbcfd/src/server/game/Miscellaneous/SharedDefines.h#L250-L274

There are 22 `Powers` listed in the enum in `SharedDefines.h` which means you need to find a way to tell this script to skip/ignore the 13 powers not handled by this switch. I don't know what that code should look like, but maybe some of the other developers will know.

Jildor: Need add a default in switch and add tabs in  cases

tkrokli: OK, I get what you mean that it needs a
```cpp
			default:
				break;
```
at the end, but not sure what you mean by `add tabs in cases`

Jildor:
Add tabulations in  all cases:
```diff
-    case X
+        case X
```

Sorry mobile :smile:

src/server/scripts/Commands/cs_modify.cpp
Jildor: Add tabulations in  all cases:
```diff
-    case X
+        case X
```

Sorry mobile :smile:

tkrokli: OK, I see. Add correct indent to each case,

tkrokli: BTW; changes have been made, but they are all below this comment section
(and hence they don't get hidden by the update).

tkrokli:
OK - so far, so good. The C++ part compiles without errors.
Only the `sql/base/auth_database.sql` part left to fix to have travis-ci show up green.

tkrokli:
@Hadifaren : I hope you take notes on my changes so you can use that knowledge later. 😃

Hadifaren:
@tkrokli of course your comments are so helpful ^_^

Hadifaren:
of course your comments always are helpful ^_^

Seyden:
A player can have multiple powers.. what you do then? What if i want to change the power that i got after i used a spell? With your change, it only uses my current power.

Also the power multiplier isn't always the same for each power type.

tkrokli:
@Seyden : Looks like your question is answered in the closing commit (reading the added content & commit message). 😄

Issue #13467, opened 2014-10-27
[Dev] WIP Catmullrom waypoint movement


Pitcrawler:
I started to play around a little with the core and made the following change to WaypointMovementGenerator.cpp

``` diff
diff --git a/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp b/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
index 942996b..5b8dd8b 100755
--- a/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
@@ -153,6 +153,16 @@ bool WaypointMovementGenerator<Creature>::StartMove(Creature* creature)
             trans->CalculatePassengerPosition(formationDest.x, formationDest.y, formationDest.z, &formationDest.orientation);
     }

+   if (creature->IsFlying())
+   {
+       for (uint32 i = 0; i < i_path->size() - i_currentNode; i++)
+       {
+           init.Path().push_back(G3D::Vector3(i_path->at(i_currentNode + i)->x, i_path->at(i_currentNode + i)->y, i_path->at(i_currentNode + i)->z));
+       }
+       init.SetSmooth();
+       init.Launch();
+       return true;
+   }
     //! Do not use formationDest here, MoveTo requires transport offsets due to DisableTransportPathTransformations() call
     //! but formationDest contains global coordinates
     init.MoveTo(node->x, node->y, node->z);
```

This made all the flying NPCs that have waypoint data use catmullrom movement. But there are some things that certainly need improvements which I can't take care of because I have really little knowledge when it gets to core programming:
At each WP the remaining path is rebuild which is unnecessary (memory consumption) and not blizzlike.
At the last WP there is no next WP so no path is build and point movement is used instead of catmullrom.
Flying creatures that have inhabitType!=4 (i.e. 5 or 7) don't use catmullrom movement. I guess the core doesn't handle them correctly so they don't get their movementflags.
Swimming creatures don't use catmullrom movement.
##

jackpoz:
any update on this ?

Kittnz:
No one is working on it, hope someone picks it up.

click:
Ok, let's continue on this one ...
**Single pathing**
Example with 20 "real" waypoints:
1-2-3...18-19-20-**21** - 21 is the same as the "final" waypoint (20) - describes that we're on the endpoint.

[...]

PR #20845, opened 2017-11-12
Core/Waypoints: Attempt to fix "smooth" waypoints

gerripeach:
WIP

this will send the next 3 wp verticies to the client.

- [X] 3.3.5
- [ ] master

#13467

- [ ] check if formations still work as intended

### DIFF
--- a/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
@@ -197,7 +197,26 @@ bool WaypointMovementGenerator<Creature>::StartMove(Creature* creature)
 
     //! Do not use formationDest here, MoveTo requires transport offsets due to DisableTransportPathTransformations() call
     //! but formationDest contains global coordinates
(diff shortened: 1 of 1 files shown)


Ovahlord:
formations are broken. the troup is following only if a point has been reached. I think formations need a constant update rather than just on point reached

Ovahlord:
Also another small issue: 
if the creature's next wayoint makes her turn arround by about 180 degrees for example the creature skips a bit of her path. you might check that out to make those waypoints not affected by the smoothing

Krudor:
I think thats more a matter of the path reaching its destination and resetting.
Paths should generally end where the creature needs to do a 180 turn.

lachtanek:
Maybe I'm wrong, I don't really understand all the waypoint code, but does MovementInform still work for the points that were sent together?
Also, the first condition if the for loop could be part of the loop's condition

gerripeach:
will update it later

ccrs:
unfinished, but the concept is correct

gerripeach:
haveing problems with movement inform.. fixes leading to crashes..

PR #20851, opened 2017-11-13
Core/Auras: implement Aura 379 (SPELL_AURA_MOD_MANA_REGEN_PCT)

DoctorKraft:
**Changes proposed:**

- Update mana regen with amount based on GT
- Remove old system with calculation based on spirit
- Implent Aura 379 (SPELL_AURA_MOD_MANA_REGEN_PCT)
Found on each passive spell of class/spec

- [X] master

### DIFF
--- a/src/server/game/Entities/Unit/StatSystem.cpp
+++ b/src/server/game/Entities/Unit/StatSystem.cpp
@@ -757,10 +757,7 @@ void Player::UpdateManaRegen()
     // Apply PCT bonus from SPELL_AURA_MOD_MANA_REGEN_PCT
     base_regen *= GetTotalAuraMultiplierByMiscValue(SPELL_AURA_MOD_MANA_REGEN_PCT, POWER_MANA);
 
-    // Apply PCT bonus from SPELL_AURA_MOD_MANA_REGEN_INTERRUPT
-    int32 modManaRegenInterrupt = GetTotalAuraModifier(SPELL_AURA_MOD_MANA_REGEN_INTERRUPT);
-
-    SetStatFloatValue(UNIT_FIELD_POWER_REGEN_INTERRUPTED_FLAT_MODIFIER + manaIndex, base_regen * modManaRegenInterrupt);
+    SetStatFloatValue(UNIT_FIELD_POWER_REGEN_INTERRUPTED_FLAT_MODIFIER + manaIndex, base_regen);
     SetStatFloatValue(UNIT_FIELD_POWER_REGEN_FLAT_MODIFIER + manaIndex, base_regen);
 }
 
--- a/src/server/game/Entities/Unit/StatSystem.cpp
+++ b/src/server/game/Entities/Unit/StatSystem.cpp
@@ -23,6 +23,7 @@
 #include "Pet.h"
 #include "Creature.h"
 #include "GameTables.h"
+#include "ObjectMgr.h"
 #include "SharedDefines.h"
 #include "SpellAuras.h"
 #include "SpellAuraEffects.h"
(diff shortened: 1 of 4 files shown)


Hadifaren:
Doctor Do you know how to implement aura AURA_MOD_MAX_POWER 418

Naios:
Is there anything missing for this PR? Otherwise it could be merged I think.

DoctorKraft:
@Naios All are good for me, works well

src/server/game/Entities/Unit/StatSystem.cpp
```diff
@@ -743,19 +744,24 @@ void Player::UpdateManaRegen()
...
+
+    // Apply PCT bonus from SPELL_AURA_MOD_MANA_REGEN_PCT
+    base_regen *= GetTotalAuraMultiplierByMiscValue(SPELL_AURA_MOD_MANA_REGEN_PCT, POWER_MANA);
+
+    // Apply PCT bonus from SPELL_AURA_MOD_MANA_REGEN_INTERRUPT
     int32 modManaRegenInterrupt = GetTotalAuraModifier(SPELL_AURA_MOD_MANA_REGEN_INTERRUPT);
 
-    SetStatFloatValue(UNIT_FIELD_POWER_REGEN_INTERRUPTED_FLAT_MODIFIER + manaIndex, base_regen + CalculatePct(spirit_regen, modManaRegenInterrupt));
-    SetStatFloatValue(UNIT_FIELD_POWER_REGEN_FLAT_MODIFIER + manaIndex, 0.001f + spirit_regen + base_regen);
+    SetStatFloatValue(UNIT_FIELD_POWER_REGEN_INTERRUPTED_FLAT_MODIFIER + manaIndex, base_regen * modManaRegenInterrupt);
```
Shauren: This is definitely wrong. SPELL_AURA_MOD_MANA_REGEN_INTERRUPT is a percentage (but im sure this aura isnt used anymore for calculating this field, all chars i tested had equal values in both)

DoctorKraft: Yeah i agree, but i have found some spells with this aura for NPC. Maybe move this aura for unit (creature) only? or remove?

Shauren: Remove it

DoctorKraft: Done

src/server/game/Entities/Unit/StatSystem.cpp
```diff
@@ -743,19 +744,21 @@ void Player::UpdateManaRegen()
...
+    // Apply PCT bonus from SPELL_AURA_MOD_POWER_REGEN_PERCENT
+    base_regen *= GetTotalAuraMultiplierByMiscValue(SPELL_AURA_MOD_POWER_REGEN_PERCENT, POWER_MANA);
 
-    SetStatFloatValue(UNIT_FIELD_POWER_REGEN_INTERRUPTED_FLAT_MODIFIER + manaIndex, base_regen + CalculatePct(spirit_regen, modManaRegenInterrupt));
-    SetStatFloatValue(UNIT_FIELD_POWER_REGEN_FLAT_MODIFIER + manaIndex, 0.001f + spirit_regen + base_regen);
+    // Apply PCT bonus from SPELL_AURA_MOD_MANA_REGEN_PCT
+    base_regen *= GetTotalAuraMultiplierByMiscValue(SPELL_AURA_MOD_MANA_REGEN_PCT, POWER_MANA);
+
+    SetStatFloatValue(UNIT_FIELD_POWER_REGEN_INTERRUPTED_FLAT_MODIFIER + manaIndex, base_regen);
+    SetStatFloatValue(UNIT_FIELD_POWER_REGEN_FLAT_MODIFIER + manaIndex, base_regen);
```
joschiwald: regen and interrupted regen is always the same?

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -444,7 +444,7 @@ pAuraEffectHandler AuraEffectHandler[TOTAL_AURAS]=
     &AuraEffect::HandleNULL,                                      //376 SPELL_AURA_MOD_CURRENCY_GAIN_FROM_SOURCE
     &AuraEffect::HandleNULL,                                      //377 SPELL_AURA_CAST_WHILE_WALKING_2
     &AuraEffect::HandleNULL,                                      //378
-    &AuraEffect::HandleNULL,                                      //379
+    &AuraEffect::HandleNoImmediateEffect,                         //379 SPELL_AURA_MOD_MANA_REGEN_PCT implemented in Player::UpdateManaRegen
```
joschiwald: i think on apply/remove you should call UpdateManaRegen

src/server/game/Entities/Unit/StatSystem.cpp
DoctorKraft: Yes, shauren has also mentionned this in top "all chars i tested had equal values in both"

src/server/game/Spells/Auras/SpellAuraEffects.cpp
DoctorKraft: rly? because they fields are correctly updated after a change spec

joschiwald: what happen if such aura is applied spec change independent

src/server/game/Entities/Unit/StatSystem.cpp
joschiwald: what happend to SPELL_AURA_MOD_MANA_REGEN_INTERRUPT, there are still auras of that type

Shauren: iirc there are only 4 and none of them are usable

src/server/game/Spells/Auras/SpellAuraEffects.cpp
DoctorKraft: Fixed

PR #20949, opened 2017-11-21
Fixed less and less objects from pools being spawned the longer the server is running

### LINKED ISSUE #19257: Spawns in Pools set only at server restart, not on respawn

**Description:**
Bloodstone Ore (ID 4278) mining nodes (ID 2653) do not appear in Drywhisker Gorge (Arathi Highlands) for the quest "Favor for Krazek" (ID 627)

**Branch(es):**
3.3.5

**TC rev. hash/commit:** 
TrinityCore rev. 8e865a60c815 2017-03-07 20:20:19 -0300 (3.3.5 branch) (Win64, Release, Static)

**TDB version:** 
TDB 335.62

**Operating system:**
Windows 10 Server & Client

lachtanek:
**Changes proposed:**

- Problem with old implementation is that it doesn't check already spawned objects when choosing new ones to spawn and it can choose the same one incorrectly multiple times.
- Rewritten it so that it now takes into account if object is already spawned (still not 100% chance)

**Target branch(es):**

- [x] 3.3.5
- [x] master ? (not sure)

**Issues addressed:** Closes #11141, maybe #19257

**Tests performed:**
- running on live server

- [ ] I don't have any idea how it worked on retail, but figured out that possibility of choosing only some objects was added on purpose, so I kinda reimplemented it (PoolMgr.cpp:356, subject to changes)
- [ ] Obviously with this implementation it happens more frequently that for example 3 veins are spawned on top of one another (titanium, rich saronite, saronite), but I don't think this is up to pool system implementation to fix.
- [ ] Too long title ^_^

### DIFF
--- a/src/server/game/Pools/PoolMgr.cpp
+++ b/src/server/game/Pools/PoolMgr.cpp
@@ -153,34 +153,6 @@ bool PoolGroup<T>::CheckPool() const
     return true;
 }
 
-template <class T>
-PoolObject* PoolGroup<T>::RollOne(ActivePoolData& spawns, uint32 triggerFrom)
-{
-    if (!ExplicitlyChanced.empty())
-    {
-        float roll = (float)rand_chance();
-
-        for (uint32 i = 0; i < ExplicitlyChanced.size(); ++i)
-        {
-            roll -= ExplicitlyChanced[i].chance;
-            // Triggering object is marked as spawned at this time and can be also rolled (respawn case)
-            // so this need explicit check for this case
-            if (roll < 0 && (ExplicitlyChanced[i].guid == triggerFrom || !spawns.IsActiveObject<T>(ExplicitlyChanced[i].guid)))
-               return &ExplicitlyChanced[i];
-        }
-    }
-    if (!EqualChanced.empty())
-    {
-        uint32 index = urand(0, EqualChanced.size()-1);
-        // Triggering object is marked as spawned at this time and can be also rolled (respawn case)
-        // so this need explicit check for this case
-        if (EqualChanced[index].guid == triggerFrom || !spawns.IsActiveObject<T>(EqualChanced[index].guid))
-           return &EqualChanced[index];
-    }
-
-    return nullptr;
-}
-
 // Main method to despawn a creature or gameobject in a pool
 // If no guid is passed, the pool is just removed (event end case)
 // If guid is filled, cache will be used and no removal will occur, it just fill the cache
@@ -340,7 +312,6 @@ void PoolGroup<Pool>::RemoveOneRelation(uint32 child_pool_id)
 template <class T>
 void PoolGroup<T>::SpawnObject(ActivePoolData& spawns, uint32 limit, uint32 triggerFrom)
 {
-    uint32 lastDespawned = 0;
     int count = limit - spawns.GetActiveObjectCount(poolId);
 
     // If triggered from some object respawn this object is still marked as spawned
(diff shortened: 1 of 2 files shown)


lachtanek:
@offl Probably both, edited main post

offl:
Theoretically it closes both issues, because now more objects will be spawned.
But #19257 is about this:
> Spawns in pools only gets set at server restart not on respawn so the same will always spawn until restart then will spawn differently.

lachtanek:
My bad then, I misunderstood that one

src/server/game/Pools/PoolMgr.cpp
```diff
@@ -349,32 +320,70 @@ void PoolGroup<T>::SpawnObject(ActivePoolData& spawns, uint32 limit, uint32 trig
...
         {
-            ReSpawn1Object(obj);
-            triggerFrom = 0;
-            continue;
+            while (count && ExplicitlyChanced.size() > rolledObjects.size())
+            {
+                --count;
+                float roll = (float)rand_chance();
+
+                for (auto obj : ExplicitlyChanced)
```
Shauren: Two things here - first, you are creating a copy of `PoolObject` every time (no ref on type) and second, don't use auto if the type is not `iterator` (also avoids the first mistake you made since it makes you think about type you are using)

src/server/game/Pools/PoolMgr.cpp
```diff
@@ -349,32 +320,70 @@ void PoolGroup<T>::SpawnObject(ActivePoolData& spawns, uint32 limit, uint32 trig
...
+            }
+
+            Trinity::Containers::RandomResize(rolledObjects, count);
         }
-        spawns.ActivateObject<T>(obj->guid, poolId);
-        Spawn1Object(obj);
 
-        if (triggerFrom)
+        // try to spawn rolled objects
+        for (auto obj : rolledObjects)
```
Shauren: Don't auto

src/server/game/Pools/PoolMgr.cpp
```diff
@@ -349,32 +320,70 @@ void PoolGroup<T>::SpawnObject(ActivePoolData& spawns, uint32 limit, uint32 trig
...
+                        break;
+                    }
+                }
+            }
+        }
+        else if (!EqualChanced.empty())
+        {
+            rolledObjects = EqualChanced;
+
+            for (auto itr = rolledObjects.begin(); itr != rolledObjects.end();)
```
Shauren: and here auto is ok, its `iterator`

Shauren (changes requested):

Shauren:
And for the second point you mentioned  (spawns on top of each other), that is a problem with how the author made db data (he fucked up), such spawns are supposed to be put into a pool of pools where the innermost pool has data for a single spawn position with different object types only (and the outer pool determines which point is active)

src/server/game/Pools/PoolMgr.cpp
lachtanek: TIL don't rely on compiler optimizations too much. I guess auto& would also work, I just like it more than writing types when I don't have to.

lachtanek:
@Shauren thanks for review, will update the code.
About second point, just wanted to mention it here, to prevent unwanted questions.

src/server/game/Pools/PoolMgr.cpp
Shauren: This here isnt really about compiler optimizations btw but TC style guide - no auto except iterators (or other long nested types)

Shauren:
To sum it up, this fixes #11141 or #19257?

Shauren (approved):

offl:
> Spawns in pools only gets set at server restart not on respawn so the same will always spawn until restart then will spawn differently.

If something here fixes this then this PR fixes #19257 and #11141, otherwise only #11141

kelno:
Reading the code, what happens if the triggerFrom object is chosen? It is removed from map here : https://github.com/TrinityCore/TrinityCore/blob/869affa89300c24b85896a050b58241714786454/src/server/game/Entities/GameObject/GameObject.cpp#L794
How is is respawned on map?

lachtanek:
@kelno not sure how the code you linked relates with pools, code I edited is called after object (GO/creature) respawn, not on despawn

kelno:
Sorry for the lack of clarity. If I'm understanding this correctly, taking an ore vein for example :
1- Player mine the veins, veins get removed from Map and respawnTime set (linked code in previous comment)
2- respawnTime expires and PoolMgr is notified [here](https://github.com/TrinityCore/TrinityCore/blob/869affa89300c24b85896a050b58241714786454/src/server/game/Entities/GameObject/GameObject.cpp#L570)
3- PoolMgr decides which object to respawn and in the case it decides to respawn the triggering vein, call to `Respawn1Object` [here](https://github.com/TrinityCore/TrinityCore/blob/869affa89300c24b85896a050b58241714786454/src/server/game/Pools/PoolMgr.cpp#L373). 
I believe in this scenario the triggering vein does not get respawned on map but is still counting in the ActiveObject list. Note that I've not tested this yet.

tkrokli:
If a part of the code can never be reached, I think it would show up in the Coverity scan:
- https://scan.coverity.com/projects/4656
- https://scan6.coverity.com/reports.htm#v35230/p10834

lachtanek:
I think @kelno is right, not sure why I didn't notice it. Thank you

PR #20950, opened 2017-11-21
Core/Scripting: add QuestScript with 2 first hooks

Traesh:
**Changes proposed:**

-  DB linked script for quests

**Target branch(es):**

- master

No test performed yet

- [ ] Add more hooks ?

### DIFF
--- a/src/server/game/Scripting/ScriptMgr.cpp
+++ b/src/server/game/Scripting/ScriptMgr.cpp
@@ -118,6 +118,10 @@ template<>
 struct is_script_database_bound<SceneScript>
     : std::true_type { };
 
+template<>
+struct is_script_database_bound<QuestScript>
+    : std::true_type { };
+
 enum Spells
 {
     SPELL_HOTSWAP_VISUAL_SPELL_EFFECT = 40162 // 59084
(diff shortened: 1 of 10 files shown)


funjoker:
OnQuestComplete
OnQuestReward
OnQuestObjectiveComplete
OnQuestObjectiveProgress(Some objectives have a bar with % filled up specifying the amount which should be reached to get triggered as parameter)
OnQuestFailed
OnQuestAccept(maybe a non NPC/GO/ITEM/SPELL bound, general way)
OnQuestAbandon

These are prolly all hooks which could be useful ^^

Seyden:
We only need 1 hook. OnQuestStatusChange... then you can check which is the new status..

funjoker:
You never rly fix quest zones. I say it's useful to have those

Seyden:
it doesnt make any sense to implement this hooks.. you have OnQuestStatusChanged.. what do you want more? This hook is a bundle of all these hooks that you want

Shauren:
I agree with @Seyden - these 2 new hooks give you old and new statuses as arguments (and are called everywhere)

Shauren (approved):

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -748,6 +752,35 @@ class ScriptRegistrySwapHooks<SceneScript, Base>
     bool swapped;
 };
 
+/// This hook is responsible for swapping QuestScript's
```
joschiwald: can you explain when we need these ScriptRegistrySwapHooks? i was thinking that we don't need such hooks for not "living" objects and scripts not bound to any object

malcrom: Traesh, can you respond to this?

Traesh: I think joschiwald is right, maybe @Naios could confirm

Naios: Yes, the hook isn't needed here.
It's only needed for scripts with a custom unload logic such as database bound scripts.

Traesh: Well this script is database bound, it's just not entity linked

Naios: @Traesh Yes, you are right, my bad. The code is fine.
I was just wondering the fact that you only implemented the assertions,
but this is equal to the hooks of `InstanceMapScript` and maybe those could be merged in the future.
:+1:

PR #20955, opened 2017-11-22
Core/Entities: add extra use of UPDATETYPE_CREATE_OBJECT2

sirikfoll:
Mixup of commits TrinityCore@533517b and  cmangos/mangos-wotlk@89c2c8a

**Changes proposed:**

Use UPDATETYPE_CREATE_OBJECT2 for newly created objects of type:
-  Player, Pet, Corpse, DynamicObject
-  Summoned Units and Vehicles
-  Summoned GameObject wich has player as summoner

- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -646,9 +646,9 @@ bool Map::AddToMap(T* obj)
 
     //something, such as vehicle, needs to be update immediately
     //also, trigger needs to cast spell, if not update, cannot see visual
-    obj->SetItsNewObject(true);
+    obj->SetIsNewObject(true);
     obj->UpdateObjectVisibilityOnCreate();
-    obj->SetItsNewObject(false);
+    obj->SetIsNewObject(false);
     return true;
 }
 
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -646,7 +646,9 @@ bool Map::AddToMap(T* obj)
 
     //something, such as vehicle, needs to be update immediately
     //also, trigger needs to cast spell, if not update, cannot see visual
+    obj->SetItsNewObject(true);
     obj->UpdateObjectVisibilityOnCreate();
+    obj->SetItsNewObject(false);
     return true;
 }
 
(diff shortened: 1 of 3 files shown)


tkrokli:
if you want AppVeyor to show correct result, you need https://github.com/TrinityCore/TrinityCore/commit/5241c1ad71bade65f431ce9b1c4e142c8386bf2b (build fix)

Keader:
nice, fix alot of animations

Shauren:
I'm sure @r00ty-tc has something to say about this

r00ty-tc:
Yeah, I did see this. But I was working at the time. 

I'm unsure whether we should take the cmangos commit as-is, or just go with what to me, seems to make sense. That is, Create2 is used when a new object is created (any kind) and the packet signifies the creation and Create1 is used when an existing object is being sent to players it has now become relevant to. 

Not getting round to testing all object types to see if it made sense (and matched sniffs) is the main reason mine is sitting as a commit on my own repo. My suspicion is that either way is likely to have the desired effect.

If we go with a port from cmangos, it's probably fine. However I really dislike the naming. I'd suggest changing "itsnewobject" to something a little more like similar names "isNewObject" perhaps.

sirikfoll:
It's completely up to you, I didn't do any research about this, just got the ideas from existing code and changed it to work for all summons, because I need it for the work i'm doing in Kologarn 😄

src/server/game/Maps/Map.cpp
```diff
@@ -646,7 +646,9 @@ bool Map::AddToMap(T* obj)
 
     //something, such as vehicle, needs to be update immediately
     //also, trigger needs to cast spell, if not update, cannot see visual
+    obj->SetItsNewObject(true);
```
tkrokli: (removed the `/home/travis/build/TrinityCore/TrinityCore/` part from the error message)
```
src/server/game/Maps/Map.cpp:649:10:
 fatal error: no member named 'SetItsNewObject' in 'Corpse'
    obj->SetItsNewObject(true);
    ~~~  ^
src/server/game/Maps/Map.cpp:3635:32:
 note: in instantiation of function template specialization 'Map::AddToMap<Corpse>' requested here
template TC_GAME_API bool Map::AddToMap(Corpse*);
                               ^
1 error generated.
```

r00ty-tc:
Yeah, to be honest I'm happy to go with this way, until we do some more research of our own. My belief is that a lot of the conditions that are there based on type are because you will simply see type1 or type2 more for that kind of object. 

For example the check for GO for owner is player. Of course sniffs will show that because if a player spawns a gameobject from spell etc, they will see the type 2, and then anyone coming into range subsequently will see type 1.

But until we prove it for sure, this does solve the main issue, which is the correct go/creature anims.

jackpoz:
@r00ty-tc so this can be merged ?

Issue #19496, opened 2017-04-23
[master] Hunter Spell Auto-Shot


MartGon:
**Description:**

Spell bug that happens when auto-shot can't hit it's target.

**Current behaviour:**

The spell works correctly until you cannot hit the target for whatever reason, looking back or when you cannot see the target. After that, the spells gets stuck, an error message appears on the client ("you're in the wrong zone") and it keeps flickering on the action bar, but any shot won't be launch. Then you'll have to reactivate the spell or target something else to fix it.

**Expected behaviour:**

Auto-shot should keep launching ranged attacks if the target is now visible and in front of you, even if you couldn't hit it before.

**Steps to reproduce the problem:**

1. Create a hunter.
2. Right-click a target dummy to start auto-shot.
3. Look back like 3 seconds, error message appears on the client
4. Look again to the dummy.

**Branch(es):**

master

**TC rev. hash/commit:** 

TrinityCore rev. 5233caf09206 2017-04-20 19:53:34 +0200 (master branch) (Win64, RelWithDebInfo, Static) (worldserver)

**TDB version:**  7.2.0

**Operating system:** Windows 7 Professional x64

Killyana:
This issue is fixed on 3.3.5: https://github.com/TrinityCore/TrinityCore/pull/14889

MartGon:
Yep, that correctly fixes the issue. I knew that this spell worked properly on 3.3.5, I thought it was a problem  of the new core, I didn't know it wasn't implemented. 
I'm sorry I didn't found this fix myself. Thx for your help!

ariel-:
Don't close issue until it's ported to master

aquadeus:
ths was ported to master time ago

PR #20960, opened 2017-11-23
Core/Entities: fix Hunter Autoshot

DoctorKraft:
**Changes proposed:**

Since cata, autoshot of hunter can be cast when moving, currently the autorepeat system break this feature because only spells with the result SPELL_CAST_OK can be done. Also he don't have a cast result when player are LoS of the target or bad orientation, etc..

**Target branch(es):**

- [ ] 3.3.5
- [X] master

- [ ] Fix double cast result (the one with "Interrupted" )
- [ ] Fix spam error when target are out of vision (building for example)

This PR are a mix with https://github.com/TrinityCore/TrinityCore/pull/18918


offl:
Closes #19496?

DoctorKraft:
@offl exactly i have copied the fix and adjust with "moving" problem on master branch

PR #20981, opened 2017-11-27
Core/Entities: Creature powers

### LINKED ISSUE #12487: [335] Hunter pets should not have 0 focus at login
**Core:** c28345e279f31af2fa32df2411325ccfabe06fc8
**DB:** TDB_full_335.52 + updates

**BUG:**

When you log in or change zones, Hunter pet focus gets set to 0 and slowly recharges back to 100% even if the pet already had 100%.

**Expected:**

When you log in or change zones a pet should have 100% focus if it had 100% focus when you logged out or initiated the zone change.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3108576-335-hunter-pets-should-not-have-0-focus-at-login?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #17476: Death Knight Pet has no energy
**Description**:
Death Knigth Pet is summoned but has no energy
I am able to summon the risen ghoul but this has no energy so it cannot cast any spells , maybe due to this
Pet should have a energy bar
1. Use Raise Dead
2. pet has no energy bar
3. 

6.x
**TC hash/commit**:  
68473cc 
**TDB version**:  
TDB_full_6.04_2016_04_11.7z
**Operating system**:  
windows 10

Mihc3:
**Changes proposed:**
- Power type handling for creatures (fixes creatures to show no power bar, vehicles to show energy, some pets etc.):
    - By default we won't set max power (makes creatures with unit class 1, 4 have no power displayed - max power = 0).
    - Creatures with unit class 2, 8 have max mana power set.
    - Vehicle initialization sets power type on its own from dbc (already scripted).
    - Guardian UpdateAllStats handles setting power type (hunter pet has focus, dk ghoul has energy, everything else has mana - **let me know if any pet type is missing**)

- 3.3.5 ❓ idk
- master ✔️

**Issues addressed:**
Closes #17476
Closes #12487

Build OK. Check if units display powers correctly: *(tested on 7.3.2 - 13 Jan 2018)*
- [x] Player with mana: priest ✔️
- [x] Player with rage: warrior ✔️
- [x] Player with energy: rogue ✔️
- [x] Player with focus: hunter ✔️
- [x] Player with runic power: DK ✔️
- [ ] Warlock: Soul Shard regeneration ❌ 
- [x] Druid forms: mana ✔️, energy ✔️, rage ✔️, astral power ✔️
- [x] Pet with energy: ghoul ✔️
- [x] Pet with focus: hunter's pet ✔️
- [x] Creature with no power.
- [x] Creature with mana.
- [x] Creature with energy: Spatial Anomaly ✔️
- [x] Vehicle without power: Living Vine ✔️
- [x] Vehicle with energy: Ammunae, Rajh ✔️

Am I missing something?

* Soul Shard regeneration does not work for Warlocks.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -1273,23 +1273,18 @@ void Creature::UpdateLevelDependantStats()
 
     switch (getClass())
     {
-        case CLASS_WARRIOR:
-            SetPowerType(POWER_RAGE);
+        case CLASS_PALADIN:
+        case CLASS_MAGE:
+            SetMaxPower(POWER_MANA, mana);
+            SetFullPower(POWER_MANA);
             break;
-        case CLASS_ROGUE:
-            SetPowerType(POWER_ENERGY);
-            break;
-        default:
-            SetMaxPower(POWER_MANA, mana); // MAX Mana
-            SetPower(POWER_MANA, mana);
+        default: // We don't set max power here, 0 makes power bar hidden
             break;
     }
 
     SetModifierValue(UNIT_MOD_HEALTH, BASE_VALUE, (float)health);
-    SetModifierValue(UNIT_MOD_MANA, BASE_VALUE, (float)mana);
 
     // damage
-
     float basedamage = stats->GenerateBaseDamage(cInfo);
 
     float weaponBaseMinDamage = basedamage;
--- a/src/server/game/Entities/Creature/TemporarySummon.h
+++ b/src/server/game/Entities/Creature/TemporarySummon.h
@@ -21,6 +21,24 @@
 
 #include "Creature.h"
 
+enum PetEntry
+{
+    // Warlock pets
+    PET_IMP             = 416,
+    PET_FEL_HUNTER      = 691,
+    PET_VOID_WALKER     = 1860,
+    PET_SUCCUBUS        = 1863,
+    PET_DOOMGUARD       = 18540,
+    PET_FELGUARD        = 30146,
+
+    // Death Knight pets
+    PET_GHOUL           = 26125,
+    PET_ABOMINATION     = 106848,
+
+    // Shaman pet
+    PET_SPIRIT_WOLF     = 29264
+};
+
 struct SummonPropertiesEntry;
 
 class TC_GAME_API TempSummon : public Creature
@@ -59,8 +77,22 @@ class TC_GAME_API Minion : public TempSummon
         Unit* GetOwner() const { return m_owner; }
         float GetFollowAngle() const override { return m_followAngle; }
         void SetFollowAngle(float angle) { m_followAngle = angle; }
-        bool IsPetGhoul() const {return GetEntry() == 26125;} // Ghoul may be guardian or pet
-        bool IsSpiritWolf() const {return GetEntry() == 29264;} // Spirit wolf from feral spirits
+
+        // Warlock pets
+        bool IsPetImp() const { return GetEntry() == PET_IMP; }
+        bool IsPetFelhunter() const { return GetEntry() == PET_FEL_HUNTER; }
+        bool IsPetVoidwalker() const { return GetEntry() == PET_VOID_WALKER; }
+        bool IsPetSuccubus() const { return GetEntry() == PET_SUCCUBUS; }
+        bool IsPetDoomguard() const { return GetEntry() == PET_DOOMGUARD; }
+        bool IsPetFelguard() const { return GetEntry() == PET_FELGUARD; }
+
+        // Death Knight pets
+        bool IsPetGhoul() const { return GetEntry() == PET_GHOUL; } // Ghoul may be guardian or pet
+        bool IsPetAbomination() const { return GetEntry() == PET_ABOMINATION; } // Sludge Belcher dk talent
+
+        // Shaman pet
+        bool IsSpiritWolf() const { return GetEntry() == PET_SPIRIT_WOLF; } // Spirit wolf from feral spirits
+
         bool IsGuardianPet() const;
     protected:
         Unit* const m_owner;
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -226,7 +226,6 @@ bool Pet::LoadPetFromDB(Player* owner, uint32 petEntry, uint32 petnumber, bool c
             SetSheath(SHEATH_STATE_MELEE);
             SetByteFlag(UNIT_FIELD_BYTES_2, UNIT_BYTES_2_OFFSET_PET_FLAGS, fields[9].GetBool() ? UNIT_CAN_BE_ABANDONED : UNIT_CAN_BE_RENAMED | UNIT_CAN_BE_ABANDONED);
             SetUInt32Value(UNIT_FIELD_FLAGS, UNIT_FLAG_PVP_ATTACKABLE); // this enables popup window (pet abandon, cancel)
-            SetPowerType(POWER_FOCUS);
             break;
         default:
             if (!IsPetGhoul())
@@ -780,7 +779,6 @@ bool Pet::CreateBaseAtTamed(CreatureTemplate const* cinfo, Map* map)
     if (!Create(map->GenerateLowGuid<HighGuid::Pet>(), map, cinfo->Entry))
         return false;
 
-    SetPowerType(POWER_FOCUS);
     SetUInt32Value(UNIT_FIELD_PET_NAME_TIMESTAMP, 0);
     SetUInt32Value(UNIT_FIELD_PETEXPERIENCE, 0);
     SetUInt32Value(UNIT_FIELD_PETNEXTLEVELEXP, uint32(sObjectMgr->GetXPForLevel(getLevel()+1)*PET_XP_FACTOR));
@@ -860,13 +858,12 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
     for (uint8 i = SPELL_SCHOOL_HOLY; i < MAX_SPELL_SCHOOL; ++i)
         SetModifierValue(UnitMods(UNIT_MOD_RESISTANCE_START + i), BASE_VALUE, float(cinfo->resistance[i]));
 
-    //health, mana, armor and resistance
+    // Health, Mana or Power, Armor
     PetLevelInfo const* pInfo = sObjectMgr->GetPetLevelInfo(creature_ID, petlevel);
     if (pInfo)                                      // exist in DB
     {
         SetCreateHealth(pInfo->health);
-        if (petType != HUNTER_PET) //hunter pet use focus
-            SetCreateMana(pInfo->mana);
+        SetCreateMana(pInfo->mana);
 
         if (pInfo->armor > 0)
             SetModifierValue(UNIT_MOD_ARMOR, BASE_VALUE, float(pInfo->armor));
@@ -887,6 +884,17 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
         SetCreateStat(STAT_INTELLECT, 28);
     }
 
+    // Power
+    if (petType == HUNTER_PET) // Hunter pets have focus
+        SetPowerType(POWER_FOCUS);
+    else if (IsPetGhoul() || IsPetAbomination()) // DK pets have energy
+        SetPowerType(POWER_ENERGY);
+    else if (IsPetImp() || IsPetFelhunter() || IsPetVoidwalker() || IsPetSuccubus() || IsPetDoomguard() || IsPetFelguard()) // Warlock pets have energy (since 5.x)
+        SetPowerType(POWER_ENERGY);
+    else 
+        SetPowerType(POWER_MANA);
+
+    // Damage
     SetBonusDamage(0);
     switch (petType)
     {
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -1786,6 +1786,7 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         void UpdateArmor() override;
         void UpdateMaxHealth() override;
         void UpdateMaxPower(Powers power) override;
+        uint32 GetPowerIndex(Powers power) const override;
         void UpdateAttackPowerAndDamage(bool ranged = false) override;
         void ApplySpellPowerBonus(int32 amount, bool apply);
         void UpdateSpellDamageAndHealingBonus();
--- a/src/server/game/Entities/Unit/StatSystem.cpp
+++ b/src/server/game/Entities/Unit/StatSystem.cpp
@@ -300,8 +300,17 @@ void Player::UpdateMaxHealth()
     SetMaxHealth((uint32)value);
 }
 
+uint32 Player::GetPowerIndex(Powers power) const
+{
+    return sDB2Manager.GetPowerIndexByClass(power, getClass());
+}
+
 void Player::UpdateMaxPower(Powers power)
 {
+    uint32 powerIndex = GetPowerIndex(power);
+    if (powerIndex == MAX_POWERS || powerIndex >= MAX_POWERS_PER_CLASS)
+        return;
+
     UnitMods unitMod = UnitMods(UNIT_MOD_POWER_START + power);
 
     float value = GetModifierValue(unitMod, BASE_VALUE) + GetCreatePowers(power);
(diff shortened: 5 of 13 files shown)


Shauren:
How does this PR affect nonpet, nonvehicle and nonplayer creatures that should be using powers other than mana? (Creatures with energy are a thing)

Mihapro:
That check should really not be there, but then every mob in Elwynn Forest has rage power (because of unit_class).

What could make creature hide its power bar? 🤔

Ovahlord:
shape shift auras

Mihapro:
Trying now that by default power type is not set (except for paladin and mage unit class). Vehicle have their own init where they set power type and so do pets/guardians. If any other creature uses power bar different than mana, then it should have an mod power display aura (179) - something like that?

src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -860,13 +860,23 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
...
+    //health, mana/power, armor and resistance
     PetLevelInfo const* pInfo = sObjectMgr->GetPetLevelInfo(creature_ID, petlevel);
     if (pInfo)                                      // exist in DB
     {
         SetCreateHealth(pInfo->health);
-        if (petType != HUNTER_PET) //hunter pet use focus
-            SetCreateMana(pInfo->mana);
+
+        // Choose power type based on unit_class, hunter pet already set to focus 
+        if (petType != HUNTER_PET) switch (cinfo->unit_class)
```
Traesh: afaik this is not a TC code standard, you should creature a new line for the switch

tkrokli: Yes, please move the
```cpp
switch (cinfo->unit_class)
```
part down to a new line with correct indent.
(look up existing  `switch` sections in TC code if you are uncertain)

Mihapro:
Is it really needed that creatures use power indexing / use `ChrClassesXPowerTypes.dbc`? It's probably mean't for players only, no?

Shauren:
Interesting point, I'll need to investigate

tkrokli:
You probably need to rebase your PR branch on current master branch to get rid of the errors outside of your PR changes.

src/server/game/Entities/Player/Player.h
```diff
@@ -22,6 +22,7 @@
 #include "Unit.h"
 #include "CUFProfile.h"
 #include "DatabaseEnvFwd.h"
+#include "DB2Stores.h"
```
Shauren: absolutely NO.
Move GetPowerIndex to Player.cpp

Shauren (changes requested):

Mihapro:
I will be away from home till Sunday, so I kindly ask you guys to update the PR. I commented on lines that need to be changed, it’s not hard :P

src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -887,6 +886,15 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
         SetCreateStat(STAT_INTELLECT, 28);
     }
 
+    // Mana/power
+    if (petType == HUNTER_PET) // Hunter pets have focus
+        SetPowerType(POWER_FOCUS);
+    if (IsPetGhoul()) // Ghouls have energy
```
Mihapro: ‘else if’ to make it look nicer.

Shauren:
Ok, I looked into creatures with multiple powers and here is the result
Revert the virtual split for all functions except GetPowerIndex and override it for creatures like this
```
if (power == GetPowerType())
    return 0;
if (power == POWER_ALTERNATE_POWER)
    return 1;
if (power == POWER_COMBO_POINTS)
    return 2;
return MAX_POWERS;
```

Shauren (changes requested):

src/server/game/Entities/Unit/Unit.h
```diff
@@ -19,6 +19,8 @@
 #ifndef __UNIT_H
 #define __UNIT_H
 
+#include "Player.h"
+#include "Creature.h"
```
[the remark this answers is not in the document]
Mihapro: Okay, I’ll move power functions into StatSystem.cpp. Might be that’s why VS build failed.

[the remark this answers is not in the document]
Shauren: No, don't move anything - build fails exactly because you added these 2 includes here

Mihapro:
Got it, thanks! 😁

src/server/game/Entities/Unit/StatSystem.cpp
```diff
@@ -846,11 +850,29 @@ void Creature::UpdateMaxHealth()
...
+    if (power == POWER_ALTERNATE_POWER)
+        return 1;
+    if (power == POWER_COMBO_POINTS)
+        return 2;
+    return MAX_POWERS;
+}
+
 void Creature::UpdateMaxPower(Powers power)
 {
+    if (power != GetPowerType())
```
Shauren: Always compare using GetPowerIndex(power) != MAX_POWERS

Mihapro: yup, == tho 👍

src/server/game/Entities/Unit/Unit.h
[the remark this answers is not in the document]
Mihapro: Many reverts, so I rebased everything - these includes were reverted too and it builds fine now. ℹ️

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -4051,16 +4051,7 @@ void AuraEffect::HandleAuraModIncreaseEnergyPercent(AuraApplication const* aurAp
...
-    if (apply)
-    {
-        target->HandleStatModifier(unitMod, TOTAL_PCT, amount, apply);
-        target->ModifyPowerPct(powerType, amount, apply);
-    }
-    else
-    {
-        target->ModifyPowerPct(powerType, amount, apply);
-        target->HandleStatModifier(unitMod, TOTAL_PCT, amount, apply);
-    }
```
joschiwald: why is this removed?

Mihapro: In case unit does not allow modifying stats (Unit::CanModifyStats()), we shouldn't force modify power here (I see in some player cases modify stats is set to false). I tested aura application/removal (for +pct mana and energy) and it works just fine, except if I'm missing something - not sure what the idea was here except to get the modified amount, which isn't used anyway.

joschiwald: when you gain this aura, current power should be increased too, on remove you lose the power again

Mihapro: Not if CanModifyStats is set to false (it makes sense that it should work like that), check Unit:9207, HandleStatModifier already updates max power. That's why it works exactly the same like before, this ugly code is not needed here :P

joschiwald: i'm not talking about max power, current power is affected

Mihapro: if you decrease max power, (current) power is set to max power (Unit.cpp, line 9513); if you increase max power, (current) power remains the same - expected behavior, no?

joschiwald: following that code, no

Mihapro: changing currect power here makes no sense, we should not give power, it should regenerate

joschiwald: example https://wow.gamepedia.com/Hymn_of_Hope

Mihapro: you are right, I only looked at 7.x dbc 😞 i'll revert this

Mihapro: rewritten tho, reducing for pct of max power that is already increased by pct is not right 😀

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -4039,28 +4039,23 @@ void AuraEffect::HandleAuraModIncreaseEnergyPercent(AuraApplication const* aurAp
...
+    float pct = float(GetAmount());
+
+    // Calculate the change
+    int32 change = target->GetMaxPower(powerType);
+    ApplyPct(change, apply ? pct : 100 - 100 / (1 + pct / 100)); // on removal (!apply) GetMaxPower is already increased by pct
```
Shauren: Beware of operator ?: precedence!
Current code behaves as if you wrote `(apply ? pct : 100) - 100 / (1 + pct / 100)`
Did you intend `apply ? pct : (100 - 100 / (1 + pct / 100))`?

Mihapro: Yep, thank you, forgot brackets 👍 😄

src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -887,6 +884,15 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
         SetCreateStat(STAT_INTELLECT, 28);
     }
 
+    // Power
+    if (petType == HUNTER_PET) // Hunter pets have focus
+        SetPowerType(POWER_FOCUS);
+    else if (IsPetGhoul() || IsPetAbomination()) // DK pets have energy
+        SetPowerType(POWER_ENERGY);
+    else 
+        SetPowerType(POWER_MANA);
```
Shauren: pretty sure this one is wrong - all warlock pets use energy (http://www.wowhead.com/imp-abilities)

Mihapro: Ok, this will be the only difference for now between 3.3.5a and 7.x. 3.3.5 could use this PR aswell, right?

Shauren: No idea about that

src/server/game/Entities/Creature/CreatureData.h
```diff
@@ -447,7 +447,7 @@ struct TC_GAME_API CreatureBaseStats
 
     uint32 GenerateHealth(CreatureTemplate const* info) const
     {
-        return uint32(ceil(BaseHealth[info->HealthScalingExpansion] * info->ModHealth * info->ModHealthExtra));
+        return uint32(std::lroundf(BaseHealth[info->HealthScalingExpansion] * info->ModHealth * info->ModHealthExtra));
```
joschiwald: ceil was used here, because some npcs have have very low ModHealth and to prevent generated health is 0

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -5172,33 +5172,11 @@ void Unit::SetPowerType(Powers new_powertype)
...
-            SetPower(POWER_FOCUS, uint32(std::ceil(GetCreatePowers(POWER_FOCUS) * powerMultiplier)));
-            break;
-        case POWER_ENERGY:
-            SetMaxPower(POWER_ENERGY, uint32(std::ceil(GetCreatePowers(POWER_ENERGY) * powerMultiplier)));
-            break;
-    }
+            pet->SetGroupUpdateFlag(GROUP_UPDATE_FLAG_PET_POWER_TYPE);*/
+    
+    UpdateMaxPower(new_powertype);
+    SetPower(new_powertype, 0);
```
joschiwald: does this cause you will lose all your mana on power type change, like druid shapeshift?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9482,10 +9473,8 @@ void Unit::SetPower(Powers power, int32 val)
             player->SetGroupUpdateFlag(GROUP_UPDATE_FLAG_CUR_POWER);
     }
     /*else if (Pet* pet = ToCreature()->ToPet()) TODO 6.x
-    {
         if (pet->isControlled())
-            pet->SetGroupUpdateFlag(GROUP_UPDATE_FLAG_PET_CUR_POWER);
-    }*/
+            pet->SetGroupUpdateFlag(GROUP_UPDATE_FLAG_PET_CUR_POWER);*/
```
joschiwald: it looks like the whole GroupUpdateFlag stuff is unused @Shauren ?

src/server/game/Entities/Unit/Unit.cpp
Mihapro: Hmm, true, moonkin keeps the same mana I think. But other forms do reset power to 0 (energy, rage), no? I could extend arguments here: `void Unit::SetPowerType(Powers new_powertype, bool resetCurrentPower = true)`

src/server/game/Entities/Creature/CreatureData.h
Mihapro: Then only 0 should be changed to 1. But rounding is always the case: https://docs.google.com/spreadsheets/d/12Tevx7OKlQqEOLe8dXJ8iBo47a_4YMx-W1A4P-5KEFI/edit?usp=sharing

src/server/game/Entities/Unit/Unit.cpp
funjoker: if power is the same dont reset maybe ?

src/server/game/Entities/Creature/CreatureData.h
Shauren: either way this is unrelated to powers - shouldnt be here

src/server/game/Entities/Unit/Unit.cpp
joschiwald: just restore the old way

src/server/game/Entities/Unit/Unit.cpp
Shauren: iirc we handle it completely wrong right now (and flags should be used in SMSG_PARTY_MEMBER_STATE_UPDATE which we dont ever send)

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1228,23 +1228,17 @@ void Creature::UpdateLevelDependantStats()
...
-        case CLASS_ROGUE:
-            SetPowerType(POWER_ENERGY);
-            break;
-        default:
-            SetMaxPower(POWER_MANA, mana); // MAX Mana
-            SetPower(POWER_MANA, mana);
+        case CLASS_PALADIN:
+        case CLASS_MAGE:
+            SetMaxPower(POWER_MANA, mana);
+            SetFullPower(POWER_MANA);
```
Shauren: add break;

src/server/game/Entities/Creature/TemporarySummon.h
```diff
@@ -59,8 +59,22 @@ class TC_GAME_API Minion : public TempSummon
         Unit* GetOwner() const { return m_owner; }
         float GetFollowAngle() const override { return m_followAngle; }
         void SetFollowAngle(float angle) { m_followAngle = angle; }
-        bool IsPetGhoul() const {return GetEntry() == 26125;} // Ghoul may be guardian or pet
-        bool IsSpiritWolf() const {return GetEntry() == 29264;} // Spirit wolf from feral spirits
+
+        // Warlock pets
+        bool IsPetImp() const { return GetEntry() == 416; }
```
MitchesD: could be a good idea to remove these magic numbers

src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -1037,6 +1045,7 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
 
     SetFullHealth();
     SetFullPower(POWER_MANA);
+    SetFullPower(POWER_FOCUS);
```
joschiwald: what about energy?

src/server/game/Entities/Creature/TemporarySummon.h
tkrokli: `NPC_IMP             = 416` is already enumerated in this file.

tkrokli:
`src/server/game/Entities/Creature/TemporarySummon.h` is ranked before
- `/src/server/scripts/Spells/spell_generic.cpp` and
- `src/server/scripts/Kalimdor/CavernsOfTime/CullingOfStratholme/culling_of_stratholme.cpp`

-- hence the 3 "fatal errors" in travis-ci regarding the direction of enum duplication.

src/server/game/Entities/Pet/Pet.cpp
Mihapro: Removing this line since `Unit::SetPowerType` was reverted and makes both focus and energy full again

Mihapro:
Description updated - I tested everything on 7.3.2 today. Only Soul Shard regeneration for Warlocks seems to be broken or not yet scripted.

funjoker:
this will just partly close #13510.

funjoker:
Soulshar regeneration is not a power issue directly.
The regeneration is as far as i know a periodic aura.
Correct me if i am wrong :D

Shauren (approved):

PR #21008, opened 2017-11-29
Core/Creature: implement extra flag to use offhand attacks

Ovahlord:
fixed initial offhand attack timer so the main and offhand attacks will now act blizzlike instead of getting executed at the same time

Note: since some creature equipment items cannot be identified as weapon since they are only visual items there is no way to detect this automaticly

**Changes proposed:**

-  added extra flag to allow creatures to use offhand attacks without having to enable it via c++ script
-  fixed offhand attack timer so the creature will launch its offhand attack after 50% of the attack time of the main attack

- [x] 3.3.5
- [x] master

**Tests performed:** tested ingame


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -557,6 +560,11 @@ bool Creature::UpdateEntry(uint32 entry, CreatureData const* data /*= nullptr*/,
 
     SetUInt32Value(UNIT_DYNAMIC_FLAGS, dynamicflags);
 
+    if (cInfo->flags_extra & CREATURE_FLAG_EXTRA_USE_OFFHAND_ATTACK)
+        SetCanDualWield(true);
+    else if (CanDualWield())
```
Ovahlord: For understanding reasons: CanDualWield() does not check if you can dual wield, instead it returns if it is enabled. So kinda bad naming for the function

Shauren: whats the harm with calling SetCanDualWield(false) anyway?

Ascathor: I am not entirely sure what the bit operator is returning, but assuming it's a boolean value, shouldn't something like this work ?

SetCanDualWield(cInfo->flags_extra & CREATURE_FLAG_EXTRA_USE_OFFHAND_ATTACK);

This would remove redundancy.

Ovahlord: Good catch, sir. Will change

joschiwald:
do you have an example where the npc does not have a dual wield aura?

Keader:
Illidan : https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Outland/BlackTemple/boss_illidan.cpp#L484

Ovahlord:
most npc's does not have any aura. simple example: the orcs at the human starting zone.

Shauren:
Dual wielding is not even an aura - its a spell effect

offl:
I think most(if not all) of NPCs which currently uses Dual Wield aura shouldn't have it. It looks like hack to force NPC use two weapons.
```sql
SELECT * FROM `smart_scripts` WHERE `comment` LIKE '%Dual Wield%';
```

Teppic1:
I just checked VanCleef as I'd coded the cast of dual wield taken from the SAI, and in the sniff I have there's no mention of the spell anywhere, either as an aura or a direct cast.

Seyden:
A lot of spells aren't casted in sniff.. somehow blizzard has a way to use spells without sending them

offl:
VanCleef and some other NPCs uses player spell(674). Some other uses 30798 with effect Learn Spell 674. Dunno how Blizzard can did that.
42459 looks like creature-spell.

Keep in mind that 98% of EAI(SAI) and CPP are not based on sniffs. Instead on wowhead and player cache(spell1-spell8).

Shauren (approved):

Shauren:
Anyone else has comments?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -5889,9 +5889,9 @@ bool Unit::Attack(Unit* victim, bool meleeAttack)
         SetUInt32Value(UNIT_NPC_EMOTESTATE, EMOTE_ONESHOT_NONE);
     }
 
-    // delay offhand weapon attack to next attack time
+    // delay offhand weapon attack by 50% of the base attack time
     if (haveOffhandWeapon() && GetTypeId() != TYPEID_PLAYER)
-        resetAttackTimer(OFF_ATTACK);
+        setAttackTimer(OFF_ATTACK, CalculatePct(GetAttackTime(OFF_ATTACK), 50));
```
Treeston: I am concerned with this - doesn't it have the potential to cause strange (faster than usual) attack patterns when switching targets?

You also seem to assume that `OFF_ATTACK` and `BASE_ATTACK` have the same attack timer here - the code doesn't match the PR description otherwise.

Ovahlord: they have the same attack timer since offhand attacks of creatures does not rely on weapon attack timers like players do. Instead they use the base attack timer as well. What I did here was only delaying the swing. The timers will still reset after execution.

Treeston:
I'd prefer something along these lines:
```cpp
// delay off-hand attack if we're starting to attack - don't hit with both weapons at the same time
if (haveOffhandWeapon() && GetTypeId() == TYPEID_UNIT && !getAttackTimer(BASE_ATTACK) && !getAttackTimer(OFF_ATTACK))
    setAttackTimer(OFF_ATTACK, uint32(GetFloatValue(UNIT_FIELD_BASEATTACKTIME)*0.5f))
```

PS: That's _assuming_ that this delay behavior is something that actually happens on retail. What are you basing this claim on?

Ovahlord:
Keep in Mind that GetFloatValue(UNIT_FIELD_BASEATTACKTIME) is pretty mich the same as GetAttackTime(OFF_ATTACK)
GetAttackTime(OFF_ATTACK) returns the regular base attack time of the offhand attack which gets set on creature creation. My change does not change anything regarding that mechanic. I corrected the first execution of the offhand attack. The rest remains as it was before. the setAttackTimer is not SetAttackTime

Treeston:
It's only "pretty much the same" until your creature has a haste aura applied to it.

PS: The core already has this kind of logic for players [here](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Player/Player.cpp#L1154-L1161). What makes you think creatures should use 50% of main hand speed (which might be upwards of 1.5 seconds) instead of the 200ms players use?

Ovahlord:
Good point regarding this. I could change that in this case

You can check it pretty easy by using a starter account and create a human character. Just tackle one those orcs or goblins with their daggers. They appear (at least to me and my connection) to attack at about 50% after the first main hand attack.

Treeston:
> They appear (at least to me and my connection) to attack at about 50% after the first main hand attack.

That is not how retail verification works. Sniff it.

Ovahlord:
retail
is that combat log from retail enough validation? Can't sniff that scenario cuz no sniffer for 7.3.2 but it should make clear enough that it happens after 50% of the first attack swing. (Base attack time = 2000)

Treeston:
That's rounded to second precision. How accurate is `/combatlog`?

Ovahlord:
`
21:46:45.138  SWING_DAMAGE,Creature-0-1461-0-24-49874-00002460AF,"Blackrock
21:46:46.148  SWING_DAMAGE,Creature-0-1461-0-24-49874-00002460AF,"Blackrock 
21:46:47.148  SWING_DAMAGE,Creature-0-1461-0-24-49874-00002460AF,"Blackrock 


`

Treeston:
OK, good enough. Still add some check so it only fires on first engage though, and not while switching when some of the weapons are already on swing timer.

Ovahlord:
>     // delay offhand weapon attack to 50% of the main hand attack time
>     if (haveOffhandWeapon() && GetTypeId() != TYPEID_PLAYER)
>     {
>         resetAttackTimer(OFF_ATTACK);
>         setAttackTimer(OFF_ATTACK, CalculatePct(GetFloatValue(UNIT_FIELD_BASEATTACKTIME), 50));
>     }

how about this?

Treeston:
`resetAttackTimer` is superfluous, and current attack time check is missing. See my comment above.

Ovahlord:
>     // delay offhand weapon attack to 50% of the main hand attack time
>     if (haveOffhandWeapon() && GetTypeId() != TYPEID_PLAYER)
>         setAttackTimer(OFF_ATTACK, getAttackTimer(BASE_ATTACK) + CalculatePct(GetFloatValue(UNIT_FIELD_BASEATTACKTIME), 50));

Treeston:
`setAttackTimer(OFF_ATTACK, std::max(getAttackTimer(OFF_ATTACK), getAttackTimer(BASE_ATTACK) + ...);`

Treeston (approved):

Faq:
If no more comments then merge?

tkrokli:
Thanks again @Ovahlord for the contribution! 👍
(also thanks to reviewers and contributors)

PR #21015, opened 2017-12-01
Core/Spells: add use for proc hitmask PROC_HIT_INTERRUPT

Ovahlord:
-  added use for PROC_HIT_INTERRUPT so spells with spell_effect_interrupt will now properly proc auras that require successful spell cast interrupts

**Target branch(es):** both

- [x] 3.3.5
- [x] master


Keader (approved):
looks good to me

Shauren (approved):

PR #21019, opened 2017-12-02
[DB/Trainer] Riding trainers fixed with new DB tables + correction

DantestyleXD:
**Changes proposed:**

-  Riding and Flying trainers fixed with new tables
-  Trainers gossip_menu fixed with new tables
-  Roxi Ramrocket script doesn't need anymore (Conditions fix)

- [ ] 3.3.5
- [x] master

Build ok, no errors.
Ingame: all trainers tested (training and gossip) and all of them work.
With them, Riding skill lvl required to learn a new skill is required now and work.

**Note:**
I commented that script because it may serve in future (My little opinion, that can be delete...as you wish)

### DIFF
--- a/sql/updates/world/master/9999_99_99_99_world.sql
+++ b/sql/updates/world/master/9999_99_99_99_world.sql
@@ -1,0 +1,372 @@
+-- Riding `trainer`s fix with new tables
+
+-- Eversong Forest
+-- Perascamin
+SET @Pesca=16280;
+DELETE FROM `creature_default_trainer` WHERE `CreatureID`=@Pesca;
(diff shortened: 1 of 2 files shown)


DoctorKraft:
It's totally false, now the trainerid it's not the id of npc but of the list (multiple trainer have same list of trainer)

Aokromes:
Better post sql fix using sniffs on existing ticket, we will drop c++ part.

DantestyleXD:
There is no a packet parser for the actually wow version, so I can not obtain that sniff.
With this file trainers work as well, with sniffs I can improve this fix.
Without them...well, there's no much to do...

PR #21022, opened 2017-12-02
.modify Command int16 limit removal

Hristiyanii:
**Changes proposed:**

-  Updated .morph command to support int64.
-

- [ ] 3.3.5

**Tests performed:*Works, tested in game.* (Does it build, tested in-game, etc.)

### DIFF
--- a/src/server/scripts/Commands/cs_modify.cpp
+++ b/src/server/scripts/Commands/cs_modify.cpp
@@ -1021,7 +1021,7 @@ public:
         if (!*args)
             return false;
 
-        uint64 display_id = (uint64)atoi((char*)args);
+        uint32 display_id = (uint32)atoi((char*)args);
 
         Unit* target = handler->getSelectedUnit();
         if (!target)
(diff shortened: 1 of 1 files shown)


src/server/scripts/Commands/cs_modify.cpp
```diff
@@ -1021,7 +1021,7 @@ class modify_commandscript : public CommandScript
         if (!*args)
             return false;
 
-        uint16 display_id = (uint16)atoi((char*)args);
+        uint64 display_id = (uint64)atoi((char*)args);
```
Shauren: It's still wrong - atoi does not return values > INT_MAX and this will cause a /W4 warning on msvc about argument truncation from uint64 to uint32 (SetDisplayId accepts uint32)

Change to uint32 and it will be good.

jackpoz:
Please take your time to test the code changes and debug them to ensure they do what you think they should do.

src/server/scripts/Commands/cs_modify.cpp
Hristiyanii: It works as off uint64 for me.

Hristiyanii: Changed (probably fixed if any issues occur) uint64 to 32

Shauren: Of course it works, it would just trigger a warning during compilation (and we want to avoid that)

PR #21033, opened 2017-12-08
C++14, interface targets and cotire for PCH

Naios:
**Changes proposed:**

-  Use interface targets rather than setting global variables for target properties
-  Make use of CMake target features in order to set the C++ standard automatically
-  Enable C++14, upgraded travis
-  Use the [cotire library](https://github.com/sakra/cotire) for precompiled headers (remove the old custom PCH code)
-  Make C++14 chrono literals globally available
-  remove the custom code for make_unique
-  upgrade fmtlib since it did yield some warnings (because the warning level is different now for the main project)

- [x] 3.3.5
- [x] master

\-

- [x] Builds

- [x] Correct some compiler flags marked with *TODO*

Feedback is welcome!


jackpoz:
which distro version support C++14 by default ? (i.e. does Ubuntu 16.04 LTS have it without adding additional repos ?)
I was wondering about upgrading travis to 16.04 instead of adding additional repos (but this is for another issue/PR)

Naios:
GCC 5.x supports C++14 partially (actually to some degree which is enough for our needs).
No, Ubuntu 16.04 LTS only provides GCC 6.x as part of the devel repo.

As far as I know GCC 6.x is provided by default in:
- Debian 9+
- Ubuntu 16.10+

Aokromes:
I changed wiki to make clang default compiler :)

Naios:
@Shauren could you take a short look on this?

Two points which need an extra look:
- The conversion of the compiler flags
- The removal of the /Zm MSVC workaround

Shauren:
With ninja on msvc (default when using "Open folder") this warning is back
`D:\Users\Shauren\Desktop\Sources\TrinityCore_3.3.5a\cl : Command line warning D9025: overriding '/W3' with '/W0'`, caused by removing stuff from  dep/CMakeLists.txt

Keader:
What will happen with unary_functions?

Shauren:
They are useless, you could go and remove their uses in C++03 code and it would still work as before

Naios:
@Shauren could you test whether the warning is still shown?
I corrected and tested it with ninja, it should be fixed now.

Shauren:
Was almost fixed, you will need to adjust it taking 3bb166dd5c014df9d5e9bea1479773ee406c0088 into account

Naios:
Thanks, I added the correction.
I'm going to merge this PR by today.

Aokromes:
RIP VS2015/GCC 4.9.

Issue #18576, opened 2016-12-24
[DEV] Core: Monster sparring


Killyana:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:** Initial support for monster sparring was implemented here https://github.com/TrinityCore/TrinityCore/pull/17673

Only melee attack are working now, spells still deal damage, Also some mechanism like deal damage only until health% is not implemented.

Old issue: https://github.com/TrinityCore/TrinityCore/issues/13258

**Branch(es):** 3.3.5 / master (Tell us which branch(es) this issue affects.)

**TC rev. hash/commit:**  rev. 3e51583e6cde

[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing C++, create a Pull Request.)

Killyana:
After further investigation all known npcs that mustn't take damage under 85% hp from other creature have the flag CREATURE_DIFFICULTYFLAGS_4_UNK4 = 0x00000080 -- 128
name from binary: No NPC damage below 85%" flag for unit(s)

Kittnz:
Then we just need a check?

kelno:
@Killyana How do you even get those names?

illfated:
@kelno : Looks like they are available in the source code in the `master` branch:

https://github.com/TrinityCore/TrinityCore/blob/f48b87054ed923f0af40065da90fea8bb233aff0/src/server/game/Entities/Creature/CreatureData.h#L34-L280

Not sure if they are available in 3.3.5 source, maybe one of the TC devs can tell you more.

kelno:
Thanks, but I've seen that, the " No NPC damage below 85%" flag for unit(s)" part is what I'm wondering about, since Killyana mention "name from binary"

illfated:
Not sure about that, maybe it is something from the DBC files?

PR #21043, opened 2017-12-10
Core/Creature: implement base support for sparring between creatures

Ovahlord:
*note: this system basicly blocks any damage once the defined health limit in the table has been reached.

AttackerEntry is the entry of the creature that deals damage while the VictimEntry is the creature that gets damaged. HealthLimitPct defines the health percentage at which the attacker entry cannot deal damage anymore. This handled by nullifying all damage. This allows us to define individual health limits for different entries.

Thanks to r00ty-TC for helping me with the unorderedMap part.

**Changes proposed:**

-  implemented database side defineable limits for damage in combat between creatures to prevent them killing each other without having to use c++ scripts every time

**Target branch(es):** master

- [x] master

Tested ingame, works so far

- [ ] DoT's are still reducing the health clientside. While the health stays correct serverside, the client still thinks that health is being lost.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.h
+++ b/src/server/game/Entities/Creature/Creature.h
@@ -226,6 +226,10 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         Player* SelectNearestPlayer(float distance = 0) const;
         Unit* SelectNearestHostileUnitInAggroRange(bool useLOS = false) const;
 
+        CreatureSparring const* GetSparringData(uint32 attackerEntry, uint32 victimEntry) const;
+        bool CanSparWith(Creature* victim) const;
+        float GetSparringHealthLimitPctFor(Creature* victim) const;
+
         void DoFleeToGetAssistance();
         void CallForHelp(float fRadius);
         void CallAssistance();
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -703,6 +703,17 @@ void Unit::DealDamageMods(Unit const* victim, uint32 &damage, uint32* absorb) co
 
 uint32 Unit::DealDamage(Unit* victim, uint32 damage, CleanDamage const* cleanDamage, DamageEffectType damagetype, SpellSchoolMask damageSchoolMask, SpellInfo const* spellProto, bool durabilityLoss)
 {
+    // Sparring Checks
+    if (Creature* me = ToCreature())
+        if (Creature* target = victim->ToCreature())
+            if (me->CanSparWith(target))
+            {
+                if (damage >= victim->GetHealth()) // safety check to prevent creatures getting accidentally killed when the percentage is very small
+                    damage = victim->GetHealth() - 1;
+                else if (target->GetHealthPct() <= me->GetSparringHealthLimitPctFor(target))
+                    damage = 0;
+            }
+
     if (victim->IsAIEnabled)
         victim->GetAI()->DamageTaken(this, damage);
 
@@ -1878,6 +1889,14 @@ void Unit::CalcAbsorbResist(DamageInfo& damageInfo)
             DealDamage(caster, splitDamage, &cleanDamage, DIRECT_DAMAGE, damageInfo.GetSchoolMask(), (*itr)->GetSpellInfo(), false);
             log.damage = splitDamage;
             log.absorb = split_absorb;
+
+            // Sparring Checks
+            if (Creature* me = caster->ToCreature())
+                if (Creature* target = damageInfo.GetVictim()->ToCreature())
+                    if (me->CanSparWith(target))
+                        if (target->GetHealthPct() <= me->GetSparringHealthLimitPctFor(target))
+                            log.damage = 0;
+
             SendSpellNonMeleeDamageLog(&log);
 
             // break 'Fear' and similar auras
@@ -1974,6 +1993,14 @@ void Unit::AttackerStateUpdate(Unit* victim, WeaponAttackType attType, bool extr
         CalculateMeleeDamage(victim, 0, &damageInfo, attType);
         // Send log damage message to client
         DealDamageMods(victim, damageInfo.damage, &damageInfo.absorb);
+
+        // Sparring Checks
+        if (Creature* me = ToCreature())
+            if (Creature* target = victim->ToCreature())
+                if (me->CanSparWith(target))
+                    if (target->GetHealthPct() <= me->GetSparringHealthLimitPctFor(target))
+                        damageInfo.HitInfo |= HITINFO_FAKE_DAMAGE;
+
         SendAttackStateUpdate(&damageInfo);
 
         DamageInfo dmgInfo(damageInfo);
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -629,6 +629,58 @@ void ObjectMgr::LoadCreatureTemplateAddons()
     TC_LOG_INFO("server.loading", ">> Loaded %u creature template addons in %u ms", count, GetMSTimeDiffToNow(oldMSTime));
 }
 
+void ObjectMgr::LoadCreatureSparringTemplate()
+{
+    uint32 oldMSTime = getMSTime();
+
+    //                                               0              1             2
+    QueryResult result = WorldDatabase.Query("SELECT AttackerEntry, VictimEntry, HealthLimitPct FROM creature_sparring_template");
+
+    if (!result)
+    {
+        TC_LOG_INFO("server.loading", ">> Loaded 0 creature template sparring definitions. DB table `creature_sparring_template` is empty.");
+        return;
+    }
+
+    uint32 count = 0;
+    do
+    {
+        Field* fields = result->Fetch();
+
+        uint32 entry = fields[0].GetUInt32();
+        uint32 victim = fields[1].GetUInt32();
+        float healthPct = fields[2].GetFloat();
+
+        if (!sObjectMgr->GetCreatureTemplate(entry))
+        {
+            TC_LOG_ERROR("sql.sql", "Creature template (Entry: %u) does not exist but has a record in `creature_sparring_template`", entry);
+            continue;
+        }
+
+        if (!sObjectMgr->GetCreatureTemplate(victim))
+        {
+            TC_LOG_ERROR("sql.sql", "Creature template (Entry: %u) does not exist but has a record in `creature_sparring_template`", entry);
+            continue;
+        }
+
+        if (healthPct > 100.0f)
+        {
+            TC_LOG_ERROR("sql.sql", "Sparring entry (Entry: %u, Victim: %u) exceeds the health percentage limit. Setting to 100.", entry, victim);
+            healthPct = 100.0f;
+        }
+
+        if (healthPct <= 0.0f)
+        {
+            TC_LOG_ERROR("sql.sql", "Sparring entry (Entry: %u, Victim: %u) has a negative or too small health percentage. Setting to 0.1.", entry, victim);
+            healthPct = 0.1f;
+        }
+
+        _creatureSparringTemplateStore[entry].emplace_back(CreatureSparring(victim, healthPct));
+    } while (result->NextRow());
+
+    TC_LOG_INFO("server.loading", ">> Loaded %u creature sparring templates in %u ms", count, GetMSTimeDiffToNow(oldMSTime));
+}
+
 void ObjectMgr::LoadCreatureScalingData()
 {
     uint32 oldMSTime = getMSTime();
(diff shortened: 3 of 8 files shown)


ccrs:
ummmm seems wrong
the full damage should be logged always

Ovahlord:
if you do so you fuck up the clientside health because the combat log alone is enough to mess up the clientside health

ccrs:
on retail damage logs keep appearing so... unkown flag/thing/we?

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -637,6 +637,52 @@ void ObjectMgr::LoadCreatureTemplateAddons()
     TC_LOG_INFO("server.loading", ">> Loaded %u creature template addons in %u ms", count, GetMSTimeDiffToNow(oldMSTime));
 }
 
+oid ObjectMgr::LoadCreatureSparringTemplate()
```
tkrokli:
```
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Globals/ObjectMgr.cpp:640:1:
 fatal error: unknown type name 'oid'; did you mean 'void'?

oid ObjectMgr::LoadCreatureSparringTemplate()
^~~
void

1 error generated.
```

meji46:
Before reaching health percentage:
> ServerToClient: SMSG_ATTACKER_STATE_UPDATE (0x2749) Length: 68
> HasLogData: False
> Size: 63
> HitInfo: HITINFO_AFFECTS_VICTIM, HITINFO_RAGE_GAIN (8388610)
> Attacker Guid: Full: 0x2015A4002025FD800000380000BBD705 Creature/0 R1385/S56 Map: 1 Entry: 38902 Low: 12310277
> Target Guid: Full: 0x2015A4002025FC0000003800003EEE45 Creature/0 R1385/S56 Map: 1 Entry: 38896 Low: 4124229
> Damage: 1027
> OverDamage: -1
> HasSubDmg: true
> SchoolMask: 1
> Float Damage: 1027
> Int Damage: 1027
> VictimState: VICTIMSTATE_NORMAL (1)
> Unk Attacker State 0: -1
> Melee Spell ID: 0 (0)
> Rage Gained: 35

After reaching health percentage:

> ServerToClient: SMSG_ATTACKER_STATE_UPDATE (0x2749) Length: 73
> HasLogData: False
> Size: 68
> HitInfo: HITINFO_AFFECTS_VICTIM, HITINFO_UNK10, HITINFO_UNK11, HITINFO_UNK12, HITINFO_RAGE_GAIN, HITINFO_FAKE_DAMAGE (25172994)
> Attacker Guid: Full: 0x2015A4002025FD8000003800033BD6FF Creature/0 R1385/S56 Map: 1 Entry: 38902 Low: 54253311
> Target Guid: Full: 0x2015A4002025FC0000003800003EEE21 Creature/0 R1385/S56 Map: 1 Entry: 38896 Low: 4124193
> Damage: 0
> OverDamage: -1
> HasSubDmg: true
> SchoolMask: 1
> Float Damage: 0
> Int Damage: 0
> VictimState: VICTIMSTATE_NORMAL (1)
> Unk Attacker State 0: 0
> Melee Spell ID: 0 (0)
> Rage Gained: 35
> Unk Float: 0.05385441

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -637,6 +637,52 @@ void ObjectMgr::LoadCreatureTemplateAddons()
...
+        }
+
+        if (!sObjectMgr->GetCreatureTemplate(victim))
+        {
+            TC_LOG_ERROR("sql.sql", "Creature template (Entry: %u) does not exist but has a record in `creature_sparring_template`", entry);
+            continue;
+        }
+
+        if (healthPct > 100.0f)
+            healthPct = 100.0f;
```
jackpoz: it would be worth to show an error about wrong values

Ovahlord:
So one of the HITINFO_UNK10, HITINFO_UNK11, HITINFO_UNK12, flags seems to be the trigger for keeping the health clientside while it seems correct to nullify the damage on periodic aura damage

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -704,6 +704,15 @@ void Unit::DealDamageMods(Unit const* victim, uint32 &damage, uint32* absorb) co
 
 uint32 Unit::DealDamage(Unit* victim, uint32 damage, CleanDamage const* cleanDamage, DamageEffectType damagetype, SpellSchoolMask damageSchoolMask, SpellInfo const* spellProto, bool durabilityLoss)
 {
+    // Sparring Checks
+    if (Creature* me = ToCreature())
+        if (Creature* target = victim->ToCreature())
+            if (me->CanSparrWith(target))
+                if (damage >= victim->GetHealth) // safety check to prevent creatures getting accidentally killed when the percentage is small
```
tkrokli:
```
src/server/game/Entities/Unit/Unit.cpp:711:39: fatal error:
 reference to non-static member function must be called;
 did you mean to call it with no arguments?
                if (damage >= victim->GetHealth) // safety check to prevent creatures getting accidentally killed when the percentage is small
                              ~~~~~~~~^~~~~~~~~
                                               ()
```

Ovahlord: I really dont know what is happening there.... looks like missing header includes but I dont wanna pollute the includes again

soulfrost: GetHealth  -> GetHealth()

Faq:
Isn't this used also for 335, for example fighting npcs at the outland gate?

Ovahlord:
If they are not dying and keep their health after a hitting at specific mark, yes.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -704,6 +704,15 @@ void Unit::DealDamageMods(Unit const* victim, uint32 &damage, uint32* absorb) co
 
 uint32 Unit::DealDamage(Unit* victim, uint32 damage, CleanDamage const* cleanDamage, DamageEffectType damagetype, SpellSchoolMask damageSchoolMask, SpellInfo const* spellProto, bool durabilityLoss)
 {
+    // Sparring Checks
+    if (Creature* me = ToCreature())
+        if (Creature* target = victim->ToCreature())
+            if (me->CanSparrWith(target))
+                if (damage >= victim->GetHealth()) // safety check to prevent creatures getting accidentally killed when the percentage is small
+                    damage = victim->GetHealth() - 1;
+                else if (target->GetHealthPct() <= me->GetSparringHealthLimitPctFor(target))
```
tkrokli:
```
src/server/game/Entities/Unit/Unit.cpp:713:17:
 fatal error: add explicit braces to avoid dangling else [-Wdangling-else]

                else if (target->GetHealthPct() <= me->GetSparringHealthLimitPctFor(target))
                ^
```
I presume this suggests braces around the lines 711-714 in your `Unit.cpp`

Ovahlord:
Sooo let's get this one here done. Aside from the travis warnings is there something else to say about it?

src/server/game/Entities/Unit/Unit.cpp
tkrokli: ... or maybe braces are needed after the `else` (uncertain about this choice)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -704,6 +704,15 @@ void Unit::DealDamageMods(Unit const* victim, uint32 &damage, uint32* absorb) co
 
 uint32 Unit::DealDamage(Unit* victim, uint32 damage, CleanDamage const* cleanDamage, DamageEffectType damagetype, SpellSchoolMask damageSchoolMask, SpellInfo const* spellProto, bool durabilityLoss)
 {
+    // Sparring Checks
+    if (Creature* me = ToCreature())
+        if (Creature* target = victim->ToCreature())
+            if (me->CanSparrWith(target))
```
Jildor: If you put else, you need add braces here

tkrokli: OK, so "around the lines 711-714" like I mentioned in my comment above, then.

Ovahlord:
So what is going on? I have been testing those changes so far and it appears that everything works fine so far ingame. So how shall we proceed?

tkrokli:
We need to wait for the main devs to merge the PR or comment if anything else needs more work.

Aokromes:
Conflicting files
src/server/game/Spells/Spell.cpp

tkrokli:
Weird conflict, it does not list any output from master conflicting with the PR content:
```cpp
<<<<<<< sparringPrototype

        m_damage = damageInfo.damage;

        caster->DealSpellDamage(&damageInfo, true);

        // Sparring Checks
        if (Creature* me = caster->ToCreature())
            if (Creature* target = damageInfo.target->ToCreature())
                if (me->CanSparrWith(target))
                    if (target->GetHealthPct() <= me->GetSparringHealthLimitPctFor(target))
                        damageInfo.damage = 0;

        // Send log damage message to client
        caster->SendSpellNonMeleeDamageLog(&damageInfo);
=======
>>>>>>> master
```
I thought this would be seen as a normal PR insert?

src/server/game/Entities/Creature/Creature.h
```diff
@@ -219,6 +219,10 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         Player* SelectNearestPlayer(float distance = 0) const;
         Unit* SelectNearestHostileUnitInAggroRange(bool useLOS = false) const;
 
+        CreatureSparring const* GetSparringData(uint32 attackerEntry, uint32 victimEntry) const;
+        bool CanSparrWith(Creature* victim) const;
```
Treeston: nitpick, but this should `CanSparWith`

tkrokli:
@Ovahlord : if you rebase your PR on the current `master` branch (140 commits behind), the conflict may go away (even if there are no visible conflicting lines in your base source).

Ovahlord:
I have updated the PR to the lates branch, squashed all commits together. Should be all fine now.

Killyana:
Closes #18576

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -1974,6 +1993,14 @@ void Unit::AttackerStateUpdate(Unit* victim, WeaponAttackType attType, bool extr
         CalculateMeleeDamage(victim, 0, &damageInfo, attType);
         // Send log damage message to client
         DealDamageMods(victim, damageInfo.damage, &damageInfo.absorb);
+
+        // Sparring Checks
+        if (Creature* me = ToCreature())
+            if (Creature* target = victim->ToCreature())
+                if (me->CanSparWith(target))
+                    if (target->GetHealthPct() <= me->GetSparringHealthLimitPctFor(target))
+                        damageInfo.HitInfo |= HITINFO_FAKE_DAMAGE;
```
ariel-: Instead of modifying Unit::AttackerStateUpdate use Unit::FakeAttackerStateUpdate which does just this (sparring vs other creature should be checked elsewhere, for example at AI level)

src/server/game/Spells/Spell.cpp
```diff
@@ -2531,6 +2531,13 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
 
             caster->DealSpellDamage(&damageInfo, true);
 
+            // Sparring Checks
```
ariel-: DealSpellDamage calls DealDamage which you already modified with this same check, double checking here is unnecessary

Ovahlord: not exactly. I am modifiying the damageInfo damage here to make the spell damage log packet also send zero damage because the spell log damage packet is responsible to handle the clientside health damage.
So if I don't do this here, the core will indeed have the correct health, while the spell damage will still deal damage clientside.

src/server/game/Entities/Unit/Unit.cpp
Ovahlord: No, it doesn't. The fake attacker state sends a hacky packet. According to sniffs the damage etc. does not get modified, only this hitinfo flag gets added. The client will know then to not use the damage. The rest must be handled by the core as I did.

ariel-: Then feel free to add damage information to fake_damage but iirc I saw sniffs with hitinfo fake and 0 damage

Treeston: At that point, would there be sufficient difference between `AttackerStateUpdate` and `FakeAttackerStateUpdate` to support keeping them separate?

I am very much not fond of the code duplication involved here.

Ovahlord: I will review this, when I am back home. in two weeks. I will probably add more fields to damageinfo so we dont have to call the checks multiple times. This might make the old fakedamage features obsolete so I can drop them afterwards

Issue #20504, opened 2017-10-01
[3.3.5] Priest - Hymn of Hope


Jildor:
**Description:**

ID - 64901 Hymn of Hope
```Trigger spell (64904) Hymn of Hope. Chance = 101```

ID - 64904 Hymn of Hope
```
Effect 0: Id 137 (SPELL_EFFECT_ENERGIZE_PCT)
BasePoints = 3
Targets (22, 30) (TARGET_SRC_CASTER, TARGET_UNIT_SRC_AREA_ALLY)
Radius (Id 23) 40,00

Effect 1: Id 6 (SPELL_EFFECT_APPLY_AURA)
BasePoints = 20
Targets (22, 30) (TARGET_SRC_CASTER, TARGET_UNIT_SRC_AREA_ALLY)
Aura Id 35 (SPELL_AURA_MOD_INCREASE_ENERGY), value = 20, misc = 0 (0), miscB = 0, periodic = 0
Radius (Id 23) 40,00
```

Possible cause?
https://github.com/TrinityCore/TrinityCore/commit/2a462e93093d177a0e0bedf3e94ac2cf4e66a3c3

**Current behaviour:**

When you use 64901 Hymn of Hope and gain aura 64904 Hymn of Hope, if you cancel 64904 Hymn of Hope, you still having the mana increased

**Expected behaviour:**

You should lost the mana increased.

**Steps to reproduce the problem:**

1. Reduce your mana (by example 1000/6000)
2. Cast 64901 Hymn of Hope
3. When you obtain 64904 Hymn of Hope, remove this aura and see how your mana still (then if you remove all ticks of 64904 Hymn of Hope, you are full restored the mana)

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 4d654f548b29bf61552d9f90da8254ecea91b63e

**TDB version:**  TDB335.63

**Operating system:** Debian

Eliminationzx:
```diff
diff --git a/src/server/game/Spells/Auras/SpellAuraEffects.cpp b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -3743,7 +3743,7 @@ void AuraEffect::HandleAuraModIncreaseEnergyPercent(AuraApplication const* aurAp
     }
 
[...]

PR #21061, opened 2017-12-14
Core/Spells: Rewrite mod energy by pct aura effect

Mihc3:
Separated from: Core/Entities: Creature powers #20981.

**Changes proposed:**
- Correct amount of mana reduced when current max power is already increased by pct (formula in line no. 4047).

- 3.3.5 ✔️
- master ✔️

Build OK.
- [ ] I don't have 3.3.5 to test https://wow.gamepedia.com/Hymn_of_Hope. 
- [x] On 7.x there are only small values, but it works.

- None.

### DIFF
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -4043,17 +4043,18 @@ void AuraEffect::HandleAuraModIncreaseEnergyPercent(AuraApplication const* aurAp
     float pct = float(GetAmount());
 
     // Calculate the change.
-    int32 change = target->GetMaxPower(powerType);
-    ApplyPct(change, apply ? pct : (100 - 100 / (1 + pct / 100))); // On removal (!apply) GetMaxPower is already increased by pct.
+    float change = target->GetMaxPower(powerType);
+    ApplyPercentModFloatVar(amount, pct, apply);
+    change -= target->GetMaxPower(powerType);
 
     if (!apply) // We reduce power before removal of aura effect.
-        target->ModifyPower(powerType, -change);
+        target->ModifyPower(powerType, int32(change));
 
     // The actual aura effect for max power.
     target->HandleStatModifier(unitMod, TOTAL_PCT, pct, apply);
 
     if (apply) // We increase power after application of aura effect.
-        target->ModifyPower(powerType, change);
+        target->ModifyPower(powerType, int32(change));
 }
 
 void AuraEffect::HandleAuraModIncreaseHealthPercent(AuraApplication const* aurApp, uint8 mode, bool apply) const
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -4040,21 +4040,14 @@ void AuraEffect::HandleAuraModIncreaseEnergyPercent(AuraApplication const* aurAp
     Powers powerType = Powers(GetMiscValue());
     UnitMods unitMod = UnitMods(UNIT_MOD_POWER_START + powerType);
 
-    float pct = float(GetAmount());
+    int32 oldPower = target->GetPower(powerType);
+    int32 oldMaxPower = target->GetMaxPower(powerType);
 
-    // Calculate the change.
-    float change = target->GetMaxPower(powerType);
-    ApplyPercentModFloatVar(change, pct, apply);
-    change -= target->GetMaxPower(powerType);
+    target->HandleStatModifier(unitMod, TOTAL_PCT, float(GetAmount()), apply);
 
-    if (!apply) // We reduce power before removal of aura effect.
-        target->ModifyPower(powerType, int32(change));
-
-    // The actual aura effect for max power.
-    target->HandleStatModifier(unitMod, TOTAL_PCT, pct, apply);
-
-    if (apply) // We increase power after application of aura effect.
-        target->ModifyPower(powerType, int32(change));
+    int32 change = target->GetMaxPower(powerType) - oldMaxPower;
+    change = (oldPower + change) - target->GetPower(powerType);
+    target->ModifyPower(powerType, change);
 }
 
 void AuraEffect::HandleAuraModIncreaseHealthPercent(AuraApplication const* aurApp, uint8 mode, bool apply) const
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -4037,28 +4037,23 @@ void AuraEffect::HandleAuraModIncreaseEnergyPercent(AuraApplication const* aurAp
         return;
 
     Unit* target = aurApp->GetTarget();
-
     Powers powerType = Powers(GetMiscValue());
-    // do not check power type, we can always modify the maximum
-    // as the client will not see any difference
-    // also, placing conditions that may change during the aura duration
-    // inside effect handlers is not a good idea
-    //if (int32(powerType) != GetMiscValue())
-    //    return;
-
     UnitMods unitMod = UnitMods(UNIT_MOD_POWER_START + powerType);
-    float amount = float(GetAmount());
 
-    if (apply)
-    {
-        target->HandleStatModifier(unitMod, TOTAL_PCT, amount, apply);
-        target->ModifyPowerPct(powerType, amount, apply);
-    }
-    else
-    {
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -4037,28 +4037,23 @@ void AuraEffect::HandleAuraModIncreaseEnergyPercent(AuraApplication const* aurAp
...
-    else
-    {
-        target->ModifyPowerPct(powerType, amount, apply);
-        target->HandleStatModifier(unitMod, TOTAL_PCT, amount, apply);
-    }
+    float pct = float(GetAmount());
+
+    // Calculate the change.
+    int32 change = target->GetMaxPower(powerType);
+    ApplyPct(change, apply ? pct : (100 - 100 / (1 + pct / 100))); // On removal (!apply) GetMaxPower is already increased by pct.
```
joschiwald: what does this calculation change?

Mihapro: increases pct for pct 😃 because base (max power, line above) is already increased by pct

joschiwald: can you explain that is clearlier

Mihapro: Imagine u have +30% max mana, lets say its 13000. Old code would reduce power by 30% of current max (before effect is removed), thats 13000*0,30 = 39k, correct should be 30k. I hope its clear enough :P

Mihapro: Sorry, 3,9k, correct should be 3k*

joschiwald: in your example i get a change of -28.8k

Mihapro: https://www.onlinegdb.com/r1kMH0GzM

Mihapro: Another valid formula is `pct * 100 / (pct + 100)` - same as `100 - 100 / (1 + pct / 100)` ... But then it depends on how many times we want to use pct in a formula.

joschiwald: i still don't get what was wrong with `ModifyPowerPct`, if there is something wrong, `ApplyPercentModFloatVar` must be wrong too

Mihapro: there's nothing wrong with `ModifyPowerPct`, it's wrong that it reduces max by pct: reducing 13000 for 30% should get us back to 10000, but 13000 * 0.30 =/= 3000 - because we have to reduce for 3000, the same amount of power we gave as we increased from 10000 to 13000 - 3000! 😄 instead that 3000 is alot bigger, 3900.

Another example from 3.3.5:
At 30k mana, priest casting hymn of hope (I believe it's 30% for the priest, 15% for members, but it doesn't matter) would get **9k max mana + 9k mana**, but at the end of effect he would lose **11.7k mana + 9k max mana**. See what's wrong?

Warpten: To clarify : the problem here (talking out of my ass, didn't check functions) is that in unapply the already increased pool was supposedly considered to be 100%. Taking 30% away from 100% is not the same as taking 30% from 130%. Functions themselves aren't wrong, usage was (?).

Mihapro:
Good, correct logic 👍

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -4037,28 +4037,24 @@ void AuraEffect::HandleAuraModIncreaseEnergyPercent(AuraApplication const* aurAp
...
-    else
-    {
-        target->ModifyPowerPct(powerType, amount, apply);
-        target->HandleStatModifier(unitMod, TOTAL_PCT, amount, apply);
-    }
+    float pct = float(GetAmount());
+
+    // Calculate the change.
+    float change = target->GetMaxPower(powerType);
+    ApplyPercentModFloatVar(change, pct, apply);
```
ariel-: Note in 3.3.5 I deleted most uses of ApplyPercentModFloatVar due to it losing information about base amount, also not blizzlike (ie a -100 aura reduces amount to 1 instead of 0 because of division by zero issues)

Check c69a7d1223b2ce108a221e5cc70f76cd80ae6675 if you want to see my approach

Mihapro: Also I think @joschiwald ruined it again - we shouldn't apply pct directly to increased max power ...

Mihapro: Example: maxpower = 13.000; pct = 30; apply negative pct to it - 13.000 * -0.3 => 3.900 again some value we can't use.

ariel-: No

```c++
void ApplyPercentModFloatVar(float& var, float val, bool apply)
{
    var *= (apply ? (100.0f + val) / 100.0f : 100.0f / (100.0f + val));
}
```

var = 13000, val = -30, apply = true ->
var = 13000 * (100 - 30) / 100 = 13000 * 0.7 = 9100

Mihapro: continue: 9100 - 13000 to get the change = 3900
you lose mana buff, your max mana falls from 13.000 to 10.000, but u lose 3900 mana, not 3000..... :/

Mihapro: ... or in another way, would you be satisfied with 9100 / 10000 mana after the effect ends?

joschiwald:
Wtf is going on here, tomorrow I'll explain it to you

ariel-:
Cherry-pick 2a462e93093d177a0e0bedf3e94ac2cf4e66a3c3?

src/server/game/Spells/Auras/SpellAuraEffects.cpp
joschiwald: @Mihapro: i have no idea what are you trying to calculate here, just see my example https://ideone.com/vhcnmS

Mihapro: Your last calculation is exactly what I am trying to explain... you got 9100 mana instead of 10000 when doing (13000, -30.f)

joschiwald:
@ariel-: no, your commit removed the whole current power handling, and the other commit just restored debuff handling, so hymn of hope is still broken, see #20504

it is always great to have so big differences of 14 months between master and 335...............

src/server/game/Spells/Auras/SpellAuraEffects.cpp
joschiwald: and that is correct in apply case of -30%, 13000 is your base and not 10000

Mihapro:
3.3.5 code: `// do not reduce current power on buff removal (Hymn of Hope`
`Treeston: Current power should remain increased when a +%mana effect is removed (that's intentional).`
Btw @Treeston are you 100% sure about this? I'll try to find a video where blizz takes current mana back when Hymn of Hope is removed.

Mihapro:
https://www.youtube.com/watch?v=3RMVAN-w1Ek&t=269s
At 4:45 it is seen that mana drops by 5.5k

Treeston:
I am 99.9% certain it worked that way as of Cataclysm. I agree with your video showing the opposite on 3.3 - adjust to match.

Mihapro:
"Was a thing" - more like exploit? 🤣 Then for sure we have to take back power/health on unapply...

Treeston:
No. If it worked on retail 3.3.5, it should work on TC 3.3.5. It is not our job to re-design the game.

joschiwald:
now it should be correct

Mihapro:
I tested it, calculation is now correct. https://onlinegdb.com/SJyUF-4Qz ✔️ 

And it's the same to my initial commit, just that now it's a similar mess to existing code in 3.x. I made this PR with nice comments and better logic to handle power unapply before the actual effect - and there were comments to make it easy to read... But vars to remember old power and shit, meh..

joschiwald:
due to rounding issues are there is no other way around, even your first aproach is more logic, now this code is compatible with 3.3.5

Faq:
ready for merge?

joschiwald:
i'm cherry-picking fix from 3.3.5

PR #21064, opened 2017-12-15
Core/Player: Implement aura interrupt 14 (AURA_INTERRUPT_FLAG_CHANGE_SPEC)

Mihc3:
**Changes proposed:**

- Removes some auras and talent buffs when changing specialization
(AuraInterruptFlags2 - 0x00004000).

- 3.3.5 ✔️
- master ✔️

Build OK. Test OK.

None.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -26519,6 +26519,8 @@ void Player::ActivateTalentGroup(ChrSpecializationEntry const* spec)
     ExitVehicle();
     RemoveAllControlled();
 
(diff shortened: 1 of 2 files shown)


ariel-:
Doesn't seem to be the case in 3.3.5 (no spec related spells using that flag):

Shauren:
Flag is correct but implementation couldn't have been more wrong (throw it all away)

Two options here
* Create new enum for SpellInterruptsEntry::AuraInterruptFlags[1]
* Extend SpellInfo::AuraInterruptFlags, SpellInfo::ChannelInterruptFlags and Unit::m_interruptMask to be a two element array
* Update code using m_interruptMask to allow more interrupt flags

Issue #18661, opened 2016-12-29
pool ram usage


Rochet2:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:**
I inspected some of the RAM usage on startup and noticed that a huge amount was consumed by poolmgr. Like 1/10 of memory was consumed by it.
http://i.imgur.com/YKnOWeD.png

Further inspection revealed that most of the reserved memory is not probably used for anything.
Some small testing showed that using map reduces memory usage to maybe 1/10 of original and startup time was reduced by 30 seconds on debug mode while having the profiling tools on (so on release maybe some seconds? just a guess)

It was also seen that some parts of the core seem to loop over the whole structures when saving.
Or looping from 0 to max entry. (about 200000 iterations)
changing the structure may make these loops relatively more faster as there are just some 4000 entries instead of 200000.

The root cause of the issue is here:
https://github.com/TrinityCore/TrinityCore/blob/bf33159a7009f64a78cf2a1309eb5182fcd3f7e3/src/server/game/Pools/PoolMgr.h#L143-L160
https://github.com/TrinityCore/TrinityCore/blob/bf33159a7009f64a78cf2a1309eb5182fcd3f7e3/src/server/game/Pools/PoolMgr.cpp#L560-L571
As seen vectors are used and they are resized to DB max entry value which is 202482.
If DB is looked into there are nowhere near that many entries there.

**Current behaviour:** (Tell us what happens.)
poolmgr consumes high amounts of memory for little benefit.
the "empty values" are not used for anything from what I can see.
vectors are used.
There are "max(entry) from pool_template" elements reserved for the vectors.
There are huge gaps in the entries which causes the waste of memory
Not sure if this is intended, but when checking for pools on load it is only checked that entry < max entry. It is never checked if the entries used actually exist in pool_template table which means there can be pool IDs used that dont exist and dont produce errors.

**Expected behaviour:** (Tell us what should happen instead.)
poolmgr should use just what it uses or needs for effective use.
the memory used should have purpose.
maybe maps should be used because of the huge gaps.
properly check pool entries on load.

**Steps to reproduce the problem:**

0. how I inspected ram usage:
1. Start visual studio (2015 enterprise) as if you were compiling
2. Clean solution (this is important because at least for me VS crashes if I use rebuilds and diagnostics)
3. Build solution
[...]

PR #21075, opened 2017-12-17
Core/Pools: fix pool memory usage

Rochet2:
**Changes proposed:**

-  Change std::vector to std::unordered_map to save memory due to only sparse keys used from the container
- This should slightly speed up startup
- This should slightly speed up saving weekly and daily quests
- This should reduce ram usage quite significantly (10% saved after startup)
- Pool loading in startup can look slower in the log messages due to unordered map find and space allocated in pool loading rather than initialization
- This may slightly slow the pool system since vector random access changed to unordered map find
- Remove max_pool_id since it is unused
- Add some startup checks that log pool IDs that are not in pool_template (before only < maxpoolid was checked). **The ones that do not exist in pool template are skipped**
- **Assumption**: All pool_id are supposed to be in pool_template
- One of the alternative fixes to changing code was to fix the DB data instead of core code. Read related issue comments: https://github.com/TrinityCore/TrinityCore/issues/18661#issuecomment-269690058

If the DB approach is preferred, then some guidance on how the data transformation could be done would be nice.
For example should all existing rows be deleted and new data be inserted, or should we just use SET statements to set the pool ids to new ones? What about custom data?

One idea could be to make a script that takes all custom data, puts it to the side, reorders TC data by pool_id, appends custom data on top of that (with reorder).

- [x] 3.3.5
- [x] master

**Issues addressed:**
Closes https://github.com/TrinityCore/TrinityCore/issues/18661

- Compiles.
- Logged ingame, looted gameobjects and herbs and waited for them to respawn. (silverpine forest, nagrand, hellfire peninsula)
- Checked that startup times after change were slightly faster. (this can vary significantly since there can be even 30 seconds of difference between two startups of same server) Here is a before and after of the pool loading: https://i.imgur.com/ZoSSodh.png and it disregards initialization time of pools for before.
- Checked that startup errors popped up about entries that are not in pool_template but are used in other pool tables.
- Checked that RAM consumed was around 1833 MB before and around 1649 MB after change. (multiple startups of the server)

- [x] Choose between DB and core edits.

### DIFF
--- a/src/server/game/Pools/PoolMgr.h
+++ b/src/server/game/Pools/PoolMgr.h
@@ -141,11 +141,11 @@ class TC_GAME_API PoolMgr
         template<typename T>
         void SpawnPool(uint32 pool_id, uint32 db_guid_or_pool_id);
 
-        typedef std::unordered_map<uint32, PoolTemplateData>       PoolTemplateDataMap;
-        typedef std::unordered_map<uint32, PoolGroup<Creature> >   PoolGroupCreatureMap;
-        typedef std::unordered_map<uint32, PoolGroup<GameObject> > PoolGroupGameObjectMap;
-        typedef std::unordered_map<uint32, PoolGroup<Pool> >       PoolGroupPoolMap;
-        typedef std::unordered_map<uint32, PoolGroup<Quest> >      PoolGroupQuestMap;
+        typedef std::unordered_map<uint32, PoolTemplateData>      PoolTemplateDataMap;
+        typedef std::unordered_map<uint32, PoolGroup<Creature>>   PoolGroupCreatureMap;
+        typedef std::unordered_map<uint32, PoolGroup<GameObject>> PoolGroupGameObjectMap;
+        typedef std::unordered_map<uint32, PoolGroup<Pool>>       PoolGroupPoolMap;
+        typedef std::unordered_map<uint32, PoolGroup<Quest>>      PoolGroupQuestMap;
         typedef std::pair<uint32, uint32>           SearchPair;
         typedef std::map<uint32, uint32>            SearchMap;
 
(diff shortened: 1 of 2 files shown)


jackpoz:
is the increased ram usage caused by those high entry values in the database ? could the ram usage issue be fixed by compacting the ids in the db ?

Shauren:
Yes but tbh using a plain vector for this is a terrible idea when ids don't have to be contiguous

Takenbacon:
@Shauren It's always a trade off with performance vs ram usage, how much slower is pooling now with having map lookups?

src/server/game/Pools/PoolMgr.h
```diff
@@ -141,12 +141,11 @@ class TC_GAME_API PoolMgr
...
         void SpawnPool(uint32 pool_id, uint32 db_guid_or_pool_id);
 
-        uint32 max_pool_id;
-        typedef std::vector<PoolTemplateData>       PoolTemplateDataMap;
-        typedef std::vector<PoolGroup<Creature> >   PoolGroupCreatureMap;
-        typedef std::vector<PoolGroup<GameObject> > PoolGroupGameObjectMap;
-        typedef std::vector<PoolGroup<Pool> >       PoolGroupPoolMap;
-        typedef std::vector<PoolGroup<Quest> >      PoolGroupQuestMap;
+        typedef std::unordered_map<uint32, PoolTemplateData>       PoolTemplateDataMap;
+        typedef std::unordered_map<uint32, PoolGroup<Creature> >   PoolGroupCreatureMap;
```
ariel-: Since this is new code and we support c++11 there is no need to leave a space between template brackets

Rochet2:
@Takenbacon Vector and map access time is the same when doing 1-10 accesses in succession.
100000000 accesses takes 0ms with vector and 1 second with map, so over the course of maybe 20 days you could lose one second if the pool system maps are actively called (like 50 times every second).


I did not make any exhaustive performance analysis on the pooling system, but based on this I would say its not really slower.

Shauren:
The point is, you don't iterate over pool vectors (which is what would give it the largest advantage)

jackpoz:
so let's merge it ?

jackpoz:
@Rochet2 please don't go on vacation this Xmas in case some issues arise :)

PR #21077, opened 2017-12-17
CI Master #21033

Naios:
#21033 -> master


Shauren:
VS build failed in 'proto' project which previously had all warnings disabled because it contains only generated code
```
C:\projects\trinitycore\dep\protobuf\src\google/protobuf/wire_format_lite_inl.h(383): error C2220: warning treated as error - no 'object' file generated [C:\projects\trinitycore\build\src\server\proto\proto.vcxproj]
  C:\projects\trinitycore\dep\protobuf\src\google/protobuf/wire_format_lite_inl.h(413): note: see reference to function template instantiation 'bool google::protobuf::internal::WireFormatLite::ReadPackedFixedSizePrimitive<google::protobuf::uint32,google::protobuf::internal::WireFormatLite::TYPE_FIXED32>(google::protobuf::io::CodedInputStream *,google::protobuf::RepeatedField<google::protobuf::uint32> *)' being compiled
C:\projects\trinitycore\dep\protobuf\src\google/protobuf/wire_format_lite_inl.h(383): warning C4018: '<': signed/unsigned mismatch [C:\projects\trinitycore\build\src\server\proto\proto.vcxproj]
```

Naios:
I was working on this today. The current issue is that I'm not capable of disable the warning level inherited by the common library. Probably we fully have to reorder the warning interfaces fdor the core targets.

@Shauren I'm still wondering why there are some warnings present in protobuffers, were those fixed in later versions of the library?

Naios:
@Shauren I encountered the issue that MSVC did yield warnings although `\W0` was set correctly.
Thus I enabled `_SCL_SECURE_NO_WARNINGS` pivately for protobuf projects, see 95a837717be55a72a1ebc9c00ce3a41bf5df76c8.

The port should build now correctly for the master branch.

Issue #19744, opened 2017-05-10
[3.3.5] Time Watchers are not in caverns of time.


dikkedeur:
**Description:** The Npc Time Watcher that are normally in the caverns of time, are after a server restart not in the caverns of time anymore, but on the surface in tanaris.
After some time they drop through the ground, so it takes roughly 15 minutes or longer before they are not on the surface anymore

**Current behaviour:** see the describtion.

Tell us what happens.:
also mentioned in the describtion, after server restart those time watchers are not at the right place.

**Expected behaviour:**they should be in the caverns and not on the surface.

Tell us what should happen instead.they should be in the right spot.

**Steps to reproduce the problem:**

Step 1 npc id= 19918
2. restart server.. go to tanaris around the caverns of time, you will see those time watchers above the ground instead of the cavern.
3. Step 3

![schermafbeelding 2017-05-10 om 08 35 28](https://cloud.githubusercontent.com/assets/11842306/25886192/c8e3e44e-355c-11e7-9fc8-27872df3f75f.png)

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. e036d2232dae+ 2017-05-10 02:41:50 +0200 (3.3.5 branch) (Unix, Debug, Static) (worldserver-daemon)

**TDB version:**  TDB 335.63

**Operating system:**ubuntu server 16.04 LTS

Killyana:
You're using a different SAI that despawn the npcs, this issue happen only after a respawn, and it's related to the inhabittype.

Killyana:
Looks like mmaps issue.

Ovahlord:
Probably a mmaps issue. Cuz the ground finder is normalizing the Z-position from a higher point and goes down until it finds a ground. Since the cave ground is below this ground it is most likely that the Z position stops at the surface.

Killyana:
I added CREATURE_FLAG_EXTRA_IGNORE_PATHFINDING to this npcs but the issue still happen, also I tested with mmap.enablePathFinding = 0 and nothing changed.

Golrag:
In ```RandomMovementGenerator<Creature>::SetRandomLocation(Creature* owner)``` ```owner->MovePositionToFirstCollision(position, distance, angle);``` Gets called. Then ```destz = NormalizeZforCollision(this, destx, desty, pos.GetPositionZ());``` is called which results in unintend destz
Note:
```

ccrs:
yeap, thats the reason

[...]

Issue #20028, opened 2017-07-14
[335] Vmaps latests changes and Wintergrasp


Jildor:
**Description:**

With latests changes of Vmaps, Wintergrasp have stranges bugs
d2a30fef1371cf904e56f96042d686c9da439175
a409287feffae22b9690cfc58d6e3e877fe39e28
cfb0f9fb19c03677b127f09407a6bd00bfefbe55
a2c123b5296b2723ae2a0211e7aaafe37fb9246d
f6c849729b27b77228704b595de3adaf24da2c10
c35793941b3aae9b54dce92389aa528d40050754

I think the causant is:
f6c849729b27b77228704b595de3adaf24da2c10
( I can't comment in the commit because is closed https://github.com/TrinityCore/TrinityCore/commit/f6c849729b27b77228704b595de3adaf24da2c10#commitcomment-22449905 )

I revert the commits and extract Vmaps and works fine

Wintergrasp vehicles count don't work
Wintergrasp Capture Points in Workshop don't work
...

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  445c5a00e8734f410ce1abc222ed847ead6741ed

**TDB version:**  TDB335.63

**Operating system:**  Debian

Jildor:
@Treeston can you take a look here?

Treeston:
No idea - I have no clue how WG code works, and other places that use the same core logic work fine (EotS capture points).

Jildor:
I'm not sure, but in Wintergrasp the phasemask changes, then can be possible worldobjects are in different phase with this change?

Aokromes:
wintergrasp needs all spawns moved into db to work properly.

Kittnz:
Dynamic spawns need to be implemented first, then move everything to db..

Jildor:
@Aokromes I think this not related to this

I'm saying the worldobjects with this: https://github.com/TrinityCore/TrinityCore/commit/f6c849729b27b77228704b595de3adaf24da2c10#diff-29afb15ead2a8204cac07d94f074f260R2561
now gets phasemask (I'm not sure if before too or not xD)

Then, in Wintergrasp, the workshops zones are differents phases and change with defender team

Jildor:
Hellfire Hellfire outdoor capture points seems don't work too :(

tkrokli:
[...]

PR #21125, opened 2017-12-26
Core/Entities: Fix some weird movement due to los issues

### LINKED ISSUE #21121: Tools/vmap: AoE randomly hits out of LoS in caves
**Description:**

DeathKnight AoE spells randomy does not work in caves. 

**Current behaviour:**

DeathKnight AoE spells randomy does not work in caves. Video of this bug: https://youtu.be/p_qnxZ0Pt9g

`vmap.enableLOS = 0` fix problem.

**Expected behaviour:**

CHANGEME Tell us what should happen instead.

1. CHANGEME Step 1 include entries of affected creatures / items / quests with a link to the relevant wowhead page.  
2. Step 2
3. Step 3

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  f0614b57a59a8525a6634e1465a83ef8fa27839e

Golrag:
**Changes proposed:**

- Made LoS check use collisionHeight instead of midsection. Value was too low.
- Gnomes will now have a breath bar more quickly than for example a tauren.
- Changes have been made to checking for ground z as well, some tweeking might be needed but removed most of the scattered +2.0f/+0.5f/we. Which I think is better now.

- 3.3.5
- master

**Issues addressed:** Closes #21121  #19744

**Tests performed:**
- Builds (local)
- Ran around  in the world to find some issues but didn't really notice any.
- Checked out the new beasts of northrend fight, no issues.
- The bronze drakes didn't pop out of the ground in tanaris.

- [ ] There is a place in dun morogh up a hill with trolls where a troll sometimes goes undermap when chasing you, will looking into that.
  - Not sure if my changes are related to this. -5067.1655237 -245.251724 442.975647 and -5070.35791 -245.135208 442.975647 are in LoS with each other, while they most likely shouldn't.

- Still not very sure what to do when collisionheight is 0, added a default one for now.

Any help & feedback is welcome, I feel like I might have removed too much of those 0.5f's...

Happy holidays :)

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2889,7 +2889,7 @@ void Creature::UpdateMovementFlags()
     // Set the movement flags if the creature is in that mode. (Only fly if actually in air, only swim if in water, etc)
     float ground = GetFloorZ();
 
-    bool isInAir = (G3D::fuzzyGt(GetPositionZMinusOffset(), ground) || G3D::fuzzyLt(GetPositionZMinusOffset(), ground)); // Can be underground too, prevent the falling
+    bool isInAir = (G3D::fuzzyGt(GetPositionZMinusOffset(), ground + GROUND_HEIGHT_TOLERANCE) || G3D::fuzzyLt(GetPositionZMinusOffset(), ground - GROUND_HEIGHT_TOLERANCE)); // Can be underground too, prevent the falling
 
     if (GetCreatureTemplate()->InhabitType & INHABIT_AIR && isInAir && !IsFalling())
     {
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -1036,7 +1036,6 @@ bool Creature::Create(ObjectGuid::LowType guidlow, Map* map, uint32 phaseMask, u
         TC_LOG_ERROR("entities.unit", "Creature::Create(): given coordinates for creature (guidlow %d, entry %d) are not valid (X: %f, Y: %f, Z: %f, O: %f)", guidlow, entry, pos.GetPositionX(), pos.GetPositionY(), pos.GetPositionZ(), pos.GetOrientation());
         return false;
     }
-    UpdatePositionData();
 
     // Allow players to see those units while dead, do it here (mayby altered by addon auras)
     if (cinfo->type_flags & CREATURE_TYPE_FLAG_GHOST_VISIBLE)
@@ -1044,6 +1043,8 @@ bool Creature::Create(ObjectGuid::LowType guidlow, Map* map, uint32 phaseMask, u
 
     if (!CreateFromProto(guidlow, entry, data, vehId))
         return false;
+
+    UpdatePositionData();
 
     if (GetCreatureTemplate()->flags_extra & CREATURE_FLAG_EXTRA_DUNGEON_BOSS && map->IsDungeon())
         m_respawnDelay = 0; // special value, prevents respawn for dungeon bosses unless overridden
@@ -2888,7 +2889,7 @@ void Creature::UpdateMovementFlags()
     // Set the movement flags if the creature is in that mode. (Only fly if actually in air, only swim if in water, etc)
     float ground = GetFloorZ();
 
-    bool isInAir = (G3D::fuzzyGt(GetPositionZMinusOffset(), ground + 0.05f) || G3D::fuzzyLt(GetPositionZMinusOffset(), ground - 0.05f)); // Can be underground too, prevent the falling
+    bool isInAir = (G3D::fuzzyGt(GetPositionZMinusOffset(), ground) || G3D::fuzzyLt(GetPositionZMinusOffset(), ground)); // Can be underground too, prevent the falling
 
     if (GetCreatureTemplate()->InhabitType & INHABIT_AIR && isInAir && !IsFalling())
     {
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -1044,8 +1044,6 @@ bool Creature::Create(ObjectGuid::LowType guidlow, Map* map, uint32 phaseMask, u
     if (!CreateFromProto(guidlow, entry, data, vehId))
         return false;
 
-    UpdatePositionData();
-
     if (GetCreatureTemplate()->flags_extra & CREATURE_FLAG_EXTRA_DUNGEON_BOSS && map->IsDungeon())
         m_respawnDelay = 0; // special value, prevents respawn for dungeon bosses unless overridden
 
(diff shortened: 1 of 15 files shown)


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1045,6 +1044,8 @@ bool Creature::Create(ObjectGuid::LowType guidlow, Map* map, uint32 phaseMask, u
     if (!CreateFromProto(guidlow, entry, data, vehId))
         return false;
 
+    UpdatePositionData();
```
jackpoz: what is this change for ?

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2888,7 +2889,7 @@ void Creature::UpdateMovementFlags()
     // Set the movement flags if the creature is in that mode. (Only fly if actually in air, only swim if in water, etc)
     float ground = GetFloorZ();
 
-    bool isInAir = (G3D::fuzzyGt(GetPositionZMinusOffset(), ground + 0.05f) || G3D::fuzzyLt(GetPositionZMinusOffset(), ground - 0.05f)); // Can be underground too, prevent the falling
+    bool isInAir = (G3D::fuzzyGt(GetPositionZMinusOffset(), ground) || G3D::fuzzyLt(GetPositionZMinusOffset(), ground)); // Can be underground too, prevent the falling
```
jackpoz: what was the purpose of +/- 0.05 fuzzyGt/Lt and why is it removed now ?

src/server/game/Entities/Creature/Creature.cpp
Golrag: If I call it earlier assertion will trigger in GetColllisionHeight, or crash in GetUInt32Value. Can't remember which one it was.

jackpoz:
after this PR I would suggest to find a place for 2.03128f so that it is defined only in a single place (no need to worry about that now, it's more important to fix the LoS issues)

src/server/game/Entities/Creature/Creature.cpp
Shauren: It was definitely intended here, don't want creatures toggling flight/walk like mad based on inaccurancies during spline update

Golrag: Will place it back then, but I would like to have those 0.05f's defined somewhere for consistency, any good place for it ?

Golrag:
Alright, the value is from legion db2's btw. 335 has 2.031 so I chose the legion one.

src/server/game/Entities/Creature/Creature.cpp
Faq: SharedDefines.h is not ok to store such?

Killyana:
It could fix https://github.com/TrinityCore/TrinityCore/issues/19744 also ?

Golrag:
Seems like it

src/server/game/Entities/Creature/Creature.cpp
Golrag: Seems like https://github.com/TrinityCore/TrinityCore/issues/20028 is back now with these changes so, I will maybe revert those changes and let the default param do the job

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1464,6 +1455,15 @@ bool Creature::CreateFromProto(ObjectGuid::LowType guidlow, uint32 entry, Creatu
     if (!UpdateEntry(entry, data))
         return false;
 
+    UpdatePositionData();
+    SetZoneScript();
+    if (GetZoneScript() && data)
+    {
+        entry = GetZoneScript()->GetCreatureEntry(guidlow, data);
```
Shauren: great, you completely broke GetCreatureEntry hook.

Golrag: Will still trigger an assert then, doing it in UpdateEntry doesnt feel right to me as we are not really updating a position. Reverting it and leaving the default collisionheight deal with it sounds like a better option to me

Keader:
Can you check if it fix Killing Spree too?
If target is so close of you, Effect1 is not working
Stop work after 95456ab5d8bd623481d315ae55dfbb44b45ba9f5
if you run a little with target, spell work fine

```
Effect 1: Id 3 (SPELL_EFFECT_DUMMY)
BasePoints = 1
Targets (18, 16) (TARGET_DEST_CASTER, TARGET_UNIT_DEST_AREA_ENEMY)
Radius (Id 13) 10,00
```

Golrag:
Can't reproduce it like in the video :)

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2567,7 +2550,26 @@ float WorldObject::GetFloorZ() const
...
         return m_staticFloorZ;
-    return std::max<float>(m_staticFloorZ, GetMap()->GetGameObjectFloor(GetPhaseMask(), GetPositionX(), GetPositionY(), GetPositionZ()));
+    return std::max<float>(m_staticFloorZ, GetMap()->GetGameObjectFloor(GetPhaseMask(), GetPositionX(), GetPositionY(), GetPositionZ() + GetCollisionHeight()));
+}
+
+float WorldObject::GetMapWaterOrGroundLevel(float x, float y, float z, float* ground/* = nullptr*/) const
+{
+    if (Unit const* unit = ToUnit())
+    {
+        *ground = z;
```
jackpoz: is this some new code that doesn't handle the case of ground = nullptr or was it just copied from somewhere ? I'd prefer to handle that case too to avoid crashes (or ASSERT() is ground is never supposed to be null)

jackpoz:
should we merge this PR and let the users report back any new issue ?

src/server/game/Entities/Object/Object.cpp
Golrag: It should be removed and let Map::GetWaterOrGroundLevel deal with it

Golrag:
Might still be some issues like I described in my PR. I'm still not very sure if my previous PR caused it or this one as ```GetMap()->isInLineOfSight``` with those coordinates returns true.

jackpoz:
Brace yourself

jackpoz:
This PR is most likely causing https://github.com/TrinityCore/TrinityCore/issues/21969#issuecomment-770276557 and https://github.com/TrinityCore/TrinityCore/issues/26024 

1. This PR assumes a lot of height map/vmap functions are used only for LoS but they are actually used also to find the floor Z and using a full collision height might pick the floor above instead
2. A lot of +0.5f/+2.0f were not a default "half collision heigth" but an empyrical delta that worked well enough when taking the Z of mmaps (which was approximated) and trying to match it to the right floor map/vmap Z

I would say full collision height should used only for LoS checks and either half or a fixed amount should be used when getting the floor Z for waypoints/paths/etc.

illfated:
Thank you for the notice. Nice to know. 👍🏻

PR #21160, opened 2018-01-02
Core/Spells: Mana potions in druid forms

DanVS:
**Changes proposed:**
-  Using mana potions in bear or cat druid forms doesn't restore mana.
-  PR fixes this behaviour by adding adding check to the energize function.

- [x] 3.3.5
- [ ] master

**Tests performed:** builds, works in game.

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -1730,7 +1730,7 @@ void Spell::EffectEnergize(SpellEffIndex effIndex)
 
     Powers power = Powers(m_spellInfo->Effects[effIndex].MiscValue);
 
-    if (unitTarget->GetTypeId() == TYPEID_PLAYER && unitTarget->getPowerType() != power && !(m_spellInfo->SpellFamilyName == SPELLFAMILY_POTION) 
+    if (unitTarget->GetTypeId() == TYPEID_PLAYER && unitTarget->getPowerType() != power && m_spellInfo->SpellFamilyName != SPELLFAMILY_POTION
         && !m_spellInfo->HasAttribute(SPELL_ATTR7_CAN_RESTORE_SECONDARY_POWER))
         return;
 
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -1730,7 +1730,7 @@ void Spell::EffectEnergize(SpellEffIndex effIndex)
 
     Powers power = Powers(m_spellInfo->Effects[effIndex].MiscValue);
 
-    if (unitTarget->GetTypeId() == TYPEID_PLAYER && unitTarget->getPowerType() != power && !m_spellInfo->SpellFamilyName == SPELLFAMILY_POTION 
+    if (unitTarget->GetTypeId() == TYPEID_PLAYER && unitTarget->getPowerType() != power && !(m_spellInfo->SpellFamilyName == SPELLFAMILY_POTION) 
         && !m_spellInfo->HasAttribute(SPELL_ATTR7_CAN_RESTORE_SECONDARY_POWER))
         return;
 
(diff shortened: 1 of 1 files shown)


jackpoz:
Could you please test ingame both cases, as in the case where the check is true and returns and the case where the check is false and the function keeps executing ? Also please describe more in detail the tests executed. To me the added check looks wrong.

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -1730,7 +1730,8 @@ void Spell::EffectEnergize(SpellEffIndex effIndex)
 
     Powers power = Powers(m_spellInfo->Effects[effIndex].MiscValue);
 
-    if (unitTarget->GetTypeId() == TYPEID_PLAYER && unitTarget->getPowerType() != power && !m_spellInfo->HasAttribute(SPELL_ATTR7_CAN_RESTORE_SECONDARY_POWER))
+    if (unitTarget->GetTypeId() == TYPEID_PLAYER && unitTarget->getPowerType() != power && !m_spellInfo->SpellFamilyName == SPELLFAMILY_POTION
```
joschiwald: `!m_spellInfo->SpellFamilyName == SPELLFAMILY_POTION` should be `m_spellInfo->SpellFamilyName != SPELLFAMILY_POTION`

jackpoz:
@joschiwald do you mind letting the author of this PR to learn something on his own ? thank you

DanVS:
Tests I made:
1. Druid in bear/cat form or without them using [Major Rejuvenation Potion](http://www.wowhead.com/item=18253);
3. Druid using [Arcane Torrent](http://www.wowhead.com/spell=28730) to restore mana in bear/cat form or without them (spell without attribute).

jackpoz:
could you please check if the SPELL_ATTR7_CAN_RESTORE_SECONDARY_POWER attribute check still works ? to me (and to the automatic build tools) it looks like the code you added is not C++ correct and it sounds strange that ingame everything is still working correctly. I'd suggest to debug the code with a debugger, not just checking if the edge case you are trying to fix has been solved.

DanVS:
Yes, attribute works. I mean [Replenishment](https://dbwotlk.com/?spell=57669) (hasn't attribute) works only on mana, while [Improved Leader of the Pack](https://dbwotlk.com/?spell=34300) (has) is working even if mana is second resource.

tkrokli:
@DanVS : I presume you have looked up the warnings ("errors") in AppVeyor and travis-ci and can deal with them. Otherwise, feel free to ask.

jackpoz:
Using item http://db.vanillagaming.org/?item=9397 in Dire Bear form (POWER_RAGE type) triggers spell http://db.vanillagaming.org/?spell=5405 which has Family SPELLFAMILY_MAGE and doesn't have SPELL_ATTR7_CAN_RESTORE_SECONDARY_POWER but still grants mana.

As suggested above a few times already, please check the build warnings pointed out by Travis and AppVeyor, your code is not C++ correct.

DanVS:
Fixed. I guess Travis fails not because of this commit.

gmillerd:
Odd, its dreamwalker from [this](https://github.com/TrinityCore/TrinityCore/commit/9f9507e6a1fd50a5ce643a4096c1712700244a61#commitcomment-26632225) that seems to be failing not this. From [travis](https://travis-ci.org/TrinityCore/TrinityCore/builds/324918053#L1836)

```
/home/travis/build/TrinityCore/TrinityCore/src/server/scripts/Northrend/IcecrownCitadel/boss_valithria_dreamwalker.cpp:867:18: fatal error: 'MovementInform' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
            void MovementInform(uint32 /*type*/, uint32 id)
                 ^
/home/travis/build/TrinityCore/TrinityCore/src/server/game/AI/ScriptedAI/ScriptedCreature.h:169:10: note: overridden virtual function is here
    void MovementInform(uint32 /*type*/, uint32 /*id*/) override { }
         ^
```

jackpoz:
Thanks @DanVS . Travis warning is not about this PR so it's all ok. 
For reference I find http://en.cppreference.com/w/cpp/language/operator_precedence to be very useful when wondering about operators.

I'll merge this PR later on (I think I'll change the "!( == )" to "!=" )

jackpoz:
thanks for the PR

PR #21164, opened 2018-01-02
Core/Spells: Improved Stormstrike proc

DanVS:
**Changes proposed:**
-  [Improved Stormstrike](https://dbwotlk.com/?spell=51522) doesn't restore mana on creature kill (works on hit).
-  PR fixes this behaviour by applying spell fix.

- [x] 3.3.5
- [ ] master

**Tests performed:** builds, works in game.

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -3093,6 +3093,15 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->ProcChance = 0;
     });
 
+    // Improved Stormstrike
+    ApplySpellFix({ 
+        51521, // (Rank 1)
+        51522  // (Rank 2)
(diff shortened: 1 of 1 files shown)


Keader:
Hack.
It need be handled by a AuraScript i belive.

jackpoz:
what are the current proc flags of the spells ?

DanVS:
04 - Successful attack by Spell that use melee weapon.

joschiwald:
you should not add PROC_FLAG_KILL, the spell should proc even the npc dies, if not, the issue is somewhere else

ariel-:
The preferred way to modify procs is through DB `spell_proc`. And as said this is not the correct solution. Moving the proc to cast phase should work.

ariel-:
Issue fixed in 4e9d34691997b7dded45c7857b67413af7ff7c6f

Issue #21032, opened 2017-12-07
Core/Pooling: fishing school gameobjects not visible


Strazhnik:
**Description:**
Gameobject don't visible where id=182956, 182954, 182953
http://www.wowhead.com/object=182956/school-of-darter
 
**Current behaviour:**
Not visible

**Expected behaviour:**
Should be visible

**Steps to reproduce the problem:**
1. go xyz where gameobject id=182956, 182954, 182953 

**Branch(es):**
3.3.5

**TC rev. hash/commit:** 
123e8a95ba60b4706617da03eb047efadc32e0e9

**TDB version:**  63

**Operating system:**  Win 7

Killyana:
There's an issue with the pooling, I don't know if the "chance" is wrong or something related to the last changes
Example the pool 311 has max_limit=1 but you find none of the gobs spawned:
``` 
guid, pool_entry, chance
64547,  311,        0, 
64585,  311,        0, 
64704,  311,       20,

Killyana:
@lachtanek can you check this issue?

lachtanek:
@Killyana isn't here an error during server load, along the lines of
```Pool Id X has no equal chance pooled entites defined and explicit chance sum is not 100.```
because by a quick look that pool doesn't seem valid to me, when you specify a chance for one member, you need to specify it for the others aswell.

Killyana:
No errors on db or server log.

lachtanek:
Thanks for clarification, looks like my code is wrong because I didn't understand it correctly before. 
How do I do this, open a new PR?

Killyana:
Yes please.

Strazhnik:
@lachtanek, may be revert?

lachtanek:
@Strazhnik sorry, kind of forgot about this, I'll open a new PR in a few days with the fix

ariel-:
c50aefac963aca5daeacf409deca9692c6ca3924

PR #21170, opened 2018-01-03
Core/Pools: Fixed a bug in explicitly + equally chanced pools

lachtanek:
Fixed a bug introduced in https://github.com/TrinityCore/TrinityCore/commit/de80cd2e0de470522a734e8e7a65e81b82bfc5fb (https://github.com/TrinityCore/TrinityCore/issues/21032).
@Killyana, can you please test it?
I can't do it right now, but I think this is simple enough fix to not need any extensive testing.

### DIFF
--- a/src/server/game/Pools/PoolMgr.cpp
+++ b/src/server/game/Pools/PoolMgr.cpp
@@ -328,36 +328,27 @@ void PoolGroup<T>::SpawnObject(ActivePoolData& spawns, uint32 limit, uint32 trig
         // roll objects to be spawned
         if (!ExplicitlyChanced.empty())
         {
-            while (count && ExplicitlyChanced.size() > rolledObjects.size())
-            {
-                --count;
-                float roll = (float)rand_chance();
+            float roll = (float)rand_chance();
 
-                for (PoolObject& obj : ExplicitlyChanced)
+            for (PoolObject& obj : ExplicitlyChanced)
+            {
+                roll -= obj.chance;
+                // Triggering object is marked as spawned at this time and can be also rolled (respawn case)
+                // so this need explicit check for this case
+                if (roll < 0 && (obj.guid == triggerFrom || !spawns.IsActiveObject<T>(obj.guid)))
                 {
-                    roll -= obj.chance;
-                    // Triggering object is marked as spawned at this time and can be also rolled (respawn case)
-                    // so this need explicit check for this case
-                    if (roll < 0 && (obj.guid == triggerFrom || !spawns.IsActiveObject<T>(obj.guid)))
-                    {
-                        rolledObjects.push_back(obj);
-                        break;
-                    }
+                    rolledObjects.push_back(obj);
+                    break;
                 }
             }
         }
 
         if (!EqualChanced.empty() && rolledObjects.empty())
         {
-            rolledObjects = EqualChanced;
-
-            for (auto itr = rolledObjects.begin(); itr != rolledObjects.end();)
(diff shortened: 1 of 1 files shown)


Treeston:
Please explain what the PR does and how it achieves it.

lachtanek:
In my first PR, I fixed a bug in pools which resulted in less and less objects being spawned the more were respawned basically. Sadly, I misunderstood one part of the pooling system (explained here https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130271/pool+gameobject#pool_gameobject-chance) and broke this. there was apparently a bug that triggering object couldn't spawn (https://github.com/TrinityCore/TrinityCore/pull/20949#issuecomment-350083960), fixed that one aswell.
This PR should fix bugs.
Sorry for not clarifying this in original post.

Killyana:
No change, this gobs still not spawned.

lachtanek:
Okay, thanks, will look into that probably tomorrow

lachtanek:
Should be fixed now, tested on pool 311, spawned ok after server start and respawned after 1st loot.

Killyana:
Yes this fix the issue #21032 but #19257 still happen, probably not related to this PR.

lachtanek:
What exactly is a problem there? I can't seem to understand it from the comments, they seem to be talking about multiple different issues.

Killyana:
It's pretty simple: .go c 86075
You will see 4 mobs in this pit, everyone has a pool at the exact same position it could be Entry: 4131or Entry: 4133
Kill and respawn this 4 npcs many times, you will notice at some point they will not respawn.

lachtanek:
The issue is that the creature is part of new spawning system.
What happened before was that when PoolMgr::SpawnObject was called, the creature was already in world (called here: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Creature/Creature.cpp#L2066) and pool system either selected it and left it spawned, or not and despawned it.
What happens with new spawning system, old creature is deleted and a new one is spawned. The problem is, if Map::CheckRespawn finds out that it is pooled creature, it doesn't do anything and leaves the respawn on pool system. Same thing will probably happen to GOs.

One solution is for map to respawn the creature anyway and pool system might despawn it right away, if it feels that way (or if it is not rolled), which feels hacky (same kind of hacky as current system).
Other solution is to somehow pass the info about whether creature is dynamically spawned or not to pool system, or let it find out somehow, which doesn't feel right either.
Third solution may be to wait for new pool system -_-
Should I try to fix it regardless?

offl:
> The issue is that the creature is part of new spawning system.

This bug was here since the pool implementation. This bug occurs on other mangos-based projects(vanilla, for example).

> Should I try to fix it regardless?

It's a different issue, maybe in a new PR?

Killyana:
I tried to reproduce it in an older revision without dynamic spawn, and the issue never happen, it's definitively related to dynamic spawns, maybe @r00ty-tc can help with that.

Shauren:
Since this PR is confirmed to fix #21032 and does not even attempt to fix #19257 then its good to go

shamage:
Hi, two days ago I went to the repair - see the attachment, and I did a new compilation of the DB, but now after the update I will not start the server and write the error (see the picture in the attachment). Do you know how to fix me? Thank you for your advice!

tkrokli:
This is not a forum. Please use the TrinityCore forum pages for questions about personal configuration, or ask in the **#trinity** help channel on IRC (see https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130200/IRC for details).

PR #21172, opened 2018-01-03
Core/Creature: improve DB linked_respawn handling

kelno:
**Description:**
Fix `SetCreatureLinkedRespawn` function possibly erasing/replacing a GameObject link (instead of only creature ones)

**Changes proposed:**

-  Specify link type 0 (creature to creature) in linked_respawn db requests related to creatures.
- Also properly delete linked_respawn entry when deleting a creature

- [x] 3.3.5

-  This prevent incorrectly removing a gameobject link with the '.npc set link' command.

- Tested ig, command still work as intended + deleting an npc removes his linked_respawn entry.

- /


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1688,6 +1688,10 @@ void Creature::DeleteFromDB()
     stmt->setUInt32(0, m_spawnId);
     trans->Append(stmt);
 
+    stmt = WorldDatabase.GetPreparedStatement(WORLD_DEL_CRELINKED_RESPAWN);
+    stmt->setUInt32(0, m_spawnId);
+    trans->Append(stmt);
```
joschiwald: is this missing in gameobject too?

kelno: Yes! I noticed it afterwards but had already done the pull request. Do you think I should add it here too?

Shauren: yes

Shauren:
Hmm, not too sure about this fix
What if linkType is 1 or 3? linking creatures to gameobjects is possible too
https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Globals/ObjectMgr.h#L1761

kelno:
Those two cases (1 and 3) are missing from the delete too, I'll look into it.
Not sure if you were pointing to that, but SetCreatureLinkedRespawn is currently usable for links between creatures only, so no changes on that side.

src/server/game/Globals/ObjectMgr.h
```diff
@@ -1678,10 +1678,10 @@ class TC_GAME_API ObjectMgr
 
         enum CreatureLinkedRespawnType
```
kelno: I also added indexes here since they are fixed values

src/server/database/Database/Implementation/WorldDatabase.cpp
```diff
@@ -24,8 +24,11 @@ void WorldDatabaseConnection::DoPrepareStatements()
         m_stmts.resize(MAX_WORLDDATABASE_STATEMENTS);
 
     PrepareStatement(WORLD_SEL_QUEST_POOLS, "SELECT entry, pool_entry FROM pool_quest", CONNECTION_SYNCH);
-    PrepareStatement(WORLD_DEL_CRELINKED_RESPAWN, "DELETE FROM linked_respawn WHERE guid = ?", CONNECTION_ASYNC);
-    PrepareStatement(WORLD_REP_CREATURE_LINKED_RESPAWN, "REPLACE INTO linked_respawn (guid, linkedGuid) VALUES (?, ?)", CONNECTION_ASYNC);
+    PrepareStatement(WORLD_DEL_CREATURE_LINKED_RESPAWN, "DELETE FROM linked_respawn WHERE guid = ? AND linkType IN (0,1)", CONNECTION_ASYNC); //CREATURE_TO_CREATURE, CREATURE_TO_GO
```
joschiwald: i would hide the values of the linkType from db layer and fill the values when you use the prepared queries

kelno:
Tested again, should be okay. Removing either the slave or the master of the link now deletes the linked_respawn entry.

src/server/database/Database/Implementation/WorldDatabase.cpp
kelno: Make sense

kelno: Would you prefer, for example in the case of  WORLD_DEL_CREATURE_LINKED_RESPAWN:
- Always feeding two values of CreatureLinkedRespawnType to the same query. But that doesn't look very generic to me.
- Calling the query two times to remove both CREATURE_TO_CREATURE and CREATURE_TO_GO types. This also implies we have no less than four queries just to remove a creature, since we also must add two for the *_MASTER queries.

Maybe go with the second option but make it a function that takes a reference to a transaction to avoid cluttering the DeleteFromDB functions with 4 queries ?

kelno: Would you prefer, for example in the case of WORLD_DEL_CREATURE_LINKED_RESPAWN:

Always feeding two values of CreatureLinkedRespawnType to the same query. But that doesn't look very generic to me.
Calling the query two times to remove both CREATURE_TO_CREATURE and CREATURE_TO_GO types.
I'd go with the latter option but this also implies we have no less than four queries just to remove a creature, since we also must add two for the *_MASTER queries.
Maybe go with the latter option but make it a RemoveLinkedRespawn function that takes a reference to a transaction to avoid cluttering the DeleteFromDB functions with 4 queries ?

kelno: Would you prefer, for example in the case of WORLD_DEL_CREATURE_LINKED_RESPAWN:

Always feeding two values of CreatureLinkedRespawnType to the same query. But that doesn't look very generic to me.
Calling the query two times to remove both CREATURE_TO_CREATURE and CREATURE_TO_GO types.
I'd go with the latter option but this also implies we have no less than four queries just to remove a creature, since we also must add two for the *_MASTER queries.
Maybe go with the latter option but make it a RemoveLinkedRespawn function that takes a reference to a transaction to avoid cluttering the DeleteFromDB functions with 4 queries ?

joschiwald:
```sql
DELETE FROM linked_respawn WHERE guid = ? AND linkType = ?
```
that way would be fine
don't care about the 4 queries, it is the job of the dbms

src/server/game/Maps/SpawnData.h
```diff
@@ -73,4 +73,12 @@ struct SpawnData
     SpawnData(SpawnObjectType t) : type(t) {}
 };
 
+enum CreatureLinkedRespawnType
```
joschiwald: `LinkedRespawnType`

PR #21193, opened 2018-01-07
Core/Movement: Correct the allowed distance to target before a repositioning is necessary. #20173

### LINKED ISSUE #20173: [3.3.5] Mobs don't move to melee range when barely out of range so they don't autoattack
**Description:** Every single creature from the world has a range where they bug out and don't perform autoattacks neither walk closer to be able to do it. This turns out into an easy exploitable way to defeat any encounter on any instance/raid.

**Current behaviour:**

https://gfycat.com/AdvancedSlipperyAnemonecrab

You're able to get every single creature on the world into a "deadzone" where they do not autoattack nor walk towards their objective.

**Expected behaviour:** Creatures should be autoattacking you until you walk out of range, then, they should follow you to get in range again.

1. Get towards any creature.  
2. Aggro it.
3. Start walking away until you meet the range where the creature doesn't attack nor follows you.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** https://github.com/TrinityCore/TrinityCore/commit/76fc9ed6356ca34a603d4c716d2850b5b15a4164

**TDB version:**  TDB 3.3.5 63

**Operating system:** Ubuntu 17.04

Tagging @Shauren cause maybe this is because of combat range changes, this is pretty huge imo.

chaodhib:
Correct the allowed distance to target before a repositioning is necessary. #20173

Also getting rid of the wordserver config parameter 'TargetPosRecalculateRange' since it is no longer needed.

**Changes proposed:**

Let's say a creature needs to AA a non-moving unit:
1. it moves closer to its target until it is in melee range. 
2. Once the chaser is close enough to its (non-moving until this point) target to attack it, it stops. 
3. Then, let's say the target moves away from the chaser, the chaser will not repositioning immediately: it will wait until the distance between itself and its target is beyond a certain threshold to start repositioning. 

The waiting is an optimization, used in order to reduce the number of times the path is recomputed when the target is moving. The issue is as follow: the threshold's value was too high. The target could move away to a certain distance, outside of melee range and the mover will not reposition.

With this PR, this is now fixed. The threshold new value is correctly set to the melee attack range.

I'm also talking about the issue [HERE](https://github.com/TrinityCore/TrinityCore/pull/17850#issuecomment-355764994) and [HERE](https://github.com/TrinityCore/TrinityCore/pull/17787#issuecomment-239819554).

**Target branch(es):**

- [X] 3.3.5
master. Maybe?

**Issues addressed:** Closes #20173

**Tests performed:** Tested IG with a melee attacker (The Lich King in The Frozen Throne)

- [X] Need to investigate to see if this change will be also valid for range attackers (like archer NPCs).  EDIT: **Done. There should be no impact**.
- [x] This change might not be needed for MoveFollow (a unit following another unit). Maybe I will rollback the change for that type of movement since the new value incur a performance cost. EDIT: **Done**.


ccrs:
The interesting research here is to decide if the distance should be adjusted depending on the target situation.
Example, if its not in melee range, look for half the melee range distance to ensure the result.

Or even, make an algorithm to search for the most suitable point to move, that is, the point in which it should be attacking the target, not the one that is closer.

Many ideas come, yet regarding this PR, I think its a nice improvement to the current situation. Though take a look into what I said first, making the distance sort of dynamic depending on the situation.

jackpoz:
Slightly unrelated to the PR: I was thinking a couple of days ago if we should use the result of MeleeAttacks/Spells about the target being out of range/out of LoS and handle it next update instead of proactively checking every now and then (100 ms) if the target is still in range/in LoS.

For now reducing the distance sounds good enough. Could you please change the "4.0f / 3.0f" magic numbers to something more self-explainatory ? even a #define keeping the same formula would be good, at first sight it's not clear what 4 and 3 mean in that context.

chaodhib:
I made a mistake in the new formula. It is fixed now. This guarantees that the target will always be in melee range.

@ccrs I'm not sure I understand what you mean in the first part.

@jackpoz Regarding the magic numbers, it is no longer need after latest changes. Regarding the first, could be an interesting optimization. The only downside is that if the chaser and the target have the same speed, the target will be able to kit the mover by alternating between moving and stopping (thus getting in out of range, making the attack miss time to attack).

chaodhib:
All the remaining points are solved. Imo, it's ready to be merged. Will merge in a few days if no further comments.

Treeston (approved):

ccrs:
What I meant is, whats better, moving to the point in which we exactly are in melee range, or move half the distance closer so the repositioning is less likely? That is, if the conditions are met, we detect a slope or obstacles or w/e.

But like I said, this is a nice improvement for current state.

ccrs (approved):

chaodhib:
@ccrs It's already the case: NPCs move at `mover->CombatReach + target->CombatReach + 0.5y` away from the target. meanwhile, the melee range is `mover->CombatReach + target->CombatReach + 1.333y`. So the target can move up to 0.8333 yards before trigerring a repositioning of the mover.

Issue #18190, opened 2016-11-03
[7.x] Conditions for phase area under zone are not executed


Lopfest:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:** conditions for area phases are ignored when there is a phase for whole zone

**Current behaviour:** phase area conditions are not checked at all, when there is an entry in phase_area for whole zone and you make condition only for specific area under that zone. Even if you have condition in DB, the core evaluates it always like following:
```cpp
bool ConditionMgr::IsObjectMeetToConditions(ConditionSourceInfo& sourceInfo, ConditionContainer const& conditions) const
{
    if (conditions.empty())
        return true;          <------
    ...
}
```
Function `void WorldObject::UpdateAreaAndZonePhase()` is not passing any phase.Conditions argument here `if (sConditionMgr->IsObjectMeetToConditions(this, phase.Conditions))`

EDIT: I have found out in DBErrors.log that this kind of condition is not even loaded on server startup
[Condition SourceType: 26 (Phase), SourceGroup: 1329, SourceEntry: 5848] Area 5848 does not have phase 1329.

**Expected behaviour:** evaluate every existing phase_area condition when you have zoneId (not areaId) in phase_area and condition only for area within that zone.
Below code needs to be changed, so it checks also areas within entered zoneId.
```
// condition loading helpers
std::vector<PhaseInfoStruct>* GetPhasesForAreaOrZoneForLoading(uint32 areaOrZone)
{
    auto itr = _phases.find(areaOrZone);
    return itr != _phases.end() ? &itr->second : nullptr;
}
```

Aditionally, switch `cond->SourceGroup, cond->SourceEntry` to `cond->SourceEntry, cond->SourceGroup` in `TC_LOG_ERROR("sql.sql", "%s Area %u does not have phase %u.", cond->ToString().c_str(), cond->SourceGroup, cond->SourceEntry);` because it is now vice-versa.

**Steps to reproduce the problem:**

1. create phase for zone
2. restrict this phase in specific area with condition
3. enter the area when condition should apply= condition is not working

Sample SQL:
```sql
DELETE FROM `phase_area` WHERE `AreaId`=5736 AND `PhaseId`=1329;
INSERT INTO `phase_area` (`AreaId`, `PhaseId`, `Comment`) VALUES
[...]

PR #21214, opened 2018-01-09
Core/Object: fix logic in WorldObject::UpdateAreaAndZonePhase()

### LINKED ISSUE #21196: GetAreaAndZonePhases is loading wrong areas
**Description:**
when loading phases from DB, they are loaded this way: `_phases[area].push_back(phase);`
however this "area" index is (probably) never used when deciding which areas to load during player's area change.

**Current behaviour:**
currently `GetAreaAndZonePhases` gets ALL phases from DB, regardless of which area or zone the player is located in. if there are multiple areas with the same phase, only the last phase which is loaded is applied. others are overwritten by the result of last loaded phase check - which always results to remove phase (`_phases.erase(id)` in method `SetInPhase`)

here is some output from the console, where i have put some logs:
image

**Expected behaviour:**
GetAreaAndZonePhases should load ONLY phases for area or zone, where player is currently located.

1. insert phase into DB `INSERT INTO phase_area (AreaId, PhaseId, Comment) VALUES (5736, 169, '');`
2. go to the wandering isle, other area than Shang Xi Training Grounds, for example here: `.go xyz 1000 3600 193`
3. phase is not applied

**Branch(es):** master

**TC rev. hash/commit:** 7d00ae40458fe4399ca414c79a9b0a88b863267a

**TDB version:**  TDB 720.00

Traesh:
**Changes proposed:**

- Fix logic in Updateareaphases
- Phases priority won't be determined by areaId order anymore
- Allow to remove a zone phase on a specific area (phases on zone & area, condition on area)

**Target branch(es):** 
- [x] master

**Issues addressed:** Closes #21196, #18190

**Tests performed:** Build, tested by Lopfest ingame

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2888,35 +2888,27 @@ bool WorldObject::HasInPhaseList(uint32 phase)
 void WorldObject::UpdateAreaAndZonePhase()
 {
     bool updateNeeded = false;
-    PhaseInfo const& phases = sObjectMgr->GetAreaAndZonePhases();
-    for (PhaseInfo::const_iterator itr = phases.begin(); itr != phases.end(); ++itr)
-    {
-        uint32 areaOrZoneId = itr->first;
+    PhaseInfo const& allAreasPhases = sObjectMgr->GetAreaAndZonePhases();
+    uint32 zoneAndArea[] = { GetZoneId(), GetAreaId() };
+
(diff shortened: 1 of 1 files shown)


Shauren:
Small cleanups plus
* fixed copying every element of allAreasPhases in first loop
* fixed copying PhaseInfoStruct in 2nd loop

Shauren (approved):

PR #21217, opened 2018-01-10
Scripts/Spells: limit use of Slam with Bloodsurge talent

tje3d:
**Changes proposed:**

Warrior should just be able to use 1 Instant Slam with [Bloodsurge](https://wotlk.evowow.com/?spell=46915) talent.
Currently the aura stays for 5 seconds and a player can spam use of slam.

**Target branch(es):**

- [x] 3.3.5


tje3d:
this will break Santified Ymirjar Lord's, second process. 
this should remove 1 bloodsurge aura for one slam usage.
how i can do that?

tje3d:
4861864 should do the job

joschiwald:
this is not good, spell 46916 should handle the charge drop if not you need to adjust the spell_proc entry in database

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -911,10 +911,19 @@ class spell_warr_slam : public SpellScriptLoader
             {
                 if (!GetHitUnit())
                     return;
-                GetCaster()->RemoveAurasDueToSpell(SPELL_WARRIOR_SLAM_AURA);
                 CastSpellExtraArgs args(TRIGGERED_FULL_MASK);
                 args.SpellValueOverrides.AddBP0(GetEffectValue());
                 GetCaster()->CastSpell(GetHitUnit(), SPELL_WARRIOR_SLAM, args);
+
+                Aura* aura = GetCaster()->GetOwnedAura(SPELL_WARRIOR_SLAM_AURA);
+                if (aura) {
```
tkrokli: TC codestyle:
```cpp
                if (aura)
                {
                    uint8 auraCount = aura->GetCharges();
                    if (auraCount > 1)
                        aura->SetCharges(auraCount - 1);
                    else
                        GetCaster()->RemoveAurasDueToSpell(SPELL_WARRIOR_SLAM_AURA);
                }
```
(possibly with a blank line between 920 & 921 for readability)

joschiwald: that won't make the code better

tkrokli:
Very well, next step is to follow up on joschiwald's comment https://github.com/TrinityCore/TrinityCore/pull/21217#issuecomment-357278895

src/server/scripts/Spells/spell_warrior.cpp
tje3d: @joschiwald All suggestions are welcome, how i can make it better ? cause i cant find anything to reduce aura one by one. anyway i'm going to fix this using spell_proc db

ariel-:
What is this I don't even

ariel-:
Hmph. Slam has SPELL_ATTR6_DONT_CONSUME_PROC_CHARGES defined in #18950

Aura procs but it doesn't drop the charge from Slam! buff

Shauren:
Thats just a triggering dummy. 50783 which does damage doesn't have that attribute and also has SpellFamilyFlags set

tje3d:
@Shauren 50783 works well ( using .cast 50783, it just remove aura one by one )

ariel-:
And 50783 has SPELL_ATTR3_TRIGGERED_CAN_TRIGGER_PROC_2, I'm starting to think we don't handle the attribute correctly (and never did)

Issue #15224, opened 2015-08-05
Core/Vehicles - Problem with Installing Accessories


sirikfoll:
hash 747ea7c
3.3.5 branch
TDB 335.59

Add accessories with vehicle_template_accessory doesn't work for some vehicles, such as:

VehicleId: 356 - Thorim's Hammer Targetting Reticle (http://www.wowhead.com/npc=33364)
VehicleId: 357 - Freya's Ward Targetting Reticle (http://www.wowhead.com/npc=33366)
VehicleId: 347 - Hodir's Fury Targetting Reticle (http://www.wowhead.com/npc=33108)

the behavior is little bit different for
VehicleId: 341 - Flame Leviathan Seat (http://www.wowhead.com/npc=33114),
if you spawn the seat with .npc add 33114 it will install the accessories. 
But in normal conditions(33114 is a accessory of 33113) the accessories seems to be installed and despawned instantly

Killyana:
After https://github.com/TrinityCore/TrinityCore/commit/4ada913d02bb5376914e46843873b3285116154f the 2 chairs are ejected on evade.
Can you confirm?

sirikfoll:
to me, it looks the same as before

Killyana:
For me they are just ejected on evade

sirikfoll:
The problem isn't with seats being ejected, but the not install of the npcs IN the flame leviathan seat(33114), try this
```sql
DELETE FROM `vehicle_template_accessory` WHERE `entry` IN (33114);
INSERT INTO `vehicle_template_accessory` (`entry`, `accessory_entry`, `seat_id`, `minion`, `description`, `summontype`, `summontimer`) VALUES
(33114, 33142, 1, 1, 'Flame Leviathan Seat', 6, 30000),
(33114, 33143, 2, 1, 'Flame Leviathan Seat', 6, 30000);
```

.npc add temp 33114   will show the npcs normally
but note that in Flame leviathan, the seat appears empty

Killyana:
We don't this accessory on TCDB, for this reason we don't have the same issue.

sirikfoll:
I'm testing on a fresh TDB, with addiction of the sql above
`faction`=2110
`unit_flags`=33570816

ccrs:
Is this the same bug as Pit of Saron Drake Riders?

http://www.wowhead.com/npc=36891/
http://www.wowhead.com/npc=31260/

Killyana:
[...]

PR #21220, opened 2018-01-11
Core/Vehicles: Fix accessories disappearing on evade and multiple join events on seat -1

lachtanek:
**Changes proposed:**

- Move Vehicle::Reset from CreatureAI::EnterEvadeMode to CreatureAI::JustReachedHome
- Check all pending join events on add passenger with seat -1

**Description:**
First issue can be reproduced by .go creature 33113 on 10man and calling .npc evade on him (Flame Leviathan). The seats and his turret should disappear (not entirely sure since I already worked on 10 different versions of this). The problem is, Vehicle::InstallAccessories is called when creature evades, so it spawns accessories, they try to cast vehicle ride spell on leviathan, but it is evaded so they are immediatelly despawned.

Second issue can be reproduced by going to Ulduar 25man somewhere, logging 2nd account and going to the same place, then .npc add temp 32934, targetting right arm and .cast self 63981. The arm looks like it's grabbing both players, but in the end only grabs one.

This PR is needed for my upcoming (soon™) rewrite of Kologarn and Flame Leviathan.
It shouldn't break anything else, maybe fix some more things, I can't think of any right now through, since both are pretty rare cases.

- [x] 3.3.5
- [ ] master ??

**Issues addressed:**
 probably none

builds, tested

- [ ] Debug and fix vehicles on vehicles with accessories, ~~for example Flame Leviathan having seats and seats should have overload device and attackable creature on them, but they are despawned immediatelly~~ Better described in #15224. I know calling Vehicle::Reset(true) for seats a few milliseconds after they spawn helps, but that is not a fix. (will come soon, hopefully)

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -167,7 +167,10 @@ void CreatureAI::EnterEvadeMode(EvadeReason why)
     }
 
     Reset();
+}
 
+void CreatureAI::JustReachedHome()
+{
     if (me->IsVehicle()) // use the same sequence of addtoworld, aireset may remove all summons!
         me->GetVehicleKit()->Reset(true);
 }
--- a/src/server/game/AI/CreatureAI.h
+++ b/src/server/game/AI/CreatureAI.h
@@ -142,7 +142,7 @@ class TC_GAME_API CreatureAI : public UnitAI
         void OnCharmed(bool isNew) override;
 
         // Called at reaching home after evade
-        virtual void JustReachedHome() { }
+        virtual void JustReachedHome();
 
         void DoZoneInCombat(Creature* creature = nullptr);
 
--- a/src/server/game/AI/ScriptedAI/ScriptedCreature.h
+++ b/src/server/game/AI/ScriptedAI/ScriptedCreature.h
@@ -324,7 +324,7 @@ class TC_GAME_API BossAI : public ScriptedAI
         void Reset() override { _Reset(); }
         void JustEngagedWith(Unit* /*who*/) override { _JustEngagedWith(); }
         void JustDied(Unit* /*killer*/) override { _JustDied(); }
-        void JustReachedHome() override { _JustReachedHome(); }
+        void JustReachedHome() override { ScriptedAI::JustReachedHome(); _JustReachedHome(); }
 
         bool CanAIAttack(Unit const* target) const override;
 
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_professor_putricide.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_professor_putricide.cpp
@@ -283,7 +283,7 @@ class boss_professor_putricide : public CreatureScript
 
             void JustReachedHome() override
             {
-                _JustReachedHome();
+                BossAI::JustReachedHome();
                 me->SetWalk(false);
                 if (events.IsInPhase(PHASE_COMBAT_1) || events.IsInPhase(PHASE_COMBAT_2) || events.IsInPhase(PHASE_COMBAT_3))
                     instance->SetBossState(DATA_PROFESSOR_PUTRICIDE, FAIL);


Killyana:
Can you test if it fixes https://github.com/TrinityCore/TrinityCore/issues/15591 ?

joschiwald:
about the second point, i had the same idea long time ago, but it was classified as hack
see my initial approach https://gist.github.com/joschiwald/36da1450dcea394a9ca3

lachtanek:
Why was it? I can't think of a better way to fix it, scheduling the spells to land after each other seems a lot worse, maybe update VehicleSeat struct with info about scheduled join, but that didn't seem right to me either.

joschiwald:
can remember why, but i'm using my patch since indrudion of the async vehicle behavoir and it fixed a lot of problems

jackpoz:
@joschiwald should you push your patch then or open a PR about it to resume the discussion about being a hack or not ?

Killyana:
Fixes https://github.com/TrinityCore/TrinityCore/issues/21410 probably

sirikfoll:
Any updates? I'm using Joschiwald's solution plus the AI changes from this PR, it helps a lot with the issues..
I can test the final solution if needed

jackpoz:
so which one should we apply, the @joschiwald patch or this PR ?

lachtanek:
if @joschiwald wants to create MR for his patch I can modify this one to only be the 1st fix, but in the year I forgot about most of the inner workings of vehicle systems and the problems it has I have debugged, because no time anymore, so I'm very unlikely to develop a different solution in the near future

jackpoz:
https://github.com/TrinityCore/TrinityCore/pull/23233 contains @joschiwald "code" (it's not a patch, it's not a diff, it references a commit hash that doesn't exist, it's something else) .
The 2nd issue will have its own PR later on (or I could edit this one and merge it with you as author)

jackpoz:
Merge latest 335 and remove the multiple vehicle join changes that got applied by joschivald patch already

Killyana:
I tested the Leviathan, Mimiron, Kologarn and for this 3 accessories despawns on evade and new one are installed when the unit reach home position
For Putricide The 2 tentacles are ejected on evade but never installed.

jackpoz:
> First issue can be reproduced by .go creature 33113 on 10man

this teleports me to Kalimdor

jackpoz:
> The problem is, Vehicle::InstallAccessories is called when creature evades, so it spawns accessories, they try to cast vehicle ride spell on leviathan, but it is evaded so they are immediatelly despawned.

could you provide more details about this, like which lines causes the issue (and maybe a callstack) ?

Killyana:
Here a test I did in 2 old year core where this issue doesn't exist:
.go xyz 160.0196, 26.3005, 409.8037 603
.gm on
.npc follow the Leviathan
Once the Leviathan is in range hit it to make it evade
Try to click the leviathan when he's in evade mode before reaching the home position
What happen in the old core: you could ride the leviathan
What happen in the current revision: you cannot ride the leviathan

It looks like the effect ride vehicle musn't be ignored during evade

lachtanek:
@jackpoz this improvised "callstack" should provide insight into what's happening

- Creature enters evade, `CreatureAI::EnterEvadeMode(...)` called, we are interested in these 2 lines https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/CreatureAI.cpp#L171
- `Vehicle::Reset(true)`
- `Vehicle::InstallAllAccessories(true)`
- for each `minion` accessory `Vehicle::InstallAccessory(...)` (Flame Leviathan's seats for example are  minions)
- `Unit::HandleSpellClick(...)` called for vehicle
- `Unit::CastCustomSpell(...)` called for spawned accessory (`target=vehicle`) casting spell with apply aura `SPELL_AURA_CONTROL_VEHICLE`
- I don't want to dive through all Spell.cpp methods here but a few calls later we get to...
- `Spell::DoSpellHitOnUnit(...)` called (`target=vehicle`) which returns SPELL_MISS_EVADE

I could devote some time to compiling the core again and posting full callstack from gdb if you'd like and also posting reproduce steps as I did not write them in original PR text for some reason

Thanks for looking into this!

Killyana:
Restoring: if (unit != m_caster && !GetSpellInfo()->HasAura(SPELL_AURA_CONTROL_VEHICLE))      before https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/Spell.cpp#L2630
Fixes the issue, I cannot fin the commit that removed this line.

jackpoz:
I missed the SPELL_MISS_EVADE while debugging, that will surely help. I think the best way is to fix the issue there

jackpoz:
Please try https://github.com/jackpoz/TrinityCore/tree/3.3.5-vehicle_accessories_homemovegen , in particular https://github.com/jackpoz/TrinityCore/commit/eb959030da75f0e8d925a1979a57630621996951

jackpoz:
2nd issue fixed with https://github.com/TrinityCore/TrinityCore/commit/6dae4a747da558ca46a777186d4519c32fc1bf04

Issue #20974, opened 2017-11-26
The Storm King's Vengeance


jerbookins:
**Description:**
http://www.wowhead.com/quest=12919/the-storm-kings-vengeance
http://www.wowhead.com/spell=55421/gymers-throw
http://www.wowhead.com/spell=55426/gymers-smash

**Current behaviour:**
Scourge damaged or killed by Gymer's Throw do not fulfill quest criteria. Only scourge killed exclusively with Gymer's Smash fulfill quest criteria.

**Expected behaviour:**
Any scourge NPC should fulfill quest criteria when killed by Gymer.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** e85a48179e3c

**TDB version:**  335.63

**Operating system:** Windows 10

Atis6:
Try this:
`
-- Vargul Blighthound
SET @ENTRY := 29889;
SET @SOURCETYPE := 0;

DELETE FROM `smart_scripts` WHERE `entryorguid`=@ENTRY AND `source_type`=@SOURCETYPE;
UPDATE creature_template SET AIName="SmartAI" WHERE entry=@ENTRY LIMIT 1;
INSERT INTO `smart_scripts` (`entryorguid`,`source_type`,`id`,`link`,`event_type`,`event_phase_mask`,`event_chance`,`event_flags`,`event_param1`,`event_param2`,`event_param3`,`event_param4`,`action_type`,`action_param1`,`action_param2`,`action_param3`,`action_param4`,`action_param5`,`action_param6`,`target_type`,`target_param1`,`target_param2`,`target_param3`,`target_x`,`target_y`,`target_z`,`target_o`,`comment`) VALUES 
(@ENTRY,@SOURCETYPE,0,0,0,0,100,0,6000,10000,16000,23000,11,50047,0,0,0,0,0,2,0,0,0,0.0,0.0,0.0,0.0,"Vargul Blighthound - In Combat - Cast 'Broken Bone' (Phase 1) (No Repeat)"),
(@ENTRY,@SOURCETYPE,1,0,0,0,100,0,2000,15000,26000,33000,11,50046,0,0,0,0,0,2,0,0,0,0.0,0.0,0.0,0.0,"Vargul Blighthound - In Combat - Cast 'Gnaw Bone' (Phase 1) (No Repeat)"),
[...]

PR #21245, opened 2018-01-16
Core/SmartScripts: TARGET_INVOKER not targeting passengers of vehicle

tje3d:
This will fix quest’s like “The Black Knight's Orders”, “A Rough Ride”

**Target branch(es):** 3.3.5

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -2687,13 +2687,13 @@ void SmartScript::GetTargets(ObjectVector& targets, SmartScriptHolder const& e, 
             }
             break;
         }
-        case SMART_TARGET_VEHICLE_ACCESSORY:
+        case SMART_TARGET_VEHICLE_PASSENGER:
         {
             if (me && me->IsVehicle())
-            {
-                if (Unit* target = me->GetVehicleKit()->GetPassenger(e.target.vehicle.seat))
-                    targets.push_back(target);
-            }
+                for (auto seatItr = me->GetVehicleKit()->Seats.begin(); seatItr != me->GetVehicleKit()->Seats.end(); ++seatItr)
+                    if (!e.target.vehicle.seatMask || (e.target.vehicle.seatMask & (1 << seatItr->first)))
+                        if (Unit* u = ObjectAccessor::GetUnit(*me, seatItr->second.Passenger.Guid))
+                            targets.push_back(u);
             break;
         }
         case SMART_TARGET_POSITION:
--- a/src/server/game/AI/SmartScripts/SmartScriptMgr.h
+++ b/src/server/game/AI/SmartScripts/SmartScriptMgr.h
@@ -1187,7 +1187,7 @@ enum SMARTAI_TARGETS
     SMART_TARGET_CLOSEST_FRIENDLY               = 26,   // maxDist, playerOnly
     SMART_TARGET_LOOT_RECIPIENTS                = 27,   // all players that have tagged this creature (for kill credit)
     SMART_TARGET_FARTHEST                       = 28,   // maxDist, playerOnly, isInLos
-    SMART_TARGET_VEHICLE_ACCESSORY              = 29,   // seat number (vehicle can target it's own accessory)
+    SMART_TARGET_VEHICLE_PASSENGER              = 29,   // seatMask (0 - all seats)
 
     SMART_TARGET_END                            = 30
 };
@@ -1315,7 +1315,7 @@ struct SmartTarget
 
         struct
         {
-            uint32 seat;
+            uint32 seatMask;
         } vehicle;
     };
 };
--- a/sql/updates/world/3.3.5/2018_01_16_00_world.sql
+++ b/sql/updates/world/3.3.5/2018_01_16_00_world.sql
@@ -1,0 +1,1 @@
+UPDATE `smart_scripts` SET `target_param1`=(1 << `target_param1`) WHERE `target_type`=29;
--- a/src/server/game/AI/SmartScripts/SmartScriptMgr.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScriptMgr.cpp
@@ -484,7 +484,7 @@ bool SmartAIMgr::IsTargetValid(SmartScriptHolder const& e)
         case SMART_TARGET_STORED:
         case SMART_TARGET_LOOT_RECIPIENTS:
         case SMART_TARGET_FARTHEST:
-        case SMART_TARGET_VEHICLE_ACCESSORY:
+        case SMART_TARGET_VEHICLE_PASSENGER:
             break;
         default:
             TC_LOG_ERROR("sql.sql", "SmartAIMgr: Not handled target_type(%u), Entry %d SourceType %u Event %u Action %u, skipped.", e.GetTargetType(), e.entryOrGuid, e.GetScriptType(), e.event_id, e.GetActionType());


Treeston:
Why should any invoker target  target all passengers?

Aokromes:
i suppose quest / achievements credits.

jackpoz:
would it make more sense to add a new TARGET_* ?

tje3d:
@Treeston just what Aokromes said. For example in "The Black Knight's Orders" quest, Target of event_type WAYPOINT_REACHED is always null and that's why the quest will not be completed ( Same for "A Rough Ride" ).
This is just a try to find proper way.

Aokromes:
Maybe even this can fix achievement credits on ulduar on flame leviatan.

Treeston:
> would it make more sense to add a new TARGET_* ?
^this. `INVOKER` arbitrarily targeting vehicle passengers is counter-intuitive.

Killyana:
Fix https://github.com/TrinityCore/TrinityCore/issues/20974 probably.

Treeston (changes requested):
As was said: Create a new target type `SMART_TARGET_PASSENGER`, preferably with a bitmask indicating which passengers to target.

Aokromes:
What about the ones on c++? (flame leviathan)

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -2438,6 +2438,12 @@ void SmartScript::GetTargets(ObjectVector& targets, SmartScriptHolder const& e,
         case SMART_TARGET_ACTION_INVOKER:
             if (scriptTrigger)
                 targets.push_back(scriptTrigger);
+
+            if (me && IsUnit(me))
```
Shauren: but remove it for INVOKER

Shauren (changes requested):

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -2696,6 +2696,15 @@ void SmartScript::GetTargets(ObjectVector& targets, SmartScriptHolder const& e,
             }
             break;
         }
+        case SMART_TARGET_PASSENGER:
+        {
```
Treeston: Why braces?

src/server/game/AI/SmartScripts/SmartScriptMgr.h
```diff
@@ -1188,8 +1188,9 @@ enum SMARTAI_TARGETS
     SMART_TARGET_LOOT_RECIPIENTS                = 27,   // all players that have tagged this creature (for kill credit)
     SMART_TARGET_FARTHEST                       = 28,   // maxDist, playerOnly, isInLos
     SMART_TARGET_VEHICLE_ACCESSORY              = 29,   // seat number (vehicle can target it's own accessory)
+    SMART_TARGET_PASSENGER                      = 30,   // passenger's of vehicle
```
Treeston: passengers

Treeston (commented):
would still like a bitmask that lets you pick seats to target

tje3d:
@Treeston I've no idea how to implement bitmask 🤔

Treeston:
`if (!mask || (mask & (1 << seatId)))`

tje3d:
@Treeston is this ok?

Shauren:
@Treeston don't you think this makes SMART_TARGET_VEHICLE_ACCESSORY redundant?

Shauren (approved):

Treeston:
@Shauren are you assuming I understand smartscript?

_(but yes, probably)_

src/server/game/AI/SmartScripts/SmartScriptMgr.h
```diff
@@ -1317,6 +1318,11 @@ struct SmartTarget
         {
             uint32 seat;
         } vehicle;
+
+        struct
+        {
+            int8 seatId;
```
Treeston: `uint32`

src/server/game/AI/SmartScripts/SmartScriptMgr.h
```diff
@@ -1317,6 +1318,11 @@ struct SmartTarget
         {
             uint32 seat;
         } vehicle;
```
Treeston: what is this union member used for?

tje3d: Used by SMART_TARGET_VEHICLE_ACCESSORY:
`
if (Unit* target = me->GetVehicleKit()->GetPassenger(e.target.vehicle.seat))
`

Treeston: how many scripts in current 3.3.5 tdb use this target?

tje3d: Nothing at all

Treeston: remove it then (replace it in the enum with `_UNUSED`)

Killyana:
SMART_TARGET_VEHICLE_ACCESSORY is exactly the same target type, also "The Black Knight's Orders" and  "A Rough Ride" were working correctly months ago, better to find what change broke this.

Shauren:
ok, taking over for merge

tje3d:
@Killyana but the current SMART_TARGET_PASSENGER can do it better now

In those case, WaypointReached always pass `nullptr` as unit, so the TARGET_TYPE_ACTION_INVOKER would not works ( but TARGET_TYPE_PLAYER_RANGE, TARGET_TYPE_PLAYER_DISTANCE and ... would works )

Killyana:
>In those case, WaypointReached always pass nullptr as unit, so the TARGET_TYPE_ACTION_INVOKER would not works.
And it's normal?

Treeston:
I see no reason why `WaypointReached` should get any invoker passed to it.

tje3d:
I will send pull request's for those quest's after the pull req merged.

Killyana:
If I understand you're trying to fix this issue https://github.com/TrinityCore/TrinityCore/issues/20310  by modifying the existing SAI ?

tje3d:
@Killyana Yes at the first 😁

Killyana:
Bad Idea, the issue must be fixed on core side.

Shauren:
I have no objections to extending this sai target, even if it does't fix the issue

tje3d:
@Killyana but as @Treeston  said, maybe its not good idea to use ACTION_INVOKER with WaypointReached.
If it should works with existing AI, then i think we have to open new PULL.

Treeston:
@Killyana please present your reasoning for why `INVOKER` should arbitrarily select the vehicle's passenger for `WAYPOINT_REACHED`, which by design is not invoked by anyone?

Treeston (approved):

Killyana:
A rough ride: the player cast 51258 which trigger 51257, it will summon the npc 28308 (it's considered as a guardian) on waypoint reached the vehicle will execute quest explored (action 15) on target 7, but after the last changes on movement generator it doesn't work any more, and it's a common script used by many quests.

Treeston:
How does the owner of the guardian become the invoker of the waypoint reached event? Invoker is only set for interaction events.

meji46:
But there isn't invoker for waypoint reached event.

It's better to use SMART_TARGET_OWNER_OR_SUMMONER or SMART_TARGET_STORED in these cases.

Killyana:
All I know it was working before:
``` sql
SELECT * FROM smart_scripts WHERE entryorguid=28308;

Treeston:
> All I know it was working before:

Don't rely on bugs.

tje3d:
@Treeston  Is it good idea to check EVENT_TYPE in SMART_TARGET_ACTION_INVOKER? so we can check if the event was SMART_EVENT_WAYPOINT_* then we use OWNER_OR_SUMMONER as target? 
And this would fix current SAI ( What @Killyana  wants 😉 )

Treeston:
Could we? Yes.
Should we? Hell no, that's utterly counter-intuitive.

PR #21254, opened 2018-01-17
Core/AI: Fix assertion triggered with PetAI derived scripts

### LINKED ISSUE #21247: [CRASH] .npc add <pet creature id> crash
**Description:**
Executing command ".npc add 19668" and attacking the spawned npc triggers an assert
>D:\wow\TrinityCore\src\server\game\AI\CoreAI\PetAI.cpp:546 in PetAI::CanAttack ASSERTION FAILED:
me->GetCharmInfo()

1. ".npc add 19668"
2. Attack the spawned creature
3. Crash

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. 3d7166c1c78e 2018-01-16 14:44:04 -0300 (3.3.5 branch) (Win64, RelWithDebInfo, Static)

**TDB version:**  335.something

[3d7166c1c78e_worldserver.exe_[16-1_20-42-30].txt](https://github.com/TrinityCore/TrinityCore/files/1636581/3d7166c1c78e_worldserver.exe_.16-1_20-42-30.txt)

jackpoz:
**Changes proposed:**

-  Fix assert triggered when spawning a Creature with AI inheriting from PetAI. The Creature will have a default AI instead.
- Remove unneeded null check since there's a check already few lines above

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #21247

- built locally in Debug mode
- spawned Shadowfiend ".npc add 19668" and worldserver showed the error log. Attacking the NPC didn't trigger any assert.
- cast Shadowfiend spell, no error log appeared (meaning the PetAI is still loaded when allowed)
- tamed a Creature as a Pet (even if this test should be unrelated to the code involved)

none

### DIFF
--- a/src/server/game/AI/AIException.cpp
+++ b/src/server/game/AI/AIException.cpp
@@ -1,19 +1,19 @@
 /*
-* Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
-*
-* This program is free software; you can redistribute it and/or modify it
-* under the terms of the GNU General Public License as published by the
-* Free Software Foundation; either version 2 of the License, or (at your
-* option) any later version.
-*
-* This program is distributed in the hope that it will be useful, but WITHOUT
-* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
-* FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
-* more details.
-*
-* You should have received a copy of the GNU General Public License along
-* with this program. If not, see <http://www.gnu.org/licenses/>.
-*/
+ * Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
 
 #include "AIException.h"
 
--- a/src/server/game/AI/AIException.cpp
+++ b/src/server/game/AI/AIException.cpp
@@ -1,0 +1,24 @@
+/*
+* Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+* Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
+*
+* This program is free software; you can redistribute it and/or modify it
+* under the terms of the GNU General Public License as published by the
+* Free Software Foundation; either version 2 of the License, or (at your
+* option) any later version.
+*
+* This program is distributed in the hope that it will be useful, but WITHOUT
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+* FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+* more details.
+*
+* You should have received a copy of the GNU General Public License along
+* with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "AIException.h"
+
+InvalidAIException::InvalidAIException(char const* const _message)
+{
+    message().assign(_message);
+}
--- a/src/server/game/AI/AIException.cpp
+++ b/src/server/game/AI/AIException.cpp
@@ -1,6 +1,5 @@
 /*
 * Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
-* Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
 *
 * This program is free software; you can redistribute it and/or modify it
 * under the terms of the GNU General Public License as published by the
--- a/src/server/game/AI/AIException.cpp
+++ b/src/server/game/AI/AIException.cpp
@@ -17,7 +17,7 @@
 
 #include "AIException.h"
 
-InvalidAIException::InvalidAIException(char const* const _message)
+InvalidAIException::InvalidAIException(char const* _message)
 {
     message().assign(_message);
 }
--- a/src/server/game/AI/AIException.h
+++ b/src/server/game/AI/AIException.h
@@ -19,6 +19,8 @@
 #define TRINITY_AIEXCEPTION_H
 
 #include "Define.h"
+#include <exception>
+#include <string>
 
 class TC_GAME_API InvalidAIException : public std::exception
 {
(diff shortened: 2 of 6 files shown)


ariel-:
What about passing/checking CharmInfo on PetAI constructor?

jackpoz:
and what do you plan to do in the constructor if CharmInfo is null ?

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -1599,7 +1600,19 @@ CreatureAI* ScriptMgr::GetCreatureAI(Creature* creature)
     ASSERT(creature);
 
     GET_SCRIPT_RET(CreatureScript, creature->GetScriptId(), tmpscript, nullptr);
-    return tmpscript->GetAI(creature);
+    CreatureAI* ai = tmpscript->GetAI(creature);
+
+    // Don't allow PetAI if there's no charm info in the creature
+    if (dynamic_cast<PetAI*>(ai))
```
Treeston: "meh"

can't we make a new virtual method on `CreatureAI` that's called `IsValid` or something?

Treeston (changes requested):

src/server/game/Scripting/ScriptMgr.cpp
Shauren: no, because then you are putting control back in hands of script and the point of this validation is to harden the core against bad scripts

jackpoz: @Shauren what if we do
```c++
virtual bool CreatureAI::IsAIAllowed(Creature* creature) { return true;}
```
and then
```c++
bool PetAI::IsAIAllowed(Creature* creature) final override { return !!creature->GetCharmInfo(); }
```
using final override we should keep the check in the core without allowing any script to override it

Treeston: @Shauren but a script has to _active_ suppress an inherited `IsAIAllowed`, while currently just inheriting from `PetAI` _passively_ causes crashes.

ariel-: Then set PetAi as final so no one can inherit it

Treeston: that'd lead to tons of code duplication

Shauren: how about throwing an exception from PetAI constructor when charminfo is missing?

Treeston: and catching that in `ScriptMgr`? that's reasonable, :+1:

Treeston: if we catch it there we'd be leaking the half-constructed object, no?

ariel-:
What if we fix the shadowfiend script instead? It's the only script inheriting from PetAI anyways

src/server/game/Scripting/ScriptMgr.cpp
Shauren: no, thats not how exceptions from constructors work, you do not return anything, you dont even get a pointer to use with operator delete (tl;dr, dont worry about it)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -10450,7 +10450,7 @@ void Unit::UpdateCharmAI()
                     if (isPossessed() || IsVehicle())
                         i_AI = new PossessedAI(ToCreature());
                     else
-                        i_AI = new PetAI(ToCreature());
+                        i_AI = new PetAI(ToCreature()); // <-- is there always a valid charm info at this point ?
```
jackpoz: is there always a valid charm info at this point or should be handle the exception ?

jackpoz:
@Shauren leaving aside for a moment the fact that I used std::exception instead of a custom one (that should be somehow generic like InvalidAIException or so), is the current code what you had in mind ?
Tested ingame and debugged with breakpoints, both spawning the npc with command and with Priest spell, it worked correctly.

Shauren:
Yes, thats exactly what I meant

shamage:
@Shamage

yes in this sense, now it's okay :) thank you

src/server/game/AI/AIException.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include "Define.h"
+
+class TC_GAME_API InvalidAIException : public std::exception
+{
+public:
+    InvalidAIException(char const* const _message);
+    ~InvalidAIException() throw() { }
+
+    char const* what() const throw() override { return msg_.c_str(); }
```
Shauren: replace `throw()` with `noexcept`

Shauren (changes requested):

src/server/game/AI/AIException.h
```diff
@@ -0,0 +1,39 @@
...
+
+#ifndef TRINITY_AIEXCEPTION_H
+#define TRINITY_AIEXCEPTION_H
+
+#include "Define.h"
+
+class TC_GAME_API InvalidAIException : public std::exception
+{
+public:
+    InvalidAIException(char const* const _message);
```
Shauren: second const is unneccessary here (the one after *)

src/server/game/AI/CreatureAISelector.cpp
```diff
@@ -84,8 +85,16 @@ namespace FactorySelector
...
 
         // scriptname in db
-        if (CreatureAI* scriptedAI = sScriptMgr->GetCreatureAI(creature))
-            return scriptedAI;
+        try
+        {
+            if (CreatureAI* scriptedAI = sScriptMgr->GetCreatureAI(creature))
+                return scriptedAI;
+        }
+        catch (const InvalidAIException& e)
```
Shauren: const after type

src/server/game/AI/AIException.h
```diff
@@ -0,0 +1,39 @@
+/*
+* Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+* Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
```
Shauren: and mangos shouldnt be here, its our own new file

Shauren (approved):

src/server/game/AI/AIException.cpp
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "AIException.h"
+
+InvalidAIException::InvalidAIException(char const* _message)
+{
+    message().assign(_message);
```
Treeston: do we really need a `.cpp` for this? `<string>` is visible in the header

src/server/game/AI/AIException.h
```diff
@@ -0,0 +1,40 @@
...
+class TC_GAME_API InvalidAIException : public std::exception
+{
+public:
+    InvalidAIException(char const* _message);
+    ~InvalidAIException() throw() { }
+
+    char const* what() const noexcept override { return msg_.c_str(); }
+
+protected:
+    std::string & message() throw() { return msg_; }
```
Treeston: is it required that this be a non-const ref?

if not, `msg_` could be a `const` member

src/server/game/AI/AIException.cpp
jackpoz: it always makes sense to declare in .h and then implement in .cpp

Treeston: it doesn't if you're adding a whole new compilation unit for no particular reason...

src/server/game/AI/AIException.h
Treeston: is that a yes or a no? I don't care if it's the same as some other class...

PR #21263, opened 2018-01-19
Core/Spells: Suppresser not channeling (Suppression) and instead attacking players

tje3d:
Suppresser suppression channel interrupt instantly and the npc tries to hit ppl instead.

**Changes proposed:**

- Prevent Suppression interrupt ( i think npc turning was the interruption reason )
- Prevent npc to attack ppl

**Target branch(es):** 3.3.5

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -4194,6 +4194,12 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->RangeEntry = sSpellRangeStore.LookupEntry(157); // 90yd
     });
 
+    // Suppression ( Valithria Dreamwalker )
+    ApplySpellFix({ 70588 }, [](SpellInfo* spellInfo)
+    {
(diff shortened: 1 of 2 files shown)


tje3d:
@tkrokli [Suppresser](http://www.wowhead.com/npc=37863/suppresser) is the correct name xD

tkrokli:
:-P ah, very well. I would have seen it if you had included that link in your PR description.
- http://www.wowhead.com/npc=37863/suppresser

I changed it because auto correct suggested I should do so. TIL - never mind.

tkrokli:
BTW, I think we should find a way to say in the title what is fixed, not what the issue is.

ariel-:
I don't like this, the script worked just fine. The issue must be on core side

tkrokli:
Aha. Maybe it is caused by a threat or movement handling issue in the core.

Aokromes:
this bug still exists?

Killyana:
Yes still happen.

PR #21271, opened 2018-01-20
Core/Item: fix null pointer ref in Item::Create()

### LINKED ISSUE #21256: Enabling AHbot seller on Linux causes Segmentation Fault
**Description:**
Enabling the Ahbot seller causes a Segmentation Fault a few seconds after server initialization has completed.

**Current behaviour:**
Using GDB, I traced this to line 322 in TrinityCore/src/server/game/Entities/Item/Item.cpp.  It currently reads:
```
if (spellInfo->HasEffect(SPELL_EFFECT_GIVE_ARTIFACT_POWER))
                if (uint32 artifactKnowledgeLevel = owner->GetCurrency(CURRENCY_TYPE_ARTIFACT_KNOWLEDGE))
                    SetModifier(ITEM_MODIFIER_ARTIFACT_KNOWLEDGE_LEVEL, artifactKnowledgeLevel + 1);
```
At this point in execution, owner has the value 0x0.  Calling GetCurrency causes the segmentation fault as it tries to access an unordered_map that does not exist.
 
**Expected behaviour:**
 
After looking through the call stack for a good place for a fix, I added a test for a valid owner to Item.cpp, changing line 322 to read:
```
 if (owner && spellInfo->HasEffect(SPELL_EFFECT_GIVE_ARTIFACT_POWER))
                if (uint32 artifactKnowledgeLevel = owner->GetCurrency(CURRENCY_TYPE_ARTIFACT_KNOWLEDGE))
                    SetModifier(ITEM_MODIFIER_ARTIFACT_KNOWLEDGE_LEVEL, artifactKnowledgeLevel + 1);

```
This has fixed the immediate problem.  There is no segmentation fault, and the auction house seller is now working.  I'm new to Trinity, and haven't spent enough time with the code yet to know if this is an architecturally correct fix, so I'm just reporting it as a bug for now. 

Set 
1. Build fresh worldserver on Ubuntu 16.04
2. Set AuctionHouseBot.Seller.Enabled = 1 in worldserver.conf
3. Start worldserver and wait 'till it starts adding items to the AH.  The segmentation fault will occur within a minute or two.

**Branch(es):**
master 

**TC rev. hash/commit:** 
TrinityCore rev. d1bd21233953 2018-01-15 08:43:13 +0100 (master branch) (Unix, Debug, Static) 

**TDB version:**  Using World DB: TDB 720.00

**Operating system:** Ubuntu 16.04

zranger1:
**Changes proposed:**
In /src/server/game/Entities/Item/Item.cpp:
    Add test for NULL owner value before evaluating newly created item for artifact effects.

**Target branch(es):** master

**Issues addressed:** Closes #21256 

**Tests performed:**
Built and ran debug for several hours, then repeated with release+symbols.   The segmentation fault does not occur with either build, test users reported that AH seller is now working. 

> Test system config:
HP Proliant DL360 G6
2x Intel Xeon x5670
80GB RAM
Ubuntu 16.04.3 LTS
GNU GCC v7.2.0, Boost 1.65.1, MySQL 5.7.20, etc

The NULL owner pointer is passed in as a default argument from the ahbot code and `Item::Create` tests its value several times.
- [ ] Look at rearranging the code to minimize the number of tests.

### DIFF
--- a/src/server/game/Entities/Item/Item.cpp
+++ b/src/server/game/Entities/Item/Item.cpp
@@ -319,7 +319,7 @@ bool Item::Create(ObjectGuid::LowType guidlow, uint32 itemId, Player const* owne
         if (i < 5)
             SetSpellCharges(i, itemProto->Effects[i]->Charges);
         if (SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(itemProto->Effects[i]->SpellID))
-            if (spellInfo->HasEffect(SPELL_EFFECT_GIVE_ARTIFACT_POWER))
+            if (owner && spellInfo->HasEffect(SPELL_EFFECT_GIVE_ARTIFACT_POWER))
                 if (uint32 artifactKnowledgeLevel = owner->GetCurrency(CURRENCY_TYPE_ARTIFACT_KNOWLEDGE))
                     SetModifier(ITEM_MODIFIER_ARTIFACT_KNOWLEDGE_LEVEL, artifactKnowledgeLevel + 1);
     }


jackpoz:
Thanks for the PR :)

PR #21282, opened 2018-01-20
Core/Creature: implement automatic distance correction for melee combat

Ovahlord:
The following PR adds support for creatures to correct their distance to their target automaticly when standing too close to each other. According to retail research they use a internal 3 seconds timer to check the distance. That's why the creatures sometimes react very fast and sometimes they need their time.

**Changes proposed:**
-  add support for creatures to reposition theirself when they are too close to their current victim by getting their new contactpoint and moving to there

**Target branch(es):** 3.3.5
- [x] 3.3.5

https://www.youtube.com/watch?v=sy-d8Zl0c5I

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -382,7 +382,7 @@ void CreatureAI::SetBoundary(CreatureBoundary const* boundary, bool negateBounda
     me->DoImmediateBoundaryCheck();
 }
 
-void CreatureAI::CheckDistanceToCurrentVictim()
+void CreatureAI::CheckMeleeRepositionRequirements()
 {
     if (Unit* victim = me->GetVictim())
     {
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -382,6 +382,19 @@ void CreatureAI::SetBoundary(CreatureBoundary const* boundary, bool negateBounda
     me->DoImmediateBoundaryCheck();
 }
 
+void CreatureAI::CheckDistanceToCurrentVictim()
+{
+    if (Unit* victim = me->GetVictim())
+    {
+        Position victimPos;
+        victimPos.Relocate(victim->GetPosition());
+
+        // If we are closer than 50% of the combat reach we are going to reposition ourself
+        if (me->GetDistance(victimPos) < CalculatePct(me->GetCombatReach(), 50))
+            me->MoveAdvanceTo(victim);
+    }
+}
+
 Creature* CreatureAI::DoSummon(uint32 entry, Position const& pos, uint32 despawnTime, TempSummonType summonType)
 {
     return me->SummonCreature(entry, pos, summonType, despawnTime);
--- a/src/server/game/AI/CreatureAI.h
+++ b/src/server/game/AI/CreatureAI.h
@@ -229,7 +229,7 @@ class TC_GAME_API CreatureAI : public UnitAI
         static bool IsInBounds(CreatureBoundary const& boundary, Position const* who);
 
         // Called every 3 Seconds to check if the creature needs to reposition itself
-        void CheckDistanceToCurrentVictim();
+        void CheckMeleeRepositionRequirements();
 
     protected:
         virtual void MoveInLineOfSight(Unit* /*who*/);
--- a/src/server/game/AI/CreatureAI.h
+++ b/src/server/game/AI/CreatureAI.h
@@ -228,6 +228,9 @@ class TC_GAME_API CreatureAI : public UnitAI
 
         static bool IsInBounds(CreatureBoundary const& boundary, Position const* who);
 
+        // Called every 3 Seconds to check if the creature needs to reposition itself
+        void CheckDistanceToCurrentVictim();
+
     protected:
         virtual void MoveInLineOfSight(Unit* /*who*/);
 
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -237,7 +237,7 @@ bool ForcedDespawnDelayEvent::Execute(uint64 /*e_time*/, uint32 /*p_time*/)
 }
 
 Creature::Creature(bool isWorldObject): Unit(isWorldObject), MapObject(), m_groupLootTimer(0), lootingGroupLowGUID(0), m_PlayerDamageReq(0), m_lootRecipient(), m_lootRecipientGroup(0), _pickpocketLootRestore(0),
-    m_corpseRemoveTime(0), m_respawnTime(0), m_respawnDelay(300), m_corpseDelay(60), m_respawnradius(0.0f), m_boundaryCheckTime(2500), m_andvanceMovementTime(3000), m_combatPulseTime(0), m_combatPulseDelay(0), m_reactState(REACT_AGGRESSIVE),
+    m_corpseRemoveTime(0), m_respawnTime(0), m_respawnDelay(300), m_corpseDelay(60), m_respawnradius(0.0f), m_boundaryCheckTime(2500), m_advanceMovementTime(3000), m_combatPulseTime(0), m_combatPulseDelay(0), m_reactState(REACT_AGGRESSIVE),
     m_defaultMovementType(IDLE_MOTION_TYPE), m_spawnId(0), m_equipmentId(0), m_originalEquipmentId(0), m_AlreadyCallAssistance(false), m_AlreadySearchedAssistance(false), m_cannotReachTarget(false), m_cannotReachTimer(0),
     m_AI_locked(false), m_meleeDamageSchoolMask(SPELL_SCHOOL_MASK_NORMAL), m_originalEntry(0), m_homePosition(), m_transportHomePosition(), m_creatureInfo(nullptr), m_creatureData(nullptr), _waypointPathId(0), _currentWaypointNodeInfo(0, 0),
     m_formation(nullptr), m_triggerJustAppeared(true), m_respawnCompatibilityMode(false), m_focusSpell(nullptr), m_focusDelay(0), m_shouldReacquireTarget(false), m_suppressedOrientation(0.0f), _lastDamagedTime(0),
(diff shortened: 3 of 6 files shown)


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -844,6 +852,27 @@ void Creature::Update(uint32 diff)
     }
 }
 
+void Creature::MoveAdvanceTo(Unit* target)
+{
+    if (!target) // Just in case
+        return;
+
+    // Do not reposition ourself when we are not allowed to move
+    if (HasUnitState(UNIT_STATE_CASTING) || isMoving() || !CanFreeMove())
```
Riztazz: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L3336

src/server/game/AI/CreatureAI.h
```diff
@@ -228,6 +228,9 @@ class TC_GAME_API CreatureAI : public UnitAI
 
         static bool IsInBounds(CreatureBoundary const& boundary, Position const* who);
 
+        // Called every 3 Seconds to check if the creature needs to reposition itself
+        void CheckDistanceToCurrentVictim();
```
Shauren: misleading function name - it also repositions

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -470,6 +470,27 @@ void Unit::MonsterMoveWithSpeed(float x, float y, float z, float speed, bool gen
...
+    if (IsMovementPreventedByCasting() || isMoving() || !CanFreeMove())
+        return;
+
+    float x, y, z;
+    GetNearPoint(target, x, y, z, 0.0f, 0.0f, target->GetAngle(this));
+    Movement::MoveSplineInit init(this);
+    init.MoveTo(x, y, z, true, false);
+
+    // Beasts move backwards instead of turning arround
+    if (ToCreature() && ToCreature()->GetCreatureTemplate()->type == CREATURE_TYPE_BEAST)
```
Shauren: Not strictly limited to beasts, more like beast models. Example: http://www.wowhead.com/npc=91847/risen-hound also moves backwards (undead dog)

Shauren (changes requested):

src/server/game/Entities/Unit/Unit.cpp
Ovahlord: Any idea how to detect this?

Shauren: not really atm, maybe there is something in creaturemodel/displayinfo related db2s

Ovahlord: Hmm, I rememember the automatic trigger model detection system is using something like this to determine the triger models. Would that be an acceptable solution?

joschiwald:
is it possible to move this code into the ChaseMovementGenerator?

Ovahlord:
Dunno. I don't think so because this behaivor counts only for uncontrolled creatures. As soon as mindcontrol comes into play or any other AI interfering case this should not happen anymore.

joschiwald:
the movementgenerator can have timers, just define one like other movementgenerators

mindcontrol does not have nothing to do with the ChaseMovementGenerator, it affects the combat movement only

i'm thinking that this "independend system" will affect all movement, including waypoint or idle movement, just because you have a victim

Ovahlord:
Not in my implementation here because I strictly disabled this behaivior for already moving creatures. I have not seen any scenarios with waypoint or idle movement in this regard on retail. This will probably more a pathingfing thing since Blizzard's pathfinding in general has some sort of "forced distance to a collision object"

joschiwald:
what about idle movement or disabled combat movement? they start moving, no?

src/server/game/Entities/Creature/Creature.h
```diff
@@ -380,6 +380,7 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         uint32 m_corpseDelay;                               // (secs) delay between death and corpse disappearance
         float m_respawnradius;
         uint32 m_boundaryCheckTime;                         // (msecs) remaining time for next evade boundary check
+        uint32 m_andvanceMovementTime;                      // (msecs) remaining time for next reposition update to avoid creatures standing inside each other
```
joschiwald: advance

src/server/game/AI/CreatureAI.cpp
```diff
@@ -382,6 +382,19 @@ void CreatureAI::SetBoundary(CreatureBoundary const* boundary, bool negateBounda
...
 
+void CreatureAI::CheckMeleeRepositionRequirements()
+{
+    if (Unit* victim = me->GetVictim())
+    {
+        Position victimPos;
+        victimPos.Relocate(victim->GetPosition());
+
+        // If we are closer than 50% of the combat reach we are going to reposition ourself
+        if (me->GetDistance(victimPos) < CalculatePct(me->GetCombatReach(), 50))
```
joschiwald: me->GetDistance(victimPos) includes me->GetCombatReach(), that means if your exact distance is less than combat reach, it will return 0.0f

funjoker:
This can be used for master too.
Just needs the addition of CombatReach and BoundRadius

src/server/game/AI/CreatureAI.cpp
Ovahlord: Huh? I see      float d = GetExactDist(&pos) - GetObjectSize(); in that function. So where exactly is the combat reach getting taken into account so I can adapt the formular?

joschiwald: see how GetObjectSize() is implemented

Ovahlord: Ye. Did so. I see no problem there.  It was intended to get the distance to be unaffected by combat reach because I wanted to make sure that we take the actual distance rather than checking when the target enters my combat reach. The backwards movement it not suposed to make the target move to the very outer circle radius just like the initial chase movement does. On retail they just move a bit back. This is also why I don't use any range values in the next step to get the contact point.

Ovahlord:
Allright, there is another case that blocks that movement and this is the size of the npc.
The current silithus event for example is spawning giant scarabeus and they do not move back.
So the question here is: how do we detect this. rank is not the case because other elite npc's will move back such as hogger for example.

ccrs:
this needs to be on the movementgenerator, or else two logics will apply at the same time, makes no sense at all.

Ovahlord:
I see but I can't help you with that because I am not famililar with the movement generator systems at all so I am kinda lost with it. I don't wanna touch things that I don't understand because this would be more work for you as well to correct my mistakes over and over again

src/server/game/AI/CreatureAI.cpp
Ovahlord: I don't see a problem there. It works exactly how I wanted it to work: compare distance between the two positions regardless of their combat-reach, if the victim is too close the creature will step back. If I would be a big dragon and a small player attacks me and moves closer it would trigger my repositioning when he comes closer than 50% of my combat-reach as actual distance between those two positions. In the next step I use the ContactPoint helper to get a good distance and so the dragon will move back.

Since the attack range is calculated by using  AttackerCombatreach + VictimCombatreach there is no issue with that because the reposition will place you far enough to prevent a double triggering.

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -237,7 +237,7 @@ bool ForcedDespawnDelayEvent::Execute(uint64 /*e_time*/, uint32 /*p_time*/)
 }
 
 Creature::Creature(bool isWorldObject): Unit(isWorldObject), MapObject(), m_groupLootTimer(0), lootingGroupLowGUID(0), m_PlayerDamageReq(0), m_lootRecipient(), m_lootRecipientGroup(0), _pickpocketLootRestore(0),
-    m_corpseRemoveTime(0), m_respawnTime(0), m_respawnDelay(300), m_corpseDelay(60), m_respawnradius(0.0f), m_boundaryCheckTime(2500), m_combatPulseTime(0), m_combatPulseDelay(0), m_reactState(REACT_AGGRESSIVE),
+    m_corpseRemoveTime(0), m_respawnTime(0), m_respawnDelay(300), m_corpseDelay(60), m_respawnradius(0.0f), m_boundaryCheckTime(2500), m_advanceMovementTime(3000), m_combatPulseTime(0), m_combatPulseDelay(0), m_reactState(REACT_AGGRESSIVE),
```
ariel-: I'm starting to think Creature should have it's own EventMap/EventProcessor/TaskScheduler to handle all timers

src/server/game/AI/CreatureAI.cpp
joschiwald: in case of your last picture (bottom right):
big combat reach: 10
small combat reach: 1
-> exact dist: 11
-> distance `float d = GetExactDist(&pos) - GetObjectSize()`: 11 - 10 = 1
-> 1 < 10 * 50%
=> repositioning
or I'm wrong?

Ovahlord: It is more like:
-> distance between point 1 and point 2 (ignoring combatreach entirely)
-> distance < AttackerCombatreach * 50%
--> if yes: reposition

It's a bit tricky how I applied that handling. I do not compare the combat-reach distance from those creatures. I use the attacker combatreach only to know when the creatures comes too close. Not by using combatreach, more by using exact positions.

src/server/game/AI/CreatureAI.h
```diff
@@ -228,6 +228,9 @@ class TC_GAME_API CreatureAI : public UnitAI
 
         static bool IsInBounds(CreatureBoundary const& boundary, Position const* who);
 
+        // Called every 3 Seconds to check if the creature needs to reposition itself
```
offl: On retail this time is about 1s, maybe less. Maybe it could be changed?

Ovahlord: My tests on retail have clearly shown me that it is more like 3 seconds. Maybe you went close to the creature when the timer has been started already

offl: Probably it was changed, but I don't know when. Maybe in TBC or WotLK.
Vanilla: https://youtu.be/SwLjJH8LXfc?t=566 https://youtu.be/SwLjJH8LXfc?t=161

src/server/game/Entities/Unit/Unit.cpp
Nitair: @Shauren are you sure thats also for WotLk? Your URL refers to an ID which indicates it was added after WotLk (exactly at patch 7.0.3.22290). I don't remember any creature which was an exception in this case (in WotLk).

Nitair (approved):

Ovahlord:
I am going to rework this PR soon (tm). Moved the checks to targetMovementGenerator and added some new exceptions and improvements.

CraftedRO:
any planns to "revive" this ?

Issue #20106, opened 2017-08-03
[3.3.5] - Random equipped items disappears


Jildor:
**Description:**

I'm not sure what happens, but I have a lot of players reports months ago, the equipped items disappears when log in or log out.
Not related with crash or rollback.
The item disappear in DB from character_inventory but still in item_instance and not appear in mail_items

Anyone know anything about this problem??

In this rev the problem does not happen: d92d5c0d17c7bd3b03522c6146ba91d3adaea223

And here I nothiced the problem appear: 73ec3a1d3b34b31ae67ea4215ce3d25a2adf392b

**Steps to reproduce the problem:**

By the moment I don't know how to reproduce it.
Happens when players logout or login, I don't know

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 445c5a00e8

**TDB version:**  TDB335.63

**Operating system:** Debian

Demonid:
I got this issue too, really annoying, from time to time items start to disappear from the inventory, i didn't check the DB, but this issue is for real.

Keader:
I'm having the same issue too.

Jildor:
Yep, I don't know how is the cause...no idea...but I think this need priority

Takenbacon:
Log the output of the `sql.sql` logger in case it's some sort of database error.

https://github.com/TrinityCore/TrinityCore/blob/master/src/server/database/Database/MySQLConnection.cpp#L230

Mack36:
Try to enable(if not yet) some worldserver logs, AFAIK there should be smth related to items loading in those:
- #Logger.entities.player.items
- #Logger.entities.player.loading
- #Logger.entities.player.character

Jildor:
I found a lot deadlocks related with character_inventory in the server.log
example
```
2017-08-06_00:56:33 SQL(p): REPLACE INTO character_inventory (guid, bag, slot, item) VALUES (138847, 14517911, 3, 267840328)
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
2017-08-06_00:56:33 SQL(p): REPLACE INTO character_inventory (guid, bag, slot, item) VALUES (138847, 14517911, 5, 267840390)
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
2017-08-06_00:56:33 SQL(p): REPLACE INTO character_inventory (guid, bag, slot, item) VALUES (138847, 14517911, 4, 267840362)
[...]

PR #21296, opened 2018-01-23
[3.3.5] Core/Database: attempt to fix 'Random equipped items disappears'

Langerz82:
**Changes proposed:**
Adds a random delay before retrying the transaction.  

- [x] 3.3.5

**Issues addressed:** Closes no issue (but targets #20106), just hopefully minimizes deadlocks.

**Tests performed:** Cannot test as do not have sufficient server population.

- [ ] Check if it works OK
- ASSERT Deadlock SQL Transaction should not appear if working OK
- items should not disappear


src/server/database/Database/Transaction.cpp
```diff
@@ -67,16 +69,26 @@ bool TransactionTask::Execute()
...
+    {
+        m_trans->Cleanup();
         return true;
+    }
 
     if (errorCode == ER_LOCK_DEADLOCK)
     {
         // Make sure only 1 async thread retries a transaction so they don't keep dead-locking each other
-        std::lock_guard<std::mutex> lock(_deadlockLock);
+        //std::lock_guard<std::mutex> lock(_deadlockLock);
```
jackpoz: why did you comment this line ?

src/server/database/Database/Transaction.cpp
```diff
@@ -67,16 +69,26 @@ bool TransactionTask::Execute()
...
         uint8 loopBreaker = 5;  // Handle MySQL Errno 1213 without extending deadlock to the core itself
         for (uint8 i = 0; i < loopBreaker; ++i)
+        {
+            std::this_thread::sleep_for(std::chrono::milliseconds(24 + rand() % 24));
```
jackpoz: how do you expect randomness to fix a issue ? just by pure luck ?

src/server/database/Database/Transaction.cpp
```diff
@@ -67,16 +69,26 @@ bool TransactionTask::Execute()
...
             if (!m_conn->ExecuteTransaction(m_trans))
+            {
+                m_trans->Cleanup();
                 return true;
+            }
+        }
+        ASSERT(false, "Deadlock SQL Transaction");
```
jackpoz: shutting down the whole worldserver just for a deadlock sounds too much

jackpoz (changes requested):
I left some comments that I'd like to discuss before merging this PR

src/server/database/Database/Transaction.cpp
jackpoz: reading more in detail https://github.com/TrinityCore/TrinityCore/issues/20106 I would agree that giving more time to other SQL transactions to finish should help resolve the deadlock, I would prefer to change the loop from looping only 5 times to loop for a configurable amount of time, default to 30/60 seconds

src/server/database/Database/Transaction.cpp
Langerz82: Ariel said: "static analysis tools will complain for sleeping (and with reason!) with a locked mutex" so I commented it out.

jackpoz:
> Cannot test as do not have sufficient server population.

I created https://github.com/jackpoz/BotFarm for these kind of tests

src/server/database/Database/Transaction.cpp
Langerz82: The randomness was added just in case 2 threads are conflicting at the same time that they are offset from each other, and given time to complete.

src/server/database/Database/Transaction.cpp
Langerz82: It shutdowns the server from the assert? How do I make it just a warning?

src/server/database/Database/Transaction.cpp
Langerz82: can we add a sleep with the lock enabled, just so it doesn't hammer SQL requests too often.

src/server/database/Database/Transaction.cpp
jackpoz: It doesn't work like that, you don't edit the code to make static analysis warnings disappear. This is a very delicate part of the core that requires a lot of attention so every modified line requires to fully understand what's happening.

The mutex is there to make sure 2 different threads deadlocking on the same SQL rows will not keep trying again at the same time to submit the transaction, resulting again in a deadlock.

I would change the loop counter to a timed loop that keeps trying for a fixed amount of time instead without the thread_sleep

src/server/database/Database/Transaction.cpp
jackpoz: we use TC_LOG_ERROR("category", "message") macro, for example TC_LOG_ERROR("sql.sql", "Deadlock SQL Transaction") .
I'm not sure if current dependency allow to use logs in this file.

Speaking of logs, I would like to log much more info on each retry (how much time it passed since first deadlock, what number of retry are we at, thread ID, etc).

Thanks for the heads-up about the issue in this code, I'll give it some care in the next days when I find time :)

Langerz82:
@jackpoz ok cool, thanks for the heads up about all mentioned. I'll close off this PR.

jackpoz:
Let it stay open so it'll remind me about it :)
Also feel free to start implementing what I wrote above if you have time

Langerz82:
Hey, I made those changes you recommended please see my patch:
https://gist.github.com/Langerz82/56e9569617c4af871e2f1944f028ca6a
If all good ill create a PR.

Keader:
You dont need open a new pr, you can just make your changes here ^^

Langerz82:
New PR made.

src/server/database/Database/Transaction.cpp
```diff
@@ -63,20 +67,51 @@ void Transaction::Cleanup()
...
+        case SQL_ELEMENT_PREPARED:
+            msg += data.element.stmt->getQueryString() + ';';
+            break;
+        case SQL_ELEMENT_RAW:
+            msg += data.element.query + ';';
+            break;
+        }
+
+    }
+    return msg.c_str();
```
Rochet2: Return the std::string instead to avoid crashes. Use c_str when you print the value only.

src/server/database/Database/Transaction.cpp
```diff
@@ -63,20 +67,51 @@ void Transaction::Cleanup()
...
 }
 
+const char* Transaction::ToString()
+{
+    std::string msg;
+    for (SQLElementData const& data : m_queries)
+    {
+        switch (data.type)
+        {
+        case SQL_ELEMENT_PREPARED:
```
Jildor: Indentation

src/server/database/Database/Transaction.cpp
```diff
@@ -63,20 +67,51 @@ void Transaction::Cleanup()
...
-        uint8 loopBreaker = 5;  // Handle MySQL Errno 1213 without extending deadlock to the core itself
-        for (uint8 i = 0; i < loopBreaker; ++i)
+        uint32 oldMSTime = getMSTime();
+        uint32 loopDuration = 0;
+        uint32 loopExpireTime = STATEMENT_DURATION;
+        for (; loopDuration <= loopExpireTime; loopDuration = GetMSTimeDiffToNow(oldMSTime))
+        {
             if (!m_conn->ExecuteTransaction(m_trans))
+            {
+                TC_LOG_INFO("deadlock.txt", "Deadlock SQL Transaction: delay:%d, SQL:%s", loopDuration, m_trans->ToString());
```
ariel-: This is not how logs work, you don't specify a filename

sirikfoll:
The other loop in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/database/Database/DatabaseWorkerPool.cpp#L278 doesn't need this change too?

src/server/database/Database/Transaction.cpp
jackpoz: This is not the time to check indentation anyway

tkrokli: @Jildor : `case` / `public` / `private` are labels and don't necessarily need to be indented.

src/server/database/Database/Transaction.cpp
```diff
@@ -63,20 +67,51 @@ void Transaction::Cleanup()
...
+    std::string msg;
+    for (SQLElementData const& data : m_queries)
+    {
+        switch (data.type)
+        {
+        case SQL_ELEMENT_PREPARED:
+            msg += data.element.stmt->getQueryString() + ';';
+            break;
+        case SQL_ELEMENT_RAW:
+            msg += data.element.query + ';';
```
tkrokli: From travis-ci :
```cpp
src/server/database/Database/Transaction.cpp:81:39:
 fatal error: adding 'char' to a string pointer does not append to the string [-Wstring-plus-char]
            msg += data.element.query + ';';
                   ~~~~~~~~~~~~~~~~~~~^~~~~
 note: use array indexing to silence this warning
            msg += data.element.query + ';';
                                      ^
                   &                  [    ]
```
(this warning is from debug mode, so Travis treats it as "fatal error")

Langerz82: So exciting. ^__^ I'll make those changes and release very soon.

Langerz82:
ok new PR done

Langerz82:
Our power here is going off today so I wont be able to check messages in 1.5 hours or so.

tkrokli:
Please check https://travis-ci.org/TrinityCore/TrinityCore/builds/332480718?utm_source=github_status&utm_medium=notification and use `Ctrl-F` to search for `fatal`

Langerz82:
@tkrokli my bad I forgot to include that change in my source.

Langerz82:
@tkrokli as far as I am aware:
```
DatabaseWorkerPool<T>::DirectCommitTransaction(SQLTransaction& transaction)
```
Is never called. But it doesn't hurt to add the safety mechanisms just in case.

tkrokli:
OK, fair enough. Travis is there for you and us to see how the PR source handles during debug, so we don't add code to the main branch if it contains errors in debug mode.

jackpoz:
> The other loop in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/database/Database/DatabaseWorkerPool.cpp#L278 doesn't need this change too?

I would prefer to keep the 2 things separated, focusing first on the timeout thing and then applying the solution to the sync queries too (which should be debugged and tested)

src/server/database/Database/Transaction.cpp
```diff
@@ -63,20 +67,51 @@ void Transaction::Cleanup()
...
-        uint8 loopBreaker = 5;  // Handle MySQL Errno 1213 without extending deadlock to the core itself
-        for (uint8 i = 0; i < loopBreaker; ++i)
+        uint32 oldMSTime = getMSTime();
+        uint32 loopDuration = 0;
+        uint32 loopExpireTime = STATEMENT_DURATION;
+        for (; loopDuration <= loopExpireTime; loopDuration = GetMSTimeDiffToNow(oldMSTime))
+        {
             if (!m_conn->ExecuteTransaction(m_trans))
+            {
+                TC_LOG_INFO("sql.sql", "Deadlock SQL Transaction: delay:%d, SQL:%s", loopDuration, m_trans->ToString().c_str());
```
jackpoz: at this point the transaction got executed successfully, this log should be moved outsite of the if() case

src/server/database/Database/Transaction.cpp
```diff
@@ -63,20 +67,51 @@ void Transaction::Cleanup()
...
+                m_trans->Cleanup();
                 return true;
+            }
+        }
+        TC_LOG_ERROR("sql.sql", "Deadlock SQL Transaction: delay:%d, SQL:%s", loopDuration, m_trans->ToString().c_str());
```
jackpoz: this log should state more in detail that the deadlock was fatal and the transaction couldn't be completed even after waiting all that time

src/server/database/Database/Transaction.cpp
```diff
@@ -63,20 +67,51 @@ void Transaction::Cleanup()
...
+    }
+    return msg;
+}
+
 bool TransactionTask::Execute()
 {
     int errorCode = m_conn->ExecuteTransaction(m_trans);
     if (!errorCode)
+    {
+        m_trans->Cleanup();
```
jackpoz: could you explain this change more in detail ?

src/server/database/Database/Transaction.cpp
Langerz82: It should report that there was a deadlock, but it was successful, as we need to know if deadlocks are occurring and where we may have to fix the code.

src/server/database/Database/Transaction.cpp
Langerz82: Agree more info needs to be provided.

src/server/database/Database/Transaction.cpp
Langerz82: I have no idea about this, but I assumed it would be beneficial to free up the memory on even successful transactions. Is this not the case?

jackpoz: I think this PR currently has few changes that are not needed. It's a good proof of concept, I'll polish and test it this weekend when I should have some free time

jackpoz:
@Langerz82 as a rule of thumb please create a branch when opening a PR, it makes it easier for everyone else to open it locally

src/server/database/Database/Transaction.cpp
jackpoz: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/database/Database/Transaction.h#L39

Cleanup() is called automatically by the destructor

src/server/database/Database/Transaction.cpp
jackpoz: deadlocks are already logged
>[ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction

src/server/database/Database/DatabaseWorkerPool.cpp
```diff
@@ -35,6 +35,9 @@
 #endif
 #include <mysql.h>
 #include <mysqld_error.h>
+#include <Timer.h>
+
+#define STATEMENT_DURATION 60000;
```
jackpoz: don't add a ";" in defines

jackpoz:
Slightly changed the async code, just need to test it and then it can be merged :)

jackpoz:
>11 commits into TrinityCore:3.3.5 from unknown repository

Continuing on https://github.com/TrinityCore/TrinityCore/pull/21386

Issue #21291, opened 2018-01-22
Quest "No Mere Dream" The Escort NPC Does Not Start Moving


tje3d:
**Description:**

"Clintar Dreamwalker's Spirit" Does not start moving when a player take "No Mere Dream" quest.
Seems like a movement system issue.

**Expected behaviour:**

"Clintar Dreamwalker's Spirit" is Escort NPC and should walk his path

**Steps to reproduce the problem:**

1. .go creature 42360
2. take "No Mere Dream" quest

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 514c847

**TDB version:** TDB 335.63

**Operating system:** Ubuntu 17.10

Killyana:
Same issue as https://github.com/TrinityCore/TrinityCore/issues/20310

Issue #21042, opened 2017-12-10
DK Minecart Quest Issue


sigbinium:
Description:
http://www.wowhead.com/quest=12701/massacre-at-lights-point

Current Behaviour:
Scarlet Miner wont say his dialogue and drag the minecart towards the ships.

Expected Behaviour:
Scarlet Miner says his dialogue, drags cart towards the ships at Light's Point.

Steps to Reproduce the Problem:
1. Clone repository
2. Compile
3. Run Server

Branch: 3.3.5a

TC Rev. Hash/Commit: 54d0b31 

TDB Version: 335.63

OS: Windows 8.1

Aokromes:
..... search! we closed a ton of duplicate tickets of this.

sigbinium:
My bad senpai! Been working 16 hour shifts lately but will search.

tkrokli:
**Core/Movement: waypoint system issues #20310**
No need to search further about this issue. It has been pointed out before that this specific issue (as well as related ones) are caused by a _general waypoint issue_ so you don't need to open new issue tickets about it or add more info about it in the existing ones. It will be fixed when the waypoint system issues are solved.

PR #21313, opened 2018-01-25
Core/ScriptedAI: skip STATE_ESCORT_NONE in EscortAI::JustAppeared()

tje3d:
**Target branch(es):** 3.3.5

**Issues addressed:** #21291 #21042 ( It may also solve some other issues )

Fixed one the problem listed in: #20310


tkrokli:
Interesting solution as well as the simplicity of it.
Have you checked that nothing breaks because of this change?

ccrs:
wrong, look at the related "discussion" in other issues, whats needed to be done is already known

tje3d:
It works and didn't break anything else, so i count it as a good temporary fix

tkrokli:
No, just wait until the weekend, need some more feedback first.

tje3d:
closed e4e6e22

PR #21317, opened 2018-01-27
Core/Spells: Add flags to spell_area

### LINKED ISSUE #20934: spell_area QUEST_STATUS_REWARDED update
**Description:**
When player rewards a quest, related spell_area auras are immediately updated. In some cases, it is unwanted behavior to update auras right after quest is rewarded. Instead, they should be updated after re-entering area. (it was working like this some time ago)
Similar discussion was [here](https://github.com/TrinityCore/TrinityCore/pull/18146), related to phase_area update after QUEST_STATUS_REWARDED
[Video](https://youtu.be/LLatvrnML2g?t=1m24s) - after rewarding quest, the audience NPCs are there clapping and cheering. On TC they disappear because auras are updated.

**Current behaviour:**
Player rewards a quest and auras are immediately updated.

**Expected behaviour:**
In quests like [Stronger Than Reeds](http://www.wowhead.com/quest=29662), [Stronger Than Wood](http://www.wowhead.com/quest=29771) and [Stronger Than Bone](http://www.wowhead.com/quest=29782) it is necessary to keep the zone auras active until player leaves the area.
New item in disables table could solve this issue.

1. go to the quest giver and reward quest
2. see that all NPCs that should stay there until player leaves area now disappear immediately

**Branch(es):** master

**TC rev. hash/commit:** b12b37058c863bcfecfefc965da45f31b604ee3b

**TDB version:**  TDB 720.00

funjoker:
If the player is in zone/area and the entry has flag `SPELL_AREA_FLAG_AUTOREMOVE` then it will be removed (SendQuestUpdate).

**If the player is leaving zone/area it is always removed no matter if entry has flag. (I think this is the most blizzlike method)**

The flag SPELL_AREA_FLAG_AUTOCAST is handling same way like autocast bool.

**Changes proposed:**

-  Adding flags to control behavior
- Remove autocast bool (uses flag instead)
- Chance to add more flags to specify it even more

MYSQL PART:
```sql
ALTER TABLE `spell_area` CHANGE COLUMN `autocast` `flags` TINYINT(3) UNSIGNED NOT NULL DEFAULT '3' AFTER `gender`;

UPDATE `spell_area` SET `flags` = `flags` | 2;
```

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes #20934

### DIFF
--- a/src/server/game/Spells/SpellMgr.h
+++ b/src/server/game/Spells/SpellMgr.h
@@ -450,6 +450,12 @@ class TC_GAME_API PetAura
 };
 typedef std::map<uint32, PetAura> SpellPetAuraMap;
 
+enum SpellAreaFlag
+{
+    SPELL_AREA_FLAG_AUTOCAST   = 0x1, // if has autocast, spell is applied on enter
+    SPELL_AREA_FLAG_AUTOREMOVE = 0x2, // if has autoremove, spell is remove automatically inside zone/area (allways removed on leaving area or zone)
+};
+
 struct TC_GAME_API SpellArea
 {
     uint32 spellId;
@@ -461,7 +467,7 @@ struct TC_GAME_API SpellArea
     Gender gender;                                          // can be applied only to gender
     uint32 questStartStatus;                                // QuestStatus that quest_start must have in order to keep the spell
     uint32 questEndStatus;                                  // QuestStatus that the quest_end must have in order to keep the spell (if the quest_end's status is different than this, the spell will be dropped)
-    bool autocast;                                          // if true then auto applied at area enter, in other case just allowed to cast
+    uint8 flags;                                            // if SPELL_AREA_FLAG_AUTOCAST then auto applied at area enter, in other case just allowed to cast || if SPELL_AREA_FLAG_AUTOREMOVE then auto removed inside area (will allways be removed on leaved even without flag)
 
     // helpers
     bool IsFitToRequirements(Player const* player, uint32 newZone, uint32 newArea) const;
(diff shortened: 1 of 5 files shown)


sql/updates/world/master/9999_99_99_99_world .sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `spell_area` CHANGE COLUMN `autocast` `flags` TINYINT(3) UNSIGNED NOT NULL DEFAULT '3' AFTER `gender`;
+UPDATE `spell_area` SET `flags` = `flags` + 2;
```
joschiwald:
```sql
UPDATE `spell_area` SET `flags` = `flags` & 2;
```

Shauren: flags | 2, its intended here that they are added

joschiwald: yeah, thats what i meant ^^

funjoker: Ya ok :D.

Shauren (approved):

src/server/game/Spells/SpellMgr.h
```diff
@@ -450,6 +450,12 @@ class TC_GAME_API PetAura
 };
 typedef std::map<uint32, PetAura> SpellPetAuraMap;
 
+enum SpellAreaFlag
+{
+    SPELL_AREA_FLAG_AUTOCAST = 0x1, // if has autocast, spell is applied on enter
```
tkrokli: cosmetic/codestyle: align with the line below by adding 3 spaces before `= 0x1`

ghost (approved):

PR #21319, opened 2018-01-27
[WIP] [335] Core/AI: correct InitializeAI - JustAppeared duplicated logic

ccrs:
Removing the duplicated logic/calls from the core.

InitializeAI -> called after AI creation, helper that allows modifications/calls that may not be suitable before.
JustAppeared -> called everytime after a creature appears (respawn, grid load, etc), standard point to call AI Reset.

Note: The only two inherits from UnitAI are CreatureAI and PlayerAI, so its safe to remove the InitializeAI implementation from UnitAI.

Scripts and AIs will be adjusted slowly, but eventually this will be merged, heads up for anyone interested.

- [X] 3.3.5
- [ ] master

tagging several issues, but on top of that, trying to stop the nonesense blaming.

### DIFF
--- a/src/server/game/AI/CoreAI/UnitAI.cpp
+++ b/src/server/game/AI/CoreAI/UnitAI.cpp
@@ -41,12 +41,6 @@ void UnitAI::AttackStart(Unit* victim)
     }
 }
 
-void UnitAI::InitializeAI()
-{
-    if (!me->isDead())
-        Reset();
-}
-
 void UnitAI::AttackStartCaster(Unit* victim, float dist)
 {
     if (victim && me->Attack(victim, false))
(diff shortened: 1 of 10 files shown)


ccrs:
GameObjectAI out of this, since it doesn't have any "appeared" ai hook.

Shauren:
before you continue
correct existing sai scripts if you go with this change

ccrs:
thats the idea

ccrs:
well

```
[20:22:35] <%ccrs> doing Reset on the next tick
[20:22:49] <%ccrs> can open nice fuckery
[20:23:19] <%ccrs> like creature spawn -> LOS -> aggro with Reset not called
[20:40:09] <%ccrs> LOS is checked right after AddToWorld, except for summons (called after InitSummon)
[20:40:13] <%ccrs> meeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeh
[20:40:17] <%ccrs> what to do here
[20:41:05] <%ccrs> back to InitializeAI as the Reset holder
[20:41:10] <%ccrs> feels like it
```

closing then

PR #21320, opened 2018-01-28
Fixes wrong condition in SMART_EVENT_MOVEMENTINFORM

tje3d:
**Changes proposed:**

Fixes a lot of Quest's/NPC's issues that are using this SMART_EVENT type.

You cant test the changes on this quest ( which is not working now ): [QuestLink](http://www.wowhead.com/quest=12069/return-of-the-high-chief)

**Target branch(es):** 3.3.5


Shauren:
Are you sure about this? Previous logic was `e.event.movementInform.type = 0` then allow ANY type
Your new logic disallows using `e.event.movementInform.type = 0` in database (if this is intended then you should all validation log at startup and skip checking it during execution)

tje3d:
@Shauren Ah you right, but the i'm pretty sure about the id.

tje3d:
Never mind, i was completely wrong.

PR #21321, opened 2018-01-28
Scripts/ICC: Rotface Ooze size growth

PatricNox:
**Changes proposed:**

-  Implement the spell usage for ooze growth, occuring on large-ooze merges
  

**Target branch(es):**

 3.3.5

**Tests performed:** 

Tested in game, 2018-01-28, revision 52fc5a572b
========== Build: 5 succeeded, 0 failed, 11 up-to-date, 2 skipped ==========

### DIFF
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_rotface.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_rotface.cpp
@@ -64,6 +64,7 @@ enum Spells
     SPELL_UNSTABLE_OOZE_EXPLOSION           = 69839,
     SPELL_STICKY_OOZE                       = 69774,
     SPELL_UNSTABLE_OOZE_EXPLOSION_TRIGGER   = 69832,
+    SPELL_OOZE_GROW                         = 74996,
 
     // Precious
     SPELL_MORTAL_WOUND                      = 71127,
@@ -655,13 +656,18 @@ class spell_rotface_large_ooze_buff_combine : public SpellScriptLoader
 
             void HandleScript(SpellEffIndex /*effIndex*/)
             {
-                if (!(GetHitCreature() && GetHitCreature()->IsAlive()))
+                Unit* caster = GetCaster();
+
+                if (!(GetHitCreature() && GetHitCreature()->IsAlive() && caster))
                     return;
 
-                if (Aura* unstable = GetCaster()->GetAura(SPELL_UNSTABLE_OOZE))
+                if (Aura* unstable = caster->GetAura(SPELL_UNSTABLE_OOZE))
                 {
                     uint8 newStack = uint8(unstable->GetStackAmount()+1);
                     unstable->SetStackAmount(newStack);
+
+                    if (unstable->GetStackAmount() > 2)
+                        caster->SetAuraStack(SPELL_OOZE_GROW, caster, unstable->GetStackAmount() - 2);
 
                     // explode!
                     if (newStack >= 5)


src/server/scripts/Northrend/IcecrownCitadel/boss_rotface.cpp
```diff
@@ -655,14 +656,19 @@ class spell_rotface_large_ooze_buff_combine : public SpellScriptLoader
 
             void HandleScript(SpellEffIndex /*effIndex*/)
             {
-                if (!(GetHitCreature() && GetHitCreature()->IsAlive()))
+                Unit* caster = GetCaster();
+
+                if (!(GetHitCreature() && GetHitCreature()->IsAlive() && caster))
```
joschiwald: caster never null here

Shauren:
Made me revisit old sniffs for this
There is no such mechanic, they never change scale.

PatricNox:
@Shauren they should actually, watch any retail video featuring a gameplay with the rotface encounter and you'll see.

I cannot find any writings confirming it, but just as aforementioned you'll find them growing in any retail video. Go PTR on Retail now, if you can, to confirm (?)

src/server/scripts/Northrend/IcecrownCitadel/boss_rotface.cpp
PatricNox: Thanks!

src/server/scripts/Northrend/IcecrownCitadel/boss_rotface.cpp
```diff
@@ -655,14 +656,19 @@ class spell_rotface_large_ooze_buff_combine : public SpellScriptLoader
...
                     return;
 
-                if (Aura* unstable = GetCaster()->GetAura(SPELL_UNSTABLE_OOZE))
+                if (Aura* unstable = caster->GetAura(SPELL_UNSTABLE_OOZE))
                 {
                     uint8 newStack = uint8(unstable->GetStackAmount()+1);
                     unstable->SetStackAmount(newStack);
 
+                    if (unstable->GetStackAmount() > 2)
+                        caster->SetAuraStack(SPELL_OOZE_GROW, caster, unstable->GetStackAmount() - 2);
```
astrixoblix: Shouldn't both `unstable->GetStackAmount()` be `newStack`

Shauren:
Tested on retail 5 min ago, no size change.

PatricNox:
because infection doesn't apply on MT

Shauren:
thats false actually (and the script follows this mechanic) - infection doesnt apply to players hit by SPELL_GREEN_ABOMINATION_HITTIN__YA_PROC (a few seconds)
you can spawn the ooze when solo (which is what i did for this test)

PatricNox:
Oh I see, well I could not reproduce to spawn more than two oozes by using just one character.

Had to have 2 chars to make it possible to see oozes being merged as if it was gameplay.

PatricNox:
this is the most reliable non-text source I could obtain within my time limit..
https://www.youtube.com/watch?v=0V_rk3FGFNw&t=93s

Shauren:
Definitely odd but I checked sniffs made during 3.3.0, 3.3.5, 4.3.4 and even verified on live realms (7.3.5) and none of that changed size

PatricNox:
See if you can find anything with that used spell ID, might be a connection somewhere.

It may also only be reproducable as duo or more, as discussed

Shauren:
Maybe they only grow on 10 man? I tested 25man versions and only found aaab7ba9f4e08ba5b7d7ac2f4e37e07d80c8f339

PatricNox:
Doubtfully, as the idea behind it seems to be "growing till they explode".

I really do not have any other source than retail videos to go with, so unless it suffices or anyone drops by with a written statement to confirm it then I guess it's all for nothing..

src/server/scripts/Northrend/IcecrownCitadel/boss_rotface.cpp
PatricNox: newstack starts 1,
don't want spell to be executed unless an actual merge with the big ooze occurs

PatricNox:
https://github.com/TrinityCore/TrinityCore/blob/aaab7ba9f4e08ba5b7d7ac2f4e37e07d80c8f339/sql/updates/world/3.3.5/2018_01_28_03_world.sql#L5

Shauren:
in that context "growing" = "becoming more", not growing in literal sense

Shauren:
Bonus screenshots at 2 and 3 stacks
wowscrnshot_013118_225817
wowscrnshot_013118_225826

Issue #21255, opened 2018-01-18
Core/Movement/Maps: Spells ignoring textures


Keader:
**Description:**  Fear movement is broken. Blink Destination is broken. Shadow step destination is broken. Charge/Intersept destination is broken

**Current behaviour:** Players affected by fear, just ignore map. They start fly, enter in textures, a real mess. Maybe MMaps/Vmaps issue?
Same to Blink, Shadow step, Charge/Intersept.

**Expected behaviour:** Players affected by fear should run around like a normal player. Players teleported by Blink should not drop of map ( same in Shadow Step/Charge )

**Steps to reproduce the problem:**

1. I tested in Warsong map
2. just add Fear aura in some place, and see how broken it is (.aura 5782)

Check: https://youtu.be/-V0H3wx1TU8 (fear)
Check2: https://youtu.be/Z75TuzmaKj4 (blink)

PS: Vmaps/MMaps is active

Edit: Bug intoduced by https://github.com/TrinityCore/TrinityCore/commit/e42903ec16f4849b71bc8736e3151a1e11d3cd25

**Branch(es):** 3.3.5

**TC rev. hash/commit:** ccf0807be0974dfb8080cb2fc6ea616cc96f062c

Keader:
Hapening with Shadowstep too.

Persephone:
Confirmed, really ancient bug

Keader:
With fear is ancient, but now alot spells is droping you of map

Foereaper:
This happens with Mage Blink and Warrior Charge as well, not sure if it is relevant to the above issue, but I have noticed it happening more recently than before.

Jildor:
Maybe @Golrag commit
https://github.com/TrinityCore/TrinityCore/commit/e42903ec16f4849b71bc8736e3151a1e11d3cd25

Keader:
No @Jildor  already checked it.
I reverted localy https://github.com/TrinityCore/TrinityCore/commit/95456ab5d8bd623481d315ae55dfbb44b45ba9f5 and https://github.com/TrinityCore/TrinityCore/commit/e42903ec16f4849b71bc8736e3151a1e11d3cd25 and is not the cause.

Jildor:
@Keader can you try to revert  @jackpoz mmaps commits? like this: https://github.com/TrinityCore/TrinityCore/commit/f8fd0b7a2c0a9acd084f011c614d529e49397479

Keader:
[...]

PR #21322, opened 2018-01-28
Core/Entities: Reduce the probability of units dropping under the map

Golrag:
**Changes proposed:**

-  Attempt to fix some undermap issues due to selecting a bad height from Map::GetHeight in UpdateAllowedPositionZ when calling MovePositionToFirstCollision.
- Some "clean up" by moving things to their own function.

- 3.3.5
- master

**Issues addressed:**
Updates #21255
maybe more.

Lot's of blink casts & fears in warsong gulch, some blink casts in dragonblight etc, will provide small videos soon.

| TestCase | rev |
| :--- | :---: |
| https://puu.sh/zbJFa/594a5df317.mp4 | https://github.com/TrinityCore/TrinityCore/pull/21322/commits/8f64004c86db1ae4605c735c7ec9376a8a8f481c |
| https://puu.sh/zbJPc/32412fe11e.mp4 | https://github.com/TrinityCore/TrinityCore/pull/21322/commits/8f64004c86db1ae4605c735c7ec9376a8a8f481c |
| https://puu.sh/zfXN0/96a4301c2b.mp4 | https://github.com/TrinityCore/TrinityCore/pull/21322/commits/10ed574a94a62cea423604c54d334c24af7df27b |
| https://puu.sh/zfYdX/b2beb6c1cb.mp4 | https://github.com/TrinityCore/TrinityCore/pull/21322/commits/10ed574a94a62cea423604c54d334c24af7df27b |

- [ ] it's a fucking mess, there should be a way easier way to do this I think that I was unable to see.
- [ ] It's not finished  (blink/fear will sometimes target something too low/high)

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2317,7 +2317,8 @@ float WorldObject::SelectBestZForDestination(float x, float y, float z, bool exc
     float myX, myY, myZ;
     GetPosition(myX, myY, myZ);
 
-    float const myCollisionHeight = excludeCollisionHeight ? 0.0f : GetCollisionHeight();
+    float const myCollisionHeight = GetCollisionHeight();
+    float const destCollisionHeight = excludeCollisionHeight ? 0.0f : myCollisionHeight;
 
     float const myGridHeight = GetMap()->GetGridMapHeight(myX, myY);
     float const myVmapFloor = std::max(GetMap()->GetVMapFloor(myX, myY, myZ, 150.0f, GetCollisionHeight()),
@@ -2325,39 +2326,37 @@ float WorldObject::SelectBestZForDestination(float x, float y, float z, bool exc
 
     // which of these 3 do I want ?
     float const destGridHeight = GetMap()->GetGridMapHeight(x, y);
-    float const destCeil = GetMap()->GetCeil(GetPhaseMask(), x, y, z, 150.0f, myCollisionHeight);
-    float const destVmapFloor = std::max(GetMap()->GetVMapFloor(x, y, z, 150.0f, myCollisionHeight),
-        GetMap()->GetGameObjectFloor(GetPhaseMask(), x, y, z, 150.0f, myCollisionHeight));
+    float const destCeil = GetMap()->GetCeil(GetPhaseMask(), x, y, z, 150.0f, destCollisionHeight);
+    float const destVmapFloor = std::max(GetMap()->GetVMapFloor(x, y, z, 150.0f, destCollisionHeight),
+        GetMap()->GetGameObjectFloor(GetPhaseMask(), x, y, z, 150.0f, destCollisionHeight));
 
     bool const hasVmapFloor = myVmapFloor > INVALID_HEIGHT;
     bool const hasDestGridHeight = destGridHeight > INVALID_HEIGHT;
     bool const hasDestVmapCeil = destCeil < VMAP_INVALID_CEIL && destCeil != destVmapFloor;
     bool const hasDestVmapFloor = destVmapFloor > INVALID_HEIGHT;
     bool const destBetweenVmaps = hasDestVmapCeil && hasDestVmapFloor;
-    bool const underVmap = !hasDestVmapFloor;
-    bool const aboveAllVmap = !hasDestVmapCeil;
+    bool const noVmap = !hasDestVmapFloor && !hasDestVmapCeil;
 
     bool const isOnVmap = hasVmapFloor &&
         ((myZ < myGridHeight && std::fabs(myVmapFloor - myZ) < std::fabs(myGridHeight - myZ)) ||
         (myZ > myGridHeight && myVmapFloor > myGridHeight));
 
-    bool const hasToFollowGridHeight = hasDestGridHeight && ((underVmap && aboveAllVmap) ||
-        (destBetweenVmaps && !isOnVmap && ((destGridHeight > destVmapFloor && destGridHeight < destCeil))) ||
-        (hasDestVmapFloor && !hasDestVmapCeil && !isOnVmap && (destGridHeight > destVmapFloor)) ||
-        (hasDestVmapFloor && !hasDestVmapCeil && isOnVmap && z > destGridHeight && destGridHeight > destVmapFloor) ||
-        (hasDestVmapFloor && !hasDestVmapCeil && isOnVmap && myZ > myGridHeight && destVmapFloor < destGridHeight) ||
-        (hasDestVmapCeil && !hasDestVmapFloor && !isOnVmap && (destCeil < destGridHeight || destCeil - destGridHeight > GetCollisionHeight())));
+    bool const hasToFollowGridHeight = hasDestGridHeight && (noVmap ||
+        (z > destGridHeight && destGridHeight > destVmapFloor) ||
+        (z < destGridHeight && hasDestVmapFloor && !hasDestVmapCeil) ||
+        (z < destGridHeight && !hasDestVmapFloor) ||
+        (destBetweenVmaps && !isOnVmap && ((destGridHeight > destVmapFloor && destGridHeight < destCeil))));
 
     float result = INVALID_HEIGHT;
     if (hasToFollowGridHeight)
     {
         result = destGridHeight;
         if (hasDestVmapFloor)
-            if (std::fabs(destVmapFloor - destGridHeight) < GetCollisionHeight())
+            if (std::fabs(destVmapFloor - destGridHeight) < myCollisionHeight)
                 result = std::max(destVmapFloor, destGridHeight);
 
         if (hasDestVmapCeil)
-            if (std::fabs(destCeil - destGridHeight) < GetCollisionHeight())
+            if (std::fabs(destCeil - destGridHeight) < myCollisionHeight)
                 result = std::max(destCeil, destGridHeight);
     }
     else if (hasDestVmapFloor)
@@ -2370,7 +2369,7 @@ float WorldObject::SelectBestZForDestination(float x, float y, float z, bool exc
             return result;
 
     LiquidData liquidData;
-    ZLiquidStatus const liquidStatus = GetMap()->GetLiquidStatus(x, y, result, MAP_ALL_LIQUIDS, &liquidData, myCollisionHeight);
+    ZLiquidStatus const liquidStatus = GetMap()->GetLiquidStatus(x, y, result, MAP_ALL_LIQUIDS, &liquidData, destCollisionHeight);
     switch (liquidStatus)
     {
         case LIQUID_MAP_ABOVE_WATER:
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2321,8 +2321,8 @@ float WorldObject::SelectBestZForDestination(float x, float y, float z, bool exc
     float const destCollisionHeight = excludeCollisionHeight ? 0.0f : myCollisionHeight;
 
     float const myGridHeight = GetMap()->GetGridMapHeight(myX, myY);
-    float const myVmapFloor = std::max(GetMap()->GetVMapFloor(myX, myY, myZ, 150.0f, GetCollisionHeight()),
-        GetMap()->GetGameObjectFloor(GetPhaseMask(), myX, myY, myZ, 150.0f, GetCollisionHeight()));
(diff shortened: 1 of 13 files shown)


src/server/game/Entities/Object/Object.cpp
```diff
@@ -2305,13 +2305,90 @@ void WorldObject::MovePosition(Position &pos, float dist, float angle)
...
+    bool const hasDestVmapFloor = destVmapFloor > INVALID_HEIGHT;
+    bool const destBetweenVmaps = hasDestVmapCeil && hasDestVmapFloor;
+    bool const underVmap = !hasDestVmapFloor;
+    bool const aboveAllVmap = !hasDestVmapCeil;
+
+    bool const isOnVmap = hasVmapFloor &&
+        ((myZ < myGridHeight && std::fabs(myVmapFloor - myZ) < std::fabs(myGridHeight - myZ)) ||
+        (myZ > myGridHeight && myVmapFloor > myGridHeight));
+
+    bool const hasToFollowGridHeight = hasDestGridHeight && ((underVmap && aboveAllVmap) ||
```
Golrag: help :'(

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2352,20 +2410,59 @@ void WorldObject::MovePositionToFirstCollision(Position &pos, float dist, float
...
-    pos.SetOrientation(GetOrientation());
+    if (vmapCollision)
+    {
+        x = vmapCollisionPos.m_positionX;
+        y = vmapCollisionPos.m_positionY;
+        z = vmapCollisionPos.m_positionZ;
+        return true;
+    }
+
+    if (gameObjectCollision)
```
ccrs: Shouldnt this be over vmap?

src/server/game/Entities/Object/Object.h
```diff
@@ -273,7 +273,9 @@ class TC_GAME_API WorldObject : public Object, public WorldLocation
         void GetClosePoint(float &x, float &y, float &z, float size, float distance2d = 0, float angle = 0) const;
         void MovePosition(Position &pos, float dist, float angle);
         Position GetNearPosition(float dist, float angle);
+        float SelectBestZForDestination(float x, float y, float z, bool excludeCollisionHeight) const;
```
ccrs: move to private?

src/server/game/Entities/Object/Object.cpp
ccrs: ugh

A brief description on whats being decided here might be helpful

Shauren:
damn, looking at this I think I preferred that +2 to z

src/server/game/Entities/Object/Object.cpp
Golrag: oh boi.
I will see if I can get back to paint + provide in game experiences

src/server/game/Entities/Object/Object.cpp
Golrag: Already checked above if both occured and checked for closest one, so I don't think the order here is relevant

Golrag:
Pretty sure it's not the +2's but the NormalizeZforCollision that got removed
I should have re-add it like it was from the beginning maybe lol

src/server/game/Entities/Object/Object.cpp
Golrag: image
We want to follow grid height in this example. purple is vmap/gob height data and green is grid

So, we have a destVmapFloor as you can see, which, for this example, is closer to the vmap than to the grid. Obviously we don't want to end there and we should select gridheight based on these conditions:
- We have a destvmap floor, but are not on a vmap currently.
- There is no vmapceil
- The grid height is higher than the vmap floor we found.

It's not finished, and I'm afraid it never will. But I think you can follow the other cases.
It tried to note everything but got quickly lost in everything and didn't really find a good way to keep track of it.

Tbh, I want this gone. The only tools I have is some modified .gps command which returns the value I want, and the debugger and searching for every corner case is such a fucking pain.

Golrag:
Added some videos for tests done + rev i did the tests in

jackpoz:
> Attempt to fix some undermap issues

could you describe more in detail how you implemented the fix ?

Cannix:
You may wanna consider to remove [https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellEffects.cpp#L4446](url) this line as position should already have a valid position and this just distorts previous calculations.

Removed this line on my server some time ago to fix false z positions.

Golrag:
@Cannix already did that

@jackpoz sure,
I will try to break it up in small parts.
Functional changes to Map::GetHeight are done because it could sometimes return -20000.0f while there was a gridmap height. I  moved collisionheight to it's own variable so that it's only included when doing vmap height data and check if we are above grid height. Which would lead our z position in UpdateAllowdPositionZ to the same we already had. Which could cause going through textures when calling this method in NormalizePath.

For Object::MovePositionToFirstCollision, UpdateAllowedPositionZ is not really an option. We want to select our z based on where we currently are. So I decided to create a new function called SelectBestZForDestination. They could probably be merged together, but I would prefer to wait for the hover PR to be merged (which includes a clean up of that function)  and then see if it would work out cleanly.

That function's only purpose is to choose between 3 positions (and later check for water).
Our raycast did not hit anything, but we might have gone through terrain so we are possibly under the map. At that dest location we might have a vmap above it or under it. We also might have none of them but still have a grid height.

It might have hit something, but still under terrain (like the top of a tunnel while we are blinking/getting feared above it). It's all about finding those cases and then select what's the best result of it.

jackpoz:
summing up all changes made so far to vmaps including this PR, does fear cause more or less undermaps than before ? with all these follow-up PRs I lost track if now we have more or less issues than before.

Golrag:
My PR that caused this is #21125 which removed "NormalizeZforCollision" which is what caused NPC's/Players to not go undermap previously (in most cases). But, it also made it possible to target a point above you (see the issue with the dragons in tanaris). Because we were selecting the closest point to z between 'ground', which would have represented gridHeight, and 'floor', which would have represented the vmap height which was not present in 50.0f so -20000.0f.

Can't say with 100% that it is better now, I would need a lot of tests and always try to compare it, which could take some time.

jackpoz:
@Golrag is this ready to be merged to get some more feedback from the users ?

Golrag:
Hm I was able to get an undermap in dalaran sewers but it was really hard to get it. Nearly impossible while debugging to see the values / what happened etc.
The undermap didn't happen because of selecting grid height instead of vmapfloor/ceil. What happened before this PR was that if there was collision it would have moved back by CONTACT_DISTANCE. But if I'm not mistaken, ```getObjectHitPos``` should already do that so I removed it. Now it is possible that we moved not enough back, so I might make that value higher (x2), and hope it get's fixed.
Could also have been path generator that didn't return a correct path, not sure.

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2320,52 +2396,70 @@ void WorldObject::MovePositionToFirstCollision(Position &pos, float dist, float
...
-    {
-        // move back a bit
-        destx -= CONTACT_DISTANCE * std::cos(angle);
-        desty -= CONTACT_DISTANCE * std::sin(angle);
         dist = std::sqrt((pos.m_positionX - destx)*(pos.m_positionX - destx) + (pos.m_positionY - desty)*(pos.m_positionY - desty));
-    }
 
-    // check dynamic collision
-    col = GetMap()->getObjectHitPos(GetPhaseMask(), pos.m_positionX, pos.m_positionY, pos.m_positionZ, destx, desty, destz, destx, desty, destz, -0.5f);
+    destz = SelectBestZForDestination(destx, desty, destz, col);
```
ccrs: > // You should check for collision again after this one has been called.

Golrag: Yes, still looking into this. I didn't like how I did it before as we might had to reselect a better z until no more collision but that seemed to much for me. Also got infinity loops while doing it lol.
So "ignoring" that comment makes the behavior a bit more like what happened previously

Golrag: I'm going to re add it anyway as it should fix more than break things (no more blinking of the roof in WSG!), with a comment tho

Persephone:
Any news on getting this merged?

jackpoz:
@Persephone did it fix underground issues on your local checkout ?

Wilcalaf:
Will a bounty help speed this up??
Actually idk how to put bounties on PRs, is it even possible?

Golrag:
No it just needs more test results to see if something is still wrong. (And if possible, see if it did happen before my previous PR which removed NormalizeZForCollision)

jackpoz:
@Wilcalaf I think right now the most helpful thing we need is feedback to see if this PR fixes the issues, please test it locally and let us know if you find any issue ingame about it.

Wilcalaf:
Hello 
It looks working (at least for blink)
Before this PR
https://streamable.com/zdsrx

After This PR
https://streamable.com/vfa59

Also i tried go havoc in Warsong to see if it happens, could not fall underground

https://streamable.com/xxrbr

So far i was not able to fall underground with blink no matter what i did. but im still testing this.

jackpoz:
@Golrag is it ready to be merged for you ?

Golrag:
I think it is

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2305,13 +2305,91 @@ void WorldObject::MovePosition(Position &pos, float dist, float angle)
     pos.SetOrientation(GetOrientation());
 }
 
+// Only call this function when you already checked for collision with vmap/gameobject.
+// You should check for collision again after this one has been called.
+// excludeCollisionHeight should only be true if you had collision, it wont add it to raycasts for dest position.
+float WorldObject::SelectBestZForDestination(float x, float y, float z, bool excludeCollisionHeight) const
+{
+    if (Unit const* unit = ToUnit())
+        if (unit->IsFlying()) // Also check if in air ?
```
ccrs: if its not expensive, it should

also, skipping any correction for flying units feels scary

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2320,52 +2398,72 @@ void WorldObject::MovePositionToFirstCollision(Position &pos, float dist, float
...
+        desty -= step * std::sin(angle);
+        // There should not be any collision between our position and destx, desty, pos.m_positionZ at this point.
+        // Use pos.m_positionZ here because destz was not good.
+        destz = SelectBestZForDestination(destx, desty, pos.m_positionZ, col);
     }
 
-    float step = dist / 10.0f;
+    Trinity::NormalizeMapCoord(destx);
+    Trinity::NormalizeMapCoord(desty);
+    // We might want to loop until there is no more collision with a better z position. (And/or until a fixed #attemps have been made).
```
ccrs: where is the attempt limit?

src/server/game/Entities/Object/Object.cpp
Golrag: It will only skip terrain because I didn't find any way to check it :/

I could just take the isInAir check that's used somewhere in updatemovementflags iirc and use that one for the comment tho

src/server/game/Entities/Object/Object.cpp
Golrag: I did not implement the loop to keep it like it was before I broke it, so the attempt limit is not there (yet). If there is really a need for it right now I will implement it in this PR.

Persephone:
Any news on the merge?

jackpoz:
@Golrag I see you added some commits after it was marked as ready, is it ready now :P ?

Wilcalaf:
@jackpoz Hey NP guys ill try to test it asap.

ccrs:
since  ```SelectBestZForDestination``` can be called several times in one execution, are its internall calls expensive? remember getarea or w/e its called, its result had to be cached for that reason.

Wilcalaf:
Guys were the changes he applied supposed to change something in game? 
its working like it was before ( which is good i assume 👍 )

tkrokli:
@Wilcalaf : Golrag knows this PR a lot better than me, but AFAIK this is meant to fix when Blink lands the caster to a point below the map Z floor (and thereby falling into empty space/dying from fall damage). The secondary goal might be to avoid NPCs dropping under the map when moving around too, but I don't know if this is implemented (I forgot to ask earlier).

Golrag:
@jackpoz while I made some comments in the code about potentional improvements, I think it's ready. Unless someone decides it's not and says they have to be implemented in this PR.

@ccsr they might be expensive, but I don't really get the idea behind caching it since it's different for every position I might be missing something here

Sorry for late response, busy weeks.

Issue #21295, opened 2018-01-23
FreezerDetector:Handler


Gooyeth:
**Description:**

Crash occurs for no apparent reason, it happens in Release and Debug.

`MaxCoreStuckTime = 120`

https://gist.github.com/Gooyeth/de7465acc70ed12342b0fc342e167962

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 514c847

**TDB version:**  TDB 335.63

**Operating system:** Debian 9 - **RAM**: 64 GB - Using NVMe

ariel-:
Please use the contrib/debugger/crashreport.gdb to generate crashlogs on gdb, freeze crashes are useless to debug without all threads backtrace

Gooyeth:
Ready, thanks.
[crash_freezer.txt](https://github.com/TrinityCore/TrinityCore/files/1656896/crash_freezer.txt)

Aokromes:
Maybe the same as https://github.com/TrinityCore/TrinityCore/issues/21187

Shauren:
@Aokromes definitely no
From that crash log it looks like map update simply took a long time, what is your freeze limit set to?

ariel-:
> `MaxCoreStuckTime = 120`

Gooyeth:
I have disabled the spell "Mirror Image" and the crash has decreased considerably, with stability for more than 8 hours without it.

Shauren:
@Treeston  Could it be that new threat manager is very slow or has some infinite loop?

Treeston:
Unlikely, though obviously possible.

Treeston:
(ED: `MaxCoreStuckTime` is in what unit? 120 _seconds_?)

Gooyeth:
Yes, seconds.

Even with more than 15 minutes no more response is obtained.

jackpoz:
the only thread doing something there is calling CombatManager::SetInCombatWith(), it's probably not a coincidence

Treeston:
@Gooyeth did it crash at all with mirror images disabled?

Gooyeth:
The crash is still, but less frequently.

Without disable the spell, it was stable for 30 minutes - 1 hour.

Treeston:
What sort of player count are we talking about? Would it be feasible to figure out what players were doing at the time of the crash?

Gooyeth:
It is produced even with 10 active players, I could not say exactly what action they were performing.

Treeston:
[...]

PR #21324, opened 2018-01-29
[3.3.5] Core/Creature: Remove legacy creature behavior

Treeston:
Some adjustments to `Creature::SelectVictim`. PR for brief review and awareness.

- No more automatic combat -> threat promotion. This matches anecdotal retail behavior - spells already in transit will not cause mob to chase you if its primary target vanishes before the spell hits.
- Prevent mobbing by stopping creatures from auto-acquiring a nearby target just because the primary target vanished.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -1112,32 +1112,8 @@ Unit* Creature::SelectVictim()
 {
     Unit* target = nullptr;
 
-    ThreatManager& mgr = GetThreatManager();
-
-    if (mgr.CanHaveThreatList())
-    {
-        target = mgr.SelectVictim();
-        while (!target)
-        {
-            Unit* newTarget = nullptr;
-            // nothing found to attack - try to find something we're in combat with (but don't have a threat entry for yet) and start attacking it
-            for (auto const& pair : GetCombatManager().GetPvECombatRefs())
-            {
-                newTarget = pair.second->GetOther(this);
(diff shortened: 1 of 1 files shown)


Treeston:
should fix #21295

Treeston:
@Shauren give this a quick look-over please when you have time - I am confident that these changes are correct and match retail behavior, but maybe there's some edge case you think of that I haven't.

Shauren:
Wait what, does that second point (REACT_AGGRESSIVE removal) mean that going invis could make the creature attack any other nearby player? Interesting and not something I can say with 100% confidence doesn't happen on retail

Treeston:
It _currently_ can do that, yeah. If you kite a mob near another player and vanish, it will aggro that other player. Retail doesn't do that with 99% confidence (but if someone can test it that'd be nice).

Treeston:
See https://gfycat.com/CoordinatedShamefulCoral

Nyeriah:
That doesn't happen on retail unless the creature is hit or there's some sort of interaction with the player vanishing (buffing or healing). I play a hunter and a rogue, you definitely can't pull mobs into people like that, it would be fun if you could tho.

Nitair:
As far as I know, you get aggro if mobs does cleave. If "player A" pulls a mob and does a cleave which damages "player B". After a vanish or sth. like that from "player A", "player B" will be attacked

Treeston:
I'm fairly confident that only applies if player B had some retaliatory effect (thorns etc) on them.

PR #21336, opened 2018-01-30
Core/SmartScripts: new SAI action SMART_ACTION_SET_MOVEMENT_SPEED

Lopfest:
**Changes proposed:**
-  implement new SMART_ACTION_SET_MOVEMENT_SPEED
- action_param_1 = movementType from UnitMoveType enum
- action_param_2 = sniff speed value (only integer part, before decimal point)
- action_param_3 = sniff speed value (only fraction part, after decimal point if any)

- [x] master

**Issues addressed:** -

**Known issues and TODO list:** -

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -2873,6 +2873,23 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
             delete targets;
             break;
         }
+        case SMART_ACTION_SET_MOVEMENT_SPEED:
+        {
+            ObjectList* targets = GetTargets(e, unit);
+            if (!targets)
+                break;
+
+            uint32 speedInteger = e.action.movementSpeed.speedInteger;
+            uint32 speedFraction = e.action.movementSpeed.speedFraction;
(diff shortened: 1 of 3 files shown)


Infinito47:
Woul this also affect swim speed?

Shauren:
Why do you see the need to use 2 separate fields? (speedDivider)

Lopfest:
more precise speed adjustment.
RunSpeed: 8 in sniffs is 1.14286 in DB right?
When you want to set (sniff) speed to 14, you will set action_param_2=2, action_param_3=1
8/1,14286=7 - should be basic run coefficient
14/7=2
Thats ok, but what if I want set (sniff) speed to something indivisible by 7, for example 10,5 for simplicity?
10,5/7=1,5 - what two whole numbers give result 1,5?
3/2=1,5 - action_param_2=3, action_param_3=2

im not saying it will be accurate, but setting speed only to whole numbers is not always what you want.

Shauren:
right, i forgot they are ints in db, not floats

Lopfest:
@Shauren, should I add these parameters also to SMART_ACTION_SET_FLY and SMART_ACTION_SET_SWIM?

Shauren:
Wouldnt it be better to extract this into a new SMART_ACTION_SET_MOVEMENT_SPEED with your 2 params + movement type?

Lopfest:
ok, i dont have problem with that. ill do it today

Aokromes:
@Shauren this can be merged?

PR #21383, opened 2018-02-08
Core/Spells: fixed a logic mistake in IsMovementPreventedByCasting()

Ovahlord:
**Changes proposed:**

-  fixed stupid inverted logic that was causing some channelled spells getting interrupted because the caster was moving while he should not be allowed to move.

- [x] 3.3.5

Tested ingame, works for me (tm)

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -3329,17 +3329,17 @@ int32 Unit::GetCurrentSpellCastTime(uint32 spell_id) const
 bool Unit::IsMovementPreventedByCasting() const
 {
     // can always move when not casting
-    if (!HasUnitState(UNIT_STATE_CASTING))
-        return false;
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -3329,17 +3329,17 @@ int32 Unit::GetCurrentSpellCastTime(uint32 spell_id) const
 bool Unit::IsMovementPreventedByCasting() const
 {
     // can always move when not casting
-    if (!HasUnitState(UNIT_STATE_CASTING))
-        return false;
+    if (HasUnitState(UNIT_STATE_CASTING))
```
ariel-: No. you aren't checking attributes if you do this.

Treeston: is the second change correct? (seems like it to me)

PR #21386, opened 2018-02-09
3.3.5 deadlocks

jackpoz:
Continuing from https://github.com/TrinityCore/TrinityCore/pull/21296


src/server/database/Database/Transaction.cpp
```diff
@@ -63,6 +67,25 @@ void Transaction::Cleanup()
...
 
+std::string Transaction::ToString()
+{
+    std::string msg;
+    for (SQLElementData const& data : m_queries)
+    {
+        switch (data.type)
+        {
+            case SQL_ELEMENT_PREPARED:
+                msg += data.element.stmt->getQueryString();
```
jackpoz: the whole ToString() concept doesn't work because it throws an exception if called before the transaction is executed and returns deleted data after the transaction has been executed (errored or not)

PR #21392, opened 2018-02-11
[WIP] Core/Movement: fix the enter cyclic movement procedure

### LINKED ISSUE #20238: MoveCirclePath always skips first movepoint
**Description:**
MoveCirclePath generates correct movepoint list, but the first movepoint id is always skipped and current creature position before MoveCirclePath is used instead of it.

in ``int32 MoveSplineInit::Launch()`` first movepoint is overwritten by ``args.path[0] = real_position;``

**Current behaviour:**
Lets say we want a NPC to move like this, starting from some start(spawn) position and then moving in circle:
image

MoveCirclePath generates path list that corresponds our needs, which looks like this:
image

But our NPC is moving always this way, skipping first generated point id and using its spawn position instead:
image

**Expected behaviour:**
Use first movepoint generated by MoveCirclePath

1. change ``npc_quel_delar_sword`` script from ``_events.ScheduleEvent(EVENT_QUEL_DELAR_LAND, 15000);`` to ``_events.ScheduleEvent(EVENT_QUEL_DELAR_LAND, 150000);`` for example
2. .gm on
3. .go xyz 5309 2006 712 668
4. .npc add temp 37158
5. watch as it skips first movepoint

**Branch(es):** both

**TC rev. hash/commit:** 2d0e65386693c35ae8f00fc0a1beee032a5ccf18

**TDB version:**  720.00

### LINKED ISSUE #21387: MoveCirclePath is missing proper enter_cycle support
**Description:**
the MoveCirclePath spline movement is currently not using the enter_cycle spline flag to make creatures join a circle path without using their current position as waypoint. Many cyclic moving creatures use that enter_cycle spline flag to move to a circle path without having their current position getting added to the path.

**Current behaviour:**
MoveCirclePath as well as manually setting up cyclic movements always take the start position into their path. This should be skipable by using the enter_cycle spline flag.

**Expected behaviour:**
MoveCirclePath as well as manual MoveSplineInit calls should be able to allow the mover to enter the circle path from anywhere without having his current position being added to the path.
cycle
The left picture is the current scenario. The right one is what should happen.

Here is a packet that shows the process. The first two spline waypoints are used for entering the cycle, the others are the actual path.
[23] Spline flags: 4209152 (Flying, Catmullrom, Cyclic, EnterCycle, UncompressedPath)
[23] Spline Time: 3631
[23] [0] Spline Waypoint: X: -193.1041 Y: -442.9865 Z: 53.70924
[23] [1] Spline Waypoint: X: -193.2778 Y: -442.0017 Z: 53.70924
[23] [2] Spline Waypoint: X: -193.4514 Y: -441.0169 Z: 55.70924
[23] [3] Spline Waypoint: X: -192.7042 Y: -441.1826 Z: 55.70924
[23] [4] Spline Waypoint: X: -192.293 Y: -441.8281 Z: 55.70924
[23] [5] Spline Waypoint: X: -192.4586 Y: -442.5753 Z: 55.70924
[23] [6] Spline Waypoint: X: -193.1041 Y: -442.9865 Z: 55.70924
[23] [7] Spline Waypoint: X: -193.8514 Y: -442.8209 Z: 55.70924
[23] [8] Spline Waypoint: X: -194.2626 Y: -442.1754 Z: 55.70924
[23] [9] Spline Waypoint: X: -194.0969 Y: -441.4282 Z: 55.70924
[23] [10] Spline Waypoint: X: -193.4514 Y: -441.0169 Z: 55.70924
[23] [11] Spline Waypoint: X: -192.7042 Y: -441.1826 Z: 55.70924
[23] Spline Duration Multiplier Next: 1
[23] Spline Full Time: 30379
[23] Spline Duration Multiplier: 1
[23] Spline Id: 1966
[23] Spline Endpoint:: X: -193.4514 Y: -441.0169 Z: 55.70924

1. spawn a creature somewhere in the world
2. use a MoveCirclePath call somewhere above the creature's position
3. see how the creature will always fly down to it's starting position. This should be skipable by using enter_cycle.

If you need some examples where exactly this is needed:
- Twilight Hatchlings in Grim Batol Erudax encounter when the hatchlings go into the air and move arround cyclic there also with the enter_cycle flag
- Glubtok in Deadmines --> His firewall starts at a spawn positon, after the wall trigger moves to a circle path. This is where this packet above came from.

**Branch(es):**
both

**TC rev. hash/commit:** 
91aa8e1193db690e5a50c700280cc7103e2f10b4

Ovahlord:
**Changes proposed:**
- rework the cyclic movement part of the spline movement so we will get retail-like behaivior now

**Target branch(es):** both

- [x] 3.3.5
- [x] master

**Issues addressed:**
 Closes #20238
 Closes #21387

**Todo:**
- [x] fix packet structure to send the first vertex only once
- [ ] make the cyclic path ignore the first vertext on repeating
- [ ] update scripts to apply changes correcrly

**Tests performed:** tested ingame; issues noticed. need to fix

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -510,6 +510,7 @@ void MotionMaster::MoveCirclePath(float x, float y, float z, float radius, bool 
         init.SetFly();
         init.SetCyclic();
         init.SetEnterCycle(enterCycle);
+        init.SetUncompressed();
         init.SetAnimation(Movement::ToFly);
     }
     else
@@ -517,6 +518,7 @@ void MotionMaster::MoveCirclePath(float x, float y, float z, float radius, bool 
         init.SetWalk(true);
         init.SetCyclic();
         init.SetEnterCycle(enterCycle);
+        init.SetUncompressed();
     }
 
     init.Launch();
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -510,7 +510,6 @@ void MotionMaster::MoveCirclePath(float x, float y, float z, float radius, bool 
         init.SetFly();
         init.SetCyclic();
         init.SetEnterCycle(enterCycle);
-        init.SetUncompressed();
         init.SetAnimation(Movement::ToFly);
     }
     else
@@ -518,7 +517,6 @@ void MotionMaster::MoveCirclePath(float x, float y, float z, float radius, bool 
         init.SetWalk(true);
         init.SetCyclic();
         init.SetEnterCycle(enterCycle);
-        init.SetUncompressed();
     }
 
     init.Launch();
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -482,7 +482,7 @@ void MotionMaster::MoveJump(float x, float y, float z, float o, float speedXY, f
     Mutate(new EffectMovementGenerator(id), MOTION_SLOT_CONTROLLED);
 }
 
-void MotionMaster::MoveCirclePath(float x, float y, float z, float radius, bool clockwise, uint8 stepCount)
+void MotionMaster::MoveCirclePath(float x, float y, float z, float radius, bool clockwise, uint8 stepCount, bool enterCycle)
 {
     float step = 2 * float(M_PI) / stepCount * (clockwise ? -1.0f : 1.0f);
     Position const& pos = { x, y, z, 0.0f };
@@ -490,7 +490,8 @@ void MotionMaster::MoveCirclePath(float x, float y, float z, float radius, bool 
 
     Movement::MoveSplineInit init(_owner);
 
-    for (uint8 i = 0; i < stepCount; angle += step, ++i)
+    uint32 totalStepCount = enterCycle ? stepCount + 1 : stepCount;
+    for (uint8 i = 0; i < totalStepCount; angle += step, ++i)
     {
         G3D::Vector3 point;
         point.x = x + radius * cosf(angle);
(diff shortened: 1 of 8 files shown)


Keader:
```
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Movement/MotionMaster.cpp:513:14: fatal error: no member named 'SetUncompressed' in 'Movement::MoveSplineInit'
        init.SetUncompressed();
```

Ovahlord:
Could need a little bit of help on that. Right now the last remaining issue is that the creature should start to join its circle from its current position but it should not move to that location anymore.
The process works like this:
1. path gets calculated: current creature position + circle path
2. packet gets sent with enter_cycle flag to tell the client to delete the first vertex after first cycle is complete. the first vertex is the creature position when we started to move. That one needs get removed from the path

Ovahlord:
Well, while I am at it, I might drop that old behaivior from the old movecircle calls so all scripts use the new correct handling.

Ovahlord:
I need to close this PR for now. After multiple internal tests, I saw that the coreside cyclic handling does not work exactly like it should. It updates the positions faster than the client so you get asynch positions with the creature and the core.

For documentations:

- The monster_move packet may not send the first vertex twice for cyclic movement. This is wrong. The client erases the first vertex automaticly so the cyclic movement will become perfect clientside after that change.
- The core however needs two changes: 
1. instead of correcting the first vertex in MoveSplineInit::Launch() the position needs to be shuffled BEHIND the cyclic path. 
2. MoveSpline::_updateState needs an extra handling to remove the first vertex from the spline once it has been reached. So you need to recalculate the path from there on.
- Cyclic movement always has the spline flags: enter_cycle, cyclic, catmullrom and on 4.x and above: Uncompressedpath
- After the initial monster_move or update_object packet with the path inside there is NO other packet sent according to sniffs. So resending the monster_move packet for example is hacky and wrong. The client handles the movement automaticly with those flags. The splinesystem coreside needs tweaks to reflect that behaivior.

PR #18885, opened 2017-01-18
Core/GameObject: GAMEOBJECT_TYPE_GOOBER not casting spells

giridhar123:
[//]: # (***************************)
[//]: # (** FILL IN THIS TEMPLATE **)
[//]: # (***************************)

**Changes proposed:** **Description:** Every gameobject with type 10 (GAMEOBJECT_TYPE_GOOBER) with spell id in data10 field should cast a spell when you click on it.

**Current behaviour:** The spell isn't cast

**Expected behaviour:** The gameobject should cast its spell

**Steps to reproduce the problem:**

1. Just find any gameobject in game with type 10 and data10 field different from 0 (you can run this query to find them): `SELECT * FROM gameobject_template WHERE TYPE = 10 AND data10 > 0;`
2. Right click on it and you will see that the object doesn't cast the spell (for some gameobject you can only open it)

https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/GameObject/GameObject.cpp#L1464 : `spellCaster = nullptr;`

Why is the spellCaster set to nullptr?

https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/GameObject/GameObject.cpp#L1887-L1890
```c++
    if (spellCaster)
        spellCaster->CastSpell(user, spellInfo, triggered);
    else
CastSpell(user, spellId);
```

I tried adding comment to `spellCaster = nullptr;` and it works and it also works if you change the line 
`CastSpell(user, spellId);` to `user->CastSpell(user, spellId);`

**Target branch(es):** master

**Issues addressed:** Closes #18752

**Tests performed:** Tested in game and it works with no crashes.

**Known issues and TODO list:**
Nothing (I think), please check if something is missing.

Issue #21142, opened 2017-12-30
[3.3.5] Spell Health Funnel removing more life that should


Keader:
Current script remove warlock health in 2 points.
Here: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L5941
And Here: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/Auras/SpellAuras.cpp#L706
The second point, is called 2 times: First call is caused by Warlock Aura. The second is caused by Pet Aura.
And current script, dont check any SPELLMOD_COST, so reduce cost part of Improved Health Funnel not work.

And **Improved Health Funnel** talent is not working (reduce health part)

More info here: #20559

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  527ef19ae8a0c82b1c0d7798c15c3f3f22162106

ariel-:
Needs #21395 first (because of two aura objects applied, one on pet, other on warlock)

CraftedRO:
8ad43140305e8aacd8c27f891d5f701a1452ca61 not really sure if still valid but it can kill you and I doubt it shoud.

CraftedRO:
2c07f34c996fc41e78e44902b590d15992de0538 still valid, here's a temporary tested (for months on my server) fix, git version in case someone need it from previous unaccepted PR:
```
-- WARLOCK FUNNEL FIX
curl https://github.com/TrinityCore/TrinityCore/commit/8e6f0582e6042eb3229aa54aeea400a958c9be98.diff | git apply
```

Issue #20560, opened 2017-10-06
Core/Spells: Cyclone vs Positive Auras


Keader:
**Description:** You can hit positive auras on friendly target with Cyclone.

**Current behaviour:** You can use spells like: Lifebloom, Rejuvenation, etc. in friendly targets affected by Cyclone. When cyclone is removed, your auras become active on target

**Expected behaviour:** All positive spells should return "IMMUNE" if friendly target is affected by Cyclone

**Steps to reproduce the problem:**

1. Apply cyclone aura in a friendly target (.aura  33786)
2. Try cast any positive spell on him (like  48441)
3. After Cyclone fade, your rejuvenation will be on target.

You can use it to "refresh" lifebloom, is a little gamebreak in pvp

PS: Is working if is negative aura, is missing some check to positve auras too.

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  67795685bae686094840c8805fefbe2cf0835009

Keader:
Change this line: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/Spell.cpp#L2509

To: ```if (unit->IsImmunedToSpell(m_spellInfo, m_caster))```
looks fix this issue, but i dont think that this is right place

Jildor:
I don't know why should be only delayed spells, @ariel- should know it.
Maybe should be all delayed spells and positive non delayed spells??

I found this commit: https://github.com/TrinityCore/TrinityCore/commit/9f5df023b746d324588f175e264c62205b69e165

Keader:
ariel is buzy with university, he should not appear here in this semester xD
Btw i cant understand why is checked only to delayed spells.

ariel-:
check UNIT_STATE_ISOLATED

CraftedRO:
835e2918319e Seems no longer valid:


https://user-images.githubusercontent.com/24683355/214086637-2ca7a69b-cc94-4bd7-b4a4-855272b8faf7.mp4

Issue #19518, opened 2017-04-28
[3.3.5/Spells] Devour Magic skipping duel check


Keader:
**Description:** Spell of Warlock Pet (Felhunter) Devour Magic is skipping duel check.

**Current behaviour:** You can dispel a target with Devour Magic, even if target is in duel with other player.

**Expected behaviour:** Spell should fail

**Steps to reproduce the problem:**

1. Put 2 characters in Duel
2. Use Corruption in a target .add aura  47813
3. With a warlock out of duel, use Devour Magic to dispel Corruption.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 4431a1149da7bc48ad1047fc8c6135ed02b2e77b

Raydor:
https://github.com/TrinityCore/TrinityCore/issues/14967#issuecomment-129504965

Keader:
you can do it without party too lol

Keader:
Ok i have a hack for it :)

```diff

diff --git a/src/server/scripts/Spells/spell_warlock.cpp b/src/server/scripts/Spells/spell_warlock.cpp
index b1c897b..b631dec 100644
--- a/src/server/scripts/Spells/spell_warlock.cpp
+++ b/src/server/scripts/Spells/spell_warlock.cpp
@@ -1484,6 +1484,43 @@ class spell_warl_unstable_affliction : public SpellScriptLoader
         }
 };
 
+// -19505 - Devour Magic
+class spell_warl_devour_magic : public SpellScriptLoader
+{
+public:
+    spell_warl_devour_magic() : SpellScriptLoader("spell_warl_devour_magic") { }
+
+    class spell_warl_devour_magic_SpellScript : public SpellScript
+    {
+        PrepareSpellScript(spell_warl_devour_magic_SpellScript);
+
+        SpellCastResult CheckExplicitTarget()
+        {
+            Unit* caster = GetCaster();
+            if (Unit* target = GetExplTargetUnit())
+            {
+                if (caster->IsFriendlyTo(target) && caster->IsValidAssistTarget(target))
+                    return SPELL_CAST_OK;
+
+                else if (caster->IsValidAttackTarget(target))
+                    return SPELL_CAST_OK;
+            }
+
+            return SPELL_FAILED_BAD_TARGETS;
+        }
+
+        void Register() override
+        {
+            OnCheckCast += SpellCheckCastFn(spell_warl_devour_magic_SpellScript::CheckExplicitTarget);
+        }
[...]

PR #21395, opened 2018-02-11
[Core/Spells] [Request extensive testing] Spell internals rework

### LINKED ISSUE #8844: [ICC/Mutated Transformation]Disease bug
Mutated Transformation ticks refreshes Spirit of Redemption duration so the priest with this talant can never die if there is an abomination alive. Indeed the priest dies from these ticks and come alive in the same second. It is possible to use this mechanic for honorable kills grind if mixed(horde+alliance) raids are enabled.

Mutated Transformation: http://wotlk.openwow.com/?spell=70402
Spirit of Redemption: Spirit of Redemption
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/104295-icc-mutated-transformation-disease-bug?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #15088: [3.3.5][Core/Spell] Mind Control - Diminishing return issue
Mind control applies a diminishing for buff and debuff of the spell. 

Issue: 
- Open 3 wow clients and login with different characters (Priest and 2 more random characters)
- Go to Ring of Trials (so they all will be enemies)
- Cast Mind Control on random player 1 (Mind Control buff is applied to the priest, and mind control debuff is applied to the random player 1 (current target))
- Cast Mind Control on random player 2 right after this: 

THERE is the issue, the buff for priest is at half duration, but the debuff for random player 2 is on full duration. 

When the buff for the priest expires, random player 2 is still in Mind Control, but priest now wont be able to cancelaura Mind Control because the buff for him is already gone bcuz of the diminishing from previous mind control on random player 1.

Priest's MC buff diminishing should be equal to current target diminishing state for MECHANIC_CHARM

3.3.5a
Hash: a6c7f78

### LINKED ISSUE #18395: Spell Effects applied in wrong order
**Description:** Spell Effects applied in wrong order.

**Expected behaviour:** Spell Effect sould be applied in correct order: EFFECT_0 -> EFFECT_1 -> EFFECT_2

**Current behaviour:** EFFECT_1 is applied before EFFECT_0.

1. Target self
2.  ```.cast 41377```
3. You will not receive damage

> Effect 0: Id 2 (SPELL_EFFECT_SCHOOL_DAMAGE)
> BasePoints = 7444 to 7557
> Targets (25, 0) (TARGET_UNIT_TARGET_ANY, NO_TARGET)
> 
> Effect 1: Id 6 (SPELL_EFFECT_APPLY_AURA)
> BasePoints = 1
> Targets (25, 0) (TARGET_UNIT_TARGET_ANY, NO_TARGET)
> Aura Id 40 (SPELL_AURA_DAMAGE_IMMUNITY), value = 1, misc = 127 (127), miscB = 0, periodic = 0

Damage should be applied before Immunity.

In sniffs packet order is:

> ServerToClient: SMSG_SPELL_NON_MELEE_DAMAGE_LOG (0x4315) Length: 45 ConnIdx: 2 Time: 05/30/2012 10:34:48.876 Number: 432310
> Target GUID: xxxxxxxxx
> Caster GUID: Full: 0xF1305B7C000050B4 Type: Creature Entry: 23420 (Essence of Anger) Low: 20660
> Spell ID: Spite (41377)
> Damage: 7525
> Overkill: -1
> SchoolMask: 8
> Absorb: 0
> Resist: 0
> Show spellname in log: false
> Unk byte: 1
> Blocked: 0
> HitType: SPELL_HIT_TYPE_UNK3 (4)
> Debug output: false
> 
> ServerToClient: SMSG_AURA_UPDATE (0x4707) Length: 30 ConnIdx: 2 Time: 05/30/2012 10:34:48.876 Number: 432311
> GUID: Full: xxxxxxxxxxx
> [0] Slot: 25
> [0] Spell ID: Spite (41377)
> [0] Flags: EffectIndex1, Duration, Negative (162)
> [0] Level: 73
> [0] Charges: 0
> [0] Caster GUID: Full: 0xF1305B7C000050B4 Type: Creature Entry: 23420 (Essence of Anger) Low: 20660
> [0] Max Duration: 2000
> [0] Duration: 2000

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  881cd78ae32d8078e247f64a73c651959a76e746

### LINKED ISSUE #21244: Core/Spells: "Placing Beacon Torch" (ID 17016) in quest "All Along The Watchtowers"
Description:
Accepted the quest from the quest starter and then headed of to the 4 towers and then used the item needed at the door of the towers and can't get the item to be used and cannot complete the objectives.

Current behaviour:
Go to any of the 4 towers and if you use the Beacon Torch item that is given to you by the quest giver at the tower but not the door it says required Andorhal Tower but then go to the door (slightly inside) it says You have no target. It gives a slight effect on the character but keeps all the charges on the beacon torch and nothing happens. If it is suppose to give a circle to place on the floor like a spell it does not and no objectives complete at any of the 4 towers.

Expected behaviour:
Go to the 4 towers and use the Beacon Torch at the doorway and it should give a objective complete for that tower. Not sure if it is suppose to leave a beacon on the floor.

The steps I have taken to reproduce the problem from abandon the quest and get it again was.
1) Level 58 Human Paladin (not sure if that is relevant)
2) get the quest from the plaguelands
3) go to tower 1,2,3 or 4 as all of them do the same and use the beacon torch in the doorway outside of combat

Branch(es):
3.3.5a

TC rev. hash/commit:
TrinityCore rev. 6892404b270f+ 2017-04-26 04:20:38 -0300 (3.3.5 branch) (Win64, Release, Static) (worldserver-daemon)

TDB version: TDB 335.63

Operating system: Windows 10 (64bit)

Extra information:
I have looked on the issues list and there was 1 similar issue but closed and I added the second DB code and added it but same result.

30 minutes later: This issue was bugging me so much I looked into the DB and found out the beacon torch item has a spell id of 17016 but could not find the spell id of 17016 in the DB

### LINKED ISSUE #21330: Core/GameObject/Spells: incorrect handling of spell executing GObjects
**Description:**

The behavior in TC is totaly different to Retail and hacked.
This causes a lot of cosmetic and logic error which need to fixed by hacks again.
This issue is caused by TC spell system.
This system needs a rewrite to support gameobject casts.

I donno if there are other related issues. Or if there are other systems which need to be rewirten for this working properly.

**Current behaviour:**

At the moment, gameobjects which execute a spell, make the player cast the spell.
This way targets are incorrect and visuals are getting lost, because gameobjects in TC can't cast spells.
This causes many gobjects not to work properly.

**Expected behaviour:**
On Retail gameobjects are able to cast spells. Proven a lot of times in sniffs.

1. Use some newer unhacked fixed gameobjects which should cast a spell on you.

**Branch(es):**

3.3.5, master

**TC rev. hash/commit:** 

0f5b8447d0c82120e5ff58eb1931e80911017cdc

**TDB version:**  
TDB 3.3.5
TDB 7.2.0

**Operating system:** Win

ariel-:
**Changes proposed:**

- Implements long overdue changes to the spell system (e.g. Gameobject casting)
- Fixes some old bugs and clears the way to fix others
- Removed assertion from GetFactionTemplateEntry as it was being null-checked and still needs to be analyzed if Gameobjects need to have a faction defined
- Improved positive detection logic (positive/negative stuff was being used to mark spells as debuffs on UI only, but things like dealing damage were considered "positive")
- Split up target selection logic (Target check only determines what kind of entities we can target, spell positivity determines if that entity is valid for attack/assist) ie when targetting a negative spell to an ally
- Spells now only handle at most one UnitAura and one DynObjAura during its lifetime (caused issues with DR as aura objects were present on both target and caster)
- Iterate over effects first, then over targets, this is blizzlike and overdue change

- [x] 3.3.5
- [x] master (after it has been tested)

**Issues addressed:** 
Closes #8844 
Closes #15088
Closes #18395 
Closes #18885
Closes #21244 
Closes #21330 
Ref #18752 
Ref #19518 
Ref #20560
Ref #21142 
Ref #21330

**Tests performed:**
- Tested summon effects and summon RAF player, they all work properly across different maps
- Tested basic interaction with gameobjects (ie hunter traps) they work properly. All other cases of gameobjects (GOOBER. world traps, need to be tested and fixed, this only opens the way for those)
- Locally compared lists of positive/negative spells before and after the change, 

- [x] GameObjects without faction casting spells (see above description). Does it need to be a part of this PR?

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2671,7 +2671,8 @@ SpellMissInfo Spell::DoSpellHitOnUnit(Unit* unit, uint32 effectMask, bool scaleA
                 {
                     for (uint8 i = 0; i < MAX_SPELL_EFFECTS; ++i)
                     {
-                        if ((effectMask & (1 << i)) && !aurSpellInfo->IsPositiveEffect(i))
+                        // mod duration only for effects applying aura!
+                        if ((aura_effmask & (1 << i)) && !aurSpellInfo->IsPositiveEffect(i))
                         {
                             positive = false;
                             break;
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -806,9 +806,9 @@ void Spell::SelectSpellTargets()
             }
 
             uint8 mask = (1 << i);
-            for (std::list<TargetInfo>::iterator ihit = m_UniqueTargetInfo.begin(); ihit != m_UniqueTargetInfo.end(); ++ihit)
+            for (auto ihit = m_UniqueTargetInfo.begin(); ihit != m_UniqueTargetInfo.end(); ++ihit)
             {
-                if (ihit->effectMask & mask)
+                if (ihit->EffectMask & mask)
                 {
                     m_channelTargetEffectMask |= mask;
                     break;
@@ -818,20 +818,19 @@ void Spell::SelectSpellTargets()
         else if (m_auraScaleMask)
         {
             bool checkLvl = !m_UniqueTargetInfo.empty();
-            for (std::list<TargetInfo>::iterator ihit = m_UniqueTargetInfo.begin(); ihit != m_UniqueTargetInfo.end();)
+            m_UniqueTargetInfo.erase(std::remove_if(std::begin(m_UniqueTargetInfo), std::end(m_UniqueTargetInfo), [&](TargetInfo const& targetInfo) -> bool
             {
                 // remove targets which did not pass min level check
-                if (m_auraScaleMask && ihit->effectMask == m_auraScaleMask)
+                if (m_auraScaleMask && targetInfo.EffectMask == m_auraScaleMask)
                 {
                     // Do not check for selfcast
-                    if (!ihit->scaleAura && ihit->targetGUID != m_caster->GetGUID())
-                    {
-                         m_UniqueTargetInfo.erase(ihit++);
-                         continue;
-                    }
+                    if (!targetInfo.ScaleAura && targetInfo.TargetGUID != m_caster->GetGUID())
+                        return true;
                 }
-                ++ihit;
-            }
+
+                return false;
+            }), std::end(m_UniqueTargetInfo));
+
             if (checkLvl && m_UniqueTargetInfo.empty())
             {
                 SendCastResult(SPELL_FAILED_LOWLEVEL);
@@ -1232,11 +1231,11 @@ void Spell::SelectImplicitAreaTargets(SpellEffIndex effIndex, SpellImplicitTarge
         case TARGET_REFERENCE_TYPE_LAST:
         {
             // find last added target for this effect
-            for (std::list<TargetInfo>::reverse_iterator ihit = m_UniqueTargetInfo.rbegin(); ihit != m_UniqueTargetInfo.rend(); ++ihit)
+            for (auto ihit = m_UniqueTargetInfo.rbegin(); ihit != m_UniqueTargetInfo.rend(); ++ihit)
             {
-                if (ihit->effectMask & (1 << effIndex))
+                if (ihit->EffectMask & (1 << effIndex))
                 {
-                    referer = ObjectAccessor::GetUnit(*m_caster, ihit->targetGUID);
+                    referer = ObjectAccessor::GetUnit(*m_caster, ihit->TargetGUID);
                     break;
                 }
             }
@@ -1687,8 +1686,7 @@ void Spell::SelectImplicitTrajTargets(SpellEffIndex effIndex, SpellImplicitTarge
 
 void Spell::SelectEffectTypeImplicitTargets(uint8 effIndex)
 {
-    // special case for SPELL_EFFECT_SUMMON_RAF_FRIEND and SPELL_EFFECT_SUMMON_PLAYER
-    /// @todo this is a workaround - target shouldn't be stored in target map for those spells
+    // special case for SPELL_EFFECT_SUMMON_RAF_FRIEND and SPELL_EFFECT_SUMMON_PLAYER, queue them on map for later execution
     switch (m_spellInfo->Effects[effIndex].Effect)
     {
         case SPELL_EFFECT_SUMMON_RAF_FRIEND:
@@ -1696,11 +1694,31 @@ void Spell::SelectEffectTypeImplicitTargets(uint8 effIndex)
             if (m_caster->GetTypeId() == TYPEID_PLAYER && m_caster->GetTarget())
             {
                 WorldObject* target = ObjectAccessor::FindPlayer(m_caster->GetTarget());
-
                 CallScriptObjectTargetSelectHandlers(target, SpellEffIndex(effIndex), SpellImplicitTargetInfo());
 
-                if (target && target->ToPlayer())
-                    AddUnitTarget(target->ToUnit(), 1 << effIndex, false);
+                // scripts may modify the target - recheck
+                if (target && target->GetTypeId() == TYPEID_PLAYER)
+                {
+                    // target is not stored in target map for those spells
+                    // since we're completely skipping AddUnitTarget logic, we need to check immunity manually
+                    // eg. aura 21546 makes target immune to summons
+                    Player* player = target->ToPlayer();
+                    if (player->IsImmunedToSpellEffect(m_spellInfo, effIndex, nullptr))
+                        return;
+
+                    target->GetMap()->AddFarSpellCallback(std::bind([](Map* map, Spell* spell, uint8 effIndex, ObjectGuid const& targetGuid)
+                    {
+                        Player* player = ObjectAccessor::GetPlayer(map, targetGuid);
+                        if (!player)
+                            return;
+
+                        // check immunity again in case it changed during update
+                        if (player->IsImmunedToSpellEffect(spell->GetSpellInfo(), effIndex, nullptr))
+                            return;
+
+                        spell->HandleEffects(player, nullptr, nullptr, effIndex, SPELL_EFFECT_HANDLE_HIT_TARGET);
+                    }, std::placeholders::_1, this, effIndex, target->GetGUID()));
+                }
             }
             return;
         default:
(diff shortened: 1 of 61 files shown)


Treeston:
> Gameobject casting
mind blown

Treeston:
> Gameobject casting

mind blown

Treeston:
> Iterate over effects first, then over targets, this is blizzlike and overdue change

Can you explain in detail what bugs this fixed and/or how this changes behavior?

> GameObjects without faction casting spells (see above description). Does it need to be a part of this PR?

How does this work in the current core? Assuming it's not a regression (and you leave code in a state that is easily extended in a follow-up) I wouldn't say this needs to be in this PR.

ariel-:
> Can you explain in detail what bugs this fixed and/or how this changes behavior?

From https://github.com/TrinityCore/TrinityCore/issues/18395#issuecomment-264652281:
> It makes a huge difference if one of the effects is INSTAKILL and spell targets multiple units including caster (and caster being the target for INSTAKILL). Also in 7.x each effect can have different landing times

> How does this work in the current core? Assuming it's not a regression (and you leave code in a state that is easily extended in a follow-up) I wouldn't say this needs to be in this PR.

Has no effect on current core because gameobjects don't really cast spells. Faction used is the faction from the trigger who casts those on their behalf, and it's a hardcoded 14 for hostile spells and 35 for positive.

Now that gameobjects do cast, faction is used for reaction checks (IsHostile, IsFriendly), if faction == 0, the code defaults to neutral reaction (which may cause unintended behavior, that is why I removed the assertion but left the log, so it'll basically be a runtime log spam if a gameobject is supposed to cast but hasn't correct faction defined in DB)

Treeston:
How much effort would it be to replicate the 14/35 thing?

Treeston:
A lot of this PR appears to be `Unit*` -> `WorldObject*` refactors. Would it be feasible to split these off into a separate sub-PR (akin to #19966) to make reviewing the actual changes made easier?

ariel-:
I tried to split the PR in as much whole chunks as I could, naturally the gameobject casting is the bigger one (and is the one responsible of the Unit -> WorldObject refactor)

Treeston:
Is there any downside to having current TC simply pass `WorldObject*` around (even if it's only ever gonna be `Unit*` in practice?)

ariel-:
Code won't simply compile (ie it would require lots of `->ToUnit()`)

Treeston:
alright then

joschiwald:
what happens with all our spellscripts where caster is guarantee to be never null?

ariel-:
> what happens with all our spellscripts where caster is guarantee to be never null?

GameObjects only cast a handful of (as of now) unscripted spells, which should be handled gracefully by default effect handlers. I see no reason for any of the existing scripts to fail, as they're spells casted by players/creatures.

ariel-:
> How much effort would it be to replicate the 14/35 thing?

That was a hack, I don't intend to replicate it, what needs to be done is fix those gameobjects on a per-case basis

Treeston:
Yes, but as a fallback with error log to avoid regressions until we can fix them.

ariel-:
Error log already done here:
https://github.com/TrinityCore/TrinityCore/pull/21395/files#diff-29afb15ead2a8204cac07d94f074f260R2617

Ovahlord:
Does this PR also covering checking spell_attributes for each single effect? Like ignoring player_only attributes for area_entry target effects for example?

ariel-:
9106c0f27208cc2eb15626e975b929edbf62ad84?

joschiwald:
No, implementation of area aura targets is completely wrong like I
mentioned in several issues (and out of scope of this pr)

>

ariel-:
This is not touching area auras, only regular auras which have more than one target (for example Mind Control which applies a dummy on the caster and the posess aura on the enemy)

Faq:
TBD is planned in this PR?

tkrokli:
The Dark Below? I thought he meant "To Be Determined" 😆

ariel-:
If there's no opposition I'll merge this in one week

PS: @Laintime still waiting your constructive feedback after https://github.com/TrinityCore/TrinityCore/commit/cb9e72e521d3cc415dd15bf6912c87f89e41b92a#commitcomment-26866970

Seyden:
@joschiwald what do you say about it?
@ariel- i would let joschi recheck the PR, before you merge.. he is good in finding things :D

joschiwald:
@Seyden impossible to merge on master, too big commits

src/server/game/Spells/Spell.cpp
```diff
@@ -2225,312 +2238,373 @@ void Spell::AddItemTarget(Item* item, uint32 effectMask)
...
+    // Reset damage/healing counter
+    spell->m_damage = Damage;
+    spell->m_healing = Healing;
 
-    // Do not take combo points on dodge and miss
-    if (missInfo != SPELL_MISS_NONE && m_needComboPoints && m_targets.GetUnitTargetGUID() == target->targetGUID)
-        m_needComboPoints = false;
+    // Get original caster (if exist) and calculate damage/healing from him data
+    // Skip if m_originalCaster not available
+    Unit* caster = spell->m_originalCaster ? spell->m_originalCaster : spell->m_caster->ToUnit();
```
D0d0: this is problem when casting from gameobjects. When spell is created Spell.cpp:559 function '->toUnit()' returns 'NUL'. So the reset of this function is not executed -> no damge or cast from gameObject

jackpoz: are you working on a fix ?

jackpoz: @ariel- any idea about how to fix this ? should half of the Unit:: spell related functions be moved to WorldObject ?

ariel-: @jacpoz One of my original ideas was to create another entity (SpellCaster) from which both Unit and GameObject would inherit from. I don't remember why I didn't follow that path. Anyways, from what I can recall about this specifically: unit spellcasting functions deal with spells from units, while gameobject spell casting functions deal with spells from GOs.

The functionality is different by definition, and the GO casting is really limited to some trap spells on its majority. I say no to moving functions to WorldObject, because of those differences.

Laintime:
@ariel- thx am back)need you view)
https://github.com/TrinityCore/TrinityCore/issues/23031

Rushor:
wb @Laintime

src/server/game/Spells/Spell.cpp
jackpoz: @D0d0 do you have any detailed how to reproduce steps ?

D0d0: @jackpoz As far as I remember, just go to Heigan in Naxx and stay in green floor

Issue #19547, opened 2017-05-03
[3.3.5] Eternal Water Dupe exploit (works only with Eternal Water and other elements)


Tonghost:
**Description:**

This bug allows you to clone items

**Steps to reproduce the problem:**

1. .add 37705 11
2. Quickly right-click on an item. Try to press a second time before the item is deleted. You will need several attempts if server running on your localhost 
3. With a certain probability you will get two items (35622) instead of one

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

5cdbbc29ab

Demonid:
Confirmed on 5cdbbc29abdceab67af6f398279a81c426d0f758

I was actually going to report this a week ago but i forgot :P

jerbookins:
Guess you have to have pretty high ping for this to occur

jackpoz:
sounds like a nice case for a PR

jackpoz:
> With a certain probability you will get two items (35662) instead of one

is the item id correct ? http://reinosdeleyenda.org/aowow/?item=35662 seems quite unrelated

Tonghost:
Typo, thx. Correct id 35622

Jildor:
.die @xinef1
https://github.com/TrinityCore/TrinityCore/commit/464d3ad5d09d58136911f4ad09335f53049ac628

Foereaper:
https://www.youtube.com/watch?v=f32v7ld7CR8

Video showcasing the dupe, I'd look more into trying to fix it but it's a bit over my head.

Rochet2:
Main issue:
The spell the item casts to transform to another item has SPELL_ATTR2_IGNORE_ITEM_CHECK flag in DBC in AttributesEx2. This causes Spell::CheckItems to return before checking anything meaningful, so the reagents are not checked at all! This means that players could hack and send packets to do the transform too at any time without trying to "glitch it".

Removing the flag from the spell will make the core check the items and avoid the exploit. Not sure how to properly fix this issue though as removing the flag is probably not the right way.

Secondary issue: (not 100% sure on this)
The reagents are not checked before removing them since the spell is instant cast and cast(!m_casttime); causes skipCheck to be true in `Spell::_cast`. This is some kind of attempt at optimization to avoid checking reagents twice for instant spells. In Spell::prepare items are supposed to be already checked, but checking the items (Spell::prepare) and taking them from the player (Spell::_cast) are separated by a timed event, which means that the items could no longer be available when they are removed.
[...]

PR #21396, opened 2018-02-11
Core/Spells: Execute item casts immediately 

### LINKED ISSUE #11644: Aura Stuck
By simultaneously pressing two used trinkets and replacing them with other trinkets, buff them not disappear. In the original buff from the thing that is not on the character should not be (buff should disappear in a moment of change trinket).

4k+ spd and unreal DPS.

http://4put.ru/pictures/max/862/2648718.jpg <- 2 trinkets (use)
http://4put.ru/pictures/max/862/2648722.jpg <- change trinket and 4 auras.

/use Maghia's Misguided Quill
/equipslot 14 Charred Twilight Scale

/use Talisman of Resurgence
/equipslot 13 Phylactery of the Nameless Lich

Need skill to use 2 trinket before the fight at the together, without training can use only one of them because of GCD.

rev: https://github.com/TrinityCore/TrinityCore/commit/f7bc86a17cc0dd1c59c338539c1a7f1d7d6e5f64
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1617696-aura-stuck?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #15492: [3.3.5] Inconsistent behavior of simultanous trinket use
Branch 3.3.5, revision 6cdfe9dc095b, TDB 335.59.

When a character has two trinkets with an on-use buff equipped and uses one of them, the other is put on a shared cooldown equal to the duration of the used trinket's buff. This works as intended.

However, when both trinkets are (attempted to be) used in quick succession by a macro such as

```
/use 13
/use 14
```

the results are inconsistent between attempts. In most cases, the _second_ command (the slot 14 trinket) appears to be the one that is used, however I have also seen a case where _both_ trinkets managed to activate at the same time, which should never happen.

Expected behavior is that macro use commands are strictly evaluated in the order listed. If both trinkets are ready, the first line should always succeed (inventory slot 13 trinket should be used) and the second line should then subsequently fail due to the shared cooldown. The second line should only ever succeed if the first trinket is not ready.

### LINKED ISSUE #19524: [3.3.5] Core/Spell/Item: Item casts delayed

**Description:**

At the moment item casts are delayed as you can see by using an item and an ability in the same macro:

/use Item
/cast Ability

The item cast wont succeed but instead will be interrupted. 

The problem lies in

Spell::prepare
```
        if (!m_casttime && /*!m_spellInfo->StartRecoveryTime && */!m_castItemGUID && GetCurrentContainer() == CURRENT_GENERIC_SPELL)
            cast(true);
```
All item casts wont be executed immediately but launched from Spell::update, the problem at hand is when you cast a spell or use a second item directly after using an item the next spell will call Spell::prepare before the previous item spell calls Spell::update, therefore 
`m_caster->SetCurrentCastSpell(this);`
in Spell::prepare will interrupt the item cast.

You can solve the issue by removing '!m_castItemGUID' from the cast check in Spell::prepare,
the question is: Is this some old code remnant which isn't necessary anymore or is the comment 'second cast causes crash' still legit for some edge cases which i can't figure out. In all standard cases removing this part isn't causing any problems.

Issue causes:
#15492
Second item cast will interrupt the first one.

#11644
Cast will be launched after switching the items.

**Current behaviour:**

Item casts are delayed.

**Expected behaviour:**

Item casts shouldn't be delayed.

Use a macro like:

/use Item
/cast Ability

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

231ec8331b2ba3cded9f0d67c77c949cd45f5a60

TrinityCore rev. 231ec8331b2b+ 2017-04-28 19:12:48 -0300 (3.3.5 branch) (Win64, Debug, Static) (worldserver-daemon) ready...

**Operating system:** Win

Cannix:
**Changes proposed:**

-  Execute item casts immediately instead of in the next Spell:udpate call

- [x] 3.3.5
- [ ] master

Closes #19524
Closes #15492
Closes #11644

builds, tested in-game

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -3038,12 +3038,11 @@ void Spell::prepare(SpellCastTargets const& targets, AuraEffect const* triggered
         if (!(_triggeredCastFlags & TRIGGERED_IGNORE_GCD))
             TriggerGlobalCooldown();
 
-        //item: first cast may destroy item and second cast causes crash
         // commented out !m_spellInfo->StartRecoveryTime, it forces instant spells with global cooldown to be processed in spell::update
         // as a result a spell that passed CheckCast and should be processed instantly may suffer from this delayed process
         // the easiest bug to observe is LoS check in AddUnitTarget, even if spell passed the CheckCast LoS check the situation can change in spell::update
         // because target could be relocated in the meantime, making the spell fly to the air (no targets can be registered, so no effects processed, nothing in combat log)
-        if (!m_casttime && /*!m_spellInfo->StartRecoveryTime && */!m_castItemGUID && GetCurrentContainer() == CURRENT_GENERIC_SPELL)
+        if (!m_casttime && /*!m_spellInfo->StartRecoveryTime && */ GetCurrentContainer() == CURRENT_GENERIC_SPELL)
             cast(true);
     }
 }


Treeston:
What, if any, precautions do you take to avoid the issue in the comment you removed?
If none, why do you believe it is unnecessary to do so?

Cannix:
There seems to be no issue, i tested every possible item cast i could think of and it runs on my server for weeks without crashing. Its apparently an old mangos code remnant.

I opened #19524 almost a year ago and there seems to be no one who rly knows if its necessary so i decided to PR it.

Treeston:
My guess would be something like two separate stacks of Eternal Earth both right-clicked within a single update tick (or something like that).

Aokromes:
Plz, avoid to make Pull requests from main branches, create one branch for each feature.

jackpoz:
does it fix https://github.com/TrinityCore/TrinityCore/issues/19547 too ?

jackpoz:
Thanks for the PR :)

jackpoz:
I would say https://github.com/TrinityCore/TrinityCore/commit/720c3cbbd0f0b34284484d743801afe281e7272d made the check obsolete

Jildor:
>does it fix #19547 too ?
This issue I think not

jackpoz:
why not ?

Jildor:
Try to reproduce it with latency or  use .reload all command, I remember the exploit still

jackpoz:
did you do that already ingame after applying this PR or is it just speculation ?

Jildor:
Yep, I was apply this fix few months ago

PR #21400, opened 2018-02-12
[3.3.5][WIP] Spell maybe-pointer refactor

Treeston:
Currently, `Spell` (and related classes, such as `Aura`), hold a GUID + pointer for units that may go away - the pointer is updated on each tick from the GUID.

In this PR, @ariel- and myself codify this behavior into a new template class that I've dubbed `MemoryOf<T>`, which natively casts to `ObjectGuid` and `T*`, as well as (explicitly) to `bool`. Via operator overloads, it also supports the methods of both `ObjectGuid` (through `.`) and `T` (through `->`, not-`nullptr` asserted).

This allows us to pass `MemoryOf<T> const&` in place of `T*` through most of the call stack all the way into AI hooks, making original GUID information available even if the unit has gone away in the meantime.

**PR status:**
* `MemoryOf<T>` added
* `Spell` refactored
* `Aura` and derived in the progress of refactoring
* `SpellScript` untouched for now

_(**Note:** this PR may or may not build at any given point in time. Please limit feedback to components listed as finished.)_

### DIFF
--- a/src/server/game/AI/CoreAI/PassiveAI.cpp
+++ b/src/server/game/AI/CoreAI/PassiveAI.cpp
@@ -75,7 +75,7 @@ void PossessedAI::OnCharmed(bool /*apply*/)
     me->IsAIEnabled = false;
 }
 
-void CritterAI::DamageTaken(Unit* /*done_by*/, uint32&)
+void CritterAI::DamageTaken(MemoryOf<Unit> const& /*done_by*/, uint32&)
 {
     if (!me->HasUnitState(UNIT_STATE_FLEEING))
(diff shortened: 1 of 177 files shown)


tkrokli:
Uncertain if this PR classifies as **Core/Spells:** or **Core/Entities:**
(sometimes the deciding part is given by the content rather than the files).

jackpoz:
> even if the unit has gone away in the meantime

is this already implemented in the current code ? could you point at which lines handle ?

Shauren:
ObjectGuid inheritance...

Issue #20600, opened 2017-10-12
[3.3.5] Odd quest reputation rewards


PolarCookie:
**Description:**
Odd reputation rewards with 2 quest/repeatable-quest pairs, maybe more.
[The Sporelings' Plight](https://wow.gamepedia.com/Quest:The_Sporelings%27_Plight) rewarded 250 rep with Sporeggar (should be 750).
[More Spore Sacs](https://wow.gamepedia.com/Quest:More_Spore_Sacs) rewarded 750 rep with Sporeggar.
[Fierce Enemies](https://wow.gamepedia.com/Quest:Fierce_Enemies) rewarded 500 rep with Kurenai.
[More Warbeads](https://wow.gamepedia.com/Quest:More_Warbeads) rewarded 250 rep with Kurenai (should have been 500).

The odd part is that all 4 of these had RewardFactionValue1 set to 5 and RewardFactionOverride1 to 0 in the database.

**Branch(es):**
3.3.5

**TC rev. hash/commit:** 
TrinityCore rev. 1750ca501f7a 2017-10-11 00:02:37 +0200 (HEAD branch) (Win64, Debug, Static) (worldserver-daemon)

**TDB version:**
335.63

**Operating system:**
Win 7

Wyreth:
Can confirm.

Old Wowhead links:
- [The Sporelings' Plight](http://web.archive.org/web/20100409120414/http://www.wowhead.com/quest=9739)
- [More Warbeads](http://web.archive.org/web/20100411071755/http://www.wowhead.com:80/quest=10477)

Aokromes:
maybe it needs data on reputation_spillover_template

kaldown:
```
UPDATE quest_template SET RewardFactionValue1 = 7 WHERE id = 10476;
UPDATE quest_template SET RewardFactionValue2 = 7, RewardFactionID2 = 970 WHERE id = 9739;
```
It would be like:
Sporegarr increased erputation by 250
Sporegarr increased erputation by 500
Which in sum gives desired 750.
Will prepare screenshots.

kaldown:
Ok, so in bounded PR I've got some useful information, and @tkrokli asked me to move conversation here.
According https://community.trinitycore.org/topic/58-wdb-fields/, I have no rights to modify RewardFactionValueID, but I can do it with RewardFactionOverride.
The problem is, that when RewardFactionOverride used for a repratable quest (id=10476 from this issue), you wouldn't get bonuses. 
[...]

PR #21404, opened 2018-02-12
Issue/20600 PR

kaldown:
a PR for https://github.com/TrinityCore/TrinityCore/issues/20600

### DIFF
--- a/sql/updates/world/3.3.5/2018_02_12_01_world_335.sql
+++ b/sql/updates/world/3.3.5/2018_02_12_01_world_335.sql
@@ -1,0 +1,2 @@
+UPDATE quest_template SET RewardFactionValue1 = 7 WHERE id = 10476;
+UPDATE quest_template SET RewardFactionValue2 = 7 RewardFactionID2 = 970 WHERE id = 9739;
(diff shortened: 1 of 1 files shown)


tkrokli:
This is a so-called "SQL only" fix, post it in a code block on the issue page itself.
Pull Requests are for C++ code or C++ files _and_ SQL together (not SQL alone).

Aokromes:
Also, we don't change RewardFactionValue(x) without sniffs.

kaldown:
@Aokromes what is "sniff" ?

tkrokli:
- https://www.lifewire.com/introduction-to-packet-sniffing-2486803
- https://www.lifewire.com/what-is-a-packet-sniffer-2487312
- https://en.wikipedia.org/wiki/Packet_analyzer

kaldown:
Yeap, thanks. I kinda understand what is sniff, but why you use this word for apply sql patch?
I mean, what are you going to sniff?
Maybe I can do that, but I don't know what to search for and how.
I tested a output after db changes, asked some questions in IRC and thought it would be enough to solve an issue.

tkrokli:
What you need to do, if nobody has got that info already, is to use a trusted package sniffer (like the one on from TC forum pages) to produce sniffed packet data from when you hand in the quest on a retail server. Remember that there is a big amount of info packets received during game play, so you would need to limit the sniffer to singular interactions where you get the reputation change added to your character.

kaldown:
But there is no retail wotlk now. I can only try to sniff on a legion.
But I still quite unsure why to do this, because people say that it must be 750 reputation.
If, accepting this quest. Character reputation truelly increase by 750 (which will be printed out) - so why then we need a sniffer?

tkrokli:
Ask in the issue page. I usually follow what the owners ask of me, when dealing with requirements.

Aokromes:
It's outdated but we never change wdb fields without sniffs, and RewardFactionValue is one of them.
https://community.trinitycore.org/topic/58-wdb-fields/

tkrokli:
OK, I see. The columns used to be named `RewRepValue1` / `RewRepValue2` and `RewRepFaction2`.

PR #21423, opened 2018-02-15
Core/Unit: do not stop movement on root and freeze when target is affected by parabolic spline before call of affected functions

Ovahlord:
**Changes proposed:**

-  add a exception for root and freeze effects to not stop movement when the target is currently affected by parabolic splines like knockbacks, pulls or leaps.
There are some spells that do a pull and a root / freeze at the same time so with the current default handling the pull effect is getting supressed immediately. This fix here will handle this.
If you need an example spell: ID - 82168

- [x] 3.3.5
- [x] master

**Tests performed:**
- Tested ingame

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -11280,7 +11280,9 @@ void Unit::SetStunned(bool apply)
         SetTarget(ObjectGuid::Empty);
         SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_STUNNED);
 
-        StopMoving();
+        // Do not stop movement when the unit is affected by parabolic spline movement (knockbacks, pulls, scripted leaps etc.)
+        if (!movespline->isParabolic())
+            StopMoving();
 
         if (GetTypeId() == TYPEID_PLAYER)
             SetStandState(UNIT_STAND_STATE_STAND);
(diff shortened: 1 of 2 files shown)


ariel-:
Should this be handled by PendingXxxx spline flags?

Ovahlord:
What do you mean exactly?

ariel-:
In retail you receive a PendingXXX (e.g. PendingRoot spline flag) when you have a pending spline stop but it's not immediate because like your example, you are currently processing some controlled movement such as knockback, death grip, etc

Ovahlord:
I wonder if this wouldn't just lead to the same result. I mean the player cannot move anyways when finishing the parabolic move.

ccrs:
same result != correct way

tripleslash:
The way this is handled in the client is:

1. Server sends SMSG_FORCE_MOVE_ROOT to the client that is currently falling or in a spline
2. Client responds immediately with CMSG_FORCE_MOVE_ROOT_ACK with MOVEMENTFLAG_PENDING_ROOT set, but not MOVEMENTFLAG_ROOT set
3. Endpoint of spine is reached with CMSG_MOVE_SPLINE_DONE, the flag will still be MOVEMENTFLAG_PENDING_ROOT but not MOVEMENTFLAG_ROOT
4. CMSG_MOVE_SPLINE_DONE will always have MOVEMENTFLAG_FALLING set, because its impossible to stop a spline without being in falling state
5. Fall is ended by CMSG_MOVE_FALL_LAND in CGUnit_C::OnCollideFallLandNotify, which internally calls CGUnit_C::UpgradePendingRoot
6. After the fall is stopped (MOVEMENTFLAG_FALLING was removed) the server should verify that MOVEMENTFLAG_ROOT is set, and MOVEMENTFLAG_PENDING_ROOT was removed. If this is not true the client is cheating.

Fall can also be stopped by CMSG_MOVE_SET_FLY and CMSG_MOVE_START_SWIM, same checks should apply as in CMSG_MOVE_FALL_LAND.

PR #21469, opened 2018-02-21
Core/Entities: Phasing rewrite

### LINKED ISSUE #15163: [master] Core/Phases: base vmaps are used instead of phased (50$ bounty)
Rev: 8b3f424af8d0e645a579d07bb36d6b5ceef36450

When a npc is placed over a wmo in a phased map which does not exist in the base map the npc usually falls down as if no vmap was loaded. 

This may be due to base vmaps being used instead of the ones for the current phase.

### LINKED ISSUE #16758: [6.x] Phased areatriggers and gameobjects not working
**Current behaviour**:
Areariggers: When you trigger an areatrigger in other phase than 1, most of areatriggers are not triggered at all (verified by turning on .debug areatrigger and sniffs)
Gameobjects: Gameobjects in other phase than 1 which have a spell to be casted on use (i.e. GO 210986 with spell 114580), the spell is not casted at all. (verified by sniffs)

**Expected behaviour**: All areatriggers should be triggerable in any phase, all gameobjects should cast appropriate spell in any phase.

1. phase any area
2. try to trigger areatrigger/use gameobject (type 6 or 10)

**Branch(es)**: 6x

**TC hash/commit**: 52bb6485417494e58361a75fcf73cf5f0a1d09c5

**TDB version**: TDB 6.03

**Operating system**: Win 64

### LINKED ISSUE #16909: [6.x] Core/Terrain Swap: Areas using Terrain Swap have issues with npc pathfinding
**Description**:

**Current behaviour**: NPC falls through the ground after taking a few steps, not to be found again until a server restart happens.  It's as if the new swapped area is not decently being used by the pathfinding of the NPC's.

**Expected behaviour**: The NPC should just walk and not fall through the ground?

1. Enable Terrain Swap on a location like Lion's Landing, Dominion Point , this is done through the DB table terrain_swap_defaults and terrain_phase_info. (If you don't know the TerrainSwapMap id's, it's 1061 for Dominion Point and 1062 for Lion's Landing)
2. Go to said location that uses Terrain Swaps (like Lion's Landing, Dominion Point),.
3. Have any NPC move from their position (!cometome 1), they will fall through the ground after taking a few steps.

**Branch(es)**: 6x

**TC hash/commit**: https://github.com/TrinityCore/TrinityCore/commit/363731aac827f3eab051064ba466cef49f6cc111 I'm still using 6.2.3, reluctant to go 6.2.4 yet. But this would not be fixed even if I was on 6.2.4

**TDB version**: latest TDB 6.03

**Operating system**: Windows 10 Pro, 64-bit

### LINKED ISSUE #21119: PhaseShiftFlags 8 and 16
**Description:**
PhaseShiftFlags 8 and 16 are not implemented.

**Current behaviour:**
Hardcoded value in MiscPackets.cpp `_worldPacket << uint32(PhaseShifts.size() ? 0 : 8);`.
Unphased creatures and gameobjects disappear when player has at least one active phase.

**Expected behaviour:**
During my long observations on Wandering Isle, i found out that PhaseShiftFlag 8 could be "see unphased creatures" and PhaseShiftFlag 16 "see unphased gameobjects". There is no common phase in the zone (now I use a common phase for creatures and GOs in this PR #17021, but it is wrong).

In below quests (there are more than these I mention here) and related SMSG_PHASE_SHIFT_CHANGE packets, player doesn't have PhaseShiftFlag 8 and he does NOT see generic creatures (which I think should be unphased). In all other phases on Wandering Isle (and also in Durotar after completing Wandering Isle), player has PhaseShiftFlag 8 and 16.

[The Way of the Tushui](http://www.wowhead.com/quest=29414/the-way-of-the-tushui) - during this quest you have phase 838 and you can't see for example [Amberleaf Scamp](http://www.wowhead.com/npc=54130/amberleaf-scamp), [Corsac Fox](http://www.wowhead.com/npc=53704/corsac-fox), ...
Since phase 838 has PhaseShiftFlags 16 you should be able to see unphased GOs, like [The Dimming Flame](http://www.wowhead.com/object=209364/the-dimming-flame) and Scroll Post (entry: 209363) - the objects that Aysa sits before.
```
ServerToClient: SMSG_PHASE_SHIFT_CHANGE (0x137C) Length: 41 ConnIdx: 1 Time: 01/22/2016 19:26:31.578 Number: 38923
PhaseShiftFlags: 16
PhaseShiftCount: 2
PersonalGUID: Full: 0x0
[0] PhaseFlags: 0
[0] Id: 838
[1] PhaseFlags: 1
[1] Id: 1323
PreloadMapIDsCount: 0
UiWorldMapAreaIDSwap: 0
VisibleMapIDsCount: 2
[0] VisibleMapID: 683 (683)
```
[Battle for the Skies](http://www.wowhead.com/quest=29786/battle-for-the-skies) - during this quest (when taken) you have pahse 524 and when you complete or reward it you have phases 536 and 1526. You also can't see any generic creatures in the area. For example [Fe-Feng Wiseman](http://www.wowhead.com/npc=55601/fe-feng-wiseman)
```
ServerToClient: SMSG_PHASE_SHIFT_CHANGE (0x137C) Length: 41 ConnIdx: 1 Time: 01/22/2016 21:26:41.726 Number: 139150
PhaseShiftFlags: 16
PhaseShiftCount: 2
PersonalGUID: Full: 0x0
[0] PhaseFlags: 0
[0] Id: 524
[1] PhaseFlags: 1
[1] Id: 1326
PreloadMapIDsCount: 0
UiWorldMapAreaIDSwap: 0
VisibleMapIDsCount: 2
[0] VisibleMapID: 683 (683)

ServerToClient: SMSG_PHASE_SHIFT_CHANGE (0x137C) Length: 45 ConnIdx: 1 Time: 01/22/2016 21:29:07.009 Number: 140704
PhaseShiftFlags: 16
PhaseShiftCount: 3
PersonalGUID: Full: 0x0
[0] PhaseFlags: 0
[0] Id: 536
[1] PhaseFlags: 1
[1] Id: 1326
[2] PhaseFlags: 1
[2] Id: 1526
PreloadMapIDsCount: 0
UiWorldMapAreaIDSwap: 0
VisibleMapIDsCount: 2
[0] VisibleMapID: 683 (683)

ServerToClient: SMSG_PHASE_SHIFT_CHANGE (0x137C) Length: 41 ConnIdx: 1 Time: 01/22/2016 21:29:31.283 Number: 140816
PhaseShiftFlags: 16
PhaseShiftCount: 2
PersonalGUID: Full: 0x0
[0] PhaseFlags: 0
[0] Id: 536
[1] PhaseFlags: 1
[1] Id: 1327
PreloadMapIDsCount: 0
UiWorldMapAreaIDSwap: 0
VisibleMapIDsCount: 2
[0] VisibleMapID: 683 (683)
```

Just a thought: maybe we should have also phase_template table which would hold phase flags?

none

**Branch(es):** master

**TC rev. hash/commit:** e86a2c439aa2a032ce338d974dae1e5311bcb18f

**TDB version:** TDB 720.00

Shauren:
* Optimized phase visibility checking
* Handle all phase flags

PhaseUseFlags field in db has 2 possible flags
https://github.com/Shauren/TrinityCore/blob/e4d6c97620528aa0d63dc64c67837940e87b13d0/src/server/game/DataStores/DBCEnums.h#L825-L832

PHASE_USE_FLAGS_ALWAYS_VISIBLE - any phasing is ignored for this object, can be seen by everyone, no matter what phase they are in or this object is in
PHASE_USE_FLAGS_INVERSE - object can be seen if there is at least one phase which is NOT shared by this object and player(other object)

**Target branch(es):** master

**Issues addressed:**
Closes #15163
Closes #16758
Closes #16909
Closes #21119

**TODO list:**
- [x] map phasing
- [x] vmap phasing
- [x] mmap phasing


src/server/game/DataStores/DB2Stores.cpp
```diff
@@ -1922,13 +1922,14 @@ uint32 DB2Manager::GetQuestUniqueBitFlag(uint32 questId)
...
 
-std::set<uint32> DB2Manager::GetPhasesForGroup(uint32 group) const
+std::set<uint32> const& DB2Manager::GetPhasesForGroup(uint32 group) const
 {
     auto itr = _phasesByGroup.find(group);
     if (itr != _phasesByGroup.end())
         return itr->second;
 
-    return std::set<uint32>();
+    static std::set<uint32> const empty;
```
Shauren: TODO

Warpten: Doesn't `return {}` work (still horrible)

Shauren: I kind of want want to just keep the uses intact - `for (uint32 phaseId : sDB2Manager.GetPhasesForGroup(phaseGroupId))`

Warpten: what about returning iterators

DoctorKraft:
Nice works! `transports` table need maybe also phaseflag/terrainswap?

src/server/game/DataStores/DB2Stores.cpp
Shauren: IteratorPair? that would probably require default constructed iterators to be comparable (not all iterators work like that boost filesystem directory iterator where default constructed is .end)

Shauren:
@DoctorKraft could use phaseUseFlags but terrainSwapMap is unneded, creatures/gameobjects need it for vmaps and mmaps but transports have a fixed path and don't need it

Warpten:
Heads up: I might start working on proper editing of mmap/vmap tiles in 3D view soon ish (maybe this weekend if I'm not completely drained again)

src/server/game/DataStores/DB2Stores.cpp
Warpten: Adds boilerplate but I think that's the safest route if you don't want to go about rewriting range for loops...

runningnak3d:
@Shauren RE: mmap phasing. If you don't mind, can I ask is this where a creature is doing pathfinding, and changes phase so that path also changes, so they are then able to continue down the (now phased) path? 

It may seem like an obvious question, but my experience with mmaps is extremely lacking.

Shauren:
@runningnak3d creatures (not pets) never change their phase (which is why their terrain phase is a db field), its a matter of picking the swapped terrain (by map id) while remaining on a different map (tl;dr, loading vmap and mmap tiles for a different map id than normal)

Warpten:
Does recast support tile (dis|re) connection? @jackpoz

Shauren:
I planned to generate mesh for swapped maps using base map for missing tiles (in mmap generator), doing this in worldserver is too slow (together with missing vmap support is why I removed this code from master)

runningnak3d:
@Shauren Thank you very much for the insightful explanation -- it makes total sense now.

Warpten:
Pretty much what I thought would be best too

src/common/Collision/DynamicTree.h
```diff
@@ -41,11 +42,11 @@ class TC_COMMON_API DynamicMapTree
...
 
-    bool isInLineOfSight(G3D::Vector3 const& startPos, G3D::Vector3 const& endPos, std::set<uint32> const& phases) const;
-    bool getIntersectionTime(std::set<uint32> const& phases, G3D::Ray const& ray, G3D::Vector3 const& endPos, float& maxDist) const;
-    bool getObjectHitPos(std::set<uint32> const& phases, G3D::Vector3 const& startPos, G3D::Vector3 const& endPos, G3D::Vector3& resultHitPos, float modifyDist) const;
+    bool isInLineOfSight(G3D::Vector3 const& startPos, G3D::Vector3 const& endPos, PhaseShift const& phaseShift) const;
+    bool getIntersectionTime(G3D::Ray const& ray, G3D::Vector3 const& endPos, PhaseShift const& phaseShift, float& maxDist) const;
+    bool getObjectHitPos(G3D::Vector3 const& startPos, G3D::Vector3 const& endPos, G3D::Vector3& resultHitPos, float modifyDist, PhaseShift const& phaseShift) const;
 
-    float getHeight(float x, float y, float z, float maxSearchDist, std::set<uint32> const& phases) const;
+    float getHeight(float x, float y, float z, float maxSearchDist, PhaseShift const& phaseShift) const;
```
joschiwald: set include is unused now?

Palabola:
Vehicle system issues are also occured(and caused Crash) in old Phase system.

Shauren:
I present you: VMAP PHASING
1. no terrain
no_terrain_swap
2. .mod phase 0 976 on both caster and target
swap_976_both

DoctorKraft:
I have found something about phasing, idk if it's the right place. Gob with `GAMEOBJECT_TYPE_PHASEABLE_MO` used for vindicaar, garrison zone post, dalaran intro of legion etc.. are not displayed. They have a displayid etc.. A idea?

Shauren:
First time I hear about that :P very likely to be broken(-ish)

Shauren:
regular maps now swap too (see .gps in gilneas, the part that goes under water)

Shauren:
mmaps done

funjoker:
Can you give a short explanations of how to sue flags etc ?

Shauren:
PhaseUseFlags field in db has 2 possible flags
https://github.com/Shauren/TrinityCore/blob/e4d6c97620528aa0d63dc64c67837940e87b13d0/src/server/game/DataStores/DBCEnums.h#L825-L832

PHASE_USE_FLAGS_ALWAYS_VISIBLE - any phasing is ignored for this object, can be seen by everyone, no matter what phase they are in or this object is in
PHASE_USE_FLAGS_INVERSE - object can be seen if there is at least one phase which is NOT shared by this object and player(other object)

Thats it really

DJScias:
I'm going to take a wild guess and assume this probably would also fix #16909? Considering it's very similiar (or perhaps even the same) to #15163.

Shauren:
yes, it should fix that as well

Shauren:
@Lopfest fixed that extractor crash you were having

Ovahlord:
So far the test results:
- TempSummon* Map::SummonCreature seems to be missing a nullptr check for InheritPhaseShift. The summoner can be null when the map summons the creature so InheritPhaseShift is called and executed without any nullptr check.

- Player::SendQuestUpdate triggers PhasingHandler::OnConditionChange(WorldObject* object) The first
 if (!sConditionMgr->IsObjectMeetToConditions(srcInfo, *itr->AreaConditions)) is causing a crash with crash log message
`ConditionMgr::IsObjectMeetToConditions (this=0x3bb0bd8 <ConditionMgr::instance()::instance>, sourceInfo=..., conditions=<error reading variable: Cannot access memory at address 0x8>) `

Crash logs:
https://pastebin.com/MJhVb8Ax
https://pastebin.com/pa9U2z90
https://pastebin.com/cgpecyKn

Lopfest:
I was not able to extract mmaps this time :smile: 
```
91% [Map 0547] Building tile [13,29]
[Map 0547] [13,29]:  Building movemap tiles...
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.1
1.25503\include\vector(1796) : Assertion failed: vector subscript out of range
```

Ovahlord:
Test results for VMap swapping for documentation:
- The LOS swapping works fine. I did the following steps to test the functionality:
1. I want to Stonetalon Mountains to the Tal'Darah Grove where a Bomb is causing a terrainswap.
2. I spawned a training dummy in front of the big tree and went inside to test the LOS by trying to cast a spell: target not in LOS, so working as intended.
3. I changed the terrain now by using .mod phase 0 731 so I was able to see the terrain swap. I tried to cast on the same dummy: target not on LOS. Works fine because the dummy was using the LOS checks from the original terrain.
4. I spawned a new dummy while the terrain was swapped at the same location as the dummy and move to the same casting location: spell was using the new terrain LOS. Working fine.

---
Issues discovered while testing VMaps LOS:
-  I was using this query here to set up a terrain swap in the db:  https://paste2.org/O09KMXn4 
This database entry however has not shown any effect so terrain_swap_defaults is suspectet to be not working as intended right now.

Shauren:
@Lopfest do you have a more complete log of that? Map 547 worked just fine for me

Ovahlord:
More test results. This time for Map-Swapping:
- After some porting fails, the code differences has been solved and so far the swapping of maps seems to be working fine. Switching terrains does also trigger the liquid change. We used Gilneas as test map and used .mode phase 0 656 to sink duskhaven underwater. The liquid has been correctly changed and fatigue is applied properly as well.

Discovered issues:
- Graveyard locations are not getting updated. For example when drowning in the sunken Duskhaven and you release your spirit, you will be placed in the air where the original terrain's graveyard used to be. So you will pop up in mid air.

Lopfest:
crashlog from mmaps_generator.exe

[540548a066af_mmaps_generator.exe_[12-3_8-4-26].txt](https://github.com/TrinityCore/TrinityCore/files/1803442/540548a066af_mmaps_generator.exe_.12-3_8-4-26.txt)

Shauren:
I think I fixed that mmaps_generator crash (threading issue)

src/tools/vmap4_extractor/wdtfile.cpp
```diff
@@ -32,40 +33,45 @@ char * wdtGetPlainName(char * FileName)
 
 extern CASC::StorageHandle CascStorage;
 
-WDTFile::WDTFile(char* file_name, char* file_name1):WDT(CascStorage, file_name), gnWMO(0)
+WDTFile::WDTFile(char const* storagePath, std::string mapName, bool cache)
+    : _file(CascStorage, storagePath), _mapName(std::move(mapName))
 {
-    filename.append(file_name1,strlen(file_name1));
+    if (cache)
+    {
```
Ovahlord: Is this due to coding standard or why not use  std::make_unique here and skip the include of common here?

Shauren: including common is no longer bad since cleanup + fixing header dependencies (it no longer does crazy things like pull in mysql dep into extractors)
that and simply consistency - using the same thing everywhere

leviathonbeast:
is this safe to merge yet?

Ovahlord:
Not yet. There are still tests required. For example if a single target aura crash has been really reintroduced as I suspect and mmaps also need more tests

Ovahlord:
Allright, some more results:
- The Mmaps phasing works fine. I spawned a creature on normal terrain and on swapped terrain, both use their current terrain's pathfinding, so it is fine.
- While doing some tests in gilneas with the phasing auras, I have noticed that not all gameobjects and creatures are getting updated properly. Sometimes a gameobject or creature remains visible while the player should not see it. Getting fixed when getting out of range and back so the creatures are not getting re-send again. I think it's just a missing visbility updat call.

Ovahlord:
Here an example:
- The Rebel Lord's Arsenal is causing a aura change uppon completing but as you can see there, the main NPC that is in the previous phase is still visible while the new phases objects are also visible:
wowscrnshot_031418_190610
- Also here a list aura display so show that the character has only one phase ID due to an aura so a duplicate aura is not the case. This screenshot also includes the first cellar door, that should not be in the new phase as well.
wowscrnshot_031418_190928

Shauren:
Graveyards handled by 5bd61e91bf0a4530598a5969047d0374fa761ff8

Shauren:
things left to check
[ ] visibility update from auras
[ ] transports
[ ] vehicle aura crash

Ovahlord:
Yup. Can't find any other issues. If those are truly related to the phasing those points are the last ones from my side. Pretty sure that wont be other issues

Shauren:
transport issue isnt related to phasing

Ovahlord:
I found a few more evidences about the crash that I mentoined earlier:
- The crash seems to be caused by trying to delete childMaps of an already deleted map when teleporting in battlegrounds. According to my debugger the core enters  bool MapInstanced::DestroyInstance(InstancedMaps::iterator &itr)
before entering  Map::~Map() and somehow this leads to the consequence that delete m_childTerrainMaps; is returning a nullptr, I guess.

Shauren:
Fixed that m_childTerrainMaps crash

Ovahlord:
Allright, another point that >might< be related to the phasing: gameobject LOS. For example the lich king does not wanna move on his platform after aggro and his summon spells are summoning below the platform. Not sure if this is due to the vmaps changes or not.

Shauren:
That needs to be tested on master, not on 4.3.4 which gets most of stuff from 3.3.5 (including the imo horribly broken recent height checking)

Ovahlord:
Agreed. I mean, I updated those called checks to match the master functionality but sure, I can backport the master checks and test again. But still, a test on master would be also nice. I might know a few people who can test this

PR #21475, opened 2018-02-22
Core/Characters: Upgrade limitation to 16 characters per realm

DoctorKraft:
**Changes proposed:**

-  [Four Extra Character Slots Per Server Now Live](http://www.wowhead.com/news=281324/four-extra-character-slots-per-server-now-live)

**Target branch(es):**

- [X] master

**Tests performed:**

Tested IG & build


tkrokli:
Seems simple enough. Is it the correct way of doing it too?

DoctorKraft:
@tkrokli Haha 👍 It's only the client limitation, TC can support so more

PR #21476, opened 2018-02-22
Core/Players: Fix display for new race

DoctorKraft:
**Changes proposed:**

**MaleDisplayID** and **FemaleDisplayID** in ChrRaces are uint32 now
Fix new display for allied races

**Target branch(es):**

- [X] master

**Tests performed:**

Tested IG & build

**TODO**
After a expired polymorph or hex (SPELL_AURA_TRANSFORM), new race get a model broken for example Void Elf get tatoo and face of Lightforged Draenei. I have no idea about this, all others variable are correct.
- [ ] Fixed

### DIFF
--- a/src/server/game/Globals/ObjectMgr.h
+++ b/src/server/game/Globals/ObjectMgr.h
@@ -581,8 +581,8 @@ struct PlayerInfo
     float positionY;
     float positionZ;
     float orientation;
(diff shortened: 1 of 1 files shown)


DoctorKraft:
@Shauren Do you have an idea about the model broken after a polymorph or hex?

Issue #21461, opened 2018-02-19
[3.3.5] Mounted movement above indoor area (while still outdoors) dismounts instantly


Persephone:
**Description:**

No idea what commit caused this, just got it reported today and it was true lel. Running on a mount over the tunnel in WSG makes your char dismount.

**Current behaviour:**

Video: https://www.dropbox.com/s/m9yqiwwx9mxw2w2/IMG_0643.MOV?dl=0

**Expected behaviour:**

You should not dismount on terrain above "no mount" area. Only if you go inside that place

**Steps to reproduce the problem:**

1. .go xyz 1421.112183 1470.923828 349.519348 489
2. Mount up
3. Run forward towards the "lamp" thingie, once your above the tunnel below you insta dismount.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/1cc3d33cfe7e51d66925d5df536a952165a88094

**TDB version:**  64

**Operating system:** ubuntu

tkrokli:
If this is a recent problem, that it did not happen a week ago, check commit https://github.com/TrinityCore/TrinityCore/commit/9e0faace9a5114fc2324c2c601ba943272e0d6ff / **Core/Entities: Reduce the probability of units dropping under the map** #21322

Persephone:
Could be related to that i will run a test tomorrow

Gooyeth:
Duplicate #21338

PR #21479, opened 2018-02-22
Core/Object: Add IsOutdoors to Object and pass z with collision height

### LINKED ISSUE #21338: [VMaps] Mounts outdoor.
**Description:**

The vmaps does not recognize some outdoor areas.

**Current behaviour:**

Warsong Gulch: http://oi66.tinypic.com/fodq1h.jpg

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 1acfefb973895b0e3df3c8d9b43b920d2a1b5392

**TDB version:**  TDB 335.63

**Operating system:** Debian 9

Golrag:
**Changes proposed:**

- Add IsOutdoors to Object
- Add Collision height to to z

- 3.3.5
- master

**Issues addressed:** Closes #21338

**Tests performed:**
- Really quick and brief test in warsong gulch like in the issue

- Might be some more calls to check
- Not sure how or why our z is sometimes under the map maybe we should fix that, don't know

Not much time :(

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -6379,8 +6379,8 @@ void Player::CheckAreaExploreAndOutdoor()
     if (IsInFlight())
         return;
 
-    bool isOutdoor;
-    uint32 areaId = GetBaseMap()->GetAreaId(GetPositionX(), GetPositionY(), GetPositionZ(), &isOutdoor);
+    bool isOutdoor = IsOutdoors();
+    uint32 areaId = GetAreaId();
     AreaTableEntry const* areaEntry = sAreaTableStore.LookupEntry(areaId);
 
     if (sWorld->getBoolConfig(CONFIG_VMAP_INDOOR_CHECK) && !isOutdoor)
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2671,6 +2671,11 @@ float WorldObject::GetMapHeight(float x, float y, float z, bool vmap/* = true*/,
     return GetMap()->GetHeight(GetPhaseMask(), x, y, z, vmap, distanceToSearch, GetCollisionHeight());
 }
 
+bool WorldObject::IsOutdoors() const
+{
+    return GetMap()->IsOutdoors(GetPositionX(), GetPositionY(), GetPositionZ() + GetCollisionHeight());
+}
+
 template TC_GAME_API void WorldObject::GetGameObjectListWithEntryInGrid(std::list<GameObject*>&, uint32, float) const;
 template TC_GAME_API void WorldObject::GetGameObjectListWithEntryInGrid(std::deque<GameObject*>&, uint32, float) const;
 template TC_GAME_API void WorldObject::GetGameObjectListWithEntryInGrid(std::vector<GameObject*>&, uint32, float) const;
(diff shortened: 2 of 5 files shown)


Treeston:
So what is the functional impact of this change? We now check whether the top (rather than the bottom) of the character's model is outdoors?

Persephone:
Nice job @Golrag !! thanks

tkrokli:
@Persephone - have you tested this content change and found it to be working for you?

Persephone:
@tkrokli  will do give me 10 minutes

Golrag:
@Treeston yea iirc in the functions it calls it did some + 2's a bit scattered around, I removed some of those  in a previous PR where I noted that I might have removed too much of them. Now it's kinda back but in one function

Persephone:
@tkrokli  I compiled, and am online locally, i still dismount if i run over the tunnel at alliance base and horde base :/

Persephone:
Checkcast only executes when a spell is cast right? I get dismounted while running over the tunnel on top of it @Golrag

Persephone:
I guess the issues werent duplicated then, the one this one fixes is appearntly when you try to use the spell, the issue i posted is while you are mounted you get dismounted when you run over the tunnel. I added a video in the original issue:

https://github.com/TrinityCore/TrinityCore/issues/21461

Golrag:
I badly documented my test case in the description I checked before changes with gps, returned indoors, after it returned outdoors. I kinda forgot where mounts are on 3.3.5 lol so i didn't really test that

Persephone:
Yeah the gamebreaking issue is moving on a mount dismounts you, no warning or anything, something just calls "dismount" :/ Would you have any idea on that perhaps?

Golrag:
Player::CheckAreaExploreAndOutdoor

Persephone:
@Golrag  if i set vmap.enableIndoorCheck = 1 to 0 i dont get dismounted when running over the tunnel, but then i can run inside the tunnel with a mount tho :P

Golrag:
This should have fixed it, thanks for testing :)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -6379,8 +6379,8 @@ void Player::CheckAreaExploreAndOutdoor()
     if (IsInFlight())
         return;
 
-    bool isOutdoor;
-    uint32 areaId = GetBaseMap()->GetAreaId(GetPositionX(), GetPositionY(), GetPositionZ(), &isOutdoor);
```
Treeston: Hrm. What is the call stack on this? It being included in `GetAreaId` implies that we get this from the vmap lookup... so we can just fetch it in the liquid state update, which happens on relocation for all units anyway.

I will investigate.

Persephone:
@Golrag Neaaat!! will test :D

src/server/game/Entities/Player/Player.cpp
Golrag: This whole function is called in Player::UpdatePosition
Which checks for m_needsZoneUpdate and updates the zone if set to true (I did not check if that implementation is correct tho)

Persephone:
confirmed it works now @Golrag  :)!

Persephone:
Is this getting merged?

tkrokli:
Sooner or later, yes. When TC owners or members who can tell if it is good enough finds it ready to be merged.

Treeston:
I don't like the extra vmap lookup on every relocate. We already do one (for the liquid update). I'll need to look into whether we can get the outdoors state from that same lookup.

Golrag:
@Treeston if that's the only thing, can't we merge it already ? I didn't really add any extra vmap lookup in this PR.
GetAreaId() returns the cached one. Or am I completely missing something here ? :p

Shauren:
Pretty sure he wants outdoor status to be added to `WorldObject::UpdatePositionData`

PR #21493, opened 2018-02-24
Core/DataStores: Set default appearance if one does not exist

### LINKED ISSUE #21483: Some items not showing in character selection
**Description:**

`Item::GetDisplayId` returns 0 for some items.

Causing some specific items do not show up in char selection screen.
Example item: http://www.wowhead.com/item=116244/spaulders-of-the-iron-crucible

This issue may be appearing only on/after 7.3.5.25996
It seems they removed some (possibly duplicate data) rows from `ItemModifiedAppearance.db2` in the update on this patch. Rows that used to have 0,1,4 appearancemodid may now only have 4.

**Current behaviour:**

Item::GetDisplayId returns 0 for some items.
This causes the item to not appear in character screen in character selection window when logging in, since 0 is saved to DB (characters.equipmentCache) as displayid to show on char screen.

**Expected behaviour:**

Item::GetDisplayId should return the displayid of the item.
The shoulder item should show in char selection screen.

1. create draenei paladin and log in
2. use .level 110
3. use .additem 116244
4. equip the item and log out
5. notice there is no shoulders equipped in char screen

**Branch(es):** master

**TC rev. hash/commit:** 8467503b90915e95ae4270e30a01a0cb00715db1

**TDB version:**  735.00_2018_02_19 + updates

**Operating system:** windows 10

Rochet2:
**Changes proposed:**

-  Sets default appearance if one does not exist
-  Uses lowest appearancemodid of the ones available

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #21483

- builds
- Fixes the issue of char selection screen not showing the equipped item (tested item 116244)
- Fixes aquiring transmogrification and transmogrifying to the item (tested item 116244)

- [ ] I dont like the implicitness of this edit
- [ ] Unsure if using lowest modid is correct way to handle getting default appearance
- [ ] Unsure if we should fill in the other appearances too, like appearancemodid 1 that used to exist.

### DIFF
--- a/src/server/game/DataStores/DB2Stores.cpp
+++ b/src/server/game/DataStores/DB2Stores.cpp
@@ -855,6 +855,10 @@ void DB2Manager::LoadStores(std::string const& dataPath, uint32 defaultLocale)
     {
         ASSERT(appearanceMod->ItemID <= 0xFFFFFF);
         _itemModifiedAppearancesByItem[appearanceMod->ItemID | (appearanceMod->AppearanceModID << 24)] = appearanceMod;
+
+        auto it = _itemModifiedAppearancesByItem.find(appearanceMod->ItemID);
+        if (it == _itemModifiedAppearancesByItem.end() || appearanceMod->AppearanceModID < it->second->AppearanceModID)
+            _itemModifiedAppearancesByItem[appearanceMod->ItemID] = appearanceMod;
     }
 
     for (ItemSetSpellEntry const* itemSetSpell : sItemSetSpellStore)


Shauren:
Something like this existed in past but it's not correct, it breaks either artifacts or some LFR versions of tier sets

Rochet2:
Would be great to have some examples of items that were broken with the old code or could be broken.

I compiled a list of item entries and their appearancemodids that this PR affects:
https://pastebin.com/mFK5uwKq
But there are quite a lot of them to check all of them one by one.

Shauren:
Not sure if this is a bug anymore - example item given in issue has only a single source - LFR. Need to double check bonuses coming from ItemBonusTreeNode.db2 for BonusTreeModID = 4 (thats the LFR mod)

Might work correctly if obtained from its intended source (for TC you would need to enter that map on that difficulty and loot it, not .add)

PR #21521, opened 2018-02-28
Core/Entities: update Scaling to 7.3.5

funjoker:
**Changes proposed:**

-  Scaling creatures have a range in delta
- [#351](https://github.com/TrinityCore/WowPacketParser/pull/351)

- [x] master

- [x] ~~ScalingMin and Max might be taken from SandboxScaling.db2 since SandboxScalingId is part of Unit~~ _In BfA DB2 got renamed (it's only for item,quest,worlquest scaling_

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -1226,20 +1226,22 @@ void Creature::SelectLevel()
 {
     CreatureTemplate const* cInfo = GetCreatureTemplate();
 
-    if (!HasScalableLevels())
+    // level
+    uint8 minlevel = std::min(cInfo->maxlevel, cInfo->minlevel);
+    uint8 maxlevel = std::max(cInfo->maxlevel, cInfo->minlevel);
+    uint8 level = minlevel == maxlevel ? minlevel : urand(minlevel, maxlevel);
+    SetLevel(level);
+
+    if (HasScalableLevels())
     {
-        // level
-        uint8 minlevel = std::min(cInfo->maxlevel, cInfo->minlevel);
-        uint8 maxlevel = std::max(cInfo->maxlevel, cInfo->minlevel);
-        uint8 level = minlevel == maxlevel ? minlevel : urand(minlevel, maxlevel);
-        SetLevel(level);
-    }
-    else
-    {
-        SetLevel(cInfo->levelScaling->MaxLevel);
         SetUInt32Value(UNIT_FIELD_SCALING_LEVEL_MIN, cInfo->levelScaling->MinLevel);
         SetUInt32Value(UNIT_FIELD_SCALING_LEVEL_MAX, cInfo->levelScaling->MaxLevel);
-        SetUInt32Value(UNIT_FIELD_SCALING_LEVEL_DELTA, cInfo->levelScaling->DeltaLevel);
+
+        int8 mindelta = std::min(cInfo->levelScaling->DeltaLevelMax, cInfo->levelScaling->DeltaLevelMin);
+        int8 maxdelta = std::max(cInfo->levelScaling->DeltaLevelMax, cInfo->levelScaling->DeltaLevelMin);
+        int8 delta = mindelta == maxdelta ? mindelta : irand(mindelta, maxdelta);
+
+        SetInt32Value(UNIT_FIELD_SCALING_LEVEL_DELTA, delta);
     }
 
     UpdateLevelDependantStats();
@@ -2566,7 +2568,7 @@ uint8 Creature::GetLevelForTarget(WorldObject const* target) const
         // between UNIT_FIELD_SCALING_LEVEL_MIN and UNIT_FIELD_SCALING_LEVEL_MAX
         if (HasScalableLevels())
         {
-            uint8 targetLevelWithDelta = unitTarget->getLevel() + GetCreatureTemplate()->levelScaling->DeltaLevel;
+            uint8 targetLevelWithDelta = unitTarget->getLevel() + GetInt32Value(UNIT_FIELD_SCALING_LEVEL_DELTA);
 
             if (target->IsPlayer())
                 targetLevelWithDelta += target->GetUInt32Value(PLAYER_FIELD_SCALING_PLAYER_LEVEL_DELTA);
--- a/sql/updates/world/master/2018_04_12_00_world.sql
+++ b/sql/updates/world/master/2018_04_12_00_world.sql
@@ -1,0 +1,5 @@
+ALTER TABLE `creature_template_scaling`
+ CHANGE `LevelScalingDelta` `LevelScalingDeltaMin` smallint(5) NOT NULL DEFAULT '0' AFTER `LevelScalingMax`,
+ ADD `LevelScalingDeltaMax` smallint(5) NOT NULL DEFAULT '0' AFTER `LevelScalingDeltaMin`;
+
+UPDATE `creature_template_scaling` SET `LevelScalingDeltaMax`=`LevelScalingDeltaMin`;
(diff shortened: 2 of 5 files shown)


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1231,20 +1231,22 @@ void Creature::SelectLevel()
...
-        uint8 maxlevel = std::max(cInfo->maxlevel, cInfo->minlevel);
-        uint8 level = minlevel == maxlevel ? minlevel : urand(minlevel, maxlevel);
-        SetLevel(level);
-    }
-    else
+    // level
+    uint8 minlevel = std::min(cInfo->maxlevel, cInfo->minlevel);
+    uint8 maxlevel = std::max(cInfo->maxlevel, cInfo->minlevel);
+    uint8 level = minlevel == maxlevel ? minlevel : urand(minlevel, maxlevel);
+    SetLevel(level);
```
funjoker: Sniffes confirmed, levels are allways set.
Did not check legion creatures yet

Seyden: I don't know how it is in 735 but.. in 7.2 and lower.. if the creature got a scaling level, his NORMAL level was always MAX.. which you removed now

funjoker: Normal Level in sniffs is Standard randomized min/max Level

funjoker:
might need a total rework

funjoker:
The gap between min and maxlevel in template describes the gap between deltamin and deltamax.
(only for rank 0 npc)
Example:
minlevel = 35
maxlevel = 38

gap = 38-35 = 3

=> this means the gap at delta must be 3 too. It seems like blizzard randomized the delta around the basedelta 0. So there are 2 possible deltadata which could accure.

DeltaOption 1: 
DeltaMin = -2
DeltaMax = 1

DeltaOption 2:
DeltaMin = -1
DeltaMax = 2

All delta combinations i've found are just following this "gap rule".
Class, type, family does not effect it.

The question now is:
Does Blizzard randomize their deltas by every start or do they have fixed values for it?


For Creatures which scale to maxlevel 20 and have rank  > 0 the deltamin and max is 1
For Creatures which scale to maxlevel 60 and have rank  > 0 the deltamin and max is 2
Missing other data for now

It also seems like the min and max is taken by SandboxScaling.db2
The flag 4 is handled serversidely. Flag 2 means deltamax = deltamin

src/server/game/Entities/Creature/Creature.cpp
Traesh: Now that's a not totally useless and duplicate data database scheme

Traesh:
@funjoker Flag 4 seem to say no scaling at all, creatures come back to legacy min/max system

Traesh:
@funjoker Flag 4 do not scale at all, levels are set just as before, which is weird

src/server/game/Entities/Creature/Creature.cpp
Traesh: This is wrong, levels are not random between min & max, all outland creature UNIT_FIELD_LEVEL (so scaling 64-80, 1690 query) where between 64 & 68, doesn't seem random to me

Shauren:
I agree with flag 4, see the check here based on client - https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Item/Item.cpp#L2195

funjoker:
@Traesh i know. but still this flag is used for something. maybe it's just to keep.

Correction to Flag 2...

```
__int64 __fastcall GetMaxSandboxScalingLevel(DB2::Struct::SandboxScaling *row)
{
  DWORD maxLevel; // ebx
  unsigned __int64 expansion; // rdx

  maxLevel = row->MaxLevel;
  if ( row->Flags & 2 )
  {
    expansion = (unsigned __int8)GetServerExpansion();
    if ( expansion >= 7 )
      LODWORD(expansion) = 6;
    maxLevel += *(_DWORD *)&sandboxScalingMaxLevelMagic[/*2 * */(signed int)expansion];
  }
  return maxLevel;
}
```
```
sandboxScalingMaxLevelMagic[0] = 60 // Vanilla
sandboxScalingMaxLevelMagic[1] = 80 // TBC
sandboxScalingMaxLevelMagic[2] = 80 // WotLK
sandboxScalingMaxLevelMagic[3] = 90 // Cataclysm
sandboxScalingMaxLevelMagic[4] = 90 // MoP
sandboxScalingMaxLevelMagic[5] = 100 // WoD
sandboxScalingMaxLevelMagic[6] = 110 // Legion

```

Flag 2 sets maxlevel depending on server expansion -> So with BfA they scale to 120 automatically?

```
        [Parser(Opcode.SMSG_INITIAL_SETUP)]
        public static void HandleInitialSetup(Packet packet)
        {
            packet.ReadByte("ServerExpansionLevel"); // <-- this is not the level anymore, this is the ExpensansionId
            packet.ReadByte("ServerExpansionTier");
        }
```

ServerExpansionLevel  must be ExpansionId

funjoker:
I think SandboxScaling.db2 is redundant with ScalingData

Shauren:
fyi the name "expansionlevel" does not refer to character level (and never did)

funjoker:
rly? well the name is then a bit irritating.

Anyway that's all i could find on informatio.

What do you think about the randomized DeltaRange? So the server sets the DeltaRange randomly around 0. Because i couldn't find any info that would "creatures witch class xyz starts at -1 DeltaMin".
Also i couldn't find a proof that blizz sets the delta randomly on every startup :/

funjoker:
@Traesh levels are set like no scaling is there => that's why i changed it in this PR.

Might change a bit more so made it [WIP] because i got some ideas and theories i need to proof first :/

src/server/game/Entities/Creature/Creature.cpp
funjoker: Traesh.. 
UNIT_FIELD_LEVEL is untouched by blizzard. This Field is Set like no scaling exist and that is What i say since this PR is opened.
We are talking about normal Level and not scaling.
In the actual scaling System Level is similar to scalingmax and that would be wrong. That's why i Set it that Level is Set Normaly.

Traesh: Sure, perfectly understand why you change it and you're right when you say it's not equal to maxlevel anymore

I'm just saying that in the full outland sniff i have, all UNIT_FIELD_LEVEL are between 64 and 68, so they are not totally random, because if they were indeed random they would statistically be between 64 and 80

The other option is that Blizzard stopped setting UNIT_FIELD_LEVEL to max scaling level, and updated all standard creature_template min&max level to 64-68, for no reason at all. Which would be very weird (but still possible when it come to Blizzard)

funjoker: I dont think they updated ay template levels.. 
Couldnt find anything in whole azeroth

funjoker: > Blizzard stopped setting UNIT_FIELD_LEVEL to max scaling level

They never did, it just seemed to be like that, because Legion Creatures did never have any minmaxlevel because they were never unscaled.

funjoker: > and updated all standard creature_template min&max level to 64-68

Can be possible, but unsure.
They did not touch any levels at Kalimdor or Eastern Kingdoms.

funjoker:
> I'm just saying that in the full outland sniff i have, all UNIT_FIELD_LEVEL are between 64 and 68, so they are not totally random, because if they were indeed random they would statistically be between 64 and 80

I just checked that and you are wrong. UNIT_FIELD_LEVEL didn't change.
There are only few things changed...

The have ScalingId and they have ScalingDatas.

And Delta is not a FIX value so there is a DeltaMin and a DeltaMax.

PR #21533, opened 2018-03-02
[3.3.5] Transitional Cryptography update for OpenSSL 1.1

Warlockbugs:
Support for both OpenSSL 1.0 LTS and OpenSSL 1.1 versions.

Many Linux distributions are still on 1.0 and will stay on LTS for quite
some time.

Direct port of my CMaNGOS commit: https://github.com/cmangos/mangos-wotlk/commit/e1b0048f052eda46bb27d20224d0339960816ac2

**Changes proposed:**

-  Adds transitional hybrid support for OpenSSL 1.0.x and 1.1.x versions.
-  Allows use and development in cutting edge Linux distros out of the box.

- [x] 3.3.5

This is a direct port, only build was performed.

None to date.

### DIFF
--- a/src/common/Cryptography/HMACSHA1.cpp
+++ b/src/common/Cryptography/HMACSHA1.cpp
@@ -21,38 +21,53 @@
 #include "Errors.h"
 #include <cstring>
 
+#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER < 0x10100000L
+HMAC_CTX* HMAC_CTX_new()
+{
+    HMAC_CTX *ctx = new HMAC_CTX();
+    HMAC_CTX_init(ctx);
+    return ctx;
+}
+
+void HMAC_CTX_free(HMAC_CTX* ctx)
+{
+    HMAC_CTX_cleanup(ctx);
+    delete ctx;
+}
+#endif
+
 HmacHash::HmacHash(uint32 len, uint8* seed)
 {
-    HMAC_CTX_init(&m_ctx);
-    HMAC_Init_ex(&m_ctx, seed, len, EVP_sha1(), nullptr);
+    m_ctx = HMAC_CTX_new();
+    HMAC_Init_ex(m_ctx, seed, len, EVP_sha1(), nullptr);
     memset(m_digest, 0, sizeof(m_digest));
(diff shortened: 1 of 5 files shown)


Shauren:
Interesting but try that on master branch - I expect it to fail because of gsoap and boost asio ssl intergration

jackpoz:
> This is a direct port, only build was performed.

Please test it ingame too

src/common/Cryptography/HMACSHA1.cpp
```diff
@@ -23,36 +23,61 @@
...
+    HMAC_CTX_free(m_ctx);
+#else
     HMAC_CTX_cleanup(&m_ctx);
+#endif
 }
 
 void HmacHash::UpdateData(std::string const& str)
 {
+#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x10100000L
+    HMAC_Update(m_ctx, (uint8 const*)str.c_str(), str.length());
```
jackpoz: most of the changes seems related to adding a "&" for the old version, maybe there is a better way to handle it than having the same preprocessor check "#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x10100000L" over and over across the whole file

jackpoz (changes requested):

src/common/Cryptography/HMACSHA1.cpp
Warlockbugs: The intent of such a patch was to be a transitional, not a permanent solution, so it was a concious decision to not bend any class APIs to a single dependency's quirks. It is elaborate enough to not raise any questions about what happened there after some time passes, ridiculously easy to revert, easy to adapt to any particular OpenSSL API changes on per-case basis.
While it doesnt look pretty, there is no denying it, many FOSS projects who are in migration/transition to new OpenSSL versions chose the preprocessor solution to stick with.

Shauren: @jackpoz don't complain, after all we have this monster too dfd2660a85e4f0891c63009ee8425b2796586409

jackpoz (approved):

PR #21535, opened 2018-03-02
Core/Player: fix HasItemFitToSpellRequirements for SPELL_ATTR8_ARMOR_SPECIALIZATION

Traesh:
**Changes proposed:**

-  Fix HasItemFitToSpellRequirements when it come to SPELL_ATTR8_ARMOR_SPECIALIZATION

- [x] master

**Issues addressed:** ?

**Tests performed:**
Tested with spell 89744, didn't apply before, apply now

- [x] Spell don't get removed if you remove a required item

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -24666,7 +24666,7 @@ bool Player::HasItemFitToSpellRequirements(SpellInfo const* spellInfo, Item cons
                 for (uint8 i : {EQUIPMENT_SLOT_HEAD, EQUIPMENT_SLOT_SHOULDERS, EQUIPMENT_SLOT_CHEST, EQUIPMENT_SLOT_WAIST, EQUIPMENT_SLOT_LEGS, EQUIPMENT_SLOT_FEET, EQUIPMENT_SLOT_WRISTS, EQUIPMENT_SLOT_HANDS})
                 {
                     Item* item = GetUseableItemByPos(INVENTORY_SLOT_BAG_0, i);
-                    if (!item || !item->IsFitToSpellRequirements(spellInfo))
+                    if (!item || item == ignoreItem || !item->IsFitToSpellRequirements(spellInfo))
                         return false;
                 }
 
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -24669,6 +24669,8 @@ bool Player::HasItemFitToSpellRequirements(SpellInfo const* spellInfo, Item cons
                     if (!item || !item->IsFitToSpellRequirements(spellInfo))
                         return false;
                 }
+
+                return true;
             }
 
             break;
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -24669,6 +24669,8 @@ bool Player::HasItemFitToSpellRequirements(SpellInfo const* spellInfo, Item cons
                     if (!item || !item->IsFitToSpellRequirements(spellInfo))
                         return false;
                 }
+
+                return true;
```
joschiwald: according to the logic of the other cases in that function and the functionname, it should be:
```cpp
for (uint8 i : {EQUIPMENT_SLOT_HEAD, EQUIPMENT_SLOT_SHOULDERS, EQUIPMENT_SLOT_CHEST, EQUIPMENT_SLOT_WAIST, EQUIPMENT_SLOT_LEGS, EQUIPMENT_SLOT_FEET, EQUIPMENT_SLOT_WRISTS, EQUIPMENT_SLOT_HANDS})
{
    if (Item* item = GetUseableItemByPos(INVENTORY_SLOT_BAG_0, i))
        if (item != ignoreItem && item->IsFitToSpellRequirements(spellInfo))
            return true;
}
```

Traesh: This is a special case where all slots must have a specific item type, so the logic here is good, it was only missing a specific return true; at the end

Another spell example : http://www.wowhead.com/spell=86538

joschiwald: ok, got it
about your known issues, try this:
```diff
diff --git a/src/server/game/Entities/Player/Player.cpp b/src/server/game/Entities/Player/Player.cpp
index 5c28e7b72b..240d0668b8 100644
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -24666,9 +24666,11 @@ bool Player::HasItemFitToSpellRequirements(SpellInfo const* spellInfo, Item cons
                 for (uint8 i : {EQUIPMENT_SLOT_HEAD, EQUIPMENT_SLOT_SHOULDERS, EQUIPMENT_SLOT_CHEST, EQUIPMENT_SLOT_WAIST, EQUIPMENT_SLOT_LEGS, EQUIPMENT_SLOT_FEET, EQUIPMENT_SLOT_WRISTS, EQUIPMENT_SLOT_HANDS})
                 {
                     Item* item = GetUseableItemByPos(INVENTORY_SLOT_BAG_0, i);
-                    if (!item || !item->IsFitToSpellRequirements(spellInfo))
+                    if (!item || item == ignoreItem || !item->IsFitToSpellRequirements(spellInfo))
                         return false;
                 }
+
+                return true;
             }
 
             break;
```

Traesh: Worked indeed, thanks @joschiwald

PR #21554, opened 2018-03-04
Core/Map: fix outdoors logic part 2

Golrag:
**Changes proposed:**

-  Change it to how it works in GetAreaId, which means that IsOutdoorWMO would only be called when GetAreaInfo returned true (which is kinda what happens in this check)

- 3.3.5
- master

**Tests performed:**
- Running around in warsong gulch with .gps and checking if outdoors/indoors was correct

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -2686,9 +2686,14 @@ void Map::GetFullTerrainStatusForPosition(float x, float y, float z, PositionFul
     //  - the vmap floor is above the gridmap floor
     // or
     //  - we are below the gridmap floor
+    data.outdoors = true;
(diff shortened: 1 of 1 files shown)


Shauren:
but now you put it under Z check...

Golrag:
Which kinda is what the return value of GetAreaInfo was if I understood it correctly

Persephone:
this fixed the checks tho, dismounts properly in houses in wsg but not above tunnel.

Golrag:
@Persephone to me your phrasing is a bit misleading at first, so just to be sure, are you telling me you are NOT dismountend when above the tunnel and correctly dismountend in the houses ?

@Shauren could you explain a bit more in detail what should happen now ? I just tried to make it behave like before with the GetAreaId (with the indoors param) which I think it is like that now

Persephone:
@Golrag Yes. Right now i am correctly dismounted INSIDE houses and tunnel, and i am not dismounted ABOVE the tunnel :) But i do get A LOT of "error unable to load mmap/vmap for X map/area" etc. Is something wrong with vmaps/mmaps?

PR #21606, opened 2018-03-13
Core/Battleground: resolve auto distribution of Arena Points

Gooyeth:
**Changes proposed:**

Fixed the timer in the distribution, previously every week the time increased changing the distribution time

- [x] 3.3.5
- [x] master

### DIFF
--- a/src/server/game/Battlegrounds/BattlegroundMgr.cpp
+++ b/src/server/game/Battlegrounds/BattlegroundMgr.cpp
@@ -171,7 +171,7 @@ void BattlegroundMgr::Update(uint32 diff)
             if (GameTime::GetGameTime() > m_NextAutoDistributionTime)
             {
                 sArenaTeamMgr->DistributeArenaPoints();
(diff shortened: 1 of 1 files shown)


jackpoz:
could you explain a bit more in detail how the 2 variables end up holding different values ?

src/server/game/Battlegrounds/BattlegroundMgr.cpp
```diff
@@ -171,7 +171,7 @@ void BattlegroundMgr::Update(uint32 diff)
             if (GameTime::GetGameTime() > m_NextAutoDistributionTime)
```
jackpoz: should this one be changed too ?

Gooyeth:
1. Let's suppose that the distribution schedule is a thursday at 10 a.m.
2. m_NextAutoDistributionTime get the value of GameTime: https://github.com/Gooyeth/TrinityCore/blob/69fb35b50cf1774b7687f13b7203439dec8ba057/src/server/game/Battlegrounds/BattlegroundMgr.cpp#L639
3. The distribution time changes value (for example: thursday 10:05 am) and keeps increasing every week.
4. Changing to `sWorld->getWorldState(WS_ARENA_DISTRIBUTION_TIME)` get the value of the previous week, without modifying the timestamp.

Gooyeth:
@jackpoz, any change here?

PR #21640, opened 2018-03-16
[3.3.5] Core/Spells: Refactor LoS Disables for spells

DanVS:
**Changes proposed:**
-  Change comment for all spells LOS check disables, fix some typos.
-  Convert attribute `SPELL_ATTR2_CAN_TARGET_NOT_IN_LOS` from core to disables in the DB - they work the same.

- [x] 3.3.5
- [ ] master

**Tests performed:** Builds


sql/updates/world/3.3.5/9999_99_99_99_world.sql
```diff
@@ -0,0 +1,95 @@
...
+(0, 62042, 64, '', '', 'Disable LOS - Stormhammer'),
+(0, 62266, 64, '', '', 'Disable LOS - Trigger 3 Adds'),
+(0, 64767, 64, '', '', 'Disable LOS - Stormhammer'),
+(0, 66636, 64, '', '', 'Disable LOS - Rising Anger'),
+(0, 69922, 64, '', '', 'Disable LOS - Temper Quel Delar'),
+(0, 69956, 64, '', '', 'Disable LOS - Return Tempered Quel Delar'),
+(0, 70460, 64, '', '', 'Disable LOS - Coldflame Jets'),
+(0, 71024, 64, '', '', 'Disable LOS - Throw Bomb'),
+(0, 71599, 64, '', '', 'Disable LOS - Explosion (Chemical Wagon)'),
+(0, 75509, 64, '', '', 'Disable LOS - Twilight Mending');
```
Aokromes: missing new line

tkrokli: (Fixed now, but github can't hide comments after last line in the files)

ariel-:
> Convert attribute SPELL_ATTR2_CAN_TARGET_NOT_IN_LOS from core to disables in the DB - they work the same.

What is the intent behind this change? If an attribute exists for the spell it should be preferred rather than serverside data

tkrokli:
Maybe DanVS thinks there has been more volatility in the core spell system than in DB and that he thinks DB is more static in this case? :trollface:

DanVS:
I think that it is better to keep things in one place. And isn't adding attribute, when it isn't present in the DBC considered as hack?

ariel-:
No if it's shown through videos/sniffs that the spell is indeed ignoring LoS. It's a known fact that blizz uses different dbcs than the ones it distributed to the client. Only for those adding the attribute in SpellMgr it's acceptable to move them to disables table.

DanVS:
OK, I understood. Should I move SQL changes to issue and close PR?

ariel-:
No need to close the PR, because there'll be cpp changes (ie, removal of the attr in spellmgr)

DanVS:
But they are wrong as I underastand. Current logic is acceptable.

ariel-:
Ok, so the sql are spells which are currently missing the attribute from DB/DBC?

DanVS:
There are spells that were in the table before (readded because of comment, mostly) + spells moved from core.

DanVS:
> No if it's shown through videos/sniffs that the spell is indeed ignoring LoS. It's a known fact that blizz uses different dbcs than the ones it distributed to the client. Only for those adding the attribute in SpellMgr it's acceptable to move them to disables table.

Although, if both spells doesn't have attribute, but one is in the DB (2825) and another is in the core (57143). What is the preferred way to keep them? Isn't it better to keep them both in one place?

tkrokli:
> Only for those adding the attribute in SpellMgr it's acceptable to move them to disables table.

Looks like you can move those 2 to the `Disables` table.

Aokromes:
sooo, this can be merged?

DanVS:
I think so (as an author).

tkrokli:
Unless ariel- has got any more comments to add, it looks good to merge for me.

joschiwald:
The core part is good as it is. There are also good descriptions of the reason why the attribute is added. No need to move it to database.
If you just want to change the comment in database, use update queries.

PR #21646, opened 2018-03-16
Core/Player: Fix equipment set GUID handling

roc13x:
**Changes proposed:**

- Take transmog_outfits into consideration when setting high GUID at server load.
- Generate new GUID before storing equipment sets in memory. Fixes new sets overwriting eachother.

**Target branch(es):**

- [x] master

- Built
- Tested in-game

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -6728,7 +6728,7 @@ void ObjectMgr::SetHighestGuids()
     if (result)
         sArenaTeamMgr->SetNextArenaTeamId((*result)[0].GetUInt32()+1);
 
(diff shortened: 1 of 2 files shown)


src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -6663,7 +6663,11 @@ void ObjectMgr::SetHighestGuids()
 
     result = CharacterDatabase.Query("SELECT MAX(setguid) FROM character_equipmentsets");
     if (result)
-        _equipmentSetGuid = (*result)[0].GetUInt64()+1;
+        _equipmentSetGuid = (*result)[0].GetUInt64();
```
joschiwald: you need the `+1` if your next db query has an empty resultset

Aokromes: @roc13x

Issue #21516, opened 2018-02-27
Quest: The Caverns of Time


jerbookins:
**Description:**
http://www.wowhead.com/quest=10277/the-caverns-of-time

**Current behaviour:**
Nothing happens.

**Expected behaviour:**
There is supposed to be a Custodian that is spawned at the start of the quest and gives you a tour of the Caverns of Time (similar to the custodian that guides you around Shattrath).

He should walk you around the area and has this script:

>Greetings, &lt;name&gt;. I will guide you through the cavern. Please try and keep up.

>We do not know if the Caverns of Time have always been accessible to mortals. Truly, it is impossible to tell as the Timeless One is in perpetual motion, changing our timeways as he sees fit. What you see now may very well not exist tomorrow. You may wake up and have no memory of this place.

>It is strange, I know... Most mortals cannot actually comprehend what they see here, as often, what they see is not anchored within their own perception of reality.

>Follow me, please. &lt;The Custodian leads you to the entrance of the Old Hillsbrad timeway.&gt;

>There are only two truths to be found here: First, that time is chaotic, always in flux, and completely malleable and second, perception does not dictate reality.

>As custodians of time, we watch over and care for Nozdormu's realm. The master is away at the moment, which means that attempts are being made to dramatically alter time. The master never meddles in the affairs of mortals but instead corrects the alterations made to time by others. He is reactionary in this regard.

>For normal maintenance of time, the Keepers of Time are sufficient caretakers. We are able to deal with most ordinary disturbances. I speak of little things, such as rogue mages changing something in the past to elevate their status or wealth in the present.

>These tunnels that you see are called timeways. They are infinite in number. The ones that currently exist in your reality are what the master has deemed as 'trouble spots.' These trouble spots may differ completely in theme but they always share a cause. That is, their existence is a result of the same temporal disturbance. Remember that should you venture inside one...

>This timeway is in great disarray! We have agents inside right now attempting to restore order. What information I have indicates that Thrall's freedom is in jeopardy. A malevolent organization known as the Infinite Dragonflight is trying to prevent his escape. I fear without outside assistance, all will be lost.

>Follow me, please. &lt;The Custodian leads you to the entrance of the Black Morass timeway.&gt;

>We have very little information on this timeway. Sa'at has been dispatched and is currently inside. The data we have gathered from his correspondence is that the Infinite Dragonflight are once again attempting to alter time. Could it be that the opening of the Dark Portal is being targeted for sabotage? Let us hope not...

[...]

PR #21658, opened 2018-03-17
Core/Entities: Reduce probability of going under the map

### LINKED ISSUE #21625: Time Watchers out of place
**Description:**
www.wowhead.com/npc=19918/time-watcher

**Current behaviour:**
There are about a dozen dragons that all fly around at surface level (some come from underground) and land and just sit around in Tanaris.

https://streamable.com/xebe5

**Expected behaviour:**
These should be way under Tanaris in the Caverns of Time, pretty sure they fly down, turn into men, and then walk through a portal somewhere.

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  b978b673a905

**TDB version:** 335.64

Golrag:
**Changes proposed:**

-  Some tweaking in the check for following gridheight in SelectBestZForDestination
-  New function which adds the functionality I was kinda looking for (looking for vmap z without passing search dist). Which is just based of GetAreaInfo, kinda too tired of this to make a clean function for it.
-  Trying to not break tanaris dragons (which don't have vmap/gob above their head so they just have the gridheight by adding a check in updateallowedpositionz

- 3.3.5

**Issues addressed:** Closes #21625 #21516

- https://puu.sh/zJFCm/52feb76e76.mp4
- https://puu.sh/zJFH9/8a3bb51e51.mp4
- https://puu.sh/zJFTs/d7caa21ecb.mp4
- https://puu.sh/zJFY0/627e62fb80.mp4
- https://puu.sh/zJG8p/03061d99ca.mp4
- https://puu.sh/zJGkd/312d2d4f2c.mp4

- Pathing is sometimes weird (not sure if this is on the list to fix as it might be related to pathfinder)
- If you blink on the lich king platform WHILE it is rebuilding, you blink into the void

if this makes things worse, i'm thinking about re-adding NormalizeZForCollision.
Might take a while for me to respond, sorry in advance, am tired

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1498,9 +1498,11 @@ void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z) const
         }
         else
         {
-            float ground_z = GetMapHeight(x, y, z) + unit->GetHoverOffset();
-            if (z < ground_z)
-                z = ground_z;
+            float ground_z = GetMap()->GetGridMapHeight(x, y) + unit->GetHoverOffset();
+            float vmap_z = std::max(GetMap()->GetAreaInfoFloorZ(x, y, z + GetCollisionHeight()), GetMap()->GetGameObjectFloor(GetPhaseMask(), x, y, z, DEFAULT_HEIGHT_SEARCH, GetCollisionHeight())) + unit->GetHoverOffset();
+            if (vmap_z <= INVALID_HEIGHT)
+                if (z < ground_z)
+                    z = ground_z;
         }
     }
     else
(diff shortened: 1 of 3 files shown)


tkrokli:
Regarding the **"Core/Entities:**" part of the title (please forgive my nitpicking) - I see that the first of the 3 files (`Object.cpp`) in this PR is in the sub-directory `Entities`, but that part is mostly about vmaps, whereas the remaining 2 files are `Map.cpp` / `Map.h` in the `Maps` directory. Maybe I'm too slow in my gray cells to understand this logically, but wouldn't it be more precise to use **Core/Maps:** ? (I could be wrong, just trying to think logically here.)

Golrag:
most impactful change is in object.cpp but tbh i don't think it matters that much, they are both good choices

ccrs:
GetAreaInfoFloorZ -> vmap lookup on each UpdateAllowedPositionZ and GetAreaInfoFloorZ 😨

Gooyeth:
@Golrag: There is also a bug, where you give fear and the player goes flying, caused by the previous fix.

Golrag:
I will need some coordinates to look it u p, if it is because of pathgenerator making a shortcut I'm afraid I won't know where to look at

Gooyeth:
@Golrag:

.go xyz -8389.067 -4242.715 -207.3428 1
.aura 5782

Gooyeth:
Exactly, I thought it was not solved yet and I was telling you

Aokromes:
@jackpoz @Shauren  this can be merged?

PR #21682, opened 2018-03-22
Core/Spells: don't save Magma aura.

Gooyeth:
**Changes proposed:**

Avoid if a pet dies by magma, when invoked always die for magma.

**Target branch(es):** 3.3.5/master

- [x] 3.3.5
- [x] master

**Tests performed:** Build

PR #21683, opened 2018-03-23
Core/Spells: do not save Liquid auras

Gooyeth:
**Changes proposed:**

Avoid if a pet dies by magma, when invoked always die for magma.

- [x] 3.3.5
- [x] master


tkrokli:
Replaces PR #21682

tkrokli:
@ariel- Do you find this solution acceptable & practical for TC?

ariel-:
There must be some generic way. There are other liquid auras being saved besides this one. So the logic would be, any aura that depends on environment shouldn't be saved (ie area auras, liquid auras)

Gooyeth:
Ready, thanks.

Tested ingame 👍

src/server/game/Spells/Auras/SpellAuras.cpp
```diff
@@ -1161,6 +1160,11 @@ bool Aura::CanBeSaved() const
     if (GetCastItemGUID() && IsPermanent())
         return false;
 
+    // don't save liquid auras
+    for (LiquidTypeEntry const* liquid : sLiquidTypeStore)
```
Shauren: thats not a scalable approach

Gooyeth: So what can be placed instead?

joschiwald: maybe a custom spell attribute which is added on startup?

Gooyeth: Ready, thanks

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -2946,6 +2946,15 @@ void SpellMgr::LoadSpellInfoCustomAttributes()
         }
     }
 
+    // add attribute custom to liquid auras
```
tkrokli: Cosmetic/grammar: `// add custom attribute to liquid auras`

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -2946,6 +2946,12 @@ void SpellMgr::LoadSpellInfoCustomAttributes()
...
+    for (LiquidTypeEntry const* liquid : sLiquidTypeStore)
+        if (liquid->SpellId)
```
ariel-: We use enclosing braces `{ }` following a for/if whenever the next instruction is a block

Gooyeth: Ready, thanks

sql/updates/characters/3.3.5/9999_99_99_99_characters.sql
```diff
@@ -0,0 +1,2 @@
+--
+DELETE FROM `pet_aura` WHERE `spell` IN (28801, 37025, 42201, 57634);
```
ariel-: One last thing before merge: the sql (and the characters_database.sql changes) can be dropped.

The core will automatically avoid loading any aura (and pet aura) which doesn't pass Aura::CanBeSaved

Gooyeth: Ready

ariel-:
Welp, github isn't cooperating today.

Thanks for the contribution :)

PR #21754, opened 2018-04-01
Core/Battleground: implement previous opponent timer on Arena queue

### LINKED ISSUE #21599: [3.3.5] Arena: Queue same team time
**Description:**
seems in 335, you only can play against with the same opposite team when pass 2 minutes in queue.

Here I found somethink about this:
https://blue.mmo-champion.com/topic/3829-feeback-new-q---mmr-system-post-32/

>Kalgan
Blue Poster
Target Source#167 - 2009/08/13 12:35:32 AM
There is currently a 2-minute queue time addition to play against a team you just played against (so, 2 minutes plus normal queue time to be matched against them). This time value was 5 minutes when 3.2 was first released, but was lowered to 2 minutes in a hotfix last Friday. 

and here talk about this too:
https://us.battle.net/forums/en/wow/topic/3988669565

I think this system needs implement.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 31f14dadde46445ae7e3e0330a3811b185bd8a61

**TDB version:**  TDB335.64

**Operating system:** Debian

Gooyeth:
**Changes proposed:**

-  The arena teams only compete with the previous opponent after 2 minutes in queue

**Target branch(es):** 3.3.5
- [x] 3.3.5

**Issues addressed:** Closes #21599


tkrokli:
My bad, I picked the label [Feedback-FixOutdatedMissingWIP] based on the travis-ci error,
which in fact is caused by the current `3.3.5` source content.
Do you plan to make any changes to the current content of the PR, or is it finished from your point of view? :-)

Gooyeth:
I was thinking about a config option to change the timer.

tkrokli:
Very well. We will keep the labels as they are for now, then. :-)

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2736,6 +2736,14 @@ Arena.MaxRatingDifference = 150
 
 Arena.RatingDiscardTimer = 600000
 
+#
+#    Arena.PreviousOpponentsDiscardTimer
+#        Description: Time (in milliseconds) after which ignore the previous opponents
```
tkrokli: grammar suggestion:
```
#        Description: Time (in milliseconds) after which the previous opponents will be ignored.
```

src/server/game/Battlegrounds/BattlegroundQueue.h
```diff
@@ -51,6 +51,7 @@ struct GroupQueueInfo                                       // stores informatio
     uint32  IsInvitedToBGInstanceGUID;                      // was invited to certain BG
     uint32  ArenaTeamRating;                                // if rated match, inited to the rating of the team
     uint32  ArenaMatchmakerRating;                          // if rated match, inited to the rating of the team
+    uint32  OpponentsTeamId;                                // for compete again
```
tkrokli: comment field grammar suggestion:
```cpp
    uint32  OpponentsTeamId;                                // for instant rematch with the same team
```

tkrokli: (in case I have misunderstood the purpose of this function, please explain)

Gooyeth: Gets the value of the previous opponent to exclude it from the queue

tkrokli: OK, so that is to store the team ID for a new match, right?
Maybe `// rematch with the current team`
(`OpponentsTeamId` itself explains the rest)

Gooyeth: It is not the current one, it is from the previous opponent. It will be excluded from the queue until the timer is met and they can compete again.

tkrokli: Could be a useful renaming, makes more sense in this case.
`// team previously excluded from the queue`

Gooyeth: hmm... It is excluded for the current queue

sql/updates/characters/3.3.5/9999_99_99_99_characters.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `arena_team` ADD `previousOpponents` int(10) unsigned NOT NULL DEFAULT 0;
```
Krudor: Do we really need to store this in the database?
Personally, it seems like an acceptable loss of data in case of a server crash/restart.

src/server/game/Battlegrounds/BattlegroundQueue.cpp
```diff
@@ -924,6 +925,9 @@ void BattlegroundQueue::BattlegroundQueueUpdate(uint32 /*diff*/, BattlegroundTyp
         // this has to be signed value - when the server starts, this value would be negative and thus overflow
         int32 discardTime = GameTime::GetGameTimeMS() - sBattlegroundMgr->GetRatingDiscardTimer();
 
+        // timer for previous opponents
+        int32 discardOpponentsTime = GameTime::GetGameTimeMS() - sWorld->getIntConfig(CONFIG_ARENA_PREV_OPPONENTS_DISCARD_TIMER);
```
Krudor: Shouldn't this be calculated when the team is queue'd up, not each time the queue updates?

sql/updates/characters/3.3.5/9999_99_99_99_characters.sql
tkrokli: Well, does it work any better, worse or the same with this PR code added to base TC source?
(presuming that you have tested this with multiple players, of course)

Gooyeth: @tkrokli: It's certainly useless, I'll eliminate it.

src/server/game/Battlegrounds/BattlegroundQueue.cpp
```diff
@@ -957,6 +961,7 @@ void BattlegroundQueue::BattlegroundQueueUpdate(uint32 /*diff*/, BattlegroundTyp
                 if (!(*itr3)->IsInvitedToBGInstanceGUID
                     && (((*itr3)->ArenaMatchmakerRating >= arenaMinRating && (*itr3)->ArenaMatchmakerRating <= arenaMaxRating)
                         || (int32)(*itr3)->JoinTime < discardTime)
+                    && ((*itr_teams[0])->ArenaTeamId != (*itr3)->PreviousOpponentsTeamId || ((int32)(*itr3)->JoinTime < discardOpponentsTime))
```
Shauren: I realize this is a copypaste of rating discarding but this doesn't behave like your quote describes - what this does is prevent queueing into the same team for the first 2 minutes of being in queue, not a 2 minute addition to average wait time

Aokromes: @Gooyeth

Aokromes: @Gooyeth

jackpoz:
Would anyone like to take over this PR ?

Wyrserth:
I'll be taking this up after I finish some other stuff I have in progress :P

src/server/game/Battlegrounds/BattlegroundQueue.cpp
Wyrserth: > what this does is prevent queueing into the same team for the first 2 minutes of being in queue

This seems like the correct behavior, the blue post says 

> There is currently a 2-minute queue time addition to play against a team you just played against (so, 2 minutes plus normal queue time to be matched against them).

Shauren: No, example:
average wait time: 5 minutes
intended behavior: waiting for ~7 minutes before seeing the same team
current code: restricted only for 2 minutes from the moment of joining queue

Wyrserth: Isn't "average wait time" only a calculated value based on how much time the previous ten teams spent in queue?

I mean, if one team spends five minutes in queue before another team joins, then they'll contribute five minutes to the "average wait time".

Once two teams are in queue and allowed to match up, the arena match starts. If both join at the same time, their wait time is basically zero.

We forbid the same two teams from facing each other for the first two minutes in queue, but since they're both in queue, their wait time should be two minutes (arenas start when the system finds two available teams,

Unless I'm missing some big piece in the picture, in that case I apologize :P

PR #21774, opened 2018-04-04
Core/AI: added CreatureAI hook for successful spell casts

Ovahlord:
**Description:**

Implement a new AI hook to allow us to notify a scripted creature when a spell cast has been successfully finished. Needed for future boss implementations that need interrupt immunities applied when their spell cast has been finished (or interrupted which has been supported by my last AI hook PR). This will save us from a ton of OnEffectLaunch spell scripts in the future.

Example bosses:
- Temple Guardian Anhuur
- Rajh
- Ammunae
- Baron Ashbury
- Lord Godfrey
- and many, many more.

**Todo:**
- Maybe we should add a target pointer as well to save even more spell scripts (?)

### DIFF
--- a/src/server/game/AI/CreatureAI.h
+++ b/src/server/game/AI/CreatureAI.h
@@ -160,6 +160,9 @@ class TC_GAME_API CreatureAI : public UnitAI
         // Called when a spell cast gets interrupted
         virtual void OnSpellCastInterrupt(SpellInfo const* /*spell*/) { }
 
+        // Called when a spell cast has been successfully finished
+        virtual void OnSuccessfulSpellCast(SpellInfo const* /*spell*/) { }
+
         // Called at reaching home after evade
         virtual void JustReachedHome() { }
 
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -3417,6 +3417,11 @@ void Spell::cast(bool skipCheck)
         hitMask |= PROC_HIT_NORMAL;
 
     m_originalCaster->ProcSkillsAndAuras(nullptr, procAttacker, PROC_FLAG_NONE, PROC_SPELL_TYPE_MASK_ALL, PROC_SPELL_PHASE_CAST, hitMask, this, nullptr, nullptr);
+
+    // Call CreatureAI hook OnSuccessfulSpellCast
+    if (Creature* caster = m_originalCaster->ToCreature())
+        if (caster->IsAIEnabled)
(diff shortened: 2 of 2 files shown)


Krudor:
Sweet!

Slightly unrelated to this PR though, shouldn't OnSpellCastInterrupt deliver some interrupt info?
Some creatures do stuff like cast enrage spells when a specific spell is interrupted by a set of specific spells/spell effects.

Ovahlord:
I think if there are specific spells that have a fixed interrupt mechanic SpellHit serves the same purpose. Might be an idea though to create a creature helper "MakeInterruptable" and also use a variable to return the current interruptable state to make it work like you wanna have it.

Pesthuf:
What about making CastSpell return a pointer to the casted spell?
Then one could add JavaScript like callbacks to spell object, so that you could write something like

```c++
  Spell* spell = me->CastSpell(...);
  spell->OnCastResult = [=](SpellCastResult result) {
        if(result == SPELL_CAST_OK)
        {
            // say something, remove immunities, reschedule the spell with cooldown...
        }
        else if (result == SPELL_FAILED_INTERRUPTED)
        {
            // remove some immunity
        }
        else
        {
            // immediately re-schedule the spell event
        }
    }
```
Then your scripted effects are right beside the spell cast that triggered them and you won't have to jump around to a different method to understand what happens.

This is needed for NPCs with Melee abilities anyway. To give an example, let's imagine an NPC with a Melee ability running towards you. When the timer reaches 0, the event gets triggered and the NPC does CastSpell(SPELL_MELEE) on you. However, you are not in range, so the spell fails - silently. The next line in the script then usually re-sets the cooldown. When the NPC then reaches you, it won't use its ability, because it's on cooldown despite not having been successfully cast. As far as I know, that's not how it happens on offi. If you kite around an NPC there, it will always hit you with its melee spell as soon as it's in range.

Sure, we could currently solve this by wrapping every melee spell in a distance check and immediately reschedule it if it fails, but distance isn't the only thing that can make spells fail. Things like missing auras, disarm etc. can also cause spells to fail. Instead of poorly reinventing the cast check system, maybe we could use it? 

I think most ranged spells do the same. Try finding an NPC on offi with a targeted ranged ability with a cooldown and a cast time. If you force them to cancel their cast by running out of range or out of LOS and then go back in, they'll usually begin casting it again, with no cooldown. The cooldown only happens when they successfully cast it or get kicked. I think when I tested it, most didn't even get a colldown when they were stunned during a cast.

Shauren (approved):

PR #21776, opened 2018-04-05
Core/Player: character teleported by GM on first login skips intro cinematic

Gooyeth:
**Changes proposed:**

-  If the player is teleported before the first login, the cinematic is not shown

- [x] 3.3.5
- [x] master

**Tests performed:** Build


tkrokli:
Could you describe in more detail what this scenario is and why it would be useful to have this setting?

What about having this as a config option, if that is possible?

Gooyeth:
Example:

1. Create a character with the name "Trinity" (does not connect yet)
2. A GM uses the command `.tele name Trinity Orgrimmar` or the GM is in Orgrimmar and use the command `.summon Trinity`. Also in faction changes.
3. The character enters the game in the new position, and the cinematic is shown.
4. Since the cinematic does not match the area, then it is not shown.

The configuration option I think is not necessary, because it is only triggered by a GM or that it changes faction before logging on.

tkrokli:
OK, even though it sounds a bit like a custom action, it should be valid on default core. BTW, this kind of description / info is always nice to put in the first PR post, so other users don't have to ask for more info.

PR #21809, opened 2018-04-10
Core/Handler: Update HandleCharCreateOpcode to minimize name duplication

### LINKED ISSUE #21656: Create two characters with the same name
**Description:**
When creating a character, you can duplicate it. there will be 2 players with the same name.

**Current behaviour:**
When creating a character, repeatedly press enter and click "Accept" simultaneously, you will get disconnected, log back and see 2 players created with the same name. Try a few times, seem to be quite easy to reproduce.
Each will have it's own GUID, but same name.

wowscrnshot_031718_015424

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 282886984415

WildCoffee:
**Changes proposed:**

Add a character cache check instead of the async DB name check at the start, then check the database for a name conflict after all other checks have been made. By doing a sync query for the name closer to the insert it should further minimize the race condition.

**Target branch(es):** 3.3.5

- [ ] 3.3.5
- [ ] master

**Issues addressed:** Closes #21656 

- Built
- Tested by disabling AntiDOS and rapidly clicking accept and hitting enter to cancel. I could no longer add 2 characters with the same name.

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -380,6 +380,13 @@ void WorldSession::HandleCharCreateOpcode(WorldPacket& recvData)
         return;
     }
 
+    // check if already exists in cache first, check DB later
+    if (sCharacterCache->GetCharacterGuidByName(createInfo->Name) != ObjectGuid::Empty)
+    {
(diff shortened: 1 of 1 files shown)


illfated:
Looks like a good job. Tick the box(es) too?

Shauren:
Not bad but I preferred to keep the async

PR #21826, opened 2018-04-14
IsInAreaTriggerRadius overload based on areatrigger ID

Lopfest:
**Changes proposed:**
-  add overloaded method IsInAreaTriggerRadius with input areatrigger ID
-  in case there are multiple overlapping areatriggers, you can check with this method if you are within any of those areatriggers

- [x] master

**Issues addressed:** -

**Known issues and TODO list:** -

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -2182,6 +2182,13 @@ bool Player::IsInAreaTriggerRadius(const AreaTriggerEntry* trigger) const
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


Warpten:
You can check overlapping ATs with the entries just as well, so I don't get what this provided besides an helpful helper that should probably be inlined?

Lopfest:
i wasnt able to find out any logic that would help me do [this](https://gist.github.com/Lopfest/ef8b87c7de96d23789ecb22dd8576c05) without requested method

Shauren:
seriously? just pass `areaTrigger` there

Shauren:
just use sAreaTriggerStore there

PR #21837, opened 2018-04-16
Core/Bnet: Improve client ban messages

roc13x:
**Changes proposed:**

- Don't hide banned game accounts from list.
- Display proper ban messages with suspension time remaining.
- More appropriate error codes for bans and locks

**Target branch(es):**

- [x] master

**Tests performed:**
- Built, tested on client

### DIFF
--- a/src/server/bnetserver/Server/Session.cpp
+++ b/src/server/bnetserver/Server/Session.cpp
@@ -52,11 +52,12 @@ void Battlenet::Session::AccountInfo::LoadResult(PreparedQueryResult result)
 
 void Battlenet::Session::GameAccountInfo::LoadResult(Field* fields)
 {
-    // a.id, a.username, ab.unbandate > UNIX_TIMESTAMP() OR ab.unbandate = ab.bandate, ab.unbandate = ab.bandate, aa.gmlevel
+    // a.id, a.username, ab.unbandate, ab.unbandate = ab.bandate, aa.gmlevel
     Id = fields[0].GetUInt32();
     Name = fields[1].GetString();
-    IsBanned = fields[2].GetUInt64() != 0;
-    IsPermanenetlyBanned = fields[3].GetUInt64() != 0;
+    UnbanDate = fields[2].GetUInt32();
+    IsPermanenetlyBanned = fields[3].GetUInt32() != 0;
+    IsBanned = IsPermanenetlyBanned || UnbanDate > time(nullptr);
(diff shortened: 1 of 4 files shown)


src/server/game/Server/WorldSocket.cpp
```diff
@@ -754,7 +754,7 @@ void WorldSocket::HandleAuthSessionCallback(std::shared_ptr<WorldPackets::Auth::
     {
         if (account.BattleNet.LastIP != address)
         {
-            SendAuthResponseError(ERROR_DENIED);
+            SendAuthResponseError(ERROR_RISK_ACCOUNT_LOCKED);
```
Shauren: it does seem like a better choice but you need to triple check it ingame - wow client doesnt handle all of these codes from giant enum

src/server/bnetserver/Server/Session.cpp
```diff
@@ -52,11 +52,12 @@ void Battlenet::Session::AccountInfo::LoadResult(PreparedQueryResult result)
...
 void Battlenet::Session::GameAccountInfo::LoadResult(Field* fields)
 {
-    // a.id, a.username, ab.unbandate > UNIX_TIMESTAMP() OR ab.unbandate = ab.bandate, ab.unbandate = ab.bandate, aa.gmlevel
+    // a.id, a.username, ab.unbandate, ab.unbandate = ab.bandate, aa.gmlevel
     Id = fields[0].GetUInt32();
     Name = fields[1].GetString();
-    IsBanned = fields[2].GetUInt64() != 0;
-    IsPermanenetlyBanned = fields[3].GetUInt64() != 0;
+    UnbanDate = fields[2].GetUInt32();
+    IsPermanenetlyBanned = fields[3].GetBool();
```
Shauren: GetBool uses GetUInt8 internally - doesnt work with strict database type checking enabled

src/server/game/Server/WorldSocket.cpp
roc13x: Just checked now, it seems to handle that code okay: https://i.imgur.com/KSr8f9H.png

Shauren: iirc that one comes from bnet, you need to disable all ban/lock checks in bnetserver, then try again

roc13x: That's what I had done in the screenshot. Logged into bnet to pass the check. Banned the account while sitting on realmlist, then tried connecting.

Shauren (approved):

PR #21853, opened 2018-04-18
Core/Items: Add loot scaling support

roc13x:
**Changes proposed:**

- Implement ITEM_BONUS_SCALING_STAT_DISTRIBUTION_FIXED.
- Fix auctionhouse search with scaled items.
- Fix RequiredLevel enforcement for scaled items.
- Fix item enchants not scaling properly with bonuses.

**Target branch(es):**

- [x] master

- Built & Tested in-game


src/server/game/Entities/Player/Player.cpp
```diff
@@ -12335,6 +12340,9 @@ void Player::MoveItemToInventory(ItemPosCountVec const& dest, Item* pItem, bool
     // only set if not merged to existing stack (pItem can be deleted already but we can compare pointers any way)
     if (pLastItem == pItem)
     {
+        pItem->SetFixedLevel(getLevel());
```
Shauren: Pretty sure this should never be called again after initial item creation (makes next line unneeded as well?)

src/server/game/Entities/Item/Item.cpp
```diff
@@ -947,7 +956,15 @@ void Item::SetItemRandomProperties(ItemRandomEnchantmentId const& randomPropId)
 
 void Item::UpdateItemSuffixFactor()
 {
-    uint32 suffixFactor = GenerateEnchSuffixFactor(GetEntry());
+    if (!GetTemplate()->GetRandomSuffix())
```
Shauren: and maybe ITEM_FIELD_PROPERTY_SEED should be saved to db and this function removed?

Shauren (changes requested):

src/server/game/Entities/Item/Item.cpp
```diff
@@ -771,6 +759,22 @@ bool Item::LoadFromDB(ObjectGuid::LowType guid, ObjectGuid ownerGuid, Field* fie
     SetModifier(ITEM_MODIFIER_SCALING_STAT_DISTRIBUTION_FIXED_LEVEL, fields[43].GetUInt32());
     SetModifier(ITEM_MODIFIER_ARTIFACT_KNOWLEDGE_LEVEL, fields[44].GetUInt32());
 
+    if (Player* owner = GetConnectedOwner())
```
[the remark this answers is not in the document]
Shauren: also, at this point player isnt registered in ObjectAccessor yet. happens much later during login process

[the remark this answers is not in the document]
roc13x: I too would prefer an sql solution, but how would I go about that? The min/max scale level is constrained by DB2 data, I can't just update the level to the owners player level

src/server/game/Entities/Item/Item.cpp
roc13x: That would work for most cases yes. But some items (such as heirlooms) keep changing their item level, which keeps changing the seed value.

src/server/game/Entities/Player/Player.cpp
roc13x: I put that in as a fallback measure, in case the player somehow received an already-created item that has not been locked to a level yet. Perhaps unclaimed loot from mail or something?

src/server/game/Entities/Item/Item.cpp
Shauren: heirlooms dont have a random suffix, so no

src/server/game/Entities/Item/Item.cpp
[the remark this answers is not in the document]
Shauren: you can just create temporary tables with partial db2 data

src/server/game/Entities/Player/Player.cpp
Shauren: unclaimed loot isnt implemented yet but it should already have it set (remember that item modifiers are sent in loot packet - you can see that on creature corpse already)

src/server/game/Entities/Item/Item.cpp
roc13x: Fair enough about the heirlooms. Although I'd like to do the suffix factor changes in a followup PR, with some other item creation refactors/cleanups, if that's okay?

Shauren (approved):

PR #21888, opened 2018-04-26
[3.3.5] [NEEDS TESTING] Core/Movement: expand Movement Slot implementation

ccrs:
basic structure changes, nothing behavioural
future PRs will have the real juice


Shauren:
What exactly is your plan with this?

ccrs:
I still think having 3 slots is something actually desirable, see more pros than cons

but then again, we can swift to something like u once suggested, priority_queue, and forget about slots

ccrs:
I feel like having different slots makes it actually easier/more efficient in regards of accessing a group of generators, or even a single one.... a multimap can also do the trick I guess

src/server/game/Movement/MotionMaster.h
```diff
@@ -112,33 +122,32 @@ class TC_GAME_API MotionMaster
...
-        MovementGenerator* _slot[MAX_MOTION_SLOT];
-        bool _initialize[MAX_MOTION_SLOT];
-        MovementList _expireList;
         Unit* _owner;
-        int _top;
+        MovementSlot* _slots[MAX_MOTION_SLOT];
+        MovementList _expireList;
+        uint8 _top;
         uint8 _cleanFlag;
+        std::unordered_multimap<uint32, MovementGenerator*>  _baseUnitStatesMap;
```
ccrs: the idea is to have an exact control of the unitstates that the owner must have and must remove.

example:
 * moment 1
SLOT 2 -> PointMovement -> ROAMING State
SLOT 3 -> Empty
* moment 2
SLOT 2 -> PointMovement -> ROAMING State
SLOT 3 -> new Generator -> PointMovement -> ROAMING State
* moment 3
SLOT 2 -> want to erase -> PointMovement -> Finalize -> wants to remove ROAMING State
SLOT 3 -> still active PointMovement -> still wants ROAMING State

ccrs:
the main structure can be seen now... more or less, thoughts @Shauren @Treeston @ariel- ?

src/server/game/Movement/MotionMaster.h
```diff
@@ -38,21 +40,30 @@ namespace Movement
     class MoveSplineInit;
 }
 
+class MovementSlot
+{
+    public:
+        MovementSlot(uint8 type);
+        ~MovementSlot();
```
illfated: Should `MovementSlot` be renamed to `MovementSlots` here too (+ `class` , `uint8 type`) ?
build error: https://travis-ci.org/TrinityCore/TrinityCore/builds/371823289#L1385-L1391

ccrs:
Sounds good, will keep new things out of this PR though, I just want to set the structure for future changes

About UnitStates, I still think a detailed control is desirable, think about a jump spline while feared, you still are feared, cannot perform any action that is not allowed while feared.

Or the case I pointed before https://github.com/TrinityCore/TrinityCore/pull/21888#discussion_r184567495. Example, while a creature is moving towards a point, long distance travel, it can charge for example another creature, long distance charge too, and while charging maybe the AI tells MotionMaster to clear/delete the point movement, erasing ROAMING when it should still be applied because of the charge.

Shauren:
Yeah but you don't need to store it, just erase "all movement related states" when current movegen changes and compute (just take it off highest new generator) new state

src/server/game/Movement/MotionMaster.h
ccrs: dont worry about build errors, yet

ccrs:
maybe its just too obvious for me to notice, but sorry for being stubborn, cant see what you mean

on phases there are a couple reference counters to know what to do https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Phasing/PhaseShift.cpp#L148
and for auras usually a getter is used to know if there are any other auras using the same mechanic or w/e, like ```GetAuraEffectsByType``` or https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L11133

Shauren:
why are unit states for movement needed in the first place? arent they the current way of trying to solve movement type priority issue?

ccrs:
they signal the different movements that are active on the creature I guess, in the end the check is usually done against UNIT_STATE_MOVING

Shauren:
I bet most of these could be replaced with a real moving check (active spline)

ccrs:
most likely, the only use is in

```c++
bool IsStopped() const { return !(HasUnitState(UNIT_STATE_MOVING)); }
```

Treeston:
movement unit states are a bitmask, keep that in mind.

a lot of movegens use them to check whether another unit has a specific kind of movegen active (chase uses this to prevent infinite repositioning when two units try to be behind each other, for example...)

ccrs:
eitherway, the removal or not of those unitstates is out of the scope of this PR

Treeston:
Is there any reason to actually still have three slots?

If this is transformed into a `priority_queue` as @Shauren suggests (and I concur), I don't see any reason to keep the slots around at all.

ccrs:
@Treeston https://github.com/TrinityCore/TrinityCore/pull/21888#issuecomment-384887006

Shauren:
@Treeston making a lower priority movement type be on top for scripts?

Treeston:
`{ PRIORITY_CONTROLLED, PRIORITY_CHASE_MOVE }`?

PS: Though I disagree with using `_CONTROLLED` for this - fourth priority (`_SCRIPT_OVERRIDE`) or whatever should exist.

Shauren:
Thing is, I don't see a need to have 4 slots then

ccrs:
I see Treeston's point, actually I think its a good idea. Let me do the changes and we can talk about it with actual code

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -190,27 +191,51 @@ void MotionMaster::Update(uint32 diff)
...
 
 void MotionMaster::Clear(bool reset/*= true*/)
 {
-    if (HasCleanFlag(MOTIONMMASTER_CLEANFLAG_UPDATE))
-    {
-        reset ? AddCleanFlag(MOTIONMMASTER_CLEANFLAG_RESET_DEFAULT) : RemoveCleanFlag(MOTIONMMASTER_CLEANFLAG_RESET_DEFAULT);
-        AddCleanFlag(MOTIONMMASTER_CLEANFLAG_CLEAR_PENDING);
-    }
+    if (HasFlag(MOTIONMMASTER_FLAG_UPDATE))
+        _delayedActions.push_back(MotionMasterDelayedClear(this, reset));
```
Shauren: you are slicing the object here - will result in calling compiler generated "pure virtual" (basically, std::terminate)

ccrs: ouch, somehow I missed it >.<

src/server/game/Movement/MotionMaster.h
```diff
@@ -114,31 +136,67 @@ class TC_GAME_API MotionMaster
...
-        MovementGenerator* _slot[MAX_MOTION_SLOT];
-        bool _initialize[MAX_MOTION_SLOT];
-        MovementList _expireList;
         Unit* _owner;
-        int _top;
-        uint8 _cleanFlag;
+        MovementGeneratorPointer _defaultGenerator;
+        std::multiset<MovementGeneratorPointer, MovementGeneratorComparator> _generators;
+        std::unordered_multimap<uint32, MovementGenerator const*> _baseUnitStatesMap;
+        std::deque<std::shared_ptr<MotionMasterDelayedAction>> _delayedActions;
```
Shauren: shouldn't this be unique_ptr instead?

ccrs: well now that u say it... since its never _shared_ ... makes sense 💃

ccrs:
Being able to call Initialize and Clear on MovementInform... POGGERS
you can review MotionMaster again now btw 👍

Shauren:
Why do you have pointer-to-pointer for Top()?

ccrs:
now that you say it... the public getter can be used in all the cases in which now Top() is used... maybe leave it there for future cases?

for reference:

```c++
public:
    MovementGenerator const* GetCurrentMovementGenerator() const;
private:
    MovementGeneratorPointer const* Top() const;
````

**Comment by ccrs** [comment]
umm nvm, the public one is a const pointer to the generator while the other is not 👍

**Comment by ccrs** [comment]
On some places the public one can be used, most are just simple == operations.
But on a couple ones like

```
if (!top->get()->Update(_owner, diff))
...
movement->get()->UnitSpeedChanged();
...
Top()->get()->Initialize(_owner);
...
```

the methods invoked are not const and therefore not legal for a const pointer. Illegal example:

```
if (!GetCurrentMovementGenerator()->Update(_owner, diff))
```

**Comment by ccrs** [comment]
but then again, is it even necessary to protect the generator from outside changes?

**Comment by jackpoz** [comment]
please make sure this PR fixes https://github.com/TrinityCore/TrinityCore/issues/18471#issuecomment-386813404

**Comment by Treeston** [comment]
> @ccrs: SplineChainMovementGenerator, that was fun 

you're welcome

**Comment by illfated** [comment]
Nice to see that the number of build errors is growing smaller too, even if build errors are not your main concern at the moment. Lots of good work, **ccrs**! 👍

File: src/server/game/Movement/MovementGenerator.h [hunk 5]
```diff
@@ -28,22 +28,56 @@ class Unit;
...
         virtual void Initialize(Unit*) = 0;
-        virtual void Finalize(Unit*) = 0;
+        // on top reassign
         virtual void Reset(Unit*) = 0;
+        // on top on MotionMaster::Update
         virtual bool Update(Unit*, uint32 diff) = 0;
+        // on current top if another movement replaces
+        virtual void Deactivate(Unit*) = 0;
+        // on movement delete
+        virtual void Finalize(Unit*, bool, bool) = 0;
```
illfated: Looks like travis-ci is unhappy with your use of `Finalize` here,
compared to the existing line `ChaseMovementGenerator.h:39`
https://github.com/TrinityCore/TrinityCore/blob/7e0c0e6b79f82e164b5270974418f2829fa8e982/src/server/game/Movement/MovementGenerators/ChaseMovementGenerator.h#L39
is this an include problem or does it need a different name?

ccrs: nah, I'm not done, still need to check chase generator

ccrs:
Ok lurkers, it is time, time to test, testttttttttt for the love of god

I'll test too... but eventually the merge WILL come 💃

ccrs:
@Shauren why though?

nothing _should_ change, that was the big point of the PR, that nothing changes, and thats whats should be tested that everything remains the same and keeps working correctly

Shauren:
Because you gain NOTHING from doing that and only lose performance (it also makes the code harder to read)

ccrs:
@Shauren its the price of using unique_ptr on a container, the reference constructor is deleted (on contrast with shared_ptr, where its not) so iterating and erasing elements is imposible

morpheus4hire:
ccrs are we ok to test this or are you making suggested changes?

ccrs:
ah, sorry about that, probably tomorrow

morpheus4hire:
Np wanted to check and thanks for the great work all of you.

Treeston:
Do you actually use the uniqueness property of `std::set` anywhere?

Treeston:
what operations do you do on it? what's stored in it?

ccrs:
nooooow it can be tested, the first ones I made seem pretty promising

need some help boyzzz 👍

ccrs:
I've put it through serious testing, and cant seem to find any weird behaviour, very promising.
I'll take a look to some memory logs though.

Code review and extensive testing is very much appreciated, the merge is actually very likely right now.

ccrs:
Did a couple profiling memory and cpu sessions with VS, everything seems fine. Some interesting screenshots:
https://gyazo.com/53b97dab8de6f0fbdcfce1e435a37df0
https://gyazo.com/ecf73f22d90dc370beffb8be1b358fb0
https://gyazo.com/15655864ac84ec580974da770de4d97d
https://gyazo.com/49746517de9e5e4cc7b8d981b938127a
https://gyazo.com/36ea5d9aa61a47ec6734dc974b3318da

lets give this... lets say a week before merging 😄

ccrs:
couple examples of interesting points to test:
- Flight paths resume
- Flight paths between maps (such as the one from Plaguelands to Silvermoon)
- High priority movements overlaping other High priority movements (Fear a Knockedback target)
- Add/Erase calls in MovementInform (Sindragosa does this for example)
- High priority movements on PetAI (Fear, Knockback, etc)
- SAI Waypoint resume
- SAI Cast-Move flag
- SAI Escort
- SAI MovementInform
- SAI Path resume for non-escorting
- SAI Path MovementInform for non-escorting
- SAI Movement Action on Waypoint MovementInform

jackpoz:
did you use raw pointers in the end for the DirectDelete() issue ?

ccrs:
Nope, everything is working with unique_ptr ^^

jackpoz:
that requires boost 1.62 tho, so a few scenarios come up:
1) remove the code that requires boost 1.62
2) upgrade boost in travis (boost 1.62 on ubuntu 14.04 or ubuntu 18.04 loaded through docker)
3) remove travis (i.e. AppVeyor recently added linux VM support https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/ )

ccrs:
2 - do you know how to do the ubuntu 18 or 16 option?

ccrs:
feels like this should do it https://hub.docker.com/_/ubuntu/

ccrs:
image
 managed to apt-get update... yay....

Shauren:
Ok so... you:

- Used a container from 3rd party library
- Increased required library version
- replaced entire CI environment
- build said library from source

All of that just to be able to use unique_ptr in a container? Please, stop and use raw pointer.

ccrs:
Chill, Im just trying options

Shauren:
That in itself is fine but not after you made a "im done with functionality, can merge" claim

ccrs:
I wont merge till the problem with CI is finished dw, also its a problem with one of the two CI we're using hehe

ccrs:
And "third party container"? XD its boost not some squetchy random lib, many things use boost co tainers

ccrs:
All test done on my part (ingame and ASan). I consider it merge ready.

.travis.yml
```diff
@@ -10,18 +10,19 @@ git:
...
     packages:
       - g++-6
-      - libboost1.58-dev
-      - libboost-filesystem1.58-dev
-      - libboost-iostreams1.58-dev
-      - libboost-program-options1.58-dev
-      - libboost-regex1.58-dev
-      - libboost-system1.58-dev
-      - libboost-thread1.58-dev
+      - libboost1.62-dev
```
jackpoz: are these changes still required or can they be reverted ?

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -2230,16 +2232,16 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                     target->ToUnit()->RemoveAllGameObjects();
             break;
         }
-        case SMART_ACTION_STOP_MOTION:
+        case SMART_ACTION_REMOVE_MOVEMENT:
```
Shauren: Is the change to this action backwards compatible? If not then you need to update db data to keep behavior the same

src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
```diff
@@ -26,8 +26,12 @@
 #include "Random.h"
 
 template<class T>
-ConfusedMovementGenerator<T>::~ConfusedMovementGenerator()
+ConfusedMovementGenerator<T>::ConfusedMovementGenerator() : _timer(0), _x(0.f), _y(0.f), _z(0.f)
 {
+    MovementGenerator::Mode = MOTION_MODE_DEFAULT;
```
Shauren: pretty sure you are supposed to use `this->` in this situation

src/server/game/Movement/MovementGenerators/IdleMovementGenerator.cpp
```diff
@@ -70,74 +114,133 @@ bool RotateMovementGenerator::Update(Unit* owner, uint32 diff)
...
+{
+    AddFlag(MOVEMENTGENERATOR_FLAG_DEACTIVATED);
+}
+
+void DistractMovementGenerator::Finalize(Unit* /*owner*/, bool/* active*/, bool/* movementInform*/)
+{
+    AddFlag(MOVEMENTGENERATOR_FLAG_FINALIZED);
+
+    /*
+     *  TODO: ugh
```
Shauren: ugh

src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
```diff
@@ -139,21 +136,51 @@ template<>
...
+template<>
+void RandomMovementGenerator<Creature>::DoFinalize(Creature* owner, bool active, bool/* movementInform*/)
+{
+    AddFlag(MOVEMENTGENERATOR_FLAG_FINALIZED);
+    if (active)
+    {
+        owner->ClearUnitState(UNIT_STATE_ROAMING_MOVE);
+        owner->StopMoving();
+
+        // TODO: ???????????
```
Shauren: ?????????????

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -56,144 +59,465 @@ inline bool IsStatic(MovementGenerator* movement)
...
-        if (curr)
-            DirectDelete(curr);
+        case MOTION_SLOT_DEFAULT:
+            if (_defaultGenerator && filter(_defaultGenerator.get()))
+                movement = _defaultGenerator.get();
+            break;
+        case MOTION_SLOT_ACTIVE:
+            if (!_generators.empty())
+            {
+                MotionMasterContainer::const_iterator itr = std::find_if(_generators.begin(), _generators.end(), [filter](MovementGenerator const* a) -> bool
```
Shauren: auto (iterator), also capture filter by ref

Shauren: auto (its an iterator)
also you can rewrite it as simply `auto itr = std::find_if(_generators.begin(), _generators.end(), std::ref(filter));` // ref is here to avoid copying the function into find_if

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -56,144 +59,465 @@ inline bool IsStatic(MovementGenerator* movement)
...
+    switch (slot)
+    {
+        case MOTION_SLOT_DEFAULT:
+            if (_defaultGenerator && filter(_defaultGenerator.get()))
+                value = true;
+            break;
+        case MOTION_SLOT_ACTIVE:
+            if (!_generators.empty())
+            {
+                MotionMasterContainer::const_iterator itr = std::find_if(_generators.begin(), _generators.end(), [filter](MovementGenerator const* a) -> bool
```
Shauren: same here, auto itr = `std::find_if(_generators.begin(), _generators.end(), std::ref(filter));`

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -56,144 +59,465 @@ inline bool IsStatic(MovementGenerator* movement)
...
+    if (!movement || IsInvalidMovementSlot(slot))
         return;
 
-    if (_cleanFlag & MOTIONMMASTER_CLEANFLAG_UPDATE)
-        DelayedClean(slot);
-    else
-        DirectClean(slot);
+    if (HasFlag(MOTIONMASTER_FLAG_UPDATE))
+    {
+        std::function<void()> action = [this, movement, slot]() {
```
Shauren: codestyle, `{` placement (applies to all std::function objects you added for _delayedActions)

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -56,144 +59,465 @@ inline bool IsStatic(MovementGenerator* movement)
...
+    switch (slot)
+    {
+        case MOTION_SLOT_DEFAULT:
+            if (_defaultGenerator && _defaultGenerator.get() == movement)
+                DirectClearDefault();
+            break;
+        case MOTION_SLOT_ACTIVE:
+            if (!_generators.empty())
+            {
+                MotionMasterContainer::iterator itr = _generators.find(movement);
```
Shauren: auto with iterators

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -56,144 +59,465 @@ inline bool IsStatic(MovementGenerator* movement)
...
+    switch (slot)
+    {
+        case MOTION_SLOT_DEFAULT:
+            if (_defaultGenerator && _defaultGenerator->GetMovementGeneratorType() == type)
+                DirectClearDefault();
+            break;
+        case MOTION_SLOT_ACTIVE:
+            if (!_generators.empty())
+            {
+                MotionMasterContainer::iterator itr = std::find_if(_generators.begin(), _generators.end(), [type](MovementGenerator const* a) -> bool
```
Shauren: auto iterator

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -640,72 +978,113 @@ void MotionMaster::MovePath(uint32 pathId, bool repeatable)
...
+    std::for_each(_generators.begin(), _generators.end(), [filter, &cache](MovementGenerator* a)
+    {
+        if (filter(a))
+            cache.push_back(a);
+    });
+
+    if (cache.empty())
+        return;
+
+    for (MovementGenerator* movement : cache)
```
Shauren: you can do this without a temporary container, just use iterator directly instead of this loop syntax

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -727,131 +1106,120 @@ void MotionMaster::Mutate(MovementGenerator *m, MovementSlot slot)
...
+                    {
+                        _generators.erase(pointer);
+                        Delete(pointer, true, false);
+                    }
+                    else
+                        pointer->Deactivate(_owner);
+                }
+                else
+                {
+                    MotionMasterContainer::iterator itr = std::find_if(_generators.begin(), _generators.end(), [movement](MovementGenerator const* a) -> bool
```
Shauren: auto

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -727,131 +1106,120 @@ void MotionMaster::Mutate(MovementGenerator *m, MovementSlot slot)
...
-        MovementGenerator* curr = top();
-        pop();
-        DelayedDelete(curr);
-    }
+    if (!movement || !movement->BaseUnitState)
+        return;
 
-    while (!empty() && !top())
-        --_top;
+    _baseUnitStatesMap.insert(std::make_pair(movement->BaseUnitState, movement));
```
Shauren: emplace should work here without make_pair step

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -727,131 +1106,120 @@ void MotionMaster::Mutate(MovementGenerator *m, MovementSlot slot)
...
+void MotionMaster::ClearBaseUnitState(MovementGenerator const* movement)
 {
-    if (IsStatic(curr))
+    if (!movement || !movement->BaseUnitState)
         return;
 
-    curr->Finalize(_owner);
-    delete curr;
-}
+    std::pair<MotionMasterUnitStatesContainer::iterator, MotionMasterUnitStatesContainer::iterator> bounds = _baseUnitStatesMap.equal_range(movement->BaseUnitState);
```
Shauren: Trinity::Containers::MultimapErasePair is designed exactly for what you are doing here, after erasing follow with
```
if (_baseUnitStatesMap.count(movement->BaseUnitState)
    _owner->ClearUnitState(movement->BaseUnitState);
```

Shauren (changes requested):

.travis.yml
ccrs: up to you/all to decide, not required indeed

src/server/game/AI/SmartScripts/SmartScript.cpp
ccrs: there is not, currently, any script using the action

> backwards compatible?
I guess so... the old one make 0 sense and hardly ever worked

src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
ccrs: that sounds right, it did never make too much sense to mee either, like it works, but thats how u declare a local variable? or call a static one? dunnow, will change 👍

src/server/game/Movement/MovementGenerators/IdleMovementGenerator.cpp
ccrs: aka will change with time and future work, something more descriptive would be better though xd

src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
ccrs: same, but dont tell me its not worth the meme

src/server/game/Movement/MotionMaster.cpp
ccrs: the function is referenced, there was no copy there, right?
eitherway, much cleaner your way 👍

src/server/game/Movement/MotionMaster.cpp
ccrs: 👍

src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
Shauren: Because this (or base) class is a template its "dependent name lookup", by qualifying it with `this->` you change when compiler resolves it (template two phase stuff)

src/server/game/Movement/MotionMaster.cpp
ccrs: tru tru

src/server/game/Movement/MotionMaster.cpp
ccrs: I guess, since the container size wont ever be > 20 (and thats maybe too much to say), I thought this was """"""safer"""""" and the setback was minimal. Will change though 👍

src/server/game/Movement/MotionMaster.cpp
ccrs: did I try that? I cant remember it was like 3-4 weeks ago, will try again 👍 
plus the current one is a copy insert sooo 💃

src/server/game/Movement/MotionMaster.cpp
ccrs: ow, thats a nice one 👍

src/server/game/Movement/MotionMaster.cpp
Shauren: There is a copy, default lambda capture mode is by value, it does not matter what the actual captured type is (you would need capture by ref, [&filter])

src/server/game/Movement/MotionMaster.cpp
Shauren: Should work (would be a different story if you tried to construct pair elements using their own constructors passing even more args, std::piecewise_construct would be needed)

src/server/game/Movement/MotionMaster.cpp
ccrs: reference the reference, the more you know 💃

src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
ccrs: got it 👍

.travis.yml
jackpoz: maybe save them somewhere in case we have to upgrade boost but remove these changes from the PR, ty

ccrs: 👍

ccrs: https://gist.github.com/ccrs/53d049c0dd8b392b3aa3f3c5de91a04b

Shauren (approved):

ccrs:
doing some last test after the last commit before merging 👍

Gooyeth:
@ccrs: This is causing problems with Harvest Soul of The Lich King, the characters die for fall damage.

1. `.go xyz 529.302 -2124.49 840.857 631`
2. `.npc add temp 36598`

Keader:
Open a ticket @Gooyeth

TheWinchesters:
Is there an open ticket about whay Gooyeth says or do I open one?

jackpoz:
Open one please and references this PR, easy way to track things

PR #21943, opened 2018-05-17
Core/Player: Proper handling mount and shapeshift on Flight Path start and end.

LluisMas:
**Changes proposed:**

- Making the player dismount automatically before taking a flying path.

- [x] master

Tested in-game

** Information of the problem: ** As you can see on the video below, it should dismount automatically when picking a flying route. Currently you get an annoying "Already mounted" message and you have to dismount automatically.

Video from retail: https://youtu.be/CjKve6enXX8

**TODO**
- [x] Reset druid shapeshift displayid and apply it again when the route is finished.

PS: It's my first pull request here I hope I did okay.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -22070,13 +22070,12 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc 
     if (npc)
     {
         // not let cheating with start flight mounted
-        if (IsMounted())
-        {
-            GetSession()->SendActivateTaxiReply(ERR_TAXIPLAYERALREADYMOUNTED);
-            return false;
-        }
+        RemoveAurasByType(SPELL_AURA_MOUNTED);
 
-        if (IsInDisallowedMountForm())
+        if (GetDisplayId() != GetNativeDisplayId())
+            RestoreDisplayId(true);
+
+        if (IsDisallowedMountForm(getTransForm(), FORM_NONE, GetDisplayId()))
         {
             GetSession()->SendActivateTaxiReply(ERR_TAXIPLAYERSHAPESHIFTED);
             return false;
@@ -22094,8 +22093,8 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc 
     {
         RemoveAurasByType(SPELL_AURA_MOUNTED);
 
-        if (IsInDisallowedMountForm())
-            RemoveAurasByType(SPELL_AURA_MOD_SHAPESHIFT);
+        if (GetDisplayId() != GetNativeDisplayId())
+            RestoreDisplayId(true);
 
         if (Spell* spell = GetCurrentSpell(CURRENT_GENERIC_SPELL))
             if (spell->m_spellInfo->Id != spellid)
--- a/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
@@ -404,6 +404,7 @@ void FlightPathMovementGenerator::DoFinalize(Player* player)
     }
 
     player->RemoveFlag(PLAYER_FLAGS, PLAYER_FLAGS_TAXI_BENCHMARK);
+    player->RestoreDisplayId();
 }
 
 #define PLAYER_FLIGHT_SPEED 30.0f
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9112,11 +9112,16 @@ bool Unit::IsInFeralForm() const
 
 bool Unit::IsInDisallowedMountForm() const
 {
-    if (SpellInfo const* transformSpellInfo = sSpellMgr->GetSpellInfo(getTransForm()))
+    return IsDisallowedMountForm(getTransForm(), GetShapeshiftForm(), GetDisplayId());
+}
+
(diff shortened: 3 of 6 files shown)


illfated:
Recommended PR title format: **Core/Player: automatically dismount player on Flight Path start**
Github also recommends shortening titles compared to your current title.
(TC members will confirm if your suggested solution is blizzlike and acceptable)

mdX7:
iirc you only get visually dismounted and remounted when path ends. For druid forms I'm 100% sure that your displayid 'resets' to normal char before flight starts and the displayid of the current form is reapplied when flight ends. The shapeshift form aura is still active during the whole flight though

LluisMas:
Forgot to attach a video from retail. You can see in the video how the player is dismounted and it’s not mounted again at the end of the route. Adding druid shapeshift removal to TODO list

mdX7:
"Remove druid shapeshift aura." is not correct. Only the displayid changes/resets, aura is still active. Didn't check behaviour in sniffs tho

mdX7:
Keep in mind that some morph auras are > shapeshift. E.g. suramar disguises

src/server/game/Entities/Player/Player.cpp
```diff
@@ -22069,17 +22069,17 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc
...
     if (npc)
     {
+
         // not let cheating with start flight mounted
         if (IsMounted())
         {
-            GetSession()->SendActivateTaxiReply(ERR_TAXIPLAYERALREADYMOUNTED);
-            return false;
+            Dismount();
+            RemoveAurasByType(SPELL_AURA_MOUNTED);
```
Shauren: Why do you use both Dismount and RemoveAurasByType?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -22069,17 +22069,17 @@ bool Player::ActivateTaxiPathTo(std::vector<uint32> const& nodes, Creature* npc
...
         }
 
-        if (IsInDisallowedMountForm())
+        if (GetDisplayId() != GetNativeDisplayId())
```
Shauren: This isn't correct, transforms that allow mounting will not be removed
What about transform auras expiring mid-flight when there is more than one transform aura remaining?

src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
```diff
@@ -404,6 +404,16 @@ void FlightPathMovementGenerator::DoFinalize(Player* player)
     }
 
     player->RemoveFlag(PLAYER_FLAGS, PLAYER_FLAGS_TAXI_BENCHMARK);
+
+    if (ShapeshiftForm form = player->GetShapeshiftForm())
+    {
+        SpellShapeshiftFormEntry const* shapeshift = sSpellShapeshiftFormStore.LookupEntry(form);
+        if (!shapeshift)
```
Shauren: Why is this check needed?

src/server/game/Entities/Player/Player.cpp
LluisMas: Yes, maybe only  RemoveAurasByType(SPELL_AURA_MOUNTED); will work.

src/server/game/Entities/Player/Player.cpp
LluisMas: Didn't think about that, what will be the correct form then?

src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
LluisMas: To check if the player is shapeshifted to return his displayid

Shauren: But player->GetShapeshiftForm() already gives you all thats needed

LluisMas: Fixed

src/server/game/Entities/Player/Player.cpp
LluisMas: Fixed

src/server/game/Entities/Player/Player.cpp
LluisMas: I don't really know how to do that because I can't just remove the aura since it needs to be reapplied when the path finishes.

src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
```diff
@@ -404,6 +404,9 @@ void FlightPathMovementGenerator::DoFinalize(Player* player)
     }
 
     player->RemoveFlag(PLAYER_FLAGS, PLAYER_FLAGS_TAXI_BENCHMARK);
+
+    if (ShapeshiftForm form = player->GetShapeshiftForm())
```
Shauren: Replace this with `player->RestoreDisplayId()`, will take care of transform and shapeshifts

src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
```diff
@@ -405,8 +405,10 @@ void FlightPathMovementGenerator::DoFinalize(Player* player)
 
     player->RemoveFlag(PLAYER_FLAGS, PLAYER_FLAGS_TAXI_BENCHMARK);
 
-    if (ShapeshiftForm form = player->GetShapeshiftForm())
+   /* if (ShapeshiftForm form = player->GetShapeshiftForm())
```
Aokromes: don't comment the code if you don't go to use it, delete it.

LluisMas:
Tested it on local and works properly

Issue #10105, opened 2013-06-26
Players can not be queued for Battlegrounds and Arenas at the same time


Bezo:
how it work:
now you can que only for bg or arena. 

how it should work:
"Players can now be queued for battlegrounds and arenas at the same time."
as from patch 3.3.3
link to source : http://www.wowwiki.com/Patch_3.3.3_%28undocumented_changes%29
under UI section.
##

Gooyeth:
It is currently implemented

tkrokli:
Ref. #3752, #21487, #21491

Killyana:
Not fixed, you still get a message that tels you cannot queue fo BG or arena.

Gooyeth:
Yes it is working, however, there is a limitation on random. The note talks about "Battlegrounds", but does not mention the random mode.

Killyana:
Random or not the player must be able to tag in both queue.

Jildor:
If you are in BG queue, you can't join to arena

Killyana:
One issue remains, but I'm not sure if it's an issue or how it must work:
If you queue into a battleground as party (party leader), then you leave the queue, only your char will leave the queue, other players will remains tagged.

Jildor:
Another issue is: (I'm not sure if it's an issue or how it must work)

When you join to bg queue and arena queue, when you get called to enter (bg or arena), you should be removed from another queue?

by example:
I'm in arena queue and bg queue, call to enter arena, I should be removed from bg queue??

Gooyeth:
@Jildor: I'm not sure, but vice versa does not happen, you can also be on a battleground and be called to arenas.

PR #21963, opened 2018-05-25
Core/Battleground: fixed the Arena queue if it was in Random battleground queue

Gooyeth:
Fixed a problem that did not let Arenas queue if it was in queue of Random Battleground

**Target branch(es):** 3.3.5

**Issues addressed:** Closes https://github.com/TrinityCore/TrinityCore/issues/10105#issuecomment-388317123

**Tests performed:** Build


jackpoz:
let's merge and see what happens.

PR #21999, opened 2018-06-04
Core/Spells: Allow Frenzied Regeneration to crit

### LINKED ISSUE #20798: [3.3.5a] Core/Spells: Frenzied Regeneration should crit
**Current behaviour:** [Frenzied Regeneration](http://www.wowhead.com/spell=22842) doesn't crit at all.

**Expected behaviour:** This spell must have a crit chance ([see comment](http://www.wowhead.com/spell=22842/frenzied-regeneration#comments:id=388268)). It's unstated whether it should inherit spell or agility crit, but according to the worldoflogs, I'd assume that it's spell crit.

**Branch(es):** 3.3.5a

**TC rev. hash/commit:**  TrinityCore rev. 6612ec47f4cd+

**TDB version:**  TDB335

**Operating system:** Ubuntu 16.04

DanVS:
**Changes proposed:**
Frenzied Regeneration spell (22845) has `SPELL_SCHOOL_MASK_NORMAL` and `SPELL_DAMAGE_CLASS_NONE` that makes it not having chance to crit.

PR enforces school mask and damage class like it is done with some similar healing spells and it makes Frenzied Regeneration to work offlike.

**Target branch(es):**
- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #20798

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -3072,6 +3072,7 @@ void SpellMgr::LoadSpellInfoCorrections()
     ApplySpellFix({
         379,   // Earth Shield
         33778, // Lifebloom Final Bloom
+        22845, // Frenzied Regeneration
 
         52042, // Healing Stream Totem
         // this one is here because we have no SP bonus for dmgclass none spell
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/SpellMgr.cpp
```diff
@@ -3088,6 +3089,12 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->DmgClass = SPELL_DAMAGE_CLASS_MAGIC;
     });
 
+    // Frenzied Regeneration
+    ApplySpellFix({ 22845 }, [](SpellInfo* spellInfo)
+    {
+        spellInfo->SchoolMask = SPELL_SCHOOL_MASK_NATURE;
```
Aokromes: [22:15:59] <Shauren> afaik school is sent in combat log packets so that might not be valid

DanVS: `SPELL_SCHOOL_MASK_NORMAL` makes critical chance to be zero according to [`Unit::SpellCritChanceDone`](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L6940-L6941). Adding an `if` here could be a solution?

PR #22000, opened 2018-06-04
Core/Spells: Reset Mangle cooldown in Berserk

DanVS:
**Changes proposed:**
As of [patch 3.0.3](https://wow.gamepedia.com/Patch_3.0.3) Berserk must reset Mangle (bear) cooldown on use.

> [Berserk]: Now clears the cooldown on [Mangle] (Bear).

PR creates spell script that handles that behaviour.

**Target branch(es):**
- [x] 3.3.5
- [ ] master

P.S. 22000 PR!

### DIFF
--- a/sql/updates/world/3.3.5/9999_99_99_99_world.sql
+++ b/sql/updates/world/3.3.5/9999_99_99_99_world.sql
@@ -1,3 +1,3 @@
-DELETE FROM `spell_script_names` WHERE `spell_id` = -50334;
+DELETE FROM `spell_script_names` WHERE `ScriptName` = 'spell_dru_berserk';
 INSERT INTO `spell_script_names`(`spell_id`, `ScriptName`) VALUES
 (-50334,'spell_dru_berserk');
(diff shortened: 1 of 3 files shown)


sql/updates/world/3.3.5/9999_99_99_99_world.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM `spell_script_names` WHERE `spell_id` = -50334;
```
Keader: always delete by ScriptName

src/server/scripts/Spells/spell_druid.cpp
```diff
@@ -177,6 +182,38 @@ class spell_dru_bear_form_passive : public SpellScriptLoader
...
 };
 
+// -50334 - Berserk
+class spell_dru_berserk : public SpellScriptLoader
+{
+    public:
+        spell_dru_berserk() : SpellScriptLoader("spell_dru_berserk") { }
+
+            class spell_dru_berserk_AuraScript : public AuraScript
+        {
```
Keader: Identation...
Will be nice if you use the new Script Registers to do it like:
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Spells/spell_mage.cpp#L661-L686

Keader:
Thanks

PR #22012, opened 2018-06-08
Core/Spells: fixed a edge case crash which happens if a player  is casting from via PassengerBoarded hook straight after exiting vehicle.

Ovahlord:
**Changes proposed:**
-  this little fix is solving a crash that happens if PassengerBoarded is being used with apply being marked as false to cast a spell which requires the passenger to cast a spell. For some reason the hook is being called before the vehicle is actually set to nullptr so the core is trying to get vehicle data while the player doesn't have a vehicle anymore. While this PR covers the crash only the real issue is the cause of the crash.

- [x] master


illfated:
Possible way to shorten the title, if you want to show more of the key points:
**Core/Spells: fix edge case crash (player cast spell via PassengerBoarded hook, apply=false)**

Ovahlord:
I prefer proper explaining titles above cased ones. And hi Trokli

Palabola:
Confirmed, it work.

PR #22013, opened 2018-06-08
Core/Spells: Skip shapeshift removal while in Shadow Dance

DanVS:
**Changes proposed:**
[Shadow Dance](https://dbwotlk.com/?spell=51713) (51713) is a shapeshift effect. It means that applying [Stealth](https://dbwotlk.com/?spell=1784) or [Vanish](https://dbwotlk.com/?spell=26889) cancels Shadow Dance prematurely.

PR allows to skip shapeshifts removal while rogue is in Shadow Dance state.

**Target branch(es):**
- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -1713,7 +1713,9 @@ void AuraEffect::HandleAuraModShapeshift(AuraApplication const* aurApp, uint8 mo
         }
 
         // remove other shapeshift before applying a new one
-        target->RemoveAurasByType(SPELL_AURA_MOD_SHAPESHIFT, ObjectGuid::Empty, GetBase());
+        // skip on rogue shadow dance
+        if (!target->HasAura(51713))
+            target->RemoveAurasByType(SPELL_AURA_MOD_SHAPESHIFT, ObjectGuid::Empty, GetBase());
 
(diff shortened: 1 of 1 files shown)


kvipka:
better just : 
if (target->getClass() != CLASS_ROGUE)

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -1713,7 +1713,9 @@ void AuraEffect::HandleAuraModShapeshift(AuraApplication const* aurApp, uint8 mo
         }
 
         // remove other shapeshift before applying a new one
-        target->RemoveAurasByType(SPELL_AURA_MOD_SHAPESHIFT, ObjectGuid::Empty, GetBase());
+        // skip on rogue shadow dance
+        if (!target->HasAura(51713))
```
kvipka: if (target->getClass() != CLASS_ROGUE)

ccrs:
hack, use a spellscript

DanVS:
ccrs, which method should I use in such case?

ccrs:
its complicated, not an expert in spell system, but, since what you want is to prevent a default behaviour in one handler, logic tells to use a spellscript to override it and define what you want to happen. though, again, I dont know without actually digging into it if its even something appropiate for this issue.

DanVS:
Will rework.

Issue #19770, opened 2017-05-17
[3.3.5] Core/Pet: Action bar


Cannix:
**Description:**

If you summon a dead hunter pet you wont get the pet action bar.

**Current behaviour:**

No action bar.

**Expected behaviour:**

You should get an action bar.

**Steps to reproduce the problem:**

1. Kill your pet
2. Run out of range / wait for despawn
3. Summon it

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. edd2c1c03995 2017-05-17 08:08:46 +0200 (3.3.5 branch) (Win64, Debug, Static) (worldserver-daemon) 

**TDB version:**  CHANGEME Version of the TrinityCore database

**Operating system:** 
Win

kaldown:
![2018-02-14--01 05 31_1366x768_scrot](https://user-images.githubusercontent.com/3754017/36176355-62e612ae-1123-11e8-887a-8c002f97a2a2.png)

tkrokli:
@Langerz82 : Just a minor question, does the pet bar appear even when the pet is dead, or does it appear after the pet is revived? (I know, I should have tested it, I just haven't been focused enough to test it for myself yet.)

Langerz82:
For me it disappears when your out of range and the pet is dead, but casting revive the bar appeared again. Maybe I am missing something in the test case.

tkrokli:
OK, fair enough. I will test this later today and see if it is possible to reproduce the issue or not.

tkrokli:
TrinityCore rev. 93357631fa8f 2018-02-18 19:52:58 +0000 (3.3.5 branch) (Win64, Release, Static)
TDB 335.64 + updates up to and including `2018_02_19_18_world_335.sql`

Took me a bit longer than I thought, but I just managed to reproduce the issue under some dubious circumstances.

I entered the Alterac Valley battleground with `.debug bg` so I could check it out on my own (playing locally on my computer). While fighting Vanndar Stormpike `<Stormpike General>`, my hunter pet got killed during one of the attempts and I chose not to resurrect it until I completed the BG and used the final "leave battleground" function from the end credit UI. After reviving my pet, I did not see the pet bar while moving around in the AV entrance cave.

![wowscrnshot_022018_234840](https://user-images.githubusercontent.com/4903600/36454250-658c0400-169b-11e8-9804-66f5117cf22e.jpg)

[...]

PR #22014, opened 2018-06-09
Core/Entities: Handle partial state change for pets

DanVS:
**Changes proposed:**
Some spells like [Feral Spirit](https://dbwotlk.com/?spell=51533) (51533) summon a set of pets under player control. But after any pet of the set dies, player looses control on remaining.

PR enables `PetHandler` to iterate through all of the player controlled pets instead of getting just one. Therefore, partial set of pets changes are processed. List is made by `Unit::ControlList` iterating and saved in `std::vector` container to be used in pet `PetSetAction` and `PetSpellAutocastOpcode` handlers.

- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/server/game/Entities/Creature/TemporarySummon.cpp
+++ b/src/server/game/Entities/Creature/TemporarySummon.cpp
@@ -300,6 +300,21 @@ void Minion::RemoveFromWorld()
     TempSummon::RemoveFromWorld();
 }
 
+void Minion::setDeathState(DeathState s)
+{
+    Creature::setDeathState(s);
+    if (s == JUST_DIED && IsGuardianPet())
+        if (Unit* owner = GetOwner())
+            if (owner->GetTypeId() == TYPEID_PLAYER && owner->GetMinionGUID() == GetGUID())
+                for (Unit* controlled : owner->m_Controlled)
+                    if (controlled->GetEntry() == GetEntry() && controlled->IsAlive())
+                    {
+                        owner->SetMinionGUID(controlled->GetGUID());
+                        owner->SetPetGUID(controlled->GetGUID());
+                        owner->ToPlayer()->CharmSpellInitialize();
+                    }
+}
+
 bool Minion::IsGuardianPet() const
 {
     return IsPet() || (m_Properties && m_Properties->Category == SUMMON_CATEGORY_PET);
--- a/src/server/game/Entities/Creature/TemporarySummon.cpp
+++ b/src/server/game/Entities/Creature/TemporarySummon.cpp
@@ -303,16 +303,22 @@ void Minion::RemoveFromWorld()
 void Minion::setDeathState(DeathState s)
 {
     Creature::setDeathState(s);
-    if (s == JUST_DIED && IsGuardianPet())
-        if (Unit* owner = GetOwner())
-            if (owner->GetTypeId() == TYPEID_PLAYER && owner->GetMinionGUID() == GetGUID())
-                for (Unit* controlled : owner->m_Controlled)
-                    if (controlled->GetEntry() == GetEntry() && controlled->IsAlive())
-                    {
-                        owner->SetMinionGUID(controlled->GetGUID());
-                        owner->SetPetGUID(controlled->GetGUID());
-                        owner->ToPlayer()->CharmSpellInitialize();
-                    }
+    if (s != JUST_DIED || !IsGuardianPet())
+        return;
+
+    Unit* owner = GetOwner();
+    if (!owner || owner->GetTypeId() != TYPEID_PLAYER || owner->GetMinionGUID() != GetGUID())
+        return;
+                
+    for (Unit* controlled : owner->m_Controlled)
+    {
+        if (controlled->GetEntry() == GetEntry() && controlled->IsAlive())
+        {
+            owner->SetMinionGUID(controlled->GetGUID());
+            owner->SetPetGUID(controlled->GetGUID());
+            owner->ToPlayer()->CharmSpellInitialize();
+        }
+    }
 }
 
 bool Minion::IsGuardianPet() const
--- a/src/server/game/Entities/Creature/TemporarySummon.cpp
+++ b/src/server/game/Entities/Creature/TemporarySummon.cpp
@@ -317,6 +317,7 @@ void Minion::setDeathState(DeathState s)
             owner->SetMinionGUID(controlled->GetGUID());
             owner->SetPetGUID(controlled->GetGUID());
             owner->ToPlayer()->CharmSpellInitialize();
+            break;
         }
     }
 }
--- a/src/server/game/Handlers/PetHandler.cpp
+++ b/src/server/game/Handlers/PetHandler.cpp
@@ -526,14 +526,12 @@ void WorldSession::HandlePetSetAction(WorldPacket& recvData)
     }
 
     std::vector<Unit*> pets;
-    for (Unit::ControlList::iterator itr = GetPlayer()->m_Controlled.begin(); itr != GetPlayer()->m_Controlled.end(); ++itr)
-        if ((*itr)->GetEntry() == pet->GetEntry() && (*itr)->IsAlive())
-            pets.push_back(*itr);
+    for (Unit* controlled : _player->m_Controlled)
+        if (controlled->GetEntry() == pet->GetEntry() && controlled->IsAlive())
+            pets.push_back(controlled);
 
-    for (std::vector<Unit*>::iterator itr = pets.begin(); itr != pets.end(); ++itr)
+    for (Unit* pet : pets)
     {
-        Unit* pet = *itr;
-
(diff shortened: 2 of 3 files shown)


illfated:
Rare to see a PR starting off with everything correct test-wise in its first commit. Good job, I say. 👍

Shauren:
I don't know how is this supposed to behave tbh - in legion feral spirit does not get a pet bar (no multi-summon does)

src/server/game/Handlers/PetHandler.cpp
```diff
@@ -525,66 +525,76 @@ void WorldSession::HandlePetSetAction(WorldPacket& recvData)
         }
     }
 
-    // check swap (at command->spell swap client remove spell first in another packet, so check only command move correctness)
-    if (move_command)
+    std::vector<Unit*> pets;
+    for (Unit::ControlList::iterator itr = GetPlayer()->m_Controlled.begin(); itr != GetPlayer()->m_Controlled.end(); ++itr)
```
Shauren: `for (Unit* controlled : _player->m_Controlled)`

src/server/game/Handlers/PetHandler.cpp
```diff
@@ -525,66 +525,76 @@ void WorldSession::HandlePetSetAction(WorldPacket& recvData)
...
+        if ((*itr)->GetEntry() == pet->GetEntry() && (*itr)->IsAlive())
+            pets.push_back(*itr);
+
+    for (std::vector<Unit*>::iterator itr = pets.begin(); itr != pets.end(); ++itr)
```
Shauren: `for (Unit* pet : pets)`

src/server/game/Handlers/PetHandler.cpp
```diff
@@ -738,12 +748,22 @@ void WorldSession::HandlePetSpellAutocastOpcode(WorldPacket& recvPacket)
...
-        ((Pet*)pet)->ToggleAutocast(spellInfo, state != 0);
-    else
-        pet->GetCharmInfo()->ToggleCreatureAutocast(spellInfo, state != 0);
+    std::vector<Unit*> pets;
+    for (Unit::ControlList::iterator itr = GetPlayer()->m_Controlled.begin(); itr != GetPlayer()->m_Controlled.end(); ++itr)
+        if ((*itr)->GetEntry() == pet->GetEntry() && (*itr)->IsAlive())
+            pets.push_back(*itr);
 
-    charmInfo->SetSpellAutocast(spellInfo, state != 0);
+    for (std::vector<Unit*>::iterator itr = pets.begin(); itr != pets.end(); ++itr)
```
Shauren: `for (Unit* pet : pets)`

src/server/game/Handlers/PetHandler.cpp
```diff
@@ -738,12 +748,22 @@ void WorldSession::HandlePetSpellAutocastOpcode(WorldPacket& recvPacket)
         return;
     }
 
-    if (pet->IsPet())
-        ((Pet*)pet)->ToggleAutocast(spellInfo, state != 0);
-    else
-        pet->GetCharmInfo()->ToggleCreatureAutocast(spellInfo, state != 0);
+    std::vector<Unit*> pets;
+    for (Unit::ControlList::iterator itr = GetPlayer()->m_Controlled.begin(); itr != GetPlayer()->m_Controlled.end(); ++itr)
```
Shauren: `for (Unit* controlled : _player->m_Controlled)`

Shauren (changes requested):

DanVS:
Shauren, in 3.3.5 they've action bar. For example, [see this video](https://www.youtube.com/watch?v=fm4mHH5i2Yw).

And switching to ranged `for` breaks this for some reason 0_o (does compile, but do not work anymore). Can you please check this out?

Shauren:
Of course it does not work. If you kill the primary pet (select it from pet frame) then you can no longer command them because client doesn't send any pet control packets at all, this is not something you can fix by modifying said packet handlers (loops work as intended when they are alive)

DanVS:
Well, I made additional testing with this PR, and results are following:

- killing non-primary pet still allows to save control on remaining, so everything is OK there;
- killing primary pet actually makes action bar to **temporally** disappear, but after some time it turns back and you can still control remaining. However, this time is pretty, so most of the summon time they are not responding.

I'd say that PR works 50/50 now. Probably it's not worth to merge?

Shauren:
When I tested it, killing primary pet simply greyed out the pet bar and it remained unusable for the duration of feral spirit

DanVS:
If you press something when bar is greyed it disappear, but then reappear again.

jackpoz:
this is how I handled it 8 years ago: https://github.com/arcemu/arcemu/commit/1e4b99d889314d6578e1f72f33aed6e464990221#diff-12a46c1c88d79d5ab384cc6949e2734eR810

When the main summon died (say wolf A in the Feral Spirit case), I set the main summon to the other summon still alive and sent again the spells to the client.

DanVS:
Jackpoz, thanks! But is seems that this is already handled in [Unit.cpp](https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Unit/Unit.cpp#L5961-L5992).

illfated:
I presume you mean the 3.3.5 version, although it is identical:
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L5947-L5978
Then again, are we sure that it is handled as intended or the way jackpoz suggested?

DanVS:
I think I figured it out. Action bar (and primary pet) reappear after corpse of died summon had removed.  I can't find a way to change that.

DanVS:
Should be fixed now.

src/server/game/Entities/Creature/TemporarySummon.cpp
```diff
@@ -300,6 +300,21 @@ void Minion::RemoveFromWorld()
     TempSummon::RemoveFromWorld();
 }
 
+void Minion::setDeathState(DeathState s)
+{
+    Creature::setDeathState(s);
+    if (s == JUST_DIED && IsGuardianPet())
```
Seyden: That just looks so horrible.. couldn't you please do it like:

if (s != JUST_DIED || !IsGuardianPet())
    return;

and so on.. that would look a lot better :D

Seyden: Smth like that

`void Minion::setDeathState(DeathState s)
{
    Creature::setDeathState(s);
    if (s != JUST_DIED || !IsGuardianPet())
        return;

    Unit* owner = GetOwner();
    if (!owner || !owner->IsPlayer() || owner->GetMinionGUID() != GetGUID())
        return;

    for (Unit* controlled : owner->m_Controlled)
    {
        if (controlled->GetEntry() == GetEntry() && controlled->IsAlive())
        {
            owner->SetMinionGUID(controlled->GetGUID());
            owner->SetPetGUID(controlled->GetGUID());
            owner->ToPlayer()->CharmSpellInitialize();
        }
    }
}
`

Killyana:
DanVS, could test if this issue is fixed or not with your commit https://github.com/TrinityCore/TrinityCore/issues/19770

DanVS:
Is doesn't fix an issue given.

src/server/game/Entities/Creature/TemporarySummon.cpp
```diff
@@ -300,6 +300,27 @@ void Minion::RemoveFromWorld()
...
+    if (!owner || owner->GetTypeId() != TYPEID_PLAYER || owner->GetMinionGUID() != GetGUID())
+        return;
+                
+    for (Unit* controlled : owner->m_Controlled)
+    {
+        if (controlled->GetEntry() == GetEntry() && controlled->IsAlive())
+        {
+            owner->SetMinionGUID(controlled->GetGUID());
+            owner->SetPetGUID(controlled->GetGUID());
+            owner->ToPlayer()->CharmSpellInitialize();
```
jackpoz: Should there be a break; here ?

DanVS: It should. Thanks!

src/server/game/Entities/Creature/TemporarySummon.cpp
DanVS: Fixed. Thanks!

DanVS:
@Shauren, @jackpoz is it okay now?

jackpoz:
The little I commented about has been fixed so that's good, I didn't test it ingame tho.

illfated:
In case you were wondering about the Codacy status:
Codacy does not support MySQL yet, so SQL "errors" can be ignored.
The .cpp "error" might be a false positive or not applicable in TC source.

DanVS:
@Shauren, are some changes still requested?

jackpoz:
let's see what happens

jackpoz:
thanks for the PR :)

Issue #21755, opened 2018-04-01
Fedora 27 make error: MYSQL_SERVER_VERSION was not declared in this scope


maramound:
**Description:**

Fedora server 27 (64-bit), mariadb-server-3:10.2.13-2.fc27.x86_64, Trinity 3.3.5.
Fatal error during make process.

**Current behaviour:**
```
/root/TrinityCore/src/common/Debugging/Errors.h:50:110: note: in definition of macro ‘WPFatal’
 #define WPFatal(cond, ...) ASSERT_BEGIN do { if (!(cond)) Trinity::Fatal(__FILE__, __LINE__, __FUNCTION__, ##__VA_ARGS__); } while(0) ASSERT_END
                                                                                                              ^~~~~~~~~~~
make[2]: *** [src/server/database/CMakeFiles/database.dir/build.make:178: src/server/database/CMakeFiles/database.dir/Database/DatabaseWorkerPool.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1125: src/server/database/CMakeFiles/database.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```
**Expected behaviour:**

No error?

**Steps to reproduce the problem:**

- install Fedora server 27 (64-bit)
- dnf install mariadb-server
- systemctl enable mariadb
- systemctl start mariadb
- git clone -b 3.3.5 git://github.com/TrinityCore/TrinityCore.git
- cmake ../ -DCMAKE_INSTALL_PREFIX=/root/trinity-server -DTOOLS=1 
- make

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

commit 6939f4eb7b023a6b2f8940ab12484523e1904860 (HEAD -> 3.3.5, origin/3.3.5)

Copy the result of server debug command (if you need to run it from client get prat addon)

sorry, don't know how to get prat addon. i do have these however:
[cmake-output.txt](https://github.com/TrinityCore/TrinityCore/files/1866708/cmake-output.txt)
[make-output.txt](https://github.com/TrinityCore/TrinityCore/files/1866709/make-output.txt)

**Operating system:** Fedora server 27 (64 bit)

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you and add the missing information to **this issue**. Thanks.

tkrokli:
[...]

PR #22031, opened 2018-06-13
Fixes for building with MariaDB.

adeutscher:
**Changes Proposed**:

Adjust 2 information printouts and load one additional header if MariaDB is detected.

* `DatabaseWorkerPool.cpp`: Adjust MySQL version check error message to use a different constant if MariaDB is detected. The error message is the component being adjusted. The error check itself remains unchanged (goes off of `MYSQL_VERSION_ID`, which is available in MariaDB as-is and is not a problem).
* `cs_server.cpp`: Adjust MySQL version announcement to use a different constant if MariaDB is detected.
* `MySQLConnection.cpp`: If MariaDB is detected, then also load in `server/errmsg.h`. The file contains necessary constants that TC normally expects to be in `errmsg.h`.

Build changes all hinge on `MARIADB_VERSION_ID` being defined. Information printouts will else back to the pre-existing message using `MYSQL_SERVER_VERSION` if `MARIADB_VERSION_ID` is not defined.

**Target branch(es)**: 3.3.5

**Issues addressed**: Brought up in #21755.

Strictly-speaking, MariaDB is not supported by TC. Putting this forward for consideration based on the affected sections being mostly console printouts.

If this PR gets rejected on those grounds, then putting this code here will at least provide a clear paper trail for people to be able to Google and work into their own codebase at their discretion.

**Tests performed**: Built auth and world servers using `mariadb-devel` on Fedora 28. Successfully authenticated, signed in, mugged an NPC, and signed out.

### DIFF
--- a/src/server/database/Database/DatabaseWorkerPool.cpp
+++ b/src/server/database/Database/DatabaseWorkerPool.cpp
@@ -56,8 +56,16 @@ DatabaseWorkerPool<T>::DatabaseWorkerPool()
 {
     WPFatal(mysql_thread_safe(), "Used MySQL library isn't thread-safe.");
     WPFatal(mysql_get_client_version() >= MIN_MYSQL_CLIENT_VERSION, "TrinityCore does not support MySQL versions below 5.1");
+    #ifdef MARIADB_VERSION_ID
+    // Server appears to be using MariaDB instead of MySQL
+    // Hackfix for slightly different constant names.
+    WPFatal(mysql_get_client_version() == MYSQL_VERSION_ID, "Used MySQL library version (%s) does not match the version used to compile TrinityCore (%s). Search on forum for TCE00011.",
+        mysql_get_client_info(), MARIADB_BASE_VERSION);
+#else
+    // Normal MySQL server.
     WPFatal(mysql_get_client_version() == MYSQL_VERSION_ID, "Used MySQL library version (%s) does not match the version used to compile TrinityCore (%s). Search on forum for TCE00011.",
         mysql_get_client_info(), MYSQL_SERVER_VERSION);
+#endif
 }
 
 template <class T>
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -25,6 +25,9 @@
 #include "Transaction.h"
 #include "Util.h"
 #include <errmsg.h>
+#ifdef MARIADB_VERSION_ID // hack for MariaDB storing necessary values in server/errmsg.h instead of the expected `errmsg.h` from MySQL.
+#include <server/errmsg.h>
+#endif
 #ifdef _WIN32 // hack for broken mysql.h not including the correct winsock header for SOCKET definition, fixed in 5.7
 #include <winsock2.h>
 #endif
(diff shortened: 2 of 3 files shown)


Aokromes:
I am compiling with maria db under debian 9 without any modification dunno....

adeutscher:
Odd, perhaps it's a quirk between distributions? The version of the `mariadb-devel` package on F28 at the moment is 10.2.14-1.

Aokromes:
Maybe...

mariadb  Ver 15.1 Distrib 10.1.26-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

adeutscher:
My main worry now is that the file structure might be a bit different between our two versions. On your system, does `/usr/include/mysql/server/errmsg.h` exist (not to be confused to `/usr/include/mysql/errmsg.h`, which TC already expects)?

If `/usr/include/mysql/server/errmsg.h` does not exist under Debian's MariaDB, then that would be a split that would absolutely bring this whole PR crashing down, leaving the header change to be revised.

Aokromes:
nope no /usr/include/mysql/server/errmsg.h

adeutscher:
@Aokromes I set up a Debian VM to do some double-checking on how Debian sets things up.

To confirm your last comment for other people reading this, the people who maintain MariaDB for Debian have everything in `/usr/include/mysql` completely flat with no sub-folders `errmsg.h` looks like it includes all of the stuff in Fedora's `server/errmsg.h`. It turns out that Debian also doesn't use `MARIADB_VERSION_ID` that I hinge on my changes on, so my branch should actually build on a Debian system without incident.

This PR should definitely stay dead because I'm not familiar enough with how the MariaDB team works with the Fedora package maintainers (if they aren't the same in cases). I can't say for certain that this won't change and break the build for Debian when they bump up to MariaDB 10.2. "Stubborn Fedora users who want to stick with MariaDB" is a pretty niche slice of TC users, as well. When I posted this PR I was going mostly off of my old install notes and comments in issues like #21755 , and didn't realize that `libmariadbclient-dev` was outright in the [modern Debian install instructions](https://trinitycore.atlassian.net/wiki/spaces/tc/pages/10977288/Linux+Requirements).

Now that the I have a better understanding of the problem as a distro-specific thing, a better long term fix for everyone would be for me try contacting the Fedora package maintainers in order to request minor changes in the name of making Fedora behavior consistent with Debian. The changes would be:

* In `errmsg.h`, include `server/errmsg.h`.
* Make `mysql_version.h` (a symlink to `mariadb_version.h`) more consistent with `server/mysql_version.h`.
  * Defining `MYSQL_SERVER_VERSION` in `mariadb_version.h` is all that really needs to be done in TC's case.
  * `server/mysql_version.h does include `MYSQL_SERVER_VERSION`, but has a different value to `MYSQL_VERSION_ID` that could mess up your server at run time instead of build time.

If anyone is reading this in the future (send back a sports almanac!) and the Fedora-MariaDB problem still hasn't been fixed, the above points can be used as suggestions for really ugly manual hackfixes to the MySQL headers without having to make your own branch (or having a Debian VM just for development that might be merged into the main repo). The hackfixes built and ran on my Fedora 28 machine without a problem.

jackpoz:
@adeutscher how did it go with Fedora ?

Aokromes:
gentoo sugests:

 Ultimately, MYSQL_SERVER_VERSION should be removed and use the MYSQL_VERSION_ID integer for version identification.

adeutscher:
@jackpoz I'm afraid I never got around to reaching out to anyone responsible for the Fedora package, but this problem seems to have been resolved. It must have interfered with someone's project enough that they went and became the squeaky wheel or fixer that this problem needed.

Back when this PR was written, I made an Ansible playbook to apply the necessary kludges to `mysql.h`. This was a very ugly approach, but it got me off the ground. I've been able to build on Fedora 31 without any special steps.

PR #22041, opened 2018-06-15
[WIP] Grouped reference references fixups

### LINKED ISSUE #21725: GroupId issue when building a reference entry from reference entries
**Description:**
Time for some referenception:
Some creatures act like they have 1 drop pool that consists of some sub-droppools.
For example:
- 10% chance to get a item from a item pool that has a 60% chance to contain a poor quality item, a 30% uncommon item chance and a 10% rare item chance.

So the correct setup would be:
3 reference loot templates for all 3 rarity levels
1 reference to select one item out of those 3 references
To achieve this, you use a groupId that all three references share in this single entry.

However, now the issue: when you try to drop a single item out of 3 references for example, the group ID is getting ignored and you still can get up to 3 items within a single drop. This is wrong. The group ID should restrict the reference to return a single item only when the count is set to 1. Here is an example sql:
https://paste2.org/LXxB934z

**Current behaviour:**
- The core ignores the groupID restriction when combinding multiple reference entries into another reference entry.

**Expected behaviour:**
- The reference should return one item from those three referenced drop pools.

- use my sql file
- go to nordrassil and kill some of those scalding rock elementals
- witness how sometimes multiple gear items are dropping.

Here is a graphic to explain how the sql should behaive:
reference

**Branch(es):**
both

**TC rev. hash/commit:** 
Rev.: 161bdad06c27833094f15ef4c08ff0d693f3bcc7

Ovahlord:
**Changes proposed:**

- Disable reference processing for grouped references as this requires a seperate handling
-  _planned_ :fix processing grouped references. Need to find a proper way to gather all grouped references and process them in one strike

- [x] 3.3.5
- [x] master

**Issues addressed:**
Closes #21725 when finished

- [x] disable processing of grouped references 
- [ ] add handling for grouped references

### DIFF
--- a/src/server/game/Loot/LootMgr.cpp
+++ b/src/server/game/Loot/LootMgr.cpp
@@ -584,7 +584,7 @@ void LootTemplate::Process(Loot& loot, bool rate, uint16 lootMode, uint8 groupId
         if (!item->Roll(rate))
             continue;                                           // Bad luck for the entry
 
-        if (item->reference > 0)                            // References processing
+        if (item->reference > 0 && item->groupid == 0)          // Ungrouped references processing
         {
             LootTemplate const* Referenced = LootTemplates_Reference.GetLootFor(item->reference);
             if (!Referenced)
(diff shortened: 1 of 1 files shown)


Ovahlord:
Well, it really seems the whole loot storing procedure would need a rework for the issues to get fixed because they are stored pretty weird sooo I can't do much about it atm until the lootsystem got an rework

illfated:
Thanks for trying, even if this attempt did not pan out the way we wanted.
"You can't always win" seems to be the appropriate phrase now.

PR #17256, opened 2016-06-01
Core/Gameobject: Gameobjects that are spawned by spell despawn after …

runningnak3d:
**Changes proposed**:
- To stop irritating people that beat a boss and don't get their loot because a spell summons a GO

**Target branch(es)**: 335/6x

**Issues addressed**: Fixes #12842 

**Tests performed**: builds. Tested with .add quest 12842 / .cast 46310. Didn't notice any adverse affects to statically spawned GOs.

**Known issues and TODO list**:
- [ ]  Need to test more spell summoned GOs that have no loot
- [ ] Need to test more spell summoned GOs that should only be lootable by a single player
- [x] Need to make sure that statically spawned GOs are not affected. Statically spawned GOs never get anywhere near that code -- BUT -- had to test anyway.

**NOTE** If your Pull Request is SQL only create a ticket instead

**SUGESTION** If your Pull Request contains SQL give it one imposible date, for example 9999_99_99_99_database.sql on that way it will be free on merge.

…looting. If a GO contains quest loot it should not de-spawn until the de-spawn timer from Spell.dbc expires.

Closes #12842

Issue #12842, opened 2014-08-13
[3.3.5][6.x] Core/Loot/Gameobject: Tempspawned Gameobjects despawn after looting


kldxzzhc2:
**How it is:**
Any tempspawned chest object which is spawned with a spell from a creature (like http://www.wowhead.com/spell=46310) will despawn if **ONE single** player lootet the object.

**How it should be:**
Any tempspawned chest object which is spawned with a spell from a creature (like http://www.wowhead.com/spell=46310) should **ONLY** despawn **if the timer from the spelldescription expires** (like in the case of spell 46310 it should despawn after 1 minute)
During this time every player should be able to loot that gameobject multiple times even without a group.

Examples for quests/gos:

[Piece of Krom'zar's Banner](http://wotlk.openwow.com/item=11227)
Kaw the Mammoth Destroyer http://www.wowhead.com/spell=46310
http://www.wowhead.com/quest=10997
http://www.wowhead.com/quest=10995
http://www.wowhead.com/quest=10996
http://de.wowhead.com/quest=12713
http://www.wowhead.com/spell=18661
http://www.wowhead.com/quest=11529
http://www.wowhead.com/quest=12498
[Scythe of Antiok](http://www.wowhead.com/item=38305/scythe-of-antiok) from [Scythe of Antiok](http://www.wowhead.com/object=190354/scythe-of-antiok) for [On Ruby Wings](http://www.wowhead.com/object=190354/scythe-of-antiok).
[Drakuru's Skull](http://www.wowhead.com/item=40425/drakurus-skull) from [Drakuru's Skull](http://www.wowhead.com/object=191458/drakurus-skull).

tkrokli:
Unless anyone else has a good reason why I should not do so, I think I need to try to use this in a PR.

tkrokli:
[Brian](https://github.com/runningnak3d) / @runningnak3d , is there any chance you still have got the "proper fix" available on your "box at work"? When or if you see this message, could you please give us a hint what that fix would be, or even post the code? :smirk_cat:

Rushor:
just noticed, that this is the same issues https://github.com/TrinityCore/TrinityCore/issues/12077

12077 can be closed :D

tkrokli:
> just noticed, that this is the same issues #12077 -- 12077 can be closed :D

Well, your issue is older, has no updates and is not as specific as this issue. I am not sure if you intended to find a solution for all of those quests with one PR, did you? :smiling_imp:

[...]

PR #22124, opened 2018-07-04
[3.3.5] Core/Loot/Gameobject: Tempspawned Gameobjects after looting

PolarCookie:
This time PR to the correct branch!

Thanks to @runningnak3d #17256 for the groundwork.

This fixes summoned chest game objects for both quests where the chest should stay after looting and where it should despawn on the first loot.

Target branch(es): 3.3.5/master
- 3.3.5

Issues addressed: Closes # (insert issue tracker number)
#12842

Tests performed: (Does it build, tested in-game, etc.)
builds and works in-game

Quests tested
https://wotlk.evowow.com/?quest=5803
https://wotlk.evowow.com/?quest=10997
https://wotlk.evowow.com/?quest=10995
https://wotlk.evowow.com/?quest=10996
https://wotlk.evowow.com/?quest=4021
Game object stays and can be relooted.

https://wotlk.evowow.com/?quest=10514
https://wotlk.evowow.com/?quest=12828
https://wotlk.evowow.com/?quest=10629
https://wotlk.evowow.com/?quest=4726
Game object despawns on loot.

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -752,8 +752,8 @@ void GameObject::Update(uint32 diff)
             //! The GetOwnerGUID() check is mostly for compatibility with hacky scripts - 99% of the time summoning should be done trough spells.
             if (GetSpellId() || GetOwnerGUID())
             {
-                //Don't delete spell spawned chests
-                if (m_respawnTime > 0 && GetGoType() == GAMEOBJECT_TYPE_CHEST)
+                //Don't delete spell spawned chests, which are not consumed on loot
+                if (m_respawnTime > 0 && GetGoType() == GAMEOBJECT_TYPE_CHEST && !GetGOInfo()->IsDespawnAtAction())
                 {
                     UpdateObjectVisibility();
                     SetLootState(GO_READY);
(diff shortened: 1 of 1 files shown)


illfated:
What about the quest "Counterattack!" (https://woehead.way-of-elendil.fr/?quest=4021) and the object "Krom'zar's Banner" (https://woehead.way-of-elendil.fr/?object=164690) containing "Piece of Krom'zar's Banner" (https://woehead.way-of-elendil.fr/?item=11227) ? Will that issue be solved too? 😃

PolarCookie:
Let me test this, @illfated

PolarCookie:
Banner sticks around and can be relooted.

Seyden:
You're able now to loot it multiple times? Doesn't that mean, that a pickup quest will be faster to finish when these gobs do have these flags?

PolarCookie:
@Seyden, no; because the items from gobs that can be relooted are always of the unique kind.
By reloot I mean I loot the item, destroy it, then loot it again successfully since the gob it still around.

The case for gobs with non-unique items are safely handled by having them despawn instantly on the first time they're looted.

The flag that determines this is data3 in gameobject_template, which for chest type gobs mean consumable.
If there is a case then of some item that can be looted over and over incorrectly, just flip data3 in the database.

Seyden:
But what happens if you do have a gob which should be lootable, should not despawn at first loot and shouldn't be relootable for players who already got loot out of it? I only know a case for master branch but flipping WDB fields is a hack

PolarCookie:
Got a specific quest example?

runningnak3d:
@Seyden What you are talking about is already handled by the loot system. Just because a GO is summoned by a spell doesn't change how the loot system works.

Seyden:
Is it? what makes you so sure? iirc i was able to loot some gobs multiple times and was able to exploit it.. but was some years ago..

PolarCookie:
this is for spell summoned gobs only.

jackpoz:
thank you for this PR :)

PR #22145, opened 2018-07-10
Core/GameObject: Make TYPEID_UNIT handle DestroyGate

### LINKED ISSUE #22138: BG/IoC : Bosses Doors bug & Bosses missing
**Description:**

Some bugs related to bosses and keeps of both faction (alliance & horde) in IoC.

**Current behaviour:**

There are no faction bosses in keeps
The door that allow players to access bosses is not opening the right way (not opening at all)

**Expected behaviour:**
There should be a boss in alliance keep and in the horde keep
The door that allow players to access bosses should be opened from as soon as one of the keep doors has been destroyed.

1. Start a IoC BG (in debug mode) 
2. Destroy one of the ennemy keep doors using siege machine
You will see that the door to access boss is still closed
4. Activate gm mode and click on the door to open it
5. Walk to the location of the boss
You will see that the boss isn't spawned.

**Branch(es):**

3.3.5 (maybe on master, i'm not working on the master one at the moment so...)

**TC rev. hash/commit:** 
https://github.com/TrinityCore/TrinityCore/commit/076c7a7b32c0804013f24f46596fe316258bde1a

**Operating system:**
Linux/Debian 9

Maphaze:
**Changes proposed:**

-  Handle Battleground DestroyGate event when player is in a siege machine
- This PR is fixing IoC final phase (doors & bosses spawns) contained in `BattlegroundIC::DestroyGate`

- [x] 3.3.5
- [ ] master

**Issues addressed:**
Closes #22138

- [x] Build
- [x] Tested in-game

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -2192,9 +2192,9 @@ void GameObject::SetDestructibleState(GameObjectDestructibleState state, WorldOb
             EventInform(m_goInfo->building.destroyedEvent, attackerOrHealer);
             AI()->Destroyed(attackerOrHealer, m_goInfo->building.destroyedEvent);
 
-            if (attackerOrHealer && attackerOrHealer->GetTypeId() == TYPEID_PLAYER)
-                if (Battleground* bg = attackerOrHealer->ToPlayer()->GetBattleground())
-                    bg->DestroyGate(attackerOrHealer->ToPlayer(), this);
+            if (Player* player = attackerOrHealer ? attackerOrHealer->GetCharmerOrOwnerPlayerOrPlayerItself() : nullptr)
+                if (Battleground* bg = player->GetBattleground())
+                    bg->DestroyGate(player, this);
 
             RemoveFlag(GAMEOBJECT_FLAGS, GO_FLAG_DAMAGED);
             SetFlag(GAMEOBJECT_FLAGS, GO_FLAG_DESTROYED);


src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -2195,6 +2195,13 @@ void GameObject::SetDestructibleState(GameObjectDestructibleState state, WorldOb
             if (attackerOrHealer && attackerOrHealer->GetTypeId() == TYPEID_PLAYER)
                 if (Battleground* bg = attackerOrHealer->ToPlayer()->GetBattleground())
                     bg->DestroyGate(attackerOrHealer->ToPlayer(), this);
+            // Fix for https://github.com/TrinityCore/TrinityCore/issues/22138
```
Aokromes: no need of the fix for github part

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -2195,6 +2195,13 @@ void GameObject::SetDestructibleState(GameObjectDestructibleState state, WorldOb
...
+            // When player is in a siege machine, the attackerOrHealer->GetTypeId() is equal to TYPEID_UNIT. The first if statement at line 2196 (above) is not executed.
+            // So, the bg->DestroyGate, at line 2198, isn't called
```
Aokromes: lines will change on the future xd

Maphaze:
@Aokromes comments has been removed.

Aokromes:
No need of delete the whole comment, only github and line part :)

illfated:
Cosmetic detail: the PR title can be simplified to only use **Core/GameObject:**

Maphaze:
@Aokromes Well, it should be could like that ^^
@illfated Thanks for the tip !

illfated:
👍 Heh, now that I think about it, I have got a pet peeve for the general usage of "Fix" in PR and commit titles.
I prefer reading a slightly more detailed description (even if abbreviated) in the titles.
In this particular case, I would have written **Core/GameObject: include DestroyGate for TYPEID_UNIT**
Then again, I often pick too much on details and overestimate the importance of titles. 😄

Maphaze:
That's a good point, what about **Core/GameObject: handle DestroyGate for TYPEID_UNIT** ? 😄

illfated:
Fair enough, although in English it is better to say "make TYPEID_UNIT handle DestroyGate" (although that probably makes less sense in programming). I think it makes more sense to say something to the effect that you are adding the call to `DestroyGate` for `TYPEID_UNIT` (compared to the already existing one for players).

BTW: in theory, you could also have used "Core/Battleground: " in the title and still be correct.

Maphaze:
Title updated !

Yeah, i choose Core/GameObject because i thought it was not only for Battleground. Maybe this PR can have other effects on certain situations outside of Battlegrounds.

Shauren (approved):

Issue #22126, opened 2018-07-05
Testing Item Query Locale combined


Langerz82:
**Description:**

I noticed the Item Query Data runs a bit slow, and in searching through the code I noticed that we have an WorldPacket object for each Locale which I dont think is needed.
So I went about making a Locale Arrays for item name & description that holds the 9 languages and this is what I came up with:
https://gist.github.com/Langerz82/c769f2b56e982e791f13f309ab6daec1

old method profile:
https://ibb.co/gho9Wd

new method profile:
https://ibb.co/gja6Bd

edit:
This needs to be test on non-US clients to make sure everything works fine.

**Current behaviour:**

Server instantiates a WorldPacket for each ItemTemplate locale.

**Expected behaviour:**

Server instantiates one WorldPacket for each ItemTemplate.

**Steps to reproduce the problem:**

1. Apply Patch
2. Use Profiler

**Branch(es):**

CHANGEME 3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. 2e6100919d77 2018-06-25 00:42:30 +0200 (3.3.5 branch) (Win64, Debug, Static) (worldserver-daemon)

**Operating system:** Win 10

Langerz82:
If this is good I can do the other ones too.

Shauren:
This is not good at all, what you are doing here is removing locales entirely and always sending enUS

Langerz82:
I just need a method to update the locale when the packet is sent then. This has a significant boost in startup performance so please dont disclude it just cause it's currently bugged.

Shauren:
What you are proposing is either updating global state (terrible design idea, can never be fast & threadsafe) or making a copy (completely removing the point of cache). On top of that this is an OPTIONAL feature that can be turned off in config for development

Langerz82:
I'll try and code it so cache is preserved still.

Langerz82:
Here is what I got, does it work on non-US locales? It's meant too.
I'm not positive it's thread safe either, though I tried to make it so.

See:
https://gist.github.com/Langerz82/23ae52a6e4d098b7a735e0f728a277b2

Shauren:
reflect.diff? doesnt look related to this

Langerz82:
I got a problem the Quests arent tracking with my changes. :( When I fix it ill submit.

Langerz82:
Memory corruption due to not passing the LocaleConstant properly. :D

Langerz82:
[...]

PR #22146, opened 2018-07-10
[3.3.5] Core/Entities: Template Optimizer for Locales

Langerz82:
**Changes proposed:**
What if we created the users default locale QueryData, and created the others as requested during run-time? Just an idea, I think it could work with some changes to my code I made.
That way caching would work and it would have fast-loading times still, and a smaller memory footprint unless of course u have many locales playing during run-time.
ref:
https://github.com/TrinityCore/TrinityCore/issues/22126  

- [x] 3.3.5

https://github.com/TrinityCore/TrinityCore/issues/22126

Tested in-game with a enGB client.

- Needs to be tested on non-US clients and make sure Locales are still working fine for Items, Quests, Points of interest, Creature Names, and GameObjects.


Langerz82:
It didnt add the new files, and I removed them. Gonna have to do some of them again.

Langerz82:
I got two "styles" of code the Quest templates uses an old style and arrays, the other use vectors to store the locale data and GetLocaleString. Let me know which one you think is better and I can change.

If it does not work on other client Locales its bugged and will need further fixing (I cant test on non-GB as I cant find the client).

Langerz82:
errors bahhhh lemme fix

src/server/game/Server/Packets/QuestPackets.cpp
```diff
@@ -23,6 +23,11 @@ void WorldPackets::Quest::QueryQuestInfo::Read()
 }
 
 WorldPacket const* WorldPackets::Quest::QueryQuestInfoResponse::Write()
+{
+    return Write(LOCALE_enUS);
```
ccrs: ?

src/server/game/Entities/Creature/CreatureTemplate.cpp
```diff
@@ -0,0 +1,171 @@
+/*
+* Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+* Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
```
illfated: Lines 2 through 17 should have 1 space indentation for all the Copyright headers.
I presume the code is unchanged (copied from other files), for this MaNGOS line.

illfated:
Title suggestion: **[3.3.5] Core/Entities: Template Optimizer for Locales**

Langerz82:
Cleaned up some of my changes a bit. Does anyone know why cs_npc.cpp is still generating that error with appveyor?

illfated:
Sorry, I don't know. One of the real C++ coders in the TC team might know.
I can only guess it is an `#include` problem, but I could easily be wrong.

src/server/game/Entities/Creature/CreatureTemplate.h
```diff
@@ -0,0 +1,175 @@
...
+#define TRINITYCORE_CREATURETEMPLATE_H
+
+#include "DBCEnums.h"
+#include "Common.h"
+#include "SharedDefines.h"
+
+class WorldPacket;
+namespace WorldPackets {
+    namespace Query {
+        class QueryCreatureResponse;
```
illfated: Please follow TC C++ codestyle, curly brackets get their own separate lines in core files:
```cpp
namespace WorldPackets
{
    namespace Query
    {
        class QueryCreatureResponse;
    }
}
```

src/server/game/Entities/GameObject/GameObjectData.h
```diff
@@ -22,558 +22,18 @@
...
-            case GAMEOBJECT_TYPE_BARBER_CHAIR: return false;
-            case GAMEOBJECT_TYPE_QUESTGIVER: return questgiver.allowMounted != 0;
-            case GAMEOBJECT_TYPE_TEXT: return text.allowMounted != 0;
-            case GAMEOBJECT_TYPE_GOOBER: return goober.allowMounted != 0;
-            case GAMEOBJECT_TYPE_SPELLCASTER: return spellcaster.allowMounted != 0;
-            default: return false;
-        }
+namespace WorldPackets {
+    namespace Query {
+        class QueryGameObjectResponse;
```
illfated: codestyle

src/server/game/Entities/GameObject/GameObjectTemplate.cpp
```diff
@@ -0,0 +1,80 @@
+/*
+* Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+*
```
illfated: codestyle

src/server/game/Entities/GameObject/GameObjectTemplate.h
```diff
@@ -0,0 +1,586 @@
+/*
+* Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+*
```
illfated: codestyle

src/server/game/Entities/Item/ItemTemplate.h
```diff
@@ -21,10 +21,16 @@
...
 #include "SharedDefines.h"
-#include "WorldPacket.h"
 #include <vector>
 
 class ObjectMgr;
+class WorldPacket;
+
+namespace WorldPackets {
+    namespace Query {
+        class QueryItemSingleResponse;
```
illfated: codestyle

Langerz82: I'm not worried about semantics atm, the core structure of the code is the main concern but thanks for the heads up. ^_^

src/server/game/Entities/Creature/CreatureTemplate.h
```diff
@@ -0,0 +1,175 @@
+/*
+* Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+* Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
```
illfated: codestyle

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -4260,21 +4258,22 @@ void ObjectMgr::LoadQuests()
...
-        typedef void(Quest::*QuestLoaderFunction)(Field* fields);
+        typedef void(Quest::*QuestLoaderFunction)(LocaleConstant lc, Field* fields);
 
         char const* QueryFields;
         char const* TableName;
         char const* TableDesc;
         QuestLoaderFunction LoaderFunction;
     };
 
+
```
illfated: redundant blank line with whitespace

src/server/game/Quests/QuestDef.h
```diff
@@ -278,8 +269,8 @@ class TC_GAME_API Quest
 
         // multiple values
 
-        std::string ObjectiveText[QUEST_OBJECTIVES_COUNT];
-
+        std::string ObjectiveText[TOTAL_LOCALES][QUEST_OBJECTIVES_COUNT];
+
```
illfated: whitespace

illfated:
Sorry for not being able to comment on the coding, but cleaning up the visual oddities makes it more readable for real coders.

Seyden:
Uhm.. why did you move so much code between the files? what purpose had it?

illfated:
Seems to be a consolidating operation, to isolate closely related code into their own files where the code seems to fit more naturally.

Langerz82:
@Seyden it made sense to make some of them in there own Template files along with its data structures that were tightly coupled to the class.
Has anyone tested it on a non-US locale and is it working ok is the main issue.... Make sure you have emptied the cache folder in your Wow install directory so it tries to retrieve new non-US locales.

jackpoz:
The idea of this PR sounds quite complicated to me, I would leave it simple with the choices "cache everything at start" or "don't cache anything because we don't have enough ram".

WildCoffee:
It looks like this PR does two things. One, it splits a couple of internal-to-the-file classes into separate files. They're really only used by the classes of the files they already reside in. I don't see where this adds anything because they're not going to be used elsewhere. Two, it makes it so the server can cache multiple locale strings. I think a better approach to this would be to add a server variable that sets the default cached locale instead of defaulting to enUS. Most people connecting to a server are going to be a certain language (that's what Blizzard defines). This just seems unnecessarily complex.

Langerz82:
1. I have to fix up the ItemTemplates to not use arrays as it's a bit of a hack.

2. Maybe we just have it so it caches the main language only or not at all. That way it's much faster doing one iteration per data item instead of 9  (I think we support atm). I agree that most Servers will be targeting one locale only.

3. Generating them on the fly is pretty damn fast i believe, since we fetch all language data from the DB and store in RAM anyway.

4. This patch was mainly trying to avoid duplicate non-locale data from being stored eating up ram. Agreed it's native data types but considering the amount of records the DB has eats quite a chunk.

lost-illusion:
I think CACHE_DATA_QUERIES was made for "GET ME ALREADY BUILDED PACKET RIGHT NOW", 
If we cache queries by request it same as CACHE_DATA_QUERIES = 0, yes, later, we have ready cache for most popular items and performace should be ok, but... why ? We can cache, or didn't cache. 
If you have reasons for this (low RAM + long uptime), you can make for self, i mean, it dont "production" solution.
But maybe you can add this as one more config option ?
@Langerz82

Shauren:
I still don't get the point of this PR - the config is meant to shift time spent building the packets from runtime to startup and its doing exactly that, there is nothing to fix. If you value fast startups for development then disable it

Langerz82:
Now that I've nearly finished it I'm seeing it as being a little redundant too as you can either cache all languages or not at all. It's a nice feature just to cache only the main language you use but not really needed for the production server. Perhaps we keep the pull open and if people end up liking this feature then we re-consider it?

illfated:
1. Why are you changing line 3 in the file `contrib/conf_merge/README` from `authserver.conf.dist` to `bnetserver.conf.dist` for the `3.3.5` branch?
2. Why are you adding `CREATE DATABASE hotfixes` to the 3.3.5 branch in `sql/create/create_mysql.sql` ?
3. Why have you removed NewLine/CR/LF at the end of the following 3 files:
    `sql/old/3.3.5a/TDB44_to_TDB45_updates/world/2011_12_13_00_world_misc.sql`
    `sql/old/3.3.5a/TDB44_to_TDB45_updates/world/2011_12_16_00_world_creature_template.sql`
    `sql/old/3.3.5a/TDB44_to_TDB45_updates/world/2011_12_16_00_world_spell_bonus_data.sql`

Much of this looks unrelated to the PR, does it really belong here?

Langerz82:
Yep it's a mistake while using git, I'll fix soon thanks for pointing it out.

lost-illusion:
You can add config option to config - which locales should be cached on start, one, two, or all, or nothing.
And for all non-cached locales we can use another config option - build every time, or cache for future.
And, if we cache all locales at start - why we should for every opcode search in map and etc. 
This is strange.

PR #22198, opened 2018-07-28
Scripts/Spells: Glyph of Holy Light

DanVS:
**Changes proposed:**
- [Heal](https://woehead.way-of-elendil.fr/?spell=54968) can not be applied to the caster itself.
- Heal amount is a bit tricky part. Before changes it just took 10% from all healing done by [Holy Light](https://woehead.way-of-elendil.fr/?spell=48782). Instead, if effective healing of Holy Light is greater or equal to it's base part then Glyph of Holy Light should take 10% from effective amount. And only in other cases 10% from all healing done is taken.

**Target branch(es):** 3.3.5

**Tests performed:** Tested in game.

### DIFF
--- a/src/server/scripts/Spells/spell_paladin.cpp
+++ b/src/server/scripts/Spells/spell_paladin.cpp
@@ -792,6 +792,7 @@ class spell_pal_glyph_of_holy_light : public SpellScriptLoader
             {
                 uint32 const maxTargets = GetSpellInfo()->MaxAffectedTargets;
 
+                targets.remove(GetCaster());
                 if (targets.size() > maxTargets)
                 {
                     targets.sort(Trinity::HealthPctOrderPred());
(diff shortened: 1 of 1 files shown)


joschiwald:
can you explain the amount part a bit clearlier?

DanVS:
Sure. This is based on comparing effective Holy Light amount to casted Holy Light BasePoints (for rank 13, for instance max BasePoints = 5444). For example:
1. Casting Holy Light (rank 13) with effective heal value of 7K will result in **7000 >= 5444**, so Glyph of Holy Light will use 10% of it.
2. Casting Holy Light (rank 13) with no effective heal (7K goes in overheal) will result in **7000 < 5444**, so Glyph of Holy Ligh will use 10% of heal including overheal.

Hope, this helps.

src/server/scripts/Spells/spell_paladin.cpp
```diff
@@ -833,8 +834,15 @@ class spell_pal_glyph_of_holy_light_dummy : public SpellScriptLoader
                 if (!healInfo || !healInfo->GetHeal())
                     return;
 
+                uint32 basePoints = healInfo->GetSpellInfo()->Effects[EFFECT_0].BasePoints + healInfo->GetSpellInfo()->Effects[EFFECT_0].DieSides;
```
joschiwald: use SpellEffectInfo::CalcValue

DanVS: This is would return random value, while maximal is needed.

Kittnz:
@DanVS why wouldn't it be applied to caster?

DanVS:
See [this log](http://www.worldoflogs.com/reports/d15o9xbb6a4axu3s/xe/?s=2527&e=2679&x=spell+%3D+"Holy+Light"+OR+spell+%3D+"Glyph+of+Holy+Light"%0D%0A).

jackpoz:
Thanks for the PR :)

ccrs:
so... this benefits overheal over timed heals? makes no sense to me, but if its confirmed... blizz logic I guess

PR #22218, opened 2018-08-02
Removed a condition which blocked resummoning permanent pets after teleport.

psychobilli:
**Changes proposed:**
Removed a condition checking m_teleport_options and a constant which always seems to test false.  Original intent of the condition is unknown, although the alternate code which fires when a new map is loaded does not have this condition.

- [ x] 3.3.5
- [ ] master

**Issues addressed:** 
I didn't see any issues related to this.  I can add one if necessary.

Builds successfully.  
Pets are successfully resummoned after teleport.  
Player death and pet death correctly require the pet to be resummoned.  
Behavior of non-combat pets is unchanged between commits - companions must be resummoned in all cases.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -1717,7 +1717,6 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
                 UnsummonPetTemporaryIfAny();
         }
 
-
         if (!(options & TELE_TO_NOT_LEAVE_COMBAT))
             CombatStop();
 
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -1710,12 +1710,9 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
             return true;
         }
 
-        if (!(options & TELE_TO_NOT_UNSUMMON_PET))
-        {
-            //same map, only remove pet if out of range for new position
-            if (pet && !pet->IsWithinDist3d(x, y, z, GetMap()->GetVisibilityRange()))
-                UnsummonPetTemporaryIfAny();
-        }
+        //same map, only remove pet if out of range for new position
+        if (pet && !pet->IsWithinDist3d(x, y, z, GetMap()->GetVisibilityRange()))
+            UnsummonPetTemporaryIfAny();
 
         if (!(options & TELE_TO_NOT_LEAVE_COMBAT))
             CombatStop();
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -1710,9 +1710,13 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
             return true;
         }
 
-        //same map, only remove pet if out of range for new position
-        if (pet && !pet->IsWithinDist3d(x, y, z, GetMap()->GetVisibilityRange()))
-            UnsummonPetTemporaryIfAny();
+        if (!(options & TELE_TO_NOT_UNSUMMON_PET))
+        {
+            //same map, only remove pet if out of range for new position
+            if (pet && !pet->IsWithinDist3d(x, y, z, GetMap()->GetVisibilityRange()))
+                UnsummonPetTemporaryIfAny();
+        }
+
 
         if (!(options & TELE_TO_NOT_LEAVE_COMBAT))
             CombatStop();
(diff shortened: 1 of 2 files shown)


illfated:
I am trying to understand why this change is needed.
How can I test that summon pet or recall pet stops working?
(before your change is added to the existing source code)

psychobilli:
My mistake.

1. Use warlock or hunter and ensure a pet is active in the world.
2. Set a hearth to Dalaran.
3. Travel to a relatively distant place.  The ground below Dalaran in Cystalsong Forest is fine.
4. Hearth.  Pet does not get resummoned.

Alternatively, repeat these steps but hearth from another major city and the pet will correctly resummon after the hearth is complete.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -1710,12 +1710,9 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
             return true;
         }
 
-        if (!(options & TELE_TO_NOT_UNSUMMON_PET))
```
joschiwald: removing this option is hack

illfated: Thanks for pointing it out. The change looked a bit odd to me, but I didn't know why.

psychobilli: I thought that might be too easy, but pets definitely don't teleport correctly.

The calling method is structure as such:
```
ToPlayer()->TeleportTo(target, TELE_TO_NOT_LEAVE_TRANSPORT | TELE_TO_NOT_LEAVE_COMBAT | TELE_TO_NOT_UNSUMMON_PET | (casting ? TELE_TO_SPELL : 0));
```
in Unit.cpp.  For hearthing, portals and item teleports it appears to evaluate to 30 consistently.  I could change that condition to check ```if (!(options == 30 ...```, which would correct this.

Although I don't quite follow the usage of the enum in ```if (!(options & TELE_TO_NOT_UNSUMMON_PET))```
Is that just to act as a label?

Shauren: This is meant to not unsummon any pets when doing a short distance teleport like a blink

psychobilli: I see.  I did some reading on bit masks and the condition makes a lot more sense now, although I'm still not convinced it's necessary.  The line immediately following the condition is checking pet proximity and blocking the unsummon if it's within range of the new location.  When I hearth from outside the inn to inside the inn on my warlock, my imp just turns around and runs over to me.  No un/resummon.

It seems like the only case where this might be useful is when the pet is.
1. Out of range.
2. Not intended to be resummoned after a teleport.

I can't think of a single use case where that would apply.

Although if you're that intent on keeping the condition, the correct change would probably be to remove the enum from the calling method in Unit.cpp.  That should be removed either way.  I'll do some more digging to see if I can find any hidden side effects, but I've been driving this for a few of days now and it looks good.

psychobilli:
I updated the changes to restore the enumeration check and removed the enum from the calling method.  I've also written a test report which can be found here: https://github.com/psychobilli/TrinityCore/wiki/PetTeleportFix-Test-Notes

src/server/game/Entities/Player/Player.cpp
```diff
@@ -1717,6 +1717,7 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
                 UnsummonPetTemporaryIfAny();
         }
 
+
```
illfated: Redundant blank line, please remove
(unless there is more code to be added later).

psychobilli: Done.

sanctum32:
what about this?
```
void Unit::NearTeleportTo(Position const& pos, bool casting /*= false*/)
{
    DisableSpline();

    {
        WorldLocation const target(GetMapId(), pos);
        uint32 tele_options = TELE_TO_NOT_LEAVE_TRANSPORT | TELE_TO_NOT_LEAVE_COMBAT;

        // Determine range
        float checkRange;
        if (GetMap()->IsBattlegroundOrArena())
            checkRange = sWorld->GetMaxVisibleDistanceInBGArenas();
        else if (GetMap()->IsDungeon() || GetMap()->IsRaid())
            checkRange = sWorld->GetMaxVisibleDistanceInInstances();
        else
            checkRange = sWorld->GetMaxVisibleDistanceOnContinents();

        if (GetDistance(pos) < checkRange)
            tele_options |= TELE_TO_NOT_UNSUMMON_PET;

        if (casting)
            tele_options |= TELE_TO_SPELL;

        ToPlayer()->TeleportTo(target, tele_options);
    }
    else
    {
        SendTeleportPacket(pos);
        UpdatePosition(pos, true);
        UpdateObjectVisibility();
    }
}
```

Killyana:
This PR was intended to fix https://github.com/TrinityCore/TrinityCore/issues/23325

Killyana:
@sanctum32 it' better to open a PR, you will get more help and reviews.

CraftedRO:
cmangos pushed same change two years later https://github.com/cmangos/mangos-classic/pull/421/files so probably is not so hack since it works.

PR #22268, opened 2018-08-14
[3.3.5] Map-wide update ticks

Treeston:
We currently do a bunch of iterating over all players and active objects and transports in the map, and then we iterate over all grids near each of these, and then we check if that grid has had an update tick already, and _then_ if it hasn't we finally do the update tick.

All of that is a ton of pointless work, and still it leads to eternal bugs.

We're ripping all of that out. Updates now happen for everything that's loaded.

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1069,19 +1069,9 @@ void WorldObject::setActive(bool on)
         return;
 
     if (on)
-    {
-        if (GetTypeId() == TYPEID_UNIT)
-            map->AddToActive(ToCreature());
-        else if (GetTypeId() == TYPEID_DYNAMICOBJECT)
-            map->AddToActive((DynamicObject*)this);
-    }
+        map->AddToActive(this);
     else
-    {
-        if (GetTypeId() == TYPEID_UNIT)
-            map->RemoveFromActive(ToCreature());
-        else if (GetTypeId() == TYPEID_DYNAMICOBJECT)
-            map->RemoveFromActive((DynamicObject*)this);
-    }
+        map->RemoveFromActive(this);
 }
 
 void WorldObject::SetFarVisible(bool on)
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -25610,14 +25610,8 @@ void Player::RemoveAtLoginFlag(AtLoginFlags flags, bool persist /*= false*/)
 
 void Player::ResetMap()
 {
-    // this may be called during Map::Update
-    // after decrement+unlink, ++m_mapRefIter will continue correctly
-    // when the first element of the list is being removed
-    // nocheck_prev will return the padding element of the RefManager
-    // instead of nullptr in the case of prev
-    GetMap()->UpdateIteratorBack(this);
     Unit::ResetMap();
-    GetMapRef().unlink();
+    m_mapRef.unlink();
 }
 
 void Player::SetMap(Map* map)
--- a/src/server/game/Grids/Dynamic/TypeContainer.h
+++ b/src/server/game/Grids/Dynamic/TypeContainer.h
@@ -86,23 +86,15 @@ template<class OBJECT_TYPES>
 class TypeMapContainer
 {
     public:
-        template<class SPECIFIC_TYPE> size_t Count() const { return Trinity::Count(i_elements, (SPECIFIC_TYPE*)nullptr); }
+        template<class SPECIFIC_TYPE> size_t Count() const { return TypeContainer::Count(i_elements, (SPECIFIC_TYPE*)nullptr); }
 
         /// inserts a specific object into the container
         template<class SPECIFIC_TYPE>
         bool insert(SPECIFIC_TYPE *obj)
         {
-            SPECIFIC_TYPE* t = Trinity::Insert(i_elements, obj);
+            SPECIFIC_TYPE* t = TypeContainer::Insert(i_elements, obj);
             return (t != nullptr);
         }
-
-        ///  Removes the object from the container, and returns the removed object
-        //template<class SPECIFIC_TYPE>
-        //bool remove(SPECIFIC_TYPE* obj)
-        //{
-        //    SPECIFIC_TYPE* t = Trinity::Remove(i_elements, obj);
-        //    return (t != nullptr);
-        //}
 
         ContainerMapList<OBJECT_TYPES> & GetElements(void) { return i_elements; }
         const ContainerMapList<OBJECT_TYPES> & GetElements(void) const { return i_elements;}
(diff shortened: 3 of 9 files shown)


Treeston:
_If it wasn't obvious, this is still WIP. A lot of logic still exists and is just dummied out to various extents to make the core build. Comment accordingly._

Ovahlord:
In how far is this affecting the performance for huge maps like Eastern Kingdoms and Kalimdor? Because if all objects are getting updated on those maps, including waypoints, pathfinding etc. wouldn't this cause some heavy load or at least increasing the startup time by alot when loading all maps right at the beginning?

Takenbacon:
🍿

Treeston:
re-enabled grid unload. going with just global update ticks for loaded objects.

Treeston:
Grid unloading and loading all grids are not the same, btw.

Palabola:
@Ovahlord I guess 32bit support even droped on Smart Phones, but this change doesn't affect the start time or the base memory usage.

jackpoz:
I'm curious to know what's the max update time diff across all maps after loading all mapcells

Treeston:
Uses about 9gb of ram on current core, so not really feasible either way.

pedroevil51:
Testing this, my diff got up by about 200-300

Kittnz:
How about an area update ? And load and unload nearby area's (not grids).
But having a general map update is great too if we finally can make global world events that need updates.

Shauren:
Area update is out of the question, that would need reading maps (and vmaps), that can't be fast. No logic is the fastest

Kittnz:
I guess thats why blizz has different servers for each continent.

Treeston:
@pedroevil51 to confirm, your diff went up after just applying this PR (no config changes)?

pedroevil51:
@Treeston Yes, i always used default config on the files
I applied the PR, I opened the server for my players to test, with more or less 100-200 players online, my diff was oscillating between 200 and 300 without stopping, I did not see climbing more than that, but also I could not test with PVP because the server was already catching enough, with delays of up to 3 seconds

Treeston:
what was your diff before applying the PR?

pedroevil51:
@Treeston About 60, high spikes at 113
i never changed any settings so i don't know how to decrease this. but the max was 113 that i saw

Shauren:
Higher diff is expected with this change all grids that have been loaded but no longer have players in them are still updated (and that is the point of this change) however 300ms seems a bit excessive

Takenbacon:
@Shauren Keep in mind that's only with content up to WoTLK, if you start including newer expansion content it will scale out of control pretty quickly.

Krudor:
Putting current issues with the diff aside, it doesn't really matter if more content impacts it.
It shouldn't be designed around running on a single machine, even though configuration could allow for it.

xvwyh:
Updating the entire world all the time is overkill. Not updating the entire world is bad as well: spawns with long waypoint paths that span across entire zones will never have the chance to travel them, if there isn't a player constantly walking nearby, keeping the grids active. Both are too extremes, but may I suggest a more moderate alternative?

#### Approach A, "update in waves":
Designate every grid cell an update "order". Update cells near players every server tick, as usual, but update the rest of the map in that designated order. Say, let's make them 4 (in reality you want a higher number), the grid would look like this:

| 1 | 2 | 3 | 4 | 1 |
|---|---|---|---|---|
| 2 | 3 | 4 | 1 | 2 |
| 3 | 4 | 1 | 2 | 3 |
| 4 | 1 | 2 | 3 | 4 |

In the first update tick we update only cells with order of 1, in the second update tick we update only those with 2, and so on. And so by the end of the update cycle each cell would be updated once, but the process of updates would be more evenly spread than performing all the updates all the time (huge perf hit) or performing all the updates occasionally (lagspikes). Each cell would, of course, have to accumulate the amount of time passed between its updates to pass it to Unit::Update(...) etc.

#### Approach B, "in order of decreasing importance":
Have an internal counter in every cell that will count the amount of time passed since a player was last nearby. The more time passes - the less frequently the cell gets updated. That way cells adjacent to players can still be updated all the time, and the rest of the cells will gradually "fade out" as the players leave them, until they eventually reach something like "update once every 2 seconds" for open world and "update never" for instances. By having this gradation, each cell will automatically be on a different update cycle from the rest, thus also evenly spreading update load and preventing lagspikes. But it will also let cells that the players have just left from pausing updates entirely, causing issues like players being unable to reenter instances due to BossState not having resetted because the boss hasn't yet returned to their home position etc.

---

This all will become important once you implement AOIs ("area of interest", or in more common terms, "custom visibility distance"), if you haven't already, and notice that creatures in faraway cells aren't moving, acting or getting respawned, even though the players can see them and so on.

---

> While I would love to see grids becomming deprecated

Yeeah, that will never happen. Grids, be it 2D or 3D, and octrees are the industry standard ways to spatially partition your game's objects, and you NEED that partitioning. You wouldn't want to have to iterate over tens of thousands of creatures and objects on the entire map every time you need to check which enemies did your 10yd AoE hit, do you?

Treeston:
Note: This PR does _not_ update the entire world all the time. Even _loading_ the entire world all the time is not feasible for a large section of the user base (doing so takes ~7GB of memory).

All this does is change the iteration mode for the _loaded_ parts of the world.

xvwyh:
Well I'm actually leaning towards the idea that a proper **server** should load the entire world at once AND never unload it AND keep it all updated, at least every once in a while. Otherwise you can't ensure that creature with long paths will ever even spawn, not to mention complete their journey. You can't make proper large/gigantic/infinite AOI, if the world isn't getting loaded farther than one grid away from players (they would only start being visible once a player visits that area at least once, and even then, if you have grid unloading enabled, they will stop being visible shortly). I understand that 7 GB might be a bit too much for a general programmer's workstation, and you of course wouldn't need all that loaded for your daily bugfixing tasks, but the support should be there, in case the server would be running on an actual server machine, to which 32-64 GB of RAM is a regular occurrence.

Treeston:
The support _is_ there. You can set `BaseMapLoadAllGrids = 1` and `GridUnload = 0` to make the core perpetually load and update all grids.

PS: This isn't really about _us_ not being able to load the core. Most of us can afford to dump 9GB of RAM into debugging, especially with swap slowdowns not really mattering with a small handful of characters connected. You'd be surprised how many people run TC on smaller servers to play with family/friends.

xvwyh:
The support to load and prevent unloading the grid is there, sure. But with this PR, if I understand it correctly, all the loaded grids (which under aforementioned setting will essentially mean "all the grids") will get updated every tick, which would be an insane performance hit. My comment was about some ways to alleviate that hit, to spread the stress around over a couple of seconds while only losing some update precision in the grids that aren't directly visited by players.

ccrs:
Given the current state of the core's implementation and the given points disclosed in this discussion and other similar ones, I do believe this is an improvement.

In my opinion, the desing and implementation, which I cannot review sadly, should be still worked on and eventually merged.

jackpoz:
I liked the idea of updates in waves, we can simplify it splitting the grid in a fixed configurable number of waves A where every B world updates we update the next wave, not caring if a player was there or not. "CellX + CellY % A == current wave" looks like a simple check

robinsch:
I'd rather would like to see Approach B which worked out for me in most terms of quick performance gain. I pretty much use the same system to not frequently update creature relocation notifiers, afk continent players or even player out of sight visibility notifiers. This worked out for me most of the time.

jackpoz:
Closing this PR

PR #22296, opened 2018-08-21
[WIP] [3.3.5] Core/AI: AI update throttling

Treeston:
Retail doesn't update its AI all the time. We do.

After this, we don't. That's it really. This is part 1 of a PR chain.


Treeston:
Note: Next part will be moving `MoveInLineOfSight` to `UpdateAI`. Immediately NPC aggro is not blizzlike either.

TODO on this: Kill `DoMeleeAttackIfReady` with fire, that thing should really not exist in its current form (and it breaks entirely with this PR)

Faq:
So there will be first PR to remove DoMeleeAttackIfReady and fix scripts where it is used?

WildCoffee:
Is this going to allow grids with no players to have a lower update frequency? This might help with server load problems because they won't need to be updated as frequently.

src/server/game/Entities/Creature/Creature.h
```diff
@@ -62,7 +62,6 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
 {
     public:
         explicit Creature(bool isWorldObject = false);
-        virtual ~Creature();
```
soulfrost: deleting virtual destructor???

Ali-Khazaee: It's virtual :-)

soulfrost: i see

Palabola:
This one is very effective actually, we are using much more and frequent AI tick than on retail. There are AI update just in every 500ms.

Ovahlord:
Pretty sure though this will also requires alot of AI reworks in order to fix event timings for scripts unless eventMap processing is getting moved to a own routine.

Ovahlord:
Summoning at @Krudor . I guess that's your time to shine?
Throwing this in here from a discussion with him and me:  https://www.youtube.com/watch?v=DU4r_eXUG_s  This kinda shows how the AI reset and interval works

Krudor:
I'd say the video merely demonstrates the aggro check update interval.
Whether or not it coincides with the update interval I can't say, but I would assume so.

Another thing about AI update intervals is porting all boss scripts away from using the DamageTaken hook for checking percentage health into an update event.
While that is not depending on this or future parts of this PR, it's related in a way to slowing down the update rate, so thought I'd just throw the idea in here.

Love these changes by the way.

src/server/game/Entities/Creature/Creature.h
soulfrost: https://www.quantstart.com/articles/C-Virtual-Destructors-How-to-Avoid-Memory-Leaks

Treeston: `Unit::~Unit()` is virtual

lachtanek:
I imagine there are some creatures that have scheduled timers for less than 1 second, it's gonna be pretty hard (or, rather, time-consuming) finding those, wouldn't it be better to enable this on a case-by-case basis, so no AI is just left broken?
Also, what about UpdateVictim, isn't it a problem that it will happen much less frequently now? (I forgot how it works exactly with threat and stuff, so maybe not)

Traesh:
CONFIG_INTERVAL_AIUPDATE should be 500, there is lot of scripts, there is a lot of scripts required that precision (maybe even more for some scripts)

Example for LK, data taken from DBM addon but there are many other examples :

```
local timerCombatStart		= mod:NewCombatTimer(53.5)
local timerPhaseTransition	= mod:NewTimer(62.5, "PhaseTransition", 72262, nil, nil, 6)
local timerSoulreaper	 	= mod:NewTargetTimer(5.1, 69409, nil, "Tank|Healer")
local timerSoulreaperCD	 	= mod:NewNextTimer(30.5, 69409, nil, "Tank|Healer", nil, 5, nil, DBM_CORE_TANK_ICON)
local timerHarvestSoul	 	= mod:NewTargetTimer(6, 68980)
local timerHarvestSoulCD	= mod:NewNextTimer(75, 68980, nil, nil, nil, 6)
local timerInfestCD			= mod:NewNextTimer(22.5, 70541, nil, nil, nil, 5, nil, DBM_CORE_HEALER_ICON)
local timerNecroticPlagueCleanse = mod:NewTimer(5, "TimerNecroticPlagueCleanse", 70337, "Healer", nil, 5, DBM_CORE_HEALER_ICON)
local timerNecroticPlagueCD	= mod:NewNextTimer(30, 70337, nil, nil, nil, 3)
local timerDefileCD			= mod:NewNextTimer(32.5, 72762, nil, nil, nil, 3, nil, DBM_CORE_DEADLY_ICON)
local timerEnrageCD			= mod:NewCDTimer(20, 72143, nil, "Tank|RemoveEnrage", nil, 5, nil, DBM_CORE_ENRAGE_ICON)
local timerShamblingHorror 	= mod:NewNextTimer(60, 70372, nil, nil, nil, 1)
local timerDrudgeGhouls 	= mod:NewNextTimer(20, 70358, nil, nil, nil, 1)
local timerRagingSpiritCD	= mod:NewNextTimer(22, 69200, nil, nil, nil, 1)
local timerSummonValkyr 	= mod:NewCDTimer(45, 71844, nil, nil, nil, 1)
local timerVileSpirit 		= mod:NewNextTimer(30.5, 70498, nil, nil, nil, 1)
local timerTrapCD		 	= mod:NewNextTimer(15.5, 73539, nil, nil, nil, 3, nil, DBM_CORE_DEADLY_ICON)
local timerRestoreSoul 		= mod:NewCastTimer(40, 73650, nil, nil, nil, 6)
local timerRoleplay			= mod:NewTimer(162, "TimerRoleplay", 72350, nil, nil, 6)
```

Krudor:
> CONFIG_INTERVAL_AIUPDATE should be 500, there is a lot of scripts requiring that precision (maybe even more for some scripts)

I would say spell timers shouldn't run on the AI updater interval, but a separate one.

It's pretty easy to see that the aggro check interval for example isn't running on the same interval as spell casts.
Spell casts timers are much more precise.

lachtanek:
that would require changing almost all the scripts tho

Treeston:
Merged AI management changes into base 3.3.5 in 042f551 and b4cdf55. As a result, this PR now has a very manageable changeset 😂

Ovahlord:
I think UpdateAI() should be seperated from the regular AI. So we still update events etc in our regular manner for control's sake but keep things like aggro updates etc on this new level.
sth like UpdateAI() --> UpdateScriptedAI() instead.

joschiwald:
what happen to TaskScheduler::Async in scripts? it is not longer async, it is delayed for 1 sec

xvwyh:
>Retail doesn't update its AI all the time. We do.

Retail's AI is a complex system that automatically casts spells, choosing them based on their priority, possibility of cast (e.g. having mana, not being silenced or interrupted), necessity (might cast assist spells on their allies if enabled and there are those in range), determines where and how far to move, and so on.

TC's AI is a hardcoded behavior that determines creatures' actions exactly and gives them no free will to make decisions whatsoever.

Retail's AI "thinks", and that thinking doesn't have to be performed every time. The core takes care of executing the consequences of that thinking.

TC's AI executes a carefully pre-planned script where everything has to be checked and performed exactly on time (eventmap updates, scheduler updates, health % updates, melee attack).

Retail's AI is data-driven.

TC's AI is behavior-driven.

That's the difference. Until we get there, there's no point in denying us the tools to tightly control the script, otherwise blizzlike scripting will be impossibly, as there will simply be not enough functionality to implement all the minor details of boss's retail behavior.

Bosses on retail have no tight scripting, they have multiple sets of spells they can cast (which scripts are able to switch around); they have event-driven trigger sets (which scripts can also switch) which cause actionsets (essentially SAI's actionlists, but with wait conditions) to execute... and that's about it. Need to perform an intricate selection of targets for a spell? They just cast a "primer" spell that hits all the players, and spell's script decides which of the affected targets the spell should hit, and triggers the proper spell on them afterwards (or the boss's triggerset catches the primer spell hit and casts the proper spell in turn, just like our OnSpellHitTarget hook). Said primer spell will most often be serverside, so we would have no access to it, we won't even know it exists. By missing all this critical info, the best we can hope for, is to hardcode whatever behavior we do know about the boss, and hope it doesn't break.

Krudor:
Wow, yeah.

Extremely well written and agree with you 100%.

PR #22325, opened 2018-08-27
NPC Vendor Challenge

GibbsDEV:
I change the npc vendor, because alliance race can not buy sets from this npc

**Changes proposed:**

-  I change the npc vendor, because alliance race can not buy sets from this npc
-  
-  

- [ ] 3.3.5
- [ ] master

- [ ] 
- [ ]


Aokromes:
thx but we don't use c++ for sql only fixes, also, we don't insert without delete, and we don't delete creature_templates, instead update. plz open a ticket with proper updates.

PR #22350, opened 2018-08-30
[WIP] Core/Chat Add a "CommandArgs" helper to deal with command parameters

Traesh:
**Changes proposed:**

- Add a "CommandArgs" class, usable in every command script
- Provide a simpler way to get command parameters, by declaring what you expect and checking what you got
- Support for optional args
- Update HandleGoGraveyardCommand to use this class in order to provide an use case

**Target branch(es):**

- [X] master

**Issues addressed:** None

**Tests performed:** Build, tested ingame with .go graveyard

- I did not use built-in tokenizer because of limitation when it come to quoted args (iirc you have to know if it's a quoted arg before calling tokenizer, but i may be wrong)
- No issue known, would require testing & improvement ideas (support more "ARG_...", code optimization, ...)

### DIFF
--- a/src/server/game/Chat/Chat.cpp
+++ b/src/server/game/Chat/Chat.cpp
@@ -1211,43 +1211,54 @@ void CommandArgs::Initialize(std::initializer_list<CommandArgsType> argsType)
         // Finally, we cast all our args to their types
         for (uint8 i = 0; i < argsTypeVector.size(); ++i)
         {
-            switch (argsTypeVector[i])
+            if ((argsTypeVector[i] & COMMAND_ARG_INT) && isSignedNumeric(argsVector[i].c_str()))
             {
-                case ARG_INT:
-                case ARG_INT_OPTIONAL:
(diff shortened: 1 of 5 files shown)


src/server/game/Chat/Chat.cpp
```diff
@@ -1187,3 +1187,161 @@ LocaleConstant CliHandler::GetSessionDbLocaleIndex() const
...
+                case ARG_PLAYER:
+                case ARG_PLAYER_OPTIONAL:
+                {
+                    PlayerResult result;
+                    _handler->extractPlayerTarget((char*)argsVector[i].c_str(), &result.PlayerPtr, &result.Guid, &result.Name);
+                    _args.push_back(result);
+                    break;
+                }
+                case ARG_OPTIONAL_BEGIN:
+                    ASSERT(false, "Cannot use ARG_OPTIONAL_BEGIN as arg type");
```
soulfrost: mby static_assert? for checking typo in usage while compile

PR #22194, opened 2018-07-27
Scripts/Items: Goblin Gumbo Kettle

DanVS:
**Changes proposed:**
Implement [Goblin Gumbo Kettle](http://www.wowhead.com/item=33219) buff that causes player to occasionally belch green smoke. I did not find exact timings, so I chose 10 to 45 seconds long trigger period.

**Target branch(es):** 3.3.5

**Tests performed:** Does build and works in-game.

PR #21280, opened 2018-01-20
Core/Player: fix periodic eating and drinking emotes

Ovahlord:
Fixes drinking and eating periodic emotes. Blizzlike timings, blizzlike first timer. So ye. This is fixed. Mangos did it wrong, I did it right. Sniffs confirm it.

**Changes proposed:**
-  you may now feel like on retail when pointlessly drinking or eating stuff

**Target branch(es):** 3.3.5
- [x] 3.3.5

**Issues addressed:** Closes #12567

**Tests performed:** 
https://www.youtube.com/watch?v=xKGFQ3EF1rc
434 client so differences may occur.

PR #22380, opened 2018-09-04
Core/Spells: implemented proc flag PROC_FLAG_ON_TICK

mdX7:
**Changes proposed:**

-  Implement PROC_FLAG_ON_TICK, was previously named PROC_FLAG_KILLED, which seems to be wrong

- [ ] 3.3.5
- [x] master

https://github.com/TrinityCore/TrinityCore/pull/22194 (has to be ported, but it works the same way in 335)

Does build, tested ingame with:
42760 Goblin Gumbo
42556 Headless Horseman Climax - Body Regen (removed on death)

TODO:
- [ ] Verify tickrate with multiple sniffs / might also be the case that it's just PROC_ASAP instead of every tick (Headless Horseman would be best to check this I think)
- [ ] Check whether aura or unit update ticks are used for this proc on retail

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -10981,7 +10981,7 @@ void Unit::Kill(Unit* victim, bool durabilityLoss)
     }
 
     if (!victim->IsCritter())
-        ProcSkillsAndAuras(victim, PROC_FLAG_KILL, PROC_FLAG_KILLED, PROC_SPELL_TYPE_MASK_ALL, PROC_SPELL_PHASE_NONE, PROC_HIT_NONE, nullptr, nullptr, nullptr);
+        ProcSkillsAndAuras(nullptr, PROC_FLAG_KILL, PROC_FLAG_NONE, PROC_SPELL_TYPE_MASK_ALL, PROC_SPELL_PHASE_NONE, PROC_HIT_NONE, nullptr, nullptr, nullptr);
 
     // Proc auras on death - must be before aura/combat remove
     victim->ProcSkillsAndAuras(victim, PROC_FLAG_NONE, PROC_FLAG_DEATH, PROC_SPELL_TYPE_MASK_ALL, PROC_SPELL_PHASE_NONE, PROC_HIT_NONE, nullptr, nullptr, nullptr);
--- a/src/server/game/Spells/Auras/SpellAuras.cpp
+++ b/src/server/game/Spells/Auras/SpellAuras.cpp
@@ -345,7 +345,7 @@ Aura* Aura::Create(SpellInfo const* spellproto, ObjectGuid castId, uint32 effMas
 Aura::Aura(SpellInfo const* spellproto, ObjectGuid castId, WorldObject* owner, Unit* caster, Item* castItem, ObjectGuid casterGUID, ObjectGuid castItemGuid, int32 castItemLevel) :
 m_spellInfo(spellproto), m_castGuid(castId), m_casterGuid(!casterGUID.IsEmpty() ? casterGUID : caster->GetGUID()),
 m_castItemGuid(castItem ? castItem->GetGUID() : castItemGuid), m_castItemLevel(castItemLevel), m_spellXSpellVisualId(caster ? caster->GetCastSpellXSpellVisualId(spellproto) : spellproto->GetSpellXSpellVisualId()),
-m_applyTime(time(NULL)), m_owner(owner), m_timeCla(0), m_updateTargetMapInterval(0),
+m_applyTime(time(NULL)), m_owner(owner), m_timeCla(0), m_updateTargetMapInterval(0), m_procOnTickTimer(0),
 m_casterLevel(caster ? caster->getLevel() : m_spellInfo->SpellLevel), m_procCharges(0), m_stackAmount(1),
 m_isRemoved(false), m_isSingleTarget(false), m_isUsingCharges(false), m_dropEvent(nullptr),
 m_procCooldown(std::chrono::steady_clock::time_point::min()),
@@ -358,6 +358,9 @@ m_lastProcAttemptTime(std::chrono::steady_clock::now() - Seconds(10)), m_lastPro
 
     if (!m_periodicCosts.empty())
         m_timeCla = 1 * IN_MILLISECONDS;
+
+    if (GetSpellInfo()->ProcFlags & PROC_FLAG_ON_TICK)
+        m_procOnTickTimer = 1 * IN_MILLISECONDS;
 
     m_maxDuration = CalcMaxDuration(caster);
     m_duration = m_maxDuration;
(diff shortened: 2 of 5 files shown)


xvwyh:
The flag should be called "PROC_FLAG_HEARTBEAT". Info taken from alpha client, here are the strings:

g_spellProcEvents
**"Proc - Deal Cmbt Dmg"**
**"Proc - Take Cmbt Dmg"**
**"Proc - Kill"**
**"Proc - Heartbeat"**
"Proc - Dodge"
"Proc - Parry"
"Proc - Block"
"Proc - Swing"
"Proc - Spell Cast"
"Proc - Spell Cast Sch."
"Proc - Spell Hit"
"Proc - Spell Hit Sch."

As you can see, the first 3 match the 2-4 proc flags we have currently in a backwards order, hence I'm making the assumption that the 4th - "Heartbeat" - is the very first one which we don't have implemented.

As for the interval, my assumption is that it should be 5 seconds. With 33% proc chance on "Goblin Gumbo", if the interval was 1 sec as in the PR, we would've seen it proc waaaaay more often in whatever videos with Goblin Gumbo we have on youtube. But instead it's seen to be procing in seemingly multiples of 5 seconds.

Another (and, in fact, the main one) source that leads me to believe it should be 5 seconds is the constant `SECONDS_PER_HEARTBEAT` from alpha, which is equal to 5.0f.

Now, the only reason I didn't end up implementing this myself, was that I couldn't know who should "own" the heartbeat timer. Whether it should be the aura or the unit, or maybe it should even be mapwide. Regardless, I'm leaning more into the "it should be owned by unit" option, because of this: there is another thing that's triggered exactly every 5 seconds on live servers: `SMSG_FLIGHT_SPLINE_SYNC` (TC hasn't yet implemented it to my knowledge, since they don't fully support cyclic splines yet (with enter_cycle flag, yes, it MUST be supported for them to work properly)). THAT timer does work on per-unit basis, different creatures can send packets at different times, but they always happen exactly every 5 seconds. And thinking logically, "procs" mean that auras react to external events. If the timer was owned by the aura, why not just make it periodic_dummy or periodic_trigger_spell with 5000 amplitude? If it was the unit's heartbeat timer that triggers the proc, then it would make sense to have it as a proc.

xvwyh:
BTW, #21280 should also be implemented as a tick of a heartbeat timer.

mdX7:
naming it HEARTBEAT makes sense to me. But still gotta recheck the timer with new sniff. I had 1s delays between the procs @ headless horseman in the sniff I got.

xvwyh:
I assume you're talking about the spell 42556 "Headless Horseman Climax - Body Regen (removed on death)"? Cause it's the only one with that flag except for 43105, and that once doesn't have any visible proc effect. Please do note, that the spell it's triggering - 42587 - is SPELL_ATTR0_ON_NEXT_SWING, and hence isn't executed immediately upon being triggered but delayed until the next melee swing. It's entirely possible that the timeline of events looked like this, assuming his melee attack timer is 1sec (otherwise 2 consecutive procs of Horseman's Cleave within 1 second would be very strange):

| Time   | Action                                                                  |
|--------|-------------------------------------------------------------------------|
| ...    |                                                                         |
| X+0.5s | Aura procs and Horseman's Cleave is "prepared" to be cast on next swing |
| X+1.0s | Headless Horseman has no enemy nearby                                   |
| X+2.0s | Headless Horseman has no enemy nearby                                   |
| X+3.0s | Headless Horseman has no enemy nearby                                   |
| X+4.0s | Headless Horseman has no enemy nearby                                   |
| X+5.0s | Headless Horseman swings and Horseman's Cleave is executed              |
| X+5.5s | Aura procs and Horseman's Cleave is "prepared" to be cast on next swing |
| X+6.0s | Headless Horseman swings and Horseman's Cleave is executed              |
| ...    |                                                                         |

Shauren:
You should not take alpha proc flags as valid for wotlk, dodge/parry/block are not flags found in dbc anymore

xvwyh:
Of course they aren't valid, but these have been the basis for WoW's proc system. Over time it got changed, "dodge/parry/block" have been extracted into a different field which isn't present in DBC (i.e. our procEx), since they aren't mutually exclusive with current proc flags, but rather extend them with additional conditions. "Positive Magic Spell", "Negative Magic Spell" are each a group of their own, but "Hit", "Crit", "Miss" aren't, they just augment the former ones: "Hit with a Negative Magic Spell", "Crit with a Negative Magic Spell", "Miss with a Negative Magic Spell", so it would make logical sense to extract them into a different field. And the remaining `g_spellProcEvents` got shifted back, ultimately retaining their meaning to this day, as one could extrapolate from the fact 3 of them match the current guessed (and proven to be correct) values.

xvwyh:
To add to my earlier comment that was referencing #21280:

These are, once again, spell script hooks from alpha:
g_spellScriptHooks
"Validate"
"Precast"
"Targeted"
"Cast"
"Aura Applied"
"Aura Removed"
"Check Interrupt"
"Get Duration"
"Get Cast Time"
**"Heartbeat"**
"Take Combat Hit"
"Deal Combat Hit"
"Script Effect"
"Take Spell Hit"

Looks pretty similar to ours. One thing to note is that they had a hook for heartbeat, leading me to believe that some kind of `OnHeartbeat` method should be called on all auras applied on a unit, which will trigger a proc, if applicable, and call the respective `AuraScript` hook. The visual effects of eating, implemented in #21280, could simply be a spell script shared by all the food auras that makes the visual effect play on "Heartbeat" hook.

jackpoz:
what is the status of this PR ?

mdX7:
still stuff to do, mostly verifying stuff from sniffs. Will do that in ~1 month.

xvwyh:
Ha! Accidentally stumbled across an undeniable proof:
image
This is an excerpt from 3.3.5's `sub_579E50` which is a function that constructs tooltip for spells. If procFlags&1, then the result of $t/$T will be 5000, in all other cases it will be aura period. From this we can conclude that procFlag&1 is indeed a periodic proc, that it has a period of 5 seconds, all of which matches with my theory about it being the "Heartbeat" proc flag.

---
Another curious bit of info from it:

image
If `m_effectAuraPeriod` (which is what we SHOULD be calling the fields responsible for aura period, but instead some idiot skipped a field and called them `EffectAmplitude`, when in fact the NEXT field, `EffectValueMultiplier`, is what should really be called `m_effectAmplitude`, and it remained like this in private servers code for ages, aaaargh) is 0, the client fallbacks to 5000. We don't have this implemented to my knowledge, and such auras with 0 periodic spam like crazy if applied. There's a couple of such auras in game, and their descriptions prove this even further, as spell creators precalculated some damage/heals amounts over durations and put them in descriptions as plain text.

Warpten:
For whatever reason that 5000 is 5200 on Cata. I stumbled upon this when researching spell duration maths for clipping. The proc type bit is nowhere to be seen though, but I think heartbeat is gone from Cata+?

Shauren:
heartbeat proc flag is definitely still a thing, legion spell ID - 207438 Aggramar's Stride

Warpten:
Correction: proc type bit is still there in cata (only on items) and set to 5200 ms

xvwyh:
I have an idea of why they changed it.
https://www.wowhead.com/bluetracker?topic=13087818929#131814318748
As we know from a dev post, prior to Draenor retail has spellbatching done in increments of 400ms. Perhaps in some way heartbeat was also subject to batching, in which case it makes sense that they changed it from 5000 to 5200, because 5000 isn't divisible by 400, but 5200 is. Or maybe they haven't changed it in Cata, but it was always the case, and they only noticed it during Cata development and made a quick correction to the tooltip generation so it better reflects the real numbers players would get.

Warpten:
I am leaning towards "we oopsed it" because wrath did have spell batching. I chose to keep it as 5200 in my spell implementations

Keader:
@mdX7 

Shauren:
Heartbeat timer is on unit, not on aura

PR #22381, opened 2018-09-04
Core/Movement Handle Movement Forces

Traesh:
**Changes proposed:**

- Add a system to add and remove movement forces from unit
- Will allow to script spells using this system (in panda/wod/legion dungeons mainly)

**Target branch(es):** master
- [x] master

**Issues addressed:**
None

**Tests performed:**
None on this TC update, lot on Ashamane using the same code


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -13632,6 +13632,69 @@ void Unit::SendSetVehicleRecId(uint32 vehicleId)
...
+    moveRemoveMovementForce.MoverGUID = GetGUID();
+    moveRemoveMovementForce.SequenceIndex = m_movementCounter++;
+    moveRemoveMovementForce.ID = source;
+    SendMessageToSet(moveRemoveMovementForce.Write(), true);
+}
+
+void Unit::RemoveAllMovementForces()
+{
+    // We need to copy the map because RemoveMovementForce method will delete from original map
+    std::unordered_map<ObjectGuid, WorldPackets::Movement::MovementForce> movementForcesCopy = _movementForces;
```
Warpten: while (!empty()) { delete front }

Shauren:
Note to self when fixing it up for merge: opcodes are all wrong, need to send _UPDATE to nearby and the one with SequenceIndex field is self-only

Seyden:
Btw is movement force really only clientside? doesnt make kind of sense to me if you could block it clientside to prevent a lot of mechanics

Shauren:
@Seyden do you currently see any checks on regular movement in TC? No, you don't (almost, except for cases than can crash/freeze other players), anticheat patches do that

Shauren:
Fixing it turned out to be more than a few lines here and there

PR #22408, opened 2018-09-09
Fixed Wow Crash in TicketHandler

HamedMohsenavi:
Once a ticket submitted containing something like '|cFFDDD000|Hquest:|h|h|r' it'll crash Wow process.

First: Clone WowAddin (WoW Console) from https://github.com/tomrus88/WowAddin
Then: Use below code as a method and register the command in order to function

```
BOOL TicketUpdateText(char const*, char const* args)
{
	CDataStore Data(CMSG_GMTICKET_UPDATETEXT);
	Data.PutString("|cFFDDD000|Hquest:|hSomething|h|r");
	Data.Finalize();
	ClientServices::SendPacket(&Data);
}

```
After That: Try to compile and Inject it to Wow.exe Process using a DLL Injector
Then: Create a ticket contains something
Finally: Open console by pressing ` and enter the registered command.

The ticket will be updated to invalid characters and once a GM views the ticket. it'll crash its Wow.exe

### DIFF
--- a/src/server/game/Handlers/TicketHandler.cpp
+++ b/src/server/game/Handlers/TicketHandler.cpp
@@ -32,6 +32,8 @@
 #include "WorldPacket.h"
 #include <zlib.h>
 
+static const char InvalidCharacters[] = ":-\\";
+
 void WorldSession::HandleGMTicketCreateOpcode(WorldPacket& recvData)
 {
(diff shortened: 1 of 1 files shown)


Treeston:
Tickets should be able to contain hyperlinks. What you posted is just an example of an invalid hyperlink (which should be rejected by the server).

Hyperlink system is in the process of being rewritten, and this will be corrected as part of that.

PR #22412, opened 2018-09-10
Deleted usages of no longer existing DB2s

Palabola:
**Target branch(es):** bfa

- [ ] bfa

- [ ] check if this is correct, especially the removal of WorldMapArea and WorldMapTransforms


Shauren:
Did I *EVER* remove uses of old db2s without fully replacing their functionality (if it was possible) when doing new patches?
No.

PR #22426, opened 2018-09-13
Core/Movement: Fixed MotionMaster::MoveChase default parameters

kelno:
**Changes proposed:**

When calling MotionMaster::MoveChase with a distance, the angle was intended to be optional but actually wasn't.
In the declaration:
https://github.com/TrinityCore/TrinityCore/blob/70955ed3a68a58701ae50c8f0f9afdec4f60853d/src/server/game/Movement/MotionMaster.h#L144
Angle was never optional because the Optional object was always initialized at 0.0f if no value was provided by the caller.

**Target branch(es):** 3.3.5

**Issues addressed:** 
This fixes for example ranged PlayerAI's chasing at an angle instead of chasing directly towards target.
This should also fix a lot of other seemingly strange chase behavior in scripts, when a chase distance is used.

**Tests performed:**
Does build, tested with a hunter character mind controlled by a creature.

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -595,16 +595,6 @@ void MotionMaster::MoveChase(Unit* target, Optional<ChaseRange> dist, Optional<C
     Add(new ChaseMovementGenerator(target, dist, angle));
 }
 
-void MotionMaster::MoveChase(Unit* target, float dist, float angle)
-{
-    MoveChase(target, Optional<ChaseRange>(dist), Optional<ChaseAngle>(angle));
-}
-
-void MotionMaster::MoveChase(Unit* target, float dist)
-{
-    MoveChase(target, Optional<ChaseRange>(dist));
-}
-
 void MotionMaster::MoveConfused()
 {
     if (_owner->GetTypeId() == TYPEID_PLAYER)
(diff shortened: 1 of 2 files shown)


jackpoz:
I would add an overload with no angle instead of using 0 as magic value

illfated:
a couple of suggestions for the PR title (select or edit as you see fit):
**Core/Movement: improve MotionMaster::MoveChase default parameters**
**Core/MotionMaster: improve MotionMaster::MoveChase default parameters**

kelno:
I changed the overloads as suggested. Any other issues with this?

Treeston:
Actually - is this overload even necessary?

`Optional<ChaseRange>` should be constructible from `float` after all, as should `Optional<ChaseAngle>`...

kelno:
Yes I thought so too, that's what I tried at first but you get `cannot convert argument 2 from 'float' to 'boost::optional<ChaseRange>'`.
Maybe there is some clever way to get around this.

Treeston:
For future reference, never `git merge` 3.3.5 into your PR branch.
Always `git rebase` your PR branch on 3.3.5.

jameyboor:
`Optional<ChaseRange>` isn't constructible from `float` . It requires 2 implicit user-defined conversions. The compiler is only allowed to make one.

PR #22427, opened 2018-09-13
Core/Spell: Fix Debuff vs Debuff calculation for level scaling Auras

HelloKitty:

- [x] 3.3.5
- [ ] master

Some auras that put negative effects on a unit, like a sunder armor, are treated as buffs and removable by the player too. These should be treated as debuffs.

Closed an issue on a fork of mine related to this: https://github.com/KurthosWoW/TrinityCore/issues/30 and https://github.com/KurthosWoW/TrinityCore/issues/28

This fix makes the following assumptions.

1. That any aura will only ever be a debuff or a buff invariant of level of caster/target.
2. That level scaling is the only scaling factor to cover checks for.

Addresses the spell ID provided in the issue, 15572. Doesn't seem to cause any regressions with some random buffs and debuffs I tried.

### DIFF
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -3392,6 +3392,8 @@ bool _isPositiveEffectImpl(SpellInfo const* spellInfo, uint8 effIndex, std::unor
 
     visited.insert({ spellInfo->Id, effIndex });
 
+    //We need scaling level info for some auras that compute bp 0 or positive but should be debuffs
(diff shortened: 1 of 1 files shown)


Shauren (approved):

HelloKitty:
Just to provide more information for future reference, spell 15572 has an effect aura mod SPELL_AURA_MOD_RESISTANCE. With the mod being 0 + Level * -12.00. The calculation was based around when Level was 0, and that may have affected a wider range of debuffs too as many other aura mods used the same logic.

PR #22461, opened 2018-09-18
[Pre-WIP] [3.3.5] The AI system redesign

Treeston:
It's Happening™. Eventually. Keep reading.

So.

Our current AI system micromanages practically everything. From making sure the creature keeps swinging in melee, to attacking a new victim, to updating eventmaps or checking health. It also has a bunch of hooks that get invoked from the outside on the most miniscule of events (spell hit, _any_ damage taken, _any movement_ at all).

It's also ridiculously rigid. Events are timed down to the millisecond (+- one server tick) and will always execute at exactly that time.

Retail.... doesn't do any of that. Retail AI is (mostly) asynchronous, and written in a way where delaying AI callbacks a few ticks doesn't make much of a difference. Their scripts set behavior flags for common behavior (should auto attack, should be unkillable, etc.) and those behavior flags are then handled by the core. The result is a _large_ reduction in script complexity, and the core can also make assumptions about what scripts will or won't do. Scripts become less powerful, which is actually a good thing.

_(@xvwyh actually wrote a very nice summary of some more differences [over here](https://github.com/TrinityCore/TrinityCore/pull/22296#issuecomment-420112383) - you should read that)_

This PR is where all of that changes. We're throwing out the old AI system in its entirety.
_(Realistic aside: it'll survive as `LegacyCreatureAI` or something for now, likely with reduced capabilities.)_

Let's talk about the replacement. Timetable for actual implementation is, most likely, christmas break, because I'm about to be consumed by the upcoming semester, so there's that. Until then, let's figure out what _won't_ work in the proposed model, and then change the proposed model so it _does_ work.

Here's how I see this working out.

# The broad strokes _(aka: How Do I Do Things)_

It's a callback-based model. The script registers callbacks (functions) to happen upon certain events.

Examples of events are health thresholds and timers. Replicating the current update loop with these is possible, but not the intended usage.

Callbacks can be registered/unregistered on the fly. The idea is that any given callback sets up the next "steps".

For example, if a boss changes phases at 60% and 40%, this would look like this:

* initial setup:
  * set self unkillable (behavior flag, see further down)
  * setup callback 1 on engage
* callback 1:
  * setup callback 2 at health threshold 60%
  * setup any necessary periodic callbacks
* callback 2:
  * cancel any unneeded periodic callbacks (i'm considering a validator model so this isn't needed)
  * setup callback 3 at health threshold 40%
  * setup periodic stuff for phase 2
* callback 3:
  * same as above
  * set self killable

Not that different from our current scripts, right? The main difference is that all of the handling of callbacks is done core-side, and we don't need to call into the actual script on every tick anymore.

_(Aside: Evading would also be core-side, and that'd let us get rid of quite a few hacks. I'm also intending to fully reset the "AI object" (I'm not sure I'd even call it that - "Behavior object"?) when reaching home after an evade.)_

# Behavior flags _(aka: How Do I Do When I'm Not Doing)_

Pretty much all of our scripts do the same things. Yes, depending on the boss, there's a few special things (that's why callbacks exist) but overall they all do the same thing. We can handle those things coreside using what I call "behavior flags" that tell the core how the creature should act.

Below is a list of behavior flags I'm currently considering for implementation. Please comment if you can think of things you can't do with these.

## The combat flags _(aka: How Do I Do When I'm Doing The Do-Do)_
* Auto-engage?
  * `true`
    * if not engaged, periodically check for nearby targets we can engage
    * (this is the behavior of current `REACT_AGGRESSIVE` in base CAI)
  * `false`
    * never engage on our own, wait for other things to engage us first
* Auto-attack?
  * `true`/`false` - whether we should ever do melee swings
  * even if this is `true` we might use spells instead depending on priority!
* Using threat?
  * `NO`/`FIXATE`/`YES` - whether we should care about the threat list
  * if this is set to `NO` or `FIXATE`, the core will never override the victim selection
  * if this is set to `FIXATE` and current victim goes away, snaps back to `YES`
* Can evade?
  * `true`/`false` - what it says on the tin; whether we should evade if there's no targets to punch anymore
  * Is this even necessary? Feels like the previous flag might handle this just fine.

## The interaction flags _(aka: How The Do-Do Is Done To Me)_
* Can I take damage?
  * `true`/`false`
* Unkillable?
  * `true`/`false`
  * Will remain at 1 HP without dying if `true` is set - bosses should set this outside their final phase

## Spell priority system _(aka: I Do The Do-Do Without Any Do!)_
* Inspired by PlayerAI
* Creature has a list of spells it can cast
  * Each spell comes with properties (priority, "cooldown", target selection...)
  * Kind of like SAI, just flexible!
* This is for non-boss mobs that should be even more dynamic
* Some randomness in the selection of spell to cast

# Some more thoughts on callbacks _(aka: How Do I Do It Do the Doing)_
I'm envisioning three different kinds of callbacks that can be used more-or-less interchangably:
* "Regular" C++ callback
  * I am quite tempted to restrain what these can do, but it may be hard
  * Restraining can bite you, we might need the flexibility
  * With Great Power etc
* Lua callback (IT'S HAPPENING)
  * Because we only go into the callback rarely, we can afford the overhead now!
  * Lua is sandboxed, so less possibility for hacks 😊 
    * Relevant behavior manipulation functions exposed in the sandbox
    * Core itself isn't exposed - because callbacks shouldn't touch the core! Bad!
* DB-based callback
  * Kind of like current SAI already works!
  * Mostly for stuff like quest credit
  * "Native" spell casts should be handled by spell priority system!

Callbacks don't _have to_ be real time, and shouldn't depend on this! If a callback happens to happen a second or two late, this shouldn't break anything.

_(Aura scripts are for precise stuff. Those are staying as is.)_

# What I Want You To Do(-Do)

Think about scripts you've worked on. What things do they do that doesn't cleanly fit the model above? Bring it up. I would prefer to find out now rather than halfway into the implementation.

What things would you like it to do it currently doesn't? Do you think some things could be more streamlined? More input on how you've found retail does things? I'm interested.

Like anything above a bunch? Tell me, so I can make sure not to cut it if possible. This is a draft after all.

PS: I am truly a bit of a masochist, I suspect.

### DIFF
--- a/src/server/game/Behavior/ActionScript/ActionScript.cpp
+++ b/src/server/game/Behavior/ActionScript/ActionScript.cpp
@@ -16,41 +16,41 @@
  */
 
 #include "ActionScript.h"
+#include "ActionThreadStep.h"
+#include "Errors.h"
 #include "GameTime.h"
 
-ActionScriptThread::ActionScriptThread(ActionScript const& script, size_t initialStep)
-    : _script(script)
+ActionThread::ActionThread(Unit* owner, ActionScript const& script, size_t initialStep)
+    : _owner(ASSERT_NOTNULL(owner)), _script(script)
 {
     StepTo(initialStep);
 }
+ActionThread::~ActionThread() {}
 
-void ActionScriptThread::Update()
+void ActionThread::Update()
 {
     if (!_currentStep)
         return;
 
-    if (Optional<bool> success = EvaluateCurrentStep())
+    if (Optional<bool> success = _currentStep->Evaluate())
     {
         if (*success)
-            StepTo(_currentStep->GotoSuccess);
+            StepTo(_currentStep->GetTemplate().GotoSuccess);
         else
-            StepTo(_currentStep->GotoFailure);
+            StepTo(_currentStep->GetTemplate().GotoFailure);
     }
     else if (_stepTimeout <= GameTime::Now())
-        StepTo(_currentStep->GotoTimeout);
+    {
+        _currentStep->Abort();
+        StepTo(_currentStep->GetTemplate().GotoTimeout);
+    }
     else
         return;
 
     return Update();
 }
 
-// true means success, false means failure, none means no result
-Optional<bool> ActionScriptThread::EvaluateCurrentStep()
-{
-    return {};
-}
-
-void ActionScriptThread::StepTo(Optional<size_t> step)
+void ActionThread::StepTo(Optional<size_t> step)
 {
     if (!step)
     {
@@ -58,7 +58,7 @@ void ActionScriptThread::StepTo(Optional<size_t> step)
         return;
     }
 
-    _currentStep = &_script[*step];
-    _stepTimeout = GameTime::Now() + _currentStep->Timeout;
+    _currentStep = ActionThreadStep::StepTo(*this, _script[*step]);
+    _stepTimeout = GameTime::Now() + _currentStep->GetTemplate().Timeout;
     // @todo execute step
 }
--- a/src/server/game/Behavior/ActionScript/ActionScript.cpp
+++ b/src/server/game/Behavior/ActionScript/ActionScript.cpp
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "ActionScript.h"
+#include "GameTime.h"
+
+ActionScriptThread::ActionScriptThread(ActionScript const& script, size_t initialStep)
+    : _script(script)
+{
+    StepTo(initialStep);
+}
+
+void ActionScriptThread::Update()
+{
+    if (!_currentStep)
+        return;
+
+    if (Optional<bool> success = EvaluateCurrentStep())
+    {
+        if (*success)
+            StepTo(_currentStep->GotoSuccess);
+        else
+            StepTo(_currentStep->GotoFailure);
+    }
+    else if (_stepTimeout <= GameTime::Now())
+        StepTo(_currentStep->GotoTimeout);
+    else
+        return;
+
+    return Update();
+}
+
+// true means success, false means failure, none means no result
+Optional<bool> ActionScriptThread::EvaluateCurrentStep()
+{
+    return {};
+}
+
+void ActionScriptThread::StepTo(Optional<size_t> step)
+{
+    if (!step)
+    {
+        _currentStep = nullptr;
+        return;
+    }
+
+    _currentStep = &_script[*step];
+    _stepTimeout = GameTime::Now() + _currentStep->Timeout;
+    // @todo execute step
+}
--- a/src/server/game/Behavior/ActionScript/ActionScript.cpp
+++ b/src/server/game/Behavior/ActionScript/ActionScript.cpp
@@ -27,38 +27,40 @@ ActionThread::ActionThread(Unit* owner, ActionScript const& script, size_t initi
 }
 ActionThread::~ActionThread() {}
 
-void ActionThread::Update()
+ActionThread::ActionThreadState ActionThread::Update()
 {
-    if (!_currentStep)
-        return;
+    ActionThreadStep& step = GetCurrentStep();
 
-    if (Optional<bool> success = _currentStep->Evaluate())
+    Optional<size_t> stepTo;
+
+    if (Optional<bool> success = step.Evaluate())
     {
         if (*success)
-            StepTo(_currentStep->GetTemplate().GotoSuccess);
+            stepTo = step.GetTemplate().GotoSuccess;
         else
-            StepTo(_currentStep->GetTemplate().GotoFailure);
+            stepTo = step.GetTemplate().GotoFailure;
     }
     else if (_stepTimeout <= GameTime::Now())
     {
-        _currentStep->Abort();
-        StepTo(_currentStep->GetTemplate().GotoTimeout);
+        step.Abort();
+        stepTo = step.GetTemplate().GotoTimeout;
     }
     else
-        return;
+        return STATE_RUNNING;
 
-    return Update();
+    step.~ActionThreadStep();
+    if (stepTo)
+    {
+        StepTo(*stepTo);
+        return Update();
+    }
+    else
+        return STATE_FINISHED;
 }
 
-void ActionThread::StepTo(Optional<size_t> step)
+void ActionThread::StepTo(size_t step)
 {
-    if (!step)
-    {
-        _currentStep = nullptr;
-        return;
-    }
-
-    _currentStep = ActionThreadStep::StepTo(*this, _script[*step]);
-    _stepTimeout = GameTime::Now() + _currentStep->GetTemplate().Timeout;
-    // @todo execute step
+    ActionScriptStep const& stepTemplate = _script[step];
+    ActionThreadStep::StepTo(&_currentStep[0], *this, stepTemplate);
+    _stepTimeout = GameTime::Now() + stepTemplate.Timeout;
 }
--- a/src/server/game/Behavior/ActionScript/ActionThreadStep.cpp
+++ b/src/server/game/Behavior/ActionScript/ActionThreadStep.cpp
@@ -29,15 +29,16 @@ class NullStep : public ActionThreadStep
         void Abort() override {}
 };
 
-/*static*/ std::unique_ptr<ActionThreadStep> ActionThreadStep::StepTo(ActionThread const& thread, ActionScriptStep const& stepTemplate)
+/*static*/ void ActionThreadStep::StepTo(unsigned char* ptr, ActionThread const& thread, ActionScriptStep const& stepTemplate)
 {
-    ActionThreadStep* step;
+    if (ptr[ACTIONSCRIPT_MAX_STEP_SIZE])
+        reinterpret_cast<ActionThreadStep*>(ptr)->~ActionThreadStep();
+
     switch (stepTemplate.StepType)
     {
         case ACTIONSCRIPT_NULL:
-            step = new NullStep(thread, stepTemplate);
+            new(ptr) NullStep(thread, stepTemplate);
             break;
     }
-    step->Initialize();
-    return std::unique_ptr<ActionThreadStep>(step);
+    reinterpret_cast<ActionThreadStep*>(ptr)->Initialize();
 }
--- a/src/server/game/Behavior/ActionScript/ActionThreadStep.cpp
+++ b/src/server/game/Behavior/ActionScript/ActionThreadStep.cpp
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "ActionThreadStep.h"
+#include "ActionScript.h"
+
+class NullStep : public ActionThreadStep
+{
+    public:
+        NullStep(ActionThread const& thread, ActionScriptStep const& temp) : ActionThreadStep(thread, temp) {}
+        void Initialize() override {}
+        Optional<bool> Evaluate() override {
+            return {};
+        }
+        void Abort() override {}
+};
+
+/*static*/ std::unique_ptr<ActionThreadStep> ActionThreadStep::StepTo(ActionThread const& thread, ActionScriptStep const& stepTemplate)
+{
+    ActionThreadStep* step;
+    switch (stepTemplate.StepType)
+    {
+        case ACTIONSCRIPT_NULL:
+            step = new NullStep(thread, stepTemplate);
+            break;
+    }
+    step->Initialize();
+    return std::unique_ptr<ActionThreadStep>(step);
+}
--- a/src/server/game/Behavior/ActionScript/ActionThreadStep.h
+++ b/src/server/game/Behavior/ActionScript/ActionThreadStep.h
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef TRINITY_ACTIONTHREADSTEP_H
+#define TRINITY_ACTIONTHREADSTEP_H
+
+#include "ActionScriptDefines.h"
+#include "Optional.h"
+#include <memory>
+
+class ActionThread;
+struct ActionScriptStep;
+class TC_GAME_API ActionThreadStep
+{
+    public:
+        virtual ~ActionThreadStep() {}
+
+        static std::unique_ptr<ActionThreadStep> StepTo(ActionThread const& thread, ActionScriptStep const& stepTemplate);
+
+        virtual void Initialize() = 0;
+        virtual Optional<bool> Evaluate() = 0;
+        virtual void Abort() = 0;
+
+        ActionThread const& GetThread() const { return _thread; }
+        ActionScriptStep const& GetTemplate() const { return _template; }
+
+    protected:
+        ActionThreadStep(ActionThread const& thread, ActionScriptStep const& stepTemplate) : _thread(thread), _template(stepTemplate) {}
+
+        ActionThread const& _thread;
+        ActionScriptStep const& _template;
+};
+
+#endif
--- a/src/server/game/Behavior/ActionScript/ActionThreadStep.h
+++ b/src/server/game/Behavior/ActionScript/ActionThreadStep.h
@@ -29,7 +29,7 @@ class TC_GAME_API ActionThreadStep
     public:
         virtual ~ActionThreadStep() {}
 
-        static std::unique_ptr<ActionThreadStep> StepTo(ActionThread const& thread, ActionScriptStep const& stepTemplate);
+        static void StepTo(unsigned char* ptr, ActionThread const& thread, ActionScriptStep const& stepTemplate);
 
         virtual void Initialize() = 0;
         virtual Optional<bool> Evaluate() = 0;
--- a/sql/updates/world/3.3.5/9999_99_99_99_world_ACTIONSCRIPT_TABLES.sql
+++ b/sql/updates/world/3.3.5/9999_99_99_99_world_ACTIONSCRIPT_TABLES.sql
@@ -1,0 +1,11 @@
+-- 
+DROP TABLE IF EXISTS `actionscript_assign`;
+CREATE TABLE `actionscript_assign` (
+    `spawnType` TINYINT UNSIGNED NOT NULL,
+    `entry` INT UNSIGNED NOT NULL,
+    `index` INT UNSIGNED NOT NULL,
+    `scriptId` INT UNSIGNED NOT NULL,
+    PRIMARY KEY USING BTREE (`spawnType`,`entry`,`index`)
+);
+
+DROP TABLE IF EXISTS `actionscript_data`;
--- a/src/common/Utilities/Duration.h
+++ b/src/common/Utilities/Duration.h
@@ -32,6 +32,10 @@ typedef std::chrono::minutes Minutes;
 /// Hours shorthand typedef.
 typedef std::chrono::hours Hours;
 
+/// time_point shorthand typedefs
+typedef std::chrono::steady_clock::time_point TimePoint;
+typedef std::chrono::system_clock::time_point SystemTimePoint;
+
 /// Makes std::chrono_literals globally available.
 using namespace std::chrono_literals;
 
--- a/src/server/game/Achievements/AchievementMgr.cpp
+++ b/src/server/game/Achievements/AchievementMgr.cpp
@@ -2294,7 +2294,7 @@ bool AchievementGlobalMgr::IsRealmCompleted(AchievementEntry const* achievement)
     // it may allow more than one group to achieve it (highly unlikely)
     // but apparently this is how blizz handles it as well
     if (achievement->Flags & ACHIEVEMENT_FLAG_REALM_FIRST_KILL)
-        return (GameTime::GetGameTimeSystemPoint() - itr->second) > Minutes(1);
+        return (GameTime::GetSystemTime() - itr->second) > Minutes(1);
 
     return true;
 }
@@ -2304,7 +2304,7 @@ void AchievementGlobalMgr::SetRealmCompleted(AchievementEntry const* achievement
     if (IsRealmCompleted(achievement))
         return;
 
-    _allCompletedAchievements[achievement->ID] = GameTime::GetGameTimeSystemPoint();
+    _allCompletedAchievements[achievement->ID] = GameTime::GetSystemTime();
 }
 
 //==========================================================
--- a/src/server/game/Behavior/ActionScript/ActionScript.h
+++ b/src/server/game/Behavior/ActionScript/ActionScript.h
@@ -19,10 +19,10 @@
 #define TRINITY_ACTIONSCRIPT_H
 
 #include "ActionScriptDefines.h"
-#include "Define.h"
 #include "Duration.h"
 #include "Optional.h"
 #include <array>
+#include <memory>
 #include <vector>
 
 struct TC_GAME_API ActionScriptVariableContent
@@ -66,22 +66,25 @@ class TC_GAME_API ActionScript
     friend class ActionScriptManager;
 };
 
-class TC_GAME_API ActionScriptThread
+class ActionThreadStep;
+class TC_GAME_API ActionThread
 {
     public:
-        ActionScriptThread(ActionScript const& script, size_t initialStep);
+        ActionThread(Unit* owner, ActionScript const& script, size_t initialStep);
+        ~ActionThread();
         // @todo this needs a return value indicating whether behavior should control
         void Update();
 
+        Unit* GetOwner() const { return _owner; }
         
     private:
+        Unit* const _owner;
         ActionScript const& _script;
-        ActionScriptStep const* _currentStep;
+        std::unique_ptr<ActionThreadStep> _currentStep;
         TimePoint _stepTimeout;
 
         std::vector<ActionScriptVariableContent> _variables;
 
-        Optional<bool> EvaluateCurrentStep();
         void StepTo(Optional<size_t> step);
 };
 
--- a/src/server/game/Behavior/ActionScript/ActionScript.h
+++ b/src/server/game/Behavior/ActionScript/ActionScript.h
@@ -25,6 +25,16 @@
 #include <array>
 #include <vector>
 
+struct TC_GAME_API ActionScriptVariableContent
+{
+
+};
+
+struct TC_GAME_API ActionScriptVariableReference
+{
+
+};
+
 struct TC_GAME_API ActionScriptStepArgument
 {
 
@@ -68,6 +78,8 @@ class TC_GAME_API ActionScriptThread
         ActionScript const& _script;
         ActionScriptStep const* _currentStep;
         TimePoint _stepTimeout;
+
+        std::vector<ActionScriptVariableContent> _variables;
 
         Optional<bool> EvaluateCurrentStep();
         void StepTo(Optional<size_t> step);
--- a/src/server/game/Behavior/ActionScript/ActionScript.h
+++ b/src/server/game/Behavior/ActionScript/ActionScript.h
@@ -23,6 +23,7 @@
 #include "Optional.h"
 #include <array>
 #include <memory>
+#include <type_traits>
 #include <vector>
 
 struct TC_GAME_API ActionScriptVariableContent
@@ -67,25 +68,30 @@ class TC_GAME_API ActionScript
 };
 
 class ActionThreadStep;
+constexpr size_t ACTIONSCRIPT_MAX_STEP_SIZE = 32;
 class TC_GAME_API ActionThread
 {
     public:
         ActionThread(Unit* owner, ActionScript const& script, size_t initialStep);
         ~ActionThread();
-        // @todo this needs a return value indicating whether behavior should control
-        void Update();
+
+        // @todo return value needs to indicate whether behavior should control
+        enum ActionThreadState { STATE_RUNNING, STATE_FINISHED };
+        ActionThreadState Update();
 
         Unit* GetOwner() const { return _owner; }
         
     private:
         Unit* const _owner;
         ActionScript const& _script;
-        std::unique_ptr<ActionThreadStep> _currentStep;
         TimePoint _stepTimeout;
+
+        ActionThreadStep& GetCurrentStep() { return *reinterpret_cast<ActionThreadStep*>(_currentStep); }
+        unsigned char _currentStep[ACTIONSCRIPT_MAX_STEP_SIZE];
 
         std::vector<ActionScriptVariableContent> _variables;
 
-        void StepTo(Optional<size_t> step);
+        void StepTo(size_t step);
 };
 
 #endif
--- a/src/server/game/Behavior/ActionScript/ActionScript.h
+++ b/src/server/game/Behavior/ActionScript/ActionScript.h
@@ -1,0 +1,76 @@
+/*
+ * Copyright (C) 2008-2018 TrinityCore <https://www.trinitycore.org/>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef TRINITY_ACTIONSCRIPT_H
+#define TRINITY_ACTIONSCRIPT_H
+
+#include "ActionScriptDefines.h"
+#include "Define.h"
+#include "Duration.h"
+#include "Optional.h"
+#include <array>
+#include <vector>
+
+struct TC_GAME_API ActionScriptStepArgument
+{
+
+};
+
+struct TC_GAME_API ActionScriptStep
+{
+    ActionScriptStepType StepType;
+    std::array<ActionScriptStepArgument, ACTIONSCRIPT_STEP_ARGS_MAX> StepArgs;
+
+    Milliseconds Timeout;
+
+    Optional<size_t> GotoSuccess;
+    Optional<size_t> GotoFailure;
+    Optional<size_t> GotoTimeout;
+};
+
+class TC_GAME_API ActionScript
+{
+    public:
+        ActionScriptStep const& operator[](size_t i) const { return steps.at(i); }
+
+        auto begin() const { return steps.cbegin(); }
+        auto end() const { return steps.cend(); }
+        size_t size() const { return steps.size(); }
+
+    private:
+        std::vector<ActionScriptStep> steps;
+    friend class ActionScriptManager;
+};
+
+class TC_GAME_API ActionScriptThread
+{
+    public:
+        ActionScriptThread(ActionScript const& script, size_t initialStep);
+        // @todo this needs a return value indicating whether behavior should control
+        void Update();
+
+        
+    private:
+        ActionScript const& _script;
+        ActionScriptStep const* _currentStep;
+        TimePoint _stepTimeout;
+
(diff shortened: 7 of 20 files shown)


Scytheria23:
A great proposal. What would make it really powerful is the ability to make templates that can be applied to more than one creature with level scalability.  For example, if a template is made for a 'Frost Mage NPC', that template could be applied to innumerable different NPCs of different levels.  Does that make sense?

Treeston:
Kind of @Scytheria23, but what would this template consist of? If it's just the frostbolt spell, what's the point in having a template?

sirikfoll:
Some behavior I saw on 3.3.5 sniffs of kologarn, there were a "cast sequence", measured in seconds passed from fight start:
18s - Casts 'Arm Sweep'
28s - Casts 'Overhead Smash' (Failed due to category cooldown)
29s - Casts overhead Smash
34s - Casts Stone Grip
Those spells share category in dbc, with the cooldowns as follow: Arm Sweep(10000ms), Overhead(4000ms), Stone Grip(8000ms)
I would expect those cooldowns to be accounted for, although it seems they were not, anyways, spell category cooldowns probably needs to be considered in the new system

Treeston:
Hrm, a success/failure result for callbacks, perhaps?

xvwyh:
Since I was pinged here...
I currently don't have time to read this entire thread, but later I will do a writeup about everything I managed to gather so far about Blizzard's own scripting sytem. They do something similar to SAI's actionlists, but with waitable actions (for example "Idle for 5 seconds and move to next action"), but apparently also LUA? (So far I'm inclined to believe it's only used for spell scripts, but I'm probably wrong). IMO if you decided to design/redesign yet another AI system, might as well try do it the way Blizzard does it.

funjoker:
Do not forget master

Warpten:
> (Aside: Evading would also be core-side, and that'd let us get rid of quite a few hacks. I'm also intending to fully reset the "AI object" (I'm not sure I'd even call it that - "Behavior object"?) when reaching home after an evade.)

Keep in mind that some (most now?) bosses do not evade on party wipe, but despawn and respawn after a fixed timer.

> if this is set to `FIXATE` and current victim goes away, snaps back to `YES`

A good check for this behavior is yors'ahj in DS, adds during the black phase are set to fixate, what do they do when their target dies? (I **think** they used to ignore feign death too)

I also share concerns about callback hell. Tracing call stacks when bogus situations happen is also going to become a **lot** more difficult.

They use some sort of SAI tool indeed (It's integrated into WowEdit, and they leaked it with a screenshot of the first Onyxia - at work, but the screenshot is out there somewhere), and some NPCs in that gun'drak dungeon with the dinosaur boss).

Treeston:
> Keep in mind that some (most now?) bosses do not evade on party wipe, but despawn and respawn after a fixed timer.

Yeah, what I mean is evade and respawn would (in terms of behavior object) do exactly the same thing.

Naios:
In my opinion this issue should be seperated into two main problems:
* Unnesseccary calls to `UpdateAI` where it should be known when the AI is updated next (at the next auto attack or occuring event). I admit that there is overhead for the `EventMap` as well as the function calls for each AI, however, this overhead has not shown to be problematic in the past. Actually one should performance measure how much time is spent in `UpdateAI` relative to a whole world update.
* Really complicated and difficult to use C++ scripting API.

I did some research on this topic in the past (from which the `TaskScheduler` and the hotswap system originated from) and for me I came to the conslusions that:

* C++ could be highly suitable for scripting in case we provide a new C++ API for scipting:
  * Callbacks don't neccessarily mean that we will encounter a callback hell, we could wrap them nicely into [promises and futures](https://github.com/Naios/continuable) - *self promotion warning*. However for our use case they will lead to a frequent memory allocation for the type erasure, the impact on this also relies on the underlying executor we will use since boost asio erases the callbacks by default.
  * Fibers (stackful coroutines) are a much better option here, we could allocate small stacks a 10kb for scripted tasks that can be suspended and canceled at any time which would make it easy to embed suspendable actions even in existing scripts. We could wrap them also nicely as promises and futures that are not shared across cores which makes it possible to implement them nicely with cross referenced pointers.
  * Stackless (compiler supported coroutines) the support for this seems not good enough currently and also those could lead to inefficient memory allocation.
* A scripting language would also be an option however I highly vote for JavaScript (v8 engine) over Lua since there is much more effort put into this nowadays (Typescript for instance).

I did some experiments with boost fibers int the past actually I got the [following example to work](https://github.com/Naios/tcfiber/blob/master/test/base/main.cpp). It is similar to the design proposed by @Treeston with a single entry point:

```cpp
class MyAI : public AsyncCreatureAI
{
  public:
    void Reset() override
    {
        await OnEnterCombat();

        Async([this] {
            await OnDespawn();

            await this->CastSpell(47474);
        });

        Async([this] {
            while (auto damage = await OnDamageReceived())
            {
                // Do something
            }
        });

        while (true)
        {
            RandomEvent(
                [this] {
                    await Wait(10s, 15s);
                    await CastSpell(28373);
                },
                [this] {
                    await Wait(8s, 13s);

                    while ((await CastSpell(3746)) != SpellCastResult::Ok)
                        ;
                });
        }

        Async([this] {
            // ...
            await OnDespawn();
        });
    }
};
```

ratkosrb:
> They use some sort of SAI tool indeed (It's integrated into WowEdit, and they leaked it with a screenshot of the first Onyxia - at work, but the screenshot is out there somewhere), and some NPCs in that gun'drak dungeon with the optional dinosaur boss.

Here are the screenshots of blizzard's tools that @Warpten mentions. They were posted on mmo champion.
Creature Editor - https://static.mmo-champion.com/mmoc/images/news/2009/august/raidsdungeons_020.jpg
Creature Spells - https://static.mmo-champion.com/mmoc/images/news/2009/august/raidsdungeons_021.jpg

You can see that in the Creature Editor, which presumably is for editing blizzard's equivalent of the `creature_template` table, there is a button for "Spells/Abilities", and Onyxia has the currently selected entry from the Creature Spells editor screenshot assigned to her creature template. In the screenshot of the Creature Spells editor we can also see there are multiple "spell\ability templates" for Onyxia, for her different phases, so presumably the used "spell\ability template" can be changed in combat by the AI script.

Treeston:
Especially the creature spells editor is _super_ interesting.

ratkosrb:
We can try to deduce the following about the "creature spell lists" from that editor screenshot.
- Both normal and boss creatures use these "spell/ability" lists.
- A single creature spell/ability list can contain information for up to 10 spells.
- The actual use of the abilities is possibly done by the AI script, since there is no option to set the target of the cast or other conditions in the Creature Spells editor.
- Managing the cooldown of abilities is separate from the AI script, as cooldowns are defined in the creature spell/ability list entry.
- Cooldowns are defined in whole seconds, not milliseconds.
- Cooldowns are a range, there's a minimum and a maximum.
- The initial delay to use an ability is separate from the repeat cooldown.
- The `Availability` field seems to be a percent chance, as the default is 100.
- Creatures can have multiple spell/ability lists, for the different phases of the fight, with one being the default that is assigned to the creature_template entry.
- The Blizzard convention for naming these spell/ability list entries is:
 `<Zone Name> - <Creature Name> <Phase Name>`

I have no idea what the `Probablity` field means, it seems to have a default value of 1. Possibly a mask, specifying the allowed difficulty modes like Normal/Heroic?

Up top there is a box titled "Percent chance each tick", which lets you set a "Chance Support Action" and "Chance Ranged Attack". The value "-1" probably means "not applicable".

There is an "Edit" button next to the spell name button, so maybe clicking that brings up a pop up window that lets you set additional data for that spell like some kind of flags. Or maybe it would bring up a "Spell Editor" to edit the spell itself.

The version of the editor that was posted on mmo champion is from 2009, based on the URL, possibly shown at Blizzcon. Here is an [earlier screenshot](https://ksr-ugc.imgix.net/assets/020/990/240/30ee67f2028362a27e0205b3f01d8cd5_original.jpg?ixlib=rb-1.1.0&w=680&fit=max&v=1524519426&auto=format&q=92&s=cc3043f4e740a3f02e39e739ec8acdd6) of the tool from vanilla. The early version is missing 3 of the fields on the right, the rest is the same.

Treeston:
I would disagree with some of the points there.

* The main entry is likely _not_ just a spell ID (note it says `Creature - Dragon Tail Sweep (Onyxia) (Tail Sweep)`, not just `Tail Sweep´) - I would assume that targeting information is contained in there.
* Probability is... a probability. I'm not sure why you think it's a mask. Range `0~1` inclusive. Probability to actually do the action, maybe? Hard to tell.
* Availability.... maybe only certain instances of the creature have a spell? Seems unlikely. My guess is as good as yours.
* Initial/Repeat are almost certainly how often the action happens. This, incidentally, is the same way our SAI repeat timers work.
* I do not think the "AI script" is actually involved with the first four spells, as outlined above. The _fifth_ spell (Deep Breath) with `????` fields is probably something that's AI triggered.

xvwyh:
CreatureSpellData.dbc, hint-hint.
Sadly Blizzard has hid all the info from us except for those "spells records" used by pets.

xvwyh:
> I do not think the "AI script" is actually involved with the first four spells, as outlined above. The fifth spell (Deep Breath) with ???? fields is probably something that's AI triggered.

No, that's just a joke on players, who were always confused about the timing of that particular Onyxia attack. Blizzard teased them by showing a screenshot of Onyxia's spells, but deliberately hiding the Deep Breath values.

xvwyh:
> The main entry is likely not just a spell ID (note it says Creature - Dragon Tail Sweep (Onyxia) (Tail Sweep), not just `Tail Sweep´) - I would assume that targeting information is contained in there.

Nah, it IS spell ID, but blizzard has internal names for everything. You think any of the designers would be happy to work with 200 spells named "Fireball"? Even the creature editor there shows a field with "Internal Name".

xvwyh:
I haven't even begun writing the theoretical part yet (it will contain my thoughts on WowEdit screenshots), but I've assembled the scripting actions for you to get started:

https://docs.google.com/spreadsheets/d/1Wh1e1ZXYM-sw3wWrY9dGZ2ngK8oa3PA2XIftQJuf5R0/edit?usp=sharing

If you don't understand the meaning of some of the things in there, there's a high chance I will explain them in my writeup, but feel free to ask anyway.

ratkosrb:
> - Probability is... a probability. I'm not sure why you think it's a mask. Range 0~1 inclusive. Probability to actually do the action, maybe? Hard to tell.
> - Availability.... maybe only certain instances of the creature have a spell? Seems unlikely. My guess is as good as yours.

If the values for Probability and Availability were reversed, i would think that too. Probability is chance to cast, and Availability is possibly a mask for difficulty. But instead you can see that Availability has a value of 100 for all the spells, which seems like percent chance to me, while Probability is 1. The ??? on Deep Breath is a joke for sure btw.

Also i just noticed, from that vanilla screenshot we can also see that the "creature spell lists" contain self buffs - "Demon Armor III" is highlighted.

Treeston:
> No, that's just a joke on players, who were always confused about the timing of that particular Onyxia attack. Blizzard teased them by showing a screenshot of Onyxia's spells, but deliberately hiding the Deep Breath values.

I feel very stupid right now.

Kittnz:
687474703a2f2f692e696d6775722e636f6d2f5951434f4d766b2e6a7067

Said screenshot?

ratkosrb:
This is where the screenshots are taken from if you want to watch the panel.
https://youtu.be/nt8BffVACQE?t=2h10m5s

Scytheria23:
_What would make it really powerful is the ability to make templates that can be applied to more than one creature with level scalability.  For example, if a template is made for a 'Frost Mage NPC', that template could be applied to innumerable different NPCs of different levels._  

Well, I'm thinking for coverage beyond Blizz-like, which I know isn't TCs objective but could be considered.  Many devs have quite complex custom NCP behaviours scripted with smartAI, but these have to be individually scripted for each creature type.  A more generic template could be applied to a range of NPC, with abilities scaling by level.

Warpten:
These strings look horribly familiar and I don't know why

xvwyh:
They are in every Wow.exe. They are probably located in some source code that's shared between client, server and WowEdit (e.g. our SharedDefined.h), and thus got compiled into the exe, even though it's never used anywhere.

lost-illusion:
Stupid thing but maybe availability it's a possible count of casts ?

Treeston:
The more I think about it - `probability` might be _relative_ (the same way our `PlayerAI` works, incidentally).

If two spells have the same probability, 50% chance of each. If one has `1` and one has `2`, 33% of former, 66% of latter, and so forth.

xvwyh:
I agree. That's how probability works for selection of CreatureDisplayInfo for a creature, and for selection of BroadcastText from RandomText (aka npc_text), no reason to think why it would be any different here.

lost-illusion:
Which point for probability if we have timers ? Or it's only for "default ai", like cast one of this spells sometimes ? If we have timers and cooldown, we just can get situation with two spell at same time.

Treeston:
If multiple spells are available, which one gets cast? It's random, which is where the probability comes in.

lost-illusion:
Seems legit. But it looks like our current system (like SAI) - we have event map, and just execute.
Anyway we should have timers and other stuff, example
setup callback 2 at health threshold 60% is equal to check every taken damage for ready to call (like current system), but yes, looks like we should handle some items by another way.

Treeston:
> setup callback 2 at health threshold 60% is equal to check every taken damage for ready to call (like current system), but yes, looks like we should handle some items by another way.

No, it's not though. That's the thing. The core can just check every "so often" whether our health is below the threshold, and execute the callback if it is.

Keader:
What about boss that have specific target selector?
Like Unchained Magic (Sindragosa) [This is not implemented in current TC yet, if i remember right, ccrs is working on it] , Blood Queen Bite, General Vezax (Shadow Crash - Current implementation is little wrong)
And how we should handle with some particular movements, like Illidan when go to air phase, need select a trigger to move and particular stuffs like that?

xvwyh:
Unchained Magic is AoE already, if some target filtering has to be made, it can and should be done in the spell script.
Vampiric Bite has spell 71837 which is an AoE dummy. I would imagine it serves the targeting purpose. Boss casts this spell, be it via his periodic spell set or via scripted actions, the script takes case of filtering targets and casts the real Vampiric Bite on the appropriate one.
Shadow Crash also has 63284 "Shadow Crash Targeting" which is serverside and is missing effects, but I'm willing to bet it's also an AoE just like the above.

Keader:
Yeah, i belive that some spells that we filter using AI, blizzard use some serverside spells to do it.
But still, has alot of particular cases, behaviors that only 1 boss have... idk how it can work in some "sandbox" system xd

xvwyh:
Me neither, and so a lot of thinking has to be done before committing to build such a system...

Treeston:
> Like now. Unit take damage -> Call hook -> Calc Percentage -> Execute, in new system it will be Unit take damage -> CalcPercentage -> Callback for this percentage if exists, it's simplified, but generally used same scheme - calc percentage on damage and execute some code for this percentage

No, no and no. You are still missing the point.

1. Unit takes damage.
2. Later, on the 1hz AI update tick: "oh, we're at 59% HP, and are supposed to do something when we hit 60%". Call the callback.

xvwyh:
https://www.reddit.com/r/wow/comments/9havni/ama_friday_21st_7pm_edt_former_wow_developers/

You could try inconspicuously asking your questions to the main Wowedit developer today.

BAndysc:
Interesting, actually it is not that much work to make a basic version of this - I feel like adding ability to add callbacks on certain hp level + some kind of spell settings would already reduce UpdateAI a lot in most scripts.

But when it comes to writing new AI, maybe it could work well with something like Unreal Blueprints that I currently work on https://github.com/BAndysc/WoWDatabaseEditor, https://www.youtube.com/watch?v=esSUJpwSyW4 

> Whether it will even be feasible to develop such scripts without proper editors.

I don't think we should create the system so that new behaviors are easily writable by hand, because there are some limits - the more powerful it is, the harder it is to write, so do we want to create *not powerful* system? Contrary to this, I think the new system should be created along with good and nice tools to make scripting easy. Not creating system first and then wait for someone to create editor - create system together with editor.

jameyboor:
I think a lot _lot_ of thought has to be put into what the cons and pros are by completely overhauling the AI system. It's a big big task if you want to go from how it is now to only have data in the scripts and logic all core-side. Is it worth it? What are the benefits? What are the cons? Are we sure we have enough information to separate both logic and data?

This sounds like a good and ambitious idea but you should worry about it not getting too ambitious and ending up with at 50% completed system that breaks compiles, causes headaches and gives nothing back in return.

arazi47:
@jameyboor 
This project aims to do things the way Blizzard does things. Worth it? Yes. Benefits? Yes. Do we have enough information to not need C++ scripting? Well, probably not, but this PR is a step in the right direction.

ratkosrb:
@jameyboor Having a scripting system that works the same way Blizzard's system worked means people will have to implement things the way they would have been implemented on retail, and thus leading to a final result that is much closer to the original content. The more the systems differ, the harder it is to implement all of the details and quirks of the original content, sometimes forcing you to partially recreate the behavior of Blizzard's event based scripting system inside the AI script of individual creatures in order to properly script the encounter.

By the way, here is an interesting excerpt from the reddit AMA linked earlier. Apparently during vanilla wow, they did not even use a scripting language. So their SAI equivalent was powerful enough to script all of the original content, most likely edited only through Wowedit.
vanilla scripting

jameyboor:
@ratkosrb Indeed, but we don't know how Blizzard's system worked, just vaguely. So writing this new system it's just guessing how it might've been done. That doesn't get you closer to the retail experience.

Killyana:
Boss on Vanilla are pretty simple compared to wotlk, almost all could be scripted by SAI, I remember a discussion between a blizz dev and some fans, he said they use lua for almost everything an c++ for complicated boss, it was on cata or mop if I'm not wrong.

lachtanek:
I also think it is worth considering just using SAI instead of implementing completely new and potentially completely different system  
SAI is currently really powerful and adding some more flags for behavior, implementing some new stuff and using knowledge of actual scripts for the events instead of checking every update (on X% HP events and stuff like that) would make it even more powerful and performant and there may not even be need for implementing something new

Treeston:
SAI is an absolute clusterfuck of hacks in the backend. Not happening.

lachtanek:
it may be, but there still is a LOT of stuff made in it and there are also some editors for it already, just pointing that out

Killyana:
It's better to keep SAI for simple stuff like quests, and create a powerful tool for more complicated scripts.
The problem with SAI is you can end up with a hacky working script if you don't use it correctly.

Treeston:
> It's better to keep SAI for simple stuff like quests

Not happening. SAI is going away with this PR.

PS: There will be a conversion script.

BAndysc:
SAI lacks one feature - there should be separate "source" of action parameters. Because now often "target" of action is actually "source" of it - it makes reading and creating script hard (unless editor helps you) or source of action is actually "me" - base object of script. That forces to create hacks like "send target to target" and using "set data" just to make other creature do something. If there was "source" of action, it would be really powerful and easy to use.

Treeston:
The resulting DB-based system will probably look very similar to how SAI looks right now on the data side, which is why I'm very confident that conversion will not be a complication.

The backend is going away entirely, though.

Pharonax:
Yeah, I'm retarded. Lol. Must've been tired while going through this. Sorry.

Killyana:
What about this 3 tables: https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130258/scripts
The event_scripts and spell_scripts are more blizzlike than SAI they even use data from DBC (for event ID for example)

xvwyh:
There. I think I put everything I wanted to into words.

https://docs.google.com/document/d/12H-IMOLv2Vp8yHjmNEH00Iy9WbK0LXswLErOYcj9Z7U/edit?usp=sharing

It's supposed to go hand-in-hand with the spreadsheet I linked earlier:

https://docs.google.com/spreadsheets/d/1Wh1e1ZXYM-sw3wWrY9dGZ2ngK8oa3PA2XIftQJuf5R0/edit?usp=sharing

Read them if you want an educated guess at how Blizzard might be doing their scripting on the server side. Whether or not TC's AI will be redesigned to work like this I don't know, and, frankly, don't care much. But I hope it will at least be useful as an insight into what hooks, action and options in general need to be provided to match Blizzard's functionality, and the closest we get to it, the less (hopefully) hacks and micro-control would be necessary.

Please comment on the docs and give your suggestions, if you still don't understand certain aspects of what written/listed, I might have an answer.

searcas:
Holy moly

Warpten:
Script effect spells are definitely Lua as could be seen during that PTR build where we had access to console worldport (we got old scholomance sniffed!) and more notably any LUA error was fed back to the client. I'm waiting to receive my copy of John Staats's book to get (hopefully) not blurred out sensible information.

BAndysc:
Reading what @xvwyh has written makes me feel that SAI **is in fact really close** to what Blizzard has! 
 * "waits" can be achieved using timed events and some editors will show it as a single "wait" action (example: https://imgur.com/a/9yCK6G0 - this is TC SAI), so we already have this,
 * multiple phases can already be reached using smartphases, so we already have this
 * casting combat spells should be moved to a separate table so that there is no more "every x-y seconds: cast spell" - that's true
 * then we just need few more events (for example - "on 50% HP threshold reached")

You have mentioned in SAI there are targets, which are not present in blizzard system, but keep in mind **blizz does everything on spells**. There are a lot of dummy spells that are **only required to pass targets to other units!** And what is more, **a lot of spells are not present in client database**, because they are only dummy spells to pass targets around. So without ability to use arbitrary targets, we won't be able to script much.

Treeston:
`spell_dbc` exists for exactly that reason.

xvwyh:
> "waits" can be achieved using timed events and some editors will show it as a single "wait" action (example: https://imgur.com/a/9yCK6G0 - this is TC SAI), so we already have this,

No it cannot. It's impossible to accurately predict how long an action will take. If you want to, say, move the creature to a point and cast a spell there, then you can't anticipate how many seconds it will take for the creature to move to that point: it might have been kited outside of the room, it might be slowed by players, it might be stunned or rooted along the way etc. The idea behind waitable actions is that they will pause execution until the action is confirmed to be completed, not just for an arbitrary amount of milliseconds. We can mimic this with "On Movement Inform" event, we can mimic timed "waits" with timed actionlist, but you would agree that it's way less convenient than just writing the actions linearly, right?

(I've always wanted to have the ability to specify delays on `SMART_EVENT_LINK` to minimize the need for actionlists in simple situations)

streetrat:
my old (1.5y old) sai*revamp ideas, some stuff was based on malcroms ideas (like the pathing and actionlist)
note, there was a drag&drop app planned for it to make it readable and editable

ratkosrb:
I don't know if you saw this already, but the original Monster and Encounter designer for WoW talked about creature scripting on the Countdown to Classic podcast. The "Availability" field mystery is finally solved.

The relevant part starts at 1:48:00.
https://countdowntoclassic.com/2018/10/15/episode-75-ex-wow-dev-alex-brazie-on-monster-raid-encounter-design-blizzard-classic/

Question:
> During a panel at Blizzcon 2009, we were shown some screenshots of the tools Blizzard used to make creatures. One of them was a window displaying all of Onyxia's abilities and their respective cooldowns, but there were also some additional fields for each of her spells whose purpose is less obvious, such as Availability and Probability. And even more surprisingly, there was no option to select the target on which the ability should be used. Can you talk a bit about the inner workings of Creature AI in WoW, and how creatures actually decided when and on who to use their abilities.

Answer:
> Yes, I totally can. This was basically what my first year at blizzard was doing, [it] was editing those fields and tweaking them.
>
> Okay, so you have to realize there were multiple parts to a creature, right? So you have the Creature Record which defines how much health it is, how much damage it does, and this was basically a number. It's 1.0 for a normal creature, 2.0 for a really high damage or a high health creature, etc. So a raid boss might be a 50 or 60, right?
>
> So the next part here is the Spell Record, and what you're looking at there is a list of basically each ability. That's the character data sheet. I can't remember the exact record name, but it's basically a list of what spells they can cast. Then each spell, you know, as you can go to MMO Champion and go see the data on each spell, and each spell is all of these things like how much damage it does, and how it works, etc.
>
> And then the final thing is this thing called the Action List. The Action List are the special, hand-scripted rules that a designer created using data fields provided by the programmers. So for example that record you were looking at there is a list of all of Onyxia's spells.
>
> Availability is what percentage chance does she have to receive that ability when she spawns in. For example, we could create an Onyxia who has... uhm, i forgot what the screaming roar or the thing that makes you run in fear, what was the name of that... let's just say Dragon Roar. So let's say you can have an Onyxia that only 10% of the time has Dragon Roar. And if you happen to spawn in an Onyxia with Dragon Roar, you are really, really out of luck that week.
>
> Probability is how heavily does it try to use that ability. Let's pretend that she has Deep... uh wanted to say Deep Breath, but that's too random. Let's just say that she has Bite and Stomp, right? Those are not real abilities, but just bear with me. And Probability on Bite is 1, and Stomp is 2. That means there is a 1 out of 3 chance that... a dice roll from 0 to 3... or 1 to 3 actually. If it's a 1 or a 2, she is going to Bite, and if it's a 3 she is going to Stomp, right? That basically says the bigger that number, the more likely [it is] to choose that ability, assuming both are available.
>
> Finally, you have the cooldown. So let's say... both of them are set 2 and 2, right? So they have equal chance to happen. And one has a cooldown of 0, and the other has a cooldown of 30 seconds, what that means is you have a 50/50 chance of picking an ability. Once the ability with the cooldown is used, that ability can't be ran again by Onyxia for at least 30 seconds. No matter what the cooldown of the ability is if you had possessed the creature, or had cast the spell manually as a designer. This was important because we wanted the ability to mind control monsters, right, which was an ability everyone had in their debug menu. You would mind control a monster, then you would just press the Spell Record to test each ability individually because it's really hard to test abilities otherwise. And so you'd put a 0 cooldown on the Spell Record, and on the Spell List inside the creature you'd give the actual cooldown you wanted there. Over time we discovered this was a really bad idea because it made it harder and harder for designer to figure out where cooldown, or where the frequency of an ability was decided. So instead over time we migrated away from that data, and started putting it in the Action Records. So, an ability your creature could use basically willy nilly, and we didn't really care about how frequently you can use it, we would just put those on the record. But things we wanted to happen at like 25% health, or every 30 seconds on the dot, we would put into the Creature Action Record.
>
> So, now to answer your final piece of the question, which is how did it choose targeting. If it was attached to the Action Record, the hand crafted piece, then there was this thing called the Target Rule. And this was a database value, it was never shared outside of Blizzard databases. And it basically would say "Is My Target", "Is Not My Target", "Is Far Away", "Is Close", "Is One Of The Top 3 Far Away", "Is One Of The Top 2 Closest". And that's how it would prioritize using those abilities. This field also existed in a box called AI Hints on the Spell Record. And this was meant for things were we had like a Fireball spell that's always supposed to be cast on your current target, and we just wanted to be like - look, just always cast this on your current target. And we want all Fireball spells to work the same way, so that was defined there.
>
> Are you confused yet? Good, because so was I. There were too many ways of doing the same thing. And what became best practice down the road... uh actually Ion Hazzikostas made a really big splash on this when he came in and said - Hey look, we need to fix Wowedit. And we specifically need to make the rules for how to target monsters way, way easier to use.
>
> During Burning Crusade I had sat down and written probably about 200-300 scripts that did things like pick 1 of 3 at random, pick 1 of 4 at random, pick one of the healers, pick three of these... right? And over time these scripts were copied, and then oh pick one of these healers with this debuff, pick someone who doesn't have this debuff. And the amount of scripts you were writing had just sheerly exploded on the server. And that meant that when you fixed a bug in one of the scripts you had to go and open all of the other scripts that had been made as a copy of that bugged script and modify it there. Anyone who has ever worked on a software engineering project will immediately begin wincing because this means hours and hours of copy and paste work, hoping you don't accidentally reintroduce a bug across what might be a hundred or two hundred spells.

xvwyh:
You think you're not, but you do.
It's inevitable.

xvwyh:
Adobe called, they want their trademark back.

src/server/game/Behavior/ActionScript/ActionThreadStep.cpp
```diff
@@ -0,0 +1,43 @@
...
+        void Abort() override {}
+};
+
+/*static*/ std::unique_ptr<ActionThreadStep> ActionThreadStep::StepTo(ActionThread const& thread, ActionScriptStep const& stepTemplate)
+{
+    ActionThreadStep* step;
+    switch (stepTemplate.StepType)
+    {
+        case ACTIONSCRIPT_NULL:
+            step = new NullStep(thread, stepTemplate);
```
xvwyh: If "steps" are anything like SAI's actions or even events, dynamic allocations every time a script is going to execute an action (or event) seems very detrimental to performance. Consider using std::variant, so no allocations will need to be done and all possible steps can live within the same already-allocated memory. Although it will probably make it very inconvenient to declare said variant, since you would need to list all possible step implementations in it. Another way would be to have a pre-allocated memory arena that will host all steps (with "placement new"s), however that would require defining strict limitations on step sizes (nothing a static_assert can't handle) to make sure they can even fit inside the arena.

Treeston: Good observations. Variant would be a complete pain to work with.

Looking into placement new into a pre-allocated region in the `ActionThread`.

src/server/game/Behavior/ActionScript/ActionThreadStep.h
```diff
@@ -0,0 +1,48 @@
...
+        virtual void Abort() = 0;
+
+        ActionThread const& GetThread() const { return _thread; }
+        ActionScriptStep const& GetTemplate() const { return _template; }
+
+    protected:
+        ActionThreadStep(ActionThread const& thread, ActionScriptStep const& stepTemplate) : _thread(thread), _template(stepTemplate) {}
+
+        ActionThread const& _thread;
+        ActionScriptStep const& _template;
```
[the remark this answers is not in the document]
Treeston: Historically, not having access to (something like) the `ActionThread` has always turned out to be a maintenance nightmare. A single extra pointer is not going to kill us.

`ActionScriptStep` is only stored here, not in the thread. We need access to it.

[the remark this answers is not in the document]
xvwyh: That's 2 pointers already which are going to be duplicated on every creature in the world using ActionScript. Well, I guess as long as only the "current" steps are being stored, the memory usage might be within reason, but if at any point it's going to be like SAI, where the entire script is copied into every of its instances because it might need to be changed (SMART_ACTION_CREATE_TIMED_EVENT, SMART_ACTION_INSTALL_AI_TEMPLATE etc), this could blow up.

I still think it's inefficient to store the reference to thread if the step can never be interacted with by anything other than the thread. Store just the state of the step, pass everything else needed for it to function as parameter(s). Unless you intend to do something like `unit->GetCurrentStep().Cancel()`, which would require the step to have all the necessary information (thread, namely) to function in such independent context, but IMO that's not the best design.

But sure, this kind of optimization can be done when the design is finalized and proven to be working, if it doesn't clash any of the implemented functionality.

src/server/game/Behavior/ActionScript/ActionThreadStep.cpp
```diff
@@ -29,15 +29,16 @@ class NullStep : public ActionThreadStep
         void Abort() override {}
 };
 
-/*static*/ std::unique_ptr<ActionThreadStep> ActionThreadStep::StepTo(ActionThread const& thread, ActionScriptStep const& stepTemplate)
+/*static*/ void ActionThreadStep::StepTo(unsigned char* ptr, ActionThread const& thread, ActionScriptStep const& stepTemplate)
```
xvwyh: Is passing `ptr` really needed here, if you already have access to the thread? `&thread.GetCurrentStep()` and voila, no need for additional `reinterpret_cast`s either. Would require `thread` to not be `const`, of course.

src/server/game/Behavior/ActionScript/ActionScript.cpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+    else if (_stepTimeout <= GameTime::Now())
+    {
+        step.Abort();
+        stepTo = step.GetTemplate().GotoTimeout;
+    }
+    else
+        return STATE_RUNNING;
+
+    step.~ActionThreadStep();
```
jackpoz: it doesn't happen often to see a destruction being called explicitly

src/server/game/Behavior/ActionScript/ActionThreadStep.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+/*static*/ void ActionThreadStep::StepTo(unsigned char* ptr, ActionThread const& thread, ActionScriptStep const& stepTemplate)
+{
+    if (ptr[ACTIONSCRIPT_MAX_STEP_SIZE])
+        reinterpret_cast<ActionThreadStep*>(ptr)->~ActionThreadStep();
+
+    switch (stepTemplate.StepType)
+    {
+        case ACTIONSCRIPT_NULL:
+            new(ptr) NullStep(thread, stepTemplate);
```
jameyboor: Why placement new for this?

src/server/game/Behavior/ActionScript/ActionThreadStep.h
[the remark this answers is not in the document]
jameyboor: By storing refs you also just implicitly deleted the other special member functions, maybe not too handy for later.

src/server/game/Behavior/ActionScript/ActionThreadStep.cpp
Treeston: Because memory fragmentation is not a good thing to have in code that runs all the time, and we don't need malloc here.

Treeston:
Alright, so this is gonna be a two-parter comment. This here is the first part, where I talk about some more design considerations I've been making. Part two, probably somewhere below by the time you're reading this, is where I engage in some Real Talk™ and tell you why I'll be closing this for the time being. It's not for the reasons you think.

So, AI design. If you've been following the PR, you will have already noticed the outlines of a SAI replacement taking shape. I called it ActionScript, name subject to change. Maybe ActionSequence would be better. I don't like calling it "Script", it sounds rigid. Maybe it's accurate, though. Either way.

In my vision, this true kinda-synchronous "scripting" system would only be necessary for truly "scripted" passages in a fight. Let's take Lich King, for example - ActionScript would handle the transition phases. Triggered at a previously-set HP threshold by the core, the script would take control, stop auto-attacking, move LK to the center of the platform, start channeling big bad AoE, and start summoning stuff periodically (maybe there's even some aura that does this), wait X seconds, stop channeling, cast "break the platform stuff" spell, end of script.

Here's why - one major problem I keep running into, which I _really_ do not think should _be_ a problem, is the consideration of what happens if multiple "things" try to directly control a given unit at the same time. What if ActionScript tells the creature to move to point A, but the creature behavior wants to chase to point B? You can see how this gets annoying really quickly.

So, none of that. In my vision, **only a single "thing" has control of a given unit at any one time**. That could be ActionScript, or it could be the creature behavior engine, or something else, but never multiple of these. The creature starts executing an ActionScript, normal creature behavior stops doing anything until you finish. Once you finish, behavior comes back.

Now you will notice that I said "has control" up there. In my vision, there's two ways you can run an ActionScript - I call them "synchronous" and "asynchronous", because the names fit. Synchronously running an ActionScript is what I described above - the script takes control of the creature, and does stuff with it, and normal behavior is paused. Asynchronously running the script lets it run _concurrently_ to normal creature behavior, but you _cannot_ control the unit (this is enforced by the engine). You _cannot_ tell the unit to move to position X, or to cast spell Y, or to stop attacking. You _can_ modify behavior flags or the creature's spell list, though, and the behavior of the creature will change based on that - so you _can't_ directly cast spell Y, but you _can_ add spell Y to the creature's spell list, and the creature will cast the spell whenever behavior deems it appropriate. Stuff like "cast spell Y if condition Z is met" uses this.

Oh, and if that wasn't obvious: no two synchronous ActionScripts can run at the same time. Not sure on the details, but either the first one gets aborted, or the second one gets delayed until the first is done. Tending towards the latter.

With that, we're done with the technical stuff. Real Talk follows in the next comment.

Treeston:
Alright, part two, the Real Talk™ part.

You may recall that I promised I'd do work on this over the break, and that work has only manifested to a limited degree.

See, and this is where we veer right into Real Talk. We all make commitments sometimes - but there's a point where there are too many, and you are constantly just dealing with the next expectation looming over your head, where this gets really shitty for your mental health.

For some more context - and some of you might be aware of this already - besides being a somewhat prolific contributor here, I'm also a second-in-command for a [tiny internet spaceship alliance](http://wiki.eveuniversity.org/) of a few thousand real people, I routinely obsess over card rulings for pointless card games on reddit, and also manage the full workload of being a student and TA at the same time. Apparently, when I have hobbies, I have them _way too much_.

So yeah, for the last few years, I would wake up, and be on my phone checking Slack and Discord within the first five minutes of my day. After that, I'd go through my mail and deal with my Github inbox. After that, reading through _all the threads_ about internet spaceships. After that, over to reddit and make sure _all the card game rulings_ had accurate replies. Before I'd know it, it'd be time to leave for work, where you can guess how I spent my breaks. Once I got home, more of the same. Just before heading to bed, I'd be checking Slack, Discord and forums.

Sounds kind of like your daily routine? As it turns out, constantly being under pressure to perform _even more_ is not good for your mental health. Like, at all.

There's this thing called "burnout". Yes, I know, we don't talk about it. It makes us seem less like the perfect, efficient, always functional, well-oiled rational machines we've all been trained to be. If you don't work hard, you're letting other people down and showing you're not worth keeping around, right?

Yeah, fuck that. I spent my "break" still being torn between four different commitments, one of which is writing a new AI system. And, you know what? I still wanted to write a new AI system. Like, consciously, I wanted to. That's why there's the whole comment above with stuff I thought about.

But, mentally? I can't anymore. I'm unable to focus on the work in front of me. In other words, I've got what we'd probably call "Mental Issues™". You know, the things you're not supposed to have, because they make you seem like a less valuable job applicant, and that's all we care about anymore.

Maybe we should all face some truths about us not being all too perfectly functioning all the time sometimes, hm?

Anyway, that's my Real Talk for today. No big lessons or great wisdom in here. Just a plea to watch your own mental state, before you end up where I am right now. I'll get better, I'm sure - but it won't be today, or in a week, or maybe even in the near future.

Right now, more work is not something I am consciously deciding not to deal with, so I'm closing this.

If I ever feel like working on it again, it may come back in the future - which is decidedly _not_ a commitment. If that happens, it'll happen on my terms, because I actually _want to_ work on it, not because I set myself up for yet another commitment that pressures me into meeting expectations.

Mic is dropped. Tree out.

jackpoz:
Hej, sorry to hear to went through what many others went through already, hope you'll find your balance between work, family and hobbies. It's fun to have hobbies and it's very easy to get totally sucked up into those.

Small tip from a guy who is trying to build a whole automated AI that is supposed to test all possible already fixed TC issues and find new ones on its own: when you have an idea that sounds very interesting to work on, take a note about it, write what you would like (or "dream of") to do about it, just like this AI redesign. Maybe it will take 10 years to finish, maybe it will take 1 year to even start writing the first line of code, it will just be a background thread running when you are bored enough to work on it and not exhausted by everything else. It will allow you to be curious about things while keeping your sanity and having time for your life :)

Kittnz:
Iv found out that sometimes you have to pause stuff and let it sit a for a while and when you feel up to work on something, do it. But don't do it because you promised it. Keep your head above water @Treeston .

ccrs:
I really feel you.
Each person is unique, of course, but I'd say that I went through something almost identical last year.
In my opinion, looking at my experiences and results, the decision you've taken is the most wise and mature. As a quick reference, I took the path of tasks/commitments/obligations prioritization and division, and have to say I'm really happy with the results.

Even if we had our differences in opinion in the past, I have to say it was a pleasure debating, discussing and even arguing about literally everything. I really think I've learned from that, hope it was a two-way thing.

Really wish everything goes well for you, I mean it.
Dont hesitate to contact me any time.

Cris.

illfated:
Thank you for all the things you already have provided to TC.
I struggle with autism, so I can only imagine the stress you feel.
I wish you all the best in your endeavours, as well as finding a healthier life.

jameyboor:
Remember this is all open-source free contributions you're making here. You should never feel pressured to write anything, as you've already contributed a lot. Take some time off @Treeston . Thanks for your work.

nawuko:
Hey Treeston, you have done awesome work for the community and seem all arround like a really nice guy. Had to deal with the same issue aswell in the past, and it will get better, but like you said, it takes time. Wish you all the best ❤️

Issue #22467, opened 2018-09-20
Core: Gossip Select Breaks After Creature::UpdateEntry


HelloKitty:
**Description:**

When Creature::UpdateEntry is set CMSG_GOSSIP_SELECT_OPTION packet will fail to be fully handled the next time it is sent. Meaning it takes two times to select the gossip selection before it will work. It fails because the scriptid has changed and, see line 147 in MiscHandler.cpp.

**Current behaviour:**

Creature::UpdateEntry breaks CMSG_GOSSIP_SELECT_OPTION handling.

**Expected behaviour:**

Creature::UpdateEntry should not affect the ability to gossip after updating.

**Branch(es):**

3.3.5

**TC hash/commit**: 5843724debc3642434c055e5cf6f29a1eaf65358

**Operating system:** Win7

I will make a pull request with a suggested change, but I don't know the best approach. There are a couple ways to address it. Such as not returning from the handler after setting the script id but also setting the last script id to the current one in UpdateEntry. Not 100% sure what is the best way.

Initially reported here: https://github.com/KurthosWoW/TrinityCore/issues/194

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you and add the missing information to **this issue**. Thanks.

HelloKitty:
5843724debc3642434c055e5cf6f29a1eaf65358

xvwyh:
There's also another issue in the code that can lead to similar results.
In `ScriptMgr::GetDialogStatus` for some weird reason there's a call to `player->PlayerTalkClass->ClearMenus();`. `GetDialogStatus` is called whenever the client queries info about whether there are quests available from all nearby NPCs. So any time an NPC with questgiver npcflag would appear within client's visibility distance, the client will request "dialog status" for that NPC and that will in turn erase all serverside info about the gossip menu currently sent to the client. This can potentially be abused by summoning creatures that can be questgivers (e.g. Archmage Vargoth, Oracles/Wolvars NPCs, Children's Week NPCs etc) and breaking gossip for all surrounding players, if they happened to be in the process of it.

That code was there in MaNGOS even as far back as 2006, I doubt it will even be possible to find who added it and why.

jackpoz:
https://github.com/TrinityCore/TrinityCore/commit/d5888a5f8f498790dccde186325ab5bec87b066d

PR #22469, opened 2018-09-20
Fix issue in Creature::UpdateEntry and CMSG_GOSSIP_SELECT_OPTION

HelloKitty:
Address #22467 making it so that script doesn't look like it was reload
and thus causing gossip to fail once after update.

**Changes proposed:**

-  Make last script ID same as current script when UpdateEntry is called

- [x] 3.3.5
- [ ] master

#22467

In a fork of mine I use UpdateEntry in DM:E on Old Ironbark. Before it took two gossip select attempts and now it only takes one so I believe it works.

If this effects other things I do not know, but it seems like it shouldn't? It seems like it's just for detecting potential script reloads? I don't know.

None, other than QA but don't know the surface area of what this change actually touches.


Keader:
Treeston always say: in 99% of cases that use UpdateEntry is hack.

HelloKitty:
@Keader Maybe, I would not know. I only used it because of what VMangos was doing for DM:E Iron Bark.

Either way this is still probably a bug, I don't think it's the desired behavior that select gossip should fail once after an UpdateEntry.

Keader:
Instances cases, UpdateEntry need be changed like https://github.com/TrinityCore/TrinityCore/commit/c78f454c0ee74000ab7f77463e8b44ff7480fdec

HelloKitty:
@Keader that makes sense I think, but DM:E creature does not change until a boss dies somewhere in the instance. Suppose to morph into a similar NPC. I don't know if the same approach?

Is it possible to make it so that GetCreatureEntry returns the newer entry by checking if a boss is dead? I don't know exactly what the return value for GetCreatureEntry is used for, what do you think?

Also, even if I did change my use it still seems like this is a bug in the end and should be fixed even if I end up moving away from UpdateEntry.

Keader:
GetCreatureEntry is called before to create ANY creature inside of instances.
So you can put any logic there.

HelloKitty:
But there are cases such as Dire Maul: East Old Ironbark where it's not determined at creation time. It's a deterministic event yes, it should happen when Zevrim dies, but not dependent on anything but that and happens in the middle of the instance run.

Maybe I misunderstand and overriding GetCreatureEntry is the solution but it seems to me that you're saying it's involved at creation time. But this particular case requires it long after creation time.

Either way though, even if I don't use UpdateEntry I still think it would be a bad idea for it to temporarily break Gossip as it currently does.

Keader:
Oh, it should be created with EntryX, and after fight should change to EntryY?

HelloKitty:
@Keader yes yes exactly, it's odd but in DM:E after one of the bosses dies that creature is suppose to turn from Old Ironbark to Relieved Ironbark or something, don't recall the name but basically yea. It changes in the middle of the instance. Created with EntryX and changed to EntryY after one of the bosses dies.

Keader:
Hmm, so i agree with this change.
@Treeston can be merged?

ccrs:
of topic: @HelloKitty can you join #trinitycorelegacy on irc?

Treeston:
Change as proposed looks like a hack to me.

`LastUsedScriptID` apparently exists to make sure gossip IDs still have the same meaning when the gossip option is processed. You don't want script A to send a gossip menu like

```
[ ] yes, please give me 50,000 gold
[ ] nah, i'm good
```

and by the time the player selects option 0 the loaded script expects the gossip ID for option 0 to belong to this gossip menu

```
[ ] yes, i want to donate 50,000 gold
[ ] i'll pass for now
```

tl;dr `LastUsedScriptID` should be set in whatever function sends the gossip menu to the player. IMO it should also be renamed to something akin to `LastGossipScriptID`.

HelloKitty:
You bring up a good point @Treeston but the current system does not actually protect players from what you describe. For instance, imagine 4 people have the gossip window open and the entry changes with UpdateEntry. A 5th person comes along, tries to gossip and the packet handler then sets the new script id.

The gossip will fail for the 5th person, he will then see the reflected changes in gossip when he opens the window BUT the other 4 players will encounter the same problem you describe. It is not protecting people in the way you describe really. If anything, the current protection in CMSG_GOSSIP_SELECT_OPTION is a hack too then.

If it's possible to close every opened gossip for an NPC when it changes entry that could work, but I don't think the currently gossiping players are tracked, are they?

jackpoz:
forgot to remove the empty line changes, please try to keep those away from PRs

PR #22472, opened 2018-09-21
Scripts/BlackrockDepths: Open the gates to Magmus only once both Shad…

ghost:
…owforge Braziers are lit

**Changes proposed:**

-  The doors at BRD's Lyceum leading to Magmus should only open once both Shadowforge Braziers are lit. Currently each brazier opens its corresponding door, which is wrong.

This is how it still works on retail up to this day.

- [X ] 3.3.5
- [X ] master

Builds, works in-game, etc.


ccrs:
delay this a bit, would be better to have a full instancescript update before 👍

Issue #22435, opened 2018-09-14
Incorrect calculation of the first "fake" Catmull-Rom spline point


xvwyh:
This is how the first point is currently calculated on the server (present in the code twice, just a heads up):
`Spline.cpp`
```cpp
points[0] = controls[0].lerp(controls[1], -1);
```
This is incorrent. Here's the fixed code that matches what the client is doing:
```cpp
points[0] = controls[0] - G3D::Vector3 { cos(initialOrientation), sin(initialOrientation), 0.0f };
```
(of course this requires initialOrientation from MoveSplineInitArgs to be passed to spline initialization)

This fixes slight position desync for flying units. My players were constantly nagging me about Val'kyr Shadowguards on The Lich King encounter always teleporting forward a bit whenever they were stunned, this issue is the reason why that was happening: due to larger distance between the first "fake" and "real" points, the "easing-in" of unit's velocity happens much faster on the server, thus making the unit travel through the spline faster than it would visually appear on the client.

I won't be providing the proof for this, because the decompiled code looks more confusing than helpful, just trust me on this one. It's still in CGUnit_C::OnMonsterMovePacket if you want to find it yourself, look for sin/cos calls.

And just to please @TrinityCoreBot: 0f3156d324ff8134171bada01b40f5c23f5c43cc

Kittnz:
Offtopic: just a question. Is there a reason why you won't make a PR for the issues you are making?

xvwyh:
I don't have TC code neither pulled nor compiled (and currently don't fancy jumping through the hoops to get boost or VS2017 to work), so I can't test any of these changes, and I would rather not commit anything I can't test. I am, however, working with a codebase that branched off of TC many years ago, so all the issues I post are still relevant. I do check everything I post on whether or not it has already been fixed here.

That, and I initially wanted to post all this on the forums like a year ago or so, but some bugs happened that prevented me from doing that (some crap with permissions, I did ask on TC's IRC but nobody volunteered to help) and I lost the motivation.

Warpten:
If you're getting triggered by SSE instructions it's a lot easier to see when you ignore hexrays. Also hexrays sometimes misses branches for no reasons (around SSE instructions)

xvwyh:
Oh well, here's more or less what you should look for: https://i.imgur.com/vudpI4A.png

Aokromes:
5e8b934

PR #22582, opened 2018-10-06
Core/Movement: Properly calculate the first Catmull-Rom spline point

DanVS:
**Changes proposed:**
Implement #22435.
Credit @xvwyh.

**Target branch(es):** 3.3.5

**Tests performed:** _Somewhat_ tested in-game.

### DIFF
--- a/src/server/game/Movement/Spline/Spline.cpp
+++ b/src/server/game/Movement/Spline/Spline.cpp
@@ -255,14 +255,14 @@ void SplineBase::InitCatmullRom(Vector3 const* controls, index_type count, index
         if (cyclic_point == 0)
             points[0] = controls[count-1];
         else
-            points[0] = controls[0] - G3D::Vector3{ cos(initialOrientation), sin(initialOrientation), 0.0f };
+            points[0] = controls[0] - G3D::Vector3{ std::cos(initialOrientation), std::sin(initialOrientation), 0.0f };
 
         points[high_index+1] = controls[cyclic_point];
         points[high_index+2] = controls[cyclic_point+1];
     }
     else
     {
-        points[0] = controls[0] - G3D::Vector3{ cos(initialOrientation), sin(initialOrientation), 0.0f };
+        points[0] = controls[0] - G3D::Vector3{ std::cos(initialOrientation), std::sin(initialOrientation), 0.0f };
         points[high_index+1] = controls[count-1];
     }
 
--- a/src/server/game/Movement/Spline/Spline.cpp
+++ b/src/server/game/Movement/Spline/Spline.cpp
@@ -199,18 +199,20 @@ float SplineBase::SegLengthBezier3(index_type index) const
     return length;
 }
 
-void SplineBase::init_spline(const Vector3 * controls, index_type count, EvaluationMode m)
+void SplineBase::init_spline(const Vector3 * controls, index_type count, EvaluationMode m, float orientation)
 {
     m_mode = m;
     cyclic = false;
+    initialOrientation = orientation;
 
     (this->*initializers[m_mode])(controls, count, 0);
 }
 
-void SplineBase::init_cyclic_spline(const Vector3 * controls, index_type count, EvaluationMode m, index_type cyclic_point)
+void SplineBase::init_cyclic_spline(const Vector3 * controls, index_type count, EvaluationMode m, index_type cyclic_point, float orientation)
 {
     m_mode = m;
     cyclic = true;
+    initialOrientation = orientation;
 
     (this->*initializers[m_mode])(controls, count, cyclic_point);
 }
@@ -253,14 +255,14 @@ void SplineBase::InitCatmullRom(Vector3 const* controls, index_type count, index
         if (cyclic_point == 0)
             points[0] = controls[count-1];
         else
-            points[0] = controls[0].lerp(controls[1], -1);
+            points[0] = controls[0] - G3D::Vector3{ cos(initialOrientation), sin(initialOrientation), 0.0f };
 
         points[high_index+1] = controls[cyclic_point];
         points[high_index+2] = controls[cyclic_point+1];
     }
     else
     {
-        points[0] = controls[0].lerp(controls[1], -1);
+        points[0] = controls[0] - G3D::Vector3{ cos(initialOrientation), sin(initialOrientation), 0.0f };
         points[high_index+1] = controls[count-1];
     }
 
--- a/src/server/game/Movement/Spline/MoveSpline.cpp
+++ b/src/server/game/Movement/Spline/MoveSpline.cpp
@@ -126,10 +126,10 @@ void MoveSpline::init_spline(MoveSplineInitArgs const& args)
         // MoveSplineFlag::Enter_Cycle support dropped
         //if (splineflags & SPLINEFLAG_ENTER_CYCLE)
         //cyclic_point = 1;   // shouldn't be modified, came from client
-        spline.init_cyclic_spline(&args.path[0], args.path.size(), modes[args.flags.isSmooth()], cyclic_point);
+        spline.init_cyclic_spline(&args.path[0], args.path.size(), modes[args.flags.isSmooth()], cyclic_point, args.initialOrientation);
     }
     else
-        spline.init_spline(&args.path[0], args.path.size(), modes[args.flags.isSmooth()]);
+        spline.init_spline(&args.path[0], args.path.size(), modes[args.flags.isSmooth()], args.initialOrientation);
 
     // init spline timestamps
     if (splineflags.falling)
--- a/src/server/game/Movement/Spline/Spline.h
+++ b/src/server/game/Movement/Spline/Spline.h
@@ -50,6 +50,7 @@ protected:
 
     uint8 m_mode;
     bool cyclic;
+    float initialOrientation;
 
     enum{
         // could be modified, affects segment length evaluation precision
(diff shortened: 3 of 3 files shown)


Kittnz:
Any tests preformed?

DanVS:
Well, seems to be working, but right now I am unable to perform tests with multiple accounts, so this probably requires attention.

DanVS:
Works on flying mobs from Icecrown and The Storm Peaks.

src/server/game/Movement/Spline/Spline.cpp
```diff
@@ -253,14 +255,14 @@ void SplineBase::InitCatmullRom(Vector3 const* controls, index_type count, index
         if (cyclic_point == 0)
             points[0] = controls[count-1];
         else
-            points[0] = controls[0].lerp(controls[1], -1);
+            points[0] = controls[0] - G3D::Vector3{ cos(initialOrientation), sin(initialOrientation), 0.0f };
```
Shauren: use std::sin and std::cos (also in the other line)

Shauren (changes requested):

src/server/game/Movement/Spline/Spline.cpp
DanVS: Fixed. Thanks!

Shauren (approved):

Ovahlord:
@xvwyh summoning you here for a semi related discussion. I'm currently digging into splines again and I have noticed something while prototyping smooth waypoint transitions and that is that creatures tend to have some small setback hickups when stopping or launching new splines while another spline is going on. 

So far I've figured out that's simply because the clientside position is ahead of the serverside one which makes sense considering there are things like latency so how does Blizz actually prevent these hickups? Do they have some 'lookahead value' internally which they use to offset splines so they compute a position that's somewhat more sync with the client?

Or do they send splines with a slightly increased duration instead so the unit moves a tad slower to mitigate that?

xvwyh:
What do you mean by "setback hickups"? Also what spline type are we talking about: linear or catmull-rom?

I've dug into the client and found the code that makes the creature "backtrack" to the first point of a linear spline if creature's clientside position differs too much from the first path point. E.g. if we have the following situation:
```
<----C----S-----




          X
```
(where C is clientside's position of a creature that's currently moving left and is ahead of the server for whatever reason, S is its serverside's position and X is the destination of a new movement)
You might think that the client will plot the new spline like this, matching the path that the server sent to the client:
```
     C    S     
          |
          |
          |
          V
          X
```
but it won't, because that will cause the creature to teleport abruptly to it's serverside's starting position.

So then you might think that the client will plot the new spline like this:
```
     C    S     
      \
       \
        \
         V
          X
```
but it won't, because that will cause severe discrepancy in the creature's positions throughout that path from that of the server (the server will send the creature straight downwards in this situation)

What happens instead is the creature's current clientside position is added to the front of the path:
```
     C--->S     
          |
          |
          |
          V
          X
```
so the creature will backtrack to where it's supposed to be from server's PoV, and then proceed with the intended path. This happens without changing the duration of the spline! So yes, it will be sped up, and it won't precisely match server's positions throughout the movement, but at least the path taken will be roughly the same, and the creature will arrive in its destination at the same time as on the server.

The pseudocode for it is roughly
```cpp
if (GetDistanceBetween(GetClientsidePosition(), path.front()) >= 0.027777777)
    path.push_front(GetClientsidePosition());
else
    path.clear(); // This seems weird? Keep reading, I made a mistake, but I'm not rewriting this wall of text
```
image

(`points` here is a pointer pointing to the SECOND element of an array of 3d vectors, the FIRST element is initially unused and reserved specifically for this case: it gets filled by the unit's current clientside position, count of points is incremented, and then the pointer to the first element is returned, allowing it to be used in the spline)

But. There's always a but. This works only when the path consists of just a start and destination points. No other intermediate points.

When the path has more points, some **very** convoluted code is executed, that searches for some specific point in that path. There's a similar code in alpha client, and it's way more straightforward in there - it simply searches for the point that's closest to unit's current position. The code in WotLK client is much more complicated, and it seems to even have some loop unrolling in it, so it's a nightmare to read, and I couldn't figure out what it was doing.  But at the end of it, depending on the index of the point that was found, one of several cases can happen:
a) Same thing happens as written above: current clientside position gets added to the front of the path (it checks if 2D (!) distance between the current position and the found point is >= 1)
b) The whole path is replaced with just 2 points: current clientside position and final destination (it checks if 3D distance between the current position and the found point is >= 0.027777777)
c) The whole path is erased and no movement happens
d) Even more convoluted code is executed that searches the path for yet another point, checking sometimes 2D sometimes 3D distance and then the path gets cut at that point and either the former or latter part of the path is discarded and jesus fucking christ I'm done, I'm out, I don't want to see this ever again.

Actually, no, I lied, the case I described with only start and destination points should actually be the case when there is NO movement at all, or the distance between start and dest points is < 0.027777777, in which case no movement happens either. But, yeah, if the clientside position is too far from it - the unit WILL move back to where it should be based on the data from the packet, and now that `path.clear()` make sense. But I'm too lazy to rewrite everything I wrote already, because I doubt it will help you anyway.

xvwyh:
And what do you mean by "smooth waypoint transitions"? Did you mean simply that you want to prevent the unit from stopping in each point for a short time awaiting a new movement packet from the server? In that case just send the whole path in one single movement packet and you're done...

Ovahlord:
Well, as far as I'm concerned, it happens for all splines. Imagine a creature chases you and you you kill that creature during chasing, there is a very high chance that the unit will turn arround and look towards its opposite direction because that's where the server says where the creature is suposed to be.

I went ahead and did some packet research and the first thing that I noticed is that the spline duration in retail packets slightly differs from what you get when doing manual calculations (something between 1 - 80ms differences).

However, I did further experiments by computing positions ahead of time which solves these turnarrounds but causes ugly snapping. At first I just added the mstime between the current time and the last gametime update cycle which makes the spline positions pin point accurate in most scenarios but then again there is this latency factor. The longer the packet takes to arrive at the client, the higher the chance that this turnarround happens again but knowing retails laggy nature there must be some trick to avoid that

xvwyh:
The thing depicted on the video is creatures stopping on death, that's not exactly the same as going from spline to spline. Stop packets have heir own special handling in the code, and way less nightmarish:
image
(case 1:)
Checks whether the distance between the current clientside position and the position from the spline is >= cvar "pathDistTol" (1 yard by default). If it's smaller than the cvar, it simply teleports the unit to the position. If it's larger than the cvar - it proceeds onwards with creating a linear spline that causes the creature to walk to the position and thus turn around.

But here's the thing: I looked at a few sniffs, and whenever I saw a creature dying in them - it was sending a movement packet with facing type set to 0, aka it's NOT a stop packet, but a simple movement packet with a completely empty path, so only the starting position is present in the packet. Maybe it behaves differently, give it a try. Basically replace
https://github.com/TrinityCore/TrinityCore/blob/cc8287927b9c9f0bf20143972fd4d3c338bff688/src/server/game/Movement/Spline/MoveSplineInit.cpp#L186
with
https://github.com/TrinityCore/TrinityCore/blob/cc8287927b9c9f0bf20143972fd4d3c338bff688/src/server/game/Movement/Spline/MoveSplineInit.cpp#L140
and maybe make some more adjustments if it won't work out of the box, but I think it should.

Ovahlord:
Yeah, I figured out already that dying creatures use a spline with facing 0. But their path is not empty. It's actually a path with a single point that's the exact same point as their starting point:

> ServerToClient: SMSG_ON_MONSTER_MOVE (0x6E17) Length: 50 ConnIdx: 2 Time: 08/01/2012 17:34:40.140 Number: 26112
> MoverGUID: Full: 0xF130C1840004707C Type: Creature Entry: 49540 (Stormwind Rat) Low: 290940
> VehicleExitVoluntary: 0
> Position: X: -8720.3125 Y: 456.77148 Z: 97.642105  << initial starting position
> (MovementMonsterSpline) Id: 13147604
> (MovementMonsterSpline) (MovementSpline) Face: 0 (Normal)
> (MovementMonsterSpline) (MovementSpline) Flags: 0 (None)
> (MovementMonsterSpline) (MovementSpline) MoveTime: 0
> (MovementMonsterSpline) (MovementSpline) PointsCount: 1
> (MovementMonsterSpline) (MovementSpline) Points: X: -8720.3125 Y: 456.77148 Z: 97.642105  << spline path destination point

Anyways, I've already replicated these packets with the exact same data but these turnarrounds are still the same. To be precise, they got even worse now and are much more consistent. Now pretty much every single dying creatures turns arround so I'm a bit at a loss here.

Little proof that I send it the same way as retail:

> ServerToClient: SMSG_ON_MONSTER_MOVE (0x6E17) Length: 48 ConnIdx: 0 EP: 127.0.0.1:60143 Time: 11/19/2020 02:32:30.138 Number: 396
> MoverGUID: Full: 0xF13001DB0000007A Type: Creature Entry: 475 (Kobold Tunneler) Low: 122
> VehicleExitVoluntary: 0
> Position: X: -9824.555 Y: 192.54974 Z: 12.730363
> (MovementMonsterSpline) Id: 256
> (MovementMonsterSpline) (MovementSpline) Face: 0 (Normal)
> (MovementMonsterSpline) (MovementSpline) Flags: 0 (None)
> (MovementMonsterSpline) (MovementSpline) MoveTime: 0
> (MovementMonsterSpline) (MovementSpline) PointsCount: 1
> (MovementMonsterSpline) (MovementSpline) Points: X: -9824.555 Y: 192.54974 Z: 12.730363

----
Edit: I solved the mystery. It's SMSG_ATTACK_STOP. It has a 3rd field that should be set to 1 when the attacker has died.

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Server/Packets/CombatPackets.cpp#L40
there are two mistakes that need to be fixed. The first one is that NowDead has to be 1 when the attacker has died, not the victim. (sniff confirmed). This field preserves the facing angle of the unit when set to 1. The 2nd one is that AttackStop is being sent before the unit's deathState gets updates so this field never gets set to 1.

xvwyh:
Huh, in my sniffs there are no points, just the initial position. Must've changed between expansions, the oldest ones I checked were from 21463.

Hilarious that movement-related stuff is controlled from SMSG_ATTACK_STOP.

Ovahlord:
The one I used was from 15595 so that's funny, ye

PR #22654, opened 2018-10-15
Fix Compilation issues

TiTaKaZ:
**Changes proposed:**

-  Updates the strings from DBCStores.cpp (somewhere at line 380) to char*s
-  
-  

- [*] 3.3.5

- [*] Compiles normally
- [/] Does not work


TiTaKaZ:
Update: does not work :/

Issue #20493, opened 2017-09-30
Trainer greetings aren't implemented.


Wyreth:
**Description:**

Yet another obscure gossip text variation from Blizzard! Aren't you happy?

On top of the trainer window for profession and class trainers, there's a line of text.

This is always the same for class trainers, but different for profession trainers depending on what they train (and some of them have unique ones, to boot).

[A video example here](https://youtu.be/UxxcAh8iqrM?t=25).

Here are the ones I could find. Once the list is complete I'll make a pull request to implement it core-side.

```sql
Cooking trainers: "Can I teach you how to turn the meat you find on beasts into a feast?"
Kylene (19186) (Cooking trainer): "There's more than one way to cook a fish."
Fishing trainers: "I can teach you how to use a fishing pole to catch fish."
Marcia Chase (28742) (Fishing trainer): "I can teach you the finer points of fishing."
First Aid trainers: "Here, let me show you how to bind those wounds...."
Enchanting trainers: "Enchanting is the art of improving existing items through magic. " -- trailing space intended
Jewelcrafting trainers: "Greetings!  Can I teach you how to cut precious gems and craft jewelry?"
Mining trainers: "You have not lived until you have dug deep into the earth."
Skinning trainers: "It requires a steady hand to remove the leather from a slain beast."
Engineering trainers: "Engineering is very simple once you grasp the basics."
Leatherworking trainers: "Greetings!  Can I teach you how to turn beast hides into armor?"
Inscription trainers: "Would you like to learn the intricacies of inscription?"
Blacksmithing trainers: "Care to learn how to turn the ore that you find into weapons and metal armor?"
Alchemy trainers: "With alchemy you can turn found herbs into healing and other types of potions."
Linzy Blackbolt (28703) (Alchemy trainer): "The herbs of Northrend can be brewed into powerful potions."
Herbalism trainers: "Searching for herbs requires both knowledge and instinct."
Tailoring trainers: "Greetings!  Can I teach you how to turn found cloth into cloth armor?"
Riding trainers: "Hello!  Can I teach you something?"
```

What's missing:
- [x] Herbalism trainers
- [x] Alchemy trainers
- [x] Inscription trainers
- [x] Tailoring trainers
- [x] Leatherworking trainers
- [x] Blacksmithing trainers

Sources: [Inscription](https://youtu.be/8sWDEZ5WSeA?t=21), [Blacksmithing](https://youtu.be/znhkYk3qAZA?t=31).

[...]

PR #22661, opened 2018-10-17
Core/Trainer: Port trainer changes from master

Ghaster:
**Changes proposed:**

-  Port https://github.com/TrinityCore/TrinityCore/commit/0cd5179726f91cc163910dd0806d622fe86d9a8c from master to 3.3.5

- [x] 3.3.5
- [ ] master

**Issues addressed:** Updates #20493 (Greetings need to added to 'trainer')

Builds, works ingame, etc etc.

- [x] Basically everything database related (porting data from 'npc_trainer' to 'trainer_spell', update commands, drop removed fields etc.)
- [x] Generate ReqAbility* fields from SpellID

For a quick recap of current changes:
- Table 'npc_trainer' is gone, replaced with 'trainer_spell'
- Creature and trainers are linked using creature_default_trainer. As the trainerId field didn't exist in WotLK, we just reuse the current values for template trainers and use the same creatureId as trainerId for the others.
- 'trainer_type', 'trainer_spell', 'trainer_class' and 'trainer_race' have been moved from 'creature_template' to 'trainer'. The last 3 columns have been condensed into a simple column called 'Requirement'. If e.g. the type has been set to 1, the 'Requirement' field can be used to select which race this trainer is allowed to.
- Note that the 'trainer' table is linked to the trainerId and not the creature id.


src/server/game/Server/WorldSession.h
```diff
@@ -74,6 +74,11 @@ class RBACData;
 
 namespace WorldPackets
 {
+	namespace NPC
+	{
+		class Hello;
+		class TrainerBuySpell;
+	}
```
Shauren: tabs

src/server/game/Entities/Creature/CreatureData.h
```diff
@@ -150,7 +150,6 @@ struct TC_GAME_API CreatureTemplate
     uint32  unit_flags2;                                    // enum UnitFlags2 mask values
     uint32  dynamicflags;
     CreatureFamily  family;                                 // enum CreatureFamily values (optional)
-    uint32  trainer_type;
     uint32  trainer_spell;
```
Shauren: You should move remaining trainer_ fields into trainer table as well

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1206,10 +1197,10 @@ bool Creature::isCanInteractWithBattleMaster(Player* player, bool msg) const
     return true;
 }
 
-bool Creature::isCanTrainingAndResetTalentsOf(Player* player) const
+bool Creature::CanResetTalents(Player* player) const
 {
     return player->getLevel() >= 10
-        && GetCreatureTemplate()->trainer_type == TRAINER_TYPE_CLASS
+        && sObjectMgr->GetTrainer(GetEntry())->GetTrainerType() == Trainer::Type::Class
```
Shauren: Is it possible to reach this line of code if the creature is not a trainer?

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -8702,122 +8689,139 @@ void ObjectMgr::LoadMailLevelRewards()
...
+    // For reload case
+    _trainers.clear();
 
-    if (GetTalentSpellCost(SpellID))
+    std::unordered_map<int32, std::vector<Trainer::Spell>> spellsByTrainer;
+    //if (QueryResult trainerSpellsResult = WorldDatabase.Query("SELECT ID, SpellID, MoneyCost, ReqSkillLine, ReqSkillRank, ReqAbility1, ReqAbility2, ReqAbility3, ReqLevel FROM trainer_spell"))
+    //if (QueryResult trainerSpellsResult = WorldDatabase.Query("SELECT b.ID, a.SpellID, a.MoneyCost, a.ReqSkillLine, a.ReqSkillRank, a.ReqLevel FROM trainer_spell AS a INNER JOIN trainer_spell AS b ON a.ID = -(b.SpellID) UNION SELECT * FROM trainer_spell WHERE SpellID > 0"))
+    if (QueryResult trainerSpellsResult = WorldDatabase.Query("SELECT b.ID, a.SpellID, a.MoneyCost, a.ReqSkillLine, a.ReqSkillRank, a.ReqAbility1, a.ReqAbility2, a.ReqAbility3, a.ReqLevel, a.VerifiedBuild FROM trainer_spell AS a "
+        "INNER JOIN trainer_spell AS b ON a.ID = -(b.SpellID) "
+        "UNION SELECT * FROM trainer_spell WHERE SpellID > 0"))
```
Shauren: Don't do `SELECT *`

Shauren: Also: wouldn't it be better to first migrate all trainers to custom ids lists (basically, make everything refer to these negative spellids) and then do the same as master branch (id, not creatureId, linked by creature_default_trainer)? That would allow WPP to keep the same output code for both versions

src/server/game/Server/Packets/NPCPackets.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+WorldPacket const* WorldPackets::NPC::TrainerList::Write()
+{
+    _worldPacket << TrainerGUID;
+    _worldPacket << TrainerType;
+
+    _worldPacket << int32(Spells.size());
+    for (TrainerListSpell const& spell : Spells)
+    {
+        _worldPacket << spell.SpellID;
```
Shauren: Keep explicit typecasts for everything thats not a guid/string

Shauren (approved):

src/server/game/Globals/ObjectMgr.cpp
Ghaster: But how would WPP then know what to export the Id field as?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -14074,8 +14074,15 @@ void Player::PrepareGossipMenu(WorldObject* source, uint32 menuId /*= 0*/, bool
                 case GOSSIP_OPTION_QUESTGIVER:
                     canTalk = false;
                     break;
-                case GOSSIP_OPTION_GOSSIP:
                 case GOSSIP_OPTION_TRAINER:
+                    if (getClass() != sObjectMgr->GetTrainer(creature->GetEntry())->GetTrainerRequirement() && sObjectMgr->GetTrainer(creature->GetEntry())->GetTrainerType() == Trainer::Type::Class)
```
Ghaster: Are we only denying the wrong class here? Or should wrong race/missing spell also be checked here?

src/server/game/Globals/ObjectMgr.cpp
Shauren: WPP can autogenerate it with our standard `@ID` placeholder

src/server/game/Entities/Player/Player.cpp
Shauren: no idea. armor/weaponsmiting was removed on retail, only profession trainers are left

illfated: Well, race needs checking for riding trainers, so you have got a point there.

illfated:
BTW, are you familiar with travis-ci and its error messages, or do you want error snippets pasted into the PR?

Ghaster:
All existing trainer data has now been ported to the new system, so this PR should be more or less "ready".

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1206,11 +1197,13 @@ bool Creature::isCanInteractWithBattleMaster(Player* player, bool msg) const
...
+bool Creature::CanResetTalents(Player* player) const
 {
-    return player->getLevel() >= 10
-        && GetCreatureTemplate()->trainer_type == TRAINER_TYPE_CLASS
-        && player->getClass() == GetCreatureTemplate()->trainer_class;
+    Trainer::Trainer const* trainer = sObjectMgr->GetTrainer(GetEntry());
+    if (!trainer)
+        return false;
+
+    return player->getLevel() >= 10 && trainer->IsTrainerValidForPlayer(player);
```
kelno: I think this will return true for every professions trainers as well
`&& trainer->GetTrainerType() == Trainer::Type::Class;`

Ghaster:
> 
> Just out of personal curiosity (maybe unnecessary): I see some of the inserts in the big SQL are somewhat unsorted, what is your method of making sure the data is kept correct? Is it simply unsorted output extracted from the existing TDB?

The trainer data has been converted directly from the existing data in the DB, so it should work 100% the same as before.
However, as far as i can tell, the way the core handles ReqAbility is pretty bogus? The only "requirements" that gets added, are the profession requirements in ObjectMgr. That means that right now, someone can simply train a rank 9 spell, and get all the previous spell ranks for free. This wasn't an issue before, due to the trainer only being able to teach one spell per train.

src/server/game/Entities/Creature/Trainer.cpp
```diff
@@ -0,0 +1,243 @@
...
+        if (player->getLevel() < trainerSpell->ReqLevel)
+            return SpellState::Unavailable;
+
+        // check ranks
+        bool hasLearnSpellEffect = false;
+        bool knowsAllLearnedSpells = true;
+        for (uint8 i = 0; i < MAX_SPELL_EFFECTS; ++i)
+        {
+            SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(trainerSpell->SpellId);
+            if (!spellInfo || !spellInfo->Effects[i].Effect == SPELL_EFFECT_LEARN_SPELL)
```
Shauren: Incorrect condition here - `!spellInfo->Effects[i].Effect == SPELL_EFFECT_LEARN_SPELL`, this should be `spellInfo->Effects[i].Effect != SPELL_EFFECT_LEARN_SPELL`

Shauren (changes requested):

Ghaster:
> No, this is supposed to take care of it
> https://github.com/TrinityCore/TrinityCore/pull/22661/files#diff-b21e8837ee77c02aab389930793ca21eR154

In the code it's working fine yes, but we are missing the ReqAbility values in the DB. These were generated on the fly beforehand when loading the trainers, but only for professions.

Shauren (approved):

Ghaster:
Thx @Shauren for fixing the remaining stuff. Two small things I noticed:

This needs to be updated:
https://github.com/TrinityCore/TrinityCore/blob/baebb2d602ca2f10159ea0039ec3a6d03155c3bd/sql/base/auth_database.sql#L1777

This can be safely removed:
https://github.com/TrinityCore/TrinityCore/blob/baebb2d602ca2f10159ea0039ec3a6d03155c3bd/src/server/game/Globals/ObjectMgr.h#L528

PR #22699, opened 2018-10-21
Core/Entities Rework Area/Zone storage

Traesh:
**Changes proposed:**

- Rework how the current area & zone are stored
- This is an optimization because most "GetAreaId/GetZoneId" which were calculating area based on position now just return id from current stored Area on the entity, saving CPU time
- Use an "Area" object rather than simply store areaId & zoneId
- Futureproofer because of the mutliple sub-zones (an area with a parent which have a parent)
- Will allow to store script by zone
- Will allow some "by zone" logic (example : automatically generate pools for gathering nodes, already done on Ashamane i'll PR it here if this one is accepted)

**Target branch(es):** master

- [X] master

**Tests performed:**
Build, Tested ingame by myself (on simple area/zones like elwynn, and "parent with parent" like northshire) & implentend on Ashamane live server for player tests

Issues : None so far

Todo : There are probably methods where we both send areaId & zoneId, now only sending the Area object would be enough


src/server/game/World/Area.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if (!areaId)
+        return nullptr;
+
+    auto itr = m_areas.find(areaId);
+    if (itr != m_areas.end())
+        return itr->second;
+
+    if (AreaTableEntry const* areaEntry = sAreaTableStore.LookupEntry(areaId))
+    {
+        Area* area = new Area(areaEntry);
```
Naios: The area never gets deleted, maybe use `std::unique_ptr<Area>` as container type

Shauren: I already gave my feedback on IRC about this, its not going to be merged with current design

PR #22723, opened 2018-10-26
Add rotation to .gob info

### LINKED ISSUE #22714: Print gameobject rotation
**Description:**

There is no way to view the full rotation of gameobjects with a command. You can only view orientation (yaw), but not pitch or roll even if you can set them with `.gobject turn <guid> <yaw> <pitch> <roll>`.

Suggestions

1. `.gps <link>` prints current rotation as a part of the other info
2. `.gobject info guid <link/guid>` prints current rotation as a part of the other info
3. `.gobject turn <link/guid>` prints current rotation instead of setting one when no rotation arguments given

**Current behaviour:**

There is no way to view the full rotation of gameobjects.

**Expected behaviour:**

There would be a command to view the current rotation (yaw, pitch and roll) of a gameobject.

**Branch(es):**

3.3.5, master

**TC rev. hash/commit:** 

0a0312b705ecb1a318fd3abe545b571a3a349a9c

Rochet2:
**Changes proposed:**

- Add rotation (yaw, pitch, roll) to .gob info command
- Replace LANG_SPAWNINFO_DISTANCEFROMPLAYER with LANG_SPAWNINFO_ROTATION
- Fix description of .distance to describe the command arguments (the more you know..)
- Fix .gob info guid targetting nearest gameobject instead of finding with given guid
- Reorder the info data slightly
- Print location and rotation data regardless of object found in world
- Add error messages for when object is not found by guid or entry
- Remove showing info of nearest object with given entry in 30 yards when command invoked with entry only

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**
Closes #22714 

- Builds
- Tested ingame with `.gob info <entry>` `.gob info guid <guid>` in scenarios where object does not exist, exists but is not spawned, exists and is spawned nearby.


jackpoz:
How come the 30 yards part has been removed ?

illfated:
Good point. `.gameobj near` (or `.gob near`) is a more intuitive command to look for nearby objects.

Warpten:
I think the intent was to have data for the nearest gob in case you forgot to specify a guid (which means you just had to negate the if)

Shauren:
Except "nearest gob" is almost always useless thanks to doodad/collision/etc gobs scattered all over the place

jackpoz:
thank you for the PR :)

PR #22728, opened 2018-10-28
Spell/Script: Curse of agony

PolarCookie:
**Changes proposed:**
Glyph of Curse of Agony was not working correctly.
Using it would give a sequence of 4x1/24, 5x1/12, 5x1/8 ticks, while it should have been, 4x1/24, 4x1/12, 4x1/8, 2x1/6.

https://www.wowhead.com/item=42456

Moved both Agony and Drain Soul to spellscripts, instead of hacks.

**Target branch(es):**
3.3.5

**Tests performed:**
Builds and tested ingame.


src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -495,8 +544,26 @@ class spell_warl_drain_soul : public SpellScriptLoader
...
 
+            void HandleTick(AuraEffect const* aurEff)
+            {
+                Unit* caster = GetCaster();
+                Unit* target = GetTarget();
+                if (caster && caster->GetTypeId() == TYPEID_PLAYER && caster->ToPlayer()->isHonorOrXPTarget(target))
+                {
+                    if (roll_chance_i(20))
+                    {
+                        caster->CastSpell(caster, SPELL_CREATE_SOULSHARD, this);
```
Keader: caster->CastSpell(caster, SPELL_CREATE_SOULSHARD, aurEff);

src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -495,8 +544,26 @@ class spell_warl_drain_soul : public SpellScriptLoader
...
+                        // Glyph of Drain Soul - chance to create an additional Soul Shard
+                        if (AuraEffect* aur = caster->GetAuraEffect(SPELL_GLYPH_OF_DRAIN_SOUL_AURA, 0))
```
Keader: EFFECT_0, instead 0

sql/updates/world/3.3.5/9999_99_99_99.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM `spell_script_names` WHERE `spell_id` = -980;
```
Keader: always delete by ScriptName xD

src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -460,6 +502,13 @@ class spell_warl_demonic_empowerment : public SpellScriptLoader
         }
 };
 
+enum DrainSoul
+{
+    SPELL_CREATE_SOULSHARD         = 43836,
+    SPELL_GLYPH_OF_DRAIN_SOUL_AURA = 58070,
+    SPELL_GLYPH_OF_DRAIN_SOUL_PROC = 58068
```
Keader: add those in validate

sql/updates/world/3.3.5/9999_99_99_99.sql
PolarCookie: Still think it's a terrible idea, and still haven't found this design document you guys never link.

Keader: 1 spell can have multiple scripts.

PolarCookie: and many spells can share the same script, which is more common.

Keader: When you delete by spell_id, you will delete all scripts assigned to it... and sometimes it's not what you want...
We delete By ScriptName, because if you are adding that Script, you will always deleted only script that you changing/adding

Jildor: I prefer delete by ScriptName and spell_id 😄

PolarCookie: There are a total of 17 spell ids with multiple ScriptNames, compared to 253 spell ids which share ScriptNames.
And of those 17 I see "clever" solutions which are unnecessary and come from developer laziness for the most part (hello, those paladin seals).
Those 17 are special exception cases, do not let them dictate a very unhygienic practice.

PolarCookie: If you want to keep your sanity in databases you refer to objects by their unique explicit identifier, which in this case is the spell id given by Blizzard, not the arbitrary and mutable name given by TrinityCore.

Keader: Well maybe @Shauren can explain better.

PolarCookie:
I'm fine with it.

jackpoz:
Thank you for the PR :)

src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -92,6 +95,48 @@ enum WarlockSpellIcons
...
+        }
+    }
+
+    void Register() override
+    {
+        AfterEffectApply += AuraEffectApplyFn(spell_warl_curse_of_agony::ApplyEffect, EFFECT_0, SPELL_AURA_PERIODIC_DAMAGE, AURA_EFFECT_HANDLE_REAL_OR_REAPPLY_MASK);
+        OnEffectUpdatePeriodic += AuraEffectUpdatePeriodicFn(spell_warl_curse_of_agony::HandleEffectPeriodicUpdate, EFFECT_0, SPELL_AURA_PERIODIC_DAMAGE);
+    }
+private:
+    uint32 _tick_amount;
```
jackpoz: just as a note, always initialize all fields either in the ctor or in the field declaration itself to avoid static analysis issues

PR #22734, opened 2018-10-31
[WIP] Core/Misc Handle CMSG_SET_WAR_MODE

Traesh:
**Changes proposed:**

- Handle the new "War Mode" system

**Target branch(es):**

- [X] bfa

**Issues addressed:**
None

**Tests performed:**
Build, not tested ingame yet

### DIFF
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -404,7 +404,7 @@ enum PlayerFlags
     PLAYER_FLAGS_CONTESTED_PVP          = 0x00000100,       // Player has been involved in a PvP combat and will be attacked by contested guards
     PLAYER_FLAGS_IN_PVP                 = 0x00000200,
     PLAYER_FLAGS_HIDE_HELM              = 0x00000400,
-    PLAYER_FLAGS_HIDE_CLOAK             = 0x00000800,
+    PLAYER_FLAGS_WAR_MODE               = 0x00000800,
     PLAYER_FLAGS_PLAYED_LONG_TIME       = 0x00001000,       // played long time
     PLAYER_FLAGS_PLAYED_TOO_LONG        = 0x00002000,       // played too long time
     PLAYER_FLAGS_IS_OUT_OF_BOUNDS       = 0x00004000,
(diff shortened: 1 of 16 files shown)


src/server/game/Entities/Player/Player.h
```diff
@@ -404,7 +404,7 @@ enum PlayerFlags
     PLAYER_FLAGS_CONTESTED_PVP          = 0x00000100,       // Player has been involved in a PvP combat and will be attacked by contested guards
     PLAYER_FLAGS_IN_PVP                 = 0x00000200,
     PLAYER_FLAGS_HIDE_HELM              = 0x00000400,
-    PLAYER_FLAGS_HIDE_CLOAK             = 0x00000800,
+    PLAYER_FLAGS_WAR_MODE               = 0x00000800,
```
Shauren: so, which one is it

Shauren:
I need to ask, what is the point of just handling this opcode? War mode implies a lot of other things on retail, such as being on completely separate servers and never being able to see nonwarmode players outside instances

src/server/game/Entities/Player/Player.h
Traesh: In my sniff it seem that 0x800 is re-used for warmode (hide_cloak & hide_helm are unused since this is now part of the transmog system)

Traesh:
iirc separe server is the only other thing implied on retail, and of course we won't be able to handle that correctly any time soon.

This still allow for a player to choose his game style (with or without savage PVP), it activate the pvp talents, and the war mode give 10% experience boost & world quest currency gain for those who choose to play PVP-style.

The fact that a PVP player will be able to see a PVE player isn't a big deal imho (but maybe i should add a temp check so PVE player can't just attack PVP player whenever they want)

ReyDonovan:
_player->AddAura(269083) -- correct spell test in game

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1157,3 +1157,16 @@ void WorldSession::HandleCloseInteraction(WorldPackets::Misc::CloseInteraction&
...
+
+void WorldSession::HandleSetWarModeOpcode(WorldPackets::Misc::SetWarMode& warMode)
+{
+    uint32 const warModeSpellId = 282559; // Enlisted
+
+    if (_player->GetZoneId() != ZONE_STORMWIND_CITY && _player->GetZoneId() != ZONE_ORGRIMMAR)
+        return;
+
+    if (warMode.Enabled)
+        _player->CastSpell(_player, warModeSpellId, true);
```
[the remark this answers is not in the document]
ReyDonovan: test in game work

buddyqc69:
do you have any update for this ?

PR #22759, opened 2018-11-03
Core/Unit: Patch 3.0.8 tagging

PolarCookie:
**Changes proposed:**
http://wowwiki.wikia.com/wiki/Patch_3.0.8
> Tapping: All player spells which cause a creature to become aggressive to you will now also immediately cause the creature to be tapped.

https://kotaku.com/5135695/wow-patch-308-brings-important-changes-to-clams
> and the fact that damage over time spells that don't immediately damage monsters now tag the critters as yours anyway is a great boon for several casting classes,

This PR makes all attack interactions that engages a mob tag that mob if possible.

I wanted to remove [the old](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L804-L805) for cleanness, but this commit addition doesn't tag on reflect damage (tried retri aura and thorns) or `.damage X`.

**Target branch(es):** 3.3.5

**Tests performed:** 
Builds and tags by any warlock spell hit I tried.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8183,6 +8183,11 @@ void Unit::AttackedTarget(Unit* target, bool canInitialAggro)
     if (Unit* targetOwner = target->GetCharmerOrOwner())
         targetOwner->EngageWithTarget(this);
 
+    //Patch 3.0.8: All player spells which cause a creature to become aggressive to you will now also immediately cause the creature to be tapped.
+    if (Creature* creature = target->ToCreature())
+        if (!creature->hasLootRecipient() && GetTypeId() == TYPEID_PLAYER)
+            creature->SetLootRecipient(this);
+
     Player* myPlayerOwner = GetCharmerOrOwnerPlayerOrPlayerItself();
     Player* targetPlayerOwner = target->GetCharmerOrOwnerPlayerOrPlayerItself();
     if (myPlayerOwner && targetPlayerOwner && !(myPlayerOwner->duel && myPlayerOwner->duel->Opponent == targetPlayerOwner))


illfated:
Title suggestion: **Core/Unit: patch 3.0.8 tagging taps aggroed creatures**

PolarCookie:
That sentence hurts my brain, I'll meet you halfway.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -8183,6 +8183,11 @@ void Unit::AttackedTarget(Unit* target, bool canInitialAggro)
     if (Unit* targetOwner = target->GetCharmerOrOwner())
         targetOwner->EngageWithTarget(this);
 
+    //Patch 3.0.8: All player spells which cause a creature to become aggressive to you will now also immediately cause the creature to be tapped.
+    if (Creature* creature = target->ToCreature())
+        if (!creature->hasLootRecipient() && GetTypeId() == TYPEID_PLAYER)
```
Shauren: what about pets (pet taunt)

PolarCookie: Pets are not tagging based on damage atm, should they?
I didn't play pet classes on retail.

illfated: Any relation to the case when pets do 100% damage, hence no loot for the player?

PolarCookie: no, pets don't tag at all right now.

PolarCookie: anyway, this PR doesn't touch pets, and i'm not going to either until some documentation shows up.

Shauren (approved):

PolarCookie:
For future reference:
https://gaming.stackexchange.com/questions/12077/in-wow-is-it-better-to-powerlevel-a-character-in-group-or-out-of-group
https://us.battle.net/forums/en/wow/topic/7004402721

PR #22768, opened 2018-11-05
[3.3.5] SmartEnum (again)

Treeston:
The previous macro-based approach slammed into a brick wall thanks to Boost PP's 64-element limit on PP tuples. We have enums that are larger than that.

This time, we stop trying to do stuff at compile time and just pre-generate an additional header file.


jackpoz:
does this actually add python as additional TC dependency for every single user (instead of say, just a TC dev that can run locally the script and update the files and push them) ?

Warpten:
s/runtime/compile-time. c++ is not a JIT language.

This makes more sense in being an additional, optional build step that is only triggered when the files targeted change. CMake does not appear to have file monitoring either, looking through both the docs and the message boards. Plus, the header would rarely get touched.

Treeston:
The current cmakefile in this branch should only re-run the generation if either the generator script or the input headers change. See [add_custom_command](https://cmake.org/cmake/help/v3.0/command/add_custom_command.html).

Warpten:
Will it still retarget any .h file (my python is spotty, but looking through I see it looks for any enum marked with a comment in any .h file) or just the ones that changed?

Treeston:
Just the ones that had any `// ANNOTATE ME` tags the last time cmake was rerun.

jackpoz:
Btw what is the purpose and utility of these enums ?

Treeston:
Iteration over enum values and enum value to string conversion, same as the previous iteration.

jackpoz:
ok but when do we need to iterate over enums in TC atm ?

Treeston:
Plenty of places would like it. Anywhere that involves making values human-readable, e.g. almost every info command.

It's also a mandatory feature for the dbedit branch, and by extension also for #22461.

jackpoz:
ah ok. having the python script in contrib (like it is now 👍 ) we could add some magic scripts to travis/appveyor to report if it needs to be rerun :)

Treeston:
alright, now build passes, everything should be good

@Shauren review + merge please

Treeston:
> And which enums are larger that 64 fields? I can't even imagine that.

Spell effects, among others.

PR #22782, opened 2018-11-10
Core/Conversation Fix UpdateFlag

Traesh:
**Changes proposed:**

-  Update "conversation" update flag (should be true only in case of "static conversation" with window at bottom center of screen)

**Target branch(es):** 
- [X] bfa

**Issues addressed:** None

**Tests performed:**
Many sniffs checked

None


Shauren:
Gets a nope
First, it doesn't matter if its set or not - client doesnt run any extra code based on this bit alone
Second, it should be tied to fields under it being set (if anything), not actors

PR #22783, opened 2018-11-10
Core/Misc: Log more information in asserts

jackpoz:
**Changes proposed:**

-  Add a new function GetDebugInfos() to types that could trigger an ASSERT() to easily include more useful information in crashlogs.
-  If the type doesn't have the function, the global default one is picked which doesn't log anything.
- Types supported:
  - [x] Object
  - [x] WorldLocation
  - [x] WorldObject
  - [x] Unit
  - [x] Creature
  - [x] GameObject
  - [x] Player
  - [x] Item
  - [x] Bag
  - [x] Spell
  - [x] Aura
  - [x] Map
  - [x] InstanceMap
  - [x] base-AI classes
  - [x] Transport
  - [x] TempSummon
  - [x] Minion
  - [x] Pet
  - [x] Guardian
  - [ ] Missed anything ?

On my journey across a lot of different TC classes I found some weird code that I chose to fix in this PR because I was too lazy to switch to another branch.

- [x] 3.3.5
- [ ] master

No issues

It builds, still need to trigger a few asserts on purpose to properly test it

- Add GetDebugInfos() function to more types

### DIFF
--- a/src/common/Debugging/Errors.cpp
+++ b/src/common/Debugging/Errors.cpp
@@ -69,20 +69,20 @@ namespace
 namespace Trinity
 {
 
-void Assert(char const* file, int line, char const* function, char const* message)
+void Assert(char const* file, int line, char const* function, std::string debugInfos, char const* message)
 {
-    std::string formattedMessage = StringFormat("\n%s:%i in %s ASSERTION FAILED:\n  %s\n", file, line, function, message);
+    std::string formattedMessage = StringFormat("\n%s:%i in %s ASSERTION FAILED:\n  %s\n", file, line, function, message) + debugInfos + '\n';
     fprintf(stderr, "%s", formattedMessage.c_str());
     fflush(stderr);
     Crash(formattedMessage.c_str());
 }
 
-void Assert(char const* file, int line, char const* function, char const* message, char const* format, ...)
+void Assert(char const* file, int line, char const* function, std::string debugInfos, char const* message, char const* format, ...)
 {
     va_list args;
     va_start(args, format);
 
-    std::string formattedMessage = StringFormat("\n%s:%i in %s ASSERTION FAILED:\n  %s\n", file, line, function, message) + FormatAssertionMessage(format, args) + '\n';
+    std::string formattedMessage = StringFormat("\n%s:%i in %s ASSERTION FAILED:\n  %s\n", file, line, function, message) + FormatAssertionMessage(format, args) + '\n' + debugInfos + '\n';
     va_end(args);
 
     fprintf(stderr, "%s", formattedMessage.c_str());
@@ -138,3 +138,8 @@ void AbortHandler(int sigval)
 }
 
 } // namespace Trinity
+
+std::string GetDebugInfos()
+{
+    return "";
+}
--- a/src/common/Debugging/Errors.cpp
+++ b/src/common/Debugging/Errors.cpp
@@ -69,20 +69,20 @@ namespace
 namespace Trinity
 {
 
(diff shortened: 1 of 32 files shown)


illfated:
I would like to point out that "infos" (with the plural S) is a modernism predominantly found in non-English languages (typically French). In English, the word "information" is an uncountable noun, as well as the English understanding of "info". Not saying that you should change your command name `debugInfos` -- there may be other reasons for why you chose that word form. I simply wanted to put it into perspective.

Faq:
Hirsoshi would be good tester as lately he has many problems :)

ccrs:
Really necessary

jackpoz:
Please suggest what fields to log for each type (like what should we log for Unit that is not logged already for WorldObject ?)

jackpoz:
Creature:
>Assertion message: 
d:\wow\trinitycore\src\server\game\entities\creature\creature.cpp:295 in Creature::AddToWorld ASSERTION FAILED:
  false
MapID: 571 X: 4948.98 Y: 2937.79 Z: 550.517 O: 1.81514
GUID Full: 0xf130003b6e000001 Type: Creature Entry: 15214  Low: 1 Entry 15214
IsAIEnabled: 1 DeathState: 0
AIName: SmartAI ScriptName: WaypointPath: 0

Shauren:
add SpawnId to creature log

ccrs:
Maybe add motionmaster debug information to unit? take a look to `.debug move` command 👍

jackpoz:
GetUnitMovementFlags() and GetExtraUnitMovementFlags() ?

ccrs:
my mistake, I was thinking about https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Commands/cs_misc.cpp#L2109
but actually thats useful information too 💃

jackpoz:
Updated example of Creature:
>Assertion message: 
d:\giacomo\wow\trinitycore\src\server\game\entities\creature\creature.cpp:295 in Creature::AddToWorld ASSERTION FAILED:
  false
MapID: 571 X: 4948.98 Y: 2937.79 Z: 550.517 O: 1.81514
GUID Full: 0xf130003b6e000001 Type: Creature Entry: 15214  Low: 1 Entry 15214
IsAIEnabled: 1 DeathState: 0 UnitMovementFlags: 1024 ExtraUnitMovementFlags 0
AIName: SmartAI ScriptName: WaypointPath: 0 SpawnId: 0

ccrs:
what about "IsInInstanceSript" and "Instance BossStates" or something like that ?

jackpoz:
they don't sound that much useful, the mapid is already logged and that's enough to know if there is an instancescript

illfated:
Nice progression. 👍

Warpten:
Things off the top of my head for spells:

targets (gameobjects, units, explicit and implicit) (not always set depending on where it crashes!)
caster
cast item
cast flags

jackpoz:
For spells I chose the lazy way of doing
```cpp
"Id: " << GetSpellInfo()->Id << " OriginalCaster: " << m_originalCasterGUID.ToString() << " State: " << getState();
```
logging the caster guid instead of checking all the different types of casters. As for the targets, that looked like a LOT of work to get things right! Maybe the spellId will already be enough (after all this is for asserts only, not for crashes). More info can be logged in the future whenever we get an assertion triggered.

illfated:
Just a little Friday joke: Wouldn't it be funny if either a boss or instance NPC causing a worldserver crash would say "This world is coming to an end. I can sense it." when a crash is imminent? 😄

Seyden:
Someone merge this into master! GO GO GO! :D

Issue #22786, opened 2018-11-12
[3.3.5] Warlock: Haunt


TheWinchesters:
English version:
**Description:**
Haunt (https://www.wowhead.com/spell=48181/haunt):
"When the Haunt spell ends or is dispelled, the soul is returned to you, healing you for 100% of the damage it did to the target".

**Current behaviour:**
Haunt is not renewed if it is casted on a target that already has it.
Now, if you cast Haunt continually, when Haunt returns the heal effect, it will heal the sum of all the damage done by all the times that Haunt was casted on the objective before the effect ends.
EDIT: I think heal returned is too big to be just "the sum of the damage made by all the times that Haunt was casted".

**Steps to reproduce the problem:**
1. Pick some objective.
2. .cheat cooldown
3. Spam Haunt and wait till heal effect is returned to the caster

-------------------------------------------------------------------------------------------------------------
Spanish version:
**Descripción:**
Poseer tiene un efecto de sanación cuando se acaba o es disipado, sanando un 100% del daño infligido.

**Comportamiento actual:**
Poseer no se renueva en un target que ya lo tiene.
Si se lanza continuamente sobre el target, cuando Poseer en el objetivo se acabe, la sanación es la suma del daño de todos los "Poseer" que se han logrado lanzar en el objetivo antes de que se acabe el efecto.
Edit: Creo que la sanación que vuelve es algo más que sólo la suma del daño de todo los "Poseer" lanzados, porque es un número realmente grande.

**Pasos para reproducir el problema:**
1. Buscar un target
2. .cheat cooldown
3. Spamear Poseer hasta que devuelva el efecto de sananción

**Branch(es):**
Branch: 3.3.5
TC rev: 257ae44
OS: Win 8.1

PolarCookie:
Sorta fixed in https://github.com/TrinityCore/TrinityCore/commit/a0999744ef02bf1494233411d873552d3d564d8a, haunt is not stacking the heal and refreshing duration properly, but when I refresh this happens
![wowscrnshot_111318_053616](https://user-images.githubusercontent.com/1158801/48391298-44822500-e706-11e8-8a9e-21ce99ddcb77.jpg)

Langerz82:
Just by looking at the code it may seem that m_spellInfo->Speed is equal to 0 for Haunt. Can someone test this who has more knowledge? I don't want to just remove the conditional as I'm sure it's used for something.

PolarCookie:
[...]

Issue #22051, opened 2018-06-18
Remove Curse "Nothing to dispell" 3.3.5


Jasonqt:
**Description:**

Remove Curse wont dispell target if you have the same curse applied

**Current behaviour:**

Remove Curse doesn't work properly. When you try to remove a curse from a friendly target when you both have the same curse applied, it says "Nothing to dispell". However, if you dont have said curse, and your friendly target does, It works. (Tested on Mage and Druid) 
If this is a crash, post the crashlog (upload to https://gist.github.com/).

**Expected behaviour:**

Even if you have same curse as your partymember(s), you should be able to dispell target.

**Steps to reproduce the problem:**

1. Go to molten core / .go creature 56737
2. Attack Gehennas, and wait for her to apply the debuff to you and your target
3. Try to dispell your partymember(s)

**Branch(es):**

Tested only on 3.3.5

**TC rev. hash/commit:** 718742005c3b

Debug command:
TrinityCore rev. 718742005c3b+ 2018-06-17 17:03:10 +0200 (3.3.5 branch) (Win64, Release, Static)
Using SSL version: OpenSSL 1.0.2l  25 May 2017 (library: OpenSSL 1.0.2l  25 May 2017)
Using Boost version: 1.65.1
Using MySQL version: 10.1.30-MariaDB
Using CMake version: 3.11.2
Compiled on: Windows 6.1.7601
Automatic database updates are disabled for all databases!
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Disabled
MMAPs status: Disabled
maps directory located in ./maps. Total size: 252191207 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS 

**Operating system:** Tested on Debian 9, Windows 7 & 10.

Langerz82:
I created a party of two. Casted "Curse of the elements" on both party members, used "Remove curse" on target party member and its working as intended. Whats the debuff spell id the creature is casting?

Jasonqt:
I forgot to mention, if you do it that way it will work. You have to make a mob cast it on you for it not to work.

Langerz82:
Can you give an example thats not in an instance, with commands to get there. (Im a WoW noob) XD

illfated:
GM character can use `.go creature Gehennas` and then `.summon (player name)`
There are mobs in Stratholme who cast curses too, you may want to check there.
[...]

Issue #21810, opened 2018-04-12
Broken spell auras logic


lost-illusion:
**Description:**

Now spell auras totally broken. Part with static applications.

**Current behaviour:**

Example one:
- aoe spell with dispelable debuff
- you cant dispel this spell from target with static applications
- if you dispel aura from owner - all targets lost aura

Example two:
- spell with speed and stackeble effect on caster, or target
- if spell have speed - we always get exist aura and cant stack

Example three:
- stackable aoe auras stacks on all targets at same time, and lost stacks from all targets

Other strange reports, but cant explain their here

**Expected behaviour:**

Need rework part with static applications. Lot of spells require single aura for single target, without static applications, in additional, part with speed check looks incorrect

**Branch(es):** 3.3.5

**TC rev. hash/commit:** any rew after #21395 or 1e25b21

**Operating system:** any

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you and add the missing information to **this issue**. Thanks.

Faq:
Isnt better in issue to mention atleast 1 spell example?

lost-illusion:
@Faq Use filters in spellwork. But for first case: 32722

lost-illusion:
This a generic problem in logic. Without any spell examples, one aura + many static applications = bad idea (if applied for all spell auras)

Killyana:
I will try to explain this issue:
If 2 npcs casts the same spell and this spell will give the player a DOT/HOT that stacks, if the first npc apply its dots the second npc will remove them when it cast the same spell instead adding an new stack.

To reproduce the issue:
1- .npc add temp 16901
2- aggro and wait until getting 2 stacks of  spell=32722
3- .npc add temp 16901 a second one
4- notice when the second npc casts 32722 you will loose all stacks and they will be replaced by the one from the second npc

According to lost-illusion this happen to spells with the same effect and not only the same ID

TecDian:
Fighting Anub'arak in Azjol'nerub I get a Crash stating
```
CreateAuraApplication ASSERTION FAILED:
  !Aura->IsRemoved()
```

TecDian:
Fighting Anub'arak in Azjol'nerub I get a Crash stating
CreateAuraApplication ASSERTION FAILED:
  !Aura->IsRemoved()

Jildor:
@ariel- we need you 😘

[...]

Issue #21609, opened 2018-03-13
[Crash] - UnitAura::AddStaticApplication


Strazhnik:
**Description:**

Crash in Valithria Dreamwalker

**Current behaviour:**

https://gist.github.com/Strazhnik/d1ef145c429e453fc6f066e76d703cfb

**Steps to reproduce the problem:**

1. Config: hp = 0.51, damage = 0.1 (and default)
2. Several time heal Valithria and continue to kill mobs 
3. Crash

**Branch(es):**
3.3.5

**TC rev. hash/commit:** 
b978b673a9053bb0d7bd63ed9261117dc7845742

**TDB version:** 
 b978b673a9053bb0d7bd63ed9261117dc7845742

**Operating system:**
Ubuntu

ariel-:
> Heal and resurrections Velithria several times

Do you know with what heal it happens?
When you say resurrect, is it related to Valithria?

Strazhnik:
Any time heal Valithria and continue to kill mobs = server crash

ariel-:
No crashes here (b978b673a9053bb0d7bd63ed9261117dc7845742 too)

Strazhnik:
Now crash on Lich and Saurfang. I will make a crash log later.

tkrokli:
@Strazhnik : Please confirm with core rev. hash on the already registered issue
**Crash Npc High Overlord Saurfang** #21644 (if your crash log has the same kind of crash)

Referenced crash is probably in
https://github.com/TrinityCore/TrinityCore/blob/1c8d9af0846036fddef3aa8daf7506d0ebadddb9/src/server/scripts/Northrend/IcecrownCitadel/boss_deathbringer_saurfang.cpp#L692

PR #22794, opened 2018-11-15
Core/Spell: SpellAura Redux

PolarCookie:
**Changes proposed:**
Revisiting https://github.com/TrinityCore/TrinityCore/commit/28288698441524081d583af95f19cb1c9fb90387 and https://github.com/TrinityCore/TrinityCore/commit/a0999744ef02bf1494233411d873552d3d564d8a.
The original issue was that server would crash on multi target delayed spells, because unit auras was being shared between targets and it would cause segmentation faults at times.
Those two commits tried to remedy the problem, but broke spells in different ways.

Solution B would be to delete `Spell::_spellAura` all together, and use `Spell::TargetInfo::HitAura` instead, so far the code is split between their use (notably `SpellScript` uses `Spell::_spellAura`), but they're functionally the same.

**Target branch(es):** 3.3.5

**Issues addressed:** Closes
https://github.com/TrinityCore/TrinityCore/issues/22786

**Tests performed:**
Builds and so far tested
https://www.wowhead.com/spell=33657/resonance
https://www.wowhead.com/spell=48181/haunt
https://www.wowhead.com/spell=59108/glutinous-poison
https://www.wowhead.com/spell=50894/zuldrak-rat
They work, units have their individual stacks and haven't crashed anything yet.

I've not tested Valithria Dreamwalker as reported in https://github.com/TrinityCore/TrinityCore/issues/21609 because I don't know that fight, @Strazhnik could you test this please?


src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -773,7 +773,7 @@ class spell_warl_haunt : public SpellScriptLoader
             {
                 if (Aura* aura = GetHitAura())
                     if (AuraEffect* aurEff = aura->GetEffect(EFFECT_1))
-                        aurEff->SetAmount(CalculatePct(aurEff->GetAmount(), GetHitDamage()));
+                        aurEff->SetAmount(CalculatePct(GetHitDamage(), aurEff->GetAmount()));
```
Warpten: Nice.

PolarCookie: No it doesn't make a difference, but yes it does matter.

jackpoz:
Could you check with @ariel- if anything is missing too ?

kvipka:
this PR will make a trouble with vehicle spell : 
https://dbwotlk.com/?spell=46598

for example q : [Where in the World is Hemet Nesingwary?](https://dbwotlk.com/?quest=12521) 

will instant entervehicle and exitvehicle

PolarCookie:
@kvipka nice catch!

PolarCookie:
Yes, I have a suspicion might push a commit later today if successful.

PolarCookie:
Update; I've been busy and lazy and haven't put in the time I thought I would, but getting back into it.
What I've found so far; https://db.rising-gods.de/?spell=59319 works.
Note that it's a single effect spell, this will be important.
https://db.rising-gods.de/?spell=46598, https://db.rising-gods.de/?spell=49256 and most other mount spells behave like this:
```
Effect 0: Id 6 (SPELL_EFFECT_APPLY_AURA)
BasePoints = 0
Targets (25, 0) (TARGET_UNIT_TARGET_ANY, NO_TARGET)
Aura Id 236 (SPELL_AURA_CONTROL_VEHICLE), value = 1, misc = 0 (0), miscB = 0, periodic = 0

Effect 1: Id 6 (SPELL_EFFECT_APPLY_AURA)
BasePoints = 0
Targets (1, 0) (TARGET_UNIT_CASTER, NO_TARGET)
Aura Id 4 (SPELL_AURA_DUMMY), value = 1, misc = 0 (0), miscB = 0, periodic = 0
```
Target gets the mount aura, player gets some dummy aura.
The issue comes from https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L3258
All SPELL_AURA_CONTROL_VEHICLE auras are hardcoded to be defined as single target, meaning when player applies his dummy aura the control aura applied to the actual mount is dispelled.
This didn't happen before because of sharing auras between targets.

PolarCookie:
To test, try going to Wyrmrest Temple and do https://db.rising-gods.de/?quest=12372.
Try to mount a the dragon then `.list auras` drake and yourself.

Shauren:
But that is exactly why sharing auras was introduced in the first place - removing CONTROL_VEHICLE from target should also clear the DUMMY from player, the same situation as priest's mind control applying aura to both target and caster

PolarCookie:
Could it be done in other ways? Sharing auras was causing seg faults.

PolarCookie:
I wouldn't call revealing hacks breaking stuff.

PolarCookie:
Entering vehicles work now, I've tested a bunch of them.
Residual dummy auras are not an issue, couldn't find any, because https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L2976 exists.

Mind controls and possessions work too without any dummy auras left behind, tested most SPELL_AURA_MOD_POSSESS and SPELL_AURA_AOE_CHARM.

In the case that something was missed, it's easy to fix this with spellscripts, not core hacks.

(reserved, testing hunter pet possession)

Shauren:
FYI, you are now reverting 1addad4dbd38d4690c24ad9c07456e06e443c78e

PolarCookie:
Cool, https://github.com/TrinityCore/TrinityCore/issues/9293#issuecomment-19501634 can you reproduce this crash?

Shauren:
I cant remember exactly what you had to do but it might involve that quest in storm peaks that makes you jump between proto drakes

PolarCookie:
Tested The Drakkensryd, could jump between drakes and didn't crash or anything, no residual auras.

jackpoz:
is there a list of spells that one should check when changing this part of the code ? Like a spell that the first PR fixed, another spell that was broken by that PR and that another one fixed etc ? Just to test it once and for all (or until we find another spell that becomes broken)

PolarCookie:
Stuff to test; in no particular order some spells known to have been broken at some point.

**Multi target delayed spells, some with stacking aura**
https://www.wowhead.com/spell=33657/resonance
https://www.wowhead.com/spell=30531/soul-transfer
(```DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId` = 13 AND `SourceEntry` = 30531;``` if you want to test outside of Mag's Lair)
https://www.wowhead.com/spell=59108/glutinous-poison

**Single target stacking aura**
https://www.wowhead.com/spell=50894/zuldrak-rat

https://woehead.way-of-elendil.fr/?spell=63165/decimation
https://www.wowhead.com/spell=48181/haunt

**Assorted quests with either possession or mount**
Don't need to test the quest itself, just if you can mount/possess without leaving any auras on you afterwards, check with `.list auras`
https://www.wowhead.com/quest=9433/a-dip-in-the-moonwell
https://www.wowhead.com/quest=12521/where-in-the-world-is-hemet-nesingwary
https://www.wowhead.com/quest=12983/the-last-of-her-kind
https://www.wowhead.com/quest=12886/the-drakkensryd
https://www.wowhead.com/quest=13086/the-last-line-of-defense
https://www.wowhead.com/quest=11310/warning-some-assembly-required
https://www.wowhead.com/quest=11470/there-exists-no-honor-among-birds
https://www.wowhead.com/quest=12498/on-ruby-wings
https://www.wowhead.com/quest=10612/the-fel-and-the-furious
(consoles are not scripted right, have to manually cast one of the six https://db.rising-gods.de/?spell=38003)
https://www.wowhead.com/quest=12050/lumber-hack
https://www.wowhead.com/item=22728/steam-tonk-controller

Step into Ulduar and try jumping around the vehicle seats there.


**Valithria Dreamwalker**
https://github.com/TrinityCore/TrinityCore/issues/21609

PolarCookie:
Was able to complete Valithria Dreamwalker without crashing, but dunno if it applies to #21609
@ariel- are you still around?

PolarCookie:
https://db.rising-gods.de/?spell=49078 is the vehicle spell for https://db.rising-gods.de/?quest=12326
https://db.rising-gods.de/?spell=47963 is triggered by it and sticks around after leaving the tank.
Found that, and ~40 other vehicle spells I'm going to test a bit further, mostly following the same pattern of a 2nd aura triggered by the vehicle spell which sticks around. 
Busy tonight, but will look into spellscripting it off.

jackpoz:
Nice list, good job :)

PolarCookie:
I've tested almost all spells with `SPELL_AURA_CONTROL_VEHICLE` effect which could have had side effects.
Except Kologarn's Grip (it's WIP in a different PR), and the following list which I couldn't find where were.
ID - 44249 Rope Swing
ID - 55951 Eye Gouge
ID - 60682 Ride Vehicle
ID - 60683 Ride Vehicle
ID - 61423 Mammoth Ride
ID - 61124 Ingest
Most likely they're fine.

Jumping into the Ulduar Salvaged Demolisher catapult causes issue with leftover auras, but that is broken in general and requires it's own PR.

jackpoz:
@PolarCookie is this PR ready to be merged ? are you available to fix some issues if we merge it in the next days ?

PolarCookie:
@jackpoz yes and yes if Netherwing keeps being unstable.

PolarCookie:
Tested Ulduar Vehicles again, specifically Salvaged Siege Engine.
Entering the Siege Engine, logging off and logging on again will leave a residual https://db.rising-gods.de/?spell=65266 on the player. This doesn't happen if you exit the vehicle normally and happens on both this PR and 3.3.5.
https://db.rising-gods.de/?spell=65031 is the aura which links the gear scaling.

The Oculus drakes did not have this issue however
https://db.rising-gods.de/?spell=49464
the player does not receive the linked aura ever for those drakes.
I think the issue is that the player is not supposed to get the 65266 at all, it's for the vehicle.

kvipka:
try to look xt-002:
[code]DoCast(xt002, SPELL_RIDE_VEHICLE_EXPOSED, true);[/code]

ID - 63313 Ride Vehicle

For me it's crash-assert:
ASSERT(!m_vehicle);

jackpoz:
@kvipka could you post the whole crashlog as Gist ? thanks

Aokromes:
maybe unrelated to PR
https://github.com/TrinityCore/TrinityCore/issues/18565

kvipka:
okay
https://gist.github.com/kvipka/ce0c28f42214a23dfaee384da3a12f36

kvipka:
instantly, yes.
Just give few damage at boss and wait while he starting "Tired" Phase and summon Heart.

change : ASSERT(!m_vehicle);
for : if (m_vehicle)

ofc fix it, but it's wrong way =)

jackpoz:
@PolarCookie can you take a look at the crash ? does it happen only with this PR btw ?

Keader:
It will fix https://github.com/TrinityCore/TrinityCore/issues/22920 too.

PolarCookie:
I'll try to take a look at it sometime before Sunday, I did recreate the XT crash but didn't investigate far before getting busy.

PolarCookie:
Based on https://github.com/TrinityCore/TrinityCore/pull/22972#pullrequestreview-194609327 I unmade the spell effect reordering which was the initial plan to reset Spell::_spellAura between targets, later uneeded for this PR by focusing on TargetInfo::HitAura. See the commit log for details.

Now to retest everything, starting with XT

src/server/game/Spells/Spell.cpp
```diff
@@ -3457,7 +3458,7 @@ void Spell::DoProcessTargetContainer(Container& targetContainer)
             if (target.EffectMask & (1 << i))
                 target.DoTargetSpellHit(this, i);
 
-    for (TargetInfoBase& target : targetContainer)
+    for (uint8 i = 0; i < MAX_SPELL_EFFECTS; ++i)
```
Shauren: this is definitely not right (does it even build?)

PolarCookie: no, working on amending it, forgot to compile first

PolarCookie:
The crash happens at 
https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Unit/Unit.cpp#L12700
Context is that the XT Heart is trying to switch vehicle seating from slot 0 (the chest) to slot 1 (the exposed point in front of XT's knees). The Heart is "entering" the exposed position while still seated in the chest, that's why `m_vehicle` is not NULL. What should happen is for the heart to first "exit" the chest before "entering" the exposed position.

The reason this doesn't happen in the 3.3.5 branch is because of https://github.com/TrinityCore/TrinityCore/pull/22794/commits/241992b5d890ee7e06932e81360824a5be3f99d0
When `SPELL_AURA_CONTROL_VEHICLE` auras were considered Single Target any of them would be considered single target with any other of them, meaning `SPELL_AURA_CONTROL_VEHICLE` aura A would cancel out `SPELL_AURA_CONTROL_VEHICLE` aura B.
In this particular case https://db.rising-gods.de/?spell=63313 and https://db.rising-gods.de/?spell=63852.

This change was made because most `SPELL_AURA_CONTROL_VEHICLE` spells apply the control aura to their vehicle and a dummy aura to the caster.

Solution might be to expand https://github.com/TrinityCore/TrinityCore/blob/241992b5d890ee7e06932e81360824a5be3f99d0/src/server/game/Entities/Unit/Unit.cpp#L12448-L12456 to handle seat swapping.

Killyana:
It fixes #22051 #22272 also

Aokromes:
imho we can merge it since it looks no one does ulduar looking to the almost 1 year old mimiron unable to be killed and the client freeze xd

PolarCookie:
Not the prettiest, but it works (tm).
Should be reconsidered whenever this happens
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L12564-L12574

Are there any other cases where npc units switch seats on their vehicle?

Killyana:
I just tested this PR and this issue i not fixed https://github.com/TrinityCore/TrinityCore/issues/21810

PolarCookie:
are you sure it's within the scope of this PR? That predates when I started working on the stacking auras.

PolarCookie:
went back to the genesis for me, throwing Zul'Drak Rats still stacks up on the Basilisks.

Killyana:
The issue is about 2 npcs casting the same spell that must stack on player, for example npc=
16901 cast spell=32722

PolarCookie:
I get that, but I don't feel motivated at the time to look into it, can it be given it's own PR?
It doesn't sound like aura sharing which this PR disables, more like aura collision.

Killyana:
Yes it could be related something else.

jackpoz:
@PolarCookie could you merge 335 into this branch ? is there anything else to do on it or is it ready to be merged ? (some things can be delayed to another PR)

illfated:
Rebase this branch on 3.3.5 is just as valid, I presume.

PolarCookie:
retesting tomorrow, have to remake vmaps it seems now.

PolarCookie:
Finished testing stuff, can't find any mistakes from the rebases.
@Killyana this can be merged now; I fixed the bug that I wanted from the beginning and the new crash with XT.

https://github.com/TrinityCore/TrinityCore/pull/22794/files#diff-4522eeffcbc961602e86620011a2d579R12504
should there be a TC_LOG_DEBUG line here same as the exit vehicle some lines earlier?

jackpoz:
Sure, feel free to add some TC_LOG_DEBUG lines here and there. Let me know when it's ready to be merged so I can merge it

Killyana:
There's an issue related to vehicles, when you kill an accessory it's no longer ejected from the vehicle for example: npc 29698 and 27213, riders remains dead on the vehicle once killed, I don't know if it's related to this PR.

PolarCookie:
I'll take a look

PolarCookie:
Happens in 3.3.5, so unrelated.
I'll keep it in mind though.

PolarCookie:
It's happening in 3.3.5 so unrelated.
I'll keep this in mind though.

A thought, this could be the use case for the dummy aura on caster for every vehicle control spell.

PolarCookie:
No, I went through with the debugger; killing the rider triggers the normal dismount path.
It might be related to the same issue that I noticed while fixing "That's Abominable!"; the undead get knocked back and 1 second later die midair, where they suddenly stop moving leaving a host of undead corpses hovering.
There's something about dead units not updating their movement in general?

Killyana:
Yes flying npc remains in the air once dead if they have "CanFly" instead of "DisableGravity", but I don't think it's related as in this case the accessory remains on the vehicle.

PolarCookie:
Doesn't seem to be any error on the vehicle, his passenger guid gets set to null, same with the riders m_vehicle.

PolarCookie:
Here is the issue
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L8674-L8725
At line 8688 ExitVehicle produces a GenericMovementGenerator for the jump off.
Then at 8708-8709 it's replaced by a IdleMovementGenerator

PolarCookie:
No idea what to do with it, so ignoring for this PR.

PolarCookie:
No idea why the dead rider follows with the mount when mount moves, since from the mount's point of view the rider is not there anymore.

illfated:
> No, I went through with the debugger; killing the rider triggers the normal dismount path.
> It might be related to the same issue that I noticed while fixing "That's Abominable!"; the undead get knocked back and 1 second later die midair, where they suddenly stop moving leaving a host of undead corpses hovering.
> There's something about dead units not updating their movement in general?

Thanks for noticing that and mentioning it. I am fairly sure you are right in your assumptions. Something similar can be seen when doing the horde quest(s) https://wow.gamepedia.com/Blinding_the_Eyes_in_the_Sky / https://wow.gamepedia.com/Keeping_the_Alliance_Blind (Daily), at least 1 piece of the "Alliance" airplane stays "dead" in the air at the height where the plane used to be, when the plane visual falls down to the ground.

Keader:
@PolarCookie it's caused by this PR or previous one?
If not... make a ticket (issue) about it...
This PR is ready to merge?

PolarCookie:
@Keader, this PR can be merged, the movement issues is related to something else.
Rider not falling off the mount on death is "fixed" by commenting out 
https://github.com/TrinityCore/TrinityCore/blob/3bc1900018fd38d23e656237e751cbe1ebb8552d/src/server/game/Entities/Unit/Unit.cpp#L8708-L8709
Knockbacked units dying and freezing in mid-air is "fixed" by additionally commenting out
https://github.com/TrinityCore/TrinityCore/blob/3bc1900018fd38d23e656237e751cbe1ebb8552d/src/server/game/Entities/Unit/Unit.cpp#L8711-L8712

Killyana:
There's already 2 issues for that, it's related to moments issues not spells

illfated:
Looks like the PR is complete, then. Movement issues belong to tickets outside of this PR.

jackpoz:
Thanks for the PR :)

PolarCookie:
Thanks for merging!
now do #21772 to complete the circle!

PR #22800, opened 2018-11-16
[Bfa] HotfixBlob for more client support

funjoker:
**Changes proposed:**

-  To feed the client with hotfixes which are unecessary for coreside we send plain data from sniff (no need of structs and loadings)

To create table use:
```sql
CREATE TABLE `hotfix_blob` (
	`TableHash` INT(10) UNSIGNED NOT NULL,
	`RecordId` INT(11) NOT NULL,
	`Size` INT(10) UNSIGNED NOT NULL DEFAULT '0',
	`Blob` BLOB NULL,
	PRIMARY KEY (`TableHash`, `RecordId`)
)
COLLATE='utf8_general_ci'
ENGINE=MyISAM;
```
**Target branch(es):**

- [x] bfa

### DIFF
--- a/src/server/game/DataStores/DB2Stores.cpp
+++ b/src/server/game/DataStores/DB2Stores.cpp
@@ -348,6 +348,7 @@ namespace
 
     StorageMap _stores;
     std::map<uint64, int32> _hotfixData;
+    std::map<std::pair<uint32 /*tableHash*/, int32 /*recordId*/>, std::vector<uint8>> _hotfixBlob;
 
     AreaGroupMemberContainer _areaGroupMembers;
     ArtifactPowersContainer _artifactPowers;
@@ -1305,9 +1306,9 @@ void DB2Manager::LoadHotfixData()
         uint32 tableHash = fields[1].GetUInt32();
         int32 recordId = fields[2].GetInt32();
         bool deleted = fields[3].GetBool();
-        if (_stores.find(tableHash) == _stores.end())
+        if (_stores.find(tableHash) == _stores.end() && _hotfixBlob.find(std::make_pair(tableHash, recordId)) == _hotfixBlob.end())
         {
-            TC_LOG_ERROR("sql.sql", "Table `hotfix_data` references unknown DB2 store by hash 0x%X in hotfix id %d", tableHash, id);
+            TC_LOG_ERROR("sql.sql", "Table `hotfix_data` references unknown DB2 store by hash 0x%X and has no reference to `hotfix_blob` in hotfix id %d", tableHash, id);
             continue;
         }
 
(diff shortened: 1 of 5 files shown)


src/server/game/DataStores/DB2Stores.cpp
```diff
@@ -1282,6 +1283,48 @@ DB2StorageBase const* DB2Manager::GetStorage(uint32 type) const
...
+
+    uint32 count = 0;
+    do
+    {
+        Field* fields = result->Fetch();
+
+        uint32 tableHash = fields[0].GetUInt32();
+
+        HotfixBlobData blobData;
+        blobData.RecordId = fields[1].GetInt32();;
```
P-Kito: copy pasta double semicolon I guess? :-P

src/server/game/DataStores/DB2Stores.cpp
```diff
@@ -1282,6 +1283,48 @@ DB2StorageBase const* DB2Manager::GetStorage(uint32 type) const
...
+        blobData.Size = fields[2].GetUInt32();;
+        blobData.Blob = fields[3].GetBinary();;
```
P-Kito: I may be missreading this, please help me: Why is a vector needed if you don't need the different indexes?

src/server/game/DataStores/DB2Stores.h
```diff
@@ -258,7 +267,10 @@ class TC_GAME_API DB2Manager
     DB2StorageBase const* GetStorage(uint32 type) const;
 
     void LoadHotfixData();
+    void LoadHotfixBlob();
     std::map<uint64, int32> const& GetHotfixData() const;
+    std::multimap<uint32 /*hash*/, HotfixBlobData> const& GetHotfixBlob();
```
P-Kito: Another question... I need clariification: One hash has multiple HotfixBlobData? Is that the reason why you choose a multimap?

src/server/game/DataStores/DB2Stores.cpp
funjoker: ReadBinary Output is vector

src/server/game/DataStores/DB2Stores.cpp
```diff
@@ -1282,6 +1283,54 @@ DB2StorageBase const* DB2Manager::GetStorage(uint32 type) const
...
+            ++count;
+        }
+    } while (result->NextRow());
+
+    TC_LOG_INFO("server.loading", ">> Loaded %u hotfix blob records in %u ms", count, GetMSTimeDiffToNow(oldMSTime));
+}
+
+HotfixBlobData const* DB2Manager::GetHotfixBlobData(uint32 tableHash, uint32 recordId)
+{
+    for (HotfixBlobContainer::const_iterator _itr = _hotfixBlob.equal_range(tableHash).first; _itr != _hotfixBlob.equal_range(tableHash).second; ++_itr)
```
mdX7: save equal_range return value in local variable

src/server/game/Handlers/HotfixHandler.cpp
```diff
@@ -78,6 +79,16 @@ void WorldSession::HandleHotfixRequest(WorldPackets::Hotfix::HotfixRequest& hotf
                 hotfixData.Data = boost::in_place();
                 storage->WriteRecord(hotfixData.RecordID, GetSessionDbcLocale(), *hotfixData.Data);
             }
+            else if (blobData)
```
Shauren: storage will be NULL for db2s we dont load (will crash on HasRecord)

src/server/game/DataStores/DB2Stores.cpp
P-Kito: Ahhhh... thanks

src/server/game/Handlers/HotfixHandler.cpp
mdX7: We'd only reach this block in case db2 hasn't that record: `if (storage->HasRecord(hotfixData.RecordID))`, never the less this else if block doesn't work properly this way. Switching the blocks should do tho. So we kinda primarily use `if (blobData)` and only use `storage->HasRecord(hotfixData.RecordID)` when there is no blobdata available.

Btw, shouldn't we differ between locales here aswell? Quite hard for some tables when just sending pure data - yes, but just thinking about this.

Shauren: No, being at this line requires `if (int32 const* hotfix = Trinity::Containers::MapGetValuePtr(hotfixes, hotfixId))` to pass. That data is loaded from hotfix_data BUT entries that point to db2s that are not loaded are SKIPPED. If you change that to be not skipped then you get a crash because `storage` is NULL

Aokromes:
add sql to pr!

welder1976:
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'Blob FROM hotfix_blob ORDER BY TableHash' at line 1
Error while parsing SQL. Core fix required.

Shauren (approved):

PR #22804, opened 2018-11-17
Core/Vehicles: define and implement VEHICLE_SEAT_FLAG_DISABLE_GRAVITY

Ovahlord:
**Changes proposed:**
Implement VEHICLE_SEAT_FLAG_DISABLE_GRAVITY. After some research I saw that there are several camera issues when using some certain vehicles such as the gargoyle in Zul'Drak that is taking you on a ride arround. Also sniff data seem to confirm that following some vehicle joins there is an disable gravity opcode involved probably to have exactly this effect.

I also encountered some issues while doing shit in the Throne of the Four Winds with vehicles so I started doing some trial and errors. I picked all vehicle Id's that were having that named camera issue and started to compare the seat flags of all those seats and the short story: this flag is the only one that all seats have in common. It might also be that this seat flag is somehow related to inheriting movementflags of the vehicle but I did not see any clear case for that.

- master

- Tested ingame.

### DIFF
--- a/src/server/game/DataStores/DBCEnums.h
+++ b/src/server/game/DataStores/DBCEnums.h
@@ -1009,7 +1009,7 @@ enum VehicleSeatFlags
 {
     VEHICLE_SEAT_FLAG_HAS_LOWER_ANIM_FOR_ENTER                         = 0x00000001,
     VEHICLE_SEAT_FLAG_HAS_LOWER_ANIM_FOR_RIDE                          = 0x00000002,
-    VEHICLE_SEAT_FLAG_UNK3                                             = 0x00000004,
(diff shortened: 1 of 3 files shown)


Keader:
https://travis-ci.org/TrinityCore/TrinityCore/builds/456424127#L1471

Faq:
label is wrong, this is for master

Shauren (approved):

Killyana:
It must be also pushed to 3.3.5.

Killyana:
@Ovahlord on 3.3.5 the flag VEHICLE_SEAT_FLAG_UNK3 is not used and this issue about the camera fallen from the vehicle doesn't happen.

So probably this "disable gravity thing" is implemented somewhere else on 3.3.5 or VEHICLE_SEAT_FLAG_UNK3 musn't be used for this purpose.

PR #22727, opened 2018-10-28
Spell/Script: Scourge Disguise and Haunting Phantoms

PolarCookie:
**Changes proposed:**
Found by lucky concidence https://github.com/TrinityCore/TrinityCore/blob/master/src/server/scripts/EasternKingdoms/ShadowfangKeep/shadowfang_keep.cpp#L324-L361.
Used it as a base to solve https://github.com/TrinityCore/TrinityCore/issues/22687 https://github.com/TrinityCore/TrinityCore/issues/22662

**Target branch(es):**
3.3.5

**Issues addressed:**
https://github.com/TrinityCore/TrinityCore/issues/22687 can be closed.
https://github.com/TrinityCore/TrinityCore/issues/22662 this PR does not make the NPCs actually cast this spell, just fixes the spell script itself.

**Tests performed:**
Builds and works in-game.

**Known issues and TODO list:**
- [x] When Scourge Disguise Expiring is triggered, a warning should flash across the screen; "Scourge Disguise Failing! Find a safe place!".
https://www.wowhead.com/item=38699/ensorcelled-choker#comments:id=458778

PR #22806, opened 2018-11-18
Quest: Ruthless Cunning and Returning the Favor

PolarCookie:
**Changes proposed:**
Fixes quests [Ruthless Cunning](https://www.wowhead.com/quest=9927/ruthless-cunning) and [Returning the Favor](https://www.wowhead.com/quest=9931/returning-the-favor).
Proper spellscripts now and can only be used on the limited targets, who are dead and don't have a banner on already without despawning.

@Shauren if you don't like `CONDITION_NEAR_GAMEOBJECT` because it destroys performance, make it so that I can do `target->SetRealData(1,1);` on an arbitrary unit and then use `CONDITION_HAS_REAL_DATA (1,1)`.
Would be even better than `CONDITION_NEAR_GAMEOBJECT` as right now the spell fails if mobs are to close.
I'm just working with the tools at hand here.

Edit: And make a note of it on the wiki!
See https://github.com/TrinityCore/TrinityCore/pull/22727#discussion_r228800994 also

**Target branch(es):**
3.3.5

**Tests performed:**
tested


jackpoz:
What tools do you need more ? It would be a good idea to take this chance to add them

PolarCookie:
A way to tag creatures with something that can be read by conditions, the functions I suggested in the second paragraph would work but I'm not picky if someone can think of something better.

src/server/scripts/Outland/zone_nagrand.cpp
```diff
@@ -636,6 +636,40 @@ class go_warmaul_prison : public GameObjectScript
...
+
+// 32314 - Plant Kil'Sorrow Banner
+class spell_q9931_plant_kilsorrow_banner : public SpellScript
+{
+    PrepareSpellScript(spell_q9931_plant_kilsorrow_banner);
+
+    void HandleHit(SpellEffIndex /*effIndex*/)
+    {
+        if (Player* caster = GetCaster()->ToPlayer())
+            caster->RewardPlayerAndGroupAtEvent(18393, caster);
```
P-Kito: I think you should name these 2 numbers

```cpp
 enum PlantBanners
 {
     NPC_XYZ                     = ...,
     QUEST_XYZ                   = ...,
};
```

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -3630,15 +3630,6 @@ void Spell::EffectScriptEffect(SpellEffIndex effIndex)
                     unitTarget->RemoveMovementImpairingAuras(true);
                     break;
                 }
-                // Plant Warmaul Ogre Banner
-                case 32307:
-                    if (Player* caster = m_caster->ToPlayer())
-                    {
-                        caster->RewardPlayerAndGroupAtEvent(18388, unitTarget);
-                        if (Creature* target = unitTarget->ToCreature())
-                            target->DespawnOrUnsummon();
```
P-Kito: What about the despawn? It was ninja removed, is it not needed anymore?

sirikfoll:
I'm not sure about the standard about this, but personally I would prefer if you just update the event_scripts and drop both the SpellScripts, the conditions, in lack of a better way, are fine for now
```sql
UPDATE `event_scripts` SET `delay`=0 WHERE `id`=11669;
UPDATE `event_scripts` SET `datalong2`=1 WHERE `id`=11674;
```

src/server/scripts/Outland/zone_nagrand.cpp
illfated:
```cpp
enum PlantBannerCredits
{
    NPC_SHADOW_COUNCIL_CREDIT_MARKER = 18388,
    NPC_WARMAUL_OGRE_CREDIT_MARKER   = 18393
};

Comment (P-Kito) [reply src/server/scripts/Outland/zone_nagrand.cpp]: Thanks, was too lazy and I thought he could figure out by himself 🎃

Comment (illfated) [reply src/server/scripts/Outland/zone_nagrand.cpp]: Maybe busy with more "important" stuff than looking into these details.

Comment (PolarCookie) [reply src/server/scripts/Outland/zone_nagrand.cpp]: Busy with being sick and work, but have some time off now.

**Comment by jackpoz** [comment]
is this ready to be merged now ?

**Comment by ccrs** [comment]
check `ConditionScript`, will do whatever you need to do

**Comment by PolarCookie** [comment]
@jackpoz ready to be merged now.

I've rewritten the SAIs, only difference is that now Warmaul Shamans cast Frost Shock as they should, tho the rank is probably wrong.
https://www.wowhead.com/npc=18064/warmaul-shaman#comments:id=245347

Kil'Sorrow Ritualists are no longer included for this quest as they're not at Kil'sorrow Fortress and I don't want to try translating their SAI.
https://www.wowhead.com/npc=18658/kilsorrow-ritualist#drops

File: src/server/scripts/Outland/zone_nagrand.cpp [hunk 3]
```diff
@@ -636,6 +635,323 @@ class go_warmaul_prison : public GameObjectScript
...
+            if (!UpdateVictim())
+                return;
+
+            scheduler.Update(diff, [this]
+            {
+                DoMeleeAttackIfReady();
+            });
+        }
+
+        bool bannered;
```
jackpoz: please expose this field with an accessor method like bool IsBannered() and make the field private

src/server/scripts/Outland/zone_nagrand.cpp
```diff
@@ -636,6 +635,323 @@ class go_warmaul_prison : public GameObjectScript
...
+        {
+            if (!has_fleed && me->GetHealth() > damage && me->HealthBelowPctDamaged(15, damage))
+            {
+                me->DoFleeToGetAssistance();
+                has_fleed = true;
+            }
+        }
+
+    private:
+        bool has_fleed;
```
jackpoz: please always initialize all fields in the constructor

src/server/scripts/Outland/zone_nagrand.cpp
```diff
@@ -636,6 +635,323 @@ class go_warmaul_prison : public GameObjectScript
...
+        {
+            if (!used_bloodthirst && me->GetHealth() > damage && me->HealthBelowPctDamaged(50, damage))
+            {
+                DoCastVictim(SPELL_BLOODTHIRST);
+                used_bloodthirst = true;
+            }
+        }
+
+    private:
+        bool used_bloodthirst;
```
jackpoz: please initialize all fields in the constructor

src/server/scripts/Outland/zone_nagrand.cpp
```diff
@@ -636,6 +635,323 @@ class go_warmaul_prison : public GameObjectScript
...
+        {
+            if (!used_transform && me->GetHealth() > damage && me->HealthBelowPctDamaged(65, damage))
+            {
+                DoCastVictim(SPELL_GISELDA_TRANSFORM_DND);
+                used_transform = true;
+            }
+        }
+
+    private:
+        bool used_transform;
```
jackpoz: please initialize all fields in the constructor

src/server/scripts/Outland/zone_nagrand.cpp
```diff
@@ -636,6 +635,323 @@ class go_warmaul_prison : public GameObjectScript
...
+        {
+            if (!used_healing && me->GetHealth() > damage && me->HealthBelowPctDamaged(50, damage))
+            {
+                DoCastSelf(SPELL_HEALING_WAVE);
+                used_healing = true;
+            }
+        }
+
+    private:
+        uint32 used_healing;
```
jackpoz: please initialize all fields in the constructor

jackpoz (changes requested):

Killyana:
No need to convert this event_scripts into cpp spell script:'spell_q9927_plant_warmaul_ogre_banner',
 'spell_q9931_plant_kilsorrow_banner'

Just add the sql that sirikfoll mentioned: https://github.com/TrinityCore/TrinityCore/pull/22806#issuecomment-440889427

jackpoz (approved):

jackpoz:
@Killyana does it look good now ?

illfated:
I don't know if it could be used in other parts of the source than in `zone_nagrand.cpp`,
but if there is any chance at all that the function
```cpp
    private:
        bool has_fleed;
```
could be used elsewhere, I would recommend using standard English when naming the function.

Killyana:
Yes looks fine for me.

jackpoz:
Thanks for the PR :)

src/server/scripts/Outland/zone_nagrand.cpp
```diff
@@ -636,6 +635,292 @@ class go_warmaul_prison : public GameObjectScript
...
+        void UpdateAI(uint32 diff) override
+        {
+            if (!UpdateVictim())
+                return;
+
+            interrupt_cooldown += diff;
+            if (me->HasUnitState(UNIT_STATE_CASTING))
+                return;
+
+            if (me->GetVictim()->HasUnitState(UNIT_STATE_CASTING) && interrupt_cooldown > 25000)
```
jackpoz: Note: this currently crashes because UpdateVictim() returns true even if me->GetVictim() returns nullptr, see https://github.com/TrinityCore/TrinityCore/pull/22875 for a proposed change to change that behavior

jackpoz: Adding a NULL check in https://github.com/TrinityCore/TrinityCore/pull/22878

PR #22816, opened 2018-11-20
Core/Objects: properly use UPDATETYPE_CREATE_OBJECT2

### LINKED ISSUE #22419: Implement UPDATETYPE_CREATE_OBJECT2
`UPDATETYPE_CREATE_OBJECT2` is currently not implemented correctly.

The better name for it would be `UPDATETYPE_SPAWN_OBJECT`. It should be sent whenever an entity first spawns into the world, as opposed to `UPDATETYPE_CREATE_OBJECT`, which is sent when entities enter visibility range.

Possible implementation would be to have an internal flag on objects which is temporarily set and removed in `Map::AddToMap` before and after triggering the initial `UpdateObjectVisibility`. If the flag is set, `updateType` will be changed to `UPDATETYPE_SPAWN_OBJECT` in `Object::BuildCreateUpdateBlockForPlayer`. All the current code that deals with `UPDATETYPE_CREATE_OBJECT2` can be safely removed. In my case I also had it added around the visibility update in `Creature::Respawn` and `GameObject::Spawn`, but that shouldn't be relevant to TC because GUIDs are no longer reused for respawns and because you lack the latter method.

The only way I managed to discover that this affects the client, is that the entity will play the "Birth" animation if available, e.g. ghouls will dig themselves out of the ground etc (without the use for `SPELL_EFFECT_SPAWN` of course).

P.S. `UPDATETYPE_NEAR_OBJECTS` which still exists in 3.3.5 branch has only one effect on the client - it calls the `Reenable` virtual method on every entity which GUID is sent, i.e. `CGUnit_C::Reenable`, `CGGameObject_C::Reenable`, etc, which is pretty pointless for already enabled objects. I guess it could be used to reenable disabled objects (those that have been removed with `UPDATETYPE_OUT_OF_RANGE_OBJECTS` but have not yet been garbage collected by the client and persist in memory in hopes of reentering visibility soon) to examine them by the devs, but if it ever had a proper use at some point, I cannot imagine what it could've been. The server cannot know which objects have or have not yet been garbage collected by the client, for it to send `UPDATETYPE_NEAR_OBJECTS` for those haven't when they reenter visibility...

Is this kind of issue report acceptable? If so, I will have a lot of useful and relevant info for you. I won't be submitting PRs since I don't work with TrinityCore's codebase, but with a project that branched off of it many-many years ago, but I'm willing to share my work and research in textual form.

Ovahlord:
**Changes proposed:**

-  get rid of this extra case madness for UPDATETYPE_CREATE_OBJECT2. This update type is used for every single object that is getting created for the first time. Thanks to some previous work in the dynamic spawn system this can be handled properly now. The cases were also wrong because there are some non-summoner gameobject spawns for example that require this update type to handle spawn animations.

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #22419

Tested ingame, checked packet logs, works as intended.

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -167,72 +167,18 @@ void Object::BuildCreateUpdateBlockForPlayer(UpdateData* data, Player* target) c
     if (!target)
         return;
 
-    uint8  updateType = UPDATETYPE_CREATE_OBJECT;
+    uint8  updateType = m_isNewObject ? UPDATETYPE_CREATE_OBJECT2 : UPDATETYPE_CREATE_OBJECT;
     uint16 flags      = m_updateFlag;
 
     /** lower flag1 **/
     if (target == this)                                      // building packet for yourself
         flags |= UPDATEFLAG_SELF;
 
-    if (m_isNewObject)
+
+    if (isType(TYPEMASK_UNIT))
     {
-        switch (GetGUID().GetHigh())
-        {
-            case HighGuid::Player:
-            case HighGuid::Pet:
-            case HighGuid::Corpse:
-            case HighGuid::DynamicObject:
-                updateType = UPDATETYPE_CREATE_OBJECT2;
-                break;
-            case HighGuid::Unit:
-            case HighGuid::Vehicle:
-            {
-                if (ToUnit()->IsSummon())
-                    updateType = UPDATETYPE_CREATE_OBJECT2;
-                break;
-            }
-            case HighGuid::GameObject:
-            {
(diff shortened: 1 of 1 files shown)


Kittnz:
Any examples of "The cases were also wrong because there are some non-summoner gameobject spawns for example that require this update type to handle spawn animations." ?

Ovahlord:
Any manually summoned gameobject with an spawn animation.

Ovahlord:
Ozruk's Rupture rock spike lines which is suposed to play the create animation as well.

Shauren (approved):

PR #22846, opened 2018-12-01
Core/Scripting: Implement Script reloading

Seyden:
**Changes proposed:**

-  Implementing the way to be able to reload new scripts
-  Not removing unlinked scripts out of the script system

**Target branch(es):** BfA/Master

- [x] BfA

- [x] Builds
- [x] Tested with Creature_template scripts
- [x] Tested with Spell_script_names


illfated:
PR title suggestion: **Core/Scripting: implement script reloading**
("implemented" looks as if it was done a long time ago)

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9225,20 +9225,34 @@ std::string const& ObjectMgr::GetScriptName(uint32 id) const
...
+        _scriptNamesStore.push_back(name);
+        return _scriptNamesStore.size() - 1;
+    }
 
     return uint32(itr - _scriptNamesStore.begin());
 }
 
+bool ObjectMgr::FindScriptId(std::string const& name) const
+{
+    // use binary search to find the script name in the sorted vector
```
joschiwald: you don't do it anymore

Seyden: Thx updated

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9225,20 +9225,32 @@ std::string const& ObjectMgr::GetScriptName(uint32 id) const
...
     return uint32(itr - _scriptNamesStore.begin());
 }
 
+bool ObjectMgr::FindScriptId(std::string const& name) const
+{
+    if (name.empty())
+        return 0;
+
+    ScriptNameContainer::const_iterator itr = std::find(_scriptNamesStore.begin(), _scriptNamesStore.end(), name);
+    return itr != _scriptNamesStore.end() && *itr == name;
```
jackpoz: How come you are checking twice for equality ?

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9225,20 +9225,32 @@ std::string const& ObjectMgr::GetScriptName(uint32 id) const
...
     // assume "" is the first element
     if (name.empty())
         return 0;
 
-    ScriptNameContainer::const_iterator itr = std::lower_bound(_scriptNamesStore.begin(), _scriptNamesStore.end(), name);
+    ScriptNameContainer::const_iterator itr = std::find(_scriptNamesStore.begin(), _scriptNamesStore.end(), name);
     if (itr == _scriptNamesStore.end() || *itr != name)
-        return 0;
+    {
+        _scriptNamesStore.push_back(name);
```
jackpoz: Is this thread safe ?

jackpoz (changes requested):

sql/base/auth_database.sql
```diff
@@ -1888,7 +1888,7 @@ INSERT INTO `rbac_permissions` VALUES
 (695,'Command: reload spell_loot_template'),
 (696,'Command: reload spell_linked_spell'),
 (697,'Command: reload spell_pet_auras'),
-(698,'Command: reload spell_proc_event'),
+(698,'Command: reload spell_script_names'),
```
Shauren: Do not use 698, taken by `character changeaccount` (these ids MUST match between branches)

src/server/game/Globals/ObjectMgr.cpp
Shauren: Because this is a copypaste from GetScriptId done without understanding how it works

Also a function named FindScriptId should return something more than just a bool "does it exist"

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -913,17 +920,6 @@ class SpecializedScriptRegistry<ScriptType, true>
             sScriptRegistryCompositum->SetScriptNameInContext(script->GetName(),
                 sScriptMgr->GetCurrentScriptContext());
         }
-        else
```
Shauren: So you are now leaking memory when script isnt assigned to anything?

Seyden: Well thats exactly what i do, if you as example create a new script via hotswap and you don't got the database entry yet, the script will be removed. Thats why i dont delete them. Better idea how i can get them back if they're removed? 🤔

Shauren: But you don't need to. You seem to misunderstand how hotswap works, it completely reloads all scripts dlls (that means replacing all the code, not just AIs) and when new one is loaded, all AddSC functions are called

Seyden: yeah but thats the problem, if you as example first save the new script, which triggers the hotswap system and calls all AddSC functions, afterwards you put your scriptname assignment into the database and you want to reload, but there isn't any script which is assigned to this scriptname

src/server/game/Globals/ObjectMgr.cpp
Seyden: No its probably not. If the core uses multiple threads to spawn entities it would probably create a issue here. Any ideas about how to make it thread safe?

Shauren: > 
> 
> No its probably not. If the core uses multiple threads to spawn entities it would probably create a issue here. Any ideas about how to make it thread safe?

Wrong. The only time this function can possibly be called when the scriptname isnt known is during reload triggered by gm/console command running in main thread (its safe)

Seyden: For fucks sake Shauren xD you're right again, just checked, its not called on spawn or smth, only when you're reloading or loading at core start. Well yeah you already said the other thing.

src/server/game/Scripting/ScriptMgr.cpp
Seyden: So it this always a issue?

jackpoz: > if you as example first save the new script, which triggers the hotswap system and calls all AddSC functions, afterwards you put your scriptname assignment into the database and you want to reload, but there isn't any script which is assigned to this scriptname

what about triggering again HotSwap script unloading and re-loading then in that case ?

Seyden: Could do that, but in that case every reloading method has to call that method. Thats an option too.

Naios: No, it's unaccetpable to remove the deferred deletion of the script.
Why you can't keep it?`It is still valid to delete the script in case no database scriptname was found.

Naios:
The big issue of the reload command approach used here is that it leaves old scriptnames inside the core when removing it from the database, I think that's why I decided once not to support new scriptnames for database bound entities from scratch.

For spell or aura scripts which keep their corresponding dll alive until destroyed (lazy unloading) this isn't an issue but for the other script types you've updated like achievement scripts, battlegrounds etc. this is a huge issue.

What you could do is to add the get or add semantics only to spell scripts while keeping the old behaviour for other scripts.

Seyden:
> The big issue of the reload command approach used here is that it leaves old scriptnames inside the core when removing it from the database, I think that's why I decided once not to support new scriptnames for database bound entities from scratch.
> 
> For spell or aura scripts which keep their corresponding dll alive until destroyed (lazy unloading) this isn't an issue but for the other script types you've updated like achievement scripts, battlegrounds etc. this is a huge issue.
> 
> What you could do is to add the get or add semantics only to spell scripts while keeping the old behaviour for other scripts.

Whats about Creatures? Don't they hold onto their script too until destroyed? I mean isn't basically each script holding on to their script until they're destroyed?

Seyden:
> The big issue of the reload command approach used here is that it leaves old scriptnames inside the core when removing it from the database, I think that's why I decided once not to support new scriptnames for database bound entities from scratch.
> 
> For spell or aura scripts which keep their corresponding dll alive until destroyed (lazy unloading) this isn't an issue but for the other script types you've updated like achievement scripts, battlegrounds etc. this is a huge issue.
> 
> What you could do is to add the get or add semantics only to spell scripts while keeping the old behaviour for other scripts.

Whats about Creatures? Don't they hold onto their script too until destroyed? I mean isn't basically each script holding on to their script until they're destroyed?

Naios:
The AIs of creatures are swapped immediately when the corresponding script project changes

Seyden:
Ahhh i remember a version where the creature actually stayed alive. Now the creature immediatley despawns in order to reload his script. Well thats shitty because using this system shouldn't be restricted to spells. Its actually only a developer feature.

jackpoz:
I never understood the big deal about restarting worldserver.exe

Seyden:
> I never understood the big deal about restarting worldserver.exe

If you code multiple dungeons, fix hundreds of spells, you will notice it. Restarting the server after a new script, takes like 20-40 sec. You get kicked out of the world and you have to enter your credentials again, login again and enter the world again. Then you have to teleport yourself to the dungeon or what ever, give yourself god mode and all that stuff, its annoying tbh :P

jackpoz:
That's just common real world development

P-Kito:
We are the first to make a better Game Developing World 🥇

Aokromes:
Conflicting files
sql/base/auth_database.sql
src/server/game/Accounts/RBAC.h

Issue #22817, opened 2018-11-20
[335] Enchanting Blood Draining is not healing


devfrom:
When you enchant your weapon with enchanting Blood Draining

Type: Enchant Weapon

Spell: Blood DrainDraining

Wowhead: http://www.wowhead.com/spell=64579

Description: The problem is that the buff does not cure the result when it drops below 35% HP

How this should work: Sometimes your weapon is due to a Blood Reserve (http://www.wowhead.com/spell=64568) when striking an enemy or dealing damage with bleeding attacks. When you fall below 35% health, the Blood Reserve restores from 360 to 440 health.

Branch: 3.3.5

TC rev: d8fee3c
Operating system: Windows 2008

PR #22862, opened 2018-12-14
DB/Spell: fix mistake in Blood Draining's CheckProc and remove Attack Power scaling

Rynem:
**Changes proposed:**

- ~~Add missing entry in the spell_proc table for the spell [Blood Reserve](https://www.wowhead.com/spell=64568) (triggered after every damage received), which allows the [Blood Draining](https://www.wowhead.com/spell=64579) weapon enchant to work properly.~~
Plot twist: it was working fine all along.

- Fix small issue in the AuraScript. The proc is checked _after_ the damage is dealt, so using HealthBelowPctDamaged doesn't make sense (it would subtract the same damage from the current health before computing the percentage of health left, double-dipping).

- Don't allow the heal proc to scale with attack power.

**Target branch(es):** both.

**Issues addressed:** None.

**Tests performed:** builds, tested and working properly.

Issue #22817 was probably fixed by an earlier commit, or it was a wrong report. Should be closed.

### DIFF
--- a/sql/updates/world/3.3.5/9999_99_99_99_world.sql
+++ b/sql/updates/world/3.3.5/9999_99_99_99_world.sql
@@ -1,5 +1,1 @@
-DELETE FROM `spell_proc` WHERE `SpellId`=64568;
-INSERT INTO `spell_proc` (`SpellId`, `SchoolMask`, `SpellFamilyName`, `SpellFamilyMask0`, `SpellFamilyMask1`, `SpellFamilyMask2`, `ProcFlags`, `SpellTypeMask`, `SpellPhaseMask`, `HitMask`, `AttributesMask`, `ProcsPerMinute`, `Chance`, `Cooldown`, `Charges`) VALUES
-(64568, 0, 0, 0, 0, 0, 0x100000, 0, 0, 0, 0, 0, 0, 0, 0);
-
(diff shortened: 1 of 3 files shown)


sql/updates/world/3.3.5/9999_99_99_99_world.sql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM `spell_proc` WHERE `SpellId`=64568;
+INSERT INTO `spell_proc` (`SpellId`, `SchoolMask`, `SpellFamilyName`, `SpellFamilyMask0`, `SpellFamilyMask1`, `SpellFamilyMask2`, `ProcFlags`, `SpellTypeMask`, `SpellPhaseMask`, `HitMask`, `AttributesMask`, `ProcsPerMinute`, `Chance`, `Cooldown`, `Charges`) VALUES
+(64568, 0, 0, 0, 0, 0, 0x100000, 0, 0, 0, 0, 0, 0, 0, 0);
```
Keader: i dont like it.
Why change proc flag is needed?

Rynem:
That's strange. I could've sworn it wasn't working until I made a spell_proc entry for it, but now I tried without that and it still works.
I'm not sure what went wrong, but it appears to work fine now.

illfated:
Some times changes don't show up when testing, either because of old data in the client Cache folder or maybe not fully reloaded DB.

PR #22872, opened 2018-12-17
Core/Player: update player phase before updating area depending auras

Ovahlord:
**Changes proposed:**
-  Update the player phase when updating an area before updating the area auras. After some research for Gilneas' transition into chapter 2 I have seen that Blizzard is using a own phase for that intro (Phase 105) This phase contains several dummy npc's which serve as summon targets for the intro stuff. Since this intro needs to be retro-active for players who have logged out for example they are using spell_area like mechanics to trigger the intro over again which requires the player to be in that phase before doing the corresponding area cast.

- [ ] 3.3.5
- [X] master

Tested ingame with own custom implementations of Gilneas on my fork.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -6974,8 +6974,8 @@ void Player::UpdateArea(uint32 newArea)
     pvpInfo.IsInFFAPvPArea = area && (area->Flags[0] & AREA_FLAG_ARENA);
     UpdatePvPState(true);
 
-    UpdateAreaDependentAuras(newArea);
     PhasingHandler::OnAreaChange(this);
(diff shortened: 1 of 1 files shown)


Shauren:
But `UpdateAreaDependentAuras` affects only spell_area and that one doesnt have any phase condition so what exactly does this change?

Ovahlord:
SpellArea needs an up to date phasing first to make things like db targets that are in a different phase being targetable first.

As I have mentoined in Duskhaven:
- the intro dummy targets are suposed to be in phase 105. When the player enters that area he needs to be in that phase so spell_area can handle the summon part and also to make sure that the summons of the player wont turn invisible when the player changes the phase after the summon.

PR #22875, opened 2018-12-19
Core/AI: Clarify CreatureAI::UpdateVictim() return values

jackpoz:
**Changes proposed:**

-  Change CreatureAI::UpdateVictim() to return true only if me->GetVictim() would return a valid object. This is the expected behavior by a lot of scripts:
```cpp
{
  if (!UpdateVictim())
    return;
  me->GetVictim()->DoStuff();
}
```

- [x] 3.3.5
- [ ] master

**Issues addressed:**

 None

No build or tests performed

No todo list

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -193,7 +193,7 @@ bool CreatureAI::UpdateVictim()
     else if (me->GetVictim())
         me->AttackStop();
 
-    return true;
+    return me->GetVictim() != nullptr;
 }
 
 bool CreatureAI::_EnterEvadeMode(EvadeReason /*why*/)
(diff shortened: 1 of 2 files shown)


jackpoz:
Summoning @Treeston

jackpoz:
Is that the new behavior or has it always been like that ?

Treeston:
Always been like that to the best of my knowledge.

Shauren:
Except `UpdateVictim` currently does not make it enter evade mode in the last elseif branch

jackpoz:
then maybe those scripts that expect UpdateVictim() to return true when there's a victim should call something like UpdateAndCheckVictim() (or UpdateAndHasVictim() or whatever name) instead

Treeston:
...or just `if (!UpdateVictim() || !me->GetVictim()) return;`?

jackpoz:
let's keep that as new suggested pattern for scripts then

Shauren:
Scripts werent using UpdateVictim to check if there is one but as a "IsInCombat" check combined with updating the current target

Issue #21681, opened 2018-03-22
[Master] Default view distance not blizzlike


Daribon:
**Description:**
Default view distance is currently lower than it is on retail.

**Current behaviour:**
Currently the default view distance is set to 90.
Instances are set to 170 and battlegrounds/arenas are set to 533.
This is not correct.

**Expected behaviour:**
The default view distance outdoor should be 100. I no longer have the documents that truly shows this, however I'm certain that 100 yards is correct since I've tested it on retail since MoP and it always turn out to be 100 yards.
Instanced view distance should be between 300 and 400 from my tests.
My results from retail for some instances:
399.8383 -- Deadmines
299 -- Gnomeregan
299.7412 -- WC
I do lack my old testing docs on instances,  however I've been able to find my battleground docs.

From my tests, the default battleground view distance should be way higher.
Here is from Netherstorm recorded on in October 2017:
```<Npc>
    <Entry>0</Entry>
    <Name>Pajin</Name>
    <Position>
      <X>1805.883</X>
      <Y>1530.591</Y>
      <Z>1248.403</Z>
      <Type>Flying</Type>
    </Position>
    <Faction>Alliance</Faction>
    <Type>None</Type>
    <ContinentId>NetherstormBG</ContinentId>
  </Npc>

Distance: 738.4362

  <Npc>
    <Entry>0</Entry>
    <Name>Aisda</Name>
    <Position>
      <X>1795.463</X>
      <Y>1540.44</Y>
      <Z>1252.986</Z>
      <Type>Flying</Type>
    </Position>
    <Faction>Alliance</Faction>
    <Type>None</Type>
    <ContinentId>NetherstormBG</ContinentId>
  </Npc>

Distance: 746.4011
Distance2D: 746.4009
  <Npc>
    <Entry>0</Entry>
    <Name>Postulat</Name>
    <Position>
      <X>1790.371</X>
      <Y>1539.031</Y>
      <Z>1253.573</Z>
      <Type>Flying</Type>
    </Position>
    <Faction>Alliance</Faction>
    <Type>None</Type>
    <ContinentId>NetherstormBG</ContinentId>
  </Npc>

Distance: 753.1147
Distance2D: 753.1145
```
From Arathi Basin:
```
  <Npc>
    <Entry>0</Entry>
    <Name>Onimoosha</Name>
    <Position>
      <X>703.449</X>
      <Y>685.8611</Y>
      <Z>-16.33741</Z>
[...]

Issue #21111, opened 2017-12-23
[3.3.5] Missing a way to change max visibility distance of creatures/game objects


Keader:
**Description:** We need implement a way to change max visibility distance of a specific creature / game object

Something like: https://github.com/TrinityCore/TrinityCore/pull/18067

**Current behaviour:** We're stuck in default value

It need to bosses like Razorscale (3.3.5), Ragnaros (4.x), etc

**Branch(es):** 3.3.5

**TC rev. hash/commit:** d3d003d69e91fb9b019a06b0402acebae5d6d250

tkrokli:
Closed issues #9205 and #9036 -- may contain useful comments.

offl:
Afaik you can see any creature at any distance in instances.

Keader:
Yes, you can see XT002 on Ulduar entrance

Ovahlord:
"Afaik you can see any creature / object at any distance in instances / PVP locations"

This doesn't seem to be the case for instances. I have seen multiple destory object packets in dungeons for alive units

offl:
Then you can control visibility by column in some table, for each map. It was implemented for private servers 8 years ago.
I don't think someone will test real values for each map on retail.

Shauren:
Im 99% sure its a limited list of ranges and not just "any" value  - in which case i might have that list somewhere

tje3d:
and Lord Marrowgar's should be visible from entrance

offl:
It's a combined system, for example [here](https://youtu.be/NRpheQ6IPwk?t=798) you can see the boss, but you can't see normal mobs(second walking blue and two groups of mana wyrms).
But [here](https://youtu.be/aC49oVSt2oQ?t=1554) you can see everything. Also those mobs are located only at the end of instance, so their visibility can be 250. Same for other parts of this and another instances, because in most cases some mobs are located only in specific place.
Same for [Lord Marrowgar](https://youtu.be/RwRD9mzQl2E?t=9) and everything before him, but in Lord Marrowgar's case 250 is not enough :)

Shauren:
> Infinite - well... you guess
Gigantic - 400 yards
Large - 200 yards
Normal - 100 yards
Small - 50 yards
Tiny - 25 yards

Langerz82:
Does something like this work ok for you? Try and set Creatures distance_visibility to 5 (for very far), or '0' for tiny distance.

See:
[...]

PR #22891, opened 2018-12-26
Core/Objects: increase sight range of objects & correct general default visibility distance

Kittnz:
This can be used for anything really.
Note: This does not make the object active.

Some info on range
https://trinitycore.atlassian.net/wiki/spaces/tc/pages/74950924/Lua+Scene+Script+Documentation#LuaSceneScriptDocumentation-ActorAOIRange

#21111 #21681


src/server/game/Entities/GameObject/GameObjectData.h
```diff
@@ -571,6 +571,23 @@ struct GameObjectTemplate
         }
     }
 
+    bool IsLargeGameObject() const
+    {
+        switch (type)
+        {
+        case GAMEOBJECT_TYPE_BUTTON: return button.large != 0;
```
Ovahlord: tabs are missing

src/server/game/Entities/GameObject/GameObjectData.h
```diff
@@ -575,14 +575,14 @@ struct GameObjectTemplate
...
         {
-        case GAMEOBJECT_TYPE_BUTTON: return button.large != 0;
-        case GAMEOBJECT_TYPE_QUESTGIVER: return questgiver.large != 0;
-        case GAMEOBJECT_TYPE_GENERIC: return _generic.large != 0;
-        case GAMEOBJECT_TYPE_TRAP: return trap.large != 0;
-        case GAMEOBJECT_TYPE_SPELL_FOCUS: return spellFocus.large != 0;
-        case GAMEOBJECT_TYPE_GOOBER: return goober.large != 0;
-        case GAMEOBJECT_TYPE_SPELLCASTER: return spellcaster.large != 0;
-        case GAMEOBJECT_TYPE_CAPTURE_POINT: return capturePoint.large != 0;
+        case GAMEOBJECT_TYPE_BUTTON:            return button.large != 0;
```
Ovahlord: tabs are partially still missing :D TC standard is using one additional tab before the case starts. See the other helpers in that struct

illfated:
Cosmetic detail in the PR title: "it's" means "it is", whereas you probably meant to write "its" which is the possessive form, but only in the singular form.
Then again, it would be preferable to use the plural form in the title, such as:
**Core/Objects: Have all large game objects' sight range increased & correct general default visibility distance**
(the possessive form when a noun ends in a plural S is the apostrophe - "objects' ")

src/server/game/Entities/Object/ObjectDefines.h
```diff
@@ -28,7 +28,12 @@
 #define TRADE_DISTANCE              11.11f
 #define MAX_VISIBILITY_DISTANCE     SIZE_OF_GRIDS           // max distance for visible objects
 #define SIGHT_RANGE_UNIT            50.0f
-#define DEFAULT_VISIBILITY_DISTANCE 90.0f                   // default visible distance, 90 yards on continents
+#define SIGHT_GIGANTIC              400.0f
+#define SIGHT_LARGE                 200.0f
```
Faq: why exact 200?

Kittnz: https://trinitycore.atlassian.net/wiki/spaces/tc/pages/74950924/Lua+Scene+Script+Documentation#LuaSceneScriptDocumentation-ActorAOIRange

Kittnz: From official sources.

xvwyh:
Notice how "large" is the same property that got renamed to "GiganticAOI" in later expansions. That raises a question: did they just decide to increase visibility distance of all existing GOs from 200 to 400, or perhaps in WotLK times "large" actually meant 400, and "gigantic" wasn't a thing at all?

https://www.youtube.com/watch?v=P5Mbzodoxmo
Here's a WotLK video in which you can quite nicely judge what's supposed to be the distance of "large" GOs. Now how we approach this depends on whether we use 2D or 3D distance. Right after the player is launched from the cannon you can see ONE of the THREE cone-shaped tents visible at Darkmoon Faire grounds.
darkmoonfaire
That one tent seems to be at a distance of around 245 yards in 2D (275 in 3D). The other two are at a distance of about 270 (305 in 3D) and 305 (340 in 3D). I never had a reason to think blizzard uses 3D distance for visibility, correct me on this if I'm wrong. So assuming it's 2D, the "large" distance seems to actually be, I'd say, 250 yards? Or 300, on the off chance they did use 3D distance.

More videos proving the same:
https://youtu.be/ywEpt0u0Qa8?t=12
https://youtu.be/h3hfgBXT7Oc?t=8
https://youtu.be/n5kBOJogtnU?t=8

Sadly this doesn't give us a chance to prove anything about whether post-WotLK distances are indeed 200 and 400. You'd think why would they feel the need to decrease "large" from 250 to 200. But that would, however, explain why they replaced "large" GO property with "GiganticAOI": to ensure that old GOs would still be visible at the distance designers might have intended.

Ovahlord:
Gigantic should be correct. I told @Kittnz  about that already in IRC. I have been porting the AOI naming fields from master to 4.3.4 since it contains Gigantic and Infinite AOI data as well (Gilneas has some weird acting doors in sniff which confirm this) and I have also noticed that on the master branch all large fields have been replaced by Gigantic so ye. Scourgeholme in Icecrown also have a plaguespreader and a watch tower that needs to be visible via gigantic aoi. And I doubt that Blizz decided midway that they want to change it to Gigantic since 200 is clearly too less range for certain gameobjects.
wowscrnshot_122818_124543

For 4.3.4 I have quite alot of gameobjects that confirm that the large field should be Gigantic like the magic barrier at the Lost City of the Tol'Vir.
wowscrnshot_122818_124737

xvwyh:
@Ovahlord, you don't have to "prove" that GO property values changed, because they are sniffed. You don't have to "prove" that GO property definitions changed, because all the info about them (seen in the comments of GameObjectData.h) is contained within the exe even to this day. In Cata, yes, they already switched "large" property to "Gigantic AOI", so you should be looking at #22892 for anything Cata-related. This PR, however, is for WotLK, in which said property was still "large", so visibility distance could've been different from what it became on later expansions.

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -1101,6 +1110,13 @@ void ObjectMgr::LoadCreatureAddons()
             creatureAddon.emote = 0;
         }
 
+        if (creatureAddon.visibilityDistanceType >= VisibilityDistanceType::Max)
+        {
+            TC_LOG_ERROR("sql.sql", "Creature (GUID: " UI64FMTD ") has invalid visibilityDistanceType (%u) defined in `creature_addon`.",
+                guid, AsUnderlyingType(creatureAddon.visibilityDistanceType));
```
illfated: I hope you have got a plan for dealing with this feedback from travis-ci :
```cpp
/home/travis/build/TrinityCore/TrinityCore/src/server/game/Globals/ObjectMgr.cpp:1116:17:
 fatal error: 
      format specifies type 'unsigned long' but the argument has type
      'ObjectGuid::LowType' (aka 'unsigned int') [-Wformat]
                guid, AsUnderlyingType(creatureAddon.visibilityDistanceType));
~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Kittnz: fixed

illfated:
Sorry for mentioning this again, but I think the title could be improved / simplified
(just like your other PR to the `master` branch):
**Core/Objects: increase sight range of objects & correct general default visibility distance**

(Feel free to ignore this if you find it redundant.)

ghost (approved):

src/server/game/Globals/ObjectMgr.cpp
illfated: (you may mark this conversation as resolved if you want)

qyh214:
I have test and cool but maybe missing some bosses,for example:
```sql
-- ICC Lord Marrowgar 
-- ICC Lord Marrowgar 10
INSERT INTO `creature_template_addon` (`entry`, `visibilityDistanceType`) VALUES ('36612', '5'); 
-- ICC Lord Marrowgar 10h
INSERT INTO `creature_template_addon` (`entry`, `visibilityDistanceType`) VALUES ('37957', '5'); 
-- ICC Lord Marrowgar 25
INSERT INTO `creature_template_addon` (`entry`, `visibilityDistanceType`) VALUES ('37958', '5'); 
-- ICC Lord Marrowgar 25h
INSERT INTO `creature_template_addon` (`entry`, `visibilityDistanceType`) VALUES ('37959', '5'); 
```

qyh214 (approved):

Kittnz:
If you want to add it, add it to creature_addon. The entrys that re in creature_template_addon are sniffed. Shouldn't be changed unless we have more details.

PR #22892, opened 2018-12-27
Core/Objects: increase sight range of all Infinite, Gigantic and Large objects & correct general default visibility distance

Kittnz:
Note: This does not make the object active.

Some info on range
https://trinitycore.atlassian.net/wiki/spaces/tc/pages/74950924/Lua+Scene+Script+Documentation#LuaSceneScriptDocumentation-ActorAOIRange

#21111 #21681 #22891


Palabola:
Confirmed, this should be work something like this.

illfated:
Regarding the PR title, like I mentioned in your other PR: "it's" means "it is", but maybe the title could be simplified to say:
**Core/Objects: increase sight range of all Infinite, Gigantic and Large game objects & correct general default visibility distance**

Faq:
me summons @xvwyh for some comments on those magic numbers 200, 100 50  etc.

Wolffenstein:
The values may need adjustments. Gateways Murketh/Shaadraz in Hellfire Peninsula would have 400yd sight range according to these changes, but they should have at least 500-600+ from what I remember on retail (400yd is definitely not enough considering you should see them immediately after going through The Dark Portal). 

Perhaps it needs to be multiplied by object's size (a pure guess based on the name of GameObject 179066 - "Smoke Emitter, Large AOI, **scale 2**").

Kittnz:
Please check out https://github.com/TrinityCore/TrinityCore/issues/21111 for more details for ranges.
600+ is not possible.. #define SIZE_OF_GRIDS            533.3333f

xvwyh:
@Wolffenstein, there's a slight problem with using these Gateways as an example, because as of the recent expansions they are now entirely clientside. Many if not all GAMEOBJECT_TYPE_GENERIC objects have been moved to GameObjects.db2 since they're just static uninteractible and unchanging spawns, so blizzard decided to cut on traffic and let the client display them on its own. On the positive side - if you were to dig into modern clients, you could discover how they handle these large/gigantic/infinite properties and what AOI does it assign to them. But on the flip side - whatever memories you have about them need to be prefaced by what expansion were the memories from.

xvwyh:
@Kittnz, but are you sure that the hardcap on notifier visit distance wasn't invented by private server developers? I've found a few occasions where it's just wrong. Many spells that are meant to hit all players map-wide (in dungeons) with 50000 radius straight up don't work properly. And there's at least one case where a spell in Throne of the Four Winds with NEARBY target is supposed to target a trigger that's more than 533 yards away.

I preferred not to remove the cap (although I really wanted to) to prevent these spells with 50000 from hanging the server, but I did make an exception for visibility and update notifiers, so I could have spawns with more than 533 visibility distance (invisible vehicle creatures on Icecrown transports that draw a vehicle icon on the world map no matter where in Northrend you are, Fel Reaver in Outland being visible from the other side of Hellfire Peninsula, etc).

Wolffenstein:
@xvwyh That's very interesting, I didn't know that! I am positive the gateways were visible back in the Burning Crusade days as well though as I have seen videos of them, they should still be usable as an example.

Shauren:
@xvwyh that hard cap is definitely for performance reasons to protect against stupid scripts doing stupid things

Kittnz:
@xvwyh Yes, i also think they just invented that size.
But we don't handle grids the same way as blizz does it.
From my tests they have zone wide size of grid.

Warpten:
Isn't there some Sha of whatever visible on the entire map in pandaria?

Kittnz:
Only zone wide, not the whole map.

Keader:
Sha of Anger, not map visible

xvwyh:
Might still be of "infinite" AOI, but because each zone is served by a different server (see CRZ) it would make an impression of only being visible in said zone.

Wolffenstein:
I remember you could select a Whale Shark (or some other Giant mob) from Vashj'ir even if you were standing in Stormwind.

Shauren:
Oh, you could select the shark all the way from Undercity

Ovahlord:
So the question that remains for creatures now is: how shall waypoints be handled this way? Naming the Fel Reaver on Hellfire Peninsula or the other one on Argus who can be seen from far away and also have waypoints. If the grid is not getting loaded when those high distance creatures show up they will just stand arround idle until someone gets close enough to load the grid.

xvwyh:
@Ovahlord https://github.com/TrinityCore/TrinityCore/pull/22268

Kittnz:
Out of scope of this PR tho Elon

xvwyh:
@Ovahlord
> which will increase the startup time by alot.

No shit, this isn't a small indie game world we're talking about but a giant MMO. Of course it can't, and shouldn't, boot in a second.

> I would prefer setting active creatures via database so only the necessary grids are being loaded instead of having all of them loaded at once.

This is actually what I did. I made "active object" to be a a field of flags instead of a boolean, so I could toggle different reasons why an object would be active or not and not have it interfere with others. So I can manually make certain creature and gameobject templates be active all the time, or make them active because they're using waypoints, or when they're on transport, or if they're a source of an area aura etc. It serves me quite well so far.

```cpp
// Reasons for why object was flagged as active
enum class ActiveFlags : uint32
{
    InCombat                = 0x00000001, // Unit is active due to being in combat and needing to evade back to home position (Default reason for backwards compatibility)
    Script                  = 0x00000002, // General purpose script reason
    ScriptPassive           = 0x00000004, // General purpose script reason (skipped if no players are in the map, use this primarily in instances)
    IsWorldObject           = 0x00000008, // Object is active due to being a world object
    SharedVision            = 0x00000010, // Unit is active because a player is sharing its vision (i.e. Mind Vision spell)
    EAI                     = 0x00000020, // Object was flagged as active from an EventAI script
    SAI                     = 0x00000040, // Object was flagged as active from a SmartAI script
    OnTransport             = 0x00000080, // Object is active due to being on a MapObject transport
    AuraGenerator           = 0x00000100, // Object is active because it emits an area aura which needs to eventually be removed when targets are out of range
    CustomVisibility        = 0x00000200, // Object was flagged as active in `object_visibility` table
    SplineWaypoints         = 0x00000400, // Object is active due to using a WaypointMovementGenerator with spline path which needs to be continuously updated to match what clients see

    // If object has ONLY any of these flags set - it will not trigger grid updates when there are no players in the map
    OnlyInNonEmptyMapsMask  = ScriptPassive | CustomVisibility | SplineWaypoints,

    None = 0,
};
```

Kittnz:
Thats the thing, tc isn't production software. It's for testing and learning. However we should have the ability to have these features.

Ovahlord:
Throwing in a small test report. So far the functionality itself seems to be just fine. There is a strange bug though. Sometimes it happens that all objects with a custom visbility are ALL turning invisible when they should show up. I did a test by setting the visbility range of the Fel Reaver to Infinite and Giant and in 3 out of 4 tests the Fel Reaver was invisible and other objects that had custom visbilities were invisible too like the three Capture Point gameobjects on Hellfire Peninsula.
This did not happen when the override variable was not added yet so I suspect this somehow being involved in that.

xvwyh:
I might know of a bug in TC's custom visibility implementation. Do me a favor. Put one character far away from a creature with large visibility distance that's standing still. If you can see the creature - try moving your character some distance away from their previous position. Say, 10 yards. The creature might disappear. Now take another character in a second client instance, go to the creature and cause it to move by e.g. engaging in combat. After the creature moves, the first character will start seeing the creature as long as the character doesn't move far from their current position. Report back if that's what you're observing.

xvwyh:
Oh, this PR **IS** the custom visibility distance implementation... I thought I remembered seeing commits already implementing it in the past... I'll write that off as hallucinations, I guess...

OK, so here goes.

Simply changing `GetVisibilityRange()`, alas, is not enough. The problem is - it's not always the creatures/GOs themselves that trigger visibility updates, but player can do that too. Go see how `Trinity::VisibleNotifier` is implemented. It takes a list of GUIDs already visible to the client (`VisibleNotifier::vis_guids`), visits all nearby grid cells (at PLAYER'S sight range, which would be the same 100 yards) and for every object it encounters it removed it's GUID from that list. At the end, when `SendToSelf()` is called, every GUID that's still present in the list is considered to have become invisible to the player. The logic is sound - we didn't find the object when visiting nearby grid cells, means it vanished - moved outside of the range, to a different phase, became invisible or stealthed, or despawned altogether. But here's where its behavior interferes with custom visibility - creatures with higher than 100 yards visibility distances can be in cells farther away from the player than his sight range with reach, but should still be visible. But because they weren't encountered while visiting cells-at-100-yards-range - the server decides they are no longer visible to the player. If the object moves, however, it triggers a visibility update of self for players in nearby cell at OBJECT'S sight range (calls `Player::UpdateVisibilityOf` for every encountered player, which succeeds, because the object is indeed at a suitable distance from the player for the latter to be inside of its visibility distance), and because of that, if the object moves a bit, it becomes visible to players again.

There's no easy and great solution. The best way to fix this would be to have multiple grid containers each corresponding to a separate AOI range. So when visibility needs to be updated, `Trinity::VisibleNotifier` can visit "Normal" grids with 100 radius, then "Large" grids with 200 radius, "Gigantic" grids with 400 radius and, well, all the "Infinite" grids. "Infinite" would probably be better implemented not as a grid, but as a simple list/set/vector, since it will always need to be traversed entirely. This will inevitably be hard to implement, because then creatures would need to keep track of which cell they are in for multiple grids at once, swap between them when when moving, and we got enough hassle managing one grid already.

I opted for a more hacky solution that sort of works and doesn't require reengineering the entire visibility system, but isn't ideal for performance. It still involves new containers, but they are plain-old `set`s, although even a `vector` would do, and probably be better performance-wise, but not convenience-wise. I have one container for map-wide visibility, and additional containers for each zone on the map. When `Trinity::VisibleNotifier` is used, after `VisitAllObjects` I also additionally visit these containers, the map-wide one, and the one corresponding to the zone the player is in. By "visit" in this case I do mean call `UpdateVisibilityOf` for EVERY object in those containers, so you can see how this isn't too great for performance, because of the lack of spacial partitioning. That way distant objects are visited as well, their GUIDs are removed from `vis_guids` and thus the server doesn't think they are out of range and doesn't hide them from the client. Which zone container the objects will get added to depends on where they spawn, and moving between zones doesn't move them to a different container, so this solution is also limited in this regard. And whether the object will be added to map-wide or zone-specific containers I can control independently, so if there's a big fancy object located near the zone boundary that I want to be visible from adjacent zones as well, I can flag it to be added to the map-wide container instead.

Same applies to 3.3.5 PR as well, of course. Same visibility system in both branches.

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2063,6 +2082,8 @@ float WorldObject::GetVisibilityRange() const
 {
     if (isActiveObject() && !ToPlayer())
         return MAX_VISIBILITY_DISTANCE;
+    else if (IsVisibilityOverride() && !ToPlayer())
+        return *m_visibilityDistanceOverride;
```
xvwyh: Why not make the override have priority over `MAX_VISIBILITY_DISTANCE` for `isActiveObject()`? Why do active objects have different visibility distance in the first place???

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2075,6 +2096,8 @@ float WorldObject::GetSightRange(const WorldObject* target) const
         {
             if (target && target->isActiveObject() && !target->ToPlayer())
                 return MAX_VISIBILITY_DISTANCE;
+            else if (target && target->IsVisibilityOverride() && !target->ToPlayer())
+                return *m_visibilityDistanceOverride;
```
xvwyh: Same here.

robinsch:
@xvwyh Is this zone based container thing a CRZ thing? Why don't you have them map based or do you visit nearby zones? I basicly use the same implementation. Vector for each AIO type and then just visit above 533y containers in VisibleNotifier. I remember having some issues with ghosts gameobjects and player pets when using the above way with client GUIDs, so I split UpdateVisibilityOf to two functions that update visibility and out of range GUIDs (you can be a bit lazy on the latter for performance gains)

src/server/game/Entities/Object/Object.cpp
Ovahlord: That's a missing cherrypick from 335. In 335 they have splitted active objects and far visible objects. So instead of isActiveObject you would have IsFarVisible here which is basically the previous version of InfiniteAOI

Shauren:
Thats also the cause of that "weird" disappearing LK platform reported somewhere else (that whole visit 100 from player)

Ovahlord:
I also had some funny issues with raid platforms like in Throne of the Four Winds like the platform turns invisible when going .gm on or when druids do shapeshifting etc. when outside of the regular visbility range while using FarVisible / setActive. (It's also a destructible gameobject just like the LK platform)

xvwyh:
@robinsch, it makes an impression of CRZ if you notice it, but I didn't intend to replicate CRZ, just to optimize it in some way. Here are some stats from a running server, in Northrend there are currently:
- 132 map-wide objects
- 449 in Icecrown
- 439 in Storm Peaks
- 86 in Zul'Drak
- 162 in Dragonblight
- 28 in Crystalsong Forest
- 167 in Grizzly Hills
- 138 in Howling Fjord
- 248 in Borean Tundra
- 48 in Wintergrasp
- 117 in Sholazar Basin
You can imagine how visiting just 218 objects when you're in Zul'Drak, for example, is better for performance than visiting all 2014. Once I implemented the visibility, I measured performance impact, found it to be negligible, and went complete nuts with increasing visibility. Flew around every map looking for big creatures and GOs and increasing distance for them to what I deemed to be acceptable (that was before I realized "large" GO property should be responsible for that), even going as far as making regular Kodos visible at 200 yd.

@Ovahlord, wow, that's a... not good actually... Something is very very wrong if even nearby objects are disappearing...

xvwyh:
@Kinzcool 
> This is why I think that it would be better to "scavenge" all the data present in this table that would still matter for the master branch (and there should be a lot).

I would personally go the opposite way. Get rid of everything that can be replaced with this data and use the data instead, adapting all the existing code to work with it instead. Yes, it's a long endeavor which I doubt I'll even do myself, and yes, it might be wrong for pre-MoP expansions, or missing for post-MoP, but in that case we can edit the data manually if we have sufficient proof, or create temporary fake entries (like those in `gossip_menu`, `npc_text` with high IDs) until official data is available.

> About unit_flags, the thing is that it seems that there are static and dynamic flags in that field.

No, everything in `unit_flags` is changeable at run time. That's why those flags are in an UpdateFields, instead of being part of static cache, like `type_flags`. And if by "dynamic" you mean "different per-player", then that's wrong as well: only the `UNIT_DYNAMIC_FLAGS` and other fields marked as `DYNAMIC` contains data that's constructed per-player. `UNIT_FIELD_FLAGS` is not one of them. `dynamic_flags` isn't needed in the DB at all, its only use (for me at least) it to know if a creature ever had "tapped by all threat-list" flag, but whether or not the creature is even subject to such behavior is determined by one of the CreatureDifficultyFlags. `InhabitType` is most definitely not needed, "Aquatic", "Amphibious", "Sessile" flags (and the eventual "flying" or "hovering" flags that are yet to be discovered) make up for it in its entirety.

Shauren:
`InhabitType` was already replaced in 3.3.5, you should take a look at it

xvwyh:
@Shauren oh nice, yet another table that will be made obsolete by CreatureDifficultyFlags then! 😄

Kinzcool:
@xvwyh 
Trust me, it would make things a lot more confusing that way. All sniff data is composed of "classic" type_flags and unit_flags. You would then have to convert those so that they would fit CD data in the DB. Basically, one would have to deal with two flag values representing the same thing which adds needless confusion.

Also, by dynamic_flags, I meant flags that can vary on a per-creature basis (from a spawned NPC to another of the same ID). At least, that's how I recall dynamic_flags to be.

There are indeed "flying" flags. I was just not sure what they represented exactly related to movement flags and to data already in the DB.

Kinzcool:
Oh, and another problem of using directly CD would be that people would have to update its flags fields with non-blizzlike data. (E.g., someone would have to make a creature have X InhabitType, so they would change data in CD.) This non-blizzlike data would be along blizzlike flag data such as type_flags.

xvwyh:
@Kinzcool then let's agree to disagree. I value the pursuit of using blizzard's data more than convenience. But I'm also not the one working on TC, so my personal feeling won't impact your development in any way.

> Also, by dynamic_flags, I meant flags that can vary on a per-creature basis (from a spawned NPC to another of the same ID). At least, that's how I recall dynamic_flags to be.

Errr, npc_flags do vary on per-creature basis. A boss can make itself ImmuneToPC during combat, or Unselectable. Not to mention Stunned, Fleeing etc also being unit_flags that change during gameplay. But a tamable creature cannot make itself not tamable, because THAT flag is in type_flags, which are static.

`dynamic_flags` are a field that can contain different data depending on who's looking at it. If two players are observing a corpse of a creature that was killed by one of them, then the killer would see "lootable" in dynamic flags field, but the other player would not. And thus the creature would only be lootable by the killer. All the other fields that aren't `DYNAMIC` as defined in UpdateFields.h, don't have such functionality. Or, rather, shouldn't. We do it all the time on private servers, by swapping displayIds for creatures depending if the player is in GM mode or not, for example. But officially only `DYNAMIC` fields support such behavior.

> Oh, and another problem of using directly CD would be that people would have to update its flags fields with non-blizzlike data. (E.g., someone would have to make a creature have X InhabitType, so they would change data in CD.) This non-blizzlike data would be along blizzlike flag data such as type_flags.

Welp. Can't do anything about that. It would be nice to have two of each table, one containing official data, and the other containing custom overrides, but that wouldn't be terribly convenient, nor would many users actually observe such rules (does anyone bother making `VerifiedBuild` value negative when they edit something in DB? I bet almost nobody).

illfated:
> IIRC, the data is from patch 5.4.8., so be cautious if you want to use it for another patch.

If that CreatureDifficulty data is from the same time as the current TC 3.3.5 `broadcast_text` bruteforce data, it is from build 18019 (https://wow.gamepedia.com/Patch_5.4.7) - if you wanted to know.

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -1162,6 +1171,13 @@ void ObjectMgr::LoadCreatureAddons()
             creatureAddon.meleeAnimKit = 0;
         }
 
+        if (creatureAddon.visibilityDistanceType >= VisibilityDistanceType::Max)
+        {
+            TC_LOG_ERROR("sql.sql", "Creature (GUID: " UI64FMTD ") has invalid visibilityDistanceType (%u) defined in `creature_addon`.",
```
Ovahlord: %u is missing here too

Kittnz: There is nothing missing here

Kittnz: they do? https://github.com/TrinityCore/TrinityCore/blob/a78a85e8b8b5cf02df486c776fff578a153a9906/src/server/game/Globals/ObjectMgr.cpp#L1170

PR #22902, opened 2019-01-01
[3.3.5] Entities/Unit: Targeting revamp (AI refactor prep part 1)

Treeston:
The first batch of refactors to keep #22461 manageable.

At its core, we split the current targeting apart somewhat - from two targets (auto-attack victim and currently selected unit) to three (auto-attack victim, primary target, and currently selected unit).

As a consequence, I also cleaned up function signatures, and made spell focusing less ugly.

## Selected unit 
* `GetTarget` -> `GetSelectedUnitGUID`
* MOVED `GetSelectedUnit` (from `Player` to `Unit`)
* MOVED `GetSelectedPlayer` (from `Player` to `Unit`)
* `SetTarget` REMOVED
* ADDED `UpdateSelectedUnit`

## Primary target
* ADDED `SetPrimaryTarget`
* ADDED `StopTargetingMe`
* ADDED `GetPrimaryTarget`

## Auto-attack victim
* getAttackerForHelper -> GetAttackerForHelper
* `Attack` -> `AutoAttackStart`
* `AttackStop` -> `AutoAttackStop`
* `GetVictim` -> `GetAutoAttackVictim`
* `EnsureVictim` REMOVED
* `RemoveAllAttackers` -> `StopAutoAttackingMe`
* `ValidateAttackersAndOwnTarget` -> `ValidateAutoAttackStates`

## Spell focus system
* `IsFocusing` -> `HasSpellFocusTarget`
* ADDED `GetSpellFocusTarget`
* `FocusTarget` -> `SetSpellFocus`
* `ReleaseFocus` -> `ReleaseSpellFocus`
* ADDED `ClearSpellFocus`
* REMOVED `MustReacquireTarget`
* REMOVED `DoNotReacquireTarget`

### DIFF
--- a/src/server/game/AI/CoreAI/CombatAI.cpp
+++ b/src/server/game/AI/CoreAI/CombatAI.cpp
@@ -156,7 +156,7 @@ void CasterAI::UpdateAI(uint32 diff)
 
     events.Update(diff);
 
-    if (me->GetVictim() && me->EnsureVictim()->HasBreakableByDamageCrowdControlAura(me))
+    if (me->GetAutoAttackVictim() && me->GetAutoAttackVictim()->HasBreakableByDamageCrowdControlAura(me))
     {
         me->InterruptNonMeleeSpells(false);
         return;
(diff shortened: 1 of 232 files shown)


jackpoz:
EnsureVictim() was added to handle some static analysis warnings about GetVictim() being dereferenced and to trigger an assert instead of a null deref crash. I hope those static analysis warnings won't get reopened.

Treeston:
If they are, they should be gracefully handled in the code. Adding asserts just masks the problem.

Treeston:
Also, I am not happy with the state of this yet, upon further thought.

I want players to use just the GUID field while player-controlled, and reserve primary victim to server-side control (and also tie it to combat state). That'll let me clean up a bunch of logic all over the place.

PS: Something I noticed. If a player possesses a creature, and then targets something, should the creature's displayed target change, or should the player's? Right now it's the latter, and I suspect it should be the former.

Treeston:
Fixed the above. Targeting now changes both own target + moved unit's target as appropriate.

PR #22904, opened 2019-01-02
Core/Creatures: Profession trainer fixes

### LINKED ISSUE #22903: Professions
**Description:**

For some reason I can learn more than 2 primary professions.

**Current behaviour:**

I just went and learned Alchemy and Herbalism and for some reason it'll let me learn Tailoring, Blacksmithing and Engineering. The whole lot pretty much.

**Expected behaviour:**

Should not be able to have more than 2 professions learned unless of course you change the setting from 2 to whatever in the config.

1. Go to any profession trainer and you should (well I can) learn all the primary professions.

**Branch(es):**

3.3.5a

**TC rev. hash/commit:** 

TrinityCore rev. 66720e5fd940 2018-12-31 11:50:16 +0100 (3.3.5 branch) (Win64, RelWithDebInfo, Static) (worldserver-daemon)

**Operating system:** Windows Server 2016

Ghaster:
**Changes proposed:**

- Correctly check TriggerSpell for professions instead of ReqAbility.
- Fixed an issue, where profession ranks counted towards the max profession amount.

- [x] 3.3.5

**Issues addressed:** Closes #22903 

**Tests performed:** Builds, tested by attempting to train more than two professions, along with training next rank of a profession.

- Cleanup?

### DIFF
--- a/src/server/game/Entities/Creature/Trainer.cpp
+++ b/src/server/game/Entities/Creature/Trainer.cpp
@@ -48,10 +48,17 @@ namespace Trainer
             if (!player->IsSpellFitByClassAndRace(trainerSpell.SpellId))
                 continue;
 
+            SpellInfo const* trainerSpellInfo = sSpellMgr->AssertSpellInfo(trainerSpell.SpellId);
+            if (!trainerSpellInfo)
+                continue;
+
             bool primaryProfessionFirstRank = false;
-            for (int32 reqAbility : trainerSpell.ReqAbility)
+            for (uint8 i = 0; i < MAX_SPELL_EFFECTS; ++i)
             {
-                SpellInfo const* learnedSpellInfo = sSpellMgr->GetSpellInfo(reqAbility);
+                if (trainerSpellInfo->Effects[i].Effect != SPELL_EFFECT_LEARN_SPELL)
+                    continue;
+
+                SpellInfo const* learnedSpellInfo = sSpellMgr->GetSpellInfo(trainerSpellInfo->Effects[i].TriggerSpell);
                 if (learnedSpellInfo && learnedSpellInfo->IsPrimaryProfessionFirstRank())
                     primaryProfessionFirstRank = true;
             }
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Creature/Trainer.cpp
```diff
@@ -48,10 +48,17 @@ namespace Trainer
             if (!player->IsSpellFitByClassAndRace(trainerSpell.SpellId))
                 continue;
 
+            SpellInfo const* trainerSpellInfo = sSpellMgr->AssertSpellInfo(trainerSpell.SpellId);
+            if (!trainerSpellInfo)
```
jackpoz: AssertSpellInfo() is supposed to never return null

src/server/game/Entities/Creature/Trainer.cpp
```diff
@@ -132,9 +139,19 @@ namespace Trainer
             return false;
 
         SpellInfo const* trainerSpellInfo = sSpellMgr->AssertSpellInfo(trainerSpell->SpellId);
-        if (trainerSpellInfo->IsPrimaryProfessionFirstRank() && !player->GetFreePrimaryProfessionPoints())
+        if (!trainerSpellInfo)
```
jackpoz: AssertSpellInfo() is supposed to never return null (here too)

jackpoz (changes requested):

Treeston:
Could do with some more explanation. What logic error does this address?

`ReqAbility` is what? What is the changed check supposed to test for, and what did it (incorrectly) test for instead?

Changes look correct to me, but could do with elaboration.

Ghaster:
When adding a spell to a trainer, one can add up to 3 spells, that the player needs to have, before they can learn that spell. That is what `ReqAbility` is for.

Before the trainer revamp, `ReqAbility` was not stored anywhere, but was rather generated by the core from the SpellID. An issue was that the `ReqAbility` that was stored in the core, was not the same `ReqAbility` that the server would send to the client (as seen here): https://github.com/TrinityCore/TrinityCore/blob/59e94fd9bb1ff82d9a34a6c693b680039c3a68b9/src/server/game/Handlers/NPCHandler.cpp#L199-L223

With the trainer revamp, this has now been fixed, as `ReqAbility` is now stored alongside all the other requirements. However, some of the checks related to making sure the player does not learn more than two professions, were not updated to take advantage of the "new" `ReqAbility` values. This PR fixes that.

LewisPotato:
How do I know when this fix has been implemented? Cause I plan on recompiling when the issue is fixed.

msoky:
> How do I know when this fix has been implemented? Cause I plan on recompiling when the issue is fixed.

it will be there https://github.com/TrinityCore/TrinityCore/commits/3.3.5

src/server/game/Entities/Creature/Trainer.cpp
Shauren: This is impossible to fail (Assert would be correct) - thanks to startup checks

jackpoz (approved):

Treeston (approved):

PR #22909, opened 2019-01-05
Add quick events for easier event handling

jameyboor:
**Changes proposed:**
Add another container that plays nicely with lambdas to ease event development.

Instead of having to create a whole seperate class that derives from `BasicEvent` this addition will allow simple future actions to be added by just adding a lambda to the EventProcessor instead.

**Target branch(es):** 3.3.5

- [x] 3.3.5

### DIFF
--- a/src/common/Utilities/EventProcessor.cpp
+++ b/src/common/Utilities/EventProcessor.cpp
@@ -78,6 +78,17 @@ void EventProcessor::Update(uint32 p_time)
         // the next update tick
         AddEvent(event, CalculateTime(1), false);
     }
+
+    auto itr = m_quickEvents.begin();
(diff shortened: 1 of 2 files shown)


Ovahlord:
While you're at it add an 2nd header entry with std::chrono timers just like AddEventAtOffset does.

Keader:
I love it

jackpoz:
I see the way the m_quickEvents are iterated is totally different from m_events, would it make more sense to have the same for both ?

Shauren:
Implemented differently in 813f693768ee94864b0d24423b9265bde2c9903d

jameyboor:
@Shauren Oh you must be feeling really cool right now. HMMMMMMMMM

Issue #22917, opened 2019-01-07
[Crash] spell_pri_power_word_shield


HAANAAE:
Description: I do not know, crash log a lot ...

Crash log / 33.3MB

https://ufile.io/ujall

Branch(es): 3.3.5
TC rev. hash/commit: https://github.com/TrinityCore/TrinityCore/commit/3b5014fdae1b57add7fd204a9f9f10f3c2092b83
Operating system: debian

Aokromes:
use https://gist.github.com/ instead that shitty hosting plz.

Aokromes:
+ also compile on debug mode or releasewithdbg

HAANAAE:
ican't upload gist ! error not allowed , and on issues

You can't comment at this time — your comment is too long (maximum is 65536 characters).

jackpoz:
This is an infinite loop triggered by spell_pri_power_word_shield script

Keader:
Wait, caster and target has the shield
Damn it

jackpoz:
Please note Reflective Shield talent https://wotlk.evowow.com/?spell=33201 is required to trigged the issue

Keader:
Fixed by 545bee04961450465a851196412f56b5578d9837

PR #22918, opened 2019-01-07
Scripts/Spells: Fix infinite loop in Power Word Shield

jackpoz:
Fix infinite loop in Power Word Shield triggering back and forth the damage spell between 2 players, both with Power Word Shield and Reflective Shield talent https://wotlk.evowow.com/?spell=33201

**Changes proposed:**

-  Don't try to reflect the reflecting spell  

- [x] 3.3.5
- [ ] master

Closes https://github.com/TrinityCore/TrinityCore/issues/22917

None

- [x] Need to repro the issue to confirm the steps
- [x] Need to test ingame that the fix works

### DIFF
--- a/src/server/scripts/Spells/spell_priest.cpp
+++ b/src/server/scripts/Spells/spell_priest.cpp
@@ -1029,6 +1029,10 @@ class spell_pri_power_word_shield : public SpellScriptLoader
                 if (dmgInfo.GetAttacker() == target)
                     return;
 
+                // Don't try to reflect the reflect spell
+                if (dmgInfo.GetSpellInfo() && dmgInfo.GetSpellInfo()->Id == SPELL_PRIEST_REFLECTIVE_SHIELD_TRIGGERED)
+                    return;
+
                 if (AuraEffect* talentAurEff = target->GetAuraEffectOfRankedSpell(SPELL_PRIEST_REFLECTIVE_SHIELD_R1, EFFECT_0))
                 {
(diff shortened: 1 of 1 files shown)


Keader:
Why it's trying reflect self in first place? o.O

jackpoz:
because 2 players have Power Word Shield both:
- Player A casts a damaging spell to Player B
- Player B absords damage with Power Word Shield and casts back the reflective spell to Player A
- Player A absords damage with Power Word Shield and casts back the reflective spell to Player B
- on and on and on

jackpoz:
A better fix would be for spell SPELL_PRIEST_REFLECTIVE_SHIELD_TRIGGERED to have a flag not to be able to be reflected. Which other spells can reflect damage back ?

Keader:
But if both is getting reflect damage, the shield should break no?

jackpoz:
the triggered spell is cast in AfterEffectAbsorb which is before the actual damage has been done so no, the shield doesn't break

Keader:
So fix looks correct to me.

ccrs:
I dont know whats the blizzlike behaviour, but this is a good countermeasure

Keader (approved):

jackpoz:
it seems like the absorbed damage gets weaker and weaker at every call (like 80, then 36, then 16, etc) and at 0 it stops the recursive call so the issue is not always reproduceable

Keader:
Yeap, but big amount of spell power can make it take a while to happen xD

PR #22936, opened 2019-01-11
Scripts/Quest: make Wounded Skirmisher reply talk show class & race

ghost:
Daily PvP quest 12288: Overwhelmed!
- https://www.wowhead.com/quest=12288/overwhelmed
- Upon finishing the healing quest spell, Wounded Skirmisher (npc_wounded_skirmisher) talks in gratitude for the healing, but shows **$r** and **$c** in their text. This is because of a missing target argument. hence the Wounded Skirmisher is not targeting the player on the quest (or in this case, the specific caster).

Text used by the NPC **Wounded Skirmisher**:
```sql
SELECT * FROM `creature_text` WHERE `CreatureID`=27463;
```
```
CreatureID  GroupID      ID  Text                                          Type  Language  Probability   Emote  Duration   Sound  BroadcastTextId  TextRange  comment
----------  -------  ------  ------------------------------------------  ------  --------  -----------  ------  --------  ------  ---------------  ---------  ---------------------------------
     27463        0       0  Ahh.... better.                                 12         0          100       0         0       0            26657          0  wounded skirmisher RANDOM_SAY_1
     27463        0       1  Whoa - I nearly died there. Thank you, $r!      12         0          100       0         0       0            26659          0  wounded skirmisher RANDOM_SAY_2
     27463        0       2  Thank you, $c.                                  12         0          100       0         0       0            26656          0  wounded skirmisher RANDOM_SAY_3
```

**Changes proposed:**

-  Add the missing target name, **caster**, to the `Talk()` call arguments.

- [X] 3.3.5
- [X] master

None that I could find. I discovered this issue last night, but I already knew what the problem & solution would be.

Tested in-game, works as intended & expected.

- [ ] if you try to attack the wounded Westfall infantry soldiers (targets of the Alliance version of the same quest), they say the same gratitude text lines as the Horde NPCs, evade and start running, then despawn at the end of the run. I don't think this should happen, but I also do not know a simple solution to this separate issue (outside the scope of this PR). I have not tested the Alliance version of this quest, but it is not part of `zone_grizzly_hills.cpp` and probably handled by DB script (SAI).
- [x] ~~The healing spell does not cause a visible healing to the "wounded" NPCs, but this is probably a combination of the fact that the script should set the Out-Of-Combat HP of the wounded NPCs to much lower than 100% health, as well as how the spell is handled. (This is mostly visual and also outside the scope of this PR.)~~ Reading the wowhead comments, it looks like this is how it actually worked on retail. :-P

### DIFF
--- a/src/server/scripts/Northrend/zone_grizzly_hills.cpp
+++ b/src/server/scripts/Northrend/zone_grizzly_hills.cpp
@@ -501,7 +501,7 @@ public:
                 && caster->ToPlayer()->GetQuestStatus(QUEST_OVERWHELMED) == QUEST_STATUS_INCOMPLETE)
             {
                 DoCast(caster, SPELL_KILL_CREDIT);
-                Talk(SAY_RANDOM);
+                Talk(SAY_RANDOM, caster);
                 if (me->IsStandState())
                     me->GetMotionMaster()->MovePoint(1, me->GetPositionX()+7, me->GetPositionY()+7, me->GetPositionZ());
                 else


Aokromes:
I wonder how much Talk( needs that parameter.

illfated:
Good question, although it is not always evident from just reading the code (in my experience).
Most of the time, you don't discover it before you are in the scripted scenario.

illfated:
Well, thinking about that question a bit more, it could in theory be automated if the code checking for the missing parameter would be able to compare every script using `Talk(SAY_*) ` , looking for the NPC/creature ID (`creature_template.entry` & `creature_text.CreatureID`), then checking that DB text for occurrences of `$r`/`$c`/`$n` in the text associated with said creature. Maybe it needs to be the other way around, checking `creature_text` for occurrences of `$r`/`$c`/`$n` and find out if it is used directly in a C++ script with `Talk()`. Some of the challenges would be to separate what is gossip and what is the `creature_text` values.

I am not sure what that C++ code would look like and if it needs to be incorporated or just used as a separate tool on the side (like whitespace lookup).

Issue #21872, opened 2018-04-22
[MMAP] NPC movement hiccups


Toomaa:
**Description:**

NPC's movement is strange:

1. moving over obstacles (objects too high to walk over)
2. moving up/down (hiccupping) inside houses, caves, instances (wmo's ?)
3. moving up/down (hiccupping) when underwater

**Current behaviour:**

Incorrect movement. I'm pretty sure this all worked correctly in one of he previous revisions - not sure which (~month ago)

**Expected behaviour:**

1. NPC's should NOT walk on top of / over obstacles when their movement is set as random (not sure what should happen for waypoint movement)
2. NPC's shouldn't hiccup when stepping from terrain to house and vice versa. Narrow places cause NPC's to move into the ceiling...
3. underwater/water NPC's should swim or walk on the ocean floor (like crabs) 

**Steps to reproduce the problem:**

1. A great example for this bug(s) is Darkshire. Observe various moving NPC's there:
Town Crier (entry 468) for example, is showing first two symptoms: [https://gph.is/2qTpgLI](https://gph.is/2qTpgLI) 
Watcher Jordan (entry 887) moves over the fountain.
Night Watch Guard (entry 10038, spawn id: 6133) gets stuck in house wall above the ground (at one point he disappeared inside the wall completely).
2. Various crabs at Zoram Strand (and other underwater places): [https://gph.is/2F7V0So](https://gph.is/2F7V0So)
Blackfathom Deeps' underwater parts exhibit all these problems. Try attacking one of the murlocs/nagas and swim underwater and above the water.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. f1ab32fbb7d5 2018-04-22 17:22:26 +0200 (3.3.5 branch) (Win64, RelWithDebInfo, Static)
Using SSL version: OpenSSL 1.0.1u  22 Sep 2016 (library: OpenSSL 1.0.2g  1 Mar 2016)
Using Boost version: 1.60.0
Using MySQL version: 5.5.58
Using CMake version: 3.10.2
Compiled on: Windows 10.0.16299
Automatic database updates are enabled for the following databases: Auth, Characters, World
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in ./maps. Total size: 252191207 bytes
[...]

PR #22937, opened 2019-01-12
Core/Movement: Fix some undermap issues with random movement/fear/blink

Golrag:
**Changes proposed:**

- Only move to point if there is a path that is not a shortcut (which will make the unit move through terrain)
- ~Added new function to check if there is a vmap floor without search distance~
- Units that can fly, are underground but far above the vmap floor will stay underground (bronze drakes in tanaris)
- Don't remove PATHFIND_SHORTCUT from path type in some cases

**Target branch(es):**

- 3.3.5

**Issues addressed:** Updates #21255

**Tests performed:**
- Builds
- Blinking around in WSG
- Fearing in WSG
- Fearing in Icecrown (mountains in the centre)
- Blinking in Icecrown

- ~Unsure about performance. I needed to check if there was a vmap way lower than the default search distance would let me for the drakes in tanaris/cot~

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -3308,8 +3308,8 @@ void WorldObject::MovePositionToFirstCollision(Position &pos, float dist, float 
     {
         if (Unit const* unit = ToUnit())
         {
-            // flying, ignore.
-            if (unit->IsFlying())
+            // unit can fly, ignore.
+            if (unit->CanFly())
                 return;
 
             // fall back to gridHeight if any
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1422,40 +1422,29 @@ void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z) const
 
     if (Unit const* unit = ToUnit())
     {
-        if (!unit->CanFly())
-        {
-            bool canSwim = unit->CanSwim();
-            float ground_z = z;
-            float max_z;
-            if (canSwim)
-                max_z = GetMapWaterOrGroundLevel(x, y, z, &ground_z);
-            else
-                max_z = ground_z = GetMapHeight(x, y, z);
+        // Unit is flying, ignore.
+        if (unit->IsFlying())
+            return;
 
-            if (max_z > INVALID_HEIGHT)
-            {
-                // hovering units cannot go below their hover height
-                float hoverOffset = unit->GetHoverOffset();
-                max_z += hoverOffset;
-                ground_z += hoverOffset;
-
-                if (z > max_z)
-                    z = max_z;
-                else if (z < ground_z)
-                    z = ground_z;
-            }
-        }
+        bool canSwim = unit->CanSwim();
+        float ground_z = z;
+        float max_z;
+        if (canSwim)
+            max_z = GetMapWaterOrGroundLevel(x, y, z, &ground_z);
         else
+            max_z = ground_z = GetMapHeight(x, y, z);
+
+        if (max_z > INVALID_HEIGHT)
         {
-            float const gridHeight = GetMap()->GetGridHeight(x, y) + unit->GetHoverOffset();
-            float const vmapHeight = GetMap()->GetVmapDataFloorZ(x, y, z); // don't add hover offset here. if check might return unintended value
-            // Check if vmap under us
-            if (vmapHeight <= INVALID_HEIGHT)
-            {
-                // no vmap found, check if unit is under grid height
-                if (z < gridHeight)
-                    z = gridHeight;
-            }
+            // hovering units cannot go below their hover height
+            float hoverOffset = unit->GetHoverOffset();
+            max_z += hoverOffset;
+            ground_z += hoverOffset;
+
+            if (z > max_z && !unit->CanFly())
+                z = max_z;
+            else if (z < ground_z)
+                z = ground_z;
         }
     }
     else
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1418,7 +1418,12 @@ void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z, float* grou
 {
     // TODO: Allow transports to be part of dynamic vmap tree
     if (GetTransport())
+    {
+        if (groundZ)
+            *groundZ = z;
+
         return;
+    }
 
     if (Unit const* unit = ToUnit())
     {
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1414,7 +1414,7 @@ void WorldObject::UpdateGroundPositionZ(float x, float y, float &z) const
         z = new_z + (isType(TYPEMASK_UNIT) ? static_cast<Unit const*>(this)->GetHoverOffset() : 0.0f);
 }
 
-void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z) const
+void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z, float* groundZ) const
 {
     // TODO: Allow transports to be part of dynamic vmap tree
     if (GetTransport())
@@ -1422,29 +1422,40 @@ void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z) const
 
     if (Unit const* unit = ToUnit())
     {
-        // Unit is flying, ignore.
-        if (unit->IsFlying())
-            return;
-
-        bool canSwim = unit->CanSwim();
-        float ground_z = z;
-        float max_z;
-        if (canSwim)
-            max_z = GetMapWaterOrGroundLevel(x, y, z, &ground_z);
-        else
-            max_z = ground_z = GetMapHeight(x, y, z);
-
-        if (max_z > INVALID_HEIGHT)
+        if (!unit->CanFly())
         {
-            // hovering units cannot go below their hover height
-            float hoverOffset = unit->GetHoverOffset();
-            max_z += hoverOffset;
-            ground_z += hoverOffset;
+            bool canSwim = unit->CanSwim();
+            float ground_z = z;
+            float max_z;
+            if (canSwim)
+                max_z = GetMapWaterOrGroundLevel(x, y, z, &ground_z);
+            else
+                max_z = ground_z = GetMapHeight(x, y, z);
 
-            if (z > max_z && !unit->CanFly())
-                z = max_z;
-            else if (z < ground_z)
+            if (max_z > INVALID_HEIGHT)
+            {
+                // hovering units cannot go below their hover height
+                float hoverOffset = unit->GetHoverOffset();
+                max_z += hoverOffset;
+                ground_z += hoverOffset;
+
+                if (z > max_z)
+                    z = max_z;
+                else if (z < ground_z)
+                    z = ground_z;
+            }
+
+            if (groundZ)
+                *groundZ = ground_z;
+        }
+        else
+        {
+            float ground_z = GetMapHeight(x, y, z) + unit->GetHoverOffset();
+            if (z < ground_z)
                 z = ground_z;
+
+            if (groundZ)
+               *groundZ = ground_z;
         }
     }
     else
@@ -1452,6 +1463,9 @@ void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z) const
         float ground_z = GetMapHeight(x, y, z);
         if (ground_z > INVALID_HEIGHT)
             z = ground_z;
+
+        if (groundZ)
+            *groundZ = ground_z;
     }
 }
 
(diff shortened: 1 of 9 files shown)


yvomsx:
Vouch +1

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1428,9 +1428,15 @@ void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z) const
         }
         else
         {
-            float ground_z = GetMapHeight(x, y, z) + unit->GetHoverOffset();
-            if (z < ground_z)
-                z = ground_z;
+            float const gridHeight = GetMap()->GetGridHeight(x, y) + unit->GetHoverOffset();
+            float const vmapHeight = GetMap()->GetVmapDataFloorZ(x, y, z); // don't add hover offset here. if check might return unintended value
+            // Check if vmap under us
+            if (vmapHeight <= INVALID_HEIGHT)
```
Shauren: Why did you split these calls? GetMapHeight uses GetMap()->GetHeight which is supposed to return the correct height value, either from grid or vmaps

src/server/game/Maps/Map.cpp
```diff
@@ -2441,24 +2441,11 @@ float Map::GetWaterOrGroundLevel(uint32 phasemask, float x, float y, float z, fl
...
-
-    float vmapHeight = VMAP_INVALID_HEIGHT_VALUE;
-    if (checkVMap)
-    {
-        VMAP::IVMapManager* vmgr = VMAP::VMapFactory::createOrGetVMapManager();
-        if (vmgr->isHeightCalcEnabled())
-            vmapHeight = vmgr->getHeight(GetId(), x, y, z, maxSearchDist);
-    }
+    float const mapHeight = GetGridHeight(x, y);
+    float const vmapHeight = checkVMap ? GetVmapDataFloorZ(x, y, z) : VMAP_INVALID_HEIGHT_VALUE;
```
Shauren: Why replace vmgr->getHeight with vmgr->getAreaAndLiquidData?

src/server/game/Entities/Object/Object.cpp
Golrag: The changes made in Map::GetHeight will make it so that it will select the closest one, not caring if z is lower than grid. The drakes in the caverns of time will rise up again if I use that function.

I'm thinking about reverting that part so we can still use the same function but not sure yet if it could break things again

src/server/game/Maps/Map.cpp
Golrag: Was looking for a function that could return the floor without a max search distance, as the default one was not enough so it would return an invalid vmap height iirc

Need to check it again to be sure

Aokromes: @Golrag jump into irc :)

yvomsx:
@Golrag Managed to resolve it? i can also take a look for u if u wish.

src/server/game/Maps/Map.cpp
yvomsx: @Aokromes Any news on this? 👍

Aokromes: @yvomsx test it yourself, and give feedback.

yvomsx: @Aokromes Give me a little, will review and post results.

Killyana:
There's an issue about random movements https://github.com/TrinityCore/TrinityCore/issues/21872
You can notice it IG with this npcs: .go c id 23929

yvomsx:
Checked this pull, compiled and tested, appears to be working so far.

Persephone:
@Golrag Can you work your magic on shadowstep as well? If you try to shadowstep a target that is on a higher ground than yourself, you get placed underground. 

As an example i spawned a dummy (id 32666) in guru here:

http://prntscr.com/m837ek

and when i use shadowstep i end up here:

http://prntscr.com/m837rt

Lorenc95 (approved):
Tested, working been running 2 days, without a crash or whatsoever, issue is not completely gone i guess but its a major improvement of what it used to be.

IllInuz:
@Killyana Above comment is what i meant::

> I found an issue (with blink)

.go xyz -8173.074219 -4719.009766 27.010540 1 4.943924
end position by path finder: -8168.48438 -4738.47656 29.2423859
--> Grid z: invalid
--> vmap z: invalid (we are under vmap)

--> undermap

illfated:
>  [ ... ] issue is not completely gone [ ... ]

Yes, I pretty much took it to mean that there are still some bugged coordinates, but it is (of course) better to include actual references to which bugs are still present.

Illuminati2739:
Hello, is this going to be applied or do we have access to to this patch?

Golrag:
@Shauren I might have an approach that removes the call to getAreaAndLiquidData. Could just ignore UpdateAllowedPositionZ for flying units, not units who can fly but really are flying. For units that can fly we can keep the code that's currently present in !unit->CanFly() and modify the max_z check.

This would mean that flying units can fly through mountains. While previously they could go all the way to the top and then go back down to reach you.

```diff
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1422,40 +1422,29 @@ void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z) const
 
     if (Unit const* unit = ToUnit())
     {
-        if (!unit->CanFly())
-        {
-            bool canSwim = unit->CanSwim();
-            float ground_z = z;
-            float max_z;
-            if (canSwim)
-                max_z = GetMapWaterOrGroundLevel(x, y, z, &ground_z);
-            else
-                max_z = ground_z = GetMapHeight(x, y, z);
+        // Unit is flying, ignore.
+        if (unit->IsFlying())
+            return;
 
-            if (max_z > INVALID_HEIGHT)
-            {
-                // hovering units cannot go below their hover height
-                float hoverOffset = unit->GetHoverOffset();
-                max_z += hoverOffset;
-                ground_z += hoverOffset;
-
-                if (z > max_z)
-                    z = max_z;
-                else if (z < ground_z)
-                    z = ground_z;
-            }
-        }
+        bool canSwim = unit->CanSwim();
+        float ground_z = z;
+        float max_z;
+        if (canSwim)
+            max_z = GetMapWaterOrGroundLevel(x, y, z, &ground_z);
         else
+            max_z = ground_z = GetMapHeight(x, y, z);
+
+        if (max_z > INVALID_HEIGHT)
         {
-            float const gridHeight = GetMap()->GetGridHeight(x, y) + unit->GetHoverOffset();
-            float const vmapHeight = GetMap()->GetVmapDataFloorZ(x, y, z); // don't add hover offset here. if check might return unintended value
-            // Check if vmap under us
-            if (vmapHeight <= INVALID_HEIGHT)
-            {
-                // no vmap found, check if unit is under grid height
-                if (z < gridHeight)
-                    z = gridHeight;
-            }
+            // hovering units cannot go below their hover height
+            float hoverOffset = unit->GetHoverOffset();
+            max_z += hoverOffset;
+            ground_z += hoverOffset;
+
+            if (z > max_z && !unit->CanFly())
+                z = max_z;
+            else if (z < ground_z)
+                z = ground_z;
         }
     }
     else
```
note: changes in map::getheight for vmapheight are reverted. still keeping the gridheight change

jackpoz:
@Golrag please post the link to a branch, not diff files

Aokromes:
@Golrag if this pr is intended to fix also shadow step, still fails

Golrag:
I don't know yet what's happening with shadowstep, will try to take a look at it tomorrow. I noticed it goes through MovePositionToFirstCollision without checking if there is a path, but I'm not sure if it returns a good position or not

natnat201193:
if this is of any help, i think it started happening more often when https://github.com/TrinityCore/TrinityCore/pull/21322 was reverted. but i could be wrong.

natnat201193:
Reverted here: https://github.com/TrinityCore/TrinityCore/commit/2dadbda24ac1c2e6ff3dad650c7e25f4b3322342#diff-d64b193f1697462f5500332fb47092ee

Golrag:
Could I get some examples of where this issue occurs ? I tried getting it in gurubashi with the example above but could not reproduce it :/ a .gps of the npc would be great

Ovahlord:
Well, I've been testing the changes so far and normal terrain just seems to be working fine. wmo terrain however seems to be struggling a bit. The best place to see that is going into Naxxramas and doing a blink towards one of the stairways up. When doing so you should be partially standing inside the stairs. If there wouldn't be a 2nd collision model below the stairs you would still fall straight through them. So yeah. Blink is more or less halfway fixed but still struggles.

Golrag:
I'm sorry I just can't reproduce it. I might be missing something..
https://puu.sh/CQWse/07f7ee930a.mp4
https://puu.sh/CQWt6/e97cc7ed50.mp4

Can't know for sure what's going on, but I think that those issues are not in scope of this PR :/
will try again after I updated tools with latest changes

VengefulWoW:
The best place is Gurubashi Arena right outside. You can blink into abyss in many different spots. It is bugged as hell. People constantly fall through a map

Illuminati2739:
Is this going to be applied? This fixes blink and shadow step however i do not see this as an actual commit

illfated:
The whole point of a Pull Request is to deal with unanswered questions or unsolved issues in the content of the PR before it gets merged. I recommend showing some patience with regards to when this will be merged. It will be merged when reviews and requested changes have been sorted out and solved (and the owners of TrinityCore are happy with the results).

Illuminati2739:
Of course, no rush was intended or referenced in my post, It was a simple question, which was not posed by me or anyone else before hand. Reason for asking was because "All Checks" were completed. 
Thank you

illfated:
Very well, fair enough. "All checks have passed" is mostly an indicator that the code passes automatic review and that no obvious errors need to be fixed. TC members are still free to discuss the content and manual testing is needed to see if the in-game behaviour actually stays blizzlike. Real coders will also be able to spot other technical reasons to request changes.

jackpoz:
@Shauren do the new changes look better ? can it be merged now ?

src/server/game/Maps/Map.cpp
```diff
@@ -2444,14 +2444,7 @@ float Map::GetWaterOrGroundLevel(uint32 phasemask, float x, float y, float z, fl
...
     // find raw .map surface under Z coordinates
-    float mapHeight = VMAP_INVALID_HEIGHT_VALUE;
-    if (GridMap* gmap = const_cast<Map*>(this)->GetGrid(x, y))
-    {
-        float gridHeight = gmap->getHeight(x, y);
-        if (G3D::fuzzyGe(z, gridHeight - GROUND_HEIGHT_TOLERANCE))
-            mapHeight = gridHeight;
-    }
-
+    float const mapHeight = GetGridHeight(x, y);
```
Shauren: Whats the reasoning behind this change? The point of `z` argument is to never return a height that is higher than supplied `z` (and vmgr->getHeight follows this restriction)

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -536,14 +536,14 @@ void PathGenerator::BuildPointPath(const float *startPoint, const float *endPoin
         /// @todo check the exact cases
         TC_LOG_DEBUG("maps", "++ PathGenerator::BuildPointPath FAILED! path sized %d returned\n", pointCount);
         BuildShortcut();
-        _type = PATHFIND_NOPATH;
+        _type = PathType(_type | PATHFIND_NOPATH);
```
jackpoz: what does this change btw ?

src/server/game/Maps/Map.cpp
Golrag: I noticed some issues with blink without it... might have been because of https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/Spell.cpp#L1368 allowing shortcut paths, not sure if that's supposed to happen. Can try to remove it and change the check to only allow normal paths

src/server/game/Movement/PathGenerator.cpp
Golrag: Prevents removing PATHFIND_SHORTCUT from the path type. I agree it's a bit weird to combine it with NOPATH. As most of the checks are already done on NOPATH

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1422,33 +1422,28 @@ void WorldObject::UpdateAllowedPositionZ(float x, float y, float &z) const
...
-        {
-            bool canSwim = unit->CanSwim();
-            float ground_z = z;
-            float max_z;
-            if (canSwim)
-                max_z = GetMapWaterOrGroundLevel(x, y, z, &ground_z);
-            else
-                max_z = ground_z = GetMapHeight(x, y, z);
+        // Unit is flying, ignore.
+        if (unit->IsFlying())
```
Ovahlord: You need to revisit the logic in this part here. The changes break hover ground z coordinates for units that can fly and have ground hover movement (for example the valkyrs in ICC). So with the changes made here the valkyrs for example never get a proper ground Z returned because they can fly and they can hover above the ground. The IsFlying check is effectively disabling all hover checks in this case so the valkyr can fly into the ground ignoring the hover offset entirely.

robinsch:
WMOs are still bad. For example in Stormwind: https://gyazo.com/a7d9834cb689ffabc5a877c3b169d407

```
.go xyz -8843 634 94
```

Duratans (approved):

src/server/game/Entities/Object/Object.cpp
```diff
@@ -3283,10 +3297,27 @@ void WorldObject::MovePositionToFirstCollision(Position &pos, float dist, float
...
+    UpdateAllowedPositionZ(destx, desty, pos.m_positionZ, &groundZ);
     pos.SetOrientation(GetOrientation());
+
+    // position has no ground under it (or is too far away)
+    if (groundZ <= INVALID_HEIGHT)
+    {
+        if (Unit const* unit = ToUnit())
+        {
+            // flying, ignore.
+            if (unit->IsFlying())
```
Ovahlord: better go with  if (unit->CanFly()) because, once again, IsFlying also returns true for units with disabled gravity which are creatures which should still consider ground heights when getting too close to it. (Naming the icc Valykr scenario again)

Cobus420:
No issue longer with blink and Fear. It works nice here!

illfated:
@Cobus420 : Some more details would be helpful, like if you mean this PR content and which commit rev. hash from the TC source you based it on.

jackpoz:
is this PR ready to be merged ? could the ones who have been testing it test latest version too and provide some feedback so we know if we can merge it ? thanks 👍

Aokromes:
imho yes, it works far better than old code, i don't have got more reports of people falling under floor.

jackpoz:
Thanks for the PR :)

PR #22954, opened 2019-01-16
Core/Misc: Wisp Spirit minor edit

KobeSagami:
Made a change to have the Wisp Spirit racial to work on anyone who has the spell. A convenience for custom projects and makes no difference on a blizzlike project.

**Changes proposed:**

-  Make the Wisp Spirit racial passive work for anyone who has the racial, not just Night Elves.
-  This is a convenience for custom projects who may want to use the racial on other races.
-  Has no affect on blizzlike projects.

- [ ] 3.3.5
- [ ] master

- N/A

- Builds and works in-game.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -4235,7 +4235,8 @@ void Player::BuildPlayerRepop()
     packet.PlayerGUID = GetGUID();
     GetSession()->SendPacket(packet.Write());
 
-    if (getRace() == RACE_NIGHTELF)
+    // If the player has the Wisp racial then cast the Wisp aura on them
+    if (HasSpell(20585))
         CastSpell(this, 20584, true);
     CastSpell(this, 8326, true);
 


src/server/game/Entities/Player/Player.cpp
```diff
@@ -4164,8 +4164,11 @@ void Player::BuildPlayerRepop()
     packet.PlayerGUID = GetGUID();
     GetSession()->SendPacket(packet.Write());
 
-    if (getRace() == RACE_NIGHTELF)
+    // If the player has the Wisp racial then cast the Wisp aura on them
+    if (HasSpell(20585))
```
Kittnz: magic number

Shauren: so are both spell casts below...

Rynem: Is it really needed to have enums for every spell, considering they're not exactly "magic numbers" as it's evident that they're spells and there are various lookup tools for them?

mdX7: Yes, because you save that lookup time when using enums. And you have to remind which spellid is related to which spell when reading longer code. With enums you directly know whats happening.

PR #22964, opened 2019-01-18
Account\Pcopy Command

### LINKED ISSUE #22921: pcopy command
**Description:**

pcopy command will be the fast way for pdump + pload

**Current behaviour:**

When we have to test bugs we must levelup characters, equip them, learn talents and spells or pdump a character and load to one new account

**Expected behaviour:**

with pcopy it will allow use pcopy "source character" "target account" we will have a copy of one existing character.

1. go test some bug

**Branch(es):**

CHANGEME both

**TC rev. hash/commit:** 

N/R d12f0841f198

**Operating system:** N/R

Lorenc95:
**Changes proposed:**

-  Adds a command to copy a character
-  
-  

- [√] 3.3.5
- [ ] master

**Issues addressed:** Closes #22921 

**Tests performed:** None

- [ ] add the actual copying of the character to the new account
- [ ] Leave the actual character and account owner unharmed/untouched/unaltered.
- [ ] Use the same GUID? if not, Copy all character info. 
- [ ] Force the rename on the character using a DB query and update the value of forcerename

This is just a quick edit, will update, was to get the general idea of what i had in mind going!!

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9563,7 +9563,7 @@ void Unit::UpdateCharmAI()
         UnitAI* newAI = nullptr;
         if (GetTypeId() == TYPEID_PLAYER)
         {
-            if (Unit* charmer = GetCharmer())
+            if (Unit* charmer = GetCharmer() || GetCharmer() == TYPEID_PLAYER)
             {
                 // first, we check if the creature's own AI specifies an override playerai for its owned players
                 if (Creature* creatureCharmer = charmer->ToCreature())
--- a/src/server/scripts/Commands/cs_misc.cpp
+++ b/src/server/scripts/Commands/cs_misc.cpp
@@ -116,6 +116,7 @@ public:
             { "unstuck",          rbac::RBAC_PERM_COMMAND_UNSTUCK,           true, &HandleUnstuckCommand,          "" },
             { "wchange",          rbac::RBAC_PERM_COMMAND_WCHANGE,          false, &HandleChangeWeather,           "" },
             { "mailbox",          rbac::RBAC_PERM_COMMAND_MAILBOX,          false, &HandleMailBoxCommand,          "" },
+            { "pcopy",            rbac::RBAC_PERM_COMMAND_PCOPY,            false, &HandlPCopyCommand,          "" },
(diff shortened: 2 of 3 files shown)


mdX7:
`Use the same GUID? if not, Copy all character info.`
Using the same GUID of the char being copied shouldn't be possible since it already exists. Or I did not get what you mean :D

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9563,7 +9563,7 @@ void Unit::UpdateCharmAI()
         UnitAI* newAI = nullptr;
         if (GetTypeId() == TYPEID_PLAYER)
         {
-            if (Unit* charmer = GetCharmer())
+            if (Unit* charmer = GetCharmer() || GetCharmer() == TYPEID_PLAYER)
```
Aokromes: i suppose not intended change

illfated: Anyway, that change is the reason why the build fails.

IllInuz: Non intented change, will revert.

src/server/scripts/Commands/cs_misc.cpp
```diff
@@ -116,6 +116,7 @@ class misc_commandscript : public CommandScript
             { "unstuck",          rbac::RBAC_PERM_COMMAND_UNSTUCK,           true, &HandleUnstuckCommand,          "" },
             { "wchange",          rbac::RBAC_PERM_COMMAND_WCHANGE,          false, &HandleChangeWeather,           "" },
             { "mailbox",          rbac::RBAC_PERM_COMMAND_MAILBOX,          false, &HandleMailBoxCommand,          "" },
+            { "pcopy",            rbac::RBAC_PERM_COMMAND_PCOPY,            false, &HandlPCopyCommand,          "" },
```
kuskmen: Typo here should be &HandlePCopyCommand?

kuskmen:
What happened with this PR why it is closed?

illfated:
Only @IllInuz knows what actually happened. Maybe something went wrong during his attempt to fix the mistake above? Maybe he was unable to open a new PR or did not have enough time? I don't know.

Issue #21731, opened 2018-03-29
Core/Creature: Creature keep following the leader once charmed


Killyana:
**Description:** If an npc is following a leader (formation) even once charmed by a player it will keep the follow movement instead being controlled by the player.

**Steps to reproduce the problem:**

1. .go c id 28605
2. ride on of the horses that follows spawnID= 129212
3. notice it will keep following the leader

TrinityCore rev. b274c4b5c346 2018-03-28 04:15:09 +0200 (3.3.5 branch) (Win32, Release, Static)
Using SSL version: OpenSSL 1.0.1e 11 Feb 2013 (library: OpenSSL 1.0.1e 11 Feb 2013)
Using Boost version: 1.59.0
Using MySQL version: 5.5.39
Using CMake version: 3.10.1
Compiled on: Windows 
Automatic database updates are disabled for all databases!
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in ./maps. Total size: 252191207 bytes
vmaps directory located in ./vmaps. Total size: 588247501 bytes
mmaps directory located in ./mmaps. Total size: 2229000480 bytes
Using frFR DBC Locale as default. All available DBC locales: frFR 
Using World DB: TDB 335.64

Langerz82:
If the leader is stolen the formation waypoints break, but thats a seperate known issue.

```diff
diff --git a/src/server/game/Entities/Unit/Unit.cpp b/src/server/game/Entities/Unit/Unit.cpp
index 6e8a96d72f..60a3476809 100644
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12482,6 +12482,10 @@ void Unit::_EnterVehicle(Vehicle* vehicle, int8 seatId, AuraApplication const* a
     if (!IsAlive() || GetVehicleKit() == vehicle || vehicle->GetBase()->IsOnVehicle(this))
         return;
 
+    Unit* unitVehicle = vehicle->GetBase();
+    if (unitVehicle && unitVehicle->GetTypeId() == TYPEID_UNIT && unitVehicle->ToCreature()->GetFormation())
+        unitVehicle->ToCreature()->GetFormation()->RemoveMember(unitVehicle->ToCreature());
+
     if (m_vehicle)
[...]

PR #22970, opened 2019-01-20
Core/Creature: Handle Creature Vehicle from Formation

Sorikoff:
**Changes proposed:**
Remove creature as vehicle from formation on entering vehicle, and restore formation on exiting vehicle.

**Target branch(es):** 3.3.5

**Issues addressed:** Ref #21731 (applies to vehicles, not sure about actual charming)


Sorikoff:
Nope. This is wrong.

Issue #19909, opened 2017-06-18
[DEV] Core/Movements: Add the possiblity to choose run/walk random move


Killyana:
**Description:** Some NPCs must have Run speed on with random movements, This cannot be done with the actual DB scripting.

Proposed solution: Add random_movement_type on creature table: 0= walk, 1=run

**Branch(es):** both

**TC rev. hash/commit:**  rev. 9fc35785307e

Faq:
Any example npc with such behavior?

Killyana:
This https://www.youtube.com/watch?v=Tm_SxTkr9eU

Killyana:
Work done by Rushor: https://gist.github.com/Killyana/acdfedd5a8c7f87fa310397dd23d1493

jackpoz:
strange that it's a float but it's handled as a boolean

Killyana:
Imagine 20 spawns, 10 use walk speed and 10 run speed, in this case we must add 10 extra SAI per guid instead of just set a Boolean into a field.

Killyana:
@Sorikoff Can you open PR similar to the one that implemented chasing speed, using this https://gist.github.com/Killyana/acdfedd5a8c7f87fa310397dd23d1493 as start ?

Sorikoff:
@Killyana, sure I could try. But as far as I understand this is for master branch nor 3.3.5?

ccrs:
both

Sorikoff:
You would 3 options, correct? Walk, CanRun, AlwaysRun?

Killyana:
Yes

Killyana:
Implemented on https://github.com/TrinityCore/TrinityCore/pull/23081

PR #22973, opened 2019-01-21
Core/Movement: Respect walk state when chasing

### LINKED ISSUE #20938: [3.3.5] Creatures always run when chasing
**Description:**

Creatures always use move type MOVE_RUN when MoveChase() is issued. 

**Current behaviour:**

Creatures always use MOVE_RUN when MoveChase() is issued. regardless of whether they are set to walk or not, which is the default for any attacking creature. 

**Expected behaviour:**

Not all creatures should run when chasing their targets. Thaladred the Darkener (http://www.wowhead.com/npc=20064/thaladred-the-darkener) for example, should walk while chasing players, but currently there is no way to achieve that, because https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L220 always disables walk mode.

1. Make Thaladred the Darkener walk instead running - SetWalk(true)
2. Start Kael'thas encounter
3. Watch as Thaladred still runs after his targets at full speed

**Branch(es):**

3.3.5/likely both

https://github.com/TrinityCore/TrinityCore/commit/a422dad8702c779b5665891d9e651aef575f33f7

Sorikoff:
**Changes proposed:**
* Chase Movement Generator is modified to not to set walk mode to false on initialization.
* Chase Movement Generator is also modified to take into account current walk mode during each update tick, therefore allowing to chase in walking mode only when needed and set to do so.

**Target branch(es):** 3.3.5

**Issues addressed:** Closes #20938

### DIFF
--- a/src/server/scripts/Northrend/FrozenHalls/HallsOfReflection/halls_of_reflection.cpp
+++ b/src/server/scripts/Northrend/FrozenHalls/HallsOfReflection/halls_of_reflection.cpp
@@ -707,7 +707,7 @@ class npc_jaina_or_sylvanas_intro_hor : public CreatureScript
                             falric->AI()->Talk(SAY_FALRIC_INTRO_1);
                             falric->SetWalk(true);
                             falric->GetMotionMaster()->MovePoint(0, FalricPosition[1]);
-                            marwyn->SetWalk(false);
+                            falric->SetWalk(false);
                         }
                         _events.ScheduleEvent(EVENT_INTRO_LK_9, 5000);
                         break;
--- a/src/server/scripts/Northrend/FrozenHalls/HallsOfReflection/halls_of_reflection.cpp
+++ b/src/server/scripts/Northrend/FrozenHalls/HallsOfReflection/halls_of_reflection.cpp
@@ -697,6 +697,7 @@ class npc_jaina_or_sylvanas_intro_hor : public CreatureScript
                             marwyn->AI()->Talk(SAY_MARWYN_INTRO_1);
                             marwyn->SetWalk(true);
                             marwyn->GetMotionMaster()->MovePoint(0, MarwynPosition[1]);
+                            marwyn->SetWalk(false);
                         }
                         _events.ScheduleEvent(EVENT_INTRO_LK_8, 1000);
                         break;
@@ -706,6 +707,7 @@ class npc_jaina_or_sylvanas_intro_hor : public CreatureScript
                             falric->AI()->Talk(SAY_FALRIC_INTRO_1);
                             falric->SetWalk(true);
                             falric->GetMotionMaster()->MovePoint(0, FalricPosition[1]);
+                            marwyn->SetWalk(false);
                         }
                         _events.ScheduleEvent(EVENT_INTRO_LK_9, 5000);
                         break;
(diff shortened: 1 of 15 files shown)


Shauren:
What about all existing scripts that modify walk/run mode? I would like you be sure you thought about them (and adjusted if needed, for example by removing walk flag if it was set for RP intro or something similar) when making a change to generic system like this and not pretended they don't exist.

Sorikoff:
Fair enough. Adjusted walk mode for creatures using it. Needs some additional testing.

src/server/scripts/Northrend/FrozenHalls/HallsOfReflection/halls_of_reflection.cpp
```diff
@@ -706,6 +707,7 @@ class npc_jaina_or_sylvanas_intro_hor : public CreatureScript
                             falric->AI()->Talk(SAY_FALRIC_INTRO_1);
                             falric->SetWalk(true);
                             falric->GetMotionMaster()->MovePoint(0, FalricPosition[1]);
+                            marwyn->SetWalk(false);
```
illfated: Copy/paste error, replace `marwyn` with `falric` :-P

Sorikoff:
Tested. Everything works now.

src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
```diff
@@ -1235,6 +1235,7 @@ class npc_tirion_fordring_tft : public CreatureScript
                     me->RemoveFlag(UNIT_NPC_FLAGS, UNIT_NPC_FLAG_GOSSIP);
                     me->SetWalk(true);
                     me->GetMotionMaster()->MovePoint(POINT_TIRION_INTRO, TirionIntro);
+                    me->SetWalk(false);
```
ccrs: uh? this should be called when chase movement starts, right?

ccrs (changes requested):
This actually needs to check all `SetWalk` uses in scripts and even core.

`SetWalk(false)` should be used just before `MoveChase` call.
And if there isnt, analyze where/when the unit should actually start running.

src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
Shauren: tirion never chases anything

ccrs: edited the comment

jackpoz:
I wonder if "Running when chasing" should be the default behavior and then some sort of flag changes the behavior to "Respect setWalk() when chasing too"

robinsch:
It's distance based. Same as the "more speed the more distance you have" when following. Most video sources I found are about < 3 yards the pet walks or chases in walking mode.

Sorikoff:
@jackpoz, I like your approach. However, I am unsure whether it would be better to keep flag in DB or it would be just fine to make additional method for C++ scripting? It looks to me that first option could be better since it would make adjusting of this a way more easier.

jackpoz:
db flag usually sounds better for these kind of things. 
do you have a list of units where it would be already needed ? just to get an idea if the db solution will fit them all.
anyway it should be possible to add support to C++ too later on if needed (or maybe it's already supported to change db flags at runtime from C++ scripts)

Sorikoff:
I know a couple of them:
1. [Dark Rider of Acherus](https://www.wowhead.com/npc=28768/dark-rider-of-acherus) (walks in scripted event)
3. [Thaladred the Darkener](https://www.wowhead.com/npc=20064/thaladred-the-darkener) (walks always)
2. [Lava Scion](https://www.wowhead.com/npc=53231/lava-scion) (does not belong to 3.3.5 branch)

Killyana:
I asked for a similar future to be able to choose between walk/run for random_movements https://github.com/TrinityCore/TrinityCore/issues/19909 maybe the gist could help.

PR #22994, opened 2019-01-29
Core/Movement: Implement move time skipped handler

chaodhib:
Implement CMSG_MOVE_TIME_SKIPPED handler and move it to MovementHandler.cpp

**Changes proposed:**

The way the MOVE_TIME_SKIPPED packet works is as follow:
- The client sends it when the player [1] is moving but the client froze for a certain period of time.
- The packet only contains two information: the GUID of the moving unit. And the time elasped since the last movement packet has been sent.
- The meaning of the packet from the point of the view of the client that sends it is basically: I didn't run the physical simulation of the unit that I control for a certain period of time. Please, just play at a later time the last movement that was sent (new timestamp = previous timestamp + "time skipped").
- the server upon receiving that packet should 
	* retransmit the packet to the other clients that have a unit nearby.
	* update the movement timestamp of that unit (new timestamp = previous timestamp + "time skipped").
- the clients upon receiving also update the movement timestamp of that unit (new timestamp = previous timestamp + "time skipped") so that the previous move sent is "played" at a later time.

[1] or more generaly the unit currently moved by that client: it can be another player (in case of mind control for example), a vehicle, ...

More information on the packet exchange can be found in [this sniff extract](https://gist.github.com/chaodhib/e66010a8ecb77b406ff86f6bbe93a5e0#file-move-time-skipped-sniff). In it, we can see 3 packets sent by the client. The unit is moving between the first and third packet (and also before the first and after the third but that's not important here). The distance traveled between the first and third packet is only 500ms * run speed. Yet, the time difference between the 1st & 3rd packet is 500ms + 86ms. 86ms is the "time skipped". This confirms that the client didn't run the physical simulation during that duration.

- [X] 3.3.5
- [ ] master

**Issues addressed:** This is retail like. The implementation of that packet handler was missing. The movement should feel more smooth. Particularly when the [anti jitter buffer](https://trinitycore.atlassian.net/wiki/spaces/tc/pages/721256449/Movement) in the client will be utilized by TC, which is not the case at the moment.

Tested in game.

The line `ASSERT(mover != nullptr);` could cause problems if mover control change (client A looses control of player B. Or client A gains control of Vehicle C, ...) is badly implemented. I didn't test these edge cases enough.

### DIFF
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -619,11 +619,19 @@ void WorldSession::HandleMoveTimeSkippedOpcode(WorldPacket& recvData)
     recvData >> timeSkipped;
 
     Unit* mover = GetPlayer()->m_unitMovedByMe;
-    ASSERT(mover != nullptr);                      // there must always be a mover
+
+    if (mover == nullptr)
+    {
+        TC_LOG_WARN("entities.player", "WorldSession::HandleMoveTimeSkippedOpcode wrong mover state from the unit moved by the player %s", GetPlayer()->GetGUID().ToString().c_str());
+        return;
+    }
 
     // prevent tampered movement data
     if (guid != mover->GetGUID())
+    {
+        TC_LOG_WARN("entities.player", "WorldSession::HandleMoveTimeSkippedOpcode wrong guid from the unit moved by the player %s", GetPlayer()->GetGUID().ToString().c_str());
         return;
+    }
 
     mover->m_movementInfo.time += timeSkipped;
 
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -608,3 +608,27 @@ void WorldSession::HandleSummonResponseOpcode(WorldPacket& recvData)
 
     _player->SummonIfPossible(agree);
 }
+
+void WorldSession::HandleMoveTimeSkippedOpcode(WorldPacket& recvData)
+{
+    TC_LOG_DEBUG("network", "WORLD: Received CMSG_MOVE_TIME_SKIPPED");
+
+    ObjectGuid guid;
+    uint32 timeSkipped;
+    recvData >> guid.ReadAsPacked();
+    recvData >> timeSkipped;
+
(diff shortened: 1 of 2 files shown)


src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -608,3 +608,27 @@ void WorldSession::HandleSummonResponseOpcode(WorldPacket& recvData)
...
+{
+    TC_LOG_DEBUG("network", "WORLD: Received CMSG_MOVE_TIME_SKIPPED");
+
+    ObjectGuid guid;
+    uint32 timeSkipped;
+    recvData >> guid.ReadAsPacked();
+    recvData >> timeSkipped;
+
+    Unit* mover = GetPlayer()->m_unitMovedByMe;
+    ASSERT(mover != nullptr);                      // there must always be a mover
```
jackpoz: can't this just be a null check ? why does it have to be an assert ?

Shauren: You must NEVER assert in client packet handlers

chaodhib: Indeed. Good point. In that case though, existing code will need to be modified as well since there already are a few asserts in MovementHandler.

jackpoz: half of "crashes" we get seems to be triggered asserts :P

kelno: Well that's a good thing, they do outline abnormal behavior instead of failing silently

chaodhib: @kelno well yes and no. the issue is that an attacker could, by sending custom packets, intentionally crash a server running on TC. Example here: If a client sends a CMSG_MOVE_TIME_SKIPPED packet while his character is feared, the condition `mover != nullptr` won't be met and the server will crash.

kelno: Yes in this case it's totally bad I agree, I was making a more general comment in response for the last comment!

chaodhib:
Done. Anything else before merging?

jackpoz:
Looks good to me

jackpoz:
Thanks for the PR :)

PR #23029, opened 2019-02-07
Refactor SendSysMessage, SendGlobalSysMessage and SendGlobalGMSysMessage

Heaven31415:
**Changes proposed:** Refactor SendSysMessage, SendGlobalSysMessage and SendGlobalGMSysMessage using Tokenizer and cpp instead of c functions like strtok and strchr.

**Target branch:** 3.3.5

**Tests performed:** It builds correctly. I tested those changes during game runtime.

**TODO list:** There are many more places that are using c functions like strtok and similar instead of cpp (with no good reason in my opinion, cpp is much easier to understand), so if those changes are okay for you, I'm prepared to refactor even more code.

### DIFF
--- a/src/server/game/Chat/Chat.cpp
+++ b/src/server/game/Chat/Chat.cpp
@@ -168,75 +168,46 @@ bool ChatHandler::hasStringAbbr(char const* name, char const* part)
 
 void ChatHandler::SendSysMessage(const char *str, bool escapeCharacters)
 {
+    std::string msg{ str };
+
+    // Replace every "|" with "||" in msg
+    if (escapeCharacters && msg.find('|') != std::string::npos)
+    {
+        Tokenizer tokens{msg, '|'};
+        std::ostringstream stream;
+        for (size_t i = 0; i < tokens.size() - 1; ++i)
+            stream << tokens[i] << "||";
+        stream << tokens[tokens.size() - 1];
+
+        msg = stream.str();
+    }
+
     WorldPacket data;
-
-    // need copy to prevent corruption by strtok call in LineFromMessage original string
-    char* buf;
-    char* pos;
-
-    if (escapeCharacters && strchr(str, '|'))
-    {
-        size_t startPos = 0;
-        std::ostringstream o;
-        while (char const* charPos = strchr(str + startPos, '|'))
-        {
-            o.write(str + startPos, charPos - str - startPos);
-            o << "||";
-            startPos = charPos - str + 1;
-        }
-        o.write(str + startPos, strlen(str) - startPos);
-        buf = strdup(o.str().c_str());
-    }
-    else
-    {
-        buf = strdup(str);
-    }
-
-    pos = buf;
-
-    while (char* line = LineFromMessage(pos))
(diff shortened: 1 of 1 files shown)


Shauren:
Don't implement a split, we already have it... its called `Tokenizer` class, also found in Util.h

Shauren:
Don't implement a split, we already have it... its called `Tokenizer` class, also found in Util.h

Heaven31415:
Okay, closing this PR, gonna refactor this using `Tokenizer`.

Heaven31415:
Reopening because I have a question about `Tokenizer`.

`Tokenizer` is not the only way to split string into tokens, there is also `boost::tokenizer` with almost the same functionality. Which should I use?

Shauren:
`Tokenizer`

Heaven31415:
Ok, I changed it to use Tokenizer instead of my own function, any additional tips are welcome (connected with style for example).

src/server/game/Chat/Chat.cpp
```diff
@@ -168,75 +168,46 @@ bool ChatHandler::hasStringAbbr(char const* name, char const* part)
...
-    }
-    else
-    {
-        buf = strdup(str);
-    }
+        Tokenizer tokens{msg, '|'};
+        std::ostringstream stream;
+        for (size_t i = 0; i < tokens.size() - 1; ++i)
+            stream << tokens[i] << "||";
+        stream << tokens[tokens.size() - 1];
```
illfated: Looks to me as if there should be braces + equal indent for the 3 `stream` lines in this `for` loop.

Heaven31415: I'm not sure what do you mean exactly, but do you want me to write this instead of for loop from 178 line?
```cpp
for(size_t i = 0; i < tokens.size() - 1; ++i) 
{
  stream << tokens[i];
  stream << "||";
}
```

Warpten: Jesus Christ no

Heaven31415: Then what?

Warpten: nothing, its fine

illfated: Sorry for being less specific than usual. I thought it would look like this:
```cpp
        for (size_t i = 0; i < tokens.size() - 1; ++i)
        {
            stream << tokens[i] << "||";
            stream << tokens[tokens.size() - 1];
            msg = stream.str();
        }
```

Heaven31415: It would completely change semantics of those lines and those new semantics wouldn't make any sense.

illfated: OK, thanks for verifying my suspicion. I am not a real coder, so I do make mistakes.

jackpoz:
is this ready to be merged ? were last changes tested too ?

Heaven31415:
Yes, they were tested and it's ready to be merged.

jackpoz:
Thanks for the PR :)

PR #23039, opened 2019-02-10
[3.3.5] Commands: lookup player shows which character is online

TheWinchesters:
**Changes proposed:**
-  Show which character is online when using lookup player commands

- [x] 3.3.5
- [ ] master

Note: The way the code checks if player is online is the best one I found, but not sure if it is the best way.


JoNMii:
Don't use comments in Spanish.

illfated:
Cosmetic detail in your PR comment (top):
To make the MarkDown tick box work as intended, remove the blank space after the X,
or save the comment with the box unchecked and put the tick mark in manually.

> **Target branch(es):** 3.3.5/master
> 
> * [X] 3.3.5
> * [ ]  master
>

Shauren (dismissed):
looks good to me

sql/updates/world/3.3.5/9999_99_99_world.sql
```diff
@@ -0,0 +1,2 @@
+-- Show if player is online in lookup player commands
+UPDATE `trinity_string` SET `content_default` = '  %s (GUID %u) |cff00ff00%s|r' WHERE (`entry` = '329');
```
ccrs: I'm curious, whats the output if the player is offline?

TheWinchesters: I have used the same behavior that the "(Offline)" in the .pinfo command but in this case, it shows the opposite (If player is online).
In the case of .pinf command
```cpp
handler->PSendSysMessage(LANG_PINFO_PLAYER, target ? "" : handler->GetTrinityString(LANG_OFFLINE).....
```
In this case:
```cpp
`handler->PSendSysMessage(LANG_LOOKUP_PLAYER_CHARACTER, name.c_str(), guid, online ? "Online" : "");
```
So, if player is offline, output is "".
Once this shows which player is online, I do not see any reason to show something if player is offline. It would look redundant.

Shauren: Hmm, I take back my approval, make that "Online" string possible to localize (use trinity_string)

TheWinchesters:
Shauren, I did not like that "online" in that way either, but I am not sure about which `entry` value I should use to move it to trinity_string.

sql/updates/world/3.3.5/9999_99_99_world.sql
ccrs: My questions was actually oriented into what happens to the trinity_string if the given parameter is empty, as in the charater is offline. Server sends the color "pattern" with empty content?

ccrs:
either "recycle" one existing or just create a new one

sql/updates/world/3.3.5/9999_99_99_world.sql
TheWinchesters: I will remove the color when I move it to trinity_string
Is it better?

TheWinchesters:
Updated:
- Moved to `trinity_string`
- color was removed
- "Online" now is (online) to match the existing (offline)
- I used the unused entry 48 because it is the most close to (offline) and it looks 'ordered'

ccrs (approved):

jackpoz:
Thanks for the PR :)

TheWinchesters:
:)
Thanks to all of you for this project.

Issue #23047, opened 2019-02-14
Weird script base implementation


Romain-P:
**Description:**

We can see `CreatureScript` and `PlayerScript` classes are both derived of `UnitScript`.
 * https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Scripting/ScriptMgr.cpp#L2187
 * https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Scripting/ScriptMgr.cpp#L2280

When the parent class is called in the constructor, the second parameter is set to `false`. That means the UnitScript won't be added.
Why ?

**Current behaviour:**

I guess it was to avoid doubloons of the same instance in 2 differents registries (`UnitScripts` and `PlayerScripts`).  
But now assuming I'm creating a custom `CreatureScript` (which also is a derived class of `UnitScript`), the `UnitScript` hooks can be overrided but will never be triggered. This is true for every derived class of `UnitScript` calling the constructor with `false` that actually means `Do Not Add To UnitScripts Registry`.  

`PlayerScript` is indeed an `UnitScript` also called with `false` in the constructor, but a hack patch has been released long time ago, I guess, see:
 * https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Scripting/ScriptMgr.cpp#L2084

We can notice that for each `UnitScript` hook, a new line that shouldn't be here is added to fix PlayerScripts overriding UnitScript hooks.

```cpp
void ScriptMgr::OnHeal(Unit* healer, Unit* reciever, uint32& gain)
{
    FOREACH_SCRIPT(UnitScript)->OnHeal(healer, reciever, gain);
    FOREACH_SCRIPT(PlayerScript)->OnHeal(healer, reciever, gain);
}
```

This is actually a hack.

**Expected behaviour:**

2 possibilities:
 * Add new hacks for all `UnitScript` derived class
 * Put `true` to the constructor so the script instance will be in both registries

Or maybe a third I don't see right now.

**Steps to reproduce the problem:**

N/A

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 58c5dfac1d8d314b9243fde52466b3c933fb7d7f

LATEST

**Operating system:** ALL

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you and add the missing information to **this issue**. Thanks.

[...]

PR #23048, opened 2019-02-14
Core/Scripts: Change PlayerScript and CreatureScript inheritance

jackpoz:
**Changes proposed:**

Remove UnitScript from PlayerScript and CreatureScript classes as that's how the original system that was PR'd worked, see https://github.com/TrinityCore/TrinityCore/pull/7867 for reference.
Please note these are used as global hooks and should really just invoke stateless functions.
If you need to hook methods from PlayerScript/CreatureScript and UnitScript, just define 2 different scripts (this is how the original system worked and how it was designed to work)

See https://github.com/TrinityCore/TrinityCore/pull/7867#issuecomment-8854396 for more info as why this change was applied in the first place

- [x] 3.3.5
- [ ] master

- https://github.com/TrinityCore/TrinityCore/issues/23047
- It also fixes PlayerScript overridden methods defined in UnitScript not being called (as now those don't exist anymore and will give a build error to existing scripts IF the use properly used the override keyword)

```cpp
#include "ScriptMgr.h"
#include <cstdio>

class TestPlayerScript : PlayerScript
{
public:
    TestPlayerScript() : PlayerScript("TestPlayerScript")
    {
        printf("TestPlayerScript ctor\n");
    }

    void OnLogin(Player* /*player*/, bool /*firstLogin*/) override
    {
        printf("Logged in\n");
    }

    virtual ~TestPlayerScript()
    {
        printf("TestPlayerScript dtor\n");
    }
};

class TestCreatureScript : CreatureScript
{
public:
    TestCreatureScript() : CreatureScript("TestCreatureScript")
    {
        printf("TestCreatureScript ctor\n");
    }

    CreatureAI* GetAI(Creature* /*creature*/) const override
    {
        printf("GetAI\n");
        return nullptr;
    }

    virtual ~TestCreatureScript()
    {
        printf("TestCreatureScript dtor\n");
    }
};

class TestUnitScript : UnitScript
{
public:
    TestUnitScript() : UnitScript("TestUnitScript")
    {
        printf("TestUnitScript ctor\n");
    }

    void OnHeal(Unit* /*healer*/, Unit* /*reciever*/, uint32& /*gain*/) override
    {
        printf("Heal\n");
    }

    virtual ~TestUnitScript()
    {
        printf("TestUnitScript dtor\n");
    }
};

void AddCustomScripts()
{
    new TestPlayerScript();
    new TestCreatureScript();
    new TestUnitScript();
}
```

- It would be nice to just call functions instead of create class instances that are not thread-safe and that get half the methods called with nothing overriding the default base empty method.

### DIFF
--- a/src/server/game/Scripting/ScriptMgr.cpp
+++ b/src/server/game/Scripting/ScriptMgr.cpp
@@ -2084,31 +2084,26 @@ void ScriptMgr::OnGroupDisband(Group* group)
 void ScriptMgr::OnHeal(Unit* healer, Unit* reciever, uint32& gain)
 {
     FOREACH_SCRIPT(UnitScript)->OnHeal(healer, reciever, gain);
-    FOREACH_SCRIPT(PlayerScript)->OnHeal(healer, reciever, gain);
 }
 
 void ScriptMgr::OnDamage(Unit* attacker, Unit* victim, uint32& damage)
 {
     FOREACH_SCRIPT(UnitScript)->OnDamage(attacker, victim, damage);
-    FOREACH_SCRIPT(PlayerScript)->OnDamage(attacker, victim, damage);
 }
 
 void ScriptMgr::ModifyPeriodicDamageAurasTick(Unit* target, Unit* attacker, uint32& damage)
 {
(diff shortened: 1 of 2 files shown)


jackpoz:
Add a test script, worked like a charm and destructor was called once (not 0 times, not 2 times)

PR #23057, opened 2019-02-19
Fixed a logic mistake for ground detection in MovePositionToFirstCollision

Ovahlord:
**Changes proposed:**
fixed an issue that was causing some units getting summoned in the air instead of the ground. We use the collision to find anything between z + halfheight and z. destZ has been normalized already so we don't have to add another halfheight to them as it already has the correct ground height.

- [x] 3.3.5
- [ ] master

Tested ingame.

Summon behaivior prior to the change:
wowscrnshot_021919_174348

Summon behaivior after the change:
wowscrnshot_021919_174550

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -3234,7 +3234,7 @@ void WorldObject::MovePositionToFirstCollision(Position &pos, float dist, float 
     UpdateAllowedPositionZ(destx, desty, destz);
     bool col = VMAP::VMapFactory::createOrGetVMapManager()->getObjectHitPos(GetMapId(),
         pos.m_positionX, pos.m_positionY, pos.m_positionZ + halfHeight,
-        destx, desty, destz + halfHeight,
+        destx, desty, destz,
         destx, desty, destz, -0.5f);
 
     // collision occured
@@ -3249,7 +3249,7 @@ void WorldObject::MovePositionToFirstCollision(Position &pos, float dist, float 
     // check dynamic collision
     col = GetMap()->getObjectHitPos(GetPhaseMask(),
         pos.m_positionX, pos.m_positionY, pos.m_positionZ + halfHeight,
-        destx, desty, destz + halfHeight,
+        destx, desty, destz,
         destx, desty, destz, -0.5f);
 
     // Collided with a gameobject


illfated:
> **Changes proposed:**
> fixed an issue that was causing some units getting summoned in the air instead of the ground. We use the collision to find anything between z + halfheight and z. destZ has been normalized already so we don't have to add another halfheight to them as it already has the correct ground height.

Maybe I have misread something here, but judging by your before & after images, it looks like your fix makes the NPC hover properly above ground, not sink into the ground. Perhaps I have completely misunderstood the "causing some units getting summoned in the air instead of the ground" description, but if you could enlighten me, I would appreciate it.

Judging by your images alone, I started to think that this could perhaps fix other flying/hovering spawns starting off spawned halfway into the ground instead of hovering or flying properly, but maybe your fix only targets NPCs summoned by players?

Ovahlord:
What you see there is an earthquake. I think that explains your first part.

2nd this fix affects everything that has a caster destination target. The collision logic is working by 'looking from top down to find something between the two points. In this case the half height is being used as reference to see if there is uneven terrain for example. The old logic was basically looking at itself so of course the returned values would make no sense.

illfated:
Thanks. I see now that I misjudged the images and that there is clear difference between the images regarding the height of the spell visual as well as the summoned NPC. My bad.

Killyana:
I tested the 3 issues closed by https://github.com/TrinityCore/TrinityCore/commit/b9c6bbb51df0ed10f74617482740fedb91ff5366#diff-29afb15ead2a8204cac07d94f074f260 an every thing works fine

Shauren:
> 
> 
> What you see there is an earthquake. I think that explains your first part.
> 
> 2nd this fix affects everything that has a caster destination target. The collision logic is working by '**looking from top down**' to find something between the two points. In this case the half height is being used as reference to see if there is uneven terrain for example. The old logic was basically looking at itself so of course the returned values would make no sense.

Except this code is not meant to be used for looking top down. Its purpose is to shorten the distance (for a example of blink) on X/Y coords only. In your case I would investigate why `UpdateAllowedPositionZ` at the very end of `MovePositionToFirstCollision` is not doing its job grabbing the ground.

Ovahlord:
I think that can be explained easily. if (std::fabs(pos.m_positionZ - destz) > 6.0f)
This means literally anything below 6.0f is getting skipped. And oh suprirse halfheight is almost always  < 6.0f in every case.

jackpoz:
What repro steps did you use ingame btw ?

Shauren:
So, how to reproduce this on 3.3.5? Need spellid + caster coords + target coords

Killyana:
Take this quest https://www.wowhead.com/quest=12053/the-might-of-the-horde and go to Icemist Village and use the item, the summoned Warsong Battle Standard will appears above the ground

Shauren:
Ok I figured it out, its only bugged for casters that can fly and this PR doesnt fix it if vmaps actually detects an obstacle in `getObjectHitPos`

jackpoz:
Thanks for the PR anyway :)

PR #23061, opened 2019-02-21
Core/Units: Only update height in SetHover if unit is bellow HoverHeight

sirikfoll:
**Changes proposed:**

- Only update height in SetHover if unit is bellow HoverHeight, this is needed when unit already is above hoverHeight, and SetHover is turned on later, then it's supposed to stay at the same position, take Miniron's Aerial unit for example.

No idea if GetFloorZ() is the correct option here, it worked as intended testing in Mimiron's fight.

- [x] 3.3.5
- [x] master


jackpoz:
Hi @sirikfoll , GetFloorZ() looks like the best choice, please keep an eye anyway or issues reported in the next days if anyone sees something strange about hovering :)

jackpoz:
Thanks for the PR :)

PR #23069, opened 2019-02-24
Core/Movement: Allow using walk when chasing

Sorikoff:
**Changes proposed:**
Allow passing additional parameter to Chase Movement Generator in order to determine whether unit should walk. Currently it looks like that: 
```c
me->GetMotionMaster()->MoveChase(horse, {}, {}, true);
```

**Target branch(es):** 3.3.5

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -585,14 +585,14 @@ void MotionMaster::MoveFollow(Unit* target, float dist, ChaseAngle angle, Moveme
     Add(new FollowMovementGenerator(target, dist, angle), slot);
 }
 
-void MotionMaster::MoveChase(Unit* target, Optional<ChaseRange> dist, Optional<ChaseAngle> angle)
+void MotionMaster::MoveChase(Unit* target, Optional<ChaseRange> dist, Optional<ChaseAngle> angle, bool walk)
 {
     // Ignore movement request if target not exist
     if (!target || target == _owner)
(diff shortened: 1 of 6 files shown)


jackpoz:
Thanks for the PR :)

Killyana:
If we have a creature that always use walk when chasing the target, how can we do that without a cpp?

Sorikoff:
Which creature we are talking about?

jackpoz:
@Killyana I would add a new column somewhere in the DB "chase_walk_mode" with:
- 0 always run, default, current behavior, doesn't break any script
- 1 respect current SetWalk() state
- 2 always walk, might not be needed but while at it we can implement it

ccrs:
noooo, just add a unitflagextraorhowheveritscalled and you will cover all cases

Killyana:
https://www.wowhead.com/npc=37881/wretched-ghoul related to The Sacred and the Corrupt quest, use always walk.

ccrs:
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Creature/CreatureData.h#L35

that

Sorikoff:
`CREATURE_FLAG_EXTRA_UNUSED_13` is free? Or may be I could alter `creature_template_movement`?

Shauren:
You have a table dedicated to movement, why would you even suggest shoving it into extra flags?

ccrs:
that is actually true, I forgot about it, much better 👍

PR #23070, opened 2019-02-24
Core/Objects: use the actual collision height of players and creatures to get accurate liquid status results

### LINKED ISSUE #17987: [3.3.5] Breath Timer when swimming while mounted
**Description**:
Breath Timer gets activated when swimming on the surface while mounted.

Breath Timer starts ticking
Camera gets underwater

https://youtu.be/dGhlHz2HXGs?t=1m40s
Breath Timer should not start
Camera should stay above water

1. Mount
2. Enter Water
3. (Jump until surface is reached)

3.3.5 / 6.x

**TC hash/commit**:  
429700e614c27627becc50bcc45cfa5315b9d4ed

Related functions:
Player::SetInWater
Map::IsUnderWater

I'm pretty sure it's something about GetCollisionHeight that fucks it up.

Ovahlord:
**Changes proposed:**
-  use the hoverheight argument in GetFullTerrainStatusForPosition properly so we use actual collision height values now instead of the default value.
-  Moved UpdatePositionData in Creature::Create a bit down to make sure that the ObjectScale field has been initialized before checking for collision heights so we wont have a crash here
- SetZoneScript will now use the map to get its zoneID so we can call UpdatePositionData later to get the most precise results

- [x] 3.3.5
- [ ] master

Tested ingame.

Underwater when not mounted: 
wowscrnshot_022419_232310

Still underwater but mounted up with bigger collision height due to the increased collision height of the mount:
wowscrnshot_022419_232318

**Closed issues**

closes #17987

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -1022,7 +1022,12 @@ bool Creature::Create(ObjectGuid::LowType guidlow, Map* map, uint32 phaseMask, u
         TC_LOG_ERROR("entities.unit", "Creature::Create(): given coordinates for creature (guidlow %d, entry %d) are not valid (X: %f, Y: %f, Z: %f, O: %f)", guidlow, entry, pos.GetPositionX(), pos.GetPositionY(), pos.GetPositionZ(), pos.GetOrientation());
         return false;
     }
-    UpdatePositionData();
+    {
+        // area/zone id is needed immediately for ZoneScript::GetCreatureEntry hook before it is known which creature template to load (no model/scale available yet)
+        PositionFullTerrainStatus data;
+        GetMap()->GetFullTerrainStatusForPosition(GetPositionX(), GetPositionY(), GetPositionZ(), data, MAP_ALL_LIQUIDS, DEFAULT_COLLISION_HEIGHT);
+        ProcessPositionDataChanged(data);
+    }
 
     // Allow players to see those units while dead, do it here (mayby altered by addon auras)
     if (cinfo->type_flags & CREATURE_TYPE_FLAG_GHOST_VISIBLE)
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1059,7 +1059,7 @@ void WorldObject::_Create(ObjectGuid::LowType guidlow, HighGuid guidhigh, uint32
 void WorldObject::UpdatePositionData()
 {
     PositionFullTerrainStatus data;
-    GetMap()->GetFullTerrainStatusForPosition(GetPositionX(), GetPositionY(), GetPositionZ(), data, MAP_ALL_LIQUIDS);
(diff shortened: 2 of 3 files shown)


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -1018,8 +1018,7 @@ bool Creature::Create(ObjectGuid::LowType guidlow, Map* map, uint32 phaseMask, u
         TC_LOG_ERROR("entities.unit", "Creature::Create(): given coordinates for creature (guidlow %d, entry %d) are not valid (X: %f, Y: %f, Z: %f, O: %f)", guidlow, entry, pos.GetPositionX(), pos.GetPositionY(), pos.GetPositionZ(), pos.GetOrientation());
         return false;
     }
-    UpdatePositionData();
```
Shauren: This breaks SetZoneScript() (called in CreateFromProto). area/zone info must already be present there

Shauren (changes requested):

src/server/game/Entities/Creature/Creature.cpp
Ovahlord: So we have a little problem here then. Object_field_scale must be initialized before calling for collision height which also happens in CreateFromProto otherwise the CollisionHeight call will crash on startup when creating battlefield npc's

jackpoz: would that be a problem only with battlefield npcs ? maybe then that part has to be changed ?

Ovahlord: It has been changed already. Instead of using cached data the map helper is being used now.

Ovahlord:
Pushing this a bit. Requested changes has been made a while ago already so what's the matter to keep the PR pending?

jackpoz:
@Shauren any other comment about this PR or can it be merged ?

ForesterDev:
I've got a crash when create new character. I think, you should also move UpdatePositionData() after InitializeDisplayIds() for players too

jackpoz:
>at /home/sources/server.lk/src/server/game/Entities/Unit/Unit.cpp:19420

does your Unit-cpp have 19420+ lines ?

Killyana:
Confirm the crash in a clean core, when you click "create the character"

illfated:
Thank you, that means the crash information belongs to a separate issue, right?

Killyana:
No this PR introduced the crash

illfated:
Thank you, I obviously misinterpreted the first version of your previous comment.

Shauren (approved):

Shauren:
Well, I couldn't find the relation between breath bar and collision height on retail, its slightly less than collision but different for each race (not less by a constant and not a % of collision height)

Killyana:
I tested with a Tauren male, and it looks like the calculations were made exactly for him.

With a Tauren female it's almost hard not getting the breath bar upon hitting X once the bar appears.

With a female blood elf you can go slightly under water  without getting the breath bar.

If I morph my character the behavior doesn't change, so it doesn't looks to be related to the model.

On retail they made some changes and the breath bar is no longer triggered at the same time with the camera being blocked under water.

jackpoz:
summing up, can this PR be merged ?

jackpoz:
> summing up, can this PR be merged ?

jackpoz:
Thanks for the PR :)

PR #23072, opened 2019-02-25
Core/GameObject: Set loot state to Not ready when a gob is deactivated

### LINKED ISSUE #22364: Zelfrax is Summoned Twice
**Description:**

Cleansing Witch Hill Zelfrax summon twice.

**Current behaviour:**

The script is working correctly but after you kill Zelfrax, its like the spell loops by itself and Zelfrax is summoned again for a second time making you have to kill the mob twice. 

**Expected behaviour:**

Zelfrax should only be summoned one time.

Have the quest Cleansing Witch Hill (11183)
Use the torch to summon Zelfrax
Kill the first Zelfrax, and you will notice the beam starts again and a second one is summoned.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

ca7ee982db4d25c464acf6daadc150d58dafa016

**Operating system:**

Windows Server 2016

Killyana:
**Changes proposed:** Set loot state to Not ready when a gob is deactivated, this will fix the call of loot state ready when a gameobject despawns.

**Target branch(es):** 3.3.5

**Issues addressed:** Closes #22364

**Tests performed:**Yes

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -779,7 +779,7 @@ void GameObject::Update(uint32 diff)
                 return;
             }
 
-            SetLootState(GO_READY);
+            SetLootState(GO_NOT_READY);
 
             //burning flags in some battlegrounds, if you find better condition, just add it
             if (GetGOInfo()->IsDespawnAtAction() || GetGoAnimProgress() > 0)


jackpoz:
No clue what this could break (if anything)

Killyana:
This SetLootState(GO_READY) (on respawn/despawn) is called only by gameobjects that have negative respawntime, for other gobs, this will never be called when a gob despawn or respawn.

I tested all possible cases with SAI and DB scripts, and every thing looks fine.
For cpp scripts I cannot find any gameobject with negative respawn time and a hook triggered with loot state ready.

jackpoz:
I would like to test this with different gameobject types, in particular the ones handled in https://github.com/TrinityCore/TrinityCore/blob/b967ba1e31e011814a69ebe1d370542dfe4fba2e/src/server/game/Entities/GameObject/GameObject.cpp#L431

Killyana:
I made lot of tests with gobs type list handled case GO_NOT_READY, nothing changed.

jackpoz:
Thanks for the PR :)

jackpoz:
@Killyana is this the cause of recent gameobject issues ?

Killyana:
What issues?

PR #23081, opened 2019-03-03
Core/Movement: Allow using run when moving randomly 

Sorikoff:
**Changes proposed:**
Allow setting random run state for each creature through DB table `creature_template_movement`/`creature_movement_override`. Values for new column `Random` are following:

0. walk always;
1. respect script settings;
2. run always.

**Target branch(es):** 3.3.5

**Tests performed:** Does build


jackpoz:
could you (or anyone) check ingame if it works too ? then I think we can merge it

Sorikoff:
Works ingame.

illfated:
Suggestion for cosmetic change of the SQL file (the way Kilyana usually commits SQL content):
```sql
--
ALTER TABLE `creature_template_movement` ADD `Random` tinyint(3) UNSIGNED NOT NULL DEFAULT '0';
ALTER TABLE `creature_movement_override` ADD `Random` tinyint(3) UNSIGNED NOT NULL DEFAULT '0';
```
(remove the blank line at the end and add empty comment line at the start)

PR #23090, opened 2019-03-04
Core/Players: Fix logic in CanSeeSpellClickOn

### LINKED ISSUE #22689: Core/Creature: Vehicle get the npcflag Spellclick automaticly
**Description:** if a npc has vehicleid !=0 he will get the npc flag = 16777216 even if npcflag=0 in the DB.

1. .go c id 23693
2. notice the spell click is activated event if the  npcflag=0 and the creature doesn't have any spell_click

There's a unit_flags2 UNIT_FLAG2_PREVENT_SPELL_CLICK= 8192 it disable the spell click on the creature but it also effect the auto attack, you will not be able to start an attack with a right click
UPDATE `creature_template` SET unit_flags2=unit_flags2|8192 WHERE `entry`=23693;

**Branch(es):** both

**TC rev. hash/commit:** rev. cf2ca013f7a

sirikfoll:
**Changes proposed:**

-  Block CanSeeSpellClickOn for the npc if there is no entry for it in SpellClick data.

**Isssues**:  Closes #22689

- [x] 3.3.5
- [x] master

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -25243,7 +25243,7 @@ bool Player::CanSeeSpellClickOn(Creature const* c) const
 
     auto clickBounds = sObjectMgr->GetSpellClickInfoMapBounds(c->GetEntry());
     if (clickBounds.begin() == clickBounds.end())
-        return true;
+        return false;
 
     for (auto const& clickPair : clickBounds)
     {


jackpoz:
Is there any example or how to reproduce case ?

sirikfoll:
https://github.com/TrinityCore/TrinityCore/issues/22689 .go c id 23693

jackpoz:
shouldn't this be managed on startup ? Does the check in ObjectMgr::LoadNPCSpellClickSpells() log anything at all ? or how come does this issue happen at all ?

sirikfoll:
Vehicle.cpp adds it again
https://github.com/TrinityCore/TrinityCore/blob/f039c30cb8cf0a4970b326dcd4c1939458029955/src/server/game/Entities/Vehicle/Vehicle.cpp#L51-L53

jackpoz:
It sounds quite complicated how the data is handled (removed, re-added, handled like if it wasn't there) but as long as it doesn't brake anything it should be alright

PR #23110, opened 2019-03-12
Revert "Core/AuctionHouse: use DisableMgr to disable specific items"

jackpoz:
Reverts TrinityCore/TrinityCore#19289


jackpoz:
To do: restore the sql that github deleted, add another sql to delete disable type 9 from db

Issue #20466, opened 2017-09-29
Minor cosmetic issue with items created by spells


Wyreth:
**Description:**

TL;DR: some items get the "Created By" tag when they shouldn't. Some items appear in chat log as "Created item" instead of "Received item" in cases where they shouldn't.

More details:

Currently, almost all the items created by spells are given the "Created By" tag and show the "Created Item X" in chat. This is handled [here](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellEffects.cpp#L1610)

However, only some items created via professions should have that. Items like [Benediction](https://youtu.be/rb2LhTzv-eY?t=440) and [Eternal Fire](https://youtu.be/U4YPNFpuFR8?t=24) shouldn't, yet in TC they both show that behaviour.

Apparently, only profession-creaftable items that aren't consumables and aren't obtainable elsewhere have the "Created By" tag. Gems created by Alchemy don't, sockets created by jewelcrafters do. 

And only player class spells (Conjure Food, Healthstone, Soul Drain) and profession spells should trigger the "Created item <x>" chat message. Everything else should show the "Received item <x>" chat message.

I don't know where leahters crafted with Leatherworking and things like Engineering intermediary items fall in.

Maybe there's some flag for this behaviour, but I haven't been able to find anything yet.

**Branch(es):** both.

**TC rev. hash/commit:** 3994286

**TDB version:** 335.63

Treeston:
Hmm... how about this classification.

Items created with `CREATE_ITEM` that are either:

* Conjured items
* Equippable

should have a `<created by>` tag. All other items should not.

PS: Benediction does not use `CREATE_ITEM`

Shauren:
Hmm thats not it, example
enchanting rods: get <created> tag
engineering gnomish army knife: no tag

Aokromes:
humm maybe related to that flag: tailoring stackable things have made by thing and they don't must have.

illfated:
Same goes for Leatherworking, creating cured hides or higher leather from lower leather also adds the green text `<Made by (Name)>` to the items.

PR #23119, opened 2019-03-17
Core/Item: Item Signature

PolarCookie:
**Changes proposed:**
items
Many items are getting incorrectly signed.
Items that should have signatures are profession crafted items that do not stack.

Disclaimer; this is mostly guesswork based on foggy memory and it's hard to find supporting evidence other than a feeling that it's incorrect for stacked items to have signatures, I swear it wasn't like that on retail.

What I could find
https://www.youtube.com/watch?v=leCvyUCqTcI
at 1:03 he hovers over a Bolt of Frostweave (item class 7, subclass 5, stacks to 20) without signature
https://www.youtube.com/watch?v=pgckZnEeq7w
at 2:52 he hovers over a Box of Bombs (item class 7, subclass 2, stacks to 1) with signature

**Target branch(es):**
3.3.5

Builds and works

### DIFF
--- a/src/server/game/Entities/Item/ItemTemplate.cpp
+++ b/src/server/game/Entities/Item/ItemTemplate.cpp
@@ -29,7 +29,7 @@ bool ItemTemplate::HasSignature() const
         Class != ITEM_CLASS_CONSUMABLE &&
         Class != ITEM_CLASS_QUEST &&
         (Flags & ITEM_FLAG_NO_CREATOR) == 0 &&
-        ItemId != 6948 /*Hearthstone;*/
+        ItemId != 6948; /*Hearthstone*/
 }
 
 bool ItemTemplate::CanChangeEquipStateInCombat() const
--- a/src/server/game/Entities/Item/ItemTemplate.cpp
+++ b/src/server/game/Entities/Item/ItemTemplate.cpp
@@ -23,6 +23,15 @@
 
 #include "Packets/QueryPackets.h"
 
+bool ItemTemplate::HasSignature() const
+{
+    return GetMaxStackSize() == 1 &&
+        Class != ITEM_CLASS_CONSUMABLE &&
+        Class != ITEM_CLASS_QUEST &&
+        (Flags & ITEM_FLAG_NO_CREATOR) == 0 &&
+        ItemId != 6948 /*Hearthstone;*/
+}
+
 bool ItemTemplate::CanChangeEquipStateInCombat() const
 {
     switch (InventoryType)
--- a/src/server/game/Entities/Item/Item.cpp
+++ b/src/server/game/Entities/Item/Item.cpp
@@ -730,6 +730,21 @@ bool Item::CanBeTraded(bool mail, bool trade) const
     return true;
 }
 
+bool Item::HasSignature() const
+{
+    const ItemTemplate * temp = GetTemplate();
+
+    if (temp->Stackable == 1)
+        if (temp->Class == ITEM_CLASS_CONTAINER ||
+            temp->Class == ITEM_CLASS_WEAPON ||
+            temp->Class == ITEM_CLASS_ARMOR ||
+            temp->Class == ITEM_CLASS_GEM ||
+            temp->Class == ITEM_CLASS_TRADE_GOODS ||
+            temp->Class == ITEM_CLASS_QUIVER)
+            return true;
+    return false;
+}
+
 bool Item::HasEnchantRequiredSkill(Player const* player) const
 {
     // Check all enchants for required skill
--- a/src/server/game/Entities/Item/Item.cpp
+++ b/src/server/game/Entities/Item/Item.cpp
@@ -730,28 +730,6 @@ bool Item::CanBeTraded(bool mail, bool trade) const
     return true;
 }
 
-bool Item::HasSignature() const
-{
-    const ItemTemplate * temp = GetTemplate();
-
-    if (temp->Stackable > 1)
-        return false;
-
-    if (temp->Class == ITEM_CLASS_CONTAINER ||
-        temp->Class == ITEM_CLASS_WEAPON ||
-        temp->Class == ITEM_CLASS_ARMOR ||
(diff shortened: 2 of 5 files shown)


Aokromes:
indeed, profession stackable items don't must have <made by> thing

Killyana:
Closes https://github.com/TrinityCore/TrinityCore/issues/20466

illfated:
I agree as well. Stackable items should not have the creator name added in the flavor text.
BTW, would you consider using the PR title standard **Core/Item:** as prefix?
I would also like to suggest a longer title: **Core/Item: don't add creator signature to stackable items**
(if you have something else in mind, feel free to use that instead)

PolarCookie:
@Killyana good thing you're a living encyclopedia of this repository.
Benediction/Anathema: no tag
Thorium Rifle: all the intermediate materials did not get a tag, but the rifle did
Fel Iron Rod: tagged
Gnomish Army knife: tagged, and @Shauren said it shouldn't so I dunno

What's up with the 
https://github.com/TrinityCore/TrinityCore/blob/4ce6d6b9f643253092812fe6c6f172a41a6c51a7/src/server/game/Entities/Item/ItemTemplate.h#L191
https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130222/item+template
says it's unk8

illfated:
The wiki page is outdated compared to the source code. See
https://github.com/TrinityCore/TrinityCore/commit/ef29f32965921f80f5c1ac21fa74e92a6a98babd#diff-cd45d8479fb6b19e7fed3c2cd11675a2R189
for details about when it was added (by Shauren).

PolarCookie:
that flag must have meaning for some other context
`SELECT entry, class, subclass, name, stackable, flags FROM item_template WHERE flags & 131072 = 131072;`
didn't give very convincing results.

illfated:
Maybe the flag needs to be added to more items than the ones currently using it?

Shauren:
No, the meaning of that flags is correct - these items are all created by spells

PolarCookie:
but is it related to what should be signed by the creator?

Shauren:
Yes, probably used in addition to other rules (such as stackable)

PolarCookie:
The 2nd commit here could be better.

PolarCookie:
What items should be signed which were not from trade skills?

PolarCookie:
@Killyana PR done for my part, important thing for me was the stacking signature.
This current version is the same as before but disallowed signed stacks or those flagged as NO_CREATOR and Soul Shard is signed because I did find this one at 1:38
https://www.youtube.com/watch?v=g_8Bg2pRsLY

src/server/game/Entities/Item/ItemTemplate.cpp
```diff
@@ -23,6 +23,15 @@
 
 #include "Packets/QueryPackets.h"
 
+bool ItemTemplate::HasSignature() const
+{
+    return GetMaxStackSize() == 1 &&
+        Class != ITEM_CLASS_CONSUMABLE &&
+        Class != ITEM_CLASS_QUEST &&
+        (Flags & ITEM_FLAG_NO_CREATOR) == 0 &&
+        ItemId != 6948 /*Hearthstone;*/
```
Killyana: Why the hearthstone? It doesn't have any signature.

PolarCookie: It's going to get a signature unless that line is there because of https://db.rising-gods.de/?spell=3286

Killyana: But the spell is casted by the npc on the player, in this case the name of the npc appears in the signature?

PolarCookie: https://github.com/TrinityCore/TrinityCore/commit/fdd8d2f3cc7108043f6bad036b994770f904866f#diff-6e2ad27531ee8eb7bafce8a41909816aR3072
originated here, HS is signed with the player name.

PolarCookie: WoWScrnShot_031919_145154

PolarCookie: That's with IdemId != 6948 commented out.

PolarCookie:
https://www.youtube.com/watch?v=QGKjFtCNHHA
https://github.com/TrinityCore/TrinityCore/commit/d01d4498ede1f0ff57e762f5c99f39012d9adcd6
heh

jackpoz:
Is there any sniff about soul shard ?

Shauren:
Managed to find one wotlk sniff of soul shard in our storage
```
[70] UpdateType: CreateObject1
[70] GUID: Full: 0x428000025FF03B03 Type: Item Low: 10199513859
[70] Object Type: 1 (Item)
[70] Update Flags: 16 (LowGuid)
[70] Low GUID: 1609579267
[70] OBJECT_FIELD_GUID: Full: 0x428000025FF03B03 Type: Item Low: 10199513859
[70] OBJECT_FIELD_TYPE: 3
[70] OBJECT_FIELD_ENTRY: 6265
[70] OBJECT_FIELD_SCALE_X: 1
[70] ITEM_FIELD_OWNER: 45718766/2.727338E-37
[70] ITEM_FIELD_OWNER + 1: 41943040/1.880791E-37
[70] ITEM_FIELD_CONTAINED: 904815653/1.776153E-06
[70] ITEM_FIELD_CONTAINED + 1: 1115684866/64.00002
[70] ITEM_FIELD_CREATOR: 45718766/2.727338E-37
[70] ITEM_FIELD_CREATOR + 1: 41943040/1.880791E-37
[70] ITEM_FIELD_STACK_COUNT: 1/1.401298E-45
[70] ITEM_FIELD_FLAGS: 32769/4.591915E-41
[70] ITEM_FIELD_PROPERTY_SEED: 76937824/3.525966E-36
[70] ITEM_FIELD_CREATE_PLAYED_TIME: 2678265/3.753049E-39
```

CREATOR is set.

jackpoz:
Thanks for the PR :)

PR #23125, opened 2019-03-21
Cherrypicking db 3.3.5a -> master

funjoker:
**Changes proposed:**

-  Cherrypicks

- [ ] 3.3.5
- [x] master


jackpoz:
ren't _335 sql files supposed to be used only in 335 branch ?

illfated:
Yes, they as good as always contain game content removed in Cataclysm.

funjoker:
Not all stuff changed in Cata. These npcs are actually still inside Game.  But untested tbh

msoky:
probably we should take a look on all "world_335.sql" which was added before years if they are part of master branch now ? or it is case for only this 5 commits?

funjoker:
Yeah take a Look. Actually i‘m more interessted into Core fixes.

msoky:
is this mergable?

illfated:
Seeing that this is only SQL content, it should be tested before merging by someone with current master branch compiled + up to date master TDB already running.

PR #23135, opened 2019-03-26
Propose new config Gm.AllowAchievements

d0d0:
**Changes proposed**:

- this PR proposes to Enable/Disable earning achievements for GMs. Current situation blocks achievements only when `.gm on`. This can be issue eg for new servers when GM does `.level 79` without `.gm on` and Realm Firsts are gone

- new worldserver config `GM.AllowAchievements`

**Target branch(es)**: 335

**Tests performed**: tested in-game, results bellow in table

|         | Config True    | Config False   |
|---------|----------------|----------------|
| .gm on  | No achievement | No achievement |
| .gm off | Achievement    | No achievement |

Player accounts are not affected (tested)


illfated:
Nice and useful feature, I like the idea a lot. In my opinion, `0 - (Disabled)` should be the default, because it is too easy to forget to disable it when creating a new GM character, as well as the fact that it is easier to repeat the action to get the achievement on a GM char than to remove it / reset the world achievements. Then again, the current setting keeps it unchanged from what it has been up to now. Looking forward to read what the TC owners will recommend.

Aokromes:
> Nice and useful feature, I like the idea a lot. In my opinion, `0 - (Disabled)` should be the default, because it is too easy to forget to disable it when creating a new GM character, as well as the fact that it is easier to repeat the action to get the achievement on a GM char than to remove it / reset the world achievements. Then again, the current setting keeps it unchanged from what it has been up to now. Looking forward to read what the TC owners will recommend.

deleting the character who won the achievement makes it available for other characters.

illfated:
> deleting the character who won the achievement makes it available for other characters.

OK, that is a valid argument (I forgot that part). Is that how it works for all "Realm First" achievements?

ccrs:
I support this, really useful 👍

D0d0:
I do not have many experiences with RBAC, but I think it is used primary for permissions not for "disabling" things, so conf is maybe better way

illfated:
Yes and no. RBAC can work as a disable function too by creating a permission to deny or grant the ability to get achievements.

Using RBAC to deny or grant access to achievements can be switched on and off from the command-line while the server is running,
A conf setting requires a server reboot to change enable / disable it.

Using RBAC, you can also choose to have it enabled regardless of `.gm on` / `.gm off` state of the character,
which in some scenarios would also be preferable in order to test things unrelated to achievements.

D0d0:
Well @illfated, you propose to switch from config file to RBAC and add function for disabling achievements with `.gm on` / `.gm off`?

illfated:
Good question, I was a bit tired when I wrote my previous comment.
I don't have a strong opinion on how it should be done.
Maybe one of the TC members can suggest how it should be done.
I just wanted to share my thoughts on the matter, even if I don't know the solution.

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2064,6 +2064,14 @@ GM.InWhoList.Level = 3
 
 GM.StartLevel = 1
 
+#
+#    GM.AllowAchievements
+#        Description: Allow GM accounts to have achievements.
```
jackpoz: maybe explain a bit more in detail that setting this to 1 but having ".gm on" will not give gms achievements

src/server/game/Achievements/AchievementMgr.cpp
```diff
@@ -511,8 +511,8 @@ void AchievementMgr::ResetAchievementCriteria(AchievementCriteriaCondition condi
 {
     TC_LOG_DEBUG("achievement", "AchievementMgr::ResetAchievementCriteria(%u, %u, %u)", condition, value, evenIfCriteriaComplete);
 
-    // disable for gamemasters with GM-mode enabled
-    if (m_player->IsGameMaster())
+    // disable for gamemasters with GM-mode enabled or when achievements for GM are disabled
+    if (m_player->IsGameMaster() || (m_player->CanBeGameMaster() && !sWorld->getBoolConfig(CONFIG_ALLOW_GM_ACHIEVEMENTS)))
```
jackpoz: I would prefer a new method in AchievementMgr accepting a Player* as parameter and returning bool as a wrapper around these checks since they are repeated already in 3 different places

src/server/worldserver/worldserver.conf.dist
illfated: Maybe something like the wording in the comment in `AchievementMgr.cpp` ?

jackpoz:
Hej @D0d0 , are you doing any additional change to the PR ? otherwise I'm going to wrap it up and merge it

D0d0:
@jackpoz no, I think it is done from my side. If you do not have any suggestions, you can merge it.

Aokromes:
I think i have a better idea for this, how about simply don't store gm wining realm 1st achievements?

ccrs:
@Aokromes I think this is actually ok.

Though I do really prefer a new RBAC permission over a new config.
It just makes more sense, you can even use it to specify wich type of GM cannot earn achievements.

Aokromes:
imho my idea is better, the only achievements causing "troubles" are the realm 1st ones, if we dont store gms wining them we don't even need configs.

illfated:
Could you perhaps make up some kind of pseudo code for how it should work, according to your idea?

Aokromes:
```
if achievement isrealmd1st && player!=gmaccount
     save to achievements
else
      ignore.

illfated:
Fair enough. I thought your idea included some form of RBAC classification, but OK.
My preference would be to allow a gm account (with GM OFF) to get achievements,
but then again, I am only a single person playing as if it was a standalone game.

Aokromes:
on my way gms can get achievements, and normal achievements gets saved, and realm 1st, the one causing this pr to exist arent' saved so they are available for players and we can get rid of one config.

illfated:
Fair enough. Lone players should not worry too much about "Realm First" achievements anyway -- unless they suffer from serious Dissociative identity disorder of some kind.

ccrs:
Again, consistency, use RBAC, its not that difficult to implement 💃

illfated:
I still agree with you, but I presume it can be tricky to see the outlines of the RBAC solution for casual players like me, as well as @D0d0 .

jackpoz:
> Again, consistency, use RBAC

RBAC bound to account level or ... ?

ccrs:
What we do usually is just to add it to the existing default permission sets

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/sql/base/auth_database.sql#L298
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/sql/base/auth_database.sql#L330
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Accounts/RBAC.h#L61

Issue #23133, opened 2019-03-24
[3.3.5] Crash: StopCastingCharm


TheWinchesters:
**Description:**
There are closed issues about this crash but it looks like it is happening again.
I updated the core yesterday (it was a two or three months old core) and I did not have this crash before.
I went from commit 4337eefc8f77ec2f64f2bda747292ea4e50199ea to commit 858782f68a33ca8f2aacee3e1d9b1435b2edfb30
There are comments about this crash in "Spellaura redux" https://github.com/TrinityCore/TrinityCore/pull/22794.
I do not know if it is related or it is something else between the two commits mentioned above.

Crashlog: https://gist.github.com/TheWinchesters/1062cb582a77dec9b9bbf3bcb2d77c7a

Note 1: This happened in a server with basic custom things not related to this crash (Transmog, Passive anticheat, ...)
Note 2: Tested in a clean core too

**Branch(es):** 3.3.5
**TC rev. hash/commit:** 858782f68a33ca8f2aacee3e1d9b1435b2edfb30
**Operating system:** Linux

TheWinchesters:
How to reproduce:
1. .go c id 30500
2. .quest add 13045
3. Do a spam of clicks on the npc 30500 (A drake mount)

Killyana:
npc_spell_click_spell=https://woehead.way-of-elendil.fr/?spell=56922
it forcecast https://woehead.way-of-elendil.fr/?spell=56679
That trigger https://woehead.way-of-elendil.fr/?spell=56678

I don't think the spell_script 'spell_gen_vehicle_control_link'  handle this case
@PolarCookie

Killyana:
I tested in 2 years old core and spamming the npc spawns also a vehicle per click, so in this case we have 2 issues, the crash and the spellclick condition

jackpoz:
There are 2 TC_LOG_FATAL and 1 ABORT in same control flow, that's quite overkill.

PolarCookie:
https://github.com/TrinityCore/TrinityCore/blob/ac8238c6342675041800eb49a3b5a0980da1e29d/src/server/game/Entities/Vehicle/Vehicle.cpp#L763
Put a breakpoint here; the times when the crash happens this execute is called twice, and it fails on the second. Normally it's only called once and everything's fine.
No idea why it randomly gets called twice.

Killyana:
It's called twice if you spam the spell_click,.
When the crash happen you get this in the server.log:
[...]

PR #23136, opened 2019-03-26
Core: VehicleJoinEvent crash fix

PolarCookie:
**Changes proposed:**
fix for https://github.com/TrinityCore/TrinityCore/issues/23133

**Issues addressed:** 
#23133

### DIFF
--- a/src/server/game/Entities/Vehicle/Vehicle.cpp
+++ b/src/server/game/Entities/Vehicle/Vehicle.cpp
@@ -785,6 +785,11 @@ bool VehicleJoinEvent::Execute(uint64, uint32)
         return true;
     }
 
+    //It's possible that multiple vehicle join
+    //events are executed in the same update
+    if (Passenger->GetVehicle())
+        Passenger->ExitVehicle();
+
     Passenger->SetVehicle(Target);
     Seat->second.Passenger.Guid = Passenger->GetGUID();
     Seat->second.Passenger.IsUnselectable = Passenger->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_NOT_SELECTABLE);


Shauren:
I'll be blunt here (keep https://github.com/TrinityCore/TrinityCore/issues/23133#issuecomment-476837142 in mind). I hate this fix. I want vehicle auras to be single target again

PolarCookie:
https://github.com/TrinityCore/TrinityCore/pull/22794
Because a lot of vehicle auras are of the type
Effect 1: vehicle control on vehicle
Effect 2: dummy aura on the caster
and this only worked when aura applications were shared, but this caused issues with other multi target spells.

PolarCookie:
And I'll be blunt that it's not a good idea to hack vehicle auras onto a single target system where they don't belong.

Shauren:
They clearly do - there can only be one control vehicle ever

PolarCookie:
They clearly are not flagged as such, and until I see the Blizzard source code I don't assume how they did things.

jackpoz:
any update on this ?

PolarCookie:
As far as I'm concerned it's finished as is; the issue was that `Passenger->SetVehicle(target)` was called while player was already set to a previous vehicle, and that caused issues with the charmer/charmed variables being still set to the previous vehicle later.

src/server/game/Entities/Vehicle/Vehicle.cpp
```diff
@@ -785,6 +785,11 @@ bool VehicleJoinEvent::Execute(uint64, uint32)
         return true;
     }
 
+    //It's possible that multiple vehicle join
+    //events are executed in the same update
```
illfated: Cosmetic detail, but seeing the length of the code before & after this addition, maybe the comment could be reduced to 1 line?

jackpoz:
wouldn't it be better to do some check when queueing a vehicle join event instead of when executing it ?

PolarCookie:
That's the issue, multiple vehicle join events can get queued up before any are executed, all will gladly pass any test because of this.

PolarCookie:
The test you want is already happening https://github.com/TrinityCore/TrinityCore/blob/f7290509005d3b93bb62149faf3add0bb7f54f51/src/server/game/Entities/Unit/Unit.cpp#L12677-L12686

jackpoz:
Well that doesn't check if there is any join event in the queue. Also is it ok to execute only the first one or should only the last one be executed ?

src/server/game/Entities/Vehicle/Vehicle.cpp
illfated: What I want to suggest here is both to drag back line 2 (789) into line 1 (788) and to add a blank space after the comment backslashes (`//`) like this:
```cpp
    // It's possible that multiple vehicle join events are executed in the same update
```
( I simply thought that this format would follow the code style standard a bit more.)

PolarCookie:
While you're at it, delete `Unit::_EnterVehicle` all together, move all modification done there to  `VehicleJoinEvent::Execute`. Don't think you even need to search the queue that way, would be a significant overhead.

jackpoz:
merging this for now, then we see if there's a better way to handle it

jackpoz:
Thanks for the PR :)

PR #23150, opened 2019-04-02
Fix for the new version of MariaDB

vbelousoff:
CR_INVALID_CONN_HANDLE was dropped since MariaDB 10.2

**Changes proposed:**

-  Checking definition of CR_INVALID_CONN_HANDLE

**Target branch(es):**

- [ ] 3.3.5
- [ ] master

**Issues addressed:**

- N/A

**Tests performed:** 

- Builds

### DIFF
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -512,7 +512,9 @@ bool MySQLConnection::_HandleMySQLErrno(uint32 errNo, uint8 attempts /*= 5*/)
     {
         case CR_SERVER_GONE_ERROR:
         case CR_SERVER_LOST:
+#ifdef CR_INVALID_CONN_HANDLE
         case CR_INVALID_CONN_HANDLE:
+#endif
         case CR_SERVER_LOST_EXTENDED:
         {
             if (m_Mysql)


jackpoz:
where was it dropped from ? which package ? strange because I found it at https://github.com/MariaDB/server/blob/5abc79dd7ab2fccb4b05ca38a512ec816d2f8e52/include/errmsg.h#L99

vladbelousoff:
> where was it dropped from ? which package ? strange because I found it at https://github.com/MariaDB/server/blob/5abc79dd7ab2fccb4b05ca38a512ec816d2f8e52/include/errmsg.h#L99

seems like it pulls the headers from https://github.com/MariaDB/mariadb-connector-c/blob/master/include/errmsg.h

I used yaourt for installing package mysql and it does not contain CR_INVALID_CONN_HANDLE
Probably it's mysql package distribution bug.

Shauren:
I looked at https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/mariadb and the downloaded source does have `CR_INVALID_CONN_HANDLE` in its errmsg.h
Am I looking at wrong place?

Shauren:
I did a bit of digging, looks like arch pulls the source from a mirror of this https://downloads.mariadb.org/mariadb/10.3.13/ - zip source structure looks to be taken from github repository (MariaDB/server) which in turn has 'libmariadb' directory linked as git submodule pointing to mariadb-connector-c

In my opinion this is a directly a mariadb issue, not any distro package fault

Shauren:
Not sure this is the right approach, I just checked documentation of all 3 functions that can return errors that we handle, `mysql_query`, `mysql_stmt_bind_param`, `mysql_stmt_execute` and none of them return `CR_INVALID_CONN_HANDLE`

vladbelousoff:
> Not sure this is the right approach, I just checked documentation of all 3 functions that can return errors that we handle, `mysql_query`, `mysql_stmt_bind_param`, `mysql_stmt_execute` and none of them return `CR_INVALID_CONN_HANDLE`

Should we just delete the line then or not?

Shauren:
Probably

jackpoz:
let's just delete it then ?

PR #23157, opened 2019-04-03
Core support for breadcrumb quests

### LINKED ISSUE #22264: [DEV]Core/Quest: Breadcrumb quests requirements
**Description:** RewardNextQuest  was used to handle the availability of Breadcrumb quests, but it's no longer the case since some changes done years ago, example:

Quest A and Quest B are Breadcrumb for quest C, in this case we put RewardNextQuest=C for A and B. 
But  RewardNextQuest  isn't supposed to handle this kind of requirements:
``` 
RewardNextQuest: (Old field name: NextQuestIdChain)
The quest entry from a creature or gameobject that ends a quest and starts a new one. The result is, that if you end the quest, the new quest instantly appears from the quest giver.
```
So we need a new field on `quest_template_addon` called `Breadcrumb`, that makes the quest not available if the quest on `Breadcrumb` is taken/completed/rewarded.

PS: I know there's the table conditions that could handle this, but it's not the correct way.

Ref: https://github.com/TrinityCore/TrinityCore/issues/20452 https://github.com/TrinityCore/TrinityCore/issues/22216

**Branch(es):** both

**TC rev. hash/commit:** rev. 41982aa30026

PolarCookie:
**Changes proposed:**
@Killyana 
Solution to #22264, support breadcrumb quests at the core level.
Follows the pattern mentioned here https://github.com/TrinityCore/TrinityCore/issues/22264#issuecomment-452668865

**Target branch(es):** 3.3.5
3.3.5

**Issues addressed:** Closes #22264 

Tested with Rejold's New Brew/Shimmer Stout, which is included in this commit.


Killyana:
Nice, for the DB data we can probably take a lot from RewardNextQuest

PolarCookie:
I was thinking of translating https://github.com/TrinityCore/TrinityCore/issues/22264#issuecomment-429515457 when/if this gets merged, but that idea works too.

PolarCookie:
As a test, added a quest with two breadcrumbs, which are already in an exclusive group with eachother, working as expected.

PolarCookie:
I'm working on translating @ZenoX92's list, but ran into an issue.
https://db.rising-gods.de/?quest=10605 is a breadcrumb for 
https://db.rising-gods.de/?quest=1472 which is in an exclusive group with 
https://db.rising-gods.de/?quest=1507
If you pick the Orgrimmar questline for Succubus, what should happen with the first breadcrumb?

Killyana:
The actual data is wrong as:  https://wow.gamepedia.com/Carendin_Summons is for blood elfs only that leads to https://wow.gamepedia.com/Devourer_of_Souls_(Undercity) as breadcrumb
Once the player get one of the "Devourer of Souls" quest, the second one turns not available.

PolarCookie:
Ye, but if a Blood Elf picks up the Org Devourer of Souls, thus closing the Undercity version, should Caredin Summons be disabled too?

Killyana:
Of course.

Killyana:
If we get 4 quests like this:
A1 breadcrumb for B1
A2 breadcrumb for B2
B1 and B2  are in an ExclusiveGroup > 0
What mut happen is: taking any of the 4 quests must make the 3 others not available.
ExclusiveGroup already do that of B1 an B2, but for A1 and A2 I don't know what will happen with this PR.

PolarCookie:
I'm testing a solution for Caredin Summons, got an example I could test for a 4 quest set like what you described? the Caredin Summons situation is missing the A2 so to speak.

Killyana:
I don't remember any quest chain that must behave like this, but I'm sure there's many.
Jut add a any quest as breadcrumb for https://db.rising-gods.de/?quest=1507 to test this situation.

PolarCookie:
Both Devourer of Souls disable Caredin Summons now, and it should be working for any Ax

PolarCookie:
Not done yet, have to disable B2 as long as A1 is in the log, right?

Killyana:
If any of the quest is in the log the 3 others cannot be taken.
If one of the breadcrumb quests is rewarded  the 3 others are available.
If A1 and A2 quests are rewarded,  B1 and B2 are available.
If B1 or B2 is in taken, complete or rewarded other quest are not available.

PolarCookie:
@Killyana if A1 and A2 are breadcrumb for B2, should A2 be available if A1 is rewarded?
Not obvious where/if there should be any exclusive group with breadcrumb quests.

Currently working on Redridge Rendezvous which looks like this, have to implement some recursive check on arbitrary breadcrumb trails.

9f78dbfd-ae20-41ac-9e7b-bc9f7fa5c867

Killyana:
Yes A2 must remains available

PolarCookie:
Feature is complete, I'd recommend testing out with the rogue questline pictured, any of the You're Hired! breadcrumbs or Tinkee Steamboil (both breadcrumb and target quest has the same prequest).

Notes for the wiki:
1) Breadcrumb loops are strictly forbidden.
2) Breadcrumbs to non-existant quests are strictly forbidden.
3) If BreadcrumbForQuestId is assigned; PrevQuestId can be assigned but NextQuestId and ExclusiveGroup must be 0.

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -4991,6 +4994,20 @@ void ObjectMgr::LoadQuests()
                 nextQuestItr->second.DependentPreviousQuests.push_back(qinfo->GetQuestId());
         }
 
+        if (uint32 breadcrumbForQuestId = std::abs(qinfo->_breadcrumbForQuestId))
+        {
+            auto targetQuestItr = _questTemplates.find(breadcrumbForQuestId);
+            if (targetQuestItr == _questTemplates.end())
+            {
+                TC_LOG_FATAL("sql.sql", "Quest %u is a breadcrumb for quest %u, but no such quest exists", qinfo->_id, breadcrumbForQuestId);
+                exit(1);
```
jackpoz: please just gracefully handle the case and log only a TC_LOG_ERROR, there's not need to shutdown worldserver for just a quest

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -5019,6 +5036,36 @@ void ObjectMgr::LoadQuests()
...
+        std::set<uint32> questSet;
+
+        while(breadcrumbForQuestId)
+        {
+            //a previously visited quest was found as a breadcrumb quest
+            //breadcrumb loop found!
+            if (!questSet.insert(qinfo->_id).second)
+            {
+                TC_LOG_FATAL("sql.sql", "Breadcrumb quests %u and %u are in a loop", qid, breadcrumbForQuestId);
+                exit(1);
```
jackpoz: please just gracefully handle the case and log only a TC_LOG_ERROR, there's not need to shutdown worldserver for just a quest

jackpoz (changes requested):
Looks good, just please replace the 2 exit() calls with some input sanitizing and let worldserver go on

src/server/game/Globals/ObjectMgr.cpp
PolarCookie: No, this one will infinite recursive loop the server later if there's a loop.
I'm of the opinion that all these checks should hard fault and tell the devs to fix their damn tables.

src/server/game/Globals/ObjectMgr.cpp
PolarCookie: This one can be handled gracefully.

src/server/game/Globals/ObjectMgr.cpp
jackpoz: The point is that we are the devs, not the users using TC. If we push a faulty SQL update, we will break all users' TC for no apparent reason as we can just force _breadcrumbForQuestId to 0 when a loop is found.

jackpoz: Let's hope none will add bad data then, but please at least fix the one above about quest not found

PolarCookie: As you wanted; both cases handled "gracefully".
As to TC devs breaking stuff, that happens all the time and it's what the snail pace review process is for.
Strict error warnings are only a benefit and would immediately eliminate those kinds of errors as the only solution is for devs to fix their damn tables (can't accidentally push a faulty sql when the server won't even start), especially as the wiki documentation is woefully out of date on many issues.

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -4991,6 +4994,16 @@ void ObjectMgr::LoadQuests()
                 nextQuestItr->second.DependentPreviousQuests.push_back(qinfo->GetQuestId());
         }
 
+        if (uint32 breadcrumbForQuestId = std::abs(qinfo->_breadcrumbForQuestId))
+        {
+            if (_questTemplates.find(breadcrumbForQuestId) == _questTemplates.end())
+                TC_LOG_ERROR("sql.sql", "Quest %u is a breadcrumb for quest %u, but no such quest exists", qinfo->_id, breadcrumbForQuestId);
```
jackpoz: should we set qinfo->_breadcrumbForQuestId to 0 in this case ? that way we don't have to check it anymore at runtime

PolarCookie: suppose you could, would remove https://github.com/TrinityCore/TrinityCore/blob/ea71f6bebc5c6b3097832b14a0540d7288c314a1/src/server/game/Entities/Player/Player.cpp#L15463

jackpoz (approved):

jackpoz:
Thanks for the PR :)

kelno:
Maybe this should be updated?
https://trinitycore.atlassian.net/wiki/spaces/tc/pages/34897928/quest+template+addon

Killyana:
Wiki updated

PR #23180, opened 2019-04-12
Core/Logging: Split logs for unknown and zero triggered spells

### LINKED ISSUE #23164: Core/Spell: Log spam from Periodic Trigger Spell
**Description:** when an aura with the effect Periodic Trigger Spell is apply to a unit it will spam the server log if there's no spell to be triggered.

example:
https://woehead.way-of-elendil.fr/?spell=40113
https://woehead.way-of-elendil.fr/?spell=66656

1. .go c 85662 (with gm on and stay at the ground)
2. notice the spam
3. .aura 66656 an notice the spam
AuraEffect::HandlePeriodicTriggerSpellAuraTick: Spell 66656 has non-existent spell 0 in EffectTriggered[0] and is therefore not triggered.

AuraEffect::HandlePeriodicTriggerSpellAuraTick: Spell 40113 has non-existent spell 0 in EffectTriggered[0] and is therefore not triggered.

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  rev. c52eb49

ForesterDev:
**Changes proposed:**

-  Move logs for zero triggered spells to debug level

**Target branch(es):** 3.3.5

- [x] 3.3.5

**Issues addressed:** Closes #23164

Build, tested


jackpoz:
Maybe debug is too little?

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -5005,7 +5005,13 @@ void AuraEffect::HandlePreventResurrection(AuraApplication const* aurApp, uint8
 
 void AuraEffect::HandlePeriodicTriggerSpellAuraTick(Unit* target, Unit* caster) const
 {
-    uint32 triggerSpellId = GetSpellInfo()->Effects[GetEffIndex()].TriggerSpell;
+    uint32 triggerSpellId = GetSpellInfo()->Effects[m_effIndex].TriggerSpell;
```
jackpoz: what's the m_effIndex change about ?

Keader: Nothing is just direct access instead call a method that return it.

illfated: Sounds like an improvement to me. 👍

ForesterDev: Changed to match the code in other parts of file

jackpoz:
I would prefer these kind of errors that can be annoying to have their own category so it's easy to filter them out, instead of changing the status to debug because they happen to often. Even better if someone takes care of fixing the actual errors in the first place

Aokromes:
Conflicting files
src/server/game/Spells/Auras/SpellAuraEffects.cpp

illfated:
@ForesterDev : got time to rebase your PR branch?

jackpoz:
I moved the "spell Id is 0" logs to TC_LOG_WARN and kept the "spell Id is invalid" to TC_LOG_ERROR.
There are a few categories but they are shared for both types of errors
- spells.aura.effect.nospell when it's in AuraEffect (SpellAuraEffects.cpp)
- spells.effect.nospell when it's in Spell (SpellEffect.cpp)

jackpoz:
Thanks for the PR :)

PR #23187, opened 2019-04-14
Core/Movement: Implement stack with priority in MotionMaster

### LINKED ISSUE #17981: Core/Unit: Guardians stop follow the player after a combat
**Description**:
A new issue has arisen with charmed guardians, where when the NPC has entered into combat, then after combat it will not move from its spot once combat has finished, On further investigation, it looks like [unit flag 2048 `UNIT_FLAG_PET_IN_COMBAT`](https://trinitycore.atlassian.net/wiki/display/tc/creature_template#creature_template-unit_flags) is not getting cleared when the pet leaves combat.

This has been confirmed with all [Sholazar Guardians](http://www.wowhead.com/npc=28069/sholazar-guardian) and Taunka'le Evacuee ([26159](http://www.wowhead.com/npc=26159/taunkale-evacuee) [26167](http://www.wowhead.com/npc=26167/taunkale-evacuee)) taken [Across Transborea](http://www.wowhead.com/quest=11930/across-transborea) in [Borean Tundra](http://www.wowhead.com/zone=3537/borean-tundra) or [Gods like Shiny Things](http://www.wowhead.com/quest=12572/gods-like-shiny-things) in [Sholazar Basin](http://www.wowhead.com/zone=3711/sholazar-basin), and summon pet (for [Across Transborea](http://www.wowhead.com/quest=11930/across-transborea) you must take quest from [NPC 26158](http://www.wowhead.com/npc=26158/mother-tauranook)) and take pet and let it enter into combat, kill the attacking NPCs and then notice how the NPC just stands there and wont move again.  If you type `.npc info` you will notice NPC still has [unit flags 2048 pet in combat](https://trinitycore.atlassian.net/wiki/display/tc/creature_template#creature_template-unit_flags) even though is not in combat.

Once combat has finished, the guardian should once again follow player and in combat, unit flags should be removed. 

For [Across Transborea](http://www.wowhead.com/quest=11930/across-transborea), you will need the following, as current script no longer works at all:

``` sql
UPDATE `smart_scripts` SET `event_flags`=`event_flags`|0x200 WHERE `entryorguid`IN(2616700,2616701,2616702,2616703) AND `source_type`=9;

UPDATE `smart_scripts` SET `target_type`=12, `target_param1`=1 WHERE  `entryorguid`=2616703 AND `source_type`=9 AND `id`=6 AND `link`=0;

DELETE FROM `smart_scripts` WHERE `entryorguid`=26167 AND `source_type`=0;
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES 
(26167, 0, 0, 0, 25, 0, 100, 512, 0, 0, 0, 0, 11, 46669, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 'Taunka\'le Evacuee - On Spawn - Cast Evacuee Random Action'),
(26167, 0, 1, 0, 8, 0, 100, 512, 46663, 0, 0, 0, 87, 2616700, 2616701, 2616702, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 'Taunka\'le Evacuee - On Spellhit (Evacuee Random Action) - Run Random Script'),
(26167, 0, 2, 0, 8, 0, 100, 513, 46677, 0, 0, 0, 80, 2616703, 2, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 'Taunka\'le Evacuee - On Spellhit (Evacuee Reaches Dragonblight) - Run Script'),
(26167, 0, 3, 0, 1, 0, 100, 512, 0, 0, 0, 0, 29, 0, 0, 0, 0, 0, 0, 12, 1, 0, 0, 0, 0, 0, 0, 'Taunka\'le Evacuee - On Reset - Follow Target'),
(26167, 0, 4, 0, 54, 0, 100, 512, 0, 0, 0, 0, 64, 1, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, 'Taunka\'le Evacuee - On Just Summoned Store Target');
```

With the above SQL, once NPC enters into combat, it will not move from its spot.

now add this line

``` sql
(26167, 0, 5, 0, 0, 0, 100, 512, 30000, 30000, 30000, 30000, 19, 2048, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 'Taunka\'le Evacuee - IC - Remove unit flag Pet In Combat');
```

Now you will notice the guardian follows all the time but combat fails (obviously)

**Branch(es)**: Tested on 3.3.5, can't confirm on 6.x.x 

**TC hash/commit**:  
https://github.com/TrinityCore/TrinityCore/commit/8d2f17b2a8b20bbb2728c5d0a8fc9eed4e398205

jackpoz:
Handle movement generators as stack with priority, deactivating current one when adding a new movement generator instead of deleting current one.
Fixes #17981

**Changes proposed:**

-  Handle movement generators as stack with priority, deactivating current one when adding a new movement generator instead of deleting current one.
-  ChaseMovementGenerator now doesn't wait for the target to be removed from world to be deactivated anymore but does that as soon as the target dies.

- 3.3.5

**Issues addressed:**
Closes #17981

- .cast 46657
- attack a mob
- after the mob dies, notice how the guardian now follows you again

- As this is a quite big change, it requires testing is other cases. Some thoughts:
  - How big is _generators going to grow if every time a npc attacks a creature pushes a new ChaseMovementGenerator ?
  - Should all the calls of MotionMaster::Add() be modified not to create a new generator everytime but maybe use an existing one already in _generators ?
- SmartAI::AttackStart() calls me->PauseMovement() that in case of an active generator does nothing to it. Should it deactivate it ? Or is mCanCombatMove always true anyway for creatures that have a movement generator other than the idle one anyway ?

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -594,7 +594,7 @@ void SmartAI::AttackStart(Unit* who)
 
     if (who && me->Attack(who, mCanAutoAttack))
     {
-        me->GetMotionMaster()->Clear(MOTION_PRIORITY_NORMAL);
+        //me->GetMotionMaster()->Clear(MOTION_PRIORITY_NORMAL);
         me->PauseMovement();
 
         if (mCanCombatMove)
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -1027,7 +1027,7 @@ void MotionMaster::LaunchMoveSpline(Movement::MoveSplineInit&& init, uint32 id/*
 void MotionMaster::Pop(bool active, bool movementInform)
 {
     MovementGenerator* pointer = *_generators.begin();
-    _generators.erase(pointer);
+    _generators.erase(_generators.begin());
     Delete(pointer, active, movementInform);
 }
 
@@ -1121,18 +1121,15 @@ void MotionMaster::DirectAdd(MovementGenerator* movement, MovementSlot slot/* = 
                 AddFlag(MOTIONMASTER_FLAG_STATIC_INITIALIZATION_PENDING);
             break;
         case MOTION_SLOT_ACTIVE:
+        {
+            auto insertPosition = _generators.begin();
+
             if (!_generators.empty())
             {
                 if (movement->Priority >= (*_generators.begin())->Priority)
                 {
                     MovementGenerator* pointer = *_generators.begin();
-                    if (movement->Priority == pointer->Priority)
-                    {
-                        _generators.erase(pointer);
-                        Delete(pointer, true, false);
-                    }
-                    else
-                        pointer->Deactivate(_owner);
+                    pointer->Deactivate(_owner);
                 }
                 else
                 {
@@ -1141,20 +1138,16 @@ void MotionMaster::DirectAdd(MovementGenerator* movement, MovementSlot slot/* = 
                         return a->Priority == movement->Priority;
                     });
 
-                    if (itr != _generators.end())
-                    {
-                        MovementGenerator* pointer = *itr;
-                        _generators.erase(pointer);
-                        Delete(pointer, false, false);
-                    }
+                    insertPosition = itr;
                 }
             }
             else
                 _defaultGenerator->Deactivate(_owner);
 
-            _generators.emplace(movement);
+            _generators.emplace_hint(insertPosition, movement);
             AddBaseUnitState(movement);
             break;
+        }
         default:
             break;
     }
--- a/src/server/game/Movement/MovementGenerators/ChaseMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/ChaseMovementGenerator.cpp
@@ -90,7 +90,7 @@ bool ChaseMovementGenerator::Update(Unit* owner, uint32 diff)
 
     // our target might have gone away
     Unit* const target = GetTarget();
-    if (!target || !target->IsInWorld())
+    if (!target || !target->IsInWorld() || !target->IsAlive())
         return false;
 
     // the owner might be unable to move (rooted or casting), pause movement


jackpoz:
@ccrs since you worked with Movement Generators, could you fine tune this PR with whatever comes to mind ?

src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -594,7 +594,7 @@ void SmartAI::AttackStart(Unit* who)
 
     if (who && me->Attack(who, mCanAutoAttack))
     {
-        me->GetMotionMaster()->Clear(MOTION_PRIORITY_NORMAL);
+        //me->GetMotionMaster()->Clear(MOTION_PRIORITY_NORMAL);
```
jackpoz: should we do something about this in case mCanCombatMove is false ? maybe that's the only case when to clear the motion generator ?

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -1027,7 +1027,7 @@ void MotionMaster::LaunchMoveSpline(Movement::MoveSplineInit&& init, uint32 id/*
 void MotionMaster::Pop(bool active, bool movementInform)
 {
     MovementGenerator* pointer = *_generators.begin();
-    _generators.erase(pointer);
+    _generators.erase(_generators.begin());
```
jackpoz: in case of Chase and Follow, this was deleting both

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -1141,20 +1138,16 @@ void MotionMaster::DirectAdd(MovementGenerator* movement, MovementSlot slot/* =
...
-                        Delete(pointer, false, false);
-                    }
+                    insertPosition = itr;
                 }
             }
             else
                 _defaultGenerator->Deactivate(_owner);
 
-            _generators.emplace(movement);
+            _generators.emplace_hint(insertPosition, movement);
```
jackpoz: this will make sure the new generator is added in front of all others with the same priority

src/server/game/Movement/MovementGenerators/ChaseMovementGenerator.cpp
```diff
@@ -90,7 +90,7 @@ bool ChaseMovementGenerator::Update(Unit* owner, uint32 diff)
 
     // our target might have gone away
     Unit* const target = GetTarget();
-    if (!target || !target->IsInWorld())
+    if (!target || !target->IsInWorld() || !target->IsAlive())
```
jackpoz: I really struggle to understand why the IsAlive() check wasn't here before

src/server/game/Movement/MotionMaster.cpp
Shauren: .erase by value will use comparator to determine what to remove (see custom comparator for `_generators`)

jackpoz: yeah, the current comparator only checks for type (default, active and passive) and priority, that didn't sound right for this case

Shauren: Are there more suspicious erases like that in code?

ccrs: Umm there are several actually, didnt know that erase by value will use the custom comparator.
This is actually a important issue... IIRC almost all erases rely on "I wan to delete this exact pointer".

What solutions are suitable for this?
- Create a unique identifier for each movement (modifying the container declaration) and erase using that?
- ... ?

src/server/game/Movement/MovementGenerators/ChaseMovementGenerator.cpp
ccrs: ummm good one, I cant remember exactly, but there is actually a reason

all I can remember right now is:
- unit's death is handled before movementgenerator update
- it is _guaranteed_ that movements are cleared on unit's death
- checking for alive shouldn't be needed...

Also, what if a dev needs a death creature chasing someone? (the odds are same as the earth being actually flat, but who knows right?)

src/server/game/Movement/MotionMaster.cpp
Shauren: erase by iterator

ccrs: I'll perform and commit the changes then.

But can you enlighten me about this, please?
* What happens in concurrency scenarios? Example: there are several threads updating the same map context, each thread updates different _parts_, example: map instance script and map entity visitor.
  * Can that actually happen?
  * Can each thread perform actions on the same object at the same time?
  * Is this actually possible? --> Can one thread perform one motionmaster delete at the same time as another thread?
    * This is actually my main worry about iterator erase
    * Is it a mutex or locking variable or deletion queue or any other countermeasure even necessary?

Shauren: All of these questions can be answered with 1 word:
No.

Map is strictly bound to a single thread, there can be no issues like what you described

ccrs: Neat! Ty! 😃

jackpoz: TC schedules a map update to the map updater thread pool, waits for all maps to finish the updates, then does some single threaded stuff and restarts. Some events like Object deletion are delayed to avoid race conditions

src/server/game/Movement/MovementGenerators/ChaseMovementGenerator.cpp
jackpoz: The check is for the target, with the old code creatures with ChaseMovementGenerator would stare at the dead body of the target until it was removed from world. Now they go back at doing what they are supposed to do

ccrs: umm then I think the problem is within chase movement comprobation/creation:

- My target dies, combat manager does it's thing.
  - Has my target changed? Yes? Am I chasing that new target? Start chasing
  - Am I now in combat? No? Stop chasing

And more use cases that I dont have the time to fully detail 😢

jackpoz: But those checks don't work because when target dies the creature is not engaged anymore and that skips the checks that are supposed to trigger evade or whatever. Just debug it and you'll see how broken the current implementation is

ccrs: I see what you mean... damn

jackpoz:
Replaced by https://github.com/TrinityCore/TrinityCore/pull/23466

PR #23207, opened 2019-04-25
Core/GameObject: Fix the logic used to set the flag GO_FLAG_NODESPAWN…

### LINKED ISSUE #15730: Core/Gameobject: Gameobjects despawn once clicked
Hello,

server: 8ff619ddc40fee1739125e91d20b519078c3f2fb
db all updates

The apple vat are to disappear when i loot the apple in the Helloween days for ca 3 min.

Killyana:
… for gobs having !GetDespawnPossibility OR !IsDespawnAtAction

Keep Goobers that are summoned, spawned until the timer expire, like Romantic Basket case

**Changes proposed:**

-  Fix the logic used to set the flag GO_FLAG_NODESPAWN for gobs having !GetDespawnPossibility OR !IsDespawnAtAction
-  Keep Goobers that are summoned spawned until the timer expire, like Romantic Basket case

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #15730

**Tests performed:** Yes

- [ ]  Require reviews

### DIFF
--- a/src/server/game/Entities/GameObject/GameObjectData.h
+++ b/src/server/game/Entities/GameObject/GameObjectData.h
@@ -487,7 +487,8 @@ struct GameObjectTemplate
             case GAMEOBJECT_TYPE_GOOBER:     return goober.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGSTAND:  return flagstand.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGDROP:   return flagdrop.noDamageImmune != 0;
-            default: return true;
+            case GAMEOBJECT_TYPE_CHEST:      return true;
+            default: return false;
         }
     }
 
--- a/src/server/game/Entities/GameObject/GameObjectData.h
+++ b/src/server/game/Entities/GameObject/GameObjectData.h
@@ -477,7 +477,7 @@ struct GameObjectTemplate
         }
     }
 
-    bool GetDespawnPossibility() const                      // despawn at targeting of cast?
+    bool CannotBeUsedUnderImmunity() const                      // The object cannot be use under Immunity like divine shield -- to be implemented
     {
         switch (type)
         {
@@ -487,7 +487,7 @@ struct GameObjectTemplate
             case GAMEOBJECT_TYPE_GOOBER:     return goober.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGSTAND:  return flagstand.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGDROP:   return flagdrop.noDamageImmune != 0;
-            default: return true;
+            default: return false;
         }
     }
 
--- a/src/server/game/Entities/GameObject/GameObjectData.h
+++ b/src/server/game/Entities/GameObject/GameObjectData.h
@@ -487,7 +487,7 @@ struct GameObjectTemplate
             case GAMEOBJECT_TYPE_GOOBER:     return goober.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGSTAND:  return flagstand.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGDROP:   return flagdrop.noDamageImmune != 0;
-            default: return false;
+            default: return true;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/GameObject/GameObjectData.h
```diff
@@ -487,7 +487,7 @@ struct GameObjectTemplate
             case GAMEOBJECT_TYPE_GOOBER:     return goober.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGSTAND:  return flagstand.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGDROP:   return flagdrop.noDamageImmune != 0;
-            default: return true;
+            default: return false;
```
jackpoz: what does this change ?

Killyana: Just the default value, CannotBeUsedUnderImmunity is not used anywhere, it could be renamed and changed after.

Killyana:
Using this to make the gob not despawning once used looks wrong:
SetFlag(GAMEOBJECT_FLAGS, GO_FLAG_NODESPAWN);
m_respawnDelayTime = 0;
m_respawnTime = 0;

In sniffs the flag GO_FLAG_NODESPAWN is not set to this gobs.

And m_respawnDelayTime = 0; + m_respawnTime = 0 is literally respawning the gob instantly once despawned, we can remove SetFlag(GAMEOBJECT_FLAGS, GO_FLAG_NODESPAWN); and the result will be the same.

The problem when the gob respawns instantly it will acquire flags that was removed by scripts like : https://github.com/TrinityCore/TrinityCore/commit/aff5229cd1777581bd85079d064037a57bf17799#diff-81ee15a2fe57681a2874d268555b8265R223

We need just something that disallow despawn and reset the state of the gob to make it usable again once clicked.

Killyana:
@PolarCookie Can you help with this issue?
with this check summoned chests an goobers depawns correctly when the summon time expires
if (m_respawnTime > 0 && GetGoType() == (GAMEOBJECT_TYPE_CHEST || GAMEOBJECT_TYPE_GOOBER) && !GetGOInfo()->IsDespawnAtAction()

But If I remove : m_respawnTime > 0   they remains visible with GM ON

To Test this issue .go c 75022  and .die then notice the summoned gob behavior

illfated:
> - Keep Goobers that are summoned spawned until the timer expire, like Romantic Basket case

Maybe I have misunderstood that, but will it solve the despawn issue of https://www.wowhead.com/object=164690/kromzars-banner ?

Killyana:
There's no issue with kromzars-banner.

illfated:
Sorry, I forgot. The solution was supposedly in the PR #22124 and I forgot about it.

PolarCookie:
I can try taking a look at it, but not promising anything; been meaning to push a new PR but been so unproductive lately.

src/server/game/Entities/GameObject/GameObjectData.h
xvwyh: Careful! For `GAMEOBJECT_TYPE_CHEST` this function should always return `true` **unless** "DamageImmuneOK" property is set, which was only created in later expansions:
https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/GameObject/GameObjectData.h#L112
And thus for 3.3.5 chests can be considered to always return `true`.

Killyana: Both DamageImmuneOK and noDamageImmune  are not implemented I just renamed the bool.

xvwyh: No, you also inverted the (unimplemented) result of the method for all types apart from the ones used in switch. Whether or not it wasn't used before, for chests it was returning the correct value, and now it isn't, so it would be great if you fixed this unintended side-effect. Implementing it is just a matter of time, it isn't difficult, but if you let this mistake remain the code, it might confuse future implementers.

Killyana: Reverted

xvwyh: Why revert it instead of just adding `case GAMEOBJECT_TYPE_CHEST: return true;`...

Killyana: It must be fine with the last change

xvwyh: Wonderful, thank you!

Issue #23205, opened 2019-04-25
LOCK_KEY_SPELL implementation


xvwyh:
As of 3.3.5, Lock.dbc contains 4 different types of locks: 3 already implemented, one - not. In the case of this not-yet-implemented lock type, LockEntry::Index field contains the ID of a spell that needs to be cast by the player to facilitate opening of the lock, be it for special conditions or appropriate visuals, or just a more fancy name on the cast bar. Hence I called it `LOCK_KEY_SPELL`, but a name like `LOCK_KEY_CUSTOM_SPELL` or `LOCK_KEY_SPELLCAST` would also be alright.

There are only 3 locks that use this type in 3.3.5:
- 1810
  - not used?
  - opened by a spell with a wonderful ID of 10000 - Douse Lock

- 1819
  - used by GO 190752, 195235, 195331 - battleground Seaforium Bombs
  - opened by a spell 1843 - Disarm

- 1845
  - used by GO 194264 - Lever in Ulduar's Thorim encounter
  - opened by a spell 68398 - Opening - for extra long 10 secs cast time!

Unlike `LOCK_KEY_SKILL`, which requires the player to have learned an appropriate spell of an appropriate skill level to open the lock (be it for lockpicking's sake, or for profession-specific mechanics like herb gathering, mining etc), `LOCK_KEY_SPELL` does not require the player to know the spell - but the client will attempt to cast the spell regardless. This is a problem, because currently `CMSG_CAST_SPELL` handler would refuse such attempts, since the player doesn't "know" the spell they attempt to cast, and thus opening such a gameobject becomes impossible.

**Implementation:**
SharedDefines.h
```diff
 enum LockKeyType
 {
     LOCK_KEY_NONE  = 0,
     LOCK_KEY_ITEM  = 1,
-    LOCK_KEY_SKILL = 2
+    LOCK_KEY_SKILL = 2,
+    LOCK_KEY_SPELL = 3,
 };
```
(why oh why do you omit the trailing comma in your codestyle, such diffs would be one-liners otherwise)

GameObject::GetSpellForLock
This was just recently ported in #23179, but the author, I assume, thought this was a typo when the code wouldn't build, and replaced it with what they could find instead. Let's restore it to the way it was meant to be.
```diff
-        if (lock->Type[i] == LOCK_KEY_SKILL)
+        if (lock->Type[i] == LOCK_KEY_SPELL)
[...]

PR #23179, opened 2019-04-12
Core/Object: Range check

d0d0:
**Changes proposed:**

-  better range checking on gobs targets

**Target branch(es):** 3.3.5/master

- [x] 3.3.5

**Issues addressed:**

Closes #23062

**Tests performed:** (Does it build, tested in-game, etc.)

Tested in game on known test cases (in issue #23062)

Issue #23062, opened 2019-02-22
CheckRange for gameobject targets


robinsch:
**Description:**

CheckRange needs to include m_targets.GetGOTarget() checks. Allows tons of exploits.

**Current behaviour:**

Doesn't

**Steps to reproduce the problem:**

Use a Key at a big distance on a locked door.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

951f130f20ef6e023c5d5c220eefb0ed9544ddef

Suggested:

```diff
diff --git a/src/server/game/Spells/Spell.cpp b/src/server/game/Spells/Spell.cpp
index 63af3ae2e..1624da1bf 100755
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -6194,6 +6194,16 @@ SpellCastResult Spell::CheckRange(bool strict) const
             (m_spellInfo->FacingCasterFlags & SPELL_FACING_FLAG_INFRONT) && !m_caster->HasInArc(static_cast<float>(M_PI), target))
             return SPELL_FAILED_UNIT_NOT_INFRONT;
     }
+    
+    if (GameObject* go = m_targets.GetGOTarget())
+    {
+        if (m_caster->GetExactDistSq(go) > maxRange)
+            return SPELL_FAILED_OUT_OF_RANGE;
+
+        if (minRange > 0.0f && m_caster->GetExactDistSq(go) < minRange)
+            return SPELL_FAILED_OUT_OF_RANGE;
+    }
 
     if (m_targets.HasDst() && !m_targets.HasTraj())
     {
```

sanctum32:
Thank you for this! Tested path and works perfectly

jackpoz:
Could you open a PR ?

jackpoz:
Please take a look at https://github.com/TrinityCore/TrinityCore/commit/76cc94f325ab68f66a4fa7b5d30316508e27c0ce#r32450315

jackpoz:
anyone got time to test https://github.com/TrinityCore/TrinityCore/commit/270a144083d85d18cba8aff861dd6effbd04c476 ?

Killyana:
Still not accurate:
.add 27991
.go xyz -3559.5385, 4951.6186, -101.04364 530    -- you will get out of range
.go xyz -3572.1037, 4942.8466, -101.04618 530    -- use the key in the door and it will open 

[...]

Issue #22530, opened 2018-09-26
DB/Quest: Learning to Leave and Return: the Magical Way


TheWinchesters:
**Description:**
In the quest Learning to Leave and Return: the Magical Way (https://www.wowhead.com/quest=12790) you have to use the Teleport to Violet Stand Crystal and the Teleport to Dalaran Crystal.

**Current behaviour:**
Teleport to Violet Stand Crystal (https://www.wowhead.com/object=191229) does not teleport the character.
And since you need to use that one to be able to use Teleport to Dalaran Crystal (https://www.wowhead.com/object=191230) I do not know if this one works.

**Expected behaviour:**
You should be teleported after using the crystals.

**Steps to reproduce the problem:**
1. .q a 12790
2. .go c id 29156
3. Try using the crystal

Branch: 3.3.5
TC Rev.: 7baf5b24525551ddf00c88084edb2a1bafeb758d
OS: Windows 8.1 / Debian 9

nagrandbugtracker:
It does work but you literally need to stand below the crystals.

illfated:
That is interesting. Maybe this has got something to do with
GameObject usage radius, similar to the issue discussed in https://github.com/TrinityCore/TrinityCore/issues/15453 ?
([**Core/GameObject: Large GOs and gossip availability**](https://github.com/TrinityCore/TrinityCore/issues/15453))

kaiyazeera:
I just ran into this issue, too. The quest object was marked completed, but I wasn't teleported. It did work however to stand directly below the crystals. Kinda irritating, would probably not have noticed if it wasn't for @Ontari784ever mentioning it (thanks!)

PR #23226, opened 2019-04-30
Core/Object: Range check vol. 2

d0d0:
**Changes proposed:**

-  implement reviews from https://github.com/TrinityCore/TrinityCore/pull/23179

- [x] 3.3.5

https://github.com/TrinityCore/TrinityCore/issues/15453 
https://github.com/TrinityCore/TrinityCore/issues/23062#issuecomment-486745022
https://github.com/TrinityCore/TrinityCore/issues/22530
https://github.com/TrinityCore/TrinityCore/issues/23205

**Tests performed:** 

Ingame with all cases from original issue 
https://github.com/TrinityCore/TrinityCore/issues/23062
https://github.com/TrinityCore/TrinityCore/issues/15453
(quests like https://github.com/TrinityCore/TrinityCore/issues/17872  https://github.com/TrinityCore/TrinityCore/issues/22530 Polishing the Helm)


src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -2554,14 +2554,34 @@ float GameObject::GetInteractionDistance() const
...
+        case GAMEOBJECT_TYPE_MINI_GAME:
+            return 5.5555553f;
+        case GAMEOBJECT_TYPE_BINDER:
             return 10.0f;
+        case GAMEOBJECT_TYPE_CHAIR:
+        case GAMEOBJECT_TYPE_BARBER_CHAIR:
+            return 3.0f;
         case GAMEOBJECT_TYPE_FISHINGHOLE:
         case GAMEOBJECT_TYPE_FISHINGNODE:
             return 20.0f + CONTACT_DISTANCE; // max spell range
```
xvwyh: I took all the values from the client executable, so you might as well have left 100 yards there.
image
image
`GAMEOBJECT_TYPE_FISHINGNODE` is initialized to 100 yards, but `GAMEOBJECT_TYPE_FISHINGHOLE` reuses the same constructor as `GAMEOBJECT_TYPE_SPELL_FOCUS` (which is also reused by `GAMEOBJECT_TYPE_DUEL_ARBITER` and `GAMEOBJECT_TYPE_AURA_GENERATOR`, hey, compiler can optimize things however it wants) which just calls the base constructor and initializes interaction distance to default 5 yards. Which is because player never really interacts with a fishing hole, only with a fishing node, which on the server checks if there are any holes nearby and sends their loot instead of the fishing loot for the current zone.

But I won't insist, since realistically the only time this will be relevant is if a player decides to fish from a moving transport, like a ship or a turtle taxi in Northrend. 😄

Killyana:
I tested all know cases, and everything looks fine.

Faq:
so only range correction for GAMEOBJECT_TYPE_FISHINGNODE missing

jackpoz:

xvwyh:
Thank you, github behaved this time and didn't unsub me randomly.
I don't have any other problems with this PR apart from codestyle, but I'll leave that critique to someone better versed in TC's codestyle than me. Functionality-wise, it can have my worthless imaginary seal of approval.

Shauren:
Codestyle looks fine to me

xvwyh:
I meant the things like comma at the end of enum values (which seems to be diligently expunged in TC, which is something I personally strongly disagree on); `case GAMEOBJECT_TYPE_FISHINGHOLE` not being dropped down to "non blizzlike values"; redefinition of the meaning of `IsWithinDistInMap` without actually having said "dist" as a parameter thus making the function name an odd choice at first glance (and IMO it's a better idea to put IsInMap, IsInPhase checks into IsAtInteractDistance, which is actually something I should do in my server too now that I think about it, brb); also I wish I had came up with a better name for IsAtInteractDistance, but WorldObject having all these similar methods like IsInRange, IsInRange2d, IsInRange3d, IsWithinDistInMap and so on already makes it confusing enough; not to mention that GameObject still has the now obsolete method IsInRange (which overloads WorldObject::IsInRange with a completely different implementation), which probably should've been replaced with IsAtInteractDistance's contents instead of adding it alongside one another, yadda yadda yadda.

But none of this detracts from the PR's intent and all this can be fixed later provided someone will care enough.

jackpoz:
tl;dr I merged it :)
Thanks for the PR :)

PR #22332, opened 2018-08-28
Core/Scripts: Move some spells to scripts

DanVS:
**Changes proposed:**
Move generic school damage spells (from `@todo`) to scripts. Also applied to four horsemen script.

**Target branch(es):** 3.3.5

**Tests performed:** Does build and partial tests shown that works in-game.

PR #23228, opened 2019-05-01
Scripts/Spells: Fix implementation of #22332

### LINKED ISSUE #21435: Scripts/Spells: Ebon Gargoyle - Attack Power too low
**Description:**

Ebon Gargoyle hitting too low, should hit around 2K ( and its damage currently does not scale with the owner attackpower )

screen shot 1396-11-28 at 2 10 12 am

**Branch(es):** 3.3.5a

**TC rev. hash/commit:** d3d0640

**TDB version:**  TDB 335.63

**Operating system:** Debian9

jackpoz:
Change the way #22332 moved spell damage calculations from EffectSchoolDMG to scripts, using OnEffectLaunchTarget and SetEffectValue instead

**Changes proposed:**

-  use SetEffectValue() instead of SetHitDamage() since EffectSchoolDMG modifies "damage", not "m_damage"
-  use OnEffectLaunchTarget instead of OnEffectHitTarget, so the script is called right before SpellEffectDMG and all the bonus code in EffectSchoolDMG is called with the "damage" set in the script
-  this restores 100% the code flow of the old code before #22332 

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #21435

I debugged spell_pet_dk_gargoyle_strike and checked that the code flow worked like intended. I don't know how much damage it's supposed to do, I got around 600 damage with a brand new level 80 character.
I didn't test the other spells, just applied the same changes.

None

### DIFF
--- a/src/server/scripts/Kalimdor/RuinsOfAhnQiraj/boss_rajaxx.cpp
+++ b/src/server/scripts/Kalimdor/RuinsOfAhnQiraj/boss_rajaxx.cpp
@@ -145,12 +145,12 @@ class spell_rajaxx_thundercrash : public SpellScript
         if (damage < 200)
             damage = 200;
 
-        SetHitDamage(damage);
+        SetEffectValue(damage);
     }
 
     void Register() override
     {
-        OnEffectHitTarget += SpellEffectFn(spell_rajaxx_thundercrash::HandleDamageCalc, EFFECT_0, SPELL_EFFECT_SCHOOL_DAMAGE);
+        OnEffectLaunchTarget += SpellEffectFn(spell_rajaxx_thundercrash::HandleDamageCalc, EFFECT_0, SPELL_EFFECT_SCHOOL_DAMAGE);
     }
 };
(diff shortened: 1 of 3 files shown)


ccrs:
seems good to me
just asking, have you considered using beforehit hook?

jackpoz:
Doesn't https://github.com/TrinityCore/TrinityCore/blob/24fbbee4b9af7b5226772378dd83b78c103d969d/src/server/game/Spells/SpellEffects.cpp#L317 mean the code after will be executed only if current mode is SPELL_EFFECT_HANDLE_LAUNCH_TARGET (so OnEffectLaunchTarget is the latest hook called before) ?

Shauren:
BeforeHit is a bad idea, you run into the same issue where you fully take over damage calculation (meaning no bonuses get applied if you dont do that yourself) OnEffectLaunchTarget + SetEffectValue is the only way to keep the bonuses handled in EffectSchoolDMG

PR #23233, opened 2019-05-02
Core/Vehicles: Fix multiple vehicle join events on same seat

jackpoz:
**Changes proposed:**

-  Apply https://gist.github.com/joschiwald/36da1450dcea394a9ca3 from https://github.com/TrinityCore/TrinityCore/pull/21220#issuecomment-357062554

- [x] 3.3.5
- [ ] master

**Issues addressed:**
It fixes 1 of the 2 issues that https://github.com/TrinityCore/TrinityCore/pull/21220 fixes (the 2nd issue will be fixed later on)

**Tests performed:** None

- [ ] Someone should test it

### DIFF
--- a/src/server/game/Entities/Vehicle/Vehicle.cpp
+++ b/src/server/game/Entities/Vehicle/Vehicle.cpp
@@ -324,7 +324,7 @@ SeatMap::const_iterator Vehicle::GetNextEmptySeat(int8 seatId, bool next) const
     if (seat == Seats.end())
         return seat;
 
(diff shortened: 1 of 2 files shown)


Keader:
Some feedback here: https://github.com/TrinityCore/TrinityCore/pull/21220#issuecomment-466262231

jackpoz:
>issue can be reproduced by going to Ulduar 25man somewhere, logging 2nd account and going to the same place, then .npc add temp 32934, targetting right arm and .cast self 63981. The arm looks like it's grabbing both players, but in the end only grabs one.

This should be the issue fixed by this PR

jackpoz:
Well it works ingame :)

PR #23234, opened 2019-05-03
Fix player turn at the start of the game for the race of Orcs

IIushka:
Fix player rotation at the start of the game for the race of Orcs


Aokromes:
thx but we don't use Pull request for sql only fixes, open a ticket instead.

Issue #23227, opened 2019-05-01
[Core/Spells] Make SpellCast methods return CastResult


Keader:
**Description:** 

We need a simple way to know if a SpellCast fail or not.
It's need to reschedule events (creature scrits) when cast fail.
In retail, normally if a spell fail, npc try cast again a sec later.
In TC, cast is skipped and will repeat in normal time instead a sec later.
Maybe needed to master too.

**TC rev. hash/commit:**  69c15f5ae769612c89346c6a308655d0987caae7

ccrs:
dunnow if this is possible and reliable actually

Keader:
I remember to see somewhere in sunwell, something called: CastSpellIfPossible(target, spellid, triggerFlags); that return bool i think
Or maybe a new hook?!

ccrs:
spellhit?

Keader:
Still seems weird a new hook only for handle with it :/

illfated:
Thank you to @brotalnia for solving this issue so quickly. 👍

PR #23236, opened 2019-05-03
Core/Spells: Return CheckCast result from CastSpell

brotalnia:
**Changes proposed:**

-  Return result of initial cast check from WorldObject::CastSpell methods

- [x] 3.3.5
- [ ] master

Fixes https://github.com/TrinityCore/TrinityCore/issues/23227

**Tests performed:** 

It compiles.

### DIFF
--- a/src/server/game/AI/CoreAI/UnitAI.cpp
+++ b/src/server/game/AI/CoreAI/UnitAI.cpp
@@ -125,7 +125,7 @@ float UnitAI::DoGetSpellMaxRange(uint32 spellId, bool positive)
     return spellInfo ? spellInfo->GetMaxRange(positive) : 0;
 }
 
-void UnitAI::DoCast(uint32 spellId)
+SpellCastResult UnitAI::DoCast(uint32 spellId)
 {
     Unit* target = nullptr;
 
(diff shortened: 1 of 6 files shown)


jackpoz:
What kind of spells does this handle ? Just instant or the ones with cast time too ?

brotalnia:
It returns the *initial* cast check for all spells. That tells you if the requirements are met to *begin* casting the spell. These are checks like "Is the target in range", "Do you have enough mana", "Are you silenced", etc.

Thus for instant spells, you would also know the final result of the cast, but for spells with a cast time, you won't know if the mob was interrupted before the cast finished for example. Only that it was able to *begin* casting the spell.

If you want to know when a spell cast actually completes, there already is an AI event for that:
https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/AI/CreatureAI.h#L164

jackpoz:
That clears up a lot any doubts, thanks :)

illfated:
Nice clarification of the code, as well as more useful output. 👍

jackpoz:
@Keader does this look good for you ?

Keader:
Yes i think so.
But it's needed in AI methods too (UnitAI.h)
Like:

```
        void DoCast(uint32 spellId);
        void DoCast(Unit* victim, uint32 spellId, CastSpellExtraArgs const& args = {});
        void DoCastSelf(uint32 spellId, CastSpellExtraArgs const& args = {}) { DoCast(me, spellId, args); }
        void DoCastVictim(uint32 spellId, CastSpellExtraArgs const& args = {});
        void DoCastAOE(uint32 spellId, CastSpellExtraArgs const& args = {}) { DoCast(nullptr, spellId, args); }
```

brotalnia:
Updated the AI methods to also return the cast result.

Keader:
@Shauren you can see any problem here ?

jackpoz:
Thanks for the PR :)

Keader:
Ty

PR #23237, opened 2019-05-03
Scripts/Spells: Gordunni Trap

### LINKED ISSUE #23232: Spell/Script: spell_gordunni_trap
**Description:**  For this spell script: spell_gordunni_trap
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Kalimdor/zone_feralas.cpp#L149

The 2 gamobjects must be summoned by spells:
https://woehead.way-of-elendil.fr/?spell=11756
https://woehead.way-of-elendil.fr/?spell=19394

Quest: https://woehead.way-of-elendil.fr/?quest=2987

**Branch(es):** 3.3.5

**TC rev. hash/commit:** rev. 69c15f5ae769

kuskmen:
**Changes proposed:**

- Gameobjects are now spawned by spell rather than manually.  

**Target branch(es):** 3.3.5

- [x] 3.3.5

**Issues addressed:** Closes #23232 

**Known issues and TODO list:** Not that I am aware of.

### DIFF
--- a/src/server/scripts/Kalimdor/zone_feralas.cpp
+++ b/src/server/scripts/Kalimdor/zone_feralas.cpp
@@ -149,10 +149,10 @@ public:
 ## spell_gordunni_trap
 ######*/
 
-enum GordunniTrap
+enum GordunniTrapSpells
 {
-    GO_GORDUNNI_DIRT_MOUND_1 = 144064,
-    GO_GORDUNNI_DIRT_MOUND_2 = 177681
+    SPELL_GORDUNNI_DIRT_MOUND_CHEST = 11756,
+    SPELL_GORDUNNI_DIRT_MOUND_JUNK = 19394
 };
 
 class spell_gordunni_trap : public SpellScriptLoader
@@ -167,8 +167,7 @@ class spell_gordunni_trap : public SpellScriptLoader
             void HandleDummy()
             {
                 GameObject* caster = GetGObjCaster();
-                if (GameObject* chest = caster->SummonGameObject(urand(0, 1) ? GO_GORDUNNI_DIRT_MOUND_1 : GO_GORDUNNI_DIRT_MOUND_2, *caster, QuaternionData(), 0))
-                    chest->SetSpellId(GetSpellInfo()->Id);
+                caster->CastSpell(caster, urand(0, 1) ? SPELL_GORDUNNI_DIRT_MOUND_CHEST : SPELL_GORDUNNI_DIRT_MOUND_JUNK);
             }
 
             void Register() override


Killyana:
There's an issue with this Spells, they have target self with Radius: 3 yards that make the summoned gob not spawning in the position of the caster (sometimes undergound)

As this spells are from spell_dbc, if there's no way fix it in the spell script
it could be fixed like this:
``` sql
UPDATE `spell_dbc` SET `RangeIndex`=1, `EffectImplicitTargetA1`=18, `EffectImplicitTargetA2`=18, `EffectRadiusIndex1`=0, `EffectRadiusIndex2`=0 WHERE `id` IN (11756);
UPDATE `spell_dbc` SET `RangeIndex`=1, `EffectImplicitTargetA1`=18, `EffectRadiusIndex1`=0, `EffectRadiusIndex2`=0 WHERE `id` IN (19394);

illfated:
Cosmetic detail, PR title suggestion:
either **Scripts/Feralas: Gordunni Trap spells** -- or even possibly **Scripts/Spells: Gordunni Trap**

xvwyh:
The range on the effect shouldn't be changed. Lootable GO should be spawned not at the trap's location or underneath a player, but in front of the player: https://youtu.be/ljMb__6xPnw?t=68

The current script for dummy effect however is causing the GO to cast the spell, when it should instead be causing the player to cast the spell. It seems reasonable that when you dig a hole, you'll dig it next to yourself, and not somewhere up to 13 yards away, as it currently happens (the theoretical maximum: 10 yards range of the disturb trap spell 11757 + 3 yards offset added by the summon effect of 11756/19394), the video above confirms that Mound appeared in front of the player. May be a coincidence, of course, but good luck finding another example.

Full order of events as I see it:
- Player casts "Digging for Cobalt"
- If there are no GO "Gordunni Trap" spawned around - nothing happens, otherwise...
- "Digging for Cobalt" hits GO "Gordunni Trap", causing the trap to get "disturbed" and cast its spell
- GO "Gordunni Trap" casts "Gordunni Trap" back at player (as per "spell" property) and despawns (because "charges" property being equal to 1) to be respawned later when the respawn time expires, to prevent the player from being able to dig in the same spot over and over until they complete the quest
- "Gordunni Trap" hits the player, its script causes the player to cast either "Summon Gordunni chest (COBALT)" or "Summon Gordunni chest (JUNK)"

Perhaps when changed to force the player to cast the spell the issue of GOs being summoned underground will fix itself, if position adjustments are bound to caster and the caster will now be a non-flying unit. If not then something weird is happening in TARGET_DEST_CASTER_FRONT handler and the resulting position isn't getting adjusted to match terrain/vmaps.

The remark about needing to manually despawn "Gordunni Cobalt Visual (DND)" is valid, though. It's odd that blizzard chose to manually summon another GO alongside the Mound, when the same effect could be achieved via "linkedTrap" property, but the video does show that the blue aura despawns when the Mound is looted.

offl:
>As this spells are from spell_dbc, if there's no way to fix it in the spell script it could be fixed like this:

They shouldn't be changed in nearly all cases if ID < ~39000 since almost all of them has correct data. Those spells were taken from DBC of old builds before Blizz made them serverside, that's why only < ~39000 has effects. And nearly all other spells with ID > ~39000 has correct RangeIndex and few other fields, attributes, duration but effects and stuff like that should be added manually

That's sad we treat spell_dbc like that

How it works now https://youtu.be/5_ZPdLEprMg?t=1547

Issue #23231, opened 2019-05-02
TrinityCore does not compile on aarch64


sindastra:
**Description:**

TrinityCore does not compile on aarch64

**Current behaviour:**

TrinityCore does not compile on aarch64

**Expected behaviour:**

It should compile on aarch64

**Steps to reproduce the problem:**

Try to compile on aarch64

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

24fbbee4b9af7b5226772378dd83b78c103d969d

**Operating system:**

Linux 4.19 aarch64 Debian Stretch

**Notes:**

```
/home/trinitycore/TrinityCore/dep/g3dlite/include/G3D/AtomicInt32.h:122:29: error: invalid output constraint '=a' in asm
                          : "=a" (nz)
                            ^
/home/trinitycore/TrinityCore/dep/g3dlite/include/G3D/AtomicInt32.h:150:29: error: invalid output constraint '=a' in asm
                          : "=a" (ret)
                            ^
/home/trinitycore/TrinityCore/dep/g3dlite/include/G3D/System.h:504:15: error: invalid output constraint '=a' in asm
            : "=a" (timelo),
              ^
```

Those parts are x86 assembly which obviously won't work on ARM.

Aokromes:
trinitycore NEEDS SSE2.

Artox:
No it does not, especially not in caps.
The only place using SSE2 is the 2006-2008 implementation of the Fast Mersenne Twister algorithm.
This is easy enough to replace, with a performance impact as is usually the first step when porting applications to other cpu architectures.

However all of this is not important because one of the original author has been improving this implementation since then, including support for arm, with **and** without neon (which for you is the replacement for sse2 on arm platforms)!

So this is a non-issue and 100% qualifies as a feature.

Aokromes:
we tried to drop SSE2 requirement and performance dropped a lot, so , yes trinitycore NEEDS SSE2 and we don't go to drop that requirement to make tc to run on underpowered machines.

Shauren:
We are not willing to sacrifice x86/x64 performace for the sake of supporting architectures nobody on the team cares about

Artox:
@Aokromes I think you should stop judging feature requests, and systems so quickly. It is toxic to potential future contributors, and lacks facts.

Otherwiese, please everyone remind me how we got from a feature request to a lets make x86 slow by removing a feature?

[...]

PR #23238, opened 2019-05-03
g3d: remove and gate some x86-specific code 

Artox:
**Changes proposed:**

-  g3d: use compiler builtins for integer math on *all* platforms
-  g3d: build cpuid code only for x86

- [x] 3.3.5
- [x] master

- Enhances #23231

- master: built on Linux amd64, gcc
- master: login, create character, enter world
- 3.3.5: cherry-picked and built on Linux amd64, gcc

- needs testing on Windows and compilers other than gcc

### DIFF
--- a/dep/g3dlite/include/G3D/AtomicInt32.h
+++ b/dep/g3dlite/include/G3D/AtomicInt32.h
@@ -70,25 +70,13 @@ public:
 
     /** Returns the old value, before the add. */
     int32 add(const int32 x) {
-#       if defined(G3D_WINDOWS)
-
+        /** select implementation of atomic operation based on compiler */
+#       if defined(_MSC_VER)
             return InterlockedExchangeAdd(&m_value, x);
-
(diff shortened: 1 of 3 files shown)


jackpoz:
I would say these https://github.com/TrinityCore/TrinityCore/tree/master/cmake/compiler are the compilers we support

Artox:
> I would say these https://github.com/TrinityCore/TrinityCore/tree/master/cmake/compiler are the compilers we support

So those are gcc, icc, clang, msvc. Both icc and clang claim gcc compatiblity, and support the __sync_* atomic builtins.
And for msvc the Interlocked* are used.

Shauren:
I would also say you definitely don't need to check that gcc version, we require 6.3 or newer for --std=c++14

Warpten:
Since this is g3d, please include a patch diff in the directory tree for these changes

Shauren:
Already supported in 3.3.5 - to be cherry-picked

PR #23240, opened 2019-05-03
update SFMT to version 1.5.1

Artox:
**Changes proposed:**

-  update SFMT
- --> SIMD acceleration for x86, ppc, armv7, aarch64
- --> plain C fallback for any other architecture

- [x] 3.3.5
- [x] master

- Enhances #23231

- built on Linux amd64, Linux aarch64 - master+3.3.5
- login, create character, enter world, on Linux aarch64 - master

- This is a work in Progress!
- needs build testing on anything but Linux gcc
- needs benchmarking on x86 (to avoid performance degredation)
- should probably be tuned for performance


Shauren:
I'm not sure we need all these makefiles, htmls and params that were included in original repository

Artox:
> I'm not sure we need all these makefiles, htmls and params that were included in original repository

I thought about that, too. But I figured the sane thing to do is importing their full release tarball with all the files. That way, if anyone ever comes across those files, or if the upstream ever goes away - we have a copy of all the documentation.

jackpoz:
I too prefer to include only the files strictly necessary to build the dep and keep everything out.

Issue #19886, opened 2017-06-14
[DEV] DB/Condition: CONDITION_SOURCE_TYPE_CREATURE_TEMPLATE_VEHICLE


Killyana:
**Description:** On retail if the player enter a prohibited area with a vehicle there's a message that appears (boss emote) then after some seconds the vehicle despawns.

On TC the vehicle despawns without any notification or delay.

So I propose to add: some extra to CONDITION_SOURCE_TYPE_CREATURE_TEMPLATE_VEHICLE:

- SourceGroup= Broadcast_Text_Id

If SourceGroup>0 :  the Broadcast_Text_Id will be displayed as a boss emote and a delay of 10 secs will be used before the vehicle despawns, IF it still in the prohibited area.

If SourceGroup=0 : the condition will work like now

I'm not sure if this change must be done on CONDITION_SOURCE_TYPE_CREATURE_TEMPLATE_VEHICLE or on CONDITION_AREAID = 23

**Branch(es):** 3.3.5

**TC rev. hash/commit:** rev. b8535dab46dd

Treeston:
Feels quite a bit hacky to use condition system for heavy logic like this.

Shauren:
Thats not how it always works, some vehicles get a debuff when outside of allowed area which will perform the despawn on expiration

Wyrserth:
Maybe the list of allowed areas can be pulled from the summoning spell?

Call of the Frostbrood (53173) that summons the vehicle for the quest [An End To All Things...](https://www.wowhead.com/quest=12779) has this in dbc:
`
Allowed areas:
4355 - Crypt of Remembrance (Map: 609)
4356 - Death's Breach (Map: 609)
4347 - Havenshire (Map: 609)
4348 - Havenshire Farms (Map: 609)
4349 - Havenshire Lumber Mill (Map: 609)
4350 - Havenshire Stables (Map: 609)
4343 - New Avalon (Map: 609)
4377 - New Avalon Forge (Map: 609)
4345 - New Avalon Orchard (Map: 609)
4346 - New Avalon Town Hall (Map: 609)
4351 - Scarlet Hold (Map: 609)
4360 - Scarlet Overlook (Map: 609)
4394 - Scarlet Tavern (Map: 609)
4362 - Sinner's Folly (Map: 609)
`

Killyana:
Other spells like https://woehead.way-of-elendil.fr/?spell=55028 doesn't have such data

Issue #18089, opened 2016-10-16
Summoned vehicles despawn immediately after exit


malcrom:
**Summoned vehicles despawn immediately after exit**:

**Current behaviour**: Summoned vehicles despawn immediately after exit

**Expected behaviour**: Summoned vehicles should not despawn immediately after exit

**Fixing issue will solve problem**:
1.  While scripting the "Enter the Maelstrom" quest at a point during the ride player must change seats to change the camera. When spell 86967 is cast the player leaves seat 2 and the vehicle despawns.

**Branch(es)**: 3.3.5 / 6.x

**TC hash/commit**:  

**TDB version**:  

**Operating system**:

Treeston:
When should they despawn?

Krudor:
Why treat them differently at all?
We have PassengerBoarded(Unit* who, int8 seatId, bool apply), to use for the cases where they should actually despawn.

tkrokli:
Sounds almost like the issue 3.3.5 once had when passing a Flight Point, where the player would appear to fly without a mount and then re-mount mid-air, almost like a druid changing to flight form.

Treeston:
That is completely unrelated.

ariel-:
It's due to VehicleAI despawning, probably

dr-j:
Another example which I mentioned before is the version of gymer which is summoned for the stormkings vengence the npc should say a line and despawn a few seconds later it is scripted to do this but this never occurs as vehicle despawns instant when player dismounts

malcrom:
Yeah there should be a despawn timer.

Kittnz:
Same like the Traveler's mammoth, instant despawn.

malcrom:
And still nothing...

ariel-:
This only happens if said vehicle is conditioned (using conditions.SourceTypeOrReferenceId = 16)

gonzo1247:
Some news about this? The bug is still exist

jackpoz:
@Faq is https://github.com/TrinityCore/TrinityCore/issues/24722#issuecomment-698574244 related to this issue too ?

Faq:
About exiting, valid for 335 on a770ad0476a32007a356c3539309f7a111893145
for https://github.com/TrinityCore/TrinityCore/issues/18089#issuecomment-254069707 
Steps:
.gm on
.go c id 29647
.q add 12919
 Mount him and use exit button, he will despawn asap without saying that text.

[...]

PR #23251, opened 2019-05-07
Core/Unit: don't despawn summoned vehicles or vehicle accessories when player ejects

Wyrserth:
**Changes proposed:**

As visible [in this video](https://youtu.be/LQMLOThA2Js?t=63), a vehicle summoned via spellclick (related to the quest [Finding the Phylactery](https://www.wowhead.com/quest=11956)) dismounts the player but stays around a few seconds.

Right now the core despawns them in the next world tick, which makes any kind of event scripting impossible (that quest has the only one, as far as I know).

This is the best approach I can think of, as I do recall vehicles lingering for a few seconds before despawning - but maybe each script should handle its own vehicles' despawn instead on relying on this global despawn.

- [x] 3.3.5

**Tests performed:** works.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12605,18 +12605,11 @@ void Unit::_ExitVehicle(Position const* exitPosition)
     if (player)
         player->ResummonPetTemporaryUnSummonedIfAny();
 
-    if (vehicle->GetBase()->HasUnitTypeMask(UNIT_MASK_MINION) && vehicle->GetBase()->GetTypeId() == TYPEID_UNIT)
-        if (((Minion*)vehicle->GetBase())->GetOwner() == this)
-            vehicle->GetBase()->ToCreature()->DespawnOrUnsummon(3000);
-
     if (HasUnitTypeMask(UNIT_MASK_ACCESSORY))
     {
         // Vehicle just died, we die too
         if (vehicle->GetBase()->getDeathState() == JUST_DIED)
             setDeathState(JUST_DIED);
-        // If for other reason we as minion are exiting the vehicle (ejected, master dismounted) - unsummon
-        else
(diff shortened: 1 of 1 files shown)


Kittnz:
Can we have a sniffed value?

Killyana:
Related to https://github.com/TrinityCore/TrinityCore/issues/19886  and https://github.com/TrinityCore/TrinityCore/issues/18089

Wyrserth:
@Killyana: thanks, didn't notice those!

https://github.com/TrinityCore/TrinityCore/issues/18089#issuecomment-254069707 shows another example of scripted event on eject, which also lasts only a few seconds.

Do you know any others?

Killyana:
Every vehicle has his own aura like https://woehead.way-of-elendil.fr/?spell=69641
And the way this auras works is not the same sometime it triggers a spell if not in the in tended area and sometimes it must be one by script, an the delay changes from a vehicle to an other.
Doing it like explained in #19886 make things easier.

Wyrserth:
So there are basically two issues about summoned vehicles, one is this one (the despawn timer after player ejects) and the other one is about area restrictions that should force-eject the player.

They're different issues, but there aren't a lot of vehicle-related quests and maybe each of them handles their own despawning, without generic systems like the DespawnOrUnsummon() here.

Aokromes:
maybe put 3 by default but allow the value to be overwrited?

Wyrserth:
In the end, I think each vehicle and vehicle accessory should handle its own despawning. [Here](https://youtu.be/IWxQLen9Hp8?t=8)'s a scripted event+delayed despawn involving NPC vehicle accessories. I wouldn't be surprised if there are more.

I'll try to go through the vehicle summoning spells and have them despawn correctly, for now this PR causes them to stick around until restart or grid unload.

Wyrserth:
Will make a new PR when I've got the scripts checked and fixed to manually despawn the vehicles.

PR #23259, opened 2019-05-08
Core/Quest: Quests related to events must be removed when the event ends

### LINKED ISSUE #23123: Core/Quest: Quests related to events must be removed when the event ends
**Description:** If you have a quest related to an event like https://woehead.way-of-elendil.fr/?quest=11955 when the event ends the quest must be removed from the journal

**Current behavior:** Quests related to events are not removed from the journal when the event ends

**Expected behavior:** Quests related to events must be removed from the journal when the event ends.

1. .event start 1
2. .quest add 11955
3. .event stop 1
4. Notice the quest remains in the journal event if you disconnect/reconnect your char

**Branch(es):** both

**TC rev. hash/commit:**  rev. 995aabe4bbbe

kuskmen:
**Changes proposed:**
-  Removes all quests related to not active events from player's quest log.
- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #23123 

**Tests performed:** 
.event start 1
.quest add 11955
.event stop 1
relog **twice**

- [ ] `PlayerScript::OnLogin` fires query that removes the quest for a character and only after second login quest is removed.
- [ ] Kinda worried about performance issues, I mean it seems to me that this script might turn out to be overkill, open for suggestions.


src/server/scripts/World/game_event_quest_cleanup.cpp
```diff
@@ -0,0 +1,73 @@
...
+    GAME_EVENT_MIDSUMMER_FIRE_FESTIVAL = 1,
+    GAME_EVENT_WINTER_VEIL = 2,
+    GAME_EVENT_LUNAR_FESTIVAL = 7,
+    GAME_EVENT_LOVE_IS_IN_THE_AIR = 8,
+    GAME_EVENT_NOBLEGARDEN = 9,
+    GAME_EVENT_CHILDRENS_WEEK = 10,
+    GAME_EVENT_HARVEST_FESTIVAL = 11,
+    GAME_EVENT_HALLOWS_END = 12,
+    GAME_EVENT_BREWFEST = 24,
+    GAME_EVENT_PILGRIMS_BOUNTY = 26,
```
Killyana: Day of the Dead = 51

ccrs:
I'd just integrate this into quest load/save system
and add a logic related to event ending -> remove current online players quests

kuskmen:
> I'd just integrate this into quest load/save system
> and add a logic related to event ending -> remove current online players quests

What happens to people not being online at the moment of event end?

kuskmen:
> I'd just integrate this into quest load/save system
> and add a logic related to event ending -> remove current online players quests

Can you elaborate more, I am still not at the point where I can understand without giving more specific instructions :/

Where is this quest load/save system and how can I iterate over all online players?

Kittnz:
This seems like a custom thing, i can't remember event quest getting removed when the event ended.

Jildor:
I don't know if this is blizz or not, but, onLogin seems bad idea for performance.
Maybe do a cleanup on startup server if a conf option is enable?

jackpoz:
> onLogin seems bad idea for performance.

well it depends on what code is being run and what impact it has

illfated:
> This seems like a custom thing, i can't remember event quest getting removed when the event ended.

Fair point, neither can I (or is this a Mandela Effect, that we remember this incorrectly?).

Killyana:
When an event starts quests related to this event are reset, and the player can do this quests again.
So if he keep a quest from a chain it doesn't make cense to reward this quest then take the qeust that starts this chain.

illfated:
> When an event starts quests related to this event are reset, and the player can do this quests again.
> So if he keep a quest from a chain it doesn't make sense to reward this quest then take the quest that starts this chain.

Fair point, I didn't notice that the outcome would be that the event quest would be rewarded when removed.

src/server/scripts/World/game_event_quest_cleanup.cpp
```diff
@@ -0,0 +1,77 @@
...
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <ScriptMgr.h>
+#include <Player.h>
+#include <GameEventMgr.h>
+
+enum GameEvents : uint16
+{
+    GAME_EVENT_MIDSUMMER_FIRE_FESTIVAL    = 1,
```
jackpoz: maybe a flag in the db would be better

src/server/scripts/World/game_event_quest_cleanup.cpp
```diff
@@ -0,0 +1,77 @@
...
+    GAME_EVENT_PILGRIMS_BOUNTY            = 26,
+    GAME_EVENT_DAY_OF_THE_DEAD            = 51
+};
+
+class GameEventQuestCleanupScript : public PlayerScript
+{
+    public:
+        GameEventQuestCleanupScript() : PlayerScript("GameEventQuestCleanupScript") { }
+
+        void OnLogin(Player* player, bool /*firstLogin*/) override
```
jackpoz: if we choose to keep the OnLogin() PlayerScript (not that each PlayerScript has a lot of hooks and even if empty, they add up), I would like you to profile a bit this code and find the most performant way to achieve the same result.
Maybe iterating the active player quests first, then check the game event mgr ? or maybe that's slow ? I have no clue.

jackpoz: any follow up on this ?

illfated:
@kuskmen : I hope you will be able to look into this PR of yours, now that there are some holidays ahead.

kuskmen:
@illfated hehe quite some holidays ahead I admit, I am gonna have to re-read all the comments and see what this stage was left off. If you guys say we still need this I can do it asap.

msoky:
@kuskmen could you check if there is sometnig needed to finish it?

kuskmen:
I think I read somewhere that this is not blizzlike behaviour so I thought it should be canceled? Am I wrong?

illfated:
I am thinking maybe it would be better to move it first to the https://github.com/TrinityCore/TrinityCoreCustomChanges repository, let users play with it there for a while, then get it back here on 3.3.5 when it gets some more feedback after actual usage. What do you others think, is that a fair suggestion?

jackpoz:

Issue #16416, opened 2016-01-28
 [3.3.5][6.x] Core: Creature equipement


Killyana:
equipment_id field on creature table is not working correctly, if it's set to 0 the npc still spawn with the equipement.
rev. cdf7ac60965f

Saben65:
Isn't that by design because everything else in the `creature` table is inherited from the `creature_template` if left empty??

Killyana:
Yes but there's nothing related to equipments on creature_template

Saben65:
Ah sorry, was thinking of `creature_equip_template`.

Kizarr:
It is located in creature_equip_template because I fixed this issue before manually I had to find the weapons and shield ID from wowhead for this npc's became TrinityCore refuses to acknowledge it.

Langerz82:
I changed the creature_equip_template on the db and used "reload all" and it had a stale entry and wasnt reloaded from the db. However restarting the Server the equipped weapons are showing.

Langerz82:
This simple patch should allow reloading of the creature equip template.

See:
https://gist.github.com/Langerz82/1d3c48e06f15f4d505b1f0c11c76b949

PR #23273, opened 2019-05-13
Core/Creature: equipment_id field on creature table is not working co…

Aokromes:
…rrectly, if it's set to 0 the npc still spawn with the equipement.

By Malcrom
Closes TrinityCore#16416

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -489,9 +489,11 @@ bool Creature::InitEntry(uint32 entry, CreatureData const* data /*= nullptr*/)
     SetNativeDisplayId(displayID);
 
     // Load creature equipment
-    if (!data || data->equipmentId == 0)
-        LoadEquipment(); // use default equipment (if available)
-    else                // override, 0 means no equipment
+    if (!data)
+        LoadEquipment();  // use default equipment (if available) for summons
+    else if (data->equipmentId == 0)
+        LoadEquipment(0); // 0 means no equipment for creature table
+    else                         
     {
         m_originalEquipmentId = data->equipmentId;
         LoadEquipment(data->equipmentId);


Killyana:
Tested and works as intended.
For DB spawns:
If equipment_id=0 no equipment loaded
If equipment_id=n>0  equipment=n is loaded
If equipment_id=-1 a random equipment is loaded

For a summoned npc: the first equipment in the DB will be loaded

For npcs that use scripts to load equipments it works fine.

malcrom:
But why would a summoned npc have an equipment 2 and no equipment 1?

Killyana:
Maybe there's no equipment=1

malcrom:
Shouldn't 2 be changed to 1 then? These are not sniffed values.

Killyana:
Yes we can change this cases, in 3.3.5 there's only 2

malcrom:
UPDATE `creature_equip_template` SET `ID` = 1 WHERE `CreatureID`=23784;

Killyana:
It' not a summoned npc so it doen't matter

jackpoz:
If this PR requires some SQL scripts, please include those in the PR.
By git blaming the modified code, it seems like it went through quite many changes in the past, so please check how this will affect other creatures too, not just the one reported in https://github.com/TrinityCore/TrinityCore/issues/16416

Killyana:
I already tested all cases, every thing works correctly, I will not provide any sql before this pr is merged as it requires lots time, and it looks this pr is refused.

jackpoz:
But "as it requires lots of time" means that if we merge it now, 335 will be broken until you decide to push whatever SQL is required. While if this PR gets compete properly, it can be merged.

malcrom:
Update creature SET equipment_id=1 WHERE equipment_id=0 AND  id = (SELECT CreatureID  FROM creature_equip_template WHERE ID=1);

Killyana:
Push the PR, if all equipments are not fixed in both branchs until tomorrow, revert it.

illfated:
Do we have a known list of NPCs not wearing/using/wielding equipment, but should be able to do so after this PR gets merged? Or will we need to apply DB changes (SQL files) after this PR is merged to see actual changes? (At the moment, the only NPC I can remember not using his intended tool/weapon is [Awilo Lon'gomba `<Grand Master Cooking Trainer>`](https://woehead.way-of-elendil.fr/?npc=29631) (NPC ID 29631) in Dalaran.)

malcrom:
ATM there are npcs with equipment that shouldn't have it. This along with Kilyanas SQL will put it at the same state but give us the ability to fix the NPCs that shouldn't have equipment.

Issue #23284, opened 2019-05-15
Quest/Script: A Visit With The Ancestors


Killyana:
**Description:** When an objective is completed 2 https://woehead.way-of-elendil.fr/?npc=18904 spawns and whisper something to the player
This npcs are summoned by https://woehead.way-of-elendil.fr/?spell=33122

This npcs: 18840, 18841, 18842, 18843 have this script "npc_creditmarker_visit_with_ancestors" to just give quest credit, we must add to extra cast done by the player with 2 secs interval for spell=33122
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Outland/zone_nagrand.cpp#L225

The issue here I'm not sure we have a something to check quest objective as the player must cast spell=33122 only if the quest is not complete and the objective related to the npc is not fulfill.

``` sql
UPDATE `creature_template` SET `flags_extra`= `flags_extra`|128 WHERE `entry` IN (18840, 18841, 18842, 18843);
UPDATE `creature_template` SET `AIName`="SmartAI",`unit_flags`= `unit_flags`|768 WHERE `entry` IN (18904);
DELETE FROM `smart_scripts` WHERE `entryorguid` IN (18904) AND source_type=0;
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(18904,0,0,0,54,0,100,0,0,0,0,0,1,0,0,0,0,0,0,7,0,0,0,0,0,0,0,"Vision of the Forgotten - On Just summoned - Say text");

DELETE FROM `creature_text` WHERE `CreatureID`=18904;
INSERT INTO `creature_text` (`CreatureID`, `GroupID`, `ID`, `Text`, `Type`, `Language`, `Probability`, `Emote`, `Duration`, `Sound`, `BroadcastTextId`, `TextRange`, `comment`) VALUES
(18904,0,0,"Turn back, mortal... This is not your battle.",15,0,100,0,0,0,16007,0,"Vision of the Forgotten"),
[...]

Issue #16037, opened 2015-12-17
[3.3.5][6.x] DB/Quest: The Great Hunter's Challenge


aletson:
b967047

[Quest](http://www.wowhead.com/quest=12592)

[Drostan](http://www.wowhead.com/npc=28328) doesn't come out and talk to you while you're killing stuff.

[This video](https://www.youtube.com/watch?v=cAKV8NhfP7E) could be useful to see at which points he comes out and taunts you.

In #11003 this is mentioned, but the issue was closed before that part of the bug could be fixed. So, new issue! Please let me know if it's better practice to reopen the old issue.

tkrokli:
Suggestion: when you post links to the creature/quest/item/spell page on wowhead, show the link itself:

http://www.wowhead.com/quest=12592/the-great-hunters-challenge

it displays both name from the page and the ID in 1 single line, more visually useful than "Link" etc.

velinath:
Sounds good; I'll start doing that in the future.

Rushor:
maybe there is sth like a aura for the player which handles the summon/despawn of Drostan

Killyana:
Spell: http://wotlk.openwow.com/spell=52556
Npc: www.wowhead.com/npc=28857/drostan

Rushor:
hm but how should http://wotlk.openwow.com/spell=52559 behave?

if the player is 5 seconds in combat, the triggered effect http://wotlk.openwow.com/spell=52556 should be executed and drostan should spawn?

or maybe another solution:

every questmob has an sai - it the questmob in 5 seconds in combat and player has quest: "The Great Hunter's Challenge" active, drostan will spawn

Rushor:
all involved creatures:

http://www.wowhead.com/npc=28009/shardhorn-rhino
http://www.wowhead.com/npc=28001/dreadsaber
http://www.wowhead.com/npc=28011/emperor-cobra
http://www.wowhead.com/npc=28002/mangal-crocolisk
http://www.wowhead.com/npc=28380/shattertusk-bull
http://www.wowhead.com/npc=28297/shango
http://www.wowhead.com/npc=28379/shattertusk-mammoth
http://www.wowhead.com/npc=28378/primordial-drake

i will try to add a script later

tkrokli:
Maybe worth considering: Drostan doesn't appear 60 times (once for each animal). Quote:
[...]

PR #23286, opened 2019-05-17
Core/Script: implement a PlayerScript hook called when a quest's objective receives progress

Wyrserth:
**Changes proposed:**

This implements a PlayerScript hook that's called whenever the player obtains progress in a quest that uses quest_template.RequiredNpcOrGoX fields.

It allows access to the player object, the id of the npc or gameobject target of the quest, and the new objective's count (if a quest requires to kill 25 of the same npc, the player is at 14/25 and kills one, the hook is called with count=15 and creatureOrGoEntry=whatever the npc entry is). 

- [x] 3.3.5
- [ ] master

**Issues addressed:** adds core support to allow proper fixing of #23284 and #16037.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -16766,6 +16766,8 @@ void Player::SendQuestUpdateAddCreatureOrGo(Quest const* quest, ObjectGuid guid,
     uint16 log_slot = FindQuestSlot(quest->GetQuestId());
     if (log_slot < MAX_QUEST_LOG_SIZE)
         SetQuestSlotCounter(log_slot, creatureOrGO_idx, GetQuestSlotCounter(log_slot, creatureOrGO_idx) + add_count);
+
+    sScriptMgr->OnPlayerReachedQuestObjectiveCount(this, quest->GetQuestId(), creatureOrGO_idx, old_count + add_count);
 }
 
 void Player::SendQuestUpdateAddPlayer(Quest const* quest, uint16 old_count, uint16 add_count)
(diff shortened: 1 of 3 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -16766,6 +16766,8 @@ void Player::SendQuestUpdateAddCreatureOrGo(Quest const* quest, ObjectGuid guid,
     uint16 log_slot = FindQuestSlot(quest->GetQuestId());
     if (log_slot < MAX_QUEST_LOG_SIZE)
         SetQuestSlotCounter(log_slot, creatureOrGO_idx, GetQuestSlotCounter(log_slot, creatureOrGO_idx) + add_count);
+
+    sScriptMgr->OnPlayerReachedQuestObjectiveCount(this, quest->GetQuestId(), creatureOrGO_idx, old_count + add_count);
```
Shauren: The value you are using here, `creatureOrGO_idx`, is not creature/gameobject entry but index of the objective being completed in quest_template

Now, I think it is actually better to use the objective index because it will let you also grab required count from  quest template

Shauren: The value you are using here, `creatureOrGO_idx`, is not creature/gameobject entry but index of the objective being completed in quest_template

Now, I think it is actually better to use the objective index because it will let you also grab required count from  quest template

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -1013,6 +1016,7 @@ class TC_GAME_API ScriptMgr
         void OnPlayerUpdateZone(Player* player, uint32 newZone, uint32 newArea);
         void OnQuestStatusChange(Player* player, uint32 questId);
         void OnPlayerRepop(Player* player);
+        void OnPlayerReachedQuestObjectiveCount(Player* player, uint32 questId, uint32 creatureOrGoEntry, uint8 count);
```
Shauren: I would rename this to `OnQuestObjectiveProgress`, and change arguments
* `uint32 questId` to `Quest const* quest`
* `uint32 creatureOrGoEntry` to `uint32 objectiveIndex` (see the other comment)
* `uint8 count` to `uint16 progress` (rename + change type, updatefields store uint16s)

Shauren (changes requested):

Wyrserth:
Done, thanks!

illfated:
Cosmetic detail in the PR title: **Core/Scripting:** would be a suitable prefix.
(feel free to ignore this comment if you find it redundant)

Shauren (approved):

PR #23319, opened 2019-05-26
Core/Misc: don't allow players to use/activate/loot non-allowed gameobjects while under the effect of a damage immunity aura

Wyrserth:
**Changes proposed:**

- Set UNIT_FLAG_HAS_IMMUNITY_AURA (previously UNIT_FLAG_UNK_31) on units affected by damage immunity auras and prevents them from interacting with certain gameobjects while said aura is active.
This flag disables the client from interacting with certain gameobjects while it's set.

- Also add some anti-cheating measures.

- Remove flag usage from some boss scripts, now it's set and unset appropriately when the relative auras are applied/removed.

Prevents, for example, Paladins from looting chests (like the Gurubashi Arena one) while bubbled.

Thanks to xvwyh for detailed post about this issue [here](https://github.com/TrinityCore/TrinityCore/pull/23207#discussion_r279888053).

- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/server/game/Entities/GameObject/GameObjectData.h
+++ b/src/server/game/Entities/GameObject/GameObjectData.h
@@ -498,10 +498,11 @@ struct GameObjectTemplate
             case GAMEOBJECT_TYPE_DOOR:       return door.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_BUTTON:     return button.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_QUESTGIVER: return questgiver.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_CHEST:      return true;                           // All chests cannot be opened while immune on 3.3.5a
             case GAMEOBJECT_TYPE_GOOBER:     return goober.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGSTAND:  return flagstand.noDamageImmune != 0;
             case GAMEOBJECT_TYPE_FLAGDROP:   return flagdrop.noDamageImmune != 0;
-            default: return true;
+            default: return false;
         }
     }
 
--- a/src/server/game/Entities/GameObject/GameObjectData.h
+++ b/src/server/game/Entities/GameObject/GameObjectData.h
@@ -491,6 +491,20 @@ struct GameObjectTemplate
         }
     }
 
+    bool CannotBeUsedUnderImmunity() const // Cannot be used/activated/looted by players under immunity effects (example: Divine Shield)
+    {
+        switch (type)
+        {
+            case GAMEOBJECT_TYPE_DOOR:       return door.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_BUTTON:     return button.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_QUESTGIVER: return questgiver.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_GOOBER:     return goober.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_FLAGSTAND:  return flagstand.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_FLAGDROP:   return flagdrop.noDamageImmune != 0;
+            default: return true;
+        }
+    }
+
     uint32 GetCharges() const                               // despawn at uses amount
     {
         switch (type)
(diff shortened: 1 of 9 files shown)


src/server/game/Entities/Unit/UnitDefines.h
```diff
@@ -152,7 +152,7 @@ enum UnitFlags : uint32
     UNIT_FLAG_UNK_28                = 0x10000000,
     UNIT_FLAG_UNK_29                = 0x20000000,           // used in Feing Death spell
     UNIT_FLAG_SHEATHE               = 0x40000000,
-    UNIT_FLAG_UNK_31                = 0x80000000,
+    UNIT_FLAG_HAS_IMMUNITY_AURA     = 0x80000000,
```
Shauren: Not researched this on later versions but it might be used for more things than just auras, better to simply call it "IMMUNE"

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -3162,7 +3162,12 @@ void AuraEffect::HandleAuraModSchoolImmunity(AuraApplication const* aurApp, uint
         target->RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_IMMUNE_OR_LOST_SELECTION);
 
     if (apply)
+    {
+        target->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_HAS_IMMUNITY_AURA);
         target->GetThreatManager().EvaluateSuppressed();
+    }
+    else
```
Shauren: Remove case needs the typical `if (!target->HasAuraType(...))` check found in all aura handlers that set/remove flags

Shauren (changes requested):

src/server/game/Entities/GameObject/GameObjectData.h
```diff
@@ -491,6 +491,20 @@ struct GameObjectTemplate
...
+    {
+        switch (type)
+        {
+            case GAMEOBJECT_TYPE_DOOR:       return door.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_BUTTON:     return button.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_QUESTGIVER: return questgiver.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_GOOBER:     return goober.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_FLAGSTAND:  return flagstand.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_FLAGDROP:   return flagdrop.noDamageImmune != 0;
+            default: return true;
```
Shauren: GAMEOBJECT_TYPE_CHEST must return true in all cases,
default must return false

Shauren (changes requested):

src/server/game/Entities/GameObject/GameObjectData.h
```diff
@@ -491,6 +491,20 @@ struct GameObjectTemplate
...
+    bool CannotBeUsedUnderImmunity() const // Cannot be used/activated/looted by players under immunity effects (example: Divine Shield)
+    {
+        switch (type)
+        {
+            case GAMEOBJECT_TYPE_DOOR:       return door.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_BUTTON:     return button.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_QUESTGIVER: return questgiver.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_GOOBER:     return goober.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_FLAGSTAND:  return flagstand.noDamageImmune != 0;
+            case GAMEOBJECT_TYPE_FLAGDROP:   return flagdrop.noDamageImmune != 0;
```
Killyana: Add GAMEOBJECT_TYPE_CHEST when data22: Damage Immune =0

Wyrserth:
Done, thanks!

@Killyana I thought that only applies to newer expansions, data for chests stops being relevant past 17 and in 3.3.5a they're all set to 0. Might have to change that when/if it's ported to master.

Killyana:
Chests in master have data22: Damage Immune and all have data22=0
Chest were not usable under immunity in all expansions.

jackpoz:
@Shauren anything more or can this be merged ?

jackpoz:
@Wyrserth is this ready to be merged ?

Wyrserth:
@jackpoz yeah there's nothing left to do that I know of.

jackpoz:
Thanks for the PR :)

PR #23321, opened 2019-05-26
Core/Console: start CliRunnable thread after all log messages have been printed

Wyrserth:
**Changes proposed:**

This is a very cosmetic change. The command line input thread was started (which outputs "TC>" and waits for input) before all log messages were printed, so the console would show "TC>[git revision] ready..." on the same line.

It makes more sense that the CLI input becomes available after the very last loading message has been printed to console.

- [x] 3.3.5
- [ ] master

**Tests performed:** builds, launched worldserver a few times to make sure nothing breaks.

### DIFF
--- a/src/server/worldserver/Main.cpp
+++ b/src/server/worldserver/Main.cpp
@@ -307,17 +307,6 @@ extern int main(int argc, char** argv)
         ClearOnlineAccounts();
     });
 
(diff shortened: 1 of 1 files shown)


MaxtorCoder:
Finally someone changes that 😂😂😂

Shauren (approved):

PR #20447, opened 2017-09-26
Core/SAI: implement SMART_ACTION_EXIT_VEHICLE and fix orientation in _ExitVehicle().

Wyreth:
**Changes proposed:**

-  Implement SMART_ACTION_EXIT_VEHICLE, allowing to choose the exit position for passenger by using SMART_TARGET_POSITION, or choose which entity should eject vehicle, using other target types.

This is needed for cases where the only vehicle aura applied is [Ride Vehicle Hardcoded](http://www.wowhead.com/spell=46598) (unapplying it doesn't dismount entity) and where the exit position is not default (example: [this quest](https://youtu.be/eGsEu1rPcOU?t=38).

- Fix orientation used in _ExitVehicle(). Previously it was using the original vehicle's orientation instead of the one set in the Position parameter.

**Target branch(es):**

- [x] 3.3.5

**Tests performed:** works fine.

PR #23322, opened 2019-05-27
Core/Script: add script hook to allow overriding of a vehicle passenger's exit position

Wyrserth:
**Changes proposed:**

Allow an UnitScript to override/modify the exit position of a vehicle's passenger when it exits a vehicle.

Some vehicles have a fixed exit position (#20447), others have an exit position relative to the vehicle's position ([Traveler's Tundra Mammoth](https://youtu.be/IWxQLen9Hp8?t=172)) and most vehicles eject the player to a nearby position (maybe it has something to do with bounding radius in that case?). DBC data doesn't seem to be enough to calculate a position, so I think this is the best alternative (being an UnitScript it works for both player and creature passengers and does not interfere with AI scripts).

- [x] 3.3.5
- [ ] master

**Tests performed:** builds, works.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12579,13 +12579,6 @@ void Unit::_ExitVehicle(Position const* exitPosition)
 
     SetControlled(false, UNIT_STATE_ROOT);      // SMSG_MOVE_FORCE_UNROOT, ~MOVEMENTFLAG_ROOT
 
-    Position pos;
-    if (!exitPosition)                          // Exit position not specified
-        pos = vehicle->GetBase()->GetPosition();  // This should use passenger's current position, leaving it as it is now
-                                                // because we calculate positions incorrect (sometimes under map)
-    else
-        pos = *exitPosition;
-
     AddUnitState(UNIT_STATE_MOVE);
 
     if (player)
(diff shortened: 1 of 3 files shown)


jackpoz:
do you have an example of use case ? or do you plan on having one ? (it doesn't have to be inside this PR but just wondering)

Wyrserth:
@jackpoz yeah, [this one](https://gist.github.com/Wyrserth/c7a897e58d526354e9e76ba55df07580) to fix the Traveler's Tundra Mammoth's NPCs, as they're supposed to eject behind the player and in their respective left/right direction based on player's position.

Would be roughly the same for #20447's usecase but with absolute coordinates assigned to pos instead of relative ones.

I think it's better to have separate multiple scripts that can be placed in the related NPCs' zone script file, but one monolithic unit would be possible as well.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12597,6 +12590,14 @@ void Unit::_ExitVehicle(Position const* exitPosition)
         SendMessageToSet(&data, false);
     }
 
+    // Default exit position to vehicle position
```
jackpoz: any reason to move this code down here from where it was before ?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12606,7 +12607,7 @@ void Unit::_ExitVehicle(Position const* exitPosition)
         init.SetFall();
 
     init.MoveTo(pos.GetPositionX(), pos.GetPositionY(), height, false);
-    init.SetFacing(GetOrientation());
+    init.SetFacing(pos.GetOrientation());
```
jackpoz: should this take into account the current unit orientation like it did before ? at least when scripts don't change it ? especially in case no exit position is being passed as parameter

src/server/game/Entities/Unit/Unit.cpp
Wyrserth: I moved it because I think it makes sense to declare pos right before it's actually used, no functionality change.

src/server/game/Entities/Unit/Unit.cpp
Wyrserth: You're right, I meant to have pos contain the current orientation when I changed that, but forgot to actually do it. Thanks!

jackpoz:
Thanks for the PR :)

PR #23336, opened 2019-06-01
Core/Groups: do not allow lfg groups to change the loot mode

Ovahlord:
**Changes proposed:**

-  add a new check to the loot method opcode handler to block any attempt of changing the loot mode while inside a lfg group. The UI itself blocks the method setting but you can still run scripts that will trigger that opcode so we will block it from now on.

- [x] 3.3.5
- [x] master

Tested ingame by some mates, works fine so far.


MaxtorCoder:
Maybe good to push this one into `master` too? Since it is marked for `3.3.5` and `master`

illfated:
Should be an easy cherry-pick, but I don't remember if there are other rules or prerequisites for that to happen.

PR #20977, opened 2017-11-26
Core/CreatureTemplate: Move resistances and spells into seperate tables

Seyden:
**Changes proposed:**

Move resistances and spells into seperate tables to reduce the table size and to have a structure
Outsourced resistances and spells into `creature_template_spell` and `creature_template_resistance`

**Target branch(es):** 3.3.5/master
- [x] master

**Issues addressed:** Closes #  (insert issue tracker number)

**Tests performed:** (Does it build, tested in-game, etc.)
- [x] Builds

PR #23343, opened 2019-06-02
Core/CreatureTemplate: Outsourced resistances and spells into creatur…

jackpoz:
Core/CreatureTemplate: Outsourced resistances and spells into creature_template_spell and creature_template_resistance

**Changes proposed:**

-  Apply https://github.com/TrinityCore/TrinityCore/pull/20977 to 335

- [x] 3.3.5
- [ ] master

- [ ] Needs to be tested
- [ ] GitHub diff is a mess and shows like if fields are not loaded anymore


brotalnia:
If you are moving the spells to a separate table, why not also implement the other creature spell data fields we know exist in Blizzard's DB based on the shared WowEdit screenshots and developer interviews. Its not just a bunch of Spell IDs, there is also initial delay, repeat delay, probability and availability.

The "creature spell lists" were discussed here:
https://github.com/TrinityCore/TrinityCore/pull/22461

Screenshot of blizzard's creature spells table editor:
https://static.mmo-champion.com/mmoc/images/news/2009/august/raidsdungeons_021.jpg

Explanation of extra creature spell entry fields:
Availability - Percent chance for the creature to spawn in with that spell.
Probability - Likelihood to cast this particular spell if multiple are available on the same tick.
Initial Delay (Min/Max) - Seconds to wait after entering combat before being allowed to cast the spell.
Repeat Delay (Min/Max) - Cooldown in seconds before being allowed to cast this spell again.

There should also be a name for each creature spells list, since creatures can switch to a different list on complex encounters. You can see from the screenshot that Onyxia has a separate spells list for each phase of the fight. And there are also 2 other fields on the list itself to specify the chance to "consider" casting a spell marked as Support or Ranged upon each tick (spell lists are updated every ~1.2 seconds).

Suggested table structure:

Table `creature_spell_template`:
**ID** - uint32, primary key
**Name** - string
**ChanceSupportAction** - uint8
**ChanceRangedAttack** - uint8

Table `creature_spell`:
**ListID** - uint32, primary key
**Index** - uint8, primary key
**SpellID** - uint32
**Availability** - uint8
**Probability** - uint8
**InitialDelayMin** - uint32
**InitialDelayMax** - uint32
**RepeatDelayMin** - uint32
**RepeatDelayMax** - uint32

There should also be a default `SpellListID` field on the `creature_template` record, which is the spells list that gets assigned to the creature on spawn.

You can see there is indeed a default Spells/Abilities List ID that can be assigned to a creature template record on this WowEdit screenshot:
https://static.mmo-champion.com/mmoc/images/news/2009/august/raidsdungeons_020.jpg

But it's not the only spell list a creature can have as discussed earlier. It should be possible to switch to a different spells list from a script.

AI_GROUP_UNIT_SPELLS - "Set the spells record for unit(s)"

From the datamined WowEdit creature script commands here:
https://docs.google.com/spreadsheets/d/1Wh1e1ZXYM-sw3wWrY9dGZ2ngK8oa3PA2XIftQJuf5R0/edit#gid=740715654

Aokromes:
this pr ain't for the creatures using the spells, instead for mind controlled creatures spell availability.

jackpoz:
Note: this will not load resistance/spells when reloading a single creature template with the chat command ".reload creature_template <entry>" , feel free to implement that. Just reload all creature templates.

PR #23345, opened 2019-06-02
Core/SAI: add action_param3 to SMART_ACTION_REMOVEAURASFROMSPELL to allow it to remove only auras from spells that were cast by the entity itself

Wyrserth:
**Changes proposed:**

There are cases where a script must remove an aura from a creature, but said creature has other copies of the same aura coming from different casters. In that case we might want to only remove the aura owned by the entity that is executing said script.

Setting action_param3 to 1 allows this behavior, while 0 (the default value) retains original behavior.

- [x] 3.3.5


jackpoz:
Thanks for the PR :)

PR #23346, opened 2019-06-02
Script/Quest: implement event for the quest Orabus the Helmsman

Wyrserth:
**Changes proposed:**

Original script [here](https://github.com/TrinityCore/TrinityCore/pull/20447#issuecomment-332803817), I just adapted it to newer TC changes and added a condition for the quest item.

- [x] 3.3.5

**Tests performed:** builds, works like [this retail video](https://www.youtube.com/watch?v=eGsEu1rPcOU).


src/server/scripts/Northrend/zone_borean_tundra.cpp
```diff
@@ -2503,6 +2503,46 @@ class npc_bloodmage_laurith : public CreatureScript
...
+    npc_orabus_the_helmsman_ship_exit_pos() : UnitScript("npc_orabus_the_helmsman_ship_exit_pos") { }
+
+    void ModifyVehiclePassengerExitPos(Unit* passenger, Vehicle* /*vehicle*/, Position& pos)
+    {
+        if (passenger->GetTypeId() == TYPEID_UNIT)
+        {
+            switch (passenger->GetEntry())
+            {
+            case NPC_ORABUS:
+                pos.Relocate(2807.858f, 7038.571f, 7.075813f, 4.73481f);
```
jackpoz: are these supposed to be hardcoded or is there any offset that might make more sense ?

Wyrserth: I thought hardcoded made more sense as the boat is always on the same position so the exit pos will always be the same regardless.

[I found a sniff](https://gist.github.com/Wyrserth/8f7e650d25f65eefb31f1db17b93b992) lying around and all the coordinates match with retail (I actually made a mistake and inverted the coordinates of some npc entries that were correct in the original script, fixing it now).

So I would leave them hardcoded as I don't think I can replicate the same coordinates with offsets, if that's alright.

jackpoz:
Thanks for the PR :)

Issue #23366, opened 2019-06-05
DB/SAI: don't make NPCs involved in quest A Lesson in Fear disappear when the quest item is used on them


Wyrserth:
**Description:**

As visible in [this video](https://youtu.be/D5cmNfyejZE?t=11), using the [Forsaken Banner](https://www.wowhead.com/item=33563) during the quest [A Lesson in Fear](https://www.wowhead.com/quest=11282) does not cause the mob's corpse to despawn.

On TC it currently despawns instantly, which makes it far more exasperating than it needs to be.

**Steps to reproduce the problem:**

1. .quest add 11282
2. .go creature id 24161
3. kill mobs until one of the quest targets appears, kill him
4. use quest item

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  1f85348

**Suggested fix:**
```sql
DELETE FROM `smart_scripts` WHERE `entryorguid` IN (24016, 24161, 24162) AND `source_type`=0;
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(24016, 0, 0, 0, 9, 0, 100, 0, 0, 5, 7000, 12000, 0, 11, 48288, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, "Ulf the Bloodletter - Within 0-5 Range - Cast 'Mace Smash' (Phase 1) (No Repeat)"),
(24016, 0, 1, 0, 6, 0, 100, 513, 0, 0, 0, 0, 0, 22, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, "Ulf the Bloodletter - On Just Died - Set Event Phase 1 (Phase 1) (No Repeat)"),
[...]

Issue #21772, opened 2018-04-03
Core/Spell: Quest items created by spellclick


ghost:
**Description:**

The quest spell **[Rifle the Bodies: Create Mage Hunter Personal Effects](http://www.wowhead.com/spell=47096/rifle-the-bodies-create-mage-hunter-personal-effects)** does not create any loot item.

The quest **[Rifle the Bodies](http://www.wowhead.com/quest=11999/rifle-the-bodies)** can not be completed, because [**Mage Hunter Personal Effects**](http://www.wowhead.com/item=35792/mage-hunter-personal-effects#contains) never drops and therefore **[Moonrest Gardens Plans](http://www.wowhead.com/item=35783/moonrest-gardens-plans)** can not be looted.

Related links:
- https://wow.gamepedia.com/Quest:Rifle_the_Bodies_(Horde)
- https://wow.gamepedia.com/Quest:Rifle_the_Bodies_(Alliance)
- http://www.wowhead.com/quest=11999/rifle-the-bodies
- http://www.wowhead.com/quest=12000/rifle-the-bodies
- http://www.wowhead.com/npc=26477/dead-mage-hunter#comments
- http://www.wowhead.com/spell=47096/rifle-the-bodies-create-mage-hunter-personal-effects
- http://www.wowhead.com/item=35792/mage-hunter-personal-effects
- http://www.wowhead.com/item=35783/moonrest-gardens-plans

I noticed in `creature_template` that creature entry 26477 does not have any loot table, but I presume that is because a corpse can't have conditional loot based on whether the quest is active or not.

See also https://woehead.way-of-elendil.fr/?spell=47096 for condition details.

@Killyana : I noticed that https://woehead.way-of-elendil.fr/?item=35792 says **"This item is not available to players."** -- in contrast to the wowhead item page. Does that mean it has reacted to the item never dropping on current TC 3.3.5 source, seeing that woehead.way-of-elendil.fr seems to be pretty up to date with TC 3.3.5 content?

**Current behaviour:**

[...]

PR #23367, opened 2019-06-05
Core/Creature: prevent dead creatures from setting a focus target.

Wyrserth:
**Changes proposed:**

Apply the SQL script in #23366 and try to do the quest. You'll notice the dead creature will set a focus on the player and turn around when the player moves, after using the Forsaken Banner.

This PR prevents dead creatures from obtaining a focus target, as they aren't supposed to move around.

- [x] 3.3.5

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -3011,6 +3011,10 @@ void Creature::FocusTarget(Spell const* focusSpell, WorldObject const* target)
     if (m_focusSpell)
         return;
 
+    // Prevent dead creatures from setting a focus target, so they won't turn
+    if (!IsAlive())
+        return;
+
     // some spells shouldn't track targets
     if (focusSpell->IsFocusDisabled())
         return;


ccrs:
Can you please check if ```FocusTarget``` uses have any sort of check before the actual call?
I find it really odd that this got through after all this time (since its impl and several corrections)

Killyana:
This also happen with this quest https://www.wowhead.com/quest=11999/rifle-the-bodies when you click the npc 26477  it will turn to the player then reset orientation, even if it's considered as dead.
The issue was reported one year ago https://github.com/TrinityCore/TrinityCore/issues/21772

Wyrserth:
Only checks are to make sure the creature isn't possessed (UNIT_FLAG_POSSESSED) and is only turning for real spell casts and not on ON_NEXT_SWING spells.

I'll try to get a callstack, I'm not sure what the spell cast by the quest item is doing internally, I suppose it makes the creature cast a spell back to the player so that's where the focus issue comes from.

Honestly there are so few of those cases, and it's just a cosmetic issue, so I'm not surprised it hasn't been noticed in the past :P

Wyrserth:
@Killyana can https://github.com/TrinityCore/TrinityCore/issues/21772#issuecomment-438279657 be merged? I've checked the NPCs, they don't turn when right-clicked, despawn properly and I wasn't able to replicate the bugs mentioned in the issue (having five ghoul drools properly disables spellclick).

Killyana:
I need to check sniffs, because in older core it was working properly.

PolarCookie:
I don't remember exactly at hand, but the NPC's turning for "Rifle the Bodies" was caused by https://github.com/TrinityCore/TrinityCore/issues/21772#issuecomment-417919623, which was (most likely) incorrect scripting.

jackpoz:
Thanks for the PR :)

RR2739-VCN:
could this be a reason why all of a sudden Battleground Spirit Healers are not targeting players when they die? I can be wrong but now BG Spirit Healers are not targeting players

Wyrserth:
@RR2739-VCN can you specific which npc gives issues? I've tried with WSG ones and couldn't find any issue with them.

RR2739-VCN:
I tried AB and EOTS. 99% of the time if you do not target a spirit healer it wont resurrect you.

Wyrserth:
Can't reproduce, tried in both BGs with two logged players, I get ressed as long as the "Resurrection in x Seconds" prompt is active, which is blizzlike behavior.

Could you make a video of your issue?

jackpoz:
and maybe try to reproduce it at http://tcubuntu.northeurope.cloudapp.azure.com too while at it, just to be sure it's not because of any custom change

RR2739-VCN:
I am testing this the following way:

EOTS | Die at a base that is now controlled by opposing faction | You are now stuck at the base, cannot target Spirit Healer obviously because it's controlled by another faction, so instead of being teleported to another base, you are stuck running to it.

Killyana:
This is a different issue, it's an old one where the player is not teleport to an other graveyard when the one where he respawn get controlled by the opposite faction.

RR2739-VCN:
I also did the test in WSG, 20 players, Regular Battleground, still the same issue, you do not get the Resurrection Popup unless you target the NPC, this happens non stop sometimes, mostly happens in the middle of battleground. That is why i thought this is related because you are not being targeted for resurrection, I cannot say which revision broke the Spirit Healers even more, but the revision we were using before 49a9cbf20a483da006e45f2ff953583c75c689f0 did not have this issue.

Issue #18373, opened 2016-12-01
[3.3.5a] Core/Spell: Focus is not cleared if the cast is interrupted


Tonghost:
Rel. #16572
190fdfd

![1](https://cloud.githubusercontent.com/assets/13364438/20805503/9c999364-b7ff-11e6-90b0-3da2cb901063.jpg)
![2](https://cloud.githubusercontent.com/assets/13364438/20805500/99f5bcf0-b7ff-11e6-9c8f-877bf465daca.jpg)
![3](https://cloud.githubusercontent.com/assets/13364438/20805502/9b2d89f4-b7ff-11e6-9f98-52e48dec98f1.jpg)

tkrokli:
Please include your `TrinityCore rev.` line from your Worldserver console, either line 1 or last startup line.

Tonghost:
@tkrokli 
190fdfd

tkrokli:
Right. Sorry, I didn't notice that you put it in your comment in the other issue. It is still recommended to put it in the original issue post, as well as using the issue template to describe the issue. I suppose you used images to explain it if your English isn't good enough to explain it properly. Anyway, testing the issue now.

tkrokli:
TC>TrinityCore rev. 190fdfdf63c0 2016-12-01 11:48:13 -0300 (3.3.5 branch) (Win64, Release, Static) / TDB 335.62

I can partially reproduce your issue with a random level 1 alliance character (Draenei male shaman). I tested both of the Sunreaver Guardian Mage guards, the male guard (on your right hand side as you try to enter) releases his focus after teleporting the alliance character onto the sewer cover / manhole cover every time.

The female guard (to your left as you try to enter the Sunreaver's Sanctuary), on the other hand, only releases her focus after the first attempt (or first 2), but then she seems to get her focus stuck on you.

You can also make the male guard face the wrong way after a row of attempts at entering Sunreaver's Sanctuary, but also make him turn back the right way after repeating the trespassing attempt some more.

Killyana:
This issue could also happen with many boss like Lana'thel, Saurfang, Sindragosa ect..., so it's not related to the AI.
The easiest way to reproduce the issue is: http://i.imgur.com/xOqPme9.jpg
The npc keep the focus event if the target is dead or on gm mode.

Killyana:
[...]

PR #23373, opened 2019-06-05
Script/Dalaran: don't allow Dalaran faction guards to teleport a player that has the Trespasser! debuff

Wyrserth:
**Changes proposed:**

The issue with the guards changing orientation is caused by the fact that the teleport spell cast is triggered two times for the same guard, it has nothing to do with spells being interrupted which was mentioned in #18373.

The spell focus system works by storing the original orientation, casting the spell, and restoring the orientation.

If the cast is triggered twice (possibly due to a delay in near teleport, in this case), the creature first casts the spell, stores the original orientation, but then before it can turn back to the original orientation it is already casting the second spell - which overrides the original orientation with the orientation it is facing right now (usually toward the area you tried to access).

Let's say the NPC has orientation=0. It turns to orientation=3 to cast a spell to a player, and saves the original orientation 0 somewhere to recover it later. But now it casts the spell again: orientation is still 3, but that stored 0 is overwritten by the new "original" orientation, a whopping 3.

Now, I believe this behavior is intended - it would be very ugly if a creature restored its original orientation every time before being allowed to cast a spell again.

Regardless of that, the issue is actually with the script - in no case a guard should teleport you when you already have the "teleported" aura.

- [x] 3.3.5

**Issues addressed:** Related to #18373, still have to check the other issues listed.

### DIFF
--- a/src/server/scripts/Northrend/zone_dalaran.cpp
+++ b/src/server/scripts/Northrend/zone_dalaran.cpp
@@ -89,7 +89,9 @@ public:
             if (!player || player->IsGameMaster() || player->IsBeingTeleported() ||
                 // If player has Disguise aura for quest A Meeting With The Magister or An Audience With The Arcanist, do not teleport it away but let it pass
                 player->HasAura(SPELL_SUNREAVER_DISGUISE_FEMALE) || player->HasAura(SPELL_SUNREAVER_DISGUISE_MALE) ||
-                player->HasAura(SPELL_SILVER_COVENANT_DISGUISE_FEMALE) || player->HasAura(SPELL_SILVER_COVENANT_DISGUISE_MALE))
+                player->HasAura(SPELL_SILVER_COVENANT_DISGUISE_FEMALE) || player->HasAura(SPELL_SILVER_COVENANT_DISGUISE_MALE) ||
+                // If player has already been teleported, don't try to teleport again
+                player->HasAura(SPELL_TRESPASSER_A) || player->HasAura(SPELL_TRESPASSER_H))
                 return;
 
             switch (me->GetEntry())


Takenbacon:
Using guard orientation is so silly, the script needs to be reworked to use areaid's instead.

Shauren:
AreaId won't work - its updated after teleport _ACK

jackpoz:
Thanks for the PR :)

ccrs:
How nice, finally "the mistery" is resolved, nice work.
Though this brings up yet another issue on focus system. Should previous orientation be preserved?

xvwyh:
@Takenbacon actually it should work via GO traps:
191372 - Dalaran Alliance Guard Teleport Trap (casts spell 54126)
191373 - Dalaran Horde Guard Teleport Trap (cast spell 54129)
The problem is that spawns of these traps presumably cannot be sniffed because, I guess, they have invisibility applied to them, invisibility that players cannot detect. So even if we were to use them - they would have to be spawned by hand. It wasn't until far beyond Cata that we even knew about these GO's existence, because if they cannot be seen - they wouldn't ever be stored in client's cache. Don't know if these GO records were even added to 3.3.5 TDB yet, but you can find them in master-branch's DB dumps.

Aokromes:
"It wasn't until far beyond Cata that we even knew about these GO's existence, because if they cannot be seen - they wouldn't ever be stored in client's cache. Don't know if these GO records were even added to 3.3.5 TDB yet, but you can find them in master-branch's DB dumps."

We known about spell focus since WOTLK.

Killyana:
I did an initial work here https://github.com/TrinityCore/TrinityCore/issues/23381

xvwyh:
@Aokromes, I don't know about what spellfocus you're talking. We did have many "generic" ones in WotLK (e.g. cooking-enabling foci), but the vast majority of quest-related foci weren't present in caches. Those ones that had IDs 300000+ were custom-made foci based on the data from SpellFocusObject.dbc, the rest - radius and other properties - were filled by hand. Only after lots more GO records were obtained from newer expansions did TC begin the process of removing 300000+ spawns and replacing them with proper foci, but spawns themselves remain non-sniffable.

Takenbacon:
@xvwyh Welcome to emulation, there's plenty of stuff that we flat out cannot do the same way blizzard does. The way things "should" work is usually irrelevant.

xvwyh:
@Takenbacon Welcome to programming, where stuff that we cannot do can eventually be done with enough effort and information. I really dislike your kind of attitude among emu developers, if we can't do something *right now*, doesn't mean we shouldn't strive to it, or try to get as close as we can. We couldn't implement these guards the way they're meant to work in past, because we couldn't have known about these traps' existence, but we do now, and that's a good enough reason to rework and remove old hacks. But everyone has their own goals, to some it's "get shit done", and to some it's "get shit done properly".

Takenbacon:
@xvwyh WoW emulation would still be sandboxes if the goal was to do everything properly and blizzlike. Pick your battles.

PR #23377, opened 2019-06-06
Core/SAI: Reimplement EVENT_FRIENDLY_HEALTH_PCT

### LINKED ISSUE #20536: Core/SAI: SMART_EVENT_FRIENDLY_HEALTH_PCT is broken

jackpoz:
This is the continuation of https://github.com/TrinityCore/TrinityCore/pull/20694 which cannot be modified anymore because the author deleted the branch in his fork.

**Changes proposed:**

EVENT_FRIENDLY_HEALTH_PCT is fundamentally broken, as any ACTION that uses a target will not work with it. It attempts to implement an action into an event, which doesn't work.

This change will reimplement it to be consistent with other events (like EVENT_FRIENDLY_HEALTH) and will allow the use of TARGET_ACTION_INVOKER, which is also consistent with other events.

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #20536

None

- [ ] All current SAI using this event is broken anway, but to fix it, all it needs is the target type changed to TARGET_ACTION_INVOKER, as the reimplementation is otherwise compatible with existing fields.
- [ ] The range should not be hardcoded but moved to SAI
- [ ] The existing scripts need to set a range
- [ ] DoSelectLowestHpFriendly(..) returns a single Unit with the lowest hp but then the health % is compared to minHpPct. This means if there is an event with minHpPct set to 50% and 2 allies, one with 20% and one with 60%, the event will not be triggered. 
- [x] --Same goes if there are 2 allies and the lowest hp one is not in combat due to the IsInCombat() check.-- The searcher already checks for IsInCombat()
- [ ] Another broken thing: the parameters were copied from EVENT_FRIENDLY_HEALTH_PCT with min/max but then the implementation of EVENT_FRIENDLY_HEALTH was left (which uses only 1 parameter validation as HP deficit). This means that DoSelectLowestHpFriendly() will always find the lowest hp unit around, even if we set max % to 10

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -3285,40 +3285,20 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
             if (!me || !me->IsEngaged())
                 return;
 
-            ObjectVector targets;
-            switch (e.GetTargetType())
+            // Get any friendly creature missing health within 30 yard radius
+            Unit* target = DoSelectLowestHpFriendly(30.0f, 1);
+            if (!target || !target->IsInCombat())
             {
-                case SMART_TARGET_CREATURE_RANGE:
-                case SMART_TARGET_CREATURE_GUID:
-                case SMART_TARGET_CREATURE_DISTANCE:
-                case SMART_TARGET_CLOSEST_CREATURE:
-                case SMART_TARGET_CLOSEST_PLAYER:
-                case SMART_TARGET_PLAYER_RANGE:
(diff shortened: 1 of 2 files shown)


jackpoz:
Note to myself: check how EVENT_FRIENDLY_HEALTH works and try to make these 2 events as similar as possible

jackpoz:
Actually the PR has just as many issues as the ones it tries to fix, I will just close it as rejected.

jackpoz:
If anyone needs this event to work, please handle the issues described here and in the PR that this one is based on.

PR #23396, opened 2019-06-13
Core/Session: don't remove players from party groups when they log out

Wyrserth:
**Changes proposed:**

From the 3.3.0 [undocumented changes](https://wowwiki.fandom.com/wiki/Patch_3.3.0_(undocumented_changes)):

- You no longer leave your party upon logging out.

I vaguely recall seeing this change in-game too back then, and I'm quite sure there was a timer that would give party leader to someone else if they logged out, but I don't have data to back that up.

**Target branch(es):** 3.3.5

- [x] 3.3.5

### DIFF
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -548,11 +548,6 @@ void WorldSession::LogoutPlayer(bool save)
         ///- If the player is in a group (or invited), remove him. If the group if then only 1 person, disband the group.
         _player->UninviteFromGroup();
 
-        // remove player from the group if he is:
-        // a) in group; b) not in raid group; c) logging out normally (not being kicked or disconnected)
-        if (_player->GetGroup() && !_player->GetGroup()->isRaidGroup() && m_Socket)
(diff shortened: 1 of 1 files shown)


Aokromes:
one point Kilyana pointed xd since if we merge that pr and leader disconects and doesn't return party will be needed to be remade how about implement 1st the if leader is offline for 2 minutes change to another member f1st? (on separate PR)

Shauren:
Wait, is automatic changing of group leader not implemented at all? (for Group code it makes no difference if its a raid or party)

Aokromes:
no it's not xd

Wyrserth:
Does anyone know what the behavior should be when every player in party logs out, or when the last online player leaves a group where everyone else is offline? I think it should be disbanded, but I'm not sure.

Aokromes:
it's not *always* automatically disbanded, i have seen PTR server crashes and party being still there at relog and others fully disbanded xd also, never checked the time to disband after all members logged out and relogged.

Wyrserth:
Superseded by #23398, closing this.

PR #23400, opened 2019-06-13
Core/Player: allow AutoStoreLoot to specify if the looted item shouldbe shown as if created by the player

Wyrserth:
**Changes proposed:**

In some cases we might want the player to receive loot stored in spell_loot_template as "You created [item]" instead of "You looted [item]", for example in Northrend Inscription Research and other similar profession-related spells.

**Target branch(es):** 3.3.5

- [x] 3.3.5

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -24590,7 +24590,7 @@ bool Player::IsBaseRuneSlotsOnCooldown(RuneType runeType) const
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


jackpoz:
Northrend Inscription Research bit is not implemented yet though, right ? this is jump in preparation of that ?

Wyrserth:
Not here, I found a few other bugs while fixing that one so I thought I should make small-sized PRs for each separate issue instead of cramping it all together. This PR only adds the functionality to fix this cosmetic bug.

Shauren (approved):
This is fine

PR #23401, opened 2019-06-14
Core/Spell: allow Northrend Inscription Research to unlock three recipes the first time it is cast

Wyrserth:
**Changes proposed:**

From the [3.1.0 patch notes](https://wowwiki.fandom.com/wiki/Patch_3.1.0):

> Players will now learn 3 recipes the very first time they perform Northrend Inscription Research. This does not apply to players who have already discovered recipes from Northrend Inscription Research (sorry). 

Also don't call UpdateCraftSkill() in the script - it's already called elsewhere and results in double skill points awarded.

**Target branch(es):** 3.3.5

- [x] 3.3.5

### DIFF
--- a/src/server/scripts/Spells/spell_generic.cpp
+++ b/src/server/scripts/Spells/spell_generic.cpp
@@ -2820,6 +2820,11 @@ class spell_gen_pet_summoned : public SpellScript
     }
 };
 
+enum ProfessionResearch
+{
+    SPELL_NORTHREND_INSCRIPTION_RESEARCH = 61177
+};
+
 class spell_gen_profession_research : public SpellScript
 {
     PrepareSpellScript(spell_gen_profession_research);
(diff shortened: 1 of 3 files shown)


src/server/scripts/Spells/spell_generic.cpp
```diff
@@ -2845,11 +2852,15 @@ class spell_gen_profession_research : public SpellScript
...
+        // Players will now learn 3 recipes the very first time they perform Northrend Inscription Research (3.3.0 patch notes)
+        if (spellId == SPELL_NORTHREND_INSCRIPTION_RESEARCH && !HasDiscoveredAnySpell(spellId, caster))
+            for (int i = 0; i < 2; ++i)
+                if (uint32 discoveredSpellId = GetExplicitDiscoverySpell(spellId, caster))
+                    caster->LearnSpell(discoveredSpellId, false);
+
         if (uint32 discoveredSpellId = GetExplicitDiscoverySpell(spellId, caster))
             caster->LearnSpell(discoveredSpellId, false);
-
-        caster->UpdateCraftSkill(spellId);
```
Keader: why removed?

Wyrserth: Because it's already called in Spell::EffectCreateItem2(), so leaving this in the script will give skill points twice.

Keader: ah is handled in other place too
Cool :)

jackpoz:
Thanks for the PR :)

PR #23404, opened 2019-06-14
Core/Spells: don't allow spells with SPELL_EFFECT_CREATE_ITEM_2 to be cast if there isn't enough space in inventory

Wyrserth:
**Changes proposed:**

SPELL_EFFECT_CREATE_ITEM_2 is used by profession recipes that should give a random item picked from a pool.

Right now, for the recipes that use a dummy item to show flavor text on mouseover like [Northrend Inscription Research](https://wowwiki.fandom.com/wiki/Northrend_Inscription_Research) do not check for inventory space, so players can craft them and obtain no item because their inventories were full.

Because those spells can award a random item of a random quantity, we have to make sure the player has at least one free bag/backpack space before being allowed to cast it.

This PR also fixes a cosmetic issue: items created in this way must show the chat message "You created [item]" instead of "You received loot [item]".

- [x] 3.3.5

### DIFF
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -1033,6 +1033,7 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         Item* GetItemByPos(uint8 bag, uint8 slot) const;
         Item* GetUseableItemByPos(uint8 bag, uint8 slot) const;
         Bag*  GetBagByPos(uint8 slot) const;
+        uint32 GetFreeInventorySpace() const;
         Item* GetWeaponForAttack(WeaponAttackType attackType, bool useable = false) const;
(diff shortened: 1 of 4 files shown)


src/server/game/Entities/Player/Player.h
```diff
@@ -2148,6 +2148,8 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
 
         std::string GetDebugInfo() const override;
 
+        uint32 GetFreeInventorySpace() const;
```
ccrs: Instead of at the end, try to place it near similar/related methods please

Wyrserth: Done, should be fine now.

jackpoz:
Thanks for the PR :)

Issue #22980, opened 2019-01-23
Trial Of The Crusader boss kill status


qyh214:
**Description:** 

kill boss and loot,Pickup status( It seems can loot again) is still displayed, but it is ineffective.

all bosses in Trial Of The Crusader have this issue.

AND I know The body needs to be kept on display after killed by player.

**Current behaviour:**
like Description

**Expected behaviour:**

After loot,Should no longer display lootable status.

**Steps to reproduce the problem:**

1. enter in Trial Of The Crusader
2. kill boss ,loot

**Branch(es):**

3.3.5, maybe both

**TC rev. hash/commit:** 

71fd5f19dc99acec6fdeb9decfc4a241134f68d2

**Operating system:** Win10 x64

Killyana:
Same issue as https://github.com/TrinityCore/TrinityCore/issues/12086
The boss will drop Alliance and horde items but players will be able to see only items related to their faction. If you activate cross faction you can make mixed party and you will be able to loot everything.

illfated:
I presume you are referring to the conf setting `AllowTwoSide.Interaction.Group` - or do we need to check other cross-faction settings too when this happens?

Aokromes:
a1b2aa9

PR #23408, opened 2019-06-14
Core/Loot: make Loot::AddItem() honor LootItem::AllowedForPlayer()

### LINKED ISSUE #13940: [3.3.5] Core/Entities: Sometimes unable to loot skinnable mobs
Hash: https://github.com/TrinityCore/TrinityCore/commit/9bd7b0e5714475d05d74c15bad500e0a59768edd

This is easily noticeable with creatures that can be skinned.
- Get Skinning
- Kill creatures
- Atleast one in every hundred will be both unlootable and impossible to skin (will say that the corpse has not been looted).

Point is, sometimes mobs drop loot that is not lootable. No sparkles around the corpse, but unable to skin them (which means that loot is actually there).

Also happens with creatures that always drop a certain item: even if the drop chance is 100%, sometimes one of the creatures will be unlootable.

Seems that every mob in game is affected by this.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7878072-3-3-5a-sometimes-unable-to-loot-mobs?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Wyrserth:
**Changes proposed:**

Currently Loot::AddItem() will increase unlootedCount without checking LootItem::AllowedForPlayer's result first. This means that items that cannot be looted by any player can be added to the loot. Because nobody can loot them, they're never removed and the creature will show an empty loot window and can't be skinned as a result.

However, in some cases the dropped item is not visible by some players (because they're not allowed to loot it) but should start a roll if any other player in party/raid can loot it. This case is also fixed in this PR.

Fixes #13940.

**Easy test case:**
WARNING: this empties the loot table of one creature. Don't run it if you care about your world db
```sql
DELETE FROM `creature_loot_template` WHERE `Entry`=2958;
INSERT INTO `creature_loot_template` (`Entry`, `Item`, `Reference`, `Chance`, `QuestRequired`, `LootMode`, `GroupId`, `MinCount`, `MaxCount`, `Comment`) VALUES
(2958, 44564, 0, 100, 0, 1, 0, 1, 1, "");
```
The creature (Prairie Wolf - 2958) will drop an Alchemy-only recipe that will prevent skinning without this PR.

- [x] 3.3.5

**Known issues**: if a player without Alchemy 400+ (in the example case) exits the loot window before the player with Alchemy 400+, the creature will not be lootable again and the recipe will be unlootable. I don't know what's the best way to fix this, and it's a global issue with the loot system and not only with the part changed by this PR.

### DIFF
--- a/src/server/game/Loot/Loot.cpp
+++ b/src/server/game/Loot/Loot.cpp
@@ -144,6 +144,24 @@ void Loot::AddItem(LootStoreItem const& item)
         lootItems.push_back(generatedLoot);
         count -= proto->GetMaxStackSize();
 
+        // In some cases, a dropped item should be visible/lootable only for some players in group
+        bool canSeeItemInLootWindow = false;
+        if (Player* player = ObjectAccessor::FindPlayer(lootOwnerGUID))
+        {
+            if (Group* group = player->GetGroup())
+            {
+                for (GroupReference* itr = group->GetFirstMember(); itr != nullptr; itr = itr->next())
+                    if (Player* member = itr->GetSource())
+                        if (generatedLoot.AllowedForPlayer(member))
+                            canSeeItemInLootWindow = true;
+            }
+            else if (generatedLoot.AllowedForPlayer(player))
+                canSeeItemInLootWindow = true;
+        }
+
+        if (!canSeeItemInLootWindow)
+            continue;
+
         // non-conditional one-player only items are counted here,
         // free for all items are counted in FillFFALoot(),
         // non-ffa conditionals are counted in FillNonQuestNonFFAConditionalLoot()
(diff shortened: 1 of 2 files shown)


Killyana:
Fixes https://github.com/TrinityCore/TrinityCore/issues/22980

Wyrserth:
This needs to check if any party or raid member should be able to loot the item too - pushing in a moment.

jackpoz:
> WARNING: this empties the loot table of one creature. Don't run it if you care about your world db

thank God we have a monthly TDB 335 now so it's easy to get started again :D

jackpoz:
Does this take into account Personal Loot case too ? did the "known issue" exist before this PR already ?

Killyana:
Yes the "known issue" is an old one, for personal loot it works fine.

jackpoz:
Thanks for the PR :)

PR #23413, opened 2019-06-15
Core/Loot: implement automatic pass in group loot for items that cannot be looted

Wyrserth:
**Changes proposed:**

When the party/raid loot settings are set to "group loot" and a dropped item cannot be looted by some players (because it's an unique item and they can't store any more of those, or because it's a soulbound recipe they already learned/they lack the profession and skill for), they should receive a "You automatically passed on: [item] because you cannot loot that item." chat message and automatically pass the loot.

Also don't start the loot if all the players automatically passed because of unlootable item or because of autopass setting being turned on, prevents loot rolls being stuck until timeout.

- [x] 3.3.5


jackpoz:
Thanks for the PR :)

PR #23415, opened 2019-06-15
Core/Creatures: fix some other cases where a creature should not be able to set focus

Wyrserth:
**Changes proposed:**

Creatures should not be able to set focus while dead, feigning death or if they're stunned, mostly in case of permanent cosmetic auras related to quests.

- [x] 3.3.5


Killyana:
Fixes some cases like https://github.com/TrinityCore/TrinityCore/issues/21772

jackpoz:
Thanks for the PR :)

PR #23416, opened 2019-06-15
Core/Commands: implement commands to look up a map/item/quest name for a given id

Wyrserth:
**Changes proposed:**

It's useful to be able to get a map/item/quest name from the related map/item/quest id.

(while editing auth_database.sql I changed the speeds of some old updates to 0 for consistency, let me know if it's an issue)

- [x] 3.3.5

**Tests performed:** it works (both from console and from in-game).


jackpoz:
Thanks for the PR :)

PR #23420, opened 2019-06-15
Core/Creature: drop current target when evading while a spell cast is in progress

### LINKED ISSUE #18373: [3.3.5a] Core/Spell: Focus is not cleared if the cast is interrupted

Wyrserth:
**Changes proposed:**

Currently a creature's target is not properly cleared when they exit combat in the middle of casting (.gm on). This PR fixes that so creatures will properly drop their target when they should.

- [x] 3.3.5

**Issues addressed:** closes #18373


jackpoz:
Thanks for the PR :)

Issue #23419, opened 2019-06-15
DB/Quest: fix end event for On Brann's Trail and fix prevquest for Sniffing Out the Perpetrator


Wyrserth:
**Description:**

Quest [On Brann's Trail](https://www.wowhead.com/quest=12854) is missing a scripted end event. Also the quest [Sniffing Out the Perpetrator](https://www.wowhead.com/quest=12855) should become available once [On Brann's Trail](https://www.wowhead.com/quest=12854) is rewarded.

Closes #22647.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 5c019db6a7d91e10eb074e8776620527796a3b46

**Suggested fix:**
```sql
UPDATE `quest_template_addon` SET `PrevQuestID`=12854 WHERE `ID`=12855;

UPDATE `creature_template` SET `AIName`="SmartAI" WHERE `entry`=29650;
DELETE FROM `smart_scripts` WHERE `entryorguid` IN (29650, 2965000) AND `source_type` IN (0, 9);
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(29650, 0, 0, 0, 20, 0, 100, 0, 12854, 0, 0, 0, 0, 80, 2965000, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, "Archaeologist Andorin - On Quest 'On Brann's Trail' Rewarded - Run Script"),
(2965000, 9, 0, 0, 0, 0, 100, 0, 500, 500, 0, 0, 0, 66, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, "Archaeologist Andorin - On Script - Face Invoker"),
(2965000, 9, 1, 0, 0, 0, 100, 0, 500, 500, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, "Archaeologist Andorin - On Script - Talk 0"),
[...]

PR #23421, opened 2019-06-15
Core/SAI: don't allow to start a new SAI actionlist while the entity is already running one

Wyrserth:
**Changes proposed:**

When an actionlist is running, we assume that it will properly finish (unless the creature is killed mid-script). Right now, especially for events that happen when a quest is turned in, other players can turn in the same quest and cause the current actionlist to be dropped and start a new one.

For example, apply the SQL in #23419 (sorry, can't think of a good example with a reasonably long script right now) and have multiple players turn in the quest [On Brann's Trail](https://www.wowhead.com/quest=12854). The script will restart and the NPC will say the first lines of creature_text again.

- [x] 3.3.5

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -3801,6 +3801,10 @@ void SmartScript::SetTimedActionList(SmartScriptHolder& e, uint32 entry, Unit* i
         return;
     }
 
+    // Do NOT allow to start a new actionlist if a previous one is already running. We need to always finish the current actionlist
+    if (!mTimedActionList.empty())
+        return;
+
     mTimedActionList.clear();
     mTimedActionList = sSmartScriptMgr->GetScript(entry, SMART_SCRIPT_TYPE_TIMED_ACTIONLIST);
     if (mTimedActionList.empty())


ZenoX92:
And what if an anctionlist is meant to be interrupted by another one?

Let's say there's an NPC who runs an actionlist during a quest event, and this actionlist ends with the NPC dying and you failing the quest. And the only way you can complete this quest is making the NPC to interrupt the actionlist by running a second one that will end with you getting the quest credit. How would you make this to happen with your proposed changes?

There are other ways to prevent the issue you are talking about from happening: removing NPC flags, using event phases, or this one you're working on https://github.com/TrinityCore/TrinityCore/pull/23422

I'm just saying, if your pull is finally merged, I think there should be a way to drop an actionlist in favor of another one. Maybe adding an extra action parameter (like the interrupt cast flag for spells).

ZenoX92:
What do you think about this @Killyana?

Killyana:
An action_list mustn't be override by an other, the example you provided, is just bad concept as we have SMART_ACTION_CALL_SCRIPT_RESET that allow us to interrupt an action list.
If there's some scripts that use this bad concept they must be fixed.

jackpoz:
Thanks for the PR :)

PR #23422, opened 2019-06-16
Core/SAI: add cooldownMin (event_param2) and cooldownMax (event_param3) to SMART_EVENT_ACCEPTED_QUEST and SMART_EVENT_REWARD_QUEST

### LINKED ISSUE #23342: Core/SAI: Repeat Min/Max for SMART_EVENT_ACCEPTED_QUEST & SMART_EVENT_REWARD_QUEST
**Description:** We need  RepeatMin | RepeatMax for both SMART_EVENT_ACCEPTED_QUEST & SMART_EVENT_REWARD_QUEST.

This will avoid having the same event triggering when there's already one in going.

**Branch(es):** Both

**TC rev. hash/commit:**  rev. 2d5387ea425f

Wyrserth:
**Changes proposed:**

SMART_EVENT_ACCEPTED_QUEST and SMART_EVENT_REWARD_QUEST need cooldown timers in order to control how often the script should execute, as to not run improperly when multiple players turn in the quest at the same time.

- [x] 3.3.5

**Issues addressed:** closes #23342

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -3188,6 +3188,7 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
         {
             if (e.event.quest.quest && var0 != e.event.quest.quest)
                 return;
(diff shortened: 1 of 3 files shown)


jackpoz:
Thanks for the PR :)

PR #23428, opened 2019-06-16
Core/FlightPath: teleport players at destination on floor Z when the flight ends

### LINKED ISSUE #10051: [Core/Taxi] Incorrect behavior at path end
I stumbled over this issue when working on a quest. http://www.wowhead.com/spell=46814 is related. 
When the flight is performed and the player reaches the last node from TaxiPathNode.dbc the flight ends and the player falls to the ground. Especially on this path the above mentioned node and the endnode 237 in TaxiNodes.dbc for this path have quite different coordinates which makes the problem quite obvious.
Instead of falling to the ground the player is supposed to teleport to the coordinates of the endnode. 

You can see this here at about 4:33 https://www.youtube.com/watch?v=XHi9Z15QE6Q.
Keep an eye on the minimap and notice how the position changes at the moment the drake lands.

hash: d163c58258160b65a2d785ef552653fc2c8f5eb1
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1413135-core-taxi-incorrect-behavior-at-path-end?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1310&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Wyrserth:
**Changes proposed:**

Right now, when a player ends a flight path, they dismount in their actual position, which is usually a few yards above the ground. Instead they should be teleported to the destination position, which makes it look better cosmetically AND prevents cheating that could occur by creating high lag before the flight ends, since the player will be forcefully teleported back in any case.

- [x] 3.3.5

**Issues addressed:** closes #10051

### DIFF
--- a/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
@@ -151,8 +151,9 @@ void FlightPathMovementGenerator::DoFinalize(Player* owner, bool active, bool/* 
         // when client side flight end early in comparison server side
         owner->StopMoving();
         owner->SetFallInformation(0, owner->GetPositionZ());
-        // When the player reaches the last flight point, teleport to destination at floor Z
-        owner->TeleportTo(_path[GetCurrentNode()]->MapID, _path[GetCurrentNode()]->LocX, _path[GetCurrentNode()]->LocY, owner->GetFloorZ(), owner->GetOrientation());
+        // When the player reaches the last flight point, teleport to destination at map height
+        float mapHeight = owner->GetMap()->GetHeight(_path[GetCurrentNode()]->LocX, _path[GetCurrentNode()]->LocY, _path[GetCurrentNode()]->LocZ);
+        owner->TeleportTo(_path[GetCurrentNode()]->MapID, _path[GetCurrentNode()]->LocX, _path[GetCurrentNode()]->LocY, mapHeight, owner->GetOrientation());
     }
 
     owner->RemoveFlag(PLAYER_FLAGS, PLAYER_FLAGS_TAXI_BENCHMARK);
--- a/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
@@ -150,9 +150,9 @@ void FlightPathMovementGenerator::DoFinalize(Player* owner, bool active, bool/* 
         // this prevent cheating with landing  point at lags
         // when client side flight end early in comparison server side
         owner->StopMoving();
-        owner->SetFallInformation(0, owner->GetPositionZ());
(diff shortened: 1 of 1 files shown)


src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
```diff
@@ -151,6 +151,8 @@ void FlightPathMovementGenerator::DoFinalize(Player* owner, bool active, bool/*
         // when client side flight end early in comparison server side
         owner->StopMoving();
         owner->SetFallInformation(0, owner->GetPositionZ());
+        // When the player reaches the last flight point, teleport to destination at floor Z
+        owner->TeleportTo(_path[GetCurrentNode()]->MapID, _path[GetCurrentNode()]->LocX, _path[GetCurrentNode()]->LocY, owner->GetFloorZ(), owner->GetOrientation());
```
jackpoz: Not sure I like the fact that we use X and Y of the path but Floor Z of the Player (which could be different than Floor Z of the path)

Ovahlord: Also I would ask for a sniff validation for this behaivior because for me this smells like a hack tbh.

Wyrserth: All the cases I have seen (videos) cause a teleport exactly on the ground, so I went with it as it made sense to me.

Could use the endpoint's FloorZ instead, they should be the same.

Ovahlord: Videos do not necessarily validate this solution. You need to search for a teleport packet inside a sniff for the behaivior. If it does not exist, the problem is located somewhere else which is probably going to be the flight path spline movement.

Wyrserth: Confirmed with sniffs, client sends MSG_MOVE_TELEPORT_ACK at the end of the flight. The endpoint serverside is higher on the ground, but teleports instantly instead of falling.

This is the flow:
```
Taxi creature finished moving to endpoint: SMSG_ON_MONSTER_MOVE: X: 2793.49 Y: 909.4843 Z: 24.54242 (Northrend)
Control is given to player: SMSG_CONTROL_UPDATE
Collision height is reset because the player dismounted: SMSG_MOVE_SET_COLLISION_HEIGHT
Teleport. MSG_MOVE_TELEPORT Destination: X: 2792.45 Y: 908.96 Z: 22.33 (FloorZ pos)
```

Wyrserth:
This can be merged, behavior was confirmed with sniffs.

jackpoz:
Thanks for the PR :)

PR #23429, opened 2019-06-16
Script/ZulGurub: rewrite High Priest Thekal's encounter script

### LINKED ISSUE #17605: [3.3.5] High priest thkal
High priest thkal dont change tiger form

https://github.com/TrinityCore/TrinityCore/commit/931a7623b24831adf679e0581735375077c8d252
ubuntu
3.3.5

Wyrserth:
**Changes proposed:**

This rewrite gets Thekal's encounter as close to blizzlike as I can manage.

First things first, I don't have any sniff to use as source so some things might not be 100% blizzlike.

- Fixed various cosmetic issues.
- Implemented the correct revival of adds (the 10-second timer starts when the first add is killed).
- Use Feign Death rather than actually kill the adds in the first phase. That's far easier to manage.
- Avoid any wrong case of evade when the creatures become invulnerable due to encounter mechanics (the original script breaks here, there's no second phase because the adds reset and Thekal dies in phase one).
- Added missing emotes.
- Fixed enrage mechanic - Thekal enrages in phase two only while below 10% health, not at the start of the fight. Also the creature size is handled by the enrage aura, so the script should never manually call SetObjectSize().

Closes #17605.

- [x] 3.3.5 (encounter removed in Cataclysm)

**Tests performed:** works fine, encounter can be properly completed. [Example video for comparison](https://www.youtube.com/watch?v=j3p92SSrRN4).

- [x] Zealot Lor'khan should cast her Greater Heal on one of the three creatures (herself, Zath and Thekal) that has the lowest health AND that is not currently dead. Right now it rolls a 50% chance between Zath and Thekal. Also it only casts the heal if her target is in melee target, which is wrong.

- [ ] I can't find any source on how the resurrection spell should be handled. The Zath and Lor'khan can be stunned and interrupted, but the resurrection should be handled regardless of that. So for the time being I changed the Resurrection spell to be instant cast.


offl:
Closes #17605?

Wyrserth:
Yeah it does, thanks!

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
```diff
@@ -85,56 +98,147 @@ class boss_thekal : public CreatureScript
 
         struct boss_thekalAI : public BossAI
         {
-            boss_thekalAI(Creature* creature) : BossAI(creature, DATA_THEKAL)
+            boss_thekalAI(Creature* creature) : BossAI(creature, DATA_THEKAL), _instance(creature->GetInstanceScript())
```
jackpoz: BossAI already stores instance script in "instance"variable https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp#L450

Wyrserth: Right, I think I realized that halfway through the script (using instance everywhere else) but forgot to remove this :P Thanks!

Wyrserth:
This is as good as it will be from me I believe, can't find any info on how resurrection should work (and the encounter was reworked in Cataclysm so any info beyond that is not applicable).

Even then, it is only a cosmetic detail that has no effect on the actual encounter.

I think this can be merged - it is a big improvement over the current TC's broken script.

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -4030,6 +4030,12 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->RecoveryTime = 20000;
     });
 
+    // Resurrection - Zul'Gurub, High Priest Thekal
+    ApplySpellFix({ 24173 }, [](SpellInfo* spellInfo)
+    {
+        spellInfo->CastTimeEntry = sSpellCastTimesStore.LookupEntry(1);
```
Keader: add a comment with how much time is it (just to make life easy when someone read it)

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
```diff
@@ -93,48 +108,139 @@ class boss_thekal : public CreatureScript
...
+                _isEvadeDisabled = false;
+            }
+
+            void EnterEvadeMode(EvadeReason why) override
+            {
+                if (_isEvadeDisabled)
+                    return;
+                if (!_EnterEvadeMode(why))
+                    return;
+                me->AddUnitState(UNIT_STATE_EVADE);
```
Keader: any reason for hardcode it?

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
```diff
@@ -93,48 +108,139 @@ class boss_thekal : public CreatureScript
...
+                    Creature* creature = nullptr;
+                    if (data == NPC_HIGH_PRIEST_THEKAL)
+                        creature = me;
+                    else if (data == NPC_ZEALOT_LORKHAN)
+                        creature = instance->GetCreature(DATA_LORKHAN);
+                    else if (data == NPC_ZEALOT_ZATH)
+                        creature = instance->GetCreature(DATA_ZATH);
+
+                    // Resurrect
+                    creature->RemoveAurasDueToSpell(SPELL_PERMANENT_FEIGN_DEATH);
```
Keader: missing check if creature pointer is valid?
Can happens if instance cant find the creature.

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
```diff
@@ -93,48 +108,139 @@ class boss_thekal : public CreatureScript
...
+                    Talk(TALK_FRENZY);
+                    _enraged = true;
+                }
+            }
 
+            void UpdateAI(uint32 diff) override
+            {
                 events.Update(diff);
 
-                if (me->HasUnitState(UNIT_STATE_CASTING))
```
Keader: why removed?

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
```diff
@@ -148,66 +254,62 @@ class boss_thekal : public CreatureScript
...
-                            events.ScheduleEvent(EVENT_FRENZY, 30s, 0, PHASE_TWO);
+                        }
+                        case EVENT_CHANGE_PHASE_1:
+                            Talk(TALK_TIGER_PHASE);
+                            events.ScheduleEvent(EVENT_CHANGE_PHASE_2, 1s);
                             break;
+                        case EVENT_CHANGE_PHASE_2:
+                        {
+                            // Trigger phase change
+                            DoCast(me, SPELL_TIGER_FORM);
```
Keader: change all "DoCast(me, xxxx);" to -> DoCastSelf(xxx);

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
```diff
@@ -270,147 +353,135 @@ class boss_thekal : public CreatureScript
...
-
-                    if (me->IsWithinMeleeRange(target))
-                        DoCast(target, SPELL_GREATERHEAL);
-
-                    _greaterHealTimer = 15000 + rand32() % 5000;
-                }
-                else
-                    _greaterHealTimer -= diff;
+                // Don't update AI if faking death
+                if (me->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_PC))
```
Keader: if is not updating AI, should not be above UpdateVictim() call?

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
Wyrserth: Added asserts, it's guaranteed that those creatures will exist at this point in the script unless something very bad is going on.

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
Wyrserth: Mistake on my part, thought he didn't have any spells with cast time :P

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
Wyrserth: Mistake in the comment, should be "don't update target". It avoids the creatures evading when they are immune to pc/npc and feigning death, but allows them to evade when the players wipe.

Wyrserth:
Fixed the issues and some more, thanks for your review!

src/server/scripts/EasternKingdoms/ZulGurub/boss_thekal.cpp
jackpoz: Do they exist at this point even if one calls ".npc delete" on them halfway through the fight ?

Wyrserth: It will crash if a GM deletes the creature after it begins feigning death and before it is resurrected.

Didn't think about harmful GMs, going to remove asserts and silently fail in that case. Thanks :P

Rushor:
on latest rev the boss resets infinite

PR #23434, opened 2019-06-18
Core/Packet: send SMSG_QUESTGIVER_STATUS_MULTIPLE when the player levels up

Wyrserth:
**Changes proposed:**

Make sure nearby questgivers are updated at levelup, so players can accept new quests that became available from a nearby questgiver without having to leave the area and come back.

- [x] 3.3.5

**How to reproduce:** (add/remove lines as needed)

- Create a Horde character
- .char level 77
- .npc add 31151 (will spawn a questgiver with gray !)
- .levelup (will bring you to 78, the minimum level required for this questgiver's quest)
- Notice that the ! is still gray and you can't obtain any quest, without this change
- .gm on
- .gm off
- The ! will become yellow because changing GM state will trigger an update on nearby questgivers, too.


jackpoz:
Thanks for the PR :)

PR #23437, opened 2019-06-18
Core/Misc: don't allow FindNearestCreature() and FindNearestGameObject() searches to return the entity that called them

### LINKED ISSUE #23334: Core/Conditions: The unit must exclude itself from CONDITION_NEAR_CREATURE & CONDITION_NEAR_GAMEOBJECT
**Description:** If you make a condition with CONDITION_NEAR_CREATURE or CONDITION_NEAR_GAMEOBJECT and use the same entry as the unit (ConditionTarget= Object) the unit will include it self in the check and it will always return true

1. .go c 107417
2. Notice the npc is doing some OOC combat movements
3. if you kill the npc in front of it he must stop doing this movement a there's a condition requires one of this npc near (23663 or 23664 or 23665 or 23666)  but as it has the same entry it will continue executing the SAI ooc
``` sql 
SELECT * FROM `conditions` WHERE `SourceTypeOrReferenceId`=22 AND `SourceEntry` IN (23663,23664,23665,23666);
```

**Branch(es):** both

**TC rev. hash/commit:** rev. 8bc545186457

Wyrserth:
**Changes proposed:**

When a creature/gameobject searches for a nearby creature/gameobject with a specific entry, it does not make sense to allow to return that same creature/gameobject in the search results.

- [x] 3.3.5

**Issues addressed:** closes #23334.

### DIFF
--- a/src/server/game/Grids/Notifiers/GridNotifiers.h
+++ b/src/server/game/Grids/Notifiers/GridNotifiers.h
@@ -740,7 +740,7 @@ namespace Trinity
 
             bool operator()(GameObject* go)
             {
-                if (go->GetEntry() == i_entry && i_obj.IsWithinDistInMap(go, i_range))
+                if (go->GetEntry() == i_entry && go->GetGUID() != i_obj.GetGUID() && i_obj.IsWithinDistInMap(go, i_range))
                 {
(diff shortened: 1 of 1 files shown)


jackpoz:
Thanks for the PR :)

PR #23438, opened 2019-06-18
Core/Conditions: implement CONDITION_QUEST_OBJECTIVE_PROGRESS to allow to apply conditions based on the player's quest objective progress

Wyrserth:
**Changes proposed:**

With this condition, you can create conditions that require the player to have a certain amount of creatures killed/gameobjects triggered for a specific objective. Because most quests use triggers in their RequiredNPCOrGo fields, this covers most quests.

Value1 is the quest entry, value2 is the objective index (a value from 0 to 3 depending on the field you want to control, from RequiredNPCOrGO1 to RequiredNPCOrGO4), value3 is the counter.

**Quick and dirty test case:**

WARNING: this will replace an existing NPC's script with testing data

```sql
DELETE FROM smart_scripts WHERE entryorguid=28566;
INSERT INTO smart_scripts VALUES
(28566, 0, 0, 0, 6, 0, 100, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, ""),
(28566, 0, 1, 0, 6, 0, 100, 0, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, ""),
(28566, 0, 2, 0, 6, 0, 100, 0, 0, 0, 0, 0, 0, 1, 2, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, ""),
(28566, 0, 3, 0, 6, 0, 100, 0, 0, 0, 0, 0, 0, 1, 3, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, "");

DELETE FROM conditions WHERE SourceEntry=28566;
INSERT INTO conditions VALUES
(22, 1, 28566, 0, 0, 48, 0, 12592, 0, 1, 0, 0, 0, "", ""),
(22, 2, 28566, 0, 0, 48, 0, 12592, 0, 2, 0, 0, 0, "", ""),
(22, 3, 28566, 0, 0, 48, 0, 12592, 0, 3, 0, 0, 0, "", ""),
(22, 4, 28566, 0, 0, 48, 0, 12592, 0, 4, 0, 0, 0, "", "");
```
.quest add 12592
.npc add 28566
Kill that creature over and over to get different creature texts when you reach 1, 2, 3 and 4 kills.

- [x] 3.3.5 (can't be applied to master due to different handling of quest objectives)

### DIFF
--- a/src/server/game/Conditions/ConditionMgr.cpp
+++ b/src/server/game/Conditions/ConditionMgr.cpp
@@ -2311,9 +2311,9 @@ bool ConditionMgr::isConditionTypeValid(Condition* cond) const
                 return false;
             }
 
-            if (cond->ConditionValue3 > quest->RequiredNpcOrGo[cond->ConditionValue2])
+            if (cond->ConditionValue3 > quest->RequiredNpcOrGoCount[cond->ConditionValue2])
             {
-                TC_LOG_ERROR("sql.sql", "%s has quest objective count %u in value3, but quest %u has a maximum objective count of %u in RequiredNPCOrGO%u, skipped.", cond->ToString(true).c_str(), cond->ConditionValue3, cond->ConditionValue2, quest->RequiredNpcOrGo[cond->ConditionValue2], cond->ConditionValue2);
+                TC_LOG_ERROR("sql.sql", "%s has quest objective count %u in value3, but quest %u has a maximum objective count of %u in RequiredNPCOrGOCount%u, skipped.", cond->ToString(true).c_str(), cond->ConditionValue3, cond->ConditionValue2, quest->RequiredNpcOrGoCount[cond->ConditionValue2], cond->ConditionValue2);
                 return false;
             }
             break;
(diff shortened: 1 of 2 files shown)


src/server/game/Conditions/ConditionMgr.cpp
```diff
@@ -2270,6 +2290,34 @@ bool ConditionMgr::isConditionTypeValid(Condition* cond) const
...
+        case CONDITION_QUEST_OBJECTIVE_PROGRESS:
+        {
+            const Quest* quest = sObjectMgr->GetQuestTemplate(cond->ConditionValue1);
+            if (!quest)
+            {
+                TC_LOG_ERROR("sql.sql", "%s points to non-existing quest (%u), skipped.", cond->ToString(true).c_str(), cond->ConditionValue1);
+                return false;
+            }
+
+            if (cond->ConditionValue2 < 0 || cond->ConditionValue2 > 3)
```
Wyrserth: Could change it to handle values between 1 and 4 instead of between 0 and 3. First one is less confusing based on quest_template fields, but second one is more correct because indexing always starts at 0 everywhere.

jackpoz:
Thanks for the PR :)

PR #23447, opened 2019-06-19
Core/Conditions: fix mistake in CONDITION_QUEST_OBJECTIVE_PROGRESS

Wyrserth:
**Changes proposed:**

Forgot to check if the player actually has the quest whose conditions are checked for, in #23438. It didn't cross my mind that this would be actually used with a progression value of 0, sorry :P

- [x] 3.3.5


jackpoz:
Thanks for the PR :)

PR #23448, opened 2019-06-19
Spell/Fishing: fix some Fishing bugs

Wyrserth:
**Changes proposed:**

This PR fixes two issues with Fishing:

- Right now you can spam the fishing spell and eventually get an instantly lootable fishing bobber: this is because the duration to wait before looting is randomly picked between static timeframes (3/7/13/17 seconds of channeling remaining). Because 17 was a valid choice, and the channeling spell lasts exactly 17 seconds, it would cause an instantly lootable bobber (blizzlike times are 3/7/13).

- When the channeling of Fishing ends without the player clicking on the bobber (can happen if you get a hook four seconds before the channeling ends) , it should despawn. Right now it stays for one more second because FISHING_BOBBER_READY_TIME (five seconds) is always added to the duration and can result in a 18-second duration (channel time is 17).
Casting fishing again in that one second causes a desync where a new fishing bobber is spawned, becomes unlootable and gives a "fish got away" even after the player walked away.

- [x] 3.3.5


jackpoz:
Thanks for the PR :)

PR #23449, opened 2019-06-19
Core/Loot: don't allow to regenerate chest loot inside instances and raids

Wyrserth:
**Changes proposed:**

Right now, loot regenerates inside any partially looted chest, based on spawntimesecs. This was implemented in 7011aabb56f061fc889ec42ce32b7605ceb35b53 and was supposed to help with chest behavior in the continent maps, but it also created an issue where loot regenerates inside chests that contain rewards from bosses.

The spawntimesecs is also the refill timer, however raids can be extended multiple times, so eventually those chests would refill.

- [x] 3.3.5


jackpoz:
Thanks for the PR :)

PR #23454, opened 2019-06-20
Scripts/Spells: Migrate More Scripted Spells to Scripts

Sorikoff:
**Changes proposed:**
Migrate even more spells to scripts system. This PR brings very minimal changes, since it is not the aim of this PR.

**Target branch(es):** 3.3.5

**Tests performed:** Tested each migrated spell.

### DIFF
--- a/src/server/scripts/Spells/spell_priest.cpp
+++ b/src/server/scripts/Spells/spell_priest.cpp
@@ -38,6 +38,7 @@ enum PriestSpells
     SPELL_PRIEST_GLYPH_OF_PRAYER_OF_HEALING_HEAL    = 56161,
     SPELL_PRIEST_GUARDIAN_SPIRIT_HEAL               = 48153,
     SPELL_PRIEST_ITEM_EFFICIENCY                    = 37595,
+    SPELL_PRIEST_LIGHTWELL_CHARGES                  = 59907,
     SPELL_PRIEST_MANA_LEECH_PROC                    = 34650,
     SPELL_PRIEST_PENANCE_R1                         = 47540,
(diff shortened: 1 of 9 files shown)


jackpoz:
As I had a bad experience with https://github.com/TrinityCore/TrinityCore/pull/22332 that caused https://github.com/TrinityCore/TrinityCore/issues/21435 (and I had to do https://github.com/TrinityCore/TrinityCore/pull/23228 ) , could you check that not only the spells don't crash worldserver but that every little detail about them still works ?

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -678,6 +679,43 @@ class spell_pri_item_t6_trinket : public SpellScriptLoader
...
 
+// 60123 - Lightwell
+class spell_pri_lightwell : public SpellScript
+{
+    PrepareSpellScript(spell_pri_lightwell);
+
+    void HandleScriptEffect(SpellEffIndex /* effIndex */)
+    {
+        Unit* caster = GetCaster();
+        if (!caster->ToCreature()->IsSummon())
```
jackpoz: please don't assume that caster is a Creature

jackpoz:
Actually, thinking about it, this PR changes 8 spells which are 7 too many for a PR.
Please open a PR moving only 1 spell at a time so that it can be clearly reviewed ( see for example https://github.com/TrinityCore/TrinityCore/pull/23454#discussion_r295807680 , it's a pain to have to scroll up and time to compare which script is related to which spell and see if the code is the same)

Sorikoff:
Yeah, this makes sense. Will make individual PRs.

PR #23455, opened 2019-06-20
Scripts/Spells: Migrate Lightwell to Script

Sorikoff:
**Changes proposed:**
Migrate Lighwell to scripts system.

I have noticed that Lighwell heals caster only once (object becomes inaccessible after the first usage), but the same behaviour has been present before this change.

**Target branch(es):** 3.3.5

**Tests performed:** Works as it did.


sql/updates/world/3.3.5/9999_99_99_99_world.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM `spell_script_names` WHERE `ScriptName` IN ('spell_pri_lightwell');
+INSERT INTO `spell_script_names` VALUES
+(60123,'spell_pri_lightwell');
```
Aokromes: missing newline.

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -678,6 +680,48 @@ class spell_pri_item_t6_trinket : public SpellScriptLoader
...
+    void HandleScriptEffect(SpellEffIndex /* effIndex */)
+    {
+        Creature* caster = GetCaster()->ToCreature();
+        if (!caster || !caster->IsSummon())
+            return;
+
+        uint32 spellId = 0;
+        switch (caster->GetEntry())
+        {
+            case 31897: spellId = 7001; break;
```
Keader: why not transform those magic numbers in enum?

Sorikoff: They all indicate the same (Lightwell/Lightwell Renew), and the only difference is level. I am unsure if it is needed to enum each Lightwell1, Lightwell2, and so on.

Faq: Lightwell_LightwellRenew = 7001 ?

Sorikoff: Updated PR. Thanks!

jackpoz:
Thanks for the PR :)

PR #23459, opened 2019-06-20
Core/RBAC: add permission that prevents accounts from earning all achievements or only realm firsts

Wyrserth:
**Changes proposed:**

This PR implements "Cannot earn achievements" and "Cannot earn realm first achievements" permissions:
- The first automatically includes the second - the account cannot earny ANY achievement.
- The second only affects realm firsts, and is used by default to prevent accounts with security 1/2/3 (moderator/gamemaster/admin) from earning realm-first achievements.

Because of RBAC permissions being inherited from lower ranked accounts (which is a good thing overall), the permissions are of the "deny" kind rather than "allow".

- [x] 3.3.5
- [x] master

**Issues addressed:** supersedes PR #23135.


illfated:
Thank you for giving me a chuckle when reading your commit title "Kill evil spaces"
-- it reminded me of this old thing:
my_foot_tells_me_things_____evil_things_by_thedraweronetwothree_d90qwu8

jackpoz:
Thanks for the PR :)

Issue #23451, opened 2019-06-20
Core/GameObject: Restock for chests


Killyana:
**Description:** The restock for gobs type chest is not implemented or implemented wrongly: https://github.com/TrinityCore/TrinityCore/commit/7011aabb56f061fc889ec42ce32b7605ceb35b53

GAMEOBJECT_TYPE_CHEST = 3
data2: chestRestockTime (time in seconds)
data3: consumable (State: Boolean flag)

Chest that must Restock/despawn:

- The ones with data2=0 AND data3=0: this chests didn't despawn and regenerate loots instantly, they all contains unique quest item
``` sql
SELECT * FROM gameobject_template WHERE TYPE=3 AND data2=0 AND data3=0;
```

- The ones with data2>0 AND data3=0: this chests didn't despawn and regenerate loots after X secs= data2 , they could contains quest item or any kind of item, If the chest contain 2 items or more and only one is looted the restock must reset the looted item inside the gob after X secs= data2
``` sql
SELECT * FROM gameobject_template WHERE TYPE=3 AND data2>0 AND data3=0;
```

- The ones with data2>0 AND data3=1: this chests despawns once empty and regenerate loots after X secs= data2  if not all items are looted (it depawns once empty), they could contains quest items or any kind of item, If the chest contains 2 items or more and only one is looted the restock must reset the looted item inside the gob after X secs= data2
``` sql
SELECT * FROM gameobject_template WHERE TYPE=3 AND data2>0 AND data3=1;
```

- The ones with data2=0 AND data3=1, this chests despawns once empty and doesn't regenerate loots.
``` sql
SELECT * FROM gameobject_template WHERE TYPE=3 AND data2=0 AND data3=1;
```

This must be implemented with https://github.com/TrinityCore/TrinityCore/issues/15730#issuecomment-486457812 to avoid having conflict with respawn/Restock

**Branch(es):** both

**TC rev. hash/commit:** rev. 9083fc149d61

Wyrserth:
BfA retail cannot be used as a reliable source because all the quest-related gameobjects are personal spawns and other players cannot affect it. From sniffs the only thing I could find out is that, after it is looted, the gameobject loses the interactable+sparkle dynamicflags to gain them in the next update tick. I don't yet know if this is just how it worked serverside back then or if it triggers a specific behavior clientside.

Wyrserth:
[...]

PR #23469, opened 2019-06-23
Core/GameObject: don't allow non-consumable goobers to despawn on use

### LINKED ISSUE #15730: Core/Gameobject: Gameobjects despawn once clicked

Wyrserth:
**Changes proposed:**

This PR fixes an issue where goober gameobjects ([Romantic Picnic Basket](https://www.wowhead.com/item=34480), the Naxxramas portals) would despawn after a player right-clicked on them.

They are goobers flagged as non-consumable (data3 == 0) so they should not despawn in any case unless they were summoned by a spell and they expired.

- [x] 3.3.5

**Issues addressed:** closes #15730, updates #23451.

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -767,22 +767,20 @@ void GameObject::Update(uint32 diff)
 
             loot.clear();
 
-            //! If this is summoned by a spell with ie. SPELL_EFFECT_SUMMON_OBJECT_WILD, with or without owner, we check respawn criteria based on spell
-            //! The GetOwnerGUID() check is mostly for compatibility with hacky scripts - 99% of the time summoning should be done trough spells.
-            if (GetSpellId() || GetOwnerGUID())
+            // Do not delete goobers that are not consumed on loot, while still allowing them to despawn when they expire if summoned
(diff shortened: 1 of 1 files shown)


Killyana:
There's 2 cases that must be checked:
1- https://woehead.way-of-elendil.fr/?item=34480/romantic-picnic-basket -- works fine
2- .go game 6166 Naxx portal -- works fine
3- .quest add 12069  & .go c id 26608  kill the npc -- Anub'et'kan's Carapace must not despawn upon looted, this was working correctly before this PR

Wyrserth:
Forgot to keep the old behavior for summoned chests. Fixed both issues, thanks!

Killyana:
Everything looks fine, it could be pushed one the restock is done.

Wyrserth:
I made this PR to fix the major issue (goobers despawning) until we have enough data to understand how the restock timer works for gameobjects involved in quests.

BfA retail cannot be used as a reliable source because all the quest-related gameobjects are personal spawns and other players cannot affect it. From sniffs the only thing I could find out is that, after it is looted, the gameobject loses the interactable+sparkle dynamicflags to gain them in the next update tick. I don't yet know if this is just how it worked serverside back then or if it triggers a specific behavior clientside.

PR #23493, opened 2019-06-23
Core/SAI: do not pass old creatureData to UpdateEntry() and keep current health when changing a NPC's entry

Wyrserth:
**Changes proposed:**

Currently we pass the old CreatureData to SMART_ACTION_UPDATE_TEMPLATE, before updating the entry. That causes some values (including modelid) to not update with the correct ones from the new creature entry.

Also the health should remain at the same percentage as it was before, instead of resetting (changing entry doesn't grant a creature a free Lay on Hands, it's the same creature with the same damage inflicted on it).

Test case: #23494. 

- [x] 3.3.5


Wyrserth:
[Proof from BfA 8.2 PTR here](https://imgur.com/a/UD3V9LM).

PR #23495, opened 2019-06-23
Core/Quest: fix money reward for quests turned in at level cap

### LINKED ISSUE #20445: [3.3.5] Wrong gold reward for certain quests at level cap
**Description:**

Some quests give a lot more gold than they should at level cap. I think this affects quests with a quest level equal or superior to 80.

This applies to, at least, every Icecrown/Storm Peaks daily quest.

Let's take [Blood of the Chosen](http://www.wowhead.com/quest=13330) as example.

In DB, RewardMoney is 74000. RewardBonusMoney is 132300.

- When a level 80 tries to accept that quest, they should apparently see only the RewardBonusMoney value ([video here](https://youtu.be/DBdknwcQQJA?t=456)). But in TC, RewardMoney is sent instead.

- When the player opens the quest log and checks the quest, they see the RewardBonusMoney value in both TC and retail. This is probably the client handling the value to show.

- When the quest is turned in, in TC they receive RewardMoney+RewardBonusMoney gold.

This only seems to affect high-level quests. For lower level quests, the correct value at level cap is RewardMoney+RewardBonusMoney.

I'd like to know if someone else has any insight on the matter.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 948cc64

**TDB version:** 335.63

Wyrserth:
**Changes proposed:**

At level cap (at least during WotLK), any experience from quests earned at level cap is converted to money at a rate of 6 coppers for each point of experience ([source](https://wowwiki.fandom.com/wiki/Money)).

Right now, the RewardBonusMoney QuestInfo field is used. However, it behaves erratically: for some quests it contains the full sum that the quest should give, in some cases it has to be added to RewardMoney, and in other cases the two values do not yield any meaningful value.

Using those values results in quests giving way too much money than they're supposed to. For comparison, a normal level 80 daily quest is supposed to give 13g 50s, while on TC they give upwards of 20g.

I believe that field is a relic of the past, from when Blizzard used it to send the experience reward to the client. They switched to DBC sometime in WotLK, but that field remained.

This PR uses the standard calculation of 1 exp point = 6 coppers, which returns the right values for every quest I could find a video/other proof of it being completed at level cap.

Closes #20445.

- [x] 3.3.5

### DIFF
--- a/src/server/game/Quests/QuestDef.cpp
+++ b/src/server/game/Quests/QuestDef.cpp
@@ -199,7 +199,7 @@ void Quest::LoadQuestMailSender(Field* fields)
     _rewardMailSenderEntry = fields[1].GetUInt32();
 }
 
-uint32 Quest::XPValue(Player* player) const
+uint32 Quest::GetXPReward(Player const* player) const
 {
     if (player)
     {
@@ -227,14 +227,17 @@ uint32 Quest::XPValue(Player* player) const
     return 0;
 }
 
-int32 Quest::GetRewOrReqMoney() const
+int32 Quest::GetRewOrReqMoney(Player const* player) const
 {
     // RequiredMoney: the amount is the negative copper sum.
-    if (_rewardMoney <= 0)
+    if (_rewardMoney < 0)
         return _rewardMoney;
 
     // RewardMoney: the positive amount
-    return int32(_rewardMoney * sWorld->getRate(RATE_MONEY_QUEST));
+    if (!player || player->GetLevel() < sWorld->getIntConfig(CONFIG_MAX_PLAYER_LEVEL))
+        return int32(_rewardMoney * sWorld->getRate(RATE_MONEY_QUEST));
+    else // At level cap, the money reward is the maximum amount between normal and bonus money reward
+        return std::max(int32(GetRewMoneyMaxLevel()), int32(_rewardMoney * sWorld->getRate(RATE_MONEY_QUEST)));
 }
 
 uint32 Quest::GetRewMoneyMaxLevel() const
--- a/src/server/game/Quests/QuestDef.h
+++ b/src/server/game/Quests/QuestDef.h
@@ -207,7 +207,7 @@ class TC_GAME_API Quest
         void LoadQuestTemplateAddon(Field* fields);
         void LoadQuestMailSender(Field* fields);
 
-        uint32 XPValue(Player* player) const;
+        uint32 GetXPReward(Player const* player) const;
 
         bool HasFlag(uint32 flag) const { return (_flags & flag) != 0; }
         void SetFlag(uint32 flag) { _flags |= flag; }
@@ -257,7 +257,7 @@ class TC_GAME_API Quest
         std::string const& GetRequestItemsText() const { return _requestItemsText; }
         std::string const& GetAreaDescription() const { return _areaDescription; }
         std::string const& GetCompletedText() const { return _completedText; }
-        int32  GetRewOrReqMoney() const;
+        int32  GetRewOrReqMoney(Player const* player = nullptr) const;
         uint32 GetRewHonorAddition() const { return _rewardHonor; }
         float GetRewHonorMultiplier() const { return _rewardKillHonor; }
         uint32 GetRewMoneyMaxLevel() const; // use in XP calculation at client
--- a/src/server/game/Entities/Creature/GossipDef.cpp
+++ b/src/server/game/Entities/Creature/GossipDef.cpp
@@ -473,8 +473,8 @@ void PlayerMenu::SendQuestGiverQuestDetails(Quest const* quest, ObjectGuid npcGU
                 data << uint32(0);
         }
 
-        data << uint32(quest->GetRewOrReqMoney());
-        data << uint32(quest->XPValue(_session->GetPlayer()) * sWorld->getRate(RATE_XP_QUEST));
+        data << uint32(quest->GetRewOrReqMoney(_session->GetPlayer()));
+        data << uint32(quest->GetXPReward(_session->GetPlayer()) * sWorld->getRate(RATE_XP_QUEST));
     }
 
     // rewarded honor points. Multiply with 10 to satisfy client
@@ -587,8 +587,8 @@ void PlayerMenu::SendQuestGiverOfferReward(Quest const* quest, ObjectGuid npcGUI
             data << uint32(0);
     }
 
-    data << uint32(quest->GetRewOrReqMoney());
-    data << uint32(quest->XPValue(_session->GetPlayer()) * sWorld->getRate(RATE_XP_QUEST));
+    data << uint32(quest->GetRewOrReqMoney(_session->GetPlayer()));
+    data << uint32(quest->GetXPReward(_session->GetPlayer()) * sWorld->getRate(RATE_XP_QUEST));
 
     // rewarded honor points. Multiply with 10 to satisfy client
     data << uint32(10 * quest->CalculateHonorGain(_session->GetPlayer()->GetQuestLevel(quest)));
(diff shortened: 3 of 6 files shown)


Killyana:
Closes https://github.com/TrinityCore/TrinityCore/issues/20445  ?

Wyrserth:
Yeah, it does. Thanks!

src/server/game/Quests/QuestDef.cpp
```diff
@@ -227,14 +227,17 @@ uint32 Quest::XPValue(Player* player) const
...
-    if (_rewardMoney <= 0)
+    if (_rewardMoney < 0)
         return _rewardMoney;
 
     // RewardMoney: the positive amount
-    return int32(_rewardMoney * sWorld->getRate(RATE_MONEY_QUEST));
+    if (!player || player->GetLevel() < sWorld->getIntConfig(CONFIG_MAX_PLAYER_LEVEL) || HasFlag(QUEST_FLAGS_NO_MONEY_FROM_XP))
+        return int32(_rewardMoney * sWorld->getRate(RATE_MONEY_QUEST));
+    else
+        return GetXPReward(player) * 6 * sWorld->getRate(RATE_MONEY_MAX_LEVEL_QUEST); // At level cap, any potential experience rewarded by the quest is converted to money at a rate of 6 coppers per 1 point of exp
```
Treeston: should this not be added to the reward value listed above?

src/server/game/Quests/QuestDef.h
```diff
@@ -257,7 +257,7 @@ class TC_GAME_API Quest
         std::string const& GetRequestItemsText() const { return _requestItemsText; }
         std::string const& GetAreaDescription() const { return _areaDescription; }
         std::string const& GetCompletedText() const { return _completedText; }
-        int32  GetRewOrReqMoney() const;
+        int32  GetRewOrReqMoney(const Player* player = nullptr) const;
```
Treeston: `Player const*`

src/server/game/Quests/QuestDef.cpp
Wyrserth: Nope, the 6 * exp is the total reward, not the bonus added to the base money reward.

Treeston: are there any quests which have a very high "regular" money reward? should it be the maximum of the two?

src/server/game/Quests/QuestDef.h
Wyrserth: Fixed, thanks!

(I'll never get used to that codestyle rule, I've been reading and using const type* for years :P)

src/server/game/Quests/QuestDef.cpp
Wyrserth: I just found a similar case. The [Amphitheather of Anguish](https://www.wowhead.com/quest=12932) quests give 32100 exp, that converts into 19g 26s 00c. The quest itself gives 19g 50s 00c, and from [this video](https://youtu.be/JZb2JF9SNHA?t=10) it seems that the player at level 80 gets 19g 50s.

Treeston: So it should be `std::max` of the two then?

Wyrserth: Yeah, it makes sense. I'm searching for more similar cases but that seems the way it works, because it doesn't make sense that you can get less money for the same quest completed at a higher level.

Wyrserth:
Putting this on hold for a while. I'm digging through old screenshots to see if I have some factual examples of level 71-75 quests being turned in at level cap.

Wyrserth:
So indeed it takes the maximum value between the quest reward and the bonus reward, thanks Treeston!

Some quests give the same money reward regardless of level, like [Crackin' Some Skulls](https://www.wowhead.com/quest=10009) ([screenshot](https://imgur.com/a/GlX5dLP)).

There's actually no need to calculate the reward, the field RewardBonusMoney already contains the correct precalculated value, and the client uses it to show the correct money reward in the quest log.

HOWEVER! TC has wrong values for some quests, probably because they changed between TBC and WotLK.

Here's some proof:

- The quest [Now That We're Friends...](https://www.wowhead.com/quest=9726) should reward 66 silvers at level 80. TC has 5g 16s set in RewardBonusMoney. Sniffs say RewardBonusMoney: 6600. [Screenshot from WotLK Retail](https://imgur.com/a/GWx632i).

- From the same screenshot as above, [Bring Me A Shrubbery!](https://www.wowhead.com/quest=9715) rewards 1g 32s. TC has 13g 20s. Sniffs say RewardBonusMoney: 13200.

I've compared all the bonus values for all quests with their calculated value (exp given at level 80 * 6) and came out with [this list](https://gist.github.com/Wyrserth/ea2ca197e6da0cc757a045bf310f851f). 267 quests out of 9464 have a potentially wrong/outdated value set in RewardBonusMoney.

Fixing those values by comparing them to sniffs is outside the scope of this PR - and it's not vital. Even if the current values are kept, TC already gives rewardmoney+rewardbonusmoney at quest reward. It can't get worse than that.

Treeston:
You mention that "the client uses (the values) to show the correct money reward in the quest log".

Are you saying that the client displays correct rewards despite TC's database being incorrect? If yes, where is the client getting its information from?

Wyrserth:
By "the values" I meant the ones in DB, sorry if it wasn't clear :P

The client uses RewardBonusMoney and RewardMoney from DB obtained via SMSG_QUEST_QUERY_RESPONSE to show the correct calculated value (if the player is at level 80, it shows the highest between reward and bonus, otherwise it only shows the normal reward money).

Shauren:
Want me to poke around the client at some point to see what it does?

Wyrserth:
If you can find the code that shows the money reward in the quest log, it'd be very helpful!

Shauren:
Here it is
<spoiler>
```
signed int __cdecl Script_GetQuestLogRewardMoney(int a1)
{
  WGUID playerGuid; // rax@1
  int player; // eax@1
  bool isMaxLevel; // esi@3
  int questInfo; // eax@5
  int rewardMoney; // edx@5
  int rewardBonusMoney; // eax@9
  WGUID guid; // [sp+Ch] [bp-Ch]@5
  int finalRewardMoney; // [sp+14h] [bp-4h]@5

  playerGuid = ClntObjMgrGetActivePlayer();
  player = (int)ClntObjMgrGetObjectPtr(
                  playerGuid,
                  TYPEMASK_PLAYER,
                  (int)"d:\\BuildServer\\WoW\\1\\work\\WoW-code\\branches\\wow-patch-3_3_5_A-BNet\\WoW\\Source\\Object/Ob"
                       "jectClient/Player_C.h",
                  160);
  isMaxLevel = player && *(_DWORD *)(*(_DWORD *)(player + 208) + 192) >= *(_DWORD *)(*(_DWORD *)(player + 4104) + 4524);// UNIT_FIELD_LEVEL >= PLAYER_FIELD_MAX_LEVEL
  guid.guid_low = 0;
  guid.guid_high = 0;
  questInfo = DBCache<QuestCache,int,HASHKEY_DWORD,NullXInfo>::GetRecord(
                &g_questDBCache,
                CGQuestLog::m_selectedQuest,
                &guid,
                0,
                0,
                0);
  rewardMoney = 0;
  finalRewardMoney = 0;
  if ( questInfo )
  {
    if ( *(_DWORD *)(questInfo + 52) > 0 )
    {
      rewardMoney = *(_DWORD *)(questInfo + 52);
      finalRewardMoney = *(_DWORD *)(questInfo + 52);
    }
    if ( isMaxLevel )
    {
      rewardBonusMoney = *(_DWORD *)(questInfo + 56);
      if ( rewardMoney <= rewardBonusMoney )
        finalRewardMoney = rewardBonusMoney;
    }
  }
  lua_pushnumber(a1, (double)finalRewardMoney);
  return 1;
}
```
</spoiler>

Wyrserth:
Thanks a lot! Indeed it takes the maximum value between RewardMoney and BonusRewardMoney without calculating anything. Needed the proof as I only had a few old screenshots to work with.

So most quests will give the correct rewards. Some have probably outdated/wrong data in quest_template and will give the wrong money reward until fixed (it is still better than the current implementation that just sums the two fields).

PR #23467, opened 2019-06-22
Core/Spells: implement SPELL_AURA_OVERRIDE_AUTOATTACK_WITH_MELEE_SPELL

Ovahlord:
**Changes proposed:**
-  implement SPELL_AURA_OVERRIDE_AUTOATTACK_WITH_MELEE_SPELL which is suposed to override default basic melee attacks with spell casts instead

**Target branch(es):** 3.3.5/master

- [ ] 3.3.5
- [x] master

**Tests performed:** (Does it build, tested in-game, etc.)
Tested ingame. Video:
https://www.youtube.com/watch?v=B-5Q-3pv7xQ
Notice the green claw swipes when the boss is doing melee attacks. these are the indicator for him doing attacks with that aura. Garothi will follow as seperate PR soon(tm).

PR #23497, opened 2019-06-23
Scripts/Antorus: implement Garothi Worldbreaker encounter 

Ovahlord:
**Changes proposed:**
-  implement Garothi Worldbreaker encounter for normal and heroic difficulty
-  added missing template data and encounter related spawns
-  added instance script and encounter data for Garothi

- [ ] 3.3.5
- [x] master

Tested ingame:
https://www.youtube.com/watch?v=XBkXH6mjCcA

- [x] Mythic Difficulty. Not within scope of this PR though. 
- [x] need merged PR for full functionality: https://github.com/TrinityCore/TrinityCore/pull/23496
- [x] need merged PR for full functionality: https://github.com/TrinityCore/TrinityCore/pull/23467

### DIFF
--- a/src/server/scripts/Argus/AntorusTheBurningThrone/boss_garothi_worldbreaker.cpp
+++ b/src/server/scripts/Argus/AntorusTheBurningThrone/boss_garothi_worldbreaker.cpp
@@ -1,0 +1,890 @@
+/*
+* Copyright (C) 2008-2019 TrinityCore <http://www.trinitycore.org/>
+*
+* This program is free software; you can redistribute it and/or modify it
+* under the terms of the GNU General Public License as published by the
+* Free Software Foundation; either version 2 of the License, or (at your
+* option) any later version.
+*
+* This program is distributed in the hope that it will be useful, but WITHOUT
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+* FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+* more details.
+*
+* You should have received a copy of the GNU General Public License along
+* with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "GridNotifiers.h"
+#include "AreaTriggerAI.h"
+#include "ScriptMgr.h"
+#include "ScriptedCreature.h"
+#include "SpellAuras.h"
+#include "SpellScript.h"
+#include "SpellAuraEffects.h"
+#include "antorus_the_burning_throne.h"
+
+enum Texts
+{
+    // Garothi Worldbreaker
+    SAY_AGGRO                       = 0,
+    SAY_DISENGAGE                   = 1,
+    SAY_ANNOUNCE_APOCALYPSE_DRIVE   = 2,
+    SAY_APOCALYPSE_DRIVE            = 3,
+    SAY_ANNOUNCE_ERADICATION        = 4,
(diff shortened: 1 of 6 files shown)


src/server/scripts/Argus/AntorusTheBurningThrone/boss_garothi_worldbreaker.cpp
```diff
@@ -0,0 +1,884 @@
...
+    }
+
+    void OnUnitEnter(Unit* unit) override
+    {
+        if (!unit->IsPlayer())
+            return;
+
+        _playerCount++;
+
+        if (Creature* annihilation = at->FindNearestCreature(NPC_ANNIHILATION, 3.0f, true))
```
Krudor: The annihilation npc is the summoner of the areatrigger right?
Isn't it possible to get the summoner of this areatrigger in that case?

src/server/scripts/Argus/AntorusTheBurningThrone/boss_garothi_worldbreaker.cpp
```diff
@@ -0,0 +1,884 @@
...
+                events.ScheduleEvent(EVENT_REENGAGE_PLAYERS, 3s + 500ms);
+                HideCannons();
+                me->RemoveUnitFlag(UNIT_FLAG_NOT_SELECTABLE);
+                break;
+            default:
+                break;
+        }
+    }
+
+    void DamageTaken(Unit* /*attacker*/, uint32& damage) override
```
Krudor: Wouldn't it be better to just initialize two private variables that is initialized to 65/60 and 35/20 depending on difficulty and using those, instead of duplicating logic and checking difficulty every time damage is taken?

src/server/scripts/Argus/AntorusTheBurningThrone/boss_garothi_worldbreaker.cpp
Ovahlord: The areatrigger AI class does not have a summoner member. That's why I had to go this way.

src/server/scripts/Argus/AntorusTheBurningThrone/boss_garothi_worldbreaker.cpp
Ovahlord: Well, it's just a getter and a switch here so I don't see any major reason to just dump the numbers in advance unless you want me to do micro-optimizations

Shauren: Difficulty switching inside is not a thing for instances that have normal/heroic/mythic modes, it only applies to 10/25/10h/25h ones

Ovahlord: Allright. Please stall the pr for a little longer then. I'm currently working on Mythic difficulty mechanics as well which will take a little longer because I'm also getting rid of some hook usages in the meantime because I got complaints that SpellHit and SpellHitTarget should no longer be used. I personally would disagree because they are great for AI based actions but well.

src/server/scripts/Argus/AntorusTheBurningThrone/boss_garothi_worldbreaker.cpp
mdX7: at->GetCaster()

PR #23500, opened 2019-06-24
Core/Mail: include 100g in the mail sent after turning in the quest The Wrath of Neptulon

Wyrserth:
**Changes proposed:**

After turning in the quest [The Wrath of Neptulon](https://www.wowhead.com/quest=8729), the player will receive a mail (after 24 hours) that should contain 100 gold. [Source](https://www.wowhead.com/quest=8729/the-wrath-of-neptulon#comments:id=83809).

- [x] 3.3.5

### DIFF
--- a/src/server/game/Mails/Mail.cpp
+++ b/src/server/game/Mails/Mail.cpp
@@ -100,7 +100,7 @@ void MailDraft::prepareItems(Player* receiver, SQLTransaction& trans)
 
     m_mailTemplateItemsNeed = false;
 
-    // The mail sent after turning in the quest The Wrath of Neptulon contains 100g
+    // The mail sent after turning in the quest The Good News and The Bad News contains 100g
     if (m_mailTemplateId == 123)
         m_money = 1000000;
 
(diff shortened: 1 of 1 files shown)


src/server/game/Mails/Mail.cpp
```diff
@@ -100,6 +100,10 @@ void MailDraft::prepareItems(Player* receiver, SQLTransaction& trans)
 
     m_mailTemplateItemsNeed = false;
 
+    // The mail sent after turning in the quest The Wrath of Neptulon contains 100g
+    if (m_mailTemplateId == 123)
```
jackpoz: maybe there is a more generic way to specify money for specific mail template id ?

Shauren: Is it really directly attaching gold? Blizz tends to create bags with gold loot for this purpose

Wyrserth: The closest info I found is from [Wowwiki](https://wowwiki.fandom.com/wiki/Quest:The_Good_News_and_The_Bad_News) and [Wowhead](https://www.wowhead.com/quest=8729) where nobody mentions a bag. Also tried to search in item_template (the entries close to the quest items, as Blizzard usually does them sequentially) for potential items, but nothing.

I thought about making this a table (mail_money_reward) that takes mailTemplateId and the amount of coppers to attach. Is it worth it to have a whole table for it when this is the only blizzlike use case for it?

PR #23524, opened 2019-06-26
Scripts/Stockades: Cleanup the scripts

Artamedes:
**Changes proposed:**

- Scripts looked ugly before, now less ugly
- Move Mortimer Moloch and Warden Thelwater to use TaskScheduler rather than EventMap
- Use new script registers
- Cleanup some other code

- [ ] 3.3.5
- [x] master

Builds, not tested ingame

### DIFF
--- a/src/server/scripts/EasternKingdoms/TheStockade/boss_hogger.cpp
+++ b/src/server/scripts/EasternKingdoms/TheStockade/boss_hogger.cpp
@@ -60,12 +60,16 @@ Position const wardenThelwaterSpawnPosition = { 138.369f, 78.2932f, -33.85627f, 
 // Hogger - 46254
 struct boss_hogger : public BossAI
 {
-public:
     boss_hogger(Creature* creature) : BossAI(creature, DATA_HOGGER), _hasEnraged(false) { }
 
-    void EnterCombat(Unit* /*who*/) override
+    void Reset() override
     {
-        _EnterCombat();
+        _hasEnraged = false;
+    }
+
+    void EnterCombat(Unit* who) override
+    {
+        BossAI::EnterCombat(who);
 
         Talk(SAY_PULL);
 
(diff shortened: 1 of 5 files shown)


src/server/scripts/EasternKingdoms/TheStockade/boss_hogger.cpp
```diff
@@ -98,7 +100,7 @@ class boss_hogger : public CreatureScript
             if (me->HasUnitState(UNIT_STATE_CASTING))
                 return;
 
-            while (uint32 eventId = events.ExecuteEvent())
+            if (uint32 eventId = events.ExecuteEvent())
```
jackpoz: Could you comment this change a bit more?

jackpoz (commented):
Could you test it ingame ?

src/server/scripts/EasternKingdoms/TheStockade/boss_hogger.cpp
Artamedes: Don't cast 2 spells at same update (if it somehow does happen), because of casting check above

src/server/scripts/EasternKingdoms/TheStockade/boss_hogger.cpp
```diff
@@ -111,6 +113,7 @@ class boss_hogger : public CreatureScript
                         events.Repeat(Seconds(15), Seconds(20));
                         break;
                 }
+
                 if (me->HasUnitState(UNIT_STATE_CASTING))
```
mdX7: should not be required anymore due to the change above

src/server/scripts/EasternKingdoms/TheStockade/boss_hogger.cpp
kelno: Those functions redundancy can be reduced a lot by using the BossAI::ExecuteEvent hook.
The whole UpdateAI function could be replaced with
```
void ExecuteEvent(uint32 eventId) override
{
    switch (eventId)
    {
    case EVENT_VICIOUS_SLICE:
        DoCastVictim(SPELL_VICIOUS_SLICE);
        events.Repeat(Seconds(10), Seconds(14));
        break;
     case EVENT_MADDENING_CALL:
        DoCast(SPELL_MADDENING_CALL);
        events.Repeat(Seconds(15), Seconds(20));
        break;
    }
}
```

PR #23526, opened 2019-06-27
Core/GameObject: implement restock mechanic for non-consumable gameobjects

### LINKED ISSUE #23451: Core/GameObject: Restock for chests

Wyrserth:
**Changes proposed:**

Some gameobject chests should never despawn (data3 = 0) but instead they become inactive until their loot is restocked (data2 > 0).

This PR implements this behavior, and fixes the following cases:

-  Only one quest item inside the gameobject, it is looted: chest becomes inactive (not usable) until the restock time passes. Then it regens the loot and becomes activable.
-  Two quest items inside the gameobjects, one is looted: the other remains available and the gameobject remains active. When the restock timer elapses, the other item is regenerated.

- [x] 3.3.5

**Issues addressed:** closes #23451 by fixing the remaining issues listed there.

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -797,7 +797,7 @@ void GameObject::Update(uint32 diff)
                 UpdateObjectVisibility();
                 return;
             }
-            else
+            else if (GetOwner() || GetSpellId())
             {
                 SetRespawnTime(0);
                 Delete();
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -510,6 +510,13 @@ void GameObject::Update(uint32 diff)
                     }
                     return;
                 }
+                case GAMEOBJECT_TYPE_CHEST:
+                    if (m_restockTime > GameTime::GetGameTime())
+                        return;
+                    // If there is no restock timer, or if the restock timer passed, the chest becomes ready to loot
+                    m_lootState = GO_READY;
+                    SendLootStateUpdateToNearbyPlayers();
+                    break;
                 default:
                     m_lootState = GO_READY;                         // for other GOis same switched without delay to GO_READY
                     break;
@@ -694,6 +701,13 @@ void GameObject::Update(uint32 diff)
                         }
                         else m_groupLootTimer -= diff;
                     }
+
+                    // Gameobject was partially looted and restock time passed, restock all loot now
+                    if (GameTime::GetGameTime() >= m_restockTime)
+                    {
+                        m_lootState = GO_READY;
+                        SendLootStateUpdateToNearbyPlayers();
+                    }
                     break;
                 case GAMEOBJECT_TYPE_TRAP:
                 {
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -2321,6 +2348,27 @@ void GameObject::SetLootState(LootState state, Unit* unit)
     }
 }
 
+void GameObject::SendLootStateUpdateToNearbyPlayers()
```
Shauren: What exactly is the point of this function here

Wyrserth: Gameobjects don't send update field changes when their loot state changes, this causes some issues like gameobjects being still active after a player loots the last quest item. This forces an update to all nearby players so they see the gameobject as activable or not activable depending on state.

Shauren: Just call `AddToObjectUpdateIfNeeded()` to schedule the update at the end of Map::Update, you are talking about GAMEOBJECT_DYNAMIC field which is always sent (if object gets its update sent), don't force immediate packet

Wyrserth: Fixed, thanks!

Killyana:
Everything looks fine, there's an issue about the gob state not being updated on client side when the player fulfill a quest objective, but it's an old issue that could be fixed later.
Some gobs have wrong loots and must be fix to avoid abuse like:
https://woehead.way-of-elendil.fr/?object=1765
``` sql
DELETE FROM `gameobject_loot_template` WHERE `entry`=1570 AND `Item`!=3706;

PR #23527, opened 2019-06-27
Core/Conditions: allow spellclick conditions to properly work for aura apply/remove

### LINKED ISSUE #23168: Core/Conditions: Conditions using auras are not updated correctly
**Description:** If you make a condition with ConditionTypeOrReference=CONDITION_AURA the result will not be updated until the units involved in the conditions are refreshed for the client.

For example if a npc has a spell click and this spell click has a condition using the aura from the spell, the player will not be able to click the same npc again until this spawn is reloaded by the client, and for other npc (same entry)  the IconeName will display an available spell_click but once licked it does nothing.

1. .quest add 13045
2. .go c id 30500
3. click a drake
4. Notice the clicked Drake no longer display the spell_click in mouse-over
5. Notice other drakes display the spell click but if you click it does nothing
6. dismount, notice the spell click remain not available in the first drake

**Branch(es):** both

**TC rev. hash/commit:** rev. 1374658d727d

Wyrserth:
**Changes proposed:**

Currently, conditions involving spellclicks will not work properly when used with CONDITION_AURA.

The spellclick will not be updated clientside until the creatures are reloaded by going out-of-range and back, or by them respawning.

This PR calls UpdateVisibleGameobjectsOrSpellClicks() (formerly UpdateForQuestWorldObjects(), renamed because its behavior is useful for cases outside of quests too) to trigger an update for all visible creatures using spellclicks, so that the client will see if a spellclick is correctly available or not.

- [x] 3.3.5

**Issues addressed:** closes #23168

### DIFF
--- a/src/server/game/Conditions/ConditionMgr.cpp
+++ b/src/server/game/Conditions/ConditionMgr.cpp
@@ -1026,6 +1026,11 @@ bool ConditionMgr::IsObjectMeetingVendorItemConditions(uint32 creatureId, uint32
     return true;
 }
 
+bool ConditionMgr::IsSpellUsedInSpellClickConditions(uint32 spellId) const
+{
+    return SpellsUsedInSpellClickConditions.find(spellId) != SpellsUsedInSpellClickConditions.end();
+}
+
 ConditionMgr* ConditionMgr::instance()
 {
     static ConditionMgr instance;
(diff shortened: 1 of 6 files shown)


Shauren:
This one gets a no from me for performance reasons. This really needs to be done with a known list of units to update, not search for update candidates on every aura apply

Shauren:
This version looks better

PR #23533, opened 2019-06-28
Scripts/Spells: Remove Collapsing Cave Aura

Sorikoff:
**Changes proposed:**
Migrate spell Remove Collapsing Cave Aura to spell scripts system.

**Target branch(es):** 3.3.5

**Tests performed:** Works


src/server/scripts/Northrend/zone_storm_peaks.cpp
```diff
@@ -1436,6 +1436,27 @@ class spell_player_mount_wyrm : public SpellScriptLoader
         }
 };
 
+enum CollapsingCave
+{
+    SPELL_COLLAPSING_CAVE = 55486
```
jackpoz: is this the same as m_spellInfo->Effects[effIndex].CalcValue() ?

Sorikoff: Yes. Since this is constant value I thought it could be better to keep it in the separate variable in this case.

jackpoz:
Thanks for the PR :)

how many spells are left ? xD

Sorikoff:
Can't say. Right now definitely about 5, but probably more will follow.

PR #23542, opened 2019-06-29
Scripts/ICC: Corrections around DB data save [WARNING]

ccrs:
So, long story short, currently the DB saving behaviour is all fucked up.

Data containing each boss state must mach the expected in forms of quantity and order (defined in a dbc).

Why is this scenario actually fcked up? Because that same "boss state data" will be used in different operations, such as:

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Instances/InstanceScript.cpp#L796
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Maps/Map.cpp#L3813

Have you ever wondered why the hell that "instance save warning" looked always wrong on ICC?

So yay, all is good finally, we all cheer in joy.

But wait, its a trap.

Redefining the way the save is stored in the database is a huge pain in the ass for "active" servers.
Imagine corrupting on-going saves for the whole population of the server? 💃 

So, instead of commiting right away, I'll leave a grace period for you to prepare.
Yes you, you lurker pserver admin.
Comming soon, you better prepare for the situation, or die drowned by player reports.

Also, for extra points, what about being a good lad and sharing the counter measure here for others to rejoice?

**Target branch(es):**

- [x] 3.3.5
- [ ] master


ccrs:
<iframe width="475" height="250" src="https://vclock.com/embed/timer/#countdown=72:00:00&date=2019-07-03T01:42:38&title=Armageddon&theme=0&ampm=1&sound=xylophone" frameborder="0" allowfullscreen></iframe>

ccrs:
<figure class="video_container">
<iframe width="475" height="250" src="https://vclock.com/embed/timer/#countdown=72:00:00&date=2019-07-03T01:42:38&title=Armageddon&theme=0&ampm=1&sound=xylophone" frameborder="0" allowfullscreen></iframe>
<figure class="video_container">

jackpoz:
Isn't there a way to make it backward compatible ? With a sql query or so ? None will read this PR description and the changes are hidden in codestyle changes

Shauren (changes requested):
Revert ALL of the codestyle changes and moving code

src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
```diff
@@ -1319,28 +1321,31 @@ class instance_icecrown_citadel : public InstanceMapScript
...
+
                 data >> HeroicAttempts;
 
-                uint32 temp = 0;
                 data >> temp;
-                if (temp == IN_PROGRESS)
-                    ColdflameJetsState = NOT_STARTED;
-                else
-                    ColdflameJetsState = temp ? DONE : NOT_STARTED;
+                (temp == DONE) ? ColdflameJetsState = DONE : ColdflameJetsState = NOT_STARTED;
```
Shauren: No.

ccrs: I get the general concept of what you are saying, better leave operations outside ternary operators, but, could you please elaborate on why, specifically?

like, is it cause of some undesired or missing or whatever compiler optimization? :)

ccrs:
> Revert ALL of the codestyle changes and moving code

ummm I'm not so sure I'll do that, main reason being the actual objective of this PR.
I did it just to try spread some awareness of the possible chaos
(even if it lasts for just one week, the time for the instance save to reset, chaos is still chaos)

So... you know I'm all aboard the correctness and best practices train, but here it seems like overkill for me.

I want people to be prepared, not actual review of the code... makes sense?

but of course, still any review or guidance or help is pretty much welcome 👍

src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
jackpoz: I honestly prefer if/else on multiple lines so I can put a breakpoint on the statement I want

ccrs:
> Isn't there a way to make it backward compatible ? With a sql query or so ? None will read this PR description and the changes are hidden in codestyle changes

I dont really know if that's an option.
See, the save is stored as a concatenation of characters stored in a simple string...

So, how can we modify an existing one into the new format using just SQL?

I mean, I'm not that "skilled", to put it in some way, in SQL queries/procedures to do that, probably someone esle knows to do it though :)

jackpoz:
> I want people to be prepared, not actual review of the code... makes sense?

0% of people will be prepared if we warn them through PR. Also what should they do to avoid whatever bad thing to happen ? Should they update their core right after instance reset time ?

ccrs:
> > I want people to be prepared, not actual review of the code... makes sense?
> 
> 0% of people will be prepared if we warn them through PR. Also what should they do to avoid whatever bad thing to happen ? Should they update their core right after instance reset time ?

thats exactly why I wanted to create this PR, I have not the slightlest clue on what's the best approach to this situation 😅

src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
ccrs: I agree on that too, but since the operations are rather simple, the break would only make sense to know the actual `temp` value...
and that's done a line before, right?

src/server/scripts/Northrend/IcecrownCitadel/icecrown_citadel.cpp
```diff
@@ -15,290 +15,159 @@
...
 enum ICCSpells
 {
     // Rotting Frost Giant
-    SPELL_DEATH_PLAGUE              = 72879,
-    SPELL_DEATH_PLAGUE_AURA         = 72865,
-    SPELL_RECENTLY_INFECTED         = 72884,
-    SPELL_DEATH_PLAGUE_KILL         = 72867,
-    SPELL_STOMP                     = 64652,
-    SPELL_ARCTIC_BREATH             = 72848,
+    SPELL_DEATH_PLAGUE = 72879,
```
Keader: remove enum identation for?

ccrs: codestyle consistency, all of them where fcked up, but I can add em back correctly :)

ccrs:
Turns out, armageddon can be postponed.

You see, boss states and completedencounters mask are two separate mechanics that do not interfere with each other.

So, right now, I have no proof that this changes are actually necessary.
Svalna can keep its bossai :)

PR #23546, opened 2019-06-30
Local-only mode fix

### LINKED ISSUE #21309: Realm - No resolve address
**Description:**

If the authserver remains active for a long time, for no reason it has this error:

> Could not resolve address 127.0.0.1 for realm "Trinity" id 1

From there, the realm is not shown, then:

> Could not bind to 127.0.0.1:3724 Address already in use

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 514c847881a9fff504e90f1186a1209c0b8a3674

**TDB version:**  TDB 335.63

**Operating system:** Debian 9

SolarisSDK:
In response to issue #21309

**Changes proposed:**

Add flags to resolver to work in local-only cases.

**Target branch(es):**

- [v ] 3.3.5
- [v ] master  
Made for 3.3.5 but master seems the same.

**Issues addressed:** Closes  #21309  (in case of boost 1.65 or older)

**Tests performed:** Yes. Works flawlessly.

### DIFF
--- a/src/common/Asio/Resolver.h
+++ b/src/common/Asio/Resolver.h
@@ -31,13 +31,15 @@ namespace Trinity
         {
             boost::system::error_code ec;
 #if BOOST_VERSION >= 106600
-            boost::asio::ip::tcp::resolver::results_type results = resolver.resolve(protocol, host, service, ec);
+            boost::asio::ip::resolver_base::flags flagsResolver = boost::asio::ip::resolver_base::all_matching;
+            boost::asio::ip::tcp::resolver::results_type results = resolver.resolve(protocol, host, service, flagsResolver, ec);
             if (results.begin() == results.end() || ec)
                 return {};
 
             return results.begin()->endpoint();
 #else
-            boost::asio::ip::tcp::resolver::query query(std::move(protocol), std::move(host), std::move(service), boost::asio::ip::tcp::resolver::query::canonical_name);
+            boost::asio::ip::resolver_query_base::flags flagsQuery = boost::asio::ip::tcp::resolver::query::all_matching;
+            boost::asio::ip::tcp::resolver::query query(std::move(protocol), std::move(host), std::move(service), flagsQuery);
             boost::asio::ip::tcp::resolver::iterator itr = resolver.resolve(query, ec);
             boost::asio::ip::tcp::resolver::iterator end;
             if (itr == end || ec)
--- a/src/common/Asio/Resolver.h
+++ b/src/common/Asio/Resolver.h
@@ -37,7 +37,7 @@ namespace Trinity
 
             return results.begin()->endpoint();
 #else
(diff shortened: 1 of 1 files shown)


jackpoz:
Which boost versions range support that change ?

jackpoz:
I checked better the PR, this is only for Boost <= 1.65 . Does the issue happen with Boost 1.66+ too and does that need any additional flag too ?

SolarisSDK:
>  Does the issue happen with Boost 1.66+ too and does that need any additional flag too ?
Sorry, can't test it just now. Currently in business trip with "stable-only" notebook. But for the function used in 1.66 there is the same type of flags in overridden list of "resolve" function.
By the way, why are deprecated function is used for 1.66?
https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ip__tcp/resolver.html
As you can see it has (deprecated) status. Maybe someone got it messed up and this part of code should be used in <1.65 and query in >1.66 ??
The same applies to master branch cause files are the same.

SolarisSDK:
For the resolve function syntax in case used for 1.66 flag goes BEFORE error code.
results_type resolve(
    const protocol_type & protocol,
    string_view host,
    string_view service,
    resolver_base::flags resolve_flags, <<< here it goes
    boost::system::error_code & ec);

https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ip__basic_resolver/resolve.html

jackpoz:
I would say the deprecation is for the query overload, not for the overloads we use, see https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ip__basic_resolver/resolve.html (which is the same page you linked too).

If possible when you are back from the business trip try to reproduce the issue with Boost 1.66 and see if it needs the flag too (or if you know already that it requires them, then just add them)

Aokromes:
well, since boost under 1.66 where non-recommended for visual studio 2017 maybe we can simply increase requirement to 1.66

SolarisSDK:
> well, since boost under 1.66 where non-recommended for visual studio 2017 maybe we can simply increase requirement to 1.66

That's not a boost or boost version problem at all. In 1.66 there might be a same behavior on same circumstances.
I'll try to test it before friday on vm.
Dropping support will hit current LTS of ubuntu (1.65) and Debian stable (1.62) .

Aokromes:
"for visual studio"  anyway debian 10 is 4 days from now, and we don't support ubuntu LTS.

SolarisSDK:
Update. Had some research and figured that most useful flag would be "::all_matching;@"
Tested on my laptop as well on my testing server. Resolution work fine.

jackpoz:
LGTM (although I didn't test it)

jackpoz:
Thanks for the PR :)

SolarisSDK:
Happy to help!
I'll try to look in some other old issues.

PR #23547, opened 2019-06-30
Scripts/Spells: Dimensional Ripper - Everlook

Sorikoff:
**Changes proposed:**
Migrate spell `Dimensional Ripper - Everlook` to spell scripts system, therefore removing all hacks from teleporting post-effects.

**Target branch(es):** 3.3.5

**Tests performed:** Not tested in-game (**should** work)


Sorikoff:
Tested in-game. All spells are applied when using spell.

PR #23571, opened 2019-07-07
Core/Movement: MoveSmoothPath will now insert the mover's current position before the spline waypoints to avoid losing the first spline point

Ovahlord:
The spline movement system will use the first path vertex as the current position and since we are only passing over the path of the npc we need to add the first vertex manually to prevent the loss of the first spline waypoint. This will ensure that we will always have nice and complete results.

-  add a additional path point at the beginning of the path since the spline system uses the first point as starting position of the moving unit

- [x] 3.3.5
- [ ] master

- tested ingame via several scripts


Treeston:
Behavior is intentional, splinechains contain the current position.

PR #23574, opened 2019-07-08
Core/Weather: Port weather changes from master

ForesterDev:
**Changes proposed:**

-  Port 6f7d048 and 6eb9973 from master to 3.3.5
-  convert SMSG_WEATHER to new system
-  Fixed weather updates to be thread safe

- [x] 3.3.5

**Tests performed:** (Does it build, tested in-game, etc.) Build (on Windows), tested


jackpoz:
what does this PR fix in 335 ? are there any weather issues ?

Shauren:
Thread safety - WeatherMgr::AddWeather in Player::Update

src/server/game/Server/WorldSession.h
```diff
@@ -81,6 +81,10 @@ namespace WorldPackets
         class Hello;
         class TrainerBuySpell;
     }
+    namespace Misc
+    {
+        class Weather;
```
Shauren: still not a proper cherry-pick - this line does not exist in master and forward declaring in WorldSession is not needed for SMSG packets

Shauren: I'll just do it myself later today

Shauren:
Side note - the includes you have in MiscPackets.h are from the pre-include-cleanup times, this is what you get when doing cherry picks completely out of order

ForesterDev:
@Shauren, when I cherry-picked commit 6f7d048765f102110cbf9ad4653c078fbadaa105, I get the whole files MiscPackets with other unnecessary packet. I've cleaned they by hands, but headers left untouched

PR #23575, opened 2019-07-08
Core/Movement: add a velocity argument for the spline chain meta table

Ovahlord:
**Changes proposed:**

-  add a velocity argument for the spline chain meta table so we can now tell the creature at which speed it shall move instead of only using the default calculated speed values. This behaivior has been observed especially at flying creatures as they fly at many different speeds while their flight speed field remains unchanged.

For the wiki:
velocity = 0 - use default speeds
velocity > 0 - use the provided velocity for that spline path

- [x] 3.3.5
- [ ] master


Keader:
@Treeston good for you?

PR #23606, opened 2019-07-18
Core/Arenas: Remove doors in RL, NA and BE when arena has begun

Jildor:
**Changes proposed:**

- Remove doors in Ruins of Lordaeron Arena, Nagrand Arena and Blade's Edge Arena when arena has begun (5 secs after)
- https://www.youtube.com/watch?v=EwLOcI4h_kg&t=1205 <-- min 20:05 start arena and 5 secs after, the doors are removed
- https://www.warcraftmovies.com/movieview.php?id=163193 <-- min 17:23
- I don't see if should be the same in Dalaran Sewers Arena, but you can't abuse there
- Ring of Valor doesn't have doors

- [X] 3.3.5


jackpoz:
Thanks for the PR :)

PR #23630, opened 2019-07-24
Core/Defines: define UNIT_FLAG_SWIMMING

ForesterDev:
-  define UNIT_FLAG_SWIMMING
-  based on WPP https://github.com/TrinityCore/WowPacketParser/blob/master/WowPacketParser/Enums/UnitFlags.cs#L24 and cmangos https://github.com/cmangos/mangos-wotlk/blob/master/src/game/Entities/Unit.h#L584 and ingame behaviour
-  

**Target branch(es):** 

- [x] 3.3.5

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12713,7 +12713,7 @@ bool Unit::CanSwim() const
         return false;
     if (IsPet() && HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_PET_IN_COMBAT))
         return true;
-    return HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_RENAME | UNIT_FLAG_UNK_15);
+    return HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_RENAME | UNIT_FLAG_SWIMMING);
 }
 
 void Unit::NearTeleportTo(Position const& pos, bool casting /*= false*/)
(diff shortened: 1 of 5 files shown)


Shauren:
So... which one is it, SWIMMING or ONLY_SWIM (wpp)

ForesterDev:
I think, the right definition is SWIMMING. Because, there are a lot of creatures with this flags that can also walk on ground

Treeston:
if it only shows swimming animation _in water_, maybe `UNIT_FLAG_CAN_SWIM` is better?

ForesterDev:
But this flag not does not allow unit to swim (see Unit::CanSwim). UNIT_FLAG_CAN_SWIM may be misleading (I think)

Ovahlord:
You would be surprised if I tell you that the walkmode spline flag is actually CanSwim which handles the swimming animation. (Walkmode does not even exist because the walk animation / speed is determined by the spline length and duration.

so only_swim looks more reasonable to me since the splines handle swimming animations already.

ForesterDev:
\elysium developers defined this flag as 

`UNIT_FLAG_USE_SWIM_ANIMATION    = 0x00008000,`

jackpoz:
did you guys agree to the name yet ?

jackpoz:
> did you guys agree to the name yet ?

anyone ?

illfated:
I vote for `UNIT_FLAG_SWIMMING` (present participle **-ing** indicating current and ongoing activity).

jackpoz:
@ForesterDev is this PR ready ?

illfated:
Browsing the PR content, it looks like pure text replacement + 2 comment lines added.
Unless any more changes are needed, I would presume it is ready to use.
Would be nice to hear from the PR author, though.

Rushor:
we are all swimming and dancing like joker

PR #23631, opened 2019-07-24
Scripts/Defines: cleanup some magic numbers in factions

ForesterDev:
**Changes proposed:**

-  define some Faction temapates and cleanup magic numbers from scripts

**Target branch(es):** 

- [x] 3.3.5


jackpoz:
Thanks for the PR :)

PR #23633, opened 2019-07-24
[3.3.5] Somewhat more secure storage of TOTP PSK

Treeston:
We currently implement TOTP 2FA (think Google Authenticator) in TC. That's good.

We currently store pre-shared secrets in the database in plain. If a backup of your auth database goes places it shouldn't, then your 2FA is suddenly a 1FA. That doesn't _defeat_ the point, but kind of _hurts_ the point.

We can give you some ways to not shoot yourself in the foot as easily. Hopefully, you'll be a little less risky with your `authserver.conf`, because that's the only other (always-available, no I'm not supporting HSMs) "secure", big airquotes, location I can think of to put a secret to protect your secrets with.

---

* Argon2 and AES primitives added
  * ~Yes, the next PR will be Argon2 for passwords, because we're still using SHA1 and this is 2019. Duh.~
  * ~As @Shauren tells me, we can't actually do that.~
  * We will do that anyway.
* `TOTPTokenSecret` key in `authserver.conf` lets you specify an AES128 key in hex
  * We use that key to encrypt the TOTP PSKs stored in `accounts`. Don't lose it.
  * We also keep a hash of that key in `secret_digest` to ensure you don't shoot yourself in the foot.
  * When you want to change secret (or remove the encryption - just leave the key blank), we'll transition your data on startup for you if you specify `TOTPTokenOldSecret` in the `.conf`
* Store TOTP secrets in binary instead of as a base32 string. Conversion script included. (That was fun.)
* Added a +- 1 interval grace window for TOTP auth tokens while I was at it
  * Yeah, this is 2019 and everyone has NTP, but time drift happens. It doesn't hurt anyone.

And yes, that's just 3 major bullets for a PR this size. A lot of the diff lines are just library files being added, plus some cleanup of really rusty old code.


Treeston:
ps i'll be disappeared for the weekend, so the earliest i'll respond to comments is monday

just wanted to get this pr'd up before

Treeston:
@Shauren rewrote it to use openssl's aes implementation now as per your wishes

not entirely happy with it, documentation for that implementation is utterly nonexistent, and it's incredibly clunky to use.

also it appends a random block to the ciphertext and i have no idea what it does, none of the docs mention that block being appended (padding is turned off, and my messages are multiples of the block size) yet it insists on doing it.

blegh.

PS i'm overflowing a vector somewhere and it crashes, let me know if you see where i'm too tired right now

Shauren:
Im pretty sure I mentioned copypasting AES from master branch, not creating completely new api... (it definitely has in.size() == out.size())

src/common/Cryptography/AES.cpp
```diff
@@ -0,0 +1,130 @@
...
+        iv[BLOCK_SIZE_BYTES - i] = data.back();
+        data.pop_back();
+    }
+    return iv;
+}
+
+/*static*/ void AES::Encrypt(std::vector<uint8>& data, BigNumber const& bnKey)
+{
+    // add padding
+    AddPadding(data);
```
Warpten: You don't need to pad your input, OpenSSL EVP has padding enabled by default (PKCS#7 too).

I would also use `_ex` variants of `Init` and `Final`.

Treeston: i know, but i wanted to know in advance how much space i would need to allocate for the output (if i pad manually, input size should equal output size)

Warpten: And if you don't pad manually then output size is just input.length + block_size - input.length() % block_size.

Treeston:
AES now copy-pasted from master branch with some additions and cleanup

src/server/authserver/Secrets/SecretMgr.h
```diff
@@ -0,0 +1,55 @@
...
+#include "Common.h"
+#include "Optional.h"
+#include <array>
+#include <string>
+
+enum Secrets
+{
+    SECRET_TOTPTOKEN_KEY = 0,
+
+    // only add new indices right below this line
```
Shauren: you mean above

Treeston: i indeed do

Treeston:
conversion script done, what a fucking mess

Treeston:
tagging back to wip, want to move secretsmgr to common

Treeston:
done from my end, pending CI checks and review @Shauren

src/server/shared/SharedDefines.h
```diff
@@ -3770,6 +3770,15 @@ enum LineOfSightChecks
...
 };
 
+enum ServerProcessTypes
+{
+    SERVER_PROCESS_AUTHSERVER = 0,
+    SERVER_PROCESS_WORLDSERVER = 1,
+
+    NUM_SERVER_PROCESS_TYPES
+};
+extern TC_SERVERS_API ServerProcessTypes THIS_SERVER_PROCESS;
```
Shauren: No, don't do it like that. Make it internal to shared, and export a function that you will call in authserver and worldserver to initialize to proper value

Treeston: what's the benefit?

Shauren: Perhaps I wasn't clear, you have to do it this way or it will not compile with dynamic linking. (see appveyor)

Treeston:
all checks passing, pending review

src/common/Cryptography/Argon2.h
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Trinity
+{
+namespace Crypto
+{
+struct TC_COMMON_API Argon2
+{
+    static constexpr uint32 HASH_LEN = 16;                    // 128 bits, in bytes
+    static constexpr uint32 ENCODED_HASH_LEN = 100;           // in chars
+    static constexpr uint32 DEFAULT_ITERATIONS = 10;          // determined by dice roll, guaranteed to be secure (not really)
```
Warpten: :)

Warpten (approved):

Shauren (approved):

PR #23641, opened 2019-07-26
Core/MMaps: Avoid logging false positive errors about missing MMap tiles

jackpoz:
**Changes proposed:**

- Allow to specify a list of known missing MMap tiles that mmaps_generator doesn't create, while still reporting tiles that are actually missing as errors.
- Add a std::hash specialization for std::tuple with 1, 2 and 3 members, required by the change above

- [x] 3.3.5
- [ ] master

Lot of issues reported by users that were closed as "Intended"

- entered map 602 Halls of Lightning
- noticed the error log
  >Could not load MMAP name:Halls of Lightning, id:602, x:29, y:30 (mmap rep.: x:29, y:30)
Could not load MMAP name:Halls of Lightning, id:602, x:30, y:30 (mmap rep.: x:30, y:30)
Could not load MMAP name:Halls of Lightning, id:602, x:28, y:30 (mmap rep.: x:28, y:30)
- applied the PR
- the error log disappeared
- tried .debug loadcell
- new error logs appeared

- [ ] Add tiles from other maps, at least the most common ones

Please note that this is not supposed to include all tiles that .debug loadcell tries to load, just the ones that normal player behavior triggers.


Warpten:
Fyi Boost has `boost/functional/hash.hpp` which supports hashing tuples of any size

jackpoz:
I don't want to use any boost if possible, ty

Shauren:
No, boost/hash is extremely expensive to include and that alone is the reason why our Hash.h exists

Warpten:
Then maybe this? https://godbolt.org/z/5deS2E

(I'm fine with the PR as it is, just discussing if general-size tuple hashing support being provided would be nice)

jackpoz:
> just discussing if general-size tuple hashing support being provided would be nice

I like to stay away from any C++ meta-programming so feel free to commit whatever code that allows to put tuples in unordered_set (watch out for compiler supports, maybe open a PR so travis/circle ci/appveyor test it)

Warpten:
Simplified version: https://godbolt.org/z/DEG4Af

Does not match the behavior of your proposed hashers because it works on the tuple in reverse order. Can probably be tuned to go in direct order by tweaking `enable_if`s, but meh. I like 0-termination better.

jackpoz:
we decided to check if at least a neighbour tile exists and skip the log in that case, instead of hardcoding all the false positive tiles. It might not log some actual errors but it will be alright.

PR #23645, opened 2019-07-27
Core/AI: fixed creatures occasionally interrupting their channeled spells

Ovahlord:
**Changes proposed:**
-  expand the focusing check in UpdateVictim() to casting related movement blockers in general to avoid channeled spells getting interrupted because of missing checks

- [x] 3.3.5
- [ ] master

- Tested ingame, discovered places no longer have these issues.

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -243,7 +243,7 @@ bool CreatureAI::UpdateVictim()
     if (!me->HasReactState(REACT_PASSIVE))
     {
         if (Unit* victim = me->SelectVictim())
-            if (!me->IsFocusing(nullptr, true) && victim != me->GetVictim())
+            if (!me->IsMovementPreventedByCasting() && victim != me->GetVictim())
                 AttackStart(victim);
 
         return me->GetVictim() != nullptr;


jackpoz:
does the new code do the same check of the old one plus something more or does it replace it ?
There are other places where "if (!me->IsFocusing(nullptr, true) && victim != me->GetVictim())" line is used, should those be updated too ?

Ovahlord:
The change expands on the focus check to all casting movement preventing checks. Right now the only thing between preventing a creature from attacking you is the focusing mechanic instead of respecting all casting related checks. From my personal tests I have seen no regressions but it would be cool to have someone else to check for issues with that change.

jackpoz:
coudl you check the other lines too then while at it ?

jackpoz:
> could you check the other lines too then while at it ?

@Ovahlord did you have time to do this ? Could you fix the merge conflicts too ?

Ovahlord:
Hey there. I actually have sth else in mind to address that issue. I gave @Killyana a patch a few days ago regarding spell focusing that should take care of that

Aokromes:
Conflicting files
src/server/game/AI/CreatureAI.cpp

jackpoz:
should you update this PR then ? Including the other places where the old code was used in scripts ?

Ovahlord:
I would more say close it because the new change has a different scope

Issue #23210, opened 2019-04-25
Core/Vehicle: Vehicle accessories must changes phase with the main vehicle


Killyana:
**Description:** if a vehicle has accessories and changes phase it will not update his accessories with the new phase even if they remains installed

**Steps to reproduce the problem:**

1. .go c id 33060 
2. .mo pha 2 the vehicle
3. .mo pha 2 your char you will find the vehicle in phase 2 without the accessory
4. .mo pha 1 the vehicle
5. .mo pha 1 your char you will find the vehicle in phase 1 with the accessory installed

**Branch(es):** both

**TC rev. hash/commit:** rev. 28b502fed11a

jackpoz:
does this happen only when using the .mo pha command or with other ways too ?

Killyana:
It could happen if a player is controlling a vehicle then move to an area where the phase will shift for him, in this case only the vehicle will shift phase not the accessories

Killyana:
Here a special case: https://woehead.way-of-elendil.fr/?quest=13309
The player ride an accessory of the vehicle then the vehicle starts a waypoints, if at some moment the player shit phase, only the accessory will switch phase with him, the main vehicle will remind in the old phase.
What must happen:
- If the vehicle is not controlled: all accessories must shift phase with the main vehicle
- If the vehicle is controlled by the player all accessories must shift phase with the main vehicle (the vehicle will shift phase correctly with the player)
- If only an accessory is controlled by  the player: main vehicle and other accessories must shift  phase with the player

jackpoz:
https://github.com/TrinityCore/TrinityCore/pull/23646 I opened a pull request

PR #23646, opened 2019-07-27
Core/Vehicle: Change vehicle accessories' phase according to the player on the vehicle

jackpoz:
**Changes proposed:**

-  Core/Vehicle: Change vehicle accessories' phase according to the player on the vehicle

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes https://github.com/TrinityCore/TrinityCore/issues/23210

Tested various .mo pha ingame, while mounting and not mounting the vehicle or any of the accessories

- [ ] Sometimes I have to move very close to the vehicle for the accessories to appear
- [ ] There might be some quirks to the player interface when being on an accessory and that changes phase


Killyana:
About: "Sometimes I have to move very close to the vehicle for the accessories to appear" it's not related to the distance, just the install of the accessory i delay client side to avoid being loaded before the main vehicle, that cause the accessory not being installed correctly for the client.

I tested the PR, and it works fine when the player is installed in the main vehicle, but in the case where the player is installed inside an accessory the phases shift still fails (I'm not sure how it mut work in this case.

jackpoz:
> but in the case where the player is installed inside an accessory the phases shift still fails

did you change the phase of the Player, of the Vehicle or of the Accessory when testing ? Just to know which case to test

Killyana:
I changed player's phase when he's inside an accessory.

ccrs:
quest on argent crusade initial area involving cannons and killing lots of undeads.

when all the requirements are met, the area phase changes -> player changes phase

but, the player is still riding the cannon... -> this impl makes the cannon change phase too, right? -> it leaves the now "old" phase without a cannon till grid unload

Killyana:
No the phase change in quest rewarded not complete: https://youtu.be/8sD5UN067m8?t=114

Wyrserth:
Doesn't the phase change for [that quest](https://www.wowhead.com/quest=13086) happen when it's rewarded, and not when the objectives are complete?

illfated:
Good point, I thought so too (on rewarded). Then again, I could be remembering this incorrectly due to playing on TC instead of retail for so long.

Wyrserth:
Yeah I sent mine a few seconds after Killyana without realizing it :P

jackpoz:
I updated the PR, please try again. I pretty much tested all the cases:
- Player on vehicle, .mo pha the player
- Player on vehicle, .mo pha the vehicle
- Player on accessory, .mo pha the player
- Player on accessory, .mo pha the accessory
- Player on accessory, .mo pha the vehicle

Killyana:
I tested many cases and every thing looks fin for me.

PR #23657, opened 2019-07-29
Scripts/Misc: Simplify Code

Sorikoff:
**Changes proposed:**
Clean up code by _simplifying_ `return` expressions in some cases. And by _simplifying_ I mean that this PR replaces structures like:
```
if (statement)
    return true;

return false;
```

with single line:
```
return statement;
```

I am targetting only `spell_generic.cpp` in this PR for easier review.

**Target branch(es):** 3.3.5

~~**Issues addressed:**~~

**Tests performed:** No tests performed.

~~**Known issues and TODO list:** (add/remove lines as needed)~~

### DIFF
--- a/src/server/scripts/Spells/spell_generic.cpp
+++ b/src/server/scripts/Spells/spell_generic.cpp
@@ -2654,10 +2654,7 @@ class spell_gen_proc_below_pct_damaged : public SpellScriptLoader
 
                 int32 pct = GetSpellInfo()->Effects[EFFECT_0].CalcValue();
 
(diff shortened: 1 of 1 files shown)


jackpoz:
one thing that the old code allows is to easily put a breakpoint on only 1 result, so just in the false case or on the true case.
it also allows to add more cases in the future without having to rewrite any code.

Sorikoff:
I think that newer code is faster to read, and this operation happens more frequently than debugging, so code can be adjusted only when the debug is actually needed in a way developer wants that (additional blocks, new variables or conditional breakpoints).

illfated:
I appreciate your idea, but unfortunately, it does not fit into the TC strategy of keeping it simple to debug (as jackpoz said).

PR #23627, opened 2019-07-22
[3.3.5] Quest pooling rewrite

Treeston:
Partial PR of #23603 to make reviewing easier.

PR #23675, opened 2019-08-03
[3.3.5] range iterators

Treeston:
Because it's something I made for #23627, and there's no reason not to have it generalized in case we need it.


Warpten:
Awfully lacks comments tbf

msoky:
@Treeston is this finished and can be merged?

PR #17614, opened 2016-07-17
Core/Spells: Fixed warlock's Banish cancel if target was already bani…

Temppus:
**Changes proposed**:

Fixed removal of banish from banished targets.

**Target branch(es)**: 335/6x

**Issues addressed**: Closes #17581, but requires #17613 to be applied first.

**Tests performed**: Build OK + tested in game.

PR #17613, opened 2016-07-17
Core/Spells: Add SpellMissInfo argument to BeforeHit hooks.

Temppus:
**Changes proposed**:
- Title said it all.
  **Target branch(es)**: 335/6x

**Issues addressed**: Partially solve one problem with #17581.

**Tests performed**: Build OK+ tested in game.

Issue #18479, opened 2016-12-14
[3.3.5] Spell/Druid: Cyclone on targets under cyclone effects should be immune


Noryad:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:**
Cyclone: Tosses the enemy target into the air, preventing all action **but making them invulnerable** for up to 6 sec...

**Current behaviour:** 
In current revision when the target is under effects of cyclone and the druid casts cyclone again, the target is affected.

**Expected behaviour:** 
The targets under cyclonic effects should be immune to cyclone while the effect is active.

**Steps to reproduce the problem:**
1. Druid casts cyclone to target under cyclonic effects.
2. Target is affected again.

**Branch(es):** 3.3.5 

**TC rev. hash/commit:**  e1a2b1a

**TDB version:**  TDB_full_world_335.62_2016_10_17 + updates

**Operating system:**  Windows

[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing C++, create a Pull Request.)

ariel-:
Part of 560c882b364630bf95e1c6859ceaa6fa3cb6ed2e...

If IsRankOf is removed, breaks Banish (because effect shouldn't be immune to itself), if it left it breaks Cyclone (because effect SHOULD be immune to itself)

Now, comparing both Cyclone and Banish, the only differences are:
Dispel Type: Cyclone has NONE; Banish has MAGIC

Spell Attributes: 
Cyclone 
Attributes: 0x00010000 (SPELL_ATTR0_NOT_SHAPESHIFT)
AttributesEx4: 0x20000800 (SPELL_ATTR4_UNK11, SPELL_ATTR4_UNK29)

Banish
Attributes: 0x40010000 (SPELL_ATTR0_NOT_SHAPESHIFT, SPELL_ATTR0_HEARTBEAT_RESIST_CHECK)
AttributesEx2: 0x00000040 (SPELL_ATTR2_CANT_TARGET_TAPPED)

PR #23697, opened 2019-08-08
Core/Spells: Fixed warlock's Banish cancel if target was already banished

ForesterDev:
**Changes proposed:**

-  Core/Spells: Add SpellMissInfo argument to BeforeHit hooks and call them also when the spell doesn't hit. (#17613)

(cherry picked from commit 8ff5b35)

-  Core/Spells: Fixed warlock's Banish cancel if target was already banished (#17614)

(cherry picked from commit 4587b5d)

**Target branch(es):**

- [x] 3.3.5

**Tests performed:** (Does it build, tested in-game, etc.) Build, tested

### DIFF
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -1346,7 +1346,7 @@ bool SpellInfo::CanPierceImmuneAura(SpellInfo const* auraSpellInfo) const
         // ...but not these (Divine shield, Ice block, Cyclone and Banish for example)
         if (auraSpellInfo->Mechanic != MECHANIC_IMMUNE_SHIELD &&
                auraSpellInfo->Mechanic != MECHANIC_INVULNERABILITY &&
-               (auraSpellInfo->Mechanic != MECHANIC_BANISH || (IsRankOf(auraSpellInfo) && auraSpellInfo->Dispel != DISPEL_NONE))) // Banish shouldn't be immune to itself, but Cyclone should
+               auraSpellInfo->Mechanic != MECHANIC_BANISH)
             return true;
     }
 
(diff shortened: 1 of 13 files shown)


src/server/game/Spells/SpellInfo.cpp
```diff
@@ -1346,7 +1346,7 @@ bool SpellInfo::CanPierceImmuneAura(SpellInfo const* auraSpellInfo) const
         // ...but not these (Divine shield, Ice block, Cyclone and Banish for example)
         if (auraSpellInfo->Mechanic != MECHANIC_IMMUNE_SHIELD &&
                auraSpellInfo->Mechanic != MECHANIC_INVULNERABILITY &&
-               (auraSpellInfo->Mechanic != MECHANIC_BANISH || (IsRankOf(auraSpellInfo) && auraSpellInfo->Dispel != DISPEL_NONE))) // Banish shouldn't be immune to itself, but Cyclone should
+               auraSpellInfo->Mechanic != MECHANIC_BANISH)
```
ForesterDev: Wihout this change, banish pierce throw banish, so script won't work

jackpoz: is this needed in master too ?

ForesterDev: Don't test in master, but I think - yes

jackpoz: could you check that there is no regression about https://github.com/TrinityCore/TrinityCore/issues/18479 and also that Cyclone still works the way it's supposed to ?

ForesterDev: image
Works fine

ccrs:
I know this is just a port from master, but this implementation is so weird..

also `Spell::_handle_immediate_phase()` is called before `Spell::DoProcessTargetContainer` inside `Spell::handle_immediate()`..
that means, this is called before any script hook
```c++
    // handle effects with SPELL_EFFECT_HANDLE_HIT mode
    for (uint32 j = 0; j < MAX_SPELL_EFFECTS; ++j)
    {
        // don't do anything for empty effect
        if (!m_spellInfo->Effects[j].IsEffect())
            continue;

        // call effect handlers to handle destination hit
        HandleEffects(nullptr, nullptr, nullptr, j, SPELL_EFFECT_HANDLE_HIT);
    }
```

so... for `SPELL_EFFECT_HANDLE_HIT` the default handler cannot be overriden with `BeforeHit` script hook... ? makes any sense?
@Shauren

PR #23700, opened 2019-08-09
[3.3.5] Social Links rewrite

Treeston:
Because I don't have enough PRs open yet.

Current SocialMgr immediately saves everything to database on change. That's bad.
It currently also iterates over all online players whenever somebody logs in or out. That's also pretty bad.

I toss the whole thing and rewrite it, and yes this is starting to be a bit of a trend by now.

SocialMgr now stores all social links authoritatively, which means it can also consistently maintain reverse lookups despite DB data only being saved periodically. Each player has pointers to its own maps so we don't even need to do hashmap lookups most of the time.

Enjoy.


ccrs:
> I toss the whole thing and rewrite it, and yes this is starting to be a bit of a trend by now.

👍

src/server/game/Entities/Player/SocialMgr.cpp
```diff
@@ -17,319 +17,332 @@
...
+    ASSERT(isFriend || (flag == SOCIAL_FLAG_IGNORED), "Invalid social flag %u to AddSocialLink", uint32(flag));
 
-    if (!itr->second.Flags)
+    WorldSession* playerSession = player->GetSession();
+    FriendsResult result;
+    Player const* targetPlayer = nullptr;
+    CharacterCacheEntry const* target = sCharacterCache->GetCharacterCacheByName(targetName);
     {
-        PreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_DEL_CHARACTER_SOCIAL);
+#define SEND_RESULT(v1,v2) do { result = (isFriend ? v1 : v2); goto send; } while(0);
```
ccrs: ummm what about defining a std::function localy? that goto should burn 🔥

src/server/game/Entities/Player/Player.h
```diff
@@ -2386,7 +2389,9 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         uint32 m_questRewardTalentCount;
 
         // Social
-        PlayerSocial* m_social;
```
Shauren: This is a bad change, don't bloat `Player` class

lachtanek: Maybe you could've used the old PlayerSocial, just improve/refactor/optimize it, seems better to me than throwing new methods/properties onto this huge thing that calls itself Player

jackpoz:
https://github.com/TrinityCore/TrinityCore/issues/23825 might be related to this PR

src/server/game/Entities/Player/Player.h
jackpoz: Marking this PR as Feedback-FixOutdatedMissingWIP until this change is reverted

Shauren:
Closing as abandoned and known to cause crashes, #23825

PR #23705, opened 2019-08-10
Core/Spells: fixed a edge case in spell focusing for channeled spells

Ovahlord:
**Changes proposed:**

-  Fixes channeled spells getting interrupted by the focus release when being cast within the focus release delay and if the channeled spell has no focus target on its own (self channeled casts)

- [x] 3.3.5
- [ ] master

- Tested ingame


Ovahlord:
another pr will follow later that will include more fixes for pretty much all focus cases

PR #23719, opened 2019-08-16
[3.3.5] Fix spell_area not checking for zoneID for quests

### LINKED ISSUE #23718: [3.3.5] area_spell is broken for quests
**Description:**

There are some area spells which are actually using **ZoneID** instead of **AreaID**
For example take the quest "Into the Realm of Shadows" (ID: 12687), when you accept the quest it supposed to cast spell id 52693 instantly but it will not, because in spell_area table it used zoneID (**and it has to** use zoneID because area is so small for that quest)

**Current behaviour:**

SendQuestUpdate only searches for spells based on areaID.
```cpp
void Player::SendQuestUpdate(uint32 questId)
{
    uint32 zone = 0, area = 0;
    GetZoneAndAreaId(zone, area);

    SpellAreaForQuestAreaMapBounds saBounds = sSpellMgr->GetSpellAreaForQuestAreaMapBounds(**area**, questId); // <-- why area only???

    if (saBounds.first != saBounds.second)
    {
        for (SpellAreaForQuestAreaMap::const_iterator itr = saBounds.first; itr != saBounds.second; ++itr)
        {
            if (!itr->second->IsFitToRequirements(this, zone, area))
                RemoveAurasDueToSpell(itr->second->spellId);
            else if (itr->second->autocast)
                if (!HasAura(itr->second->spellId))
                    CastSpell(this, itr->second->spellId, true);
        }
    }

    UpdateVisibleGameobjectsOrSpellClicks();
}
```

The commit that breaks the system and made this bug:
https://github.com/TrinityCore/TrinityCore/commit/d7e47de264a20dc689b4b30663c2de09613b791f#diff-5589ee47b3a7ad220cff105a8fe7eac3R15743

**Expected behaviour:**

It supposed to check for zoneID too, as it does in SpellArea::IsFitToRequirements function.
Proof: (https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Player/Player.cpp#L23933)
```cpp
void Player::UpdateZoneDependentAuras(uint32 newZone)
{
    // Some spells applied at enter into zone (with subzones), aura removed in UpdateAreaDependentAuras that called always at zone->area update
    SpellAreaForAreaMapBounds saBounds = sSpellMgr->GetSpellAreaForAreaMapBounds(newZone);
    for (SpellAreaForAreaMap::const_iterator itr = saBounds.first; itr != saBounds.second; ++itr)
        if (itr->second->autocast && itr->second->IsFitToRequirements(this, newZone, 0))
            if (!HasAura(itr->second->spellId))
                CastSpell(this, itr->second->spellId, true);
}

void Player::UpdateAreaDependentAuras(uint32 newArea)
{
    // remove auras from spells with area limitations
    for (AuraMap::iterator iter = m_ownedAuras.begin(); iter != m_ownedAuras.end();)
    {
        // use m_zoneUpdateId for speed: UpdateArea called from UpdateZone or instead UpdateZone in both cases m_zoneUpdateId up-to-date
        if (iter->second->GetSpellInfo()->CheckLocation(GetMapId(), m_zoneUpdateId, newArea, this, false) != SPELL_CAST_OK)
            RemoveOwnedAura(iter);
        else
            ++iter;
    }

    // some auras applied at subzone enter
    SpellAreaForAreaMapBounds saBounds = sSpellMgr->GetSpellAreaForAreaMapBounds(newArea);
    for (SpellAreaForAreaMap::const_iterator itr = saBounds.first; itr != saBounds.second; ++itr)
        if (itr->second->autocast && itr->second->IsFitToRequirements(this, m_zoneUpdateId, newArea))
            if (!HasAura(itr->second->spellId))
                CastSpell(this, itr->second->spellId, true);
}
```

1. create dk class
2. get quest from Salanar the Horsemen (Into the realm of shadows quest id: 12687)
3. you will see that the spell will not cast on you and you have to get that aura using gossip menu or change your area so it triggers onZone/AreaChange event

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

43c08b176115505b76c027667047f9a5bfcc4e11

**Operating system:** 

Win32 (x64)

P.S. such an important bug exists for a long time, did not anyone find it yet? WOW

0x416c69:
**Changes proposed:**

-  Fix spell_area not checking for zoneID for quests

- [x] 3.3.5
- [ ] master

Closes #23718 

It built.
Tested, working and fixed the bug.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -15956,17 +15956,27 @@ void Player::SendQuestUpdate(uint32 questId)
 
     if (saBounds.first != saBounds.second)
     {
-        uint32 zone = 0, area = 0;
+        bool dontRemoveAura = false;
+        uint32 zone = 0, area = 0, auraRemove = 0;
         GetZoneAndAreaId(zone, area);
 
         for (SpellAreaForQuestMap::const_iterator itr = saBounds.first; itr != saBounds.second; ++itr)
         {
             if (!itr->second->IsFitToRequirements(this, zone, area))
-                RemoveAurasDueToSpell(itr->second->spellId);
+                auraRemove = itr->second->spellId;
             else if (itr->second->autocast)
+            {
+                dontRemoveAura = true; // We don't want to remove this aura anymore because the player is inside the required area of a certain quest
+                                       // Another quest could have the same spell aura and because it doesn't fit the requirements (could not be in that area) it will try to remove the aura
+                                       // And that aura is required for another quest so we should not remove this aura
+
                 if (!HasAura(itr->second->spellId))
                     CastSpell(this, itr->second->spellId, true);
+            }
         }
+
+        if (auraRemove && !dontRemoveAura)
+            RemoveAurasDueToSpell(auraRemove);
     }
 
     UpdateVisibleGameobjectsOrSpellClicks();
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -15956,27 +15956,46 @@ void Player::SendQuestUpdate(uint32 questId)
 
     if (saBounds.first != saBounds.second)
     {
-        bool dontRemoveAura = false;
-        uint32 zone = 0, area = 0, auraRemove = 0;
+        std::set<uint32> aurasToRemove, aurasToCast;
+        uint32 zone = 0, area = 0;
         GetZoneAndAreaId(zone, area);
 
         for (SpellAreaForQuestMap::const_iterator itr = saBounds.first; itr != saBounds.second; ++itr)
         {
             if (!itr->second->IsFitToRequirements(this, zone, area))
-                auraRemove = itr->second->spellId;
+                aurasToRemove.insert(itr->second->spellId);
             else if (itr->second->autocast)
-            {
-                dontRemoveAura = true; // We don't want to remove this aura anymore because the player is inside the required area of a certain quest
-                                       // Another quest could have the same spell aura and because it doesn't fit the requirements (could not be in that area) it will try to remove the aura
-                                       // And that aura is required for another quest so we should not remove this aura
-
-                if (!HasAura(itr->second->spellId))
-                    CastSpell(this, itr->second->spellId, true);
-            }
+                aurasToCast.insert(itr->second->spellId);
         }
 
-        if (auraRemove && !dontRemoveAura)
-            RemoveAurasDueToSpell(auraRemove);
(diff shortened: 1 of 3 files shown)


Killyana:
Check if there's a regression for the 2 issues fixed by https://github.com/TrinityCore/TrinityCore/pull/17509

0x416c69:
The issues are back by this PR of mine, however what #17509 tried to fix:

> So, we can iterate spell_auras by player's current area instead, since that's what we really want to update.That way, spells can safely be removed when the checks fail and it will work as intended™.

This is really a hacky way to fix this issue I think, it will fail if two quests are in the same place too (which I don't know if it exists or not)

I think the way to solve that problem is to prevent users to take quests which use the same spell in spell_area as the first one and ask them to abandon the first one if they want to take that quest (?)

Killyana:
I added this at line 15957-15958:
``` cpp
    UpdateAreaDependentAuras(area);
    UpdateZoneDependentAuras(zone);
```
without adding anything and the issue looks to be fixed, still need to check if there's a regression.

0x416c69:
No, that's really hacky also unnecessary processing because UpdateArea/ZoneDependentAuras have to be triggered when Zone/Area is changed. I already fixed the issue. I will push now.

0x416c69:
I tested it, quest A Cold Front Approaches is working perfect now.

illfated:
Looks good to me, although it needs to be reviewed by real coders.

0x416c69:
Now that I think more, I think my code also creates another bug.

What if one quest has 2 spells and it updates, 1 spell need to be removed and 1 spell need to remain be casted so in this scenario:

It will try to remove aura n1: Player does not fit so we should remove
And then it will try to remove aura n2: Player fit so we should not remove any aura

aura n1 and n2 are two different spells so it's fine for code to do its old same thing so I'm gonna make another modification to my PR.

0x416c69:
There y'all go, seems perfect to me now.

illfated:
While not being a real coder, I would like to offer a suggestion for the comment text in Player.cpp:
```cpp
        // Auras matching the requirements will be inside the aurasToCast container.
        // Auras not matching the requirements may prevent using auras matching the requirements.
        // aurasToRemove will erase conflicting auras in aurasToCast to handle spells used by multiple quests.
```
This is purely a suggestion, you can use the content however you want to, either as an inspiration for modifications, partial or complete usage.
(if the suggested text contains some wrong wording due to incorrect assumptions, feel free to modify or skip the change)

jackpoz:
@Killyana do you have any other question ? how does this PR look ?

Killyana:
Looks fine.

jackpoz:
Thanks for the PR :)

PR #23740, opened 2019-08-25
DB/Creature: Kurzen Commando

Nightprince:
Fix spell id Kurzen Commando for hide, spell id 30831 is wrong and not work must 22766

https://www.wowhead.com/npc=938/kurzen-commando#abilities


Aokromes:
we don't use pull requests for sql only fixes, plz open a ticket, also we don't put ' around numbers.

PR #23757, opened 2019-09-03
Scripts/Commands: Allow .cheat god to target other players

Artamedes:
**Changes proposed:**

-  Allow cheat god to target other players, useful for testing bosses without other gm chars

- [x] 3.3.5

Builds, works ingame

- [x] must be test!

### DIFF
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -35,6 +35,7 @@
 #include "RBAC.h"
 #include "Transport.h"
 #include "Vehicle.h"
+#include "World.h"
 #include "WorldPacket.h"
 #include "WorldSession.h"
 
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -16,23 +16,28 @@
  * with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "AccountMgr.h"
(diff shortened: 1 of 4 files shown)


jackpoz:
This PR will allow any GM to make their friends invincible. Please consider putting it in a RBAC disabled by default

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -204,6 +210,27 @@ void WorldSession::HandleMoveWorldportAck()
 
     //lets process all delayed operations on successful teleport
     player->ProcessDelayedOperations();
+
+    if (player->GetCommandStatus(CHEAT_GOD) && !HasPermission(rbac::RBAC_PERM_COMMAND_CHEAT_GOD))
```
jackpoz: what does this part actually do ? it wasn't included in the PR description

Shauren:
I don't think this is a good idea

jackpoz:
Something I often use when I don't want to die but don't want to use .cheat god is https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=30839 together with https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?item=12947

jackpoz:
Hi, thanks for the PR. Even if these changes can help a lot test instances, I'm not still convinced that we want to have players running around with ".cheat god", nor have some logs that iterate all the current online players.
I would prefer to see the issue that this PR targets to be fixed in a different way.

PR #23779, opened 2019-09-08
Build/CMake: Set some policies to NEW

jackpoz:
**Changes proposed:**

-  Set some CMake policies to NEW. Each policy has its own commit

- [x] 3.3.5
- [ ] master

**Issues addressed:** None

**Tests performed:** Compared Visual Studio 2019 build files win WinMerge

- [ ] Check CI for any build issue


illfated:
Thank you for working on this. I was wondering how long it would take before it would be even looked at closer, instead of being dismissed in the issue tracker.

jackpoz:
By the way @illfated I wanted to thank you for playing on https://tcubuntu.northeurope.cloudapp.azure.com/ and testing issues there. Please let me know (just @ me) if you need any higher permission on the test server :)

illfated:
Thank you, I am doing just fine as it is. My time barely stretches to testing with GM powers, so I there's no need for more at the moment. But I will keep it in mind in case any unexpected problem comes up (no idea what that would be, though).

PR #23797, opened 2019-09-14
Core/Calendar: Add some additional validation when creating events

### LINKED ISSUE #14923: [3.3.5][6.x] Calendar events cap
- [ ] Implement cap of 30 events for player
- [ ] Implement cap of 100 events for guild
- [ ] Autodelete old events that are older than 30 days
- [ ] An event can be created only every 5 seconds

jackpoz:
**Changes proposed:**

- Allow only 30 player events and 100 guild events to be created.
- Don't allow to create guild events if player is not in guild.
- Send some more error messages to the client (not blizzlike errors but better than nothing).
- Autodelete old events that are older than 30 days
- An event can be created only every 5 seconds
- HandleCalendarCopyEvent should check that the creator is the same as the player copying it or that it's a guild event and the player is in the same guild
- Fix debug assertion triggered when deleting a character with calendar events
- Avoid double std::set lookup when deleting a character with calendar events for each event

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #14923

- This the message displayed if an event is created in the past (feel free to find a better message code)
image
- This is the message displayed if creating a Guild Event/Announcement while not in guild
image
- Created 30 events, received this at the 31st
image
- Tried to copy an event into the 31st one, received
image
- Created 100 guild events, received this at the 101st
image
- Tried to copy a guild event into the 101st one, received
image

- [x] Autodelete old events that are older than 30 days
- [x] An event can be created only every 5 seconds
- [x] HandleCalendarCopyEvent should check that the creator is the same as the player copying it or that it's a guild event and the player is in the same guild

### DIFF
--- a/src/server/game/Calendar/CalendarMgr.cpp
+++ b/src/server/game/Calendar/CalendarMgr.cpp
@@ -175,13 +175,24 @@ void CalendarMgr::RemoveEvent(uint64 eventId, ObjectGuid remover)
         return;
     }
 
+    RemoveEvent(calendarEvent, remover);
+}
+
+void CalendarMgr::RemoveEvent(CalendarEvent* calendarEvent, ObjectGuid remover)
+{
+    if (!calendarEvent)
+    {
+        SendCalendarCommandResult(remover, CALENDAR_ERROR_EVENT_INVALID);
+        return;
+    }
+
     SendCalendarEventRemovedAlert(*calendarEvent);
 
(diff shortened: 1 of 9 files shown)


jackpoz:
NB: The whole CalendarMgr/CalendarHandler code should be checked line by line for bugs/crashes/exploits.

jackpoz:
Please note next startup might take some time as old calendar events will be deleted

Issue #15989, opened 2015-12-12
[3.3.5] DB/Quest: Where in the World is Hemet Nesingwary? (12521)


Treeston:
Branch 3.3.5, revision 01beba21b5210860ddd6a37471c6b2bea77a369e

Quest [Where in the World is Hemet Nesingwary](http://www.wowhead.com/quest=12521/where-in-the-world-is-hemet-nesingwary)

The [Archmage Penturus' Flying Machine](http://www.wowhead.com/npc=28192/archmage-pentarus-flying-machine) that you enter on the Sholazar Basin side of the portal is incredibly, incredibly slow. It takes over 10 minutes to get to the quest waypoint.

Killyana:
The problem is: sometimes the teleportation of the vehicle fails, in this case the waypoint starts from Dalaran to Sholazar, even if you're seeing the flying machine in Sholazar it's not the real position, if the speed it too slow look to the map to see the real position of the flying machine.

Treeston:
Still an issue on 8e6fb3b1c50f

Oh, and you don't get a parachute at the end, meaning you die. Horribly.

genion18:
Hy,

Very Horribly, the Flights end here and when logout/login is the Charakter at the Air in Dalaran to the death flight at the ground.

greets

genion18:
Hy,

Very Horribly, the Flights end here and when logout/login is the Charakter at the Air in Dalaran to the death flight at the ground.

greets

genion18:
Hy,

Very Horribly, the Flights end here and when logout/login is the Charakter at the Air in Dalaran to the death flight at the ground.

greets

![WoWScrnShot_080619_222346](https://user-images.githubusercontent.com/321832/62575365-dfc1b880-b89a-11e9-9b1b-cab6675b0a8b.jpg)

Duratans:
confirm, sometimes you end up completely in a weird location flying in the abyss.,

jackpoz:
Fixed by https://github.com/TrinityCore/TrinityCore/commit/afbc47891965f9c31c4b31c66931a2dbc9846f93

PR #23807, opened 2019-09-20
Fixed possible desync issues caused by vehicles teleport

sanctum32:
**Changes proposed:**

-  Fixed desync issues when vehicle is teleported with seat acessories which are filled by players  

**Target branch(es):** 

- 3.3.5

**Issues addressed:** 

When vehicle is ported and if player uses any of its seats, both units are teleported but just visual, their real positions remains same. Best practise to reproduce this is by taking quest https://www.wowhead.com/quest=12521/where-in-the-world-is-hemet-nesingwary and by calling this quest vehicle event. 
This change will make sure that grid is loaded before units are teleported. Desync happened because units were attempting to have teleport before destination grid activates (if grid was not activated before)

**Tests performed:** 

build, tested in-game

Note: its my first PR on this project

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12663,6 +12663,9 @@ void Unit::NearTeleportTo(Position const& pos, bool casting /*= false*/)
     }
     else
     {
+        if (!GetMap()->IsGridLoaded(pos.GetPositionX(), pos.GetPositionY()))
+            GetMap()->LoadGrid(pos.GetPositionX(), pos.GetPositionY());
+
         SendTeleportPacket(pos);
         UpdatePosition(pos, true);
         UpdateObjectVisibility();


jackpoz:
was it failing in Map::CreatureRelocation() because of !getNGrid(new_cell.GridX(), new_cell.GridY()) check ?

sanctum32:
@jackpoz i added 
`
   if (!respawnRelocationOnFail && !getNGrid(new_cell.GridX(), new_cell.GridY()))
   {
        printf("FAILED Map::CreatureRelocation\n");
        return;
    }`
IT wasn't triggered

jackpoz:
Could you find the place in the code where it returns early instead of loading the grid ? Maybe that would be a better place for the fix

sanctum32:
hm, long debugging process, alright, I will
btw since my english is terrible, i made video how this issue goes https://www.youtube.com/watch?v=y0Zsh5P-bzQ&feature=youtu.be

Killyana:
Fixes: https://github.com/TrinityCore/TrinityCore/issues/15989

sanctum32:
I made test of
`void Map::PlayerRelocation(Player* player, float x, float y, float z, float orientation)
{
    ASSERT(player);

    Cell old_cell(player->GetPositionX(), player->GetPositionY());
    Cell new_cell(x, y);

    player->Relocate(x, y, z, orientation);
    printf("Map::PlayerRelocation call 1\n");
    if (player->IsVehicle())
    {
        printf("Map::PlayerRelocation call 2\n");
        player->GetVehicleKit()->RelocatePassengers();
    }

    if (old_cell.DiffGrid(new_cell) || old_cell.DiffCell(new_cell))
    {
        TC_LOG_DEBUG("maps", "Player %s relocation grid[%u, %u]cell[%u, %u]->grid[%u, %u]cell[%u, %u]", player->GetName().c_str(), old_cell.GridX(), old_cell.GridY(), old_cell.CellX(), old_cell.CellY(), new_cell.GridX(), new_cell.GridY(), new_cell.CellX(), new_cell.CellY());

        player->RemoveFromGrid();

        printf("Map::PlayerRelocation call 3\n");
        if (old_cell.DiffGrid(new_cell))
        {
            EnsureGridLoadedForActiveObject(new_cell, player);
            printf("Map::PlayerRelocation call 4\n");
        }

        printf("Map::PlayerRelocation call 5\n");
        AddToGrid(player, new_cell);
    }

    player->UpdatePositionData();
    player->UpdateObjectVisibility(false);
    printf("Map::PlayerRelocation call 6\n");
}`
1 and 6 was at output, still debugging (will edit this msg)

jackpoz:
https://www.codeproject.com/Articles/79508/Mastering-Debugging-in-Visual-Studio-2010-A-Beginn gives some nice tips to learn how to debug without relying on printf

sanctum32:
stopped using visual studio like 1 year now since it was crashing 24/7 for me, Since 2019 version i'm getting constant crashes on it (windows was reinstalled with all tools by several times, crashes mostly are related to intelisence caching). Since then using visual tudio code/atom editors and doing msvc builds via command line. Still not sure how to do debugging in code/atom editors for c++ code, so using printfs for now. Anyways thanks for suggestion, i'll keep it in bookmarks.

sanctum32:
i'm unable to find any other call which would return unit from pre-activating cells. At this moment only guess is - grid activation is just missing before unit relocation...

jackpoz:
Before merging this PR I would like to know more about why the current code doesn't work and if there is a better place where we can load the grid instead of Unit::NearTeleportTo()

jackpoz:
@sanctum32 were you able to fix your development environment to debug the code ?

sanctum32:
I'll do at 2019-11-02 once after job, yes @jackpoz i fixed remaining issues with visual studio, update to latest version fixed most of issues (rarelly turning them since they tend to break things usually, if it works, it works)

jackpoz:
anyone had time to debug why the current 3.3.5 code doesn't handle this case ?

jackpoz:
Map::CreatureRelocation() simply calls AddCreatureToMoveList() on the vehicle (which is actually a Minion), while Unit::SendTeleportPacket() is called right away.
Map::CreatureCellRelocation() then doesn't load any grid because the Minion is just a simple creature (even if IsCharmedOwnedByPlayerOrPlayer() returns true, but the code doesn't check that).

jackpoz:
Something like https://github.com/jackpoz/TrinityCore/commit/daf0c26024ac90008d643bce4c8d167a59a6a66a should fix the issue. Is anyone available to test it ?

jackpoz:
This PR has been replaced by https://github.com/TrinityCore/TrinityCore/commit/afbc47891965f9c31c4b31c66931a2dbc9846f93 which is a more generic fix.
Thanks for the investigation done about the issue

PR #23817, opened 2019-09-22
Core/Movement: provide an extra spline path vertex for taxi pathings …

Ovahlord:
**Changes proposed:**
-  added an additional spline path vertex for the spline system to use for calculating player start positions so we use correct pathings now and wont lose the first taxi path vertex for the player starting position. This effectively and correctly fixes the crash that was intended to be fixed by 680e1cbd6cf7c52abe104fb87d8fa6dba83fb633 which is only safeguarding the core from a mistake.

Oh, and on top of that taxi pathings such as the one starting at stormwing will no longer clip through walls because of that. You're welcome.

- [x] 3.3.5
- [x] master

- tested ingame, works for me (tm)

### DIFF
--- a/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
@@ -75,7 +75,7 @@ void FlightPathMovementGenerator::DoReset(Player* owner)
     owner->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_REMOVE_CLIENT_CONTROL | UNIT_FLAG_TAXI_FLIGHT);
 
     Movement::MoveSplineInit init(owner);
-    // Providing an starting vertex since the taxi paths do not provide such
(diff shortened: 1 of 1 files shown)


jackpoz:
I have 1 case where this should be checked:
- player starts a taxi
- player joins a battleground
- player gets teleported inside the battleground
- player leaves the battleground
- TC code checks what's the closest taxi node to make the player continue the taxi. A node before/behind player position is chosen

With the previous code, if the selected node would be before/behind the player was chosen, that node would be replaced by current player position. With this PR code will the taxi be "1) current position 2) before/behind node 3) next node" ?

Ovahlord:
The player will continue his path at his current position before he joined the battleground as it suposed to be. All I did with this change was feeding the spline system with the necessary waypoint data so it can do its work even if we are one path vertex away from reaching our taxi node.

jackpoz:
Thanks for the PR :)

jackpoz:
Might have caused https://github.com/TrinityCore/TrinityCore/issues/23829

Issue #23829, opened 2019-09-26
Visual bug on last flight master point


TerrorBringer:
**Description:**

When you use a flight master taxi, on the last destination (when you land), the flying taxi does not come near the floor, you just go from mid-air to spawning besides the Flight Master.

Probably caused by this commit https://github.com/TrinityCore/TrinityCore/commit/cf3277c641bd877b9fa1939da5da7cde1778561b

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/c8dc9fdc418d9703893e21603b7593b015f08fea

**Operating system:** Win 10 x64

jackpoz:
could you try reverting that commit locally and checking the behavior ?

Killyana:
It's the correct behavior implemented by https://github.com/Wyrserth/TrinityCore/commit/aeed80b71b3e0c10c8364ee3a15c343f4bd02e46
Once at the last node the player will be teleport to the ground instead of falling.

Killyana:
@TerrorBringer in which fly you noticed that?

illfated:
> Edit: After [cf3277c](https://github.com/TrinityCore/TrinityCore/commit/cf3277c641bd877b9fa1939da5da7cde1778561b) the teleportation is done before reaching the last node

Hm, interesting. Does it need an additional spline to reach the destination flightpoint, then?

jackpoz:
@Ovahlord can you take a look at this ?

Ovahlord:
That's not caused by this. The change does not affect the destination of a spline. Kilyana already named what's causing this and apparently this seems to be intended behaivior. Nothing wrong here on my side

TerrorBringer:
In https://github.com/TrinityCore/TrinityCore/commit/8e3862cfbd9ee677866fbbee60c3b96ee3f4868a, there is no issue in the last node.

Shauren:
It actually does affect the destination, if you look at `FlightPathMovementGenerator::DoUpdate`, it increments `_currentNode` based on points given to spline init but runs MovementGenerator finalize logic based on `_path` which does not have any extra points (returning false from DoUpdate)
Result of that is that spline has correct destination but MovementGenerator interrupts it in DoFinalize

Killyana:
@Ovahlord after https://github.com/TrinityCore/TrinityCore/commit/cf3277c641bd877b9fa1939da5da7cde1778561b the last node is skipped.

[...]

PR #23830, opened 2019-09-26
Core/Movement: fixed taxi pathings getting finalized earlier than intended

Ovahlord:
**Changes proposed:**

-  followup fix for cf3277c641bd877b9fa1939da5da7cde1778561b discussed in https://github.com/TrinityCore/TrinityCore/issues/23829

- [x] 3.3.5

### DIFF
--- a/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
@@ -94,8 +94,9 @@ bool FlightPathMovementGenerator::DoUpdate(Player* owner, uint32 /*diff*/)
     if (!owner)
         return false;
 
-    uint32 pointId = owner->movespline->currentPathIdx() < 0 ? 0 : owner->movespline->currentPathIdx();
-    if (pointId > _currentNode && _currentNode < _path.size())
+    // skipping the first spline path point because it's our starting point and not a taxi path point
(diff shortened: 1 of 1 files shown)


jackpoz:
Tested ingame etc ?

Ovahlord:
Kilyana tested it and said it seems to be fine.

jackpoz:
does https://github.com/TrinityCore/TrinityCore/pull/23830/files#diff-d0720ce3f8edece58e8e6e5483a786beR126 need to be changed too ? or why did the array range check change to _currentNode < _path.size() - 1 ?

Ovahlord:
Since we now deal with a path that is 1 point larger than intended we also have adapt the path.size changes accordingly so the checks will remain intact otherwise we risk getting out of range errors

jackpoz:
If _currentNode is used as index of path,  it doesn't matter how you got it,  with or without the additional start point, < path.size() condition will always stop all out of range cases

Ovahlord:
yes, and with the new spline point _currentNode will be incremented also when we pass the first spline point that is not a actual taxi path point which leads to the problem that we trigger the finalization earlier than intended and all the incremental path size checks also need to be changed accordingly what I just did. Example: prior to my addition we had 13 path points and now since we consider the player position, we have 14 instead. So path.size() - 1 is covering that problem up so all checks will do their old job again.

jackpoz:
so this is ready to be merged, latest version tested too ?

jackpoz:
Thanks for the PR :)

PR #23835, opened 2019-09-28
Core/Taxis: teleport players to the destination taxi node location instead of their current ground position

Ovahlord:
**Changes proposed:**
-  the initial implementation in 6a0a800535f6de9f6028a5375abfaf1dae8c7c28 is wrong. Taxis use the node DBC coordinates as teleport destination instead. Clear evidence can be found here: https://youtu.be/xtXePB4TVbo?t=55

- [x] 3.3.5

- Tested ingame, perfectly identical result: https://www.youtube.com/watch?v=mWRi-bTuVig

### DIFF
--- a/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/FlightPathMovementGenerator.cpp
@@ -161,7 +161,7 @@ void FlightPathMovementGenerator::DoFinalize(Player* owner, bool active, bool/* 
         if (TaxiNodesEntry const* node = sTaxiNodesStore.LookupEntry(taxiNodeId))
         {
             owner->SetFallInformation(0, node->z);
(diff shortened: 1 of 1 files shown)


Killyana:
I tested this PR and it works fine

Ovahlord:
Btw I tested this on the ptr a few days ago, it's still the same behaivior so you may merge this :)

Warpten:
Friendly reminder that wrath != live

Ovahlord:
Whelp, in the end I've just corrected a already implemented feature. I've checked multiple videos and sniffs for all kinds of versions and either classic lets you drop from where the flight path ends up to you get teleported to the destination taxi node. the current implementation does neither of it. It straightup uses a vmap lookup to teleport you to the ground which I haven't seen in a single cases

Killyana:
Here proof from retail:
2006 vanilla: https://youtu.be/jo-cfBB4BY8?t=350
2010 wotlk: https://youtu.be/tWr7tjXY8FE?t=260

PR #23842, opened 2019-10-01
[3.3.5] Core / MMAPS Float Value

JustZerooo:
**Changes proposed:**

-  Changing the float value to a non .f;
-  
-  

- [X] 3.3.5
- [ ] master

 - None

- None

- None
- [ ] 
- [ ]

### DIFF
--- a/src/tools/mmaps_generator/MapBuilder.h
+++ b/src/tools/mmaps_generator/MapBuilder.h
@@ -95,7 +95,7 @@ namespace MMAP
     class MapBuilder
     {
         public:
-            MapBuilder(float maxWalkableAngle   = 70.f,
+            MapBuilder(float maxWalkableAngle   = 70.0f,
                 bool skipLiquid          = false,
                 bool skipContinents      = false,
                 bool skipJunkMaps        = true,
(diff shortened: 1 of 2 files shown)


Shauren:
But why? This does nothing

TheNeoTV:
For correct coding standarts and normalized float values. Some of you need corect of all work so do it corect.

MaxtorCoder:
> 
> 
> For correct coding standarts and normalized float values. Some of you need corect of all work so do it corect.

Still makes no sense to do it lol

TheNeoTV:
it makes sense to see you do correct code work and not half work. floats are in general used as 123.456f and not as 123.f it makes sense to have a complete float and not a half float

MaxtorCoder:
And it makes no sense to do `1.0f` since `1.f` is faster.

jackpoz:
Thanks for the PR.
Could you please update the PR description with a more detailed description about the reason behind these changes ? As you mentioned coding standards, which coding standards ? Is writing 70.f a bad practice ?

Meltie2013:
Adding a zero at the end doesn't make any difference, besides the readability by the human readers. Compilers don't really care if its `1.f` or `1.0f`. Works the same way to my knowledge.

This is more of a code styling PR then a PR to fix something.

jackpoz:
I personally find 70.f clearer as it states we don't care about specifying the first decimal (and for me 70 is 70.00000000 already without having to write any 0). The .f part is just a programming-specific notation that we are forced to use when using division in these cases (and to avoid warnings about double conversion if we use 70.0 instead).

This PR will probably be closed but I was still interested into the reason behind it.

PR #23864, opened 2019-10-09
Core/Movement: implement smooth waypoint transition

Ovahlord:
**Changes proposed:**
-  implement missing handling to make a creature smoothly transition from one waypoint path into another. Retails tends to use the first spline point of the following waypoint path in the prior path to make transitions smoother without appearing laggy. This PR is literally doing this by calculating the path of the next waypoint in advance and picking the first spline point from it for the current waypoint path.

- [x] 3.3.5

Tested ingame: https://www.youtube.com/watch?v=osnXEEXwUH4

The small hickup you can see at the npc Thomas Miller on the left side is not an issue of the smooth waypoint transition but a issue with the path generator that tends to occasionally return stupid spline points which is an issue outside of the scope of this PR.

### DIFF
--- a/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
@@ -29,7 +29,7 @@
 #include "Transport.h"
 #include "WaypointManager.h"
 
-WaypointMovementGenerator<Creature>::WaypointMovementGenerator(uint32 pathId, bool repeating) : _nextMoveTime(0), _transitionPointId(0), _pathId(pathId), _repeating(repeating), _loadedFromDB(true)
+WaypointMovementGenerator<Creature>::WaypointMovementGenerator(uint32 pathId, bool repeating) : _nextMoveTime(0), _pathId(pathId), _transitionPointId(0), _repeating(repeating), _loadedFromDB(true)
 {
     Mode = MOTION_MODE_DEFAULT;
     Priority = MOTION_PRIORITY_NORMAL;
@@ -37,7 +37,7 @@ WaypointMovementGenerator<Creature>::WaypointMovementGenerator(uint32 pathId, bo
     BaseUnitState = UNIT_STATE_ROAMING;
 }
 
-WaypointMovementGenerator<Creature>::WaypointMovementGenerator(WaypointPath& path, bool repeating) : _nextMoveTime(0), _transitionPointId(0), _pathId(0), _repeating(repeating), _loadedFromDB(false)
+WaypointMovementGenerator<Creature>::WaypointMovementGenerator(WaypointPath& path, bool repeating) : _nextMoveTime(0), _pathId(0), _transitionPointId(0), _repeating(repeating), _loadedFromDB(false)
 {
     _path = &path;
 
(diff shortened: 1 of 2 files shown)


jackpoz:
I'm curious about the pathfinding part. Maybe it uses pathfinding in a bad way:
- does it try to get a path from each waypoint to next ?
- does it expect it to be a straight line without setting the straight line flag ?

if you have some how to reproduce steps, please open an issue and I might take a look into it

Ovahlord:
From what I can see that hickup is caused by asynch clientside / serverside spline handling. The unit turns arround because the serverside position is behind the clientside one I assume because there is currently nothing in place that keeps the serverside and clientside spline synch. I assume that the initial spline point shouldn't exactly be the creature position but instead a position that's a bit ahead

jackpoz:
Which line are you referring to ?

xXBentRealityXx:
Tried importing this to latest TC 3.3.5 branch and now they just twitch before they make a sharp turn

jackpoz:
@Ovahlord with "path generator" you mean "Movement::MoveSplineInit" ?

Ovahlord:
No. Literally PathGenerator itself. It uses a unit object as owner merely for checking unit states and liquid status. Making the path generator unit independend from units it will make it much easier to handle. could be changed to pass unit states and liquid status via arguments instead

msoky:
can be merged please?

Killyana:
This is not done yet.

Seyden:
@joschiwald didn't we already implemented all this stuff for 4.3.4? I think we even fixed the event script notification bugs. There was no hiccups and i think everything for that to work was already implemented. I wonder what the big spline initialization is for.

@Ovahlord btw you can't implement it for every waypoint in the game, back in 4.3.4 i researched if every waypoint was smooth like that, but there were a lot of creatures and npcs who didn't use smooth curves.

Ovahlord:
There are npc's ye, but they are handled quite simple: they have a tiny delay between their points to prevent a smooth transition. You can see this when parsing waypoints and compare their spline duration against the timestamp of the next movement.
That's pretty much it.

jackpoz:
> Making the path generator unit independend from units it will make it much easier to handle

There, I did that :)

jackpoz:

Ovahlord:
not ready. The current systems are not supporting the necessary circumstances to get a proper result. Aside from that the the spline system has yet another bug that it's falling behind the serverside position of the npc from time to time (which causes these turnarround hickups)

Krejza9:
Any news after a year ?

Nyr97:
This was fixed almost 2 months ago by @Ovahlord on the 4.3.4 branch. The only problem now consists in having to sniff the delay of each waypoint so that the transition is blizzlike, which is a task not everyone can do right now. Though adding custom delays works.

Krejza9:
> 
> 
> This was fixed on November with some additions and corrections over time by @Ovahlord on the 4.3.4 branch [The-Cataclysm-Preservation-Project@d5f9b8c](https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/commit/d5f9b8c26f50e2fa085de28d8da1cc4c169ea2ae). The only problem now consists in having to sniff the delay of each waypoint so that the transition is blizzlike, which is a task not everyone can do right now. Though adding custom delays works.

but what about version 3.3.5 ?

funjoker:
Dead

offl:
> I can't imagine Blizz adds those delays -1312 -155 -520 -763 234 525 862 924 1402 1230 to every waypoint

Random movement too is smooth, just saying

PR #23877, opened 2019-10-20
Core/Spells - Rogue: Fix Unfair Advantage range

### LINKED ISSUE #23873: DB/Spell | Talent : Rogue's Unfair Advantage (51672) does not counter-attack.
**Description:**

Talent Unfair Advantage does not trigger.

**Current behaviour:**

If you learned Talend Unfair Advantage and evade an enemies attack, Unfair Advantage nearly never triggers. I only got 1 or 2 situations where it proccs - no matter how often i evade attacks.
.list aura does show it correctly.

**Expected behaviour:**

Should work as described - evading an attack should result in a counter-attack in 100% of all cases with Talent rank 2 but only once per second.

Level a rogue
Learn Unfair Advantage Rank 2
Fight a mob
Watch combat

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

30823f9b79b4ad16f96e90692a900f68a7f63c0a

Jildor:
**Changes proposed:**

-  Fix range of spell, if target distance is > 0 yards, Unfair Advantage can't hit with a range of 0 yards

- [X] 3.3.5
- [ ] master

**Issues addressed:** Closes #23873

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -3942,6 +3942,14 @@ void SpellMgr::LoadSpellInfoCorrections()
     });
 
     ApplySpellFix({
+        51675,  // Rogue - Unfair Advantage (Rank 1)
+        51677   // Rogue - Unfair Advantage (Rank 2)
+    }, [](SpellInfo* spellInfo)
+    {
+        spellInfo->RangeEntry = sSpellRangeStore.LookupEntry(2); // 5 yards
+    });
+
+    ApplySpellFix({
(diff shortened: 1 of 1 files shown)


jackpoz:
should the triggered spell ignore range checks in this case ?

Jildor:
I think 5 yards it's enough, because it's melee hit

Shauren:
No, triggered spells should not ignore range checks (this was verified on retail with various trinket procs)

jackpoz:
Thanks for the PR :)

PR #23895, opened 2019-10-28
Core: added aarch64 definition for jemalloc

damaex:
**Changes proposed:**

-  aarch64 definition for jemalloc
    - as defined by jemalloc on raspberry pi

- [x] 3.3.5

- [x] build
- [x] in game (on raspberry pi 4)


jackpoz:
I would prefer to move the architecture checks to https://github.com/TrinityCore/TrinityCore/blob/b6d3708a220766e54a7fc1447f333745d5afbb9e/dep/jemalloc/CMakeLists.txt#L15 , similar to https://github.com/TrinityCore/TrinityCore/commit/10ad43916fb67eaf59c5a676c83b36ec8506beec

illfated:
Just out of curiosity; is it a visual Github bug that the code looks like it has been commented out (gray) instead of showing black text or syntax highlighting?

damaex:
> Just out of curiosity; is it a visual Github bug that the code looks like it has been commented out (gray) instead of showing black text or syntax highlighting?

The file is `.h.in.cmake` and github doesn't know how to handle C++ Header in CMake files.

illfated:
Aha. I forgot that bit. Thanks for refreshing my memory. 👍

jackpoz:
is there some difference between __arm and __aarch64__ ? Which one should we target ? Should we move the platform checks to https://github.com/TrinityCore/TrinityCore/tree/master/cmake/platform so we can reuse it ?

The jemalloc-specific code itself looks good :)

Issue #21615, opened 2018-03-13
[3.3.5] Core/Spells: Green Buff in WSG does nothing


VincentVanclef:
**Description:**

The buff applies from the gameobject, to the player, but the buff itself does nothing, no heal, no anything. You can see in combat log nothing is displayed at all.

Image: http://prntscr.com/iqqg6t

I get the buff, animation, but buff does nothing.

**Expected behaviour:**

should restore some health and mana

**Steps to reproduce the problem:**

1. Walk into green buff in WSG
2. Notice that the character does not get any spell benefits

**Branch(es):**  3.3.5

**TC rev. hash/commit:**  https://github.com/TrinityCore/TrinityCore/commit/28288698441524081d583af95f19cb1c9fb90387

**TDB version:**  335.64

**Operating system:** ubuntu

Persephone:
@ariel- Is this related to gameobject casting?

Persephone:
@ariel-  Nope tested casting the buff on myself (23493), nothing is happening i do get visual but it aint healing

Killyana:
It could be related to the faction not set correctly to gobs that are not spawned (summoned, or temp spawn like BG spawns)
If you .gob add 180147 the gob will cast the spell on you char
If you .gob add temp 180147 the gob will not cast the spell

Persephone:
Its not related to the other issue xD ffs! Already explained this. Green buff script is broken has nothing to do with stealth! Reopen the other

Killyana:
Chickens https://woehead.way-of-elendil.fr/?npc=620 can get the auras from this gobs.

Hir0shi:
@confirm

please if can fixed!

old time is buged this issue

Killyana:
Caused by https://github.com/TrinityCore/TrinityCore/commit/34967e9c32b2c7e871bb93f41609a3b08ad92931#diff-fe5c8ebb7ad8e663836b00f39eb524d6

Jildor:
So, when will @ariel- appear? 😢

Keader:
It's a good question xD

Hir0shi:
we can help of project sunwell is fixed this bug?!

copy cod!

Hir0shi:
guys

can give me patch for fix this bug!?

:disappointed:

illfated:
@Hir0shi : This is not a forum.
Please post only useful comments.

Hir0shi:
[...]

PR #23936, opened 2019-11-24
Core/GameObjects: Skip LoS checks for traps

### LINKED ISSUE #23142: Core/Spells: Spells cast by traps should not miss
**Description:** When a gob cast a spell it mustn't miss.
the issue could be caused by data1=0 an considered as caster level
GAMEOBJECT_TYPE_TRAP = 6
data1: level (npc equivalent level for casted spell)

1. .gob add temp 179526
2. move around the gob an you will get many misses
3. with a character level 5 it will miss less

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  rev. a926a3088d39

jackpoz:
**Changes proposed:**

-  Disable LoS checks for GameObjects of type GAMEOBJECT_TYPE_TRAP

- [x] 3.3.5
- [ ] master

**Issues addressed:** Ref #21615

**Tests performed:** Tested ingame, couldn't get it to work. It seems #23142 has to be fixed too

- [ ] Need to fix #23142 too to make it work properly

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -1291,6 +1291,14 @@ uint8 GameObject::GetLevelForTarget(WorldObject const* target) const
     if (Unit* owner = GetOwner())
         return owner->GetLevelForTarget(target);
 
+    if (GetGoType() == GAMEOBJECT_TYPE_TRAP)
+    {
+        if (GetGOInfo()->trap.level != 0)
+            return GetGOInfo()->trap.level;
+        if (const Unit* targetUnit = target->ToUnit())
+            return targetUnit->GetLevel();
+    }
+
     return 1;
 }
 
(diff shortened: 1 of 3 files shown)


jackpoz:
@Killyana even with both patches I still don't get healed. GameObject::GetLevelForTarget() is not even called.

@joshwhedon I don't know how many of the changes scattered around different issues are needed to get this to work and I can't keep track of them. Please apply your changes on a branch and submit a PR (it can even be done on GitHub and takes 5 minutes).

jackpoz:
@Killyana when the Aura applied by the GameObject is triggered, https://github.com/TrinityCore/TrinityCore/blob/94f186ee8342824da57eb0fb24378dacc67cbd64/src/server/scripts/Spells/spell_generic.cpp#L2876 check if false so no heal is done. Did you have any custom change there or to spell https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=23493 script binding ?

jackpoz:
Found the last piece of the puzzle (3 different code snippets posted in 3 different places)

PR #23955, opened 2019-12-06
Update extractor.bat

JustZerooo:
More informations, nicer design, a bit updated :)

**Changes proposed:**

-  
-  
-  

- [x] 3.3.5
- [x] master

- [ ] 
- [ ]

### DIFF
--- a/contrib/extractor.bat
+++ b/contrib/extractor.bat
@@ -140,8 +140,9 @@ GOTO MENU
 CLS
 COLOR 0A
 ECHO.
-ECHO Now extracting Cameras, please do not exit the program!
+ECHO Now extracting VMAPS, please do not exit the program!
 ECHO.
+RMDIR /s /q .\Buildings
 RMDIR /s /q .\vmaps
 vmap4extractor.exe
 ECHO.
--- a/contrib/extractor.bat
+++ b/contrib/extractor.bat
@@ -53,10 +53,12 @@ ECHO.
 ECHO Now extracting dbcs and maps, please do not exit the program!
 ECHO.
 RMDIR /s /q .\buildings
-RMDIR /s /q .\vmaps
+RMDIR /s /q .\Cameras
+RMDIR /s /q .\db2
+RMDIR /s /q .\dbc
 RMDIR /s /q .\maps
 RMDIR /s /q .\mmaps
-RMDIR /s /q .\dbc
+RMDIR /s /q .\vmaps
 mapextractor.exe
 ECHO.
 ECHO Dbcs and maps are finished extracting... Now extracting vmaps...
@@ -73,13 +75,12 @@ ECHO Vmaps are now assembled... Now ready to build mmaps...
 ECHO.
 ECHO.
 MKDIR mmaps
-MOVE .\buildings\* .\vmaps
 mmaps_generator.exe
 PAUSE
 
 CLS
 ECHO.
-ECHO All maps/dbc/vmaps/mmaps completed!
+ECHO All /camera/maps/db2/dbc/vmaps/mmaps completed!
 ECHO.
 ECHO You will now need to move the folders dbc, vmaps, maps, and mmaps to your server directory. 
 ECHO.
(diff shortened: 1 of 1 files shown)


Shauren:
20 hours? What kind of potato machine is that

Setraxis:
Some Chinese oder India Computer :D Thats a random value, i saw someoe who tells that :O

Rushor:
Potatoe Machines ftw!

Aokromes:
some people may want to simply extract basic files to test server instead all.

Setraxis:
Fine for these peoples :)

Jinnaix:
No, not fine for these people, your script change doesnt cover that case.

In addition, i think its misleading telling this : 
``` 
ECHO       Before Continuing, PLEASE check to ensure your 
ECHO.
ECHO          Config.wtf is set at accounttype "BFA"
ECHO.
ECHO     (They should be located in: "\World of Warcraft\_retail_\WTF\")
```
The extractors are used for WotLK too, i would say the majority use it for that. Telling something about account type will confuse people.

Setraxis:
Check last change. Already changed ;)

contrib/extractor.bat
```diff
@@ -1,57 +1,76 @@
...
-start /b /w vmap4assembler.exe Buildings vmaps
-rmdir Buildings /s /q
-GOTO MENU
+CLS
+COLOR 0A
+ECHO.
+ECHO.
+ECHO                             ***NOTE*** 
+ECHO This entire process will take anywhere from 2 hours to 8 hours or MORE, 
+ECHO                depending upon the speed of your machine. (If you have a potato machine *Shauren says that*)
```
illfated:
```suggestion
ECHO depending upon the speed of your machine.
ECHO  (if you have a potato machine. *Shauren says that*)
```
You want to keep info text lines at max. 80 char. length, if end user has default size cmd window.

Jinnaix:
Again - and @Aokromes already pointed that too - this whole change deletes the option to extract something specific :

```
SET /P M=Type 1, 2, 3, 4 or 5 then press ENTER:
IF %M%==1 GOTO MAPS
IF %M%==2 GOTO VMAPS
IF %M%==3 GOTO MMAPS
IF %M%==4 GOTO ALL
```

with the change in this pr you can only extract everything and this is a bad change only to make a console window look more "pretty".

illfated:
Thanks, I was too tired to remember that you pointed that out. I will have a further look into that sub-routine call.

contrib/extractor.bat
```diff
@@ -1,57 +1,76 @@
...
-ECHO 1 - Extract dbc/db2 and maps
-ECHO 2 - Extract vmaps (needs maps to be extracted before you run this)
-ECHO 3 - Extract mmaps (needs vmaps to be extracted before you run this, may take hours)
-ECHO 4 - Extract all (may take hours)
-ECHO 5 - EXIT
-ECHO.
-SET /P M=Type 1, 2, 3, 4 or 5 then press ENTER:
-IF %M%==1 GOTO MAPS
-IF %M%==2 GOTO VMAPS
-IF %M%==3 GOTO MMAPS
```
illfated: Add these features back for people who don't want to extract mmaps every time they use this batch script.
My slow 2012 computer uses more than 6 hours to extract those for WotLK, 8-12 hours for Warlords/Legion.

contrib/extractor.bat
Setraxis: i use default size and all fine, no change needed

Setraxis:
NgCduzJ

Now all you need ...

Jinnaix:
Script looks good for me now.

One more thing :

The following was written for a reason because people came here with reporting this as a problem/bug all the time :

```
ECHO The vmaps extractor output text below is intended and not an error:
ECHO ..........................................
ECHO Extracting World\Wmo\Band\Final_Stage.wmo
ECHO No such file.
ECHO Couldn't open RootWmo!!!
ECHO Done!
ECHO ..........................................
ECHO.
```

I wouldn't recommend removing it but it might influence your goal of making the extractor more pretty

Setraxis:
Has nothing todo wit my bat file. open a new ticket for this issue. this has 0 todo with the bat file.

PR #23986, opened 2019-12-21
Core/Movement: Fix wrong orientation set when pausing movement

### LINKED ISSUE #23272: Creature/Movement: Orientation changes wrongly and not updated in client side
**Description:** When a npc use a waypoints via SAI then it pause in some points, orientation will change to the opposite without being updated visually in client side.
Example : Npc: 16708
He has a waypoint: 16708 that start via SAI and stops in 2 spots to execute some cosmetics:
Dekin - ON Waypoint 6 Reached - Run Script
Dekin - ON Waypoint 12 Reached - Run Script

When the npc reach point 6 the orientation is 5.179513 but upon paused it changes to 1.985356

When the npc reach point 12 the orientation is : Orientation: 4.167312 but upon paused it changes to 0.886503

If you move from the creature until it disappear then come back the orientation will get updated for the client

1. .go c id 16708 
2. wait until the npc stops in one oh the 2 spots
3. .gps the npc and compare the orientation

Orientation not updated in client side
image

Orientation updated in client side
image

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  rev. cf37f357df5c

jackpoz:
**Changes proposed:**

-  Fix wrong orientation set when pausing movement. The current Unit position is set as calculated from the spline if no time has passed since the spline was launched, i.e. time_passed == 0

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #23272

- Tested ingame and debugged to see that the orientation of the npc of #23272 https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=16708 is correct
- Talked to the npc of #23272 https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=16708 while it's walking, the position got updated correctly (even if there is another issue that the old orientation is set when talking to the npc, relogging shows the correct orientation, it should be checked if 335 had this issue already).

-  None

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -10171,7 +10171,8 @@ void Unit::StopMoving()
         return;
 
     // Update position now since Stop does not start a new movement that can be updated later
-    UpdateSplinePosition();
+    if (movespline->Started())
+        UpdateSplinePosition();
     Movement::MoveSplineInit init(this);
     init.Stop();
 }
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -10171,7 +10171,7 @@ void Unit::StopMoving()
         return;
 
     // Update position now since Stop does not start a new movement that can be updated later
-    if (movespline->Started())
+    if (movespline->HasStarted())
         UpdateSplinePosition();
     Movement::MoveSplineInit init(this);
     init.Stop();
--- a/src/server/game/Movement/Spline/MoveSpline.h
+++ b/src/server/game/Movement/Spline/MoveSpline.h
@@ -124,7 +124,7 @@ namespace Movement
 
         bool onTransport;
         std::string ToString() const;
-        bool Started() const
+        bool HasStarted() const
         {
             return time_passed > 0;
         }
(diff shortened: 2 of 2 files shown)


jackpoz:
@ccrs @Shauren please take a look at this

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -10171,7 +10171,8 @@ void Unit::StopMoving()
         return;
 
     // Update position now since Stop does not start a new movement that can be updated later
-    UpdateSplinePosition();
+    if (movespline->Started())
```
Ovahlord: isn't !movespline->Finalized()  the common helper used for checking if a spline is currently in progress?

jackpoz: Does that tell if the spline started yet ?

Ovahlord: It does return true as long as a spline is currently in effect, yes. It's the common check used in every movement generator to check if a creature is actually moving or not. (It's also the common check to protect against nullptrs: no spline in effect --> movespline will return nullptr)

jackpoz: But I don't need to know if it's in effect or finished,  I need to know if the Unit walked at least 1 step already in three current spline (hence the name Started and not something like InProgress)

jackpoz:
@Shauren anything against merging this ?

PR #23990, opened 2019-12-22
Fix mob flee speed too fast, Fixes #21899

### LINKED ISSUE #21899: Creature flee speed when "run-away" triggers is too fast!
**Description: Creature flee speed when "run-away" triggers is too fast!**

When fighting mobs that run away when under a certain health level, more often than not they run away at 100%+ full speed. Sometimes they "hobble"/walk away as intended, or do the running animation but look snared.

**Current behaviour: Runs away at 100% speed**

Mob runs away too quickly to rally nearby enemies for support.

**Expected behaviour:**

Mob speed should be closer to that of walking than full on sprinting AFAIK. This does not feel blizzlike.

1. Blackrock Sentry - 4065 (all blackrock), Redridge Mystic - 430 (all redridge gnolls).
2. Fight mob until it triggers creature flee at X hp.
3. Be annoyed that it's already out of your range and running away full speed pulling more mobs that do the same which exponentially grows as more start fleeing full speed in all directions. Die. Repeat.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. 88130d86b0ab 2018-03-29 20:18:52 +0200 (3.3.5 branch) (Win64, R
elease, Static)
Using SSL version: OpenSSL 1.1.0h  27 Mar 2018 (library: OpenSSL 1.1.0g  2 Nov 2
017)
Using Boost version: 1.65.1
Using MySQL version: 5.6.39
Using CMake version: 3.11.0
Compiled on: Windows 10.0.16299
Automatic database updates are disabled for all databases!
MMAPs status: Enabled
maps directory located in data/maps. Total size: 252191207 bytes
vmaps directory located in data/vmaps. Total size: 588247501 bytes
mmaps directory located in data/mmaps. Total size: 2152621724 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS
Using World DB: TDB 335.64

**Operating system: Windows 10**

*** Tried googling this issue and searching the issues on here, found nothing that seemed related. ***

ryancheung:
**Changes proposed:**

-  Creature flee speed should be 66% of its run speed.

- [x] 3.3.5
- [ ] master

#21899

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Movement/MovementGenerators/PointMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/PointMovementGenerator.cpp
@@ -181,6 +181,24 @@ template void PointMovementGenerator<Creature>::DoFinalize(Creature*, bool, bool
 
 //---- AssistanceMovementGenerator
 
+void AssistanceMovementGenerator::Initialize(Unit* owner)
+{
+    RemoveFlag(MOVEMENTGENERATOR_FLAG_INITIALIZATION_PENDING | MOVEMENTGENERATOR_FLAG_DEACTIVATED);
+    AddFlag(MOVEMENTGENERATOR_FLAG_INITIALIZED);
+
+    if (owner->HasUnitState(UNIT_STATE_LOST_CONTROL | UNIT_STATE_NOT_MOVE))
+        return;
+
+    if (!owner->IsStopped())
+        owner->StopMoving();
+
+    owner->AddUnitState(UNIT_STATE_ROAMING | UNIT_STATE_ROAMING_MOVE);
+    Movement::MoveSplineInit init(owner);
+    init.MoveTo(_x, _y, _z);
+    init.SetVelocity(owner->GetSpeed(MOVE_RUN) * 0.66);
+    init.Launch();
+}
+
 void AssistanceMovementGenerator::Finalize(Unit* owner, bool active, bool movementInform)
 {
     AddFlag(MOVEMENTGENERATOR_FLAG_FINALIZED);
--- a/src/server/game/Movement/MovementGenerators/PointMovementGenerator.h
+++ b/src/server/game/Movement/MovementGenerators/PointMovementGenerator.h
@@ -42,7 +42,7 @@ class PointMovementGenerator : public MovementGeneratorMedium<T, PointMovementGe
 
         uint32 GetId() const { return _movementId; }
 
-    private:
+    protected:
         void MovementInform(T*);
 
         uint32 _movementId;
@@ -58,6 +58,7 @@ class AssistanceMovementGenerator : public PointMovementGenerator<Creature>
     public:
         explicit AssistanceMovementGenerator(uint32 id, float x, float y, float z) : PointMovementGenerator<Creature>(id, x, y, z, true) { }
 
+        void Initialize(Unit* owner) override;
         void Finalize(Unit*, bool, bool) override;
         MovementGeneratorType GetMovementGeneratorType() const override;
 };


src/server/game/Movement/MovementGenerators/PointMovementGenerator.cpp
```diff
@@ -181,6 +181,24 @@ template void PointMovementGenerator<Creature>::DoFinalize(Creature*, bool, bool
 
 //---- AssistanceMovementGenerator
 
+void AssistanceMovementGenerator::Initialize(Unit* owner)
+{
+    RemoveFlag(MOVEMENTGENERATOR_FLAG_INITIALIZATION_PENDING | MOVEMENTGENERATOR_FLAG_DEACTIVATED);
+    AddFlag(MOVEMENTGENERATOR_FLAG_INITIALIZED);
+
+    if (owner->HasUnitState(UNIT_STATE_LOST_CONTROL | UNIT_STATE_NOT_MOVE))
+        return;
```
offl: This means if creature is in CC, it will not flee after CC fades?

jackpoz:
Could you fill the remaining fields in the template about "Tests performed" and "Known issues and TODO list" ?
Also
>/home/travis/build/TrinityCore/TrinityCore/src/server/game/Movement/MovementGenerators/PointMovementGenerator.cpp:197:17: fatal error: 
      **use of undeclared identifier '_destination'**
    init.MoveTo(_destination.GetPositionX(), _destination.GetPositionY()...
                ^
1 error generated.
src/server/game/CMakeFiles/game.dir/build.make:2569: recipe for target 'src/server/game/CMakeFiles/game.dir/Movement/MovementGenerators/PointMovementGenerator.cpp.o' failed
make[2]: *** [src/server/game/CMakeFiles/game.dir/Movement/MovementGenerators/PointMovementGenerator.cpp.o] Error 1

jackpoz:
This PR seems to modify more code than the description says.
What about passing the speed in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Movement/MotionMaster.cpp#L938 and keeping the rest of the code the same ? Something like https://github.com/jackpoz/TrinityCore/commit/b931fc9dca13475805d347220045617a3f31992f ?

Killyana:
FYI, the issue is not that the npc always run when it "run-away", in 50% of cases the speed is correct, it's just sometimes something fails to change the speed.

ryancheung:
> This PR seems to modify more code than the description says.
> What about passing the speed in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Movement/MotionMaster.cpp#L938 and keeping the rest of the code the same ? Something like [jackpoz@b931fc9](https://github.com/jackpoz/TrinityCore/commit/b931fc9dca13475805d347220045617a3f31992f) ?

That sounds much intuitive, and I prefer passing the speed in Point Movement constructor, not the Assistance movement constructor.

jackpoz:
well, I would prefer to keep using DoInitialize() instead of overriding Initialize(), while also not duplicating the code of PointMovementGenerator<T>::DoInitialize(T* owner)

ryancheung:
> FYI, the issue is not that the npc always run when it "run-away", in 50% of cases the speed is correct, it's just sometimes something fails to change the speed.

This implementation is from vmangos core. I haven't test it in TC but tested in TrinityTBC core and it works, maybe this would fix that in TC also.

src/server/game/Movement/MovementGenerators/PointMovementGenerator.cpp
```diff
@@ -181,6 +181,24 @@ template void PointMovementGenerator<Creature>::DoFinalize(Creature*, bool, bool
...
+
+    if (!owner->IsStopped())
+        owner->StopMoving();
+
+    owner->AddUnitState(UNIT_STATE_ROAMING | UNIT_STATE_ROAMING_MOVE);
+    Movement::MoveSplineInit init(owner);
+    init.MoveTo(_x, _y, _z);
+    init.SetVelocity(owner->GetSpeed(MOVE_RUN) * 0.66);
```
Ovahlord: This smells like a ass ugly hack to me. Where did you pull this number from?

ryancheung: The number is from the issue comments and it seems reasonable, feels like what's in classic wow.

Killyana: Already done here with the same value: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L8490-L8491

illfated: Does it make much difference with or without `speed *= 0.66f` (float indicator)?

jackpoz:
As stated by someone else, it seems the issue is in Creature::DoFleeToGetAssistance() and Unit::AttackStop() updating both the Creature speed and some error in the logic causes the wrong speed to be set.
It would be better to fix the broken logic in those 2 functions

jackpoz:
I opened a PR at https://github.com/TrinityCore/TrinityCore/pull/23997

ryancheung:
TC members found the way to fix the issue now, I'm closing this pr.

jackpoz:
thanks for the effort put into the PR (the issue actually turned out to affect more bits of code than one would have hoped)

PR #23991, opened 2019-12-22
Core/Movement: Add more LoS checks

jackpoz:
**Changes proposed:**

-  Add more LoS checks to ensure Units don't cross trees/walls when fleeing/moving. I was able to reproduce only the Random movement one, with a Wolf attempting to walk into a tree, but I couldn't reproduce it with Confused one (polymorph). Still, some extra checks are better. MovePositionToFirstCollision() does some checks too but it doesn't seem to guarantee that the destination will be in LoS

- [x] 3.3.5
- [ ] master

**Issues addressed:** None

- Tested ingame with Polymorph with a breakpoint in the new code

- None


illfated:
Thank you for your efforts. Looking forward to see tcubuntu.northeurope.cloudapp.azure.com updated with all the commits after 2019-12-19.

jackpoz:
Server updated

PR #24004, opened 2019-12-26
Core/Spells: Implement a way for scripts to define spell corrections.

Warpten:
**Changes proposed:**

- Stop littering `SpellMgr` with hacks and litter scripts instead, to keep hacks contained.

- [ ] 3.3.5
- [x] master

Will target 3.3.5 if this ships on master or passes validation.

**Issues addressed:** None

**Tests performed:** Builds and runs.

**TODO:**

- [ ] Hotswap ? Should work out of the box, but untested (I only build in static)
- [ ] Possible side effects if code executed during startup expects a modification to a spell to have been applied
- [ ] Migrate corrections out of `SpellMgr` to where they rightfully belong. 

`LoadSpellInfoCorrections` should **not** disappear, since not every spell is a script spell ; some are core mechanics (Hearthstone, Unstuck, Skinning, you name it)

### DIFF
--- a/src/server/game/World/World.cpp
+++ b/src/server/game/World/World.cpp
@@ -2210,6 +2210,10 @@ void World::SetInitialWorldSettings()
     TC_LOG_INFO("server.loading", "Loading scenario poi data");
     sScenarioMgr->LoadScenarioPOI();
 
+    // We do it last to make sure everything a script might call in ShouldLoad() is initialized.
+    TC_LOG_INFO("server.loading", "Loading script-specific spell info corrections...");
+    sScriptMgr->ApplySpellCorrections();
+
     // Preload all cells, if required for the base maps
     if (sWorld->getBoolConfig(CONFIG_BASEMAP_LOAD_GRIDS))
     {
--- a/src/server/game/Scripting/ScriptMgr.cpp
+++ b/src/server/game/Scripting/ScriptMgr.cpp
@@ -1373,6 +1373,11 @@ SpellScriptLoader* ScriptMgr::GetSpellScriptLoader(uint32 scriptId)
     return ScriptRegistry<SpellScriptLoader>::Instance()->GetScriptById(scriptId);
 }
 
+void ScriptMgr::ApplySpellCorrections() const
+{
+    FOREACH_SCRIPT(SpellCorrectionLoader)->ApplyCorrections();
+}
+
 void ScriptMgr::OnNetworkStart()
 {
     FOREACH_SCRIPT(ServerScript)->OnNetworkStart();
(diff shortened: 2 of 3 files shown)


Keader:
Hmm, is not better have all "spell hacks" in same place?

src/server/game/World/World.cpp
```diff
@@ -2210,6 +2210,10 @@ void World::SetInitialWorldSettings()
     TC_LOG_INFO("server.loading", "Loading scenario poi data");
     sScenarioMgr->LoadScenarioPOI();
 
+    // We do it last to make sure everything a script might call in ShouldLoad() is initialized.
```
Shauren: This doesn't work, you need to process spell data before anything else starts reading SpellInfo, for example checking duration for loading creature_addon auras

Shauren (changes requested):
And I'm not sure about putting all changes all over the place in scripts

Warpten:
Idk i feel like have a 7 billion lines switch isnt great.

src/server/game/World/World.cpp
Warpten: That makes sense, I honestly didn't think about it too much and just assumed nothing needed to access SpellInfo on load

Mesielecat:
I think is better keep all hack into the file of the script, when you have lot of correction you need go to spellMgr, and review the file, but if you don't use spellMgr and you change the entire switch of the spellMgr to the script this gonna keep it clean the source and ordened. Is a clean way of make a hack haha

Jildor:
I prefer keep "hackfix dbc corrections" in spellMgr.cpp, I think it's better to have the shit in one place than all the shit scattered in all corners 😄 

an alternative could be add those corrections in DB, in the end only modify a value of a field.
I know, we have spell_dbc table, but this table is for: "This table contains data regarding serverside spells which are not to be found in Client DBC files."
and if I remember right, we should add all fields to modify one spell in this table.
So, maybe we can create by example spell_dbc_corrections and change by example:
```c++
    // Crafty's Ultra-Advanced Proto-Typical Shortening Blaster
    ApplySpellFix({ 51912 }, [](SpellInfo* spellInfo)
    {
        spellInfo->Effects[EFFECT_0].Amplitude = 3000;
    });

    // Desecration Arm - 36 instead of 37 - typo? :/
    ApplySpellFix({ 29809 }, [](SpellInfo* spellInfo)
    {
        spellInfo->Effects[EFFECT_0].RadiusEntry = sSpellRadiusStore.LookupEntry(EFFECT_RADIUS_7_YARDS);
    });
```

```sql
SET @CORRECTION_AMPLITUDE := 32;
SET @CORRECTION_RADIUS    := 30;
DELETE FROM `spell_dbc_corrections` WHERE `SpellId` IN (51912,29809);
INSERT INTO `spell_dbc_corrections` (`SpellId`,`CorrectionId`,`CorrectionValue`.`Effect`) VALUES
-- Crafty's Ultra-Advanced Proto-Typical Shortening Blaster
(51912,@CORRECTION_AMPLITUDE,3000,0),
-- Desecration Arm - 36 instead of 37 - typo? :/
(29809,@CORRECTION_RADIUS,37,0);
```

So, we can reload spell_dbc_corrections with command and apply changes without compile and restart server.

Aokromes:
i love jildor's suggestion.

Shauren:
And I don't, in the end when looking at database you will see (51912, 32, 3000, 0) - completely meaningless

Mesielecat:
Well that point is the same with smart_script, you gonna see lot of value in a table, but when you use sai editor you can work really faster, someone could add a extention to the spellwork and use as editor for this table

Aokromes:
comment field :P

VincentVanclef:
jildors suggestion + a comment field is the way to go imo

Warpten:
This is inherently not reloadable because the effects of previous corrections cannot be reverted (unless you want to reload ALL the spells again from DBC)

VincentVanclef:
And what would be so bad about reloading 57.000 rows from dbcs? itll take a second. :3

Shauren:
Pointers to data rows cached everywhere? Suddenly being invalidated does not seem like a good idea

Issue #21190, opened 2018-01-05
[3.3.5] Core/Waypoint: Waypoint pause via gossip hello mustn't be executed during escorts


Funkeln:
**Description:**

Some NPCs wait very long before starting their waypoints during escorts (i.e Arthas @ CoS and Thrall @ Old hillsbrad.

**Current behaviour:**

For example, after talking to Arthas to start the waves (after he moves into the city) it takes full 3 minutes until he starts doing anythiing. 

**Expected behaviour:**

It should not be like that, there should be a small pause when waypoints start, but atm its absurdly high. It makes achievements like the timed on in CoS impossible because waypoint takes too long.

**Steps to reproduce the problem:**

1. Go to CoS
2. Play through the lore yadda-yadda stuff until Arthas move into the city
3. Select his gossip to start waves
4. Notice as you wait forever for him to do anything

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/e0b609a178528995959e5ea54255a3f856e62913

ccrs:
most likely previously the timers set by the scripts were just ignored/not honored

tje3d:
Waves start after about 1 or 2 minute, that wait forever is another bug.

Aokromes:
https://github.com/TrinityCore/TrinityCore/pull/17860

Funkeln:
That stratholme rewrite will not fix this. I checked all of his timers, both in his script and script_waypoint. None of them are ever over 20 seconds. This 2 minute wait is coming from elsewhere, something in the core processing the waypoint movement wait timers went wrong.

tje3d:
@Funkeln its not done yet

Nyeriah:
@ccrs The issue is that when a creature has a waypoint sequence in progress, you pause them whenever they are talked to with CONFIG_CREATURE_STOP_FOR_PLAYER (Creature.MovingStopTimeForPlayer - defaulted at 3 minutes) whenever they are talked to. So, Arthas, Thrall and any other escort NPCs that have gossips mid script will have that 3 minutes pause sent OnGossipHello (whenever they are talked to).

There is another issue, however. Even if we lower that setting, and I did try lowering it to 10 seconds, it is still broken. Because then the waypoint is forced to unpause once that setting timer expires, completely ignoring the script. So simply talking to Arthas, without selecting the gossip, still forces him to start moving.

[...]

PR #24005, opened 2019-12-27
Core/Movement: Fix WaypointMovementGenerator ignoring EscortAI::SetEscortPaused() on Gossip Hello

jackpoz:
**Changes proposed:**

- Fix WaypointMovementGenerator ignoring EscortAI::SetEscortPaused() when talking to a NPC, scheduling a 3 minutes pause on the WaypointMovementGenerator even if it was already disabled with no timer, waiting for the player to select a gossip entry to resume the movement.

- [x] 3.3.5
- [ ] master

**Issues addressed:** Ref #21190

- tried the Thrall @ Old hillsbrad escort event. When Thrall stops after killing some kind of boss guard at the Keep gate (and before mounting up), he waits for the player to select a gossip. As soon as I selected the gossip, he started walking (and the WaypointMovementGenerator wasn't paused twice, I debugged the code)

None

### DIFF
--- a/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/WaypointMovementGenerator.cpp
@@ -56,6 +56,10 @@ void WaypointMovementGenerator<Creature>::Pause(uint32 timer/* = 0*/)
 {
     if (timer)
     {
+        // Don't try to paused an already paused generator
+        if (HasFlag(MOVEMENTGENERATOR_FLAG_PAUSED))
+            return;
+
         AddFlag(MOVEMENTGENERATOR_FLAG_TIMED_PAUSED);
         _nextMoveTime.Reset(timer);
         RemoveFlag(MOVEMENTGENERATOR_FLAG_PAUSED);


jackpoz:
@ccrs you might want to take a look at this

Ovahlord:
That's because the Gossip hello opcode handler does not pause the movement generator but instead it stops the entire movement by calling StopMoving. probably a oversight since the rework of waypoint movement.

ccrs:
this is a huge ummmm

the real fix for `EscortAI` would be to include a `Resume` inside here
```c++
void EscortAI::SetEscortPaused(bool on)
{
    if (!HasEscortState(STATE_ESCORT_ESCORTING))
        return;

    if (on)
    {
        AddEscortState(STATE_ESCORT_PAUSED);
        if (MovementGenerator* movementGenerator = me->GetMotionMaster()->GetCurrentMovementGenerator(MOTION_SLOT_DEFAULT))
            movementGenerator->Pause(0);
    }
    else
    {
        RemoveEscortState(STATE_ESCORT_PAUSED);
        _resume = true;
    }
}
```

now, why on earth didn't I do that?

ccrs:
also, your proposal should also be taken into consideration, should we prevent timed pause if a pause is already active?

give me a week to take a look into this, if I'm MIA feel free to modify the PR, or merge it and I'll review it whenever I can

jackpoz:
EscortAI already calls Resume(0) at https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/ScriptedAI/ScriptedEscortAI.cpp#L203-L207

The thing that Escort AI cannot control is the fact that something else might remove the "all movement must be paused until further notice" and replace it (from GossipHello) with "resume movement in 3 minutes, even if the Player hasn't selected anything from the Gossip). So keep in mind the issues are 2:
1. Movement must resume as soon as Resume is called (or EscortAI::SetEscortPaused(false) in our case)
2. Movement must stay paused when a script calls Pause(0) until the script itself is ready to Resume it

I would say the "pause the movement until further notice" takes precedence over "pause the movement for a limited amount of time".

ccrs:
feels like the AI should be able to decide

Faq:
Machine learning yay

ccrs:
I will most likely move the "pause action" from the gossip handler and similar places to `UnitAI`

jackpoz:
@ccrs should I merge the PR the way it is for now and let you work on it when you have more time ?

ccrs:
hold

jackpoz:
enough time has passed

Issue #23326, opened 2019-05-29
[3.3.5/Core]Warrior cannot charge in water


soniczhang11:
**Description:**

Warrior cannot charget in water.

**Current behaviour:**
When warrior try to cast charge in water, notice said that it cannot find a path.
**Expected behaviour:**

CHANGEME Tell us what should happen instead.

**Steps to reproduce the problem:**

1. Create a Warrior and levelup.
2. Swim in water and try to find some mobs.
3. Try to cast charge to those mobs.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

16bf11ab20fcd5e3984dd90d011d13f56288d1ea

**Operating system:** CentOS

Killyana:
Probably related https://github.com/TrinityCore/TrinityCore/issues/22253

jackpoz:
please provide a clear example with .go xyz coordinates to reproduce the issue

jackpoz:
https://github.com/TrinityCore/TrinityCore/pull/24010

Issue #11526, opened 2014-01-28
[Core/Spell] Warrior Charge


Nephelion:
https://github.com/TrinityCore/TrinityCore/commit/1b48c7b977b19f990b7f1389f48a0a45cdde9784

when charge with warr some time go underground and buged ...
##

lainimer:
confirm baged

SQLfix:
Any news ?

Aokromes:
No.

jackpoz:
Please try after https://github.com/TrinityCore/TrinityCore/commit/91c399cad239b20dac68629587e7c8b4ec612be3

jackpoz:
any feedback ?

ccrs:
I tested it and got same results as https://github.com/TrinityCore/TrinityCore/commit/91c399cad239b20dac68629587e7c8b4ec612be3 comments

jackpoz:
No more charge into the void tho ? Can we close this issue?

ccrs:
it's so restrictive that I guess yes xd

Runico:
No, it still charge into void

jackpoz:
@Runico post gps coordinates of start and end position, even better if you post them in the .go xyz format

Frytiks:
Add IsArena check With pathfind_nopath should be tempfix

Runico:
@jackpoz this is coordinates of start and end position
```
Start Point:
Map: 530 (Outland) Zone: 3522 (Blade's Edge Mountains) Area: 3784 (Forge Camp: Terror) Phase: 4294967295
X: 1453.244507 Y: 7314.201660 Z: 364.589569 Orientation: 3.726342
grid[34,45]cell[5,5] InstanceID: 0
 ZoneX: 28.232836 ZoneY: 81.707520
GroundZ: 364.590057 FloorZ: 364.590057 Have height data (Map: 1 VMap: 1 MMap: 1)
```
End Point:
```
Map: 530 (Outland) Zone: 3522 (Blade's Edge Mountains) Area: 3784 (Forge Camp: Terror) Phase: 4294967295
X: 1437.925903 Y: 7303.247070 Z: 372.203339 Orientation: 3.702782
grid[34,45]cell[5,5] InstanceID: 0
 ZoneX: 28.434763 ZoneY: 82.131081
GroundZ: 372.205139 FloorZ: 372.205139 Have height data (Map: 1 VMap: 1 MMap: 1)
```

[...]

PR #24010, opened 2019-12-28
Core/MMaps: Fix charge underwater/falling

### LINKED ISSUE #22253: Warrior Charge Underwater Facing the Wrong Way
**Description:**

Warrior charge underwater causes the warrior to charge at the mob but faces the wrong way.

**Current behaviour:**

When your underwater fighting turtles for instance and you use charge the charge works but when you stop at the mob you are facing the wrong way.

**Expected behaviour:**

You should be able to use the charge underwater and stop right in front of the mob so you can instantly attack without having to move around to get the mob in front of you.

1. Go anywhere where there is deep water and mobs
2. Cast charge on any mob and your toon will run at them.
3. Then you are greeted by the error that you are facing the wrong way.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

6726126664db+

**Operating system:** 

Windows Server 2012

jackpoz:
**Changes proposed:**

- Fix charge sometimes returning "no path available" when swimming/underwater.
- Fix charge sometimes returning "no path available" when falling, now it allows to charge downwards.
- Fix charge selecting a destination point 4 yards above swimming/underwater targets.
- Disabled raycast charge as that seems to be unblizzlike. This might fix some falling underground issues.
- Removed the Z 4 yards offset that never made any sense to me.

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #22253 , #23326 , #24011 , maybe #11526

Performed the tests described at https://github.com/TrinityCore/TrinityCore/issues/22253#issuecomment-465990651 

- [ ] More tries should be done in water (shallow and deep)
- [ ] Any try should be done with flying caster/targets

### DIFF
--- a/src/server/game/Movement/PathGenerator.cpp
+++ b/src/server/game/Movement/PathGenerator.cpp
@@ -205,23 +205,19 @@ void PathGenerator::BuildPolyPath(G3D::Vector3 const& startPos, G3D::Vector3 con
         TC_LOG_DEBUG("maps.mmaps", "++ BuildPolyPath :: farFromPoly distToStartPoly=%.3f distToEndPoly=%.3f", distToStartPoly, distToEndPoly);
 
         bool buildShotrcut = false;
-        if (_sourceUnit->GetTypeId() == TYPEID_UNIT)
-        {
-            Creature* owner = (Creature*)_sourceUnit;
 
-            G3D::Vector3 const& p = (distToStartPoly > 7.0f) ? startPos : endPos;
-            if (_sourceUnit->GetBaseMap()->IsUnderWater(p.x, p.y, p.z))
-            {
-                TC_LOG_DEBUG("maps.mmaps", "++ BuildPolyPath :: underWater case");
-                if (owner->CanSwim())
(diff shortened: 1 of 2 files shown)


Killyana:
This fix both issues, I noticed if you are falling you cannot charge, it returns no path available, but it's not relate to this PR

jackpoz:
Fixed charge not working when falling, now it allows to charge downwards

jackpoz:
Disabled raycast charge as that seems to be unblizzlike

Killyana:
Charge in water, and when the player is falling are fixed.

jackpoz:
any chance it fixed the falling underground also ?

Killyana:
Yes I cannot reproduce an undermap issue, also it's better to remove: https://github.com/TrinityCore/TrinityCore/pull/24010/files#diff-952c711803610aa78110e4f92ff2516aR5515-R5516 this 4 yards are now useless, thy just cause a weird jump at the end of the charge

jackpoz:
Removed the Z 4 yards offset that never made any sense to me

PR #24025, opened 2019-12-31
Fix fleeing speed to 66% run speed

jackpoz:
**Changes proposed:**

-  Fix fleeing speed to 66% run speed until the creature starts returning home

- [x] 3.3.5
- [ ] master

The tests described at https://github.com/TrinityCore/TrinityCore/issues/21899#issuecomment-568303556

- Might aswell fix some more fleeing issues in this PR while at it

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -950,7 +950,6 @@ void Creature::DoFleeToGetAssistance()
         Cell::VisitGridObjects(this, searcher, radius);
 
         SetNoSearchAssistance(true);
-        UpdateSpeed(MOVE_RUN);
 
         if (!creature)
             /// @todo use 31365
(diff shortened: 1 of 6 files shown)


Killyana:
It looks pretty good, slow effects from spells are handled correctly, the only issue I get is if you make a npc to call for help a second time it will get his speed reduced too much, this extra reduction is removed when it get back to combat: https://imgur.com/a/eMgFuDN

jackpoz:
Some notes for the people testing it:
>.go c 8786 will flee
.go c 28334 will call for help

jackpoz:
@Killyana flag 64 calls https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/SmartScripts/SmartScript.cpp#L571 which then calls https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/SmartScripts/SmartAI.cpp#L810 , changing the mob from run to walk depending on what _run variable contains (shared variable across all SmartScripts for current mob.

I don't change that in this PR and chances are the issue happens before too.

jackpoz:
there is currently an issue in existing 335 that forces SetWalk(true) after the mob survives a fleeing. It will not be fixed in this PR.

PR #24036, opened 2020-01-03
Core/PathGenerator: Fix path generator returning shortcuts when start and end are on the same polygon

jackpoz:
**Changes proposed:**

-  Fix path generator returning shortcuts when start and end are on the same polygon by handling this case as if start and end were on 2 different polygons. This will ensure BuildPointPath() gets called which calls FindSmoothPath(), making sure each step is not longer than SMOOTH_PATH_STEP_SIZE (4 yards)

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes the issue reported at https://github.com/TrinityCore/TrinityCore/commit/88a14251e2765880e90cb41be7919cd39e3308e2#commitcomment-36645636

Checked the locations reported at https://github.com/TrinityCore/TrinityCore/commit/88a14251e2765880e90cb41be7919cd39e3308e2#commitcomment-36645636

None

### DIFF
--- a/src/server/game/Movement/PathGenerator.cpp
+++ b/src/server/game/Movement/PathGenerator.cpp
@@ -245,19 +245,14 @@ void PathGenerator::BuildPolyPath(G3D::Vector3 const& startPos, G3D::Vector3 con
     // *** poly path generating logic ***
 
     // start and end are on same polygon
-    // just need to move in straight line
+    // handle this case like if they were 2 different polygons, building a line path split in few points
     if (startPoly == endPoly)
     {
         TC_LOG_DEBUG("maps.mmaps", "++ BuildPolyPath :: (startPoly == endPoly)");
 
-        BuildShortcut();
-
         _pathPolyRefs[0] = startPoly;
-        _polyLength = 1;
-
-        _type = farFromPoly ? PATHFIND_INCOMPLETE : PATHFIND_NORMAL;
-        TC_LOG_DEBUG("maps.mmaps", "++ BuildPolyPath :: path type %d", _type);
-        return;
+        _pathPolyRefs[1] = endPoly;
+        _polyLength = 2;
     }
 
     // look for startPoly/endPoly in current path
--- a/src/server/game/Movement/PathGenerator.cpp
+++ b/src/server/game/Movement/PathGenerator.cpp
@@ -245,7 +245,7 @@ void PathGenerator::BuildPolyPath(G3D::Vector3 const& startPos, G3D::Vector3 con
     // *** poly path generating logic ***
 
     // start and end are on same polygon
-    // handle this case like if they were 2 different polygons, building a line path split in few points
+    // handle this case as if they were 2 different polygons, building a line path split in some few points
     if (startPoly == endPoly)
     {
         TC_LOG_DEBUG("maps.mmaps", "++ BuildPolyPath :: (startPoly == endPoly)");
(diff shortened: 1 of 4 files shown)


jackpoz:
@VincentVanclef this PR fixes your charge issue, you might want to test it and report some feedback :)

ccrs (approved):

Killyana:
Charges are much better, but I found a case where you will undermap
start position:  .go xyz  1128.96643, -4991.62988, 18.356369 1
target: .go xyz 1134.28735, -4997.28515, 26.179434 1

VincentVanclef:
> @VincentVanclef this PR fixes your charge issue, you might want to test it and report some feedback :)

Tested it, and it does indeed fix the issue. However, as @Killyana says, i can still go underground at:

start position: .go xyz 1128.96643, -4991.62988, 18.356369 1
target: .go xyz 1134.28735, -4997.28515, 26.179434 1

but really nice and swift work @jackpoz ! 👍

src/server/game/Movement/PathGenerator.cpp
```diff
@@ -245,19 +245,14 @@ void PathGenerator::BuildPolyPath(G3D::Vector3 const& startPos, G3D::Vector3 con
     // *** poly path generating logic ***
 
     // start and end are on same polygon
-    // just need to move in straight line
+    // handle this case like if they were 2 different polygons, building a line path split in few points
```
illfated: (I know this is just a comment line, but it would be nice to be able to read it in correct English)
```suggestion
    // handle this case as if they were 2 different polygons, building a line path split in some few points
```

jackpoz:
it might be that a Player shouldn't be able to charge with that slope. The slope angle was changed in https://github.com/TrinityCore/TrinityCore/commit/93a68a66c3009ea86569076ea05a39ce87dfb78d from 55° to 70° but maybe it should be like that only in instances (we can't have 2 different angles for charge and for creature movement)

jackpoz:
I reverted https://github.com/TrinityCore/TrinityCore/commit/93a68a66c3009ea86569076ea05a39ce87dfb78d for continents, please re-extract mmaps for maps 0, 1, 530 and 571 (or just the map you want to try).

Powershell script:
```powershell
.\mmaps_generator 0
.\mmaps_generator 1
.\mmaps_generator 530
.\mmaps_generator 571
```

I didn't change mmaps version so the old mmaps work fine too

VincentVanclef:
> I reverted [93a68a6](https://github.com/TrinityCore/TrinityCore/commit/93a68a66c3009ea86569076ea05a39ce87dfb78d) for continents, please re-extract mmaps for maps 0, 1, 530 and 571 (or just the map you want to try).
> 
> Powershell script:
> 
> ```powershell
> .\mmaps_generator 0
> .\mmaps_generator 1
> .\mmaps_generator 530
> .\mmaps_generator 571
> ```
> 
> I didn't change mmaps version so the old mmaps work fine too

But, doesnt this mean that we can still charge underground in instances?

Killyana:
The angle must be 70° at least, with this last change I wasn't able to charge in this spot but it works in retail https://imgur.com/a/lz72tCG

src/tools/mmaps_generator/MapBuilder.cpp
```diff
@@ -1016,6 +1022,11 @@ namespace MMAP
         config.detailSampleDist = config.cs * 16;
         config.detailSampleMaxError = config.ch * 1;
 
+        if (isContinentMap(mapID))
+        {
+            config.walkableClimb = m_bigBaseUnit ? 2 : 4;
```
jackpoz: @Killyana how about I revert any mmaps_generator change and just push the charge fix that seems to be an improvement over current 3.3.5 branch ? we can continue research about angle and slope later on

Killyana: Yes sounds good.

PR #24046, opened 2020-01-10
Core/DynamicObjects: Restore the ability of DynamicObjects to be world objects

### LINKED ISSUE #17081: [3.3.5] Core/Spell: Far Sight Updates
**Description**: Locations viewed via Far Sight to not show updates.

Discovered while discussing #17041, marking as a separate issue to note as a to-do. The problem is similar to #17041, but the PR does not have an effect on Far Sight.

**Current behaviour**: 

Distant locations (out of normal caster sight) that are scryed via Far Sight display a "freeze-frame" of current activity.

Examples:
- Player movement in the target area will not display.
- NPCs will finish their current waypoint, then appear to the caster to stand in place.

Information will be updated if the spell is re-cast, but I think that this is more on account of sight going back to the caster for so long during the cast time.

**Expected behaviour**:

The caster should be able to see activities play out in the remote location as though they were standing there.

1. Create a shaman, and teach them Far Sight (Spell 6196).
2.  Use Far Sight to observe a remote location that the caster usually would not be able to see units in due to range. To observe this problem, pick a target location with a co-operating player and/or an NPC that constantly patrols.
3. Wait for an NPC to patrol or get the player in the target area to move. The caster's view will not reflect player movement unless they re-cast Far Sight.

**Branch(es)**: 335

**TC hash/commit**:  b91bed0

**TDB version**:  335.61

**Operating system**:  Linux

**Linking model**: dynamic

xvwyh:
… which was accidentally removed in f0f4a620fbeef9cc450d1294c1964fe98e45645d breaking Far Sight

**Changes proposed:**

-  Restore the code that allowed DynamicObjects to be added to world objects container, which is a necessity for them to be included in packet broadcast recipients. Before the aforementioned commit this was working as expected because `Map::AddToGrid` for DynamicObjects was using the generic template of the method which already implements this behavior.

**Target branch(es):**

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes #17081

**Tests performed:** [Tested to be working](https://github.com/TrinityCore/TrinityCore/issues/17081#issuecomment-573047250) by @Keader at whose request this PR is made.

Albeit not tested, the code that creates this issue exists in master branch as well, so it makes sense to apply the fix there as well.

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -336,7 +336,10 @@ template<>
 void Map::AddToGrid(DynamicObject* obj, Cell const& cell)
 {
     NGridType* grid = getNGrid(cell.GridX(), cell.GridY());
-    grid->GetGridType(cell.CellX(), cell.CellY()).AddGridObject(obj);
+    if (obj->IsWorldObject())
+        grid->GetGridType(cell.CellX(), cell.CellY()).AddWorldObject(obj);
+    else
+        grid->GetGridType(cell.CellX(), cell.CellY()).AddGridObject(obj);
 
     obj->SetCurrentCell(cell);
 }


Keader:
@Shauren

Warpten:
Corollary: why is `SetCurrentCell` called for players, gameobjects, and dynamic objects **only** ? Were it not for this difference, you could just collapse these specializations. Why do some types need to know their current cell?

xvwyh:
It's because only some entity types inherit from `MapObject`. It's required to support delayed relocation between cells, without which any kind of iteration on cell's creatures etc will break if that creature decides to move to a different cell (and thus modify the container that's currently being iterated on). You need to somehow [store which cell the object was in previously](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Maps/MapObject.h#L47), to be able to [figure out if it had moved to a different cell at all](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Maps/Map.cpp#L1080).

Until transports were implemented, only players and creatures supported delayed relocation, because they were the only types capable of movement. With transports, gameobjects and dynobjects can now be attached to transports, hence the need appeared to allow them to be moved as well, and so they now also inherit from `MapObject` and delayed relocation methods were added for them as well.

But you're right, seems like by now the only type that DOESN'T support movement is Corpse, and I would argue that it should. I think corpses need to be attachable to transports as well, because:

- opcode `MSG_CORPSE_QUERY` contains space for guid of the transport that the corpse resides on;
- opcode `CMSG_CORPSE_MAP_POSITION_QUERY` exists, which the client sends periodically if the corpse is located on a transport that's on a different map to query the server about the whereabouts of said transport, since it's not visible to the client.

If this is implemented, then yeah, all objects that can be placed in maps will support movement, hence removing the need for `MapObject` (the functionality of which could then be pushed to `WorldObject`) and for separate specializations of `Map::AddToGrid`.

Warpten:
Right, it's that junk. You can't specialize for `MapObject` either to reduce code duplication here, so, bleh.

LGTM.

jackpoz:
Thanks for the PR :)

Issue #22448, opened 2018-09-17
Implement MoveSplineFlag::Enter_Cycle


xvwyh:
#### The problem

Currently enter cycle spline flag is not being handled, or, as the comment in the code claims, "MoveSplineFlag::Enter_Cycle support dropped". But it's an important flag without which cyclic splines are broken on the server and the units end up moving in a different path than they are on the client.

What this flag does is that it specifies that the path contains an additional point at the beginning (after the fake catmull-rom point, of course) which is the unit's starting position, before it entered a looping path. It essentially informs that client that the first point in the spline is not part of the loop and must be discarded after one cycle:
![entercycle1](https://user-images.githubusercontent.com/43143822/45637501-5010f100-bab3-11e8-9e0a-1f5c84ab0034.png)
Client does that, and thus the unit visually moves as intended, both on the first cycle and on every subsequent one. After the first cycle is done, it discards the first point and the spline now only contains the points of a looping path:
![entercycle2](https://user-images.githubusercontent.com/43143822/45637500-5010f100-bab3-11e8-83d1-0feb7ab5aecc.png)
But the server doesn't. Server doesn't handle the flag and removes neither the flag, nor the first "entry" point from the path. This results in serverside unit's path being different from that on the client, and this also incorrectly informs all other clients, that encountered the unit on its subsequent cycles, that the path is still on its first loop. Both of these issues cause position desyncs, the severity of which are proportional to how far the unit was initially away from the first path point:
![entercycle3](https://user-images.githubusercontent.com/43143822/45637499-5010f100-bab3-11e8-9ac7-4b6606dd1da2.png)
Too see the issue you would have to make a way to visualize unit's real serverside position, for example by spawning and despawning some creature every so often.

#### The solution

Don't blindly apply this patch, it has a few caveats which you can read below, and may require some additional work for improved experience. It's also a patch for my core, not for TC, but it should be trivial to port it.

```diff
diff -r b7421dfd4ba0 -r 7df6d3cd6702 src/server/game/Entities/Unit/Unit.cpp
--- a/src/server/game/Entities/Unit/Unit.cpp    Mon Dec 12 23:13:32 2016 +0300
+++ b/src/server/game/Entities/Unit/Unit.cpp    Fri Dec 16 00:43:26 2016 +0200
@@ -407,6 +407,20 @@
     movespline->updateState(t_diff);
     bool arrived = movespline->Finalized();
 
+    if (movespline->isCyclic())
+    {
[...]

PR #24049, opened 2020-01-12
Core/Movement: Implement MoveSplineFlag::Enter_Cycle flag

Faq:
by xvwyh

**Changes proposed:**

-  
-  
-  

- [x] 3.3.5
- [ ] master

#22448
#20238

Some testing was performed by Ovahlord described in #22448

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -42,6 +42,7 @@
 #include "LootMgr.h"
 #include "MotionMaster.h"
 #include "MovementGenerator.h"
+#include "MovementPacketBuilder.h"
 #include "MoveSpline.h"
 #include "MoveSplineInit.h"
 #include "ObjectAccessor.h"
@@ -484,19 +485,28 @@ void Unit::UpdateSplineMovement(uint32 t_diff)
     movespline->updateState(t_diff);
     bool arrived = movespline->Finalized();
 
+    if (movespline->isCyclic())
+    {
+        m_splineSyncTimer.Update(t_diff);
+        if (m_splineSyncTimer.Passed())
+        {
+            m_splineSyncTimer.Reset(5000); // Retail value, do not change
+
+            WorldPacket data(SMSG_FLIGHT_SPLINE_SYNC, 4 + GetPackGUID().size());
+            Movement::PacketBuilder::WriteSplineSync(*movespline, data);
+            data.appendPackGUID(GetGUID());
+            SendMessageToSet(&data, true);
+        }
+    }
+
     if (arrived)
         DisableSpline();
 
-    m_movesplineTimer.Update(t_diff);
-    if (m_movesplineTimer.Passed() || arrived)
-        UpdateSplinePosition();
+    UpdateSplinePosition();
 }
 
 void Unit::UpdateSplinePosition()
 {
-    static uint32 const positionUpdateDelay = 400;
-
-    m_movesplineTimer.Reset(positionUpdateDelay);
     Movement::Location loc = movespline->ComputePosition();
 
     if (movespline->onTransport)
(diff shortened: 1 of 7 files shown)


Ovahlord:
Tested this stuff for a long time now, no issues to complain about.

jackpoz:
@Ovahlord could you add the changes that Shauren mentioned in irc ?

Faq:
Would be good to add here changes he mentioned on irc.

jackpoz:
>[2020-01-14 20:09:45] <Shauren> eeeeeeeeh, im pretty sure you can do that calculation in a better way than initializing a duplicate spline
[2020-01-14 20:11:22] <Elon> u mean u can recalc the new spline duration and length after completing the first cycle without tearing half of the spline system apart?
[2020-01-14 20:13:09] <Shauren> totally
[2020-01-14 20:14:24] <Shauren> since you only need to drop first point, you basically just need to shorten the duration by the time it takes to move from first to second point
[2020-01-14 20:15:03] <Elon> does the spline system actually have a segment info?
[2020-01-14 20:15:11] <Elon> like duration per point?
[2020-01-14 20:15:46] <Shauren> which means you need to simply use spline.length(1, spline.last()) // where spline is the internal member of MoveSpline
[2020-01-14 20:16:03] <Shauren> that returns the duration with the first segment removed

jackpoz:
@Faq any followup to the changes mentioned above ?

Faq:
Sorry, no time. Anyone can do that.

jackpoz:
One day someone will implement what Shauren suggested

jackpoz:
Thanks for the PR :)

xvwyh:
Not sure that the thing @Shauren is suggesting is achievable as easily as he thinks. If we were talking about linear splines - absolutely. But with curved splines, removing a point can radically alter the curvature of surrounding segments, and thus their length. And even the angle of approach to the first point in the loop will be much different with and without the enter_cycle point, which will also affect the curvature and length. Unless Catmull-Rom splines have some magical properties that I don't know about.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -484,19 +485,28 @@ void Unit::UpdateSplineMovement(uint32 t_diff)
...
+            data.appendPackGUID(GetGUID());
+            SendMessageToSet(&data, true);
+        }
+    }
+
     if (arrived)
         DisableSpline();
 
-    m_movesplineTimer.Update(t_diff);
-    if (m_movesplineTimer.Passed() || arrived)
```
jackpoz: could this timer be restored to avoid calling UpdateSplinePosition() every update ?

Ovahlord: It's less about updating the position, it's about the following triggered actions such as vmap based position data updates. Long story short, this change is not only updating spline positions but also causing vmap lookups which are now being spamed with every single update tick. Not really a healthy thing to do

xvwyh: Then maybe that's the thing you want to optimize instead? Throttle (or completely block) movement flags recalc for creatures who are using random or waypoint movement out of combat, throttle visibility updates and MoveInLineOfSight calls. IMO that's a better solution than causing server/client desyncs.

And as for dynamic spawns, I don't know how they work, but I cannot see why can zone be changeable in real time: blizzard used separate servers for separate zones from the beginning, as they themselves said, and I highly doubt that it makes sense for spawns to be able to transition from zone to zone under these circumstances, as that would mean they would have to be moved to a different server responsible for that zone, which is just a technical overkill.

Jildor: this can be related with this issue?
https://github.com/TrinityCore/TrinityCore/issues/24356

xvwyh: If the claim "seems related with pathfinding" is true - then no. Pathfinding would be executed in the same amounts with or without this position update delay. This delay is only responsible for controlling how often is the unit being physically moved along the already generated path.

PR #24057, opened 2020-01-13
Scripts/Spells: Nourish

Sorikoff:
**Changes proposed:**
Migrate spell `Nourish` to spell scripts system.

**Target branch(es):** 3.3.5

~~**Issues addressed:**~~

**Tests performed:** Seems to be working correct, however, additional testing of output numbers would be appreciated.

~~**Known issues and TODO list:**~~


jackpoz:
@Sorikoff could you do the "additional testing of output numbers" too ?

Sorikoff:
Just checked. Difference with/without this PR for Nourish is ~1%.

Keader:
It should be changed in Launch hook, no?

Sorikoff:
Well, from what I see most of `HandleHeal` spells are using `OnEffectHitTarget`.

jackpoz:
did you 2 agree if the current PR is ok or if it has to be changed ?

Sorikoff:
I think it works the same as before. I can re-check if required but later.

Sorikoff:
So tested again (no items, 80 lvl, all talents and spells + Glyph of Nourish learned). Difference with/without PR (5 casts of Nourish when Regrowth active) is about 2% (crits still matter).

src/server/scripts/Spells/spell_druid.cpp
```diff
@@ -967,6 +968,39 @@ class spell_dru_moonkin_form_passive : public AuraScript
...
+        // Glyph of Nourish
+        if (AuraEffect const* aurEff = caster->GetAuraEffect(SPELL_DRUID_GLYPH_OF_NOURISH, EFFECT_0))
+        {
+            uint32 auraCount = 0;
+            int32 heal = GetHitHeal();
+
+            Unit::AuraEffectList const& periodicHeals = GetHitUnit()->GetAuraEffectsByType(SPELL_AURA_PERIODIC_HEAL);
+            for (AuraEffect const* hot : periodicHeals)
+            {
+                if (caster->GetGUID() == hot->GetCasterGUID())
```
ccrs: I know that in the original impl there was none, but, should we actually filter here for druid spells only?

jackpoz:
Thanks for the PR :)

PR #24058, opened 2020-01-13
Fixed PvP Combat timer never getting refreshed

### LINKED ISSUE #22851: PvP Combat timer never gets refreshed while meleeing
**Description:**

PvP combat will always expire after ~5s, timer never gets refreshed even if you're currently attacking in melee.

**Current behaviour:**

(description)

**Expected behaviour:**

PvP Combat timer should get reset to max duration after each attack.

1. Get two characters to duel
2. Get character A to melee character B
3. See that even if A hits every time, combat will still expire after ~5s. 
(Players will still re enter combat after the next hit)

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 
e6f9f182019b1e194d840e78c4b2583b2b57b0bd 

**Operating system:** Windows x64

**Additional notes:**
It seems to fail here:
https://github.com/TrinityCore/TrinityCore/blob/2b332d570708e1cb762d008e3e487992272c5cd1/src/server/game/Entities/Unit/Unit.cpp#L8167-8168
The refresh would happen if the code reached the SetInCombatWith a few lines below.

kelno:
**Changes proposed:**

-  Small tweak to allow combat timer to be refreshed in PvP.  

**Target branch(es):**

- [x] 3.3.5

**Issues addressed:** 
Closes #22851

Code running on sunstrider core for about a year. (sunstrider being basically TC for TBC at this point).

/

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8146,9 +8146,14 @@ void Unit::EngageWithTarget(Unit* enemy)
         return;
 
     if (CanHaveThreatList())
-        m_threatManager.AddThreat(enemy, 0.0f, nullptr, true, true);
+    {
+        if (!IsEngagedBy(enemy))
+            GetThreatManager().AddThreat(enemy, 0.0f, nullptr, true, true);
+    }
     else
+    {
         SetInCombatWith(enemy);
+    }
 }
 
 void Unit::SetImmuneToAll(bool apply, bool keepCombat)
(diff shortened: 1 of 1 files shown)


kelno:
PvP combat needs other rework as well, the combat manager is always working on the hypothesis that combat is always mutual, which is actually untrue in PvP. Example: If your target drop combat by vanishing it shouldn't affect your own combat state. A lot of more serious cases are affected by this but this one is easier to explain.

jackpoz:
@Treeston please take a look at this when you have time

jackpoz:
@kelno: actually you missed some code from https://github.com/SunstriderEmu/BugTracker/commit/06d99c5e8fbf34ad3bb8968b586ff88ade657b69

jackpoz:
I applied the feedback provided by Treeston

Treeston (approved):

Killyana:
Tested and works fine.

jackpoz:
Thanks for the PR :)

Issue #21078, opened 2017-12-18
Investigation on MiscValue at SPELL_EFFECT_ACTIVATE_OBJECT


Mihc3:
Hi! Since MiscValue is marked as "needs research" in Spell::EffectActivateObject, I'll share what I've found out so far.

I'm not sure how to handle some already written cases below, but I tried. 😆 
```c++
void Spell::EffectActivateObject(SpellEffIndex /*effIndex*/)
{
    if (effectHandleMode != SPELL_EFFECT_HANDLE_HIT_TARGET)
        return;

    if (!gameObjTarget)
        return;

    ScriptInfo script;
    script.command = SCRIPT_COMMAND_ACTIVATE_OBJECT;

    int32 action = effectInfo->MiscValue;
    int32 value = effectInfo->BasePoints;
    int32 id = effectInfo->MiscValueB;

    switch (action)
    {
        case 1: // Animation?
            script.command = SCRIPT_COMMAND_PLAY_ANIMKIT;
            script.PlayAnimKit.AnimKitID = value;
            return;
        case 2: // Simply activates?
            // 28373 Communique, Necropolis-to-Proxies
            // 154003 Debug - Garrison - Activate Object - Custom 1
            break;
        case 3: // Maybe user damages the GO? BasePoints = 0 + spellPower * 1
            // Firework objects here.
            break;
        case 4: // Only: 8202 Sapta Sight. BasePoints = 0
            break;
        case 5: // Sets some stat? BasePoints = 0 + spellPower * 1
            // Some kind of a throw (to destTarget?).
            break;
        case 7: // Sets health, dmg? Targets nearby GO. BasePoints = 0 + spellPower * 1
            // 24934 Summon RC Tank
            // 25005 Summon RC Blimp
            break;
        case 8: // Breaks ice. Trap door opens. Lights (Dalaran?) lamp.
            break;
        case 9: // Opens door. Open Doors.
            gameObjTarget->SetGoState(GO_STATE_ACTIVE);
            return;
        case 10: // Explosion. Trap door closes.
            gameObjTarget->SetGoState(GO_STATE_READY); // Not sure.
            return;
        case 12: // Puts on fire?
            break;
        case 15: // Despawn.
            gameObjTarget->SetLootState(GO_JUST_DEACTIVATED);
            return;
        case 16: // Deactivate (switched ON, switch OFF disabled).
            gameObjTarget->SetLootState(GO_READY);
[...]

PR #24062, opened 2020-01-14
Core/Gameobject: Rename GO_STATE_ACTIVE_ALTERNATIVE to GO_STATE_DESTROYED

Faq:
**Changes proposed:**
Change to proper name of state: Rename GO_STATE_ACTIVE_ALTERNATIVE to GO_STATE_DESTROYED
by xvwyh

- [x] 3.3.5
- [ ] master

ref #21078

none needed

- [ ] 
- [ ]

### DIFF
--- a/src/server/shared/SharedDefines.h
+++ b/src/server/shared/SharedDefines.h
@@ -1650,7 +1650,7 @@ enum GOState : uint8
 {
     GO_STATE_ACTIVE             = 0,                        // show in world as used and not reset (closed door open)
     GO_STATE_READY              = 1,                        // show in world as ready (closed door close)
-    GO_STATE_ACTIVE_ALTERNATIVE = 2                         // show in world as used in alt way and not reset (closed door open by cannon fire)
+    GO_STATE_DESTROYED          = 2                         // show in world as used in alt way and not reset (closed door open by cannon fire)
 };
 
 #define MAX_GO_STATE              3
--- a/src/server/shared/SharedDefines.h
+++ b/src/server/shared/SharedDefines.h
@@ -1650,7 +1650,7 @@ enum GOState : uint8
 {
     GO_STATE_ACTIVE             = 0,                        // show in world as used and not reset (closed door open)
     GO_STATE_READY              = 1,                        // show in world as ready (closed door close)
-    GO_STATE_DESTROYED          = 2                         // show in world as used in alt way and not reset (closed door open by cannon fire)
+    GO_STATE_DESTROYED          = 2                         // show the object in-game as already used and not yet reset (e.g. door opened by a cannon blast)
 };
 
 #define MAX_GO_STATE              3
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -1459,7 +1459,7 @@ void GameObject::SwitchDoorOrButton(bool activate, bool alternative /* = false *
         RemoveFlag(GAMEOBJECT_FLAGS, GO_FLAG_IN_USE);
 
     if (GetGoState() == GO_STATE_READY)                      //if closed -> open
-        SetGoState(alternative ? GO_STATE_ACTIVE_ALTERNATIVE : GO_STATE_ACTIVE);
+        SetGoState(alternative ? GO_STATE_DESTROYED : GO_STATE_ACTIVE);
     else                                                    //if open -> close
         SetGoState(GO_STATE_READY);
 }
--- a/src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
+++ b/src/server/scripts/EasternKingdoms/Deadmines/instance_deadmines.cpp
@@ -175,7 +175,7 @@ class instance_deadmines : public InstanceMapScript
             {
                 if (GameObject* pIronCladDoor = instance->GetGameObject(IronCladDoorGUID))
                 {
-                    pIronCladDoor->SetGoState(GO_STATE_ACTIVE_ALTERNATIVE);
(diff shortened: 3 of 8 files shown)


src/server/shared/SharedDefines.h
```diff
@@ -1650,7 +1650,7 @@ enum GOState : uint8
 {
     GO_STATE_ACTIVE             = 0,                        // show in world as used and not reset (closed door open)
     GO_STATE_READY              = 1,                        // show in world as ready (closed door close)
-    GO_STATE_ACTIVE_ALTERNATIVE = 2                         // show in world as used in alt way and not reset (closed door open by cannon fire)
+    GO_STATE_DESTROYED          = 2                         // show in world as used in alt way and not reset (closed door open by cannon fire)
```
illfated: Since this is essentially all that is being changed here in this PR, I recommend using a precise description in the comment line, to avoid confusion or misinterpretation.
Preliminary suggestion:
```suggestion
    GO_STATE_DESTROYED          = 2                         // show the object in-game as already used and not yet reset (e.g. door opened by a cannon blast)
```
(of course, this may be a tad much, so feel free to suggest a shorter version)

Aokromes:
sometimes the GO aint destroyed instead changed of shape. that's the reason of that name.

illfated:
You mean the original name `GO_STATE_ACTIVE_ALTERNATIVE` ?

Jinnaix:
Still, the current naming is confusing. How about something like `GO_STATE_ACTIVE_DONE` or `GO_STATE_ACTIVE_ACTIVATED`?

illfated:
Fair enough. Maybe make a list of all known GO state names and compare them?

`GO_DESTRUCTIBLE_INTACT`
`GO_DESTRUCTIBLE_DAMAGED`
`GO_DESTRUCTIBLE_DESTROYED`
`GO_JUST_DEACTIVATED`
`GO_NOT_READY`
`GO_READY`
`GO_STATE_ACTIVE`
`GO_STATE_READY`

... and so on.

illfated:
> sometimes the GO aint destroyed instead changed of shape. that's the reason of that name.

Fair point. The GO is not necessary _destroyed_ as such, it can be (as in the comment) simply that a door is changed from a locked state to opened, but without the ability to go back to its previous state. `GO_STATE_DESTROYED` would be a misnomer in those instances, but I guess there are other examples of GO descriptions imprecisely named like that. I presume you will discuss any possible alternatives.

Faq:
@illfated 

> door is changed from a locked state to opened, but without the ability to go back to its previous state. 

So in RL we call such thing broken/damaged/destroyed.
GO_STATE_ACTIVE_DAMAGED
GO_STATE_DAMAGED
GO_STATE_ACTIVE_DESTROYED
GO_STATE_DESTROYED

Word ACTIVE sounds like not needed in such combination.

illfated:
> So in RL we call such thing broken/damaged/destroyed.
> GO_STATE_ACTIVE_DAMAGED
> GO_STATE_DAMAGED
> GO_STATE_ACTIVE_DESTROYED
> GO_STATE_DESTROYED
> 
> Word ACTIVE sounds like not needed in such combination.

Fair & good point.

Faq:
@xvwyh why GO_STATE_DESTROYED ?

> <%Shauren> i dont see why you want to rename it
<%Shauren> i mean, did you do some digging in client to try figure out what that state actually is?

Shauren:
This is correct. IDA verified for 3.3.5

xvwyh:
@Faq, look in https://github.com/TrinityCore/TrinityCore/issues/21078#issuecomment-420115205
The names in the comments are taken directly from Wow.exe files, and they've been present there since alpha till at least 7.2.3 (the latest version I had downloaded), growing in number as new actions were added. These are the names of all the actions that blizzard's scripts and spells can perform on GOs. You'll see that there are a bunch of actions to play custom animations, then actions that deal with `GO_FLAG_LOCKED`, then actions that deal with opening and closing of doors, and then suddenly "Destroy" and "Rebuild". These action have been in the game since at least 0.5.3 alpha, loooong before destructible buildings were introduced into the game with WotLK, so we conclude that it's not related to them. If you were to search for the list of spells with `SPELL_EFFECT_ACTIVATE_OBJECT` and MiscValue=12 ("Destroy") you would find spells like 11195 "Blow Zul'Farrak Door". Guess what can happen to the door leading to the final boss in Zul'Farrak? [Weedli Blastfuse](https://wow.gamepedia.com/Weegli_Blastfuse) can blow up the door at players' request, changing its state to `GO_STATE_ACTIVE_ALTERNATIVE`, which puts the door into a visually destroyed state. Connect the dots together and you should come to the same conclusion as I did.

Faq:
@xvwyh thnx for clarification
```
enum class GameObjectState : uint8
{
     Open        = 0,
     Closed      = 1,
     Destroyed   = 2
}
```
Therefore those 2 are not precise names. @Shauren  should I rename them too? Or something else here missing and that's why not merged yet?
    GO_STATE_ACTIVE -> GO_STATE_OPEN
    GO_STATE_READY -> GO_STATE_CLOSED

xvwyh:
I imagine those are sprinkled around the codebase so much, that changing them at this point is just pointless...

jackpoz:
is this ready to be merged ?

illfated:
Looks like it is ready, based on the fact that it is only a renaming of the enum name and that there are no strong arguments that it should not be changed.

jackpoz:
@illfated https://en.wikipedia.org/wiki/Law_of_triviality is kicking in in this PR

jackpoz:
Thanks for the PR :)

Issue #22986, opened 2019-01-24
Core: Visibility distance calculation


Killyana:
**Description:** The visibility distance calculation is wrong for gobs and creatures.

For this reason we get some weird result like this: https://imgur.com/a/88vi0XA and visibility distance is different from retail.

Visibility range:
100 - (Visibility.Distance.Continents)
170 - (Visibility.Distance.Instances)
533 - (Visibility.Distance.BGArenas)

visibilityDistanceType:
#define VISIBILITY_DISTANCE_GIGANTIC         400.0f
#define VISIBILITY_DISTANCE_LARGE               200.0f
#define VISIBILITY_DISTANCE_NORMAL          100.0f
#define VISIBILITY_DISTANCE_SMALL               50.0f
#define VISIBILITY_DISTANCE_TINY                   25.0f

The visibility distance must be calculated like this for game objects:
Final Visibility range = Visibility range + radius of the game object+ visibilityDistanceType

*visibilityDistanceType=0 the gob doesn't have any related data

Example: https://woehead.way-of-elendil.fr/?object=194484

This gob has: radius =200 visibility=200 and it' spawned into an instance so default view distance=170

This means final visibility distance= 200+200+170 =570m ( this is confirmed on retail)

For creature the radius could probably be ignored and just use this formula:
visibility= visibilityDistanceType  + default distance

This also explains the Tiny and small visibility distance used by some gobs.

**Branch(es):** both

**TC rev. hash/commit:** rev. 717e050edb73

xvwyh:
For instances and battlegrounds these visibility distances are most definitely different. In SotA sniffs, when you arrive on boats, I saw objects with Gigantic property (sigils, e.g. 192685) appearing as far away as 1450 yards, maybe even further. NPCs with Large AOI flag (23472, verifiable through CreatureDifficulty dumps Kinzcool posted recently) became visible as far as 1080 yards. One could make some sort of sniff analyzer that would read packets sequentially, keep track of player's position, and record which objects became visible at what distance.

Kittnz:
Yeah, it's not really clear at this point. Who said there may not be a specific distance difference for a specific bg or instances?

Faq:
Anything here left after https://github.com/TrinityCore/TrinityCore/commit/b0db728c49f0b70d2c4f18270c389e4445161b7f ?

Killyana:
[...]

Issue #22984, opened 2019-01-24
Core/GameObject: Visiblity distance


Killyana:
**Description:** On 3.3.5 some Gameobjects must have increased visibility distance.

GAMEOBJECT_TYPE_DOOR:              
GAMEOBJECT_TYPE_FLAGSTAND:         
GAMEOBJECT_TYPE_FLAGDROP:          
GAMEOBJECT_TYPE_DUNGEON_DIFFICULTY:
GAMEOBJECT_TYPE_TRAPDOOR:          

The visibility must be set by Gob type and not by the boolean: large gob (they doesn't have it in 3.3.5 anyway)

**Branch(es):**3.3.5

**TC rev. hash/commit:** rev. 717e050edb73

Killyana:
On retail the visibility type for doors is infinite.

Kittnz:
Retail has a lot of visibility changes for gobs, look at the data that some gobs have. it's totally different from 335.

offl:
Logically creature and gameobject visibility must be equal. So player always can see both creature and door behind creature.
https://youtu.be/OM-TmH_3UVM?t=142

Killyana:
Let's take Uduar as example:

This gobs are doors and visible from almost any distance (infinite):

[Flame Leviathan Gate] GUID: 35413 ID: 194630
[Mimiron's Targetting Crystal] GUID: 45229 ID: 194705
[Thorim's Targetting Crystal] GUID: 45230 ID: 194706
[Freya's Targetting Crystal] GUID: 45225 ID: 194704
[Hodir's Targetting Crystal] GUID: 45500 ID: 194707
[Hodir's Storm Generator] GUID: 43114 ID: 194665
[Thorim's Weather Generator] GUID: 44883 ID: 194666
[Mimiron's Storm Generator] GUID: 43113 ID: 194664
[Freya's Storm Generator] GUID: 42960 ID: 194663

On master they have: 
data8: Door is opaque= 1
data9: Gigantic AOI =0
data10: Infinite AOI =0
On 3.3.5 nothing related to visibility distance

I think blizzard implemented visibility range for doors like other types after wotlk, and they used Door is opaque before adding Gigantic AOI and Infinite AOI  in later extensions.

At the moment this gobs have normal visibility in both branchs.

Some videos from wotlk where we can see this gobs from far away: 
https://youtu.be/hYGwNFtJUDM?t=36  -- crystals
[...]

PR #24074, opened 2020-01-17
Core/Object: Add infinite gob check

Kittnz:
- Add default visibility range with distance override
- Add Infinite gameobject check

- [x] 3.3.5
- [x] master

https://github.com/TrinityCore/TrinityCore/issues/22986
https://github.com/TrinityCore/TrinityCore/issues/22984


Killyana:
Tested IG, everything looks good.

Killyana:
Revert the addition for the visibility distance and keep only the gob change, I will try to get correct data from retail, as current one are wrong

PR #24075, opened 2020-01-18
Scripts/Quests: Reports from the Field

Sorikoff:
**Changes proposed:**
Changed script so that it uses gossips from DB. Also moved it to new registration system.

**Target branch(es):** 3.3.5

~**Issues addressed:**~

**Tests performed:** Works as intended

~**Known issues and TODO list:**~


illfated:
Thank you for making this change. I was thinking about making this change myself, but had not found enough time to get started yet. Well done. 👍

jackpoz:
Thanks for the PR :)

PR #24083, opened 2020-01-19
Core/PathGenerator: Fix path generator returning shortcuts when start and end are on the same polygon

jackpoz:
**Changes proposed:**

Second try of handling paths on same poly in a better way than just building a shortcut.
- Skip looking for closest point on polygon boundary when path is on same poly
- Add a new status PATHFIND_FARFROMPOLY for callers to decide what to do if the start or end positions are far from mmap polygon, i.e. player on unwalkable mesh reached with flying mount
- Skip paths that have PATHFIND_FARFROMPOLY in ConfusedMovementGenerator<T>, FleeingMovementGenerator<T> and RandomMovementGenerator<Creature> . The movement generator will have to either try again to look for another path if end position is far from the polygon, otherwise the Unit will not move at all and just stand still if start position is far from the polygon. This fixes Players falling under meshes when being on unwalkable meshes, i.e. pillars in Blade's Edge Arena in Outland map (not the actual Arena map).

- [x] 3.3.5
- [ ] master

- [x] Charge from .go xyz -3268.020752 -2486.342529 20.014091 0 to .go xyz -3282.670654 -2478.257324 24.3308734 0
- [x] .go xyz 1291.276733, -4012.389648, 50.746246 1 and .aura 52061 you will go through the map
- [x]  .go xyz 1362.246704, -4174.677246, 95.796837 1 and .aura 52061 you will go through the map
- [x] .go xyz 2849.104004, 5943.180176, 11.228345 530 and .aura 52061 you will go through the map

### DIFF
--- a/src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/ConfusedMovementGenerator.cpp
@@ -107,7 +107,9 @@ bool ConfusedMovementGenerator<T>::DoUpdate(T* owner, uint32 diff)
         }
 
         bool result = _path->CalculatePath(destination.GetPositionX(), destination.GetPositionY(), destination.GetPositionZ());
-        if (!result || (_path->GetPathType() & PATHFIND_NOPATH) || (_path->GetPathType() & PATHFIND_SHORTCUT))
+        if (!result || (_path->GetPathType() & PATHFIND_NOPATH)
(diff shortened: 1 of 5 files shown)


Killyana:
For me it looks fine, there's some issues about maxWalkableAngle, but we can change it in a future PR once we have more info

illfated:
Cosmetic/semantics: Not sure if it is caused by autocorrect, but in your "Tests performed:" section you may want to replace "throw" with "through" as in "you will go through the map".

PR #24093, opened 2020-01-22
Core/Movement: improved random movement pause handling to get closer to retail behaivior

Ovahlord:
**Changes proposed:**
-  tweaked the pause handling for random movements to get closer to retail behaivior. Instead of going with a plain 50% chance of triggering a waiting moment we will now pre-determine a number of splines that must be launched before a pause time window kicks in. While at it I also corrected the pause timer.
By the looks of it I would say I have pretty much nailed it so yeah. 110% blizzlike.

- [x] 3.3.5
- [ ] master

- tested ingame on my 434 fork and of course compared it against retail

### DIFF
--- a/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
@@ -171,7 +171,7 @@ void RandomMovementGenerator<Creature>::SetRandomLocation(Creature* owner)
     init.SetWalk(walk);
     int32 splineDuration = init.Launch();
 
-    _wanderSteps--;
+    --_wanderSteps;
     if (_wanderSteps) // Creature has yet to do steps before pausing
         _timer.Reset(splineDuration);
     else
--- a/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
@@ -25,7 +25,7 @@
 #include "Random.h"
 
 template<class T>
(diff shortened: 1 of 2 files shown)


src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
```diff
@@ -168,8 +169,17 @@ void RandomMovementGenerator<Creature>::SetRandomLocation(Creature* owner)
     Movement::MoveSplineInit init(owner);
     init.MovebyPath(_path->GetPath());
     init.SetWalk(walk);
-    time_t traveltime = init.Launch();
-    _timer.Reset(traveltime + resetTimer);
+    int32 splineDuration = init.Launch();
+
+    _wanderSteps--;
```
ccrs: nitpicking master attacks again 💃 

by default its better to use inline modifying operators, or w/e the name is, as a pre-operation
I'm making up all these names, I know

I mean ```--_wanderSteps;``` is better than ```_wanderSteps--;```

ccrs: https://en.cppreference.com/w/cpp/language/operator_incdec

meh, I didnt go too far from the real term 💃

ccrs: > Because a temporary copy of the object is constructed during post-increment and post-decrement, pre-increment or pre-decrement operators are usually more efficient in contexts where the returned value is not used.

There, that's why

Kittnz:
You might wanna add a random walk/run toggle too

Ovahlord:
outside of the scope of this pr

ccrs:
agreed

PR #24100, opened 2020-01-24
Core/PacketIO: convert some packets to to new packet class

ForesterDev:
**Changes proposed:**

-  Convert to new packet class packets:
-  SMSG_INVALIDATE_PLAYER
-  SMSG_BINDER_CONFIRM
-  SMSG_BIND_POINT_UPDATE
-  SMSG_PLAYER_BOUND
-  SMSG_LOGIN_SET_TIME_SPEED

Based on master branch

- [x] 3.3.5

**Tests performed:** (Does it build, tested in-game, etc.) Build, tested in-game


jackpoz:
for future reference I think it would be easier to have each packet on its own PR for easier review and faster merge

src/server/game/Entities/Player/Player.cpp
```diff
@@ -22592,52 +22601,62 @@ void Player::SetGroup(Group* group, int8 subgroup)
...
+    /// SMSG_EQUIPMENT_SET_LIST
     SendEquipmentSetList();
 
-    data.Initialize(SMSG_LOGIN_SETTIMESPEED, 4 + 4 + 4);
-    data.AppendPackedTime(GameTime::GetGameTime());
-    data << float(0.01666667f);                             // game speed
-    data << uint32(0);                                      // added in 3.1.2
-    SendDirectMessage(&data);
+    /// SMSG_LOGIN_SET_TIME_SPEED
+    static float const TimeSpeed = 0.01666667f;
```
ForesterDev: I have a suggestion: move this to own function in WorldSession class and call this method here

Shauren: do not do that in port-only commits as these would be skipped when going in the other direction

jackpoz:
This PR affects quite a few packets. Please keep an eye open in the issue list if any user reports issues related to these packets and fix the reported issues :)

jackpoz:
Thanks for the PR :)

PR #24112, opened 2020-01-28
Core/Vehicles: implement vehicle seat addon table to specify seat ori…

Ovahlord:
**Changes proposed:**

-  implement a table to specify special vehicle seat behaivior
-  the current implementation comes with the following features:

1. a field to override vehicle seat attachment orientations. This allows people to change vehicle seat orientations by default to make them look into different directions. A commonly used feature, especially in Cataclysm

2. fields to handle passenger exit destinations properly. Instead of using a ugly ass scripting hook we can now specify either exit position offsets or entire coordinates which gives us full control over the vehicle exit destinations

**Target branch(es):**
- [x] 3.3.5

**Tests performed:**
- tested on my 434 fork in multiple occasions: Magmaw and his pincers, Traveler's Tundra Mammoth, Ozumat in Throne of the Tides and a Stormwind Charger Horse involved in the cata version of the quest 'Further Concerns'

- [ ] Snap ModifyVehiclePassengerExitPos out of existance. You don't want this hook.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12545,6 +12545,7 @@ void Unit::_ExitVehicle(Position const* exitPosition)
         return;
 
     // This should be done before dismiss, because there may be some aura removal
+    VehicleSeatAddon const seatAddon = m_vehicle->GetSeatAddonForSeatOfPassenger(this);
     Vehicle* vehicle = m_vehicle->RemovePassenger(this);
 
     Player* player = ToPlayer();
@@ -12575,7 +12576,15 @@ void Unit::_ExitVehicle(Position const* exitPosition)
         // Set exit position to vehicle position and use the current orientation
         pos = vehicle->GetBase()->GetPosition();
         pos.SetOrientation(GetOrientation());
+
+        // To-do: snap this hook out of existance
         sScriptMgr->ModifyVehiclePassengerExitPos(this, vehicle, pos);
+
+        // Change exit position based on seat entry addon data
+        if (seatAddon.ExitParameter == VehicleExitParameters::VehicleExitParamOffset)
+            pos.RelocateOffset({ seatAddon.ExitParameterX, seatAddon.ExitParameterY, seatAddon.ExitParameterZ, seatAddon.ExitParameterO });
+        else if (seatAddon.ExitParameter == VehicleExitParameters::VehicleExitParamDest)
+            pos.Relocate({ seatAddon.ExitParameterX, seatAddon.ExitParameterY, seatAddon.ExitParameterZ, seatAddon.ExitParameterO });
     }
 
     float height = pos.GetPositionZ() + vehicle->GetBase()->GetCollisionHeight();
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12545,7 +12545,7 @@ void Unit::_ExitVehicle(Position const* exitPosition)
         return;
 
     // This should be done before dismiss, because there may be some aura removal
-    VehicleSeatAddon const seatAddon = m_vehicle->GetSeatAddonForSeatOfPassenger(this);
+    VehicleSeatAddon const* seatAddon = m_vehicle->GetSeatAddonForSeatOfPassenger(this);
     Vehicle* vehicle = m_vehicle->RemovePassenger(this);
 
     Player* player = ToPlayer();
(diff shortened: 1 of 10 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12545,6 +12545,7 @@ void Unit::_ExitVehicle(Position const* exitPosition)
         return;
 
     // This should be done before dismiss, because there may be some aura removal
+    VehicleSeatAddon const seatAddon = m_vehicle->GetSeatAddonForSeatOfPassenger(this);
```
Warpten: temporary copy

src/server/game/Entities/Vehicle/Vehicle.cpp
```diff
@@ -41,7 +41,8 @@ UsableSeatNum(0), _me(unit), _vehicleInfo(vehInfo), _creatureEntry(creatureEntry
         if (uint32 seatId = _vehicleInfo->m_seatID[i])
             if (VehicleSeatEntry const* veSeat = sVehicleSeatStore.LookupEntry(seatId))
             {
-                Seats.insert(std::make_pair(i, VehicleSeat(veSeat)));
+                VehicleSeatAddon addon = sObjectMgr->GetVehicleSeatAddon(seatId);
```
Warpten: temporary copy

src/server/game/Entities/Vehicle/Vehicle.cpp
```diff
@@ -345,6 +346,30 @@ SeatMap::const_iterator Vehicle::GetNextEmptySeat(int8 seatId, bool next) const
...
+ *
+ * @author Ovahlord
+ * @date 28-1-2020
+ *
+ * @param passenger Identifier for the current seat user
+ *
+ * @return The seat addon data for the currently used seat of a passenger
+ */
+
+VehicleSeatAddon const Vehicle::GetSeatAddonForSeatOfPassenger(Unit const* passenger) const
```
Warpten: temporary copy

src/server/game/Entities/Vehicle/Vehicle.cpp
```diff
@@ -809,6 +834,7 @@ bool VehicleJoinEvent::Execute(uint64, uint32)
     Passenger->RemoveAurasByType(SPELL_AURA_MOUNTED);
 
     VehicleSeatEntry const* veSeat = Seat->second.SeatInfo;
+    VehicleSeatAddon const veSeatAddon = Seat->second.SeatAddon;
```
Warpten: temporary copy

Warpten (changes requested):

src/server/game/Entities/Vehicle/VehicleDefines.h
```diff
@@ -65,16 +73,32 @@ struct PassengerInfo
...
+    float ExitParameterY = 0.f;
+    float ExitParameterZ = 0.f;
+    float ExitParameterO = 0.f;
+    VehicleExitParameters ExitParameter = VehicleExitParameters::VehicleExitParamNone;
+};
+
 struct VehicleSeat
 {
-    explicit VehicleSeat(VehicleSeatEntry const* seatInfo) : SeatInfo(seatInfo)
+    explicit VehicleSeat(VehicleSeatEntry const* seatInfo, VehicleSeatAddon const seatAddon) : SeatInfo(seatInfo), SeatAddon(seatAddon)
```
Warpten: more copies

Warpten:
Missing removal of npc script

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12575,7 +12576,15 @@ void Unit::_ExitVehicle(Position const* exitPosition)
         // Set exit position to vehicle position and use the current orientation
         pos = vehicle->GetBase()->GetPosition();
         pos.SetOrientation(GetOrientation());
+
+        // To-do: snap this hook out of existance
         sScriptMgr->ModifyVehiclePassengerExitPos(this, vehicle, pos);
```
Mesielecat: I don't think, that this hook is necesary to remove, could exist some vehicle where the exit position could be changed in a dynamic way.

Ovahlord: That hook, or better said all scripts of the scriptmgr that use the same method are absolutely just custom script features. This hook is being slapped on every single vehicle. The more scripts use it, the more scripts will be called with every single vehicle that's ejecting a passenger. That's absolutely not okay and should get dropped as soon as possible. We have a proper way now.

sirikfoll:
Very nice feature!
Could this also include the x, y, z for the `attachmentOffset`? (yes, i know those should be extracted from client's m2 bone attachment, but until..)

PR #24133, opened 2020-02-03
Core/Creature: Use proper name for wander distance.

ratkosrb:
**Changes proposed:**

Change the misleading name used for the random movement generator's max wandering distance. For some reason it was named "respawn radius" in the Creature class, even though it is nothing of the sort. There is no such thing as a respawn radius.

- [x] 3.3.5
- [x] master

- None


illfated:
Hm, fair enough that the original name lacks an ideal description. Speaking for myself, I have always interpreted the original name as "circular area of movement with given radius length from (re)spawn point". Maybe I am just being conservative, but I struggle a bit to see the description improvement. Is it obvious that creatures with limited movement distance will choose random angles within a full circle?

ratkosrb:
The misleading part is the respawn part, not the radius part. The current name makes it seem as if the respawn point can vary and be anywhere in this radius, when there is no such thing in the game.

The only place this value is used is Random Movement Generator, and it is called "wanderDistance" there, so I simply updated the name in Creature class to match that.

illfated:
Fair enough. I presume it will be acceptable to seasoned coders. I just wanted to share my thoughts as a long time TC source user and a small contributor.

Kittnz:
I would also change the column name for spawndist. To make it more clear DB side as well.

ccrs:
yeah, also check commands texts, trinity strings and such

src/server/scripts/Commands/cs_npc.cpp
```diff
@@ -163,7 +163,7 @@ class npc_commandscript : public CommandScript
             { "model",      rbac::RBAC_PERM_COMMAND_NPC_SET_MODEL,     false, &HandleNpcSetModelCommand,         "" },
             { "movetype",   rbac::RBAC_PERM_COMMAND_NPC_SET_MOVETYPE,  false, &HandleNpcSetMoveTypeCommand,      "" },
             { "phase",      rbac::RBAC_PERM_COMMAND_NPC_SET_PHASE,     false, &HandleNpcSetPhaseCommand,         "" },
-            { "spawndist",  rbac::RBAC_PERM_COMMAND_NPC_SET_SPAWNDIST, false, &HandleNpcSetSpawnDistCommand,     "" },
+            { "wander_distance",  rbac::RBAC_PERM_COMMAND_NPC_SET_SPAWNDIST, false, &HandleNpcSetSpawnDistCommand,     "" },
```
jackpoz: we rarely have _ in commands, maybe remove the character ?

ratkosrb: I removed it.

jackpoz:
Could you do one last check ingame to ensure everything works ? Chat commands included

ratkosrb:
Tested it in game and found one more text that had to be updated. Should be all good now.

illfated:
I know, cosmetic detail, but would you mind realigning the internal columns in `static std::vector<ChatCommand> npcSetCommandTable` in `cs_npc.cpp` ?

sql/updates/world/3.3.5/2020_99_99_99_world_335.sql
```diff
@@ -0,0 +1,7 @@
+-- Use more appropriate name for distance used by random movement generator.
+ALTER TABLE `creature`
+	CHANGE COLUMN `spawndist` `wander_distance` FLOAT NOT NULL DEFAULT '0' AFTER `spawntimesecs`;
+
+-- Update name used by chat command.
+UPDATE `trinity_string` SET `content_default`='Wander distance changed to: %f' WHERE `entry`=297;
+UPDATE `command` SET `name`='npc set wanderdistance', `help`='Syntax: .npc set spawndist #dist\r\n\r\nAdjust wander distance of selected creature to dist.' WHERE `name`='npc set spawndist';
```
jackpoz: There is a spawndist in this line still

ratkosrb: Fixed it.

illfated:
Good alignment. Thank you.

jackpoz:
Thanks for the PR :)

PR #24134, opened 2020-02-04
Use boost::optional and boost::none instead of smelly pointer

Lillecarl:
**Changes proposed:**

-  Use boost::optional and boost::none instead of smelly pointer,
reasoning behind this must've been to allow 0 to be a usable value as well. boost does this better.

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**
Codesmell

**Tests performed:**
Builds, works ingame (tested the usage on the function in another branch of mine)

- [ ] bump minimum C++ version to C++17 and use the std variant instead (wontfix)
- [ ] look for other smells like this (one PR at a time)

### DIFF
--- a/src/server/game/AI/CoreAI/GameObjectAI.h
+++ b/src/server/game/AI/CoreAI/GameObjectAI.h
@@ -52,7 +52,7 @@ class TC_GAME_API GameObjectAI
         static int32 Permissible(GameObject const* go);
 
         // Called when the dialog status between a player and the gameobject is requested.
-        virtual Optional<QuestGiverStatus> GetDialogStatus(Player* /*player*/) { return boost::none; }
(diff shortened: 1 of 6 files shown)


Lillecarl:
https://github.com/TrinityCore/TrinityCore/pull/24134 When this is merged we can replace boost::option with std::option

jackpoz:
TC has a Optional<T> type for this scenario

Shauren:
Not only you should use Optional type (#include "Optional.h", actually a wrapper for boost::optional currently) but also there is no need to set its default value to boost::none, simply writing `= {}` is enough to set it to empty

Lillecarl:
Force pushed an update, thanks for the guidance.

When implementing a wrapper template like this, is it to ease refactorability when it goes C++17 or what's the reasoning?

jackpoz:
Thanks for the PR :)

PR #24138, opened 2020-02-09
WIP: Modernize codebase with Clang-Tidy modernize-loop-convert

Lillecarl:
I haven't checked the fixes manually yet, one thing i noticed straight away is that they use auto & instead of auto& which should be corrected before merge.

I only ran it on the "src/" directory where actual TrinityCore code lives, dependencies shall not be messed with.

**Changes proposed:**

-  Range based loops everywhere

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:** 
none

**Tests performed:**
builds locally with clang

- [ ] 
- [ ]


jackpoz:
Test it ingame too at least

Lillecarl:
Definitely, just haven't had access to the wow client yet. I'll keep you posted.

jackpoz:
Don't use unsigned long, use our uint8/16/32/64 types

Lillecarl:
Ah yes, clang-tidy didn't give rats ass about the typedefs, I'll make a search and replace.

soulfrost:
resharper power 💃

Lillecarl:
@soulfrost Clang-Tidy, it understands C++ better as it uses the clang parsing engine to understand C++ rather than some homebrew parsing engine.

Lillecarl:
@soulfrost response to your edit: if you look at the diff you can clearly see that it has picked the variable name that was used to dereference the iterator inside the loop, how is that unacceptable?

soulfrost:
just wait for Shauren, he will give clear answer

Lillecarl:
That renders your comments 100% redundant, as you can see this is a WIP marked PR. Types will be easily corrected, then we'll await further feedback from those with constructive criticism.

soulfrost:
okay okay, clearing them now, see discord/irc

Ovahlord:
Uhm, I think we avoid auto as much as possible for the sake of readability. Having auto everywhere for each simple iterator makes it a pain in the ass to read later on.

jackpoz:
I would say running the cland-tidy tool on a subfolder would be easier to review, instead of having 159 files affected

Shauren:
Objection. It uses auto for everything

Lillecarl:
@jackpoz I used Qt Creator to run it, so honestly I've got no clue.

If you have access to a Linux or Mac box It's piece of cake, if not how would you like me to proceed?

I could make another branch and "cherry pick" just a handful of files.

Lillecarl:
@Shauren it doesn't use auto exclusively, my understanding seems to be that it uses auto when iterating an enum, unsigned int for regular for loops, and auto for iterating over containers.

If I were to split this into several prs and expand auto into the type, would it then be interesting?

Shauren:
It's probably not worth your time to manually go through all of the changes and fix them up

Also it creates loops that break our codestyle (const placement, spacing around &)

Lillecarl:
I'd like to see the codebase fresh with new features to motivate others to follow that behaviour (at least for those ugly 30 char iterators).

But your word is final, I'll leave this here if you change your mind, it's only about 1.5k fixes, so it wouldn't be too much of a hassle to do it in steps of folders if the work pays off.

jackpoz:
it would be better if we learnt how to use clang-tidy, like passing parameters that end up updating the sources in the exact way we want

Lillecarl:
@jackpoz I'll have a look when I'm home.
As you'll see here https://clang.llvm.org/extra/clang-tidy/checks/modernize-loop-convert.html is that there are barely any parameters to pass to this fix, but once it's a range loop it wouldn't change to auto if ran again.

So this is kinda a one-off, though it'd be cool to learn indeed and make circle run this and complain if there's a non range loop where it judges there can be.

Then there would be several others to deploy as well, such as not using c arrays and such.

See this list for possibilities https://clang.llvm.org/extra/clang-tidy/checks/list.html

Shauren:
I am not against using tools like this, what I don't like is the result of running it all on default settings

Lillecarl:
I would like to note that other than C++ tooling lagging a bit I do not see the need to see the type, it's redundant information considering the compiler already has all the information. Specifying type is good for literals though.

soulfrost:
loop

u can try to use Resharper c++ to converte loops, in file, folder etc, easyer to formate it later

Lillecarl:
@Shauren This particular tool has only one parameter, which is the risk factor. I submitted a link in a previous post.

Therefore to adhere to TrinityCore standards I'm willing to put the hours to expand these :)

Once implemented we could automate circleci to check these and complain if someone uses and old style iterator where a range is possible.

Lillecarl:
@soulfrost resharper can't be run in CI, also the assumption that I have windows, visual studio and resharper is wrong.

soulfrost:
look for `ReSharper Command Line Tools`

Lillecarl:
Thanks, seems like we could indeed integrate this proprietary tool in TrinityCores otherwise open-source tool chain.

Lillecarl:
https://clang.llvm.org/extra/clang-tidy/checks/performance-implicit-conversion-in-loop.html

A clang fix relevant to this case.

src/server/game/Battlegrounds/BattlegroundMgr.cpp
```diff
@@ -133,13 +133,13 @@ void BattlegroundMgr::Update(uint32 diff)
...
 
-        for (uint8 i = 0; i < scheduled.size(); i++)
+        for (unsigned long i : scheduled)
         {
-            uint32 arenaMMRating = scheduled[i] >> 32;
-            uint8 arenaType = scheduled[i] >> 24 & 255;
-            BattlegroundQueueTypeId bgQueueTypeId = BattlegroundQueueTypeId(scheduled[i] >> 16 & 255);
-            BattlegroundTypeId bgTypeId = BattlegroundTypeId((scheduled[i] >> 8) & 255);
-            BattlegroundBracketId bracket_id = BattlegroundBracketId(scheduled[i] & 255);
+            uint32 arenaMMRating = i >> 32;
```
illfated:
```cpp
C:\projects\trinitycore\src\server\game\Battlegrounds\BattlegroundMgr.cpp(138): error C2220: warning treated as error - no 'object' file generated [C:\projects\trinitycore\build\src\server\game\game.vcxproj]
C:\projects\trinitycore\src\server\game\Battlegrounds\BattlegroundMgr.cpp(138): warning C4293: '>>': shift count negative or too big, undefined behavior [C:\projects\trinitycore\build\src\server\game\game.vcxproj]
```

Lillecarl: Can you share your cmake params so I can steal and use the same?

This is very weird code, I'll look into it

illfated: This is based on the CMake parameters used in AppVeyor:
https://ci.appveyor.com/project/DDuarte/trinitycore/builds/30668901#L9

Shauren: This breaks because clang and msvc disagree on size of unsigned long.
For clang `unsigned long` is 64 bits (uint64)
For msvc `unsigned long` is 32 bits (uint32)

Clearly not using our uint64 typedef (which is guaranteed 64 bits on all platforms) breaks it

illfated: Also notice that AppVeyor is set to make & build a 32 bits core.

Shauren: No, it is not. Don't spread misinformation

illfated: OK, well, I see. so `DCMAKE_C_FLAGS="/DWIN32` is irrelevant, I guess?

Shauren: It is. That define is a generic compile definition that various third party libraries use to detect that they are compiled for windows, not specifically 32 bit windows

Lillecarl: I'm still willing to go through the diff and correct these where they're wrong, along with the autos. If you want the change in TC once it's done (probably will make other PR's per folder or smth to get less bulk work, about 150 changes each or smth).

jackpoz:
160 files changed makes it quite impossible to review

Lillecarl:
@jackpoz Don't think about this one for now. Once I'm done I'll split it up for easier review :)

Lillecarl:
Closing this in favor of the the smaller batches.
https://github.com/TrinityCore/TrinityCore/pull/24164
https://github.com/TrinityCore/TrinityCore/pull/24165

PR #24140, opened 2020-02-09
clang-tidy readability-delete-null-pointer

Lillecarl:
**Changes proposed:**
-  Remove unneeded code

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**
none

**Tests performed:**
builds

- [x] none

### DIFF
--- a/src/server/game/Guilds/Guild.cpp
+++ b/src/server/game/Guilds/Guild.cpp
@@ -2400,8 +2400,8 @@ void Guild::DeleteMember(SQLTransaction& trans, ObjectGuid guid, bool isDisbandi
     // Call script on remove before member is actually removed from guild (and database)
     sScriptMgr->OnGuildRemoveMember(this, player, isDisbanding, isKicked);
 
-    if (Member* member = GetMember(guid))
-        delete member;
+    delete (GetMember(guid));
+
     m_members.erase(lowguid);
 
     // If player not online data in data field will be loaded from guild tabs no need to update it !!
--- a/src/server/game/Guilds/Guild.cpp
+++ b/src/server/game/Guilds/Guild.cpp
@@ -2400,9 +2400,12 @@ void Guild::DeleteMember(SQLTransaction& trans, ObjectGuid guid, bool isDisbandi
     // Call script on remove before member is actually removed from guild (and database)
     sScriptMgr->OnGuildRemoveMember(this, player, isDisbanding, isKicked);
 
-    delete (GetMember(guid));
-
-    m_members.erase(lowguid);
+    auto memberItr = m_members.find(guid);
+    if (memberItr != m_members.end())
+    {
+        delete memberItr->second;
+        m_members.erase(memberItr);
+    }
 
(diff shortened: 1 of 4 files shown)


src/server/game/Guilds/Guild.cpp
```diff
@@ -2400,8 +2400,8 @@ void Guild::DeleteMember(SQLTransaction& trans, ObjectGuid guid, bool isDisbandi
     // Call script on remove before member is actually removed from guild (and database)
     sScriptMgr->OnGuildRemoveMember(this, player, isDisbanding, isKicked);
 
-    if (Member* member = GetMember(guid))
```
Shauren: To prevent double lookup (first in GetMember, then in .erase), refactor as
```
auto memberItr = m_members.find(guid);
if (memberItr != m_members.end())
{
    delete memberItr->second;
    m_members.erase(memberItr);
}
```

Shauren (approved):

src/server/game/Guilds/Guild.cpp
Lillecarl: I do see your point regarding auto here, QtCreator (without ClangCodeModel because it newer works) could not autocomplete memberItr. I know recent Visual Studio's have no issues whatsoever without auto though.

Shauren: There is another, bigger, problem with `auto` that seems to always be ignored by everyone. Your IDE is not the only place where you read code, github pull requests is another place and you have none of the tools available to determine variable type

Lillecarl: I don't see the problem there though considering you generally don't write code using GH and we have CI to say if members of a class variable exists or not.

src/server/game/Guilds/Guild.cpp
```diff
@@ -2400,9 +2400,12 @@ void Guild::DeleteMember(SQLTransaction& trans, ObjectGuid guid, bool isDisbandi
...
-        delete member;
-    m_members.erase(lowguid);
+    auto memberItr = m_members.find(guid);
```
Rochet2: Hmm, how does this work if m_members is `std::unordered_map<uint32, Member*>` and guid is `ObjectGuid`?
I dont think there is a cast implemented for that. Am I missing something?

Lillecarl: I didn't test it in game yet, so it probably doesn't.

Shauren: it compiled with everything? that would be weird

Lillecarl: Yup.

Lillecarl: I basically just did what Shauren wrote, Sry for that.

Shauren: Edit: i know why it compiles, implicit operator conversion to uint64 which then gets truncated to uint32 (works for player guids, not portable across branches)

PR #24142, opened 2020-02-09
WIP: Remove C-style arrays

Lillecarl:
**Changes proposed:**
-  Remove C-style arrays

**Target branch(es):**
- [x] 3.3.5
- [ ] master

**Issues addressed:**
- Removes type unsaftey

**Tests performed:**
- builds

- [ ] Remove the rest of the about 400 C-style arrays poisioning the codebase

***Notes***
I started a little shy at this, since there's no automatic fix for this i have to change it all automatically and would like feedback before i continue.

### DIFF
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -29,6 +29,7 @@
 #include "PetDefines.h"
 #include "PlayerTaxi.h"
 #include "QuestDef.h"
+#include "TradeData.h"
 #include <memory>
 #include <queue>
 #include <unordered_set>
@@ -1075,7 +1076,7 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         InventoryResult CanTakeMoreSimilarItems(uint32 entry, uint32 count, uint32* itemLimitCategory = nullptr) const { return CanTakeMoreSimilarItems(entry, count, nullptr, nullptr, itemLimitCategory); }
         InventoryResult CanStoreNewItem(uint8 bag, uint8 slot, ItemPosCountVec& dest, uint32 item, uint32 count, uint32* no_space_count = nullptr) const;
         InventoryResult CanStoreItem(uint8 bag, uint8 slot, ItemPosCountVec& dest, Item* pItem, bool swap = false) const;
-        InventoryResult CanStoreItems(Item** items, int count, uint32* itemLimitCategory) const;
+        InventoryResult CanStoreItems(std::array<Item*, TRADE_SLOT_TRADED_COUNT> items, int count, uint32* itemLimitCategory) const;
         InventoryResult CanEquipNewItem(uint8 slot, uint16& dest, uint32 item, bool swap) const;
         InventoryResult CanEquipItem(uint8 slot, uint16& dest, Item* pItem, bool swap, bool not_loading = true) const;
 
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -1076,7 +1076,7 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         InventoryResult CanTakeMoreSimilarItems(uint32 entry, uint32 count, uint32* itemLimitCategory = nullptr) const { return CanTakeMoreSimilarItems(entry, count, nullptr, nullptr, itemLimitCategory); }
         InventoryResult CanStoreNewItem(uint8 bag, uint8 slot, ItemPosCountVec& dest, uint32 item, uint32 count, uint32* no_space_count = nullptr) const;
         InventoryResult CanStoreItem(uint8 bag, uint8 slot, ItemPosCountVec& dest, Item* pItem, bool swap = false) const;
-        InventoryResult CanStoreItems(std::array<Item*, TRADE_SLOT_TRADED_COUNT> items, int count, uint32* itemLimitCategory) const;
+        InventoryResult CanStoreItems(std::array<Item*, TRADE_SLOT_TRADED_COUNT>& items, int count, uint32* itemLimitCategory) const;
         InventoryResult CanEquipNewItem(uint8 slot, uint16& dest, uint32 item, bool swap) const;
         InventoryResult CanEquipItem(uint8 slot, uint16& dest, Item* pItem, bool swap, bool not_loading = true) const;
 
--- a/src/server/game/Handlers/SpellHandler.cpp
+++ b/src/server/game/Handlers/SpellHandler.cpp
@@ -640,7 +640,7 @@ void WorldSession::HandleMirrorImageDataRequest(WorldPacket& recvData)
         data << uint8(player->GetFacialStyle());
         data << uint32(player->GetGuildId());
 
-        static std::array const itemSlots =
+        static std::array<EquipmentSlots, 12>  const itemSlots =
         {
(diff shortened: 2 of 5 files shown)


src/server/game/Handlers/TradeHandler.cpp
```diff
@@ -237,7 +237,7 @@ static void clearAcceptTradeMode(TradeData* myTrade, TradeData* hisTrade)
     hisTrade->SetInAcceptProcess(false);
 }
 
-static void clearAcceptTradeMode(Item* *myItems, Item* *hisItems)
+static void clearAcceptTradeMode(std::array<Item*, TRADE_SLOT_TRADED_COUNT> myItems, std::array<Item*, TRADE_SLOT_TRADED_COUNT> hisItems)
```
illfated: This change seems to cause a warning(error) in AppVeyor for line 357:
```cpp
c:\projects\trinitycore\src\server\game\handlers\tradehandler.cpp(357): error C2220: warning treated as error - no 'object' file generated [C:\projects\trinitycore\build\src\server\game\game.vcxproj]
c:\projects\trinitycore\src\server\game\handlers\tradehandler.cpp(357): warning C4700: uninitialized local variable 'hisItems' used [C:\projects\trinitycore\build\src\server\game\game.vcxproj]
c:\projects\trinitycore\src\server\game\handlers\tradehandler.cpp(357): warning C4700: uninitialized local variable 'myItems' used [C:\projects\trinitycore\build\src\server\game\game.vcxproj]
```
https://github.com/TrinityCore/TrinityCore/pull/24142/files#diff-5521b6495c30ba901d8ed225feba9bf7L357
```cpp
        setAcceptTradeMode(my_trade, his_trade, myItems, hisItems);
```

Lillecarl: Cheers, I'll take a look

src/server/game/Entities/Player/Player.h
```diff
@@ -1075,7 +1076,7 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         InventoryResult CanTakeMoreSimilarItems(uint32 entry, uint32 count, uint32* itemLimitCategory = nullptr) const { return CanTakeMoreSimilarItems(entry, count, nullptr, nullptr, itemLimitCategory); }
         InventoryResult CanStoreNewItem(uint8 bag, uint8 slot, ItemPosCountVec& dest, uint32 item, uint32 count, uint32* no_space_count = nullptr) const;
         InventoryResult CanStoreItem(uint8 bag, uint8 slot, ItemPosCountVec& dest, Item* pItem, bool swap = false) const;
-        InventoryResult CanStoreItems(Item** items, int count, uint32* itemLimitCategory) const;
+        InventoryResult CanStoreItems(std::array<Item*, TRADE_SLOT_TRADED_COUNT> items, int count, uint32* itemLimitCategory) const;
```
Shauren: Copying arrays is straight up no good

Shauren (changes requested):

src/server/game/Handlers/SpellHandler.cpp
```diff
@@ -640,7 +640,7 @@ void WorldSession::HandleMirrorImageDataRequest(WorldPacket& recvData)
         data << uint8(player->GetFacialStyle());
         data << uint32(player->GetGuildId());
 
-        static EquipmentSlots const itemSlots[] =
+        static std::array const itemSlots =
```
Shauren: This looks like `auto` to me (array of what?), since we only just enabled c++17, not sure how to proceed

src/server/game/Server/WorldSession.h
```diff
@@ -29,6 +29,7 @@
 #include "Packet.h"
 #include "QueryCallbackProcessor.h"
 #include "SharedDefines.h"
+#include "TradeData.h"
```
Shauren: Here is why I prefer to NOT use std::array, you aren't bloating includes

Lillecarl: Ah yes i mentioned this in the commit message as the following "I know people will not like the extra includes here, but to be safe we either need to move the enum or use another container to be safe."
We could always use a vector in this case and reserve the size, the C style stuff is really unsafe if someone comes around to touching it.

Shauren: Nobody is ever going to touch trade data size because its not a variable you can tweak, it must match client expectations (this can never change in 3.3.5 and in master branch I will be the person to do it, if it ever does (unlikely))

src/server/game/Entities/Player/Player.h
Lillecarl: Should probably be a reference indeed.

src/server/game/Handlers/SpellHandler.cpp
Lillecarl: It's technically auto indeed. If we don't specify type nor size the compiler will do what it was designed to do and deduct the type for us along with the size.
https://www.learncpp.com/cpp-tutorial/6-15-an-introduction-to-stdarray/ That's the rules apparently, i could specify the size there as well.

src/server/game/Server/WorldSession.h
Lillecarl: Good points. I just want to keep the codebase up2date now that I'm doing TC for a hobby, which is where TC shines in comparision with the "competitors" (Way more up2date) :)

If trade size is a value that'll never change we could move it to shareddefines which is pretty much included everywhere already, and get rid of those new includes.

jackpoz:
were these reported by clang tidy ? which parameters does it need for this particular warning ?

Lillecarl:
https://clang.llvm.org/extra/clang-tidy/checks/modernize-avoid-c-arrays.html It doesn't have a fixit included. But just a warning to fix.

jackpoz:
we might have to agree first if this is the direction we want to take or not before merging this PR and continuing with the other 400 arrays

Lillecarl:
Yes definitely. To be honest as it looks I'd say we rather convert these usecases where extra header includes are required into vectors and handle them dynamically, or move the std::array definitions into some generic header, but that's ew.

jackpoz:
@Shauren does this look ok, even with the additional include ?

Lillecarl:
Considering it's not really a hot code path a change of container to vector would probably be the better choice than an array.

Naios:
Propably the right way to do it in modern C++ would be to change the C array to `std::array` and passing it to other functions as `std::span` https://en.cppreference.com/w/cpp/container/span. The overhead is minimal (2 pointers instead of one) and also brings most benifits as `std::array` does like bounds checking and O(1) size retrieval without specifying its size explicitly in its type. Through that it also won't matter whether you are accessing a c-like array, a `std::array` or `std::vector` in the codebase and its representation can stay untouched while its container type may change.

src/server/game/Server/WorldSession.h
jackpoz: Should we go with the "move the enum to shareddefines" then, include 1 less header and merge the PR ?

illfated: Sounds like a good idea to me.

Shauren: No, I would leave the code as-is (until std::span)

jackpoz:
let's wait for std::span.

Thanks anyway for the PR :)

PR #24151, opened 2020-02-13
Implement previously implicit ctor explicitly

Lillecarl:
Declare them default explicitly as current behaviour was deprecated in C++17

**Changes proposed:**

-  Declare copy ctor default in Position and WorldLocation classes.

**Target branch(es):** 

- [x] 3.3.5
- [ ] master

**Issues addressed:**
Warnings in Position.h

**Tests performed:**
Builds


Shauren:
Taking a closer look at Position and WorldLocation classes... do they even need user defined copy constructors and assignments? (fix would change from adding assignment operator to removing copy constructor)

Lillecarl:
Well they're used right now, didn't wanna change behaviour. We could investigate this further.

Shauren:
I was hinting at default compiler generated ones doing exactly the same as our "custom" implementation does

Shauren (approved):

PR #24159, opened 2020-02-15
Remove dependency on undefined behaviour

Lillecarl:
**Changes proposed:**
-  Stop relying on undefined behaviour to convert between float and int raw bits.

- [x] 3.3.5

**Issues addressed:**
Use legal C++ rather than not.

Builds, ran vmap4assembler w/o issues

### DIFF
--- a/src/common/Collision/BoundingIntervalHierarchy.h
+++ b/src/common/Collision/BoundingIntervalHierarchy.h
@@ -29,29 +29,26 @@
 #include <algorithm>
 #include <limits>
 #include <cmath>
+#include "string.h"
 
 #define MAX_STACK_SIZE 64
 
+// https://stackoverflow.com/a/4328396
+
 static inline uint32 floatToRawIntBits(float f)
 {
-    union
-    {
-        uint32 ival;
-        float fval;
-    } temp;
-    temp.fval=f;
-    return temp.ival;
+    static_assert(sizeof(float) == sizeof(uint32), "Size of uint32 and float must be equal for this to work");
+    uint32 ret;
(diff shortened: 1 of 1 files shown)


src/common/Collision/BoundingIntervalHierarchy.h
```diff
@@ -29,29 +29,26 @@
...
-    temp.fval=f;
-    return temp.ival;
+    static_assert(sizeof(float) == sizeof(uint32), "Size of uint32 and float must be equal for this to work");
+    uint32 ret;
+    memcpy(&ret, &f, sizeof(float));
+    return ret;
 }
 
-static inline float intBitsToFloat(uint32 i)
+static inline uint32 intBitsToFloat(uint32 i)
```
jackpoz: is this supposed to be uint32 ?

Lillecarl: absolutely not, how tf can i fail like this. /shame

jackpoz:
was this spotted by clang-tidy too ? could you paste which parameters you passed to spot this ?

Lillecarl:
@jackpoz No this one wasn't it was found by me.

jackpoz:
Please keep an eye on the issues list if anyone report strange behavior. We will be able to use https://en.cppreference.com/w/cpp/numeric/bit_cast when we will support C++20

Lillecarl:
@jackpoz Yes, could you add a C++20 tag to this issue so the information doesn't get lost (As i think it'll be about 2 years before we're C++20 ready).

Thanks :)

PR #24165, opened 2020-02-16
Modernize codebase with Clang-Tidy range based loops, part 2

Lillecarl:
**Changes proposed:**
-  Range based loops everywhere

**Target branch(es):**
- [x] 3.3.5
- [ ] master

**Issues addressed:** 
Keeping TrinityCore up2date with the C++ standards

**Tests performed:**
Builds locally with clang

- [ ] Finish the rest

### DIFF
--- a/src/server/game/AI/CoreAI/PetAI.cpp
+++ b/src/server/game/AI/CoreAI/PetAI.cpp
@@ -169,9 +169,9 @@ void PetAI::UpdateAI(uint32 diff)
                 // No enemy, check friendly
                 if (!spellUsed)
                 {
-                    for (GuidSet::const_iterator tar = _allySet.begin(); tar != _allySet.end(); ++tar)
+                    for (ObjectGuid target : _allySet)
                     {
-                        Unit* ally = ObjectAccessor::GetUnit(*me, *tar);
+                        Unit* ally = ObjectAccessor::GetUnit(*me, target);
 
                         //only buff targets that are in combat, unless the spell can only be cast while out of combat
                         if (!ally)
@@ -218,8 +218,8 @@ void PetAI::UpdateAI(uint32 diff)
         }
 
         // deleted cached Spell objects
-        for (TargetSpellList::const_iterator itr = targetSpellStore.begin(); itr != targetSpellStore.end(); ++itr)
-            delete itr->second;
+        for (std::pair<Unit*, Spell*> const& unitspellpair : targetSpellStore)
+            delete unitspellpair.second;
     }
 
     // Update speed as needed to prevent dropping too far behind and despawning
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -169,12 +169,12 @@ Creature* SmartScript::FindCreatureNear(WorldObject* searchObject, ObjectGuid::L
 void SmartScript::OnReset()
 {
     ResetBaseObject();
-    for (SmartAIEventList::iterator i = mEvents.begin(); i != mEvents.end(); ++i)
+    for (SmartScriptHolder& event : mEvents)
     {
-        if (!((*i).event.event_flags & SMART_EVENT_FLAG_DONT_RESET))
+        if (!(event.event.event_flags & SMART_EVENT_FLAG_DONT_RESET))
         {
-            InitTimer((*i));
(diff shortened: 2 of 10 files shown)


jackpoz:
wait for the other PRs to be merged before opening more :)

Lillecarl:
Okies, i'll pause

Fabi:
> wait for the other PRs to be merged before opening more :)

Merge faster then

MaxtorCoder:
> Merge faster then
Lmaoooooooo

jackpoz:
2 range-loop PRs are enough, no need to open 100 and review them all at the same time, maybe with the same issues in each PR.

src/server/game/AI/CoreAI/PetAI.cpp
```diff
@@ -169,9 +169,9 @@ void PetAI::UpdateAI(uint32 diff)
                 // No enemy, check friendly
                 if (!spellUsed)
                 {
-                    for (GuidSet::const_iterator tar = _allySet.begin(); tar != _allySet.end(); ++tar)
+                    for (ObjectGuid target : _allySet)
```
jackpoz: image
I wonder what's the "standard" way to have a ObjectGuid iterator

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -2067,9 +2067,8 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                             if (!path || path->nodes.empty())
                                 continue;
 
-                            for (auto itr = path->nodes.begin(); itr != path->nodes.end(); ++itr)
+                            for (WaypointNode waypoint : path->nodes)
```
jackpoz: should this have & too ?

src/server/game/AI/CoreAI/PetAI.cpp
Lillecarl: Well it's not really an iterator but yeah, I recon objectguid const& should be the default

src/server/game/AI/SmartScripts/SmartScript.cpp
Lillecarl: Probably, I'll go though it all again. Just need to set up my devenv again 😊

src/server/game/Achievements/AchievementMgr.cpp
```diff
@@ -1692,15 +1683,15 @@ bool AchievementMgr::CanUpdateCriteria(AchievementCriteriaEntry const* criteria,
 
 bool AchievementMgr::ConditionsSatisfied(AchievementCriteriaEntry const* criteria) const
 {
-    for (uint32 i = 0; i < MAX_CRITERIA_REQUIREMENTS; ++i)
+    for (auto /* unnamed struct */ AdditionalRequirement : criteria->AdditionalRequirements)
```
jackpoz: "auto /* unnamed struct */" some leftovers ?

Lillecarl: Nope, see where it's declared, there's no named type for it. I put the comment here to please Shauren.

src/server/game/AI/CoreAI/PetAI.cpp
jackpoz: >Shauren: its definitely better to just use value for guid, not reference. at least in 3.3.5
ModoX: On master reference be better since we're using 128 bit guids over there
Shauren: @modox, yes and no. i half expect the compiler to use xmm registers like it does in wow client. in which case a copy would be better
ModoX: Good point

so copy it is :)

src/server/game/Achievements/AchievementMgr.cpp
jackpoz: I think it's more readable without the comment in the code (some comments fit better in GitHub as they are just part of the PR review)

src/server/game/AI/CoreAI/PetAI.cpp
Lillecarl: Awesome, not that this holds only when not modifying the guid (and we wanna save it) 😁

jackpoz (changes requested):
Please fix the comments above and fix the build warnings/errors that the CI checks reported

jackpoz:
Thanks for the PR :)

PR #24178, opened 2020-02-19
Core/Spell: Proper SPELL_EFFECT_PULL_TOWARDS_DEST implementation for players

### LINKED ISSUE #23203: Proper SPELL_EFFECT_PULL_TOWARDS_DEST implementation for players
**Description:**

`SPELL_EFFECT_PULL_TOWARDS_DEST` is implemented as a parabolic spline jump, but it should in fact be implemented as a knockback for players, which can be verified by examining sniffs.

This prevents at least the following spells from working correctly (those are the only 3 use cases of the effect in 3.3.5):
https://www.wowhead.com/spell=46230 - Mu'ru in Sunwell Plateau, apparently
https://www.wowhead.com/spell=47764 - Grand Magus Telestra in The Nexus
https://www.wowhead.com/spell=68645 - Rocket Pack on ICC Gunships encounter

**Implementation:**

Revert 9a1282aac6befe6b3e26deb39d066fdc04452f4a

Spell.h
```diff
         void EffectPullTowards(SpellEffIndex effIndex);
+        void EffectPullTowardsDest(SpellEffIndex effIndex);
```
SpellEffects.cpp
```diff
-    &Spell::EffectPullTowards,                              //145 SPELL_EFFECT_PULL_TOWARDS_DEST                      Black Hole Effect
+    &Spell::EffectPullTowardsDest,                          //145 SPELL_EFFECT_PULL_TOWARDS_DEST
```
Spell::EffectPullTowards
```diff
-    float speedZ = (float)(m_spellInfo->Effects[effIndex].CalcValue() / 10);
-    float speedXY = (float)(m_spellInfo->Effects[effIndex].MiscValue/10);
+    float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
+    float speedZ = damage / 10.0f;
     Position pos;
-    if (m_spellInfo->Effects[effIndex].Effect == SPELL_EFFECT_PULL_TOWARDS_DEST)
-    {
-        if (m_targets.HasDst())
-            pos.Relocate(*destTarget);
-        else
-            return;
-    }
-    else //if (m_spellInfo->Effects[i].Effect == SPELL_EFFECT_PULL_TOWARDS)
-    {
-        pos.Relocate(m_caster);
-    }
+    pos.Relocate(m_caster);
```
```diff
+void Spell::EffectPullTowardsDest(SpellEffIndex effIndex)
+{
+    if (effectHandleMode != SPELL_EFFECT_HANDLE_HIT_TARGET)
+        return;
+
+    if (!unitTarget)
+        return;
+
+    if (!m_targets.HasDst())
+    {
+        LOG_ERROR_in_whateverWayYou_do_itTheseDays("Spell %u with SPELL_EFFECT_PULL_TOWARDS_DEST has no dest target", m_spellInfo->Id);
+        return;
+    }
+
+    Position const* pos = m_targets.GetDstPos();
+    float distXY = unitTarget->GetExactDist(pos); // This is a blizzlike error: this should be 2D distance according to projectile motion formulas, but Blizzard erroneously used 3D distance
+    float distZ = pos->GetPositionZ() - unitTarget->GetPositionZ();
+
+    float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
+    float speedZ = (2 * speedXY * speedXY * distZ + Movement::gravity * distXY * distXY) / (2 * speedXY * distXY);
+
+    unitTarget->JumpTo(speedXY, speedZ, true, pos);
+}
+
```
(`Unit::JumpTo` is actually a "knockback" for players, in case you're confused by my statements. It's only a "jump" for creatures)

This will result in knockback data identical to retail servers for players - players will get knocked in such a way that they will **pass through** the dest target. The caveat, however, is that for creatures this isn't enough. Perhaps the parabolic jump spline is calculated incorrectly for creature knockbacks in `Unit::JumpTo`, or perhaps this effect requires special handling, but these changes don't fix pulls for creatures properly. You would need to somehow calculate a ballistic trajectory that goes **through** the desired point (dest target) and then figure out when that trajectory will intersect with the floor (which isn't a flat plane) AND not let the creature escape the navmesh with such a jump. That's the ideal scenario, but whether or not blizzard does it that way - I don't know, I didn't research creatures, I had spent way too many hours just trying to figure out the formula for players. Could be that they just project the trajectory onto the same floor plane (z) as the creature, find where they intersect, and then nudge the point along the navmesh from creature's position towards the intersection point as far as possible without leaving the navmesh.

This will also require changes to `spell_igb_rocket_pack`: in `HandlePeriodic` you would need to check if the player is missing the "falling" movement flag (after first waiting for it to be applied, as there is no guarantee the player won't send more movement updates between the spell cast and knockback acknowledgement) instead of checking if the spline is finished, as there won't be any splines used for the pull anymore.

As for Magus Telestra's case, be advised that the effect won't look nice if you're not updating player's serverside position every time a `CMSG_MOVE_KNOCK_BACK_ACK` packet is received. Pull is triggered every 200ms there, so if the player's position isn't getting updated often (due to `CMSG_MOVE_KNOCK_BACK_ACK` being sent constantly, the client won't be very eager to also send movement heartbeat updates often), the effect will look as if the player is constantly being pulled from the same place. And the pull destination also needs to be shifted up 3 yards in a spell script, but that's beyond the scope of this issue.

**Branch(es):**

both

**TC rev. hash/commit:** 

32e1de39a26628dcb64bc21ad415afb2ad938925 (irrelevant, but hi, bot)

Faq:
Closes #23203 and also reverts 9a1282a
thnx to xvwyh

**Changes proposed:**
SPELL_EFFECT_PULL_TOWARDS_DEST is implemented as a parabolic spline jump, but it should in fact be implemented as a knockback for players, which can be verified by examining sniffs.

This prevents at least the following spells from working correctly (those are the only 3 use cases of the effect in 3.3.5):
https://www.wowhead.com/spell=46230 - Mu'ru in Sunwell Plateau, apparently
https://www.wowhead.com/spell=47764 - Grand Magus Telestra in The Nexus
https://www.wowhead.com/spell=68645 - Rocket Pack on ICC Gunships encounter

- [x] 3.3.5
- [ ] master

**Issues addressed:** 
Closes #23203

Compiles and tests performed by Rushor

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12305,11 +12305,11 @@ uint32 Unit::GetModelForForm(ShapeshiftForm form, uint32 spellId) const
     return modelid;
 }
 
-void Unit::JumpTo(float speedXY, float speedZ, bool forward, Position const* dest)
+void Unit::JumpTo(float speedXY, float speedZ, bool forward, Optional<Position> dest)
 {
     float angle = forward ? 0 : float(M_PI);
     if (dest)
-        angle += GetRelativeAngle(dest);
+        angle += GetRelativeAngle(*dest);
 
     if (GetTypeId() == TYPEID_UNIT)
         GetMotionMaster()->MoveJumpTo(angle, speedXY, speedZ);
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12305,9 +12305,12 @@ uint32 Unit::GetModelForForm(ShapeshiftForm form, uint32 spellId) const
     return modelid;
 }
 
-void Unit::JumpTo(float speedXY, float speedZ, bool forward)
+void Unit::JumpTo(float speedXY, float speedZ, bool forward, Position const* dest)
 {
     float angle = forward ? 0 : float(M_PI);
+    if (dest)
+        angle += GetRelativeAngle(dest);
+
     if (GetTypeId() == TYPEID_UNIT)
         GetMotionMaster()->MoveJumpTo(angle, speedXY, speedZ);
     else
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -207,7 +207,7 @@ SpellEffectHandlerFn SpellEffectHandlers[TOTAL_SPELL_EFFECTS] =
     &Spell::EffectTriggerSpell,                             //142 SPELL_EFFECT_TRIGGER_SPELL_WITH_VALUE
     &Spell::EffectUnused,                                   //143 SPELL_EFFECT_APPLY_AREA_AURA_OWNER
     &Spell::EffectKnockBack,                                //144 SPELL_EFFECT_KNOCK_BACK_DEST
-    &Spell::EffectPullTowards,                              //145 SPELL_EFFECT_PULL_TOWARDS_DEST                      Black Hole Effect
+    &Spell::EffectPullTowardsDest,                          //145 SPELL_EFFECT_PULL_TOWARDS_DEST        Black Hole Effect
     &Spell::EffectActivateRune,                             //146 SPELL_EFFECT_ACTIVATE_RUNE
     &Spell::EffectQuestFail,                                //147 SPELL_EFFECT_QUEST_FAIL               quest fail
     &Spell::EffectTriggerMissileSpell,                      //148 SPELL_EFFECT_TRIGGER_MISSILE_SPELL_WITH_VALUE
@@ -4392,23 +4392,37 @@ void Spell::EffectPullTowards(SpellEffIndex effIndex)
     if (!unitTarget)
         return;
 
+    float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
+    float speedZ = damage / 10.0f;
     Position pos;
-    if (m_spellInfo->Effects[effIndex].Effect == SPELL_EFFECT_PULL_TOWARDS_DEST)
-    {
-        if (m_targets.HasDst())
-            pos.Relocate(*destTarget);
-        else
-            return;
-    }
-    else //if (m_spellInfo->Effects[i].Effect == SPELL_EFFECT_PULL_TOWARDS)
-    {
-        pos.Relocate(m_caster);
-    }
-
-    float speedXY = float(m_spellInfo->Effects[effIndex].MiscValue) * 0.1f;
-    float speedZ = unitTarget->GetDistance(pos) / speedXY * 0.5f * Movement::gravity;
+    pos.Relocate(m_caster);
 
     unitTarget->GetMotionMaster()->MoveJump(pos, speedXY, speedZ);
+}
+
+void Spell::EffectPullTowardsDest(SpellEffIndex effIndex)
+{
+    if (effectHandleMode != SPELL_EFFECT_HANDLE_HIT_TARGET)
+        return;
+
+    if (!unitTarget)
+        return;
+
+    if (!m_targets.HasDst())
+    {
+        TC_LOG_ERROR("spells", "Spell %u with SPELL_EFFECT_PULL_TOWARDS_DEST has no dest target", m_spellInfo->Id);
+        return;
+    }
+
+    Position const* pos = m_targets.GetDstPos();
+    // This is a blizzlike mistake: this should be 2D distance according to projectile motion formulas, but Blizzard erroneously used 3D distance
+    float distXY = unitTarget->GetExactDist(pos);
+    float distZ = pos->GetPositionZ() - unitTarget->GetPositionZ();
+
+    float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
+    float speedZ = (2 * speedXY * speedXY * distZ + Movement::gravity * distXY * distXY) / (2 * speedXY * distXY);
+
+    unitTarget->JumpTo(speedXY, speedZ, true, pos);
 }
 
 void Spell::EffectDispelMechanic(SpellEffIndex effIndex)
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -4422,7 +4422,7 @@ void Spell::EffectPullTowardsDest(SpellEffIndex effIndex)
     float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
     float speedZ = (2 * speedXY * speedXY * distZ + Movement::gravity * distXY * distXY) / (2 * speedXY * distXY);
 
-    unitTarget->JumpTo(speedXY, speedZ, true, pos);
+    unitTarget->JumpTo(speedXY, speedZ, true, *pos);
 }
 
 void Spell::EffectDispelMechanic(SpellEffIndex effIndex)
--- a/src/server/game/Entities/Unit/Unit.h
+++ b/src/server/game/Entities/Unit/Unit.h
@@ -1108,7 +1108,7 @@ class TC_GAME_API Unit : public WorldObject
         void UpdateHeight(float newZ);
 
         void KnockbackFrom(float x, float y, float speedXY, float speedZ);
-        void JumpTo(float speedXY, float speedZ, bool forward = true, Position const* dest = nullptr);
+        void JumpTo(float speedXY, float speedZ, bool forward = true, Optional<Position> dest = {});
         void JumpTo(WorldObject* obj, float speedZ, bool withOrientation = false);
 
         void MonsterMoveWithSpeed(float x, float y, float z, float speed, bool generatePath = false, bool forceDestination = false);
--- a/src/server/game/Entities/Unit/Unit.h
+++ b/src/server/game/Entities/Unit/Unit.h
@@ -1108,7 +1108,7 @@ class TC_GAME_API Unit : public WorldObject
         void UpdateHeight(float newZ);
 
         void KnockbackFrom(float x, float y, float speedXY, float speedZ);
-        void JumpTo(float speedXY, float speedZ, bool forward = true);
+        void JumpTo(float speedXY, float speedZ, bool forward = true, Position const* dest = nullptr);
         void JumpTo(WorldObject* obj, float speedZ, bool withOrientation = false);
 
         void MonsterMoveWithSpeed(float x, float y, float z, float speed, bool generatePath = false, bool forceDestination = false);
--- a/src/server/game/Spells/Spell.h
+++ b/src/server/game/Spells/Spell.h
@@ -342,6 +342,7 @@ class TC_GAME_API Spell
         void EffectSendTaxi(SpellEffIndex effIndex);
         void EffectKnockBack(SpellEffIndex effIndex);
         void EffectPullTowards(SpellEffIndex effIndex);
+        void EffectPullTowardsDest(SpellEffIndex effIndex);
         void EffectDispelMechanic(SpellEffIndex effIndex);
         void EffectResurrectPet(SpellEffIndex effIndex);
         void EffectDestroyAllTotems(SpellEffIndex effIndex);


Killyana:
Not working well for me: https://imgur.com/a/B7nvkrX

Mesielecat:
> Not working well for me: https://imgur.com/a/B7nvkrX
> Tested with https://www.wowhead.com/spell=68645/rocket-pack

Killyana do you try in other place out of the airship?

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -186,7 +186,7 @@ SpellEffectHandlerFn SpellEffectHandlers[TOTAL_SPELL_EFFECTS] =
     &Spell::EffectWeaponDmg,                                //121 SPELL_EFFECT_NORMALIZED_WEAPON_DMG
     &Spell::EffectUnused,                                   //122 SPELL_EFFECT_122                      unused
     &Spell::EffectSendTaxi,                                 //123 SPELL_EFFECT_SEND_TAXI                taxi/flight related (misc value is taxi path id)
-    &Spell::EffectPullTowards,                              //124 SPELL_EFFECT_PULL_TOWARDS
+    &Spell::EffectPullTowardsDest,                          //145 SPELL_EFFECT_PULL_TOWARDS_DEST
```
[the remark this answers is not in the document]
Faq: ups

Killyana:
Yes, even if I'm not on the transport the result is the same

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -4392,23 +4392,37 @@ void Spell::EffectPullTowards(SpellEffIndex effIndex)
...
 
-    float speedXY = float(m_spellInfo->Effects[effIndex].MiscValue) * 0.1f;
-    float speedZ = unitTarget->GetDistance(pos) / speedXY * 0.5f * Movement::gravity;
+    Position const* pos = m_targets.GetDstPos();
+    // This is a blizzlike mistake: this should be 2D distance according to projectile motion formulas, but Blizzard erroneously used 3D distance
+    float distXY = unitTarget->GetExactDist(pos);
+    float distZ = pos->GetPositionZ() - unitTarget->GetPositionZ();
 
-    unitTarget->GetMotionMaster()->MoveJump(pos, speedXY, speedZ);
+    float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
```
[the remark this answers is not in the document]
xvwyh: It's functionally identical to the previous implementation, so at the very least nothing will change for the worse:
`float speedXY = float(m_spellInfo->Effects[effIndex].MiscValue) * 0.1f;`

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -4392,23 +4392,37 @@ void Spell::EffectPullTowards(SpellEffIndex effIndex)
     if (!unitTarget)
         return;
 
+    float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
```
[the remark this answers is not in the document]
xvwyh: It's functionally identical to the previous implementation, so at the very least nothing will change for the worse:
`float speedXY = float(m_spellInfo->Effects[effIndex].MiscValue) * 0.1f;`

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -4392,23 +4392,37 @@ void Spell::EffectPullTowards(SpellEffIndex effIndex)
...
+    float speedZ = damage / 10.0f;
```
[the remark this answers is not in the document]
xvwyh: Yes, in which case speedZ will be 0, I don't see a problem there.

xvwyh:
@Killyana, for starters - try removing the script "spell_igb_rocket_pack" from the spell. I can't imagine how its presence could be interfering with something movement-related, but who knows. It needs to be rewritten anyway, as splines will no longer be used for that jump after this patch. The script will need to monitor player's movement flags to keep track of when the player is airborne and then no longer airborne to determine when to trigger the "landing" effect.

Killyana:
I removed `ScriptName`='spell_igb_rocket_pack' from the DB, but still the same result.

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -4392,23 +4392,37 @@ void Spell::EffectPullTowards(SpellEffIndex effIndex)
...
+    Position const* pos = m_targets.GetDstPos();
+    // This is a blizzlike mistake: this should be 2D distance according to projectile motion formulas, but Blizzard erroneously used 3D distance
+    float distXY = unitTarget->GetExactDist(pos);
+    float distZ = pos->GetPositionZ() - unitTarget->GetPositionZ();
 
-    unitTarget->GetMotionMaster()->MoveJump(pos, speedXY, speedZ);
+    float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
+    float speedZ = (2 * speedXY * speedXY * distZ + Movement::gravity * distXY * distXY) / (2 * speedXY * distXY);
+
+    unitTarget->JumpTo(speedXY, speedZ, pos);
```
xvwyh: Hm, I see, TC doesn't have the parameter for this method that serves to specify the direction of the knockback. Here `pos` is being erroneously passed to the `bool forward` parameter.

Here it is:
Unit.h
```diff
-        void JumpTo(float speedXY, float speedZ, bool forward = true);
+        void JumpTo(float speedXY, float speedZ, bool forward = true, Position* dest = nullptr);
```

Unit.cpp
```diff
-void Unit::JumpTo(float speedXY, float speedZ, bool forward)
+void Unit::JumpTo(float speedXY, float speedZ, bool forward, Position* dest)
 {
     float angle = forward ? 0 : M_PI;
+    if (dest)
+        angle += GetRelativeAngle(dest);
+
     if (GetTypeId() == TYPEID_UNIT)
         GetMotionMaster()->MoveJumpTo(angle, speedXY, speedZ);
```

Killyana: No change

xvwyh: Not surprising.
I don't have a slightest clue why it's not working for you. The code is very simple, I can't imagine what can it depend on that I'm not seeing.
I'll wait for someone else to test it and report their results.

Killyana: My bad I forget to apply the last commit of this PR, So this is the locale diff: https://pastebin.com/qrAEFR5W
And this is the result: https://imgur.com/i1M7cfP

xvwyh: Yes, NOW you're missing this piece of code the I posted above. Obviously you must also change the call to `unitTarget->JumpTo(speedXY, speedZ, true, pos);`.

xvwyh:
Went searching in sniffs for what could be the default value for when the spell has MiscValue=0. Found one case with spell 31703, SpeedXY was 30 in this case.

You might also notice that that spells's effect is `SPELL_EFFECT_PULL_TOWARDS`, and as it turns out - yes, it too must be implemented as a clientside knockback for players, just like `SPELL_EFFECT_PULL_TOWARDS_DEST`, instead of spline as it currently is. Luckily, it seems to work just like `SPELL_EFFECT_PULL_TOWARDS_DEST`, so it can be implemented in the same way:

```cpp
void Spell::EffectPullTowards(SpellEffIndex effIndex)
{
    if (effectHandleMode != SPELL_EFFECT_HANDLE_HIT_TARGET)
        return;

    if (!unitTarget)
        return;

    Position pos;
    pos.Relocate(m_caster);

    float distXY = unitTarget->GetExactDist(&pos); // This is a blizzlike error: this should be 2D distance, but Blizzard erroneously used 3D distance
    float distZ = pos.GetPositionZ() - unitTarget->GetPositionZ();
    float speedXY = m_spellInfo->Effects[effIndex].MiscValue ? m_spellInfo->Effects[effIndex].MiscValue / 10.0f : 30.0f;
    float speedZ = (2 * speedXY * speedXY * distZ + Movement::gravity * distXY * distXY) / (2 * speedXY * distXY);
    unitTarget->JumpTo(speedXY, speedZ, true, &pos);
}

void Spell::EffectPullTowardsDest(SpellEffIndex effIndex)
{
    if (effectHandleMode != SPELL_EFFECT_HANDLE_HIT_TARGET)
        return;

    if (!unitTarget)
        return;

    if (!m_targets.HasDst())
    {
        sLog->outError("Spell %u with SPELL_EFFECT_PULL_TOWARDS_DEST has no dest target", m_spellInfo->Id);
        return;
    }

    Position const* pos = m_targets.GetDstPos();
    float distXY = unitTarget->GetExactDist(pos); // This is a blizzlike error: this should be 2D distance, but Blizzard erroneously used 3D distance
    float distZ = pos->GetPositionZ() - unitTarget->GetPositionZ();
    float speedXY = m_spellInfo->Effects[effIndex].MiscValue ? m_spellInfo->Effects[effIndex].MiscValue / 10.0f : 30.0f;
    float speedZ = (2 * speedXY * speedXY * distZ + Movement::gravity * distXY * distXY) / (2 * speedXY * distXY);
    unitTarget->JumpTo(speedXY, speedZ, true, pos);
}
```

Yes, this removes the usage of spell points (aka `damage`), I don't know what do they affect, if anything. I don't have any sniffs with spells that have it not be 0.

Additionally I believe that `SPELL_EFFECT_PULL_TOWARDS` must be pulling units not to caster's position, but next to them, e.g. to the edge of their hitbox. Otherwise, what is the point in them being separate effects? However I cannot prove this yet, so it's up to you if you want to change it or not. In case you do: replace `pos.Relocate(m_caster);` with `m_caster->GetFirstCollisionPosition(pos, m_caster->GetCombatReach(), m_caster->GetRelativeAngle(unitTarget));` or something along those lines.

Just bear in mind that `SPELL_EFFECT_PULL_TOWARDS` is used in many more cases than `SPELL_EFFECT_PULL_TOWARDS_DEST`, and I can't vouch for all of them being correct after these changes, because I literally just discovered this.

src/server/game/Spells/SpellEffects.cpp
Killyana: I get an error: c:\trinitysource355\src\server\game\spells\spelleffects.cpp(4425): error C2661: 'Unit::JumpTo' : aucune fonction surchargée ne nécessite 4 arguments (cannot take 4 arguments)

xvwyh: Then you must be wrong in saying
> So this is the locale diff: https://pastebin.com/qrAEFR5W

Because if you had changed Unit::JumpTo method as shown, you wouldn't be getting that error...

Killyana: Yes I revert it to see if it will change something or not.
After reapplying everything I get this error:
``` 
Erreur	C2664	'void Unit::JumpTo(WorldObject *,float,bool)' : impossible de convertir l'argument 4 de 'const Position *' en 'Position *'	game	c:\trinitysource355\src\server\game\spells\spelleffects.cpp	4425

Comment (Killyana) [reply src/server/game/Spells/SpellEffects.cpp]: I revert it to see if it changes something.
After applying everything I get this error:
``` 
Erreur	C2664	'void Unit::JumpTo(WorldObject *,float,bool)' : impossible de convertir l'argument 4 de 'const Position *' en 'Position *'	game	c:\trinitysource355\src\server\game\spells\spelleffects.cpp	4425

xvwyh: Change the parameter from `Position* dest` to `Position const* dest`. I'll edit the comment.

Killyana:
Does sniffs of this spells 30010 63025 64233 74799 will help to understand the use of basepoint on SPELL_EFFECT_PULL_TOWARDS?

Killyana:
Does the orientation of the player is supposed to be changed or not https://imgur.com/b4nvB0j ?

xvwyh:
> Does sniffs of this spells 30010 63025 64233 74799 will help to understand the use of basepoint on SPELL_EFFECT_PULL_TOWARDS?

Hopefully. It all depends on the circumstances. When the player and/or the target is moving - it makes it much harder to understand which positions were used by server for calculations.

> Does the orientation of the player is supposed to be changed or not https://imgur.com/b4nvB0j ?

Since it works like a knockback - the orientation will not change.

Killyana:
Here sniffs of Spell ID: 63025 https://pastebin.com/HMLTimkq

xvwyh:
> It's probably the base point=140

No, it's the MiscValue. MiscValue controls horizontal speed (which is what's called Fall Speed in the sniff because naming is hard), this is 100% correct.

I see that points are almost always either 0, or the same as MiscValue, which gives me reason to speculate that some designers didn't know which field to put their value in, so they tried both.

From your sniff, caster was at 825.9041 -49.36711 409.8036, target was at 840.3483 -35.66744 409.8036. The distance is therefore 19.90768. SpeedXY is obviously 14, clearly visible in the knockback packet. If we were to plug it into my formula...
`(2 * speedXY * speedXY * distZ + Movement::gravity * distXY * distXY) / (2 * speedXY * distXY)`
`(2 * 14 * 14 * 0 + 19.2911 * 19.90768 * 19.90768) / (2 * 14 * 19.90768)`
SpeedZ would end up being 13.71575, which isn't equal to the vertical speed from the knockback packet (12.58898). Two possible explanations:
1. BasePoints somehow affect something. Couldn't yet find what and how.
2. Pull wasn't towards the caster's position, but to a contact point next to the caster.

"2." seem more likely, for the reason I explained above: by pulling someone "to you" you expect that someone to end up being in front of you, not directly inside you. For that we have `SPELL_EFFECT_PULL_TOWARDS_DEST`. Assuming that, using some math we can calculate what the pull distance would be with the values from the sniff, and it ends up being 18.27223. 18.27223 is smaller than 19.90768 by 1.63545 yards. This doesn't look like a player's combat reach, which should be 1.5. So I don't know... Maybe the players were enlarged by something? Would be nice to see PPPPPPPP's latest UNIT_FIELD_BOUNDINGRADIUS and UNIT_FIELD_COMBATREACH up until that point. Otherwise I have no idea where that number came from.

> And this is the one from spell ID=74799 (base point=150)

This is, alas, useless without knowing the caster's position at that time.

Killyana:
``` 
At the biguinning of the raid:
[0] UNIT_FIELD_BOUNDINGRADIUS: 0.347
[0] UNIT_FIELD_COMBATREACH: 1.5

Just after the cast of Spell ID: 63025 
[5] GUID: Full: 0x60000000 Type: Player Low: ppppppp Name: ppppppp
[5] OBJECT_FIELD_SCALE_X: 1.3
[5] UNIT_FIELD_BASEATTACKTIME: 1030/1.443337E-42
[5] UNIT_FIELD_UNK63: 1288/1.804872E-42
[5] UNIT_FIELD_BOUNDINGRADIUS: 0.4511
[5] UNIT_MOD_CAST_SPEED: 1061105662/0.7468261

**Comment by Faq** [comment]
hmm

**Comment by xvwyh** [comment]
I'm sorry, @Killyana, I don't know.

If base points has any effect on the skill, I still can't figure it out. I tend to believe that it doesn't, and it only pops up in modern spells because devs just copy-paste them, with previous devs' mistakes intact.

No matter how I try to calculate the distance the players in those sniffs are supposed to travel - it never matches up with the real distance to caster. In both of these cases both the casters and the targets are players, so combat reaches should be relatively similar, but in the first sniff the real traveled distance ends up being 1.63545 yards shorter than the distance to caster, and in the second sniff it ends up being 5.98 and 4.8 yards shorter for respective targets. More than that, in videos of Halion fight I quickly found on youtube, I saw that players were pulled a very-very short distance, no matter how far they were from the player who dropped down Soul Consumption. Something along the lines of 5 yards, even shorter than it is in your sniffs. So, for whatever reason, some spells must pull targets to caster (they even say so in the description), but others only slightly nudge them in the direction of the caster.

Still up to you, TC devs, if you want to apply the changes to `SPELL_EFFECT_PULL_TOWARDS` handler as well or not, at the very least it will *improve* the behavior of the effect (make it clientside, allowing players to collide with solid surfaces, not fall through the ground, work on moving transports etc), but it still won't be anywhere near blizzlike, because thus far I don't have the slightest clue how is it even supposed to look like. Each spell behaves differently.

**Comment by Faq** [comment]
@Rushor You tested this some time ago, any feedback?

**Comment by Killyana** [comment]
The rocket pack works correctly, even in retail the player don't change orientation.

File: src/server/game/Entities/Unit/Unit.cpp [hunk 6]
```diff
@@ -12305,9 +12305,12 @@ uint32 Unit::GetModelForForm(ShapeshiftForm form, uint32 spellId) const
     return modelid;
 }
 
-void Unit::JumpTo(float speedXY, float speedZ, bool forward)
+void Unit::JumpTo(float speedXY, float speedZ, bool forward, Position const* dest)
```
jackpoz: should we use Optional<Position> for this ?

xvwyh: That would make for better semantics, yes.

jackpoz (changes requested):
Please change the optional parameter pointer to Optional<T>

jackpoz:
Thanks for the PR :)

PR #24181, opened 2020-02-20
Core/Server: correct timestamp format for shutdown/restart notification broadcasts

### LINKED ISSUE #24174: Correcting the Shutdown/Restart Notification Broadcasts ($5 Bounty)
**Description:**

The shutdown/restart warning notification system uses the wrong timestamp format. It should be like: [SERVER] Shutdown in 15:00 and NOT like: [SERVER] Shutdown in 15 Minute(s)

So, the currently the shutdown/restart notifications currently use the universal timestamp format that is used for things like viewing the server uptime in .server info, which you can modify at: C:\Source\src\common\Utilities\Util.cpp

Which is not recommended as it breaks many things that use it. So there must be another way of changing it, maybe by modifying the World.cpp file in C:\Source\src\server\game\World\World.cpp

VIDEO EVIDENCE:
 https://www.youtube.com/watch?v=f6v_yRupABE
https://www.youtube.com/watch?v=z8za1_ozG-U

Can provide more if you think that is not enough.

**Current behaviour:**

So we are using the incorrect timestamp format to print the shutdown/restart notifications to chat.

**Expected behaviour:**

We should be using the correct timestamp format to print the shutdown/restart notifications to chat. E.g. for setting 15 minutes on the timer.

[SERVER] Shutdown in 15:00
[SERVER] Shutdown in 14:00
[SERVER] Shutdown in 13:00
[SERVER] Shutdown in 12:00
[SERVER] Shutdown in 11:00
[SERVER] Shutdown in 10:00
[SERVER] Shutdown in 9:00
[SERVER] Shutdown in 8:00
[SERVER] Shutdown in 7:00
[SERVER] Shutdown in 6:00
<-- When the timer gets down to 05:00 it will then start to display seconds as shown below. -->
[SERVER] Shutdown in 5:00
[SERVER] Shutdown in 4:45
[SERVER] Shutdown in 4:30
[SERVER] Shutdown in 4:15
[SERVER] Shutdown in 4:00
[SERVER] Shutdown in 3:45
[SERVER] Shutdown in 3:30
[SERVER] Shutdown in 3:15
[SERVER] Shutdown in 3:00
[SERVER] Shutdown in 2:45
[SERVER] Shutdown in 2:30
[SERVER] Shutdown in 2:15
[SERVER] Shutdown in 2:00
[SERVER] Shutdown in 1:45
[SERVER] Shutdown in 1:30
[SERVER] Shutdown in 1:15
[SERVER] Shutdown in 1:00
[SERVER] Shutdown in 0:45
[SERVER] Shutdown in 0:30
[SERVER] Shutdown in 0:15

1. Login any character
2. Use either .server shutdown 900 or .server restart 900
3. See that the timestamp that is used is 15 Minute(s) and not 15:00

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. 92ac6ca0fee0 2020-02-17 00:25:23 +0100 (3.3.5 branch) (Win64, Debug, Static) (worldserver-daemon)

**Operating system:** 

Windows Server 2016 Standard

ForesterDev:
**Changes proposed:**

-  change the time output format for restart/shutdown notifications 
-  also check where we should use singular or plural forms for words: Day, Hour, Minute, Second

- [x] 3.3.5

**Issues addressed:** Closes #24174 

**Tests performed:** (Does it build, tested in-game, etc.) Build, tested
image

### DIFF
--- a/src/common/Common.h
+++ b/src/common/Common.h
@@ -69,13 +69,6 @@ enum TimeConstants
     IN_MILLISECONDS = 1000
 };
 
-enum TimeFormats
-{
-    TF_FULL_TEXT    = 1,    // 1 Days 2 Hours 3 Minutes 4 Seconds
-    TF_SHORT_TEXT   = 2,    // 1d 2h 3m 4s
-    TF_NUMERIC      = 3     // 1:2:3:4
-};
-
 enum AccountTypes
 {
     SEC_PLAYER         = 0,
--- a/src/common/Common.h
+++ b/src/common/Common.h
@@ -69,6 +69,13 @@ enum TimeConstants
     IN_MILLISECONDS = 1000
 };
 
+enum TimeFormats
+{
+    TF_FULL_TEXT    = 1,    // 1 Days 2 Hours 3 Minutes 4 Seconds
+    TF_SHORT_TEXT   = 2,    // 1d 2h 3m 4s
+    TF_NUMERIC      = 3     // 1:2:3:4
+};
+
 enum AccountTypes
 {
     SEC_PLAYER         = 0,
--- a/src/common/Utilities/Util.cpp
+++ b/src/common/Utilities/Util.cpp
@@ -16,6 +16,7 @@
  */
 
 #include "Util.h"
+#include "Common.h"
 #include "Containers.h"
 #include "IpAddress.h"
 #include <utf8.h>
--- a/src/common/Utilities/Util.cpp
+++ b/src/common/Utilities/Util.cpp
@@ -107,7 +107,7 @@ time_t GetLocalHourTimestamp(time_t time, uint8 hour, bool onlyAfterTime)
     return hourLocal;
 }
 
-std::string secsToTimeString(uint64 timeInSecs, TimeFormats timeFormat, bool hoursOnly)
+std::string secsToTimeString(uint64 timeInSecs, TimeFormat timeFormat, bool hoursOnly)
 {
     uint64 secs    = timeInSecs % MINUTE;
     uint64 minutes = timeInSecs % HOUR / MINUTE;
@@ -118,11 +118,11 @@ std::string secsToTimeString(uint64 timeInSecs, TimeFormats timeFormat, bool hou
     if (days)
     {
         ss << days;
-        if (timeFormat == TF_NUMERIC)
+        if (timeFormat == TimeFormat::Numeric)
             ss << ":";
-        else if (timeFormat == TF_SHORT_TEXT)
+        else if (timeFormat == TimeFormat::ShortText)
             ss << "d";
-        else // if (timeFormat == TF_FULL_TEXT)
+        else // if (timeFormat == TimeFormat::FullText)
         {
             if (days == 1)
                 ss << " Day ";
@@ -133,11 +133,11 @@ std::string secsToTimeString(uint64 timeInSecs, TimeFormats timeFormat, bool hou
     if (hours || hoursOnly)
     {
         ss << hours;
-        if (timeFormat == TF_NUMERIC)
+        if (timeFormat == TimeFormat::Numeric)
             ss << ":";
-        else if (timeFormat == TF_SHORT_TEXT)
+        else if (timeFormat == TimeFormat::ShortText)
             ss << "h";
-        else // if (timeFormat == TF_FULL_TEXT)
+        else // if (timeFormat == TimeFormat::FullText)
         {
             if (hours <= 1)
                 ss << " Hour ";
@@ -150,11 +150,11 @@ std::string secsToTimeString(uint64 timeInSecs, TimeFormats timeFormat, bool hou
         if (minutes)
         {
             ss << minutes;
-            if (timeFormat == TF_NUMERIC)
+            if (timeFormat == TimeFormat::Numeric)
                 ss << ":";
-            else if (timeFormat == TF_SHORT_TEXT)
+            else if (timeFormat == TimeFormat::ShortText)
                 ss << "m";
-            else // if (timeFormat == TF_FULL_TEXT)
+            else // if (timeFormat == TimeFormat::FullText)
             {
                 if (minutes == 1)
                     ss << " Minute ";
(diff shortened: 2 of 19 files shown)


LordWaycrest:
Can confirm this works, good job. :-)

src/common/Common.h
```diff
@@ -69,6 +69,13 @@ enum TimeConstants
     IN_MILLISECONDS = 1000
 };
 
+enum TimeFormats
```
Shauren: This enum belongs together with the function inside Util.h file, do not put it here and #include "Common.h" in Util.h (more includes is bad)

Shauren (changes requested):

illfated:
> LordWaycrest wrote:
> EDIT: Can I just ask, what are the other irrelevant commits you've included in this pull request? I'm not trying to be funny I just can't understand why.

Those commits are there to keep the PR branch up to date with the 3.3.5 branch (at least for the content of this PR), to avoid file conflicts.

LordWaycrest:
> LordWaycrest wrote:
> EDIT: Can I just ask, what are the other irrelevant commits you've included in this pull request? I'm not trying to be funny I just can't understand why.
> 
> Those commits are there to keep the PR branch up to date with the 3.3.5 branch (at least for the content of this PR), to avoid file conflicts.

Hi, thank you for clarifying! :-)

illfated:
> Hi, thank you for clarifying! :-)

You're welcome. The commits don't look quite like TC commits, but there may have been some back-and-forth during the creation of his PR, so I am guessing he didn't start off with a fully updated source before creating the PR branch.

LordWaycrest:
> Hi, thank you for clarifying! :-)
> 
> You're welcome. The commits don't look quite like TC commits, but there may have been some back-and-forth during the creation of his PR, so I am guessing he didn't start off with a fully updated source before creating the PR branch.

I can't wait for this PR to be merged if it ever happens. I've been wanting to get this sorted for a long time. I'm all about Blizz-like! :-)

illfated:
Sure, I can see that. I enjoy seeing people making constructive changes toward blizzlike content, especially when the end result gives the player a feeling of being in a game environment  with a very close to retail experience.

LordWaycrest:
> Sure, I can see that. I enjoy seeing people making constructive changes toward blizzlike content, especially when the end result gives the player a feeling of being in a game environment  with a very close to retail experience.

That is honestly my aim, people say to me "Why on earth would you focus on silly things like that? It all counts!

illfated:
@ForesterDev : would you mind adding the content changes from commit https://github.com/TrinityCore/TrinityCore/commit/9009c82a8db2bdb433cba17820af8b0cef8e28e3 to have all 4 PR tests working as intended?

src/common/Utilities/Util.h
```diff
@@ -26,6 +26,13 @@
 #include <utility>
 #include <vector>
 
+enum TimeFormats
```
Shauren: Make this `enum class`, drop TF_ prefix from each value and use CamelCase for naming

src/common/Utilities/Util.cpp
```diff
@@ -116,15 +116,75 @@ std::string secsToTimeString(uint64 timeInSecs, bool shortText, bool hoursOnly)
...
+            }
+        }
+        else
+        {
+            if (timeFormat == TimeFormat::Numeric)
+                ss << "0:";
+        }
+        if (secs || (!days && !hours && !minutes))
+        {
+            ss << secs;
```
Shauren: Will this format seconds value as a single digit when between 0 and 9?

ForesterDev: Yes, but for restart/shutdown notifications this situation is impossible, that why I did not process this case

Warpten: just use fmt

src/common/Utilities/Util.cpp
```diff
@@ -116,15 +117,75 @@ std::string secsToTimeString(uint64 timeInSecs, bool shortText, bool hoursOnly)
 
     std::ostringstream ss;
     if (days)
-        ss << days << (shortText ? "d" : " Day(s) ");
+    {
+        ss << days;
+        if (timeFormat == TimeFormat::Numeric)
```
soulfrost: why not switch?

soulfrost:
```c++
    switch (timeFormat)
    {
    case TimeFormat::Numeric:
        ss << ":";
        break;
    case TimeFormat::ShortText:
        ss << "d";
        break;
    case TimeFormat::FullText:
        [[fallthrough]];
    default:
        if (days == 1)
            ss << " Day ";
        else
            ss << " Days ";
        break;
    }
```

src/common/Utilities/Util.cpp
Shauren: @Warpten not worth using fmt separately for multiple parts of this function (and you can't wrap it all in one big format)

jackpoz:
Thanks for the PR :)

LordWaycrest:
Thank you those who helped, especially @ForesterDev :-)

LordWaycrest:
While this has been resolved, I actually have an idea that we could implement, later on, so obviously we can't make this format work for anything over 59 minutes, else it would look messed up. So I thought maybe we could limit the countdown timer to 59 minutes to solve that matter?

illfated:
Well, that might be useful on custom servers, because I can't remember seeing retail servers ever announcing an upcoming restart or shutdown in more than 30 or 45 minutes. Whoever chooses to use a warning for more than 59 minutes should consider this a special (custom) case. (At least that is what it looks like to me.)

LordWaycrest:
> Well, that might be useful on custom servers, because I can't remember seeing retail servers ever announcing an upcoming restart or shutdown in more than 30 or 45 minutes. Whoever chooses to use a warning for more than 59 minutes should consider this a special (custom) case. (At least that is what it looks like to me.)

Once their automated shutdown countdown kicked in early which resulted in it printing random messages to chat. People were seeing a 53:00 minute warning.

Blizzard_CSEU on Twitter even apologized for it.

QKoQKS1

Think I could make it a feature request like you could configure it in the worldserver.conf?

illfated:
Feature request ticket sounds fair enough. (I am not a TC member, though. Just speaking from general experience.)

PR #24188, opened 2020-02-23
Fixed a crash in spell_energize_aoe

kelno:


Shauren:
Do you have a crash log to back it up? `container.erase(itr++)` is a perfectly valid pattern

ccrs:
is it? 🤔

kelno:
I don't know about that.
This is on my fork so the stack will differ a little bit but:

This was detected as an issue with clang AddressSanitizer, for which I have a log:
[sanitizer.txt](https://github.com/TrinityCore/TrinityCore/files/4241223/sanitizer.txt)
(frame 7 in dustwallow_marsh.cpp is the erase line)

Then after restarting without the AddressSanitizer the same player made it crash for real. Also have a log:
image

Shauren:
I noticed your stack trace uses std::vector, we have std::list
Do you have more differences there than just container swap?

p.s. https://godbolt.org/z/629W-e

jackpoz:
might aswell just link the branch in your fork with all the changes when reporting an issue with modified code

kelno:
Okay, I'll detail under this but it's actually an invalid thing to do for a vector, so indeed it's not a crash on TC. But I think it's still more of a hack way to do it, we should use the same pattern everywhere and use a one that's not working only for some containers. Your choice then! Sorry for the false alarm.

And for the c++ enthusiasts, the why it's not okay for vector:
Removing the first element will offset everything in that vector back one position while the itr won't move, so it will point at the next element already. Then the itr++ will move it further again, skipping one element. The example code you gave works but only by chance, it will incorrectly skip the iteration on the second element. Inserting a second two in the container after the first one will make it obvious:  https://godbolt.org/z/_KP9MH 
I don't know why it does crashes exactly but in any case it's good that it does not allow it, it would be quite an invisible error than only fails sometimes depending on the order of the element.

Shauren:
Your explanation is not correct, vector.erase simply invalidates all iterators at or past the erased element, so `itr` is no longer valid on next iteration (no skipped element, simply invalid)
As for why the code was written this way - pre c++11, erase did not return a new iterator

kelno:
Alright thanks for the clarification

jackpoz:
@kelno http://www.cplusplus.com/reference/list/list/erase/ 
>Iterator validity
Iterators, pointers and references referring to elements removed by the function are invalidated.
All other iterators, pointers and references keep their validity.

The code "targets.erase(itr++);" does these things:
1. "itr" value is saved to a temporary variable (see operator ++ implementation). Let's call this A.
1. "itr" is incremented, pointing to the element next to the one that we want to remove (or .end() if there are no more elements). Let's call this B
1. the original "itr" value (A) is passed to targets.erase() . This removes A from the list. As std::list is implemented as a linked list, this means pretty much making the element before A point to B.
1. as "itr" was pointing to B already, targets.erase() doesn't invalidate it (see the quote above from http://www.cplusplus.com/reference/list/list/erase/  )

So tl;dr the current TC code works because it follows the rules of the standard, no undefined behavior involved.

jackpoz:
We'll keep the current code, thanks for the PR anyway

PR #24193, opened 2020-02-24
Core/Server: improve timestamp format output for large time values

### LINKED ISSUE #24191: [FEATURE REQUEST] [CONFIG OPTION] Make 59 minutes the maxmium a GM can set to issue a shutdown/restart notification
**Description:**

So we made changes to the shutdown/restart notifications to use a decimal format instead. So we managed to achieve this however it appears we can not make it work out for anything over 59 minutes, so I had an idea maybe we could add an option in the worldserver.conf to make 59 minutes the maximum a GM can issue a shutdown/restart notice. Or we could just completely limit it to avoid users making issues thinking it is a bug.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. 69231581e4f2 2020-02-22 21:31:37 +0100 (3.3.5 branch) (Win64, RelWithDebInfo, Static)

**Operating system:**

Windows Server 2016 Standard

ForesterDev:
**Changes proposed:**

-  impove time format for server restart/shutdown notifications if remain time > 1h or >1d

- [x] 3.3.5

**Issues addressed:** Closes #24191

**Tests performed:** (Does it build, tested in-game, etc.) Tested, build

### DIFF
--- a/src/common/Utilities/Util.cpp
+++ b/src/common/Utilities/Util.cpp
@@ -19,6 +19,7 @@
 #include "Common.h"
 #include "Containers.h"
 #include "IpAddress.h"
+#include "StringFormat.h"
 #include <utf8.h>
 #include <algorithm>
 #include <iomanip>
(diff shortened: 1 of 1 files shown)


Kitzunu:
Works great I think, got DD:HH:MM:SS
image

But if https://github.com/TrinityCore/TrinityCore/issues/24191#issuecomment-590219744 is correct, then we just need to make it hidden until 30:00

Shauren:
I would not bother with that, after all we will never know if they schedule the shutdown weeks in advance and only show countdown 30 (15 for live) minutes earlier or shutdown command is sent exactly at 30/15 minutes

Kitzunu (approved):

LordWaycrest:
Hi sorry I was working late last night, I can confirm it works properly now! :-)

jackpoz:
Thanks for the PR :)

PR #24200, opened 2020-02-26
Core/PacketIO: update SMSG_EMOTE / CMSG_EMOTE packets to new system

ForesterDev:
**Changes proposed:**

-  convert SMSG_EMOTE / CMSG_EMOTE packets to new system

**Target branch(es):**

- [x] 3.3.5

**Tests performed:** (Does it build, tested in-game, etc.) Tested, build

### DIFF
--- a/src/server/game/Handlers/ChatHandler.cpp
+++ b/src/server/game/Handlers/ChatHandler.cpp
@@ -22,6 +22,7 @@
 #include "Channel.h"
 #include "ChannelMgr.h"
 #include "Chat.h"
+#include "ChatPackets.h"
 #include "DatabaseEnv.h"
 #include "DBCStores.h"
 #include "GameTime.h"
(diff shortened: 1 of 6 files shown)


src/server/game/Server/Packets/ChatPackets.h
```diff
@@ -0,0 +1,51 @@
...
+    namespace Chat
+    {
+        class Emote final : public ServerPacket
+        {
+        public:
+            Emote() : ServerPacket(SMSG_EMOTE, 4 + 8) { }
+
+            WorldPacket const* Write() override;
+
+            int32 EmoteID = 0;
```
jackpoz: should this be uint32 ?

src/server/game/Handlers/ChatHandler.cpp
```diff
@@ -568,20 +569,19 @@ void WorldSession::HandleMessagechatOpcode(WorldPacket& recvData)
     }
 }
 
-void WorldSession::HandleEmoteOpcode(WorldPacket& recvData)
+void WorldSession::HandleEmoteOpcode(WorldPackets::Chat::EmoteClient& packet)
 {
-    if (!GetPlayer()->IsAlive() || GetPlayer()->HasUnitState(UNIT_STATE_DIED))
```
jackpoz: it's easier to review a PR if there are very few changes. this one looks more like some refactoring of the whole function than just using the new packt structure.

src/server/game/Server/Packets/ChatPackets.h
Ovahlord: yes. Even the client magic enum says it's uint32:

https://gist.github.com/Ovahlord/47695c854d14a26215de77e4c3a29e40#file-gistfile1-hpp-L5815

Shauren: no, its signed in client (but it doesnt matter because as you can see in the handler, only a few values are acceptable)

src/server/game/Handlers/ChatHandler.cpp
Ovahlord: Actually this is fine. The check was previously blocking the packet reading but now that the packet is always being read the check was just looking ugly up here so he moved it down to the other check. Just a cosmetic improvement, not a refactor.

jackpoz:
Thanks for the PR :)

Issue #22975, opened 2019-01-22
Core/SAI: Move to position and orientation


Killyana:
**Description:** The action SMART_ACTION_MOVE_TO_POS could use the target SMART_TARGET_POSITION that has x, y, z and orientation.

The issue here is the orientation is always ignored, so an extra action "set orientation" must be usedd once the position reached.

We need the orientation to be set for the npc once the position reached if the orientation is not 0

**Branch(es):** both

**TC rev. hash/commit:**  rev. 813cf761fcf8

Rushor:
yeah that would be really nice. in the other scriptsystem with `waypoints` the set of the orientation works fine.

Killyana:
We don't have orientation in waypoints table neither, it was added in waypoint_data only, adding orientation field in waypoints table will avoid using set orientation every time we need the npc to change orientation after a pause or at the end of the path.

offl:
MoveType (walk \ run \ fly) will be also useful in waypoints table & SMART_ACTION_MOVE_TO_POS
WaitTime in waypoints table would be also useful

Rushor:
Actually using waypoint data for sai is a good solution and add a sai action for calling waypoint-data. It's strange that we have 3 scripts systems for pathing: waypoints, waypoint-data, script-waypoints

offl:
> It's strange that we have 3 scripts systems for pathing: waypoints, waypoint-data, script-waypoints

And they are different from each other :rofl: 

The best we can use today is CMaNGOS' creature_movement & creature_movement_template with deletion of all another systems except splines and including script_waypoint.

They have the [PathID column](https://github.com/cmangos/mangos-tbc/commit/652382c43bee18affc2535172cbfe3aa6e208fae), this allow you not to create waypoints with ID*100 (888800 or 888801 or 888802). Just always 8888 + PathID column. Plus now you can load waypoint by changing creature_template.MovementType to 2. But now you will have to create waypoints for every creature even if they uses the same waypoint.
Plus you can [summon creature with the PathID parameter](https://github.com/cmangos/mangos-tbc/commit/652382c43bee18affc2535172cbfe3aa6e208fae#diff-08c5c4dca930b7acf7d5c045dc6f496bR1611). Without this you will never script [some NPCs properly without](https://github.com/cmangos/tbc-db/commit/b6e118eb01dcfab7493474715cc92be62a62b381#diff-e1b28cee751dc1b4835cb396492a25faR134) C++.

[...]

PR #24201, opened 2020-02-26
Core/SmartAI: Allow setting orientation for SMART_ACTION_MOVE_TO_POS

Sorikoff:
**Changes proposed:**
Pass orientation to MotionMaster when raw coordinates in SMART_ACTION_MOVE_TO_POS are given.

**Target branch(es):** 3.3.5

**Issues addressed:** Ref #22975

**Tests performed:** Does build, tested with custom SQL

~**Known issues and TODO list:**~

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -1476,7 +1476,7 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                     if (TransportBase* trans = me->GetDirectTransport())
                         trans->CalculatePassengerPosition(dest.x, dest.y, dest.z);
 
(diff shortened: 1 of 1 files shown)


Killyana:
Can you add : action_param5=movement_type=0/1 (run/walk)

Sorikoff:
In separate PR.

Ovahlord:
I'm still  not a fan of shoving orientation into move commands because blizzlike approaches would be creating a 2nd move call after reaching the desired destination. So IF you want to do both in one function call, make sure to call two splines. one to move to the position and one to face to the desired direction. That's how blizz does it.

Sorikoff:
Could you please clarify then why orientation can even be passed to MoveGen if it is not blizzlike approach?

Ovahlord:
Because someone probably thought that this is a nice way to do while it's actually wrong. As far as I have seen movement packets in sniffs, facing with orientation values are never being sent together with a destination. There is only one case where facing is being used together and that's chase movement. But even in this face the chase movement does not use facing angles but a facing target guid.

Sorikoff:
Thanks for clarification.

PR #24215, opened 2020-02-29
Add bash script for extracting client files

Teppic1:
**Changes proposed:**

Simple bash script to automate extracting/copying client files on a Linux/Unix system.

It follows the instructions on the wiki but additionally copies camera files if found. It includes a few simple checks.

**Target branch(es):** 3.3.5

I was able to extract and copy all files from my client.

### DIFF
--- a/contrib/extractor.sh
+++ b/contrib/extractor.sh
@@ -1,0 +1,180 @@
+#!/bin/bash
+
+# This is where the WoW client files are located
+CLIENT_DIR=/mnt/disk/client
+# This is where the installed server files are located
+SERVER_DIR=${HOME}/server
+
+# Check main dirs
+if [[ ! -d ${CLIENT_DIR} ]]
+then
+	echo "${CLIENT_DIR} client directory not found"
+	echo "Enter directory now, or press ctrl-c and edit the top of this script"
+	read CLIENT_DIR
+	echo
+fi
+
+if [[ ! -d ${SERVER_DIR} ]]
+then
+	echo "${SERVER_DIR} server directory not found"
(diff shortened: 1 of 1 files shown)


Faq:
# 22779

contrib/extractor.sh
```diff
@@ -0,0 +1,180 @@
+#!/bin/bash
+
+# This is where the WoW client files are located
+CLIENT_DIR=/mnt/disk/client
```
jackpoz: I think it's better to ask for the paths in this script

Teppic1: Updated.

contrib/extractor.sh
```diff
@@ -0,0 +1,184 @@
...
+VMAP4ASSEMBLER=${SERVER_DIR}/bin/vmap4assembler
+MMAPS_GENERATOR=${SERVER_DIR}/bin/mmaps_generator
+
+check_files() {
+for i in $MAPEXTRACTOR $VMAP4EXTRACTOR $VMAP4ASSEMBLER $MMAPS_GENERATOR
+do
+	if [[ ! -x $i ]]
+	then
+		echo "Invalid executable: $i"
+		echo "Check tools are installed correctly"
```
illfated: suggestion (for better understanding & readability in proper English):
```suggestion
		echo "Check that the extraction tools are installed correctly"
```

Teppic1: This is normal shell script error language, it looks a bit weird if you put errors into full normal English :)

jackpoz:
Thanks for the PR :)

PR #24216, opened 2020-02-29
Core: build on aarch64 with gcc

damaex:
**Changes proposed:**

-  fixed GCC build on aarch64
    - missing `mm_malloc.h` on gcc for aarch64
        - added functions `_mm_malloc` and `_mm_free`
        - like https://clang.llvm.org/doxygen/mm__malloc_8h_source.html
    - don't force SSE2 flags for aarch64 with gcc

**Target branch(es):**

- [x] 3.3.5

**Tests performed:**
- [x] build on raspberry pi 4 4GB (ubuntu 19.10 64bit + GCC 9.2.1)
- [x] connected and played 5 min with 2 accounts in parallel

### DIFF
--- a/src/common/Utilities/SFMTRand.cpp
+++ b/src/common/Utilities/SFMTRand.cpp
@@ -20,34 +20,39 @@
 #include <array>
 #include <functional>
 #include <random>
-#if defined(__clang__) && defined(__aarch64__)
-#include <mm_malloc.h>
-#elif defined(__GNUC__) && !defined(__clang__) && defined(__aarch64__)
-static __inline__ void *__attribute__((__always_inline__, __nodebug__, __malloc__))
-_mm_malloc(size_t __size, size_t __align) {
-    if (__align == 1) {
-        return malloc(__size);
-    }
-
-    if (!(__align & (__align - 1)) && __align < sizeof(void *))
-        __align = sizeof(void *);
-
-    void *__mallocedMemory;
-
-    if (posix_memalign(&__mallocedMemory, __align, __size))
-        return NULL;
-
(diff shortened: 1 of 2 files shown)


ghost (approved):
LGTM

src/common/Utilities/SFMTRand.cpp
```diff
@@ -20,8 +20,30 @@
 #include <array>
 #include <functional>
 #include <random>
-#if defined(__aarch64__)
+#if defined(__clang__) && defined(__aarch64__)
```
jackpoz: how about
```cpp
#if defined(__aarch64__)
  #if defined(__clang__)
  ...
  #elif defined(__GNUC__)
  ...
  #else
  "some error here"
  #endif
#else
...
```

jackpoz (changes requested):

src/common/Utilities/SFMTRand.cpp
```diff
@@ -20,12 +20,39 @@
...
 #include <random>
+#include <ctime>
+
 #if defined(__aarch64__)
-#include <mm_malloc.h>
+    #if defined(__clang__)
+        #include <mm_malloc.h>
+    #elif defined(__GNUC__)
+        static __inline__ void *__attribute__((__always_inline__, __nodebug__, __malloc__))
+        _mm_malloc(size_t __size, size_t __align) {
```
illfated: TC C++ codestyle: put opening curly brackets ` { ` on a new line (see lower part of the file for examples)

ghost (approved):

jackpoz:
Thanks for the PR :)

PR #24227, opened 2020-03-01
Make MySQL variant usage 100% typesafe

Lillecarl:
No longer rely on external enum, let the compiler sort everything out with usage of variant visitor.

**Changes proposed:**
-  Improve type saftey.

- [x] 3.3.5
- [ ] master

**Issues addressed:**
https://github.com/TrinityCore/TrinityCore/commit/3ebcb71c3dc80e0bea9a36207401a033712bda88#commitcomment-37555060
Better type saftey

**Tests performed:**
Builds, logged ingame, added an item, relogged, item was still there.

Notes:
See https://en.cppreference.com/w/cpp/utility/variant/visit

### DIFF
--- a/src/server/database/Database/MySQLPreparedStatement.cpp
+++ b/src/server/database/Database/MySQLPreparedStatement.cpp
@@ -231,51 +231,22 @@ std::string MySQLPreparedStatement::getQueryString() const
         pos = queryString.find('?', pos);
         std::stringstream ss;
 
-        switch (data.type)
-        {
-            case TYPE_BOOL:
-                ss << uint16(std::get<bool>(data.data)); // stringstream will append a character with that code instead of numeric representation
-                break;
-            case TYPE_UI8:
-                ss << uint16(std::get<uint8>(data.data)); // stringstream will append a character with that code instead of numeric representation
-                break;
-            case TYPE_UI16:
-                ss << std::get<uint16>(data.data);
-                break;
-            case TYPE_UI32:
-                ss << std::get<uint32>(data.data);
-                break;
-            case TYPE_I8:
-                ss << int16(std::get<int8>(data.data)); // stringstream will append a character with that code instead of numeric representation
-                break;
-            case TYPE_I16:
-                ss << std::get<int16>(data.data);
-                break;
-            case TYPE_I32:
-                ss << std::get<int32>(data.data);
-                break;
-            case TYPE_UI64:
-                ss << std::get<uint64>(data.data);
-                break;
-            case TYPE_I64:
-                ss << std::get<int64>(data.data);
-                break;
-            case TYPE_FLOAT:
-                ss << std::get<float>(data.data);
-                break;
-            case TYPE_DOUBLE:
-                ss << std::get<double>(data.data);
-                break;
-            case TYPE_STRING:
-                ss << '\'' << std::get<std::string>(data.data) << '\'';
-                break;
-            case TYPE_BINARY:
+        std::visit([&](auto&& data) {
+            // Convert some types to get correct string representation
+            using T = std::decay_t<decltype(data)>;
+            if constexpr (std::is_same_v<T, bool>)
+                ss << uint16(data);
+            else if constexpr (std::is_same_v<T, uint8>)
+                ss << uint16(data);
+            else if constexpr (std::is_same_v<T, int8>)
+                ss << int16(data);
+            else if constexpr (std::is_same_v<T, std::vector<uint8>>)
                 ss << "BINARY";
-                break;
-            case TYPE_NULL:
+            else if constexpr (std::is_same_v<T, decltype(nullptr)>)
                 ss << "NULL";
-                break;
-        }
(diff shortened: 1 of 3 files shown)


illfated:
I might be a bit overzealous, but maybe your test could include restarting the servers in addition to the relogging (inevitable, of course).

src/server/database/Database/MySQLPreparedStatement.cpp
```diff
@@ -231,51 +231,22 @@ std::string MySQLPreparedStatement::getQueryString() const
...
+                ss << uint16(data);
+            else if constexpr (std::is_same_v<T, uint8>)
+                ss << uint16(data);
+            else if constexpr (std::is_same_v<T, int8>)
+                ss << int16(data);
+            else if constexpr (std::is_same_v<T, std::vector<uint8>>)
                 ss << "BINARY";
-                break;
-            case TYPE_NULL:
+            else if constexpr (std::is_same_v<T, decltype(nullptr)>)
```
Naios: `decltype(nullptr)` -> `std::nullptr_t`

src/server/database/Database/PreparedStatement.cpp
```diff
@@ -36,51 +36,39 @@ void PreparedStatement::BindParameters(MySQLPreparedStatement* stmt)
...
+            else if constexpr (std::is_same_v<T, int32>)
+                stmt->setInt32(pos, data);
+            else if constexpr (std::is_same_v<T, int64>)
+                stmt->setInt64(pos, data);
+            else if constexpr (std::is_same_v<T, float>)
+                stmt->setFloat(pos, data);
+            else if constexpr (std::is_same_v<T, double>)
+                stmt->setDouble(pos, data);
+            else if constexpr (std::is_same_v<T, std::string>)
+                stmt->setString(pos, data);
```
Naios: `stmt->setString(pos, data);` -> `stmt->setString(pos, std::forward<decltype(data)>(data));`

src/server/database/Database/PreparedStatement.cpp
```diff
@@ -36,51 +36,39 @@ void PreparedStatement::BindParameters(MySQLPreparedStatement* stmt)
...
+            else if constexpr (std::is_same_v<T, std::vector<uint8>>)
+                stmt->setBinary(pos, data);
```
Naios: `stmt->setBinary(pos, data);` -> `stmt->setBinary(pos, std::forward<decltype(data)>(data));`

Naios (changes requested):
All in all the PR is fine, GJ! Sadly the heavy `constexpr if` clauses can't be avoided without refactoring the `stmt->setUInt32` setters.

Lillecarl:
> 
> 
> I might be a bit overzealous, but maybe your test could include restarting the servers in addition to the relogging (inevitable, of course).

Well I've started the server more than once which i would call "restarting". I put the PR's out early to see if it's likeable or not.

src/server/database/Database/PreparedStatement.cpp
Lillecarl: May i ask why this and the string method should have this whereas you don't want it for the primitive types?

Lillecarl:
> 
> 
> All in all the PR is fine, GJ! Sadly the heavy `constexpr if` clauses can't be avoided without refactoring the `stmt->setUInt32` setters.

I was looking at the mariadb C++ client library and since you set a type that's different on every datatype it seems like we would just move the ugliness elsewhere if we did this. One workaround could possibly be using some kind of association from the variant index to get the MySQL type, though that'd be rather iffy, and it wouldn't help with string and binary.

But yes there might still be room for improvement, if we could somehow associate variant index with MySQL type at compile time that'd be great, so if you have any clues I'd like to follow through.

Shauren:
I have more ideas for this

Lillecarl:
> 
> 
> I have more ideas for this

Please enlighten us bare mortals :p

src/server/database/Database/PreparedStatement.cpp
Naios: You can do it for the other types as well but it won't bring any benefit since you would prefer copying the smaller types sizeof(type)<= sizeof(void*) anyway. If you miss the perfect forwarding you will create a copy of the string or vector.

Also don't forget to do

```cpp
std::visit(..., std::move(data));
//                     ^^^^^
```

if you intend to fully consume the variant through the visit.

Lillecarl: @Naios Are you on IRC or Discord? I still don't really see the thing considering we're using auto&& in the visitor, discussion is a bit hard here and I'd like to learn.

Using I'm unsure about when we're finished with the variant or not, I don't know how we should proper safeguard never fetching the querystring twice. Or am i missunderstanding something here?

Thanks :smile:

Shauren:
30482038559d65648d9bcfa29cd426a6f266eeba

Lillecarl:
Thanks for the credits for the idea :p @Shauren

PR #24232, opened 2020-03-02
Core/Objects: MovePositonToFirstCollision will now use detour raycasts to determine terrain obstacles

Ovahlord:
**Changes proposed:**

-  MovePositionToFirstCollision will no longer rely on inaccurate vmap checks to detect terrain collisions. Instead we will now let the mmaps do their part by going with a detour raycast instead of spaming vmap height checks all over our target distance.
This will eliminate false positive collisions on uneven terrain with ocasional height bumps and overall return absolute precise collision results as long as the mmap quality is fine.
Hi @jackpoz 

- [x] 3.3.5
- [ ] master

- tested on my 434 fork.


Killyana:
3>c:\trinitysource355\src\server\game\entities\object\object.cpp(3304): error C2065: 'step' : Undeclared identifier
3>c:\trinitysource355\src\server\game\entities\object\object.cpp(3305): error C2065: 'step' : Undeclared identifier

Ovahlord:
@Killyana  if you do copy pasting by hand, please do it correctly :D There is not deathstar... eeeh step variable used anywhere anymore

Killyana:
I made some tests specially with blink, and results are not good, probably all movements spells that use vmaps hacks must be updated

illfated:
> [ ... ] :D There is no death star... [ ... ]

"These aren't the droids you're looking for." XD

Ovahlord:
@Killyana mind to give me some more info about "not good"?

Killyana:
Actually I get exactly the same issue before and after this PR, also blink don't use any script.
https://imgur.com/qKbCIXS

Ovahlord:
Oh that's because the entire wotlk Orgrimmar is a wmo and detour/vmaps have some issues with collision detection on wmo terrain.
Maybe a case for @jackpoz. Not directly related to this PR

Killyana:
Same issue here: .go xyz 989.5202, -4291.4619, 23.964901 1
image
If there wmo terrain angle is >0 it will not work properly.
But as you said it's not related to this PR.

jackpoz:
how about something like https://github.com/TrinityCore/TrinityCore/compare/3.3.5...jackpoz:fixes/3.3.5-better-blink ?

jackpoz:
Moved mmaps recast before vmaps checks

jackpoz:
Thanks for the PR :)

PR #23603, opened 2019-07-17
[3.3.5] Pooling system rewrite

Treeston:
## Guess who's back, baby.

So, spawn pooling. You take a bunch of spawns, and you put them in a pool. And then you tell the pool to spawn `x` spawns, and it spawns `x` random spawns from the pool based on the chances specified for each of them. And everything plays nice if grids get loaded and unloaded and stuff, and if one spawn in the pool goes away then when its respawn timer expires another random spawn from the pool is spawned. And you can also put pools in other pools, so you can have a pool that decides which of its sub-pools to spawn, and then each sub-pool picks something to spawn from itself. Gathering nodes do that one.

Anyway. I tore the whole thing out, since it was built around old persistent entities (like, old _old_ entities, pre-dynspawn old), and remade the whole thing since we have this fancy concept of a respawn timer that's not tied to any object now and stuff, which makes stuff _way_ easier by comparison. The old system had to hack its way around a lot of things (and was probably broken in many small and creative ways, if we're being entirely honest).

---

# Changelist

* New `SPAWN_TYPE_POOL (2)`.
  * Multiple respawn timers of the same spawnId are legal now.
* Split quest pooling logic from spawn pooling logic. They both use `pool_template`, though I might split it off too before this is over.
  * Standardized weekly and monthly respawns to also happen at the daily reset hour while I was at it
* Pools can no longer be cross-map. Thread safety and all that.
  * Pools are now per-map. You can have pools inside instances if you feel like it.
* Tons of on-load sanity checking of pool spawns. All those pool fixes pushed to 3.3.5 recently? Yeah.
  * Did we know our Alterac Mountains pool contained nodes in the Barrens? It did!
  * Our Duskwood pool had some nodes in Desolace!
  * _(we call them happy little features)_
* Some cleanups to pooling chance specification
  * Chances no longer need to add to `100.0` - weighting is relative ( `2.0` is twice as likely as `1.0`)
  * No more implicit equal chancing - specify explicit chances for everything (there's a conversion script)
* Merged all respawn timers into `respawn` (from old `creature_respawn` and `gameobject_respawn`) [374597c]
* Merged all spawn pool members into `pool_members` (from old `pool_creature`, `pool_gameobject` and `pool_pool`) [45bc91c]

---

Apart from those, everything _should_ work the same.

You know the drill, it probably doesn't. Go break things.

# To do:
- [ ] database update queries still assume one timer per spawnid, needs fix

PR #24283, opened 2020-03-12
Core/Pools: Attempt to workaround pools being broken since 2015

### LINKED ISSUE #21331: Core/Pools: Objects in the pool stop spawning
**Description:** Creatures (probably GameObjects too) that belong to a pool, will stop to respawn after some time. I can confirm this happens with Webbed Crusaders(30268 30273) and Serfex the Reaver(28083), for example.
I only monitored it for a few hours, but I think they won't respawn again until a server restart.

**Current behaviour:** After some number os respawns, the creature no longer spawn.

**Expected behaviour:** Creatures should keep respawning within the right time.

To force it happen:
1. Execute: DELETE FROM characters.respawn;
                  UPDATE world.creature SET spawntimesecs = 3 where id IN (30273,30268);
2. Start server
3. .gm on
4. .go xyz 6394 229 395 571
5. .cast 44659
6. Repeat step 3 a few times, usally after the second/third time the Webbed Crusaders won't spawn anymore.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** e45d54d80e3f

jackpoz:
**Changes proposed:**

-  Attempt to workaround pools being broken since 2015

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #21331

- Steps described in #21331

- This is not the most perfect fix of the universe that some want but 5 years of broken code is long enough to try fixing it.
- It attempts to fix only Creatures, it doesn't attempt to fix GameObjects


Killyana:
I made some tests, and I managed to get 2 spawns from the same pool spawned at the same time:
image

What I did is just killing all creature in this room and restarted the server, upon login  SpawnID: 21318 was already spawned, after a while SpawnID: 86077 spawned with the 3 others.

``` sql
SELECT * FROM `creature` WHERE `guid` IN (21318,86077);
SELECT * FROM `pool_members` WHERE `SpawnID` IN (21318,86077);
SELECT * FROM `pool_template` WHERE `entry` IN (32493);

Faq:
Wasnt better with #23603?

jackpoz:
@Faq sure, just fix the merge conflicts and fix the remaining todo point then. Also do that with approval from the original author.

Killyana:
So I figure out that if the unit that was chosen to respawn is different from the one that get killed, it will respawn instantly in sever restart, and the other unit will also respawn on top of it

sirikfoll:
I can confirm this works as a fix for #21331. Even if only for creatures, still a huge improvement

joshwhedon:
I am not convinced we need to do it via respawn. We don't want to respawn, we want to call Spawn1Object and not despawn it at the end of the process. Also doing so would fix gameobjects too.

jackpoz:
@joshwhedon have you tried with Spawn1Object ?

jackpoz:
I added a different attempt at https://github.com/TrinityCore/TrinityCore/pull/24949

jackpoz:
Replaced by https://github.com/TrinityCore/TrinityCore/pull/24949

PR #24300, opened 2020-03-14
Scripts/BlackfathomDeeps: Kelris will now channel his spell

### LINKED ISSUE #24292: Creature/Script: Boss Kelris
**Description:** Kelris must channel the spell 8734 on reset.
If use this diff, it will only be channeled correctly on server start, but on reset it will get interrupted.
``` diff
diff --git a/src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp b/src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
index 272817a24b..5c6f67bf5c 100644
--- a/src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
+++ b/src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
@@ -21,8 +21,9 @@
 
 enum Spells
 {
-    SPELL_MIND_BLAST    = 15587,
-    SPELL_SLEEP         = 8399,
+    SPELL_MIND_BLAST                    = 15587,
+    SPELL_SLEEP                         = 8399,
+    SPELL_BLACKFATHOM_CHANNELING        = 8734
 };
 
 enum Texts
@@ -61,6 +62,11 @@ public:
             _JustDied();
         }
 
+        void Reset() override
+        {
+            DoCast(me, SPELL_BLACKFATHOM_CHANNELING);
+        }
+
         {

                 if (!UpdateVictim())
``` 
Also if you aggro it when he channel the spell it will not get interrupted, and the boss remains rooted.

1. Apply the diff
2. .go c id 4832

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  rev. a40b742c3ae8

Lumber-Jack-2:
**Changes proposed:**

-  Kelris will now channel his spell
-  https://github.com/TrinityCore/TrinityCore/issues/24292
-  

- 3.3.5

-  https://github.com/TrinityCore/TrinityCore/issues/24292

yes

- [ ] nothing
- [ ]

### DIFF
--- a/src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
+++ b/src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
@@ -51,13 +51,13 @@ public:
         void Reset() override
         {
             _Reset();
-            DoCast(me, SPELL_BLACKFATHOM_CHANNELING, true);
+            DoCastSelf(SPELL_BLACKFATHOM_CHANNELING);
         }
 
         void JustReachedHome() override
         {
             _JustReachedHome();
-            DoCast(me, SPELL_BLACKFATHOM_CHANNELING, true);
+            DoCastSelf(SPELL_BLACKFATHOM_CHANNELING);
         }
 
         void JustEngagedWith(Unit* who) override
--- a/src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
+++ b/src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
@@ -21,8 +21,9 @@
 
 enum Spells
 {
-    SPELL_MIND_BLAST    = 15587,
-    SPELL_SLEEP         = 8399,
+    SPELL_MIND_BLAST             = 15587,
+    SPELL_SLEEP                  = 8399,
+    SPELL_BLACKFATHOM_CHANNELING = 8734
 };
 
 enum Texts
@@ -47,10 +48,23 @@ public:
     {
         boss_kelrisAI(Creature* creature) : BossAI(creature, DATA_KELRIS) { }
 
+        void Reset() override
+        {
+            _Reset();
+            DoCast(me, SPELL_BLACKFATHOM_CHANNELING, true);
+        }
+
+        void JustReachedHome() override
+        {
+            _JustReachedHome();
+            DoCast(me, SPELL_BLACKFATHOM_CHANNELING, true);
+        }
+
         void JustEngagedWith(Unit* who) override
         {
             BossAI::JustEngagedWith(who);
             Talk(SAY_AGGRO);
+            me->RemoveAurasDueToSpell(SPELL_BLACKFATHOM_CHANNELING);
             events.ScheduleEvent(EVENT_MIND_BLAST, 2s, 5s);
             events.ScheduleEvent(EVENT_SLEEP, 9s, 12s);
         }


Killyana:
Look a little bit hacky, npcs using SAI don't have any issue with this spell, it's cast by the event 25 (reset) and nothing else

Rushor:
i think bossai and smartai are different in handling spellcasts in Reset()

Ovahlord:
Use DoCastSelf instead of DoCast(me
Also remove that RemoveAurasDueToSpell call there because the aura already has interrupt flags that will take care of the interrupt.

Rushor:
Nope the aura will Not be removed on aggro

Killyana:
I removed everything and kept only the reset hook, to be sure nothing mess with the spell cast, butt still the channeling is interrupted upon cast (on reset), and if you use .cast self 8734 then aggro it will not interrupt it.

``` cpp
#include "ScriptMgr.h"
#include "ScriptedCreature.h"
#include "blackfathom_deeps.h"

enum Spells
{
    SPELL_MIND_BLAST             = 15587,
    SPELL_SLEEP                  = 8399,
    SPELL_BLACKFATHOM_CHANNELING = 8734
};

class boss_kelris : public CreatureScript
{
public:
    boss_kelris() : CreatureScript("boss_kelris") { }

    struct boss_kelrisAI : public BossAI
    {
        boss_kelrisAI(Creature* creature) : BossAI(creature, DATA_KELRIS) { }
		
        void Reset() override
        {
            _Reset();
            DoCastSelf (SPELL_BLACKFATHOM_CHANNELING, true);
        }
    };

    CreatureAI* GetAI(Creature* creature) const override
    {
        return GetBlackfathomDeepsAI<boss_kelrisAI>(creature);
    }
};

void AddSC_boss_kelris()
{
    new boss_kelris();
}
```

Rushor:
hmm and what happens if we remove the triggered flag? aka 
DoCastSelf (SPELL_BLACKFATHOM_CHANNELING);

Killyana:
Exactly the same result.

Rushor:
ID - 8734 Blackfathom Channeling
Interrupt Flags: 0x00000000, AuraIF 0x0000101F, ChannelIF 0x0000101B

sth with the aura is wonky maybe? but that does not explain why it works with sai

Killyana:
It works just fine with SAI, the spell is just cast on reset, nothing else https://imgur.com/a/WAHt2W1

Rushor:
ah yeah if you look into smartai.cpp we have this:

GetScript()->OnReset();

it is calles in: 
void SmartAI::JustAppeared()
void SmartAI::JustReachedHome()
void SmartAI::EnterEvadeMode(EvadeReason /*why*/)

in Bossai we do not have that and musst apply  that code manually there

Also:

void SmartAI::JustEngagedWith(Unit* enemy)

calls me->InterruptNonMeleeSpells(false); // must be before ProcessEvents

which we also do not have in Bossai

That is why we have to add all that shit from sai into bossai

Rushor:
So all those changes are not hack, but instead the valid way of adding and removing the aura

src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
```diff
@@ -47,10 +48,23 @@ class boss_kelris : public CreatureScript
...
+        void Reset() override
+        {
+            _Reset();
+            DoCast(me, SPELL_BLACKFATHOM_CHANNELING, true);
+        }
+
+        void JustReachedHome() override
+        {
+            _JustReachedHome();
+            DoCast(me, SPELL_BLACKFATHOM_CHANNELING, true);
```
Keader: this looks a hacky.
If i remember right it will work like:

Evade -> JustReachedHome -> Reset

When it happens, SPELL_BLACKFATHOM_CHANNELING will called twice.
And still, dont use DoCast(me, xxxx, ?true/false);
use: DoCastSelf(xxx) method.

Rushor: yes it should cast that - but it actually does not, therefore i added it here

will update the DoCastSelf

Keader: You are treating the symptom not the disease

Rushor: oh i just see

void BossAI::_JustReachedHome()
{
    me->setActive(false);
}

so no reset call?

Keader: Hmm, someone removed It ?
Put a breakpoint in debug mode to be sure.

Rushor: yep just tested. reset is called once the creature starts evading, but not again once the creature reaches the homeposition

Keader: So still has the Double call

Rushor: yep - sai in general has more calls and connections to other functions than bossai - they are not synchronous

Killyana: @Keader here https://imgur.com/a/a6qL2Ng the spell is cast on reset, but it get interrupted because the reset is called when the boss starts evading and not when it reach home position.

Keader: you guys checked on sniffs, if has no delay in cast?
because... to me, looks like just missing a little event (eventmap) to handle with it.
and still is weird... Reset was called when creature reachs home position, not before, when it changed?

Killyana: I don't know, but the script of this boss doesn't really matter (also I checked sniffs and it just channel the spell)
What matter is fixing the reset for (ScriptedAI, BossAI, WorldBossAI)

Rushor: yep i would suggest to merge the functions that are called from sai to worldbossai and bossai. sai works the best for event stuff

Killyana: It must be implemented in ScriptedAI as worldbossai and bossai inherit every thing from there.

src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
```diff
@@ -47,10 +48,23 @@ class boss_kelris : public CreatureScript
     {
         boss_kelrisAI(Creature* creature) : BossAI(creature, DATA_KELRIS) { }
 
+        void Reset() override
+        {
+            _Reset();
+            DoCastSelf(SPELL_BLACKFATHOM_CHANNELING);
+        }
+
+        void JustReachedHome() override
```
Killyana: Remove this as the reset is fixed

src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
```diff
@@ -47,10 +48,23 @@ class boss_kelris : public CreatureScript
...
+        {
+            _JustReachedHome();
+            DoCastSelf(SPELL_BLACKFATHOM_CHANNELING);
+        }
+
         void JustEngagedWith(Unit* who) override
         {
             BossAI::JustEngagedWith(who);
             Talk(SAY_AGGRO);
+            me->RemoveAurasDueToSpell(SPELL_BLACKFATHOM_CHANNELING);
```
Killyana: Remove as it will be done by the native AI

src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
illfated: Took me a little while to realize that this means "including the next 5 lines".

src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
```diff
@@ -47,10 +48,23 @@ class boss_kelris : public CreatureScript
...
+        {
+            _Reset();
+            DoCastSelf(SPELL_BLACKFATHOM_CHANNELING);
+        }
+
+        void JustReachedHome() override
+        {
+            _JustReachedHome();
+            DoCastSelf(SPELL_BLACKFATHOM_CHANNELING);
+        }
```
Faq:
```suggestion

```

src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
Faq: Use suggestion function in github

src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
Faq:
```suggestion
```

Faq (changes requested):

src/server/scripts/Kalimdor/BlackfathomDeeps/boss_kelris.cpp
illfated: Very well, although I think this suggestion needs to include the blank line after the current code lines.

Killyana:
No need to change any thing in this PR after https://github.com/TrinityCore/TrinityCore/commit/92447425279151373e6608e18214c7166bc7f46e
It can be push as it is.

ghost (approved):

Rushor:
why was the general fix not working @Killyana (cannot find the discussion in irc) :o

Killyana:
It caused some issue with some boss after a reset.

jackpoz:
Thanks for the PR :)

PR #24345, opened 2020-03-22
Core/AI: Fix reset and interruption of non melee spells

Killyana:
**Changes proposed:**

-  Move the reset to JustReachedHome() instead of EnterEvadeMode()
-  Interrupt non melee spells in EngagementStart()
-  

- [x] 3.3.5
- [ ] master

**Issues addressed:** It fixes issues encountered in this PR https://github.com/TrinityCore/TrinityCore/pull/24300

**Tests performed:** Does it build: Yes, tested in-game: Yes

- [ ]  Review

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -205,6 +205,11 @@ void CreatureAI::JustAppeared()
     }
 }
 
+void CreatureAI::JustReachedHome()
+{
+    Reset();
+}
+
 void CreatureAI::JustEnteredCombat(Unit* who)
 {
     if (!IsEngaged() && !me->CanHaveThreatList())
@@ -234,7 +239,6 @@ void CreatureAI::EnterEvadeMode(EvadeReason why)
         }
     }
 
-    Reset();
 }
 
 bool CreatureAI::UpdateVictim()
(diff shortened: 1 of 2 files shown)


illfated:
This makes sense. After all, `JustReachedHome()` is called at the end of any evade or reset scenario, so this logic looks good to me.

Rushor:
nice!

src/server/game/AI/CreatureAI.cpp
```diff
@@ -276,6 +280,7 @@ void CreatureAI::EngagementStart(Unit* who)
     }
     _isEngaged = true;
 
+    me->InterruptNonMeleeSpells(false);
```
offl: Some creatures must continue channeling after aggro

Killyana: Interrupting the channeling is the rule, continue is the exception, I cannot remember a creature or boss that continue channeling once aggro.

offl: Isn't there a attribute to don't interrupt channeling on aggro?
I'm 100% sure it is invalid for master but I can't remember any wotlk creature with exception. Cmangos does it manually in each script even in tbc branch

Maybe it affects Magtheridon now?

PR #24347, opened 2020-03-23
CPack Support

myuzhobcplidtkieno:
**Changes proposed:**

-  Add the ability to use CPack
-  Defaults to generating deb packages if the 'package' build target is specified

- [x] 3.3.5
- [ ] master

- [x] Dynamic scripts aren't supported
- [x] Hard-coded download location of full database

### DIFF
--- a/cmake/package.cmake
+++ b/cmake/package.cmake
@@ -1,0 +1,88 @@
+# This file is part of the TrinityCore Project. See AUTHORS file for Copyright information
+#
+# This file is free software; as a special exception the author gives
+# unlimited permission to copy and/or distribute it, with or without
+# modifications, as long as this notice is preserved.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY, to the extent permitted by law; without even the
+# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+if("${SCRIPTS}" MATCHES "dynamic")
+  message(FATAL_ERROR "Packaging doesn't support dynamic scripts.")
+  return()
+endif()
+
+# Define install candidates for SQL files
+include(GNUInstallDirs)
+install(DIRECTORY "${CMAKE_SOURCE_DIR}/sql" DESTINATION "${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME}")
+
+# Download the full database
+set(rev_database_version "${rev_version_major}${rev_version_minor}${rev_version_patch}")
+set(FULL_DATABASE_URL "https://github.com/TrinityCore/TrinityCore/releases/download/TDB335.${rev_database_version}/TDB_full_world_335.${rev_database_version}_${rev_database_date}.7z")
+message("* Package database URL   : ${FULL_DATABASE_URL}")
+get_filename_component(FULL_DATABASE_NAME "${FULL_DATABASE_URL}" NAME)
+string(REGEX REPLACE "\\.[^.]*$" "" FULL_DATABASE_NAMEWE ${FULL_DATABASE_NAME})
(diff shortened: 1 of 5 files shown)


jackpoz:
https://github.com/jackpoz/TrinityCoreCron/blob/3.3.5-TDB/release-tdb.sh might need to be updated too

cmake/package.cmake
```diff
@@ -0,0 +1,91 @@
...
+endif()
+
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${FULL_DATABASE_NAMEWE}.sql" DESTINATION "${CONF_DIR}")
+
+if(NOT CPACK_GENERATOR)
+  set(CPACK_GENERATOR "DEB")
+endif()
+if(SERVERS AND TOOLS)
+  set(CPACK_PACKAGE_NAME "trinity")
+  set(CPACK_DEBIAN_PACKAGE_CONFLICTS "trinity-server, trinity-tools")
```
Shauren: You should make this depend on branch too so that both 3.3.5 and master can be installed side by side

myuzhobcplidtkieno:
I've made a PR for TrinityCoreCron to keep it inline with these changes: https://github.com/jackpoz/TrinityCoreCron/pull/4

cmake/package.cmake
myuzhobcplidtkieno: I think that would require renaming the binaries too, otherwise they would conflict

revision_data.h.in.cmake
```diff
@@ -9,7 +9,7 @@
  #define _SOURCE_DIRECTORY          R"(@CMAKE_SOURCE_DIR@)"
  #define _BUILD_DIRECTORY           R"(@BUILDDIR@)"
  #define _MYSQL_EXECUTABLE          R"(@MYSQL_EXECUTABLE@)"
- #define _FULL_DATABASE             "TDB_full_world_335.20031_2020_03_16.sql"
+ #define _FULL_DATABASE             "TDB_full_world_335.@rev_version_major@@rev_version_minor@@rev_version_patch@_@rev_database_date@.sql"
```
jackpoz, line 12: does this mean the version naming will be changed ? Like what would 20031 become with the new system ?

jackpoz: since https://github.com/TrinityCore/TrinityCore/commit/3358cc2a4772aedf6415b5ba7ca3fdbbf808da24#commitcomment-32815761 (scroll down at the comment at the bottom, it's a huge commit) we changed the TDB versioning to follow this syntax:
| Header | TDB | {branch} | . | {last 2 digits of the year} | {2 digits of the month} | {counter of TDB releases for the month  starting from 1} |
| --- | --- | --- | --- | --- | --- | --- |
| Today's release | TDB | 335 | . | 19 | 03 | 1 |
| Next week's release (if any) | TDB | 335 | . | 19 | 03 | 2 |
| Next month's release | TDB | 335 | . | 19 | 04 | 1 |

I would like to keep the {year}{month}{counter} format if possible without having to handle a major.minor.patch that for TC development really doesn't have much of a meaning (we don't have a stable release schedule and we dropped counting TDB with incremental numbers).

Is major.minor.patch a requirement by CPack ?

cmake/package.cmake
Shauren: Kind of want this fixed actually, binaries dont need to be renamed if you just install to a different directory (CMAKE_INSTALL_PREFIX or CPACK_PACKAGE_INSTALL_DIRECTORY)

jackpoz:
Please check https://github.com/TrinityCore/TrinityCore/pull/24347#discussion_r436260978 comment

jackpoz:
Converting this PR to draft as there is still stuff to do and the author doesn't seem to be active

PR #24350, opened 2020-03-23
Core/Dep: Rethrow the original exception object using an empty throw.…

danthespal:
… Otherwise a copy of e is created and the original exception type is lost and replaced by its (potential) super-type ParseError.

**Changes proposed:**

-  
-  
-  

- [ ] 3.3.5
- [ ] master

- [ ] 
- [ ]

### DIFF
--- a/dep/g3dlite/source/AnyTableReader.cpp
+++ b/dep/g3dlite/source/AnyTableReader.cpp
@@ -12,7 +12,7 @@ AnyTableReader::AnyTableReader(const std::string& name, const Any& a) : m_any(a)
         // invoked automatically.
         m_any.~Any();
         m_alreadyRead.~Set();
-        throw e;
+        throw;
     }
 }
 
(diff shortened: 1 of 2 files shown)


jackpoz:
hey, thanks for the PR. We rarely improve the code in the /dep folder as they are supposed to be kept as close as possible to the original external code (dep stands for external dependencies, in this case https://casual-effects.com/g3d/www/index.html ).
It might be that we don't even use that function and it could be there just to be able to build it.

We are currently using this version of G3D: https://github.com/TrinityCore/TrinityCore/blob/12c10220309b22bbeabe4070023cf0c94ac4dcae/dep/PackageList.txt#L29

It might be a more recent version has fixed the same issue too but we decided not to update for other reasons (breaking changes and additional libraries in latest version)

danthespal:
ok thanks :D I will look at the other things then :D

illfated:
Thank you for the interesting input, might be nice to know when dealing with rare exceptions.

PR #24365, opened 2020-03-29
Core/Spells: fixed a possible client crash when casting pull towards dest spells with misc value 0

Ovahlord:
**Changes proposed:**

-  <Shauren> one does not simply divide by 0

The pull towards dest spell effect handler was not protected against potential 0 dividers when the misc value of the effects is 0. This means the packet will contain a infinite speedZ value which will ultimately cause a client crash. That's fixed now.

- [x] 3.3.5
- [ ] master

- tested ingame on 434 which has plenty of such scenarios

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -4422,7 +4422,7 @@ void Spell::EffectPullTowardsDest(SpellEffIndex effIndex)
     float distXY = unitTarget->GetExactDist(pos);
     float distZ = pos->GetPositionZ() - unitTarget->GetPositionZ();
 
-    float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
+    float speedXY = m_spellInfo->Effects[effIndex].MiscValue ? m_spellInfo->Effects[effIndex].MiscValue / 10.0f : 30.0f;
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/SpellEffects.cpp
```diff
@@ -4422,7 +4422,7 @@ void Spell::EffectPullTowardsDest(SpellEffIndex effIndex)
     float distXY = unitTarget->GetExactDist(pos);
     float distZ = pos->GetPositionZ() - unitTarget->GetPositionZ();
 
-    float speedXY = m_spellInfo->Effects[effIndex].MiscValue / 10.0f;
+    float speedXY = m_spellInfo->Effects[effIndex].MiscValue ? m_spellInfo->Effects[effIndex].MiscValue / 10.0f : 30.0f;
```
jackpoz: is the 30.0f some sort of standard value ?

Ovahlord: https://github.com/TrinityCore/TrinityCore/blob/6835fb9dcb418f8ced09cc1c8244d646cd2abb98/src/server/game/Spells/SpellEffects.cpp#L4400 yes. It's just missing here because nobody was thinking about that it's possible here as well

jackpoz:
Thanks for the PR :)

illfated:
> - one does not simply divide by 0

... also, thanks for the meme-fied bullet point list item 😄

PR #24372, opened 2020-03-30
Entities/Player: Restore old checks for UpdatePvPFlag

### LINKED ISSUE #24280: PvP mode is automatically turned off
Hello! There is such a problem: if you turn on pvp mode, and a player from the opposite fraction tries to attack you, then he turns on pvp mode for a second and turns off automatically.

Branch(es): 3.3.5
TDB: 335.20021
Ubuntu 18.04
Rev. e7d2489

Faq:
**Changes proposed:**

-  restore old checks what were wiped in https://github.com/TrinityCore/TrinityCore/commit/671052575a36624b06379ee5e13a7fea26870194
-  remove some not working code in https://github.com/TrinityCore/TrinityCore/commit/671052575a36624b06379ee5e13a7fea26870194
-  now pvp can be engaged

- [x ] 3.3.5

Closes #24280

Compiles and tested with horde and alliance char in Alliance territory. 
Alliance char attacked Horde and got PVP flag, but after 5min flag wear off and Horde char became Hostile. 
https://wowwiki.fandom.com/wiki/PvP_flag

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -20590,14 +20590,19 @@ void Player::ResetContestedPvP()
 
 void Player::UpdatePvPFlag(time_t currTime)
 {
+    if (!IsPvP())
+        return;
+
+    if (!pvpInfo.EndTimer || (currTime < pvpInfo.EndTimer +300) || pvpInfo.IsHostile)
+        return;
+
     if (pvpInfo.EndTimer && pvpInfo.EndTimer <= currTime)
     {
         pvpInfo.EndTimer = 0;
         RemoveFlag(PLAYER_FLAGS, PLAYER_FLAGS_PVP_TIMER);
     }
 
-    if (IsPvP() && !pvpInfo.IsHostile && !HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_IN_PVP | PLAYER_FLAGS_PVP_TIMER))
-        UpdatePvP(false);
+    UpdatePvP(false);
 }
 
 void Player::UpdateDuelFlag(time_t currTime)
@@ -21906,14 +21911,9 @@ void Player::UpdateHomebindTime(uint32 time)
 
 void Player::InitPvP()
 {
+    // pvp flag should stay after relog
     if (HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_IN_PVP))
         UpdatePvP(true, true);
-    else if (HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_PVP_TIMER))
-    {
-        UpdatePvP(true, true);
-        if (!pvpInfo.IsHostile)
-            pvpInfo.EndTimer = GameTime::GetGameTime() + 300;
-    }
 }
 
 void Player::UpdatePvPState(bool onlyFFA)
@@ -21940,7 +21940,7 @@ void Player::UpdatePvPState(bool onlyFFA)
     if (onlyFFA)
         return;
 
-    if (pvpInfo.IsHostile)                               // in hostile area
+    if (pvpInfo.IsHostile)                                  // in hostile area
     {
         if (!IsPvP() || pvpInfo.EndTimer)
             UpdatePvP(true, true);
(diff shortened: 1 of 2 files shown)


jackpoz:
is the original issue "Properly restore PvP-enabled state on login" that the commit fixed still working too ?

Faq:
There is no issue linked like that, no exact information what to test with repo steps.

Tested this: Enable PVP-> Log off-> log on-> there is still PVP fag on-> turned off PVP-> Got in log msg that in 5min will be switched off PVP flag->After 5min it was off.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -20537,6 +20537,12 @@ void Player::ResetContestedPvP()
 
 void Player::UpdatePvPFlag(time_t currTime)
 {
+    if (!IsPvP())
+        return;
+
+    if (!pvpInfo.EndTimer || currTime < pvpInfo.EndTimer || pvpInfo.IsHostile)
+        return;
```
jackpoz: these checks make most of the checks after this lines redundant. could you clean up the function a bit (while keeping the fix) ?

Faq: This is no go, need more work.

Faq: rechecked and commited changes 8 days ago

Faq:
@Treeston any comments about this? :)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -21940,15 +21940,15 @@ void Player::UpdatePvPState(bool onlyFFA)
...
+    if (pvpInfo.IsHostile)                                  // in hostile area
     {
         if (!IsPvP() || pvpInfo.EndTimer)
             UpdatePvP(true, true);
     }
     else                                                    // in friendly area
     {
         if (IsPvP() && !HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_IN_PVP) && !pvpInfo.EndTimer)
-            pvpInfo.EndTimer = GameTime::GetGameTime() + 300;                  // start toggle-off
+            pvpInfo.EndTimer = GameTime::GetGameTime();     // start toggle-off
```
Faq, line 21951: No need for +300, 5min timer will start anyway, was added in 671052575a36624b06379ee5e13a7fea26870194

src/server/game/Entities/Player/Player.cpp
```diff
@@ -21906,14 +21911,9 @@ void Player::UpdateHomebindTime(uint32 time)
 
 void Player::InitPvP()
 {
+    // pvp flag should stay after relog
     if (HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_IN_PVP))
         UpdatePvP(true, true);
-    else if (HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_PVP_TIMER))
```
Faq, line 21911: removing not working part, added in 671052575a36624b06379ee5e13a7fea26870194
Not working as bugged in client, mentioned by shauren

src/server/game/Entities/Player/Player.cpp
```diff
@@ -21968,7 +21968,7 @@ void Player::UpdatePvP(bool state, bool _override)
     }
     else
     {
-        pvpInfo.EndTimer = GameTime::GetGameTime() + 300;
```
Faq, line 21971: same

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -464,7 +464,7 @@ void WorldSession::HandleTogglePvP(WorldPackets::Misc::TogglePvP& togglePvP)
     else
     {
         if (!GetPlayer()->pvpInfo.IsHostile && GetPlayer()->IsPvP())
-            GetPlayer()->pvpInfo.EndTimer = GameTime::GetGameTime() + 300;     // start toggle-off
```
Faq, line 467: not needed, handled by client

Killyana:
The PR fix the issue.
But doesn't the UI must reflect the status of the PVP flag?
For example here, I disabled the pvp and after 5min It got off, then I attacked a player so PVP turned ON,  but the check mark is still in "Active"
image

Faq:
That kinda should be handled by client. Like this it was before all these commits.

jackpoz:
@Killyana "Deactivated" means the player wants it deactivated so it will expire by itself after 5 minutes.
Setting it to "Activated" means you always want to have PvP On, no matter how long ago you pvp'd last time.

Killyana:
In this case it's fine.

jackpoz:
Thanks for the PR :)

PR #24378, opened 2020-03-31
Scripts/Spells: prevent Drain Soul from cancelling prematurely

### LINKED ISSUE #24368: Core/Spells: Drain Soul cancels prematurely
**Description:**

Drain Soul cancels prematurely when another enemy dies.

**Current behaviour:**

Drain Soul will cancel if another enemy dies while it is channeling.

**Expected behaviour:**

Drain Soul should only be cancelled if the duration runs out or if the target dies.

1. Create a Warlock
2. Add a rank of Drain Soul (`.learn 1120`)
3. For convenience:
   * Learn a high-rank Corruption: `.learn 11672`
   * Learn Summon Voidwalker for a meatshield: `.learn 697`
   * Add Soul Shard for Voidwalker: `.additem 6265`
4. Enter combat with 2 experience-granting enemies, which we will call 'Creature A' and 'Creature B'.
5. Cast enough DoTs on Creature A so that it will die.
6. Before Creature A dies, cast Drain Soul on Creature B.
7. When Creature A dies, Drain Soul will be cancelled.
   * This happens with or without Improved Drain Soul
   * This happens with any rank of Drain Soul

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

7258d00f

**Operating system:**

Fedora 31

adeutscher:
**Changes proposed:**

-  Prevent Drain Soul from proccing early when the warlock deals a killing blow to a unit other than the victim of drain soul.

- [x] 3.3.5
- [ ] master

**Issues addressed:** Fixes #24368

**Tests performed:**

Two-target test:

1. Create a warlock.
2. Teach the warlock Drain Soul: `.learn 1120`
3. Give the warlock a high-rank of Corruption for convenient enemy killing: `.learn 11672`
4. Engage two experience-granting targets in combat.
5. Cast Corruption on target A.
6. Before target A dies, cast Drain Soul on target B.
7. When Target A dies, Drain Soul should still be casting on target B.
8. Kill Target A while channeling Drain Soul on it. A soul shard should be generated.

Continued proc test:

1. Level the same warlock to level 16: .level 15
2. Get Improved Drain Soul from the Affliction Tree
3. Have your "What happened to me?" view of the combat log ready.
4. Engage two experience-granting targets in combat.
5. Cast Corruption on target A.
6. Before target A dies, cast Drain Soul on target B.
7. When target A dies, observe that Drain Soul has not procced to give the caster mana back. As with the original test, Drain Soul should not be interrupted by the death of Target A.
8. Kill target B while channeling Drain Soul on it. A soul share should be generated, and the caster should receive mana back.

### DIFF
--- a/src/server/scripts/Spells/spell_warlock.cpp
+++ b/src/server/scripts/Spells/spell_warlock.cpp
@@ -516,6 +516,8 @@ class spell_warl_drain_soul : public SpellScriptLoader
 
             bool CheckProc(ProcEventInfo& eventInfo)
             {
+                // Drain Soul's proc tries to happen each time the warlock lands a killing blow on a unit while channeling.
+                // Make sure that dying unit is afflicted by the caster's Drain Soul debuff in order to avoid a false positive.
                 return eventInfo.GetProcTarget()->GetAuraApplicationOfRankedSpell(SPELL_WARLOCK_DRAIN_SOUL, GetCaster()->GetGUID()) != 0;
             }
 
--- a/src/server/scripts/Spells/spell_warlock.cpp
+++ b/src/server/scripts/Spells/spell_warlock.cpp
@@ -519,8 +519,8 @@ class spell_warl_drain_soul : public SpellScriptLoader
                 // Drain Soul's proc tries to happen each time the warlock lands a killing blow on a unit while channeling.
                 // Make sure that dying unit is afflicted by the caster's Drain Soul debuff in order to avoid a false positive.
 
-                Unit * caster = GetCaster();
-                Unit * victim = eventInfo.GetProcTarget();
+                Unit* caster = GetCaster();
+                Unit* victim = eventInfo.GetProcTarget();
 
                 if (caster && victim)
                     return victim->GetAuraApplicationOfRankedSpell(SPELL_WARLOCK_DRAIN_SOUL, caster->GetGUID()) != 0;
--- a/src/server/scripts/Spells/spell_warlock.cpp
+++ b/src/server/scripts/Spells/spell_warlock.cpp
@@ -518,7 +518,14 @@ class spell_warl_drain_soul : public SpellScriptLoader
             {
                 // Drain Soul's proc tries to happen each time the warlock lands a killing blow on a unit while channeling.
                 // Make sure that dying unit is afflicted by the caster's Drain Soul debuff in order to avoid a false positive.
-                return eventInfo.GetProcTarget()->GetAuraApplicationOfRankedSpell(SPELL_WARLOCK_DRAIN_SOUL, GetCaster()->GetGUID()) != 0;
+
+                Unit * caster = GetCaster();
+                Unit * victim = eventInfo.GetProcTarget();
+
+                if (caster && victim)
+                    return victim->GetAuraApplicationOfRankedSpell(SPELL_WARLOCK_DRAIN_SOUL, caster->GetGUID()) != 0;
+
+                return false;
             }
 
             void HandleProc(AuraEffect const* aurEff, ProcEventInfo& eventInfo)
--- a/src/server/scripts/Spells/spell_warlock.cpp
+++ b/src/server/scripts/Spells/spell_warlock.cpp
@@ -35,6 +35,7 @@
 
 enum WarlockSpells
 {
+    SPELL_WARLOCK_DRAIN_SOUL                        = 1120,
     SPELL_WARLOCK_CREATE_SOULSHARD                  = 43836,
     SPELL_WARLOCK_CURSE_OF_DOOM_EFFECT              = 18662,
     SPELL_WARLOCK_DEMONIC_CIRCLE_SUMMON             = 48018,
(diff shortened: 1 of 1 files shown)


src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -513,6 +514,13 @@ class spell_warl_drain_soul : public SpellScriptLoader
                 });
             }
 
+            bool CheckProc(ProcEventInfo& eventInfo)
+            {
+                // Drain Soul's proc tries to happen each time the warlock lands a killing blow on a unit while channeling.
+                // Make sure that dying unit is afflicted by the caster's Drain Soul debuff in order to avoid a false positive.
+                return eventInfo.GetProcTarget()->GetAuraApplicationOfRankedSpell(SPELL_WARLOCK_DRAIN_SOUL, GetCaster()->GetGUID()) != 0;
```
jackpoz: is there a chance for GetCaster() to return null at this point ?

adeutscher: I'm not entirely familiar with the scripting system, so I'm can't say it wouldn't with 100% certainty. I definitely got caught up in the idea of a smooth-looking one-liner.

I'll make the following adjustments:

* Switch from `GetCaster()` to `eventProc.GetActor()`.
* Store pointers for `GetProcTarget` and `GetActor` instead of the one-liner.
* Wrap the final `if` in a safety check for the above stored pointers.

src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -513,6 +514,20 @@ class spell_warl_drain_soul : public SpellScriptLoader
                 });
             }
 
+            bool CheckProc(ProcEventInfo& eventInfo)
+            {
+                // Drain Soul's proc tries to happen each time the warlock lands a killing blow on a unit while channeling.
+                // Make sure that dying unit is afflicted by the caster's Drain Soul debuff in order to avoid a false positive.
+
+                Unit * caster = GetCaster();
```
jackpoz: just minor change, we usually write
```cpp
Unit* caster =
```
with the * near the type

illfated:
Cosmetic PR title detail, it can be changed to reflect its location in the source:

**Scripts/Spells: prevent Drain Soul from cancelling prematurely**

(You are of course free to ignore this, I simply wanted to suggest this based on the conservative nomenclature in this repository for PR titles.)

src/server/scripts/Spells/spell_warlock.cpp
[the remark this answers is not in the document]
adeutscher: Confirmed, adjusted with 03607c2

Keader:
We are fixing a symptom, not a cause.

illfated:
> We are fixing a symptom, not a cause.

Fair point. Do you know what the underlying cause is, and whether it can be solved, if we know where to look?

Keader:
I'm out of TC for a while.
I'm installing everything here again, i will build tc tomorrow morning. I think that i have fix for it.
was something about one effect of drain soul be positive and proc early

Keader:
Wait... i pushed a fix for this months ago
43c08b176115505b76c027667047f9a5bfcc4e11

illfated:
@adeutscher : Have you noticed the content in the commit linked by Keader, from August 2019?

Keader:
hmm, looks like not same issue...
this issue looks new

adeutscher:
I think what happened was that the mechanics of how procs are handled has changed since 43c08b1.

If we've entered `spell_warl_drain_soul::HandleProc`, then it's already too late for the aura being procced. The aura will be removed afterwards, per [this](https://github.com/TrinityCore/TrinityCore/blob/abe4d59526fe0e7d5a96810d44c7d17fd0d0afbd/src/server/game/Spells/Auras/SpellAuras.cpp#L1843) check. This check and the part that removes the charges is from 6+ years ago, so this wouldn't have been a problem with Keader's commit.

What I noticed next is that my world database has an AttributeMask of 1, compared with Keader's 48. It sounds like this problem could have come up for me because my database missed the update bus. Keader's update currently lives in `sql/old/3.3.5a/world/19071_2019_08_16/2019_08_15_00_world.sql`. I'll load up a fresh world database tonight to confirm this.

For a related question, where in the C++ code is the enum that aligns with the AttributeMask values on the `spell_proc` table?

illfated:
I don't know, but have you looked through the lists in

https://github.com/TrinityCore/TrinityCore/blob/e61c54dce85a9df6b6fefd722533d29466269b68/src/server/shared/SharedDefines.h#L397-L695

to see if there is anything useful in that file?

Shauren:
Its these
https://github.com/TrinityCore/TrinityCore/blob/e61c54dce85a9df6b6fefd722533d29466269b68/src/server/game/Spells/SpellMgr.h#L239-L247

adeutscher:
Thanks for the nudge. SpellAttr6's 'SPELL_ATTR6_DONT_CONSUME_PROC_CHARGES' aligns with the 0x30/48 mask.

I won't be able to check myself until this evening Pacific time, but what does your world database have to say about the following query:

```sql
SELECT AttributesMask FROM spell_proc WHERE SpellId = -1120;
```

Keader:
ty

adeutscher:
@Shauren `ProcAttributes` makes a lot more sense than our SharedDefines dive. Thanks!

PR #24379, opened 2020-04-01
Core/Movement: Formation Rewrite

### LINKED ISSUE #15792: [DEV] Core: Creature formation rewrite
Taken from an irc conversation with Malcrom

Formations npcs should have destination updated every 1.6 sec.
They have to move to a position at the described angle of the leader.
Atm they are told to move to a position near the next waypoint on offi they are not.

Here is a pathed creature http://i.imgur.com/BU65oKX.png
Here is a follower http://i.imgur.com/93qfHPC.png

ATM the follower is told to move next to the leaders next waypoint, This is wrong.
Why would there be 50 waypoints to 14 on offi if that was correct.
Also you see a follower usually has turn points jutting off because it is moving to where it thinks it should be if the leader stays moving the same direction.

So what needs to be done is first have the followers destination point updated every 1.6 sec,
then replace the calculation with the proper one to determine movepoint and any speed adjustment required due to range.
Then your follower is updated every 1.6 sec rather than heading to what it thinks is the waypoint and not being corrected until it reaches it.
Nothing has to change with the db table

A creature should be able to exist in water only with gravity disabled.
So you should be able to set it in creature_template .
But it should not be used to determine if a creature is flying, it is being used that way in tc and cause issues.

### LINKED ISSUE #22919: Core/Movements: Incorrect formation behavior
**Description:** For formations when the follower has a lower speed than the leader, once it gets  too far it will run to the next waypoint and reach it before the leader and wait for him.

This is definitively not the correct behavior, probably the follower must use the same speed as the leader and if he gets too far (for some reason) he must run to the contact point with the leader+ the follow distance, and not to the next waypoint.

1. UPDATE creature_template SET speed_walk=0.5 WHERE entry IN (5701);
2. .go c id 5701
3. the npc 5701 (Selina Pickman) is following the leader, just notice the behavior
You can also use ".wp show on"  on the leader to see the waypoints

**Branch(es):** both probably

**TC rev. hash/commit:**  rev. dd81978f432f

Killyana:
**Changes proposed:**

-  Rewrite formation

- [x] 3.3.5
- [ ] master

**Issues addressed:**
Closes #15792 
Closes #22919

**Tests performed:** It compiles and works fine IG

- Review [ ]

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -338,7 +338,7 @@ bool Creature::IsFormationLeader() const
     return m_formation->IsLeader(this);
 }
 
-void Creature::SignalFormationMovement(Position const& destination, uint32 id/* = 0*/, uint32 moveType/* = 0*/, bool orientation/* = false*/)
+void Creature::SignalFormationMovement()
 {
     if (!m_formation)
         return;
@@ -346,7 +346,7 @@ void Creature::SignalFormationMovement(Position const& destination, uint32 id/* 
     if (!m_formation->IsLeader(this))
         return;
 
-    m_formation->LeaderMoveTo(destination, id, moveType, orientation);
+    m_formation->LeaderStartedMoving();
 }
 
 bool Creature::IsFormationLeaderMoveAllowed() const
--- a/src/server/game/Entities/Creature/Creature.h
+++ b/src/server/game/Entities/Creature/Creature.h
@@ -315,7 +315,7 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         CreatureGroup* GetFormation() { return m_formation; }
         void SetFormation(CreatureGroup* formation) { m_formation = formation; }
         bool IsFormationLeader() const;
-        void SignalFormationMovement(Position const& destination, uint32 id = 0, uint32 moveType = 0, bool orientation = false);
+        void SignalFormationMovement();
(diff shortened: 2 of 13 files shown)


jackpoz:
>/home/circleci/project/src/server/game/Entities/Creature/CreatureGroups.cpp:293:120: fatal error: unused parameter 'movement' [-Wunused-parameter]
        MovementGenerator const* moveGen = member->GetMotionMaster()->GetMovementGenerator([](MovementGenerator const* movement)->bool { return true; }, MOTION_SLOT_DEFAULT);

ckegg:
Just a quick thought about formation and aggros

Taking example of Garr in Molten Core; 8 firesworns linked to Garr with formation.
1. Have one person pull one of them, yes they all come.
2. Get another person do damage on one of them before they reach to the first person
3. Mobs reach to first person and killed him
4. Mobs return to home except the one has threat with second person
5 repeat above until only Garr left and start solo it

Is it possible keep all mobs linked in formation also links on aggro?

offl:
Combat pulse is not working(?) + killed NPC must respawn(not always, some trash mobs mustn't respawn)(see https://github.com/cmangos/issues/wiki/creature_linking)

Ovahlord:
Outside of the scope of this PR. This part is movement and not combat related.

ckegg:
> 
> 
> Outside of the scope of this PR. This part is movement and not combat related.

Totally understood just bring it up and people getting aware of it

jackpoz:
Thanks for the PR :)

PR #24387, opened 2020-04-02
Core/Battlegrounds: Use fields from DB2/battleground_template

Golrag:
We can use the following fields from DB2/battleground_template values in the Battleground class: 
- isArena
- name
- queueId
- min & max players (per team)
- map id
- script id
- start max dist
- type id
- min & max level

**Changes proposed:**

-  Drop fields from the Battleground class that are present in the BattlegroundTemplate
-  Drop fields from battleground_template that are set in the BattlemasterListEntry

- master

**Issues addressed:** Closes #  none

- Builds
- Server starts up
- Not loading invalid battleground templates (example: invalid start location)

- [x] Issues with .reload battleground_template

Any feedback & help is welcome :)

### DIFF
--- a/src/server/game/DataStores/DB2Stores.cpp
+++ b/src/server/game/DataStores/DB2Stores.cpp
@@ -837,6 +837,20 @@ void DB2Manager::LoadStores(std::string const& dataPath, uint32 defaultLocale)
     for (AzeriteUnlockMappingEntry const* azeriteUnlockMapping : sAzeriteUnlockMappingStore)
         azeriteUnlockMappings[azeriteUnlockMapping->AzeriteUnlockMappingSetID].push_back(azeriteUnlockMapping);
 
+    for (BattlemasterListEntry const* battlemaster : sBattlemasterListStore)
+    {
+        if (battlemaster->MaxLevel < battlemaster->MinLevel)
+        {
+            TC_LOG_ERROR("db2.hotfix.battlemaster_list", "Battlemaster (%u) contains bad values for MinLevel (%u) and MaxLevel (%u). Swapping values.", battlemaster->ID, battlemaster->MinLevel, battlemaster->MaxLevel);
+            std::swap(*(int8*)&battlemaster->MaxLevel, *(int8*)&battlemaster->MinLevel);
+        }
+        if (battlemaster->MaxPlayers < battlemaster->MinPlayers)
+        {
+            TC_LOG_ERROR("db2.hotfix.battlemaster_list", "Battlemaster (%u) contains bad values for MinPlayers (%u) and MaxPlayers (%u). Swapping values.", battlemaster->ID, battlemaster->MinPlayers, battlemaster->MaxPlayers);
+            std::swap(*(int8*)&battlemaster->MaxPlayers, *(int8*)&battlemaster->MinPlayers);
+        }
+    }
+
     ASSERT(BATTLE_PET_SPECIES_MAX_ID >= sBattlePetSpeciesStore.GetNumRows(),
         "BATTLE_PET_SPECIES_MAX_ID (%d) must be equal to or greater than %u", BATTLE_PET_SPECIES_MAX_ID, sBattlePetSpeciesStore.GetNumRows());
 
--- a/sql/updates/world/master/2020_05_02_00_world.sql
+++ b/sql/updates/world/master/2020_05_02_00_world.sql
@@ -1,0 +1,5 @@
+ALTER TABLE `battleground_template`
+   DROP COLUMN `MinPlayersPerTeam`,
+   DROP COLUMN `MaxPlayersPerTeam`,
+   DROP COLUMN `MinLvl`,
+   DROP COLUMN `MaxLvl`;
--- a/src/server/game/Battlegrounds/Arena.cpp
+++ b/src/server/game/Battlegrounds/Arena.cpp
@@ -27,7 +27,7 @@
 #include "WorldStatePackets.h"
 #include "WorldSession.h"
 
-Arena::Arena()
+Arena::Arena(BattlegroundTemplate const* battlegroundTemplate) : Battleground(battlegroundTemplate)
 {
     StartDelayTimes[BG_STARTING_EVENT_FIRST]  = BG_START_DELAY_1M;
     StartDelayTimes[BG_STARTING_EVENT_SECOND] = BG_START_DELAY_30S;
--- a/src/server/game/Battlegrounds/Arena.h
+++ b/src/server/game/Battlegrounds/Arena.h
@@ -48,7 +48,7 @@ enum ArenaWorldStates
 class TC_GAME_API Arena : public Battleground
 {
     protected:
-        Arena();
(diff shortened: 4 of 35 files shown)


Golrag:
updated to https://github.com/TrinityCore/TrinityCore/commit/e5b12a76dcd3348ff5ddc3e79da4d4a9f6837351

sql/updates/world/master/yyyy_mm_dd_xx_world.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE `battleground_template`
+   DROP COLUMN `MinPlayersPerTeam`,
```
Aokromes, line 2: imho no, better allow small servers to customize requirements.

Golrag: That should still be doable with hotfixes

Golrag:
```sql
DELETE FROM `battlemaster_list` WHERE `ID` = 7;
INSERT INTO `battlemaster_list` (`ID`, `Name`, `GameType`, `ShortDescription`, `LongDescription`, `InstanceType`, `MinLevel`, `MaxLevel`, `RatedPlayers`, `MinPlayers`, `MaxPlayers`, `GroupsAllowed`, `MaxGroupSize`, `HolidayWorldState`, `Flags`, `IconFileDataID`, `RequiredPlayerConditionID`, `MapID1`, `MapID2`, `MapID3`, `MapID4`, `MapID5`, `MapID6`, `MapID7`, `MapID8`, `MapID9`, `MapID10`, `MapID11`, `MapID12`, `MapID13`, `MapID14`, `MapID15`, `MapID16`, `VerifiedBuild`) VALUES
(7, 'Eye of the Storm', 'Domination', NULL, NULL, 3, 15, 120, 15, 5, 10, 1, 5, 2851, 24, 525912, 0, 566, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 32638);

DELETE FROM `hotfix_data` WHERE `Id` = 1 AND `TableHash` = 1435332686 AND `RecordId` = 7;
INSERT INTO `hotfix_data` (`Id`, `TableHash`, `RecordId`, `Deleted`) VALUES (1, 1435332686, 7, 0);
```

Will result in an Eye of the storm with max 10 players per team & min 5 players per team

Golrag: With hotfixes the client will also display the correct value
image

Aokromes: then nothing to say :)

Golrag:
I've some more changes in mind for battlegrounds (other PR's will be made after this one). No content changes tho, as that would require lots of sniffs and i only have some legion ones

Golrag:
update to https://github.com/TrinityCore/TrinityCore/commit/fb0a29131d30e15d8d323271490069860580817d and restore some checks:
- Max level must be equal or greater than Min level
- Max players must be equal or greater than Min players

Golrag:
Could I get some feedback/review ? :p

src/server/game/DataStores/DB2Stores.cpp
```diff
@@ -837,6 +837,12 @@ void DB2Manager::LoadStores(std::string const& dataPath, uint32 defaultLocale)
     for (AzeriteUnlockMappingEntry const* azeriteUnlockMapping : sAzeriteUnlockMappingStore)
         azeriteUnlockMappings[azeriteUnlockMapping->AzeriteUnlockMappingSetID].push_back(azeriteUnlockMapping);
 
+    for (BattlemasterListEntry const* battlemaster : sBattlemasterListStore)
+    {
+        ASSERT(battlemaster->MaxLevel >= battlemaster->MinLevel, "Battlemaster (%u) contains bad values for MinLevel (%u) and MaxLevel (%u)", battlemaster->ID, battlemaster->MinLevel, battlemaster->MaxLevel);
+        ASSERT(battlemaster->MaxPlayers >= battlemaster->MinPlayers, "Battlemaster (%u) contains bad values for MinPlayers (%u) and MaxPlayers (%u)", battlemaster->ID, battlemaster->MinPlayers, battlemaster->MaxPlayers);
```
Shauren, line 843: never assert on db2 data like this, the only time it needs to be done if some compile time constant depends on it like array sizes, this case here should be handled like sGameObjectDisplayInfoStore

Shauren (changes requested):

src/server/game/DataStores/DB2Stores.cpp
Golrag: Done! :)

src/server/game/DataStores/DB2Stores.cpp
```diff
@@ -837,6 +837,20 @@ void DB2Manager::LoadStores(std::string const& dataPath, uint32 defaultLocale)
     for (AzeriteUnlockMappingEntry const* azeriteUnlockMapping : sAzeriteUnlockMappingStore)
         azeriteUnlockMappings[azeriteUnlockMapping->AzeriteUnlockMappingSetID].push_back(azeriteUnlockMapping);
 
+    for (BattlemasterListEntry const* battlemaster : sBattlemasterListStore)
+    {
+        if (battlemaster->MaxLevel < battlemaster->MinLevel)
+        {
+            TC_LOG_ERROR("sql.sql", "Battlemaster (%u) contains bad values for MinLevel (%u) and MaxLevel (%u). Swapping values.", battlemaster->ID, battlemaster->MinLevel, battlemaster->MaxLevel);
```
Shauren, line 844: Maybe pick a different logger category for this?

Golrag:
I kinda forgot about .reload battleground_template this command might break some things now.

src/server/game/DataStores/DB2Stores.cpp
Golrag: updated! Using hotfixes.battlemasterlist now.

Golrag:
So there are a couple of things we can do to support reloading battleground_templates at runtime:

- Let the Battleground class have a copy of the BattlegroundTemplate. This will have another issue tho as new battlegrounds are copied from one base battleground instance. So they will never receive the updated battleground_template.
- Not clearing the _battlegroundTemplates container in BattlegroundMgr when loading battlegrounds (while also avoiding to erase them when new bad data is loaded). You won't be able to delete battleground_templates at runtime anymore (kinda like creature_template).
- Revert the changes made in Battleground class. We can keep the changes made to the battleground_template table.
- ... other suggestions

Shauren:
I like not being able to delete

Golrag:
Should be working fine now.
changing a record to use an invalid WorldSafeLocation then reloading the table -> Message saying the change will be ignored.
Starting the server with invalid WorldSafeLocation -> battleground_template not loaded.
Fixing that with a valid WorldSafeLocation -> will load the battleground_template.

Inserting a new battleground_template that does not have a Battleground class will never be loaded.

PR #24388, opened 2020-04-02
CI/Circle CI: Prevent conflict with diff branches

funjoker:
**Changes proposed:**

-  reanmed keys for cache

- [x] 3.3.5
- [ ] master

- [ ] 
- [ ]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,9 +56,9 @@ jobs:
             /scripts/checkout.sh
       - restore_cache:
           keys:
-            - nopch-{{ .Branch }}-{{ .Revision }}
-            - nopch-{{ .Branch }}
-            - nopch
+            - 3.3.5-nopch-{{ .Branch }}-{{ .Revision }}
+            - 3.3.5-nopch-{{ .Branch }}
+            - 3.3.5-nopch
       - run:
           name: Setup
           command: |
@@ -85,7 +85,7 @@ jobs:
             ./authserver --version
             ./worldserver --version
       - save_cache:
-          key: nopch-{{ .Branch }}-{{ .Revision }}
+          key: 3.3.5-nopch-{{ .Branch }}-{{ .Revision }}
           paths:
             - /home/circleci/.ccache
 workflows:


.circleci/config.yml
```diff
@@ -56,9 +56,9 @@ jobs:
             /scripts/checkout.sh
       - restore_cache:
           keys:
-            - nopch-{{ .Branch }}-{{ .Revision }}
-            - nopch-{{ .Branch }}
-            - nopch
+            - 3.3.5-nopch-{{ .Branch }}-{{ .Revision }}
```
Shauren, line 59: isnt having branch twice kind of redundant?

Shauren: But thanks to that last component (revision) it doesn't matter because a commit will never have the same hash on 3.3.5 and master, to achieve that you would need to run `git merge` (cherry-picks create new hash)

funjoker: For that revision thing I recommend to ask @jackpoz

funjoker: it works like a fallback methodic

jackpoz: @Shauren take for example https://circleci.com/gh/TrinityCore/TrinityCore/6329?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

>No cache is found for key: nopch-pull/24385-e0144f7af251631b0e61e4446bb072edbdb22e14
No cache is found for key: nopch-pull/24385
Found a cache from build 6326 at nopch

1. First nopch-{{ .Branch }}-{{ .Revision }} is checked. For a new PR, there is no cache yet so this is empty
2. Then nopch-{{ .Branch }} is checked. For a new PR, there is no cache yet so this is empty too.
3. Then nopch is checked. That exists as it's updated with every commit (it's shared across the whole Circle CI)

So the "nopch" cache is used by any new branch and in particular any new PR. After the first successful build, the PR will use the nopch-{{ .Branch }} cache.

Shauren: Ok, then doesn't that make the `-{{ .Revision }}` part useless (as in, it would never be pulled)? We are never building the same commit multiple times (except manual triggers if build fails for some random unrelated reason, in which case the cache wouldn't have been stored)

jackpoz: That's pretty much how it works. The revision bit came from https://circleci.com/docs/2.0/caching/#source-caching .

Only 1 cache is saved anyway
>Creating cache archive...
Uploading cache archive...
Stored Cache to nopch-pull/24385-e0144f7af251631b0e61e4446bb072edbdb22e14
/home/circleci/.ccache

so having more keys doesn't really have any bad effect

jackpoz:
Thanks for the PR :)

PR #24399, opened 2020-04-06
Core/Objects: reduce unnecessary vmap lookups by using cached liquid status data pulled from position updates

Ovahlord:
**Changes proposed:**

-  instead of launching a seperate vmap check for liquid status checks such as being in water or underwater, we will now pull our data from our already done vmap lookup. 
-  dropped the already redundant player override for underwater checks and use the default helper. It already does consider collision heights so now we get accurate underwater height results such as gnomes being underwater earlier than taurens for example.

- [x] 3.3.5
- [ ] master

- tested ingame, results are fine

Hi Trokli. I know you're gonna read this :)


illfated:
Oh man. I feel really predictable now. 😉  😜
Wish I could be of help somehow, but I don't have the coding skills.

PR #24404, opened 2020-04-07
Core/Spells: implement SMSG_MOUNT_RESULT and use it for transformed mounting cases

Ovahlord:
**Changes proposed:**

-  implement mount result opcode handling which got introduced with 4.x and use it in the currently only known sniffable case when trying to mount up while being transformed

- [ ] 3.3.5
- [x] master

- tested on my 434 fork
WoWScrnShot_040720_184548


src/server/game/Miscellaneous/SharedDefines.h
```diff
@@ -6508,4 +6508,19 @@ enum class GameError : uint32
     ERR_NOT_IN_PET_BATTLE                                           = 1047,
 };
 
+enum class MountResult : uint32
+{
+    InvalidMountee = 0,
+    TooFarAway     = 1,
+    AlreadyMounted = 2,
+    NotMountable   = 3,
+    NotYourPet     = 4,
```
Faq, line 6517: In what cases this used?

src/server/game/Miscellaneous/SharedDefines.h
```diff
@@ -6508,4 +6508,19 @@ enum class GameError : uint32
...
+    Other          = 5,
+    Looting        = 6,
```
Faq, line 6519: How can Mount result be " Looting" ?

Ovahlord: These values are from wpp. These values are not only for smsg_mount_result but also for smsg_dismount_result so some of these values are used for the other opcode as well which I have not discovered in sniffs yet.

ForesterDev:
Why in 4.x? It's present in 3.3.5 - SMSG_MOUNTRESULT

Faq:
@Ovahlord Can you make PR also for 335?

src/server/game/Miscellaneous/SharedDefines.h
illfated: Maybe parked mounts without vehicle access?

src/server/game/Miscellaneous/SharedDefines.h
Shauren: Values are correct, comment is wrong. They are used exclusively for SMSG_MOUNT_RESULT.

PR #24417, opened 2020-04-09
Core/Gameobjects: Improve stealth detection

### LINKED ISSUE #24224: [3.3.5] Gameobjects and stealth interaction
**Description:**

Walking over a gameobject(for example BG buffs, or hunter traps) in stealth triggers the object, so it dissapears but the player receives no effect from GO trigger. Unfortunately #23936 didn't solve this issue.

**Expected behaviour:**

GO should have standart behavior when player triggers them from stealth.

1. Walk over BG buff, or hunter trap in stealth
2. GO triggers, but player receive no effect

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

ce449f6b5332ab466d935de0077bccfdde716d1b

Jildor:
**Changes proposed:**

-  all explained here: https://github.com/TrinityCore/TrinityCore/issues/24224

@joshwhedon quotes
>The main problem with the go detection is that IsValidAttackTarget doesn't only check the visibility, it uses the trap owner to check all the target validity (there is no reference to the trap itself at the moment of the check).
At first glance, it should be possible to change NearestAttackableNoTotemUnitInObjectRangeCheck to take the gameobject as parameter, and to tweak IsValidAttackTarget to handle gameobjects better (for example when looking up the affected player, need to take the gameobject case into account)

>Oh it looks like ReputationRank WorldObject::GetReactionTo(WorldObject const* target) const does not handle gameobjects very well.

- Skip gobs check HasInArc in WorldObject::CanDetectStealthOf
- Traps should detect stealth always

Author: @joshwhedon

- [X] 3.3.5
- [ ] master

**Issues addressed:** Closes #24224

**To-Do:**
~~Maybe we should remove unused i_funit in NearestAttackableNoTotemUnitInObjectRangeCheck (only used for hunter traps)~~

### DIFF
--- a/src/server/game/Grids/Notifiers/GridNotifiers.h
+++ b/src/server/game/Grids/Notifiers/GridNotifiers.h
@@ -887,7 +887,7 @@ namespace Trinity
     class NearestAttackableNoTotemUnitInObjectRangeCheck
     {
         public:
-            NearestAttackableNoTotemUnitInObjectRangeCheck(WorldObject const* obj, Unit const* funit, float range) : i_obj(obj), i_funit(funit), i_range(range) { }
+            NearestAttackableNoTotemUnitInObjectRangeCheck(WorldObject const* obj, float range) : i_obj(obj), i_range(range) { }
 
             bool operator()(Unit* u)
             {
@@ -912,7 +912,6 @@ namespace Trinity
 
         private:
             WorldObject const* i_obj;
-            Unit const* i_funit;
             float i_range;
     };
 
--- a/src/server/game/Grids/Notifiers/GridNotifiers.h
+++ b/src/server/game/Grids/Notifiers/GridNotifiers.h
@@ -903,7 +903,7 @@ namespace Trinity
                 if (!u->isTargetableForAttack(false))
                     return false;
 
-                if (!i_obj->IsWithinDistInMap(u, i_range) || !i_funit->IsValidAttackTarget(u))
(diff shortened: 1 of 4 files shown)


src/server/game/Grids/Notifiers/GridNotifiers.h
```diff
@@ -903,7 +903,7 @@ namespace Trinity
                 if (!u->isTargetableForAttack(false))
                     return false;
 
-                if (!i_obj->IsWithinDistInMap(u, i_range) || !i_funit->IsValidAttackTarget(u))
```
Faq, line 906: fatal error: private field 'i_funit' is not used

src/server/game/Spells/Spell.cpp
```diff
@@ -5186,7 +5186,7 @@ SpellCastResult Spell::CheckCast(bool strict, uint32* param1 /*= nullptr*/, uint
     {
         // Check explicit target for m_originalCaster - todo: get rid of such workarounds
         WorldObject* caster = m_caster;
-        if (m_originalCaster)
+        if (m_originalCaster && !caster->ToGameObject())
```
jackpoz, line 5189: could you please add a code comment here to explain why the check is needed and what it does ?

jackpoz (changes requested):

jackpoz:
I change a bit the code style, it should not have broken anything.

Jildor:
sure, feel free to modify whatever you want, thx 👍

jackpoz (approved):

PR #24440, opened 2020-04-11
Core/Packets: updated SMSG_LEVELUP_INFO

ForesterDev:
(cherry picked from commit bf9024c34180b1e35afb3cc9ef4cf1f3133dedfd)

**Changes proposed:**

-  update SMSG_LEVELUP_INFO to new packet class

- [x] 3.3.5

Build

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -2593,23 +2593,23 @@ void Player::GiveLevel(uint8 level)
     PlayerClassLevelInfo classInfo;
     sObjectMgr->GetPlayerClassLevelInfo(GetClass(), level, &classInfo);
 
-    // send levelup info to client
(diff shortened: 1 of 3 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -2593,23 +2593,23 @@ void Player::GiveLevel(uint8 level)
     PlayerClassLevelInfo classInfo;
     sObjectMgr->GetPlayerClassLevelInfo(GetClass(), level, &classInfo);
 
-    // send levelup info to client
```
Faq, line 2596: Would be good to restore comments

ForesterDev: I would prefer to move the packet to a separate method

Ovahlord: Would make sense if it would be used outside of this method but it's not and also this is primarily a conversion, not a refactor

jackpoz:
Thanks for the PR :)

Issue #22824, opened 2018-11-22
[3.3.5] Onyxia spells failing, Eruption missing, Deep Breath not casting


wetbrownsauce:
Hi, sorry if this report is unwanted, I'm trying to figure all this stuff out still. 

**Description:**
Onyxia has many bugs during the encounter.

**Current behaviour:**
Phase 1:
-Onyxia's Flame Breath and Wing Buffet suffer push back while she is casting them if a player is damaging her.
-Flame Breath and Wing Buffet will sometimes be interrupted by her own auto-attacks.

Phase 2:
-Onyxia will sometimes bug out when using her Deep Breath. The announcement of the ability will still happen on screen, but she will not cast it and then completely bugs out her movement, never flying to a new location until phase 3 when she lands. 
[Gif of the above happening](https://giant.gfycat.com/EasygoingWeeklyArawana.webm)
I'm not too familiar with trinitycore's code, but this could perhaps be a result of the following:

```
void SetNextRandomPoint()
        {
            uint8 MaxCount = sizeof(MoveData) / sizeof(OnyxMove);

            uint8 iTemp = urand(0, MaxCount - 1);

            if (iTemp >= MovePoint)
                ++iTemp;

            MovePoint = iTemp;
        }
```

Unless I'm misunderstanding this, it seems that if iTemp was randomized to 7, and Onyxia's previous MovePoint ended on 7 via `MovePoint = PointData->LocIdEnd;`, then iTemp would be incremented to 8 resulting in a MovePoint of 8 which then returns null later in the code when called by `GetMoveData()`

Please ignore that rambling if I'm entirely wrong, I'm pretty new to this. 

-When landing after phase 2, Onyxia floats until kited around a bit.
[Gif of the above happening](https://giant.gfycat.com/WarlikeDistantBarracuda.webm)

Phase 3:
-The Eruption of lava from cracks in the floor does no damage. In the combat log it repeatedly says "Lava Fissure's Eruption misses Onyxia" and once Onyxia is dead, it begins saying this about the player. 
[Image of combat log in phase 3](https://i.imgur.com/qJMNR5Y.png)
-Flame Breath, Wing Buffet, AND Bellowing Roar all suffer the same push back and interrupts during this phase.
[Gif of the above happening, this occurs in both phase 1 & 3](https://giant.gfycat.com/KlutzyScentedBoa.webm) 
-Whelps do not spawn in phase 3 at all, even though they should. (I think this is already known though).

[...]

PR #24472, opened 2020-04-18
Scripts/OnyxiasLair: Fix Onyxia Eruption spell

Jildor:
**Changes proposed:**

-  Set boss as originalCaster for proper spellcast
-  Ternary Operator for spell difficulty isn't needed, spell have spell difficulty and works properly

For some reason, spell hit onyxia too, but this is another issue, explained here: #22824

- [X] 3.3.5
- [ ] master

**Issues addressed:** Update #22824 (don't close)

### DIFF
--- a/src/server/scripts/Kalimdor/OnyxiasLair/instance_onyxias_lair.cpp
+++ b/src/server/scripts/Kalimdor/OnyxiasLair/instance_onyxias_lair.cpp
@@ -111,7 +111,9 @@ public:
                 //THIS GOB IS A TRAP - What shall i do? =(
                 //Cast it spell? Copyed Heigan method
                 floorEruption->SendCustomAnim(floorEruption->GetGoAnimProgress());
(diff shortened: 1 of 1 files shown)


jackpoz:
Should we wait for the issue of the spell hitting Onyxia to be fixed too ? Or will that be a sql only fix anyway ?

Jildor:
I think can be sql fix that, but I think is better do it separately, because I don't know what can be the proper fix for that

Rushor:
you can add a spellscript to not hit onyxia

Killyana:
No it will be a hack, there's already an issue about gob no handling factions correctly.

Rushor:
well then just let's merge this and wait for the otehr gameobject fix

jackpoz:
anyone would like to start working on the "other gameobject fix" while at it ? (for another PR ofc)

jackpoz:
ah well, let's hope someone will work on the other issue too

jackpoz:
Thanks for the PR :)

PR #24407, opened 2020-04-07
Core/PacketIO: Updated CMSG_TOGGLE_PVP opcode

Faq:
**Changes proposed:**

-  Cherry-picked from adabac9eddc5e344640ea1ed9c6c6e5c82a1de5f and 455959c

**Target branch(es):** 3.3.5/master

- [x] 3.3.5

**Tests performed:** (Does it build, tested in-game, etc.)
Compiles

**Known issues and TODO list:** (add/remove lines as needed)

PR #24476, opened 2020-04-19
Core/PacketIO: updated CMSG_TOGGLE_PVP opcode

ForesterDev:
**Changes proposed:**

-  update CMSG_TOGGLE_PVP  to new packet class
-  It can't be cherry-picket from master branch beacouse of difference between 3.3.5 and bfa. Look at https://github.com/TrinityCore/TrinityCore/pull/24407#issuecomment-610909699

- [x] 3.3.5

Build, tested in game with api functions TooglePvP() and SetPVP(0) / SetPvP(1) and UI functions


Aokromes:
supersedes https://github.com/TrinityCore/TrinityCore/pull/24407

src/server/game/Server/Packets/MiscPackets.h
```diff
@@ -285,6 +285,18 @@ namespace WorldPackets
...
         };
 
+        class TogglePvP final : public ClientPacket
+        {
+        public:
+            TogglePvP(WorldPacket&& packet) : ClientPacket(CMSG_TOGGLE_PVP, std::move(packet)) { }
+
+            void Read() override;
+
+            bool HasPvPStatus() const { return GetSize() == 1; }
```
ForesterDev, line 295: Don't like this way to check, if packet has data. I think we need generic way for clientpacket class

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -442,15 +442,21 @@ void WorldSession::HandleLogoutCancelOpcode(WorldPacket& /*recvData*/)
...
-void WorldSession::HandleTogglePvP(WorldPacket& recvData)
+void WorldSession::HandleTogglePvP(WorldPackets::Misc::TogglePvP& togglePvP)
 {
     // this opcode can be used in two ways: Either set explicit new status or toggle old status
-    if (recvData.size() == 1)
+    if (togglePvP.HasPvPStatus())
     {
-        bool newPvPStatus;
-        recvData >> newPvPStatus;
-        GetPlayer()->ApplyModFlag(PLAYER_FLAGS, PLAYER_FLAGS_IN_PVP, newPvPStatus);
```
jackpoz, line 452: Is ApplyModFlag() not working with the new packet structure ?

src/server/game/Server/Packets/MiscPackets.h
jackpoz: Enable could be Optional<bool>

src/server/game/Handlers/MiscHandler.cpp
Shauren: It is, it was changed in master for a different reason (incompatible with 8.1+ updatefield format) and since this is a port from master branch, it has the `if ()` separation too

src/server/game/Server/Packets/MiscPackets.h
Shauren: I agree with Optional but not with operator>> suggestion, it is better to leave initializing it decoupled with reading the field

src/server/game/Handlers/MiscHandler.cpp
ForesterDev: I changed this because in my opinion logic has become easier to understand than before

jackpoz: I think I would prefer to keep ApplyModFlag as a helper for setting/removing flags based on bool variable.
Maybe it's not so easy to understand because it's not used that much around the code.

src/server/game/Server/Packets/MiscPackets.h
jackpoz: ><%Shauren> its probably easier to just slap a getsize check on it with current state of bytebuffer :D
<%jackpoz> what about a CanRead<T>() ?
return _rpos + sizeof(T) <= size();
<%Shauren> if you want to add it, make another overload that accepts (size_t pos) too for parity with read overloads
but sounds good

ForesterDev: I think, it should be done outside the pr, and after I will update this PR using Optional for Enable flag

jackpoz:
Thanks for the PR :)

PR #24487, opened 2020-04-22
Core/Database: fix wrong password logging

### LINKED ISSUE #24486: SQL Error
**Description:**

When a player tries to login with an incorrect password it normally logs it to the logs_ip_actions table in the authserver databse (if enabled), but since Pull Request #24453 was implemented, the query is trying to insert the ip into the realm_id column which is resulting in this SQL error:

SQL(p): INSERT INTO logs_ip_actions (account_id, character_guid, realm_id, type, ip, systemnote, unixtime, time) VALUES (2, 0, '127.0.0.1', 1, 'Login to WoW Failed - Incorrect Password', 0, unix_timestamp(NOW()), NOW())
 [ERROR]: [1265] Data truncated for column 'realm_id' at row 1
Unhandled MySQL errno 1265. Unexpected behaviour possible.

This is also causing the entry to not be created in the DB, It seems to only be affecting wrong password logging since account logins, char logins/logouts seem to be logging just fine with no error.

**Current behaviour:**

What I put above ^

**Expected behaviour:**

It should insert data into the correct columns in logs_ip_actions

1. Enable wrong password logging in authserver.conf
2. Try logging in with an incorrect password
3. Check authserver logs and you should see the error

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. 740afc71358c 2020-04-21 22:22:31 +0200 (3.3.5 branch) (Win64, RelWithDebInfo, Static) (authserver)

**Operating system:**

Windows Server 2016 Standard

ForesterDev:
**Changes proposed:**

-  fix format for prepared statement

- [x] 3.3.5

**Issues addressed:** Closes #24486  (insert issue tracker number)

No, but I hope it works fine :)


Faq:
Thanx for the PR :)

LordWaycrest:
Thank you for the Pull Request. :-)

Issue #24490, opened 2020-04-23
Possible to move while casting 


HelloKitty:
**Description:**

It's possible to cast a spell and then move as a player with an unmodified client. This can be abused without external tools but can be significantly more effective with the right tools.

**Current behaviour:**

Currently, in Spell::Update a check for canceling a spell's cast is done by checking the current (aka LAST handled) MovementInfo from the last movement packet. It is possible with this logic for a client to turn its mouse to send various non-movement MSG_MOVE_* packets after moving such that the next Player::Update and Spell::Update will appear as though the player isn't moving yet did move since the last update. This is reproducible by hand within the client.

**Expected behaviour:**

Spell::Update should interrupt spells if a player has moved since the last update and not only if the current Player's MovementInfo appears to be moving.

**Steps to reproduce the problem:**

1. Start casting a spell
2. Move while casting, strafing works well.
3. Then immediately turn the rotation of the player to send a movement packet that appears not to be moving.

**Branch(es):**

3.3.5 and maybe master

**TC rev. hash/commit:** 

c2dd8d02388cc1e297b648099ab824c988a3ba03

**Operating system:** Windows

HelloKitty:
Having trouble replicating this on TC head. Was working on TC fork from 2 months ago. Will investigate again when I have more time.

Killyana:
I will close this issue as impossible to reproduce

PR #24491, opened 2020-04-23
Core/Spell: Fix stutter step cast while moving exploit

HelloKitty:
**Changes proposed:**

-  Aggregate movement flags sent in movement packets in currently casted spells.
-  Check aggregated movement flags in Spell update and not rely exclusively on the snapshot of movement info in IsMoving() check

- [x] 3.3.5
- [ ] master

https://github.com/TrinityCore/TrinityCore/issues/24490

Tested locally, it fixes the ability to stutter step cast.

None

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -26895,3 +26895,10 @@ std::string Player::GetDebugInfo() const
     sstr << Unit::GetDebugInfo();
     return sstr.str();
 }
+
+void Player::InterruptSpellsByMovement()
+{
+    for (uint32 i = CURRENT_FIRST_NON_MELEE_SPELL; i < CURRENT_MAX_SPELL; i++)
+        if (m_currentSpells[i] && m_currentSpells[i]->IsInterruptedByMovement())
+            InterruptSpell(static_cast<CurrentSpellTypes>(i), false);
+}
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -2171,6 +2171,8 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
 
         std::string GetDebugInfo() const override;
 
+        void InterruptSpellsByMovement();
+
     protected:
         // Gamemaster whisper whitelist
         GuidList WhisperList;
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -35,6 +35,7 @@
 #include <boost/accumulators/statistics/variance.hpp>
 #include <boost/accumulators/accumulators.hpp>
 #include <boost/accumulators/statistics.hpp>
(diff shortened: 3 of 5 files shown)


HelloKitty:
There are plenty of other ways to address tracking if a player moved since last update but this is how I had originally done it afew weeks ago on a fork of mine. So, I made the PR with that.

jackpoz:
shouldn't ToPlayer()->isMoving() return true if the player is moving ?

HelloKitty:
@jackpoz It does return true if the player is moving, but it doesn't return true if the player moved since last Player::Update and then stopped moving which can happen since 100 packets can be handled in WorldSession::Update. This means you can move inbetween updates and not interrupt the casting, if you do it the right way.

Ovahlord:
Long story short: multiple movement packets processed in one update tick override each other's movement flags instead of collecting them all for movement based actions. That might almost deserve a baseline member instead of having a spell specific special snowflake

HelloKitty:
@Ovahlord Yea, it might be useful to aggregate these for other purposes too. My concern about not tying these to currently casted spells is outlined in above though.

Basically, if you move, stop, cast a spell and *dont* handle it specific to spells as you propose then the spell will be interrupted even though you weren't moving after the spell was cast but did move during the last WorldSession::Update.

Shauren:
I really need to ask this... why are doing this in such complicated way instead of interrupting on movement instantly?

HelloKitty:
@Shauren Good question, my approach was based on the assumption that someone put movement interrupting spells in Spell::Update for a reason. I'm unsure who or for what reason but I decided to make as minor a change as possible.

I agree interrupting spells immediately would be not only simpler but also ideal for other cases such as responsiveness to client on casting state, especially when there is high diff, and also fixes the case where you move while casting and try to cast right away in the same tick, but it says you're still casting because it hasn't been interrupted yet.

Would you like me to change the PR to interrupt casting in the packet handler immediately instead?

Shauren:
Yes

HelloKitty:
@Shauren As requested, this new commit will immediately interrupt spell casting if the new MovementInfo from the movement packet indicates they are moving and if the spell is interruptable by movement.

It was not as straight forward as I had hoped it'd be, since there is more things to consider than just if the movement flag is set on the MovementInfo and to interrupt all spells. Feel free to adjust this further, I have not tested this version because I do not have a working local based on current TrinityCore Head.

jackpoz:
could someone test this ?

HelloKitty:
I am actually having trouble replicating this on TC 3.3.5 Head, even though the fork I identified this being a bug on is only a month or so old and I try to keep it mostly up to date. CMSG_CANCEL_CAST seems to be interrupting it currently. But that relies on client sending it. This was not happening before though.

Something may have changed since a month or two ago, not 100% what. When I have more free time I'll investigate again.

jackpoz:
well, isn't this PR a safeguard against people not sending CMSG_CANCEL_CAST ?

Issue #24226, opened 2020-03-01
SAI/Movement: Evade after a waypoint


Killyana:
**Description:** if you call an evade after a waypoints (via SAI) to reset the creature to spawn position it will no longer work, the npc will reset instead at the last way point. (no combat involved here)

I don't know if we need to update all SAI scripts or if this can be fix directly in the core.
Example: https://github.com/TrinityCore/TrinityCore/issues/23917 
This npc will use a waypoint, after a while an evade is called (the waypoint is finish) but instead returning to the home position, it will stay at the last point of the waypoint.
``` sql 
SELECT *  FROM `smart_scripts` WHERE entryorguid IN (-102327);
```
**Steps to reproduce the problem:**

1. Do the quest https://www.wowhead.com/quest=12140/all-hail-roanauk
2. Notice at the end of the event Taunka Soldiers didn't return to their initial position
3. even using .npc evade

**Branch(es):** 3.3.5

**TC rev. hash/commit:** rev. ce449f6b5332

jackpoz:
What about when a Creature engages combat during a waypoint path ? In this case evade should return to last waypoint.

Killyana:
The old behavior for creatures using waypoints (not waypoint_data) was:
1- If the creature never enter combat during the wp it will evade to the spawn position
2- If it evades after a combat it will evade to the position where the combat starts (every combat start will update the home position)
But after the movement rewrite all waypoints "types" were unified and now have the same behaviors, so scripts that use this differences are no longer working properly (same for the waypoints pause via the gossip)

Killyana:
We can add 2 params to the action:
Param1: 0/1 (return to spawn pos/return to the current home pos)
Param2: 0/1 (run/walk)

Rushor:
Why has this movement rewrite been merged without SAI tests actually?

Rushor:
this also happens if you use move point action. seems like after reaching a specific point, the new reached position is set to the homeposition

Killyana:
I cannot reproduce the issue with move_to_pos
.tele test
.npc add 321
``` sql
UPDATE `creature_template` SET `AIName`="SmartAI" WHERE entry =321;
DELETE FROM `smart_scripts` WHERE `entryorguid` IN (321) AND `source_type`=0;
[...]

Issue #23917, opened 2019-11-15
DB/Quest: All Hail Roanauk!


citizenwow:
Description:

In quest https://www.wowhead.com/quest=12140/ There are no phrases completely and it is impossible to pass the quest. https://www.wowhead.com/npc=26810/

Branch(es): 3.3.5

TC rev. hash/commit:

TrinityCore rev. c79571a 2019-11-03 17:00:24 +0100

Operating system: CentOS 7

dr-j:
Cannot reproduce take quest from overlord agmar he will say line and walk to where roanauk is when agmar is there he will be display gossip if you add quest with command or try to do at a later date this will not work because agmar wont be near him this is a condition of the gossip that agmar is near by

citizenwow:
I take this quest without any command.
![Без названия](https://user-images.githubusercontent.com/57755620/68990103-2f335f80-0860-11ea-8832-239182719090.png)

dr-j:
Overlord agmar has got to be close by for the option to be displayed as said he starts wp on quest accept if you go to roanauk before agmar gets there he won’t display option

dr-j:
https://youtu.be/m9IkxjhAuqc

As said works fine if you try speaking while overlord agmar is too far away he wont display gossip option when you speak to Roanauk Overlord agmar has to be within 30 yards of him otherwise wont display gossip option for event

Killyana:
I remember I case where the quest could be stuck (if the event starts and no one talk to Roanauk ), also some minor visuals must be fixed.

Rushor:
so adding an evade after some minutes should fix that (+active flag)

Killyana:
Something like:
``` sql
DELETE FROM `smart_scripts` WHERE `entryorguid`=26379 AND `source_type`=0 AND `id` IN (5,6,7,8,9,10);
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
[...]

PR #24492, opened 2020-04-23
 Game/AI: Implement option to move a creature back to Homeposition on Evade.

Lumber-Jack-2:
…Evade.

* If ToHomePosition (action param 1) is 1 > Move to last stored position
* If ToHomePosition (action param 1) is 0 > Move to homeposition
* Ref: https://github.com/TrinityCore/TrinityCore/issues/24226

**Changes proposed:**

* If ToHomePosition (action param 1) is 1 > Move to last stored position
* If ToHomePosition (action param 1) is 0 > Move to homeposition
* Ref: TrinityCore#24226

- [x] 3.3.5

https://github.com/TrinityCore/TrinityCore/issues/24226
yes

- [x] nothing
- [ ]


Killyana:
Can you add an other param walk/run ?

Rushor:
Why not adding it with the existing Action setwalk/run?

Should it run by default?

Killyana:
Yes it must run by default, but as we will update existing SAI it's easier to just update the current line than adding an extra one for walk.

Rushor:
okay cool, i will see if it is possible

Rushor:
@Killyana added, if param walkmode 0 run by default

Killyana:
I tested this issue https://github.com/TrinityCore/TrinityCore/issues/23917
The taunkas around Roanauk must evade to home position walking when the event ends.
But the just stay there:
image

``` sql
UPDATE `smart_scripts` SET action_param1=0, action_param2=1 WHERE `entryorguid` IN( -102327,-102329,-102326,-102330,-102328,-102341,-102333) AND action_type=24;

Rushor:
@Killyana okay i think the homeposition was not stored here - and name was wrong.

what we want: if paramter is 0 > move creature to the spawnposition.

what i changed:

if action_param1 (ToRespawnPosition) = 0, store new homeposition from respawnposition and move to it.

Killyana:
It works fine now, thanks.
I need to some tests for other situations to be sure nothing get broken.

Killyana:
Npcs always run on evade whatever the value of action_param2

Rushor:
hm i cannot see the reason why they always run

jackpoz:
> hm i cannot see the reason why they always run

https://github.com/TrinityCore/TrinityCore/blob/dc7856644f48d6b313e973197fed69ef56d6b5e0/src/server/game/Movement/MovementGenerators/HomeMovementGenerator.cpp#L79

jackpoz:
should the walk/run bit be ignored and merge the PR the way it is now ?

illfated:
Could be a useful incentive to fix the mentioned walk/run issue if the current functionality becomes available to all. From what we have seen too often, testing does not take place until the code is published.

Rushor:
hm yeah then let's merge that "remove to home" fix first. i will remove the option with walk/run again.

or i must copy the function of GetMotionMaster()->MoveTargetedHome(); and define the 

what would you suggest?

Killyana:
In thins case make the npc return to home pos if there's no escort, and evades to the last point if it evades after a combat, no need to add an extra param

jackpoz:

jackpoz:
>jackpoz commented on Jun 6

2nd try: what's the status of this PR ?

illfated:
Hello, @Rushor . Thanks for the code to the "Force of Neltharaku" PR which is merged now. 
I see you have been active about 2 days ago. 
Everything OK, I hope?

jackpoz:
Setting the status of this PR back to Draft as there are still some things left to do. Please mark it as ready for review once they have been handled.

jackpoz:
Thanks for the PR :)

PR #24506, opened 2020-04-27
Core\GameObject: corrected check for IsWithinDistInMap

ForesterDev:
**Changes proposed:**

-  Before change checks IsInMap(this) && InSamePhase(this) are always return true, because we check, if gameObject is in same phase or map with himself

- [x] 3.3.5

Build

Don't know, just logical change

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -2730,7 +2730,7 @@ bool GameObject::IsAtInteractDistance(Position const& pos, float radius) const
 
 bool GameObject::IsWithinDistInMap(Player const* player) const
 {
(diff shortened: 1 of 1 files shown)


Faq:
ref to PR #23226

Jildor:
is needed add
```c
return player && ...
```
like WorldObject::IsWithinDistInMap
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Object/Object.cpp#L1199

??

jackpoz:
@Jildor IsInMap() already returns false if the input is null, see
https://github.com/TrinityCore/TrinityCore/blob/411033ddf819db9e8ac5c1e6c540cf529138b5e6/src/server/game/Entities/Object/Object.cpp#L1165-L1170

jackpoz:
Thanks for the PR :)

PR #24516, opened 2020-04-29
Fix movement map generation

stoneharry:
**Changes proposed:**
Fixes movement map generation.

I ran mmaps with `--debugOutput true` to examine the navmesh generated using Recast. However I noticed that none of the navmeshes were actually being generated. After some debugging I found this culprit - skipping tiles that were perfectly valid, and trying to run on invalid tiles.

- [ ] 3.3.5
- [ ] master

- Builds and tested in game.


Shauren:
No, current code is correct.
If file doesnt exist: return false so we generate new one
If file has invalid format, outdated version: return false so we generate a new one

Shauren:
What you probably want is to always delete current file if --debugOutput was given

PR #24523, opened 2020-04-30
Core/Objects: further improvements for MovePositionToFirstCollision

Ovahlord:
- the PathGenerator will now normalize incomplete destinations as well
- normalize destination positions before launching detour raycasts to get better direction data
- exclude unwanted pathfinding results from further use to avoid unintended behaivior

- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -3270,19 +3270,43 @@ void WorldObject::MovePositionToFirstCollision(Position &pos, float dist, float 
         return;
     }
 
+    // Update our desired destination height
+    UpdateAllowedPositionZ(destx, desty, destz);
+
     // Use a detour raycast to get our first collision point
     PathGenerator path(this);
     path.CalculatePath(destx, desty, destz, false, true);
+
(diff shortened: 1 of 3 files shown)


jackpoz:
Tested locations:
- .go xyz 1001.908142 -4580.841797 55.079449 1 (this still allows to get on the stone in front if close enough)
- .go xyz 5989 2167 593 571 going up and down the stairs

Killyana:
Some issues not related to this PR: https://imgur.com/0np4EMR

Could be reproduce it in:  1434.846, -4417.37988, 25.46262, 0.362067 1

jackpoz:
@Ovahlord take a look at https://github.com/TrinityCore/TrinityCore/blob/d3578c34743fe1016bf657d9c13d509534ff2c7b/src/server/game/Spells/Spell.cpp#L1358-L1374 
It calls MovePositionToFirstCollision (which does a mmaps check now) and then calls CalculatePath() on the resulting end point. CalculatePath() is pointless in this case, could you please check all the cases like this with CalculatePath() call after MovePositionToFirstCollision ?

jackpoz:
The spell thing will be fixed in another PR

PR #24548, opened 2020-05-03
Core/DBC: Implement ServerMessage.dbc and generalize SMSG_SERVER_MESSAGE

HelloKitty:
**Changes proposed:**

-  Load ServerMessages.dbc.
-  Generalize SMSG_SERVER_MESSAGE to not assume that a message is formattable based on known blizzlike DBC entry ids.

The intention of this commit, like my other recent ones, is to help TrinityCore be slightly more generalized as an MMORPG framework since I use it as such.

- [x] 3.3.5
- [ ] master

Broadcast messages still work, I tested them. I did not test with any non-blizzlike DBCs though.

This is my first time loading a DBC so maybe someone will find some issues.

### DIFF
--- a/src/server/game/World/World.cpp
+++ b/src/server/game/World/World.cpp
@@ -3000,8 +3000,12 @@ void World::SendServerMessage(ServerMessageType type, const char *text, Player* 
 {
     WorldPacket data(SMSG_SERVER_MESSAGE, 50);              // guess size
     data << uint32(type);
-    if (type <= SERVER_MSG_STRING)
+
+    if (sServerMessagesStore.AssertEntry(type)->IsFormattableMessage())
+    {
+        ASSERT_NOTNULL(text);
         data << text;
+    }
 
     if (player)
         player->SendDirectMessage(&data);
--- a/src/server/game/World/World.cpp
+++ b/src/server/game/World/World.cpp
@@ -2998,13 +2998,26 @@ uint32 World::ShutdownCancel()
 /// Send a server message to the user(s)
 void World::SendServerMessage(ServerMessageType type, const char *text, Player* player)
 {
+    ServerMessagesEntry const* messageEntry = sServerMessagesStore.LookupEntry(type);
+
+    if (!messageEntry)
+    {
+        TC_LOG_ERROR("misc", "ServerMessages.dbc entry %u missing. Failed to send SMSG_SERVER_MESSAGE in World::SendServerMessage.", static_cast<uint32>(type));
+        return;
+    }
+
     WorldPacket data(SMSG_SERVER_MESSAGE, 50);              // guess size
     data << uint32(type);
 
-    if (sServerMessagesStore.AssertEntry(type)->IsFormattableMessage())
+    if (messageEntry->IsFormattableMessage())
     {
-        ASSERT_NOTNULL(text);
-        data << text;
+        if (text)
+            data << text;
+        else
+        {
+            TC_LOG_WARN("misc", "ServerMessages.dbc entry %u is formattable but no input was provided. Empty string used.", static_cast<uint32>(type));
+            data << std::string();
+        }
     }
 
     if (player)
--- a/src/server/shared/DataStores/DBCStructure.h
+++ b/src/server/shared/DataStores/DBCStructure.h
@@ -1904,7 +1904,7 @@ struct ServerMessagesEntry
     uint32 ID;                                              // 0
     char* Text;                                             // 1 Message, it's localized.
 
-    inline bool IsFormattableMessage() const { return std::strstr(Text, "%s") != nullptr; }
+    inline bool IsFormattableMessage() const { return strstr(Text, "%s") != nullptr; }
 };
 
 #pragma pack(pop)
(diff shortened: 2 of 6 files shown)


HelloKitty:
Hmmm, works on MSVC. Seems like I need a replacement for std::strstr for clang?

src/server/game/World/World.cpp
```diff
@@ -3000,8 +3000,12 @@ void World::SendServerMessage(ServerMessageType type, const char *text, Player*
 {
     WorldPacket data(SMSG_SERVER_MESSAGE, 50);              // guess size
     data << uint32(type);
-    if (type <= SERVER_MSG_STRING)
+
+    if (sServerMessagesStore.AssertEntry(type)->IsFormattableMessage())
```
Ovahlord, line 3004: that's  very unstable because it only takes someone to use the wrong type and all will blow up. Better use a LookupEntry call here and just cancel this whole helper execution if the entry is not found

HelloKitty: I'm not sure what the desired behavior should be though. In the scenario you propose though, someone has written code that tries to send a message that doesn't exist. That calling code is technically wrong. But should it crash/assert? Should it silently ignore? Should it log an error?

I can change it but I am just curious what the best approach would be.

In a perfect world we would auto-generate these types of enums from the DBC somehow and wouldn't be able to send messages that don't exist.

Ovahlord: I would just look up the entry at the beginning of the function and if the entry does not exist, just throw an error log and block further execution. Since the core is being used by all different kinds of people and there is no documentation, having it user friendly should be prioritized imo

HelloKitty: As requested I have removed the strict assertions and warn/error log instead. I think a better approach would be to make the server message type an enum class and check at startup that entries exist for all enumerated values. But I wouldn't know how to do that in C++, in C# that'd be easily done.

Anyway, it is more friendly now I guess.

Shauren:
This one is unneccessary:

* the string should always be present in the packet (even empty one as null-terminator is also sent)
* passing extra args that wont be used to printf-family functions is safe
* 3.3.5 client will use `[%d]: %s` (id, text) when entry is not found in dbc (so preventing sending isnt even bad if you want to send custom text formatted like that)

HelloKitty:
@Shauren Ohhh, that is very helpful information! I will adjust the PR accordingly, I think there is still some value that can be salvaged after some adjustments.

jackpoz:

src/server/shared/DataStores/DBCStructure.h
```diff
@@ -24,6 +24,7 @@
 #include "Util.h"
 #include <set>
 #include <map>
+#include <string.h>
```
jackpoz, line 27: this should be <cstring>

src/server/shared/DataStores/DBCStructure.h
```diff
@@ -1897,6 +1898,15 @@ struct WorldStateUI
 };
 */
 
+//ServerMessages.dbc
+struct ServerMessagesEntry
+{
+    uint32 ID;                                              // 0
+    char* Text;                                             // 1 Message, it's localized.
```
Carbenium, line 1905:
```suggestion
    char* Text[16];                                         // 1-16
    //uint32 Text_lang_mask;                                // 17
```

src/server/shared/DataStores/DBCStructure.h
```diff
@@ -1897,6 +1898,15 @@ struct WorldStateUI
...
+
+    inline bool IsFormattableMessage() const { return strstr(Text, "%s") != nullptr; }
```
Carbenium, line 1907:
```suggestion
    inline bool IsFormattableMessage() const { return strstr(Text[0], "%s") != nullptr; }
```

src/server/shared/DataStores/DBCStructure.h
Carbenium:
```suggestion
#include <cstring>
```

src/server/shared/DataStores/DBCfmt.h
```diff
@@ -140,5 +140,6 @@ char constexpr WMOAreaTableEntryfmt[] = "niiixxxxxiixxxxxxxxxxxxxxxxx";
 char constexpr WorldMapAreaEntryfmt[] = "xinxffffixx";
 char constexpr WorldMapOverlayEntryfmt[] = "nxiiiixxxxxxxxxxx";
 char constexpr WorldSafeLocsEntryfmt[] = "nifffxxxxxxxxxxxxxxxxx";
+char constexpr ServerMessagesEntryfmt[] = "nsxxxxxxxxxxxxxxxx";
```
Carbenium, line 143:
```suggestion
char constexpr ServerMessagesEntryfmt[] = "nssssssssssssssssx";
```

Carbenium:
After some discussion we decided to not pull this change as per Shauren's coment (https://github.com/TrinityCore/TrinityCore/pull/24548#issuecomment-623076372).

Regardless, thanks for the PR!

PR #24235, opened 2020-03-04
force charset utf-8 compiling for unicode charset on VS

dxzhan:
add compile option utf-8 for unicode on vs

PR #24572, opened 2020-05-06
force charset utf-8 compiling for unicode charset on VS

qyh214:
**Changes proposed:**

-  cmake:force charset utf-8 compiling for unicode charset on VS
-  console:fix chinese message show ???

- [x] 3.3.5
~~- [ ] master~~

**Issues addressed:** N/A

**Tests performed:** BUILD、CI、play in game

N/A

continued #24235 by @dxzhan

Note:
- When visual studio compiles, if the cpp file contains Chinese characters, an error will be reported after compile or the Chinese characters in the game will display `???`. The reproduction mode is as follows:

Add a custom script to output an system message and written in Chinese characters when onlogin(). Before PR, vs may not pass the compilation, or after compilation and enter the game. you will see the system message shows `???`

- second console fix:trinity_string.content_loc4 is Chinese(zhCN),you translate. if this message shows in console.before this pr it will show `???`

### DIFF
--- a/src/server/worldserver/CommandLine/CliRunnable.cpp
+++ b/src/server/worldserver/CommandLine/CliRunnable.cpp
@@ -28,7 +28,9 @@
 #include "Log.h"
 #include "Util.h"
 
-#if TRINITY_PLATFORM != TRINITY_PLATFORM_WINDOWS
+#if TRINITY_PLATFORM == TRINITY_PLATFORM_WINDOWS
+#include <Windows.h>
+#else
 #include <readline/readline.h>
 #include <readline/history.h>
 #include "Chat.h"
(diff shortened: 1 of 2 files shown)


src/server/worldserver/CommandLine/CliRunnable.cpp
```diff
@@ -91,7 +93,9 @@ void utf8print(void* /*arg*/, char const* str)
     if (!Utf8toWStr(str, strlen(str), wtemp_buf, wtemp_len))
         return;
 
-    wprintf(L"%s", wtemp_buf);
+    char temp_buf[6000];
```
jackpoz, line 96: does a wchar_t[6000] fit in a char[6000] ? is there any C++17 feature or so that might replace the platform-specific CharToOemBuffW call ?

Shauren: hell no it doesnt, sizeof(wchar_t) is 2 on windows, 4 on linux

Warpten: wchar is cancer, avoid at all cost.

Warpten: Also this is going to warn about stack size

PR #24618, opened 2020-05-13
Core/CreatureAI: ICC/Shambling Horror re-cast Enrage as soon as it is not stun

Metalaka:
**Changes proposed:**

-  If you stun the Shambling Horror during the cast of Enrage, they will re-cast it as soon as the stun fades

- [x] 3.3.5
- [ ] master (should probably apply too ?)

build, tested IG in 10 & 25

- I have no proof that the Shambling Horror's Shockwave work the same way, so it's not part of the changes.

**Sources**
https://www.wowhead.com/spell=72143/enrage#comments
If you stun the Shamblers they will re-cast it as soon as the stun fades. Even if stunned part way through the cast. Even though the spell data has a cooldown of 7s they never use it more than once every 20 on the Lich King 25 encounter


jackpoz:
Thanks for the PR :)

PR #24631, opened 2020-05-15
Fixes/3.3.5 aura infinite loop

jackpoz:
**Changes proposed:**

-  Core/Auras: Attempt to fix infinite loop with aura 18950 on map unload
-  Core/Auras: Log a detailed error and assert when failing to remove all auras after a few tries

- [x] 3.3.5
- [ ] master

Some crashes

None

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -4130,7 +4130,7 @@ void Unit::RemoveAllAuras()
 {
     // this may be a dead loop if some events on aura remove will continiously apply aura on remove
     // we want to have all auras removed, so use your brain when linking events
-    while (!m_appliedAuras.empty() || !m_ownedAuras.empty())
+    for (int counter = 0; !m_appliedAuras.empty() || !m_ownedAuras.empty(); counter++)
     {
         AuraApplicationMap::iterator aurAppIter;
         for (aurAppIter = m_appliedAuras.begin(); aurAppIter != m_appliedAuras.end();)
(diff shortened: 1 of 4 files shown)


kelno:
Might be interesting:
I just had the same issue happen to me on my live server but in RemoveAllAurasOnDeath
Some aura was being re applied at each remove.

jackpoz:
did you get a crashlog with some information about which unit/aura ?

kelno:
I have rewritten bot airforce scripts which were causing this issue:

- The airforce bots put spell 38067 on players, which has aura effect SPELL_AURA_MOD_STALKED
- If players dies, he loses the aura, which triggers a visibility update (in HandleAuraModStalked handler), which triggered a new spell cast from the airforce bot MoveInLineOfSight
Infinite loop ensues just like in this pull request.

So, not currently an issue on TC, but super similar to this case and not hard to do in any script.

sirikfoll:
Got the crash when unloading npc [Stonevault Cave Lurker](https://www.wowhead.com/npc=4850/stonevault-cave-lurker) in mapId 70 (Uldaman).
Npc had aura [22766](https://www.wowhead.com/spell=22766/sneak) (SPELL_AURA_MOD_STEALTH and SPELL_AURA_MOD_DECREASE_SPEED), didn't catch any extra logs :/

jackpoz:
No crashlog or server log ?

sirikfoll:
No, got this crash some time ago and just remembered to inform, will try to activate the logs, but it's a very very rare crash..

PR #24634, opened 2020-05-16
Core/Units: creatures will now also consider manually set movement flags for CanFly and CanHover

Ovahlord:
**Changes proposed:**

-  Creatures will now also be able to benefit from flight movement when their movement flags have been changed outside of their template movement (flight auras, scripts etc)
-  renamed IsLevitating to IsGravityDisabled because that's what the movement flag implies

- [x] 3.3.5
- [ ] master

-tested ingame, need more test results

### DIFF
--- a/src/server/game/Entities/Creature/Creature.h
+++ b/src/server/game/Entities/Creature/Creature.h
@@ -98,8 +98,8 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         CreatureMovementData const& GetMovementTemplate() const;
         bool CanWalk() const { return GetMovementTemplate().IsGroundAllowed(); }
         bool CanSwim() const override { return GetMovementTemplate().IsSwimAllowed() || IsPet(); }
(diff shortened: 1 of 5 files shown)


joshwhedon:
The part about checking if the creature is flying to decide if it can fly looks good, and this thing has been making creatures behave weirdly. And renaming IsLevitating is consistant because ... levitating doesn't mean much in the context.

joshwhedon (approved):
It looks good, good call on renaming IsLevitating, it makes more sense this way. I guess we will have to wait and see if this breaks stuff in the end.

jackpoz:
@Ovahlord is this ready to be merged ?

Jildor:
Without this PR Onyxia fly, Queen Lana'thel and other bosses scripts are behave weird, maybe caused by this: https://github.com/TrinityCore/TrinityCore/commit/fdcb7388a64bf61a691535933a9fbae0504b80cc
I think can be merge

PR #24637, opened 2020-05-16
Tools/vmap_extractor: Fixed typos reading MAIN and MAID chunks

Ghaster:
**Changes proposed:**

-  vmap_extractor is currently discarding many ADTs, due to reading the MAIN chunk array incorrectly. It is supposed to be read as [y][x], not [x][y]. 

- [ ] 3.3.5
- [x] master

- Tested by generating completely new vmaps and mmaps, fixes worldserver shutting down due too missing vmaps in the starting zone.

### DIFF
--- a/src/tools/vmap4_extractor/wdtfile.cpp
+++ b/src/tools/vmap4_extractor/wdtfile.cpp
@@ -156,13 +156,13 @@ ADTFile* WDTFile::GetMap(int32 x, int32 y)
     if (_adtCache && _adtCache->file[x][y])
         return _adtCache->file[x][y].get();
 
-    if (!(_adtInfo.Data[x][y].Flag & 1))
+    if (!(_adtInfo.Data[y][x].Flag & 1))
         return nullptr;
 
     ADTFile* adt;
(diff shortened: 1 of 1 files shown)


Shauren:
I have to know what are you smoking since worldserver does NOT shut down to missing vmaps in starting zone at all

Ghaster:
Then that might have been something else, going to double check. However it is definitely stored as [y][x] in the WDT (atleast in the MAIN chunk).

Shauren:
It seems you are right

Shauren:
Ok I dug even deeper into this
Turns out everything is working as intended anyway thanks to another mistake in the same function (x and y should also be swapped when accessing `_adtFileDataIds->Data[y][x]`) but because vmapextractor works by iterating entire 64x64 grid anyway, it enters this function again later with x and y having the opposite values so the adt is extracted anyway

Ghaster:
Ah, that would explain the non-working server startup for me, since I was testing stuff out with non-filedataID WDTs.

Updated with reading _adtFileDataIds correctly aswell.

Issue #22729, opened 2018-10-29
DB/Quest: Extinguishing the Idol


TheWinchesters:
**Description:**
Quest Extinguishing the Idol (https://www.wowhead.com/quest=3525/extinguishing-the-idol) 
End: Belnistrasz's Brazier (https://www.wowhead.com/object=152097/belnistraszs-brazier)

**Current behaviour:**
Belnistrasz's Brazier does not exist and does not appear after escorting Belnistrasz (https://www.wowhead.com/npc=8516/belnistrasz)

**Expected behaviour:**
Not sure, but I guess the object should spawn after Belnistrasz event.

1. .go c id 8516
2. Accept the quest 3525
3. Escort npc 8516

Branch: 3.3.5
TC rev: 9b02fe6e79199cde122894c9c784648aee11cd98
OS: Win 8.1

illfated:
That's odd. I distinctly remember completing that quest at least a couple of times on TC source during the last few years, but possibly half a year ago since last time I did it.
Could there be something in the latest chrono changes in the file where that object should be spawned?
https://github.com/TrinityCore/TrinityCore/blob/5b0f4009df4cde413c92c73467ca9a63e79a772b/src/server/scripts/Kalimdor/RazorfenDowns/razorfen_downs.cpp#L204-L228

illfated:
TrinityCore rev. 257ae44a20ed 2018-10-28 10:07:56 +0100 (3.3.5 branch) (Win64, RelWithDebInfo, Static)
Confirmed, `me->SummonGameObject(GO_BELNISTRASZS_BRAZIER` does not happen.

@DanVS : could you take a look and see if there could be a timing issue here?
(or if there are any other gameobject related changes causing this)

illfated:
Yes, the room shakes for a second, although it feels like it is too short and happens just a second before Belnistrasz despawns himself.
After Belnistrasz despawns, there is no gameobject near (`.gob near`) at the spot Belnistrasz stood:
`Found nearby gameobjects (distance 10.000000): 0`

I am not sure, but for some reason I seem to remember there used to be enough time to talk to Belnistrasz before he despawned in the older TC revisions.

celinia:
This quest is still bugged.

TrinityCore rev. 8a9f80909387 2020-05-16 21:22:30 +0200 (3.3.5 branch) (Win64, RelWithDebInfo, Static)

After completing the summoning event no brazier spawns.

illfated:
[...]

PR #24655, opened 2020-05-18
Core/Object: Extend SummonGameObject function to control when the obj…

Faq:
…ect will despawn

**Changes proposed:**

-  Extend SummonGameObject function to control when the object will despawn, not to limit just for SAI ca4f1e33

- [x] 3.3.5

ref #22729

Compiles and tested with #22729 where summoner is despawned, gob despawns only after provided timer.

- [check if breaks anything else? ]


illfated:
Looks promising so far. Looking forward to what the TC members have to say about this update.

PR #24802, opened 2020-06-13
Worldserver/CLI: console printf fix

ipriver:
**Changes proposed:**

-  Some CLI printf functions had different output because of a missing space.
-  Removed extra console printf for *nix systems
Screen Shot 2020-06-13 at 22 07 24

These issues only visible if Console is Enabled in the worldserver config. 

- [X] 3.3.5
- [ ] master 

Cannot check master branch, but the CLI code is the same, so I assume this PR should also fix it.

**Tests performed:**  Compiled, manually tested in the console.

### DIFF
--- a/src/server/worldserver/CommandLine/CliRunnable.cpp
+++ b/src/server/worldserver/CommandLine/CliRunnable.cpp
@@ -32,7 +32,16 @@
 #include <readline/readline.h>
 #include <readline/history.h>
 #include "Chat.h"
+#endif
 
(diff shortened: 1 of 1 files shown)


jackpoz:
are all the changes needed or is there some refactoring included too in the changes ? mixing the 2 might make it harder to review the PR

ipriver:
I'd say that the minor refactoring here is the part of the fix.

jackpoz:
Thanks for the PR :)

PR #24809, opened 2020-06-14
Log number of Creatures and GameObject per map

jackpoz:
**Changes proposed:**

-  Log number of Creatures and GameObject per map in InfluxDB/Grafana. This should help spot strange cases like instances with a lot of creatures (i.e. Violet Hold)

image

- [x] 3.3.5
- [ ] master

**Issues addressed:** None

Tested on Windows

- [ ] Create a proper Grafana dashboard
- [ ] Update the Grafana dashboards included in TC (and maybe to latest Grafana version too)


src/common/Metric/Metric.h
```diff
@@ -100,7 +105,7 @@ class TC_COMMON_API Metric
     void Update();
 
     template<class T>
-    void LogValue(std::string const& category, T value)
+    void LogValue(std::string const& category, T value, std::vector<MetricTag> const& tags)
```
Shauren, line 108: is there a reason you didn't want tags to be moved instead of copied?

jackpoz:
Default (no stacking)
image

With stacking
image

jackpoz:
Each instance id now has its own line in the graph

image

PR #24812, opened 2020-06-15
CMake: Additional cmake entry for configs

ipriver:
**Changes proposed:**
I prefer to have clean project directories and use -DTOOLS=0 to stop compiling unnecessary binaries, but there is no option to disable copying configuration files in a cmake, and every time after the build it copies default config files into a config directory. 
Yeah, you can make a trick only available to *nix systems and set DCONF_DIR to a value of "/dev/null", but it anyway adds an extra step.
So the proposal is to use new -DCONF=0 (by default works as before and is set to 1) entry that removes copying step for both *nix and Win systems.
-  Additional cmake entry that allows disabling copying of the configuration files

image

- [x] 3.3.5
- [x] master

Build works on *nix systems

- [ ] Need to test changes on a Windows system
- [ ] master branch requires additional changes in a bnetserver CMakeLists


cmake/options.cmake
```diff
@@ -52,3 +52,4 @@ option(WITH_STRICT_DATABASE_TYPE_CHECKS "Enable strict checking of database fiel
 set(WITH_SOURCE_TREE    "hierarchical" CACHE STRING "Build the source tree for IDE's.")
 set_property(CACHE WITH_SOURCE_TREE PROPERTY STRINGS no flat hierarchical hierarchical-folders)
 option(WITHOUT_GIT      "Disable the GIT testing routines"                            0)
+option(CONF             "Copy authserver and worldserver .conf.dist files to the project dir"      1)
```
Carbenium, line 55: `INSTALL_CONF` would make the meaning of the variable more clear.

Carbenium (changes requested):
Thanks for the PR!

Please address the inline comment and rebase+squash the PR. Afterwards it should be good to merge.

cmake/options.cmake
Aokromes: imho better copy_conf xd

Carbenium: @ipriver feel free to choose one suggestion ;) As long as it's more than just `CONF`, I'm fine with it.

ipriver: hah, I named it `CONF` because there are `SERVER` and `TOOLS` variables that are used to toggle compiling and I thought it would make sense to have the same one-world variable for configs.
Though if you still want to change the name of it I like `COPY_CONF` more :)

Carbenium: Then `COPY_CONF` it is. Makes it definitely easier to search for it.

ipriver: Done. Made a commit amend, rebased to the latest commit in 3.3.5 and force pushed to clean my previous commits in this branch.

cmake/showoptions.cmake
```diff
@@ -12,6 +12,15 @@ message("* Install core to        : ${CMAKE_INSTALL_PREFIX}")
 if( UNIX )
   message("* Install configs to     : ${CONF_DIR}")
 endif()
+
+if ( COPY_CONF )
```
Carbenium, line 16: Just small nit-pick (sorry!): we don't use spaces around parentheses in CMake
```suggestion
if( COPY_CONF )
```

cmake/showoptions.cmake
```diff
@@ -12,6 +12,15 @@ message("* Install core to        : ${CMAKE_INSTALL_PREFIX}")
...
+  if ( UNIX )
```
Carbenium, line 17:
```suggestion
  if( UNIX )
```

src/server/authserver/CMakeLists.txt
```diff
@@ -78,10 +78,14 @@ endif()
 
 if( UNIX )
   install(TARGETS authserver DESTINATION bin)
-  install(FILES  authserver.conf.dist DESTINATION ${CONF_DIR})
+  if ( COPY_CONF )
```
Carbenium, line 81:
```suggestion
  if( COPY_CONF )
```

src/server/authserver/CMakeLists.txt
```diff
@@ -78,10 +78,14 @@ endif()
...
+    install(FILES  authserver.conf.dist DESTINATION ${CONF_DIR})
+  endif()
 elseif( WIN32 )
   install(TARGETS authserver DESTINATION "${CMAKE_INSTALL_PREFIX}")
-  install(FILES authserver.conf.dist DESTINATION "${CMAKE_INSTALL_PREFIX}")
+  if ( COPY_CONF )
```
Carbenium, line 86:
```suggestion
  if( COPY_CONF )
```

src/server/worldserver/CMakeLists.txt
```diff
@@ -90,10 +90,14 @@ endif()
 
 if( UNIX )
   install(TARGETS worldserver DESTINATION bin)
-  install(FILES worldserver.conf.dist DESTINATION ${CONF_DIR})
+  if ( COPY_CONF )
```
Carbenium, line 93:
```suggestion
  if( COPY_CONF )
```

src/server/worldserver/CMakeLists.txt
```diff
@@ -90,10 +90,14 @@ endif()
...
+    install(FILES worldserver.conf.dist DESTINATION ${CONF_DIR})
+  endif()
 elseif( WIN32 )
   install(TARGETS worldserver DESTINATION "${CMAKE_INSTALL_PREFIX}")
-  install(FILES worldserver.conf.dist DESTINATION "${CMAKE_INSTALL_PREFIX}")
+  if ( COPY_CONF )
```
Carbenium, line 98:
```suggestion
  if( COPY_CONF )
```

Carbenium (changes requested):
Just the small nit-pick and good to go.

illfated:
@Carbenium : As I have understood it, there is no longer a requirement to squash PR commits, because the TC member who merges the PR will perform the squash action on merge (just F.Y.I.).

ipriver:
Hmm, there are different styles for if statements and variables in CMake files that's what confused me. Though most of the code is using `if(STATEMENT)` or `set(variable)` without spaces at all.
I suggest merging this PR if it doesn't have any other issues and later I'll create an extra PR where I go through all CMake files and fix indentation to be in one style. Don't want to mix style changes into this PR.

Carbenium:
dc467ee0f6d0a49be4bd20b621f141228dfec736

PR #24813, opened 2020-06-15
CMake: removed unused code for LIBSDIR entry

ipriver:
**Changes proposed:**

During the build, cmake informs that it is using a separate directory for libraries (as a default on *nix systems)

image

During the runtime we can see that new running server process is using shared libraries from the default system lib directory.

image

And that's because there is no code that creates a directory or links libs to the compiled binary. 

-  Remove legacy code related to LIBSDIR entry, so it won't confuse users during the build as it doesn't have an implementation.

- [x] 3.3.5
- [x] master

Build works on *nix systems

### DIFF
--- a/cmake/platform/unix/settings.cmake
+++ b/cmake/platform/unix/settings.cmake
@@ -4,12 +4,6 @@ if( NOT CONF_DIR )
   message(STATUS "UNIX: Using default configuration directory")
 endif()
 
-# set default library directory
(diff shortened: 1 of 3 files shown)


Aokromes:
what about dynamic build?

Carbenium (approved):
LGTM

jackpoz:
Thanks for the PR :)

PR #24821, opened 2020-06-16
Script/Spell: properly round damage of Seal of Righteousness (on-hit proc)

Rothend:
**Changes proposed:**

Currently, Seal of Righteousness' damage for the on-hit proc doesn't match the tooltip.
Example: human paladins, with their _massive_ 27 attack power, should deal 2 damage of SoR proc at level 1. Instead it deals 1 damage.
This is because the spell script truncates the calculated value, instead of rounding.

Video of human paladin [here](https://www.youtube.com/watch?v=oLy8hMW-e48). Sometimes it deals less damage, but that's because of the creature resisting some of the spell damage.

- [x] 3.3.5
- [ ] master

**Tests performed:** tested, works. Damage done (when not partially resisted) matches tooltip.

### DIFF
--- a/src/server/scripts/Spells/spell_paladin.cpp
+++ b/src/server/scripts/Spells/spell_paladin.cpp
@@ -2054,7 +2054,7 @@ class spell_pal_seal_of_righteousness : public SpellScriptLoader
                 float mws = GetTarget()->GetAttackTime(BASE_ATTACK);
                 mws /= 1000.0f;
 
-                int32 bp = int32(mws * (0.022f * ap + 0.044f * sph));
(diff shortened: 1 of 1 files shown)


jackpoz:
what is used in similar code ? int32(round(...)) ?

Rothend:
[Eviscerate](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellEffects.cpp#L569)'s calculation uses std::lroundf() too.

jackpoz:
Thanks for the PR :)

PR #24837, opened 2020-06-18
Core/Spells: SPELL_AURA_MOD_ATTACK_POWER_PCT should be negative if TargetType or Basepoints are negative

Jildor:
**Changes proposed:**

-  check TargetType and Basepoints for SPELL_AURA_MOD_ATTACK_POWER_PCT to determine if should be positive or negative, by example spell: ID - 61044 Demoralizing Shout should be negative

**Target branch(es):**

- [X] 3.3.5
- [ ] master

**Issues addressed:** Closes nothing I think

### DIFF
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -3641,6 +3641,7 @@ bool _isPositiveEffectImpl(SpellInfo const* spellInfo, uint8 effIndex, std::unor
             case SPELL_AURA_MOD_RANGED_ATTACK_POWER:
             case SPELL_AURA_MOD_DAMAGE_PERCENT_DONE:
             case SPELL_AURA_MOD_SPEED_SLOW_ALL:
+            case SPELL_AURA_MOD_ATTACK_POWER_PCT:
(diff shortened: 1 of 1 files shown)


Killyana:
@Jildor Maybe it's the same for this spell https://classic.wowhead.com/spell=3150
 If:  Mod Health Regen or  Mod Stat (any) has negative bp the spell should be negative.

Jildor:
well that case (unrelated with SPELL_AURA_MOD_ATTACK_POWER_PCT) needs more reseach, because SPELL_AURA_MOD_HEALTH_REGEN_PERCENT you have spells like Health Funnel, I don't know if Health Funnel should be negative, so maybe in that case (SPELL_AURA_MOD_HEALTH_REGEN_PERCENT) should be checked only TargetType and not BP (add here for fix: https://github.com/TrinityCore/TrinityCore/blob/075a3994444014e4264cdd7369c7207c0c2a09e6/src/server/game/Spells/SpellInfo.cpp#L3716), but better new issue and new PR for that

jackpoz:
Thanks for the PR :)

PR #24862, opened 2020-06-21
Core/Spells: SPELL_AURA_MOD_HEALING should be negative if TargetType is negative

### LINKED ISSUE #24854: Core/Spell: spells with the effect Apply Aura: Mod Healing Taken and negative bp must be displayed as debuff and not buff
**Description:** For example: https://woehead.way-of-elendil.fr/?spell=32858
Probably must be fixed like https://github.com/TrinityCore/TrinityCore/commit/7d88891983d5c7d2673be760e11bff12df958d41

1. find any hostile creature
2. .cast back 32858
3. the aura will be displayed as buff and can be removed by a right click

**Branch(es):** 3.3.5

**TC rev. hash/commit:** rev. e6b06c4f039f

Jildor:
**Changes proposed:**

-  SPELL_AURA_MOD_HEALING should be negative if TargetType is negative
-  Mage Dampen/Amplify Magic specific check (https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellInfo.cpp#L3432-L3434) still be necessary because those spells have SPELL_AURA_MOD_DAMAGE_TAKEN and if have BP > 0 (https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellInfo.cpp#L3648) Amplify Magic will be negative and it shouldn't be

- [X] 3.3.5
- [ ] master

**Issues addressed:** Closes #24854

### DIFF
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -3714,6 +3714,7 @@ bool _isPositiveEffectImpl(SpellInfo const* spellInfo, uint8 effIndex, std::unor
             case SPELL_AURA_MOD_ATTACKER_RANGED_CRIT_DAMAGE:
             case SPELL_AURA_MOD_ATTACKER_SPELL_AND_WEAPON_CRIT_CHANCE:
             case SPELL_AURA_DUMMY:
+            case SPELL_AURA_MOD_HEALING:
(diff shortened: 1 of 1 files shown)


jackpoz:
Thanks for the PR :)

PR #24865, opened 2020-06-21
Core/EscortAI: set maximum distance allowed between player and escortee to default vision range

Rothend:
**Changes proposed:**

When an escort NPC gets further than 100 yards from the player that started the escort, it despawns. The current distance for that behaviour in TC is 50 yards.

- [x] 3.3.5
- [ ] master


jackpoz:
Thanks for the PR :)

PR #24868, opened 2020-06-21
Misc: Add code style settings for CLion

Carbenium:
**Changes proposed:**

-  Add code style settings for CLion

- [x] 3.3.5
- [x] master

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,8 +17,14 @@ CMakeLists.txt.user
 *.BASE.*
 *.LOCAL.*
 nbproject/*
-.idea/*
 .browse.VC*
 .vscode
 cmake-build-*/
 .vs
+
+.idea/**/workspace.xml
+.idea/**/tasks.xml
+.idea/**/usage.statistics.xml
+.idea/**/dictionaries
+.idea/**/shelf
+.idea/misc.xml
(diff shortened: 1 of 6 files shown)


ipriver:
Not sure if it's a good option to pollute project directory with configuration files for specific IDEs.
There is already a generic [editorconfig](https://editorconfig.org/) solution that works everywhere. 
IMHO

Carbenium:
I agree that a universal way to do this would be nice.
On the other hand it is quite handy to have more or less our [styleguide](https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130103/C+Development+Standards) (including CMake code style) directly loaded by the IDE. My hope is that this lets new contributors focus more on the code then having to check whether they follow the styleguide (at least CLion users).

I'll let this PR sit here until tomorrow or so. Let's see if some other comments pop up, otherwise I'm still tending towards merging.

PS: An editorconfig is already in the repo.

ipriver:
The problem I see here is that it only affects CLion users, other ppl who use VS (.vs), VSCode (.vscode), Vim, etc. gonna require additional configuration directories and it will become hard to support these rules for different users. It's like a cross-platform development, but for IDEs.

Maybe it could be a good idea to bring a linter to enforce syntax rules into the project and attach it to the CI pipeline.  (e.g. http://clang.llvm.org/extra/clang-tidy/)  though I'm not very familiar with C++ linters.

Btw currently [.editorconfig](https://github.com/TrinityCore/TrinityCore/blob/master/.editorconfig) has very basic rules.

Shauren:
I don't like it, rather not put any ide-specific settings in repo

Carbenium:
Convinced :)

I like the `clang-tidy` idea. Let's see if that is a viable solution.

PR #24869, opened 2020-06-21
Replace Trinity::make_unique with std

Carbenium:
**Changes proposed:**

-  Replace Trinity::make_unique with std

- [x] 3.3.5
- [x] master

**Tests performed:** compile


Faq:
Just wondering why it was implemented this way (to use **Trinity::make_unique**) in first place.

Shauren:
@Faq std::make_unique was added in c++14, Trinity::make_unique is older (and had different implementation before)

PR #24874, opened 2020-06-22
Core/SAI: implement SMART_EVENT_SUMMONED_UNIT_DIES (82), useable by creatures and gameobjects

### LINKED ISSUE #24831: Core/SAI: SMART_EVENT_SUMMON_DIES is not implemented
SMART_EVENT_SUMMON_DESPAWNED triggers only after creature despawn. Pretty useless. In most cases we should use SMART_EVENT_SUMMON_DIES because not all creatures should despawn instantly after death.

**Branch:** both
**TC rev. hash/commit:** d81a9e5bc3b3

Rothend:
**Changes proposed:**

This script hook already exists in CreatureAI and GameObjectAI. This PR exposes it to SAI via event type 82.
Param1 is either 0 (fires for all creatures) or the entry of the summoned creature we want to trigger this even when it dies.
Param2 and param3 are min-max cooldowns.
Basically the same parameters as SMART_EVENT_SUMMONED_UNIT are used, because they fit this event nicely.

This is useful, for example, when scripting an escort quest where the escorted NPC pauses WP, summons hostile mobs and resumes WP when all those hostile mobs are killed. This would be achieved with a combination of this event and SMART_ACTION_SET_COUNTER/SMART_EVENT_COUNTER_SET.

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes #24831


src/server/game/AI/SmartScripts/SmartScriptMgr.h
```diff
@@ -1498,7 +1503,12 @@ const uint32 SmartAIEventMask[SMART_EVENT_END][2] =
     {SMART_EVENT_FRIENDLY_HEALTH_PCT,       SMART_SCRIPT_TYPE_MASK_CREATURE },
     {SMART_EVENT_DISTANCE_CREATURE,         SMART_SCRIPT_TYPE_MASK_CREATURE },
     {SMART_EVENT_DISTANCE_GAMEOBJECT,       SMART_SCRIPT_TYPE_MASK_CREATURE },
-    {SMART_EVENT_COUNTER_SET,               SMART_SCRIPT_TYPE_MASK_CREATURE + SMART_SCRIPT_TYPE_MASK_GAMEOBJECT }
+    {SMART_EVENT_COUNTER_SET,               SMART_SCRIPT_TYPE_MASK_CREATURE + SMART_SCRIPT_TYPE_MASK_GAMEOBJECT },
+    {SMART_EVENT_SCENE_START,               SMART_SCRIPT_TYPE_MASK_CREATURE },
```
Rothend, line 1507: I wasn't sure on what to put in here, since those events are only for the master branch. Let me know if it's wrong.

Ovahlord: comment them out then if they have no use or purpose at all.

Rothend: Can't really do that though, this array is indexed with the event id so it would be an access violation if every possible id wasn't accounted for.

jackpoz:
Thanks for the PR :)

Issue #21586, opened 2018-03-11
Quest: Ending Their World (9759)


jerbookins:
**Description:**
http://www.wowhead.com/quest=9759/ending-their-world

**Current behaviour:**
- Quest is missing some details and scripted elements, and is not completed after killing Sironas.
- Demolitionist Legoso walks at an annoyingly slow pace.
- Demolitionist Legoso doesn't always agro nearby creatures.
- After the first cluster of bombs goes off, there are no fire effects on the lower half of the Vector Coil.
- There are no electric-style effects coming from the tips of the giant crystals, down into Sironas' hands.
- Sironas isn't in a pose, holding her hands up in the air collecting the energy. She's just standing around.
- After the second cluster of bombs goes off, 
- When you defeat Sironas the quest is not completed.

**Expected behaviour:**
https://www.youtube.com/watch?v=4IhYoGu3LZU
super shitty quality but old and looks reliable.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 078b621c746f

**TDB version:** 335.64

**Operating system:** Windows 10

Killyana:
Please update your core and test again, the issue related to spells cannot be casted could be the reason off all that.

jerbookins:
You're half-right. A lot of stuff worked correctly.

Still broken stuff:
- Demolitionist Legoso walks at an annoyingly slow pace and he doesn't always agro nearby creatures. (Specifically, he ignores non blood-elf creatures, as well as an NPC named Fenissa the Assassin.) https://streamable.com/7dbf4
- The written portion of the script doesn't properly convert $n into the player name.
- After the first cluster of bombs goes off, there are no fire effects on the lower half of the Vector Coil. 
![wowscrnshot_031318_022127](https://user-images.githubusercontent.com/16745918/37330464-c4844e38-2666-11e8-8408-5173de3b9828.jpg)

offl:
The only fix I have, heh. WIP.

1. He now uses hacky waypoint pause for 5min because of #24759. So yeah, waypoint pauses ticks in combat. And if he will be in combat for too long, some weird things will happen. But the real problem is with the last pause. He is supposed to call action list after reaching home and if Sironas was killed. But without waypoint pause with timers it will not work and I don't know if #24759 will fix it.

2. Right now he uses event type 2 to cast the spell only on self.  Is he really supposed to cast the heal on his escort?
He will not cast his healing spell anymore on himself and player regardless of event type (14 or 74).
[...]

PR #24877, opened 2020-06-23
Core/SAI: do not allow the waypoint pause timer to update while in combat

Rothend:
**Changes proposed:**

Currently, the SAI waypoint pause timer keeps ticking down while the creature is in combat. If you pause the path right before the creature enters combat, it will start pathing straight away after combat ends.

In most escort quests, the creature summons hostile mobs, and after combat it stays paused, says/does something and then resumes pathing. If the pause timer keeps ticking in combat, this behavior requires various hacks to implement.

If a creature is in combat, it doesn't run any OOC/RP events. It makes sense that if we specify a pause timer, we want it to be paused for that amount even if the creature engages hostiles in the meantime.

- [x] 3.3.5
- [x] master

**Issues addressed:** couldn't find a specific issue about this. It is mentioned in https://github.com/TrinityCore/TrinityCore/issues/21586#issuecomment-647054110

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -933,13 +933,13 @@ void SmartAI::UpdatePath(uint32 diff)
     // handle pause
     if (HasEscortState(SMART_ESCORT_PAUSED) && (_waypointReached || _waypointPauseForced))
     {
-        if (_waypointPauseTimer <= diff)
+        if (!me->IsInCombat() && !HasEscortState(SMART_ESCORT_RETURNING))
         {
-            if (!me->IsInCombat() && !HasEscortState(SMART_ESCORT_RETURNING))
+            if (_waypointPauseTimer <= diff)
                 ResumePath();
+            else
+                _waypointPauseTimer -= diff;
         }
-        else
-            _waypointPauseTimer -= diff;
     }
     else if (_waypointPathEnded) // end path
     {


jackpoz:
Thanks for the PR :)

offl:
Not sure about this anymore. There's two types of cases:
In first you need to pause waypoint with no time to resume (permanent pause) and later resume waypoint manually(used in long events in escort scripts or in other waypoints events).
In second waypoint is resumed on timer.
How can you know if it's permanent pause or timed pause? Well, if it's long event, it's permanent. If not, it's hard to know.

I encountered one interesting case where event starts on timer and timer ticks in combat, but the event may executed only while out of combat. It's waypoint script and the event is not called On Reached Home(confirmed by videos). So the only way to implement it properly is let waypoint pause tick in combat. Or make tricky and messy actions.

Maybe anyone knows if waypoint pauses should really tick in combat? Can be checked in non-long pauses (5-10 seconds).
Gonna try later to check it.

offl:
I checked one case, seems like WP pause should not update in combat.
> I encountered one interesting case where event starts on timer and timer ticks in combat, but the event may executed only while out of combat. It's waypoint script and the event is not called On Reached Home(confirmed by videos). So the only way to implement it properly is let waypoint pause tick in combat. Or make tricky and messy actions.

Anyway turned out WP pause cannot be used to script something properly in this case or in another.

Faq:
`How can you know if it's permanent pause or timed pause? Well, if it's long event, it's permanent. If not, it's hard to know.`
How much event there, where permanent pause needed? Guessing in such cases are some actions what will cancel pause, for example if used some user action, so after it is done, pause is canceled and event continues.

So in the end if someone can enlist all needs of different cases, then can also modify behavior of this pause thing.

offl:
Permanent pause is definitely needed in large event scripts for better event synchronization.
Also pause cannot be canceled by some actions if I'm not wrong.

PR #24883, opened 2020-06-23
Next round of 3.3.5 -> master cherry-picks

Carbenium:

- [ ] 3.3.5
- [x] master


funjoker:
You gonna go whole June ? :P

funjoker:
We have to CP some header clean up changes too, because some are 335 specific changes and still required in master

src/server/game/Maps/Map.cpp
```diff
@@ -2899,6 +2903,103 @@ ZLiquidStatus Map::getLiquidStatus(PhaseShift const& phaseShift, float x, float
     return result;
 }
 
+void Map::GetFullTerrainStatusForPosition(float x, float y, float z, PositionFullTerrainStatus& data, uint8 reqLiquidType) const
+{
+    VMAP::IVMapManager* vmgr = VMAP::VMapFactory::createOrGetVMapManager();
+    VMAP::AreaAndLiquidData vmapData;
+    vmgr->getAreaAndLiquidData(GetId(), x, y, z, reqLiquidType, vmapData);
```
Shauren, line 2918: you need to add `uint32 terrainMapId = PhasingHandler::GetTerrainMapId(phaseShift, this, x, y);` and replace `GetId()` with it

also drop `const` from function signature (been done for all Map functions that use GetGrid)

src/server/game/Maps/Map.cpp
```diff
@@ -2899,6 +2903,103 @@ ZLiquidStatus Map::getLiquidStatus(PhaseShift const& phaseShift, float x, float
...
+    if (vmapData.areaInfo)
+        data.areaInfo = boost::in_place(vmapData.areaInfo->adtId, vmapData.areaInfo->rootId, vmapData.areaInfo->groupId, vmapData.areaInfo->mogpFlags);
+
+    GridMap* gmap = const_cast<Map*>(this)->GetGrid(GetId(), x, y);
```
Shauren, line 2922: `terrainMapId` will go here (and const_cast removal)

src/server/game/Entities/Object/Object.h
```diff
@@ -21,6 +21,7 @@
 #include "Common.h"
 #include "GridReference.h"
 #include "GridRefManager.h"
+#include "Map.h"
```
Shauren, line 24: needs to be removed, `enum LineOfSightChecks` moved to SharedDefines.h (this is a change from include cleanup that was initially done in master then ported to 3.3.5, so it will not be picked anywhere else)

src/server/game/Maps/Map.cpp
```diff
@@ -2899,6 +2903,103 @@ ZLiquidStatus Map::getLiquidStatus(PhaseShift const& phaseShift, float x, float
     return result;
 }
 
+void Map::GetFullTerrainStatusForPosition(float x, float y, float z, PositionFullTerrainStatus& data, uint8 reqLiquidType) const
```
Shauren, line 2914: Needs `PhaseShift const&` arg (as first)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -2471,7 +2421,7 @@ void Player::GiveLevel(uint8 level)
     UpdateAllStats();
 
     if (sWorld->getBoolConfig(CONFIG_ALWAYS_MAXSKILL)) // Max weapon skill when leveling up
-        UpdateSkillsToMaxSkillsForLevel();
+        UpdateWeaponsSkillsToMaxSkillsForLevel();
```
Shauren, line 2424: this commit needs to drop all CONFIG_ALWAYS_MAXSKILL related bits

Shauren:
drop when rebasing
83df323caa448d5914f57b9cfa4d63fa4f850a77

src/server/game/Entities/Object/Object.h
Carbenium: Wouldn't it make sense to drop this when we port the SharedDefines.h commit to master? That's why I didn't bother to find a better temporary place for the enum.

Shauren: But you will not be porting that SharedDefines.h commit, thats why

msoky:
👍

PR #24884, opened 2020-06-24
Core/Player: do not consider FLAG_EXTRA_NO_XP_AT_KILL or pet status when deciding if a creature counts as a "gives experience or honor" target

### LINKED ISSUE #23860: [3.3.5] - Spell/Proc: Proc spells vs CREATURE_FLAG_EXTRA_NO_XP_AT_KILL
**Description:**

https://www.wowhead.com/npc=39190/wicked-spirit

NPC Wicked Spirit of ICC - LK encounter have flags_extra CREATURE_FLAG_EXTRA_NO_XP_AT_KILL
With this flag, doesn't get exp when die, then, can't proc Rapid Killing.

in this log we can see how should proc Rapid Killing after kill Wicked Spirit:
http://www.worldoflogs.com/reports/b0e71guud7hcynf1/xe/?s=9921&e=10899&x=targetName%3D%22Rogerbrown%22+AND+spell+%3D+%22Rapid+Recuperation%22+OR+spell+%3D+%22Rapid+Killing%22+OR+targetName+%3D+%22Wicked+Spirit%22+&page=1
```
[20:41:47.479] Rogerbrown Chimera Shot Wicked Spirit *8000* (O: 7279)
[20:41:47.732] Wicked Spirit dies
[20:41:47.789] Rogerbrown's Rapid Killing fades from Rogerbrown
[20:41:47.789] Rogerbrown's Rapid Recuperation is refreshed by Rogerbrown
[20:41:47.789] Rogerbrown gains Rapid Killing from Rogerbrown
```

**Fix:**

```sql
UPDATE `creature_template` SET `flags_extra`=`flags_extra` &~64 WHERE `entry` IN (39190,39287,39288,39289);
```

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 88317a4ec8a4fae2f914fb699e8fb0fff1da3205

Rothend:
**Changes proposed:**

When a spell/talent/ability says "targets that grant experience or honor", the "experience" part is not to be taken literally. It simply means that the creature must be in the level range where it appears at green/yellow/red "difficulty" (ie: not trivial) for the player.

So it shouldn't take into consideration FLAG_EXTRA_NO_XP_AT_KILL, and those abilities should also work on pets.

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes #23860

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -23843,7 +23843,7 @@ bool Player::isHonorOrXPTarget(Unit* victim) const
 
     if (Creature const* creature = victim->ToCreature())
     {
-        if (!creature->CanGiveExperience())
+        if (creature->IsCritter() || creature->IsTotem())
             return false;
     }
     return true;


ForesterDev:
I think, we need to rename isHonorOrXPTarget() method for a clearer definition.

Rothend:
What name would you suggest?

ForesterDev:
Maybe something like IsProcOnDeathTarget ... @illfated maybe you can help with grammar?

illfated:
Hmm, interesting conundrum. I thought some of these methods were named based on some of the decompiled game client contents, but that is probably only true for a limited amount of methods.

At the moment, it looks to me like there is no obvious connection between the names `isHonorOrXPTarget()` and `IsProcOnDeathTarget()`, so I think I need to look up the existing method details first.

illfated:
> It simply means that the creature must be in the level range where it appears at green/yellow/red "difficulty" (ie: not trivial) for the player.
>

I could be mistaken, but I thought this was the same as the definition of which mobs give XP or honor to the player, unless DB or script flags are set on the targets to block players from receiving XP on kill (or critters not giving XP, even to players level 1-5).

Maybe I am just being too literal in the wrong sense, but in my opinion, the XP part of "targets that grant experience or honor" is meant to be taken literally, except for the separate cases where it is obvious that you shouldn't get XP, either because you are using an extreme weapon or cannon vehicle to kill an endless amount of NPCs, random decorative critters or because the targets are temp summons, quest events or faction guards (hostile PvP).

Rothend:
I kinda think the name is fine? It fits the wording Blizz uses on the spells/talents and it does exactly what is on the tin.

illfated:
Yes, I concur. I don't quite see why `isHonorOrXPTarget()` should be renamed, unless someone could come up with a similar short word combination with obvious clarity and better readability, to replace it. At the moment, I struggle to find anything that is not a conjoined sentence (at least twice as many letters).

jackpoz:
Thanks for the PR :)

PR #24886, opened 2020-06-24
Core/Skill: properly scale current value of hidden skills with player level

### LINKED ISSUE #20979: [3.3.5] Core/Spell: Incorrect tooltip value for spells scaling with level of player
**Description:**
The tooltip of player spells scaling with the level show the value of the spell as if the player is level 0. When using these spells, the correct value calculated with the level of the player is however used. So this bug is only that the formula is incorrect in the tooltip when hovering the spell icon in your Actionbar/Spellbook.

**Current behaviour:**
Spells scaling with player level do not correctly show there value in the tooltip. For example Mind Control 605 with formula 32 + Level * 1.00 will always display 32 in the tooltip instead of 32+level of player. Other example Blood Fury (Racial) orc 20572 formula 6 + Level * 4.00 will always display 6 in the tooltip.

**Expected behaviour:**
The value in the tooltip should be calculated using the level of the player. For example for Blood Fury, the tooltip should display 326 at level 80.

 - First you must have enabled "Enhanced Tooltips" in Interface/Help
 - You can see the bug with an Orc warrior. Even if his level is greater than 1, when you mouse hover the Blood Fury spell, the value displayed is 6 regardless of the level.

**Branch(es):**
3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. bd560fd1609c 2017-11-25 10:08:58 -0300 (3.3.5 branch) (Win64, RelWithDebInfo, Static)

### LINKED ISSUE #24801: [3.3.5] Rogue - Disarm Trap
**Description:**

When you try to disarm a Hunter Trap by right click on trap, you get error msg:
```
Requires Disarm Trap 400
```
If you use spell Disarm Trap and select the trap, you can disarm the trap
If you decreases hunter level you can see how decreases the requirement until you can disarm it if the difference in levels is wide
If you use command .setkill 39 400 on the rogue, you can disarm trap by right click

**Expected behaviour:**

Rogue must be able to disarm the trap by right click
this was working fine on TC rev about 2017

I'm not sure when and where did this break

1. Get a hunter and a rogue
2. put a trap with hunter
3. Use Stealth with rogue and try to right click trap

**Branch(es):**

 3.3.5

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/aeb41e67a92b7cd0c75683a269fb0ca75309e9a0

Rothend:
**Changes proposed:**

There are a lot of skills (among those, one for each race plus three for each class) that aren't visible in the client but are still used to handle things like tooltips whose values scale with player level, or the Rogue skill [Subtlety](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?skill=39) that the client uses when the player right-clicks a Hunter trap.

The current code in Player::UpdateSkillsForLevel() correctly updates the maximum value of those skills, however the minimum value is never updated, hence all those skills (for example [Blood Fury](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=20572) and [Gift of the Naaru](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=28880)) will show the level 1 values on their tooltips.

This PR updates all those skills, minus weapon skills and Lockpicking, to player level * 5, whenever a player logs in or levels up (so there's nothing to manually update in the characters database).

- [x] 3.3.5
- [ ] master

**Issues addressed:**
Closes #20979
Closes #24801

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -5939,7 +5939,7 @@ void Player::UpdateSkillsForLevel()
         if (max != 1)
         {
             /// maximize skill always
-            if (alwaysMaxSkill || rcEntry->Flags & SKILL_FLAG_ALWAYS_MAX_VALUE)
+            if (alwaysMaxSkill || (rcEntry->Flags & SKILL_FLAG_ALWAYS_MAX_VALUE))
             {
                 SetUInt32Value(valueIndex, MAKE_SKILL_VALUE(maxSkill, maxSkill));
                 if (itr->second.uState != SKILL_NEW)
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -5938,8 +5938,8 @@ void Player::UpdateSkillsForLevel()
         /// update only level dependent max skill values
         if (max != 1)
         {
-            // Some skills (for example racial, class skills) have their current and maximum values scale at a rate of 5 skill points for each level
-            if (alwaysMaxSkill || SkillLevelScalesWithPlayerLevel(itr->first))
+            /// maximize skill always
+            if (alwaysMaxSkill || rcEntry->Flags & SKILL_FLAG_ALWAYS_MAX_VALUE)
             {
                 SetUInt32Value(valueIndex, MAKE_SKILL_VALUE(maxSkill, maxSkill));
                 if (itr->second.uState != SKILL_NEW)
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -5938,8 +5938,8 @@ void Player::UpdateSkillsForLevel()
         /// update only level dependent max skill values
         if (max != 1)
         {
-            /// maximize skill always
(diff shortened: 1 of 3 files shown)


Warpten:
Why not just use SKILL_RANGE_LEVEL on skills that need to behave like that and manually update the database? This feels like duplicated logic

Rothend:
I don't understand, skills are saved in DB when the player is saved so it shouldn't be manually updated I think?

I changed the existing code that's used for CONFIG_ALWAYS_MAX_SKILL_FOR_LEVEL to also work for skills that aren't lockpicking/weapon skills (by the point that part of code is reached, UpdateSkillsForLevel() has already filtered out every skill that doesn't depend on player level). Should it be handled somewhere else? I can't think of a better place, please do tell if you have suggestions!

Shauren:
I think what you want here is `SKILL_FLAG_ALWAYS_MAX_VALUE` from SkillRaceClassInfo flags field

Rothend:
You're right, thanks! Had no idea that existed. Right now it's used only in Player::LearnDefaultSkill().

Shauren (approved):

PR #24895, opened 2020-06-25
Core/Unit: disable player skill upgrades in pvp

ipriver:
**Changes proposed:**
image

-  Disabled weapon/defense skill upgrades in pvp if you are attacking player's pet or totems.
-  Disabled defense skill upgrades in pvp when being attacked by a player or pet. 

- [x] 3.3.5
- [ ] master

**Tests performed:** 
Compiled, manually tested

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9987,6 +9987,9 @@ uint32 createProcHitMask(SpellNonMeleeDamage* damageInfo, SpellMissInfo missCond
 
 void Unit::ProcSkillsAndReactives(bool isVictim, Unit* procTarget, uint32 typeMask, uint32 hitMask, WeaponAttackType attType)
 {
+    if (!procTarget)
+        return;
+
     // Player is loaded now - do not allow passive spell casts to proc
     if (GetTypeId() == TYPEID_PLAYER && ToPlayer()->GetSession()->PlayerLoading())
         return;
@@ -9995,7 +9998,12 @@ void Unit::ProcSkillsAndReactives(bool isVictim, Unit* procTarget, uint32 typeMa
     if (typeMask & MELEE_BASED_TRIGGER_MASK && procTarget)
     {
         // Update skills here for players
-        if (GetTypeId() == TYPEID_PLAYER)
+        // only when you are not fighting other players or their pets/totems (pvp)
+        if (GetTypeId() == TYPEID_PLAYER &&
+                procTarget->GetTypeId() != TYPEID_PLAYER &&
+                !(procTarget->IsTotem() && procTarget->ToTotem()->GetOwner()) &&
+                !(procTarget->IsPet() && procTarget->ToPet()->GetOwner())
+           )
         {
             // On melee based hit/miss/resist need update skill (for victim and attacker)
             if (hitMask & (PROC_HIT_NORMAL | PROC_HIT_MISS | PROC_HIT_FULL_RESIST))
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9998,8 +9998,8 @@ void Unit::ProcSkillsAndReactives(bool isVictim, Unit* procTarget, uint32 typeMa
         // only when you are not fighting other players or their pets/totems (pvp)
         if (GetTypeId() == TYPEID_PLAYER &&
                 procTarget->GetTypeId() != TYPEID_PLAYER &&
-                !(procTarget->IsTotem() && procTarget->ToTotem()->GetOwner()->GetControllingPlayer()) &&
-                !(procTarget->IsPet() && procTarget->ToPet()->GetOwner())
+                !(procTarget->IsTotem() && procTarget->ToTotem()->GetOwner()->IsPlayer()) &&
+                !procTarget->IsPet()
            )
         {
             // On melee based hit/miss/resist need update skill (for victim and attacker)
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9987,9 +9987,6 @@ uint32 createProcHitMask(SpellNonMeleeDamage* damageInfo, SpellMissInfo missCond
 
 void Unit::ProcSkillsAndReactives(bool isVictim, Unit* procTarget, uint32 typeMask, uint32 hitMask, WeaponAttackType attType)
 {
-    if (!procTarget)
-        return;
-
     // Player is loaded now - do not allow passive spell casts to proc
     if (GetTypeId() == TYPEID_PLAYER && ToPlayer()->GetSession()->PlayerLoading())
         return;
@@ -10001,7 +9998,7 @@ void Unit::ProcSkillsAndReactives(bool isVictim, Unit* procTarget, uint32 typeMa
         // only when you are not fighting other players or their pets/totems (pvp)
         if (GetTypeId() == TYPEID_PLAYER &&
                 procTarget->GetTypeId() != TYPEID_PLAYER &&
-                !(procTarget->IsTotem() && procTarget->ToTotem()->GetOwner()) &&
+                !(procTarget->IsTotem() && procTarget->ToTotem()->GetOwner()->GetControllingPlayer()) &&
                 !(procTarget->IsPet() && procTarget->ToPet()->GetOwner())
            )
         {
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9987,6 +9987,9 @@ uint32 createProcHitMask(SpellNonMeleeDamage* damageInfo, SpellMissInfo missCond
 
 void Unit::ProcSkillsAndReactives(bool isVictim, Unit* procTarget, uint32 typeMask, uint32 hitMask, WeaponAttackType attType)
 {
+    if (!procTarget)
```
Shauren, line 9990: this change does nothing, possibly only enables code analysis warnings about it being checked twice and always true later in the function

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9995,7 +9998,12 @@ void Unit::ProcSkillsAndReactives(bool isVictim, Unit* procTarget, uint32 typeMa
     if (typeMask & MELEE_BASED_TRIGGER_MASK && procTarget)
     {
         // Update skills here for players
-        if (GetTypeId() == TYPEID_PLAYER)
+        // only when you are not fighting other players or their pets/totems (pvp)
+        if (GetTypeId() == TYPEID_PLAYER &&
+                procTarget->GetTypeId() != TYPEID_PLAYER &&
+                !(procTarget->IsTotem() && procTarget->ToTotem()->GetOwner()) &&
```
Shauren, line 10004: GetOwner checks are pointless here, a totem or pet cannot exist without owher

ipriver: From the source files, GetOwner `Player* GetOwner() const;` Does it mean that any NPC can be passed as a Player pointer? The idea's to check that pet/totem belongs to a player and not to NPC.

src/server/game/Entities/Unit/Unit.cpp
ipriver: it checks if procTarget pointer is not null so that the code below won't cause a segmentation fault.

src/server/game/Entities/Unit/Unit.cpp
Shauren: Totem inherits Minion https://github.com/TrinityCore/TrinityCore/blob/3ac1992afbeb1cd7669b300b8ee4131d39f35595/src/server/game/Entities/Creature/TemporarySummon.h#L76 and thats Unit*, not Player*

`Pet` cannot be created for a creature owner at all, so if you have one you know its owned by a player
* pretty sure totems dont grant weapon skills in any case, not just when owned by a player

src/server/game/Entities/Unit/Unit.cpp
Shauren: As I said, this is already checked 5 lines below

src/server/game/Entities/Unit/Unit.cpp
ipriver: Yeah, you are right about Totems, I gotta check if there's another way to find a player who summoned it.

But can't agree with you about pets:
https://github.com/TrinityCore/TrinityCore/blob/3ac1992afbeb1cd7669b300b8ee4131d39f35595/src/server/game/Entities/Unit/Unit.h#L1656
https://github.com/TrinityCore/TrinityCore/blob/3ac1992afbeb1cd7669b300b8ee4131d39f35595/src/server/game/Entities/Pet/Pet.h#L148

src/server/game/Entities/Unit/Unit.cpp
ipriver: I see it now, somehow missed it

src/server/game/Entities/Unit/Unit.cpp
Shauren: The point I was trying to make here is that no totem should grant weapon skillups, not just player owned one

ipriver:
Thanks for your comments, now Totems should be working only for player, gonna describe it below.

For Totems:
- `IsTotem()` checks if the target is a totem
- `ToTotem()` converts target to a totem 
https://github.com/TrinityCore/TrinityCore/blob/6533de31005fc9680d9f2b61a4d8ea4958fc8edc/src/server/game/Entities/Unit/Unit.h#L1984
> Totem inherits Minion
> 
> https://github.com/TrinityCore/TrinityCore/blob/3ac1992afbeb1cd7669b300b8ee4131d39f35595/src/server/game/Entities/Creature/TemporarySummon.h#L76

- `GetOwner()` returns Unit* and we can use `GetControllingPlayer()` to check if the totem was created by a player

https://github.com/TrinityCore/TrinityCore/blob/a43cceb19cce93fe31e71086145636126593d67b/src/server/game/Entities/Unit/Unit.cpp#L5829-L5839

For Pets:
- `IsPet()` checks if the target is a pet
- `ToPet()` converts target to a pet
-  `GetOwner()` returns Player* for `Pet`
https://github.com/TrinityCore/TrinityCore/blob/a43cceb19cce93fe31e71086145636126593d67b/src/server/game/Entities/Pet/Pet.h#L148

> `Pet` cannot be created for a creature owner at all, so if you have one you know its owned by a player

Anyway I'd like to keep a check that a pet has an owner (who is a `Player`)

> also I'm pretty sure totems don't grant weapon skills in any case, not just when owned by a player

I assume totems that were created by NPC should grant the same skill points as the NPC itself.
Now because of `GetControllingPlayer()` players can gain weapon skills only when attacking totems created by NPC and not when attacking Player's totems.

Checked both cases against Players and NPCs in-game and debugged through the console.

src/server/game/Entities/Unit/Unit.cpp
ipriver: @Shauren I couldn't find any confirmation on the internet about how it really should work with weapon skill leveling proc chance while fighting against NPC totems. Though I feel like in ZF on Classic retail I was able to level up weapon skills by hitting totems, but currently, I don't have a subscription to check that.
If you still want to disable proc chance for everyone except NPC mobs then I could simplify the weapon proc if-check. Otherwise please check my comment below where I described how the code works.

Shauren:
Pets: I care about not adding code that does nothing
`procTarget->IsPet() && procTarget->ToPet()->GetOwner()`
If IsPet() is true then GetOwner will always return a valid Player*

ipriver:
ok, I just bought a subscription to check if totems could increase your weapon skills in Classic. I bet it worked the same way in 3.3.5 retail like in vanilla-classic.

image

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9995,7 +9995,12 @@ void Unit::ProcSkillsAndReactives(bool isVictim, Unit* procTarget, uint32 typeMa
     if (typeMask & MELEE_BASED_TRIGGER_MASK && procTarget)
     {
         // Update skills here for players
-        if (GetTypeId() == TYPEID_PLAYER)
+        // only when you are not fighting other players or their pets/totems (pvp)
+        if (GetTypeId() == TYPEID_PLAYER &&
+                procTarget->GetTypeId() != TYPEID_PLAYER &&
+                !(procTarget->IsTotem() && procTarget->ToTotem()->GetOwner()->GetControllingPlayer()) &&
```
Shauren, line 10001: I would change `GetControllingPlayer()` to just `IsPlayer()`, you probably don't want to have totems suddenly starti granting skills when their owner gets mind controlled by a boss (and I'm not aware of any player-owner pets/minions that can summon totems)

illfated:
Cosmetic PR detail: if you would like to use an old-school PR title for this PR, you could write something like this:

**Core/Entities: disable player skill upgrades in PvP**
 -- or --
**Core/Player: disable player skill upgrades in PvP**
(although this PR does include both the Unit & Player subdirectories)

src/server/game/Entities/Unit/Unit.cpp
ipriver: > you probably don't want to have totems suddenly starti granting skills when their owner gets mind controlled by a boss

This won't be like that because of `GetTypeId() == TYPEID_PLAYER`. 
Anyway, I agree that `IsPlayer()` seems like a more suitable check here, but because `IsPlayer()` doesn't exist for `Unit` I had to change `GetOwner()` to `GetOwnerGUID`.

Shauren: `IsPlayer()` was ported back from master branch recently and now exists in 3.3.5 as well (as member of `Object`), you just can't see it since this PR is based on older commit

ipriver: oh, ok. Rebased to bf62901053d20da790e78394cd9b2a3129d28afa and made commit amend 9e952fde7b770d60b1732f211ff41032a2f54b37.

Shauren (approved):

PR #24907, opened 2020-06-27
Script/Quest: fix (More) Torch Catching

### LINKED ISSUE #24893: [3.3.5] [Quest] Torch Catching - Midsummer Fire Festival
**Description:**
Currently players cannot do the quest **[Torch Catching](https://www.wowhead.com/quest=11657/torch-catching)** since using the provided item **[Unlit Torches](https://www.wowhead.com/item=34833/unlit-torches)** does not trigger the required behavior. Meaning that after the player uses the item, he will not do the throw animation and neither will the torch actually be thrown.

**Branch(es):**
3.3.5

**TC rev. hash/commit:** 
TrinityCore rev. 3ac1992afbeb 2020-06-24 22:53:41 +0200 (3.3.5 branch) (Unix, Release, Static)

**Operating system:** 
Ubuntu 19.04

Rothend:
**Changes proposed:**

The quests Torch Catching ([Alliance](https://www.wowhead.com/quest=11657), [Horde](https://www.wowhead.com/quest=11923)) and More Torch Catching ([Alliance](https://www.wowhead.com/quest=11924), [Horde](https://www.wowhead.com/quest=11925)) aren't completable. The quest item does nothing when used.

Note: apparently there should be some bunny/trigger NPC to use as target for the throwing torch, however I couldn't find anything useful in sniffs. The behavior is achieved by picking a random position to throw the torch in, instead.

- [x] 3.3.5
- [x] master

**Issues addressed:** Closes #24893


src/server/scripts/Spells/spell_holiday.cpp
```diff
@@ -1768,6 +1770,144 @@ class spell_midsummer_torch_catch : public SpellScript
...
+            return;
+
+        uint8 requiredCatches = 0;
+        // Number of required catches depends on quest - 4 for the normal quest, 10 for the daily version
+        if (player->GetQuestStatus(QUEST_TORCH_CATCHING_A) == QUEST_STATUS_INCOMPLETE || player->GetQuestStatus(QUEST_TORCH_CATCHING_H) == QUEST_STATUS_INCOMPLETE)
+            requiredCatches = 3;
+        else if (player->GetQuestStatus(QUEST_MORE_TORCH_CATCHING_A) == QUEST_STATUS_INCOMPLETE || player->GetQuestStatus(QUEST_MORE_TORCH_CATCHING_H) == QUEST_STATUS_INCOMPLETE)
+            requiredCatches = 9;
+
+        if (player->GetAuraCount(SPELL_TORCHES_CAUGHT) >= requiredCatches)
```
jackpoz, line 1869: What should happen if requiredCatches is 0 here ?

Rothend: Having the quest item without one of those quests' status incomplete should be impossible normally. The quest item is removed when the quest is completed. Unless a GM adds it to the player manually, in which case all bets are off I believe. Right now it'd let the quest complete regardless of the aura, which I think is fine for this not-normally-possible case :P

Shauren:
`26188 ([PH] Torch Catching Target Bunny)`

src/server/scripts/Spells/spell_holiday.cpp
jackpoz: I would

jackpoz:

offl:
You can merge it and add comment for future devs to use 26188 ([PH] Torch Catching Target Bunny) instead of picking random position

src/server/scripts/Spells/spell_holiday.cpp
jackpoz: could you please handle the case I mentioned ?

PR #24908, opened 2020-06-27
Core/Gameobjects: Traps shouldn't activate in Sanctuary

Jildor:
**Changes proposed:**

-  Traps shouldn't activate in Sanctuary
-  go to dalaran with a hunter, find a hostile npc with movement and put a trap or freezing arrow at this npc

- [X] 3.3.5
- [ ] master


Keader:
It will not affect the duel area?

Jildor:
nope, you can hit players in duel with traps

Killyana:
This will effect gobs like bonfire that have traps and deals fire damage to players once close?

Jildor:
well, in case of the bonfire I can add that the gameobject has owner and owner is player, but in the case of that particular npc I don't see a way to avoid it without hackfix, so I think I will close the PR :(

Killyana:
Maybe summoned traps must be scripted like pets are, to handle the sanctuary thing

Faq:
I might be wrong, but traps are temp spawns, bonfires not. This could be a check?

Jildor:
bonfire not is a problem, we can check trap owner, the attackable npc can be a problem as Killyana says, so I don't know what will be the right way

PR #24911, opened 2020-06-28
Core/Gameobjects: Traps shouldn't ignore IsImmuneToNPC/IsImmuneToPC and IsInSanctuary rules if have owner

Jildor:
**Changes proposed:**

-  Traps shouldn't ignore IsImmuneToNPC/IsImmuneToPC and IsInSanctuary rules if has owner 
-  #24908 followup
-  go to dalaran and use hunter trap or freezing arrow on a hostile moving npc or player

- [X] 3.3.5
- [ ] master

**Issues addressed:** None I think


jackpoz:
Thanks for the PR :)

PR #24913, opened 2020-06-28
Add locales to hotfix_blob

funjoker:
**Changes proposed:**

-  Added locale column to hotfix_blob

- [ ] 3.3.5
- [x] master

### DIFF
--- a/src/server/game/DataStores/DB2Stores.cpp
+++ b/src/server/game/DataStores/DB2Stores.cpp
@@ -352,6 +352,8 @@ typedef std::vector<TalentEntry const*> TalentsByPosition[MAX_CLASSES][MAX_TALEN
 typedef std::unordered_set<uint32> ToyItemIdsContainer;
 typedef std::tuple<uint16, uint8, int32> WMOAreaTableKey;
 typedef std::map<WMOAreaTableKey, WMOAreaTableEntry const*> WMOAreaTableLookupContainer;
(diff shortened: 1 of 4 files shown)


src/server/game/DataStores/DB2Stores.cpp
```diff
@@ -1438,7 +1437,9 @@ void DB2Manager::LoadHotfixBlob()
         }
 
         int32 recordId = fields[1].GetInt32();
-        _hotfixBlob[std::make_pair(tableHash, recordId)] = fields[2].GetBinary();
+        LocaleConstant locale = GetLocaleByName(fields[2].GetString());
```
Shauren, line 1440: needs to be validated

funjoker: We never validated locale in other loadings too.
Also GetLocaleByName always returns enUS if nothing is found (kinda stupid imo).

Warpten: IIRC you can (and should) use the locale that I think can be specified in worldserver.conf

Shauren: no, that "thing" will be deleted

Shauren: this will get fixed, im changing GetLocaleByName to return TOTAL_LOCALES if invalid string is passed and also adding `IsValidLocale(LocaleConstant)` that checks `locale < TOTAL_LOCALES && locale != LOCALE_none;`

PR #24949, opened 2020-07-05
Core/Pools: Attempt to workaround pools being broken since 2015

### LINKED ISSUE #21331: Core/Pools: Objects in the pool stop spawning

jackpoz:
Attempt n°2

**Changes proposed:**

-  Do not attempt to respawn pooled objects right away as it doesn't work anyway. Instead spawn a different one

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #21331

Steps described at #21331

- [ ] Someone with more knowledge about pools and spawns should test it

### DIFF
--- a/src/server/game/Pools/PoolMgr.cpp
+++ b/src/server/game/Pools/PoolMgr.cpp
@@ -288,7 +288,7 @@ void PoolGroup<T>::SpawnObject(ActivePoolData& spawns, uint32 limit, uint32 trig
 
         if (!EqualChanced.empty() && rolledObjects.empty())
         {
-            std::copy_if(EqualChanced.begin(), EqualChanced.end(), std::back_inserter(rolledObjects), [triggerFrom, &spawns](PoolObject const& object)
+            std::copy_if(EqualChanced.begin(), EqualChanced.end(), std::back_inserter(rolledObjects), [/*triggerFrom, */&spawns](PoolObject const& object)
             {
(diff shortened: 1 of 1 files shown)


jackpoz:
@Killyana could you please test this and see if it fixes anything and/or adds new issues ? Thanks

Killyana:
The only Issue I get is when you have two pooled creature A and B, if you kill A, the next respawn will always be B and same for B once killed the next respawn will always be A, so respawn sequence: . . . A B A B A . . .

jackpoz:
well that's still an improvement over "if you kill A, nothing will respawn" 😄

PR #24968, opened 2020-07-07
.

JustZerooo:
1

### DIFF
--- a/sql/updates/world/master/2020_07_07_00_world.sql
+++ b/sql/updates/world/master/2020_07_07_00_world.sql
@@ -1,2 +1,2 @@
 DELETE FROM `spell_script_names` WHERE `spell_id` = 109132;
-INSERT INTO `world`.`spell_script_names` (`spell_id`, `ScriptName`) VALUES ('109132', 'spell_monk_roll');
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES ('109132', 'spell_monk_roll');
--- a/sql/updates/world/master/2020_07_07_00_world.sql
+++ b/sql/updates/world/master/2020_07_07_00_world.sql
@@ -1,0 +1,2 @@
+DELETE FROM `spell_script_names` WHERE `spell_id` = 109132;
+INSERT INTO `world`.`spell_script_names` (`spell_id`, `ScriptName`) VALUES ('109132', 'spell_monk_roll');
--- a/src/server/scripts/Spells/spell_monk.cpp
+++ b/src/server/scripts/Spells/spell_monk.cpp
@@ -39,6 +39,11 @@ enum MonkSpells
     SPELL_MONK_SOOTHING_MIST                            = 115175,
     SPELL_MONK_STANCE_OF_THE_SPIRITED_CRANE             = 154436,
     SPELL_MONK_SURGING_MIST_HEAL                        = 116995,
+    SPELL_MONK_ROLL                                     = 109132,
+    SPELL_MONK_ROLL_TRIGGER                             = 107427,
+    SPELL_MONK_ITEM_PVP_GLOVES_BONUS                    = 124489,
+    SPELL_MONK_DISABLE                                  = 116095,
+    SPELL_MONK_DISABLE_ROOT                             = 116706,
 };
 
 // 117952 - Crackling Jade Lightning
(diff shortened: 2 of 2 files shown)


sql/updates/world/master/2020_07_07_00_world.sql
```diff
@@ -0,0 +1,2 @@
+DELETE FROM `spell_script_names` WHERE `spell_id` = 109132;
+INSERT INTO `world`.`spell_script_names` (`spell_id`, `ScriptName`) VALUES ('109132', 'spell_monk_roll');
```
Jinnaix, line 2: This query would never apply to a database named anything else than "world" - please adjust this by removing it.

Jinnaix (changes requested):

sql/updates/world/master/2020_07_07_00_world.sql
DerEnderman: No, because it is in world folder and is correct. Sry i dont change

Ovahlord:
You should update the spell script to the modern scripting model to keep the standard for that file going.

DerEnderman:
> 
> 
> You should update the spell script to the modern scripting model to keep the standard for that file going.

Working fine no need to update and not interested. it works and that is all we need if you dont like then dont use

MaxtorCoder:
> 
> 
> > You should update the spell script to the modern scripting model to keep the standard for that file going.
> 
> Working fine no need to update and not interested. it works and that is all we need if you dont like then dont use

It's about that people follow the simple coding standards, if you cannot follow that then why bother contributing in the first place. :)

src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -159,9 +164,63 @@ class spell_monk_provoke : public SpellScript
...
+  public:
+    spell_monk_roll() : SpellScriptLoader("spell_monk_roll") { }
+
+    class spell_monk_roll_SpellScript : public SpellScript
+    {
+        PrepareSpellScript(spell_monk_roll_SpellScript);
+
+        bool Validate(SpellInfo const* /*spellInfo*/) override
+        {
+            return ValidateSpellInfo({ SPELL_MONK_ROLL });
```
Shauren, line 179: You don't need to validate existence of the spell that is being scripted (checked externally in ObjectMgr)
You need to validate that other spells that you use in this script.

src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -159,9 +164,63 @@ class spell_monk_provoke : public SpellScript
...
+        {
+            Aura* aur = GetCaster()->AddAura(SPELL_MONK_ROLL_TRIGGER, GetCaster());
+            if (!aur)
+                return;
+
+            AuraApplication* app = aur->GetApplicationOfTarget(GetCaster()->GetGUID());
+            if (!app)
+                return;
+
+            app->ClientUpdate();
```
Shauren, line 192: Why is this part needed at all?

src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -159,9 +164,63 @@ class spell_monk_provoke : public SpellScript
...
+        }
+
+        void HandleAfterCast()
+        {
+            Unit* caster = GetCaster();
+            if (!caster || caster->GetTypeId() != TYPEID_PLAYER)
```
Shauren, line 198: caster in SpellScript is never nullptr
nothing in this script requires caster to also be a player

src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -159,9 +164,63 @@ class spell_monk_provoke : public SpellScript
     }
 };
 
+// Roll - 109132 or Roll (3 charges) - 121827
```
Shauren, line 167: Why does the comment mention 121827 when that spell is not given the script in database?

src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -159,9 +164,63 @@ class spell_monk_provoke : public SpellScript
...
+        void HandleAfterCast()
+        {
+            Unit* caster = GetCaster();
+            if (!caster || caster->GetTypeId() != TYPEID_PLAYER)
+                return;
+
+            caster->CastSpell(caster, SPELL_MONK_ROLL_TRIGGER, true);
+
+            if (caster->HasAura(SPELL_MONK_ITEM_PVP_GLOVES_BONUS))
+                caster->RemoveAurasByType(SPELL_AURA_MOD_DECREASE_SPEED);
```
Shauren, line 204: Shouldn't this be done by casting 125127 instead of manually removing aura type?

src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -39,6 +39,11 @@ enum MonkSpells
     SPELL_MONK_SOOTHING_MIST                            = 115175,
     SPELL_MONK_STANCE_OF_THE_SPIRITED_CRANE             = 154436,
     SPELL_MONK_SURGING_MIST_HEAL                        = 116995,
+    SPELL_MONK_ROLL                                     = 109132,
+    SPELL_MONK_ROLL_TRIGGER                             = 107427,
+    SPELL_MONK_ITEM_PVP_GLOVES_BONUS                    = 124489,
+    SPELL_MONK_DISABLE                                  = 116095,
```
Shauren, line 45: What do disable spells have to do with roll?

src/server/scripts/Spells/spell_monk.cpp
Shauren: Actually, no item has this effect attached to it, why is it here at all

Shauren (changes requested):

PR #24989, opened 2020-07-10
Core/Player: Limit leveling for expansions

### LINKED ISSUE #24959: Leveling beyond account expansion restrictions
**Description:**

My account is expansion 0 so it doesn't have access to the content beyond vanilla. When I hit level 60 after regularly leveling (7x experience rate) the EXP bar disappeared as expected but I still continued to gain experience and eventually leveled up to 61.

**Expected behaviour:**

Experience gain should stop at the appropriate level for each expansion if the account doesn't have access to the next content.

1. Step 1 Have an account with expansion value 0. ( only vanilla ) 
2. Step 2 Have a character at level 60, supposed max level for expansion value 0.
3. Step 3 Do appropriate level experience granting activities
4. Step 4 Eventually level up beyond 60 without having access to it.

**Branch(es):**

3.3.5

ipriver:
**Changes proposed:**

Added additional functions to check max available level based on account expansion and worldserver config value. 
Changes don't affect accounts that have default expansions or custom MaxLevel configs.

-  Disabled XP for killing mobs, exploring, and completing quests.
-  Players at max level now should correctly receive additional gold for completing quests. 
-  It also enables arena team registration for these players.
-  Pets already are bound to players level, so there's no workaround needed there.

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #24959

**Tests performed:** 

Build works, tested in-game

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -2538,7 +2538,7 @@ void Player::GiveXP(uint32 xp, Unit* victim, float group_rate)
     sScriptMgr->OnGivePlayerXP(this, xp, victim);
 
     // XP to money conversion processed in Player::RewardQuest
-    if (level >= sWorld->getIntConfig(CONFIG_MAX_PLAYER_LEVEL))
+    if (IsMaxLevel())
         return;
 
     uint32 bonus_xp;
@@ -2555,11 +2555,11 @@ void Player::GiveXP(uint32 xp, Unit* victim, float group_rate)
     uint32 nextLvlXP = GetUInt32Value(PLAYER_NEXT_LEVEL_XP);
     uint32 newXP = GetXP() + xp + bonus_xp;
 
-    while (newXP >= nextLvlXP && level < sWorld->getIntConfig(CONFIG_MAX_PLAYER_LEVEL))
+    while (newXP >= nextLvlXP && !IsMaxLevel())
     {
         newXP -= nextLvlXP;
 
-        if (level < sWorld->getIntConfig(CONFIG_MAX_PLAYER_LEVEL))
+        if (!IsMaxLevel())
             GiveLevel(level + 1);
 
         level = GetLevel();
@@ -6474,7 +6474,7 @@ void Player::CheckAreaExploreAndOutdoor()
 
         if (areaEntry->area_level > 0)
         {
-            if (GetLevel() >= sWorld->getIntConfig(CONFIG_MAX_PLAYER_LEVEL))
+            if (IsMaxLevel())
             {
                 SendExplorationExperience(areaId, 0);
             }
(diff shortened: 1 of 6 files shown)


src/server/game/Entities/Player/Player.h
```diff
@@ -2499,6 +2501,8 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         uint32 healthBeforeDuel;
         uint32 manaBeforeDuel;
 
+        uint8 _acc_expansion;
```
Shauren, line 2504: This addition is useless, only duplicates data

src/server/game/Entities/Player/Player.cpp
```diff
@@ -2665,6 +2667,31 @@ void Player::GiveLevel(uint8 level)
...
+{
+    uint8 max_lvl = GetMaxLevelForExpansion(_acc_expansion);
+
+    // if _acc_expansion stores some undefined value, GetMaxLevelForExpansion
+    // returns 0 and it should be treated like the latest expansion
+    // with deafult max lvl
+    return max_lvl != 0 ? max_lvl : DEFAULT_MAX_LEVEL;
+}
+
+bool Player::IsMaxLevel() const
```
Shauren, line 2680: Replace the entire function body with just `return GetLevel() >= GetUInt32Value(PLAYER_FIELD_MAX_LEVEL);`

src/server/game/Entities/Player/Player.cpp
```diff
@@ -2665,6 +2667,31 @@ void Player::GiveLevel(uint8 level)
...
 }
 
+uint8 Player::GetExpansionLevelLimit() const
+{
+    uint8 max_lvl = GetMaxLevelForExpansion(_acc_expansion);
+
+    // if _acc_expansion stores some undefined value, GetMaxLevelForExpansion
+    // returns 0 and it should be treated like the latest expansion
+    // with deafult max lvl
+    return max_lvl != 0 ? max_lvl : DEFAULT_MAX_LEVEL;
```
Shauren, line 2677: Don't do weird corrections of returned value here, change GetMaxLevelForExpansion to return DEFAULT_MAX_LEVEL instead (and remove this function)

Shauren (changes requested):
Finally, in `Player::InitStatsForLevel` change PLAYER_FIELD_MAX_LEVEL to `GetMaxLevelForExpansion(GetSession()->Expansion())` instead of reading config

src/server/game/Entities/Player/Player.cpp
ipriver: good catch

src/server/game/Entities/Player/Player.cpp
ipriver: > change PLAYER_FIELD_MAX_LEVEL to GetMaxLevelForExpansion(GetSession()->Expansion())

I like the idea of this variable as it solves the reason why I added `uint8 _acc_expansion` to exclude extra function calls. 
But it won't work simply like that as it requires checks between `GetMaxLevelForExpansion` and `MaxPlayerLevel` in the config.

Shauren: Avoiding extra non virtual function calls that simply read one field on a non hot code path?

Shauren (approved):

Issue #11877, opened 2014-04-12
[Scripts/Onyxia] Onyxia errors in third phase


MitchesD:
Rev.: 870d5d9c45ddf8331f7029072f44f4b1cfa3b54b
TDB 335.53

```
http://s13.postimg.org/k1gpgsqxz/Wo_WScrn_Shot_041114_235342.jpg
```

DBErrors spam:

```
Entry 426151096 not found in `trinity_string` table.
```
##

illfated:
Ref. closed issue #23789 .

RR2739-VCN:
And the issue again dissapears and becomes dormant. Shame, TC needs to stop closing issues frim 2019 by referencing issues from 2014. I mean wtf is the point? So i reported this issue and TC closed it and referenced this old one. Whats the point? That is why i dont report anymore because Akro closes everything and loves to reference 6 year old open issues.

Keader:
@RR2739-VCN because has a open issue about same issue already.
And already has more information in old issue.

RR2739-VCN:
I agree merging issues is a great idea. However closing it and saying " hey someone opened the same issue 5 years ago, closed" is not the right way to go about it. I believe merging issues is the right way to go about it.

Aokromes:
contributing guidelines states to search before posting new issues writing onyxia shows this same ticket on 1st possition.

RR2739-VCN:
Contributing becomes simply tediouswhen you close issues and referencing issues from 5 years ago. You would be correct the issue reported already exists in recent reports but you close issues and referencing 5 years old FORGOTTEN issue sweeping new reports under a rug. Stop closing reports from the community and stop referencing 5 years old issues and do as @Keader said, merge them. Away you approach this is incorrect. Once again, rethink your strategy here abd do what people suggested "merge , dont close"

Killyana:
No need for any condition as this traps are listed in the cpp.

xvwyh:
Have you tested what would happen if there are any player-spawned gameobjects around Onyxia? They would also get "activated".

xvwyh:
The cause of this issue is invalid memory reads. The horrific mess that is SPELL_EFFECT_ACTIVATE_OBJECT creates a temporary object on the stack and passes a reference to it to MapScripts handler, but MapScripts handler can save a pointer to it and execute it *later*. By that time the object on the stack is already destroyed and the pointer dereferences into garbage, which is what's causing these weird error messages about non-existent commands and values.

[...]

PR #24997, opened 2020-07-11
Core/Spells: Implement SPELL_EFFECT_ACTIVATE_OBJECT. (#23)

jackpoz:
Core/Spells: Implement SPELL_EFFECT_ACTIVATE_OBJECT.

- [x] 3.3.5
- [ ] master

None

- [ ] Scripts that override SPELL_EFFECT_ACTIVATE_OBJECT should be checked if they are still needed at all

### DIFF
--- a/sql/updates/world/3.3.5/2020_99_99_99_world.sql
+++ b/sql/updates/world/3.3.5/2020_99_99_99_world.sql
@@ -3,8 +3,7 @@ ALTER TABLE `gameobject_template_addon`
 	ADD COLUMN `artkit0` INT NOT NULL DEFAULT 0 AFTER `maxgold`,
 	ADD COLUMN `artkit1` INT NOT NULL DEFAULT 0 AFTER `artkit0`,
 	ADD COLUMN `artkit2` INT NOT NULL DEFAULT 0 AFTER `artkit1`,
-	ADD COLUMN `artkit3` INT NOT NULL DEFAULT 0 AFTER `artkit2`,
-	ADD COLUMN `artkit4` INT NOT NULL DEFAULT 0 AFTER `artkit3`;
+	ADD COLUMN `artkit3` INT NOT NULL DEFAULT 0 AFTER `artkit2`;
 
 -- Note: All of these should be targettable by spells 46904 and 46903, but conditions are only set for Stormwind (damn Horde fanatics)
 UPDATE gameobject_template SET artkit0 = 121, arkit1 = 122 WHERE entry IN (
--- a/sql/updates/world/3.3.5/2020_99_99_99_world.sql
+++ b/sql/updates/world/3.3.5/2020_99_99_99_world.sql
@@ -5,8 +5,8 @@ ALTER TABLE `gameobject_template_addon`
 	ADD COLUMN `artkit2` INT NOT NULL DEFAULT 0 AFTER `artkit1`,
 	ADD COLUMN `artkit3` INT NOT NULL DEFAULT 0 AFTER `artkit2`;
 
--- Note: All of these should be targettable by spells 46904 and 46903, but conditions are only set for Stormwind (damn Horde fanatics)
-UPDATE gameobject_template SET artkit0 = 121, arkit1 = 122 WHERE entry IN (
+-- Note: All of these should be targetable by spells 46904 and 46903, but conditions are only set for Stormwind (damn Horde fanatics)
+UPDATE gameobject_template_addon SET artkit0 = 121, arkit1 = 122 WHERE entry IN (
 	188352, -- Flame of Shattrath
 	188129, -- Flame of Silvermoon
 	188128, -- Flame of the Exodar
--- a/sql/updates/world/3.3.5/2020_99_99_99_world.sql
+++ b/sql/updates/world/3.3.5/2020_99_99_99_world.sql
@@ -1,0 +1,42 @@
+--
+ALTER TABLE `gameobject_template_addon`
+	ADD COLUMN `artkit0` INT NOT NULL DEFAULT 0 AFTER `maxgold`,
+	ADD COLUMN `artkit1` INT NOT NULL DEFAULT 0 AFTER `artkit0`,
+	ADD COLUMN `artkit2` INT NOT NULL DEFAULT 0 AFTER `artkit1`,
+	ADD COLUMN `artkit3` INT NOT NULL DEFAULT 0 AFTER `artkit2`,
+	ADD COLUMN `artkit4` INT NOT NULL DEFAULT 0 AFTER `artkit3`;
+
+-- Note: All of these should be targettable by spells 46904 and 46903, but conditions are only set for Stormwind (damn Horde fanatics)
+UPDATE gameobject_template SET artkit0 = 121, arkit1 = 122 WHERE entry IN (
+	188352, -- Flame of Shattrath
+	188129, -- Flame of Silvermoon
+	188128, -- Flame of the Exodar
+	181567, -- Flame of the Wetlands
+	181566, -- Flame of Hillsbrad
+	181565, -- Flame of Westfall
+	181564, -- Flame of Silverpine
+	181563, -- Flame of Darkshore
+	181562, -- Flame of Stonetalon
+	181561, -- Flame of Ashenvale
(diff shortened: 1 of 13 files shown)


sql/updates/world/3.3.5/2020_99_99_99_world.sql
```diff
@@ -0,0 +1,42 @@
+--
+ALTER TABLE `gameobject_template_addon`
+	ADD COLUMN `artkit0` INT NOT NULL DEFAULT 0 AFTER `maxgold`,
+	ADD COLUMN `artkit1` INT NOT NULL DEFAULT 0 AFTER `artkit0`,
+	ADD COLUMN `artkit2` INT NOT NULL DEFAULT 0 AFTER `artkit1`,
+	ADD COLUMN `artkit3` INT NOT NULL DEFAULT 0 AFTER `artkit2`,
+	ADD COLUMN `artkit4` INT NOT NULL DEFAULT 0 AFTER `artkit3`;
```
Warpten, line 7: No `artKit4` for 3.3.5

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -1432,6 +1432,32 @@ void GameObject::UseDoorOrButton(uint32 time_to_restore, bool alternative /* = f
     m_cooldownTime = time_to_restore ? (GameTime::GetGameTimeMS() + time_to_restore) : 0;
 }
 
+void GameObject::PlayAnimKit(int32 animKit)
+{
+    WorldPacket data(SMSG_GAME_OBJECT_ACTIVATE_ANIM_KIT, 4 + 2);
```
Warpten, line 1437: Opcode does not exist in 3.3.5.

Anim kit is set via `GAMEOBJECT_BYTES_1[1]`

src/server/game/Entities/GameObject/GameObjectData.h
```diff
@@ -669,4 +672,47 @@ struct GameObjectData : public SpawnData
...
+    Despawn,                        // Despawn
+    MakeInert,                      // Make Inert                       // Disables interactions
+    MakeActive,                     // Make Active                      // Enables interactions
+    CloseAndLock,                   // Close + Lock                     // Sets GO_STATE_READY and sets GO_FLAG_LOCKED
+    UseArtKit0,                     // Use ArtKit0                      // 46904: 121
+    UseArtKit1,                     // Use ArtKit1                      // 36639: 81, 46903: 122
+    UseArtKit2,                     // Use ArtKit2
+    UseArtKit3,                     // Use ArtKit3
+    SetTapList,                     // Set Tap List
+    GoTo1stFloor,                   // Go to 1st floor
```
Warpten, line 702: `SetTapList` is the last enumeration value

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -7671,8 +7671,8 @@ void ObjectMgr::LoadGameObjectTemplateAddons()
 {
     uint32 oldMSTime = getMSTime();
 
-    //                                                0       1       2      3        4
-    QueryResult result = WorldDatabase.Query("SELECT entry, faction, flags, mingold, maxgold FROM gameobject_template_addon");
+    //                                                0       1       2      3        4       5        6        7        8        9
+    QueryResult result = WorldDatabase.Query("SELECT entry, faction, flags, mingold, maxgold, artkit0, artkit1, artkit2, artkit3, artkit4 FROM gameobject_template_addon");
```
Warpten, line 7675: `artKit4` gone

Warpten (changes requested):

Warpten:
Related issue: #21078

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -7717,6 +7720,20 @@ void ObjectMgr::LoadGameObjectTemplateAddons()
             }
         }
 
+        uint32 artKitIndex = 0;
+        for (uint32& artKitID : gameObjectAddon.artKits)
```
Shauren, line 7724: This code is stupid, you use range loop syntax to avoid having to introduce a separate index variable... but then you have that variable anyway because you need it for logging, just use classic `for (std::size_t i = 0; i < gameObjectAddon.artKits.size(); ...`

sql/updates/world/3.3.5/2020_99_99_99_world.sql
```diff
@@ -0,0 +1,42 @@
...
+
+-- Note: All of these should be targettable by spells 46904 and 46903, but conditions are only set for Stormwind (damn Horde fanatics)
```
illfated, line 9: targetable or TargetTable?

src/server/game/Entities/GameObject/GameObject.cpp
Shauren: No, since the script action enum for 3.3.5 ends before anim kits, this function isnt needed at all

src/server/game/Globals/ObjectMgr.cpp
Shauren: Also why isnt it just checked in the same loop as assignment from query result?

src/server/game/Entities/GameObject/GameObject.cpp
Warpten: Yes i'm retarded and keep getting confused between art and anim

src/server/game/Globals/ObjectMgr.cpp
Warpten: /shrug I just liked having related errors together

src/server/game/Entities/GameObject/GameObjectData.h
```diff
@@ -669,4 +672,33 @@ struct GameObjectData : public SpawnData
     uint8 artKit = 0;
 };
 
+enum class GameObjectActions : uint32
+{
+    // Name from client executable      // Comments
```
xvwyh, line 677: If you don't mind, please indent this line for it to be aligned with the comments on the subsequent lines. It's the first column of the comments that's meant to be "Name from client executable".

xvwyh:
Pulling the definition of enum GameObjectActions to GameObjectData.h is indeed a good idea, because these actions can be used on more places than just this spell effect. The entire switch statement should be extracted into its own method in GameObject too, frankly, because Blizzard also uses them in their scripts: for [ai actions](https://docs.google.com/spreadsheets/d/1Wh1e1ZXYM-sw3wWrY9dGZ2ngK8oa3PA2XIftQJuf5R0/edit#gid=740715654&range=H41), [world state actions](https://docs.google.com/spreadsheets/d/1Wh1e1ZXYM-sw3wWrY9dGZ2ngK8oa3PA2XIftQJuf5R0/edit#gid=663503057&range=C38), and for events triggered on gameobjects ("Step 4" [here](https://github.com/TrinityCore/TrinityCore/issues/21078#issuecomment-420115205)).

xvwyh:
Oh, and to answer the question @Warpten left in the code:

> // What's the difference with Open?

I don't know anymore. Initially "Disturb" was used to trigger traps - cause them to cast their spell, play anim, change state, and despawn if they have limited charges. "Open" was used to change state of other GO types. But as I was going through all spells with that effect to set up implicit targeting conditions for them, I noticed some that were clearly meant to interact with traps, but now they used "Open". Perhaps "Open" also performed the same actions as "Distub" on retail, just like our `GameObject::Use`, some designers used "Open" because it sounded more reasonable than a weirdly names "Disturb", found it to work, and left it that way...

xvwyh:
Actually, double-checking it now, I only have 1 spell with "Open" that's targeting a trap: spell 28278 targeting GO 181136. Which does have a spell that casts when the trap is disturbed... But then again, maybe they could've just made the GO cast a spell from a script in reaction to event "Open", and the GO properly remains unused... Anyway, bottom line is - I don't know, not enough info to make a definitive statement on that.

jackpoz:
@Killyana could you test if this fixes https://github.com/TrinityCore/TrinityCore/issues/11877 ?

Warpten (approved):

jackpoz:
we'll test in production like usual 😃

PR #25050, opened 2020-07-15
Core/Random: Replace Boost TLS with C++11 one

Carbenium:
**Changes proposed:**

-  Core/Utilities: Add a wrapper for lazy initialization
-  Core/Random: Replace Boost TLS with C++11 one

- [x] 3.3.5
- [ ] master

**Tests performed:** build + debug session with 2 map threads confirmed, that SFMT is initialized per thread

Credits go to @jackpoz and @Naios (ref #15782)

### DIFF
--- a/src/common/Utilities/Random.cpp
+++ b/src/common/Utilities/Random.cpp
@@ -18,24 +18,11 @@
 #include "Random.h"
 #include "Errors.h"
 #include "SFMTRand.h"
-#include <boost/thread/tss.hpp>
+#include "Util.h"
 #include <random>
 
-static boost::thread_specific_ptr<SFMTRand> sfmtRand;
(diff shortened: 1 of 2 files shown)


Shauren:
@jackpoz might have something to say here about thread_local messing with his c++/cli stuff for botfarm

Carbenium:
As far as I understood from the original PR, boost TSS was messing with it. Let's see what he has to say.

jackpoz:
Both boost tss and std tls cause the same issue in C++/CLI so I just commented out the whole Random.cpp code (and never reopened the PR as I forgot).

Looks good to me 👍

jackpoz:
We might end up merging https://github.com/TrinityCore/TrinityCore/pull/15782 instead

Issue #25006, opened 2020-07-12
Investigate the usage of "[[fallthrough]]"


Carbenium:
**Description:**

The currently used method to mark implicit falltroughs in switch statements is a comment of the form
``` /* fallthrough */``` or ```/* fallthrough - reason*/```.

This works well for GCC since it is a supported method to surpress ```-Wimplicit-fallthrough``` warnings (see [Docs](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)).
Clang has no support for these types of comments, but relies on the ```[[fallthrough]]``` attribute (see [Docs](https://clang.llvm.org/docs/AttributeReference.html#fallthrough)).

Investigate/discuss whether we should switch to the standardized ```[[fallthrough]]``` attribute.

**Pros :**

- better compiler support (GCC + clang)
- future-proof since standardized

**Branch(es):**
both

**TC rev. hash/commit:** c92950b3e1f6366d85d707365a8ad2caddafeecc

jackpoz:
I would like every fallthrough to have a comment about why it's there,  whether it's in the attribute or as comment

Carbenium:
First step of converting existing fallthrough marks is handled in https://github.com/TrinityCore/TrinityCore/pull/25054.
clang though is a bit pickier than GCC, so some additional warnings popped up: #25055

When those are fixed we can enable `-Wimplicit-fallthrough` for clang as well.

Carbenium:
85b5b842ca6c05d4e51081e6c3282940a80f3761
e6e5775be83d07c3221d6690f11184246bec4bf1

PR #25054, opened 2020-07-15
Misc: Use [[fallthrough]] attribute instead of comment to mark intentional fallthroughs

Carbenium:
**Changes proposed:**

-  Misc: Use [[fallthrough]] attribute instead of comment to mark intentional fallthroughs

- [x] 3.3.5
- [ ] master

**Issues addressed:** Related: #25006 

**Tests performed:** build

### DIFF
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -545,8 +545,8 @@ bool MySQLConnection::_HandleMySQLErrno(uint32 errNo, uint8 attempts /*= 5*/)
                 mysql_close(m_Mysql);
                 m_Mysql = nullptr;
             }
+            [[fallthrough]];
         }
-        /* fallthrough */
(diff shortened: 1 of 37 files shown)


src/server/database/Database/MySQLConnection.cpp
```diff
@@ -545,8 +545,8 @@ bool MySQLConnection::_HandleMySQLErrno(uint32 errNo, uint8 attempts /*= 5*/)
                 mysql_close(m_Mysql);
                 m_Mysql = nullptr;
             }
+            [[fallthrough]];
```
Shauren, line 548: Does this work? GCC complained if the comment wasnt placed right before next case label

Carbenium: Got a clean build on GCC 10 and clang 10 (with the changes from #25056 applied). So at least works for me (TM). CI also doesn't seem to mind.

PR #25056, opened 2020-07-15
Build: Enable -Wimplicit-fallthrough on clang

### LINKED ISSUE #25006: Investigate the usage of "[[fallthrough]]"

Carbenium:
Depends on #25054

**Changes proposed:**

-  Misc: Add some missing breaks (no-ops) to satisfy clang
-  Build: Enable -Wimplicit-fallthrough on clang

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #25006 and #25055

**Tests performed:** build on clang

### DIFF
--- a/src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
+++ b/src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
@@ -223,7 +223,7 @@ class boss_kologarn : public CreatureScript
                         break;
                     case NPC_RUBBLE:
                         summons.Summon(summon);
-                        // absence of break intended
+                        [[fallthrough]];
                     default:
                         return;
                 }
--- a/cmake/compiler/clang/settings.cmake
+++ b/cmake/compiler/clang/settings.cmake
@@ -9,6 +9,7 @@ if(WITH_WARNINGS)
       -W
       -Wall
       -Wextra
+      -Wimplicit-fallthrough
       -Winit-self
       -Wfatal-errors
       -Wno-mismatched-tags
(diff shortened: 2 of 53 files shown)


src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
```diff
@@ -223,7 +223,7 @@ class boss_kologarn : public CreatureScript
                         break;
                     case NPC_RUBBLE:
                         summons.Summon(summon);
-                        // absence of break intended
+                        [[fallthrough]];
```
Faq, line 226: isnt this better to be includes in #25054 ?

Carbenium: True. I was just lazy :) Let's fix that.

Carbenium: Fixed.

PR #25066, opened 2020-07-17
Core/Unit: Allow gaining Weapon Skill when hit is missed with ranged weapon

Jildor:
**Changes proposed:**

-  with this change: https://github.com/TrinityCore/TrinityCore/commit/e20e51a3855595817df89ac69494b9fd41e1acb3 if you miss hit with a ranged weapon, you can't gain skill, Player::UpdateWeaponSkill shouldn't return if !victim

**Target branch(es):** 3.3.5

- [X] 3.3.5
- [ ] master

**Issues addressed:** None

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -5837,8 +5837,6 @@ bool Player::UpdateSkillPro(uint16 SkillId, int32 Chance, uint32 step)
 void Player::UpdateWeaponSkill(WeaponAttackType attType)
 {
     Unit* victim = GetVictim();
-    if (!victim)
-        return;
 
     if (IsInFeralForm())
         return;                                             // always maximized SKILL_FERAL_COMBAT in fact


ipriver:
Yeah, I see that it doesn't update skills for ranged weapons though not only when you miss but also when you hit.

Previously there was a `victim && victim->GetTypeId() == TYPEID_PLAYER` check that I moved to another funciton. I don't think that removing `!victim` check is a correct fix here.
If the removal of this check allows gaining ranged weapon skill that means that when you hit with a ranged weapon the victim is a `nullptr`. I feel like a problem could be there.

Jildor:
why? allways was working without !victim check
the only that needs check victim is: https://github.com/TrinityCore/TrinityCore/blob/692c0686ce731015f8309bda916d0102e724af43/src/server/game/Entities/Player/Player.cpp#L5847

and is checked with victim && ... I don't see the problem

ipriver:
https://github.com/TrinityCore/TrinityCore/blob/692c0686ce731015f8309bda916d0102e724af43/src/server/game/Entities/Player/Player.cpp#L5847
> 
> and is checked with victim && ... I don't see the problem

Because you will be always gaining weapon skills for targets that `CREATURE_FLAG_EXTRA_NO_SKILLGAIN` when you are using ranged weapons.

Jildor:
So that null target with ranged weapon should be investigated and find a solution to that in another way, but I still think that this change should be applied

ipriver:
As a **quick-fix** to enable ranged weapons leveling, your fix should work, though it doesn't fix the origin problem that has been revealed.

I tried to find the way to fix `nullptr` value for the victim:

https://github.com/TrinityCore/TrinityCore/blob/00b16992f1ae3bf14ab5fe6366028a2b8648bfa0/src/server/game/Entities/Unit/Unit.h#L814

`m_attacking` is set to `nullptr` at the initialization and when you start attacking someone `bool Unit::Attack(Unit* victim, bool meleeAttack)` function is being called

https://github.com/TrinityCore/TrinityCore/blob/00b16992f1ae3bf14ab5fe6366028a2b8648bfa0/src/server/game/Entities/Unit/Unit.cpp#L5562-L5566

and is resetted when you stop attacking via `bool Unit::AttackStop()`

https://github.com/TrinityCore/TrinityCore/blob/00b16992f1ae3bf14ab5fe6366028a2b8648bfa0/src/server/game/Entities/Unit/Unit.cpp#L5611-L5617

And these functions only being called for Melee attacks, that's why `m_attacking` is always `nullptr` for ranged weapons and wands.

Some more experienced TC developers probably already knew that Ranged attacks and Wand attacks work like spells and have different target methods. I got confused and haven't found any good/simple solution to integrate into how Spells can update `m_attacking` so that I decided to write what I've found and maybe it could help someone.


https://github.com/TrinityCore/TrinityCore/blob/00b16992f1ae3bf14ab5fe6366028a2b8648bfa0/src/server/game/Entities/Unit/Unit.cpp#L2875-L2916

Jildor:
That is what I am trying to say, the problem should be investigated, but, that !target does not seem the most appropriate, so I consider that leaving that part as it was and investigating the other old problem in a new issue calmly is the most coherent.

Shauren:
0b616e4865f89b1d15ae20dccd83825cddb5c288

Issue #23799, opened 2019-09-15
Boss/Script: Rotface


Killyana:
**Description:** After Rotface encounter players will remains in combat with Entry: 37013
Puddle Stalker

**Steps to reproduce the problem:**

1. .go c id 36627
2. start the combat and go inside the green goo
3. Kill the boss
4. You will remain in combat with some triggers:
Combat refs: (Combat state: 1 | Manager state: 1)
[PvE] Puddle Stalker (SpawnID 201653)

**Branch(es):** 3.3.5

**TC rev. hash/commit:** rev. 215a6cee2489

RR2739-VCN:
Confirmed. You will remain in combat until you relog or gm clears your combat.

illfated:
Is this regardless of normal/heroic and 10/25 man modes?

OcelotGear:
@illfated yes

Jildor:
Then should be fixed like kologarn https://github.com/TrinityCore/TrinityCore/pull/23866

OcelotGear:
With that method if you enter the flood when the boss has already died, you turn back into combat.

Jildor:
So those triggers should be despawned when rotface die?

Killyana:
If you kill Rotface and get stuck in combat with this triggers, go to Festergut and kill it and it will stop the combat, something in Festergut script or Putricide script stop the combat with this npcs triggers once Festergut dies.

jackpoz:
are those Puddle Stalkers supposed to be despawned or be passive after Rotface dies or ?

offl:
We need to check sniffs, usually we don't despawn triggers and other invisible helpers

Shauren:
These triggers are permanently spawned on retail (before and after fight) but they should never enter combat themselves

Jildor:
related with: #24156 and #24290
also can happens by example with temp triggers summons like ICC - lanathel Swarming Shadows.
before combat/threat rewrite those NPCs didn't get stuck in combat, I don't know what can be the best way to fix this.

jackpoz:
> These triggers are permanently spawned on retail (before and after fight) but they should never enter combat themselves

ah right, that's due to the combat rewrite, right ? a lot of things entering combat when they shouldn't.

jackpoz:
Please test https://github.com/TrinityCore/TrinityCore/pull/25086

PR #25086, opened 2020-07-20
Core/Combat: Disable triggers entering combat

### LINKED ISSUE #24290: stuck on combat with icecrown citadel traps (for sure it needs generic fix)
**Description:**

people get stuck on combat if they walk on the traps when enabled and they don't die.

**Current behaviour:**

you end stuck on combat.

**Expected behaviour:**

you don't enter combat with them

1. .go xyz 4228.416504 2768.950195 350.963074 631
2. .cheat god on
3. .gm off
3. walk around enabled traps

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

70569ae3e66e

**Operating system:** debian 10 x64

jackpoz:
**Changes proposed:**

-  Core/Combat: Disable triggers entering combat by adding a new creature_template.flags_extra value 0x00002000 CREATURE_FLAG_EXTRA_NO_COMBAT

- [x] 3.3.5
- [ ] master

**Issues addressed:** Closes #24290 , #23799

- Tested #24290
- Tested #23799

- [ ] List the triggers that need to be disallowed to enter combat


Jildor:
this can break things like halion controller: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/ChamberOfAspects/RubySanctum/boss_halion.cpp#L608-L638

and blood prince council controller: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/IcecrownCitadel/boss_blood_prince_council.cpp#L258-L307

and probably valithria controller: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/IcecrownCitadel/boss_valithria_dreamwalker.cpp#L565

I'm not sure how many triggers like this can be, if we can't find a way for get all controller and assing a custom flag, a lot of boss encounters and maybe quests (I'm not sure), can break.

Another safer way would be to make a custom flag that would avoid entering combat, but it would have to be assigned to many triggers.

Keader:
Old Behavior: Triggers can enter combat and remove combat like any other npc.
So 90% of scripts is based on that.
After treeston threat rewrite, triggers cant automatic enter in combat (you need force they enter in combat and you need force they  remove combat).
I belive that both issues here, is caused by some negative spell (casted by trigger) put they in combat and nothing remove (caused by treeston rewrite).
To merge this issue, need check all current scripts that use triggers to control fights.
We have Illidari Council (Black Temple) and Yogg Saron (Ulduar), with this behavior too.
Anyway... if i remember right, has a combat trigger (that keep players in combat with him) in Northrend Beasts (TOC) and Reliquary of Souls/Lost (Black Temple) and they're blizzlike keep players in combat while fighting.

jackpoz:
Do those triggers use spells to put players in combat or scripts ? We could skip entering combat when casting spells for triggers.

Ovahlord:
It varies. There are some trigger npcs that engage players via aggro creator spell but there are also simple trigger npcs that use the   combat pulse mechanic of bosses.

jackpoz:
I would rather have triggers not enter combat by default and allow that with a flag for the ones that need to do that (with the AI taking care of adding/removing the flag as needed)

Warpten:
Summoned triggers probably enter combat as they are summoned since they usually despawn on encounter end

jackpoz:
Implemented a new creature_template.flags_extra value CREATURE_FLAG_EXTRA_NO_COMBAT 0x00002000 that doesn't allow the creature to enter combat.
This can be added to all triggers with a sql query, or only to the triggers we want if we get a list. For now only the triggers in the 2 known issues that this PR addresses are handled (please list more)

offl:
In case anyone wants to add this flag to more creatures, use https://gist.github.com/Kinzcool/e81137dab3bc828631ada717f7da5159
And
```sql
SELECT * FROM creature_template a LEFT JOIN dbc_creaturedifficulty b ON a.entry=b.creatureid WHERE b.flags1 & 0x02000000 /*CREATURE_DIFFICULTYFLAGS_IGNORE_COMBAT*/
```
Note: some of them may be not valid at all or can broke some things

sql/updates/world/3.3.5/9999_99_99_99_world.sql
```diff
@@ -0,0 +1,5 @@
+--
+UPDATE creature_template SET flags_extra = flags_extra | 0x00002000 /* CREATURE_FLAG_EXTRA_NO_COMBAT */ WHERE entry IN
+(37744 -- ICC Frost Freeze Trap
+,37013 -- ICC Rotface Puddle Stalker
```
Jildor, line 4:
```diff
+,30648 -- Obsidian Sanctum Fire Cyclone
```

offl: I think we can open an issue with help-wanted label and ask people for help to add this flag to more creatures, I can add it mostly to TBC without screwing up everything

jackpoz: On top of that, we don't have to find all creatures before pushing the sql update. Even if we add 1000 sql updates, as they are applied automatically by worldserver it's all fine

PR #25090, opened 2020-07-21
Core/Scripts: Fix timing issue of EVENT_FLY_TO_RANDOM_PILLAR during Illidan encounter

Carbenium:
The intention here is to add additional 30 seconds of delay to the current delay of EVENT_FLY_TO_RANDOM_PILLAR.
Instead of EventMap::GetNextEventTime, EventMap::GetTimeUntilEvent has to be used which returns the currently
scheduled delay until the event occurs.
EventMap::GetNextEventTime instead exposes the timer value internal to EventMap at which the event should occur.

**Changes proposed:**

-  Core/Scripts: Fix timing issue of EVENT_FLY_TO_RANDOM_PILLAR during Illidan encounter

- [x] 3.3.5
- [ ] master

**Tests performed:** build + in-game


Keader (approved):
looks correct to me

Issue #22913, opened 2019-01-06
Security Vulnerability


DennisWG:
**Description:**

Anyone with read-access to the account database (by legit means or otherwise) can log into any account they want to.

This is due to the fact that all the major emulators currently store the hash of USERNAME:PASSWORD - which is refered to as `x` in the SRP6 protocol - within their database, namely within the `sha_pass_hash` field inside the `account` table. Using this value, you can bypass the SRP6 authentication process, because it is used to calculate the shared secret (session key). Here's the critical step (as described per [rfc5054](https://tools.ietf.org/html/rfc5054)):

```
S = pow(B - k * (pow(g, x) % N), a + u * x) % N

where
B is known to both server and client
k is known to both server and client
g is known to both server and client
N is known to both server and client
a is known to the client
u is known to both server and client
x is calculated from the username & password and should only be known to the user
```
As you can see, the only value that isn't known to anyone, except the user, is `x` and this is what keeps this whole process safe.

Here's a demo video I've quickly thrown together demonstrating that I'm able to query the realm list of a cmangos based realm (this happens after the SRP6 authentication sequence) using only the account name and the value `x`.

https://www.youtube.com/watch?v=JAVDuzq3Jg0

It is important to note that this value is not used anywhere in the entire code-base, which makes me wonder why it was put there in the first place.

I'm not sure if this applies to the master branch, but at least until WotLK I know that WoW still uses SRP6.

**How to fix this:**
Luckily this is very simple:
Remove the `sha_pass_hash` field within the `account` table inside the database.

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you and add the missing information to **this issue**. Thanks.

Shauren:
> It is important to note that this value is not used anywhere in the entire code-base, which makes me wonder why it was put there in the first place.

This is incorrect, RASession.cpp

DennisWG:
My mistake about this not being used.
Depending on how hard it is to remove this from your code-base, it might be worth looking into just putting more stuff into the hash. What I mean is don't just store Hash(USERNAME:PASSWORD), store Hash(USERNAME:PASSWORD:SOMEOTHERVALUE) instead. Again, I don't know if this applies to TrinityCore, but using it like this will make it perfectly safe.

Treeston:
See #25157 for deprecation timeline of `sha_pass_hash`.

[...]

PR #25093, opened 2020-07-22
Core/Authserver: Auth cleanup phase 1

Treeston:
~~Nothing too huge yet, just some minor refactors. I'm trying to keep PR size low so they can be merged in a reasonable timeframe.~~

OK yeah nevermind it ended up getting bloated. Refactors have a tendency to lead to more refactors. Oops.

It's all building up to an eventual fix for #22913 in case that wasn't sufficiently implied.

---

## `AuthSession`
* Aggressive `std::array`-ification of fields
* Actually readable now due to aggressive use of `::GetDigestOf`
* A bunch of `BigNumber`s that were only ever used as byte arrays are now... well, byte arrays
  * In particular the session key, both here and in `WorldSession` (and related logic)

## `ARC4`
* Moved to `Trinity::Crypto`
* Normalized parameter order from `(int, uint8*)` to `(uint8*, size_t)`
* Added generic container overloads
* Moved seed length from ctor to `::Init`, I have no idea why this was in the ctor before

## `BigNumber`
* Added `BigNumber(uint32)` ctor
* Added `BigNumber(std::string const&)` ctor
* Added `SetBinary(C const&)` for generic containers
* Added `SetHexStr(std::string const&)`
* Removed all the `AsByteArray` variants
* Added `GetBytes` to retrieve bytes into a provided fixed-length buffer
* Added `ToByteArray<N>` to retrieve bytes as a fixed-size array
* Added `ToByteVector` to retrieve bytes into a vector

## `CryptoRandom`
* New header that's mostly just a wrapper around `RAND_bytes`
* `void GetRandomBytes(uint8*, size_t)` for arbitrary buffers
* `void GetRandomBytes(Container&)` for pre-sized containers
* `std::array<uint8, S> GetRandomBytes<S>()` to give you a new array

## `SessionKeyGenerator`
* The old `SHA1Randx`, used for Warden key generation
* It's now called `SessionKeyGenerator` because that's what it's called in master
* It's now templated on the hash function to use (because that's how it works in master)
* It is now aggressively more readable because of the next section

## `SHA1` & `HMAC_SHA1`
* Renamed `SHA1Hash` to `SHA1`, and `HmacHash` to `HMAC_SHA1`
* Moved both to the `Trinity::Crypto` namespace where they belong
* Both now expose a unified interface:
  * `static constexpr size_t DIGEST_LENGTH`
  * `using Digest = std::array<uint8, DIGEST_LENGTH>`
  * `void UpdateData(uint8 const*, size_t)` for arbitrary data
  * `void UpdateData(C const&)` for various container types (including `std::string`)
  * `static Digest GetDigestOf(Ts&&...)` for objectless hashing of one or more supported types

## Other changes
* `ByteArrayToHexStr` (in `Util.h`) now takes `size_t` instead of `int` for its length parameter
* `ByteArrayToHexStr` now has an overload for generic containers


src/common/Cryptography/SHA1.h
```diff
@@ -29,21 +30,36 @@ class TC_COMMON_API SHA1Hash
...
         void UpdateData(const std::string &str);
 
+        template <size_t SIZE, size_t... SIZES, typename... Ts> void UpdateBigNumbers(BigNumber const& bn, Ts&&... tail)
+        {
+            static_assert(sizeof...(SIZES) == sizeof...(Ts));
+            UpdateBigNumber<SIZE>(bn);
+            if constexpr (sizeof...(Ts) > 0)
+                UpdateBigNumbers<SIZES...>(std::forward<Ts>(tail)...);
+        }
+
```
Warpten, line 46:
```cpp
    template <
        typename... Ts>
    auto UpdateBigNumbers(Ts&&... ts) -> std::enable_if_t<
        (std::is_same_v<foo, std::decay_t<Ts>> && ...)
    , void>
    {
        return (UpdateBigNumber(std::forward<Ts>(ts)), ...);
    }
```

You can tweak the decay bit, i was feeling lazy

src/common/Utilities/Util.cpp
```diff
@@ -611,7 +611,7 @@ bool Utf8ToUpperOnlyLatin(std::string& utf8String)
     return WStrToUtf8(wstr, utf8String);
 }
 
-std::string ByteArrayToHexStr(uint8 const* bytes, uint32 arrayLen, bool reverse /* = false */)
+std::string ByteArrayToHexStr(uint8 const* bytes, size_t arrayLen, bool reverse /* = false */)
```
Warpten, line 614: cue anti-boost protests

`boost::algorithm::(un)hex`?

src/common/Utilities/Util.h
```diff
@@ -304,7 +304,8 @@ TC_COMMON_API bool IsIPAddress(char const* ipaddress);
 TC_COMMON_API uint32 CreatePIDFile(std::string const& filename);
 TC_COMMON_API uint32 GetPID();
 
-TC_COMMON_API std::string ByteArrayToHexStr(uint8 const* bytes, uint32 length, bool reverse = false);
+TC_COMMON_API std::string ByteArrayToHexStr(uint8 const* bytes, size_t length, bool reverse = false);
+template <typename C> std::string ByteArrayToHexStr(C const& container, bool reverse = false) { return ByteArrayToHexStr(std::data(container), std::size(container), reverse); }
```
Warpten, line 308: TC_COMMON_API lacking here kind of defeats the purpose of this method being here (and you can't add it)

Treeston: not sure i follow, `TC_*_API` shouldn't be necessary for any method defined in the header

src/common/Utilities/Util.cpp
Treeston: sure, i honestly just wanted to replace the parameter, we can change that

src/common/Cryptography/SHA1.h
Treeston: asserting can produce a nicer error message though

src/common/Utilities/Util.h
Warpten: IIRC `TC_*_API` macros are for hotswap
except you can't have declspec(dllexport) (which is what those macros do) on template methods for obvious reasons

so you broke hotswap builds

Warpten: actually i don't know i just know TC_API macro breaks it here if its added and apparently it builds on nix for obscure reasons so ... ???

src/common/Cryptography/SHA1.h
Warpten: sure. fold expressions though

src/common/Utilities/Util.h
Treeston: `TC_COMMON_API` resolves to `export` inside common, and to `import` outside common

but that only matters if the definition is only in common, if the definition is in the header then it's in every compilation unit anyway

src/common/Cryptography/SHA1.h
Treeston: ha sure we can do it

src/common/Cryptography/SHA1.h
```diff
@@ -29,21 +30,34 @@ class TC_COMMON_API SHA1Hash
...
-        uint8 mDigest[SHA_DIGEST_LENGTH];
+        static void _CheckedBNToBytes(BigNumber const& bn, uint8* arena, size_t sz);
+        template <size_t SIZE> void UpdateBigNumber(BigNumber const& bn)
+        {
+            std::array<uint8, SIZE> arena;
+            _CheckedBNToBytes(bn, arena.data(), SIZE);
+            UpdateData(arena.data(), SIZE);
+        }
+
+        SHA_CTX _mC;
```
Shauren, line 59: thats a not properly executed codestyle adjustment, old name `mC` was a variation of `m_C` without the underscore, you should have removed the m as well

src/common/Cryptography/SHA1.h
```diff
@@ -29,21 +30,22 @@ class TC_COMMON_API SHA1Hash
     public:
         SHA1Hash();
         ~SHA1Hash();
+        static constexpr size_t HASH_LEN = SHA_DIGEST_LENGTH;
 
-        void UpdateBigNumbers(BigNumber* bn0, ...);
 
         void UpdateData(const uint8 *dta, int len);
         void UpdateData(const std::string &str);
+        void UpdateData(char const* str) { UpdateData(std::string(str)); } /* explicit overload to avoid using the container template */
```
Warpten, line 38: <details>
<summary>std::string_view overload?</summary>

cue angry shauren noises
</details>

src/common/Cryptography/WardenKeyGeneration.h
```diff
@@ -25,18 +25,16 @@
 class SHA1Randx
 {
 public:
-    SHA1Randx(uint8* buff, uint32 size)
+    SHA1Randx(std::array<uint8, 40> const& buf)
```
Shauren, line 28: Size of this array isnt fixed and taking half of it in code below is the correct behavior, see master branch: https://github.com/TrinityCore/TrinityCore/blob/8f9a69c9fa812b82af97809e445fc176b28f90eb/src/common/Cryptography/SessionKeyGeneration.h#L28 
It was renamed and is being used with two different sizes right now, 40 and 32

Carbenium:
@Treeston Just a side-note since you are touching the crypto stuff: The OpenSSL low-level API (`SHA*_` and so on) is being officially deprecated with OpenSSL 3.0 (https://wiki.openssl.org/index.php/OpenSSL_3.0#Low_Level_APIs).
Might be a good time to switch to the `EVP_*` family of functions.

src/common/Cryptography/HMACSHA1.h
```diff
@@ -19,27 +19,76 @@
...
 class BigNumber;
 
-#define SEED_KEY_SIZE 16
-
-class TC_COMMON_API HmacHash
+namespace Trinity
+{
+namespace Crypto
+{
+class TC_COMMON_API HMAC_SHA1
```
Shauren, line 33: I don't like how this gets a refactor going in a different direction than in master branch

https://github.com/TrinityCore/TrinityCore/blob/b3ba029791a9003880d77a048b16e9e593485fe9/src/common/Cryptography/HmacHash.h#L49-L50

Carbenium:
@Shauren Can we drop OpenSSL 1.0 support? Even Debian Oldstable (Stretch) is on 1.1

src/common/Cryptography/HMAC.h
```diff
@@ -0,0 +1,138 @@
...
+namespace Trinity {
+    namespace Impl {
+        struct HMACImpl
+        {
+            typedef EVP_MD const* (*HashCreator)();
+
+#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER < 0x10100000L
+            static HMAC_CTX* MakeCTX()
+            {
+                HMAC_CTX* ctx = new HMAC_CTX();
```
Carbenium, line 39: https://www.openssl.org/docs/manmaster/man3/EVP_DigestSignInit.html

These are the corresponding `EVP_*` functions.

Treeston: Can you find any documentation on the `PKEY` functions for openssl < 1.1.1? I couldn't.

src/common/Cryptography/SHA1.h
```diff
@@ -15,60 +15,18 @@
...
-#define _AUTH_SHA1_H
+#ifndef TRINITY_SHA1_H
+#define TRINITY_SHA1_H
 
-#include "Define.h"
-#include <array>
-#include <string>
-#include <string_view>
+#include "GenericHash.h"
 #include <openssl/sha.h>
```
Carbenium, line 22:
```suggestion
#include <openssl/evp.h>
```

src/common/Cryptography/HMAC_SHA1.h
```diff
@@ -15,23 +15,18 @@
...
-#include <stdarg.h>
+#ifndef TRINITY_HMAC_SHA1_H
+#define TRINITY_HMAC_SHA1_H
 
-Trinity::Crypto::SHA1::SHA1()
-{
-    SHA1_Init(&_ctx);
-}
+#include "HMAC.h"
+#include <openssl/sha.h>
```
Carbenium, line 22:
```suggestion
#include <openssl/evp.h>
```

Carbenium: Actually there is one problem: `SHA_DIGEST_LENGTH` is in `<openssl/sha.h>`
Couple of options:
  * keep `sha.h` (not in favour)
  * define it ourselves
  * use `EVP_MD_size/EVP_MD_CTX_size`

Treeston: The third option isn't an option, these aren't `constexpr`.

src/common/Cryptography/ARC4.cpp
```diff
@@ -17,34 +17,26 @@
...
 #include "ARC4.h"
 
-ARC4::ARC4(uint32 len) : m_ctx(EVP_CIPHER_CTX_new())
+Trinity::Crypto::ARC4::ARC4() : _ctx(EVP_CIPHER_CTX_new())
 {
-    EVP_CIPHER_CTX_init(m_ctx);
-    EVP_EncryptInit_ex(m_ctx, EVP_rc4(), nullptr, nullptr, nullptr);
-    EVP_CIPHER_CTX_set_key_length(m_ctx, len);
+    EVP_CIPHER_CTX_init(_ctx);
+    EVP_EncryptInit_ex(_ctx, EVP_rc4(), nullptr, nullptr, nullptr);
```
Carbenium, line 23: Maybe some asserts sprinkled in here wouldn't be the worst idea.

Treeston: ah yes, i missed those

src/common/Cryptography/CryptoHash.h
```diff
@@ -0,0 +1,111 @@
...
+namespace Trinity
+{
+    namespace Impl
+    {
+        struct GenericHashImpl
+        {
+            typedef EVP_MD const* (*HashCreator)();
+
+#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER < 0x10100000L
+            static EVP_MD_CTX* MakeCTX() { return EVP_MD_CTX_create(); }
```
Carbenium, line 40: same here.

src/common/Cryptography/CryptoHash.h
```diff
@@ -0,0 +1,111 @@
...
+            void DestroyCTX(EVP_MD_CTX* ctx) { EVP_MD_CTX_destroy(ctx); }
```
Carbenium, line 41: I don't see any calls to `GenericHashImpl::DestroyCTX`

src/common/Cryptography/CryptoHash.h
```diff
@@ -0,0 +1,111 @@
...
+                void UpdateData(char const* str) { UpdateData(std::string_view(str)); } /* explicit overload to avoid using the container template */
+                template <typename Container> void UpdateData(Container const& c) { UpdateData(std::data(c), std::size(c)); }
+
+                void Finalize()
+                {
+                    uint32 length;
+                    int result = EVP_DigestFinal_ex(_ctx, _digest.data(), &length);
+                    ASSERT(result == 1);
+                    ASSERT(length == DIGEST_LENGTH);
+                    _ctx = nullptr;
```
Carbenium, line 93: Either call `EVP_MD_CTX_free` or reuse the same ctx (has to first inited with `EVP_DigestInit_ex` again).

src/common/Cryptography/BigNumber.cpp
```diff
@@ -185,15 +179,14 @@ bool BigNumber::IsNegative() const
     return BN_is_negative(_bn);
 }
 
-bool BigNumber::AsByteArray(uint8* buf, std::size_t bufsize, bool littleEndian) const
+void BigNumber::GetBytes(uint8* buf, std::size_t bufsize, bool littleEndian) const
 {
     int nBytes = GetNumBytes();
-    ASSERT(!(nBytes < 0));
+    ASSERT(!(nBytes < 0), "Bignum has negative number of bytes (%d).", nBytes);
```
Carbenium, line 185:
```suggestion
    ASSERT(nBytes >= 0, "Bignum has negative number of bytes (%d).", nBytes);
```
if you like. I find it more readable.

src/common/Cryptography/CryptoHash.h
```diff
@@ -0,0 +1,120 @@
...
+        };
+
+        template <GenericHashImpl::HashCreator HashCreator, size_t DigestLength>
+        class GenericHash
+        {
+            public:
+                static constexpr size_t DIGEST_LENGTH = DigestLength;
+                using Digest = std::array<uint8, DIGEST_LENGTH>;
+
+                static Digest GetDigestOf(uint8 const* dta, int len)
```
Carbenium, line 55:
```suggestion
                static Digest GetDigestOf(uint8 const* data, size_t len)
```

src/common/Cryptography/CryptoHash.h
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    GenericHash hash;
+                    hash.UpdateData(dta, len);
```
Carbenium, line 58:
```suggestion
                    hash.UpdateData(data, len);
```

src/common/Cryptography/CryptoHash.h
```diff
@@ -0,0 +1,120 @@
...
+
+                ~GenericHash()
+                {
+                    if (!_ctx)
+                        return;
+                    GenericHashImpl::DestroyCTX(_ctx);
+                    _ctx = nullptr;
+                }
+
+                void UpdateData(uint8 const* dta, size_t len)
```
Carbenium, line 85:
```suggestion
                void UpdateData(uint8 const* data, size_t len)
```

src/common/Cryptography/CryptoHash.h
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    int result = EVP_DigestUpdate(_ctx, dta, len);
```
Carbenium, line 87:
```suggestion
                    int result = EVP_DigestUpdate(_ctx, data, len);
```

src/common/Cryptography/HMAC.h
```diff
@@ -0,0 +1,132 @@
...
+    {
+        struct HMACImpl
+        {
+            typedef EVP_MD const* (*HashCreator)();
+
+#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER < 0x10100000L
+            static HMAC_CTX* MakeCTX()
+            {
+                HMAC_CTX* ctx = new HMAC_CTX();
+                int result = HMAC_CTX_init(ctx);
```
Warpten, line 43:
```suggestion
                [[maybe_unused]] int result = HMAC_CTX_init(ctx);
```

keeps the linters happy

src/common/Cryptography/CryptoHash.h
```diff
@@ -0,0 +1,120 @@
...
+            private:
+                EVP_MD_CTX* _ctx;
+                Digest _digest;
+        };
+    }
+
+    namespace Crypto
+    {
+        using SHA1 = Trinity::Impl::GenericHash<EVP_sha1, Constants::SHA1_DIGEST_LENGTH_BYTES>;
+        using SHA256 = Trinity::Impl::GenericHash<EVP_sha256, Constants::SHA256_DIGEST_LENGTH_BYTES>;
```
Warpten, line 116: This is vastly irrelevant but i'd use openssl defines here since you're already using openssl types, just for consistency

Carbenium: We discussed this previously. See https://github.com/TrinityCore/TrinityCore/pull/25093#discussion_r460142261 . 
That's why we ended up here.

Carbenium:
Alright, if Warpten is even going for linter warnings, then I have also another nit-pick ;)

How about C++17 syntax for nested namespaces?
```c++
namespace Trinity::Crypto
{
...
```

src/common/Cryptography/CryptoHash.h
Warpten: Fair

src/common/Cryptography/HMAC.h
Carbenium: Should be solved when we drop <1.1.1.

src/common/Cryptography/HMAC.h
Treeston: no that one actually doesn't have a return value, that's why appveyor died

(smh openssl 1.0.2)

Carbenium (approved):
Just some more nit-picks ;) Otherwise LGTM!

src/server/game/Server/WorldSocket.cpp
```diff
@@ -501,17 +498,17 @@ void WorldSocket::HandleAuthSessionCallback(std::shared_ptr<AuthSession> authSes
     }
 
     // Check that Key and account name are the same on client and server
-    uint32 t = 0;
+    uint8 t[4] = { 0x00,0x00,0x00,0x00 };
```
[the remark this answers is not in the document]
Carbenium: no love for whitespaces? :cry:

Warpten:
missed a few more of these `[[maybe_unused]]` on ossl result codes

src/common/Cryptography/HMAC.h
```diff
@@ -0,0 +1,128 @@
...
+                hash.Finalize();
+                return hash.GetDigest();
+            }
+
+            GenericHMAC(uint8 const* seed, size_t len) : _ctx(HMACImpl::MakeCTX())
+            {
+                int result = HMAC_Init_ex(_ctx, seed, len, HashCreator(), nullptr);
+                ASSERT(result == 1);
+            }
+            template <typename C> GenericHMAC(C const& container) : GenericHMAC(std::data(container), std::size(container)) {}
```
Shauren, line 86: Why are you placing the template part on the same line as everything else? We don't do that anywhere

illfated:
Great job. Love your style.

PR #25095, opened 2020-07-22
Add support for Catch2 as unit test framework

Carbenium:
**Changes proposed:**

-  Add catch2 as unit test framework which is called through ctest
-  Add some basic tests for EventMap

**How to run unit tests**:
 - Run CMake with `-DBUILD_TESTING=1`
 - Build tests with `make all` or `make tests-common`
 - Run tests with `make test`

- [x] 3.3.5
- [ ] master

**Tests performed:** build and **unit tests**

- [x] Wire it up for CI
- [ ] Add some docs how to execute unit tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
           command: |
             mkdir bin
             cd bin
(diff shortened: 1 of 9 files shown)


Shauren:
too big

Carbenium:
@Shauren Well, in principle it's a header-only lib. It would be possible to pull-in the one header file only (https://raw.githubusercontent.com/catchorg/Catch2/master/single_include/catch2/catch.hpp) and the 1 or 2 CMake scripts.

I, personally am also okay with leaving the sources completely out-of-tree. Not sure how you would solve it on Windows.

Warpten:
> we want less deps as source
Git submodule is calling

Carbenium:
@Naios Thanks for the suggestion! I implemented it with `FetchContent` now. Could you take a look if this is fine for you?

Naios:
@Carbenium LGTM! Maybe you could even use the archive github download link since it will be faster than via git: `https://github.com/catchorg/Catch2/archive/v2.13.0.zip`.

Issue #24356, opened 2020-03-27
Update Time Diff Unstable


jjo72ddd:
Update time diff sometimes has high spikes.
- [x] CMSG_ADD_FRIEND https://github.com/TrinityCore/TrinityCore/issues/24356#issuecomment-666295422
- [x] Loading mails sync
- [x] Loading pets sync
- [ ] SAI scripts with SetActive() action https://github.com/TrinityCore/TrinityCore/issues/24356#issuecomment-674500552
- [ ] Delete chars only on startup https://github.com/TrinityCore/TrinityCore/issues/25483
- [x] Spikes when loading mmaps
  - In general loading tiles/grids will always take time and if one cares about lag, they should be never unloaded (GridUnload = 0) and maybe even preloaded on startup (BaseMapLoadAllGrids = 1) . If that's not possible, consider checking https://github.com/TrinityCore/TrinityCore/commit/b6e205f7ef32f7b82644043d82a808e67cc9acb2 and https://github.com/TrinityCore/TrinityCore/pull/25520 to reduce the load times of MMaps (which can sum up even to 2 seconds for 1 player logging in)

Available tools to investigate high update time diff:
- Grafana https://trinitycore.atlassian.net/wiki/spaces/tc/pages/19955721/Monitoring+a+TrinityCore+server , i.e. https://tcubuntu.northeurope.cloudapp.azure.com/grafana/d/IRRL03nMk/performance-profiling?orgId=1&refresh=1m
  - Enable CMake flag WITH_DETAILED_METRICS and uncomment the settings in worldserver.conf at https://github.com/TrinityCore/TrinityCore/blob/6c7837f947ff4eb5110a116a371daa6f9e2b3bbe/src/server/worldserver/worldserver.conf.dist#L4080-L4081 to enable more metrics (the default ones are always enabled)
- Enable CMake flag WITH_COREDEBUG to spot sync database queries during World Update loop (they should be async instead)

jackpoz:
You can use Grafana to monitor the update time , https://trinitycore.atlassian.net/wiki/spaces/tc/pages/19955721/Monitoring+a+TrinityCore+server

jjo72ddd:
Thank you, I will try to install Grafana, but what could be the reason for such a jump from 50 to 800 Update Time Diff?

jjo72ddd:
I noticed that the worldserver process uses up to 90% one core. Can I distribute across all available cores?

Killyana:
[...]

PR #25098, opened 2020-07-23
Core/Metric: Add more metrics about World::Update() loop

jackpoz:
**Changes proposed:**

- Add more metrics to see where World::Update() spends most of its time
- Add new cmake setting WITHOUT_METRICS that disables all metrics
- Remove old deprecated performance profiling features
  - "server set difftime" command
  - "RecordUpdateTimeDiffInterval" and "MinRecordUpdateTimeDiff" worldserver config settings
  - related code (the whole UpdateTime class could be removed too.)

- [x] 3.3.5
- [ ] master

**Issues addressed:** Ref #24356 

See https://snapshot.raintank.io/dashboard/snapshot/YodeVT0A5EYQgRTM2IekERCNN3z3nzr0
image

None


Carbenium:
I'd say drop the whole `UpdateTime` stuff if it isn't needed any more. Then nobody is tempted to use it.

jackpoz:
UpdateTime is used in .server info and changing that requires to mess with the command response string in the db. It will be in another pr.

PR #25109, opened 2020-07-24
Build: Add option to use gold linker

Carbenium:
**Changes proposed:**

-  Build: Add WITH_LD_GOLD option to use gold linker on unix

- [x] 3.3.5
- [ ] master

Credits for the idea go to @jackpoz


Faq:
But why it is needed?

Carbenium:
@Faq The gold linker has a better performance than the standard ld (cuts re-linking time on the large debug builds for me in half).
The initial idea was to enable it for the CI builds. Unfortunately the time reduction there is not really significant. It seems the major speed-up is during re-linking.
To keep the default config users are using in sync with CI I decided not to enable it for now. I will keep using it locally to gather some more experience and maybe we can enable it as default linker at some point.

jackpoz:
https://www.youtube.com/watch?v=ntG50eXbBtc

PR #25110, opened 2020-07-24
common/Crypto: ifdef OpenSSL thread locking code

Carbenium:
**Changes proposed:**

The code in question is not needed for versions 1.1.0 and above
and are no-ops. ifdef the parts until we drop support for < 1.1.0.

Fixes a -Wunused-value warning on GCC 7.5

Ref: https://github.com/openssl/openssl/commit/2e52e7df518d80188c865ea3f7bb3526d14b0c08

- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/common/Cryptography/OpenSSLCrypto.h
+++ b/src/common/Cryptography/OpenSSLCrypto.h
@@ -19,6 +19,7 @@
 #define OPENSSL_CRYPTO_H
 
 #include "Define.h"
+#include <openssl/opensslv.h>
 
 /**
 * A group of functions which setup openssl crypto module to work properly in multithreaded enviroment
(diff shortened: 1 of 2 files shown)


Treeston:
make the no-op versions visible in the header please (so the optimizer sees them)

Carbenium:
@Treeston This should do it.

src/common/Cryptography/OpenSSLCrypto.h
```diff
@@ -19,17 +19,25 @@
...
 namespace OpenSSLCrypto
 {
+
+#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER < 0x1010000fL
     /// Needs to be called before threads using openssl are spawned
     TC_COMMON_API void threadsSetup();
     /// Needs to be called after threads using openssl are despawned
     TC_COMMON_API void threadsCleanup();
+#else
+    TC_COMMON_API void threadsSetup() { };
```
Treeston, line 37: `TC_COMMON_API` not needed here, it's only needed if the definition is in cpp

Carbenium: This is also true. Fixed.

Treeston (approved):

PR #25116, opened 2020-07-25
Core/EventMap: Finish std::chrono-fication

Carbenium:
**Changes proposed:**

-  EventMap: finish std::chrono-fication
-  EventMap: Unify semantics of DelayEvents
`DelayEvents(Milliseconds delay)` had different semantics than `DelayEvents(Milliseconds delay, uint32 group)`.
The first method delayed the events only in the case the internal timer already ticked at least for the amount of delay. In contrast the latter method delayed events regardless of the internal timer value.
Use the latter semantics for `DelayEvents(Milliseconds delay)` as well which makes the outcome more predictable.
Callers were examined. No usage was found which depends on this edge case.

-  EventMap: Add some more tests

- [x] 3.3.5
- [ ] master

**Tests performed:** build + unit tests

### DIFF
--- a/src/common/Utilities/EventMap.cpp
+++ b/src/common/Utilities/EventMap.cpp
@@ -94,7 +94,16 @@ uint32 EventMap::ExecuteEvent()
 
 void EventMap::DelayEvents(Milliseconds delay)
 {
-    _time = delay < _time - _time.min() ? _time - delay : TimePoint::min();
+    if (Empty())
+        return;
+
+    EventStore delayed = std::move(_eventMap);
+    for (EventStore::iterator itr = delayed.begin(); itr != delayed.end();)
+    {
+        EventStore::node_type node = delayed.extract(itr++);
+        node.key() = node.key() + delay;
+        _eventMap.insert(_eventMap.end(), std::move(node));
+    }
 }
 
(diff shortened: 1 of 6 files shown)


src/common/Utilities/EventMap.cpp
```diff
@@ -94,7 +94,18 @@ uint32 EventMap::ExecuteEvent()
...
+
+    EventStore delayed;
+
+    for (EventStore::iterator itr = _eventMap.begin(); itr != _eventMap.end();)
+    {
+        delayed.insert(EventStore::value_type(itr->first + delay, itr->second));
+        _eventMap.erase(itr++);
+    }
+
+    _eventMap.insert(delayed.begin(), delayed.end());
```
Shauren, line 108: Maybe something like this?

```suggestion
    EventStore delayed = std::move(_eventMap);

    for (EventStore::iterator itr = delayed.begin(); itr != delayed.end();)
    {
        EventStore::node_type node = delayed.extract(itr++);
        node.key() = node.key() + delay;
        _eventMap.insert(_eventMap.end(), std::move(node));
    }
```

Carbenium: Hah, `extract`. That's a new one. Thanks for the suggestion!

PR #25135, opened 2020-07-26
Authserver cleanup phase 3 - the account table unfuckening

Treeston:
_(This is part 3 of 5 in the deprecation process of `sha_pass_hash`, as outlined in #25157.)_

## Changes
* `salt` and `verifier` are now `BINARY(32)`, from `VARCHAR(64)`
* `SessionKey` is now `BINARY(40)`, from `VARCHAR(80)`

## Backwards compatibility
* `sha_pass_hash` will still be updated, and external tools can still set it
* `sha_pass_hash` will only be read if backwards-compatibility fields `s`/`v` were changed by external tools


jackpoz:
> * `sha_pass_hash` is no longer updated, though external tools can still set it

this breaks https://github.com/TrinityCore/minimanager , in particular login using TC as auth. See https://github.com/TrinityCore/aowow/blob/master/includes/user.class.php#L286 .

It would be nice if that tool (and other tools under the TC umbrella) would be updated before merging this PR

Treeston:
@jackpoz we want some of the changes in this PR before updating external tools

re-added sha_pass_hash updating (for now)

src/server/authserver/Server/AuthSession.cpp
```diff
@@ -415,8 +419,8 @@ void AuthSession::LogonChallengeCallback(PreparedQueryResult result)
     {
         Trinity::Crypto::SRP6::Salt salt;
         Trinity::Crypto::SRP6::Verifier verifier;
-        HexStrToByteArray(fields[11].GetString(), salt, true); /* this is actually flipped in the DB right now, old core did hexstr (big endian) -> bignum -> byte array (little-endian) */
-        HexStrToByteArray(fields[12].GetString(), verifier);
+        fields[11].GetBinary(salt);
```
Shauren, line 422: Can you use the other overload here (and everywhere else), this looks just weird/wrong

Shauren (changes requested):

jackpoz:
I would add https://github.com/TrinityCore/account-creator also to the list of apps that would be nice to fix before merging this

src/server/authserver/Server/AuthSession.cpp
```diff
@@ -120,6 +121,41 @@ std::array<uint8, 16> VersionChallenge = { { 0xBA, 0xA3, 0x1E, 0x99, 0xA0, 0x0B,
...
+{
+    TC_LOG_INFO("server.authserver", "Updating password hashes...");
+    uint32 const start = getMSTime();
+    uint32 c = 0;
+
+    bool hadWarning = false;
+    // the auth update query nulls salt/verifier if they cannot be converted
+    // if they are non-null but s/v have been cleared, that means a legacy tool touched our auth DB (otherwise, the core might've done it itself, it used to use those hacks too)
+    QueryResult result = LoginDatabase.Query("SELECT id, sha_pass_hash, IF((salt IS null) AND (verifier IS null), 0, 1) AS shouldWarn FROM account WHERE s != DEFAULT(s) OR v != DEFAULT(v)");
+    if (result) do
```
Shauren, line 134: dont mess with me

PR #25151, opened 2020-07-27
 Core/PacketIO: update CMSG_LFG_JOIN and CMSG_LFG_LEAVE packet structures

Faq:
**Changes proposed:**

-  attempting to CP  CMSG_LFG_JOIN and CMSG_LFG_LEAVE from [4.x branch](https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/commit/6abeb365842c506f2151e01e87e2981ee7ff0a7e)

- [x] 3.3.5
- [ ] master

compiles
Tested with 5 accounts: lvl 80 chars, queued, one from them used leave button and got cd, other chars got in dungeon. After timeout passed, joined in same group.

- [x] tests finished

### DIFF
--- a/src/server/game/Handlers/LFGHandler.cpp
+++ b/src/server/game/Handlers/LFGHandler.cpp
@@ -15,7 +15,9 @@
  * with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "DBCStores.h"
 #include "LFGMgr.h"
+#include "LFGPackets.h"
 #include "Log.h"
 #include "GameTime.h"
 #include "Group.h"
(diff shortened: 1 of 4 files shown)


src/server/game/Server/Packets/LFGPackets.cpp
```diff
@@ -0,0 +1,33 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "LFGPackets.h"
+
+void WorldPackets::LFG::LFGJoin::Read()
+{
+    _worldPacket >> Roles;
+    for (uint8 i = 0; i < 3; ++i)
```
Shauren, line 23: This is not the same code you removed from https://github.com/TrinityCore/TrinityCore/pull/25151/files#diff-ee1b5ab275f949c7fb79bc7ea2e24287L62-L64, obviously breaks LFG

jackpoz:
I would suggest to stick to less complicated CP (sometimes it's easier to just do a brand new fresh commit instead of attempting cp'ing complicated stuff)

src/server/game/Server/Packets/LFGPackets.cpp
Faq: Should be ok with https://github.com/TrinityCore/TrinityCore/pull/25151/commits/0fe5155db39ecd1cd1a3b43aaf16a41a9dd56586

Carbenium (changes requested):
See inline comments to make this fully compatible with 335.

src/server/game/Handlers/LFGHandler.cpp
```diff
@@ -46,61 +47,44 @@ void BuildPartyLockDungeonBlock(WorldPacket& data, lfg::LfgLockPartyMap const& l
...
+    if (!lfgJoin.Slots)
     {
         TC_LOG_DEBUG("lfg", "CMSG_LFG_JOIN %s no dungeons selected", GetPlayerInfo().c_str());
-        recvData.rfinish();
         return;
     }
 
     lfg::LfgDungeonSet newDungeons;
-    for (int8 i = 0; i < numDungeons; ++i)
+    for (int8 i = 0; i < lfgJoin.Slots; ++i)
```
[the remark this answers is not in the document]
Faq: you sure about uint32? as now it is int8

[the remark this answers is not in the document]
Carbenium: The counter was and still is (u)int8 (see down below)
```c++
Slots.resize(_worldPacket.read<uint8>());
```
`slot` is not the counter but the content itself, which was previously `uint32` as well.

src/server/game/Server/Packets/LFGPackets.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "LFGPackets.h"
+
+void WorldPackets::LFG::LFGJoin::Read()
+{
+    _worldPacket >> Roles;
+    _worldPacket.read_skip<uint16>();
+    _worldPacket >> Slots;
+    for (int8 i = 0; i < Slots; ++i)
+        _worldPacket >> dungeonId;
```
Warpten, line 26: This is broken anyways

jackpoz: >This is utterly broken anyways

The message above is broken too because it doesn't explain in any way which part is broke, why, how it should be instead or how to fix it.

jackpoz:
As I wrote some days ago:
>I would suggest to stick to less complicated CP (sometimes it's easier to just do a brand new fresh commit instead of attempting cp'ing complicated stuff)

The code review of this PR is getting just as complicated as rewriting from scratch the code. The point of cherry-picking commits is to apply as-is commits that are compatible with different branches. Packet structures are not and require manual work.

In addition to this, every change to packet structures has a chance on breaking something that is working fine, adding no value to the user (and actually removing value by introducing bugs).

I would suggest once again to avoid complicated CP like this one altogether, either focusing on other cherry-picks or just writing the changes from scratch (and inspecting very carefully that everything read/written from/to packets is exactly the same as before, ignoring other branches)

Carbenium (approved):

PR #25159, opened 2020-07-28
Update create_mysql.sql

jhskim78:
Grant to create user is deprecated.  Also, trinity user should not need access to existing databases, which the first grant was giving access to.

**Changes proposed:**

-  
-  
-  

- [x] 3.3.5
- [ ] master

- [ ] 
- [ ]


Shauren:
IF NOT EXISTS was added to mysql in version 5.7.6 which is newer than our minimum required version, this cannot be used.

PR #25168, opened 2020-07-29
Core/Units: add new functionality for units to block or enable combat interactions

Ovahlord:
**Changes proposed:**
As a followup to 6440c3bcac85a40de5c34aef1d8a8856966cc140 and the discussions about it, these changes will add a larger span of possibilities.
-  all units will now be able to ignore combat entirely via member variable in the unit class. The added helpers allow us to change that ignore state during runtime.

- [x] 3.3.5
- [ ] master

tested on 4.x

- [ ] (optional) https://paste2.org/p8aK1KZF
this is a list of all creatures from creaturedifficulty dbc table up to expansion 3 that have the ignore combat difficulty flag. There are MANY creatures that are set to ignore combat that should not have set it by default so there must be some indicator whenever this flag is suposed to be applied right on creation or applied during runtime.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -614,6 +614,9 @@ bool Creature::UpdateEntry(uint32 entry, CreatureData const* data /*= nullptr*/,
         ApplySpellImmune(0, IMMUNITY_EFFECT, SPELL_EFFECT_ATTACK_ME, true);
     }
 
+    if (cInfo->flags_extra & CREATURE_FLAG_EXTRA_NO_COMBAT)
+        SetIgnoringCombat(true);
+
     LoadTemplateRoot();
     InitializeMovementFlags();
 
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -614,8 +614,7 @@ bool Creature::UpdateEntry(uint32 entry, CreatureData const* data /*= nullptr*/,
         ApplySpellImmune(0, IMMUNITY_EFFECT, SPELL_EFFECT_ATTACK_ME, true);
     }
 
-    if (cInfo->flags_extra & CREATURE_FLAG_EXTRA_NO_COMBAT)
-        SetIgnoringCombat(true);
+    SetIgnoringCombat((cInfo->flags_extra & CREATURE_FLAG_EXTRA_NO_COMBAT) != 0);
 
     LoadTemplateRoot();
     InitializeMovementFlags();
--- a/src/server/game/Entities/Unit/Unit.h
+++ b/src/server/game/Entities/Unit/Unit.h
@@ -1686,6 +1686,11 @@ class TC_GAME_API Unit : public WorldObject
 
         float GetCollisionHeight() const override;
 
+        // returns if the unit is ignoring any combat interaction
+        bool IsIgnoringCombat() const { return _isIgnoringCombat; }
+        // enables/disables combat interaction of this unit.
+        void SetIgnoringCombat(bool apply) { _isIgnoringCombat = apply; }
+
         std::string GetDebugInfo() const override;
     protected:
         explicit Unit (bool isWorldObject);
@@ -1832,6 +1837,8 @@ class TC_GAME_API Unit : public WorldObject
 
         SpellHistory* m_spellHistory;
         PositionUpdateInfo _positionUpdateInfo;
+
+        bool _isIgnoringCombat;
 };
 
 namespace Trinity
--- a/src/server/game/Combat/CombatManager.cpp
+++ b/src/server/game/Combat/CombatManager.cpp
@@ -44,12 +44,9 @@
         return false;
     if (a->HasUnitState(UNIT_STATE_IN_FLIGHT) || b->HasUnitState(UNIT_STATE_IN_FLIGHT))
         return false;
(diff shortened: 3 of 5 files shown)


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -614,6 +614,9 @@ bool Creature::UpdateEntry(uint32 entry, CreatureData const* data /*= nullptr*/,
         ApplySpellImmune(0, IMMUNITY_EFFECT, SPELL_EFFECT_ATTACK_ME, true);
     }
 
+    if (cInfo->flags_extra & CREATURE_FLAG_EXTRA_NO_COMBAT)
```
ipriver, line 617: What if it'd been already modified to `_isIgnoringCombat = true`? Unnecessary call of `SetIgnoringCombat(true)` function.

src/server/game/Entities/Unit/Unit.h
```diff
@@ -1686,6 +1686,11 @@ class TC_GAME_API Unit : public WorldObject
 
         float GetCollisionHeight() const override;
 
+        // returns if the unit is ignoring any combat interaction
+        bool IsIgnoringCombat() const { return _isIgnoringCombat; }
```
ipriver, line 1690: jfyi: could it better to do smth like `return _isIgnoringCombat || flags_extra & CREATURE_FLAG_EXTRA_NO_COMBAT` instead of a getter function?

Ovahlord: No, because this state is suposed to be modifyable during runtime. A flag check would defy that purpose because extra flags are static and cannot be changed.

src/server/game/Entities/Creature/Creature.cpp
Ovahlord: Wrong. UpdateEntry is being called on creature creation and transforms which both are suposed to consider this flag since transform entries can have different flags than the original entry.

jackpoz:
Thanks for the PR :)

xvwyh:
I am getting more and more confident that `CREATURE_DIFFICULTYFLAGS_IGNORE_COMBAT` really should just behave like our passive reactstate. I looked at Ayamiss the Hunter fight in AQ20 video (not opened on my realm yet), https://classic.wowhead.com/npc=15546 spawn throughout the fight and they have this flag initially. They just fly around the arena, don't chase players if attacked, but they can be killed. And later on in the fight the flag is presumably removed from them, because they start chasing players, unless they were killed. Sadly, they also have "NoAutoRegen" flag on them too, so I can't reliably say if they enter combat or not, but it would be very awkward if they weren't. I'll be able to see it for myself when I get a chance to do that raid.

I'm still not convinced that `CREATURE_FLAG_EXTRA_NO_COMBAT` should exist... But I can't deny that Icecrown frost trap doesn't cause the player to enter combat, and the creatures only have "Ignore Combat" and "Unselectable" flags, and the damaging spell doesn't have a single attribute that would allow it to skip threat/aggro, so I don't know...

xvwyh:
>  (optional) https://paste2.org/4nvVANOk
> UPDATE `creature_template` SET `flags_extra`= `flags_extra` | 0x2000 WHERE `entry` IN (914, 4279, 4516, 7430, 
WHAT ARE YOU DOING? These are IDs of CreatureDifficulty.db2 rows! Not IDs of actual creatures!
Row with ID 914 points to creature 1182, row with ID 4952 points to creature 4952, and so on...

Ovahlord:
I am aware of that and I'm way ahead of you and have created a little excel doc table where we are currently checking for units that may be able to attack on their own and which not. This flag really is so weird. This stuff gets even more confusing when checking it against sniffs. Like C'thun for example does enter combat in phase two but he's flagged with the difficulty flag for both, his default and transformed entry. So I'd say this flag primarily blocks entering combat from the creature's side but does not prevent combat in general.

xvwyh:
> Like C'thun for example does enter combat in phase two but he's flagged with the difficulty flag for both, his default and transformed entry.

Well, since the flag can be turned off in a script, I don't see why this is of any concern. OK, it doesn't give us any info about whether creatures change this flag when "changing mode" (this is how blizzard calls our `UpdateEntry`), but there's nothing stopping them from removing "Ignore Combat" flag manually in the script when transitioning to phase 2.

xvwyh:
Having "Ignore Combat" in CreatureDifficulty means that the creature spawns with that flag set. Afterwards - anything can happen via scripts. They may as well remove the flag in "Spawn" event, if they wanted to, for whatever weird reason.

Ovahlord:
well, I went with the 'do not enter combat by yourself' approach like this here https://paste2.org/H4AN6fFe  and to my surprise this works surprisingly well in many situations.  The unit reacts like usual when attacked by something else but the unit itself no longer does any default combat reaction and this seems to come quite close to what I also observed in sniffs

xvwyh:
I don't know how the new Treeston-redesigned™ combat system in TC works, so I'll take your word for it.

But can you please confirm that you corrected your SQL and that you're not confusing CreatureDifficulty ID with Creature ID? And if so, please remove or correct the pastebin from your main post. The last thing I want if for you to be looking at a completely wrong list of creatures and trying to find reason in nonsense.

Ovahlord:
I did https://paste2.org/p8aK1KZF and just as I've said before, I've also built a excel table for it to check for creatures that might need a few extra checks and fixups when that flag is applied.

xvwyh:
Thanks, I'm relieved now. Good luck!

PR #25174, opened 2020-07-30
Log sync db queries in World::Update() loop

jackpoz:
**Changes proposed:**

-  Log sync db queries in World::Update() loop when building with WITH_COREDEBUG

- [x] 3.3.5
- [ ] master

**Issues addressed:** None

Logged in, spotted a few sync queries on login

- None

### DIFF
--- a/src/server/database/Database/DatabaseWorkerPool.cpp
+++ b/src/server/database/Database/DatabaseWorkerPool.cpp
@@ -33,6 +33,9 @@
 #include "Transaction.h"
 #include "MySQLWorkaround.h"
 #include <mysqld_error.h>
+#ifdef TRINITY_DEBUG
+#include <boost/stacktrace.hpp>
+#endif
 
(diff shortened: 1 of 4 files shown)


jackpoz:
Reported sync queries:
- https://www.irccloud.com/pastebin/biGPB9qA/
- https://www.irccloud.com/pastebin/q60AUmiz/
- https://www.irccloud.com/pastebin/ZLsZ13C4/

robinsch:
Maybe instead of debug use PERFORMANCE_PROFILING

jackpoz:
well WITH_COREDEBUG can be enabled in RelWithDebInfo also. Keep in mind boost::stacktrace::stacktrace() takes seconds which means you don't want to always have this On anyway.

On top of that, none should ever use PERFORMANCE_PROFILING , that's strictly for running worldserver with a profiler attached.

PR #25199, opened 2020-08-03
Scripts/BlackwingLair: Suppression Device Room

Jinnaix:
**Changes proposed:**

-  Fully working Suppression Device Room

See https://github.com/TrinityCore/TrinityCore/pull/25199#issuecomment-670730662 as for the update after taking this PR back into draft.

- [x] 3.3.5

**Issues addressed:** 
None

**Tests performed:** 
- [x] builds
- [x] tested in-game

The Rogue spell "Disarm Trap" despawns the target by default. This should not happen to Suppression Devices.
We prevent this currently with an action OnLootStateChange.
This is considered as a hackfix but this is a task for another PR.

I dont know and cant confirm if the option to interact with these gameobjects is aviable for anyone else than Rogues.
Right now, other players have the option to interact with it (idicated by the small gearwheel), but only stealthed Rogues with Disarm Trap can really interact.

While we _could_ fix the currently wrong animation loop by splitting cast and animation,we decided to keep it like this since it was like that back then. Therefore, compared to current WoW Classic, we have a delay in our animation loop.


src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +122,86 @@ class boss_broodlord : public CreatureScript
...
+                _events.ScheduleEvent(EVENT_SUPPRESSION_ANIM, Seconds(2));
+            }
+
+            void UpdateAI(uint32 diff) override
+            {
+                if (_instance->GetBossState(DATA_VAELASTRAZ_THE_CORRUPT) != DONE)
+                    return;
+
+                if (_instance->GetBossState(DATA_VAELASTRAZ_THE_CORRUPT) == DONE)
+                    {
```
MaxtorCoder, line 144: formatting

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +122,86 @@ class boss_broodlord : public CreatureScript
...
+                        if (me->GetGoState() == GO_STATE_DESTROYED)
+                        me->SetLootState(GO_READY); // NOT doing this will despawn the gameobject after using Disarm Trap
+                    }
+
+                if (_instance->GetBossState(DATA_BROODLORD_LASHLAYER) == DONE)
+                    {
```
MaxtorCoder, line 150: formatting

MaxtorCoder (changes requested):

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +122,86 @@ class boss_broodlord : public CreatureScript
...
+
+                _events.Update(diff);
+
+                while (uint32 eventId = _events.ExecuteEvent())
+                {
+                    switch (eventId)
+                    {
+                        case EVENT_SUPPRESSION_AURA:
+                            if (me->GetGoState() == GO_READY)
+                            me->CastSpell(nullptr,SPELL_SUPPRESSION_AURA,true);
```
MaxtorCoder, line 164: formatting as well

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +122,86 @@ class boss_broodlord : public CreatureScript
...
+                            _events.ScheduleEvent(EVENT_SUPPRESSION_AURA, Seconds(6)); //I couldnt find out if there is a time between the aura effect and the recast - this results in constantly being under effect of the aura
+                            break;
+                        case EVENT_SUPPRESSION_ANIM:
+                            if (me->GetGoState() == GO_READY)
+                            me->SendCustomAnim(0);
```
MaxtorCoder, line 169: here again

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
Jildor: these checks running each update diff does not seem the most appropriate

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +122,86 @@ class boss_broodlord : public CreatureScript
...
+                            if (me->GetGoState() == GO_STATE_DESTROYED)
+                            me->SetGoState(GO_STATE_READY);
+                            me->SetLootState(GO_READY);
+                            me->RemoveFlag(GAMEOBJECT_FLAGS, GO_FLAG_NOT_SELECTABLE);
+                            break;
+                    }
+                }
+            }
+
+            void OnLootStateChanged(uint32 state, Unit* who) override
```
TranceAccount, line 182: `fatal error: unused parameter 'who' [-Wunused-parameter]`

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +122,86 @@ class boss_broodlord : public CreatureScript
...
+class go_suppression_device : public GameObjectScript
+{
+    public:
+        go_suppression_device() : GameObjectScript("go_suppression_device") { }
+
+        struct go_suppression_deviceAI : public GameObjectAI
+        {
+            go_suppression_deviceAI(GameObject* go) : GameObjectAI(go), _instance(go->GetInstanceScript())
+            {
+                _events.ScheduleEvent(EVENT_SUPPRESSION_AURA, Seconds(6));
```
jackpoz, line 134: I would prefer this to be somewhere else than the constructor

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +122,86 @@ class boss_broodlord : public CreatureScript
...
+                _events.ScheduleEvent(EVENT_SUPPRESSION_ANIM, Seconds(2));
+            }
+
+            void UpdateAI(uint32 diff) override
+            {
+                if (_instance->GetBossState(DATA_VAELASTRAZ_THE_CORRUPT) != DONE)
+                    return;
+
+                if (_instance->GetBossState(DATA_VAELASTRAZ_THE_CORRUPT) == DONE)
```
jackpoz, line 143: The check above returns if not done, isn't this check redundant ?

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +122,86 @@ class boss_broodlord : public CreatureScript
...
+                {
+                    if (me->GetGoState() == GO_STATE_DESTROYED)
+                        me->SetLootState(GO_READY); // NOT doing this will despawn the gameobject after using Disarm Trap
+                }
```
jackpoz, line 147:
```suggestion
                if (me->GetGoState() == GO_STATE_DESTROYED)
                    me->SetLootState(GO_READY); // NOT doing this will despawn the gameobject after using Disarm Trap
```

Jinnaix: thank you :)

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
Jinnaix: thank you :)

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
Jinnaix: thank you :)

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
Jinnaix: Im up for suggestions here but i couldnt come up with something else.
The intention was:
Do nothing until VAEL is dead.
Execute some actions and then do nothing after BROOD is dead.

The check inbetween these 2 is needed, if we dont do this, the gameobject is despawned by 'Disarm Trap'.

Jinnaix:
Thanks for everyones feedback 👍 

For me, unless there is something i am missing, this is done and can be merged if everyone is satisfied or rejected if this isnt the solution that fits for this gameplay mechanic.

For the future reader:

Im not really happy with my checks here:

```cpp
            {
                if (_instance->GetBossState(DATA_VAELASTRAZ_THE_CORRUPT) != DONE)
                    return;

                if (me->GetGoState() == GO_STATE_DESTROYED)
                    me->SetLootState(GO_READY); // NOT doing this will despawn the gameobject after using Disarm Trap

                if (_instance->GetBossState(DATA_BROODLORD_LASHLAYER) == DONE)
                {
                    me->SetGoState(GO_STATE_DESTROYED);
                    me->SetFlag(GAMEOBJECT_FLAGS, GO_FLAG_NOT_SELECTABLE);
                    return;
                };
```

The intention here was/is:

1) Do nothing until Vaelastraz is dead.
2) Set to state destroyed, make yourself unselectable and do nothing as soon as Broodlord is dead.

These kind of game objects don't wait for a player interaction but instead do something all the time.

While there might be smarter solutions for that, the middle check for GoState Destroyed bothers me the most.
Without it, Rogues 'Disarm Trap' will despawn the Suppression Device.

I dont know why exactly this  prevents the gob from being despawned but it does and i couldnt figure out another way of getting the same result.

offl:
What about
data3: 22247 spell (Spell Id from Spell.dbc)
data6: 3600000 autoCloseTime

Jinnaix:
From a players point of view, what i did in this pr is as close to retail as i could get, judging from a dozen videos ive seen from current classic bwl runs. Sure - there are differences (especially in how disarm traps works in classic compared to wotlk), but still.

We utilize the correct spell here and the auto close time, even if seen in sniffs, can be ignored at this point because it doesnt matter for this encounter/mechanic. The Supression Devices have to be active until you kill Broodlord.

Before doing it in c++ i tried countless other ways with sai, always failing at one point or another - mostly at the gameobject despawn other times on the gameobject animation.

But i can look into it again, there are other gob`s that have simliar properties without a c++ script or sai attached to them.

image

offl:
At this point better not to return to SAI since it will be impossible or hard enough
I just wonder can we use data3 and data6 here, maybe it could be scripted more easy, I didn't tried yet to just apply faction change and click on gob without any script

Jinnaix:
Well, it has a lock id - you can only interact with it by disarm trap. And without doing anything else, disarm trap will by default despawn the gameobjekt. So here we are again :D
But I'm totally up for suggestions if you want to try something specific.

Jinnaix:
And yes - i tried it with changing a lot of the database values like gob_temp_addon no despawn flag and other stuff.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +122,85 @@ class boss_broodlord : public CreatureScript
...
+                _events.ScheduleEvent(EVENT_SUPPRESSION_AURA, Seconds(6));
+                _events.ScheduleEvent(EVENT_SUPPRESSION_ANIM, Seconds(2));
+            }
+
+            void UpdateAI(uint32 diff) override
+            {
+                if (_instance->GetBossState(DATA_VAELASTRAZ_THE_CORRUPT) != DONE)
+                    return;
+
+                if (me->GetGoState() == GO_STATE_DESTROYED)
```
Shauren, line 145: This will override state back to GO_READY on next UpdateAI even if DATA_BROODLORD_LASHLAYER is DONE, you would need to move this part after that check

Jinnaix: Thank you! :)

Faq:
> 
> ... disarm trap will by default despawn the gameobjekt. So here we are again :D
> 

Then mby needed some changes in core where trap is not despawning gob in some cases?
@xvwyh any input?

xvwyh:
Don't know, sorry. When I was implementing it years ago, Classic didn't exist yet and I had very little info to work with, Disarm Trap also makes them despawn on my server and I couldn't have known that it shouldn't work like that. I had only one late-expansion sniff to work with, and that one had a grant total of zero disarms, so I simply have no idea of what's happening to them under the hood when they are disarmed. However, I did learn something else from that sniff, so, as per my tradition, I'm gonna throw in some more headache for you.

There are 3 World Invisible Triggers (12999) constantly walking across the suppression room, patrolling back and forth, at a leisurely walking speed.

image
<details>
<summary>Screenshots from Waypoint Creator tool</summary>
image
image
image
</details>

They do absolutely nothing in the sniffs by themselves, just patrol around.

But another fascinating fact is that the traps do occasionally (and seemingly randomly) disarms themselves! Temporarily. For about 6 seconds. They change state to `GO_STATE_ACTIVE` and set `GO_FLAG_IN_USE`, stop casting the spell and playing animations, then, 6 seconds later, they restore back to `GO_STATE_ACTIVE` and one more second later they restore `GO_FLAG_IN_USE`. This happens constantly throughout the room.

What I think is happening here is that World Invisible Triggers deactivate the traps as they walk over them. I'm pretty sure I checked this back in the day, but I can't verify this quickly right now because that movement of World Invisible Triggers is one long continuous spline, so I can't know for certain where they are located at any particular time.

Additionally, now that I know traps occasionally reset from being disarmed after playing Classic, I think this isn't just a random timer, but is also tied to those World Invisible Triggers. When they walk over them - they temporarily deactivate them, even if they're already disarmed, but after 6 seconds they reset back to undisarmed state, and since players can't see these World Invisible Triggers, to them it looks like they are popping back up randomly, hence the huge randomness in the reset timer (30-120 seconds in this PR).

I'll check my video recordings of BWL runs, but I just wanted to drop this here in the meantime, to make your day worse. You're welcome!

xvwyh:
The most fascinating thing is that I cannot see all that I mentioned above happening on Classic! But it's most definitely present in my sniff. Maybe this was a late addition to aid players in solo farming BWL who aren't rogues with disarm trap? Throws a wrench into my send-to-last paragraph in the previous comment, if there were no World Invisible Triggers during vanilla, then yeah, it must've just been a dumb random timer.

Additionally, from my sniff, Suppression Devices are casting and playing their animation exactly every 4 or 5 seconds. Randomly either 4 or 5 seconds, but always exactly on the second, never in-between (which may as well speak to how often their AI is updated, I dunno). Cast and animation always go hand-in-hand, there are no 2 separate timers as in this PR:
> 03:52:43.985 cast & anim
> 03:52:48.985 cast & anim
> 03:52:53.985 cast & anim
> 03:52:58.985 cast & anim
> 03:53:03.985 cast & anim
> 03:53:09.000 cast & anim
> zone out, zone in
> 03:53:34.922 cast & anim
> 03:53:39.875 cast & anim
> 03:53:42.172 disabled by patrolling trigger
> 03:53:48.250 restored state
> 03:53:49.563 removed in_use flag
> 03:53:50.657 cast & anim
> 03:53:53.094 disabled by patrolling trigger
> 03:53:59.188 restored state
> 03:54:00.516 removed in_use flag
> 03:54:01.579 cast & anim
> 03:54:04.063 disabled by patrolling trigger
> 03:54:10.125 restored state
> 03:54:11.454 removed in_use flag
> 03:54:12.532 cast & anim
> 03:54:17.532 cast & anim
> 03:54:22.532 cast & anim
> 03:54:27.532 cast & anim
> 03:54:32.532 cast & anim
> 03:54:37.547 cast & anim
> 03:54:42.563 cast & anim
> 03:54:47.547 cast & anim
> 03:54:52.563 cast & anim
> 03:54:57.563 cast & anim

This causes a small visual problem, tho. The animation lasts longer than 5 seconds, and it doesn't restart if animation packet is send before its finished. So in reality the animation will only play once every 2 casts (once every 10 seconds). This was fixed in Classic, tho, the animation restarts even if the previous one wasn't fully finished, so the next wave of smoke particles is emitted even before the previous one expired.

Jinnaix:
> This causes a small visual problem, tho. The animation lasts longer than 5 seconds, and it doesn't restart if animation packet is > sent before it's finished.

That is the main reason i separated the cast from the animation.
I didnt had any sniffs for this PR, but @Shauren could confirm the faction change in irc. While i watched a truckload of Videos, i also noticed that in classic right now, the animations arent  synced for all gobs,  but also arent random. Without overextending it, this was the best i could come up with.

For the 3 triggers, i assume that this was a later addition and not present in vanilla/tbc/wotlk as you also stated in your second post. 

I would like to at least decrease the cast time down to 4-5 seconds.
While this might be not technically correct in terms of what we see in sniffs, there is now only a barely noticeable difference from this PR to how it is in classic/would be in wotlk from a players point of view.
The spell is correct, the mechanic to disable them is correct, their activity is bound to Broodlord.

Replicating whatever Blizzard might have done here without correct data from back then might be tough.

Thoughts?

xvwyh:
> I would like to at least decrease the cast time down to 4-5 seconds.

I dunno what I was smoking an hour ago, but now that I've looked at the timers I myself copy-pasted in my previous comment, it's always every 5 seconds, no randomness. Sorry.

https://www.youtube.com/watch?v=nZVUNVKOmoE

Here's a video from the real vanilla. Sometimes the animation repeats after 5 seconds, sometimes - after 10 seconds. I don't know anymore... How rich would it be to attribute it to server/network lags? :D Anyway, in my recordings of modern Classic 5 seconds seems to be reliable, while on a 3.3.5 client it doesn't work because the previous animation is still playing. Sniff confirms the interval is 5 seconds. So make it 5, and be what may.

> Thoughts?

Yeah, I guess forget about my first comment entirely. Must've been a late addition.

https://youtu.be/8al6ZjVDvOY?t=452
This is 7.2.5, you can see how one of the devices is deactivated, despite the player going through the room for the first time:
image
And there's one deactivating on its own a few ~~seconds~~ frames earlier.
https://youtu.be/wQ3X7Nd9aoU?t=637
Here's another one, deactivating and reactivating on its own in the distance, and then, right before the mage blinks, you can see the next one on the Trigger's path deactivating too. And one more example in the next room.

Anyway, this is just to prove to myself that I'm not going mad. We don't know if these Triggers were added before or after 3.3.5.

xvwyh:
> the animations arent synced for all gobs, but also arent random. Without overextending it, this was the best i could come up with.

Just randomize the initial timer, the one you currently have in the constructor. Make it random between 0s and 5s. That will be enough.

> the mechanic to disable them is correct

Dunno how correct is your implementation without seeing any sniffs that feature disarm trap being used on them, sniffs which I don't possess. Using GO_STATE_DESTROYED smells like a hack. But it's not like TC is free from hacks. :D

> their activity is bound to Broodlord.

One thing I may add to that is to add one more `if` to avoid unnecessary work:
```cpp
                if (_instance->GetBossState(DATA_BROODLORD_LASHLAYER) == DONE)
                 {
                     if (me->GetGoState() != GO_STATE_DESTROYED)
                     {
                         me->SetGoState(GO_STATE_DESTROYED);
                         me->SetFlag(GAMEOBJECT_FLAGS, GO_FLAG_NOT_SELECTABLE);
                     }
                     return;
                 };
```
Also makes it just a little bit more clear that this is supposed to be a one-time operation, not something that should be performed continuously on every trap when Broodlord is dead.

Jinnaix:
>Dunno how correct is your implementation without seeing any sniffs that feature disarm trap being used on them, sniffs which I don't possess. Using GO_STATE_DESTROYED smells like a hack. But it's not like TC is free from hacks. :D

Well, correct in terms of : This is the only state that puts them down into the floor. Not necessarily correct in terms of "this is how it was done" xD

>(ideally it should be moved to InstanceScript::SetBossState, but it just complicates the code (you will now have to store GUIDs of all spawned traps) without achieving any real performance benefit)

Yes, i had that thought too. But i didnt do it because of the same reason - the benefit doesnt really justify the work - in the end, these are only 38 gobs times number of active BWL raids.
In the not so far future, we will contribute our BWL Boss script rewrites. Maybe i will touch it again then.

Thank you very much @xvwyh for your time and insight in this topic.

xvwyh:
> This is the only state that puts them down into the floor.

`GO_STATE_ACTIVE` also puts them down in the floor. `GO_STATE_ACTIVE` is used for temporary deactivation in later expansions, but yeah, we can't say for certain what should be used without sniffs.

Also address this part, please, before you're done with this PR:

> Just randomize the initial timer, the one you currently have in the constructor. Make it random between 0s and 5s. That will be enough.
> 
> I mean get rid of EVENT_SUPPRESSION_ANIM entirely, randomize the initial EVENT_SUPPRESSION_AURA and use it both to cast the spell and play the anim, in that order. They always come together.

Jinnaix:
@xvwyh 
While combining these 2 actions into one still let the spell fire correctly, the animation is now way longer as it should - roughly ~9 seconds before the loop begins. Still want me to push it that way?

xvwyh:
Yes. This is what I've been saying all along:

> This causes a small visual problem, tho. The animation lasts longer than 5 seconds, and it doesn't restart if animation packet is sent before it's finished. So in reality the animation will only play once every 2 casts (once every 10 seconds). **This was fixed in Classic**, tho, the animation restarts even if the previous one wasn't fully finished, so the next wave of smoke particles is emitted even before the previous one expired.

> https://www.youtube.com/watch?v=nZVUNVKOmoE
> 
> Here's a video from the real vanilla. Sometimes the animation repeats after 5 seconds, sometimes - after 10 seconds. I don't know anymore... How rich would it be to attribute it to server/network lags? :D Anyway, in my recordings of modern Classic 5 seconds seems to be reliable, while on a 3.3.5 client it doesn't work because the previous animation is still playing. Sniff confirms the interval is 5 seconds. So make it 5, and be what may.

This bug is on blizzard, don't bother fixing it. :P

xvwyh:
Ok, now you've made a mess. Let me fix it for you:
```cpp
                        case EVENT_SUPPRESSION_CAST:
                             if (me->GetGoState() == GO_STATE_READY)
                             {
                                 me->CastSpell(nullptr, SPELL_SUPPRESSION_AURA, true);
                                 me->SendCustomAnim(0);
                             }
                             _events.ScheduleEvent(EVENT_SUPPRESSION_CAST, Seconds(5));
                             break;
```
C++ isn't python, indentation doesn't matter, you group multiple statements by surrounding them with { }.
I also put cast before the animation, as that it how it appears in sniffs.

Also, one more thing:
```diff
-                if (me->GetGoState() == GO_STATE_DESTROYED)
+                if (me->getLootState() == GO_JUST_DEACTIVATED)
                      me->SetLootState(GO_READY); // NOT doing this will despawn the gameobject after using Disarm Trap
```
This should still be working, but not causing unnecessary work when loot state is already `GO_READY`.

Still unsure if `GO_STATE_DESTROYED` is needed, I think `GO_STATE_ACTIVE` would do just fine. Try replacing all occurrences of `GO_STATE_DESTROYED` with `GO_STATE_ACTIVE` in your script and check if everything still works fine.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +121,85 @@ class boss_broodlord : public CreatureScript
...
+
+                while (uint32 eventId = _events.ExecuteEvent())
+                {
+                    switch (eventId)
+                    {
+                        case EVENT_SUPPRESSION_CAST:
+                            if (me->GetGoState() == GO_STATE_READY)
+                            {
+                                me->SendCustomAnim(0);
+                                me->CastSpell(nullptr,SPELL_SUPPRESSION_AURA,true);
```
Carbenium, line 165:
```suggestion
                                me->CastSpell(nullptr, SPELL_SUPPRESSION_AURA, true);
```

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +121,85 @@ class boss_broodlord : public CreatureScript
...
+                            }
+                            _events.ScheduleEvent(EVENT_SUPPRESSION_CAST, Seconds(5));
```
Carbenium, line 168:
```suggestion
                            _events.ScheduleEvent(EVENT_SUPPRESSION_CAST, 5s);
```

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
```diff
@@ -113,7 +121,85 @@ class boss_broodlord : public CreatureScript
...
+    public:
+        go_suppression_device() : GameObjectScript("go_suppression_device") { }
+
+        struct go_suppression_deviceAI : public GameObjectAI
+        {
+            go_suppression_deviceAI(GameObject* go) : GameObjectAI(go), _instance(go->GetInstanceScript()) { }
+
+            void InitializeAI() override
+            {
+                _events.ScheduleEvent(EVENT_SUPPRESSION_CAST, Seconds(0), Seconds(5));
```
Carbenium, line 135:
```suggestion
                _events.ScheduleEvent(EVENT_SUPPRESSION_CAST, 0s, 5s);
```

Carbenium (commented):
Just some stylistic suggestions.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
Jinnaix: Doesnt changes like this create a std::chrono overload warning?

Jinnaix:
Now i fail with @ mentions in commits. Shouldnt do this while being sober, where is my beer?... 
Thanks once again everyone. Should contain all feedback now. Still works as intended.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_broodlord_lashlayer.cpp
Carbenium: No, they are exactly the same type. 

The literal is defined like something along these lines:
```c++
constexpr Seconds operator ""s(unsigned long long s)
{
    return Seconds(s);
}
```
What wouldn't work is using a type with greater precision than milliseconds, since the `ScheduleEvent` accepts `Milliseconds` as parameter. In that case `std::chrono::duration_cast` has to be used to force the conversion to the less precise type.

Jinnaix: Thanks for that explanation :)

Jinnaix:
Just a small note for @jackpoz -  if you look over this: from my perspective, this PR is ready and can be merged.
I have covered all feedback and gave it a ingame check everytime i changed something.

Jinnaix:
Please don't merge yet - I will try do rework some stuff related to feedback from treestone

Jinnaix:
Update:
We get rid of all the boss state checks on every update diff (and the anti-despawn too) and use clean functions for all wanted actions now.
Functionality is the same as before, this only changes technical stuff.
Thank you very much @Treeston for all the time spend and help you provided on that.

Treeston (approved):

illfated:
Thank you for your PR! 👍

PR #25202, opened 2020-08-04
Scripts/Naxxramas: Frogger event improvements

Jildor:
**Changes proposed:**

-  Prevent Living Poison accumulation
-  Now Living Poison cast Explode
-  Updates: https://github.com/TrinityCore/TrinityCore/issues/19072#issuecomment-667949712

~~Edit: There is still a problem when you are in the instance, but you are away from the npc spawn zone, because event still run but NPC doesn't despawn, I have no idea how to solve that properly, so I prefer not to touch that.~~

Edit2: I found a way, I hope can be valid ;)

**Target branch(es):** 3.3.5/master

- [X] 3.3.5
- [ ] master

**Issues addressed:** None

**Tests performed:** tested in-game

PR #25214, opened 2020-08-06
Scripts/Naxxramas: Adjust Frogger event timer

Jildor:
**Changes proposed:**

-  Adjust summonTime duration
-  Adjust Event Timer depending on _nextFroggerWave
- NPC speed updated

According to @offl comment https://github.com/TrinityCore/TrinityCore/pull/25202#issuecomment-668746460
here a video of how it should work: https://www.youtube.com/watch?v=EhkrER775E8
and here result after this PR: https://imgur.com/sWqba34

- [X] 3.3.5
- [ ] master

**Issues addressed:** None


Ovahlord:
Let me give you a hand with cleaning up this mess:
https://paste2.org/XahYLBc5

So first of all, all these timers are wrong. The waves are spawned in 6 second intervals . I've just documented one of them since they are pretty much identical. Next I've corrected the spline movement behaivior and got rid of the waypoint_data part since waypoints are poorly implemented anyways. Then I've corrected the despawn mechanics and use the correct movement behaivior.

Jildor:
Thanks, that makes more sense, I tried to adjust the timer watching the video so that it stays the same, but it's not the same :)
let me try

Jildor:
updated ;)
https://imgur.com/a/JVSRHEu

src/server/scripts/Northrend/Naxxramas/naxxramas.cpp
```diff
@@ -46,19 +89,26 @@ struct npc_frogger_trigger_naxx : public ScriptedAI
...
-                {
-                    std::list<TempSummon*> spawns;
-                    me->SummonCreatureGroup(_nextFroggerWave, &spawns);
-                    if (!spawns.empty())
+                case EVENT_SUMMON_LIVING_POISON:
+                    for (uint8 i = 0; i < MaxLivingPoisons; ++i)
                     {
-                        spawns.front()->GetMotionMaster()->MovePath(10 * NPC_FROGGER + _nextFroggerWave, false);
-                        spawns.front()->UpdateSpeed(MOVE_SWIM); // We need update swim speed because spawn in "water"
+                        if (Creature* slime = DoSummon(NPC_LIVING_POISON, LivingPoisonSpawnPositions[i], 8500/*8s + 500ms*/, TEMPSUMMON_CORPSE_TIMED_DESPAWN))
```
Ovahlord, line 95: yeah, that should be 8s + 500ms for 335 before carbenium comes crawling out of his cave and slaps me again

Carbenium: `8500ms` would have been also fine :yum:

Jildor: which do you prefer?
do I change it or leave it like that?

Carbenium: Leave it. That's totally fine with me. The non-chrono version actually shouldn't even compile, since those overloads are gone

PR #25220, opened 2020-08-06
Core/Updater: Fix detection of the mysql binary when a directory is given as path

### LINKED ISSUE #25216: Worldserver crashes with "Caught Signal 22"
**Description:**

Deviating from the official guide on Atlassian, I tried to use my self compiled OpenSSL and Boost libraries. I managed to compile TrinityCore (3.3.5) without errors after some adaptions in cmake which were:

LIB_EAY_DEBUG="S:/Installationsfrei/wowps/Dependencies/OpenSSL/Debug/lib/libcrypto.lib"
LIB_EAY_RELEASE="S:/Installationsfrei/wowps/Dependencies/OpenSSL/lib/libcrypto.lib"
SSL_EAY_DEBUG="S:/Installationsfrei/wowps/Dependencies/OpenSSL/Debug/lib/libssl.lib"
SSL_EAY_RELEASE="S:/Installationsfrei/wowps/Dependencies/OpenSSL/lib/libssl.lib"

I compiled OpenSSL two times. Once with and once without the "--debug" option. To Compile OpenSSL on Windows, I basically followed these instructions: https://stackoverflow.com/questions/45494630/how-to-build-openssl-on-windows-with-visual-studio-2017
But adapted to Visual Studio 2019.

Also, I used the .zip-File version of MySQL to avoid installing it.

**Expected behaviour:**

After testing the connection to the MySQL server, configuring worldserver.conf accordingly and starting worldserver.exe, the databases auth, world and character should be populated.

**Actual behaviour:**
worldserver.exe crashes with this error: "Caught Signal 22" (See also the crashlog at the end).

1. Follow the steps described here: https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130077/Installation+Guide
2. But compile OpenSSL and Boost from source.
3. Adapt setting in CMake.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. 53effff9b2ee 2020-08-05 20:43:11 +0300 (3.3.5 branch) (Win64, RelWithDebInfo, Static) (worldserver-daemon)

**Operating system:**
Windows 10 64-bit (2004)

**Crashlog:**
https://pastebin.com/URUVuSJ5

Carbenium:
**Changes proposed:**

Fix detection of the mysql binary when a directory is given as path.
This ensures that we never pass a directory path StartProcess which results in a crash.

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #25216

**Tests performed:**

started worldserver with updates to apply

### DIFF
--- a/src/server/database/Updater/DBUpdater.cpp
+++ b/src/server/database/Updater/DBUpdater.cpp
@@ -40,10 +40,10 @@ std::string DBUpdaterUtil::GetCorrectedMySQLExecutable()
 bool DBUpdaterUtil::CheckExecutable()
 {
     boost::filesystem::path exe(GetCorrectedMySQLExecutable());
-    if (!exists(exe))
+    if (!is_regular_file(exe))
     {
         exe = Trinity::SearchExecutableInPath("mysql");
-        if (!exe.empty() && exists(exe))
+        if (!exe.empty() && is_regular_file(exe))
         {
             // Correct the path to the cli
             corrected_path() = absolute(exe).generic_string();


src/server/database/Updater/DBUpdater.cpp
```diff
@@ -40,10 +40,10 @@ std::string DBUpdaterUtil::GetCorrectedMySQLExecutable()
 bool DBUpdaterUtil::CheckExecutable()
 {
     boost::filesystem::path exe(GetCorrectedMySQLExecutable());
-    if (!exists(exe))
+    if (is_directory(exe))
```
Shauren, line 43: shouldnt this be !is_regular_file?

Carbenium: Hmm, I think we should even do something like
```c++
if (is_regular_file(exe))
    return true;
if (is_directory(exe))
{
    ...
}
return false;
```
With that we would be able to catch the `is_block_file`, `is_character_file`, etc. cases.

Naios: I think `is_regular_file` would be the best choice here too.

Shauren: Im pretty sure we only care that is_regular_file(exe) -> good
else search in path ignoring config

possibly maybe is_symlink? (but then you need to check link target) nothing else should be treated as valid entry

Carbenium: Argh, sorry. I misinterpreted `SearchExecutableInPath`

In this case I agree, let's do `!is_regular_file`

Issue #24912, opened 2020-06-28
[3.3.5] Caster monsters don't attempt to melee after counterspell


Trystanosaurus:
**Description:**

Currently, if you counterspell (and presumably use other spell school locking abilities) on a caster type monster they won't attempt to melee your character, this makes silence pulls impossible

**Expected behaviour:**

You counterspell an enemy while casting.
They run at you ant try to melee attack.

**Steps to reproduce the problem:**

1. Create a mage, level 20+ with counterspell (https://www.wowhead.com/spell=2139/counterspell)
2. Go to an area with caster enemies, e.g. the Ango'rosh Souleaters in Ango'rosh Stronghold (.tele Angorosh)
3. Cast Counterspell while they're casting and see that they just stand there.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

54c701cf0db8

**Operating system:**

Windows 10

offl:
Why https://github.com/TrinityCore/TrinityCore/issues/20222#issuecomment-327015832
is completely different from https://github.com/TrinityCore/TrinityCore/commit/1e8227eda7f9c6a989fe6c928c857d76491e008c

offl:
Eww, it was updated in https://github.com/TrinityCore/TrinityCore/commit/668df7fd01a2b3bdac1ea4600ed17df03e51bd60
So problem is in another place

jackpoz:
works fine for me:
https://i.imgur.com/3CgGA82.gifv

offl:
Tried multiple times with Counterspell and it always does not work
Spelllock works but not always

jackpoz:
did you try that same mob or a different one ? different sai action flags (combat_move) ? or npcs that aren't even SAI npcs ?

offl:
I tried that same mob and too tempspawned him like you did

jackpoz:
Tried again on the test realm with https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=18121 , it works fine:
https://imgur.com/8Sr2mA1

jackpoz:
turns out countspell silences only a school by default but with the talent improved counterspell it silences the creature.
I created a PR

jackpoz:
Fixed silence for a single school, the only case left is spell being interrupted mid cast

PR #25226, opened 2020-08-08
Core/SAI: Fix creatures casting with flag SMARTCAST_COMBAT_MOVE not switching to melee when the school of the spell they are trying to cast gets silenced (and other cases of spell failure)

### LINKED ISSUE #24914: Core/SAI: When timer expires in combat, it should not reset until the cast is successful
**Description:**

Currently timers expires if spell was not used because the cast was not succesful(creature was stunned, for example). Or player was out of spell range. Or because creature used another spell at this moment: this is why we interrupt one spell to cast another and more important(healing spells, for example). Blizzard never interrupts one spell to cast another. They even don't interrupt one phase to enter in new.

**Expected behaviour:**

When timer expires in combat, it should not reset until the cast is successful.

apply sql
update smart_scripts set event_param1 = 5000, event_param2 = 5000, event_param3 = 15000, event_param4 = 15000 where entryorguid=26426 and source_type=0;
1. .tele test and .gm off  and .mod speed 10
2. use clocks but don't start the timer
3. .npc add temp 26426
4. aggro creature when you are in melee and start timer
5. wait for first cast, creature will cast spell after 5 sec
6. now go away from creature for next 20 sec, don't allow it to be in melee range with you
7. after 20 sec go back in melee range
here creature must to cast spell on you but it doesn't

**Branch(es):** both

**TC rev. hash/commit:** 3ac1992afbeb

jackpoz:
**Changes proposed:**

-  Fix creatures casting with flag SMARTCAST_COMBAT_MOVE not switching to melee when the school of the spell they are trying to cast gets silenced (and other cases of spell failure)
- Retry casting after 500 ms if a cast failed

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Improves #24912
Fixes #24914

**Tests performed:**

- https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=18121 using https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=2139 with and without https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=11255

- [ ] It would be nice if the mobs retried to cast when interrupted

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -552,23 +552,25 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                         if (e.action.cast.castFlags & SMARTCAST_INTERRUPT_PREVIOUS)
                             me->InterruptNonMeleeSpells(false);
 
+                        SpellCastResult result = me->CastSpell(target->ToUnit(), e.action.cast.spell, triggerFlag);
+
                         if (e.action.cast.castFlags & SMARTCAST_COMBAT_MOVE)
                         {
                             // If cast flag SMARTCAST_COMBAT_MOVE is set combat movement will not be allowed unless target is outside spell range, out of mana, or LOS.
                             bool allowMove = false;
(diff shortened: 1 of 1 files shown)


offl:
> It would be nice if the mobs retried to cast when interrupted

Probabably https://github.com/TrinityCore/TrinityCore/issues/24914, I'm not sure

PR #25229, opened 2020-08-09
Core/AI: Remove code that removed UNIT_DYNFLAG_LOOTABLE from possessed units

### LINKED ISSUE #24060: When a creature charmed by a player kills another creature, the killed creature cannot be looted.
Current behaviour:
Charming a creature and killing another creature with it will remove the dyno lootable flag without checking if the charmer has met the damage requirement to receive loot.

Expected behaviour:
Charmed unit should check if the charming player has met the damage requirement before disabling the lootable flag.
However, I'm not sure if this is correct for all cases.

How to reproduce:
1. Go to Naxxramas 10 man Razuvious.
2. Damage Razuvious to low health.
3. Charm an understudy using the crystal and kill Razuvious with it.
4. No lootable flag.

Branch: 3.3.5, not sure about master.
TrinityCore rev. f1c338f1371e 2020-01-12 22:39:09 +0100 (3.3.5 branch)

Notes:
The issue for this case lies here:
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/CoreAI/PassiveAI.cpp#L75

```
void PossessedAI::KilledUnit(Unit* victim)
{
    // We killed a creature, disable victim's loot
    if (victim->GetTypeId() == TYPEID_UNIT)
        victim->RemoveFlag(UNIT_DYNAMIC_FLAGS, UNIT_DYNFLAG_LOOTABLE);
}
```

This check does not check for charmer damage requirements and this check should also probably be done by the killed unit, not the killer.

jackpoz:
**Changes proposed:**

-  Remove code that removed UNIT_DYNFLAG_LOOTABLE from possessed units

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #24060

**Tests performed:**

None, it needs to be tested

- [ ] Needs to be tested
- [ ] No idea why the code was there (as usual no comments at all in the code about WHY the code is there in the first place)

### DIFF
--- a/src/server/game/AI/CoreAI/PassiveAI.cpp
+++ b/src/server/game/AI/CoreAI/PassiveAI.cpp
@@ -67,6 +67,12 @@ void PossessedAI::UpdateAI(uint32 /*diff*/)
     }
 }
 
+void PossessedAI::JustDied(Unit* /*u*/)
+{
+    // We died while possessed, disable our loot
+    me->RemoveFlag(UNIT_DYNAMIC_FLAGS, UNIT_DYNFLAG_LOOTABLE);
+}
+
 void CritterAI::JustEngagedWith(Unit* /*who*/)
 {
     if (!me->HasUnitState(UNIT_STATE_FLEEING))
--- a/src/server/game/AI/CoreAI/PassiveAI.cpp
+++ b/src/server/game/AI/CoreAI/PassiveAI.cpp
@@ -67,19 +67,6 @@ void PossessedAI::UpdateAI(uint32 /*diff*/)
     }
 }
 
-void PossessedAI::JustDied(Unit* /*u*/)
-{
-    // We died while possessed, disable our loot
-    me->RemoveFlag(UNIT_DYNAMIC_FLAGS, UNIT_DYNFLAG_LOOTABLE);
-}
-
-void PossessedAI::KilledUnit(Unit* victim)
-{
-    // We killed a creature, disable victim's loot
-    if (victim->GetTypeId() == TYPEID_UNIT)
-        victim->RemoveFlag(UNIT_DYNAMIC_FLAGS, UNIT_DYNFLAG_LOOTABLE);
-}
-
 void CritterAI::JustEngagedWith(Unit* /*who*/)
 {
     if (!me->HasUnitState(UNIT_STATE_FLEEING))
--- a/src/server/game/AI/CoreAI/PassiveAI.h
+++ b/src/server/game/AI/CoreAI/PassiveAI.h
@@ -45,9 +45,6 @@ class TC_GAME_API PossessedAI : public CreatureAI
         void UpdateAI(uint32) override;
         void EnterEvadeMode(EvadeReason /*why*/) override { }
 
-        void JustDied(Unit*) override;
-        void KilledUnit(Unit* victim) override;
-
         static int32 Permissible(Creature const* /*creature*/) { return PERMIT_BASE_NO; }
 };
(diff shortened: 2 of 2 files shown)


illfated:
OK, so the code removal builds without any issues (at least for Travis & AppVeyor), so I guess all we need to find out is if this change causes any loot to become available incorrectly.

Jildor:
added here: https://github.com/TrinityCore/TrinityCore/commit/44bdb135f4ed5bb834795cfb928ccbb0504cf834#diff-bd2a90cd7b17b69966c8094fd8255f54R113
can it affect players' pets?

offl:
Maybe it was added to prevent players possess overpowered mobs and kill unlimited count of lootable mobs?

illfated:
> Maybe it was added to prevent players possess overpowered mobs and kill unlimited count of lootable mobs?

Hmm, interesting thought. Isn't there any limitation of how high level mob a player can charm or possess? (I have not used that skill much myself, at least not to know the answer to this question.)

Jinnaix:
Without any further research, I would suggest making an exception for dungeons and only have it active for base maps (azeroth, kalimdor, Outland, northrend). We currently don't know of any abusive potential. There might have been a reason for this code in the first place. Thoughts?

jackpoz:
The whole https://github.com/TrinityCore/TrinityCore/issues/24060 issue is about Naxx, which is what this PR is targetting

Jinnaix:
While the current state of the code is clearly wrong, this PR will fix problems in Naxx but creates another wrong state too.

As taken from https://wow.gamepedia.com/Mind_Control
>Mobs must take damage from players in order to generate loot; should they be killed solely by a controlled mob they will drop no items - the same rule as with any other pet. In addition, a controlled mob will generate no loot if it dies while still controlled; it must be released if you wish to take its items.

This information was also aviable on older revisions (2010 https://wow.gamepedia.com/index.php?title=Mind_Control&oldid=2378715 , 2008 https://wow.gamepedia.com/index.php?title=Mind_Control&oldid=1560268) and i cant find anything that states it as wrong.

jackpoz:
@Jinnaix I updated the PR restoring the PossessedAI::JustDied() handler to remove the lootable flag if the unit is possessed on death

Jinnaix:
image

offl:
Why? It's not implemented
> Mobs must take damage from players in order to generate loot; should they be killed solely by a controlled mob they will drop no items - the same rule as with any other pet.

It probably creates dangerous exploits

Jinnaix:
Sorry, this Pull Request is missing conversation from Irc/Discord.
The logic exists (for players->groups->raids) and applies here too. Therefore we can safely remove the KilledUnit function.
If you solely kill a mob with a possessed unit, you dont get lootaccess. If you damage the mob, possess a unit and kill the damaged mob with the said possessed unit, you get access to this creatures loot. (This also works if the damage comes from your dots later on or from members of your group/raid)

offl:
So it's implemented somewhere else?

Jinnaix:
Yes, we have a function for PlayerDamageReq in creature.cpp that applies to this case. The former code that gets removed in this PR just killed the whole logic. What we want to keep is, that you have to release the possessed unit before it dies to prevent exploiting (and because of - its blizzlike).

natnat201193:
As I said on the original report, it's not checking for charmer damage requirements as it should.

"This check does not check for charmer damage requirements and this check should also probably be done by the killed unit, not the killer."

Thanks for looking into it :)

PR #25233, opened 2020-08-10
Switch from instance->GetPlayers loop to InstanceScript::DoOnPlayers

Traesh:
**Changes proposed:**

-  Replace players loop by new InstanceScript DoOnPlayers method
-  Use this new method in a few scripts as example
-  Goal is to provide cleaner code and allow to do action on all players instances without creatint new method in InstanceScript each time

**Target branch(es):**

- [ ] 3.3.5
- [X] master

**Issues addressed:**

None

**Tests performed:**

Build

None


src/server/game/Instances/InstanceScript.h
```diff
@@ -209,6 +209,9 @@ class TC_GAME_API InstanceScript : public ZoneScript
         // Respawns a GO having negative spawntimesecs in gameobject-table
         void DoRespawnGameObject(ObjectGuid guid, uint32 timeToDespawn = MINUTE);
 
+        // Execute the parameter function for all players in instance
+        void DoOnPlayers(std::function<void(Player*)>&& function);
```
Treeston, line 213: Why is this not on `Map`?

src/server/game/Instances/InstanceScript.cpp
```diff
@@ -535,99 +535,87 @@ void InstanceScript::DoRespawnGameObject(ObjectGuid guid, uint32 timeToDespawn /
...
 
 // Send Notify to all players in instance
 void InstanceScript::DoSendNotifyToInstance(char const* format, ...)
 {
-    InstanceMap::PlayerList const& players = instance->GetPlayers();
+    va_list ap;
+    va_start(ap, format);
+    char buff[1024];
+    vsnprintf(buff, 1024, format, ap);
+    va_end(ap);
```
Treeston, line 563: No.

src/server/game/Instances/InstanceScript.cpp
```diff
@@ -535,99 +535,87 @@ void InstanceScript::DoRespawnGameObject(ObjectGuid guid, uint32 timeToDespawn /
         TC_LOG_DEBUG("scripts", "InstanceScript: DoRespawnGameObject failed");
 }
 
-void InstanceScript::DoUpdateWorldState(uint32 uiStateId, uint32 uiStateData)
+void InstanceScript::DoOnPlayers(std::function<void(Player*)>&& function)
```
Treeston, line 538: My gut disagrees with needing to pun to std::function here.

If this is moved to `Map`, it can be fully in the header without issue, since `Map::PlayerList` is visible in that header.

```c++
template <typename T>
void DoOnPlayers(T&& fn)
{
    for (auto const& ref : GetPlayers())
        if (Player* player = ref.GetSource())
            fn(player);
}
```

PS is it even possible for source to be nullptr there?

src/server/game/Instances/InstanceScript.cpp
MaxtorCoder: MonkaS

Shauren: This is a refactor of existing method already doing thing, don't complain, its not new code

src/server/game/Instances/InstanceScript.cpp
Shauren: No, but this mistake is repeated everywhere

Treeston: Sounds like a perfect opportunity to fix this then.

Shauren: Well, there is one case when this can be nullptr and thats when during such a loop, player object is freed (which is never)

Treeston: Aren't player objects only freed at the end of the tick?

Is it possible to cause iterator invalidation during this loop (by teleporting players etc.?)

Shauren: No, none of the things you mentioned happen in: 1) map update 2) a loop iterating player list on map

Treeston: Well, assuming some script does something silly where it does `DoOnPlayers([](Player* p) { p->TeleportTo(wherever); })`

jackpoz:
@Traesh do you have time to address the review comments ?

Nyr97:
Any update on this draft or is it abandoned?

Issue #23035, opened 2019-02-08
[3.3.5] Warden


lilirara:
**Description:**

Warden implemented but not used for reasons we expect.

**Current behaviour:**

I can see multiple opened issues with the same subject that were actually dodged by the TrinityCore developers. TrinityCore is open-source project, and I really thank you guys for the work done. Still in this feeling of open-source where people are co-helping/working, I'm trying to create this issue hoping someone might understand the need and try to fix it.

Nowadays (on 3.3.5 servers at least), most of the players are using PQR/Gagarin softwares that injects the wow client to access or unlock protected LUA API.
These injectors are not UPDATED since a while (about more than 5+ years at least) so it could be easly decompiled I guess, and be understood like TrinityCore did for developing this educational game-server.

It's obviously related to TrinityCore since warden is existing for that kind of cheats.

**Expected behaviour:**

I would like you guys trying to commit a fix for PQR which is actually used by every random players.
This software is destroying the game.

Nowadays, it's impossible to find any tutorial about warden. Of course, github issues are not a forum, but exists for improve TrinityCore and players experiences. PQR is like a game leak which must be fixed.

**Steps to reproduce the problem:**

/

**Branch(es):**

3.3.5
All Systems

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you and add the missing information to **this issue**. Thanks.

Meltie2013:
@lilirara Keep in mind, this is a educational development and not meant for public servers. For something small as signatures for Warden are not hard too update by yourself.

Here's some examples that I have:
```
(810, 243, '', '', 7728616, 2, '7503', 'Gagarin Fly 1'), 
(811, 243, '', '', 9993834, 2, '7452', 'Gagarin Fly 2'), 
(812, 243, '', '', 7248789, 6, '8DB0F4010000', 'Gagarin Warden Disabler'), 
(813, 243, '', '', 7277083, 6, '0F8691000000', 'Gagarin Speed'), 
(814, 243, '', '', 9995134, 2, '755C', 'Gagarin Fix Fly'), 
(815, 243, '', '', 7714472, 2, '7506', 'Gagarin Wall Climb'), 
(816, 243, '', '', 5419615, 6, '0F8CEDFDFFFF', 'Gagarin AntiAFK'), 
[...]

PR #25235, opened 2020-08-11
Core/Warden: Warden refactors

Treeston:
Make Warden readable and maintainable.

_(Did you know current Warden allocates a 12kb heap array for every single player, and they all contain the same (constant) data that never changes? [Yeah.](https://github.com/TrinityCore/TrinityCore/blob/47ba15808bb47ec8265b39174aa5c6cc8715d68a/src/server/game/Warden/WardenWin.cpp#L69-L88))_


Treeston:
Just left it running until all 36 memory and 750 other checks in the current world DB had been requested and passed by an unmodified 3.3.5 client.

Keader:
since you're looking into it... check if you can find any info that helps with https://github.com/TrinityCore/TrinityCore/issues/23035

Treeston:
This just cleans up existing logic, doesn't actually touch any functionality.

I don't necessarily want to spend time reverse engineering Warden.

PR #25247, opened 2020-08-15
Fix for LUA_STR_CHECK

scizzydo:
The check above is what I use to actually get a real result for LUA_STR_CHECK, which when I used variables that would have a string to them it actually fails the check. When checking for items like a function if it exists of course it won't work... In addition, fixing the Unk2 to Unk3 causes people to be able to adjust the address of the check to the address 0x00419210 to make warden perform the FrameScript__Execute to run lua on the client. This seems to be the big request like in https://github.com/TrinityCore/TrinityCore/issues/23035

**Changes proposed:**

-  Fix the real issue with the wrong value passed to the BuildChecksum for the function2 warden check
-  Adjusted the check return for how I was testing on my server, since it's literally checking for a string value I assumed the warden check was to fail if there is an actual string value. Of course it won't pass for things like functions, and what not, but it will return a string value if the item has one.
-  

- [X] 3.3.5
- [ ] master

**Issues addressed:**

https://github.com/TrinityCore/TrinityCore/issues/23035

**Tests performed:**

Tested in game with warden check 139, the str I initially had was ABCDEFG, which when nil or as a function passed the check, but when I did /script ABCDEFG = "word" the LUA_STR_CHECK failed as it was now a string. 

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Warden/WardenWin.cpp
+++ b/src/server/game/Warden/WardenWin.cpp
@@ -105,7 +105,7 @@ void WardenWin::InitializeModule()
     Request.String_library2 = 0;
     Request.Function2 = 0x00419D40;                         // 0x00400000 + 0x00419D40 FrameScript::GetText
     Request.Function2_set = 1;
-    Request.CheckSumm2 = BuildChecksum(&Request.Unk2, 8);
+    Request.CheckSumm2 = BuildChecksum(&Request.Unk3, 8);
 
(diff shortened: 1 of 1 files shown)


Warpten:
* `Unk3` change is fine. (that whole structure is disgusting too, but that's besides the point)
* I don't know about the check itself since that module is half broken in regards to how LUA checks should go but I guess if it works then it works.

Warpten (approved):

PR #25260, opened 2020-08-16
Scripts/Commands: Convert argument parsing of most debug commands to new system

Carbenium:
**Changes proposed:**

-  Scripts/Commands: Convert argument parsing of most debug commands to new system
Commands which accept enums aren't included.

- [x] 3.3.5
- [ ] master

**Tests performed:**

build + the more complicated ones ingame


sql/updates/world/3.3.5/2020_08_xx_xx_world.sql
```diff
@@ -0,0 +1,10 @@
+UPDATE `command` SET `help`='Syntax: .debug transport [1/0]
```
Treeston, line 1: I don't like `[1/0]` here. Use `start/stop`.

sql/updates/world/3.3.5/2020_08_xx_xx_world.sql
```diff
@@ -0,0 +1,10 @@
...
+
+ Allows to stop (0) a transport at its nearest wait point and start movement (1) of a stopped one. Not all transports can be started or stopped.' WHERE `name`='debug transport';
+
+UPDATE `command` SET `help`='Syntax: .debug boundary 1/0 [duration]
```
Treeston, line 5: Same, use optional exact sequence "fill" + duration.

sql/updates/world/3.3.5/2020_08_xx_xx_world.sql
```diff
@@ -0,0 +1,10 @@
...
+Flood fills the targeted unit''s movement boundary and marks the edge of said boundary with debug creatures.
+Specify ''1'' as first parameter to fill the entire area with debug creatures.' WHERE `name`='debug boundary';
+
+UPDATE `command` SET `help`='Syntax: .debug neargraveyard 1/0
```
Treeston, line 9: Same, use optional exact sequence "linked".

sql/updates/world/3.3.5/2020_08_xx_xx_world.sql
Treeston: Also, _what_ transport? current transport?

Carbenium: > Also, _what_ transport? current transport?

Yes, the one you're on. I'll clarify that bit.

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -131,19 +131,9 @@ class debug_commandscript : public CommandScript
         return commandTable;
     }
 
-    static bool HandleDebugPlayCinematicCommand(ChatHandler* handler, char const* args)
+    // cinematicId - ID from CinemaicSequences.dbc
```
Treeston, line 134:
```suggestion
    // cinematicId - ID from CinematicSequences.dbc
```

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -579,13 +514,8 @@ class debug_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleDebugGetItemStateCommand(ChatHandler* handler, char const* args)
+    static bool HandleDebugGetItemStateCommand(ChatHandler* handler, std::string itemState)
```
Treeston, line 517: keep in mind that this only gets the input up to the first whitespace - is this intended?

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -1061,23 +991,25 @@ class debug_commandscript : public CommandScript
...
+        if (seatIdArg)
+        {
+            entry = entryOrSeatId;
+            seatId = *seatIdArg;
+        }
+        else
+        {
+            entryOrSeatId = 0;
+            seatId = entryOrSeatId;
+        }
```
Treeston, line 1013: Is this a change of functionality? As I see it, before it was always `entry seatId`, and now it is `[entry] seatId`?

PS: This entire check looks screwy to me. Consider making `entry` optional instead of what you currently do.

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -1152,29 +1071,15 @@ class debug_commandscript : public CommandScript
...
-            return false;
-
-        uint32 PhaseShift = atoi(args);
-        handler->GetSession()->SendSetPhaseShift(PhaseShift);
+        handler->GetSession()->SendSetPhaseShift(phaseShift);
         return true;
     }
 
-    static bool HandleDebugGetItemValueCommand(ChatHandler* handler, char const* args)
+    static bool HandleDebugGetItemValueCommand(ChatHandler* handler, uint32 guidArg, uint32 index)
```
Treeston, line 1080:
```suggestion
    static bool HandleDebugGetItemValueCommand(ChatHandler* handler, ObjectGuid::LowType guid, uint32 index)
```

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -1220,16 +1113,9 @@ class debug_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleDebugItemExpireCommand(ChatHandler* handler, char const* args)
+    static bool HandleDebugItemExpireCommand(ChatHandler* handler, uint32 guidArg)
```
Treeston, line 1116:
```suggestion
    static bool HandleDebugItemExpireCommand(ChatHandler* handler, ObjectGuid::LowType guid)
```

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -1191,21 +1096,9 @@ class debug_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleDebugSetItemValueCommand(ChatHandler* handler, char const* args)
+    static bool HandleDebugSetItemValueCommand(ChatHandler* handler, uint32 guidArg, uint32 index, uint32 value)
```
Treeston, line 1099:
```suggestion
    static bool HandleDebugSetItemValueCommand(ChatHandler* handler, ObjectGuid::LowType guid, uint32 index, uint32 value)
```

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -1302,18 +1188,8 @@ class debug_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleDebugSetValueCommand(ChatHandler* handler, char const* args)
+    static bool HandleDebugSetValueCommand(ChatHandler* handler, uint32 index, Variant<uint32, float> value)
```
Treeston, line 1191: I actually don't like this. It makes `.debug setvalue 0 10` and `.debug setvalue 0 10.0` do completely different things, and I think this is very much not obvious.

Treeston: Also it reminds me that I want to adjust `Variant` to use `std::variant`... might do that later

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -1582,45 +1411,40 @@ class debug_commandscript : public CommandScript
...
+            transport->EnableMovement(false);
+        }
 
-        handler->PSendSysMessage("Transport %s %s", transport->GetName().c_str(), start ? "started" : "stopped");
+        handler->PSendSysMessage("Transport %s %s", transport->GetName().c_str(), *start ? "started" : "stopped");
         return true;
     }
 
-    static bool HandleDebugLoadCellsCommand(ChatHandler* handler, char const* args)
+    static bool HandleDebugLoadCellsCommand(ChatHandler* handler, int32 mapId)
```
Treeston, line 1440:
```suggestion
    static bool HandleDebugLoadCellsCommand(ChatHandler* handler, uint32 mapId)
```

src/server/scripts/Commands/cs_debug.cpp
Carbenium: I'd say so. The expected possible strings are all without whitespace.

src/server/scripts/Commands/cs_debug.cpp
Carbenium: I would argue that it's a debug command which should be used by knowledgeable people, making it an acceptable use-case. So this should really be an exception.

Additionally the command output makes it quite clear what was done (int/float is explicitly specified in the text).

src/server/scripts/Commands/cs_debug.cpp
Carbenium: Yep, I'm stupid.

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -1900,7 +1683,7 @@ class debug_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleDebugOutOfBounds(ChatHandler* handler, CommandArgs* /*args*/)
+    static bool HandleDebugOutOfBounds(ChatHandler* handler)
```
Treeston, line 1686: I fundamentally disagree with this function existing in our source code.

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -258,33 +220,16 @@ class debug_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleDebugSendSpellFailCommand(ChatHandler* handler, char const* args)
```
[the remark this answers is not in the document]
Carbenium: Good catch. I'll revert this part and do it separately with the other enums.

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -491,21 +436,13 @@ class debug_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleDebugUpdateWorldStateCommand(ChatHandler* handler, char const* args)
+    static bool HandleDebugUpdateWorldStateCommand(ChatHandler* handler, uint32 variable, uint32 value)
```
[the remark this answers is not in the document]
Carbenium: Actually it should be `variableId` since that's how the packet field is named. So rather keep like it is or go for `variableId`.

https://github.com/TrinityCore/TrinityCore/blob/a9a4f05a282fd85e4f22923ec251c92e79581a47/src/server/game/Server/Packets/WorldStatePackets.h#L56

src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -1061,24 +991,12 @@ class debug_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleDebugEnterVehicleCommand(ChatHandler* handler, char const* args)
+    static bool HandleDebugEnterVehicleCommand(ChatHandler* handler, uint32 entry, int8 seatId)
```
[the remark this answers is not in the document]
Carbenium: Yes, you're right...

src/server/scripts/Commands/cs_debug.cpp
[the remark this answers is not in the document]
Treeston: hm, never looked at the packet really

fine by me

src/server/scripts/Commands/cs_debug.cpp
Carbenium: I guess let's see what @jackpoz has to say

Treeston: too late e18f7aa

Treeston (approved):

PR #25276, opened 2020-08-18
Scripts/Commands: Convert argument parsing of deserter commands to new system

Carbenium:
**Changes proposed:**

-  Scripts/Commands: Convert argument parsing of deserter commands to new system

- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/server/scripts/Commands/cs_deserter.cpp
+++ b/src/server/scripts/Commands/cs_deserter.cpp
@@ -76,7 +76,7 @@ public:
     * selected player, with the provided duration in seconds.
     *
     * @param handler The ChatHandler, passed by the system.
-    * @param args The provided duration in seconds.
+    * @param time The provided duration in seconds.
     * @param isInstance provided by the relaying functions, so we don't have
     * to write that much code :)
     *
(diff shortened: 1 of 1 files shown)


src/server/scripts/Commands/cs_deserter.cpp
```diff
@@ -89,26 +89,15 @@ class deserter_commandscript : public CommandScript
     * .deserter bg add 3600 (one hour)
     * @endcode
     */
-    static bool HandleDeserterAdd(ChatHandler* handler, char const* args, bool isInstance)
+    static bool HandleDeserterAdd(ChatHandler* handler, uint32 time, bool isInstance)
```
Treeston, line 92: Might be useful to allow taking std::chrono types to make the unit of time explicit?

Carbenium: Yes, that would be definitely nice. I would like do that in a second step since we have to implement duration string parsing and the consumer APIs have to be converted as well.

Treeston: sure, works for me

Treeston (approved):

PR #25282, opened 2020-08-18
Tweaks to detect PQR for example

scizzydo:
**Changes proposed:**

-  Change LUA_STR_SEARCH to be FrameScript__Execute instead of FrameScript__GetText
-  Modify the checks to send 1 of the lua commands per warden check (otherwise it may be a long time till it's called again)
-  Modify the chat handler to log if the sender is the recipient and the message contains Warden

- [X] 3.3.5
- [ ] master

**Issues addressed:**

https://github.com/TrinityCore/TrinityCore/issues/23035

**Tests performed:**

It builds. The check I used was:
Check ID 139 with the str being:
"if CastSpellByID and not CSBID_H then hooksecurefunc('CastSpellByID', function() if debugstack():find('PQR') then SendAddonMessage("Warden", "PQR Found", "WHISPER", UnitName("player")) end end) CSBID_H = true end"

The reason I decided to do hooksecurefunc and check the debugstack is because CastSpellByID/CastSpellByName should never come from a call outside of the Interface\AddOn call path and wouldn't include PQR in the debugstack

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Handlers/ChatHandler.cpp
+++ b/src/server/game/Handlers/ChatHandler.cpp
@@ -221,6 +221,18 @@ void WorldSession::HandleMessagechatOpcode(WorldPacket& recvData)
     if (msg.size() > 255)
         return;
 
+    // Handle the warden SendAddonMessage here
+    if (lang == LANG_ADDON)
+    {
+        if (to == sender->GetName())
+        {
+            if (msg.find("Warden") != std::string::npos)
+            {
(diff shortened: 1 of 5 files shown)


scizzydo:
https://imgur.com/a/Cii15N1

There's the screen shots for example with the log showing the message of PQR detected, the logging settings, etc

Treeston:
Thanks!

PR #25285, opened 2020-08-18
Core/ChatCommands: Check whether a passed numeric enum value is valid

Carbenium:
**Changes proposed:**

-  Core/ChatCommands: Check whether a passed numeric enum value is valid

- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/common/Utilities/SmartEnum.h
+++ b/src/common/Utilities/SmartEnum.h
@@ -59,6 +59,22 @@ class EnumUtils
         template <typename Enum>
         static uint32 ToIndex(Enum value) { return Trinity::Impl::EnumUtils<Enum>::ToIndex(value);}
 
+        template<typename Enum>
+        static bool IsValid(Enum value)
+        {
+            try
+            {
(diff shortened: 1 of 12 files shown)


src/common/Utilities/SmartEnum.h
```diff
@@ -23,7 +23,9 @@
 
 struct EnumText
 {
-    EnumText(char const* c, char const* t, char const* d) : Constant(c), Title(t), Description(d) {}
+    EnumText(uint32 i, char const* c, char const* t, char const* d) : Index(i), Constant(c), Title(t), Description(d) { }
```
Carbenium, line 26: I'm not quite happy with `EnumText` and `ToString` anymore if we do it like this.

@Treeston suggestions?

Treeston: no need to add it to enumtext

literally just make it the opposite of what FromIndex already is

Treeston (approved):

PR #25302, opened 2020-08-21
Core/ChatCommands: Add std::wstring argument type

Carbenium:
**Changes proposed:**

-  Core/ChatCommands: Add std::wstring argument type

- [x] 3.3.5
- [ ] master


Treeston (approved):
Code is fine from my end, figure out with @Shauren if this is something we want.

PR #25318, opened 2020-08-23
Core/SAI: Implement SMART_TARGET_ACTION_INVOKER in SMART_EVENT_FRIENDLY_HEALTH_PCT

### LINKED ISSUE #20536: Core/SAI: SMART_EVENT_FRIENDLY_HEALTH_PCT is broken

jackpoz:
**Changes proposed:**

-  Core/SAI: Implement SMART_TARGET_ACTION_INVOKER in SMART_EVENT_FRIENDLY_HEALTH_PCT
- Added a 5th parameter to SMART_EVENT_FRIENDLY_HEALTH_PCT "range" that is used when target is SMART_TARGET_ACTION_INVOKER to look for low hp % friendly creatures. **SAI scripts must set this to a value != 0**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #20536 , ref https://github.com/TrinityCore/TrinityCore/pull/20694 , ref https://github.com/TrinityCore/TrinityCore/pull/23377

**Tests performed:**

Still testing

- [ ] The existing SMART_EVENT_FRIENDLY_HEALTH_PCT with SMART_TARGET_* not SMART_TARGET_ACTION_INVOKER should be rechecked and maybe even updated as ideally SMART_EVENT_FRIENDLY_HEALTH_PCT should always just call the same function to get the lowest hp % target in range and then GetTarget() for the action could be whatever, just like how SMART_EVENT_FRIENDLY_HEALTH works already

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -3333,7 +3333,7 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
             if (!me || !me->IsEngaged())
                 return;
 
-            ObjectVector targets;
+            Unit* unitTarget = nullptr;
             switch (e.GetTargetType())
             {
                 case SMART_TARGET_CREATURE_RANGE:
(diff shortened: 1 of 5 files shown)


offl:
Let's talk about additional params to target player too in the same event type in another issue, maybe I'll find a way to implement it more correctly

offl:
Works perfectly, now we can update all actions to use target INVOKER and migrate event HEALTH to HEALTH_PTC

PR #25321, opened 2020-08-23
Core/Config: implement custom config

Winfidonarleyan:
**Changes proposed:**
- Load config options from DB.

**Target branch(es):**
- [x] 3.3.5
- [ ] master

**Tests performed:**
- Build and in-game

- [ ]

### About system
1. This system well suited for custom scripts (no need change `worldserver.conf`)
2. Сan replaces ~90% options `worldserver.conf`. (Example full replace this section)
https://github.com/TrinityCore/TrinityCore/blob/27b1930d97d541c241655d8b820cbaebe6eec44e/src/server/game/World/World.cpp#L470
And delete config in `World` class.
3. This system is completely independent, it does not affect default config in `World` class. That is, there will be no conflicts


Treeston:
Can you elaborate on how this would behave? Would it replace `authserver.conf`? `worldserver.conf`? Both?

What happens if the setting in `.conf` conflicts with the setting in the DB?

`world` is designed to be disposable. Why is this data stored in `world`, instead of `character`?

Winfidonarleyan:
@Treeston 
1. At this stage, it will be best only for `worldserver.conf`
2. There can be no conflict, because one thing is used (`sWorld->get***Config()` or `sGameConfig->Get***Config())
3. To be honest, I did not understand the question, I think that the config should be stored in the world DB

Jildor:
I prefer keep config in *.conf files

src/server/game/Config/GameConfig.cpp
```diff
@@ -0,0 +1,210 @@
...
+    _floatOptions.clear();
+    _stringOptions.clear();
+}
+
+void GameConfig::AddBoolOption(std::string const& optionName, bool value /*= false*/)
+{
+    auto const& itr = _boolOptions.find(optionName);
+    if (itr != _boolOptions.end())
+    {
+        TC_LOG_FATAL("server.loading", "> Bool option (%s) is exist!", optionName.c_str());
```
jackpoz, line 43: >Bool option (%s) is exist!

I think there is typo somewhere in this sentence, should it be "Bool option (%s) exists already!" or so ?

src/server/game/Config/GameConfig.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+void GameConfig::AddOption(std::string const& optionName, GameConfigType type, std::string const& value, std::string const& defaultValue)
+{
+    switch (type)
+    {
+    case GameConfigType::GAME_CONFIG_TYPE_BOOL:
+        AddBoolOption(optionName, value.empty() ? StringToBool(defaultValue) : StringToBool(value));
+        break;
+    case GameConfigType::GAME_CONFIG_TYPE_INT:
+        AddIntOption(optionName, value.empty() ? std::stoi(defaultValue) : std::stoi(value));
```
jackpoz, line 94: please check if some recent changes in 3.3.5 branch added new features to transform a string to a int/float

src/server/game/Config/GameConfig.cpp
```diff
@@ -0,0 +1,210 @@
...
+        break;
+    case GameConfigType::GAME_CONFIG_TYPE_FLOAT:
+        AddFloatOption(optionName, value.empty() ? std::stof(defaultValue) : std::stof(value));
+        break;
+    case GameConfigType::GAME_CONFIG_TYPE_STRING:
+        AddStringOption(optionName, value.empty() ? defaultValue : value);
+        break;
+    default:
+        ABORT();
```
jackpoz, line 103: does this just crash worldserver with not useful message about why it did that ?

src/server/game/Config/GameConfig.h
```diff
@@ -0,0 +1,66 @@
...
+
+class TC_GAME_API GameConfig
+{
+public:
+    static GameConfig* instance();
+
+    void Load();
+
+    // Get config options
+    bool GetBoolConfig(std::string const& optionName);
```
jackpoz, line 42: maybe these functions could accept a default value like the sWorld ones ?

jackpoz:
I left some comments. Please take a look at them.

I wonder also if we should change the name of the table to something like custom_* to make clear TC is not using this by default

src/server/game/Config/GameConfig.cpp
Winfidonarleyan: Maybe, my english is bad :D

src/server/game/Config/GameConfig.cpp
Winfidonarleyan: Yes yes, okay

src/server/game/Config/GameConfig.cpp
Winfidonarleyan: Hm, i agree

Treeston:
See #25329, we will not be adopting the PR at this time.

jackpoz:
I'm still quite interested into merging this as long as TC doesn't actually move any config in there but only custom changes choose to use it.

Treeston:
In that case, my previous comment still stands

> `world` is designed to be disposable. Why is this data stored in `world`, instead of `character`?

jackpoz:
It seems there was a time when one could create his own .conf file and load it from C++ together. It might be worth to see if that could be another solution too.

src/server/game/Config/GameConfig.h
Winfidonarleyan: I'm not sure this is necessary because the system works as follows, if you don't set a custom value, the function will return the default value

src/server/game/Config/GameConfig.cpp
Winfidonarleyan: Hm ho to use `Trinity::StringTo` for `float` or this type doesn't support?

Treeston: float support soon, give me a day or two, i wanted to merge #25323

Winfidonarleyan: Very good, i wait

Treeston: #25364

sql/updates/world/3.3.5/9999_99_99_99_custom_config.sql
```diff
@@ -0,0 +1,20 @@
...
+  PRIMARY KEY (`OptionName`) USING BTREE
+) ENGINE = InnoDB CHARACTER SET = utf8 COLLATE = utf8_general_ci ROW_FORMAT = Dynamic;
+
+-- ----------------------------
+-- Records of custom_config
+-- ----------------------------
+INSERT INTO `custom_config` VALUES ('TestOption.Bool', 'bool', '1', NULL, 'Test option bool');
+INSERT INTO `custom_config` VALUES ('TestOption.Float', 'float', '1.5232', NULL, 'Test option float');
+INSERT INTO `custom_config` VALUES ('TestOption.Int', 'int', '100', NULL, 'Test option int');
+INSERT INTO `custom_config` VALUES ('TestOption.String', 'string', 'test string', NULL, 'Test option string');
```
illfated, line 20: As long as you are inserting rows into the same table, the SQL only needs 1 insert command:
```suggestion
INSERT INTO `custom_config` (`OptionName`, `Type`, `DefaultValue`, `CustomValue`, `Description`) VALUES
('TestOption.Bool', 'bool', '1', NULL, 'Test option bool'),
('TestOption.Float', 'float', '1.5232', NULL, 'Test option float'),
('TestOption.Int', 'int', '100', NULL, 'Test option int'),
('TestOption.String', 'string', 'test string', NULL, 'Test option string');
```
TC SQL files normally use a `DELETE FROM` line before the insert. I don't remember if it can be omitted for new tables.

jackpoz:
> It seems there was a time when one could create his own .conf file and load it from C++ together. It might be worth to see if that could be another solution too.

https://github.com/TrinityCore/TrinityCore/blob/28353e920b531b7dd221ac926049f5a68b75e7ec/src/common/Configuration/Config.h#L35 is the function that scripts can call on startup to load additional config files other than worldserver.conf

Winfidonarleyan:
Maybe no need this pr? Hm

PR #25326, opened 2020-08-24
Core/Totem: Immune to all positive spells.

HoustonCore:
**Changes proposed:**

- Immune to all positive spells, except of stoneclaw totem absorb and sentry totem bind sight
- Totems positive spells have unit_caster target

- [√ ] 3.3.5

**Tests performed:**

**Tests:**
[Look video](https://youtu.be/qw6ySrBLgho)

### DIFF
--- a/src/server/game/Entities/Totem/Totem.cpp
+++ b/src/server/game/Entities/Totem/Totem.cpp
@@ -159,8 +159,8 @@ bool Totem::IsImmunedToSpellEffect(SpellInfo const* spellInfo, uint32 index, Wor
 {
     // immune to all positive spells, except of stoneclaw totem absorb and sentry totem bind sight
     // totems positive spells have unit_caster target
-    if (spellInfo->Effects[index].Effect != SPELL_EFFECT_DUMMY && 
-        spellInfo->Effects[index].Effect != SPELL_EFFECT_SCRIPT_EFFECT && 
+    if (spellInfo->Effects[index].Effect != SPELL_EFFECT_DUMMY &&
+        spellInfo->Effects[index].Effect != SPELL_EFFECT_SCRIPT_EFFECT &&
         spellInfo->IsPositive() && spellInfo->Effects[index].TargetA.GetTarget() != TARGET_UNIT_CASTER &&
         spellInfo->Effects[index].TargetA.GetCheckType() != TARGET_CHECK_ENTRY && spellInfo->Id != SENTRY_STONECLAW_SPELLID && spellInfo->Id != SENTRY_BIND_SIGHT_SPELLID)
         return true;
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Totem/Totem.cpp
```diff
@@ -157,9 +157,13 @@ void Totem::UnSummon(uint32 msTime)
 
 bool Totem::IsImmunedToSpellEffect(SpellInfo const* spellInfo, uint32 index, WorldObject const* caster) const
 {
-    /// @todo possibly all negative auras immune?
-    if (GetEntry() == 5925)
-        return false;
+    // immune to all positive spells, except of stoneclaw totem absorb and sentry totem bind sight
+	// totems positive spells have unit_caster target
```
Faq, line 161:
```suggestion
    // totems positive spells have unit_caster target
```

src/server/game/Entities/Totem/Totem.cpp
```diff
@@ -157,9 +157,13 @@ void Totem::UnSummon(uint32 msTime)
...
+    if (spellInfo->Effects[index].Effect != SPELL_EFFECT_DUMMY && 
+        spellInfo->Effects[index].Effect != SPELL_EFFECT_SCRIPT_EFFECT && 
+        spellInfo->IsPositive() && spellInfo->Effects[index].TargetA.GetTarget() != TARGET_UNIT_CASTER &&
+        spellInfo->Effects[index].TargetA.GetCheckType() != TARGET_CHECK_ENTRY && spellInfo->Id != 55277 && spellInfo->Id != 6277)
```
Faq, line 165: avoid using magic numbers, like 55277, 6277

src/server/game/Entities/Totem/Totem.cpp
MokiGitHub: Done

src/server/game/Entities/Totem/Totem.cpp
MokiGitHub: > avoid using magic numbers, like 55277, 6277

Do I declare them in Totem.h? I'm just learning)

Jildor:
any proof about that totems should be immune?

MokiGitHub:
> any proof about that totems should be immune?

Edit: Try to google "buffing (or healing) shaman totem wow" there are zero instances where this occured on retail, you will not find anything relevant. If you ever engaged a shaman who used Fire Elemental Totem you would always kill the totem rather than the elemental because the totem couldn't be healed while the elemental could. If this was a thing you could just put Earth Shield on the totem to protect it from being killed.
Source: my 350 days /played as a shaman in TBC/Wotlk/Cata

Yes, and I know some servers have issues with Chain Heal jumping to the totems. Couldn't reproduce it here but it's something to keep in mind. Chain Heal and Prayer of Mending should ignore them as targets.

P.S. This commentnot my.

src/server/game/Entities/Totem/Totem.cpp
MokiGitHub: Done, check pls.

MokiGitHub:
Can someone test it?

Shauren:
Yeah totems are immune to being buffed and healing as well, this is still the case even in bfa

jackpoz:
are there any quest spells that might be affected by this change ?

MokiGitHub:
> are there any quest spells that might be affected by this change ?

I think no.

jackpoz:
Thanks for the PR :)

PR #25327, opened 2020-08-24
Tokenizer -> Trinity::Tokenize

Treeston:
100% less `new char[]`, because this isn't 2005 anymore

### DIFF
--- a/src/server/game/Entities/Item/Item.cpp
+++ b/src/server/game/Entities/Item/Item.cpp
@@ -31,6 +31,7 @@
 #include "ScriptMgr.h"
 #include "SpellInfo.h"
 #include "SpellMgr.h"
+#include "StringConvert.h"
 #include "Player.h"
 #include "TradeData.h"
 #include "UpdateData.h"
@@ -421,7 +422,10 @@ bool Item::LoadFromDB(ObjectGuid::LowType guid, ObjectGuid owner_guid, Field* fi
 
     ItemTemplate const* proto = GetTemplate();
     if (!proto)
+    {
+        TC_LOG_ERROR("entities.item", "Invalid entry %u for item %s. Refusing to load.", GetEntry(), GetGUID().ToString().c_str());
         return false;
+    }
 
     // set owner (not if item is only loaded for gbank/auction/mail
     if (owner_guid)
@@ -441,10 +445,17 @@ bool Item::LoadFromDB(ObjectGuid::LowType guid, ObjectGuid owner_guid, Field* fi
         need_save = true;
     }
 
-    Tokenizer tokens(fields[4].GetString(), ' ', MAX_ITEM_PROTO_SPELLS);
+    std::vector<std::string_view> tokens = Trinity::Tokenize(fields[4].GetCString(), ' ', false);
     if (tokens.size() == MAX_ITEM_PROTO_SPELLS)
+    {
         for (uint8 i = 0; i < MAX_ITEM_PROTO_SPELLS; ++i)
-            SetSpellCharges(i, atoi(tokens[i]));
+        {
+            if (Optional<int32> charges = Trinity::StringTo<int32>(tokens[i]))
+                SetSpellCharges(i, *charges);
+            else
+                TC_LOG_ERROR("entities.item", "Invalid charge info '%s' for item %s, charge data not loaded.", std::string(tokens[i]).c_str(), GetGUID().ToString().c_str());
+        }
+    }
 
     SetUInt32Value(ITEM_FIELD_FLAGS, fields[5].GetUInt32());
     // Remove bind flag for items vs NO_BIND set
@@ -454,7 +465,11 @@ bool Item::LoadFromDB(ObjectGuid::LowType guid, ObjectGuid owner_guid, Field* fi
         need_save = true;
     }
 
-    _LoadIntoDataField(fields[6].GetString(), ITEM_FIELD_ENCHANTMENT_1_1, MAX_ENCHANTMENT_SLOT * MAX_ENCHANTMENT_OFFSET);
+    if (!_LoadIntoDataField(fields[6].GetString(), ITEM_FIELD_ENCHANTMENT_1_1, MAX_ENCHANTMENT_SLOT * MAX_ENCHANTMENT_OFFSET))
+    {
+        TC_LOG_ERROR("entities.item", "Invalid enchantment data '%s' for item %s, refusing to load.", fields[6].GetCString(), GetGUID().ToString().c_str());
+        return false;
+    }
     SetInt32Value(ITEM_FIELD_RANDOM_PROPERTIES_ID, fields[7].GetInt16());
     // recalculate suffix factor
     if (GetItemRandomPropertyId() < 0)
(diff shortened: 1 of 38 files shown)


Warpten:
Why did you have to rewrite so much, just make the iterator element type into std::string_view

Treeston:
All the callers hand it off to `atoi` and similar shit, none of which play nice with `string_view`.

Warpten:
But you can just call .data() and all is fine, especially since these views **are** null-terminated (tokenizer copies and replaces delimiter with \0)

Treeston:
...and I do not want `Tokenizer` to copy if it doesn't have to, and I _definitely_ do not want to start a trend of assuming `string_view`s are null-terminated.

Warpten:
And that's OK too, because atoi and friends don't care if there's non alnum or whatever characters after the number

Treeston:
I'm not going to assume _anything_ about the data outside a `string_view`. That includes not assuming that there is a `\0` immediately after it, but _also_ includes not assuming that there is no `0`-`9` immediately after it.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -1540,27 +1541,40 @@ bool Player::BuildEnumData(PreparedQueryResult result, WorldPacket* data)
...
         }
 
+        ItemTemplate const* proto = nullptr;
+        if (Optional<uint32> itemId = Trinity::StringTo<uint32>(equipment[visualBase]))
+            proto = sObjectMgr->GetItemTemplate(*itemId);
+
+        if (!proto)
+        {
+            TC_LOG_ERROR("entities.player.loading", "Player %u has invalid equipment '%s' in `equipmentcache`. Load aborted.", guid, std::string(equipment[visualBase]).c_str());
+            return false;
```
Shauren, line 1563: Do not skip building character enum for a silly reason like this, just send empty slot like it used to

src/server/game/Entities/Player/PlayerTaxi.cpp
```diff
@@ -61,16 +62,21 @@ void PlayerTaxi::InitTaxiNodesForLevel(uint32 race, uint32 chrClass, uint8 level
         SetTaximaskNode(213);                               //Shattered Sun Staging Area
 }
 
-void PlayerTaxi::LoadTaxiMask(std::string const& data)
+bool PlayerTaxi::LoadTaxiMask(std::string const& data)
```
Shauren, line 65: Dont fail. 0 on invalid tokens

src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -2068,18 +2069,27 @@ void WorldSession::HandleCharFactionOrRaceChangeCallback(std::shared_ptr<Charact
...
-                    knownTitles[index] = atoul(tokens[index]);
+                for (uint32 index = 0; index < knownTitles.size(); ++index)
+                {
+                    if (Optional<uint32> thisMask = Trinity::StringTo<uint32>(tokens[index]))
+                        knownTitles[index] = *thisMask;
+                    else
+                    {
+                        TC_LOG_ERROR("entities.player", "%s has invalid title data '%s' at index %u - faction change aborted", GetPlayerInfo().c_str(), std::string(tokens[index]).c_str(), index);
+                        SendCharFactionChange(CHAR_CREATE_ERROR, factionChangeInfo.get());
+                        return;
```
Shauren, line 2090: same here, don't fail the entire faction change process if old data cannot be converted, just log and skip it

Treeston: I didn't want to proceed since it has the potential for irreversible data loss.

src/server/shared/Packets/ByteBuffer.h
```diff
@@ -203,20 +203,22 @@ class TC_SHARED_API ByteBuffer
...
-            if (size_t len = (str ? strlen(str) : 0))
-                append((uint8 const*)str, len);
-            append((uint8)0);
-            return *this;
+            return operator<<(std::string_view(str));
+        }
+
+        ByteBuffer &operator<<(char const* str)
+        {
+            return operator<<(std::string_view(str));
```
Shauren, line 221: old code handled nullptr safely, this does not

Treeston: it did not - old code called `append(nullptr, 0)` which would assert out

src/server/game/Entities/Player/PlayerTaxi.cpp
Treeston: Same as faction change, this has the potential for irreversible data loss once the character gets saved back to DB.

src/server/game/Entities/Player/Player.cpp
Treeston: This one I am a bit more conflicted on, but generally I feel that if character data is invalid, I'd rather fail safely (refuse to load) rather than fail deadly (potentially lose data somewhere down the line).

Shauren: Except deleting invalid items is the intended behavior here

src/server/game/Entities/Item/Item.cpp
```diff
@@ -454,7 +465,11 @@ bool Item::LoadFromDB(ObjectGuid::LowType guid, ObjectGuid owner_guid, Field* fi
         need_save = true;
     }
 
-    _LoadIntoDataField(fields[6].GetString(), ITEM_FIELD_ENCHANTMENT_1_1, MAX_ENCHANTMENT_SLOT * MAX_ENCHANTMENT_OFFSET);
+    if (!_LoadIntoDataField(fields[6].GetString(), ITEM_FIELD_ENCHANTMENT_1_1, MAX_ENCHANTMENT_SLOT * MAX_ENCHANTMENT_OFFSET))
```
Shauren, line 468: Do not fail loading item when enchantments are invalid. Clear enchantments and trigger item save.

Returning false from Item::LoadFromDB permanently deletes the entire item by design.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -1540,27 +1541,40 @@ bool Player::BuildEnumData(PreparedQueryResult result, WorldPacket* data)
...
+        }
+
         SpellItemEnchantmentEntry const* enchant = nullptr;
 
-        uint32 enchants = GetUInt32ValueFromArray(equipment, visualBase + 1);
+        Optional<uint32> enchants = Trinity::StringTo<uint32>(equipment[visualBase + 1]);
+        if (!enchants)
+        {
+            TC_LOG_ERROR("entities.player.loading", "Player %u has invalid enchantment info '%s' in `equipmentcache`. Load aborted.", guid, std::string(equipment[visualBase + 1]).c_str());
+            return false;
```
Shauren, line 1572: same here, let it build packet with a 0 and clear enchantments on loadfromdb

jackpoz:
Would it be possible to add a debug log with a new category to get a message about all failed Trinity::StringTo() conversions ? The old atoi/atoll/atoul kept parsing as long as valid characters were found so we might have some database data that has to be fixed (i.e. "123," instead of "123")

Treeston:
Call sites are forced to explicitly deal with failures (`Optional<T>` does not implicitly convert to `T`).

All current call sites log failures to some extent.

PR #25329, opened 2020-08-24
Implement WorldConfigMgr: now all configs are placed in DataBase

kvipka:
Move configs from overloaded worldserver.conf in database. Code based on work of @Winfidonarleyan

**Changes proposed:**

-  fix missed configs from World.h
-  readable code in World.cpp
-  implement correct handle for configs which have dependancy

- [√] 3.3.5
- [ ] master

**Tests performed:**


Treeston:
Does this supersede #25321?

Treeston:
`world` is designed to be discardable. `drop database world; create database world;` should not lose your settings.
Move them to `characters`.

Treeston:
I see no reason to have defaults in DB updates. Have worldserver push them to the DB on startup if a requested key is missing.

Treeston:
The `IDInTypeGroup` is a disaster waiting to happen. Just use string identifiers.

kvipka:
> The `IDInTypeGroup` column is a disaster waiting to happen. Just use string identifiers in DB, and load them into the enum-indexed arrays from worldserver as usual.

I previously have thought to do just 5 tables for each type.

kvipka:
> `world` is designed to be discardable. `drop database world; create database world;` should not lose your settings.
> Move them to `characters`.

Can't understand. What you mean under "should not lose your settings."
You Set in world_config in field custom_value your parameters, like this:


And lately just make sql-query for nessesary your fields

Ovahlord:
But why even do all of it? It removes any kind of properly readable documentation and makes it even more vulnerable to hack attacks by exposing it to database injections.

Viste:
> makes it even more vulnerable to hack attacks by exposing it to database injections

all main params still on worldserver.conf as i see. also in real world only newcomers do the:
`GRANT ALL PRIVILEGES ON dbname.* TO 'username'@'%';` 
You can always allow access to the web part, only to the tables you need for certain actions, and keep access to this table for trinity@localhost(just example) and keep DB protected, never hear about database injections through wow client

Jildor:
I'm with @Ovahlord
The conf file is better for readability and keep order grouping confs

kvipka:
As i understand, Treeston mean, that configs should be easy to install.
And if someone will drop world database - configs should working.

For this in db we have two values (default column and custom, where you set nessesary values). Just save queries with your settings. For example: 


Just save your queries and implement it after *drop db - set db*

MaxtorCoder:
> 
> 
> As i understand, Treeston mean, that configs should be easy to install.
> And if someone will drop world database - configs should working.
> 
> For this in db we have two values (default column and custom, where you set nessesary values). Just save queries with your settings. For example:
> 
> 
> Just save your queries and implement it after _drop db - set db_
> something like this :
> 
> ```
> UPDATE `world_config` SET `CustomValue` = 'BLA BLA CAR'WHERE `OptionName` = 'Motd';
> UPDATE `world_config` SET `CustomValue` = '1' WHERE `OptionName` = 'PlayerSave.Stats.MinLevel';
> ```

That's just useless in my opinion, that would be even more work to reset the database for your core. 😄 Configs should just stay in the .conf files and not in the database.

Viste:
> Configs should just stay in the .conf files and not in the database

When you put in db config ptions like rates and etc and all other stay on file config why not? its normal case

MaxtorCoder:
> 
> 
> > Configs should just stay in the .conf files and not in the database
> 
> When you put in db-config options like rates and etc and all other stay on file-config why not? its normal case

But why would you want that tho, 1 it takes longer to start the WorldServer since it has to fetch all the configs + actually assign them. I really don't see a use case to database configs.

kvipka:
yep, longer... =)

MaxtorCoder:
> 
> 
> yep, longer... =)
> 

Still doesn't change the fact that I find it a useless feature to have.

Shauren:
No, configs should be in file

jackpoz:
Thanks for the PR.

We currently have no plans on changing the way users access the configs and before merging this PR we should agree on doing that. As usual, when there is a disagreement on a change, no-change is the best option.

Moving some custom configs to DB "might" be ok (but then that ends up having configs spread around), having all configs in the DB doesn't sound ok. On top of that, the transition is just calling for issues (what happens to the existing modified values in worldserver.conf of existing servers ? are they migrated ? do they take priority ? what if one changes the value in both ?).

If enough people show interest in this, it could be moved to https://github.com/TrinityCore/TrinityCoreCustomChanges/ (or we could just add a link to the wiki).

PR #25335, opened 2020-08-25
StringConvert.h

Treeston:
Centralize String-to-type conversion in a single header.

More importantly, add a workaround for a compiler bug in clang 7 (which we still support, thanks Debian 10).

clang will fail to link if `std::from_chars` is used for a 64-bit parameter. This would break the build if any chat commands used 64-bit arguments, for example. Awkward.

cmake now tests for existence of the bug, and we work around it if found (by falling back to `std::sto(u)ll`).

### DIFF
--- a/src/common/Utilities/StringConvert.h
+++ b/src/common/Utilities/StringConvert.h
@@ -31,7 +31,7 @@ namespace Trinity::Impl::StringConvertImpl
 {
     template <typename T, typename = void> struct For
     {
-        static_assert(!std::is_same_v<T,T>, "Unsupported type used for ToString or StringTo");
+        static_assert(Trinity::dependant_false_v<T>, "Unsupported type used for ToString or StringTo");
         /*
         static Optional<T> FromString(std::string_view str, ...);
         static std::string ToString(T&& val, ...);
@@ -76,7 +76,7 @@ namespace Trinity::Impl::StringConvertImpl
 
         static std::string ToString(T val)
         {
-            std::string buf(20,0); /* 2^64 is 20 decimal characters, -(2^63) is 20 including the sign */
+            std::string buf(20,'\0'); /* 2^64 is 20 decimal characters, -(2^63) is 20 including the sign */
             char* const start = buf.data();
             char* const end = (start + buf.length());
             std::to_chars_result const res = std::to_chars(start, end, val);
--- a/src/common/Utilities/StringConvert.h
+++ b/src/common/Utilities/StringConvert.h
@@ -92,6 +92,13 @@ namespace Trinity::Impl::StringConvertImpl
     };
 
 #ifdef TRINITY_NEED_CHARCONV_WORKAROUND
+    /*
+        If this is defined, std::from_chars will cause linkage errors for 64-bit types.
+        (This is a bug in clang-7.)
+
+        If the clang requirement is bumped to >= clang-8, remove this ifdef block and its
+        associated check in cmake/compiler/clang/settings.cmake
+    */
     template <>
     struct For<uint64, void>
     {
--- a/src/common/Utilities/StringConvert.h
+++ b/src/common/Utilities/StringConvert.h
@@ -31,7 +31,7 @@ namespace Trinity::Impl::StringConvertImpl
 {
     template <typename T, typename = void> struct For
     {
-        static_assert("Unsupported type used for ToString or StringTo");
+        static_assert(!std::is_same_v<T,T>, "Unsupported type used for ToString or StringTo");
         /*
         static Optional<T> FromString(std::string_view str, ...);
         static std::string ToString(T&& val, ...);
@@ -56,11 +56,6 @@ namespace Trinity::Impl::StringConvertImpl
                     base = 2;
                     str.remove_prefix(2);
                 }
-                else if (str.substr(0, 1) == "0")
-                {
-                    base = 8;
-                    str.remove_prefix(1);
-                }
                 else
                     base = 10;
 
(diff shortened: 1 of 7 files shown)


src/common/Utilities/StringConvert.h
```diff
@@ -0,0 +1,178 @@
...
+#include <charconv>
+#include <string>
+#include <string_view>
+#include <type_traits>
+
+namespace Trinity::Impl::StringConvertImpl
+{
+    template <typename T, typename = void> struct For
+    {
+        static_assert("Unsupported type used for ToString or StringTo");
```
Warpten, line 34: This becomes static_assert(true)

src/common/Utilities/StringConvert.h
```diff
@@ -0,0 +1,178 @@
...
+
+        static std::string ToString(T val)
+        {
+            std::string buf(20,0); /* largest unsigned 64-bit value is 20 decimal characters, largest signed 64-bit value is 19 */
+            char* const start = buf.data();
+            char* const end = (start + buf.length());
+            std::to_chars_result const res = std::to_chars(start, end, val);
+            ASSERT(res.ec == std::errc());
+            buf.resize(res.ptr - start);
+            return buf;
```
Warpten, line 90: Have you considered `fmt::format("{}", val)` here

src/common/Utilities/StringConvert.h
```diff
@@ -0,0 +1,178 @@
...
+        {
+            return (val ? "1" : "0");
+        }
+    };
+}
+
+namespace Trinity
+{
+    template <typename Result, typename... Params>
+    Optional<Result> StringTo(std::string_view str, Params&&... params) { return Trinity::Impl::StringConvertImpl::For<Result>::FromString(str, std::forward<Params>(params)...); }
```
Warpten, line 172: line returns please, same L175

src/common/Utilities/StringConvert.h
Shauren: static_assert(!std::is_same_v<T, T>, "bla");

src/common/Utilities/StringConvert.h
```diff
@@ -0,0 +1,178 @@
...
+            std::from_chars_result const res = std::from_chars(start, end, val, base);
+            if (res.ptr == end)
+                return val;
+            else
+                return std::nullopt;
+        }
+
+        static std::string ToString(T val)
+        {
+            std::string buf(20,0); /* largest unsigned 64-bit value is 20 decimal characters, largest signed 64-bit value is 19 */
```
Shauren, line 84:
```suggestion
            std::string buf(20,'\0'); /* largest unsigned 64-bit value is 20 decimal characters, largest signed 64-bit value is 19 */
```

Also, comment isnt right because "longest" 64 bit signed value is also 20, dont forget the '-' character

src/common/Utilities/StringConvert.h
```diff
@@ -0,0 +1,178 @@
...
+                {
+                    base = 16;
+                    str.remove_prefix(2);
+                }
+                else if (StringEqualI(str.substr(0, 2), "0b"))
+                {
+                    base = 2;
+                    str.remove_prefix(2);
+                }
+                else if (str.substr(0, 1) == "0")
```
Shauren, line 59: I'm not sure we should support base 8 with just a "0" prefix, I'd rather not support it at all (or at least make the prefix "0o" if you have to have it)

src/common/Utilities/StringConvert.h
Warpten: consider adding

```cpp
template <typename T>
struct dependant_false { constexpr static const bool value = false; };

template <typename T>
constexpr static const bool dependant_false_v = dependant_false<T>::value;
```

to advstd?

src/common/Utilities/StringConvert.h
Treeston: meant 19 without sign

src/common/Utilities/StringConvert.h
Treeston: This behavior matches `std::sto(u)ll`, if we do not want this that's fine too.

src/common/Utilities/StringConvert.h
Treeston: If we ever choose to change our fmtlib uses to `{}` specifiers, I'd either adjust this to use fmtlib for everything, or use fmtlib to use this for everything (no need to duplicate code).

Right now, we don't, though.

src/common/Utilities/StringConvert.h
Treeston: is that actually in c++20 @Warpten?

Treeston: and yeah i brainfarted and forgot the is_same_v

src/common/Utilities/StringConvert.h
Treeston: in my defense that line was a lot simpler when i initially added it...

it kind of just kept growing

src/common/Utilities/StringConvert.h
Warpten: Nah it's not, but it's so convenient we might as well. Or add another header, at least the intent is clearer that way

src/common/Utilities/StringConvert.h
Shauren: btw you missed the `0` to `'\0'` change

src/common/Utilities/StringConvert.h
Warpten: Yeah templates do that

PR #25336, opened 2020-08-26
Core/ObjectMgr: Solve global storage issue

illunix:
**Changes proposed:**
 
* Add a condition for normalizePlayerName to not search for existing names with a blank space " " in the end of the string.
 
**Target branch(es):**
 
* [x]   3.3.5
 
**Tests performed:**
 
* Builds, ingame

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -142,6 +142,9 @@ bool normalizePlayerName(std::string& name)
     if (name.empty())
         return false;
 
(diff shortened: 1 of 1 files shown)


Shauren:
Before making a change like this, what does it fix? Is there an issue?

FireEmerald:
If it was possible before this change to create a character with following spaces behind the name, then this change makes sense 🤔

Shauren:
No, it was never possible, it is checked at character creation in `ObjectMgr::CheckPlayerName`

illunix:
Okey the nvm

Issue #24218, opened 2020-03-01
Quest "Digging for Prayer Beads" problem


aleigood:
Description:
After the Honor Hold was attacked by Infernal Rain, NPC did not return to its original location, and the Digging for Prayer Beads quest could not be completed, The dog Fei Fei took the wrong path.

Steps to reproduce the problem:
1. .character level 60
2. .go xyz -684.60302 2626.43994 89.15503 530
3. .quest add 10916
4. .additem 31799
5. wait for Infernal Rain
6. complete the quest

Branch(es):
3.3.5

TC rev. hash/commit:
94186a5e79d5d3fe29ac80245992af6365999a93

Operating system:
CentOS

dr-j:
I dont understand what issue is here other than your instructions to reproduce been wrong as there is no item 10916 its quest not item

as for dog been in wrong place this could be due to previous run of script and grid has been unloaded by there been no player in area before script completed  and thus when next player came was in wrong position

And I have no idea what you are exactually going on about when you say infernal rain

dr-j:
It is also possible there maybe more than one path fei fei may use

aleigood:
> It is also possible there maybe more than one path fei fei may use as think I was last person to do any work on this and that was from sniff but there was a prior implementation of this in tc before

Sorry, I've updated the steps

![image](https://user-images.githubusercontent.com/3367999/75626257-03ff3100-5c01-11ea-84df-34db49dd2352.png)

Killyana:
Related to https://github.com/TrinityCore/TrinityCore/issues/21106

Issue #22362, opened 2018-08-31
Honor Hold Fear Bug


WOW-Reborn:
**Description:**

Honor Hold is totally messed up with the infernal hellfire.

**Current behaviour:**

CHANGEME Tell us what happens.
When hellfire is triggered in Honor Hold all the NPCs run around like crazy and get stuck in random places.  Also it makes your character run around and not able to control movement at all.

**Expected behaviour:**

From what I remember the NPCs moving around is normal but they are not returning to their location after the fear ends one example is the flight master.  But I'm not totally sure if the actual player is supposed to be feared around like crazy.

**Steps to reproduce the problem:**

Go to Honor Hold in Hellfire Peninsula 
Wait for the Infernal Hellfire to start raining down.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

8a45191e92a277ee9aafebbd0c1aa082c6be4298

**Operating system:** 

Windows Server 2016

WOW-Reborn:
It does exist, I don't know why your saying it doesn't but the infernal hellfire does rain down.

Killyana:
Yes I have to wait for a moment.
Anyway it's a general issue https://github.com/TrinityCore/TrinityCore/issues/21106

WOW-Reborn:
Question though, I know the NPCs are suppose to flee but was it normal for it to make the players toon run in fear as well?  I can't remember that far back.

Killyana:
This issue is closed be cause the main issue is a duplicate, the second issue is not really an issue as there's no proof.

Also it's better to have only one bug in the ticket not 2 or 3.

PR #25347, opened 2020-08-28
Core/Unit: send npcs home after fear runs out

### LINKED ISSUE #21106: Core/Movement: Feared NPCs do not return to their home position
**Description:** Feared NPCs do not return to their home position.

After you cast spell id 33815 (AoE fear) on any NPC in a crowd, they do not return home after the last fear move.
They keep standing at the last fear waypoint.

**Expected behaviour:**

They should return home.

1. `./cast self 33815` on an NPC in a crowd
2. watch the NPC until it gets released from Fear

**Branch(es):** Both

**TC rev. hash/commit:**  b08280fcfee500acf8b951ec2c05e0760d4b9955

Jinnaix:
**Changes proposed:**

-  send creatures back to home position after fear

**Target branch(es):**

- [x] 3.3.5

**Issues addressed:**

Closes #21106

Also gets rid of problems around the Honor Hold fear event (#22362 #24218)

**Tests performed:**

- [x] Builds
- [x] Creature out of combat before being feared and out of combat after fear runs out -> returns home
- [x] Creature out of combat before being feared but taken into combat before fear runs out -> continue combat
- [x] Creature in combat before being feared but out of combat before fear runs out -> return home
- [x] Creature in combat before being feared and in combat when fear runs out -> continue combat

none that im aware of

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -11392,6 +11392,10 @@ void Unit::SetFeared(bool apply)
             if (GetVictim())
                 SetTarget(EnsureVictim()->GetGUID());
         }
+        if (!(GetTypeId() == TYPEID_PLAYER) && !IsInCombat())
+        {
+            GetMotionMaster()->MoveTargetedHome();
+        }
     }
 
     // block / allow control to real player in control (eg charmer)
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -11391,10 +11391,8 @@ void Unit::SetFeared(bool apply)
             GetMotionMaster()->Remove(FLEEING_MOTION_TYPE);
             if (GetVictim())
                 SetTarget(EnsureVictim()->GetGUID());
-        }
-        if (!(GetTypeId() == TYPEID_PLAYER) && !IsInCombat())
-        {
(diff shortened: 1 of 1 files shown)


Jildor:
could it affect controlled units?

Jinnaix:
> could it affect controlled units or pets?

Can you give me a example that you would like to have tested with the expected result?

Jildor:
Fear on pet and leave combat and see what happens when remove fear effect

I'm not sure if when you control a unit (by example mind control) and unit get fear effect, lose mind control or not, in case mind control continue, try leave combat and see what happens when remove fear effect.

Jinnaix:
Both scenarios are very unlikely to happen, but here are the results:
A)
If you fear a pet and get out of combat, the pet will return home with its owner.
B)
If you mindcontrol a unit (this takes you in combat with this unit), and it gets feared, it will break your mind control, stay in combat with you and attack you after fear runs out.
If you mindcontrol a unit and that unit gets feared and you get ot of combat in this situation, the fear effect is removed immediately and the unit will return home.

jackpoz:

Jinnaix:
I have nothing left to add here - all tests successful. 

I will create a new issue and try to solve https://github.com/TrinityCore/TrinityCore/issues/21106#issuecomment-491614049 afterwards since this is not related to fear but should be handled the same.

jackpoz:
Thanks for the PR :)

PR #25353, opened 2020-08-28
[3.3.5] Cleanup unit tests + more of them

### LINKED ISSUE #25349: Unit tests do not have C++17 enabled
3.3.5 bf8b34946db51e27a25947316e2d4bb7a58a0e75

Microsoft Visual Studio Community 2019 v16.7.2

If you try to include a header that uses C++17 features (such as `Optional.h` or `StringConvert.h`) in unit tests, they will fail to build.

Treeston:
Change folder structure to be more scalable.

Add C++17 support (closes #25349).

Add unit tests for `StringConvert.h`


tests/CMakeLists.txt
```diff
@@ -9,20 +9,26 @@
...
+  TEST_SOURCES
 )
 
-add_executable(tests-common ${COMMON_SOURCES})
+GroupSources(${CMAKE_CURRENT_SOURCE_DIR})
 
-target_link_libraries(tests-common
+add_executable(tests ${TEST_SOURCES})
+
+set_property(TARGET tests PROPERTY CXX_STANDARD 17)
```
jackpoz, line 20: Have you tried targetting  trinity-feature-interface ? I don't think we set CXX_STANDARD anywhere else

Treeston: is L24 not enough? `trinity-core-interface` targets `trinity-default-interface` targets `trinity-feature-interface`

Treeston: @Shauren do cmake magic please

jackpoz: @Carbenium can you please take a lookt at this to avoid setting CXX_STANDARD in multiple places in the sources ?

image

It seems Catch2 likes to set its own CXX_STANDARD

PR #25364, opened 2020-08-30
Trinity::StringTo<double> support

Treeston:


src/common/Utilities/StringConvert.h
```diff
@@ -171,6 +170,91 @@ namespace Trinity::Impl::StringConvertImpl
             return (val ? "1" : "0");
         }
     };
+
+#if TRINITY_COMPILER == TRINITY_COMPILER_MICROSOFT
```
Treeston, line 174: @Shauren is there a better way to feature test from_chars support for `double`?

Shauren: using cmake compile or run test but i dont think bloating those is a good idea, this quick ifdef is fine

src/common/Utilities/StringConvert.h
```diff
@@ -171,6 +170,91 @@ namespace Trinity::Impl::StringConvertImpl
...
+            try {
+                if (str.empty())
+                    return std::nullopt;
+
+                if ((base == 10) && StringEqualI(str.substr(0, 2), "0x"))
+                    return std::nullopt;
+
+                std::string tmp;
+                if (base == 16)
+                    tmp.append("0x");
```
Shauren, line 240: What if:
1) base = 0, input = 0x0, msvc = 1
2) base = 0, input = 0x0, msvc = 0
3) base = 16, input = 0x0, msvc = 1
4) base = 16, input = 0x0, msvc = 0
5) base = 16, input = 0, msvc = 1
6) base = 16, input = 0, msvc = 0

does switching msvc on cause change in behavior?

Treeston: No. `0x0` is never parsed as hex, not even with `chars_format::hex`. It will be parsed as `0` with tail `x0`, which we then fail.

_(There are some cases where the compatibility layer may be more permissive than the `from_chars` implementation. There should be no cases where it is less permissive.)_

PR #25384, opened 2020-09-03
Core/Movement: Remove MOVEMENTFLAG_MASK_MOVING and MOVEMENTFLAG_FLYIN…

HoustonCore:
**Changes proposed:**

- Remove flag can fly when mobs falling.

- [√ ] 3.3.5

**Tests performed:**

**Follow-up**
- We need more checks and feedback.
- Please correct me in the standards.

[Video](https://www.youtube.com/watch?v=dmKeSj49Inc&feature=youtu.be)

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -911,14 +911,20 @@ void MotionMaster::MoveFall(uint32 id/* = 0*/)
     if (_owner->HasUnitState(UNIT_STATE_ROOT | UNIT_STATE_STUNNED))
         return;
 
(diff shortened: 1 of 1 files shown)


jackpoz:
which scenario does this fix ?

HoustonCore:
> which scenario does this fix ?
> could you check what happens to the corpses of flying creatures when they die ?

Now I will try to record a video before and after)

HoustonCore:
> which scenario does this fix ?
> could you check what happens to the corpses of flying creatures when they die ?

Done [video](https://www.youtube.com/watch?v=dmKeSj49Inc&feature=youtu.be)

Shauren:
This is a completely wrong way to fix that.

HoustonCore:
Ok

PR #25388, opened 2020-09-03
[CI] .npc command conversion

Treeston:


Shauren:
Did you have to re-indent all the namespaces?

Treeston:
now that we have nested-namespace (thanks c++17) i feel they make it way more explicit that these definitions all live in a namespace rather than the global namespace

PR #25404, opened 2020-09-05
Core/Unit: Fix units on vehicles not dismounting on death

### LINKED ISSUE #23095: Core/Unit: prevent the full cleanup of movements on death
**Description:** When you kill an accessory of a vehicle it remains installed and moves with the vehicle.

1.  .go c id 29698 OR  27213
2. Kill the rider
3. Notice the rider remains in the vehicle once dead

**Branch(es):** 3.3.5

**TC rev. hash/commit:** rev. 78d6115fdc

jackpoz:
**Changes proposed:**

- Fix units on vehicles not dismounting on death by allowing the dismount movement to finish
- Issue added in 982643cd96790ffc54e7a3e507469649f3b074d2

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #23095

**Tests performed:**

Tested #23095 steps ingame

- [ ] No clue if there is a case where this would bug out, with a dead unit still moving. Maybe in that case we should call GetMotionMaster()->Clear() before

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8663,6 +8663,8 @@ void Unit::setDeathState(DeathState s)
     // Death state needs to be updated before RemoveAllAurasOnDeath() is called, to prevent entering combat
     m_deathState = s;
 
+    bool isOnVehicle = GetVehicle() != nullptr;
+
     if (s != ALIVE && s != JUST_RESPAWNED)
     {
         CombatStop();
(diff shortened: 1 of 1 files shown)


Shauren:
Ref your TODO: unit on a vehicle should never have any movement generator active for itself

You are also checking isOnVehicle twice

jackpoz:
fixed the double check of isOnVehicle

Jildor:
I do not remember well, but I think that Mimiron (ulduar boss) mount 3 npc, one on top of the other forming a totem, I do not remember if in the script they really die or not, I do not know if this could affect him or is unrelated

Faq:
> 
> 
> I do not remember well, but I think that Mimiron (ulduar boss) mount 3 npc, one on top of the other forming a totem, I do not remember if in the script they really die or not, I do not know if this could affect him or is unrelated

They not die, is set HP 1.

Faq:
Tested with travalers Tundra mommoth and the 2 npcs on it, works just fine (felt on ground).

jackpoz:
Thanks for the tests @Faq 🚀

PR #25406, opened 2020-09-05
Core/Maps: Always update the grid of player summons even if far away

### LINKED ISSUE #24212: Shaman's totem will not disappear
Description:
Shaman releases the totem, and then goes as far as possible (can also see the totem), waiting for the end of the totem countdown, the countdown will stay at 0 seconds, the totem will not disappear,  still valid, can be recycled, can attack.

Expected behaviour:
Totem can disappear correctly.

- releases a totem
- goes as far as possible (can also see the totem)
- waiting for the end of the totem countdown

Branch(es):
3.3.5

TC rev. hash/commit:
cbe765d

Operating system:
CentOS

jackpoz:
**Changes proposed:**

-  Always update the grid of player summons even if far away

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #24212

**Tests performed:**

None yet

- [ ] Some more type of summons/guardians should be added

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -830,8 +830,10 @@ void Map::Update(uint32 t_diff)
                 VisitNearbyCellsOf(unit, grid_object_update, world_object_update);
         }
 
-        { // Update player's guardians
+        { // Update player's summons
             std::unordered_set<Unit*> toVisit;
+
+            // Totems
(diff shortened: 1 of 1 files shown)


jackpoz:
Before:
image

jackpoz:
Note: the totem will still show as 0s until the player moves a bit (any direction is fine) to trigger visibility update (or object update or whatever) but the totem is no longer in world already anyway (it can be checked with .npc info)

PR #25417, opened 2020-09-07
Move some files to CoreAI dir

illunix:
**Changes proposed:**
 
*Move some files to CoreAI dir

**Target branch(es):**
 
* [x]  3.3.5
 
**Tests performed:**
 
Builds


Shauren:
I disagree, CoreAI holds "final" ai classes that have no additional scripted behavior and work with just db data

CreatureAI and all the stuff you moved is a base for implementing both CoreAI and ScriptedAI

PR #25430, opened 2020-09-09
Scripts/Commands: Convert argument parsing of lfg commands to new system

Carbenium:
**Changes proposed:**

-  Scripts/Commands: Convert argument parsing of lfg commands to new system

- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/server/scripts/Commands/cs_lfg.cpp
+++ b/src/server/scripts/Commands/cs_lfg.cpp
@@ -91,8 +91,8 @@ public:
 
         Group* groupTarget = nullptr;
 
-        if (player->IsConnected())
-            groupTarget = player->GetConnectedPlayer()->GetGroup();
+        if (Player* target = player->GetConnectedPlayer())
+            groupTarget = target->GetGroup();
         else
         {
             CharacterDatabasePreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_SEL_GROUP_MEMBER);
--- a/src/server/scripts/Commands/cs_lfg.cpp
+++ b/src/server/scripts/Commands/cs_lfg.cpp
@@ -27,7 +27,9 @@
 #include "Player.h"
 #include "RBAC.h"
 
-void GetPlayerInfo(ChatHandler* handler, Player* player)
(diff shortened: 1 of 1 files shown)


src/server/scripts/Commands/cs_lfg.cpp
```diff
@@ -64,48 +66,45 @@ class lfg_commandscript : public CommandScript
...
+        if (!player)
+            player = PlayerIdentifier::FromTargetOrSelf(handler);
+        if (!player)
             return false;
 
         Group* groupTarget = nullptr;
 
-        if (playerTarget)
-            groupTarget = playerTarget->GetGroup();
+        if (player->IsConnected())
```
Treeston, line 94: `if (Player* target = player->GetConnectedPlayer())`

src/server/scripts/Commands/cs_lfg.cpp
```diff
@@ -115,47 +114,36 @@ class lfg_commandscript : public CommandScript
...
-            int32 tmp = atoi(str);
-            if (tmp > -1)
-                options = tmp;
-        }
-
-        if (options != -1)
+        if (optionsArg)
         {
-            sLFGMgr->SetOptions(options);
+            sLFGMgr->SetOptions(*optionsArg);
```
Treeston, line 133: what are the values for this? are they some enum?

src/server/scripts/Commands/cs_lfg.cpp
Carbenium: Does it make sense to introduce a local for that though I'm only interested in the group?

src/server/scripts/Commands/cs_lfg.cpp
Carbenium: It's a bitmask. Same as the `DungeonFinder.OptionsMask` config value.

src/server/scripts/Commands/cs_lfg.cpp
Treeston: Well, you use that same value in the next line, so...

Treeston (approved):

PR #25431, opened 2020-09-09
Core/ChatCommands: Drop PlayerIdentifier's cast operators which could return nullptr

Carbenium:
**Changes proposed:**

-  Core/ChatCommands: Drop PlayerIdentifier's cast operators which could return nullptr

- [x] 3.3.5
- [ ] master

### DIFF
--- a/src/server/game/Chat/ChatCommands/ChatCommandTags.h
+++ b/src/server/game/Chat/ChatCommands/ChatCommandTags.h
@@ -149,12 +149,11 @@ namespace Trinity::ChatCommands
         PlayerIdentifier() : _name(), _guid(), _player(nullptr) {}
         PlayerIdentifier(Player& player);
 
-        operator ObjectGuid() { return _guid; }
-        Player* operator->() const { return _player; }
-        explicit operator bool() const { return (_player != nullptr); }
-        bool operator!() const { return (_player == nullptr); }
+        operator ObjectGuid() const { return _guid; }
+        operator std::string const&() const { return _name; }
+        operator std::string_view() const { return _name; }
 
-        std::string const& GetName() { return _name; }
+        std::string const& GetName() const { return _name; }
         ObjectGuid GetGUID() const { return _guid; }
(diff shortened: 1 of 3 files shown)


src/server/game/Chat/ChatCommands/ChatCommandTags.h
```diff
@@ -149,9 +149,6 @@ namespace Trinity::ChatCommands
         PlayerIdentifier() : _name(), _guid(), _player(nullptr) {}
         PlayerIdentifier(Player& player);
 
-        template <typename T>
-        operator std::enable_if_t<std::is_base_of_v<T, Player>, T*>() const { return static_cast<T*>(_player); }
-        operator value_type() const { return _player; }
         operator ObjectGuid() { return _guid; }
         Player* operator->() const { return _player; }
         explicit operator bool() const { return (_player != nullptr); }
```
Treeston, line 154: Get rid of these too please (replace them with calls to `IsConnected` if present).

Also add conversion operators for `std::string const&` and `std::string_view`.

Treeston: by these i mean the ! and bool operators.

arguably get rid of the -> operator too

src/server/scripts/Commands/cs_character.cpp
```diff
@@ -713,22 +712,22 @@ class character_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleCharacterEraseCommand(ChatHandler* handler, PlayerIdentifier player)
+    static bool HandleCharacterEraseCommand(ChatHandler* handler, PlayerIdentifier playerIdent)
```
Treeston, line 715: I don't like this, other commands use `PlayerIdentifier player` and `Player* target`.

Treeston: like i'm not saying it needs to stay like that but try to keep them unified i guess?

Issue #23269, opened 2019-05-11
Core/Spell: Orientation changes after casting a spell in self with TARGET_UNIT_TARGET_ALLY


Killyana:
**Description:**  If an creature cast a spell that have targets (21, 0) (TARGET_UNIT_TARGET_ALLY, NO_TARGET) on self it will change orientation to 0 But it will not be visible to the client until the unit is reloaded (leave the area and come back), Even if .gps shows 0 as orientation, you can test with the spell 33482.

**Steps to reproduce the problem:**
1. .go c 66839
2. .gps and notice the orientation
3. .cast self 33482
4. .gps and notice the orientation changed to 0 even if visually it's still the same
5. .tele test   and   .rec
6. Notice the npc changed the orientation visually

**Branch(es):** 3.3.5

**TC rev. hash/commit:** rev. cf37f357df5c

Killyana:
Could also be seen in Skorn - Howling Fjord with this npcs (23663,23664,23665,23666)

offl:
> But it will not be visible to the client until the unit is reloaded (leave the area and come back)

Maybe I didn't understood what means 'not be visible to the client' but the case below is visible without leaving the area:

Apply https://github.com/TrinityCore/TrinityCore/issues/18653#issuecomment-629174231 (if not yet applied)
.go c id 23316
.q a 11061
.q c 11061
turn in the quest
Creature will cast on you the spell  41109 Periodic - Summon Torkus Jr and will change the orientation

<details>

```text
ID - 41109 Periodic - Summon Torkus Jr.
=================================================
Description: Trigger that causes the player to summon Torkus Jr.
Category = 0, SpellIconID = 2158, activeIconID = 0, SpellVisual = (0,0), SpellPriority = 0
Family SPELLFAMILY_GENERIC, flag [0] 0x00000000 [1] 0x00000000 [2] 0x00000000

SpellSchoolMask = 1 (SPELL_SCHOOL_MASK_NORMAL)
DamageClass = 0 (SPELL_DAMAGE_CLASS_NONE)
PreventionType = 0 (SPELL_PREVENTION_TYPE_NONE)
=================================================
Attributes: 0x00000180 (SPELL_ATTR0_HIDDEN_CLIENTSIDE, SPELL_ATTR0_HIDE_IN_COMBAT_LOG)
[...]

PR #25440, opened 2020-09-10
Core/Spells: Cleaned up and improved spell focusing behavior

Ovahlord:
**Changes proposed:**

-  removed a good chunk of hacks and wrong handlings and improved overall handlings for spell focusing, especially for channeled spells

- [x] 3.3.5
- [ ] master

**Tests performed:**
- tested on 4.x ingame

** For folks that wanna check for quality**
- Ingvar the Plunderer's Slam attacks should not turn during cast and hit the right direction
- Spells such as Drain Life casted by creatures should now always return correct internal orientation values
- Channeled spells that are not suposed to track targets will no longer cause the caster to face you during cast

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -3116,16 +3116,15 @@ SpellCastResult Spell::prepare(SpellCastTargets const& targets, AuraEffect const
         }
     }
 
-    // focus if not controlled creature
-    if (m_caster->GetTypeId() == TYPEID_UNIT && !m_caster->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_POSSESSED))
+    // Creatures focus their target when possible
+    if (m_casttime && m_caster->IsCreature() && !m_spellInfo->IsNextMeleeSwingSpell() && !IsAutoRepeat() && !m_caster->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_POSSESSED))
     {
-        if (!(m_spellInfo->IsNextMeleeSwingSpell() || IsAutoRepeat()))
-        {
-            if (m_targets.GetObjectTarget() && m_caster != m_targets.GetObjectTarget())
-                m_caster->ToCreature()->SetSpellFocus(this, m_targets.GetObjectTarget());
-            else if (m_spellInfo->HasAttribute(SPELL_ATTR5_DONT_TURN_DURING_CAST))
-                m_caster->ToCreature()->SetSpellFocus(this, nullptr);
-        }
+        // Channeled spells and some triggered spells do not focus a cast target. They face their target later on via channel object guid and via spell attribute or not at all
+        bool const focusTarget = !m_spellInfo->IsChanneled() && !(_triggeredCastFlags & TRIGGERED_IGNORE_SET_FACING);
+        if (focusTarget && m_targets.GetObjectTarget() && m_caster != m_targets.GetObjectTarget())
+            m_caster->ToCreature()->SetSpellFocus(this, m_targets.GetObjectTarget());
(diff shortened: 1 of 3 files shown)


offl:
Closes #23269, just tested all cases

jackpoz:
@Ovahlord if you want to test my circle ci fix it helps if you actually merge that commit into your branch 😄 
besides, the fix has no effect if the PR has already a successful build, it affects only new PRs

src/server/game/Spells/Spell.cpp
```diff
@@ -3814,6 +3806,17 @@ void Spell::update(uint32 difftime)
...
                             unit->RemoveOwnedAura(m_spellInfo->Id, m_originalCasterGUID, 0, AURA_REMOVE_BY_CANCEL);
                 }
+                else
+                {
+                    // Update serverside orientation of channeling units
+                    if (Unit* unitCaster = m_caster->ToUnit())
+                    {
+                        ObjectGuid const channelGuid = unitCaster->GetChannelObjectGuid();
+                        if (!channelGuid.IsEmpty() && channelGuid != unitCaster->GetGUID() && GetSpellInfo()->HasAttribute(SPELL_ATTR1_CHANNEL_TRACK_TARGET))
+                            if (WorldObject const* objectTarget = ObjectAccessor::GetWorldObject(*unitCaster, channelGuid))
```
Treeston, line 3816: do we truly need this on every tick?

is it not sufficient to do it whenever the channel would "do" anything (trigger a spell, periodic tick on an aura, etc.)?

are there any channels that truly need _constant_ facing updates?

Ovahlord: Yes. channeled spells that periodically trigger a spell that for example cone target spells that constantly need a up to date angle

Treeston: sure, but couldn't we just update the facing when the channel aura triggers a spell?

i.e. before processing any periodic event of the channel aura, update facing

Ovahlord:
Aight, place your final complaints before I'm hitting the button

src/server/game/Spells/Spell.cpp
```diff
@@ -3116,16 +3116,15 @@ SpellCastResult Spell::prepare(SpellCastTargets const& targets, AuraEffect const
         }
     }
 
-    // focus if not controlled creature
-    if (m_caster->GetTypeId() == TYPEID_UNIT && !m_caster->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_POSSESSED))
+    // Creatures focus their target when possible
+    if (m_casttime && m_caster->IsCreature() && !m_spellInfo->IsNextMeleeSwingSpell() && !IsAutoRepeat() && !m_caster->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_POSSESSED))
```
Treeston, line 3120: did you want to change this possessed flag? i remember you asking about it

Treeston (approved):

src/server/game/Spells/Spell.cpp
Ovahlord: Yeah, but I don't know how possessed focusing is suposed to look so I keep it for the time being until I know more

src/server/game/Spells/Spell.cpp
Jildor: I think these comments should be considered, maybe this has broken things: https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/commit/d606fa4688c01652948be36097014c029b43f1a4?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDExODI3MTMxNzI6NTIxMDM2

PR #25443, opened 2020-09-10
[3.3.5] Error messages from Chat command argument parsing

Treeston:
Argument handlers now return `ChatCommandResult` instead of `Optional<std::string_view /*Tail*/>`.

`ChatCommandResult` cosplays as `std::optional` - it constructs from `std::string_view`, and `std::nullopt`. However, it also constructs from `std::string&&` and `char const*`, allowing you to specify custom error messages in argument parsers.

This means invalid character names or item IDs will no longer just hit you with the command usage, but will display a meaningful error message, too.

Yay for progress!


Treeston:
@Carbenium this shouldn't interfere with your PRs if you only touch cs_ files, but maybe don't touch the internals until this is done.

Shauren (changes requested):
1) All of error messages **must** be localized with trinity_string (like every other command string outside of cs_debug)
2) Make a helper function `std::string(_NOT_VIEW) Demangle(std::type_info const& typeInfo)` with implementation in a .cpp file (don't include anything from `<boost>` in a header if you can avoid it)

Shauren (approved):

PR #25447, opened 2020-09-11
Dep/Fmt: update fmt library to 7.0.3

Winfidonarleyan:
**Changes proposed:**
-  Update `fmt` lib to https://github.com/fmtlib/fmt/releases/tag/7.0.3

**Target branch(es):**
- [x] 3.3.5

**Tests performed:**
- Build only


jackpoz:
please test it a bit more than just "build only"

Shauren:
I will take a look at this later but I would prefer not to have library upgrades done in PRs at all

Winfidonarleyan:
> please test it a bit more than just "build only"

Okay

PR #25455, opened 2020-09-14
Script/Commands: Add ".pdump copy" command

### LINKED ISSUE #22921: pcopy command

jackpoz:
**Changes proposed:**

- Add ".pdump copy" command
Syntax: .pdump copy $playerNameOrGUID $account [$newname] [$newguid]
Copy character with name/guid $playerNameOrGUID into character list of $account with $newname, with first free or $newguid guid.

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #22921

**Tests performed:**

Everything was tested, but then more code changes have been applied so need to re-test it again.

- [ ] ".pdump write" needs to be tested
- [ ] ".pdump load" needs to be tested
- [ ] ".pdump copy" needs to be tested


Treeston:
is there ever a case where $newname can be optional considering the character already exists?

Aokromes:
> is there ever a case where $newname can be optional considering the character already exists?

core already forces to rename duplicated names.

Treeston:
Oh it will force rename on login? Nice.

jackpoz:
This command is just a shortcut for ".pdump write" + ".pdump load" without passing through a File stored on disk, keeping every validation check that each single command does.

src/server/game/Tools/PlayerDump.h
```diff
@@ -19,6 +19,7 @@
 #define _PLAYER_DUMP_H
 
 #include <string>
+#include <istream>
```
Shauren, line 22:
```suggestion
#include <iosfwd>
```

src/server/game/Tools/PlayerDump.cpp
```diff
@@ -820,14 +826,13 @@ DumpReturn PlayerDumpReader::LoadDump(std::string const& file, uint32 account, s
...
-                break;
+        input.getline(buf, BUFFER_SIZE);
+        if (input.eof())
+            break;
+        if (input.fail())
             return DUMP_FILE_BROKEN;
-        }
 
         std::string line;
         line.assign(buf);
```
Shauren, line 838: All of this could be just
```
std::string line;
while (std::getline(input, line))
{
    // do stuff with line
}
if (input.fail())
    return DUMP_FILE_BROKEN;

// normal success code... think it commits transaction there?
```

jackpoz: looks like some nice refactoring to me (that maybe I don't want to include in this PR)

Shauren: You are already killing FILE* and replacing it with as-similar-as-possible looking code using istream (which is not how you should be using istreams)

PR #25490, opened 2020-09-24
Core/Spells: SPELL_AURA_MOD_HEALING_PCT should be negative if TargetType is enemy or BasePoints are negative

### LINKED ISSUE #24166: bladestorm
**Current behaviour:**
During bladestorm , when you cast frostbolt or ice chains on warrior they count as buff instead of debuff (you can spellsteal .eg) and when bladestorm finishes they wont slow the warrior until the buff duration is ended (you can cast those spells and refresh the buff)

CHANGEME Tell us what should happen instead.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 
https://github.com/TrinityCore/TrinityCore/commit/5f20715e2f38a6eb721f443df9e13c84ecb11fbb

Windows 10 64bit

Jildor:
**Changes proposed:**

-  SPELL_AURA_MOD_HEALING_PCT should be negative if TargetType is enemy or BasePoints are negative

- [X] 3.3.5
- [ ] master

**Issues addressed:**

Closes #24166

**Tests performed:**

tested in-game


jackpoz:
Thanks for the PR :)

PR #25522, opened 2020-10-01
Core/Spells: unify OnSpellCastInterrupt and OnSuccessfulSpellCast hooks 

Ovahlord:
**Changes proposed:**

-  OnSpellCastInterrupt and OnSuccessfulSpellCast  now share one hook and are getting distinguished by a new enum value.
-  Additionally a new hook call scenario has been added which gets triggered when a channeled spell ends

- [x] 3.3.5
- [ ] master

**Tests performed:**
- tested and used on 4.x on many different boss encounters

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -3825,7 +3825,7 @@ void Spell::update(uint32 difftime)
                 // We call the hook here instead of in Spell::finish because we only want to call it for completed channeling. Everything else is handled by interrupts
                 if (Creature* creatureCaster = m_caster->ToCreature())
                     if (creatureCaster->IsAIEnabled())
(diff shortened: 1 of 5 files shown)


src/server/game/Spells/Spell.cpp
```diff
@@ -3821,6 +3821,10 @@ void Spell::update(uint32 difftime)
             {
                 SendChannelUpdate(0);
                 finish();
+
+                if (Creature* creatureCaster = m_caster->ToCreature())
```
ccrs, line 3825: random thought, why not inside Spell::finish() ?

Ovahlord: because there is no need for that anymore. spell_finish_successful cast already covers all cases where a spell is getting flagged as 'finished' straight afterwards. This case here is explicitely only for finished channel cases

PR #25527, opened 2020-10-02
First boss of VoTW dungeon Legion 

searcas:
**Changes proposed:**

-  
-  
-  

**Target branch(es):** master

- [x] master

**Issues addressed:**

**Tests performed:**
[Hatred] Doesn't work at all.
[Darkstrike] Need to attack with both weapons.
[Swoop] Just jumping to target atm, need to damage everyone in his path and knockback all players
[Swoop] Speed of swoop/jump need to be increased in core.

- [ Need more scripts on spells] 
- [ Need polishing timers]


jackpoz:
I would suggest to create a new branch when working on something instead of using the default one "master" (or "3.3.5"). I allows to work on different things at the same time if needed.

PR #25555, opened 2020-10-10
Core/ChatCommands: Fix a potential crash

Jildor:
**Changes proposed:**

-  Well, I don't see a need to crash server just to avoid log a command linked to rbac role player in the GM log
-  Therefore I decide to avoid this check in case the rbac does not exist and log an error
-  I know it's stupid to change or delete the default roles, but I don't see a need to crash

- [X] 3.3.5
- [ ] master

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/game/Chat/ChatCommands/ChatCommand.cpp
+++ b/src/server/game/Chat/ChatCommands/ChatCommand.cpp
@@ -140,8 +140,14 @@ static void LogCommandUsage(WorldSession const& session, uint32 permission, std:
     if (AccountMgr::IsPlayerAccount(session.GetSecurity()))
         return;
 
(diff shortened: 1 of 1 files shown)


Shauren:
hmm... this guy definitely did just that. deleted default role

Jildor:
Of course, I had to delete it to check the log, a colleague told me that has (for some reason) different permission id and roles and he got the crash

Treeston:
I feel like this is not the appropriate location to deal with this.

The default roles existing is something the core gets to assume. If they do not exist, that is something that should be checked (and fixed) and startup.

Jildor:
oh, I see, then I should check if the rbac defined in the core exist in the DB at startup.
In that case I prefer to create an issue indicating this and that someone with more experience can do it.
I'll close this issue, thanks;)
#25573

PR #25560, opened 2020-10-11
Script/Commands: .character changeaccount now kick the player before being transferred in case of being online

Jildor:
**Changes proposed:**

-  kick the player before being transferred in case of being online to avoid problems (I have even been able to generate a crash doing the idiot with the command before this PR)

- [X] 3.3.5
- [ ] master

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/scripts/Commands/cs_character.cpp
+++ b/src/server/scripts/Commands/cs_character.cpp
@@ -494,8 +494,7 @@ public:
         }
 
         if (Player* onlinePlayer = player->GetConnectedPlayer())
-            if (WorldSession* session = onlinePlayer->GetSession())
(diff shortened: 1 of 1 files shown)


src/server/scripts/Commands/cs_character.cpp
```diff
@@ -493,6 +493,10 @@ class character_commandscript : public CommandScript
             }
         }
 
+        if (Player* onlinePlayer = player->GetConnectedPlayer())
+            if (WorldSession* session = onlinePlayer->GetSession())
```
Shauren, line 497: Unneccessary, existing `Player` will always have a nonnull worldsession

Jildor: thx, done ;)

Treeston (approved):

PR #25625, opened 2020-10-31
Improve multithreading of mmaps_generator

jackpoz:
**Changes proposed:**

-  Improve multithreading of mmaps_generator by letting threads handle single tiles instead of whole maps
- Add various new CMake options to support different *San sanitizer tools when using Clang

- [x] 3.3.5
- [ ] master

**Issues addressed:**

None

**Tests performed:**

- Tested map 30 with 4 threads, the result looks good in RecastDemo

- [x] Run *San tests
- [ ] Check ingame that mmaps still work
- [x] Fix mmaps_generator not closing when processing 1 single tile


src/tools/mmaps_generator/MapBuilder.cpp
```diff
@@ -176,40 +180,46 @@ namespace MMAP
...
 
             if (_cancelationToken)
                 return;
 
-            buildMap(mapId);
+            dtNavMesh* navMesh = dtAllocNavMesh();
+            if (!navMesh->init(&tileInfo.m_navMeshParams))
+            {
+                printf("[Map %03i] Failed creating navmesh for tile %i,%i !\n", tileInfo.m_mapId, tileInfo.m_tileX, tileInfo.m_tileY);
+                return;
```
Shauren, line 194: leaks navmesh if init fails

Faq:
`**Store TileBuilder as raw pointer for now, it will be changed later on** `
When?

jackpoz:
> `Store TileBuilder as raw pointer for now, it will be changed later on`
> Out of this PR?

It depends, first I want to make sure it works and there are no race conditions. There are a few ToDo left and some are about polishing, like moving TileBuilder class to its own file, so I would say modern C++ refactoring would fit better in another PR

jackpoz:
This PR is now complete from my side

PR #25684, opened 2020-11-30
Core/Player: auto-attacked player stands up from sit or sleep state

aurelio4:
**Changes proposed:**
-  When auto attacking a sitting/sleeping/kneeling player, they should stand up.

- [x] 3.3.5
- [ ] master

**Issues addressed:**
Fixed a bug that caused a sitting/sleeping/kneeling player to not stand up when getting auto attacked.

**Tests performed:**
builds, tested, and works.

**Known issues and TODO list:** N/A

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -946,6 +946,10 @@ bool Unit::HasBreakableByDamageCrowdControlAura(Unit* excludeCasterChannel) cons
         }
     }
 
+    // check to see if victim is sitting
+    if (victim->IsSitState())
+        victim->SetStandState(UNIT_STAND_STATE_STAND);
+
     return damage;
 }
 
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -947,7 +947,7 @@ bool Unit::HasBreakableByDamageCrowdControlAura(Unit* excludeCasterChannel) cons
     }
 
     // check to see if victim is sitting
-    if (victim->IsSitState())
+    if (victim->GetStandState())
         victim->SetStandState(UNIT_STAND_STATE_STAND);
 
     return damage;


offl:
What if player sleeps?

illfated:
Fair point. I know from previous experience on a retail server back in 2005-2006 that when I sat down while being attacked by PvP players next to Hammerfall, my character was always forced to stand up. 
(Soon after that I relocated to a non-PvP realm, finding the PvP experience not suitable to my own taste for gameplay.)

aurelio4:
> What if player sleeps?

This is also a problem. I'm working on a fix for sleeping and sitting later tonight. I'll update when that's done.

offl:
Shouldn't we just check if stand state is not stand?

illfated:
> Shouldn't we just check if stand state is not stand?

Sounds reasonable, if the same rule applies to sleeping state (using `/sleep`) and is confirmed.

aurelio4:
Updated and tested. Melee auto attacks will now cause the victim to stand when the victim is sitting, sleeping, or kneeling.

illfated:
OK. May I also suggest a PR title, using the old-school format based on source tree and content change?

**Core/Player: auto-attacked player stands up from sit or sleep state**

aurelio4:
> OK. May I also suggest a PR title, using the old-school format based on source tree and content change?
> 
> **Core/Player: auto-attacked player stands up from sit or sleep state**

I've gone ahead and changed it, thanks for catching that.

jackpoz:
should this affect already players, in case some script set a stand state to a boss during a fight ?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -946,6 +946,10 @@ bool Unit::HasBreakableByDamageCrowdControlAura(Unit* excludeCasterChannel) cons
         }
     }
 
+    // check to see if victim is sitting
+    if (victim->GetStandState())
```
jackpoz, line 950: should this affect already players, in case some script set a stand state to a boss during a fight ?

jackpoz (changes requested):

src/server/game/Entities/Unit/Unit.cpp
jackpoz: @aurelio4 do you have a reply to my comment above ?

offl: Yeah, it's safer to check for player only

aurelio4 (commented):
@jackpoz 

> i suppose if the boss would cause the player to sit and then crit the player, the player would stand up. is there any boss in the game that does this though? even if there is, this is a blizzlike feature that should be implemented.

it's not a pvp specific feature. i quoted the wowwiki before, and it states:

> Melee attacks against a sitting target automatically hit and cause double damage. The first damage taken while sitting will cause a target to immediately stand. 

it doesn't specify whether it's for pvp or pve, but i'm willing to believe it's for both.
if there's anything you wish for me to change in the code to improve this, i'm open to anything. just found this feature that wasn't in the core and wanted to raise awareness to it.

jackpoz:
Thanks for the PR :)

PR #25690, opened 2020-12-06
Core/Visibility: Split BG/Arena visibility settings

Jildor:
**Changes proposed:**

-  Split BG/Arenas visibility settings to be able to handle different values if necessary
-  the maximum playable distance in arenas is ~180yd, but I prefer to keep the default value it had before (533)

- [X] 3.3.5
- [ ] master

**Tests performed:**

tested in-game


jackpoz:
I think it's ok if this PR doesn't add backward compatibility to the previous setting as it shouldn't affect that much

jackpoz (approved):

jackpoz:
Thanks for the PR :)

PR #25698, opened 2020-12-09
Attempt to fix crashes of chat commands added last summer

### LINKED ISSUE #25692: World Server Instant Crash upon typing .send items
**Description:**

I've recently compiled the newest build of 3.3.5, and i have found that with Rbac Permission ID: 484 - Send Items.
it doesn't display syntax like the other commands, instead it just instantly crashes the world server, no auth, DBErrors, World Server logs written.

**Expected behaviour:**

I was expecting to see the following after i had typed ".send items" and not have the worldserver self terminate without leaving a reason as to why.

Syntax: .send items #playername "#subject" "#text" itemid1[:count1] itemid2[:count2] ... itemidN[:countN] Send a mail to a player. Subject and mail text must be in "". If for itemid not provided related count values then expected 1, if count > max items in stack then items will be send in required amount stacks. All stacks amount in mail limited to 12.

due to ".send items" isn't the complete syntax to send an actual item.

1. Possibly create a fresh compile of 3.3.5 and configure everything that matches your requirements. 
2. Once you've got both auth and world server running load into the world and make sure you have gmlevel/securitylevel 3.
3. in the commands in game type .send items.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. d665c68cfcf1 2020-12-04 12:00:11 +0200 (3.3.5 branch) (Win64, Release, Static)

**Operating system:** Windows Server 2019 Standard - 64 bit

jackpoz:
**Changes proposed:**

-  Fix crashes

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #25692

**Tests performed:**

None

- [ ] Fix more crashes

### DIFF
--- a/src/common/Utilities/Util.cpp
+++ b/src/common/Utilities/Util.cpp
@@ -59,6 +59,11 @@ static char* strtok_static_context = nullptr;
 
 char* strtok_static(char* str, const char* delim)
 {
+    // Safeguard against calling strtok(nullptr,..) as first call as it would result in a crash otherwise.
+    // The first strtok() call must have a non-null pointer as first argument.
+    if (!str && !strtok_static_context)
+        return nullptr;
+
     return strtok_r(str, delim, &strtok_static_context);
 }
 
--- a/src/common/Utilities/Util.cpp
+++ b/src/common/Utilities/Util.cpp
@@ -55,7 +55,19 @@ std::vector<std::string_view> Trinity::Tokenize(std::string_view str, char sep, 
     return tokens;
 }
 
+static char* strtok_static_context = nullptr;
+
+char* strtok_static(char* str, const char* delim)
+{
+    return strtok_r(str, delim, &strtok_static_context);
+}
+
 #if (defined(WIN32) || defined(_WIN32) || defined(__WIN32__))
+char* strtok_r(char* str, const char* delim, char** saveptr)
+{
+    return strtok_s(str, delim, saveptr);
+}
+
 struct tm* localtime_r(time_t const* time, struct tm *result)
 {
     localtime_s(result, time);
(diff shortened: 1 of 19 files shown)


jackpoz:
> [ccrs](/ccrs) requested a review from [Treeston](/Treeston) [10 minutes ago](#event-4095224427)

@ccrs no thanks

Treeston:
I'm still going to tell you that your PR is horrible, whether you like it or not.

You are, yet again, not fixing any issues, just preventing them from being found by hiding them with that bizarre strtok construction. Stuff is still broken, it just fails silently now.

The actual issue in #25692 is that a lot of the old command handler code is relying on the `strtok` buffer being set by unrelated logic. `HandleSendMailCommand`, in particular, relies on `extractPlayerTarget` setting it. However, `extractPlayerTarget` shortcuts for empty arguments and never calls `strtok`, but still returns non-`nullptr`. This results in the crash in #25692 for empty arguments.

The immediate fix for #25692 is thus making sure that `extractPlayerTarget` populates the `strtok` buffer if it returns non-`nullptr` (until such a time when all of this `strtok` mess can be removed for good).

Making sure other commands don't crash is done by either moving them to the new system (preferred), or just making sure that `strtok(nullptr, ...)` calls are preceded by something setting the strtok buffer - either a non-`nullptr` call to `strtok`, or a call to one of the `extractX` methods on `ChatHandler` which returned non-`nullptr`. I went through all the existing `strtok(nullptr, ...)`s in commands, and this appears to be the case for all of them.

jackpoz:
Just for reference, this PR was in draft status. There's not much to review at this stage, especially since it was going to be modified with whatever local changes I had.

In general fixing a crash takes precedence over fixing a functionality. As time is limited, one pushes the crashfix, takes note of the broken functionalities and fixes them when there are no more urgent issues.

Nice to see you found a quite elegant solution that fixes both crash and broken functionality with just a couple of changed lines.

PR #25709, opened 2020-12-12
Core/Config: Fixed StartPlayLevels of DK, DH & RAF

RealLushen:
**Changes proposed:**

-  Modify the worldserver.conf so its up to date with 9.0.2

- master

**Issues addressed:**

- Console Errors due to wrong values

**Tests performed:**

- Yes it builds and works.


Shauren:
Not that simple, allied race (and pandaren) DKs start at 10, other races at 8
For sql you need to make update file that does `ALTER TABLE`

osleyder85:
Beware, the Death Knight Class player does not start at level 10, he starts at level 8 just like the Demon Hunter. This I just checked at Blizzard.

Aokromes:
read what shauren said.

illfated:
I assume that both the worldserver.conf settings and its matching core functions need to be updated to differentiate between (Allied Races + Pandaren) versus the Core Races, so this PR needs to be expanded to include extended core settings (pun intended).

PR #19231, opened 2017-03-01
Core/Misc : Personal spawn

Traesh:
**Changes proposed:**

-  Add a parameter "visibleBySummonerOnly" to the "SummonCreature" function to allow summon to be visible only by summoner
- Automatically set "visibleBySummonerOnly" to true for spell with SummonPropertiesFlag & 0x10 (see PR conversation for more informations)
- If summon is personnal spawn, all aggro is redirected to summoner, unless victim is also a personnal spawn with same summoner

**Target branch(es):** 3.3.5/master

- [ ] 3.3.5
- [x] master

**Issues addressed:**
- Closes #18254

**Tests performed:** (Does it build, tested in-game, etc.)
Compiled, tested ingame

**Known issues and TODO list:** 

- None

PR #25765, opened 2020-12-25
Core/Spells: Implemented personal summons (#19231)

jackpoz:
**Changes proposed:**

-  cherry pick commit b7bb5e6a98119512702519adecf559ecedc20743

- [x] 3.3.5
- [ ] master

**Issues addressed:**

None

**Tests performed:**

None

- Let CI check the build

### DIFF
--- a/src/server/game/Combat/ThreatManager.cpp
+++ b/src/server/game/Combat/ThreatManager.cpp
@@ -320,6 +320,23 @@ void ThreatManager::AddThreat(Unit* target, float amount, SpellInfo const* spell
             return;
         amount = 0.0f;
     }
+    else if (TempSummon* tempSummonVictim = target->ToTempSummon())
+    {
+        if (tempSummonVictim->IsVisibleBySummonerOnly())
+        {
+            if (Unit* tempSummonSummoner = tempSummonVictim->GetSummonerUnit())
+            {
+                // Personnal Spawns from same summoner can aggro each other
+                if (!_owner->ToTempSummon() ||
+                    !_owner->ToTempSummon()->IsVisibleBySummonerOnly() ||
+                    tempSummonVictim->GetSummonerGUID() != GetOwner()->ToTempSummon()->GetSummonerGUID())
+                {
+                    AddThreat(tempSummonSummoner, amount, spell, ignoreModifiers, ignoreRedirects);
+                    amount = 0.0f;
+                }
+            }
+        }
+    }
 
     // if we cannot actually have a threat list, we instead just set combat state and avoid creating threat refs altogether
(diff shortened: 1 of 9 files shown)


jackpoz:
@Treeston could you compare https://github.com/TrinityCore/TrinityCore/commit/b7bb5e6a98119512702519adecf559ecedc20743#diff-0e95a31f2d0f9af5b93027cf1c8eddd47d5678cf344be4eecdc54df3ef410700 with https://github.com/TrinityCore/TrinityCore/pull/25765/files#diff-0e95a31f2d0f9af5b93027cf1c8eddd47d5678cf344be4eecdc54df3ef410700 and see if I ported the changes correctly for ThreatManager ?

It doesn't have to be the best solution etc, I just need to know if it's a similar place as the original author. If the code needs improvements, it could be improved in another PR (as this is supposed to only be a cherry-pick with the least amount of changes to allow any improvements to be ported back to master branch too)

Treeston:
Taking a look. Do I understand the check correctly - if the threatening unit is a tempsummon that is only visible to its summoner, unless the threatened unit is also a tempsummon that is only visible to that same summoner, it redirects all threat to that summoner instead?

Treeston:
Move that entire check further up, just before/after the vehicle check that starts on line 315. Use the same logic it uses if the check passes.

Treeston:
(You can't really keep the original structure from `master`, it abuses redirect logic which wouldn't work in `3.3.5`, our redirect system is different because it supports multiple redirections.)

jackpoz:
I moved the code above and copied the behavior of Vehicle with setting the threat to 0 after redirecting it. Please let me know if now it's ok

Treeston:
Looks reasonable as a cherry pick.

(I have my doubts whether it is _correct_ in its behavior, though.)

funjoker:
There are probably small tweaks which can be adjusted in future. It’s basic support.

jackpoz:
And it works:
image

Treeston:
Neat.

jackpoz:
PvP doesn't seem to work that good as the summon killed an enemy player with the enemey player being clueless
image

Treeston:
That was kind of what I was alluding to above. All the current threat logic is doing is preventing the summon from threatening other units.

I think this should go in one of the CanEnterCombat (or whatever it's called, it's the big static check function on combatmanager). Actual behavior would need to be determined (it should be (I guess?) able to fight creatures, but not other players?)

jackpoz:
I think it should fit in WorldObject::CanSeeOrDetect() even, the personal summon should never target an enemy player.

Treeston:
That sounds reasonable, yes.

PR #25775, opened 2020-12-28
Core/Vehicle: add a way to delay the Vehicle despawn Time if its needed

gonzo1247:
**Changes proposed:**
-  Some Vehicles need a delayed despawn Time for some action after Player/Creature exit the vehicle. With this PR is this now possible. It's add a new Smart Action 'SMART_ACTION_SET_DELAY_VEHICLE_DESPAWN'  to delayed the despawn in SmartAI.
With the function 'SetTimedVehicleDespawn()' can this also in C++ Code used.
-  
-  

- [x] master

**Issues addressed:**

https://github.com/TrinityCore/TrinityCore/issues/18089
Now this issue can closed

**Tests performed:**
Build works, if ingame testet with different time with SmartAI and C++

- [ ] 
- [ ]


Shauren:
Wouldnt it be better to introduce a new db table `vehicle_template` that would have a column like that?

jackpoz:
did you miss to push the change where GetTimedVehicleDespawn() is used ? I wonder if the new field should be in Vehicle class itself

gonzo1247:
> 
> 
> did you miss to push the change where GetTimedVehicleDespawn() is used ? I wonder if the new field should be in Vehicle class itself

yea i missed to stage this change. its in the follow commit

gonzo1247:
> 
> 
> Wouldnt it be better to introduce a new db table `vehicle_template` that would have a column like that?

this is also possible - i change it and push it again.
That would have the advantage that the timer is there from the start.

gonzo1247:
Its been changed and works over a new table

src/server/game/Entities/Vehicle/Vehicle.cpp
```diff
@@ -885,3 +885,12 @@ void VehicleJoinEvent::Abort(uint64)
     if (Passenger->IsInWorld() && Passenger->HasUnitTypeMask(UNIT_MASK_ACCESSORY))
         Passenger->ToCreature()->DespawnOrUnsummon();
 }
+
+Milliseconds Vehicle::GetDelayDespawnTime()
+{
+    if (VehicleTemplateList const* temp = sObjectMgr->GetVehicleTemplate(this))
+        for (VehicleTemplateList::const_iterator itr = temp->begin(); itr != temp->end(); ++itr)
```
Shauren, line 892: What exactly was your plan by making this a list/vector?

gonzo1247: My first plan was that the hold more as one data - there was with a ForceRespawn Timer but this was discarded and i didn't change the ground part.

gonzo1247:
Or gives a bether/other way to impelemt this?
my idea is, that so can easyer expand if the vehicle_template becomes new fields

PR #25802, opened 2021-01-04
Scripts/Spells: Implement effect "Party Time" & Add Rejuvenation when eating Deviate Fish

r4d1sh:
**Changes proposed:**

-  Added one missed effect from eating Deviate Fish (https://classic.wowhead.com/item=6522/deviate-fish) - "Rejuvenation" (https://classic.wowhead.com/spell=8070/rejuvenation)
-  Implemented "Party Time" (https://classic.wowhead.com/spell=8067/party-time) effect from eating Deviate Fish. This effect causes your character to play a random emotion once in 5, 10 or 15 seconds. List of possible emotions if player moves: applaud, laugh, cheer, chicken. Also possible dance if player stands.

- [x] 3.3.5
- [ ] master

**Tests performed:**

Builded and tested on Win64


illfated:
Based on old-school title composing and this PR content, a suitable name could be something like:

**Scripts/Spells: implement effect "Party Time" & add Rejuvenation when eating Deviate Fish**

Alternative phrasing could also be used.

r4d1sh:
> Based on old-school title composing and this PR content, a suitable name could be something like:
> 
> **Scripts/Spells: implement effect "Party Time" & add Rejuvenation when eating Deviate Fish**
> 
> Alternative phrasing could also be used.

@illfated Thanks for your help in choosing suitable name for this commit and PR.

illfated:
You are welcome. To me, it helps me understand the individual changes when looking at the TC commit list later, when it has become part of the TC source.

src/server/scripts/Spells/spell_item.cpp
```diff
@@ -821,6 +822,60 @@ class spell_item_deviate_fish : public SpellScript
...
+        if (!player)
+            return;
+
+        _event = new PartyTimeEmoteEvent(player);
+        player->m_Events.AddEventAtOffset(_event, RAND(5s, 10s, 15s));
+    }
+
+    void HandleEffectRemove(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+    {
+        if (!_event)
```
jackpoz, line 862: Are auras removed before events when logging out so that _event will never point to deleted memory ?

sql/updates/world/3.3.5/XXXX_XX_XX_XX_world.sql
```diff
@@ -0,0 +1,2 @@
+DELETE FROM `spell_script_names` WHERE `ScriptName` = 'spell_item_party_time';
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES (8067, 'spell_item_party_time');
```
illfated, line 2: codestyle/assurance: add comment on first line, in case previously loaded SQL file does not have NewLine at EOF
```suggestion
-- 
DELETE FROM `spell_script_names` WHERE `ScriptName` = 'spell_item_party_time';
INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES (8067, 'spell_item_party_time');
```

src/server/scripts/Spells/spell_item.cpp
Keader: i wonder... in logout all m_Events are not deleted anyway?

r4d1sh: @jackpoz good point. I can override `BasicEvent::IsDeletable()`, but your idea with aura check inside event looks better.
@Keader they are deleted. This is the problem - my pointer will point to deleted memory.

r4d1sh:
Builded and tested in-game.

jackpoz:
Thanks for the PR :)

illfated:
Ditto, I concur. Thank you for contributing.

illfated:
BTW, if I can make a wish for something cosmetic regarding your **PR titles**, 
next time you contribute: Please, please, please avoid using uppercase " **i** ", 
because it looks like lowercase L in most Sans Serif fonts (typeface).

To me, it looks like

#### LMPLEMENT

in every PR title.

PR #20206, opened 2017-08-28
Core/Spells: implement Heartbeat Resist mechanic

Eliminationzx:
**Changes proposed:**
> _Spells with flag SPELL_ATTR0_HEARTBEAT_RESIST_CHECK will have a chance to break on each tick._
> _Creature will try to break aura based on time left 25, 50, 75% with a base chance of 5% and 0,02% extra per resistance point into applied school lvl 70 up to 10%._

**Chronology of changes in this mechanics:**
- _Patch 2.2.0 (25-Sep-2007): Duration against PvP targets has been reduced to 10 seconds_
- _Patch 2.0.3 (09-Jan-2007): There is no longer a random chance the effect of "Sap" will end (i.e. removed the heartbeat resist check). The effect will still end early when the target is damaged_
- _Patch 2.0.1 (05-Dec-2006): In PvP, Crowd Control effects will last no longer than 12 seconds instead of the full duration, with a chance of a heartbeat resist. Casting a lower rank polymorph on a target that is already under the effect of a higher rank will cause an error message and not consume mana_
- _Patch 1.4.0 (2005-05-05): Charm and Polymorph mechanics can now be reactively countered by spells and abilities (e.g. Druid Shapeshifting will now cancel Polymorph effects)_
- _Patch 1.2.0 (18-Dec-2004): Players now have an increasing chance to break free of the effect, such that it is unlikely the effect will last more than 15 seconds_

As a result, we can conclude that with the release of patch 2.0.1 Heartbeat resistance stopped working in PvP.

**Selected formula for the implementation of this mechanics (for 3.3.5 branch)**:

When `SpellSchoolMask = SPELL_SCHOOL_MASK_NORMAL`:
_break_chance = base_break_chance + 100 * (aura_max_duration - aura_duration) * (aura_max_duration - aura_duration) / (aura_max_duration * aura_max_duration)_

Otherwise:
_break_chance = base_break_chance + ((target_resistance /(target_level^1.441 * 0.10)) * 100)_

**When resistance was removed, formula changed (for master branch):**
> _Heartbeat Resist is the ability for a mob to randomly break free of CC. This chance is lowered based on how much +hit you have, but never 100%. This is why healers would be a last-resort CC, because they have 0 hit rating causing mobs to break free more often._
[Source](https://www.mmo-champion.com/threads/820741-The-complete-crowd-control-Guide)

[WowGamepedia](https://wow.gamepedia.com/Heartbeat_resist)
[...]

PR #25822, opened 2021-01-08
Core/Unit: Implemented unit-based internal heartbeat-timer

r4d1sh:
**Changes proposed:**

-  Implemented unit-based internal heartbeat-timer. I guess for WotLK it is exactly 5 seconds
-  Move SMSG_FLIGHT_SPLINE_SYNC and food/drink periodically animation to heartbeat system
-  Re-implemented PROC_FLAG_KILLED as PROC_FLAG_HEARTBEAT
-  Create AuraScript hook OnTargetHeartbeat and re-implement "Party Time!" by using it
-  Implemented heartbeat-periodic auras like "Poison Mushroom" (6727)

- [x] 3.3.5
- [ ] master

**Tests performed:**

Builded and tested on Win64

- [ ] There is a problem with heartbeat-periodic auras. The existing system assumes the same proc time of the aura for all targets, but heartbeat-periodic auras should proc at different times for different targets. Currently this PR don't support `GetTickNumber()`, `GetRemainingTicks()` and `GetTotalTicks()` methods for heartbeat-periodic auras.
- [x] TODO: Implement heartbeat resist mechanic based on #20206


src/server/game/Spells/Auras/SpellAuras.h
```diff
@@ -138,6 +138,8 @@ class TC_GAME_API Aura
         void UpdateOwner(uint32 diff, WorldObject* owner);
         void Update(uint32 diff, Unit* caster);
 
+        virtual void OnTargetHeartbeat(AuraApplication* aurApp) { }
```
Shauren, line 141:
```suggestion
        virtual void OnTargetHeartbeat(AuraApplication* /*aurApp*/) { }
```

Comment out the argument to silence compile warning

Aokromes:
this is the same as https://github.com/TrinityCore/TrinityCore/pull/22380 ?

src/server/game/Spells/SpellInfo.cpp
```diff
@@ -850,6 +850,8 @@ SpellInfo::SpellInfo(SpellEntry const* spellEntry)
     _auraState = AURA_STATE_NONE;
 
     _allowedMechanicMask = 0;
+
+    _isHeartbeatProcOrPeriodic = (ProcFlags & PROC_FLAG_HEARTBEAT) != 0;
```
Shauren, line 854: This seems unneccessary, you could just write `(ProcFlags & PROC_FLAG_HEARTBEAT) != 0` directly inside `SpellInfo::IsHeartbeatProcOrPeriodic`

Shauren:
Not same and that one is wrong

src/server/game/Spells/Auras/SpellAuras.cpp
```diff
@@ -2718,6 +2731,24 @@ void UnitAura::AddStaticApplication(Unit* target, uint8 effMask)
...
 }
 
+void UnitAura::OnTargetHeartbeat(AuraApplication* aurApp)
+{
+    // Eat/Drinking heartbeat animation; Need to find a way to move this in spell script
+    {
+        SpellSpecificType specific = GetSpellInfo()->GetSpellSpecific();
+        if (specific == SPELL_SPECIFIC_FOOD || specific == SPELL_SPECIFIC_FOOD_AND_DRINK)
+            aurApp->GetTarget()->SendPlaySpellVisual(SPELL_VISUAL_KIT_FOOD);
+        else if (specific == SPELL_SPECIFIC_DRINK)
```
Ovahlord, line 2741: you can remove the else while at it. You can play both animations if you eat and drink at the same time

r4d1sh: @Ovahlord These are mutually exclusive conditions. The spell cannot be both ("food" or "food_and_drink") and "drink". But I calling `OnTargetHeartbeat` for all applied auras. If you eat and drink at the same time `OnTargetHeartbeat` will be called for both auras and both animations will be sent

xvwyh: On Classic right now if you both eat and drink at the same time - both visuals are sent (I can hear both the munching and the drinking sounds at the same time).

But as the comment states, eventually this needs to be remade into two generic spellscripts and manually applied on all respective food and drink spells. Then there won't be a way to make them mutually exclusive, and both scripts will execute their Heartbeat hooks and send both visuals together.

xvwyh:
I wasn't sure why was it needed to allow the usual periodic effects to execute on heartbeat (the call to `PeriodicTick()`), but after examining the spells further... I'm left even more confused.

So there's a bunch of spells in the game that have periodic auras, but the period is 0. As I pointed out in https://github.com/TrinityCore/TrinityCore/pull/22380#issuecomment-499149176, 3.3.5 client contains code that changes the period to 5000 if it's 0. The code is still there in 4.3.4 at the very least (but 5200). So there's reason to think that 5000 is just a default fallback for when the designer forgot to set the period themselves.

However. All such **vanilla** spells also have the procflag applied to them, without any reason for it (it doesn't look like they need to do anything at heartbeats, be it scripted action or proc): 4312, 6727, 7288, 7291, 7331. Spells from **after vanilla** no longer have this procflag, but they still have 0 period, so they either weren't working at all, were ticking on every world update, or were falling back to 5000 and ticking once every 5 seconds: 30400, 34589, 37504, 42416, 43327, 52562, 57550, 64345, 65755.

I'd ignore post-vanilla spells, since they're not very related to this PR. But we're still left with 5 vanilla spells that *are* related, since they use the procflag. 3 out of 5 bear the name "(TEST)", so they aren't used in the game. 4312 doesn't seem to be used. 6727, however, IS used, and reportedly IS supposed to be ticking every 5 secs.

<details>
<summary>Story time</summary>
I went to check in my Classic combatlog if I had ever eaten the mushroom that applies aura 6727. And sure enough - I did. I allowed the debuff to stay on me for 10 secs before dispeling it, and during those 10 seconds it dealt no damage to me, even though it should've, at least once, if we're going with the assumption that the interval should be 5 secs. But it didn't.

I went to wowhead, and [found out](https://classic.wowhead.com/item=8526/violet-tragan#comments:id=4940:reply=1342176) that it's broken in Classic and doesn't deal any damage. According to the other comments - it was working during vanilla. But [according to the comments on the spell itself](https://www.wowhead.com/spell=6727/poison-mushroom#comments:id=360070) - it wasn't working during WotLK. Go figure.

Regardless, Blizzard actually fixed this spell on Classic just recently, in the Naxx update. And they fixed it by... removing the procflag and actually specifying 5000 period. So there you have it. This tech is so ancient that it's broken even on retail.
</details>

So because it's broken on retail - we have no way to verify how it's supposed to work. Whether it triggers from heartbeat ticks, or is it just a regular periodic aura with 5000 period. With the current PR it will be triggering on heartbeats. I don't personally like it, and I think it unnecessarily overcomplicates the code, but I have no hard evidence in support of either option... Best I can give you is description of 7291 where it's "hardcoded" into the description that the spell is supposed to deal a total of 60 healing over 30 secs, with each tick being 10, so that means there must be 6 ticks. If the spell is supposed to tick on heartbeats, that basically means that there will be 5 ticks most of the time, except for maybe some edge case when the aura matches the heartbeat timer perfectly. But this spell isn't used in the game, so it's hardly reliable.

Besides this, there is one real issue in this implementation: if the spell has both the procflag AND a proper periodic aura with non-zero period, then this PR will prevent the aura from ticking on its own, and it will only tick on heartbeats. This definitely needs to be fixed. The rest is up for debate.

src/server/game/Entities/Unit/Unit.h
```diff
@@ -40,6 +40,8 @@
 #define MAX_AGGRO_RESET_TIME 10 // in seconds
 #define MAX_AGGRO_RADIUS 45.0f  // yards
 
+#define SECONDS_PER_HEARTBEAT 5
```
Carbenium, line 43: Please use `std::chrono` types for new code.

src/server/game/Spells/Auras/SpellAuras.cpp
r4d1sh: Current PR code behavior is absolutly identical as if there was a generic spellscript for all food and drink spells with `OnTargetHeartbeat` hook. `UnitAura::OnTargetHeartbeat` will be called for all food and drinks auras applied to your character and if you eat and drink at the same time you get both visual effects. This `else-if` statement only needed to avoid unnecessary check. If the checked aura has `SPELL_SPECIFIC_FOOD` or `SPELL_SPECIFIC_FOOD_AND_DRINK` specific then it cannot have `SPELL_SPECIFIC_DRINK`.

r4d1sh: Moved to spellscript in 5a8285e

r4d1sh:
On retail WoW classic servers time of sending player's update to out of range group members coincides with heartbeat so I move it's code to `Unit::Heartbeat`.

Ovahlord:
Fine from my side. I was actually thinking about that just yesterday lol

xvwyh:
Just want to note that, while someone somewhere might benefit from having these hooks in PlayerScript and UnitAI, in reality they *should never be needed* to script any blizzlike content. To my knowledge, only spell scripts can react to heartbeat on retail. There are no action triggers (creature-record-wide scripting) for heartbeat anywhere up to 7.3.2, which was the last executable I checked.
<details>
<summary>g_actionTriggers</summary>

```
"- NONE -"
"OnReaction"
"OnEnterCombat"
"OnLeaveCombat"
"OnCombatTick"
"OnHealthRange"
"OnEnergyRange"
"OnDeath"
"OnSpell"
"OnKill"
"OnSpawn"
"OnEmote"
"OnMelee"
"OnInteract"
"OnCombatTrigger"
[^ Alpha]
"OnSpellCast"
"OnPickPocket"
"OnSkinned"
"OnCombatReturn"
"OnPathFailing"
"OnHealthRangeRandom"
"OnEnergyRangeRandom"
[^ 1.12.1]
"OnGeneralTrigger"
"OnDespawn"
"OnSpellFailed"
"OnCharmBreak"
"OnPassengerControlEnd"
"OnVehicleReturn"
"OnVehicleRide"
"OnVehicleAbandon"
[^ 2.4.3]
"OnSpellStart"
"OnAuraApplied"
"OnAuraRemoved"
"OnPassengerRide"
"OnPassengerAbandon"
"OnPassengerSpawn"
[^ 3.3.5]
[^ 4.3.4]
"OnLootLockReleased"
"OnChannelStart"
"OnChannelInterrupted"
"OnChannelFinished"
"OnHealthDepleted"
[^ 7.3.2]
```
</details>
Likewise, there are no events (spawngroup-wide scripting) for heartbeats either, but on that front I have no info past the alpha client.
<details>
<summary>g_eventTypes</summary>

```
"Spawn"
"General event (designer defined)"
"Time event"
"Quest event"
"Area trigger event"
```
</details>
Spell scripts, however, did have a heartbeat hook even in alpha.
<details>
<summary>g_spellScriptHooks/g_spellArgs/g_spellArgsNeeded</summary>

```
"Validate"                  "(caster, spellrec)"
"Precast"                   "(caster, cast, ...)"
"Targeted"                  "(caster, castinfo)"
"Cast"                      "(caster, castinfo)"
"Aura Applied"              "(unit, aura)"
"Aura Removed"              "(unit, aura)"
"Check Interrupt"           "(caster, cast, value)"
"Get Duration"              "(caster, spellrec)"
"Get Cast Time"             "(caster, spellrec)"
"Heartbeat"                 "(unit, aura, slot)"
"Take Combat Hit"           "(victim, attacker, spell, value, value2)"
"Deal Combat Hit"           "(victim, attacker, spell, value, value2)"
"Script Effect"             "(unit, caster, spell)"
"Take Spell Hit"            "(victim, attacker, spell, value, value2)"
```
</details>

Shauren:
I'll add that units aren't the only objects that have heartbeat, proof for that found in later expansion client: 
https://github.com/TrinityCore/TrinityCore/blob/468c406740ab4a1043328d30813b29a6d6122671/src/server/game/Entities/GameObject/GameObjectData.h#L464

xvwyh:
Can confirm for capture points. WorldState update is pitch perfect every multiple of 5.25 secs, which is the the interval of the heartbeat we're experiencing in Classic.
image

I have a feeling that ritual-type GOs also check the number of channelers on heartbeat, as it always felt very random when will the channeling end when summoning players in Classic. I'll try to confirm it.

Wolffenstein:
Capture Points indeed do update every ~5s.

A question though - is the heartbeat a global (realm) shared timer or is it a separate unique one for every object?

xvwyh:
It's a separate timer for every object, reasonably assumed to start when the objects spawns. Players who logged in at different times will have heartbeats at different offsets from one another.

illfated:
Even though I am not a real coder, this feels like one of the quite cool changes to the source in later years.

src/server/game/Entities/Object/Object.cpp
```diff
@@ -974,6 +974,18 @@ m_currMap(nullptr), m_InstanceId(0), m_phaseMask(PHASEMASK_NORMAL), m_notifyflag
...
     m_serverSideVisibility.SetValue(SERVERSIDE_VISIBILITY_GHOST, GHOST_VISIBILITY_ALIVE | GHOST_VISIBILITY_GHOST);
     m_serverSideVisibilityDetect.SetValue(SERVERSIDE_VISIBILITY_GHOST, GHOST_VISIBILITY_ALIVE);
+
+    m_heartBeatTimer.Reset(SECONDS_PER_HEARTBEAT);
+}
+
+void WorldObject::Update(uint32 time_diff)
+{
+    m_heartBeatTimer.Update(time_diff);
+    while (m_heartBeatTimer.Passed())
```
ccrs, line 984: while?

r4d1sh: @ccrs Yep. If since the last update we have missed one or more "heartbeats", we must handle them in the current update.

ccrs: interesting, can you elaborate on the reasoning for this? I cant seem to find it myself

r4d1sh: @ccrs Heartbeat should fire every 5 seconds. In addition, we need to make sure that if the server has been lagging for the last 10 seconds, Heartbeat will be called twice. We should not miss heartbeats due to lags, because then some heartbeat-periodic aura can proc fewer times than was intended. Or we'll miss something else important.

Timer of normal periodic auras works like this. You can see here:  https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L820

xvwyh: Imagine that we're about to execute the tick, but the server freezes for 7 seconds. Now we're two ticks behind on heartbeat, the first one was the one we were about to execute, and the second one was meant to come 5 secs later, but the server was frozen at that time. With the "if" approach we have:
```
// timer is at -7s
if (timer <= 0)
{
    tick();
    timer = 5s;
}
```
"if" block gets executed once, timer is now set to 5s. We ticked only once, even tho we should've ticked twice. The timer is now at 5s, instead of 3s, as it should've been after 7s of lag.
Now the "while" approach:
```
// timer is at -7s;
while (timer <= 0)
{
    tick();
    timer += 5s;
}
```
"while" block get executed the first time, timer is now at -2s, "while" block gets executed the second time, timer is now at 3s. We ticked twice as intended. Yes, these ticks are not spaced out in time as intended, but at least their NUMBER is consistent, and the state of timer after the lag is preserved so it doesn't get out of sync.

This is the idea behind using "while" for timer checks. Is it needed here? Not really. There are a lot of other places where it would be more useful, for example in periodic auras. There is just one problem with this approach: if the lag was CAUSED by the tick - it will only make things spiral out of control even further and snowball the lag to the point where it would just freeze the server completely. Using "if" sort of prevents this from escalating too much, but at the cost of consistency.

ccrs: I see thanks, for the explaination.

 I did miss the bellow statement
```c++
m_heartBeatTimer.Reset(m_heartBeatTimer.GetExpiry() + SECONDS_PER_HEARTBEAT);
```

neat, I'm now considering if this approach should be also applied to movement generators respective timers...

r4d1sh:
> Besides this, there is one real issue in this implementation: if the spell has both the procflag AND a proper periodic aura with non-zero period, then this PR will prevent the aura from ticking on its own, and it will only tick on heartbeats. This definitely needs to be fixed. The rest is up for debate.

Not sure about this point. I played WoW Classic today and found an interesting spell "Lifestone Regeneration" (https://classic.wowhead.com/spell=5707/lifestone-regeneration). This spell has both the heartbeat procflag (at least in WotLK) AND a proper periodic aura with 5000ms period.

As you can see, Wowhead trusts the period value from the game files and displays it in the tooltip: "_Equip: Restores 10 health every 5 sec._" (https://classic.wowhead.com/item=833/lifestone)

image

But in-game you can see a different value that equals 5.2 seconds. This value is approximately equal to the heartbeat period in WoW Classic.

image

This can serve as proof that the presence of a heartbeat flag is always of higher priority than an explicitly specified period in DBC.

Shauren:
If you own this item, you should look at combat log healing timestamps, I'm going to repeat this everywhere: DO NOT BLINDLY TRUST SPELL/ITEM TOOLTIPS

xvwyh:
Alas, judging by my combatlog, in all my time of playing Classic I never encountered a player with that item. But even if I did, I doubt we could gleam anything useful from it: if Violet Tragan (with procflag and 0 period) wasn't working on Classic at all until they "fixed" it by changing period to 5000, then I have no reason to believe that the procflag would suddenly be working for Lifestone. At best I expect it to just adhere to its periodic timer of 5 secs. If only we had actual Vanilla servers, not this modern engine garbage...

But even then...

> 2005/03/08 (Patch 1.2.4): i ate it and got a 30s poison effect which did **10 dmg every 3 or 5 seconds**.
> 2005/05/06 (Patch 1.4.2): For those of you who read the tooltip and are as weak minded as I am, it's something like **8 nature damage every 5 seconds** for 30 seconds.
> 2006/11/02 (Patch 1.12.1): Hmm... I ate one and it put a "Poison Mushroom" debuff on me, and I went green, but **no dmg from it**.
> 2006/11/28 (Patch 1.12.1): Eating this causes you to be affected by a weak poison dealing **10 damage per tick**, lasting 30 seconds.
> 2007/02/10 (Patch 2.0.6): As of February 2007, the debuff caused the green "poison" cloud debuff, but it dealt **no damage to my character**, who had no armor on, although she did have 452 armor from "ice armor". I repeatedly picked and ate the mushroom and never received any damage.
> 2008/08/24 (Patch 3.0.1): The effect does **nothing but turn you green**.
> 2019/12/04 (Classic): It may be buggy in Classic, since it **just turned me green**.

There are reports of it being broken as early as 1.12 Vanilla. Maybe we shouldn't bother implementing it in WotLK at all.

r4d1sh:
This spell has hide-in-combat-log attribute so we are unlikely to see something there.

Shauren:
UNIT_HEALTH lua events then?

xvwyh:
Update fields broadcasting has its own delay, which will make it much harder to figure out when exactly the health changes. We're talking a 200ms difference here.

Shauren:
Update fields yes, but health (and powers) are special, SMSG_HEALTH_UPDATE/SMSG_POWER_UPDATE

Wolffenstein:
I'm not sure about classic, but I believe there was no SMSG_HEALTH_UPDATE before WOTLK. Health was updated in update object batches every 400ms.

Shauren:
Classic is using a bfa client, its 100% there

Wolffenstein:
> 
> 
> Classic is using a bfa client, its 100% there

Yeah but they made some changes to mimic original batching and I don't know if they adjusted this as well so just wanted to mention it. If it's in classic sniffs then you can just ignore what I said.

xvwyh:
Classic may be using BFA client, but health/power prediction is purposefully disabled there. I dunno how it works on modern clients, but by looking into WotLK exe - that packet modified a field in CGUnit_C that's responsible for storing **predicted** health. So it's possible that even if those packets do get sent to the client - they're effectively useless. At the very least I can say that on Classic both UNIT_HEALTH and UNIT_HEALTH_FREQUENT are fired always simultaneously.

Wolffenstein:
> 
> 
> Classic may be using BFA client, but health/power prediction is purposefully disabled there. I dunno how it works on modern clients, but by looking into WotLK exe - that packet modifies a field in CGUnit_C that's responsible for storing **predicted** health. So it's possible that even if those packets do get sent to the client - they're effectively useless. At the very least I can say that on Classic both UNIT_HEALTH and UNIT_HEALTH_FREQUENT are fired always simultaneously.

This means they probably did it to mimic original vanilla/tbc behaviour where health was processed every 400ms along with update object - that's why addons like Instant Health existed. They probably needed it to keep consistency with vanilla, this comment should give you a bit more insight:

image

r4d1sh:
Default health regeneration stops in combat. With this, I was able to observe health regeneration from spell proc in combat with low-level NPC. Based on the fact that the proc times have not shifted relative to my five-second stopwatch over time, it can be assumed that the spell does indeed proc every 5000 ms.

Faq:
> 
> UPD. Timers with alert every 5.2 / 5.25 seconds quickly desync with actual proc timer. So I'm sure the spell tooltip is really lying. 😕

Saw Shauren mentioning several times in chat: "Don't trust tooltips".

Any chance some part of this could be merged or it will be one from those PRs what will be labeled with Closed-Abandoned?

offl:
There is an new difficulty flag added only in classic and named 'Use Fast Classic Heartbeat'

jackpoz:
@r4d1sh what is the status of this PR ?

SNiTEBoBy (approved):

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -557,6 +545,34 @@ void Unit::InterruptMovementBasedAuras()
         RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_MOVE);
 }
 
+void Unit::Heartbeat()
+{
+    SendFlightSplineSyncIfNeeded();
```
killerwife, line 550: I advise calling WorldObject::Heartbeat for future compatibility

r4d1sh:
> @r4d1sh what is the status of this PR ?

Maybe soon I will continue to work on this PR.

killerwife:
Good work, merging it to cmangos.

r4d1sh:
Just a side note: Seems like aura from spell "Brittle Bones" (ID: 32441) should proc on heartbeat and trigger spell "Rattled" (ID: 32437) with a certain probability. I don't know at all with what probability, but I guess that about 15%.

I think there are more similar spells. If description of aura says that something should happen periodically, but period is not specified, and the type of aura, and it's just SPELL_AURA_DUMMY, then most likely this is proc-on-heartbeat aura.

Aokromes:
Conflicting files
src/server/game/Scripting/ScriptMgr.h

if no answer before 1 april i will close as abandoned.

Ovahlord:
Just a headsup for documentation purposes. Most of this PR's implementation has been ported over into master branch in tiny steps.
WorldObject heartbeat: bf9cd88cebf752ce8614fbac8a7c6d1f0de7fee8
PartyUpdate heartbeat: 80d72a2c913657d3bbea538cfb842de3fa55dc50
Cyclic Spline Sync heartbeat: c2fedbb6cc89510002e9f56751e7f8427e306487
(New) Indoor/Outdoor aura check heartbeat: 5a169f6eae45a7578d6e312cdba9825f6fc24fa5

Food emote will remain generic core behavior and will not be moved to spell script like this PR suggested. The proc flag implementation will soon follow. I decided to skip the periodic aura calculation and spell script hook additions for now since the master branch db2 contain no actively used spells anymore that use any of that. So this remains a task for a possible 3.3.5 cherry pick in the future.

What might be tackled later as well is moving control zone gameobjects to heartbeats as well but that's out of scope for the time being

Aokromes:
It’s [279 commits](https://github.com/TrinityCore/TrinityCore/compare/cbabaa0f...3.3.5) behind (base commit: https://github.com/TrinityCore/TrinityCore/commit/cbabaa0f3a605ed8ca0742c1b4a2eb4f741a01a9)
Use the [web editor](https://github.com/TrinityCore/TrinityCore/pull/25822/conflicts) or the command line to resolve conflicts before continuing.

src/server/game/Entities/Unit/Unit.cpp
src/server/scripts/EasternKingdoms/Karazhan/karazhan.cpp

Aokromes:
src/server/game/Entities/Object/Object.h
src/server/game/Entities/Unit/Unit.h
src/server/game/Spells/SpellMgr.cpp
src/server/scripts/EasternKingdoms/Karazhan/karazhan.cpp

PR #25825, opened 2021-01-08
Core/Commands: Fix .go zonexy command

Robingadko:
**Changes proposed:**

-  Change formulae for calculation of x and y coordinates as the current ones seem to be calculating wrongly
-  Use inversion formulae of DB2Manager::Map2ZoneCoordinates()

- [ ] 3.3.5
- [x] master

**Issues addressed:**

None

**Tests performed:**

Built and tested in-game in such zones as Darkshore, Stonetalon Mountains, Southern Barrens, Krokuun, Zuldazar, Revendreth.

None

### DIFF
--- a/src/server/game/DataStores/DB2Stores.cpp
+++ b/src/server/game/DataStores/DB2Stores.cpp
@@ -3119,7 +3119,7 @@ bool DB2Manager::GetUiMapPosition(float x, float y, float z, int32 mapId, int32 
 
     DBCPosition2D uiPosition
     {
-        // x any y are swapped
+        // x and y are swapped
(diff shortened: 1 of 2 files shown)


illfated:
At first, I was thinking about suggesting that the PR title could start with "Scripts/Commands:" because `Commands/cs_go.cpp` is located within the directory `src/server/scripts/`, but since the second file `DataStores/DB2Stores.cpp` is belongs to the Core category (`src/server/game/` & similar), it can just stay as it is.

As a grammatist, I usually find it more precise to use the verb form "Fix" instead of "Fixed", because the infinite or present tense form describes something about to happen when the commit is applied, instead of having already been applied to the source. Then again, I guess that is more a question of semantics and taste, than to have any importance at all. No worries for you, the PR looks good to me.

PR #25832, opened 2021-01-09
Expose binaries built on Circle CI as docker images in the Circle CI artifacts

jackpoz:
**Changes proposed:**

- Expose binaries built on Circle CI as docker images in the Circle CI artifacts
  - The artifact is always called docker.tar.gz
  - It contains all the binaries (authserver, worldserver, extractors)
  - It can be imported in docker with ```docker load```, more info in the README.md included
  - The docker image is tagged with "github_useranme/github_repository-lowercasebranchname-commitsha1" and github_useranme/github_repository-lowercasebranchname-latest (not sure if the -latest tag works though)
- Include some operational instructions
- Change Circle CI pch job build settings

- [x] 3.3.5
- [ ] master

**Issues addressed:**

None, but now PRs can be tested without building locally the sources. The images shouldn't be used in production but only for testing/debugging

**Tests performed:**

Logged to character selection screen

- [x] ~~It would be useful to create 2 circle ci TC base images, one for the runtime and one for building TC, to reduce the image size~~
    - Circle CI base images are already quite big (400 MB) and the current size with binaries is 560 MB so it's not worth the effort
- [ ] It would be useful to push the images to dockerhub
  - [ ] Maybe this could be optional and only when DOCKERHUB_* credentials are set in circle ci
- [x] It would be nice if autoupdater worked
  - [x] Test importing TDB

### DIFF
--- a/contrib/Docker/README.md
+++ b/contrib/Docker/README.md
@@ -22,7 +22,22 @@ The instructions below expect basic knowledge about how to configure TrinityCore
         ```
         docker run --entrypoint=worldserver --volume=/host/path/to/configs:/etc/TrinityCore--volume=/var/run/mysqld/mysqld.sock:/var/run/mysqld/mysqld.sock -p=8085:8085 "image name" --config /etc/TrinityCore/worldserver.conf
         ```
-    Change the ports and other parameters as needed. Consult Docker documentation for additional details about possible configuration settings
+    Change the ports and other parameters as needed. Consult Docker documentation for additional details about possible configuration settings.
+
+## Content
+The image contains:
+- authserver
+- mapextractor
+- mmaps_generator
+- vmap4asembler
+- vmap4extractor
+- worldserver
+- README&#46;md
+
+You can explore the image using
+```
+docker run --entrypoint=/bin/bash -it "image name"
+```
 
 ## Limitations:
 - Database auto-updater: no instructions are given in this README to get the database auto-updater to work. Either apply the updates manually or investigate how to get it to work (most likely some additional --volume parameters should be passed)
--- a/contrib/Docker/README.md
+++ b/contrib/Docker/README.md
@@ -39,10 +39,9 @@ You can explore the image using
 ```
 docker run --entrypoint=/bin/bash -it "image name"
 ```
-
-You can export the logs from the container with
+Note that the WORKDIR is set to /home/circleci and by default all logs will be saved in that folder. You can export the logs from the container with
 ```
-docker cp "container name":name.log name.log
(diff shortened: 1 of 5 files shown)


jackpoz:
image

Autoupdater works (just need to test importing TDB)

illfated:
The source view for the file `contrib/Docker/Dockerfile` shows a missing NewLine at EOF. Is it safe to assume that it is not an important detail?

jackpoz:
I always get confused about which file extension complains when there is a new line (I think .yml doesn't always like it) but I will add it and check how it goes

jackpoz:
logs are created at /home/circleci by default and can be copied from the container to the host with commands like
```
docker cp "container name":/home/circleci/Server.log server.log
```
image

illfated:
Cool. I really hope those Docker fans will check out this feature.

jackpoz:
I will for sure use this next time I need to host a PR on the test realm as it makes it much easier (I don't have to build the sources anymore, I can just download the docker image, load it, start it up)

contrib/Docker/README.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+For more instructions please check docker official documentation.
+
+## Limitations:
+- Database connection: the instructions provided expect MySQL to run on the host machine. Change ```docker run``` parameters and .conf settings to fit your scenario
+- To import TDB using the autoupdater:
+  1. Download the TDB sql file from GitHub
+  1. ```chmod 666 TDB_name.sql``` in the host
+  1. Map it with ```--volume=/path/to/TDB_full_name.sql:/home/circleci/TDB_full_name.sql``` added to the commands specified in the main steps above
```
illfated, line 54: Based on my MarkDown experience, it is recommended to use only 1 pair of MD backticks for horizontal string encapsulation. I haven't noticed any rationale for using the triple pair backtick encapsulation other than for multi-line code block fencing.  Are there any documents or pages recommending that MD backticks should be used like this in certain circumstances?

jackpoz: I didn't even know 2 were enough, thanks for the tip

jackpoz:
@illfated could you please check if I missed any of the triple ` and maybe even suggest any other change so I can accept them ? thanks

illfated:
Sorry, I fell asleep around dinner time and forgot to check my GitHub feed until just now (binged YouTube for a few hours). There are a couple of double pairs of backticks in `contrib/Docker/README.md` which could be simplified. It is only a "readme" file, so I can easily edit & suggest the change to simplify the MD backticks now, if you like.

PR #25880, opened 2021-01-17
Core/Creature: Add new db field InteractionPauseTimer

### LINKED ISSUE #25647: Core/Waypoints: Waypoint stop timer after gossip hello
**Description:**

Waypoint stop timer after gossip hello is definitely wrong.
Treat as 20sec pause:

ClientToServer: CMSG_GOSSIP_HELLO (0x017B) Length: 8 ConnIdx: 0 Time: 10/06/2010 17:23:54.000 Number: 8278
GUID: Full: 0xF1300039E70042EC Type: Creature Entry: 14823 Low: 17132

ServerToClient: SMSG_ON_MONSTER_MOVE (0x00DD) Length: 49 ConnIdx: 0 Time: 10/06/2010 17:24:16.000 Number: 8521
GUID: Full: 0xF1300039E70042EC Type: Creature Entry: 14823 Low: 17132

Yeah, I know anyone can change that timer in config (Creature.MovingStopTimeForPlayer = 180000) but the thing is if my memory serves me well that timer can vary from creature to creature. It depends on something.

**Expected behaviour:**

Timer should be lowered and investigated if it really can vary or not.

**Branch(es):** both
**TC rev. hash/commit:** 05fa1a46b36f81c177a503f7f5de14eb34830208

jackpoz:
**Changes proposed:**

-  Add new db field InteractionPauseTimer to allow to define different interaction pause timers for each creature (the previous system was a global setting in worldserver.conf Creature.MovingStopTimeForPlayer)
- Make all the fields in creature_template_movement optional. The default value specified in C++ will be used if this table contains no row or NULL value

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #25647

**Tests performed:**

None

- [ ] Finish the PR


jackpoz:
This PR is a draft and is missing required changes that I have not done yet

src/server/game/Entities/Creature/CreatureData.h
```diff
@@ -114,14 +114,15 @@ enum class CreatureRandomMovementType : uint8
 struct TC_GAME_API CreatureMovementData
 {
     CreatureMovementData() : Ground(CreatureGroundMovementType::Run), Flight(CreatureFlightMovementType::None), Swim(true), Rooted(false), Chase(CreatureChaseMovementType::Run),
-        Random(CreatureRandomMovementType::Walk) { }
+        Random(CreatureRandomMovementType::Walk), InteractionPauseTimer(180000) { }
```
jackpoz, line 117: this default value might have to be changed

jackpoz:
Example:
```sql
INSERT INTO creature_template_movement (CreatureId, InteractionPauseTimer) VALUES (14508, 5000) ON DUPLICATE KEY UPDATE InteractionPauseTimer = 5000;
```
image

https://user-images.githubusercontent.com/1153754/105240584-650f5a80-5b6d-11eb-93d9-0c12f6a8e098.mp4

@offl how does this look ?

offl:
Should be 0

jackpoz:
In case it's 0 do you want to skip stopping the movement completely ? As PauseMovement() is called with force = true

jackpoz:
New example with 0:
```sql
INSERT INTO creature_template_movement (CreatureId, InteractionPauseTimer) VALUES (14508, 0) ON DUPLICATE KEY UPDATE InteractionPauseTimer = 0;
```

https://user-images.githubusercontent.com/1153754/105411594-e03b4400-5c33-11eb-8154-3130b47b1752.mp4

offl:
Good to go then

jackpoz:
I added support to InteractionPauseTimer to creature_movement_override too and changed the fields of creature_movement_override to allow NULL.

When creature_movement_override has a NULL column, creature_template_movement is checked and if it has a value for that column, that's the one that gets used. If creature_template_movement has a NULL column too or no row for the creature, then the default C++ value is used.

You can see an example here:
- creature_template_movement
```sql
INSERT INTO creature_template_movement (CreatureId, InteractionPauseTimer) VALUES (14508, 5000) ON DUPLICATE KEY UPDATE InteractionPauseTimer = 5000;
SELECT * FROM creature_template_movement WHERE CreatureId = 14508;
```
image
- creature_movement_override
```sql
INSERT INTO creature_movement_override (SpawnId, Chase) VALUES (637, 1) ON DUPLICATE KEY UPDATE Chase = 1;
SELECT SpawnId, Ground, Swim, Flight, Rooted, Chase, Random, InteractionPauseTimer FROM creature_movement_override WHERE spawnid = 636;
```
image
- result
image

This query can be used to retrieve the content of the 2 tables as loaded by worldserver
```sql
SELECT cmo.SpawnId,COALESCE(cmo.Ground, ctm.Ground),COALESCE(cmo.Swim, ctm.Swim),COALESCE(cmo.Flight, ctm.Flight),COALESCE(cmo.Rooted, ctm.Rooted),COALESCE(cmo.Chase, ctm.Chase),COALESCE(cmo.Random, ctm.Random),COALESCE(cmo.InteractionPauseTimer, ctm.InteractionPauseTimer)FROM creature_movement_override AS cmoLEFT JOIN creature AS c ON c.guid = cmo.SpawnIdLEFT JOIN creature_template_movement AS ctm ON ctm.CreatureId = c.id
```

This allows to specify in the sql updates only the columns that needs to be overridden, leaving the rest NULL as a clear "the default value is used for this". `INSERT ... ON DUPLICATE KEY UPDATE ...` means one doesn't even have to care if there is any existing row at all.

@offl what do you think about this ? we can talk about it on irc or discord if you have time. Of course if one prefers to always specify all values and never use NULL it's still possible.

offl:
I'm no longer interested in updating that table fully
Because I don't want to spent tens of hours in attempt to create one of the biggest update and at the same time spent energy for arguing with people who thinks that table has lot of potential

I don't know how it will be in practice, I used full statements will all columns in case of that table
I also avoid to use NULL almost always because to me they looks less readable especially if you have some columns with 0 and some with NULL
But it's me, some people may say they prefer NULLs

malcrom:
ick

src/server/game/Entities/Creature/CreatureData.h
Ovahlord: Use the worldserver config default value here instead. Making default handling more flexible
`sWorld->getIntConfig(CONFIG_CREATURE_STOP_FOR_PLAYER)`

malcrom:
Why not add it to the AI system to pause npc when gossip opens rather than add a field to db?

Ovahlord:
Because it's a movement meta data since it has nowhere been observed that this value changes during runtime. So if it's constant and never changes, there is no need to bloat the AI with more helpers that nobody uses.

PR #25887, opened 2021-01-18
Fix typo in SendAuthWaitQue method name

chipzz:
Que is misspelled :P


Shauren (approved):

illfated:
As Manuel in the TV series Fawlty Towers used to say when he did not understand something: - "¿Que?" 😄

PR #25911, opened 2021-01-22
[Do Not Merge] Move Argus to BrokenIsles - Discuss about it!

JustZerooo:
**Changes proposed:**

-  Move Argus to BrokenIsles
-  
-  

- [ ] 3.3.5
- [X] master

**Issues addressed:**

No Issues involved

**Tests performed:**

No Tests Needed

This PR serves to illustrate why Argus actually belongs to the broken islands and should not be treated as a separate folder. The principle here is actually quite simple. Blizzard has a category in its map system and expands it accordingly into sub-categories.

Here is an example for Sturmwin (World -> Azeroth -> Eastern Kingdoms -> Stormwind City)

As you can see, everything has a certain hierarchy. And after that, in my opinion, the System folder should also be kept here. That has nothing to do with whether you want to know better, but the aspect of how it is handled is decisive, not the sub-map in any DBC / DB2 file, this is not the decisive point here.

Argus ((World -> Azeroth -> Broken Isles -> Argus))


MaxtorCoder:
Useless.

Aokromes:
nope argus is separate planet. being show on same map doesn't makes same continent.
Also if you want discusion https://github.com/TrinityCore/TrinityCore/discussions

PR #25914, opened 2021-01-22
IsAllowingDeadTargets now checks effects targets as well

### LINKED ISSUE #25900: 9.0.2 - (Priest) Mass resurrection does not work
**Description:**

Mass resurrection does not resurrect players in a group, nor does it resurrect anyone at all.

The spell has 2 effects, namely:
SpellEffect - 173 (EffectResurrectWithAura) = implemented and working
SpellTarget - 119 (TARGET_CORPSE_SRC_AREA_RAID) = partially implemented, but does not work in the spell

**Expected behaviour:**

If the player is in a group and is dead, mass resurrection should prompt the dead player to respawn

1. Create any character as a priest class
2. Invite another ally player to the group
3. The invited player must die and not leave the body
4. Player priest must use Mass Resurrection

**Branch(es):**

master

**TC rev. hash/commit:** 

bb24c6a

**Operating system:** 

Windows 10 x64

matanshukry:
**Changes proposed:**

-  IsAllowingDeadTargets() now checks the `_effects` target and not just it's own target

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #25900 

**Tests performed:**

- [x] It builds
- [x] Tests in-game

### DIFF
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -1550,9 +1550,7 @@ bool SpellInfo::IsRequiringDeadTarget() const
 bool SpellInfo::IsAllowingDeadTarget() const
 {
     if (HasAttribute(SPELL_ATTR2_CAN_TARGET_DEAD) || Targets & (TARGET_FLAG_CORPSE_ALLY | TARGET_FLAG_CORPSE_ENEMY | TARGET_FLAG_UNIT_DEAD))
-    {
         return true;
-    }
 
     for (SpellEffectInfo const* effect : _effects)
     {
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -1549,7 +1549,24 @@ bool SpellInfo::IsRequiringDeadTarget() const
 
 bool SpellInfo::IsAllowingDeadTarget() const
 {
-    return HasAttribute(SPELL_ATTR2_CAN_TARGET_DEAD) || Targets & (TARGET_FLAG_CORPSE_ALLY | TARGET_FLAG_CORPSE_ENEMY | TARGET_FLAG_UNIT_DEAD);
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/SpellInfo.cpp
```diff
@@ -1549,7 +1549,24 @@ bool SpellInfo::IsRequiringDeadTarget() const
...
+    {
+        return true;
+    }
+
+    for (SpellEffectInfo const* effect : _effects)
+    {
+        if (!effect)
+            continue;
+
+        switch (effect->TargetA.GetTarget())
```
Shauren, line 1562: Replace the switch with `if (effect->TargetA.GetObjectType() == TARGET_OBJECT_TYPE_CORPSE)`
And add TargetB

Shauren (changes requested):

src/server/game/Spells/SpellInfo.cpp
shdr212: The reason I used a switch is because I saw few more enum values with `CORPSE` in them, but wasn't sure which I should include

Are you saying I shouldn't include anything else?

matanshukry: The reason I used a switch is because I saw few more enum values with CORPSE in them, but wasn't sure which I should include

Are you saying I shouldn't include anything else?

Shauren: Look closer, I'm suggesting replacing `GetTarget()` with `GetObjectType()`

matanshukry: @Shauren my bad.

Added TargetB and also using `GetObjectType()` now like you suggested.

Shauren (changes requested):
Good, code is fine, now time to take care of style

We don't use brackets for single line statement (exceptions apply in if/else constructs)

matanshukry:
> Good, code is fine, now time to take care of style
> 
> We don't use brackets for single line statement (exceptions apply in if/else constructs)

done

Shauren (approved):

PR #25920, opened 2021-01-23
Scripts/Commands: gm cmd modify gender: set correct customization

### LINKED ISSUE #25899: Changing gender breaks character model
After change character gender with command .modify gender character's model breaks.
Change gender with commands .character customize/changerace/changefaction don't work and leads to client creation error.

1. Join Game, 9.0.2 client
2. Type .mod gender
3. Logout to character select screen
4. See your character model breaks
5. Try to change gender to opposite in customization menu
6. Getting client creation error

**Branch(es):**

master

**Operating system:** 

Windows 10

matanshukry:
**Changes proposed:**

-  When using the gm command to modify gender, customization should also be changed since they are not compatible between different genders. This change will select the 1st customization from every option. 

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #25899

**Tests performed:**

- [x] it builds
- [x] tested in-game

### DIFF
--- a/src/server/scripts/Commands/cs_modify.cpp
+++ b/src/server/scripts/Commands/cs_modify.cpp
@@ -35,6 +35,8 @@ EndScriptData */
 #include "SpellMgr.h"
 #include "SpellPackets.h"
 #include "WorldSession.h"
+#include "CharacterCache.h"
+#include "UpdateFields.h"
 
 class modify_commandscript : public CommandScript
 {
@@ -891,6 +893,30 @@ public:
         // Change display ID
         target->InitDisplayIds();
 
+        target->RestoreDisplayId(false);
+        sCharacterCache->UpdateCharacterGender(target->GetGUID(), gender);
+
+        // Generate random customizations
+        std::vector<UF::ChrCustomizationChoice> customizations;
+
+        const std::vector<const ChrCustomizationOptionEntry*> * options = sDB2Manager.GetCustomiztionOptions(target->getRace(), gender);
+        for (const ChrCustomizationOptionEntry *option : *options)
+        {
+            std::vector<ChrCustomizationChoiceEntry const*> const* choicesForOption = sDB2Manager.GetCustomiztionChoices(option->ID);
+            // take the 1st option
+            if (choicesForOption->size() > 0)
+            {
+                const ChrCustomizationChoiceEntry* choiceEntry = choicesForOption->at(0);
+                UF::ChrCustomizationChoice choice;
+                choice.ChrCustomizationOptionID = option->ID;
+                choice.ChrCustomizationChoiceID = choiceEntry->ID;
+                customizations.push_back(choice);
+            }
+        }
+
+        target->SetCustomizations(Trinity::Containers::MakeIteratorPair(customizations.begin(), customizations.end()));
+
+        //
         char const* gender_full = gender ? "female" : "male";
 
         handler->PSendSysMessage(LANG_YOU_CHANGE_GENDER, handler->GetNameLink(target).c_str(), gender_full);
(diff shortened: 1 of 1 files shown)


illfated:
Your PR title is quite OK as it is. I just thought maybe it would be possible to incorporate the "Scripts/Commands" part in the title, if it would also be possible to shorten the existing title a little bit. The closest I could come up with on short notice is as follows: 
**Scripts/Commands: switch gender gm command random customization** 
(or modify it to add a better description for the actual source command change)

Note: this is only a suggestion, allowing for your own personal ideas and choices, not a request to change the title unless you are happy with changing it. I am only thinking what it will look like in the final merge and appearance in the commits list. Feel free to keep the title as it is, or pick any intermediate edit that you like.

Shauren:
@illfated does not matter. When someone clicks squash button on github, he can override commit message

illfated:
Fair point. Thank you.

matanshukry:
I still like the point @illfated said here; naming is important (and the most difficult job for a programmer).
done.

Shauren:
Does this work fine for all race/class/gender combinations? There are things like dependent choices - for example you can only pick choice A for option A if you selected choice B for option B. You can see the logic for that in `WorldSession::ValidateAppearance` and `WorldSession::MeetsChrCustomizationReq`. In case it doesn't validate properly, that character will be forced to recustomize on next login

src/server/scripts/Commands/cs_modify.cpp
```diff
@@ -891,6 +893,30 @@ class modify_commandscript : public CommandScript
         // Change display ID
         target->InitDisplayIds();
 
+        target->RestoreDisplayId(false);
+        sCharacterCache->UpdateCharacterGender(target->GetGUID(), gender);
+
+        // Generate random customizations
+        std::vector<UF::ChrCustomizationChoice> customizations;
+
+        const std::vector<const ChrCustomizationOptionEntry*> * options = sDB2Manager.GetCustomiztionOptions(target->getRace(), gender);
```
Shauren, line 902: [TYPE] [const][* or &][space][VARIABLE NAME]

matanshukry:
> Does this work fine for all race/class/gender combinations? There are things like dependent choices - for example you can only pick choice A for option A if you selected choice B for option B. You can see the logic for that in `WorldSession::ValidateAppearance` and `WorldSession::MeetsChrCustomizationReq`. In case it doesn't validate properly, that character will be forced to recustomize on next login

done

src/server/scripts/Commands/cs_modify.cpp
matanshukry: done

src/server/scripts/Commands/cs_modify.cpp
```diff
@@ -891,6 +893,41 @@ class modify_commandscript : public CommandScript
...
+
+            // Loop over the options until the first one fits
+            std::vector<ChrCustomizationChoiceEntry const*> const* choicesForOption = sDB2Manager.GetCustomiztionChoices(option->ID);
+            for (ChrCustomizationChoiceEntry const* choiceForOption : *choicesForOption)
+            {
+                ChrCustomizationReqEntry const* choiceReq = sChrCustomizationReqStore.LookupEntry(choiceForOption->ChrCustomizationReqID);
+                if (choiceReq && !worldSession->MeetsChrCustomizationReq(choiceReq, playerClass, false, MakeChrCustomizationChoiceRange(customizations)))
+                    continue;
+
+                const ChrCustomizationChoiceEntry* choiceEntry = choicesForOption->at(0);
```
Shauren, line 919: missed a const here

matanshukry: Fixed now.

Shauren (approved):

PR #25924, opened 2021-01-23
Core/Misc: Fix static analysis issues

jackpoz:
**Changes proposed:**

-  Core/Misc: Fix static analysis issues

- [x] 3.3.5
- [ ] master

**Issues addressed:**

None

**Tests performed:**

None

- [ ] Fix more issues

### DIFF
--- a/src/server/scripts/Commands/cs_debug.cpp
+++ b/src/server/scripts/Commands/cs_debug.cpp
@@ -274,43 +274,57 @@ public:
         {
             char commentToken[2];
             ifs.get(commentToken[0]);
-            if (commentToken[0] == '/' && !ifs.eof())
+            if (ifs.eof())
+                break;
+            if (commentToken[0] == '/')
             {
                 ifs.get(commentToken[1]);
-                // /* comment
-                if (commentToken[1] == '*')
+                if (!ifs.eof())
                 {
-                    while (!ifs.eof())
+                    // /* comment
+                    if (commentToken[1] == '*')
                     {
-                        ifs.get(commentToken[0]);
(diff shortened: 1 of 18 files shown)


src/server/scripts/Commands/cs_debug.cpp
```diff
@@ -274,44 +274,58 @@ class debug_commandscript : public CommandScript
...
             }
             parsedStream.put(commentToken[0]);
         }
         ifs.close();
 
-        uint32 opcode;
+        uint32 opcode = 0;
         parsedStream >> opcode;
 
+        if (opcode == 0)
```
jackpoz, line 326: Need to change this to !opcode

illfated:
Looks quite ordered and easy to follow.

I assume real coders don't find it difficult to read, but for me it is somewhat hard to follow the logic in the change in `cs_debug.cpp`. I presume the logic stems from a Coverity entry, and is therefore needed to avoid issues.

jackpoz:
all these fixes are from pvs-studio analysis at https://tcubuntu.northeurope.cloudapp.azure.com/pvsstudio/analysis/ . The cs_debug.cpp in particular is https://tcubuntu.northeurope.cloudapp.azure.com/pvsstudio/analysis/sources/cs_debug.cpp_16.html#ln273 , saying
>The 'ifs' stream is checked for EOF before reading from it, but is not checked after reading. Potential use of invalid data.

I was actually able to trigger an infinite loop in that code, now that's fixed

illfated:
> I was actually able to trigger an infinite loop in that code, now that's fixed

Progress! 😄

Issue #25870, opened 2021-01-17
tele name <player> npcid / npcguid <id/guid>


Aokromes:
**Description:**

tele name (player) npcid / npcguid (id/guid)
example:

tele name aokromes npcid 10000
or
tele name aokromes npcguid 1000

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

0b7b7f1

**Operating system:** 

N/R

illfated:
Wouldn't it be more appropriate to modify the `.appear` command to include the ID/GUID parameter?

```
appear  Syntax: .appear [$charactername]

        Teleport to the given character. 
        Either specify the character name or click on the character's portrait,
        e.g. when you are in a group. 
        Character can be offline.
```

Jinnaix:
What aokromes wants here is to teleport any player to a specific npc without first teleport to that creature as gm and then summon the player afterwards. Also being able to do this from console would be nice too.

illfated:
Oh. I see. Did not think about that scenario. Very well, that makes my comment above redundant.

Shauren:
.tele name npc id [player name] [creatureid]
.tele name npc id [player name] [guid]
.tele name npc id [player name] [creature name

PR #25933, opened 2021-01-24
 Scripts/Commands: implement teleport character to NPC

matanshukry:
**Changes proposed:**

-  Added gm command to teleport a character to an npc.

| Syntax | Example |
| ------- | -- |
| `.tele name2 <char_name> id <npc_id>` | `.tele name2 msh id 63847` |
| `.tele name2 <char_name> guid <npc_guid>` | `.tele name2 msh guid 63847` |
| `.tele name2 <char_name> name <npc_name>` | `.tele name2 msh id Incineratus` |

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Referenced #25870

**Tests performed:**

- [x] It builds
- [x] tested in-game


src/server/game/Chat/Chat.h
```diff
@@ -127,8 +129,10 @@ class TC_GAME_API ChatHandler
         GameTele const* extractGameTeleFromLink(char* text);
         bool GetPlayerGroupAndGUIDByName(const char* cname, Player*& player, Group*& group, ObjectGuid& guid, bool offline = false);
         std::string extractPlayerNameFromLink(char* text);
+        std::string extractPlayerNameFromLink(std::string const &text);
         // select by arg (name/link) or in-game selection online/offline player or self if a creature is selected
         bool extractPlayerTarget(char* args, Player** player, ObjectGuid* player_guid = nullptr, std::string* player_name = nullptr);
+        bool extractPlayerTarget(std::string const &nameStr, Player** player, ObjectGuid* player_guid = nullptr, std::string* player_name = nullptr);
```
Shauren, line 135: Don't touch this class to add std::string overloads (will be ported from 3.3.5 later)

src/server/scripts/Commands/cs_tele.cpp
```diff
@@ -35,6 +35,9 @@ EndScriptData */
 #include "Player.h"
 #include "RBAC.h"
 #include "WorldSession.h"
+#include <boost/tokenizer.hpp>
```
Shauren, line 38: We have a Tokenizer class in Util.h

src/server/scripts/Commands/cs_tele.cpp
```diff
@@ -343,9 +347,203 @@ class tele_commandscript : public CommandScript
...
+            return false;
+
+        // build query
+        std::ostringstream whereClause;
+        bool isValid = false;
+        switch (npcTargetType)
+        {
+            case 0: // id
+            {
+                uint32 npcId = std::stol(entry);
```
Shauren, line 427: Use `atoul`, std::stol can throw

src/server/scripts/Commands/cs_tele.cpp
```diff
@@ -343,9 +347,203 @@ class tele_commandscript : public CommandScript
...
+                if (isValid = !!npcId)
+                    whereClause << "WHERE id = '" << entry << '\'';
+                break;
+            }
+            case 1: // guid
+            {
+                ObjectGuid::LowType guidLow = std::stoll(entry);
```
Shauren, line 434: Use `atoull`

Shauren (changes requested):

src/server/scripts/Commands/cs_tele.cpp
```diff
@@ -343,9 +347,203 @@ class tele_commandscript : public CommandScript
...
+        }
+
+        if (!isValid)
+        {
+            handler->SendSysMessage(LANG_BAD_VALUE);
+            handler->SetSentErrorMessage(true);
+            return false;
+        }
+
+        QueryResult result = WorldDatabase.PQuery("SELECT position_x, position_y, position_z, orientation, map FROM creature %s", whereClause.str().c_str());
```
MaxtorCoder, line 458: What if the NPC has a waypoint and isn't on that position. Iirc there's a method to retrieve the Creature/Unit entity dynamically without having to query DB.

src/server/game/Chat/Chat.h
matanshukry: I didn't really understand what you mean here.
Is the problem with overloads? That is, if I rename the methods is that fine? or should I not create any methods with std::string in the header?

src/server/scripts/Commands/cs_tele.cpp
matanshukry: To be honest I pretty much just combined 2 existing gm commands, implementation-wise.
That is, the reason I used this it's because it used in the other gm command - `go creature`.

But if you know another way I'm willing to take a look?

src/server/scripts/Commands/cs_tele.cpp
matanshukry: Shauren why do we have it / what are the advantages of it?
from quickly I looking I can see one advantage for boost, which is it parses the data on demand as the tokens are accessed rather than the constructor, and it only support multiple characters.

src/server/scripts/Commands/cs_tele.cpp
MaxtorCoder: Tbf I cannot remember the method that required GUID/Entry MonkaW

illfated:
PR title suggestion: **Scripts/Commands: implement teleport character to NPC**
(I prefer the present tense instead of past tense, but that is just my personal preference)

jackpoz:
As the issue is about 3.3.5, could you please implement it on that branch ? Last PRs that I ported there had a lot of issues/code differences that required manual handling to be ported

matanshukry:
@jackpoz I can port it without actually testing it in-game, if that still works?

Jinnaix:
> @jackpoz I can port it without actually testing it in-game, if that still works?

You can ask for other people to check out your PR and check ingame functionality since you seem to not have a WotLK client / environment .

src/server/scripts/Commands/cs_tele.cpp
Shauren: Not using boost is the advantage - compile times don't suffer

Generally if we have code for it, it should be preferred over directly using 3rd party libraries

src/server/scripts/Commands/cs_tele.cpp
```diff
@@ -48,6 +51,7 @@ class tele_commandscript : public CommandScript
             { "add",   rbac::RBAC_PERM_COMMAND_TELE_ADD,   false, &HandleTeleAddCommand,   "" },
             { "del",   rbac::RBAC_PERM_COMMAND_TELE_DEL,    true, &HandleTeleDelCommand,   "" },
             { "name",  rbac::RBAC_PERM_COMMAND_TELE_NAME,   true, &HandleTeleNameCommand,  "" },
+            { "name2",  rbac::RBAC_PERM_COMMAND_TELE_NAME,   true, &HandleTele2Command,  "tele2 name <character> ((npcid <npcid>) | (npcguid <npcguid>))" },
```
Shauren, line 54: Help text should go to world.command table

src/server/game/Chat/Chat.cpp
```diff
@@ -964,15 +1000,75 @@ ObjectGuid::LowType ChatHandler::extractLowGuidFromLink(char* text, HighGuid& gu
 std::string ChatHandler::extractPlayerNameFromLink(char* text)
 {
     // |color|Hplayer:name|h[name]|h|r
-    char* name_str = extractKeyFromLink(text, "Hplayer");
-    if (!name_str)
+    std::string name_str = extractKeyFromLink(text, "Hplayer");
```
Shauren, line 1003: This here is still calling `char*     extractKeyFromLink(char* text, char const* linkType, char** something1 = nullptr);` which can return nullptr - assigning it to std::string will crash

src/server/game/Chat/Chat.cpp
```diff
@@ -964,15 +1000,75 @@ ObjectGuid::LowType ChatHandler::extractLowGuidFromLink(char* text, HighGuid& gu
...
 
-    return name;
+    return name_str;
+}
+
+std::string ChatHandler::extractPlayerNameFromLink(std::string const &text)
+{
+    // |color|Hplayer:name|h[name]|h|r
+    std::string name_str = extractKeyFromLink(text, "Hplayer");
+    normalizePlayerName(name_str);
```
Shauren, line 1017: Ignoring errors from `normalizePlayerName`

src/server/game/Chat/Chat.h
Shauren: This has to do with significant improvements (black-magic like) done to command argument handling done in the other branch (you would be adding code that will definitely be unused later)

Another thing are the issues I mentioned in other comments

(all of current, "old" handling is based on `strtok` and relies on having calls to  `strtok` done)

src/server/scripts/Commands/cs_tele.cpp
Shauren: not ideal but "fine", if its a copy of logic from other commands

src/server/game/Chat/Chat.h
matanshukry: Right I saw all the strtok handling, I thought making some new methods with std::string and tokenizer is gonna make it better.

I'm still not sure what I can use though; can I just move my methods into the .cpp file and make them non-member methods? is that fine?

src/server/game/Chat/Chat.cpp
matanshukry: Yeah I shouldn't of touched that function, that was a mistake.

src/server/scripts/Commands/cs_tele.cpp
matanshukry: I was talking about parsing that's done in runtime, not compile time.

And I disagree with that; boost isn't just any 3rd party library, it's widely known and used. I'd even argue that if there are any bugs in the code, it's more likely in ours than theirs, since their product is more widely used and tested.
We're just reinventing the wheel here.

But either way I removed it.

illfated:
Maybe
```diff
  #include "Language.h"
  #include "MapManager.h"
  #include "MotionMaster.h"
+ #include "ObjectAccessor.h"
  #include "ObjectMgr.h"
  #include "PhasingHandler.h"
  #include "Player.h"
```
in `cs_tele.cpp` ?

matanshukry:
@illfated are you saying it because of the result from circleci?
I was surprised to see that so I regenerated the proejct with cmake, cleaned all, rebuilt all. "It works on my machine".

I'll wait for the others CI to see what they show, but would love if someone else on master branch can pull and build this as well.

illfated:
> @illfated are you saying it because of the result from circleci?

Yes, it is about the result from ci/circleci: **nopch** (nopch build). The pch builds will complete without issues, I presume.

matanshukry:
@illfated not completely sure why it's not showing up in the pch version, but like you said that include completed those errors when building with no_pch.
I'll say ahead locally I had other errors as well (on no_pch only), but I didn't see thme on the circleci so I ignored them. I'll fixthen  if they come up too.

illfated:
I don't have the necessary experience to explain the coding difference between pch and no_pch, I only know that "pch" means 'pre-compiled headers' and that it refers to compiling with or without PCH.

matanshukry:
@illfated I know what it means, I just don't know why one would want to do 2 build versions, one without and with.

illfated:
Fair enough. The TC members usually know stuff like that, especially those who handle a lot of PRs. I am thinking more like "why not, if you have the option to check for multiple scenarios".

Shauren:
Committed to 3.3.5 d2be584f99a63a165cfd1d047f460a8ebfb87ac3

PR #25934, opened 2021-01-24
Core/Player: remaining XP overflow correction

matanshukry:
**Changes proposed:**

- Fix an issue with exp overflow when the experience required for next level is less than the experience you have (how is that even possible, I'm not really sure)

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [] tested in-game
   - I don't know how to get to a situation where the player has more experience than required for that level. Even at the max level the required experience is somewhat high, and the experience is set to 0.

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -1353,7 +1353,7 @@ void WorldSession::HandlePlayerLogin(LoginQueryHolder* holder)
 
     TC_LOG_INFO("entities.player.character", "Account: %u (IP: %s) Login Character: [%s] (%s) Level: %d, XP: %u/%u (%u left)",
         GetAccountId(), GetRemoteAddress().c_str(), pCurrChar->GetName().c_str(), pCurrChar->GetGUID().ToString().c_str(), pCurrChar->getLevel(),
-        _player->GetXP(), _player->GetXPForNextLevel(), _player->GetXPForNextLevel() - _player->GetXP());
+        _player->GetXP(), _player->GetXPForNextLevel(), std::max(0, (int)_player->GetXPForNextLevel() - (int)_player->GetXP()));
 
     if (!pCurrChar->IsStandState() && !pCurrChar->HasUnitState(UNIT_STATE_STUNNED))
         pCurrChar->SetStandState(UNIT_STAND_STATE_STAND);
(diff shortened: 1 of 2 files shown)


matanshukry:
@Aokromes comments? specifically, how do you get a character to have more xp than xp required?

Aokromes:
have level 80 (on 3.3.5a branch)

matanshukry:
Doesn't work on master. Not 60 (max level) nor 80.

Aokromes:
well, i did the comment on 3.3.5a branch, not on master branch.

matanshukry:
@Aokromes I wasn't complaining about you or anything like that, just wrote that to clarify I can't fully test it on master.

illfated:
title suggestion: **Core/Player: remaining XP overflow correction**
(or what you prefer, just to follow title codestyle)

Aokromes:
> @Aokromes I wasn't complaining about you or anything like that, just wrote that to clarify I can't fully test it on master.

i mean, i only have seen the bug on 3.3.5a branch, idk on master branch because i don't use it, every level 80 char shows that huge amount of xp left.

src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -1353,7 +1353,7 @@ void WorldSession::HandlePlayerLogin(LoginQueryHolder* holder)
 
     TC_LOG_INFO("entities.player.character", "Account: %u (IP: %s) Login Character: [%s] (%s) Level: %d, XP: %u/%u (%u left)",
         GetAccountId(), GetRemoteAddress().c_str(), pCurrChar->GetName().c_str(), pCurrChar->GetGUID().ToString().c_str(), pCurrChar->getLevel(),
-        _player->GetXP(), _player->GetXPForNextLevel(), _player->GetXPForNextLevel() - _player->GetXP());
+        _player->GetXP(), _player->GetXPForNextLevel(), std::max(0, (int)_player->GetXPForNextLevel() - (int)_player->GetXP()));
```
Shauren, line 1356: cast to int32 (never use raw built in int types)

Aokromes:
Level: 80, XP: 57387/0 (4294909909 left)
 Level: 80, XP: 632/0 (4294966664 left)
 Level: 80, XP: 901/0 (4294966395 left)

jackpoz:
@matanshukry on 335 GetXPForNextLevel() returns 0 when maxed, that's why the difference is negative. In a way checking if GetXPForNextLevel() is not 0 should be enough, something like
```cpp
_player->GetXPForNextLevel() ? _player->GetXPForNextLevel() - _player->GetXP() : 0
```

Shauren:
@jackpoz this version works for all branches and is best to keep that way to have same code everywhere possible

(as for why it behaves differently on master - xp per level is now found in client files and is provided for all levels, even beyond the cap)

Shauren (approved):

PR #25938, opened 2021-01-24
Core/Quests: automatically add quests with AUTO_PUSH flag to player

matanshukry:
**Changes proposed:**

-  automatically push quests with relevant flag on player login

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tests in game

- [ ] Currently a bug: need to filter the quests and somehow and not push them all. Possibly by zone/area id.


src/server/game/Server/Packets/QuestPackets.cpp
```diff
@@ -746,5 +746,6 @@ void ChoiceResponse::Read()
     _worldPacket >> ChoiceID;
     _worldPacket >> ResponseID;
 }
+
```
Aokromes, line 749: unneeded change

src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -1365,6 +1365,8 @@ void WorldSession::HandlePlayerLogin(LoginQueryHolder* holder)
 
     sScriptMgr->OnPlayerLogin(pCurrChar, firstLogin);
 
+
+
```
Aokromes, line 1369: unneeded change.

illfated:
PR title suggestion: **Core/Quests: push quests automatically with AUTO_PUSH flag**

matanshukry:
Thanks @mdX7 for helping me figuring out how to push it to the player!

which looks a lot simpler now that we know what the solution is

src/server/game/Quests/QuestDef.h
```diff
@@ -531,6 +531,8 @@ class TC_GAME_API Quest
         uint32 CalculateHonorGain(uint8 level) const;
         bool   CanIncreaseRewardedQuestCounters() const;
 
+        bool IsAutoPush() const { return HasFlagEx(QuestFlagsEx::QUEST_FLAGS_EX_AUTO_PUSH); }
```
Shauren, line 534: `QuestFlagsEx` is a regular enum, not enum class, so don't add `QuestFlagsEx::`

src/server/game/Entities/Player/Player.cpp
```diff
@@ -18742,9 +18742,34 @@ bool Player::LoadFromDB(ObjectGuid guid, CharacterDatabaseQueryHolder* holder)
...
     return true;
 }
 
+void Player::_PushQuests()
+{
+    ObjectMgr::QuestMap const& questMap = sObjectMgr->GetQuestTemplatesAutoPush();
+
+    for (ObjectMgr::QuestMap::const_iterator it = questMap.cbegin(); it != questMap.cend(); ++it)
+    {
+        QuestStatusMap::const_iterator result = m_QuestStatus.find(it->first);
```
Shauren, line 18756: All of the checks (including quest status) should be replaced with just a simple function call `CanTakeQuest(it->second, false)`

src/server/game/Globals/ObjectMgr.h
```diff
@@ -1701,6 +1702,7 @@ class TC_GAME_API ObjectMgr
         std::map<HighGuid, std::unique_ptr<ObjectGuidGeneratorBase>> _guidGenerators;
 
         QuestMap _questTemplates;
+        QuestMap _questTemplatesAutoPush;
```
Shauren, line 1705: You can just turn this into a `std::vector<Quest const*>`

src/server/game/Entities/Player/Player.h
```diff
@@ -2574,6 +2574,8 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         /***                    QUEST SYSTEM                   ***/
         /*********************************************************/
 
+        void _PushQuests();
```
Shauren, line 2577: remove `_` from function name

src/server/game/Entities/Player/Player.cpp
```diff
@@ -18742,9 +18742,34 @@ bool Player::LoadFromDB(ObjectGuid guid, CharacterDatabaseQueryHolder* holder)
 
     m_achievementMgr->CheckAllAchievementCriteria(this);
     m_questObjectiveCriteriaMgr->CheckAllQuestObjectiveCriteria(this);
+
+    _PushQuests();
```
Shauren, line 18746: Any other cases when quests are pushed to players? Levelup? Zone change?

Shauren (changes requested):

src/server/game/Entities/Player/Player.h
matanshukry: sure. I noticed other quests that are starting with _, I thought it's a notion for private methods.
When should one use _ for method names then?

src/server/game/Entities/Player/Player.cpp
matanshukry: I added zone change before I saw this, I forgoet levelup though. I'll do that too.

src/server/game/Entities/Player/Player.h
illfated: Usually it is reserved for class member variables, but looking at the start of this file, I presume there is an exception of some description for the utterly few methods starting with underscore.

illfated:
Good job so far.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -18756,9 +18759,37 @@ bool Player::LoadFromDB(ObjectGuid guid, CharacterDatabaseQueryHolder* holder)
...
     return true;
 }
 
+void Player::PushQuests()
+{
+    std::vector<Quest const*> const& questsAutoPush = sObjectMgr->GetQuestTemplatesAutoPush();
+
+    for (Quest const *quest : questsAutoPush)
+    {
+        QuestStatusMap::const_iterator result = m_QuestStatus.find(quest->GetQuestId());
```
Shauren, line 18773: Don't check `m_QuestStatus` here, `CanTakeQuest` already does that for you

Shauren (changes requested):
I have investigated the quest list that have this flag and here is what I found:
* All of them are candidates for pushing, regardless of the zone you are in (remove that condition)
* Almost all of them are quest types currently unsupported (emissary quests)
* The condition to filter out unsupported quests is `!quest->GetQuestTag() || quest->GetQuestTag() == QuestTagType::Tag` (GetQuestTag() returns Optional)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -18756,9 +18759,26 @@ bool Player::LoadFromDB(ObjectGuid guid, CharacterDatabaseQueryHolder* holder)
...
+    std::vector<Quest const*> const& questsAutoPush = sObjectMgr->GetQuestTemplatesAutoPush();
+
+    for (Quest const *quest : questsAutoPush)
+    {
+        //
+        if (quest->GetQuestTag() && quest->GetQuestTag() != QuestTagType::Tag)
+            continue;
+
+        if (!quest->IsUnavailable() && CanTakeQuest(quest, false))
+            AddQuest(quest, nullptr);
```
Shauren, line 18778: Shouldn't this be `AddQuestAndCheckCompletion(quest, nullptr);`?

matanshukry: Usually I would say no, but I noticed the SL quest progress is kept (both tc and on retail), so yes. Either way it's probably safer to use it than not.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -18756,9 +18759,26 @@ bool Player::LoadFromDB(ObjectGuid guid, CharacterDatabaseQueryHolder* holder)
...
+    {
+        //
+        if (quest->GetQuestTag() && quest->GetQuestTag() != QuestTagType::Tag)
+            continue;
+
+        if (!quest->IsUnavailable() && CanTakeQuest(quest, false))
+            AddQuestAndCheckCompletion(quest, nullptr);
+    }
+}
+
```
Shauren, line 18781:
```suggestion
void Player::PushQuests()
{
    for (Quest const* quest : sObjectMgr->GetQuestTemplatesAutoPush())
    {
        if (quest->GetQuestTag() && quest->GetQuestTag() != QuestTagType::Tag)
            continue;

        if (!quest->IsUnavailable() && CanTakeQuest(quest, false))
            AddQuestAndCheckCompletion(quest, nullptr);
    }
}

```

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -4214,6 +4215,9 @@ void ObjectMgr::LoadQuests()
 
         Quest* newQuest = new Quest(fields);
         _questTemplates[newQuest->GetQuestId()] = newQuest;
+
+        if (newQuest->IsAutoPush())
+            _questTemplatesAutoPush.push_back(newQuest);
```
Shauren, line 4220:
```suggestion
        if (newQuest->IsAutoPush())
            _questTemplatesAutoPush.push_back(newQuest);
```

src/server/game/Entities/Player/Player.cpp
```diff
@@ -18756,9 +18759,26 @@ bool Player::LoadFromDB(ObjectGuid guid, CharacterDatabaseQueryHolder* holder)
...
     return true;
 }
 
+void Player::PushQuests()
+{
+    std::vector<Quest const*> const& questsAutoPush = sObjectMgr->GetQuestTemplatesAutoPush();
+
+    for (Quest const *quest : questsAutoPush)
+    {
+        //
```
Shauren, line 18773:
```suggestion
void Player::PushQuests()
{
    for (Quest const* quest : sObjectMgr->GetQuestTemplatesAutoPush())
    {
```

Shauren (approved):

PR #25944, opened 2021-01-25
Core/Creature: Implement a way to respawn creatures with random time

gonzo1247:
…Creatures. Now we can say a creature should respawn between 1 Hour and 5 Hour and the core generate a random time between this two given times.

**Changes proposed:**

-  Some creatures (mostly rare) has not a fixed respawn time after x Hour. The has a random respawn time between two different times. Like between 2 Hours and 7 Hours. With the new colum 'spawntimesecmax' it's possible to add a Max spawn time and allow  the core to generate a random respawn time between spawntimesecs and spawntimesecmax.
-  If the colum 'spawntimesecmax' zero (0) that not used and core use only 'spawntimesecs'
-  Idk if this the best way to handle this because i has no idea if blizz handle this. If they use specific steps for create the respawn time or so. 

- [x] master

**Issues addressed:**

**Tests performed:**

Build works and ingame it's testet. If 'spawntimesecmax' in creature table set with a higher value as spawntimesecs the core generate a random time and use this for the respawn.

- [ ] 
- [ ]


offl:
I see [here](https://www.gamersglobal.de/sites/gamersglobal.de/files/userupload/user131/wowedit_1024.jpg) `Respawn Overrides` with
`Repop Min`
d h m s
`Repop Max`
d h m s

So retail uses no spawntimesecs or spawntimesecsmax for each spawn and respawn time is two different constants in continents and dungeons + raids, let's say 5 minutes for continents and 2 hours for dungeons. Or something like that
For us this means we indeed need spawntimesecsmax

gonzo1247:
Ok thats looks in Interesting.
For us we can change the spawntimesecs and spawntimesecmax to a Time type that the respawn time can set by HH:MM::SS with a range of 838:38:38 (whats like nearly 35 days). That should enough for the creatures be.
In the core that can recalculate to seconds and use the exist function for calculate the respawn time. 

or we let it as seconds in the colum - here should other say what the think is better...

offl:
> but overall we need this change, even as seconds in colum or with the Time type

Sure, Blizz uses random respawn time and we knew it even without that screenshot

Shauren:
This can't work - you only read `spawntimesecmax` value from Creature::SaveToDb which is executed from .npc add (and a bunch of other gm commands) but never on a "normal" path

illfated:
Please rename the SQL (currently named `2021_01_25_01_world.sql`) in this PR. That name is already taken.
Use a generic unused file name (using an impossible date), typically `9999_99_99_00_world.sql`.

gonzo1247:
okey i will rename the name of sql.
And i will change the rest - i test it indeed only with new spawn.

gonzo1247:
@Shauren i look and i read `spawntimesecmax`from `Creature::SaveToDB()`
I add a spawn into the `creature` table with this time: `spawntimesecs` = `7200` and `spawntimesecmax` = '28800`
after killing the creature this unix time `1611655271` was set to the respawn table. `Realzeit: 26.01.2021 - 11:01:11`
Thats looks for me that its works for spawned creature.
I think for spawns with .gm commands that needs more changes, like spawntimesecs and spawntimesecmax set via the .npc add command or so.

Shauren:
I said that this doesn't work for normal spawns - the ones you test with gm commands work.

gonzo1247:
I add a creature to `creature` table via sql - no .gm commands or other stuff, based with a fresh guid. Or is a "normal" spawn now in other table as in the `creature` table in the world database?

Shauren:
Doesn't work, closing

PR #25971, opened 2021-01-28
Core/AreaTriggers Implementation for server side area triggers

matanshukry:
**Changes proposed:**

-  Implemented server side area triggers. 
-  Incluedes new table `areatrigger_srv_position` to position the area trigger on the server

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] tested in-game

**Sql Implementatino**

```sql

ALTER TABLE `areatrigger_template` 
DROP PRIMARY KEY,
ADD COLUMN `IsServer` tinyint(1) NOT NULL AFTER `Id`,
ADD PRIMARY KEY (`Id`, `IsServer`);

ALTER TABLE `areatrigger_template_actions` 
DROP PRIMARY KEY,
ADD COLUMN `IsServer` tinyint(1) NOT NULL AFTER `AreaTriggerId`,
ADD PRIMARY KEY (`AreaTriggerId`, `IsServer`);

CREATE TABLE `areatrigger_positions` (
  `SpawnId` int(12) unsigned NOT NULL,	
  `Id` int(10) unsigned NOT NULL,
  `IsServer` tinyint(1) NOT NULL,
  `MapId` int(10) NOT NULL,
  `PosX` float NOT NULL,
  `PosY` float NOT NULL,
  `PosZ` float NOT NULL,
  `PhaseId` int(10) DEFAULT 0,
  `PhaseGroup` int(10) DEFAULT 0,
  `PhaseUseFlags` tinyint(3) DEFAULT 0,
  `Comment` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`SpawnId`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8;

```

**Example of content**

- Add the following scripts to have the mage tower works, inside and outside (should not be added, will create a different issue for it)

```sql
DELETE FROM `areatrigger_positions` WHERE `SpawnId` IN (1, 2);
INSERT INTO `areatrigger_positions` (`SpawnId`, `Id`, `IsServer`, `MapId`, `PosX`, `PosY`, `PosZ`, `PhaseId`, `PhaseGroup`, `PhaseUseFlags`, `Comment`) VALUES
  (1, 1, 1, 0, -9015.774, 877.223, 148.61871, 0, 0, 0, 'Stormwind Mage Portal Entrance'),
  (2, 2, 1, 0, -8999.866, 864.13995, 65.88978, 0, 0, 0, 'Stormwind Mage Portal Exit');

DELETE FROM `areatrigger_template` WHERE `Id` in (1, 2) AND `IsServer` = 1;
INSERT INTO `areatrigger_template` (`Id`, `IsServer`, `Type`, `Flags`, `Data0`, `Data1`, `Data2`, `Data3`, `Data4`, `Data5`, `ScriptName`, `VerifiedBuild`) VALUES
  (1, 1, 1, 0, 1.2597655, 1.2792665, 3.7021635, 0, 0, 0, '', 0),
  (2, 1, 1, 0, 1.2597655, 1.2792665, 3.7021635, 0, 0, 0, '', 0);


DELETE FROM `areatrigger_template_actions` WHERE `AreaTriggerId` IN (1, 2) AND `IsServer` = 1;
INSERT INTO `areatrigger_template_actions` (`AreaTriggerId`, `IsServer`, `ActionType`, `ActionParam`, `TargetType`) VALUES
  (1, 1, 2, 3631, 5),
  (2, 1, 2, 3630, 5);


UPDATE `world_safe_locs` SET LocX=-9014.864258, LocY=874.324890, LocZ=148.618713 WHERE `id` = 3630;
```
Note: the update at the end is due to conflict with positions. Discussed below

- Example of usage with smart script:
```sql
DELETE FROM `areatrigger_template` WHERE `Id` in (1, 2) AND `IsServer` = 1;
INSERT INTO `areatrigger_template` (`Id`, `IsServer`, `Type`, `Flags`, `Data0`, `Data1`, `Data2`, `Data3`, `Data4`, `Data5`, `ScriptName`, `VerifiedBuild`) VALUES
  (2, 1, 1, 0, 1.2597655, 1.2792665, 3.7021635, 0, 0, 0, 'SmartTriggerServer', 0);
  

DELETE FROM `smart_scripts` WHERE `entryorguid` = 2 AND `source_type` = 11;
INSERT INTO `smart_scripts` (`entryorguid`,`source_type`,`id`,`link`,`event_type`,`event_phase_mask`,`event_chance`,`event_flags`,`event_param1`,`event_param2`,`event_param3`,`event_param4`,`action_type`,`action_param1`,`action_param2`,`action_param3`,`action_param4`,`action_param5`,`action_param6`,`target_type`,`target_param1`,`target_param2`,`target_param3`,`target_x`,`target_y`,`target_z`,`target_o`,`comment`) VALUES
(2, 11, 0, 0, 46, 0, 100, 0, 0, 0, 0, 0, 62, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, -9016.31, 876.244, 148.619, 314.621, "On areatriggerserver 1 - teleport to outside of mage sw tower");
```
Note: the sql script above assume there is not areatrigger template with id 2, like there is in the script above. Hence don't use blindly together.

- [ ] I'm not sure why but multiple times I got to a state where `Update()` wasn't called on `AreaTriggerServer`, which I usually just ended up restarting. I'll try to reproduce it, but if someone knows why would that happen feel free to share.
- [ ] Currently if you have overlapping areas you could wind up in an infinite loop (ats 1 teleport to location A, location A is where ats 2, ats 2 teleport to location B, location B is where ats 1 is at). We can try to handle it by not triggering the AreaTrigger on teleport (if possible, didn't check yet). Otherwise, we can simply ignore this and verify before we put loop coordinates.


MaxtorCoder:
PLEAAAAASE do not create a new class for serverside ATs, it's completely useless and best to keep everything in 1 impl (so merge the changes into the AT class)

Also best to name the db table `area_trigger_positions` and not have abbreviations in table names.

mdX7:
> Currently if you have overlapping areas you could wind up in an infinite loop (ats 1 teleport to location A, location A is where ats 2, ats 2 teleport to location B, location B is where ats 1 is at). We can try to handle it by not triggering the AreaTrigger on teleport (if possible, didn't check yet). Otherwise, we can simply ignore this and verify before we put loop coordinates.
This case should never happen Blizzard is using a nice trick to prevent this: you're never actually teleported on the at itself, but a few yards offset.

matanshukry:
> PLEAAAAASE do not create a new class for serverside ATs, it's completely useless and best to keep everything in 1 impl (so merge the changes into the AT class)
> 
> Also best to name the db table `area_trigger_positions` and not have abbreviations in table names.

1. I thought it would be better to separate it at first, but looking at it again you might be write. I'll do that
2. the current areatrigger table don't have _ between area and trigger. What do you think about `areatrigger_positions` ?

> > Currently if you have overlapping areas you could wind up in an infinite loop (ats 1 teleport to location A, location A is where ats 2, ats 2 teleport to location B, location B is where ats 1 is at). We can try to handle it by not triggering the AreaTrigger on teleport (if possible, didn't check yet). Otherwise, we can simply ignore this and verify before we put loop coordinates.
> 
> This case should never happen Blizzard is using a nice trick to prevent this: you're never actually teleported on the at itself, but a few yards offset (hardcoded coords for each teleport).
> Random example: when entering an dungeon you're not teleported directly into the portal inside the dungeon. Same thing when going outside the dungeon.
So the 2nd option I presented, cool.

> 
> Also we should not use a new class, yes. That way your first issue should also be gone.
>
Agreed like above.
 
> Also i think we do not need that serverside boolean in `areatrigger_template_actions` since it references an at in _template and its already flagged there. Same thing for `areatrigger_srv_position`.
>
it's not referenced there, it's part of the primary key now. The reason for that is you could create server side ats with ids that in the future will be overwritten by data from retail, so you need to distinguish between data we create, and data we get.
At which point you still need `is server side` boolean in `areatrigger_template_actions` and `areatrigger_srv_position`.

> That position table should also not be explicitly be named for serverside triggers, just in case we want (whyever) spawn an areatrigger which is indeed visible to the player and spawned w/o a spell.
I'm not sure if that's applicable. If the area trigger is visible then we can get data from retail about it, and it should have another implementation. But sure I guess it's fine, we can just call it `areatrigger_positions`

MaxtorCoder:
> the current areatrigger table don't have _ between area and trigger. What do you think about areatrigger_positions ?

That's better, yes.

matanshukry:
Updated with single entity and changed the table names. A lot better and a lot less changes now.
Review again please.

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -212,11 +213,60 @@ AreaTrigger* AreaTrigger::CreateAreaTrigger(uint32 spellMiscId, Unit* caster, Un
     return at;
 }
 
+bool AreaTrigger::CreateServer(Map* map, AreaTriggerTemplate const* areaTriggerTemplate, AreaTriggerServerPosition const& position)
```
MaxtorCoder, line 216: Maybe an idea to merge this functionality in `AreaTrigger::Create`? Add an extra bool as argument and then use the position passed into that function as server position.

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -212,11 +213,60 @@ AreaTrigger* AreaTrigger::CreateAreaTrigger(uint32 spellMiscId, Unit* caster, Un
...
+    {
+        return false;
+    }
+
+    return true;
+}
+
+AreaTrigger* AreaTrigger::CreateAreaTriggerServer(Map* map, AreaTriggerTemplate const* areaTriggerTemplate, AreaTriggerServerPosition const& position)
+{
+    AreaTrigger*at = new AreaTrigger();
```
MaxtorCoder, line 250:
```suggestion
    AreaTrigger* at = new AreaTrigger();
```

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
matanshukry: This is the current argument list of the `AreaTrigger::Create` function:
`uint32 spellMiscId, Unit* caster, Unit* target, SpellInfo const* spell, Position const& pos, int32 duration, SpellCastVisual spellVisual, ObjectGuid const& castId, AuraEffect const* aurEff`

The only one that can be provided is the position, and even with just that and the bool info would be lacking since it's getting a `AreaTriggerMiscTemplate` from the spell, and using that to get the script Id, which is irrelevvant with the server side scripts.

MaxtorCoder: Hmmm, okay, we'll see what Shauren or any other contributer has to say heh :D

illfated:
Cosmetic & codestyle: PR title adjustment suggestion, specifically typo in `sever`, but also codestyle of placing a colon `:` after the keywords (**Core/AreaTriggers:**)

**Core/AreaTriggers: server side area triggers implementation**

illfated:
Regarding **nopch** build: maybe `#include Battleground.h` in `AreaTrigger.cpp` ?

matanshukry:
@illfated why `Battleground.h` ?
Anyway I added `ObjectMgr.h` and it solved it locally.

I will also say I'm seeing these errors locally:
```
2>D:\dev\tc\src\server\game\Movement\MovementGenerator.h(69,1): error C2440: 'static_cast': cannot convert from 'Unit *' to 'T *'
2>        with
2>        [
2>            T=Player
2>        ] (compiling source file D:\dev\tc\src\server\game\AI\CreatureAIRegistry.cpp)
2>D:\dev\tc\src\server\game\Movement\MovementGenerator.h(69,54): message : Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or function-style cast (compiling source file D:\dev\tc\src\server\game\AI\CreatureAIRegistry.cpp)
2>D:\dev\tc\src\server\game\Movement\MovementGenerator.h(68): message : while compiling class template member function 'bool MovementGeneratorMedium<Player,FlightPathMovementGenerator>::Update(Unit *,uint32)' (compiling source file D:\dev\tc\src\server\game\AI\CreatureAIRegistry.cpp)
2>D:\dev\tc\src\server\game\Movement\MovementGenerators\WaypointMovementGenerator.h(101): message : see reference to class template instantiation 'MovementGeneratorMedium<Player,FlightPathMovementGenerator>' being compiled (compiling source file D:\dev\tc\src\server\game\AI\CreatureAIRegistry.cpp)
2>D:\dev\tc\src\server\game\Movement\MovementGenerator.h(69,1): error C2440: 'static_cast': cannot convert from 'Unit *' to 'T *'
2>        with
2>        [
2>            T=Creature
2>        ] (compiling source file D:\dev\tc\src\server\game\Handlers\MovementHandler.cpp)
2>D:\dev\tc\src\server\game\Movement\MovementGenerator.h(69,54): message : Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or function-style cast (compiling source file D:\dev\tc\src\server\game\Handlers\MovementHandler.cpp)
2>D:\dev\tc\src\server\game\Movement\MovementGenerator.h(68): message : while compiling class template member function 'bool MovementGeneratorMedium<Creature,WaypointMovementGenerator<Creature>>::Update(Unit *,uint32)' (compiling source file D:\dev\tc\src\server\game\Handlers\MovementHandler.cpp)
2>D:\dev\tc\src\server\game\Movement\MovementGenerators\WaypointMovementGenerator.h(54): message : see reference to class template instantiation 'MovementGeneratorMedium<Creature,WaypointMovementGenerator<Creature>>' being compiled (compiling source file D:\dev\tc\src\server\game\Handlers\MovementHandler.cpp)
2>Done building project "game.vcxproj" -- FAILED.

```

However they do'nt up on the CI or on pch builds, so I haven't touched them. I'll handle them if the CI will report it, of course.

illfated:
> @illfated why `Battleground.h` ?
> Anyway I added `ObjectMgr.h` and it solved it locally.

Very well, my mistake (hence the question, I was uncertain).

matanshukry:
> > @illfated why `Battleground.h` ?
> > Anyway I added `ObjectMgr.h` and it solved it locally.
> 
> Very well, my mistake (hence the question, I was uncertain).
No worries - thanks for trying to help.

> 
> > I will also say I'm seeing these errors locally:
> 
> Huh. Not what I would expect, as long as the build checks here on GitHub don't complain. Sorry I don't know your set of errors. Real coders may be able to see what happened on your side.
> 
> edit: Oh. I forgot to take one thing into consideration: are you sure that your local source is just as up to date as the GitHub TC source?

Yeah. I'm updating it to master every few hours. But yeah as long asi t works on GitHub..

illfated:
> Yeah. I'm updating it to master every few hours. But yeah as long as it works on GitHub..

Oh well. Unless incorrect versions of the local dependency software caused those issues, I don't know. I am clueless this time.

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
```diff
@@ -71,6 +72,16 @@ enum AreaTriggerActionUserTypes
     AREATRIGGER_ACTION_USER_MAX    = 6
 };
 
+struct AreaTriggerServerPosition
```
Shauren, line 75: Would `AreaTriggerSpawn` be a better name?

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
```diff
@@ -71,6 +72,16 @@ enum AreaTriggerActionUserTypes
...
+{
+    uint32 Id;
+    bool IsServer;
```
Shauren, line 78: You should replace these two with either direct pointer to AreaTriggerTemplate or encapsulate using AreaTriggerTemplateKey

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -1670,6 +1670,15 @@ bool ScriptMgr::OnAreaTrigger(Player* player, AreaTriggerEntry const* trigger, b
...
 
+bool ScriptMgr::OnAreaTriggerServer(Player* player, AreaTriggerTemplate const* triggerTemplate, bool entered)
+{
+    ASSERT(player);
+    ASSERT(triggerTemplate);
+
+    GET_SCRIPT_RET(AreaTriggerScript, triggerTemplate->ScriptId, tmpscript, false);
+    return tmpscript->OnTriggerServer(player, triggerTemplate, entered);
+}
+
```
Shauren, line 1681: ScriptMgr changes are unneccessary, scripts can use existing `AreaTriggerAI` by just giving it the scriptname in db

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
```diff
@@ -71,6 +72,16 @@ enum AreaTriggerActionUserTypes
...
+    WorldLocation Location;
```
Shauren, line 79: A few more fields are needed here (and in database)
* uint64 SpawnId; // guid, PK (to support multiple spawns with the same template/script - imaginary scenario: a few creatures with areatrigger around each of them doing the same thing for "go near one of them" type of script)
* uint8 PhaseUseFlags; // phasing requirement - for example to only do thing when you are phased on a quest
* uint32 PhaseId;
* uint32 PhaseGroup;

src/server/game/AI/SmartScripts/SmartScript.h
```diff
@@ -122,6 +123,7 @@ class TC_GAME_API SmartScript
         GameObject* go;
         ObjectGuid goOrigGUID;
         AreaTriggerEntry const* trigger;
+        AreaTriggerTemplate const* areaTriggerTemplate;
```
Shauren, line 126: Don't support SAI for now, that would require `class AreaTrigger` to support SAI in the first place (you would need two new SMART_SCRIPT_TYPE_)

src/server/game/Globals/AreaTriggerDataStore.cpp
```diff
@@ -16,17 +16,48 @@
...
+            Trinity::hash_combine(hashVal, value.IsServer);
+            return hashVal;
+        }
+    };
+}
+
 namespace
 {
-    std::unordered_map<uint32, AreaTriggerTemplate> _areaTriggerTemplateStore;
+    std::unordered_map<uint32, std::vector<AreaTriggerServerPosition>> _areaTriggersByMap;
```
Shauren, line 59: This will become `std::unordered_map<ObjectGuid::LowType, AreaTriggerSpawn> _areaTriggerSpawns;` (see comments on AreaTriggerServerPosition)

src/server/game/Globals/AreaTriggerDataStore.cpp
```diff
@@ -131,22 +173,73 @@ void AreaTriggerDataStore::LoadAreaTriggerTemplates()
...
+                    position.Id, position.IsServer);
+                continue;
+            }
+
+            // Add each area trigger to the map
+            auto triggerIter = _areaTriggersByMap.emplace(mapId, std::vector<AreaTriggerServerPosition>()).first;
+            triggerIter->second.push_back(position);
+
+            // call OnCreateMap to create the trigger if the map already exist
+            Map* map = sMapMgr->FindMap(mapId, 0);
```
Shauren, line 237: Do not access map here (linked to other suggested changes)

src/server/game/Globals/AreaTriggerDataStore.cpp
```diff
@@ -272,3 +365,27 @@ AreaTriggerDataStore* AreaTriggerDataStore::Instance()
     static AreaTriggerDataStore instance;
     return &instance;
 }
+
+void AreaTriggerDataStore::OnCreateMap(class Map* map)
```
Shauren, line 369: You cannot simply preload all objects on map creation, this is not resistant to grid unloads.

You must add guids to a new container in CellObjectGuids (ObjectMgr.h), see `ObjectMgr::AddCreatureToGrid`/`ObjectMgr::RemoveCreatureFromGrid` for an example of what I'm talking about. Then work with `ObjectGridLoader` to get your areatriggers to spawn (api is predefined, must be `AreaTrigger::LoadFromDB(ObjectGuid::LowType spawnId, Map* map, bool addToMap, bool allowDuplicate)`

Shauren (changes requested):

src/server/game/AI/SmartScripts/SmartScript.h
matanshukry: but it does?
That is, I tested SAI and it works with the server side impelmentation as I currently did it?

Shauren: It worked but you had id conflict with old db2 based areatriggers

matanshukry: > It worked but you had id conflict with old db2 based areatriggers
What do you mean exactly by that?
I mean yes the old area trigger would still trigger, but I'm pretty sure they had no effectss.
I use a new source type for the SmartAI, and only the new area trigger i've created in the db (id=2, IsServer=1) has "SmartTriggerServer" on it's script name.

matanshukry: I took a 2nd look and now I understand what you're saying.

I implemented all the thing that are required for SAI, but removed them in one of the commits. You can see them here:
https://github.com/TrinityCore/TrinityCore/pull/25971/commits/e7764cd341788ee4746b9ebe32a7f049c734833b

I can bring it back it shouldn't be much problem, but if you prefer we can leave it for another issue.

Shauren: Up to you if you want to support it fully or not at all, just don't leave this partially broken code

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -212,11 +218,60 @@ AreaTrigger* AreaTrigger::CreateAreaTrigger(uint32 spellMiscId, Unit* caster, Un
...
+
+    Object::_Create(ObjectGuid::Create<HighGuid::AreaTrigger>(GetMapId(),
+        areaTriggerTemplate->Id, GetMap()->GenerateLowGuid<HighGuid::AreaTrigger>()));
+
+    SetEntry(areaTriggerTemplate->Id);
+
+    SetObjectScale(1.0f);
+
+    if (position.PhaseId || position.PhaseGroup || position.PhaseUseFlags)
+        if (!PhasingHandler::InDbPhaseShift(this, position.PhaseUseFlags, position.PhaseId, position.PhaseGroup))
```
Shauren, line 254: Typo here, should be `PhasingHandler::InitDbPhaseShift`

src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
```diff
@@ -71,6 +72,19 @@ enum AreaTriggerActionUserTypes
     AREATRIGGER_ACTION_USER_MAX    = 6
 };
 
+struct AreaTriggerServerPosition
+{
+    uint32 SpawnId;
```
Shauren, line 77: Reason I suggested uint64 here is for consistency with other types of spawns

src/server/game/Globals/AreaTriggerDataStore.h
```diff
@@ -19,15 +19,20 @@
...
 class AreaTriggerTemplate;
 class AreaTriggerMiscTemplate;
+struct AreaTriggerServerPosition;
 
 class TC_GAME_API AreaTriggerDataStore
 {
 public:
     void LoadAreaTriggerTemplates();
 
+    CellGuidSet const* GetAreaTriggersForMapAndCell(uint32 mapId, uint32 cellId) const;
```
Shauren, line 33: If you are not storing this in CellObjectGuids then just use std::set<ObjectGuid::LowType> directly, not with typedef (tl;dr don't include ObjectMgr.h in headers)

src/server/game/AI/SmartScripts/SmartScript.h
matanshukry: I'm gonna leave tihs out of this PR and I've created a new ticket to add SAI to this when it's done

illfated:
(ref. #25972)

PR #25988, opened 2021-01-29
Core/Auras: Implement SPELL_AURA_397 & SPELL_AURA_398

### LINKED ISSUE #24794: Spell aura 397 & 398
**Description:**

`SPELL_AURA_397` and `SPELL_AURA_398` are not implemented. They appear to be linked to `SMSG_BATTLEGROUND_PLAYER_POSITIONS` as their miscvalue is linked to  the arena slot information.

Right now we handle `SMSG_BATTLEGROUND_PLAYER_POSITIONS` in the battleground class but it's not linked to the aura's arena slot value

I'm not sure what's the difference between these two auras.

**Expected behaviour:**

We should name them correctly and use their miscvalue to fill in the arenaslot.

1. Check spellwork

**Branch(es):**

- master
- maybe 3.3.5 ?

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/1cdd1d0249ad49c860e117e1e39d451e1a3fbe43

**Operating system:** win 10

Golrag:
**Changes proposed:**

-  Define SPELL_AURA_397 & SPELL_AURA_398. More info can be found in #24794
-  Remove GetPlayerPositionData virtual function in Battleground class
-  Save the players in a vector in the base battleground class.

-  3.3.5 maybe
- master

**Issues addressed:**

Closes #24794

**Tests performed:**
- Builds
- Went in game, queued for Eye of the Storm (specific), cast 34976 and then with debugger went through everything. Also removed the buff by clicking on it and went through the debugger again. Also checked map if the flag was showing up, it did.

- Names might not be very blizzlike
- Some coding standards?

**Notes:**
It's been a while.
I went for a vector because the order of insertion is being kept in the packets. (first player who takes flag is the first one in the position).
Unsure about making a new struct for it... Maybe I should have just used the one from the WorldPackets and updated the position each time. Not sure if that's something thats acceptable tho (in case of updating packets etc might get a little more complex)

Edit: Using the packet class makes things a lot cleaner tho

Any feedback is welcome :)

### DIFF
--- a/src/server/game/Battlegrounds/Battleground.cpp
+++ b/src/server/game/Battlegrounds/Battleground.cpp
@@ -248,7 +248,20 @@ void Battleground::_ProcessPlayerPositionBroadcast(uint32 diff)
         m_LastPlayerPositionBroadcast = 0;
 
         WorldPackets::Battleground::BattlegroundPlayerPositions playerPositions;
-        GetPlayerPositionData(&playerPositions.FlagCarriers);
+
+        for (BattlegroundPlayerPositionSlotInfo const& info : _playerPositionInfo)
+        {
+            if (Player* player = ObjectAccessor::GetPlayer(GetBgMap(), info.Guid))
+            {
+                WorldPackets::Battleground::BattlegroundPlayerPosition position;
+                position.Guid = player->GetGUID();
+                position.Pos = player->GetPosition();
+                position.IconID = info.IconId;
+                position.ArenaSlot = info.ArenaSlot;
+                playerPositions.FlagCarriers.push_back(position);
+            }
+        }
+
         SendPacketToAll(playerPositions.Write());
     }
 }
--- a/src/server/game/Battlegrounds/Battleground.cpp
+++ b/src/server/game/Battlegrounds/Battleground.cpp
@@ -249,17 +249,13 @@ void Battleground::_ProcessPlayerPositionBroadcast(uint32 diff)
 
         WorldPackets::Battleground::BattlegroundPlayerPositions playerPositions;
 
-        for (BattlegroundPlayerPositionSlotInfo const& info : _playerPositionInfo)
+        for (WorldPackets::Battleground::BattlegroundPlayerPosition& playerPosition : _playerPositionInfo)
         {
-            if (Player* player = ObjectAccessor::GetPlayer(GetBgMap(), info.Guid))
-            {
-                WorldPackets::Battleground::BattlegroundPlayerPosition position;
-                position.Guid = player->GetGUID();
-                position.Pos = player->GetPosition();
-                position.IconID = info.IconId;
-                position.ArenaSlot = info.ArenaSlot;
-                playerPositions.FlagCarriers.push_back(position);
-            }
+            // Update position data if we found player.
+            if (Player* player = ObjectAccessor::GetPlayer(GetBgMap(), playerPosition.Guid))
+                playerPosition.Pos = player->GetPosition();
+
+            playerPositions.FlagCarriers.push_back(playerPosition);
         }
 
         SendPacketToAll(playerPositions.Write());
--- a/src/server/game/Battlegrounds/Battleground.cpp
+++ b/src/server/game/Battlegrounds/Battleground.cpp
@@ -138,8 +138,6 @@ Battleground::~Battleground()
 
     for (BattlegroundScoreMap::const_iterator itr = PlayerScores.begin(); itr != PlayerScores.end(); ++itr)
         delete itr->second;
-
-    _playerPositions.clear();
 }
 
 void Battleground::Update(uint32 diff)
@@ -251,13 +249,13 @@ void Battleground::_ProcessPlayerPositionBroadcast(uint32 diff)
 
         WorldPackets::Battleground::BattlegroundPlayerPositions playerPositions;
 
-        for (std::shared_ptr<WorldPackets::Battleground::BattlegroundPlayerPosition> position : _playerPositions)
+        for (WorldPackets::Battleground::BattlegroundPlayerPosition& playerPosition : _playerPositions)
         {
             // Update position data if we found player.
-            if (Player* player = ObjectAccessor::GetPlayer(GetBgMap(), position->Guid))
-                position->Pos = player->GetPosition();
+            if (Player* player = ObjectAccessor::GetPlayer(GetBgMap(), playerPosition.Guid))
+                playerPosition.Pos = player->GetPosition();
 
-            playerPositions.FlagCarriers.push_back(*position);
+            playerPositions.FlagCarriers.push_back(playerPosition);
         }
 
         SendPacketToAll(playerPositions.Write());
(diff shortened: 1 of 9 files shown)


illfated:
Looks interesting. Should be interesting to test for anyone who uses TC master branch and also enjoys Battlegrounds.

illfated:
Curious error message in AppVeyor (warning treated as error): `'WorldPackets::Battleground::BattlegroundPlayerPosition *const ': unknown size`

Golrag:
Dont have that error locally, so might be a little tricky to resolve

Golrag:
Well.. something I could try is to include the BattlegroundPackets.h in Battleground.h but I would like to avoid that. Anyone has other suggestions?

src/server/game/Battlegrounds/Battleground.cpp
```diff
@@ -1669,6 +1678,21 @@ void Battleground::PSendMessageToAll(uint32 entry, ChatMsg msgType, Player const
...
 }
 
+void Battleground::AddPlayerPosition(WorldPackets::Battleground::BattlegroundPlayerPosition const position)
+{
+    _playerPositions.push_back(position);
+}
+
+void Battleground::RemovePlayerPosition(ObjectGuid guid)
+{
+    auto const& itr = std::remove_if(_playerPositions.begin(), _playerPositions.end(), [guid](WorldPackets::Battleground::BattlegroundPlayerPosition const info)
```
Shauren, line 1688: Do not const& this, the type here is `std::vector<WorldPackets::Battleground::BattlegroundPlayerPosition>::iterator` - you are making a reference to temporary (gets destroyed after `;`)

Mesielecat:
I think that the 398 the misc value is directly the iconId, if you check the 121164 Orb of Power or 121164 Orb of Power or 121176 Orb of Power or 121177 Orb of Power, this are out of the arena slot

Mesielecat:
Reviewing sniff the value of arenaSlot is the same that the misc value, then this arenaSlot is confuse and need update the name of the variable, maybe flagSlot

Golrag:
In my early bfa sniff I can see that temple of kot mogu always has IconID 1 or 2 and that the misc value is the ArenaSlot. IconID depends on faction and reverses it with the corresponding aura effect. I wrote that down in the issue I linked

Golrag:
To fix the build on appveyor i could try having a vector of pointers or unique_ptrs, will play around with it tomorrow

illfated:
> To fix the build on appveyor i could try having a vector of pointers or unique_ptrs, will play around with it tomorrow

Seems like your changes did the trick. Thank you, well done. 👍🏻

Shauren:
Why shared_ptr?

Golrag:
Had some compile issues with a unique_ptr :/ can try to look at it again or go with just a pointer

src/server/game/Spells/Auras/SpellAuraDefines.h
```diff
@@ -469,8 +469,8 @@ enum AuraType : uint32
     SPELL_AURA_SHOW_CONFIRMATION_PROMPT                     = 394,
     SPELL_AURA_AREA_TRIGGER                                 = 395,  // NYI
     SPELL_AURA_TRIGGER_SPELL_ON_POWER_AMOUNT                = 396,  // NYI Triggers spell when health goes above (MiscA = 0) or falls below (MiscA = 1) specified percent value (once, not every time condition has meet)
-    SPELL_AURA_397                                          = 397,
-    SPELL_AURA_398                                          = 398,
+    SPELL_AURA_UPDATE_BATTLEGROUND_PLAYER_POSITION          = 397,
+    SPELL_AURA_UPDATE_BATTLEGROUND_PLAYER_POSITION_2        = 398,
```
Shauren, line 473: Proposed names
SPELL_AURA_BATTLEGROUND_PLAYER_POSITION_FACTIONAL = 397 (inspired by BATTLEMASTER_LIST_FLAG_FACTIONAL)
SPELL_AURA_BATTLEGROUND_PLAYER_POSITION = 398

Golrag: done!

Golrag:
The BattlegroundMgr is using the copy constructor for Battlegrounds in CreateNewBattleground. I think that's why I'm unable to have a vector of unique_ptr's?
```
error C2280: 'std::unique_ptr<WorldPackets::Battleground::BattlegroundPlayerPosition,std::default_delete<WorldPackets::Battleground::BattlegroundPlayerPosition>>::unique_ptr(const std::unique_ptr<WorldPackets::Battleground::BattlegroundPlayerPosition,std::default_delete<WorldPackets::Battleground::BattlegroundPlayerPosition>> &)': attempting to reference a deleted function (compiling source file H:\Trinity\source\src\server\game\Battlegrounds\BattlegroundMgr.cpp)
````

**Comment by Shauren** [comment]
Convert it back to value and also:
* Declare copy constructor in Battleground class
* Create "implementation" in Battleground.cpp `Battleground::Battleground(Battleground const&) = default;`

**Comment by Shauren** [comment]
`_Copy_assign` errors in appveyor, add `Battleground& operator=(Battleground const&) = delete;`?

**Comment by Golrag** [comment]
That seems to work :)

File: src/server/game/Battlegrounds/Battleground.cpp [hunk 3]
```diff
@@ -1669,6 +1684,21 @@ void Battleground::PSendMessageToAll(uint32 entry, ChatMsg msgType, Player const
...
 }
 
+void Battleground::AddPlayerPosition(WorldPackets::Battleground::BattlegroundPlayerPosition const position)
+{
+    _playerPositions.push_back(position);
+}
+
+void Battleground::RemovePlayerPosition(ObjectGuid guid)
+{
+    auto itr = std::remove_if(_playerPositions.begin(), _playerPositions.end(), [guid](WorldPackets::Battleground::BattlegroundPlayerPosition const info)
```
Shauren, line 1694: add reference to lambda argument type

src/server/game/Battlegrounds/Battleground.cpp
```diff
@@ -1669,6 +1684,21 @@ void Battleground::PSendMessageToAll(uint32 entry, ChatMsg msgType, Player const
     va_end(ap);
 }
 
+void Battleground::AddPlayerPosition(WorldPackets::Battleground::BattlegroundPlayerPosition const position)
```
Shauren, line 1687: add reference to arg

Golrag:
Warsong Scramble is up now and I wonder how it behaves on retail. Does the last one that picked up the flags that is put in that arenaslot? Does it update automatically if that same player droppes the flag? Not sure how that all works

Side note: Warsong Scramble TLDR: 3 flags in each base, 10 points max

Golrag:
Okay, no idea about packets but I did it to get some quick answers:
- It updates automatically
- Its not the last one that picks it up or the first one that shows up. It can overwrite current ArenaSlot. (ex, i had flag, then druid, druid showed up on arena slot. Swapped it around then).

Shauren:
It just does this
```
ServerToClient: SMSG_BATTLEGROUND_PLAYER_POSITIONS (0x2928) Length: 80 ConnIdx: 1 Time: 02/07/2021 17:10:40.454 Number: 42649
BattlegroundPlayerPositionCount: 4
[0] Guid: Full: 1 Player
[0] Pos: X: 1130.923 Y: 1448.4581
[0] IconID: 1
[0] ArenaSlot: 3
[1] Guid: Full: 2 Player
[1] Pos: X: 1058.4755 Y: 1458.3544
[1] IconID: 1
[1] ArenaSlot: 3
[2] Guid: Full: 3 Player
[2] Pos: X: 1075.4467 Y: 1460.2341
[2] IconID: 1
[2] ArenaSlot: 3
[3] Guid: Full: 4 Player
[3] Pos: X: 1243.1815 Y: 1439.6768
[3] IconID: 2
[3] ArenaSlot: 2
```

Golrag:
Okay, well I think the code in this PR already supports this then. Maybe the client chooses the "correct" target

Shauren:
Before I merge - what happens with this PR if a player with one of these auras disconnects and reconnects?

Golrag:
I think that should trigger an aura remove, right?
Not sure what happens on retail. I think the character stays in the world for a while but unsure if the flag drops. Will test what happens in the PR

Shauren:
What I want to know is whether that creates duplicate entries for same guid in _playerPositions or not

Golrag:
GetBattleground is not nullptr and thus the player gets removed from the vector. BUT I noticed something that might be an issue.
I wait to log in till my character is removed from world, then login -> I get the aura back. Could this lead to possible other issues? Thinking about having multiple flag carriers suddenly. Might just be me using .cast instead of clicking the flag. Think its out of scope of this PR tho

Golrag:
And... I found a way to make battleground return nullptr: teleport the player while he is offline and carrying a flag

Golrag:
So this might happen when a player gets teleported out (hearthstone/teleport from gm). Which calls Battleground::RemovePlayerAtLeave which removes the battleground reference from player.

I have a fix for that one

PR #25989, opened 2021-01-29
Core/Creature: Allow to skip Rate.Corpse.Decay.Looted when calling SetCorpseDelay()

### LINKED ISSUE #25629: Core/SAI: Timer of SMART_ACTION_SET_CORPSE_DELAY behavior
**Description:**

I set it to 3, creature despawns instantly, I set it to 30, creature despawns in 15 sec

**Expected behaviour:**

Corpse delay 3 should mean 3 seconds and not 0
30 should mean 30, not 15

1. .go c id 25332
2. kill, notice instant despawn instead of 3 seconds

**Branch(es):** 3.3.5
**TC rev. hash/commit:** c2a0b78a051ea3301d8769b2c11fae00e14fe095

jackpoz:
**Changes proposed:**
Core/Creature: Allow to skip Rate.Corpse.Decay.Looted when calling SetCorpseDelay()
Add a second parameter to SetCorpseDelay() that specifies if Rate.Corpse.Decay.Looted setting should be ignored, false by default (aka don't ignore by default).
Add a second parameter to SMART_ACTION_SET_CORPSE_DELAY to specify if Rate.Corpse.Decay.Looted should be included, false by default (aka ignore by default).

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #25629

**Tests performed:**

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -2170,7 +2170,7 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
             for (WorldObject* const target : targets)
             {
                 if (IsCreature(target))
-                    target->ToCreature()->SetCorpseDelay(e.action.corpseDelay.timer);
+                    target->ToCreature()->SetCorpseDelay(e.action.corpseDelay.timer, !e.action.corpseDelay.includeDecayRatio);
             }
 
             break;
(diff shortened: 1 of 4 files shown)


offl:
Looks like
```sql
UPDATE `smart_scripts` SET `action_param2` = 1 WHERE `action_type` = 116;
```
.go c id 25332
kill

jackpoz:
By default with action_param2 = 0 the corpse is removed after the time specified.
With action_param2 = 1 then corpse decay ratio kicks in and the corpse is removed after 1 second.

Basically you don't need to do any sql update to get SAI to force the corpse removal to the current values stored in SAI

offl:
Oh, ok

illfated:
First time I see this type of error message in AppVeyor (but maybe you have seen it before):
`Error uploading artifact the storage: Unable to read data from the transport connection: The connection was closed.`
At least it is not build related, so basically all the builds are OK.

PR #25998, opened 2021-01-31
Core/PathFinding: Allow to skip calling UpdateAllowedPositionZ() per creature/per spawn

### LINKED ISSUE #21969: Core/Movements: Creatures can walk up but cannot walk down, when mmap.enablePathFinding = 1.
**Description:**

Creatures can walk but cannot walk down, when mmap.enablePathFinding = 1.

**Current behaviour:**

When fight with malchezaar, I was hit to the corner. then malchezaar run to the low wall, and never come down again.

**Expected behaviour:**

If his goal is at the corner, he should stand in a reasonable place and attack the target instead of putting on the wall. Even on the wall should be able to come down. 

1. Step 1 
    open GM mode(.gm on).    type ".go xyz -10962 -2018 276 532"  teleport to malchezaar
2. Step 2
    close GM mode(.gm off).   attack Malchezaar and leading him to the corner
3. Step 3
    Waiting for Malchezaar to walk to the wall, if he don't go, just move left and right .

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. f8427ccfb2d8+ 2018-05-26 22:11:52 +0200 (3.3.5 branch) (Unix, RelWithDebInfo, Dynamic)
Using SSL version: OpenSSL 1.0.2n  7 Dec 2017 (library: OpenSSL 1.0.2n  7 Dec 2017)
Using Boost version: 1.62.0
Using MySQL version: 10.1.26-MariaDB
Using CMake version: 3.7.2
Compiled on: Linux 4.9.0-6-amd64
MMAPs status: Enabled
maps directory located in ./maps. Total size: 252190595 bytes
vmaps directory located in ./vmaps. Total size: 588247333 bytes
mmaps directory located in ./mmaps. Total size: 2152621560 bytes
Using zhCN DBC Locale as default. All available DBC locales: zhCN 
Using World DB: TDB 335.64

**Operating system:** debian9

jackpoz:
**Changes proposed:**

Core/PathFinding: Allow to skip calling UpdateAllowedPositionZ() per creature/per spawn

Add a new column creature_template_movement.SkipHeightNormalization and creature_movement_override.SkipHeightNormalization to skip calling UpdateAllowedPositionZ() when generating a path with MMaps. In some cases UpdateAllowedPositionZ() picks the floor above, causing the creature to end up in the wrong place and causing further issues later on as the wrong MMap poly gets selected.

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Allows to fix #21969

**Tests performed:**

https://imgur.com/a/YZKV8lP

- [ ] Another way could be checking why UpdateAllowedPositionZ() gives wrong Z in the first place


jackpoz:
To fix Malchezaar the table has to be filled with the required data:
```sql
INSERT INTO `creature_template_movement` (`CreatureId`, `SkipHeightNormalization`) VALUES (15690, 1);
```

It can be reloaded at runtime with `.reload creature_template 15690` to see the difference

Macsimka:
Also related to collision height - https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Object/Object.cpp#L3570.
Example: https://i.imgur.com/pvUIYMd.jpg
Maybe we shouldn't use CreatureModelScale in Collision height calculation here, don't know.

jackpoz:
A lot of those + collision height were added https://github.com/TrinityCore/TrinityCore/commit/e42903ec16f4849b71bc8736e3151a1e11d3cd25 instead of using fixed values to "fix LoS issues" but it seems they broke a lot of mmaps uses

jackpoz:
For now let's use https://github.com/TrinityCore/TrinityCore/pull/26092 instead

PR #26019, opened 2021-02-02
Core/Spells: Fix Seed of Corruption (Warlock) target selection

### LINKED ISSUE #26010: Seed of corruption's explosion damages through walls and structures.
**Description:**

Seed of corruption's explosion will hit creatures in other rooms through walls, this is especially bad in dungeons, a swarm of creatures aggroed from another room will run at you.

Here's an example in scholomance it hits the skeletons in a room under the ground
image

**Expected behaviour:**

Seed of corruption should not hit mobs behind walls or out of LOS

1. spawn a mob anywhere with another mob behind a wall, or udner it, make seed explode, see if it damages the other mobs. Or just try it in scholomance

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

92c8ffc047ea35ea9a534a878cd4be5726f20974

**Operating system:** 

windows

Jildor:
**Changes proposed:**

-  Patch 2.3.0 (2007-11-13): Seed of Corruption detonation will now obey line of sight: https://wow.gamepedia.com/Seed_of_Corruption
-  Seed of Corruption detonation should also hit seed of corruption actionTarget (this was working fine because GetExplTargetUnit() can't get that target here, I only remove that unnecessary check): http://www.worldoflogs.com/reports/ak6ponc2cou5kufz/xe/?s=1062&e=1349&x=spell+%3D+%22Seed+of+Corruption%22%0D%0AAND+sourceName+%3D+%22Plazma%22

- [X] 3.3.5
- [ ] master

**Issues addressed:**

Closes #26010

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/scripts/Spells/spell_warlock.cpp
+++ b/src/server/scripts/Spells/spell_warlock.cpp
@@ -1156,7 +1156,7 @@ class spell_warl_seed_of_corruption : public SpellScript
         targets.remove_if([&](WorldObject const* target)
         {
             if (Unit const* unitTarget = target->ToUnit())
-                if (WorldLocation* dest = const_cast<WorldLocation*>(GetExplTargetDest()))
+                if (WorldLocation const* dest = GetExplTargetDest())
                     if (!unitTarget->IsWithinLOS(dest->GetPositionX(), dest->GetPositionY(), dest->GetPositionZ()))
                         return true;
 
--- a/src/server/scripts/Spells/spell_warlock.cpp
+++ b/src/server/scripts/Spells/spell_warlock.cpp
@@ -1147,31 +1147,27 @@ class spell_warl_seduction : public SpellScriptLoader
 };
 
 // -27285 - Seed of Corruption
-class spell_warl_seed_of_corruption : public SpellScriptLoader
+class spell_warl_seed_of_corruption : public SpellScript
 {
-    public:
-        spell_warl_seed_of_corruption() : SpellScriptLoader("spell_warl_seed_of_corruption") { }
+    PrepareSpellScript(spell_warl_seed_of_corruption);
 
-        class spell_warl_seed_of_corruption_SpellScript : public SpellScript
+    void FilterTargets(std::list<WorldObject*>& targets)
+    {
+        targets.remove_if([&](WorldObject const* target)
         {
-            PrepareSpellScript(spell_warl_seed_of_corruption_SpellScript);
+            if (Unit const* unitTarget = target->ToUnit())
+                if (WorldLocation* dest = const_cast<WorldLocation*>(GetExplTargetDest()))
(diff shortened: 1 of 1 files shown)


src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -1147,31 +1147,27 @@ class spell_warl_seduction : public SpellScriptLoader
...
         {
-            PrepareSpellScript(spell_warl_seed_of_corruption_SpellScript);
-
-            void FilterTargets(std::list<WorldObject*>& targets)
-            {
-                if (GetExplTargetUnit())
-                    targets.remove(GetExplTargetUnit());
-            }
+            if (Unit const* unitTarget = target->ToUnit())
+                if (WorldLocation* dest = const_cast<WorldLocation*>(GetExplTargetDest()))
```
jackpoz, line 1159: do you need the const_cast for something in particular here ?

jackpoz:
does the LoS exclusion actually need a spell script ? isn't there an out of the box way to achieve the same ? or would that be considered a "hack" etc ?

Jildor:
can remove:
```
AttributesEx2: 0x00000004 (SPELL_ATTR2_CAN_TARGET_NOT_IN_LOS)
```
 but it would surely be hacky, so I have decided to filter by spellscript, I don't know if there would be a more proper way

src/server/scripts/Spells/spell_warlock.cpp
Jildor: copypaste from: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp#L2294

offl:
```cpp
    // Seed of Corruption
    ApplySpellFix({ 27285 }, [](SpellInfo* spellInfo)
    {
        spellInfo->AttributesEx2 &= ~SPELL_ATTR2_CAN_TARGET_NOT_IN_LOS;
    });
```

Jildor:
@offl As I have indicated to jackpoz, I think that would be considered hacky, if you remove that attribute, what happens if the caster is out of LoS when the proc occurs?  Maybe the aoe would not be launched in the actiontarget position as it is not in LoS with caster?

offl:
I'm not really a spell stuff guy
Just pointed to possible scenario in case you meant something different by 'can remove'

src/server/scripts/Spells/spell_warlock.cpp
mdX7: in the linked case the WorldLocation gets relocated, we're just getting the coords here => const_cast not required

Jildor: yep sorry, I couldn't make the change yesterday, thanks to both

jackpoz:
Thanks for the PR :)

PR #26028, opened 2021-02-03
Core/Spells Adding object_type to checker to early-eliminate alive units

### LINKED ISSUE #25994: Core/Spells Massress hits living player and self
**Description:**

If you cast a massress spell like https://www.wowhead.com/spell=212036/mass-resurrection you hit everything in your group like Living ppl and yourself

**Expected behaviour:**

This spell should only hit Dead People in your group/raid

1. create a priest and skill to holy/disci
2. use massress
3. see that the "Yellow" animation hits everything in your group

**Branch(es):**

master

**TC rev. hash/commit:** 

23144d0665504e2e24ee790e900bb8de0df4aa08

**Operating system:** 

Win10

matanshukry:
**Changes proposed:**

- Added object type into checker that is used by grid search to early eliminate alive units when looking for corpses
- Removed script 'spell_hun_pet_carrion_feeder'. Ability doesn't seem to exist and neither the spells that were attached/related to it

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #25994

**Tests performed:**

- [x] It builds
- [x] Tested in-game


sql/updates/world/master/2021_02_03_00_world_remove_pet_carrion.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM spell_script_names WHERE ScriptName='spell_hun_pet_carrion_feeder';
```
mdX7, line 1: you actually updated that script.. so why remove it from db then? :D

src/server/game/Spells/Spell.cpp
```diff
@@ -7961,6 +7961,17 @@ bool WorldObjectSpellTargetCheck::operator()(WorldObject* target)
             default:
                 break;
         }
+
+        switch (_objectType)
+        {
+            case TARGET_OBJECT_TYPE_CORPSE:
+            case TARGET_OBJECT_TYPE_CORPSE_ALLY:
+            case TARGET_OBJECT_TYPE_CORPSE_ENEMY:
+                if (unitTarget->IsAlive())
```
mdX7, line 7970: possible crash: you gotta check if `unitTarget` is != null, its a cast `Unit* unitTarget = target->ToUnit();` which may result in null value.

You might just put it inside the `if (unitTarget)` block above. Even above the other switch-case block for even faster reaction.

sql/updates/world/master/2021_02_03_00_world_remove_pet_carrion.sql
matanshukry: I thought I did :-O
I'll delete it again I guess

src/server/game/Spells/Spell.cpp
matanshukry: It's hard to tell due to github collapsing code areas, but this block is actually inside a `if (unitTarget)` block, so there's no need to check that.

PR #26029, opened 2021-02-03
Core/Spells Add EffectAttributesFlags with NoScaleWithStack (0x40

### LINKED ISSUE #26021: Core/Spells Prayer of Mending aura (41635) value is multiplied by stack amount
**Description:**

When casting spell 41635 (prayer of mending aura) with spell modifiers with SPELLVALUE_BASE_POINT0 and SPELLVALUE_AURA_STACK, it will show the text `Heals for <value> the next time you take damage.`.
Instead of <value> though, it will show the result of BasePoint * AuraStack. It should only show the BasePoint.

The place where we multiply the value by the stack amount is this:
https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L670

**Expected behaviour:**

The value should not be multiplied by the stack amount.
Not for this spell specifically anyway, and probably others too - depending on some condition.

1. Cast spell 41635 with both modifiers as specified above
2. Note the actual tooltip shown in game is a product of the 2 values, rather than just the base points values

**Branch(es):**

master

**TC rev. hash/commit:** 

23144d0665504e2e24ee790e900bb8de0df4aa08

**Operating system:** 

Windows 10

matanshukry:
**Changes proposed:**

- Defined enum EffectAttributesFlags with NoScaleWithStack value (0x40) and implemeneted

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #26021 

**Tests performed:**

- [x] It builds
- [x] Tested in-game

@mdX7 Thanks for helping figuring this out!

### DIFF
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -372,6 +372,7 @@ SpellImplicitTargetInfo::StaticData  SpellImplicitTargetInfo::_data[TOTAL_SPELL_
 };
 
 SpellEffectInfo::SpellEffectInfo(SpellInfo const* spellInfo, SpellEffectEntry const* _effect)
+    : EffectAttributes(EFFECT_ATTRIBUTES_FLAGS_NO_SCALE_WITH_STACK)
 {
     ASSERT(spellInfo);
     ASSERT(_effect);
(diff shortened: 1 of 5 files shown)


src/server/game/Spells/SpellInfo.cpp
```diff
@@ -372,6 +372,7 @@ SpellImplicitTargetInfo::StaticData  SpellImplicitTargetInfo::_data[TOTAL_SPELL_
 };
 
 SpellEffectInfo::SpellEffectInfo(SpellInfo const* spellInfo, SpellEffectEntry const* _effect)
+    : EffectAttributes(EFFECT_ATTRIBUTES_FLAGS_NO_SCALE_WITH_STACK)
```
mdX7, line 375: we should initialize it with _NONE

matanshukry: yeah that should of been `EFFECT_ATTRIBUTES_FLAGS_NONE`. Was definitely a mistake. I'll fix that

Shauren (approved):

PR #26041, opened 2021-02-05
Core/Spells: GetPrevSpellChain check disabled spells + riding skills

matanshukry:
**Changes proposed:**

- GetPrevSpellChain can now get a unit to return only non-disabled spells for that unit
- Player will now check DisableMgr on learning spells
- added reload command to disables
- SQL file to add artisan riding spell as disabled spells for players and update the trainer's riding spells 

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -3136,6 +3136,8 @@ void Player::LearnSpell(uint32 spell_id, bool dependent, int32 fromSkill /*= 0*/
         SendDirectMessage(packet.Write());
     }
 
+    disabled |= DisableMgr::IsDisabledFor(DISABLE_TYPE_SPELL, spell_id, this, 0);
+
     // learn all disabled higher ranks and required spells (recursive)
     if (disabled)
     {
--- a/src/server/scripts/Commands/cs_disable.cpp
+++ b/src/server/scripts/Commands/cs_disable.cpp
@@ -77,7 +77,7 @@ public:
         return commandTable;
     }
 
-    static bool HandleReloadDisablesCommand(ChatHandler* handler, char const* args)
+    static bool HandleReloadDisablesCommand(ChatHandler* /*handler*/, char const* /*args*/)
     {
         DisableMgr::LoadDisables();
         return true;
--- a/src/server/scripts/Commands/cs_disable.cpp
+++ b/src/server/scripts/Commands/cs_disable.cpp
@@ -66,14 +66,21 @@ public:
         };
         static std::vector<ChatCommand> disableCommandTable =
         {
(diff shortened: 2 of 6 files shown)


src/server/scripts/Commands/cs_disable.cpp
```diff
@@ -76,6 +77,12 @@ class disable_commandscript : public CommandScript
         return commandTable;
     }
 
+    static bool HandleReloadDisablesCommand(ChatHandler* handler, char const* args)
```
illfated, line 80: "unused parameter" as usual.

src/server/scripts/Commands/cs_disable.cpp
```diff
@@ -76,6 +77,12 @@ class disable_commandscript : public CommandScript
         return commandTable;
     }
 
+    static bool HandleReloadDisablesCommand(ChatHandler* /*handler*/, char const* /*args*/)
```
Shauren, line 80: This command already exists in the `.reload` group

src/server/game/Entities/Player/Player.cpp
```diff
@@ -3136,6 +3136,8 @@ void Player::LearnSpell(uint32 spell_id, bool dependent, int32 fromSkill /*= 0*/
         SendDirectMessage(packet.Write());
     }
 
+    disabled |= DisableMgr::IsDisabledFor(DISABLE_TYPE_SPELL, spell_id, this, 0);
```
Shauren, line 3139: `disabled` for known spells is a completely different thing, its a leftover from 3.3.5 when actual spell ranks were a thing you had to learn from trainers and was used to hide learned higher ranks of active talents (it also does nothing here)

Shauren:
sql pushed in c6c583a63e188c1d8d81d0141a0f188479a98d0f

matanshukry:
> sql pushed in [c6c583a](https://github.com/TrinityCore/TrinityCore/commit/c6c583a63e188c1d8d81d0141a0f188479a98d0f)

What about the cpp part of it? ..

illfated:
Maybe check how much of it is needed, based on the current master branch content, then build a new PR based on what you think is still missing?

matanshukry:
> Maybe check how much of it is needed, based on the current master branch content, then build a new PR based on what you think is still missing?

Looked at commits and found a commit around 10h ago from Shauren that made the cpp part of it not relevant.

I don't understand how this repo is working; just manually pushing commits without any PRs/issues/informative comments, so that other people PRs (and hence their time and effort) becomes meaningless? ..

illfated:
Well, sometimes the TC owners want a different solution than the one presented in a PR, so they do it like this. Sure, it would have been nice to have an explanatory comment about what is the chosen solution, but sometimes we don't get the explanation before we see the replacement commit(s). Feel free to ask in IRC or Discord, since I don't know the specific answer for this case.

PR #26043, opened 2021-02-06
Scripts/Spells: Touch of the Magi - accumulate damage and DoT duration

### LINKED ISSUE #26040: Core/Spells Touch of the Magi not work
**Description:**

Spellid: 321507

When used on an enemy, it applies an aura of 210824 to him and for 8 seconds, accumulates a charge and then deals damage to the enemy. Damage is calculated for 25% of the damage done to the enemy

**Expected behaviour:**

When used, it must impose on the target the aura 210824, which for 8 seconds accumulates damage dealt to the character, and releases 25% of the damage to the target on which the aura hangs

1. Create a character class mage
2. Change character spec to Arcane
3. Raise character level to 33
4. Cast spell 321507 on enemy target

**Branch(es):**

master

**TC rev. hash/commit:** 

db18aac

**Operating system:** 

Windows 10 x64

matanshukry:
**Changes proposed:**

- Touch of the magi will now accumulate damage and explode for said damage over time

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #26040 

**Tests performed:**

- [x] It builds
- [x] Tested in-game


Faq:
From where did you know that it was not working?

matanshukry:
> From where did you know that it was not working?

#26040
Updated op as well.

illfated:
title suggestion: **Scripts/Spells: Touch of the Magi**
(due to my linguistic & grammar orientation, I find the word "fixed" to be more ambiguous than useful)

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -802,6 +804,50 @@ class spell_mage_water_elemental_freeze : public SpellScript
...
+    void HandleProc(AuraEffect const* /*aurEff*/, ProcEventInfo& eventInfo)
+    {
+        DamageInfo* damageInfo = eventInfo.GetDamageInfo();
+        if (damageInfo)
+            if (damageInfo->GetAttacker() == GetCaster() && damageInfo->GetVictim() == GetTarget())
+            {
+                uint32 extra = CalculatePct(damageInfo->GetDamage(), 25);
+                if (extra > 0)
+                {
+                    _sum += CalculatePct(damageInfo->GetDamage(), 25);
```
mdX7, line 828: you can use the `extra` variable here

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -802,6 +804,50 @@ class spell_mage_water_elemental_freeze : public SpellScript
...
+                    AuraEffect *auraEffect = GetTarget()->GetAuraEffect(SPELL_MAGE_TOUCH_OF_THE_MAGI_AURA, 0, GetCaster()->GetGUID());
```
mdX7, line 829: use `*` always directly after type, no space inbetween => `AuraEffect* auraEffect`

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -802,6 +804,50 @@ class spell_mage_water_elemental_freeze : public SpellScript
...
+                    _sum += extra;
+                    AuraEffect* auraEffect = GetTarget()->GetAuraEffect(SPELL_MAGE_TOUCH_OF_THE_MAGI_AURA, 0, GetCaster()->GetGUID());
+                    auraEffect->ChangeAmount(_sum);
+                }
+            }
+    }
+
+    void AfterRemove(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+    {
+        if (_sum == 0 || GetTargetApplication()->GetRemoveMode() != AURA_REMOVE_BY_EXPIRE)
```
Shauren, line 837: `_sum` in this function is the same as `aurEff->GetAmount()`

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -802,6 +804,50 @@ class spell_mage_water_elemental_freeze : public SpellScript
...
+        DamageInfo* damageInfo = eventInfo.GetDamageInfo();
+        if (damageInfo)
+            if (damageInfo->GetAttacker() == GetCaster() && damageInfo->GetVictim() == GetTarget())
+            {
+                uint32 extra = CalculatePct(damageInfo->GetDamage(), 25);
+                if (extra > 0)
+                {
+                    _sum += extra;
+                    AuraEffect* auraEffect = GetTarget()->GetAuraEffect(SPELL_MAGE_TOUCH_OF_THE_MAGI_AURA, 0, GetCaster()->GetGUID());
+                    auraEffect->ChangeAmount(_sum);
```
Shauren, line 830: `auraEffect == aurEff`

src/server/scripts/Spells/spell_mage.cpp
mdX7: so we can drop `_sum` completely, pog

src/server/scripts/Spells/spell_mage.cpp
matanshukry: Is that smart? what if down the line we decide aurEff argument will be one that actually shouldn't be changed?
Otherwise - what would you say about removing the const from the function signature?

Shauren: But your code is doing the same - since both point to the same AuraEffect object

matanshukry: the result is the same currently because the const that is passed is the same as the one from `GetAuraEffect` and like you said, it doesn't have to be const. But in the future if someone will want to change the variable that is passed, they won't go looking into each script and check no one is doing a const_cast.

Either way `const_cast` isn't something we should do. Like I said, if it's the same I can remove the const from the signature, assuming you agree.

Shauren: Remove const from argument, I agree.

matanshukry:
> title suggestion: **Scripts/Spells: Touch of the Magi - accumulate damage and DoT duration**
> (due to my linguistic & grammar orientation, I find the word "fixed" to be more ambiguous than useful)

Just saw it now. Better late than never I guess

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -802,6 +804,45 @@ class spell_mage_water_elemental_freeze : public SpellScript
...
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_MAGE_TOUCH_OF_THE_MAGI_EXPLODE });
+    }
+
+    void HandleProc(AuraEffect* aurEff, ProcEventInfo& eventInfo)
+    {
+        DamageInfo* damageInfo = eventInfo.GetDamageInfo();
+        if (damageInfo)
+            if (damageInfo->GetAttacker() == GetCaster() && damageInfo->GetVictim() == GetTarget())
```
Shauren, line 821: codestyle: wrap in brackets (you do it on first required inner level, then all the way to the outermost one)

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -802,6 +804,45 @@ class spell_mage_water_elemental_freeze : public SpellScript
...
+            {
+                uint32 extra = CalculatePct(damageInfo->GetDamage(), 25);
+                if (extra > 0)
+                    aurEff->ChangeAmount(aurEff->GetAmount() + extra);
+            }
+    }
+
+    void AfterRemove(AuraEffect const* aurEff, AuraEffectHandleModes /*mode*/)
+    {
+        int amount = aurEff->GetAmount();
```
Shauren, line 831: dont use plain `int`, its `int32`

ghost (approved):
Looks pretty good to me. A good amount of `const` simplification, which I presume is good too.

PR #26044, opened 2021-02-06
Core/GameObjects: Add GameObjectPlaySpellVisual

Golrag:
**Changes proposed:**

-  Allow GameObjects to send SMSG_GAME_OBJECT_PLAY_SPELL_VISUAL
-  Also updates the correct update field

- master

**Issues addressed:**

Closes: none?

**Tests performed:**

- Builds

- none

**Notes:**
- Based on the structure of WPP
- Can be used for Capture Points

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -2758,6 +2758,17 @@ void GameObject::SetAnimKitId(uint16 animKitId, bool oneshot)
     SendMessageToSet(activateAnimKit.Write(), true);
 }
 
+void GameObject::SetSpellVisualId(int32 spellVisualId, ObjectGuid activatorGuid)
+{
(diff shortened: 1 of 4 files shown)


src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -2777,6 +2777,16 @@ class GameObjectModelOwnerImpl : public GameObjectModelOwnerBase
     GameObject* _owner;
 };
 
+void GameObject::SetSpellVisual(int32 spellVisualId, ObjectGuid activatorGuid)
+{
+    SetUpdateFieldValue(m_values.ModifyValue(&GameObject::m_gameObjectData).ModifyValue(&UF::GameObjectData::SpellVisualID), spellVisualId);
```
Shauren, line 2782: should it really always set the visual permanently?

Golrag: I only checked the capture points in Arathi Basin and Battle for Gilneas at the end of the legion expansion, the value was send in the update object packet after. Not sure if other types of gameobject have different behaviour

PR #26055, opened 2021-02-08
Core/SmartScripts: new FlagsArgument for SmartActionSummonCreature

matanshukry:
**Changes proposed:**

Added flags for smart action summon creature: 
- prefer unit (important due to the source in smart events)
- personal spawn

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game


matanshukry:
@Shauren review this please.

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -1182,10 +1182,14 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
         }
         case SMART_ACTION_SUMMON_CREATURE:
         {
-            WorldObject* summoner = GetBaseObjectOrUnit(unit);
+            EnumFlag<SmartActionSummonCreatureFlags> flags(static_cast<SmartActionSummonCreatureFlags>(e.action.summonCreature.flags));
+            bool preferUnit = flags.HasFlag(SmartActionSummonCreatureFlags::PreferUnit);
+            WorldObject* summoner = preferUnit ? (unit ? unit : GetBaseObject()) : GetBaseObjectOrUnit(unit);
```
BAndysc, line 1187: I do not like it works in quite nondeterministic way. 

The purpose of the flag is to use action invoker as the summoner. So why fallback to BaseObject when invoker is null?

I’d rather keep it simple and obvious - instead of `prefer unit`, imho it is better to make the flag `use invoker as summoner`. And if there is no invoker, then simply do not summon unit.

I do not see any reason why fallback to BaseObject will be good. If someone needs to use invoker as summoner, then he has valid reason not to use BaseObject (for instance making it personal spawn)

matanshukry: Cool, I agree.
I'm keeping the other side (without the flag) with the fallback though, due to compatibility.

illfated:
Good PR title. Thank you. 👍🏻

PR #26065, opened 2021-02-09
Core/SmartAI: AI change on charmed

### LINKED ISSUE #23582: [3.3.5]Spell: Enslave Demon
**Description:**

Bug with warlock's Enslave Demon

**Current behaviour:**

1. Warlocks can cast Enslave Demon, even they have not dismissed their pets.
2. Warlock cannot control enslaved demon to attack, but can let them follow, stop and use their spell.
3. Enslaved demons' spell have no gcd. For example, when you enslave a elite Felguard in Blasted Lands, it can spell Mortal Strike as fast as your button clicking.

**Expected behaviour:**

1. Warlocks can only ensalve Demon when their pets are already dismissed.
2. Enslaved demons should be able to attcak.
3. Enslaved demons' spell should have gcd and cd.

1. .go c id 8717
2. enslave it
3. try above issues.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

4dbdc22b3b5770ff8252ea057bf7a7a92dcb31cb

**Operating system:** CentOS 7

### LINKED ISSUE #25982: Priest - Mind Control VS mobs : no melee attack ?
**Description:**

Mind control currently works on humanoids mobs (spells, movements, behaviour) but no melee attack is performed. 

**Expected behaviour:**

Mind controled mobs should be able to perform melee attack ('claws' icon). Is this something I've never noticed ? Everything else is fine.

1. Mind control an humanoid mob
2. Attack some other mob
3. Try to initiate a melee attack =>  nothing happens
4. Spells, movements, etc are working as expected

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TC> TrinityCore rev. 1ff4a7e75b56+ 2021-01-27 02:04:30 +0200 (3.3.5 branch) (Unix, RelWithDebInfo, Static)
Using SSL version: OpenSSL 1.1.1d  10 Sep 2019 (library: OpenSSL 1.1.1d  10 Sep 2019)
Using Boost version: 1.67.0
Using MySQL version: 100327
Using CMake version: 3.13.4
Compiled on: Linux 5.4.78-2-pve
MMAPs status: Enabled
maps directory located in ../../data/maps. Total size: 251987687 bytes
vmaps directory located in ../../data/vmaps. Total size: 658130721 bytes
mmaps directory located in ../../data/mmaps. Total size: 1962380076 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS
Using World DB: TDB 335.21011

**Operating system:** 

Debian Buster 10.7

engilas:
**Changes proposed:**

-  SmartAI calls base method UnitAI::OnCharmed to update AI on charmed

Without this change, charmed creature can't update AI.

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #25982 (melee attack works correctly)
Partially #23582 (part 2: Enslaved demons should be able to attcak)

**Tests performed:**

Built and tested with warlock's 'Enslave Demon' spell, priest's 'Mind control' spell.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -704,7 +704,8 @@ void SmartAI::OnCharmed(bool isNew)
 
     GetScript()->ProcessEventsFor(SMART_EVENT_CHARMED, nullptr, 0, 0, charmed);
 
-    UnitAI::OnCharmed(isNew);
+    if (!GetScript()->HasAnyEventWithFlag(SMART_EVENT_FLAG_WHILE_CHARMED)) // we can change AI if there are no events with this flag
+        UnitAI::OnCharmed(isNew);
 }
 
 void SmartAI::DoAction(int32 param)
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -669,7 +669,7 @@ void SmartAI::PassengerBoarded(Unit* who, int8 seatId, bool apply)
     GetScript()->ProcessEventsFor(apply ? SMART_EVENT_PASSENGER_BOARDED : SMART_EVENT_PASSENGER_REMOVED, who, uint32(seatId), 0, apply);
 }
 
-void SmartAI::OnCharmed(bool /*isNew*/)
+void SmartAI::OnCharmed(bool isNew)
 {
     bool const charmed = me->IsCharmed();
     if (charmed) // do this before we change charmed state, as charmed state might prevent these things from processing
@@ -703,6 +703,8 @@ void SmartAI::OnCharmed(bool /*isNew*/)
     }
 
     GetScript()->ProcessEventsFor(SMART_EVENT_CHARMED, nullptr, 0, 0, charmed);
+
+    UnitAI::OnCharmed(isNew);
 }
 
 void SmartAI::DoAction(int32 param)
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -60,6 +60,7 @@ SmartScript::SmartScript()
     mCurrentPriority = 0;
     mEventSortingRequired = false;
     mNestedEventsCounter = 0;
+    mAllEventFlags = 0;
 }
 
 SmartScript::~SmartScript()
@@ -3836,15 +3837,13 @@ void SmartScript::FillScript(SmartAIEventList e, WorldObject* obj, AreaTriggerEn
 
         if (scriptholder.event.event_flags & SMART_EVENT_FLAG_DIFFICULTY_ALL)//if has instance flag add only if in it
         {
-            if (obj && obj->GetMap()->IsDungeon())
-            {
-                if ((1 << (obj->GetMap()->GetSpawnMode()+1)) & scriptholder.event.event_flags)
-                {
-                    mEvents.push_back(scriptholder);
-                }
-            }
-            continue;
+            if (!(obj && obj->GetMap()->IsDungeon()))
+                continue;
+
+            if (!(1 << (obj->GetMap()->GetSpawnMode() + 1) & scriptholder.event.event_flags))
+                continue;
         }
+        mAllEventFlags |= scriptholder.event.event_flags;
         mEvents.push_back(scriptholder);//NOTE: 'world(0)' events still get processed in ANY instance mode
     }
 }
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9511,7 +9511,6 @@ void Unit::ScheduleAIChange()
         RestoreDisabledAI();
         PushAI(nullptr); //This could actually be PopAI() to get the previous AI but it's required atm to trigger UpdateCharmAI()
     }
-    RefreshAI();
 }
 
 void Unit::RestoreDisabledAI()
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9511,6 +9511,7 @@ void Unit::ScheduleAIChange()
         RestoreDisabledAI();
         PushAI(nullptr); //This could actually be PopAI() to get the previous AI but it's required atm to trigger UpdateCharmAI()
     }
+    RefreshAI();
 }
 
 void Unit::RestoreDisabledAI()
--- a/src/server/game/AI/SmartScripts/SmartScript.h
+++ b/src/server/game/AI/SmartScripts/SmartScript.h
@@ -56,6 +56,7 @@ class TC_GAME_API SmartScript
         uint32 GetPathId() const { return mPathId; }
         WorldObject* GetBaseObject() const;
         WorldObject* GetBaseObjectOrPlayerTrigger() const;
+        bool HasAnyEventWithFlag(uint32 flag) const { return mAllEventFlags & flag; }
         static bool IsUnit(WorldObject* obj);
         static bool IsPlayer(WorldObject* obj);
         static bool IsCreature(WorldObject* obj);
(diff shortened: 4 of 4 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9511,6 +9511,7 @@ void Unit::ScheduleAIChange()
         RestoreDisabledAI();
         PushAI(nullptr); //This could actually be PopAI() to get the previous AI but it's required atm to trigger UpdateCharmAI()
     }
+    RefreshAI();
```
jackpoz, line 9514: the whole point of ScheduleAIChange() is to not change it during this world update tick but on next one to avoid crashes

jackpoz (changes requested):

src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -703,6 +703,8 @@ void SmartAI::OnCharmed(bool /*isNew*/)
     }
 
     GetScript()->ProcessEventsFor(SMART_EVENT_CHARMED, nullptr, 0, 0, charmed);
+
+    UnitAI::OnCharmed(isNew);
```
jackpoz, line 707: did you check CreatureAI::OnCharmed() too ? any idea about which one of the 2 should be called (as CreatureAI is the base class of SmartAI) and what kind of issues the 2 might create ?

jackpoz: By looking at https://github.com/TrinityCore/TrinityCore/blob/e33384d4ddbd4477f8cc5cae6e0679ab914595a5/src/server/game/AI/CoreAI/UnitAI.h#L148-L152 it seems UnitAI::OnCharmed(isNew) is does the same as what the comment says, so calling ScheduleAIChange() , without having to trigger anything more that CreatureAI might do.
So this change should be ok.

src/server/game/Entities/Unit/Unit.cpp
engilas: Yes, refresh is unnecessary, tested without it.

jackpoz:
one thing that I wonder is if for some reason SmartAI should still stay even when charmed.
@offl do you see any issue with this PR ? is it ok if a SmartAI creature loses the SmartAI while being charmed and then regains it when the charm goes away ?

offl:
> is it ok if a SmartAI creature loses the SmartAI while being charmed and then regains it when the charm goes away ?

We have SMART_EVENT_FLAG_WHILE_CHARMED (Event can occur while player controlled)
Means we want SmartAI being active even if creature is charmed

engilas:
It's too many logic in PetAI for commands, I think I can close this PR.

jackpoz:
we can leave it open for a while as a why to discuss about the issue. it makes no sense to me that every single SAI creature doesn't change AI when charmed but it might take some time before we can merge this

jackpoz:
@engilas we came up with a compromise: SmartAI::OnCharmed() should schedule an AI change if there are no events with SMART_EVENT_FLAG_WHILE_CHARMED flag (since there are quests that require SAI to trigger some script stuff when charmed).

This should fix the issue in most of the cases, except for some cases where we still need SAI.

I don't know if there is some sort of "IsThereAnyEventWithFlag(flag)" that we can use or if we need to implement it, is that something you would like to work on or should I assign it to myself ?

engilas:
@jackpoz Yes, I can try to implement it

engilas:
@jackpoz added check for flag SMART_EVENT_FLAG_WHILE_CHARMED

ccrs:
SmartAI cant go, must stay, this way smart scripts can keep control on what happens.

The problem is that, as discussed here and a couple years ago, there is no implementation for PetAI-like features, like commands and such. I remember rising this myself. Nothing was finally done, left as is.

ccrs:
> @engilas we came up with a compromise: SmartAI::OnCharmed() should schedule an AI change if there are no events with SMART_EVENT_FLAG_WHILE_CHARMED flag (since there are quests that require SAI to trigger some script stuff when charmed).
> 
> This should fix the issue in most of the cases, except for some cases where we still need SAI.
> 
> I don't know if there is some sort of "IsThereAnyEventWithFlag(flag)" that we can use or if we need to implement it, is that something you would like to work on or should I assign it to myself ?

Aaaaand I didnt read that, sorry. Feels ok I guess, though number of world creatures without SAI should be quite low, right?

BAndysc:
@jackpoz @engilas 
> SmartAI::OnCharmed() should schedule an AI change if there are no events with SMART_EVENT_FLAG_WHILE_CHARMED flag (since there are quests that require SAI to trigger some script stuff when charmed).

IMHO this introduces hard to follow behaviour when writing script. A small flag in one single event can change the behaviour of the whole script! I do not like it. I can imagine someone copies  an event from another script which has that flag and unintentionally changes the behaviour when charmed (even if event seems unrelated). Or the other way around: someone fix the script by changing the events and unintentionally removes the event with the flag, because it is hard to spot.

Wouldn't it be better to add action SMART_ACTION_USE_SMART_AI_WHEN_CHARMED that you would use in AIIntialize method? Thanks to this, there is no way someone will miss the fact the script has such behaviour.

And you can automatically at such action to all scripts where at least one event has the flag to mimic the behaviour. But in the future scripts, it will be clear user choice to decide

offl:
Personally I treat it as temp solution because it's indeed not really what we need simply because we are checking if creature has event with CHARMED flag but it's not correct in first place because that flag is linked to action id and not to creature entry. And there should be no such flag linked to creature entry because we already know some of those charmed creatures with SAI should attack in melee.

SAI just should not break Charmed AI logic but since it does and breaks hundreds and thousands creatures, that solution is better than nothing.

BAndysc:
Still, if we want to temporary hack, then imho it is better to do it explicitly (new action) than implicitly (with event flag).

offl:
What is the actual problem with making creature attack in melee and use SAI if charmed? Too hard to implement?
The thing is even if we remove SMART_EVENT_FLAG_WHILE_CHARMED from nearly all creatures(I'll remove it anyway today with or withot this PR) and fix melee attack for charmed SAI creatures, some of them will be broken because not all of them should attack in melee. However some of them will be fixed. The broken ones should be fixed manually by adding passive reactstate or their melee attack should be disabled, depends on case

This is not about flags and only about conflicts between SAI logic and Charmed logic

ccrs:
I agree with @offl , this temporal solution feels the most suitable
- The flag is already present and nothing needs to be changed.
- When it was introduced additional work was done in form of update scripts and manual updates to adapt all existing SAI scripts
- Code modification is minimal

a new event is overkill imo

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -3836,15 +3837,13 @@ void SmartScript::FillScript(SmartAIEventList e, WorldObject* obj, AreaTriggerEn
...
-                }
-            }
-            continue;
+            if (!(obj && obj->GetMap()->IsDungeon()))
+                continue;
+
+            if (!(1 << (obj->GetMap()->GetSpawnMode() + 1) & scriptholder.event.event_flags))
+                continue;
         }
+        mAllEventFlags |= scriptholder.event.event_flags;
```
jackpoz, line 3846: I wonder if we should reset mAllEventFlags before this loop in case SAI table is reloaded at runtime

jackpoz:
I honestly don't want to re-implement PetAI (and PossessedAI) in SmartAI, have 2 different implementation for the same logic.

Charming a creature is not just about "melee attacks", there's a lot more logic into it, which is why there are AI classes that do only that.

This PR is supposed to be a compromise for the time being until someone feels like untangling this mess that SmartAI is based on, with "let's reimplement everything in SmartAI implementation ignoring everything that CreatureAI already handles". I've seen this already with starting/pausing/stopping a waypoint path.

An alternative could be have SmartAI instantiate PossessedAI/PetAI and call its hooks and process the SMART_EVENT_FLAG_WHILE_CHARMED events, but it's just an idea. I'm not even sure if the scripters should be able to override PossessedAI/PetAI actions but that's another story.

src/server/game/AI/SmartScripts/SmartScript.cpp
engilas: We don't update instance of SmartScript, instead we create new instances
https://github.com/TrinityCore/TrinityCore/blob/bf0975f658d5777f0a2812507ab0cca462b18ea7/src/server/game/AI/SmartScripts/SmartScriptMgr.cpp#L335-L343
I think the creatures still have the old version of SAI after reload

ccrs:
we're on the same page, I agree

jackpoz (approved):

ghost (approved):
Nice.

jackpoz:
Thanks for the PR :)

Issue #15870, opened 2015-11-16
[3.3.5] Sentry Totem


Keader:
spell 6495 = [Sentry Totem]

When using the spell Sentry Totem, the shaman gains the place of vision where the totem is left.
The view changes with the player by right-clicking on the buff Sentry Totem.

Issue:

After using the spell the player and the totem gain buff Bind Sight (PT) [id: 6277]
However if the shaman uses any spell or use any mount, loses Bind Sight (PT) and the view can not be changed for the Totem.
I believe also that the aura Bind Sight (PT) should be invisible.

Vid: https://www.youtube.com/watch?v=ameF6Wl9pSs (Originally posted by Killyana)

Original Issue: #9197

TC Hash: eddc759cbe49d8a1170660ca825eb4af8b02caa3

Vortalex:
Quick note, all of the issues listed can be fixed by adjusting attributes on the spell in the dbc.

But to add to the issue, Sentry Totem has other bugs...
• Buff does not disappear upon a new Sentry Totem being placed or the old one being killed
• Totem does not ping map upon being attacked

PR #26068, opened 2021-02-10
Spells/Shaman: Sentry Totem - Fix summon position.

Nalafaer:
Shaman's Sentry Totem (and only Sentry Totem) is being placed far away from the caster, which is a bug. This PR fixes that.

Spell description:
_Summons an immobile Sentry Totem with 100 health **at your feet** for 5 min that allows vision of nearby area and warns of enemies that attack it. Right-Click on buff to switch back and forth between totem sight and shaman sight._

**Changes proposed:**

- Remove `HasRadius()` check and just use the default totem summon distance when placing totems with `TARGET_DEST_CASTER_<FRONT|BACK>_<LEFT|RIGHT>` targeting, as totems should always be placed next to the caster.

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Tests performed:**

Tested on all totems placable by Shaman. They are all placed and work as intended, Sentry Totem included.

**Notes:**

- This is unrelated to #15870 which is another Sentry Totem bug.
- I wonder why the `HasRadius()` check was used for this kind of target type. Do we use `TARGET_DEST_CASTER_<FRONT|BACK>_<LEFT|RIGHT>` for anything other than totems? In that case this may have unintended side effects.

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -1398,8 +1398,7 @@ void Spell::SelectImplicitCasterDestTargets(SpellEffIndex effIndex, SpellImplici
                 case TARGET_DEST_CASTER_BACK_RIGHT:
                 {
                     static float const DefaultTotemDistance = 3.0f;
-                    if (!m_spellInfo->Effects[effIndex].HasRadius())
-                        dist = DefaultTotemDistance;
+                    dist = DefaultTotemDistance;
                     break;
                 }
                 default:


Shauren:
This change is incorrect because these target types are not exclusive to totems

obraz

Shauren:
One more example that this would break are portals during Valithria Dreamwalker encounter
```
ID - 72225 Summon Dream Portal
Effect 0: Id 28 (SPELL_EFFECT_SUMMON)
BasePoints = 1
Targets (44, 86) (TARGET_DEST_CASTER_FRONT_LEFT, TARGET_DEST_DEST_RANDOM)
EffectMiscValueA = 38186
EffectMiscValueB = 64
Radius (Id 10) 30,00
```

PR #26078, opened 2021-02-12
Core/Movement: add trace message for handle movement opcode to ease debug

matanshukry:
**Changes proposed:**

- Added trace message for handle movement opcode to ease debug
- Call SetCanTransitionBetweenSwimAndFly() when aura flight has changed
- Using GetDisplayId() instead of GetNativeDisplayId() for shapeshift forms
- Merged together druid form spell select
- Changed swift flight from level >= 71 to riding skill >= 300

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tests in-game

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -2039,9 +2039,9 @@ bool Player::IsUnderWater() const
         GetPositionZ() < (GetMap()->GetWaterLevel(GetPhaseShift(), GetPositionX(), GetPositionY()) - 2);
 }
 
-void Player::SetInWater(bool apply)
+void Player::SetInWater(bool inWater)
 {
-    if (m_isInWater == apply)
+    if (m_isInWater == inWater)
         return;
 
     //define player in water by opcodes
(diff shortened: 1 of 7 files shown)


matanshukry:
@Shauren

Do note I'm not completely sure about the `GetDisplayId` instead of `GetNativeDisplayId`.

That is, without it there are still issues with detecting when player is out of the water, and it is the fix with druid shapeshift form, I'm just not sure about everything else.

Shauren:
I will need to recheck that `GetNativeDisplayId` change because when I tested it on retail, both aquatic form and non-shapeshifted forms for highmountain tauren have the same collision height so next time I will try on a troll

Shauren:
It doesn't look like this change is correct
Model data
```
DisplayID: 2428 (Aquatic Form)
CreatureModelDataID: 8406
CollisionHeight: 2.031278
```

Packet when shapeshifting into aquatic form:
```
ServerToClient: SMSG_MOVE_SET_COLLISION_HEIGHT (0x2E11) Length: 30 ConnIdx: 1 Time: 02/12/2021 18:42:04.812 Number: 21129
MoverGUID: Full: 0xERASED
SequenceIndex: 32
Height: 2.427413225173950195
Scale: 1
Reason: 2
MountDisplayID: 0
ScaleDuration: 100
```

Troll female native model data:
```
DisplayID: 59223
CreatureModelDataID: 7793
CollisionHeight: 2.4274132
```

matanshukry:
Shauren then perhaps something else that is related to that area.

Tested with female troll on the local tc env.

When using the bigger height value (2.4) with `GetNativeDisplayId()` you will only be detected out of water if you'll try to jump while being on the water level rather than being below.

illfated:
PR title suggestion: **Core/Movement: add trace message to handle movement opcode and ease debug**
(you may of course write the title differently, using my suggestion as a guide)

PR #26082, opened 2021-02-13
Scripts/Spells: corrections to Death Knight spell Blood Mastery

matanshukry:
**Changes proposed:**

-  Using EffectLaunch rather than OnEffectHitRather and with the correct effect type

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

- [ ] Missing mechanic: the heal should be done for 25% of the damage received in the last 5 seconds. I'm not aware of any mechanics we have to do it, nor have I seen any proc flags on the spells.

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -1654,6 +1654,7 @@ void SpellMgr::LoadSpellProcs()
     }
 
     isTriggerAura[SPELL_AURA_DUMMY] = true;
+    isTriggerAura[SPELL_AURA_PERIODIC_DUMMY] = true;
     isTriggerAura[SPELL_AURA_MOD_CONFUSE] = true;
     isTriggerAura[SPELL_AURA_MOD_THREAT] = true;
     isTriggerAura[SPELL_AURA_MOD_STUN] = true; // Aura does not have charges but needs to be removed on trigger
--- a/src/server/scripts/Spells/spell_dk.cpp
+++ b/src/server/scripts/Spells/spell_dk.cpp
@@ -29,6 +29,7 @@
 #include "SpellHistory.h"
 #include "SpellMgr.h"
 #include "SpellScript.h"
+#include <numeric>
 
 enum DeathKnightSpells
 {
@@ -63,6 +64,7 @@ enum DeathKnightSpells
     SPELL_DK_RUNIC_RETURN                       = 61258,
     SPELL_DK_SLUDGE_BELCHER                     = 207313,
     SPELL_DK_SLUDGE_BELCHER_SUMMON              = 212027,
+    SPELL_DK_DEATH_STRIKE_ENABLER               = 89832, // Server Side
     SPELL_DK_TIGHTENING_GRASP                   = 206970,
     SPELL_DK_TIGHTENING_GRASP_SLOW              = 143375,
     SPELL_DK_UNHOLY                             = 137007,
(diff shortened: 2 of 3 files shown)


matanshukry:
travis-ci is failing, but looks like a cache issue to me?

illfated:
Yes, it looks like something happened to one of the travis-ci Linux files after the build started. Should be possible to resolve by triggering a new build. (I seem to remember the travis-ci page having an option to restart the build, but I can't find that option now, so it might be removed for all I know.)

illfated:
BTW, PR title suggestion: **Scripts/Spells: corrections to Death Knight spell Blood Mastery**

src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -585,12 +585,12 @@ class spell_dk_death_strike : public SpellScriptLoader
             void HandleHeal(SpellEffIndex /*effIndex*/)
             {
                 Unit* caster = GetCaster();
-                //TODO: heal = std::min(10% health, 20% of all damage taken in last 5 seconds)
-                int32 heal = CalculatePct(caster->GetMaxHealth(), GetSpellInfo()->GetEffect(EFFECT_4)->CalcValue());
-                caster->CastCustomSpell(SPELL_DK_DEATH_STRIKE_HEAL, SPELLVALUE_BASE_POINT0, heal, caster, true);
+                // heal for 25% taken for last 5 seconds, for min 7% of max health, up to
```
illfated, line 588: Comment content improvement suggestion:
```suggestion
                // Heal 25% damage taken during the last 5 seconds, up to a minimum of 7% of max health.
```
(in case I have misunderstood the spell description, please adjust the text accordingly)

matanshukry: actually I don't think the usage of "up to" with "minimum" is a good one in genereal (which is also what I did). I'll simply use the actual tooltip text:
"heals you for 25% of all damage taken in the last 5 sec, minimum 7.0% of maximum health."

illfated: Fair & good point. So maybe the following will be acceptable?
```suggestion
                // Heals you for 25% of all damage taken in the last 5 sec, minimum 7.0% of maximum health.
```

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -6410,6 +6410,27 @@ void AuraEffect::HandleBattlegroundPlayerPosition(AuraApplication const* aurApp,
         bg->RemovePlayerPosition(target->GetGUID());
 }
 
+void AuraEffect::HandleModArmorPctFromStat(AuraApplication const* aurApp, uint8 mode, bool apply) const
+{
+    // Increase armor by <AuraEffect.BasePoints> % of your <primary stat>
+
```
illfated, line 6416: I could be wrong, but it looks like this blank line is redundant.
```suggestion
```

matanshukry: I wanted to make sure the comment isn't assosiated with the line after it but rather with the entire body;
Any better suggestions than this empty newline?

illfated: Aha, I see what you mean. So the comment does not belong to the next line.
How about outside the brackets, either right before or right after `void AuraEffect:: [...]` ?

matanshukry: Haven't seen any comments outside methods but I do like it;
Moved that code to a different PR, but I'll just do that there.

ghost (approved):
Looks fairly good to me, mostly minor changes (refactoring?)

matanshukry:
> Looks fairly good to me, mostly minor changes (refactoring?)

Yup, still missing the main part about it (in the TODO list). I'll probably have to implement server-side proc effects first and then use it on this one.

sql/updates/world/master/2021_99_99_99_world_dk_blood_mastery.sql
```diff
@@ -0,0 +1,17 @@
+-- 89832 Death Strike Enabler
+DELETE FROM `serverside_spell` WHERE `Id` = 89832 AND `DifficultyID` = 0;
+INSERT INTO `serverside_spell` (`Id`, `DifficultyID`, `SpellName`, `Attributes`, `CastingTimeIndex`, `ProcFlags`, `ProcChance`, `DurationIndex`, `RangeIndex`, `SchoolMask`) VALUES
+(89832, 0, 'Spell Strike Enabler', 0x1d0, 1 /* CastingTimeIndex: 0/0/0 */, 0xAAAA8 /*ProcFlags: Damage Taken */, 100 /*Proc Chance*/, 21 /*Duration: -1/0/-1 */, 1 /*Range: Self Only*/, 0x1 /* Physical */);
```
Shauren, line 4: should have been `Death Strike Enabler`?

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -1652,6 +1652,7 @@ void SpellMgr::LoadSpellProcs()
     }
 
     isTriggerAura[SPELL_AURA_DUMMY] = true;
+    isTriggerAura[SPELL_AURA_PERIODIC_DUMMY] = true;
```
Shauren, line 1655: This should not be needed, it is only there to automatically generate SpellProcEntry data based on SpellInfo alone, without using `spell_proc` table at all

Shauren (changes requested):

src/server/game/Spells/SpellMgr.cpp
matanshukry: Right, although it won't automatically generate it, it will only do so if it has proc flags.
Any reason we don't want that?

Without it I need to create a dedicated spell_proc entry, while using the exact same values that are already on the spell.

sql/updates/world/master/2021_99_99_99_world_dk_blood_mastery.sql
matanshukry: yup

src/server/game/Spells/SpellMgr.cpp
Shauren: This is actually fine

Shauren (approved):

src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -485,31 +498,74 @@ class spell_dk_death_strike : public SpellScript
...
+    }
 
-            void Register() override
-            {
-                OnEffectHitTarget += SpellEffectFn(spell_dk_festering_strike::HandleScriptEffect, EFFECT_2, SPELL_EFFECT_DUMMY);
-            }
-        };
+    void Register() override
+    {
+        OnEffectHitTarget += SpellEffectFn(spell_dk_festering_strike::HandleScriptEffect, EFFECT_2, SPELL_EFFECT_DUMMY);
```
ZaDarkSide, line 566: Here should be EFFECT_1

matanshukry: this PR isn't related to this script, just removed indent.

ZaDarkSide: Yes, but still it's a bug :)

matanshukry: If you want to comment on every bug in the code we're going to be in here for a long time :)

I also wouldn't call it a bug as much as outdated code; it was right at some point in the past, just never got updated.

ZaDarkSide: Maybe I should start doing PRs :) instead of commenting on lines of code :)

matanshukry: Always welcome to do PRs - that would be great :)
But I do appreciate the comments on code! if you ask me - do both xd

PR #26083, opened 2021-02-13
Scripts/Events: Fix achievement in Love is in the air: "Lonely?"

### LINKED ISSUE #24139: [3.3.5] [Achievement] [1291] Lonely
**Description:**
For some reason the achievement can't be completed.

**Current behaviour:**
Currently nothing happen except the basket and the umbrella are spawn.

**Expected behaviour:**

When two characters are joining together with [Romantic Picnic Basket](https://wotlk.evowow.com/?item=34480) they should fall in love. I guess the effect is called like that.
They should get hearth above they heads and then, when they start eating [Buttermilk Delight](https://wotlk.evowow.com/?item=22236) they should get the achievement [Lonely?](https://wotlk.evowow.com/?achievement=1291)

1. Go to Dalaran.
2. Use [Romantic Picnic Basket](https://wotlk.evowow.com/?item=34480)
3. Use second character to join your picnic table.
4. Wait for the hearth effect.
5. Eat [Buttermilk Delight](https://wotlk.evowow.com/?item=22236) 

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

5fa77744e4495a77dea734330d5ae9e113b8a7f1

**Operating system:**

Ubuntu 16.0.4 LTS

Jildor:
**Changes proposed:**

-  Getcaster() is null, we don't need check caster in the script, so remove it

- [X] 3.3.5
- [ ] master

**Issues addressed:**

Closes #24139

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/scripts/Spells/spell_holiday.cpp
+++ b/src/server/scripts/Spells/spell_holiday.cpp
@@ -90,9 +90,6 @@ class spell_love_is_in_the_air_romantic_picnic : public AuraScript
         Cell::VisitWorldObjects(target, searcher, INTERACTION_DISTANCE * 2);
         for (std::list<Player*>::const_iterator itr = playerList.begin(); itr != playerList.end(); ++itr)
         {
-            if (foundSomeone)
-                break;
-
             if (Player* playerFound = (*itr))
             {
                 if (target != playerFound && playerFound->HasAura(GetId()))
@@ -100,6 +97,7 @@ class spell_love_is_in_the_air_romantic_picnic : public AuraScript
                     playerFound->CastSpell(playerFound, SPELL_ROMANTIC_PICNIC_ACHIEV, true);
                     target->CastSpell(target, SPELL_ROMANTIC_PICNIC_ACHIEV, true);
                     foundSomeone = true;
+                    break;
                 }
             }
         }
--- a/src/server/scripts/Spells/spell_holiday.cpp
+++ b/src/server/scripts/Spells/spell_holiday.cpp
@@ -91,7 +91,7 @@ class spell_love_is_in_the_air_romantic_picnic : public AuraScript
         for (std::list<Player*>::const_iterator itr = playerList.begin(); itr != playerList.end(); ++itr)
         {
             if (foundSomeone)
-                continue;
+                break;
 
             if (Player* playerFound = (*itr))
             {
--- a/src/server/scripts/Spells/spell_holiday.cpp
+++ b/src/server/scripts/Spells/spell_holiday.cpp
@@ -46,11 +46,23 @@ class spell_love_is_in_the_air_romantic_picnic : public AuraScript
 {
     PrepareAuraScript(spell_love_is_in_the_air_romantic_picnic);
 
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo(
+        {
+            SPELL_BASKET_CHECK,
+            SPELL_MEAL_PERIODIC,
+            SPELL_MEAL_EAT_VISUAL,
+            SPELL_DRINK_VISUAL,
+            SPELL_ROMANTIC_PICNIC_ACHIEV
+        });
+    }
+
     void OnApply(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
     {
         Unit* target = GetTarget();
         target->SetStandState(UNIT_STAND_STATE_SIT);
-        target->CastSpell(target, SPELL_MEAL_PERIODIC, false);
+        target->CastSpell(target, SPELL_MEAL_PERIODIC);
     }
 
     void OnPeriodic(AuraEffect const* /*aurEff*/)
@@ -61,13 +73,13 @@ class spell_love_is_in_the_air_romantic_picnic : public AuraScript
         // If our player is no longer sit, remove all auras
         if (target->GetStandState() != UNIT_STAND_STATE_SIT)
         {
-            target->RemoveAura(SPELL_ROMANTIC_PICNIC_ACHIEV);
+            target->RemoveAurasDueToSpell(SPELL_ROMANTIC_PICNIC_ACHIEV);
             target->RemoveAura(GetAura());
             return;
         }
 
-        target->CastSpell(target, SPELL_BASKET_CHECK, false); // unknown use, it targets Romantic Basket
-        target->CastSpell(target, RAND(SPELL_MEAL_EAT_VISUAL, SPELL_DRINK_VISUAL), false);
+        target->CastSpell(target, SPELL_BASKET_CHECK); // unknown use, it targets Romantic Basket
+        target->CastSpell(target, RAND(SPELL_MEAL_EAT_VISUAL, SPELL_DRINK_VISUAL));
 
         bool foundSomeone = false;
         // For nearby players, check if they have the same aura. If so, cast Romantic Picnic (45123)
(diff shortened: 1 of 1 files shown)


ghost (approved):
Looks quite good to me.

xParad0x:
Tested in game, looks good @Jildor

xParad0x (approved):

src/server/scripts/Spells/spell_holiday.cpp
```diff
@@ -42,75 +42,65 @@ enum SpellsPicnic
...
 
-            void OnPeriodic(AuraEffect const* /*aurEff*/)
-            {
-                // Every 5 seconds
-                Unit* target = GetTarget();
-                Unit* caster = GetCaster();
+        // If our player is no longer sit, remove all auras
+        if (target->GetStandState() != UNIT_STAND_STATE_SIT)
+        {
+            target->RemoveAura(SPELL_ROMANTIC_PICNIC_ACHIEV);
```
Keader, line 64: RemoveAuraDueToSpell

src/server/scripts/Spells/spell_holiday.cpp
```diff
@@ -42,75 +42,65 @@ enum SpellsPicnic
...
+        }
 
-                // If our player is no longer sit, remove all auras
-                if (target->GetStandState() != UNIT_STAND_STATE_SIT)
-                {
-                    target->RemoveAura(SPELL_ROMANTIC_PICNIC_ACHIEV);
-                    target->RemoveAura(GetAura());
-                    return;
-                }
+        target->CastSpell(target, SPELL_BASKET_CHECK, false); // unknown use, it targets Romantic Basket
```
Keader, line 69: triggered = false is default, so can be removed here and below

src/server/scripts/Spells/spell_holiday.cpp
```diff
@@ -42,75 +42,65 @@ enum SpellsPicnic
...
+        }
 
-            void Register() override
-            {
-                AfterEffectApply += AuraEffectApplyFn(spell_love_is_in_the_air_romantic_picnic_AuraScript::OnApply, EFFECT_0, SPELL_AURA_PERIODIC_DUMMY, AURA_EFFECT_HANDLE_REAL);
-                OnEffectPeriodic += AuraEffectPeriodicFn(spell_love_is_in_the_air_romantic_picnic_AuraScript::OnPeriodic, EFFECT_0, SPELL_AURA_PERIODIC_DUMMY);
-            }
-        };
+        if (!foundSomeone && target->HasAura(SPELL_ROMANTIC_PICNIC_ACHIEV))
+            target->RemoveAura(SPELL_ROMANTIC_PICNIC_ACHIEV);
```
Keader, line 96: RemoveAuraDueToSpell too

src/server/scripts/Spells/spell_holiday.cpp
```diff
@@ -42,75 +42,65 @@ enum SpellsPicnic
     SPELL_ROMANTIC_PICNIC_ACHIEV    = 45123, // Romantic Picnic periodic = 5000
 };
 
-class spell_love_is_in_the_air_romantic_picnic : public SpellScriptLoader
+class spell_love_is_in_the_air_romantic_picnic : public AuraScript
 {
-    public:
-        spell_love_is_in_the_air_romantic_picnic() : SpellScriptLoader("spell_love_is_in_the_air_romantic_picnic") { }
+    PrepareAuraScript(spell_love_is_in_the_air_romantic_picnic);
```
Keader, line 47: missing validate spells

jackpoz:
> * Getcaster() is null, we don't need check caster in the script, so remove it

I don't totally understand this sentence. Is GetCaster() supposed to be null ? is that how things are supposed to work ? how does the script change when not using GetCaster() ? what is it used instead ?

Jildor:
> > * Getcaster() is null, we don't need check caster in the script, so remove it
> 
> I don't totally understand this sentence. Is GetCaster() supposed to be null ? is that how things are supposed to work ? how does the script change when not using GetCaster() ? what is it used instead ?

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/GameObject/GameObject.cpp#L768-L781
the picnic gob (187267) is goober and when player click, player get the aura 45102
the owner of the aura is the player, but for some reason, GetCaster() here return null (also return null GetGObjCaster()), you can test with worldobject GetOwner() and you get player clicker.
in the script we don't need GetCaster() because we can use target and playerfound for spellcast (is the current use of GetCaster in this script)

src/server/scripts/Spells/spell_holiday.cpp
Jildor: thx, done ;)

src/server/scripts/Spells/spell_holiday.cpp
Jildor: thx, done ;)

src/server/scripts/Spells/spell_holiday.cpp
Jildor: thx, done ;)

src/server/scripts/Spells/spell_holiday.cpp
Jildor: thx, done ;)

jackpoz:
@Keader anything else to do or does this look good ?

Keader:
looks good to me

src/server/scripts/Spells/spell_holiday.cpp
```diff
@@ -42,75 +42,77 @@ enum SpellsPicnic
...
+        // For nearby players, check if they have the same aura. If so, cast Romantic Picnic (45123)
+        // required by achievement and "hearts" visual
+        std::list<Player*> playerList;
+        Trinity::AnyPlayerInObjectRangeCheck checker(target, INTERACTION_DISTANCE*2);
+        Trinity::PlayerListSearcher<Trinity::AnyPlayerInObjectRangeCheck> searcher(target, playerList, checker);
+        Cell::VisitWorldObjects(target, searcher, INTERACTION_DISTANCE * 2);
+        for (std::list<Player*>::const_iterator itr = playerList.begin(); itr != playerList.end(); ++itr)
+        {
+            if (foundSomeone)
+                continue;
```
jackpoz, line 94: shouldn't this be "break;" instead of "continue;" ?

src/server/scripts/Spells/spell_holiday.cpp
```diff
@@ -42,75 +42,77 @@ enum SpellsPicnic
...
-                            caster->CastSpell(*itr, SPELL_ROMANTIC_PICNIC_ACHIEV, true);
-                            caster->CastSpell(target, SPELL_ROMANTIC_PICNIC_ACHIEV, true);
-                        }
-                        foundSomeone = true;
-                        // break;
-                    }
+                    playerFound->CastSpell(playerFound, SPELL_ROMANTIC_PICNIC_ACHIEV, true);
+                    target->CastSpell(target, SPELL_ROMANTIC_PICNIC_ACHIEV, true);
+                    foundSomeone = true;
                 }
```
jackpoz, line 103: or even better, how about adding a "break;" here after foundSomeone so we get out of the loop as soon as we are done ?

src/server/scripts/Spells/spell_holiday.cpp
Jildor: Yes, it seems more optimal, thx

src/server/scripts/Spells/spell_holiday.cpp
Jildor: ok, thx

jackpoz:
just to be sure: is the achievement supposed to support 1 target player only ? https://github.com/TrinityCore/TrinityCore/pull/26083/files#diff-1b0804f59f113e7f4bf3733cba9666d276f450f70956a2790edf3a7d21f5df54R99

Jildor:
I do not have data on how it would work with more players in retail, I don't know.
It seems to me little relevant in short, since the achievement would be completed anyway

jackpoz:
Thanks for the PR :)

PR #26084, opened 2021-02-13
Core/Auras: Implemented aura to increase armor by percentage of primary stat

matanshukry:
**Changes proposed:**

-  Implemented aura to increase armor by percentage of primary stat

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -7740,7 +7740,7 @@ void Player::_ApplyItemBonuses(Item* item, uint8 slot, bool apply)
 
     if (uint32 armor = proto->GetArmor(itemLevel))
     {
-        HandleStatFlatModifier(UNIT_MOD_ARMOR, BASE_VALUE, float(armor), apply);
+        HandleStatFlatModifier(UNIT_MOD_ARMOR, TOTAL_VALUE, float(armor), apply);
         if (proto->GetClass() == ITEM_CLASS_ARMOR && proto->GetSubClass() == ITEM_SUBCLASS_ARMOR_SHIELD)
             SetUpdateFieldValue(m_values.ModifyValue(&Player::m_activePlayerData).ModifyValue(&UF::ActivePlayerData::ShieldBlock), apply ? int32(armor * 2.5f) : 0);
     }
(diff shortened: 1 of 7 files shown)


src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -6403,6 +6403,24 @@ void AuraEffect::HandleBattlegroundPlayerPosition(AuraApplication const* aurApp,
...
+
+    // only players have primary stats
+    Player* player = aurApp->GetTarget()->ToPlayer();
+    if (!player)
+        return;
+
+    Stats primaryStat = player->GetPrimaryStats();
+    int32 armorAmount = CalculatePct(player->GetStat(primaryStat), GetAmount());
+    if (apply)
+        for (uint8 x = SPELL_SCHOOL_NORMAL; x < MAX_SPELL_SCHOOL; ++x)
```
Shauren, line 6420: This implementation is incorrect/incomplete for two reasons
* Armor is UNIT_MOD_ARMOR, not UNIT_MOD_RESISTANCE_*
* Armor bonus must update dynamically when stats change - requires calling UpdateArmor() in Player::UpdateStats and moving this calculation inside UpdateArmor() (StatSystem.cpp)
* The stat to use is  `MiscValueA != -2 ? Stats(MiscValueA) : player->GetPrimaryStats()`

src/server/game/Entities/Player/Player.cpp
```diff
@@ -28867,3 +28867,34 @@ uint8 Player::GetItemLimitCategoryQuantity(ItemLimitCategoryEntry const* limitEn
 
     return limit;
 }
+
+Stats Player::GetPrimaryStats() const
+{
+    switch (getClass())
```
Shauren, line 28873: I will provide a better implementation for this function that does not hardcode classes and specializations

Shauren (changes requested):

src/server/game/Spells/Auras/SpellAuraEffects.cpp
matanshukry: Those are 3 reasons

matanshukry: Verify I did it correctly, but know tihs:

The client doesn't seem to get the updated armor. When looking on the character window, with equipment and stats, armor is always 5, even when ironfur is on and it adds around 180 armor.

I also verified the armor is being added in the server by debugging an attack.

Shauren: Not good, it should be showing up on client

matanshukry: Found the issue, sort of. Seems like the client can only see the `Resistances` and not the `BonusResistanceMods`.
Also, specifically with `IronFur` which has this effect it will update the `Resistances`.

I've changed the PR to reflect that, so that `SPELL_AURA_MOD_ARMOR_PCT_FROM_STAT` will now affect baseValue rather than totalValue.

Shauren (approved):

PR #26085, opened 2021-02-13
Core/Auras: implement mana cost modification aura

matanshukry:
**Changes proposed:**

- Implemented mana cost modiciation aura

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game


illfated:
PR title typo correction: **Core/Aura: implement mana cost modification aura**

Shauren:
Merge 331b9ccd6a3be12e92e7c2aab7b793739b405d2f and remove changes from UpdateFields files, I added a note explaining those are not meant to be touched

src/server/game/Spells/SpellInfo.h
```diff
@@ -432,6 +432,9 @@ struct SpellPowerCost
 {
     Powers Power;
     int32 Amount;
+
```
Shauren, line 435: Is there a reason for this change?

matanshukry: I had a piece of code locally removed that used it and forgot to return it.
Returned now.

src/server/game/Spells/SpellInfo.cpp
```diff
@@ -3901,7 +3906,9 @@ std::vector<SpellPowerCost> SpellInfo::CalcPowerCost(Unit const* caster, SpellSc
             powerCost += CalculatePct(powerCost, (*i)->GetAmount());
         }
 
-        if (power->PowerType == POWER_HEALTH)
+        if (power->PowerType == POWER_MANA)
+            powerCost = std::max(0.0f, (float)powerCost * (1.0f + caster->GetManaCostMultiplier())); // Ensure >= 0
```
Shauren, line 3910: There is no requirement that cost must be >= 0, costs can actually be negative (and grant power)

This probably also makes the change above (comparison against -1) incorrect

matanshukry: So if a resto druid uses heart of the wild together with innervate, and then cast a spell (e.g. regrowth with cost 240 mana), he will get 120 extra mana ?

Shauren: Resto druid does not have access to Restoration Affinity talent so they cannot benefit from cost reduction with Heart of the Wild

matanshukry: Balance druid*

Shauren: Nonhealers cannot be given innervate

matanshukry: Well I don't know the game well enough to give other examples.
Since the idea here is to "Reduce mana", it doesn't sound logical to give the player mana from it.

Do you have an example where multiple spells can be applied and the combined mana reduction is above 100%?

PR #26092, opened 2021-02-14
Core/Maps: Use a fixed offset instead of full collision height when retrieving floor Z

jackpoz:
**Changes proposed:**

-  Use a fixed offset 0.5f instead of full collision height when retrieving floor Z as a full collision height ended up on the floor above a few times. It makes more sense to cast the ray just a bit higher up than using a full collision height (which by default is 2.0f)

- [x] 3.3.5
- [ ] master

**Issues addressed:**

- Fixes https://github.com/TrinityCore/TrinityCore/issues/21969#issuecomment-467837750
- Replaces https://github.com/TrinityCore/TrinityCore/pull/25998

**Tests performed:**

https://github.com/TrinityCore/TrinityCore/issues/21969#issuecomment-467837750

- [ ] Test it more around the world
- [ ] Test the cases of https://github.com/TrinityCore/TrinityCore/pull/21322
- [ ] Test the cases of https://github.com/TrinityCore/TrinityCore/pull/21125

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -3554,7 +3554,7 @@ float WorldObject::GetFloorZ() const
 {
     if (!IsInWorld())
         return m_staticFloorZ;
-    return std::max<float>(m_staticFloorZ, GetMap()->GetGameObjectFloor(GetPhaseMask(), GetPositionX(), GetPositionY(), GetPositionZ() + GetCollisionHeight()));
+    return std::max<float>(m_staticFloorZ, GetMap()->GetGameObjectFloor(GetPhaseMask(), GetPositionX(), GetPositionY(), GetPositionZ() + Z_OFFSET_FIND_HEIGHT));
 }
 
 float WorldObject::GetMapWaterOrGroundLevel(float x, float y, float z, float* ground/* = nullptr*/) const
@@ -3567,7 +3567,7 @@ float WorldObject::GetMapWaterOrGroundLevel(float x, float y, float z, float* gr
 float WorldObject::GetMapHeight(float x, float y, float z, bool vmap/* = true*/, float distanceToSearch/* = DEFAULT_HEIGHT_SEARCH*/) const
 {
     if (z != MAX_HEIGHT)
-        z += GetCollisionHeight();
+        z += Z_OFFSET_FIND_HEIGHT;
 
     return GetMap()->GetHeight(GetPhaseMask(), x, y, z, vmap, distanceToSearch);
 }
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -2457,7 +2457,7 @@ float Map::GetWaterOrGroundLevel(uint32 phasemask, float x, float y, float z, fl
     if (const_cast<Map*>(this)->GetGrid(x, y))
     {
         // we need ground level (including grid height version) for proper return water level in point
-        float ground_z = GetHeight(phasemask, x, y, z + collisionHeight, true, 50.0f);
+        float ground_z = GetHeight(phasemask, x, y, z + Z_OFFSET_FIND_HEIGHT, true, 50.0f);
         if (ground)
             *ground = ground_z;
 
--- a/src/server/shared/SharedDefines.h
+++ b/src/server/shared/SharedDefines.h
@@ -24,6 +24,7 @@
 #include "SmartEnum.h"
 
 float const GROUND_HEIGHT_TOLERANCE = 0.05f; // Extra tolerance to z position to check if it is in air or on ground.
+constexpr float Z_OFFSET_FIND_HEIGHT = 0.5f;
 
 enum SpellEffIndex : uint8
 {


Jildor:
@jackpoz after this change you can fall under textures more frecuently, by example when you use blink in a small elevation of the ground, por ejemplo:
```
.go xyz 604.184937 139.763580 394.707001 649 6.246275
cast blink
```

```
.go xyz 627.301392 156.685043 140.341751 649 3.227158
cast blink
```

example after this change:
https://imgur.com/SFK4GHy

example before this change:
https://imgur.com/3AfO8PW

I suggest revert this change and looking for a better solution

jackpoz:
Is that the map with mmaps disabled ? Blink uses mmaps so no wonder it doesn't work. We should probably just disable blink in places without mmaps.

Jinnaix:
image

Jildor:
yep, seems on maps with mmaps disabled, I just tried warsong gulch for example and I have not seen problems, but blink should work in places without mmaps, because is useful for avoid spells, not a solution disable blink

Aokromes:
maybe we can change instance wide disable mmaps for creature wide for the ones on top of arena.

Jildor:
add CREATURE_FLAG_EXTRA_IGNORE_PATHFINDING to all creatures of those two maps and remove disable mmaps for Trial of the crusader and Trial of the champion could be a way?

Aokromes:
that's what i said, but only for the ones on top of the arena, not the ones on bottom, no need of ignore pathfinding there.

jackpoz:
Could anyone do that and send a fix ?

Jildor:
I've been doing some tests removing disable mmaps in trial of the crusader and it doesn't seem like a valid way, when cast blink you don't move, same with charge or intercept (warrior), you can't use it because not path avaliable.

It doesn't seem like the best result to me, I keep thinking that maybe reverting this commit and looking for a better solution is a good option

jackpoz:
Blink has been using mmaps since March 2020 (1 year ago) with https://github.com/TrinityCore/TrinityCore/pull/24232 , charge has been using mmaps for years. Not having mmaps working properly in a specific place does of course create issues.

Steps to fix your issue:
- re-enable mmaps in trial of the crusader
- implement whatever missing feature in mmaps to have trial of the crusader work 100%

Focusing on trying to get blink/charge to work properly without mmaps is a wasted effort (that people chose to work on for lot of years when we already have mmaps for that exact reason).

Ovahlord:
No more temp/hackfixes on that. It only boils down to create new issues and people like me have to clean that mess up later. Pathfinding is currently in a somewhat solid state without major duc-taping so let's keep it that way. If you wanna solve that problem properly then implement gameobject pathfinding, especially for destructuble gameobjects. That's the only issue in that regard.

jackpoz:
@Jildor the cases linked in the PR description imo are just as bad as yours, i.e. Karazhan last boss where the ceiling is picked instead of the floor as Z.

Even if restoring the previous code works in your test cases, there will always be other cases where it causes issues.

You could try with 1.0f instead of 0.5, testing your ToC case and the cases this PR referenced and see if that has a better result (what is the collision height for a player ?)

We can take the chance to finally get ToC fixed in a better way, i.e. including the GameObject floor in the mmtile and generate another tile with whatever other status the gameobject goes to and swap tiles at runtime. Bonus points for keeping tile references values the same.

Jildor:
Tested with Z_OFFSET_FIND_HEIGHT = 1.0f;
About this cases:
https://github.com/TrinityCore/TrinityCore/pull/26092#issuecomment-798882900
- first case I can't reproduce it, rarely I can fall under arena (I found one case here: .go xyz 546.304871 103.062332 394.680878 649 4.946443)
- second case I can reproduce it, I can fall in same place (not relevant that case because not a boss combat zone)

Malchezaar issue: https://github.com/TrinityCore/TrinityCore/issues/21969#issuecomment-467837750 still working fine

I don't see problems in warsong gulch map

other issue cases about tanaris cavern of time or orgrimmar dummies seems fine


Also I tested with Z_OFFSET_FIND_HEIGHT = 1.5f;
all above cases seems fine I can't fall in trial of crusader cases, malchezaar seems fine, etc.
I'm not sure what an already higher value like 1.5f could affect, the truth is, I don't know what else to check, but with the 1.0f value it's pretty good

Jildor:
Another case with Z_OFFSET_FIND_HEIGHT = 0.5f; (default PR)
```
.gm on
.go boss lana
.mod hp 100000000
.gm off and agro lana'thel
go to the stairs and use blink when gameobject door (red smoke) is active
```
example: https://imgur.com/WwFYmRm

this doesn't happens before this PR and with Z_OFFSET_FIND_HEIGHT = 1.0f; rarely occurs

Maybe I'm posing another nonsense, but leaving collisionHeight as it was before this PR and limiting it to a maximum value of for example 1.5f could it be a valid way?

jackpoz:
@Jildor about the stair case, increasing the height for that case is not a good thing because the way the blink destination point is found using map/vmap only is "take XYZ of current position , increase XY by the blink distance, then use the same Z as the start position and hope that the slope is not too steep that we went underground. Add to Z some random 0.5/1.0/1.5 hoping none will find a case that requires 5.0".

Adding Collision Height is wrong because the Z we should add doesn't depend on the player height but depends on the Z difference between the start and the end point (which MMAPS can find much better as we walk on the polygons, we don't just randomly add a fixed distance to XY and hope for the best).

Of course it looks like Collision Height works better ingame, but that's just because you didn't find yet a case where it doesn't work. Pretty sure you will find a whole lot of cases where this whole logic fails if you disable mmaps completely.

The only safe way to not fall underground is to use mmaps with raycast. Maybe we could increase the Z by some extra safety offset in those places where mmaps are disabled, so we keep the clean solution that we want and avoid having players falling in ToC.

ccrs:
but mmaps are active in ICC, right?

jackpoz:
".mmac loc" is the way to go to know if the tile is loaded at a specific location

Jildor:
I guess the ccrs question is for the case I have detailed of the stairs in lana'thel room(https://github.com/TrinityCore/TrinityCore/pull/26092#issuecomment-804781384), it's important to remember that happens when the door (gameobject red smoke) is active and obstructs the player.

That is why I have made that suggestion in that same comment, because I have seen that there may be cases in maps with active mmaps after this PR.

I leave the result of ".mmap loc" of that specific case in case it is necessary for some reason
```
mmap tileloc:
6312326.mmtile
tileloc [26, 23]
Calc   [11, 12]
Dt     [11,12]
```

Jildor:
Also can happens same as ICC stairs case in battleground Strand of the Ancients with destructible doors
Tested with default PR ( Z_OFFSET_FIND_HEIGHT = 0.5f;), before this PR I can't reproduce it.
https://imgur.com/a/ZnvdA3d
https://imgur.com/DqnKZuv
https://imgur.com/vhSHDrA

I leave a location of one of this examples, but I can reproduce it in all doors
```
Map: 607 (Strand of the Ancients) Zone: 4384 (Strand of the Ancients) Area: 4604 (Gate of the Red Sun) Phase: 129
X: 1236.454590 Y: -208.137436 Z: 54.091614 Orientation: 3.469470
grid[34,31]cell[2,4] InstanceID: 1
 ZoneX: 57.097488 ZoneY: 55.645477
GroundZ: 54.091637 FloorZ: 54.091637 MinHeight: -500.000000 Have height data (Map: 1 VMap: 1 MMap: 1)

mmap tileloc:
6072932.mmtile
tileloc [32, 29]
Calc   [04, 02]
Dt     [04,02]
```

jackpoz:
One thing to keep in mind is that Z_OFFSET_FIND_HEIGHT was defined in a single place to make it easy to tune, compared to the old + 0.5f used in the past.

If you have had better results with 1.0f and still cannot reproduce the Karazhan boss issue, please open a PR with the new value. We will implement everything else that has been discussed here in a later time.

Faq:
Any reason why this conversation was not in https://github.com/TrinityCore/TrinityCore/discussions ?

illfated:
Most likely some form of laziness and ease of access to the relevant PR above (main discussion topic).

PR #26093, opened 2021-02-15
Scripts/Spells: Ferocious bite implemented

matanshukry:
**Changes proposed:**

- Implemented Ferocious bite

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

**Dependant On:**

#26097


illfated:
Utterly and purely cosmetic, but for the record: "Ferocious Bite" is the name of the spell, so you could choose to let "implemented" or "implement" be the only non-capitalized word in the title.

src/server/scripts/Spells/spell_druid.cpp
```diff
@@ -1510,6 +1510,40 @@ class spell_dru_wild_growth : public SpellScriptLoader
...
+        float ratio = -(float)powerGain / eff1->BasePoints; // map from [0..BasePoints] to [0..1]
+        dmg *= 1.0f + ratio;
+
+        // Set the damage to m_damage on spell
+        SetHitDamage((int32)dmg);
+    }
+
+    void Register() override
+    {
+        OnEffectHitTarget += SpellEffectFn(spell_dru_ferocious_bite::HandleHitTarget, EFFECT_1, SPELL_EFFECT_POWER_BURN);
```
Shauren, line 1543: This script is incrrect: EFFECT_1 (power burn) hits the caster so `GetHitDamage()` and `SetHitDamage()` are damage values dealt to caster from burned power

src/server/game/Spells/Spell.cpp
```diff
@@ -5380,7 +5380,15 @@ SpellCastResult Spell::CheckCast(bool strict, uint32* param1 /*= nullptr*/, uint
             case SPELL_EFFECT_POWER_DRAIN:
             {
                 // Can be area effect, Check only for players and not check if target - caster (spell can have multiply drain/burn effects)
-                if (m_caster->GetTypeId() == TYPEID_PLAYER)
+                // effect target can be caster and can be different than spell target (e.g. Ferocious Bite)
```
Shauren, line 5383: Don't you have a separate PR for this part?

Shauren (changes requested):
Required hooks without hacking up CheckCast
`OnEffectLaunchTarget` for SPELL_EFFECT_POWER_BURN for caster - manually apply EFFECT_1 value from 102543 to GetEffectValue() (up to this point in calculation this will be your "max extra consumed power"). Damage multiplier will be `1.0f + std::min<float>(GetPower(POWER_ENERGY), maxConsumedExtraEnergy) / maxConsumedExtraEnergy`  and store this calculated value as a script class member (float _damageMultiplier). You do need make it "as if" ModifyPower was used without using it, we need it to actually use default effect handler later for proper packet contents

`OnEffectHitTarget` for SPELL_EFFECT_POWER_BURN for caster - manually apply EFFECT_1 value from 102543 to GetEffectValue()/SetEffectValue() and let it do its default thing later

`OnEffectHitTarget` for SPELL_EFFECT_SCHOOL_DAMAGE applying the multiplier saved from `OnEffectLaunchTarget` (with SetHitDamage)

src/server/game/Spells/Spell.cpp
matanshukry: I do, but this PR depends on it.
That's wyh I added the "Dependant On:" part in the OP and linked the other issue.

src/server/scripts/Spells/spell_druid.cpp
matanshukry: Thanks for the thorough explaination! :)

Shauren (approved):

PR #26094, opened 2021-02-15
Core/Quests: NPC shows player Quest Window for singular Quest

### LINKED ISSUE #22010: Auto-accepting quests does not work correctly
**Description:**
When accessing the NPC, auto-accepting quests should take place in the starting locations, but it does not work correctly, when the address is displayed, a menu with the name of the quest is displayed, and after it is clicked, an auto-acceptance of the quest takes place.
And it should work as follows: when accessing, the quest itself opens immediately and its auto-acceptance occurs.
As an example, the quest 28806 NPC951
wowscrnshot_060718_190234
wowscrnshot_060718_190329

And should work like this:
wowscrnshot_060718_190536

1. Step 1 Click on the NPC 951 
2. Step 2 Click quest name and quest auto-accepted, but it's not right

**Branch(es):**
master

**TC rev. hash/commit:** 
https://github.com/TrinityCore/TrinityCore/commit/36ff7668823f56d81e422c4baeca43b8a32ba810

**Operating system:**
Windows 10 Pro

matanshukry:
**Changes proposed:**

-  When player interacts with npc it will show the quest window if there's only one quest

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #22010

**Tests performed:**

- [x] It builds
- [x] Tested in-game

- [] WorldPackets::Quest::QuestGiverRequestItems.StatusFlags currently we only use `canComplete ? 0xFF : 0xFD;`. However even when the quest isn't complete, player will get the message "You are already on that quest". `AutoLaunched` seems to not have a direct influence on that. I'm guessing there are other flags we're missing

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -15139,7 +15139,7 @@ void Player::SendPreparedQuest(WorldObject* source)
         // Auto open
         if (Quest const* quest = sObjectMgr->GetQuestTemplate(questId))
         {
-            if (qmi0.QuestIcon == GOSSIP_ICON_INTERACT_1)
+            if (qmi0.QuestIcon == 4)
                 PlayerTalkClass->SendQuestGiverRequestItems(quest, source->GetGUID(), CanRewardQuest(quest, false), true);
             // Send completable on repeatable and autoCompletable quest if player don't have quest
             /// @todo verify if check for !quest->IsDaily() is really correct (possibly not)
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -15139,33 +15139,24 @@ void Player::SendPreparedQuest(WorldObject* source)
         // Auto open
         if (Quest const* quest = sObjectMgr->GetQuestTemplate(questId))
         {
-            if (qmi0.QuestIcon == 4)
-            {
+            if (qmi0.QuestIcon == GOSSIP_ICON_INTERACT_1)
                 PlayerTalkClass->SendQuestGiverRequestItems(quest, source->GetGUID(), CanRewardQuest(quest, false), true);
-                return;
-            }
             // Send completable on repeatable and autoCompletable quest if player don't have quest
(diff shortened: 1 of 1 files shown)


illfated:
Since the PR title has been cut short, maybe rephrase it to make it quicker to read:

**Core/Quests: NPC shows Player Quest Window for singular Quest**

(This is merely a suggestion, use how much or little as you want. The main idea is to simplify the title as much as possible.)

Shauren:
All of these changes are unneccessary, all that you had to do was remove `source->GetTypeId() != TYPEID_UNIT || source->ToUnit()->HasNpcFlag(UNIT_NPC_FLAG_GOSSIP)` from `Player::SendPreparedQuest`

Shauren (changes requested):
All of these changes are unneccessary, all that you had to do was remove `source->GetTypeId() != TYPEID_UNIT || source->ToUnit()->HasNpcFlag(UNIT_NPC_FLAG_GOSSIP)` from `Player::SendPreparedQuest`

matanshukry:
> All of these changes are unneccessary, all that you had to do was remove `source->GetTypeId() != TYPEID_UNIT || source->ToUnit()->HasNpcFlag(UNIT_NPC_FLAG_GOSSIP)` from `Player::SendPreparedQuest`

Yup "sadly" (not really) this is better; my bad.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -15139,33 +15139,24 @@ void Player::SendPreparedQuest(WorldObject* source)
         // Auto open
         if (Quest const* quest = sObjectMgr->GetQuestTemplate(questId))
         {
-            if (qmi0.QuestIcon == 4)
-            {
+            if (qmi0.QuestIcon == GOSSIP_ICON_INTERACT_1)
```
Shauren, line 15142: I would like you to revert this change - GOSSIP_ICON_INTERACT_1 is not the correct enum to use (there is no defined enum for this field)

Shauren (approved):

Issue #24677, opened 2020-05-22
[Master] Power regeneration issue with SPELL_AURA_MOD_POWER auras


Loopeless:
**Description:**

Basically, when you have auras that decrease power regeneration such as SPELL_AURA_MOD_POWER_REGEN_PERCENT / SPELL_AURA_MOD_POWER_REGEN as you are regenerating power, look closely at your power bar and see that it is going back and forth between values.

To be more specific it looks that it's only the issue with focus and energy powers. In case if those auras are supposed to increase regeneration, then there seems not to be an issue. However, if they are supposed to decrease regeneration then you start seeing this issue to appear.

Somewhere in legion or BFA they (maybe even before that) they added auras that handle power regeneration differently for each specialization. In this example let's use a hunter.

Hunter in survival and marksman has about 50-70% decreased focus regeneration because playstyle is different etc and BeastMastery has regular focus regeneration (no increase nor decrease and it works fine).

I also discovered that the power regeneration is handled client-wise in most cases and there has to be some flag that is missing atm or incorrectly used.

Here is a flag that is handling this visual part for the client at the moment.

UNIT_FIELD_POWER_REGEN_FLAT_MODIFIER

The problem with UNIT_FIELD_POWER_REGEN_FLAT_MODIFIER seems to appear only when you are trying to decrease its value more than the default value. (Which is with specs that have different power regeneration as I already explained).

It is worth mentioning that it technically works fine server-wise but the only problem is the visual part of it and it is a huge issue because a person can't normally determine current power value.

**Expected behavior:**

Regeneration timers should match on both client and server-side when they are decreased or increased, generally speaking.

**Steps to reproduce the problem:**

1. Create a hunter and put it to the max level (120) just for testing purposes.
2. Then you can either use abilities or I suggest switching back and forth between specialization to reset focus to 0 to see it regenerate. (Issue only appear with marksman and survival)
3. Watch the power regeneration bar (for focus) and how it starts to have weird visual problems.

**Branch(es):** master

**TC rev. hash/commit:** 3e833739

**Operating system:** Win10 64x OS

Loopeless:
Maybe it could be an issue with packet `SMSG_POWER_UPDATE`
The one that gets triggered within `SetPower `function (This one is used for regeneration functionality)

`void Unit::SetPower(Powers power, int32 val)`

This part:

```
    if (IsInWorld())
    {
        WorldPackets::Combat::PowerUpdate packet;
[...]

PR #26097, opened 2021-02-15
Core/Spells: POWER_BURN should not cancel when target isn't caster

matanshukry:
**Changes proposed:**

-  POWER_BURN should not cancel when target isn't caster

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

- [ ] Damage isn't applied at the moment. `Spell::EffectPowerBurn` will call `effectInfo->CalcValueMultiplier`, but it uses amplitutde which is 0 in the case of Ferocious Bite. Even if it wasn't, the result is multiplied by the amount of energy taken (25), rather than by the amount of damage dealt.

Not directly related (I might just open different issues for them):
- [ ] DisplayPower: seems to be like a single power that is displayed but you can have multiple such as energy + combopoints. Perhps we should convert it to a list rather than single? We could also use GetMaxPower(power) > 0, I believe.


Robingadko:
>  Incorrect match between server and client energy.

Looks like there's already an issue #24677 for that. The same problem also for hunter focus and pet powers.

matanshukry:
> > Incorrect match between server and client energy.
> 
> Looks like there's already an issue #24677 for that. The same problem also for hunter focus and pet powers.

Cool, I'll remove it from th todo list here and take a closer look at that issue then. Thanks!

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -534,14 +534,6 @@ void Spell::EffectSchoolDMG(SpellEffIndex effIndex)
             }
             case SPELLFAMILY_DRUID:
             {
-                // Ferocious Bite
-                if (m_caster->GetTypeId() == TYPEID_PLAYER && m_spellInfo->SpellFamilyFlags[3] & 0x1000)
```
Shauren, line 538: Belongs in different PR

src/server/game/Spells/Spell.cpp
```diff
@@ -5380,7 +5380,15 @@ SpellCastResult Spell::CheckCast(bool strict, uint32* param1 /*= nullptr*/, uint
             case SPELL_EFFECT_POWER_DRAIN:
             {
                 // Can be area effect, Check only for players and not check if target - caster (spell can have multiply drain/burn effects)
-                if (m_caster->GetTypeId() == TYPEID_PLAYER)
```
Shauren, line 5383: Just completely remove all of it for both SPELL_EFFECT_POWER_BURN and SPELL_EFFECT_POWER_DRAIN

Shauren (changes requested):

src/server/game/Spells/SpellEffects.cpp
matanshukry: Is it? the idea was this code depends on the fact we're not handling the power_burn, and this PR fixed it so the code isn't needed anymore.
no?

Shauren: To me this belongs to ferocious bite PR

src/server/game/Spells/Spell.cpp
Shauren: Oh, I wanted you to just remove the entire switch case for those two effects

PR #26100, opened 2021-02-16
Core/Unit: Improve Glancing Blow calculation

Metalaka:
Hi

**Changes proposed:**

-  Add Glancing Blow vs mob with the same level as the attacker
-  Set a minimum of 1% damage reduction for glancings

- [x] 3.3.5
- [ ] master

**Issues addressed:**

**Tests performed:**

hash: 2f9e38612f90f2b438df26eb10f465e2ac53b1f7
I've just read the code and found that a 80 vs 80 can't do Glancing Blow, not tested at all.

I don't know yours coding standards, feel free to comment/edit the PR ;)

**Proof:**

Chance to glancing vs same level: 6%
http://www.worldoflogs.com/reports/k9lp4yw68u6s1asl/details/15/?s=34&e=196

Damage reduction vs same level: 1 - 10%
http://www.worldoflogs.com/reports/k9lp4yw68u6s1asl/xe/?x=targetName%3D%22The+Damned%22%0D%0Aand+%28%0D%0A%28%0D%0AsourceName%3D%22Oranoss%22+and+%28isGlancing+%3D+TRUE%0D%0Aor+fullType+%3D+SWING_DAMAGE%29%0D%0A%29%0D%0Aor+spell%3D%22Sunder+Armor%22%0D%0Aor+spell%3D%22Faerie+Fire%22%0D%0A%29%0D%0A%0D%0A

NB: I've checked the formula against 81-83 btw, everything seems like in the logs.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -1324,7 +1324,7 @@ void Unit::CalculateMeleeDamage(Unit* victim, CalcDamageInfo* damageInfo, Weapon
             // against boss-level targets - 24% chance of 25% average damage reduction (damage reduction range : 20-30%)
             // against level 82 elites - 18% chance of 15% average damage reduction (damage reduction range : 10-20%)
             int32 const reductionMax = leveldif * 10;
-            int32 const reductionMin = std::min(1, reductionMax - 10);
+            int32 const reductionMin = std::max(1, reductionMax - 10);
             float reducePercent = 1.f - irand(reductionMin, reductionMax) / 100.0f;
 
             for (uint8 i = 0; i < MAX_ITEM_PROTO_DAMAGES; ++i)
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -1324,7 +1324,7 @@ void Unit::CalculateMeleeDamage(Unit* victim, CalcDamageInfo* damageInfo, Weapon
             // against boss-level targets - 24% chance of 25% average damage reduction (damage reduction range : 20-30%)
             // against level 82 elites - 18% chance of 15% average damage reduction (damage reduction range : 10-20%)
             int32 const reductionMax = leveldif * 10;
-            int32 const reductionMin = reductionMax - 10;
+            int32 const reductionMin = std::min(1, reductionMax - 10);
             float reducePercent = 1.f - irand(reductionMin, reductionMax) / 100.0f;
 
             for (uint8 i = 0; i < MAX_ITEM_PROTO_DAMAGES; ++i)
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -2171,7 +2171,7 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(Unit const* victim, WeaponAttackTy
     }
 
     // 4. GLANCING
-    // Max 40% chance to score a glancing blow against mobs that are higher level (can do only players and pets and not with ranged weapon)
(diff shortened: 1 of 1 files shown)


offl:
Shouldn't we update comments too?

Faq:
Your rev hash is for master branch, but you point to the 335 branch.

Metalaka:
> Your rev hash is for master branch, but you point to the 335 branch.

Updated, well seen

Metalaka:
> Shouldn't we update comments too?

I'm not used to put toon of comments inside my code…
We can add the expected behavior for level 80 & 81 of course, if you find it useful

offl:
I mean aren't existing comments outdated now?

Metalaka:
Nope, comments are still goods.

Actually, the behavior cover glancings for `level +1/2/3` (81-83/boss). Comments explains levels 82/83.
The PR add glancings for same levels (eg: 80vs80).

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -1316,13 +1316,15 @@ void Unit::CalculateMeleeDamage(Unit* victim, CalcDamageInfo* damageInfo, Weapon
...
+            if (leveldif == 0)
+                leveldif = 1;
             if (leveldif > 3)
                 leveldif = 3;
 
             // against boss-level targets - 24% chance of 25% average damage reduction (damage reduction range : 20-30%)
             // against level 82 elites - 18% chance of 15% average damage reduction (damage reduction range : 10-20%)
             int32 const reductionMax = leveldif * 10;
-            int32 const reductionMin = reductionMax - 10;
+            int32 const reductionMin = std::min(1, reductionMax - 10);
```
jackpoz, line 1327: does this code handle a negativ leveldif ?

Metalaka: a glancing can only occur against same or higher level, can't be negative.

jackpoz: Well but before your change, in case of same level reductionMax was 0 and reductionMin was -10.

Btw is that std::min correct ? With a leveldif of 3, before reductionMin was 20, now it's 1

Metalaka: Before my changes, glancing didn't happened against same level. So it was fine.

You're totally right, that's a mistake.
The goal was to obtain a reduction range between 1% and 10% when leveldif is 1. So it's the max of 1 and 0 not the min…

```suggestion
            int32 const reductionMin = std::max(1, reductionMax - 10);
```

xParad0x (approved):
Looks good. Tested

jackpoz:
Thanks for the PR :)

src/server/game/Entities/Unit/Unit.cpp
jackpoz: @Metalaka remember when I asked "does this code handle a negativ leveldif ?" ?

https://gist.github.com/Aokromes/7e65ac8c8d3ed269eb9aef0214941603

>reductionMax = -70
reductionMin = 413603904

Random.cpp:37 in irand ASSERTION FAILED

It's better to handle in the code all the cases so crashes don't happen

jackpoz:
This PR causes a crash https://gist.github.com/Aokromes/7e65ac8c8d3ed269eb9aef0214941603

jackpoz:
This PR has been reverted, please fix the crash, test it ingame, handle the PR review comments.

Please open PRs in draft status if they are not tested.

src/server/game/Entities/Unit/Unit.cpp
Metalaka: This was not intended. That negative reduction allow to silently do more damage with glancings that critics.
A level 80 vs a vanilla-boss (lvl63) won an insane 2.75 modifier on glancings.
It's fixed in https://github.com/TrinityCore/TrinityCore/pull/26120

Feel free to add some checks on `leveldif` if you want, but it contiue to seem overkill to me.

Issue #17563, opened 2016-07-10
[3.3.5] Creature Formations [25$]


lively158:
**Description:** If you kill the leader of a formation and then move away from the members of the formation, they are not able to move. They get stuck as if they are trying to follow the leader still.

**Current behaviour:** Standing still while trying to attack.

**Expected behaviour:** Continue attacking and following as normal e.g. as soon as entering combat, forgot the formations until exiting combat.

**Steps to reproduce the problem:**

  A good example is Garr inside of Molten Core, aggro him and kill him while leaving his Firesworn alive and then move as if you were stood in fire, they will remain stuck but keeping aggro

**Branch(es):** 3.3.5

**TC hash/commit:** 22d29ce

**TDB version:** TDB_full_335.61_2016_04_11

**Operating system:** Windows 10 Pro x64

TrinityCoreBot:
This issue is not considered valid because of the following reasons:
- No proper commit hash (revision)

Please read http://bit.ly/tc-issuetracker-and-you and add the missing information to **this issue**. Thanks.

lively158:
Corrected?

DDuarte:
Yes

MrSmite:
> They get stuck as if they are trying to follow the leader still.

They are, they're trying to follow him into death. What are you waiting for? ;)

spectrez:
Any word on this? 

not being able to create proper formations because of this bug is really hampering

Langerz82:
Trying to work on this.

Langerz82:
Ready for testing see:
https://gist.github.com/Langerz82/88137d801c9514d7eef19b1ccf06e93c

Killyana:
Did you take in consideration the case where the leader respawns, what happen in this case?

Kittnz:
it should rejoin the group and take lead again

ccrs:
thats so, so, wrong, u cannot copy motion slots like that, just create new movement on the new leader using old's information.

Langerz82:
@ccrs how would I go about creating a new MotionMaster using the leaders old info?

ccrs:
create new movement, new movement generator, not a new motion master, thats like killing a bee with a rocket.

Langerz82:
give me some time I'll try and work it out ^_^

Langerz82:
@ccrs can you give me a quick dirty example of how I would go about doing such a thing? I think this issue may be over my head but I'm eager to work it out.

[...]

Issue #25494, opened 2020-09-24
Core/Formations: Waypoints movement is blocked if creature is not the leader of formation


offl:
**Description:**

If creature is not the leader of formation and if it has waypoint, waypoint movement will be blocked. In other words creature will not move

**Expected behaviour:**

Waypoint movement should be not blocked even if it's not the leader of formation

I have the case where creatures are pooled, there are 2 pooled creatures per pack of mobs and both of them should walk on path. I can't make them both leaders and I don't want to give leader to pooled creature at all

**Steps to reproduce the problem:**

```sql
DELETE FROM `creature_formations` WHERE `leaderGUID` IN (83289);
INSERT INTO `creature_formations` (`leaderGUID`, `memberGUID`, `dist`, `angle`, `groupAI`, `point_1`, `point_2`) VALUES
(83289,83289,0,0,3,0,0),
(83289,78323,0,0,3,0,0),
(83289,83266,0,0,3,0,0),
(83289,84845,0,0,3,0,0),
(83289,84844,0,0,3,0,0),
(83289,49348,0,0,3,0,0),
(83289,83283,0,0,3,0,0);
```
1. apply sql from above
2. .go c 78323
3. notice creature does not move

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 9fda67d01679

offl:
To be applied after fix, part of https://github.com/TrinityCore/TrinityCore/commit/63b6e9054f2cec29bf00f2dbb600aa4b1e1b034f

<details>

```sql
DELETE FROM `creature_formations` WHERE `leaderGUID` IN (83289,49354,12163,83287,83284,12162,49153,49159,83293);
INSERT INTO `creature_formations` (`leaderGUID`, `memberGUID`, `dist`, `angle`, `groupAI`, `point_1`, `point_2`) VALUES
(83289,83289,0,0,3,0,0),
(83289,78323,0,0,3,0,0),
(83289,83266,0,0,3,0,0),
(83289,84845,0,0,3,0,0),
(83289,84844,0,0,3,0,0),
(83289,49348,0,0,3,0,0),
(83289,83283,0,0,3,0,0),
(49354,49354,0,0,3,0,0),
(49354,83267,0,0,3,0,0),
(49354,29119,0,0,3,0,0),
(49354,83290,0,0,3,0,0),
(49354,49353,0,0,3,0,0),
(49354,86084,0,0,3,0,0),
(49354,86083,0,0,3,0,0),
[...]

Issue #24764, opened 2020-06-07
Core/Creature: Possibility to use formations for creature entry is missing


offl:
We need another table `creature_formations` but for creature entries. Kinda  `creature_formations_template`.
This is needed to make groups of summoned creatures, to make them aggro together.
Also can be used to more easily make groups of static spawns. For example right now if a boss should aggro with 100 trash NPCs, you have to make formation for each creature guid.

Basically we need analogue of https://github.com/cmangos/issues/wiki/creature_linking_template with columns `map` and `search_range`. At this moment we don't need `dist`, `angle`, `point_1`, `point_2`. Just `leaderEntry`, `memberEntry`, `groupAI`,  `map` and `search_range`.

**TC rev. hash/commit:** d81a9e5bc3b3

dr-j:
well think ive said before some creatures should have multiple formations and change formation via script examples are script for dimensius (still awaiting commit) where during saheeds escort and engaging dimensius formation of the npcs which help changes at least 3 times

Other examples are probably bodyguard for hire/gizzleton caravan where order of formation should change depending on direction they are travelling in or depending on which quest is currently active (by direction) as to which takes the lead atm there is no formations there both npcs and both kodos have individual waypoints rather than using formation which results in that all of those can get seperated from each other

offl:
Let's wait first for creature_formations_template and its basic functionaliy without anything related to dist and angle. Because it will be easier for devs.

The things you mentioned probably requires some sort of dynamic formations and right now related only to the creature_formations table. Maybe one day I will open an issue about it. But right now I feel like I forgot everything related to this thing.

Ovahlord:
Any example of where you would actually need this without hacking the shit out of a script?

Jinnaix:
I might be wrong here but i remember that in several Dungeons/Raids if you sneak past the packs to a boss and attack him, the area that belongs to the boss gets pulled at once. Or skipping a boss and attacking the final boss leads to the skipped boss getting aggroed.

Problem with this and Trinitycore is - we can only put a creature into one formation. So trash linking and a pull mechanic for attacking the boss is currently not possible. See attumen in karazhan vor example - Every creature on his way is linked to him. Meanwhile, you cant pack them together with creature formations.

Side note - creature formation AI is still broken:
https://github.com/TrinityCore/TrinityCore/issues/17563#issuecomment-615197753

Aokromes:
[...]

PR #26101, opened 2021-02-16
Scripts/BlackrockDepths: Small correction for final boss room

Jinnaix:
**Changes proposed:**

-  Removing call for help on Boss
- Adding creature formation to all units instead 

Call for help only targets units visible for the boss - this leaves out several units in the final room.

**Target branch(es):**

- [X] 3.3.5

**Issues addressed:**

None

**Tests performed:**

Builds and tested

Irrelevant in this case but just a reminder: https://github.com/TrinityCore/TrinityCore/issues/17563


src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockDepths/boss_emperor_dagran_thaurissan.cpp
```diff
@@ -64,7 +64,7 @@ class boss_emperor_dagran_thaurissan : public CreatureScript
             void JustEngagedWith(Unit* /*who*/) override
             {
                 Talk(SAY_AGGRO);
-                me->CallForHelp(VISIBLE_RANGE);
+                // me->CallForHelp(VISIBLE_RANGE); -- Only calls in line of sight, leaves out the Units behind the pillars
```
Aokromes, line 67: delete thee row, no comment it.

Jildor:
maybe better modify CallForHelp method and add bool ignoreLoS after float radius?

offl:
This way we will be not able to create formations for packs(primary key will not allow it)
Instead we need to create another param for CallForHelp to allow it to select creatures not in LoS

Or update formations
There are cases https://github.com/TrinityCore/TrinityCore/issues/24786#issuecomment-643774655 which cannot be done with CallForHelp and only with formations

Jinnaix:
> This way we will be not able to create formations for packs(primary key will not allow it)
> Instead we need to create another param for CallForHelp to allow it to select creatures not in LoS
> 
> And update formations
> There are cases [#24786 (comment)](https://github.com/TrinityCore/TrinityCore/issues/24786#issuecomment-643774655) which cannot be done with CallForHelp and only with formations
> 
> In this particular case better to use CallForHelp

I did this change some months ago on my private fork and im sure i checked if the unit packs are standing close enough to each other to correctly be pulled as a pack on aggro.

offl:
Point still stands, we will be not able to use formations or CallForHelp in other cases
And I'm able to pull less than I should

Jinnaix:
Im fine with this getting closed. I personally see the need for sub-formations in this and other cases.
Especially in tbc dungeons there was a lot of situations where previous bosses get pulled if you sneak past them and aggro the final boss, while the previous bosses also had a room pull mechanic.

offl:
We barely can use both options currently
If someone will want to make formations, he will need to remove your formations and restore CallForHelp
In case of Pandemonius I can't use neither CallForHelp or formations

jackpoz:
Should a new issue be created about the "ignoreLoS" parameter ?

Jildor:
That is what I think would be more useful, but I don't know if it is what offl needs in the case of pandemonius or if need something else, because if it is only for this case (thaurissan) I don't see it necessary

offl:
I remember https://youtu.be/UcOi3-Hj-Y4?t=385 but CallForHelp will also pull Whitemane out of door and that's not good

offl:
> Instead we need to create another param for CallForHelp to allow it to select creatures not in LoS

Nah
It's a temp solution, we don't really need it
So yeah, it will be used in few cases

jackpoz:
> It's a temp solution, we don't really need it

@offl what is the long term solution ? I'm just trying to figure out if you need someone to fix some C++ issues or if you have everything in place and the fix is going to be a SQL-only update

Jildor:
in case you need to add the option to ignore LoS in CallForHelp, let me know, I can do a PR

Jinnaix:
Just a reminder that as of right now, the boss room behavior in BRD is not correct.
But i think generating aggro with CallForHelp would trigger a creature formation, correct? Maybe in that case we group the trash packs in that room.

offl:
> what is the long term solution ?

I wish I knew

1. We indeed can use CallForHelp in some cases, it's fine
2. CallForHelp ignoreLoS probably will be used few times, is it worth it?
3. we currently can't use formations if not by pathing creatures because they are broken https://github.com/TrinityCore/TrinityCore/issues/17563#issuecomment-753526821 https://github.com/TrinityCore/TrinityCore/issues/25494
4. even if 3 will be fixed right now I don't know how to use formations to link packs of trash and at the same time link them to boss in the same formation

Jildor:
Apart from some boss script, would there be NPCs with smartAIs that use the CallForHelp action with ignoreLoS param?
If positive I can open a PR with the change, otherwise it might not be worth it?

offl:
> would there be NPCs with smartAIs that use the CallForHelp action with ignoreLoS param

I doubt

If we want to solve this particular case right now without formations, the best option is CallForHelp with ignoreLoS, yes
But it's indeed will be used in few cases

> 4. even if 3 will be fixed right now I don't know how to use formations to link packs of trash and at the same time link them to boss in the same formation

https://github.com/TrinityCore/TrinityCore/issues/24764 but again can't be used always

We are far from correct implementation for things related to linked aggro in general and especially for linked aggro in case of bosses

PR #26102, opened 2021-02-16
Core/Conditions: CONDITION_NEAR_CREATURE will exclude other's personal spawns

matanshukry:
**Changes proposed:**

-  CONDITION_NEAR_CREATURE will exclude personal spawns that aren't the target's owner

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #

**Tests performed:**

- [x] It builds
- [x] Tested in-game

### DIFF
--- a/src/server/game/Grids/Notifiers/GridNotifiers.h
+++ b/src/server/game/Grids/Notifiers/GridNotifiers.h
@@ -28,6 +28,7 @@
 #include "Player.h"
 #include "Spell.h"
 #include "SpellInfo.h"
+#include "TemporarySummon.h"
 #include "UnitAI.h"
 #include "UpdateData.h"
 
@@ -1312,7 +1313,7 @@ namespace Trinity
 
             bool operator()(Creature* u)
             {
-                if (u->getDeathState() != DEAD && u->GetEntry() == i_entry && u->IsAlive() == i_alive && i_obj.IsWithinDistInMap(u, i_range))
+                if (u->getDeathState() != DEAD && u->GetEntry() == i_entry && u->IsAlive() == i_alive && i_obj.IsWithinDistInMap(u, i_range) && !TempSummon::IsPersonalSummonOfAnotherPlayer(u, i_obj.GetGUID()))
                 {
                     i_range = i_obj.GetDistance(u);         // use found unit range as new range limit for next check
                     return true;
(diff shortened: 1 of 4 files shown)


src/server/game/Grids/Notifiers/GridNotifiers.h
```diff
@@ -1321,10 +1322,18 @@ namespace Trinity
...
 
         private:
-            WorldObject const& i_obj;
-            uint32 i_entry;
-            bool   i_alive;
-            float  i_range;
+            bool IsPersonalSpawnOfAnoter(Creature* creature)
+            {
+                if (TempSummon const* tempSummon = creature->ToTempSummon())
+                    if (tempSummon->IsVisibleBySummonerOnly() && i_obj.GetGUID() != tempSummon->GetSummonerGUID())
```
Ovahlord, line 1328: GetSummonerGUID is not a good idea because there are tempoary summons that do not get a summonerguid or creatorguid assigned (wild summons for example) so you actually gotta check for  GetSummoner()->GetGUID

src/server/game/Grids/Notifiers/GridNotifiers.h
```diff
@@ -1321,10 +1322,18 @@ namespace Trinity
...
+                        return true;
+                return false;
+            }
+
+            WorldObject const&      i_obj;
+            uint32                  i_entry;
+            bool                    i_alive;
+            float                   i_range;
```
illfated, line 1336: Looks useful & tidy. I'm sort of torn regarding these elements and whether the tidy format follows TC codestyle (I can't remember seeing this arrangement before).

src/server/game/Grids/Notifiers/GridNotifiers.h
Shauren: This calls `TempSummon::GetSummonerGUID()`, it is fine

src/server/game/Grids/Notifiers/GridNotifiers.h
matanshukry: It was used before the chageI made but just on the 3 variables rather than the 4.
It's also used extensively in the code, although mostly in enums. I don't know if it's the current standard since I believe it was changed over the years, but here are few examples where it's used:

- https://github.com/TrinityCore/TrinityCore/blob/master/src/server/database/Database/MySQLConnection.h#L101
- https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Battlegrounds/Zones/BattlegroundAB.h#L359
- https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Creature/Creature.h#L441

Shauren: The current standard is "as few changes as possible"

PR #26109, opened 2021-02-17
Core/Spells: Modable max stack amount

matanshukry:
**Changes proposed:**

-  Modable max stack amount

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] tested in-game

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -22414,7 +22414,7 @@ void Player::ApplySpellMod(uint32 spellId, SpellModOp op, T& basevalue, Spell* s
         return;
 
     float totalmul = 1.0f;
-    int32 totalflat = 0;
+    T totalflat = 0;
 
     // Drop charges for triggering spells instead of triggered ones
     if (m_spellModTakingSpell)
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -22414,7 +22414,7 @@ void Player::ApplySpellMod(uint32 spellId, SpellModOp op, T& basevalue, Spell* s
         return;
 
     float totalmul = 1.0f;
-    T totalflat = 0;
+    int32 totalflat = 0;
 
     // Drop charges for triggering spells instead of triggered ones
     if (m_spellModTakingSpell)
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -1834,7 +1834,7 @@ void Unit::CalcAbsorbResist(DamageInfo& damageInfo)
         absorbAurEff->GetBase()->CallScriptEffectAfterAbsorbHandlers(absorbAurEff, aurApp, damageInfo, tempAbsorb);
 
         // Check if our aura is using amount to count damage
-        if (absorbAurEff->GetAmount() >= 0)
+        if (absorbAurEff->GetAmount() > 0)
         {
             // Reduce shield amount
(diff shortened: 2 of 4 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -22414,7 +22414,7 @@ void Player::ApplySpellMod(uint32 spellId, SpellModOp op, T& basevalue, Spell* s
         return;
 
     float totalmul = 1.0f;
-    int32 totalflat = 0;
+    T totalflat = 0;
```
Shauren, line 22417: Why change this?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -1834,7 +1834,7 @@ void Unit::CalcAbsorbResist(DamageInfo& damageInfo)
         absorbAurEff->GetBase()->CallScriptEffectAfterAbsorbHandlers(absorbAurEff, aurApp, damageInfo, tempAbsorb);
 
         // Check if our aura is using amount to count damage
-        if (absorbAurEff->GetAmount() >= 0)
+        if (absorbAurEff->GetAmount() > 0)
```
Shauren, line 1837: Why change this?

src/server/game/Spells/Auras/SpellAuras.cpp
```diff
@@ -924,12 +924,21 @@ void Aura::SetStackAmount(uint8 stackAmount)
     SetNeedClientUpdateForTargets();
 }
 
+uint32 Aura::GetMaxStackAmount() const
```
Shauren, line 927: Could you rename this to `CalcMaxStackAmount` for consistency with CalcMaxCharges and to make it clear that it does not simply read a cached field?

src/server/game/Spells/Auras/SpellAuras.cpp
```diff
@@ -191,7 +191,7 @@ void AuraApplication::BuildUpdatePacket(WorldPackets::Spells::AuraInfo& auraInfo
 
     // send stack amount for aura which could be stacked (never 0 - causes incorrect display) or charges
     // stack amount has priority over charges (checked on retail with spell 50262)
-    auraData.Applications = aura->GetSpellInfo()->StackAmount ? aura->GetStackAmount() : aura->GetCharges();
+    auraData.Applications = aura->GetMaxStackAmount() ? aura->GetStackAmount() : aura->GetCharges();
```
Shauren, line 194: I would rather not do unneccessary caster lookup to calculate max stacks since we are not interested in that value here, just to know if aura uses stacks or charges

src/server/game/Entities/Player/Player.cpp
matanshukry: I thought it was a mistake.
Isn't it? shouldn't it be T?

src/server/game/Entities/Unit/Unit.cpp
matanshukry: When the amount is 0 and the spell is passive it will still remove, although it shouldn't.
Tested on brambles, 203953.

src/server/game/Spells/Auras/SpellAuras.cpp
matanshukry: What if the spell has 0 on both stacks and charges, but can be "moded" (is that a verb?) to have more stacks? or is that not a possible scenario?

matanshukry:
Looks like travis doesn't like me.
It looks as if it's complaining about code from a different branch, that doesn't exist at all on this PR.

Either way everything else has passed though

MaxtorCoder:
It's probably shauren's code because he did a commit on Block percentages.

PR #26129, opened 2021-02-23
* Core/Conversation Allow to define 'nearest creature id' for convers…

Mistixx:
…ation actors

**Changes proposed:**

-  
-  
-  

- [ ] 3.3.5
- [ ] master

**Issues addressed:**

**Tests performed:**

- [ ] 
- [ ]


MaxtorCoder:
What real usage has this? Example?

illfated:
Looks like the PR owner has not taken the time to fill in the PR comment template. Could it be that this PR targets incorrect repository?

src/server/game/Globals/ConversationDataStore.cpp
```diff
@@ -106,7 +107,8 @@ void ConversationDataStore::LoadConversationTemplates()
             uint32 conversationId         = fields[0].GetUInt32();
             uint32 actorId                = fields[1].GetUInt32();
             ObjectGuid::LowType actorGuid = fields[2].GetUInt64();
-            uint16 idx                    = fields[3].GetUInt16();
+            uint32 actorNearId            = fields[3].GetUInt32();
+            uint16 idx = fields[4].GetUInt16();
```
matanshukry, line 111: indent

illfated: Yes, preferably keep the codestyle from the lines above.

matanshukry:
@Mistixx how is the SC script loader relevant?
Also, more importantly - can you fill in the relevent details and describe what is the use case for this?

I also have an issue in this area, and it would be better to think of a solution that applies to both use cases.

MaxtorCoder:
> 
> 
> @Mistixx how is the SL script loader relevant?
> Also, more importantly - can you fill in the relevent details and describe what is the use case for this?
> 
> I also have an issue in this area, and it would be better to think of a solution that applies to both use cases.

It totally isn't relevant at all.

Mistixx:
I'm sorry is that I accidentally made a commit to another repository and had the path set for trinity

Mistixx:
It has been an error so you can delete this pull request

MaxtorCoder:
> 
> 
> It has been an error so you can delete this pull request

You can close it yourself as well.

PR #26133, opened 2021-02-23
Core/Spells: Modify cooldowns when cooldown aura starts

matanshukry:
**Changes proposed:**

-  Modify cooldowns when cooldown aura starts

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

**Notes:**

- The implementation is based on the assumption that the % of cooldown reduction is from the spell cooldown, and not from the cooldown left, regardless of the fact it is done while the spell is on cooldown.

**Example for an issue:**

1. Use a druid with guardian spec
2. Use Thrash. It will start the 6s cd
3. Use Berserk. Your cd should be lowered by 75%, although It is not.


Faq:
Any example where this could be usable?

matanshukry:
> Any example where this could be usable?

Added as an example to the OP

Shauren:
<%Shauren>  i just checked that bear berserk on retail
<%Shauren> they just completely reset thrash and growl cooldown (its not a generic behavior of cooldown mod aura)

PR #26139, opened 2021-02-23
Group level up

### LINKED ISSUE #26071: group level
**Description:**

how about create one command group level to raise all players on a group?

The reason for such command is allow a gm create a fresh group of the needed lvl to test things without have to level 5, 10, 25, 40, 80 characters one each time or change configuration to create all characters of the desired level.

**Expected behaviour:**

all group levels

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

bf0975f658d5

**Operating system:** 

N/R

J-Cieplinski:
**Changes proposed:**

-  Added command to level up all players in group

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #26071

**Tests performed:**

- [x] Does it build 
- [x] Tested in-game


J-Cieplinski:
I wanted to add ressurect for offline players but it requires GUID and It seems that I cannot get one  from `GroupReference` where source is offline (as it's going to be nullptr?). If it's doable then please point me in the right direction.

Also for the record. I think that current solution of adding specified amount of levels to current level is unintuitive. IMO it should be setting level to the specified one. But I left it as is to keep it consistent with non-group version

jackpoz:
@J-Cieplinski there are 2 different "level" commands, ".level" and ".char level", one adds the number of levels, the other sets it to the specified number (I don't remember which is which).

Please check them ingame and see which one fits best this new command.

J-Cieplinski:
> @J-Cieplinski there are 2 different "level" commands, ".level" and ".char level", one adds the number of levels, the other sets it to the specified number (I don't remember which is which).
> 
> Please check them ingame and see which one fits best this new command.

Oh I did not notice the `.char level` command. `.level` adds the number of levels.

I would say functionality wise `.char level` would be better but It seems to me like group commands should be both functionality and naming wise similar to their regular counterparts. To that end `.group char level` seems unintuitive.

If it were up to me I would go with `char level` functionality but with `.group level` command. Dunno if there are some conventions/guidelines for things like that. If it's ok then I would do it this way.

Aokromes:
never used char level myself xd i use only level to raise level on the amount of levels i need to test (usualy 79 levels more)

J-Cieplinski:
> never used char level myself xd i use only level to raise level on the amount of levels i need to test (usualy 79 levels more)

In that case I will leave it as is since the ticket this PR closes is yours :)

jackpoz:
I personally find the behavior of ".level" very messy, every time I create a char and want to level to 80 I end up writing ".level 80", reaching level 81 and having to write ".level -1".

So when adding a new ".group level" command, I would find more reasonable to have the ".char level" behavior, so if I write ".group level 80" I know every member will be level 80 without the need of writing a table with math formulas "so Player A is level 53 so to reach to level 80 I need to add 17, no wait, 27, right, Player B is level 54 so if I use 27 he will go 81, that's not good, let me..." and so on.

@J-Cieplinski please hold on a moment until I convince @Aokromes and then we can have ".group level" behave like ".char level" 😃

Aokromes:
rename it to group leveladjust (short for it will be group level) XDDDDD if someone wants group setlevel he can create a new issue :trollface:

jackpoz:
I would say let's have ".group level 80" and then we will think about how to make @Aokromes happy (maybe with a sub command or so).

J-Cieplinski:
Ok so I think i managed to make everyone happy.

There are now 2 group commands for leveling group up.
`group level` which just adds specified number of levels to the current one
`group set level` which sets the level to specified one.

Is that ok?

jackpoz:
".group level 79" makes no sense to me and it's a command behavior that we are keeping in ".level" only due to backward compatibility.
When looking for the command, none will think that ".group set" is where the level is, considering ".char level" already sets the level.

Please change ".group level" so that when someones writes "80", it will set the level of the group to 80.

illfated:
> Please change ".group level" so that when someone writes "80", it will set the level of the group to 80.

Maybe a level increase could be performed by using a command like `.group add level 79` to add levels to the existing one?

Aokromes:
> > Please change ".group level" so that when someone writes "80", it will set the level of the group to 80.
> 
> Maybe a level increase could be performed by using a command like `.group add level 79` to add levels to the existing one?

no, because level adds or removes levels

while the char level set levels and when i opened ticket i din't know about char level and i don't care about it, for me "char level" group equivalent must be "group set level" or "group level set"

J-Cieplinski:
Ok gents, I'm unsubscribing from this PR. Please @ me after you hash it out between yourselves :)

Aokromes:
i close my request, you can set it like jackpoz wants. for me it will be useless.

Aokromes:
i close my request, you can set it like jackpoz wants.

illfated:
@J-Cieplinski : Sorry for causing you discomfort there. I guess I added too much information, and Aokromes corrected me somewhat. I stand down.

jackpoz:
Thanks for the PR :)

Issue #24411, opened 2020-04-08
Dungeon Finder fails


reisa089:
**Description:**

The dungeon finder fails when you have all options assigned after a period of time of several hours or days

**Current behaviour:**

Currently every once in a while the dungeon finder starts to malfunction. When in theory you have all the search engine options assigned (Tank, heal, dps), the search engine indicates that another specialization is missing. I will attach 2 images to you to see the situation.

https://cdn.discordapp.com/attachments/577987817540681730/697372408520376320/WoWScrnShot_040820_110804.jpg

https://cdn.discordapp.com/attachments/577987817540681730/697372410470465606/WoWScrnShot_040820_110807.jpg

**Expected behaviour:**

The dungeon finder should work correctly regardless of the time the online server spends.

**Steps to reproduce the problem:**

I am not sure how the problem originates but we believe that the situation occurs when some user performs alt + f4 in the middle of searching for a game but it is a theory. We have not been able to find the cause for now.

**Branch(es):**

CHANGEME 3.3.5, master or both

**TC rev. hash/commit:** 
last committ implemented:
DB/Quest: "Attunement to Dalaran" Remove non repeatable flag from the quest credit event // d5f57342412608de5233d1e576eedb73e7bcd0f9

**Operating system:** Debian 10

jackpoz:
could you continue the investigation, checking if the alt + f4 always triggers the issue ? It will help us reproducing and fixing it

reisa089:
No problem jackpoz. I hope to confirm exactly where it fails in the coming days. Greetings and thank you

Aokromes:
use lfg queue debug to see if you find any "bad states"

reisa089:
Hello good day. We have been doing ingame tests at a time where we had 180 online. In the tests carried out, it has given us the feeling that the dungeon finder system has an unexpected behavior when more than 5 people with similar levels enter. In tests carried out on each user, the search engine indicated that different functions were missing to score. If it is true that when it happens that there are 3 dps, 1heal and a tank, the search engine works.

The conclusion that we can draw is that the search engine does not process the information well as there are more than 5 people listed, indicating that different functions are missing from each one and therefore giving the feeling of failure of the search engine itself.

We would love to be able to confirm this point 100% but it is quite difficult since many factors come into play. Anyway we have followed the indications of Aokromes and we have taken a log of the lfg of several hours.

We are going to leave the log active for several days, so I will upload the most up-to-date log shortly if you like

[...]

PR #26145, opened 2021-02-24
Core/LFG: Fixed a bug where rolecheck timer was milliseconds

robinsch:
**Changes proposed:**

-  roleCheck.cancelTime is time_t (seconds precision) causing role check to not fail for 12h
See: https://github.com/TrinityCore/TrinityCore/blob/57e63bb6a8699494f427efa8b5b03d65239967b5/src/server/game/DungeonFinding/LFGMgr.cpp#L298

- [x ] 3.3.5

### DIFF
--- a/src/server/game/DungeonFinding/LFGMgr.h
+++ b/src/server/game/DungeonFinding/LFGMgr.h
@@ -44,7 +44,7 @@ enum LfgOptions
 
 enum LFGMgrEnum
 {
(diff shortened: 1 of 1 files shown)


Jildor:
Maybe related issue: #24411

illfated:
Let's hope you're right, would be nice to resolve that issue ticket.

BTW, cosmetic/GitHub tech: Remember to put the X without any space in your tick box, like this:

MarkDown:
```
- [X] 3.3.5
```

Result:
- [X] 3.3.5

PR #26147, opened 2021-02-24
Core/Spells: Fixed various paladin spells and removed various unused as well

matanshukry:
**Changes proposed:**

-  Fixed various paladin spells and removed various unused as well

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

### DIFF
--- a/src/server/scripts/Spells/spell_paladin.cpp
+++ b/src/server/scripts/Spells/spell_paladin.cpp
@@ -22,9 +22,11 @@
  */
 
 #include "ScriptMgr.h"
+#include "DB2Stores.h"
 #include "Group.h"
 #include "Player.h"
 #include "Random.h"
+#include "Spell.h"
 #include "SpellAuraEffects.h"
 #include "SpellHistory.h"
 #include "SpellMgr.h"
@@ -33,7 +35,7 @@
 enum PaladinSpells
 {
     SPELL_PALADIN_AVENGERS_SHIELD                = 31935,
-    SPELL_PALADIN_AURA_MASTERY_IMMUNE            = 64364,
+    SPELL_PALADIN_AVENGING_WRATH                 = 31884,
     SPELL_PALADIN_BEACON_OF_LIGHT                = 53563,
     SPELL_PALADIN_BEACON_OF_LIGHT_HEAL           = 53652,
     SPELL_PALADIN_BLESSING_OF_LOWER_CITY_DRUID   = 37878,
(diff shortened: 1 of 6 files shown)


src/server/scripts/Spells/spell_paladin.cpp
```diff
@@ -746,219 +558,180 @@ class spell_pal_hand_of_sacrifice : public SpellScriptLoader
...
-                    {
-                        if (sSpellMgr->GetSpellInfo((*i)->GetAmount(), GetCastDifficulty()))
-                        {
-                            spellId = (*i)->GetAmount();
-                            break;
-                        }
-                    }
-                }
+// 40470 - Paladin Tier 6 Trinket
+class spell_pal_item_t6_trinket : public AuraScript
```
mdX7, line 689: legacy set effects were removed in wod or legion afaik

src/server/scripts/Spells/spell_paladin.cpp
```diff
@@ -1430,34 +1032,55 @@ class spell_pal_t8_2p_bonus : public SpellScriptLoader
...
+    PrepareAuraScript(spell_pal_zeal);
+
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_PALADIN_JUDGEMENT_PROT_RET_R3, SPELL_PALADIN_JUDGEMENT_GAIN_HOLY_POWER });
+    }
+
+    void HandleEffectProc(AuraEffect* aurEff, ProcEventInfo& /*procInfo*/)
+    {
+        Unit* caster = GetCaster();
```
mdX7, line 1047: caster may be null

src/server/scripts/Spells/spell_paladin.cpp
Shauren: This is not a set bonus effect, its a trinket (spell is literally named "Paladin Tier 6 Trinket")

illfated:
I am guessing that your PR branch is behind the current TC master branch, hence the file conflict.

matanshukry:
> I am guessing that your PR branch is behind the current TC master branch, hence the file conflict.

Sort of; just made specific new PRs split from this one to make this one look less horrible, which is what caused the conflicts. Will resolve in a minute.

src/server/scripts/Spells/spell_paladin.cpp
Ovahlord: GetTarget() is what you wanna use for auras that equal getcaster in 9/10 cases

matanshukry: Yeah I actually fixed that already after @mdX7 comment. Realized I should of used GetTarget + corrected the validate method.

Shauren:
I made some minor adjustments to some of them, one change remains to be done

Righteous Verdict needs to be removed from Templar's Verdict script and converted to a proc aura script (267610 has proc flags in SpellInfo)

matanshukry:
@Shauren you used the Spell version of HasPowerTypeCost / CalcPowerCost on the spells, but is that what we want?

Specifically for selfless healer, when divine storm is free (due to another talent) - shouldn't it still proc?
Tooltip makes me think it should ("power spending abilities"), although currently it won't since the power cost of it will be 0.

Shauren:
If you were worried about that then it's not a problem - all costs are in that container, even those that calculated to 0

matanshukry:
> If you were worried about that then it's not a problem - all costs are in that container, even those that calculated to 0

I'll do the remaining change later on tonight then.
And thanks!

PR #26163, opened 2021-02-25
Core/ThreadManager: delete double include "TemporarySummon.h"

Winfidonarleyan:
**Changes proposed:**
- Delete double include "TemporarySummon.h"

- [x] 3.3.5

### DIFF
--- a/src/server/game/Combat/ThreatManager.cpp
+++ b/src/server/game/Combat/ThreatManager.cpp
@@ -31,7 +31,6 @@
 #include "ObjectAccessor.h"
 #include "WorldPacket.h"
 #include <algorithm>
-#include "TemporarySummon.h"
 
 #include "Hacks/boost_1_74_fibonacci_heap.h"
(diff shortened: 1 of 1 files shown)


jackpoz:
Thanks for the PR :)

illfated:
Wouldn't avoiding situations like this be more obvious if all the include lists were sorted Lexicographically (alphabetically) ascending?

jackpoz:
@illfated the header was added when copying a commit from master and I didn't really check if the header was there already, I just checked if it built.

The order of headers follows a technical logic (linked header of the .cpp, TC headers, 3rd party headers, std headers), it's just not obvious at first sight.

Aokromes:
> @illfated the header was added when copying a commit from master and I didn't really check if the header was there already, I just checked if it built.
> 
> The order of headers follows a technical logic (linked header of the .cpp, TC headers, 3rd party headers, std headers), it's just not obvious at first sight.

maybe tc headers, 3rd party headers and std headers must be separated by one empty row.

illfated:
Thanks for the insight. I forgot that there is more than one sort order for the include lists.

Issue #24961, opened 2020-07-07
Innkeeper cannot display zhCN


xstellae:
**Description:**
The innkeeper cannot display zhCN. I added zhCN records to gossip_menu_option_locale but it didn't work. I found that 4 locales had been written to npc_innkeeper.cpp and never load from database.

**Expected behaviour:**

Display zhCN currently.

**Steps to reproduce the problem:**

Use a zhCN client and find a innkeeper such as creature_template entry 9501, then it just display English.

**Branch(es):**

3.3.5

Paraciusz:
I find that if I set creature_template.ScriptName to empty string, it will display correctly. What's the difference between them？

illfated:
The innkeeper script only supports 4 languages:

- LOCALE_frFR
- LOCALE_deDE
- LOCALE_esES
- LOCALE_enUS


https://github.com/TrinityCore/TrinityCore/blob/7987ec103e3b611daf9528be5859befa2be402cf/src/server/scripts/World/npc_innkeeper.cpp#L54-L168

illfated:
I don't remember how much work is needed to get rid of that old script and use DB linking instead.

offl:
This is why scripts are linked to quest id or gossip id, not to creature. Currently to use all locales and original gossip menus we need to create 150 scripts for creatures instead of 15 gossip scripts linked to gossip id.

This is exactly why they constantly duplicates similar objects and quests, because they holds unique scripts.
Moving everything to one system was not a good idea. However we still can implement new source type for gossips.

offl:
We need to use broadcast texts directly from table for gossip menu options in such scripts. When our scripts will be tied directly to gossip options, we will be able to properly convert all inkeepers to DB

It's needed also in npc_professions.cpp because this thing is not convertable to DB due to lack of info & really big amount of stuff to duplicate(each gossip menu and its conditions needs to be duplicated multiple times), the best we can do in this case now is use texts directly from table and update all those scripts

illfated:
Thanks for pointing out those details. I usually forget about those details, unless I start looking into what it takes to convert those scripts, because it is quite a mess.

matanshukry:
@offl gossip_menu_option has a `OptionBroadcastTextId` so you should be able to use a gossip menu option with broadcast text?

Although I couldn't find any entry with 'trick or treast' in BroadcastText.


Regarding making this script into db: I think with some enhancement to the SAI system we can make it.
Specifically:

[...]

PR #26170, opened 2021-02-26
Core/SAI: Implement quest source for SAI

matanshukry:
**Changes proposed:**

-  Implement quest source for SAI

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

**Use case:**

Pushed quests (such as wqs) where conversations happens when a quest has been given or completed to player

**Example Usage**

Currently if you level a character to 48 you'll get the quest,
so this quest will teleport you to stormwind once the player has that quest.

```sql
SET @QUEST_SL := 60545; -- A chilling summon

DELETE FROM `smart_scripts` WHERE `entryorguid` = @QUEST_SL AND `source_type` = 5;
INSERT INTO `smart_scripts` (`entryorguid`,`source_type`,`id`,`link`,`event_type`,`event_phase_mask`,`event_chance`,`event_flags`,`event_param1`,`event_param2`,`event_param3`,`event_param4`,`event_param5`,`action_type`,`action_param1`,`action_param2`,`action_param3`,`action_param4`,`action_param5`,`action_param6`,`target_type`,`target_param1`,`target_param2`,`target_param3`,`target_x`,`target_y`,`target_z`,`target_o`,`comment`) VALUES
(@QUEST_SL, 5, 0, 1, 47, 0, 100, 0, 0, 0, 0, 0, 0, 62, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, -8833.07, 622.778, 93.9317, 0.6771, "On QUEST_SL accepted - teleport to stormwind");

DELETE FROM `quest_template_addon` WHERE `ID` = @QUEST_SL;
INSERT INTO `quest_template_addon` (`ID`, `ScriptName`) VALUES
(@QUEST_SL, 'SmartQuest');
```


offl:
Great
Did you tried to use action lists with different target types(quest-giver 1 and player-invoker 7)?

@jackpoz maybe you want to port this to 3.3.5?

matanshukry:
> Did you tried to use action lists with different target types(quest-giver 1 and player-invoker 7)?

I didn't, but I believe it would only work if the events are linked.
I think those restrictions also applies to scene and area triggers though?

Also, currently (before this PR) you can still handle events on quests using a specific NPC and quest id, this PR is needed for pushed quests where there is no quest giver. Is there anyway to have multiple targets in that case?

offl:
Yes, you're right, areatriggers can't load action list but can use linked events. I thought initially (https://github.com/TrinityCore/TrinityCore/issues/24961#issuecomment-726807474) SMART_SCRIPT_TYPE_QUEST, SMART_SCRIPT_TYPE_GOSSIP and SMART_SCRIPT_TYPE_EVENT are supposed to act as event holders, like creatures or objects do. So they can load action lists and target type 1 (self) will refer to event holder(quest-giver, gossip-menu owner)
But looks like it requires more changes

jackpoz:
> @jackpoz maybe you want to port this to 3.3.5?

@offl sure but only after it has been merged to master

offl:
>Yes, you're right, areatriggers can't load action list but can use linked events. I thought initially (#24961 (comment)) SMART_SCRIPT_TYPE_QUEST, SMART_SCRIPT_TYPE_GOSSIP and SMART_SCRIPT_TYPE_EVENT are supposed to act as event holders, like creatures or objects do. So they can load action lists and target type 1 (self) will refer to event holder(quest-giver, gossip-menu owner)
But looks like it requires more changes

Technically we can load an action list via SMART_SCRIPT_TYPE_QUEST or SMART_SCRIPT_TYPE_GOSSIP to quest-giver and gossip-owner. In some cases it's exactly what those script types are supposed to do. Benefits: we are avoinding to script quest taken\rewarded events or gossip select events in main script because it's ugly and may cause problems in future.
For example we want to convert all scripts to free ids predefined templates like Blizz did it. Let's say we want to convert all combat actions to predefined templates, but what's left? Exactly, all those gossip scripts which are supposed to be linked to gossip entry, all quest scripts originally linked to quest entries. All waypoints scripts originally linked to point ids(biggest problem ever). Spellhit events are kinda combat actions but mostly all of them can be linked to script ids(btw SAI also can legitimately replace spell_linked_spell which is indeed hack but not because those scripts are scripted in DB but because that system is not how it's supposed to work. There are 700 linked spells in 3.3.5 and I really doubt someone wants to make 700 scripts just to cast another spell after OnRemove or AfterCast. Spell scripts like creature scripts should be used only to make something complicated, like boss scripts or player spells. To cast another spell or cast random spell you don't need to create C++ spell script because it's insane).

However we have also SMART_SCRIPT_TYPE_EVENT and that thing is more complex because they are map scripts and can be updated even if event starter is dead or despawned(lots of events are called when object despawns). At least it's how event_scripts always worked. In some cases there is no creature or gameobject invoker(taxi events)

So SMART_SCRIPT_TYPE_QUEST or SMART_SCRIPT_TYPE_GOSSIP too should be able to call action lists to quest-giver or gossip-owner otherwise those scripts cannot be used the way they are used in retail
Just thoughts, I'm not asking to implement that thing

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -3815,6 +3832,12 @@ void SmartScript::OnInitialize(WorldObject* obj, AreaTriggerEntry const* at, Sce
         sceneTemplate = scene;
         TC_LOG_DEBUG("scripts.ai", "SmartScript::OnInitialize: source is Scene with id %u", scene->SceneId);
     }
+    else if (qst)
+    {
+        mScriptType = SMART_SCRIPT_TYPE_QUEST;
+        quest = qst;
+        TC_LOG_DEBUG("scripts.ai", "SmartScript::OnInitialize: source is Quest with id %u", scene->SceneId);
```
Shauren, line 3839: Will crash if debug logger is enabled (using nullptr scene pointer instead of quest)

Shauren (changes requested):

PR #26180, opened 2021-02-28
Core/Spells: Treat orientation 0 as valid value in EffectTeleportUnits

### LINKED ISSUE #25273: Core/Spells: EffectTeleportUnits uses player's current orientation if in DB teleport orientation is set to 0
**Description:**

3.3.5
```cpp
    if (!targetDest.GetOrientation() && m_targets.GetUnitTarget())
        targetDest.SetOrientation(m_targets.GetUnitTarget()->GetOrientation());
```
master
```cpp
    if (!orientation && m_targets.GetUnitTarget())
        orientation = m_targets.GetUnitTarget()->GetOrientation();
```

**Expected behaviour:**

Player's orientation must be changed to 0 if in DB teleport orientation is 0

1. .cast 17609
2. wait for teleport, change player's orientation and cast it again
3. notice orientation is not set to 0 (sniffed value)

**Branch(es):** both
**TC rev. hash/commit:** bbcf00257ebd

offl:
**Changes proposed:**

Currently EffectTeleportUnits uses player's current orientation if in DB teleport orientation is set to 0. Sometimes Blizzard uses 0 and it's shown in sniffs

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #25273

**Tests performed:**

Tested long time ago, heh. Can test av build later
Edit: tested av build, works

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -1059,9 +1059,6 @@ void Spell::EffectTeleportUnits(SpellEffIndex /*effIndex*/)
     if (targetDest.GetMapId() == MAPID_INVALID)
         targetDest.m_mapId = unitTarget->GetMapId();
 
-    if (!targetDest.GetOrientation() && m_targets.GetUnitTarget())
-        targetDest.SetOrientation(m_targets.GetUnitTarget()->GetOrientation());
-
     if (targetDest.GetMapId() == unitTarget->GetMapId())
         unitTarget->NearTeleportTo(targetDest, unitTarget == m_caster);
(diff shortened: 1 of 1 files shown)


Ovahlord:
I'd give it the same treatment as I did for waypoints and make orientation a optional value

offl:
There may be cases where orientation should not be changed(core will take current player orientation) but it'll be hard to find at least one case
So 'just in case' orientation in table should be nulled like here https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/commit/171d37d99e41b4d7b691e696f14f1a13c628ec4a

PR #26183, opened 2021-02-28
Core/Spells: implement SpellInfo helper to filter for relevant mechanic immunities in spell_start packet

Ovahlord:
**Changes proposed:**

-  we no longer send immunities for instant cast spells in SpellStart packets
-  filter out mechanic immunities of a caster for relevant ones which currently are MECHANIC_SILENCE and  MECHANIC_INTERRUPT as sniffs have only shown these two values so far.

The actual benefit from this is that spells sometimes are being falsely shown as not interruptable because all the other immunities may confuse the client about if the spell may be interrupted or not. We now send correct data so these issues should be solved.

- [x] 3.3.5
- [ ] master

**Tests performed:**
- tested in 4.x

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -4176,8 +4176,8 @@ void Spell::SendSpellStart()
     uint32 mechanicImmunityMask = 0;
     if (Unit* unitCaster = m_caster->ToUnit())
     {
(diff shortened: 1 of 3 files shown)


illfated:
Nice and proper PR title. Thank you. 👍🏻

jackpoz:
@Ovahlord is this ready to be merged ?

jackpoz:
Thanks for the PR :)

PR #26188, opened 2021-03-01
Core/Spells: Handle auras 328 & 396 (SPELL_AURA_TRIGGER_SPELL_ON_POWER_*)

Robingadko:
(cherry picked from 37bb37b8d4a0f932df2a80dbb205067ac8625821 and extended)

**Changes proposed:**

-  Add hadlers for `SPELL_AURA_TRIGGER_SPELL_ON_POWER_PCT `and `SPELL_AURA_TRIGGER_SPELL_ON_POWER_AMOUNT`
-  Add `Unit::CheckPowerProc` method to decide when auras should proc

- [ ] 3.3.5
- [x] master

**Issues addressed:**

None

**Tests performed:**

Built and tested in-game with https://wowhead.com/spell=190488/fulmination-controller

None

### DIFF
--- a/src/common/Utilities/Util.h
+++ b/src/common/Utilities/Util.h
@@ -72,6 +72,13 @@ inline T CalculatePct(T base, U pct)
     return T(base * static_cast<float>(pct) / 100.0f);
 }
 
+template <class T>
+inline float GetPctOf(T value, T max)
+{
+    ASSERT(max && value <= max);
+    return float(static_cast<float>(value) / static_cast<float>(max) * 100.0f);
+}
+
 template <class T, class U>
 inline T AddPct(T &base, U pct)
 {
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9955,6 +9955,7 @@ void Unit::SetPower(Powers power, int32 val)
     if (maxPower < val)
         val = maxPower;
 
+    uint32 oldPower = GetPower(power);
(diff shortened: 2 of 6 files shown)


Ovahlord:
I think the old naming  SPELL_AURA_PROC_ON_POWER_AMOUNT is describing the behaivior much more accurately as it's a real proc that's only firing off when meeting the criteria of the aura effect. I'm also a bit surprised that master doesn't have its implementation anymore as it's present in 4.x already and iirc it was a official implementation

Traesh:
First Shauren point is handled in the original code (no removal)

Second point is handled in both original and PR code, as we check for old amount value here : https://github.com/TrinityCore/TrinityCore/pull/26188/files#diff-b4a23722c5b449d326b1c0e5a75f3b51de931e42e688793fd9f1d233536a02c6R10037

For example on 190427 if you go from 0 to 1 you'll trigger the proc because oldValue < 1 && newValue >= 1
If you go from 1 to 2 newValue is indeed >= 1 but oldValue is not < 1 then first effect is not triggered.

Shauren:
Then what are those lines doing here? https://github.com/TrinityCore/TrinityCore/pull/26188/files#diff-b4a23722c5b449d326b1c0e5a75f3b51de931e42e688793fd9f1d233536a02c6R10040-R10041

Shauren:
Closing in favor of me manually redoing the cherry-pick
If you are going to do it, at least don't break the original (and correct) logic

56e9560661adca81fc1e9a297cb6823cf7a6cb22

PR #26190, opened 2021-03-01
Scripts/The Eye: Update instance script

offl:
**Changes proposed:**

-  update instance script and header to modern standarts
-  reorder data types and rename them to make less ugly, we don't need full name of spells or creatures
-  had to remove DATA_ASTROMANCER, I have no idea how that thing exists. It's not linked to any creature. I changed it to DATA_SOLARIAN since npc_solarium_priest indeed should cast spell on themselves or on her

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

none

**Tests performed:**

Builds, tested in-game every encounter with .instance getboss

none


ccrs:
iirc existing savedata can be corrupted by this change

offl:
Server owners can update core after reset

Ovahlord:
@ccrs  since when do we care about private servers? Also no, because he merely changed the creature guid accessors which are not being saved to db.

offl:
Speaking of encounter states they should never be changed without checks and should always be numbered. They're not events types which you can rotate as you want and nothing will be broken. It also pisses off DB devs when encounter states are not numbered because if there are 10 or 15 encounter states, you need to count them to link conditions and you need to count them every time you check conditions
Only additional data may be not numbered
```cpp
    // Encounter States
    DATA_ALAR                           = 0,
    DATA_SOLARIAN                       = 1,
    DATA_VOID_REAVER                    = 2,
    DATA_KAELTHAS                       = 3,
```
It's also extremely good to leave a comment near every encounter state if it's used in SAI, gossip text or option conditions because it's not easy to check if encounter state is used in conditions
```cpp
    DATA_TWIN_EMPERORS      = 4,      // Used in conditions for gossip menu options & texts (menu 6644)
    DATA_CTHUN              = 5,      // Used in conditions for SAI (creatures 15378 - 15380)
    DATA_BUG_TRIO           = 6,
    DATA_VISCIDUS           = 7,
    DATA_OURO               = 8,      // Used in conditions for gossip menu options & texts (menu 6644)
```

ccrs:
chill, this change does affect people, some consideration should at least be given to that. I didnt go further than that.
I'm not against this change.

at least give a warning or something.

illfated:
**Scripts/Tempest Keep:**

offl:
Which scripts? Are they DB or C++?

illfated:
Sorry, I should have specified. I meant that the old-school PR title format would be written like that, "Scripts" from the directory category `/src/server/scripts/` and the subsection would either be the name of the instance, instance area, or spell category.

offl:
Well, I have a lot of questions but ok

illfated:
Thank you. "Core" is mostly the other directories outside of `/src/server/scripts/` (in particular, `/src/server/game/` - but also many of the others).

PR #26197, opened 2021-03-03
Remove non-existent logger types: MoveSplineInitArgs and DoZoneInCombat

Smerdokryl:

- [ ] 3.3.5
- [x] master


Aokromes:
nope, they are subcategories to allow you to don't log those.

Smerdokryl:
> nope, they are subcategories to allow you to don't log those.

But they aren't in the worldserver config, and also they show up even if `Logger.misc` is disabled

Aokromes:
> > nope, they are subcategories to allow you to don't log those.
> 
> @Aokromes But they aren't in the worldserver config, and also they show up even if `Logger.misc` is disabled

you missed the word ALLOW. loggers are designed to allow you to configure to your needs.

PR #26201, opened 2021-03-04
Core/Summon: Force personal summon when the summoner is also a pesronal summon

matanshukry:
**Changes proposed:**

-  Force personal summon when the summoner is also a pesronal summon

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1826,7 +1826,20 @@ TempSummon* WorldObject::SummonCreature(uint32 entry, Position const& pos, TempS
 {
     if (Map* map = FindMap())
     {
-        if (TempSummon* summon = map->SummonCreature(entry, pos, nullptr, despawnTime, ToUnit(), 0, vehId, visibleBySummonerOnly))
+        Unit* summoner = ToUnit();
+
+        // If we are a TempSummon use our own summoner as the summoner
+        if (visibleBySummonerOnly)
+        {
+            Unit* selfUnit = ToUnit();
+            TempSummon* tempSummon = selfUnit ? selfUnit->ToTempSummon() : nullptr;
+
+            if (tempSummon->IsVisibleBySummonerOnly())
+                if (Unit* ownSummoner = tempSummon->GetSummoner())
+                    summoner = ownSummoner;
+        }
+
+        if (TempSummon* summon = map->SummonCreature(entry, pos, nullptr, despawnTime, summoner, 0, vehId, visibleBySummonerOnly))
         {
             summon->SetTempSummonType(despawnType);
             return summon;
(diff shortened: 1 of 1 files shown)


Faq:
Any example where this is needed?

matanshukry:
> Any example where this is needed?

in the SL intro quest Darion, which is a personal spawn, summons a gate.
Without this the gate isn't a personal spawn and is visible by everyone.

Even without the use case though - there isn't any situation where the opposite is true;
That is, the following doesn't make sense:
- Player A summons a personal spawn creature B
- Creature B then summons creature C
- Creature C should *not* be a personal spawn

Faq:
Personal spawns minion :)

matanshukry:
> Personal spawns minion :)

I don't know what you mean by that

Ovahlord:
@Faq  personal summons are all kinds of summons that are exclusively visible to the summoner so of course it's logical to make summons that are summoned by the personal summon npc to have them visible to the player owner as well. That kind of logic is also being used by normal tempoary summons by passing down creatorguids to their childsummons

matanshukry:
> The current implementation of 'minions' and 'guardians' is one gíant clusterfuck that mixes up so many elements that should not be mixed up.

I'm not familiar with that; Could you open an issue with info about it? such as what was the reason behind the current implementation separation (if you know), what's the difference etween minions and guardians, and how you think it should be implemented?

Ovahlord:
Minions were originally suposed to be the counterpart to "allied" summons which is determined by SummonProperties field 'Control'.

The hierarchy is the following:

1. TempSummon (any summon, including wild ones)
2. Minion (child of TempSummon for allied summons. These ones have the CreatorGUID assigned in all cases)
3. Guardian (child of Minion, controlable units which have a CreatorGUID as well as the OwnerGUID indicating that the player has full control over these units. There are only controlable guardians.

The issue why it all became clusterfucked is that some people never understood that system and thought that summons such as Warlock Infernals or Guardian of Ancient Kings or DK Gargoyles are guardians but in fact, they are not. They are 'Minions' as they are allied, but not controlled. Been digging through alot of sniffs to validate that.

matanshukry:
@Ovahlord Awesome! Both for the explanation and the cleanup you've already done!
I'll also see if I can make a PR to merge this to tc.
Unless you want to do it yourself, of course.

Ovahlord:
These changed are highly experimental and will break stuff. They are not ready to be released into the wild yet as the entire guardian statsystem basically breaks since many summons that used to benefit from these stats no longer do so because they are no guardians anymore.

ZaDarkSide:
> I've already done the most basic cleanup for that on my 4.x fork. https://git.io/Jt8bp

`src/server/game/Entities/Creature/TempoarySummons` <= you have typo in the folder name in the commit

matanshukry:
Closing in favor of #26245

PR #26202, opened 2021-03-04
Scripts/Ahn'kahet: Herald Volazj texts

offl:
**Changes proposed:**

Changed the way his texts are handled using original spells and added all texts
His whisper texts are translations for his yell texts and are handled in separate system(by spells) and actually in retail are delayed from yell texts(I skipped that part, in TC they are instant). Because of that they cannot be grouped like it's done in other cases

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Tests performed:**

Builds, tested in-game, works

Didn't noticed any so far, I think I screwed up somewhere with triggered cast flag


illfated:
**Scripts/Azjol-Nerub:** would be a more precise title, but I guess "Do what thou wilt" shall be the whole of the Law. (Thelema)

illfated:
Thanks.

jackpoz:
I did a small change to initialize a local variable (just to make sure no uninitialized value ends up being used in the future).

Is there anything missing before this PR can be merged ? The cpp part looks fine.

jackpoz (approved):

PR #26205, opened 2021-03-05
Core/Spells: Implement aura allow swapping talents

matanshukry:
**Changes proposed:**

- Implement aura allow swapping talents

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

### DIFF
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -2547,7 +2547,7 @@ void AuraEffect::HandleAuraAllowTalentSwapping(AuraApplication const* aurApp, ui
 
     if (apply)
         target->AddUnitFlag2(UNIT_FLAG2_ALLOW_CHANGING_TALENTS);
-    else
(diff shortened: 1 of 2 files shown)


src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -2536,6 +2536,21 @@ void AuraEffect::HandleAuraModSkill(AuraApplication const* aurApp, uint8 mode, b
...
+    if (!(mode & (AURA_EFFECT_HANDLE_REAL)))
+        return;
+
+    Player* target = aurApp->GetTarget()->ToPlayer();
+    if (!target)
+        return;
+
+    if (apply)
+        target->AddUnitFlag2(UNIT_FLAG2_ALLOW_CHANGING_TALENTS);
+    else
```
Shauren, line 2550: As with all aura types that apply a flag, it needs this kind of check to properly handle stacking
```
    if (!apply && target->HasAuraType(GetAuraType()))
        return;
```

Shauren (changes requested):

PR #26235, opened 2021-03-13
Core/Cheats: Power - set max power on all powers initially + on regenerate

matanshukry:
**Changes proposed:**

-  Set max power on all powers when applying cheat_power
-  On regenerate, if cheat_power is on set value to max power

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

- [x] It builds
- [x] Tested in-game

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -1841,6 +1841,9 @@ void Player::Regenerate(Powers power)
         }
     }
 
+    if (IsPlayer() && ToPlayer()->GetCommandStatus(CHEAT_POWER))
+        curValue = maxPower;
+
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -1841,6 +1841,9 @@ void Player::Regenerate(Powers power)
         }
     }
 
+    if (IsPlayer() && ToPlayer()->GetCommandStatus(CHEAT_POWER))
```
Shauren, line 1844: I would say that `this` in a function belonging to `Player` class is always a `Player`

Shauren (changes requested):

PR #26240, opened 2021-03-13
Core/OutdoorPvP: pick Silithyst Geyser with PvP ON

### LINKED ISSUE #15906: [Quest \ Silithus] Scouring the Desert
TrinityCore rev. 5df0bcdd4e2d (clean installation)
http://www.wowhead.com/quest=9419/scouring-the-desert (Alliance side - prev quest 9415)
http://www.wowhead.com/quest=9422/scouring-the-desert (Horde side - prev quest 9416)
Both quests can't be completed:
Silithyst can't be obtained from Silithys Geyser
* Note:
Tested with a normal account / player <-- quest doesn't work
Tested with a GM account / player <-- quest do work
Tested with a GM account changed to normal account <-- quest do work

ghost:
**Description:**

As suggested by Rushor a year ago, this change makes it possible to pick up a Silithyst Geyser object while the PvP flag is turned ON.

Before this change, it is only possible to pick up a Silithyst Geyser object when the player character's PvP status is OFF.

Thanks to @Rushor for suggesting this change with the affirmation that it solves the PvP status issues focused on in issue ticket #15906.

**Changes proposed:**

-  change `return true;` to `return false;` in `src/server/game/OutdoorPvP/OutdoorPvP.cpp`, line number 498.

- [X] 3.3.5
- [ ] master

**Issues addressed:**

Closes #15906

**Tests performed:**

Builds without any issues, only 1 Boolean value changed from `true` to `false`.

Tested in-game and found to be working as intended and expected.

- [X] Does not turn off player PvP status when the 5 minutes disable time runs out. (Could this be another enGB game client issue?)
- [ ] Not sure if this will affect other PvP activity negatively. (Maybe this needs testing by active PvP players.)

### DIFF
--- a/src/server/game/OutdoorPvP/OutdoorPvP.cpp
+++ b/src/server/game/OutdoorPvP/OutdoorPvP.cpp
@@ -495,7 +495,7 @@ bool OPvPCapturePoint::HandleCustomSpell(Player* player, uint32 /*spellId*/, Gam
 {
     if (!player->IsOutdoorPvPActive())
         return false;
-    return true;
+    return false;
 }
 
 bool OutdoorPvP::HandleOpenGo(Player* player, GameObject* go)


Shauren:
This is a wrong fix. Real issue is that all OutdoorPvP hooks are called for all types - so this breaks because `HandleCustomSpell` is called first on `OutdoorPvPHP` (hellfire zone), it returns true so `HandleCustomSpell` of `OutdoorPvPSI` is not called. 

The fix should be to only call `HandleCustomSpell` on `OutdoorPvP` object for current zone, not to break OutdoorPvP even further by changing return value

PR #26271, opened 2021-03-19
Core/Languages: Cross-Languages

### LINKED ISSUE #25980: 9.0.2 - The languages ​​of different races are understood by all players regardless of race or faction
**Description:**

Players from different factions or races, when using their own language, must be understood by the opponent if it belongs to another race or faction. 

In my server configurations, the interaction between different factions is completely disabled, but the players still understand each other.

(If you know how to solve this problem, write below and I will close this issue.)

**Expected behaviour:**

Example: a human race player, does not understand an orc race player, due to different languages

1. Create characters of different factions.
2. Teleport two players of the opposite faction to the same location.
3. Let the players try to communicate using different languages ​​typical of their race

**Branch(es):**

master

**TC rev. hash/commit:** 

d4ad987

**Operating system:** 

Windows 10 x64

matanshukry:
**Changes proposed:**

- Attempt at translating the text from different factions

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #25980

**Tests performed:**

- [x] It builds
- [x] Tested in-game

**Notes:**

I based the translation on the following source: https://wowdev.wiki/DB/LanguageWords .


src/server/game/Chat/LanguageMgr.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+uint32 LanguageMgr::SStrHash(char const* string, bool caseInsensitive, uint32 seed) const
+{
+    assert(string);
+
+    uint32 shift = 0xEEEEEEEE;
+    while (*string) {
+        char c = *string++;
+
+        if (caseInsensitive)
```
mdX7, line 112: should be `if (!caseInsensitive)` according to wowdev.wiki

matanshukry: To clarify I reverted the condition since I don't like double negative (variable named `not_` and then using boolean not).
But I also should of passed true rather than false to this method, which I now did.

Warpten:
Was researched by me and schlumpf, ref is ok (but is accurate up to magic since that's when they moved chat translate to the server)

src/server/game/Chat/LanguageMgr.h
```diff
@@ -0,0 +1,56 @@
...
+#include "Define.h"
+#include "Language.h"
+#include <string>
+#include <unordered_map>
+
+class TC_GAME_API LanguageMgr
+{
+    typedef std::pair<uint32, uint8> WordKey;
+    typedef std::vector<char const*> WordList;
+    typedef std::unordered_map<WordKey, WordList> WordsMap;
```
Warpten, line 30: Try with an ordered_map and add a member to WordKey that is dbc Id, use that to sort keys, and see if that fixes bad translate

matanshukry: What would be the usage of it then?

Currently I'm not using any order of WordsMap but finding exactly what I need, so sorting it won't matter at the moment.

matanshukry: I tried to change the WordList to include the dbc id of the word and then sorted all the WordList after loading them.
Sadly it didn't change anything.

Warpten: That's how the old client did it. Unfortunately if that changed due to a DBC reorder that may mean they changed the hash function or bucket size.

matanshukry: I remember it looks very similar between the 2 versions, so I went back and tried to see what's different; I then noticed it's exactly the same (for language id 1, word group of length 3).

Found a bug in the code and now it's working :) Thanks!

Warpten:
Some bits of code from my old Legion-era implementation (Yes, I know, I never released it, hush).

<details>
```cpp
int32 ChatTranslate::SStrHash(const wchar_t* str, bool fixString /*=true*/)
{
    unsigned long seed1 = 0x7FED7FED;
    unsigned long seed2 = 0xEEEEEEEE;

    const wchar_t* c = str;
    while (*c != '\0')
    {
        int ch = *c++;
        if (fixString)
        {
            if (ch >= 0x61 && ch <= 0x7A) // [ 'a', 'z' ]
                ch -= 0x20;
            else if (ch == 0x2F)        // '/'
                ch = 0x5C;              // '\\'
        }

        seed1 = (hashTable[ch >> 4] - hashTable[ch & 0xF]) ^ (seed1 + seed2);
        seed2 = ch + 0x21 * seed2 + seed1 + 3;
    }

    if (seed1 == 0)
        return 1;

    return seed1;
}
```

```cpp
    // Bit of code that organizes words in buckets
    for (LanguageWordsEntry const* languageEntry : sLanguageWordsStore)
        _languageWords[languageEntry->Language][strlen(languageEntry->Word)].push_back(languageEntry);
```

```cpp
std::vector<LanguageWordsEntry const*> const* DB2Manager::GetLanguageWords(Language language, uint32 wordSize) const
{
    auto itr = _languageWords.find(language);
    if (itr != _languageWords.end())
    {
        auto itr2 = itr->second.find(wordSize);
        if (itr2 != itr->second.end())
            return &itr2->second;
    }

    return nullptr;
}
```

And to find a word replacement:

```cpp
        uint32 wordHash = SStrHash(word.c_str());

        std::vector<LanguageWordsEntry const*> const* languageWords = nullptr;
        for (; wordLength >= 1 && languageWords == nullptr; --wordLength)
            languageWords = sDB2Manager.GetLanguageWords(targetLanguage, wordLength);

        if (languageWords == nullptr)
            continue;

        std::wstring newWord;
        if (!Utf8toWStr((*languageWords)[wordHash % languageWords->size()]->Word, newWord))
            continue;
        
        replacementMap[word] = newWord;
```
</details>

matanshukry:
> ```c++
> SStrHash
> ```

The main differences I see are:
1. SStrHash is different; only by this line though: `seed2 = ch + 0x21 * seed2 + seed1 + 3;`. Seems like very different to me actually
2. using `Utf8toWStr` at the end.

However it seems the current code I implemented based on your early code in wowdev.wiki is working.
Can you give me an example of where it doesn't work, where the new code will fix it?

src/server/game/Chat/LanguageMgr.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    std::stringstream result;
+    Tokenizer tokens(msg, ' ');
+    bool first = true;
+    for (char const* str : tokens)
+    {
+        uint32 wordHash = SStrHash(str, true);
+        const char* nextPart = str;
+        if (wordHash % 300 < targetPlayerLanguageSkill)
+            continue;
```
matanshukry, line 83: Shouldn't go to next token but rather add current nextPart (str) to `result`.

matanshukry: nvm, ended up fixing it.

Shauren:
Is the translation part here complete? Asking because of this point in #26329
> Include translation PR #26271 (but fully)
What does "but fully" mean?

matanshukry:
> Is the translation part here complete? Asking because of this point in #26329
> 
> > Include translation PR #26271 (but fully)
> > What does "but fully" mean?

I don't remember the difference by now, but looking at the code I can find at least one place where this code isn't correct (might even be a bug iirc). I'm simply going to close this and add the mentioned PR to #26329

PR #26290, opened 2021-03-21
Core/Player: Fix output of .additem command with negative count value

r4d1sh:
**Changes proposed:**

Now we have a problem with .additem command output.

Steps to reproduce:
1. .additem 43102 10
2. .additem 43102 -1
3. get error "No such item exists in target's inventory"

This is because `Player::DestroyItemCount` returns an invalid value (most often this is zero for cases where the player has more items than needs to be removed)

This pull request will fix this bug.

- [x] 3.3.5
- [ ] master

**Tests performed:**

Builded and tested (3.3.5)


jackpoz:
Thanks for the PR :)

PR #26294, opened 2021-03-22
Scripts/Quest: Rework 'Basic Orders' (25199)

offl:
**Changes proposed:**

Wanted to rewrite script but turned out it's partially handled by spells which are supposed to target Gnomeregan Trainee and force them to perform emotes linked to specific spells(or remove emote) (previously handled in SAI)
Then turned out spells has no radius at all in DBC so I added ApplySpellFix thing with values from wowhead

And yes, auras are removed almost instantly after being applied and after creature performs emote in sniff, so I guess they are removed after first periodic tick(sniff without milliseconds)

Other changes are related to SAI only:
now instead of messing with phases we are using original auras to create conditions(player should perform specific emote in a given time when NPC asks and credit can be received only if aura is still active)
slightly adjusted timers and added missing actions(emotes)

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

none

**Tests performed:**

Builds, tested in-game, works


bakoorahnin:
very good

PR #26321, opened 2021-03-26
Core/Maps: increase value of fixed offset when retrieving floor Z

Jildor:
**Changes proposed:**

-  increase value of fixed offset when retrieving floor Z

well here is explained a bit: #26092
after that change, you can fall under textures more frecuently in certain situations, because before with collisionHeight we have higher value than 0.5f

I added a extra print collisionHeight in ".npc info" and ".pinfo" commands to see values, karazhan boss (malchezaar) has 8 collisionHeight, players varies depending on race, about 2 collisionHeight (gnome 1.2, tauren 2.3), can be higher/smaller depending on scale (.mod scale or auras that modify scale)
Z_OFFSET_FIND_HEIGHT = 0.5f is not sufficient, so I increase to 1.5f to avoid this problems

- [X] 3.3.5
- [ ] master

**Tests performed:**

tested all cases described here: #26092

### DIFF
--- a/src/server/shared/SharedDefines.h
+++ b/src/server/shared/SharedDefines.h
@@ -24,7 +24,7 @@
 #include "SmartEnum.h"
 
 float const GROUND_HEIGHT_TOLERANCE = 0.05f; // Extra tolerance to z position to check if it is in air or on ground.
-constexpr float Z_OFFSET_FIND_HEIGHT = 0.5f;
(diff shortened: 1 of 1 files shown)


jackpoz:
Good change for the time being, when I will have time I will check the ICC issue you mentioned

jackpoz:
Thanks for the PR :)

PR #26326, opened 2021-03-27
Core/Maps: Change .map file version from FourCC to uint32

jackpoz:
**Changes proposed:**

-  Core/Maps: Change .map file version from FourCC to uint32
- Re-extracting maps is required (not vmaps/mmaps)

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Allows to easily increment map version

**Tests performed:**

Extracted maps and loaded them.
Loading the current maps without re-extracting them shows this error:
>Map file './maps/0004331.map' is from an incompatible map version (MAPS v959328630), MAPS v9 is expected. Please pull your source, recompile tools and recreate maps using the updated mapextractor, then replace your old map files with new files. If you still have problems search on forum for error TCE00018.

None

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -55,7 +55,7 @@
 BOOST_1_74_FIBONACCI_HEAP_MSVC_COMPILE_FIX(RespawnListContainer::value_type)
 
 u_map_magic MapMagic        = { {'M','A','P','S'} };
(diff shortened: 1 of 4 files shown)


Faq:
Just wondering why v9, not v2?

jackpoz:
because the current version is v1.9 and uint32 doesn't support decimal numbers

PR #26348, opened 2021-04-06
Core/Movement: Improve client control logic

chaodhib:
This PR serves as a necessary prelude for [Core/Movement: Improve player movement ](https://github.com/TrinityCore/TrinityCore/pull/18771).

[Core/Movement: Improve player movement ](https://github.com/TrinityCore/TrinityCore/pull/18771) needed proper client transfer control. Up until now, two fields served that purpose: `Unit::m_playerMovingMe` and `Player::m_unitMovedByMe`. There are 2 issues with them:
* They are difficult to understand and reason with (just look at the current implementation of `Player::SetMovedUnit()` if you don't believe me).
* They are not updated when client A looses control over Player B because Player B is under the effect of a CC such as Sheep or Fear. `player_B->m_unitMovedByMe` will still points toward player_B. What is missing here is what I call an API that gives the "real-time client control status" of player_B.

In order to fix these problems, this PR introduces 3 notions:
1. **Base client control**: assuming I'm correct (please correct me if I'm wrong here), there are only 2 ways a client can have 'client control' over a unit in the game: a) the unit is a player and the client logs into the world with that player (I call this Player unit, the "base player" of that client). b) a (base) player charms another player or unit (this includes a Priest MCing another unit or a Player mounting a vehicle). Therefore, by looking at a unit's `UNIT_FIELD_CHARM` and `UNIT_FIELD_CHARMEDBY` values, we can tell who/what is **really** moving that unit. [1]

2. **Real-time client control**: On top of the base client control, the server will send to the client `SMSG_CONTROL_UPDATE` messages when the client should loose/regain control over a unit because a CC is applied/removed from that unit. A client could therefore have base client control but temporarily loose actual control while the CC is active. I call this real-time client control. Knowing the "real-time client control status" of a unit is needed when we need to change the speed of a unit for example: if the server controls the unit, we need to send a `SMSG_MOVE_SPLINE_SET_RUN_SPEED` message to the client to inform of the speed of change. if the client has real-time client control, the server needs to send a `SMSG_FORCE_RUN_SPEED_CHANGE` message (and expect an ACK response).

3. A **GameClient entity**. Instead of `player_B->m_playerMovingMe` pointing to `player_B`, we now have `gameClient_A->_activeMover` pointing to `player_B`. And if control is lost, `gameClient_A->_activeMover` is `NULL`. Same for `player_B->m_unitMovedByMe` which also points to `player_B`. Instead, we have `player_B->_gameClientMovingMe` which points to `gameClient_A`. This is much more understandable than the cyclic references `m_playerMovingMe` and `m_unitMovedByMe`.

Base client control API:
- Unit::isCharmerOrSelfPlayer
- Unit::GetCharmerOrSelf()
- Unit::GetCharmerOrSelfPlayer
- Unit::GetCharmedOrSelf()

Real time client control API:
* Unit::IsMovedByClient()
* Unit::IsMovedByServer()
* Unit::GetGameClientMovingMe()
* GameClient::GetActiveMover()
* GameClient::IsAllowedToMove(Unit* unit)

Real time client control is implemented using the exchange of the following packets:
1. S->C `SMSG_CONTROL_UPDATE`
2. C-> `CMSG_SET_ACTIVE_MOVER` and/or `CMSG_MOVE_NOT_ACTIVE_MOVER`

[1] If the unit is a creature
* and if `UNIT_FIELD_CHARMEDBY == 0`, then nobody has "client control" over that unit (in other words, the server has control over that unit).
* if `UNIT_FIELD_CHARMEDBY` is set, then that player has "client control" over that unit (vehicle or mind control).

If the unit is a player
* and if `UNIT_FIELD_CHARMEDBY == 0`, then we are in a "base player" situation.
* if `UNIT_FIELD_CHARMEDBY` is set, then that player has "client control" over that unit (e.g. mind control).

- [X] 3.3.5
- [ ] master

**Tests performed:**

Tested in game. Vehicles work. Weird scenarios like `player_A (a priest) MCs player_B and then player_C fears player_B` also work.


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -8627,15 +8654,17 @@ void Unit::SetSpeedRate(UnitMoveType mtype, float rate)
...
-        ++ToPlayer()->m_forced_speed_changes[mtype];
+        ++m_forced_speed_changes[mtype];
 
         if (!IsInCombat())
             if (Pet* pet = ToPlayer()->GetPet())
                 pet->SetSpeedRate(mtype, m_speed_rate[mtype]);
     }
 
-    if (Player* playerMover = Unit::ToPlayer(GetUnitBeingMoved())) // unit controlled by a player.
+    if (IsMovedByClient()) // unit controlled by a player.
```
Faq, line 8664: This just sounds weird. I mean the naming "IsMovedByClient" + "controlled by player". So according naming, client is in control not player :)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -26960,3 +26955,8 @@ std::string Player::GetDebugInfo() const
     sstr << Unit::GetDebugInfo();
     return sstr.str();
 }
+
+GameClient* Player::GetGameClient() const
+{
+    return GetSession()->GetGameClient();
```
jackpoz, line 26961: should there be a null check for GetSession() ? are there cases where a Player entity still exists while the session doesn't anymore ? I don't remember

jackpoz:
does this just need to be tested, merged and then every new bug will be fixed shortly after that as they get reported ?

src/server/game/Entities/Player/Player.cpp
Shauren: no

chaodhib:
Yes. Let's first see if problems can be found through code review. Then in 2 weeks or so, let's merge it and I will take care of the bug reports that emerge.

src/server/game/Entities/Unit/Unit.cpp
chaodhib: Context is key here. "MovedByClient" emphasis that the unit is not "MovedByServer". Also, if a (human) player controls a game client which controls a unit then by the transitive property of control, the human player controls the unit and the naming still works :)

illfated: Also, even though it sounds a bit cheesy, you could choose to view the player as client of the server operator (i.e. Blizzard).

ccrs:
love to see you back @chaodhib great work! ❤️

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -6270,6 +6273,30 @@ bool Unit::isPossessing() const
         return false;
 }
 
+bool Unit::isCharmerOrSelfPlayer() const
```
ccrs, line 6276: why the lowerCamelCase ?

chaodhib: It's a typo. I'll fix it. Thanks!

chaodhib:
@ccrs Thank you. It's good see familiar ~~faces~~  fellow members :)

jackpoz:
@chaodhib do you need this hosted at https://tcubuntu.northeurope.cloudapp.azure.com/test/ for some days ?

chaodhib:
Sure but without real players or testers on the server, that will be of limited value

chaodhib:
If there is no other comments, let's merge it?

ccrs:
I'd like to take a look if its not too much trouble.
Give me a couple days and if not just ignore me 💃

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -6270,6 +6273,30 @@ bool Unit::isPossessing() const
...
+
+Unit* Unit::GetCharmerOrSelf()
+{
+    if (IsCharmed())
+        return GetCharmer();
+    else
+        return this;
+}
+
+Player* Unit::GetCharmerOrSelfPlayer()
```
Treeston, line 6292: `return ToPlayer(GetCharmerOrSelf())`, this can be inline in the header

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -6270,6 +6273,30 @@ bool Unit::isPossessing() const
         return false;
 }
 
+bool Unit::IsCharmerOrSelfPlayer() const
```
Treeston, line 6276: isn't this just `GetCharmerOrSelf()->IsPlayer()`?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12163,23 +12179,13 @@ void Unit::UpdateObjectVisibility(bool forced)
...
-    }
-
-    if (!player)
+    if (IsMovedByServer())
     {
         GetMotionMaster()->MoveKnockbackFrom(x, y, speedXY, speedZ);
     }
     else
     {
+        Player* playerMover = GetGameClientMovingMe()->GetBasePlayer();
```
Treeston, line 12188: so, is a "mover" the player moving me...?

src/server/game/Server/GameClient.h
```diff
@@ -0,0 +1,52 @@
...
+        GameClient(WorldSession* sessionToServer);
+
+        void AddAllowedMover(Unit* unit);
+        void RemoveAllowedMover(Unit* unit);
+        bool IsAllowedToMove(Unit* unit) const;
+        bool IsAllowedToMove(ObjectGuid guid) const;
+        void SetMovedUnit(Unit* target, bool allowMove);
+
+        Unit* GetActiveMover() const { return _activeMover; }
+        void SetActiveMover(Unit* activeMover) { _activeMover = activeMover; }
```
Treeston, line 38: ...or is a "mover" something that is moved _by_ us?

This naming seems really confusing (very similar to the old naming scheme which was replaced in #17819). I think it's extremely likely to cause more issues over time. I'd recommend renaming.

chaodhib:
Good to see you again @Treeston! Thanks for the review!

src/server/game/Entities/Unit/Unit.cpp
chaodhib: indeed, fixed

src/server/game/Entities/Unit/Unit.cpp
chaodhib: No because the function is `const` and `GetCharmerOrSelf` is not

Treeston:
Well, you _did_ request one. I didn't really have time to look over the whole thing yet, but these are the things that stood out upon skimming.

src/server/game/Entities/Unit/Unit.cpp
chaodhib: Here the naming is bad. that Player unit might not even be moving. I changed it to basePlayer. We only need this reference to send the message to a specific client. It might be more readable if I were to add `SendDirectMessage` to GameClient as well

src/server/game/Server/GameClient.h
chaodhib: The active mover of a game client is the unit that a client is currently moving. The naming actually come from the opcodes `CMSG_SET_ACTIVE_MOVER ` and `CMSG_MOVE_NOT_ACTIVE_MOVER` so I feel like it is sensible.

The client sends these kind of messages when it starts/stops moving a unit (after it has been authorized/unauthorized to do so by the server using `SMSG_CONTROL_UPDATE`.

chaodhib:
I wasn't sure if you were still active in wow emulation. In any case, thanks for your time!

src/server/game/Entities/Unit/Unit.cpp
chaodhib: done

src/server/game/Entities/Player/Player.h
```diff
@@ -1652,6 +1652,7 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         void RemovedInsignia(Player* looterPlr);
 
         WorldSession* GetSession() const { return m_session; }
+        GameClient* GetGameClient() const;
```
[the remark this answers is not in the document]
chaodhib: fixed

src/server/game/Server/GameClient.h
Treeston: Even if the opcodes are called that, the name is still horribly un-intuitive. Do consider that this logic will be used by folks five years down the line, who won't be able to phone up you or me to explain it to them.

I'd recommend naming them `GetUnitBeingMoved` or something (isn't that what we currently have?)

#17819 was in part caused due to bugs, which had been caused by people assuming "mover" meant what one might intuitively think it means (the unit moving me).

src/server/game/Entities/Unit/Unit.cpp
Treeston: well, then why is `GetCharmerOrSelf` not const...?

src/server/game/Server/GameClient.h
chaodhib: What about `activeMovedUnit` ? Multiple opinions would be appreciated. I'm also fine with `activeMover` since a gameClient cannot be "moved" by a unit. Only a unit can be "moved" by a game client. Removing the cyclic reference already defuses a lot of confusion imo.

src/server/game/Entities/Unit/Unit.cpp
chaodhib: `GetCharmerOrSelf` calls `this`. Making the function `const` would make this returns const Unit* which is not what I want from the function.

Treeston: `const_cast` it away in `GetCharmerOrSelf`? seems like a perfectly valid use case for `const_cast`

src/server/game/Server/GameClient.h
```diff
@@ -38,6 +38,8 @@ class TC_GAME_API GameClient
         void SetActiveMover(Unit* activeMover) { _activeMover = activeMover; }
 
         Player* GetBasePlayer() const { return _sessionToServer->GetPlayer(); }
+
+        void SendDirectMessage(WorldPacket const* data) const { GetBasePlayer()->SendDirectMessage(data); }
```
jackpoz, line 42: this line breaks no-pch and I would say having it in .cpp is better than including Player.h in this header (no idea why people are obsessed of moving everything to headers as 1 line functions that are a pain to edit and debug with breakpoints)

src/server/game/Entities/Unit/Unit.cpp
chaodhib: Done

src/server/game/Server/GameClient.h
chaodhib: Done

jackpoz:
PR https://github.com/TrinityCore/TrinityCore/pull/26501 has been merged, thanks for waiting

src/server/game/Server/GameClient.h
chaodhib: Done. Renamed the field to `activelyMovedUnit`.

ccrs:
looking on it right now, though I have to say that there are several things to look into, the new ack opcode handling + gameclient + controlling revisit

src/server/game/Entities/Unit/Unit.h
```diff
@@ -1193,12 +1195,17 @@ class TC_GAME_API Unit : public WorldObject
         CharmInfo* InitCharmInfo();
         void DeleteCharmInfo();
 
-        // all of these are for DIRECT CLIENT CONTROL only
-        void SetMovedUnit(Unit* target);
-        // returns the unit that this player IS CONTROLLING
-        Unit* GetUnitBeingMoved() const { return m_unitMovedByMe; }
-        // returns the player that this unit is BEING CONTROLLED BY
-        Player* GetPlayerMovingMe() const { return m_playerMovingMe; }
+        // base client control of this unit (possess effects, vehicles and similar). Not affected by temporary CC.
```
ccrs, line 1198: I find this a bit concerning, meaning I at least dont know if this can cause any undesired effect.
Does the logic that uses these methods take into account the fact that the unit is actuallly still in client control, but temporally disabled due to a CC?
Is there a way to know if the unit is client controlled, even with any temporal CC aura?

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -26,12 +26,14 @@
...
 #include "MovementGenerator.h"
 #include "MoveSpline.h"
+#include "ObjectAccessor.h"
 #include "Transport.h"
 #include "Battleground.h"
 #include "InstanceSaveMgr.h"
 #include "ObjectMgr.h"
 #include "Vehicle.h"
 #include "GameTime.h"
+#include "GameClient.h"
```
ccrs, line 36: while you are at it you could order these alphabetically with worldsession on top 😝

src/server/game/Server/GameClient.h
```diff
@@ -0,0 +1,54 @@
...
+
+        Unit* GetActivelyMovedUnit() const { return _activelyMovedUnit; }
+        void SetActivelyMovedUnit(Unit* activelyMovedUnit) { _activelyMovedUnit = activelyMovedUnit; }
+
+        Player* GetBasePlayer() const { return _sessionToServer->GetPlayer(); }
+
+        void SendDirectMessage(WorldPacket const* data) const;
+    private:
+        // describe all units that this client has direct control over. Example, a player on a vehicle has client control over himself and the vehicle at the same time.
+        GuidSet _allowedMovers;
```
ccrs, line 45: why ordered set?

src/server/game/Server/GameClient.cpp
```diff
@@ -0,0 +1,63 @@
...
+void GameClient::RemoveAllowedMover(Unit* unit)
+{
+    _allowedMovers.erase(unit->GetGUID());
+    if (unit->GetGameClientMovingMe() == this)
+        unit->SetGameClientMovingMe(nullptr);
+}
+
+bool GameClient::IsAllowedToMove(Unit* unit) const
+{
+    return _allowedMovers.count(unit->GetGUID());
```
ccrs, line 44: container.find(element) != container.end() feels more appropiate

ccrs: https://en.cppreference.com/w/cpp/container/set/contains

available con c++20, not sure if we're on it

src/server/game/Server/Protocol/Opcodes.cpp
```diff
@@ -960,7 +960,7 @@ void OpcodeTable::Initialize()
     /*0x33D*/ DEFINE_SERVER_OPCODE_HANDLER(SMSG_MOTD,                      STATUS_NEVER);
     /*0x33E*/ DEFINE_SERVER_OPCODE_HANDLER(SMSG_MOVE_SET_CAN_TRANSITION_BETWEEN_SWIM_AND_FLY, STATUS_NEVER);
     /*0x33F*/ DEFINE_SERVER_OPCODE_HANDLER(SMSG_MOVE_UNSET_CAN_TRANSITION_BETWEEN_SWIM_AND_FLY, STATUS_NEVER);
-    /*0x340*/ DEFINE_HANDLER(CMSG_MOVE_SET_CAN_TRANSITION_BETWEEN_SWIM_AND_FLY_ACK, STATUS_NEVER, PROCESS_INPLACE,  &WorldSession::Handle_NULL                     );
+    /*0x340*/ DEFINE_HANDLER(CMSG_MOVE_SET_CAN_TRANSITION_BETWEEN_SWIM_AND_FLY_ACK, STATUS_LOGGEDIN, PROCESS_THREADSAFE, &WorldSession::HandleMoveSetCanTransitionBetweenSwinAndFlyAck);
```
ccrs, line 963: why does it qualify as threadsafe? honestly don't know myself

src/server/game/Server/GameClient.cpp
jackpoz: We are on C++17

src/server/game/Server/GameClient.h
chaodhib: Because I needed a data structure to hold guids without any duplicates.  also, i don't really care about the order on which the guids were inserted. both of these requirements make a good case for using a set.

src/server/game/Server/Protocol/Opcodes.cpp
chaodhib: it's like the other movement packet handlers: a thread safe message is a message which can be handled in one of the Map update threads while other Map updates are running concurrently. Basically, everything that only concern things happen in one map (for example dealing damage to a unit or movement) is thread safe. Things that cross map boundaries (chat, mail, LFG, ...) are non thread safe.

src/server/game/Entities/Unit/Unit.h
chaodhib: > Does the logic that uses these methods take into account the fact that the unit is actuallly still in client control, but temporally disabled due to a CC?

The PR aim is to not alter any behaviour in that regard. It is only to introduce functions that give the 'real time client control' status of a unit (IsMovedByClient() for example). These functions are needed to properly implement player movement.

> Is there a way to know if the unit is client controlled, even with any temporal CC aura?

Yes, by getting the "base client control" status of the unit; IsCharmerOrSelfPlayer()

src/server/game/Handlers/MovementHandler.cpp
chaodhib: done

src/server/game/Server/GameClient.h
ccrs: then how about moving into unordered? 😃

chaodhib: Oh I see. Coming from Java, I thought the standard std::set was unordered mb. Thank you. I changed it.

Treeston: tbf, the asymptotic runtime improvement for a set that's going to hold single-digit number of GUIDs at best is utterly irrelevant 😛

chaodhib: True of course :) It's still useful for future devs to know that the ordering doesn't matter in that data struct.

chaodhib:
I'm merging. Thank you for the review!

w5860363:
https://github.com/TrinityCore/TrinityCore/issues/27563

PR #26358, opened 2021-04-09
Core/Entities: Move GridObject to its own header

chipzz:
**Changes proposed:**

-  Move GridObject to its own header

- [x] 3.3.5
- [ ] master

**Issues addressed:**

**Tests performed:**

Did a rebuild

### DIFF
--- a/src/server/game/Entities/Corpse/Corpse.h
+++ b/src/server/game/Entities/Corpse/Corpse.h
@@ -19,6 +19,7 @@
 #define TRINITYCORE_CORPSE_H
 
 #include "Object.h"
+#include "GridObject.h"
 #include "DatabaseEnvFwd.h"
 #include "GridDefines.h"
 #include "Loot.h"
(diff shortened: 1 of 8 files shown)


Faq:
Expected benefit from these changes?

chipzz:
It's a header reorganization, so no direct performance or stability improvements are to be expected. The main point of this is to make TrinityCore more 3rd party friendly, so it's easier to copy and embed parts of it, and potentially long-term turning parts of TrinityCore into a library you can link to.
Since Object.h is included (indirectly) in practically every file, this should also improve build times for non-PCH builds.
When I was trying to embed Object.h in my own project, without this I was at 59 files and counting and ~20K LOC before I gave up and made this change. With this change, it's closer to 21 files and ~10K LOC.
Additional changes are needed for the .cpp files, but I will address that in a separate issue.

jackpoz:
Thanks for the PR :)

PR #26352, opened 2021-04-08
Core/Misc: Fixed utf8 encoding in console input/output.

r4d1sh:
**Changes proposed:**

- Correct translation from console encoding to utf8 and vice versa
- Fixed the possibility of console input and output for at least Russian on Windows

**Target branch(es):** 3.3.5/master

- [x] 3.3.5
- [ ] master

**Tests performed:**

Builded and tested on Windows

PR #26402, opened 2021-04-14
Core/Console: Improve ReadWinConsole logic and cosmetic changes

r4d1sh:
**Changes proposed:**

As @silviu20092 pointed out, when the worldserver is shutting down, it crashes. I tried to investigate the problem and found out that interrupting waiting for console input by calling `CancelSynchronousIo` does not leads to return the corresponding error, and `ReadConsoleW` returns true as if the console input was successfully read. I don't know if this is a Microsoft bug, but people are saying something similar here: https://github.com/microsoft/terminal/issues/334

Fortunately, we can determine if the waiting for console input is canceled by the unchanged variable `numberOfCharsRead`, even though ReadConsoleW returns no errors.

Anyway, crash was already fixed by @Shauren in 677952f. I am just suggesting a logical improvement to the ReadWinConsole function. A situation where ReadConsoleW did not read anything should be considered an error and ReadWinConsole should return false.

Also I removed one unused variable and one unnecessary CLI prefix printing.

- [x] 3.3.5
- [ ] master

**Tests performed:**

Builded and tested on Windows


ccrs:
Would you like to take a look @Shauren ?

r4d1sh:
Problem with weird characters reported by @silviu20092 (https://github.com/TrinityCore/TrinityCore/pull/26352#issuecomment-821985111) fixed in 3c4ffd1.

If anyone is wondering, the problem was with `OemToCharBuffW` function that was called from` consoleToUtf8` and its interaction with CRLF.

Some German programmer writes about it here: http://blog.m-ri.de/index.php/2011/08/22/oemtocharbuffw-macht-nicht-das-gleiche-wie-oemtocharbuffa-und-anschliesender-umwandlung-nach-unicode/

src/server/worldserver/CommandLine/CliRunnable.cpp
```diff
@@ -152,18 +149,13 @@ void CliThread()
 
         if (!command.empty())
         {
-            std::size_t nextLineIndex = command.find_first_of("\r\n");
-            if (nextLineIndex != std::string::npos)
+            auto nextLineIndex = RemoveCRLF(command);
```
Shauren, line 152: do not use auto

r4d1sh:
A few more possible errors leading to the appearance of weird characters in the console fixed in In 36717c7.  All the problems I knew about should be fixed. If there are no more known issues, the PR is ready to merge.

sernels:
see https://github.com/TrinityCore/TrinityCore/issues/26463

sernels:
pls use virtual machine after change

r4d1sh:
> see #26463
> pls use virtual machine on test after change (virtualBOX)

Something similar happened on Windows before my fixes. These changes should not affect the server behavior on UNIX systems. Most likely, this behavior was before my changes.

Could you check this with an earlier revision like 6bdfa91f?

Shauren:
To fix printing Difficulty variables, you have to cast them to uint32/int32, not uint8/int8

r4d1sh:
> To fix printing Difficulty variables, you have to cast them to uint32/int32, not uint8/int8

For technical or ideological reasons? I checked the Trinity :: StringFormat function, and it turned out that it works with uint8, but it does not work with enumerations, even if their type is explicitly defined.

Shauren:
For compatibility with c printf reasons, int8/uint8 are treated as chars instead of numbers

jackpoz:
anything left to do in this PR ?

r4d1sh:
If there are no other comments, everything is ready.

jackpoz:
this was not tested on linux right ?

r4d1sh:
> this was not tested on linux right ?

Not tested for Linux, but changes should only affect Windows.

jackpoz:
personal to do: test it on linux using docker image from https://app.circleci.com/pipelines/github/TrinityCore/TrinityCore/4426/workflows/072486cb-94cc-4ac4-a433-f9db291838ca/jobs/16092/artifacts

jackpoz:
>TC> account create åöä åöä
TC> An account with this name already exists!
TC> account delete åöä
TC> Account deleted: åöä
TC> account create åöä åöä
TC> Account created: åöä
TC> account delete åöä
TC> Account deleted: åöä
Halting process...

looks like it's working on Ubuntu

PR #26412, opened 2021-04-14
Move WorldSession::WriteMovementInfo to MovementInfo

chipzz:
**Changes proposed:**

-  Move WriteMovementInfo method out of WorldSession

- [x] 3.3.5
- [ ] master

**Issues addressed:**

**Tests performed:**

Rebuilt

- This doesn't belong in WorldSession (and neither does ReadMovementInfo for that matter, but it has a dependency on Player, so I'll address that in a separate PR). Where it needs to end up and in what shape/ form should be discussed. As a first step I've made it a simple member method in the header file, but it could be operator << too.

### DIFF
--- a/src/server/game/Entities/Object/MovementInfo.h
+++ b/src/server/game/Entities/Object/MovementInfo.h
@@ -20,6 +20,8 @@
 
 #include "ObjectGuid.h"
 #include "Position.h"
+#include "WorldPacket.h"
+#include "UnitDefines.h"
 
 struct MovementInfo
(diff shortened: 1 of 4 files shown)


Shauren:
No. Any moving packet building code should go into "packet classes", like this https://github.com/TrinityCore/TrinityCore/blob/9619f7c2832c550d4606410a8cdbde5ee54c761d/src/server/game/Server/Packets/MovementPackets.cpp#L25

PR #26426, opened 2021-04-17
Scripts/RubySanctum: Improve Halion boundary

Jildor:
**Changes proposed:**

-  adjust Halion circleBoundary size
boundary radius was a little smaller (tested with .debug boundary command)

-  improve DoCheckEvade
in this commit: https://github.com/TrinityCore/TrinityCore/commit/0b33a8281f6be3458ef3698349f00a1f6aabb0b5#diff-4dbcd97850a76ba57320148f9e3ed30e52f4302b362a63ddddb1a1007cb14f50R760-R770 an event was added for check if all alive players are inside boss combat zone, otherwise, evade and despawn.
But npc_halion_controller doesn't have boundary, so that check does not work properly

- [X] 3.3.5
- [ ] master

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/scripts/Northrend/ChamberOfAspects/RubySanctum/boss_halion.cpp
+++ b/src/server/scripts/Northrend/ChamberOfAspects/RubySanctum/boss_halion.cpp
@@ -572,6 +572,7 @@ class npc_halion_controller : public CreatureScript
                 _materialCorporealityValue = 5;
                 _materialDamageTaken = 0;
                 _twilightDamageTaken = 0;
+                SetBoundary(_instance->GetBossBoundary(DATA_HALION));
             }
 
             void JustAppeared() override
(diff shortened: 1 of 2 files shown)


src/server/scripts/Northrend/ChamberOfAspects/RubySanctum/boss_halion.cpp
```diff
@@ -572,6 +572,7 @@ class npc_halion_controller : public CreatureScript
                 _materialCorporealityValue = 5;
                 _materialDamageTaken = 0;
                 _twilightDamageTaken = 0;
+                SetBoundary(_instance->GetBossBoundary(DATA_HALION));
```
jackpoz, line 575: is it common to call SetBoundary() in the constructor or what do the other scripts do ?

jackpoz: it seems so

jackpoz:
Thanks for the PR :)

src/server/scripts/Northrend/ChamberOfAspects/RubySanctum/boss_halion.cpp
Jildor: Yep

PR #26448, opened 2021-04-18
Core/Creature: Create & use CREATURE_FLAG_EXTRA_IGNORE_FEIGN_DEATH

offl:
**Changes proposed:**

Currently CREATURE_FLAG_EXTRA_GUARD is responsible for IsGuard (PERMIT_BASE_PROACTIVE & checks if creature can detect feign death)
This change creates separate flag CREATURE_FLAG_EXTRA_IGNORE_FEIGN_DEATH because now we have that corresponding difficulty flag and because using one flag for multiple completely different tasks(now we know not only guards can ignore feign death) is wrong

***

That flag does not what I expected because it simply allows creature attack affected by feign death target. But if you use feign death mid combat, creature simply will evade and that looks logically wrong to me(especially because creature re-enters in combat with you after reaching home if you're too close)
To me it looks like something directly in spell effect should be adjusted but that's something for another time

Also original flag probably is not only about 'can attack or not' because some of creatures with that flag can't attack anyone in first place. For example that flag is used by all triggers which calls guards on you but they can't attack you. Looks like they still should call guards on you even if you are affected by feign death

- [x] 3.3.5
- [ ] master

**Issues addressed:**

none

**Tests performed:**

Builds, tested before & after

### DIFF
--- a/sql/updates/world/3.3.5/2021_04_18_55_world.sql
+++ b/sql/updates/world/3.3.5/2021_04_18_55_world.sql
@@ -1,0 +1,400 @@
+--
+UPDATE `creature_template` SET `flags_extra` = `flags_extra` | 0x00010000 WHERE `entry` IN
+(68, -- Stormwind City Guard
+1756, -- Stormwind Royal Guard
+2614, -- Air Force Alarm Bot (Alliance)
+2615, -- Air Force Alarm Bot (Horde)
+3083, -- Honor Guard
+3084, -- Bluffwatcher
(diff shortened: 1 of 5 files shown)


ccrs:
looks good, any other missing creature entry that you can think of? vanilla bosses or something like that

offl:
That flag is used in CreatureDifficulty table, it's original data but from MoP
So there may be missing entries and some may be added only in MoP but it's fine in this specific case and we can't check if some should be added or removed

PR #26459, opened 2021-04-21
Scripts/Naxxramas: Slightly update Grobbulus

offl:
**Changes proposed:**

-  add missing emotes
-  summon Fallout slimes by spell instead of manually and add 25 mode spell SPELL_SLIME_SPRAY_H to SpellHitTarget to restore broken summon after one of old changes(RAID_MODE was removed, so no summon in 25 mode, tell me if you want RAID_MODE instead)
-  create script for Fallout slimes, it does everything I saw in sniff but I doubt AddThreat is used always, it's hard to figure out why it was shown in sniff and why they behaves differently in movies. Can be changed to DoZoneInCombat or something like that
-  move npc_grobbulus_poison_cloud to NullCreatureAI, it's better than changing react state to passive and call other old stuff
-  codestyle
-  timers

> RAND(25700ms, 27700ms, 30000ms));

This may look weird but I call them 'set of timers'. Some time ago I noticed that timers are not just random, there are multiple timers and they repeats from sniff to sniff. Sets of timers are used everywhere but it's really hard and not really needed to find them. Moreover it's quite dumb to use them without checking 30+ timers of each spell. In this particular case I got enough casts(not for SLIME_SPRAY) but still can't figure out why few timers are completely different from other commonly used

They're mostly weird, like 3300, 2800, 15400, 18800. Sometimes like 5000, 10000, 15000

Moreover... it doesn't look like they're just randomized, each timer has its own chance to be used...
Anyway, this is something from another universe, added as an experiment

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

none

**Tests performed:**

Build, tested in-game


src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
```diff
@@ -15,35 +15,51 @@
  * with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+/*
+ * Slime Spray timers must always looks like ~30 (initial) ~30 (first repeat) 55 (second repeat) ~30 55 ~30 55 ~30 55 ~30 55
+  so after second cast timers are always flipped
+ * SPELL_BOMBARD_SLIME should be used out of combat (waypoint script?)
+ * He should call all stitched giants after aggro
+ */
```
ccrs, line 23: usually these type of comments are put near the thing it references, not a the top

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
```diff
@@ -15,35 +15,51 @@
...
+
 #include "ScriptMgr.h"
 #include "ScriptedCreature.h"
+#include "PassiveAI.h"
+#include "ObjectAccessor.h"
 #include "SpellScript.h"
 #include "SpellAuraEffects.h"
 #include "naxxramas.h"
```
ccrs, line 31: while you are at it, you could also reorder these header includes as intended, that is shared header on top an the rest on alphabetical. This means ``naxxramas.h` should be on top.

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
```diff
@@ -53,16 +69,16 @@ struct boss_grobbulus : public BossAI
...
+        events.ScheduleEvent(EVENT_INJECT, 12s, 15s);
+        events.ScheduleEvent(EVENT_SPRAY, RAND(25700ms, 27700ms, 30000ms));
         events.ScheduleEvent(EVENT_BERSERK, 12min);
     }
 
     void SpellHitTarget(WorldObject* target, SpellInfo const* spellInfo) override
     {
-        if (spellInfo->Id == SPELL_SLIME_SPRAY)
-            me->SummonCreature(NPC_FALLOUT_SLIME, *target, TEMPSUMMON_TIMED_DESPAWN_OUT_OF_COMBAT);
+        if (spellInfo->Id == SPELL_SLIME_SPRAY || spellInfo->Id == SPELL_SLIME_SPRAY_H)
```
ccrs, line 80: nice catch, people usually forget to check difficulty entries.
though I recall a helper method to fetch the normal entry for a spell in an instance, given the difficulty entry...

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
```diff
@@ -101,22 +123,65 @@ struct boss_grobbulus : public BossAI
...
+        me->SetReactState(REACT_PASSIVE);
+        DoCastSelf(SPELL_DISEASE_CLOUD);
+
+        _scheduler.Schedule(2s, [this](TaskContext /*task*/)
+        {
+            me->SetReactState(REACT_AGGRESSIVE);
+
+            if (Unit* summoner = ObjectAccessor::GetUnit(*me, _summonerGUID))
+                // maybe only if threat list was empty?
+                AddThreat(summoner, 100000.0f);
```
ccrs, line 157: I recommend using the "FixateTarget" method, it's really cool.

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
```diff
@@ -101,22 +123,65 @@ struct boss_grobbulus : public BossAI
...
+    }
+
+    void IsSummonedBy(WorldObject* summoner) override
+    {
+        _summonerGUID = summoner->GetGUID();
+    }
+
+    void EnterEvadeMode(EvadeReason /*why*/) override
+    {
+        // they don't despawn when Grobbulus dies
```
ccrs, line 168: you can manually call Grobbulus summon hook to link each summon with the boss

something like `grobbulus->AI()->IsSummonedBy()`
I dont remember exactly which summon hook it is, maybe I'm wrong with this one

BossAI will take care to despawn and handle the summons

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
```diff
@@ -101,22 +123,65 @@ struct boss_grobbulus : public BossAI
...
+{
+    npc_fallout_slime(Creature* creature) : ScriptedAI(creature)
     {
-        SetCombatMovement(false);
-        creature->SetReactState(REACT_PASSIVE);
+        Initialize();
     }
 
-    void IsSummonedBy(WorldObject* /*summoner*/) override
+    void Initialize()
```
ccrs, line 145: IMO JustAppeared is a better fit for these initializations, you could drop the Initialize method altogether

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
ccrs: `JustSummoned` this is the one

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
offl: That's kinda more correct version of old `SDComment:`.  Right now I don't have a clear vision of how it's supposed to look. I still add some comments inside code but I think about something like 'global TO-DO per file'

All that stuff wasn't bad but always looked to me useless. I mean, one line for comments usually filled with comments like 'timers, something is broken'. TC mostly removes all old ScriptData but we still need some global TO-DO per file
```cpp
/* ScriptData
SDName: Boss_High_King_Maulgar
SD%Complete: 90
SDComment: Correct timers, after whirlwind melee attack bug, prayer of healing
SDCategory: Gruul's Lair
EndScriptData */
```

Not sure about slime timers but comment about SPELL_BOMBARD_SLIME can be moved to enum
And part about call for help cannot be moved anywhere

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
offl: They don't despawn when boss is killed because they are not related to boss in any way because they are summoned by player
Originally they just stuck after leaving combat(literally, they don't move), then they despawns after few seconds. I didn't recreated that behavior

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
offl: They are not fixated, just added 100k threat in one sniff but didn't in other(no clue why)
In movies they sometimes may move to another target(no clue why)

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
offl: >  better fit for these initializations

Even for SetCorpseDelay?

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
ccrs: thats why I recommended the manual hook call, place a getcreature from instance to get grobbulus 👍

offl: `grobbulus->AI()->JustSummoned(slime)` will despawn them after boss death

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
Ovahlord: you can set corpse delay anywhere, even in the ctor of the script. there is like no standard for it at all. JustAppeared() is not really a good spot for such initializations because it's being called after the first update tick of the creature instead of right at spawn. InitializeAI would be the correct spot for immediate changes.

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
ccrs: ah I finally realize what's the intended behaviour lul

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
ccrs: ummm true, JustAppeared is delayed one tick

ccrs: eitherway, IMO, Casts feel like they should be handled on JustAppeared
dont know, it really depends on each scenario

offl:
And this is why I don't like to touch headers

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
```diff
@@ -101,22 +122,63 @@ struct boss_grobbulus : public BossAI
...
+        me->DespawnOrUnsummon(61s);
+    }
+};
+
+struct npc_fallout_slime : public ScriptedAI
 {
-    npc_grobbulus_poison_cloud(Creature* creature) : ScriptedAI(creature)
+    npc_fallout_slime(Creature* creature) : ScriptedAI(creature) { }
+
+    void InitializeAI()
```
Jinnaix, line 141: missing override

Treeston:
the timer thing sounds like a consequence of AI tick rate, so I have some academic objections to hackfixing them this way in AI (plus I think it's wasted effort)

offl:
AI tick rate which repeats from sniff to sniff?
I think it's completely unknown territory for emu. We mostly use random timers but they're way much complicated sometimes. But yeah, no matter what it is, it's wasted effort, especially without parser setting which will parse all events and their cooldowns + really large sniffs(10+minute of fight)

src/server/scripts/Northrend/Naxxramas/boss_grobbulus.cpp
```diff
@@ -15,35 +15,48 @@
...
 };
 
-enum Events
-{
-    EVENT_BERSERK               = 1,
-    EVENT_CLOUD                 = 2,
-    EVENT_INJECT                = 3,
-    EVENT_SPRAY                 = 4
-};
+#define SPELL_SLIME_SPRAY RAID_MODE<uint32>(28157, 54364)
```
ccrs, line 52: `sSpellMgr->GetSpellIdForDifficulty(SPELL_UNBOUND_PLAGUE, GetCaster())``

this is what I meant

offl: Nice thing, thanks

offl:
If this will be merged soon, please don't forget to change sql name

ccrs:
if you tested ingame the last changes, it feels ready to me 👍

Treeston:
> AI tick rate which repeats from sniff to sniff?

yeah, blizz updates AI at set intervals, not every tick like we do; this will end up with certain "rounding" for AI-controlled actions, like spell casts. the result would end up looking like what you're seeing.

see some of the discussion on #22296 and #22461.

offl:
@Treeston, I'm not sure if I understood everything correctly but if after #22296 you can't receive 4 casts after exactly 3600ms and 600 casts with 1200ms(repeat timer) interval, then it's not correct simply because it will not match sniffs
But even if you adjust timers to 3500 and 1s and still will not receive always 3600 and 1200(especially if they will be completely random), then again something is wrong

Moreover some timers are exact values, without ms. Like 10 or 20 or 50. If with your change they will be 11 or 20600 or 53423, it's absolutely not correct

Hunters and casters(shoot spell, shadowbolt, firebolt, frostbolt) are interesting, there are hundreds of them and they all uses exactly same timers. They again look like 2400 3600 3000 3200 3800, these timers were used since TBC and were not even sniffed but gained from combat logs. Now I can confirm that these timers are absolutely correct but originally are used as sets of timers

PR #26485, opened 2021-04-26
Core: Replace unchecked Unit::GetVictim calls with Unit::EnsureVictim

HelloKitty:
**Changes proposed:**

- Use Unit::EnsureVictim everywhere Unit::GetVictim is referenced but not checked for nullptr

- [x] 3.3.5
- [ ] master

**Issues addressed:**

**Tests performed:**

Compiles

None

### DIFF
--- a/src/server/game/AI/PlayerAI/PlayerAI.cpp
+++ b/src/server/game/AI/PlayerAI/PlayerAI.cpp
@@ -657,7 +657,7 @@ PlayerAI::TargetedSpell SimpleCharmedPlayerAI::SelectAppropriateCastForSpec()
             }
             VerifyAndPushSpellCast(spells, SPELL_ENRAGED_REGEN, TARGET_NONE, 3);
             VerifyAndPushSpellCast(spells, SPELL_INTIMIDATING_SHOUT, TARGET_VICTIM, 4);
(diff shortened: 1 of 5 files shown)


Ovahlord:
I don't know what you mean that they are not checked for nullptr because in all your changed lines there literally IS a pointer check right before the action is being called.

HelloKitty:
@Ovahlord That's reasonable, but assumes that the calling code has complete knowledge of the implementation of Unit::GetVictim while I don't think it should be written with that assumption. If you're not going to assign the returned Unit pointer to a local variable then calling the asserting version of GetVictim should be preferred imo.

I am aware that the behavior does not actually change in these cases but this seems the better way.

Ovahlord:
I dunno, to me it seems like wasted performance and inconsistent code using two different helpers with one calling a macro that's completely useless in that scenario.

Shauren:
No. EnsureVictim only exists to shut up static analysis tools, you absolutely do not need it in places where GetVictim was already checked for nullptr

HelloKitty:
I disagree, the calling could should not have the assumption that it knows the exact implementation of Unit::GetVictim. It should not assume it returns the same exact Unit pointer even though we do know that it will. But I guess it really is a trivial matter and I don't care much about pursuing the matter further so I'll delete the branch.

PR #26486, opened 2021-04-26
Core/Map: Added Map::SetZoneOverrideLight std::chrono overload

HelloKitty:
**Changes proposed:**

-  Add new overload Map::SetZoneOverrideLight taking std::chrono milliseconds
- Replace existing usages with std::chrono literals using the new overload

- [x] 3.3.5
- [ ] master

**Issues addressed:**

**Tests performed:**

Compiles

None


Shauren:
If you are also switching existing uses to use chrono overload, is there any reason to keep the old one?

HelloKitty:
Hmmm, I can delete it but wasn't sure if I should delete the existing API. Want me to delete?

Shauren:
At no point in history we ever guaranteed api stability, go ahead and delete it

HelloKitty:
Okie dokie, I've gone ahead and remove it.

PR #26494, opened 2021-04-29
Core/Game: Fix dynamic library cache path with git flow branches

ihm-tswow:
**Description**
Some common git workflows (`git flow` specifically) typically has forward slashes in their branch names. Since we use branch names for cache directories, the directory may actually be multiple directories deep, so we need to use `create_directories` when we do to avoid errors.

**Changes proposed:**

- Change `create_directory` to `create_directories` when creating the temporary library directory.

- [x] 3.3.5
- [x] master

**Tests performed:**

Builds fine, and libraries load fine on 3.3.5 even with slashes in the branch names. master not tested, but `create_directories` should not fail where `create_directory` doesn't.


Shauren (approved):

jackpoz:
Thanks for the PR :)

Issue #25258, opened 2020-08-16
Convert script for They're Alive! Maybe to SAI 


dr-j:
https://www.wowhead.com/quest=9670/theyre-alive-maybe

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Kalimdor/zone_bloodmyst_isle.cpp#L44

Current CPP script has possible creature summons hard coded in script and does not use summon spells and there are some named creatures included which should not be part of this

These are all the possible creatures which can currently be spawned by cpp script for this quest

17322 = Infected Wildkin
17661 = Deathclaw
17496 = Cruelfin <Rawgrlgrlgrlgrlgrrgle>
17522 = Myst Spinner
17340 = Axxarien Shadowstalker
17352 = Corrupted Treant
17333 = Wrathscale Screamer
17524 = Nazzivus Summoner
17654 = Tagged Blacksilt Scout
17348 = Elder Brown Bear
17339 = Nazzivus Felsworn
17345 = Brown Bear
17359 = Tel'athion the Impure
17353 = Corrupted Stomper
17336 = Wrathscale Sorceress
17550 = Void Anomaly
17330 = Blacksilt Seer
17701 = Lord Xiz
17321 = Bristlelimb Warrior
17325 = Blacksilt Forager
17320 = Bristlelimb Shaman
17683 = Zarakh
17342 = Axxarien Hellcaller
17715 = Atoph the Bloodcursed
17334 = Wrathscale Marauder
17341 = Axxarien Trickster
17338 = Nazzivus Rogue
17337 = Nazzivus Satyr
17346 = Mutated Tangler
17344 = Mutated Constrictor
17327 = Blacksilt Tidecaller

These are the ones which there are actually free webbed creature spells in spell_dbc for 

17330 - Blacksilt Seer
17340 - Axxarien Shadowstalker
17342 - Axxarien Hellcaller
17346 - Mutated Tangler
17348 - Elder Brown Bear
17350 - Royal Blue Flutterer
17353 - Corrupted Stomper
17527 - Enraged Ravager
17607 - Sunhawk Defender
17608 - Sunhawk Pyromancer
17681 - Expedition Researcher

~~~sql
UPDATE `creature_template` SET `AIName`='SmartAI', `ScriptName`='' WHERE  `entry`=17680;
DELETE FROM `smart_scripts` WHERE `entryorguid`=17680 AND `source_type`=0;
[...]

PR #26554, opened 2021-05-25
Scripts/Quest: Rework cocooned creatures

offl:
**Changes proposed:**

Created spell scripts to make those quests work as correct as possible. They're quite similar so it'll be not hard to review all at once.
There are 2 types of spells - one is used on death of a cocoon if player has no quest and summons various hostile creatures. 
Second is too used on death but only if player has quest and summons various hostile creatures or quest creature. Also it targets player and forces him to summon quest creature because we(and Blizz) as usually want to make player owner of creature(to easily use player's guid in creature script)
I may edit sql later, so please just let me know if spell scripts are good to go

Ovahlord suggested to use std::array but I got error I wasn't able to fix
*Pretends he forgot to validate all spells like other people did*

**Target branch(es):** 3.3.5

**Issues addressed:**

none

**Tests performed:**

Builds, tested in-game


jackpoz:
Isn't there some RAND() macro to get a random spell from a container ? Instead of urand() as index

Ovahlord:
It's fine using an array here because they are just one ID apart and one spell only has a weird ID. It would make more sense to use RAND here if we had completely wild values here

illfated:
( back ref. #25258 )

src/server/scripts/Kalimdor/zone_bloodmyst_isle.cpp
```diff
@@ -786,12 +786,17 @@ class spell_free_webbed : public SpellScript
 
     bool Validate(SpellInfo const* /*spellInfo*/) override
     {
-        return ValidateSpellInfo(SummonSpells);
+        return ValidateSpellInfo(
+        {
+            SPELL_FREE_WEBBED_1, SPELL_FREE_WEBBED_2, SPELL_FREE_WEBBED_3, SPELL_FREE_WEBBED_4,
```
jackpoz, line 791: is this needed because ValidateSpellInfo doesn't support std::array ?

src/server/scripts/Kalimdor/zone_bloodmyst_isle.cpp
```diff
@@ -773,7 +773,7 @@ enum FreeWebbedBloodmyst
     SPELL_FREE_WEBBED_11     = 31010
 };
 
-uint32 const SummonSpells[10] =
+std::array<uint32, 10> const SummonSpells =
```
jackpoz, line 776: was this needed to use Trinity::Containers::SelectRandomContainerElement( ) ? it should work even with the old code

Ovahlord: no but we don't live under a rock anymore so yay for 'modern' code.

src/server/scripts/Kalimdor/zone_bloodmyst_isle.cpp
offl: Not supported or I don't know how to do it properly

src/server/scripts/Kalimdor/zone_bloodmyst_isle.cpp
jackpoz: @Ovahlord does ValidateSpellInfo() support std::array ?

Ovahlord: Nope. It's a list initializer which is a glorified vector so what he would have to do is copy the array content which is way too overkill for that check. Just slap a maintainer info on it to keep the enums and the array sync.

offl: As you said I will be not able to validate both spell list and one spell which is not inside spell list
Currently all spells are validated but manually, looks not that bad as I expected

jackpoz: I'd stick to the old C-style array and just call `ValidateSpellInfo(array) && ValidateSpellInfo({EXTRA_SPELL});`

Issue #18582, opened 2016-12-25
Core/Movement: Acknowledgement packets should be handled differently


chaodhib:
**Description:**
The issue has been described by @tripleslash [HERE](https://github.com/TrinityCore/TrinityCore/issues/16808#issuecomment-211121355) and [HERE](https://github.com/TrinityCore/TrinityCore/issues/16808#issuecomment-236642299).

I also tried to describe the problem using sequence diagrams:
[current situation](https://www.lucidchart.com/documents/view/7e8c8f20-c622-470a-b301-ce20778c46e6)
[expected behavior](https://www.lucidchart.com/documents/view/acc15c74-eb12-43b4-853b-a998dbd5dfff)

**Branch(es):** 3.3.5 (and master too I suppose)

**TC rev. hash/commit:**  
https://github.com/TrinityCore/TrinityCore/commit/e8314e2c4aa6f12f36daee60fc6933ecdd0fcb9b

RussellNixon6:
Delete

Rushor:
delete?

Aokromes:
this issue is still valid?

chaodhib:
Yes

Issue #26550, opened 2021-05-24
Mod Decrease Speed aura will not restore speed after


ckegg:
**Description:**
After Core/Movement: Improve client control logic, https://github.com/TrinityCore/TrinityCore/pull/26348, any spell with Mod Decrease Speed aura will not restore speed after, for example: [Blind](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=42972)

However, https://github.com/TrinityCore/TrinityCore/pull/26348 actually fixed https://github.com/TrinityCore/TrinityCore/issues/21044

**Expected behaviour:**
Speed restored after Mod Decrease Speed aura removed.

**Steps to reproduce the problem:**
1. .go creature id 23956
2. Let Dragonflayer Strategist hit you with [Blind](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=42972)
3. Your character will wander confused for 3 seconds plus speed decrease, and your speed will not restore after.

**Branch(es):**
3.3.5

**TC rev. hash/commit:** 
Commit: 302c60d3dd8e2734b32d239142c72b1add22aad9

**Operating system:** 
debian 10.4

jackpoz:
Assigning this to @chaodhib

chaodhib:
Thanks for the report. I'll have a look this week, maybe today if I can.

chaodhib:
Confirmed. I'm working on a fix. A proper fix involves some of the changes introduced in https://github.com/TrinityCore/TrinityCore/pull/18771. I'll see if I can quickly produce a PR with a smaller scope.

Aokromes:
d337fb9

jackpoz:
https://github.com/TrinityCore/TrinityCore/commit/d337fb99ed81dc3522bdea17e762febd989a6956

PR #26538, opened 2021-05-20
Core/Movement: Hotfix the crashes that can occur in the movement handlers

chaodhib:
Hotfix the crashes that can occur in the movement handlers since the latest client control logic rework

**Changes proposed:**

-  Clean up potentially dangling pointers when `GameClient::SetMovedUnit` is called.
- This PR is a hotfix in the sense that the full fix is not implemented yet. The change in src/server/game/Server/WorldSession.cpp should be temporary.

**Target branch(es):** 3.3.5/master

- [X] 3.3.5
- [ ] master

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/26531

**Tests performed:**

Tested IG. Feedback and testing from you guys is welcome.

Issue #21044, opened 2017-12-11
Core/Movements: Speed exploit


Killyana:
**Description:** If a player use Nitro Boosts or any temp speed boost then ride a vehicle, if the speed aura fades once the player is inside the vehicle he will keep the speed boost.

**Current behaviour:** Player speed is not reset if a speed boost aura fades once he's inside a vehicle.

**Expected behaviour:** Player speed must be reset in all cases once the aura fades.

**Steps to reproduce the problem:**

1. .go c id 33109  (requires gm on to enter the raid)
2. .cast 54861 (on your char)
3. ride the vehicle before the aura fades
4. Once the aura fades dismount and notice your char's speed.

**Branch(es):** both

**TC rev. hash/commit:**  rev. 2286da2166dd

**TDB version:**  CHANGEME Version of the TrinityCore database

**Operating system:** CHANGEME OS

Keader:
Idk if is the proper fix, but works.
```diff
diff --git a/src/server/game/Entities/Unit/Unit.cpp b/src/server/game/Entities/Unit/Unit.cpp
index bd4c71c06e..57467c0cc8 100644
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9400,7 +9400,7 @@ void Unit::SetSpeedRate(UnitMoveType mtype, float rate)
                 pet->SetSpeedRate(mtype, m_speed_rate[mtype]);
     }
 
-    if (Player* playerMover = GetPlayerBeingMoved()) // unit controlled by a player.
+    if (Player* playerMover = GetPlayerBeingMoved() ? GetPlayerBeingMoved() : ToPlayer()) // unit controlled by a player.
     {
         // Send notification to self. this packet is only sent to one client (the client of the player concerned by the change).
         WorldPacket self;
```

Keader:
@Shauren place is correct to do it? https://github.com/TrinityCore/TrinityCore/issues/21044#issuecomment-350610437

Jildor:
tested and works @Keader 
What are you waiting for?
 😄

Keader:
"Is not correct form to deal with it" they say...
xD

Jildor:
😿

PR #26561, opened 2021-05-28
Core/Movement: Implement proper player speed change

chaodhib:
**Changes proposed:**

Implement a part of the changes proposed in https://github.com/TrinityCore/TrinityCore/pull/18771: the speed changes

Taken from that PR, here are the changes visually (only speed changes were made smoother in this PR. Knock backs were kinda implemented already. The other types of changes will come in later PRs):
Before: https://youtu.be/KnsQ_Zqz0kQ
After: https://youtu.be/MwpWViAxhbs

- [X] 3.3.5

**Issues addressed:**

Closes: 
* https://github.com/TrinityCore/TrinityCore/issues/26550
* The issue mentioned [here](https://github.com/TrinityCore/TrinityCore/pull/26538#issuecomment-848529459)
* Partially closes https://github.com/TrinityCore/TrinityCore/issues/18582

**Tests performed:**

Tested IG. No issue found so far.

No known issue. However, the PR is still in WIP. But early feedback and **testing would be highly appreciated** !

EDIT: known issues found:
- [x] trying to reproduce https://github.com/TrinityCore/TrinityCore/issues/21044 results in a kick (with reason: "Took too long to ack a movement change"). EDIT: this is fixed now.


chaodhib:
@ ckegg @ shenhuyong testing would be highly appreciated!

shenhuyong:
@chaodhib Nice work, I tested it, Judgement of Justice (20184) works fine now.

src/server/worldserver/worldserver.conf.dist
```diff
@@ -4034,6 +4035,24 @@ PacketSpoof.BanMode = 0
...
+# Settings to tweak various anti cheat measures.
+#
+#    AntiCheats.PendingMoveChangesTimeoutTime
+#        Description: Time each client has to acknowledge a movement change until they are kicked.
+#        Note:        A lower time will give cheaters less advantage when using a very specific
+#                     type of hack. However, the cost is that innocent but laggy players might
+#                     sometimes get kicked.
+#        Default:     2000 (2 seconds)
+
+AntiCheats.PendingMoveChangesTimeoutTime = 2000
```
Ovahlord, line 4053: I think it's more consistent to not use plural here as we don't count anti-cheat as an entity but as a type of mechanic

chaodhib: Done

jackpoz:
could you please check https://github.com/TrinityCore/TrinityCore/issues/21044 too ? to see if it reappear after this PR.

chaodhib:
> 
> 
> could you please check #21044 too ? to see if it reappear after this PR.

@jackpoz Sure. It's in the "known issue / todo" section of the PR. trying to reproduce results in the player getting kicked. I'm looking into it.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -4507,13 +4507,14 @@ void Player::BuildPlayerRepop()
 {
     WorldPackets::Misc::PreRessurect packet;
     packet.PlayerGUID = GetGUID();
-    GetSession()->SendPacket(packet.Write());
+    SendMessageToSet(packet.Write(), true);
```
Shauren, line 4510: This seems like a correct change to me (after all this packet handler is responsible for clearing in-flight spell missiles) but this belongs in a separate PR

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -52,6 +52,8 @@
 #include "ObjectMgr.h"
 #include "Opcodes.h"
 #include "OutdoorPvP.h"
+#include "MovementPacketSender.h"
+#include "MoveSplineInit.h"
```
Shauren, line 56: Why is this header neccessary here?

Shauren (changes requested):

src/server/worldserver/worldserver.conf.dist
```diff
@@ -4037,6 +4038,23 @@ PacketSpoof.BanDuration = 86400
...
+# Settings to tweak various anti cheat measures.
+#
+#    AntiCheats.PendingMoveChangesTimeoutTime
+#        Description: Time each client has to acknowledge a movement change until they are kicked.
+#        Note:        A lower time will give cheaters less advantage when using a very specific
+#                     type of hack. However, the cost is that innocent but laggy players might
+#                     sometimes get kicked.
+#        Default:     2000 (2 seconds)
+
+AntiCheat.PendingMoveChangesTimeoutTime = 5000
```
jackpoz, line 4053: can we have this disabled by default ?

jackpoz (changes requested):

Faq:
could you check also this? https://github.com/TrinityCore/TrinityCore/issues/20136#issuecomment-687663047

src/server/game/Entities/Player/Player.cpp
chaodhib: true. i removed that part

src/server/game/Entities/Unit/Unit.cpp
chaodhib: it wasn't. it's the remnants of a rebase I assume. Anyway, I removed it.

chaodhib:
@jackpoz https://github.com/TrinityCore/TrinityCore/issues/21044 remains fixed with this PR. The crash is fixed.

@Faq I checked briefly. The remaining issue doesn't seem to be linked to this PR. Looks like a NPC spawning issue.

src/server/worldserver/worldserver.conf.dist
chaodhib: Maybe we could change this right before the next release. But before that, for the sake of testing, I would suggest to keep this enabled by default. A player getting kicked is an instant feedback that some assumption in the code is wrong or that there is some kind of bug. It would be nice to catch these issues early.

jackpoz: I'd like to get the issues caused by recent prs fixed first, then we can experiment again with users and ask feedback about new features.

chaodhib: If i'm not mistaken, all the issues created by https://github.com/TrinityCore/TrinityCore/pull/26348 or https://github.com/TrinityCore/TrinityCore/pull/26538 are either fixed already or they are fixed by this PR. 

However, this PR doesn't just fix issues created recently, it is the first step into making player movement smoother. That's why user feedback on this would be cool.

shenhuyong:
I found a bug, players will be kick when they rides their mounts teleport to a map where mount is not allowed.
`2021-05-29_22:07:58 INFO  [cheat] Unit::CheckPendingMovementAcks: Player GUID: GUID Full: 0x0000000000012509 Type: Player Low: 75017 took too long to acknowledge a movement change. He was therefore kicked.`
How to reproduce it:
- Use command `.go xyz 12890 -7326 66 530`
- Ride a mount
- Go into the instance(Magisters Terrace)
- Then you can see your character moves faster and you will be kick later.

chaodhib:
@shenhuyong thanks! will look into it

src/server/worldserver/worldserver.conf.dist
jackpoz: Would it be ok to leave it disabled by default for some time and then we change it ? we can find other ways to get feedback, like actively asking users to enable the setting, but I would prefer this feature to be opt-in for a while (which would allow to merge this PR earlier as we don't have to test every case).

Maybe like merging the PR, letting 1 month pass with no issues related to it reported, then we can enable it by default ? I can enable it on the test realm too and users can test it there if they want to help.

ckegg:
> @ckegg @shenhuyong testing would be highly appreciated!
Speed Mod issue seems to be resolved after tests.

src/server/worldserver/worldserver.conf.dist
chaodhib: Sure let's do that. I'll make the kick disabled by default, right before merging the PR. I'll keep it enable while the PR is open though, to get as much as feedback as possible, if you don't mind

jackpoz: is that something that can be implemented with "when AntiCheat.PendingMoveChangesTimeoutTime is set to 0 don't kick, > 0 kick" ?

chaodhib: Yes, that's what I was thinking too

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12919,6 +12897,91 @@ void Unit::UpdateHeight(float newZ)
...
+        even with zero latency, the server will receive an ack from the client only after 20s.
+
+        For this reason and because the current implementation is simple (you dear reader, feel free to improve it if you can), we will just ignore checking for
+        near teleport acks (for now. @todo).
+        */
+        if (oldestChangeToAck.movementChangeType == MovementChangeType::TELEPORT)
+            return;
+
+        GameClient* controller = GetGameClientMovingMe();
+        controller->GetWorldSession()->KickPlayer("Took too long to ack a movement change");
```
ccrs, line 12944: I'm guessing the string can be moved to db

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -450,78 +451,107 @@ void WorldSession::HandleForceSpeedChangeAck(WorldPacket &recvData)
...
-        case CMSG_FORCE_SWIM_SPEED_CHANGE_ACK:          move_type = MOVE_SWIM;          force_move_type = MOVE_SWIM;        break;
-        case CMSG_FORCE_SWIM_BACK_SPEED_CHANGE_ACK:     move_type = MOVE_SWIM_BACK;     force_move_type = MOVE_SWIM_BACK;   break;
-        case CMSG_FORCE_TURN_RATE_CHANGE_ACK:           move_type = MOVE_TURN_RATE;     force_move_type = MOVE_TURN_RATE;   break;
-        case CMSG_FORCE_FLIGHT_SPEED_CHANGE_ACK:        move_type = MOVE_FLIGHT;        force_move_type = MOVE_FLIGHT;      break;
-        case CMSG_FORCE_FLIGHT_BACK_SPEED_CHANGE_ACK:   move_type = MOVE_FLIGHT_BACK;   force_move_type = MOVE_FLIGHT_BACK; break;
-        case CMSG_FORCE_PITCH_RATE_CHANGE_ACK:          move_type = MOVE_PITCH_RATE;    force_move_type = MOVE_PITCH_RATE;  break;
+        TC_LOG_INFO("cheat", "WorldSession::HandleForceSpeedChangeAck: Player %s from account id %u kicked because no movement change ack was expected from this player",
+            _player->GetName().c_str(), _player->GetSession()->GetAccountId());
+        client->GetWorldSession()->KickPlayer("No movement change ack was expected from this player");
```
ccrs, line 503: same in this file

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -450,78 +451,107 @@ void WorldSession::HandleForceSpeedChangeAck(WorldPacket &recvData)
...
+        TC_LOG_WARN("misc", "The computed movement time using clockDelta is erronous. Using fallback instead");
+        movementInfo.time = GameTime::GetGameTimeMS();
+    }
+    else
+    {
+        movementInfo.time = (uint32)movementTime;
     }
+
+    mover->m_movementInfo = movementInfo;
+    mover->UpdatePosition(movementInfo.pos);
```
ccrs, line 550: just a doubt. In case of splines sent by movementgenerator or manually, there will be two UpdatePositions right? one done by the updatespline method and one on movement ack here. right?

src/server/game/Movement/MovementPacketSender.cpp
```diff
@@ -0,0 +1,106 @@
...
+    controller->SendDirectMessage(&data);
+}
+
+void MovementPacketSender::SendSpeedChangeToObservers(Unit* unit, UnitMoveType mtype, float newSpeedFlat)
+{
+    GameClient* controller = unit->GetGameClientMovingMe();
+
+    ASSERT(controller);
+
+    WorldPacket data;
```
ccrs, line 87: maybe this is a nice chance to move it to the new packet format, if you feel like it

src/server/game/Entities/Unit/Unit.cpp
jackpoz: that's just a log, we rarely move to db logs

src/server/game/Handlers/MovementHandler.cpp
funjoker: check jackpoz's response above :)

src/server/game/Movement/MovementPacketSender.cpp
funjoker: yeah please do so :)

src/server/game/Entities/Unit/Unit.cpp
illfated: The alternative would be to look for an identical string in the OS or external libs, but I guess the chance of finding a close to identical text is pretty slim in generic files.

src/server/game/Movement/MovementPacketSender.cpp
Faq: Yeah, but that's another PR.

ccrs: "if you feel like it"

Ovahlord: Nah, since we're using array based packet building here, having a structure build helper here is cleaner that squeezing it all into one class imo. Also out of scope of this PR

src/server/game/Entities/Unit/UnitDefines.h
```diff
@@ -269,25 +269,40 @@ enum MovementFlags : uint32
...
+    MOVEMENTFLAG2_INTERPOLATED_TURNING                  = 0x00000800,
+    MOVEMENTFLAG2_INTERPOLATED_PITCHING                 = 0x00001000,
+    MOVEMENTFLAG2_UNK14                                 = 0x00002000,
+    MOVEMENTFLAG2_CAN_TRANSITION_BETWEEN_SWIM_AND_FLY   = 0x00004000,
+    MOVEMENTFLAG2_UNK16                                 = 0x00008000
 };
 
+enum UnitMoveType
+{
+    MOVE_WALK = 0,
```
Ovahlord, line 293: tabs pls

src/server/game/Entities/Unit/UnitDefines.h
```diff
@@ -269,25 +269,40 @@ enum MovementFlags : uint32
...
+    MOVE_RUN_BACK = 2,
+    MOVE_SWIM = 3,
+    MOVE_SWIM_BACK = 4,
+    MOVE_TURN_RATE = 5,
+    MOVE_FLIGHT = 6,
+    MOVE_FLIGHT_BACK = 7,
+    MOVE_PITCH_RATE = 8
+};
+
+#define MAX_MOVE_TYPE     9
```
Ovahlord, line 304: let's move away from defines and go with `static constexpr uint8 MAX_MOVE_TYPE = 9;` instead

src/server/game/Entities/Unit/UnitDefines.h
illfated: TC codestyle recommends against Tabs.

Ovahlord: yes and that's no problem when one properly sets up his visual studio to use  tabs4spaces which is what I mean. 
And hi trokli, the stalker of everyone, the one that notices everything and everyone.

Shauren: He didnt mean tabs literally but aligning the = to the same column

illfated: > He didnt mean tabs literally but aligning the = to the same column

Thanks, that interpretation went over my head. I approve.

src/server/game/Entities/Unit/UnitDefines.h
jackpoz: the old code had `#define MAX_MOVE_TYPE     9` , let's try to keep as few changes as possible in this PR (and in every PR) so we can focus on the actual issues that it fixes

jackpoz:
it would be nice to update every single line of TC to modern codestyle but let's do that in another PR. This PR is supposed to fix issues that were added in a recent PR, nothing more.

src/server/game/Handlers/MovementHandler.cpp
chaodhib: Sorry, i'm not sure I follow you. if you could give more context to your question, that would be appreciated

src/server/game/Entities/Unit/UnitDefines.h
chaodhib: done

src/server/game/Movement/MovementPacketSender.cpp
chaodhib: so won't do here (but everyone is welcome to improve this in a later PR)

src/server/game/Handlers/MovementHandler.cpp
chaodhib: so won't do here (but everyone is welcome to improve this in a later PR)

src/server/game/Entities/Unit/Unit.cpp
chaodhib: so, won't do

src/server/game/Entities/Unit/UnitDefines.h
chaodhib: so won't do here (but everyone is welcome to improve this in a future PR)

chaodhib:
[This bug](https://github.com/TrinityCore/TrinityCore/pull/26561#issuecomment-850841398) should be fixed now and there is no pending suggestion. If there is nothing else, let's merge this weekend. 

Before I merge, I will change `AntiCheat.PendingMoveChangesTimeoutTime` default value to 0, like said above.

src/server/game/Handlers/MovementHandler.cpp
Ovahlord: While a unit is being moved by a spline (player charge or knockbacks) the spline system does its own position relocation in the Unit::Update loop. So what he probably asks for is disabling movementInfo based position relocations while the spline is doing its job and use the spline's position instead.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12919,6 +12897,101 @@ void Unit::UpdateHeight(float newZ)
...
+    return std::find_if(m_pendingMovementChanges.begin(), m_pendingMovementChanges.end(),
+        [changeType](PlayerMovementPendingChange const& pendingChange)
+    {
+        return pendingChange.movementChangeType == changeType;
+    }) != m_pendingMovementChanges.end();
+}
+
+void Unit::CheckPendingMovementAcks()
+{
+    if (sWorld->getIntConfig(CONFIG_PENDING_MOVE_CHANGES_TIMEOUT) <= 0)
```
Faq, line 12928: Why "<"? Timeout can be negative in config?

src/server/game/Handlers/MovementHandler.cpp
ccrs: not trying to imply any solution myself though, I was just stating the scenario. Don't know what's more appropiate, two updatepositions, the one on updatespline or the one on ack.

Ovahlord: IMO we should give the client as little control over such important value as possible so while we do have the chance of using spline data, we should do so let alone to avoid packet spoofing possibilities.

jackpoz:
@chaodhib just as a reminder of how config files work, you'll have to set the value to 0 in worldserver.conf and in the .cpp file where the config is read

chaodhib:
@jackpoz sure, will do. thanks

src/server/game/Entities/Unit/Unit.cpp
chaodhib: to be honest, i don't know. will check

src/server/game/Handlers/MovementHandler.cpp
chaodhib: These 2 lines are in no way different from the lines https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Handlers/MovementHandler.cpp#L396 & https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Handlers/MovementHandler.cpp#L415 in the `HandleMovementOpcodes` handler. The client is allowed to submit positional changes for the units it controls.  The ACK messages carry completely valid positions. The unit did went through these locations (at some point in the past). Disregarding them can actually lead to jerky movements when observing other players.

> the spline system does its own position relocation in the Unit::Update loop

For the majority of the time, spline will only run on units controlled by the server. And here, i'm talking about "real-time client control" (see [this PR](https://github.com/TrinityCore/TrinityCore/pull/26348)). The client is not allowed to send movement messages about them (if it does, they are ignored). So most of the time, there won't be a conflict. Player charges are indeed an exception. But even in retail, charges can lead to jerky movements. It is how the movement system works in WoW. Clients are authoritative on movement for the units under their controls.

PS: Knockback doesn't require a spline (the server can send SMSG_MOVE_KNOCK_BACK to the client, then the physical simulation of the knockback is run on the client).

shenhuyong:
@chaodhib  [This bug](https://github.com/TrinityCore/TrinityCore/pull/26561#issuecomment-850841398) has not been completely fixed.
Another way to reproduce it:
- Use command `.go xyz 3478 266 -120`
- Create a group and convert it to be a raid group.
- Mount a flying mount
- Fly into the instance (Obsidian Sanctum)
- Then you can see your character moves faster and you will be kicked later.

chaodhib:
@shenhuyong Thanks. I'll look into it Monday (I'm not free until then)

jackpoz:
@chaodhib no idea if this is related to the recent changes or not, but if I enter vehicle https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=34775 and exit it without moving further, other players will see me on top of the vehicle:
image

chaodhib:
Sorry for the delay.

@shenhuyong I can't reproduce the bug. For me, it works fine, even with a flying mount, weird. Are you sure you checked out the latest commit of the branch?

@jackpoz I checked and what you are observing is also on the 3.3.5 branch. So while it is clearly a movement bug, it is not related to this PR.

Ovahlord:
Ye, the vehicle exit thing is unrelated to this. It's some bullshit going on in the passenger removal which I think is the spline packet that does this.

Aokromes:
what's missing here to be merged? i got reports of people being slowed after vomit on Festergut

chaodhib:
Nothing as far as I can tell. I'll merge tomorrow evening, if no objection until then.

jackpoz:
could you please apply the "disable the setting by default" with a bit of warning so we can review the final set of changes before merging it 😃 ?

chaodhib:
@jackpoz Done, let me know when I can merge :)

jackpoz:
The default settings look good to me

chaodhib:
Cool, I'm merging. Thank you for the review guys!

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -450,78 +451,105 @@ void WorldSession::HandleForceSpeedChangeAck(WorldPacket &recvData)
...
+        case MovementChangeType::SPEED_CHANGE_SWIM_BACK:            moveTypeSent = MOVE_SWIM_BACK; break;
+        case MovementChangeType::RATE_CHANGE_TURN:                  moveTypeSent = MOVE_TURN_RATE; break;
+        case MovementChangeType::SPEED_CHANGE_FLIGHT_SPEED:         moveTypeSent = MOVE_FLIGHT; break;
+        case MovementChangeType::SPEED_CHANGE_FLIGHT_BACK_SPEED:    moveTypeSent = MOVE_FLIGHT_BACK; break;
+        case MovementChangeType::RATE_CHANGE_PITCH:                 moveTypeSent = MOVE_PITCH_RATE; break;
         default:
-            TC_LOG_ERROR("network", "WorldSession::HandleForceSpeedChangeAck: Unknown move type opcode: %u", recvData.GetOpcode());
+            TC_LOG_INFO("cheat", "WorldSession::HandleForceSpeedChangeAck: Player %s from account id %u kicked for incorrect data returned in an ack",
+                _player->GetName().c_str(), _player->GetSession()->GetAccountId());
+            _player->GetSession()->KickPlayer("incorrect data returned in an ack");
```
jackpoz, line 523: Please wrap this kick around the config check too so users can disable it if they don't want it

Aokromes: 2021-06-16_21:36:17 INFO  [cheat] WorldSession::HandleForceSpeedChangeAck: Player from account id kicked for incorrect data returned in an ack
2021-06-16_21:36:17 INFO  [network.kick] Account: 97003 Character: '' GUID Full: Type: Player Low: kicked with reason: incorrect data returned in an ack

it looks people is being kicked while enter on raid when dead.

PR #26564, opened 2021-05-30
Added "thread" library in file RASession.cpp to fix the build issue in GCC glibc

KoNickss:
**Changes proposed:**

-  Fixed bug when using GCC for compilation, GCC needs the `thread` library included for use of the sleep_for, so I added that it so it can build, otherwise the build crashes at 99%, other compilers don't seem to have this issue but since this seems to be intended behaivour with GCC, we need to fix this ourselves

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #12597

Before this the issue was closed improperly, as the issue was still present on both macOS and Linux.

**Tests performed:**
It compiles succesfully, as opposed to the error shown previously, I tried it also on a clean VM and seems like this fixes the issue entirely
None

### DIFF
--- a/src/server/worldserver/RemoteAccess/RASession.cpp
+++ b/src/server/worldserver/RemoteAccess/RASession.cpp
@@ -27,6 +27,7 @@
 #include <boost/asio/buffer.hpp>
 #include <boost/asio/read_until.hpp>
 #include <memory>
+#include <thread>
 
 using boost::asio::ip::tcp;
 


Shauren:
I would like to see the compiler log as it builds just fine with GCC

KoNickss:
`error: ‘sleep_for’ is not a member of ‘std::this_thread’

then the build fails

illfated:
Maybe it seems unnecessary to you, but it would be useful to let the PR title give a short description of what this PR is about. "3.3.5" is too generic in this repository.

Suggestion: **Build/GCC: include \<thread\> in RASession.cpp**

You can of course use something else, this is just what I could think of at short notice. ( I used the backslash character to escape the brackets in `\<thread\>`. )

Shauren:
> 
> 
> `error: ‘sleep_for’ is not a member of ‘std::this_thread’
> 
> then the build fails

More details please. OS (and version, distro name), GCC version

KoNickss:
I'm trying to compile latest version of branch 3.5.5

GCC version 11.1.0

Archlinux

If you want I can give you the complete compilation output after I get home

Shauren:
I looked through sources of libstdc++ versions 10 and 11, it looks like some code was reorganized and this PR is indeed neccessary for version 11

KoNickss:
> I looked through sources of libstdc++ versions 10 and 11, it looks like some code was reorganized and this PR is indeed neccessary for version 11
> 
> The culprit was `<future>` header, included in RASession.h, in version 10 it included , in version 11 it no longer does

Indeed, thanks for merging, I'll also open a request on master

PR #26565, opened 2021-05-30
Fixed libstdc++ version 11 future lib removal of thread

KoNickss:
**Changes proposed:**

-  Fixed issue when compiling on distros with libstdc++ 11 and up, where \<future\> no longer contains thread, so you get a build error at 99%, I also made a similar pull request for 3.5.5 and it got merged

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes new issue with libstdc++ 11

**Tests performed:**
It builds, it runs, it's fine

### DIFF
--- a/src/server/worldserver/RemoteAccess/RASession.cpp
+++ b/src/server/worldserver/RemoteAccess/RASession.cpp
@@ -26,6 +26,7 @@
 #include <boost/asio/buffer.hpp>
 #include <boost/asio/read_until.hpp>
 #include <memory>
+#include <thread>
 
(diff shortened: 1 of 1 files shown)


KoNickss:
I'll add some more notes here: on libstdc++ version 11, the library "future" no longer contains \<thread\>, which is neccesary for the `sleep_for` func, since the header file uses the "future" library it breaks on systems with stdlibc++ 11 and up, this can be easily fixed by manually importing the library, it works on my host as well as a clean slate VM I made for this

PR #26571, opened 2021-06-02
Scripts/Zone: Update IoC scripts to new register model

offl:
**Changes proposed:**

-  Update IoC Lands scripts to new register model

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

none

**Tests performed:**

Builds, not tested

none

### DIFF
--- a/src/server/scripts/Northrend/IsleOfConquest/isle_of_conquest.cpp
+++ b/src/server/scripts/Northrend/IsleOfConquest/isle_of_conquest.cpp
@@ -31,50 +31,39 @@
 // TO-DO: This should be done with SmartAI, but yet it does not correctly support vehicles's AIs.
 //        Even adding ReactState Passive we still have issues using SmartAI.
 
-class npc_four_car_garage : public CreatureScript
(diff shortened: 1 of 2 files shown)


src/server/scripts/Northrend/IsleOfConquest/isle_of_conquest.cpp
```diff
@@ -31,50 +31,39 @@
 // TO-DO: This should be done with SmartAI, but yet it does not correctly support vehicles's AIs.
```
illfated, line 31: Cosmetic/grammar: The location of "yet" affects its effective meaning in a sentence.
```suggestion
// TO-DO: This should be done with SmartAI, but it does not yet correctly support vehicles' AIs.
```
Also: Remove redundant possessive 'S' which is represented by the apostrophe after plural nouns ending with 'S'.

offl: Not for this PR, even such small changes

PR #26578, opened 2021-06-05
Scripts/Quest: Rework 'Taken by the Scourge' (11611)

offl:
**Changes proposed:**

Rework 'Taken by the Scourge' (11611) by implementing spell scripts and SAI. This one is different from the ones I previously updated but logic is the same: if you're not on quest, by killing cocoon you can get only guardians but only 3. If you're on quest, you can get guardians or peons which gives quest credit. But again you can't get more than 3 guardians(it's controlled by a spell player casts every time guardian is summoned) and once you get 3 guardians, all other freed creatures will be only peons

After guardian dies or executes despawn event, it casts a dispel spell to remove one stack debuff which controlls can player summon guardian or not(less than 3 stacks = can summon guardian)

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

none

**Tests performed:**

Builds, tested in-game

Leaved a bunch of comments about guardians, don't have enough info to make them work better


offl:
I'll merge in next few hours

PR #26586, opened 2021-06-08
Scripts/Commands: Minor improvements in .pinfo and .unban character commands

Jildor:
**Changes proposed:**

-  Now .pinfo command show correct bantime in case of not being permanent (currently show always "permanent")
-  Now .pinfo command show bantype Account in case of being bantype account instead of showing "unknown" (not sure if I defined in the proper number LANG_ACCOUNT)
-  Now .unban character command show success text with unban player name

- [X] 3.3.5
- [ ] master

**Tests performed:**

tested in-game


jackpoz:
Looks good

jackpoz:
Thanks for the PR :)

PR #26616, opened 2021-06-22
Clear unused #undef leftover in a refactor

kelno:
Previously this function was using a macro that was #undef at the end of the function. Since then it uses a lambda function instead and the undef is not needed anymore.

(Please do tell if that kind of tiny PR is not welcome)


jackpoz:
> (Please do tell if that kind of tiny PR is not welcome)

small PRs are always welcome 👍

jackpoz:
Thanks for the PR :)

PR #26627, opened 2021-06-26
Build: Remove fine-grained CXX feature checks

Carbenium:
**Changes proposed:**
With C++17 enabled these don't serve any purpose.
See https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html

- [x] 3.3.5
- [ ] master

**Issues addressed:**

**Tests performed:**
build

### DIFF
--- a/cmake/macros/ConfigureBaseTargets.cmake
+++ b/cmake/macros/ConfigureBaseTargets.cmake
@@ -19,20 +19,7 @@ add_library(trinity-feature-interface INTERFACE)
 
 target_compile_features(trinity-feature-interface
   INTERFACE
-    cxx_std_17
-    cxx_alias_templates
-    cxx_auto_type
-    cxx_constexpr
-    cxx_decltype
-    cxx_decltype_auto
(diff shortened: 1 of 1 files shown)


jackpoz:
Did you try with an old enough compiler? As some might implement c++17 partially and not implement all the features we want

Carbenium:
There are no fine-grained feature checks for C++17 anymore. All listed features are C++14 and below (see the linked doc in the description).

jackpoz:
I was thinking about https://github.com/TrinityCore/TrinityCore/commit/cd30e0b86ce6ee88386a91cebdf353fc55805c57#diff-ad412c7191ac12642c1f67b5a159dc30bf4a1727833017a08b2d9d0e8a6fda5c as we didn't find any CMake guard to test that out of the box

Carbenium:
Ah, yes. I remember that one.
Grepping through the CMake's source there is nothing related. I suppose our home-made check is the best we can do.

PR #26630, opened 2021-06-26
Scripts/Commands: Convert title commands to new system

Carbenium:
**Changes proposed:**

-  Scripts/Commands: Convert title commands to new system

- [x] 3.3.5
- [ ] master

**Tests performed:**
build + in-game


jackpoz:
Last year some commands had their behaviour modified during the conversion, even if that was discovered months later. Just as a heads up, please try to keep the behaviour the same (i.e. if the command checks for additional parameters or options)

Carbenium:
https://github.com/TrinityCore/TrinityCore/pull/26631 is (unfortunately) going to be such an candidate. It's not really possible to represent old behavior with the new system. We will have to decide if that change is acceptable or we have to retain the manual parsing for those commands.

PR #26631, opened 2021-06-26
Scripts/Commands: Convert rbac commands to new system

Carbenium:
**Changes proposed:**

-  Scripts/Commands: Convert rbac commands to new system

- [x] 3.3.5
- [ ] master

**Tests performed:**

build + in-game

**Note**: The invocation semantics of the `rbac account *` commands slightly changed. When > 1 arguments are passed the current target isn't considered in account resolution. Instead only the first argument is used as account identifier.

**Note 2**: To enable rbac commands for administrators permissions 202-206 have to be linked to permission 196.

Sample SQL for auth DB:
```sql
INSERT INTO `rbac_linked_permissions` VALUES (196, 202), (196, 203), (196, 204), (196, 205), (196, 206);
```


src/server/scripts/Commands/cs_rbac.cpp
```diff
@@ -32,349 +32,256 @@ EndScriptData */
...
-
-                if (param1)
-                    id = atoi(param1);
-
-                useSelectedPlayer = true;
-            }
-            else
-            {
-                id = atoi(param2);
-                realmId = atoi(param3);
```
Shauren, line 115: What is/was `param1` when this branch is taken?

Carbenium: The account name e.g `rbac acc rev acc1 2 3` is account name "acc1", permission ID 2 and realm 3

jackpoz:
> The invocation semantics of the `rbac account *` commands slightly changed. When > 1 arguments are passed the current target isn't considered in account resolution. Instead only the first argument is used as account identifier.

could you write a couple of examples of how it worked before and how it will work after to get an idea of what this change means ?

Carbenium:
So, I tried to summarize the old and new behavior in the table below.

The **Note 1** from above is related to case 9/10. The old behavior changed how parameters where interpreted depending whether a target was selected or not. Now we don't do this anymore and use the first parameter always as account identifier if more than 1 parameter is passed.

 Case| Target | Param 1 | Param 2 | Param 3 | Old | New
 -|---------|-------------|-------------|-------------|------|-------
1| \- | - | - | - | "wrong parameter" | syntax error
2| self/other | - | - | - | "wrong parameter" | syntax error
3| \- | `perm` | - | - | self granted `perm` | syntax error
4| self/other | `perm` | - | - | self/other granted `perm` on realm `-1` | self/other granted `perm` on realm `-1`
5| \- | `account` | - | - | "wrong parameter" | syntax error
6| self/other | `account` | - | - | "wrong parameter" | syntax error if `account` can't be parsed as `perm`, otherwise `account` parsed as perm id is granted to self/other on realm `-1`
7| \- | `account` | `perm` | - | "wrong parameter" | `account` granted `perm` on realm `-1`
8| self/other | `account` | `perm` | - | "wrong parameter" | `account` granted `perm` on realm `-1`
9| \- | `perm` | `realmid` | - | self granted `perm` on realm `realmid`  | syntax error
10| self/other | `perm` | `realmid` | - | self/other granted `perm` on realm `realmid` | parsed as **case 8**.
11| \- | `account` | `perm` | `realmid` | `account` granted `perm` on realm `realmid` | `account` granted `perm` on realm `realmid`
12| self/other | `account` | `perm` | `realmid` | `account` granted `perm` on realm `realmid` |  `account` granted `perm` on realm `realmid`

jackpoz:
Thanks for the documentation, feel free to merge it whenever you want. If someone reports a strange behavior, we can link them the table above.

PR #26639, opened 2021-06-27
CI/GitHub: Use issue form and add contact links

jackpoz:
**Changes proposed:**

-  CI/GitHub: Use issue form and add contact links
-  See https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/

This is how https://github.com/TrinityCore/TrinityCore/issues/new will look:
image

This is how "Generic issue" choice will look (mind the small screenshot):
image

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Forces users to fill all the required fields

**Tests performed:**

- Tested the template in another repo
- Tested the github actions that set the labels
- Tested the contact links

- In the future we will modify the links and the templates to fit our needs

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue.yml
+++ b/.github/ISSUE_TEMPLATE/issue.yml
@@ -49,7 +49,7 @@ body:
         Copy the result of server debug command (if you need to run it from client get prat addon)
     validations:
       required: true
-  - type: textarea
+  - type: input
     id: os
     attributes:
       label: Operating system
--- a/.github/ISSUE_TEMPLATE/issue.yml
+++ b/.github/ISSUE_TEMPLATE/issue.yml
@@ -1,0 +1,68 @@
+name: Generic issue
+description: Report an issue, a suggestion, SQL only fix.
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this form! Please make sure to fill all fields, including the Title above.
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description: |
+        Description of the problem or issue here.
+        If this is a crash, post the crashlog (upload to https://gist.github.com/).
+    validations:
+      required: true
+  - type: textarea
+    id: expected
+    attributes:
+      label: Expected behaviour
+      description: |
+        Tell us what should happen instead.
+    validations:
+      required: true
+  - type: textarea
+    id: reproduce
+    attributes:
+      label: Steps to reproduce the problem
+      description: |
+        1. Step 1 include entries of affected creatures / items / quests with a link to the relevant wowhead or https://tcubuntu.northeurope.cloudapp.azure.com/aowow/ page.  
+        2. Step 2
+        3. Step 3
+    validations:
+      required: true
+  - type: dropdown
+    id: branch
+    attributes:
+      label: Branch
+      options:
+        - 3.3.5
+        - master
+    validations:
+      required: true
+  - type: textarea
+    id: commit
+    attributes:
+      label: TC rev. hash/commit
+      description: |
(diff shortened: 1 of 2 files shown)


jackpoz:
It can be tested at https://github.com/jackpoz/TrinityCore/issues/new/choose until it gets merged

Faq:
If merged it will work also for 335. like demo shows it?

Shauren:
@Faq yes, this is set on per-repository level, not per-branch

.github/ISSUE_TEMPLATE/issue.yml
```diff
@@ -0,0 +1,68 @@
...
+      required: true
+  - type: textarea
+    id: commit
+    attributes:
+      label: TC rev. hash/commit
+      description: |
+        Copy the result of server debug command (if you need to run it from client get prat addon)
+    validations:
+      required: true
+  - type: textarea
```
Carbenium, line 52: Would a single-line `input` field be enough? Just to reduce the vertical extent a bit.

jackpoz: That could be an idea.

The issue template actually suggests to "Copy the result of server debug command (if you need to run it from client get prat addon)", this is the result of that command:
>TrinityCore rev. c8ca48823d45 2021-06-26 14:21:18 +0200 (3.3.5 branch) (Unix, RelWithDebInfo, Static)
Using SSL version: OpenSSL 1.1.1j  16 Feb 2021 (library: OpenSSL 1.1.1j  16 Feb 2021)
Using Boost version: 1.74.0
Using MySQL version: 80025
Using CMake version: 3.18.4
Compiled on: Linux 5.11.0-1006-azure
MMAPs status: Enabled
maps directory located in /trinity/wowdata/maps. Total size: 291014951 bytes
vmaps directory located in /trinity/wowdata/vmaps. Total size: 657797442 bytes

If we want to allow pasting this, then we need a multi-row field.

Carbenium: I suppose there was a misunderstanding. My comment was meant for the `os` field.
I agree that we need multiline for the `server debug` output.

.github/ISSUE_TEMPLATE/issue.yml
```diff
@@ -0,0 +1,68 @@
...
+        - 3.3.5
+        - master
+    validations:
+      required: true
+  - type: textarea
+    id: commit
+    attributes:
+      label: TC rev. hash/commit
+      description: |
+        Copy the result of server debug command (if you need to run it from client get prat addon)
```
Faq, line 49:
```suggestion
        Copy the output of server console (when used ".server debug" command) or from game client (by using "Prat" addon)
```

.github/ISSUE_TEMPLATE/issue.yml
jackpoz: @Carbenium I picked the field above instead of the field below. OS single line sounds good

jackpoz, line 52:
```suggestion
  - type: input
```

.github/ISSUE_TEMPLATE/issue.yml
jackpoz: should `when used ".server debug" command` be `when using ".server debug" command` ?

Faq: using sounds better :)

offl: Suggested version is a bit confusing

jackpoz: As the PR has been merged, feel free to open a new PR changing the text. Both versions sounds a bit confusing to me, maybe we will come up with something that is even clearer :)

Issue #26642, opened 2021-06-28
Crash log on HandleForceSpeedChangeAck


ckegg:
**Description:**
Not sure if it is related to https://github.com/TrinityCore/TrinityCore/pull/26561
worldserver.conf : AntiCheat.PendingMoveChangesTimeoutTime = 4000

[crash-2021-06-27-21-09-00.log](https://github.com/TrinityCore/TrinityCore/files/6722560/crash-2021-06-27-21-09-00.log)
[crash-2021-06-27-23-11-33.log](https://github.com/TrinityCore/TrinityCore/files/6722561/crash-2021-06-27-23-11-33.log)
[crash-2021-06-27-23-58-04.log](https://github.com/TrinityCore/TrinityCore/files/6722562/crash-2021-06-27-23-58-04.log)
[crash-2021-06-28-05-00-45.log](https://github.com/TrinityCore/TrinityCore/files/6722563/crash-2021-06-28-05-00-45.log)

**Expected behaviour:**
No crash

**Steps to reproduce the problem:**
Tried to reproduce but was unable to. Tried levitate, slow fall or whatever but dunno how crashes happened. Maybe lag?

**Branch(es):**
3.3.5

**TC rev. hash/commit:** 
c8ca48823d45dadb042a80637eff954483dd5328

**Operating system:** 
debian 11

jackpoz:
@chaodhib could you please check this ?

jackpoz:
```c++
REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_FALLING_SLOW) && !mover->HasAuraType(SPELL_AURA_FEATHER_FALL),
        MOVEMENTFLAG_FALLING_SLOW);
```

`mover` is null at that point

chaodhib:
@jackpoz will do

ckegg:
Crash log with crashreport.gdb script 1.4
3.3.5 b596ea298a7b
[crash-2021-07-01-10-58-17.log](https://github.com/TrinityCore/TrinityCore/files/6745134/crash-2021-07-01-10-58-17.log)

Aokromes:
maybe this is druids or shamans with knock back spells?

ckegg:
>
> [crash-2021-07-01-10-58-17.log](https://github.com/TrinityCore/TrinityCore/files/6745134/crash-2021-07-01-10-58-17.log)

I dig deeper on all guids and they all were hunter characters.
[...]

PR #26644, opened 2021-06-28
Core/Movement: Fix movement handler crash during ReadMovementInfo check

chaodhib:
**Changes proposed:**

-  Fix a rare edge case that can lead to a null pointer crash

- [X] 3.3.5
- [ ] master

**Issues addressed:**

Should close https://github.com/TrinityCore/TrinityCore/issues/26642

**Tests performed:**

Tested IG.

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -928,6 +928,7 @@ void WorldSession::ReadMovementInfo(WorldPacket &data, MovementInfo* mi)
     #endif
 
     Unit* mover = GetPlayer()->GetGUID() == mi->guid ? GetPlayer() : ObjectAccessor::GetUnit(*GetPlayer(), mi->guid);
+    ASSERT(mover, "if the server allows this unit to be moved by the client, the unit should still exist!");
 
     if (!GetPlayer()->GetVehicleBase() || !(GetPlayer()->GetVehicle()->GetVehicleInfo()->Flags & VEHICLE_FLAG_FIXED_POSITION))
         REMOVE_VIOLATING_FLAGS(mi->HasMovementFlag(MOVEMENTFLAG_ROOT), MOVEMENTFLAG_ROOT);
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -38,6 +38,7 @@
 #include "Map.h"
 #include "Metric.h"
 #include "MoveSpline.h"
+#include "ObjectAccessor.h"
 #include "ObjectMgr.h"
 #include "Opcodes.h"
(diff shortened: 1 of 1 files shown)


src/server/game/Server/WorldSession.cpp
```diff
@@ -926,7 +927,7 @@ void WorldSession::ReadMovementInfo(WorldPacket &data, MovementInfo* mi)
             mi->RemoveMovementFlag((maskToRemove));
     #endif
 
-    Unit* mover = GetGameClient()->GetActivelyMovedUnit();
+    Unit* mover = GetPlayer()->GetGUID() == mi->guid ? GetPlayer() : ObjectAccessor::GetUnit(*GetPlayer(), mi->guid);
```
jackpoz, line 930: could you please add an if check on mover after this line and handle that case too ? whether it's return + log error or ASSERT (even if a return + log is always better if possible so it doesn't crash the whole server)

Shauren: mi->guid is client controlled, it must not be ASSERT

chaodhib: @jackpoz will do tonight.

@Shauren 

> how does this change affect cheating prevention? 

WorldSession::ReadMovementInfo is called in a movement handler only after all the permission checks are made

Shauren: Wouldnt this mean that there is a data consistency error in GameClient state? `_allowedMovers` has a guid but `_activelyMovedUnit` is not set?

chaodhib: @jackpoz done. i choose the assert approach and not just returning because otherwise weird things will happen after the function will return.

@Shauren It might. I'll look into it. It could also be explained by the following edge case:
on login, there is a window of time (after the server initialize the player and sends `CreateObject2` to the client and before the clients sends `CMSG_SET_ACTIVE_MOVER`) during which the GameClient's `_allowedMovers` contains the player but `_activelyMovedUnit` is null. If the client sends an ACK message during that time window, we get the situation you described

chaodhib: In the meantime, let's merge this already?

jackpoz:
@chaodhib does https://github.com/TrinityCore/TrinityCore/issues/26642#issuecomment-872005053 help ?

src/server/game/Server/WorldSession.cpp
jackpoz: Did you check the crashlogs mentioned below ?

jackpoz:
it seems @offl and @Aokromes have a lot of faith in this PR, let's merge it then and see what happens next

Issue #26650, opened 2021-07-03
DB/Locales: Add missing ruRU texts (Part 4)


Freeman256:
### Description

Adding missing ruRU texts.
Hopefully I've filled out this new error reporting form correctly. Sorry and correct me if this is not the case.

### Expected behaviour

Use the code below to add missing ruRU texts.

```
-- Untranslated name of the quest (it may have already been inserted from that large database that I transferred ** offl ** recently). / Непереведённое название квеста (возможно, оно уже вставилось из той большой базы, которую я передал **offl** недавно).
UPDATE `quest_template_locale` SET `ObjectiveText1`='Убейте Морбента Скверна' WHERE  `ID`=55 AND `locale`='ruRU';

-- Correction of an incorrect phrase. / Исправление неверной фразы.
UPDATE `quest_request_items_locale` SET `CompletionText`='Ну что, $N, тебе удалось обнаружить лагерь охотников?' WHERE  `ID`=201 AND `locale`='ruRU';

-- Not sure if there is a need to insert these missing strings, but maybe localized strings will reference them. / Не уверен, что есть необходимость вставлять эти отсутствующие строки, но, возможно, на них будут ссылаться локализованные строки.
DELETE FROM `quest_greeting` WHERE `ID` IN (2277,7802)
INSERT INTO `quest_greeting` (`ID`, `Greeting`) VALUES 
(2277, 'Greetings, $N.'),
(7802, 'Greetings, $N.');

-- Russian text is missing. The quest gives a game object, so I made the type as in the English version. / Отсутствует русский текст. Квест даёт игровой объект, поэтому сделал тип как у английской версии.
DELETE FROM `quest_greeting_locale` WHERE `ID` IN (186420) AND `locale` = 'ruRU';
INSERT INTO `quest_greeting_locale` (`ID`, `Type`, `locale`, `Greeting`) VALUES 
(186420, '1', 'ruRU', 'Эта коллекция свитков содержит различную логистическую и стратегическую информацию, а также закодированную корреспонденцию.');

[...]

PR #26651, opened 2021-07-03
CI/GitHub: Add issue form for SQL fixes

jackpoz:
**Changes proposed:**

-  Add issue form for SQL fixes

- [ ] 3.3.5
- [x] master

**Issues addressed:**

None

**Tests performed:**

Tested at https://github.com/jackpoz/TrinityCore/issues/new/choose

None


Faq:
My suggestion is add an examples for some points (link to correct sql commit, or just add in the issue text), so there anyone talks in same language and there are no misunderstandings.
Now it is like don't do that, or that, but after would be nice to see correct line how things should be done.

offl:
That is something I'm thinking about but not sure if
> link to correct sql commit, or just add in the issue text

is the proper way to do it

jackpoz:
@Faq right now I just copied the old template and pasted it as an issue form. It sounds good to have a reference example of correctly filled form, either with text or even just with a screenshot

Faq:
Yeah the main idea is so there is not needed much explaining like for issue https://github.com/TrinityCore/TrinityCore/issues/26650#issuecomment-873396215 so any other reporter knows how it should be done.
True, it will look a bit like guide.

jackpoz:
we could include a link to https://github.com/TrinityCore/TrinityCore/wiki with a page that we can update separately, with every information and step we want to give users

jackpoz:
@Faq how about this ?
image

It points to https://github.com/TrinityCore/TrinityCore/wiki/SQL-Fix that can be modified without the need of PRs or commits or so

PR #26655, opened 2021-07-06
Core/Movement: allow waypoints to use 0 as valid facing value

### LINKED ISSUE #26037: Core/Waypoints: orientation field 0 is wrongly treated as no orientation
**Description:**

`orientation` field 0 is wrongly treated as no orientation while 0 is valid orientation and is used by Blizz
Of course we can use 6.2831853072 but it's not the same

**Expected behaviour:**

treat 0 as valid orientation

**Branch(es):** both
**TC rev. hash/commit:** 3cb70ee62a3f30ae492a6985b5306052d90c89c4

Ovahlord:
**Changes proposed:**

-  make waypoint orientation values nullable to support angle values of 0 as well as there actually are some cases that use such values

Closes #26037

**Target branch(es):**
- 3.3.5

**Tests performed:**
- tested on 4.x, no issues discovered

### DIFF
--- a/sql/updates/world/3.3.5/2021_07_06_xx_world.sql
+++ b/sql/updates/world/3.3.5/2021_07_06_xx_world.sql
@@ -1,0 +1,2 @@
+ALTER TABLE `waypoint_data` CHANGE `orientation` `orientation` FLOAT DEFAULT NULL NULL;
+UPDATE `waypoint_data` SET `orientation`= NULL WHERE `orientation`= 0;
--- a/sql/updates/world/3.3.5/2022_04_16_00_world.sql
+++ b/sql/updates/world/3.3.5/2022_04_16_00_world.sql
@@ -1,2 +1,5 @@
 ALTER TABLE `waypoint_data` CHANGE `orientation` `orientation` FLOAT DEFAULT NULL NULL;
 UPDATE `waypoint_data` SET `orientation`= NULL WHERE `orientation`= 0;
+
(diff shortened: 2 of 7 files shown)


Macsimka:
Also need made changes here, to prevent use 0.0f for all scripted waypoints.
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Scripting/ScriptSystem.cpp#L83

ccrs (approved):

jackpoz:
> Also need made changes here, to prevent SetFacing to 0.0f for all scripted waypoints.

are there other places that need to be checked or is this PR ready to be merged ?

Ovahlord:
apparently smartai waypoints also need a bit love. will take care of it soon(tm)

offl:
Can be merged without additions to `waypoints`
335 was without orientation column in `waypoints` for 10 years, now it's clear that orientation should be always scripted

Shauren:
> Can be merged without additions to `waypoints` 335 was without orientation column in `waypoints` for 10 years, now it's clear that orientation should be always scripted

Actually it cannot - you might think it's not neccessary however this isn't the full story. This PR modifies code shared by both waypoint systems and without explicitly supporting SAI in it, it will make all SAI waypoints behave as if they are intended to face 0 on each point

Shauren:
github borked, its merged (4747515872775553de35cc3267e1ccd5e0076ca4)

PR #26702, opened 2021-07-14
Core/Objects: Units possessing another ones must use the detection of the latter

### LINKED ISSUE #23528: [3.3.5] Core: Players visibility should use invisibility detection from the currently controlled unit
Title. In cases of auras like invisibility being applied on the owner the possessed unit should not have the vision restrictions applied on it.
Also to note, the client will also use the indivisibility glow filter effect but shouldn't.
Example spell: Coax Marmot (38544) - applies an invisibility aura on the caster and the controlled unit shares the vision effects.
https://youtu.be/Amsn4iBbxAI?t=14

Steps to reproduce:
1. .quest add 10720
2. .go xyz 3346.222168, 5783.050781, -7.563022, 530
3. /use Kodohide Drum

rev. eeba12778b330f799eab1530da14bcd8a01b8269 (yesterday's)

meji46:
**Changes proposed:**

-  Units with invisibility auras should be able to see units that cannot detect them.

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #23528

**Tests performed:**
Builds, tested in-game

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1650,8 +1650,10 @@ bool WorldObject::CanDetect(WorldObject const* obj, bool ignoreStealth, bool che
     if (Unit const* thisUnit = ToUnit())
     {
         if (thisUnit->isPossessing())
+        {
             if (Unit* charmed = thisUnit->GetCharmed())
                 seer = charmed;
+        }
         else if (Unit* controller = thisUnit->GetCharmerOrOwner())
             seer = controller;
     }
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1645,10 +1645,16 @@ bool WorldObject::CanDetect(WorldObject const* obj, bool ignoreStealth, bool che
 {
     WorldObject const* seer = this;
 
+    // If a unit is possessing another one, it uses the detection of the latter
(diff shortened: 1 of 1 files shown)


jackpoz:
Does this PR still work with the case of https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=66 mentioned in the removed comment ?

meji46:
> Does this PR still work with the case of https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=66 mentioned in the removed comment ?



> Does this PR still work with the case of https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=66 mentioned in the removed comment ?

I guess not. i have seen the following comment "Since 4.0.1, invisible Mages can see other players, although they cannot see other invisible mages (unless party members)".

Perhaps this mechanic should only be applied for INVISIBILITY_GENERAL (0)?

kelno:
I think the current logic is fine, it's just missing that it should use detection from the currently controlled unit rather than always the player character.

meji46:
@kelno I prefer your approach so I have changed the logic. Now the unit uses detection from the currently controlled unit.

Tested with https://www.wowhead.com/spell=38544/coax-marmot and https://www.wowhead.com/spell=66/invisibility

jackpoz:
Thanks for the PR :)

Issue #26721, opened 2021-07-18
[3.3.5] Game Event: Fall of the Infernals in Honor Hold


Freeman256:
### Description

Infernals periodically drop in Honor Hold. Because of this, the characters enter the state of battle, however, after the end of the fall of the infernals, the state of the battle is not removed for a very long time (more than 20 minutes, I timed it), or it may not be removed at all. 

### Expected behaviour

After the fall of the Infernals in Honor Hold, the battle condition should be removed soon (ideally immediately after the end). 

### Steps to reproduce the problem

1. Create an alliance character;
2. **.go xyz  -680 2700  95 530**;
3. Wait for the beginning of the fall of the infernals in Honor Hold;
4. Make sure that your character enters a state of battle when an infernal falls nearby (in fact, it is not an infernal that falls, but just a stone); 
5. Make sure that your character does not leave the combat state after the end of the fall of the infernals in Honor Hold;
6. Use **.gm on** or **Alt+F4** if you want to eat, drink, use the flight master or if you want to mount.

### Branch

3.3.5

### TC rev. hash/commit

https://github.com/TrinityCore/TrinityCore/commit/a47926ceab7091dd0c06b6bb16a0d1f64bd30a63
TrinityCore rev. 2021-07-16 ‏‎18:51:35 +0300 (3.3.5 branch) (Win64, Release, Static)

### Operating system

Win7 SP1 x64 with all updates (build 6.1.7601).

### Custom changes

None.

jackpoz:
There is agm command to check with which npc you are still in combat with, maybe we have to add a no-combat flag to that (or maybe an old script has to be changed)

Jinnaix:
Use ".debug combat" on yourself to get the information we are looking for in this case.
Im personally interested in this, since i thought its all good now with the honor hold event, after i fixed the fear bug that existed before.

Jinnaix:
Valid.

Caused by the Infernal Rain NPC´s (Entry 18729).
There are 5 static one (spawn id 86525 - 86529) and 10 moving around (spawn id 86515 - 86524).

The static ones take the moving ones as targets for spells.

After the event is done, several NPC and potentially the player himself can get stuck in combat with the 5 static Infernal Rain NPC.

Example Picture taken from Spawnid 86525 after event is over:

![image](https://user-images.githubusercontent.com/37972361/126171806-09f2ab20-61b2-45bb-8589-9704eafc404a.png)

[...]

PR #26744, opened 2021-07-24
Core/Unit: Changes to stunned unit behavior

Jinnaix:
**Changes proposed:**

-  Send units back home after stun in case there is no combat
-  should get rid of remaining problems outlined here: https://github.com/TrinityCore/TrinityCore/issues/26721#issuecomment-882577305

With a very high chance, several units could get stuck on their current position like shown in the screenshot below.
With this change, affected units return home in all test cases (10/10)
<details>

WoWScrnShot_072421_121331

</details>

- [x] 3.3.5

**Issues addressed:**

Closes  https://github.com/TrinityCore/TrinityCore/issues/26721

**Tests performed:**

- [x] Builds
- [x] tested ingame

- [ ] If someone is aware of any situation where this could break something, please let me know or check it with this PR. I cant think of any case where the current behavior would be wanted.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -11326,6 +11326,10 @@ void Unit::SetStunned(bool apply)
 
             RemoveUnitMovementFlag(MOVEMENTFLAG_ROOT);
         }
+
+        // sends a unit back home in case the effect ends and the unit is not in combat
+        if (IsAlive() && !IsPlayer() && !IsInCombat())
+            GetMotionMaster()->MoveTargetedHome();
     }
 }
 


Ovahlord:
This will break basically any generic events that use self stun auras for cosmetic reasons. I'd say you better investigate honor hold event sniffs if this out of combat stuff is even suposed to happen.

offl:
I already found one of them https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=19481#sai-actionlist-1948100

Jinnaix:
Yes, just tested this quest/sai script with these changes and it breaks.
I´ll continue to find another way for honor hold, as we talked about it yesterday.
Thanks for your feedback!

Issue #26592, opened 2021-06-12
Quest: Supplying the Front (1578)


Freeman256:
https://classic.wowhead.com/quest=1578/supplying-the-front
https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?quest=1578

Description:
Quest not working. To complete the task, the player needs to have the blacksmithing profession raised by 35. If you bring the required items ([Copper Axe](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?item=2845) and [Copper Chain Belt](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?item=2851)) to [Thorvald Deepforge](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=6030), then the quest cannot be turned in. I tried restart the game with deleting the cache, dropping one of the required items and crafting it again - to no avail, restarting the server also didn't fix the problem. 

Expected behaviour:
After making the above items in the indicated quantities, the [Thorvald Deepforge](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=6030) must allow this quest to be completed.

Steps to reproduce the problem:
1. 	Create an Alliance character and raise the level of blacksmithing to 35. 
2.	Take the quest [Supplying the Front](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?quest=1578) from the [Tormus Deepforge](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=6031) in Ironforge.
3.	Craft the [Copper Axe](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?item=2845) and [Copper Chain Belt](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?item=2851) by 6.
4.	Take these items to [Thorvald Deepforge](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=6030) in the southern guard tower of Loch Modan.
5.	Make sure that, despite the conditions being met, the quest cannot be completed.
6.	Become sad.

Branch:
3.3.5

TC rev. hash/commit: https://github.com/TrinityCore/TrinityCore/commit/2acdd163fea281f467f074d267e45660fc9bf99d

TrinityCore rev. 2021-06-11 17:28:34 +0300 (3.3.5 branch) (Win64, Release, Static)

Some logs from auth/world servers:
[...]

PR #26758, opened 2021-07-27
Core/Quests: Improve ItemRemovedQuestCheck logic

### LINKED ISSUE #25603: [3.3.5] Quest item in inventory does not count when completing the quest
**Description:**

This is a bit tricky to explain:
Lets say you have a quest where you need to collect 20 linen cloth. But you collect more than 20, you collect 30. 
Then you sell the first 10 that you collected, so that you still have 20 in your backpack. 
If you try to complete the quest now, it will tell you that you have not enough linen cloth in your backpack. 

As a workaround you can sell those 20 linen cloth and rebuy them, or put them on your bank and then back in your inventory. 
Then it will count them as collected. 

**Expected behaviour:**

It shouldn't matter if those 20 linen cloth are the first 20 you collected, or the last 20. 

1. Get any quest where you need to collect something and where you are able to collect more than you need (e.g. you need 20 linen cloth)
2. Collect more of those things than you need.  (like 30 linen cloth) 
3. Sell the first 10 that you collected (not the last one's you got, but the first ones, thats very important!). 
4. Try to complete the quest. it will tell you that it's not possible. 
5. Put the 20 linen cloth on your bank and back in your inventory. 
6. Now you can complete the quest.

**Branch(es):**

3.3.5 rev. 07845b4

Jildor:
**Changes proposed:**

-  Improve ItemRemovedQuestCheck logic
-  we don't need reduce by count when GetItemCount is called after 449bd9058bce87ade6831ba38d1a0b83b9cbfb18

- [X] 3.3.5
- [ ] master

**Issues addressed:**

Closes #25603

**Tests performed:**

tested in-game with quest detailed here: #26592

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -16392,14 +16392,14 @@ void Player::ItemRemovedQuestCheck(uint32 entry, uint32 count)
             {
                 QuestStatusData& q_status = m_QuestStatus[questid];
 
-                uint32 reqitemcount = qInfo->RequiredItemCount[j];
-                uint16 curitemcount = q_status.ItemCount[j];
+                uint32 reqItemCount = qInfo->RequiredItemCount[j];
+                uint16 questStatusItemCount = q_status.ItemCount[j];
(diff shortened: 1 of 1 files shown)


jackpoz:
Thanks for the PR :)

Issue #25628, opened 2020-11-02
Core/Spells: Applied out of combat periodic damaging auras causes combat


offl:
**Description:**

Auras like https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=17467 are applied out of combat and triggered damaging spell causes combat when player stays in range of spell. Creature doesn't start attack (and should not) and after leaving spell range player stucks in combat so only way to remove it is to kill creature.

**Expected behaviour:**

Such auras should not cause combat.

**Steps to reproduce the problem:**

1. .go x 302.035461 94.35424 22.281729 552 1.559
2. .debug combat

**Branch(es):** 3.3.5
**TC rev. hash/commit:**  c2a0b78a051ea3301d8769b2c11fae00e14fe095

illfated:
Out of curiosity, do you know if the information in https://classic.wowhead.com/spell=17467/unholy-aura has changed compared to the timeline in 2005-2011?

offl:
In this particular case probably no

offl:
Another example:
.go c id 20869 or .npc add temp 20869
ensure you're level 80
ensure distance between you and creature is ~14 but not lower than 10
.gm off
wait for spell to hit you
once it happens, you will be always in combat with creature until you kill it

CraftedRO:
319434544baa018339ba6284ebb910eec1418a4f still valid.

Legoso:
Links from the OP comments above, with an updated and available main address:

- https://aowow.trinitycore.info/?spell=17467/unholy-aura
- https://aowow.trinitycore.info/?spell=36717/energy-discharge

PR #26762, opened 2021-07-28
Core/Creature: rename CREATURE_FLAG_EXTRA_IGNORE_COMBAT into CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT

offl:
**Changes proposed:**

-  Make No Combat flag purely custom without any reference to original Ignore Combat flag
-  Create Ignore Combat flag which just changes react state to passive

Some quick facts to better understand these changes:

To me it looks like we need improvements here instead of making another custom flag:

```cpp
bool SpellInfo::HasInitialAggro() const
{
    return !(HasAttribute(SPELL_ATTR1_NO_THREAT) || HasAttribute(SPELL_ATTR3_NO_INITIAL_AGGRO));
}
```
For example we have #25628.
Maybe we need to check if more auras should not cause combat? Because we already used that flag several times just to make triggers with auras not cause combat at all to prevent combat stuck. I can't check some of those auras at this moment.

Slightly after that flag was implemented in TC, Ovahlord made some changes to handle it differently: https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/commit/a2fa0c2ffae36bc60c0fec68a76a9dd800b21781. If I understand correctly with those changes if creature casts immediate damage spell, it will not enter in combat with player. I checked one creature with that flag and it entered in combat with player after casting such spell. So...
Also it's a DB flag, if we use it wrongly, we'll have to change its implementation later. But since we used it for hundreds of creatures, we'll just break lots of them. And yes, since we can't change that flag at runtime, we'll have to change reactstate directly and that is too not safe. At the same time react state of creatures with that flag will be changed only few times. At least in 3.3.5. But yeah, better to create a completely new function which will change only reactstate only from passive to aggressive to match original AI action(see below). Because if by any chance implementation of that flag will be changed and it will not change react state but something else, we'll have to change all scripts where creature with that flag changes its reactstate. Otherwise those scripts will not work.

The reason why it's named Ignore Combat is because creature... ignores it, does not respond to attacks, that's it.

I was too naive to treat that flag as original Ignore Combat flag and used it only if creature has corresponding difficulty flag. Slighly later I finally understood that that flag even cannot be used where Blizz used it, it simply breaks scripts. I renamed it to match original name, I used it only if creature has corresponding difficulty flag but I can't use it in all cases where Blizz used it. Nonsense.
With current implementation we can't remove effect of that flag via SAI. We should. Some of units with that flags spawns passive and then can became aggressive. Blizz uses actions: Reset "Ignore combat" flag for unit(s) / Set "Ignore combat" flag for unit(s)

**Target branch(es):**

3.3.5

**Issues addressed:**

none

**Tests performed:**

Build, tested in-game

none

### DIFF
--- a/src/server/game/Combat/CombatManager.cpp
+++ b/src/server/game/Combat/CombatManager.cpp
@@ -44,8 +44,8 @@
         return false;
     if (a->HasUnitState(UNIT_STATE_IN_FLIGHT) || b->HasUnitState(UNIT_STATE_IN_FLIGHT))
         return false;
-    // ... both units must not be ignoring combat
-    if (a->IsIgnoringCombat() || b->IsIgnoringCombat())
+    // ... both units must be allowed to enter combat
+    if (a->IsCombatDisallowed() || b->IsCombatDisallowed())
         return false;
     if (a->IsFriendlyTo(b) || b->IsFriendlyTo(a))
         return false;
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -621,7 +621,7 @@ bool Creature::UpdateEntry(uint32 entry, CreatureData const* data /*= nullptr*/,
         ApplySpellImmune(0, IMMUNITY_EFFECT, SPELL_EFFECT_ATTACK_ME, true);
     }
 
-    SetIgnoringCombat((cInfo->flags_extra & CREATURE_FLAG_EXTRA_IGNORE_COMBAT) != 0);
+    SetIsCombatDisallowed((cInfo->flags_extra & CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT) != 0);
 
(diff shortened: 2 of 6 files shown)


offl:
>I can create a function to remove effect of Ignore Combat extra flag but with it we'll have 2 actions in SAI, one will change react state to any and second(my) will change SetIgnoreCombat which will change react to passive or aggressive
>
>so on one hand in scripts we'll call a function to change react state from passive to aggressive instead of changing react state by calling SetReactState. this way we'll do it in way Blizz does it. it's also way much more safe. on other hand there will be 2 ways to change react state and that is too not good
>
>makes me wonder why they don't have anything related to defensive react state
>
>it's probably not even needed

Anyways all that stuff is for future

Faq:
So this PR only for rename?

offl:
Yep

PR #26786, opened 2021-08-08
Core/Spells Implement target 135 : TARGET_UNIT_FIRST_IN_LINE

Traesh:
**Changes proposed:**

-  Handle target 135 : TARGET_UNIT_FIRST_IN_LINE

- [ ] 3.3.5
- [X] master

**Issues addressed:**

None that i know

**Tests performed:**

Built, tested ingame with spell 320782 (triggered by 320780)

None


Shauren:
This is indeed a line target, however its implementation is not quite correct
It does not form a line in front of caster but a line between caster and DEST
I propose this target to be named `TARGET_UNIT_FIRST_ON_LINE_CASTER_TO_DEST`
Also, target 134 is the same, without stopping on first unit (`TARGET_UNIT_LINE_CASTER_TO_DEST`)

src/server/game/Spells/SpellInfo.cpp
```diff
@@ -350,9 +350,9 @@ SpellImplicitTargetInfo::StaticData  SpellImplicitTargetInfo::_data[TOTAL_SPELL_
...
     {TARGET_OBJECT_TYPE_DEST, TARGET_REFERENCE_TYPE_TARGET, TARGET_SELECT_CATEGORY_DEFAULT, TARGET_CHECK_ALLY,     TARGET_DIR_NONE},        // 132 TARGET_DEST_TARGET_ALLY
-    {TARGET_OBJECT_TYPE_NONE, TARGET_REFERENCE_TYPE_NONE,   TARGET_SELECT_CATEGORY_NYI,     TARGET_CHECK_DEFAULT,  TARGET_DIR_NONE},        // 133
-    {TARGET_OBJECT_TYPE_NONE, TARGET_REFERENCE_TYPE_NONE,   TARGET_SELECT_CATEGORY_NYI,     TARGET_CHECK_DEFAULT,  TARGET_DIR_NONE},        // 134
-    {TARGET_OBJECT_TYPE_NONE, TARGET_REFERENCE_TYPE_NONE,   TARGET_SELECT_CATEGORY_NYI,     TARGET_CHECK_DEFAULT,  TARGET_DIR_NONE},        // 135
+    {TARGET_OBJECT_TYPE_DEST, TARGET_REFERENCE_TYPE_CASTER, TARGET_SELECT_CATEGORY_CONE,    TARGET_CHECK_ALLY,     TARGET_DIR_NONE},        // 133 TARGET_UNIT_LINE_CASTER_TO_DEST_ALLY
+    {TARGET_OBJECT_TYPE_DEST, TARGET_REFERENCE_TYPE_CASTER, TARGET_SELECT_CATEGORY_CONE,    TARGET_CHECK_ENEMY,    TARGET_DIR_FRONT},       // 134 TARGET_UNIT_LINE_CASTER_TO_DEST_ENEMY
+    {TARGET_OBJECT_TYPE_DEST, TARGET_REFERENCE_TYPE_CASTER, TARGET_SELECT_CATEGORY_CONE,    TARGET_CHECK_DEFAULT,  TARGET_DIR_FRONT},       // 135 TARGET_UNIT_FIRST_ON_LINE_CASTER_TO_DEST
```
Shauren, line 355: These targets will not work like this, that is not what I meant when I said DEST in earlier comment

`TARGET_OBJECT_TYPE_UNIT` was correct (units are added to target list)
`TARGET_DIR_***` doesn't do anything with `TARGET_SELECT_CATEGORY_CONE`
All 3 of these targets need changes in `WorldObjectSpellConeTargetCheck` to change where the cone/line is aimed at (instead of always being in front or back)

Shauren: obraz

Traesh: @Shauren Should be cleaner now, let me know if i got something wrong with the spell target system
 
 On your illustration the line continue even after DEST, that's not what i did here (i check between caster & dest). Should dest only be used as angle ?

Shauren: Yes, DEST only determines the angle, radius (or range? same as cones) determines distance

Shauren:
Hmm, I might have found a counter-example for target 135 being only FIRST... (might be a line that stops on DEST and doesnt go beyond)
https://www.wowhead.com/spell=186124/wrought-chaos

Traesh:
@Shauren You're right, after checking a few spells135 isn't only first, it's probably just "TARGET_UNIT_LINE_CASTER_TO_DEST"

"Only first" behaviour is either handled by MaxTargets (example : 343986 Ritual of Woe from HoA last boss) or by SPELL_ATTR5_SINGLE_TARGET_SPELL (example : 320782 Power Overwhelming from DoS Manastorms)

Internally SPELL_ATTR5_SINGLE_TARGET_SPELL probably also set MaxTargets to 1 since there is no reason to have multiple targets

The only thing to do here then is order targets by distance before resizing to MaxTargets (is WorldObjectListSearcher result orderer by distance already or is it random ?) and set MaxTargets to 1 on loading if spell IsSingleTarget()

PR #26788, opened 2021-08-08
adding packet for legendary creator

Mistixx:
**Changes proposed:**

-  the legendary creator interface pack was made like this which I am going to show
-  
-  

- [ ] 3.3.5
- [ ] master

**Issues addressed:**

**Tests performed:**

- [ ] 
WoWScrnShot_071021_230241
WoWScrnShot_071121_010621


Shauren:
I have not actually checked if the structure is correct but I will say something else

We do not add packets that are unused

Mistixx:
Yes, check and the structure is still the same even in 9.1 but ok I understand is that I made these packets for TrinityCore because I spoke with funjoker and I told him that I was doing this for Trinity ok no problems shauren then I will close the commit

PR #26797, opened 2021-08-09
Core/SAI: Remove SMART_ACTION_INSTALL_AI_TEMPLATE

offl:
**Changes proposed:**

It looks like we can create a new table where we will store predefined events or sets of events(the ones which are currently stored directly in core) and install them by SMART_ACTION_INSTALL_AI_TEMPLATE. Those events probably can be stored even in main table but with new source type

However what we need is not install predefined sets of events in our scripts but tie them directly to creature guid or entry.
In other words currently that action is useless because of bad design, it could be used if improved, but we need different scripting methods

**Target branch(es):**

3.3.5

**Issues addressed:**

none

**Tests performed:**

didn't checked anything


ccrs (approved):
lookin good

offl:
Those templates are heavily outdated, we don't use such scripting methods anymore and even if they were updated, it's not how such things are supposed to be used because you don't need a template with predefined action set but without spell id and timers(which you set in that action) but already created action or sets of actions you can tie to any creature. For example instead of having 30 similar scripts with Shoot spell with same id you can create one and tie it to different creatures

PR #26816, opened 2021-08-14
deps/fmt: Upgrade to 7.1.3

Carbenium:
Printing enum values as integers was broken before 7.1.0.
Upgrade the library so we include commits

https://github.com/fmtlib/fmt/commit/20d4f2e83677be035a15693127e404c4a5d68ac9
https://github.com/fmtlib/fmt/commit/86287b8d5625ef390da6ef10fc77a872b8ad41ac

which fix the issue.

Minimal repo:
```c++
#include <fmt/printf.h>

enum Values : uint8_t {

    VAL_0 = 0,
};

int main() {
    fmt::printf("%d | asd", VAL_0);
}
```
Godbolt: https://godbolt.org/z/nnb8x4xdx (Change library version to reproduce)

Output with 7.0.0: `| asd`
Output with 7.1.3: `0 | asd`

- [x] 3.3.5
- [ ] master

**Issues addressed:**
`.instance listbinds` lead to incomplete output due to this issue.

**Tests performed:**
* build
* run server
* `.instance listbinds` shows the expected output

### DIFF
--- a/dep/PackageList.txt
+++ b/dep/PackageList.txt
@@ -22,7 +22,7 @@ efws (Entropia File System Watcher - crossplatform file system watcher)
 
 fmt (a small, safe and fast formatting library)
   https://github.com/fmtlib/fmt
-  Version: 7.0.3 https://github.com/fmtlib/fmt/releases/tag/7.0.3
+  Version: 7.1.3 https://github.com/fmtlib/fmt/releases/tag/7.1.3
 
 G3D (a commercial-grade C++ 3D engine available as Open Source (BSD License)
   http://g3d.sourceforge.net/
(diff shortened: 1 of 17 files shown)


Shauren:
Well, we usually always explicitly cast all uint8 based enums to uint32 when printing...

jackpoz:
Please update https://github.com/TrinityCore/TrinityCore/blob/3.3.5/dep/PackageList.txt too

Carbenium:
@jackpoz Done.
@Shauren Also true. Then again if fmt is doing the conversion to the underlying type, we should be safe, right?

Winfidonarleyan:
https://github.com/fmtlib/fmt/releases/tag/8.0.1

Shauren:
@Winfidonarleyan not changing major version, sticking to 7.x for now

Winfidonarleyan:
Okay. I just wanted to say that there is a version 8 :)

PR #26827, opened 2021-08-16
Core/AreaTrigger Switch polygon vertices from Template to SpellMisc

Traesh:
**Changes proposed:**

-  Switch polygon vertices from AreatriggerTemplate to AreaTriggerMiscTemplate
-  A single AreatriggerId can have different vertices depending the spell
-  Proof from sniff with at entry 23953 having 3 different possible vertices configuration depending on the spell : https://pastebin.com/8AYEk3DX
- For those who know SL dungeons, this is the 3 spells used by DoS last boss for Master of Death

- [ ] 3.3.5
- [X] master

**Issues addressed:**

None

**Tests performed:**

Built, test ingame

- Maybe double check the transition SQL request, not sure it handle everything correctly


sql/updates/world/master/2021_08_16_01_world.sql
```diff
@@ -0,0 +1,12 @@
+RENAME TABLE `areatrigger_template_polygon_vertices` TO `spell_areatrigger_vertices`
```
Shauren, line 1: missing `;`

sql/updates/world/master/2021_08_16_01_world.sql
```diff
@@ -0,0 +1,14 @@
...
+ALTER TABLE `spell_areatrigger_vertices` ADD COLUMN `SpellMiscId` INT(10) NULL FIRST;
+
+UPDATE `spell_areatrigger_vertices`
+LEFT JOIN `spell_areatrigger` ON `spell_areatrigger_vertices`.AreaTriggerId = `spell_areatrigger`.AreaTriggerId
+SET `spell_areatrigger_vertices`.SpellMiscId = `spell_areatrigger`.SpellMiscId;
+
+DELETE FROM `spell_areatrigger_vertices` WHERE SpellMiscId IS NULL;
+
+ALTER TABLE `spell_areatrigger_vertices` DROP COLUMN `SpellMiscId`,
+                                         CHANGE `AreaTriggerId` `SpellMiscId` INT(10) UNSIGNED NOT NULL,
```
Shauren, line 12: This query doesn't make sense, that column still holds areatrigger ids

Ovahlord:
This explains why Garothi Worldbreaker's surging fel was using 1 as area trigger id while it's most likely a own thing

Shauren:
0, but yes

Issue #26728, opened 2021-07-19
DEV/DB: SMART_ACTION_WP_START reactState param


offl:
### Description

I'm going to remove reactState param of SMART_ACTION_WP_START. If react state indeed should be changed when creature starts waypoint, I'll change it in separate SMART_ACTION_SET_REACT_STATE and change reactState param to 3. If it should not be changed, reactState param will be changed to 3 without additional changes
When reactState param will be changed to 3 in all 855 actions, we'll remove that param completely
These changes can break some scripts or make them buggy because I can't always know if react state of those 550 creatures which were made defensive and passive should really be changed or not

### Expected behaviour

none

### Steps to reproduce the problem

none

### Branch

3.3.5

### TC rev. hash/commit

60379cf2c15d9cee1414fbaec2b217b89b8dc67e

### Operating system

Win 8.1

### Custom changes

none

offl:
I knew it will end like this. Removed the rest without checking code, with no doubt breaks some things. Will be fixed when reported. Any way I already removed accidentally passive react state of all vehicles with waypoints. What a waste of free time of 2 weeks. Literally same result as if I removed it without checking something.

This is probably the most dumbest thing I've ever seen and here's why.
1. In nearly all cases you don't need to change react state at all when creature starts waypoint. However it was changed in 550 cases(out of 850) to defensive or passive.
2. You can't skip react state change, best you can do is use aggressive react state, that's how it works. 0 is not DO NOTHING but REACT_PASSIVE. In escort quests if creature changes faction before waypoint is loaded, original react state from faction will be fucked up. Even if you change it to closest to faction react state, it will override react state from faction in a weird way.
3. There's no such thing as action to change react state to defensive, there's no even such thing as change react state because we use pet react states which are supposed to be used for pets and vehicles and not for every single creature. There is a flag which acts like passive react state but it too has nothing to do with react states.

CraftedRO:
So in a few words, you wasted two weeks to break some of 550 cases (because you think they were hacks), ** instead of focusing on already existing issues **, yea it sound legit.

CraftedRO:
So in a few words, you wasted two weeks to break some of 550 cases (because you think they were hacks), **instead of focusing on already existing issues**, yea it sound legit.

offl:
Where do you see the word 'hack'?
[...]

Issue #25764, opened 2020-12-25
Core/Misc: Error if unit flag is not allowed to be set manually


offl:
**Description:**

We need IsAllowedUnitFlag with possibility to remove or add unit flag from that list and throw error if any flag from that list was added in DB.
The reason why is because almost all unit flags should never be in DB because some of them are handled by server, some by client and all other by spells. We can't just remove all of them because in some cases it will broke stuff so we'll need to replace it by auras(stun, pacify, immune). In some other cases flags can be removed without changes(in combat, for example).

Full list of flags, as start we can make UNIT_FLAG_SKINNABLE not allowed. Other I can add later
```cpp
    UNIT_FLAG_SERVER_CONTROLLED     = 0x00000001,           // set only when unit movement is controlled by server - by SPLINE/MONSTER_MOVE packets, together with UNIT_FLAG_STUNNED; only set to units controlled by client; client function CGUnit_C::IsClientControlled returns false when set for owner
    UNIT_FLAG_NON_ATTACKABLE        = 0x00000002,           // not attackable
    UNIT_FLAG_REMOVE_CLIENT_CONTROL = 0x00000004,           // This is a legacy flag used to disable movement player's movement while controlling other units, SMSG_CLIENT_CONTROL replaces this functionality clientside now. CONFUSED and FLEEING flags have the same effect on client movement asDISABLE_MOVE_CONTROL in addition to preventing spell casts/autoattack (they all allow climbing steeper hills and emotes while moving)
    UNIT_FLAG_PLAYER_CONTROLLED     = 0x00000008,           // controlled by player, use _IMMUNE_TO_PC instead of _IMMUNE_TO_NPC
    UNIT_FLAG_RENAME                = 0x00000010,
    UNIT_FLAG_PREPARATION           = 0x00000020,           // don't take reagents for spells with SPELL_ATTR5_NO_REAGENT_WHILE_PREP
    UNIT_FLAG_UNK_6                 = 0x00000040,
    UNIT_FLAG_NOT_ATTACKABLE_1      = 0x00000080,           // ?? (UNIT_FLAG_PLAYER_CONTROLLED | UNIT_FLAG_NOT_ATTACKABLE_1) is NON_PVP_ATTACKABLE
    UNIT_FLAG_IMMUNE_TO_PC          = 0x00000100,           // disables combat/assistance with PlayerCharacters (PC) - see Unit::IsValidAttackTarget, Unit::IsValidAssistTarget
[...]

Issue #24245, opened 2020-03-05
The monster is frozen by the hunter's Freezing Trap. It can still cast skills.


aleigood:
Description:
The monster is frozen by the hunter's Freezing Trap. It can still cast skills.

Expected behaviour:
Monsters cannot cast any skills.

Steps to reproduce the problem:
.character level 40
.go xyz -12236.657227 154.116531 6.689152 0
Release the frozen trap.
Trap the Lashtail Raptor and get as close it as possible.

Branch(es):
3.3.5

TC rev. hash/commit:
d1080af8db63c78986a7fcc94fcfb2fdfe817734

Operating system:
CentOS

Killyana:
At first look creatures using SmartAI are able to cast spells while stunned

Killyana:
I will remove this wrong cast_flags from combat scripts.

Rushor:
well sometimes the triggered flag is really needed for sai spells, else the spell is not casted at all.

several sai scripts have a spell with costs. some creatures have not mana or enery or sth and would not be able to cast the spell without the triggered flag (at least that is the state that i remember from some years ago)

Rushor:
```diff
diff --git a/src/server/game/AI/SmartScripts/SmartScript.cpp b/src/server/game/AI/SmartScripts/SmartScript.cpp
index 7f38d0f1b4..47df2f75c4 100644
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -576,7 +576,8 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                             ENSURE_AI(SmartAI, me->AI())->SetCombatMove(allowMove);
                         }
 
-                        me->CastSpell(target->ToUnit(), e.action.cast.spell, triggerFlag);
+                        if (!me->HasAuraType(SPELL_AURA_MOD_STUN))
+                            me->CastSpell(target->ToUnit(), e.action.cast.spell, triggerFlag);
                     }
                     else if (go)
                         go->CastSpell(target->ToUnit(), e.action.cast.spell, triggerFlag);
```

but this will break a lot of scripts i guess

Rushor:
```diff
diff --git a/src/server/game/AI/SmartScripts/SmartScript.cpp b/src/server/game/AI/SmartScripts/SmartScript.cpp
[...]

Issue #20648, opened 2017-10-17
DB/Instance: Wailing Caverns


ZenoX92:
**Description:**

Most of bosses in Wailing Caverns are not scripted.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. 9fa6885 2017-08-08 11:36:46 -0300 (3.3.5 branch) (Win64, Release, Static) (worldserver-daemon)

**TDB version:**  335.63

**Operating system:** Windows 7

**Suggested Fix:**
```sql
-- Lady Anacondra
UPDATE `creature_template` SET `ManaModifier`=4 WHERE `entry`=3671;
DELETE FROM `smart_scripts` WHERE `entryorguid`=3671 AND `source_type`=0 AND `id`>0;
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(3671,0,1,0,0,0,100,2,0,0,3400,4700,11,9532,64,0,0,0,0,2,0,0,0,0,0,0,0,"Lady Anacondra - In Combat - Cast Lightning Bolt"),
(3671,0,2,0,0,0,100,2,8000,12000,12000,25000,11,700,32,0,0,0,0,6,0,0,0,0,0,0,0,"Lady Anacondra - In Combat - Cast Sleep"),
(3671,0,3,0,0,0,100,2,1000,1000,50000,60000,11,8148,0,0,0,0,0,1,0,0,0,0,0,0,0,"Lady Anacondra - In Combat - Cast Thorns Aura"),
(3671,0,4,0,1,0,100,3,1000,1000,0,0,11,13236,0,0,0,0,0,1,0,0,0,0,0,0,0,"Lady Anacondra - Out of Combat - Cast Nature Channeling"),
(3671,0,5,0,2,0,100,2,0,50,15000,26000,11,23381,0,0,0,0,0,1,0,0,0,0,0,0,0,"Lady Anacondra - Between 0-50% Health - Cast Healing Touch"),
[...]

PR #26833, opened 2021-08-16
Core/SAI: Remove no longer used event types & params

offl:
**Changes proposed:**

-  remove SMART_EVENT_TARGET_HEALTH_PCT, reasons here 41275c056e7a03d140e36924b42f043625fe1a0d
-  remove SMART_EVENT_FRIENDLY_HEALTH, no longer used since 2020
-  remove SMART_EVENT_IS_BEHIND_TARGET, no longer used since 2020
-  remove params of SMART_EVENT_RESPAWN (SMART_SCRIPT_RESPAWN_CONDITION), (0c3d11c71a056acdd392ddc219185444f63b6372), we already have `conditions`

**Target branch(es):** 

3.3.5

**Issues addressed:**

none

**Tests performed:**

didn't checked

none, so far

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -2868,16 +2868,6 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
             ProcessTimedAction(e, e.event.minMaxRepeat.repeatMin, e.event.minMaxRepeat.repeatMax);
             break;
         }
-        case SMART_EVENT_TARGET_HEALTH_PCT:
-        {
-            if (!me || !me->IsEngaged() || !me->GetVictim() || !me->EnsureVictim()->GetMaxHealth())
-                return;
-            uint32 perc = (uint32)me->EnsureVictim()->GetHealthPct();
-            if (perc > e.event.minMaxRepeat.max || perc < e.event.minMaxRepeat.min)
-                return;
-            ProcessTimedAction(e, e.event.minMaxRepeat.repeatMin, e.event.minMaxRepeat.repeatMax, me->GetVictim());
-            break;
-        }
         case SMART_EVENT_MANA_PCT:
         {
             if (!me || !me->IsEngaged() || !me->GetMaxPower(POWER_MANA))
@@ -2925,22 +2915,6 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
                         return;
 
             ProcessTimedAction(e, e.event.targetCasting.repeatMin, e.event.targetCasting.repeatMax, me->GetVictim());
-            break;
-        }
-        case SMART_EVENT_FRIENDLY_HEALTH:
-        {
-            if (!me || !me->IsEngaged())
-                return;
-
-            Unit* target = DoSelectLowestHpFriendly((float)e.event.friendlyHealth.radius, e.event.friendlyHealth.hpDeficit);
-            if (!target || !target->IsInCombat())
-            {
-                // if there are at least two same npcs, they will perform the same action immediately even if this is useless...
-                RecalcTimer(e, 1000, 3000);
-                return;
-            }
-
-            ProcessTimedAction(e, e.event.friendlyHealth.repeatMin, e.event.friendlyHealth.repeatMax, target);
             break;
         }
         case SMART_EVENT_FRIENDLY_IS_CC:
@@ -3040,18 +3014,6 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
 
             ProcessAction(e, unit, var0, var1, bvar, spell, gob);
             break;
-        case SMART_EVENT_IS_BEHIND_TARGET:
-            {
-                if (!me)
-                    return;
-
-                if (Unit* victim = me->GetVictim())
-                {
-                    if (!victim->HasInArc(static_cast<float>(M_PI), me))
-                        ProcessTimedAction(e, e.event.behindTarget.cooldownMin, e.event.behindTarget.cooldownMax, victim);
-                }
-                break;
-            }
         case SMART_EVENT_RECEIVE_EMOTE:
             if (e.event.emote.emote == var0)
             {
@@ -3136,10 +3098,7 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
         {
             if (!GetBaseObject())
                 return;
-            if (e.event.respawn.type == SMART_SCRIPT_RESPAWN_CONDITION_MAP && GetBaseObject()->GetMapId() != e.event.respawn.map)
-                return;
-            if (e.event.respawn.type == SMART_SCRIPT_RESPAWN_CONDITION_AREA && GetBaseObject()->GetZoneId() != e.event.respawn.area)
-                return;
+
             ProcessAction(e);
             break;
         }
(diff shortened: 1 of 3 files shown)


src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -3136,10 +3098,7 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
         {
             if (!GetBaseObject())
```
jackpoz, line 3099: this might not be needed anymore as there are no more checks to do with GetBaseObject()

ccrs:
not a fan, I see no gain from removing those events simply because they are not used right now.

about the SMART_EVENT_RESPAWN, I see it reasonable, nothing to complain there

jackpoz:
that's a fair point, even if we don't use some of those SAI events/actions it doesn't hurt to leave them there (at least the ones that don't have another implementation already)

offl:
I was a little tired and didn't explained why I want to remove them. They are removed not because they're not used but because I know that they should not be used

I spent a lot of time to convert all SMART_EVENT_FRIENDLY_HEALTH to PTC and don't want to let people continue to use it for absolutely no reason. That event was always really bad because it's simply dumb to use HP value instead of HP percentage. Especially because we have different dungeon modes with different HP values and there are no signs that Blizz checks HP value and not percentages

Regarding SMART_EVENT_TARGET_HEALTH_PCT and SMART_EVENT_IS_BEHIND_TARGET, well, I have to say that I became a little obsessed recently with Blizz's original AI actions, hooks and stuff like that. I didn't found such AI hooks. I'm aware it does not mean that they don't have them.

It may be risky to remove those event types, especially SMART_EVENT_TARGET_HEALTH_PCT, I'm aware of it. We can restore them later if they will be needed. However at this point I don't see why we should leave posibility to use them. Especially because they are useless, I mean core does not care when WE ARE TRYING TO use those spells(both events are used only for spells), it already checks attributes(SMART_EVENT_IS_BEHIND_TARGET) and aura states(SMART_EVENT_TARGET_HEALTH_PCT) of spells used in those events and does not care when you want you cast those spells. By using those event types we check everything twice. Also I prefer to remove things completely rather than wait to people to use them again in scripts for no reason because converting them again takes time

***

>events are tools to be able to script stuff, why having less tools?

I have completely different opinion for that now. Some people say 'SAI is too much limited'. That is not entirely correct. I mean there indeed should be more things but at the same time there should be less things or some things should be handled completely differently or should be strongly limited and not only for DB(#25764). Speaking of safety, some parts of SAI are old ticking bombs that explodes periodically. Solving one problem takes lot of time and breaks tens of scripts(#26728) and that will happen over and over again

We have all Blizz's AI actions, hooks and even params of their analogue of SAI for years(actually from beginning, from alpha). After checking that stuff, including leaked recently enums, I ended up with such conclusion:

Some of their flags has comments like 'TALK TO A PROGRAMMER FIRST'. This is my favorite so far:
>Always, ALWAYS tries to stand right on top of his move to target. ALWAYS!!	

Some things are limited to programmers only
>(Programmer Only) Don't consume charge if item deleted

So their development team has programmers and designers. Because of all those comments, I think their designers probably don't know anything about how core works(same with 90%+ of emu DB devs). Designers uses analogue of our SAI. Do they script bosses? Maybe only if they know how to do it, I mean if they not create only quests, creatures and stuff like that? Maybe there is another type of designers or maybe they work in team and everybody does only what they know? Maybe only programmers script bosses? Unfortunately I don't know much about all that stuff but unexpectantly here it really does matter to know how development process of a game works

At the same time their analogue of SAI looks to me as system which is used not only for simple scripts but even for bosses because of the way that system constructed. It affects all creatures and objects in game, not only the ones we script only in DB. It doesn't look like they want to have duplicating logic and avoid to use that system for bosses. But at the same time it doesn't look like their analogue of SAI is used to script everything we script by using BossAI because it's too limited

Designers can do in scripts only limited amount of things(so do we, DB devs can operate only with limited amount of actions and event types), as it's supposed to be. But even that way Blizz's devs somehow can make really weird things in their DB scripts(so do we)

>We have all Blizz's AI actions, hooks and even their params of their analogue of SAI for years.

Would be nice to focus on that when creating something new or adjusting something old. Simply to not create monsters like #26728. The only exceptions are iternal systems we don't even have implemented, in other words in some cases we have no chose and have to create and use event types and AI actions which Blizz never used. But in lots of other cases I simply see no reason to create something different that will just cause problems in future

jackpoz:
SMART_EVENT_TARGET_HEALTH_PCT and SMART_EVENT_IS_BEHIND_TARGET are not bound to cast spells, one could even just send a chat message when those events happen for whatever reason.
SMART_EVENT_FRIENDLY_HEALTH has been the only working health related event for a long time until SMART_EVENT_TARGET_HEALTH_PCT was implemented properly.

I would prefer to keep these 3 events and remove the rest. We don't have to use them in TC, we don't have to approve fixes by users that include them, but it costs very little to keep them implemented.

offl:
>SMART_EVENT_TARGET_HEALTH_PCT and SMART_EVENT_IS_BEHIND_TARGET are not bound to cast spells, one could even just send a chat message when those events happen for whatever reason.

Should never happen. This is not even a thing

>We don't have to use them in TC, we don't have to approve fixes by users that include them, but it costs very little to keep them implemented.

#20648 dude literally changed WDB fields because creatures had too less mana to cast spells from his scripts
That was pushed to the source, nobody checked that, nobody paid attention
#24245 too nobody checked. Dude used that flag in nearly all his scripts for months. Nobody paid attention, all that stuff was added to the source. How I removed it knowing it's needed in some cases? Well, I removed it from DB and then spawned all those creatures and checked if they still can cast spell without that flag. Had a damn good time
#26728 why is this even a problem? Why people made 500 creatures passive or defensive? What was on their mind? Of course in lots of cases they just used default value(0) but I can swear some people used 0 or 1 because they were thinking they have to do it. Or they will overwrite default react state(faction, civilian flag) which is not even possible. I checked those 500 creatures and I know what I'm talking about
What I think about SAI's scripting library? Can't even chose what to say. Let's say hundreds of scripts are so awful that they requires no updates but full rewrite. Thousands of scripts should updated with modern standards

Meh, 2 hours of writing for nothing
No regrets, now I know what that Blizz's SAI is and who uses it, sadly you don't understand that

ccrs:
Its nice to see how you raise points, answer them yourself and then extract your own assuptions based on those autoanswered points about the perception you have on people's knowledge and awareness

jackpoz:
After talking about the issues that these events can cause, especially when used incorrectly, and how long it takes for us to fix those issues, it looks important enough for us to not support these events anymore.

jackpoz:
One thing that really makes no sense with these events is to use them to replicate the checks of Spell::CheckCast:
- take an example with a spell that requires 1k mana, target with hp < 20% and caster behind target
- which event would you pick ? SAI_EVENT_TARGET_HP_PCT ? SAI_EVENT_TARGET_BEHIND ? SAI_EVENT_HAS_MANA ?
- whatever is picked, it will not be enough to apply ALL the checks that Spell::CheckCast does, making it an half-baked solution where the event is triggered but the action fails

What would make sense in this would be a SAI_EVENT_CAN_CAST that does all the checks that Spell::CheckCast() does, but then that's the same as just casting the spell and checking the result.

@offl thanks for helping me realize how non-sensical using these events as "spell checks" is

jackpoz:
I will work on implementing a warning on SAI actions/events that we suggest not to use. It will allow TC to get notified if a contribution uses them and at the same time allow users to keep using them if they want. Being a log, it can be turned off through configs.

offl:
>take an example with a spell that requires 1k mana, target with hp < 20% and caster behind target

There is no such spells, SMART_EVENT_TARGET_MANA_PCT is not even used currently(surprising)

PR #26836, opened 2021-08-18
Remove combo points on a creature when it evades

kelno:
**Changes proposed:**

-  Remove combo points on a creature when it evades

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Whenever a player engages a creature and gets combo points on it, they're not removed and the player can start combat again starting with >0 combo points.

**Tests performed:**

Does build & tested in game.

None

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -303,7 +303,7 @@ bool CreatureAI::_EnterEvadeMode(EvadeReason /*why*/)
     }
 
     me->RemoveAurasOnEvade();
-
+    me->ClearComboPointHolders(); // Remove all combo points targeting this unit
     me->CombatStop(true);
     me->LoadCreaturesAddon();
     me->SetLootRecipient(nullptr);


jackpoz:
Is that how it worked on retail ?

kelno:
Hmm I didn't even think about about checking it, I had a very strong assumption it made no sense.
Might need confirmation then.

mdX7:
On current retail they're decaying, ref: https://www.youtube.com/watch?v=19iOKOxg-Zs


> 
> 
> Is that how it worked on retail ?

seems to be that way on classic tbc ptr, ref: https://www.youtube.com/watch?v=CZbSCtrm8dQ
they might've changed that in wotlk, but i dont think so

jackpoz (approved):

jackpoz:
Looks good to me, if none has anything to day I will merge it later on

ccrs:
you should check overrides to EnterEvadeMode, there are many, and they most likely should have this too

jackpoz:
makes one wonder why people chose to copy&paste EnterEvadeMode() body instead of using parameters or anything better than that...

jackpoz:
let's start with this implementation

jackpoz:
Can this PR be merged or is there anything left to do ? Were the last changed tested ?

jackpoz:
Thanks for the PR 🚀 😃

PR #26837, opened 2021-08-18
Don't count character pending deletion towards character per account limit

kelno:
**Changes proposed:**

- Title
- Also prevent to list more than 10 characters in char enum, in case a character gets undeleted.

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

- Fix the shown character number in realm selection.
- Don't prevent players to create characters on their account when they have less than the limit.

**Tests performed:**

Does build.

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -245,7 +245,7 @@ void WorldSession::HandleCharEnum(PreparedQueryResult result)
                 ++num;
             }
         }
-        while (result->NextRow() && num < sWorld->getIntConfig(CONFIG_CHARACTERS_PER_REALM)); // client shows an error if more than 10 characters are listed
+        while (result->NextRow() && num < MAX_CHARACTERS_PER_REALM); // client shows an error if more than 10 characters are listed
     }
 
     data.put<uint8>(0, num);
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -245,7 +245,7 @@ void WorldSession::HandleCharEnum(PreparedQueryResult result)
                 ++num;
             }
         }
-        while (result->NextRow());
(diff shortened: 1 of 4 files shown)


src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -245,7 +245,7 @@ void WorldSession::HandleCharEnum(PreparedQueryResult result)
                 ++num;
             }
         }
-        while (result->NextRow());
+        while (result->NextRow() && num < sWorld->getIntConfig(CONFIG_CHARACTERS_PER_REALM)); // client shows an error if more than 10 characters are listed
```
jackpoz, line 248: if 10 shows an error, shouldn't this check for 10 (or a const set to 10) instead of a config value ?

kelno: I did have the same question while typing this. I think both are valid:
CONFIG_CHARACTERS_PER_REALM is enforced to be no more than 10 at config load time.

I'd say it's a game design decision here:
With this right now, if a server owner sets the limit to 5 the player won't see the 6th character either, I think it makes sense to not allow someone to have more character available than defined in any case, but it's really not obvious that the character is just hidden and still exists.
Thus I prefer the constant solution actually. Your call, I can change it and use the constant at config load times as well, 10 is hardcoded there as well.

ccrs: I'd leave the constant 10

jackpoz: I would like to have the "10" here and where the config range is enforced to be replaced by a constant defined once and used in both places.

kelno: Updated now. I wasn't sure where to place the constant though.

jackpoz:
gcc build is broken but that's due to last 335 change, ignore it

jackpoz:
Thanks for the PR 🚀 😃

PR #26859, opened 2021-08-24
Core/Combat Improve extra attacks handling

trickerer:
**Changes proposed:**

- Fix targeting inconsistency. Many creatures use extra attack spells directly, not as passive aura. Such casts may cause caster being selected as damage target.
- Add proper chain proc rules.
- Trigger actual attacks on next tick to prevent infinite loop.

- [x] 3.3.5
- [ ] master

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -444,8 +444,7 @@ void Unit::Update(uint32 p_time)
 
     m_combatManager.Update(p_time);
 
-    if (_lastDamagedTargetGuid)
-        _lastDamagedTargetGuid = ObjectGuid::Empty;
+    _lastDamagedTargetGuid = ObjectGuid::Empty;
     if (_lastExtraAttackSpell)
     {
         while (!extraAttacksTargets.empty())
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -314,7 +314,6 @@ Unit::Unit(bool isWorldObject) :
     m_modAttackSpeedPct[OFF_ATTACK] = 1.0f;
     m_modAttackSpeedPct[RANGED_ATTACK] = 1.0f;
 
-    m_extraAttacks = 0;
     m_canDualWield = false;
 
     m_movementCounter = 0;
@@ -444,6 +443,22 @@ void Unit::Update(uint32 p_time)
     ASSERT(!m_procDeep);
 
     m_combatManager.Update(p_time);
+
+    if (_lastDamagedTargetGuid)
+        _lastDamagedTargetGuid = ObjectGuid::Empty;
+    if (_lastExtraAttackSpell)
+    {
+        while (!extraAttacksTargets.empty())
+        {
+            auto itr = extraAttacksTargets.begin();
+            ObjectGuid targetGuid = itr->first;
+            uint32 count = itr->second;
+            extraAttacksTargets.erase(itr);
+            if (Unit* victim = ObjectAccessor::GetUnit(*this, targetGuid))
+                HandleProcExtraAttackFor(victim, count);
+        }
+        _lastExtraAttackSpell = 0;
+    }
 
     // not implemented before 3.0.2
     if (uint32 base_att = getAttackTimer(BASE_ATTACK))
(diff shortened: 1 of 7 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -445,6 +444,22 @@ void Unit::Update(uint32 p_time)
 
     m_combatManager.Update(p_time);
 
+    if (_lastDamagedTargetGuid)
+        _lastDamagedTargetGuid = ObjectGuid::Empty;
```
jackpoz, line 448: should the if check be removed here ? in the end the result is always ObjectGuid::Empty

offl:
> Many creatures use extra attack spells directly, not as passive aura.

Yes, in nearly all or all* cases they should proc from auras

> Such casts may cause caster being selected as damage target.

Yes, *in some cases Thrash aura is not shown in sniffs and if you try to cast Thrash(proc) directly using original target type(self), creature will damage itself and not victim. Works if you cast it on victim

src/server/game/Entities/Unit/Unit.cpp
trickerer: The purpose of resetting it is that extra attack spells direct casts add extra attacks on current target and not some dead / non-existant / last attacked unit. See Unit::AddExtraAttacks().
Imagine a situation where mob hits player1and then gets taunted by player2 30 yds away. Mob targets player2 and casts Thrash. It will trigger extra attacks on player1 unless `lastDamagedTargetGuid` is empty.

jackpoz: I don't think you understoot what I mean.

Take `if (_lastDamagedTargetGuid)`, it can either be true or false:
- if it's true, `_lastDamagedTargetGuid = ObjectGuid::Empty;` will be called, meaning the stored value after the `if` check will be `ObjectGuid::Empty`
- if it's false, it means the stored value is already `ObjectGuid::Empty`

No matter if the `if` check evaluates to true or false, the value of `_lastDamagedTargetGuid` is always `ObjectGuid::Empty` at the end.

It's more of nitpicking, it might be that this PR is big enough to skip these small changes and focus instead on the big picture, my bad.

trickerer: Ah, sorry, didn't read carefully enough.

Yeah, I agree. Performance is pretty much unrelated here, so this check is unnecessary.

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -989,8 +989,22 @@ bool AuraEffect::CheckEffectProc(AuraApplication* aurApp, ProcEventInfo& eventIn
...
+                    Unit const* source = eventInfo.GetActor();
+                    uint32 lastExtraAttackSpell = source->GetLastExtraAttackSpell();
+
+                    // Patch 1.12.0(?) extra attack abilities can no longer chain proc themselves
+                    if (lastExtraAttackSpell == triggerSpellId)
+                        return false;
+
+                    // Patch 2.2.0 Sword Specialization (Warrior, Rogue) extra attack can no longer proc additional extra attacks
+                    // 3.3.5 Sword Specialization (Warrior), Hack and Slash (Rogue)
+                    if (lastExtraAttackSpell == 16459 || lastExtraAttackSpell == 66923)
```
jackpoz, line 1004: Not so nice to have hardcoded spell ids. Having enums/constexpr is an improvement, is there any other more generic way ?

src/server/game/Entities/Unit/Unit.h
```diff
@@ -1883,6 +1888,10 @@ class TC_GAME_API Unit : public WorldObject
         int8 m_comboPoints;
         std::unordered_set<Unit*> m_ComboPointHolders;
 
+        uint32 _lastExtraAttackSpell;
```
jackpoz, line 1891: please initialize this field in the constructor

src/server/game/Spells/Auras/SpellAuraEffects.cpp
trickerer: Actually these spells (including talents themselves) are the only ones with spell icon 1462 (no spell family or flags). Not the best way to check spells but still better than hardcoded values I guess.

stoneharry: > Actually these spells (including talents themselves) are the only ones with spell icon 1462 (no spell family or flags). Not the best way to check spells but still better than hardcoded values I guess.

Checking spell icons is a horrible way to identify the spell if you are working on custom content - much better to hardcode the spell ID if it really cannot be done in a generic way, at least then the scope is locked to the entries it was intended for.

jackpoz:
Thanks for the PR 🚀 😃

PR #20787, opened 2017-11-04
Scripts/Spells: Sheath of Light - overheal scaling fix

DanVS:
**Changes proposed:**

-  Sheath of Light (53501) should scale only from effective heal done (see http://www.warcraftmovies.com/movieview.php?id=164185 at 0:15 for proof).

**Target branch(es):** 3.3.5/master

- [x] 3.3.5
- [ ] master

**Issues addressed:** ...

**Tests performed:** Tested in game.

**Known issues and TODO list:**  ...

PR #26870, opened 2021-08-28
[Core/Spells]: Use GetHeal instead of GetEffectiveHeal in Sheath of Light AuraScript

lineagedr:
**Changes proposed:**

I've noticed that when you proc Paladin talent  [Sheath of Light]  the effect does nothing.
After some debugging I've found out that GetEffectiveHeal returns 0 and GetHeal returns the heal amount

-  Use GetHeal instead of GetEffectiveHeal in Sheath of Light AuraScript

**Target branch(es):** 3.3.5

- [ ] 3.3.5

**Issues addressed:**

**Tests performed:**
Compiled and tested and it's working as intended.

None

### DIFF
--- a/src/server/scripts/Spells/spell_paladin.cpp
+++ b/src/server/scripts/Spells/spell_paladin.cpp
@@ -1850,7 +1850,7 @@ class spell_pal_sheath_of_light : public AuraScript
         Unit* target = eventInfo.GetProcTarget();
 
         SpellInfo const* spellInfo = sSpellMgr->AssertSpellInfo(SPELL_PALADIN_SHEATH_OF_LIGHT_HEAL);
-        int32 amount = CalculatePct(static_cast<int32>(healInfo->GetEffectiveHeal()), aurEff->GetAmount());
+        int32 amount = CalculatePct(static_cast<int32>(healInfo->GetHeal()), aurEff->GetAmount());
 
         ASSERT(spellInfo->GetMaxTicks() > 0);
         amount /= spellInfo->GetMaxTicks();


Shauren:
What about overhealing? GetEffectiveHeal is not supposed to count that

jackpoz:
Did you use "git blame" to find when and why it was set to use GetEffectiveHeal() ? Because I did and I found https://github.com/TrinityCore/TrinityCore/pull/20787 which is a PR that had the opposite changes.

In which case does GetEffectiveHeal() return 0 ? Did you check why it returns 0 ?

lineagedr:
@Shauren I'm not sure I follow. I'm replacing GetEffectiveHeal with GetHeal

Talent Healing part Description 3/3
```
Your critical healing spells heal the target for 60% of the healed amount over 12 seconds.
```

GetHeal returns the healed amount which we need. When I tested the talent GetEffectiveHeal was returning 0.

lineagedr:
@jackpoz thanks I'll check it out

lineagedr:
It seems to be working only if you're not full heal.

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L8341
Both if statements fail and it returns 0

ccrs:
but that's what effective heal means, right? do you have any source that backs up the change into full healing effect?

jackpoz:
Maybe https://github.com/TrinityCore/TrinityCore/pull/26870/files#diff-5155d22bb7dd572ce21909738f5b15c44dab9a8f29bed22ac9d651b80bdace4dR1846 should be changed from GetHeal() to GetEffectiveHeal() too

lineagedr:
@ccrs I guess you're right even though when the paladin is full health and he procs the talent he'll get the buff 

hea

For 0 hp.

Keader:
Guys... https://github.com/TrinityCore/TrinityCore/pull/20787

jackpoz:
> Maybe https://github.com/TrinityCore/TrinityCore/pull/26870/files#diff-5155d22bb7dd572ce21909738f5b15c44dab9a8f29bed22ac9d651b80bdace4dR1846 should be changed from GetHeal() to GetEffectiveHeal() too

@lineagedr changing that line should stop the talent from proccing when being full hp, could you please try that ?

jackpoz:
Created a PR at https://github.com/TrinityCore/TrinityCore/pull/26987 that should fix the "0 hp" case

PR #26873, opened 2021-08-29
Core/Units: removed an incorrect addition of SPELL_AURA_MOD_CRIT_CHANCE_FOR_CASTER benefits

Ovahlord:
**Changes proposed:**

-  SPELL_AURA_MOD_CRIT_CHANCE_FOR_CASTER is no longer going to be considered for basic melee attacks as this effect should only affect spells. This fixes an issue that was causing these aura effect types to increase crit chances for all melee attacks instead of only the abilities that are specified in the class mask of the effect.
- 

- 3.3.5

**Tests performed:**
- tested on 4.x with the enhancement shaman storm strike buff


Jildor:
just for document a bit, implemented here: 2feb53ecb69d66bb6df3f48d9a5f4e6a00aec1f7
on 3.3.5 only affect Point of no Escape hunter talent that affects Frost Trap, Freezing Trap and Freezing Arrow.

jackpoz:
Thanks for the PR 🚀 😃

PR #26874, opened 2021-08-29
AI/PassiveAI: explicitly prevent all triggers to enter combat

ccrs:
**Changes proposed:**

Prevent triggers to enter combat except the ones that control encounters.

Everything up to debate, dont hesitate into giving your point, idea, w/e

- [X] 3.3.5
- [ ] master

**Issues addressed:**

**Tests performed:**

- [ ] 
- [ ]

### DIFF
--- a/sql/updates/world/3.3.5/2021_0X_XX_XX_world.sql
+++ b/sql/updates/world/3.3.5/2021_0X_XX_XX_world.sql
@@ -1,0 +1,1 @@
+UPDATE creature_template SET flags_extra = flags_extra &~ 8192 WHERE (flags_extra & 128) > 0 AND (flags_extra & 8192) > 0;
--- a/sql/updates/world/3.3.5/2021_0X_XX_XX_world.sql
+++ b/sql/updates/world/3.3.5/2021_0X_XX_XX_world.sql
@@ -1,1 +1,2 @@
-UPDATE creature_template SET flags_extra = flags_extra &~ 8192 WHERE (flags_extra & 128) > 0 AND (flags_extra & 8192) > 0;
+--
+UPDATE creature_template SET flags_extra = flags_extra | 8192 WHERE (flags_extra & 128) > 0;
--- a/src/server/game/AI/CoreAI/PassiveAI.h
+++ b/src/server/game/AI/CoreAI/PassiveAI.h
@@ -44,7 +44,6 @@ class TC_GAME_API PossessedAI : public CreatureAI
         void JustStartedThreateningMe(Unit*) override { }
         void UpdateAI(uint32) override;
         void EnterEvadeMode(EvadeReason /*why*/) override { }
-
         void JustDied(Unit*) override;
 
         static int32 Permissible(Creature const* /*creature*/) { return PERMIT_BASE_NO; }
@@ -74,7 +73,6 @@ class TC_GAME_API CritterAI : public PassiveAI
 
         void JustEngagedWith(Unit* /*who*/) override;
         void EnterEvadeMode(EvadeReason why) override;
-
         void MovementInform(uint32 type, uint32 id) override;
 
         static int32 Permissible(Creature const* creature);
@@ -84,6 +82,7 @@ class TC_GAME_API TriggerAI : public NullCreatureAI
 {
     public:
         explicit TriggerAI(Creature* creature) : NullCreatureAI(creature) { }
+
         void IsSummonedBy(WorldObject* summoner) override;
 
         static int32 Permissible(Creature const* creature);
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -621,8 +621,7 @@ bool Creature::UpdateEntry(uint32 entry, CreatureData const* data /*= nullptr*/,
         ApplySpellImmune(0, IMMUNITY_EFFECT, SPELL_EFFECT_ATTACK_ME, true);
     }
 
-    if (((cInfo->flags_extra & CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT) != 0) || IsTrigger())
-        SetIsCombatDisallowed(true);
+    SetIsCombatDisallowed((cInfo->flags_extra & CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT) != 0);
 
     LoadTemplateRoot();
     InitializeMovementFlags();
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -621,7 +621,8 @@ bool Creature::UpdateEntry(uint32 entry, CreatureData const* data /*= nullptr*/,
         ApplySpellImmune(0, IMMUNITY_EFFECT, SPELL_EFFECT_ATTACK_ME, true);
     }
 
-    SetIsCombatDisallowed((cInfo->flags_extra & CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT) != 0);
+    if (((cInfo->flags_extra & CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT) != 0) || IsTrigger())
+        SetIsCombatDisallowed(true);
 
     LoadTemplateRoot();
     InitializeMovementFlags();
--- a/src/server/scripts/Northrend/ChamberOfAspects/RubySanctum/boss_halion.cpp
+++ b/src/server/scripts/Northrend/ChamberOfAspects/RubySanctum/boss_halion.cpp
@@ -1280,7 +1280,7 @@ class npc_living_inferno : public CreatureScript
 
         struct npc_living_infernoAI : public ScriptedAI
         {
-            npc_living_infernoAI(Creature* creature) : ScriptedAI(creature) { }
+            npc_living_infernoAI(Creature* creature) : ScriptedAI(creature), _instance(creature->GetInstanceScript()) { }
 
             void IsSummonedBy(WorldObject* /*summoner*/) override
             {
@@ -1290,7 +1290,7 @@ class npc_living_inferno : public CreatureScript
                 // SMSG_SPELL_GO for the living ember stuff isn't even sent to the client - Blizzard on drugs.
                 if (me->GetMap()->GetDifficulty() == RAID_DIFFICULTY_25MAN_HEROIC)
                 {
-                    scheduler.Schedule(Seconds(3), [this](TaskContext /*context*/)
+                    _scheduler.Schedule(Seconds(3), [this](TaskContext /*context*/)
                     {
                         me->CastSpell(me, SPELL_SPAWN_LIVING_EMBERS, true);
                     });
@@ -1303,6 +1303,14 @@ class npc_living_inferno : public CreatureScript
                 }
             }
 
+            bool CanAIAttack(Unit const* target) const override
+            {
+                if (CreatureBoundary const* boundary = _instance->GetBossBoundary(DATA_HALION))
+                    return CreatureAI::IsInBounds(*boundary, target);
+
+                return true;
+            }
+
             void JustDied(Unit* /*killer*/) override
             {
                 me->DespawnOrUnsummon(1ms);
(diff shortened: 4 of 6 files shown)


Jildor:
Wouldn't it be better to add flag_extra that prevent enter combat to NPC triggers so that we know that they should not enter combat and nothing is going to break instead of doing it in that unnecessary way that can break a lot of scripts?

ccrs:
scripts will be checked dont worry, valithria is just the first one.

point is, any trigger out there will get into combat if something causes the combat. nothing will ever bring out the combat.
that's not acceptable.

either I implement some sort of timed out of combat logic in triggerai

or just simply prevent combat with triggers as the rule

ccrs:
like, do/did you experience combat for X seconds in retail when a spell casted by a trigger hits you? I don't feel thats the case if you ask me

ccrs:
also there is the argument that before the combat rewrite triggers never ever get into combat unless defined/scripted to do so

Keader:
Check Yogg Saron (Ulduar) about it too

Jildor:
but I mean that a simple update in creature_template adding all the NPC triggers that we want, excluding those that can break scripts seems better to me than doing it like this by core and having to modify NPC scripts

ccrs:
> Check Yogg Saron (Ulduar) about it too

everything will be checked, dont worry, really

ccrs:
> but I mean that a simple update in creature_template adding all the NPC triggers that we want, excluding those that can break scripts seems better to me than doing it like this by core and having to modify NPC scripts

point is, there are like what? 30 scripts tops? and there are like 1826 trigger templates?

1796 trigger templates with no combat flag and 30 templates without it?

plus the scripts do still need to be checked, at least standarization is needed

ccrs:
well now I see that having the flag in the DB will give control for those in the future that want to script something... tough choice

Jildor:
Let's see how this evolves, at the moment this way of doing it does not convince me, I do not trust much that everything will Be checked and no scripts end up broken

ccrs:
these 3 look good, I've tested em for quite a bit, but feedback is welcome!

offl:
Let's explain **half of things**:
* Nearly all triggers already has CREATURE_FLAG_EXTRA_TRIGGER which is used already to do **10 tasks**, including change of react state to passive
* Now we'll add to nearly all triggers CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT which blocks combat completely
* Some day I will add Ignore Combat extra flag to tons of triggers, totems and other stuff(original flag, not custom) to creatures with corresponding difficulty flag which is original data. It will change react state to passive or something like that
* 3 flags applied to exactly same creatures will do almost same job and probably will cause mess and conflicts now and in future. 2 flags are custom and 1 original, just saying
* One day someone will have to remove both CREATURE_FLAG_EXTRA_TRIGGER and CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT or change their implementation(I personally gonna do it for CREATURE_FLAG_EXTRA_TRIGGER) because they are custom and causes real problems, it will be a really long and hard road
* Let me explain why it's hard to remove that monstrosity of flags and replace by original flag. Since CREATURE_FLAG_EXTRA_TRIGGER does 10 things at once, whoever will do it will have to know what exactly that flag does and will have to check at least 3 times all creatures with that flag which is not easy. For example that flag adds Not Selectable flag which should never happen. It also changes react state to passive, there is only one flag in retail which can change react state to passive, actually it probably has nothing to do with react states and changes variable which is checked somewhere in code. So you want to remove that flag from some creatures. First you need to check for creatures which has no not selectable flag in DB right now, then you need to check sniffs of each creature without not selectable flag and apply it. Luckily we also can check CreatureDifficulty for that flag but we can't really trust it because it's from MoP. Then you need to check if creature should have original Ignore Combat flag(react state passive) and add it. Also we need take care of other things every time we want to remove that flag. It's how it works when we create custom flag and use it in 10 places of code
Oh right, it's also used in MagicSpellHitResult. Dissallows threat list. Changes faction and level if has summoner. Also UpdateTriggerVisibility. The amount of work to be done to remove that flag and handle all that stuff correctly is just bigger than to update DB side of 10 instances.
* I have a strong feeling based on original flags Blizz uses that there is no such thing as 'no combat' flag, I said that multiple times already
* Multiple creatures with CREATURE_FLAG_EXTRA_TRIGGER can enter in combat in retail. What this PR does is solves a problem by applying generic rule to tons of creatures at once. Never ends good, causes problems and is eventually reverted leaving tons of things broken. Happens all the time, nothing new.
* To solve #25628 someone will use custom attributes which will be applied to tons of triggered auras or will block completely combat from triggered auras. Just to make you understand, triggered auras is one of the reasons players stuck in combat with triggers. But instead of doing investigation, we want to restrict combat globally of all triggers no matter where they are used and no matter if they actually can enter in combat
* Current implementation of CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT can break spells which creatures cast on triggers
* It also can break encounters and simple creatures, nobody will check 1800 creatures. It's used not only in encounters but also in tons of quests and personally I don't know what to expect from that flag.
* I don't get move 'apply flag and then remove it by SetIsCombatDisallowed(false); in scripts if it's not needed', it's just not how it's supposed to work
* I'm sick of having to describe all that stuff 10 times

jackpoz:
What are the benefits of this PR ? I see mostly drawbacks.

jackpoz:
Just a heads up about PRs: please don't force-push, it makes the review harder. Thank you 👍

ccrs:
counter arguments:

@offl half the text just describes the problem that exists related to CREATURE_FLAG_EXTRA_TRIGGER, even if related indirectly, its not the focus of the PR.
also, in some way this is a minimal, tiny, step forward on that issue, since we are separating the combat prevention effect into another clear and really easy to see flag, but w/e.

> Multiple creatures with CREATURE_FLAG_EXTRA_TRIGGER can enter in combat in retail. What this PR does is solves a problem by applying generic rule to tons of creatures at once. Never ends good, causes problems and is eventually reverted leaving tons of things broken. Happens all the time, nothing new.

I already know this, I said it many times already, but I'll say it again, this wont be rushed, and it'll be thoroughly tested to ensure no rollbacks. Even if I have to run every single dungeon and raid.

> Multiple creatures with CREATURE_FLAG_EXTRA_TRIGGER can enter in combat in retail

give me an example of "trigger" creatures outside dungeons and raids that should be in combat and I'll adjust the PR.

> Just to make you understand, triggered auras is one of the reasons players stuck in combat with triggers

Leave the fucking "Just to make you understand" aside please. As I also have already said before, I know there is a problem there, but to state that triggered auras are the, dunnow, 95% of the origin of undesired combat with triggers, is a lie. The amount of "actions" that can cause combat is huge.

> But instead of doing investigation

ugh

> Current implementation of CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT can break spells which creatures cast on triggers

show me examples and I'll adjust the PR

> It also can break encounters and simple creatures, nobody will check 1800 creatures. It's used not only in encounters but also in tons of quests and personally I don't know what to expect from that flag.

prior the threat/combat system rewrite no trigger was allowed to enter combat, none, only those defined in scripts to do so. I expect very few edge cases, but we'll see.

> I don't get move 'apply flag and then remove it by SetIsCombatDisallowed(false); in scripts if it's not needed', it's just not how it's supposed to work

It's an extra flag, we define how it works, I don't get the message here "its not how is supposed to work".
The flag is added to the DB so that there are options to work with while developing.
Scripts right now use the direct call SetIsCombatDisallowed(false) as a fail safe. You see, a creature that has a script that is binded to a instance can be spawned manually outside the instance. The script wont load, the flag will still be there, happy outcome for the person who is toying with the spawns.

Again, if the consensus is that it should not be in the DB at all, I'll not include it.

@jackpoz 
> What are the benefits of this PR ? I see mostly drawbacks.
You should pretty much know it urself, since you did manual assignments to some of the triggers.

The situation is:
current combat system does not prevent explict or implicit combat to "trigger" creatures.
therefore, combat can be caused with a explict call to setincombatwith, like dozoneincombat, or with an implicit result of an action.
these actions can be spells, direct spells or auras, or even summoner to owner combat assist.
combatmgr does not rely on something like a "periodic update" that checks "should I be in combat?" each X update ticks.
it relies on external agents that caused the combat status to remove it, and I think it should be like that actually.

so, what can we do?

(this was all debated in IRC btw)
option 1: 
adjust triggerai to have an internal timer to remove combat each X seconds.
--> do all triggers get into combat and then remove it, dunnow, 5 seconds later? not to my knowledge

option 2:
adjust combatmgr to have a fail-safe implementation to prevent combat in specific situations if the owner is signaled as such?
but wait something ~very~ similar already exists right?

option 3:
use a determined set-up of the already existing CREATURE_FLAG_EXTRA_CANNOT_ENTER_COMBAT

option 4:
simply hardcode combat prevention to all triggers


> Just a heads up about PRs: please don't force-push, it makes the review harder. Thank you

exactly why?

offl:
>half the text just describes the problem that exists related to CREATURE_FLAG_EXTRA_TRIGGER, even if related indirectly, its not the focus of the PR.
also, in some way this is a minimal, tiny, step forward on that issue, since we are separating the combat prevention effect into another clear and really easy to see flag, but w/e.

2 flags to set react state to passive, 1 of them disallows threat list and 1 blocks combat completely. Does that sound good?

>I already know this, I said it many times already, but I'll say it again, this wont be rushed, and it'll be thoroughly tested to ensure no rollbacks. Even if I have to run every single dungeon and raid.

What's the point? You added that flag to 1800 creatures, we know 0 creatures should have that flag. How many creatures causes combat stuck? 10?

>give me an example of "trigger" creatures outside dungeons and raids that should be in combat and I'll adjust the PR.
>show me examples and I'll adjust the PR

Why should I do that and spent more time on that when I know that there is simply no such thing as flag which blocks combat nor 'disallow threat list'?

ccrs:
> How many creatures causes combat stuck? 10?

Much more sadly. Though only encounter triggers are reported.

> 2 flags to set react state to passive, 1 of them disallows threat list and 1 blocks combat completely. Does that sound good?

Yeah it does, a new react passive flag is the natural step for me too

> Why should I do that and spent more time on that when I know that there is simply no such thing as flag which blocks combat nor 'disallow threat list'?

One good example would actually be really helpful. This is not the ideal solution, aka the blizzlike way, but it is to me a nice step forward.

offl:
>One good example would actually be really helpful. This is not the ideal solution, aka the blizzlike way, but it is to me a nice step forward.

npc_fear_controller, we checked it few weeks ago

ccrs:
Sorry I did not explain myself correctly, each trigger that has a script assigned will be checked, the example I was talking about is a trigger that does not have any script, something that requires combat and simply relies on the default TriggerAI

Faq:
Why not make a log what specially logs those troublesome triggers, so in the end is knows exactly amount and id?

Aokromes:
> Why not make a log what specially logs those troublesome triggers, so in the end is knows exactly amount and id?

i asked jackpoz that yesterday xd i think it maybe cpu intensive.

jackpoz:
> Just a heads up about PRs: please don't force-push, it makes the review harder. Thank you

1. Harder to see the new changes since last update on github
2. All comments get marked as old as they reference commit hashes that don't exist anymore

offl:
Any example of trigger which causes combat stuck and is not a controller and has no periodic aura and located in instance?

jackpoz:
> [@ccrs](/ccrs) [ccrs](/ccrs) [force-pushed](/TrinityCore/TrinityCore/compare/8b42a0bef09b400057f7e29cb872c54e083c28ab..0e9bfe437dc44cc9d5be8fc968b6d6865ae9d056) the combat-trigger branch from [ `8b42a0b` ](/TrinityCore/TrinityCore/commit/8b42a0bef09b400057f7e29cb872c54e083c28ab) to [ `0e9bfe4` ](/TrinityCore/TrinityCore/commit/0e9bfe437dc44cc9d5be8fc968b6d6865ae9d056) [ 1 hour ago](#event-5260736746)

image


> > Just a heads up about PRs: please don't force-push, it makes the review harder. Thank you
> 
> 1. Harder to see the new changes since last update on github
> 2. All comments get marked as old as they reference commit hashes that don't exist anymore

Add to the list "people and bots get notified like if all the commits are new, while in fact they are not"

Shauren:
> Add to the list "people and bots get notified like if all the commits are new, while in fact they are not"

Just don't make branches for your own PR in TrinityCore/TrinityCore

jackpoz:
> > Add to the list "people and bots get notified like if all the commits are new, while in fact they are not"
> 
> Just don't make branches for your own PR in TrinityCore/TrinityCore

I agree with this, creating branches on TrinityCore IMO confuses users, especially if every developer created all their branches in TC. A draft PR from a personal fork is already a good way to share the progress in a public way

PR #26875, opened 2021-08-30
[Exploit] Do not allow to queue for rated arena without a group

robinsch:
**Changes proposed:**

-  Don't trust packet send from client

- [x] 3.3.5
- [ ] master

**Issues addressed:**
If you send CMSG_BATTLEMASTER_JOIN_ARENA with rated=1 and asgroup=0 you will be placed in rated arena queue without a group. (Client will display skirmish because of how we build SMSG_BATTLEFIELD_STATUS but in fact you end up in rated arena queue container.

**Tests performed:**
tested in-game

### DIFF
--- a/src/server/game/Handlers/BattleGroundHandler.cpp
+++ b/src/server/game/Handlers/BattleGroundHandler.cpp
@@ -620,6 +620,10 @@ void WorldSession::HandleBattlemasterJoinArena(WorldPacket& recvData)
 
     recvData >> guid >> arenaslot >> asGroup >> isRated;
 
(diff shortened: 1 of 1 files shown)


jackpoz:
does this work still with ".debug arena" ?

robinsch:
No logic is changed.

Keader:
> does this work still with ".debug arena" ? not that we have to keep an exploit if an existing command gets affected but more for information on the behavior

.debug arena do not affect rated arenas (if i remember right)

jackpoz:
Thanks for the PR 🚀 😃

PR #26883, opened 2021-09-02
Scripts/Creature: Update Dancing Flames

offl:
**Changes proposed:**

```cpp
            WorldPacket data;                       //send update position to client
            me->BuildHeartBeatMsg(&data);
            me->SendMessageToSet(&data, true);
```
-  creature should be summoned on top of object she summons after spawn. That ancient stuff above makes creature behave weirdly(it starts running somewhere) and I think we should not call something like that. However I'm not sure if calling Relocate is enough because in other scripts after relocating creature it forced also to move to new position. I had high hopes on SetDest in summoning spell script but forgot that she also need to summon her brazier on the ground and not at her current position. And changing SetDest z to +1.05 in one summoning spell and z to -1.05 in other is more weird than changing it once in creature script. So... leave old Relocate
-  removed or replaced few things, for example she indeed changes orientation but not to player but to angle(sniff) and only if you /dance
-  previously used bool and timer were supposed to add cooldown to her emotes but it does not work(it's simply not used for anything). Leaved a bunch of comments about how she acts on retail. It looks like a tricky script but I think for Blizz it's something regular and we just miss something in scripting system. Tried to reproduce [that I saw in video](https://youtu.be/GudPapkz5-I?t=25) with new implementation but I think something is still missing. Anyways, better than before 

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

none

**Tests performed:**

Builds, tested in-game

none


src/server/scripts/World/npcs_special.cpp
```diff
@@ -299,101 +299,79 @@ class npc_chicken_cluck : public CreatureScript
...
-        }
-
-        bool Active;
-        uint32 CanIteract;
+        DoCastSelf(SPELL_SUMMON_BRAZIER, true);
+        DoCastSelf(SPELL_BRAZIER_DANCE, false);
+        me->SetUInt32Value(UNIT_NPC_EMOTESTATE, EMOTE_STATE_DANCE);
+        float x, y, z;
+        me->GetPosition(x, y, z);
+        me->Relocate(x, y, z + 1.05f);
```
ccrs, line 318: try updateposition

offl:
```cpp
        float x, y, z, o;
        me->GetPosition(x, y, z, o);
        me->UpdatePosition(x, y, z + 1.05f, o);
```
Works same way, but why it's used only in ancient scripts and new ones uses Relocate?

ccrs: updateposition is the legit way of changing a poisition, since it goes through all the "update position" related logic

relocate is a straight serverside position change. position is later on sent to client through other means

though maybe a specific situation does not require the full "update position" logic, relocate should be used knowing the somewhat bypass that it implies.

offl: Then Relocate sounds better here

PR #26895, opened 2021-09-08
Core/Vendors: Support for vendors with empty inventory list

meji46:
**Changes proposed:**

-  Added support for vendors with empty inventory list. Example: https://www.wowhead.com/npc=45407/ibdil-the-mender

- [ ] 3.3.5
- [x] master

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game

None

### DIFF
--- a/src/server/game/Handlers/ItemHandler.cpp
+++ b/src/server/game/Handlers/ItemHandler.cpp
@@ -708,6 +708,8 @@ void WorldSession::SendListInventory(ObjectGuid vendorGuid)
     // Resize vector to real size (some items can be skipped due to checks)
     packet.Items.resize(count);
 
+    packet.Reason = AsUnderlyingType(count ? VendorInventoryReason::None : VendorInventoryReason::Empty);
+
     SendPacket(packet.Write());
 }
 
--- a/src/server/game/Entities/Creature/Creature.h
+++ b/src/server/game/Entities/Creature/Creature.h
@@ -54,6 +54,12 @@ typedef std::list<VendorItemCount> VendorItemCounts;
 
 #define MAX_VENDOR_ITEMS 150                                // Limitation in 4.x.x item count in SMSG_LIST_INVENTORY
 
+enum class VendorInventoryReason : uint8
+{
+    None    = 0,
+    Empty   = 1
+};
+
 static constexpr uint8 WILD_BATTLE_PET_DEFAULT_LEVEL = 1;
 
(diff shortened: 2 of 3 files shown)


src/server/game/Handlers/ItemHandler.cpp
```diff
@@ -702,6 +702,10 @@ void WorldSession::SendListInventory(ObjectGuid vendorGuid)
     // Resize vector to real size (some items can be skipped due to checks)
     packet.Items.resize(count);
 
+    // Repair-only NPC
+    if (!count && vendor->HasNpcFlag(UNIT_NPC_FLAG_REPAIR))
+        packet.Reason = 1;
```
Ovahlord, line 707: I'm not sure if this applies only for repair vendors. Guild Vendors should also show an empty window when you're not in a guild or if you do not have unlocked guild rewards yet.

meji46:
```
ServerToClient: SMSG_VENDOR_INVENTORY (0x25B4) Length: 17 ConnIdx: 1 Time: 09/09/2021 15:46:17.141 Number: 1896
VendorGUID: Full: 0x20425C00003249C00000610000380FB6 Creature/0 R4247/S97 Map: 0 (Reinos del Este) Entry: 51495 (Steeg Haskell) Low: 3674038
Reason: 1
VendorItems: 0
```

You are right. I guess then it will be enough to check that count is 0.

Ovahlord: So now it's the time where we might need a reverse engineering guy who can suggest names for that reason value as magic numbers are bad.

Ovahlord: So yeah. Apparently this field is being used for all cases where a vendor window is needed while no items are available. So the check for  GOSSIP_OPTION_VENDOR can be removed entirely imo.

Ovahlord: https://git.io/Juzp3 that's my take on that

meji46: Are there no other possible values for this field?

meji46: Perfect. Then I will use your same constexpr definition and update the pull request

meji46: Done. Thanks for your help 😄

PR #26908, opened 2021-09-13
Flood Control Addon Messages

stoneharry:
**Changes proposed:**
-  Flood control addon messages (anti-spam / rate limit)

- [x] 3.3.5
- [ ] master

**Issues addressed:**
n/a

**Tests performed:**

I opened up a project I am working on for a public test a few months ago and encountered people able to lock-up/freeze the server by spamming addon messages. I am using addon messages as a transport layer to send data to the client and receive data back, but these are only small lightweight requests.

I don't know why addon messages are not flood protected in the first place, is there a good historical reason for this?

Here is me testing locally, with no changes:
BeforeFloodProtection

Now if I make the change this pull request proposes:
AfterFloodProtection

Even better, if I enable `PacketSpoof.Policy = 1` and set the WorldSession maxPacketCounterAllowed for AddonMessages to 75, then we get this because more than 75 addon messages are sent from the player in a second and they are kicked:
AfterMaxPacketCounter

Again, no idea what the historical reasons are behind not rate limiting addon messages. To me I feel like they should be treated the same as any other chat message.


Ovahlord:
I feel like it would be enough to just add the cmsg_addon_message opcode to the ddos protection handling which kicks players immediately after sending too many opcodes of that type per second.

Riztazz:
> I feel like it would be enough to just add the cmsg_addon_message opcode to the ddos protection handling which kicks players immediately after sending too many opcodes of that type per second.

Addon messages are simple chat messages with lang addon though

Also there are some addons that do some heavy talking between each other, iirc carbonite does a lot via addon channels

jackpoz:
what about https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Server/WorldSession.cpp#L419 ? isn't that supposed to process only 100 packets from each player for each session ?

isn't your change going to show to the player some error/warning message while being clueless about why (as addon users are not addon developers) ?

stoneharry:
> what about https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Server/WorldSession.cpp#L419 ? isn't that supposed to process only 100 packets from each player for each session ?
> 
> isn't your change going to show to the player some error/warning message while being clueless about why (as addon users are not addon developers) ?

Yes, and I think this behaviour would be good actually. Hear me out:
- Although a lot of addon messages can be sent if you have a lot of addons, I think it would be hard to hit this rate limiter through purely legitimate messages, but I have no evidence to backup this up.
- If a player has been rate limited, I think it would be better to notify them that they have been muted. That way it's very obvious that something is going wrong, and they can start to ask for help or triage it themselves - maybe try turning some addons off.

An alternative approach might be to go with something like Riztazz suggested where we try to set the limit based on the number of addons that they have, but since we have no guarantee what the best case/worst case legitimate addon will be this feels to me like we are overcomplicating the solution.

Riztazz:
You can use https://github.com/tripleslash/wowscout to queue 100 packets and send them at once

HelloKitty:
I agree that rate limiting CMSG_MESSAGECHAT will break blizzlike 3.3.5 addons. Instead I suggest you enqueue these chat requests and handle them at the rate you would prefer. I think Blizzlike emulation should not worry about the performance implications of running Eluna on a custom server and using it for to create psuedo-gameplay packets.

stoneharry:
> I agree that rate limiting CMSG_MESSAGECHAT will break blizzlike 3.3.5 addons. Instead I suggest you enqueue these chat requests and handle them at the rate you would prefer. I think Blizzlike emulation should not worry about the performance implications of running Eluna on a custom server and using it for to create psuedo-gameplay packets.

Do you have any evidence that it will break 3.3.5 blizzlike addons?

Also this can crash the server even if you have zero custom processing on the addon messages. Perhaps config can fix this rather than a code change, but I have not had time to re-measure this yet.

Takenbacon:
Something that should be mentioned is there should be a limit on incoming data to the server. I am unsure if the blizzlike approach is a socket limit (e.g. fixed size read buffer) or something more targeted. See [https://wowwiki-archive.fandom.com/wiki/ChatThrottleLib](url)

I have literally seen first hand people create WoW client botnets to successfully DDoS players via chat message addon.

Riztazz:
There is also an issue of remote control via chat messages[1], while what i'll describe is out of scope of this PR, if you were to implement a thread for chat (on top of whatever limiters there should be), it would open a way to fix it without big performance hit

[1] https://www.mmorpg.com/news/new-chat-exploit-making-the-rounds-2000090767
Works for 3.3.5, although with different function names, RemoveExtraSpaces doesn't exist in 3.3.5a client but other functions can be replaced

jackpoz:
@Takenbacon https://github.com/TrinityCore/TrinityCore/pull/26908#issuecomment-918935980 is supposed to do exactly what you said, processing only 100 packets at every world update from a single player

Riztazz:
He meant a botnet, multiple sources targeting single player with addon messages

stoneharry:
I suspect that the 100 packets per update is working but the 10 second sampling of the metrics means that we do not see this in the graph. Regardless, you shouldn't be able to queue up millions of packets from a single player in second(s).

There is some evidence that Blizzard also throttles addon messages here: https://wowwiki-archive.fandom.com/wiki/ChatThrottleLib

I still plan on re-testing this over the weekend and getting some better analysis and the break-down of how I measured this. Please bear with me.

Jinnaix:
It's a fact that there have been interface addons in the past being sold for the sole purpose of either nuking specific player clients or whole servers. An open minded discussion with some facts and measurements is very much welcome.

Takenbacon:
@jackpoz It takes a lot less to drop a client than you'd think. I don't think the client its self is capable of handling a significant amount of incoming packets, even if the network had the bandwidth to do so but I'm not completely sure as I've never tested.

If we take your 100 packets per update for example, let's assume the server ticks every 25ms and someone has 25 characters (could be many more) "attacking" a player, that would result in approx. 100,000 packets per second sent to a player (which is also a SIGNIFICANT amount of data as well, btw, if we roughly assume we are sending ~300 bytes per packet with a max length chat message + packet data overhead + wow protocol overhead, not even including tcp headers and such)

The examples I have of a successful attack was "only" sending ~15,000 packets per second.

jackpoz:
You know, even if we process 10 packets per update, you can still use 250 players to send packets ( https://github.com/jackpoz/BotFarm can easily spawn 1000 bots on an average desktop) and achieve the same result.

Riztazz:
What about extending current dos protection to track packets over time?

stoneharry:
I ran a very basic test using a fork of 3.3.5 TC - spam a macro in a client for about ~5 seconds: `/script for i=0,1000 do SendClientRequest("feaihf", 5, 1); end` which constructs and sends addon messages with random contents.

Metrics

The server takes ~2 minutes to finish processing all of the packets that are queued up at an average of 1.6k packets per metric sample.

The code for adding this metric can be found here: https://github.com/stoneharry/TrinityCore/commit/a0629fc704878378afcbbf00e9bd83b44a40f8f8

Now is this acceptable? Can it be fixed through config? Those are both questions I'll leave you with to decide what you want to do. I'm happy to convert this into an issue and abandon this PR if you want to continue the discussion there. For my use-cases I have fixed it with the changes described in the original post.

I noticed that there is a newish Warden module that sends addon messages to feed the results back to the server - it would be interesting to target this code path and see if the impact is greater there.

jackpoz:
@stoneharry out of curiosity what is the raw value in the addon_messages table (without any group by) ? Is it below 100 ?

stoneharry:
Metrics

That is running:
`SELECT sum("value") FROM "addon_messages" WHERE ("realm" = 'Development') AND $timeFilter GROUP BY time($__interval) fill(null)`

Takenbacon:
> 
> 
> You know, even if we process 10 packets per update, you can still use 250 players to send packets ( https://github.com/jackpoz/BotFarm can easily spawn 1000 bots on an average desktop) and achieve the same result.

I am well aware how scaling works, thank you. It's a problem regardless.

stoneharry:
...and without any group by (sorry for the delay, not that used to Grafana):

Metrics

HelloKitty:
@stoneharry 

> > I agree that rate limiting CMSG_MESSAGECHAT will break blizzlike 3.3.5 addons. Instead I suggest you enqueue these chat requests and handle them at the rate you would prefer. I think Blizzlike emulation should not worry about the performance implications of running Eluna on a custom server and using it for to create psuedo-gameplay packets.
> 
> Do you have any evidence that it will break 3.3.5 blizzlike addons? From my testing thus far it all seems fine.

Most addons take a fire and forget approach, it's going to hamper their ability to communicate.

> 
> Also this can freeze the server even if you have zero custom processing on the addon messages. Perhaps config can fix this rather than a code change, but I have not had time to re-measure this yet.

I just don't understand this, all WorldSessions are clamped to 100 packets per update tick. No server, even in debug, should crawl to a halt under this circumstance.

> 
> I don't really see how Eluna comes into play here. I can disable the chat hook to prove that it doesn't make a difference if you would like, but that will be at the weekend most likely. The addon messages I sent were just wrapped in a client side Lua script for loop macro, and did not have any meaning or use. Pretend I didn't mention the word custom in this thread, that was just background for how I discovered the issue.

You're right, mentioning Eluna was presumptious but it's based on the what types of servers you seem to be developing. Surely you can forgive me for assuming based on your latest custom WoW server videos.


@Takenbacon 

> Something that should be mentioned is there should be a limit on incoming data to the server. I am unsure if the blizzlike approach is a socket limit (e.g. fixed size read buffer) or something more targeted. See https://wowwiki-archive.fandom.com/wiki/ChatThrottleLib
> 
> I have literally seen first hand people create WoW client botnets to successfully DDoS players via chat message addon.

Is this a specific addon message targeting a specific addon to DDoS players or is this just a bunch of messages with ADDON language sent to the target player?

The client seems quite capable of handling a significant amount of packets (but maybe not individually) which I once verified by packaging up all broadcasted set packets into one mega packet for continents and sending it to players with like 3,000 players in a map. This handling of 10,000s of packets a Map tick did not seem to disconnect players but did contribute to some reduction in frame rate. A fun lil expirment. Chat messages though are definitely going to raise a lot of LUA Signals/Events, and many addons may attempt to parse them so this opcode as you say may be a significant issue when it comes to target attacks. But I wonder if they are targeting a specific addon?

Anyway, as a treat the fun little custom opcode used to batch up all Broadcast Set packets into one massive packet pet tick:

 
image


Anyway, while I agree with @Takenbacon that this is an issue when considering server->client in the form of DDoS attacks I really am not convinced a TrinityCore 3.3.5 should suffer any significant performance issues when handling CMSG_MESSAGECHAT. I mean, this opcode can even be brought to map threads with alittle bit of locking for Channels and consideration for afew other things. It's not a performance issue on a server with 5,000+ people talking and with a significant amount of addon messages to boot! Something else is probably wrong if your server is lagging due to CMSG_MESSAGECHAT.

stoneharry:
> @stoneharry
> 
> > > I agree that rate limiting CMSG_MESSAGECHAT will break blizzlike 3.3.5 addons. Instead I suggest you enqueue these chat requests and handle them at the rate you would prefer. I think Blizzlike emulation should not worry about the performance implications of running Eluna on a custom server and using it for to create psuedo-gameplay packets.
> > 
> > 
> > Do you have any evidence that it will break 3.3.5 blizzlike addons? From my testing thus far it all seems fine.
> 
> Most addons take a fire and forget approach, it's going to hamper their ability to communicate.

They absolutely do take a fire and forget approach, but I still do not believe they will hit the thresholds proposed here. Blizzard certainly limits the amount of addon messages that can be sent before disconnecting you, hence the throttling library I linked to.

> > Also this can freeze the server even if you have zero custom processing on the addon messages. Perhaps config can fix this rather than a code change, but I have not had time to re-measure this yet.
> 
> I just don't understand this, all WorldSessions are clamped to 100 packets per update tick. No server, even in debug, should crawl to a halt under this circumstance.

That might be the case, my testing and observations have been pretty rudimentary - but is it correct for a server to spend 2 minutes processing messages that a player can send from the client in 5 seconds? That really does not sound right to me. With the limit to the amount of packets it can process per update, it is going to take a while to dequeue but this is still filling up a buffer with a lot of data which could not be legitimate traffic. It would be interesting to do further tests targeting the Warden code path where it is parsing for patterns in the addon messages, and/or to use one of the bot-clients linked to scale up.

> > I don't really see how Eluna comes into play here. I can disable the chat hook to prove that it doesn't make a difference if you would like, but that will be at the weekend most likely. The addon messages I sent were just wrapped in a client side Lua script for loop macro, and did not have any meaning or use. Pretend I didn't mention the word custom in this thread, that was just background for how I discovered the issue.
> 
> You're right, mentioning Eluna was presumptious but it's based on the what types of servers you seem to be developing. Surely you can forgive me for assuming based on your latest custom WoW server videos.

You are correct, I do use Eluna. I'm just overly defensive from historic attitudes in the community.

> Anyway, while I agree with @Takenbacon that this is an issue when considering server->client in the form of DDoS attacks I really am not convinced a TrinityCore 3.3.5 should suffer any significant performance issues when handling CMSG_MESSAGECHAT. I mean, this opcode can even be brought to map threads with alittle bit of locking for Channels and consideration for afew other things. It's not a performance issue on a server with 5,000+ people talking and with a significant amount of addon messages to boot! Something else is probably wrong if your server is lagging due to CMSG_MESSAGECHAT.

I have not been able to reproduce the lagging that I initially saw when testing, more metrics and better testing methods would be needed to get a thorough analysis. But it still doesn't sound right to me that so many addon message packets can be queued up in the first place, it doesn't sound right on the face of it and probably with the right combination of factors can lead to some detrimental effect.

Let's put another spin on this - if a player is allowed to spam addon messages as much as they want because the server should be able to dequeue it at a reasonable rate without falling over, what's stopping people looking at what popular addons parse messages for and flooding players with those expensive messages? Scale this up and it is going to have an effect.

jackpoz:
@stoneharry isn't this PR still receiving the thousands of spammed addon messages which still end up in the opcode handler and still get throttled by 100 at a time ? I mean, the only way to avoid this is to stop the messages from even reaching worldserver, so either stopping that from the client or kicking the player

stoneharry:
> @stoneharry isn't this PR still receiving the thousands of spammed addon messages which still end up in the opcode handler and still get throttled by 100 at a time ? I mean, the only way to avoid this is to stop the messages from even reaching worldserver, so either stopping that from the client or kicking the player

Yes, I think we have definitely strayed away the exact code changes proposed by this PR as we have discussed this. I also changed the WorldSession maxPacketCounterAllowed which does cause the player to get kicked (although a higher threshold than 75 would probably be safer).

Changes

jackpoz:
By the way the reason it takes 2 minutes to process the packets it's a sign that the throttling works :) keeping the game playable for other players (and it can be lowered ofc).

As the antidos kicks based on packet count for each opcode, it would make sense to skip to next worldsession if there are too many of the same opcode.

P.s.: one legit guild owner took 2 minutes to login because he had too many addons (I think it was Prat with a nickname module enabled) and his worldsession was getting throttled. @Aokromes reported this issue.

HelloKitty:
One option, and something we did, was replace LockedQueue's internals with a higher performance lockless queue that also supports measuring the size of the queue. You could simply check the queue size if you did this and disconnect the WorldSession if they had some insane amount of queued up packets.

jackpoz:
> As the antidos kicks based on packet count for each opcode, it would make sense to skip to next worldsession if there are too many of the same opcode.

this is still worth to be implemented

PR #26911, opened 2021-09-14
Core/Misc: Improvement item durability repair logic.

### LINKED ISSUE #25937: guild bank adds one row per item repaired instead per repair
**Description:**

guild bank adds one row per item repaired instead per repair, making log quite useless because if 2 persons repairs it fills visible log.

**Expected behaviour:**

it must show one row per repair.

1. make one guild 
2. add some funds
3. break your gear
4. notice you see 1 row per repaired item

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

a06a755f4c52

**Operating system:** 

debian 10 x64.

r4d1sh:
**Changes proposed:**

* Move item repair cost calculation logic from Player::DurabilityRepair to Item::CalculateDurabilityRepairCost
* Make item repair cost calculation more congruent with client logic (std::round)
* Do not repair items if player does not have enough money to pay for the repair of all items at once
* Withdrawing money from guild bank only once to avoid spam in the guild bank log and reduce number of transactions in the database

- [x] 3.3.5
- [ ] master

**Issues addressed:**

Closes #25937

**Tests performed:**

Successfully builded and tested on Windows 10


src/server/game/Entities/Player/Player.cpp
```diff
@@ -4874,117 +4874,117 @@ void Player::DurabilityPointLossForEquipSlot(EquipmentSlots slot)
...
-uint32 Player::DurabilityRepairAll(bool cost, float discountMod, bool guildBank)
+void Player::DurabilityRepairAll(bool takeCost, float discountMod, bool guildBank)
 {
-    uint32 TotalCost = 0;
-    // equipped, backpack, bags itself
-    for (uint8 i = EQUIPMENT_SLOT_START; i < INVENTORY_SLOT_ITEM_END; i++)
-        TotalCost += DurabilityRepair(((INVENTORY_SLOT_BAG_0 << 8) | i), cost, discountMod, guildBank);
+    std::list<std::pair<Item*, uint32>> itemRepairCostStore;
+    // Сollecting all items that can be repaired and repair costs
+    {
```
jackpoz, line 4881: is this an additional local scope ?

src/server/game/Handlers/NPCHandler.cpp
```diff
@@ -734,7 +734,7 @@ void WorldSession::HandleRepairItemOpcode(WorldPacket& recvData)
 
         Item* item = _player->GetItemByGuid(itemGUID);
         if (item)
-            _player->DurabilityRepair(item->GetPos(), true, discountMod, guildBank != 0);
```
jackpoz, line 737: is guildBank ignored now in this case or ?

r4d1sh: > is guildBank ignored now in this case or ?

Yes. We can't repair one specific item with guild money, only for the player's own money. If specific item guid received we can ignore `guildBank` parameter

src/server/game/Entities/Player/Player.cpp
r4d1sh: > is this an additional local scope ?

Yes, just to highlight this separate piece of code. Technically, it is not necessary to do this.

jackpoz: We usually don't use scopes to group code in a single function but mostly only to trigger a destructor. Could you please remove the additional scope in this case :) ?

jackpoz:
The PR shows quite a lot of code as modified but most of it was just moved around. Could you maybe add some GitHub comments to highlight the actual changes you added, in particular in Player.cpp ? It would help the PR review and speed up the merge process

r4d1sh:
> The PR shows quite a lot of code as modified but most of it was just moved around. Could you maybe add some GitHub comments to highlight the actual changes you added, in particular in Player.cpp ? It would help the PR review and speed up the merge process

The changes are mainly made to `Player::DurabilityRepairAll` method. Changes to "Player::DurabilityRepair" are mainly moving the code to more appropriate places. It shouldn't be hard to compare before and after code if you look only at these two methods.

jackpoz:
Thanks for the PR 🚀 😃

Issue #26912, opened 2021-09-14
Core/Misc: unit_flags names


offl:
### Description

#25764 created [discussion](https://github.com/TrinityCore/TrinityCore/issues/25764#issuecomment-758059124) (thanks, @xvwyh) about original names of flags, decided to store all that info here to not lose it since #25764 can be closed anytime
Included wpp names since someone spent some time and added more proper names to some flags, I found some of them really helpful

Legend:
` -` current name is fine
` +` can be renamed
`?` to be discussed

Would be nice if more people dig into this, unfortunately I can't check where some of those flags are used in client and because of that some of them are marked as `?`
Some of them I'll rename in next few days, however in some cases new names should be approved first and in some cases I simply don't know what to do

```txt
flag          | 3.3.5 name                                            | wpp name                                       | client             | static flag    | comment
?  1          | UNIT_FLAG_SERVER_CONTROLLED     = 0x00000001,         | NotClientControlled              = 0x1,        | IsDisconnected ?   | -              | -
+  2          | UNIT_FLAG_NON_ATTACKABLE        = 0x00000002,         | PlayerCannotAttack               = 0x2,        | IsSpawning         | -              | UNIT_FLAG_SPAWNING, we simply can't rename it at this moment because that crap is used in tens of scripts, I'll open issue about how exactly that flag is used originally and leave a comment in code to rename that flag once it will be removed from all scripts
?  4          | UNIT_FLAG_REMOVE_CLIENT_CONTROL = 0x00000004,         | RemoveClientControl              = 0x4,        | IsClientLocked ?   | -              | -
-  8          | UNIT_FLAG_PLAYER_CONTROLLED     = 0x00000008,         | PlayerControlled                 = 0x8,        | IsPlayerControlled | -              | -
+  16         | UNIT_FLAG_RENAME                = 0x00000010,         | Unk4                             = 0x10,       | -                  | -              | UNIT_FLAG_EVADING, set when creature evades, removed when it reaches home. Checked multiple creatures, pets too. Not that I really know what it does, just wanna rename it because current name is confusing and pisses me off https://github.com/TrinityCore/TrinityCore/pull/24020/files#r363048708
?  32         | UNIT_FLAG_PREPARATION           = 0x00000020,         | Preparation                      = 0x20,       | -                  | -              | -
?  64         | UNIT_FLAG_UNK_6                 = 0x00000040,         | Unk6                             = 0x40,       | IsPlusMob ?        | -              | Read below
[...]

PR #26954, opened 2021-09-27
Core/SAI: Split set/remove unit_flags actions

offl:
**Changes proposed:**

-  Deprecate SMART_ACTION_SET_UNIT_FLAG & SMART_ACTION_REMOVE_UNIT_FLAG
-  Create new bool actions to set ImmunePC, ImmuneNPC, Uninteractible (former not selectable)

The plan was to use here SetImmuneTo and create SetUninteractible. However I noticed that SetImmuneTo has param keepInCombat which allows to keep creature in combat after making it immune to PC\NPC. Not sure but looks like only if creature has passive react state. Later I checked how it works without calling SetImmuneTo, by changing flags directly(how it always was in SAI). Turned out it doesn't matter how flag was set, creature will remain in combat if it's passive

I have no idea when it started but to me it looks like dependency created by people who think they need to make creature immune to PC\NPC during combat at all cost. From my experience in case of bosses Blizz makes them passive or uninteractible and passive. Probably that's why UNIT_FLAG_NON_ATTACKABLE is so popular in C++ scripts(200 cases, should be 0). Simply because it keeps creature in combat but at the same time makes creature non-attackable. Those scripts were not written from sniffs. Same with scripts which calls SetImmuneTo mid combat

On top of that:
- I have a feeling we should not use **pet react states** to make normal creature passive or defensive
- There's no such param in original actions. Even if system which uses those action is not used for really complicated bosses, it doesn't matter because it's really bad to have differently handled functions in different systems, especially if that function is used in thousands of scripts. They should be same no matter where they are used. We have huge difference between how core scripts and DB scripts works but on retail they all are same, no matter if it's boss or normal mob. The only difference between bosses and simple creatures is you can make bosses more complex but in game they acts in same way
- In https://github.com/TrinityCore/TrinityCore/issues/26912#issuecomment-922333781 I made an assumption that combat-related flags are messed up and their behavior is not entirely correct. You also can see how flags works

Pretty sure creature leaves combat instantly when flags are set, no matter if it's passive or not. I know it looks ugly but I don't want to use here SetImmuneTo because I have no idea if its effect is same or not. Maybe later I'll try to figure out if it can be used

**Target branch(es):**

- [x] 3.3.5
- [ ] master

**Issues addressed:**

none

**Tests performed:**

not tested, didn't tried to build

should be tested + sql

### DIFF
--- a/sql/updates/world/3.3.5/2021_09_27_00_world.sql
+++ b/sql/updates/world/3.3.5/2021_09_27_00_world.sql
@@ -1,0 +1,7 @@
+--
+UPDATE `smart_scripts` SET `action_type` = 144, `action_param1` = 1 WHERE `action_type` = 18 AND `action_param1` = 256;
+UPDATE `smart_scripts` SET `action_type` = 144, `action_param1` = 0 WHERE `action_type` = 19 AND `action_param1` = 256;
+UPDATE `smart_scripts` SET `action_type` = 145, `action_param1` = 1 WHERE `action_type` = 18 AND `action_param1` = 512;
+UPDATE `smart_scripts` SET `action_type` = 145, `action_param1` = 0 WHERE `action_type` = 19 AND `action_param1` = 512;
+UPDATE `smart_scripts` SET `action_type` = 146, `action_param1` = 1 WHERE `action_type` = 18 AND `action_param1` = 33554432;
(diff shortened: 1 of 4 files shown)


offl:
All currently deprecated action and event types will be removed after 2 weeks

PR #26974, opened 2021-09-30
Core/Characters: Added allied race creation data, implemented intro s…

mdX7:
…cenes & added config option to disable achievement requirements for allied races

**Changes proposed:**

-  Added playercreateinfo data for allied races
-  Added config option to disable required achievements for allied races
-  Implemented scene intros

**Target branch(es):** 3.3.5/master

- [ ] 3.3.5
- [x] master

**Issues addressed:**

Closes #  (insert issue tracker number)

**Tests performed:**

(Does it build, tested in-game, etc.)
Builds, tested ingame

**Depends on:** (add/remove lines as needed)
#26975 to fix blackscreens after skipping intro

PR #26975, opened 2021-09-30
Core/Scenes: defined flags and implemented some

mdX7:
**Changes proposed:**

-  defined scene flags
-  implemented some of them
-  

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

build, tested in combination with #26974

**Notes:** 
i'm assuming `SceneFlag::PlayAsLoginCinematic` is only used serverside and not sent to client, i cannot think of any scenes which are used on login besides the first login intros, which do not have the flag set. Handling the flag like it is and adding it to the scene_template in #26974 results in proper cancelling of the scene without being stuck in a black screen.


src/server/game/Globals/ObjectMgr.h
```diff
@@ -32,6 +32,7 @@
 #include "Position.h"
 #include "QuestDef.h"
 #include "RaceMask.h"
+#include "SceneMgr.h"
```
Shauren, line 35: Nope.

Shauren (changes requested):

PR #26977, opened 2021-10-01
Scripts/Commands: avoid crash on lookup spell id

Smerdokryl:
GetEffect assertion fails for spells with no linked SpellEffect entries.

**Changes proposed:**

-  Check GetEffects().size() before calling GetEffect

- [ ] 3.3.5
- [x] master

**Tests performed:**

Built, tested in game - works as expected from 'normal' spells

### DIFF
--- a/src/server/scripts/Commands/cs_lookup.cpp
+++ b/src/server/scripts/Commands/cs_lookup.cpp
@@ -959,10 +959,15 @@ public:
 
             bool known = target && target->HasSpell(id);
 
-            SpellEffectInfo const& spellEffectInfo = spellInfo->GetEffect(EFFECT_0);
-            bool learn = spellEffectInfo.IsEffect(SPELL_EFFECT_LEARN_SPELL);
+            bool learn = false;
+            SpellInfo const* learnSpellInfo;
+            if (spellInfo->GetEffects().size()) {
+                SpellEffectInfo const& spellEffectInfo = spellInfo->GetEffect(EFFECT_0);
+                learn = spellEffectInfo.IsEffect(SPELL_EFFECT_LEARN_SPELL);
 
-            SpellInfo const* learnSpellInfo = sSpellMgr->GetSpellInfo(spellEffectInfo.TriggerSpell, DIFFICULTY_NONE);
+                if (learn)
+                    learnSpellInfo = sSpellMgr->GetSpellInfo(spellEffectInfo.TriggerSpell, DIFFICULTY_NONE);
+            }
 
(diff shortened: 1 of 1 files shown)


mdX7:
Is this for custom spells only or do we have spells like that in db2 aswell? Never saw a spell without effects tbh

src/server/scripts/Commands/cs_lookup.cpp
```diff
@@ -959,18 +959,23 @@ class lookup_commandscript : public CommandScript
...
-            bool learn = spellEffectInfo.IsEffect(SPELL_EFFECT_LEARN_SPELL);
-
-            SpellInfo const* learnSpellInfo = sSpellMgr->GetSpellInfo(spellEffectInfo.TriggerSpell, DIFFICULTY_NONE);
+            bool learn = false;
+            SpellInfo const* learnSpellInfo;
+            if (spellInfo->GetEffects().size()) {
+                SpellEffectInfo const& spellEffectInfo = spellInfo->GetEffect(EFFECT_0);
+                learn = spellEffectInfo.IsEffect(SPELL_EFFECT_LEARN_SPELL);
+
+                if (learn)
```
mdX7, line 968: You can remove this variable and use IsEffect directly

src/server/scripts/Commands/cs_lookup.cpp
```diff
@@ -959,18 +959,23 @@ class lookup_commandscript : public CommandScript
 
             bool known = target && target->HasSpell(id);
 
-            SpellEffectInfo const& spellEffectInfo = spellInfo->GetEffect(EFFECT_0);
-            bool learn = spellEffectInfo.IsEffect(SPELL_EFFECT_LEARN_SPELL);
-
-            SpellInfo const* learnSpellInfo = sSpellMgr->GetSpellInfo(spellEffectInfo.TriggerSpell, DIFFICULTY_NONE);
+            bool learn = false;
+            SpellInfo const* learnSpellInfo;
+            if (spellInfo->GetEffects().size()) {
```
mdX7, line 964: `{` always on next line according to our guidelines

Lordron: and i think !empty is more preferable

src/server/scripts/Commands/cs_lookup.cpp
Smerdokryl: `learn` is used again on line 995/1000

Shauren:
Thank you for your contribution but I will not be merging it

* Doesn't fix the other command also crashing
* Possibly causes a far worse crash due to uninitialized pointer
* Doesn't fix the command logic always assuming learn effect is in first slot

PR #26991, opened 2021-10-01
Scripts/Commands: avoid crash on lookup spell (id)

Smerdokryl:
GetEffect assertion fails for spells with no linked SpellEffect entries.

**Changes proposed:**

-  Loop over effects instead of getting only EFFECT_0
-  Ensure (implicitly via the above) that the spell has effects

- [ ] 3.3.5
- [x] master

**Tests performed:**

Built, tested in game - works as expected from 'normal' spells

**Potential Improvements:**

- The two lookup spell methods are mostly the same. Perhaps they should be merged or remade into wrappers of a common method?

### DIFF
--- a/src/server/scripts/Commands/cs_lookup.cpp
+++ b/src/server/scripts/Commands/cs_lookup.cpp
@@ -959,10 +959,15 @@ public:
 
             bool known = target && target->HasSpell(id);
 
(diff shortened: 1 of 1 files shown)


Smerdokryl:
:eyes: :neutral_face:

Too little too late, huh!

PR #27021, opened 2021-10-04
Core/Conditions: implemented conditions for areatriggers

mdX7:
* also added example fix for Death's Power Grows quest

**Changes proposed:**

-  implement conditions for areatriggers
-  fix Death's Power Grows quest

- [ ] 3.3.5
- [x] master

**Issues addressed:**

**Tests performed:**

builds, tested ingame


src/server/game/Conditions/ConditionMgr.h
```diff
@@ -286,6 +288,7 @@ class TC_GAME_API ConditionMgr
         bool IsObjectMeetingVehicleSpellConditions(uint32 creatureId, uint32 spellId, Player* player, Unit* vehicle) const;
         bool IsObjectMeetingSmartEventConditions(int64 entryOrGuid, uint32 eventId, uint32 sourceType, Unit* unit, WorldObject* baseObject) const;
         bool IsObjectMeetingVendorItemConditions(uint32 creatureId, uint32 itemId, Player* player, Creature* vendor) const;
+        bool IsObjectMeetingAreaTriggerConditions(std::pair<uint32, bool> const& areaTriggerId, WorldObject* object) const;
```
mdX7, line 291: personally i kinda dont like std::pair here since we could use AreaTriggerId directly, but that either requires us to include AreaTriggerTemplate.h in ConditionMgr (which feels way more worse) or move AreaTriggerId to a new file, which feels weird. Or we could use a pointer and just deref it inside + forward declare in ConditionMgr.h, but that also feels weird.

Shauren: But the rule to allow forward declaring is not "use pointer", its "don't use value"

src/server/game/Conditions/ConditionMgr.cpp
```diff
@@ -1048,6 +1053,17 @@ bool ConditionMgr::IsObjectMeetingVendorItemConditions(uint32 creatureId, uint32
     return true;
 }
 
+bool ConditionMgr::IsObjectMeetingAreaTriggerConditions(uint32 areaTriggerId, bool isServerSide, WorldObject* object) const
+{
+    ConditionEntriesByAreaTriggerIdMap::const_iterator itr = AreaTriggerConditionContainerStore.find({ areaTriggerId, isServerSide });
```
Shauren, line 1058: This is exactly where you DO use `auto`, or preferably `Trinity::Containers::MapGetValuePtr` since this is a map lookup

src/server/game/Conditions/ConditionMgr.cpp
```diff
@@ -1284,6 +1300,13 @@ void ConditionMgr::LoadConditions(bool isReload)
                 case CONDITION_SOURCE_TYPE_PHASE:
                     valid = addToPhases(cond);
                     break;
+                case CONDITION_SOURCE_TYPE_AREATRIGGER:
+                {
+                    AreaTriggerConditionContainerStore[{ cond->SourceGroup, cond->SourceEntry}].push_back(cond);
```
Shauren, line 1305:
```suggestion
                    AreaTriggerConditionContainerStore[{ cond->SourceGroup, cond->SourceEntry }].push_back(cond);
```

src/server/game/Conditions/ConditionMgr.cpp
```diff
@@ -1886,6 +1909,13 @@ bool ConditionMgr::isSourceTypeValid(Condition* cond) const
                 return false;
             }
             break;
+        case CONDITION_SOURCE_TYPE_AREATRIGGER:
+            if (!sAreaTriggerDataStore->GetAreaTriggerTemplate({ uint32(cond->SourceGroup), bool(cond->SourceEntry) }))
```
Shauren, line 1913: Validate that `cond->SourceEntry` is exactly `0` or `1` too

CraftedRO:
this migt be one day cherry picked to 3.3.5

PR #27029, opened 2021-10-07
Core/SAI: added new action type SMART_ACTION_ADD_TO_STORED_TARGET_LIST

mdX7:
**Changes proposed:**

-  added new action type SMART_ACTION_ADD_TO_STORED_TARGET_LIST
-  
-  

- [ ] 3.3.5
- [x] master

**Tests performed:**

build, tested in combination with #27027 and:
```sql
DELETE FROM `conversation_line_template` WHERE `Id` IN (32915, 32916, 32917, 32918, 32919, 32926);
INSERT INTO `conversation_line_template` (`Id`, `StartTime`, `UiCameraID`, `ActorIdx`, `Flags`, `VerifiedBuild`) VALUES
(32915, 0, 0, 0, 0, 40120),
(32916, 9878, 0, 0, 0, 40120),
(32917, 22389, 0, 0, 0, 40120),
(32918, 33855, 0, 0, 0, 40120),
(32919, 43842, 0, 0, 0, 40120),
(32926, 55409, 0, 1, 1, 40120);

DELETE FROM `conversation_template` WHERE `Id`=13254;
INSERT INTO `conversation_template` (`Id`, `FirstLineID`, `LastLineEndTime`, `TextureKitId`, `VerifiedBuild`) VALUES
(13254, 32915, 55409, 0, 40120);

UPDATE `creature_template` SET `AIName`='SmartAI' WHERE `entry`=161709;
DELETE FROM `smart_scripts` WHERE `entryorguid`=161709 AND `source_type`=0;
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `event_type`, `event_param1`, `action_type`, `action_param1`, `action_param2`, `target_type`, `target_param1`, `comment`) VALUES 
(161709, 0, 0, 62, 25082, 134, 316983, 0, 1, 0, 'Darion Mograine - On Gossip Option Select - Invoker Cast - Self'),
(161709, 0, 1, 62, 25082, 144, 0, 0, 1, 0, 'Darion Mograine - On Gossip Option Select - Add to Stored Target List - Self'),
(161709, 0, 2, 62, 25082, 144, 0, 0, 7, 0, 'Darion Mograine - On Gossip Option Select - Add to Stored Target List - Invoker'),
(161709, 0, 3, 62, 25082, 143, 13254, 1, 12, 0, 'Darion Mograine - On Gossip Option Select - Create Conversation - Stored Targets'),
(161709, 0, 4, 62, 25082, 72, 0, 0, 0, 0, 'Darion Mograine - On Gossip Option Select - Close Gossip');
```


BAndysc:
Instead of `Add to Stored Target List - Self'` the first "add" should be `Store Target List: Self`, so that the stored list is always cleared on gossip. SAI lines will always be executed sequentially, so it is not a problem when two players select gossip at the same time (which is btw. very unlikely)

mdX7:
> Instead of Add to Stored Target List - Self' the first "add" should be Store Target List: Self, so that the stored list is always cleared on gossip
Yep realized that aswell, but this is just a test anyway :P 
> so it is not a problem when two players select gossip at the same time (which is btw. very unlikely)
Sounds good then

PR #27043, opened 2021-10-11
Core/Scripting: Implement Script Reloading

Seyden:
**Changes proposed:**

-  Implementing the way to be able to reload new scripts

- [x] master

**Issues addressed:**

**Tests performed:**

- Builds
- Tested with Creature_template scripts
- Tested with Spell_script_names
- Tested with AreaTrigger_template Serverside Areatriggers

- [ ] 
- [ ]


acidmanifesto:
> I would like to bash this for all the people out there with no internet.

i got free wifi at mcdonalds.

Naios (changes requested):
Overall your PR is a good point to start, however I would change a few things:

- Currently it doesnt seem that your PR supports removing scriptnames from the database such that a scriptname delete results in unloading the scriptname form entities.
- The scriptname datastructure should be changed to a map rather than an unsorted vector to allow removing elements in the middle. Currently the vector assumes that script ids are assigned in alphabetical sorted order and this is no longer the case.

Shauren:
Id assignments inside AI classes are wrong for creatures and gameobjects. AI classes can be created using AIName database field directly or dynamically by a core mechanic (charm/mind control) or even based on creature type (pets) - in these cases current code will assign incorrect script id to them

Naios:
@Shauren Thanks for your hint. Therefore, I corrected the previous behaviour by propagating the name of the AIs created by core mechanic to the script. Does this look right to you?

Shauren (changes requested):

Naios (approved):

PR #27057, opened 2021-10-13
Core/BattlePets: Misc fixes

meji46:
**Changes proposed:**

-  Rename WorldPackets::BattlePet::BattlePet::CollarID to WorldPackets::BattlePet::BattlePet::DisplayID.
-  Use the DisplayID field to store the model of the battle pet. If the species has BattlePetSpeciesFlags::RandomDisplay, nothing is stored. Otherwise a random model is chosen from those available.
- Stored DisplayID in ITEM_MODIFIER_BATTLE_PET_DISPLAY_ID when caging the battle pet (previously the CreatureID was being stored).
- Modified SendPlaySpellVisual function to allow sending Target and TargetPosition at the same time.
- Added SpellVisual when uncaging.
- Set NoRename field to false. I've never seen it set to true and the current value definition is incorrect.

- [ ] 3.3.5
- [x] master

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

- I haven't found the use of the CollarID field in battle pet slots.


Shauren:
No, that one actually means collar, whatever it is supposed to be

Aokromes:
Conflicting files
src/server/database/Database/Implementation/LoginDatabase.cpp

PR #27178, opened 2021-10-25
Core/BattlePets: Journal lock improvements

meji46:
**Changes proposed:**

The journal lock restricts access to the journal functions to the character of an account (associated with the battlenet account) that enters the world. If a character from a second account (also associated with the battlenet account) enters the world, he will have most of the journal functions blocked.
- Enabled opcode SMSG_BATTLE_PET_JOURNAL_LOCK_DENIED
- Enabled opcode CMSG_BATTLE_PET_REQUEST_JOURNAL_LOCK
- Function added to check if any account of the same bnet account has acquired the journal lock (IsJournalLockAcquired).
- Function to acquire/release the journal lock. (ToggleJournalLock)
- Added function to try to acquire the lock and send the correct opcode (SendJournalLockStatus) (e.g. enter the world, open the journal, start a pet battle)
- WorldPackets::BattlePet::BattlePetJournal.HasJournalLock is now initialized with the value returned by HasJournalLock
- Added HasJournalLock checks in several functions (client locks these actions without acquired journal lock)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


Shauren:
If this is supposed to be a part of future work on blocking access to pet journal from two accounts at the same time then it is not going to work. Each WorldSession gets its own BattlePetMgr - and each player gets his separate WorldSession (so two accounts from the same bnet account logged in will have two separate BattlePetMgrs, this lock will not work)

Shauren (changes requested):
If this is supposed to be a part of future work on blocking access to pet journal from two accounts at the same time then it is not going to work. Each WorldSession gets its own BattlePetMgr - and each player gets his separate WorldSession (so two accounts from the same bnet account logged in will have two separate BattlePetMgrs, this lock will not work)

meji46:
> If this is supposed to be a part of future work on blocking access to pet journal from two accounts at the same time then it is not going to work. Each WorldSession gets its own BattlePetMgr - and each player gets his separate WorldSession (so two accounts from the same bnet account logged in will have two separate BattlePetMgrs, this lock will not work)

I've mainly tried to mirror the logic of the journal lock in retail.

Regarding the point of making all WorldSession of a bnet account have the same journal lock information, I don't know exactly what would be the best solution. I had thought of temporarily storing the player's guid who has the lock in db but, a second player (same bnet account) who doesn't have the lock, has to constantly check if the lock is free or not.

PR #27189, opened 2021-10-27
Scripts/Uldaman: Fix Ironaya crash

Jildor:
**Changes proposed:**

-  Fix Ironaya crash when ResetThreat: https://github.com/TrinityCore/TrinityCore/blob/ed939325374216d6a049e778d6ad864c65283ae6/src/server/scripts/EasternKingdoms/Uldaman/boss_ironaya.cpp#L76-L84
SPELL_KNOCKAWAY (ID - 10101 Knock Away) has SPELL_EFFECT_WEAPON_DAMAGE, so can kill the victim, in that case, when ResetThreat do EnsureVictim and ASSERT is triggered:
https://github.com/TrinityCore/TrinityCore/blob/1403de8dafbadb92dfd16c47b63156f9216a2bdd/src/server/game/Entities/Unit/Unit.h#L849-L854
-  Little code cleanup, because is ancient script and I prefer check HealthBelowPct inside DamageTaken instead every UpdateAI tick.
I don't want do a fully instance rewrite, the focus of this PR is fix the crash, anyone can do rewrite later

**Issues addressed:**

Closes: None, maybe .die Treeston again :D (https://github.com/TrinityCore/TrinityCore/commit/e2a1ccd118d129b96e09ff1a15ed0adb1d4a3897#diff-096cfd16b5b45a8321f215ef98d7cf24dd99013dec4983ef2db15b86aa140e8c)

**Tests performed:**

Tested in-game

### DIFF
--- a/src/server/scripts/EasternKingdoms/Uldaman/boss_ironaya.cpp
+++ b/src/server/scripts/EasternKingdoms/Uldaman/boss_ironaya.cpp
@@ -33,82 +33,70 @@ enum Ironaya
     SPELL_WSTOMP                = 11876
 };
 
-class boss_ironaya : public CreatureScript
+struct boss_ironaya : public ScriptedAI
 {
-    public:
+    boss_ironaya(Creature* creature) : ScriptedAI(creature)
+    {
+        Initialize();
+    }
 
-        boss_ironaya()
-            : CreatureScript("boss_ironaya")
+    void Initialize()
+    {
+        _hasCastKnockaway = false;
+        _hasCastWstomp = false;
+    }
+
+    void Reset() override
+    {
+        _scheduler.CancelAll();
+        Initialize();
+    }
+
+    void DamageTaken(Unit* /*attacker*/, uint32& damage, DamageEffectType /*damageType*/, SpellInfo const* /*spellInfo = nullptr*/) override
+    {
+        if (!_hasCastKnockaway && HealthBelowPct(50) && me->GetVictim())
         {
+            _hasCastKnockaway = true;
+            DoCastVictim(SPELL_KNOCKAWAY, true);
+            ResetThreat(me->GetVictim(), me);
         }
 
-        struct boss_ironayaAI : public ScriptedAI
+        if (!_hasCastWstomp && HealthBelowPct(25))
         {
-            boss_ironayaAI(Creature* creature) : ScriptedAI(creature)
-            {
-                Initialize();
-            }
-
-            void Initialize()
-            {
-                uiArcingTimer = 3000;
-                bHasCastKnockaway = false;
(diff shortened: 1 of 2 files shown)


Jildor:
> > I prefer check HealthBelowPct inside DamageTaken instead every UpdateAI tick.
> 
> It's not that bad because otherwise every action inside DamageTaken ignores checks from UpdateAI(primarily casting state) And even if we add them to DamageTaken, it will not work as in retail because you'll have to damage creature again to again trigger DamageTaken
> 
> On retail creature can't switch to another phase if it currently casts a spell, on TC they can because transitions are handled in DamageTaken. Good example is Kil'jaeden, he changes phase 5 times and each time he casts a 30s spell. And if you damage him too fast, all he will do is casting that spell 5 times in a row

I do not agree with what you say, checking that in UpdateAI seems to me to excessive checks every Tick.
About what you say that in DamageTaken it ignores if a boss is casting and enters phase earlier than it should, you simply can launch an Event in DamageTaken, so that event will not run until it finishes casting, for example:
```c++
    void DamageTaken(Unit* /*who*/, uint32& /*damage*/, DamageEffectType /*damageType*/, SpellInfo const* /*spellInfo = nullptr*/) override
    {
        if (me->HealthBelowPct(90) && !_checkPhase)
        {
            _checkPhase = true;
            events.ScheduleEvent(EVENT_CHANGE_PHASE, 1ms);
        }
    }

    {
        if (!UpdateVictim())
            return;

        events.Update(diff);

        if (me->HasUnitState(UNIT_STATE_CASTING))
            return;

        while (uint32 eventId = events.ExecuteEvent())
        {
            switch (eventId)
            {
                case EVENT_CHANGE_PHASE:
                    _phase = PHASE_X;
                    DoSomething...
                    break;
                case EVENT_CAST_SPELL_Y:
                    DoCastVictim(SPELL_Y);
                    events.ScheduleEvent(SPELL_Y, 10s);
                    break;
```

and about Kil'jaeden script, he check phases in UpdateAI but doesn't check casting state:

https://github.com/TrinityCore/TrinityCore/blob/c4741a131a0a28b69c228d3637beab7090e5f1e3/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kiljaeden.cpp#L865-L910

offl:
I don't remember if I tried to launch events in DamageTaken
But yeah, sounds better

Keader:
i belive that blizzard dont have a "OnDamageTaken" hook
If you sniff a creature that active some event at X% of HP, it always has a variance on amount...
i realy belive that blizzard check with "events" like Jildor did

offl:
They have OnHealthRange and OnHealthRangeRandom

PR #27213, opened 2021-10-30
Core/SAI: Add actions to reset unit_flags to default

offl:
**Changes proposed:**

-  Add actions to reset unit_flags to default
-  Change target of unit_flags actions from unit to creature, we don't want to accidentally make player immune(how it never happened)

**Issues addressed:**

none

**Tests performed:**

none

- [ ] test it if build will be not broken

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScriptMgr.h
+++ b/src/server/game/AI/SmartScripts/SmartScriptMgr.h
@@ -587,10 +587,13 @@ enum SMART_ACTION
     SMART_ACTION_SET_HEALTH_PCT                     = 142,    // percent
     SMART_ACTION_CREATE_CONVERSATION                = 143,    // don't use on 3.3.5a
     SMART_ACTION_SET_IMMUNE_PC                      = 144,    // 0/1
-    SMART_ACTION_SET_IMMUNE_NPC                     = 145,    // 0/1
-    SMART_ACTION_SET_UNINTERACTIBLE                 = 146,    // 0/1
+    SMART_ACTION_RESET_IMMUNE_PC                    = 145,    //
+    SMART_ACTION_SET_IMMUNE_NPC                     = 146,    // 0/1
+    SMART_ACTION_RESET_IMMUNE_NPC                   = 147,    //
+    SMART_ACTION_SET_UNINTERACTIBLE                 = 148,    // 0/1
+    SMART_ACTION_RESET_UNINTERACTIBLE               = 149,    //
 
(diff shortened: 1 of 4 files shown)


src/server/game/AI/SmartScripts/SmartScriptMgr.h
```diff
@@ -587,10 +587,13 @@ enum SMART_ACTION
     SMART_ACTION_SET_HEALTH_PCT                     = 142,    // percent
     SMART_ACTION_CREATE_CONVERSATION                = 143,    // don't use on 3.3.5a
     SMART_ACTION_SET_IMMUNE_PC                      = 144,    // 0/1
-    SMART_ACTION_SET_IMMUNE_NPC                     = 145,    // 0/1
-    SMART_ACTION_SET_UNINTERACTIBLE                 = 146,    // 0/1
+    SMART_ACTION_RESET_IMMUNE_PC                    = 145,    //
```
jackpoz, line 590: Would it be possible to keep the old values for existing actions ?

offl: If necessary, yes

jackpoz:
@offl is this ready to be merged ?

Shauren:
No, it is not

offl:
There's 20+ flags which has such action, one action for each flag
It should be done by calling ResetFlag(flag). All those flags are static, so when all flags will be in one place, we'll be able to reset more of them. However only specific flags can be reset and only one flag at once to let people think twice what they're doing because every Reset flag action no matter which flag it resets is quite dangerous and should be used wisely because unlike with factions which are already default values and will be never changed in 99% cases and in 99% cases it's safe to reset faction to default, our flags are not default in lots of cases

At the same time by doing it with ResetFlag(flag), it will allow to call it in /scripts like with SetFlag\RemoveFlag which leads to dumb things in scripts. I suggested earlier to purge non-static flags from /scripts like I did for SAI and dissalow calling SetFlag in /scripts by replacing it with unique functions. So probably it's not that bad to have 20 reset and 20 set   functions, they're quite small because inside them all you do is call ResetFlag and SetFlag(off\on). We already use SetImmuneTo, for example. All other SetFlag in /scripts in case of unit_flags are just hacks

Btw I'm not sure if it should take care of flags in `creature`, because Blizz simply does not add flag if at least one spawn will not have it. Maybe not always but in lots of cases it's how it looks like in CreatureDifficulty. Makes me think we are not ready for something like that, maybe we should not even implement anything like that

Issue #27218, opened 2021-10-30
[Spell] Warrior - No Rage generation with parry 


w5860363:
### Description

https://wowwiki-archive.fandom.com/wiki/Parry

`When a warrior's special attack is parried, he does not lose the rage cost associated with that spell (even though the spell fails). Miss, dodge, or block events will consume the full rage cost when the spell fails.
`
Now, when parry, it consumes all your Rage.

Maybe issue:https://github.com/TrinityCore/TrinityCore/issues/24567
Maybe related: 484c681

### Expected behaviour

Warrior returns some of his Rage when he casts spell and is parried.

### Steps to reproduce the problem

Get Warrior
Get Damage Hits
parry  the hits
no rage

### Branch

3.3.5

### TC rev. hash/commit

https://github.com/TrinityCore/TrinityCore/commit/9b354c6bebdf9fe9c9842330447162edc63cc751

### Operating system

Windows 10 x64,

### Custom changes

None

PR #27220, opened 2021-10-31
Core/Spells: Warrior spell is knocked back with DODGE when returning to rage consumption

w5860363:
test https://github.com/TrinityCore/TrinityCore/issues/27218
https://wowwiki-archive.fandom.com/wiki/Parry The above statement is not the same as https://www.bluetracker.gg/wow/topic/eu-en/12911577-18-02-05-kalgans-response- to-warriors/ above does not match
Unable to determine if SPELL_MISS_DODGE is included and returns full cost or 80%

**Changes proposed:**

- Warrior cast spell
-  DODGE and PARRY the hits 
-  return rage cost

**Issues addressed:**
https://github.com/TrinityCore/TrinityCore/issues/27218
**Tests performed:**
 tested in-game,

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2559,6 +2559,8 @@ void Spell::TargetInfo::DoDamageAndTriggers(Spell* spell)
         // Do not take combo points on dodge and miss
         if (MissCondition != SPELL_MISS_NONE && spell->m_needComboPoints && spell->m_targets.GetUnitTargetGUID() == TargetGUID)
             spell->m_needComboPoints = false;
+        else if (caster->GetTypeId() == TYPEID_PLAYER && MissCondition == SPELL_MISS_PARRY && MissCondition == SPELL_MISS_DODGE && spell->m_spellInfo->PowerType == POWER_RAGE)
+            caster->ModifyPower(POWER_RAGE, spell->m_powerCost);
 
         // _spellHitTarget can be null if spell is missed in DoSpellHitOnUnit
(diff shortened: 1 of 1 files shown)


Ovahlord:
changed the title to be a... little bit more on point.

src/server/game/Spells/Spell.cpp
```diff
@@ -2559,6 +2559,8 @@ void Spell::TargetInfo::DoDamageAndTriggers(Spell* spell)
         // Do not take combo points on dodge and miss
         if (MissCondition != SPELL_MISS_NONE && spell->m_needComboPoints && spell->m_targets.GetUnitTargetGUID() == TargetGUID)
             spell->m_needComboPoints = false;
+        else if (caster->GetTypeId() == TYPEID_PLAYER && MissCondition == SPELL_MISS_PARRY && MissCondition == SPELL_MISS_DODGE && spell->m_spellInfo->PowerType == POWER_RAGE)
+            caster->ModifyPower(POWER_RAGE, spell->m_powerCost);
```
jackpoz, line 2563: Are you sure the && are correct ? Did you debug the code changes with different test cases to ensure it does what you are trying to make it do ?

Ovahlord:
This change is wrong. If you want to prevent losing rage when the hit result is mismatching, you should check Spell::TakePower  because that's the place where the power consumption happens.

w5860363:
Understood

jackpoz:
You can still continue the work in this PR, integrating the comments in your changes

PR #27240, opened 2021-11-02
Scripts/Creature: Kill trigger_periodic

offl:
**Changes proposed:**

-  Replaced trigger_periodic with unique scripts

**Issues addressed:**

none

**Tests performed:**

none

- [x] test


src/server/scripts/Northrend/Naxxramas/boss_sapphiron.cpp
```diff
@@ -581,6 +593,42 @@ class spell_sapphiron_summon_blizzard : public SpellScript
...
 };
 
+// 29330 - Sapphiron's Wing Buffet Despawn
+class spell_sapphiron_wing_buffet_despawn_periodic : public AuraScript
+{
+    PrepareAuraScript(spell_sapphiron_wing_buffet_despawn_periodic);
+
+    void PeriodicTick(AuraEffect const* /*aurEff*/)
+    {
+        if (GetTarget()->GetTypeId() != TYPEID_UNIT)
```
Ovahlord, line 603: you can just go with  
```
if (Creature* creature = target->ToCreature())
    creature->DespawnOrUnsummon();
```

offl:
Since that void zone is summoned by 4 spells and uses only one damaging spell, looks like I need to check UNIT_CREATED_BY_SPELL of caster directly in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/Naxxramas/boss_four_horsemen.cpp#L702 and use damage from effect2 of summon spells https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=16697#summoned-by-spell

Because doesn't look like I can do same thing as in https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/Ulduar/Ulduar/boss_xt002.cpp#L663-L674 because here we don't cast triggered spell directly and use periodic aura (same should be done for Ulduar)

PR #27256, opened 2021-11-07
Scripts/Spells: Implement Demon Hunter Base Spell "Glide"

### LINKED ISSUE #18303: [Demon Hunter] Glide

**Description:** Demon Hunters' glide does not dispel upon touching the ground

**Current behaviour:** The Demon Hunter keeps gliding around on the ground 

**Expected behaviour:** Upon touching the ground the glide aura should be removed from the Demon Hunter

1. Create a Demon Hunter
2. Jump 3 times or activate Glide 

**Branch(es):** master 

**TC rev. hash/commit:** https://github.com/TrinityCore/TrinityCore/commit/9f31a6aa3cd72f25d3fe49c6547f599772ddaeda

**TDB version:**  703.00

**Operating system:**  Windows 10 x64

Seyden:
**Changes proposed:**

-  Implement Demon Hunter Spell "Glide"
-  Remove Movementflag check in SetFeatherFall due to the reason that anticheat removes the flag under a certain condition and not informing the client which results in that if a feather fall effect is removed, that the check says: you're not feather falling but we also didn't inform the client which results in a endless feather fall
-  Implement SendForcedCooldown to inform the client about a small forced cooldown

**Issues addressed:**

Closes #18303 

**Tests performed:**

- [ ]
- [ ]


src/server/scripts/Spells/spell_dh.cpp
```diff
@@ -231,6 +363,82 @@ struct areatrigger_dh_sigil_of_chains : AreaTriggerAI
...
+    }
+
+    void OnRemoveGlide(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+    {
+        GetTarget()->RemoveAura(SPELL_DH_GLIDE_DURATION);
+    }
+
+    void Register() override
+    {
+        if (m_scriptSpellId == SPELL_DH_GLIDE_DURATION)
```
Shauren, line 435: These are two completely separate scripts, different `OnRemove*` functions, different params to `AuraEffectRemoveFn`, so why did you shove them both into the same class, especially when `spell_dh_glide` should apply to only one of them?

Shauren (changes requested):

src/server/scripts/Spells/spell_dh.cpp
Seyden: alright alright.. i just wanted to save the extra code

PR #27258, opened 2021-11-08
Core/SAI: Remove deprecated events and actions

offl:
**Changes proposed:**

-  Remove deprecated events and actions

**Issues addressed:**

none

**Tests performed:**

none

- [x] run worldserver

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -352,11 +352,6 @@ bool SmartAI::IsEscortInvokerInRange()
     return true;
 }
 
-void SmartAI::WaypointStarted(uint32 nodeId, uint32 pathId)
-{
-    GetScript()->ProcessEventsFor(SMART_EVENT_WAYPOINT_START, nullptr, nodeId, pathId);
-}
-
 void SmartAI::WaypointReached(uint32 nodeId, uint32 pathId)
 {
     if (!HasEscortState(SMART_ESCORT_ESCORTING))
@@ -742,16 +737,6 @@ void SmartAI::SetRun(bool run)
 void SmartAI::SetDisableGravity(bool fly)
 {
     me->SetDisableGravity(fly);
-}
-
-void SmartAI::SetCanFly(bool fly)
-{
-    me->SetCanFly(fly);
-}
-
-void SmartAI::SetSwim(bool swim)
-{
-    me->SetSwim(swim);
 }
 
 void SmartAI::SetEvadeDisabled(bool disable)
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -1326,28 +1326,12 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
             ENSURE_AI(SmartAI, me->AI())->SetDisableGravity(e.action.setDisableGravity.disable != 0);
             break;
         }
-        case SMART_ACTION_SET_CAN_FLY:
-        {
-            if (!IsSmart())
-                break;
-
-            ENSURE_AI(SmartAI, me->AI())->SetCanFly(e.action.setFly.fly != 0);
-            break;
-        }
         case SMART_ACTION_SET_RUN:
         {
             if (!IsSmart())
                 break;
 
             ENSURE_AI(SmartAI, me->AI())->SetRun(e.action.setRun.run != 0);
-            break;
-        }
-        case SMART_ACTION_SET_SWIM:
-        {
-            if (!IsSmart())
-                break;
-
-            ENSURE_AI(SmartAI, me->AI())->SetSwim(e.action.setSwim.swim != 0);
             break;
         }
         case SMART_ACTION_SET_COUNTER:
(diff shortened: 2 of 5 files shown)


src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -2251,13 +2182,6 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
             ProcessEventsFor((SMART_EVENT)SMART_EVENT_TIMED_EVENT_TRIGGERED, nullptr, eventId);
             break;
         }
-        case SMART_ACTION_REMOVE_ALL_GAMEOBJECTS:
```
Shauren, line 2254: Why is this one deprecated/removed?

offl: Because I can't imagine a scenario where creature need to remove all owned gameobjects

src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -352,11 +352,6 @@ bool SmartAI::IsEscortInvokerInRange()
     return true;
 }
 
-void SmartAI::WaypointStarted(uint32 nodeId, uint32 pathId)
-{
-    GetScript()->ProcessEventsFor(SMART_EVENT_WAYPOINT_START, nullptr, nodeId, pathId);
```
ccrs, line 357: and this one? also another one you cant imagine where to use it? my god

offl:
In case of some actions or events and their params removing them or adding heavily depends on what we want from that system. Have a system which does not allow to do completely weird things and is quite limited(in a good way) or have a **uncontrollable** system which may allow more but in an dangerous way, in lots of cases without real reason

It may be hard to decide what never will be needed and what may be needed. List of original AI hooks\actions and params does not always help, for example some things should be always done by spells but it may be really hard to know which spell should be used(especially in master). Those actions probably will stay here forever. Even if I remove something without knowing some facts, events or actions can be restored. Also nothing is removed because not used but because I did some research and turned out it's not needed/should not happen/horrible/dangerous/can be scripted in a more legit way in SAI even today. Or re-implemented in new action or event and old ones were removed

For example SMART_EVENT_EVENT_PHASE_CHANGE probably was used first time by me a year ago, then turned out Blizz uses no phases at all and the scripts where that event type was used can be more clean without it. The system which can fully replace phases at least in SAI is quite interesting and solves many problems
SMART_EVENT_WAYPOINT_START was broken for years, when it was fixed, turned out it even cannot be used the way we want to use it. Then turned out Blizz uses no hooks for waypoints in DB scripts. On top of that WaypointStarted was always barely used because it's simply not necessary to use it in general
Looks like SMART_EVENT_CHARMED_TARGET was never even handled. I'm afraid of even thinking what can be scripted when creature charms another creature or a player

```cpp
        case SMART_ACTION_SET_SWIM:
        case SMART_ACTION_OVERRIDE_SCRIPT_BASE_OBJECT:
        case SMART_ACTION_RESET_SCRIPT_BASE_OBJECT:
        case SMART_ACTION_SET_CAN_FLY:
```
SetSwim and SetCanFly should not be called by scripts, CanFly should not be stored in DB. SetCanFly currently is used only in C++ scripts and should be replaced by DisableGravity
```cpp
        case SMART_ACTION_REMOVE_AURAS_BY_TYPE:
        case SMART_ACTION_SET_SIGHT_DIST:
        case SMART_ACTION_FLEE:
        case SMART_ACTION_REMOVE_ALL_GAMEOBJECTS:
```
As I said some things really depends on what we want from SAI and for which branch
SMART_ACTION_REMOVE_AURAS_BY_TYPE is really something we want to use in DB scripts? Aren't you afraid? Blizz is afraid to give something like that to their designers and they can remove only one aura at once, not all auras, not stack of aura, not all auras on evade, not all passive auras, not all owned auras
SMART_ACTION_SET_SIGHT_DIST is a part of Change radius record for unit(s) action but radius record is not only sight dist but also call for help radius, leash radius and stuff like that. It's probably not even possible to notice if radius record was changed at runtime. The reason why someone added that action is because sight dist should be different from creature to creature but all of that originally is stored in DB and not in core as static number. Better to do it in creature_template otherwise we're making our scripts 'heavy' and there already are like 2k actions which changes states and stuff like that but all of that is supposed to be stored as flags and not called every time directly in scripts when creature spawns. SAI with its linked events and the way those linked events are used doesn't allow to remove anything in an easy way without spending lot of time
SMART_ACTION_FLEE flee is not supposed to be called like that
SMART_ACTION_REMOVE_ALL_GAMEOBJECTS looks like a way to make something crazy in scripts. As I said I don't know where do you need this, it's used few times in C++ scripts and with huge chance it's something what is supposed to make things easier and making things easier in emu usually leads to disaster. It's not used in SAI currently and is used only few times in core. But there are actions and events both in core and SAI which are used hundreds of times and in lots of cases the way they are implemented does not allow to remove them easily or adjust. And in lots of cases they blocks new features. For example action to change npc flags to 0 is used 300 times, who was thinking it's a good idea to not know what you're changing I don't know, but now without cleaning those 300 scripts we can't stop storing npc flags in DB and make them trully dynamic. If I see something what should not be used, I'd rather remove it now than wait before it will be used not in few cases but in tens or hundreds of cases. That case with npc flags is one of tens if not of hundreds

src/server/game/AI/SmartScripts/SmartAI.cpp
offl: SMART_EVENT_WAYPOINT_START was broken for years, when it was fixed, turned out it even cannot be used the way we want to use it. Then turned out Blizz uses no hooks for waypoints in DB scripts. On top of that WaypointStarted was always barely used because it's simply not necessary to use it in general

You can read about all of that here https://github.com/TrinityCore/TrinityCore/issues/25179#issuecomment-941750597 and below

PR #27273, opened 2021-11-13
Core/Spells: Implemented SummonPropertiesFlags: UseSummonerFaction & IgnoreSummonerPhase

meji46:
…d SummonPropertiesFlags::IgnoreSummonerPhase

**Changes proposed:**

-  Defined enum class SummonPropertiesFlags.
- Implemented use of SummonPropertiesFlags::IgnoreSummonerPhase.
- Added use of SummonPropertiesFlags::UseSummonerFaction in TempSummon::InitStats.
- Removed several calls to SetFaction(faction) in functions related to the summon of a creature.

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

- [ ] Determine priority between flag SummonPropertiesFlags::UseSummonerFaction and faction defined in SummonProperties.db2.
- [ ] This [line](https://github.com/meji46/TrinityCore/blob/441599a79d8f1e6a6b67971c532bc50048acebfc/src/server/game/Spells/SpellEffects.cpp#L2052) (Spell::EffectSummonType) uses a summon method that doesn't pass the SummonProperties as a parameter. I don't know if it could be changed to the same one used in the rest of the function.
- [ ] In [Minion::InitStats](https://github.com/meji46/TrinityCore/blob/441599a79d8f1e6a6b67971c532bc50048acebfc/src/server/game/Entities/Creature/TemporarySummon.cpp#L296) the summoner's faction is forced. This overrides the use of the flag when the creature has the masks: UNIT_MASK_GUARDIAN, UNIT_MASK_TOTEM, UNIT_MASK_PUPPET or UNIT_MASK_MINION (Vehicles and critters).
    - In SummonProperties.db2 the only summon type that doesn't use the flag (according to the Control field), is Possessed (SUMMON_CATEGORY_PUPPET). I assume that, in this case, the faction is forced by applying the charm. In the other cases, the flag may appear so possibly this SetFaction isn't correct.
    - There are several cases of vehicles (SUMMON_CATEGORY_VEHICLE) without the flag that, according to sniffs, use the summoner's faction. Perhaps there is a seat flag that serves to identify when to set the summoner's faction.
    - Summons with slot 5 (Critter) possibly set the summoner's faction.

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1789,9 +1789,8 @@ TempSummon* Map::SummonCreature(uint32 entry, Position const& pos, SummonPropert
         return nullptr;
     }
 
-    if (summoner)
-        if (summoner->IsPlayer() || !(properties && properties->GetFlags().HasFlag(SummonPropertiesFlags::IgnoreSummonerPhase)))
-            PhasingHandler::InheritPhaseShift(summon, summoner);
+    if (summoner && !(properties && properties->GetFlags().HasFlag(SummonPropertiesFlags::IgnoreSummonerPhase)))
+        PhasingHandler::InheritPhaseShift(summon, summoner);
 
     summon->SetCreatedBySpell(spellId);
 
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1752,7 +1752,7 @@ TempSummon* Map::SummonCreature(uint32 entry, Position const& pos, SummonPropert
                     mask = UNIT_MASK_MINION;
                     break;
                 default:
-                    if (properties->Flags & 512) // Mirror Image, Summon Gargoyle
+                    if (properties->GetFlags().HasFlag(SummonPropertiesFlags::JoinSummonerSpawnGroup))
                         mask = UNIT_MASK_GUARDIAN;
                     break;
                 }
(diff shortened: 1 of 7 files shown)


src/server/game/Entities/Transport/Transport.cpp
```diff
@@ -472,7 +472,12 @@ TempSummon* Transport::SummonPassenger(uint32 entry, Position const& pos, TempSu
         return nullptr;
     }
 
-    PhasingHandler::InheritPhaseShift(summon, summoner ? static_cast<WorldObject*>(summoner) : static_cast<WorldObject*>(this));
+    WorldObject* phaseShiftOwner = static_cast<WorldObject*>(this);
```
Shauren, line 475: static_cast is no longer neccessary outside of ` ? : ` expression

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1789,9 +1789,9 @@ TempSummon* Map::SummonCreature(uint32 entry, Position const& pos, SummonPropert
         return nullptr;
     }
 
-    // Set the summon to the summoner's phase
     if (summoner)
-        PhasingHandler::InheritPhaseShift(summon, summoner);
+        if (summoner->IsPlayer() || !(properties && properties->GetFlags().HasFlag(SummonPropertiesFlags::IgnoreSummonerPhase)))
```
Shauren, line 1793: Why is player getting a special case?

meji46: The full name is "Ignore Summoner's Phase (Wild Only)" and I thought it meant it only affected summoners that are creatures and not players.

Is it better if I remove that part?

Shauren: "Wild" is a type of summon

Ovahlord:
> Determine priority between flag SummonPropertiesFlags::UseSummonerFaction and faction defined in SummonProperties.db2.

Well, this seems to be pretty simple. If the flag is given and there actually is a summoner (you can still summon creatures without summoners) you pull his faction. If no summoner is provided, fall back to the dbc value.

meji46:
> Well, this seems to be pretty simple. If the flag is given and there actually is a summoner (you can still summon creatures without summoners) you pull his faction. If no summoner is provided, fall back to the dbc value.

This is the logic I've followed but I left it stated in case someone has a retail example to confirm.

PR #27277, opened 2021-11-15
Scripts/Mechanar: Rework Nethermancer Sepethrea

offl:
**Changes proposed:**

-  Boss and summons now uses only correct spells, other small adjustments
Summons also chases random non-tank target, however this part is far from absolutely correct but close enough under specific conditions. Hard to say what exactly they do. 1000000 threat is original way to support that, sniffed

Took some ideas from mangos

**Issues addressed:**

none

**Tests performed:**

Builds, tested in-game


src/server/scripts/Outland/TempestKeep/Mechanar/boss_nethermancer_sepethrea.cpp
```diff
@@ -15,77 +15,92 @@
...
-        Talk(SAY_SLAY);
+        _JustReachedHome();
+        DoCastSelf(SPELL_FROST_ATTACK);
     }
 
     void JustDied(Unit* /*killer*/) override
     {
-        _JustDied();
+        events.Reset();
+        instance->SetBossState(DATA_NETHERMANCER_SEPRETHREA, DONE);
```
Ovahlord, line 101: why did you remove _JustDied()  here?  This will lead to summons not getting despawned when the encounter ends so you will have orphaned spawns here.

offl: They're killed by SPELL_QUELL_RAGING_FLAMES

Ovahlord: Then don't store them at all.  

`void JustSummoned(Creature* /*summon*/) override { }`

This way you can cast the spell easily on evade and death and use the default ai hooks.

offl: I still despawn them on evade with summons.DespawnAll();

```cpp
    void EnterEvadeMode(EvadeReason why) override
    {
        // Fails because target is in evade mode (yes, she kills them on evade too)
        // DoCastSelf(SPELL_QUELL_RAGING_FLAMES, true);
        summons.DespawnAll();
```

Ovahlord: And why did you comment the spell then instead of using it?

offl: If it will be still safe to call both kill and despawn when kill will be fixed, I can uncomment it

offl: That's probably dumb logic, I'll uncomment it later

PR #27279, opened 2021-11-15
Change order in HandleLootOpcode

acidmanifesto:
I just can not understand why we have a check occuring after SendLoot. All checks should be before it in this case. Viewing gitblame, this was last touched back about 13 years ago and with no notes explaining the reasoning. Would love a further explaination in the reason why interrupt cast if statement is occurring after a send loot and not before.

**Changes proposed:**

-  Change Order of GetPlayer()->SendLoot(guid, LOOT_CORPSE); to occur after checks and not inbetween checks within the HandleLootOpcode.

**Issues addressed:**

Closes #  My OCD Sanity

**Tests performed:**

It does build with no issues in game from what i can see.
I am able to craft multiple items with tayloring, and able to loot corpses just fine.
cast spells or while crafting does get interrupted when i decide to loot something

### DIFF
--- a/src/server/game/Handlers/LootHandler.cpp
+++ b/src/server/game/Handlers/LootHandler.cpp
@@ -227,11 +227,11 @@ void WorldSession::HandleLootOpcode(WorldPacket& recvData)
     if (!GetPlayer()->IsAlive() || !guid.IsCreatureOrVehicle())
         return;
 
-    GetPlayer()->SendLoot(guid, LOOT_CORPSE);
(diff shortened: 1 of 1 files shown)


Shauren (approved):
seems legit

Jinnaix (approved):
As we already spoke about this and there does not seem to be an issue with crafting or other stuff, this looks fine.

As a general guidance - conditional if checks should be executed before the related action.

PR #27280, opened 2021-11-16
Core/Battlegrounds: Avoid reset BG and Arena scoreboard stats and BG map achievement criterias when relogin if player was already in the BG

Jildor:
**Changes proposed:**

-  Avoid reset BG and Arena scoreboard stats and BG map achievement criterias when relogin if player was already in the BG
If during a BG or Arena in progress you lost the connection (or Alt + F4) and had to relogin, you lose the BG / Arena record of damage, healing, bases captured, etc.
The same with achievements that required to meet X objectives in the same BG, such as:
https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?achievement=583
https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?achievement=584
If you relogin, you lost the progress you made.

I do not know if you think the way I have done it is good or you think it is better to do it in another way

I have noticed that in SotA when you relogin you are teleported, should this case also be avoided or is it blizzlike (I doubt it is blizzlike, more like an exploit if you are a defender and want to go back sooner)?
https://github.com/TrinityCore/TrinityCore/blob/c4741a131a0a28b69c228d3637beab7090e5f1e3/src/server/game/Battlegrounds/Zones/BattlegroundSA.cpp#L483-L491
If so, I can create another PR after this one so as not to mix both things if you prefer.

**Issues addressed:**

None I think

**Tests performed:**

Tested in-game

### DIFF
--- a/src/server/game/Battlegrounds/Arena.cpp
+++ b/src/server/game/Battlegrounds/Arena.cpp
@@ -73,8 +73,10 @@ Arena::Arena()
 
 void Arena::AddPlayer(Player* player)
 {
+    bool const isInBattleground = IsPlayerInBattleground(player->GetGUID());
     Battleground::AddPlayer(player);
-    PlayerScores[player->GetGUID().GetCounter()] = new ArenaScore(player->GetGUID(), player->GetBGTeam());
+    if (!isInBattleground)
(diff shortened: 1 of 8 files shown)


jackpoz:
Thanks for the PR 🚀 😃

jackpoz:
@Jildor ping me on github next time it passes a month without a PR getting merged :)

Issue #27289, opened 2021-11-19
DB/Quest: destroy naberius


w5860363:
### Description

The target's spell 36146 should not be able to control all players but should be a random target

### Expected behaviour

Cast spell Chains of Naberius (36146) on Random hostile

### Steps to reproduce the problem

1.login two player
2.go cr 72506
3.Controlled Target out of combat

### Branch

3.3.5

### TC rev. hash/commit

https://github.com/TrinityCore/TrinityCore/commit/9cb86a42a9cc2c44507d1ce604a1856b91bb385e

### Operating system

Windows 10 x64,

### Custom changes

None

w5860363:
And you can't leave the house after the battle.

PR #27291, opened 2021-11-19
Try to solve spell  Chains of Naberius(36146)

w5860363:
test https://github.com/TrinityCore/TrinityCore/issues/27289

### DIFF
--- a/sql/custom/world/2021_11_17_02_world.sql
+++ b/sql/custom/world/2021_11_17_02_world.sql
@@ -1,0 +1,1 @@
+UPDATE `smart_scripts` SET `target_type`=6 WHERE `id`=0 AND `entryorguid`=20483;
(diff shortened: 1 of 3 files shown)


offl:
This spell indeed should target only one player but this more looks like castFlag 'cast only if threat size is > 1'
On retail in lots of cases creature actually can cast spell on current target but usually it does not because threat size is 1

Atm SMART_TARGET_HOSTILE_RANDOM_NOT_TOP here provides correct spell target selection under any condition but that happens probably only because the spell is AOE

w5860363:
@offl The original TDB is smart_ TARGET_ VICTIM

I didn't find out from the scattered retail videos that he would control the current hate players

offl:
I'm not sure what will happen with 2 players but if there is 1 player and his guardians, he will mind control player (spell can target only players)

offl:
What I was trying to say is 6 is fine, it does not require changes

Issue #27226, opened 2021-10-31
Core/Spell:Hunter Spell: Eyes of the Beast (1002)


w5860363:
### Description

http://wowwiki.wikia.com/wiki/Eyes_of_the_Beast
pet cast spell,will lose control and return to the master

### Expected behaviour

Should allow me to see through my pet's eyes and control my pet. And be able to use spells normally
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/CoreAI/PetAI.cpp#L360
Not relevant to this section Commenting out still doesn't work
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/AI/CoreAI/PetAI.cpp#L640
The problem lies in this part
### Steps to reproduce the problem

1.creature hunter
2.levle 79
3.learn spell
4.cast spell Eyes of the Beast
5.Control the pet to cast a spell and attack

### Branch

3.3.5

### TC rev. hash/commit

https://github.com/TrinityCore/TrinityCore/commit/e4430cd2bedf2079961bd1a30ba9f5f1de93af82

### Operating system

Windows 10 x64

### Custom changes

None

w5860363:
`void PetAI::OnCharmed(bool isNew)
{
    if (me->IsCharmed())
        me->GetMotionMaster()->MoveFollow(me->GetCharmer(), PET_FOLLOW_DIST, me->GetFollowAngle());

    CreatureAI::OnCharmed(isNew);
}`

Commenting out can solve the problem, but I don't know if brach 9. X wants to use it

PR #27293, opened 2021-11-19
Try to solve spell Eyes of the Beast (1002)

w5860363:
Try to solve https://github.com/TrinityCore/TrinityCore/issues/27226


Shauren:
Looks close but not quite there, `CreatureAI::OnCharmed` must still be called, you should check `isPossessedByPlayer()` before `MoveFollow`

Issue #26826, opened 2021-08-15
TC can't connect to database after restart mysql server


kvolk2git:
### Description

System user trinity is not root

**The first part of the error:**
After restart Mysql server and run ./authserver I get an error:

_Could not connect to MySQL database at 127.0.0.1: RSA Encryption not supported - caching_sha2_password plugin was built with GnuTLS support
DatabasePool Login NOT opened._ 

Trinity will not be able to connect to the database until the user's console is logged in to mysql

mysql -u trinity -ptrinity
mysql> exit;

I'm launching it again ./authserver

_Opening DatabasePool 'auth'. Asynchronous connections: 1, synchronous connections: 1.
MySQL client library: 10.3.31
MySQL server ver: 8.0.26-0ubuntu0.20.04.2
Connected to MySQL database at 127.0.0.1
MySQL client library: 10.3.31
MySQL server ver: 8.0.26-0ubuntu0.20.04.2
Connected to MySQL database at 127.0.0.1
DatabasePool 'auth' opened successfully. 2 total connections running.
Started auth database connection pool.
Loading IP Location Database...
Added realm "Trinity" at 192.168.1.30:8085._

or  in config file need add "ssl" in connection string

authserver.conf  
LoginDatabaseInfo = "127.0.0.1;3306;trinity;trinity;auth;ssl"

worldserver.conf
LoginDatabaseInfo     = "127.0.0.1;3306;trinity;trinity;auth;ssl"
WorldDatabaseInfo     = "127.0.0.1;3306;trinity;trinity;world;ssl"
CharacterDatabaseInfo = "127.0.0.1;3306;trinity;trinity;characters;ssl"

I decided to go the second way and added " ssl " to the connection strings,  and trinity began to connect to the database without the need to log in to the console in mysql after restarting

**The second part of the error:**
After a few minutes, I synchronized the repository.

_create mode 100644 sql/updates/world/3.3.5/2021_08_15_01_world.sql
 create mode 100644 sql/updates/world/3.3.5/2021_08_15_02_world.sql
 create mode 100644 sql/updates/world/3.3.5/2021_08_15_03_world_335.sql_

And started the server. The server crashed with an error when connecting.Using configuration file 

_/home/trinity/server/etc/worldserver.conf.
Using SSL version: OpenSSL 1.1.1f  31 Mar 2020 (library: OpenSSL 1.1.1f  31 Mar 2020)
Using Boost version: 1.71.0
Updating Auth database...
[...]

PR #27298, opened 2021-11-20
TrinityCore & MySQL8 & SSL - fix error autoupdate server database.

kvolk2git:
**Changes proposed:**

If you are using a MySQL 8 server and trying to connect via SSL, then after updating the TC server, if there are sql update files, you will get an error similar to this:

_Applying of file '/home/trinity/TrinityCore/sql/updates/world/3.3.5/2021_11_17_99_test_world.sql' to database 'world' failed! If you are a user, please pull the latest revision from the repository. Also make sure you have not applied any of the databases with your sql client. You cannot use auto-update system and import sql files from TrinityCore repository with your sql client. If you are a developer, please fix your sql query.
Could not update the World database, see log for details._

This is due to the fact that the --ssl parameter has been removed from version 8 in the MySQL server, and now you need to use --ssl-mode=REQUIRED instead. You can read more in the changelog, here https://dev.mysql.com/doc/refman/8.0/en/mysql-nutshell.html

>The client-side --ssl and --ssl-verify-server-cert options have been removed. Use --ssl-mode=REQUIRED instead of --ssl=1 or --enable-ssl. Use --ssl-mode=DISABLED instead of --ssl=0, --skip-ssl, or --disable-ssl. Use --ssl-mode=VERIFY_IDENTITY instead of --ssl-verify-server-cert options. (The server-side --ssl option is still available, but is deprecated as of MySQL 8.0.26 and subject to removal in a future MySQL version.)

>For the C API, MYSQL_OPT_SSL_ENFORCE and MYSQL_OPT_SSL_VERIFY_SERVER_CERT options for mysql_options() correspond to the client-side --ssl and --ssl-verify-server-cert options and are removed. Use MYSQL_OPT_SSL_MODE with an option value of SSL_MODE_REQUIRED or SSL_MODE_VERIFY_IDENTITY instead._

**Issues addressed:**

Partially closes #  (#26826)
> 2. The "ssl" connection parameter should not affect the possibility of updating the server database.

### DIFF
--- a/src/server/database/Updater/DBUpdater.cpp
+++ b/src/server/database/Updater/DBUpdater.cpp
@@ -370,8 +370,17 @@ void DBUpdater<T>::ApplyFile(DatabaseWorkerPool<T>& pool, std::string const& hos
     // Set max allowed packet to 1 GB
     args.emplace_back("--max-allowed-packet=1GB");
 
+#if !defined(MARIADB_VERSION_ID) && MYSQL_VERSION_ID >= 80000
+
+    if (ssl == "ssl")
+        args.emplace_back("--ssl-mode=REQUIRED");
+
+#else
+
     if (ssl == "ssl")
         args.emplace_back("--ssl");
+
+#endif
 
     // Database
     if (!database.empty())


Shauren (approved):

src/server/database/Updater/DBUpdater.cpp
```diff
@@ -370,9 +370,18 @@ void DBUpdater<T>::ApplyFile(DatabaseWorkerPool<T>& pool, std::string const& hos
     // Set max allowed packet to 1 GB
     args.emplace_back("--max-allowed-packet=1GB");
 
+#if !defined(MARIADB_VERSION_ID) && MYSQL_VERSION_ID >= 80000
+
+    if (ssl == "ssl")
+        args.emplace_back("--ssl-mode=REQUIRED");
+
+#else
+
```
Shauren, line 379:
```suggestion
#if !defined(MARIADB_VERSION_ID) && MYSQL_VERSION_ID >= 80000
    if (ssl == "ssl")
        args.emplace_back("--ssl-mode=REQUIRED");
#else
```

src/server/database/Updater/DBUpdater.cpp
```diff
@@ -370,9 +370,18 @@ void DBUpdater<T>::ApplyFile(DatabaseWorkerPool<T>& pool, std::string const& hos
...
     if (ssl == "ssl")
         args.emplace_back("--ssl");
 
+#endif
+
```
Shauren, line 384:
```suggestion
#endif
```

PR #27308, opened 2021-11-22
(Core\CMD) Npc Support for Freeze and Unfreeze

acidmanifesto:
**Changes proposed:**

- Remove Magic Numbers dealing with 9454 (Freeze Aura)
- Added Support for GM Cmd .freeze and .unfreeze to target and work on  NPCS

**Issues addressed:**

Closes #  No Known Issues

**Tests performed:**

- [x] Builds.
- [x] Tested in Game.

**Steps to Test:**

1. Target moving npcs and type .freeze to freeze the npcss.
2. Target moving npcs and type .unfreeze to unfreeze the npcs with the frozen aura.

### DIFF
--- a/src/server/scripts/Commands/cs_misc.cpp
+++ b/src/server/scripts/Commands/cs_misc.cpp
@@ -52,6 +52,11 @@
 #include "World.h"
 #include "WorldSession.h"
 
+enum FreezeAura
+{
+    FreezeAura = 9454
+};
+
 // temporary hack until includes are sorted out (don't want to pull in Windows.h)
 #ifdef GetClassName
 #undef GetClassName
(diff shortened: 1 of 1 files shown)


MaxtorCoder:
Oof, plz change indent settings and reformat the file 😄

acidmanifesto:
> Oof, plz change indent settings and reformat the file 😄

i just have github lie to me for my confidence and i select hide whitespace changes. :)
Honestly. not certain how the indents happened truthfully.

src/server/scripts/Commands/cs_misc.cpp
```diff
@@ -2549,6 +2555,7 @@ class misc_commandscript : public CommandScript
     {
         std::string name;
         Player* player;
+        Unit* target = handler->getSelectedUnit();
```
Goatform, line 2558: null check?

src/server/scripts/Commands/cs_misc.cpp
```diff
@@ -2449,6 +2454,7 @@ class misc_commandscript : public CommandScript
 
     static bool HandleFreezeCommand(ChatHandler* handler, char const* args)
     {
+        Unit* target = handler->getSelectedUnit();
```
Goatform, line 2457: check if null check is required

src/server/scripts/Commands/cs_misc.cpp
acidmanifesto: got a check in line 2594, hopefully it is what you ment. If not then im a bit confused in how you would want the null check to be done exactly in this case.

src/server/scripts/Commands/cs_misc.cpp
acidmanifesto: got a check in line 2534, hopefully it is what you ment. If not then im a bit confused in how you would want the null check to be done exactly in this case.

jackpoz:
The if blocks are not chained together correctly.
there is `if (has aura) return; if (does not have aura) return; else`, meaning the `else` will never happen. the old code also had the `else` linked to `if (player)`. Please check this code again and make sure the cases handled before are still handled correctly.

(I think comment in the code as github is having issues)

jackpoz:
> The if blocks are not chained together correctly. there is `if (has aura) return; if (does not have aura) return; else`, meaning the `else` will never happen. the old code also had the `else` linked to `if (player)`. Please check this code again and make sure the cases handled before are still handled correctly.
> 
> (I think comment in the code as github is having issues)

@acidmanifesto this PR has some pending review for you to deal with 😃

acidmanifesto:
Lol. So it is. I been caught up on the anticheat wip i pretty much forgot about the others. When i get some free time outside of the anticheat obsession i will try to get back to this.

acidmanifesto:
I am going to close this for now and reopen it another way at a later time.

PR #27313, opened 2021-11-24
Scripts/Crypts: Rework Exarch Maladaar

offl:
**Changes proposed:**

-  new register model, modern codestyle
-  texts are now grouped and handled correctly, intro text removed(was never used)
-  updated timers, logic for spells, the way Stolen Soul is summoned is now correct
-  Stolen Soul now "correctly" clones summoner-player. I know it looks like shit and probably was completely unnecessary considering how much time that thing took but it's what Blizz did back in the days. For some reason SetDisplayId(summonerModel) and HandleAuraCloneCaster is 2 completely different things. Cloning player model like it was done before by plain SetDisplayId(summonerModel) breaks model, while with HandleAuraCloneCaster it works correctly. No clue why, maybe unit_flags2
-  Stolen Soul now has more proper timers for class abilities & logic
-  Stolen Soul now dispels debuff after death

**Issues addressed:**

none

**Tests performed:**

tested


sql/updates/world/3.3.5/2021_11_24_10_world.sql
```diff
@@ -0,0 +1,18 @@
...
+(18373,1,0,"Stare into the darkness of your soul.",14,0,100,0,0,10511,17645,0,"maladaar SAY_SOUL_CLEAVE"),
+(18373,2,0,"Rise, my fallen brothers! Take form and fight!",14,0,100,0,0,10512,15466,0,"maladaar SAY_SUMMON"),
+(18373,3,0,"You will pay with your life!",14,0,100,0,0,10513,17647,0,"maladaar SAY_AGGRO_1"),
+(18373,3,1,"There is no turning back now!",14,0,100,0,0,10514,17648,0,"maladaar SAY_AGGRO_2"),
+(18373,3,2,"Serve your penitence!",14,0,100,0,0,10515,17649,0,"maladaar SAY_AGGRO_3"),
+(18373,4,0,"These walls will be your tomb!",14,0,100,0,0,10516,17650,0,"maladaar SAY_SLAY_1"),
+(18373,4,1,"Now you'll stay... for eternity.",14,0,100,0,0,10517,17651,0,"maladaar SAY_SLAY_2"),
+(18373,5,0,"This is... where I belong.",14,0,100,0,0,10518,17646,0,"maladaar SAY_DEATH");
+
+DELETE FROM `spell_linked_spell` WHERE `spell_trigger` = 33326;
```
Riztazz, line 16: Wasn't this table considered hack and shouldn't be used with newer scripts?

offl: Yeah, I heard that multiple times
However I heard also that 'it depends' but I don't really know how exactly, heh

offl: It looks like it can be used in certain cases but in general it's abused, probably that's why it's considered as hack
I'll add a spell script later

Ovahlord: good choice. spell linked spell is deprecated and avoid using it whenever possible.

Shauren:
HandleAuraCloneCaster is "new" wotlk tech, it didnt exist in tbc

offl:
It's used in SWP https://tbc.wowhead.com/spell=45785/sinister-reflection-clone

src/server/scripts/Outland/Auchindoun/AuchenaiCrypts/boss_exarch_maladaar.cpp
```diff
@@ -15,305 +15,365 @@
...
+        BossAI::JustEngagedWith(who);
+        Talk(SAY_AGGRO);
+        events.ScheduleEvent(EVENT_SOUL_SCREAM, RAND(10s, 15s, 20s, 25s));
+        events.ScheduleEvent(EVENT_RIBBON_OF_SOULS, 1s, 4s);
+        events.ScheduleEvent(EVENT_STOLEN_SOUL, RAND(10s, 15s, 20s, 25s));
+    }
+
+    void DamageTaken(Unit* /*attacker*/, uint32& /*damage*/, DamageEffectType /*damageType*/, SpellInfo const* /*spellInfo = nullptr*/) override
+    {
+        if (!_avatarSummoned && !HealthAbovePct(25))
```
Ovahlord, line 118: me->HealthBelowPctDamaged(25, damage); to ensure that the event is being scheduled right when the health falls below 25% pct.

offl: I was thinking about that because currently it's indeed scheduled in a weird way, can't even figure out why eventually it's always scheduled
I'll try later, thanks

Just in case here it's in separate event because of https://github.com/TrinityCore/TrinityCore/pull/27189#issuecomment-954938274

Ovahlord: that's fine, just use my suggested helper and you will get more consistent results. The problem with HealthAbovePct is that it returns false AFTER taking damage because   the DamageTaken hook is being called before the damage was applied on the health. HealthBelowPctDamage instead uses the incoming damage and checks it against the current health to handle exactly that problem.

offl:
```cpp
        switch (_summonerClass)
        {
            case CLASS_WARRIOR: spellsRecord = SPELLS_RECORD_WARRIOR; break;
            case CLASS_PALADIN: spellsRecord = SPELLS_RECORD_PALADIN; break;
            ...
        }

        if (spellsRecord)
            me->SetSpellsRecord(spellsRecord);
```

PR #27336, opened 2021-11-28
Core/SAI: Misc fixes

meji46:
**Changes proposed:**

-  Fixed crash when using SMART_TARGET_CLOSEST_* in areatrigger smart script

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -1179,11 +1179,12 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
         {
             for (WorldObject* target : targets)
             {
+                Unit* invoker = me ? me : unit;
                 if (Creature* cTarget = target->ToCreature())
                 {
                     CreatureAI* ai = cTarget->AI();
                     if (IsSmart(cTarget))
-                        ENSURE_AI(SmartAI, ai)->SetData(e.action.setData.field, e.action.setData.data, unit);
+                        ENSURE_AI(SmartAI, ai)->SetData(e.action.setData.field, e.action.setData.data, invoker);
                     else
                         ai->SetData(e.action.setData.field, e.action.setData.data);
                 }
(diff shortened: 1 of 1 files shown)


Shauren:
I don't know about the SMART_ACTION_SET_DATA change, isn't that going to affect/break existing scripts?

meji46:
> I don't know about the SMART_ACTION_SET_DATA change, isn't that going to affect/break existing scripts?

Only two cases in master:
```sql
SELECT * FROM smart_scripts WHERE event_type=38 AND target_type=7;
```
But in reviewing them, the following questions have come to my mind:
- If the action is associated to an actionlist I understand that SMART_ACTION_SET_DATA must have as invoker the creature that executes the actionlist.
- Possible ambiguity. Creature: SMART_EVENT_ACCEPTED_QUEST -> SMART_ACTION_SET_DATA. Who should be the invoker? The player who accepts the quest or the questgiver?

Maybe the best is something like:
```c++
Unit* invoker = me ? me : unit;
```
If the one executing the action is a creature, it acts as an invoker. Otherwise (areatriggers or gameobjects) the invoker of the event will be the invoker of the action.

offl:
>Possible ambiguity. Creature: SMART_EVENT_ACCEPTED_QUEST -> SMART_ACTION_SET_DATA. Who should be the invoker? The player who accepts the quest or the questgiver?

It's something that is related to how targets & scripting system works, the way targets works can be observed checking targets of texts during scripted events. In case of all events with more than one creature and with player who activates event through gossip \ quest end \ quest start. Also events with multiple talking with each other creatures
Shortly after you may feels like it may be automated target selection but I don't know if that is true or not
The fact is in lots of cases target selection is weird enough to not think it's not automated or target is somehow inherited not by our SEND_TARGET action which they even don't have

offl:
>The fact is in lots of cases target selection is weird enough to not think it's not automated or target is somehow inherited not by our SEND_TARGET action which they even don't have

I forgot about https://wow.tools/dbc/?dbc=enumeratedstring&build=2.5.1.39170#page=1&colFilter[1]=tap%20list&colFilter[4]=1158
But 166+ are Wotlk+, and the way that tap list is used is probably not how we can imagine

meji46:
I've changed the logic so that it doesn't affect the previous behavior. Only the smart_event invoker will be used as invoker for SMART_ACTION_SET_DATA if the object executing the SAI isn't a creature (e.g. gameobject or areatrigger).

BAndysc:
I don't think creating a different logic for actions depending on a type of script base object is making it easier to understand. It is not _that_ bad, because it is at least predictable (creature, creature timed actionlist: always pass `me`, other: always pass `invoker`), but still it is quite subtle and can be confusing and easy to forget about.

Imho it should be always `me` (or null in case of other types) and if you want to pass an invoker, just use `SMART_ACTION_SEND_TARGET_TO_TARGET`. But that's just my opinion

Shauren:
I agree with @BAndysc. the change to SMART_ACTION_SET_DATA will not be merged but the crashfix part of the PR will

meji46:
> I agree with @BAndysc. the change to SMART_ACTION_SET_DATA will not be merged but the crashfix part of the PR will

SQL fixes and PR  updated

PR #27348, opened 2021-11-30
Core/Misc: Verify LFG role selections by class

ihm-tswow:
**Changes proposed:**

- Silently filter out invalid LFG role selections based on player classes (warriors should not be able queue as healers etc.)
- Cause the default "did not select any roles" message to display if a player selects only invalid roles

**Issues addressed:**
Currently, players can select whatever roles they want in LFG. Since this concerns an exploit, though a very minor one, I won't post the steps to reproduce it here. Maintainers may contact me if you need help setting it up.

image

**Tests performed:**
Tested the following in-game:
- Joining LFG with every class individually, checking for both false positives/negatives. All works as expected.
- Joining LFG with a group, selecting valid/invalid pairs with both party leader and other characters. All works as expected.
- Re-queuing with a group and selecting valid/invalid pairs. All works as expected.

- [ ] Decide if invalid selections should really be silent.
- [ ] Someone more familiar with LFGMgr look at if I covered all possible places this can happen.

### DIFF
--- a/src/server/game/DungeonFinding/LFGMgr.cpp
+++ b/src/server/game/DungeonFinding/LFGMgr.cpp
@@ -1872,8 +1872,10 @@ uint8 LFGMgr::FilterClassRoles(Player* player, uint8 roles)
 {
     // check classes for permitted roles
     roles &= PLAYER_ROLE_ANY;
-    if (!(LfgRoleClasses::TANK & player->GetClassMask())) roles &= ~PLAYER_ROLE_TANK;
-    if (!(LfgRoleClasses::HEALER & player->GetClassMask())) roles &= ~PLAYER_ROLE_HEALER;
+    if (!(LfgRoleClasses::TANK & player->GetClassMask()))
+        roles &= ~PLAYER_ROLE_TANK;
+    if (!(LfgRoleClasses::HEALER & player->GetClassMask()))
+        roles &= ~PLAYER_ROLE_HEALER;
     return roles;
 }
 
--- a/src/server/game/DungeonFinding/LFGMgr.cpp
+++ b/src/server/game/DungeonFinding/LFGMgr.cpp
@@ -401,8 +401,6 @@ void LFGMgr::JoinLfg(Player* player, uint8 roles, LfgDungeonSet& dungeons, const
 
     // Sanitize input roles
     roles &= PLAYER_ROLE_ANY;
-
-    // check classes for permitted roles
     roles = FilterClassRoles(player, roles);
 
     // At least 1 role must be selected
@@ -1871,7 +1869,6 @@ uint8 LFGMgr::GetTeam(ObjectGuid guid)
 
 uint8 LFGMgr::FilterClassRoles(Player* player, uint8 roles)
 {
-    // check classes for permitted roles
     roles &= PLAYER_ROLE_ANY;
     if (!(LfgRoleClasses::TANK & player->GetClassMask()))
         roles &= ~PLAYER_ROLE_TANK;
--- a/src/server/game/DungeonFinding/LFGMgr.cpp
+++ b/src/server/game/DungeonFinding/LFGMgr.cpp
@@ -399,12 +399,15 @@ void LFGMgr::JoinLfg(Player* player, uint8 roles, LfgDungeonSet& dungeons, const
     if (!player || !player->GetSession() || dungeons.empty())
         return;
 
+    // Sanitize input roles
+    roles &= PLAYER_ROLE_ANY;
+
+    // check classes for permitted roles
+    roles = FilterClassRoles(player, roles);
+
     // At least 1 role must be selected
     if (!(roles & (PLAYER_ROLE_TANK | PLAYER_ROLE_HEALER | PLAYER_ROLE_DAMAGE)))
         return;
-
-    // Sanitize input roles
-    roles &= PLAYER_ROLE_ANY;
 
     Group* grp = player->GetGroup();
     ObjectGuid guid = player->GetGUID();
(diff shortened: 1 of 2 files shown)


src/server/game/DungeonFinding/LFGMgr.cpp
```diff
@@ -710,6 +713,12 @@ void LFGMgr::UpdateRoleCheck(ObjectGuid gguid, ObjectGuid guid /* = ObjectGuid::
     // Sanitize input roles
     roles &= PLAYER_ROLE_ANY;
 
+    if (guid) {
+        Player* player = ObjectAccessor::FindPlayer(guid);
+        if (!player) return;
+        roles = FilterClassRoles(player, roles);
+    }
+
```
MaxtorCoder, line 721:
```suggestion
    if (guid)
    {
        if (Player* player = ObjectAccessor::FindPlayer(guid))
            roles = FilterClassRoles(player, roles);
    }

```

src/server/game/DungeonFinding/LFGMgr.cpp
```diff
@@ -1859,6 +1868,15 @@ uint8 LFGMgr::GetTeam(ObjectGuid guid)
     return team;
 }
 
+uint8 LFGMgr::FilterClassRoles(Player* player, uint8 roles)
+{
+    // check classes for permitted roles
+    roles &= PLAYER_ROLE_ANY;
+    if (!(LfgRoleClasses::TANK & player->GetClassMask())) roles &= ~PLAYER_ROLE_TANK;
+    if (!(LfgRoleClasses::HEALER & player->GetClassMask())) roles &= ~PLAYER_ROLE_HEALER;
```
MaxtorCoder, line 1876:
```suggestion
    if (!(LfgRoleClasses::TANK & player->GetClassMask()))
        roles &= ~PLAYER_ROLE_TANK;
    if (!(LfgRoleClasses::HEALER & player->GetClassMask()))
        roles &= ~PLAYER_ROLE_HEALER;
```

MaxtorCoder (commented):
Some code style changes

src/server/game/DungeonFinding/LFGMgr.cpp
ihm-tswow: This one is functionally different, we need to exit the function if the player doesn't exist, not allow it to continue without filtering

ihm-tswow:
As a small note, cheating players receive a glitched message that they are queued as incompatible classes if they attempt queue with both (warrior queueing as tank+healer are shown to themselves to be queued as tank+healer, but are only really queued as tank). Don't think it's fixable, but wouldn't bother even if it was.

jackpoz:
shouldn't cheating players be kicked ? or their queuing be ignored ?

src/server/game/DungeonFinding/LFGMgr.cpp
jackpoz: still, the formatting needs to be changed for these lines

src/server/game/DungeonFinding/LFGMgr.h
```diff
@@ -128,6 +128,11 @@ enum LfgRoleCheckState
     LFG_ROLECHECK_NO_ROLE                        = 6       // Someone selected no role
 };
 
+enum LfgRoleClasses {
+    TANK   = 0b10000100011, // warrior, paladin, dk, druid
```
jackpoz, line 132: is there a way for a normal person to know what does this number mean ? some other enum defined somewhere with different classes ?

src/server/game/DungeonFinding/LFGMgr.cpp
ihm-tswow: sorry, completely forgot about this pull. i'll fix it tomorrow!

src/server/game/DungeonFinding/LFGMgr.h
ihm-tswow: Good point, i'll do `(1<<(CLASS_WARRIOR-1)) | 1<<(CLASS_PALADIN-1)) ...`. Don't think there's any mask-ready enum, and that's the style used by other `SharedDefines.h` stuff.

ihm-tswow:
> shouldn't cheating players be kicked ? or their queuing be ignored ?

_shrug_. As it works currently, it will ignore queuing if the player does not select _any_ valid roles and otherwise just remove the ones that are invalid. I can change that into a full comparison and ignore/kick depending on how you want to handle it.

For some context, tswow will soon be supporting mods that change clients to do this on purpose for custom servers, so I assume some clients might start cheating this on accident if they connect to normal trinitycore servers using such a client. Because of how the mod works, i think it's also more likely to happen on accident than other types of unintentional cheating.

jackpoz:
We can skip changing anything more on this PR but usually we add a config if we want to allow changing a behaviour and we add a log just in case. We could postpone the discussion for another PR and merge this the way it is.

jackpoz:
@ihm-tswow anything left before we can merge this ? did you do another round of tests after latest changes ?

src/server/game/DungeonFinding/LFGMgr.cpp
```diff
@@ -710,6 +711,13 @@ void LFGMgr::UpdateRoleCheck(ObjectGuid gguid, ObjectGuid guid /* = ObjectGuid::
     // Sanitize input roles
     roles &= PLAYER_ROLE_ANY;
 
+    if (guid) {
+        if (Player* player = ObjectAccessor::FindPlayer(guid))
+            roles = FilterClassRoles(player, roles);
+        else
+            return;
+    }
```
jackpoz, line 719:
```suggestion
    if (guid)
    {
        if (Player* player = ObjectAccessor::FindPlayer(guid))
            roles = FilterClassRoles(player, roles);
        else
            return;
    }
```

ihm-tswow:
@jackpoz I did test the latest changes, but because of how central the system is i'd suggest someone else also try it to make sure.

jackpoz:
Thanks for the PR 🚀 😃

Rushor:
nice thank you!

PR #27352, opened 2021-12-01
Core/Objects: Added logic for PlayHoverAnim flag

meji46:
**Changes proposed:**

-  Set PlayHoverAnim flag to true for units that are flying or hovering and aren't rooted.

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


Shauren:
I don't agree with this, play hover anim is supposed to be possible to toggle separately, see 7e779a6f5955de96dc1baa2f48b1bb15ad9bc343

PR #27363, opened 2021-12-04
[Core/Spells]: Make it possible for spells with motion, that the hit hooks are delayed until motion done

Seyden:
**Changes proposed:**

-  ~~Idea 1: Changed JumpArrivalCast to JumpArrivalActions due to non standarized handling by blizzard (delaying cast or casting it instantly by impact)~~
-  ~~Idea 2: Implement a boolean to check if the spell has to get delayed until the movement is finished, this will cause the spell to fall into the SPELL_STATE_DELAYED state and will also be keeped alive until the movement generator handles the delay.
With this idea we dont need JumpArrivalCast or JumpArrivalActions, we can simply use OnEffectHit and OnEffectHitTarget Hooks. Its worth saying that this will cause the whole spell to delay the Hit Phase~~
-  

**Issues addressed:**

**Tests performed:**
- Builds
- Tested ingame with a custom script for metamorphosis

- [ ] 
- [ ]


src/server/game/Spells/SpellEffects.cpp
```diff
@@ -5896,7 +5935,16 @@ void Spell::EffectJumpCharge()
             effectExtra->ParabolicCurveId = *params->ParabolicCurveId;
     }
 
-    unitCaster->GetMotionMaster()->MoveJumpWithGravity(*destTarget, speed, params->JumpGravity, EVENT_JUMP, false, arrivalCast.get_ptr(), effectExtra.get_ptr());
+    bool hasMovementStarted = unitCaster->GetMotionMaster()->MoveJumpWithGravity(*destTarget, speed, params->JumpGravity, EVENT_JUMP, false, arrivalActions, effectExtra.get_ptr());
+    if (hasMovementStarted)
+        return;
+
+    // In case the motion master didn't start, handle the actions immediatley
```
Nyr97, line 5942: Spelling mistake :warning:

src/server/game/Spells/SpellScript.h
```diff
@@ -142,6 +142,9 @@ class TC_GAME_API _SpellScript
         // Function called when script is destroyed
         // use for: deallocating memory allocated by script
         virtual void Unload() { }
+        // Function called when spell effects with JumpArrivalAction are executed
+        // use for: create your own lambda actions which are called upon finalizing the movement
```
Nyr97, line 146: Use for: creating your own... :warning:

Seyden: that was just the format that the other methods were using, i think the "use for" is more like a lable, so i reused it

Nyr97: Right above virtual void Unload() { } you can see the option used is "use for: deallocat**ing**" :eyes:

PR #27414, opened 2021-12-16
Core/Battlegrounds: Avoid teleport players when relogin inside Strand of the Ancients

Jildor:
**Changes proposed:**

-  As I commented here: #27280
I have noticed that in SotA when you relogin you are teleported, you shouldn't be teleported every time you relogin inside Strand of the Ancients.

**Issues addressed:**

None

**Tests performed:**

Tested in-game


jackpoz:
Thanks for the PR 🚀 😃

PR #27435, opened 2021-12-19
Core/Creature: Added Helper method for Add or Disable Encounter frame…

BertiRean:
…s Engage

**Changes proposed:**

-  Use a helper method instead making a if in each case where you need add/remove an Encounter frame
-  
-  

**Issues addressed:**

**Tests performed:**

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -3122,6 +3122,18 @@ float Creature::GetAggroRange(Unit const* target) const
     return 0.0f;
 }
 
+void Creature::AddInstanceEncounterFrame(uint8 priority /*= 0*/)
+{
+    if (InstanceScript* instance = GetInstanceScript())
+        instance->SendEncounterUnit(ENCOUNTER_FRAME_ENGAGE, this, priority);
+}
+
+void Creature::RemoveInstanceEncounterFrame(uint8 priority /*= 0*/)
+{
+    if (InstanceScript* instance = GetInstanceScript())
+        instance->SendEncounterUnit(ENCOUNTER_FRAME_DISENGAGE, this, priority);
+}
+
 Unit* Creature::SelectNearestHostileUnitInAggroRange(bool useLOS) const
 {
     // Selects nearest hostile target within creature's aggro range. Used primarily by
--- a/src/server/game/Entities/Creature/Creature.h
+++ b/src/server/game/Entities/Creature/Creature.h
@@ -241,6 +241,9 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         float GetAttackDistance(Unit const* player) const;
         float GetAggroRange(Unit const* target) const;
 
+        void AddInstanceEncounterFrame(uint8 priority = 0);
+        void RemoveInstanceEncounterFrame(uint8 priority = 0);
+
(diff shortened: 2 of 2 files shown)


Ovahlord:
at this point you can just get rid of the instance script depency and send plain packets instead. Since open world bosses use it as well (since MoP?).

BertiRean:
> at this point you can just get rid of the instance script depency and send plain packets instead. Since open world bosses use it as well (since MoP?).

Well that could be done also, maybe @Shauren have a better opinion about if we need remove or not the instance dependency

Shauren:
I don't see the value in this PR. Since all (new) boss scripts are supposed to use BossAI, they already have a InstanceScript* member (so you dont need to check it), all this is going to do is change 1 line (`instance->SendEncounterUnit(ENCOUNTER_FRAME_ENGAGE, me, priority);` with another 1 line (`me->AddInstanceEncounterFrame(priority)`

BertiRean:
> 

Well for BossAI you can use the instance pointer on structure, but for scripts that are normally like trash or something. If you don't want make a InstanceScirpt* as struct member in each script, this method save a lot of time and avoid write InstanceScript _instance in each script of normal mob or boss add.

Shauren:
Right, it looks like you missed/forgot one more thing
https://github.com/TrinityCore/TrinityCore/blob/34ae77ca8685ecc0252f580a184d6f37277f33be/src/server/scripts/Northrend/IcecrownCitadel/icecrown_citadel.h#L563-L569

When you use functions like that me->GetInstanceScript() is guaranteed to return something valid (eliminating the need to check it, just like in BossAI)

BertiRean:
I've attached a video about how Serpentrix on Eye of Azshara show different frames for each head.

Shauren:
Yeah but my latest comment still applies to scripted & summoned adds

Aokromes:
Conflicting files
src/server/game/Entities/Creature/Creature.cpp

mdX7:
> at this point you can just get rid of the instance script depency and send plain packets instead. Since open world bosses use it as well (since MoP?).

Cannot confirm. Tested @ 167526 & 178958 (Shadowlands worldbosses) and 83746 (WoD worldboss)

BertiRean:
> Yeah but my latest comment still applies to scripted & summoned adds

But then the method should be on ScriptedAI or CreatureAI?. Because BossAI is only used by the bosses.

Nyr97:
> > at this point you can just get rid of the instance script depency and send plain packets instead. Since open world bosses use it as well (since MoP?).
> 
> Cannot confirm. Tested @ 167526 & 178958 (Shadowlands worldbosses) and 83746 (WoD worldboss)

@Seyden also couldn't confirm this as Galion (MoP worldboss) does not send it either.

Ovahlord:
Doolmlord Kazzak in Taanan Jungle clearly does have the frame and he is a world boss.

ImmortalusDev:
> Doolmlord Kazzak in Taanan Jungle clearly does have the frame and he is a world boss.

Where do you saw that?

mdX7:
Also Kazzak in Tanaan Jungle is not having that
<details>
<summary>Ref</summary>

grafik

</details>

ImmortalusDev:
It was always related only to instanced creatures, bosses and it's summons, nothing else.

PR #27452, opened 2021-12-22
Core/SAI: Remove SMART_ACTION_SEND_GO_CUSTOM_ANIM, make SMART_ACTION_ADD_AURA & SMART_ACTION_SET_GO_FLAG deprecated

offl:
**Changes proposed:**

-  Remove SMART_ACTION_SEND_GO_CUSTOM_ANIM, it was replaced with gameobject actions
-  SMART_ACTION_ADD_AURA is now deprecated, reasons in e96c713815d1d129db31e60a98e2ff069795b9a7
-  SMART_ACTION_SET_GO_FLAG too deprecated, it was replaced with gameobject actions, ADD and REMOVE too were replaced but will stay for a while just in case

**Issues addressed:**

none

**Tests performed:**

none

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -1791,13 +1791,6 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                     target->ToUnit()->InterruptNonMeleeSpells(e.action.interruptSpellCasting.withDelayed != 0, e.action.interruptSpellCasting.spell_id, e.action.interruptSpellCasting.withInstant != 0);
             break;
         }
-        case SMART_ACTION_SEND_GO_CUSTOM_ANIM:
-        {
-            for (WorldObject* target : targets)
-                if (IsGameObject(target))
-                    target->ToGameObject()->SendCustomAnim(e.action.sendGoCustomAnim.anim);
-            break;
-        }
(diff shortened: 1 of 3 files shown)


Shauren:
Where is the sql converting SMART_ACTION_SEND_GO_CUSTOM_ANIM to SMART_ACTION_ACTIVATE_GAMEOBJECT? (a generic one, not dependent on entryorguid+id)

offl:
https://github.com/TrinityCore/TrinityCore/commit/8516984ffe1a0241b5cfc47f2d0124291503aec1

Shauren:
Thats "wrong", it must be in this PR, not be pushed earlier (unlikely but not impossible)

offl:
Well, people wanted me to make them deprecated first and remove only after some time
If you mean I need to create a sql, create a warning-commit that in 2 weeks it will be removed, and only after 2 weeks push sql and remove action from core, then I would rather do everything without any warning, because it's not even possible to hold such changes for such a long time without pushing them

Any way that process gives nothing and I'm only wasting time on deprecating stuff and then removing. People who have custom scripts should check what they push, people who cherry pick stuff do it differently and in general lots of things simply are not cherry-pickable without investing time in that. Moreover some of them are not easily cherry-pickable for master because master may have scripts which 3.3.5 does not have

Shauren:
You just didn't need to immediately push the sql together with marking action as deprecated

offl:
Ok

PR #27456, opened 2021-12-23
Shaman cleanup spells

aquadeus:
**Changes proposed:**

-  Fix earthen rage, remove obsolete spells and cleanup

**Tests performed:**

Does it build and tested in-game


Shauren:
Rejected, Do not mix refactors and fixes in a single PR

PR #27461, opened 2021-12-24
Core/SAI: Use GetBaseObject as second param of SMART_ACTION_ACTIVATE_GAMEOBJECT & reorder GameObjectActions

offl:
**Changes proposed:**

-  Use GetBaseObject as second param of SMART_ACTION_ACTIVATE_GAMEOBJECT to make actions with unitCaster param work
-  Reorder GameObjectActions. I **clearly** understand why they were implemented like that but it only makes work harder. They has specific order which helps to understand how exactly they should be used

**Issues addressed:**

none

**Tests performed:**

none

- [ ] maybe check if GetBaseObject is null

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -1438,6 +1438,9 @@ void GameObject::ActivateObject(GameObjectActions action, WorldObject* spellCast
 
     switch (action)
     {
+        case GameObjectActions::None:
+            TC_LOG_FATAL("spell", "Spell %d has action type NONE in effect %d", spellId, effectIndex);
+            break;
         case GameObjectActions::AnimateCustom0:
         case GameObjectActions::AnimateCustom1:
         case GameObjectActions::AnimateCustom2:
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -1438,29 +1438,50 @@ void GameObject::ActivateObject(GameObjectActions action, WorldObject* spellCast
...
             break;
         case GameObjectActions::OpenAndUnlock:
             if (unitCaster)
+            {
                 UseDoorOrButton(0, false, unitCaster);
-            [[fallthrough]];
-        case GameObjectActions::Unlock:
-        case GameObjectActions::Lock:
-            ApplyModFlag(GAMEOBJECT_FLAGS, GO_FLAG_LOCKED, action == GameObjectActions::Lock);
+                RemoveFlag(GAMEOBJECT_FLAGS, GO_FLAG_LOCKED);
```
jackpoz, line 1466: Is RemoveFlag working better in this case than ApplyModFlag ?

offl: It's single case now

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -1496,8 +1513,8 @@ void GameObject::ActivateObject(GameObjectActions action, WorldObject* spellCast
 
             break;
         }
-        case GameObjectActions::None:
-            TC_LOG_FATAL("spell", "Spell %d has action type NONE in effect %d", spellId, effectIndex);
+        case GameObjectActions::SetTapList:
+            // No use cases, implementation unknown
```
Shauren, line 1517: This is supposed to restrict who can loot a chest, for example with boss loot

offl: There also Set Tapped to Challenge Players

Lots of things in GameObject::ActivateObject depends on how [step 4](https://github.com/TrinityCore/TrinityCore/issues/21078#issuecomment-420115205) should be implemented, I guess

jackpoz:
>Known issues and TODO list:
 maybe check if GetBaseObject is null

just for information, if GetBaseObject is null not crash will happen.

PR #27490, opened 2021-12-29
Make Position::HasInLine really check if the target pos is in line an…

Seyden:
…d not if its in range in the cone

**Changes proposed:**

-  Currently Position::HasInLine does check if the target position is in a distance less than width IN a line, this is not correct because we only want to know if the target position is in LINE with an infinite y axis distance
-  
-  

**Issues addressed:**

**Tests performed:**

- [ ] 
- [ ]


Seyden:
Currently HasInLine is using width to calculate the distance within the line, which shouldn't be the job of this function

image

Seyden:
(Not valid) Wow coordination system swaps x and y, therefore sinus is correct here

PR #27503, opened 2022-01-01
MMAP: Build ADT floor just like WMO floor below liquid

killerwife:
Fixes ocean floor pathfinding

To test, go to for example ocean south of booty bay and look through .mmap loc or recastdemo

image

image

Before and after images supplied (cmangos recastdemo)


jackpoz:
Thanks for the PR 🚀 😃

Issue #27527, opened 2022-01-05
Core/Quest - incorrect answer when sharing a quest to a group.


kvolk2git:
### Description

If you are trying to share a quest in a group and a player who is being offered a quest has already completed this quest and passed as a result, you will receive a message that the player already has this task instead of a message that the task has already been completed.

### Expected behaviour

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Handlers/QuestHandler.cpp

This code snippet does not work as intended

        if (!receiver->SatisfyQuestStatus(quest, false))
        {
            sender->SendPushToPartyResponse(receiver, QUEST_PARTY_MSG_HAVE_QUEST);
            continue;
        }

        if (receiver->GetQuestStatus(questId) == QUEST_STATUS_COMPLETE)
        {
            sender->SendPushToPartyResponse(receiver, QUEST_PARTY_MSG_FINISH_QUEST);
            continue;
        }

### Steps to reproduce the problem

1. create 2 character
2. take the some quest 
3. share this quest and look answer
4. on first character complete and pass the quest
5. on second character again share quest and look answer

### Branch

3.3.5

### TC rev. hash/commit

e470a23a2f09

### Operating system

any

### Custom changes

none

kvolk2git:
I started to deal with the code and came to the conclusion that the Player::SatisfyQuestStatus is not working properly.
In the file https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Quests/QuestDef.h possible quest statuses are described:

enum QuestStatus : uint8
{
    QUEST_STATUS_NONE           = 0,
    QUEST_STATUS_COMPLETE       = 1,
    //QUEST_STATUS_UNAVAILABLE    = 2,
    QUEST_STATUS_INCOMPLETE     = 3,
    //QUEST_STATUS_AVAILABLE      = 4,
    QUEST_STATUS_FAILED         = 5,
    QUEST_STATUS_REWARDED       = 6,        // Not used in DB
    MAX_QUEST_STATUS
};

In the file https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Player/Player.cpp the code snippet will return false:
bool Player::SatisfyQuestStatus(Quest const* qInfo, bool msg) const
{
    if (GetQuestStatus(qInfo->GetQuestId()) == QUEST_STATUS_REWARDED)
    {
        if (msg)
        {
[...]

PR #27528, opened 2022-01-05
Core/Quest - fix incorrect answer when sharing

kvolk2git:
**Changes proposed:**

-  fix incorrect answer when sharing quest anв quest is complete from other players

**Issues addressed:**

Closes #  (#27527)


kvolk2git:
incorrect design

PR #27539, opened 2022-01-06
Core/Auras: Implement SPELL_AURA_MOD_BONUS_ARMOR (465)

aquadeus:
**Changes proposed:**

-  Implement SPELL_AURA_MOD_BONUS_ARMOR

**Tests performed:**

Tested in-game


src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -3537,6 +3537,17 @@ void AuraEffect::HandleModArmorPctFromStat(AuraApplication const* aurApp, uint8
...
 }
 
+void AuraEffect::HandleModStatBonusArmor(AuraApplication const* aurApp, uint8 mode, bool apply) const
+{
+    if (!(mode & (AURA_EFFECT_HANDLE_CHANGE_AMOUNT_MASK | AURA_EFFECT_HANDLE_STAT)))
+        return;
+
+    Unit* target = aurApp->GetTarget();
+
+    if (target->IsPlayer() || target->IsPet())
```
mdX7, line 3547: im quite sure regular units should also be affected by it since they also have armor. I actually scrolled thru the spells with this aura and kinda only saw things which only players triggers however

aquadeus: could be, but spells like 294161 are not casted by pets? or for pets?

aquadeus: solved

PR #27541, opened 2022-01-06
spell 37135 shouldn't control all players

w5860363:
https://tbc.wowhead.com/npc=20045/nether-scryer
spell 37135 https://tbc.wowhead.com/spell=37135/domination
There is only one player in the skill description


w5860363:
@offl He now performs as a whole. He still controls all 25 accounts in the test
Even if there is a group control effect, it should not include all healing and tanks

offl:
Should control 5 random then

PR #27583, opened 2022-01-09
Core/Conditions: Update phases when quest objective is updated

mdX7:
**Tests performed:**

tested ingame

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -17143,6 +17143,8 @@ void Player::UpdateQuestObjectiveProgress(QuestObjectiveType objectiveType, int3
 
     if (anyObjectiveChangedCompletionState)
         UpdateVisibleGameobjectsOrSpellClicks();
+
+    PhasingHandler::OnConditionChange(this);
 }
 
 bool Player::HasQuestForItem(uint32 itemid) const


Shauren:
I don't doubt that this works but what about quests that don't immediately update phases and instead rely on later spellcast (or something like that) to update phases?

mdX7:
for that i'd assume the phase condition wouldnt be about the objective itself but something else (i also don't have a particular quest on my mind for that case though)

mdX7:
okay it seems no spell is cast or something, but credit is added after selecting gossip option (quest is https://www.wowhead.com/quest=57584/a-fate-most-noble btw)
After completing the `Speak to Greeter Mnemis` objective the phase is added.
```
ClientToServer: CMSG_GOSSIP_SELECT_OPTION (0x349A) Length: 24 ConnIdx: 1 Time: 01/13/2022 22:43:33.378 Number: 4548
ServerToClient: SMSG_PHASE_SHIFT_CHANGE (0x2578) Length: 91 ConnIdx: 1 Time: 01/13/2022 22:43:33.467 Number: 4555
ServerToClient: SMSG_QUEST_UPDATE_ADD_CREDIT (0x2A8C) Length: 28 ConnIdx: 1 Time: 01/13/2022 22:43:33.467 Number: 4556

```

PR #27590, opened 2022-01-10
Core/Quest: Remove RewardMoney from quest_template and send correct data

funjoker:
**Changes proposed:**

-  Remove static RewardMoney
-  Use correct values in packet

**Issues addressed:**

**Tests performed:**
It builds

- [ ] 
- [ ]


src/server/game/Entities/Creature/GossipDef.cpp
```diff
@@ -474,10 +474,14 @@ void PlayerMenu::SendQuestGiverQuestDetails(Quest const* quest, ObjectGuid npcGU
 void PlayerMenu::SendQuestQueryResponse(Quest const* quest) const
 {
     if (sWorld->getBoolConfig(CONFIG_CACHE_DATA_QUERIES))
-        _session->SendPacket(&quest->QueryData[static_cast<uint32>(_session->GetSessionDbLocaleIndex())]);
+    {
+        WorldPackets::Quest::QueryQuestInfoResponse* queryPacket = (WorldPackets::Quest::QueryQuestInfoResponse*)&quest->QueryData[static_cast<uint32>(_session->GetSessionDbLocaleIndex())];
+        queryPacket->Info.RewardMoney = _session->GetPlayer()->GetQuestMoneyReward(quest);
+        _session->SendPacket(queryPacket->Write());
```
Shauren, line 480: Very much thread unsafe and completely removes the point of this cache - its fine if we just send "as if at max level for quest" value here when cache is enabled

Shauren (changes requested):

src/server/game/Quests/QuestDef.cpp
```diff
@@ -317,12 +316,24 @@ uint32 Quest::XPValue(Player const* player) const
...
+uint32 Quest::MoneyValue(Player const* player, bool maxValue) const
 {
+    uint32 value = 0;
+
+    if (maxValue)
+        if (Optional<ContentTuningLevels> questLevels = sDB2Manager.GetContentTuningData(GetContentTuningId(), 0))
+            if (QuestMoneyRewardEntry const* money = sQuestMoneyRewardStore.LookupEntry(questLevels->MaxLevel))
+                value = money->Difficulty[GetRewMoneyDifficulty()] * GetMoneyMultiplier();
+
     if (QuestMoneyRewardEntry const* money = sQuestMoneyRewardStore.LookupEntry(player->GetQuestLevel(this)))
```
funjoker, line 328: maybe make it else if ?

Shauren: Not a maybe, definitely (this crashes)

PR #27593, opened 2022-01-11
Core/Player: Added PlayerCreateMode field to playercreateinfo_cast_sp…

mdX7:
…ell to only cast spells based on related mode

**Tests performed:**

built, tested ingame with undead in exiles reach and deathknell
(+ also used 295600 spell for exiles reach)


src/server/game/Globals/ObjectMgr.h
```diff
@@ -627,6 +628,7 @@ struct PlayerLevelInfo
 };
 
 typedef std::vector<uint32> PlayerCreateInfoSpells;
+typedef std::unordered_map<PlayerCreateMode, PlayerCreateInfoSpells> PlayerCreateInfoSpellsMap;
```
Shauren, line 631: no, just make that a plain array

Shauren (changes requested):

src/server/game/Globals/ObjectMgr.h
mdX7: done

Shauren (approved):

Issue #27540, opened 2022-01-06
Core/Boss AI  BlackTemple error spellid


w5860363:
### Description

https://tbc.wowhead.com/spell=40810/saber-lash#see-also-other

In the old version of SD2, the spell ID was 40810

40816 in the new version of the code

The new version cannot be allocated damage in any form

Refer to previous version

https://github.com/samaelsacred/TrinityCore434/blob/6de6c751ac497f5d42893170d0b0a06803859cf8/src/server/scripts/Outland/BlackTemple/boss_mother_shahraz.cpp
https://github.com/Falstring/Dalaran-Core/blob/3605a709de0b6b73c05dcb0bd2eb772d7b0bcf86/src/server/scripts/Outland/BlackTemple/boss_mother_shahraz.cpp

### Expected behaviour

Spell damage can be shared

### Steps to reproduce the problem

1.two player go cr 52760
2.Overlapping station
3.Observe the position of one person after injury
4.The damage remains the same

### Branch

3.3.5

### TC rev. hash/commit

https://github.com/TrinityCore/TrinityCore/commit/32cef906b02562ffb129b095386a9314abde82f1

### Operating system

Windows 10 x64,

### Custom changes

None

w5860363:
boss_warlord_najentus.cpp spell 39835 The spell speed is very different from the original SD2

PR #27594, opened 2022-01-11
Core/AI boss_mother_shahraz

w5860363:
.go xyz 980.47 164.25 192.84 564 
Nearby will be out of combat, so you should avoid transmitting players to nearby areas

The spell cast by the health check is wrong

This aura causes the boss to cast melee spell once a second


w5860363:
/src/server/scripts/Outland/BlackTemple/instance_black_temple.cpp#L54
The safety zone is set, but the wall is not in the safety zone

offl:
In sniff SPELL_RANDOM_PERIODIC is not used on aggro, but on 10% HP (old behavior is correct); SPELL_FRENZY is not used at all

Saber Lash procs from aura 40816 which always is applied on boss (old behavior is correct). You're testing boss without tank with tank gear and it procs on every boss melee attack. There's problem with divided damage and probably spell should have hidden cooldown in spell_proc. I'll open an issue later about that stuff

After checking videos & sniffs almost sure boss teleports players to random positions around boss (like currently (GetCaster()->GetRandomNearPosition(50.0f))
Even if she teleports players to predefined positions, that means you need way much more positions and chose only the ones which are close to boss, otherwise she will teleport players way too far(that does not happen in videos & sniffs). It's simply too complicated for Blizz to be true

>https://github.com/TrinityCore/TrinityCore/issues/27540

Comparing TC's BT which was fully updated from sniffs with another projects more likely will revert correct changes than really fix what is needed to be fixed. There are lots of things which can be updated or handled differently in BT but definitely not like that.

w5860363:
@offl https://tbc.wowhead.com/spell=40816/saber-lash have a 7 second cooldown

https://wowwiki-archive.fandom.com/wiki/Mother_Shahraz
([Saber Lash]—Very frequent frontal cleave, 20 yard range, ~30k damage on plate, divided between up to three targets, provides immunity to Fatal Attraction for 30 sec
)

 [Enrage]—at 10% she periodically increases damage and attack speed for a short time. After 10 minutes, she enrages permanently.
---------------------------------
Then refresh the immune time if it is triggered every time

Enrage Will trigger at 10%
-----------
As for GetRandomNearPosition, it's only because BossBoundaryData is triggered near the wall.
test .go xyz 980.47 164.25 192.84 564

offl:
Yes, 7 seconds is literally what I see in https://tbc.warcraftlogs.com/reports/Ax7rW18VY9tLZzFX#fight=20&view=events&type=damage-done&hostility=1

>enrage

It's not an hidden aura and can be seen in sniffs or videos but I don't see it

w5860363:
@offl https://www.youtube.com/watch?v=6dvLGBxe-ro Video before 13year 2:38-2:40
Maybe Blizzard retail has changed the display method of aura

Issue #27562, opened 2022-01-07
Core/AI illidari_council combat error


w5860363:
### Description

 boss Can not be killed by any form, in time to use .die they will also come back to life

### Expected behaviour

Normal end of the battle and the appearance of Akama

### Steps to reproduce the problem

1.go cr 52764
2.dam 10000000

### Branch

3.3.5

### TC rev. hash/commit

https://github.com/TrinityCore/TrinityCore/commit/5fc650645d9b96f7298c4c11a6a1941d629ae601

### Operating system

Windows 10 x64,

### Custom changes

None

TheEvilPotato:
Still present as of https://github.com/TrinityCore/TrinityCore/commit/9ef60d70638888c07d131d42d3746478feac2423, they also seem to despawn if they kill all players, they only return if you reboot your worldserver.exe

w5860363:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

acidmanifesto:
> 这是来自QQ邮箱的假期自动回复邮件。
>  
> 您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

Translation:
This is a holiday auto-reply email from QQ mailbox. Hi, I am currently on vacation and cannot respond to your emails in person. I will get back to you as soon as possible after the holiday.

PR #27595, opened 2022-01-11
Try to solve the problem that the boss can not be killed and does not die together

w5860363:
Try to solve https://github.com/TrinityCore/TrinityCore/issues/27562

But due to another bug https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Outland/BlackTemple/boss_illidari_council.cpp# L554 BOSS is constantly regaining health

Not yet completed

The spell will still reset creature health indefinitely

### DIFF
--- a/src/server/scripts/Outland/BlackTemple/boss_illidari_council.cpp
+++ b/src/server/scripts/Outland/BlackTemple/boss_illidari_council.cpp
@@ -176,17 +176,8 @@ struct boss_illidari_council : public BossAI
     {
         _inCombat = false;
         events.Reset();
+        
         instance->SetBossState(DATA_ILLIDARI_COUNCIL, DONE);
-
-        for (uint32 bossData : CouncilData)
-        {
-            if (Creature* council = instance->GetCreature(bossData))
-            {
-                // Allow loot
-                council->LowerPlayerDamageReq(council->GetMaxHealth());
-                council->CastSpell(council, SPELL_QUIET_SUICIDE, true);
-            }
-        }
     }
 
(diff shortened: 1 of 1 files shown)


Keader:
It looks a hack to me.
If i remember right, they should stay in "live" and die with SPELL_QUIET_SUICIDE
If they're not dying with it, need check why.

src/server/scripts/Outland/BlackTemple/boss_illidari_council.cpp
```diff
@@ -268,8 +268,18 @@ struct IllidariCouncilBossAI : public BossAI
...
-            damage = me->GetHealth() - 1;
+        if (who == me)
+            return;
+
+        for (uint32 bossData : CouncilData)
+        {
+            if (Creature* council = instance->GetCreature(bossData))
+                if (council != me && damage < council->GetHealth())
+                {
+                    council->ModifyHealth(-int32(damage));
```
Keader, line 279: hack, they should die for SPELL_QUIET_SUICIDE

Keader (changes requested):
hack 
please check why they're not dying for SPELL_QUIET_SUICIDE

w5860363:
@Keader This is just an attempt. There are many problems at present

Fighting can't be balanced. Their health value is constantly restored to full, so I think this is the reason why they can't die of suicide

Keader:
Some core change caused it, since it was working when was scripted (and script dont change)
check why they're restoring life and why quiet suicide is not killing too.

PR #27604, opened 2022-01-12
Scripts/Baradin Hold: fixed a crash in Occuthar's Focused Fire target selection

Ovahlord:
**Changes proposed:**

-  reworked the target selection for Occu'thar's Focused Fire to modernize the script and to fix a crash that was caused by storing a nullptr target, which made the core blow up since it expects a fully valid target list at this point.
- Also allow Occu'thar to actually hit his victim when no other target option is available

**Tests performed:**
- tested on 4.x

### DIFF
--- a/src/server/scripts/EasternKingdoms/BaradinHold/boss_occuthar.cpp
+++ b/src/server/scripts/EasternKingdoms/BaradinHold/boss_occuthar.cpp
@@ -266,9 +266,7 @@ class spell_occuthar_eyes_of_occuthar : public SpellScriptLoader
                 if (targets.empty())
                     return;
 
-                WorldObject* target = Trinity::Containers::SelectRandomContainerElement(targets);
-                targets.clear();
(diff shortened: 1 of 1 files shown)


src/server/scripts/EasternKingdoms/BaradinHold/boss_occuthar.cpp
```diff
@@ -266,9 +266,7 @@ class spell_occuthar_eyes_of_occuthar : public SpellScriptLoader
                 if (targets.empty())
                     return;
 
-                WorldObject* target = Trinity::Containers::SelectRandomContainerElement(targets);
-                targets.clear();
-                targets.push_back(target);
+                Trinity::Containers::RandomResize(targets, 1);
```
Shauren, line 269: This is a pessimization, not optimization, `RandomResize` always iterates entire container (to support > 1 cases)

PR #27620, opened 2022-01-16
Scripts/Spells: Fix shaman Earthquake spell

Robingadko:
**Changes proposed:**

-  Implement [Earthquake](https://www.wowhead.com/spell=61882/earthquake) scripts and set correct [damage](https://www.wowhead.com/spell=77478/earthquake).
-  Take into account the [Aftershock](https://www.wowhead.com/spell=273221/aftershock) talent and [Bottle of Swirling Maelstrom](https://www.wowhead.com/spell=329580/bottle-of-swirling-maelstrom) buff.

**Issues addressed:**

None.

**Tests performed:**

Tested in-game.

None.


src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -456,6 +462,110 @@ class spell_sha_earthen_rage_proc_aura : public AuraScript
...
+            SPELL_SHAMAN_BOTTLE_OF_SWIRLING_MAELSTROM,
+            SPELL_SHAMAN_EARTHQUAKE_KNOCKING_DOWN
+        });
+    }
+
+    void CalcDamage(SpellEffIndex /*effIndex*/)
+    {
+        const Player* player = GetCaster()->ToPlayer();
+        float versaDmgMod = 1.0f;
+        // $mawPower = ${ 1 + ($329580m5 / 100) }
```
Nyr97, line 534: Are we sure we want to start adding Torghast Anima Powers values in here?

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -456,6 +462,110 @@ class spell_sha_earthen_rage_proc_aura : public AuraScript
...
+
+        uint32 dmg = player->m_activePlayerData->ModDamageDonePos[SPELL_SCHOOL_NATURE] * 0.391f * versaDmgMod;
+
+        if (AuraEffect* aurEff = player->GetAuraEffect(SPELL_SHAMAN_ELEMENTAL_SHAMAN, EFFECT_0))
+        {
+            float dmgMod = 1.0f;
+            dmg *= AddPct(dmgMod, aurEff->GetAmount());
+        }
+
+        if (AuraEffect* aurEff = player->GetAuraEffect(SPELL_SHAMAN_BOTTLE_OF_SWIRLING_MAELSTROM, EFFECT_4))
```
Nyr97, line 546: Are we sure we want to start adding Torghast Anima Powers values in here? Should it not be wiser to make the effect of such aura complement the ticks instead rather than raw calculating on its tick spell?

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -456,6 +462,110 @@ class spell_sha_earthen_rage_proc_aura : public AuraScript
...
+    PrepareAuraScript(spell_sha_earthquake_aura);
+
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_SHAMAN_EARTHQUAKE, SPELL_SHAMAN_EARTHQUAKE_TICK });
+    }
+
+    void HandlePeriodic(AuraEffect const* /*aurEff*/)
+    {
+        std::vector<AreaTrigger*> areaTriggers = GetCaster()->GetAreaTriggers(SPELL_SHAMAN_EARTHQUAKE);
```
mdX7, line 503: possible crash: caster may be null in aurascripts => check if `GetCaster()` is null

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -456,6 +462,110 @@ class spell_sha_earthen_rage_proc_aura : public AuraScript
...
+        ({
+            SPELL_SHAMAN_ELEMENTAL_SHAMAN,
+            SPELL_SHAMAN_BOTTLE_OF_SWIRLING_MAELSTROM,
+            SPELL_SHAMAN_EARTHQUAKE_KNOCKING_DOWN
+        });
+    }
+
+    void CalcDamage(SpellEffIndex /*effIndex*/)
+    {
+        const Player* player = GetCaster()->ToPlayer();
```
mdX7, line 532: possible crash: make sure to check if player is null if you want to cast caster to it, imagine some gm letting npcs cast this spell forcefully to crash the server

src/server/scripts/Spells/spell_shaman.cpp
Robingadko: So, should it be a common rule to always check caster in any aura script?

Shauren: That already is a rule.

src/server/scripts/Spells/spell_shaman.cpp
Robingadko: Yeah, you may be right. I gonna try another solution. Thanks for commenting this.

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -456,6 +460,85 @@ class spell_sha_earthen_rage_proc_aura : public AuraScript
...
+
+// 77478 - Earthquake tick
+class spell_sha_earthquake_tick : public SpellScript
+{
+    PrepareSpellScript(spell_sha_earthquake_tick);
+
+    bool Validate(SpellInfo const* spellInfo) override
+    {
+        if (!spellInfo->GetEffects().empty())
+            const_cast<SpellEffectInfo*>(&spellInfo->GetEffect(EFFECT_0))->BonusCoefficient = 0.391f;
```
Shauren, line 524: This should be in SpellMgr::LoadSpellInfoCorrections

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -456,6 +460,85 @@ class spell_sha_earthen_rage_proc_aura : public AuraScript
...
+    PrepareSpellScript(spell_sha_earthquake);
+
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_SHAMAN_AFTERSHOCK, SPELL_SHAMAN_AFTERSHOCK_ENERGIZE });
+    }
+
+    void HandleAfterCast()
+    {
+        int32 energize = *GetSpell()->GetPowerTypeCostAmount(POWER_MAELSTROM);
```
Shauren, line 475: This is supposed to already be handled by `spell_sha_aftershock`, if it isn't working then it must be investigated why the proc doesn't work

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -456,6 +460,85 @@ class spell_sha_earthen_rage_proc_aura : public AuraScript
...
+    }
+
+    void HandlePeriodic(AuraEffect const* /*aurEff*/)
+    {
+        if (Unit* caster = GetCaster())
+        {
+            std::vector<AreaTrigger*> areaTriggers = caster->GetAreaTriggers(SPELL_SHAMAN_EARTHQUAKE);
+
+            for (AreaTrigger* at : areaTriggers)
+                caster->CastSpell(at->GetPosition(), SPELL_SHAMAN_EARTHQUAKE_TICK, true);
```
Shauren, line 506: This would synchronize all ticks on all earthquakes cast by player (not how it behaves on retail)

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -456,6 +460,85 @@ class spell_sha_earthen_rage_proc_aura : public AuraScript
...
+        if (!spellInfo->GetEffects().empty())
+            const_cast<SpellEffectInfo*>(&spellInfo->GetEffect(EFFECT_0))->BonusCoefficient = 0.391f;
+
+        return ValidateSpellInfo({ SPELL_SHAMAN_EARTHQUAKE_KNOCKING_DOWN });
+    }
+
+    void HandleOnHit()
+    {
+        if (Unit* target = GetHitUnit())
+            if (roll_chance_i(GetEffectInfo(EFFECT_1).BasePoints))
```
Shauren, line 532: Never access `BasePoints`, always go through `CalcValue()`
Also if are accessing EFFECT_1 outside of effect handler bound to that index (meaning you aren't in a function linked to `SpellEffectFn`) then it must be validated that it exists in `Validate`

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -456,6 +460,85 @@ class spell_sha_earthen_rage_proc_aura : public AuraScript
...
+class spell_sha_earthquake_tick : public SpellScript
+{
+    PrepareSpellScript(spell_sha_earthquake_tick);
+
+    bool Validate(SpellInfo const* spellInfo) override
+    {
+        if (!spellInfo->GetEffects().empty())
+            const_cast<SpellEffectInfo*>(&spellInfo->GetEffect(EFFECT_0))->BonusCoefficient = 0.391f;
+
+        return ValidateSpellInfo({ SPELL_SHAMAN_EARTHQUAKE_KNOCKING_DOWN });
```
Shauren, line 526:
```suggestion
        return ValidateSpellInfo({ SPELL_SHAMAN_EARTHQUAKE_KNOCKING_DOWN }) && spellInfo->GetEffects().size() > EFFECT_1;
```

Shauren (changes requested):

Shauren:
I admit this isn't a simple spell to fix

src/server/scripts/Spells/spell_shaman.cpp
Seyden: @Robingadko i would propose that you create a areatrigger script for this case and calculate the tick time inside the areatrigger script and then just let the Caster cast SPELL_SHAMAN_EARTHQUAKE_TICK at the AT position

src/server/scripts/Spells/spell_shaman.cpp
Robingadko: `SetEffectValue` does work but with `OnEffectLaunchTarget`. Will use it.

src/server/scripts/Spells/spell_shaman.cpp
Robingadko: Aftershock cannot proc with triggered spell. Should I correct this in spell_proc table? Cause when Earthquake is cast as non-triggered there are 5 ticks instead of 6.

Shauren: But this is the spell that player directly casts from spellbook, that is never triggered

Robingadko: Sorry, I've confused it with damage [Earthquake](https://www.wowhead.com/spell=77478/earthquake).
Aftershock doesn't proc due to mismatched proc type. It's `PROC_FLAG_DONE_SPELL_MAGIC_DMG_CLASS_POS` for earthquake and `PROC_FLAG_DONE_SPELL_MAGIC_DMG_CLASS_NEG` for aftershock. 
Will see what can be changed to make it work.

PR #27622, opened 2022-01-16
Core/Spells: Fix mage Flame on talent

### LINKED ISSUE #19583: Core/Script The talent Flame On has an issue.
**Description:**

The talent Flame On(ID:205029) given to Mage Fire at lv.60 has an issue

Flame On (passive): Reduces the cooldown of Fire Blast by 2 secconds and increases the maximum number of charges by 1

**Current behaviour:**

When you choose this talent it increase 1 charge of Fire Blast, but it doesn´t reduces the cooldown

**Expected behaviour:**

When you choose this talent it should reduces the cooldown of Fire Blast by 2 secconds and increases the maximum number of charges by 1

1. Create a Mage character
2. Reach lv.59, choose the Fire Spec
3. Check the cooldown of the spell Fire Blast
4. Reach lv.60 and choose the talent Flame On
5. Check the cooldown of the spell Fire Blast

**Branch(es):**

Master

**TC rev. hash/commit:** 

Description in wowhead http://www.wowhead.com/spell=205029/flame-on

aquadeus:
**Changes proposed:**

-  Fix talent Flame On

**Issues addressed:**

Closes #19583

**Tests performed:**

Tested in-game


aquadeus:
Idk what blizzard is doing here, for a 100% correct fix we need a float value here

mdX7:
Blizz indeed is using float values, our systems gotta be adjusted to use floats someday

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -564,6 +564,23 @@ class spell_mage_firestarter_dots : public AuraScript
...
 
+// 205029 - Flame On
+class spell_mage_flame_on : public AuraScript
+{
+   PrepareAuraScript(spell_mage_flame_on);
+
+   void CalculateAmount(AuraEffect const* /*aurEff*/, int32& amount, bool& canBeRecalculated)
+   {
+       canBeRecalculated = false;
+       amount = -16;
```
mdX7, line 575: if its just set to a constant we likely may use SpellMgr for this. Also we may calculate it directly `effect3 basepoints / rechargeTime * 100`

Shauren: Eeeh its more complicated than that.
Cooldown of Fire Blast is reduced by haste and the talent is then supposed to reduce it by flat 2 seconds (that means this aura is supposed to update value with stats (haste) changes)

mdX7: shouldn't we already take that into account by using the current rechargetime for the calculation?

Shauren: no, this is fine

PR #27631, opened 2022-01-18
[SmoothPhasing]: Implement SmoothPhasing Replace Object

Seyden:
**Changes proposed:**

-  Replace a object and restore it automatically on despawn with the original guid
-  Target selection replacement also works if the replaced object is visible and gets invisible
-  Implicit target selection replacement doesnt work like adressed in "Known issues"

**Issues addressed:**

**Tests performed:**
1. .npc add 166824
2. .learn 332402
3. attach spell_replace_object_test to spell 332402
4. target the npc you spawned and cast the spell to any position you want
5. personal npc gets spawned and the original will vanish
6. cast the spell again at any position (with your personal copy in target)
7. personal copy will despawn and will be replaced with original npc

- [x] Builds
- [x] Tested in-game

- [ ] Implicitly Target Selection replacing 
- Example:
- Named A - We have a questgiver with the entry 5000 (Visible)
- Named B - We have a hidden npc with the entry 6000 (Not visible for the player, becomes visible at some point)
- Named C - We have the temp summon of entry 6000 which gets summoned at the quest giver position
- If Player rewards a quest at NPC A, NPC B will spawn NPC C at NPCs A position, NPC C will have the replace object info for replacing NPC B (normally this will switch the target selection from NPC B to NPC C if you selected NPC B) but because NPC B is not visible, the client knows a way to check if the NPC C looks the same like NPC A or by any other marking and if the destroying of NPC A happens in the same SMSG_UPDATE_OBJECT packet like object creation of NPC C, then it will implicity replace NPC A with NPC C and therefore change the Target Selection from NPC A to NPC C
(We currently dont merge the packets, therefore the implicit target selection is not working)

### DIFF
--- a/src/server/game/Entities/Object/Object.h
+++ b/src/server/game/Entities/Object/Object.h
@@ -370,7 +370,7 @@ class TC_GAME_API Object
             }
         }
 
-        void BuildMovementUpdate(ByteBuffer* data, CreateObjectBits flags) const;
+        void BuildMovementUpdate(ByteBuffer* data, CreateObjectBits flags, Player* target) const;
         virtual UF::UpdateFieldFlag GetUpdateFieldFlagsFor(Player const* target) const;
         virtual void BuildValuesCreate(ByteBuffer* data, Player const* target) const = 0;
         virtual void BuildValuesUpdate(ByteBuffer* data, Player const* target) const = 0;
@@ -436,6 +436,12 @@ class FlaggedValuesArray32
     private:
         T_VALUES m_values[ARRAY_SIZE];
         T_FLAGS m_flags;
+};
+
+struct TC_GAME_API ReplaceObjectInfo
+{
+    ObjectGuid ReplaceObject;
+    bool StopAnimKits = true;
 };
 
 class TC_GAME_API WorldObject : public Object, public WorldLocation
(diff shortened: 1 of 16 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -24072,6 +24072,44 @@ uint8 Player::GetStartLevel(uint8 race, uint8 playerClass, Optional<int32> chara
...
+    if (isRestore)
+    {
+        RemoveReplacedObject(replacedObject);
+        m_Events.AddEvent(new ReplacedObjectRestoreCleanupDelayEvent(this, newObject), m_Events.CalculateTime(1000));
+    }
+}
+
+ReplaceObjectInfo const* Player::GetReplacedObjectForCreature(Creature const* creature) const
+{
+    auto itr = _replacedObjects.find(creature->GetGUID());
```
Seyden, line 24091: @Shauren maybe add another vector for cross referencing?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -24072,6 +24072,44 @@ uint8 Player::GetStartLevel(uint8 race, uint8 playerClass, Optional<int32> chara
...
+    if (itr != _replacedObjects.end())
+        return &itr->second;
+
+    return nullptr;
+}
+
+bool Player::HasReplaceObjectInfoFor(Creature const* creature) const
+{
+    for (auto const&[_, replaceInfo] : _replacedObjects)
```
Seyden, line 24100: @Shauren same here?

src/server/game/AI/SmartScripts/SmartScriptMgr.h
```diff
@@ -622,8 +622,9 @@ enum class SmartActionSummonCreatureFlags
     None = 0,
     PersonalSpawn = 1,
     PreferUnit = 2,
+    ReplaceObject = 3, // Requires: PersonalSpawn
```
BAndysc, line 625: it is a flag, so it can't be 3 ;)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -24166,6 +24166,30 @@ inline void BeforeVisibilityDestroy<Creature>(Creature* t, Player* p)
         t->ToPet()->Remove(PET_SAVE_NOT_IN_SLOT, true);
 }
 
+void Player::UpdateVisibilityOf(std::list<WorldObject*> targets)
```
Naios, line 24169: Do not use an owning datastructure here https://en.cppreference.com/w/cpp/container/span (if this is available to TC)

src/server/scripts/Spells/spell_generic.cpp
```diff
@@ -4732,6 +4732,45 @@ class spell_summon_battle_pet : public SpellScript
     }
 };
 
+// 332402 - Anima Orb (Just for testing until PR is finished)
+class spell_replace_object_test : public SpellScript
```
Naios, line 4736: Dont forget this

src/server/game/Entities/Object/Object.h
```diff
@@ -762,6 +783,9 @@ class TC_GAME_API WorldObject : public Object, public WorldLocation
 
         ObjectGuid _privateObjectOwner;
 
+        std::unordered_map<ObjectGuid /* Seer */, ReplaceObjectInfo> _replacedObjects;
```
Naios, line 786: It is super heavy to add two `std::unordered_maps` to every object in the world.

Naios (changes requested):

src/server/game/AI/SmartScripts/SmartScriptMgr.h
Seyden: Damn i thought it was being masked

src/server/game/Entities/Object/Object.h
Seyden: what is the alternative? the original npc can replace multiple "to be despawned personal copies" from x players and the original npc can be replaced by x personal copies for each player

src/server/game/Entities/Player/Player.cpp
BertiRean: Tc is using cpp17, not 20

src/server/game/Entities/Object/Object.h
BertiRean: @Seyden I think that a vector could be more efficient in this case, how many ReplaceObject can have a npc?, 20 30?

BertiRean: In case that you need a map, may @Naios could tell you if the boost::flat_map is a good choice for this.

Seyden: @BertiRean the count is: max players on the server since start until creature respawn

PR #27655, opened 2022-01-21
Core/Proc: Named new proc flags

aquadeus:
**Changes proposed:**

-  Name new proc flags


Shauren:
Nope, these will be the names we will be using - https://github.com/TrinityCore/SpellWork/commit/2cd4eaeabfee36c65d3e6585cb679992593cf612

xvwyh:
For reference, this is not about "consistency with spellwork", these are the official blizzard names: https://wow.tools/dbc/?dbc=enumeratedstring&build=2.5.1.39640#page=1&colFilter[4]=407

PR #27671, opened 2022-01-25
Scripts/Northrend Use std::chrono boss_ionar

holydeew:
**Changes proposed:**

-  
-  
-  

**Issues addressed:**

**Tests performed:**

- [ ] 
- [ ]


Nyr97:
You're supposed to get rid of the Timers variables and basically change them into the new Chrono system.

src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
```diff
@@ -87,10 +87,10 @@ class boss_ionar : public CreatureScript
             bIsSplitPhase = true;
             bHasDispersed = false;
 
-            uiSplitTimer = 25 * IN_MILLISECONDS;
+            uiSplitTimer = 25s;
```
Nyr97, line 90: Instead of using this, you would basically update events or tasks in UpdateAI. And then, schedule these events.

src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
```diff
@@ -87,10 +87,10 @@ class boss_ionar : public CreatureScript
...
 
-            uiStaticOverloadTimer = urand(5 * IN_MILLISECONDS, 6 * IN_MILLISECONDS);
-            uiBallLightningTimer = urand(10 * IN_MILLISECONDS, 11 * IN_MILLISECONDS);
+            uiStaticOverloadTimer = (5s, 6s);
```
Nyr97, line 92: Instead of using this, you would basically update events or tasks in UpdateAI. And then, schedule these events.

src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
```diff
@@ -87,10 +87,10 @@ class boss_ionar : public CreatureScript
...
+            uiBallLightningTimer = (10s , 11s);
```
Nyr97, line 93: Instead of using this, you would basically update events or tasks in UpdateAI. And then, schedule these events.

src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
```diff
@@ -102,10 +102,10 @@ class boss_ionar : public CreatureScript
         bool bIsSplitPhase;
         bool bHasDispersed;
 
-        uint32 uiSplitTimer;
+        Seconds uiSplitTimer;
```
Nyr97, line 105: These should be removed.

src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
```diff
@@ -226,9 +226,9 @@ class boss_ionar : public CreatureScript
             // Splitted
             if (!me->IsVisible())
             {
-                if (uiSplitTimer <= uiDiff)
+                if (uiSplitTimer <= Seconds(uiDiff))
```
Nyr97, line 229: And this type of checking should also be replaced with task/event scheduling.

Nyr97 (changes requested):

iCeROCKZ:
Thank you

Ovahlord:
How about just converting the script to use eventMap instead? that one only allow chrono literals and is far less painful to look at

src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
Shauren: This doesn't work, you are always setting it to 6 seconds

src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
Shauren: This doesn't work, you are always setting it to 11 seconds

src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
Shauren: Update diff is in milliseconds, not seconds

src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
```diff
@@ -245,36 +245,36 @@ class boss_ionar : public CreatureScript
...
             }
 
-            if (uiStaticOverloadTimer <= uiDiff)
+            if (uiStaticOverloadTimer <= Seconds(uiDiff))
             {
                 if (Unit* target = SelectTarget(SelectTargetMethod::Random, 0))
                     DoCast(target, SPELL_STATIC_OVERLOAD);
 
-                uiStaticOverloadTimer = urand(5*IN_MILLISECONDS, 6*IN_MILLISECONDS);
+                uiStaticOverloadTimer = 5s, 6s;
```
Shauren, line 266: This doesn't work, you are always setting it to 6 seconds

Shauren (changes requested):

Shauren:
Seeing as this PR doesn't really bring any benefit (only issues) I will just close it

PR #27674, opened 2022-01-25
Core/DBUpdater: Apply updates in a transaction

Shauren:
**Changes proposed:**

-  Wrap each sql update in a transaction

Useful for development, cherry-picking and in the rare case a broken sql gets pushed

**Tests performed:**

Tested on windows and with spaces in sql file name

**TODO list:**

- [x] Test on linux
- [x] Test on linux with spaces in name/path
- [x] Test on windows with spaces in path to cloned dir


jackpoz:
Just as a note, some MySQL statements executed in a transaction are applied even before COMMIT, like all the ones that alter a table structure.

Shauren:
b79a91039bba12d62e43806d8a20ad76107ac5d7 and ce4981b23f55eee427a850986ebbdd373d0683fe

PR #27678, opened 2022-01-27
DB/Loot: Rework Level 1 - 5 loot to be blizzlike

ghost:
**Changes proposed:**
-  Add extra primary keys to creature_loot_template to allow proper loot.
- Redo loot for all starting areas

Added for testing on TC.

### DIFF
--- a/sql/updates/world/3.3.5/2022_01_27_02_world.sql
+++ b/sql/updates/world/3.3.5/2022_01_27_02_world.sql
@@ -1,0 +1,891 @@
+-- Update Primary Keys for creature_loot_template
+ALTER TABLE `creature_loot_template` DROP PRIMARY KEY, ADD PRIMARY KEY (`Entry`, `Item`, `Reference`, `GroupId`) USING BTREE;
+
+SET @REF := 20000;
+DELETE FROM `reference_loot_template` WHERE `Entry` BETWEEN @REF AND @REF + 18;
+DELETE FROM `reference_loot_template` WHERE `Entry`=24072;
+
+SET @REFGREY0 := @REF;
+-- Correct ref loot for various grey level 1 to 5 Vanilla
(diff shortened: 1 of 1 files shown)


MaxtorCoder:
Make an issue for sql fixes.

offl:
Why you think same reference is assigned twice to same creature?

malcrom:
Because I'm smarter than you.

malcrom:
This is here for anyone who is interested to test. I know Shauren and Awful will have issues with it because they are assholes.

Teppic1:
It's quite common to see in classic mobs dropping items from the same reference template a second time, and things are certainly separated out by mob type (like boars vs humans). The templates are definitely different between TBC and vanilla 1-20 zones too. Something that'd be more difficult to emulate is how many mobs will drop a single item or a bunch of items from the same templates, but usually nothing in between, though it's not a big issue.

Shauren:
> It's quite common to see in classic mobs dropping items from the same reference template a second time, and things are certainly separated out by mob type (like boars vs humans). The templates are definitely different between TBC and vanilla 1-20 zones too. Something that'd be more difficult to emulate is how many mobs will drop a single item or a bunch of items from the same templates, but usually nothing in between, though it's not a big issue.

That is where you simply change *Count columns to roll another item from the same reference. There is no need to do a silly thing like modifying primary key of the table and inserting the reference twice.

malcrom:
> It's quite common to see in classic mobs dropping items from the same reference template a second time, and things are certainly separated out by mob type (like boars vs humans). The templates are definitely different between TBC and vanilla 1-20 zones too. Something that'd be more difficult to emulate is how many mobs will drop a single item or a bunch of items from the same templates, but usually nothing in between, though it's not a big issue.

This is the only way you can make it work blizzlike. Some here think they know everything without testing and are completely wrong and delete comments.

Issue #24257, opened 2020-03-06
DB/Creature: Strigid Screecher should get assists


Teppic1:
**Description:**

These owls (id 1996) will pull any others nearby during combat, on a repeat timer every few seconds. Tested in Classic retail, it's not just an aggro pull as it happens if you kite them in range of others too.

**Current behaviour:**

No script.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

c5226aec41ce1430b53cc5b4073dff644fe37dac

Note: haven't tested Strigid Hunters if they have the same behaviour. (I don't know how factions work so I haven't attempted a fix)

Aokromes:
a lot of start zone assists where removed on wow 3.

Teppic1:
I've no way of checking if it was removed as I can't even find any comments about them doing it in the first place. These are level 7-8 though, not 1-5, if that makes any difference.

illfated:
That is an interesting point. I can't remember seeing that behaviour myself, but then again I did not fight those birds much on retail back in 2006-2012. I wonder if a similar behaviour would apply to other beasts who use the emote "%s cries out for help."

Teppic1:
In this case there's no message about assistance. I noticed it because all of the owls are neutral normally, so it's obvious when another you haven't hit suddenly starts attacking you.

Killyana:
This is an old behavior that was removed after vanilla.

Teppic1:
Source would be nice...

Killyana:
I just tested in retail and it doesn't happen, blizzard made many changes in starting zones after vanilla and BC, but finding any sources would be hard.

illfated:
It is even harder to find because the behaviour is for mobs above level 1-5 (after leaving the sandbox starter area). The changes for mobs in the sandbox starter area (1-5) is documented well enough where all mobs became neutral instead of hostile.

Aokromes:
https://us.forums.blizzard.com/en/wow/t/starting-area-neutral-mobs/274573

Teppic1:
Yeah, in 3.3.0 they made the 1-5 experience easier by making the hostile mobs now neutral in all of the starter areas.

I don't know of any changes to mobs outside of the starter areas though. I don't think any hostile mobs were made neutral and I don't know of any abilities being removed. In Cata tons of mobs were changed, including abilities/loot, so current retail isn't a good source for earlier.

illfated:
[...]

PR #27700, opened 2022-01-31
Core/Scripts/SAI: Fix Silithus Wind Stones

### LINKED ISSUE #27698: Wind Stones in Silithus are not reseting after use
### Description

After using a wind stone in Silithus to call forth a Templar, Duke, etc there's supposed to be a cool down until it's usable again. 

They currently appear to be usable once and then do not reset. You have to move to another stone or camp. 

Currently, the only way to use the wind stone again is to restart the server.

### Expected behaviour

There is supposed to be automatic reset timer on them. WOWIKI claims it is "approximately 15 mins".

https://wowwiki-archive.fandom.com/wiki/Wind_stone

1. Go to Silithus and kill Twilight's until you have the 3 required garments
2. Don the gear, click the wind stone, call forth Templar
3. Defeat Templar (on lesser wind stones, dukes on wind stones, etc)
4. Wind stone will not return to a usable state.

### Branch

3.3.5

### TC rev. hash/commit

TrinityCore rev. a650bd6 2022-01-16 18:50:28 +0100 (3.3.5 branch) (Win64, RelWithDebInfo, Static)

### Operating system

Windows 7 x64

### Custom changes

Trinity Bots

anzz1:
**Changes proposed:**

-  Destroy Wind Stone on use, allow it to respawn (remove GO_FLAG_NODESPAWN on initialize) 
-  Re-check twilight set aura on GossipSelect (don't allow to summon without wearing set)
-  Fix spell Abyssal Punisment (24803) not casting when interacting with stone without required items
-  Fix miniboss not despawning after leaving combat
-  Fix possible exploit of spawning many minibosses from a single wind stone spawn
-  Aggro summoner as primary target, his group as secondary
-  Failsafe to despawn boss in 30s if unable to acquire a target

SAI Scripts for minibosses
-  On Summon: Set Root , Set Immune to PC, Create Timed Event 1 (5s), Create Timed Event 2 (30s)
-  On Timed Event 1 (5s): Unset Root, Unset Immune to PC, Add 100.0f Threat (Stored List 1), Add 50.0f Threat (Stored List 2)
-  On Reset: Set Counter 1 +1
-  On Counter 1 = 2: Despawn Self In 1000ms
-  On Timed Event 2 (30s): Despawn Self Instant
-  On Aggro: Remove Timed Event 2

(*) Stored List 1 = summoner , Stored List 2 = summoner's group

**Issues addressed:**

Closes #27698

**Tests performed:**

Builds, tested in-game, everything seems to work as it should.

- [ ] Initial threat values of 100.0f for the summoner and 50.0f for other group members are made up and not checked against sniffs or live servers.


Jinnaix:
You are overshooting here. 
Item removal is already fully and correctly covered.
Changing the punishment spellcast source from gob to player seems correct and missing.
I can not comment on the GO data change, but it would surprise me if no one jumps in and calls this a hack. 

I copy paste my notes from the last PR i made for windstones:

> Stuff to note down, mabye for a new issue, not covered by this PR
> 
> * [ ]  Wind Stone Bosses should have a grace period for about 5 Seconds (potentially for the summoning player to re-equip items after he loses the worn twillight set)
> * [ ]  Wind Stone Bosses should get aggro on the summoning player AND any other player on that players group
> * [ ]  Used Wind Stone should be destroyed (currently the spell disables the gameobject) and respawn after some time
> * [ ]  Spawned Boss should despawn immediately after combat ends (in case of attacking player/s die for example)

TQPS:
I can confirm that items are removed correctly after summon - this includes medallions and rings. This works great.

Currently ...
There is no grace period for the spawn to engage player. I wont lie, I re-equip gear before clicking the bubble - garb still disappears.

The spawns engage players, pets and bots - no noticeable issues seen.

The gameobjects remain in place after use. A usable one displays the type of wind stone when you hover over it - after you use it, it sits there dormant. Hovering over it displays nothing. Does not despawn or become available again. Requires server restart to use again.

If you die the Templar/Duek etc are supposed to despawn - currently they do not. Died after summoning one and after the corpse run, Duke was still there.

Aokromes:
yeah changing data without sniff is hack xd

Ovahlord:
> yeah changing data without sniff is hack xd

4.x sniffed template shows data5 = 0 as well. So it's true that touching the data is a hack but that one here was a lucky strike.

anzz1:
Oh okay, well if that is the case it means that GameObjectTemplate::GetDespawnPossibility() is wrong in core because it gets whether a gameobject should be able to respawn with `return questgiver.noDamageImmune != 0` for the type `GAMEOBJECT_TYPE_QUESTGIVER` (2). And looking from past videos and wiki/wowhead sources these stones should definitely destroy & respawn. 

Hmm, it did not remove the items for me without this code? Is it handled in SQL scripts or where exactly? I tried to find events tied to the stone template entries but couldn't find any, except for the gossip texts.

Jinnaix:
> And looking from past videos and wiki/wowhead sources these stones should definitely destroy & respawn.

Yes, this is true, thats what should happen.

> Hmm, it did not remove the items for me without this code? Is it handled in SQL scripts or where exactly? I tried to find events tied to the stone template entries but couldn't find any, except for the gossip texts.

Item removal is handled via a spell, the gameobject (the windstones) will force the player to cast a spell on himself when he uses a gossip item. The spell does not care if you wear the items or have them in your bag as far as i know. It should work correctly.

> That being said, I'm glad for the feedback and as a newcomer I do not yet fully grasp which things can be changed and which not, and this information can't be easily found anywhere so all this is new information for me and good for learning.

In a ideal scenario, we just build the framework and let the games original data do as much as possible on its own.

So while your attempt to change the data of that gameobject gives you the correct behavior (and would not get noticed from a players point of view what you did there) its wrong for us. These data fields have been filled with the correct data by sniffing them directly from blizzard. So we have to find another way to do this without changing the datafields (this is what people call a hack). Obviously i dont know a working way here.
Maybe you already pointed out the right place to make a change in the core, im not confident with the function you mentioned at the start.




> Also yeah, it currently does not re-check whether you have the proper auras on gossip option select, should it do that? What happens on live servers if you un-equip the items and then click the option? It does nothing? That can easily be changed if that is the case.

Im not 100% sure if it checks on live servers at click or not, but that's a change i would support.

anzz1:
With `me->RemoveFlag(GAMEOBJECT_FLAGS, GO_FLAG_NODESPAWN);` can revert the sql change, because that is what the core code does if Data5 (noDamageImmune) = 0.

offl:
That object should not have that flag
Checked classic & Cata, no flag. Maybe all GAMEOBJECT_TYPE_QUESTGIVER should not have it

anzz1:
Okay this PR is kinda borked, I just realized that some of the logic is happening somewhere else, if I go near a stone and use .cast 24762 (SPELL_DUKE_RANDOM) near a (medium "rank 2") wind stone (for example GOB template id: 180554), I get hit with the punishment spell which is probably caused by a trap underneath every wind stone (GOB template id: 180502) and it also activates the stone leaving it in the same unusable "activated" but not despawn state as before.

offl:
The fact that noDamageImmune is used for 2 checks at once is suspicious
Maybe it even is not be linked to GO_FLAG_NODESPAWN

anzz1:
Yeah if I understood the logic right noDamageImmune = 0 would mean it *is* "immune to damage". And that is why it is linked to nodespawn, that if its deleted it's not meant to ever come back. But does such questgiver objects exist?

anzz1:
Okay so with the last commit its mostly working fine (evertything else except for the instant attack), adding 1.1 second delay to the despawn lets the "some other logic" somewhere else does its' thing before despawning so items are correctly removed (and the punishment spell cast, yes it casts also when you *do* have the correct items it seems). But that is not exactly the most clean way and I also do not like that I don't know where the other logic is coming from.

anzz1:
So I thought the timer to engage after 5 sec and despawning upon leaving combat would be easy to add but these mobs actions are handled in SmartAI so I don't think I can override that, right? I don't really understand smart_scripts enough yet to work with em so that's a no go for me rn.

anzz1:
Boss now despawns after 5 seconds out-of-combat. Feel free to check in live server whether 5 seconds is right amount or not

anzz1:
Okay I got everything to work as it's supposed to with SAI actions

anzz1:
Yeah I guess that's it for now. Everything seems to work as it should. On a side note though, I noticed that if you stealth up before the miniboss aggroes, it will just keep following you and look at you but not attack until you unstealth or do some other action. This is unrelated to this one though but rather a core issue.

anzz1:
All previous concerns have been addressed and everything 'hacky' in the implementation has been changed. I cleaned up the commit to its' (hopefully) final state and in my point-of-view this PR looks ready to merge. If you have any more concerns or issues to address, please feel free to comment.

anzz1:
Okay one more "last" modification , I noticed that using `TEMPSUMMON_TIMED_DESPAWN_OUT_OF_COMBAT` also means that it despawns very fast on death and loot disappears (well, a corpse is definitely "out of combat") and thats a no-no. I reverted back the change in script to use `TEMPSUMMON_TIMED_OR_DEAD_DESPAWN` as it were and implement the despawn in SAI instead, by using a counter on reset. First reset is on spawn, but if a second reset happens the boss will despawn.

Jinnaix:
First of all, thank you for your effort.

The changes read well. I have no time right now give it an ingame test but as far as i can see, everything is covered.
Aside from some sources that state that all players in the group of the summoning player should get aggro too, but i guess this is a minor detail thing missing.

It seems it didnt took you very long to get familiar enough with SAI, assuming @123123123123123123123 is your account too :D
The solution with the counter increase is smart.

Note:
Every PR gets squashed on merge, force pushes arent needed and generally also not liked, just as a note for potential future PR.

Aokromes:
rename 2022_02_01_00_world.sql to 2022_99_99_00_world.sql

anzz1:
Yeah I don't know what happened with the account name :D, probably something wrong in my git config with the new github token auth thing. 

Thanks to all of you too for taking the time to comment and direct me to the right way of doing things. 

I'll look into the group aggro stuff.

anzz1:
Yeah I ended up doing exactly that, thanks for pointing me out to the right direction. 

I decided to use SMART_ACTION_ADD_THREAT as it seemed to work wonderfully for this, but I didn't find any examples of other scripts using that action. 

Now it correctly engages preferring the summoner as primary target, and his possible group as secondary targets. Note, I pulled these threat values (100.0f for primary, 50.0f for secondary) completely out of my ass as a seemingly reasonable numbers, but they are not checked against live servers or sniffs. So that is left as "todo".

I also added a failsafe to despawn the boss after 30s if he is unable to engage anyone in that time, like what could happen after selecting a gossip option and then in the 5 second time window the summoner (and whole group) teleports away. If a single person from the group is left though, boss will engage that person.

src/server/scripts/Kalimdor/zone_silithus.cpp
```diff
@@ -1214,9 +1217,29 @@ class go_wind_stone : public GameObjectScript
...
+                    primaryTarget.push_back(player);
+                    ENSURE_AI(SmartAI, summons->AI())->GetScript()->AddToStoredTargetList(primaryTarget, 1);
+
+                    // Add player's group (if applicable) within 50yds to SmartAI stored target list 2
+                    if (Group* group = player->GetGroup())
+                    {
+                        ObjectVector secondaryTargets;
+                        for (GroupReference* groupRef = group->GetFirstMember(); groupRef != nullptr; groupRef = groupRef->next())
+                            if (Player* member = groupRef->GetSource())
+                                if (member != player && member->IsInMap(summons) and member->GetDistance(summons) < 50.0f)
```
Shauren, line 1236: We don't use named logical operators, please replace `and` with `&&`

src/server/scripts/Kalimdor/zone_silithus.cpp
```diff
@@ -1214,9 +1217,29 @@ class go_wind_stone : public GameObjectScript
...
+                                    secondaryTargets.push_back(member);
+
+                        if (secondaryTargets.size() > 0)
+                            ENSURE_AI(SmartAI, summons->AI())->GetScript()->AddToStoredTargetList(secondaryTargets, 2);
```
Shauren, line 1240: `ENSURE_AI` will cause a crash if for some reason summon AI was not SAI, to be safe you need to check it
```
if (SmartAI* sai = CAST_AI(SmartAI, summons->AI()))
    sai->GetScript()->AddToStoredTargetList(secondaryTargets, 2);
```

src/server/scripts/Kalimdor/zone_silithus.cpp
anzz1: That was an oopsie on my part, fixed. I guess there is a bug in Visual Studio since it shouldn't allow to do that when permissive-flag is explicitly disabled.

src/server/scripts/Kalimdor/zone_silithus.cpp
anzz1: Oh, I thought that ENSURE_AI was meant to do exactly that, "Ensure AI" :D . I guess it does not. Will fix, thanks.

anzz1: Yeah not having it as SmartAI triggers an assertion and "crashes" the server. Shouldn't that happen though? Isn't it what asserts are for, to stop execution is something is very wrong, like it is if the unit does not have SmartAI.

Shauren: No, scripts are never supposed to crash the server in case of bad data in database (such as missing scriptnames/ainame assignments)

jackpoz: `ENSURE_AI` is supposed to be used only when 100% sure it will never ever not be of the requested type. This is useful to mark cases as "it will never happen" for static analysis reasons and to get an assert at the correct point instead of a null dereference check sometime further down the code.

As a rule of thumb, it's good to check the implementation of methods when using them. See https://github.com/TrinityCore/TrinityCore/blob/a8475e5d70c4a49b3cd9865f5059ac12a333a98c/src/server/game/AI/CoreAI/UnitAI.h#L29-L37 on how `ENSURE_AI` is implemented

src/server/scripts/Kalimdor/zone_silithus.cpp
```diff
@@ -1148,7 +1149,9 @@ class go_wind_stone : public GameObjectScript
             go_wind_stoneAI(GameObject* go) : GameObjectAI(go) { }
 
             private:
-                uint8 GetPlayerRank(Player* player) // For random summoning
+                bool isSummoning;
```
jackpoz, line 1152: Please initialized class member fields in the constructor to avoid uninitialized values (even if `Reset()` is assigning a value

anzz1: Yeah, I thought the way of declaring them uninitialized was a bit weird too, but I took example from the rest of the script and thought that was "The Way To Do It™". I'll change rest of them too.

src/server/scripts/Kalimdor/zone_silithus.cpp
anzz1: Yeah triggering an assertion is indeed a nuclear option, I understand now that ENSURE_AI means "absolutely ensure in matters of life and death" and not in db or script errors or such. I added the proper checks and logged errors instead of using assertion.

src/server/scripts/Kalimdor/zone_silithus.cpp
anzz1: I initialized all values found in the script's classes except for the ObjectGuid ones. By a quick glance it mostly seems that the ObjectGuids are checked to exist before trying to access them, but I didn't have a meticulous look through the whole script.

Shauren: ObjectGuid has a default constructor that ensures it is always set to 0 so it does not need any additional initialization code

anzz1: Okay, thanks for the review :)

src/server/scripts/Kalimdor/zone_silithus.cpp
```diff
@@ -1214,41 +1218,88 @@ class go_wind_stone : public GameObjectScript
...
+                        // On Summon: Set Root, Set Immune to PC, Create Timed Event 1 (5s), Create Timed Event 2 (30s)
+                        // On Timed Event 1 (5s): Unset Root, Unset Immune to PC, Add 100.0f Threat (Stored List 1), Add 50.0f Threat (Stored List 2)
+                        // On Reset: Set Counter 1 +1
+                        // On Counter 1 = 2: Despawn Self In 1000ms
+                        // On Timed Event 2 (30s): Despawn Self Instant
+                        // On Aggro: Remove Timed Event 2
+                        // In Combat: Use NPC specific spells
+                    }
+                    else {
+                        TC_LOG_ERROR("scripts", "[zone_silithus] go_wind_stoneAI: NPC `%s` (%s) has wrong AI `%s`, should be `SmartAI`.", summons->GetName().c_str(), summons->GetGUID().ToString().c_str(), summons->GetAIName().c_str());
```
jackpoz, line 1286: Nice to see error logs being added, they help a lot.

If in the future you want to log more information but don't want to write a super long logging function trying to get all the data that might be useful, a lot of classes implement `std::string GetDebugInfo() const` that logs a lot:
- position with x,y,z and map id
- guid, name, entry
- AI
- summoner/charmer
- movement

No matter how spammy this might be with 10/15 lines of data being logged, it gets logged only on errors anyway.

Sometimes I append the GetDebugInfo() result to the end of the log message, so I get the best of both (1st line having the custom log stuff I want to focus on, then the rest).

You don't have to change anything in your code, it's just me sharing information 😃

src/server/scripts/Kalimdor/zone_silithus.cpp
```diff
@@ -1328,61 +1379,124 @@ class go_wind_stone : public GameObjectScript
...
                             break;
                     }
                     return true;
                 }
+                void Reset() override
+                {
+                    // Allow despawning
+                    me->RemoveFlag(GAMEOBJECT_FLAGS, GO_FLAG_NODESPAWN);
+                    isSummoning = false;
+                }
```
jackpoz, line 1495:
```suggestion
                }

```

jackpoz: Adding an empty line between the 2 functions

src/server/scripts/Kalimdor/zone_silithus.cpp
```diff
@@ -1328,61 +1379,124 @@ class go_wind_stone : public GameObjectScript
...
+                                SummonNPC(me, player, NPC_ROYAL_AIR, SPELL_ROYAL_AIR);
+                            }
                             break;
 
                         default:
                             break;
                     }
                     return true;
                 }
+                void Reset() override
```
jackpoz, line 1490:
```suggestion

                void Reset() override
```
Adding an empty line between the 2 functions

src/server/scripts/Kalimdor/zone_silithus.cpp
```diff
@@ -1214,41 +1218,88 @@ class go_wind_stone : public GameObjectScript
...
+                        // Next actions happen in smart_scripts
+                        // On Summon: Set Root, Set Immune to PC, Create Timed Event 1 (5s), Create Timed Event 2 (30s)
+                        // On Timed Event 1 (5s): Unset Root, Unset Immune to PC, Add 100.0f Threat (Stored List 1), Add 50.0f Threat (Stored List 2)
+                        // On Reset: Set Counter 1 +1
+                        // On Counter 1 = 2: Despawn Self In 1000ms
+                        // On Timed Event 2 (30s): Despawn Self Instant
+                        // On Aggro: Remove Timed Event 2
+                        // In Combat: Use NPC specific spells
+                    }
+                    else {
```
jackpoz, line 1285:
```suggestion
                    else
                    {
```
Applying TC codestyle

jackpoz:
I added some nitpicking code suggestions for codestyle, so you can just click on github to accept them :) (hopefully I didn't add any whitespace but CI checks will tell 😄 )

anzz1:
Thanks, I did just that. I'm all for nitpicking, a well-formatted and uniformly styled code is something that makes it pleasing to work with instead of a garbled mess which just gives you a headache and even make you not want to work it (and it's especially important in open-source volunteer projects). I tried to follow the formatting but obviously missed a couple of spots, thanks for picking them up :). 

Note to self: I should look into updating my knowledge whether auto-formatters work well today, previously I had lots of issues with them touching files I didn't want to modify and also not really understanding the format properly and doing stuff wrong. Should check whether that is still the case or have they gotten better in the past few years.

src/server/scripts/Kalimdor/zone_silithus.cpp
```diff
@@ -818,15 +820,15 @@ class npc_anachronos_quest_trigger : public CreatureScript
...
 
-        int8 LiveCount;
-        uint8 WaveCount;
+        int8 LiveCount = 0;
+        uint8 WaveCount = 0;
 
-        bool EventStarted;
-        bool Announced;
-        bool Failed;
+        bool EventStarted = false;
```
offl, line 829: Better to make adjustments only to same thing in a single PR

offl:
To explain changes, here's how it works:

1. Player selects gossip
2. Object casts Apply Aura: Periodically Trigger Spell on player
3. After 1 sec player casts spell from periodic trigger above
4. Activate Object: (Make Inert) of that spell makes object unselectable. Since it's delayed for 1 sec, object remains selectable before player cast, that can be seen on retail(visually). Probably player can still click on gossip option and probably that even will execute script. I believe it's not possible to spawn 2 or more creatures at once due to how spawning and the script itself works or Blizz just does not care, they usually simply don't care about supporting all edge cases and even simpliest ones
5. Here creatures are spawned from DB(by default they're just inactive, like creatures in world events), we don't really have support for nearly all systems involved here and few things here are still unknown. Spawner here is player or player is involved in spawning, whatever it is creature should have player guid because it executes events not from gameobject script but from its own script and actions requires player guid(Blizz usually tries to make creature have player guid when spawning creature, by making player summoner). Gameobject script and creature script are separated and has to be scripted only in DB. Creatures are not spawned inside each object but at specific position(same pos each time(sniffs)). Blizz can't summon creature without spells, they only spawns them from DB. That requires 4 spawned in DB creatures for each object(52 in total), it may sound ridiculous but it's how it works. Also that's the reason why each object has unique entry. Probably same with all other duplicated objects, like those multiple cages but reason for duplicating may be different for each case
6. Spawned creature executes action list on spawn, talks, casts spell, changes orientation to player. I added 3 action lists due to the way Blizz handles texts so it will be easier to update in future. After 8 sec creature becomes attackable, attacks spawner. Why aura player casts has duration of 8 sec and what it does when applied on player remains unknown

Also creatures are immune to PC by default(not scripted), Wiki states those elementals despawns after evade and you have only one attempt to kill them, so we just make them immune to PC by default and despawn after evade. Despawn if not engaged maybe is supported or maybe is not supported, since it's not a critical issue here, I prefer to not make code more complicated using methods which we have currently and simply use different summon type

I appreciate your efforts in making the aggro thingy but it should be handled in core / only in creature script, but first it should be confirmed if creatures really enter in combat with spawner's group(and raid), maybe it's just combat pulse which actually should not be BossAI-exclusive mechanic

anzz1:
@offl I agree with your changes and that they went more in the right direction. Using "On Evade" was definitely better than using the counter, I somehow missed that option entirely. 

If by the "combat pulse" you mean the mechanic which makes everyone in the group of the "aggroer" in certain (close) range enter combat (added to the creature's threat list?), yeah I'm like 90% sure that happens with any mob anywhere, even outside instances, and not just bosses. So probably the right way to go would be exactly to only aggro the summoner like you did and look at the group mechanic as part of the core's combat mechanics and not in a script of a single creature. 

I have always mained rogue and I'm also pretty certain that once you get added to the creature's threat list (entering combat) there is no way to get off it without using spells like Vanish, Hunter's Feign Death, Mage's Invisibility etc. and the mob should find you and start attacking you even if you stay in stealth and never engage yourself. So any mob will always "see" you once you're on their threat list. Now it doesn't work that way, and this creature script is a good way to see that. 

As you can click the gossip option, enter stealth, and the mob will never find you and start attacking you even if you get added to it's threat list (like I did on my last commit). If you explicitly tell the mob to attack you while you are in stealth (the way I did it on my older commit and the way you do now on your last), it will just follow you, not attacking.

Notable is that somewhat counterintuitively (but that is the way it works on live), entering combat while in stealth does not break stealth, even if it does in the attacking creature's POV. Instead you will stay in stealth, and the stealth will only get broken after the mob hits you. Like it currently does.

I guess this should be posted as a separete issue and having a closer look to the core's combat mechanics, identifying these two issues
1) Most mobs (if not all) should use "combat pulse", adding close-by group members of the victim to the creature's threat/attack list, regardless of your stealth status
2) Once you are on this list, combat will be entered (that currently works), and this can only be removed by killing the mob, running outside of it's evade range, dying yourself or casting a spell that drops combat like Vanish or Feign Death. Staying in stealth without casting such a spell should make the creature be able to see you, find you, and attack you (this does not work, you will stay in combat indefinitely but stealthed, mob will return to its home/evade position and not try to find you, you will still stay in combat indefinitely). If you tell the mob explicitly to attack by a script, it will just follow you and not attack until your stealth is removed by other means (you removing it or attacking the mob).

anzz1:
On a sidenote: @offl I noticed you using the Run Script aka `SMART_ACTION_CALL_TIMED_ACTIONLIST` , is that preferred option instead of chain-linking smart actions together? And what are their functional differences?

offl:
>If by the "combat pulse" you mean the mechanic which makes everyone in the group of the "aggroer" in certain (close) range enter combat (added to the creature's threat list?), yeah I'm like 90% sure that happens with any mob anywhere, even outside instances, and not just bosses, regardless whether you are in stealth or not. So probably the right way to go would be exactly to only aggro the summoner like you did and look at the group mechanic as part of the core's combat mechanics and not in a script of a single creature.

I dont know if only group of the "aggroer" will be put into combat, that sounds like a really complicated mechanic, but yeah, that's what I meant

>Notable is that somewhat counterintuitively (but that is the way it works on live), entering combat while in stealth does not break stealth, even if it does in the attacking creature's POV. Instead you will stay in stealth, and the stealth will only get broken after the mob hits you. Like it currently does.

It's probably fully controlled by aura interrupt flags (like the one you added few days ago), in master they all are named from leaks, in 3.3.5 names are still old and some are not used (should be ported from master)

There's #24257, kinda related to combat pulse
That 'attack vs stealth' is weird, didn't know it works like that

>I noticed you using the Run Script aka SMART_ACTION_CALL_TIMED_ACTIONLIST , is that preferred option instead of chain-linking smart actions together? And what are their functional differences?

Linked events doesn't support timers, actions lists do
Action lists makes scripts less messy and readable + it looks like Blizz uses only action lists (we can't because our are implemented differently and in general system is different, for example creature can't run 2 or more action lists at once but it's unknown if Blizz really can run 2 or more)
Mostly linked events are used when timers are not needed and action lists when timers are needed

anzz1:
I don't think it's complicated, and I'm pretty sure it should work like this:

On Aggro: Add victim and his group members within < X yards to threat table.
Every X sec/frames: Add group members within < X yards to threat of the person who has "tag" on the mob.

That's it. It means that if my group member aggroes a mob and I'm close, I will enter combat, regardless of my stealth. It also means that if the fight has started before without me in it, but then I come and walk too close of a group member fighting a mob (which he has a tag on), I will enter combat too. If I walk past a mob fighting some random person, I obviously won't get combat. Also if my group member attacks a mob which some other group has "tag" on (grey nameplate, no loot), then I will also not get added to combat, only he will.


And yeah, to my best knowledge and iirc, stealth definitely works so that after you are placed in the threat table (enter combat), stealth won't help you anymore at all. When you aggro a mob in stealth (by walking too close, group member aggro, etc.), you will get combat, mob start to chase&attack but the stealth is not removed yet, only when the first hit lands.

So stealth should only affect whether a mob sees you or not out-of-combat, and nothing else. In combat the stealth does not have any effect.

And thanks for clearing that up about action lists & linked events.

PR #27733, opened 2022-02-05
Core/SAI: Add Offset to SET_HOME_POS

anzz1:
**Changes proposed:**

-  Add possibility to use an position offset on SMART_ACTION_SET_HOME_POS when using player/creature/gameobject as target
-  Fix comment on SMART_ACTION_CALL_TIMED_ACTIONLIST to reflect reality

**Issues addressed:**

Closes: None

**Tests performed:**

Builds, works as it should, tested in-game. Tiny change, won't break anything.
Tested with SQL query
```sql
SELECT * FROM `smart_scripts`
WHERE `action_type` = 101 AND target_type <> 8 AND (`target_x` <> 0 OR `target_y` <> 0 OR `target_z` <> 0 OR `target_o` <> 0);
```
to return nothing, meaning that no scripts had target position params set by accident before this was possible.  

(101 = is SMART_ACTION_SET_HOME_POS)
(8 = is not SMART_TARGET_POSITION)

- None

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -1874,7 +1874,7 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                 if (IsCreature(target))
                 {
                     if (e.GetTargetType() == SMART_TARGET_SELF)
-                        target->ToCreature()->SetHomePosition(me->GetPositionX()+e.target.x, me->GetPositionY()+e.target.y, me->GetPositionZ()+e.target.z, me->GetOrientation()+e.target.o);
+                        target->ToCreature()->SetHomePosition(me->GetPositionX()+e.target.x, me->GetPositionY()+e.target.y, me->GetPositionZ()+e.target.z, me->GetOrientation());
                     else if (e.GetTargetType() == SMART_TARGET_POSITION)
                         target->ToCreature()->SetHomePosition(e.target.x, e.target.y, e.target.z, e.target.o);
                     else if (e.GetTargetType() == SMART_TARGET_CREATURE_RANGE || e.GetTargetType() == SMART_TARGET_CREATURE_GUID ||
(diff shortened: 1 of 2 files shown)


offl:
Where it's needed?
Not that I really wanna change all of that but turned out Blizz uses spells (Anchor Here) to change home position, later, (not even in Cata) they implemented an action to do it in scripts without spells, it's called `Set Anchor Point`, looks like it can't use position of other unit or predefined position, only current position(same with spells)
Since we call SetHomePosition mostly after reaching waypoint, looks like something is missing in core and that should be called not manually in each script

anzz1:
Well, I actually needed it myself in #27735 :D. I mean there can be other use-cases for this, but I used it to set a home position of Z+1 so the unit is not on ground, but flying. IIRC even blizz had problems with their fly-units running in air instead of flying so the current movementgenerator seems 'blizzlike' in that way :D. 

Additional note: Regarding all SAI functions which deal with cartesian positions should probably have a flag if the wanted position is in absolute position units or a vector offset (example of which can be found in `SMART_ACTION_MOVE_OFFSET`), where +Y does not mean absolute +Y but rather +Y relative to the unit, meaning "go forward".

anzz1:
If you want to, you can test #27735 without this patch and you will immedietaly see the problem. :D
If the creature has a home point **on ground** and returns to it's previous waypoint movement **in air** after returning home, it will run to the waypoint instead of flying there. To my best knowledge this is 'blizzlike' (albeit unintended) behaviour and they also used to set a point above ground before moving to another air position (and hence the `MotionMaster::MoveTakeoff` takeoff movement type in core too). If there are any sniffs regarding fly units returning to air after point-movement in ground this could be confirmed for sure.

Shauren:
I don't have any opinion whether we should have this feature or not - but even if we do, this is not the right way to implement it. All "offset" type of positions should use `Position::RelocateOffset` that properly rotates offsets instead of doing plain addition

offl:
Let's first figure out if it's needed in #27735

anzz1:
> I don't have any opinion whether we should have this feature or not - but even if we do, this is not the right way to implement it. All "offset" type of positions should use `Position::RelocateOffset` that properly rotates offsets instead of doing plain addition

Yeah I know, hence my "Sidenote" about positions in smart actions. As there are two different offsets which the scripter could want. Relative to the world coordinates, or relative to the unit's coordinates with it's rotation. I used what seemed to be most commonly used (and thus expected) form. The problem is that it's kinda all-over-the-place right now, but it can't be fixed easily without breaking old scripts. One glaring example is when you compare `SMART_ACTION_SUMMON_CREATURE` and `SMART_ACTION_SUMMON_GO`, as Creature uses a "absolute" (relative to world) offset but Gameobject does it relative to the unit.

But yeah, I left the orientation part completely out for now. Absolute coords seem to be the most used form so this follows the current pattern. But a better way to handle this would be to add a param to every position action specifying which one is it that you want, absolute or relative coordinates.

PR #27737, opened 2022-02-06
Scripts/Spells: updated Penance

Nyr97:
**Changes proposed:**

-  removed unnecessary rank checks.
- converted script to current standard.

**Issues addressed:** None.

**Tests performed:** It builds and it was tested in-game.

**Known issues and TODO list:** None.


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -573,85 +573,64 @@ class spell_pri_mind_bomb : public AuraScript
...
+    }
 
-                uint8 rank = spellInfo->GetRank();
-                if (!sSpellMgr->GetSpellWithRank(SPELL_PRIEST_PENANCE_R1_DAMAGE, rank, true))
-                    return false;
-                if (!sSpellMgr->GetSpellWithRank(SPELL_PRIEST_PENANCE_R1_HEAL, rank, true))
-                    return false;
+    bool Load() override
+    {
+        return GetCaster()->GetTypeId() == TYPEID_PLAYER;
```
mdX7, line 591: why are we checking if its a player if we nowhere explicitly use player?

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -573,85 +573,64 @@ class spell_pri_mind_bomb : public AuraScript
...
+    }
 
-                return true;
-            }
+    SpellCastResult CheckCast()
+    {
+        Player* caster = GetCaster()->ToPlayer();
```
mdX7, line 596: `Player*` stuff is not used, you can simply use `GetCaster()`

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -573,85 +573,64 @@ class spell_pri_mind_bomb : public AuraScript
...
-        };
+    void HandleDummy(SpellEffIndex /*effIndex*/)
+    {
+        Unit* caster = GetCaster();
 
-        SpellScript* GetSpellScript() const override
+        if (Unit* target = GetHitUnit())
         {
-            return new spell_pri_penance_SpellScript;
+            if (!target->IsAlive())
```
mdX7, line 619: `SPELL_ATTR2_CAN_TARGET_DEAD` isn't set for 47540, so we shouldn't reach this point i think?

mdX7: nevermind, this cast is done while channeling (we are literally causing the waves here)

src/server/scripts/Spells/spell_priest.cpp
Nyr97: You actually do on SpellCastResult CheckCast(), maybe the target has any immune_to_pc and shouldn't be eligible for the spell since the caster is a player.

src/server/scripts/Spells/spell_priest.cpp
Nyr97: It is used to check !caster->IsFriendlyTo(target).

mdX7: Definition of `IsFriendlyTo` is in `WorldObject`, so we do not require `Player*`

Nyr97: ehh that makes sense, let me change it then

src/server/scripts/Spells/spell_priest.cpp
Nyr97: this is the dummy, the triggered spells are the channeled spells, the individual bolts are different Ids

src/server/scripts/Spells/spell_priest.cpp
mdX7: That sounds weird and doesnt make any sense. If I cast as a player and the target shouldnt be affected by a spell it required IMMUNE_TO_PC, if i cast it as a npc it requires IMMUNE_TO_NPC, theres nothing related to spellscripts at this point.
We only have to check if the caster is a player if we actually use stuff from the `Player` class itself and dont want to check it right before casting. And we do not use anything from Player in this script, just from inherited classes

PR #27738, opened 2022-02-06
3.3.5 add missing page_locale_text for zhTW and zhCN

leelf00:
**Changes proposed:**

-  add missing page_locale_text for zhTW and zhCN


Aokromes:
thx but we don't use pull request for sql only fixes, open a issue instead.

leelf00:
i18n work cause my account flag last month,any other way to prevent?

PR #27766, opened 2022-02-12
Scripts/Spells: Script one more player talk

offl:
**Changes proposed:**

-  Script one more player talk. The spell is used by vehicle and targets vehicle, in script vehicle forces passenger to talk

**Issues addressed:**

none

**Tests performed:**

builds, tested


sql/updates/world/3.3.5/2022_02_12_20_world.sql
```diff
@@ -0,0 +1,4 @@
+--
+DELETE FROM `spell_script_names` WHERE `spell_id` = 49550 AND `ScriptName` = 'spell_call_out_injured_soldier';
```
Shauren, line 2: doesnt make really a difference here but when adding a new script you just delete by scriptname alone

Shauren (approved):

sql/updates/world/3.3.5/2022_02_12_20_world.sql
offl: indeed

PR #27772, opened 2022-02-12
Script/Special: actually remove money from player when disabling or enabling experience gain from the related NPC

Alastro-Kraken:
**Changes proposed:**

- When enabling or disabling experience gain from the NPCs [Behsten](https://www.wowhead.com/npc=35365/behsten#comments) (Alliance) or [Slahtz](https://www.wowhead.com/npc=35364/slahtz) (Horde), you need to have 10 gold in order to disable/enable exp, but it won't be taken from you. That's because the gossip requring gold is only set on DB, but the script does not take money.

Note: since in order to reach this code path you need to have at least 10 gold, I did not add an additional check in the script.

**Issues addressed:**

Could not find any.

**Tests performed:**

Builds, tested and working.

### DIFF
--- a/src/server/scripts/World/npcs_special.cpp
+++ b/src/server/scripts/World/npcs_special.cpp
@@ -1750,6 +1750,8 @@ enum BehstenSlahtz
     OPTION_ID_XP_ON    = 1      // "I wish to start gaining experience again."
 };
 
+const int32 DisableEnableExperienceCost = 100000;
+
 class npc_experience : public CreatureScript
 {
 public:
@@ -1778,6 +1780,7 @@ public:
         {
             uint32 const action = player->PlayerTalkClass->GetGossipOptionAction(gossipListId);
             ClearGossipMenuFor(player);
+            player->ModifyMoney(-DisableEnableExperienceCost);
 
             switch (action)
             {


Shauren:
Isnt this supposed to be handled with "BoxMoney"?

Alastro-Kraken:
BoxMoney only checks if you have enough money, it does not take it away currently.

offl:
Maybe because of AddGossipItemFor params

Alastro-Kraken:
Oh yeah I see now that it'd take money if the gossip option was set from DB, but not if set from a script (the profession specialization scripts all have a manual ModifyMoney() call too).
I don't think it can be moved to DB with current condition options, and creating one just for this case might be overkill?

offl:
Who knows, maybe they used spells here(cast on gossip select), maybe whole thing is just scripted in DB
There's also ConditionScript, but I don't know if we should use it for such case and how row in `conditions` should look
I think ModifyMoney should be handled automatically in core but easiest way is just handle it manually, heh

Shauren:
All you had to do is to use this function that accepts more arguments instead https://github.com/TrinityCore/TrinityCore/blob/93735ae3bd926d13cff43ffce1f0640a2693649f/src/server/game/AI/ScriptedAI/ScriptedGossip.h#L90

My question instead was about retail-like factor, does it show a popup with confirmation that it will take gold or does it just silently remove money?

offl:
That function is not for database gossips
>does it show a popup with confirmation that it will take gold

Yes

PR #27783, opened 2022-02-16
Core/Auras: Fix SPELL_AURA_MOD_DAMAGE_DONE_FOR_MECHANIC for effects m…

### LINKED ISSUE #27534: Core/Auras: SPELL_AURA_MOD_DAMAGE_DONE_FOR_MECHANIC only takes spell mechanic into account and not effect mechanic
### Description

As the title states, SPELL_AURA_MOD_DAMAGE_DONE_FOR_MECHANIC only takes into account spell mechanics to increase or decrease in % a certain mechanic's power and ignores its effect mechanic.

As an example, I can provide Ranger Shot (347548) which doesn't increase in damage besides having any stacks of its own aura:

image

### Expected behaviour

SPELL_AURA_MOD_DAMAGE_DONE_FOR_MECHANIC should take both spell mechanic and effect mechanics into account.

1. Cast Ranger Shot (347548) twice on a target.
2. Notice the damage will not increase no matter the stacks.

### Branch

master

### TC rev. hash/commit

TrinityCore rev. 7f43d50e6fba+ 2022-01-03 18:36:52 +0100 (sylvanasSOD branch)

### Operating system

Windows 10 x64

### Custom changes

No custom changes besides branch sylvanasSOD.

aquadeus:
…echanics

**Changes proposed:**

-  Fix SPELL_AURA_MOD_DAMAGE_DONE_FOR_MECHANIC for effects mechanics

**Issues addressed:**

Closes #27534

**Tests performed:**

Tested in-game


Shauren (changes requested):
This is very wrong.

* You are applying SPELL_AURA_MOD_DAMAGE_DONE_FOR_MECHANIC as many times as the spell has effects in total
* Effect mechanic should have priority over spell mechanic

You need to add `SpellEffectInfo const& spellEffectInfo` argument to both functions you are modifying

Issue #22375, opened 2018-09-03
Core/Spells: Spell::SearchTargets doesn't always return full list of targets


Zhalfyr:
Description: 
it appears that  if a circle with radius equals to the range of the casted spell is entirely in your cell the visitors don't search for targets in adjacent cells.
Let me take an example : if you're standing 10 meters away from the nearest border of your cell, and your target is standing a bit further, behind that border. if the creature as an object size that make him overlap in your cell enough to be in range of an AoE you would cast, you won't hit it unless the area of the spell overlap the creature's cell, although you should hit the creature since it's objectsize makes it in range.

Steps to reproduce the problem:
Spawn a creature with huge radius (more than cellsize to be sure you're not in its cell)  go the furthest you can where your wow client says you're still in melee range (you can cast melee targeted attacks on the creature) and try to cast an AoE centered on you, you won't hit the creature.

Branch(es): master

TC rev. hash/commit: rev. 5a0ce66

OS : W10 x64

wonder2011:
Omg, that's bad !

matanshukry:
A bit old, but anyone got an actual example of such a creature with a "huge radius" ?

Zhalfyr:
Try Fallen Avatar in Tomb of Sargeras. If he is at a grid intersection, you can experience the issue.

matanshukry:
@Zhalfyr any way to make sure / see / test if I'm in a grid intersection?

Zhalfyr:
Its original location is at a grid intersection. 
Entry : 116939
X : 6606.67
Y : -795.122
Z : 1665.61
O : 3.13905
MapId : 1676

matanshukry:
@Zhalfyr Not sure how to get in?

`.go xyz 6606.67 -795.122 1665.61 1676` results in `invalid target map or coordinates..`
Tried to walk in too, but looks like areatrigger isn't implemented.

Got perhaps another example, something that can be reproduced easily?

Shauren:
insert a dummy instance_template row

matanshukry:
Still problematic to check though:
- The position is in the air. Probably should have a gameobject, doesn't exist there
- the npc it self is friendly so can't attack it either

matanshukry:
Closest thing I sort of did:
1. `.cheat god`
2. `.go xyz 6606.67 -795.122 1665.61 1676`
3. `.npc add temp 36853` ( Sindragosa )
4. Quickly attack it, hopefully will be as close position as possible
[...]

PR #27827, opened 2022-02-25
Core/Spells: Increase search radius for avoid problems with some spells that can't find units with huge combat reach

### LINKED ISSUE #27135: Halion - Combat Reach - different from sides
### Description

Halion has different hitbox from sides... this affects AOE spells like divine storm, bladestorm, fan of knives only.
from the side of Baltharus the Warborn - up to 26yd
from the side of Saviana Ragefire - up to 8yd

https://youtu.be/pVYt4lEFb5o  

### Expected behaviour

Halion should have the same hitbox around him, no matter which side you are standing

step 1.  get to the ruby sanctum and kill minibosses
step 2. .learn 53385
step 3. .cheat cooldown
step 4. .mod mana 9999999
step 5. observe different ranges from the sides

### Branch

3.3.5

### TC rev. hash/commit

ae9e583

### Operating system

Windows 10 x64

### Custom changes

none

Jildor:
**Changes proposed:**

- Increase search radius for avoid problems with some spells that can't find units with huge combat reach
- Added extra radius to area/cone spells, totems like Searing Totem and SPELL_EFFECT_APPLY_AREA_AURA_ENEMY like Totem of Wrath

Explained in this issue for master branch: https://github.com/TrinityCore/TrinityCore/issues/22375
Creatures like Halion, Sindragosa, Brain of yogg-Saron have big combat reach and if you're are in other cell, you can't find it if search radius is lesser than combat reach

I set a search radius of 40 yards, it seemed enough to me seeing the combat reach of creatures in 335.

I don't know if you think it's okay to do this like this, I can't think of another way to solve it that doesn't cause problems or is within my knowledge

**Issues addressed:**

Closes #27135

**Tests performed:**

tested in-game
before PR:
https://imgur.com/mpLweDJ
https://imgur.com/hYtD6Km

after PR:
https://imgur.com/iWcQ0i9
https://imgur.com/251EGyd

### DIFF
--- a/src/server/game/Grids/Cells/CellImpl.h
+++ b/src/server/game/Grids/Cells/CellImpl.h
@@ -60,15 +60,15 @@ inline CellArea Cell::CalculateCellArea(float x, float y, float radius)
 }
 
 template<class T, class CONTAINER>
-inline void Cell::Visit(CellCoord const& standing_cell, TypeContainerVisitor<T, CONTAINER>& visitor, Map& map, WorldObject const& obj, float radius, bool extraRadius /*= false*/) const
+inline void Cell::Visit(CellCoord const& standing_cell, TypeContainerVisitor<T, CONTAINER>& visitor, Map& map, WorldObject const& obj, float radius) const
 {
     //we should increase search radius by object's radius, otherwise
     //we could have problems with huge creatures, which won't attack nearest players etc
-    Visit(standing_cell, visitor, map, obj.GetPositionX(), obj.GetPositionY(), radius + obj.GetCombatReach(), extraRadius);
+    Visit(standing_cell, visitor, map, obj.GetPositionX(), obj.GetPositionY(), radius + obj.GetCombatReach());
 }
 
 template<class T, class CONTAINER>
-inline void Cell::Visit(CellCoord const& standing_cell, TypeContainerVisitor<T, CONTAINER>& visitor, Map& map, float x_off, float y_off, float radius, bool extraRadius /*= false*/) const
+inline void Cell::Visit(CellCoord const& standing_cell, TypeContainerVisitor<T, CONTAINER>& visitor, Map& map, float x_off, float y_off, float radius) const
 {
     if (!standing_cell.IsCoordValid())
         return;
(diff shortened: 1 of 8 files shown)


src/server/game/Grids/Cells/CellImpl.h
```diff
@@ -81,6 +81,12 @@ inline void Cell::Visit(CellCoord const& standing_cell, TypeContainerVisitor<T,
         map.Visit(*this, visitor);
         return;
     }
+
+    // We need in some cases increase search radius
+    // Allow find creatures with huge combat reach in a diferent nearby cell
```
Nyr97, line 86: Spelling.

Faq, line 86:
```suggestion
    // Allow find creatures with huge combat reach in a different nearby cell
```

Jildor: thx, done ;)

Shauren:
Why are you adding extra argument to `Cell::Visit*` when all it does is modify another of its arguments? Just do it at call site, in spell code

Jildor:
You are right, it is simpler.
I'll change it tomorrow, thanks :)

Jildor:
Updated :)
I have added a check to add the extra search radius if the default range is greater than 0, is for don't skip this check: https://github.com/TrinityCore/TrinityCore/blob/cb15c9111068f60ed28394cc056c15863ad83813/src/server/game/Grids/Cells/CellImpl.h#L79
If you think that check is not necessary, please tell me and I will remove it

Seyden:
Thats not a valid fix if we have creatures with a combat radius which is bigger than 2 cells right? Such as a lot of cataclysm bosses and stuff

Jildor:
It is focused on the 3.3.5 branch.
I have no idea of the combat reach of the creatures that are in the master branch, in case there were broader combat reach than in the 3.3.5 branch it would be to increase that value for master, but in 3.3.5 40 seems enough to me.

Shauren:
The highest combatreach value in 3.3.5 creature_model_info is 77

Jildor:
I know, but I did not see the need to increase the value for a creature that is not worth it, anyway, I can increase it to the value you tell me, I don't have any problem, but as I said at the beginning, it seems enough to me 40

jackpoz (approved):
Looks good to me, letting Shauren merge it.
Even if there are some edge cases with 77 yards that might still not work, 40 y is good enough of an improvement.

PR #27833, opened 2022-02-26
Tool/MMAP: Fixed invalid check causing certain tiles not to build

robinsch:
example: mmaps_generator 230 will cause tile 33, 29 to not build because of the check. causing NPCs to not have mmap information in some parts of the map.

.go xyz 1380 -831 -88 230 has no mmap tile information because of above check.
https://i.imgur.com/Pypeu4f.png

**Changes proposed:**

-  removed invalid check

**Issues addressed:**

**Tests performed:**

in-game

- [ ] 
- [ ]

### DIFF
--- a/src/tools/mmaps_generator/MapBuilder.cpp
+++ b/src/tools/mmaps_generator/MapBuilder.cpp
@@ -858,12 +858,10 @@ namespace MMAP
                 //printf("%sNo vertices to build tile!              \n", tileString);
                 break;
             }
-            if (!params.polyCount || !params.polys ||
-                TILES_PER_MAP*TILES_PER_MAP == params.polyCount)
+            if (!params.polyCount || !params.polys)
             {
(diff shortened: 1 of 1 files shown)


robinsch:
no clue how TC handles it, but these mmaps tiles were just missing before. so yeah you need to rerun the extractor to get them.

jackpoz:
> better increase mmaps version, no?

after all the mmaps versions, we just tell users to re-generate them and it's up to them to do it

jackpoz:
Thanks for the PR 🚀 😃

HelloKitty:
I went through TrinityCore, MaNGOS and CMaNGOS history in an attempt to find the purpose of this code. Unfortunately, a large portion of the navmesh system was commited as a single commit long ago. A good arguement for not squashing history maybe? Either way, it makes no sense why this existed and there is no explanation for it in any git history of any project. It seems like some sort of strange practical joke from long ago.

PR #27877, opened 2022-03-20
Core/Object: Fix WorldObject::SetZoneScript unsafe cast to InstanceMap

HelloKitty:
**Changes proposed:**

-  Change unsafe casting in WorldObject::SetZoneScript

**Issues addressed:**

I encountered a very special case of this being called from MapInstanced, during some rather experimental changes, and technically Map subtypes of MapInstanced will indicate that IsDungeon is true. Therefore, this unsafe/invalid cast would go through and cause a crash.

**Tests performed:**

No real tests, it seems correct just from a glance.

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1982,8 +1982,8 @@ void WorldObject::SetZoneScript()
 {
     if (Map* map = FindMap())
     {
-        if (map->IsDungeon())
-            m_zoneScript = (ZoneScript*)((InstanceMap*)map)->GetInstanceScript();
+        if (InstanceMap* instanceMap = map->ToInstanceMap())
+            m_zoneScript = reinterpret_cast<ZoneScript*>(instanceMap->GetInstanceScript());
         else if (!map->IsBattlegroundOrArena())
         {
             if (Battlefield* bf = sBattlefieldMgr->GetBattlefieldToZoneId(GetZoneId()))


HelloKitty:
It will definitely crash otherwise, an example crash callstack:

```
Call stack:
Address   Frame     Function      SourceFile
0000024AF59E1C90  000000B1378FF460  0000:0000000000000000 
00007FF7D3D2BA1F  000000B1378FF5B0  GameObject::Create+2BF  C:\Users\Glader\Documents\Github\Another.TrinityCore\src\server\game\Entities\GameObject\GameObject.cpp line 279
00007FF7D3D2E4A8  000000B1378FF6F0  GameObject::LoadFromDB+2E8  C:\Users\Glader\Documents\Github\Another.TrinityCore\src\server\game\Entities\GameObject\GameObject.cpp line 1063
00007FF7D42127F6  000000B1378FF750  LoadHelper<GameObject>+86  C:\Users\Glader\Documents\Github\Another.TrinityCore\src\server\game\Grids\ObjectGridLoader.cpp line 129
00007FF7D4213855  000000B1378FF7D0  ObjectGridLoader::Visit+D5  C:\Users\Glader\Documents\Github\Another.TrinityCore\src\server\game\Grids\ObjectGridLoader.cpp line 143
00007FF7D4212DD5  000000B1378FF800  VisitorHelper<ObjectGridLoader,GameObject,TypeList<Creature,TypeList<DynamicObject,TypeList<Corpse,TypeNull> > > >+15  C:\Users\Glader\Documents\Github\Another.TrinityCore\src\server\game\Grids\Dynamic\TypeContainerVisitor.h line 51
00007FF7D42132CE  000000B1378FF8D0  ObjectGridLoader::LoadN+9E  C:\Users\Glader\Documents\Github\Another.TrinityCore\src\server\game\Grids\ObjectGridLoader.cpp line 190
00007FF7D3EAE422  000000B1378FFA70  Map::Update+2D2  C:\Users\Glader\Documents\Github\Another.TrinityCore\src\server\game\Maps\Map.cpp line 826
00007FF7D4192519  000000B1378FFAA0  MapInstanced::Update+19  C:\Users\Glader\Documents\Github\Another.TrinityCore\src\server\game\Maps\MapInstanced.cpp line 55
00007FF7D419307B  000000B1378FFAE0  MapUpdater::WorkerThread+11B  C:\Users\Glader\Documents\Github\Another.TrinityCore\src\server\game\Maps\MapUpdater.cpp line 119
00007FF7D419285F  000000B1378FFB10  std::thread::_Invoke<std::tuple<void (__cdecl MapUpdater::*)(void),MapUpdater *>,0,1>+F  C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\thread line 56
00007FF8318F1BB2  000000B1378FFB40  _configthreadlocale+92
00007FF8324D7034  000000B1378FFB70  BaseThreadInitThunk+14
00007FF834202651  000000B1378FFBF0  RtlUserThreadStart+21
```

Generally this callstack would never happen ever in normal operation **but** no reason not to correct it anyway.

Ovahlord:
which gameobject exactly did you try to load?

HelloKitty:
> which gameobject exactly did you try to load?

None in particular. This callstack originates from some work I did involving loading all grids/cells and updating them all every tick. TrinityCore does not do this by default. If someone tried without realizing that MapInstanced would also attempt to load grids if you tried then, since it too calls the base Map::Update, then you'd end up calling WorldObject::SetZoneScript and encounter this invalid and unsafe cast. There is no problem with a specific GameObject but the code as is, even though never would run in normal cases, it incorrect. Unsafe. Leads to a crash. I suggest this PR as a fix for this.

PR #27903, opened 2022-04-06
Scripts/Trial of the Crusader: Fix achievement A Tribute to Immortality

Jildor:
**Changes proposed:**
- Now instance achievement A Tribute to Immortality status will be saved into DB
achievement should be completed without any raid member die during all bosses encounters, if any member die, but before kill last boss you leave instance and wait until is destroyed (Instance.UnloadDelay) or do a server restart, TributeToImmortalityEligible will default (true) again because it's not saved.
- Also add support to save instance achievement status of A Tribute to Dedicated Insanity (NYI)

**Issues addressed:**

None

**Tests performed:**

tested in-game


jackpoz:
Thanks for the PR 🚀 😃

Issue #24846, opened 2020-06-19
[3.3.5] Possible out-of-order Packets on Login?


HelloKitty:
**Description:**

TrinityCore 3.3.5 may be sending packets potentially out of order on login. However, since I do not have retail 3.3.5 packet logs I cannot personally verify this.

For example, when logging in Player::LoadFromDB will eventually call Player::SetDrunkValue which invokes UpdateObjectVisibility. This is before the player is even in the map, but it will trigger an update packet from SendToSelf. Packets such as SMSG_POWER_UPDATE are also triggered from LoadFromDB. I think these may have been in Player::SendInitialPacketsBeforeAddToMap before.

I'm also pretty sure I'm getting Object Create blocks for many units before SMSG_LOGIN_VERIFY_WORLD too. Does anyone have retail packet logs to confirm if this behavior is correct?

**Expected behaviour:**

Unknown but I suspect that TrinityCore should not be sending SMSG_COMPRESSED_UPDATE_OBJECT before SMSG_LOGIN_VERIFY_WORLD.

**Steps to reproduce the problem:**

1. Login.

**Branch(es):**

CHANGEME 3.3.5, master or both

**TC rev. hash/commit:** 0e9eb8e7f5eac85531ce25d214cf6018bef4114b

CHANGEME Copy the result of server debug command (if you need to run it from client get prat addon)

**Operating system:** Windows

Rothend:
Another issue related to wrong packet order at login is that often, when you create a new character and log it in, you see a flash of your character in the flashing area before the opening cinematic starts. Sometimes you also see a flash of the position the flyby camera NPC is at, with some visual glitches.

Killyana:
@Rothend it could be the cause of https://github.com/TrinityCore/TrinityCore/issues/22821

HelloKitty:
Tonight I confirmed that this issue can cause GameObjects to render incorrectly on the client

https://user-images.githubusercontent.com/5829095/163331325-17786889-d137-4391-ba5c-190368a548f5.mp4


I will PR a hack fix which can address this. Understandable if it gets rejected since there is truly a systemic issue of sending out of order packets on login/worldport. But as far as I can tell it fixes this bug.

HelloKitty:
Another example of the bug this causes:
![af12a39067285dbcc34aa7c891581317](https://user-images.githubusercontent.com/5829095/163337432-334839ce-b3ec-4846-b35b-72ea461946b1.png)

[...]

Issue #20401, opened 2017-09-23
3.3.5 gameobjects color


Frenlock:
**Description:**

Sometimes some gameobjects : mailbox , barrels , flags , indicators ( idk if all are gameobjects ) 
have a strange color like a dark-brown shadow over them . 
I tryed to re-download the game client , clear cache folder (check video settings) but didn't helped so I think that its not client-side related .
https://imgur.com/a/ID1y0

**Current behaviour:**

The gameobjects have that dark-brown color .

**Expected behaviour:**

The gameobjects shouldn't have that dark-brown color .

**Steps to reproduce the problem:**

1. Login and logout several times until the problem appear .
2. If u go through portals / use hearthstone or teleport the problem may dissapear .

**Branch(es):** 3.3.5

**TC rev. hash/commit:** 
TrinityCore rev. 4542d93 2017-09-21 16:26:32 +0200 (3.3.5 branch) (Win32, Release, Static)

**TDB version:**  TDB_full_335.63_2017_04_18

**Operating system:** Windows 10 x64

Shauren:
This is definitely client related. There is zero model rendering related code on the server.

Aokromes:
if you are running windows 10 insider build you will have this problem.

Frenlock:
Hmm I running on Windows 10 Home x64

CraftedRO:
Btw this is still happening on (same Windows 10 Home x64 21H1 (OS Build 19043.1237) with a clean wow client.
TrinityCore rev. c8061d8ba620 2021-09-19 02:15:09 +0300 (3.3.5 branch) (Unix, RelWithDebInfo, Static) - Aowow trinity server.

![WoWScrnShot_092321_130537](https://user-images.githubusercontent.com/24683355/134489965-cff436b5-473c-471e-bb00-fe20848e9536.jpg)
![WoWScrnShot_092321_130548](https://user-images.githubusercontent.com/24683355/134489976-85ca5f84-234d-46fd-86f1-e91ad2942319.jpg)

Aokromes:
it's something on your settings.
https://i.imgur.com/K9Kxg5H.jpeg

CraftedRO:
well it can be but then why with same settings if I relog / enter portals , the problem is gone ?

ihm-tswow:
What's your GPU / graphics driver version? Would be good to collect this in case it happens to more people. There's a few other known issues with modern drivers

ihm-tswow:
[...]

PR #27914, opened 2022-04-14
Core/Player: Prevent Forced UpdateObjectVisibility Before in World

HelloKitty:
It's critical for the client to not recieve SMSG_UPDATE_OBJECT object create blocks before the world is loaded. It can cause some very mind boggling and subtle bugs.

**Changes proposed:**

-  Preventing immediate handling for Player::UpdateObjectVisibility when not in world

**Issues addressed:**

This partially addresses this 2 year old issue I raised: https://github.com/TrinityCore/TrinityCore/issues/24846

https://github.com/TrinityCore/TrinityCore/issues/20401

**Tests performed:**

One of the most obvious examples of bugs this issue causes is strange lighting/coloring/rendering of GameObjects that initially spawn on your client. See the below: 

https://user-images.githubusercontent.com/5829095/163336454-4b63047c-d6c0-4990-aa53-9e6e976e7d76.mp4

af12a39067285dbcc34aa7c891581317

After this commit these objects no longer appear incorrectly colored. This probably fixes other subtle bugs too that happen around logging in or worldporting.

Obviously this change is a very low level Core change. I will have trouble fully testing the impact, sorry. Maybe this negatively affects something.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -22550,7 +22550,8 @@ template void Player::UpdateVisibilityOf(DynamicObject* target, UpdateData& data
 
 void Player::UpdateObjectVisibility(bool forced)
 {
-    if (!forced)
+    // Sending visibility updates to the client before in world causes strange behavior
+    if (!forced || !IsInWorld())
         AddToNotify(NOTIFY_VISIBILITY_CHANGED);
     else
     {


HelloKitty:
I encountered this during testing the PR, possibly a regression? But I'm not really sure if it is or not. Let me revert and test to find out: https://i.gyazo.com/4005455bf9415b742af342207e394e94.mp4

jackpoz:
is that method used to tell the player to refresh how the objects around look, or is it used to tell other objects around to refresh how the player looks ?

HelloKitty:
> is that method used to tell the player to refresh how the objects around look, or is it used to tell other objects around to refresh how the player looks ?

Afaik it basically takes the player as a reference point (if force it used) and iterates the cells nearby and calls UpdateVisibilityOf on nearby players. Otherwise they're scheduled for relocation notifying.

```
template<class T>
inline void Trinity::VisibleNotifier::Visit(GridRefManager<T> &m)
{
    for (typename GridRefManager<T>::iterator iter = m.begin(); iter != m.end(); ++iter)
    {
        vis_guids.erase(iter->GetSource()->GetGUID());
        i_player.UpdateVisibilityOf(iter->GetSource(), i_data, i_visibleNow);
    }
}
```

Are you familiar with this bug or is this a regression https://i.gyazo.com/4005455bf9415b742af342207e394e94.mp4 ? I haven't been able to reproduce it more than once.

HelloKitty:
Tbh running in place could have been because of breakpoints/debugger. I find it impossible to reproduce it a second time.

Ovahlord:
I wonder if this can be connected to cmsg_loading_screen_notify, which might be used to inform the server that the client is ready

CraftedRO:
> Issues addressed:
> 
> This partially addresses this 2 year old issue I raised

Actually :)) https://github.com/TrinityCore/TrinityCore/issues/20401

CraftedRO:
[Sigh...](https://github.com/TrinityCore/TrinityCore/issues/20401)

HelloKitty:
> [Sigh...](https://github.com/TrinityCore/TrinityCore/issues/20401)

Ohhh, thank you for linking. I didn't know the issue was reported. That issue as well as "Out of Order Packets on Login" are definitely related! Dang, 2017 is quite a long while ago.

Takenbacon:
> I wonder if this can be connected to cmsg_loading_screen_notify, which might be used to inform the server that the client is ready

I don't believe such an opcode exists in WoTLK. If I recall right during initial login loading screen there is no client acks, it just waits for server data.

For map changes I did one day accidentally learn it's super important to wait for MSG_MOVE_WORLDPORT_ACK before sending any map object related data.

jackpoz:
Is there still some research going on for this PR and should it be changed to draft or is it ready to be merged ?

HelloKitty:
> Is there still some research going on for this PR and should it be changed to draft or is it ready to be merged ?

I consider it complete, nothing left for me to do. Though I've not tested it on live servers I haven't personally encountered any regressions on my fork.

qyh214:
I use nVidia 2080, I thought it was the graphics driver issue,so it is server issue?
and also npc sometimes also shows black. but switch map will show normal.

CraftedRO:
> I use nVidia 2080, I thought it was the graphics driver issue,so it is server issue? and also npc sometimes also shows black. but switch map will show normal.

No dude it never was a gpu or windows issue , but some people just prefer to close valid issues when they doesn't understand what's about.

kaytotes:
> Is there still some research going on for this PR and should it be changed to draft or is it ready to be merged ?

I have a few users on a TS-WoW based server that could consistently reproduce this often enough to be a nuisance. 

I've deployed a build today with these changes applied and will do some rudimentary testing.

kaytotes:
So far so good. For the users who could pretty consistently reproduce this it hasn't happened again. Haven't noticed any weird after effects or stability issues.

Shauren:
Alright, I finally had time to look at this and compare code between 3.3.5 and master branches. Turns out master branch already had this fixed over 7 years ago in 3ec88b7086953685617c2ccd79207a388d20e129

In my opinion that old commit is actually a better fix - we don't want to move those visibility updates after login is completed but skip them entirely. That is because the last step of logging in is `Map::AddPlayerToMap`, which sends all the neccessary (and up-to-date) data

HelloKitty:
> Alright, I finally had time to look at this and compare code between 3.3.5 and master branches. Turns out master branch already had this fixed over 7 years ago in [3ec88b7](https://github.com/TrinityCore/TrinityCore/commit/3ec88b7086953685617c2ccd79207a388d20e129)
> 
> In my opinion that old commit is actually a better fix - we don't want to move those visibility updates after login is completed but skip them entirely. That is because the last step of logging in is `Map::AddPlayerToMap`, which sends all the neccessary (and up-to-date) data

I was unaware that someone else noticed this years ago. Seems like someone came to a similar conclusion back then but I'm not sure they realized that it fixes afew odd bugs. Either way works, the behavior should be the same. Up to you which version you want in the 3.3.5 branch. I initially worte it that way too but decided to change the behavior as minimally as possible. Setting visibility/notify bits to be handled later seemed like the minimal change.

PR #27921, opened 2022-04-21
Tool/MMAP: Fixed possible race condition in MMapManager::loadMap

robinsch:
**Changes proposed:**

-  Lock when loading mmaps

**Issues addressed:**

Can be called from multiple threads (instances loading same grid for example)

**Tests performed:**
Build


Shauren:
did you catch a stack trace?

robinsch:
No, but relocation to another grid can call this from multiple map threads writing to the same container. Same for VMaps (did not PR since I preload all maps and vmaps)

src/common/Collision/Management/MMapManager.h
```diff
@@ -24,6 +24,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <mutex>
```
jackpoz, line 27:
```suggestion
#include <mutex>
#include <string>
#include <unordered_map>
#include <vector>
```

jackpoz: we sort headers logically (TC, external libraries, std) and then alphabetically

robinsch:
Still two maps can call it at the same time @jackpoz

Shauren:
> Still two maps can call it at the same time @jackpoz

No, they cannot.

When an instance map loads a grid this happens
* instance->EnsureGridCreated
* instance->LoadMapAndVMap
* instance->LoadMap
* m_parentMap->EnsureGridCreated <---- ACQUIRES _gridLock of m_parentMap
* m_parentMap->LoadMapAndVMap
* ...
* m_parentMap->LoadMMap
* m_parentMap->EnsureGridCreated finishes <---- RELEASES _gridLock of m_parentMap
* instance->LoadVMap and instance->LoadMMap are skipped because instance->i_InstanceId is not 0

m_parentMap pointers are shared by all instances, loading map, vmap and mmap data happens under _gridLock of m_parentMap so it is threadsafe.

robinsch:
wops, I missed that lock somehow.

PR #27930, opened 2022-04-24
Core/ItemHandler: fix selling items over gold limit

okfok:
Changes proposed:

Make it impossible to sell items if it will cause exceeding gold limit

Issues:

when you sell items over gold limit that item will be sold but you wont get money for it.

Tests performed:

Build successful, 
tested in-game

### DIFF
--- a/src/server/game/Handlers/ItemHandler.cpp
+++ b/src/server/game/Handlers/ItemHandler.cpp
@@ -440,6 +440,13 @@ void WorldSession::HandleSellItemOpcode(WorldPacket& recvData)
         ItemTemplate const* pProto = pItem->GetTemplate();
         if (pProto)
         {
+            uint32 money = pProto->SellPrice * count;
+            if (_player->GetMoney() + money > MAX_MONEY_AMOUNT)               // prevent exceeding gold limit
+            {
+                _player->SendEquipError(EQUIP_ERR_TOO_MUCH_GOLD, nullptr, nullptr);
+                _player->SendSellError(SELL_ERR_UNK, creature, itemguid, 0);
+				return;
+            }
             if (pProto->SellPrice > 0)
             {
                 if (count < pItem->GetCount())               // need split items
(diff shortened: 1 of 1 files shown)


src/server/game/Handlers/ItemHandler.cpp
```diff
@@ -440,6 +440,13 @@ void WorldSession::HandleSellItemOpcode(WorldPacket& recvData)
         ItemTemplate const* pProto = pItem->GetTemplate();
         if (pProto)
         {
+            uint32 money = pProto->SellPrice * count;
+            if (_player->GetMoney() + money > MAX_MONEY_AMOUNT)               // prevent exceeding gold limit
+            {
+                _player->SendEquipError(EQUIP_ERR_TOO_MUCH_GOLD, nullptr, nullptr);
+                _player->SendSellError(SELL_ERR_UNK, creature, itemguid, 0);
+                return;
+            }
```
jackpoz, line 449: this code should be in the `if (pProto->SellPrice > 0)` body.

src/server/game/Handlers/ItemHandler.cpp
```diff
@@ -440,6 +440,13 @@ void WorldSession::HandleSellItemOpcode(WorldPacket& recvData)
         ItemTemplate const* pProto = pItem->GetTemplate();
         if (pProto)
         {
+            uint32 money = pProto->SellPrice * count;
+            if (_player->GetMoney() + money > MAX_MONEY_AMOUNT)               // prevent exceeding gold limit
```
jackpoz, line 444: Check similar code like https://github.com/TrinityCore/TrinityCore/blob/0d8672732552b0474e22b077a2318fa87446f25f/src/server/game/Handlers/TradeHandler.cpp#L299 to make sure `_player->GetMoney() + money` doesn't overflow (and please test it with the overflow case to verify it works as expected)

src/server/game/Handlers/ItemHandler.cpp
okfok: ok, i will move it

src/server/game/Handlers/ItemHandler.cpp
okfok: yep, it has trouble when item costs 1 copper

okfok: fixed

jackpoz (approved):

jackpoz:
Thanks for the PR 🚀 😃

PR #27937, opened 2022-05-03
Scripts/Naxxramas: Fix achievement The Immortal / The Undying

Jildor:
**Changes proposed:**

-  Now instance achievement status of The Immortal / The Undying will be saved into DB
achievement should be completed without any raid member die during all bosses encounters, if any member die, but before kill last boss you leave instance and wait until is destroyed (Instance.UnloadDelay) or do a server restart, playerDied will default 0 again because it's not saved.
-  this was reverted after https://github.com/TrinityCore/trinitycore/commit/ef9b4aea5e97ffbd4c0f611bef6fab6e554d82ff#diff-42d3a9126aeb07fb2ad87f5d8513c35eb4ed7082aaaf5af233de8a12e3c1728cL468

**Issues addressed:**

None

**Tests performed:**

tested in-game


joschiwald:
Great spotting, thanks.

PR #27940, opened 2022-05-04
Core/Spells: Implemented SPELL_ATTR1_IGNORE_OWNERS_DEATH

tisk:


Shauren:
This is incorrect, this attribute is not related to auras (and definitely does not make aura persist through death, tested with hunter pvp talent https://www.wowhead.com/spell=356719/chimaeral-sting)

jackpoz:
> No description provided.

It's good practice to include a description when opening a PR. There is a template that should be filled in, not deleted

PR #27973, opened 2022-05-24
Core/SAI: Added SMART_TARGET_OWNER_OR_SUMMONER_VICTIM

BertiRean:
**Changes proposed:**

-  
-  
-  

**Issues addressed:**

**Tests performed:**

- Cast the spell 201002 on self. Add a basic sai action on random hostile like cast. You will see that creature don't attack to hostile targets with your action.

- [ ]  
- [ ] 

The SMART_TARGET_HOSTILE, can't work in summons with Guardian Mask. Because the Method ThreatManager::CanHaveThreatList(Unit const* who) return false always. 

In retail the summons with property control Guardian can attack normally to hostile units.  In TC we can't have ThreatList and this makes useless SAI_TARGETS that work with threat list. 

We have some options to solve the issue:

- Add a new Target in SAI to get Victim of owner or summoner.
- Set ThreatList enabled for Guardian mobs created via Spells.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -2956,6 +2956,26 @@ void SmartScript::GetTargets(ObjectVector& targets, SmartScriptHolder const& e, 
                 targets.push_back(target);
             break;
         }
+        case SMART_TARGET_OWNER_OR_SUMMONER_VICTIM:
+        {
+            if (me)
+            {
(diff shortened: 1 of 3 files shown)


BertiRean:
This part of code is the cause of current issue. https://github.com/TrinityCore/TrinityCore/blob/aa592e431b945800928f2e43b3812cfa240fada1/src/server/game/Combat/ThreatManager.cpp#L151

PR #27976, opened 2022-05-25
Update a few SMSG packets for 3.4.0.43659

MaxtorCoder:
**Changes proposed:**

- Update SMSG_FEATURE_SYSTEM_STATUS
- Update SMSG_FEATURE_SYSTEM_STATUS_GLUE
- Update SMSG_SEASON_INFO

**Tests performed:**
- Does not build because of other changes

### DIFF
--- a/src/server/game/Server/Packets/SystemPackets.cpp
+++ b/src/server/game/Server/Packets/SystemPackets.cpp
@@ -122,7 +122,7 @@ WorldPacket const* FeatureSystemStatus::Write()
     _worldPacket.WriteBit(ChatDisabledByPlayer);
     _worldPacket.WriteBit(LFGListCustomRequiresAuthenticator);
     _worldPacket.WriteBit(RaceClassExpansionLevels.has_value());
-    _worldPacket.WriteBit(BattlegroundsEnabled);
+    _worldPacket.WriteBit(WarGameEnabled);
 
     _worldPacket.FlushBits();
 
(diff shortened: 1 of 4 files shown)


src/server/game/Server/Packets/SystemPackets.cpp
```diff
@@ -121,6 +121,8 @@ WorldPacket const* FeatureSystemStatus::Write()
     _worldPacket.WriteBit(ChatDisabledByDefault);
     _worldPacket.WriteBit(ChatDisabledByPlayer);
     _worldPacket.WriteBit(LFGListCustomRequiresAuthenticator);
+    _worldPacket.WriteBit(RaceClassExpansionLevels.has_value());
+    _worldPacket.WriteBit(BattlegroundsEnabled);
```
Ghaster, line 125: I believe this bit only enable/disable War Games and not battlegrounds entirely

MaxtorCoder: > I believe this bit only enable/disable War Games and not battlegrounds entirely

```cpp
  if ( !byte_143621604 )
  {
    CGGameUI::DisplayError(1039);
    return 0i64;
  }
```
This is the check that checks the bool value and the value 1039 corresponds to `ERR_BATTLEGROUND_NOT_ENABLED`

MaxtorCoder: Update: Indeed related to wargame, blame blizzard for not naming the error code correctly.

PR #27992, opened 2022-06-03
Core/Player: GetWeaponForAttack ranged check

Smerdokryl:
**Changes proposed:**

-  Distinguish between ranged and non-ranged weapons in GetWeaponForAttack (since ranged weapons no longer have a special equipment slot)


Shauren:
But what does this fix?

Smerdokryl:
> But what does this fix?

`Player::GetWeaponForAttack(RANGED_ATTACK)` returning a sword*

PR #27993, opened 2022-06-03
Tools/Extractors: Implement remote casc mode

Luzifix:
**Changes proposed:**
- Implement remote casc feature in mapextractor and vmapexport
- Add -c argument to activate remote casc
- Add -r argument to set the remote casc region (default is eu)
- Add -dl argument to vmapexport to select the corrent dbc locale 

**Tests performed:**
- It build and export all nesessary files

### DIFF
--- a/src/tools/extractor_common/CascHandles.cpp
+++ b/src/tools/extractor_common/CascHandles.cpp
@@ -190,6 +190,29 @@ CASC::Storage* CASC::Storage::Open(boost::filesystem::path const& path, uint32 l
     return storage;
 }
 
+CASC::Storage* CASC::Storage::OpenRemote(boost::filesystem::path const& path, uint32 localeMask, char const* product, char const* region)
+{
+    HANDLE handle = nullptr;
+    std::string cacheArgument = std::string(path.string() + ":" + product + ":" + region);
+
+    printf("Open casc remote storage...\n");
+    if (!::CascOpenOnlineStorage(cacheArgument.c_str(), localeMask, &handle))
+    {
+        DWORD lastError = GetCascError(); // support checking error set by *Open* call, not the next *Close*
+        printf("Error opening remote casc storage: %s\n", HumanReadableCASCError(lastError));
+        CascCloseStorage(handle);
+        SetCascError(lastError);
+        return nullptr;
+    }
+
+    Storage* storage = new Storage(handle);
+
+    if (!storage->LoadOnlineTactKeys())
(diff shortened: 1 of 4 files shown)


src/tools/extractor_common/CascHandles.h
```diff
@@ -40,7 +40,8 @@ namespace CASC
     public:
         ~Storage();
 
-        static Storage* Open(boost::filesystem::path const& path, uint32 localeMask, char const* product);
+		static Storage* Open(boost::filesystem::path const& path, uint32 localeMask, char const* product);
+		static Storage* OpenRemote(uint32 localeMask, char const* product);
```
MaxtorCoder, line 44:
```suggestion
		    static Storage* Open(boost::filesystem::path const& path, uint32 localeMask, char const* product);
		    static Storage* OpenRemote(uint32 localeMask, char const* product);
```

src/tools/extractor_common/CascHandles.cpp
```diff
@@ -190,6 +190,31 @@ CASC::Storage* CASC::Storage::Open(boost::filesystem::path const& path, uint32 l
     return storage;
 }
 
+CASC::Storage* CASC::Storage::OpenRemote(uint32 localeMask, char const* product)
+{
+    HANDLE handle = nullptr;
+    boost::filesystem::path cachDir = boost::filesystem::current_path() / "Cache";
```
Shauren, line 196: I recommend not naming it `Cache` but something like `CascCache` or `ExtractorCascCache` to avoid clashing with clients own `Cache` directory if ran from where the exes are

src/tools/extractor_common/CascHandles.cpp
```diff
@@ -190,6 +190,31 @@ CASC::Storage* CASC::Storage::Open(boost::filesystem::path const& path, uint32 l
...
+
+    std::string cacheArgument = std::string(cachDir.string() + ":" + product + ":eu");
```
MaxtorCoder, line 198: I also would make some sort of argument to change the locale here

Luzifix: do you mean the region eu/us?

MaxtorCoder: Yes

src/tools/extractor_common/CascHandles.h
MaxtorCoder: Still not indented correctly

Issue #25342, opened 2020-08-27
Combat animations not working properly


ghost:
**Description:**

Created a dwarf hunter and weirdly combat animations arent working properly. You only get a damage visual on how much damage the mob has done to you but you cant see any animation whatsoever. Happens only when you dont move yourself and only when you attack with ranged weapon.

**Expected behaviour:**

Well it should show all combat animation :)

**Steps to reproduce the problem:**

1. Create a character (dwarf hunter at best)
2. Shoot at the enemies with ranged weapon with quiet a distance (DONT MOVE!)
3. A high chance that combat animation wont be played but enemy will still do damage

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/be73959a0d838f71f9ff995486479fdffb7ab013

**Operating system:**  Windows 10

Aokromes:
you write 3.3.5a branch, but you link master branch hash.

felwyn:
Sorry corrected!

offl:
Sounds like something opposite to https://github.com/TrinityCore/TrinityCore/issues/11854

kaytotes:
This SetFacingToObject call: https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Unit/Unit.cpp#L2054-L2055
Added in: https://github.com/TrinityCore/TrinityCore/commit/f481ae1048aac32ba3a40491f0304d0317234295

With that present I can pretty consistently recreate the stuttering attack animation. Not every creature but maybe 1 in 4. More easily reproducable if combat starts in melee range. Numerous times within a 30 minute window, tested in human starting area against Kobold Workers.

With it removed I've done roughly 1 hour of testing in the same way and not managed to replicate the bug.

Re-added it, able to reproduce in about 5 minutes again.

@Treeston could you confirm?

Shauren:
eab6852b686ffaafa3aefb09b9c310c453db55aa

PR #28006, opened 2022-06-05
Creature Animation: Remove Set Facing Call

kaytotes:
**Changes proposed:**

-  Remove `SetFacingToObject` call from `AttackerStateUpdate`

Added in: https://github.com/TrinityCore/TrinityCore/commit/f481ae1048aac32ba3a40491f0304d0317234295

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/25342

**Tests performed:**

With that present I can pretty consistently recreate the stuttering attack animation. Not every creature but maybe 1 in 4. More easily reproducable if combat starts in melee range. Numerous times within a 30 minute window, tested in human starting area against Kobold Workers.

With it removed I've done roughly 1 hour of testing in the same way and not managed to replicate the bug.

Re-added it, able to reproduce in about 5 minutes again.

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -2096,9 +2096,6 @@ void Unit::AttackerStateUpdate(Unit* victim, WeaponAttackType attType, bool extr
     if (!extra && _lastExtraAttackSpell)
         _lastExtraAttackSpell = 0;
 
-    if (GetTypeId() == TYPEID_UNIT && !HasUnitFlag(UNIT_FLAG_POSSESSED) && !HasUnitFlag2(UNIT_FLAG2_CANNOT_TURN))
-        SetFacingToObject(victim, false); // update client side facing to face the target (prevents visual glitches when casting untargeted spells)
-
     // melee attack spell cast at main hand attack only - no normal melee dmg dealt
     if (attType == BASE_ATTACK && m_currentSpells[CURRENT_MELEE_SPELL] && !extra)
         m_currentSpells[CURRENT_MELEE_SPELL]->cast();


Treeston:
Re-test Ingvar the Plunderer with this removed, if my vague memory is correct that's why we added it.

(Have him cast his cone AoE on a second target, and see if his facing is still correct afterwards, without the attack target moving.)

kaytotes:
Does this help? https://streamable.com/ob7h00

Been a while since I've done that fight, some other encounter bugs at play but Dark Smash seems okay

Ovahlord:
the facing should be done via target update field and spamming facing splines on each melee attack is downright a hack so removing it is correct.

Treeston:
> the facing should be done via target update field and spamming facing splines on each melee attack is downright a hack so removing it is correct.

very possible, spellcast focus was my first big core PR and not something I'm too proud of in retrospect.

Shauren:
I have one question - is the serverside facing correct with this line removed? (client facing should rely on TARGET updatefield alone as @Ovahlord mentioned and doesnt need to send any packets)

kaytotes:
Is there a specific way for me to test that? I spent an hour or so massing with Ingvar and didn't notice anything majorly off.

Shauren:
I believe you would be looking at failing autoattacks due to facing if it was incorrect

kaytotes:
I will spend a few hours testing what I can keeping an eye on combat log

Shauren:
You would be wasting your time doing that, use debugger to inspect orientation value during AttackerStateUpdate
But I can save you even more time, just checked how AttackerStateUpdate is called for creatures - facing is never checked

kaytotes:
Is there anything more I can do to get this over the line? Done a fair amount of testing and not managed to spot any real issues.

Issue #26734, opened 2021-07-22
Core/Gossip: Default gossip text


offl:
### Description

If creature has no gossip_menu_id filled and has no gossip flag nor quest_greeting, when you accept two quest from the same creature, creature displays default gossip text
```cpp
    if (!gossip)
    {
        for (uint8 i = 0; i < MAX_GOSSIP_TEXT_OPTIONS; ++i)
        {
            data << float(0);
            data << "Greetings $N";
            data << "Greetings $N";
```
Since the text is not taken from npc_text, it's not translated
I don't know if in such conditions any text should be displayed or not
At the same time we currently have a bunch of creatures with gossip_menu_id not filled but with gossip NPC flag, that too forces creature to display default text. In this case we probably need that default text
```sql
SELECT * FROM creature_Template WHERE npcflag & 1 AND !gossip_menu_id
```

### Expected behaviour

If that default text should not be displayed, it should be removed. If it should be displayed, creature should send npc_text to make locales work
```txt
[...]

PR #28016, opened 2022-06-10
Core/Player: Added fallback to regular gossip menu if no quest_greeting is found

mdX7:
**Changes proposed:**

-  Fallback to showing gossip menu when no quest_greeting is found, without fallback it's not showing any text besides quest options

**Tests performed:**

tested ingame with npc ids: 39379, 64764

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -14749,7 +14749,11 @@ void Player::SendPreparedGossip(WorldObject* source)
             return;
         }
     }
+    SendPreparedGossipForced(source);
+}
 
+void Player::SendPreparedGossipForced(WorldObject* source)
+{
     // in case non empty gossip menu (that not included quests list size) show it
     // (quest entries from quest menu will be included in list)
 
@@ -14823,7 +14827,7 @@ void Player::OnGossipSelect(WorldObject* source, uint32 gossipListId, uint32 men
             break;
         case GOSSIP_OPTION_QUESTGIVER:
             PrepareQuestMenu(guid);
-            SendPreparedQuest(source);
+            SendPreparedGossip(source);
             break;
         case GOSSIP_OPTION_VENDOR:
         case GOSSIP_OPTION_ARMORER:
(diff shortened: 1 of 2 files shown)


Shauren:
But the entire point of quest greeting is to display text for creatures that dont have a gossip menu

offl:
Reminds me #26734

mdX7:
My assumption is that quest_greeting is only used for implicit emote related stuff. We also have an extra string field in quest_greeting which will be displayed if we send it as QuestGiverQuestListMessage.
Essentially the only difference between QuestGiverQuestListMessage and regular GossipMessage is that we have additional emote stuff.

The change i did still allows showing QuestGiverQuestListMessage with greeting entries. If we do not have any greeting entry its displayed using regular gossip menu.

I actually debugged thru the entries i posted above for the test and i always ran into QuestGiverQuestListMessage and it displayed no text at all although it should be using regular gossips (according to sniff for 64764)

meji46:
I've encountered the same situation with the NPC https://www.wowhead.com/npc=51934/gretchen-fenlow.

She has gossip_menu and offers several quests but the core currently sends the info in SMSG_QUEST_GIVER_QUEST_LIST_MESSAGE instead of SMSG_GOSSIP_MESSAGE

Shauren:
I have a different theory, what if it only depends on how the interaction was started?

CMSG_GOSSIP_HELLO -> SMSG_GOSSIP_MESSAGE
CMSG_QUEST_GIVER_HELLO -> SMSG_QUEST_GIVER_QUEST_LIST_MESSAGE

meji46:
Well, I was just referring to the current behavior of the core:

- NPC 51934 (Retail) (Gossip_menu)
  - CMSG_TALK_TO_GOSSIP -> SMSG_GOSSIP_MESSAGE
- NPC 51665 (Retail & TC) (Gossip_menu)
  - CMSG_TALK_TO_GOSSIP -> SMSG_QUEST_GIVER_QUEST_DETAILS (1 quest available)
  - CMSG_TALK_TO_GOSSIP -> SMSG_GOSSIP_MESSAGE (no quest available)
- NPC 51934 (TC) (Gossip_menu)
  - CMSG_TALK_TO_GOSSIP -> SMSG_QUEST_GIVER_QUEST_LIST_MESSAGE
  
Possibly how the interaction is started is what determines the response sent by the server. The only visible difference between gossip menus and quest greetings for a player is that once only 1 quest or no quests are available from the npc using the quest greeting, the quest greeting won't be visible anymore

shamage:
If I remember correctly, the current NPCs that have a greeting for the player usually only change the Quests menu (the greeting remains) - if the player has already exhausted (completed the quests) then only the player's greeting is displayed or in some cases the greeting changes to a different text. In exceptional cases, I also encountered the fact that if the player completed all the tasks for a specific NPC, no greeting was available anymore.

Aokromes:
Conflicting files

PR #28017, opened 2022-06-10
Core/Conditions: Added CONDITION_SOURCE_TYPE_POINT_OF_INTEREST

meji46:
**Changes proposed:**

-  Added CONDITION_SOURCE_TYPE_POINT_OF_INTEREST

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


mdX7:
arent those limited by conditions on gossip_menu_option? (just thinking about oribos guards atm)

Shauren:
Honestly if its just 1 special case, I would just write a creature script and manually send the POI from OnGossipSelect

meji46:
It's the only one I've found but I don't know if there are more in, for example, the Brewfest event (which has pois)

Shauren:
Yep, should be scripted

PR #28019, opened 2022-06-11
Core/Scripts: Fix hunter talent Posthaste

### LINKED ISSUE #19599: Core/Script The passive spell talent Posthaste has issues.
**Description:**

The passive spell talent ''Posthaste'' [Spell ID: 109215] of the Hunter class, has issues.

**Current behaviour:**

The passive spell talent ''Posthaste'' [Spell ID: 109215] does not remove impairing effects when using ''Disengage'' [Spell ID: 781] and it doesn't increase movement speed by 60% for 5 sec.

**Expected behaviour:**

The passive spell talent ''Posthaste'' [Spell ID: 109215] should free the Hunter from all impairing effects after using ''Disengage'' [Spell ID: 781] and then it should increase movement speed by 60% for 5 sec.

1. Create a Hunter character.
2. Reach level 10 and choose the spec Marksmanship or Beast Mastery.
3. Then, reach level 45 and choose the passive spell talent ''Posthaste'' [Spell ID: 109215]
4. Use the spell ''Disengage'' [Spell ID: 781] while under any imparing effect.

**Branch(es):**

Master

**TC rev. hash/commit:** 

Description in wowhead http://www.wowhead.com/spell=109215/posthaste

aquadeus:
**Changes proposed:**

-  Fix hunter talent Posthaste

**Issues addressed:**

Closes #19599

**Tests performed:**

Tested in-game

### DIFF
--- a/sql/updates/world/master/2022_06_16_03_world_hunter_posthaste.sql
+++ b/sql/updates/world/master/2022_06_16_03_world_hunter_posthaste.sql
@@ -1,0 +1,2 @@
+DELETE FROM `spell_script_names` WHERE `ScriptName` = 'spell_hun_posthaste';
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES (781, 'spell_hun_posthaste');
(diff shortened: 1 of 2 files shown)


Nyr97:
Leaving it here for the time being: this implementation is wrong. The spell is not cast until the player lands, no matter the distance that is run by the player after launching the Disengage movement, but it is obviously not AfterCast. 

My first idea was Blizzard used a serverside aura with AuraInterruptFlag::OnLanding to control this. There are some serverside spells called Posthaste whose spellIds were added on and post Cataclysm (when this talent was first introduced). It could be convenient to investigate when this AuraInterruptFlag was first introduced in the client's strings. If it was on Cataclysm, there's a high chance that this was the way they handle this talent.

TL;DR: we need to implement a serverside aura that uses AuraInterruptFlag::OnLanding (which also needs implementation through a new hook that is called when that Disengage's MovementType happens) and force CastSpell when OnRemove. Then, reapply this aura AfterRemove so it's always on the player for as long as he has the talent aura.

PR #28030, opened 2022-06-13
[3.3.5] OpenSSL version 3.0 support

damaex:
**Changes proposed:**

- OpenSSL version 3.0 support
- MariaDB support for Ubuntu 22.04

**Issues addressed:**

- Ubuntu 22.04 unsupported

**Tests performed:**

- [x] builds on 20.04/22.04
- [x] run, login and 5min gameplay on 22.04

- [x] none

### DIFF
--- a/cmake/macros/FindOpenSSL.cmake
+++ b/cmake/macros/FindOpenSSL.cmake
@@ -80,7 +80,7 @@ Set ``OPENSSL_MSVC_STATIC_RT`` set ``TRUE`` to choose the MT version of the lib.
 #]=======================================================================]
 
 set(OPENSSL_EXPECTED_VERSION "1.0")
-set(OPENSSL_MAX_VERSION "1.2")
+set(OPENSSL_MAX_VERSION "3.1")
 
 macro(_OpenSSL_test_and_find_dependencies ssl_library crypto_library)
   if((CMAKE_SYSTEM_NAME STREQUAL "Linux") AND
--- a/src/common/Cryptography/ARC4.cpp
+++ b/src/common/Cryptography/ARC4.cpp
@@ -18,21 +18,43 @@
 #include "ARC4.h"
 #include "Errors.h"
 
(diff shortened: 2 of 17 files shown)


Shauren:
Ill have to reject this PR, outside of RC4 (because it doesnt exist there anymore), openssl 3.0 support is done in master branch and should just be ported

Shauren:
bc87f7b337154e683369a3790ee8fd1a7d4cba98 and aa6a0f24312ce6132e55a9aed677362a3b1dc8a9

damaex:
RC4 still exists in the openssl 3.0 legacy provider `OSSL_PROVIDER_load(_libctx, "legacy");`
So, you want me to update the `master`-branch for OpenSSL 3.0 and port it back?

Shauren:
No, master branch is already updated (see commits i linked), all that remains is porting it to 3.3.5

damaex:
Okay, I have seen those and will do my updates accordingly.

Is HMAC used in the master branch??
Because:
https://github.com/TrinityCore/TrinityCore/blob/daafe1c181c47111b30b1ca4f3c3bfc628d1f207/src/common/Cryptography/HMAC.h#L121
Should be: `size_t length = DIGEST_LENGTH;`
According to: https://www.openssl.org/docs/man3.0/man3/EVP_DigestSignFinal.html

> EVP_DigestSign() signs tbslen bytes of data at tbs and places the signature in sig and its length in siglen in a similar way to EVP_DigestSignFinal(). In the event of a failure EVP_DigestSign() cannot be called again without reinitialising the EVP_MD_CTX. If sig is NULL before the call then siglen will be populated with the required size for the sig buffer. If sig is non-NULL before the call then siglen should contain the length of the sig buffer.

Shauren:
Yes, its used and that code works with both 1.1 and 3.0

damaex:
> Yes, its used and that code works with both 1.1 and 3.0

Interestingly this didn't work on my tests and i had to change it.

PR #28036, opened 2022-06-17
Scripts/Events: Selina Dourman

meji46:
**Changes proposed:**

-  Added missing AI, POIs and conditions for Selina Dourman (thanks to @mdX7 for pointing me to use CAST_AI)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None

### DIFF
--- a/src/server/scripts/Events/darkmoon_faire.cpp
+++ b/src/server/scripts/Events/darkmoon_faire.cpp
@@ -99,7 +99,7 @@ struct npc_selina_dourman : public ScriptedAI
                         poiId = POI_FORTUNE_TELLER;
                         break;
                     default:
-                        break; 
+                        break;
                 }
                 if (poiId)
                     player->PlayerTalkClass->SendPointOfInterest(poiId);
@@ -120,7 +120,7 @@ struct npc_selina_dourman : public ScriptedAI
     {
         if (!talkTarget || _talkCooldown)
             return;
-    
+
         _talkCooldown = true;
         _events.ScheduleEvent(EVENT_SELINA_TALK_COOLDOWN, 30s);
         Talk(SAY_WELCOME, talkTarget);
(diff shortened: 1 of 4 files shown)


src/server/scripts/Events/darkmoon_faire.cpp
```diff
@@ -0,0 +1,190 @@
...
+                CloseGossipMenuFor(player);
+                break;
+            default:
+                break;
+        }
+
+        return false;
+    }
+
+    void SetData(uint32 id, uint32 value) override
```
mdX7, line 126: you can move the entire code of theis function into the method which is being called by the areatrigger (in this case `SetTalkTarget`. You might want to rename it to something like `DoWelcomeTalk` or something when you've moved it there.

src/server/scripts/Events/darkmoon_faire.cpp
```diff
@@ -0,0 +1,171 @@
...
+            default:
+                break;
+        }
+
+        return false;
+    }
+
+    void DoWelcomeTalk(ObjectGuid talkTargetGuid)
+    {
+        if (!talkTargetGuid.IsEmpty())
```
mdX7, line 122: No need to check that. ObjectAccessor takes care of this

src/server/scripts/Events/darkmoon_faire.cpp
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    void UpdateAI(uint32 diff) override
+    {
+        _events.Update(diff);
+
+        switch (_events.ExecuteEvent())
+        {
+            case EVENT_SELINA_TALK_COOLDOWN:
```
mdX7, line 141: This actually looks kinda weird. I'd set a variable _talkCooldown to false when timer runs out and set it to true when doing the talk

src/server/scripts/Events/darkmoon_faire.cpp
meji46: Done!

PR #28067, opened 2022-06-29
Core/Movement: Fixed desync in knockback movement

meji46:
**Changes proposed:**

-  Fixed desynchronization of client when a player performs a knockback movement

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

Test case:
- Player 1 casts the spell 79043 on himself
- Player 2 (observer) sees the knockback move executed by player 1. 
- Before applying this change, during the upward movement of player 1 (the knockback itself), player 2 saw the movement with small jumps and it wasn't fluid

None

### DIFF
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -547,6 +547,7 @@ void WorldSession::HandleMoveKnockBackAck(WorldPackets::Movement::MoveKnockBackA
     if (_player->m_unitMovedByMe->GetGUID() != movementAck.Ack.Status.guid)
         return;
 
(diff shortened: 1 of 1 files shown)


mdX7:
according this issue also happens with npcs and not with players only (we only receive _ACK from players, so this only fixes it for players)

Shauren:
I don't believe that, npcs use spline movement for that

meji46:
Yes, in the description I indicate that it only fixes the problem for players. For npcs I'm investigating since the problem is that the movement is not validated when having speedxy == 0 (I'm getting data from retail to compare) https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Movement/MotionMaster.cpp#L778

PR #28080, opened 2022-07-03
Fix starting level for GMs creating a DK

danjohnso:
**Changes proposed:**

-  Add an additional check to setting player start level to exclude DKs from the GM.StartLevel override of StartPlayerLevel/StartDeathKnightPlayerLevel

**Issues addressed:**
I didn't open an issue with this, but it feels incorrect that setting StartPlayerLevel to 80 would override the StartDeathKnightPlayerLevel being 55 still for GMs.  If the GM.StartLevel is less than the PlayerStartLevel, the previous version of this code would always make the GM.StartLevel config value match the StartPlayerLevel at runtime, which made GM DeathKnights level 80.  If this is intended behavior, feel free to reject this PR.

**Tests performed:**

Tested in game

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -546,7 +546,7 @@ bool Player::Create(ObjectGuid::LowType guidlow, CharacterCreateInfo* createInfo
         ? sWorld->getIntConfig(CONFIG_START_PLAYER_LEVEL)
         : sWorld->getIntConfig(CONFIG_START_DEATH_KNIGHT_PLAYER_LEVEL);
 
-    if (m_session->HasPermission(rbac::RBAC_PERM_USE_START_GM_LEVEL))
+    if (GetClass() != CLASS_DEATH_KNIGHT && m_session->HasPermission(rbac::RBAC_PERM_USE_START_GM_LEVEL))
     {
         uint32 gm_level = sWorld->getIntConfig(CONFIG_START_GM_LEVEL);
         if (gm_level > start_level)
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -546,9 +546,9 @@ bool Player::Create(ObjectGuid::LowType guidlow, CharacterCreateInfo* createInfo
         ? sWorld->getIntConfig(CONFIG_START_PLAYER_LEVEL)
         : sWorld->getIntConfig(CONFIG_START_DEATH_KNIGHT_PLAYER_LEVEL);
 
-    if (GetClass() != CLASS_DEATH_KNIGHT && m_session->HasPermission(rbac::RBAC_PERM_USE_START_GM_LEVEL))
+    if (m_session->HasPermission(rbac::RBAC_PERM_USE_START_GM_LEVEL))
     {
-        uint32 gm_level = sWorld->getIntConfig(CONFIG_START_GM_LEVEL);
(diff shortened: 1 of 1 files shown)


jackpoz:
I need some more data to understand how it worked before and how it is working now:

| StartPlayerLevel | StartDeathKnightPlayerLevel | GM.StartLevel | GM DK Level without PR | GM DK Level with PR |
| --- | --- | --- | --- | --- |
| 1 | 55 | 1 | ??? | ??? |
| 1 | 55 | 10 | ??? | ??? |
| 1 | 55 | 60 | ??? | ??? |
| 1 | 60 | 50 | ??? | ??? |
| 80 | 55 | 1 | ??? | ??? |
| 80 | 55 | 70 | ??? | ??? |

danjohnso:
So part of this is affected by world startup: https://github.com/TrinityCore/TrinityCore/blob/9a9edb4dfe60f574b33c705c79fca29e278e2384/src/server/game/World/World.cpp#L1011-L1017

Anytime GM.StartLevel is less than StartPlayerLevel, it gets set to the StartPlayerLevel.  So notice the last 2 lines of your table I updated what the runtime value of GM.StartLevel is.

<html><body>


StartPlayerLevel | StartDeathKnightPlayerLevel | GM.StartLevel | GM DK Level without PR | GM DK Level with PR
-- | -- | -- | -- | --
1 | 55 | 1 | 55 | 55
1 | 55 | 10 | 55 | 55
1 | 55 | 60 | 60 | 55
1 | 60 | 50 | 60 | 60
80 | 55 | 80 | 80 | 55
80 | 55 | 80 | 80 | 55


</body>
</html>

jackpoz:
Thanks for taking the time to explain this.
To me, this is not the behavior I would expect:
StartPlayerLevel | StartDeathKnightPlayerLevel | GM.StartLevel | GM DK Level without PR | GM DK Level with PR
-- | -- | -- | -- | --
1 | 55 | 80 | 80 | 55

With GM.StartLevel set to 80 I would expect any class to be level 80, even DK. On the other hand, in this other case it makes sense:
StartPlayerLevel | StartDeathKnightPlayerLevel | GM.StartLevel | GM DK Level without PR | GM DK Level with PR
-- | -- | -- | -- | --
1 | 55 | 10 | 10 | 55

So to me GM DK level is std::max(GM.StartLevel, StartDeathKnightPlayerLevel). A DK should never be below StartDeathKnightPlayerLevel , GM or not, but at the same time no GM should be lower than GM.StartLevel . Does this make any sense to you ?

danjohnso:
Yes it makes sense, scenario 1 there is where I first ran into it and thought something needs to change.

My only thought to ignoring GM Level, for Death Knights in particular, is because of how they get their talent points from questing the start zone and not from actually being a certain level.  If I make a Death Knight at level 80 as a GM or as a player, it will only have 25 talent points until I go do all the quests which feels like it defeats the point of starting them at higher than level 55.  After giving this more thought, I am not sure why DK start level is configurable at all without a custom module like the DK Skip Module to handle adding the quest completion talent points.

jackpoz:
As a gm I often `.learn all my class` and learn every talent, even if that is not possible as a normal player. I wonder how often that 55 dk level is changed, but in the end we just give tools to our users, it's up to them what to do with those.

Do you have the chance to change the pr to implement the std::max() code ?

danjohnso:
Fair enough, I pushed the std:max version

jackpoz:
Thanks for the PR 🚀 😃

jackpoz:
This PR caused https://github.com/TrinityCore/TrinityCore/issues/28165 . The level of non-DK chars is considering DK level config, but it shouldn't.

Issue #19210, opened 2017-02-27
Core/Spell: Knock Back effect not working on npcs


Killyana:
[//]: # (***************************************************************)
[//]: # (** DON'T DELETE THIS TEMPLATE ELSE YOUR ISSUE WILL BE CLOSED **)
[//]: # (***************************************************************)

**Description:** Some spells with a Knock Back effect doesn't work on npcs

**Current behaviour:** Knock Back effect doesn't work on npcs

**Expected behaviour:** Knock Back effect doesn't work on npcs

**Steps to reproduce the problem:**

1. .cast self 41909 on any npcs
2. The npc must be Knocked Back in the air then dies after 3 sec (spell liked spell)
3. Notice the Knocked Back effect doesn't happen but the spell linked spell is triggered correctly
4. .cast 41909 in your character to see that the spell work correctly on players

**Branch(es):** 3.3.5

**TC rev. hash/commit:**  rev. 691d52daa9be

[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing C++, create a Pull Request.)

ariel-:
Took a look at this, is not a spell issue, rather a movement issue. Splines are currently not launched for vertical movements.

Vortalex:
The fix commited by ariel does not seem to work, as this is still an issue over a full year later.

Faq:
anyone got sniff of this spell?

Killyana:
``` 
ServerToClient: SMSG_SPELL_START (0x6415) Length: 37 ConnIdx: 2 Time: 06/18/2012 11:49:30.630 Number: 129328
Caster GUID: Full: 0xF13070500071D619 Type: Creature Entry: 28752 Low: 7460377
Caster Unit GUID: Full: 0xF13070500071D619 Type: Creature Entry: 28752 Low: 7460377
Cast Count: 0
Spell ID: 41909 (41909)
Cast Flags: 2 (HasTrajectory)
Time: 0
Time2: 0
Target Flags: 0 (Self)

ServerToClient: SMSG_SPELL_GO (0x6E16) Length: 48 ConnIdx: 2 Time: 06/18/2012 11:49:30.630 Number: 129329
Caster GUID: Full: 0xF13070500071D619 Type: Creature Entry: 28752 Low: 7460377
[...]

PR #28081, opened 2022-07-03
Core/Movement: Knockback improvements

meji46:
**Changes proposed:**

-  Knockbacks with negative speeds will now be validated and processed. 
   - Example: the knockback effect from spell 283677 was never executed for players (both speed are negative). Spells like 79043 were validated because at least one of their speeds was positive (players only).
-  Creatures will be able to execute knockbacks with negative speedXY (I don't know if any additional checks should be made to avoid negative speeds depending on the type of movement). This doesn't fix knockbacks with speedXY == 0.
   - Example: the knockback effect from spell 79043.
- Creatures that cast knockback spells on themselves will use their orientation to determine the direction of movement.

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


src/server/game/Movement/Spline/MoveSplineInit.h
```diff
@@ -149,7 +149,7 @@ namespace Movement
         /* Sets the velocity (in case you want to have custom movement velocity)
          * if no set, speed will be selected based on unit's speeds and current movement mode
          * Has no effect if falling mode enabled
-         * velocity shouldn't be negative
+         * velocity shouldn't be negative (except knockbacks)
```
Shauren, line 152: Is there any other place in spline code left that does not force abs(velocity)?

meji46: I've avoided modifying anything that wouldn't affect knockbacks or that I didn't know what it is used for. Checking what you comment:
- https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Movement/Spline/MoveSpline.cpp#L118 (unused?)
- In CommonInitializer there is no check and I've only changed the parameter passed in its call https://github.com/TrinityCore/TrinityCore/blob/1fc3a51db2d7248c62e0890ae1e0b40f83d9ae2c/src/server/game/Movement/Spline/MoveSpline.cpp#L168
- In MoveSpline I don't know if you prefer to save the velocity in absolute value https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Movement/Spline/MoveSpline.cpp#L181
- MotionMaster::MoveJump and MotionMaster::MoveJumpWithGravity are checking that speedXY is always positive (which I understand is correct)

Faq:
Just a side note, there exists also [19210](https://github.com/TrinityCore/TrinityCore/issues/19210)

meji46:
> Just a side note, there exists also [19210](https://github.com/TrinityCore/TrinityCore/issues/19210)

It's indicated in the second point (This doesn't fix knockbacks with speedXY == 0). It requires additional work and it's best to work on it separately from this PR

aquadeus:
any news on this? could fix the knockback for aberrus spells

Nyr97:
There was a fix a few weeks ago about knockbacks being merged. Worth looking at.

Aokromes:
src/server/game/Movement/Spline/MoveSpline.cpp

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -806,23 +806,34 @@ void MotionMaster::MoveKnockbackFrom(Position const& origin, float speedXY, floa
...
     Position dest = _owner->GetPosition();
+    float o = dest == origin ? 0.0f : _owner->GetRelativeAngle(origin) + float(M_PI);
+    if (speedXY < 0)
+    {
+        speedXY = -speedXY;
+        o = o - float(M_PI);
+    }
+
+    if (speedZ < 0)
+        speedZ = -speedZ; // doesn't seem to be supported on official servers - packet sent for knockback with positive and negative speed has the same flags and JumpGravity
```
Shauren, line 821: Tested with spell 335079

PR #28085, opened 2022-07-03
Core/TempSummon: Implement Ground/FlyingMountDisplayID usage from creature_summoned_data

Nyr97:
Recently, creature_summoned_data in https://github.com/TrinityCore/TrinityCore/commit/ebf1b6eb5cbfd1dc8f1b0d26c29ae1423784cabb was implemented which contained GroundMountDisplayID and FlyingMountDisplayID for this purpose.

**Changes proposed:**

-  TemporalSummons with an Entry in creature_summoned_data will cast the serverside spell to check whether the player IsMounted(). In such a case, if they're IsFlying(), FlyingMountDisplayID will be chosen, if not, GroundMountDisplayID will be chosen.
- The aura will trigger a periodic to check (every 500 ms,  it can be adjusted, but this is the closest to Blizzard packets) if the player IsMounted() and change its MountDisplayID depending on if the player IsFlying() or not. The summoned creature will dismount if the player gets in combat while IsMounted() but IsFlying() not.
- There won't be a SetDisplayMountID packet being sent if the current MountDisplayID equals either GroundMountDisplayID  or FlyingMountDisplayID. This is an additional check to avoid cases in which a summoned creature uses the same mount for both (e.g.: Rana and Al'dali from Zereth Mortis, added them as examples to the SQL).

**Issues addressed:**

None.

**Tests performed:**

It builds and it was tested in-game. Video of reference: https://www.youtube.com/watch?v=gPBTNGTmhFU&ab_channel=Pedris

None.


src/server/game/Entities/Creature/TemporarySummon.cpp
```diff
@@ -238,6 +238,9 @@ void TempSummon::InitStats(uint32 duration)
 
     if (m_Properties->GetFlags().HasFlag(SummonPropertiesFlags::SummonFromBattlePetJournal))
         RemoveNpcFlag(UNIT_NPC_FLAG_WILD_BATTLE_PET);
+
+    if (sObjectMgr->GetCreatureSummonedData(GetEntry()))
```
Shauren, line 242: remove this, handle with creature_template_addon

Shauren (changes requested):

src/server/game/Entities/Creature/TemporarySummon.cpp
Nyr97: via auraID?

Nyr97: done.

sql/updates/world/master/2022_xx_xx_xx_world.sql
```diff
@@ -17,3 +17,11 @@ DELETE FROM `creature_summoned_data` WHERE `CreatureID` IN (182084, 182088);
...
 (182084, 182085, 103104, 103104),
 (182088, 182089, 103104, 103104);
+
+-- Creature Addon
+DELETE FROM `creature_template_addon` WHERE `entry` IN (90382, 91911, 182084, 182088);
+INSERT INTO `creature_template_addon` (`entry`, `path_id`, `mount`, `bytes1`, `bytes2`, `emote`, `aiAnimKit`, `movementAnimKit`, `meleeAnimKit`, `visibilityDistanceType`, `auras`) VALUES
+(90382, 0, 0, 0, 1, 0, 0, 0, 0, 0, '147066'),
+(91911, 0, 0, 0, 1, 0, 0, 0, 0, 0, '147066'),
+(182084, 0, 0, 0, 1, 0, 0, 0, 0, 0, '147066'),
+(182088, 0, 0, 0, 1, 0, 0, 0, 0, 0, '147066');
```
Aokromes: missing newline.

Nyr97: it was added on [wline at end of SQL](https://github.com/TrinityCore/TrinityCore/pull/28085/commits/d47c3906bd6ba55b77253338d099468147dc66c7)

Issue #28076, opened 2022-07-02
[3.3.5] WSG Flag Capture Bug [40$]


immvp:
### Description

Currently, if you hold the Flag on top of the Capture Point on Warsong Gulch while enemy team is also holding the Flag, the flag will not capture. **This is correct.** 
However, if you're on the Capture Point when the enemy Flag gets Returned, then the Flag won't Capture. You need to get off the Capture Point, and onto it again. 

It seems like this could be fixed if you checked Flag position on Flag Return. I'm not sure tho'.

### Expected behaviour

Flag should be captured if Enemy Flag gets returned, even if you're on the capture point when this occurs.

### Steps to reproduce the problem

1. Pick up Flag on both sides
2. Get on the Capture Point with one of the Flags
3. Return the other flag
4. Now you'll see the Flag doesn't Capture, until you've walked off and onto the Capture Point.

### Branch

3.3.5

### TC rev. hash/commit

 TrinityCore rev. debd2446d6e8 2022-07-02 01:04:18 +0200 (main branch) (Unix, Release, Static)

### Operating system

Ubuntu 20.04

### Custom changes

None

immvp:
Not sure how Bounties work, but I'd pay 20$ for this to be fixed.

Aokromes:
you must put it at https://bountysource.com/

immvp:
Done @Aokromes

immvp:
Increased bounty to 40.

Shauren:
Is it even supposed to work like that?

immvp:
@Shauren That flag is supposed to cap when you're on top of capture point? Yes! I can send video evidence if needed!

acidmanifesto:
> Is it even supposed to work like that?

I submitted wsg sniffs a while ago on the forum submission. Think it will show in the sniffs?

immvp:
Maybe I explained the bug a bit bad. Essentially the flag should cap when you're on capture point, even IF you're on it pre-maturely when the flag returns.

immvp:
Is there any way this thread can get the Bounty label, that'd be great. ^_^

immvp:
I can increase the bounty if it means this will be looked at ASAP. I figued it wouldn't be anything major tho'. ^_^

If you're a developer and consider looking at this but the bounty is too small to make it worth it, please let it be known!

PR #28087, opened 2022-07-04
Fix WSG Flag Capture bug #28076

CoderKane:
**Changes proposed:**

**Issues addressed:**

#28076

**Tests performed:**

Untested. Build not checked.

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
+++ b/src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
@@ -541,7 +541,7 @@ void BattlegroundWS::EventPlayerClickedOnFlag(Player* player, GameObject* target
                 if (player->IsInAreaTriggerRadius(sAreaTriggerStore.LookupEntry(3646))) {
                     EventPlayerCapturedFlag(player);
                 }
-            }            
+            }
         }
         else
         {
@@ -581,8 +581,8 @@ void BattlegroundWS::EventPlayerClickedOnFlag(Player* player, GameObject* target
             {
               if (player->IsInAreaTriggerRadius(sAreaTriggerStore.LookupEntry(3647))) {
                     EventPlayerCapturedFlag(player);
-                }    
-            }         
+                }
+            }
         }
         else
         {
--- a/src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
+++ b/src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
@@ -540,7 +540,7 @@ void BattlegroundWS::EventPlayerClickedOnFlag(Player* player, GameObject* target
             {
                 if (player->IsInAreaTriggerRadius(sAreaTriggerStore.LookupEntry(3646))) {
                     EventPlayerCapturedFlag(player);
-                }     
+                }
             }            
         }
         else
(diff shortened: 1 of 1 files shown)


src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
```diff
@@ -551,6 +559,14 @@ void BattlegroundWS::EventPlayerClickedOnFlag(Player* player, GameObject* target
             PlaySoundToAll(BG_WS_SOUND_FLAG_RETURNED);
             UpdatePlayerScore(player, SCORE_FLAG_RETURNS, 1);
             _bothFlagsKept = false;
+
+	        //check if ALLIANCE flag is in capture zone, if so, capture it
```
Shauren, line 563: replace tab with spaces

src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
```diff
@@ -518,6 +518,14 @@ void BattlegroundWS::EventPlayerClickedOnFlag(Player* player, GameObject* target
             PlaySoundToAll(BG_WS_SOUND_FLAG_RETURNED);
             UpdatePlayerScore(player, SCORE_FLAG_RETURNS, 1);
             _bothFlagsKept = false;
+
+            //check if HORDE flag is in capture zone, if so, capture it
+            if (Player* player = ObjectAccessor::FindPlayer(GetFlagPickerGUID(TEAM_HORDE)))
```
Shauren, line 523: Because this function runs in map update context, using `ObjectAccessor::FindPlayer` is unsafe, replace it with `ObjectAccessor::GetPlayer(GetBgMap(), GetFlagPickerGUID(TEAM_HORDE))` (and do the same for alliance code)

Shauren (changes requested):

Shauren:
Code is mostly correct but what I would like to find out is that this is blizzlike behavior

immvp:
Please add my discord Mvq#0670 @CoderKane ^_^ I'd love to have you look at more bugs, at the benefit of TrinityCore of course!

immvp:
> Code is mostly correct but what I would like to find out is that this is blizzlike behavior

It has been confirmed in TBC to function like this, and can be confirmed to function like this in WoD as well. Proof:
https://youtu.be/atPfEHpW0jM?t=437 (7 min 17 seconds)

I doubt they changed the rules for flag capture in WOTLK, Cata and MoP and then randomly went back to how it was in TBC in WoD. I'll keep digging for a WOTLK example tho', if you want. ^_^

Shauren:
A WotLK example would be better because areatriggers (for example the spot where you capture the flag) were messed with in later expansions (but I dont know exactly when)

immvp:
> A WotLK example would be better because areatriggers (for example the spot where you capture the flag) were messed with in later expansions (but I dont know exactly when)
> 
> But that doesn't change the fact that there are changes that must be made before this code can be accepted

I'll find one for you. ^_^

CoderKane:
> Please add my discord Mvq#0670 @CoderKane ^_^ I'd love to have you look at more bugs, at the benefit of TrinityCore of course!

Sure. I'll ad you. I'll can have a look at some of those other bugs when I have some extra time. I'm by no means an expert of C++ but I catch on pretty quick :)

src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
Nyr97: Also, correct spelling: // Check Horde flag if it is in capture zone; if so, capture it

CoderKane: > Also, correct spelling: // Check Horde flag if it is in capture zone; if so, capture it

Fixed :)

Shauren:
> > A WotLK example would be better because areatriggers (for example the spot where you capture the flag) were messed with in later expansions (but I dont know exactly when)
> > But that doesn't change the fact that there are changes that must be made before this code can be accepted
> 
> I'll find one for you. ^_^

Any news on this? I found a questionable one from classic, was wondering if you had anything better

PR #28090, opened 2022-07-04
Scripts/Spells: Fix mage spell Ray of Frost

aquadeus:
**Changes proposed:**

-  Fix mage talent "Ray of Frost"

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -1110,6 +1112,52 @@ class spell_mage_prismatic_barrier : public AuraScript
...
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo ({ SPELL_MAGE_RAY_OF_FROST_BONUS });
+    }
+
+    void HandleEffectPeriodic(AuraEffect const* aurEff)
+    {
+        if (Unit* caster = GetCaster())
+        {
+            if (aurEff->GetTickNumber() > 1) // First tick should deals base damage
```
Nyr97, line 1150: typo

Nyr97 (changes requested):

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -1110,6 +1112,52 @@ class spell_mage_prismatic_barrier : public AuraScript
...
+class spell_mage_ray_of_frost : public SpellScript
+{
+    PrepareSpellScript(spell_mage_ray_of_frost);
+
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo ({ SPELL_MAGE_RAY_OF_FROST_FINGERS_OF_FROST });
+    }
+
+    void HandleFingersOfFrost()
```
mdX7, line 1125: would recommend naming it similar to the hook, like `HandleOnHit`

src/server/scripts/Spells/spell_mage.cpp
aquadeus: where?

Faq: or just "First tick deals base damage"

Shauren:
You need to remove SPELL_MAGE_RAY_OF_FROST_FINGERS_OF_FROST when channel gets interrupted
Don't use hotfix db to edit spells, put that in SpellMgr

aquadeus:
> You need to remove SPELL_MAGE_RAY_OF_FROST_FINGERS_OF_FROST when channel gets interrupted Don't use hotfix db to edit spells, put that in SpellMgr

Ok changed

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -1110,6 +1112,59 @@ class spell_mage_prismatic_barrier : public AuraScript
...
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo ({ SPELL_MAGE_RAY_OF_FROST_BONUS, SPELL_MAGE_RAY_OF_FROST_FINGERS_OF_FROST });
+    }
+
+    void HandleEffectPeriodic(AuraEffect const* aurEff)
+    {
+        if (Unit* caster = GetCaster())
+        {
+            if (aurEff->GetTickNumber() > 1) // First tick should deal base damage
```
Seyden, line 1150: aren't you like 1 tick too late? i mean when the 2nd tick happenend, only then you apply the bonus

mdX7: the tick number is increased before the hook is called, so its 1 for the first tick - not 0.
Ref:
https://github.com/TrinityCore/TrinityCore/blob/2a5329847636ddc080c461bda06375edfbf3815d/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L1027-L1036

Seyden: That doesn’t answer my question.. ticksDone starts at 0 and is increased to 1 and then the hook is called and GetTickNumber returns 1 and not 2.

Therefore you still need a second call in order to get this bonus or am I seeing things ?

mdX7: This hook is executes before dmg is done. So the check is fine.

Ref
https://github.com/TrinityCore/TrinityCore/blob/2a5329847636ddc080c461bda06375edfbf3815d/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L1063

PR #28093, opened 2022-07-04
Spells/Shaman: Fix loop for proc on "Earthen Rage"

aquadeus:
**Changes proposed:**

-  Since the attributes 3 were rewritten the proc of this spell enters in an endless loop of proc due to the correct implementation of the called attribute SPELL_ATTR3_DISABLE_PROC

**Tests performed:**

Tested in-game

### DIFF
--- a/src/server/scripts/Spells/spell_shaman.cpp
+++ b/src/server/scripts/Spells/spell_shaman.cpp
@@ -419,7 +419,7 @@ class spell_sha_earthen_rage_passive : public AuraScript
 
     bool CheckProc(ProcEventInfo& procInfo)
     {
-        return !(procInfo.GetSpellInfo()->Id == SPELL_SHAMAN_EARTHEN_RAGE_DAMAGE);
+        return procInfo.GetSpellInfo() && procInfo.GetSpellInfo()->Id != SPELL_SHAMAN_EARTHEN_RAGE_DAMAGE;
     }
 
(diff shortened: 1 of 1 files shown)


Ovahlord:
have you checked the spellfamilymask before? and also if the proc entry does not override a cooldown in the db2?

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -417,6 +417,11 @@ class spell_sha_earthen_rage_passive : public AuraScript
         return ValidateSpellInfo({ SPELL_SHAMAN_EARTHEN_RAGE_PERIODIC, SPELL_SHAMAN_EARTHEN_RAGE_DAMAGE });
     }
 
+    bool CheckProc(ProcEventInfo& procInfo)
+    {
+        return !(procInfo.GetSpellInfo()->Id == SPELL_SHAMAN_EARTHEN_RAGE_DAMAGE);
```
Shauren, line 422:
```suggestion
        return procInfo.GetSpellInfo() && procInfo.GetSpellInfo()->Id != SPELL_SHAMAN_EARTHEN_RAGE_DAMAGE;
```

Shauren:
> have you checked the spellfamilymask before? and also if the proc entry does not override a cooldown in the db2?

This proc has no cooldown and no familymask (it supposed to proc from all casts)

Issue #28091, opened 2022-07-04
DB/Event: Events calendar update


osleyder85:
### Description

Update of the calendar of events, in addition other events that were missing were added.

### SQL fix

```sql
ALTER TABLE `game_event` CHANGE `eventEntry` `eventEntry` smallint unsigned NOT NULL COMMENT 'Entry of the game event';

UPDATE `game_event` SET `start_time`='2022-07-04 10:00:00', `end_time`='2037-07-05 04:00:00', `length`=1080  						WHERE `eventEntry`=72;  --  62 - Fireworks Spectacular
UPDATE `game_event` SET `start_time`='2022-12-16 10:00:00', `end_time`='2037-01-02 06:00:00', `description`='Feast of Winter Veil' 	WHERE `eventEntry`=2; 	-- 141 - Feast of Winter Veil
UPDATE `game_event` SET `start_time`='2023-04-10 10:00:00', `end_time`='2037-04-17 10:00:00', `occurence`=525600					WHERE `eventEntry`=9; 	-- 181 - Noblegarden
UPDATE `game_event` SET `start_time`='2023-05-01 10:00:00', `end_time`='2037-05-08 10:00:00', `description`='Children\'s Week' 		WHERE `eventEntry`=10;	-- 201 - Children's Week
UPDATE `game_event` SET `start_time`='2022-09-06 10:00:00', `end_time`='2037-09-13 10:00:00'										WHERE `eventEntry`=11; 	-- 321 - Harvest Festival
UPDATE `game_event` SET `start_time`='2022-10-18 10:00:00', `end_time`='2037-11-01 11:00:00', `length`=20220 						WHERE `eventEntry`=12; 	-- 324 - Hallow's End
UPDATE `game_event` SET `start_time`='2023-01-20 10:00:00', `end_time`='2037-02-03 10:00:00'									   	WHERE `eventEntry`=7; 	-- 327 - Lunar Festival
UPDATE `game_event` SET `start_time`='2022-06-21 10:00:00', `end_time`='2037-07-05 10:00:00'                 					   	WHERE `eventEntry`=1;	-- 341 - Midsummer Fire Festival
UPDATE `game_event` SET `start_time`='2022-09-20 10:00:00', `end_time`='2037-10-06 10:00:00', `length`=23040 						WHERE `eventEntry`=24;  -- 372 - Brewfest
[...]

PR #28095, opened 2022-07-06
Core/Event: Add event Weekly Bonus

osleyder85:
**Changes proposed:**
Add bonus for the following events, according to the [wowhead guide](https://www.wowhead.com/bonus-events):

 561 - Arena Skirmish Bonus Event     - 186401 - Sign of the Skirmisher
 563 - Battleground Bonus Event        - 186403 - Sign of Battle
 565 - Pet Battle Bonus Event              - 186406 - Sign of the Critter
 592 - World Quest Bonus Event         - 225788 - Sign of the Emissary
1217 - Shadowlands Dungeon Event  - 225787 - Sign of the Warrior

This PR I did, according to the ID of the events added in this issues [#28091](https://github.com/TrinityCore/TrinityCore/issues/28091)

**Tests performed:**
- [X] Does it build
- [X] Tested in-game

- [X] Add weekly_bonus.cpp in Event

### DIFF
--- a/src/server/scripts/Events/weekly_bonus.cpp
+++ b/src/server/scripts/Events/weekly_bonus.cpp
@@ -20,20 +20,38 @@
 
 constexpr auto MAX_WEEKLY_EVENT = 5;
 
+enum Spells
+{
+    SPELL_SIGN_OF_THE_SKIRMISHER    = 186401,
+    SPELL_SIGN_OF_BATTLE            = 186403,
+    SPELL_SIGN_OF_THE_CRITTER       = 186406,
+    SPELL_SIGN_OF_THE_WARRIOR       = 225787,
+    SPELL_SIGN_OF_THE_EMISSARY      = 225787
+};
+
+enum Events
+{
+    ARENA_SKIRMISH_BONUS_EVENT      = 91,
+    BATTLEGROUND_BONUS_EVENT        = 99,
+    PET_BATTLE_BONUS_EVENT          = 86,
+    SHADOWLANDS_DUNGEON_EVENT       = 102,
+    WORLD_QUEST_BONUS_EVENT         = 94
+};
+
 int32 events[MAX_WEEKLY_EVENT] = {
-    91,  //  561 - Arena Skirmish Bonus Event
-    99,  //  563 - Battleground Bonus Event
-    86,  //  565 - Pet Battle Bonus Event
-    102, // 1217 - Shadowlands Dungeon Event
-    94   //  592 - World Quest Bonus Event
+    ARENA_SKIRMISH_BONUS_EVENT,
+    BATTLEGROUND_BONUS_EVENT,
+    PET_BATTLE_BONUS_EVENT,
+    SHADOWLANDS_DUNGEON_EVENT,
+    WORLD_QUEST_BONUS_EVENT
 };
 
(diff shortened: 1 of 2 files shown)


src/server/scripts/Events/weekly_bonus.cpp
```diff
@@ -0,0 +1,63 @@
...
+    {
+        for (uint8 i = 0; i < MAX_WEEKLY_EVENT; ++i)
+            if (!sGameEventMgr->IsActiveEvent(events[i]))
+                if (player->HasAura(auras[i]))
+                    player->RemoveAura(auras[i]);
+
+        for (uint8 i = 0; i < MAX_WEEKLY_EVENT; ++i)        
+            if (sGameEventMgr->IsActiveEvent(events[i]))
+            {
+                player->AddAura(auras[i], player);
```
Ovahlord, line 54: this is very likely a hack.  use CastSpell(nullptr, auras[i]); instead

Nyr97: Adding aura directly is not recommended because it skips packets that Blizzard normally uses for this type of handling. That's why he's calling it a hack. Just because we haven't gotten rid of all those hacks doesn't mean it's appropriate to add more.

Fabi: > Adding aura directly is not recommended because it skips packets that Blizzard normally uses for this type of handling. That's why he's calling it a hack. Just because we haven't gotten rid of all those hacks doesn't mean it's appropriate to add more.

Get rid of them then :D

src/server/scripts/Events/weekly_bonus.cpp
```diff
@@ -0,0 +1,81 @@
...
+    Weekly_Bonus_Event_PlayerScript() :PlayerScript("Weekly_Bonus_Event_PlayerScript") {}
+
+    void OnLogin(Player* player, bool /*firstLogin*/)
+    {
+        for (uint8 i = 0; i < MAX_WEEKLY_EVENT; ++i)
+            if (!sGameEventMgr->IsActiveEvent(events[i]))
+                if (player->HasAura(auras[i]))
+                    player->RemoveAura(auras[i]);
+
+        for (uint8 i = 0; i < MAX_WEEKLY_EVENT; ++i)
```
Fabi, line 69: I'm pretty sure you can combine both loops into one.

src/server/scripts/Events/weekly_bonus.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+    void OnLogin(Player* player, bool /*firstLogin*/)
+    {
+        for (uint8 i = 0; i < MAX_WEEKLY_EVENT; ++i)
+        {
+            if (!sGameEventMgr->IsActiveEvent(events[i]))
+                if (player->HasAura(auras[i]))
+                    player->RemoveAura(auras[i]);
+
+            if (sGameEventMgr->IsActiveEvent(events[i]))
```
Fabi, line 70: Maybe remove one IsActiveEvent check then? it's already being checked so you canse just do an else there. Which one you do first is up to you :D

Shauren:
This is needs to be implemented in a generic way using SPELL_ATTR13_GLOBAL_AURA (combined with conditions maybe?)

Issue #28088, opened 2022-07-04
[3.3.5] Warsong Gulch Flag Debuff [40$]


immvp:
### Description

Focused Assault (https://wotlkdb.com/?spell=46392) and Brutal Assault ((https://wotlkdb.com/?spell=46393) is currently triggered by both Flags being held for 10 minutes, and 15 minutes, respectively. This is **correct**. 

However, the Timer for the 10 and 15 minutes gets reset if **_bothFlagsKept = false**. This is not correct. It should reset on Flag Capture, or if both Alliance and Horde Flags have been reset to their original state.

### Expected behaviour

In Warsong Gulch, after 10 minutes of both teams being in **posession** of the flag (Or more accurately, 10 minutes of both teams NOT having the Flag Gameobject in its original position), whoever holds the Flag will get a Debuff named Focused Assault (https://wotlkdb.com/?spell=46392).

After 5 more minutes, so 15 minutes total, whoever holds the Flag will get a Debuff named Brutal Assault (https://wotlkdb.com/?spell=46393).

Timer should only reset on Flag Capture or if both Flags have been returned. (Both of which result in the Flag Gameobject being on its original position in the base)

### Steps to reproduce the problem

1. Pick up Flag on both Teams (Horde and Alliance)
2. Wait 10 minutes for the Focused Assault debuff to appear (Or Custom edit the Timer so it's shorter, for easier testing)
3. **Drop** the flag on One team, and **Return** it with the other, while still holding the Flag on the other.
4. Focused Assault debuff will now dissapear from whoever is still holding the Flag.
5. Pick the Flag up again, and wait 10 minutes. You'll see Focused Assault debuff appears again, which means the Timer has been reset from the Return.

### Branch

3.3.5

### TC rev. hash/commit

TrinityCore rev. 0d27a7bdef35 2022-07-04 02:36:22 +0200 (main branch) (Unix, Release, Static)

### Operating system

Ubuntu 20.04

### Custom changes

Transmog Patch. 

Anything related to Battleground and Battlefield has not been touched.

adeutscher:
@immvp I'd like to confirm some phrasing since these seem in conflict (or maybe I'm just misreading):

"In Warsong Gulch, after 10 minutes of both teams being in posession of the flag (Or more accurately, 10 minutes of both teams NOT having the Flag Gameobject in its original position)"

[...]

PR #28098, opened 2022-07-07
Prevent PvP trinket from clearing statuses that ignore caster auras

adeutscher:
**Changes proposed:**

-  When removing auras due to immunity, do not remove auras with the attribute `SPELL_ATTR6_IGNORE_CASTER_AURAS`.

**Issues addressed:**

Brought up in #28088 , though it wasn't the original focus.

**Tests performed:**

```
// Get and equip PVP trinket (tested with a Horde Warlock):
.additem 18852
// Mostly for laughs, give ourselves a speed aura to miss when we apply an Assault debuff to limit our speed.
.aura 1557
// Apply Focused Assault
.aura 46392
// Apply hamstring debuff
.aura 1715
```

While Hamstring is active, use the trinket. It is expected to have Hamstring go away, while Focused Assault remains.

Also tested with Brutal Assault (46393)

### DIFF
--- a/src/server/game/Spells/SpellInfo.cpp
+++ b/src/server/game/Spells/SpellInfo.cpp
@@ -2889,7 +2889,11 @@ void SpellInfo::ApplyAllSpellImmunitiesTo(Unit* target, uint8 effIndex, bool app
     {
         target->ApplySpellImmune(Id, IMMUNITY_STATE, auraType, apply);
         if (apply && HasAttribute(SPELL_ATTR1_DISPEL_AURAS_ON_IMMUNITY))
-            target->RemoveAurasByType(auraType);
+            target->RemoveAurasByType(auraType, [](AuraApplication const* aurApp) -> bool
+            {
+                // if the aura has SPELL_ATTR6_IGNORE_CASTER_AURAS, then it cannot be removed by immunity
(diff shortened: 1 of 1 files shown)


Shauren:
Unfortunately your work is based on incorrect naming of this spell attribute, making this fix invalid.

SpellAttr6 0x4 is `SPELL_ATTR6_NOT_AN_ATTACK` (leaked official name)

Shauren:
No, it means something else and is not related to immunities (we dont control what they mean, it is dbc data)

Shauren:
I disagree - some examples
* 1130 Hunter's Mark (Rank 1)
* 2096 Mind Vision (Rank 1)
and a bunch of non-aura spells

immvp:
Good point. On Wowhead, focused assault has an aura named Unaffected by invulnerability. That's probably what needs to be looked at.

adeutscher:
Dang, that is unfortunate.

In the worst case, if I can't find a better SpellInfo property condition to
filter off of, would it be acceptable to have a filter specifically for the
IDs of the WSG debuffs that got me investigating this?

Shauren:
What if you add SPELL_ATTR0_UNAFFECTED_BY_INVULNERABILITY using spellmgr hack? This spell doesn't have it in 3.3.5 by default

PR #28106, opened 2022-07-10
Improve snare immunity handling

### LINKED ISSUE #28096: [3.3.5] Hand of Freedom Bug (40$)
### Description

Hand of Freedom doesn't remove Slows, but instead makes the Slow effect not work. 
Additionally, if an effect has a Slow effect AND an additional effect, then the ability bugs out, resulting in you being slowed with Hand of Freedom up. (Example is [Frostbolt ](https://wotlkdb.com/?spell=116) with [Permafrost ](https://wotlkdb.com/?spell=12571) Talent)

### Expected behaviour

Hand of Freedom should remove Roots and Slows off the target.
If a spell has a Slow effect AND an additional effect (Like [Frostbolt ](https://wotlkdb.com/?spell=116) with [Permafrost ](https://wotlkdb.com/?spell=12571) then Hand of Freedom should keep the Frostbolt aura active, while removing the Slow effect from it.

1. Log on a Mage and spec into [Permafrost ](https://wotlkdb.com/?spell=12571)
2. Cast an abiltiy with a Chill effect like [Frostbolt](https://wotlkdb.com/?spell=116) or [Cone of Cold](https://wotlkdb.com/?spell=120) on a Paladin
3. Cast Hand of Freedom on the Paladin
4. Try to walk around - You'll notice that you're still slowed.

### Branch

3.3.5

### TC rev. hash/commit

TrinityCore rev. 5e56b14d08d1 2022-07-06 17:15:57 +0200 (main branch) (Unix, Release, Static)

### Operating system

ubuntu 20.04

### Custom changes

Transmog

adeutscher:
**Changes proposed:**

- When snare immunity is applied, remove auras whose only effect is a snare (e.g. Hamstring). Auras with multiple effects (e.g. Infected Wounds) will continue to have their snare effects floored to 0.
- If a new aura with a snare effect is applied to a target that is immune (e.g. Infected Wounds on a target with Hand of Freedom), the user should not have the snare effect of the aura applied to them.
- When snare immunity is removed from a unit that has an aura with a snare component (e.g. Hand of Freedom fading from a unit with Infected Wounds), then the value of the snare aura should be recalculated to snare the unit.

**Issues addressed:**

Closes #28096 

**Tests performed:**

Test definitions:

1. Test restoration of snare effects when immunity is removed
    1. Select a spell with a multi-effect aura. Suggested: Infected Wounds
    2. Cast the selected spell on target
    3. Observe speed, should be slower
    4. Cast Hand of Freedom on target.
    5. Observe debuffs, Multi-effect aura aura should still be on the target
    6. Observe speed, should be normal
    7. Before the snare aura ends, cancel Hand of Freedom
    8. Observe speed, should be slower again
2. Test removal of single-effect auras
    1. Select a spell that applies an aura whose only effect is a snare. Suggested: Blast Wave, Hamstring
    2. Cast the selected spell on the target
    3. Cast Hand of Freedom on the target
    4. Observe debuffs, the snare aura should be removed.
3. Test ignoring of single-effect auras with a snare effect when Hand of Freedom is already active
    1. Select a spell with a non-aura effect and a snare effect. Suggested: Blast Wave
    2. Cast Hand of Freedom on the target.
    3. While Hand of Freedom is active, cast the selected snare spell on the target.
    4. Expected behavior: Target takes damage from Blast Wave, Blast Wave aura is not applied.
4. Test ignoring of snare effects on multi-effect auras when Hand of Freedom is already active
    1. Select a spell with a multi-effect aura. Suggested: Frostbolt with Permafrost talent
    2. Cast Hand of Freedom on the target.
    3. While Hand of Freedom is active, cast the selected snare spell on the target.
    4. Expected behavior: Target takes damage from Frostbolt, Frostbolt aura is applied
    5. Observe speed, should not be reduced
    6. Before Frostbolt ends, cancel Hand of Freedom
    7. Observe speed, should be slower

Test results:

1. Success
2. Success
3. Success (this ended up working even without any changes)
4. Success

- [ ] Fix up removal/ignoring of auras with placeholder effects like Frostbolt. See below comment for an aura test list.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -4131,14 +4131,15 @@ void Unit::RemoveMovementImpairingAuras(bool withRoot)
     }
 }
 
-void Unit::RemoveAurasWithMechanic(uint32 mechanic_mask, AuraRemoveMode removemode, uint32 except)
+void Unit::RemoveAurasWithMechanic(uint32 mechanic_mask, AuraRemoveMode removemode, uint32 except, std::function<bool(AuraApplication const*, uint32)> const& check)
 {
     for (AuraApplicationMap::iterator iter = m_appliedAuras.begin(); iter != m_appliedAuras.end();)
     {
         Aura const* aura = iter->second->GetBase();
         if (!except || aura->GetId() != except)
         {
-            if (aura->GetSpellInfo()->GetAllEffectsMechanicMask() & mechanic_mask)
+            if ((aura->GetSpellInfo()->Mechanic == mechanic_mask || aura->GetSpellInfo()->GetAllEffectsMechanicMask() & mechanic_mask) &&
+                (!check || check(iter->second, mechanic_mask)))
             {
                 RemoveAura(iter, removemode);
                 continue;
@@ -4381,7 +4382,7 @@ void Unit::RestoreMovementImpairingAuras()
                 continue;
 
             AuraEffect* auraEffect = aura->GetEffect(spellEffectInfo.EffectIndex);
-            if (auraEffect->GetAmount() == 0 && auraEffect->GetCaster())
+            if (auraEffect && auraEffect->GetAmount() == 0 && auraEffect->GetCaster())
                 auraEffect->ChangeAmount(auraEffect->CalculateAmount(auraEffect->GetCaster()), false);
         }
 
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -4104,16 +4104,28 @@ void Unit::RemoveMovementImpairingAuras(bool withRoot)
     for (AuraApplicationMap::iterator iter = m_appliedAuras.begin(); iter != m_appliedAuras.end();)
     {
         Aura const* aura = iter->second->GetBase();
+
         if (aura->GetSpellInfo()->Mechanic == MECHANIC_SNARE)
         {
             RemoveAura(iter);
             continue;
         }
 
-        // turn off snare auras by setting amount to 0
+        // turn off snare effects by setting amount to 0
         for (SpellEffectInfo const& spellEffectInfo : aura->GetSpellInfo()->GetEffects())
-            if (iter->second->HasEffect(spellEffectInfo.EffectIndex) && spellEffectInfo.Mechanic == MECHANIC_SNARE)
-                aura->GetEffect(spellEffectInfo.EffectIndex)->ChangeAmount(0);
+        {
+            if (!iter->second->HasEffect(spellEffectInfo.EffectIndex))
+            {
+                continue;
+            }
+
+            if (spellEffectInfo.Mechanic != MECHANIC_SNARE)
+            {
+                continue;
+            }
+
+            aura->GetEffect(spellEffectInfo.EffectIndex)->ChangeAmount(0);
+        }
 
         ++iter;
     }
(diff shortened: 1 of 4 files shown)


adeutscher:
Discussed with @immvp and came to the decision that we should proceed to treat auras like non-permafrost Frostbolt to be removed/ignored when their only effective aura effect value is a snare.

To make sure we're not breaking anything, we put together a list of auras to test to make sure that they aren't being wrongfully kept/ignored

To keep (with neutralized snare):

* Frostfire Bolt
* Frost Armor, Infected Wounds
* Frostbolt (with Permafrost)
* Slow (Mage talent)
* Icy Chill Enchant
* Chains of Ice w/ glyph

To go:

* Blast Wave
    * Pure snares like this are already ignored/removed, but this test confirms that nothing is broken.
* Frostbolt (without Permafrost)

adeutscher:
Rephrased ToDo list in PR description

immvp:
Chains of Ice w/ Glyph should not be kept on second thought

Jildor:
We have any 3.3.5 proof about last point?
>When snare immunity is removed from a unit that has an aura with a snare component (e.g. Hand of Freedom fading from a unit with Infected Wounds), then the value of the snare aura should be recalculated to snare the unit.

Shauren:
I don't like this solution, feels too much like special-casing MECHANIC_SNARE

Things I have found:
* counting non-snare effects is unneccessary, `Unit::RemoveMovementImpairingAuras` does it mostly right (and I believe all mechanic immunities should behave like that) - remove auras that have `GetSpellInfo()->Mechanic == immuned_mechanic` and turn off effects that have `spellEffectInfo.Mechanic == immuned_mechanic`
* Expanding on "turning off effects" I identified a possible bug in how `Aura::UpdateTargetMap` works - immunity to individual effects is only checked after this line https://github.com/TrinityCore/TrinityCore/blob/13d86762e65f13c50870430f0e015a9ef387eb1d/src/server/game/Spells/Auras/SpellAuras.cpp#L662 which means that existing effects are not re-evaluated (danger: needs keeping track of immunity source for SPELL_ATTR1_DISPEL_AURAS_ON_IMMUNITY)

Keeping those two points in mind I believe the ideal fix would:
1. correct the behavior of `Unit::RemoveAurasWithMechanic` to make it only remove the aura when spell mechanic field is equal to removed mechanic and call `Aura::UpdateTargetMap` otherwise
2. reorder immunity checks in Aura::UpdateTargetMap/UnitAura::FillTargetMap
3. remove special casing of MECHANIC_SNARE in `SpellInfo::ApplyAllSpellImmunitiesTo`

By doing all of the above, you would also remove the need to manually reapply auras after immunity ends because Aura::UpdateTargetMap is called periodically

sanctum32:
i would suggest to check this code line too https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Unit/Unit.cpp#L8209, its not related to your changes but this bitmask check is incorrect there and npcs speeds are not updated properly - because bitmask index before shift is reduced by one step and it checks not intended immunity

adeutscher:
@Shauren Those changes make sense to me at a high level.

I think I've gone and burned myself out a bit, so I'm having trouble wrapping my head around UpdateTargetMap. I'm going to make a note in the bug that someone who takes this up should follow your suggestions, but I don't think I'll be able to be the one to do it at the moment.

Shauren:
> i would suggest to check this code line too https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Unit/Unit.cpp#L8209 (or https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Entities/Unit/Unit.cpp#L8599), its not related to your changes but this bitmask check is incorrect there and npcs speeds are not updated properly - because bitmask index before shift is reduced by one step and it checks not intended mechanic
> 
> i suggest to change it to `if ((immuneMask & (1 << MECHANIC_SNARE)) != 0 || (immuneMask & (1 << MECHANIC_DAZE)) != 0)`

No, they are correct. mechanic_immune_mask in database is shifted by a single bit so that 1 means immunity to MECHANIC_CHARM, not MECHANIC_NONE

sanctum32:
@Shauren i meant ```(MECHANIC_SNARE - 1)``` parts, by default, rest of core code doesn't do such index deduction before shifting. Anyways I believe its just logical mistake, because it takes wrong index to shift

sanctum32:
i understand what shifting does, for example
1 << 5, this will mean, it will go to 5th bit index, so it would be equal to 32 (or 0x20), but if we do action like 1 << (5 - 1), it will return 16 (or 0x10)

Shauren:
No, you did not understand what I said at all. I will repeat it for the third time:
To make creature immune to MECHANIC_CHARM you set `creature_template.mechanic_immune_mask = (1 << (MECHANIC_CHARM - 1)) = 1` (MECHANIC_CHARM enum value is 1)
Because that is what `Creature::LoadTemplateImmunities` expects - so all other places in code that read MechanicImmuneMask directly must also apply -1 to mechanic value

sanctum32:
hm, i missed about MECHANIC_NONE definition, so yea, you are right but still, most of core doesn't do this bitmask shifting correctly still

PR #28109, opened 2022-07-12
Combat Trigger on Spell Launch

### LINKED ISSUE #28103: [3.3.5] Combat trigger
### Description

Combat currently gets triggered the same way for Caster and Target. 
If Caster casts Fireball, Caster will be put in Combat instantly, and so will the **target**. 

### Expected behaviour

Caster should be put in combat when projectile is in the air. Proof: https://youtu.be/fLjTRifi6dE?t=206 (03:26)
Target should be put in combat when projectile hits. Proof: https://youtu.be/Jug7beeLQ1A?t=367 (06:07, the guy literally zooms in on the thrown, also this is at 19 twinks so warriors can't charge in combat)

1. Have Player A cast a projectile type ability at Player B
2. Observe Combat on Both players. They'll get Combat simultaneously.

### Branch

3.3.5

### TC rev. hash/commit

 TrinityCore rev. 5759246d73c7 2022-07-09 18:15:48 +0200 (main branch) (Unix, Release, Static)

### Operating system

Ubuntu 20.04

### Custom changes

Transmog Patch

adeutscher:
**Changes proposed:**

-  When the user fires a projectile, immediately enter the caster into combat and do not enter the target into combat until the result of the launch is announced.
-  When a spell misses, place the target in combat

**Issues addressed:**

Closes #28103 

**Tests performed:**
Setup:

1. On one account, create a character with a projectile spell. I went with a warlock for Shadow Bolt, but using Fireball would have been better
2. On another account, create a warrior to be a target for projectiles. Specifically making a warrior to test Spell Reflection in a test case.
3. Log into the characters, and put them in a setup that allows them to attack each other
4. Set up your windows so that you can conveniently watch the combat state of the caster and the target at once. 
5. Place the characters at maximum range for your spell.
6. Each of these tests assumes that the two characters are not in combat at step 1.

Test Cases:

1. Confirm that projectiles don't put the target into combat at launch (Hit).
    1. Cast a projectile at the target.
    2. While the projectile is in the air, observe that the caster is immediately placed into combat.
    3. While the projectile is in the air, observe that the target is not placed into combat.
    4. Confirm that the projectile cast has hit. If it misses, then pivot to Test 2. Test 2 is this test, but for the more unlikely case of non-hits
    5. Observe that the target is now in combat.
2. Confirm that projectiles place the caster in combat and don't put the target into combat at launch (Miss).
    1. Cast a projectile at the target.
    2. While the projectile is in the air, observe that the caster is immediately placed into combat.
    3. While the projectile is in the air, observe that the target is not placed into combat.
    4. Confirm that the projectile cast has missed. If it hits, then repeat until it does not
        * If you don't want to keep rolling the dice on a miss, then you could create a second caster with a large level distance
    6. Observe that the target is now in combat.
3. Confirm that reflected projectiles put the target in combat at time of reflection. Expecting this to act just like Test 2, but adding it here to cover our bases.
	1. Set the warrior up with a shield and teach them spell reflection
        * `.additem 32255`
        * `.learn 23920`
        * `.level 70`
    2. Set up enough Rage on the target, and cast Spell Reflection
    3. While Spell Reflection is active, cast a projectile that can be reflected at the target.
    4. While the projectile is in the air, observe that the caster is immediately placed into combat.
    5. While the projectile is in the air, observe that the target is not placed into combat.
    7. Confirm that the projectile cast has is reflected. If it does not, then repeat until it does
    8. Observe that the target is in combat BEFORE the spell projectile hits the caster.
4. Test pointer paranoia. This is probably an understanding hurdle on my part), but if it doesn't crash things then that can only be good.
    1. Fire a projectile at a target. 
    2. Before the projectile hits, sign that target out
5. Smoke Testing. Spend a session running around the world a bunch with a spellcaster. I'd encourage abusing GM powers to visit more areas and dying in an instance at least once (releasing spirit in an instance is what caused the crash that prompted this test).

Test Results:

1. Success
2. Success
3. Success
4. Success* - When this happens, the caster is told the spell is "interrupted" after the projectile is launched. Bit weird, but I'm classifying this behaviour as out-of-scope for this bug/PR. Nothing's crashing on us on account of that test.
5. Smoke tested with misc in-game play.

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2030,37 +2030,6 @@ void Spell::CleanupTargetList()
     m_delayMoment = 0;
 }
 
-class ProcImpactDelayed : public BasicEvent
-{
-public:
-    ProcImpactDelayed(Unit* owner, ObjectGuid casterGUID, const bool enterCombat, bool enablePVP) : _owner(owner), _casterGUID(casterGUID), _enterCombat(enterCombat), _enablePVP(enablePVP) { }
-
-    bool Execute(uint64 /*e_time*/, uint32 /*p_time*/) override
-    {
-        if(!_owner)
-            return true;
-
-        Unit* caster = ObjectAccessor::GetUnit(*_owner, _casterGUID);
-        if (!caster)
-            return true;
-
-        // This will only cause combat - the target will engage once the projectile hits (in DoAllEffectOnTarget)
-        if (_enterCombat)
-            _owner->SetInCombatWith(caster);
-
-        if (_enablePVP && caster->ToPlayer())
-            caster->ToPlayer()->UpdatePvP(true);
-
-        return true;
-    }
-
-private:
-    Unit*  _owner;
-    ObjectGuid _casterGUID;
-    bool _enterCombat;
-    bool _enablePVP;
-};
-
 class ProcReflectDelayed : public BasicEvent
 {
     public:
@@ -2144,12 +2113,7 @@ void Spell::AddUnitTarget(Unit* target, uint32 effectMask, bool checkIfValid /*=
 
     // Calculate hit result
     WorldObject* caster = m_originalCaster ? m_originalCaster : m_caster;
-    Unit* unitCaster = m_caster->ToUnit();
     targetInfo.MissCondition = caster->SpellHitResult(target, m_spellInfo, m_canReflect && !(IsPositive() && m_caster->IsFriendlyTo(target)));
-
-    // This will only cause combat - the target will engage once the projectile hits (in DoAllEffectOnTarget)
-    if (m_originalCaster && targetInfo.MissCondition != SPELL_MISS_EVADE && !m_originalCaster->IsFriendlyTo(target) && (!m_spellInfo->IsPositive() || m_spellInfo->HasEffect(SPELL_EFFECT_DISPEL)) && (m_spellInfo->HasInitialAggro() || target->IsEngaged()))
-        m_originalCaster->SetInCombatWith(target, true);
 
     // Spell have speed - need calculate incoming time
     // Incoming time is zero for self casts. At least I think so.
@@ -2166,15 +2130,6 @@ void Spell::AddUnitTarget(Unit* target, uint32 effectMask, bool checkIfValid /*=
         // Calculate minimum incoming time
         if (!m_delayMoment || m_delayMoment > targetInfo.TimeDelay)
             m_delayMoment = targetInfo.TimeDelay;
-
-        if (unitCaster && targetInfo.MissCondition != SPELL_MISS_EVADE)
-        {
-            bool enterCombat = !caster->IsFriendlyTo(target) && (!m_spellInfo->IsPositive() || m_spellInfo->HasEffect(SPELL_EFFECT_DISPEL)) && (m_spellInfo->HasInitialAggro() || target->IsEngaged());
-            bool enablePVP = targetInfo.IsPVPEnabling();
-
-            if (enterCombat || enablePVP)
-                target->m_Events.AddEvent(new ProcImpactDelayed(target, m_originalCasterGUID, enterCombat, enablePVP), target->m_Events.CalculateTime(Milliseconds(targetInfo.TimeDelay)));
-        }
     }
     else
         targetInfo.TimeDelay = 0ULL;
@@ -2183,7 +2138,7 @@ void Spell::AddUnitTarget(Unit* target, uint32 effectMask, bool checkIfValid /*=
     if (targetInfo.MissCondition == SPELL_MISS_REFLECT)
     {
         // Calculate reflected spell result on caster (shouldn't be able to reflect gameobject spells)
-        ASSERT(unitCaster);
+        Unit* unitCaster = ASSERT_NOTNULL(m_caster->ToUnit());
         targetInfo.ReflectResult = unitCaster->SpellHitResult(unitCaster, m_spellInfo, false); // can't reflect twice
 
         // Proc spell reflect aura when missile hits the original target
@@ -2357,9 +2312,8 @@ void Spell::TargetInfo::PreprocessTarget(Spell* spell)
     else if (MissCondition == SPELL_MISS_REFLECT && ReflectResult == SPELL_MISS_NONE)
         _spellHitTarget = spell->m_caster->ToUnit();
 
-    // Ensure that a player target is put in combat by a taunt, even if they result immune clientside
-    if ((MissCondition == SPELL_MISS_IMMUNE || MissCondition == SPELL_MISS_IMMUNE2) && spell->m_caster->GetTypeId() == TYPEID_PLAYER && unit->GetTypeId() == TYPEID_PLAYER && spell->m_caster->IsValidAttackTarget(unit, spell->GetSpellInfo()))
-        unit->SetInCombatWith(spell->m_caster->ToPlayer());
+    if (spell->m_originalCaster && MissCondition != SPELL_MISS_EVADE && !spell->m_originalCaster->IsFriendlyTo(unit) && (!spell->m_spellInfo->IsPositive() || spell->m_spellInfo->HasEffect(SPELL_EFFECT_DISPEL)) && (spell->m_spellInfo->HasInitialAggro() || unit->IsEngaged()))
+        unit->SetInCombatWith(spell->m_originalCaster);
 
     spell->CallScriptBeforeHitHandlers(MissCondition);
 
@@ -7651,10 +7605,18 @@ void Spell::HandleLaunchPhase()
 
 void Spell::DoEffectOnLaunchTarget(TargetInfo& targetInfo, float multiplier, SpellEffectInfo const& spellEffectInfo)
 {
+    Unit* targetUnit = m_caster->GetGUID() == targetInfo.TargetGUID ? m_caster->ToUnit() : ObjectAccessor::GetUnit(*m_caster, targetInfo.TargetGUID);
+    if (!targetUnit)
+        return;
+
+    // This will only cause combat - the target will engage once the projectile hits (in Spell::TargetInfo::PreprocessTarget)
+    if (m_originalCaster && targetInfo.MissCondition != SPELL_MISS_EVADE && !m_originalCaster->IsFriendlyTo(targetUnit) && (!m_spellInfo->IsPositive() || m_spellInfo->HasEffect(SPELL_EFFECT_DISPEL)) && (m_spellInfo->HasInitialAggro() || targetUnit->IsEngaged()))
+        m_originalCaster->SetInCombatWith(targetUnit, true);
+
     Unit* unit = nullptr;
     // In case spell hit target, do all effect on that target
     if (targetInfo.MissCondition == SPELL_MISS_NONE)
-        unit = m_caster->GetGUID() == targetInfo.TargetGUID ? m_caster->ToUnit() : ObjectAccessor::GetUnit(*m_caster, targetInfo.TargetGUID);
+        unit = targetUnit;
     // In case spell reflect from target, do all effect on caster (if hit)
     else if (targetInfo.MissCondition == SPELL_MISS_REFLECT && targetInfo.ReflectResult == SPELL_MISS_NONE)
         unit = m_caster->ToUnit();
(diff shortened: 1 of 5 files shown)


adeutscher:
Fixed my initialization error, but I do not currently understand the current error that I'm seeing locally with CI parameters that are more like the CI build:

```
In file included from /usr/include/signal.h:328,                                                                                                                                                                                            
                 from /home/adeutscher/dev/wow/wotlk/src/TrinityCore/build/_deps/catch2-src/single_include/catch2/catch.hpp:8039,                                                                                                           
                 from /home/adeutscher/dev/wow/wotlk/src/TrinityCore/tests/main.cpp:20:                                                                                                                                                     
/home/adeutscher/dev/wow/wotlk/src/TrinityCore/build/_deps/catch2-src/single_include/catch2/catch.hpp:10827:58: error: call to non-‘constexpr’ function ‘long int sysconf(int)’                                                             
10827 |     static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;                                                                                                                                         
      |                                                          ^~~~~~~~~~~                                                                                                                                                                
compilation terminated due to -Wfatal-errors.
```

Time will tell if if it's something weird with my environment or something that the CI environment has no issue with.

adeutscher:
The build error from my local environment with CI-matching `cmake` params did not happen with the CI build.

To whomever reviews this, is the error in my previous comment cause for concern or does it smell of "local environment screwiness"?

src/server/game/Spells/Spell.cpp
```diff
@@ -2130,6 +2166,10 @@ void Spell::AddUnitTarget(Unit* target, uint32 effectMask, bool checkIfValid /*=
         // Calculate minimum incoming time
         if (!m_delayMoment || m_delayMoment > targetInfo.TimeDelay)
             m_delayMoment = targetInfo.TimeDelay;
+
+        if (unitCaster && targetInfo.MissCondition != SPELL_MISS_EVADE) {
+            target->m_Events.AddEvent(new ProcImpactDelayed(unitCaster, targetGUID, m_spellInfo, targetInfo.IsPVPEnabling()), target->m_Events.CalculateTime(Milliseconds(targetInfo.TimeDelay)));
```
jackpoz, line 2171: Is it safe to store unitCaster pointer in an event that will be executed later on, with a chance of unitCaster having logged out in the meantime ?

adeutscher:
```c++
class ProcReflectDelayed : public BasicEvent
{
    public:
        ProcReflectDelayed(Unit* owner, ObjectGuid casterGuid) : _victim(owner), _casterGuid(casterGuid) { }

        bool Execute(uint64 /*e_time*/, uint32 /*p_time*/) override
        {
            Unit* caster = ObjectAccessor::GetUnit(*_victim, _casterGuid);
            if (!caster)
                return true;

            uint32 const typeMaskActor = PROC_FLAG_NONE;
            uint32 const typeMaskActionTarget = PROC_FLAG_TAKEN_SPELL_MAGIC_DMG_CLASS_NEG | PROC_FLAG_TAKEN_SPELL_NONE_DMG_CLASS_NEG;
            uint32 const spellTypeMask = PROC_SPELL_TYPE_DAMAGE | PROC_SPELL_TYPE_NO_DMG_HEAL;
            uint32 const spellPhaseMask = PROC_SPELL_PHASE_NONE;
            uint32 const hitMask = PROC_HIT_REFLECT;

            Unit::ProcSkillsAndAuras(caster, _victim, typeMaskActor, typeMaskActionTarget, spellTypeMask, spellPhaseMask, hitMask, nullptr, nullptr, nullptr);
            return true;
        }

    private:
        Unit* _victim;
        ObjectGuid _casterGuid;
};
```

```c++
namespace ObjectAccessor
{
    // these functions return objects only if in map of specified object
    ...
    TC_GAME_API Unit* GetUnit(WorldObject const&, ObjectGuid const& guid);
    ...
```

`ObjectAccessor` does have some methods for accessing anywhere, and we could add a FindUnit in a pinch, but they come with a warning label:

```c++
    // these functions return objects if found in whole world
    // ACCESS LIKE THAT IS NOT THREAD SAFE
    TC_GAME_API Player* FindPlayer(ObjectGuid const&);
    TC_GAME_API Player* FindPlayerByName(std::string_view name);
    TC_GAME_API Player* FindPlayerByLowGUID(ObjectGuid::LowType lowguid);
```

Another option is that there is this guy outside of the "not thread safe" cluster:

```
    TC_GAME_API Player* GetPlayer(Map const*, ObjectGuid const& guid);
```

The existing GetUnit call that's being used is:

```c++
Unit* ObjectAccessor::GetUnit(WorldObject const& u, ObjectGuid const& guid)
{
    if (guid.IsEmpty())
        return nullptr;

    // 
    if (guid.IsPlayer())
        return GetPlayer(u, guid);

    if (guid.IsPet())
        return GetPet(u, guid);

    return GetCreature(u, guid);
}

Player* ObjectAccessor::GetPlayer(WorldObject const& u, ObjectGuid const& guid)
{
    return GetPlayer(u.GetMap(), guid);
}
```

If the approach in both `ProcReflectDelayed` and the new `ProcImpactDelayed` is fundamentally unsafe, how does the following sound?

1. Instead of taking a unit, event creation takes a `Map*` and a GUID up front.
2. The event invokes a newly created `Unit* ObjectAccessor::GetUnit(Map* u, ObjectGuid const& guid)`, with the same setup as the original version that is currently being used.
3. The events in `Spell.cpp` use the new `ObjectAccessor::GetUnit` overload for everything, with a guard clause for failing to get the unit that was via pointer.

jackpoz:
it's not just what the event stores but how the event is used.

Take your code for example:
```cpp
target->m_Events.AddEvent(new ProcImpactDelayed(unitCaster, targetGUID, m_spellInfo, targetInfo.IsPVPEnabling()), target->m_Events.CalculateTime(Milliseconds(targetInfo.TimeDelay)));
```
In this case we are storing `unitCaster` in an event scheduled to `target` . This is unsafe because `unitCaster` could be invalid by the time the event is processed (for example the caster logs out).
What is safe is to store `target` as that's the same holder of the event and if `target` becomes invalid, then all events it's holding will be removed anyway.

In short, this is the safe pattern:
```cpp
somePointer->m_Events.AddEvent(new SomeEvent(somePointer, onlyGUIDHere, anotherGUIDHere));
```

As for `ProcReflectDelayed`, we'll have to check if `Unit* owner` is the same as the event holder, as that's the only safe pattern.

adeutscher:
If a unit becomes invalid, would it still be stored for retrieval in ObjectAccessor?

To confirm with a code example, I'm thinking of the following:

```c++
bool Execute(uint64 /*e_time*/, uint32 /*p_time*/) override
{
	// Use map pointer (taken from Unit->GetMap()) and GUID retrieved when the event was initialized to confirm that the unit is still valid to work with.
	Unit* caster = ObjectAccessor::GetUnit(*_mapPointer, _casterGuid);
	if(!caster)
		// Was unable to fetch caster, abort. logged out?
		return true;

	Unit* target = ObjectAccessor::GetUnit(*caster, _targetGuid);
	if (!target)
		// Was unable to fetch target, abort
		return true;

	// Do the rest of the event's logic
```

jackpoz:
ObjectAccessor can be trusted as long as the result is used only for the current update tick, so every update loop ObjectAccessor has to be called. If the unit is no longer present, it will return `nullptr`.

There's no need to pass Map, the only pointer passed should be the holder of the event. That's why `ProcReflectDelayed` has `Unit* owner` as parameter name, it's the owner of the event, no reference about being caster or target.

adeutscher:
The main reason I'd like to pass the `Map` is that the `ObjectAccessor` methods that don't have a context object are have the warning label about not being threadsafe. With what you said about the update tick I could be over-thinking things, though. The `ObjectAccessor` call currently in `ProcReflectDelayed` is a few lines of code away from using the owner unit's `GetMap()` function, so that drove my thinking.

If you think it's safe, then do you agree that it would acceptable to skip passing through the `Map*` and make a method like `TC_GAME_API Player* FindPlayer(ObjectGuid const&);`, but for finding the `Unit*`? And if so, should this approach be applied both `ProcReflectDelayed` and `ProcImpactDelayed`? Both events are scheduled to be run when the projectile result is revealed to the players.

src/server/game/Spells/Spell.cpp
Seyden: @jackpoz i guess that was a rhetorical question? :D

jackpoz:
ObjectAccessor::GetUnit() is fine, it already does what the FindUnit() one you have in mind to implement. It's thread-safe enough in our context so it's all fine.

About the other event, we should check if the holder of m_Events is the same as the Unit* parameter, otherwise we should fix that (but please let's do that in another PR)

adeutscher:
Roger.

I'm happy to keep the scope of this PR as low as possible.

I was messing about in the my test world to demo the concept to some today,
and accidentally brought down the server. I don't have proof it was my
changes, but wasn't doing anything too weird so it seems like a good
suspect. You made a good call bringing up your concerns.

With the change affecting a core part of combat, I'm going to make "mess
around in the build for a few hours" part of my testing routine on top of
my structured tests that I've written above.

src/server/game/Spells/Spell.cpp
adeutscher: Confirmed. I haven't had the bandwidth to do any TC commits in the last couple days.

adeutscher: Made the suggested changes where target is the first parameters.

I'm going to add one more specific test, just to be put my mind at ease (I think my brain might still stuck in some wrong idea of the checks and statuses): Log a character out while a projectile is in flight towards them.

adeutscher:
@jackpoz In #28103 , a user brought up similar behaviour with hunter traps. It boils down to: In 3.3.5, Explosive/Frost/Freezing trap would place both the trapper and victim in combat. This PR's branch now keeps the victim out of combat, but the trapper is still placed in combat. My understanding is that neither the trapper or the victim should go into combat.

The full fix to this would be based around one of the `SetInCombatWith` calls that I'm adjusting here (or maybe there's a Spell flag for this, I don't know off the cuff and I haven't researched it), but I'd feel more comfortable if this were addressed in another issue/branch/PR. Do you agree?

adeutscher:
@jackpoz I'd say it improves on some cases. Projectiles+PvP combat is acting much better, even if there are still weird behaviours like traps that carried over from 3.3.5.

I've just re-tested my test cases. For the next couple hours I'm going to fart around in a test build. If I don't manage to crash the server, I'll check off my final ToDo.

adeutscher:
@jackpoz Apologies if this is too forward, but I wanted to send an update ping on this.

The merge is contingent on approval of course, but I wanted to confirm that I've wrapped up my testing.

src/server/game/Spells/Spell.cpp
```diff
@@ -2030,6 +2030,37 @@ void Spell::CleanupTargetList()
...
+
+        Unit* caster = ObjectAccessor::GetUnit(*_owner, _casterGUID);
+        if (!caster)
+            return true;
+
+        // This will only cause combat - the target will engage once the projectile hits (in DoAllEffectOnTarget)
+        if (_enterCombat)
+            _owner->SetInCombatWith(caster);
+
+        if (_enablePVP)
```
jackpoz, line 2051:
```suggestion
        if (_enablePVP && caster->ToPlayer())
```

jackpoz:
@Jildor did you have any feedback about this ?

Jildor:
Hi! yes, in case of hunter trap (I know that is special case because are gameobjects) here I try to explain how should work:
https://github.com/TrinityCore/TrinityCore/issues/28103#issuecomment-1186973330
before PR: Trap hit target and owner of trap (player hunter) and target set in combat
after PR: Trap hit target and owner of trap (player hunter) set in combat, but target can't due to that unitcaster check.
How should it work: Trap hit target and target set in combat, owner of trap (player hunter) shouldn't set in combat due to that.

so the result after this PR is it's worse than before for hunter traps
in case of m_caster is a gameobject and m_originalCaster is a player, can ```target->SetInCombatWith(m_originalCaster, true);``` (instead of ```m_originalCaster->SetInCombatWith(target, true);```) so avoid hunter set in combat but probably not is a good idea

also behaviour of leave combat isn't correct because if caster enter in combat when cast spell and target enter in combat after delayedEvent is executed but both leave combat at the same time is wrong, here I explain it better how should it work: https://github.com/TrinityCore/TrinityCore/issues/28103#issuecomment-1191268267

adeutscher:
In my other post on the topic I assumed that it was better with the thinking that the desired effect was for neither party to be in combat. But if the victim IS supposed to be in combat that is a complete backslide.

PR #28112, opened 2022-07-13
Scripts/Spells: Divine Star (Priest) fixed

Nyr97:
**Changes proposed:**

- [x]  Reorganized the headers to follow alphabetical order.
- [x] Divine Star follows terrain and forces return if it finds any collision on its first path. It will ignore terrain and collision on its second path.
- [x] Divine Star will hold units' GUIDs to avoid any case in which it would hit twice a unit on a single path if they happen to get inside the areatrigger twice due to speed or abilities (e.g. Charge). Vector is cleaned upon starting its return path to mimic this behavior.
- [x] Divine Star hits any unit inside the areatrigger when starting its return path OnUnitExit since OnUnitEnter will not be triggered for units that happened to be at the end of the first path (edgy case).

**Issues addressed:**

None.

**Tests performed:**

It was built and tested in-game.

- Atonement from Discipline specialization should only trigger on the first instance of damage on any unit hit by Divine Star (ignoring this one on this PR as this goes beyond Divine Star, it affects any AoE damaging spell for the priest).
- [x] Sometimes Z height on the first path is normalized to groundZ due to VMAPS issues. This seems to happen when trying to reach an elevation on the terrain from a Z beneath it (@jackpoz field of expertise to test and debug.) It doesn't happen the other way around:
- Casting Divine Star from stairs up down: https://gyazo.com/8d14bc9a2de42a7d4551b48017bad5bb
- Casting Divine Star from stairs down up: https://gyazo.com/b10024f4408aa098297fc1e4533e6b6e


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -323,6 +328,78 @@ class spell_pri_divine_hymn : public SpellScript
...
+                    {
+                        _returningToCaster = true;
+                        _casterCurrentPosition = caster->GetPosition();
+
+                        Movement::PointsArray returnSplinePoints;
+
+                        for (uint8 i = 0; i < 4; i++)
+                        {
+                            G3D::Vector3 returnPoint;
+                            returnPoint.x = (i < 2) ? at->GetPositionX() : caster->GetPositionX();
```
Seyden, line 377: Any particular reason why these are 4 return points and the first two being the same and also the last two being the same?

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -323,6 +328,78 @@ class spell_pri_divine_hymn : public SpellScript
...
+
+                        at->InitSplines(returnSplinePoints, (at->GetDistance(caster) / 24) * 1000);
+                    }
+
+                    // NOTE: this can be adjusted, 100-150 was the closest the AT seems to update its spline to the caster's current position.
+                    task.Repeat(Milliseconds(100));
+                });
+            }
+            else
+                at->Remove();
```
Seyden, line 392: you can actually put that to the start of the method and check

if (_returningToCaster)
{
   at->Remove();
   return;
}

then you prevent clearing unneccessarily the _affectedUnits and stuff

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -323,6 +328,78 @@ class spell_pri_divine_hymn : public SpellScript
...
+
+    void OnDestinationReached() override
+    {
+        if (Unit* caster = at->GetCaster())
+        {
+            _affectedUnits.clear();
+
+            if (!_returningToCaster)
+            {
+                _scheduler.Schedule(Milliseconds(), [this, caster](TaskContext task)
```
Seyden, line 365: i would actually outsource this part of the code to a method called: ReturnToCaster

src/server/scripts/Spells/spell_priest.cpp
Nyr97: On retail, it seems to be 4 different points at which the spline redirects onto the caster when returning.

Seyden: So 4 different points, but right now you have 4 points of which half and half are the same

src/server/scripts/Spells/spell_priest.cpp
Nyr97: done

src/server/scripts/Spells/spell_priest.cpp
Nyr97: moved it somewhere else.

src/server/scripts/Spells/spell_priest.cpp
Nyr97: image
on retail first two points are duplicates and last two points are duplicates as well; first two being the AT's pos and last two being the caster's pos.

PR #28118, opened 2022-07-15
Scripts/Spells: Priest - Power of the Dark Side fixed

Nyr97:
**Changes proposed:**

-  Damage or healing modifier is applied OnEffectLaunchTarget.
-  Penance's tint will be removed once Power of the Dark Side aura is gone.

**Issues addressed:**

None.

**Tests performed:**

None.


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -585,6 +589,112 @@ class spell_pri_penance : public SpellScript
...
+        return ValidateSpellInfo({ SPELL_PRIEST_POWER_OF_THE_DARK_SIDE });
+    }
+
+    void HandleOnRemove(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+    {
+        if (Unit* caster = GetCaster())
+        {
+            if (caster->HasAura(SPELL_PRIEST_POWER_OF_THE_DARK_SIDE))
+                caster->RemoveAura(SPELL_PRIEST_POWER_OF_THE_DARK_SIDE);
+        }
```
Seyden, line 608:
```suggestion
        Unit* caster = GetCaster();
        if (!caster || !caster->HasAura(SPELL_PRIEST_POWER_OF_THE_DARK_SIDE))
            return;
            
        caster->RemoveAura(SPELL_PRIEST_POWER_OF_THE_DARK_SIDE);
```

offl: Why not just
```cpp
        caster->RemoveAura(SPELL_PRIEST_POWER_OF_THE_DARK_SIDE);
```
Or if caster = target
```cpp
    GetTarget()->RemoveAura(SPELL_PRIEST_POWER_OF_THE_DARK_SIDE);
```

Nyr97: I thought I was supposed to check if the caster has the aura first before removing it to avoid a crash when attempting to remove it if it was null 🤔

mdX7: crashes during scripting are 99% caused by having `NULL->something()` which is only possible if caster is null in this case. You can simply use RemoveAura, other checks are done inside there

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -585,6 +589,112 @@ class spell_pri_penance : public SpellScript
...
+            }
+        }
+    }
+
+    void Register() override
+    {
+        if (m_scriptSpellId == SPELL_PRIEST_PENANCE_DAMAGE)
+            OnEffectLaunchTarget += SpellEffectFn(spell_pri_penance_damage_healing::HandleLaunchTarget, EFFECT_0, SPELL_EFFECT_SCHOOL_DAMAGE);
+        if (m_scriptSpellId == SPELL_PRIEST_PENANCE_HEALING)
+            OnEffectLaunchTarget += SpellEffectFn(spell_pri_penance_damage_healing::HandleLaunchTarget, EFFECT_0, SPELL_EFFECT_HEAL);
```
Seyden, line 662: Why exactly are you using OnEffectLaunchTarget and not just OnEffectHitTarget and just add the aura on top of the already existing calculation?

src/server/scripts/Spells/spell_priest.cpp
Seyden: > 

yeah GetTarget()->RemoveAura(SPELL_PRIEST_POWER_OF_THE_DARK_SIDE); would be the best solution. I just thought it was a requirement that he has the aura or smth

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -585,6 +589,112 @@ class spell_pri_penance : public SpellScript
...
+
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_PRIEST_POWER_OF_THE_DARK_SIDE_TINT });
+    }
+
+    void HandleOnApply(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+    {
+        if (Unit* caster = GetCaster())
+            caster->CastSpell(caster, SPELL_PRIEST_POWER_OF_THE_DARK_SIDE_TINT, true);
```
Seyden, line 679: if its supposed to be casted on the target then you can just do 

GetTarget()->CastSpell(GetTarget(), SPELL_PRIEST_POWER_OF_THE_DARK_SIDE_TINT, true);

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -585,6 +589,112 @@ class spell_pri_penance : public SpellScript
...
+    }
+
+    void HandleOnRemove(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+    {
+        if (Unit* caster = GetCaster())
+        {
+            if (caster->HasAura(SPELL_PRIEST_POWER_OF_THE_DARK_SIDE_TINT))
+                caster->RemoveAura(SPELL_PRIEST_POWER_OF_THE_DARK_SIDE_TINT);
```
Seyden, line 687: Same like we discussed above

PR #28142, opened 2022-07-20
Core/Movement: Implement SetUnlimitedSpeed for MoveJumpWithGravity

Nyr97:
According to Sylvanas Windrunner's encounter sniffs, her jump spells using SPELL_EFFECT_JUMP_CHARGE can go past 50.0f speed limitation of the client.

**Changes proposed:**

- Implement SetUnlimitedSpeed that allows speed to surpass 50.0f with certain flags or 28.0f by default.
- Implement SetUnlimitedSpeed inside MoveJumpWithGravity init.

**Issues addressed:**

None.

**Tests performed:**

Not entirely sure yet whether the limitation is only based on creatures or both players and creatures. We could check whether the actor is not a player to set it just in case.

### DIFF
--- a/src/server/game/Movement/MotionMaster.cpp
+++ b/src/server/game/Movement/MotionMaster.cpp
@@ -873,6 +873,7 @@ void MotionMaster::MoveJumpWithGravity(Position const& pos, float speedXY, float
         init.SetParabolicVerticalAcceleration(gravity, 0);
         init.SetUncompressed();
         init.SetVelocity(speedXY);
+        init.SetUnlimitedSpeed();
         if (hasOrientation)
             init.SetFacing(pos.GetOrientation());
         if (effect)
--- a/src/server/game/Movement/Spline/MoveSplineInit.h
+++ b/src/server/game/Movement/Spline/MoveSplineInit.h
@@ -146,6 +146,11 @@ namespace Movement
          */
         void SetOrientationFixed(bool enable);
 
+        /* Enables no-speed limit
+         * if not set, the speed will be limited by certain flags to 50.0f, and otherwise 28.0f
+         */
+        void SetUnlimitedSpeed();
+
         /* Sets the velocity (in case you want to have custom movement velocity)
          * if no set, speed will be selected based on unit's speeds and current movement mode
          * Has no effect if falling mode enabled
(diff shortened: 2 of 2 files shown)


xvwyh:
Just gonna leave this here. In EnumeratedString there's only 2 mentions of speed limit I could find, and they're in EnumeratedStringSetIDs 200 and 342. ID 200 looks like taxi flags of some sort, but 342 looks like jump flags. Why? Here are its values:
- "No Facing Control" - can also be found in spell attribute "**JumpCharge** - no facing control" - prevents the unit from changing facing to the jump direction.
- "Speed is absolute" - can also be found in spell attribute "Charge/**JumpCharge** - Use Absolute Speed" - makes it so that the speed that comes from EffectMultiplier is treated not as a multiplier of unit's run speed, but already as the final jump speed.
- "Ignore Speed Limit" - is what we're looking for, and as mentioned in the PR - a use case of it was found in a SPELL_EFFECT_**JUMP_CHARGE** spell.
- "Fall Damage" - would make complete sense to be used as a jump flag, since originally you would be completely immune to fall damage on spline-based jumps.

You can also find in the .exe that "JumpChargeParams" is one of the possible SpellMiscValue types (137) and judging by surrounding types (it's preceded by a lot of garrison types) - it was added way back during WoD or immediately afterwards; and that there is "JumpChargeFlagFlags" enum (there are names of enums often listed in the exe, with non-flag enums being written in ALL_CAPS and with flag enums being postfixed with word "Flags" regardless of whether there is that word in the name already or not, leading to names with "FlagsFlags" often at the end).

The TL;DR is: They have a separate record type "JumpChargeParams" which is now used to store data about jump parameters like speed, height, and flags, and it's used by spells, but not by aigroupactions, at least not in 7.3.2 which was the last time I parsed them. One of these flags - the newest one - is the one that disables the spline speed limit. But, sadly, since JumpChargeParams.db2 is not included in the client, this information gives us nothing substantial to work with, except for the warm feeling of knowing that there's one less mystery.

Shauren:
Oh but we already knew about JumpChargeParams existence (master branch also has jump_charge_params in world database, columns decided by us, not leaked)

PR #28147, opened 2022-07-21
Core/Spells: add SetCastTime for spells

Nyr97:
**Changes proposed:**

-  Implement an additional method to allow spells to have a given cast time set OnPrecast hook. Sylvanas Windrunner encounter modifies Wailing's Arrow cast time to 1.5s on phase 3, being the original 3s on phase 1 with no aura involved or different spell as well, e.g.

**Issues addressed:**

None.

**Tests performed:**

None.

### DIFF
--- a/src/server/game/Spells/Spell.h
+++ b/src/server/game/Spells/Spell.h
@@ -579,6 +579,7 @@ class TC_GAME_API Spell
         UsedSpellMods m_appliedMods;
 
         int32 GetCastTime() const { return m_casttime; }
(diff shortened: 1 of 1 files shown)


Ovahlord:
CastSpellExtraArgs supports spell mods. you can just apply a spell mod for casting time

Nyr97:
there's no SpellValueMod for CastTime, only SpellModOp allows ChangeCastTime and the caster is not a player, unless I'm looking at the wrong place 🤔

PR #28153, opened 2022-07-25
Core/Spells: Implement OnEffectAbsorbHeal and AfterEffectAbsorbHeal hooks

Nyr97:
We're missing proper implementation of basic absorb heal hooks for spellscripts. Credits to @Seyden for providing the base code for the implementation.

**Changes proposed:**

- Implement OnEffectAbsorbHeal aura hook.
- Implement AfterEffectAbsorbHeal aura hook.

**Issues addressed:**

None.

**Tests performed:**

None.


Ovahlord:
how about not just restricting this to heal but insead to all kinds of absorb?

Nyr97:
there is an existent hook for absorb damage, I too thought it would be more efficient to add support to it for healing rather than splitting them, but I don't know the implications it would have and the final decision is @Shauren

Seyden:
So like what? You both know that healing has HealInfo and Damage Absorb does have DamageInfo?

Shauren:
ignore him

PR #28161, opened 2022-07-29
Fix cripple and curse spells for 25 man Noth

### LINKED ISSUE #28160: [3.3.5] Noth 25 does not curse enough players
### Description

When in the Noth encounter on 25 man, only 3 players are cursed instead 8

### Expected behaviour

8 players should be cursed

1) Engage Noth on 25 man
2) Wait for curse to go out and see players affected

### Branch

3.3.5

### TC rev. hash/commit

53cc21b1b45e+

### Operating system

Ubuntu 21.10

### Custom changes

None

danjohnso:
**Changes proposed:**

-  Fix spell cast for 25 man Noth of Curse and Cripple
-  The spell ids were labeled in source correctly, but were not taken in consideration when cast
-  

**Issues addressed:**

Closes  #28160 

**Tests performed:**

None

### DIFF
--- a/src/server/scripts/Northrend/Naxxramas/boss_noth.cpp
+++ b/src/server/scripts/Northrend/Naxxramas/boss_noth.cpp
@@ -54,8 +54,10 @@ enum Talk
 
 enum Spells
 {
-    SPELL_CURSE         = 29213, // 25-man: 54835
-    SPELL_CRIPPLE       = 29212, // 25-man: 54814
+    SPELL_CURSE         = 29213,
+    SPELL_CRIPPLE       = 29212,
+    SPELL_CURSE_25      = 54835,
+    SPELL_CRIPPLE_25    = 54814,
 
     SPELL_TELEPORT      = 29216, // ground to balcony
     SPELL_TELEPORT_BACK = 29231  // balcony to ground
@@ -210,7 +212,7 @@ struct boss_noth : public BossAI
             {
                 case EVENT_CURSE:
                 {
-                    DoCastAOE(SPELL_CURSE);
+                    DoCastAOE(RAID_MODE(SPELL_CURSE, SPELL_CURSE_25));
                     events.Repeat(randtime(Seconds(50), Seconds(70)));
                     break;
                 }
@@ -223,7 +225,7 @@ struct boss_noth : public BossAI
                     events.Repeat(Seconds(40));
                     break;
                 case EVENT_BLINK:
-                    DoCastAOE(SPELL_CRIPPLE, true);
+                    DoCastAOE(RAID_MODE(SPELL_CRIPPLE, SPELL_CRIPPLE_25), true);
                     DoCastAOE(SPELL_BLINK);
                     ResetThreatList();
                     justBlinked = true;


src/server/scripts/Northrend/Naxxramas/boss_noth.cpp
```diff
@@ -54,8 +54,10 @@ enum Talk
 
 enum Spells
 {
-    SPELL_CURSE         = 29213, // 25-man: 54835
-    SPELL_CRIPPLE       = 29212, // 25-man: 54814
+    SPELL_CURSE         = 29213,
+    SPELL_CRIPPLE       = 29212,
+    SPELL_CURSE_25      = 54835,
+    SPELL_CRIPPLE_25    = 54814,
```
Ovahlord, line 60: don't use the raid mode macro here and move the spellds to spelldifficulty so the core automatically picks the 25 player version.

Jildor:
```sql
-- Noth the Plaguebringer
DELETE FROM `spelldifficulty_dbc` WHERE `id` IN (29212, 29213);
INSERT INTO `spelldifficulty_dbc` (`id`, `spellid0`, `spellid1`, `spellid2`, `spellid3`) VALUES
(29212, 29212, 54814, 0, 0), -- Cripple
(29213, 29213, 54835, 0, 0); -- Curse of the Plaguebringer
```

danjohnso: I see the spell manager already handles the max targets piece, is it preferred to be inline @Jildor ?

Shauren: I disagree with that suggestion, spelldifficulty_dbc is incompatible with master (and wotlk classic)

Shauren (approved):

src/server/scripts/Northrend/Naxxramas/boss_noth.cpp
danjohnso: So this is the preferred way to handle this?  I have a bunch of other fixes for Naxx bosses hopefully I can finish this weekend

Jildor: Trying to unserstand why if pr is for 335 branch and spelldifficulty was the way, now we should avoid use that table? 

P.D. you could have waited to add the suggestion to limit the maximum targets for curse before fast merge xD

Shauren: Target limits are already in SpellMgr

Jildor: Yes, I saw it a moment ago, even so I don't understand what the standard is, is it deprecated in the future spelldifficulty_dbc?

And for maxtargets case if not in spellmgr, the standard is in spellmgr or spellmod in castspell?

Shauren: The way I see it yes, custom spelldifficulty_dbc rows are now deprecated
And for maxtargets, it doesn't matter

danjohnso: https://github.com/TrinityCore/TrinityCore/blob/6b2fac687d05e14ae3e7b947b1193c39cc0d43c1/src/server/game/Spells/SpellMgr.cpp#L3447-L3452
So I was thinking about the max target value for 25 man, I think it should actually be 10?  Not sure what the standard is around here for proof of original mechanics, but one of the people I was running this with said it should be 10, I found this video from 13 years back mentioning it being 10, and wowpedia also says 10.

https://wowpedia.fandom.com/wiki/Noth_the_Plaguebringer
https://www.youtube.com/watch?v=zoQli9GzXW8

PR #28170, opened 2022-08-01
Core/Players: Only send QuestUpdateAddCredit on specific objective types

Ghaster:
**Changes proposed:**

-  Don't send QuestUpdateAddCredit on QUEST_OBJECTIVE_ITEM.
-  This behaviour can also be observed in sniffs. No QuestUpdateAddCredit is sent when looting quest items.

**Issues addressed:**

**Tests performed:**
Builds, works ingame etc.

- [ ] Haven't checked QuestUpdateAddCreditSimple


Shauren:
> AE/PTR builds explicitly check if the packet is one of the three types and asserts if not.

I'm not seeing this. assertion only fires when objective with { type, objectid } pair is not found in quest template AND type is not 0, 2, 3

Client explicitly also handles 9, 10, 11, 12, 13, 14, 15, 19, 20

Ghaster:
Yeah mb, missed the missing quest template stuff. Shall I change it to only prevent sending on QUEST_OBJECTIVE_ITEM?

Shauren:
yes

Ghaster:
Updated

PR #28192, opened 2022-08-11
Core/Gossips: Added support for FriendshipFactionID

meji46:
**Changes proposed:**

-  Added support for FriendshipFactionID. It displays a reputation bar at the top of the gossip window

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -9687,6 +9687,48 @@ void ObjectMgr::LoadGossipMenuItems()
     TC_LOG_INFO("server.loading", ">> Loaded " SZFMTD " gossip_menu_option entries in %u ms", _gossipMenuItemsStore.size(), GetMSTimeDiffToNow(oldMSTime));
 }
 
+void ObjectMgr::LoadGossipMenuFriendshipFactions()
+{
+    uint32 oldMSTime = getMSTime();
+
+    _gossipMenuFriendshipFactionsStore.clear();
+
+    //                                               0       1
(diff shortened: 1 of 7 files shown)


funjoker:
why not let GetGossipMenuFriendshipFaction return int32 instead of uint32 ?

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -9687,6 +9687,48 @@ void ObjectMgr::LoadGossipMenuItems()
...
+    {
+        TC_LOG_INFO("server.loading", ">> Loaded 0 gossip_menu_friendship_faction IDs. DB table `gossip_menu_friendship_faction` is empty!");
+        return;
+    }
+
+    do
+    {
+        Field* fields = result->Fetch();
+
+        uint32 MenuID = fields[0].GetUInt32();
```
mdX7, line 9709: local variables always lowerCamelCase

meji46: That's what I get for copying the name from the SQL query 😄 Fixed

meji46:
> why not let GetGossipMenuFriendshipFaction return int32 instead of uint32 ?

I was relying on FactionEntry::ID type instead of the packet field. I changed the return of the function to avoid the explicit cast. Thanks!

mdX7 (approved):

Shauren:
Not bad, I was just having some design thoughts - maybe we shouldn't have named `gossip_menu` (the table holding menu texts) as `gossip_menu` - this way we cant easily have some sort of a "template" table for gossip menu (where this new column could exist) without causing confusion or too much data split

PR #28193, opened 2022-08-11
Script/Commands: Added .quest objectivecomplete command to complete s…

mdX7:
…ingle quest objectives

**Changes proposed:**

-  Added .quest completeobjective command to complete single quest objectives

**Tests performed:**
tested ingame


mdX7:
another option would be changing it to `.quest objectivecomplete` so people who are used to shortcuts like `.q com` dont have to adapt to this change

PR #28198, opened 2022-08-15
Core/Player: Remove name check when logging in

ihm-tswow:
**Changes proposed:**
Remove the name check that fires when players are just logging in. 

**Issues addressed:**
Name checks run in the main thread and are very expensive, but the same checks run whenever someone creates or modifies a character anyways so it never does anything unless `reserved_name` or the profanity/reserved dbcs are changed. This hits very hard when many players log in after a restart or crash.

Almost 30ms on my machine in release mode for a single player:
image

**Tests performed:**

- Creating new characters with profane/reserved names
- Renaming a character with profane/reserved names (AT_LOGIN_RENAME)
- Recustomizing a character with profane/reserved names (AT_LOGIN_CUSTOMIZE)

**Possible adjustments**
- Make this configurable in .conf to allow modifying those dbcs after initial server launch
- Keep the `reserved_name` check to allow adding things like admin names after the fact, since it's almost always next to free to do anyways.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -17183,17 +17183,6 @@ bool Player::LoadFromDB(ObjectGuid guid, CharacterDatabaseQueryHolder const& hol
 
     m_name = fields[2].GetString();
 
-    // check name limitations
-    if (ObjectMgr::CheckPlayerName(m_name, GetSession()->GetSessionDbcLocale()) != CHAR_NAME_SUCCESS ||
-        (!GetSession()->HasPermission(rbac::RBAC_PERM_SKIP_CHECK_CHARACTER_CREATION_RESERVEDNAME) && sObjectMgr->IsReservedName(m_name)))
-    {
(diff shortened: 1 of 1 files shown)


jackpoz:
If wonder if it's the GetSession()->HasPermission() bit that loads permissions and triggers the delay. Do you have the chance to profile which function inside Player::LoadFromDB() causes the delay ?

Shauren:
No, its probably checking every single regex in NamesProfanity.dbc and NamesReserved.dbc that takes time

ihm-tswow:
It's almost definitely the regex searches, there are 8000 of them just for profanity and many of them start with wildcards

ihm-tswow:
Just as a headsup, these tests were a fluke and I'm going to properly re-do my profiling of the entire function. I was relying on hazy memories of some old results I had when working in a fork where the real results were probably closer to about 3ms (in my head, 30ms souned about like 3ms), but didn't stop to think properly.

For the sake of brevity, i'll probably close this specific PR and make one that covers the entire function properly instead, and i'll make sure to recreate the actual scenarios I've encountered

ihm-tswow:
Since these numbers were wildly off and this fix does very little on its own for now, i'll close this and restart once I've gotten other systems to where it actually makes a difference

PR #28200, opened 2022-08-17
Core/Battlegrounds: Capture flag immediately when standing on capture point and your team's flag return due to drop timer ended

Jildor:
**Changes proposed:**

-  Capture flag immediately when standing on capture point and your team's flag return due to drop timer ended
>When the flag is dropped, it becomes a clickable object where it was dropped. A capturing team member can pick it up again and keep running, but the player who dropped it will not be able to personally pick it up again for 3 seconds. Or, a defending team member can click on it and it will instantly teleport back to their base (this is called "returning" the flag). If no player picks up or returns the flag for about five seconds, it will return itself.
https://wowwiki-archive.fandom.com/wiki/Warsong_Gulch

Flag can return to the base clicking on it or after a few seconds if no one click it.

first case was fixed here: https://github.com/TrinityCore/TrinityCore/pull/28087
but second case need fix it.

**Issues addressed:**

None

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
+++ b/src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
@@ -292,6 +292,8 @@ void BattlegroundWS::RespawnFlagAfterDrop(uint32 team)
 
     SetDroppedFlagGUID(ObjectGuid::Empty, GetTeamIndexByTeamId(team));
     _bothFlagsKept = false;
+    // Check opposing flag if it is in capture zone; if so, capture it
+    HandleFlagRoomCapturePoint(team == ALLIANCE ? TEAM_HORDE : TEAM_ALLIANCE);
 }
 
 void BattlegroundWS::EventPlayerCapturedFlag(Player* player)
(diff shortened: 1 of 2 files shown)


src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp
```diff
@@ -381,6 +383,13 @@ void BattlegroundWS::EventPlayerCapturedFlag(Player* player)
         _flagsTimer[GetTeamIndexByTeamId(player->GetTeam()) ? 0 : 1] = BG_WS_FLAG_RESPAWN_TIME;
     }
 }
+void BattlegroundWS::HandleFlagRoomCapturePoint(int32 team)
+{
+    Player* flagCarrier = ObjectAccessor::GetPlayer(GetBgMap(), GetFlagPickerGUID(team));
+    uint32 areaTrigger = team == TEAM_ALLIANCE ? 3647 : 3646;
```
Faq, line 389: Magic numbers? :)

Jildor: are the same here: https://github.com/TrinityCore/TrinityCore/blob/bbf4cd5ac6b12f76e719c9498868b706145be0db/src/server/game/Battlegrounds/Zones/BattlegroundWS.cpp#L674

(and same in the other bg scripts)

Faq: You can be the first one who changes that :)

Jildor: Better in other pr, I prefer not to mix

PR #28202, opened 2022-08-17
Core/Startup: Multithread database queries

ihm-tswow:
Since maintainers had some comments on the structure of this proposal, I thought it's probably best to just submit it without  polish so it can first be decided what parts if any you actually want. It should be fully functional to try out and measure as is.

**Changes proposed:**

-  Run next to all world loading queries right as loading starts to let the database threads and io do their job without waiting for the main thread between each query. This saves me between **4 and 6 seconds** of startup time in release mode. If combined with the other optimizations i've posted and disabling localization, you can load the worldserver in like 3-4 seconds total.
 
-  This requires us to store queries themselves somewhere we can access it **before** we start loading, so they can't remain inline with the loader functions anymore. In the current draft, all sql syntax has been moved to [LoadDataQueries.cpp](https://github.com/ihm-tswow/TrinityCore/blob/faster-database-load/src/server/game/Globals/LoadDataQueries.cpp), and referenced from places like `ObjectMgr.cpp` etc.
- Note that any benefit requires you to **increase world database thread counts in worldserver.conf**

**Discussion**
The biggest change here is moving all the SQL syntax out of the functions that actually use them, which some maintainers took issue with. There are a few ways this can be addressed or resolved:

- Edit in multiplate tabs. When a query becomes big enough that using the wrong index becomes a real issue, the sql code is often too big to be displayed in the same code window anyways, and it's just as easy to have two views into the same files as into two separate files. The SQL syntax is very easy to find from the loader function in any IDE.

- Cover less queries. I did this for almost every loading query so I could mess around with them and measure more, but I've found that only about 20 of them actually make a meaningful difference in its current implementation. It's possible to keep most of the optimization without nuking half of `ObjectMgr.cpp`

- Move the `LoadDataQuery` objects right above the functions that use them, e.g.
```c++
LoadDataQuery LoadCreatureTemplateQuery("select a,b,c,d,e,f,g from creature_template");
void LoadCreatureTemplate() {
     QueryResults results = LoadCreatureTemplateQuery.GetOrQueryResults();
    // do your thing
}
```

- Use some kind of loader class that ties SQL to the function that use it, e.g. something like this:

```c++
class LoadDataClass {
    virtual void Load() { blah blha blah }
    virtual std::string GetSQL() = 0;
protected:
    virtual void OnLoad(Result result) = 0; 
    // internal implementations blah blah blah
};

...

class LoadCreatureTemplate : public LoadDataClass {
    // Do this in the class directly or in cpp file or whatever
    void GetSQL() { return "select a,b,c,d,e from creature_template"; }
    void OnLoad(QueryResult result) { ... load creature templates here ... }
};

LoadDataCreatureTemplate LoadCreatureTemplateQuery; // call Execute directly from World.cpp / command handlers
```

- Or any combination of the above.

**Tests performed:**

**None** - This is currently a **draft** that will probably change significantly

**TODO list:**
- [ ] Decide if you want this optimization or not
- [ ] Decide how many tables you want this optimization for (I estimate about ~20 of them actually make a meaningful difference at this point)
- [ ] Decide how you want the patch to be formatted for those tables

And once all that's done:

- [ ] Review system
- [ ] Review/test individual tables

**Known issues / Further work**
- [ ] Busy waiting for queries to complete
- [ ] Most index comments have been stripped, i just didn't fill them in yet.
- [ ] It might be possible to be more clever with the order we load queries in to maximize CPU utilization. I haven't really analyzed this too much and probably won't bother, but keeping more rather than less tables like this would keep the door more open for further such optimizations.


src/server/game/Globals/LoadDataQueries.cpp
```diff
@@ -0,0 +1,1230 @@
...
+static std::vector<LoadDataQuery*> asyncQueries;
+
+LoadDataQuery::LoadDataQuery(std::string const& query)
+    : m_query(query)
+{}
+
+QueryResult LoadDataQuery::GetOrQueryResults()
+{
+    WaitForAsyncQuery();
+    if (m_asyncResult.has_value())
```
Lordron, line 14: can race condition occur here if GetOrQueryResults called the same time callback executed?

Lordron:
master will require PreparedStatement support to support hotfix loading

Warpten:
I personally do not think this is worthwhile to pursue (no offense), although i have little to no say in how TC runs anymore - if I ever had any.

However, I really don't like the fact that all load queries definitions are crammed in a single .cpp file. Splitting those in counterparts to the systems they relate to (`LoadQueries.Achievement.cpp` for `AchievementMgr` for example) could be an option, assuming it's ok to have potentially an explosion of the number of files related to this. The same could be done to their declarations; wether or not that will have any impact on build time is yet to be seen (I suspect not).

In my opinion, a true async implementation would be a complete rewrite of the load logic, including queries, to process rows as they come instead of waiting for the entire result set; basically a MPSC queue, which - fortunately! - I think there already is a base implementation for in the code base.

Here are my two cents. Make of that what you will.

src/server/game/Globals/LoadDataQueries.cpp
ihm-tswow: > can race condition occur here if GetOrQueryResults called the same time callback executed?

Currently, the callback is only executed from inside `WaitForAsyncQuery` (that's the only place that calls `InvokeIfReady`) so as long as any callers are single-threaded (and they are, assuming we start these queries before any tables load), that should not be possible.

ihm-tswow:
> I personally do not think this is worthwhile to pursue (no offense), although i have little to no say in how TC runs anymore

Don't worry about that, any input is valuable here 🙂. For my fork this tweak is a gamechanger because of how much we restart the worldserver, I just posted it here in case anyone, tc maintainer or not, finds it useful. I estimate that about 95%+ of the time save can be retained by reducing the amount of tables this covers from 160 to about 20, and maybe 70% if reducing it to about 5 (but that's just a guess). If rejected i'll probably reduce the tables like this for my fork to avoid future merge conflicts anyways.

> However, I really don't like the fact that all load queries definitions are crammed in a single .cpp file. Splitting those in counterparts to the systems they relate to (`LoadQueries.Achievement.cpp` for `AchievementMgr` for example) could be an option, assuming it's ok to have potentially an explosion of the number of files related to this. The same could be done to their declarations; wether or not that will have any positive impact on build time is yet to be seen (I suspect not).

That would be about 20 more translation units assuming every query is kept (and with only the 20 worst tables kept, about 3 translation units), but like 17 of these only contain one or two queries as the vast majority of queries are in `ObjectMgr`, `GameEventMgr` and `SpellMgr`. It's also possible to place definitions inside the files that use them as I described to not separate sql syntax from loader functions, but that would probably make the declarations look kinda wonky or placed in all kinds of random places.

> In my opinion, a true async implementation would be a complete rewrite of the load logic, including queries, to process rows as they come instead of waiting for the entire result set; basically a MPSC queue, which - fortunately! - I think there already is a base implementation for in the code base. And yet I don't even know if that is possible, because if I recall correctly, that is not something the database driver provides.

I'm unaware of any such functionality, but my knowledge on it is very limited. However, from my tests, doing this per-table in the way you suggest would do (surprisingly) little for total startup time:

1. It seems the biggest performance gain here comes from firing multiple queries at once (i.e. parallelism in dbthreads and/or mysql itself). I didn't really analyze this, but it can be illustrated by just waiting for all queries right after firing them, which makes very little difference for total startup time on my machine compared to interweaving it the way I do now; in both cases total time saved is between 4 and 6 seconds. Unfortunately, most bad tables are dependent on each others and thus their sql cannot be parallelized with that model. It's possible that improving the threading model for dbthreads and/or updating mysql can improve this, but I haven't really looked into it.

2. Even assuming time can be saved with cpu-bound loader work, that work rarely happens inside the loaders with very bad sql times anyways, but in unrelated functions like dbc and those dependent on the entire tables like query data. ScriptNames takes over ~1.4s to query and ~1ms to process, broadcast_text+locales takes ~1.2s in total, ~400ms to process, localization strings ~1.7s in total, ~300ms to process, gameobjects/items/creatures/smartai spend between 70/80% of their time in sql.

jackpoz:
> * Note that any benefit requires you to **increase world database thread counts in worldserver.conf**

Just for info, async db connections are handled poorly by TC as they can run out of order or cause mysql deadlocks, with the only remediation to keep a low number of db async threads (this is visible when instance reset happens with 1 transaction queued for each map affecting 3 tables, with different db threads locking the same tables/rows and spamming db deadlocks)

ihm-tswow:
I don't think deadlocks can happen during startup at least. If waiting for these before processing, these are 100% reads and I think the very few writes that can happen if processing concurrently are to unrelated tables, neither of which involve transactions.

I could look at some of the world db inserts (`npc create` etc), but i don't think any of those involve transactions either.

Lordron:
As mentioned by @Warpten trully async server load requires lots of refactoring to be made. Async query is ok, but why stop here if we can call load functions async. Well there is a big problem.

**Almost** each of the load function depends on previous (one or more) whether it altering already loaded data or loading new. Example Function2 expect FlagX be added to spells by calling Function1 before. Changes to templates in Function1 **may** alter result of Function2.

Example - You can't load creature before loading creature templates and spawn groups.
There is a small amount of load functions whose data are never used during loading process. Creature\Gameobject\Quest locales are good example. These can be easily called async by passing IO context in SetInitialWorldSettings.

```
    Trinity::Asio::post(*ioContext, []() {
        sObjectMgr->LoadCreatureLocales();
        sObjectMgr->LoadGameObjectLocales();
    });
```

But that will save us about 5 sec. Not a huge win. 

We can imagine a loadings process as a chains where you can load each chain async to other without breaking data.
To make a proper async loading we need a full view on loading chains. So i made a graph that shows all connections between tables (actually almost all, accuracy ~90, smth may missing).  URL is big, have to use tinyurl to shorten https://tinyurl.com/3x4kxa5v
As you see it's a mess. It looks like a single chain that can't be processed async. Only small amount of tiny tables. 

**Loading process need to be splited to loading and validation.**

Fast checking code showed some candidates that can be splited to loading and validation. Graph after https://tinyurl.com/4pzkpenh

- LoadItemSetNames
- LoadGameObjectTemplateAddons
- LoadCreatureTemplateAddons
- LoadReputationOnKill
- LoadCreatureClassLevelStats
- LoadCreatures
- LoadTempSummons
- LoadCreatureAddons
- LoadCreatureMovementOverrides
- LoadGameObjects
- LoadGameObjectAddons
- LoadGameObjectOverrides
- LoadGameObjectQuestItems
- LoadCreatureQuestItems
- LoadQuests
- LoadNPCSpellClickSpells
- LoadVehicleTemplateAccessories
- LoadVehicleAccessories
- LoadCreatureDefaultTrainers

ihm-tswow:
@Lordron I like that idea, I considered something similar before but stumbled into the same issues you mentioned, and I wasn't sure if we are willing to actually split validation and loading like that.

I'd be willing to investigate this if other maintainers prefer it to the current solution. However, just be aware, your proposal will be slower than my implementation since you do not cover script name loading (which is one entire second) or the locale/broad_text tables (another 1-2 seconds). That issue can probably be solved either by combining it with my solution or by working it into your proposal, I think both are possible, but we'd have to try it to know if it actually loses time or not.

However, just be aware, the biggest benefit would be keeping most of the loading structure intact (instead of moving it out like i do), and I highly doubt you will save more than a few hundred milliseconds by doing this compared to the ~20 tables I covered initially. This is because almost no time is spent actually processing these tables, so there isn't much cpu-bound work to save.

Shauren:
You basically want something like this https://github.com/taskflow/taskflow

Aokromes:
src/server/game/Globals/LoadDataQueries.cpp:44:38: fatal error: no matching literal operator for call to 'operator""ms' with argument of type 'unsigned long long' or 'const char *', and no matching literal operator template
        std::this_thread::sleep_for(5ms);
                                     ^
1 error generate

Aokromes:
Conflicting files
src/server/game/Globals/ObjectMgr.cpp

Winfidonarleyan:
Based this pr https://github.com/WarheadCore/WarheadBand/commit/5b15ba5f07a75038b7f4b761c2a321cefc1f2ef4

Aokromes:
src/server/game/Globals/ObjectMgr.cpp
src/server/game/Loot/LootMgr.cpp

if no fixed before 1 april i will close as abandoned.

Aokromes:
3 weeks left

PR #28206, opened 2022-08-19
Core/Debugging: Improve SymInitialize fail message

ihm-tswow:
**Changes proposed:**

-  Improve DbgHelp SymInitialize error message to be less confusing

**Issues addressed:**
None, but i've seen lots of confusion about this error.

**Tests performed:**
This is how new output looks (well, with a real error message of course, I just copied it out of the if clause locally to test formatting)

```
----
SYMBOL HANDLER ERROR (THIS IS NOT THE CRASH ERROR)

Couldn't initialize symbol handler for process when generating crash report
Error: The operation completed successfully.

THE BELOW CALL STACKS MIGHT HAVE INACCURATE FILE/FUNCTION NAMES

----
```

- This is all I know about the error, there might be something else we should let the user know.

### DIFF
--- a/src/common/Debugging/WheatyExceptionReport.cpp
+++ b/src/common/Debugging/WheatyExceptionReport.cpp
@@ -718,7 +718,7 @@ PEXCEPTION_POINTERS pExceptionInfo)
             Log(_T("SYMBOL HANDLER ERROR (THIS IS NOT THE CRASH ERROR)\r\n\r\n"));
             Log(_T("Couldn't initialize symbol handler for process when generating crash report\r\n"));
             Log(_T("Error: %s\r\n"), ErrorMessage(GetLastError()));
-            Log(_T("THE BELOW CALL STACKS MAY NOW HAVE MISSING OR INACCURATE FILE/FUNCTION NAMES\r\n\r\n"));
+            Log(_T("THE BELOW CALL STACKS MIGHT HAVE MISSING OR INACCURATE FILE/FUNCTION NAMES\r\n\r\n"));
             Log(_T("END OF SYMBOL HANDLER ERROR\r\n"));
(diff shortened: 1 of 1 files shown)


src/common/Debugging/WheatyExceptionReport.cpp
```diff
@@ -713,8 +713,14 @@ PEXCEPTION_POINTERS pExceptionInfo)
...
         {
-            Log(_T("\n\rCRITICAL ERROR.\n\r Couldn't initialize the symbol handler for process.\n\rError [%s].\n\r\n\r"),
-                ErrorMessage(GetLastError()));
+            Log(_T("\r\n"));
+            Log(_T("----\r\n"));
+            Log(_T("SYMBOL HANDLER ERROR (THIS IS NOT THE CRASH ERROR)\r\n\r\n"));
+            Log(_T("Couldn't initialize symbol handler for process when generating crash report\r\n"));
+            Log(_T("Error: %s\r\n"), ErrorMessage(GetLastError()));
+            Log(_T("THE BELOW CALL STACKS MIGHT HAVE MISSING OR INACCURATE FILE/FUNCTION NAMES\r\n\r\n"));
+            Log(_T("END OF SYMBOL HANDLER ERROR\r\n"));
```
Shauren, line 722: Is this line really neccessary?

ihm-tswow: Which one? the last or the first? I added the last so there is 0 confusion about where symbol messages start and begin, first one just clarifies it's related to error reporting

Shauren: the END OF

ihm-tswow: fair enough

PR #28218, opened 2022-08-25
Core/Garrisons: Implement SMSG_GARRISON_OPEN_TALENT_NPC  opcode

aquadeus:
Co-authored-by: ModoX <moardox@gmail.com>

**Changes proposed:**

-  Implement SMSG_GARRISON_OPEN_TALENT_NPC opcode and add a table to put all npcs with talent trees.

**Tests performed:**

Tested in-game


mdX7:
@meji46 just thinking of it: maybe we should actually bind FriendshipFactionID to creature instead of gossip_menu/the new table here? Chromies Death Scenario is the only case where both is used, but it would be redundant data if it's saved separately. Also I don't know of a case where one creature has two different friendshipfactions

src/server/game/Garrison/GarrisonMgr.cpp
```diff
@@ -461,3 +462,59 @@ void GarrisonMgr::LoadFollowerClassSpecAbilities()
...
+        if (FactionEntry const* faction = sFactionStore.LookupEntry(friendshipFactionID))
+        {
+            if (!sFriendshipReputationStore.LookupEntry(faction->FriendshipRepID))
+            {
+                TC_LOG_ERROR("sql.sql", "NPC entry %u has invalid FriendshipFactionID (%u) in `garrison_talent_npc`, set to 0.", npcEntry, friendshipFactionID);
+                friendshipFactionID = 0;
+            }
+        }
+
+        GarrisonTalentNPC& data = _garrisonTalentNPCs[npcEntry];
```
mdX7, line 514: Npc entry field is not being set. We MIGHT actually drop the field entirely since its used as key. Adding it to the struct mostly is for cases when the struct is passed somewhere and you might want to access the entry afterwards

src/server/game/Garrison/GarrisonMgr.cpp
```diff
@@ -461,3 +462,59 @@ void GarrisonMgr::LoadFollowerClassSpecAbilities()
...
+
+GarrisonTalentNPC const* GarrisonMgr::GetTalentNPCEntry(int32 entry) const
+{
+    return Trinity::Containers::MapGetValuePtr(_garrisonTalentNPCs, entry);
+}
+
+void GarrisonMgr::LoadTalentNPCs()
+{
+    //                                                      0                 1                    2
+    QueryResult result = WorldDatabase.Query("SELECT npcEntry, garrTalentTreeID, friendshipFactionID FROM garrison_talent_npc");
```
mdX7, line 474: you defined the fields in UpperCamelCase in sql file, you should use naming 1:1 like it is there

meji46:
> @meji46 just thinking of it: maybe we should actually bind FriendshipFactionID to creature instead of gossip_menu/the new table here? Chromies Death Scenario is the only case where both is used, but it would be redundant data if it's saved separately. Also I don't know of a case where one creature has two different friendshipfactions

@mdX7 I thought of initially binding the FriendshipFactionID to the creature but, the creatures that use it, don't send it for all their gossips. Generally it's only sent for the default gossip_menu, but you have the case of Gina (https://github.com/TrinityCore/TrinityCore/commit/1bbc8592d95978a4c681fa6177c7c457189ef485#diff-f7470730f9bc2e3c15980ac9bde308269f395e297468674803181d3c02a58c38R13) who sends it for her default gossip_menu and only for one of her multiple ActionMenuId (gossip_menu_option).

PR #28220, opened 2022-08-27
Core/Gossips: Drop OptionType for gossip_menu_option

meji46:
**Changes proposed:**

-  Rename field **OptionIcon** to **OptionNpc**  to reflect the true meaning.
-  Drop the field **OptionType** to define the behavior of gossip_menu_option. Now the field **OptionNpc** defines this behavior. 
- GOSSIP_OPTION_OUTDOORPVP defines a custom type used only by Zangarmarsh. @mdX7 will update the script to remove unnecessary functions and use a script on the specific NPC that made use of this system. #28224
- Implemented GossipOptionIcon::DisableXPGain and GossipOptionIcon::EnableXPGain. Deleted the old script "npc_experience".
- Updated GossipOptionIcon enum (and renamed to GossipOptionNpc).
- Updated fallback gossip menu options.
- Log error when using ActionMenuID or ActionPoiID with GossipOptionNpc different from GossipOptionNpc::None.

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

- The struct **GossipMenuItem** has been updated by renaming **OptionType** to **Action** to continue to reflect the use of eTradeskill::GOSSIP_ACTION_* in old scripts. In other PR I'll remove this field from the struct and update the old scripts.


Luzifix:
An icon should not define the functionality of a Gossip action. This will lead to a lot of problems with custom gossips.

Shauren:
> An icon should not define the functionality of a Gossip action. This will lead to a lot of problems with custom gossips.

You are mistaken, this field is not icon but real option type and does have different functionality in client, it is only called "icon" because someone in the past only notice that effect on client without any behavior changes

meji46:
> You are mistaken, this field is not icon but real option type and does have different functionality in client, it is only called "icon" because someone in the past only notice that effect on client without any behavior changes (yes, icons are hardcoded per option type on retail)

Should I rename the names to reflect the true meaning?
* GossipOptionIcon -> GossipNpcOption
* GossipMenuItems.OptionIcon -> OptionNpc
* GossipMenuItem.MenuItemIcon -> OptionNpc

> An icon should not define the functionality of a Gossip action. This will lead to a lot of problems with custom gossips.
We lose the possibility to generate a gossip_menu_option with an icon that has nothing to do with the assigned behavior but as Shauren says, this field is not a real icon

offl:
You can use default gossip icon for custom scripts. Especially because someone in future can also decide to drop OptionNpcFlag and if he is crazy enough start sending npc flags if creature at this current moment has specific gossip option with specific icon(type), if that is how those dynamic flags are sent.

shamage:
An error occurred during the update. Please, is it possible to fix it? "SQL syntax error"

https://gist.github.com/shamage/a0dec810b0a7428f71d5e57a06df7258

PR #28226, opened 2022-08-31
Core/Auras: Implement SPELL_AURA_MOD_BONUS_ARMOR_PCT and rename HandleModStatBonusArmor

aquadeus:
for consistency

**Changes proposed:**

-  Implement SPELL_AURA_MOD_BONUS_ARMOR_PCT

**Tests performed:**

Tested in-game


src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -3593,14 +3593,26 @@ void AuraEffect::HandleModArmorPctFromStat(AuraApplication const* aurApp, uint8
...
 
     aurApp->GetTarget()->HandleStatFlatModifier(UNIT_MOD_ARMOR, BASE_VALUE, float(GetAmount()), apply);
 }
 
+void AuraEffect::HandleModBonusArmorPercent(AuraApplication const* aurApp, uint8 mode, bool /*apply*/) const
+{
+    if (!(mode & (AURA_EFFECT_HANDLE_CHANGE_AMOUNT_MASK | AURA_EFFECT_HANDLE_STAT)))
+        return;
+
+    Player* target = aurApp->GetTarget()->ToPlayer();
```
mdX7, line 3609: target does not have to be a player, you can simply call `target->UpdateArmor()` on the `Unit*` object. Polymorphism magic does the rest

(not sure if this is relevant for creatures though)

aquadeus: thanks :D

PR #28253, opened 2022-09-15
Kezan/Quests: Implement quest 14071 (Rolling with my homies)

### LINKED ISSUE #16052: Kezan - Rolling with my Homies
The Keys to the Hot Rod Can summon the Hot Rod but I cannot mount this unit. 
http://www.wowhead.com/quest=14071/rolling-with-my-homies

Goblin Start Quest.

TrinityCore rev. 4695e23 2015-12-09 01:33:37 +0100 (6.x branch) (Unix, Release)

aquadeus:
Co-authored-by: Naddley <64811442+Naddley@users.noreply.github.com>

**Changes proposed:**

-  Implement quest "Rolling with my homies"

**Issues addressed:**

Closes #16052

**Tests performed:**

Tested in-game

- [ ] There is an error on the hot rod when you cast the spell "Punch it", it stops movement on apply and remove the aura
- [ ] There is an error here https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Spells/SpellInfo.cpp#L2328 that prevents the cast of the spell when trigger an areatrigger (sends you the error SPELL_FAILED_CANT_DO_THAT_RIGHT_NOW), so i commented lines from 2328 to 2339 to continue working on the rest of the quest.
- [ ] Implement serverside spell 66679, casted when pick the quest, is a periodic aura to check if you lost the homies (f.e. when relogging) and resummons again.

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -4631,6 +4631,15 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->AttributesEx &= ~SPELL_ATTR1_IS_CHANNELLED;
     });
 
+    // Hot Rod Radio
+    ApplySpellFix({ 66298 }, [](SpellInfo* spellInfo)
+    {
+        ApplySpellEffectFix(spellInfo, EFFECT_0, [](SpellEffectInfo* spellEffectInfo)
+        {
+            spellEffectInfo->TargetA = SpellImplicitTargetInfo(TARGET_UNIT_PASSENGER_0);
+        });
+    });
+
     for (SpellInfo const& s : mSpellInfoMap)
     {
         SpellInfo* spellInfo = &const_cast<SpellInfo&>(s);
--- a/src/server/scripts/Maelstrom/kezan.cpp
+++ b/src/server/scripts/Maelstrom/kezan.cpp
@@ -13,6 +13,191 @@
  * with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "Creature.h"
+#include "Player.h"
+#include "ScriptMgr.h"
+#include "SpellAuraEffects.h"
+#include "SpellScript.h"
+#include "Vehicle.h"
+
+enum QuestRollingWithMyHomiesMisc
+{
+    NPC_SUMMONED_ACE                = 34957,
+    NPC_SUMMONED_GOBBER             = 34958,
+    NPC_SUMMONED_IZZY               = 34959,
+
+    SPELL_SERVERSIDE_ABANDON_QUEST  = 78976,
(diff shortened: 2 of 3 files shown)


src/server/game/Spells/SpellMgr.cpp
```diff
@@ -4631,6 +4631,15 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->AttributesEx &= ~SPELL_ATTR1_IS_CHANNELLED;
     });
 
+    // Hot Rod Radio
+    ApplySpellFix({ 66298 }, [](SpellInfo* spellInfo)
+    {
+        ApplySpellEffectFix(spellInfo, EFFECT_0, [](SpellEffectInfo* spellEffectInfo)
+        {
+            spellEffectInfo->TargetA = SpellImplicitTargetInfo(TARGET_UNIT_PASSENGER_0);
```
Ovahlord, line 4639: hack.  TARGET_UNIT_TARGET_PASSENGER should target the first passenger by default. If that's not the case, the core needs a fixup.

Ovahlord: No, you manually target the passenger. That target allows any unit on the vehicle. So instead of hacking the spell to only target seat 0, just explicitely select the player.

aquadeus: i just realize there is a typo error (is not the radio but the horn honk spell), and in the script i selected the player but still says no target without this hack

src/server/scripts/Maelstrom/kezan.cpp
```diff
@@ -13,6 +13,186 @@
...
+                break;
+            case NPC_SUMMONED_GOBBER:
+                args.AddSpellBP0(SEAT_GROBBER);
+                break;
+            case NPC_SUMMONED_IZZY:
+                args.AddSpellBP0(SEAT_IZZY);
+                break;
+        }
+
+        caster->CastSpell(vehicle, VEHICLE_SPELL_RIDE_HARDCODED, args);
```
Nyr97, line 147: I'm not exactly sure if we should be forcing this spell to be cast, but rather access the vehicle so we can mount it up.

aquadeus: this spell is in the sniff

Nyr97: Not saying otherwise, but we have helpers and functions in place that already force the spell to be cast when, e.g., clicking on a mountable npc.

src/server/game/Spells/SpellMgr.cpp
Ovahlord: So this is a bug then that needs to be fixed instead.

PR #28258, opened 2022-09-18
Core/Spells: Fix glyph of stars (druid)

aquadeus:
**Changes proposed:**

-  Fix glyph of stars

**Tests performed:**

Tested in-game


Nyr97:
FYI: it is not possible to change a spell's raw value through a spellscript unless it's about damage calculations and so on. If you wish to change miscvalue, it must be done on Spellmgr.cpp 😿

aquadeus:
Then is no other way to handle this spell

Shauren:
Why aren't you hooking into `Unit::GetModelForForm`? All shapeshift customizations are scripted there

aquadeus:
but this spell has no model id, it just removes the moonkin form

Shauren:
Then make it return 0?

aquadeus:
ok changed

PR #28259, opened 2022-09-18
Feature(Smart Scripts/SMART_EVENT_RANGE): Prevent Initial timer 

acidmanifesto:
**Changes proposed:**

-  prevent initial firing of ranged event and will not fire until repeat time has expired.

**Issues addressed:**

Prevent initial firing of ranged event and will not fire until repeat time has expired.

Currently the event will cast the spell in entering combat which is not desirable.

**Tests performed:**

Tested in game with and without flag

## How to Test the Changes:
1. Recompile with changes
2. Test without flag
3. Insert in to world
```
UPDATE `smart_scripts` SET `event_type`=9, `event_chance`=75, `event_param1`=0, `event_param2`=8, `event_param3`=10000, `event_param4`=15000, `event_param5`=0, `action_param2`=1, `target_type`=1, `comment`="Boulderfist Magus - Within 0-8 Range - Cast 'Frost Nova'" WHERE `entryorguid`=2567 AND `source_type`=0 AND `id`=3;
``` 
4. .go c 14665
5. enter combat
6. Test with flag
7. Insert in to world
```
UPDATE `smart_scripts` SET `event_type`=9, `event_chance`=75, `event_param1`=0, `event_param2`=8, `event_param3`=10000, `event_param4`=15000, `event_param5`=1, `action_param2`=1, `target_type`=1, `comment`="Boulderfist Magus - Within 0-8 Range - Cast 'Frost Nova'" WHERE `entryorguid`=2567 AND `source_type`=0 AND `id`=3;
``` 

After testing you should see the Frost Nova is cast in entering combat without flag and not until repeat timer occurs with flag.

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -2819,7 +2819,7 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
             else
             {
                 if (!e.event.minMaxRepeat.controller)
-                    RecalcTimer(e, 500, 500); //  // make it predictable "This seems to be done to standardize min, max start rather than using the range values for the timer."
+                    RecalcTimer(e, 500, 500); // make it predictable "This seems to be done to standardize min, max start rather than using the range values for the timer."
                 else
                     RecalcTimer(e, e.event.minMaxRepeat.repeatMin, e.event.minMaxRepeat.repeatMax); // if param5 value is greater than 0 first action will not happen until after repeat timer fires.
             }
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -2816,13 +2816,8 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
 
             if (me->IsInRange(me->GetVictim(), (float)e.event.minMaxRepeat.min, (float)e.event.minMaxRepeat.max))
                 ProcessTimedAction(e, e.event.minMaxRepeat.repeatMin, e.event.minMaxRepeat.repeatMax, me->GetVictim());
-            else
-            {
-                if (!e.event.minMaxRepeat.controller)
-                    RecalcTimer(e, 500, 500); //  // make it predictable "This seems to be done to standardize min, max start rather than using the range values for the timer."
-                else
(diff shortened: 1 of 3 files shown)


offl:
>The Range Event is a much better

It's not, there's no such hook. The only place where designers can configure at which range to cast a spell is Creature Spells(wild guess, I'm not even sure if there are spells which should be used at specific range, maybe only spells which has no range, like Rushing Charge)
That event type was used because otherwise spells like Frost Nova indeed will be used even if it will hit nothing(always was thinking it's a bug but never checked), but seems like nobody cared that in that case spell will be used instantly first time, while on retail all those spells has both initial and repeat timers. Significantly reducing usage of that event or drop it completely and switch to In Combat event type, with both initial and repeat timers is much better

malcrom:
Frost Nova is self cast so using In Combat will cast the spell regardless of the range of the player which is not Blizzlike. 
Even if you don't like using this feature addition for that particular spell this it does nothing to affect existing scripts only adds the ability to disable firing on entering combat.

Shauren:
Some random insight on blizzards way of doing things:

* range event is not a thing
* spells are cast using in-combat event
* conditions on each event/action are done using CombatCondition.db2 which references UnitCondition.db2 - and one of the condition types is "within range"

Ignoring all that, why does range event need this change to support that scenario? Did you intend to achieve some sort of initial cooldown since start of combat?

malcrom:
Previous comment:

Even if you don't like using this feature addition for that particular spell it does nothing to affect existing scripts only adds the ability to disable initial firing on entering range.

jackpoz:
I see this PR is supposed to affect only SMART_EVENT_RANGE but then it changes SmartEvent::minMaxRepeat and that is used by many more SMART_EVENT_ . Please add a new struct instead and change SMART_EVENT_RANGE to use that new struct.

I'm unsure about what "controller" is supposed to mean in this case, it's quite unclear to me, so maybe there's a better name ?

This is on top of what everyone else wrote, which might mean the PR will have to be changed.

malcrom:
Thank you Jackpoz for suggestion. Changes incoming soon.

malcrom:
> I see this PR is supposed to affect only SMART_EVENT_RANGE but then it changes SmartEvent::minMaxRepeat and that is used by many more SMART_EVENT_ . Please add a new struct instead and change SMART_EVENT_RANGE to use that new struct.
> 
> I'm unsure about what "controller" is supposed to mean in this case, it's quite unclear to me, so maybe there's a better name ?
> 
> This is on top of what everyone else wrote, which might mean the PR will have to be changed.

Updated.

PR #28272, opened 2022-09-22
Zone/Silverpine: Added spawns in phase 169 in Forsaken High Command

Nyr97:
* added spawns for missing creatures in phase 169 in Forsaken High Command.
* implemented Guts and Gore (questId 26995).
* implemented The Gilneas Liberation Front (questId 26989).
* implemented Iterating Upon Success (questId 26998).
* implemented (Serverside/Non-DB2) Summon Forsaken Warhorse (spellId 147066).

**Changes proposed:**

^  

**Issues addressed:**

None.

**Tests performed:**

It builds and it was tested in-game.

**Known issues and TODO list:** (add/remove lines as needed)

None.

PR #28273, opened 2022-09-22
Zone/Silverpine: Implemented Horde Hauler (npcId 44731)

Nyr97:
NOTE: do not attempt to implement this PR until https://github.com/TrinityCore/TrinityCore/pull/28272 is merged. This one is dependent.

* implemented Magical Chains (spellId 84238).
* implemented Eject Passengers 3-8 (spellId 83477). [Note: I believe this one is used somewhere else, so it might be wise to add it on general purpose scripts).
* added missing pathings, formations and addons for several creatures.

**Changes proposed:**

^

**Issues addressed:**

None.

**Tests performed:**

None.


src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -39,6 +39,251 @@
...
+    }
+
+    void WaypointReached(uint32 waypointId, uint32 pathId) override
+    {
+        if (pathId == PATH_FROM_NORTH_TO_SOUTH)
+        {
+            if (waypointId == WAYPOINT_ON_FORSAKEN_HIGH)
+                _events.ScheduleEvent(EVENT_YELL_ON_FORSAKEN_HIGH, 1s);
+
+            if (waypointId == WAYPOINT_ON_SEPULCHER)
```
mdX7, line 164: else if or switch

idea behind this is:
if the first condition is met it does NOT check if the other conditions are met, while if you are using only ifs (not else ifs) its checking every conditions sequentially

Nyr97: Feedback applied!

PR #28279, opened 2022-09-23
Modernized some scripting models in Halls of Lightning

Ovahlord:
**Changes proposed:**

-  just modernized some old code in Halls of Lightning. There's gonna be more juicy stuff soon(tm)

**Tests performed:**
- built and tested on 4.x


CraftedRO:
01f0f83139e7 Just tested this on both normal and heroic difficulties and everything seems fine , except that my dk pet died in few seconds at Ionar heroic but this is because avoidance doesn't works at all kek.

PR #28281, opened 2022-09-24
Core/Units: Set UNIT_FLAG2_UNTARGETABLE_BY_CLIENT as allowed flag

meji46:
**Changes proposed:**

-  This flag is used by creatures that usually show tooltip but cannot be targeted (e.g. brewfest apple bucket).

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


offl:
Commented flags are 'to dissalow because flag isn't static (not set in db)', e.g. UNIT_FLAG2_REGENERATE_POWER
What you want is just remove it from dissalowed, same with AOI flags and UNIT_FLAG2_UNINTERACTIBLE_IF_HOSTILE, or better all flags that are marked as unused in 335. It was cherry picked without changes

offl:
Somehow UNIT_FLAG2_CANNOT_TURN is commented while in 335 commented flag is UNIT_FLAG2_UNK2, Cannot Turn shouldn't be dissalowed

meji46:
> Commented flags are 'to dissalow because flag isn't static (not set in db)', e.g. UNIT_FLAG2_REGENERATE_POWER What you want is just remove it from dissalowed, same with AOI flags and UNIT_FLAG2_UNINTERACTIBLE_IF_HOSTILE, or better all flags that are marked as unused in 335. It was cherry picked without changes

The commented out flags aren't part of the set of disallowed flags, are they? I commented out UNIT_FLAG2_UNTARGETABLE_BY_CLIENT precisely because it follows the same principle that had been applied to UNIT_FLAG2_REGENERATE_POWER, which can be static

offl:
UNIT_FLAG2_REGENERATE_POWER is set by core if creature doesn't have CREATURE_STATIC_FLAG_NO_AUTOMATIC_REGEN or CREATURE_STATIC_FLAG_5_NO_POWER_REGEN, so that unit flag is commented to make it dissalowed somehow later

UNIT_FLAG_UNK_6 is commented too, that flag with no doubts shouldn't be in DB, so when someone will find a solution how to set it without storing in DB, it can be uncommented

meji46:
> UNIT_FLAG2_REGENERATE_POWER is set by core if creature doesn't have CREATURE_STATIC_FLAG_NO_AUTOMATIC_REGEN or CREATURE_STATIC_FLAG_5_NO_POWER_REGEN, so that unit flag is commented to make it dissalowed somehow later
> 
> UNIT_FLAG_UNK_6 is commented too, that flag with no doubts shouldn't be in DB, so when someone will find a solution how to set it without storing in DB, it can be uncommented

Understood, then the flag is temporarily static until CREATURE_STATIC_FLAG_NO_AUTOMATIC_REGEN and CREATURE_STATIC_FLAG_5_NO_POWER_REGEN are implemented

Then:
- Remove UNIT_FLAG2_UNTARGETABLE_BY_CLIENT, UNIT_FLAG2_UNINTERACTIBLE_IF_HOSTILE and UNIT_FLAG2_CANNOT_TURN
- Regarding the AOI flags, currently the field visibilityDistance from creature_addon is used for this behaviour (taking into account the limits of grids). Should the flags be applied dynamically according to the value of this field or be taken as static values?

offl:
>Regarding the AOI flags, currently the field visibilityDistance from creature_addon is used for this behaviour (taking into account the limits of grids). Should the flags be applied dynamically according to the value of this field or be taken as static values?

I don't know. Despite they are not used, you get thousands of errors because they are currently disallowed

Shauren:
> > Regarding the AOI flags, currently the field visibilityDistance from creature_addon is used for this behaviour (taking into account the limits of grids). Should the flags be applied dynamically according to the value of this field or be taken as static values?
> 
> I don't know. Despite they are not used, you get thousands of errors because they are currently disallowed

They are supposed to be set by visibilityDistance, thats why they are disallowed

offl:
Those flags are handled in client? If no, it doesn't matter if they will be allowed, you just keep using visibilityDistance and add unit flags from sniffs, but without thousands of errors

meji46:
I was working on the implementation to dynamically add the AOI flags according to the VisibilityDistance associated with the creature (since this wasn't being done) but I noticed that there are many cases in instances where VisibilityDistance has been defined for creatures that have no AOI flag
```sql
SELECT c.entry, c.name, c.unit_flags2, ca.visibilityDistanceType FROM creature_template c LEFT JOIN creature_template_addon ca ON c.entry=ca.entry WHERE visibilityDistanceType!=0 AND !(c.unit_flags2&(0x40000000|0x00400000|0x00200000))
```
Is it ok to normalize the value of visibilityDistanceType according to the presence or not of AOI flags and then remove these flags from the templates? The application of the flags would then be done in the function SetVisibilityDistanceOverride
```c++
    if (GetTypeId() == TYPEID_UNIT)
    {
        if (Creature* creature = ToCreature())
        {
            UnitFlags2 AOIUnitFlag = UnitFlags2(0);

            switch (type)
            {
                case VisibilityDistanceType::Large:
                    AOIUnitFlag = UNIT_FLAG2_LARGE_AOI;
                    break;
                case VisibilityDistanceType::Gigantic:
                    AOIUnitFlag = UNIT_FLAG2_GIGANTIC_AOI;
                    break;
                case VisibilityDistanceType::Infinite:
                    AOIUnitFlag = UNIT_FLAG2_INFINITE_AOI;
                    break;
                default:
                    break;
            }

            if (AOIUnitFlag && !creature->HasUnitFlag2(AOIUnitFlag))
            {
                creature->RemoveUnitFlag2(UNIT_FLAG2_LARGE_AOI | UNIT_FLAG2_GIGANTIC_AOI | UNIT_FLAG2_INFINITE_AOI);
                creature->SetUnitFlag2(AOIUnitFlag);
            }
        }
    }
```

Shauren:
No, it is not ok. Value in `visibilityDistanceType` is more correct in our existing db content, do not look at the flag in creature_template to change visibilityDistanceType

offl:
visibilityDistanceType was added from MoP's CreatureDifficulty.db2, unit flags were added from sniffs, that's fine that there's collision, you can't sniff all those creatures with AOI unit flags but CreatureDifficulty.db2 has all that data for all MoP creatures

>Value in visibilityDistanceType is more correct in our existing db content, do not look at the flag in creature_template to change visibilityDistanceType

At the same time you probably didn't added visibilityDistanceType to creatures with AOI unit_flags, so it should be fine to add missing based on unit_flags, but not removing if creature has no unit_flags
I noticed that people add manually visibilityDistanceType, to be honest no idea if that data was based on unit flags or not, nor why you didn't switched to unit_flags yet, in theory that can be done even for 335

Shauren:
New versions of WPP set visibilityDistanceType based on unit_flags2

offl:
>Is it ok to normalize the value of visibilityDistanceType according to the presence or not of AOI flags and then remove these flags from the templates?

>New versions of WPP set visibilityDistanceType based on unit_flags2

Those are static, they are set in DB(just not like we do it), and can be stored in DB currently
Why you use that field if you can just use unit_flags?

PR #28291, opened 2022-09-27
Core/GameObjects: Added check to avoid using GAMEOBJECT_TYPE_CHAIR if already in use by a creature

meji46:
**Changes proposed:**

-  Added check to avoid using GAMEOBJECT_TYPE_CHAIR if already in use by a creature. In retail, the chairs have no special flag and the client sends CMSG_GAME_OBJ_USE and CMSG_GAME_OBJ_REPORT_USE on click. The only difference with empty chairs is that the server doesn't send SMSG_MOVE_TELEPORT.

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -2225,17 +2225,18 @@ void GameObject::Use(Unit* user)
 
                 float x_i = GetPositionX() + relativeDistance * std::cos(orthogonalOrientation);
                 float y_i = GetPositionY() + relativeDistance * std::sin(orthogonalOrientation);
+                Position const slotPos(x_i, y_i, GetPositionZ());
 
                 // Check if seat is "occupied" by creature
-                if (Creature* creature = FindNearestCreatureInRange(INTERACTION_DISTANCE))
-                    if (creature->IsSitState() && creature->GetStandState() != UNIT_STAND_STATE_SIT && creature->GetExactDist2d(x_i, y_i) < 0.1f)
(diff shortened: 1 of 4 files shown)


offl:
https://github.com/TrinityCore/TrinityCore/discussions/27046#discussioncomment-1466257 interesting stuff about emotes

Nyr97:
This is not exactly correct. In WoD garrison, if there's any peon or any other NPC sitting on a chair, you're able to click on it, sit on it and the NPC will be thrown away a few yards from it.

mdX7:
> This is not exactly correct. In WoD garrison, if there's any peon or any other NPC sitting on a chair, you're able to click on it, sit on it and the NPC will be kicked and thrown away a few yards from it.

Sounds like we require a script for such cases

Shauren:
Wouldn't that just be a script on CMSG_GAME_OBJ_REPORT_USE? Sounds easy to special case a few objects

meji46:
In this case I create a generic script and associate it with go to prevent interaction with it if it has a creature "occupying" it.

I understand that I should use OnGossipHello and not OnReportUse, right?

Shauren:
No, I was talking about garrison guys, these chairs would get a script

Nyr97:
it could be the case, not sure if this is the case anywhere else, but I've seen this behaviour there 👀

Shauren:
Closing as 277c063413e4ec54ba943abd87cc1d273ad035e5 allows creatures to actually use the chair and get inserted into ChairListSlots

Issue #28302, opened 2022-10-02
DB/Silverpine: PrevQuestID, NextQuestID, and ExclusiveGroup for every quest


Nyr97:
### Description

It fixes Silverpine Forest's entire questline for PrevQuestID, NextQuestID, and ExclusiveGroup. I also added additional information for ExclusiveGroup. Note that part [1/5] was already merged.

### SQL fix

```sql
https://gist.github.com/Nyr97/9b45ab3f97f427d6149dc6bf98c1f8a4
```

### Branch

master

### TC rev. hash/commit

Trinity rev. https://github.com/TrinityCore/TrinityCore/commit/9f084718277b01936435c66696084fc049f8f8cf

PR #28275, opened 2022-09-22
Zone/Silverpine: Implemented Waiting to Exsanguinate (questId 27045)

Nyr97:
NOTE: do not attempt to merge this PR until https://github.com/TrinityCore/TrinityCore/pull/28272 and https://github.com/TrinityCore/TrinityCore/pull/28273 are merged. This one is dependent.

* scripted the RP event for the quest Waiting to Exsanguinate.
* added support for Deathstalker Rane Yorick to skip the first pathing if the player decides to click on the Armoire before she gets in the hiding spot.
* added missing spawn for Yorick in Ivar Patch when Waiting to Exsanguinate is completed.
* implemented generic spellscript for Reverse Cast Ride Vehicle (spellId 83781).

**Changes proposed:**

^

**Issues addressed:**

None.

**Tests performed:**

It builds and it was tested in-game.

**Known issues and TODO list:**

None.

PR #28303, opened 2022-10-02
Scripts/Silverpine: quest fixes and spawns for Forsaken Rear Guard

Nyr97:
Fixed anything related to quests in Forsaken Rear Guard, North Tide's Beachhead, and The Skittering Dark. 

Note: https://github.com/TrinityCore/TrinityCore/pull/28275 must be merged first in order for It's Only Poisonous if You Ingest It to work correctly as it contains a generic spellscript that was introduced in it and it is also used here. 

I also highly recommend merging https://github.com/TrinityCore/TrinityCore/issues/28302 first to avoid seeing quests available that are supposed to be taken in a later stage of the questline.

**Changes proposed:**

* Scripted Admiral Hatcher and Warlord Tork's RP event.
* Scripted "Salty" Rocka's and "Salty" Gorgar's RP events.
* Scripted Apothecary Wormcrud's and Orc Sea Pups' RP event.
* Scripted randomized Orc Sea Pups' RP events (they will perform different actions when they respawn as they do in retail).
* Scripted Forest Ettin's and Mutant Bush Chicken's RP event for It's Only Poisonous if You Ingest It (they're set as active so that they can be roaming around the entire place).
* Scripted Skitterweb Matriarch's RP when engaging in combat.
* Added missing spawns, paths, bytes1, bytes2, emotes, animKits and auras for several creatures.
* Fixed several quests:
- [x] Fixed The Warchief's Fleet (questId 27065).
- [x] Fixed Give 'em Hell! (questId 27073).
- [x] Fixed Steel Thunder (questId 27069).
- [x] Fixed Playing Dirty (questId 27082).
- [x] Fixed It's Only Poisonous if You Ingest It (questId 27088).
- [x] Fixed Skitterweb Menace (questId 27095).
- [x] Fixed Lost in the Darkness (questId 27093).
- [x] Fixed Deeper into Darkness (questId 27094).
- [x] Fixed Orcs are in Order (questId 27096).

**Issues addressed:**

None.

**Tests performed:**

None.


src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+                        if (player->GetDistance2d(me) < 20.0f && !_playerNear)
+                        {
+                            _events.ScheduleEvent(EVENT_ROCKA_CHOOSE_CONVERSATION, 1s);
+                            _events.ScheduleEvent(EVENT_ROCKA_CONVERSATION_COOLDOWN, 180s);
+
+                            _playerNear = true;
+                        }
+                        else
+                        {
+                            _playerNear = false;
```
mdX7, line 1669: setting it in `Reset()` should be enough instead of every time this event triggers

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+
+    void Reset() override
+    {
+        _events.Reset();
+        _events.ScheduleEvent(EVENT_ROCKA_CHECK_CONVERSATION, 1s);
+    }
+
+    void MoveInLineOfSight(Unit* who) override
+    {
+        if (who->GetTypeId() == TYPEID_PLAYER)
```
mdX7, line 1644: i think this won't really work if we have > 1 player. Imagine the following scenario:
- you run into range 25.0f => you become `_playerGUID` meanwhile
- i run into LoS range => i become `_playerGUID` in that case
- you run into < 20.0f range => nothing happens because i'm `_playerGUID`
- i run into < 20.0f range => convo happens

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+        for (uint8 i = 0; i < 3; i++)
+        {
+            for (Creature* seaDogOrc : _orcSeaDogList)
+                _drunkenOrcSeaDog[i] = seaDogOrc->GetGUID();
+        }
+    }
+
+private:
+    EventMap _events;
+    ObjectGuid _playerGUID;
```
mdX7, line 1983: unused

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+                _torokGUID = torok->GetGUID();
+        }
+    }
+
+private:
+    EventMap _events;
+    ObjectGuid _playerGUID;
+    ObjectGuid _torokGUID;
+    std::list<Player*> _playerList;
+    std::vector<Creature*> _seaDogList;
```
mdX7, line 2140: `_seaDogList` is only used locally, no need for member variable

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+        if (quest->GetQuestId() == QUEST_STEEL_THUNDER)
+            player->CastSpell(player, SPELL_SEA_PUP_TRIGGER, true);
+
+        if (quest->GetQuestId() == QUEST_LOST_IN_THE_DARKNESS)
+        {
+            me->GetCreatureListWithEntryInGrid(_seaDogList, NPC_ORC_SEA_DOG, 25.0f);
+
+            for (Creature* seadog : _seaDogList)
+            {
+                if (seadog->GetOwner()->ToPlayer()->IsQuestRewarded(QUEST_LOST_IN_THE_DARKNESS))
```
mdX7, line 2049: what if owner is null? (e.g. gm spawning such a creature)
what if owner is no player?

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+// 44913 - Orc Sea Dog
+struct npc_silverpine_orc_sea_dog_not_sick : public ScriptedAI
+{
+    npc_silverpine_orc_sea_dog_not_sick(Creature* creature) : ScriptedAI(creature) {}
+
+    void JustAppeared() override
+    {
+        switch (urand(EVENT_SEA_DOG_SIT, EVENT_SEA_DOG_UNEASY))
+        {
+            case EVENT_SEA_DOG_SIT:
```
mdX7, line 2170: use different prefix, these arent event ids

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+    }
+
+    void SpellHit(WorldObject* /*caster*/, SpellInfo const* spellInfo) override
+    {
+        switch (spellInfo->Id)
+        {
+            case SPELL_SEA_PUP_TRIGGER:
+                if (Player* player = ObjectAccessor::GetPlayer(*me, _playerGUID))
+                    Talk(TALK_ORC_PUP_DELIVER_CRATES, player);
+                if (Player* player = ObjectAccessor::GetPlayer(*me, _playerGUID))
```
mdX7, line 2322: you can use `player` from the if earlier, no need to check the same condition twice in a row

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+            if (apply)
+            {
+                if (_isJustSummoned)
+                    return;
+
+                if (Player* player = ObjectAccessor::GetPlayer(*me, _playerGUID))
+                    Talk(seatId + 1, player);
+            }
+            else
+                passenger->ToCreature()->DespawnOrUnsummon(3s);
```
mdX7, line 2311: what ifr passenger is no creature? (ref gm magic)

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+            default:
+                break;
+        }
+    }
+
+    void UpdateAI(uint32 diff) override
+    {
+        if (Player* player = ObjectAccessor::GetPlayer(*me, _playerGUID))
+        {
+            if (!player->HasAura(SPELL_SUMMON_ORC_SEA_PUP))
```
mdX7, line 2339: can we shovel that into a generic spellscript with `OnRemove` somehow? i kinda dont like iterating all player auras just for this every tick

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+                    break;
+                }
+
+                case EVENT_ORC_PUP_REMOVE_PROTECTION:
+                    _isJustSummoned = false;
+                    break;
+
+                case EVENT_ORC_PUP_DELIVER_CRATES:
+                    DoCastSelf(SPELL_EJECT_ALL_PASSENGERS);
+                    me->SetStandState(UNIT_STAND_STATE_DEAD);
```
mdX7, line 2371: you might wanna remove spellclick flag here since the npc despawns 3s after player was ejected

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+{
+    npc_silverpine_orc_crate(Creature* creature) : ScriptedAI(creature) { }
+
+    void IsSummonedBy(WorldObject* summoner) override
+    {
+        if (summoner->GetEntry() == NPC_ORC_SEA_PUP)
+        {
+            me->SetUnitFlag(UNIT_FLAG_UNINTERACTIBLE);
+
+            me->EnterVehicle(summoner->ToCreature());
```
mdX7, line 2402: `EnterVehicle()` expects `Unit*`, which would cover `Player*` aswell
=> use `ToUnit()`

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+};
+
+// 83840 - Despawn All Summons
+class spell_silverpine_despawn_all_summons_rear_guard : public SpellScript
+{
+    PrepareSpellScript(spell_silverpine_despawn_all_summons_rear_guard);
+
+    void HandleScriptEffect(SpellEffIndex /*effIndex*/)
+    {
+        Unit* caster = GetCaster();
```
mdX7, line 2445: caster can be null

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+        _events.ScheduleEvent(EVENT_CHECK_ETTIN, 1s);
+    }
+
+    void IsSummonedBy(WorldObject* summoner) override
+    {
+        if (Player* player = summoner->ToPlayer())
+        {
+            _playerGUID = player->GetGUID();
+
+            if (Player* player = ObjectAccessor::GetPlayer(*me, _playerGUID))
```
mdX7, line 2491: you already have access to `player` here from first condition, this condition is unnecessary

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+struct npc_silverpine_forest_ettin : public ScriptedAI
+{
+    npc_silverpine_forest_ettin(Creature* creature) : ScriptedAI(creature)
+    {
+        Initialize();
+    }
+
+    void Initialize()
+    {
+        me->setActive(true);
```
mdX7, line 2591: setActive is very very very evil, we shouldnt use that here as it keeps the grid loaded

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+    {
+        if (apply)
+        {
+            if (seatId == SEAT_ETTIN_ARM)
+            {
+                if (passenger->GetEntry() == NPC_MUTANT_BUSH_CHICKEN)
+                {
+                    _chickenGUID = passenger->GetGUID();
+
+                    _playerGUID = passenger->GetOwner()->GetGUID();
```
mdX7, line 2619: owner can be null if i summon a chicken right into the ettin as gm (= chicken is nearest)

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+                    _events.ScheduleEvent(EVENT_CHICKEN_EXPLODE + 2, 2s);
+            }
+        }
+    }
+
+    void DoAction(int32 param) override
+    {
+        switch (param)
+        {
+            case ACTION_GRAB_CHICKEN:
```
mdX7, line 2636: i would prefer a method where we pass the Creature pointer of the chicken since we can kinda bug this out if someone places its chicken closer to the ettin than the chicken who triggered this action

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+
+    TALK_WEBBEB_ORC_FREED                       = 0
+};
+
+// 44942 - Orc Sea Dog
+struct npc_silverpine_orc_sea_dog : public ScriptedAI
+{
+    npc_silverpine_orc_sea_dog(Creature* creature) : ScriptedAI(creature)
+    {
+        Initialize();
```
mdX7, line 2829: constructors are only used to initialize variables => move that `SetReactState()` into `JustAppeared()`

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+        me->SetPower(POWER_ENERGY, 100, true);
+    }
+
+    void IsSummonedBy(WorldObject* summoner) override
+    {
+        if (Player* player = summoner->ToPlayer())
+        {
+            _playerGUID = player->GetGUID();
+
+            player->KilledMonsterCredit(NPC_ORC_SEA_DOG, _playerGUID);
```
mdX7, line 2850: `_playerGUID` is unneeded here. In `KilledMonsterCredit` the guid is used to get a creature, which is never true for players

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1596,1436 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+                    for (Creature* stalker : _stalkerList)
+                        stalker->RemoveAura(SPELL_SKITTERWEB);
+                    me->SetAIAnimKitId(ANIMKIT_MATRIARCH_INTERACT);
+                    me->SetHomePosition(me->GetPosition());
+                    _events.ScheduleEvent(EVENT_MATRIARCH_AGGRO + 2, 1s);
+                    break;
+
+                case EVENT_MATRIARCH_AGGRO + 2:
+                    me->SetReactState(REACT_AGGRESSIVE);
+                    _events.ScheduleEvent(EVENT_SUMMON_SPIDERLINGS, 2s);
```
mdX7, line 2999: you could move the scheduling of `EVENT_SUMMON_SPIDERLINGS` and `EVENT_VENOM_SPLASH` into a new method and just call it here and in `JustEngagedWith` to have the scheduling of it in one place

src/server/scripts/Spells/spell_generic.cpp
```diff
@@ -5141,6 +5141,28 @@ class spell_gen_eject_passengers_3_8 : public SpellScript
...
+// 83840 - Despawn All Summons
+class spell_gen_despawn_all_summons_owned_by_caster : public SpellScript
+{
+    PrepareSpellScript(spell_gen_despawn_all_summons_owned_by_caster);
+
+    void HandleScriptEffect(SpellEffIndex /*effIndex*/)
+    {
+        Creature* target = GetHitCreature();
+
+        if (target->GetOwner() == GetCaster())
```
mdX7: caster can be null

mdX7 (changes requested):

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Neither do I but there's missing core support for many SPELL_EFFECT_SUMMON effects which are bound to a SPELL_EFFECT_APPLY_AURA as SPELL_AURA_DUMMY. 

I don't know if there's any documentation anywhere on TC about this, but as of today, retail has dynamic handling for these in which:

- if the player somehow loses the aura, either by dying or moving out of an allowed aura (e.g. spell_area), the creature should be unsummoned.
- if the creature dies, there's a serverside timer at which the creature is resummoned (not sure if by using the same auraGUID or using a new one by removing the aura and forcing the caster to cast it again).


On TC, it works this way:

- if the summon creature happens to die or despawn due to distance while the aura persists, it won't spawn ever again until the player manually uses .cast to summon it.
- if the player happens to log off and back in into the world and meets the requirements to cast the spell through spell_area, but the player also met the requirements before logging off, the creature related to the SPELL_EFFECT_SUMMON won't be summoned. This is probably due to the fact that the aura was still on the player (though it must be noted that the NPC will be spawned if the player has the aura and they cast the spell through .cast, but not through spell_area).
- if the player doesn't meet the requirements anymore, e.g. going out of the area related to the spell on spell_area or the zone id or area id bound to the spell, the NPC won't unsummon.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: I can remove it but since visibilityDistance is 3 it will be seen stuck in one place from hundred yards away until the player gets near to it to load the grid.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied, I'm removing the spell from here since it was added in the last PR.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied. I went with a more direct approach: if any player moves in the line of sight within 20 yards, it will check if the conversation is on CD; if it's not, it will start the conversation. There's no need to check constantly in an infinite loop of check the cooldown is reset after a certain time after being triggered.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied. There was a Despawn All Summons spells being cast that I missed.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Added IsCreature() check to avoid any crash.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Removed, feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
mdX7: offl found something cool in ur other pr, maybe its the case here aswell: https://github.com/TrinityCore/TrinityCore/pull/28334#discussion_r990509726

Nyr97: image
The problem is, even with these flags, the aura still persists and the creature won't be respawned.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied. I've refactored the code in which the entire event is handled by the ettin. Since the item's spell uses a SPELL_EFFECT_DUMMY as effect 1, we can set our player's guid once the ettin is hit by such spell and basically forcing him to only look for a chicken who's owner is the SpellHit's caster.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

mdX7 (approved):

PR #28307, opened 2022-10-02
Core/SpellEffects: Implement SPELL_EFFECT_SET_CHROMIE_TIME

aquadeus:
**Changes proposed:**

-  Implement effect 277 (SPELL_EFFECT_SET_CHROMIE_TIME)
- 
**Tests performed:**

Tested in-game

### DIFF
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -2806,6 +2806,8 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
 
         void SetTransportServerTime(int32 transportServerTime) { SetUpdateFieldValue(m_values.ModifyValue(&Player::m_activePlayerData).ModifyValue(&UF::ActivePlayerData::TransportServerTime), transportServerTime); }
 
+        void SetChromieTimeExpansion(uint8 chromieTimeExpansion) { SetUpdateFieldValue(m_values.ModifyValue(&Player::m_activePlayerData).ModifyValue(&UF::ActivePlayerData::UiChromieTimeExpansionID), chromieTimeExpansion); }
+
         bool IsInFriendlyArea() const;
         bool IsFriendlyArea(AreaTableEntry const* inArea) const;
 
--- a/src/server/game/Spells/Spell.h
+++ b/src/server/game/Spells/Spell.h
@@ -403,6 +403,7 @@ class TC_GAME_API Spell
         void EffectSendChatMessage();
         void EffectGrantBattlePetExperience();
         void EffectLearnTransmogIllusion();
+        void EffectSetChromieTime();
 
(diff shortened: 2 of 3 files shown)


mdX7:
i think the spelleffect is doing a bit more than just setting the updatefield

Shauren:
Is this PR going to see more development? In current state it doesn't do anything useful

funjoker:
Doesn’t this need also the implementation of contentuning replacement for chosen zone?

MaxtorCoder:
> Doesn’t this need also the implementation of contentuning replacement for chosen zone?

Yes and it needs the starter quests for each expansion (this is fairly easy to do), the content tuning however, might be a bit more complicated, iirc you also move to the Chromie Time phase, works the same way with Warmode

fluxurion:
When we select chromie time CTROptions ContentTuningConditionMask and ExpansionLevelMask updatefield does change too

MaxtorCoder:
> When we select chromie time CTROptions ContentTuningConditionMask and ExpansionLevelMask updatefield does change too

Correct, ContentTuningConditionMask is for ContentTuning (as the name suggests) and ExpansionLevelMask comes from UiChromieTimeExpansionInfo.db2

aquadeus:
oh i see is this right? 

[0] (CtrOptions) ContentTuningConditionMask: 5
[0] (CtrOptions) Field_4: 3
[0] (CtrOptions) ExpansionLevelMask: 32

MaxtorCoder:
> oh i see is this right?
> 
> [0] (CtrOptions) ContentTuningConditionMask: 5 [0] (CtrOptions) Field_4: 3 [0] (CtrOptions) ExpansionLevelMask: 32

Yes, Field_4 is unused still, but both of those come from DB2 iirc.

aquadeus:
could be that this effect only changes UiChromieTimeExpansionID and the opcode SMSG_CHROMIE_TIME_SELECT_EXPANSION_SUCCESS is the one which sends all the data from the db2, taking the value of the miscvalue selected?

MaxtorCoder:
> could be that this effect only changes UiChromieTimeExpansionID and the opcode SMSG_CHROMIE_TIME_SELECT_EXPANSION_SUCCESS is the one which sends all the data from the db2, taking the value of the miscvalue selected?

No, `SMSG_CHROMIE_TIME_SELECT_EXPANSION_SUCCESS` is an empty packet, `CMSG_CHROMIE_TIME_SELECT_EXPANSION` has the UiChromieTimeExpansionInfo ID :)

aquadeus:
> 

Ahh makes sense now, then ill go first with the opcode

aquadeus:
yes the packets are very easy, only send guid and expansion fields

MaxtorCoder:
> yes the packets are very easy, only send guid and expansion fields

Yessir!

PR #28337, opened 2022-10-05
Core/Spells: fixed Howling Blast

Ovahlord:
**Changes proposed:**

-  handle damage for non-primary targets
-  handle visuals

**Tests performed:**
- tested ingame

- [ ] the spell proc system currently does not properly support two spell casts of the same family being executed at the same time resulting in spell mods being consumed at wrong stages which in this case results in Rime only modifying either the primary or secondary spell damage

### DIFF
--- a/src/server/scripts/Spells/spell_dk.cpp
+++ b/src/server/scripts/Spells/spell_dk.cpp
@@ -52,9 +52,11 @@ enum DeathKnightSpells
     SPELL_DK_DEATH_STRIKE_OFFHAND               = 66188,
     SPELL_DK_FESTERING_WOUND                    = 194310,
     SPELL_DK_FROST                              = 137006,
+    SPELL_DK_FROST_FEVER                        = 55095,
     SPELL_DK_GLYPH_OF_FOUL_MENAGERIE            = 58642,
     SPELL_DK_GLYPH_OF_THE_GEIST                 = 58640,
     SPELL_DK_GLYPH_OF_THE_SKELETON              = 146652,
+    SPELL_DK_HOWLING_BLAST_AREA_DAMAGE          = 237680,
     SPELL_DK_MARK_OF_BLOOD_HEAL                 = 206945,
     SPELL_DK_NECROSIS_EFFECT                    = 216974,
     SPELL_DK_RAISE_DEAD_SUMMON                  = 52150,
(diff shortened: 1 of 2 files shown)


src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -635,6 +637,70 @@ class spell_dk_glyph_of_scourge_strike_script : public SpellScript
...
+    PrepareSpellScript(spell_dk_howling_blast);
+
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_DK_HOWLING_BLAST_AREA_DAMAGE, SPELL_DK_FROST_FEVER });
+    }
+
+    void FilterTargets(std::list<WorldObject*>& targets)
+    {
+        targets.remove_if([&](WorldObject const* target)
```
Shauren, line 655: what does the sniff say about SPELL_GO hit targets for 49184

Ovahlord: only the expl target

src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -635,6 +637,70 @@ class spell_dk_glyph_of_scourge_strike_script : public SpellScript
...
+        {
+            if (GetSpellInfo()->Id != SPELL_DK_HOWLING_BLAST_AREA_DAMAGE)
```
Seyden, line 657: we literally just did it like this.. there was no need for an extra target list or smth
grafik 
grafik

src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -635,6 +637,70 @@ class spell_dk_glyph_of_scourge_strike_script : public SpellScript
...
+    {
+        if (GetSpellInfo()->Id != SPELL_DK_HOWLING_BLAST_AREA_DAMAGE)
+            GetCaster()->CastSpell(GetExplTargetUnit(), SPELL_DK_HOWLING_BLAST_AREA_DAMAGE);
+    }
+
+    void HandleSpellVisual(SpellEffIndex /*effIndex*/)
+    {
+        // Because the expl. unit target may have gone away between cast and hit, we are using the dummy effect target instead.
+        for (ObjectGuid const& guid : _areaTargets)
+            if (Unit* target = ObjectAccessor::GetUnit(*GetHitUnit(), guid))
```
Nyr97, line 682: Somehow, nopch throws a warning here because ObjectAccessor is non-existent.

src/server/scripts/Spells/spell_dk.cpp
Seyden: If you‘re using a unit to cast it, even if it’s a dest spell, it can still hold a unit. Check the code you will see that it can hold both.

Seyden: > If you‘re using a unit to cast it, even if it’s a dest spell, it can still hold a unit. Check the code you will see that it can hold both.

Ok actually i gotta recheck why it did work back in legion.. it had the same effects and target types but i honestly dont understand atm why it worked. We also didn't do any custom changes.

PR #28334, opened 2022-10-05
Scripts/Silverpine: fixed Fenris Isle quests

Nyr97:
Please, do not attempt to merge this PR until https://github.com/TrinityCore/TrinityCore/pull/28275 and https://github.com/TrinityCore/TrinityCore/pull/28303 are merged since this one depends on those two.

**Changes proposed:**
* Added missing spawns and waypoints for Fenris Isle both for base phase (169) and No Escape phase (266).
* Scripted RP event for quest No Escape (questId 27099).
* Quest fixed:
- [x] Fixed Rise, Forsaken (questId 27097).
- [x] Fixed No Escape (questId 27099).

**Issues addressed:**

None.

**Tests performed:**

It builds and it was tested in-game.

**Known issues and TODO list:** (add/remove lines as needed)

None.

PR #28340, opened 2022-10-06
Scripts/Silverpine: fixed Lordaeron (questId 27098)

Nyr97:
Please, do not attempt to merge this PR until https://github.com/TrinityCore/TrinityCore/pull/28275, https://github.com/TrinityCore/TrinityCore/pull/28303, and https://github.com/TrinityCore/TrinityCore/pull/28334 are merged since this one depends on those three.

**Changes proposed:**

* Scripted RP event for quest Lordaeron (questId 27098).
* Quest fixed:

- [x] Fixed quest Lordaeron (questId 27098).

**Issues addressed:**

None.

**Tests performed:**
 
- [ ] Currently, the player gets dismounted from their horse when the vehicle uses SPELL_RIDE_SYLVANAS_HORSE to mount on Sylvanas' horse seatId 1. This issue is triggered by how this type of summonProperty (vehicle) is handled: as of today, any vehicle summoned creature will force Hardcoded Ride Vehicle. However, auras with control effect should respect its base value if higher than 10 because it references a different control effect spell, so it should be casting that one instead. TL;DR SummonProperties.db2 needs to be implemented properly.


src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1170,6 +1178,26 @@ struct npc_silverpine_sylvanas_windrunner_high_command : public ScriptedAI
...
+
+    void OnQuestAccept(Player* player, Quest const* quest) override
+    {
+        switch (quest->GetQuestId())
+        {
+            case QUEST_LORDAERON:
+                player->CastSpell(player, SPELL_SUMMON_SYLVANAS_AND_HORSE, true);
+                player->CastSpell(player, SPELL_SUMMON_FORSAKEN_WARHORSE, true);
+                player->CastSpell(player, SPELL_LORDAERON_AURA, true);
+                player->CastSpell(player, SPELL_SUMMON_LORDAERON_ACTORS, true);
```
offl, line 1190: >```cpp
>    SPELL_LORDAERON_AURA                        = 84189, // Note: NYI, after researching thoroughly, nothing seems to point what this SPELL_AURA_DUMMY should do. Maybe it is what forces the player to cast the rest of the spells.
>```

I'm almost sure SourceSpellID (spell to cast on quest accept) is a thing on retail, despite it's not sent to client because not needed by client. That is easiest way to cast spells comparing to other possible ways. But what is done here I have no idea. Scripted auras for such stuff usually has small duration, this one infinite and if you preserved orded of casts, it does something else. All those spells also are casted by player, so maybe there's serverside master spell

Btw what happens if quest is abandoned and taken again? Player casts again all those spells? Using SourceSpellID raises question how to add condition for [this](https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?npc=20130), which is much more easier case
When this aura gets removed btw?

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1624,714 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+    {
+        OnEffectHitTarget += SpellEffectFn(spell_silverpine_summon_lordaeron_actors::HandleScriptEffect, EFFECT_0, SPELL_EFFECT_SCRIPT_EFFECT);
+    }
+
+private:
+    ObjectGuid _playerGUID;
+};
+
+// 84173 - Despawn All Summons
+class spell_silverpine_despawn_all_summons_lordaeron : public SpellScript
```
offl, line 2249: Maybe create a generic spell script, there's like 100 similar spells, spell_gen_despawn_owned or something like that

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: I checked the serverside spell list that we have access to and there's no master spell related to any of these at all. I checked around 250 ids before and after these, and there's nothing. I need to review the packets being sent because right now I'm not sure if I really preserved the order for this exact spell.

If the quest is retaken, all of the spells are recast again. The aura gets removed either when the player dismounts from the horse, which causes a skip to happen, and then the given complete quest spell is cast, or after that complete quest spell is cast.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: I'll consider it, but I'm most likely taking that approach since we can control target conditions through DB.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: I just rechecked the packets and they follow this order in the same world tick:

SMSG_SPELL_START - 84128 (Summon Sylvanas + Horse)
SMSG_SPELL_START - 84164 (Summon Forsaken Warhorse)
SMSG_SPELL_START - 85406 (Detect: Quest Invis Zone 20 (Ambermill Shield)) [This one is handled by area_spell]
SMSG_SPELL_START - 84127 (Summon Lordaeron Actors)
SMSG_SPELL_START - 84189 (Lordaeron Area Aura)
SMSG_SPELL_START - 46598 (Ride Vehicle Hardcoded)
SMSG_SPELL_GO - 84189 (Lordaeron Area Aura)
SMSG_SPELL_GO - 84128 (Summon Sylvanas + Horse)
SMSG_SPELL_GO - 85406 (Detect: Quest Invis Zone 20 (Ambermill Shield))
SMSG_SPELL_GO - 84164 (Summon Forsaken Warhorse)
SMSG_SPELL_GO - 84127 (Summon Lordaeron Actors)
SMSG_SPELL_GO - 46598 (Ride Vehicle Hardcoded)

offl: It's exactly what I usually see, master spell is sent as last

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: I have added this spell as generic in an older PR which is soon getting merged.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: So are we to assume that aura spell dummy is the masterspell for everything else?

offl: >The aura gets removed either when the player dismounts from the horse, which causes a skip to happen, and then the given complete quest spell is cast, or after that complete quest spell is cast.

So it serves some kind of purpose, because of that it should be an aura with infinite duration

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1618,654 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+
+        me->SetReactState(REACT_PASSIVE);
+    }
+
+    void IsSummonedBy(WorldObject* summoner) override
+    {
+        _playerGUID = summoner->GetGUID();
+    }
+
+    void EnterEvadeMode(EvadeReason /*why*/) override { }
```
mdX7, line 1657: add a comment why you are overriding it like this, looks kinda unused on first look although its serving a special purpose

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1618,654 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+    {
+        _events.Update(diff);
+
+        while (uint32 eventId = _events.ExecuteEvent())
+        {
+            switch (eventId)
+            {
+                case EVENT_ACTIVATE_SKIP:
+                {
+                    if (Player* player = ObjectAccessor::GetPlayer(*me, _playerGUID))
```
mdX7, line 1669: just use `if (Unit* summoner = me->GetSummonerUnit()` its doing all the checks by itself, you can drop `_playerGUID` in the process

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1618,654 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+
+    WAYPOINT_ARRIVED_TO_SEPULCHER               = 44,
+
+    POINT_CLOSE_TO_ZONE                         = 1,
+
+    SEAT_WARHORSE_SYLVANAS                      = 0
+};
+
+// 45057 - Forsaken Warhorse (Sylvanas)
+struct npc_silverpine_warhorse_sylvanas_lordaeron : public ScriptedAI
```
mdX7, line 1704: looks like a script which could be moved to sai

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1618,654 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+        {
+            case SPELL_KILL_ME:
+            {
+                if (_heartstrikeCD || me->HasUnitState(UNIT_STATE_CASTING))
+                    me->HandleEmoteCommand(EMOTE_ONESHOT_NONE);
+                else
+                {
+                    me->HandleEmoteCommand(EMOTE_STATE_HOLD_BOW);
+
+                    if (caster->ToUnit()->GetEntry() == NPC_WORGEN_RENEGADE)
```
mdX7, line 1800: `ToUnit()` cast is not required; `GetEntry()` is part of `WorldObject`

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -1596,6 +1618,654 @@ struct npc_silverpine_forsaken_bat : public VehicleAI
...
+};
+
+// 44911 - Dreadguard
+struct npc_silverpine_dreadguard_lordaeron : public ScriptedAI
+{
+    npc_silverpine_dreadguard_lordaeron(Creature* creature) : ScriptedAI(creature), _done(false) { }
+
+    void Reset() override
+    {
+        _events.Reset();
```
mdX7, line 2022: also set `_done = false;` here

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied!

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied. It was all for testing purposes, it's now gone.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied though I need to access ToTempSummon() for this.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied, in 434 branch GetEntry was only part of Unit/Creature.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: Feedback applied.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: I really want to avoid moving anything related to sAI when it comes to RP events, especially because the mounting bug is still existing.

src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
Nyr97: I've done more research on the packets, apparently, the aura is first applied and cast when No Escape quest is rewarded. It is on until Lordaeron quest is rewarded, so it cannot be the master spell for all of the rest.

offl: Any way, I have no idea how to use SourceSpellID for player cast, didn't encountered something like that in 335 yet

Nyr97: SourceSpellId in master is contained in quest_template_addon, but if I remember correctly that value is taken directly from the sniff if there's any difference between the sniff and the DB and I had no update row.

PR #28344, opened 2022-10-07
Core/Spells: Summoned creatures by SPELL_EFFECT_SUMMON should inherit phaseshift of caster

mdX7:
...phaseshift of caster

**Issues addressed:**

**Tests performed:**

built, tested in game with #28275 (removed code of adding phase manually)


mdX7:
apparently already handled based by SummonPropertiesFlag in https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Object/Object.cpp#L1849
gonna investigate

PR #28345, opened 2022-10-07
Core/Spells: Take SummonProperties into account for SPELL_EFFECT_SUMMONs with default SummonTitle

mdX7:
**Tests performed**:

built, tested in game with https://github.com/TrinityCore/TrinityCore/pull/28275 (removed code of adding phase manually)

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -2011,7 +2011,9 @@ void Spell::EffectSummonType()
                             // randomize position for multiple summons
                             pos = caster->GetRandomPoint(*destTarget, radius);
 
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/SpellEffects.cpp
```diff
@@ -2011,21 +2011,15 @@ void Spell::EffectSummonType()
                             // randomize position for multiple summons
                             pos = caster->GetRandomPoint(*destTarget, radius);
 
-                        summon = caster->SummonCreature(entry, pos, summonType, Milliseconds(duration), 0, m_spellInfo->Id, privateObjectOwner);
+                        summon = unitCaster->GetMap()->SummonCreature(entry, pos, properties, duration, unitCaster, m_spellInfo->Id, 0, privateObjectOwner);
+                        summon->SetTempSummonType(summonType);
```
Shauren, line 2015: Move this after `if (!summon)`

Shauren (changes requested):

PR #28347, opened 2022-10-08
Core/Game: Write directly to UpdateData's ByteBuffer

Lordron:
**Changes proposed:**

-  Use ByteBuffer from UpdateData and write to it instead of copying to it after. 
-  
-  

**Issues addressed:**

**Tests performed:**

Build, logged in

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -166,14 +166,14 @@ void Object::RemoveFromWorld()
 
 void Object::BuildMovementUpdateBlock(UpdateData* data, uint32 flags) const
 {
-    ByteBuffer buf(500);
+    ByteBuffer& buf = data->GetBuffer();
 
     buf << uint8(UPDATETYPE_MOVEMENT);
     buf << GetPackGUID();
 
     BuildMovementUpdate(&buf, flags);
 
-    data->AddUpdateBlock(buf);
+    data->AddUpdateBlock();
 }
 
 void Object::BuildCreateUpdateBlockForPlayer(UpdateData* data, Player* target) const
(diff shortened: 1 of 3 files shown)


MaxtorCoder:
Are you sure this won't introduce any issues?

Lordron:
Yes, AddUpdateBlock used to append ByteBuffer to the end of internal m_data. Now we just use it directly removing unneeded allocation and following copying.

jackpoz:
is there any case where the buffer is supposed to be reused for different packets, like when sending similar information to different players ?

Lordron:
Nope. Look previously it worked like this: 

1. Allocate an intermediate ByteBuffer
2. Write movement\updates to it
3. Append ByteBuffer to UpdateData::m_data in AddUpdateBlock
4. Append UpdateData::m_data to WorldPacket in BuildPacket

Now

1. Take the reference to UpdateData::m_data
2. Write movement\updates to it
3. Append UpdateData::m_data to WorldPacket in BuildPacket

Internal ByteBuffer (m_data) is never reused. 1 UpdateData allocated for 1 player. It could be reused and even with that these changes won't break anything

Possible places for reusing

1. [Move outside loop and clear](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Battlegrounds/Zones/BattlegroundSA.cpp#L310)
2. [Move outside loop and clear](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Groups/Group.cpp#L514)
3. [Move outside loop and clear](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Maps/Map.cpp#L698)

PR #28369, opened 2022-10-15
BG/ArathiBasin: Make the battleground playable

Golrag:
Add some Creature & GameObject spawns to the battleground Arathi Basin

**Changes proposed:**

-  Spawn creatures & some scripts
-  Spawn gameobjects
-  Remove the honorable defender trigger NPC from BattlegroundAB.cpp

**Issues addressed:**

Closes: none

**Tests performed:**

- Builds
- Tested in game

- [ ] Not everything is scripted (yet)
- [x] Missing some spawns for capture points (horde/alliance)
- [x] Some scripts could just be done with SAI but i'm not really familiar with it
- [x] Waypoints might not be 100% blizzlike (as in how i create the path, points are from sniffs)
- [ ] It seems like the npc's bound to world state capture points are being created again instead of just despawning when capture point changes state. They should keep the same guid and just despawn/respawn (CreateObject1 is used in sniffs)
  - This is not the case for the spirit guides if we will want to move them to db spawns. They get a new guid every time. 
- [ ] Missing alliance gossip menus & vendor items
- [x] Scores don't work (defended/assaulted), achievement works tho


Shauren:
Don't use std::extent anymore, replace with `std::size(path)`

Nyr97:
Shouldn't we just throw any paths into waypoint_data and refuse to use MoveSmoothPathing from MotionMaster directly until smoothTransition gets implemented?

src/server/scripts/EasternKingdoms/ArathiBasin/arathi_basin.cpp
```diff
@@ -0,0 +1,788 @@
...
+
+private:
+    TaskScheduler _scheduler;
+};
+
+struct npc_bg_ab_lumberjack : ScriptedAI
+{
+    npc_bg_ab_lumberjack(Creature* creature) : ScriptedAI(creature) { }
+
+    enum Spells
```
Nyr97, line 719: Enumarates are normally outside of the struct itself.

src/server/scripts/EasternKingdoms/ArathiBasin/arathi_basin.cpp
```diff
@@ -0,0 +1,788 @@
...
+        { 684.04517f, 676.04100f, -14.635073f }
+    };
+};
+
+// Lumber Mill
+struct npc_bg_ab_lumberjack_wood_carrier : ScriptedAI
+{
+    enum Spells
+    {
+        SPELL_LUMBERJACKIN = 290604,
```
Nyr97, line 249: Enumarates are normally outside of the struct itself.

src/server/scripts/EasternKingdoms/ArathiBasin/arathi_basin.cpp
```diff
@@ -0,0 +1,788 @@
...
+        { 1381.3038f, 1341.4166f, -1.76017900f }
+    };
+};
+
+struct npc_bg_ab_dominic_masonwrite : ScriptedAI
+{
+    npc_bg_ab_dominic_masonwrite(Creature* creature) : ScriptedAI(creature) { }
+
+    enum Spells
+    {
```
Nyr97, line 161: Enumarates are normally outside of the struct itself.

src/server/game/Battlegrounds/Zones/BattlegroundAB.cpp
```diff
@@ -369,19 +369,6 @@ void BattlegroundAB::_NodeOccupied(uint8 node, Team team)
...
-    if (!trigger)
-        trigger = AddCreature(WORLD_TRIGGER, node+7, BG_AB_NodePositions[node], GetTeamIndexByTeamId(team));
-
-    //add bonus honor aura trigger creature when node is accupied
-    //cast bonus aura (+50% honor in 25yards)
-    //aura should only apply to players who have accupied the node, set correct faction for trigger
-    if (trigger)
-    {
-        trigger->SetFaction(team == ALLIANCE ? FACTION_ALLIANCE_GENERIC : FACTION_HORDE_GENERIC);
-        trigger->CastSpell(trigger, SPELL_HONORABLE_DEFENDER_25Y, false);
```
Nyr97, line 383: Are you sure this spell buff should be deleted?

Golrag: Yes, the different triggers are spawned in db now with auras added in creature_addon table. You won't get the buff with gm on tho

src/server/scripts/EasternKingdoms/ArathiBasin/arathi_basin.cpp
Golrag: I kinda like them scoped to the scripts they are using tho, even making some of them now just static const ones
if this is something really desired i will change it, but i would rather do that when everything else is done and its just some code style adjustments if that's okay

Golrag:
> Shouldn't we just throw any paths into waypoint_data and refuse to use MoveSmoothPath from MotionMaster directly until smoothTransition gets implemented?

Maybe. I find all the waypoint tables etc quite confusing that's why I go with this function knowing it's not really blizzlike and add some comments to it

src/server/scripts/EasternKingdoms/ArathiBasin/arathi_basin.cpp
Nyr97: I believe it is standard in TC not to include it inside the struct. @mdX7 may know better.

Nyr97:
It is quite easy to fix DB-wise: you just take your own points and add an index to them and a pathId. Then you just use MovePath(pathId).

Golrag:
That would also mean that the one merging this PR will have to adjust path id's in sql & scripts
Will do new scripts with path in db and see what it gives, when new ones are done i will move the current ones

Golrag:
anyone has alliance gossip menu & vendor data i could use? have them for 10.0 but seems like they changed stuff with gossips & npc_vendor seems filled with wrong data

Aokromes:
src/server/scripts/EasternKingdoms/ArathiBasin/arathi_basin.cpp:422:56: fatal error: member access into incomplete type 'Map'
                    if (Creature* target = me->GetMap()->GetCreature(targetGuid))

sql/updates/world/master/yyyy_mm_dd_xx_world.sql
```diff
@@ -0,0 +1,934 @@
...
+(@CGUID+175, 1, 0, 0, 1, 0, 0, NULL), -- Arathi Horse
+(@CGUID+203, 1, 0, 0, 1, 0, 0, NULL), -- Forsaken Lumberjack
+(@CGUID+206, 1, 0, 0, 1, 0, 0, NULL), -- Forsaken Lumberjack
+(@CGUID+230, 1, 0, 0, 1, 0, 0, NULL), -- Forsaken Blacksmith
+(@CGUID+242, 1, 0, 0, 1, 0, 0, NULL), -- Arathi Blacksmith
+(@CGUID+249, 1, 0, 0, 1, 0, 0, NULL); -- Forsaken Horse
+
+DELETE FROM `creature_addon` WHERE `guid` BETWEEN @CGUID+0 AND @CGUID+295;
+INSERT INTO `creature_addon` (`guid`, `path_id`, `mount`, `bytes1`, `bytes2`, `emote`, `aiAnimKit`, `movementAnimKit`, `meleeAnimKit`, `visibilityDistanceType`, `auras`) VALUES
+(@CGUID+102, 0, 0, 0, 1, 0, 0, 0, 0, 3, '285977'), -- Bound Arathi Rager - 285977 - Earth Channel Omni
```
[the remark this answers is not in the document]
Nyr97: Players are allowed to see at a further distance in instanced zones, including BGs. You can see players from Warsong Gulch's horde hub in Alliance's one for example. At some point visibility was increased, not sure if TC ever updated it as well.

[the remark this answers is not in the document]
Golrag: yes i see something about it being discussed in #22986

Golrag:
im gonna extend the scope a bit, trying to make the battleground working & playable while still having a reviewable PR.

Golrag:
I updated the PR to have a working battleground now. I applied the code for spirit guides that is also available in my other PR for it. Would rather have that one merged before this one so this PR can be cleaned up a bit (but feel free to point out some early/important mistakes)

A small summary of these last changes (except the details of the spirit guides):
- moved the handling of graveyards to conditions table. We can set conditions on graveyard for faction/team (we dont really need that team column in `graveyard_zone`. And added extra condition for the correct world state for each team.
This means that the whole code in Battleground to select closest graveyard based on occupied node can just be removed. Im pretty sure we could do that with other battlegrounds as well but i'm not gonna touch that in this PR
- a lot is being done in `ProcessEvent`, actually every interaction with a capture point goes through that function. Its just a giant switch case. Maybe if we want to avoid that we could look into EventScripts where each event has its own script.
- updated the points given per base & adjusted the tick timer
- computing the total amount of capture points by team needs a bit of a clean up, i don't like accessing the gameobjects for that, i think we could just use the events for it and keep in a variable. Else we could compute it via world states.

PR #28372, opened 2022-10-17
Core/Scripts: Implemented Zort's Protective Elixir

r4d1sh:
**Changes proposed:**

-  Implemented [Zort's Protective Elixir](https://www.wowhead.com/wotlk/item=36770). This elixir should protect against [Corrosive Spit](https://www.wowhead.com/wotlk/spell=47447) cast by [Ice Heart Jormungar Feeder](https://www.wowhead.com/wotlk/npc=26358)

**Tests performed:**

Builded and tested in-game


r4d1sh:
> did you check sniffs if the aura is actually just removed or if the entire corrosive spit spell is just being resisted?

I saw the combat log and there corrosive spit is removed every time after applying

jackpoz:
Thanks for the PR 🚀 😃

PR #28374, opened 2022-10-17
Core/DB: Allow to pass MySQL password via env

Naios:
Originally authored by @Yehonal

This commit fixes MySQL warnings about insecure console password arguments.

See https://github.com/azerothcore/azerothcore-wotlk/pull/13404 for the description.

---

The author acknowledged to license this fix under GPL so it can be cherry-picked to TC.

---
image
---

@Ovahlord This commit needs some testing, which I'm unable to do currently.

### DIFF
--- a/src/common/Utilities/StartProcess.cpp
+++ b/src/common/Utilities/StartProcess.cpp
@@ -80,7 +80,7 @@ template<typename T>
 static int CreateChildProcess(T waiter, std::string const& executable,
                               std::vector<std::string> const& argsVector,
                               std::string const& logger, std::string const& input,
-                              bool secure)
+                              bool secure, environment envVariables)
 {
     ipstream outStream;
     ipstream errStream;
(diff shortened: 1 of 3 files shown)


Shauren:
I disagree with this, passing passwords with env variables is as insecure as on command line

MYSQL_PWD has been deprecated in mysql 8.0 according to documentation here https://dev.mysql.com/doc/refman/8.0/en/environment-variables.html

Naios:
@Shauren That's up to you, if you want to accept it or not. But yes, I agree it's not an ideal solution.

acidmanifesto:
At the moment there is a issue with some users on Azerothcore side of things with windows. A fix is being looked into.
See highlight for issue myself and another person has reported. 
image

Shauren:
Then I'm going to close this. While having a warning displayed every time any update is applied might be ugly, at least that method of authentication was not deprecated

PR #28382, opened 2022-10-20
Core/Items: Dropped deprecated ItemFlag2 ITEM_FLAG2_DONT_IGNORE_BUY_PRICE

mdX7:
**Changes proposed:**

- dropped deprecated item flag (ref EnumeratedStrings.db2: `(deprecated) Don't ignore buy prices when extended cost is specified`)
- use `std::max(1, price)` if buyprice is > 0

**Tests performed:**
tested with npc entry 45490, blacksmith hammer and fishing rod now have proper pricing and aren't given away for free anymore


Shauren:
Hmm, how are they doing this one then? https://www.wowhead.com/item=186540/rarity (yes, it costs gold + extended cost on retail), no flag

PR #28392, opened 2022-10-23
Revert "MMAP: Build ADT floor just like WMO floor below liquid"

jackpoz:
Reverts TrinityCore/TrinityCore#27503

This commit actually causes issues where the terrain mesh underwater is disconnected from the terrain outside of the water, with creatures not being able to enter/exit water as they used to do.

image

### DIFF
--- a/src/tools/mmaps_generator/TerrainBuilder.cpp
+++ b/src/tools/mmaps_generator/TerrainBuilder.cpp
@@ -493,6 +493,10 @@ namespace MMAP
                             minTLevel = h;
                     }
 
+                    // terrain under the liquid?
+                    if (minLLevel > maxTLevel)
+                        useTerrain = false;
+
                     //liquid under the terrain?
                     if (minTLevel > maxLLevel)
                         useLiquid = false;


Ovahlord:
and without it ocean floor pathfinding is broken, causing ocan floor moving creatures not being able to chase you at all. so whatever you do here, it's unbreaking one thing, and breaking another one.

jackpoz:
Ocean floor creatures will get FARFROMPOLY flag and build a shortcut by default

funjoker:
what about 4.x+ case in Vashj'ir?
Lot of creatures crawling on the ground.
You can run on the ground and actually are faster than swimming with the boots.

jackpoz:
@funjoker this PR is reverting a PR from 2022/01/01 that is causing new issues than what it tried to fix, until it gets fixed properly in a separate branch.
4.x will work in the same way as it did before 2022/01/01 .

funjoker:
is the terrainmesh seperated due to the water surface mesh ?

PR #28404, opened 2022-10-25
Scripts/Silverpine: Fixed Nowhere to Run (questId 27195)

Nyr97:
**Changes proposed:**

-  Scripted RP event for Nowhere to Run (questId 27195).
-  Added missing spawns in Deep Elem Mine.
-  Added visibility conditions for spawns.

**Issues addressed:**

None.

**Tests performed:**

- Formation and MoveFollow are deadly wrong because we require a FormationMovementGenerator and FollowMovementGenerator rework (or cherrypick 434's branch one). That's the reason why the troopers' movement is so clunky. In any case, I'm keeping a register of things that I might improve and keep updated once certain new implementations get added.
- Toss Bomb (spellId 84467) isn't able to find an appropriate target even if the condition is correctly set or a wider RangeIndex is added to the spell. In any case, there's no npcTrigger or gameobject to be found to be used as entry for TARGET_DEST_NEARBY_ENTRY. The bomb is spawned after SPELL_GO is sent, so it cannot be the entry. I changed TargetA to TARGET_DEST_DB instead.
- Toss Bomb (spellId 84468) uses SPELL_EFFECT_SEND_EVENT which is unusable in TC. I changed it to serve a better purpose as SPELL_EFFECT_SUMMON.


src/server/scripts/EasternKingdoms/zone_silverpine_forest.cpp
```diff
@@ -5301,6 +5301,275 @@ struct npc_silverpine_generic_actor_fenris_keep : public ScriptedAI
...
+        std::vector<Creature*> forsakenTrooperList;
+        GetCreatureListWithEntryInGrid(forsakenTrooperList, me, NPC_VETERAN_FORSAKEN_TROOPER, 25.0f);
+
+        for (uint32 i = 0; i < forsakenTrooperList.size(); i++)
+        {
+            _forsakenTroopersGUID[i] = forsakenTrooperList[i]->GetGUID();
+
+            if (Creature* forsakenTrooper = ObjectAccessor::GetCreature(*me, _forsakenTroopersGUID[i]))
+            {
+                forsakenTrooper->GetMotionMaster()->MoveFollow(me, float(1.5f * i), float(M_PI));
```
mdX7, line 5547: no need to use ObjectAccessor here, you have access to the index already, so you can access the related forsakenTrooper by using `forsakenTrooperList[i]`

Nyr97: Feedback applied!

PR #28410, opened 2022-10-27
Core/Object: Implement FindNearestCreatureWithAura

Nyr97:
**Changes proposed:**

-  Implement FindNearestCreatureWithAura method.
-  Implement NearestCreatureEntryWithLiveStateAndAuraInObjectRangeCheck class to support the searcher.

**Issues addressed:**

None.

**Tests performed:**

Waiting for this to be built.

None.

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2075,6 +2075,15 @@ Creature* WorldObject::FindNearestCreature(uint32 entry, float range, bool alive
     return creature;
 }
 
+Creature* WorldObject::FindNearestCreatureWithAura(uint32 entry, float range, bool alive, uint32 spellId) const
+{
+    Creature* creature = nullptr;
+    Trinity::NearestCreatureEntryWithLiveStateAndAuraInObjectRangeCheck checker(*this, entry, alive, range, spellId);
+    Trinity::CreatureLastSearcher<Trinity::NearestCreatureEntryWithLiveStateAndAuraInObjectRangeCheck> searcher(this, creature, checker);
+    Cell::VisitAllObjects(this, searcher, range);
+    return creature;
+}
+
 GameObject* WorldObject::FindNearestGameObject(uint32 entry, float range, bool spawnedOnly) const
 {
     GameObject* go = nullptr;
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2075,10 +2075,10 @@ Creature* WorldObject::FindNearestCreature(uint32 entry, float range, bool alive
     return creature;
 }
 
-Creature* WorldObject::FindNearestCreatureWithAura(uint32 entry, float range, bool alive, uint32 spellId) const
+Creature* WorldObject::FindNearestCreatureWithAura(uint32 entry, uint32 spellId, float range, bool alive) const
 {
     Creature* creature = nullptr;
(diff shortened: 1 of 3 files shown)


Ovahlord:
At this point, I'd rather just add lambda predicate support and use that one instead of bloating the core with case specific selectors

jackpoz:
> At this point, I'd rather just add lambda predicate support and use that one instead of bloating the core with case specific selectors

how do you store i_range with a lambda ? passing local variables as reference ? it's not as easy as it looks, it's more than just 1 function

Ovahlord:
> > At this point, I'd rather just add lambda predicate support and use that one instead of bloating the core with case specific selectors
> 
> how do you store i_range with a lambda ? passing local variables as reference ? it's not as easy as it looks, it's more than just 1 function

so? use the default operator with its default args (entry, range etc) then filter the returned target list with lamda with either   erase_if or remove_if and then resize to 1

Warpten:
> how do you store i_range with a lambda ?
Value capture? Nothing stops you from using lambdas with grid searchers today, all that would need to be added is deduction guides.

jackpoz:
Can any of you 2 then show an example implementing the same as this PR does ? Because theories don't always work and one finds out only when doing a reality check

Warpten:
Threw together one example on the first random snippet I found in IDE (`src/server/scripts/Northrend/IcecrownCitadel/icecrown_citadel.cpp`):

```cpp
                Creature* target = nullptr;
                // Trinity::FriendlyMostHPMissingInRange u_check(me, 60.0f, 0);
                auto check = [me = me, i_hp = 0](Unit* u) mutable -> bool {
                    if (u->IsAlive()
                        && u->IsInCombat()
                        && !me->IsHostileTo(u)
                        && me->IsWithinDistInMap(u, 60.0f)
                        && u->GetMaxHealth() - u->GetHealth() > i_hp)
                    {
                        i_hp = u->GetMaxHealth() - u->GetHealth();
                        return true;
                    }

                    return false;
                };

                Trinity::CreatureLastSearcher searcher(me, target, check);
                Cell::VisitGridObjects(me, searcher, 60.0f);
```
Didn't even need to add CTAD guides to `GridNotifiers.h` since C++17 is able to do that from constructors, but if it was needed, it'd be as simple as
```cpp
    template <typename C>
    CreatureLastSearcher(WorldObject const*, Creature*&, C&) -> CreatureLastSearcher<C>;
```
Or if you want to be pedantic about it, this also works:
```cpp
 Trinity::CreatureLastSearcher<decltype(check)> searcher(me, target, check);
```
Might take this to another PR.

Nyr97:
Would you be interested in implementing it on your own knowledge and practice instead? I would close this PR in that case. It is just out of my knowledge to put into work if that's what you guys would rather go for.

Warpten:
I mean, the lambda thing does not relate to your PR at all.

Nyr97:
The only reason I wanted to add this helper is sometimes we want to look for a certain creature with a certain aura, and rather than having 2 lines involved for it after looking for the NearestCreature, set it all into one at least.

jackpoz:
Thanks @Warpten for the example, that helps a lot.

Would it make sense to approve this PR anyway ? it's not like in the past we added these helpers only if there were at least 10 cases where they are used and the lambda is just one more possible way to do it.

Warpten:
Lgtm?

jackpoz:
Thanks for the PR 🚀 😃

PR #28411, opened 2022-10-27
Core/Creatures: Never enable gravity for creatures that can only fly

meji46:
**Changes proposed:**

-  Never enable gravity for creatures that can only fly. This occurs when this expression: https://github.com/TrinityCore/TrinityCore/commit/1b8cb366b181650bd2dabc4e249eb2e0e95419b6#diff-2222cde2ff898034f0ce38bba2d5755cbdeef050c328b2d56599bc6d6664afc1R2784 returns false.
-  It also avoids problems where some flying trigger is spawned (almost) inside a wall and GetFloorZ() returns an unexpected value close to the Z of the trigger. In these cases, the core enables gravity and the creature falls clientside (the serverside position is desync). 

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2783,7 +2783,7 @@ void Creature::UpdateMovementFlags()
     bool canHover = CanHover();
     bool isInAir = (G3D::fuzzyGt(GetPositionZ(), ground + (canHover ? *m_unitData->HoverHeight : 0.0f) + GROUND_HEIGHT_TOLERANCE) || G3D::fuzzyLt(GetPositionZ(), ground - GROUND_HEIGHT_TOLERANCE)); // Can be underground too, prevent the falling
 
-    if (GetMovementTemplate().IsFlightAllowed() && isInAir && !IsFalling())
+    if (GetMovementTemplate().IsFlightAllowed() && (isInAir || !GetMovementTemplate().IsGroundAllowed()) && !IsFalling())
     {
         if (GetMovementTemplate().Flight == CreatureFlightMovementType::CanFly)
             SetCanFly(true);


Ovahlord:
that's why most triggers have the rooted movement flag which is taking care of that problem.

meji46:
> that's why most triggers have the rooted movement flag (in sniffs) which is taking care of that problem.

But as you said, most, not all. The halloween fire triggers in towns (related to headless horseman attacks) only have disable gravity movement flag (in sniffs).

This problem can be reproduced in Brill with this commit: https://github.com/TrinityCore/TrinityCore/commit/d5d57f921875c1a0f98d6f1ff09b20a75c0de907

Ovahlord:
and do these triggers even die in the first place or is there some hack doing hacky things?

meji46:
I haven't done much research on the mechanics yet but the spell to extinguish them has the following effect:
```
Effect 0: Id 2 (SPELL_EFFECT_SCHOOL_DAMAGE)
Difficulty: Id 0 (DIFFICULTY_NONE)
BasePoints = 2.00
Targets (8, 0) (TARGET_UNIT_DEST_AREA_ENTRY, NO_TARGET)
Radius (Id 15) 3.00 (Min: 0.00 Max: 3.00)
```
And the trigger itself, when active, uses a spell to spread the fire on nearby triggers with this effect:
```
Effect 1: Id 6 (SPELL_EFFECT_APPLY_AURA)
Difficulty: Id 0 (DIFFICULTY_NONE)
BasePoints = 4.00
Targets (38, 0) (TARGET_UNIT_NEARBY_ENTRY, NO_TARGET)
Aura Id 20 (SPELL_AURA_OBS_MOD_HEALTH), value = 4, misc = 0 (0), miscB = 0, amplitude = 0, periodic = 2000
```
So they will probably die. In any case, the described problem occurs without any interaction:
1. The trigger spawns
2. The movement flags update is executed
3. In this update it is assumed that the trigger is not in the air because GetFloorZ() returns an unexpected value. I think the cause is that the trigger is (almost) inside the model of a WMO
4. The disable gravity flag is automatically removed (without initiating any fall movement)
5. The trigger is still in its correct position (serverside) but the client sees the trigger falling

Ovahlord:
Ah, I remember the mechanic. It does not kill the trigger, it only handles the size of the flame based on health. I did a prototype of it. The trigger does not despawn because the fire can re-spread. Please try to find a more clear example

meji46:
> Ah, I remember the mechanic. It does not kill the trigger, it only handles the size of the flame based on health. I did a prototype of it. The trigger does not despawn or die because the fire can re-spread. Please try to find a more clear example

Excuse me @Ovahlord, I don't understand why it's necessary to look for another example. The mechanics of the trigger have nothing to do with the objective of the PR. As I indicated in my previous message, there are no trigger mechanics involved in the incorrect removal of the disable gravity flag

Ovahlord:
I'm asking for examples because I'm doubting this mechanic is suposed to exist. That's how science works. Make a thesis and offer proof to underline it.

meji46:
I think it is self-explanatory that an npc that only inhabits the air, should in no case enable gravity.

In order to provide another similar example for triggers, we would have to update quite a few triggers (creature_template_movement) to identify those that fly and aren't rooted, and then find one that is positioned in a place where GetFloorZ() returns another unexpected value and we can reproduce this same situation (quite difficult). I found the first example I offered by chance.

Perhaps you could indicate what is the alternative to the behaviour I'm proposing or a retail case that contradicts it

PR #28481, opened 2022-11-06
Pools/GameObjects: Allow traps to be pooled and not respawn instantly

Golrag:
While trying to make battleground buffs work with pooling I noticed a behavior I was not expecting. Read more on it here: https://github.com/TrinityCore/TrinityCore/discussions/28475

**Changes proposed:**

-  Remove updatepool call in gameobject::Delete.

**Issues addressed:**

Closes: none that I know of.

**Tests performed:**

- Builds
- Battleground buffs that work with pooling (given the SQL in the discussion) work like I wanted it to work

- I have no idea why that piece of code is here, the creature counter part does not seem to have that code. 
- I *might* have broken a lot of pooling with this. I am unsure about what to test and what results are expected

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -1440,7 +1440,11 @@ void GameObject::Delete()
     if (GameObjectOverride const* goOverride = GetGameObjectOverride())
         ReplaceAllFlags(GameObjectFlags(goOverride->Flags));
 
-    AddObjectToRemoveList();
+    uint32 poolid = GetGameObjectData() ? GetGameObjectData()->poolId : 0;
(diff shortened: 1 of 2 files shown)


Shauren:
The idea here is that a herb/mining node spawns somewhere else in the zone immediately, not after current one attempts to respawn (unless ALL nodes are on respawn timer)

Golrag:
I see, that complicates things a bit more. What would be a good approach to support both cases (if mine is a viable solution)?
Add a flag to the pool template "dont spawn other pool object immediately" ?

Golrag:
I update the code to just allow traps to not respawn instantly and not break all the ore stuff
Still think a flag on the pool template might be a better idea to wait for the respawn timer of the object that got consumed but not sure if thats the way to go with the current data

Shauren:
Replaced by #28754 (that one is going in)

PR #28488, opened 2022-11-06
[Scripting/Helper]: Implement FindNearestCreatureWithOptions

Seyden:
**Changes proposed:**

-  
-  
-  

**Issues addressed:**

**Tests performed:**

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2115,6 +2115,15 @@ Creature* WorldObject::FindNearestCreature(uint32 entry, float range, bool alive
     return creature;
 }
 
+Creature* WorldObject::FindNearestCreatureWithOptions(uint32 entry, float range, FindCreatureExtraArgs const& args /*= { }*/) const
+{
+    Creature* creature = nullptr;
+    Trinity::NearestCreatureEntryWithOptionsInObjectRangeCheck checker(*this, entry, range, &args);
+    Trinity::CreatureLastSearcher<Trinity::NearestCreatureEntryWithOptionsInObjectRangeCheck> searcher(this, creature, checker);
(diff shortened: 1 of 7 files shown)


mdX7:
#28410 there was a discussion about using lambda expressions for it, that would be more powerful than this approach

Seyden:
> #28410 there was a discussion about using lambda expressions for it, that would be more powerful than this approach

it would indeed be but not everyone knows all the options regarding the core and also not every scripter knows how to write lambdas. Having predefined options, makes it easy for scripters to look for what they're looking for

Nyr97:
we can add more options as time passes by though, that's the great part of this PR!

ImmortalusDev:
> > #28410 there was a discussion about using lambda expressions for it, that would be more powerful than this approach
> 
> it would indeed be but not everyone knows all the options regarding the core and also not every scripter knows how to write lambdas. Having predefined options, makes it easy for scripters to look for what they're looking for

Warpten wrote there example, so anyone can use it

Seyden:
I found a new issue regarding phasing.. CreatureLastSearcher and every GridNotifier Searcher Visit Check has a HasInPhase Check.. so i'm not quite sure why we're even checking 2 times if the object is in the certain phase:

1 time in the Visit function of the GridNotifier Searcher
1 time in the Visitor checker for each implementation which also checks for the phase

i have to implement a new gridnotifier searcher struct (without the phase check) and will then move my FindNearestCreatureWithOptions to this impl

xvwyh:
A lambda could be added as an `std::function` field to FindCreatureOptions, killing two birds with one stone: both preserving the simplicity of common checks implemented as part of FindCreatureOptions, and adding the flexibility of custom checks via lambdas. Just depends on whether TC overlords will approve of the minor extra costs that comes with `std::function`.

PR #28508, opened 2022-11-08
Core/Battlegrounds: Move AreaSpiritGuides to script & spell system

Golrag:
**Changes proposed:**

-  Move handling of spirit guides to script and spell system
-  Remove code in battlegrounds
-  Trying to keep battlefield code out of it. Haven't researched those.
- Implement Spell Effect Teleport Graveyard
- Implement Spell Effect Spirit Heal

**Issues addressed:**

Closes: none

**Tests performed:**

- Builds
- Queued for Eye of the storm, capture base & die
- Tested with my PR for Arathi Basin (haven't pushed much of the local changes, yet).

- ~There is a small hack going on with how the teleport to other graveyard is handled. Another NPC is supposed to be summoned which casts the serverside spell~


src/server/game/Battlegrounds/Battleground.cpp
```diff
@@ -1689,18 +1573,8 @@ bool Battleground::AddSpiritGuide(uint32 type, float x, float y, float z, float
     uint32 entry = (teamId == TEAM_ALLIANCE) ? BG_CREATURE_ENTRY_A_SPIRITGUIDE : BG_CREATURE_ENTRY_H_SPIRITGUIDE;
 
     if (Creature* creature = AddCreature(entry, type, x, y, z, o, teamId))
```
Nyr97, line 1575: This lane must go as well since you're creating a pointer that is not being used anywhere.

Golrag: fixed

Golrag:
personal spirit guides are done :)

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -4477,26 +4478,31 @@ void Spell::EffectSkill()
...
+    {
+        if (!playerTarget->IsInWorld())
+            return;
+
+        // skip if player does not want to live
+        if (spiritGuide && playerTarget->GetSpiritHealer() != spiritGuide->GetGUID())
+            return;
+
+        playerTarget->ResurrectPlayer(1.0f);
+        playerTarget->CastSpell(playerTarget, 6962, true);
```
Nyr97, line 4501: Magic number, replace with an enum please!

src/server/scripts/Spells/spell_generic.cpp
```diff
@@ -5225,6 +5225,31 @@ class spell_gen_despawn_all_summons_owned_by_caster : public SpellScript
     }
 };
 
+// Personal resurrections in battlegrounds
+// 156758 - Spirit Heal
+class spell_gen_spirit_heal : public AuraScript
```
Nyr97, line 5230: Maybe we could change it to spell_gen_bg_spirit_heal? Just a small suggestion, not necessarily at all, but it's more clear what spirit heal it refers to.

src/server/scripts/World/npcs_special.cpp
```diff
@@ -2287,6 +2287,82 @@ struct npc_gen_void_zone : public ScriptedAI
...
+    void OnChannelFinished(SpellInfo const* /*spell*/) override
+    {
+        ScheduleSpiritHealChannel();
+    }
+
+    void OnDespawn() override
+    {
+        // this is a hack.
+        // A NPC is summoned (Alliance/Horde Graveyard Teleporter) and this NPC casts the spell.
+        DoCastSelf(SPELL_GRAVEYARD_TELEPORT);
```
Nyr97, line 2316: Any luck finding out what NPC is being summoned or if it's a spell summoning it? It might be serverside, but we have serverside spellIds up to Legion. If the creature is summoned, it should come up in the sniff with who created it and what spell created it.

Golrag: Possibly serverside spells:
- ID - 24237 Summon Alliance Graveyard Teleporter (SERVERSIDE) > summons 26350
- ID - 46894 Summon Horde Graveyard Teleporter (SERVERSIDE) > summons 26351

would need to figure out summon properties, need to set demon creator, needs to despawn as soon as possible (despawns at same time as spirit guide iirc) and make that npc cast the spell instead.
could also just summon the npc here without the spell and make it cast to be a bit more blizzlike

src/server/game/Spells/SpellEffects.cpp
Golrag: copied it from original code place, but cant hurt to add it to an enum

src/server/scripts/Spells/spell_generic.cpp
Golrag: i haven't researched battlefields yet. Wintergrasp changed a bit on retail and now you can choose your graveyard location etc. Not sure which spells are being used there. Might be this one as well which wouldn't fit the bg prefix then

src/server/scripts/World/npcs_special.cpp
Golrag: I updated it to summon it already, but not with spells. Added comment with it

Golrag:
The PvP world quest in azure span also has a spirit guide that automatically ressurects you again. That one also teleports you after you are alive again, doesn't seem to be any (new) spells involved, but im gonna leave that one out of scope

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -4471,26 +4472,31 @@ void Spell::EffectSkill()
...
-    unitTarget->ToPlayer()->SpawnCorpseBones();
-    */
+    Creature* spiritGuide = GetCaster()->ToCreature();
+    if (Player* playerTarget = unitTarget->ToPlayer())
+    {
+        if (!playerTarget->IsInWorld())
+            return;
+
+        // skip if player does not want to live
+        if (spiritGuide && playerTarget->GetSpiritHealer() != spiritGuide->GetGUID())
```
Golrag, line 4491: I think filtering the targets here is not the way it's supposed to be. If a player refuses to be alive again, he shouldn't be hit by the spell at all

Golrag:
Is there anything left to do for this PR before could get merged? Besides deciding on good script names, I think i've handled everything

Seyden:
What happens if you click away the timer ? Is the aura removed ? Or does the aura stay on you ?

Golrag:
Clicking away the timer removes aura 2584 (Waiting for resurrect). Which then comes back to SpellAuraEffects.cpp where we set the SpiritHealer to nullptr.

Seyden:
Do we even need then that whole handling in the core ? Why setting the spirit healer guid, you can just get it via the aura anyway ?

Golrag:
The aure is not cast by the spirit guide but by the player itself, so we can't rely on that aura for it.
What is possible however to reduce complexity is to remove the whole check if the spirit healer can resurrect the player or not, and change it to the hardcoded spell id (Waiting for resurrect). There are only 2 spells with the effect spirit heal and I can't find any on retail where those two might happen near enough each other

Seyden:
> The aure is not cast by the spirit guide but by the player itself, so we can't rely on that aura for it. What is possible however to reduce complexity is to remove the whole check if the spirit healer can resurrect the player or not, and change it to the hardcoded spell id (Waiting for resurrect). There are only 2 spells with the effect spirit heal and I can't find any on retail where those two might happen near enough each other

that sounds like a good idea in my opinion.. because i dont know a case where there are 2 spirit healers, you are either in range and have the aura on and then you get revived or you dont have the aura and then you get ignored by the revive

Nyr97:
iirc, strand of the ancient had several spirit healers and you could choose to run away from the default you were sent into and walk towards the one you wanted to be revived on

Seyden:
> iirc, strand of the ancient had several spirit healers and you could choose to run away from the default you were sent into and walk towards the one you wanted to be revived on

Yeah but that was the case for all Battlegrounds, but in that case you would probably loose the aura, therefore you wouldn’t be valid for the revive, if you move to the other spirit healer and the next batch is over, the new timer will start after all the other players were revived, and then you would get revived if you stay at the side of the other spirit healer

Golrag:
im gonna keep the code with the object guid, in case we ever need distance checks to remove the debuff when the client doesn't do it

src/server/game/Handlers/BattleGroundHandler.cpp
```diff
@@ -658,11 +659,13 @@ void WorldSession::HandleAreaSpiritHealerQueryOpcode(WorldPackets::Battleground:
     if (!unit->IsSpiritService())                            // it's not spirit service
         return;
 
-    if (Battleground* bg = _player->GetBattleground())
-        sBattlegroundMgr->SendAreaSpiritHealerQueryOpcode(_player, bg, areaSpiritHealerQuery.HealerGuid);
+    if (unit->AI()->OnSpiritHealerQuery(_player))
```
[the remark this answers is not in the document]
Golrag: added a hardcoded 20.0f. I did not check if thats the exact value, but it seems rather correct given the aoe spell is also 20y. Not sure how the client does it. Might be hardcoded 20 somewhere, doubt its looking at the spells involved

Golrag:
I tried finding a solution without introducing new AI hooks and without having too many hardcoded spell ids in the core but couldn't find one. If someone has other ideas please share :)

src/server/game/AI/CreatureAI.h
```diff
@@ -236,6 +236,12 @@ class TC_GAME_API CreatureAI : public UnitAI
         static bool IsInBounds(CreatureBoundary const& boundary, Position const* who);
         bool IsInBoundary(Position const* who = nullptr) const;
 
+        // Called when a player queries the spirit healer timer information
+        // return true to prevent automatic SMSG_AREA_SPIRIT_HEALER_TIME
+        virtual bool OnSpiritHealerQuery(Player* /*player*/) { return false; }
```
mdX7, line 241: i'd suggest to move it into a new AI class instead. We could use the SpiritGuideAI you already implemented

Golrag: done, not completely sure if that was the right way to do it but it seems to work

Golrag: to go a bit further on this. I was thinking about creating more AI classes. AreaSpiritGuideAI & PersonalSpiritGuideAI and assign them in the db with AIName. Then for the one in azure span we could use a script

PR #28516, opened 2022-11-10
[Core/Quests]: Dont immediately update object visibility at PhaseShift and wait for the AI Hooks to be called in Player::RewardQuest

### LINKED ISSUE #28506: Quest gossip closing issue
### Description

Quest gossip should close immediately before a phase change not fade out with the npc. 

### Expected behaviour

After quest reward gossip should close before phase change.

It can be tested on the exile's Reach ships and possibly in other places.

### Branch

master

### TC rev. hash/commit

TrinityCore rev. badd37e9a499+

### Operating system

Windows 10 x64

### Custom changes

none

Seyden:
**Changes proposed:**

-  Change the order in which Quest Reward Hook / Phase Shift and Update Object Visiblity is called in Player::RewardQuest
-  This change will also fix implicit replace objects
-  

**Issues addressed:**

Closes #28506

**Tests performed:**

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -14960,15 +14960,69 @@ void Player::RewardQuest(Quest const* quest, LootItemType rewardType, uint32 rew
         UpdatePvPState();
     }
 
-    SendQuestUpdate(quest_id);
-
     SendQuestGiverStatusMultiple();
+
+    SendQuestUpdate(quest_id, false);
 
     //lets remove flag for delayed teleports
     SetCanDelayTeleport(false);
 
+    bool canHaveNextQuest = !quest->HasFlag(QUEST_FLAGS_AUTOCOMPLETE) ? !questGiver->IsPlayer() : true;
+    if (canHaveNextQuest)
+    {
+        switch (questGiver->GetTypeId())
+        {
+            case TYPEID_UNIT:
+            case TYPEID_PLAYER:
+            {
+                //For AutoSubmition was added plr case there as it almost same exclute AI script cases.
+                // Send next quest
+                if (Quest const* nextQuest = GetNextQuest(questGiver->GetGUID(), quest))
+                {
+                    // Only send the quest to the player if the conditions are met
+                    if (CanTakeQuest(nextQuest, false))
+                    {
+                        if (nextQuest->IsAutoAccept() && CanAddQuest(nextQuest, true))
+                            AddQuestAndCheckCompletion(nextQuest, questGiver);
+
+                        PlayerTalkClass->SendQuestGiverQuestDetails(nextQuest, questGiver->GetGUID(), true, false);
+                    }
+                }
+
+                PlayerTalkClass->ClearMenus();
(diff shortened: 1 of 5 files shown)


malcrom:
Works

src/server/game/Entities/Player/Player.cpp
```diff
@@ -14960,15 +14960,69 @@ void Player::RewardQuest(Quest const* quest, LootItemType rewardType, uint32 rew
...
-    SendQuestUpdate(quest_id);
-
     SendQuestGiverStatusMultiple();
 
+    SendQuestUpdate(quest_id, false);
+
     //lets remove flag for delayed teleports
     SetCanDelayTeleport(false);
 
+    bool canHaveNextQuest = !quest->HasFlag(QUEST_FLAGS_AUTOCOMPLETE) ? !questGiver->IsPlayer() : true;
```
Shauren, line 14970: whats up with this random condition that wasnt there before?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -14960,15 +14960,69 @@ void Player::RewardQuest(Quest const* quest, LootItemType rewardType, uint32 rew
...
+            }
+            default:
+                break;
+        }
+    }
+
     sScriptMgr->OnQuestStatusChange(this, quest_id);
     sScriptMgr->OnQuestStatusChange(this, quest, oldStatus, QUEST_STATUS_REWARDED);
+
+    UpdateObjectVisibility();
```
Shauren, line 15025: and where does this come from?

Seyden: Like you see above, I purposefully set false for SendQuestUpdate, so the phases are sent if needed but visibility is updated at a later point after the AI hooks are done

src/server/game/Entities/Player/Player.cpp
Seyden: This condition is in QuestHandlet.cpp in a different kind of form.. we have this condition because there is a assert in CanHaveNextQuest

src/server/game/Entities/Player/Player.cpp
Shauren: reread 2nd sentence in my comment, you changed "update visibility if phases changed earlier" to "update visibility always but later"

Seyden: Ah yes that’s what I do, otherwise I would have to return a Boolean in OnConditionChanged so I know if a phase changed, the question is: is it worth it ?

src/server/game/Entities/Player/Player.cpp
Shauren: Moving that code from `WorldSession::HandleQuestgiverChooseRewardOpcode` to `Player::RewardQuest` possibly changes the guid that is passed to `SendQuestGiverQuestDetails`, did you check what is the value of `packet.QuestGiverGUID` for quests with QUEST_FLAGS_AUTOCOMPLETE?

PR #28517, opened 2022-11-10
Core/Auras: Implement SPELL_AURA_MOD_ROOT_DISABLE_GRAVITY

Nyr97:
**Changes proposed:**

-  DisableGravity will be applied or removed if the creature does not have any aura of the same kind or if it's not DisableGravity by default in its MovementTemplate.

**Issues addressed:**

None.

**Tests performed:**

None.


src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -2872,6 +2872,25 @@ void AuraEffect::HandlePreventFleeing(AuraApplication const* aurApp, uint8 mode,
...
 }
 
+void AuraEffect::HandleAuraModRootAndDisableGravity(AuraApplication const* aurApp, uint8 mode, bool apply) const
+{
+    if (!(mode & AURA_EFFECT_HANDLE_REAL))
+        return;
+
+    Unit* target = aurApp->GetTarget();
+
+    target->SetControlled(apply, UNIT_STATE_ROOT);
```
mdX7, line 2882: what if there are other root effects?

PR #28519, opened 2022-11-10
Core/SAI: Add extra action param 7

aquadeus:
**Changes proposed:**

-  Add another parameter due to some npcs are using 7 random emotes, but atm you can only add 6

**Tests performed:**

Tested in-game


src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -2553,6 +2553,7 @@ SmartScriptHolder SmartScript::CreateSmartEvent(SMART_EVENT e, uint32 event_flag
     script.action.raw.param4 = action_param4;
     script.action.raw.param5 = action_param5;
     script.action.raw.param6 = action_param6;
+    script.action.raw.param6 = action_param7;
```
jackpoz, line 2556: you copied param6 instead of param7. are you sure you tested ingame this PR ?

aquadeus:
oops i changed in localhost but i forget to upload the change

mdX7:
you might also want to change `SMART_ACTION_PARAM_COUNT` then

aquadeus:
> you might also want to change `SMART_ACTION_PARAM_COUNT` then

done

Lordron:
bloating params is not a healthy way to do it. you can use timed actions for as many emotes as you want. What if we will discover an npc with 8 or more emotes?

mdX7:
> bloating params is not a healthy way to do it. you can use timed actions for as many emotes as you want. What if we will discover an npc with 8 or more emotes?

in that case we could create a table where we reference all related values under a certain id and just reference the id here, that way we'd be future proof. CMaNGOS also does it in that way iirc

Lordron:
> > bloating params is not a healthy way to do it. you can use timed actions for as many emotes as you want. What if we will discover an npc with 8 or more emotes?
> 
> in that case we could create a table where we reference all related values under a certain id and just reference the id here, that way we'd be future proof. CMaNGOS also does it in that way iirc

and dump action_param1-6 to it leaving with 1 single actionParamId.

Shauren:
I've seen up to 8 params max for blizz AI_GROUP_ACTION

Shauren:
This was not tested AT ALL, the most imporatant piece is missing - accessing db query result

PR #28544, opened 2022-11-17
Core/Player: Vertical "Say" distance

### LINKED ISSUE #26593: vertical "say" distance
**Description:**

on trinitycore it looks we don't have vertical limit for "say"

**Expected behaviour:**

retail have a limit of around 25 yards (tested on dalaran northrend landing point)

1. go dalaran with 2 players
2. fly with one
3. say something
4. notice even you fly high the say is seen

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

2acdd163fea2

**Operating system:** 

debian 10 x64

Dekz120:
**Changes proposed:**

-  Add Say/Yell/TextEmote vertical distance limit.
-  Change` Player::Say / Player::TextEmote / Player::Yell` in /src/server/game/Entities/Player/Player.cpp
-  Add a bool (required3dDist = false by default) to Trinity::MessageDistDeliverer class. This allows to check whether calling function requires unlimited vertical message distance. and now only **Say Yell and TextEmote** functions has a vertical limit, while other functions(for ex. combat/spell log) which are uses MessageDistDeliverer class have no this limitation. (Maybe they should have this limit, but this fact need to be researched)

**Issues addressed:**

Closes #26593

**Tests performed:**

Builds. Tested in game

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -6358,21 +6358,12 @@ void Player::SendMessageToSetInRange(WorldPacket const* data, float dist, bool s
     Cell::VisitWorldObjects(this, notifier, dist);
 }
 
-void Player::SendMessageToSetInRange(WorldPacket const* data, float dist, bool self, bool own_team_only) const
+void Player::SendMessageToSetInRange(WorldPacket const* data, float dist, bool self, bool own_team_only, bool required3dDist /*= false*/) const
 {
     if (self)
         SendDirectMessage(data);
 
-    Trinity::MessageDistDeliverer notifier(this, data, dist, own_team_only);
-    Cell::VisitWorldObjects(this, notifier, dist);
-}
-
-void Player::SendChatMessageToSetInRange(WorldPacket const* data, float dist, bool self, bool own_team_only) const
-{
-    if (self)
-        SendDirectMessage(data);
-
-    Trinity::MessageDistDeliverer notifier(this, data, dist, own_team_only, nullptr, true);
+    Trinity::MessageDistDeliverer notifier(this, data, dist, own_team_only, nullptr, required3dDist);
     Cell::VisitWorldObjects(this, notifier, dist);
 }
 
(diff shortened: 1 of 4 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -20812,7 +20821,7 @@ void Player::TextEmote(std::string_view text, WorldObject const* /*= nullptr*/,
 
     WorldPacket data;
     ChatHandler::BuildChatPacket(data, CHAT_MSG_EMOTE, LANG_UNIVERSAL, this, this, _text);
-    SendMessageToSetInRange(&data, sWorld->getFloatConfig(CONFIG_LISTEN_RANGE_TEXTEMOTE), true, !GetSession()->HasPermission(rbac::RBAC_PERM_TWO_SIDE_INTERACTION_CHAT));
+    SendChatMessageToSetInRange(&data, sWorld->getFloatConfig(CONFIG_LISTEN_RANGE_YELL), true, !GetSession()->HasPermission(rbac::RBAC_PERM_TWO_SIDE_INTERACTION_CHAT));
```
jackpoz, line 20824: what does this CONFIG_ change do ?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -6367,6 +6367,15 @@ void Player::SendMessageToSetInRange(WorldPacket const* data, float dist, bool s
     Cell::VisitWorldObjects(this, notifier, dist);
 }
 
+void Player::SendChatMessageToSetInRange(WorldPacket const* data, float dist, bool self, bool own_team_only) const
```
jackpoz, line 6370: maybe this method should be called like all the other overloads SendMessageToSetInRange() and have 1 additional parameter about 3d distance ?

src/server/game/Entities/Player/Player.cpp
Dekz120: My bad. I'll switch CONFIG_s in ::Yell and ::TextEmote

src/server/game/Entities/Player/Player.cpp
Dekz120: I guess. so i'll add last default bool argument to the SendMessageToSetInRange() instead a new method creation

Dekz120: just avoided to change public method and added a private new instead 
If it's ok to change public method I will

jackpoz: In general it's ok to change any public method as long as all the references in TC are updated to fix the build.
The only case where we might want to discuss is AI and script-related hooks that might break custom scripts by users, but even then it's up to users to fix their custom code.

So please feel free to change public methods.

src/server/game/Entities/Player/Player.cpp
Dekz120: Fixed in last commit

src/server/game/Entities/Player/Player.cpp
Dekz120: Fixed in last commit

jackpoz:
Please don't force push, it makes it harder to review the PR. All merge commits will be squashed anyway when we merge the PR

jackpoz:
Thanks for the PR 🚀 😃

PR #28555, opened 2022-11-19
Core/SAI: Added param PointId to SMART_ACTION_MOVE_OFFSET

meji46:
**Changes proposed:**

-  Added param1 PointId, useful for linking the movement to the event SMART_EVENT_MOVEMENTINFORM (Type: 8)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


jackpoz:
I think this PR is missing some validation on the SAI parameters, like https://github.com/meji46/TrinityCore/blob/ddc3080d31443827542cced5f3b7878f46261bbf/src/server/game/AI/SmartScripts/SmartScriptMgr.cpp#L980 . Didn't it throw warnings on startup when you tested it ?

meji46:
I probably missed it since when I opened this PR we had hundreds of SAI warnings and errors in master branch

Shauren:
You didn't have any warnings because this PR only adds a new parameter so all existing values in database were 0 for this action, not triggering "unused value" warning

meji46:
Yes, I know that, but I was using the param for Jade Forest so the warning would have been triggered in my local

jackpoz:
well, the line I linked is what needs to be modified with the structure you use as parameters

PR #28562, opened 2022-11-23
Core/Creatures: split creature_addon and creature_template_addon's byte fields

Ovahlord:
**Changes proposed:**

-  the times of just dumping raw sniff values into creature_addon tables are over at last. We now have clear and readable fields which will replace bytes1 and bytes2. Existing data will be converted and the old columns dropped afterwards.
- This will make addons much more enjoyable to work with while also blocking access to fields which should have never been set in the DB in the first place, such as pet talents/flags and shapeshift data

**Tests performed:**
- tested ingame
-

### DIFF
--- a/sql/updates/world/3.3.5/2099_99_99_00_world.sql
+++ b/sql/updates/world/3.3.5/2099_99_99_00_world.sql
@@ -23,14 +23,14 @@ UPDATE `creature_template_addon` SET `SheathState`= 0 WHERE (`bytes2` & 0x1 | 0x
 UPDATE `creature_template_addon` SET `SheathState`= 1 WHERE (`bytes2` & 0x1) != 0;
 UPDATE `creature_template_addon` SET `SheathState`= 2 WHERE (`bytes2` & 0x2) != 0;
 
-UPDATE `creature_template_addon` SET `PvPFlags`= 0x1 WHERE (`bytes2` & 0x100) != 0;
-UPDATE `creature_template_addon` SET `PvPFlags`= 0x2 WHERE (`bytes2` & 0x200) != 0;
-UPDATE `creature_template_addon` SET `PvPFlags`= 0x4 WHERE (`bytes2` & 0x400) != 0;
-UPDATE `creature_template_addon` SET `PvPFlags`= 0x8 WHERE (`bytes2` & 0x800) != 0;
-UPDATE `creature_template_addon` SET `PvPFlags`= 0x10 WHERE (`bytes2` & 0x1000) != 0;
-UPDATE `creature_template_addon` SET `PvPFlags`= 0x20 WHERE (`bytes2` & 0x2000) != 0;
-UPDATE `creature_template_addon` SET `PvPFlags`= 0x40 WHERE (`bytes2` & 0x4000) != 0;
-UPDATE `creature_template_addon` SET `PvPFlags`= 0x80 WHERE (`bytes2` & 0x8000) != 0;
+UPDATE `creature_template_addon` SET `PvPFlags`= `PvPFlags`|0x1 WHERE (`bytes2` & 0x100) != 0;
+UPDATE `creature_template_addon` SET `PvPFlags`= `PvPFlags`|0x2 WHERE (`bytes2` & 0x200) != 0;
+UPDATE `creature_template_addon` SET `PvPFlags`= `PvPFlags`|0x4 WHERE (`bytes2` & 0x400) != 0;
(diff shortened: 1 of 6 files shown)


src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -782,6 +784,30 @@ void ObjectMgr::LoadCreatureTemplateAddons()
...
+            animTier = 0;
+        }
+        creatureAddon.bytes1 |= uint32(uint32(animTier) << (UNIT_BYTES_1_OFFSET_ANIM_TIER * 8));
+
+        if (sheathState >= MAX_SHEATH_STATE)
+        {
+            TC_LOG_ERROR("sql.sql", "Creature (Entry: %u) has invalid sheath state (%u) defined in `creature_addon`. Truncated to 0", entry, standState);
+            sheathState = 0;
+        }
+        creatureAddon.bytes2 |= uint32(uint32(sheathState) << (UNIT_BYTES_2_OFFSET_SHEATH_STATE * 8));
```
Shauren, line 806: Split them in CreatureAddon structure too (they get deconstructed into single fields anyway when loading addon in Creature class)

sql/updates/world/3.3.5/2099_99_99_00_world.sql
```diff
@@ -0,0 +1,76 @@
...
+	
+UPDATE `creature_template_addon` SET `StandState`= 1 WHERE (`bytes1` & 0x1) != 0;
+UPDATE `creature_template_addon` SET `StandState`= 2 WHERE (`bytes1` & 0x2) != 0;
+UPDATE `creature_template_addon` SET `StandState`= 3 WHERE (`bytes1` & 0x3) != 0;
+UPDATE `creature_template_addon` SET `StandState`= 4 WHERE (`bytes1` & 0x4) != 0;
+UPDATE `creature_template_addon` SET `StandState`= 5 WHERE (`bytes1` & 0x5) != 0;
+UPDATE `creature_template_addon` SET `StandState`= 6 WHERE (`bytes1` & 0x6) != 0;
+UPDATE `creature_template_addon` SET `StandState`= 7 WHERE (`bytes1` & 0x7) != 0;
+UPDATE `creature_template_addon` SET `StandState`= 8 WHERE (`bytes1` & 0x8) != 0;
+UPDATE `creature_template_addon` SET `StandState`= 9 WHERE (`bytes1` & 0x9) != 0;
```
Shauren, line 15: This is broken. `1 & 9 != 0` is true

Just write those queries as
```
UPDATE `creature_template_addon` SET `StandState`=`bytes1` & 0xFF;
UPDATE `creature_template_addon` SET `AnimTier`=(`bytes1` >> 24) & 0xFF;
UPDATE `creature_template_addon` SET `SheathState`=`bytes2` & 0xFF;
UPDATE `creature_template_addon` SET `PvPFlags`=(`bytes2` >> 8) & 0xFF;
```

Shauren (changes requested):

src/server/game/Globals/ObjectMgr.cpp
Ovahlord: kk. now while looking at it. what to do with visbility flags? (untrackable etc). are these meant to be handled by the core as well?

Shauren: For now just add columns for fields we load in Creature::LoadCreaturesAddon()

meji46: I think they should all be handled by the core. At least the untrackable and invisible (https://github.com/TrinityCore/TrinityCore/pull/28448) flags I have only seen applied in the presence of auras

Ovahlord: for now I do as requested and keep the flags in a db column until the core features are in place

Warpten:
Do we have old data that contains forbidden db flags that this would potentially break?

Ovahlord:
There shouldn't because the core never used these values from DB. The columns that I've added were the ones the core was caring about. VisibilityFlags is subject for disucssions because it's not fully clear yet if these are mean to be set during core runtime or if they are indeed pre-set

Lordron:
Can we also make NULL by default for them (for new record, without touching old ones) to make server decide what value to use. So we can adjust interesting fields keeping rest up to server.

Example: Server sets SHEATH_STATE_MELEE for every creature without addon record. If i wanna add aura\change any other fields i have to explicitly set SheathState it to 1

Shauren:
Thats a feature change and belongs in a separate PR

Ovahlord:
that's why I've defaulted it to 1.

Shauren (approved):

PR #28643, opened 2023-01-01
Core/Spells: Implement SetDelay for CastSpellExtraArgs

Nyr97:
This will allow a caster to delay Spell::prepare to a given amount of time. It also allows scripters to stop relying on new custom BasicEvents to schedule their casts when required.

**Changes proposed:**

- Implement SetDelay for CastSpellExtraArgs which will delay Spell::prepare for the spell for a given amount of time in Milliseconds.

**Issues addressed:**

None.

**Tests performed:**

Experimental.

- [ ] I'm still unsure whether targets need to be validated further down the method.

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2857,7 +2857,7 @@ SpellCastResult WorldObject::CastSpell(CastSpellTargetArg const& targets, uint32
     spell->m_customArg = args.CustomArg;
 
     if (args.Delay > 0ms)
-        this->m_Events.AddEventAtOffset([this, spell, targets, &args]() { return spell->prepare(*targets.Targets, args.TriggeringAura); }, args.Delay);
+        this->m_Events.AddEventAtOffset([spell, targets, &args]() { return spell->prepare(*targets.Targets, args.TriggeringAura); }, args.Delay);
     else
         return spell->prepare(*targets.Targets, args.TriggeringAura);
 }
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2856,9 +2856,10 @@ SpellCastResult WorldObject::CastSpell(CastSpellTargetArg const& targets, uint32
 
     spell->m_customArg = args.CustomArg;
 
-    spell->m_delay = args.Delay;
-
-    return spell->prepare(*targets.Targets, args.TriggeringAura);
+    if (args.Delay > 0ms)
+        this->m_Events.AddEventAtOffset([this, spell, targets, &args]() { return spell->prepare(*targets.Targets, args.TriggeringAura); }, args.Delay);
+    else
+        return spell->prepare(*targets.Targets, args.TriggeringAura);
 }
 
 void WorldObject::SendPlaySpellVisual(WorldObject* target, uint32 spellVisualId, uint16 missReason, uint16 reflectStatus, float travelSpeed, bool speedAsTime /*= false*/)
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2856,6 +2856,8 @@ SpellCastResult WorldObject::CastSpell(CastSpellTargetArg const& targets, uint32
 
     spell->m_customArg = args.CustomArg;
 
+    spell->m_delay = args.Delay;
+
     return spell->prepare(*targets.Targets, args.TriggeringAura);
 }
 
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -555,6 +555,7 @@ m_spellValue(new SpellValue(m_spellInfo, caster)), _spellEvent(nullptr)
     m_castItemEntry = 0;
     m_castItemLevel = -1;
     m_castFlagsEx = 0;
+    m_delay = 1ms;
 
     if (IsIgnoringCooldowns())
         m_castFlagsEx |= CAST_FLAG_EX_IGNORE_COOLDOWN;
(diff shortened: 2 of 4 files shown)


jackpoz:
> * as one world tick is very unlikely to cause crashes due to null memory.

anything that has a chance of happening, will happen.

src/server/game/Spells/Spell.h
```diff
@@ -547,6 +547,7 @@ class TC_GAME_API Spell
         ObjectGuid m_originalCastId;
         bool m_fromClient;
         uint32 m_castFlagsEx;
+        uint32 m_delay;
```
jackpoz, line 550: Always initialize all fields in the constructor

src/server/game/Spells/Spell.cpp
```diff
@@ -3318,7 +3318,7 @@ SpellCastResult Spell::prepare(SpellCastTargets const& targets, AuraEffect const
 
     // create and add update event for this spell
     _spellEvent = new SpellEvent(this);
-    m_caster->m_Events.AddEvent(_spellEvent, m_caster->m_Events.CalculateTime(1ms));
+    m_caster->m_Events.AddEvent(_spellEvent, m_caster->m_Events.CalculateTime(m_delay ? Milliseconds(m_delay) : 1ms));
```
jackpoz, line 3321: We could set m_delay to 1ms in the constructor and skip the 0 check here. what happens if m_delay is set to 0 anyway ?

src/server/game/Spells/Spell.cpp
```diff
@@ -9127,6 +9127,13 @@ CastSpellExtraArgs& CastSpellExtraArgs::SetTriggeringAura(AuraEffect const* trig
     return *this;
 }
 
+CastSpellExtraArgs& CastSpellExtraArgs::SetDelay(uint32 delay)
```
jackpoz, line 9130: isn't this implemented in the header already ?

src/server/game/Spells/Spell.cpp
Nyr97: Not sure, but any CastSpell is scheduled for the next world tick.

src/server/game/Spells/Spell.h
Nyr97: Done, I also passed m_delay to be 1ms by default.

Nyr97:
I'll look for a different approach since we shouldn't be modifying that SpellEvent's 1ms delay at all. The entire prepare hook should be delayed instead.

Ovahlord:
is there actually any case where this is even needed? delays are meant for missile spells and are not meant to be meddled with so what exactly is this suposed to 'fix' or solve?

Nyr97:
It is needed in many cases from Legion forward for scripting. Currently, there is no other way to call for CastSpell to match a SpellVisual's impact based on a pure timer which is calculated in your given script and then passed to a BasicEvent (yes, many spells do not use missiles as they used to do in older expansions).

If we want to avoid scripters adding countless BasicEvent for newer content, we must have a generic way. And the best way possible is to use a BasicEvent for Spell::prepare.

src/server/game/Spells/SpellDefines.h
```diff
@@ -462,6 +463,7 @@ struct TC_GAME_API CastSpellExtraArgs
     Difficulty CastDifficulty = Difficulty(0);
     ObjectGuid OriginalCastId = ObjectGuid::Empty;
     Optional<int32> OriginalCastItemLevel;
+    Milliseconds Delay;
```
Nyr97, line 466: Not sure if I should set a default value of 0ms here or convert it to a <Optional> value. The later won't allow us to use args.Delay to be passed for the AddEventAtOffset's Milliseconds offset.

jackpoz: Optional is used when "no value" has a different meaning than "default value" (which would be 0ms in this case)

src/server/game/Entities/Object/Object.cpp
```diff
@@ -2856,7 +2856,10 @@ SpellCastResult WorldObject::CastSpell(CastSpellTargetArg const& targets, uint32
 
     spell->m_customArg = args.CustomArg;
 
-    return spell->prepare(*targets.Targets, args.TriggeringAura);
+    if (args.Delay > 0ms)
+        this->m_Events.AddEventAtOffset([spell, targets, &args]() { return spell->prepare(*targets.Targets, args.TriggeringAura); }, args.Delay);
```
Seyden, line 2860: Could you test this with a delay of 2 secs?

Shauren: This is not good, you are passing args to delayed lambda by reference (its a local variable in the function calling CastSpell), it will point to random garbage when it executes

Shauren (changes requested):

src/server/game/Entities/Object/Object.cpp
Nyr97: Yeah, @mdX7 already told me it's not a good idea to pass args on this. I'd use the m_ variable directly, the issue is TriggeringAura changes depending on the function precious to this one 🫥

Shauren: It gets worse than that, TriggeringAura might become invalid too

Nyr97: In that case, how is it possible to pass it without ever being invalid?

Warpten: Invert the logic, delay the call to CastSpell

Nyr97: Yeah, this would go ahead of my knowledge to handle properly if it went into that direction. What about a new method for WorldObject::CastSpellWithDelay which calls for CastSpell directly  through m_Event and make sure we always pass an ObjectGuid if target is an unit to be used in the lambda?

PR #28656, opened 2023-01-02
Spells/Evoker: Implement Azure Strike target limited to 2 in area

aquadeus:
**Changes proposed:**

-  Implement Azure Strike target limited to 2 in area

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_evoker.cpp
```diff
@@ -34,6 +37,22 @@ enum EvokerSpells
...
 };
 
+// 362969 - Azure Strike (blue)
+class spell_evo_azure_strike : public SpellScript
+{
+    PrepareSpellScript(spell_evo_azure_strike);
+
+    void FilterTargets(std::list<WorldObject*>& targets)
+    {
+        Trinity::Containers::RandomResize(targets, GetEffectInfo(EFFECT_0).CalcValue());
```
Shauren, line 47: This effect must always hit the unit selected for EFFECT_0

Using `GetEffectInfo(EFFECT_0).CalcValue()` isn't correct here because it will always return 2 but there is a talent that modifes this value (369909), you need to pass caster argument to CalcValue

Shauren (changes requested):

src/server/scripts/Spells/spell_evoker.cpp
aquadeus: I didnt know that spell ty. The spell hits main target and 2 (or 3) additional in area

Nyr97: wouldnt that be fixed with an if statement for a TraitId check to modify it? or does the trait automatically modify this spell effect0's basevalue?

Seyden: I would probably extend RandomResize with a keepPredicate or copy the method and write it locally in my script and extend it by a keep predicate

Nyr97: we have quite a lot of different spells that have an increased number of targets to filter when certain TraitId is learnt, so we should use the same method all across the board anyway!

mdX7: The talent is modifying it already, so passing caster to CalcValue does the trick, without even knowing the spell id

Nyr97: where do they handle what spell's effect must modify a different spell's effect exactly?

mdX7: Db2, you can see the spells changing values in SpellWork near the top of spells output

Nyr97: Thank you!

PR #28663, opened 2023-01-04
Core/Conf: Update MaxPlayerLevel 60 > 70

AzgathCore:


Jinnaix:
Thats not enough, you also want to change DEFAULT_MAX_LEVEL to make this change correct.

image

Jinnaix:
Looks good to me. (assuming level 70 for DF is correct, I have no idea :D)

PR #28761, opened 2023-01-17
Core/Spells: Howling Blast redux

Ovahlord:
So Saifi can now shut the fuck up.

**Changes proposed:**
-  fixed howling blast using the fancy std::any arg

**Tests performed:**
- builds


Shauren:
danger zone: reference overloads of any_cast throw exceptions, try casting that triggered spell with gm command

PR #28766, opened 2023-01-19
Core/Battlegrounds: Prevent end Arena when a dead player logout

Jildor:
**Changes proposed:**

-  Prevent end Arena when a dead player logout.
By example in Arena 2vs2, Team A - player 1 is alive, Team A - player 2 is dead, player 2 lose connection and login again and Arena end losing Team A.
Arena only should end by time or if no player is alive or online.
>The Fight
Players battle it out until all the members of one team are defeated or choose to leave, or until 45 minutes have elapsed.
https://wowwiki-archive.fandom.com/wiki/Arena_PvP_system

**Issues addressed:**

None

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/game/Battlegrounds/Battleground.cpp
+++ b/src/server/game/Battlegrounds/Battleground.cpp
@@ -1097,8 +1097,8 @@ void Battleground::EventPlayerLoggedOut(Player* player)
         // drop flag and handle other cleanups
         RemovePlayer(player, guid, GetPlayerTeam(guid));
 
-        // 1 player is logging out, if it is the last, then end arena!
-        if (isArena())
+        // 1 player is logging out, if it is the last alive, then end arena!
+        if (isArena() && player->IsAlive())
             if (GetAlivePlayersCountByTeam(player->GetBGTeam()) <= 1 && GetPlayersCountByTeam(GetOtherTeam(player->GetBGTeam())))
                 EndBattleground(GetOtherTeam(player->GetBGTeam()));
     }


Jildor:
> If a dead player logs out and he is the last player online, shouldn't there be no other players online and alive anyway ?

if last alive player die, arena ends, so that shouldn't happen.
> Could you write an example where the old code works different from the new code ?

following the example that I told above:
"By example in Arena 2vs2, Team A - player 1 is alive, Team A - player 2 is dead, player 2 lose connection and login again"
Before this PR: Arena end losing Team A.
After this PR: Arena continues until last player die or logout/leave battleground or arena timer end.

CraftedRO:
Nice catch Jildor

CraftedRO:
I have no players to test this but if I understand correctly before PR arena was ending if player 1 from team A was still alive and player 2 from same team (A) (already dead) disconnected meanwhile ?

Jinnaix:
I have a hard time understanding this.
Isnt the origin of the problem here? 
```cpp
if (GetAlivePlayersCountByTeam(player->GetBGTeam()) <= 1 
```
and would also be solved with
```cpp
if (GetAlivePlayersCountByTeam(player->GetBGTeam()) < 1 
```
?

jackpoz:
@Jildor now I get why I was confused :)
I thought that
```cpp
 if (GetAlivePlayersCountByTeam(player->GetBGTeam()) <= 1
```
was actually
```cpp
 if (GetPlayersCountByTeam(player->GetBGTeam()) <= 1
```
so not checking only alive players in current team but any player in current team.

Jildor:
> I have no players to test this but if I understand correctly before PR arena was ending if player 1 from team A was still alive and player 2 from same team (A) (already dead) disconnected or logged off meanwhile ?

yep


> I have a hard time understanding this. Isnt the origin of the problem here?
> 
> ```c++
> if (GetAlivePlayersCountByTeam(player->GetBGTeam()) <= 1 
> ```
> 
> and would also be solved with
> 
> ```c++
> if (GetAlivePlayersCountByTeam(player->GetBGTeam()) < 1 
> ```
> 
> ?

No, because player that is logging out enter in that check.



> @Jildor now I get why I was confused :) I thought that
> 
> ```c++
>  if (GetAlivePlayersCountByTeam(player->GetBGTeam()) <= 1
> ```
> 
> was actually
> 
> ```c++
>  if (GetPlayersCountByTeam(player->GetBGTeam()) <= 1
> ```
> 
> so not checking only alive players in current team but any player in current team.

😄 That's why I added the IsAlive() check

francosantana (approved):
looks good :D

jackpoz:
Thanks for the PR 🚀 😃

PR #28812, opened 2023-02-13
Core/Arena: Correctly update the stats of the members of each team when the week ends and the arena points are distributed.

Jildor:
**Changes proposed:**

-  Correctly update the stats of the members of each team when the week ends and the arena points are distributed.
After this PR: https://github.com/TrinityCore/TrinityCore/pull/17989
When do FinishWeek() all WeekGames and WeekWins are set to 0 (team and members)
after that, do SaveToDB() that have check:
```c++
    for (MemberList::const_iterator itr = Members.begin(); itr != Members.end(); ++itr)
    {
        // Save the effort and go
        if (itr->WeekGames == 0)
            continue;
```
so when arena points distribution is executed, always skip save member stats into DB and that's wrong, because if you restart the server, load stats from db and don't update it before (I am referring to the weekGames and the weekWins of each member)

**Issues addressed:**

None

**Tests performed:**

Tested in game.

### DIFF
--- a/src/server/game/Battlegrounds/ArenaTeam.cpp
+++ b/src/server/game/Battlegrounds/ArenaTeam.cpp
@@ -907,7 +907,7 @@ void ArenaTeam::UpdateArenaPointsHelper(std::map<uint32, uint32>& playerPoints)
     }
 }
 
-void ArenaTeam::SaveToDB()
+void ArenaTeam::SaveToDB(bool forceMemberSave)
 {
     // Save team and member stats to db
     // Called after a match has ended or when calculating arena_points
@@ -927,7 +927,7 @@ void ArenaTeam::SaveToDB()
     for (MemberList::const_iterator itr = Members.begin(); itr != Members.end(); ++itr)
     {
         // Save the effort and go
-        if (itr->WeekGames == 0)
+        if (itr->WeekGames == 0 && !forceMemberSave)
             continue;
 
         stmt = CharacterDatabase.GetPreparedStatement(CHAR_UPD_ARENA_TEAM_MEMBER);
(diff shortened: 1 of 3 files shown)


ibakula:
Hello from the original author of the reference commit/fix,

I don't have neither a client nor a test server, but the issue seems real to my eyes now, considered that when the flush occurs, the weekly games get reset to zero by the FinishWeek function call and then following the execution of SaveToDB, the members' reset stats won't be saved. 

However, when I think of it, the mentioned lines from 929 to line 931 in SaveToDB that serve as a check might be needless/superfluous because if the team was inactive, thanks to the check in FinishWeek, the function SaveToDB call will be skipped and the needless "stack up" of SQL queries will be avoided, preventing the lag spike which my original commit was intended to fix.

My proposition is thus that you try to change your commit by removing that check entirely and see if this fixes the problem of having a large query when there are a lot of inactive teams, as it should.

My mistake might have had happened by first adding that line to test if it prevented stacking SQL queries without much consideration after a while of getting back to solving the issue.

Except for that, I am (was at the time, actually) unsure whether if the real fix should be moving large update queries onto a separate thread or to manage the execution in a different way (or a combination of both) and this solution seems (seemed) better (or at least partial) for this case (when the teams are inactive) because I was also considering what would happen if the case was that all those teams actually were active (and considering such a large update) meaning that the spike would still occur. 

Sad to send my salutation to the watchlist. @jackpoz @Kittnz

Jildor:
Hello, I have preferred to keep the check and only avoid it in the weekly distribution because when an arena is played and the save is made we can avoid if members in the team who have not played any match that it does not save, the difference in those cases it doesn't matter, but I thought it was better to leave it like that than to remove it.

In any case, they will say how they prefer the PR to be and if necessary, I will change it :)

jackpoz:
@ibakula considering that it took 6 years to find this issue, I would say your original PR was fine. It's part of software development to add bugs and fix them, so no worries.

@Jildor is this PR ready to be merged ?

Jildor:
for me it's ready :)

ibakula:
> @ibakula considering that it took 6 years to find this issue, I would say your original PR was fine. It's part of software development to add bugs and fix them, so no worries.

Okay, that's pretty slick then.

jackpoz:
Thanks for the PR 🚀 😃

PR #28815, opened 2023-02-14
Core/Transports: Fixed event timestamps for edge cases

meji46:
**Changes proposed:**

-  Fixed desync in the arrival/departure event for transports with several pause nodes in the same TransportPathLeg (for example, The Zephyr's arrival/departure event for the second pause node (Orgrimmar))

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


Ovahlord:
It's not only that but also that many transports currently calculate wrong path lenghts for some unknown reason. There are ships like the Vashjir one which are completely async due to that. (Check the time field in sniffs against what the core currently produces). So I'd say that problem should be fixed first to see if the transport events are affected by that as well.

meji46:
@Shauren can you check what Ovah says?

Shauren:
Yes, he mentioned that very long time ago

PR #28816, opened 2023-02-15
Core/SAI: Implemented new source type SMART_SCRIPT_TYPE_EVENT (3)

meji46:
**Changes proposed:**

-  Added new script hook (class EventScript) to execute actions after triggering an event (GameEventSender)
-  Implemented new source type SMART_SCRIPT_TYPE_EVENT (3) (event_scripts replacement)
   -  Table event_script_names -> ScriptName: **SmartEventTrigger**
   -  New event type SMART_EVENT_SEND_EVENT_TRIGGER (87)
   -  BaseObject may be a player, gameobject or creature
   -  Event SMART_EVENT_GO_EVENT_INFORM can be replaced with this new source_type

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


Shauren:
What are all those changes that put the source in target arg if there is no proper target?

meji46:
> What are all those changes that put the source in target arg if there is no proper target?

The idea is to always have a BaseObject in SAI when the source type isn't a WorldObject (https://github.com/TrinityCore/TrinityCore/commit/7b14b9ef22d7869d3630c708c127683c4a59f210). That's why I have tried to ensure that GameEvents::Trigger always has a source and target for any situation.

- Spell::EffectSendEvent: Target NO_TARGET. The caster is:
  - The target in GameEvents::Trigger
  - The object in sScriptMgr->OnEventTrigger
  - The BaseObject in SAI
- GAMEOBJECT_TYPE_MAP_OBJ_TRANSPORT already acts as source/target for GameEvents::Trigger so I've added the same behavior to GAMEOBJECT_TYPE_TRANSPORT.
- The core sends SetDestructibleState without attackerOrHealer when initializing Wintergrasp so when it's nullptr I send the GO as source for GameEvents::Trigger

meji46:
@Shauren , do I have to change anything you asked me or is it valid?

mdX7 (approved):

PR #28835, opened 2023-03-01
Core/Handlers: Remove player from group on Faction Change if two side interaction group is not allowed.

Jildor:
**Changes proposed:**

-  Remove player from group on Faction Change if two side interaction group is not allowed.
When player do a faction change and was in party/raid, when player enter to the realm, player continue in group with opposite faction party/raid.

**Issues addressed:**

None

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -27,6 +27,7 @@
 #include "GameTime.h"
 #include "GitRevision.h"
 #include "Group.h"
+#include "GroupMgr.h"
 #include "Guild.h"
 #include "GuildMgr.h"
 #include "InstanceSaveMgr.h"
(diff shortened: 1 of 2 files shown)


src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -1881,6 +1882,17 @@ void WorldSession::HandleCharFactionOrRaceChangeCallback(std::shared_ptr<Charact
                 Player::LeaveAllArenaTeams(factionChangeInfo->Guid);
             }
 
+            if (!sWorld->getBoolConfig(CONFIG_ALLOW_TWO_SIDE_INTERACTION_GROUP))
+            {
+                CharacterDatabasePreparedStatement* stmt2 = CharacterDatabase.GetPreparedStatement(CHAR_SEL_GROUP_MEMBER);
+                stmt2->setUInt32(0, lowGuid);
+                PreparedQueryResult resultGroup = CharacterDatabase.Query(stmt2);
```
Shauren, line 1889: bad, blocking query, LEFT JOIN group id into CHAR_SEL_CHAR_RACE_OR_FACTION_CHANGE_INFOS

Shauren (changes requested):
query

Jildor:
Done, if something else needs to be changed I'll check tomorrow ;)

Shauren (approved):

PR #28845, opened 2023-03-07
Core/Creatures: Improved movement interruption while casting spells

### LINKED ISSUE #17585: [6.x] [Spells] Channeled spells interrupted on moving
**Description**: Channeled spells that are not supposed to be interrupted when moving, are interrupted when doing so.

**Current behaviour**: Channeling a spell that does NOT have AURA_INTERRUPT_FLAG_MOVE, are still being interrupted when moving.
**Expected behaviour**: Channeled spell is not interrupted when moving.

1. Cast a channeled spell, without channel interrupt flag "AURA_INTERRUPT_FLAG_MOVE", IE SpellId: 144985, Garrosh's Whirling Corruption.
2. There's a brief cast time, once the channel has started, if you start moving, the spell will be interrupted. Even though there's no flag indicating that it should be interrupted upon moving (AURA_INTERRUPT_FLAG_MOVE).

**Branch(es)**: 6.x

**TC hash/commit**:  db85bad1ff349d2e710de34d2a3dd8ae2a2a07a8

meji46:
**Changes proposed:**

-  The current implementation of Creature::IsMovementPreventedByCasting is incorrect for two reasons:
   - Any channeling will interrupt the creature's current movement if:
      - The spell has SpellAuraInterruptFlags::Moving or SpellAuraInterruptFlags::Turning (Spell::CheckMovement check). This is correct since otherwise the creature will interrupt the cast at the moment it moves (chase during combat).
      - The caster has UNIT_STATE_CASTING (always present for active channeling). This is incorrect and automatically causes any channeling to prevent the creature from moving.
    - The current logic belongs to an old version of Unit::IsMovementPreventedByCasting that was overwritten in Creature to support the SpellFocus system (Creature::SetSpellFocus).
https://github.com/TrinityCore/TrinityCore/commit/522f537048189b40a12d68583485d1de7fcbf1d2 https://github.com/TrinityCore/TrinityCore/commit/5043639c563514c079ba6eb959dd4c1c555fa494 https://github.com/TrinityCore/TrinityCore/commit/3ea46e57afe26778704647084cc6fa55a798e510 https://github.com/TrinityCore/TrinityCore/commit/5a2f0ce29e29d934c200f617fec6dad619fab9a5
-  Creature::IsMovementPreventedByCasting implementation has been modified to use Unit::IsMovementPreventedByCasting validation while maintaining the Creature::HasSpellFocus check
-  Added additional check to Unit::IsMovementPreventedByCasting() to not prevent movement during casting of (non-channeled) spells that don't have SpellInterruptFlags::Movement.
- Modified the logic of CheckMovement() to not prevent moving creatures from casting spells with movement interruption flags. In retail, creatures stop their movement to cast spells that would be interrupted if they are moving (IsMovementPreventedByCasting will stop the movement).

**Issues addressed:**
(Maybe) it could close #17585

**Tests performed:**
Builds, tested in-game.
- Channeling 131992 while the creature (out of combat) is moving: https://gyazo.com/a394ff7de2a0ceeba3a6a3a3bb7bb9ee
- Leymor encounter: https://github.com/TrinityCore/TrinityCore/commit/1baa2f4f3993e7269f72fd143bea36d1a7ffd9f8. Before the PR, if you moved making the boss have to chase you, most of its abilities weren't cast because they had SpellInterruptFlags::Movement. With PR, the boss stops to cast his spells and then continue the chase.

None

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -3375,21 +3375,10 @@ void Creature::DoNotReacquireSpellFocusTarget()
 
 bool Creature::IsMovementPreventedByCasting() const
 {
-    // first check if currently a movement allowed channel is active and we're not casting
-    if (Spell* spell = m_currentSpells[CURRENT_CHANNELED_SPELL])
-    {
-        if (spell->getState() != SPELL_STATE_FINISHED && spell->IsChannelActive())
-            if (spell->CheckMovement() != SPELL_CAST_OK)
-                return true;
-    }
+    if (!Unit::IsMovementPreventedByCasting() && !HasSpellFocus())
+        return false;
 
-    if (HasSpellFocus())
-        return true;
-
-    if (HasUnitState(UNIT_STATE_CASTING))
-        return true;
-
-    return false;
+    return true;
 }
 
 void Creature::StartPickPocketRefillTimer()
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -3025,7 +3025,8 @@ bool Unit::IsMovementPreventedByCasting() const
         return false;
 
     if (Spell* spell = m_currentSpells[CURRENT_GENERIC_SPELL])
-        if (CanCastSpellWhileMoving(spell->GetSpellInfo()))
+        if (CanCastSpellWhileMoving(spell->GetSpellInfo()) || spell->getState() == SPELL_STATE_FINISHED ||
+            !spell->m_spellInfo->InterruptFlags.HasFlag(SpellInterruptFlags::Movement))
             return false;
 
     // channeled spells during channel stage (after the initial cast timer) allow movement with a specific spell attribute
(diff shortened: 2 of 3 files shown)


Nyr97:
> Leymor encounter: https://github.com/TrinityCore/TrinityCore/commit/1baa2f4f3993e7269f72fd143bea36d1a7ffd9f8. Before the PR, if you moved making the boss have to chase you, most of its abilities weren't cast because they had SpellInterruptFlags::Movement. With PR, the boss stops to cast his spells and then continue the chase.

This is rather interesting. I remember having to force StopMoving on creatures sometimes so that the spell scheduled isn't finished or skipped entirely, probably related to SpellInterruptFlags::Movement?

mdX7:
> What about spells that have SpellAuraInterruptFlags::Turning, but also SPELL_ATTR12_UNK11 (SPELL_ATTR12_FACE_DESTINATION)?

out of scope, but thats something to consider next

mdX7 (approved):

Lordron:
Will it handle spells like 88616 where both InterruptFlags::Movement and SPELL_EFFECT_CHARGE are present?

meji46:
> Will it handle spells like 88616 where both InterruptFlags::Movement and SPELL_EFFECT_CHARGE are present?

Note that this spell is channeled (SPELL_ATTR1_IS_SELF_CHANNELLED) and therefore, only SpellChannelInterruptFlags::Moving or SpellChannelInterruptFlags::Turning are checked (it has neither)

PR #28848, opened 2023-03-08
Scripts/Generic: Implemented generic method to cast triggered spells …

mdX7:
…on open lock effect

* used for Seat of the Aspects teleporter in Valdrakken

**Tests performed:**

tested valdrakken seat of the aspects teleporters ingame

**Note:**
could also move it to zone_valdrakken or zone_thaldraszus and rename to spell_xy_seat_of_the_aspects_teleporter; didn't encounter similar gobs yet

### DIFF
--- a/sql/updates/world/master/9999_99_99_99_world.sql
+++ b/sql/updates/world/master/9999_99_99_99_world.sql
@@ -1,3 +1,3 @@
 DELETE FROM `spell_script_names` WHERE `spell_id`=387905;
 INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
-(387905, 'spell_gen_gob_lock_triggered_cast');
+(387905, 'spell_gen_open_lock_triggered_cast_gob_spell');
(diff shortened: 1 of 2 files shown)


mdX7:
We might even just implement it in core for all lock gobs having spellids assigned. Requires further research

Shauren:
Why? SPELL_EFFECT_OPEN_LOCK calls GameObject::Use, which casts the goober spell

mdX7:
> Why? SPELL_EFFECT_OPEN_LOCK calls GameObject::Use, which casts the goober spell

indeed, issue is actually that its not casting the spell due to lock effect spell is still being cast - `IsNonMeleeSpellCast` is returning true

Ref `Spell::prepare`:
```c
    // Prevent casting at cast another spell (ServerSide check)
    if (!(_triggeredCastFlags & TRIGGERED_IGNORE_CAST_IN_PROGRESS) && m_caster->ToUnit() && m_caster->ToUnit()->IsNonMeleeSpellCast(false, true, true, m_spellInfo->Id == 75) && !m_castId.IsEmpty())
    {
        SendCastResult(SPELL_FAILED_SPELL_IN_PROGRESS);
        finish(SPELL_FAILED_SPELL_IN_PROGRESS);
        return SPELL_FAILED_SPELL_IN_PROGRESS;
    }
```

PR #28861, opened 2023-03-14
Core/Skinning: Update correct skinning skill according to creature expansion

meji46:
**Changes proposed:**

-  Update correct skinning skill according to ContentTuning expansion linked to the skinned creature (if a creature doesn't have ContentTuning, no skill will be updated)
-  Drop obsolete check to prevent skinning if you don't have the required skill level
-  Simplified Player::UpdateGatherSkill to update the gathering skill level of all expansion skills (parent skill is the base skill of the profession). Updated level values (tooltip colors) with data from retail (thanks to Naddley for extinguishing wildlife)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game. 

None

### DIFF
--- a/src/server/game/Entities/Unit/Unit.h
+++ b/src/server/game/Entities/Unit/Unit.h
@@ -852,7 +852,7 @@ class TC_GAME_API Unit : public WorldObject
         bool IsTotem() const    { return (m_unitTypeMask & UNIT_MASK_TOTEM) != 0; }
         bool IsVehicle() const  { return (m_unitTypeMask & UNIT_MASK_VEHICLE) != 0; }
 
-        uint8 GetContentTuning() const { return m_unitData->ContentTuningID; }
+        int32 GetContentTuning() const { return m_unitData->ContentTuningID; }
         uint8 GetLevel() const { return uint8(m_unitData->Level); }
         uint8 GetLevelForTarget(WorldObject const* /*target*/) const override { return GetLevel(); }
         void SetLevel(uint8 lvl, bool sendUpdate = true);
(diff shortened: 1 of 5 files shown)


src/server/game/Spells/SpellEffects.cpp
```diff
@@ -3721,6 +3721,7 @@ void Spell::EffectSkinning()
     loot->FillLoot(creature->GetCreatureTemplate()->SkinLootId, LootTemplates_Skinning, player, true);
     player->SendLoot(*loot);
 
+    // Blizzard still uses this formula
```
Shauren, line 3724: rephrase comment to remove "blizzard"

src/server/game/Entities/Unit/Unit.h
```diff
@@ -852,6 +852,7 @@ class TC_GAME_API Unit : public WorldObject
         bool IsTotem() const    { return (m_unitTypeMask & UNIT_MASK_TOTEM) != 0; }
         bool IsVehicle() const  { return (m_unitTypeMask & UNIT_MASK_VEHICLE) != 0; }
 
+        uint8 GetContentTuning() const { return m_unitData->ContentTuningID; }
```
Shauren, line 855: Must return a 32bit value (preferably same sign as m_unitData->ContentTuningID)

Shauren (changes requested):

src/server/game/Entities/Unit/Unit.h
meji46: Oops, I forgot to update the return type. Fixed

mdX7 (approved):

PR #28862, opened 2023-03-16
TaskScheduler cleanup

Lordron:
**Changes proposed:**

- Pass std::chrono types by value, they are small enough to not be passed around by const& 
- Same applied to Optional<group_t>
- Sink parameter passing for TaskScheduler::task_handler_t/std::function in Dispatch instead of passing by const& -Replace EmptyCallback with nullptr and pass by rval-ref

**Issues addressed:**

**Tests performed:**

Build and tested on NPC with TaskScheduler

- [ ] 
- [ ]

### DIFF
--- a/src/common/Utilities/TaskScheduler.cpp
+++ b/src/common/Utilities/TaskScheduler.cpp
@@ -24,21 +24,21 @@ TaskScheduler& TaskScheduler::ClearValidator()
     return *this;
 }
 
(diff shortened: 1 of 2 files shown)


Shauren (changes requested):
I disagree with the `const&` to `&&` change for callbacks in TaskScheduler::Update, TaskScheduler::Dispatch and TaskContext::Dispatch because those functions do not take ownership of the passed argument (either by copy or move)

Schedule/Task::Task are fine because they do that

PR #28883, opened 2023-03-28
Core/StatSystem: Fixed missing calls to set negative attack power mods

robinsch:
**Changes proposed:**

-  Added calls to SetAttackPowerModNeg

**Issues addressed:**

All AP mods causes the client API Script_UnitAttackPower to always send posBuff > 0 because of setting the wrong unit field.

**Tests performed:**
https://i.gyazo.com/bb92befe647e29003dc03abbe445fcff.jpg

- [ ] 
- [ ]


Shauren (approved):

jackpoz:
Thanks for the PR 🚀 😃

kelno:
If I read this correctly, if the player has say +300 AP from gear and -250AP from debuff, it's still gonna show green

PR #28884, opened 2023-03-28
Change Map::Visit to not call EnsureGridLoaded if cell.NoCreate is true

Lordron:
Was discussed with @jackpoz in discord. He proposed to test & make a PR to see where discussion would take us. Maybe there were a good reason to call EnsureGridCreated after IsGridLoaded or it just was and oversight. The motivation behind this changes is to reduce locking, for some reason MSVC in debug/releasewithdebug mode showed that this was a pretty time consuming lock with thousands of active creatures moving around and constantly using Map::Visit

**Changes proposed:**

-  When Map::Visit is called with Cell::NoCreate set to true we check that grid exists by calling IsGridLoaded(GridCoord(x, y)) and still call EnsureGridLoaded which will call EnsureGridCreated and use locking. Sure that this call is redundant because IsGridLoaded tells us that grid is created (!= null) and loaded (isGridObjectDataLoaded is set to true)
- I also cleaned up Map::isGridObjectDataLoaded/setGridObjectDataLoaded functions. Almost everywhere we had NGridType or a sequence of getNGrid calls that was reduced to just 1 call

**Issues addressed:**

**Tests performed:**

Build and tested localy on a fork

- [ ] 
- [ ] 

I also would love to know why Cell structure uses bitpacking into 4-bytes field. Does anyone remember what was the motivation for this back then? Is this some optimization technic?

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -534,11 +534,11 @@ bool Map::EnsureGridLoaded(Cell const& cell)
     NGridType *grid = getNGrid(cell.GridX(), cell.GridY());
 
     ASSERT(grid != nullptr);
-    if (!isGridObjectDataLoaded(cell.GridX(), cell.GridY()))
+    if (!grid->isGridObjectDataLoaded())
     {
         TC_LOG_DEBUG("maps", "Loading grid[%u, %u] for map %u instance %u", cell.GridX(), cell.GridY(), GetId(), i_InstanceId);
 
-        setGridObjectDataLoaded(true, cell.GridX(), cell.GridY());
+        grid->setGridObjectDataLoaded(true);
 
         ObjectGridLoader loader(*grid, this, cell);
         loader.LoadN();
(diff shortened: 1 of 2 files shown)


jackpoz (approved):

jackpoz:
@Lordron we have a saying in TC that goes:
>don't touch grid

Would you be available in the future if the apocalypse happens and some grid issue/crash has to be investigated ? Or do you plan on going to Mars anytime soon ?

Lordron:
> @Lordron we have a saying in TC that goes:
> 
> > don't touch grid
> 
> Would you be available in the future if the apocalypse happens and some grid issue/crash has to be investigated ? Or do you plan on going to Mars anytime soon ?

i'm touching this code because of a reason - i notice a slowdown invoking Map::Visit caused by a lock in EnsureGridLoaded. 
The only thing this pr is doing is prevent calling EnsureGridLoaded for NoCreate = true && IsGridLoaded case. The grid existence is guarded by if (grid) check on the bottom.

I'm not saying i'm 100% correct with this pr and it should be merged right away. If i'm mistaken please specify here or leave comments in the code why it's important to call EnsureGridLoaded right after IsGridLoaded, what the point and how then NoCreate bit is working. Then maybe we can provide another API that will work only with loaded cells for things like packet broadcasting.

I'm agree with a concept - if it ain't broken - don't fix it, but i believe it's not the case if code contains a flow - a slowdown in this case.

I was playing with "update all objects no matter what" this. The slowdown was noticed when i started the server with all grid loaded. In debug this change dropped diff from 1900ms to 1400-1500, and in release from ~300 to 200-250. Profiler confirmed that slowdown was caused by a lock at this path

jackpoz:
To me the fix is a valid fix and makes sense and I don't see anything wrong with it, but reality often proves us wrong.

Merging this and let's hope things go smooth. Maybe keep an eye out in the next days/weeks for users reporting weird issues.

PR #28895, opened 2023-04-05
Core/Units: Prevent backward melee attacks

meji46:
**Changes proposed:**

-  Prevent backward melee attacks from creatures

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.
- Possess a creature (.poss)
- Attack another creature in melee
- Face the opposite direction to the attacked creature

None

### DIFF
--- a/src/server/game/AI/CoreAI/UnitAI.cpp
+++ b/src/server/game/AI/CoreAI/UnitAI.cpp
@@ -68,6 +68,10 @@ void UnitAI::DoMeleeAttackIfReady()
     if (!me->IsWithinMeleeRange(victim))
         return;
 
+    // Check that the victim is in front of the unit
+    if (!me->HasInArc(2 * float(M_PI) / 3, victim))
+        return;
+
     //Make sure our attack is ready and we aren't currently casting before checking distance
     if (me->isAttackReady())
     {


Ovahlord:
isn't there a IsInFront helper already which might be better for readability?

mdX7:
> isn't there a IsInFront helper already which might be better for readability?

indeed

```c
bool WorldObject::isInFront(WorldObject const* target, float arc) const
{
    return HasInArc(arc, target);
}
```

Shauren:
Which saves you nothing because the angle value is not equal to default value in `isInFront`, you still call a function on the same object with the same two arguments

mdX7 (approved):
tested ingame, works fine

src/server/game/AI/CoreAI/UnitAI.cpp
```diff
@@ -68,6 +68,10 @@ void UnitAI::DoMeleeAttackIfReady()
     if (!me->IsWithinMeleeRange(victim))
         return;
 
+    // Check that the victim is in front of the unit
+    if (!me->HasInArc(2 * float(M_PI) / 3, victim))
```
Nyr97, line 72: Is this 120º degrees?

mdX7: yes

PR #28902, opened 2023-04-07
Core/Spells: Fix a double-mount bug with Magic Rooster

r4d1sh:
**Changes proposed:**

Remove SPELL_ATTR0_CASTABLE_WHILE_MOUNTED attribute from Magic Rooster (65917) spell.

Usually, if you try to mount while sitting on another mount, the client will automatically send a dismount opcode. Unfortunately, the Magic Rooster has the attribute SPELL_ATTR0_CASTABLE_WHILE_MOUNTED, most likely by mistake. If you try to mount the Magic Rooster while sitting on another mount, you will have both mounts in your buff list. The only thing we can do is remove this attribute on the server. Auto-dismount will still not work, but at least the player will receive an error if they try to mount the Magic Rooster while sitting on another mount.

**Tests performed:**

Builded and tested.

### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -3477,7 +3477,7 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->_GetEffect(EFFECT_1).Effect = SPELL_EFFECT_NONE;
     });
 
-    // // Magic Rooster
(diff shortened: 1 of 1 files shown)


Shauren:
This is apparently blizzlike... https://www.wowhead.com/wotlk/item=49290/magic-rooster-egg#comments:id=929888

PR #28908, opened 2023-04-09
Core/GameObjects: Allow certain types to be sent to the client without display

meji46:
**Changes proposed:**

-  Allow certain types to be sent to the client without display (thanks to @Shauren and @mdX7 for confirming/retrieving the list)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.
- Runeforge in Acherus
image
- .gob add temp 289688 (GAMEOBJECT_TYPE_MULTI and GAMEOBJECT_TYPE_SIEGEABLE_MULTI have a visible model in the client)

None


mdX7 (approved):
move returns to next line within switch/case block, rest is fine

PR #28923, opened 2023-04-16
Spells/Priest: Update Divine Star

Nyr97:
**Changes proposed:**

-  Update Divine Star's areatrigger script to calculate its travel distance off from its effect 1 so that it always has support for its travel distance changes. Removed hardcoded values.
-  Update Divine Star's trigger spells to properly support both Divine Star for Shadow and Holy/Discipline specs. It also offers support during Shadow Covenant aura for Discipline since it uses SPELL_AURA_OVERRIDE_ACTIONBAR_SPELLS to override the spells cast.
- Prevent Divine Star's energize effect from happening if the caster does not use Insanity as PowerDisplay.

**Issues addressed:**

None.

**Tests performed:**

None.

### DIFF
--- a/src/server/scripts/Spells/spell_priest.cpp
+++ b/src/server/scripts/Spells/spell_priest.cpp
@@ -369,8 +369,7 @@ struct areatrigger_pri_divine_star : AreaTriggerAI
         if (!spellInfo)
             return;
 
-        SpellEffectInfo const& effectSpellInfo = spellInfo->GetEffect(EFFECT_1);
-        if (!effectSpellInfo.BasePoints)
+        if (spellInfo->GetEffects().size() <= EFFECT_1)
             return;
 
         Unit* caster = at->GetCaster();
@@ -380,7 +379,7 @@ struct areatrigger_pri_divine_star : AreaTriggerAI
         _casterCurrentPosition = caster->GetPosition();
 
         // Note: max. distance at which the Divine Star can travel to is EFFECT_1's BasePoints yards.
-        _divineStarXOffSet = float(effectSpellInfo.CalcValue(caster));
+        _divineStarXOffSet = float(spellInfo->GetEffect(EFFECT_1).CalcValue(caster));
 
         Position destPos = _casterCurrentPosition;
         at->MovePositionToFirstCollision(destPos, _divineStarXOffSet, 0.0f);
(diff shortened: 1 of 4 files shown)


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -448,7 +457,7 @@ struct areatrigger_pri_divine_star : AreaTriggerAI
                 returnSplinePoints.push_back(PositionToVector3(caster));
                 returnSplinePoints.push_back(PositionToVector3(caster));
 
-                at->InitSplines(returnSplinePoints, at->GetDistance(caster) / 24 * 1000);
+                at->InitSplines(returnSplinePoints, uint32(at->GetDistance(caster) / 24 * 1000));
```
jackpoz, line 460: is the 24 in this line correct ?

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -363,8 +368,8 @@ struct areatrigger_pri_divine_star : AreaTriggerAI
         {
             _casterCurrentPosition = caster->GetPosition();
 
-            // Note: max. distance at which the Divine Star can travel to is 24 yards.
-            float divineStarXOffSet = 24.0f;
+            // Note: max. distance at which the Divine Star can travel to is EFFECT_1's BasePoints yards.
+            float divineStarXOffSet = sSpellMgr->GetSpellInfo(at->GetSpellId(), DIFFICULTY_NONE)->GetEffect(EFFECT_1).CalcValue(at->GetCaster());
```
jackpoz, line 372: is there any validation needed before or after calling GetEffect(EFFECT_1) ?

src/server/scripts/Spells/spell_priest.cpp
Nyr97: oops I need to obtain the data again, thanks for pointing out.

src/server/scripts/Spells/spell_priest.cpp
Nyr97: I know I can check whether the spell exists in db2 or not. The issue is I'm not sure how to validate a spell's effect in an areatrigger spell since we normally validate spells in SpellScripts.

`        SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(at->GetSpellId(), DIFFICULTY_NONE);
        if (!spellInfo)
            return;`

Nyr97: I guess that works to validate, check it!

PR #28931, opened 2023-04-18
Core/Creatures: Drop difficulty entries

meji46:
**Changes proposed:**

-  Dropped difficulty entries and moved all difficulty dependent fields to creature_template_scaling
   -  HealthScalingExpansion, HealthModifier, ManaModifier, ArmorModifier, DamageModifier, CreatureDifficultyID
-  Added difficulty field for table creature_questitem
-  Support for fallback difficulty in case of no data for current difficulty 

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

- In the future, loot should be linked to CreatureDifficultyID
- The query caching system requires a revamp in order to cache the difficulty dependent data. I have simplified it to only cache data for DIFFICULTY_NONE


mdX7 (approved):

src/server/game/Entities/Creature/Creature.h
```diff
@@ -469,7 +469,6 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
 
         bool DisableReputationGain;
 
-        CreatureTemplate const* m_creatureInfo;                 // Can differ from sObjectMgr->GetCreatureTemplate(GetEntry()) in difficulty mode > 0
```
Shauren, line 472: why remove the cached pointer?

meji46: I assumed that it was only used to identify difficulty entry https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Creature/Creature.cpp#L507

Shauren:
Do these sql changes cover all creature templates that have difficulty_entry set?

meji46:
> Do these sql changes cover all creature templates that have difficulty_entry set?

Yes, it moves the data from the dropped columns (base entry) to all scaling records. Then, the data from each difficulty is moved to the specific scaling record (or a default one is generated if missing)
image
DifficultyId for normal mode is wrong (should be 3) but we preserve all modifiers

src/server/game/Entities/Creature/Creature.h
meji46: Fixed

mdX7 (approved):

PR #28933, opened 2023-04-19
Scripts/Trial Of The Crusader: Fix application of spell Surge of Adrenaline (Icehowl)

Jildor:
**Changes proposed:**

-  Add spell ScriptName spell_icehowl_massive_crash to all difficulties
-  Correct application of Surge of Adrenaline only in normal mode instead only in heroic mode

Seems that Keader had a little mistake here: 2362e9c79410c689afc10266b94f7e78c88bbddc 😄 

>After a while, Icehowl will go to the middle and use Massive Crash, throwing everyone to the wall, then Icehowl will emote, glaring at a random enemy, and run to the opposite side of him, preparing to use Ferocious Butt. The whole raid gains a 50% movement buff(Surge of Adrenaline)(In normal mode only), and everyone near the spot of Icehowl's target should move away. 

https://wowpedia.fandom.com/wiki/Icehowl

**Issues addressed:**

None

**Tests performed:**

tested in-game


jackpoz:
are the 3 new spells used anywhere in the script or are there some changes needed to use the correct one for each difficulty ?

Keader:
looks correct to me :)

jackpoz:
Thanks for the PR 🚀 😃

PR #28936, opened 2023-04-23
Core/Spells: Implement SpellAuraInterruptFlags2::TouchingGround

Nyr97:
**Changes proposed:**

- Implement SpellAuraInterruptFlags2::TouchingGround. If the unit is not flying (DisableGravity or Flying) or falling, you're left with touching the ground which covers both solid ground, solid water (while waterwalking), or in the water (submerged or not).

**Issues addressed:**

None.

**Tests performed:**

None.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12483,6 +12483,9 @@ bool Unit::UpdatePosition(float x, float y, float z, float orientation, bool tel
     if (isInWater)
         RemoveAurasWithInterruptFlags(SpellAuraInterruptFlags2::Swimming);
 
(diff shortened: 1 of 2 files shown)


Seyden:
Where is the water check ? Is swimming counting as flying or falling ?

Nyr97:
We don't check water because it triggers anyway if you ever hit:

1. Solid ground.
2. Solid water:
- Above water during waterwalking.
- In water but not submerged.
- In water but submerged.

Shauren:
I think this one is wrong, what about being rooted while in air?

Nyr97:
Huh, I gotta test that, I'll come around when I get a valuable sniff.

Nyr97:
Im closing this until I find more info.

PR #28938, opened 2023-04-23
Core/Spells: Implement SpellAuraInterruptFlags2::StartOfEncounter

Nyr97:
**Changes proposed:**

- Implement SpellAuraInterruptFlags2::StartOfEncounter. Yes, battlegrounds (including arenas) are considered PvP encounters. If Solo Shuffle queue is ever implemented, it will be covered by this.

**Issues addressed:**

None.

**Tests performed:**

None.

### DIFF
--- a/src/server/game/Instances/InstanceScript.cpp
+++ b/src/server/game/Instances/InstanceScript.cpp
@@ -410,7 +410,11 @@ bool InstanceScript::SetBossState(uint32 id, EncounterState state)
                     instance->DoOnPlayers([](Player* player)
                     {
                         if (player->IsAlive())
+                        {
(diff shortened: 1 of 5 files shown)


src/server/game/Instances/InstanceScript.cpp
```diff
@@ -410,7 +410,11 @@ bool InstanceScript::SetBossState(uint32 id, EncounterState state)
                     instance->DoOnPlayers([](Player* player)
                     {
                         if (player->IsAlive())
+                        {
                             Unit::ProcSkillsAndAuras(player, nullptr, PROC_FLAG_ENCOUNTER_START, PROC_FLAG_NONE, PROC_SPELL_TYPE_MASK_ALL, PROC_SPELL_PHASE_NONE, PROC_HIT_NONE, nullptr, nullptr, nullptr);
+
+                            player->RemoveAurasWithInterruptFlags(SpellAuraInterruptFlags2::StartOfEncounter);
```
Nyr97, line 416: Not sure if we should check IsAlive() for this, might as well leave it outside of the check since we don't even do that in Battleground::RemovePlayerAtLeave to remove battleground-specific auras.

PR #28939, opened 2023-04-23
Core/Spells: Implement SpellAuraInterruptFlags2::EndOfEncounter

Nyr97:
**Changes proposed:**

- Implement SpellAuraInterruptFlags2::EndOfEncounter. Yes, battlegrounds (including arenas) are considered PvP encounters. If Solo Shuffle queue is ever implemented, it will be covered by this.

**Issues addressed:**

None.

**Tests performed:**

None.

### DIFF
--- a/src/server/game/Instances/InstanceScript.cpp
+++ b/src/server/game/Instances/InstanceScript.cpp
@@ -420,7 +420,7 @@ bool InstanceScript::SetBossState(uint32 id, EncounterState state)
 
                     instance->DoOnPlayers([](Player* player)
                     {
-                        player->RemoveAurasWithInterruptFlags(SpellAuraInterruptFlags2::EndOfEncounter);
+                        player->AtEndOfEncounter();
                     });
                     break;
                 }
@@ -437,7 +437,7 @@ bool InstanceScript::SetBossState(uint32 id, EncounterState state)
 
                     instance->DoOnPlayers([](Player* player)
                     {
-                        player->RemoveAurasWithInterruptFlags(SpellAuraInterruptFlags2::EndOfEncounter);
+                        player->AtEndOfEncounter();
                     });
                     break;
                 }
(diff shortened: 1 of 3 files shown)


src/server/game/Instances/InstanceScript.cpp
```diff
@@ -415,10 +415,19 @@ bool InstanceScript::SetBossState(uint32 id, EncounterState state)
...
                 }
                 case FAIL:
+                {
                     ResetCombatResurrections();
                     SendEncounterEnd();
+
+                    instance->DoOnPlayers([](Player* player)
+                    {
+                        if (player->IsAlive())
+                            player->RemoveAurasWithInterruptFlags(SpellAuraInterruptFlags2::EndOfEncounter);
```
Nyr97, line 425: Not sure if we should check IsAlive() for this, might as well leave it outside of the check since we don't even do that in Battleground::RemovePlayerAtLeave to remove battleground-specific auras.

src/server/game/Instances/InstanceScript.cpp
```diff
@@ -427,7 +436,14 @@ bool InstanceScript::SetBossState(uint32 id, EncounterState state)
                         DoUpdateCriteria(CriteriaType::DefeatDungeonEncounter, dungeonEncounter->ID);
                         SendBossKillCredit(dungeonEncounter->ID);
                     }
+
+                    instance->DoOnPlayers([](Player* player)
+                    {
+                        if (player->IsAlive())
+                            player->RemoveAurasWithInterruptFlags(SpellAuraInterruptFlags2::EndOfEncounter);
```
Nyr97, line 443: Not sure if we should check IsAlive() for this, might as well leave it outside of the check since we don't even do that in Battleground::RemovePlayerAtLeave to remove battleground-specific auras.

PR #28942, opened 2023-04-23
Core/Spells: Implement SpellAuraInterruptFlags2::EnteringInstance

Nyr97:
**Changes proposed:**

- Implement SpellAuraInterruptFlags2::EnteringInstance. Instanceable returns true if MAP_INSTANCE, MAP_RAID, MAP_BATTLEGROUND, MAP_ARENA and MAP_SCENARIO.
- EnteringInstance either implies:
1. there's some sort of middle ground where the transfer is in process, but not completed yet. Map::AddPlayerToMap handles it post-transfer, so it should be fine either way. In any case, let me know if there's a better handler for it.
2. or they decided to use -ing form to refer to the gerund as in "EnterInstance" and it should just happen when you're added to the map itself.

**Issues addressed:**

None.

**Tests performed:**

None.

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -457,13 +457,11 @@ bool Map::AddPlayerToMap(Player* player, bool initPlayer /*= true*/)
     player->UpdateObjectVisibility(false);
     PhasingHandler::SendToPlayer(player);
 
-    if (player->IsAlive())
-    {
-        ConvertCorpseToBones(player->GetGUID());
+    if (Instanceable())
+        player->RemoveAurasWithInterruptFlags(SpellAuraInterruptFlags2::EnteringInstance);
 
-        if (Instanceable())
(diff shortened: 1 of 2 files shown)


src/server/game/Maps/Map.cpp
```diff
@@ -458,8 +458,13 @@ bool Map::AddPlayerToMap(Player* player, bool initPlayer /*= true*/)
     PhasingHandler::SendToPlayer(player);
 
     if (player->IsAlive())
+    {
         ConvertCorpseToBones(player->GetGUID());
 
+        if (Instanceable())
+            player->RemoveAurasWithInterruptFlags(SpellAuraInterruptFlags2::EnteringInstance);
```
Nyr97, line 465: Not sure if we should check IsAlive() for this, might as well leave it outside of the check since we don't even do that in Battleground::RemovePlayerAtLeave to remove battleground-specific auras.

PR #28944, opened 2023-04-23
Core/Spells: Implement SpellAuraInterruptFlags2::AbandonVehicle

Nyr97:
**Changes proposed:**

- Implement SpellAuraInterruptFlags2::AbandonVehicle. Added on both Unit::ExitVehicle and Unit::_ExitVehicle since the later uses special handling for AuraEffect::HandleAuraControlVehicle.

**Issues addressed:**

None.

**Tests performed:**

None.


Shauren:
Removed from `Unit::ExitVehicle`, removing the aura type causes `Unit::_ExitVehicle` to be called anyway

PR #28947, opened 2023-04-27
DB/Creatures: detach non-wdb data from creature_template part 1: difficulty entries

Ovahlord:
**Changes proposed:**
this marks the beginning of a long overdue task: relocating data which isn't sent in query packets for creatures into separate tables. This means that when this journey is over, we can call creature_template sacred so no user may touch the table ever again just like we do with gameobject_template.

**Tests performed:**
- none yet.


Shauren:
I actually disagree with tiny splits like what you propose here for 1:1 relations, it makes dealing with tables very annoying

Ovahlord:
Well, i'm currently thinking about migrating difficulty into creature_template_addon to get closer to what gameobject_template does  and only create a new table for stats instead

CraftedRO:
why don't you port for example other stuffs like creature formations ?

Nyr97:
isn't it best to move it into creature_template_difficulty instead? _addon table should resemble as much as master's, and master's difficulties are handled in _scaling table which is non-existent in wotlk branches

funjoker:
IMO creature_template should only contain data from QUERY_RESPONSE and any other data taken from UFs or else should go to addon

PR #28965, opened 2023-05-09
Core/Graveyards: Drop Faction column and replace it with conditions

Golrag:
**Changes proposed:**

- Drop `Faction` column from `graveyard_zone` and replace it with `conditions`
- Push outdoor pvp graveyards to new conditions

**Issues addressed:**

Closes: none

**Tests performed:**

- Builds
- Tested conditions with my arathi PR

- [ ] SQL could be improved
- [x] I noticed some graveyards have both a faction requirement for a zone and then no faction requirement in a different zone (same graveyard). This needs a better handling (maybe the conditions need to be changed to support graveyards by zone) or this is wrong data. I could use some input on this


acidmanifesto (changes requested):
```
[ 89%] Building CXX object src/server/scripts/CMakeFiles/scripts.dir/OutdoorPvP/OutdoorPvPScriptLoader.cpp.o /home/circleci/project/src/server/scripts/OutdoorPvP/OutdoorPvPNA.cpp:32:14: fatal error: unused variable 'NA_HALAA_GRAVEYARD' [-Wunused-const-variable] uint32 const NA_HALAA_GRAVEYARD = 993; 

[ 89%] Building CXX object src/server/scripts/CMakeFiles/scripts.dir/Outland/boss_doomwalker.cpp.o /home/circleci/project/src/server/scripts/OutdoorPvP/OutdoorPvPZM.cpp:40:14: fatal error: unused variable 'ZM_GRAVEYARD_ID' [-Wunused-const-variable] uint32 const ZM_GRAVEYARD_ID = 969; 
```

mdX7:
does a case exist where we could *accidently* spawn outside battlefield area because distance to non-battlefield gy is closer? if no we could also drop `Battlefield::GetClosestGraveyard` and rely on worldstate conditions here aswell

sql/updates/world/master/2023_mm_dd_xx_world.sql
```diff
@@ -0,0 +1,48 @@
...
+    `Faction`,
+    CONCAT(
+        'Graveyard - ',
+        `ID`,
+        ' - Team ',
+        CASE
+            WHEN `Faction` = 469 THEN 'Alliance'
+            WHEN `Faction` = 67 THEN 'Horde'
+            ELSE `Faction`
+        END
```
mdX7, line 23: also add stuff from comment field from graveyard_zone

sql/updates/world/master/2023_mm_dd_xx_world.sql
```diff
@@ -0,0 +1,48 @@
...
+    )
+FROM `graveyard_zone`
+WHERE `Faction` <> 0
+GROUP BY 1, 2, 3, 4, 5;
```
mdX7, line 27: what is your intention with this group by?

mdX7 (commented):
for gy 7 and similar we would have to add graveyard zone id to conditions indeed.
In terms of battlefields the idea i had won't properly work, we'd have to give battlefields gys higher priority than regular graveyards. (Imagine being attacker in wintergrasp, dying in the top right while all bases are controlled by defender, then you'd respawn in icecrown)

sql/updates/world/master/2023_mm_dd_xx_world.sql
Golrag: toi avoid duplicates. But this will be removed later when support for ghostzone in conditions are added

PR #28974, opened 2023-05-12
Scripts/Spells: Implement Shield Charge

SargeroDeV:
Changes proposed:

    Implement Shield Charge (spellId: 385952)

Issues addressed:

None.

Tests performed:

Known issues and TODO list:

None.


src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -400,6 +401,27 @@ class spell_warr_rallying_cry : public SpellScript
...
+    PrepareSpellScript(spell_warr_shield_charge);
+
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_WARRIOR_SHIELD_CHARGE_EFFECT });
+    }
+
+    void HandleDummy(SpellEffIndex /*effIndex*/)
+    {
+        GetCaster()->CastSpell(GetHitUnit(), SPELL_WARRIOR_SHIELD_CHARGE_EFFECT);
```
Nyr97, line 416: Are you sure you don't need any TriggerCastFlags?

SargeroDeV: Yes

mdX7:
theres one remaining issue, thats that the AoE of the impact cast happens when the charge starts, not when the target is reached via charge. But that's something i'll fix soon, we also have a similar issue with Elder Leaxa in Underrot :)

thanks for the pr :+1:

PR #28978, opened 2023-05-13
Fixed the quest dialog could not be closed automatically after getting the quest reward when the quest giver has no other follow-up quests.

HuangStomach:
**Changes proposed:**

- Fixed the quest dialog could not be closed automatically after getting the quest reward when the quest giver has no other follow-up quests.
- Add the `const` keyword to functions and arguments that are semantically read-only.

**Issues addressed:**

discussion: https://github.com/TrinityCore/TrinityCore/discussions/28971#discussion-5188356

**Tests performed:**

build completed, tested in-game

None.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -14474,47 +14474,28 @@ bool Player::IsActiveQuest(uint32 quest_id) const
     return m_QuestStatus.find(quest_id) != m_QuestStatus.end();
 }
 
-Quest const* Player::GetNextQuest(ObjectGuid guid, Quest const* quest) const
+Quest const* Player::GetNextQuest(Object const* questGiver, Quest const* quest) const
 {
-    QuestRelationResult quests;
     uint32 nextQuestID = quest->GetNextQuestInChain();
+    if (!nextQuestID)
+        return nullptr;
 
-    switch (guid.GetHigh())
+    if (questGiver == this)
     {
-        case HighGuid::Player:
-            ASSERT(quest->HasFlag(QUEST_FLAGS_AUTOCOMPLETE));
-            return sObjectMgr->GetQuestTemplate(nextQuestID);
-        case HighGuid::Creature:
-        case HighGuid::Pet:
-        case HighGuid::Vehicle:
-        {
-            if (Creature* creature = ObjectAccessor::GetCreatureOrPetOrVehicle(*this, guid))
-                quests = sObjectMgr->GetCreatureQuestRelations(creature->GetEntry());
-            else
-                return nullptr;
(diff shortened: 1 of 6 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -16943,7 +16943,7 @@ void Player::SendQuestReward(Quest const* quest, Creature const* questGiver, uin
     {
         if (questGiver->IsGossip())
             packet.LaunchGossip = true;
-        else if (questGiver->IsQuestGiver())
+        else if (questGiver->IsQuestGiver() && GetQuestDialogStatus(questGiver) > QuestGiverStatus::TrivialDailyQuest)
```
Shauren, line 16946: QuestGiverStatus is a bitmask type, use appropriate comparison operators

Shauren (changes requested):

ZaDarkSide:
I don't think this fixes anything, for example quest [Join the Battle!](https://www.wowhead.com/quest=28780/join-the-battle) the 3rd Quest when playing as a Human Hunter in Northshire still gets stuck on screen after you press Accept button, curiously enough if you set in `WordServer.conf` the setting `Quests.IgnoreAutoAccept = 1` then the Quest Window no longer gets stuck on screen.

HuangStomach:
> I don't think this fixes anything, for example quest [Join the Battle!](https://www.wowhead.com/quest=28780/join-the-battle) the 3rd Quest when playing as a Human Hunter in Northshire still gets stuck on screen after you press Accept button, curiously enough if you set in the setting then the Quest Window no longer gets stuck on screen.`worldserver.conf``Quests.IgnoreAutoAccept = 1`

This change is in the `SendQuestReward` function and corrects the fact that the dialog box does not close automatically when the quest reward is obtained. I'll change the description of this pr.

ZaDarkSide:
Can you give an example of what Quest it fixes, like I did? So I can test before and after I applied your fix? Because I understand now, that my example issue is not the exact issue you are trying to solve, so my issue is a separate issue from this one.

HuangStomach:
I found this out when I first started the orc newbie quest.
For example, the first quest after choosing Exile's Reach [warming-up](https://www.wowhead.com/quest=59926/warming-up)
and the early quests in Valley of Trials [cactus-apple](https://www.wowhead.com/object=171938/cactus-apple)

HuangStomach:
I found this out when I first started the orc newbie quest.
For example, the first quest after choosing Exile's Reach [warming-up](https://www.wowhead.com/quest=59926/warming-up)
and the early quests in Valley of Trials [cactus-apple](https://www.wowhead.com/object=171938/cactus-apple)

PR #28987, opened 2023-05-14
Scripts/Spells: Implement Brutal Vitality

SargeroDeV:
Changes proposed:

Implement Brutal Vitality (spellId: 384036)

Issues addressed:

None.

Tests performed:

Known issues and TODO list:

None.


src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -92,6 +93,41 @@ class spell_warr_bloodthirst : public SpellScript
...
+    {
+        damageAmount += CalculatePct(eventInfo.GetDamageInfo()->GetDamage(), GetEffectInfo(EFFECT_0).CalcValue(GetTarget()));
+    }
+
+    void HandleDummyTick(AuraEffect const* /*aurEff*/)
+    {
+        if (damageAmount == 0)
+            return;
+
+        if (AuraEffect* ignorePainAura = GetCaster()->GetAuraEffect(SPELL_WARRIOR_IGNORE_PAIN, EFFECT_0))
```
mdX7, line 116: `GetCaster()` can be null in AuraScripts (imagine i'm moonfiring you and logout instantly)

in this case we can just use `GetTarget()` because its the same unit at all times

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -92,6 +93,41 @@ class spell_warr_bloodthirst : public SpellScript
...
+        damageAmount = 0;
+    }
+
+    void Register() override
+    {
+        AfterEffectProc += AuraEffectProcFn(spell_warr_brutal_vitality::HandleProc, EFFECT_0, SPELL_AURA_PERIODIC_DUMMY);
+        OnEffectPeriodic += AuraEffectPeriodicFn(spell_warr_brutal_vitality::HandleDummyTick, EFFECT_0, SPELL_AURA_PERIODIC_DUMMY);
+    }
+
+    uint32 damageAmount = 0;
```
mdX7, line 128: its a private member variable, we are using the convention of `_lowerCamelCase` for such.

```cpp
private:
    uint32 _damageAmount = 0;
```

mdX7 (changes requested):

Shauren (changes requested):

src/server/scripts/Spells/spell_warrior.cpp
SargeroDeV: Thanks for the explanation

SargeroDeV: Now I have a question, in this case the spell Brutal Vitality (spellId: 384036) is added to me through the talent tree, that makes me Player in his Caster and his Target, isn't it? so in case I disconnect myself the same Caster and Target both will become null anyway.

mdX7: An aura cannot exist without a target. Target is always the person having the aura (debuff/buff). And if that person disconnects aura is gone aswell, no need to think about this case on scripting end :)

PR #29003, opened 2023-05-18
Core/Graveyards: Fix graveyards in subzones when there is no valid graveyard in that subzone

Golrag:
Currently, when you die as a horde player in Northshire, your ghost doesn't get teleported to the nearest horde graveyard. This PR makes resolves this issue by looking at the parent area if there is any.

**Changes proposed:**

-  Searching for a correct graveyard through parent area when none is found

**Issues addressed:**

Closes: none?

**Tests performed:**

1. die as horde player in Northshire --> Correctly teleported away now
2. die in the Great Sea between vashjir and stormwind as a Horde player --> teleported to tirisfal glades (yep this is correct).

- None

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -7096,6 +7096,28 @@ WorldSafeLocsEntry const* ObjectMgr::GetClosestGraveyard(WorldLocation const& lo
         }
     }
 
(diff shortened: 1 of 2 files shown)


Faq:
Is there a need to to keep old `GetClosestGraveyard` and make new one? I mean not to remake old `GetClosestGraveyard`, so it handles also this case?

PR #29005, opened 2023-05-18
Scripts/Warrior: Implement Critical Thinking

SargeroDeV:
Changes proposed:

Implement Critical Thinking (spellId: 389306)

Issues addressed:

None.

Tests performed:

Known issues and TODO list:

None.

### DIFF
--- a/src/server/scripts/Spells/spell_warrior.cpp
+++ b/src/server/scripts/Spells/spell_warrior.cpp
@@ -283,14 +283,12 @@ class spell_warr_critical_thinking : public AuraScript
         return ValidateSpellInfo({ SPELL_WARRIOR_CRITICAL_THINKING_ENERGIZE });
     }
 
-    void HandleProc(AuraEffect* /*aurEff*/, ProcEventInfo& eventInfo)
+    void HandleProc(AuraEffect* aurEff, ProcEventInfo& eventInfo)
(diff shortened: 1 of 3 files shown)


src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -199,6 +200,32 @@ class spell_warr_colossus_smash : public SpellScript
...
+
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_WARRIOR_CRITICAL_THINKING_ENERGIZE });
+    }
+
+    void HandleProc(AuraEffect* /*aurEff*/, ProcEventInfo& eventInfo)
+    {
+        Unit* target = GetTarget();
+        int32 rageAmount = *eventInfo.GetProcSpell()->GetPowerTypeCostAmount(POWER_RAGE);
```
Shauren, line 216: GetPowerTypeCostAmount returns a Optional, you cannot simply access its value without checking if it has a value

Shauren (changes requested):

Shauren:
Your spell_proc data didnt work because 163201 does not have a damaging effect (it only triggers a spell that deals damage)

PR #29014, opened 2023-05-20
Players/Instances: Repop players at instance entrance if any when they release spirit

Golrag:
**Changes proposed:**

-  Change `Player::RepopAtGraveyard` to take into account `InstanceScript::GetEntranceLocationId`
-  If no entrance is found for that instance script, fall back to default graveyard lookup
-  This means every instance will need its own `graveyard_zone` entry for each entrance

**Issues addressed:**

Closes: none?

**Tests performed:**

- Builds

- none


src/server/game/Entities/Player/Player.cpp
```diff
@@ -4842,13 +4842,18 @@ void Player::RepopAtGraveyard()
...
     if (Battleground* bg = GetBattleground())
         ClosestGrave = bg->GetClosestGraveyard(this);
-    else
+    else if (Battlefield* bf = sBattlefieldMgr->GetBattlefieldToZoneId(GetMap(), GetZoneId()))
+        ClosestGrave = bf->GetClosestGraveyard(this);
+    else if (InstanceScript* instance = GetInstanceScript())
     {
-        if (Battlefield* bf = sBattlefieldMgr->GetBattlefieldToZoneId(GetMap(), GetZoneId()))
-            ClosestGrave = bf->GetClosestGraveyard(this);
+        uint32 entrance = instance->GetEntranceLocation();
```
Seyden, line 4849: just put that code into GetEntranceLocation when its 0, in that case you can just call GetEntranceLocation

Golrag: that might break scripts from people who are using this features in their own forks. That being said, maybe my previous attempt with ZoneScript wasn't so bad after all. Could have just overriden it for InstanceScript to call EntranceLocation if present else find graveyard 🤔

Seyden: Since when do we care about forks?

PR #29026, opened 2023-05-27
Enhance stack rule 3 handling

Rochet2:
**Background:**
What is stack rule 3?
The world database `spell_group_stack_rules` table's `stack_rule` column's value 3 which is defined as `SPELL_GROUP_STACK_RULE_EXCLUSIVE_SAME_EFFECT	Same effects of spells will not stack, yet auras will remain on a target` in [the DB wiki here](https://trinitycore.info/database/335/world/spell_group_stack_rules).

What does it do?
The spells in the group need to have a common effect and then that effect will be prevented from stacking between all of the spells. For example, a haste aura could be blocked from stacking on the player from similar spells. The core logic automagically selects the aura effect of the spells in the group that the rule is used for by a custom heuristic. The core logic will select the FIRST found spell effect with MOST OCCURRENCES in the spell effect list. This selection logic is called "guessing" in the core code comments.

**Issues addressed:**

There was an issue when trying to make a custom group, in which case the logic chooses wrong effect and posts an error.
```
SpellId XXX listed in `spell_group` with stack rule 3 does not share aura assigned for group YYY
```
Basically the error says that the spell id XXX does not share any aura effect with the other spells in the same group YYY.

**Why does this error occur?**
There was a group that has two spells:
```
Spell A has spell effects 41, 41, 54.
Spell B has spell effects 3, 54, 0.
```
The core logic will select the FIRST found spell effect with MOST OCCURRENCES in the spell effect list. Then it will make sure that the effect is in all the spells in the group. In this case it will select 41 as the spell effect that all spells in the group are expected to share. This is because 41 and 54 are equally common in the list of spell effects, but 41 was detected first. Then the core triggers the error because spell B does not have the expected spell effect.

Rather obviously, the intended spell effect for the stacking rule was 54. This is because it is the only spell effect that the two spells share. The selection of 41 as the spell effect for stacking rule seems relatively arbitrary and wrong.

This PR attempts to change that so that the spell effect selection will make more sense in cases like this.

**Changes proposed:**

- The old guessing code calculates the amount of occurrences of the effect types in each spell. However, there is an issue with this method when one defines custom groups. As an example, with the old code if a spell with effects (1, 1, 2) and another spell with effects (2, 3, 4) are in a group, the code would choose the wrong effect type and post an error since the first found most frequent effect would be 1. But actually the spells only share the effect 2. The new calculation only considers each effect type once from each spell (and its ranks), which fixes this issue.
- Spells like soothing kiss (27275) have the spell rank 1 in DB, but the spell effect we look for is actually only in the last rank of the spell. This means that the frequency calculation does not take into account any other spell ranks for the "guessing" even if they may be relevant. In this change we now find spell effect types from all ranks of the spell for the "guessing". This is because all ranks of the spell are affected by the group. You can see [here](https://github.com/TrinityCore/TrinityCore/blob/fb388298a0bf76402a6f1de35bdfcb33223d51bb/src/server/game/Spells/SpellMgr.cpp#L405-L408) the logic that applies the group during runtime.
- The code now treats multiple aura types that are equally frequent to both be used for the stacking rule.
- The code now finds only effects that are common between the different spells. This means that even if an effect is more frequent than another, it may not be selected if it is not shared between all other spells (or their ranks)

**Tests performed:**

- Builds,
- The changes do not affect existing data calculation on 3.3.5. This was tested by printing the results of old and new code with the following code and they provided identical results.

```cpp
    auto prettyPrintSpellMap = [](const SameEffectStackMap& spellMap) {
        for (const auto& spellPair : spellMap) {
            std::cout << "Spell Group: " << static_cast<int>(spellPair.first) << std::endl;
            std::cout << "Aura Names: ";

            const auto& auraNames = spellPair.second;
            for (uint32_t auraName : auraNames) {
                std::cout << auraName << " ";
            }

            std::cout << std::endl << std::endl;
        }
    };
    prettyPrintSpellMap(mSpellSameEffectStack);
```

- [ ] Same code could be probably applied to master branch, but I have not checked if the change affects the results of the stacking rules on master.


jackpoz:
The amount of modified code and affected logic makes this a fearful PR to review. I'm unsure if anyone will step up and review it, or if this should be one of those cases "fingers crossed 🤞 " and merge it, with worst case scenario being reverted.
Do you have an idea about what kind of issues we should keep an eye out for ? Even if they will probably be reported in 1 year at the earliest anyway.

Rochet2:
> The amount of modified code and affected logic makes this a fearful PR to review.

The code is used to load data from database into a container. By reviewing the result of old code and new code we should already see that merging the code should not affect anything. I did that review for 3.3.5 branch. Furthermore, it looks like on master branch there is no rows with `stack_rule=3` at all in `spell_group_stack_rules`, so master branch is unaffected.

It will only affect custom code if someone uses spell groups and stack_rule 3.

> Do you have an idea about what kind of issues we should keep an eye out for ?

I added error cases, so that people should get errors in logs if something is totally wrong.
Apart from having an error on startup, people cannot know if the behaviour changed on their side.
The only way to see a difference is to:
* run old code and log warning when new code produces different results
* or see in game that things are not as expected
* or think about if the logic applies differently to your own data

The differences in spells would be that certain spell effects do not stack as expected during game play - so they do not affect your stats or speed etc. Lets say two buffs would slow you 50%, then if they are grouped they should in total slow you only to 50% and not to a total halt. So spells can either now stack or not stack compared to how they used to stack before (stacking behaviour changed).

The data that would be affected is custom `spell_groups` with `stack_rule=3` in `spell_group_stack_rules`. And any of the ranks of those spells.

Rochet2:
To add to this, the documentation here https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130134/spell+group+stack+rules
says that
>3 | SPELL_GROUP_STACK_RULE_EXCLUSIVE_SAME_EFFECT | Same effects of spells will not stack, yet auras will remain on a target

But this is not exactly how it has been working or will work after this change.
It has always selected the most common effect between the aura effects. It has never applied to all effects of the spells.
Not sure if it should affect all spell effects that are shared instead of "most frequent" spell effect - which is the case with old and new code. The new code just changes how the frequency is calculated a bit to make it work in custom cases.

Aokromes:
Conflicting files
src/server/game/Spells/SpellMgr.cpp

ccrs:
the code is structured as follows:
block one
- get auratypes from group spells related info
- count the amount of times each auratype is repeated

block two --> this one is a bit iffy, I'm sure there is a much better way to achieve this result
- get the common list of all auratypes

block three
- from all the common auraTypes, get which are the most repeated, and only give me a list if all of them are repeated the same amount of times

block four
- from these most repeated common auraTypes,
- if its found within the manual exclusion subgroup "haste", include the whole group auraType as a stack --> why only validate with the front within this subgroup?
- if I have not found it in the "haste" subgroup, add everything? all the "repeated common auraTypes"? while iterating on that very same container? --> does not make much sense, I'm already changing that

ccrs:
now at least it makes sense to me? I think there is definately something salvageable here @Shauren, leaving this to your final review

edit:
- if its found within the manual exclusion subgroup "haste", include the whole group auraType as a stack --> why only validate with the front within this subgroup?
--> cause only the front is inserted previously if it was matched while iterating through spell effects
--> this makes me paranoid, but makes sense?

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -1397,100 +1397,115 @@ void SpellMgr::LoadSpellGroupStackRules()
...
+                    ++frequencyContainer[auraType];
             }
 
-            uint32 auraType = 0;
-            size_t auraTypeCount = 0;
-            for (uint32 auraName : frequencyContainer)
+            // Find aura types that all spells share
+            std::unordered_set<AuraType> commonAuraTypes;
+            for (const auto& [spellId, auraTypes] : auraTypesBySpellId)
+                commonAuraTypes.insert(auraTypes.begin(), auraTypes.end());
```
Rochet2, line 1463: >Find aura types that all spells share

This is not true if we do an union like that in cases like this:
```
SpellA: 10
SpellB: 10
SpellC: 20, 30
```

The union will produce {10,20,30} as commonAuraTypes even if it should most likely just produce {} since the types are not shared by the spells.
And since 10 is the most common one, it will be chosen.

So perhaps it should still be an intersection as it used to be
```suggestion
            // Find aura types that all spells share (intersection)
            std::unordered_set<AuraType> commonAuraTypes;
            for (const auto& [spellId, auraTypes] : auraTypesBySpellId)
            {
                if (commonAuraTypes.empty())
                    commonAuraTypes.insert(auraTypes.begin(), auraTypes.end());
                else
                {
                    std::unordered_set<AuraType> intersection;
                    for (AuraType const& auraType : auraTypes)
                    {
                        if (commonAuraTypes.contains(auraType))
                            intersection.insert(auraType);
                    }
                    commonAuraTypes = std::move(intersection);
                }
            }
```

ccrs: since I dont know the expected behavior on this scenario

SpellA: 10
SpellB: 10
SpellC: 20, 30

I cant really comment, leaving "what feels right" aside

Rochet2: >since I dont know the expected behavior on this scenario

The 335 data does not contain such scenarios.
This is why my original test passed even if the code used an intersection:

>The changes do not affect existing data calculation on 3.3.5.

Apart from checking master branch data for example, I dont know if there is any other way to decide than "what feels right" and "what works as expected". It is a bit unexpected that a stacking rule between two spells is related to a non shared effect - effectively not blocking any stacking.

Rochet2: This is basically a variation of the original issue that was being fixed in this PR.

PR #29029, opened 2023-05-28
Core/Creatures: Implement CREATURE_STATIC_FLAG_4_IGNORE_LOS_WHEN_CASTING_ON_ME

Golrag:
**Changes proposed:**

- Replace CREATURE_FLAG_EXTRA_UNUSED_22 with CREATURE_FLAG_EXTRA_IGNORE_LOS_WHEN_CASTING_ON_ME to be able to set this value
- Add extra checks to prevent LoS checking

**Issues addressed:**

Closes: none? mabye it properly fixes the old ice tomb issue

**Tests performed:**

- Builds
- Tested in game by adding the Ice Tomb gameobject and an NPC inside it and casting Heroic Throw & charge (in a raid)

-  ~Maybe this should just be placed withing Object::IsWithinLoSInMap or others.~
- [ ] This might mean that this whole calculate LoS on object edges etc is not needed/not blizzlike at all and this flag should fix it.

Feel free to tell me if any of those two todo's should be handled in this PR.


Golrag:
There currently isnt a field in database where we can set static flags directly (some are set with movement_template table, or type_flags or others even). Combining all those fields and putting them into a single static flags column (or multiple due to the size) sounds like a lot of work tho

Nyr97:
iirc Static Flags will soonTM be handled, though some are already

Shauren:
static flags can now be set in db, don't add new flags extra

Golrag:
Alright. What do I have to do to initialise it now? It's being filled in in the CreatureDifficulty but not in Creature. Do I need to 'manually' add that flag to `Creature::ApplyAllStaticFlags`  ? I thought it would add them all from creature_difficulty_template and then apply side effects

Shauren:
There are no side effects for this flag (no unitflag change, no movement type change, nothing) so this is all that you need

Golrag:
What I meant to say was that I currently fail to see where `Creature::_staticFlags`  is being set to the values of `CreatureDifficulty::StaticFlags`. So I still need to add _staticFlags.applyFlag(CREATURE_STATIC_FLAG_4_IGNORE_LOS_WHEN_CASTING_ON_ME);` in `Creature::ApplyAllStaticFlags`?

Shauren:
uh... oops?
no, `Creature::ApplyAllStaticFlags` was supposed to take the flags and shove all of them into _staticFlags

Golrag:
Alrighty :)

Golrag:
What about the TODO's I described in this PR? Is this something we should look into?

Shauren:
> `Object::IsWithinLoSInMap`

 no, that would mean that all non-spell los checks would also pass (for example aggro)

>This might mean that this whole calculate LoS on object edges etc is not needed/not blizzlike at all and this flag should fix it. 

probably

But this solution is incomplete, there are more LoS checks in Spell.cpp: chain targeting, `if (losPosition)` (for ground targeted aoe, i tested this one on retail should also ignore los, testable npc is https://www.wowhead.com/npc=67423/shieldwall-peasant)

Golrag:
I put all the LoS checks that were happening inside Spell.cpp into their own functions to catch stuff like DisableMgr & attributes at one location. Hope I caught them all now and didnt break other stuff where those checks were left out intentionally.

That being said, want me to remove the object edge LoS stuff in this PR as well?

Shauren:
No, any potential removal would be in a separate PR

PR #29041, opened 2023-05-30
Core/Creatures: Drop dynamic flags from creature_template

meji46:
**Changes proposed:**

-  Drop dynamic flags from creature_template (the current implementation forces the DB value to 0)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


Nyr97:
So basically dynamicFlags are fully controlled by the core no matter what dynFlag we set in the template? 👀

Shauren:
What about `creature`?

malcrom:
That was added at my request back in the day before we could set this any other way. Isn't this able to be done in SAI now?

Shauren:
No and no flag currently makes sense to be set in db (feign death is not in dynamicflags anymore and even if it was there, you should use one of the "permanent feign death" auras that we have scripted to automatically apply neccessary flags)

malcrom:
Back when it was added it was needed because the creatures it was needed for had no aura to make them look dead. They had flags set. That was 10 years ago or so but it was needed at the time.

malcrom:
Sorry Shauren I thought it was creature not creature template

Shauren:
Its both, there are no dynamic flags that make sense to be applied in db anymore

Nyr97:
@malcrom Modox handled several feign death auras in spell_generic.cpp, some set no emotes at all during the aura, etc.

malcrom:
> @malcrom Modox handled several feign death auras in spell_generic.cpp, some set no emotes at all during the aura, etc.

My error. I thought he was taking about dynamic flags in creature table which was added because at the time there was no other way to make some creatures appear dead that didn't have auras. I'm all for removing it.

Issue #26923, opened 2021-09-17
SQL account online


gygyti:
### Description

I will try to explain the problem. 
when you log into your account, where you choose a character to start playing for:
then
auth.account online set to 1 - ok we're really online.

however, the problem is that when you choose your character and enter the game , then auth.account online set to 1 again

### Expected behaviour

i think this is an extra load on the sql server. There is no need to update the value if it already matches. 

### Steps to reproduce the problem

check auth.account and enter the game
(you can check it with triggers)

### Branch

3.3.5

### TC rev. hash/commit

fece5bd

### Operating system

Debian 10 x64

### Custom changes

none

jackpoz:
Are you referring to some particular lines of code or ?

gygyti:
> 
> 
> Are you referring to some particular lines of code or ?

i think yes, when you enter the game, a request is made to the auth database

jackpoz:
When you say "if it matches already" you mean if one has logged already with another character ? How does one know if a value matches without checking it with a query ? In mysql SELECT+UPDATE is more expensive than a simple UPDATE

gygyti:
> 
> 
> When you say "if it matches already" you mean if one has logged already with another character ? How does one know if a value matches without checking it with a query ? In mysql SELECT+UPDATE is more expensive than a simple UPDATE

"if it matches already" I mean, the server sends a request to Mysql: UPDATE `online` = 1, but this is not required, since` online` is already 1. This happens when you log in to the game (but before that in your account) - but why? if you log into your account and the online value is already 1
update is an extra iteration when entering the game

CraftedRO:
I found the way to remove the second "enter world" update on auth.account online and it works it will only send the update query (LOGIN_UPD_ACCOUNT_ONLINE) once elsewhere (not sure where) when you login credentials but I have no idea if this is really a thing to do or not:

those 3 lines can be deleted and it will build without problem:

https://github.com/TrinityCore/TrinityCore/blob/39bebe6a653bdd65160aa7f9fef6501d43884079/src/server/game/Handlers/CharacterHandler.cpp#L1278

PR #29071, opened 2023-06-11
Remove extra online update query when login in world on character

CraftedRO:
**Changes proposed:**

-  Remove extra update query when login in world on character since is there already one sent when you enter credentials

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/26923

**Tests performed:**

(Does it build, tested in-game, tested on db side.)

- [ ]

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -845,12 +845,6 @@ void WorldSession::HandlePlayerLogin(LoginQueryHolder const& holder)
 
     CharacterDatabase.Execute(stmt);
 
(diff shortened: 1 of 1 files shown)


Shauren:
This is neccessary in case you have multiple realms running and one of them is restarted, it will clear online flag on its restart

CraftedRO:
it will clear online flag on worldserver restart with this too anyway

Shauren:
Yes but at least this code will set it back to 1 if you change characters

CraftedRO:
it will set back to 1 in AUTH db that's not supposed to be a character online status but rather account, meh anyway I don't know why I even care

PR #29098, opened 2023-06-25
Modernize some pointers and arrays in player class

Ovahlord:
**Changes proposed:**

-  converted some plain raw pointers into smart pointers to recude code size and modernize the code in general
-  converted some c-style arrays to std::array for easier initialization and reduced code size

**Tests performed:**
builds

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -17116,7 +17116,8 @@ void Player::_LoadDeclinedNames(PreparedQueryResult result)
     if (!result)
         return;
 
-    m_declinedname = std::make_shared<DeclinedName>();
+    m_declinedname.reset();
+    m_declinedname = std::make_unique<DeclinedName>();
     for (uint8 i = 0; i < MAX_DECLINED_NAME_CASES; ++i)
         m_declinedname->name[i] = (*result)[i].GetString();
 }
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -219,11 +219,10 @@ Player::Player(WorldSession* session) : Unit(true), m_sceneMgr(this)
     m_DailyQuestChanged = false;
     m_lastDailyQuestTime = 0;
 
-    for (uint8 i=0; i < MAX_TIMERS; i++)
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Player/Player.h
```diff
@@ -3133,8 +3133,8 @@ class TC_GAME_API Player : public Unit, public GridObject<Player>
         WorldLocation m_recall_location;
         uint32 m_recall_instanceId;
 
-        DeclinedName *m_declinedname;
-        Runes *m_runes;
+        std::shared_ptr<DeclinedName> m_declinedname;
```
jackpoz, line 3136: does shared_ptr fit that well here ? as that means it's ok that some object is storing the old m_declinedname object when reloading it, causing some nasty bugs hard to reproduce

Ovahlord: I was looking at the _LoadDeclinedNames logic which deletes the allocated memory during every load so I was convinced that a shared_ptr is fine here because the load method would just create a new ptr and orphan this one here which would then invoke the deallocation

Shauren: I believe the question was "why not unique_ptr?"

Ovahlord: ye, now rethinking about it, I think I went overboard with my logic. will fix

Ovahlord:
so, any complaints left?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -17139,8 +17116,8 @@ void Player::_LoadDeclinedNames(PreparedQueryResult result)
     if (!result)
         return;
 
-    delete m_declinedname;
-    m_declinedname = new DeclinedName;
+    m_declinedname.reset();
```
Shauren, line 17119: eh, reset line becomes unneccessary if the next thing you do is assign new value

Ovahlord: does unique_ptr also destroy objects when assigning a new one? Was expecting it to only do that when going out of scope

Shauren: yes

PR #29099, opened 2023-06-25
Scripts/Hunter: Implement Rapid Fire

Naddley:
**Tests performed:**

build, tested ingame


src/server/scripts/Spells/spell_hunter.cpp
```diff
@@ -430,6 +432,35 @@ class spell_hun_posthaste : public SpellScript
...
+            SPELL_HUNTER_RAPID_FIRE_DAMAGE,
+            SPELL_HUNTER_RAPID_FIRE_ENERGIZE
+        });
+    }
+
+    void HandlePeriodic(AuraEffect const* /*aurEff*/)
+    {
+        if (Unit* caster = GetCaster())
+        {
+            caster->CastSpell(GetTarget(), SPELL_HUNTER_RAPID_FIRE_DAMAGE, true);
```
Nyr97, line 453: Out of curiosity, if you start channeling this spell and you turn around (the target is no longer in your line of sight), with the current script, is the damage spell cast? On retail, it doesn't, so we shouldn't let that happen even if it uses IGNORE_LINE_OF_SIGHT. This attribute is used in case the target goes behind a wall. It's worth testing.

mdX7: shouldn't channel abort in that case?

Nyr97: > shouldn't channel abort in that case?

Channeling persists, but the damage spell is not passed if the target is not in your vision arc

mdX7: working as intended

mdX7 (approved):

PR #29100, opened 2023-06-25
Core/Spells: Fix Death Pact

aquadeus:
**Changes proposed:**

-  Fix absorb of Death Pact spell

**Tests performed:**

Tested in-game

### DIFF
--- a/src/server/scripts/Spells/spell_dk.cpp
+++ b/src/server/scripts/Spells/spell_dk.cpp
@@ -443,6 +443,11 @@ class spell_dk_death_pact : public AuraScript
 {
     PrepareAuraScript(spell_dk_death_pact);
 
+    bool Validate(SpellInfo const* spellInfo) override
+    {
+        return ValidateSpellEffect({ { spellInfo->Id, EFFECT_2 } });
+    }
+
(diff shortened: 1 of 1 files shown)


src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -446,7 +446,7 @@ class spell_dk_death_pact : public AuraScript
     void HandleCalcAmount(AuraEffect const* /*aurEff*/, int32& amount, bool& /*canBeRecalculated*/)
     {
         if (Unit* caster = GetCaster())
-            amount = int32(caster->CountPctFromMaxHealth(amount));
+            amount = int32(caster->CountPctFromMaxHealth(GetEffectInfo(EFFECT_2).CalcValue(caster)));
```
jackpoz, line 449: does this need a Validate check for EFFECT_2 now or is that checked already somewhere ?

Shauren: it does need it

aquadeus: done

mdX7 (approved):

PR #29104, opened 2023-06-28
Core/Unit: modernize some pointers and arrays in unit class

Ovahlord:
**Changes proposed:**

-  modernize most pointers and arrays in unit class where it makes sense
-  make dtor and ctor of the vehicle class public since std::unique_ptr does invoke them outside of the unit methods which is being blocked by protected

**Tests performed:**
- builds

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -327,7 +327,7 @@ Unit::Unit(bool isWorldObject) :
     m_state = 0;
     m_deathState = ALIVE;
 
-    m_currentSpells.fill(nullptr);
+    m_currentSpells = { };
 
     m_auraUpdateIterator = m_ownedAuras.end();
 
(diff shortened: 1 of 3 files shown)


Lordron:
idea outside of scope of this pr - sort Object\WorldObject\Unit\Creature etc fields by their size (and modifiers (public, private etc)?) to reduce gaps and overall size. Maybe replace internal storage of FlaggedValuesArray with map/vector to reduce size

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -327,7 +327,7 @@ Unit::Unit(bool isWorldObject) :
     m_state = 0;
     m_deathState = ALIVE;
 
-    m_currentSpells.fill(nullptr);
+    m_currentSpells = { };
```
Ovahlord, line 330: does using this array initialization method really set nullptr or just NULL?

PR #29115, opened 2023-07-01
Core/AreaTriggers: Teleport to set instance entrances when entering i…

mdX7:
…nstances via areatrigger entity

**Tests performed:**
built, tested ingame

### DIFF
--- a/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
+++ b/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
@@ -756,8 +756,14 @@ void AreaTrigger::DoActions(Unit* unit)
                     case AREATRIGGER_ACTION_TELEPORT:
                     {
                         if (WorldSafeLocsEntry const* safeLoc = sObjectMgr->GetWorldSafeLoc(action.Param))
+                        {
                             if (Player* player = caster->ToPlayer())
+                            {
+                                if (WorldSafeLocsEntry const* instanceEntrance = player->GetInstanceEntrance(safeLoc->Loc.GetMapId()))
+                                    safeLoc = instanceEntrance;
                                 player->TeleportTo(safeLoc->Loc);
+                            }
+                        }
                         break;
                     }
                     default:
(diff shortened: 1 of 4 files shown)


src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -756,8 +756,14 @@ void AreaTrigger::DoActions(Unit* unit)
                     case AREATRIGGER_ACTION_TELEPORT:
                     {
                         if (WorldSafeLocsEntry const* safeLoc = sObjectMgr->GetWorldSafeLoc(action.Param))
+                        {
                             if (Player* player = caster->ToPlayer())
+                            {
+                                if (WorldSafeLocsEntry const* instanceEntrance = player->GetInstanceEntrance(safeLoc->Loc.GetMapId()))
```
Shauren, line 762: So this makes having teleporting areatriggers impossible to use inside instances with all of them pointing to the entrance location now?

mdX7: fixed

src/server/game/Entities/Player/Player.cpp
```diff
@@ -19538,6 +19538,36 @@ void Player::AddInstanceEnterTime(uint32 instanceId, time_t enterTime)
         _instanceResetTimes.insert(InstanceTimeMap::value_type(instanceId, enterTime + HOUR));
 }
 
+WorldSafeLocsEntry const* Player::GetInstanceEntrance(uint32 targetMapId)
+{
+    if (GetMapId() == targetMapId)
+        return nullptr;
```
columbus-eristr, line 19544: Fetching the instance entrance of a player that is already inside the instance doesn't seem entirely unreasonable, feel like this scenario should be supported.

Krudor, line 19544: Seems reasonable to support fetching current entrance while inside the same map, any specific reason to why this should be prevented?

mdX7: agreed

PR #29123, opened 2023-07-04
Core/Scripts: Update eclipse script, now restore stacks in combat too

aquadeus:
**Changes proposed:**

-  Update druid eclipse script

**Tests performed:**

Tested in-game


mdX7:
stacks should be restored after eclipse buff runs out, not every 20s - ref https://www.youtube.com/watch?v=vvAJKWZhZP8

PR #29124, opened 2023-07-05
Use the [[unlikely]] keyword in all assert, warning and error common macros

Ovahlord:
**Changes proposed:**

Use the [[unlikely]] keword which has been introduced with C++20 in all debugging macros in order to improve compiler optimization. This minimizes potential performance loss when using asserts in larger quantities. This makes using macros, such as ASSERT and ASSERT_NOT_NULL less impactful as we can expect these conditions to be met at all times.

**Tests performed:**
- tested ingame


Shauren:
I'll just mention that this does absolutely nothing for MSVC, it ignores this attribute

PR #29130, opened 2023-07-08
EventMap : Allow to use uint32 as eventId

Traesh:
**Changes proposed:**

- Currently the ScheduleEvent method allow a uint32 eventId but store it internally as uint16, which lead to issues if the user pass a value greater than 65535.
- Update the EventMap to really allow uint32 eventId. It allow to simplify some scripts by removing the need to declare a spell, then the event triggering it. You can now directly do events.ScheduleEvent(SPELL_MY_SPELL_ID, 10s);
- Because storing eventId as uint32 force us to switch to uint64 (we store eventId, phase & group in the same property), i used the extra space to allow 16 bits groups & 16 bits phases instead of 8 each.

**Issues addressed:**

None that i saw

**Tests performed:**

Built, test ingame

- [ ] Don't know if switching phase&group in methods to uint16 was the right move, i preferred it that way to indicate that they will be stored on 16 bits. I can revert this change and switch groups to uint8 instead of original uint32.

### DIFF
--- a/src/common/Utilities/EventMap.cpp
+++ b/src/common/Utilities/EventMap.cpp
@@ -35,13 +35,15 @@ void EventMap::SetPhase(uint16 phase)
 
 void EventMap::ScheduleEvent(uint32 eventId, Milliseconds time, uint16 group /*= 0*/, uint16 phase /*= 0*/)
 {
+    uint64 internalEventId = uint64(eventId);
+
     if (group && group < 16)
-        eventId |= (1LL << (group + 31));
+        internalEventId |= (1LL << (group + 31));
 
     if (phase && phase < 16)
-        eventId |= (1LL << (phase + 47));
+        internalEventId |= (1LL << (phase + 47));
 
-    _eventMap.insert(EventStore::value_type(_time + time, eventId));
+    _eventMap.insert(EventStore::value_type(_time + time, internalEventId));
 }
 
 void EventMap::ScheduleEvent(uint32 eventId, Milliseconds minTime, Milliseconds maxTime, uint16 group /*= 0*/, uint16 phase /*= 0*/)
--- a/src/common/Utilities/EventMap.cpp
+++ b/src/common/Utilities/EventMap.cpp
@@ -25,37 +25,37 @@ void EventMap::Reset()
     _phase = 0;
 }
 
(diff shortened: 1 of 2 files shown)


Nyr97:
I am not fond of having people using SPELL_ declarations in enums to schedule events for two reasons:

1. most people will ignore that usage and will declare events anyway.
2. we'll have people using events declarations inside ExecuteEvent and spell declarations, making it confusing to both read and review.

Ovahlord:
yeah, I'm kinda confused why someone would even need that many IDs unless he's doing some unholy shit that deserves better design. So if you say eventIds can only reach the numeric limit of uint16, just cut down the argument instead to reflect that behavior for it is very unlikely one is ever needing that many Ids

Traesh:
As i said in the PR message, the goal would be to avoid unnecessary code. Instead of


```
enum Spells
{
    SPELL_ONE = 123456,
}

enum Events
{
    EVENT_ONE = 1,
}

events.ScheduleEvent(EVENT_ONE, 8s);
```


you could just


```
enum Spells
{
    SPELL_ONE = 123456,
}

events.ScheduleEvent(SPELL_ONE, 8s);

```

Even ignoring the fact that it allow to write less code, I personally find the second way clearer, you know that SPELL_ONE = it will cast this spell. You don't know what EVENT_ONE actually do without reading it. It's pretty clear in this example, but sometimes people are way more creative with event names (EVENT_START_SPITTING, EVENT_CONTINUE_SPITTING, ...)

It's maybe just a preference of me, if i'm the only one liking it we can just switch argument to uint16.

Ovahlord:
I'd keep spell Ids and event Ids separate for clarity purposes because the spell Ids point towards their corresponding dbc/db2 data while eventIds are specifically made for the eventMap and I'd personally keep it that way. Yes, you can reduce code size by just using spell Ids for events but there are MANY cases where plain spell casts are not enough and require additional events. So that alone would cause lots of mixed code which I'm not a fan of.

kelno:
This makes me think it's quite bad that the EventMap functions even allow to use a uint32 eventId but will fail because it only supports a uint16, seems to me that's confusing for no good reason. In the current state the interface of that class should just expose uint16 eventId for function parameters

PR #29131, opened 2023-07-09
Core/Scripts: Fix EventScript assert for GameObject::SetDestructibleState

meji46:
**Changes proposed:**

-  Fix EventScript assert for GameObject::SetDestructibleState. SetDestructibleState is used to initialize or rebuild Wintergrasp with the param attackerOrHealer as nullptr (invoker for ScriptMgr::OnEventTrigger).
-  Changed GameObjectTemplate::GetEventScriptId to GameObjectTemplate::GetEventScriptSet to list all GameEvents defined for certain types of gameobjects

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None


jackpoz:
I wonder if instead of an assert it should be a return false

PR #29133, opened 2023-07-11
Allow autoattack if channel with SPELL_ATTR5_ALLOW_ACTIONS_DURING_CHANNEL

Traesh:
**Changes proposed:**

-  Allow a npc to autoattack if current channeling spell has SPELL_ATTR5_ALLOW_ACTIONS_DURING_CHANNEL
-  Example : Used by most of "SPELL_AURA_MOD_FIXATE" spells

**Issues addressed:**

**Tests performed:**

Build, tested ingame


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -2096,6 +2096,13 @@ void Unit::AttackerStateUpdate(Unit* victim, WeaponAttackType attType, bool extr
     if (HasUnitState(UNIT_STATE_CANNOT_AUTOATTACK) && !extra)
         return;
 
+    if (HasUnitState(UNIT_STATE_CASTING))
```
mdX7, line 2099: we should consider moving this block into a separate method and then just call it here and in `UnitAI::DoMeleeAttackIfReady()` as its literally the same code :)

Shauren: autoattack code unification can wait

Shauren (approved):

PR #29134, opened 2023-07-11
CreatureAI OnHealthDepleted hook

Traesh:
**Changes proposed:**

-  Add "OnHealthDepleted" hook triggered when an unkillable creature reach 1 hp for the first time, or a normal creature reach 0 hp
- Example : npc Nullification Device (196712) should cast a spell when reaching 1 hp then despawn.

**Issues addressed:**

**Tests performed:**

Build, tested ingame

### DIFF
--- a/src/server/game/AI/CreatureAI.h
+++ b/src/server/game/AI/CreatureAI.h
@@ -99,7 +99,7 @@ class TC_GAME_API CreatureAI : public UnitAI
         virtual void JustEngagedWith(Unit* /*who*/) { }
 
         // Called when the creature reaches 0 health (or 1 if unkillable).
-        virtual void OnHealthDepleted(Unit* /*attacker*/, bool /*wasKilled*/) { }
+        virtual void OnHealthDepleted(Unit* /*attacker*/, bool /*isKill*/) { }
 
         // Called when the creature is killed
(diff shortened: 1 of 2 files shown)


Shauren:
This should be called OnHealthDepleted and should trigger for both unkillable creatures (when reaching 1 health) or on death

retail has this this hook `OnHealthDepleted: Unit reaches 0 health (or 1 if unkillable).`

mdX7 (approved):

Shauren:
I see a minor issue with the way you added it for death case, script can remove unkillable flag from OnHealthDepleted, causing it to run twice (second time on real death)

Traesh:
I see your point, multiple things can cause the hook to run multiple times when the creature gets hit again (regen health, remove flag, ...). I added a "wasKilled" parameter so the user have the information of was happen, and he can check it if his hook should only run when health is 1

mdX7:
> I see your point, multiple things can cause the hook to run multiple times when the creature gets hit again (regen health, remove flag, ...). I added a "wasKilled" parameter so the user have the information of was happen, and he can check it if his hook should only run when health is 1

sounds good to me :+1:
then we could drop JustDied afterwards

PR #29145, opened 2023-07-13
Scripts/Mage: Implemented Flurry

### LINKED ISSUE #19585: Core/Script The spell Flurry doesn´t work.
**Description:**

The spell Flurry(ID:44614) given to Mage Frost at lv.20 has an issue

Flurry: Unleach a flurry of ice shards, striking the target 3 time for a total of (x) frost damage. Each shard reduces the target´s movement speed by 70% for 1 sec

**Current behaviour:**

The spell doesn´t work at all, when you cast the spell nothing happen, you don´t attack the target and any visual effect is activated

**Expected behaviour:**

when you use the spell you should strike the target 3 times with ice shards for (x) total damage and each ice shards should reduce the movement of the target 70% for 1 sec 

1. Create a Mage character
2. Reach lv.20 and choose the Frost Spec
3. Use the spell Flurry in an enemy

**Branch(es):**

Master

**TC rev. hash/commit:** 

Description in wowhead http://www.wowhead.com/spell=44614/flurry

aquadeus:
**Changes proposed:**

-  Implement flurry talent

**Issues addressed:**

Closes #19585

**Tests performed:**

Tested in-game

### DIFF
--- a/src/server/scripts/Spells/spell_mage.cpp
+++ b/src/server/scripts/Spells/spell_mage.cpp
@@ -774,49 +774,46 @@ class spell_mage_flame_on : public AuraScript
    }
 };
 
-class FlurryEvent : public BasicEvent
+// 44614 - Flurry
+class spell_mage_flurry : public SpellScript
 {
+    class FlurryEvent : public BasicEvent
+    {
     public:
-        FlurryEvent(Unit* caster, ObjectGuid const& targetGUID, uint8 count) : _caster(caster), _targetGUID(targetGUID), _count(count) { }
+        FlurryEvent(Unit* caster, ObjectGuid const& target, ObjectGuid const& originalCastId, int32 count)
+            : _caster(caster), _target(target), _originalCastId(originalCastId), _count(count) { }
 
         bool Execute(uint64 time, uint32 /*diff*/) override
         {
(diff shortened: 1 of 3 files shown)


src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -816,6 +818,81 @@ class spell_mage_flame_on : public AuraScript
...
+    {
+        return ValidateSpellInfo
+        ({
+            SPELL_MAGE_FLURRY_DAMAGE
+        });
+    }
+
+    void EffectHit(SpellEffIndex)
+    {
+        GetCaster()->m_Events.AddEventAtOffset(new FlurryEvent(GetCaster(), GetHitUnit()->GetGUID(), GetEffectValue() - 1), randtime(200ms, 300ms));
```
Ovahlord, line 863: milliseconds like this usually imply that such events should happen alongside the main proc so better remove it alltogether and just move it into the effect handler

Shauren: No, EFFECT_1 of this spell is a SPELL_EFFECT_TRIGGER_MISSILE which also deals damage

Ovahlord: So why not attaching that part to the missile then?

Issue #19720, opened 2017-05-10
Core/Script The passive spell Ysera's Gift is not working.


HannibalRoG:
**Description:**

The druid's passive spell Ysera's Gift (ID: 145108) learned at level 34 in the specialization of Restauration, is not working.

Ysera`s Gift (passive): Heals you for 3% of your maximum health every 5 sec. If you are at full health, an injured party or raid member will be healed instead.

**Current behaviour:**

The passive spell does not heal at all, it does nothing, it has no effects.

**Expected behaviour:**

The spell should heal once every 5 sec the 3% of your maximum health, or heal an ally according to it`s description.

**Steps to reproduce the problem:**

1. Create a Restauration Druid Character.
2. Reach level 34 in the specialization of Restauration.
3. Get injured and wait 5 sec. or get near an injured party or raid member and wait 5 sec.

**Branch(es):**

Master

**TC rev. hash/commit:** 

7c66e7bbc7a401e7a2d8f2bdcead6531678262ea

**TDB version:**  7.2.0+updates

**Operating system:** Windows 10

Description in wowhead http://www.wowhead.com/spell=145108/yseras-gift

Nyr97:
Fixed by https://github.com/TrinityCore/TrinityCore/pull/29148.

PR #29148, opened 2023-07-14
Core/Spells: Implement Ysera's Gift

### LINKED ISSUE #28705: Druid Talent Ysera's Gift
### Description

https://www.wowhead.com/spell=145108/yseras-gift This talent just doesn't work.

### Expected behaviour

Heals you for 3% of your maximum health every 5 sec. If you are at full health, an injured party or raid member will be healed instead.

Just create druid and use that spell with this learned talent.

### Branch

master

### TC rev. hash/commit

https://github.com/TrinityCore/TrinityCore/commit/65e787d18be898532363fea3f8847fa4d7f21192

### Operating system

Debian 11

### Custom changes

None

aquadeus:
**Changes proposed:**

-  Implement Ysera's Gift

**Issues addressed:**

Closes #28705 #19720

Tested in-game

### DIFF
--- a/src/server/scripts/Spells/spell_druid.cpp
+++ b/src/server/scripts/Spells/spell_druid.cpp
@@ -94,7 +94,9 @@ enum DruidSpells
     SPELL_DRUID_TRAVEL_FORM                    = 783,
     SPELL_DRUID_THRASH_BEAR                    = 77758,
     SPELL_DRUID_THRASH_BEAR_AURA               = 192090,
-    SPELL_DRUID_THRASH_CAT                     = 106830
+    SPELL_DRUID_THRASH_CAT                     = 106830,
+    SPELL_DRUID_YSERAS_GIFT_HEAL_PARTY         = 145110,
+    SPELL_DRUID_YSERAS_GIFT_HEAL_SELF          = 145109
 };
 
 class RaidCheck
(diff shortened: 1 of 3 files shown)


src/server/scripts/Spells/spell_druid.cpp
```diff
@@ -1662,6 +1664,36 @@ class spell_dru_wild_growth_aura : public AuraScript
...
+        return ValidateSpellInfo
+        ({
+            SPELL_DRUID_YSERAS_GIFT_HEAL_SELF,
+            SPELL_DRUID_YSERAS_GIFT_HEAL_PARTY
+        });
+    }
+
+    void HandleEffectPeriodic(AuraEffect const* aurEff)
+    {
+        int32 healAmount = int32(GetTarget()->CountPctFromMaxHealth(aurEff->GetAmount()));
```
Nyr97, line 1683: there are effects that increase the % health amount, is there a way to pass target as in CalcValue() so that such effects are taken into account?

mdX7: thats already taken care of in `AuraEffect::CalculateAmount`

Shauren: no, we should not, aurEff amount is already modified

Shauren (changes requested):
145110 requires a target selection script to limit it to a random injured target with Trinity::SelectRandomInjuredTargets(targets, 1, true)

aquadeus:
> 145110 requires a target selection script to limit it to a random injured target with Trinity::SelectRandomInjuredTargets(targets, 1, true)

done

PR #29151, opened 2023-07-15
Core/Spells: Implement Soothing Mist

aquadeus:
**Changes proposed:**

-  Implement Soothing Mist

**Tests performed:**

Tested in-game

### DIFF
--- a/sql/updates/world/master/2023_99_99_99_monk_soothing_mist.sql
+++ b/sql/updates/world/master/2023_99_99_99_monk_soothing_mist.sql
@@ -1,0 +1,5 @@
+DELETE FROM `spell_script_names` WHERE `spell_id` IN (116670, 124682, 124081);
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
+(116670, 'spell_monk_soothing_mist'),
+(124682, 'spell_monk_soothing_mist'),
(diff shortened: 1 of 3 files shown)


Shauren:
This one is on hold, all 3 spells use a not implemented attribute SPELL_ATTR9_ALLOW_CAST_WHILE_CHANNELING, then they should probably use `CalcCastTime` script hook once that attribute is in

Nyr97:
SPELL_ATTR9_ALLOW_CAST_WHILE_CHANNELING is finally implemented in https://github.com/TrinityCore/TrinityCore/commit/3365e6a4b84f59cca5f7ef5fca71fb3979697e52.

> Script is also not quite correct, these spells are supposed to only be instant & allowed to cast if they target the same unit as Soothing Mist target

About this, there's a talent that allows you to switch the target of those spells cast through Soothing Mist, but it also forces Soothing Mist to switch the channel target to that unit when used. It also forces that new Soothing Mist cast to use the missing channel time of the former.

Shauren:
Monk spells are also a bit weird (scripted behavior), all those spells have SPELL_ATTR9_ALLOW_CAST_WHILE_CHANNELING but they still interrupt essence font channel

Nyr97:
Maybe any AuraInterruptFlag at hand, I haven't given any look to it yet.

PR #29152, opened 2023-07-16
Implement AREATRIGGER_FLAG_HAS_FOLLOWS_TERRAIN

### LINKED ISSUE #20119: Core/Area Triggers: AREATRIGGER_FLAG_HAS_FOLLOWS_TERRAIN
**Description:**

Currently AREATRIGGER_FLAG_HAS_FOLLOWS_TERRAIN has the NYI comment in the core (thus not working at all)

**Current behaviour:**

When you send an areatrigger with a spline (for example you set x1 = 0, x2 = 20, thus sending the trigger 20 yards directly forwards from the caster's position and orientation) it will go through terrain. For example if a hill is in front of you, the trigger will simply ignore terrain and pass through it.

**Expected behaviour:**

Using AREATRIGGER_FLAG_HAS_FOLLOWS_TERRAIN we should be able to generate path for the areatriggers so that they properly follow the terrain.

1. Take http://www.wowhead.com/spell=84714/frozen-orb 
2. Set (x,y,z) in spell_areatrigger_spline to (0,0,0), (40,0,0) so that it goes forward 40 yards.
3. Go somewhere where there is an uphill or walls, check how it passes through everything.

**Branch(es):**

master

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/6c60dddbc6fbc8b5179dc4c763054edc7ad487bd

**TDB version:**  7.x

**Operating system:** Win64

Traesh:
**Changes proposed:**

-  Handle AREATRIGGER_FLAG_HAS_FOLLOWS_TERRAIN
-  AT with this flag should follow the terrain z, and just stop if slope is too steep or hit a wall
-  AT without this flag should just follow their spline without any consideration to z or vmaps

**Issues addressed:**

Closes #20119

**Tests performed:**

Tested ingame : https://www.youtube.com/watch?v=xnkvjn1FZUI

- Not exactly an issue, but this will work only with small spline steps (which is what blizzard seem to do too). If you create an areatrigger with splines (0,0),(40,0) if won't work properly. (Edit : Fixed with last reviews)

### DIFF
--- a/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
+++ b/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
@@ -791,8 +791,9 @@ void AreaTrigger::InitSplineOffsets(std::vector<Position> const& offsets, uint32
     // This is needed to rotate the spline, following caster orientation
     std::vector<G3D::Vector3> rotatedPoints;
     rotatedPoints.reserve(offsets.size());
-    for (Position const& offset : offsets)
+    for (uint32 i = 0; i < offsets.size(); ++i)
     {
+        Position const& offset = offsets[i];
         float x = GetPositionX() + (offset.GetPositionX() * angleCos - offset.GetPositionY() * angleSin);
         float y = GetPositionY() + (offset.GetPositionY() * angleCos + offset.GetPositionX() * angleSin);
         float z = lastZ;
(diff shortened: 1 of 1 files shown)


Shauren:
Uhhhh... I kind of hate this

Wouldn't using WorldObject::MovePositionToFirstCollision on each point pair be better here? (and stopping the loop when distance to previous point is too small)

Traesh:
I'll update my PR in the week, as i found more informations in sniffs today

The same AT will not always have the same count of splines, which mean they are probably generated on the fly when casting, they don't have pre-established steps. We should probably just have (0,0),(0,40) as offsets and do the step by ourselves.

The TimeToTarget also change depending on the distance, so if your full TimeToTarget is 4s and you reach a wall at half your path the new TimeToTarget should be 2s.

About your message, sadly MovePositionToFirstCollision won't work. It does not handle going upstair very well, and more important, if you're at the edge of a cliff and spawn an AT it should fly in the air, slowly going down, and not be stuck on the cliff (which would happen with MovePositionToFirstCollision).

I'm mainly trying to figure out the way Blizzard calculate the steps, as they can change for a single AT, even with very close TimeToTargets. We could just hardcode it to 2 meters.

Traesh:
@Shauren I feels like it's a bit better now. I still don't use MovePositionToFirstCollision for the reason explained above but the rest of the system is redone.

"Random" values come from testing & comparing with retail

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -782,22 +782,62 @@ void AreaTrigger::UndoActions(Unit* unit)
...
+
+            rotatedPoints.emplace_back(currentPosition);
+            lastZ = currentPosition.z;
+        }
 
-        UpdateAllowedPositionZ(x, y, z);
-        z += offset.GetPositionZ();
+        // If we stopped before the end of the spline due to LOS,
+        // reduce the timeToTarget accordingly to keep initial speed
+        timeToTarget *= currentPercent;
```
Krudor, line 837: Seems like something that could be taken advantage of by purposely sticking travelling areatriggers + mobs in a corner, not sure if this is controlled by another flag, working differently, or is in line with retail behaviour.

Not saying anything is wrong here but a potential issue.

Traesh: Can confirm it works this way on retail, for the frozen orb at least. Maybe there is a "despawn on collision" flag, i'll check if i find something like this in my tests

Shauren:
I still don't think this implementation is correct

First, a few (3) retail casts
https://youtu.be/A0S9-3qgP_I
goes down following gravity

https://youtu.be/ZXA9kjigJqg
first cast stops at the bottom of the hill and thats also exactly where player movement is stopped if you approach it from a certain angle
second cast goes up a bit (different angle) but its also the same point where player movement stops

Your code would make all 3 of these casts go straight forward and, in the case of second video, straight through the terrain - maps (adt) are not part of line of sight checks (casting spells through such terrain is also allowed)

PR #29153, opened 2023-07-16
Core/AT: Fix areatrigger type disk on growing radius

aquadeus:
Credits to: meji46

**Changes proposed:**

-  Fix areatrigger type disk on growing radius (like disintegration halo)

**Tests performed:**

Tested in-game

### DIFF
--- a/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
+++ b/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
@@ -493,7 +493,7 @@ void AreaTrigger::SearchUnitInCylinder(std::vector<Unit*>& targetList)
 
 void AreaTrigger::SearchUnitInDisk(std::vector<Unit*>& targetList)
 {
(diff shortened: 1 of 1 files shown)


Seyden:
Thats a bad way of doing it, because there are probably a lot more factors playing a role here.. Scale Curve, Extra Scale Curve and and and and

mdX7:
> Thats a bad way of doing it, because there are probably a lot more factors playing a role here.. Scale Curve, Extra Scale Curve and and and and

scale curve stuff is nyi and that stuff can be added when its implemented

Traesh:
This fix won't work as some disk areatriggers just don't grow overtime.

Example with Kurog, search for "Erupting Bedrock" https://www.mythictrap.com/en/vault/kurog

He spawns multiple disks one after another but each disk is static

Nyr97:
Halls of Infusion for Khajin boss as well: we have the same entry areatrigger being sent for 3 different disk sizes (also it's a serverside areatrigger, spell is unk).

PR #29160, opened 2023-07-19
Core/Unit: Implement CannotTurn action type

Nyr97:
**Changes proposed:**

^.

**Issues addressed:**

None.

**Tests performed:**

None.

None.


Nyr97:
Not entirely sure if we prefer to keep raw updates of flags in internal core handling, and keep the setter only for scripts. Let me know what you think so I can change some of the changes I applied.

Issue #21250, opened 2018-01-17
[3.3.5] Spells with TARGET_DEST_DEST_BACK


Keader:
**Description:** Spells with TARGET_DEST_DEST_BACK is hitting targets in left/right of caster.

**Current behaviour:** In encounters like Sindragosa, you get hits of  ```SPELL_TAIL_SMASH (71077) ``` even if you not in Sindragosa back.

**Expected behaviour:** You get hits of Tail Smash only when you are in Sindragosa Back.

It's not caused by "Dragon model", it is broken to players too.

Try it putting 2 characters in Duel. use macro: .cast self 71077 in Character A, and check that you can got hits in Chracter B even when you not in Chracter A back.

**Branch(es):** 3.3.5

**TC rev. hash/commit:** ccf0807be0974dfb8080cb2fc6ea616cc96f062c

@chaodhib / @Golrag  can take a look?
maybe can be related with recent changes of range stuffs xd

Treeston:
depends on how core defines "back". i guess we use 180° right now, and it should be less - but how much? is it identical for all spells?

sirikfoll:
To me it seems to be something like 360º right now, Sindragosa's tank takes spell sometimes :P

Keader:
yeah, but right now 90º and 270º is included xD

Treeston:
i meant cone of 180° radius, so from 90° to 270° relative, yes. the fact that the cone is centered on the dragon's rearside is to be assumed.

Keader:
try with a player, i think that you can see better xd

chaodhib:
So the remaining question is, what should be the opening angle of the cone. Less than 180° most likely. But is it more than 90°? Less?

The answer lies somewhere in the sniffs. Or by experimenting with that class of spell in retail (by moving a bit, launching the spell. take note of position and if hit or not. then repeat).

Shauren:
Or, alternatively the answer lies in legion spell data (it is now in the client)

Golrag:
I haven't really looked at it very closely but it seems to me that you only get knocked back if you are within the range of the effect (20y). The effect hits at some distance at his back, like it should but does indeed just a "circle" aoe around it instead of a cone ? Seems like in Spell::SelectImplicitAreaTargets calls SearchAreaTargets which does not really care about a cone.
I might be completely misunderstanding the issue tho

Keader:
@Golrag put 2 characters side by side in duel and try cast Tail
[...]

PR #29165, opened 2023-07-19
Fix angle calculation in Unit::KnockbackFrom

Krudor:
**Changes proposed:**

- Change unit knockback angle from relative to absolute.
- A relative angle calculation would result in the unit being knocked back in a parabolic curve instead of a straight line from the source.

**Issues addressed:**

- None that I found, #21250 seemed close but wasn't quite it.

**Tests performed:**

- Builds, briefly tested in game, preview available: https://youtu.be/KnkTUS5bMZw
- Video shows current behaviour on master, and then new behaviour after the fix.

- None

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -11910,7 +11910,7 @@ void Unit::KnockbackFrom(Position const& origin, float speedXY, float speedZ, Mo
         GetMotionMaster()->MoveKnockbackFrom(origin, speedXY, speedZ, spellEffectExtraData);
     else
     {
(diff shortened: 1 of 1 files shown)


aquadeus:
nice pr, this also fixes spells like vengeful retreat of dh and dread landing from kazzara

Seyden:
Why is this even open again? This was fixed a long time ago 🤔

Krudor:
Perhaps in a difference scenario?
Didn't look to much into it but there seems to be different calculations depending on where it's being called from.

Creatures are handled by the same Unit::KnockbackFrom function for example but works differently, and already works fine.

PR #29174, opened 2023-07-23
Scripts/Generic: Unify major healing cooldown CDs modifiers for all classes

Nyr97:
**Changes proposed:**

- All major healing cooldown CDs from healing specializations have a 100 % healing increase modifier which takes in if the caster is not in a group or if their group is not a raid group. Rather than individually implementing each one, we're creating a generic solution that can be updated in the future if Blizzard happens to change the effect's index containing such dummy modifier.
- Spells scripted through this implementation:
1. Implement Tranquility talent.
2. Implement Divine Hymn talent.
3. Implement Luminous Barrier talent.
4. Implement Revival talent.
5. Implement Healing Tide Totem talent **partially** (only healing effect part since the spell must be scripted).
6. Implement Rewind talent **partially** (only healing effect part since the spell must be scripted).
- Removed Divine Hymn's Spellscript which is used to filter targets since the cap is included in MaxAffectedTargets and the target system should handle it on itself.

**Issues addressed:**

None.

**Tests performed:**

Tested.

None.


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -371,25 +373,77 @@ class spell_pri_circle_of_healing : public SpellScript
...
+    {
+        Player* player = GetCaster()->ToPlayer();
+        if (!player)
+            return;
+
+        float healBonus = 1.0f;
+
+        // Note: if caster doesn't have group or has one which is not a raid group.
+        Group const* group = player->GetGroup();
+        if (!group || (group && !group->isRaidGroup()))
```
Seyden, line 402: https://github.com/TrinityCore/TrinityCore/pull/29186/files#r1278150297

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -371,25 +373,77 @@ class spell_pri_circle_of_healing : public SpellScript
...
```
Seyden, line 435: https://github.com/TrinityCore/TrinityCore/pull/29186/files#r1278150297

Shauren:
By "raid", the tooltip means "in a raid instance", not "in a raid group"

Nyr97:
but the healing is also decreased in BGs

Shauren:
pvp places also have their own separate damage and healing modifiers, idk about that

Nyr97:
ill do some testing, iirc healing is not decreased in arenas but bgs only

Nyr97:
ready for review

PR #29175, opened 2023-07-23
DB/Creature  hyjal-summit npc (frost-wyrm id 17907 ) no longer falls through textures

6yTblJIka:
npc https://www.wowhead.com/wotlk/npc=17907/frost-wyrm , spawn under texture , fix movement flag to " can fly " and now npc spawn correctly blizzlike


sql/updates/world/3.3.5/2023_07_23_00_world.sql
```diff
@@ -0,0 +1,2 @@
+INSERT IGNORE INTO `creature_template_movement` (`CreatureId`) VALUES (17907);
+UPDATE `creature_template_movement` SET `Flight` = 2 WHERE `CreatureId` = 17907;
```
Nyr97, line 2: You're missing a space at the bottom of the SQL.

Aokromes:
we don't use pull request for sql only fixes, also, we don't use insert ignore into, use delete + insert instead on one issue.

PR #29182, opened 2023-07-26
Scripts/Druid: Implement Flourish

Nyr97:
**Changes proposed:**

- Implement Flourish talent.

**Issues addressed:**

None.

**Tests performed:**

Tested and it works.

None.

### DIFF
--- a/src/server/scripts/Spells/spell_druid.cpp
+++ b/src/server/scripts/Spells/spell_druid.cpp
@@ -896,9 +896,6 @@ class spell_dru_flourish : public SpellScript
 
     void HandleScriptEffect(SpellEffIndex /*effIndex*/) const
     {
-        if (_targetAndAuras.empty())
-            return;
-
         auto const it = _targetAndAuras.find(GetHitUnit()->GetGUID());
         if (it == _targetAndAuras.end())
             return;
--- a/src/server/scripts/Spells/spell_druid.cpp
+++ b/src/server/scripts/Spells/spell_druid.cpp
@@ -898,11 +898,7 @@ class spell_dru_flourish : public SpellScript
             if (!effect)
                 return true;
 
-            Aura* aura = effect->GetBase();
-            if (!aura)
-                return true;
-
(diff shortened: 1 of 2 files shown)


src/server/scripts/Spells/spell_druid.cpp
```diff
@@ -559,6 +559,38 @@ class spell_dru_ferocious_bite : public SpellScript
...
+
+        std::list<AuraEffect*> effectList = target->GetAuraEffectsByType(SPELL_AURA_PERIODIC_HEAL);
+
+        for (AuraEffect* aurEff : effectList)
+        {
+            // Note: exclude any aura that doesn't belong to caster or is not affected by Flourish.
+            if (aurEff->GetCasterGUID() != caster->GetGUID() || aurEff->IsAffectingSpell(GetSpellInfo()))
+                continue;
+
+            if (Aura* aura = aurEff->GetBase())
```
Seyden, line 578: How is it possible that the aura will be null here?

Nyr97:
I updated the old script so that we can filter out targets that were not supposed to be affected anyway.

Nyr97:
@Shauren old, but valid for review

src/server/scripts/Spells/spell_druid.cpp
```diff
@@ -883,6 +883,72 @@ class spell_dru_ferocious_bite : public SpellScript
...
+            Aura* aura = auraEff->GetBase();
+            if (!aura || aura->GetCasterGUID() != GetCaster()->GetGUID())
+                continue;
+
+            SpellInfo const* auraInfo = aura->GetSpellInfo();
+            if (!auraInfo || auraInfo->SpellFamilyName != SPELLFAMILY_DRUID)
+                continue;
+
+            RegisterAffectedAurasForTarget(unitTarget->GetGUID(), aura);
+            hasValidAura = true;
```
Nyr97, line 938: not sure if it'd be any better in terms of performance to check for the value to be false to be set to true, should be the same overhead but whatever

PR #29184, opened 2023-07-27
Scripts/Druid: Implement Power of the Archdruid

Nyr97:
**Changes proposed:**

- Implement Power of the Archdruid talent.

**Issues addressed:**

None.

**Tests performed:**

None.

I'm not sure how the consumption of aura 392303 should be done. Normally, we use Req Spellmod from Proc Hit Mask, but the effect doesn't have any modifier to any spell. @Shauren suggestions?


src/server/scripts/Spells/spell_druid.cpp
```diff
@@ -896,6 +899,52 @@ class spell_dru_omen_of_clarity : public AuraScript
...
+        // Note: max. targets is SPELL_DRUID_POWER_OF_THE_ARCHDRUID's EFFECT_0 BasePoints.
+        uint8 const maxTargets = powerOfTheArchdruidEffect->GetAmount();
+
+        if (targetList.size() > maxTargets)
+            Trinity::Containers::RandomResize(targetList, maxTargets);
+
+        for (Unit* chosenTarget : targetList)
+            target->CastSpell(chosenTarget, eventInfo.GetProcSpell()->GetSpellInfo()->Id, CastSpellExtraArgs(aurEff));
+
+        target->RemoveAurasDueToSpell(GetSpellInfo()->Id);
```
Seyden, line 939: You can actually call Remove() or Remove(removeMode), you dont even need to say: target->Remove(), because the aura is applied on the target anyway

`enum AuraRemoveMode
{
    AURA_REMOVE_NONE = 0,
    AURA_REMOVE_BY_DEFAULT = 1,       // scripted remove, remove by stack with aura with different ids and sc aura remove
    AURA_REMOVE_BY_INTERRUPT,         // removed by aura interrupt flag
    AURA_REMOVE_BY_CANCEL,
    AURA_REMOVE_BY_ENEMY_SPELL,       // dispel and absorb aura destroy
    AURA_REMOVE_BY_EXPIRE,            // aura duration has ended
    AURA_REMOVE_BY_DEATH
};`

PR #29185, opened 2023-07-27
Scripts/Druid: Implement Embrace of the Dream

Nyr97:
**Changes proposed:**

- Implement Embrace of the Dream talent.

**Issues addressed:**

None.

**Tests performed:**

None.

None.


src/server/scripts/Spells/spell_druid.cpp
```diff
@@ -459,6 +464,91 @@ class spell_dru_earthwarden : public AuraScript
...
+            SPELL_DRUID_REJUVENATION,
+            SPELL_DRUID_REJUVENATION_GERMINATION
+        });
+    }
+
+    void FilterTargets(std::list<WorldObject*>& targets)
+    {
+        targets.remove_if([&](WorldObject const* target)
+        {
+            Unit const* unitTarget = target->ToUnit();
```
Seyden, line 511: Can this EVER be false?

Nyr97: if the caster is not a player, yes, some playerAI are druid class and use this exact spell, see bots in arathi brawl

Seyden: But this check doesnt check if its a player or not, its checking for unit

Nyr97: spells can target gameobjects, that's why Im checking

Seyden: This spell can't target gameobjects, the TARGET_TYPE won't target GOBs.. so do you want to check for Players now or do you want to prevent targeting Gameobjects?

In both cases, your check does nothing here

PR #29194, opened 2023-07-28
3.3.5

ghost:
**Changes proposed:**

-  
-  
-  

**Issues addressed:**

**Tests performed:**

- [ ] 
- [ ]


acidmanifesto (changes requested):
You may as well get rid of the breaks at 784 and 786 then as it makes zero sense to break inbetween the cases with nothing happening.

PR #29196, opened 2023-07-28
AreaTrigger: Extend SearchUnitInSphere radius calculation

Seyden:
**Changes proposed:**

-  Implement core handling for OverrideScaleCurve and make it possible to manipulate it via AreaTrigger::SetOverrideScaleCurve
-  Implement ScaleCurveId for Spheres
-  Implement SpellMod::Radius for Spheres

**Issues addressed:**

**Tests performed:**

- [x] Tested in-game with Anduin Befouled Barrier
- [x] Builds

- [ ] Find out if OverrideScaleCurve actually needs AREATRIGGER_FLAG_HAS_DYNAMIC_SHAPE, the reason here is that the AT gets sent with an UPDATE_OBJECT create without the flag and afterwards the OverrideScaleCurve is sent in another packet without the flag being applied, it could be that the flag gets set on activating the overrideScaleCurve but the update being not sent as packet to the client
- [ ] Database support is missing for initial data (wont be handled in this PR)

### DIFF
--- a/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
+++ b/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
@@ -364,7 +364,7 @@ float AreaTrigger::GetOverrideScaleCurveProgress() const
 {
     auto const overrideScale = m_areaTriggerData->OverrideScaleCurve;
     ASSERT(overrideScale->OverrideActive, "OverrideScaleCurve must be active to evaluate it");
-    
+
     if (!GetTimeToTargetScale())
         return 1.0f;
 
--- a/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
+++ b/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
@@ -142,6 +142,7 @@ bool AreaTrigger::Create(uint32 areaTriggerCreatePropertiesId, Unit* caster, Uni
     SetUpdateFieldValue(areaTriggerData.ModifyValue(&UF::AreaTriggerData::DecalPropertiesID), GetCreateProperties()->DecalPropertiesId);
 
     SetScaleCurve(areaTriggerData.ModifyValue(&UF::AreaTriggerData::ExtraScaleCurve), GetCreateProperties()->ExtraScale);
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -441,6 +464,29 @@ void AreaTrigger::SetScaleCurve(UF::MutableFieldReference<UF::ScaleCurve, false>
     }
 }
 
+void AreaTrigger::SetOverrideScaleCurve(uint32 startTimeOffSet, std::array<DBCPosition2D, 2> points, CurveInterpolationMode mode /*= CurveInterpolationMode::Linear*/)
+{
+    AreaTriggerScaleCurveTemplate curveTemplate;
+    curveTemplate.StartTimeOffset = startTimeOffSet;
```
Traesh, line 470: Is there a situation where startTimeOffSet is not GetTimeSinceCreated ?

Seyden: No clue tbh, but I rather not force it, if you can just set it. I mean it’s not like we have mass sniffs

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -509,8 +555,20 @@ void AreaTrigger::SearchUnitInSphere(std::vector<Unit*>& targetList)
                 _shape.SphereDatas.RadiusTarget,
                 sDB2Manager.GetCurveValueAt(GetCreateProperties()->MorphCurveId, GetProgress()));
         }
+
+        if (GetCreateProperties()->ScaleCurveId)
+        {
```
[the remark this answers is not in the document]
Seyden: Because it’s more readable in my opinion for calculations

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
Traesh: Maybe allow it as optional then, with GetTimeSinceCreated as default

Shauren:
* OverrideScaleCurve is exclusive with ScaleCurveId
* Radius spellmod must be evaluated only once (to support mods with charges) and set in OverrideScaleCurve to make the areatrigger visually larger on client
* ScaleCurve progress can be evaluated even if not active

I'm going to close this PR as I've been working on improving areatrigger code

PR #28540, opened 2022-11-16
Scripts/SepulcherOfTheFirstOnes: Implemented Anduin Wrynn

AlexIndex:
Anduin Raid Encounter WIP
https://youtu.be/KdcKCYdid1w
Update:
https://youtu.be/Zf1CW3V2Qck?si=oonk0J0seFoUx8aE

Assisting NPCs
- [x] Sylvanas AI
- [x] Jaina AI
- [x] Uther AI

Conversations:
- [x] Introduction pre-fight
- [X] Arthas-Uther Conversation
- [X] Arthas-Sylvanas Conversation
- [x] Outroduction
- [X] Cinematic
- [x] Chest and Bridges

Phase 1:
- [X] Energize Willpower for LFR
- [X] Dark Zeal
- [X] Hopebreaker
- [X] Domination Word: Pain
- [x] Befouled Barrier - Requires Core Support for AT. (To be the done among the last)
- [x] Befoulment
- [X] Blasphemy 💀
- [X] Kingsmourne Hungers
- [X] Empty Vessel
- [X] Lost Soul's AI 💀
- [X] Banish Soul 💀
- [X] Wicked Star
- [X] Anduin's Hope AI
- [X] Anduin's Doubt AI
- [x] Anduin's Despair AI
- [x] Rain of Despair (Targets players outside the phase)
- [X] Force of Will

Phase Intermission:
- [X] Remnant of a Fallen King's AI
- [X] Dark Presence
- [X] Army of the Dead
- [X] Soul Reaper
- [X] Remorseless Winter
- [x] March of the Damned (needs fixing orientation change, areatrigger sniff)
- [X] Monstrous Soul's AI
- [X] Necrotic Detonation
- [X] Unraveling Frenzy
- [x] Fiendish Soul's AI
- [x] Necrotic Claws (Leap not working yet)
- [X] Soul Explosion

Phase 2:
- [X] Grim Reflections' AI
- [X] Grim Fate 💀

Phase 3:
- [x] Beacon of Hope - (Not working Mythic shrink of AT and Fragment launch)
- [x] Fragment of Hope
- [X] Purging Light
- [X] Empowered Hopebreaker
- [X] Hopelessness
- [x] Empowered Wicked Star 💀

Additional Things to be added:
Anduin's Soul AI - Release Players from Kingsmourne when adds are healed/ slain except Mythic

PR #29215, opened 2023-08-04
Add EventMap to AreaTriggerAI

Traesh:
**Changes proposed:**

-  Add EventMap & multiple "Pass parameters between AI" methods on AreatriggerAI
-  Main use would be for AT with overrideCurve who need to update every x ms (example : at_anduin_wrynn_befouled_barrier  in  [#28540](https://github.com/TrinityCore/TrinityCore/pull/28540)) to avoid relying on old-school timers
- Some (rare) AT also pulse spell periodically which would be easier with EventMap

**Issues addressed:**

None

**Tests performed:**

Build, tested ingame

### DIFF
--- a/src/server/game/AI/CoreAI/AreaTriggerAI.cpp
+++ b/src/server/game/AI/CoreAI/AreaTriggerAI.cpp
@@ -26,3 +26,11 @@ AreaTriggerAI::AreaTriggerAI(AreaTrigger* a, uint32 scriptId) : _scriptId(script
 AreaTriggerAI::~AreaTriggerAI()
 {
 }
+
+void AreaTriggerAI::OnUpdate(uint32 diff)
+{
+    events.Update(diff);
+
+    while (uint32 eventId = events.ExecuteEvent())
+        ExecuteEvent(eventId);
+}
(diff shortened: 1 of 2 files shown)


Ovahlord:
private and protected members usually follow the  _memberName style, (see the bossAI eventmap member for reference)

Traesh:
> private and protected members usually follow the _memberName style, (see the bossAI eventmap member for reference)

I actually did take BossAI as reference, the EventMap has no _ : https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/AI/ScriptedAI/ScriptedCreature.h#L350

Nyr97:
sadly some parts of the code are still not up-to-date style-wise >:

Seyden:
I honestly don’t see the use for EventMaps, if one needs it inside their script, just add it as a member of the script, it’s not like every at script needs a eventmap

Nyr97:
I too argued it would be fine to have EventMap on every ScriptedAI as a default option (since only BossAI does) and they didn't like it.

BertiRean:
Completely un needed, maybe the Hooks for pass information will be nice, but areatriggers mostly cases do just small actions.

PR #29238, opened 2023-08-15
GameObject/NewFlag: Few changes

Golrag:
**Changes proposed:**

- Expose FlagState to GameObject
- Add and expose FlagCarrier ObjectGuid
- Prevent calling ZoneScript hook on same FlagState
- Prevent clicking flag at same time/during lag to call zonescript hook multiple times and casting spells etc
- Dropped flag should only spawn when debuff is not removed by default (script effect)

**Issues addressed:**

Closes: none, needed for Warsong Gulch.

**Tests performed:**

- Tested on local warsong gulch
- Flag Drop does not happen on capture flag

- [x] Should also check into that lag/multiple people at same time click for dropped flag

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -511,6 +511,9 @@ public:
 
     void SetState(FlagState newState, Player* player)
     {
+        if (_state == newState)
+            return;
+
         FlagState oldState = _state;
         _state = newState;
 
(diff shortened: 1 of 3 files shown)


Golrag:
Maybe we could expose `GameObjectType::NewFlag` to more than just the `GameObject` class and provide some casting stuff like we have for worldobject (talking about ToCreature, ToGameObject etc)?

PR #29246, opened 2023-08-19
Core/Achievements: Allow updating criteria trees with a criteria that has correct map id

Golrag:
**Changes proposed:**

-  Allow criteria tree's that have a criteria with map id to be updated even when achievement instance id check failed
- Previously some achievements had a different map id (example: old arathi, old warsong) which is why some of these achievements wouldn't work

**Issues addressed:**

Closes: none? Achievements in battlegrounds like Warsong Gulch Perfection work now

**Tests performed:**

- Builds
- Tested on Warsong Gulch local branch
  - Warsong Gulch Perfection works now
  -  Warsong Expedience works now

- This might break some achievements? I'm not sure.


Golrag:
This does not fix achievements like Quick Cap (capture flag in 75 seconds). So I'm doubting if this is the right approach now. The asset used here is worldstate (capture flag)

Shauren:
Will be fixed in my next commit

PR #29252, opened 2023-08-20
Core/AreaTriggerActions: Define actions and provide function to call them

Golrag:
The idea is to provide a function that is the entry point for every area trigger action. This could later be used for example in smart_scripts. But also c++ scripts (see my [capture flag pr](https://github.com/TrinityCore/TrinityCore/pull/29237) and [wsg pr](https://github.com/TrinityCore/TrinityCore/pull/29247))

Actions will not have any implementation in this PR. That should be done in different ones.

**Changes proposed:**

-  Define AreaTriggerActions from [enumerated strings](https://wow.tools/dbc/?dbc=enumeratedstring&build=2.5.1.39640#page=1&colFilter[4]=746%20)
-  Provide function that can be called

**Issues addressed:**

Closes: None

**Tests performed:**

- Builds locally

- [ ] Maybe we could define a better way to pass down the params.

### DIFF
--- a/src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
+++ b/src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
@@ -22,6 +22,7 @@
 #include "EnumFlag.h"
 #include "ObjectGuid.h"
 #include "Optional.h"
+#include "SharedDefines.h"
 #include "SpawnData.h"
 #include <variant>
 #include <vector>
@@ -30,8 +31,6 @@
 #define MAX_AREATRIGGER_SCALE 7
 
 enum class AreaTriggerFlag : uint32
-enum TeamId;
-
 {
     None                           = 0x00,
     IsServerSide                   = 0x01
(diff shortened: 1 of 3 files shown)


src/server/game/Entities/AreaTrigger/AreaTriggerTemplate.h
```diff
@@ -64,6 +65,72 @@ enum AreaTriggerActionTypes
...
+    // ApplyAuraAndKeepRetrying
+    Optional<uint32> SpellID;
+
+    // SetScale
+    // IncrementScale
+    // DecrementScale
+    Optional<float> Scale;
+
+    // ActivateGameObject
+    Optional<ObjectGuid> Guid;
```
Nyr97, line 128: Can we add an Optional<std::String> StringId? Maybe we want to look for String as well.

Golrag: Yes, sounds like a good idea! I'll add that one after AreaTrigger refactor has been done

Aokromes:
Conflicting files
src/server/game/Entities/AreaTrigger/AreaTrigger.h

PR #29280, opened 2023-08-23
Save enter/exit areatrigger info in Unit

Traesh:
**Changes proposed:**

We currently have a flaw in the AT enter/exit implementation, because if two same AT overlap and a player exit one while still being in the other, we still call OnUnitExit and potentialy remove an aura that the player should still have because he's still in the second AT. 

This PR allow to check in OnUnitEnter/Exit if the player was already in, or still in, an identical AT before applying/removing the corresponding aura (or any other action).

It also allow to easily check if an Unit is inside an AT, which is used by Blizzard on fights like Dargrul The Underking in the Neltharion's Lair. An AT is spawned behind the crystal pillars and rotate following Dargrul, if you're inside when he cast his magmawave you're safe.

**Issues addressed:**

None

**Tests performed:**

Build, tested ingame

- Currently scripters would have to manually check in OnUnitExit if the unit is still in an identical AT, we could set a parameter in the hook but that would be a major breaking change.

### DIFF
--- a/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
+++ b/src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
@@ -641,6 +641,8 @@ void AreaTrigger::HandleUnitEnterExit(std::vector<Unit*> const& newTargetList)
         DoActions(unit);
 
         _ai->OnUnitEnter(unit);
+
+        unit->EnterAreaTrigger(GetEntry(), GetGUID());
     }
 
     for (ObjectGuid const& exitUnitGuid : exitUnits)
@@ -654,6 +656,8 @@ void AreaTrigger::HandleUnitEnterExit(std::vector<Unit*> const& newTargetList)
 
                 player->UpdateQuestObjectiveProgress(QUEST_OBJECTIVE_AREA_TRIGGER_EXIT, GetEntry(), 1);
             }
+
+            leavingUnit->ExitAreaTrigger(GetEntry(), GetGUID());
 
             UndoActions(leavingUnit);
 
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -5265,6 +5265,24 @@ void Unit::RemoveAllAreaTriggers()
         m_areaTrigger.front()->Remove();
 }
 
+void Unit::EnterAreaTrigger(uint32 areaTriggerId, ObjectGuid areaTriggerGUID)
+{
+    _insideAreaTriggers[areaTriggerId].insert(areaTriggerGUID);
+}
+
+void Unit::ExitAreaTrigger(uint32 areaTriggerId, ObjectGuid areaTriggerGUID)
+{
+    _insideAreaTriggers[areaTriggerId].erase(areaTriggerGUID);
+
+    if (_insideAreaTriggers[areaTriggerId].size() == 0)
+        _insideAreaTriggers.erase(areaTriggerId);
+}
+
+bool Unit::IsInAreaTrigger(uint32 areaTriggerId)
+{
+    return _insideAreaTriggers.find(areaTriggerId) != _insideAreaTriggers.end();
+}
+
 void Unit::SendSpellNonMeleeDamageLog(SpellNonMeleeDamage const* log)
 {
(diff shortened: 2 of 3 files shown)


Nyr97:
I don't think it's a flaw, it's just how the hook works. This can be easily prevented by checking an aura's modunitGUID to prevent a different areatrigger to remove, apply or reapply the same thing.

Traesh:
Not really sure of what's your referring to when you say "modunitGUID" but i don't really see how we can handle this even for Auras without this PR.

How would you handle a scenario like this ? (check out this beautiful drawing)
eozrP93
 
1) Player enter A (aura applied) then B (aura re-applied). B disappear first, aura is removed from player while he's still in A.
2) Same but this time A disappear first.

Anyway as you said even if your aura method work, anything else would still require to store infos in the Unit, hence this PR.

Seyden:
I also dont see any other way except registering the responsible AreaTriggers to the Aura.

So either every AT which would trigger the aura, really triggers it and on "reapply" it kinda says: Hello i'm also here and still alive and when the AT calls RemoveAura, it first unregisters itself from the aura and if there is no responsible worldobject left that it can depend on, the aura really vanishes.. kinda like a stack.. just more like a AreaTrigger responsible or caster stack.

Or the 2nd solution is basically what you implemented. Storing it in the Unit.

If i'm being honest, i like the approach more of just calling CastSpell or RemoveAura with the AreaTrigger Guid and that it handles the responsibility via if the aura still has a alive responsibilty holder or not.

Nyr97:
just like seyden said, for auras we can easily check who's the owner of the aura (modownerunit) by passing guid argument in hasAura, do not remove if not mine, but this creates a new issue where only the at that cast it can really remove it and therefore if a is gone first, it will remove it even if b is there

mdX7:
Using a stack wont work if an at invalidates in the middle of the stack. 
In general pr is a nice addition to scripts which I also recently thought about, but i forgot which at i had in mind at the time. Good idea :+1:

Seyden:
> Using a stack wont work if an at invalidates in the middle of the stack. We could use a counter instead or just provide more data we have sccess to like in this case. In general pr is a nice addition to scripts which I also recently thought about, but i forgot which at i had in mind at the time. Good idea 👍

You missed the point, i didn't mean stacks as in stacks.. i meant more like a "owner list"

AreaTrigger 1 spawns:
- Player 1 enters
  - Aura applied
    - AreaTrigger 1 registers itself as one of the owners in the Aura

AreaTrigger 2 spawns:
- Player 1 enters
  - Aura is "refreshed" due to it already existing
    - AreaTrigger 2 registers itself as one of the owners in the Aura

AreaTrigger 1 despawns:
- Player 1 leaves
  - AT calls RemoveAura with itself as owner information
    - Aura unregisters AreaTrigger 1 as owner
      - Aura checks if there are still owners left
        - if yes: Dont do anything
        - if no: Remove the aura

Traesh:
That's actually a good idea, but i think we still need the info in the Unit anyway.

As stated in the previous messages your solution is great for Auras, but AT doesn't handle only auras.

Some for example just add a MovementForce
Some cast a spell and (if we take my previous drawing) you don't get a new cast when you move from A to B, only when you truly exit and come back inside.
Some scripters may want to know if an unit is currently in an AreaTrigger (fe blizz use this to check los on Dargrul https://youtu.be/dXoh4Wx4n2M)
Etc...

BertiRean:
This is solved using areatrigger neighbour system instead another hook. If they're in a neighbour areatrigger with same data, you should not remove the aura.

Nyr97:
Well, your explanation is falling short a bit: a neighbor system allows you to constantly check whether the unit is also inside of a different atGUID with the same data (either from _template or _properties). This means that you're forced to relieve at(B) from handling anything related to the unit for as long as the unit is inside at(A). However, this also means that you need to be able to send this information from one at to the other: if at(A) despawns, but you're still inside at(B), the info should be passed onto at(B).

This is a bit complicated since we need to manipulate several aspects of how the behavior is handled.

BertiRean:
The message about despawns is easy to implement, because you notify when the area is despawned to other areas with same spell.

Shauren:
26fc14419142e3188370035ce4144cdb125bec05

PR #29281, opened 2023-08-24
Core/Creature: Allow creature table to override npc- and unitflags to 0

mdX7:
**Tests performed:**
tested ingame


Ovahlord:
but why? Usually the creature tables is being used to override the default flags of the template/difficulty so why doing it the other way arround this time?

mdX7:
Logic isn't changed, its only allowing 0 value to override whatever is set in template. Why? Template values in general are mostly the most shared values. So if i have 1000 spawned npcs with gossip npcflag and i require exactly one who should have npcflags 0 we can now do it after pr easily. Pre pr we would've to 0 the template and set gossip on all 1000 creatures just because one creature is a bit off. Or use a script to set npcflags on JustAppeared(). Was a fictional example, fyi

At the end of the day this pr allows 0 as valid value which wasnt possible before

mdX7:
dont have a precise example, but when looking at npcflags value 0 in creature table i'd assume it use 0 and not override from template :P

Shauren (approved):

PR #29282, opened 2023-08-24
Core/Movement: fixed an edge case between hover offsets and polygon selection during pathfinding

Ovahlord:
**Changes proposed:**

-  some units are moving above the ground by using hover movement. This can cause the unit to hover above 7.0yards which is the hardcoded distance limit at which the path generator determines that our path is incomplete due to being too far away from a polgyon. However, hovering units are technically still grounded so we are going to treat them like this and take their hover offset into consideration when checking for the distance between start point and polygon.

- this fixes all edge cases which lead to PATHFIND_INCOMPLETE being disabled in the ChaseMovementGenerator so it can be eneabled as well.

**Tests performed:**
- tested on 4.x

### DIFF
--- a/src/server/game/Movement/PathGenerator.cpp
+++ b/src/server/game/Movement/PathGenerator.cpp
@@ -209,8 +209,13 @@ void PathGenerator::BuildPolyPath(G3D::Vector3 const& startPos, G3D::Vector3 con
     }
 
     // we may need a better number here
-    bool startFarFromPoly = distToStartPoly > 7.0f;
-    bool endFarFromPoly = distToEndPoly > 7.0f;
+    float allowedPolyDist = 7.0f;
+    if (Unit const* unit = _source->ToUnit())
+        if (unit->IsHovering()) // some hovering units hover above the original 7.0f threshold so we have to take their hover offset into consideration
+            allowedPolyDist = std::max(allowedPolyDist, unit->GetHoverOffset());
+
+    bool startFarFromPoly = distToStartPoly > allowedPolyDist;
+    bool endFarFromPoly = distToEndPoly > allowedPolyDist;
     if (startFarFromPoly || endFarFromPoly)
     {
         TC_LOG_DEBUG("maps.mmaps", "++ BuildPolyPath :: farFromPoly distToStartPoly={:.3f} distToEndPoly={:.3f}", distToStartPoly, distToEndPoly);


src/server/game/Movement/PathGenerator.cpp
```diff
@@ -209,8 +209,13 @@ void PathGenerator::BuildPolyPath(G3D::Vector3 const& startPos, G3D::Vector3 con
     }
 
     // we may need a better number here
-    bool startFarFromPoly = distToStartPoly > 7.0f;
-    bool endFarFromPoly = distToEndPoly > 7.0f;
+    float allowedPolyDist = 7.0f;
+    if (Unit const* unit = _source->ToUnit())
+        if (unit->IsHovering()) // some hovering units hover above the original 7.0f threshold so we have to take their hover offset into consideration
+            allowedPolyDist = std::max(allowedPolyDist, unit->GetHoverOffset());
```
jackpoz, line 215: should this be
```cpp
allowedPolyDist = allowedPolyDist + unit->GetHoverOffset();
```
?

jackpoz: Thinking about it, this is wrong, the Z of the "shadow" on the ground should be passed throughout the whole PathGenerator class for hovering creatures, otherwise for example https://github.com/TrinityCore/TrinityCore/blob/804e0b8dcddce1db8f2a4f6ed0d291cf04e9ab84/src/server/game/Movement/PathGenerator.cpp#L124-L138 will use a hitbox in the air to find the ground poly

Ovahlord: So, what's your idea here?

jackpoz: https://github.com/TrinityCore/TrinityCore/blob/804e0b8dcddce1db8f2a4f6ed0d291cf04e9ab84/src/server/game/Movement/PathGenerator.cpp#L59 should be updated to take hovering creatures into consideration and remove their hovering offset.
Btw do hovering creatures need the hovering offset to be added to every point of the resulting path ?

Ovahlord: They do yes. UpdateAllowedPositionZ does this already when normalizing the path so we bascially have to do a regular path generation with 'un-hovered' positions and be done with it

Ovahlord: About your GetPosition suggestion: I did add a couple overloads to my cata fork where I set the start position without the hover offset, but the farfrompoly issue was still present. I was using a particular boss to test it and there is one thing that we also have to check:   the boss is hovering, but also the bosses' spawn position is below the map surface so he essentially hovers through the ground. So we have to do a vmap raycast here to normalize the Z offset

CraftedRO:
this looks abandoned

Issue #24552, opened 2020-05-03
[Movement] MovementInform is not being called in rare occasions


joshwhedon:
**Description:**

In some very rare occasions, MovementInform does not seem to be called for scripts. This does not seem specific to any motion master.
It has been observed for:
- Whitemane and mograine in scarlet monastery, sometimes the movement inform is not called, therefore she does not resurect mograine
- Lord Jaraxxus in trial of the crusaders, sometimes, his intro movement inform does not trigger, and the event gets stuck
- Wild wyrm in storm peaks, this one might be unrelated, but on fall, when reaching the ground, it is supposed to eject all passengers, sometimes it does not
- Bronjham in the forge of souls ( @Jildor ) 

I am not yet 100% certain MovementInform is the cause, but based on all the scripts the problem occurs in, it is the likely candidate.

**Current behaviour:**

MovementInform is not called

**Expected behaviour:**

MovementInform is called

**Steps to reproduce the problem:**

No clear way to reproduce the issue, it seems to occur fairly randomly, about once a day on a server with 800 players for those scripts. Very probable to occur for other scripts, but has not been reported as it does not block the encounters.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/a548d9f9b021a54cfc9013b7e104ccf524e88358

**Operating system:** Linux

Jildor:
Just for document the case of bronjahm.
When bronjahm summon Corrupted Soul: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/FrozenHalls/ForgeOfSouls/boss_bronjahm.cpp#L134
Corrupted soul do MoveFollow to bronjahm.
in case of Corrupted Soul is summoned next to bronjahm, MovementInform doesn't work and can't cast the spell:
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Northrend/FrozenHalls/ForgeOfSouls/boss_bronjahm.cpp#L241-L251

I don't know if can be related with the issue, but, maybe can help

joshwhedon:
Thanks, after having a look at the code it seems to be a different issue:
In the case of follow movement generator, MovementInform is only called when a movement occured and the new position is at distance of the target:
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Movement/MovementGenerators/FollowMovementGenerator.cpp#L98
[...]

PR #29293, opened 2023-08-29
Scripts/Trial of the Crusader: Update Read and Write Save Data

Jildor:
**Changes proposed:**

-  Update Read and Write Save Data
-  Fixes Lord Jaraxxus stuck case described here #24552
When you enter in the instance, boss states are 0 (NOT_STARTED), if you check instance table in char db, data field is empty
If you do a server restart without do anything in the instance and login again and enter in the instance, boss states are 5 (TO_BE_DECIDED)
So when Lord Jaraxxus is summoned, and do Reset(): https://github.com/TrinityCore/TrinityCore/blob/8e7b6c54e469e70d2a3d817e0f9b9d0ae46cbae6/src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/boss_lord_jaraxxus.cpp#L116-L131
ACTION_JARAXXUS_INTRO can't be executed, so: https://github.com/TrinityCore/TrinityCore/blob/8e7b6c54e469e70d2a3d817e0f9b9d0ae46cbae6/src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/boss_lord_jaraxxus.cpp#L156-L160
EVENT_INTRO don't start and stuck.

**Issues addressed:**

Updates #24552

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/instance_trial_of_the_crusader.cpp
+++ b/src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/instance_trial_of_the_crusader.cpp
@@ -573,57 +573,28 @@ class instance_trial_of_the_crusader : public InstanceMapScript
 
             void Save()
             {
-                OUT_SAVE_INST_DATA;
-
-                std::ostringstream saveStream;
-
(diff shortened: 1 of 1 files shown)


Shauren:
Even if this doesnt fix the issue, its a good move to get rid of the legacy saving code

PR #29309, opened 2023-09-09
Update barbershop opcodes to 3.4.0

lineagedr:
**Changes proposed:**

-  Update barbershop opcodes to 3.4.0

**Tests performed:**
successfully builds and works in-game

- None

Note: I've noticed that usually their hex values should be close like for example: since enable_barber_shop is 0x26B9 the barber_result should be 0x26BA but apparently 0x26BA is related to some barber shop discount system where it opens a gossip code box asking if you'd like a randomized appearance for a discounted price not sure if it was ever implemented in wotlk classic though.


Shauren:
No longer relevant, 3.4.3 is out and will be the target version of this branch

PR #29319, opened 2023-09-12
Refactors

Ovahlord:
**Changes proposed:**

-  refactored corpse location getter to return by reference instead of value to reduce uncessary copies and marked them as const in the process to prevent unwanted meddling with it
-  refactored locale detection to silence a ill-defined for loop warning (C6293)
-  added underlying types to world configs to silence some other analysis warnings

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -4512,8 +4512,7 @@ void Player::BuildPlayerRepop()
     // there must be SMSG.STOP_MIRROR_TIMER
 
     // the player cannot have a corpse already on current map, only bones which are not returned by GetCorpse
-    WorldLocation corpseLocation = GetCorpseLocation();
-    if (corpseLocation.GetMapId() == GetMapId())
+    if (GetCorpseLocation().GetMapId() == GetMapId())
     {
         TC_LOG_ERROR("entities.player", "Player::BuildPlayerRepop: Player '{}' ({}) already has a corpse", GetName(), GetGUID().ToString());
         return;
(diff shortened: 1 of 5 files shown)


Warpten:
The worldlocation stuff barely matters, it's effectively just PODs so a copy is trivial. You're just saving a few bytes off the stack for an indirection. /shrug

Ovahlord:
> The worldlocation stuff barely matters, it's effectively just PODs so a copy is trivial. You're just saving a few bytes off the stack for an indirection. You can just return a const copy if your concern is mutating it, but that will probably run into issues for relocation script hooks /shrug

yeah, visual studio doesn't like it though and marks it as guideline warning so for the sake of good habits

Shauren:
* That enum underlying type - why? I am not seeing any warnings in VS related to that
* locale detection - you are unintentionally changing the logic making the last existing locale (on the enum) "default" instead of the first one
* worldlocation - sure whatever, can merge

Shauren (changes requested):
* That enum underlying type - why? I am not seeing any warnings in VS related to that
* locale detection - you are unintentionally changing the logic making the last existing locale (on the enum) "default" instead of the first one
* worldlocation - sure whatever, can merge

Ovahlord:
> * That enum underlying type - why? I am not seeing any warnings in VS related to that
> * locale detection - you are unintentionally changing the logic making the last existing locale (on the enum) "default" instead of the first one
> * worldlocation - sure whatever, can merge

- The underlying type thing is a code quality check warning where it complains about arrays using signed enum values for indexing. That's fixed by adding an unsigned underlying type to enums

PR #29324, opened 2023-09-14
Scripts/Mardum: Implement Quest "Enter the Illidari: Ashtongue"

Naddley:
**Changes proposed:**

-  Script Quest "Enter the Illidari: Ashtongue" ID: 40378

**Issues addressed:**

None

**Tests performed:**

build, tested ingame

- [ ] DH´s on retail have Riding Skill 100% Ground Mounts after creation (level 8) which is missing on TC.


src/server/scripts/BrokenIsles/zone_mardum.cpp
```diff
@@ -622,6 +625,234 @@ struct npc_illidari_fighting_invasion_begins : public ScriptedAI
...
+            Unit* korvas = ObjectAccessor::GetUnit(*me, korvasGuid);
+            if (!korvas)
+                return;
+
+            Talk(SAY_KAYN_ACTIVATE_GATEWAY, me);
+            me->CastSpell(privateObjectOwner, SPELL_TRACK_TARGET_IN_CHANNEL, false);
+            korvas->CastSpell(privateObjectOwner, SPELL_TRACK_TARGET_IN_CHANNEL, false);
+        });
+
+        _scheduler.Schedule(7s, [this](TaskContext /*task*/)
```
Nyr97, line 708: if there is a lag, all of this is desynchronized, you must schedule every action after the next, not altogether

src/server/scripts/BrokenIsles/zone_mardum.cpp
```diff
@@ -622,6 +625,234 @@ struct npc_illidari_fighting_invasion_begins : public ScriptedAI
...
+    npc_sevis_brightflame_ashtongue_gateway_private(Creature* creature) : ScriptedAI(creature) { }
+
+    void JustAppeared() override
+    {
+        _scheduler.Schedule(1s, [this](TaskContext /*task*/)
+        {
+            Talk(SAY_SEVIS_SAY_FIND_ALLARI, me);
+        });
+
+        _scheduler.Schedule(3s, [this](TaskContext /*task*/)
```
Nyr97, line 799: same this here as the above, my only problem with scheduler is each lambda must be inside the previous and it creates a lot of indentation if multiple are required

PR #29333, opened 2023-09-18
Core/Quests: Ignore update of optional objectives if quest is completed

meji46:
**Changes proposed:**

-  Ignore update of optional objectives if quest is completed
-  This logic (https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Entities/Player/Player.cpp#L16655) causes that a completed quest is flagged as incomplete when processing an optional objective (Player::CanCompleteQuest returns false as quest is already completed)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game ([Quest 8488](https://www.wowhead.com/quest=8488/unexpected-results))

None

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -16546,7 +16546,7 @@ void Player::UpdateQuestObjectiveProgress(QuestObjectiveType objectiveType, int3
             continue;
 
         // Ignore update of optional objectives if quest is completed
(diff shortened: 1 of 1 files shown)


mdX7:
maybe we should actually move it to `Player::IsQuestObjectiveCompletable` directly?

Shauren:
Fixed differently in 4362a8c3f6aff1a8403ab9c1dee3fd23a82db339

PR #29341, opened 2023-09-28
Fix CMakeLists for tools-only build

kerhong:
**Issues addressed:**

`tools` depend on `short_alloc`, but `short_alloc` is only built if `-DSERVERS=1`.

```sh
# with current TrinityCore 3.3.5
$ mkdir build
$ cd build
$ cmake -S .. -B . -DTOOLS=1 -DSERVERS=0 -DSCRIPTS=none
$ make
... lots of lines removed
/usr/bin/ld: cannot find -lshort_alloc: No such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:136: all] Error 2
```

This PR builds `short_alloc` with both `-DTOOLS=1` and `-DSERVERS=1`

**Tests performed:**

Verified that it builds.


Shauren:
Ah damn I forgot to cherry-pick 592e01cf7a5d7b6f7ec63d30edc09dd9d4062b62 to 3.3.5

PR #29344, opened 2023-10-01
Billy

billyplus:
`isTrivial` means the level of player is greater then the level of quest plus the value of LOW_LEVEL_HIDE_DIFF


Shauren:
This PR has no changes

PR #29349, opened 2023-10-03
Core/GameObject/ControlZone: Update spawn conditions when new event is triggered

Golrag:
**Changes proposed:**

-  update the spawn conditions when the control zone fires an event

**Issues addressed:**

Closes: none

**Tests performed:**

- Builds
- Tested with local changes for eye of the storm

- None


Faq:
What happens when spawn conditions not updated?

Golrag:
They do get updated, but not instantly. We need to update them instantly so spirit guides despawn and players get teleported

PR #29355, opened 2023-10-07
Build: added support for Unity builds in all projects except for scripts

Ovahlord:
**Changes proposed:**

-  enable the unity build functionality via cmake option which allows us to merge multiple source files into a single unity files to speed up build time.

**Tests performed:**
- builds. build time is effectively cut in half

**Todo:**
- the scripts project cannot be built with unity building right now due to its unscoped enum mess which needs a big refactor to work properly


jackpoz:
how does this affect debugging symbols and crashlog lines ?

Ovahlord:
> how does this affect debugging symbols and crashlog lines ?

image
 just fine.

Ovahlord:
closing this for now as I'm gonna do an overhaul one day where scripts will be moved into namespaces so we can buil game and scripts

PR #29370, opened 2023-10-15
Core/Map: Implement several difficulty getters

Nyr97:
**Changes proposed:**

-  Implemented some map mode getters: IsMythic, IsHeroicOrMythic, and IsTimewalking. These are usable both in dungeon, raid, expedition islands, and warfronts maps since parent getters check for map flags directly. Also, these should be helpful to scripters in the future to avoid overcrowding the code with GetDifficultyID and enumerates.
- Updated a script with one of them.

**Issues addressed:**

None.

**Tests performed:**

It was tested and it builds in-game.

None.

### DIFF
--- a/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
+++ b/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
@@ -303,6 +303,41 @@ bool ScriptedAI::HealthAbovePct(uint32 pct) const
     return me->HealthAbovePct(pct);
 }
 
+bool ScriptedAI::IsLFR() const
+{
+    return me->GetMap()->IsLFR();
+}
+
+bool ScriptedAI::IsNormal() const
+{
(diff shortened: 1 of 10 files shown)


Nyr97:
Added IsMythicPlus (we may rename it to IsMythicKeystone instead to keep Blizzard's original name) since DIFFICULTY_MYTHIC_KEYSTONE does not have DIFFICULTY_FLAG_DISPLAY_MYTHIC flag, so it falls outside of the checker. Also, there are old rotational dungeons being reused for current mythic seasons that have different spawns and mechanics depending on Mythic or MythicPlus (for example, Grimrail Depot and in-coming Darkheart Thicket, Black Rook Hold, Everbloom for season 3 in Dragonflight, etc).

Nyr97:
Also added IsHeroicOrHigher to avoid having to call three different getters.

PR #29395, opened 2023-10-27
Core/Creatures: Allow interaction with trigger creatures using spellclick

meji46:
**Changes proposed:**

-  Allow interaction with trigger creatures using spellclick. There are a lot of trigger creatures with visual auras to emulate interactive objects (spellclick)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game

<details>
  <summary>Example</summary>

(.gm on)
image
(.gm off)
image

</details>

None

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -655,8 +655,8 @@ bool Creature::UpdateEntry(uint32 entry, CreatureData const* data /*= nullptr*/,
         }
     }
 
-    // trigger creature is always uninteractible and can not be attacked
-    if (IsTrigger())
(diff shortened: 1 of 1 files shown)


Nyr97:
this is a big change, have we gotten any results how many are there without the flag? we might be breaking some quests

Ovahlord:
At this point, it's worth considering to not use the trigger special flag at all for these npcs. The trigger flag used to be used to select invisible display ids and use a passive react state which is now handled by model probabilities and static flags. So with that being said, the appropriate fix here would be just unflagging the npc as trigger so it's handled like any other creature.

meji46:
Yep, we only lose the possibility to see the dummy model with gm on. These creatures use a visual aura, so it doesn't matter

PR #29401, opened 2023-10-29
Upgrade (dep): zlib package to 1.3.0.1

acidmanifesto:
**Changes proposed:**
Upgrade zlib package to 1.3.0.1 to ensure a few CVEs were patched within the library.

**Issues addressed:**
[CVE-2022-37434](https://nvd.nist.gov/vuln/detail/CVE-2022-37434)

**Tests performed:**

Cmake runs fine, Compile completes (VS 2022). 
Binaries run with no issue.

- [ ] 
- [ ]

### DIFF
--- a/dep/PackageList.txt
+++ b/dep/PackageList.txt
@@ -42,7 +42,7 @@ utf8-cpp (UTF-8 with C++ in a Portable Way)
 
 zlib (A Massively Spiffy Yet Delicately Unobtrusive Compression Library)
   http://www.zlib.net/
-  Version: 1.3.0.1
+  Version: 1.3
 
 gSOAP (a portable development toolkit for C and C++ XML Web services and XML data bindings)
   http://gsoap2.sourceforge.net/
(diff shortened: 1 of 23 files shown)


Faq:
dublicate of https://github.com/TrinityCore/TrinityCore/pull/29401

mdX7:
> dublicate of #29402

Its pr to different branch

acidmanifesto:
Given it has a NVD score of 9.8 out of 10, it was best to do it for each of the branches and see what gets merged first, wotlk classic branch can just cherry pick from master.

Shauren:
meh, TC is unaffected (read the CVE, only applications using inflateGetHeader are vulnerable, TC does not)

acidmanifesto:
I still feel that upgrading deps when capable and compatible is still worth the effort. Improvements not including the cve mention are still worth considering regardless of whether it is minor, major, or not applicable.




Although we are on 1.2.12, what concerns me is that this is noted here:

Due to the first bug fix, any installations of 1.2.12 or earlier should be replaced with 1.2.13.

Shauren:
Yes, exactly
> Due to the first bug fix...

First here being `inflateGetHeader` which I already explained that its not used anywhere in TC

This doesn't mean I'm against the upgrade, just that I suggest carefully evaluating changes instead of jumping on "CVE fix, must upgrade asap"

PR #29409, opened 2023-11-05
Core/Spells: implement spell queue

Ovahlord:
Players can now queue up their spell and item cast requests a couple of milliseconds before the global cooldown expires or previous spell cast launches which then gets executed right away

**Changes proposed:**

-  adding a queue to store a spell cast request that has been made before an ongoing cast or gcd has expired, offering a seamless and more responsive casting experience

**Tests performed:**
- builds, tested ingame, long-term tested on 4.x

- Pet casts are currently not supported as they are essentially using their own cast request bracket and since pet spell casts are kinda poorly implemented atm, it's not worth wasting time with them.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -1249,6 +1249,9 @@ void Player::setDeathState(DeathState s)
             return;
         }
 
+        // clear all pending spell cast requests when dying
+        CancelPendingCastRequest();
+
         // drunken state is cleared on death
         SetDrunkValue(0);
         SetPower(POWER_COMBO_POINTS, 0);
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -29904,13 +29904,14 @@ void Player::ExecutePendingSpellCastRequest()
 
     // check known spell or raid marker spell (which not requires player to know it)
     SpellInfo const* spellInfo = sSpellMgr->AssertSpellInfo(_pendingSpellCastRequest->CastRequest.SpellID, GetMap()->GetDifficultyID());
-    if (castingUnit->IsPlayer() && !castingUnit->ToPlayer()->HasActiveSpell(spellInfo->Id) && !spellInfo->HasAttribute(SPELL_ATTR8_SKIP_IS_KNOWN_CHECK))
+    Player* plrCaster = castingUnit->ToPlayer();
+    if (plrCaster && !plrCaster->HasActiveSpell(spellInfo->Id) && !spellInfo->HasAttribute(SPELL_ATTR8_SKIP_IS_KNOWN_CHECK))
     {
         bool allow = false;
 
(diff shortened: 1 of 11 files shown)


Nyr97:
Why would we support this if this was made to prevent high-latency users from having a bad experience gameplay-wise? Locally we have no lag whatsoever, and if I recall correctly, you rejected this specific feature for 4.3.4 branch some time ago when someone else requested or tried to implement it themselves.

heyitsbench:
Maybe embed it in a config? For development/personal servers/QoL, turn the system off, for the sake of accuracy turn it on?

Ovahlord:
> Why would we support this if Blizzard implemented this feature to ONLY prevent high-latency users from having a bad experience gameplay-wise? Locally we have no lag whatsoever. Also, if I recall correctly, you rejected this specific feature for 4.3.4 branch some time ago when someone else requested or tried to implement it themselves for that same reason.

The Client supports it, so we do so as well. That's it.

acidmanifesto:
> Also, if I recall correctly, you rejected this specific feature for 4.3.4 branch some time ago when someone else requested or tried to implement it themselves for that same reason.


Its been in tcpp for a good while.
https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/commit/4430863a1e9a787a3a702b58adaf4c7bb639fa29

Ovahlord:
> Maybe embed it in a config? For development/personal servers/QoL, turn the system off, for the sake of accuracy turn it on?

There is no reason to turn it off at all because all it does is storing packet information until it can be used and canceling spell casts as well, as the /cancelqueuedspell macro can interrupt it.

Nyr97:
> > Why would we support this if Blizzard implemented this feature to ONLY prevent high-latency users from having a bad experience gameplay-wise? Locally we have no lag whatsoever. Also, if I recall correctly, you rejected this specific feature for 4.3.4 branch some time ago when someone else requested or tried to implement it themselves for that same reason.
> 
> 
> 
> The Client supports it, so we do so as well. That's it.

Well,,, https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/issues/224 this states otherwise 👀

Ovahlord:
> > > Why would we support this if Blizzard implemented this feature to ONLY prevent high-latency users from having a bad experience gameplay-wise? Locally we have no lag whatsoever. Also, if I recall correctly, you rejected this specific feature for 4.3.4 branch some time ago when someone else requested or tried to implement it themselves for that same reason.
> > 
> > 
> > The Client supports it, so we do so as well. That's it.
> 
> Well,,, [The-Cataclysm-Preservation-Project/TrinityCore#224](https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/issues/224) this states otherwise 👀

are you here to keep bitching about it or do you want to contribute some actual feedback like testing this stuff?

Nyr97:
Both of them are fitting, but the later I'm doing currently.

mdX7 (approved):

PR #29410, opened 2023-11-05
Core/BattlegroundMap: Load all cells when created & mark all creatures as active

Golrag:
**Changes proposed:**

-  Make sure all cells are loaded for a BattlegroundMap
-  Make all creatures active on a BattlegroundMap when added
-  

**Issues addressed:**

Closes: None?

**Tests performed:**

- Implemented OnCreatureCreate hook from battleground, with a creature at other end of the map in alterac valley. Hook got called with creature
- Tested with my alterac valley changes

- [ ] Maybe we could do the same for gameobjects?


Nyr97:
I'm not sure this is... correct? This would mean we would need to set all instanced maps as active and load all their grids?

Golrag:
The issue im trying to solve in battlegrounds is that currently not all creatures are loaded, which for alterac means, there is no text that can be yelled by the herald. These herald npcs are located in the middle and are not loaded right away when there are no players nearby activating the grid.

I think for battlegrounds/arenas (and maybe even raids/dungeons) we should load everything and make sure everything stays loaded. See the CapturePoints, NewFlag type gameobjects, these also need to be active for their internal timer

jackpoz:
I would like the loadCells bit to have a config, enabled by default.

As for the setActive, I wonder if we can just setActive the whole map instead of each creature ? Which solution looks cleaner ?
In any case, please move the condition `if (GetMap()->IsBattlegroundOrArena())` to a new method, something like `IsAlwaysActive()`, so it will be easier to add more conditions in the future

src/server/game/World/World.cpp
```diff
@@ -803,6 +803,12 @@ void World::LoadConfigSettings(bool reload)
         TC_LOG_ERROR("server.loading", "InstanceMapLoadAllGrids enabled, but GridUnload also enabled. GridUnload must be disabled to enable instance map pre-loading. Instance map pre-loading disabled");
         m_bool_configs[CONFIG_INSTANCEMAP_LOAD_GRIDS] = false;
     }
+    m_bool_configs[CONFIG_BATTLEGROUNDMAP_LOAD_GRIDS] = sConfigMgr->GetBoolDefault("BattlegroundMapLoadAllGrids", true);
+    if (m_bool_configs[CONFIG_BATTLEGROUNDMAP_LOAD_GRIDS] && m_bool_configs[CONFIG_GRID_UNLOAD])
+    {
+        TC_LOG_ERROR("server.loading", "BattlegroundMapLoadAllGrids enabled, but GridUnload also enabled. GridUnload must be disabled to enable battleground map pre-loading. Battleground map pre-loading disabled");
```
jackpoz, line 809: We could just ignore GridUnload for battlegrounds when this new config is enabled

Golrag: We can just remove this piece of code here and it will work right? As creatures will be added automatically and keeping the grid active

jackpoz:
Looks good enough. Are BG maps closed and unloaded correctly still ? No crash on shutdown with an ongoing BG ?

Golrag:
haven't noticed any crashes, BattlegroundMap destructor gets called when leaving ongoing battleground (and being the only player that was in it)

jackpoz (approved):

PR #29420, opened 2023-11-07
Core/Utilities: Added timer series queuing to EventMap

mdX7:
**Changes proposed:**
- add a new scheduling method which automatically reschedules itself when the event is triggered with specified timers in a queue
- allows us to use trivial timers from dbm / bigwigs or similar instead of speculating about tons of conditions and edge cases per spell

**Tests performed:**
tested ingame with anduin pr

**Example:**
```cpp
events.ScheduleEventSeries(EVENT_WICKED_STAR, { 55s, 35s, 30s });
```
This initially schedules EVENT_WICKED_STAR for 55s,
After 55s have passed it will reschedule itself for 35s, after 35s have passed then it will reschedule for 30s and afterwards it will no longer reschedule, as the series has ended/queue is empty.
Handling may be done as usual via `events.ExecuteEvent()` e.g.:
```cpp
    {
        events.Update(diff);

        [...]

        while (uint32 eventId = events.ExecuteEvent())
        {
            switch (eventId)
            {
                case EVENT_WICKED_STAR:
                {
                    DoCastSelf(SPELL_WICKED_STAR);
                    Talk(SAY_WICKED_STAR);
                    break;
                }
```

### DIFF
--- a/src/common/Utilities/EventMap.cpp
+++ b/src/common/Utilities/EventMap.cpp
@@ -85,6 +85,7 @@ uint32 EventMap::ExecuteEvent()
             uint32 eventId = (itr->second & 0x0000FFFF);
             _lastEvent = itr->second; // include phase/group
             _eventMap.erase(itr);
+            ScheduleNextFromSeries(_lastEvent);
             return eventId;
         }
     }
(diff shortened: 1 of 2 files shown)


Nyr97:
This is goochie and I see that if we reach last index in vector, it won't schedule, right?

mdX7:
> This is goochie and I see that if we reach last index in vector, it won't schedule, right?

correct, if the entire timer series is done (as in no more timers left) it wont schedule anything

PR #29421, opened 2023-11-07
Core/Packets: Rename instance of `templat` to `templateInfo`.

heyitsbench:
**Changes proposed:**

-  Rename a usage of CharacterTemplate from `templat` to `templateInfo`. `template` is a C++ keyword, but the latter seems preferable than a misspelt word in my opinion.

**Issues addressed:**

Closes no issue.

**Tests performed:**

Untested.

- None that I am aware of.


heyitsbench:
There is a similar case in [BlackMarketMgr](https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/BlackMarket/BlackMarketMgr.cpp#L200), would maybe be better named `marketTemplate`. If desired, I can make another PR (or add to this one) to adjust that as well.

PR #29433, opened 2023-11-11
Knock Back Spells with only BasePoints can now Knock Back Units

gonzo1247:
**Changes proposed:**

-  Some Spells like 108773 or 93575, 88250 has BasePoint set but no MiscValue.  As a result, we have only given a speedz, but no speedxy. Since the calculation in line 3957 (float speedxy = float(effectInfo->MiscValue) * ratio;) turns speedxy into 0, the KnockBack Effect does not work. 
With this change it will work, even if the unit will move a little bit.

**Tests performed:**

Build and work ingame - testet with a few spells.

- [ ] 
- [ ]

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -3959,6 +3959,9 @@ void Spell::EffectKnockBack()
     if (speedxy < 0.01f && speedz < 0.01f)
         return;
 
+    if (speedxy < 0.01f)
+        speedxy = 0.05f;
+
     Position origin;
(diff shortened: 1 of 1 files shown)


mdX7:
any example spells to reproduce?

meji46:
Negative speedXY is valid. Check https://github.com/TrinityCore/TrinityCore/pull/28081

Shauren:
Awful hack

gonzo1247:
ahh I overlooked that with Negative speedXY. Then I'll close this for now, as it definitely wouldn't work with that. Until a better solution is found, then I would open it again.

PR #28208, opened 2022-08-21
Core/Creatures: Added possibility to set PlayHoverAnim state on spawn

meji46:
**Changes proposed:**

-  Added a new field **PlayHoverAnim**  in the table **creature(_template)_addon** to control this state
- Extended function **SetPlayHoverAnim** to avoid sending SMSG_SET_PLAY_HOVER_ANIM in case of spawning

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

**Known issues and TODO list:** (add/remove lines as needed)

- I'm not sure if the functions **Set*AnimKitId** should also be extended to not send **SMSG_SET_*_ANIM_KIT** in case of spawning

PR #29434, opened 2023-11-12
Core/Units: Update PlayHoverAnim state when flying or hovering

meji46:
**Changes proposed:**

-  Update PlayHoverAnim state when flying or hovering (related to https://github.com/TrinityCore/TrinityCore/pull/28208)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game

None

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -12839,6 +12839,16 @@ bool Unit::SetDisableGravity(bool disable, bool updateAnimTier /*= true*/)
             SetAnimTier(AnimTier::Ground);
     }
 
+    if (IsAlive())
+    {
+        if (IsGravityDisabled() || IsHovering())
+            SetPlayHoverAnim(true);
+        else
+            SetPlayHoverAnim(false);
+    }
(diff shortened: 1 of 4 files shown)


Ovahlord:
This is not correct to do this way because there are a ton of flying units which do not use hover anim at all and merely play flying animations by using the animTier value

meji46:
> This is not correct to do this way because there are a ton of flying units which do not use hover anim at all and merely play flying animations by using the animTier value. Same applies for hovering. Both use animTiers which then also uses the HoverHeight update field. The hover animation opcode / bit field itself is almost exclusivey used for levitating auras or special cases.

As I told you in the other PR, the anthropomorphic creature models use this value to do the anim either while flying or hovering (AnimTier not involved). Besides, all flying creatures have the bit set to true (I've only found one special case for an invisible trigger with root).

I've given several examples for cases with/without auras:
https://github.com/TrinityCore/TrinityCore/pull/28208#issuecomment-1238631848

CraftedRO:
maybe we'll get this merged this year

Nyr97:
So this change means we won't ever use or allow SetPlayHoverAnim scripting-wise, the whole usability is left to SetDisableGravity and SetHover to decide.

PR #29438, opened 2023-11-12
Core/Battleground: Rework BattlegroundScore

Golrag:
**Changes proposed:**

-  Remove subclasses of Battleground Score for each battleground zone
-  Use `PVPStat.db2` and `PVPScoreboardLayout.db2`
-  Add another function in Battleground to update PvpStats
- The only thing that was different for each battleground was the values send in PVPMatchPlayerPVPStat

**Issues addressed:**

Closes: None

**Tests performed:**

- Builds
- Tested in game

- [ ] Some clean up might still be needed


meji46:
Can't we use PVPScoreboardLayout.db2 and PVPStat.db2?

Golrag:
Eh i wasn't aware of it, lemme check if I can convert it

Golrag:
Updated to use these two DB files

mdX7 (approved):

src/server/game/Battlegrounds/BattlegroundScore.h
```diff
@@ -30,67 +33,48 @@ enum ScoreType
...
-        uint8 TeamId; // PvPTeamId
-
-        // Default score, present in every type
-        uint32 KillingBlows;
-        uint32 Deaths;
-        uint32 HonorableKills;
-        uint32 BonusHonor;
-        uint32 DamageDone;
-        uint32 HealingDone;
+    BattlegroundScore(ObjectGuid playerGuid, uint32 team, std::unordered_set<uint32> const* pvpStatIds);
```
BertiRean, line 41: What is the reason to get a pointer and not a const & to pvpStatIds 🤔

src/server/game/Battlegrounds/BattlegroundScore.h
```diff
@@ -30,67 +33,48 @@ enum ScoreType
...
+    virtual ~BattlegroundScore();
+
+    void UpdateScore(uint32 type, uint32 value);
+    void UpdatePvpStat(uint32 pvpStatID, uint32 value);
+
+    // For Logging purpose
+    virtual std::string ToString() const { return ""; }
```
BertiRean, line 48: Maybe do this a pure function will be better in order to force devs to propose a good String Representation of Score for Logging. @mdX7 What do you think ?

src/server/game/DataStores/DB2Stores.cpp
```diff
@@ -1621,6 +1626,10 @@ uint32 DB2Manager::LoadStores(std::string const& dataPath, LocaleConstant defaul
             sOldContinentsNodesMask[field] |= submask;
     }
 
+    for (PVPScoreboardLayoutEntry const* layout : sPVPScoreboardLayoutStore)
```
Shauren, line 1629: Why bother loading the Layout db2 when you are only using data from PvPStat?

Golrag: The idea was to only load the ones that are used in a layout. But this is kinda misleading with the name of the property and they have the same amount of records anyway. Will remove it

PR #29469, opened 2023-11-26
Scripts/AzureVault: Implement Telash Greywing encounter

Naddley:
**Changes proposed:**

-  implement Telash encounter

**Issues addressed:**

None

**Tests performed:**

build, tested ingame


src/server/game/Spells/SpellMgr.cpp
```diff
@@ -4746,6 +4746,24 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->AttributesEx8 |= SPELL_ATTR8_ATTACK_IGNORE_IMMUNE_TO_PC_FLAG;
     });
 
+    // Jump to Center (DNT)
+    ApplySpellFix({ 387981 }, [](SpellInfo* spellInfo)
+    {
+        ApplySpellEffectFix(spellInfo, EFFECT_0, [](SpellEffectInfo* spellEffectInfo)
+        {
+            spellEffectInfo->TargetA = SpellImplicitTargetInfo(TARGET_DEST_DEST);
```
Ovahlord, line 4754: what's the matter with hacking all these targets lately?

mdX7: *some serverside entity

Shauren (changes requested):
Kill all `PrepareSpellScript` and `PrepareAuraScript` uses, they are deprecated and do nothing (kept only for compatibility)

src/server/scripts/DragonIsles/AzureVault/boss_telash_greywing.cpp
```diff
@@ -0,0 +1,428 @@
...
+                        DoCastSelf(SPELL_ACTIVATE_VAULT_RUNE);
+                        events.CancelEvent(EVENT_FROST_BOMBS);
+                        events.CancelEvent(EVENT_ICY_DEVASTATOR);
+                    }
+                    events.Repeat(500ms);
+                    break;
+                default:
+                    break;
+            }
+            if (me->HasUnitState(UNIT_STATE_CASTING))
```
Nyr97, line 185: Out of curiosity, why are we adding this here again? The boss shouldn't be able to melee at all during casting unless any of the spells have "allow while casting" attributes, especially if we set a return for such before ExecuteEvent.

src/server/scripts/DragonIsles/AzureVault/boss_telash_greywing.cpp
```diff
@@ -0,0 +1,428 @@
...
+    }
+};
+
+// 389453 - Ice Power (periodic)
+class spell_telash_ice_power_perodic : public AuraScript
+{
+    void HandlePeriodic(AuraEffect const* /*aurEff*/)
+    {
+        Unit* target = GetTarget();
+        target->SetPower(POWER_ENERGY, target->GetPower(POWER_ENERGY) + 2);
```
Nyr97, line 263: GetPowerType, pls :D

src/server/scripts/DragonIsles/AzureVault/boss_telash_greywing.cpp
```diff
@@ -0,0 +1,428 @@
...
+#include "SpellAuraEffects.h"
+#include "SpellScript.h"
+#include "SpellAuras.h"
+#include "SharedDefines.h"
+#include "azure_vault.h"
+
+enum TelashSpells
+{
+    // Telash
+    SPELL_ICE_POWER_PERIODIC        = 389453,
```
Nyr97, line 38: Rename it to SPELL_POWER_ENERGIZE_ICE_POWER_PERIODIC, to keep the same name convention as the newest implemented bosses? :D

Nyr97 (commented):
Overall pretty good, imo. Just a few notes here and there.

src/server/scripts/DragonIsles/AzureVault/boss_telash_greywing.cpp
mdX7: Castcheck is already done implicitly in `DoMeleeAttackIfReady`.

This check is used to prevent weird overlaps which happen at the very same tick. Imagine boss saying textA due to event A, but due to while loop we also execute event B, which triggers textB, but spellcast from event B is prevented due to implicit casting check in DoCast. So we end up printing 2 texts but have only one cast.

Another way to prevent this would be using `if (... = events.ExecuteEvent())` instead of using a loop

Nyr97: So the issue is entirely made out of our choice to use a while to use ExecuteEvents, which also prevents having certain update ticks to ignore next call?

PR #29493, opened 2023-12-14
Scripts/Stormwind: Implement Quest "Battle for Azeroth: Tides of War"

Naddley:
**Changes proposed:**

-  Script Quest "Battle for Azeroth: Tides of War" ID: 46727

**Issues addressed:**

None

**Tests performed:**

build, tested ingame

- [ ] NPC: Vision of Sailor's Memory (139645) gets spawned two times instead of one
- [ ] Areatrigger misses restriction (AT should only trigger once but its not bound to Quest-Objective, serverside aura or Tracker Quest.) --> maybe a possible serverside (personal) AT which gets destroyed?


Nyr97:
I'd suggest new condition type for triggering, allowing OnEnter only if condition is met (and if it exists)

dio85:
AreaTriggerId 7045

PR #29509, opened 2023-12-21
Core/Maps: Update SpawnGroup Conditions when loading all cells

Golrag:
**Changes proposed:**

-  Update SpawnGroup Conditions when loading cells

**Issues addressed:**

Closes: none

Current issue: SpawnGroups that don't despawn on condition failure are being spawned even tho the condition to spawn has never been met.

**Tests performed:**

- Builds
- Tested in game while reworking alterac valley

- This might not be the proper fix

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -2487,6 +2487,22 @@ void Map::UpdateSpawnGroupConditions()
     }
 }
 
+void Map::InitSpawnGroupState()
+{
+    std::vector<uint32> const* spawnGroups = sObjectMgr->GetSpawnGroupsForMap(GetId());
+    if (!spawnGroups)
+        return;
+
+    for (uint32 spawnGroupId : *spawnGroups)
+    {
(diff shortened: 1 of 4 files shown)


Shauren (changes requested):
This one is rejected in current state, proper implementation would be to make a specialized version of `UpdateSpawnGroupConditions` that only marks groups as active/inactive (`SetSpawnGroupActive`, without touching WorldObjects) and call that on map construction, in MapManager::Create functions after setting bg/instancescript but before LoadAllCells

 - staying open as a reminder either until either of us do that

Golrag:
@Shauren, something like this? It seems to work on my alterac pr

Shauren:
I would make that not depend on load cells config and always call it

Golrag:
I still have it in the debug command, not sure if that was needed tho

Shauren (approved):

Shauren:
Not needed, as it was already guaranteed to have been called on map creation

PR #29516, opened 2023-12-22
Core/Misc: fixed nopch build for msvc compilers

Ovahlord:
yeah. it does just that


src/common/Asio/IpNetwork.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include "IpNetwork.h"
 #include "IpAddress.h"
+#include <algorithm>
```
Shauren, line 20: project includes
3rd party includes
std c++ includes
stc c includes

PR #29541, opened 2023-12-30
Core/Creatures: migrate creature_template_movement into static flags

Ovahlord:
With the discovery of static flags names we now know how and when to set certain creature movement states, which effectively renders all inhabittype fields in creature_template_movement obsolete. Which is why we are now going to replace them with the according static flags implementations

**Changes proposed:**

- creature_template_movement will be trimmed down to only determining movement generator behavior and initially setting hover and gravity states
-  implemented most movement related static flags
- deprecated CREATURE_FLAG_EXTRA_NO_MOVE_FLAGS_UPDATE  as dynamic switching between flight and hover states was never a thing on retail to begin with.

**Tests performed:**
- tested ingame

- [x] plain hovering is not determined by static flags. So we have to decide if we want to disable db side hover states or if we shall trust animTier values in template_addon
- [x] we have to decide if we want to apply the can_swim static flag to all creature by default and remove them based on migrated movement_template data
- [ ] CREATURE_STATIC_FLAG_3_CANNOT_PENETRATE_WATER needs support for path generation. For now it's only being used to determine if a creature is able to reach diving units
- [x] A migration sql query and sets the static flags based on existing data and then drop the columns in template_movement


Nyr97:
UNIT_FLAG_CAN_SWIM is present in most creature_template UnitFlags, as present as UNIT_FLAG2_REGENERATE_POWER for UnitFlags2. You'll need to remove every existing one from templates to avoid Worldserver throwing a whole lot of warnings 👀

Nyr97:
@mdX7 WPP should be updated to entirely ignore setting that flag on update or insert when the PR gets merged 👀

Shauren:
What is that GravityInitiallyDisabled weirdness? What about CREATURE_STATIC_FLAG_FLOATING?

Ovahlord:
> What is that GravityInitiallyDisabled weirdness? What about CREATURE_STATIC_FLAG_FLOATING?

CREATURE_STATIC_FLAG_FLOATING is permanently disabling gravity and does not determine the initial state. this flag is mostly used by trigger npcs

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12681,7 +12691,7 @@ bool Unit::SetDisableGravity(bool disable, bool updateAnimTier /*= true*/)
         SendMessageToSet(packet.Write(), true);
     }
 
-    if (IsCreature() && updateAnimTier && IsAlive() && !HasUnitState(UNIT_STATE_ROOT) && !ToCreature()->IsTemplateRooted())
+    if (IsCreature() && updateAnimTier && IsAlive() && !HasUnitState(UNIT_STATE_ROOT) && !ToCreature()->IsSessile())
```
Shauren, line 12694: Remove `IsSessile` check here, UNIT_STATE_ROOT already covers that (it is impossible to lose UNIT_STATE_ROOT when IsSessile is true)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12907,7 +12917,7 @@ bool Unit::SetHover(bool enable, bool updateAnimTier /*= true*/)
...
```
Shauren, line 12920: Remove `IsSessile` check here, UNIT_STATE_ROOT already covers that (it is impossible to lose UNIT_STATE_ROOT when IsSessile is true)

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2253,7 +2235,7 @@ void Creature::setDeathState(DeathState s)
         if (m_formation && m_formation->GetLeader() == this)
             m_formation->FormationReset(true);
 
-        bool needsFalling = (IsFlying() || IsHovering()) && !IsUnderWater();
+        bool needsFalling = (IsFlying() || IsHovering()) && !IsUnderWater() && !IsSessile();
```
Shauren, line 2238: I think this should check current root status (UNIT_STATE_ROOTED) instead of static flag (example: has root aura that is not removed on death)

Shauren (changes requested):

src/server/game/Entities/Creature/Creature.cpp
Ovahlord: agreed

PR #29544, opened 2023-12-31
Core/Maps: restored a hack in gridmap liquid status code

### LINKED ISSUE #22020: Liquid Status
**Description:**

Server is incorrectly determining the liquid status.

**Current behaviour:**

In some locations, server incorrectly determines the liquid status. Somewhere player can breathe underwater and somewhere player cannot breathe when he is not underwater.

Example of locations where player can breathe underwater:  
.go xyz -4634.93 3539.738 -114.65 0

Example of locations where player should be able to breathe:
.go xyz -6384 4078 -431 0     - Nespirah, here is also Fatigue and should not be
.go xyz -9760 -944 57 1

EDIT: with map version 1.7 some locations were fixed, so these locations were removed from example list

**Expected behaviour:**

Player should/should not be able to breathe it depends on if he is/is not underwater.

Just use some of commands above to teleport to the location.

**Branch(es):**

4.3.4

**TC rev. hash/commit:** 
ad3dff69c122

### LINKED ISSUE #29543: Inconsistent liquid status results on ocean floors
### Description

The liquid data is suffering from some precision issues as when moving closely to the ocean floor, there are spots where the character is falsely marked as 'not in water'. I guess these places are points where the player position slips below the bottom liquid geometry. So either we have to add a tiny bit more thickness to the edges of liquids or we have to take advantage of collision height values (starting at half height for example)

### Expected behaviour

Ocean floors should consistently label you as in or underwater.

1. Go to Vashjir's kelpthar forest (.tele kelpthar)
2. turn gm off to activate the breathing mirror timer
3. swim along the ocean floor and notice several spots where the breathing bar fills up again  .gps confirms the issue

### Branch

master

### TC rev. hash/commit

ee7b0e82fa076ad4338b1c362235cec6739f00e0

### Operating system

Windows 11

### Custom changes

None

Ovahlord:
**Changes proposed:**

-  restored a slightly less impactful hackfix for a precision issue in gridmap liquid code that causes players to be markes as not in water while being underwater.

**Issues addressed:**

Closes
closes #29543
closes #22020

**Tests performed:**
- tested ingame, results are promising

### DIFF
--- a/src/server/game/Maps/GridMap.cpp
+++ b/src/server/game/Maps/GridMap.cpp
@@ -593,6 +593,8 @@ float GridMap::getLiquidLevel(float x, float y) const
     return _liquidMap[cx_int*_liquidWidth + cy_int];
 }
 
+constexpr float GROUND_LEVEL_OFFSET_HACK = 0.02f; // due to floating point precision issues, we have to resort to a small hack to fix inconsistencies in liquids
+
 // Get water state on map
 ZLiquidStatus GridMap::GetLiquidStatus(float x, float y, float z, Optional<map_liquidHeaderTypeFlags> ReqLiquidType, LiquidData* data, float collisionHeight) const
 {
@@ -656,8 +658,8 @@ ZLiquidStatus GridMap::GetLiquidStatus(float x, float y, float z, Optional<map_l
 
     // Get water level
     float liquid_level = _liquidMap ? _liquidMap[lx_int*_liquidWidth + ly_int] : _liquidLevel;
-    // Get ground level (sub 0.2 for fix some errors)
-    float ground_level = getHeight(x, y);
+    // Get ground level (sub 0.02 for fix some errors)
+    float ground_level = getHeight(x, y) - GROUND_LEVEL_OFFSET_HACK;
 
     // Check water level and ground level
     if (liquid_level < ground_level || z < ground_level)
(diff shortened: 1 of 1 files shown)


src/server/game/Maps/GridMap.cpp
```diff
@@ -656,8 +658,8 @@ ZLiquidStatus GridMap::GetLiquidStatus(float x, float y, float z, Optional<map_l
 
     // Get water level
     float liquid_level = _liquidMap ? _liquidMap[lx_int*_liquidWidth + ly_int] : _liquidLevel;
-    // Get ground level (sub 0.2 for fix some errors)
-    float ground_level = getHeight(x, y);
+    // Get ground level (sub 0.02 for fix some errors)
+    float ground_level = getHeight(x, y) - GROUND_LEVEL_OFFSET_HACK;
```
Shauren, line 662: Are you sure that this is the way to go? `ground_level` variable is returned from this function in `LiquidData`

Ovahlord: You mean that this might cause issues when comparing that level against the wmo level?

Shauren: I mean that if every function in the callstack starts to "slightly adjust" inputs/outputs instead of only checks, you end up accumulating them for even greater inaccuracy

Ovahlord: fair enough. I'll relocate the check to the check below and test again

Ovahlord: tested, no regressions. change has been made

PR #29551, opened 2024-01-02
Core/Spells: Added option to override orientation for TARGET_DEST_DB

mdX7:
**Tests performed:**
tested ingame


src/server/game/Spells/SpellMgr.cpp
```diff
@@ -1193,11 +1195,14 @@ void SpellMgr::LoadSpellTargetPositions()
...
 
-        // target facing is in degrees for 6484 & 9268... (blizz sucks)
-        if (spellInfo->GetEffect(effIndex).PositionFacing > 2 * M_PI)
-            st.target_Orientation = spellInfo->GetEffect(effIndex).PositionFacing * float(M_PI) / 180;
-        else
-            st.target_Orientation = spellInfo->GetEffect(effIndex).PositionFacing;
+        if (!st.target_Orientation)
+        {
+            // target facing is in degrees for 6484 & 9268... (blizz sucks)
+            if (spellInfo->GetEffect(effIndex).PositionFacing > 2 * M_PI)
```
Ovahlord, line 1201: float cast for m_pi?

Nyr97:
So if we pass a WorldLocation as a destination containing orientation value it overrides whatever the actual facing float is?

mdX7:
> So if we pass a WorldLocation as a destination containing orientation value it overrides whatever the actual facing float is?

its for spells using `spell_target_position`, where we get the orientation from db2. With PR we can override it, if the value in db2 is incorrect (which sadly is the case sometimes)

Nyr97:
Oh, this is useful for Silverpine: Yorick's facing is entirely off because it was based on caster's orientation, but on retail she faces towards the caster when she spawns in.

PR #29554, opened 2024-01-03
Scripts/Bloodmyst isle: add missing Legoso Talk to $n pointer

ghost:
**Changes proposed:**

-  Legoso talk text now correctly references the player name, instead of showing "$n" in the text (4 talk lines updated).

**Issues addressed:**

Updates #21586

**Tests performed:**

- [x] Built multiple times without any observed issues.
- [x] Tested in-game, both in this form and non-simplified.

> ~~Not an issue or TODO as such, but I am uncertain whether this solution will be acceptable in its current form,~~
> ~~or if TC prefers to have one `if (Player* player = GetPlayerForEscort())` for each `Talk()`.~~
> ~~I don't know whether the added `if (!player) return;` creates new issues.~~

### DIFF
--- a/src/server/scripts/Kalimdor/zone_bloodmyst_isle.cpp
+++ b/src/server/scripts/Kalimdor/zone_bloodmyst_isle.cpp
@@ -357,10 +357,6 @@ public:
         {
             _events.Update(diff);
 
-            Player* player = GetPlayerForEscort();
-            if (!player)
-                return;
-
             if (UpdateVictim())
             {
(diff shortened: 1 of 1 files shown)


Shauren (changes requested):
This is a bad approach as it will cause the script to try to find a player on server every update tick, you should do it separately in each event case so that it runs only when neccessary

Legoso:
Very well. I sort of suspected something like that could be the case. I will modify it back to the form I used originally in my testing.

Shauren (approved):

PR #29562, opened 2024-01-06
Improvements to make it easier to developing/debugging battlegrounds.

r4d1sh:
**Changes proposed:**

-  Add RBAC permission RBAC_PERM_NO_BATTLEGROUND_DESERTER_DEBUFF. This solves the problem when you are constantly faced with the need to remove the Deserter while debugging/developing a battleground.
-  Do not apply Deserter debuff to players teleported from the battleground by gm command (.appear/.summon).
-  Add RBAC permission RBAC_PERM_CAN_AFK_ON_BATTLEGROUND. It allows you to be AFK on the battleground. This is also due to the convenience of working on battlegrounds.
- Implemented .bg start / .bg stop commands. Just use .bg start to skip battleground preparation time (2 mins).
- Unlink RBAC_PERM_IGNORE_IDLE_CONNECTION (7) from commands-role (196) and link it to sec-level-role (192). Just because it's right. I decided not to do a separate PR for such a small thing.

**Tests performed:**

Tested in-game


Shauren (approved):

Shauren:
Changed permission ids to avoid conflict with master branch (882 and 883 are used)

PR #29571, opened 2024-01-08
Update boss_cthun.cpp

MaroonBeard:
**Changes proposed:**
-  if/else changed to a ternary
-  add 'SetFacingTo()'

**Issues addressed:**
The Eye of C'thun and the spell Dark Glare were not visually moving, but the spell was still hitting/casting in the correct direction. This fix now reflects the correct orientation visually.

**Tests performed:**
Builds, and tested in-game. The Eye of C'thun now moves and deals damage in the correct location(s).

### DIFF
--- a/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
+++ b/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
@@ -342,7 +342,7 @@ public:
                         if (DarkGlareTickTimer <= diff)
                         {
                             //Set angle and cast
-                            float angle = (ClockWise) ? DarkGlareAngle + DarkGlareTick * float(M_PI) / 35 : DarkGlareAngle - DarkGlareTick * float(M_PI) / 35;
+                            float angle = ClockWise ? DarkGlareAngle + DarkGlareTick * float(M_PI) / 35 : DarkGlareAngle - DarkGlareTick * float(M_PI) / 35;
 
                             me->SetOrientation(angle);
                             me->SetFacingTo(angle);
--- a/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
+++ b/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
@@ -342,10 +342,10 @@ public:
                         if (DarkGlareTickTimer <= diff)
                         {
                             //Set angle and cast
-                            if (ClockWise)
-                                me->SetOrientation(DarkGlareAngle + DarkGlareTick * float(M_PI) / 35);
-                            else
-                                me->SetOrientation(DarkGlareAngle - DarkGlareTick * float(M_PI) / 35);
+                            float angle = (ClockWise) ? DarkGlareAngle + DarkGlareTick * float(M_PI) / 35 : DarkGlareAngle - DarkGlareTick * float(M_PI) / 35;
+
+                            me->SetOrientation(angle);
+                            me->SetFacingTo(angle);
 
                             me->StopMoving();
 


acidmanifesto (approved):
Lgtm

src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
```diff
@@ -342,10 +342,10 @@ class boss_eye_of_cthun : public CreatureScript
                         if (DarkGlareTickTimer <= diff)
                         {
                             //Set angle and cast
-                            if (ClockWise)
-                                me->SetOrientation(DarkGlareAngle + DarkGlareTick * float(M_PI) / 35);
-                            else
-                                me->SetOrientation(DarkGlareAngle - DarkGlareTick * float(M_PI) / 35);
+                            float angle = (ClockWise) ? DarkGlareAngle + DarkGlareTick * float(M_PI) / 35 : DarkGlareAngle - DarkGlareTick * float(M_PI) / 35;
```
jackpoz, line 345: Removing the parenthesis as they are not needed and we rarely have them in the rest of the code
```suggestion
                            float angle = ClockWise ? DarkGlareAngle + DarkGlareTick * float(M_PI) / 35 : DarkGlareAngle - DarkGlareTick * float(M_PI) / 35;
```

src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
```diff
@@ -342,10 +342,10 @@ class boss_eye_of_cthun : public CreatureScript
...
+
+                            me->SetOrientation(angle);
+                            me->SetFacingTo(angle);
```
jackpoz, line 348: I wonder if both are needed or if SetOrientation is not required when calling SetFacingTo. Could you check what happens when not calling SetOrientation ? I'll ask around too.

Fraxinus94: The direction of the spell cast isn't changed for some reason.

src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
Fraxinus94: Updated

src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
jackpoz: Well, I have no idea if all SetOrientation() calls need a SetFacingTo() and if all SetFacingTo() need a SetOrientation().

These functions are still affected by some bugs with workarounds added years ago and never properly fixed, like https://github.com/TrinityCore/TrinityCore/commit/76667e45a902f86bd341d40efa28037137e5e541 . Maybe we should call SetOrientation() in SetFacingTo() by default ? Or maybe we shouldn't ?

We leave that discussion for another day

jackpoz:
Thanks for the PR 🚀 😃

src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
Fraxinus94: Mostly another hackfix, but I think we should call SetFacingTo() in SetOrientation() here:
[Position.h](https://github.com/TrinityCore/TrinityCore/blob/e8849ca5a14719e7d395db7907bac26705f25dbf/src/server/game/Entities/Object/Position.h#L71)

PR #29582, opened 2024-01-16
Never save player's gold to db separately from their inventory.

r4d1sh:
Never save player's gold to db separately from their inventory.
Leave maintainer's contacs If additional explanation is needed.
High priority + need a quick cherry-pick to 3.3.5 branch

### DIFF
--- a/src/server/game/AuctionHouse/AuctionHouseMgr.cpp
+++ b/src/server/game/AuctionHouse/AuctionHouseMgr.cpp
@@ -1684,7 +1684,7 @@ bool AuctionHouseObject::BuyCommodity(CharacterDatabaseTransaction trans, Player
     }
 
     player->ModifyMoney(-int64(totalPrice));
-    player->SaveGoldToDB(trans);
+    player->SaveInventoryAndGoldToDB(trans);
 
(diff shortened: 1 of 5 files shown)


Shauren:
Is this one of those cases where two separate async transactions execute out of order?

r4d1sh:
> Is this one of those cases where two separate async transactions execute out of order?

No, that's a different issue. We shouldn't save gold separately from the inventory at all because essentially they are very interrelated things. If a player sells some item to a vendor, and then we save their gold without saving the inventory, what will happen if the server crashes? The sold item will return to player, but the saved gold won't disappear.

Shauren:
But that's not what I'm seeing in this PR, you are modifying operations that operate on gold only (guild bank, taking money from mail)

r4d1sh:
> But that's not what I'm seeing in this PR, you are modifying operations that operate on gold only (guild bank, taking money from mail)

Yes, because these are the places where gold saved to db separately from inventory. Knowing this, player can easily save his gold to db whenever he wants, separately from his inventory. And just wait until the server crashes and he gets his profit

CraftedRO:
> ++ need a quick cherry-pick to 3.3.5 branch

Issue #26738, opened 2021-07-23
Stratholme: Portcullis Trap


TQPS:
### Description

As you pass under the first portcullis in Stratholme on the way to live, it does not close, and the mobs do not spawn.

### Expected behaviour

Each time you pass through here during a run or re-entry to the instance, both portcullis gates close shut, trapping those inside and a bunch of little maggots spawn. Once the mobs have been disposed of the portcullis opens up again.

If you run back via this way, it does not re-close.

If you exit and re-enter the instance (example dying), it will re-close and spawns re-occur.

### Steps to reproduce the problem

1. Enter Stratholme
2. Run to live 
3. Pass under first portcullis on the way
4. Does not close, remains open and nothing occurs

<details>

Image 

![image](https://user-images.githubusercontent.com/66294837/126759073-45d0100b-122a-4e90-8ef1-3cd53680b1c7.png)
</details>

### Branch

3.3.5

### TC rev. hash/commit

TrinityCore rev. 87a4bc3 2021-07-23 00:12:58 +0300 (3.3.5 branch) (Win64, RelWithDebInfo, Static)

### Operating system

Windows 7 x64

### Custom changes

Trickerer's Bots

offl:
To make it more clear, it's about https://github.com/classicdb/database/issues/631
> The trap gates do not trigger. Video of current core: https://youtu.be/4u3Hpdq44Pc Video of working trap gates: https://youtu.be/--IiC_Wuxfg?t=411

14shagov:
I apologize for cluttering up the place with pull requests. I didn't immediately understand how it worked.

Based on cmangos/mangos-classic@eb8bf25.
I made 14shagov/TrinityCore@3aa6191.

The mechanics are almost the same as you described, but there is a difference - if one and only one of the traps is triggered, both traps are deactivated after opening the doors. I don't know how critical this is.
Also, if all enemies are not killed within two minutes, the doors will open on their own.
If there are any comments, I'd be glad to hear them.

One more thing, as described in classicdb/database#631, there is no area_trigger set for these traps on map 329, so the trap will check every 30 seconds if someone is around. The rest of the time it will be deactivated.
I don't know if there is any way to fix this other than increasing the frequency of checking.

Aokromes:
75e38a5d8f739358421ca5eece7e3701f6b8d190

PR #29596, opened 2024-01-19
Scripts/Stratholme: Portcullis Trap

14shagov:
**Changes proposed:**

- a working Portcullis Trap in Stratholme

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/26738

**Tests performed:**

Builded and tested in game

**Known issues:** 

I wrote this in the issue, but I'll repeat it.
The mechanics are almost the same, it was described, but there is a difference - if one and only one of the traps is triggered, both traps are deactivated after opening the doors. I don't know how critical that is.
Also, if all enemies are not killed within two minutes, the doors will open on their own.

One more thing, as described in classicdb/database#631, there is no area_trigger set for these traps on map 329, so the trap will check every 30 seconds if someone is around. The rest of the time it will be deactivated.
I don't know if there is any way to fix this other than increasing the frequency of checking.


src/server/scripts/EasternKingdoms/Stratholme/instance_stratholme.cpp
```diff
@@ -95,6 +110,37 @@ class instance_stratholme : public InstanceMapScript
             GuidSet abomnationGUID;
             EventMap events;
 
+            GuidVector TrapGatesGUIDs[2]; //two gate areas
+            std::map<TempSummon*, uint8> PlaguedGatesGUIDs;
```
Shauren, line 114: Do not ever store pointers to world objects for longer than "current update tick", if anything causes them to despawn, this will crash

Use ObjectGuid instead

Shauren (changes requested):
At 30 second update frequency, this is useless as a trap...

src/server/scripts/EasternKingdoms/Stratholme/instance_stratholme.cpp
```diff
@@ -119,6 +166,21 @@ class instance_stratholme : public InstanceMapScript
                         }
                         break;
                     }
+                    case NPC_PLAGUED_RAT:
+                    case NPC_PLAGUED_INSECT:
+                    case NPC_PLAGUED_MAGGOT:
+                    {
+                        auto el = PlaguedGatesGUIDs.find(who->GetGUID());
+                        uint8 uiGate = el->second;
```
Shauren, line 174: You are possibly dereferencing invalid (.end()) iterator here, all of the code here should be under `if (el != PlaguedGatesGUIDs.end())`, not just erase

src/server/scripts/EasternKingdoms/Stratholme/instance_stratholme.cpp
```diff
@@ -119,6 +166,24 @@ class instance_stratholme : public InstanceMapScript
...
+                    case NPC_PLAGUED_RAT:
+                    case NPC_PLAGUED_INSECT:
+                    case NPC_PLAGUED_MAGGOT:
+                    {
+                        auto el = PlaguedGatesGUIDs.find(who->GetGUID());
+                        if (el != PlaguedGatesGUIDs.end())
+                        {
+                            uint8 uiGate = el->second;
+                            PlaguedGatesGUIDs.erase(el);
+
```
Aokromes, line 178: src/server/scripts/EasternKingdoms/Stratholme/instance_stratholme.cpp:178:                        

Remove whitespace at the end of the lines above

Exited with code exit status 1

PR #29600, opened 2024-01-21
.account set seclevel command will update security level for online sessions

r4d1sh:
**Changes proposed:**

-  You will not need to re-login to your account after you have changed your security level using the console command.

**Tests performed:**

Builded and tested in-game.


CraftedRO:
can't be the other command we already have be updated to do this too, instead creating a new one ?

r4d1sh:
> can't be the other command we already have be updated to do this too, instead creating a new one ?

What you mean? This PR does not create any new commands.

CraftedRO:
ah ok, I guess I missread it, nice PR

Shauren (approved):

PR #29613, opened 2024-01-23
Spells/Priest: Body and Soul Proc

cristianvnt:
**Issues addressed:**

[Body and Soul](https://www.wowhead.com/spell=64129/body-and-soul) procs from almost any ability used. 
It should proc ONLY from **PW:Shield** and **Leap of Faith**.

**Tests performed:**

1. Log on a priest.
2. Have selected in the talent tree **Body and Soul**.
3. Use any ability.


Nyr97:
This is not the correct way of implementing the spell, you should use the spell_proc table, AuraScripts for such are a no-no.

Mematoru23:
Aight, didn't know that. AuraScripts should be used then for auras that dont work at all? or it depends

> This is not the correct way of implementing the spell, you should use the spell_proc table, AuraScripts for such are a no-no when it can be handled entirely on the DB.

jackpoz:
You might want to create a branch before opening a PR, so you can open multiple PRs with only the changes related to them

PR #29624, opened 2024-01-27
Spells/Priest: Power Word: Barrier

cristianvnt:
**Issues addressed:**

[Power Word: Barrier](https://www.wowhead.com/spell=62618/power-word-barrier)'s buff not aplied to players.

**Tests performed:**

1. Make a Disc Priest
2. Use Power Word: Barrier
3. Go inside the Barrier


mdX7:
Thankfully we have the table `areatrigger_template_actions` for this stuff, that way we dont have to write that code for all areatriggers as in: on enter apply buff, on leave remove buff. We have the documentation about it here:
https://trinitycore.info/database/master/world/areatrigger_template_actions

What I've noticed is that you put an INSERT statement to assign ScriptName for a row we already have in db, its prefered that you use an `UPDATE` statement for existing rows - it would look like this: 
```sql
UPDATE `areatrigger_create_properties` SET `ScriptName`='at_blabla' WHERE `Id`=1489 AND `IsCustom`=0;
```

PR #29626, opened 2024-01-28
Core/PacketIO: Implemented CMSG_CHECK_IS_ADVENTURE_MAP_POI_VALID and ...

mdX7:
…SMSG_PLAYER_IS_ADVENTURE_MAP_POI_VALID

**Tests performed:**
tested ingame as alliance char in Boralus with
```sql
UPDATE `gossip_menu_option` SET `GossipOptionID`=48786, `OptionBroadcastTextID`=95541, `GossipNpcOptionID`=30917, `VerifiedBuild`=53040 WHERE (`MenuID`=22725 AND `OptionID`=0); -- OptionBroadcastTextID: 95541 - 96633 - 98823

DELETE FROM `creature_template_gossip` WHERE (`CreatureID`=139522 AND `MenuID`=22725);
INSERT INTO `creature_template_gossip` (`CreatureID`, `MenuID`, `VerifiedBuild`) VALUES
(139522, 22725, 53040); -- Scouting Map

DELETE FROM `creature_template_addon` WHERE `entry`=139522;
INSERT INTO `creature_template_addon` (`entry`, `auras`) VALUES
(139522, '274053');

-- for testing purposes, should be conditional, based on if required quest status is != NONE, old value = 0
UPDATE `creature_template` SET `npcflag`=1 WHERE `entry`=139522;
```
with rewarded quests:
```
47099
46729
47186
```
and active quest `47189`


Shauren:
Should this also check CanTakeQuest?

mdX7:
> Should this also check CanTakeQuest?

when looking at quest status only it was always returning true when conditions were met, unrelated to quest status (none/taken/rewarded) - the client is filtering whether to show the related quest on the map or not. The other checks of CanTakeQuest are not really possible to test

PR #29629, opened 2024-01-28
Core/Quest: Implemented AcceptConversationID and RewardConversationID

mdX7:
**Tests performed:**
tested ingame

**Note:**
Maybe used for worldquests at some point or quests like 50476


mdX7:
flagging as draft until we know how we may implement wqs

PR #29649, opened 2024-02-01
Scripts/Spells: Implemented Priest talent Blaze of Light

cristianvnt:
**Issues addressed:**

[Blaze of Light](https://www.wowhead.com/spell=215768/blaze-of-light) doesn't proc at all, and it should from Penance.

**Tests performed:**

tested ingame

### DIFF
--- a/src/server/scripts/Spells/spell_priest.cpp
+++ b/src/server/scripts/Spells/spell_priest.cpp
@@ -625,17 +625,25 @@ class spell_pri_benediction : public SpellScript
 // 215768 - Blaze of Light
 class spell_pri_blaze_of_light : public AuraScript
 {
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo(
+        {
+            SPELL_PRIEST_BLAZE_OF_LIGHT_DECREASE,
+            SPELL_PRIEST_BLAZE_OF_LIGHT_INCREASE
+        });
+    }
+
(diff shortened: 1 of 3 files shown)


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -619,6 +622,30 @@ class spell_pri_benediction : public SpellScript
...
+    {
+        Unit* caster = eventInfo.GetActor();
+        Unit* target = eventInfo.GetProcTarget();
+        if (!caster || !target)
+            return;
+
+        if (caster->IsValidAttackTarget(target))
+        {
+            caster->CastSpell(target, SPELL_PRIEST_BLAZE_OF_LIGHT_DECREASE, aurEff);
+            return;
```
Nyr97, line 638: is there any reason you're returning here?

Mematoru23: theres no point since i check the target, right?

mdX7:
```c
if (caster->IsValidAttackTarget(target))
    caster->CastSpell(target, SPELL_PRIEST_BLAZE_OF_LIGHT_DECREASE, aurEff);
else
    caster->CastSpell(target, SPELL_PRIEST_BLAZE_OF_LIGHT_INCREASE, aurEff);
```

Mematoru23: i see i see

PR #29669, opened 2024-02-05
Scripts/Priest: Implemented Priest talent Indemnity

cristianvnt:
**Issues addressed:**

[Indemnity](https://www.wowhead.com/spell=373049/indemnity) doesn't increase Atonement duration by 2 when applied by Power Word: Shield.

**Tests performed:**

tested ingame


Shauren:
This had to be rolled into existing `spell_pri_atonement_effect` script because order of script execution is not determined

PR #29671, opened 2024-02-05
Scripts/Spells: Implemented Priest talent Heaven's Wrath

cristianvnt:
Discipline Priest talent

**Issues addressed:**
[Heaven's Wrath](https://www.wowhead.com/spell=421558/heavens-wrath) doesn't reduce Ultimate Penitence cooldown.

**Tests performed:**
tested ingame
- rank 1 and 4 Penance bolts => 1*4 = 4 sec cdr(from Penanace) + 2 sec cdr(from spell's cd) = 6 sec cdr(e.g.1:50->1:44)
- rank 2 and 4 Penance bolts => 2*4 = 8 sec cdr(from Penance) + 2 sec cdr(from spell's cd) = 10 sec cdr(e.g.1:50->1:40)

build used(for 2nd test): BAQA0Hr2WRGgVq7/s2iQ2HhjlABoFEk0SSIIlkSSikEAAAAAAAAAAAAS7AikEkCRSiIQ5AEJRUIA

### DIFF
--- a/sql/updates/world/master/9999_99_99_99_world_heavens_wrath.sql
+++ b/sql/updates/world/master/9999_99_99_99_world_heavens_wrath.sql
@@ -1,0 +1,7 @@
+DELETE FROM `spell_script_names` WHERE `spell_id` = 421558;
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
+(421558, 'spell_pri_heavens_wrath');
+
+DELETE FROM `spell_proc` WHERE `SpellId` IN (421558);
+INSERT INTO `spell_proc` (`SpellId`,`SchoolMask`,`SpellFamilyName`,`SpellFamilyMask0`,`SpellFamilyMask1`,`SpellFamilyMask2`,`SpellFamilyMask3`,`ProcFlags`,`ProcFlags2`,`SpellTypeMask`,`SpellPhaseMask`,`HitMask`,`AttributesMask`,`DisableEffectsMask`,`ProcsPerMinute`,`Chance`,`Cooldown`,`Charges`) VALUES
+(421558,0x02,6,0x00000000,0x00818000,0x00000080,0x00000000,0x0,0x0,0x3,0x2,0x403,0x0,0x0,0,0,0,0); -- Heaven's Wrath
--- a/sql/updates/world/master/9999_99_99_99_world_heavens_wrath.sql
+++ b/sql/updates/world/master/9999_99_99_99_world_heavens_wrath.sql
@@ -4,4 +4,4 @@ INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
 
 DELETE FROM `spell_proc` WHERE `SpellId` IN (421558);
 INSERT INTO `spell_proc` (`SpellId`,`SchoolMask`,`SpellFamilyName`,`SpellFamilyMask0`,`SpellFamilyMask1`,`SpellFamilyMask2`,`SpellFamilyMask3`,`ProcFlags`,`ProcFlags2`,`SpellTypeMask`,`SpellPhaseMask`,`HitMask`,`AttributesMask`,`DisableEffectsMask`,`ProcsPerMinute`,`Chance`,`Cooldown`,`Charges`) VALUES
-(421558,0x02,6,0x00000000,0x00818000,0x00000080,0x00000000,0x0,0x0,0x3,0x2,0x403,0x0,0x0,0,0,0,0); -- Heaven's Wrath
+(421558,0x00,6,0x00000000,0x00018000,0x00000000,0x00000000,0x0,0x0,0x3,0x2,0x403,0x0,0x0,0,0,0,0); -- Heaven's Wrath
--- a/src/server/scripts/Spells/spell_priest.cpp
+++ b/src/server/scripts/Spells/spell_priest.cpp
@@ -1264,9 +1264,16 @@ class spell_pri_heavens_wrath : public AuraScript
     {
         return ValidateSpellInfo
         ({
-            SPELL_PRIEST_HEAVENS_WRATH,
             SPELL_PRIEST_ULTIMATE_PENITENCE
         });
+    }
+
+    bool CheckProc(ProcEventInfo& eventInfo)
+    {
+        if (eventInfo.GetSpellInfo()->Id == SPELL_PRIEST_ULTIMATE_PENITENCE)
+            return false;
+
+        return true;
     }
 
     void HandleEffectProc(AuraEffect const* aurEff, ProcEventInfo& eventInfo)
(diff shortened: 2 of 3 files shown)


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -1255,6 +1257,36 @@ class spell_pri_guardian_spirit : public AuraScript
...
 };
 
+// 421558 - Heaven's Wrath
+class spell_pri_heavens_wrath : public AuraScript
+{
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo
+        ({
+            SPELL_PRIEST_HEAVENS_WRATH,
```
Shauren, line 1267: Don't put the spell you are attaching script to inside Validate (it's always validated internally before even reaching this function)

sql/updates/world/master/9999_99_99_99_world_heavens_wrath.sql
```diff
@@ -0,0 +1,7 @@
+DELETE FROM `spell_script_names` WHERE `spell_id` = 421558;
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
+(421558, 'spell_pri_heavens_wrath');
+
+DELETE FROM `spell_proc` WHERE `SpellId` IN (421558);
+INSERT INTO `spell_proc` (`SpellId`,`SchoolMask`,`SpellFamilyName`,`SpellFamilyMask0`,`SpellFamilyMask1`,`SpellFamilyMask2`,`SpellFamilyMask3`,`ProcFlags`,`ProcFlags2`,`SpellTypeMask`,`SpellPhaseMask`,`HitMask`,`AttributesMask`,`DisableEffectsMask`,`ProcsPerMinute`,`Chance`,`Cooldown`,`Charges`) VALUES
+(421558,0x02,6,0x00000000,0x00818000,0x00000080,0x00000000,0x0,0x0,0x3,0x2,0x403,0x0,0x0,0,0,0,0); -- Heaven's Wrath
```
Shauren, line 7: SchoolMask is optional (setting it to holy is also wrong because that causes this talent to not work with Power of the Dark Side - it does on retail)

SpellFamilyMask is wrong, you made it too broad
should be only 0x00018000 in `SpellFamilyMask1`, 0 in other fields

It also shouldn't proc from Ultimate Penitence bolts (they have the same SpellFamilyMask as penance bolts) - you will need to exclude that in script

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -1255,6 +1257,36 @@ class spell_pri_guardian_spirit : public AuraScript
...
+        });
+    }
+
+    void HandleEffectProc(AuraEffect const* aurEff, ProcEventInfo& eventInfo)
+    {
+        Unit* caster = eventInfo.GetActor();
+        if (!caster)
+            return;
+
+        SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(SPELL_PRIEST_ULTIMATE_PENITENCE, GetCastDifficulty());
```
Nyr97, line 1278: missing return here

Shauren (changes requested):

src/server/scripts/Spells/spell_priest.cpp
Mematoru23: so i should put eveything else i use besides the spell im attaching to

src/server/scripts/Spells/spell_priest.cpp
Shauren: wrong advice, ignore

@Nyr97 - its validated

sql/updates/world/master/9999_99_99_99_world_heavens_wrath.sql
Mematoru23: any tips on how to exclude the bolts of that spell?

Shauren: No other option other than hardcode spell ids

Mematoru23: im not sure what you mean by this

Mematoru23: > SpellFamilyMask is wrong, you made it too broad should be only 0x00018000 in `SpellFamilyMask1`, 0 in other fields

but i found Penance in the other masks too, so why wouldn't it be fine?

mdX7: check spellid in CheckProc

Shauren: > > SpellFamilyMask is wrong, you made it too broad should be only 0x00018000 in `SpellFamilyMask1`, 0 in other fields
> 
> but i found Penance in the other masks too, so why wouldn't it be fine?

You want to find a minimal mask that covers all the spells you want (and nothing else)
And that is these four: 373130, 400187, 47666 and 47750

Mematoru23: > > > SpellFamilyMask is wrong, you made it too broad should be only 0x00018000 in `SpellFamilyMask1`, 0 in other fields
> > 
> > 
> > but i found Penance in the other masks too, so why wouldn't it be fine?
> 
> You want to find a minimal mask that covers all the spells you want (and nothing else) And that is these four: 373130, 400187, 47666 and 47750

aight
so as i thought

Mematoru23:
made changes, i'd like another review

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -1255,6 +1257,43 @@ class spell_pri_guardian_spirit : public AuraScript
...
+    {
+        return ValidateSpellInfo
+        ({
+            SPELL_PRIEST_ULTIMATE_PENITENCE
+        });
+    }
+
+    bool CheckProc(ProcEventInfo& eventInfo)
+    {
+        if (eventInfo.GetSpellInfo()->Id == SPELL_PRIEST_ULTIMATE_PENITENCE)
```
Shauren, line 1273: This needs to exclude spells 421543 (damage bolt) and 421544 (healing bolt), not 421453 (the aura with cooldown)

Shauren (changes requested):

src/server/scripts/Spells/spell_priest.cpp
Mematoru23: oh, i see, makes sense

src/server/scripts/Spells/spell_priest.cpp
Nyr97: dont we need ti assert in that case?

mdX7: If the validation fails script is not loaded

Nyr97: oh I understand

PR #29676, opened 2024-02-06
Scripts/Spells: Implemented Priest talent Train of Thought

cristianvnt:
**Issues addressed:**

[Train of Thought](https://www.wowhead.com/spell=390693/train-of-thought) doesn't work.
Power Word: Shield's cd should be reduced by 1sec by Flash Heal or Renew.
Penance's cd should be reduced by 0.5sec by Smite.

**Tests performed:**

tested ingame
disc priest
build: BAQA0Hr2WRGgVq7/s2iQ2HhjlABoFEcg0SSIIlkSSikkAAAAAAAAAAAAi0ISSQKEJJiAlDQkERhA

### DIFF
--- a/src/server/scripts/Spells/spell_priest.cpp
+++ b/src/server/scripts/Spells/spell_priest.cpp
@@ -175,7 +175,6 @@ enum PriestSpells
     SPELL_PRIEST_STRENGTH_OF_SOUL_EFFECT            = 197548,
     SPELL_PRIEST_SURGE_OF_LIGHT                     = 109186,
     SPELL_PRIEST_SURGE_OF_LIGHT_EFFECT              = 114255,
-    SPELL_PRIEST_TRAIN_OF_THOUGHT                   = 390693,
     SPELL_PRIEST_TRANQUIL_LIGHT                     = 196816,
     SPELL_PRIEST_THE_PENITENT_AURA                  = 200347,
     SPELL_PRIEST_TRAIL_OF_LIGHT_HEAL                = 234946,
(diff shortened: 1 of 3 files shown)


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -2846,6 +2847,52 @@ class spell_pri_trail_of_light : public AuraScript
...
+    void HandleProc(ProcEventInfo& eventInfo)
+    {
+        Unit* caster = GetTarget();
+        if (!caster)
+            return;
+
+        if (eventInfo.GetSpellInfo()->Id == SPELL_PRIEST_SMITE)
+        {
+            SpellInfo const* penance = sSpellMgr->GetSpellInfo(SPELL_PRIEST_PENANCE, GetCastDifficulty());
+            AuraEffect const* aurEff = caster->GetAuraEffect(SPELL_PRIEST_TRAIN_OF_THOUGHT, EFFECT_1);
```
Nyr97, line 2875: you need to validate this effect, OnProc hook does not implicitly validate effects regardless of not being scripting a specific effect

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -2846,6 +2847,52 @@ class spell_pri_trail_of_light : public AuraScript
...
+            int32 cdr = aurEff->GetAmount();
+
+            caster->GetSpellHistory()->ModifyCooldown(penance, Milliseconds(cdr));
+        }
+        else
+        {
+            SpellInfo const* pws = sSpellMgr->GetSpellInfo(SPELL_PRIEST_POWER_WORD_SHIELD, GetCastDifficulty());
+            AuraEffect const* aurEff = caster->GetAuraEffect(SPELL_PRIEST_TRAIN_OF_THOUGHT, EFFECT_0);
```
Nyr97, line 2883: same thing here

Nyr97 (changes requested):

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -2846,6 +2847,52 @@ class spell_pri_trail_of_light : public AuraScript
...
+            SPELL_PRIEST_RENEW,
+            SPELL_PRIEST_SMITE,
+            SPELL_PRIEST_POWER_WORD_SHIELD,
+            SPELL_PRIEST_PENANCE
+        });
+    }
+
+    void HandleProc(ProcEventInfo& eventInfo)
+    {
+        Unit* caster = GetTarget();
```
Nyr97, line 2868: it is misleading to call our pointer caster when we're getting target (even though we know the target of the effect is the caster in this case)

Nyr97 (changes requested):

PR #29696, opened 2024-02-14
Core/Player: Save druid/warlock customisations at barbershop

### LINKED ISSUE #29429: Moonkin/Druid Customizations Don't Save
### Description

Moonkin/Druid Barbershop Customization do not save when hitting accept, and instead default back to base moonkin form.

### Expected behaviour

Moonkin/Druid form should be customizable and save after hitting accept.

1. Create Druid Character
2. Spec Balance(Or don't)
3. Use barbershop chair to customize Moonkin/Druid Forms"
4. Hit accept.
5. Shift Form or Reuse Barbershop chair

### Branch

master

### TC rev. hash/commit

97f9e18cbdb520c242f46827d2d171de37f55afa

### Operating system

Windows 10

### Custom changes

None.

Razinao:
**Changes proposed:**

-  Barbershop alter appearance packets come in multiple times, one for each character model (base character model, shapeshift forms, warlock pet forms, etc). At the moment, after the first packet comes in, the checks on being in a barbershop chair pass and then the character is forced to stand. Subsequent packets fail, so they don't get to perform SetCustomizations and save the changes.
-  Additionally, SetCustomizations currently performs a wipe of all customisations. We change this to only wipe the options relevant to the chrModel that is being modified. When the chrModel is 0, we fetch the customisations from the race + gender -> chrModel customisations. When this is non-zero, we fetch the chrModel customisations. In both cases we just delete any existing entries from the database/the Customisations UF.
-  ValidateAppearance was handling the customisations for a non playable race model as if it was the playable race model, so it was validating if option 911 (cat form) was available to character model Highmountain Tauren Female. This is not the case: we should be validating on if character model 190 has option 911. This most likely came from a DB2 change between 9.2.5 and now which has gone overlooked.
- When using the barbershop or character customise, changing gender requires us to remove all of the customisations associated with the old model. This is to prevent build up of useless datapoints and keep the ValidateAppearance customisations set clean.
- The same applies for when changing races/factions. The old race/gender combination needs to be removed entirely.
- Additionally, when using the change race/faction for a class like druid, there are forms associated with a race. We want to remove the bear/cat forms associated with the old race.
- Between 9.2.5 and now, a change was made so that all Bear Form options have been merged into one, and it seems like the best way to get which race was associated with the choices now is to use the RaceMask in ChrCustomizationReq. By checking through all of the old races choices and seeing if the new race also supports it, we can keep forms like the artifact / achievement forms from being reset. If the new race doesn't support it, we delete the old customisation option->choice
- When logging in, we need to force the druid customisations to something valid as these are not pre-determined at character creation or when a race change has been completed. So, at log in, we find the first viable candidate. This will typically be a customisation that has no unlock requirements.

**Issues addressed:**

Closes #29429 

**Tests performed:**

- Verified build.

- Verified that when using the barbershop to set druid form/warlock pet customisations, that these customisations are correctly saved.

- Verified that when using character customisation screen to set base model customisations, that the druid form/warlock pet customisations do not get wiped.

- Verified that when using the character changerace/changefaction, that commonly accessible customisations (i.e. warlock pets, artifact shapeshift models or other unlockables) do not get reset.

- Verified that when using the character changerace/changefaction, that race-only accessible customisations (i.e. base bear/cat/travel forms) are forced to a valid one for the new race.

**Known issues list:**

- [ ] When interactring with the barbershop in druid form or while a warlock pet is out, the customisation does not immediately take effect.


Shauren:
I do not agree with your proposed solution that introduces a dependency on customization packet order

Instead of having the main model request wipe all customizations, have each packet replace only customizations related to model being customized in that packet (this would also take care of doing it on character selection screen)

KamiliaBlow:
It looks like your changes are working a little incorrectly. You can change the appearance of your warlock pets or druid shapes. But at the same time, all other classes cannot change the appearance of their characters with the help of a hairdresser (the changes are not applied to the database). Characters of the warlock and druid class can change the appearance of a character and save it to the database.

Aokromes:
Conflicting files
src/server/game/Handlers/CharacterHandler.cpp

Razinao:
> It looks like your changes are working a little incorrectly. You can change the appearance of your warlock pets or druid shapes. But at the same time, all other classes cannot change the appearance of their characters with the help of a hairdresser (the changes are not applied to the database). Characters of the warlock and druid class can change the appearance of a character and save it to the database.

Thanks for catching this, this should be fixed with the latest commit.

thorton1786:
Hello sorry to be a pain I’m greatful for your work and it seems to be working great. The only issue I seem to be having still is the barbershop changes resetting for worgen druids. I haven’t had the issue with any other races as of yet.

Razinao:
> Hello sorry to be a pain I’m greatful for your work and it seems to be working great. The only issue I seem to be having still is the barbershop changes resetting for worgen druids. I haven’t had the issue with any other races as of yet.

This was being caused because the human customizations were not being removed, and so the transaction to commit new customizations was inserting partially duplicated results.

I'm not too keen on the approach that seems to be required to fix this. Skin colour option 14 is tied to ChrModel 2 (human female) but this ChrModel just so happens to be used by Worgen as well.

Either I start associating those non-worgen options to worgen, or I make a specific handler to also clear out the human model customisation options for worgen.

Nonetheless, this should be fixed in the latest commit.

KamiliaBlow:
It looks like the barber shop for druids and mages is not working again. When I get to the barbershop interface, they won't let me choose a druid or demon look.

trackmastersteve:
I have tried all of the Horde druids. And they all seem to be stuck with a white (uncolored) Haranir model. And I can't seem to change it in the barbershop. The Haranir versions seem to look right, accept the Moonkin. And again, I can't change it in the barbershop. But as least they have color! Thanks again for all of the hard work! This project is amazing!

PR #29727, opened 2024-02-20
Scripts/Spells: Implemented Priest talent Sanctuary

cristianvnt:
**Issues addressed:**

[Sanctuary](https://www.wowhead.com/spell=231682/sanctuary) doesn't work. Smite should apply on target an aura that prevents x damage dealt by enemy.

200d8ac9d7d0aa2ed34280004bc481cba179df6e

**Tests performed:**

tested ingame
discipline talent
build: BAQA0Hr2WRGgVq7/s2iQ2HhjlABopgg0SCiIlokkEJJBAAAAAAAAAAAItgkEkCRSiIQJEJRUIA


Shauren (changes requested):
This doesn't work at all, nothing about these spells do any sort of damage prevention

Mematoru23:
so just by applying the aura on target doesn't mean is gonna work right?

Nyr97:
I mean the spell passes the info correctly to the aura, but the aura itself is SPELL_AURA_DUMMY, it won't do much on itself besides showing the correct numbers on the client.

Nyr97:
Also, the spell that does the absorb is Mind Blast - 208771. Ignore the name, the didn't care about renaming it.

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -2670,58 +2671,125 @@ class spell_pri_schism : public SpellScript
...
+    void HandleEffectHit(SpellEffIndex /*effIndex*/) const
     {
-        if (GetCaster()->HasAura(SPELL_PRIEST_SANCTUARY))
-            GetCaster()->CastSpell(GetHitUnit(), SPELL_PRIEST_SANCTUARY_AURA, true);
+        Unit* caster = GetCaster();
+
+        if (caster->HasAura(SPELL_PRIEST_SANCTUARY))
+        {
+            caster->CastSpell(GetHitUnit(), SPELL_PRIEST_SANCTUARY_AURA, true);
+            caster->CastSpell(caster, SPELL_PRIEST_SANCTUARY_ABSORB, true);
```
Nyr97, line 2694: Why are you choosing caster as target? The one holding the aura should be the target of smite.

Mematoru23: from sniffs it looks like both target and caster get an aura, 208771 on caster(like a hidden one, from which the absorb happens) and 208772 on target

PR #29730, opened 2024-02-20
Core/Spells: Fix SPELLVALUE_CRIT_CHANCE

r4d1sh:
**Changes proposed:**

Why do we need this division here?

image

roll_chance_f takes a pure percentage value.


Shauren (approved):

Shauren:
I guess this is because someone wanted to support passing 10.5 as forced crit chance (shoving 1050 into spellvalue)

PR #29745, opened 2024-02-24
Core/Spells: Fixed work of sobering spells and other improvements for drunk system.

r4d1sh:
**Changes proposed:**

The first commit related to `SPELL_EFFECT_INEBRIATE` (100) effect. The problem is that there are not only inebriation but also sobering spells (with basepoint < 0). The current implementation uses an unsigned type, ignoring sobering spells. This leads to a uint8 underflow, resulting in the spell's behavior being completely unexpected: it can both inebriate and sober you up.

Example:

1. `.gobject add temp 185914`
2. Use it!
3. If you were completely sober before, you will now get maximum inebriation. Although coffee was supposed to sober you up by 20%. This happened because of uint8 underflow.

There is another problem: there are spells with an inebriate effect with basepoint values so large or small that it inevitably leads to uint8 overflow/underflow. For example, `29690` or `46874`. As intended, `29690` should make you extremely drunk, but in reality, the behavior can be unpredictable each time. If someone keeps casting this spell on you again and again, you will likely notice that sometimes you get drunk and sometimes you sober up.

The first commit completely fixes this issues.

The second commit is related to the aura effect `SPELL_AURA_MOD_FAKE_INEBRIATE` (304). The main problem is that the handler of this effect ignores the value of player's actual inebriation. In any case, there was a lot of redundant (duplicate) code here. Now the logic for updating invisibility detect has been moved to a separate method, and all known issues have been fixed.

**Tests performed:**

Builded and tested in-game.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -959,7 +959,7 @@ void Player::HandleSobering()
     SetDrunkValue(drunk);
 }
 
-DrunkenState Player::GetDrunkenstateByValue(uint8 value)
+/*static*/ DrunkenState Player::GetDrunkenstateByValue(uint8 value)
 {
     if (value >= 90)
         return DRUNKEN_SMASHED;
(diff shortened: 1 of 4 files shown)


Ovahlord:
the sobering part has been fixed in master branch by me already. so cherrypick that commit.

r4d1sh:
> the sobering part has been fixed in master branch by me already. so cherrypick that commit. [45333fc](https://github.com/TrinityCore/TrinityCore/commit/45333fcc8fd8fbd58b6168098a8445b36b9c5248)

Oh, I don't think the overflow/underflow issue is completely resolved here. Unless we hope that no one will ever create a spell with basepoint = (2147483547; 2147483647]

Ovahlord:
this has not been a problem throughout all expansions so far so I doubt that this will ever happen.

Ovahlord:
just change the int32 to int64 and your problems are gone.

r4d1sh:
> just change the int32 to int64 and your problems are gone.

Yes, but why should I do this if ready-made code has already been written that completely solves the problem. Not mine btw xd

CraftedRO:
Just tested this pr but doesn't seem to work with the given example: https://youtu.be/4N-zMfxoigQ , maybe I'm missing smth

r4d1sh:
> Just tested this pr but doesn't seem to work with the given example: https://youtu.be/4N-zMfxoigQ , maybe I'm missing smth

You drink a sobering coffee while being completely sober, and as expected, nothing happens

CraftedRO:
ah I got it, I have to drink smth before, well now it works: https://youtu.be/gdjO3SZGrRs

CraftedRO:
maybe we'll see this man's work merged this decade, just saying :smirk:

r4d1sh:
What's going on 😄

Shauren:
nothing

PR #29751, opened 2024-02-25
Build: Set CMake policy CMP0144 to NEW

ghost:
**Changes proposed:**

-  Set CMake policy CMP0144 to NEW
-  (similar to commit 4d7bf20 on master)
- Notable diff: extra documentation in comments

**Issues addressed:**

Only affects users with Cmake version 3.27 and newer (3.28 and soon to be released 3.29)
Cmake manual page: https://cmake.org/cmake/help/v3.27/policy/CMP0144.html

**Tests performed:**

Builds just fine, removes a warning in Cmake when using new versions 3.27 and above.
I have used this cherry-picked commit for several months now to avoid the old warning.

No known issues, this is basically the same content as commit 4d7bf20 on master.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,10 @@ if(POLICY CMP0074)
   cmake_policy(SET CMP0074 NEW) # find_package() uses <PackageName>_ROOT variables
 endif()
 
+if(POLICY CMP0144)              # New in version 3.27:
+  cmake_policy(SET CMP0144 NEW) # find_package() uses upper-case <PACKAGENAME>_ROOT variables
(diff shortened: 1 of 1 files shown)


Shauren:
Closing, ported together with many other build related commits

Legoso:
> Closing, ported together with many other build related commits

Fair enough. I was worried it would take another half year to see the cherry-pick.

CraftedRO:
half year it is still reasonable, I wonder if we have to create a dummy cherry pick to force a real cherry pick for each unmerged thing 😅

Legoso:
Problem solved ... for now. ^^

PR #29755, opened 2024-02-25
Core/Characters: Fix error when changing race/faction with some races

### LINKED ISSUE #29754: Faction (and race) change to Kul Tiran / Pandaren not working
### Description

When attempting to change race, the worldserver reported the error "Could not find language data for race" and errored the race change on the client. However, it still updated the characterCache and in the case of changing faction, any subsequent attempts to were met with "Must change to opposing faction" because the characterCache race no longer matched the characters actual race.

### Expected behaviour

Changing race/faction should not be blocked by the error "Could not find language data for race"

Create a horde character
Run the command .char changefaction <CharacterName>
Log out
Attempt to perform the faction change to the Kul Tiran race

### Branch

master

### TC rev. hash/commit

412610e495684cfa8eb2f4ff7c25d450e1a016ea

### Operating system

Windows

### Custom changes

None applicable to the nature of this ticket

Razinao:
**Changes proposed:**

- Add missing races to the ChangeRaceFaction opcode callback that were preventing a successful race/faction change
- Reduce repeated code

**Issues addressed:**

Closes #29754

**Tests performed:**

Confirmed that race change to Kul Tiran/Pandaren worked after these changes, rather than erroring (yet updating characterStore which broke the change faction callback until server restart)

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -2274,7 +2274,8 @@ void WorldSession::HandleCharRaceOrFactionChangeCallback(std::shared_ptr<WorldPa
         trans->Append(stmt);
 
         // Race specific languages
-        if (factionChangeInfo->RaceID != RACE_ORC && factionChangeInfo->RaceID != RACE_HUMAN && factionChangeInfo->RaceID != RACE_MAGHAR_ORC)
+        static const std::set<uint8> racesWithoutSpecificLanguages = { RACE_ORC, RACE_HUMAN, RACE_MAGHAR_ORC, RACE_KUL_TIRAN };
+        if (racesWithoutSpecificLanguages.find(factionChangeInfo->RaceID) == racesWithoutSpecificLanguages.end())
         {
             stmt = CharacterDatabase.GetPreparedStatement(CHAR_INS_CHAR_SKILL_LANGUAGE);
             stmt->setUInt64(0, lowGuid);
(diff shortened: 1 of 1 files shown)


src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -2274,7 +2274,8 @@ void WorldSession::HandleCharRaceOrFactionChangeCallback(std::shared_ptr<WorldPa
         trans->Append(stmt);
 
         // Race specific languages
-        if (factionChangeInfo->RaceID != RACE_ORC && factionChangeInfo->RaceID != RACE_HUMAN && factionChangeInfo->RaceID != RACE_MAGHAR_ORC)
+        static const std::set<uint8> racesWithoutSpecificLanguages = { RACE_ORC, RACE_HUMAN, RACE_MAGHAR_ORC, RACE_KUL_TIRAN };
```
Shauren, line 2277: use std::array, dont make it static and use std::find on it

Shauren (changes requested):

src/server/game/Handlers/CharacterHandler.cpp
Razinao: Can do, is there a particular reason to not use static though?

Shauren (approved):

src/server/game/Handlers/CharacterHandler.cpp
Shauren: Not a strong reason - using static local variables might introduce thread-safe initialization code for that variable (wasn't happening with std::array but was with std::set, and that one was also allocating memory with `new`). Since we don't need actually need this to be a shared "global" variable, a non-static one works just as well

PR #29773, opened 2024-03-02
Add Windows x86_64 CI (github actions) support.

Fabi:
**Changes proposed:**

- Windows x86_64 CI support through github actions

This is only for the master branch.
It should finish in about 25 minutes most of the time. Sometimes around 20 minutes. That's way faster than appveyor!

**Additional Info**

- I used the latest versions for boost and openssl. If you want some older, just change the link and install folder
- I did set a retention period. Left it in the yml file just in case you want to specify it
- I did not add a badge since you might want to port it to the other branches too. I only do it for master.
- This is intended to replace appveyor.
- Only one artifact with symbols is uploaded. If people use it and get crashes we want proper crash logs.

Notes for other branches:
335: branch requires copying the legacy dll. I did not test it there at all.
classic: Should work out of the box

**Tests performed:**

Tested with github actions on the master branch.


Faq:
github services not often down. It is a few hours max, or what you meant?

Aokromes:
> > * This is intended to replace appveyor.
> 
> We prefer to have different CIs so if one is down, we can rely on the others.

maybe we can have both?

Fabi:
> > * This is intended to replace appveyor.
> 
> We prefer to have different CIs so if one is down, we can rely on the others.

Then addon it as another CI for Windows should be fine no? Currently there are only advantages in doing that 🤔

PR #29775, opened 2024-03-02
Core/Quests: Handle quest 29408 more blizzlike

aquadeus:
**Changes proposed:**

-  Fixes in quest 29408 to make it more blizzlike

**Tests performed:**

Tested in-game


src/server/scripts/Pandaria/zone_the_wandering_isle.cpp
```diff
@@ -400,10 +402,34 @@ struct npc_instructor_zhi : public ScriptedAI
...
+};
+
+// 210986 - Edict of Temperance
+struct go_edict_of_temperance : public GameObjectAI
+{
+    go_edict_of_temperance(GameObject* go) : GameObjectAI(go) { }
+
+    bool OnReportUse(Player* /*player*/) override
+    {
+        me->SendGameObjectDespawn();
```
mdX7, line 417: use `me->DespawnOrUnsummon`, it also takes care of GameObjectDespawn packet. You might want to set a forced respawntime here btw

(same for spell hit)

aquadeus: ok now works as intended

PR #29793, opened 2024-03-09
Fix Envoker Initial quest pickup

ghost:
**Changes proposed:**

-  Quest Awaken, Dracthyr was set as an auto accept quest incorrectly. Fixed.
-  Ally should be a quest giver so added.
-  Added missing spell click data.

**Tests performed:**

Builds and tested in game.

- [ ] 
- [ ]

### DIFF
--- a/src/server/scripts/DragonIsles/zone_the_forbidden_reach.cpp
+++ b/src/server/scripts/DragonIsles/zone_the_forbidden_reach.cpp
@@ -24,6 +24,7 @@
 #include "SpellInfo.h"
 #include "SpellScript.h"
 #include "Log.h"
+#include "ObjectMgr.h"
 
 enum DracthyrLoginSpells
 {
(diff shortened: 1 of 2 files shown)


malcrom:
image

Dialog open and quest accepted.

src/server/scripts/DragonIsles/zone_the_forbidden_reach.cpp
```diff
@@ -151,6 +153,23 @@ class spell_dracthyr_summon_dervishian : public SpellScript
     }
 };
 
+// 366636 - Stasis
+class spell_stasis_3 : public SpellScript
+{
+    void AcceptQuest()
```
Nyr97, line 159: It's best to use Load override to check if the caster is a player, so that the spell doesn't load at all. This will prevent sObjectMgr from being called in AfterHit in case GetHitUnit() is not a player. Either that or checking if the target is the player previous to obtaining the questTemplate.

malcrom: I'm not sure what you mean.

malcrom:
It seems spell casting is garbage in TC. Cant get anything to work correctly. The current scripting is a hack and I can't get spells to cast and update before other spells are cast. The teleport spell in script does not update player position before spawning ally spell which is cast directly after. Works on official but not on TC.

Issue #25991, opened 2021-01-30
Creature: Get rid of curhealth & curmana


offl:
**Description:**

curhealth & curmana are used to set current health or mana  if you want to make unit look dying or in cases like Murmur or Valithria Dreamwalker
From what I know Blizz uses no curhealth or curmana. They uses spells or calls function directly in script to reduce health by percent. In case of spells it may be random percents. They don't change HP to 52422, they changes it to 50% of maxhealth

What I'm suggesting is handle these cases by scripts and finally remove curhealth & curmana, we don't really need 2 columns to hackily support HP of 20 creatures
```sql
SELECT * FROM creature a LEFT JOIN creature_template b ON a.id=b.entry WHERE b.regenhealth=0 order by id
```
Btw Valithria Dreamwalker is spawned twice just to support different curhealth in normal and heroic

curhealth may be used for these
```sql
SELECT * FROM smart_scripts WHERE action_Type=102
```
And if creature uses SetRegenerateHealth, they all should be checked too

**Expected behaviour:**

Migrate values to scripts and kill curhealth & curmana

**Branch(es):** both
**TC rev. hash/commit:** 1ff4a7e75b56

offl:
>curhealth may be used for these
> ```sql
> SELECT * FROM smart_scripts WHERE action_Type=102
> ```

Nope

offl:
[5] OBJECT_FIELD_GUID: Full: 0xF1300040D9000DFE Type: Creature Entry: 16601 (Dying Blood Elf) Low: 3582
[5] UNIT_FIELD_HEALTH: 1
[5] UNIT_FIELD_MAXHEALTH: 168/2.35E-43

That is not easy to support if HP is always 1
Maybe https://tcubuntu.northeurope.cloudapp.azure.com/aowow/?spell=28838 if not used in Naxx and if other spells with same name are used for same thing but I doubt it's how it works

offl:
I don't like this more and more
Note for future: don't add auras, dynflags and crap like that, just change curhp to 1 and curmana to 0 and ensure there will be no regen
I don't wanna make no mana regen in SAI, no hp regen in c_t, set mana to 0 in SAI and set curhp to 1 in creature
At least right now

[23] OBJECT_FIELD_GUID: Full: 0xF1300045686AAE08 Type: Creature Entry: 17768 (Blood Knight Stillblade) Low: 6991368
[23] UNIT_FIELD_HEALTH: 1
[...]

PR #29801, opened 2024-03-12
Core/Creatures: Changed the spawn health field to a percentage

### LINKED ISSUE #16756: Creature spawn behavior with `creature_template`.`RegenHealth`=0
https://github.com/TrinityCore/TrinityCore/commit/c9f9e93c227103b405b382bd2aadb6dd83fa3c72

If RegenHealth=0, creature takes in the health value from `creature`.`curhealth`. This kind of sucks. since creatures can be in several modes with several different health values, and the default values on the DB cause nasty issues if you set RegenHealth=0, no warnings given. You need to 'hack' and set the health value to some health larger than all modes, if you want the creature to have max health. It's not very portable.

I suggest changing `creature`.`curhealth` to a signed int, and set it to -1 for every creature that doesn't use RegenHealth in it's template. It will be equivalent to max health.

That way one could also remove Creature::setRegeneratingHealth from the cpp scripts that mostly don't have to be there, and just set the `creature_template`.`RegenHealth` values accordingly. At least without concerns for scripts breaking because creatures suddenly have 1 health.

meji46:
**Changes proposed:**

-  Move curhealth to curHealthPct
-  Drop curmana (auras/spells handle special cases)
-  Removed check from Unit::SetPowerType that caused creatures to appear with 0 mana (`GetPowerType() == new_powertype` -> During init, all creatures use DisplayPowerType Mana (0) so Unit::SetPowerType(POWER_MANA) does nothing)
-  Added spell scripts for absolute health values (special cases)
-  Update some scripts

**Issues addressed:**

Closes #16756
Updates #25991

**Tests performed:**
Builds, tested in-game.

- Cases of health < 100% are controlled by serverside spell casting. Since we have no reference of which ones are used, I've preferred to preserve the health field


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -5531,12 +5531,30 @@ void Unit::SendAttackStateUpdate(uint32 HitInfo, Unit* target, uint8 /*SwingType
     SendAttackStateUpdate(&dmgInfo);
 }
 
-void Unit::SetPowerType(Powers new_powertype, bool sendUpdate/* = true*/)
+void Unit::InitPowerType()
```
Shauren, line 5534: Why is this in Unit if only creatures use this? Also about `UnitsUseDefaultPowerOnInit` - `Unit` means Creature, not our Unit class

meji46: My idea was to use this function also for players (flag SetToMaxOnInitialLogIn)

About UnitsUseDefaultPowerOnInit, there are some PowerTypes with this flag but without the flag IsUsedByNPCs

Shauren (approved):

PR #29835, opened 2024-03-26
Scripts/DH: Fix chaos strike spell

aquadeus:
**Changes proposed:**

-  Fix chaos strike spell

**Tests performed:**

Tested in-game


Shauren:
This is completely wrong, casting 344862 should cause 162794 or 228477 to be cast instead using spell override mechanic (https://wago.tools/db2/SpecializationSpells?filter[OverridesSpellID]=344862&page=1)

PR #29836, opened 2024-03-27
Core/GameObjects: Implement destructible_hitpoint table

Golrag:
**Changes proposed:**

-  Implement table to store health info about destructible buildings

**Issues addressed:**

Closes: some todo's in code

**Tests performed:**

- Builds locally
- Tested in IoC & SotA

- None

Data is based on mop data, when the fields were still present in gameobject_template. I think the max health can still be sniffed, damaged value will be guess work


src/server/game/Entities/GameObject/GameObjectData.h
```diff
@@ -839,6 +848,8 @@ struct GameObjectTemplate
     std::string StringId;
     WorldPacket QueryData[TOTAL_LOCALES];
 
+    ::DestructibleHitpoint const* DestructibleHitpoint;
```
Shauren, line 851: I would move this to `GameObjectValue` instead

Shauren (changes requested):

src/server/game/Entities/GameObject/GameObjectData.h
Golrag: alright!

Golrag: Like this then? This might cause the console logs to be a bit more spammed now that the error message is on gameobject create instead of during load times

Shauren: You can still do validation on startup, like other gameobject_template fields that reference other tables

Golrag: should be done!

Issue #29121, opened 2023-07-04
ICC - Sindragosa should not be flying during the ground phase, spreading its wings


qq462518439:
### Description

When you engage in a battle with Sindragosa, if the boss is in the ground phase and still attacking you in a flying state.
The lower body is underground, which looks strange
![QQ截图20230704151326](https://github.com/TrinityCore/TrinityCore/assets/91646500/781e16de-775b-446a-99d0-576f84478dc3)

### Expected behaviour

Sindragosa should not be flying during the ground phase, spreading its wings

### Steps to reproduce the problem

1. Start meeting Sindragosa and wait for it to land
2. Move your cartoon characters appropriately after starting the battle
3. Meet again and watch

### Branch

3.3.5

### TC rev. hash/commit

https://github.com/TrinityCore/TrinityCore/commit/96f2a97e109befe6d8612e6e8ddc6e68a8bd1dd3

### Operating system

Windows 10 x64

### Custom changes

npcBots
Eluna

Krutok:
But this is not the original model of Sindragosa: https://wow.zamimg.com/uploads/screenshots/normal/170272-sindragosa.jpg

Shauren:
It is in chinese client (they dont have bones)

Krutok:
ahh ok

qq462518439:
![QQ截图20230712124138](https://github.com/TrinityCore/TrinityCore/assets/91646500/e01faab2-1688-48f5-ae9f-21d2d16ff983)
I tested the same situation with the original client again, so it seems to have nothing to do with the model

Ovahlord:
What flags_extra does Sindragosa have with .npc info?

qq462518439:
@Ovahlord 
Flags Extra: 268435457

PsychoJS:
any solution?

qq462518439:
> any solution?
Do you have the same problem
Perhaps refer to the source code of azerothcore?But I don't know the code

vahzxz:
I think this is a common problem. NPCs stupidly do not know how to keep the phase they are in, dragons also behave terribly. Any Dragon, if you drag it down and it descends to the ground, this problem manifests itself. It is required to implement a system that preserves the shape of the NPC when it touches the ground. I think these large-scale problems should be solved.

CraftedRO:
Fix suggested by Ovahlord
```sql
[...]

PR #29853, opened 2024-03-29
Core/Movement: fix air state condition

### LINKED ISSUE #29819: [Movement] - There is no change of state when in the air
### Description

It is required to implement a system that changes the state of mobs on the move, depending on whether they can stand with their feet on the ground or not.
Since the problem is large-scale and the dragons do not behave adequately, and in principle any NPCs in the game world. Having done this, many issues will be solved

### Expected behaviour

Any flying NPC must understand that they are on the ground and turn off the flight phase

When attacked by any flying NPC, it falls into the ground.

### Branch

3.3.5

### TC rev. hash/commit

7a4674c5ace966ecc7f067ad7687bfcab2a1777f

### Operating system

Win Server 2022

### Custom changes

No

AntonK93:
**Changes proposed:**
Problem is that calculation to detect `isInAir` has been made only using the strict calculation of `ground Z + 0.5` value which animation doesn't work for most of the flying creatures(mostly dragons) , they should change the animation type depends on the size of their model which is unique per model and distance from the ground, e.g. sindragosa during the attack has difference about 0.8f between ground and her position which leads to the flying state since it's bigger than 0.5

After the change it takes the `collisionHeight` which is 6 for sindragosa and since we need to set it to flying state when half of the model goes up we divide it by 2

and it automatically will go up as soon as it reach pos + colission > ground Z 

Closes #29819 #29121

**Tests performed:**
tested in game

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2653,7 +2653,8 @@ void Creature::UpdateMovementFlags()
     float ground = GetFloorZ();
 
     bool canHover = CanHover();
-    bool isInAir = (G3D::fuzzyGt(GetPositionZ(), ground + (canHover ? GetFloatValue(UNIT_FIELD_HOVERHEIGHT) : 0.0f) + GROUND_HEIGHT_TOLERANCE) || G3D::fuzzyLt(GetPositionZ(), ground - GROUND_HEIGHT_TOLERANCE)); // Can be underground too, prevent the falling
+    float collisionHeight = GetCollisionHeight() / 2.0f;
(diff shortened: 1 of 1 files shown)


Ovahlord:
which flags_extra does Sindragosa have?.

AntonK93:
> which flags_extra does Sindragosa have?.

`268435457`, but i have tested it against some other npc dragons with different sizes, all of the flying creatures have this issue

Ovahlord:
Add extra flag 0x200 to her.

AntonK93:
> Add extra flag 0x200 to her. Without that change that is.

you're right that it will work for the boss since the whole air phase is scripted specifically, but as i told there are many dragons in the world without any scripts and when it lands it wrongly flies when it supposed to walk, expected to add a "dragon" script for all of them ?

Ovahlord:
flight movement as it stands right now in 335 is fundamentally wrong implemented. transitions between flight and grounded states are always suposed to be scripted or part of static flags which are not implemented in 335. So the fix that I proposed with the extra flag is the correct approach for the listed issue.

CraftedRO:
is that flag in sniffs ?

PR #29877, opened 2024-04-01
Battleground/BattleForGilneas: Implementation

Golrag:
**Changes proposed:**

-  Implement Battle for Gilneas battleground

**Issues addressed:**

**Tests performed:**

- Builds locally
- Clicked on capture points, tested graveyards

- Cosmetic spawns are not implemented
- Tried using namespaces &  constexpr for variables instead of enums so we have them scoped & can track unused ones


src/server/scripts/Battlegrounds/BattleForGilneas/battleground_battle_for_gilneas.cpp
```diff
@@ -15,18 +15,299 @@
...
+#include "Battleground.h"
+#include "BattlegroundPackets.h"
 #include "BattlegroundScript.h"
 #include "ScriptMgr.h"
 
+namespace BattleForGilneas
+{
+    namespace GameObjects
+    {
+        static constexpr uint32 CapturePointLighthouse = 228050;
```
Ovahlord, line 27: why not going with enums instead? Since you already have a namespace for Battle for Gilneas you can also just go with enum GameObjects

Golrag: With this we can get warnings about unused variables, don't think we can with enums

Ovahlord: Turning enums into variables only for extra warnings doesn't sound like something that's worth the trouble, considering that you're basically allocating memory for compile-checks only, which IMO should be either handled via static asserts or runtime asserts, informing the user that something is missing. Like for example you should define a maximum number of expected capture points and throw and assert if the setup did not register all of them.

mdX7: Notice they are flagged as constexp. Their usages are literally replaced with its values at compile time, theres no memory disadvantage at runtime

Shauren (approved):

PR #29898, opened 2024-04-05
Core/Settings: Change in settings loading sequence

RioMcBoo:
Fix : sCfgCategoriesStore hasn't got data yet before using

Ported from:
https://github.com/CypherCore/CypherCore/commit/6abce678ca5b29c308afdf68f90458d1e1d00045


RioMcBoo:
hmm, i had not any errors while compile

Shauren:
Not good, this is lost when you .reload config

RioMcBoo:
> Not good, this is lost when you .reload config

Then can we just copy this code in two spots?

PR #29915, opened 2024-04-11
Core/Creatures: Allow profession tools to be used as equipment

meji46:
**Changes proposed:**
-  Allow profession tools to be used as equipment

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game: [Lee Sunspark](https://www.wowhead.com/npc=57617/lee-sunspark)

None


src/server/game/Entities/Item/ItemTemplate.h
```diff
@@ -415,6 +415,20 @@ enum InventoryType : uint8
 
 #define MAX_INVTYPE                               35
 
+constexpr std::array<InventoryType, 10> InventoryTypesEquipable =
```
Shauren, line 418: Eh, making it local in ObjectMgr::LoadEquipmentTemplates seems fine (I don't see another use for this)

meji46: True, I don't know why I was thinking about validations for SAI. Fixed

Shauren: Maybe because SAI can set random items (not validated) if action_param1 = 0

meji46: Sooo... should I move it back to where I put it? I don't like the idea of being able to try to equip any item in that SAI action

Issue #17366, opened 2016-06-14
[6.x] Bnet Authenticator


Nelidon:
**Description**: The bnetserver should be able to ask to the user to enter a digital code from Google Authenticator (example) to log in

**Current behaviour**:

**Expected behaviour**:

**Steps to reproduce the problem**:

**Branch**: 6.x

**TC hash/commit**:  2fe6fc63d79655a96ee2135a6b380ce353729088

**TDB version**:  6.04

**Operating system**:  Linux

[//]: # (This template is for problem reports, for other type of reports edit it accordingly)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing c++ create a Pull Request)

stas730:
For 3.3.5 too.

Aokromes:
Unless someone broke it for 3.3.5a was working.

Nelidon:
Not the same system.
3.3.5 use(d) token stored in database.

stas730:
@Aokromes, which source file? And how to use it?

stas730:
UPD: In previous issue, you said "We can't implement Battle.net functions". WHY?

tkrokli:
@stas730 : in issue https://github.com/TrinityCore/TrinityCore/issues/17483#issuecomment-229710801 by DDuarte :
"The 3.3.5 Battle.net module is heavily encrypted. It is out of the scope of TC to work with that."

Like Nelidon suggested in his top post, Google Authenticator or similar is an open source solution that TC can and may work with instead. If you feel so strongly about using Blizzard's authenticator, play retail or find someone else who are willing to work with you on hacking a solution for using it in your own source. As stated before, it is not within the scope of TrinityCore's work to use it like you want.

stas730:
TrinityCore team are idiots, which thinks "We can develop WoW server without sniffing".
X(
Now, I will never post issues about TrinityCore. I need to find more friendly team or open source MMORPG.

Nelidon:
@stas730 stop crying like a child and leave us.
Better for all of us.

PR #29931, opened 2024-04-19
Core/Bnet: Added missing 'next_url' field to Login.proto

superp00t:
**Changes proposed:**

This PR modifies the Login.proto file to add a string field `next_url` to `LoginResult`.

This is required if we want to implement TOTP authenticator support in master branch as well as in 3.3.5, as issue #17366 suggests.

When `LoginResult.authentication_state` is equal to `AUTHENTICATOR` after the POST request to `/bnetserver/login`, the client looks for a `next_url` field to know where to submit the authentication code.

The presence of this field will trigger the client to display a text input menu for the authenticator code.

One the client submits this code, a POST request is made to the URL in `next_url`. The body of this request is a `LoginForm`, and the response is a `LoginResult`. This form contains two input fields: a text value, `authenticator_input` which is the user's text input,
and `remember_authenticator` which the client sets to `"true"`.

The client expects the result from this last request to be the same as if the the authenticator was not required; the authenticator URL response should contain `server_evidence_M2` and `login_ticket`.

**Issues addressed:**

This PR alone does not implement #17366, but may help.

**Tests performed:**

Verified the basic process of requesting authenticator code from the client.

- TODO: An actual implementation of TOTP in Bnet that will make use of this

### DIFF
--- a/src/server/proto/Login/Login.pb.cc
+++ b/src/server/proto/Login/Login.pb.cc
@@ -171,13 +171,14 @@ void protobuf_AssignDesc_Login_2eproto() {
       ::google::protobuf::MessageFactory::generated_factory(),
       sizeof(SrpLoginChallenge));
   LoginResult_descriptor_ = file->message_type(6);
-  static const int LoginResult_offsets_[6] = {
+  static const int LoginResult_offsets_[7] = {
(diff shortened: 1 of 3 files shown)


MaxtorCoder:
You have to generate the proper files .c/.h files as well, you can do this by using [https://github.com/TrinityCore/TrinityCore/tree/master/contrib/protoc-bnet](protoc-bnet)

PR #29940, opened 2024-04-25
Core/Creature: implement overriding creature static flags based on spawnId and difficultyId

Ovahlord:
**Changes proposed:**
As discussed internally and requested by DB devs, this is a way of implementing a per spawn and difficulty function to override single static flags fields

**Tests performed:**
- none yet. will soon happen


sql/updates/world/master/2024_xx_xx_xx_world.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE `creature_static_flags_override` (
```
mdX7, line 1: missing `DROP TABLE IF EXISTS` or use a procedure which checks if table exists before executing this statement

src/server/game/Entities/Creature/Creature.h
```diff
@@ -557,6 +557,8 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         time_t _lastDamagedTime; // Part of Evade mechanics
         CreatureTextRepeatGroup m_textRepeat;
 
+        // Draws data from m_creatureDifficulty and spawn/difficulty based override data and returns a CreatureStaticFlagsHolder value which contains the data of both
+        CreatureStaticFlagsHolder const GenerateStaticFlags(CreatureDifficulty const* creatureDifficulty, ObjectGuid::LowType spawnId, Difficulty difficultyId) const;
```
Shauren, line 561: there should be no const on return by value

src/server/game/Entities/Creature/CreatureData.h
```diff
@@ -317,6 +317,15 @@ class CreatureStaticFlagsHolder
     void ApplyFlag(CreatureStaticFlags7 flag, bool apply) { if (apply) _flags7 |= flag; else _flags7 &= ~flag; }
     void ApplyFlag(CreatureStaticFlags8 flag, bool apply) { if (apply) _flags8 |= flag; else _flags8 &= ~flag; }
 
+    EnumFlag<CreatureStaticFlags> const& GetFlags() const { return _flags; }
```
Shauren, line 320: EnumFlag is trivial to copy, dont return a ref

PR #29945, opened 2024-04-27
Implemented OnHeartbeat AuraScript hook and refactored an aurascript to use it as example

Ovahlord:
**Changes proposed:**

-  OnHeartbeat is now available as global AuraScript hook
-  Amalgam's Seventh Spine is now using said hook to update its aura

**Tests performed:**
- tested ingame

### DIFF
--- a/src/server/game/Spells/SpellScript.h
+++ b/src/server/game/Spells/SpellScript.h
@@ -1048,6 +1048,7 @@ enum AuraScriptHookType
     AURA_SCRIPT_HOOK_CHECK_AREA_TARGET,
     AURA_SCRIPT_HOOK_DISPEL,
     AURA_SCRIPT_HOOK_AFTER_DISPEL,
+    AURA_SCRIPT_HOOK_ON_HEARTBEAT,
     AURA_SCRIPT_HOOK_ENTER_LEAVE_COMBAT,
     // Spell Proc Hooks
     AURA_SCRIPT_HOOK_CHECK_PROC,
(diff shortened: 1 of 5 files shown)


src/server/game/Spells/SpellScript.h
```diff
@@ -1174,6 +1175,58 @@ class TC_GAME_API AuraScript : public SpellScriptBase
...
+                    return (static_cast<ScriptClass*>(auraScript)->*reinterpret_cast<ScriptFunc>(callImpl.Member))();
+                };
+            }
+            else
+            {
+                static_assert(std::is_invocable_r_v<void, ScriptFunc>,
+                    "AuraHeartbeatHandler signature must be \"static void HandleHeartbeat()\"");
+
+                _callImpl = { .Static = reinterpret_cast<decltype(AuraHeartbeatFnType::Static)>(handler) };
+                _safeWrapper = [](AuraScript* /*auraScript*/, DispelInfo* dispelInfo, AuraHeartbeatFnType callImpl) -> void
```
Shauren, line 1214: dispelinfo?

Ovahlord: fixed

src/server/scripts/Spells/spell_item.cpp
```diff
@@ -4345,7 +4337,7 @@ class spell_item_amalgams_seventh_spine : public AuraScript
             if (target->GetPrimarySpecialization() != spec)
                 target->RemoveAurasDueToSpell(aura);
             else if (!target->HasAura(aura))
-                target->CastSpell(target, aura, aurEff);
+                target->CastSpell(target, aura, GetEffect(EFFECT_0));
```
Nyr97, line 4340: Would we not need to validate this effect since OnHeartbeat is not really hooking any and implicitly validating any?

PR #29946, opened 2024-04-27
Core/Creatures: moved combat pulse into heartbeat handling and implemented CREATURE_STATIC_FLAG_2_FORCE_PARTY_MEMBERS_INTO_COMBAT

Ovahlord:
**Changes proposed:**

-  the combat pulse mechanic is being moved into heartbeat handling and now fires once on engaging a creature
-  This mechanic will now only kick in for creatures that use CREATURE_STATIC_FLAG_2_FORCE_PARTY_MEMBERS_INTO_COMBAT
- this furthermore no longer binds the mechanic to dungeons but it can now also act in the open world since we now draw our targets from combat references instead of iterating over the entire map.

**Tests performed:**
- tested ingame with a 2nd character in a party

**Todo**
- [x] The static flag must be set in DB for all (raid) bosses now


Nyr97:
Are you sure this must be set for all bosses? Asking because I've been recently playing wotlk classic and on low level dungeons, for example, Deadmines, when engaging a boss, I wouldn't be dragged into combat if my team mates engaged it while I was drinking. And as far as I recall, engaging would break drinking.

Ovahlord:
It is.

Ghaster:
Flag was also added (at earliest) very late in Vanilla or very early TBC so there's a chance that earlier Vanilla bosses simply didn't have the flag until later.

Ovahlord:
It applies primarily for raid bosses only. Dungeon bosses are not part of this mechanic for most of the part.

Nyr97:
I see then, thanks for the clarification!

Nyr97:
Since you've removed DoZoneInCombat(), this means that, in raid encounters where there are creatures that engage the boss when players do, e.g., Archimonde in Hellfire Citadel or Sylvanas Windrunner in Sanctum of Domination, we're forced to find those creatures and call AttackStart on the bosses, is that it?

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -3644,6 +3624,9 @@ void Creature::AtEngage(Unit* target)
         ai->JustEngagedWith(target);
     if (CreatureGroup* formation = GetFormation())
         formation->MemberEngagingTarget(this, target);
+
+    // Creatures with CREATURE_STATIC_FLAG_2_FORCE_PARTY_MEMBERS_INTO_COMBAT periodically force party members into combat
+    ForcePartyMembersIntoCombat();
```
moJIto-ice, line 3629: hmmmmmmm,. group updates packets also used to be updated here ( i meen health, auras of group members etc ) or it should be called from ::Heartbeat( ) ?

Ovahlord:
> Since you've removed DoZoneInCombat(), this means that, in raid encounters where there are creatures that engage the boss when players do, e.g., Archimonde in Hellfire Citadel or Sylvanas Windrunner in Sanctum of Domination, we're forced to find those creatures and call AttackStart on the bosses, is that it?

DoZoneInCombat only put all players (and their summons) on the map in combat with the target creature. So it was just making sure that when you engage the boss (e.g. Tank start attacking) that all players were put in combat which is only correct for raid bosses and this is now done via static flag and only for party members, not for all players on the map.

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -3662,6 +3645,38 @@ void Creature::AtDisengage()
...
+        if (!player || player->IsGameMaster())
+            continue;
+
+        if (Group const* group = player->GetGroup())
+            partiesToForceIntoCombat.insert(group);
+    }
+
+    for (Group const* partyToForceIntoCombat : partiesToForceIntoCombat)
+    {
+        for (auto const* itr = partyToForceIntoCombat->GetFirstMember(); itr != nullptr; itr = itr->next())
```
Shauren, line 3669: bad auto

Shauren (changes requested):

Shauren (approved):

Issue #17564, opened 2016-07-10
Cleanup bags button freezes inventory


azeroc:
**Description**:
When you go into your backpack and click on "Cleanup bags" button (top-right corner of the bag UI) your items freezes (their icons turn gray). In order to unfreeze them you have to relog.

**Current behaviour**: Clicking on "Cleanup bags" button (top-right corner of the bag UI) your items freezes (their icons turn gray)

**Expected behaviour**: Items in the bag should be sorted or "cleaned up" like in retail or at the very least the button should not freeze your items and do nothing instead (until cleanup feature is implemented).

**Steps to reproduce the problem**:

1. Open your character's backpack
2. Click on "Cleanup bags" button at top-right corner of bag UI.

**Branch(es)**: 6.x

**TC hash/commit**:  22d29ce0494c32a73e20646d2320a44a5eb2e633

**TDB version**:  6.04

**Operating system**:  Windows 10 64bit

**P.S.** While it is quite a minor issue, my suggestion would be to disable button all-together (user presses it and nothing happens) until its functionality is implemented.
[//]: # (This template is for problem reports, for other type of reports edit it accordingly)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing c++ create a Pull Request)

tkrokli:
TrinityCore rev. 22d29ce0494c 2016-07-10 11:27:38 +0200 (6.x branch) (Win64, Release, Static)
TDB 6.04 plus updates up to and including `2016_07_05_02_world.sql`

Confirmed, any bag sorting or bag GUI utility use freezes the bag contents until reboot + delete Cache.

frozen4:
Confirmed, any bag sorting or bag GUI utility use freezes the bag contents until reboot

Kodekc:
Confirmed, 7.2.5.24461 The bug is present.

PR #29947, opened 2024-04-28
Fix reagent bag & add sorting bag

Bloodtigress:
The bag cleanup order are : 
- hearthstone
- Food and drink
- Weapon
- Armor
- All other items
- Grey (poor) quality items

**Changes proposed:**

-  Fix an issue (not logged) where a new looted bag go to the reagent bag instead of inventory if all bag slot are taken
-  Add cleanup item fonctionality for: backpack, bags, bank and reagent bank
- Replace some 'magic number' in ItemHandler with enum value 

**Issues addressed:**
#17564
#17564

**Tests performed:**
- Build OK
- Tested ingame for all functionality

- None that I know of


Shauren:
Sorting logic is a lot more complex than just those five classifications, then by entry (by the use of sorted map)
For example: 
* weapons and armor are additionally sorted by slot and itemlevel (descending)
* junk (poor items) go to the end of last bag instead of directly following all other items
* bags can be marked as "preferred" for a category of items during sorting (click on bag icon when its open)
* bags can be marked as ignored for sorting

Auto equipping bag change is problematic, it causes these bags to be unable to be looted to backpack (in case all bag slots are full and the bags are filled)

Bloodtigress:
Here are my answers
- I could do more complex logic. but I need to know what is the logic of the sorting. Right now, I don't know all the logic of the sorting. I remembered that the first item was always the hearthstone. After that I know it's the food/drink. After that comes the armors/weapons (I don't know which one comes first)

- For the junk item I can change that

- About the flag of bags, I can't do that right now because the code does nothing when it receives the change flag message (CMSG_CHANGE_BAG_SLOT_FLAG). I'd like to add the code to do taht (or at least try) but I need to know what is the message we need to send to the client. Once this is done then I think I could take into account the flag of each bag.

- Finally about the auto equip I'm working on it. I think I know how to fix this. But I need to do some test first

If you have any information about the sorting order, let me know and I'll change the code

Bloodtigress:
OK I've just push the modification of the code.
So the issue about the bag not going to the backpack is fixed
And now, the grey item will go the last bag instead of right after the other items

Once I know more about the sorting complexity, I'll change the way it works.

By the way I've seen an bug some bugs about the reagent bag :
- When you log out the bag in the reagent slot is empty and you receive the new bag in your mail box when you had a bag
- You can't equip the reagent bag when you have one in your inventory

Shauren:
bag flags implemented - cc9241784491198581a19f3f5502a42d048ea063

Bloodtigress:
First of all thanks very much for the implementation of the bag flags system. With that I was able to take the bag flags into account!
I really wanted to use them so really thanks!
It took me a while to figure out how to use them because there was a strange behaviour with those flags.
When I remove a 'Priority' flag in the game, it seems to me that another one was added. I figure out it was because only one 'Priority' flag must be set at a time (it also causes issues with the game itself like displaying only one 'Priority' flags)

I fixed that (that's why there is a modification of the HandleChangeBagSlotFlag and HandleChangeBankBagSlotFlag in the itemHandler)

After that and some testing I didn't find any issue. But if you find any, let me know so that I can fix them. 

About the last thing I need to do (the logic behind the sorting), I'll look for it later

Bloodtigress:
I update the code to add more complexity to the sorting code.
I try to make it more Blizzlike. 
It works for most of the item. But unfortunately there are still some items that are quite weird. 
For example: 
If you have Core of Earth, Elemental Fire and Essence of Fire
Then in retail you have this order Core of Earth, Elemental Fire and Essence of Fire
But with what I've done you'll have Essence of Fire, Core of Earth and Elemental Fire

There is also an issue with some special item for example Cold Iron Pick

Anyway the way the sorting is done is : 
1 - special items (hearthstone, etc.)
2 - Consummable
3 - Weapon order by Item level only in descending order (it seems to be the same way in retail)
4 - Armor order by armor type then item level in descending order
5 - All other item order by item class (ascending order) then subclass (descending order) then finally item Id (descending order)

Shauren:
Magic numbers (2c286b5fb6b23bf8354af06f884aabe98c66d254) and reagent bag slot (6f9e359701bd75705f73f23af65affb5a5bed573) were fixed, you can remove those from the PR

Bloodtigress:
I've removed the magic number from my repo as well as the reagent bag slot fix. 
I've also put the code about the storing of the item in the bag in the player files. I think it's better to put them there

KamiliaBlow:
This pull request needs to be updated due to changes in 11.2.0 related to the bank.

Bloodtigress:
@KamiliaBlow Unfortunately I don't play the game right now. So I don't know what are the changes and I can't update this PR.
You can update the code to make it work if that is what you want.

PR #29953, opened 2024-05-02
Working on guild

RioMcBoo:
Thanks Alistar, but i dk how to add him to RP


RioMcBoo:
We need also tabard hotfixes for guild saving/loading to pass tabard validation image

RioMcBoo:
I'm a little taken aback. When I change access to bank tabs (several at once), the client sends not one request, but each time a new request for each cell. 
What causes the disconnect (I assume that the socket is disconnected due to packet spam, 3 packets in a row = disc) 

When can i fix it particularly for this packets? or just test

RioMcBoo:
@jackpoz , my hope, bring to us and help to show that anybody cares for this branch

PR #29958, opened 2024-05-05
Core/Creatures: Fix level scaling for summons without SummonPropertiesFlags::UseCreatureLevel

meji46:
**Changes proposed:**

-  Fix level scaling for summons without SummonPropertiesFlags::UseCreatureLevel. Creature levels only scale between ContentTuning levels

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game

None


ipro4911:
wait so with this npcs are level scaled to u when they are spawned?

Ovahlord:
> wait so with this npcs are level scaled to u when they are spawned?

No. They use a fixed level.

PR #29960, opened 2024-05-05
Dropped support for OpenSSL 1.1, MariaDB and Pre MySQL 8.4 LTS versions

Ovahlord:
Controversal changes, might not get approved, however, someone may eventually resurrect the PR if it shouldn't get merged. At least I tried :)

**Changes proposed:**

-  The time for MySQL 5.7 has come to an end with MySQL 8.4 being the first long term support version of MySQL 8. This marks the point where we will start using MySQL 8 as default version since this is the version that has gotten rid of many functions that have been marked as deprecated for quite some time.

- Since MariaDB is still using MySQL 5.7 functions, it is no longer compatible with MySQL 8.4. Following this update, the support for MariaDB has been removed as a result of incompatability

- OpenSSL 3 is now default shipped alongside MySQL 8 and is also now the required version for users to install for consistency reasons. The backwards compatability has been removed from the core. This means that OpenSSL 1.1 has been put to rest and can no longer be used.

**Tests performed:**
- built, tested, no regressions

- [x] validate if MySQLWorkaround.h is still needed or if it can be safely removed 
- [x] tank the shitstorm of angry MariaDB and old MySQL DB users
- [ ] Maybe keep support for older MySQL 8 versions though there is no reason to do so as not even Oracle cares about them  anymore and is now focusing on 8.4 as first maintained version
- [ ] depending on if there will be something like 8.4.1 we may add some wildcard support for subversions


acidmanifesto:
Just some fyi. Openssl 1.1.x is no longer provided currently due to security exploits and has been dropped by infavor for openssl 3.x.x
The link to the openssl dependiences on the tc wiki only provides openssl 3.x.x.
As for mysql, there are alot of wide changes that will make future queries backwards compatible with earlier versions of mysql, we already seen some queries not being compatible withing 5.6 and 5.7 variation.
Mariadb should be dropped full due to the incompatibility that is growing between the two sql giants.

CraftedRO:
just logged in to wish this will not be merged to 3.3.5 at least, you had your preservation thing compatible with mysql 5.1 till almost died (has been archived) I don't see the concern about this in TC.

danielsreichenbach:
Not an active contributor, just a user. Would like to suggest to treat any external dependency just like compiler support.

Minimum version should be the versions available in the supported Linux distributions without having to add external repositories.

Effectively that means MariaDB 10.6 or newer, and for Windows one might support MySQL. I don't really know what's the issue with MariaDB but I do know that any stable non Windows system by default does not support MySQL but MariaDB. So if supporting Linux for the project is a primary object, so should being able to use MariaDB.

I do agree with dropping OpenSSL < 3.0 though since that's basically gone from any Linux distro that is considered stable.

Generally I get the sentiment of "just add external repository and get the new stuff" but it's in reality for Linux systems a huge source of frustration and problems since external repositories tend to get fucked up or dropped or ship broken packages.

So I would advise against "use the newest, cool stuff" and suggest a pragmatic approach: use what's is considered stable on the platform, and supported by default. Anything else is bad practice IMHO.

Ovahlord:
> Not an active contributor, just a user. Would like to suggest to treat any external dependency just like compiler support.
> 
> Minimum version should be the versions available in the supported Linux distributions without having to add external repositories.
> 
> Effectively that means MariaDB 10.6 or newer, and for Windows one might support MySQL. I don't really know what's the issue with MariaDB but I do know that any stable non Windows system by default does not support MySQL but MariaDB. So if supporting Linux for the project is a primary object, so should being able to use MariaDB.
> 
> I do agree with dropping OpenSSL < 3.0 though since that's basically gone from any Linux distro that is considered stable.
> 
> Generally I get the sentiment of "just add external repository and get the new stuff" but it's in reality for Linux systems a huge source of frustration and problems since external repositories tend to get fucked up or dropped or ship broken packages.
> 
> So I would advise against "use the newest, cool stuff" and suggest a pragmatic approach: use what's is considered stable on the platform, and supported by default. Anything else is bad practice IMHO.

As MariaDB is likely never going to adapt the MySQL 8 functions and is gonna stick with 5.7, which is pretty much deprecated at this point,  the end of support is inevitable. TrinityCore uses MySQL, not MariaDB. The sparse support for MariaDB was a result of it being close enough to MySQL. This has changed. MySQL 8.4 and going forward is going to deprecate and remove more and more 5.7 functionality which will lead to incompatability. In a sense, it already is to some degree.

danielsreichenbach:
Then I would suggest to consider something like Postgres instead unless you want to make it super painful for Linux folks out there. Highly doubt MySQL will return to Linux distros.

Fabi:
Finally.
Openssl 1 is eol, mysql 5.7 is eol and mariadb was never supported. So updating those to it's latest parts and dropping support for the others should have been done a long time ago.

And for linux users it is VERY common to use other repos, especially official repos for specific versions. People who say otherwise are stuck in their bubble of never using up2date software and only what an OS provides (basically very bad company like thoughts).

So basically everything that goes eol should be updated as requirement before it's eol and that is due a long time now. Do it faster :D

Ovahlord:
> Finally. Openssl 1 is eol, mysql 5.7 is eol and mariadb was never supported. So updating those to it's latest parts and dropping support for the others should have been done a long time ago.
> 
> And for linux users it is VERY common to use other repos, especially official repos for specific versions. People who say otherwise are stuck in their bubble of never using up2date software and only what an OS provides (basically very bad company like thoughts).
> 
> So basically everything that goes eol should be updated as requirement before it's eol and that is due a long time now. Do it faster :D

Well, there is one particular thing for master at least and that's the legacy openssl provider. Master is currently forced to load it because of the outdated warden implementation uses RC4 stuff, which has long been deprecated. And as long as we don't know what the new warden uses, we can't replace it. On top of one certain individual wanting to reduce differences between branches (bc 335 uses RC4 everywhere) so Arc4 is kinda stuck to stay

jackpoz:
> Then I would suggest to consider something like Postgres instead unless you want to make it super painful for Linux folks out there. Highly doubt MySQL will return to Linux distros.

https://trinitycore.info/install/requirements/linux our wiki already includes steps on how to install MySQL and we support only setups that follow our wiki instructions.

You are free to use a different DBMS and maintain any code change required by that DBMS. If you guarantee you will take care of the support, we might even accept that as a PR, but you will have to verify not just that it compiles but that any MySQL feature is supported and behaves in the same way. 

Please note that MariaDB developers say that MariaDB is not 100% compatible with MySQL at https://mariadb.com/kb/en/mariadb-vs-mysql-compatibility/#incompatibilities-between-currently-maintained-mariadb-versions-and-mysql , ask them to provide 100% compatibility and we will then support both.

jackpoz:
Considering the news about private equities being in the process of buying MariaDB https://www.theregister.com/2024/05/21/private_equity_offer_for_mariadb/ , linux users will be pissed anyway.

MySQL is not a paid product, so your example doesn't really fit. We also don't support a lot of other DBMS, like PostgreSQL, SQL Server, CosmosDB, MongoDB, JackpozDB (sad it's not supported :( ) . We cannot support every fork that MySQL has in addition to MySQL itself, we will support MariaDB when and if we will decide to switch to that,  test TC, migrate all users (which will piss off the other half that is fine with MySQL).

On the other end, all the people who like MariaDB could step up and volunteer to test all the sql queries used by TC, verify that they are compatible with each new MariaDB release and fix any incompatibility without TC devs having to do that. We don't have the time to do that.

Ovahlord:
Only because it's a default shipped software doesn't mean we need to use it. Trinity explicitetly uses MySQL and says so in the install instructions. That's - again - user laziness because fun fact: windows does not ship any database solution at all, and you know what? Nobody is complaining about having to install it in the first place. Linux default installations are not a reference point. MariaDB solely exists on Linux because it's open source. That's it.

Jinnaix:
Update your reading skills please, not a single word from ovahlord tells that mariadb is ONLY AVAILABLE ON LINUX.

Jinnaix:
Use another translator then, since that's absolutely not the meaning of that sentence in that context.

Jinnaix:
Why do you have to be this stubborn all the time?
Where does it say that it only exists for Linux? The reason why Linux is using mariadb in favor over everything else is because mariadb is open source. So it compliments the general basic of Linux. That doesn't mean it's good software.
If you care so much about mariadb, maintain its compability for TrinityCore then for the upcoming years.

Jinnaix:
convenience is a valid and fair reason to come to a decision in an open source project.
convenience is also the only valid argument you have against using MySQL since it's not pleasant for you to install MySQL on Linux.
Truth is  - manpower is a limiting factor. So the maintainers decide what dependencies and external software get used and supported.

TrinityCore can't satisfy every user and his personal preferences. 
TrinityCore supports and requires a certain software state. 
That does not mean that it does not or will not work with other software. But as in any open source project, it's up to the potential user to take a step away from the official supported way and try and test for themselves.

I personally don't give a duck about Linux at all. And a very big portion of the people who uses Linux for TrinityCore don't do it because of "performance" but because of money. Since using windows comes with additional license cost when you rent a server.

jackpoz:
> Just because a few shares are bought doesn't mean that MariaDB won't be developed further.

The whole reason MariaDB project started was because MySQL was bought by Oracle: https://web.archive.org/web/20130331001143/https://www2.computerworld.com.au/article/457551/dead_database_walking_mysql_creator_why_future_belongs_mariadb/
>The day the Sun purchase was announced, Widenius responded in the tried and true open source fashion — he forked MySQL, launching MariaDB, and took a swathe of MySQL developers with him.
"Many of the original MySQL core developers, including me, didn't believe that Oracle would be a good owner of MySQL and we wanted to ensure that the MySQL code base would be free forever," Widenius explains.

It's also not "a few shares", it's a takeover. It's in the first sentence of the article I sent
>MariaDB shareholders have backed a private equity **takeover offer** for the troubled database vendor.

So I expect JesusDB project to start as a child of MariaDB as soon as the private equity company finalizes buying MariaDB. Then all linux distros will switch to JesusDB and we will have to support 3 databases.

Espionage724:
I used MariaDB for it being advertised as more-FOSS and being relatively easy to install on Fedora Linux around 2015-2016. If I recall right, MySQL is also in Fedora default repos nowadays so I guess it's not really an issue for me to switch.

I've heard of others reporting TC having issues with MariaDB and TC devs not recommending the use of it, but kept using it on Linux since it worked fine. On Windows I've always used MySQL, and [when I was doing some install notes recently](https://wiki.realmofespionage.xyz/windows:servers:games:trinitycore_3.3.5_localhost) I went with MySQL and figured why use a fork of it when I could use the real thing for free :p

I use Fedora Linux mainly, but I wonder how officially-available MySQL is on Ubuntu and openSUSE? If it's in default repos, great! If MySQL has an official repo to add, that's cool too.



> Then I would suggest to consider something like Postgres instead unless you want to make it super painful for Linux folks out there. Highly doubt MySQL will return to Linux distros.

Forcing PostgreSQL would likely have me not bother with TC any more (I'd rather flat file or SQLite at that point). My entire stack is MySQL, and the only time I ever messed with PostgreSQL was setting up a Mastodon server, also far outside my usual stack with it involving Ruby.

Or, that'll give me another reason not to bother with a Linux server and just use MySQL on Windows :p

Espionage724:
> If I recall right, MySQL is also in Fedora default repos nowadays so I guess it's not really an issue for me to switch.

Nah that's a no-go; trying to casually install mysql-server on F40 KDE results in significant package conflicts because of the default MariaDB install. [Even Arch defaults to it](https://wiki.archlinux.org/title/MariaDB).

I'm now against dropping MariaDB compatibility. I suppose it doesn't matter on Windows, but mainstream Linux distros expect MariaDB.

Ovahlord:
Distros don't expect anything from you. It is your active choice to refuse to install software requirements.

jackpoz:
But then again, don't mainstream linux users expect to run everything in docker where the default packages don't matter at all ?

Espionage724:
> But then again, don't mainstream linux users expect to run everything in docker where the default packages don't matter at all ?

Maybe, but that's gross. I'd sooner shoe-horn MySQL on F40 somehow before using Docker :p

jackpoz:
Oh btw, https://investors.mariadb.com/news/news-details/2024/Offer-Update-eb4f65123/default.aspx
>Bidco now owns 61,263,283 MariaDB Shares, representing 88.70% of the issued share capital of MariaDB as of July 22, 2024

It's BidcoDB now 😄

Espionage724:
> > If I recall right, MySQL is also in Fedora default repos nowadays so I guess it's not really an issue for me to switch.
> 
> Nah that's a no-go; trying to casually install mysql-server on F40 KDE results in significant package conflicts because of the default MariaDB install. [Even Arch defaults to it](https://wiki.archlinux.org/title/MariaDB).
> 
> I'm now against dropping MariaDB compatibility. I suppose it doesn't matter on Windows, but mainstream Linux distros expect MariaDB.

MariaDB was causing a compiling issue for me on FreeBSD vs MySQL. I don't know how or why Fedora ties MariaDB to desktop components, but that's their ordeal, and FreeBSD proves they can be packaged separately along with TC only working with MySQL.

I'm for axing MariaDB support on the grounds of I don't seemingly need it (MySQL exists/works), and saw it cause an issue with `3.3.5` master.

Ovahlord:
superceeded by

https://github.com/TrinityCore/TrinityCore/pull/30255

LinuxJedi:
I'm not going to attempt to change your mind on this. But to clear up a few things:

* MariaDB Server is maintained by us at the MariaDB Foundation, not the for-profit entity in the links provided. That entity does pay a majority of the developers, but MariaDB Server development will continue on regardless.
* Fedora's dependency on MariaDB is due to KDE's Akonadi backend.
* If you wanted your software to run in our CI, there is a programme we are setting up for that (we already do it for some projects). We try to work with as many open source projects that use MariaDB Server as we can.
* We do work with FreeBSD, you didn't indicate what your compile issue was, so it is difficult to help you there.

I hope this helps.

-- 
Andrew Hutchings
Chief Contributions Officer
MariaDB Foundation

Ovahlord:
> I'm not going to attempt to change your mind on this. But to clear up a few things:
> 
>     * MariaDB Server is maintained by us at the MariaDB Foundation, not the for-profit entity in the links provided. That entity does pay a majority of the developers, but MariaDB Server development will continue on regardless.
> 
>     * Fedora's dependency on MariaDB is due to KDE's Akonadi backend.
> 
>     * If you wanted your software to run in our CI, there is a programme we are setting up for that (we already do it for some projects). We try to work with as many open source projects that use MariaDB Server as we can.
> 
>     * We do work with FreeBSD, you didn't indicate what your compile issue was, so it is difficult to help you there.
> 
> 
> I hope this helps.
> 
> -- Andrew Hutchings Chief Contributions Officer MariaDB Foundation

Thanks for the reply. The primary reason why we actually want to part ways with MariaDB are the increasing amount of incompatabilities or differences with MySQL. We already have several prepared statements which behave differenty and would require extensive refactors to support MariaDB which for us, given that we have three branches with very limited manpower to maintain everything, is highly inconvenient. While we are currently keeping MariaDB on life-support so that Linux folks can at least try to build and meddle with it, ideally we would ask to switch to MySQL. We are not opposed to keep MariaDB around if future versions would be aiming towards reducing differences between MySQL lts versions and Maria so that all involved parties have a happy life :)

LinuxJedi:
> Thanks for the reply. The primary reason why we actually want to part ways with MariaDB are the increasing amount of incompatabilities or differences with MySQL. We already have several prepared statements which behave differenty and would require extensive refactors to support MariaDB which for us, given that we have three branches with very limited manpower to maintain everything, is highly inconvenient. While we are currently keeping MariaDB on life-support so that Linux folks can at least try to build and meddle with it, ideally we would ask to switch to MySQL. We are not opposed to keep MariaDB around if future versions would be aiming towards reducing differences between MySQL lts versions and Maria so that all involved parties have a happy life :)

Is it possible to get details or links to these? Feel free to email me so as not too fill this ticket. The reason I ask is that we are working towards improving both MySQL and Oracle compatibility. It would be good to solve this for future users / other applications.

Shauren:
@Ovahlord we have solved issues related to mysql and mariadb returning different column types in query metadata already

@LinuxJedi it is mostly differences in C api such as these 
https://github.com/TrinityCore/TrinityCore/blob/fff12eafdf971e88ff9b3abe2c6e0a62e050b450/src/server/database/Database/MySQLHacks.h#L30-L32
https://github.com/TrinityCore/TrinityCore/blob/fff12eafdf971e88ff9b3abe2c6e0a62e050b450/src/server/database/Database/MySQLConnection.cpp#L136-L150
Or in new to mysql 8.4 mysql_stmt_bind_named_param replacing mysql_stmt_bind_param (old one is deprecated, unknown removal schedule)

jackpoz:
@LinuxJedi the main reason of dropping MariaDB compatibility is that we never really supported it and only accepted some PRs that fixed the build. The main concern is the growing list at https://mariadb.com/kb/en/mariadb-vs-mysql-compatibility/#incompatibilities-between-currently-maintained-mariadb-versions-and-mysql , where both products are diverging, so the drop-in compatibility assumption is no longer guaranteed.

We prefer to focus our efforts on 1 dbms and that as always been MySQL officially, so by inertia we are continuing with that.

Thanks for taking the time to write a comment.

LinuxJedi:
> @LinuxJedi the main reason of dropping MariaDB compatibility is that we never really supported it and only accepted some PRs that fixed the build. The main concern is the growing list at https://mariadb.com/kb/en/mariadb-vs-mysql-compatibility/#incompatibilities-between-currently-maintained-mariadb-versions-and-mysql , where both products are diverging, so the drop-in compatibility assumption is no longer guaranteed.

Ah! That makes sense, yes. We are increasingly spending time to discover things for this list, which is one reason it is growing. All this feedback helps. Many thanks!

CraftedRO:
> given that we have three branches with very limited manpower to maintain everything

"we" have limited manpower but 85 opened pull requests...

Espionage724:
> We do work with FreeBSD, you didn't indicate what your compile issue was, so it is difficult to help you there.

Thanks for the info and MariaDB in-general! My issue was with fmt: https://github.com/TrinityCore/TrinityCore/issues/30218

LinuxJedi:
> > We do work with FreeBSD, you didn't indicate what your compile issue was, so it is difficult to help you there.
> 
> Thanks for the info and MariaDB in-general! My issue was with fmt: #30218

No problem. OK, that kind of conflict makes sense. Thanks for letting us know.

I've made sure the Foundation team has seen this thread and we are learning from it.

Shauren:
> FYI, regarding fmt, we received this PR which should make things a little better there, I'm working towards getting it in: [MariaDB/server#3527](https://github.com/MariaDB/server/pull/3527)

I appreciate you notifying us but that libfmt issue has nothing to do with that MariaDB change, it is simply a matter of FreeBSD shipping a libfmt version that is not compatible with our project (breaking changes between v9 to v10 - we have v9 vendored, FreeBSD ships v10)

Espionage724:
On the new proposal:

> MySQL 5.7 and MariaDB are no longer accepted to compile or run the applications

Does that explicitly block MariaDB if it's present no matter what? I see mentions of `MIN_MYSQL_SERVER_VERSION_STRING` but I wonder that if MariaDB "presented" those MySQL-specific strings and a higher minimal version than required, would that allow it to work?

Basically if I wanted to use MariaDB and (assuming) all its code was compliant-enough to compile and work fine with TC, is TC explicitly looking to block that? I don't like the idea of blocking MariaDB on [a concept](https://github.com/TrinityCore/TrinityCore/pull/29960#issuecomment-2369289117) when it might continue to work fine code-wise.

Espionage724:
> FYI, regarding fmt, we received this PR which should make things a little better there, I'm working towards getting it in: [MariaDB/server#3527](https://github.com/MariaDB/server/pull/3527)

That issue doesn't seem to happen with FreeBSD 14.1 any more!

`mariadb114-client` and `mariadb114-server` can be installed; both don't seem to pull-in `libfmt` but it looks like something I did prior already pulled it in (it was already installed). TC 3.3.5 compiled fine as of a minute ago on FreeBSD 14.1 with `mariadb114` (11.4.2) and `libfmt` no longer seems an issue!

Espionage724:
> @LinuxJedi the main reason of dropping MariaDB compatibility is that we never really supported it and only accepted some PRs that fixed the build. The main concern is the growing list at https://mariadb.com/kb/en/mariadb-vs-mysql-compatibility/#incompatibilities-between-currently-maintained-mariadb-versions-and-mysql , where both products are diverging, so the drop-in compatibility assumption is no longer guaranteed.
> 
> We prefer to focus our efforts on 1 dbms and that as always been MySQL officially, so by inertia we are continuing with that.
> 
> Thanks for taking the time to write a comment.

Does MySQL have to be prioritized for TrinityCore?

TC is a cross-OS open-source project, (imo) appealing to the hobbyist concept of running a video game server outside of a "large entity". It sounds better to focus on the more-FOSS community-oriented option of MariaDB for compatibility, instead of proprietary business-centric MySQL?

Or rather, is MySQL more-supported code-wise to what TC's doing vs MariaDB? If so, can that be flipped? Since this is looks like a MySQL **or** MariaDB ordeal, why can't TC go all-in on MariaDB instead of MySQL?

I ran into minor issues with both MySQL and MariaDB over the years with TC 3.3.5 on different operating systems. Generally speaking, MySQL was mostly config related post-install, where MariaDB was mostly during compiling with TC's decisions with the build process. Most recently, [I ran into an update sql issue with MySQL ](https://github.com/TrinityCore/TrinityCore/issues/30287) with permissions I've been setting up fine for years, but now not good-enough all of a sudden. Had that same thing happen on Windows yesterday with MySQL, dropped-in MariaDB on Windows, and the same permissions/db set-up I used for years was back to no longer being a problem.

And this is silly; [I had this link bookmarked](https://dev.mysql.com/downloads/installer/) for MySQL downloads on Windows. Link looks good right? [The root is this](https://dev.mysql.com/downloads/), with MySQL installer for Windows at the bottom. Turns out, that's outdated. [This](https://www.mysql.com/downloads/) doesn't look anywhere near as quick-readable to find a Windows installer. At the very bottom in the footer there's a **slightly**-different [MySQL Community Server link ](https://dev.mysql.com/downloads/mysql/)that then shows MySQL 8.4 for Windows. And the only reason I caught that was because I knew MySQL 8.4 existed and I thought it was odd that first Windows page was only showing 8.0. MariaDB 11.5 server for Windows was a single button click from Google search first result.

CraftedRO:
@Espionage724 MySQL is clearly best RDBMS, so yea that's why it shoud be prioritized, and even if I like and actually even use mariadb at the moment too on my server in production that doesn't mean I won't change to mysql if I have to since they are clearly not 100% compatible eachother anymore, but if you find hard to find mysql download page here's the link you are looking for https://dev.mysql.com/downloads/mysql/ .

Espionage724:
> MySQL is clearly best RDBMS

PostgreSQL and others would probably take issue with that :p

What makes MySQL better for TC vs MariaDB?

CraftedRO:
> What makes MySQL better for TC vs MariaDB? Not asking in-general; I mean specifically with TC.

if it's in general be sure it's better for TC specifically aswell.

jackpoz:
@Smerdokryl found @Espionage724 as another volunteer to maintain MariaDB support. You guys should work together (and maybe see if there are more people interested in helping)

PR #29963, opened 2024-05-07
Scripts/HoL: Fix a crash caused by iterator invalidating in Ionar boss script

r4d1sh:
**Changes proposed:**

The crash occurs in `boss_ionar::CallBackSparks`. We are calling `Creature::DespawnOrUnsummon` for iterable creatures directly within the loop, which leads to call `boss_ionar::SummonedCreatureDespawn`, followed by `SummonList::Despawn`. The iterator becomes invalidated, resulting in a core crash.

Steps to reproduce the crash:

1. `.tele hallsoflightning`
2. Enter the dungeon and pull the boss Ionar.
3. Target Ionar and use `.damage 200000`
4. When sparks appear, kill at least one of them.
5. After a few seconds, the server crashes.

This PR fixes the issue.

**Tests performed:**

Builded and tested in-game.

### DIFF
--- a/src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
+++ b/src/server/scripts/Northrend/Ulduar/HallsOfLightning/boss_ionar.cpp
@@ -145,9 +145,12 @@ struct boss_ionar : public BossAI
 
         Position pos = me->GetPosition();
 
-        for (ObjectGuid guid : summons)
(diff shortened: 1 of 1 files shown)


Shauren:
Merging crash fix anyway but aren't sparks supposed to be immune?

Ovahlord:
Nope, you can attack and harm them. But usually they have that much HP that killing them should be impossible (if they can even die)

Shauren:
Actually...
obraz

PR #29966, opened 2024-05-08
Core/Spells: Fix dispelling movement impairing auras by PvP-trinket and similar spells.

r4d1sh:
**Changes proposed:**

After [3306a4d](https://github.com/TrinityCore/TrinityCore/commit/3306a4d06cc557967c23adcc60c2e) spells that grant immunity to mechanics do not completely remove the auras that these mechanics have on only one of their effects. Instead, these effects are only temporarily disabled.

However, there seem to be exceptions to this. For example, PvP-trinkets, or Human racial "Every Man for Himself". 
Such spells are distinguished by the duration of their aura - only 100 ms. Of course, they should completely remove the auras they grant immunity to, and not just disable some effects for 100 milliseconds.

How to reproduce:

1. Target yourself and use `.aura 31589`
2. Use PvP-trinket or "Every Man for Himself"
3. See that nothing worked as expected

This PR fixed the issue.

**Tests performed:**

Builded and tested in-game.


r4d1sh:
Spells with Duration ID 36 (1 second) have also been added to the exception list. Primarily, these are Blink and its various variations.

Shauren:
You are wrong about 29300 - I tested this on retail (both that spell and blink are unchanged) and it behaves like on TC, grants immunity only for blink mini duration and does not remove the aura

r4d1sh:
> You are wrong about 29300 - I tested this on retail (both that spell and blink are unchanged) and it behaves like on TC, grants immunity only for blink mini duration and does not remove the aura (because mechanic is on one effect instead of entire spell)

Wow, that's an amazing fact. Well, then you can merge only the first commit.

Seyden:
You sure that your fix isn't just a hack and that you should actually use SPELL_ATTR1_IMMUNITY_PURGES_EFFECT                               ?

Shauren:
Did some more testing with 29300
Using pvp trinket on it removes the stun effect only, not whole aura (threat modifier remains)

r4d1sh:
> Did some more testing with 29300
> Using pvp trinket on it removes the stun effect only, not whole aura (threat modifier remains)

Stun effect removed completely or only for 100 ms?

And do you mean wotlk classic?

Shauren:
Completely removed

r4d1sh:
Tested on WotLK Classic (Cataclysm pre-patch): Human racial "Every Man for Himself" completely removes 29300, not only stun effect.

Shauren:
Alright then, just move setting the variable to spell id cases for pvp trinket and human racial that are already there instead of doing it by duration

r4d1sh:
What about spells like:

ID - 53490 Bullheaded (duration 407, 100 ms) - pretty sure that it should completely remove the auras they grant immunity to
ID - 48020 Demonic Circle: Teleport (duration 36, 1 sec) - pretty sure too, need to check on cata classic with 29300
Blink (duration 36, 1 sec) - pretty sure too, need to check on cata classic with 29300

Shauren:
Demonic circle only removes slows, not stuns or roots, you wont test it with 29300

Shauren:
Try your luck with feral druid talent Infected Wounds

r4d1sh:
> Try your luck with feral druid talent Infected Wounds (it's the reason why current code behaves the way it does)

"Every Man for Himself" completely dispelled "Infected Wounds" on 4.4.0.54670

PR #29977, opened 2024-05-12
Scripts/RagefireChasm: Implement Dark Shaman Koranthal encounter

aquadeus:
**Changes proposed:**

-  Implement Dark Shaman Koranthal encounter

**Tests performed:**

Tested in-game


src/server/scripts/Kalimdor/RagefireChasm/boss_dark_shaman_koranthal.cpp
```diff
@@ -0,0 +1,132 @@
...
+    {
+        if (!UpdateVictim())
+            return;
+
+        events.Update(diff);
+
+        if (me->HasUnitState(UNIT_STATE_CASTING))
+            return;
+
+        while (uint32 eventId = events.ExecuteEvent())
```
mdX7, line 89: i'd prefer to use `if` here instead of `while` to prevent skipping of events as in:
- Imagine EVENT_TWISTED_ELEMENTS and EVENT_SHADOW_STORM should be triggered at the same time (due to randomness overlaps may happen)
- if we have a `while` we'd execute the first event properly during first iteration, e.g. EVENT_TWISTED_ELEMENTS and reschedule it as usual
- in the next iteration of the `while` loop we'd execute `EVENT_SHADOW_STORM`, the boss would do the talk and reschedule it. But would NOT cast the spell, as hes already casting the spell from previous iteration `SPELL_TWISTED_ELEMENTS`

By using `if` here it would first cast `SPELL_TWISTED_ELEMENTS` as usual, then wait until the cast is done (see above `UNIT_STATE_CASTING` check) and on the very next tick it would handle `EVENT_SHADOW_STORM`

mdX7 (changes requested):

src/server/scripts/Kalimdor/RagefireChasm/boss_dark_shaman_koranthal.cpp
Shauren: Thats why all scripts now repeat UNIT_STATE_CASTING check at the end of each iteration

PR #29980, opened 2024-05-12
Core/Criteria: Implemented CriteriaType CompletedLFGDungeon

BertiRean:
**Changes proposed:**

-  
-  
-  

**Issues addressed:**

**Tests performed:**

- [ ] 
- [ ]


src/server/game/Achievements/CriteriaHandler.cpp
```diff
@@ -1247,6 +1247,7 @@ bool CriteriaHandler::IsCompletedCriteria(Criteria const* criteria, uint64 requi
         case CriteriaType::Login:
             return true;
         // handle all statistic-only criteria here
+        case CriteriaType::CompletedLFGDungeon:
```
Shauren, line 1250: This is wrong, CompletedLFGDungeon is not statistic only, only 13422 is statistic, 46666, 60921, 61256 and 61357 are not

Shauren (changes requested):

src/server/game/Achievements/CriteriaHandler.cpp
BertiRean: Are these values CriteriaIds or CriteriaTree Id ? I don't found 61256

Shauren: criteria ids

PR #29989, opened 2024-05-14
Scripts/RagefireChasm: Implement Slagmaw encounter

aquadeus:
Thanks to @mdX7 and @meji46 for the help

**Changes proposed:**

-  Implement Slagmaw encounter

**Tests performed:**

Tested in-game


src/server/scripts/Kalimdor/RagefireChasm/boss_slagmaw.cpp
```diff
@@ -0,0 +1,160 @@
...
+                        DoCast(target, SPELL_LAVA_SPIT);
+                        _lavaSpitCounter++;
+                    }
+                    events.Repeat(1s);
+                    break;
+                }
+                else if (_lavaSpitCounter == 5)
+                {
+                    PhaseSubmerge();
+                    events.CancelEvent(EVENT_LAVA_SPIT);
```
mdX7, line 122: no need to cancel it here, as its not re-scheduled in this `else if` branch, its only rescheduled in `if` above

src/server/scripts/Kalimdor/RagefireChasm/boss_slagmaw.cpp
```diff
@@ -0,0 +1,176 @@
...
+};
+
+// 61463 - Slagmaw
+struct boss_slagmaw : public BossAI
+{
+    boss_slagmaw(Creature* creature) : BossAI(creature, BOSS_SLAGMAW), _lavaSpitCounter(0), _lastTeleportSpell(SPELL_MAGNAW_TELEPORT_WEST) { }
+
+    void JustAppeared() override
+    {
+        me->SetCanMelee(false);
```
Shauren, line 57: This should be a static flag in db

Shauren (changes requested):

src/server/scripts/Kalimdor/RagefireChasm/boss_slagmaw.cpp
```diff
@@ -0,0 +1,165 @@
...
+        _lavaSpitCounter = 0;
+
+        events.ScheduleEvent(EVENT_TELEPORT, 3s);
+    }
+
+    uint32 GetNextTeleportSpell()
+    {
+        std::array<uint32, 4> teleportSpells = { };
+        auto end = std::ranges::remove_copy(SlagmawTeleportSpells, teleportSpells.begin(), _lastTeleportSpell).out;
+        _lastTeleportSpell = Trinity::Containers::SelectRandomContainerElement(std::span(teleportSpells.begin(), end));
```
Shauren, line 98:
```suggestion
        std::array<uint32, 3> teleportSpells = { };
        std::ranges::remove_copy(SlagmawTeleportSpells, teleportSpells.begin(), _lastTeleportSpell);
        _lastTeleportSpell = Trinity::Containers::SelectRandomContainerElement(teleportSpells);
```

_lastTeleportSpell is always initialized to a value that exists in SlagmawTeleportSpells so this kind of shortcut (removing temp span) will work (it was initially what i suggested on discord but removed it because i didnt know about that _lastTeleportSpell constraint)

PR #29992, opened 2024-05-15
Core/Creatures: Check fallback difficulty data for creature static flags override

meji46:
**Changes proposed:**

-  Check fallback difficulty data for creature static flags override. We can override the static flags of a spawn with several difficulties without inserting the same data for each difficulty

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game.

None

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -10722,7 +10722,16 @@ JumpChargeParams const* ObjectMgr::GetJumpChargeParams(int32 id) const
 
 CreatureStaticFlagsOverride const* ObjectMgr::GetCreatureStaticFlagsOverride(ObjectGuid::LowType spawnId, Difficulty difficultyId) const
 {
-    return Trinity::Containers::MapGetValuePtr(_creatureStaticFlagsOverrideStore, std::make_pair(spawnId, difficultyId));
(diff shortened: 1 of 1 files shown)


mdX7 (approved):

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -10889,10 +10898,14 @@ void ObjectMgr::LoadCreatureStaticFlagsOverride()
             continue;
         }
 
-        if (std::find(creatureData->spawnDifficulties.begin(), creatureData->spawnDifficulties.end(), difficultyId) == creatureData->spawnDifficulties.end())
+        // DIFFICULTY_NONE is always a valid fallback
+        if (difficultyId != DIFFICULTY_NONE)
         {
-            TC_LOG_ERROR("sql.sql", "Table `creature_static_flags_override` has data for a creature that is not available for the specified DifficultyId (SpawnId: {}, DifficultyId: {}), skipped", spawnId, difficultyId);
-            continue;
+            if (std::find(creatureData->spawnDifficulties.begin(), creatureData->spawnDifficulties.end(), difficultyId) == creatureData->spawnDifficulties.end())
```
BertiRean, line 10904: You could use std::ranges instead

meji46: That's from C++23 so we can't use it yet

BertiRean: You're right, just set, map and other containers have .contains method.

PR #29995, opened 2024-05-20
Core/Players: moved zone and area updating from a standalone timer into Heartbeat handling

Ovahlord:
**Changes proposed:**

-  turns out that retail and classic update their zone data on a per-heartbeat basis as well so we move the zone and area id handling over there

**Tests performed:**
- tested ingame

Is is worth noting that because of this relocation of the updating, zones, areas and consequently phasing and spell_area may take a bit longer as the old zone update timer was ticking every second while the heartbeat only fires once every 5200ms or on certain movement actions.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -140,8 +140,6 @@
 #include <G3D/g3dmath.h>
 #include <sstream>
 
-#define ZONE_UPDATE_INTERVAL (1*IN_MILLISECONDS)
-
 // corpse reclaim times
 #define DEATH_EXPIRE_STEP (5*MINUTE)
 #define MAX_DEATH_COUNT 3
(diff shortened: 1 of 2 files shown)


Shauren:
Actually pointless code tbh (the whole periodic update)
* area/zone is updated on every relocation (Unit::UpdatePosition)
* because areatrigger packet now contains entered/exited bit (unlike 335), resting is handled entirely in WorldSession::HandleAreaTriggerOpcode

Takenbacon:
>     * area/zone is updated on every relocation (Unit::UpdatePosition)

It shouldn't be, the vmap calls to get the area/zone ID from a position isn't cheap

>     * because areatrigger packet now contains entered/exited bit (unlike 335), resting is handled entirely in WorldSession::HandleAreaTriggerOpcode

Never trust the client.

PR #29996, opened 2024-05-20
Scripts/Battlegrounds: Implement Seething shore

Golrag:
**Changes proposed:**

-  Implement Seething Shore Battleground
- Buffs spawns are random right now, not sure how they handle this on retail.
- Spawning Azerite Nodes
- Achievements

**Issues addressed:**

Closes: None?

**Tests performed:**

- Builds locally
- Tested in game by playing through the battleground

**Known issues**
- The trusth spell doesn't work. See https://github.com/TrinityCore/TrinityCore/pull/28081
- Azerite nodes are missing spawning animation (summoned by spell effect)
- The tar pits do environmental lava damage. This doesn't happen on retail
- Not all texts of the commanders are implemented, not sure when which are called.
- Texts from commanders should only be visible by own faction. This should be handled by broadcast texts player condition. I don't think this is the case right now.

**TODO List**
- [x] Vendor items are missing
- [x] Teleporting to `world_safe_locs` that are using transport offsets doesn't work right now
- [ ] Cosmetic event: cannons start firing at some point, randomly. TBD if this is in scope
- [x] Some more waypoints might be missing


src/server/scripts/Battlegrounds/SeethingShore/battleground_seething_shore.cpp
```diff
@@ -0,0 +1,820 @@
...
+        {
+            _isInformedNearVictory[teamId] = true;
+            if (teamId == TEAM_ALLIANCE)
+                battleground->SendBroadcastText(SeethingShore::Texts::AllianceNearVictory, CHAT_MSG_BG_SYSTEM_NEUTRAL);
+            else
+                battleground->SendBroadcastText(SeethingShore::Texts::HordeNearVictory, CHAT_MSG_BG_SYSTEM_NEUTRAL);
+        }
+
+        battleground->UpdatePvpStat(player, SeethingShore::PvpStats::CapturedAzerite, 1);
+        capturePoint->SetAnimKitId(2560, false);
```
Nyr97, line 263: I know it's hard, but any idea what the AnimKit does? Just to name it rather than using plain numbers.

Golrag: Animation that makes it "destroy" / "being mined", not really sure what naming would fit better

src/server/scripts/Battlegrounds/SeethingShore/battleground_seething_shore.cpp
```diff
@@ -0,0 +1,902 @@
...
+{
+    explicit transport_seething_shore() : TransportScript("transport_seething_shore") { }
+
+    void OnRemovePassenger(Transport* /*transport*/, Player* player) override
+    {
+        if (Battleground const* bg = player->GetBattleground())
+        {
+            if (bg->GetStatus() != STATUS_IN_PROGRESS)
+            {
+                WorldSafeLocsEntry const* pos = bg->GetTeamStartPosition(Battleground::GetTeamIndexByTeamId(player->GetBGTeam()));
```
mdX7, line 687: can we assume that pos is never null?

src/server/scripts/Battlegrounds/SeethingShore/battleground_seething_shore.cpp
```diff
@@ -0,0 +1,902 @@
...
+        if (pathId != npc_bg_seething_shore_air_supply_ground_dummy::PATH_ID_GROUND)
+            return;
+
+        Creature* groundDummy = me->FindNearestCreature(SeethingShore::Creatures::AirSupplyGroundDummy, 20.0f);
+        if (!groundDummy)
+            return;
+
+        QuaternionData const rot = QuaternionData::fromEulerAnglesZYX(groundDummy->GetOrientation(), 0.f, 0.f);
+        uint8 const action = urand(0, 3);
+        switch (action)
```
mdX7, line 789: you can use `if (roll_chance_i(25))` for case 3 and handle 0-2 in else branch

src/server/scripts/Battlegrounds/SeethingShore/battleground_seething_shore.cpp
Golrag: People can't enter the bg without a valid one, that said with `.reload battleground` it could be possible to fill in bad data in the database during a running battleground and cause some issues

PR #29997, opened 2024-05-21
Core/Disables: Added DISABLE_TYPE_PHASE_AREA

mdX7:
**Tests performed:**
Used the following sql and checked if any of the Refugees are visible in Stormwind harbor while gm off
```sql
DELETE FROM `disables` WHERE `sourceType`=9 AND `entry`=11595;
INSERT INTO `disables` (`sourceType`, `entry`, `flags`, `params_0`, `params_1`, `comment`) VALUES 
(9, 11595, 0, '', '', 'Disable Darnassian and Gilnean Refugees phase');
```

**Note:**
Also thought about skipping loading of objects with disabled phases entirely, but since this table is reloadable and creature/gameobject isn't i decided to just not apply the phase to the players

### DIFF
--- a/src/server/game/Conditions/DisableMgr.cpp
+++ b/src/server/game/Conditions/DisableMgr.cpp
@@ -241,7 +241,7 @@ void LoadDisables()
                     TC_LOG_INFO("misc", "Pathfinding disabled for {} map {}.", MapTypeNames[mapEntry->InstanceType], entry);
                 break;
             }
-            case DISABLE_TYPE_PHASE:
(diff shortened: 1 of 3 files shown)


Shauren:
1. Why go this ways intead of just removing phase_area?
2. Name is too generic for what it does, it disables phase_area, not the phase (still works if gained from aura)

mdX7:
> 1. Why go this ways intead of just removing phase_area?

so its easier to re-enable them without looking years of commit history back to find the related phase_area lines

Shauren (approved):

PR #30012, opened 2024-05-28
Core/Misc: Custom exceptions will use string_view for constructor and noexcept for what

BertiRean:
**Changes proposed:**

-  Change constructor param of custom exceptions to std::string_view instead std::string or const char*
-  
-  

**Issues addressed:**

**Tests performed:**
Build Pass

- [ ] 
- [ ]


src/common/DataStores/DB2FileLoader.h
```diff
@@ -178,7 +178,7 @@ struct DB2RecordCopy
 class TC_COMMON_API DB2FileLoadException : public std::exception
 {
 public:
-    DB2FileLoadException(std::string msg) : _msg(std::move(msg)) { }
+    DB2FileLoadException(std::string_view msg) : _msg(msg) { }
```
Shauren, line 181: This is a bad change. Don't just change things for the sake of changing and check WHY they are written the way they are.

Almost all of DB2FileLoadException are constructed using Trinity::StringFormat which means that old code directly moves the result of formatting into the exception object (1 instance of the string is ever created with content) while your "improved" version causes unneccessary copies

Shauren (changes requested):

src/common/DataStores/DB2FileLoader.h
BertiRean: Thanks by the comment. And for the other exceptions ?

BertiRean: I understand that format return a normal std::string, but when you call the constructor with that param is not a new copy created ?
inline std::string StringFormat(FormatString<Args...> fmt, Args&&... args)
It's not a temporal string
I mean not for the constructor
StringFormat -> Create std::string A, A is passed as param to Constructor of Exception, the constructor create a new str B, and this  B is moved to _str member in class

Shauren: Except that the argument for DB2FileLoadException is constructed using `string&&` constructor using return value of StringFormat (which moves, not copies)

PR #30024, opened 2024-06-04
Move zone message handling to map

Takenbacon:
**Changes proposed:**

-  This moves the zone specific message sending functions to map level, preventing needless iterating on the full world player list as well as improving thread safety as accessing player objects cross map is unsafe, even though it technically doesn't matter much in this specific context.

**Issues addressed:**

None

**Tests performed:**

Builds on MSVC, tested in game.

None

### DIFF
--- a/src/server/game/Weather/Weather.cpp
+++ b/src/server/game/Weather/Weather.cpp
@@ -29,7 +29,6 @@
 #include "ScriptMgr.h"
 #include "Util.h"
 #include "World.h"
-#include "Map.h"
 
 /// Create the Weather object
 Weather::Weather(Map* map, uint32 zoneId, WeatherData const* weatherChances)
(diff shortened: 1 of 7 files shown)


src/server/game/Weather/Weather.cpp
```diff
@@ -28,10 +28,11 @@
 #include "ScriptMgr.h"
 #include "Util.h"
 #include "World.h"
```
Shauren, line 30: Is this still needed? (also put Map.h in its sorted place)

Takenbacon: > Is this still needed? (also put Map.h in its sorted place)

Doesn't look like it's needed, no. I need to fix the build on GCC anyways apparently, forgot how touchy it can be.

Takenbacon: There is still one use, setting the timer from the world config.

Shauren (approved):

PR #30037, opened 2024-06-14
Core/PacketIO: increase ByteBuffer append limit from ~10mb to ~100mb

Ovahlord:
**Changes proposed:**

-  just increasing the bytebuffer::append limit so we can squeeze larger datasets into it.

**Tests performed:**
- tested for Cata Classic as their hotfixes require that increased limit.


Lordron:
this assert is incorrect btw. it checks the current size, not the size after (size() + cnt). Nothing stops me from appending whatever bytes (above the limitation) in a single call. Only the second call will trigger the ASSERT

PR #30044, opened 2024-06-17
Core/Misc: Improve and optimize account instance enter times system.

r4d1sh:
**Changes proposed:**

-  **Cleanup account instance times data from DB on account deletion.** Currently, if we delete an account (for example, using `.account delete` command), the records associated with that account will remain in `account_instance_times` table forever. The first commit fixes this.
-  **Update instance enter times only when it's really needed.** Currently, in every call of `Player::Update`, we check the instance reset times and delete the expired ones. Actually, we can do this only when the player attempts to enter an instance. It would also be good to do this immediately after loading this data from DB, to avoid having too old records in the database if player has not been going to dungeons for a long time. The second commit fixes this.

**Tests performed:**

Builded and tested in-game.

### DIFF
--- a/src/server/game/Accounts/AccountMgr.cpp
+++ b/src/server/game/Accounts/AccountMgr.cpp
@@ -122,6 +122,10 @@ AccountOpResult AccountMgr::DeleteAccount(uint32 accountId)
     stmt->setUInt32(0, accountId);
     CharacterDatabase.Execute(stmt);
 
+    stmt = CharacterDatabase.GetPreparedStatement(CHAR_DEL_ACCOUNT_INSTANCE_LOCK_TIMES);
+    stmt->setUInt32(0, accountId);
+    CharacterDatabase.Execute(stmt);
+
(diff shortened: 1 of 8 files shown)


Nyr97:
What if you're inside of the dungeon whose ID expires with these changes? If you don't check on every ::Update, it won't kick the player from the instance, right?

r4d1sh:
> What if you're inside of the dungeon whose ID expires with these changes? If you don't check on every ::Update, it won't kick the player from the instance, right?

Yes, it won't kick the player from the instance, whether we check in each `Player::Update` or not. Because it shouldn't. This system does not imply any kicking players from instances; it's designed to prevent players from entering too many dungeons within one hour.

r4d1sh:
Additionally, all this logic has been moved from `Player` to `WorldSession`, as it pertains to the account rather than the character. This means that there will no longer be a need to load data from the db every time a player changes characters without fully logging out of the account.

PR #30047, opened 2024-06-19
DB: Clean some deprecated warnings in MySQL 8+

CraftedRO:
**Changes proposed:**

-  Clean following deprecated warnings in MySQL 8+ :
```
/* Warning: (1681) Integer display width is deprecated and will be removed in a future release. */
/* Warning: (1681) UNSIGNED for decimal and floating point data types is deprecated and support for it will be removed in a future release. */
/* Warning: (1287) 'utf8mb3' is deprecated and will be removed in a future release. Please use utf8mb4 instead */
/* Warning: (3778) 'utf8mb3_general_ci' is a collation of the deprecated character set UTF8MB3. Please consider using UTF8MB4 with an appropriate collation instead. */
```

**Issues addressed:**

None

**Tests performed:**

Tests were performed on MySQL 8.4
```
Auth:
before
/* Affected rows: 1.535  Found rows: 0  Warnings: 8  Duration for 231 queries: 11,976 sec. */
after
/* Affected rows: 1.535  Found rows: 0  Warnings: 0  Duration for 231 queries: 11,942 sec. */

Characters:
before
/* Affected rows: 191  Found rows: 0  Warnings: 13  Duration for 842 queries: 00:01:25.6 (+ 0,015 sec. network) */
after
/* Affected rows: 191  Found rows: 0  Warnings: 0  Duration for 842 queries: 54,172 sec. */
```

- ~[ MySQL 8.0.16 is the first version that supports CHECK constraints so it won't work on lower versions. ]~
- [ Aditionally ALTER TABLE.sql can be created for worldb too idk.]


Shauren:
Not touching tinyint(1) - mysql has special treatment for it (its still preserved by mysqldump, other types (int, bigint) and values (2,3...) are dropped)

You also didn't do the hard part, which is figuring out how to make charset changes stick when running mysqldump again

Shauren:
There was no need to attempt to fix the build - it fails because CI checks if new sql file was added to the big ```INSERT INTO `updates` ``` query (and as a PR author you are not supposed to do that)

Issue #30054, opened 2024-06-24
Core/Character: .character changefaction shoud most likely not teleport character if it's DK


CraftedRO:
### Description

While using .character changefaction it will teleport the character to default faction zone (stormwind for alliance and orgrimmar inn for horde) but that shoud most likely not be the case for a fresh created DK that shoud have not leave that zone untill finnish questline either it will bugg out.

### Expected behaviour

Most likely DK class shoud be able to change faction but if it's still in class chain qust line shoud not leave acherus.

Probably there's need a condition to check dk phase and not teleport it from acherus if it's still there on chain quests.

### Steps to reproduce the problem

create a DK
.character changefaction dk
notice if you use it and change your faction, you will be teleported and you won't be able to complete dk class chain without gm assistance

### Branch

3.3.5

### TC rev. hash/commit

8824889f53a9d11d5b943f844900e41456edddec

### Operating system

Windows 11 x64

### Custom changes

None

PR #30057, opened 2024-06-25
Core: Add some blizzlike level restrictions while using changefaction

CraftedRO:
**Changes proposed:**

-  Adding minim required level 10 for players and 60 for Death Knights while using .character changefaction
-  https://www.bluetracker.gg/wow/topic/us-en/19717691770-helpful-faction-change-resources/

Captură de ecran 2024-06-25 035112

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/30054

**Tests performed:**

Does it build, tested in-game as can be seen below:

https://youtu.be/w0uOkyq7dgM

- [ Find a way to make this condition work while used directly in server console too, I guess... ]


Ovahlord:
the psysmessages should be converted into trinity strings so they can be localized.

Shauren:
fixed level checks for offline targets

CraftedRO:
what do you think, woud be blizzlike to also add a check for dk phase (to make sure it won't leave zone) or it's enough as it is ?

PR #30062, opened 2024-06-28
Core/Spells: Fixed possible use after free with deleted focusObject

Jildor:
**Changes proposed:**

-  Fixed possible use after free with deleted focusObject

**Issues addressed:**

None

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -576,7 +576,8 @@ m_caster((info->HasAttribute(SPELL_ATTR6_CAST_BY_CHARMER) && caster->GetCharmerO
     m_procAttacker = 0;
     m_procVictim = 0;
     m_hitMask = 0;
-    focusObjectGUID = ObjectGuid::Empty;
+    focusObject = nullptr;
+    m_focusObjectGUID.Clear();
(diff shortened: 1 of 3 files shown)


jackpoz:
Is this not handled in UpdatePointer() or whatever is called at every update loop ?

Jildor:
Updated and tested with feedback, thanks

src/server/game/Spells/Spell.cpp
```diff
@@ -7284,6 +7287,13 @@ bool Spell::UpdatePointers()
             m_originalCaster = nullptr;
     }
 
+    if (m_focusObjectGUID)
+    {
+        focusObject = ObjectAccessor::GetGameObject(*m_caster, m_focusObjectGUID);
+        if (focusObject && !focusObject->IsInWorld())
```
Shauren, line 7293: IsInWorld check is unneccessary, ObjectAccessor will never hand out pointers to objects not-in-world (other than players with FindConnectedPlayer)

Jildor: thanks, updated

jackpoz:
Thanks for the PR 🚀 😃

PR #30067, opened 2024-07-01
Core/Creature: Set PvP flags for Azuremyst NPCs

danielfinke:
**Changes proposed:**

- Allow targeting "Draenei Youngling" for quest "A Hearty Thanks!"
- Port changes from azerothcore/azerothcore-wotlk#4270

**Tests performed:**

- Tested both NPCs' behaviour in-game


CraftedRO:
In my tests there's no need of c++ part for Rescue the Survivors! quest, you can use  [Gift of the Naaru] on npc_draenei_survivor too, so only query is needed and this one is in hexa and mine it's decimal here https://github.com/TrinityCore/TrinityCore/issues/30479 both does the same thing.

Shauren (approved):

PR #30069, opened 2024-07-04
Scripts/Spells: Implemented Priest talent Shadow Word: Death

cristianvnt:
**Issues addressed:**
[Shadow Word: Death](https://www.wowhead.com/spell=32379/shadow-word-death) talent doesn't work properly.

0122470ed75488650344fc306beee234ec7bd886

**Tests performed:**
tested ingame
talent in general tree.
e.g for disc: BAQA/KVNHBYFUG61hTBXiMbcYCBoFEJJSSSQCpkkkESAAAAAAAAAAAQRJNOgkEkCRiICR5AEJRUIA


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -3079,6 +3080,45 @@ class spell_pri_shadow_mend_periodic_damage : public AuraScript
...
+            { SPELL_PRIEST_SHADOW_WORD_DEATH, EFFECT_2 },
+            { SPELL_PRIEST_SHADOW_WORD_DEATH, EFFECT_4 }
+        });
+    }
+
+    void HandleEffectHit(SpellEffIndex /*effIndex*/)
+    {
+        Unit* caster = GetCaster();
+        Unit* target = GetHitUnit();
+        if (!caster || !target)
```
mdX7, line 3100: both never null (unless hit entity can be gameobject, which isn't the case here)

Mematoru23: so no need for the check, right?

mdX7: correct. you can use the variables instead of `GetCaster()` or `GetHitUnit()` in code below, but you dont need the null checks

src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -3079,6 +3080,45 @@ class spell_pri_shadow_mend_periodic_damage : public AuraScript
...
+    }
+
+    void HandleEffectHit(SpellEffIndex /*effIndex*/)
+    {
+        Unit* caster = GetCaster();
+        Unit* target = GetHitUnit();
+
+        uint64 hitDamage = GetHitDamage();
+
+        if (target->GetHealthPct() < GetEffectInfo(EFFECT_1).BasePoints)
```
Nyr97, line 3103: never BasePoints, CalcValue(caster) in case there's anything that might be modifying this effect

Nyr97: there might be an aura of whatever sort that increases the threshold, which has been fairly common in the past

PR #30073, opened 2024-07-06
Core/Players: Fixed exploration

Ghaster:
**Changes proposed:**

-  Flipped check as we naturally should not explore any zone with an offset larger than DataFlags.

**Issues addressed:**

**Tests performed:**

- [ ] 
- [ ]


Shauren:
No, this is wrong. DataFlags is a dynamically growable field so DataFlags will increase its size to whatever is neccessary for current zone

PR #30083, opened 2024-07-14
Spells/Auras: Implement SPELL_AURA_ACT_AS_CONTROL_ZONE

Golrag:
**Changes proposed:**

-  Implement SPELL_AURA_ACT_AS_CONTROL_ZONE (401)

**Issues addressed:**

- None. But this is required for Silvershard mines

**Tests performed:**

- Builds locally
- Tested if the events were still called in thte outdoor pvp zones that contains control zones
- Tested applying the aura with some modifications made to gameobject_template

- None


Shauren:
But that makes them visible to client - GAMEOBJECT_TYPE_CONTROL_ZONE does not have a data field to make them serverside

Golrag:
Yes, but the gameobjects that are present in the 3 spells that use this aura dont have a display id, so they are never sent to client in update_object

The war within adds 2 more spells like this, but we're kinda limited on data.

Could add a check to verify display id is 0 before spawn

mdX7 (approved):

PR #30084, opened 2024-07-15
Core/GroupHander: Check if player is in a group for `CMSG_REQUEST_PARTY_MEMBER_STATS` before providing data.

### LINKED ISSUE #30085: Player tracking exploit with CMSG_REQUEST_PARTY_MEMBER_STATS
### Description

CMSG_REQUEST_PARTY_MEMBER_STATS does not check for groups.
Therefore, a malicious client can just send any players GUID and receive health, rough position, power etc back.

With a modified client this basically renders invisiblity to nothing, as you can just display the rough position after any player went invisible with this exploit.

### Expected behaviour

CMSG_REQUEST_PARTY_MEMBER_STATS should check current players and target players group, if they are not in the same group, mark it as a cheat attempt.

1.) Send a "malicious" CMSG_REQUEST_PARTY_MEMBER_STATS packet to the server.
2.) Parse and read the response SMSG_PARTY_MEMBER_STATS_FULL.

This requires of course third party tools or a modified client.

### Branch

3.3.5

### TC rev. hash/commit

TrinityCore rev. 8824889f53a9 2024-06-22 14:03:17 +0200 (3.3.5 branch) (Win64, Release, Static)

### Operating system

Windows 10 x64

### Custom changes

None

heyitsbench:
**Changes proposed:**
- Adds a group check to the handling of `CMSG_REQUEST_PARTY_MEMBER_STATS` to ensure players can't request information on other players not in their group.

**Issues addressed:**

Closes #30085

**Tests performed:**

Untested, have no way of spoofing packets like that that I know of.

- This might be the wrong approach, this might be incorrect, but I think it should probably be corrected, so I figured I'd throw my hat in the ring.

### DIFF
--- a/src/server/game/Handlers/GroupHandler.cpp
+++ b/src/server/game/Handlers/GroupHandler.cpp
@@ -923,7 +923,7 @@ void WorldSession::HandleRequestPartyMemberStatsOpcode(WorldPacket &recvData)
     recvData >> Guid;
 
     Player* player = ObjectAccessor::FindConnectedPlayer(Guid);
-    if (!player)
+    if (!player || !(player->GetGroup() != GetPlayer()->GetGroup()))
     {
         WorldPacket data(SMSG_PARTY_MEMBER_STATS_FULL, 3+4+2);
         data << uint8(0);                                   // only for SMSG_PARTY_MEMBER_STATS_FULL, probably arena/bg related
--- a/src/server/game/Handlers/GroupHandler.cpp
+++ b/src/server/game/Handlers/GroupHandler.cpp
@@ -911,7 +911,7 @@ void WorldSession::HandleRequestPartyMemberStatsOpcode(WorldPacket &recvData)
     recvData >> Guid;
 
     Player* player = ObjectAccessor::FindConnectedPlayer(Guid);
-    if (!player || !GetPlayer()->GetGroup() || !(player->GetGroup()->GetGUID() != GetPlayer()->GetGroup()->GetGUID()))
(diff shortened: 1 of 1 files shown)


jackpoz:
https://github.com/jackpoz/BotFarm can be used to send any packet,  if you have the time to learn how to set it up

src/server/game/Handlers/GroupHandler.cpp
```diff
@@ -911,7 +911,7 @@ void WorldSession::HandleRequestPartyMemberStatsOpcode(WorldPacket &recvData)
     recvData >> Guid;
 
     Player* player = ObjectAccessor::FindConnectedPlayer(Guid);
-    if (!player)
+    if (!player || !GetPlayer()->GetGroup() || !(player->GetGroup()->GetGUID() == GetPlayer()->GetGroup()->GetGUID()))
```
jackpoz, line 914: You need to check if player->GetGroup() is null also

jackpoz (changes requested):

src/server/game/Handlers/GroupHandler.cpp
heyitsbench: > You need to check if player->GetGroup() is null also.

Corrected.

> You can compare the result of GetGroup() itself, no need to compare the guid, unless you have reasons that I am not aware about to do so.

Remnant from trying to get the condition to pass when I didn't realize I had a double negative hanging around. Corrected as well.

jackpoz (approved):

src/server/game/Handlers/GroupHandler.cpp
jackpoz: One could remove 1 of the null checks now that we check for != but we keep it like that.

PR #30128, opened 2024-08-02
Core/AuctionHouse: Fixed use after free when there was no bidder and seller was offline.

lineagedr:
**Changes proposed:**

-  Remove the code that deletes the item from SendAuctionExpired since the item is used after that in RemoveAuction and it's also being deleted there.

**Issues addressed:**

**Tests performed:**

- In RemoveAuction we're not actually deleting the item so I'm not sure if thats okay.


Shauren:
This leaks memory and leaves orphaned items in database

lineagedr:
> This leaks memory and leaves orphaned items in database

How about I store the items that will be deleted in a vector and pass the vector to RemoveAuction where they'll be deleted. Does that work?

PR #30135, opened 2024-08-03
Spells/Druid: Implement Astral Communion talent with Celestial Alignment

aquadeus:
**Changes proposed:**

-  Implement Astral Communion talent energize with Celestial Alignment

**Tests performed:**

Tested in-game


Shauren:
This one is incorrect, entering eclipse with Celestial Alignment/Incarnation should only energize once, not twice

Shauren (changes requested):
This one is incorrect, entering eclipse with Celestial Alignment/Incarnation should only energize once, not twice

aquadeus:
ok i updated the pr and fixed the double energize (energize happens when wrath or starfire procs the eclipse) with celestial alignment, now missing the energize part with incarnation of elune, but that will be in another pr

PR #30150, opened 2024-08-05
Spells/Warlock: Update Unstable Affliction script

aquadeus:
**Changes proposed:**

-  Update Unstable Affliction script to 11.0

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -962,38 +962,38 @@ class spell_warl_t4_2p_bonus : public AuraScript
...
-                    bp = target->SpellDamageBonusTaken(caster, aurEff->GetSpellInfo(), bp, DOT);
-                    bp *= 9;
-
-                    // backfire damage and silence
-                    CastSpellExtraArgs args(aurEff);
-                    args.AddSpellBP0(bp);
-                    caster->CastSpell(target, SPELL_WARLOCK_UNSTABLE_AFFLICTION_DISPEL, args);
-                }
-            }
+            float sp = caster->SpellBaseDamageBonusDone(GetSpellInfo()->GetSchoolMask()) * 0.30613f;
```
mdX7, line 977: Where does this value come from?

aquadeus: Effect 1: Id 6 (SPELL_EFFECT_APPLY_AURA)
Difficulty: Id 0 (DIFFICULTY_NONE)
BasePoints = 0.00 + spellPower * 0.30613
Targets (6, 0) (TARGET_UNIT_TARGET_ENEMY, NO_TARGET)
Aura Id 3 (SPELL_AURA_PERIODIC_DAMAGE), value = 0, misc = 0 (0), miscB = 0, amplitude = 0, periodic = 2000

Shauren: Hardcoding the value was a bad idea (already changed for 11.0) - it is available as a field in SpellEffect

src/server/scripts/Spells/spell_warlock.cpp
```diff
@@ -962,38 +962,38 @@ class spell_warl_t4_2p_bonus : public AuraScript
...
+            int32 damage = sp * GetEffectInfo(EFFECT_0).CalcValue() / 100;
+            CastSpellExtraArgs args;
+            args.AddSpellBP0(damage);
+            caster->CastSpell(dispelInfo->GetDispeller(), SPELL_WARLOCK_UNSTABLE_AFFLICTION_DAMAGE, args);
```
Shauren, line 981: Not setting trigger flags will cause this to interrupt warlocks current cast

src/server/scripts/Spells/spell_warlock.cpp
aquadeus: oh didnt know that, thanks for the tip

PR #30161, opened 2024-08-10
Spells/Warlock: Implement Volatile Agony talent

aquadeus:
**Changes proposed:**

-  Implement Volatile Agony talent

**Tests performed:**

Tested in-game


Shauren:
This had to be changed to spell instead of aura script - otherwise it would keep triggering on agony stacking by itself (when fixed)

aquadeus:
oh didnt thought about that, thanks

PR #30173, opened 2024-08-20
darwin/arm64 client support for master/11.0.2+

motivewc:
**Changes proposed:**
- Enable Mac ARM clients to connect to TrinityCore

**Issues addressed:**
Mac ARM client was previously denied for being on an unsupported platform and was missing infrastructure for auth seeds

**Tests performed:**
1. Compiles
2. I was able to achieve realm list and later in-world while running a custom patched Dragonflight Mac client on Apple Silicon. [See screenshot
](https://imgur.com/I4vsLSD)

N/A


Shauren:
It is currently disallowed because no launcher exists

motivewc:
I'm working on a client patcher that I'm willing to open source once it's done. I'm still working through verifying that I'm able to get in world, though, and it looks like there might be some work to do in adding ARM support to the world server, too. I could re-open this once things are further along if you want.

Fabi:
> I'm working on a client patcher that I'm willing to open source once it's done. I'm still working through verifying that I'm able to get in world, though, and it looks like there might be some work to do in adding ARM support to the world server, too. I could re-open this once things are further along if you want.

The only thing you need are the common patches from the arctium launcher (which you maybe used idk), the bnetserver to allow mac to connect and the authseed for arm mac. that's all

motivewc:
Yeah, I'd be nowhere without the Arctium open source stuff, as well as the tons of useful information in the Discords. I'm working through the process to find the auth seed information.

src/server/game/Server/WorldSocket.cpp
```diff
@@ -713,6 +713,13 @@ void WorldSocket::HandleAuthSessionCallback(std::shared_ptr<WorldPackets::Auth::
         digestKeyHash.UpdateData(buildInfo->Win64AuthSeed.data(), buildInfo->Win64AuthSeed.size());
     else if (account.Game.OS == "Mc64")
         digestKeyHash.UpdateData(buildInfo->Mac64AuthSeed.data(), buildInfo->Mac64AuthSeed.size());
+    else if (account.Game.OS == "MacA")
+        digestKeyHash.UpdateData(buildInfo->Mac64AuthSeed.data(), buildInfo->MacArmAuthSeed.size());
```
Winfidonarleyan, line 717:
```suggestion
        digestKeyHash.UpdateData(buildInfo->MacArmAuthSeed.data(), buildInfo->MacArmAuthSeed.size());
```

motivewc: Heh, woops. Didn't notice this as I'm still extracting the seed info.

motivewc:
I've tested the auth seed included with 11.0.2.56196 successfully and been able to get in-world. If anyone's looking for a client patcher for macOS, I wrote a [quick script](https://github.com/motivewc/wowpatch) to do the deed. 

Since opening the PR, it looks like we're on build 56263 so I'll see about extracting that auth seed too. I'd like to see macOS client support become a thing, which means there probably needs to be some automation to extract out the auth seed... What would your requirements be @Shauren?

V-Cyberpunk:
tested your patcher on windows. can't get it to work to login. is there something different on windows? tested with 10.2.7 dragonflight, but rsa modulus and ed25519 pub keys are the same there.

Shauren:
While I can add neccessary columns and code changes I cannot rely on anyone to keep providing auth seeds going forward

For this to be viable, I would need to a have a tool/script that can run on windows (or linux in WSL) - which means it cannot attempt to launch the client to extract the value

motivewc:
@Shauren How would you feel about merging this w/o a guarantee of future support? I know there is some tooling being worked on to automate the extraction by @Fabi , but realistically, I think this expectation largely matches the support already there for darwin/amd64 (which is to say, no one has provided an authseed in a while). While the situation isn't ideal, future PRs would at least just be a quick SQL update.

Fabi:
> @Shauren How would you feel about merging this w/o a guarantee of future support? I know there is some tooling being worked on to automate the extraction by @Fabi , but realistically, I think this expectation largely matches the support already there for darwin/amd64 (which is to say, no one has provided an authseed in a while). While the situation isn't ideal, future PRs would at least just be a quick SQL update.

When macOS arm support is added there should also be changes for Windows arm and maybe the possibility to add an alternative seed for China binaries. With that it might be a good idea to have a different table structure where every build/platform/region? should get it's own row for flexibility

Shauren:
Yes but not in this PR

PR #30184, opened 2024-08-24
Warlock/Spells: Fix Conflagrate energize

aquadeus:
**Changes proposed:**

- Fix Conflagrate energize

**Tests performed:**

Tested in-game


Shauren:
script was wrong, Conflagrate is supposed energize per target with Havoc active

PR #30187, opened 2024-08-26
Fix/DB : Missing `scrapping_loot_template`

TrinityClass:


jackpoz:
@MOPandaria care to elaborate what you are trying to do in this PR by filling the PR template ?

Shauren:
scrapping_loot_template is not a thing we have

PR #30196, opened 2024-08-27
Core/Conf: Update MaxPlayerLevel to 80 

hwis:


CraftedRO:
most likely this was valid

PR #30199, opened 2024-08-28
Advanced Flying (dragonriding)

Traesh:
**Changes proposed:**

-  Handle advanced flying (dragonriding)
-  
-  

**Issues addressed:**

**Tests performed:**

Build, tested ingame

- [ ] Todo in another PR : Handle other talents (vigor regen, etc)


src/server/game/Entities/Object/Position.h
```diff
@@ -57,6 +57,7 @@ struct TC_GAME_API Position
     // Better to limit access to _orientation field, to guarantee the value is normalized
 private:
     float m_orientation;
+    float m_pitch;
```
Shauren, line 60: rejected, just use pitch from m_movementInfo everywhere

Traesh: Done

src/server/game/Entities/Unit/StatSystem.cpp
```diff
@@ -86,11 +86,31 @@ void Unit::UpdateDamagePhysical(WeaponAttackType attType)
     }
 }
 
+uint32 Unit::GetCreateVigor() const
+{
+    // Must be in "higher to lower" order
+    std::vector<uint32> vigorTalentSpellIds = {
```
mdX7, line 92: would prefer `static constexpr` array here. std::vector is pretty over kill for just looping over 3 values

src/server/scripts/Spells/spell_advanced_flying.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "ScriptMgr.h"
+#include "Spell.h"
+#include "SpellAuraEffects.h"
+#include "SpellHistory.h"
+#include "SpellMgr.h"
+#include "SpellScript.h"
+#include <MovementPackets.h>
+#include <G3D/g3dmath.h>
+
+enum DragonRidingSpells
```
mdX7, line 35: AdvancedFlyingSpells

sql/updates/world/master/9999_99_99_00_world_dragonriding.sql
```diff
@@ -0,0 +1,18 @@
...
+INSERT INTO spell_script_names VALUES
+(374763, "spell_af_skyward_ascent"),
+(372610, "spell_af_skyward_ascent"),
+(372773, "spell_af_energy"),
+(372608, "spell_af_surge_forward"),
+(361584, "spell_af_whirling_surge");
+
+DELETE FROM `spell_linked_spell` WHERE spell_trigger = 406095 AND spell_effect = 423624;
+DELETE FROM `spell_linked_spell` WHERE spell_trigger = 372773 AND spell_effect = 433547;
+INSERT INTO `spell_linked_spell` VALUES
```
mdX7, line 16: `spell_linked_spell` is a table we no longer use and is scheduled for dropping at some point. Please use a SpellScript instead

src/server/game/Entities/Player/Player.cpp
```diff
@@ -30836,3 +30836,44 @@ bool Player::CanExecutePendingSpellCastRequest()
...
+    SendMessageToSet(addImpulse.Write(), true);
+}
+
+void Player::InitAdvancedFly()
+{
+    FlightCapabilityEntry const* flightCapabilityEntry = sFlightCapabilityStore.LookupEntry(GetFlightCapabilityID());
+    if (!flightCapabilityEntry)
+        return;
+
+    std::vector<std::tuple<OpcodeServer, float, Optional<float>>> advFlyValues = {
```
mdX7, line 30855: personally i wouldn't use a container here but construct each packet individually, as constructing the container and deconstructing it is time we could save at this point. We could still apply functional decomposition like you did here though.

something along
```c
        auto advFlyingPacket = WorldPackets::Movement::SetAdvFlyingSpeed(SMSG_MOVE_SET_ADV_FLYING_BANKING_RATE);
        advFlyingPacket.SequenceIndex = m_movementCounter++;
        advFlyingPacket.speed = flightCapabilityEntry->BankingRateMin;
        advFlyingPacket.maxSpeed = flightCapabilityEntry->BankingRateMax;
        SendDirectMessage(advFlyingPacket.Write());
```
we could even go a step further and put everything in constructor, like
```c
SendDirectMessage(WorldPackets::Movement::SetAdvFlyingSpeed(SMSG_MOVE_SET_ADV_FLYING_BANKING_RATE, flightCapabilityEntry->BankingRateMin, flightCapabilityEntry->BankingRateMax, _movementCounter++).Write());
```
that way we'd have even less code in this method - and its way more performant

(also on a side note of using containers here: std::vector would be overkill here aswell, an array would be enough)

src/server/game/Entities/Player/Player.cpp
```diff
@@ -30836,3 +30836,44 @@ bool Player::CanExecutePendingSpellCastRequest()
...
+        { SMSG_MOVE_SET_ADV_FLYING_PITCHING_RATE_DOWN,          flightCapabilityEntry->PitchingRateDownMin,         flightCapabilityEntry->PitchingRateDownMax          },
+        { SMSG_MOVE_SET_ADV_FLYING_PITCHING_RATE_UP,            flightCapabilityEntry->PitchingRateUpMin,           flightCapabilityEntry->PitchingRateUpMax            },
+        { SMSG_MOVE_SET_ADV_FLYING_TURN_VELOCITY_THRESHOLD,     flightCapabilityEntry->TurnVelocityThresholdMin,    flightCapabilityEntry->TurnVelocityThresholdMax     },
+        { SMSG_MOVE_SET_ADV_FLYING_SURFACE_FRICTION,            flightCapabilityEntry->SurfaceFriction,             {}                                                  },
+        { SMSG_MOVE_SET_ADV_FLYING_OVER_MAX_DECELERATION,       flightCapabilityEntry->OverMaxDeceleration,         {}                                                  },
+        { SMSG_MOVE_SET_ADV_FLYING_LAUNCH_SPEED_COEFFICIENT,    flightCapabilityEntry->LaunchSpeedCoefficient,      {}                                                  },
+    };
+
+    for (auto const& tuple : advFlyValues)
```
mdX7, line 30871: (sidenote when using containers with tuples: you can name the fields `for (auto const& [opcode, speed, maxSpeed] : advFlyValues)`)

src/server/game/Entities/Unit/StatSystem.cpp
```diff
@@ -86,11 +86,31 @@ void Unit::UpdateDamagePhysical(WeaponAttackType attType)
     }
 }
 
+uint32 Unit::GetCreateVigor() const
```
mdX7, line 89: maybe a name something like `GetMaxVigor` would be more appropriate, as the talents you are using below are increasing max vigor?

src/server/game/Entities/Player/Player.cpp
Traesh: Done, i still used an proxy method because i'd rather have a single point to modify if the packet change at some point in the future, and it should have no impact on perfs with inline

src/server/game/Entities/Player/Player.cpp
Traesh: Oooh, thanks !

src/server/game/Entities/Player/Player.h
```diff
@@ -2182,6 +2183,10 @@ class TC_GAME_API Player final : public Unit, public GridObject<Player>
 
         void RemovedInsignia(Player* looterPlr);
 
+        void InitAdvancedFly();
```
Nyr97, line 2186: InitAdvancedFlight instead?

Traesh: Hum, i can switch to "InitAdvancedFlying" as this is the opcodes names

src/server/game/Entities/Unit/Unit.h
```diff
@@ -1747,6 +1753,8 @@ class TC_GAME_API Unit : public WorldObject
         virtual bool CanEnterWater() const = 0;
         virtual bool CanSwim() const;
 
+        float GetAdvFlyingVelocity() const;
```
Nyr97, line 1756: GetAdvFlightVelocity?

Traesh: No, Blizz call it "AdvFly", short for "Advanced Flying" (check opcodes names)

src/server/game/Entities/Player/Player.h
mdX7: Following opcode naming sounds good to me

hygfrtqwe:
The first time is fine, as long as you try to mount it the second time, you won't be able to fly

src/server/game/Server/Packets/MovementPackets.h
```diff
@@ -710,6 +710,35 @@ namespace WorldPackets
...
+                SequenceIndex = sequenceIndex;
+                Speed = speed;
+                MaxSpeed = maxSpeed;
+            }
+
+            WorldPacket const* Write() override;
+
+            uint32 SequenceIndex = 1;
+            float Speed = .0f;
+            Optional<float> MaxSpeed;
```
Shauren, line 728: I want this split up into two classes `SetAdvFlyingSpeed` and `SetAdvFlyingMinMaxSpeeds` since this isn't a shared reading function in client and using Optional in packets implies that it is

src/server/game/Entities/Player/Player.h
Traesh: AdvFlying it is then, it will also match the movementInfo member

Nyr97: sounds better to me too

src/server/game/Entities/Unit/Unit.h
Nyr97: oh so it's a shortening, in that case we gucci

src/server/game/Entities/Player/Player.cpp
```diff
@@ -30836,3 +30836,38 @@ bool Player::CanExecutePendingSpellCastRequest()
...
+    SendAdvFlyingSpeed(SMSG_MOVE_SET_ADV_FLYING_PITCHING_RATE_UP,           flightCapabilityEntry->PitchingRateUpMin,           flightCapabilityEntry->PitchingRateUpMax);
+    SendAdvFlyingSpeed(SMSG_MOVE_SET_ADV_FLYING_TURN_VELOCITY_THRESHOLD,    flightCapabilityEntry->TurnVelocityThresholdMin,    flightCapabilityEntry->TurnVelocityThresholdMax);
+    SendAdvFlyingSpeed(SMSG_MOVE_SET_ADV_FLYING_SURFACE_FRICTION,           flightCapabilityEntry->SurfaceFriction);
+    SendAdvFlyingSpeed(SMSG_MOVE_SET_ADV_FLYING_OVER_MAX_DECELERATION,      flightCapabilityEntry->OverMaxDeceleration);
+    SendAdvFlyingSpeed(SMSG_MOVE_SET_ADV_FLYING_LAUNCH_SPEED_COEFFICIENT,   flightCapabilityEntry->LaunchSpeedCoefficient);
+}
+
+void Player::AddMoveImpulse(Position direction)
+{
+    auto addImpulse = WorldPackets::Movement::MoveAddImpulse();
```
Shauren, line 30863: no auto

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12472,6 +12482,15 @@ bool Unit::CanSwim() const
     return HasUnitFlag(UNIT_FLAG_RENAME | UNIT_FLAG_CAN_SWIM);
 }
 
+float Unit::GetAdvFlyingVelocity() const
+{
+    auto advFlying = m_movementInfo.advFlying;
```
Shauren, line 12487: 1. no auto
2. (not a big deal here) auto doesnt create references, this is a copy

Traesh: Changed

For my personal knowledge, is auto bad ?

Nyr97: auto is fine if the type is generically known to the people working (or even reading) the code, or even if you don't really care what the type is, however, it hinders the readability when the type is infrequent to the people navigating through it

Shauren: `auto` is banned in our codestyle (with the exception of container iterators and in places where auto is mandatory such as structured bindings)

aquadeus:
i wonder if SPELL_AURA_514 means spell aura mod advance flight speed or something related (spell 430833 - Soar (Racial))

Nyr97:
SPELL_AURA_514 is a modifier to adv fly max. velocity pact.

aquadeus:
ok so SPELL_AURA_513 is mod air friction, 514 mod max velocity and 515 mod lift coefficient, only missing 518

Traesh:
518 modify ADD_IMPULSE_MAX_SPEED

Shauren:
Issues fixed so far when merging parts to master
* Incorrect packet structures (missing MoverGUID)
* Aura ids mixed up
* Incorrect aura stacking behavior

src/server/game/Entities/Unit/StatSystem.cpp
```diff
@@ -86,11 +86,31 @@ void Unit::UpdateDamagePhysical(WeaponAttackType attType)
     }
 }
 
+uint32 Unit::GetMaxVigor() const
```
aquadeus, line 89: this is removed on 11.2.7

src/server/scripts/Spells/spell_advanced_flying.cpp
```diff
@@ -0,0 +1,211 @@
...
+{
+    void HandleHitTarget(SpellEffIndex /*effIndex*/)
+    {
+        if (Player* caster = GetCaster()->ToPlayer())
+        {
+            float SURGE_SPEED = 60.0f;
+
+            float destX = caster->GetPositionX() + SURGE_SPEED * std::cos(caster->GetOrientation());
+            float destY = caster->GetPositionY() + SURGE_SPEED * std::sin(caster->GetOrientation());
+            float destZ = caster->GetPositionZ() + SURGE_SPEED * std::tan(caster->m_movementInfo.pitch);
```
Nyr97, line 192: I would actually implement a getter for m_movementInfo first, and maybe one for pitch, though we should take into account that, in wow movement and space, Blizzard uses Roll, Pitch and Yaw, and each of these manipulate the direction differently (we have examples already in AreaTrigger):

- Roll is in charge of the rotation around the Y axis (this is what makes an entity rotate sideways, best example is an airplane banking left to force a left movement).
- Pitch is in charge of the rotation around the X axis (this is what makes entities rotate up and down as if a missile was making an arch with its trajectory).
- Yaw is in charge of the rotation around the vertical Z axis (this is what m_orientation is actually).

In this case, are you sure tan(pitch) is correct? That's a slope, not a unit-direction. Given that SURGE_SPEED is your impulse's magnitude, you want a total 3D speed and the standard is yaw + pitch:

x = cos(pitch) * cos(yaw)
y = cos(pitch) * sin(yaw)
z = sin(pitch)

As it currently stands, total speed becomes wrong (SURGE_SPEED / cos(pitch)) since it gets bigger the more you look up or down. Also, regardless of pitch (say 80.f), you will still dash at full speed horizontally, and when tan(pitch) gets into account, (say 90.f), there's a crazy Z impulse occuring.

src/server/scripts/Spells/spell_advanced_flying.cpp
```diff
@@ -0,0 +1,211 @@
...
+};
+
+// 361584 - Whirling Surge
+class spell_af_whirling_surge : public SpellScript
+{
+    void HandleHitTarget(SpellEffIndex /*effIndex*/)
+    {
+        if (Player* caster = GetCaster()->ToPlayer())
+        {
+            float SURGE_SPEED = 60.0f;
```
Nyr97, line 188: this can be constexpr

PR #30204, opened 2024-09-01
Core: Add auth seed for macOS arm64 build 56421

motivewc:
**Changes proposed:**
Quick SQL update for Mac ARM64 Auth Seed

**Issues addressed:**
N/A

**Tests performed:**
Tested in-game 

**Known issues and TODO list:** (add/remove lines as needed)
N/A

PR #30207, opened 2024-09-01
Add automation around database dumps

motivewc:
**THIS IS NOT READY FOR MERGE, BUT NEEDS FEEDBACK**

# Background
This PR proposes changes to the way that TrinityCore accepts database schema changes and solicits feedback for how to reduce developer friction.

In a recent [example PR](https://github.com/TrinityCore/TrinityCore/pull/30204/files), I wanted to make a one-line insert change to the `updates/auth/master` folder. In an ideal world, I'd be able to just submit the single SQL file, but we also require that the auth DB dump be updated.

This process is manual at the moment and a bit painful to ensure that the file excludes commentary that isn't useful. In addition to this, if someone else were to merge a database change in the intervening period between review and merge, then it necessitates regeneration again. Ideally, the worst case here would be someone else stealing my update file name and my needing to update the PR to rename a single file. 

# Implementation
These changes will automate the steps to generate the auth and characters database dumps. Developers would be expected to ONLY update the `sql/updates/.../master` folder, and a periodic database dump PR would be automatically opened once a week. This is, of course, tune-able, and even the PR could be merged automatically if desired. The workflow can also be triggered on-demand.

Here's an example of what the update PR looks like on my fork: https://github.com/motivewc/TrinityCore/pull/3

The process as written also generates the world and hotfixes dumps from the dev dump, however, these don't make it into the PR. This can be changed if there's ever the desire in the future.

# Concerns
## 1. This eliminates the SQL update check in `contrib/check_updates.sh` but additional tooling may be desired in its place.
I'd suggest that in place of that step, we:
1.  run the build and verify that any updates can be executed when the daemon runs with the `----update-databases-only` flag. This would ensure that someone has ultimately written valid SQL that can run against the state of the DB.
2. optionally add a linter to enforce good SQL hygiene such as quoting column names (perhaps [SQLFluff](https://github.com/sqlfluff/sqlfluff)?)

## 2.  I re-use the previous "master" build to generate SQL dumps
This felt like a reasonable shortcut, but the automation could just build from master and then do the same thing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
           password: $DOCKERHUB_PASSWORD
         environment:
           MYSQL_ALLOW_EMPTY_PASSWORD: yes
-          MYSQL_ROOT_PASSWORD: ''
+          MYSQL_ROOT_PASSWORD: ""
           MYSQL_DATABASE: test_mysql
     steps:
       - run:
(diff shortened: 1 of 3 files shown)


jackpoz:
> Ideally, the worst case here would be someone else stealing my update file name and my needing to update the PR to rename a single file.

To clarify, you are not expected to deal with this as a contributor, the maintainers will deal with it.

.circleci/config.yml
```diff
@@ -24,24 +24,6 @@ jobs:
...
-            mysql -h 127.0.0.1 -uroot auth < sql/base/auth_database.sql
-            ./contrib/check_updates.sh auth master auth 127.0.0.1
-            mysql -h 127.0.0.1 -uroot characters < sql/base/characters_database.sql
-            ./contrib/check_updates.sh characters master characters 127.0.0.1
-            echo "Importing world database structure"
-            mysql -h 127.0.0.1 -uroot world < sql/base/dev/world_database.sql
-            echo "Importing hotfixes database structure"
-            mysql -h 127.0.0.1 -uroot hotfixes < sql/base/dev/hotfixes_database.sql
-            echo "Importing world database updates"
-            cat sql/updates/world/master/*.sql | mysql -h 127.0.0.1 -uroot world
```
jackpoz, line 42: These checks verify that any sql update has valid SQL syntax, even if they don't update the table structure, and should be kept.

.github/workflows/gcc-build.yml
```diff
@@ -8,36 +8,50 @@ jobs:
...
+          cd bin
+          make test
+      - name: Check executables
+        run: |
+          cd ${{ github.workspace }}/check_install/bin
+          ./bnetserver --version
+          ./worldserver --version
+
+      - name: Archive
+        if: github.ref == 'refs/heads/master'
```
jackpoz, line 46: is this check similar to https://github.com/TrinityCore/TrinityCore/blob/b1d8061a0f848241fe70ab0d718e829836cb5601/.circleci/config.yml#L96C21-L96C22 or ? As in, "master" is the name of the branch and multiple branches should be supported ?

Shauren:
As explained before, users submitting PRs are not expected to modify files in sql/base directory

PR #30209, opened 2024-09-05
Scripts/Command Use linked bonuses & context for .additem command

Traesh:
**Changes proposed:**

-  Let .additem commands use the bonuses & itemcontext from linked items.
-  Allow to add item directly from link with correct bonuses & context
-  

**Issues addressed:**

None

**Tests performed:**

Built, tested ingame

### DIFF
--- a/src/server/scripts/Commands/cs_misc.cpp
+++ b/src/server/scripts/Commands/cs_misc.cpp
@@ -1159,6 +1159,8 @@ public:
             return false;
 
         uint32 itemId = 0;
+        std::vector<int32> bonusListIDs;
+        ItemContext itemContext = ItemContext::NONE;
 
         if (args[0] == '[')                                        // [name] manual form
         {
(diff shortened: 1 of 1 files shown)


CraftedRO:
if this does what I think, I guess it will need some merge to 3.3.5 too.

Traesh:
@CraftedRO Not needed afaik, there is no bonus or context on items in 3.3.5

CraftedRO:
ah ok then that's another case to be able to properly add items with this command via itemname / link.

Shauren:
I like the idea but not the implementation. Ideally it would be directly accessible through typed command args (Hyperlink<item>) but im not judging it yet (no idea if its easily doable)

Shauren (changes requested):
Needs changing after 1f60e76b7f157af6a24fa3c962b59f3391b21e75

PR #30210, opened 2024-09-05
Core/Unit Allow to set triggerSpellId in jump_charge_params

Traesh:
**Changes proposed:**

-  Allow to set triggerSpellId in jump_charge_params
-  Add various jump data from sniffs

**Issues addressed:**

None

**Tests performed:**

Built, tested ingame


Nyr97:
I don't think this implementation is on the right track: 

374704 - Fiery Focus (Jump Charge) used by Chargath, Bane of Scales [Neltharus] has TriggerSpellId 375055, which is supposed to be triggered at the end of the jump spline on the target who was passed as jump destination. However, some other jump charge effects use completely different targets for the jump effect and the triggeredSpellId targets.

Traesh:
I'm not sure to see the problem here.

If the triggered spell require an unit, it is now correctly passed for the arrival cast, which was not the case before.
If the triggered spell target isn't an unit, this won't change anything as the Target won't be used by the spell target selector at all.

This is the exact implementation we already have for 41 SPELL_EFFECT_JUMP, we always set the Target as unitTarget.

Nyr97:
I understand that, however, SPELL_EFFECT_JUMP does not have as much mystery as EFFECT_JUMP_CHARGE does, most of its values are serverside, see https://github.com/TrinityCore/TrinityCore/pull/28142#issuecomment-1193408942 for reference (we need a flag field, but it's completely out of the scope of this PR).

It's impossible to know all of the fields they might be using, but it's best to keep only the ones we're pretty sure about in that table; I think it's best to create a new one instead: jump_charge_actions which allows us to define given actions for a miscValue and spell, with a field to decide what action to do (cast spell for example), when that action should be triggered (spline launch or finish), and what the target for that action should be (caster, target, for example).

mdX7 (approved):

PR #30214, opened 2024-09-07
Core/Crafting : Basis of the new crafting system

Traesh:
**Changes proposed:**

- Implement the basis for the new crafting system
- Allow to craft items and correctly determine the quality of the crafted item based on the player skills & reagent quality
- Handle optional reagent
- https://youtu.be/zt9Ew8MxuqY

**Issues addressed:**

None afaik

**Tests performed:**

Built, tested ingame

I will only try to fix the first one for this "V1" version, but i still list some things that will be required :

- [x] The item with "random properties" are not correctly generated : the created item still display "random properties" on them
- [ ] Profession talents are not implemented
- [ ] Profession items (knife, etc) are not used in the math
- [ ] Finishing reagent are not tested
- [ ] Profession sub-stats (Concentration, ingenuity, ...) are not implemented


src/server/game/DataStores/DB2Structure.h
```diff
@@ -2877,6 +2942,56 @@ struct MawPowerEntry
...
+    int32 ID;
+    int32 Field_9_0_1_33978_001;
+    int32 PlayerConditionID;
+    int32 ReagentType;
+    int8 Field_10_0_2_46091_005;
+};
+
+struct ModifiedCraftingSpellSlotEntry
+{
+    int32 ID;
```
Nyr97, line 2986: not uint? 👀

Ovahlord: there is no need to play smartass in db2 structs because the field types will be checked against the client extracted db2 meta on startup. If it loads it's correct. If not, he'll fix it. See the WIP in the title.

Nyr97: i don't know what the comment about the WIP or about playing smartass have to do with anything I pointed out, there's no reason not to be using unsigned integers when the ID field only stores positive values

Nyr97: In any case, the whole crafting system is huge and it's best to break it down to multiple PRs, so it's easier to review. Good so far!

Nyr97:
In any case, the whole crafting system is huge and it's best to break it down to multiple PRs as you pointed out (there's a lot of missing stuff on that list, however), so it's easier to review. Good so far!

src/server/game/DataStores/DB2Structure.h
MaxtorCoder: It should indeed be uint32, simply follow the rest of the structures.

Traesh:
I think i will stop here for this PR so i can already update what need to be changed before starting other part of the craft system.

@Shauren you can do your thing :p

MaxtorCoder:
Fix codestyle and nopch

Traesh:
> Fix codestyle and nopch

codestyle is SQL only, nopch in progress.

coreretail6:
@Traesh Keep up :-) a lot of users (and me) are tuned by your PRs (this and Skyriding)

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+#include "Item.h"
+#include "ItemBonusMgr.h"
+#include "SharedDefines.h"
+#include "Spell.h"
+#include "SpellInfo.h"
+#include "SpellMgr.h"
+#include "SpellPackets.h"
+#include "Player.h"
+
+SpellCastResult Crafting::DoCraft(uint32 craftingDataId)
```
Nyr97, line 31: You're mixing responsibilities like validation and computation. Given that this system is very tedious, I would break it down into smaller helper functions since it's easier to debug and maintain in the long term:

ValidateReagents
Determine/ComputeCraftQuality
ConsumeReagents

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+    // If we didn't found anything, we have a problem. Return and don't take reagent from player.
+    if (!craftedItemId)
+        return SpellCastResult::SPELL_FAILED_ERROR;
+
+    ItemContext context = _spell->GetSpellInfo()->HasAttribute(SPELL_ATTR0_IS_TRADESKILL) ? ItemContext::Trade_Skill : ItemContext::NONE;
+
+    std::vector<int32> bonusLists;
+    for (int32 bonusTreeID : bonusTreeIds)
+    {
+        std::vector<int32> treeBonusList = ItemBonusMgr::GetBonusListsForItem(craftedItemId, { context, {}, {}, bonusTreeID, qualityTier });
```
Nyr97, line 129: these are quite large structures being created and iterated in nested loops, you could use unordered_set<uint32> for quick ItemID checks

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+        bonusLists.insert(std::end(bonusLists), std::begin(treeBonusList), std::end(treeBonusList));
```
Nyr97, line 130: this is creating unnecessary copies by inserting directly, it's better to reserve and emplace_back or std::move for that matter

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+        {
+            std::vector<ModifiedCraftingReagentItemEntry const*> reagentItems = sDB2Manager.GetMCRReagentItemByCategory(category->ID);
+
+            for (ModifiedCraftingReagentItemEntry const* reagentItem : reagentItems)
+            {
+                std::set<ItemSparseEntry const*> requiredItems = sDB2Manager.GetItemSparseByMCRReagentItem(reagentItem->ID);
+
+                for (WorldPackets::Spells::SpellCraftingReagent const& providedReagent : providedReagentItr->second)
+                {
+                    bool hasItem = std::count_if(std::cbegin(requiredItems), std::cend(requiredItems), [providedReagent](ItemSparseEntry const* itemSparse) {
```
Nyr97, line 96: pretty inefficient, it's best to do .contains(providedReagent.ItemID) for 1(O) lookup if you decide to pre-process them

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+    uint32 totalReagentWeigth = CalculateTotalReagentWeights();
+
+    CraftingDifficultyEntry const* craftingDifficultyEntry = sCraftingDifficultyStore.LookupEntry(_craftingData->CraftingDifficultyID);
+    if (!craftingDifficultyEntry)
+        return skillValue;
+
+    float maximumReagentBonus = round(float(_craftingData->CraftingDifficulty) * float(craftingDifficultyEntry->CraftSkillBonusPercent) / 100.f);
+
+    for (ModifiedCraftingSpellSlotEntry const* spellSlot : sDB2Manager.GetMCRSpellSlotBySpell(_spell->GetSpellInfo()->Id))
```
Nyr97, line 161: this call is done multiple times in different functions, you could retrieve it once in DoCraft and pass it as const& to GetSkillLevelForCraft() and CalculateTotalReagentWeights() and save resources

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+    {
+        for (ModifiedCraftingCategoryEntry const* category : sDB2Manager.GetMCRCategoryByReagentSlot(spellSlot->ModifiedCraftingReagentSlotID))
+        {
+            auto providedReagentItr = _spell->m_targets.ModifiedCraftingReagents.find(spellSlot->Slot);
+            if (providedReagentItr != _spell->m_targets.ModifiedCraftingReagents.end())
+            {
+                for (WorldPackets::Spells::SpellCraftingReagent const& providedReagentForSlot : providedReagentItr->second)
```
Nyr97, line 168: providedReagentItr->second being accessed repeatedly inside deep loops, it's best to extract to a local reference directly

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+#include "SpellPackets.h"
+#include "Player.h"
+
+SpellCastResult Crafting::DoCraft(uint32 craftingDataId)
+{
+    std::vector<ModifiedCraftingSpellSlotEntry const*> spellSlots = sDB2Manager.GetMCRSpellSlotBySpell(_spell->GetSpellInfo()->Id);
+    _craftingData = sCraftingDataStore.LookupEntry(craftingDataId);
+
+    if (!_craftingData || spellSlots.empty())
+        return SpellCastResult::SPELL_FAILED_ERROR;
```
Nyr97, line 37: I understand we need to send a SpellCastResult as result, but it'd be great if we could expand it to something that better reflects debuggability with some type of enum class CraftingFailureReason where we add MissingRequiredReagent, InvalidReagentItem, QualityTierMissing, etc.

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+        for (ModifiedCraftingCategoryEntry const* category : sDB2Manager.GetMCRCategoryByReagentSlot(spellSlot->ModifiedCraftingReagentSlotID))
+        {
+            auto providedReagentItr = _spell->m_targets.ModifiedCraftingReagents.find(spellSlot->Slot);
+            if (providedReagentItr != _spell->m_targets.ModifiedCraftingReagents.end())
+            {
+                for (WorldPackets::Spells::SpellCraftingReagent const& providedReagentForSlot : providedReagentItr->second)
+                {
+                    if (CraftingReagentQualityEntry const* reagentQuality = sDB2Manager.GetCraftingReagentQualityByItem(providedReagentForSlot.ItemID))
+                    {
+                        float reagentWeight = float(category->MatQualityWeight) * float(providedReagentForSlot.Quantity) / 2.0f * float(reagentQuality->OrderIndex);
```
Nyr97, line 172: why is it divided by 2.f? no documentation as to why, in any case, moving it to a constexpr float inside a namespace for Crafting is best so we have a name for whatever this is

also, I recommend encapsulating the whole in a function like CalculateReagentWeight()

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+{
+    // If this is a simple item (no quality), we have the id
+    uint32 newItemId = _craftingData->CraftedItemID;
+
+    // If it's a reagent with quality, get id from CraftingDataItem
+    if (!newItemId)
+    {
+        std::vector<uint32> craftedItemList = sDB2Manager.GetCraftingDataItemIDByCraftingData(_craftingData->ID);
+        if (craftedItemList.size() < qualityTier)
+            return 0;
```
Nyr97, line 237: a log here would be great along with return 0

src/server/game/Craft/Crafting.h
```diff
@@ -0,0 +1,52 @@
...
+class Spell;
+
+struct CraftingDataEntry;
+
+enum SpellCastResult : int32;
+
+class TC_GAME_API Crafting
+{
+public:
+    Crafting(Player* player, Spell* spell) :
```
Nyr97, line 34: I'd set the costructor plain Crafting(Player* player, Spell* spell) and set nullptr to the pointer in the private section

src/server/game/Craft/Crafting.h
```diff
@@ -0,0 +1,52 @@
...
+        _player(player), _spell(spell), _craftingData(nullptr) { }
+
+    SpellCastResult DoCraft(uint32 craftingDataId);
```
Nyr97, line 37: I'd force non-copiable semantics by adding:

`    Crafting(Crafting const&) = delete;
    Crafting& operator=(Crafting const&) = delete;
`

src/server/game/Craft/Crafting.h
```diff
@@ -0,0 +1,52 @@
...
+struct CraftingDataEntry;
+
+enum SpellCastResult : int32;
+
+class TC_GAME_API Crafting
+{
+public:
+    Crafting(Player* player, Spell* spell) :
+        _player(player), _spell(spell), _craftingData(nullptr) { }
+
```
Nyr97, line 36: I would also enforce non-copiable semantics

```suggestion
    Crafting(Crafting const&) = delete;
    Crafting& operator=(Crafting const&) = delete;
```

src/server/game/Craft/Crafting.h
Nyr97: we could even set noexcept since both pointers are required and must outlive this object, and even explicit to prevent accidental conversions (unlikely, but it's nice practice) for one-argument constructors

```suggestion
explicit Crafting(Player* player, Spell* spell) noexcept;
```

src/server/game/Craft/Crafting.h
```diff
@@ -0,0 +1,52 @@
...
+    uint32 GetSkillLevelForCraft();
+    uint32 GetSkillIdForSpell(uint32 spellId);
+    uint32 CalculateTotalReagentWeights();
+    void InitCraftingStatModifier(Item* item);
+    uint32 GetCraftedItemIdForQuality(uint32 qualityTier);
+
+    Player* _player;
+    Spell* _spell;
+
+    CraftingDataEntry const* _craftingData;
```
Nyr97, line 49: In-class default initialisation as nullptr maybe

src/server/game/Craft/Crafting.h
```diff
@@ -0,0 +1,52 @@
...
+class TC_GAME_API Crafting
+{
+public:
+    Crafting(Player* player, Spell* spell) :
+        _player(player), _spell(spell), _craftingData(nullptr) { }
+
+    SpellCastResult DoCraft(uint32 craftingDataId);
+
+private:
+    uint32 GetSkillLevelForCraft();
```
Nyr97, line 40: GetSkillLevelForCraft(), GetSkillIdForSpell(), CalculateTotalReagentWeights(), GetCraftedItemIdForQuality() are functions return values that are almost always critical for the logic of CraftingResult, so I'd mark them as [[nodiscard]] to warn us about their possibly non-use when called

src/server/game/Craft/Crafting.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+    for (ModifiedCraftingSpellSlotEntry const* spellSlot : sDB2Manager.GetMCRSpellSlotBySpell(_spell->GetSpellInfo()->Id))
+        for (ModifiedCraftingCategoryEntry const* category : sDB2Manager.GetMCRCategoryByReagentSlot(spellSlot->ModifiedCraftingReagentSlotID))
+            totalWeight += category->MatQualityWeight * spellSlot->ReagentCount;
+
+    return totalWeight;
+}
+
+// Todo : We currently only set random stats, but crafter
+// can define stats using a "Missive" item
```
Nyr97, line 206: those "missives" fall into "Optional Reagents" and should be processed during worldpacket SpellOptionalReagent, we need to store them maybe and then iterate over them to SetModifier

KamiliaBlow:
12.0.0 - No longer works due to a change in the SpellCraftingReagent structure.

Issue #26243, opened 2021-03-14
sometimes people enters one instance dead


Aokromes:
**Description:**

i have seen this bug myself before i started to use the dynamic resurrect patch, with it i got few players reports, idk how to reproduce at all.

**Expected behaviour:**

people must resurrect on enter instance

**Steps to reproduce the problem:**

1. die
2. release
3. enter instance, sometimes you will be as ghost inside.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

1ae55803b6fb

**Operating system:** 

debian 10 x64

CraftedRO:
I think I had find a way to reproduce this but only on a gm level account , go at the entrance of fos / pos / hor for example then .die yourself there (outside of dungeon entrance) come with spirit and enter in any dungeon .

https://user-images.githubusercontent.com/24683355/111073940-7cbdec00-84e9-11eb-92af-d40816ee5775.mp4

CraftedRO:
ok now I was able to reproduce it on a normal account too , but Il edit this in a second :))

CraftedRO:
After relogging as spirit (after I died in ICC then I left party) when I log back on if someone invite me in a group / then it convert to raid again I'm able to enter ICC as a ghost and this will appear in DBErrors log :
```
SQL(p): INSERT INTO character_aura (guid, casterGuid, itemGuid, spell, effectMask, recalculateMask, stackCount, amount0, amount1, amount2, base_amount0, base_amount1, base_amount2, maxDuration, remainTime, remainCharges, critChance, applyResilience) VALUES (13, 13, 0, 63531, 6, 6, 1, 0, 3, 3, 0, -1, -1, -1, -1, 0, 0.0, 1)
 [ERROR]: [1062] Duplicate entry '13-13-0-63531-6' for key 'PRIMARY'
```

Jildor:
Maybe that can be a way to reproduce #22234

CraftedRO:
Interesting is that DBErrors appear only when doing the above steps on a Paladin indeed .

illfated:
Not too much data present in `characters`.`character_aura` for a single player:
```sql
SELECT * FROM `character_aura`;
```
```
guid  casterGuid  itemGuid   spell  effectMask  recalculateMask  stackCount  amount0  amount1  amount2  base_amount0  base_amount1  base_amount2  maxDuration  remainTime  remainCharges  critChance  applyResilience
[...]

PR #30228, opened 2024-09-13
Core/Players: Allow delayed teleports to be executed even when not alive

### LINKED ISSUE #17291: [3.3.5]: Bug with unstuck
**Description**:
Bug with unstuck. NOT blizzlike.
If we using "unstuck", we teleports to the starting location.
We need to move to some other X, Y, or Z. Like official server.
1. Create a character
2. Go to the random location
3. Use "unstuck"

3.3.5
**TC hash/commit**:  
3bba685
**TDB version**:  
Maybe, all.
**Operating system**:  
All.

Jildor:
**Changes proposed:**

-  Allow delayed teleports to be executed even when not alive

not alive check was added here: https://github.com/TrinityCore/TrinityCore/commit/4a8a89e1db1f103020ffecbb570e1c570c2e0d21#diff-de71a0bef2855962354cadbe36e913957a5017a1821e44753fae6db29a21f7ebR1393
I'm not sure in which cases this was necessary, so I'm not sure if it's still necessary.
But this check prevents the dead player using unstuck from being able to teleport to the graveyard. (RepopAtGraveyard)
https://github.com/TrinityCore/TrinityCore/blob/f2a83e8238676a030d814741467bc0a43c0d88d5/src/server/game/Spells/SpellEffects.cpp#L3707-L3718

I don't know what you think, can that check be removed or what can we do?

**Issues addressed:**

Closes #17291
Updates #26243 (https://github.com/TrinityCore/TrinityCore/issues/26243#issuecomment-1964089498)

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -1323,9 +1323,7 @@ void Player::Update(uint32 p_time)
             m_hostileReferenceCheckTimer -= p_time;
     }
 
-    //we should execute delayed teleports only for alive(!) players
-    //because we don't want player's ghost teleported from graveyard
-    if (IsHasDelayedTeleport() && IsAlive())
+    if (IsHasDelayedTeleport())
         TeleportTo(m_teleport_dest, m_teleport_options);
 }
 


CraftedRO:
any will of merging this ?

CraftedRO:
Probably was a typo and it had to be !isAlive() instead If that make sense.

Jildor:
No, the implementation of && IsAlive() was intentional. Negate that check would be an error, as it would only execute delayed teleports if the player is not alive.

I haven't found the reason why that check was set there, nor have I found a reason why it should be there currently, other than fixing the linked issues.

PR #30229, opened 2024-09-13
Core/Maps: Do not allow entering an instance if the player is not alive and does not have the corpse inside or in inner instance

Jildor:
**Changes proposed:**

-  Do not allow entering an instance if the player is not alive and does not have the corpse inside or in inner instance

I don't see any reason to allow player access to the instance if is dead and has no corpse.
I found this case in wowhead about: https://www.wowhead.com/blue-tracker/topic/servers-just-went-down-and-i-cant-get-corpse-1567936776

**Issues addressed:**

Update #26243

**Tests performed:**

tested in-game
test: https://github.com/TrinityCore/TrinityCore/issues/26243#issuecomment-1945985713
test: https://github.com/TrinityCore/TrinityCore/issues/26243#issuecomment-2226858498
also test enter in Blackwing lair, die and go in ghost to Blackrock Spire and go inside Blackrock Spire in ghost, enter in Blackwing lair and revive.
(instance_template parent_map)


CraftedRO:
any will of merging this ?

PR #30231, opened 2024-09-13
Spells/DK: Fix talent Brittle

aquadeus:
**Changes proposed:**

-  Fix talent Brittle

**Tests performed:**

Tested in-game


Shauren (changes requested):
This is incomplete, misses `spell_proc` data to restrict it to diseases only (Death and Decay ticks also trigger PROC_FLAG_DEAL_HARMFUL_PERIODIC)

aquadeus:
proc added

Issue #23290, opened 2019-05-19
Bug with a reputation from the Aldor/the Scryers.


Infernales:
**Description:**

When reputation changes, the AtWar flag is not reset.

**Current behaviour:**

When we get reputation from these two fractions, it is added to the 'character_reputation' table with the flag 81, but as soon as we bring the reputation to a negative value, the flag changes to 83. If we want to make a positive reputation again, the flag will remain 83. Accordingly, we will remain with a faction in war mode.

**Expected behaviour:**

It should not be like that.

**Steps to reproduce the problem:**

1. Create character.
2. .mod rep 932/934 9999999
3. Relog.
4. .mod rep 932/934 -9999999
5. Relog.
6. .mod rep 932/934 9999999
7. Relog.
8. Open the tab with a reputation.

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

a0d41c3

**Operating system:** CHANGEME OS

Linux

Vasago:
Confirmed in the latest Trinity core https://github.com/TrinityCore/TrinityCore/commit/c7b8bf98a91d9f4a6fc34ac3139e5c8288a6145e @Killyana - maybe a DB bug? It is looong time bug and also in newly created characters.

Rushor:
```diff
diff --git a/src/server/game/Reputation/ReputationMgr.cpp b/src/server/game/Reputation/ReputationMgr.cpp
index b870f40801..6f3e60b9d7 100644
--- a/src/server/game/Reputation/ReputationMgr.cpp
+++ b/src/server/game/Reputation/ReputationMgr.cpp
@@ -478,6 +478,18 @@ void ReputationMgr::SetAtWar(RepListID repListID, bool on)
 
 void ReputationMgr::SetAtWar(FactionState* faction, bool atWar) const
 {
+    if (atWar)
+    {
+        // Sholazar Factions
+        if ((faction->ID == 1104 && _player->GetReputationRank(1104) < REP_UNFRIENDLY) || (faction->ID == 1105 && _player->GetReputationRank(1105) < REP_UNFRIENDLY))
+            faction->Flags |= FACTION_FLAG_AT_WAR;
+        // Aldor the Scryers Faction
+        if ((faction->ID == 932 && _player->GetReputationRank(932) < REP_UNFRIENDLY) || (faction->ID == 934 && _player->GetReputationRank(934) < REP_UNFRIENDLY))
[...]

PR #30235, opened 2024-09-14
Core/Reputation: Properly reset Aldor/Scryers at warr flag

CraftedRO:
**Changes proposed:**

- Attempt to fix Aldor vs Scryer reputation issue described in the targeted issue.
- The fix was posted in [the issue comment section](https://github.com/TrinityCore/TrinityCore/issues/23290#issuecomment-787598949).
-  Credit goes to [Jinnaix](https://github.com/Jinnaix) and [ckegg](https://github.com/ckegg)

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/23290

**Tests performed:**

Does it build, tested in-game as can be seen here:

https://github.com/TrinityCore/TrinityCore/issues/23290#issuecomment-2350902619

- [ As I already mentioned, there's need a reputation state change in order to see effects so if a character does already have exaulted it won't see any difference, not sure if this is the expected result but is a lot better than nothing.]


aquadeus:
could this be applied to sholazar basin reps too?

CraftedRO:
> could this be applied to sholazar basin reps too?

well in my tests that one was working correctly even without this PR

CraftedRO:
last sync, will keep this open one more week

PR #30250, opened 2024-09-17
Spells/DK: Fix proc of Frost Fever

aquadeus:
**Changes proposed:**

-  Fix proc of Frost Fever

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -660,6 +660,25 @@ class spell_dk_festering_strike : public SpellScript
...
+class spell_dk_frost_fever_proc : public AuraScript
+{
+    bool Validate(SpellInfo const* spellInfo) override
+    {
+        return ValidateSpellEffect({ { spellInfo->Id, EFFECT_0 } });
+    }
+
+    bool CheckProc(AuraEffect const* /*aurEff*/, ProcEventInfo const& /*eventInfo*/) const
+    {
+        return roll_chance_i(GetEffect(EFFECT_0)->GetAmount());
```
Nyr97, line 673: It's always recommendable to pass CalcAmount in case any effect modifies this amount

mdX7: that is for effects which aren't within same aura, theres no need to recalculate for no reason

Nyr97: but what if a different aura modifies base value of this aura?

mdX7: thats already taken into account in `GetAmount`

Shauren (approved):

PR #30255, opened 2024-09-21
Core/Database: dropped support for MariaDB and MySQL 5.7 and set 8.0.34 as the new required minimum version

Ovahlord:
Due to MySQL 5.7 being no longer supported and MariaDB being based on 5.7, both versions will no longer be supported to compile or run the source. MySQL 8.0.34 is set as a new standard as it is the first 8.0 version that started using OpenSSL 3 which we already require as well.

This superceedes the previous PR of depcecating these versions.

- MySQL 5.7 and MariaDB are no longer accepted to compile or run the applications
- Added a static_assert that throws a compile error when the mysql version is too low to reduce user errors

There is no point in arguing around anymore. It will happen. You had plenty of time to update your setup, the grace period is over.

### DIFF
--- a/src/server/database/Database/DatabaseWorkerPool.cpp
+++ b/src/server/database/Database/DatabaseWorkerPool.cpp
@@ -41,12 +41,12 @@
 #include <boost/stacktrace.hpp>
 #endif
 
-constexpr uint64 MIN_MYSQL_SERVER_VERSION = 80834;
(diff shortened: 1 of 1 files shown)


jackpoz:
https://dev.mysql.com/doc/refman/8.0/en/upgrading.html useful link to upgrade from 5.7 to 8.0

Shauren:
Will do this in cmake

PR #30259, opened 2024-09-24
Core/Conversations: Hackfix for NextConversationLineID overflow

meji46:
**Changes proposed:**

-  Field NextConversationLineID from ConversationLine.db2 is still uint16 so there is an overflow for referenced ids > 65535. Until db2 has the type changed to uint32 we need this hack to make new conversations work
-  Example from db2: ID 66067, NextConversationLineID: 532 (value should be 65536+532=66068)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game

None

### DIFF
--- a/src/server/game/Globals/ConversationDataStore.cpp
+++ b/src/server/game/Globals/ConversationDataStore.cpp
@@ -203,9 +203,9 @@ void ConversationDataStore::LoadConversationTemplates()
     {
         if (conversationLine && conversationLine->NextConversationLineID)
         {
-            uint32 const firstLineId = 60000; // Arbitrary id to cover the affected rows
+            static constexpr uint32 FirstLineId = 60000; // Arbitrary id to cover the affected rows
 
-            if (conversationLine->ID > firstLineId && conversationLine->NextConversationLineID < (sConversationLineStore.GetNumRows() - USHRT_MAX - 1))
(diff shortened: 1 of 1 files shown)


src/server/game/Globals/ConversationDataStore.cpp
```diff
@@ -198,11 +198,27 @@ void ConversationDataStore::LoadConversationTemplates()
         TC_LOG_INFO("server.loading", ">> Loaded 0 Conversation actors. DB table `conversation_actors` is empty.");
     }
 
+    // TODO: Remove this hack when NextConversationLineID is changed to uint32
+    auto getNextConversationLineId = [&](ConversationLineEntry const* conversationLine)
+    {
+        if (conversationLine && conversationLine->NextConversationLineID)
+        {
+            uint32 const firstLineId = 60000; // Arbitrary id to cover the affected rows
```
mdX7, line 206: can we `static constexpr` that?

mdX7 (commented):
does `FirstLineId` of `conversation_template` overflow aswell or just `Next[...]`?

meji46:
Just NextConversationLineID

funjoker:
I’m wondering if client has „special handling“ in its code for this.

mdX7:
> I’m wondering if client has „special handling“ in its code for this.

client is not using the field

mdX7 (approved):

Issue #16608, opened 2016-02-14
[DB] Eastvale Peasants in Elwynn Forest


Kittnz:
Bug: these Eastvale Peasants in Elwynn Forest are standing around idle, are not cutting wood and returning the wood.

Info: 
http://wotlk.openwow.com/npc=11328

8c38bab0e3f7b280a9ac0f56b848d66cb8fd7bc0

dr-j:
well the chopping is probably just a emote state as with other mobs which perform that action as for returning the wood that needs wp.

Kittnz:
i haven't found any waypoints yet for these in sniffs or they are incomplete. I think this will need to be done by hand or if we find sniffs that hold the wp's.

Killyana:
Waypoint are done just some emotes are missing.

CraftedRO:
c3e43d3d31d8 those are remaining things I found here:
```sql
SELECT * FROM `waypoint_data` WHERE `id`=813490; -- point 3 is not blizzlike and wood cutting near waggon isn't too
```
after they put the wood in the waggon they shoud face supervisor raelen and wait a bit (~5sec) then short after they sound:
(checked on retail 10.0.7) ready to work / all right / off I go then (all 3 in row for each peasant)
and then they start to walk to the trees again.

CraftedRO:
72f148916c1b6cafa7c2f845f37c62711eb2d646 still valid.

PR #30262, opened 2024-09-26
Scripts/Elwynn Forest: Eastvale Peasant scripting

ghost:
Replaced hacky scripting for Eastvale Peasants with proper scripting.

**Changes proposed:**

-  Replace Hacky SAI and pathing with C++ script. This scripting operates the same as official, a repeating order of 5 paths.

**Tests performed:**

Builds. Tested in game.

Closes https://github.com/TrinityCore/TrinityCore/issues/16608

### DIFF
--- a/src/server/scripts/EasternKingdoms/zone_elwynn_forest.cpp
+++ b/src/server/scripts/EasternKingdoms/zone_elwynn_forest.cpp
@@ -281,9 +281,9 @@ private:
 
 enum SupervisorRaelen
 {
-    EVENT_FIND_PEASENTS  = 1,
-    EVENT_NEXT_PEASENT   = 2,
-    NPC_EASTVALE_PEASENT = 11328
+    EVENT_FIND_PEASANTS  = 1,
+    EVENT_NEXT_PEASANT   = 2,
+    NPC_EASTVALE_PEASANT = 11328
 };
 
 struct npc_supervisor_raelen : public ScriptedAI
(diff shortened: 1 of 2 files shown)


src/server/scripts/EasternKingdoms/zone_elwynn_forest.cpp
```diff
@@ -275,7 +275,260 @@ struct npc_cameron : public ScriptedAI
...
+                case EVENT_FIND_PEASENTS:
+                {
+                    GuidVector tempGUIDs;
+                    std::list<Creature*> peasents;
+                    GetCreatureListWithEntryInGrid(peasents, me, NPC_EASTVALE_PEASENT, 100.f);
+                    for (Creature* peasent : peasents)
+                    {
+                        tempGUIDs.push_back(peasent->GetGUID());
+                    }
+                    peasentGUIDs.push_back(tempGUIDs[2]);
```
Ovahlord, line 333: peasent

src/server/scripts/EasternKingdoms/zone_elwynn_forest.cpp
```diff
@@ -275,7 +275,260 @@ struct npc_cameron : public ScriptedAI
...
+        }
+
+        if (!UpdateVictim())
+            return;
+
+        DoMeleeAttackIfReady();
+    }
+private:
+    EventMap _events;
+    uint8 _PeasantId;
```
jackpoz, line 354: Please initialize this to 0 (even if it's done in Reset() also)

CraftedRO:
https://github.com/TrinityCore/TrinityCore/issues/16608 related ?

malcrom:
> #16608 related ?

Looks to be.

src/server/scripts/EasternKingdoms/zone_elwynn_forest.cpp
malcrom: done

src/server/scripts/EasternKingdoms/zone_elwynn_forest.cpp
```diff
@@ -275,7 +275,263 @@ struct npc_cameron : public ScriptedAI
...
+
+        if (uint32 eventId = _events.ExecuteEvent())
+        {
+            switch (eventId)
+            {
+                case EVENT_FIND_PEASANTS:
+                {
+                    GuidVector tempGUIDs;
+                    std::list<Creature*> peasants;
+                    GetCreatureListWithEntryInGrid(peasants, me, NPC_EASTVALE_PEASANT, 100.f);
```
Nyr97, line 331: you need to check whether container peasants is not empty, otherwise the worldserver will crash when you iterate through it and it's empty.

malcrom: AN if (peasants) for lines 332 to 341?

Nyr97: as Shauren pointed GetCreatureListWithEntryInGrid won't guarantee that the list you're providing below will be the same order you want it, for that purpose you'd need to sort it first, but, to do so, you'd need StringId, for example, which isn't ported from Master currently.

also, if GetCreatureListWithEntryInGrid fails to obtain the exact number you're referencing, in this case a container of 5 elements, the code will crash too sadly, as I mentioned earlier

malcrom: I see an issue with code. will work on it tonight.

malcrom:
I need to remove this and will push a new branch.

PR #30289, opened 2024-09-30
Scripts/Spells: Implement first aid related anti-venom item spells

### LINKED ISSUE #21518: [item/spell]
**Description:**
The item Anti-Venom (http://www.wowhead.com/item=6452/anti-venom) doesn't do anything.

**Current behaviour:**
If you are poisoned, it doesn't cure the poison.

**Expected behaviour:**
If you are poisoned, it should cure the poison.

1. Got to Mother Fang (http://www.wowhead.com/npc=471/mother-fang) and let her hit you.
2. Once poisoned, use the item.

Tested with a level 18 character with the appropriate First Aid skill. The poison (http://www.wowhead.com/spell=11918/poison) is level 20, the anti-venom should cure poisons up to level 25 (http://www.wowhead.com/spell=7932/anti-venom).

**Branch(es):** 3.3.5, master
**TC rev. hash/commit:** : c9b730c76707
**TDB version:** according to hash

Odyssey:
**Changes proposed:**

Makes the following first aid related items work:

* [Anti-Venom](https://www.wowhead.com/wotlk/item=6452/anti-venoml)
* [Strong Anti-Venom](https://www.wowhead.com/wotlk/item=6453/strong-anti-venom)
* [Powerful Anti-Venom](https://www.wowhead.com/wotlk/item=19440/powerful-anti-venom)

Note that WotlK era introduced a hard nerf to the anti-venom items, making them take in account the target's max level instead of the poison spell level, so players couldn't abuse the items to bypass higher level poison mechanics anymore. A conversation regarding this nerf and how the items behaved on higher level targets can be seen [here](https://www.wowhead.com/item=19440/powerful-anti-venom#comments:id=713457). This is reflected in the spell script.

While the PR targets 3.3.5, if someone seeing this takes this code to the master branch, the scriptname should also be assigned to the spell ids 172368 and 307165, from future expansions.

**Issues addressed:**

Closes #21518

**Tests performed:**

Builds & works ingame. Tested against the example in #21518 with one and multiple poison effects present.

### DIFF
--- a/sql/updates/world/3.3.5/XXXX_XX_XX_YY_world.sql
+++ b/sql/updates/world/3.3.5/XXXX_XX_XX_YY_world.sql
@@ -1,0 +1,6 @@
+-- Assign first aid anti-venom items spellscript
+DELETE FROM `spell_script_names` WHERE `ScriptName`='spell_item_antivenom';
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
+(7932, 'spell_item_antivenom'),
+(7933, 'spell_item_antivenom'),
+(23786, 'spell_item_antivenom');
--- a/src/server/scripts/Spells/spell_item.cpp
+++ b/src/server/scripts/Spells/spell_item.cpp
@@ -4356,13 +4356,15 @@ class spell_item_antivenom : public SpellScript
         if (Unit* target = GetHitUnit())
         {
             // Remove poison effects from the target up to X level
-            if (GetSpellInfo()->MaxLevel >= target->GetLevel())
+            uint32 maxLevel = GetSpellInfo()->MaxLevel;
+
+            target->RemoveAppliedAuras([&maxLevel](AuraApplication const* aurApp) -> bool
             {
-                target->RemoveAppliedAuras([](AuraApplication const* aurApp) -> bool
-                {
-                    return aurApp->GetBase()->GetSpellInfo()->Dispel == DISPEL_POISON;
-                });
-            }
+                Aura const* aura = aurApp->GetBase();
+                bool isPoison = aura->GetSpellInfo()->Dispel == DISPEL_POISON;
+                bool passesLevelCheck = aura->GetCasterLevel() <= maxLevel;
+                return isPoison && passesLevelCheck;
+            });
         }
     }
 
--- a/src/server/scripts/Spells/spell_item.cpp
+++ b/src/server/scripts/Spells/spell_item.cpp
@@ -4344,6 +4344,34 @@ class spell_item_titanium_seal_of_dalaran_catch : public SpellScript
     }
 };
 
+// 7932 - Anti-Venom
+// 7933 - Strong Anti-Venom
+// 23786 - Powerful Anti-Venom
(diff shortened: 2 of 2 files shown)


Ovahlord:
The question now is, if we want to go with the original balance of the game or if we really want to include classic changes

Aokromes:
if it's wotlk classic it must target wotlk classic branch, oh wait :P

Shauren:
Comments say its already nerfed in 3.1.1 (the old wotlk, not classic) and the script should be checking level of the caster of each poison aura

Odyssey:
Updated. Can quickly be tested engaging any hostile target and using .cast back 744 and .cast back 11918 or any other poison spells

Odyssey:
> the script should be checking level of the caster of each poison aura

I don't think that's true though? [This](https://www.wowhead.com/item=19440/powerful-anti-venom#comments:id=713457:reply=118683) mentions it not working against casters of level below 60.

From [here](https://www.wowhead.com/item=19440/powerful-anti-venom#comments:id=713457:reply=98510)
>  Now, it seems to check the level of the source target

>  the anti-venom won't work as long as the source target has higher level than 60

I think the source is clearly the anti-venom user, and the target is the player targetted by that anti-venom.

Odyssey:
Well, if the source is the poison source, the source target is still the player getting poisoned, which while not being an official statement, would line up with the guy saying it doesn't work in silithus against mobs under maxLevel.

CraftedRO:
> The question now is, if we want to go with the original balance of the game or if we really want to include classic changes

noone wants to include classic fun changes.

CraftedRO:
Just tested this PR, it works as intended.

Aokromes:
Conflicting files
src/server/scripts/Spells/spell_item.cpp

CraftedRO:
> Conflicting files src/server/scripts/Spells/spell_item.cpp

this is because in that place were merged the other titanium_seal_of_dalaran fix code.

CraftedRO (approved):

CraftedRO:
why does this still have wip label ?

anzz1:
Logically it makes sense that the change to the poison level would have been to use the **poison's caster** level as a drop-in change to the mechanic. It doesn't make any logical sense that you could heal yourself of high level poisons by the account of yourself being low level, and in turn that you couldn't heal yourself from low level poisons because you're higher level.

That being said, it's certainly possible that Blizz made a programming/logic error back in the day and checked against the level of the caster of the anti-venom spell, i.e. yourself, instead of the caster of the poison aura, i.e. the enemy that poisoned you, and it being unclear whether that error was fixed before/during the final 3.3.5 release.

Presumably nobody checked during WOTLK Classic how it worked there? (not that it's a 100% authority on 3.3.5 blizzlike behaviour being a backported newer engine)
Could someone check how the system currently works in Cata Classic?

My opinion is that due to 100% authoritative answer from 3.3.5 era is impossible to acquire unless someone finds some old video or has old sniffs of this particular mechanic, the most sense would be to implement it how it's logical, that the **poison's caster** level should be what determines whether the poison can be cured, not your own level.

PR #30293, opened 2024-10-02
Tools/vmap_assembler: Fix missing <mutex> header in TileAssembler.cpp

nekolisa42:
**Changes proposed:**

- Added mutex header to TileAssembler.cpp.
- Resolved build errors related to std::call_once at line 85 (now located at line 86) causing compilation issues (C3861, C2039).

**Tests performed:**

- Verified that the project successfully builds after adding the mutex header.
- No in-game testing required as this is a compilation fix.

### DIFF
--- a/src/tools/vmap4_assembler/TileAssembler.cpp
+++ b/src/tools/vmap4_assembler/TileAssembler.cpp
@@ -28,6 +28,7 @@
 #include <boost/filesystem/directory.hpp>
 #include <boost/filesystem/operations.hpp>
 #include <set>
+#include <mutex>
 
 template<> struct BoundsTrait<VMAP::ModelSpawn*>
 {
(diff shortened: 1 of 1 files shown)


Ovahlord:
what build environment do you use to reproduce the error? CI doesn't report any issues,

nekolisa42:
the build environment i’m using is as follows:

operating system: windows 11 23h2
compiler: visual studio 2022 enterprise version 17.5.0 (microsoft visual c++ 2022)
boost version: 1.78.0
cmake version: 3.30.4
active solution configuration: RelWithDebInfo

i encountered the error when building tileassembler.cpp in this environment, specifically related to the missing <mutex> header for std::call_once. the build process failed with the following errors:

C3861: 'call_once': identifier not found
C2039: 'call_once': is not a member of 'std'

adding the <mutex> header resolved the issue in my environment.

src/tools/vmap4_assembler/TileAssembler.cpp
```diff
@@ -28,6 +28,7 @@
 #include <boost/filesystem/directory.hpp>
 #include <boost/filesystem/operations.hpp>
 #include <set>
+#include <mutex>
```
Shauren, line 31: Can you move this just before <set>?

Shauren (approved):

PR #30343, opened 2024-10-12
Core/Texts: Use BroadCastText ConditionId

Golrag:
**Changes proposed:**

-  Add some extra fields to ChatPacketSender so that it can not only send texts but also handle sounds and emotes
-  ChatPacketSender will also take care of player conditions
-  Creature Text table that have a broadcast_text id filled in will use said condition id

**Issues addressed:**

Closes: none?

**Tests performed:**

- Builds locally
- Tested with my changes for Deephaul Ravine battleground. Almost all the texts have condition ids

- [ ] Code might need some clean up or refactor, feel free to suggest some things here


src/server/game/Texts/ChatTextBuilder.cpp
```diff
@@ -27,15 +27,57 @@
...
 
 void ChatPacketSender::operator()(Player const* player) const
 {
+    if (!player->MeetPlayerCondition(PlayerConditionID))
+        return;
+
+    if (SoundPacket)
+        player->SendDirectMessage(SoundPacket->GetRawPacket());
+    else if (ObjectSoundPacket)
+        player->SendDirectMessage(ObjectSoundPacket->GetRawPacket());
```
mdX7, line 76: We could merge `SoundPacket` and `ObjectSoundPacket` and use `ServerPacket` instead. Due to inheritance magic `GetRawPacket()` is valid for both and polymorphism does the rest

PR #30361, opened 2024-10-18
Core/GridNotifier: Allow more specific checks to include or exclude feign death units in IsAlive check

mdX7:
* this also extends SMART_TARGET_CLOSEST_CREATURE dead param

**Tests performed:**
tested ingame using all 4 types


src/server/game/Entities/Object/Object.h
```diff
@@ -452,12 +452,22 @@ class FlaggedValuesArray32
         T_FLAGS m_flags;
 };
 
+enum class FindCreatureAliveState : uint8
+{
+    Alive               = 0, // includes feign death
+    Dead                = 1, // includes feign death
+    EffectivelyAlive    = 2, // excludes feign death
+    EffectivelyDead     = 3, // excludes feign death
```
Shauren, line 460: I would have expected "EffectivelyDead" to mean "behaves like dead", which means `!IsAlive() || IsFeignDeath()`

Shauren (approved):

PR #30273, opened 2024-09-28
Respawn Aggro Delay

kaytotes:
**Changes proposed:**

-  Adds a 5 second grace period between when a creature spawns and when it can aggro on move in los.

**Tests performed:**

Active usage on our server since January 2023.

PR #30362, opened 2024-10-18
Core/Creature: Implement aggro grace period

mdX7:
**Issues addressed:**
ref https://github.com/TrinityCore/TrinityCore/discussions/26528
supercedes https://github.com/TrinityCore/TrinityCore/pull/30273

**Tests performed:**
creatures tested: 226909, 223799
tested db spawned & temp summoned creatures
- run into aggro range of creature after heartbeat triggered => aggro
- kill & respawn creature while camping at respawn spot (not moving) => aggro after grace period
- kill & respawn creature while camping at respawn spot (moving) => aggro after grace period
- kill & respawn creature while camping at respawn spot + attacking while grace period didnt expire yet => instant aggro


Ovahlord (approved):

Nyr97:
Does this force non-moving creatures that should be sparring with each other to aggro each other? When I was working on Silverpine, I had to force random movement from DB by 1.f for them to detect each other.

mdX7:
> Does this force non-moving creatures that should be sparring with each other to aggro each other? When I was working on Silverpine, I had to force random movement from DB by 1.f for them to detect each other. Asking so that I might remove that small hackfix I had to use back 2 years ago.

Yes, MoveInLineOfSight triggers the aggro behavior and is manually triggered after the grace period in this pr

PR #30369, opened 2024-10-21
Spells/Monk: Fix Rising Sun Kick heal reduction on Windwalker spec

aquadeus:
**Changes proposed:**

-  Fix Rising Sun Kick heal reduction on Windwalker spec

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -252,6 +254,30 @@ class spell_monk_provoke : public SpellScript
...
+
+        if (caster->HasAura(SPELL_MONK_COMBAT_CONDITIONING))
+            caster->CastSpell(GetHitUnit(), SPELL_MONK_MORTAL_WOUNDS, CastSpellExtraArgs()
+                .SetTriggeringSpell(GetSpell())
+                .SetTriggerFlags(TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR));
+    }
+
+    void Register() override
+    {
+        OnEffectHitTarget += SpellEffectFn(spell_monk_rising_sun_kick::HandleOnHit, EFFECT_0, SPELL_EFFECT_TRIGGER_SPELL);
```
mdX7, line 277: should it hit every target or just the primary target?

aquadeus: it has a 10 seconds recovery time, so you cant hit two targets at the same time

Shauren: cooldown doesnt matter (windwalker reduces it by pressing blackout kick) - aura is not unique, all targets hit get it

PR #30385, opened 2024-10-29
Spells/Scripts: Implement Leech Stat and SPELL_AURA_MOD_LEECH

aquadeus:
**Changes proposed:**

-  Implement Leech Stat and SPELL_AURA_MOD_LEECH

**Tests performed:**

Tested in-game


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -14107,6 +14109,28 @@ void Unit::SetVignette(uint32 vignetteId)
...
+    float tmpMod = 0.0f;
+
+    if (Player* player = ToPlayer())
+        tmpMod += player->m_unitData->Lifesteal;
+    else
+        tmpMod += GetTotalAuraModifier(SPELL_AURA_MOD_LEECH);
+
+    ApplyPct(amount, tmpMod);
+
+    HealInfo healInfo(this, this, amount, leechSpell, leechSpell->GetSchoolMask());
```
Nyr97, line 14127: Does this take into account that only effective healing should leech?

Nyr97:
Also, leech itself doesn't work on every instance of damage or healing occuring, it's based off GCD, it can be checked in simcraft (basically it stores the healing that should be rewarded and grant it every GCD).

You need an internal storage that updates every GCD to reward stored amount for this.

Issue #29903, opened 2024-04-08
Hunter/Spell: Flare shoud remove Stealth right away


CraftedRO:
### Description

Hunter Flare shoud remove Stealth right away while target is inside radius, atm it does nothing if casted on a stealthed target but only if target leave flare radius area and re-enter inside will be affected.

https://youtu.be/0mHpIu-5Aqo

Similar issue might be here https://github.com/TrinityCore/TrinityCore/issues/29605

### Expected behaviour

Flare shoud remove Stealth right away as you can see here (use slow playback speed): https://youtu.be/5Riwc5dJ1AE?t=30

### Steps to reproduce the problem

create a hunter and a rogue/druid
duel eachother
with rogue / druid enter Stealth / Prowl
with hunter place flare directly on rogue / druid
notice

### Branch

3.3.5

### TC rev. hash/commit

2ae8c0d32da8a5cb167de82e93c993a4f7145418

### Operating system

Windows 11 x64

### Custom changes

[Passive anticheat](https://github.com/TrinityCore/TrinityCoreCustomChanges/commits/3.3.5-passive_anticheat)

r0m1ntik:
Perhaps this commit broke https://github.com/TrinityCore/TrinityCore/commit/3306a4d06cc557967c23adcc60c2e3257811b324

CraftedRO:
[Possible fix](https://github.com/Project-Epoch/TrinityCore/commit/144769fceaf7985f2c4f96fae1b7785e65d857b6#diff-0d2a44370e18523042eca998013bf68fa21f38524f406c0ccdfd557f62342a30)

Issue #29605, opened 2024-01-22
Frost trap only slows on re-enter in PvP


thomasjteachey:
### Description

Hunter frost trap doesn't snare until someone ENTERS the radius from the outside, not when they trigger the trap.

During a duel, if someone triggers a trap, it doesn't slow them until they completely leave the radius and re-enter it.

### Expected behaviour

Frost trap should slow as soon as they enter

### Steps to reproduce the problem

1) Make a hunter (Player1)
2) Make any other character (Player2) and duel/fight the hunter
3) Hunter lays a frost trap
4) Player2 triggers the frost trap
5) Player2 is not slowed

### Branch

3.3.5

### TC rev. hash/commit

TrinityCore rev. 75016bf4da86 2024-01-22 13:58:31 +0100 (3.3.5 branch) (Win64, Debug, Static)

### Operating system

Windows 11

### Custom changes

None

PR #30422, opened 2024-11-11
Core/Spells: Allow persistent area auras to apply all of their effects upon creation.

CraftedRO:
**Changes proposed:**

- Cherrypick https://github.com/Project-Epoch/TrinityCore/commit/144769fceaf7985f2c4f96fae1b7785e65d857b6#diff-0d2a44370e18523042eca998013bf68fa21f38524f406c0ccdfd557f62342a30L1751
-  This fixes the issue where the area auras do not instantly apply their effects. Actually, they were only applying the last SPELL_EFFECT_PERSISTENT_AREA_AURA; Flare and Frost Trap have 2 and 3 SPELL_EFFECT_PERSISTENT_AREA_AURA effects respectively, so only the last one would be applied.
-  Credit goes to original authors.

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/29903

Closes https://github.com/TrinityCore/TrinityCore/issues/29605

**Tests performed:**

Does it build, tested in-game, as can be seen below:

https://youtu.be/8CgMXaAJAbg

- [ None ]

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -1601,7 +1601,7 @@ void Spell::EffectPersistentAA()
 
     ASSERT(_dynObjAura->GetDynobjOwner());
     for (size_t i = 0; i < m_spellInfo->GetEffects().size(); ++i)
-        if (!m_spellInfo->GetEffect(SpellEffIndex(i)).IsEffect(SPELL_EFFECT_NONE))
+        if (m_spellInfo->GetEffect(SpellEffIndex(i)).IsEffect(SPELL_EFFECT_PERSISTENT_AREA_AURA))
             _dynObjAura->_ApplyEffectForTargets(i);
 }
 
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/SpellEffects.cpp
```diff
@@ -1600,7 +1600,9 @@ void Spell::EffectPersistentAA()
         return;
 
     ASSERT(_dynObjAura->GetDynobjOwner());
-    _dynObjAura->_ApplyEffectForTargets(effectInfo->EffectIndex);
+    for (size_t i = 0; i < m_spellInfo->GetEffects().size(); ++i)
+        if (!m_spellInfo->GetEffect(SpellEffIndex(i)).IsEffect(SPELL_EFFECT_NONE))
```
Shauren, line 1604: You should only be calling _ApplyEffectForTargets for SPELL_EFFECT_PERSISTENT_AREA_AURA effects, not any non-empty effect

Shauren (changes requested):

src/server/game/Spells/SpellEffects.cpp
CraftedRO: like this ?

Shauren (approved):

PR #30427, opened 2024-11-12
Scripts/FoS: Ensure Bronjahm always teleports to the middle upon phase switch

### LINKED ISSUE #30421: sometimes bronjahm fails to teleport to center of room
### Description

sometimes bronjahm fails to teleport to center of room

### Expected behaviour

he must teleport to center

go forge of souls
fight bronjahm
sometimes he will fail to teleport to center on 2nd stage

### Branch

3.3.5

### TC rev. hash/commit

a377ce28c208

### Operating system

debian 12 x64

### Custom changes

n/r

Odyssey:
**Changes proposed:**

* Handle the phase switch on UpdateAI instead of DamageTaken, so the teleport is always issued after the current cast has finished, regardless if the threshold was hit while the boss was casting or not.
* Remove unneeded check because the DoMeleeAttackIfReady() below already checks for it

**Issues addressed:**

Closes #30421

**Tests performed:**

Builds & works ingame in both scenarios (health threshold reached while the boss is casting and while the boss isn't casting).

### DIFF
--- a/src/server/scripts/Northrend/FrozenHalls/ForgeOfSouls/boss_bronjahm.cpp
+++ b/src/server/scripts/Northrend/FrozenHalls/ForgeOfSouls/boss_bronjahm.cpp
@@ -109,17 +109,6 @@ struct boss_bronjahm : public BossAI
             Talk(SAY_SLAY);
     }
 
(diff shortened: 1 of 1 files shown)


Jildor:
Sorry, I took the liberty of creating a PR with a solution that I consider more appropriate. #30428

PR #30458, opened 2024-11-22
Core/Spells: Implemented priority-based target sorting via PriorityRules system

Nyr97:
Introduces a lightweight, flexible system to sort spell targets based on script-defined priority rules. This replaces hardcoded or purely random selection logic with a more precise, predictable approach using weighted scoring.

**Changes proposed:**

- Added PriorityRules structure to define scoring conditions with weights.
- Implemented SortTargetsWithPriorityRules():
1. Scores all eligible targets based on provided rules.
2. Sorts by descending score.
3. Shuffles tied entries around the cutoff to avoid deterministic bias.
- SortTargetsWithPriorityRules now requires an explicit rule set, improving clarity and avoiding hidden logic.

**Tests performed:**

- Verified Power Word: Radiance correctly prioritizes groupmates with no Atonement before other targets.
- Confirmed deterministic behavior no longer occurs on repeated casts with tied scores.

None.

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -9633,67 +9633,41 @@ void SelectRandomInjuredTargets(std::list<WorldObject*>& targets, size_t maxTarg
     std::ranges::transform(tempTargets.begin(), tempTargets.begin() + maxTargets, targets.begin(), Trinity::TupleElement<0>);
 }
 
-void SortTargetsWithPriorityRules(std::list<WorldObject*>& targets, size_t maxTargets, std::vector<PriorityRules> const& rules)
+void SortTargetsWithPriorityRules(std::list<WorldObject*>& targets, size_t maxTargets, std::span<TargetPriorityRule const> rules)
 {
     if (targets.size() <= maxTargets)
         return;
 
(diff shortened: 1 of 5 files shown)


Nyr97:
I'm thinking of just refactoring it entirely instead so that we can pass const std::vector<PriorityState>& priorityOrder instead and we define it somewhere in Spell.h, where PriorityState is also defined in there. 

We can always pass the same priority for all usages OR we can create a new priority list and include new ones if we see fit for the spell. Let me know what you think.

Nyr97:
not sure how to ask for a review so I'm pinging @Shauren or @mdX7

Shauren:
I don't like this

* Doesn't make sense as Unit members (its a spell func)
* Always selects the same targets on every cast (if their classification doesnt change between casts - sorting is now deterministic, first by priority then by the order they were added to grid)
* Uses std::function

Nyr97:
Moved it to Spell class inside Trinity namespace, made sorting less deterministic by randomizing equal-scored entries and restored old SelectRandomInjured stuff, also updated PR desc

Faq:
Perhaps when rewriting such system logic, could add some tests?

Nyr97:
Description clearly states I tested it on Radiance.

PR #30459, opened 2024-11-22
Scripts/StormwindCity: Implement quest "The King's Command" (29547)

meji46:
**Changes proposed:**

-  Implement quest The King's Command (29547) and alternative quests (49556, 60125)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game

None


mdX7 (approved):

src/server/scripts/EasternKingdoms/zone_stormwind_city.cpp
```diff
@@ -494,6 +504,58 @@ class conversation_quest_ancient_curses_accept : public ConversationScript
     EventMap _events;
 };
 
+// 130804 - The King's Command Movie Aura
+class spell_the_kings_command_movie_aura : public SpellScript
+{
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({
+            StormwindCity::Spells::FadeToBlack
```
Nyr97, line 513: maybe setting using namespace StormwindCity to reduce verbose?

meji46: But I would need to set it to every function that uses any of these variables. The other option is to add constexpr variables to the class

PR #30462, opened 2024-11-24
Scripts/Gundrak: Properly handle wipes on Drakkari Colossus encounter

Odyssey:
**Changes proposed:**

Rework encounter evade logic so that

-  If a wipe happens during a colossus stage, the colossus resets properly
-  If a wipe happens during an elemental stage, the elemental despawns and the colossus resets properly.

**Issues addressed:**

Closes #27811

**Tests performed:**

builds; tested and working in-game

### DIFF
--- a/src/server/scripts/Northrend/Gundrak/boss_drakkari_colossus.cpp
+++ b/src/server/scripts/Northrend/Gundrak/boss_drakkari_colossus.cpp
@@ -340,6 +340,13 @@ struct boss_drakkari_elemental : public ScriptedAI
         }
     }
 
+    void EnterEvadeMode(EvadeReason /*why*/) override
+    {
+        if (Creature* colossus = instance->GetCreature(DATA_DRAKKARI_COLOSSUS))
+            colossus->AI()->DoAction(ACTION_UNFREEZE_COLOSSUS);
+        me->DespawnOrUnsummon();
+    }
+
     void SpellHitTarget(WorldObject* target, SpellInfo const* spellInfo) override
     {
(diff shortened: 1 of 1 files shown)


Aokromes:
not perfect but far better than current status.

Aokromes:
looks good to me.

Aokromes (approved):
looks good to me

CraftedRO (approved):

src/server/scripts/Northrend/Gundrak/boss_drakkari_colossus.cpp
```diff
@@ -112,6 +113,20 @@ struct boss_drakkari_colossus : public BossAI
...
     }
 
+    void EnterEvadeMode(EvadeReason why) override
+    {
+        if (why == EVADE_REASON_NO_HOSTILES)
+        {
+            if (isFrozen)
+                return;
+
+            DoZoneInCombat();
```
Shauren, line 123: What is this supposed to do? Why are you forcing combat with zone on evade (!) when not frozen during elemental phase?

Odyssey: Very rarely while testing, I had a case that I couldn't really reproduce consistently where the boss would reset upon unfreezing instead of re-engaging player(s), that's why I moved it there

CraftedRO: so, did we get stucked here ?

PR #30477, opened 2024-12-01
Core/Units: Added helper methods to cancel mount/shapeshift auras

meji46:
**Changes proposed:**

-  Added helper methods to cancel mount/shapeshift auras (AURA_REMOVE_BY_INTERRUPT can be used in Unit::CancelShapeshiftForm to force a valid shapeshift instead of simply removing the current one) 
- Remove invalid auras in UpdateAreaDependentAuras using AURA_REMOVE_BY_INTERRUPT. This allows to force a valid shapeshift instead of simply removing the current one (entering a cave that only allows ground mounts in flight form should force ground shapeshift)
- Check valid mount shapeshift in Unit::UpdateMountCapability (using Unit::CancelShapeshiftForm) and update the shapeshift if necessary (e.g.  by applying SPELL_AURA_MOUNT_RESTRICTIONS)

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game
- SPELL_AURA_MOUNT_RESTRICTIONS: 421175 (Disable ALL Mounts)
- Cave that only allows ground mounts: Crumbling Depths in Deepholm 

- Log in while in flight form forces ground shapeshift. LearnDefaultSkills() is loaded after _LoadAuras() in Player::LoadFromDB so this check https://github.com/TrinityCore/TrinityCore/blob/master/src/server/scripts/Spells/spell_druid.cpp#L2110 fails. We can't call LearnDefaultSkills() before _LoadAuras() because LearnDefaultSkills() refreshes skill bonus auras

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -26024,7 +26024,7 @@ void Player::UpdateAreaDependentAuras(uint32 newArea)
     {
         // use m_zoneUpdateId for speed: UpdateArea called from UpdateZone or instead UpdateZone in both cases m_zoneUpdateId up-to-date
         if (iter->second->GetSpellInfo()->CheckLocation(GetMapId(), m_zoneUpdateId, newArea, this) != SPELL_CAST_OK)
-            RemoveOwnedAura(iter);
+            RemoveOwnedAura(iter, AURA_REMOVE_BY_INTERRUPT);
         else
             ++iter;
     }
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8229,6 +8229,11 @@ void Unit::UpdateMountCapability()
     if (IsLoading())
         return;
 
+    if (SpellShapeshiftFormEntry const* spellShapeshiftForm = sSpellShapeshiftFormStore.LookupEntry(GetShapeshiftForm()))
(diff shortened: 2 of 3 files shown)


Nyr97:
The to-do thing you mention requires implementing travel form (ID: 783) entirely which has quite a lot of ifs and buts being a dummy aura. There are many conditions which decide what form to use in Travel Form:

- Movementflag type (In water, submerged, etc.).
- Indoor state.
- Combat state.
- Riding ability.
- Allowed areas or related flags like AreaFlags::OnlyEvaluateGhostBindOnce.             
- Allowed zones, taking into account Pathfinding achievements or serverside quests.

meji46:
Yes, I am aware that the dummy aura is related to what I'm talking about

The current implementation addresses several of the conditions you mention in GetFormSpellId, but I know that it isn't complete. Feel free to open a PR given that you know all the conditions

Nyr97:
The problem is deciding when to hook this evaluation: since it's a dummy aura, we're to assume they either use OnUpdate with a very minimal tick interval or tick server, which sounds pretty exhausting in regards to resources or they use some type of movement update to do so. 

From my experience, I can tell that travel form sins the same way regular mounts do: when you get into a building, sometimes you may feel there's a delay until you get dismounted (or shapeshifted back), allowing you to fly indoors for a short time. This brings up the fact that Heartbeat may be involved at least for Stag/Flying when getting indoors.

mdX7 (approved):

PR #30481, opened 2024-12-04
Core/SAI: Allow update of disable gravity/root applied via static flags (using related action type)

meji46:
**Changes proposed:**

- Creatures that use CREATURE_STATIC_FLAG_FLOATING or CREATURE_STATIC_FLAG_SESSILE cannot enable gravity or be unrooted using SMART_ACTION_SET_DISABLE_GRAVITY or SMART_ACTION_SET_ROOT. Both actions currently use the underlying method of SetFloating/SetSessile which doesn't remove the related static flag and therefore has no effect. The actions have been modified to use SetFloating/SetSessile

**Issues addressed:**
None

**Tests performed:**
Builds, tested in-game

- As another option, we could manually apply disable gravity/root when spawning (SAI) instead of using static flags. That way static flags would be truly "static" and couldn't be toggled


src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -1361,10 +1361,9 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
         }
         case SMART_ACTION_SET_DISABLE_GRAVITY:
         {
-            if (!IsSmart())
-                break;
-
-            ENSURE_AI(SmartAI, me->AI())->SetDisableGravity(e.action.setDisableGravity.disable != 0);
+            for (WorldObject* target : targets)
```
Shauren, line 1364: Changes behavior

Does this not require sql updates to make existing actions target the same object?

meji46: All action targets were already 1 (except for one). Added sql for that case (target type 0) and custom cases

Shauren (approved):

PR #30490, opened 2024-12-06
Spells/DH: Fix talent "Fiery Brand"

aquadeus:
**Changes proposed:**

-  Fix talent "Fiery Brand"

**Tests performed:**

Tested in-game


Shauren (changes requested):
Unconditional cast is wrong, the DoT requires https://www.wowhead.com/spell=320962/fiery-brand

Shauren (approved):

PR #30501, opened 2024-12-11
Scripts/Spells: Fix demon hunter talent "Feast of Souls"

aquadeus:
**Changes proposed:**

-  Fix demon hunter talent "Feast of Souls"

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_dh.cpp
```diff
@@ -411,6 +411,33 @@ class spell_dh_eye_beam : public AuraScript
...
+    {
+        GetCaster()->CastSpell(GetCaster(), SPELL_DH_FEAST_OF_SOULS_PERIODIC_HEAL, CastSpellExtraArgsInit{
+            .TriggerFlags = TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR,
+            .TriggeringSpell = GetSpell()
+        });
+    }
+
+    void Register() override
+    {
+        AfterCast += SpellCastFn(spell_dh_feast_of_souls::HandleHeal);
```
Nyr97, line 437: Is it supposed to heal even though you may fail the spell when hitting the target with Soul Cleave? If that's how it works, this hook is fine; otherwise, OnEffectHitTarget is required.

Shauren: Soul Cleave... cleaves, cant be on hit target (its a cone target on triggered spell - its not even direct)

Nyr97: What if the target is a rogue during Evasion only? Is the cone supposed to hit even though the triggered spell is SPELL_DAMAGE_CLASS_MELEE with SPELL_SCHOOL_MASK_NORMAL?

PR #30503, opened 2024-12-12
Scripts/Spells: Fix demon hunter talent "Collective Anguish"

aquadeus:
**Changes proposed:**

-  Fix demon hunter talent "Collective Anguish"

**Tests performed:**

Tested in-game


Shauren:
This doesnt work (as expected) since both spells are channelled - needs core support

PR #30505, opened 2024-12-13
Core/Database: implement field accessing by string

Ovahlord:
**Changes proposed:**

-  implement new operator overload to allow us to directly access fields by their alias which helps managing large select queries without having to deal with index values

**Tests performed:**
- tested on cata_classic by using the provided skill loading changes

As of right now, using new benchmark results by using creature_template for mass load, I'm currently observing a ~5% performance regression compared to index based accessing which is within tolerable range

### DIFF
--- a/src/server/database/Database/QueryResult.cpp
+++ b/src/server/database/Database/QueryResult.cpp
@@ -405,7 +405,7 @@ _fields(fields)
     for (uint32 i = 0; i < _fieldCount; i++)
     {
         InitializeDatabaseFieldMetadata(&_fieldMetadata[i], &_fields[i], i, false);
-        bool success = _fieldIndexByAlias.try_emplace(_fieldMetadata[i].Alias, i).second;
+        bool success = _fieldIndexByAlias.try_emplace({ Trinity::DB::FieldLookupByAliasKey::RuntimeInit, _fieldMetadata[i].Alias }, i).second;
         ASSERT(success);
         _currentRow[i].SetMetadata(&_fieldMetadata[i]);
     }
@@ -463,7 +463,7 @@ m_metadataResult(result)
         rowSize += size;
 
         InitializeDatabaseFieldMetadata(&m_fieldMetadata[i], &field[i], i, true);
-        bool success = m_fieldIndexByAlias.try_emplace(m_fieldMetadata[i].Alias, i).second;
+        bool success = m_fieldIndexByAlias.try_emplace({ Trinity::DB::FieldLookupByAliasKey::RuntimeInit, m_fieldMetadata[i].Alias }, i).second;
         ASSERT(success);
 
         m_rBind[i].buffer_type = field[i].type;
@@ -637,7 +637,7 @@ Field const& ResultSet::operator[](std::size_t index) const
     return _currentRow[index];
 }
 
-Field const& ResultSet::operator[](std::string_view const& alias) const
+Field const& ResultSet::operator[](Trinity::DB::FieldLookupByAliasKey const& alias) const
 {
     auto itr = _fieldIndexByAlias.find(alias);
     ASSERT(itr != _fieldIndexByAlias.end());
@@ -663,7 +663,7 @@ Field const& PreparedResultSet::operator[](std::size_t index) const
     return m_rows[std::size_t(m_rowPosition) * m_fieldCount + index];
 }
 
-Field const& PreparedResultSet::operator[](std::string_view const& alias) const
+Field const& PreparedResultSet::operator[](Trinity::DB::FieldLookupByAliasKey const& alias) const
 {
     ASSERT(m_rowPosition < m_rowCount);
     auto itr = m_fieldIndexByAlias.find(alias);
--- a/src/server/database/Database/QueryResult.cpp
+++ b/src/server/database/Database/QueryResult.cpp
@@ -400,10 +400,13 @@ _result(result),
 _fields(fields)
 {
     _fieldMetadata.resize(_fieldCount);
+    _fieldIndexByAlias.reserve(_fieldCount);
     _currentRow = new Field[_fieldCount];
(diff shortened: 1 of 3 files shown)


MaxtorCoder:
Wouldn't this essentially cause optimization problems? As in queries taking a longer time to load?

Ovahlord:
> Wouldn't this essentially cause optimization problems? As in queries taking a longer time to load?

It's slower but it's not world-endingly bad and it makes maintaining large tables way more pleasant. As for asynchronous queries this shouldn't really matter if the query takes a couple milliseconds longer (if at all).

Lordron:
What about https://github.com/rbock/sqlpp11
https://github.com/Shauren/TrinityCore/commit/d295907b0b72848ae1eb2bcc2552a50cc578498a

This eventually will kill all indexes without having to maintain field names across all source base

src/server/database/Database/QueryResult.cpp
```diff
@@ -400,10 +400,13 @@ _result(result),
 _fields(fields)
 {
     _fieldMetadata.resize(_fieldCount);
+    _fieldIndexByAlias.reserve(_fieldCount);
     _currentRow = new Field[_fieldCount];
     for (uint32 i = 0; i < _fieldCount; i++)
     {
         InitializeDatabaseFieldMetadata(&_fieldMetadata[i], &_fields[i], i, false);
+        auto [_, success] = _fieldIndexByAlias.try_emplace(_fieldMetadata[i].Alias, i);
```
Shauren, line 408:
```suggestion
        bool success = _fieldIndexByAlias.try_emplace(_fieldMetadata[i].Alias, i).second;
```

src/server/database/Database/QueryResult.cpp
```diff
@@ -452,13 +455,16 @@ m_metadataResult(result)
...
     m_fieldMetadata.resize(m_fieldCount);
+    m_fieldIndexByAlias.reserve(m_fieldCount);
     std::size_t rowSize = 0;
     for (uint32 i = 0; i < m_fieldCount; ++i)
     {
         uint32 size = SizeForType(&field[i]);
         rowSize += size;
 
         InitializeDatabaseFieldMetadata(&m_fieldMetadata[i], &field[i], i, true);
+        auto [_, success] = m_fieldIndexByAlias.try_emplace(m_fieldMetadata[i].Alias, i);
```
Shauren, line 466:
```suggestion
        bool success = m_fieldIndexByAlias.try_emplace(m_fieldMetadata[i].Alias, i).second;
```

Lordron:
What if instead of this we will generate a simple view structure. Where each accessor is mapped 1:1 to db. It will required additional aliases with join queries.
So in given example we will produce a struct like this

```
struct CharacterSkillsSelect
{
    Field* _field;
  
    CharacterSkillsSelect(Field* field) : _field(field) { }
    
    uint16 Skill() const { return _field[0].GetUInt16(); }
    uint16 Value() const { return _field[1].GetUInt16(); }
    uint16 Max() const { return _field[2].GetUInt16(); }
    int8 ProfessionSlot() const { return _field[3].GetInt8(); }
};
```
Pros: we do not have to maintain name, every changed name, removed field will lead to compilation error. 
no more wrong accessors (mean GetInt8 on UInt32 field etc). All of this become compile time error\warning.
Cons: it may still make compilation slower?

MaxtorCoder:
> What if instead of this we will generate a simple view structure. Where each accessor is mapped 1:1 to db. It will required additional aliases with join queries.
> 
> So in given example we will produce a struct like this
> 
> 
> 
> ```
> 
> struct CharacterSkillsSelect
> 
> {
> 
>     Field* _field;
> 
>   
> 
>     CharacterSkillsSelect(Field* field) : _field(field) { }
> 
>     
> 
>     uint16 Skill() const { return _field[0].GetUInt16(); }
> 
>     uint16 Value() const { return _field[1].GetUInt16(); }
> 
>     uint16 Max() const { return _field[2].GetUInt16(); }
> 
>     int8 ProfessionSlot() const { return _field[3].GetInt8(); }
> 
> };
> 
> ```
> 
> Pros: we do not have to maintain name, every changed name, removed field will lead to compilation error. 
> 
> no more wrong accessors (mean GetInt8 on UInt32 field etc). All of this become compile time error\warning.
> 
> Cons: it may still make compilation slower?
> 
> 

That is quite literally what Shauren abandoned before as an idea. So no.

Lordron:
His implementation is much complex and based on sqlcpp. With a log of generic code and inheritance https://github.com/Shauren/TrinityCore/blob/07e3548bd94ab08f312b17f44c42fd094e06e688/src/server/database/Database/Implementation/HotfixDatabaseTypes.h

> > What if instead of this we will generate a simple view structure. Where each accessor is mapped 1:1 to db. It will required additional aliases with join queries.
> > So in given example we will produce a struct like this
> > ```
> > 
> > struct CharacterSkillsSelect
> > 
> > {
> > 
> >     Field* _field;
> > 
> >   
> > 
> >     CharacterSkillsSelect(Field* field) : _field(field) { }
> > 
> >     
> > 
> >     uint16 Skill() const { return _field[0].GetUInt16(); }
> > 
> >     uint16 Value() const { return _field[1].GetUInt16(); }
> > 
> >     uint16 Max() const { return _field[2].GetUInt16(); }
> > 
> >     int8 ProfessionSlot() const { return _field[3].GetInt8(); }
> > 
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Pros: we do not have to maintain name, every changed name, removed field will lead to compilation error.
> > no more wrong accessors (mean GetInt8 on UInt32 field etc). All of this become compile time error\warning.
> > Cons: it may still make compilation slower?
> 
> That is quite literally what Shauren abandoned before as an idea. So no.

PR #30531, opened 2024-12-19
Spells/DH: Fix talent "Shattered Destiny"

aquadeus:
**Changes proposed:**

-  Fix talent "Shattered Destiny"

**Tests performed:**

Tested in-game


aquadeus:
oh just realize is for every 12 fury points not  power cost of 12 fury points, will fix it tomorrow

PR #30538, opened 2024-12-24
Core/AI: Implemented conversation ai

mdX7:
* allows having one conversation ai object per conversation instead of one big shared script

**Tests performed:**
tested using aberrus convos & Get Your Bearings boralus convos

### DIFF
--- a/src/server/game/AI/CoreAI/ConversationAI.h
+++ b/src/server/game/AI/CoreAI/ConversationAI.h
@@ -64,7 +64,7 @@ class TC_GAME_API ConversationAI
         uint32 GetId() const { return _scriptId; }
 };
 
-class NullConversationAI : public ConversationAI
+class NullConversationAI final : public ConversationAI
 {
     public:
         using ConversationAI::ConversationAI;
--- a/src/server/game/AI/CoreAI/ConversationAI.h
+++ b/src/server/game/AI/CoreAI/ConversationAI.h
@@ -1,0 +1,73 @@
+/*
+ * This file is part of the TrinityCore Project. See AUTHORS file for Copyright information
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef TRINITY_CONVERSATIONAI_H
+#define TRINITY_CONVERSATIONAI_H
+
+#include "Define.h"
+#include "ObjectGuid.h"
+
+class Conversation;
+class Unit;
+class Player;
+
+class TC_GAME_API ConversationAI
+{
+        uint32 _scriptId;
+
+    protected:
+        Conversation* const convo;
+    public:
+        explicit ConversationAI(Conversation* c, uint32 scriptId = {});
+        virtual ~ConversationAI();
+
+        // Called when the Conversation has just been initialized, just before added to map
+        virtual void OnInitialize() {}
+
+        // Called when Conversation is created but not added to Map yet.
+        virtual void OnCreate([[maybe_unused]] Unit* creator) { }
+
+        // Called when Conversation is started
+        virtual void OnStart() { }
+
+        // Called when player sends CMSG_CONVERSATION_LINE_STARTED with valid conversation guid
+        virtual void OnLineStarted([[maybe_unused]] uint32 lineId, [[maybe_unused]] Player* sender) { }
+
+        // Called for each update tick
+        virtual void OnUpdate([[maybe_unused]] uint32 diff) { }
+
+        // Called when the Conversation is removed
+        virtual void OnRemove() { }
+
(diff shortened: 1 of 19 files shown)


src/server/game/AI/CoreAI/ConversationAI.h
```diff
@@ -0,0 +1,73 @@
...
+#include "Define.h"
+#include "ObjectGuid.h"
+
+class Conversation;
+class Unit;
+class Player;
+
+class TC_GAME_API ConversationAI
+{
+        uint32 _scriptId;
```
Nyr97, line 30: _scriptId is a private member, but convo is protected, do we not want to make convo private as well and ensure people use getters to access it?

uint32 GetScriptId() const { return _scriptId; } (rather than GetId only)
Conversation* GetConversation() const { return convo; }

mdX7: thats intended to enable `convo` usage similar to `me` or `at` in CreatureAI/AreaTriggerAI

src/server/game/AI/CoreAI/ConversationAI.h
```diff
@@ -0,0 +1,73 @@
...
+
+    protected:
+        Conversation* const convo;
+    public:
+        explicit ConversationAI(Conversation* c, uint32 scriptId = {});
+        virtual ~ConversationAI();
+
+        // Called when the Conversation has just been initialized, just before added to map
+        virtual void OnInitialize() {}
```
Nyr97, line 39: instead of using empty functions, we can mark them in the base class with ```cpp = default``` where desired, such as OnInitialize, OnCreate, and OnStart.

src/server/game/AI/CoreAI/ConversationAI.h
```diff
@@ -0,0 +1,73 @@
...
+class Player;
+
+class TC_GAME_API ConversationAI
+{
+        uint32 _scriptId;
+
+    protected:
+        Conversation* const convo;
+    public:
+        explicit ConversationAI(Conversation* c, uint32 scriptId = {});
```
Nyr97, line 35: we can also make scriptId an optional being defaulted to nullopt rather than handling the default value maybe

src/server/game/AI/CoreAI/ConversationAI.h
```diff
@@ -0,0 +1,73 @@
...
+        virtual uint32 GetData([[maybe_unused]] uint32 id = 0) const { return 0; }
+        virtual void SetData([[maybe_unused]] uint32 id, [[maybe_unused]] uint32 value) { }
+        virtual void SetGUID([[maybe_unused]] ObjectGuid const& guid, [[maybe_unused]] int32 id = 0) { }
+        virtual ObjectGuid GetGUID([[maybe_unused]] int32 id = 0) const { return ObjectGuid::Empty; }
+
+        // Gets the id of the AI (script id)
+        uint32 GetId() const { return _scriptId; }
+};
+
+class NullConversationAI : public ConversationAI
```
Nyr97, line 67: we could definitely set NullConversationAI as 

```cpp
final
```
since we won't be subclassing it any longer, it's pretty clear what the intention for this class is.

src/server/game/AI/CoreAI/ConversationAI.h
Shauren: What? No, only "special members" such as contructors/destructor/assignment operator/operator=/operator<=> can be defaulted

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -1405,6 +1395,15 @@ class GenericAreaTriggerEntityScript : public AreaTriggerEntityScript
 };
 #define RegisterAreaTriggerAI(ai_name) new GenericAreaTriggerEntityScript<ai_name>(#ai_name)
 
+template <class AI>
+class GenericConversationScript : public ConversationScript
+{
+public:
+    GenericConversationScript(char const* name) : ConversationScript(name) {}
+    ConversationAI* GetAI(Conversation* convo) const override { return new AI(convo); }
```
Nyr97, line 1403: at some point, we might just want to stop relying on raw pointer returning for AI scripts and favour std::unique_ptr to ensure automatic memory clean-up when it goes out of scope which would reduce and *mostly* void memory leaks:

```cpp
    std::unique_ptr<ConversationAI> GetAI(Conversation* convo) const override
    {
        return std::make_unique<AI>(convo);
    }
```

src/server/game/AI/CoreAI/ConversationAI.h
Nyr97: oh, I c

src/server/game/Scripting/ScriptMgr.h
mdX7: it is used as a unique_ptr, check `_ai`

src/server/game/AI/CoreAI/ConversationAI.h
Nyr97: as far as I know, virtual functions can be set as default if no behavior is required, unless overriding is expected

src/server/game/AI/CoreAI/ConversationAI.h
mdX7: given std::optional has a bit of overhead and we dont really need scriptid 0 to be valid i'm keeping it as is (just like it is for all the other ais)

src/server/game/Scripting/ScriptMgr.h
Nyr97: oh, I'm late to the party 😞

src/server/game/AI/CoreAI/ConversationAI.h
mdX7: https://en.cppreference.com/w/cpp/language/member_functions#Special_member_functions

PR #30551, opened 2024-12-27
Scripts/Spells: Implement demon hunter talent "Glaive Tempest"

aquadeus:
**Changes proposed:**

-  Implement demon hunter talent "Glaive Tempest" (areatrigger)

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_dh.cpp
```diff
@@ -1039,6 +1041,35 @@ class spell_dh_vengeful_retreat_damage : public SpellScript
...
+struct at_dh_glaive_tempest : AreaTriggerAI
+{
+    at_dh_glaive_tempest(AreaTrigger* areatrigger) : AreaTriggerAI(areatrigger) { }
+
+    void OnCreate(Spell const* /*creatingSpell*/) override
+    {
+        if (Unit* caster = at->GetCaster())
+            caster->CastSpell(at->GetPosition(), SPELL_DH_GLAIVE_TEMPEST, CastSpellExtraArgs(TriggerCastFlags(TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR)));
+
+        _scheduler.Schedule(200ms, 250ms, [this](TaskContext task)
```
mdX7, line 1055: at spell has SPELL_ATTR8_HASTE_AFFECTS_DURATION, that means more haste => less duration. With the current approach that'd be a disadvantage. Count the amount of ticks done and then use at max duration/<max ticks> as tickrate

Shauren: This spell ticks 7 (* 2) times, with a tick on cast (SPELL_ATTR5_EXTRA_INITIAL_PERIOD) and 500ms period (affected by melee haste)
Duration is snapshotted on cast
Haste is recalculated every tick (might cause you to gain/lose ticks!)
Every tick cast the spell twice (simulating separate main/off hand hits, no special handling for that required, just cast twice)

see 55dda32225a653d1ecdd44d88e8566f8dd2f4121 for haste handling (except use ModHaste instead of ModCastingSpeed for melee haste)

Shauren (changes requested):

Shauren (approved):

PR #30556, opened 2024-12-30
Core/DBCEnums: Update CurrencyConsts, Difficulty and DifficultyFlags

osleyder85:
**Changes proposed:**

-  Update CurrencyConsts
-  Update Difficulty
-  Update DifficultyFlags


src/server/game/DataStores/DBCEnums.h
```diff
@@ -916,56 +975,70 @@ enum class CurveInterpolationMode : uint8
...
-    DIFFICULTY_VISIONS_OF_NZOTH     = 152,
-    DIFFICULTY_TEEMING_ISLAND       = 153
+    DIFFICULTY_NONE                         = 0,
+    DIFFICULTY_NORMAL                       = 1,    // 1 (Party Dungeon),   MinPlayers:  5, MaxPlayers:  5
+    DIFFICULTY_HEROIC                       = 2,    // 1 (Party Dungeon),   MinPlayers:  5, MaxPlayers:  5
+    DIFFICULTY_10_NORMAL                    = 3,    // 2 (Raid Dungeon),    MinPlayers: 10, MaxPlayers: 10
+    DIFFICULTY_25_NORMAL                    = 4,    // 2 (Raid Dungeon),    MinPlayers: 25, MaxPlayers: 25
+    DIFFICULTY_10_HEROIC                    = 5,    // 2 (Raid Dungeon),    MinPlayers: 10, MaxPlayers: 10
+    DIFFICULTY_25_HEROIC                    = 6,    // 2 (Raid Dungeon),    MinPlayers: 25, MaxPlayers: 25
+    DIFFICULTY_LFR                          = 7,    // 2 (Raid Dungeon),    MinPlayers: 25, MaxPlayers: 25
```
Nyr97, line 985: at this point, I would rename DIFFICULTY_LFR to DIFFICULTY_LFR_OLD, since it was the one used for Flexible raid when it was initially introduced in 5.4, and DIFFICULTY_LFR_NEW to DIFFICULTY_LFR since it's the standard

Ovahlord: this enum has been pulled from interface code.

osleyder85: > this enum has been pulled from interface code.

https://wago.tools/db2/Difficulty

Nyr97: > this enum has been pulled from interface code.

more like db2, either way, both are called Looking for Raid, so it's more than reasonable to rename them as I indicated

src/server/game/DataStores/DBCEnums.h
```diff
@@ -839,10 +839,69 @@ DEFINE_ENUM_FLAG(ChrCustomizationReqFlag);
...
+    CURRENCY_TYPE_ASPECTS_DREAMING          = 2718, // Earned from Mythic difficulty Amirdrassil and Mythic Keystone dungeons from +16 and up during Dragonflight Season 3.
+    CURRENCY_TYPE_AZEROTHIAN_ARCHIVES       = 2719, // Progress with the Azerothian Archivists.
+    CURRENCY_TYPE_COUNCIL_OF_DORNOGAL       = 2897, // Grants Reputation with the Council of Dornogal.
+    CURRENCY_TYPE_HALLOWFALL_ARATHI         = 2899, // Grants Reputation with the Hallowfall Arathi.
+    CURRENCY_TYPE_THE_ASSEMBLY_OF_THE_DEEPS = 2902, // Grants Reputation with The Assembly of the Deeps
+    CURRENCY_TYPE_THE_SEVERED_THREADS       = 2903, // Grants Reputation with The Severed Threads.
+    CURRENCY_TYPE_THE_WEAVER_NOTORIETY      = 3002, // Grants Reputation with The Weaver.
+    CURRENCY_TYPE_THE_GENERAL_NOTORIETY     = 3003, // Grants Reputation with The General.
+    CURRENCY_TYPE_THE_VIZIER_NOTORIETY      = 3004, // Grants Reputation with The Vizier.
+    CURRENCY_MAX_CAP_ANCIENT_MANA           = 4000
```
meji46, line 904: Is the cap change intended?

osleyder85: > Is the cap change intended?

I set that value because I thought it should be higher than the last one. What value do you recommend?

osleyder85: > Is the cap change intended?

Do I keep the value I had?

MaxtorCoder: MaxQty is 300, so idk why it was set to 2k to begin with, unless there was another reason for that value?

meji46: > MaxQty is 300, so idk why it was set to 2k to begin with, unless there was another reason for that value?

That's default max quantity, you can increase the cap to 2000

meji46: > > Is the cap change intended?
> 
> Do I keep the value I had?

Yes, that value isn't a currency id, is the max quantity for ancient mana (after increasing cap)

Nyr97: In that case it doesn't belong in there, that enum is for types only.

osleyder85: > Yes, that value isn't a currency id, is the max quantity for ancient mana (after increasing cap)

Yes, right, I understood that and I'm updating.

meji46: Your last change is wrong. Just keep CURRENCY_MAX_CAP_ANCIENT_MANA as it was defined (2000) and remove the rest of the caps (Player::GetCurrencyMaxQuantity takes into account required db2 fields)

MaxtorCoder: You really did not have go add the max caps for every currency type either as that can simply be retrieved from DB2...

Shauren:
This is very unneccessary, we don't need all the currency ids as enum in code

PR #30571, opened 2025-01-03
Scripts/Priest: Update Shadow Covenant

Nyr97:
**Changes proposed:**

Updated Shadow Covenant to 11.0.7 patch.

**Issues addressed:**

None.

**Tests performed:**

Tested.

None.


src/server/scripts/Spells/spell_priest.cpp
```diff
@@ -2968,31 +2970,33 @@ class spell_pri_spirit_of_redemption : public AuraScript
...
+        if (!GetCaster()->HasAura(SPELL_PRIEST_SHADOW_COVENANT))
+            return false;
 
-        Trinity::SelectRandomInjuredTargets(targets, maxTargets, true);
+        return true;
+    }
 
-        if (Unit* explicitTarget = GetExplTargetUnit())
-            targets.push_front(explicitTarget);
+    void TriggerShadowCovenant() const
```
mdX7, line 2989: Would suggest renaming it in a way so it reflects the used hook, like: `HandleAfterCast`

Shauren (changes requested):
But this is wrong? It should be shadowfield/mindbender

Nyr97:
Not only did I attach it to the wrong spell, but I noticed that Voidwrath also benefits from it, apart from Mindbender and Shadowfiend.

PR #30575, opened 2025-01-03
Scripts/Spells: Implement evoker talent "Ancient Flame"

aquadeus:
**Changes proposed:**

-  Implement evoker talent "Ancient Flame"

**Tests performed:**

Tested in-game


mdX7 (commented):
set Charges = 1 in `spell_proc`, that way we shouldn't need the script

aquadeus:
the spell that must be removed has no proc

PR #30578, opened 2025-01-04
Scripts/Spells: Implement evoker talents "Ruby Essence Burst" and "Azure Essence Burst"

aquadeus:
**Changes proposed:**

-  Implement evoker talent "Ruby Essence Burst" and "Azure Essence Burst"

**Tests performed:**

Tested in-game


Nyr97:
Since both are using the same effect, and many other spells as well, why not just create a namespace with helpers to return the float chance directly from the effect?

PR #30589, opened 2025-01-08
Spells/Mage: Update ice barrier absorb formula

aquadeus:
**Changes proposed:**

-  Update ice barrier absorb formula

**Tests performed:**

Tested in-game


Shauren:
canBeRecalculated applies only to spellsteal/load from db/applying spellmod and should be set to false here (as the old script version), it does not prevent new calc on refresh

aquadeus:
thanks for the advice

PR #30602, opened 2025-01-15
Core/Creature: implement initial SummonInfo API class 

Ovahlord:
This new class will be the new centralized API for summon behavior. The plan is to eventually migrate/correct/fix the entire summons implementation over into that API to have it nicely organized and refactored in a way that it fixes fundamental design issues, such as taming hunter pets being unable to get converted from open world spawns.

The approach will be to migrate functions over one by one to avoid large regressions or creating new issues.

**Changes proposed:**

-  Added SummonInfo class and handle its initialization
-  Changed the Unit::IsSummon helper to check for an existing SummonInfo ptr instead of relying on unit_masks (which will get removed eventually as well)

**Tests performed:**
- tested ingame, no behavior changes observed

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -49,6 +49,7 @@
 #include "Spell.h"
 #include "SpellAuraEffects.h"
 #include "SpellMgr.h"
+#include "SummonInfo.h"
 #include "TemporarySummon.h"
 #include "Vehicle.h"
 #include "World.h"
@@ -3606,6 +3607,21 @@ void Creature::InitializeInteractSpellId()
         SetInteractSpellId(clickBounds.begin()->second.spellId);
     else
         SetInteractSpellId(0);
+}
+
+void Creature::InitializeSummonInfo()
+{
+    _summonInfo = std::make_unique<SummonInfo>(this);
+}
+
+SummonInfo* Creature::GetSummonInfo() const
+{
+    return _summonInfo.get();
+}
+
+bool Creature::IsSummon() const
+{
+    return _summonInfo != nullptr;
 }
 
 void Creature::BuildValuesCreate(ByteBuffer* data, UF::UpdateFieldFlag flags, Player const* target) const
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -355,6 +355,11 @@ void Creature::AddToWorld()
         if (IsVehicle())
             GetVehicleKit()->Install();
 
+        // If the creature has been summoned, register it for the summoner
+        if (IsSummon())
+            if (Unit* summoner = GetSummonInfo()->GetUnitSummoner())
+                summoner->RegisterSummon(GetSummonInfo());
+
         if (GetZoneScript())
             GetZoneScript()->OnCreatureCreate(this);
     }
@@ -364,6 +369,11 @@ void Creature::RemoveFromWorld()
 {
     if (IsInWorld())
     {
+        // If the creature about to despawn, unregister it for the summoner
+        if (IsSummon())
+            if (Unit* summoner = GetSummonInfo()->GetUnitSummoner())
+                summoner->UnregisterSummon(GetSummonInfo());
+
         if (GetZoneScript())
             GetZoneScript()->OnCreatureRemove(this);
 
@@ -3609,10 +3619,10 @@ void Creature::InitializeInteractSpellId()
         SetInteractSpellId(0);
 }
 
-void Creature::InitializeSummonInfo()
+void Creature::InitializeSummonInfo(SummonInfoArgs const& args)
 {
     ASSERT(_summonInfo == nullptr, "SummonInfo has already been initialized and must not be allocated again!");
-    _summonInfo = std::make_unique<SummonInfo>(this);
+    _summonInfo = std::make_unique<SummonInfo>(this, args);
 }
 
 SummonInfo* Creature::GetSummonInfo() const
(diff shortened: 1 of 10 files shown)


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -3608,6 +3609,21 @@ void Creature::InitializeInteractSpellId()
         SetInteractSpellId(0);
 }
 
+void Creature::InitializeSummonInfo()
+{
+    _summonInfo = std::make_unique<SummonInfo>(this);
+}
+
+SummonInfo* Creature::GetSummonInfo() const
```
Nyr97, line 3617: isn't it best to use std::optional for safer handling?

`
std::optional<SummonInfo> GetSummonInfo(uint32 summonId)
`

however, what I had in mind it would be for _summonInfo to be a std::unordered_map for faster lookup by Id

src/server/game/Entities/Creature/Creature.h
```diff
@@ -579,6 +587,9 @@ class TC_GAME_API Creature : public Unit, public GridObject<Creature>, public Ma
         uint32 _gossipMenuId;
         Optional<uint32> _trainerId;
         float _sparringHealthPct;
+
+        // The Summon API which controls the behavior of summons
+        std::unique_ptr<SummonInfo> _summonInfo;
```
Nyr97, line 592: is there any case at all where we would be fine at allowing multiple ownership at some point due to scripting reasons for this to be accessible? just asking in case it'd be convenient to switch to shared_ptr instead for the future

src/server/game/Entities/Creature/SummonInfo/SummonInfo.h
```diff
@@ -0,0 +1,39 @@
...
+
+class TC_GAME_API SummonInfo
+{
+public:
+    SummonInfo(SummonInfo const& right) = delete;
+    SummonInfo(SummonInfo&& right) = delete;
+    SummonInfo& operator=(SummonInfo const& right) = delete;
+    SummonInfo& operator=(SummonInfo&& right) = delete;
+
+    SummonInfo(Creature* summonedCreature);
```
Nyr97, line 33: I would mark this as explicit

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -3608,6 +3609,21 @@ void Creature::InitializeInteractSpellId()
         SetInteractSpellId(0);
 }
 
+void Creature::InitializeSummonInfo()
+{
+    _summonInfo = std::make_unique<SummonInfo>(this);
```
Nyr97, line 3614: shouldn't be required, because it's assumed that the void will be called just once, but shouldn't we check for it being null?

`
if (!_summonInfo)
        _summonInfo = std::make_unique<SummonInfo>(this);
`

Ovahlord: I would rather throw an assert instead to directly tell the user that this method is not meant to be called twice.

Nyr97: yeah that would be nice as well, if filled, an assert instead

src/server/game/Entities/Creature/Creature.h
Ovahlord: There is no such thing such as multiple ownership. You call the spell effect, you summon the creature, you are the summoner. If someone were to take the summon away from you, you would charm that creature (which is a whole different mechanic).

As for the optional --> I don't want to bloat the memory of the creature object any further and keep the summon API nicely contained in its own cozy space on the heap. By using a pointer, we just use additional 64 bit for the creature instead of shoving in the whole size of the API

src/server/game/Entities/Creature/SummonInfo/SummonInfo.h
Ovahlord: Not needed as this ctor will soon get more mandatory parameters which will disable the implicit conversion. And since we allocate the SummonInfo object from inside the creature object with a self reference, there is no way a conversion error can occur

Shauren: explicit applies to any number of args since `return { arg1, arg2, arg3 }` became a thing

src/server/game/Entities/Creature/SummonInfo/SummonInfoArgs.h
```diff
@@ -0,0 +1,36 @@
...
+#define _SummonInfoArgs_h__
+
+#include "define.h"
+#include "Duration.h"
+#include "Optional.h"
+#include "ObjectGuid.h"
+
+struct SummonPropertiesEntry;
+
+struct TC_GAME_API SummonInfoArgs
```
Shauren, line 28: _API definition not needed on structures that don't have member functions not visible in header

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9143,6 +9144,17 @@ uint32 Unit::GetCreatureTypeMask() const
     return (creatureType >= 1) ? (1 << (creatureType - 1)) : 0;
 }
 
+void Unit::RegisterSummon(SummonInfo* summon)
+{
+    if (std::find(_activeSummons.begin(), _activeSummons.end(), summon) != _activeSummons.end())
```
Shauren, line 9149: `advstd::ranges::contains(_activeSummons, summon)`

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -9143,6 +9144,17 @@ uint32 Unit::GetCreatureTypeMask() const
...
+        _activeSummons.push_back(summon);
+}
+
+void Unit::UnregisterSummon(SummonInfo* summon)
+{
+    std::erase_if(_activeSummons, [summon](SummonInfo const* activeSummon) { return activeSummon == summon; });
```
Shauren, line 9155: `std::erase(_activeSummons, summon);`

PR #30605, opened 2025-01-17
Core/Creature: implement initial SummonInfo API for handling summons in the future

Ovahlord:
New attempt:

**Changes proposed:**

-  added a new SummonInfo API which will serve as future place to handle summons in all kinds of way
-  for now it only supports accessing summoners. The API will get expanded as the implementation goes on. We will add more step by step to make sure we won't break anything/too much at once.
-  

**Tests performed:**
- tested ingame

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -49,6 +49,7 @@
 #include "Spell.h"
 #include "SpellAuraEffects.h"
 #include "SpellMgr.h"
+#include "SummonInfo.h"
 #include "TemporarySummon.h"
 #include "Vehicle.h"
 #include "World.h"
@@ -354,6 +355,11 @@ void Creature::AddToWorld()
         if (IsVehicle())
             GetVehicleKit()->Install();
 
+        // If the creature has been summoned, register it for the summoner
+        if (IsSummon())
+            if (Unit* summoner = GetSummonInfo()->GetUnitSummoner())
+                summoner->RegisterSummon(GetSummonInfo());
+
         if (GetZoneScript())
             GetZoneScript()->OnCreatureCreate(this);
     }
@@ -363,6 +369,11 @@ void Creature::RemoveFromWorld()
 {
     if (IsInWorld())
     {
+        // If the creature about to despawn, unregister it for the summoner
+        if (IsSummon())
+            if (Unit* summoner = GetSummonInfo()->GetUnitSummoner())
+                summoner->UnregisterSummon(GetSummonInfo());
+
         if (GetZoneScript())
             GetZoneScript()->OnCreatureRemove(this);
 
@@ -3606,6 +3617,22 @@ void Creature::InitializeInteractSpellId()
         SetInteractSpellId(clickBounds.begin()->second.spellId);
     else
         SetInteractSpellId(0);
+}
+
+void Creature::InitializeSummonInfo(SummonInfoArgs const& args)
+{
+    ASSERT(_summonInfo == nullptr, "SummonInfo has already been initialized and must not be allocated again!");
+    _summonInfo = std::make_unique<SummonInfo>(this, args);
+}
+
+SummonInfo* Creature::GetSummonInfo() const
+{
+    return _summonInfo.get();
+}
+
+bool Creature::IsSummon() const
+{
+    return _summonInfo != nullptr;
 }
 
 void Creature::BuildValuesCreate(ByteBuffer* data, UF::UpdateFieldFlag flags, Player const* target) const
--- a/src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
+++ b/src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
@@ -23,8 +23,8 @@
 #include "SummonInfoArgs.h"
 
 SummonInfo::SummonInfo(Creature* summonedCreature, SummonInfoArgs const& args) :
-    _summonedCreature(ASSERT_NOTNULL(summonedCreature)), _despawnOnSummonerLogout(false), _despawnOnSummonerDeath(false), _despawnWhenExpired(false),
-    _summonerGUID(args.SummonerGUID), _remainingDuration(args.Duration), _maxHealth(args.MaxHealth), _summonSpellId(args.SummonSpellId)
+    _summonedCreature(ASSERT_NOTNULL(summonedCreature)), _summonerGUID(args.SummonerGUID), _remainingDuration(args.Duration), _maxHealth(args.MaxHealth), _summonSpellId(args.SummonSpellId),
+    _despawnOnSummonerLogout(false), _despawnOnSummonerDeath(false), _despawnWhenExpired(false)
 {
     if (args.SummonPropertiesId.has_value())
         InitializeSummonProperties(*args.SummonPropertiesId);
--- a/src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
+++ b/src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
@@ -30,18 +30,23 @@ SummonInfo::SummonInfo(Creature* summonedCreature, SummonInfoArgs const& args) :
 
 void SummonInfo::InitializeSummonSettings(SummonInfoArgs const& args)
 {
-    _remainingDuration = args.Duration;
     _summonerGUID = args.SummonerGUID;
+    _maxHealth = args.MaxHealth;
+    _remainingDuration = args.Duration;
+    _summonSpellId = args.SummonSpellId;
 
-    if (!args.SummonPropertiesID.has_value())
+    if (!args.SummonPropertiesId.has_value())
         return;
 
-    SummonPropertiesEntry const* summonProperties = sSummonPropertiesStore.LookupEntry(*args.SummonPropertiesID);
+    SummonPropertiesEntry const* summonProperties = sSummonPropertiesStore.LookupEntry(*args.SummonPropertiesId);
     if (!summonProperties)
     {
-        TC_LOG_ERROR("entities.unit", "Creature {} has been summoned with a non-existing SummonProperties.db2 entry (RecId: {}). Possible dirty DB2 data or missing hotfix entry.", _summonedCreature->GetGUID().ToString(), *args.SummonPropertiesID);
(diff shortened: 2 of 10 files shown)


src/server/game/Entities/Creature/SummonInfo/SummonInfoArgs.h
```diff
@@ -0,0 +1,36 @@
...
+#include "Define.h"
+#include "Duration.h"
+#include "Optional.h"
+#include "ObjectGuid.h"
+
+struct SummonPropertiesEntry;
+
+struct SummonInfoArgs
+{
+    Optional<ObjectGuid> SummonerGUID;
```
Nyr97, line 30: default initialization to std::nullopt of all members of this struct maybe?

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -3608,6 +3619,22 @@ void Creature::InitializeInteractSpellId()
...
+    ASSERT(_summonInfo == nullptr, "SummonInfo has already been initialized and must not be allocated again!");
+    _summonInfo = std::make_unique<SummonInfo>(this, args);
+}
+
+SummonInfo* Creature::GetSummonInfo() const
+{
+    return _summonInfo.get();
+}
+
+bool Creature::IsSummon() const
```
Nyr97, line 3633: To improve API usability, we could use GetSummonInfo() directly to return nullptr when the creature is not a summon, rather than having IsSummon and GetSummonInfo.

src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+GameObject* SummonInfo::GetGameObjectSummoner() const
+{
+    if (!_summonerGUID.has_value())
+        return nullptr;
+
+    return ObjectAccessor::GetGameObject(*_summonedCreature, *_summonerGUID);
+}
+
+bool SummonInfo::ShouldDespawnOnSummonerLogout() const
```
Nyr97, line 73: I would rename all of the bools and setters:

ShouldDespawnOnSummonerLogout -> IsDespawningOnSummonerLogout
SetShouldDespawnOnSummonerLogout -> SetDespawnOnSummonerLogout

src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
```diff
@@ -0,0 +1,101 @@
...
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "SummonInfo.h"
+#include "Creature.h"
+#include "DB2Stores.h"
+#include "ObjectAccessor.h"
+#include "SummonInfoArgs.h"
+
+SummonInfo::SummonInfo(Creature* summonedCreature, SummonInfoArgs const& args) : _summonedCreature(summonedCreature)
```
Nyr97, line 24:
```suggestion
SummonInfo::SummonInfo(Creature* summonedCreature, SummonInfoArgs const& args) : _summonedCreature(summonedCreature)
```

Nyr97, line 24: I would split the constructor into correct initialisation of members and initialisation of despawn flags:

```suggestion
SummonInfo::SummonInfo(Creature* summonedCreature, SummonInfoArgs const& args)
    : _summonedCreature(summonedCreature),
      _remainingDuration(args.Duration),
      _summonerGUID(args.SummonerGUID)
{
    InitializeDespawnFlags(args.SummonPropertiesID);
}

void SummonInfo::InitializeDespawnFlags(Optional<uint32> summonPropertiesID)
{
    uint32 propertiesID = summonPropertiesID.value_or(0);
    if (SummonPropertiesEntry const* summonProperties = sSummonPropertiesStore.LookupEntry(propertiesID))
    {
    
    auto flags = summonProperties->GetFlags();
    _shouldDespawnOnSummonerLogout = flags.HasFlag(SummonPropertiesFlags::DespawnOnSummonerLogout);
    _shouldDespawnOnSummonerDeath = flags.HasFlag(SummonPropertiesFlags::DespawnOnSummonerDeath);
    _shouldDespawnWhenExpired = flags.HasFlag(SummonPropertiesFlags::DespawnWhenExpired);
    }
}
```

src/server/game/Entities/Creature/SummonInfo/SummonInfo.h
```diff
@@ -0,0 +1,73 @@
...
+    // Returns true when the summon will despawn after its duration has expired. If not set, the summon will just die.
+    bool ShouldDespawnWhenExpired() const;
+    // Marks the summon to despawn after its duration has expired. If disabled, the summon will just die.
+    void SetShouldDespawnWhenExpired(bool set);
+
+private:
+    Creature* _summonedCreature;
+    Optional<Milliseconds> _remainingDuration; // NYI
+    Optional<ObjectGuid> _summonerGUID;
+    bool _shouldDespawnOnSummonerLogout; // NYI
```
Nyr97, line 68: If you split the constructor, these can be set to false directly and you don't need an else to set them to false

src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
Nyr97: then auto might get Shauren-banned, but huh

src/server/game/Entities/Creature/SummonInfo/SummonInfoArgs.h
Ovahlord: Optional is nullopt by default

src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
Ovahlord: The naming for these methods stems from the summon property flags, so we will follow them.

src/server/game/Entities/Creature/Creature.cpp
Ovahlord: This is a convenience helper to improve readability. Also GetSummonInfo already does return nullptr when not set because std::unique_ptr::get() returns the plain raw pointer to the object.

src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
Ovahlord: The initialization of these settings will get expanded once I get down to implement them. There are a couple more checks that must be performed to correctly assign these settings. For example totem slot summons should always despawn when the summoner dies. But this has to wait until I get to summon slots

src/server/game/Entities/Creature/SummonInfo/SummonInfoArgs.h
Nyr97: That is if it is default-constructed, however, by explicitly initializing it with std::nullopt it's more clear to other people that might work on the API on later stages.

src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
Nyr97: But the naming of those flags is explicitly set as DespawnOn-Scenario, removing verbose is fine anyway.

src/server/game/Entities/Creature/SummonInfo/SummonInfoArgs.h
Ovahlord: IMO that's really just nitpicking as anyone who works with that struct should have at least a basic understanding of how the types work that he wants to touch.

src/server/game/Entities/Creature/SummonInfo/SummonInfoArgs.h
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef _SummonInfoArgs_h__
+#define _SummonInfoArgs_h__
+
+#include "Define.h"
+#include "Duration.h"
+#include "Optional.h"
+#include "ObjectGuid.h"
+
+struct SummonPropertiesEntry;
```
mdX7, line 26: unused

mdX7 (commented):
feels kinda weird to initialize all stuff within `InitializeSummonSettings` instead of ctor. Maybe move the non SummonProperties related stuff to ctor and use the method only to initialze SummonProperties? Or is is planned that we are able to overwrite settings using `InitializeSummonSettings` on the same object? If not i'd also move the initialize function to private

src/server/game/Entities/Unit/Unit.h
```diff
@@ -1468,6 +1469,12 @@ class TC_GAME_API Unit : public WorldObject
         std::array<ObjectGuid, MAX_SUMMON_SLOT> m_SummonSlot;
         std::array<ObjectGuid, MAX_GAMEOBJECT_SLOT> m_ObjectSlot;
 
+        std::vector<SummonInfo*> _activeSummons;
```
Shauren, line 1472: should not be public

src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -50,6 +52,8 @@ Pet::Pet(Player* owner, PetType type) :
 {
     ASSERT(GetOwner());
 
+    InitializeSummonInfo({ .SummonerGUID = GetOwner()->GetGUID()});
```
Shauren, line 55: formatting (space before closing })

src/server/game/Entities/Creature/SummonInfo/SummonInfo.cpp
```diff
@@ -0,0 +1,122 @@
...
+    if (args.SummonPropertiesId.has_value())
+        InitializeSummonProperties(*args.SummonPropertiesId);
+}
+
+void SummonInfo::InitializeSummonProperties(uint32 summonPropertiesId)
+{
+    SummonPropertiesEntry const* summonProperties = sSummonPropertiesStore.LookupEntry(summonPropertiesId);
+    if (!summonProperties)
+    {
+        TC_LOG_ERROR("entities.unit", "Creature {} has been summoned with a non-existing SummonProperties.db2 entry (RecId: {}). Possible dirty DB2 data or missing hotfix entry.", _summonedCreature->GetGUID().ToString(), summonPropertiesId);
```
Shauren, line 38: Don't `.ToString()` on guids for logging (ObjectGuid is formattable) - but this comment becomes unneccessary after taking into account the id to pointer comment

Shauren (changes requested):

PR #30640, opened 2025-01-26
Scripts/Atal'Dazar: Fix Rezan mechanics on heroic and mythic difficul…

aquadeus:
…ties and improve code

Thanks to Modox and Meji

**Changes proposed:**

-  Fix Rezan mechanics on heroic and mythic difficulties and improve code

**Tests performed:**

Tested in-game


src/server/scripts/Zandalar/AtalDazar/boss_rezan.cpp
```diff
@@ -263,10 +269,43 @@ class spell_rezan_boss_emote_at_target : public SpellScript
...
+class spell_rezan_pile_of_bones_slow : public AuraScript
+{
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_PILE_OF_BONES_SLOW });
+    }
+
+    void OnRemove(AuraEffect const* aurEff, AuraEffectHandleModes /*mode*/) const
+    {
+        Unit* target = GetTarget();
```
mdX7, line 282: we should research about a generic way to handle this entire magic, sniff is also just sending SMSG_AURA_UPDATEs only instead of recasts

maybe something like OnBeforeRemove which allows to entirely skips aura removal instead of aura effect only

PR #30644, opened 2025-01-26
Core/Chat: Fix lower security check for GM commands.

r4d1sh:
**Changes proposed:**

Many chat commands in TrinityCore support a security check. For example, you cannot use the `.summon` command on a player whose security level (or, in the old terminology, GM-Level) is higher than yours.

Unfortunately, this system is broken and has been for a very long time (possibly since first commit of TrinityCore).

The issue lies in the server incorrectly interpreting the value of the config parameter `GM.LowerSecurity`. According to its description, when enabled, it should disable the lower security check. However, in reality, the server disables the lower security check when the parameter is set to 0 (which is the default value).

Steps to reproduce:

1. Log into the game with two accounts, one with security level 3 and the other with security level 2.
2. Notice that a character with a lower security level can use the `.summon` command on a character with a higher security level.

**Tests performed:**

Builded (Ubuntu 24.04.1 LTS, clang 18.1.3) and tested in-game


CraftedRO:
https://github.com/TrinityCore/TrinityCore/issues/11908 maybe related ?

r4d1sh:
> #11908 maybe related ?

Nope, it's not related.

Shauren (changes requested):
I disagree with this change, this is a "extremely legacy" setting that always worked like this (including mangos), the only thing that changed about it was its description in config file - that is where i would change it (5753be3271d2f5a981709f013547f543c738d62d flipped the config description)

r4d1sh:
> I disagree with this change, this is a "extremely legacy" setting that always worked like this (including mangos), the only thing that changed about it was its description in config file - that is where i would change it ([5753be3](https://github.com/TrinityCore/TrinityCore/commit/5753be3271d2f5a981709f013547f543c738d62d) flipped the config description)

What do you think is the best thing to do with `RBAC_PERM_CHECK_FOR_LOWER_SECURITY`? It doesn't seem to fully match its name. In the current implementation, it actually means 'allow ignoring the check if it is disabled in the config.' This permission is granted to all access levels starting from moderator and seems unnecessary. We could either remove it or change it to 'allow ignoring the check even if it is enabled in the config' and unlink it from all levels by default.

Shauren:
> > I disagree with this change, this is a "extremely legacy" setting that always worked like this (including mangos), the only thing that changed about it was its description in config file - that is where i would change it ([5753be3](https://github.com/TrinityCore/TrinityCore/commit/5753be3271d2f5a981709f013547f543c738d62d) flipped the config description)
> 
> What do you think is the best thing to do with `RBAC_PERM_CHECK_FOR_LOWER_SECURITY`? It doesn't seem to fully match its name. In the current implementation, it actually means 'allow ignoring the check if it is disabled in the config.' This permission is granted to all access levels starting from moderator and seems unnecessary. We could either remove it or change it to 'allow ignoring the check even if it is enabled in the config' and unlink it from all levels by default.

The old (pre-rbac) implementation if commands were enabled for level 0 access, its intent was to always perform the gmlevel check, no matter what was set in the config
RBAC simply replaced that with permission check (granted to every non level 0 account just as before)

Shauren:
In other words
has permission -> follow config
no permission -> always check level

r4d1sh:
> In other words has permission -> follow config no permission -> always check level

Would it be a good decision to remove this permission and replace it with a check `!AccountMgr::IsPlayerAccount(m_session->GetSecurity())`?

r4d1sh:
- Clarified of description of `GM.LowerSecurity` config option.
- Name and description of RBAC option that allows ignoring the lower security check has been made clearer too.

PR #30646, opened 2025-01-26
Update for Ice block

Zaekor:
Fix for the Ice block spell not working when you want to cast it and not applying the Hypothermia debuff

**Tests performed:**

tested in-game

### DIFF
--- a/src/server/scripts/Spells/spell_mage.cpp
+++ b/src/server/scripts/Spells/spell_mage.cpp
@@ -68,6 +68,7 @@ enum MageSpells
     SPELL_MAGE_GIRAFFE_FORM                      = 32816,
     SPELL_MAGE_ICE_BARRIER                       = 11426,
     SPELL_MAGE_ICE_BLOCK                         = 45438,
+    SPELL_MAGE_HYPOTHERMIA                       = 41425,
     SPELL_MAGE_IGNITE                            = 12654,
     SPELL_MAGE_INCANTERS_FLOW                    = 116267,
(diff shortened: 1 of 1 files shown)


Shauren:
I'm sorry, how is this tested? You can't just touch only part of the spell and re-break what the existing script was doing... How did you not notice that Ice Block now heals you 6% per second even without Cryo-Freeze talented?

Shauren:
By the way, Hypothermia is already triggered by script `spell_gen_trigger_exclude_caster_aura_spell` which is already attached to spell 45438 (Ice Block)

Zaekor:
> I'm sorry, how is this tested? You can't just touch only part of the spell and re-break what the existing script was doing... How did you not notice that Ice Block now heals you 6% per second even without Cryo-Freeze talented?
> 
> The part you deleted was handling https://www.wowhead.com/spell=320913/everwarm-socks

I looked at how the spell works on retails and corrected it here

I disagree [Ice Block](https://www.wowhead.com/spell=45438/ice-block) does not heal (unless [Cryo-Freeze](https://www.wowhead.com/spell=382292/cryo-freeze) talent is applied) 

En ce qui concerne Everwarm Socks, si vous regarder bien les changelog il ne soigne plus depuit 11.0.5 (Tue Oct 22 2024) Build 57689


> By the way, Hypothermia is already triggered by script `spell_gen_trigger_exclude_caster_aura_spell` which is already attached to spell 45438 (Ice Block)

I don't see anything on spell_gen_trigger_exclude_caster_aura_spell

Shauren:
I have no idea what you frenched out there

PR #30649, opened 2025-01-27
Core/Player: Fixed an issue where overridden specialization spells we…

### LINKED ISSUE #30648: Core/Spells: Missing Triggered Spell Check and Action Bar Update for aura SPELL_AURA_OVERRIDE_ACTIONBAR_SPELLS (e.g., Dash of Chaos)
### Description

I have encountered two issues related to the SPELL_AURA_OVERRIDE_ACTIONBAR_SPELLS aura in TrinityCore:

1. Triggered Spells Check
This aura does not properly check for triggered spells to find the matching spell that should be overridden. As a result, some spells that should be overridden are not handled correctly.

2. Action Bar Display
When a spell is overridden  (e.g., replacing Fel Rush with Dash of Chaos), it is not shown on the action bar. The overridden spell should replace the corresponding action bar slot or be added to the action bar, ensuring that it is visible and accessible to the player.

### Expected behaviour

- The system should search for triggered spells when handling SPELL_AURA_OVERRIDE_ACTIONBAR_SPELLS to find the correct spell to override.
- Overridden spells should automatically appear on the action bar, replacing the original spell or being added to the appropriate slot.

1. Cast a spell or effect that applies SPELL_AURA_OVERRIDE_ACTIONBAR_SPELLS (e.g., Dash of Chaos ID: 427793)..
2. Observe that the system does not handle triggered spells like Fel Rush ID: 195072  correctly for overrides.
3. Notice that overridden spells do not appear on the action bar.

### Branch

master

### TC rev. hash/commit

a5c547a383e40b155a2063404c040ea0007c9ebd

### Operating system

Windows 11 x64

### Custom changes

None

Eliminationzx:
…re incorrectly removed for all specializations. Now, they are only removed for the primary specialization when changing specs

**Changes proposed:**

- Spells are now removed only for the primary specialization when the player's spec is changed.
- The loop for iterating through all specializations has been moved to call in `Player::ResetTalentSpecialization`, ensuring a more precise behavior for the primary specialization.
- Mastery spells are now properly handled for the primary specialization.

**Issues addressed:**

Closes #30648

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -2370,11 +2370,8 @@ bool Player::IsMaxLevel() const
 void Player::InitTalentForLevel()
 {
     uint8 level = GetLevel();
-    // talents base at level diff (talents = level - 9 but some can be used already)
-    if (level < MIN_SPECIALIZATION_LEVEL)
-        ResetTalentSpecialization();
-
     int32 talentTiers = DB2Manager::GetNumTalentsAtLevel(level, Classes(GetClass()));
+
     if (level < 15)
     {
         // Remove all talent points
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -27362,7 +27362,7 @@ void Player::ResetTalentSpecialization()
 
     ResetPvpTalents();
 
-    RemoveSpecializationSpells();
+    RemoveAllSpecializationsSpells();
 
     ChrSpecializationEntry const* defaultSpec = ASSERT_NOTNULL(sDB2Manager.GetDefaultChrSpecializationForClass(GetClass()));
     SetPrimarySpecialization(defaultSpec->ID);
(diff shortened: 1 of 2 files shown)


Shauren:
And what exactly changes? Swapping specialization does something like this
```
RemoveSpecializationSpells();
SetPrimarySpecialization();
LearnSpecializationSpells();
```

Eliminationzx:
> And what exactly changes? Swapping specialization does something like this
> 
> ```
> RemoveSpecializationSpells();
> SetPrimarySpecialization();
> LearnSpecializationSpells();
> ```

Before changing the specialization, we only remove the spells associated with the previous primary specialization, rather than attempting to remove spells for all specializations. While I am not entirely certain this is the optimal approach, it effectively resolves the issue in the specific case of the spell from the Destruction specialization (Destruction Warlock, ID: 137046).

Shauren:
I know exactly what the change you made is, what I want to know is why this works

Eliminationzx:
> I know exactly what the change you made is, what I want to know is why this works

I am not entirely sure why this change works yet. I am currently investigating the underlying behavior to understand the root cause and will provide more details once I have a clearer picture. If you have any insights or suggestions for specific areas to explore, I would greatly appreciate your input.

hwis:
segfault after applying this changes in SpellMgr::AddSameEffectStackRuleSpellGroups

Eliminationzx:
> segfault after applying this changes in SpellMgr::AddSameEffectStackRuleSpellGroups

its not ready yet

Shauren:
Uh... what happened here?

Eliminationzx:
> Uh... what happened to this?
I'm not entirely sure about the details regarding this PR. It ended up being a bit of a complicated issue, and I felt it was better to close it for now.

PR #30655, opened 2025-01-30
Scripts/AtalDazar: Implement Priestess Alunza encounter

aquadeus:
**Changes proposed:**

-  Implement Priestess Alunza encounter

**Tests performed:**

Tested in-game

- [x] TODO: orbs should have collision but is not working properly so they are despawning when reach fixed position
- [x] TODO: the npc should select the nearest at of tainted blood and go to his position, if not ats found should attack players atm is not working, just attacking


src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,684 @@
...
+// 130738 - Corrupted Gold
+struct npc_priestess_alun_za_corrupted_gold : public ScriptedAI
+{
+    npc_priestess_alun_za_corrupted_gold(Creature* creature) : ScriptedAI(creature) { }
+
+    void JustAppeared() override
+    {
+        DoCastSelf(SPELL_CORRUPTED_GOLD);
+
+        // TODO: orbs have collision but dont work properly so they are despawning when reach fixed position
```
Nyr97, line 265: collision you mean when hitting the WMO to bounce back?

aquadeus: yes but some of them get collisions with the cauldron and dont come out

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,684 @@
...
+    void OnRemove(AuraEffect const* aurEff, AuraEffectHandleModes /*mode*/) const
+    {
+        if (GetTargetApplication()->GetRemoveMode() != AURA_REMOVE_BY_EXPIRE)
+            return;
+
+        Unit* target = GetTarget();
+        if (!target)
+            return;
+
+        target->CastSpell(target, GetEffectInfo(EFFECT_0).CalcValue(), CastSpellExtraArgsInit{
```
Nyr97, line 431: CalcValue would only matter here if there was an aura which modifies the BasePoints of this spell's EFFECT_0

Shauren: bad suggestion, always calcvalue, never read basepoints directly

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,684 @@
...
+            return;
+
+        while (uint32 eventId = _events.ExecuteEvent())
+        {
+            switch (eventId)
+            {
+                case EVENT_CHECK_TAINTED_BLOOD:
+                {
+                    // TODO: the npc should select the nearest at of tainted blood and go to his position, if not ats found should attack players atm is not working, just attacking
+                    std::vector<AreaTrigger*> atList = me->GetAreaTriggers(SPELL_TAINTED_BLOOD_AT);
```
Nyr97, line 329: It doesn't work because GetAreaTriggers returns areatriggers owned by the unit using the getter, these ats are cast by the boss, just use the boss as the getter and it'll be fixed.

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,684 @@
...
+
+        target->CastSpell(target, GetEffectInfo(EFFECT_0).CalcValue(), CastSpellExtraArgsInit{
+            .TriggerFlags = TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR,
+            .TriggeringAura = aurEff
+        });
+    }
+
+    void Register() override
+    {
+        OnEffectRemove += AuraEffectRemoveFn(spell_priestess_alun_za_molten_gold::OnRemove, EFFECT_0, SPELL_AURA_DUMMY, AURA_EFFECT_HANDLE_REAL);
```
Shauren, line 439: General rule: don't cast spells in aura `On*` hooks, only do it in `After*`

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,684 @@
...
+
+        Talk(SAY_WIPE);
+
+        _EnterEvadeMode();
+        _DespawnAtEvade();
+    }
+
+    void KilledUnit(Unit* victim) override
+    {
+        if (!victim->IsPlayer())
```
Nyr97, line 149: I suggest using some type of time_t to prevent spamming kill texts. Some type of 3-5s of difference should be ok.

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,684 @@
...
+    {
+        return ValidateSpellInfo({ SPELL_MOLTEN_GOLD_MISSILE });
+    }
+
+    void HandleScript(SpellEffIndex /*effIndex*/) const
+    {
+        GetCaster()->CastSpell(GetHitUnit(), SPELL_MOLTEN_GOLD_MISSILE, CastSpellExtraArgsInit{
+            .TriggerFlags = TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR,
+            .TriggeringSpell = GetSpell(),
+            .OriginalCastId = GetSpell()->m_castId
```
Shauren, line 456: Setting TriggeringSpell also sets OriginalCastId, this is unneccessary (there are more cases of this in this PR)

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,684 @@
...
+        OnDestinationTargetSelect += SpellDestinationTargetSelectFn(spell_priestess_alun_za_tainted_blood::SetDest, EFFECT_0, TARGET_DEST_DEST);
+    }
+};
+
+// 255577 - Transfusion
+class spell_priestess_alun_za_tranfusion_cast : public SpellScript
+{
+    void HandleEnergy(SpellEffIndex /*effIndex*/) const
+    {
+        GetCaster()->SetPower(POWER_ENERGY, 0);
```
Nyr97, line 508: Is this necessary? Spell 255577 has 100 energy power cost, this shouldn't be required.

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
Nyr97: oh I thought we were not supposed to use CalcValue when not required, noted

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
Nyr97: well, those points you're using as destination are not really using MovePositionToFirstCollision, that would fix that issue you're experiencing

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
aquadeus: didnt realize that

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,668 @@
...
+    }
+};
+
+// 258681 - Energy Regen
+class spell_priestess_alun_za_energy_regen : public AuraScript
+{
+    void OnPeriodic(AuraEffect const* aurEff) const
+    {
+        Unit* target = GetTarget();
+        if (target)
```
mdX7, line 385: target never null

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,668 @@
...
+// 255591 - Molten Gold
+class spell_priestess_alun_za_molten_gold : public AuraScript
+{
+    void OnRemove(AuraEffect const* aurEff, AuraEffectHandleModes /*mode*/) const
+    {
+        if (GetTargetApplication()->GetRemoveMode() != AURA_REMOVE_BY_EXPIRE)
+            return;
+
+        Unit* target = GetTarget();
+        if (!target)
```
mdX7, line 428: target never null

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,668 @@
...
+// 255577 - Transfusion
+class spell_priestess_alun_za_tranfusion_cast : public AuraScript
+{
+    void OnRemove(AuraEffect const* aurEff, AuraEffectHandleModes /*mode*/) const
+    {
+        if (GetTargetApplication()->GetRemoveMode() != AURA_REMOVE_BY_EXPIRE)
+            return;
+
+        Unit* target = GetTarget();
+        if (!target)
```
mdX7, line 510: target never null

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -0,0 +1,668 @@
...
+struct at_priestess_alun_za_corrupted_gold : AreaTriggerAI
+{
+    at_priestess_alun_za_corrupted_gold(AreaTrigger* areatrigger) : AreaTriggerAI(areatrigger) { }
+
+    void OnUnitEnter(Unit* unit) override
+    {
+        if (!unit->IsPlayer())
+            return;
+
+        at->GetCaster()->CastSpell(unit, SPELL_CORRUPTED_GOLD_DAMAGE, TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR);
```
mdX7, line 643: caster might be null (possible crash)

mdX7 (changes requested):

src/server/scripts/Zandalar/AtalDazar/boss_priestess_alun_za.cpp
```diff
@@ -382,8 +382,7 @@ class spell_priestess_alun_za_energy_regen : public AuraScript
     void OnPeriodic(AuraEffect const* aurEff) const
     {
         Unit* target = GetTarget();
```
Nyr97, line 384: there's no need to make a pointer if you're not checking for null, I guess this is because the holder of the aura is the same as the caster and there's no way there's one without the other in this case?

PR #30679, opened 2025-02-08
Core/Spells: Added missing DBC data for some food effects.

### LINKED ISSUE #27192: Scripts/Spells: Food from Bobbing Apple should give Well Fed after 10s not when breaking the eating
### Description

Atm it does give you the well fed buff anytime after you break the eating.

https://user-images.githubusercontent.com/24683355/139061829-ded09edf-d9a5-4f46-b916-e4980b0f09cd.mp4

### Expected behaviour

https://tbc.wowhead.com/item=20516/bobbing-apple#comments
>  If you spend at least 10 seconds eating you will become well fed and gain Stamina and Spirit for 15 min.

.additem 20516

### Branch

3.3.5

### TC rev. hash/commit

1403de8dafba

### Operating system

Windows 11 x64

### Custom changes

None

r4d1sh:
**Changes proposed:**

There are several spells (food auras) for which it is very likely that the effect applying the Well Fed buff is missing in `Spell.dbc`. This was previously implemented using `spell_linked_spell`, but it does not work correctly (see the linked issue). A script could be written to handle this, but I believe this is a case where using the `ApplySpellFix` function is the right approach.

I have also removed some scripts for Pilgrim’s Bounty food and applied `ApplySpellFix` to them as well, as I believe this method is well-suited for this case.

**Issues addressed:**

Closes #27192

**Tests performed:**

Builded and tested in-game.


Shauren:
... its ok i guess? those spells are a mess in retail too (some but not all pilgrim's bounty foods got proper trigger spells now)

PR #26477, opened 2021-04-25
Core/Sounds - added missing victory near sound for Arathi Basin (3.3.5)

schell244:
**Changes proposed:**
-  added missing sound for arathi basin

**Target branch(es):**
- [x] 3.3.5

**Issues addressed:**
Closes https://github.com/TrinityCore/TrinityCore/issues/26474

**Tests performed:**
Builds, Tested

PR #30690, opened 2025-02-14
Core/BG: Properly update near victory sound for Arathi Basin and score treeshold

CraftedRO:
**Changes proposed:**

-  In [this PR](https://github.com/TrinityCore/TrinityCore/pull/26477) victory near sound for Arathi Basin were added, however they are inversed as you can see in following proofs:
-  [Here Horde winning](https://youtu.be/icOnwfoKoXU?t=860) and in the background we [hear this sound - PVPWarningAllianceLONG](https://www.wowhead.com/wotlk/sound=8456/pvpwarningalliancelong).
-  [Here Alliance winning](https://youtu.be/MGfcBW7Hmn4?t=870) and in the background we [hear this sound - PVPWarningHordeLONG](https://www.wowhead.com/wotlk/sound=8457/pvpwarninghordelong).
- Adtitionally, I found this [Vmangos sniffed fix](https://github.com/vmangos/core/pull/874/files).

- I will push one more commit because in the videos above from wotlk classic we can see that the treeshold for BG_AB_WARNING_NEAR_VICTORY_SCORE is actually 1440 instead of 1400 as it is on TC atm.

**Issues addressed:**

**Tests performed:**

- [ None ]


Shauren:
Can confirm near victory score requirement, found a 3.3.5 sniff

Shauren (approved):

CraftedRO:
yeah, well when bg required 2000 points, threeshold was at 1800 so probably it is on 90% from total bg score, at least for AB.

Screenshot 2025-02-14 155341

Issue #26924, opened 2021-09-18
Core/Spell: Wanding doesn't swap to player target


CraftedRO:
### Description

For some reason when you attack with a equipped wand a target and meanwhile you swap another target , you still keep attack (shoot) starting target.
I tested this on Aowow (trinity) server.

https://user-images.githubusercontent.com/24683355/133888827-7c72034c-0931-4333-bbb2-6bde43e6dcf1.mp4

### Expected behaviour

When you swithing targets while wanding , your next shoot shoud be casted on your new / current target.
[Some yt proof : link](https://youtu.be/y-4iDn8dwBU) / https://youtu.be/2wFWxbQRnpE?t=14

### Steps to reproduce the problem

Equip a wand on a mage / warlock / priest (I tested on warlock)  then start casting shoot on a target.
Switch target while casting.
Notice you are still shooting main target.

### Branch

3.3.5

### TC rev. hash/commit

4b028e599e06

### Operating system

probably debian have aowow

### Custom changes

None I guess

Jinnaix:
Same behavior for TBC(C) (just tested on TBC PTR )- switching the target while having shoot on autoattack will attack the new target. Unless this has not been changed for/in WotLK, current mechanic is wrong.

CraftedRO:
Just in case if relevant , AzerothCore commited this fix https://github.com/azerothcore/azerothcore-wotlk/commit/42019c09b12697a38fd0a9af82b9c94b62529bfe but it doesn't build to TC
![wanbuilderror](https://user-images.githubusercontent.com/24683355/134186909-47476bb7-6749-4165-8539-8328fc83cd17.PNG)

CraftedRO:
hey @Keader  if you got time maybe you can take a look here too :D

CraftedRO:
recorded today on WotLK classic PTR

https://github.com/TrinityCore/TrinityCore/assets/24683355/246ce759-180f-4bb8-861d-8b2da2c3d583

CraftedRO:
Can someone point me where the function that takes care of wanding (target) is, this issue is really annoying considering in 3.3.5 wanding does still have an important role in leveling and so on.

CraftedRO:
found a fix, unable to make a PR because license incompatibility
```diff
diff --git a/src/server/game/Handlers/MiscHandler.cpp b/src/server/game/Handlers/MiscHandler.cpp
index eda9900a26ef8..b7590c70b5eec 100644
[...]

PR #30695, opened 2025-02-16
Core/Spells: Updates autoshoot spell target in case of client-side target change

CraftedRO:
**Changes proposed:**

-  Updates autoshoot spell target in case of clientside target change
-  Fix based on Vmangos and other projects.

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/26924

**Tests performed:**

- [ None ]


src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -470,6 +470,24 @@ void WorldSession::HandleSetSelectionOpcode(WorldPacket& recvData)
...
+    // Update target of current autoshoot spell
+    if (guid)
+    {
+        if (Spell* autoReapeatSpell = _player->GetCurrentSpell(CURRENT_AUTOREPEAT_SPELL))
+        {
+            if (autoReapeatSpell->m_targets.GetUnitTargetGUID() != guid)
+            {
+                if (Unit* unit = ObjectAccessor::GetUnit(*_player, guid))
+                {
+                    if (unit->IsAlive() && !_player->IsFriendlyTo(unit) && unit->isTargetableForAttack())
```
Shauren, line 483: The correct check in all cases (replacing all three condtions you have here) is `_player->IsValidAttackTarget(unit, autoReapeatSpell->GetSpellInfo())`

Shauren:
I've done some research on this and believe this is a blizzlike bug which got fixed on classic

What controls this behavior is SPELL_ATTR4_AUTO_RANGED_COMBAT
In 3.3.5, wand shoot is missing this attribute
In classic blizz changed the spell in build 1.13.2.30073

This attributet must be present in client dbc, adding it serverside will not do anything

CraftedRO:
well there is hope, one of the biggest wotlk 3.3.5 server have this implemented, what I have to is search a valid video from back then as proof, I don't think they mess with this up, imo this shoud behave same like hunter autoshoot, but will see.

CraftedRO:
found smth here, use the lower playback speed, will search for more if needed https://youtu.be/2wFWxbQRnpE?t=14

CraftedRO:
what I'm not 100% sure is the intrerupted message, in some other video I found for example when target dies while you shoot with wand it show you "your target is dead" as error and not allways intrerupted.

Issue #20972, opened 2017-11-25
Titans Grip Allows For Dual Fishing Pole Buffs


gmillerd:
**Description:**

The warrior class talent Titans Grip with the Dual Wield spell and fishing skill allows for equipping two fishing poles. The tool-tip reads Axes, Maces, Swords and the buff allows for polearms and staves. Fishing poles are not true melee weapons and shouldn't be allowed to dual wield.

The side effect of this is that additional fishing skill buffs stack on the second weapon and its lure.

**Current behaviour:**

Equipping two fishing poles is possible with TG and DW and Fishing Skill.

**Expected behaviour:**

Not allowed to equip a second pole, similar to if TG was not talented. Only "weapons" can be dual wielded.

**Steps to reproduce the problem:**

1. Learn TG, DW, Fishing
2. Equip two Strong Poles
3. View fishing skill is 1+10=11

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

TrinityCore rev. eba1c1ff97af 2017-11-25 11:21:12 +0100 (3.3.5 branch) (Unix, RelWithDebInfo, Static) (worldserver-daemon)

**TDB version:**  TDB_full_world_335.63_2017_04_18.sql

**Operating system:** Linux

tkrokli:
- https://wow.gamepedia.com/Titan%27s_Grip
- http://www.wowhead.com/spell=46917/titans-grip
- https://wow.gamepedia.com/Dual_wield
- http://www.wowhead.com/spell=674/dual-wield
- https://wow.gamepedia.com/Strong_Fishing_Pole
- http://www.wowhead.com/item=6365/strong-fishing-pole

CraftedRO:
835e2918319e Still valid :


https://user-images.githubusercontent.com/24683355/213885214-878ff91f-c8a7-44ab-8924-0ad68e941458.mp4

CraftedRO:
An updated tested version of the above closed PR is here: https://github.com/Craftedro/TrinityCore/compare/3.3.5...dual-poles.diff https://github.com/CraftedRO/TrinityCore/commit/25c940e2749f0629b075a2eced56f2dcaa1e53b8
I won't open a yet but it works perfectly as can be seen here: https://youtu.be/KOwlqTfqmnc

PR #30747, opened 2025-03-13
Core/Player: Do not allow to wear two fishing poles when spec into Titans Grip

CraftedRO:
**Changes proposed:**

-  Well this will fix the targeted issue but I need some help with the mentioned issues in the second commit.
-  All the credit goes to [Sava Vranešević](https://github.com/svranesevic), I only updated a bit his abandoned pr to fit new tc.

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/20972

**Tests performed:**

- [ I commented out (in the second commit) the parts of the code that need improvements. ]

### DIFF
--- a/src/server/game/Entities/Item/ItemTemplate.cpp
+++ b/src/server/game/Entities/Item/ItemTemplate.cpp
@@ -139,6 +139,32 @@ uint32 ItemTemplate::GetSkill() const
     }
 }
 
+bool ItemTemplate::IsWeaponOfSubClass(ItemSubclassWeapon subClass) const
+{
+    if (Class != ITEM_CLASS_WEAPON)
+        return false;
+
+    return SubClass == subClass;
+}
+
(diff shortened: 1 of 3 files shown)


src/server/game/Entities/Item/ItemTemplate.cpp
```diff
@@ -139,6 +139,32 @@ uint32 ItemTemplate::GetSkill() const
     }
 }
 
+bool ItemTemplate::IsWeaponOfSubClass(ItemSubclassWeapon subClass) const
+{
+    if (Class != ITEM_CLASS_WEAPON)
+        return false;
+
+    return SubClass == subClass; // here we got a sign unsigned error iirc
```
CraftedRO, line 147: not sure if smth like this will work here: return SubClass == static_cast<uint32>(subClass);

CraftedRO:
But I remeber that error messages wasn't appear at all, if there shoud be any related errors on-screen.

PR #30750, opened 2025-03-14
Core/Misc: Add configuration option for setting the update interval of the Who list cache

Jildor:
**Changes proposed:**

-  Just add configuration to set the update interval of the Who list cache, nothing change in game.

**Issues addressed:**

None

**Tests performed:**

tested in-game


CraftedRO:
meh looks like on retail if I spamm refresh button, the list will be updated at 5 seconds interval (our default value), while this change won't affect anything (except that I have to add in my config) I will always be against adding almost useless features, but don't take me wrong, is just mho.

Shauren (changes requested):
Doesn't work with `.reload config`

Jildor:
Done ;)

PR #30760, opened 2025-03-16
Spells/Druid: Implement talents Umbral Embrace/Umbral Inspiration

aquadeus:
**Changes proposed:**

-  Implement talents Umbral Embrace/Umbral Inspiration

**Tests performed:**

Tested in-game


Shauren (changes requested):
Doesn't work, damage is not increased (but auras get applied/removed correctly)

393763 does not have any damage increasing effect

PR #30783, opened 2025-03-23
Core/Spells: Implement druid talent Crashing Star

aquadeus:
**Changes proposed:**

-  Implement druid talent Crashing Star

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_druid.cpp
```diff
@@ -425,6 +427,28 @@ class spell_dru_celestial_alignment : public SpellScript
...
+class spell_dru_crashing_star : public SpellScript
+{
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_DRUID_CRASHING_STAR_TALENT, SPELL_DRUID_CRASHING_STAR_DAMAGE });
+    }
+
+    void HandleDamage(SpellEffIndex /*effIndex*/)
+    {
+        if (Unit* caster = GetCaster())
```
mdX7, line 440: caster never null in `SpellScript´ unless gob target is involved

mdX7 (changes requested):
`Shooting Stars has a $s1% chance to instead call down a Crashing Star`

with the current implementation we could get a Shooting Star + Crashing Star simultaneously, i think thats not intended?

mdX7 (approved):

Issue #30103, opened 2024-07-18
Core: Authserver shoud not close even if no internet connection and hostname cannot be resolved to IP address


CraftedRO:
### Description

At the moment in my tests Authserver will close when no internet connection (hostname cannot be resolved to IP address) first time when started.

### Expected behaviour

In my opinion Authserver shoud keep running even if no internet connection (hostname cannot be resolved to IP address) first time when it's started and keep trying to resolve it as it does when ethernet disconnect after this validation.

### Steps to reproduce the problem

connect to `auth`.`realmlist` database
change adress field to smth like test.test.com (or test.test and skip next step)
unplug ethernet cable
try to start authserver
notice

### Branch

3.3.5

### TC rev. hash/commit

f57647287af1f47a1c4a503cde654d6ff8f4261c

### Operating system

Windows 11 x64

### Custom changes

None

CraftedRO:
Seems related with hosts file but that's sux when you have dynamic IP because probably there's need aditional tool/script to update that file.

CraftedRO:
Reopening this issue since I have a simple fix.

PR #30786, opened 2025-03-24
Core/Auth: Do not close Authserver at startup when no valid realms found

CraftedRO:
**Changes proposed:**

-  Do not close Authserver at startup when no valid realms found, just skip to the seek loop.

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/30103

**Tests performed:**

- [ None, maybe few more tests ]

### DIFF
--- a/src/server/authserver/Main.cpp
+++ b/src/server/authserver/Main.cpp
@@ -192,12 +192,6 @@ int main(int argc, char** argv)
 
     std::shared_ptr<void> sRealmListHandle(nullptr, [](void*) { sRealmList->Close(); });
 
-    if (sRealmList->GetRealms().empty())
-    {
-        TC_LOG_ERROR("server.authserver", "No valid realms specified.");
-        // return 1; (skip to the loop)
-    }
-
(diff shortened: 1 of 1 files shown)


CraftedRO:
Self reminder: on a separate PR maybe we can add the possibility to configure the refresh interval (that looks to be 17 seconds atm) for resolving the ip.

Shauren:
> Self reminder: on a separate PR maybe we can add the possibility to configure the refresh interval (that looks to be 17 seconds atm) for resolving the ip.

20 seconds, already configurable (RealmsStateUpdateDelay)

CraftedRO:
ah I searched in the wrong config, maybe we have to move that to auth config.

Shauren:
> ah I searched in the wrong config, maybe we have to move that to auth config.

Already there - https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/authserver/authserver.conf.dist#L99

CraftedRO:
oh, my bad

Shauren (approved):

PR #30808, opened 2025-03-31
Proxy Protocol v2 support

### LINKED ISSUE #27597: network:support proxy protocol
### Description

Sometimes,  for load balancing, security and so on,we maybe use proxy like haproxy.haproxy will use proxy_protocol v2 to send user real ip.
but I only know website how to support proxy protocol and tc is use tcp and I do not know how to support it,

### Expected behaviour

seems now we not support proxy protocol and all user login info - ip info will use 127.0.0.1.
that's wrong and maybe cause other issues.

1.one machine is public.(A)
2.second machine(B) is tc server and use haproxy and others  connect with machine A

### Branch

3.3.5

### TC rev. hash/commit

5ba93a95de911277908596d805e7fa7f38a18c3a

### Operating system

Windows 10 x64

### Custom changes

None

nexrail:
**Changes proposed:**

-  Authserver
-  Worldserver
-  Shared Network

**Issues addressed:**

[Closes #27597](https://github.com/TrinityCore/TrinityCore/issues/27597)

**Tests performed:**

Auth- and world-server tested and show/stores real client ips in account table, with haproxy frontend.

- [ ] Failed to retrieve client's remote address remote_endpoint: Transport endpoint is not connected

### DIFF
--- a/src/server/shared/Networking/Socket.h
+++ b/src/server/shared/Networking/Socket.h
@@ -274,7 +274,8 @@ private:
         auto remainingLen = packet.GetActiveSize() - 16;
         readPointer += 16; // Skip strait to address.
 
-        switch (addressFamily) {
+        switch (addressFamily)
+        {
             case PROXY_HEADER_ADDRESS_FAMILY_AND_PROTOCOL_TCP_V4:
             {
                 if (remainingLen < 12)
--- a/src/server/shared/Networking/Socket.h
+++ b/src/server/shared/Networking/Socket.h
@@ -25,6 +25,7 @@
 #include <memory>
 #include <functional>
 #include <type_traits>
+#include <boost/asio.hpp>
 #include <boost/asio/ip/tcp.hpp>
 
 using boost::asio::ip::tcp;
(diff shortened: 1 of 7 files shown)


CraftedRO:
you missed #include "Config.h" and probably more, do you have author consent for this (that project license is not compatible with ours), can you handle eventually issues caused by this ?

src/server/shared/Networking/Socket.h
```diff
@@ -34,12 +35,24 @@ using boost::asio::ip::tcp;
 #define TC_SOCKET_USE_IOCP
 #endif
 
+enum ProxyHeaderReadingState {
```
Ovahlord, line 38: we do not inline brackets. Please move them into a new line. Apply this to all the other changes you made.

nexrail:
can we merge or should i close this request?

Shauren (changes requested):
I do not like how intrusive (into base socket) your implementation is

What I would like to see is this implemented in master branch using https://github.com/TrinityCore/TrinityCore/blob/71b681bbf0f5189cd87a6cea66ef51667223f54a/src/common/network/ConnectionInitializers/SocketConnectionInitializer.h#L26 (not **yet** in 3.3.5)

This one can serve as example of a initializer that does reading from the socket before entering main read loop https://github.com/TrinityCore/TrinityCore/blob/71b681bbf0f5189cd87a6cea66ef51667223f54a/src/server/game/Server/WorldSocket.cpp#L80

If your port is successful then it can be added to 3.3.5

PR #30705, opened 2025-02-21
Scripts/Underrot: Implement Sporecaller Zancha encounter

aquadeus:
**Changes proposed:**

-  Implement Sporecaller Zancha encounter

**Tests performed:**

Tested in-game

**Known issues and TODO list:** (add/remove lines as needed)

- Volatile pods have weird modelid/animkit
- Missing achievement "Not a fun guy"

PR #30832, opened 2025-04-09
Core/AreaTrigger: Implement AreaTriggerActionSetFlag::DontDespawnWith…

mdX7:
…Creator

**Tests performed:**
tested ingame with #30705


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -5439,10 +5438,14 @@ void Unit::RemoveAreaTrigger(AuraEffect const* aurEff)
     }
 }
 
-void Unit::RemoveAllAreaTriggers()
+void Unit::RemoveAllAreaTriggers(bool force /*= true*/)
 {
```
Nyr97, line 5442: this was called at JustDied or Reset?

Nyr97: oh, wait, weren't areatriggers despawned at death by default? in order to prevent that, you'd have to use JustDied override, being empty, however, this flag could just prevent that small workaround

mdX7: Notice "force" arg. Currently at despawning isnt handled in a generic case besides RemoveFromWorld. We could add that part to BossAI::JustDied and BossAI::EnterEvadeMode however

Shauren (approved):

Issue #19294, opened 2017-03-13
[3.3.5] Pet Talent Cobra Reflexes Should Not Play Animation


MrSmite:
[//]: # (**********************************)
[//]: # (** Fill in the following fields **)
[//]: # (**********************************)

**Description:**

[Cobra Reflexes](http://wotlk.openwow.com/spell=25077) is a passive Hunter pet talent that currently incorrectly has an audio and visual component when the pet is added to the world.

**Current behaviour:**

See description

**Expected behaviour:**

Because Cobra Reflexes is a passive talent it should not have an audio / visual display when the pet is added to the world, similar to other passive talents such as [Spiked Collar](http://wotlk.openwow.com/spell=53182). Only proc talents such as [Frenzy](http://wotlk.openwow.com/spell=19624) trigger audio and visual components.

NOTE: I could not find any specifics on wowhead about this spell because of Legion and there isn't really any good "evidence" on YouTube. The basis of my report is simply that no other passive talents have an audio and visual component applied.

**Steps to reproduce the problem:**

1. Select Cobra Reflexes in the Hunter pet talent tree
2. Do something such as zoning, dismounting or call pet to trigger adding the pet to the world
3. Notice the giant cobra animation and hissing sound that is incorrectly displayed

**Branch(es):**

3.3.5

**TC rev. hash/commit:** 

https://github.com/TrinityCore/TrinityCore/commit/8c6fa411e8118752ddbefe8dc022514602859fb9

> *CHANGEME Copy the first line of the `worldserver`*
> Question: On Windows, how do you copy text out of the console? Right-click does nothing and CTRL+C stops the server...

**TDB version:**

TDB_full_world_335.61_2016_04_11 + updates

**Operating system:**

WinXP / Win7

[//]: # (This template is for problem reports. For other types of report, edit it accordingly.)
[//]: # (For fixes containing C++, create a Pull Request.)

mario-64:
Thank you for reporting this issue. It's existed for a long time. You're correct passive talents should make no animation or sound when a pet is called

MrSmite:
> It's existed for a long time.

Yep, I've been meaning to look into it myself but haven't had the time.

ariel-:
Should be added to pet using AddAura instead of CastSpell then.

> Question: On Windows, how do you copy text out of the console? Right-click does nothing and CTRL+C stops the server...

[...]

PR #30871, opened 2025-04-14
Core/Pet: Replace CastSpell with AddAura

CraftedRO:
**Changes proposed:**

-  Replace CastSpell with AddAura so passive spells like [Cobra Reflexes](https://www.wowhead.com/wotlk/spell=61682/cobra-reflexes) won't show the animation when learned or while adding pet to the world.
-  Credit goes to [MrSmite](https://github.com/MrSmite)

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/19294

**Tests performed:**

- [ Maybe only cast them for stable pets as seen in this video if blizzlike https://youtu.be/JPQon0EG7CA?t=115 ]

### DIFF
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -1481,7 +1481,7 @@ bool Pet::addSpell(uint32 spellId, ActiveStates active /*= ACT_DECIDE*/, PetSpel
     m_spells[spellId] = newspell;
 
     if (spellInfo->IsPassive() && (!spellInfo->CasterAuraState || HasAuraState(AuraStateType(spellInfo->CasterAuraState))))
-        CastSpell(this, spellId, true);
+        AddAura(spellId, this);
     else
         m_charmInfo->AddSpellToActionBar(spellInfo);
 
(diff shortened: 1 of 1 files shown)


Nyr97:
Though I understand the visual issue, AddAura should be avoided in most cases. Have you checked cast flags or if SPELL_START is sent on retail?

CraftedRO:
till now, only found a sniff with a similar spell (issue) and only those lines are present on SMSG_AURA_UPDATE_ALL:
[4] Spell ID: 50871 (Savage Rend)
[4] Flags: 9 (EffectIndex0, NotCaster)

CraftedRO:
seems not in my sniff for this specific spell

CraftedRO:
so, what's to be done in this case ?

Nyr97:
Most spells do not send SPELL_START unless the client registers the beginning of the cast. If it involves any sort of aura, update aura packet is sent. There's also proof that Blizzard can intentionally skip SPELL_START. 

Have you checked in retail if the aura is present at all without the talent? Maybe it's there by default and it just sends an update once the talent is set. In any case, I'd say wait for Shauren or Modox who might know better.

CraftedRO:
In retail these spells do not longer exist, nor on classic/cata ptr.

PR #30880, opened 2025-04-16
Core/Spell: Warrior, Thunder Clap

MoltenCrystal:
**Changes proposed:**

-  Applies slow
-  Applies Rend to targets if Rend is talented
-  

**Issues addressed:**

**Tests performed:**

Builds and works ingame like it should

- [ ] 
- [ ]


MoltenCrystal:
Need pointers on castflags/triggers, as I know I should not use "true" in any case

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -75,6 +75,9 @@ enum WarriorSpells
     SPELL_WARRIOR_TRAUMA_EFFECT                     = 215537,
     SPELL_WARRIOR_VICTORIOUS                        = 32216,
     SPELL_WARRIOR_VICTORY_RUSH_HEAL                 = 118779,
+    SPELL_WARRIOR_THUNDER_CLAP_SLOW         = 435203,
+    SPELL_WARRIOR_THUNDER_CLAP_REND         = 388539,
+    SPELL_WARRIOR_REND                      = 772,
```
aquadeus, line 80: codestyle

sql/updates/world/master/2025_04_16_01_world.sql
```diff
@@ -0,0 +1,3 @@
+-- Thunder Clap (SpellID: 6343)
+DELETE FROM `spell_script_names` WHERE spell_id = 6343;
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES (6343, 'spell_warr_thunder_clap');
```
aquadeus, line 3: add a new line here

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -833,6 +836,45 @@ class spell_warr_victory_rush : public SpellScript
...
+            SPELL_WARRIOR_THUNDER_CLAP_REND,
+            SPELL_WARRIOR_REND
+        });
+    }
+
+    void HandleSlow(SpellEffIndex /*effIndex*/)
+    {
+        // Apply movement speed reduction
+        if (Unit* target = GetHitUnit())
+        {
```
Nyr97, line 856: single liner, no bracket if possible

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -833,6 +836,45 @@ class spell_warr_victory_rush : public SpellScript
...
+            SPELL_WARRIOR_THUNDER_CLAP_SLOW,
+            SPELL_WARRIOR_THUNDER_CLAP_REND,
+            SPELL_WARRIOR_REND
+        });
+    }
+
+    void HandleSlow(SpellEffIndex /*effIndex*/)
+    {
+        // Apply movement speed reduction
+        if (Unit* target = GetHitUnit())
```
Nyr97, line 855: there's no need to check for gethitunit in OnEffectHitTarget, it's never null at that point

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -833,6 +836,45 @@ class spell_warr_victory_rush : public SpellScript
...
+        {
+            GetCaster()->CastSpell(target, SPELL_WARRIOR_THUNDER_CLAP_SLOW, true);
+        }
+    }
+
+    void HandleRend()
+    {
```
Nyr97, line 862: why not OnEffectHitTarget

Nyr97:
Also, as a rule of thumb, we don't generically handle all kind of effects on a single spellscript, say there are 20 talents which modify or do stuff under certain conditions. We create specific spellcripts for a specific purpose and then attach it to the specific spellId.

aditsa2 (approved):

mdX7:
> Besides ValidateSpellInfo, there's ValidateEffectInfo: for example, if you're hooking a given void to effect 1, but you need to access data of effect 0 of the same spellId, you need to validate that effect. The same applies if you need to check effect of a different spellId, see here for reference: https://github.com/TrinityCore/TrinityCore/pull/30841/files#diff-3743b23036148800b8e23d6255c92915cd5262167f00c13869bf1e23c95da4e5R659

Ur mistaken here, hooks are checking if they are valid themselves - if they are not valid they print an error. For reference: [`SpellScript::_Validate`](https://github.com/TrinityCore/TrinityCore/blob/3665fa2596191819d275252845dcf97b5e87af7c/src/server/game/Spells/SpellScript.cpp#L257)
`ValidateEffectInfo` has to be used if you have to access a specific effect of a spell e.g. if you have to use `.CalcValue(GetCaster())`. That may happen via. `GetEffectInfo(EFFECT_1)` or `<spellInfo object here>.GetEffect(EFFECT_1)`.

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -833,6 +836,45 @@ class spell_warr_victory_rush : public SpellScript
...
+        if (GetCaster()->HasSpell(SPELL_WARRIOR_REND))
+        {
+            // Apply Rend (388539) to the target
+            GetCaster()->CastSpell(GetHitUnit(), SPELL_WARRIOR_THUNDER_CLAP_REND, true);
+        }
+
```
aquadeus, line 868: remove

MoltenCrystal: Why remove this, this is what applies it?

src/server/scripts/Spells/spell_warrior.cpp
MoltenCrystal: Wouldn't the HandleSlow be needed to be called something simliar? Or how should I name these properly?

sql/updates/world/master/2025_04_16_01_world.sql
MoltenCrystal: So always an empty line after each "complete" statement?

src/server/scripts/Spells/spell_warrior.cpp
aquadeus: i mean empty line, the compiler gets that as an error

MoltenCrystal: Ah you meant the empty line sorry I missed that

sql/updates/world/master/2025_04_16_01_world.sql
aquadeus: yes

PR #30881, opened 2025-04-17
Core/Spells: treat spells casts triggered by SPELL_EFFECT_FORCE_CAST_2  as individual casts

Ovahlord:
**Changes proposed:**

-  spells which are triggered by SPELL_EFFECT_FORCE_CAST_2  will no longer be treated as an triggered spell cast but instead they act like a followup cast controlled by the server.

**Tests performed:**
- tested ingame on 4.4.2 with the Goblin Town in a Box. Cast data now lines up with sniff behavior

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -825,8 +825,13 @@ void Spell::EffectForceCast()
             return;
     }
 
-    CastSpellExtraArgs args(TRIGGERED_FULL_MASK & ~(TRIGGERED_IGNORE_POWER_COST | TRIGGERED_IGNORE_REAGENT_COST));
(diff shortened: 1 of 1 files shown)


mdX7:
We may run into "cast in progress" issues here, i kinda recall i tried that for Elder Leaxa pr and then just fixed it differently

mdX7:
wasn't Elder Leaxa or im just unable to find the spell anymore lul. Also checked SpellMgr hacks, didn't figure out where this was the issue. Was like 2y ago.
The issue was that FORCE_CAST_2 was executed before the cast is fully "done". Forcing interrupt prior to casting the FORCE_CAST spell may result in sending spell failed packets to clients, which may seem odd.
I mean we can go with this pr until the issue happens again and then just readjust as needed

PR #30914, opened 2025-05-02
Scripts/Spells: Implement Putrid Mushroom Primer spell

offl:
**Changes proposed:**

-  Implement Putrid Mushroom Primer spell

**Issues addressed:**

none

**Tests performed:**

builds, tested in-game


src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp
```diff
@@ -159,8 +157,36 @@ struct npc_underbog_mushroom : public ScriptedAI
...
+
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_SUMMON_UNDERBOG_MUSHROOM });
+    }
+
+    void HandleScript(SpellEffIndex /*effIndex*/)
+    {
+        Creature* caster = GetCaster()->ToCreature();
+        if (!caster || !caster->IsAIEnabled())
```
Nyr97, line 173: as far as I'm concern, there's no need to check for IsAIEnabled() to access AI class ever since Jackpot introduced AI changing status where AI is never null.

src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp
```diff
@@ -159,8 +157,36 @@ struct npc_underbog_mushroom : public ScriptedAI
...
+            return;
+
+        // This is wrong. They don't need a script to simply select random target. But what should be done here? FilterTargets? Exclude pets?
+        if (Unit* target = caster->AI()->SelectTarget(SelectTargetMethod::Random, 0))
```
Nyr97, line 177: what's the target type for this spell? I'm assuming it's an AOE chooser since it's got a radius of 100 yards on it, so the correct way would be filtering through all of those targets if you want to exclude any, Spell system on itself already randomly chooses between targets when using AOE types, even when MaxTargets is set.

offl: I can add
```cpp
    void FilterTargets(std::list<WorldObject*>& targets)
    {
        if (targets.empty())
            return;

        WorldObject* target = Trinity::Containers::SelectRandomContainerElement(targets);
        targets.clear();
        targets.push_back(target);
    }

File: src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp (hunk 1 above)
Comment (offl) [reply src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp:173]: ok, thanks

File: src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp (hunk 2 above)
Comment (Shauren) [reply src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp:177]: Yes, a target script is the preferred way to handle situations like this

Does not apply to this spell (because its serverside) but for spells visible on client doing random selection in effect hook would play spell visual on every unit

Comment (offl) [reply src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp:177]: I can simply filter targets but it's still unknown what exactly that script should do, simply selecting random target still looks too easy

Comment (Nyr97) [reply src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp:177]: since you lack information, the only way to test this is forcing that spell several times until you find a pattern, if you can't find any, random it is, for as long as you can reproduce the expected result

Comment (offl) [reply src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp:177]: I think better to check sniffs but so far I don't have any sniff

Comment (Nyr97) [reply src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp:177]: is the .empty check really necessary at the beginning? if there are no targets available it shouldn't even trigger OnObjectAreaTargetSelect iirc

Comment (Shauren) [reply src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp:177]: > is the .empty check really necessary at the beginning? if there are no targets available it shouldn't even trigger OnObjectAreaTargetSelect iirc, it would make sense if it was after some type of filtering in case you set it empty entirely

YES

Comment (Nyr97) [reply src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp:177]: if there are no targets available, why would the system force the hook processing at all?

**Comment by offl** [comment]
Now with FilterTargets

File: src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp [hunk 3]
```diff
@@ -167,20 +168,26 @@ class spell_hungarfen_putrid_mushroom_primer : public SpellScript
...
-        if (Unit* target = caster->AI()->SelectTarget(SelectTargetMethod::Random, 0))
-            target->CastSpell(target, SPELL_SUMMON_UNDERBOG_MUSHROOM, true);
+        WorldObject* target = Trinity::Containers::SelectRandomContainerElement(targets);
+        targets.clear();
+        targets.push_back(target);
+    }
+
+    void HandleScript(SpellEffIndex /*effIndex*/)
+    {
+        GetHitUnit()->CastSpell(GetHitUnit(), SPELL_SUMMON_UNDERBOG_MUSHROOM);
```
Nyr97, line 184: no CastArgs at all, or what spell is triggering the spell, etc?

src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_the_black_stalker.cpp
```diff
@@ -99,25 +99,25 @@ struct boss_the_black_stalker : public BossAI
                         EnterEvadeMode();
```
Nyr97, line 99: I don't know how to feel about this old boundary check inside EventMap, might as well update it to be checked from InstanceScript directly

offl: We need Leash to be implemented, it's original thingy

Nyr97: well, I meant encounter boundary in our own terms where we can choose what type of, radius from point, rectangle, polygon, etc.

offl: That boundary system has nothing close to original Leash field

offl: It's just radius boundary, yeah
Well, many DB things on retail can be overriden in Lua

src/server/scripts/Outland/CoilfangReservoir/TheUnderbog/boss_hungarfen.cpp
Shauren: triggered was neccessary, you don't want either the mushroom to fail because player is casting something or players cast to be interrupted

PR #30915, opened 2025-05-03
Scripts/Spells: Get rid of some database spell scripts

offl:
**Changes proposed:**

-  Get rid of some database spell scripts

**Issues addressed:**

none

**Tests performed:**

builds, tested in-game


offl:
Eh, I tried to avoid using spell scripts with args

Issue #30169, opened 2024-08-18
Tamed pet stays in combat after all mobs have been killed and is slower than the player character


F0rce1991:
### Description

The pet tamed during the 'Taming the Beast' quest is moving slower than the player character, so you always have to wait till it catched up. Another problem is, that the pet stays in combat after the mobs you are fighted at is already dead. I only tried that with a troll hunter but I think it will happen to all other classes too.

### Expected behaviour

The pet should be as fast as the player character and should stop being in combat after all enemy mobs are dead.

### Steps to reproduce the problem

1. Accept the quest 'Taming the Beast' with your hunter.
2. Tame a mob
3. Run around and get into fight with your previously tamed pet

### Branch

3.3.5

### TC rev. hash/commit

TrinityCore rev. 8938f383572a 2024-08-17 22:27:45 +0000 (3.3.5 branch) (Unix, RelWithDebInfo, Dynamic)
Using SSL version: OpenSSL 3.0.13 30 Jan 2024 (library: OpenSSL 3.0.13 30 Jan 2024)
Using Boost version: 1.74.0
Using MySQL version: 101106 (I use the MariaDB 10.11.6)
Using CMake version: 3.25.1
Compiled on: Linux 6.6.31+rpt-rpi-v8
Automatic database updates are enabled for the following databases: Auth, Characters, World
Worldserver listening connections on port {}
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in ../data/maps. Total size: 291014951 bytes
vmaps directory located in ../data/vmaps. Total size: 657797442 bytes
mmaps directory located in ../data/mmaps. Total size: 1970895648 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS 
Using World DB: TDB 335.24081
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0

### Operating system

Debian 12 (Raspberry Pi OS)

### Custom changes

None

CraftedRO:
4a6b73782beb6981a6ea48dd0da34303d5af3edb seems still valid, after taming any beast with quest item the pet receive * UNIT_FLAG_IN_COMBAT (0x80000) for some reason and indeed the tamed creature speed looks slightly lower than the owner

PR #30918, opened 2025-05-07
Core/Unit: Force leaving combat on temporary charmed unit

CraftedRO:
**Changes proposed:**

-  Workaround to force leaving combat on temporary charmed unit
-  It still doesn't properly reset and return near owner after killing an enemy but as I understand that's because it's not really charmed as a real pet so HandleReturnMovement won't fire as it shoud but I guess is a bit better than permanently in combat.
-  

**Issues addressed:**

Updates: https://github.com/TrinityCore/TrinityCore/issues/30169

**Tests performed:**

- [ None as I know ]


CraftedRO:
looks like this have to be fixed in a more generic way since same issue occur even when enslave a minion with warlock, for some reason pet keep combat state even after charmed...

CraftedRO:
I think this is all that I can do in this case, if nothing break shoud be better than nothing...

PR #30947, opened 2025-05-14
Scripts/Redridge: Implement Quest "Saving Foreman Oslow"

jasongdove:
**Changes proposed:**

- Script quest 26520 "Saving Foreman Oslow"

**Issues addressed:**

None

**Tests performed:**

Tested in-game

- [ ] Exit vehicle (vehicle seat addon?) doesn't work properly, so ettin keeps holding rock at the end of the script


TQPS:
Thank you for looking into this one.

src/server/scripts/EasternKingdoms/zone_redridge_mountains.cpp
```diff
@@ -113,147 +138,612 @@ class npc_big_earl : public CreatureScript
...
+            while (uint32 eventId = _events.ExecuteEvent())
             {
-                while (uint32 eventId = _events.ExecuteEvent())
+                switch (eventId)
                 {
-                    switch (eventId)
-                    {
                     case EVENT_STORE_GUIDS:
-                        if (Creature* oslow = me->FindNearestCreature(NPC_OSLOW, 5.0f, true))
+                        if (Creature const* oslow = me->FindNearestCreatureWithOptions(10.0f,  { .StringId = "npc_redridge_foreman_oslow_under_rock"sv }))
```
mdX7, line 241: is the stringid here required? In general we only use stringids if we cannot get the creature via entry only (e.g. bcuz the same entry is spawned 10 times and we want to use a specific one)

src/server/scripts/EasternKingdoms/zone_redridge_mountains.cpp
```diff
@@ -113,147 +138,612 @@ class npc_big_earl : public CreatureScript
...
+                        if (Creature* boulder = ObjectAccessor::GetCreature(*me, _boulderGUID))
+                            boulder->AI()->SetData(0, DATA_DONE);
+
+                        std::function<void(Movement::MoveSplineInit&)> initializer = [](Movement::MoveSplineInit& init)
+                        {
+                            Movement::PointsArray path(std::size(EttinPathUpHill));
+                            std::ranges::transform(EttinPathUpHill, path.begin(), [](Position const pos) { return PositionToVector3(pos); });
+                            init.MovebyPath(path, 0);
+                        };
+                        me->GetMotionMaster()->LaunchMoveSpline(std::move(initializer), POINT_UP_PATH, MOTION_PRIORITY_NORMAL, POINT_MOTION_TYPE);
```
mdX7, line 614: we could move this path into db `waypoint_path*` tables, then we just gotta do `me->GetMotionMaster()->MovePath(<pathid from db>);` and can remove any other operations done on the path in this file

src/server/scripts/EasternKingdoms/zone_redridge_mountains.cpp
```diff
@@ -113,147 +138,612 @@ class npc_big_earl : public CreatureScript
...
+    }
+};
+
+// 80739 - Lift Huge Boulder
+class spell_redridge_lift_huge_boulder : public SpellScript
+{
+    void HandleScriptEffect(SpellEffIndex /*effIndex*/) const
+    {
+        if (Creature* target = GetHitCreature())
+            target->CastSpell(GetCaster(), static_cast<uint32>(GetSpellInfo()->GetEffect(EFFECT_0).BasePoints), false);
```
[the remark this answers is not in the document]
mdX7: If any aura is applied we always use CastSpell in scripts to apply it, so that should be fine.
Usually only serverside spells tend to skip SPELL_GO in sniff. But in that case we'd use CastSpell aswell

src/server/scripts/EasternKingdoms/zone_redridge_mountains.cpp
```diff
@@ -23,16 +23,47 @@ Script Data End */
...
+#include <algorithm>
 
-enum DumpyKeeshan
+enum RedridgeSpellData
 {
-    NPC_BIGEARL                    = 43248,
+    SPELL_APPLY_QUEST_INVIS_ZONE_3  = 80815, // Used by npc's in Lakeshire Inn
+    SPELL_APPLY_QUEST_INVIS_ZONE_4  = 80816, // Used by npc's in Lakeshire Inn
+
+    SPELL_APPLY_QUEST_INVIS_ZONE_19 = 82099,  // Used by npc's in Lakeshire Townhall
```
Naddley, line 39: You can remove those enums, since there unused in your code.
This detect spells needs to be handled in spell_area table.

jasongdove: They were moved from other existing code in this file to keep all the spells together. Should I have left them separate?

Naddley: hmm I see. Then keep it for now. We need to remove it after rework of the other script then.

mdX7 (approved):

PR #30949, opened 2025-05-15
Scripts/Mardum: Script Fel Lord Caza for quest Hidden no more

aquadeus:
**Changes proposed:**

-  Script Fel Lord Caza for quest Hidden no more

**Tests performed:**

Tested in-game

- [x] Spell 210104 - is not completing the quest
- [ ] Spell 210105 - should be allowed to cast during a knockback like retail, which is not happening atm


src/server/scripts/BrokenIsles/zone_mardum.cpp
```diff
@@ -1969,6 +1969,139 @@ CreatureAI* DemonHunterHiddenNoMoreAISelector(Creature* creature)
...
+    }
+};
+
+// 196889 - Throw Axe
+class spell_fel_lord_caza_throwing_axe_selector : public SpellScript
+{
+    void HandleHit(SpellEffIndex /*effIndex*/)
+    {
+        Unit* caster = GetCaster();
+        caster->CastSpell(GetHitUnit()->GetPosition(), SPELL_FEL_LORD_CAZA_THROW_AXE_MISSILE, TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR);
```
mdX7, line 2096: you can `GetCaster()` directly here, no need to declare it as variable

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -5092,6 +5092,15 @@ void SpellMgr::LoadSpellInfoCorrections()
         });
     });
 
+    // Throw Axe
+    ApplySpellFix({ 196876 }, [](SpellInfo* spellInfo)
+    {
+        ApplySpellEffectFix(spellInfo, EFFECT_2, [](SpellEffectInfo* spellEffectInfo)
+        {
+            spellEffectInfo->Effect = SPELL_EFFECT_NONE;
```
mdX7, line 5100: i dont like this, are we sure 196959 is not casted anywhere?

aquadeus: yes, its only used with this npc but the triggered spell is not on sniffs, so i prevent the spam on the console for the non existing areatrigger

mdX7 (approved):

PR #30952, opened 2025-05-16
Scripts/Monk: Implement talent Song of Chi-ji

aquadeus:
**Changes proposed:**

-  Implement talent Song of Chi-ji

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_monk.cpp
```diff
@@ -432,6 +436,42 @@ class spell_monk_save_them_all : public AuraScript
...
+    {
+        SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(at->GetSpellId(), DIFFICULTY_NONE);
+        if (!spellInfo)
+            return;
+
+        Unit* caster = at->GetCaster();
+        if (!caster)
+            return;
+
+        Position destPos = at->GetFirstCollisionPosition(40.0f, 0.0f);
```
Nyr97, line 454: hardcoded distance, use the spell's max distance instead, which is how far it travels

PR #30953, opened 2025-05-17
Core/Procs: Implement PROC_FLAG_2_SUCCESSFUL_DISPEL

aquadeus:
**Changes proposed:**

-  Implement PROC_FLAG_2_SUCCESSFUL_DISPEL

**Tests performed:**

Tested in-game


Shauren (changes requested):
Seems wrong to me, it should fire procs for actor, not action target

aquadeus:
> Seems wrong to me, it should fire procs for actor, not action target

ok done

Shauren (changes requested):
This is not at all what I said that needs to be done, swap proc flags argument order

Issue #30447, opened 2024-11-17
Core/Creature: Wrong rounded health value from HealthModifier when regen is disabled


CraftedRO:
### Description

Probably really cosmetic issue and almost no relevant but well.
Smith Argus <Blacksmithing Trainer> is a npc in Goldshire, in  db / curhealth it does have 664 health but ingame 665, now probably  1 hp is not a big deal but it's a bit strange.

### Expected behaviour

Ingame it shoud also have 664 health value as can be seen here:

![Captură de ecran 2024-11-17 193415](https://github.com/user-attachments/assets/221890a6-4f62-48be-b2a4-b94475bb78f6)

https://www.youtube.com/watch?v=o3Aht1CAqZY
https://www.wowhead.com/wotlk/npc=514/smith-argus

### Steps to reproduce the problem

.go c id 514
.npc info
notice health

### Branch

3.3.5

### TC rev. hash/commit

32b54ffb66676ec47d7153c1b897770bc5d08493

### Operating system

Windows 11 x64

### Custom changes

None

CraftedRO:
For some reason any value we set in db/curhealth for this npc is ignored.

Shauren:
This column is used only if health regen is disabled

CraftedRO:
so shoud we disable that, then ?

Shauren:
No, this is probably due to rounding (HealthModifier is 1.02 in creature_template)

PR #30971, opened 2025-05-21
Core/Creature: Update creature GenerateHealth rounding formula

CraftedRO:
**Changes proposed:**

-  Attempt to fix creature GenerateHealth rounding issue
-  Fix based on https://github.com/cmangos/mangos-classic/commit/8ee39cb664f3ef8689576ed639831517f940b3cb (that shoud be ported too most likely)

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/30447

**Tests performed:**

- [ Maybe few more tests, ]

### DIFF
--- a/src/server/game/Entities/Creature/CreatureData.h
+++ b/src/server/game/Entities/Creature/CreatureData.h
@@ -403,7 +403,8 @@ struct TC_GAME_API CreatureBaseStats
 
     uint32 GenerateHealth(CreatureTemplate const* info) const
     {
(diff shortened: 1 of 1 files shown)


Shauren (changes requested):
This breaks creatures whose health would calculate between 0 and 1 (returning 0)

CraftedRO:
I think I found a variant that can be simplified, but I'll let you to do that if protection against minvalue not needed.

Issue #29706, opened 2024-02-16
If you leave party/raid while "you will be teleported to nearest gy" timer is onscreen you will be teleported to your HS location


CraftedRO:
### Description

If player logout when "You are not in this instance, you will be teleported to nearest gy" timer is onscreen, when he log back in he will be teleported to his hearthstone location (it works even as ghost).
I don't think this is the blizzlike behaviour.

https://youtu.be/piN-XLFaLMw

Players are using this as a second hearthstone :D

### Expected behaviour

Most likely player shoud be teleported somewhere outside of the instance, maybe at graveyard as stated in message.

### Steps to reproduce the problem

creater a party or raid,
enter inside a instance
leave party
when "You are not in this instance, you will be teleported to nearest gy" timer is onscreen just press logout 
notice

### Branch

3.3.5

### TC rev. hash/commit

86551ba71669712cadf82f2dc1f0c5481f0fba1b

### Operating system

Windows 11 x64

### Custom changes

[Passive anticheat](https://github.com/TrinityCore/TrinityCoreCustomChanges/commits/3.3.5-passive_anticheat)

CraftedRO:
iirc we tested on dungeons and you have to be teleported at the dungeon entrance instead, we forgot to test on raids.

CraftedRO:
oh, I even found the video, I guess will try to fix this.

PR #30982, opened 2025-05-23
Core/Misc: Properly teleport player outside of the instance

CraftedRO:
**Changes proposed:**

-  Attempt to fix described targeted issue and make this blizzlike as you can see in following retail cata classic test:
-  https://youtu.be/oLesj9H1D8Q
-  

**Issues addressed:**

Closes https://github.com/TrinityCore/TrinityCore/issues/29706

**Tests performed:**

- [ More tests ]

### DIFF
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -523,9 +523,9 @@ void WorldSession::LogoutPlayer(bool save)
         if (Battleground* bg = _player->GetBattleground())
             bg->EventPlayerLoggedOut(_player);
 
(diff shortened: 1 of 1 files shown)


CraftedRO:
most likely this check is not needed.

Shauren (approved):

CraftedRO:
while this change was needed and is good, after watching more times recoded video, looks like the player remain in instance probably like it does the .gm on in tc but that's another thing/behaviour maybe even complex so it will keep instance lock maybe and don't get kicket at all, tbd.

Shauren:
Logging in also checks if player is allowed to be in the instance (technically this could just create a new instance)

PR #31001, opened 2025-05-26
Core/Misc: Fix player outdoor status after leaving an instance

jasongdove:
**Changes proposed:**

- A regression to player indoor/outdoor status was introduced with https://github.com/TrinityCore/TrinityCore/pull/28731
- After that change, players may have to move after zoning out of an instance before they are able to mount
- This fixes that case, and potentially others; I'm not familiar enough with the teleport semantics to understand it all, but this looks similar to `Player::LoadFromDB` where `UpdatePositionData` follows `SetMap`, so I think it's an appropriate place to add this call.

**Issues addressed:**

None.

**Tests performed:**

Tested in-game.

1. `.tele ragefirechasm`
2. Zone into instance
3. Zone out of instance, but do not move
4. Attempt to mount

None.

### DIFF
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -107,6 +107,7 @@ void WorldSession::HandleMoveWorldportAck()
 
     player->ResetMap();
     player->SetMap(newMap);
+    player->UpdatePositionData();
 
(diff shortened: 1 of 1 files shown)


CraftedRO:
judging by the code similarity, both pr might fit to 3.3.5 too (if they do a difference)

PR #31013, opened 2025-05-29
Scripts/Scarlet Monastery: Move Scarlet Trainees to summon_groups

CraftedRO:
**Changes proposed:**

-  Based on my cata ptr tests and this classic video https://youtu.be/ZEIjJIZH5I0?t=644 after you kill [boss Herod](https://www.wowhead.com/wotlk/npc=3975/herod), [Scarlet Trainees](https://www.wowhead.com/wotlk/npc=6575/scarlet-trainee) should not despawn instant after reaching end path point when not in combat but rather after a while (so players will still be able to loot them for example after wipe).

**Issues addressed:**

Closes None

**Tests performed:**

- [ None as I know. ]


CraftedRO:
> The trainees should be moved to spawn groups and spawned once after herod's boss state is set to DONE.

I was afraid you woud appear here too :D

CraftedRO:
hmm, if we perma spawn them in db and then add them to spawn_group they will still be spawned since the beginning (no matter what we set in instance_spawn_groups, in an master example I found these are prevented to be spawned via CONDITION_SOURCE_TYPE_SPAWN_GROUP (33) conditions that are nyi on 3.3.5, am I wrong ?

CraftedRO:
nvm, it works but strange af

CraftedRO:
for some reason at actually herod core script, the spawn group condition in db only triggers correctly (at the boss state) when I set bossStates to 4 (special) wich is strange since we woud expect to trigger on 3.

Shauren:
Anyway I fully disagree with the spawn group suggestion specifically for this script - those trainees should only spawn exactly on death, not any time the instance is reloaded if boss is dead

CraftedRO:
shoud trainees c++ script+script_waypoint remain or removed and add some SAI after moving to creature_summon_groups ?

CraftedRO:
Updated with sniffed positions, I think this is all I can do in this case.

CraftedRO:
> > -- I don't think this ever had any waypoints, but only movementId in creature_template
> 
> movementId holds speed of creature (NYI), not id of path or something

Well, that's strange because in my tests they came down even without forcing them with SAI

Nyr97:
sus, because those creatures are not engaged in combat with you until you manually engage them, so there's no way these are moving without being forced.

it is likely they have splines in ObjectCreate packet, which usually do not result in a MONSTER_MOVE packet.

PR #31015, opened 2025-05-31
Spells/Warrior: Fix intimidating shout's menace talent

Golrag:
**Changes proposed:**

- Added root effect after knock back
- Fixed target selection
- fixed a console error
- Disable aoe root on normal fear

**Issues addressed:**

Closes: none

**Tests performed:**

- Builds locally
- Feared around some NPC with and without talent

- None


src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -482,15 +484,51 @@ class spell_warr_impending_victory : public SpellScript
...
+    }
+    void FilterTargets(std::list<WorldObject*>& unitList) const
+    {
+        unitList.remove(GetExplTargetWorldObject());
+    }
+
+    void HandleRoot(SpellEffIndex /*effIndex*/) const
+    {
+        CastSpellExtraArgs const args(GetSpell());
+        GetCaster()->m_Events.AddEventAtOffset([caster = GetCaster(), targetGUID = GetHitUnit()->GetGUID(), castArgs = args]()
```
Nyr97, line 521: Should we be making a copy of CastSpellExtraArgs here? I assume it's to avoid lifetime issues since the lambda is delayed, but I'm unsure whether CastSpellExtraArgs is temporary or not.

Nyr97: You can, however, castArgs = CastSpellExtraArgs(GetSpell()) directly on the lambda, there's no need for it to be outside.

mdX7 (approved):

Issue #30643, opened 2025-01-26
Spell/Warrior: Bloodthirst restoring 3% of your health


Zaekor:
### Description

Bloodthirst applies damage, but not restoring 3% of your health.

### Expected behaviour

Bloodthirst should apply damage and restoring 3% of your health.

### Steps to reproduce the problem

.1 Create a warrior
.2 Cast the [Bloodthirst](https://www.wowhead.com/spell=23881/bloodthirst#changelog) spell on a target

### Branch

master

### TC rev. hash/commit

TrinityCore rev. c85d12fc3f05 2025-01-25 22:48:10 +0100 (master branch) (Windows, AMD64, RelWithDebInfo, Static) (worldserver-daemon)

### Operating system

Windows 10

### Custom changes

_I don't know if this is the correct way to do it, this is my first contribution, please excuse me if this is not the right way._

### Editing spell_warrior.cpp

```
// 23881 - Bloodthirst
class spell_warr_bloodthirst : public SpellScript
{
    bool Validate(SpellInfo const* /*spellInfo*/) override
    {
        return ValidateSpellInfo({ SPELL_WARRIOR_BLOODTHIRST_HEAL });
    }

    void HandleHit()
    {
        Unit* caster = GetCaster();
        CastSpellExtraArgs args(TRIGGERED_FULL_MASK);
        GetCaster()->CastSpell(GetCaster(), SPELL_WARRIOR_BLOODTHIRST_HEAL, args);
    }

    void Register() override
    {
        AfterHit += SpellHitFn(spell_warr_bloodthirst::HandleHit);
    }
};
```

Nyr97:
Fixes for core changes are done through PRs, not this way. Also we don't set TRIGGERED_FULL_MASK anymore, we only set appropriate flags for the spell being cast.

PR #31027, opened 2025-06-03
Scripts/Spells: Fix bloodthirst heal

### LINKED ISSUE #30117: bloodthirst spell bugged
### Description

https://www.wowhead.com/spell=23881/bloodthirst 
using this ability does not heal the warrior back

### Expected behaviour

should heal 3% of your hp back

use spell on any enemy while low on hp 

### Branch

master

### TC rev. hash/commit

eb572d823ba3 2024-07-16 19:13:27

### Operating system

win 10

### Custom changes

none

Golrag:
**Changes proposed:**

-  Fix bloodthirst heal

**Issues addressed:**

Closes #30117, #30643

**Tests performed:**

- Builds locally
- heal is cast

- none


Golrag:
Converted back to a draft as it's not suppose to trigger when you hit an immune target (not sure about dodge/miss/parry). it only looks at main target, cleave targets don't count.

Nyr97:
AfterCast completely ignores whether the spell was successful (meaning that, for as long as targets is > 0, it can be dodged, parried, blocked, immuned, and still trigger whatever is inside the hook, but just once, not per target).

OnEffectHitTarget can be used to prevent an effect from applying on whatever conditions you choose. I can't remember if this spell was able to ever cleave, I don't think so; if it is the case, you can simply check whether target is explicit to trigger it if you don't want it to trigger it per target count.

mdX7 (approved):

PR #31034, opened 2025-06-04
Core/Instances: Support custom area triggers when finding instance entrance and start

jasongdove:
**Changes proposed:**

- Adds `_teleportAreaTriggerStore` to use when searching for go back/entrance locations, populated by `areatrigger_teleport` and by `areatrigger` spawns that have a teleport action
- This fixes `.go instance` for scripted instances that use custom area triggers to zone in and out

**Issues addressed:**

None.

**Tests performed:**

Tested in-game using Atal'Dazar

1. Initially, `.go instance atal` will teleport inside entrance
2. Fix parent map in `instance_template` and confirm that `.go instance atal` now teleports outside instance:

```sql
UPDATE `instance_template` SET `parent` = 1642 WHERE `map` = 1763;
```

None.


jasongdove:
Maybe it would be better to keep `areatrigger_teleport` as is, and add some logic to check spawned area triggers in `ObjectMgr::GetGoBackTrigger` and `ObjectMgr::GetMapEntranceTrigger`?

EDIT: Updated to remove all SQL changes.

jasongdove:
The area trigger ids (DB2, custom, not custom) are not used at all here - `AreaTriggerStruct` just holds info about source map, target map and target position.

I just renamed `_areaTriggerStore` because it seemed kind of ambiguous given all of the different types of area triggers that the core now has.

src/server/game/Globals/AreaTriggerDataStore.cpp
```diff
@@ -409,6 +409,26 @@ void AreaTriggerDataStore::LoadAreaTriggerSpawns()
             spawn.scriptId = sObjectMgr->GetScriptId(fields[12].GetString());
             spawn.spawnGroupData = sObjectMgr->GetLegacySpawnGroup();
 
+            // If the trigger has a teleport action, use it for checking go back/map entrance locations
+            auto teleportActionItr = std::ranges::find_if(createProperties->Template->Actions, [](AreaTriggerAction const& action) { return action.ActionType == AREATRIGGER_ACTION_TELEPORT; });
```
mdX7, line 413: maybe we should also explicitly check if destination map is != areatrigger mapid in case there are areatriggers teleporting within the same mapid

jasongdove: Yeah, if we want to continue to implicitly find these locations (vs having a list of world safe locs related to dungeon entrances/starts), it would be good to filter those out. I can make that change.

jasongdove:
I don't think continuing to search area triggers for world safe locs related to instances makes sense; it's too flaky - e.g. instances that have two exit locations (Deadmines) will sometimes pick one or the other, and there isn't a way to specify which one is appropriate. Also, some instances (Darkmaul Citadel) have no entrance area triggers.

Closing this and starting on a different plan to refactor and make things more deterministic.

Shauren:
Personally I would not bother fixing up that specific chat command, a better alternative serving "go to a location with this name" purpose already exists, and that is `.tele`

jasongdove:
That's fair - I was just pointing out another use of the area triggers that is easily tested by a single player. The primary problem is that LFG will not load dungeons that have no (DBC) area triggers https://github.com/TrinityCore/TrinityCore/blob/39c06d40e26ca7c1651e266cde16e031066610f5/src/server/game/DungeonFinding/LFGMgr.cpp#L259-L264

and dungeon exit triggers are used here

https://github.com/TrinityCore/TrinityCore/blob/39c06d40e26ca7c1651e266cde16e031066610f5/src/server/game/Entities/Player/Player.cpp#L18295

I think removing the indirection of checking area triggers (and having the world safe locs referenced in something like instance_template) would make more sense for instances.

jasongdove:
`lfg_dungeon_template` could be used to fix entrances, but it wouldn't (currently) help with teleporting out of an instance for any reason. And the functions are not limited to LFG - they're also used more generally with instances in that player section I linked.

jasongdove:
This is how the two functions appear to be used:

`ObjectMgr::GetMapEntranceTrigger` (find starting location inside instance)
	- `LFGMgr::LoadLFGDungeons` when `lfg_dungeon_template` record does not exist
	- `Player::LoadFromDB` when instance has been reset
	- `go_commandscript::HandleGoInstanceCommand`

`ObjectMgr::GetGoBackTrigger` (find location outside instance entrance)
	- `Player::LoadFromDB` when ??? (map fails to create?)
	- `Player::LoadFromDB` when dungeon encounter is in progress
	- `WorldSession::HandlePlayerLogin` when player cannot be added to map (bad coordinates?)
	- `go_commandscript::HandleGoInstanceCommand`

PR #31035, opened 2025-06-06
Core/AreaTriggers: Add AREATRIGGER_ACTION_TAVERN

jasongdove:
**Changes proposed:**

- Add `AREATRIGGER_ACTION_TAVERN` to support server-side tavern area triggers
- SQL includes an example tavern at Illidari Stand in Azsuna

**Issues addressed:**

None.

**Tests performed:**

Tested in game; client and server area triggers can both be used as taverns.

None.


src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
```diff
@@ -263,6 +264,9 @@ bool AreaTrigger::Create(AreaTriggerCreatePropertiesId areaTriggerCreateProperti
         }
     }
 
+    if (IsStaticSpawn())
+        setActive(true);
```
jasongdove: I'm not sure about this part specifically, but it resolved an issue I ran into where server area triggers keep players in their target list after logout. Because they were still in the target list, `DoActions` was not called on login, and the rest state update field (loaded from db) became out of sync with the player rest flags.

With this here, the player is removed from the area trigger target list on logout, and properly gets the rest flag as a result of the `DoActions` call on login.

Nyr97: this is bad, the underlying issue is how action targets are cleared for players logging out, that should be fixed instead.

it is not recommended to ever set objects as active, as they keep the grid loaded wherever they're added, and given that there are over 100+ tavern locations, this would stress the server unnecessarily, especially since you're setting all static ones as such, not those using the flag.

Nyr97 (commented):
comment

Shauren:
I have a comment about both this PR and #31034 - you assumed that there are two distinct areatrigger id sequences. This is incorrect, there are THREE sequences:
* AreaTrigger.db2
* areatrigger_template WHERE IsCustom = FALSE
* areatrigger_template WHERE IsCustom = TRUE

src/server/game/Entities/AreaTrigger/AreaTrigger.cpp
jasongdove: Okay, I suspected it wasn't an ideal solution. Do you have any suggestions for resolving the issue I outlined with players logging out and staying in the area trigger `_insideUnits`?

Shauren: setActive, but only while there are units inside

jasongdove: I think I have this updated to remain active only while units are inside.

jasongdove:
Thanks for the clarification about the ids, that's good to know. For this feature, though, I think we only care about DBC (we also check during player movement) vs not DBC (no extra checking needed). Updated to reflect this.

Nyr97:
As far as I am aware, tavern areatriggers are not handled clientside anymore, and it's been like that for ages now (they're not part of AreaTrigger.db2), isn't that right?

jasongdove:
At some point new taverns were no longer added to DB2, but the client still sends CMSG_AREA_TRIGGER for enter/exit existing taverns (e.g. 4265 fairbreeze village).

Nyr97:
Oh in that case the distinction is nice to keep.

PR #31038, opened 2025-06-06
Scripts/Spells: Implement Raging Blow Cooldown Reset talents

Golrag:
**Changes proposed:**

- Implement Improved Raging Blow
- Implement Wrath and Fury

**Issues addressed:**

Closes: none

**Tests performed:**

- Builds locally
- Cooldown Reduction applied
- On retail, crushing blow tootlips suggests that the 25% cooldown reduction is always present. But after testing on dummies, that doesnt seem to be the case. I assume blizzard assumes you took the talent. Either that, or I was unlucky.

- none


src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -548,6 +551,41 @@ class spell_warr_mortal_strike : public SpellScript
...
+class spell_warr_raging_blow_cooldown_reset : public SpellScript
+{
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_WARRIOR_IMPROVED_RAGING_BLOW })
+            && ValidateSpellEffect({{ SPELL_WARRIOR_WRATH_AND_FURY, EFFECT_0 }});
+    }
+
+    bool Load() override
+    {
```
Nyr97, line 652: it's better to point to the caster once and then use it than pointing to it twice (much more efficient and less redundant)

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -548,6 +551,41 @@ class spell_warr_mortal_strike : public SpellScript
...
+        return GetCaster()->HasAura(SPELL_WARRIOR_IMPROVED_RAGING_BLOW) || GetCaster()->HasAura(SPELL_WARRIOR_WRATH_AND_FURY);
+    }
+
+    void HandleResetCooldown(SpellEffIndex /*effIndex*/) const
+    {
+        // it is currently impossible to have Wrath and Fury without having Improved Raging Blow, but we will check it anyway
```
Nyr97, line 659: same applies here, there's no need to get caster 4 times, just point to it once and use it.

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -548,6 +551,41 @@ class spell_warr_mortal_strike : public SpellScript
...
+        int32 value = 0;
+        if (GetCaster()->HasAura(SPELL_WARRIOR_IMPROVED_RAGING_BLOW))
+            value = GetEffectValue();
+
+        if (AuraEffect const* auraEffect = GetCaster()->GetAuraEffect(SPELL_WARRIOR_WRATH_AND_FURY, EFFECT_0, GetCaster()->GetGUID()))
```
Nyr97: is it necessary to check the caster guid here? can the buff be applied by a different warrior onto you?

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -548,6 +551,41 @@ class spell_warr_mortal_strike : public SpellScript
...
+            value += auraEffect->GetAmount();
+
+        if (roll_chance_i(value))
+            GetCaster()->GetSpellHistory()->ResetCooldown(GetSpellInfo()->Id, true);
```
Nyr97: this works for spells that do not use charges, is any of the spells being attached using charges? cause i see both of them using ChargeCategory 1672

Nyr97:
As a rule of thumb, Blizzard now directly sets category charges with count 1 on spells if any aura coming from traits or tier sets is supposed to add charges to the affected spell. I do remember there were cases before where the non-affected spell was using cooldown, and if you had a certain aura from a talent trait, it would modify the spell (or replace it directly with a different id) with one with categoryCharge.

I don't remember if it's the case anymore, but the standard is what I mentioned above. It might be useful to implement a ResetCooldownOrCharge feature instead in the future so that we first check for any charge otherwise default to cooldown.

Aokromes:
Use the [web editor](https://github.com/TrinityCore/TrinityCore/pull/31038/conflicts) or the command line to resolve conflicts before continuing.

src/server/scripts/Spells/spell_warrior.cpp

PR #31053, opened 2025-06-09
Scripts/Spells: Implement Fresh Meat talent

Golrag:
**Changes proposed:**

-  implement Fresh meat talent

**Issues addressed:**

Closes: none

**Tests performed:**

- Builds locally
- Bloodthirst applies fresh meat debuff to target
- Also enrages if target doesn't have said debuff

- [x] I'm not sure if it should be handled by the enrage proc intead


src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -422,6 +424,46 @@ class spell_warr_frenzied_enrage : public SpellScript
...
+
+    void ApplyFreshMeat(SpellEffIndex /*effIndex*/) const
+    {
+        Unit* target = GetHitUnit();
+        if (target != GetExplTargetUnit())
+            return;
+
+        Unit* caster = GetCaster();
+
+        if (target->HasAura(SPELL_WARRIOR_FRESH_MEAT_DEBUFF, caster->GetGUID()))
```
Nyr97: currently, the implementation falls short (and partially wrong implementation), it reads: "Bloodthirst always Enrages you the first time you strike a target and it has a 15% increased chance to trigger Enrage."

the debuff itself is infinite, so this script won't ever trigger again against the same target unless it's removed manually (by death also works since I tested it), what if both of you stop combat and engage again, does it work? if it were the case, maybe adding an aura script to it to be manually removed upon losing combat with the warrior caster since it also lacks AuraInterruptFlags.

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -422,6 +424,46 @@ class spell_warr_frenzied_enrage : public SpellScript
...
+            return;
+
+        caster->CastSpell(target, SPELL_WARRIOR_FRESH_MEAT_DEBUFF, CastSpellExtraArgsInit{
+            .TriggerFlags = TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR
+        });
+
+        caster->CastSpell(nullptr, SPELL_WARRIOR_ENRAGE, CastSpellExtraArgsInit{
+            .TriggerFlags = TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR
```
Nyr97: also, spell 215568 applies SPELL_AURA_ADD_FLAT_MODIFIER at index 12 (PointsIndex1, so effect_1) for 184361 - Enrage spell of value 15; it's the % increased chance for Bloodthirsty spell, flat (so it just adds 15 to that 30%, becoming 45%, without any other modifiers).

the enrage trigger should be handled by spell proc data and an AuraScript with DoCheckProc hook, where Bloodthirsty calculates from effect 1 and Rampage always triggers it.

this forced trigger cast is just wrong.

Golrag: This modifier is already handled in https://github.com/TrinityCore/TrinityCore/commit/35214519ce12064223fb6c39fbdc2519db4569d3

src/server/scripts/Spells/spell_warrior.cpp
Golrag: oh right, i forgot about aura removal

src/server/scripts/Spells/spell_warrior.cpp
Nyr97: oh in that case, just get rid of this and it should be ok.

src/server/scripts/Spells/spell_warrior.cpp
Nyr97: i suggest testing this on retail, i don't know what the conditions are for this aura to be removed (besides death), AuraInterruptFlags usually help identifying those conditions, but this one lacks them

Nyr97: As it currently stands, the implementation now can trigger Enrage twice, both from this spellscript and from spell proc system based on Enrage spellscript's roll_chance_i(effect value) and it feels wrong. 

I would move the implementation here: https://github.com/TrinityCore/TrinityCore/commit/35214519ce12064223fb6c39fbdc2519db4569d3#diff-ebd199edd79a601267532397be4e524ee08979dc7be563b821647246b091d87dR361

where you do eventInfo.GetSpell and retrieve the spell target and then the debuff and check whether debuff not present OR chance

Shauren: Not relevant. This aura has 100% chance so it doesn't matter that the other one can also proc it

Golrag: aura removal has been added.

if we go with the proc route, I think we would need to edit the current proc data to have target info in ProcEventInfo

Nyr97: that'd be easy to add, Shauren just needs to verify on sniff if Enrage ever happens to trigger twice by the generic trigger and the talent, which I argued it shouldn't (not that it's a fact) since there are effects that trigger when Enrage does

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -424,6 +426,61 @@ class spell_warr_frenzied_enrage : public SpellScript
...
+    }
+};
+
+// 316044 - Fresh Meat
+class spell_warr_fresh_meat_aura_dummy : public AuraScript
+{
+    void OnOwnerOutOfCombat(bool isNowInCombat) const
+    {
+        if (!isNowInCombat)
+            GetTarget()->RemoveAura(GetAura());
```
mdX7:
```suggestion
            Remove();
```

src/server/scripts/Spells/spell_warrior.cpp
Golrag: enrage twice has been fixed, fresh meat completely handled in the enrage_proc script

Shauren:
Removed removal handling - after a bit of retail testing (and aura_update sniffing) it turns out its never removed except in cases that would natually remove it such as evade or death (even in pvp only death removes it)

PR #31056, opened 2025-06-10
Scripts/Spells: Implement Cold Steel, Hot Blood talent

Golrag:
**Changes proposed:**

-  Implement Coldl Steel, Hot Blood talent

**Issues addressed:**

Closes: none?

**Tests performed:**

- Builds locally
- Crits on Bloodthirst/Bloodbath give debuff

- [ ] It seems like only one target can get the debuff, while all targets on whom the crit occured should have the debuff.  
Retail behaviour: While more targets can get crit at the same time, the extra rage generated is only cast with one spell_go (no spell_start). But the rage gained is still summed up (so 2 targets critted, means double rage gained).


Golrag:
leaving this here so i dont forget:
Enrage **doesn't** trigger from bloodthirst when target is immune (even with fresh meat)
Enrage **does** trigger from rampage when the target is immune

Shauren:
Tested on retail - rage gain is always 4, no matter how many crits happened (matches expected behavior, proc aura has 500ms cooldown)

Casting 385042 should probably be a script on bloodthirst in this case

PR #31057, opened 2025-06-10
Scripts/Spells: Implement shaman talent Deluge

aquadeus:
**Changes proposed:**

-  Implement shaman talent Deluge

**Tests performed:**

Tested in-game


src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -585,6 +587,52 @@ class spell_sha_crash_lightning : public SpellScript
...
+    void CalculateHealingBonus(SpellEffectInfo const& /*spellEffectInfo*/, Unit const* victim, int32& /*healing*/, int32& /*flatMod*/, float& pctMod) const
+    {
+        if (AuraEffect const* deluge = GetCaster()->GetAuraEffect(SPELL_SHAMAN_DELUGE_TALENT, EFFECT_0))
+            if (victim->GetAura(SPELL_SHAMAN_RIPTIDE, GetCaster()->GetGUID()) || victim->GetAura(SPELL_SHAMAN_DELUGE_AURA, GetCaster()->GetGUID()))
+                AddPct(pctMod, deluge->GetAmount());
+    }
+
+    void Register() override
+    {
+        CalcHealing += SpellCalcHealingFn(spell_sha_deluge::CalculateHealingBonus);
```
Nyr97, line 609: Implementation is wrong since you're currently double-dipping: 

200076 uses SPELL_AURA_ADD_FLAT_MODIFIER_BY_SPELL_LABEL as effect 1 affecting spells' misc 3 (PointsIndex0, value effect of healing) included in Label 1703 (Healing Wave, Healing Surge, Chain Heal) by +15%; this effect already adds the correct amount the very moment you have the aura, it's just the condition for X auras to be present is not there, you'd need to inverse the condition here:

- Load function where it passes only if effect 1 is present.
- if effect 1 is present but the target doesn't have caster's riptide AND doesn't have deluge aura, reduce pct by effect 1 amount (both 0 and 1 use the same value on BasePoints, but effect 0 is only for the tooltip whereas what carries the amount is the effect 1' misc).

if effect 1 isn't present, well, the mod isn't there, so we don't need to do anything.

src/server/scripts/Spells/spell_shaman.cpp
```diff
@@ -585,6 +587,52 @@ class spell_sha_crash_lightning : public SpellScript
     }
 };
 
+// 200076 - Deluge (attached to 77472 - Healing Wave, 8004 - Healing Surge and 1064 - Chain Heal
+class spell_sha_deluge : public SpellScript
+{
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_SHAMAN_RIPTIDE, SPELL_SHAMAN_DELUGE_AURA })
+            && ValidateSpellEffect({ { SPELL_SHAMAN_DELUGE_TALENT, EFFECT_0 } });
```
Nyr97, line 597: I'd also check if SPELL_SHAMAN_DELUGE_TALENT effect 1 is aura of type SPELL_AURA_ADD_FLAT_MODIFIER_BY_LABEL with miscB being 1703 (if they ever decide to change the aura type or the Label id to point to different spells, we'll know we have to detach them from spellIds you assigned this script to)

src/server/scripts/Spells/spell_shaman.cpp
Nyr97: also, you don't need to get caster 3 different times, point to 1 once and then use it

Shauren: Your analysis is incorrect, SPELL_AURA_ADD_FLAT_MODIFIER_BY_SPELL_LABEL only modifies points of 200075

PR #31064, opened 2025-06-10
Scripts/Spells: Implement Improved Whirlwind & Whirlwind target cap

Golrag:
**Changes proposed:**

-  Implement Improved Whirlwind talent (not the damage reduction part)
-  Add target cap info to whirlwind

**Issues addressed:**

Closes: none?

**Tests performed:**

- Builds locally
- Rage regen is correct
- buff only applies when targets are hit

- Out of scope: Apply the whirlwind cleave buff to Thunder Clap/Thunder Blast
- Out of scope: damage reduction on cleaved targets
- Might need some refactor before or after merge to apply the whirlwind cleave buff (currently used in multiple spells, soon another one)


src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -589,6 +591,58 @@ class spell_warr_impending_victory : public SpellScript
...
+    }
+
+    bool Load() override
+    {
+        return GetCaster()->HasAura(SPELL_WARRIOR_IMPROVED_WHIRLWIND);
+    }
+
+    void HandleHit(SpellEffIndex /*effIndex*/) const
+    {
+        int64 const targetsHit = GetUnitTargetCountForEffect(EFFECT_0);
```
Nyr97, line 726: that's a big ass variable though, what's the likelihood to ever surpass 100 targets? but I guess we're keeping the same type as the getter

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -589,6 +591,58 @@ class spell_warr_impending_victory : public SpellScript
...
+        if (!targetsHit)
```
Nyr97, line 727: it's also not recommended to check u/int types as bools, it's best to check against int type (<= 0 or whatever).

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -589,6 +591,58 @@ class spell_warr_impending_victory : public SpellScript
...
+            return;
+
+        Player* caster = GetCaster()->ToPlayer();
+        if (!caster)
+            return;
+
+        int32 const ragePerTarget = GetEffectValue();
```
Nyr97, line 734: it's also not totally required to write const for plain primitive types, it's useful to let others know that it won't ever mutate but

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -589,6 +591,58 @@ class spell_warr_impending_victory : public SpellScript
     }
 };
 
+// 12950 - Improved Whirlwind (attached to 190411 - Whirlwind)
+class spell_improved_whirlwind : public SpellScript
+{
+    bool Validate(SpellInfo const* spellInfo) override
+    {
+        return ValidateSpellInfo({ SPELL_WARRIOR_IMPROVED_WHIRLWIND, SPELL_WARRIOR_WHIRLWIND_CLEAVE_AURA })
+            && ValidateSpellEffect({ { spellInfo->Id, EFFECT_2 }, { SPELL_WARRIOR_WHIRLWIND_ENERGIZE, EFFECT_0 } });
```
Nyr97, line 716: I would check if SPELL_WARRIOR_WHIRLWIND_CLEAVE_AURA's effect 0 is aura of type SPELL_AURA_ADD_FLAT_MODIFIER and if the misc is 17 (ChainTargets) as well, it might change in future builds

src/server/scripts/Spells/spell_warrior.cpp
Nyr97: in any case, OnEffectHitTarget won't trigger if GetUnitTargetCountForEffect is below 1 though, not sure if it needs validation at all

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -589,6 +591,58 @@ class spell_warr_impending_victory : public SpellScript
...
+        if (!targetsHit)
+            return;
+
+        Player* caster = GetCaster()->ToPlayer();
+        if (!caster)
+            return;
+
+        int32 const ragePerTarget = GetEffectValue();
+        int32 const baseRage = GetEffectInfo(EFFECT_0).CalcValue();
+        int32 const maxRage = baseRage + (ragePerTarget * GetEffectInfo(EFFECT_2).CalcValue());
```
Nyr97, line 736: why is the sqrt effect value being used though? the tooltip states it's effect 1 value per target hit, there's no indication there's a max rage you can gain, unless you tested this

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -589,6 +591,58 @@ class spell_warr_impending_victory : public SpellScript
...
+
+        caster->CastSpell(nullptr, SPELL_WARRIOR_WHIRLWIND_ENERGIZE, CastSpellExtraArgsInit{
+            .TriggerFlags = TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_DONT_REPORT_CAST_ERROR,
+            .TriggeringSpell = GetSpell(),
+            .SpellValueOverrides = {{ SPELLVALUE_BASE_POINT0, rageGained * 10 } }
+        });
+
+        SpellInfo const* whirlwindCleaveAuraInfo = sSpellMgr->AssertSpellInfo(SPELL_WARRIOR_WHIRLWIND_CLEAVE_AURA, GetCastDifficulty());
+        int32 stackAmount = static_cast<int32>(whirlwindCleaveAuraInfo->StackAmount);
+        caster->ApplySpellMod(whirlwindCleaveAuraInfo, SpellModOp::MaxAuraStacks, stackAmount);
```
Nyr97: im guessing the use of ApplySpellMod is because there's no SPELLVALUE_AURA_MAX_STACK, right?

src/server/scripts/Spells/spell_warrior.cpp
Golrag: It's used in the tooltip
```
$maxRage=${$190411s1+($190411s2*$190411s3)}
```

src/server/scripts/Spells/spell_warrior.cpp
Golrag: I think I'd rather thave this be the same as in titanic rage script for now, and when a refactor happens to extract the code in own function, we could add a helper that also checks the aura effect and misc value

src/server/scripts/Spells/spell_warrior.cpp
Nyr97: ohh that's interestinng, i missed it

src/server/scripts/Spells/spell_warrior.cpp
mdX7: "let me hit -2 targets"

mdX7 (approved):

PR #31065, opened 2025-06-11
Script/Spells: Implement Sudden Death & Juggernaut (Fury)

Golrag:
**Changes proposed:**

-  Implement Sudden Death proc handler
- Add Sudden Death proc consumption
-  Add juggernaut proc data

**Issues addressed:**

Closes: None?

**Tests performed:**

- Builds locally
- When Sudden Death procs, cast execute to see buff being consumed one charge at the time
- Juggernaut stacks go up, they should go up even if target is immune

- [ ] I'm not sure about the proc data to proc sudden death. So I left that as is


Nyr97:
Sudden Death used to have a max. proc per minutes reduced by haste (Approximately [1.7 + Haste] procs per minute), at least until early TWW, though I'm pretty sure the core already checks this implicitly by setting 0 on ProcsPerMinute field for the proc data. It might be worth asking the Warrior discord directly just in case this has changed.

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -1036,6 +1038,26 @@ class spell_warr_t3_prot_8p_bonus : public AuraScript
     }
 };
```
Nyr97: missing comment for spell name and id related

Shauren:
> Sudden Death used to have a max. proc per minutes reduced by haste (Approximately [1.7 + Haste] procs per minute), at least until early TWW, though I'm pretty sure the core already checks this implicitly by setting 0 on ProcsPerMinute field for the proc data. It might be worth asking the Warrior discord directly.

Not relevant for scripting - this is a core mechanic, handled by SpellProcsPerMinuteMod.db2

mdX7 (approved):

PR #31070, opened 2025-06-13
Scripts/Spells: Implement Ravager

Golrag:
**Changes proposed:**

-  Implement Ravager Talent

**Issues addressed:**

Closes: none

**Tests performed:**

- Builds
- Ravager does damage

- [ ] Display ID should be different for caster and non-caster. This is something missing in core (no different creature, only display id)
- [ ] The targetting is something that I left out of scope. Ravager should follow a target, not sure which one.

### DIFF
--- a/src/server/scripts/Spells/spell_warrior.cpp
+++ b/src/server/scripts/Spells/spell_warrior.cpp
@@ -70,6 +70,10 @@ enum WarriorSpells
     SPELL_WARRIOR_MORTAL_WOUNDS                     = 115804,
     SPELL_WARRIOR_POWERFUL_ENRAGE                   = 440277,
     SPELL_WARRIOR_RALLYING_CRY                      = 97463,
+    SPELL_WARRIOR_RAVAGER                           = 228920,
+    SPELL_WARRIOR_RAVAGER_RAGE_GAIN                 = 334934,
+    SPELL_WARRIOR_RAVAGER_PERIODIC_DAMAGE           = 156287,
+    SPELL_WARRIOR_RAVAGER_SUMMON                    = 227876,
     SPELL_WARRIOR_RECKLESSNESS                      = 1719,
     SPELL_WARRIOR_RUMBLING_EARTH                    = 275339,
     SPELL_WARRIOR_SHIELD_BLOCK_AURA                 = 132404,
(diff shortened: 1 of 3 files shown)


src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -1083,6 +1088,89 @@ class spell_warr_victory_rush : public SpellScript
...
+        for (Unit* summon : owner->m_Controlled)
+            if (summon->GetEntry() == static_cast<uint32>(summonedCreatureEntry))
+                summons.push_back(summon);
+
+        return summons;
+    }
+
+    void HandlePeriodic(AuraEffect const* /*auraEffect*/) const
+    {
+        std::vector<Unit*> const ravagers = FindRavagerSummons(GetTarget());
```
Nyr97, line 1446: what if the vector returns empty?

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -1083,6 +1088,89 @@ class spell_warr_victory_rush : public SpellScript
...
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_WARRIOR_RAVAGER_PERIODIC_DAMAGE })
+            && ValidateSpellEffect({ { SPELL_WARRIOR_RAVAGER_SUMMON, EFFECT_0 } });
+    }
+
+    std::vector<Unit*> FindRavagerSummons(Unit const* owner) const
+    {
+        std::vector<Unit*> summons;
+        int32 const summonedCreatureEntry = sSpellMgr->GetSpellInfo(SPELL_WARRIOR_RAVAGER_SUMMON, GetCastDifficulty())->GetEffect(EFFECT_0).MiscValue;
```
Nyr97, line 1436: why int, a valid creature is never under 1, you can static cast this instead and keep the type as unsigned if you want to avoid the warning about type change due to the getter type

src/server/scripts/Spells/spell_warrior.cpp
```diff
@@ -1083,6 +1088,89 @@ class spell_warr_victory_rush : public SpellScript
...
+class spell_warr_ravager : public SpellScript
+{
+    bool Validate(SpellInfo const* /*spellInfo*/) override
+    {
+        return ValidateSpellInfo({ SPELL_WARRIOR_RAVAGER_SUMMON });
+    }
+
+    void SummonRavager(SpellEffIndex /*effIndex*/) const
+    {
+        GetCaster()->CastSpell(*GetHitDest(), SPELL_WARRIOR_RAVAGER_SUMMON, CastSpellExtraArgsInit{
```
Nyr97, line 1412: it's generally better to avoid dereferencing GetHitDest() with *, it creates a copy and can lead to precision loss (especially with position data). 

instead, prefer accessing the position directly via GetHitDest()->GetPosition(), this helps ensure more accurate and consistent behavior.

